Issue #2768, opened 2019-12-05
Deprecate warning for some Python and CUDA codes


leofang:
See, for example, https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=101105

Python:
- `PyThread_get_key_value` (observed starting 3.7)
- `PyThread_set_key_value` (observed starting 3.7)
- `PyThread_create_key` (observed starting 3.7)

CUDA:
- `cuParamSetTexRef` (based on the doc this was already marked deprecated at least starting 8.0)
- `cudaPointerAttributes::isManaged`  (observed starting 9.2, but the doc says 10.0)
- `cudaPointerAttributes::memoryType` (observed starting 9.2, but the doc says 10.0)

niboshi:
Thank you for raising an issue!
Just a pointer to a related issue: #1940

leofang:
Thanks, @niboshi! I knew I've seen this raised somewhere!

leofang:
I think I added `cuParamSetTexRef` in #2432 but didn't really use it 😅

leofang:
With #2769 and #2770 merged, now we are left with the warnings from `cudaPointerAttributes`.

leofang:
> * `cudaPointerAttributes::isManaged`  (observed starting 9.2, but the doc says 10.0)
> * `cudaPointerAttributes::memoryType` (observed starting 9.2, but the doc says 10.0)

As far as I can tell, these attributes are never used in the codebase. How about just removing them entirely so as to suppress the deprecation warning? We're gonna remove them when they are removed from CUDA anyway.

PR #2769, opened 2019-12-05
Avoid using PyThread API

### LINKED ISSUE #1940: Avoid using PyThread TLS API
It's deprecated in favor of TSS ([PEP-539](https://www.python.org/dev/peps/pep-0539/)), and also it seems PyThread is considered as CPython's internal API, according to PEP-539.

niboshi:
Fixes #1940
Part of #2768

### DIFF
--- a/cupy/cuda/memory_hook.pyx
+++ b/cupy/cuda/memory_hook.pyx
@@ -1,22 +1,35 @@
 import collections
-from cpython cimport pythread
 import threading
 
-cdef thread_local = threading.local()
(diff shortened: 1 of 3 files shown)


emcastillo:
Jenkins, test this please

emcastillo:
CI failure is unrelated, but is it safe to merge?

niboshi:
>AttributeError: 'NoneType' object has no attribute 'ndim'

This one could be unrelated, but the test also suffers from timeout. That should not be ignored.

emcastillo:
Timeout seems to be due to MPI not correctly closing the state

PR #2712, opened 2019-11-25
Skip tests that segfault when using SciPy 1.3.x

grlee77:
Avoid two tests that segfault within SciPy when it is version 1.3.* (or 1.4.0rc1). See scipy/scipy#11125.

Hopefully this will be fixed prior to the final 1.4.0 release.

PR #2776, opened 2019-12-06
update slicing of CSR and CSC matrices for compatibility with SciPy 1.4.0

grlee77:
This is a follow up to #2712 that updates CSR and CSC sparse matrix slicing tests for 1.4.0. The prior PR just skipped the existing tests on 1.3.x to avoid segfaults. 

In SciPy 1.4.0, the behavior for this tests case is changed from raising an `IndexError` to instead return an empty slice. [For reference see this comment at at SciPy](https://github.com/scipy/scipy/pull/11127#issuecomment-560936956) (and the implementation in scipy/scipy#11166). The new behavior is consistent with slicing of NumPy arrays/matrices. 

The PR corresponding to this behavior is being backported to the 1.4.0 branch for inclusion prior to the upcoming 1.4.0 release (see scipy/scipy#11161). For the moment I have just verified that the new test cases pass vs. current SciPy master.

### DIFF
--- a/cupyx/scipy/sparse/compressed.py
+++ b/cupyx/scipy/sparse/compressed.py
@@ -285,7 +285,8 @@ class _compressed_sparse_matrix(sparse_data._data_matrix):
             raise ValueError('slicing with step != 1 not supported')
 
         if not (major_start <= major_stop):
-            raise IndexError('index out of bounds')
+            # will give an empty slice, but preserve shape on the other axis
+            major_start = major_stop
 
         start = self.indptr[major_start]
         stop = self.indptr[major_stop]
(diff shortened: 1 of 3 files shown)


takagi:
pfnCI, test this please.

grlee77:
I forgot that the existing `test_getitem_slice_start_larger_than_stop` are no longer going to pass since the old SciPy behavior of raising an error was considered a bug by the SciPy team. This PR fixed that on the CuPy end so we can no longer compare errors to SciPy in this case.

Should I just remove the old `test_getitem_slice_start_larger_than_stop` test cases and only run these tests for SciPy >= 1.4.0?

grlee77:
In other words replace the existing `test_getitem_slice_start_larger_than_stop` with the new `test_getitem_slice_start_larger_than_stop2` (removing the 2 from the name)

takagi:
Yes, I also think testing for SciPy<1.4.0 makes no sense here and it is enough for running these tests for SciPy>=1.4.0.

takagi:
pfnCI, test this please.

takagi:
LGTM!

PR #2782, opened 2019-12-08
Support loading PTX in `cupy.RawModule`

leofang:
Part of #2779. 

This PR is simply to demonstrate that it is very easy to support PTX in `cupy.RawModule`. It's basically a one-line change. 

While I do not have any opinion in whether to support this or not, I did have this possibility in mind when adding `cupy.RawModule` in #2389, and recently I heard there exists people writing PTX directly instead of in CUDA C/C++, despite it is a quite unusual and inefficient approach IMHO. 

I'll let the core devs make the decision for the PTX support.

PR #2784, opened 2019-12-09
Separate code and path arguments in `RawModule`

niboshi:
Current API can't handle a code with a comment ending with `.cubin` at the end.

### DIFF
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -217,7 +217,8 @@ cdef class RawModule:
     providing its path, and kernels therein can be retrieved similarly.
 
     Args:
-        code_or_path (str): CUDA source code or path to cubin.
+        code (str): CUDA source code. Mutually exclusive with ``path``.
+        path (str): Path to cubin. Mutually exclusive with ``code``.
         options (tuple of str): Compiler options passed to the backend (NVRTC
             or NVCC). For details, see
             https://docs.nvidia.com/cuda/nvrtc/index.html#group__options or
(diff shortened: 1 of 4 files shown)


niboshi:
Or maybe should we have separate interfaces for loading from code and file, instead of a single `__init__`?
E.g. `json.load` and `json.loads`.

emcastillo:
I think the current PR solution is good enough. Adding a different interface would make the initialization process different from `ElementwiseKernel` and `ReductionKernel` and I think is better for these clases to share a similar schema.

emcastillo:
Jenkins, test this please

emcastillo:
Needs to fix the texture tests too 😢

emcastillo:
#2782 has been merged just now, we will need to rebase ...

niboshi:
Fixed. PTAL

emcastillo:
Jenkins, test this please

leofang:
I think the change is nice as it makes the support of different kinds of valid inputs more robust. Thank you @niboshi. I am just curious about the motivation:

> Current API can't handle a code with a comment ending with `.cubin` at the end.

Who would be bothered by this? 🤣

niboshi:
Yes, it may be unlikely, but still possible. Libraries like CuPy can't ignore such possibilities.

cupy/core/raw.pyx
```diff
@@ -218,7 +218,8 @@ cdef class RawModule:
     similarly.
 
     Args:
-        code_or_path (str): CUDA source code or path to cubin/ptx.
+        code (str): CUDA source code. Mutually exclusive with ``path``.
+        path (str): Path to cubin/ptx. Mutually exclusive with ``code``.
```
leofang: Sorry @niboshi, one more question: are `code` and `path` really mutually exclusive here? Based on the changes I thought a user could specify both `code` and `path` without getting an error?

leofang: Never mind, sorry for my stupidity 😔

niboshi: Sorry, I forgot to respond. Yes, I think they are mutually exclusive.

Issue #2569, opened 2019-10-25
Support CUDA 10.1 Update 2


kmaehashi:
Currently CuPy does not work with CUDA 10.1 Update 2. It should be resolved by the final release of v7.

* Following tests are failing (
    * cupy_tests.fft_tests.test_fft.TestRfftn_param_11.test_irfftn
    * tests.cupy_tests.linalg_tests.test_product.TestMatrixPower.test_matrix_power_inv2

* #2542 broke compatibility with CUDA 10.1 Update 2. Some undocumented APIs (e.g.,  `cusparseCbsrilu02_bufferSizeExt`) are dropped in CUDA 10.1 Update 2. As they are still declared in `cusparse.h` (this is a bug of cuSPARSE), build succeeds but `import cupy` fails. (PR #2573)
```
CUDA 10.1.105-1: ok
CUDA 10.1.168-1: ok
CUDA 10.1.243-1: fail (cuSPARSE 10.3.0.243)
```

* CUDA 10.1 Update 2 now uses semantic versioning, it is better to print version of every library (cusolver, cusparse, ...) currently linked to CuPy in `cupy.show_config`. (PR #2578)
* CUDA 10.1 Update 2 changed FP16 headers (PR #2592)

jakirkham:
So this an issue with the current release or just `master`?

kmaehashi:
v7.0.0rc1 release and current `master` branches are affected.

anaruse:
> #2542 broke compatibility with CUDA 10.1 Update 2. Some undocumented APIs

Sorry, this was my mistake. I was using CUDA 10.1 Update 1 at that time and did not notice this. I checked that some undocumented APIs like `cusparseCbsrilu02_bufferSizeExt` are actually removed from libcusparse.so.10.3.0.243, though those are still in libcusparse_static.a.

kmaehashi:
Test results:

> cupy_tests.fft_tests.test_fft.TestRfftn_param_11.test_irfftn
https://jenkins.preferred.jp/job/chainer/job/chainer-test_pr/70/label=mn1-p100,test=cupy-combination/testReport/junit/tests.cupy_tests.fft_tests.test_fft/TestRfftn_param_11/test_irfftn/

> tests.cupy_tests.linalg_tests.test_product.TestMatrixPower.test_matrix_power_inv2
[...]

PR #2788, opened 2019-12-10
Use more stable input to test `linalg.matrix_power`

toslunar:
The issue is found in #2569.

The PR fixes:
- `a ** 2` is less accurate than `a * a`;
- condition number of `[[1, 4, 9], [16, 25, 36], [49, 64, 81]]` is somewhat large.

### DIFF
--- a/tests/cupy_tests/linalg_tests/test_product.py
+++ b/tests/cupy_tests/linalg_tests/test_product.py
@@ -431,19 +431,22 @@ class TestMatrixPower(unittest.TestCase):
     @testing.for_float_dtypes(no_float16=True)
     @testing.numpy_cupy_allclose(rtol=1e-5)
     def test_matrix_power_inv1(self, xp, dtype):
-        a = testing.shaped_arange((3, 3), xp, dtype) ** 2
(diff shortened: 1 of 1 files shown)


niboshi:
Jenkins, test this please

emcastillo:
It is out of the scope of this PR, but should we try to run a preconditioner in the inputs of linalg routines?

niboshi:
@emcastillo I'm not so faimiliar with preconditioning. What would you suggest, more specifically?

emcastillo:
It is used for solving systems with iterative methods such as CG. 
When the matrix is ill-conditioned (it has a large condition number) successive iterations accumulate the precision errors resulting in wrong solutions.

A Preconditioner transforms the system into an equivalent one but with a small condition number.

niboshi:
@emcastillo I understand that to that extent.
What would you suggst in this particular case?

emcastillo:
This case is fine with the current fix, but for some other routines, we could try to generate the inputs with a condition number less than a threshold (especially iterative solvers such as cholesky, lu, etc.).

niboshi:
Why not test on smaller matrix in the first place?

PR #2766, opened 2019-12-05
Do not let Python 2 users build CuPy v7+

leofang:
Close #2763. 

cc: @JanuszL

PR #2794, opened 2019-12-11
Reomve Python 3.4 matrix from Travis CI

niboshi:
Python 3.4 is no longer supported.


leofang:
This fixes https://github.com/cupy/cupy/pull/2766#issuecomment-562226721.

kmaehashi:
LGTM!

PR #2798, opened 2019-12-11
Document device synchronization

### LINKED ISSUE #2797: Document functions that may synchronize the device
Some functions (e.g. `nonzero`) inevitably synchronize the device.
That should be documented.

niboshi:
Fixes #2797

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -711,7 +711,6 @@ cdef class ndarray:
         .. warning::
 
             This function may synchronize the device.
-            Avoid using it in a performance-critical code.
 
         .. seealso::
             :func:`numpy.nonzero`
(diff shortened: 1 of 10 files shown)


cupy/core/core.pyx
```diff
@@ -708,6 +708,11 @@ cdef class ndarray:
         Returns:
             tuple of arrays: Indices of elements that are non-zero.
 
+        .. warning::
+
+            This function may synchronize the device.
+            Avoid using it in a performance-critical code.
```
asi1024: I feel we don't want to suggest to avoid using these functions without an alternative solution. How about removing this phrase and just saying the fact that these functions synchronize the device?

niboshi: Done in f22fe440bd66cefdc9d941685eaed54cafc2e77b.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #2799, opened 2019-12-11
Avoid device synchronization in `cupy.allclose`

niboshi:
Currently it synchronizes the device and returns a `bool`, but I think it can just return a 0-dim array.

### DIFF
--- a/cupy/logic/comparison.py
+++ b/cupy/logic/comparison.py
@@ -57,14 +57,14 @@ def allclose(a, b, rtol=1.e-5, atol=1.e-8, equal_nan=False):
             to NaN's in ``b``.
 
     Returns:
-        bool: if ``True``, two arrays are element-wise equal within a
-            tolerance.
+        cupy.ndarray: A boolean 0-dim array.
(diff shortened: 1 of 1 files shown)


leofang:
I guess it’s better to document this NumPy non-compatible behavior somewhere?

asi1024:
@leofang It is documented [here](https://docs-cupy.chainer.org/en/stable/reference/difference.html#reduction-methods).

leofang:
Thanks, @asi1024, but I don’t think this is clear for `allclose`. Off my head I would not know `allclose` also belongs to the reduction methods, despite it is.

niboshi:
I agree with @leofang in that it's difficult for users to realize `allclose` is a reduction.
But it could be the same for other functions, including `any` for example.
How about separately discussing how to document it in wider scope?

leofang:
We could consider adding @asi1024's link to `allclose`'s (and perhaps `any`'s?) docstring. It can be in a separate PR if you think appropriate.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #2611, opened 2019-11-06
Fix `__cuda_array_interface__` data pointer for 0-size arrays

leofang:
Close #2610.

PR #2802, opened 2019-12-12
Avoid looking up null pointers' attributes

### LINKED ISSUE #2789: `__cuda_array_interface__ ` with `NULL` pointer fails to convert to CuPy array
Recently ran into an issue with CuPy's `__cuda_array_interface__` conversion support where it fails if it holds a `NULL` pointer for the data. Here's an example using an existing CuPy array. Though this happens with any `__cuda_array_interface__` supporting object that holds a `NULL` pointer. 

  - CuPy version: 6.6.0
  - OS/Platform: Ubuntu 18.04, x86_64
  - CUDA version: 10.0
  - cuDNN/NCCL version (if applicable): 7.6.4/2.4.6
* Code to reproduce

```python
import cupy                                                             

a = cupy.empty((0,), dtype="|u1")                                       
a2 = cupy.core.core._convert_object_with_cuda_array_interface(a)
```

```python
CUDARuntimeError                          Traceback (most recent call last)
<ipython-input-3-cf9277f8c1fb> in <module>
----> 1 a2 = cupy.core.core._convert_object_with_cuda_array_interface(a)

cupy/core/core.pyx in cupy.core.core._convert_object_with_cuda_array_interface()

cupy/core/core.pyx in cupy.core.core._convert_object_with_cuda_array_interface()

cupy/cuda/memory.pyx in cupy.cuda.memory.UnownedMemory.__init__()

cupy/cuda/runtime.pyx in cupy.cuda.runtime.pointerGetAttributes()

cupy/cuda/runtime.pyx in cupy.cuda.runtime.check_status()

CUDARuntimeError: cudaErrorInvalidValue: invalid argument
```

leofang:
Follow-up of #2611. Close #2789.

This PR fixes the consumer side (while #2611 fixed the producer side) for handling zero-size arrays. 

cc: @jakirkham

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -2776,7 +2776,7 @@ not_equal = create_comparison(
 
 cpdef ndarray _convert_object_with_cuda_array_interface(a):
     cdef Py_ssize_t sh, st
-    cdef dict desc = a.__cuda_array_interface__
+    cdef object desc = a.__cuda_array_interface__
     cdef tuple shape = desc['shape']
     cdef int dev_id = -1
     cdef int nbytes
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -2776,13 +2776,16 @@ not_equal = create_comparison(
 
 cpdef ndarray _convert_object_with_cuda_array_interface(a):
     cdef Py_ssize_t sh, st
-    desc = a.__cuda_array_interface__
-    shape = desc['shape']
+    cdef dict desc = a.__cuda_array_interface__
+    cdef tuple shape = desc['shape']
+    cdef int dev_id = -1
+    cdef int nbytes
+
+    ptr = desc['data'][0]
     dtype = numpy.dtype(desc['typestr'])
-    if 'mask' in desc:
-        mask = desc['mask']
-        if mask is not None:
-            raise ValueError('CuPy currently does not support masked arrays.')
+    mask = desc.get('mask')
+    if mask is not None:
+        raise ValueError('CuPy currently does not support masked arrays.')
     strides = desc.get('strides')
     if strides is not None:
         nbytes = 0
@@ -2790,7 +2793,11 @@ cpdef ndarray _convert_object_with_cuda_array_interface(a):
             nbytes = max(nbytes, abs(sh * st))
     else:
         nbytes = internal.prod_sequence(shape) * dtype.itemsize
-    mem = memory_module.UnownedMemory(desc['data'][0], nbytes, a)
+    # the v2 protocol sets ptr=0 for 0-size arrays, so we can't look up
+    # the pointer attributes and must use the current device
+    if nbytes == 0:
+        dev_id = device.get_device_id()
+    mem = memory_module.UnownedMemory(ptr, nbytes, a, dev_id)
     memptr = memory.MemoryPointer(mem, 0)
     return ndarray(shape, dtype, memptr, strides)
 
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -129,7 +129,7 @@ cdef class UnownedMemory(BaseMemory):
                 ptr_attrs = runtime.pointerGetAttributes(ptr)
                 device_id = ptr_attrs.device
             else:
-                raise RuntimeError
+                raise RuntimeError('UnownedMemory requires explicit device ID for a null pointer.')
         self.size = size
         self.device_id = device_id
         self.ptr = ptr
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -125,11 +125,11 @@ cdef class UnownedMemory(BaseMemory):
         # ptr=0 for 0-size arrays from __cuda_array_interface__ v2:
         # we need a valid device id as null ptr can't be looked up
         if device_id < 0:
-            if ptr > 0:
-                ptr_attrs = runtime.pointerGetAttributes(ptr)
-                device_id = ptr_attrs.device
-            else:
-                raise RuntimeError('UnownedMemory requires explicit device ID for a null pointer.')
+            if ptr == 0:
+                raise RuntimeError('UnownedMemory requires explicit'
+                                   ' device ID for a null pointer.')
+            ptr_attrs = runtime.pointerGetAttributes(ptr)
+            device_id = ptr_attrs.device
         self.size = size
         self.device_id = device_id
         self.ptr = ptr
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -122,7 +122,8 @@ cdef class UnownedMemory(BaseMemory):
     def __init__(self, intptr_t ptr, size_t size, object owner,
                  int device_id=-1):
         cdef runtime.PointerAttributes ptr_attrs
-        if device_id < 0:
+        # ptr=0 for 0-size arrays from __cuda_array_interface__ v2 
+        if device_id < 0 and ptr > 0:
             ptr_attrs = runtime.pointerGetAttributes(ptr)
             device_id = ptr_attrs.device
         self.size = size
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -122,7 +122,7 @@ cdef class UnownedMemory(BaseMemory):
     def __init__(self, intptr_t ptr, size_t size, object owner,
                  int device_id=-1):
         cdef runtime.PointerAttributes ptr_attrs
-        # ptr=0 for 0-size arrays from __cuda_array_interface__ v2 
+        # ptr=0 for 0-size arrays from __cuda_array_interface__ v2
         if device_id < 0 and ptr > 0:
             ptr_attrs = runtime.pointerGetAttributes(ptr)
             device_id = ptr_attrs.device
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -122,10 +122,14 @@ cdef class UnownedMemory(BaseMemory):
     def __init__(self, intptr_t ptr, size_t size, object owner,
                  int device_id=-1):
         cdef runtime.PointerAttributes ptr_attrs
-        # ptr=0 for 0-size arrays from __cuda_array_interface__ v2
-        if device_id < 0 and ptr > 0:
-            ptr_attrs = runtime.pointerGetAttributes(ptr)
-            device_id = ptr_attrs.device
+        # ptr=0 for 0-size arrays from __cuda_array_interface__ v2:
+        # we need a valid device id as null ptr can't be looked up
+        if device_id < 0:
+            if ptr > 0:
+                ptr_attrs = runtime.pointerGetAttributes(ptr)
+                device_id = ptr_attrs.device
+            else:
+                raise RuntimeError
         self.size = size
         self.device_id = device_id
         self.ptr = ptr
--- a/tests/cupy_tests/creation_tests/test_from_data.py
+++ b/tests/cupy_tests/creation_tests/test_from_data.py
@@ -489,6 +489,16 @@ class TestCudaArrayInterface(unittest.TestCase):
         assert a.strides == b.strides
         assert a.nbytes == b.data.mem.size
 
+    @testing.for_all_dtypes()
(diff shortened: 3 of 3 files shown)


niboshi:
It seems currently `BaseMemory`, the base class of `UnownedMemory`, doen't assume negative `device_id`.
I think we should choose from either of the following:
1. Fix `BaseMemory` to allow unspecified device.
    * make `BaseMemory.device_id` a property.
    * If the device is not specified, `BaseMemory.device/device_id` should raise error if the pointer is null. If it is non-null, they can use `runtime.pointerGetAttributes` to retrieve the device.
2. Fix `UnownedMemory` to allow null pointers only if the device is specified.
    * Raise error if the pointer is null and the device is unspecified.
    * In this PR, `_convert_object_with_cuda_array_interface()` should either avoid using `UnownedMemory` or pass it a concrete device ID, if the pointer is null.

leofang:
Thanks, @niboshi, for your feedback! I think No.2 is better in that we won't touch the base that might interfere with other functionalities. Moreover, `_convert_object_with_cuda_array_interface()` should still use `UnownedMemory`, otherwise no function in CuPy would use it. In that case, the concrete device ID should be the current device when the cuda array interface protocol is executed. 

If you think this is reasonable, I will make changes. Thanks.

leofang:
@niboshi May I move toward your No.2 suggestion?

niboshi:
Sorry for delay.
Yes, please go with 2.

leofang:
Thanks, @niboshi. PTAL.

cupy/core/core.pyx
```diff
@@ -2776,21 +2776,28 @@ not_equal = create_comparison(
 
 cpdef ndarray _convert_object_with_cuda_array_interface(a):
     cdef Py_ssize_t sh, st
-    desc = a.__cuda_array_interface__
-    shape = desc['shape']
+    cdef dict desc = a.__cuda_array_interface__
```
niboshi: I think it's not safe to assume it's `dict`.
In Python you can define arbitrary type that behaves as a dictionary.

cupy/cuda/memory.pyx
```diff
@@ -122,9 +122,14 @@ cdef class UnownedMemory(BaseMemory):
...
+        # ptr=0 for 0-size arrays from __cuda_array_interface__ v2:
+        # we need a valid device id as null ptr can't be looked up
         if device_id < 0:
-            ptr_attrs = runtime.pointerGetAttributes(ptr)
-            device_id = ptr_attrs.device
+            if ptr > 0:
+                ptr_attrs = runtime.pointerGetAttributes(ptr)
+                device_id = ptr_attrs.device
+            else:
+                raise RuntimeError
```
niboshi: Let's add a message.

```suggestion
                raise RuntimeError('UnownedMemory requires explicit device ID for a null pointer.')
```

cupy/cuda/memory.pyx
```diff
@@ -122,9 +122,14 @@ cdef class UnownedMemory(BaseMemory):
     def __init__(self, intptr_t ptr, size_t size, object owner,
                  int device_id=-1):
         cdef runtime.PointerAttributes ptr_attrs
+        # ptr=0 for 0-size arrays from __cuda_array_interface__ v2:
+        # we need a valid device id as null ptr can't be looked up
         if device_id < 0:
-            ptr_attrs = runtime.pointerGetAttributes(ptr)
-            device_id = ptr_attrs.device
+            if ptr > 0:
```
niboshi: Please use `==`. `ptr` can be negative.
Also, it's better to indent the code this way:

```py
            if ptr == 0:
                raise RuntimeError('UnownedMemory requires explicit device ID for a null pointer.')
            ptr_attrs = runtime.pointerGetAttributes(ptr)
            device_id = ptr_attrs.device
```

cupy/core/core.pyx
leofang: Before I proposed the v2 protocol, I investigated and ensured all current supporting libraries (see numba/numba#4448) implemented it as a simple Python dict, so that we could safely require it to be a dict in v2 without any backward incompatibility. I believe this is a very safe and compliant change.

cupy/cuda/memory.pyx
leofang: Good point, thanks.

leofang:
@niboshi, PTAL, thanks!

cupy/core/core.pyx
niboshi: Do you mean v2 already requires it to be a `dict`?
In the specification (https://numba.pydata.org/numba-doc/dev/cuda/cuda_array_interface.html), I can only find a statement that it returns a "dictionary" which is broader than `dict`.

In any case, this is the receiving side. It may be safe not to unnecessarily assume it (even if the protocol specifies it).

leofang: > I can only find a statement that it returns a "dictionary" which is broader than dict.

That is not true. Dictionaries in Python almost always refer to the built-in `dict`. There is no ambiguity for alternative interpretation. Also, the wording and the implementation follow that of `__array_interface__` in NumPy.

I revised the wording in v2 in the hope that it can be used as a small optimization in Cython (both CuPy and mpi4py have the receiver code written in Cython), so I would very much like to keep the change as is.

niboshi: >Dictionaries in Python almost always refer to the built-in `dict`.

I can't find any (quasi-)official statement about that.
[The official glossary](https://docs.python.org/3/glossary.html#term-dictionary) describes "dictionary" as "An associative array, where arbitrary keys are mapped to values.". It's still ambiguous, I think.

Python is a duck-typing language. Without explicit wording, I wouldn't assume "dictionary" refers to `dict`.

As for performance, I did some measurement using the following code. There was no noticeable difference whether or not `d` was typed (actually it seems typed code was a bit slower, perhaps because of extra type checking).

```py
# distutils: language = c++
cpdef int func(dict d):
    cdef int s = 0
    cdef int t = 0
    for i in range(100):
        t = d['key']
        s += t
    return s
```

Also, for the purpose of optimization, an implementation of `__cuda_array_interface__` could return a specialized and faster variant of dictionary with limited set of keys, for example. Explicit typing could potentially cause problems in such a case.

leofang: > I can't find any (quasi-)official statement about that.
> [The official glossary](https://docs.python.org/3/glossary.html#term-dictionary) describes "dictionary" as "An associative array, where arbitrary keys are mapped to values.". It's still ambiguous, I think.
> 
> Python is a duck-typing language. Without explicit wording, I wouldn't assume "dictionary" refers to `dict`.

> Also, for the purpose of optimization, an implementation of `__cuda_array_interface__` could return a specialized and faster variant of dictionary with limited set of keys, for example. Explicit typing could potentially cause problems in such a case.

I disagree. The terms "dictionary" and `dict` are always used interchangeably, see, e.g., any place in the official Python documentation that refers to dictionaries:
https://docs.python.org/3.7/library/stdtypes.html?highlight=dictionary#mapping-types-dict

In practice, having any alternative dictionary implementation for `__cuda_array_interface__` can only increase the complexity of code dependency and the risk of not being interoperable. It makes little sense to be overcautious on this.


 
> As for performance, I did some measurement using the following code. There was no noticeable difference whether or not `d` was typed (actually it seems typed code was a bit slower, perhaps because of extra type checking).
> 
> ```python
> # distutils: language = c++
> cpdef int func(dict d):
>     cdef int s = 0
>     cdef int t = 0
>     for i in range(100):
>         t = d['key']
>         s += t
>     return s
> ```
> 

This is what I got:
```python
In [1]: %load_ext Cython

In [2]: class Array:
   ...:     def __init__(self):
   ...:         self.__interface__ = {'shape': (1, 2, 3), 'ptr':18}
   ...:         

In [3]: arr = Array()

In [4]: %%cython
   ...: cpdef get_face(a):
   ...:     x = a.__interface__
   ...:     shape = x['shape']
   ...:     ptr = x['ptr']
   ...:     return ptr
   ...: 

In [5]: %%cython
    ...: cpdef get_face_typed(a):
    ...:     cdef dict x = a.__interface__
    ...:     cdef tuple shape = x['shape']
    ...:     ptr = x['ptr']
    ...:     return ptr
    ...: 

In [6]: %timeit get_face(arr)
107 ns ± 0.312 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: %timeit get_face_typed(arr)
101 ns ± 0.212 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

leofang: FYI, see the original intent of the v2 protocol for `__cuda_array_interface__` being a Python `dict`: https://github.com/numba/numba/pull/4609#issue-321035175.

niboshi: >I disagree. The terms "dictionary" and `dict` are always used interchangeably, see, e.g., any place in the official Python documentation that refers to dictionaries:
https://docs.python.org/3.7/library/stdtypes.html?highlight=dictionary#mapping-types-dict

Thank you for the pointer.
I see, maybe you're correct in that the official documentation refers to `dict` by the word "dictionary".
Maybe officially there's no ambiguity, but in reality, I think there is.
Even I, who have been using Python for 8~9 years, mistook the word, because the primary mode of the language is duck-typing and this philosophy is everywhere.

Anyway, this is about the expression in the protocol specification, and I think not essential in this discussion. As I said above, even if the protocol explicitly specifies the type to be `dict`, I don't think the receiver should assume that.

>This is what I got:

The result looks the same as mine. Looks like explicit typing has no positive impact on performance.

>In practice, having any alternative dictionary implementation for `__cuda_array_interface__` can only increase the complexity of code dependency and the risk of not being interoperable. It makes little sense to be overcautious on this.

I agree it could increase the risk of not being interoperable (not sure about code dependency). 
But it's about the producer side. Receiver side would increase the risk by making an assumption, instead.

leofang: @niboshi, I thought you replied something but I can't find it now. Anyway, I am surprised about the pushback here, to end this and land the fix asap I changed the type in 230e83a.

niboshi: Thank you for understanding.
This is what I posted (modified a bit):

>>I disagree. The terms "dictionary" and `dict` are always used interchangeably, see, e.g., any place in the official Python documentation that refers to dictionaries:
>https://docs.python.org/3.7/library/stdtypes.html?highlight=dictionary#mapping-types-dict
>
>Thank you for the pointer.
>I see, maybe you're correct in that the official documentation refers to `dict` by the word "dictionary".
>Maybe officially there's no ambiguity, but in reality, I think there is.
>Even I, who have been using Python for 8~9 years, mistook the word, because the primary mode of the language is duck-typing and this philosophy is everywhere.
>
>Anyway, this is about the expression in the protocol specification, and I think not essential in this discussion. As I said above, even if the protocol explicitly specifies the type to be `dict`, I don't think the receiver should assume that.
>
>>This is what I got:
>
>I see. it might be dependent on environment. But anyway I think a few nanosecond is not so important.
> 
>
>>In practice, having any alternative dictionary implementation for `__cuda_array_interface__` can only increase the complexity of code dependency and the risk of not being interoperable. It makes little sense to be overcautious on this.
>
>I agree it could increase the risk of not being interoperable (not sure about code dependency). 
>But it's about the producer side. Receiver side would increase the risk by making an assumption, instead.

niboshi:
Jenkins, test this please

niboshi:
LGTM

jakirkham:
Thanks all! 😀

cupy/core/core.pyx
leofang: Thanks, @niboshi. I have only one comment:

> > In practice, having any alternative dictionary implementation for `__cuda_array_interface__` can only increase the complexity of code dependency and the risk of not being interoperable. It makes little sense to be overcautious on this.
> 
> I agree it could increase the risk of not being interoperable (not sure about code dependency).
> But it's about the producer side. Receiver side would increase the risk by making an assumption, instead.

I still think it is a right thing to do on the receiver side to reject any non-standard `dict` alternatives. First, this is specified in the protocol as you also agreed. Second, the receiver does not (and should not) recognize the type of any non-standard dictionary, or the receiving library would have to also depend on the library that provides the custom dictionary.

jakirkham: Have submitted PR ( https://github.com/numba/numba/pull/5049 ) in an attempt to clarify this in the spec. Feedback welcome 🙂

PR #2805, opened 2019-12-12
fix range used in cupyx.scipy.ndimage filter origin check

grlee77:
This PR makes the allowed settings for origin match SciPy's behavior.

The actual origin validation code in current scipy master is
```Python
(origin < -(lenw // 2)) or (origin > (lenw - 1) // 2)
```
I just made the minimal needed change here, but can change to match the style above exactly if you prefer that

This fix corresponds to the following PR merged for SciPy 1.2: scipy/scipy#8817. It seems older versions might segfault for some invalid `origin` settings.


asi1024:
Thanks for your PR! Jenkins, test this please.

asi1024:
LGTM!

Issue #2777, opened 2019-12-06
CUDA stream errors


xiaodong-yu:
When I play with the CUDA stream APIs in cupy, I find even running the official examples (cupy/examples/stream/cufft.py) returns the following errors:
Exception ignored in: <bound method Stream.__del__ of <cupy.cuda.stream.Stream object at 0x7fa0f7366050>>
Traceback (most recent call last):
  File "cupy/cuda/stream.pyx", line 168, in cupy.cuda.stream.Stream.__del__
  File "cupy/cuda/stream.pyx", line 50, in cupy.cuda.stream._set_current_stream
AttributeError: 'NoneType' object has no attribute 'ref'

I believe I installed cupy 6.5 for my CUDA version (10.0), anybody knows what causes this error? Thank you so much in advance!

leofang:
Possibly the same problem: #2658 

@emcastillo I was not alone! Could it be some modules have a hidden stream added for some reason, and that stream was deconstructed too late (after `cupy.cuda.runtime` is de-imported)?

emcastillo:
I am not sure but maybe https://github.com/cupy/cupy/pull/2769/ can fix this.

The problem seems to be that part of the thread-local storage gets blown up before the stream itself is destructed.

leofang:
> I am not sure but maybe #2769 can fix this.

Just tested the master branch on the machine that I observed the #2658 symptom, and it still happens ☹️

emcastillo:
The problem is object destruction order here, the globals are not guaranteed to be alive when a object `__del__` is called. I think that just wrapping it into a try/catch should suffice

leofang:
I wonder why it is not reproducible everywhere...

leofang:
@xiaodong-yu could you confirm that the error you saw happened at interpreter shutdown, i.e., the script itself ran correctly?

niboshi:
#2809

Issue #2658, opened 2019-11-20
Is `cupy.cuda.runtime` set to None too early during interpreter shutdown?


leofang:
* Conditions (you can just paste the output of `python -c 'import cupy; cupy.show_config()'`)
```shell
$ python -c 'import cupy; cupy.show_config()'
CuPy Version          : 7.0.0rc1
CUDA Root             : /usr/local/cuda
CUDA Build Version    : 10000
CUDA Driver Version   : 10000
CUDA Runtime Version  : 10000
cuDNN Build Version   : 7604
cuDNN Version         : 7604
NCCL Build Version    : None
NCCL Runtime Version  : None
```
Tested with a fresh build on the latest master in a fresh conda environment.

* Code to reproduce
```python
>>> import cupy as cp
>>> stream = cp.cuda.Stream()
>>> # ...then exit the interpreter
>>>
Exception ignored in: <bound method Stream.__del__ of <cupy.cuda.stream.Stream object at 0x7f75830870b8>>
Traceback (most recent call last):
  File "cupy/cuda/stream.pyx", line 186, in cupy.cuda.stream.Stream.__del__
AttributeError: 'NoneType' object has no attribute 'streamDestroy'
```
I don't think I've seen this error before. Looks like `runtime` is set to `None` when `stream.__del__()` is invoked, but this is unexpected because, according to [PEP 442](https://docs.python.org/3/whatsnew/3.4.html#whatsnew-pep-442), starting at Python 3.4 the modules unloaded during interpreter shutdown are no longer set to `None`, and I'm on Python 3.6.7.

Can this be reproduced? Thanks.

emcastillo:
Couldn't reproduce with python 3.7.3

leofang:
What's the Cython version that you used? I am on 0.29.14.

emcastillo:
0.29.13

leofang:
Still see the error with Python 3.6.7 + Cython 0.29.13...Not sure what happened 😥

emcastillo:
Jave you tried python 3.7.x?

leofang:
Sorry that was a typo. I tried 3.7.3. 😅

leofang:
[...]

PR #2809, opened 2019-12-13
Detect interpreter shutdown for proper `__del__` behavior

### LINKED ISSUE #2806: `__del__` behavior on interpreter shutdown
When finishing program execution, python does not guarantee the order in which objects are destroyed.

Calls inside a `__del__` method calling other objects or modules are not guaranteed to be correctly executed.

https://github.com/cupy/cupy/blob/master/cupy/cuda/stream.pyx#L185
It is known to fail because `runtime` module is already unloaded when `__del__` is called.

https://docs.python.org/3/reference/datamodel.html#object.__del__

This can affect behavior on RAII pattern.

emcastillo:
Closes #2806, #2658, and #2777 

In some python interpreters (mostly installed with pyenv)
The following one-liner causes an exception in object destruction during interpreter shutdown.
`python -c 'import cupy; stream = cupy.cuda.Stream()'`

When finishing program execution, python does not guarantee the order in which objects are destroyed.

Calls inside a __del__ method calling other objects or modules are not guaranteed to be correctly executed as those objects can be already deleted.

 [PEP442](https://docs.python.org/3/whatsnew/3.4.html#whatsnew-pep-442) should address this, but we have seen this problem persists.

This solution is obtained from [Numba/llvmlite](https://github.com/numba/llvmlite/blob/9ddb1a52b259daa418cd33f62cfe8f4992c7a2d3/llvmlite/binding/common.py#L27-L41) code.

### DIFF
--- a/cupy/random/generator.py
+++ b/cupy/random/generator.py
@@ -51,8 +51,10 @@ class RandomState(object):
         self._generator = curand.createGenerator(method)
         self.seed(seed)
 
-    def __del__(self):
+    def __del__(self, is_shutting_down=util.is_shutting_down):
         # When createGenerator raises an error, _generator is not initialized
+        if is_shutting_down():
+            return
         if hasattr(self, '_generator'):
             curand.destroyGenerator(self._generator)
 
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -13,6 +13,7 @@ from cupy.cuda import device
 from cupy.cuda import function
 from cupy.cuda import nvrtc
 from cupy.cuda import runtime
+from cupy import util
 
 _nvrtc_version = None
 _nvrtc_max_compute_capability = None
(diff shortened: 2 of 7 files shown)


niboshi:
A nitpick but how about this way?
```py
if is_shutting_down():
    return
```
No extra indentation needed, not complicating conditions, and you can write the same way for any clean up code.

cupy/random/generator.py
```diff
@@ -51,9 +51,9 @@ def __init__(self, seed=None, method=curand.CURAND_RNG_PSEUDO_DEFAULT):
         self._generator = curand.createGenerator(method)
         self.seed(seed)
 
-    def __del__(self):
+    def __del__(self, is_shutting_down=util.is_shutting_down):
         # When createGenerator raises an error, _generator is not initialized
-        if hasattr(self, '_generator'):
+        if not util.is_shutting_down() and hasattr(self, '_generator'):
```
toslunar:
```suggestion
        if not is_shutting_down() and hasattr(self, '_generator'):
```

emcastillo:
Thanks for comments.

niboshi:
LGTM
Jenkins, test this please

PR #2811, opened 2019-12-13
[backport] Detect interpreter shutdown for proper `__del__` behavior

niboshi:
Backport of #2809

PR #2812, opened 2019-12-13
[backport] Detect interpreter shutdown for proper `__del__` behavior

niboshi:
Backport of #2809


niboshi:

emcastillo:
I think we need to backport it to v7 too 😂
let me do the pr later!

niboshi:
@emcastillo There already is one: #2811

PR #2814, opened 2019-12-14
Fix split and array_split with indices overrun

takagi:
Close #2163.

`cupy.split` and `cupy.array_split` raise some error when they take indices that have an index exceeds the target axis' dimension. This PR fix it to make them return empty arrays for such sections as NumPy does.

PR #2815, opened 2019-12-15
Fix split and array_split with unordered indices supplied

takagi:
Follows #2814 .

`cupy.split` and `cupy.array_split` raise an error when they take unordered indices as `[4, 2]`. NumPy does not require indices ordered and, in such case, it return an empty array for a going-back index. I'm not sure it is meaningful, but it actually does. This PR fix the functions to follow NumPy's behavior.

```Python
>>> import numpy as np
>>> import cupy as cp
>>> np.split(np.array([0, 1, 2, 3, 4]), [4, 2])
[array([0, 1, 2, 3]), array([], dtype=int64), array([2, 3, 4])]
>>> cp.split(cp.array([0, 1, 2, 3, 4]), [4, 2])
  File "cupy/core/core.pyx", line 1341, in cupy.core.core.ndarray.__repr__
    return repr(self.get())
  File "cupy/core/core.pyx", line 1432, in cupy.core.core.ndarray.get
    a_cpu = numpy.empty(self._shape, dtype=self.dtype, order=order)
ValueError: negative dimensions are not allowed
```

### DIFF
--- a/cupy/core/_routines_manipulation.pyx
+++ b/cupy/core/_routines_manipulation.pyx
@@ -436,7 +436,7 @@ cpdef array_split(ndarray ary, indices_or_sections, Py_ssize_t axis):
         stride = 0
     for index in indices:
         index = min(index, size)
-        shape[axis] = index - prev
+        shape[axis] = max(index - prev, 0)
         v = ary.view()
         v.data = ary.data + prev * stride
         # TODO(niboshi): Confirm update_x_contiguity flags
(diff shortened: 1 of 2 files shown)


cupy/core/_routines_manipulation.pyx
```diff
@@ -435,7 +435,8 @@ cpdef array_split(ndarray ary, indices_or_sections, Py_ssize_t axis):
     if ary.size == 0:
         stride = 0
     for index in indices:
-        shape[axis] = index - prev
```
emcastillo: I think that the way min & max are implemented in python can lead to multiple device synchronization issues here. 
Python could synchronize the device when reading the result of cupy.array.__lt__ and cupy.array.__gt__

emcastillo:
Jenkins, test this please

takagi:
As #2814 is now sucessfully back-ported, would you make the back-ports of this PR as well?

emcastillo:
Yes, I was waiting for that :)

takagi:
Thanks!

PR #2818, opened 2019-12-16
Refactor array testing decorators

niboshi:
Depends on #2820 
Refactor the common code to call cupy and cupy test implementation (`_call_func_numpy_cupy()`).

PR #2819, opened 2019-12-16
Detect synchronization

### LINKED ISSUE #2808: Detection of device synchronization
Device synchronization is a source of performance degradation, but currently even some of CuPy functions implicitly (whether intentionally or not) do that (#2797).

It would be nice to have a mechanism to detect device synchronization.

### Proposal

Add a context manager within which any operation that synchronizes the device causes an error. The context manager sets a thread-local configuration variable.

```py
with cupyx.allow_synchronize(False):
    ... # operations that are not supposed to synchronize
```

In low-level CuPy code that cause synchronization like `ndarray.get`, insert `_declare_synchronize()` call as follows:

```py
cpdef _declare_synchronize():
    ... # Raise an error if within allow_synchronize(False).

cdef class ndarray:
    cpdef get(self, ...):
        _declare_synchronize()
       ...
```

### Application in unit tests

Given the above implemented, we could apply it in `testing.numpy_cupy_equal` for example. If the test code has expected device synchronization, the test implementation can enclose the code with `cupyx.allow_synchronize(True)` explicitly.

### Notes
* `ndarray.get` may copy asynchronously if certain conditions are met (if using a non-default stream and given an explicit `out` with pinned memory). If's difficult for the code to predict whether it's actually synchronous or not. Maybe it's safe to separate asynchronous behavior as another method like `ndarray.get_async`.

niboshi:
Fixes #2808
Depends on #2818

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -19,6 +19,7 @@ from cupy.core._reduction import ReductionKernel
 from cupy.core._ufuncs import elementwise_copy
 from cupy.core._ufuncs import elementwise_copy_where
 from cupy.core import flags
+from cupy.core import syncdetect
 from cupy import cuda
 from cupy.cuda import device
 from cupy.cuda import memory as memory_module
@@ -1443,6 +1444,8 @@ cdef class ndarray:
             else:
                 a_gpu = self
             a_cpu = numpy.empty(self._shape, dtype=self.dtype, order=order)
+
+        syncdetect._declare_synchronize()
         ptr = ctypes.c_void_p(a_cpu.__array_interface__['data'][0])
         with self.device:
             if stream is not None:
--- a/cupy/core/syncdetect.py
+++ b/cupy/core/syncdetect.py
@@ -1,0 +1,54 @@
+import contextlib
+import threading
+
+
+_thread_local = threading.local()
+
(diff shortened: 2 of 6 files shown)


emcastillo:
Jenkins, test this please

niboshi:
@emcastillo  Please merge #2818 first.

emcastillo:
Jenkins, test this please

emcastillo:
#2818 was merged

niboshi:
Please merge #2818.

emcastillo:
Rebase please :)

niboshi:
PTAL.

emcastillo:
Jenkins, test this please

emcastillo:
😂I guess that we have more synchronization than we expected

niboshi:
Hmm, it's too many...
I think we should separate the change to test helpers as another PR.

emcastillo:
I agree. At first, I thought that all of them were going to be fine so it was nice to do the checks all-together.

niboshi:
Done (force-pushed).

emcastillo:
Jenkins, test this please

kmaehashi:
Discussion to remove this feature ongoing in #2946

PR #2820, opened 2019-12-16
Fix decorator usage in tests 

niboshi:


cupy/testing/helper.py
```diff
@@ -137,6 +137,10 @@ def _contains_signed_and_unsigned(kw):
 
 def _make_decorator(check_func, name, type_check, accept_error, sp_name=None,
```
emcastillo: tests use `accept_error` as kwarg, should we force the arguments to be kwarg only?

niboshi: `_make_decorator` is an internal function, and there are only a few callers. I don't think there's any reason to make it keyword-only.

niboshi: If you think that increases readability, I won't object that, though.

emcastillo: with the reduced number of cases and the unlikely addition of more, it is ok to leave it as-is. I was suggesting it only to enforce using the same style in the code base, but as you said probably it won't be used in other places.

emcastillo:
Jenkins, test this please

niboshi:
Fixed test failures.
PTAL

emcastillo:
Jenkins, test this please

niboshi:
@emcastillo Backport to v6?

emcastillo:
I am trying it, but the backport script breaks when pushing the changes ...
I am looking into it now

PR #2823, opened 2019-12-17
[WIP] Support 64 bit extent `randomint`

### LINKED ISSUE #2816: random.randint support for range larger than int32
[random.randint](https://github.com/cupy/cupy/blob/master/cupy/random/generator.py#L1122) doesn't support ranges larger than int32 apparently because [RandomState._interval](https://github.com/cupy/cupy/blob/f121d14f6d3d2f5b25a58f26d2f8bca40dd10812/cupy/random/generator.py#L631) doesn't. Is there a plan to support larger numbers such as int64 and uint64?

emcastillo:
This is a horrible hack just posted to show where the problem is and discuss a clean solution.

Closes #2816

Issues to discuss

- Is it ok to just change the random number generator when we detect we trying to generate a 64 bit extent large int?
- Should we just throw an error and ask the user to use a proper generator?

@niboshi @asi1024

### DIFF
--- a/cupy/random/generator.py
+++ b/cupy/random/generator.py
@@ -617,7 +617,8 @@ class RandomState(object):
 
         if mx == 0:
             return cupy.zeros(size, dtype=numpy.int32)
-
+        generator = self._generator
+        curand_generate = curand.generate
         if mx < 0:
             raise ValueError(
                 'mx must be non-negative (actual: {})'.format(mx))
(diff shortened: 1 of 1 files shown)


niboshi:
>Is it ok to just change the random number generator when we detect we trying to generate a 64 bit extent large int?

I think this characteristic is not good as a random number generator.
We could instead sample 32bit numbers twice to generate a 64bit random number.

emcastillo:
One thing is that we need to call `generateLongLong` if the `RandomState` is `SOBOL64`
Shold we repurpose this PR to do that?

niboshi:
Can you elaborate why?

emcastillo:
`curand.generate` fails when curand generator is `CURAND RNG QUASI SOBOL64` or `CURAND RNG QUASI SCRAMBLED SOBOL64`. It only works by calling `curand.generateLongLong`

niboshi:
Oh, I see.
Then, I think `RandomState` should remember the method when creating a generator and use different generator functions depending on it.

emcastillo:
Exactly, should we do this in this PR?

niboshi:
I think it's ok.

toslunar:
#2829 closed the issue.  Thanks for your early proposal.

emcastillo:
I tried to implement the sobol generators support in RandomState but I think we should not support them. (they don't use a seed but vectors and would complicate the curand invoking code).

Doing a explicit check of the method should be enough.

Issue #2251, opened 2019-06-12
Unify tests for `cupy.random` and `cupy.random.RandomState`


toslunar:
`cupy.random` and `cupy.random.RandomState` should have almost the same interface.  Tests for them could be unified.

See #2250.

PR #2828, opened 2019-12-18
Improve compatibility of `random.randint`

toslunar:
- `high - low` should not overflow
- `low` and `high` can be of float types
- reflect empty `size` other than `0`.


tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -1192,6 +1192,19 @@ def test_randint_3(self):
         self.generate(3, 10, size=0)
 
 
+class TestRandintMisc(unittest.TestCase):
+
+    def test_randint_overflow(self):
+        cupy.random.randint(numpy.int8(-100), numpy.int8(100))
```
niboshi: How about checking the range of resulted values?

tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -1192,6 +1192,19 @@ def test_randint_3(self):
...
+
+    def test_randint_float(self):
+        cupy.random.randint(1.2, 3.4, 5)
```
niboshi: I think we should test this more thoroughly.

* check the range of generated values.
* negative numbers
* `high` is omitted
* conditions whose result would be different if `ceiling` was used

toslunar:
I'll follow https://github.com/numpy/numpy/blob/8a0858f3903e488495a56b4a6d19bbefabc97dca/numpy/random/_bounded_integers.pyx.in#L333-L334

toslunar:
> I'll follow https://github.com/numpy/numpy/blob/8a0858f3903e488495a56b4a6d19bbefabc97dca/numpy/random/_bounded_integers.pyx.in#L333-L334

No, `int(low)` etc. looks correct.  https://github.com/numpy/numpy/blob/8a0858f3903e488495a56b4a6d19bbefabc97dca/numpy/random/_bounded_integers.pyx.in#L319-L320

tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -1188,9 +1188,21 @@ def test_randint_1(self):
...
+        self.generate(3, size=(4, 0, 5))
+
+    def test_randint_overflow(self):
+        self.generate(numpy.int8(-100), numpy.int8(100))
+
+    def test_randint_float1(self):
+        self.generate(-1.2, 3.4, 5)
+
+    def test_randint_float2(self):
+        self.generate(6.7, size=(2, 3))
```
niboshi: How about checking the value range? (not only this test, but also other `randint` tests)

tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -1188,9 +1188,21 @@ def test_randint_1(self):
...
         self.generate(3, 10, size=0)
 
+    def test_randint_empty2(self):
+        self.generate(3, size=(4, 0, 5))
+
+    def test_randint_overflow(self):
+        self.generate(numpy.int8(-100), numpy.int8(100))
+
+    def test_randint_float1(self):
+        self.generate(-1.2, 3.4, 5)
```
niboshi: How about adding cases like these:
* (-1.8, 3.9)  # `round` would result in different range
* (-3.8, -1.2)  # negative number at the upper bound

tests/cupy_tests/random_tests/test_generator.py
toslunar: Currently such tests are only in `test_sample`.  It's similar to that `tests/cupyx_tests/test_scatter.py` has few tests of `cupyx.scatter_add` because `tests/cupy_tests/core_tests/test_ndarray_scatter.py` has many tests of `ndarray.scatter_add`.  BTW, I'm aware of #2251 but left it in the PR.

tests/cupy_tests/random_tests/test_generator.py
toslunar: Thanks for the concrete suggestion.  I printed the bounds generated by `test_sample.TestRandint2.test_bound_float1`.
```
-3.2163061490620386 1.4340494179096543
-1.7548240925086693 4.447761323199037
-5.952811438090274 -2.784308792168858
0.7074908472677865 4.258527551050477
-4.63293410315047 1.7509332942724996
-3.907978778828104 4.91321954312264
-4.916231095357449 -2.146717804499249
-3.8030250737500784 5.786237847073696
-4.280589872674058 4.116831490893233
-3.259262529583008 4.1622474872584
```

tests/cupy_tests/random_tests/test_generator.py
niboshi: I see, but this PR lacks output range checks for floating-point boundaries.
We need it anyway.
Maybe implement such checks in `test_sample.py`?

niboshi: I see.
I overlooked that `test_bound_float1` in `test_sample.py` is testing floating-point ranges.

tests/cupy_tests/random_tests/test_generator.py
niboshi: OK. Sorry I overlooked `test_bound_float1` is testing floating-point ranges.

tests/cupy_tests/random_tests/test_sample.py
```diff
@@ -47,6 +55,26 @@ def test_bound_2(self):
...
+        self.assertEqual(val.shape, (20,))
+        self.assertGreaterEqual(val.min(), -100)
+        self.assertLess(val.max(), 100)
+
+    @condition.repeat(3, 10)
+    def test_bound_float1(self):
+        # generate floats s.t. int(low) < int(high)
+        low, high = sorted(numpy.random.uniform(-5, 5, size=2))
+        low -= 1
+        high += 1
```
niboshi: This logic only generates pairs at least 2.0 apart and thus not exhaustive enough.
Actually, the specification of allowed spans in NumPy looks a bit complicated around zero...
* -2 < low < -1 ⇒ -1 < high
* low == -1 ⇒ -1 < high
* -1 < low < 0 ⇒ 1 <= high 
* low == 0 ⇒ 1 <= high
* 0 < low < 1 ⇒ 1 <= high
* low == 1 ⇒ 1 < high
* 1 < low < 2 ⇒ 2 <= high

I think it's better to test on these 7 configurations explicitly, instead of (or additionally to) using random values.

niboshi:
Jenkins, test this please

niboshi:
Test failure does not seem related.

PR #2829, opened 2019-12-18
Support 64 bit extent `randint`

### LINKED ISSUE #2816: random.randint support for range larger than int32

toslunar:
Fix #2816.

Compared to #2823, this PR keeps using 32-bit RNG.

### DIFF
--- a/cupy/random/generator.py
+++ b/cupy/random/generator.py
@@ -636,16 +636,13 @@ class RandomState(object):
             return cupy.array(())
 
         sample = cupy.empty((n,), dtype=dtype)
-        # 32-bit RNG will be called to fill 32-bit or 64-bit `sample`
-        generate_args = (
-            self._generator,
-            sample.data.ptr,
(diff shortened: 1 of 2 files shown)


emcastillo:
I think this is the way to go.

emcastillo:
Jenkins, test this please

emcastillo:
Seems like some of the tests requires fixes 😭

emcastillo:
Jenkins, test this please

toslunar:
I'd rather not regard the previous behavior as a bug because the error message is good.

PR #2785, opened 2019-12-09
Separate reduction code from `_kernel.pyx`

niboshi:
Removes `include 'reduction.pxi'` as it's confusing, because often you lose where functions are actually defined.

`_kernel.pyx` uses `_broadcast_core()`. Importing `_routines_manipulation.pyx` from `_kernel.pyx` causes cyclic dependencies. ~I refactored `_broadcast_core()` and `class broadcast` as `_broadcast.pyx`.~
I only moved `_broadcast_core` to `internal.pyx`.

PR #2725, opened 2019-11-27
keepdims should always preserve all dimensions in CUB-based reductions

grlee77:
closes #2720

The bug is fixed by the lines computing `out_shape` within `device_reduce`.

In order to avoid multiple calls to `_get_axis`, I moved the `_get_axis` call outside of `_preprocess_array` and call it directly in `_routines_math.pyx` and `_routines_statistics.pyx` instead. The duplicate logic could potentially be refactored into a common function

`_get_axes` will have already sorted the axes in ascending order and checked their range (It also separates out the axes to be reduced from those needed in determining the shape for keepdims).
This allows greatly simplifying the code in `_cub_device_segmented_reduce_axis_compatible` and removes the need for `_cub_device_reduce_axis_compatible` altogether.

Although not strictly part of the bug fix, I implemented a performance improvement in the segmented reduction case by replacing the following slow call

```Python
numpy.all(numpy.diff(cub_axis) == 1)
```
with a call to a new `_contig_axis` Cython helper function

PR #2598, opened 2019-11-03
Add tests for `cupy.cuda.cub`

leofang:
I think it's better to have a separate PR for CUB tests.

This PR will hopefully help #2579 once the CI is set up to test CUB. Preferably this PR should get reviewed after #2584 and #2596 are merged, but it is ***not*** blocked by those two PRs for two reasons:
1. the tests simply won't be run if `cupy.cuda.cub` is not built 
2. the original `cupy.argmin()` and `cupy.argmax()` would be tested instead, if the CUB support from #2596 is not ready. 

Note: using the `TestCUBperformance` class here I found that CUB may be slower for a few cases (such as small workloads), so I think the performance warning/documentation discussed with @emcastillo in #2562 may still be needed.

PR #2584, opened 2019-10-29
Build the `cupy.cuda.cub` module by default

leofang:
Close #3078. Close #3075. Close #3108. Close #3507.

This PR includes all the CUB v1.8.0 headers (see notes below) so that the `cupy.cuda.cub` module can be built by default. This also avoids the need of documenting how to build it.

Note that the `CUB_PATH` env variable is no longer needed.

TODO:
- [x] ~~Write a unit test for `cupy.cuda.cub`~~ (related: #2579) ***UPDATE:*** see #2598
- [x] Discuss with CuPy core devs whether we should set `cupy.cuda.cub_enabled = False` during import for backward compatibility, and whether the `CUB_DISABLED` env variable is still needed.

Note: 
1. The CUB project is open source under the BSD license, so redistributing its source code is allowed if the license file is included.
2. Since CUB is a header-only library, many of its files are actually needed. Below is the list of CUB headers included when compiling `cupy/cuda/cupy_cub.cu` (obtained via `cpp -MM -I(...) cupy/cuda/cupy_cub.cu`): 
```shell
cub/agent/agent_reduce.cuh
cub/agent/agent_reduce_by_key.cuh
cub/agent/agent_scan.cuh
cub/agent/single_pass_scan_operators.cuh
cub/block/block_discontinuity.cuh
cub/block/block_exchange.cuh
cub/block/block_load.cuh
cub/block/block_raking_layout.cuh
cub/block/block_reduce.cuh
cub/block/block_scan.cuh
cub/block/block_store.cuh
cub/block/specializations/block_reduce_raking.cuh
cub/block/specializations/block_reduce_raking_commutative_only.cuh
cub/block/specializations/block_reduce_warp_reductions.cuh
cub/block/specializations/block_scan_raking.cuh
cub/block/specializations/block_scan_warp_scans.cuh
cub/device/device_reduce.cuh
cub/device/device_segmented_reduce.cuh
cub/device/dispatch/dispatch_reduce.cuh
cub/device/dispatch/dispatch_reduce_by_key.cuh
cub/device/dispatch/dispatch_scan.cuh
cub/grid/grid_even_share.cuh
cub/grid/grid_mapping.cuh
cub/grid/grid_queue.cuh
cub/iterator/arg_index_input_iterator.cuh
cub/iterator/cache_modified_input_iterator.cuh
cub/iterator/constant_input_iterator.cuh
cub/thread/thread_load.cuh
cub/thread/thread_operators.cuh
[...]

PR #2832, opened 2019-12-18
fix delayed import in cub.pyx

grlee77:
This modification is necessary following #2785. It was not caught there because CUB isn't currently enabled on CI.

I did also try moving the import to the top of the file, but this caused an `ImportError`, so I have left it as-is.


grlee77:
If #2725 is merged first, this PR can be closed.
(#2725 is refactor of the CUB reduction code and also now uses the updated import)

leofang:
@kmaehashi Is it possible to get CUB tested by merging #2584 and #2598? With the recent refactoring work going on, the CUB module becomes a bit fragile as @grlee77 just found...It'd be nice to spot the bug asap.

emcastillo:
Will try to merge #2725 first.

emcastillo:
#2725 was merged, I am closing this

PR #2842, opened 2019-12-19
WIP: Improve the efficiency of cupyx.scipy.ndimage.convolve (and correlate)

grlee77:
This PR rewrites the n-dimensional convolution kernel to make it more efficient. This is especially true when the filter itself only has a small number of non-zero entries.

Prior to this PR, in ``n`` dimensions, the kernel consisted of `n` nested for loops that looped over the filter kernel.

In this PR, the change is that the filter kernel is first converted to a 1d array containing the non-zero coefficients and a second array containing the coordinates of these coefficients. This allows replacing the ``n`` nested for loops by a **single** loop that only traverses the non-zero coefficients.

In my initial testing this seems to give a substantial performance benefit, even in the case when the filter does not have any non-zero entries. Here is a basic benchmark:

```Python
import cupy
from cupyx.scipy.ndimage import convolve

d = cupy.cuda.Device()
x = cupy.random.randn(256, 256, 256);

# densely populated filter
w = cupy.ones((5, 5, 5))
w /= w.size
%timeit convolve(x, w); d.synchronize()  

# sparsely populated filter
w2 = cupy.zeros((5, 5, 5))
w2[0, 0, 0] = 0.5
w2[-1, -1, -1] = -0.5
%timeit convolve(x, w2); d.synchronize() 
```

Prior to this PR the times for these dense and sparsely populated filters are equivalent. I measure:
```201 ms ± 7.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)```

After this PR, I get the following for the densely populated case:
```127 ms ± 283 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)```

and for the sparsely populated case:
```3.71 ms ± 69.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)```

### DIFF
--- a/cupyx/scipy/ndimage/filters.py
+++ b/cupyx/scipy/ndimage/filters.py
@@ -107,9 +107,17 @@ def _correlate_or_convolve(input, weights, output, mode, cval, origin,
         return output
     input = cupy.ascontiguousarray(input)
     weights = cupy.ascontiguousarray(weights, cupy.float64)
+
+    # The kernel needs only the non-zero kernel values and their coordinates.
+    # This allows us to use a single for loop to compute the ndim convolution.
+    # The loop will be over only the the non-zero entries of the filter.
+    wlocs = cupy.where(weights)
+    wvals = weights[wlocs]     # (nnz,) array of non-zero values
+    wlocs = cupy.stack(wlocs)  # (ndim, nnz) array of indices for these values
+
     return _get_correlete_kernel(
-        input.ndim, mode, cval, input.shape, tuple(wshape), tuple(origin))(
-        input, weights, output)
+        input.ndim, mode, cval, input.shape, weights.shape, wvals.size,
+        tuple(origin))(input, wlocs, wvals, output)
 
 
 def _generate_boundary_condition_ops(mode, ix, xsize):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by grlee77** [comment]
I can refactor slightly to pass `wshape` and the number of nonzero entries instead of the array `w` to `_generate_correlete_kernel`. That should allow restoring the caching of kernels.

File: cupyx/scipy/ndimage/filters.py [hunk 1]
```diff
@@ -107,9 +107,17 @@ def _correlate_or_convolve(input, weights, output, mode, cval, origin,
         return output
     input = cupy.ascontiguousarray(input)
     weights = cupy.ascontiguousarray(weights, cupy.float64)
+
+    # The kernel needs only the non-zero kernel values and their coordinates.
+    # This allows us to use a single for loop to compute the ndim convolution.
+    # The loop will be over only the the non-zero entries of the filter.
+    wlocs = cupy.where(weights)
```
niboshi: This is equivalent to `cupy.nonzero(weights)` (I think it's better to write this way), so it synchronizes.

I wonder whether it's more significant in practice, compared to the improvement in this PR.

cupyx/scipy/ndimage/filters.py
```diff
@@ -107,9 +107,17 @@ def _correlate_or_convolve(input, weights, output, mode, cval, origin,
...
+    wvals = weights[wlocs]     # (nnz,) array of non-zero values
+    wlocs = cupy.stack(wlocs)  # (ndim, nnz) array of indices for these values
+
     return _get_correlete_kernel(
-        input.ndim, mode, cval, input.shape, tuple(wshape), tuple(origin))(
-        input, weights, output)
+        input.ndim, mode, cval, input.shape, weights.shape, wvals.size,
```
niboshi: Another point is that, by using `wvals.size` (which can vary depending on the values of `weights`) in the kernel code, it may cause infinitely many instances of kernel compilation.

grlee77: Yes, but wasn't that already the case before as well since the kernel can have any number of different shapes too? (The kernel dimensions get hardcoded into the kernel)

grlee77: I would guess dense kernels are quitr common but there are some interesting cases with sparse ones. For example, all nd filters in [this publication have only two non-zero entries](https://www.google.com/url?sa=t&source=web&rct=j&url=https://arxiv.org/abs/1708.08421&ved=2ahUKEwiI3qCpuMPmAhXMLs0KHTebD3IQFjABegQIBBAF&usg=AOvVaw1H6xZa4nowi3rLNgIH50YU)

I'm not sure there is a real world scenario where thousands of different kernels would be generated.

cupyx/scipy/ndimage/filters.py
asi1024: `cupy.where` returns the mask of non-zero values, so device synchronization is not needed. Please note that `cupy.nonzero` synchronizes because it returns the indices of non-zero values.

asi1024:
Why does this PR improve the performance of the densely populated inputs?

asi1024:
> Prior to this PR the times for these dense and sparsely populated filters are equivalent. I measure:
> `201 ms ± 7.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

Some values are cached internally after the first function call. Please measure the time by repeating 10 times also in this case.

asi1024:
I locally compared the performance in the densely populated case.
```py
import cupy
from cupyx.scipy.ndimage import convolve
from cupyx.time import repeat

d = cupy.cuda.Device()
x = cupy.random.randn(256, 256, 256);
w = cupy.random.randn(5, 5, 5)

perf = repeat(convolve, (x, w), n=50, n_warmup=10)
print(perf)
```

master(40b59e44d):
```
convolve            :    59.998 us   +/-54.305 (min:   46.976 / max:  339.244) us  72760.652 us   +/-79.005 (min:72704.094 / max:73062.813) us
```

This PR (bca5bba25):
```
convolve            :   359.820 us   +/-55.769 (min:  344.258 / max:  743.166) us  75956.443 us   +/-93.009 (min:75915.840 / max:76360.062) us
```

grlee77:
I cannot do it immediately, but will take a look again at performance later today or tomorrow. I did not necessarily expect it to be faster for the densely populated case (but would also not expect it to be much slower).

It would also be possible to keep the current implementation and call it instead of the newly proposed one when all (or most) entries of the filter are non-zero.

cupyx/scipy/ndimage/filters.py
niboshi: No, please check the documentation and the implementation:
https://github.com/cupy/cupy/blob/40b59e44d5bfe6296ce1b70ca54d121a0f304555/cupy/sorting/search.py#L191

cupyx/scipy/ndimage/filters.py
asi1024: If `weights` is guaranteed to have at most 2 non-zero entries, `convolve` is too generic to calculate it. It seems to be able to be achieved by just only adding two slices.

cupyx/scipy/ndimage/filters.py
asi1024: I see. it synchronizes as you say.

grlee77:
Okay, I reran with that `repeat` utility and also see a similar result to @asi1024. I think the CPU time being higher is to be expected due to the extra handling of weights to get the coordinates and values, but it looks like the GPU time is also slightly higher (212 vs 196 ms).

master(40b59e4):
```
convolve            :   234.678 us   +/-74.324 (min:  105.436 / max:  376.489) us  196379.369 us   +/-875.660 (min:194806.778 / max:197748.734) us
```
This PR (bca5bba):
```
convolve            :   759.932 us   +/-67.219 (min:  699.047 / max: 1095.894) us  213619.128 us   +/-916.371 (min:212165.634 / max:215121.918) us
```

cupyx/scipy/ndimage/filters.py
grlee77: The two-entry case was an extreme example of sparsity, but there are some other intermediate cases. Another one that I have seen are diamond shaped filters (i.e. extent 3 x 3 but the 4 corner entries are zero or 5 x 5 with 13 central non-zero entries). Those cases would also benefit from this PR.

grlee77:
Given that there is some loss of performance for the fully dense case, I no longer want to merge this as-is. So I guess there are three main options:

1.) close the PR as is
2.) retain both kernel generation functions and select between the two either in an automated fashion (probably hard to do at least some minimal overhead) or via a new kwarg.
3.) improve the proposed solution so it is also faster in the dense case (not sure this is possible)

I am interested in how best to implement kernels that use a support mask of this type as that is a common pattern for a number of other functions in `scipy.ndimage` which use a `structure` or `footprint` argument. All of the morphological operations do this for example, so I don't think it is wasted effort even if we do not choose to use it here.

Have any of you worked on or have plans to work on these other types of image operations (e.g. `binary erosion`, `minimum_filter`, etc.)?

grlee77:
I opened a separate PR with a simpler approach that I initially tried prior to this one: #2846.

niboshi:
I feel it's a waste to just throw away this optimization, because it's siginificant in almost-zero case.
As an option for (3), how about doing this optimization only if `weights` is a sparse matrix?
That wouldn't work in `scipy.ndimage`, but I think it's at least better than introducing a new argument in (2).

jakirkham:
Is this PR still needed @grlee77? Wasn't sure based on your earlier comment and the fact PR ( https://github.com/cupy/cupy/pull/2846 ) was merged.

grlee77:
> Is this PR still needed @grlee77? Wasn't sure based on your earlier comment and the fact PR ( #2846 ) was merged.

I will go ahead and close it given that some of the benefit was obtained by the simpler change in #2846.  This implementation can have some benefit for filters that are mostly zeros, but I think that is a pretty niche case, and perhaps not worth maintaining a separate code path here.

PR #2857, opened 2019-12-23
[backport] Fix split and array_split with unordered indices supplied

emcastillo:
Backport of #2815


emcastillo:

PR #2859, opened 2019-12-23
Add `ptp` ndarray method and function

grlee77:
This PR adds `ptp` (peak-to-peak). 

I have included a `dtype` argument to make it consistent with CuPy's min & max (which `ptp` calls internally), but this argument is not present in the NumPy API. If preferred it can be removed (I wasn't sure what the reasoning for its inclusion was).

Also, unlike `amin` or `amax`, `ptp` as implemented here does not currently support Fusion. I haven't looked at the Fusion code implementation, so am not sure how easy it might be to support? The current implementation calls both `_amin` and `_amax` ufuncs rather than implementing its own separate reduction kernel, so there was not a single kernel to point to as done in the `amin` function I looked at as a reference.

Finally, should we also check explicitly for `bool` dtype and raise a TypeError in that case to match NumPy's behavior?

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -660,6 +660,7 @@ from cupy.statistics.order import amin as min  # NOQA
 from cupy.statistics.order import nanmax  # NOQA
 from cupy.statistics.order import nanmin  # NOQA
 from cupy.statistics.order import percentile  # NOQA
+from cupy.statistics.order import ptp  # NOQA
 
 from cupy.statistics.meanvar import average  # NOQA
 from cupy.statistics.meanvar import mean  # NOQA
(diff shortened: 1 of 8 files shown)


niboshi:
Thank you for the PR!
I think there's no point to have `dtype` argument. Could you remove it?
As for `bool`, it's `-=` that should raise an error. I think it's fine not to check it in `ptp`.

grlee77:
I have now updated without the dtype argument

niboshi:
Thanks. Jenkins, test this please

grlee77:
Older NumPy apparently did not support `keepdims` or a tuple of axes. Please try again now.

niboshi:

niboshi:
👍

PR #2866, opened 2019-12-24
[backport] Avoid looking up null pointers' attributes

niboshi:
Backport of #2802


niboshi:

leofang:
We should also back port to v6 as this touches the consumer side.

niboshi:
@leofang Thank you for pointing out! #2868

PR #2870, opened 2019-12-24
Fix `testing.shaped_random` for shape `()`

niboshi:
Fixes #2869

PR #2872, opened 2019-12-24
Fix `argmin`/`argmax` `dtype` argument

### LINKED ISSUE #2595: Signatures and behaviors of `argmax` and `argmin` are incompatible with NumPy
First, for the signature: In CuPy the signature for `argmax()` and `argmin()` is 
```python
cupy.argm*(a, axis=None, dtype=None, out=None, keepdims=False)
```
But in NumPy it's 
```python
numpy.argm*(a, axis=None, out=None)
```
That is, `dtype` and `keepdims` should be removed.

Second, for the behavior: In CuPy the `axis` could be a tuple, but in Numpy it can only be an integer:
```python
>>> import cupy as cp
>>> a = cp.arange(60).reshape(3,4,5)
>>> a.argmax(axis=(0,1))
array([11, 11, 11, 11, 11], dtype=int64)
>>> 
>>> import numpy as np
>>> a = np.arange(60).reshape(3,4,5)
>>> a.argmax(axis=(0,1))
TypeError: 'tuple' object cannot be interpreted as an integer
```

I think both are easy to fix. Just add a few guards prior to calling the actual workhorses.

niboshi:
Fixes #2595
Requires #2870

### DIFF
--- a/cupy/core/_routines_statistics.pyx
+++ b/cupy/core/_routines_statistics.pyx
@@ -206,7 +206,7 @@ nanmax = create_reduction_func(
 
 cdef _argmin = create_reduction_func(
     'cupy_argmin',
-    tuple(['{}->{}'.format(d, r) for r in 'bhilq' for d in '?BhHiIlLqQ'])
+    tuple(['{}->{}'.format(d, r) for r in 'qlihb' for d in '?BhHiIlLqQ'])
     + (
         ('e->q', (None, 'my_argmin_float(a, b)', None, None)),
         ('f->q', (None, 'my_argmin_float(a, b)', None, None)),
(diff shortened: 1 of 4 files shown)


cupy/core/_routines_statistics.pyx
```diff
@@ -206,27 +206,27 @@ nanmax = create_reduction_func(
...
 cdef _argmin = create_reduction_func(
     'cupy_argmin',
-    ('?->q', 'B->q', 'h->q', 'H->q', 'i->q', 'I->q', 'l->q', 'L->q',
-     'q->q', 'Q->q',
-     ('e->q', (None, 'my_argmin_float(a, b)', None, None)),
-     ('f->q', (None, 'my_argmin_float(a, b)', None, None)),
-     ('d->q', (None, 'my_argmin_float(a, b)', None, None)),
-     ('F->q', (None, 'my_argmin_float(a, b)', None, None)),
-     ('D->q', (None, 'my_argmin_float(a, b)', None, None))),
+    tuple(['{}->{}'.format(d, r) for r in 'bhilq' for d in '?BhHiIlLqQ'])
```
leofang: I am a bit concerned with the order `'bhilq'`. Which return type is chosen when `argmin(..., dtype=None, ...)` is called? I thought we'd need to put `'q'` as the leading character to match NumPy default, but I could be wrong. (I don't have access now to test this PR.)

leofang:
To close #2595, it'd be nice to document the API deviation (which is a good thing I think).

cupy/core/_routines_statistics.pyx
niboshi: Thank you for pointing out! It seems you are correct. Some of the tests were failing.

emcastillo:
Waiting till #2870 is merged

niboshi:
#2870 merged.

emcastillo:
Jenkins, test this please

PR #2874, opened 2019-12-24
Disallow boolean subtraction

### LINKED ISSUE #2864: Boolean subtraction should be an error
```py
xp.array([False,True]) - xp.array([False, True])
```

If `xp` is `numpy`:
```
TypeError: numpy boolean subtract, the `-` operator, is deprecated, use the bitwise_xor, the `^` operator, or the logical_xor function instead.
```

CuPy should align with it.

(Some background:  #2859)

niboshi:
Fixes #2864

### DIFF
--- a/tests/cupy_tests/math_tests/test_arithmetic.py
+++ b/tests/cupy_tests/math_tests/test_arithmetic.py
@@ -1,6 +1,8 @@
 import itertools
-import numpy
 import unittest
+
+import numpy
+import pytest
 
 import cupy
 from cupy import testing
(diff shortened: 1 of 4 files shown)


tests/cupy_tests/math_tests/test_arithmetic.py
```diff
@@ -289,3 +290,21 @@ def test_modf(self, xp, dtype):
...
+    'xp': [numpy, cupy],
+    'shape': [(3, 2), (), (3, 0, 2)]
+}))
+@testing.gpu
+class TestBoolSubtract(unittest.TestCase):
+
+    def test_bool_subtract(self):
+        xp = self.xp
+        if xp is numpy and not testing.numpy_satisfies('>=1.13.0'):
+            raise pytest.SkipTest('NumPy<1.13.0')
```
toslunar: How about the following?
```python
@testing.with_requires('numpy>=1.13')
@testing.numpy_cupy_raises(accept_error=TypeError)
def test_bool_subtract(self, xp):
```

It's fair to skip the cupy test with numpy<1.13.

niboshi: Hm, I felt I wanted to test its behavior even with numpy<1.13.
Is it not necessary?

toslunar: It looks an implementation detail to me whether to reproduce the old behavior if old numpy is installed.

niboshi: I don't think it's an implementation detail, because the difference affects those users using old numpy.

toslunar: Then the policy to follow newest numpy could be documented (at `cupy.subtract`?) as well as it's tested.

toslunar:
Jenkins, test this please.

tests/cupy_tests/math_tests/test_arithmetic.py
toslunar:
```suggestion
            pytest.skip('NumPy<1.13.0')
```

toslunar:
Jenkins, test this please.

PR #2875, opened 2019-12-24
Remove `dtype` argument from `min`/`max`

### LINKED ISSUE #2865: Remove dtype argument from min/max?
It looks to me there's no point for these functions to have `dtype` argument.
They should be keyword-only at the very least, for numpy compatibility.

niboshi:
Fixes #2865


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #2876, opened 2019-12-24
[RFC] Disallow `testing.numpy_cupy_raises` without explicit error type

### LINKED ISSUE #2871: Disallow `testing.numpy_cupy_raises` without explicit error type?
By default this decorator passes a test with any errors raised, but I think we should never write such a test.

Problems like below could be prevented.
* #2870
* https://github.com/cupy/cupy/pull/2820/files#diff-17feca175c3f71ea15a5fe29026fa958L145-L149

`pytest.raises()` also requires explicit error type.

niboshi:
Fixes #2871

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -563,11 +563,15 @@ numpy: %s''' % (str(cupy_result), str(numpy_result))
     return decorator
 
 
-def numpy_cupy_raises(name='xp', sp_name=None, scipy_name=None,
-                      accept_error=Exception):
+def numpy_cupy_raises(
+        accept_error, *, name='xp', sp_name=None, scipy_name=None):
     """Decorator that checks the NumPy and CuPy throw same errors.
 
     Args:
+         accept_error(Exception or tuple of Exception): Specify
+             acceptable errors. When both NumPy test and CuPy test raises the
+             same type of errors, and the type of the errors is specified with
+             this argument, the errors are ignored and not raised.
          name(str): Argument name whose value is either
              ``numpy`` or ``cupy`` module.
          sp_name(str or None): Argument name whose value is either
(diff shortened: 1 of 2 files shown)


niboshi:
This inevitably breaks backward compatibility.
I want to make `accept_error` a positional required argument and others keyword-only.

cupy/testing/helper.py
```diff
@@ -563,11 +563,15 @@ def test_func(self, *args, **kw):
...
-                      accept_error=Exception):
+def numpy_cupy_raises(
+        accept_error, *, name='xp', sp_name=None, scipy_name=None):
     """Decorator that checks the NumPy and CuPy throw same errors.
 
     Args:
+         accept_error(Exception or tuple of Exception): Specify
+             acceptable errors. When both NumPy test and CuPy test raises the
+             same type of errors, and the type of the errors is specified with
+             this argument, the errors are ignored and not raised.
```
asi1024: The test with this decorator does not pass when both NumPy test and CuPy test don't raise any error.

niboshi: I agree the wording could be improved to be more unambiguous, but it's not the point of this PR.

cupy/testing/helper.py
```diff
@@ -590,6 +588,18 @@ def numpy_cupy_raises(name='xp', sp_name=None, scipy_name=None,
...
+        if (isinstance(accept_error, type)
+                and issubclass(accept_error, Exception)):
+            break
+        if (isinstance(accept_error, tuple)
+                and all(isinstance(ex, type) for ex in accept_error)
+                and all(issubclass(ex, Exception) for ex in accept_error)):
+            break
+        raise TypeError(
+            'accept_error must be a subtype of Exception or a tuple of such '
+            'types.')
```
asi1024: Why did you use a while-loop? This loop looks iterated at most once.
How about writing like as follows?
```py
if not isinstance(accept_error, tuple):
    accept_error = accept_error,
if not all(isnstance(ex, type) and issubclass(ex, Exception) for ex in accept_error):
    raise TypeError(...)
```

niboshi: IMO `whlie True:` is cleaner. It's more readable and there's no redundancy (no need to preprocess before checking).
I often write this way personally, but I will fix it if you still don't like it.

asi1024: There are two similar fragment `isinstance(...) and issubclass(...)` in your implementation, but it is avoidable by introducing preprocess before checking. I still feel my suggestion is better.

niboshi: OK. By the way, let's settle the discussion of #2871 before making a fix.

kmaehashi:
Blocked by https://github.com/cupy/cupy/pull/2987

PR #2498, opened 2019-09-26
Fix `@testing.numpy_cupy_` decorators for skips

niboshi:
Fixes #2492

PR #2394, opened 2019-08-13
Raise `DeprecationWarning` on 0-dim arrays in `numpy.nonzero` to match NumPy 1.17 behavior

asi1024:
See the description of #2387.

PR #2878, opened 2019-12-25
[backport] Refactor array testing decorators

emcastillo:
Backport of #2818

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -25,7 +25,6 @@ import cupyx.scipy.sparse
 def _call_func(self, impl, args, kw):
     try:
         result = impl(self, *args, **kw)
-        assert result is not None
         error = None
         tb_str = None
     except Exception as e:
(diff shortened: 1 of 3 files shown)


emcastillo:

emcastillo:
Seems like #2394 should be backported to v6 first 🤔,
or do we just not merge this in v6.

niboshi:
I think #2498 should be backported instead.

Current v6 has a problem that the decorators does not treat `SkipTest` correctly, so if cupy raises it, the test will be skipped even if numpy behaves differently. That shoud be fixed (by backporting #2498).

However, #2498 alone does not fix all decorators. As `numpy_cupy_array_list_equal()` is not using the common implementation, #2498 won't fix the decorator. Backporting #2498 and this PR together fixes all the decorators.

emcastillo:
Backported #2892

cupy/testing/helper.py
```diff
@@ -40,6 +39,91 @@ def _call_func(self, impl, args, kw):
...
+    cupy_axis_error = cupy.core._errors._AxisError
+    if isinstance(cupy_error, cupy_axis_error) and numpy_axis_error is None:
+        if not isinstance(numpy_error, (ValueError, IndexError)):
+            return False
+        errors = list(set(errors) - set([IndexError, ValueError]))
+
+    return all([isinstance(cupy_error, err) == isinstance(numpy_error, err)
+                for err in errors])
+
+
```
niboshi: Please remove them.

emcastillo:
PTAL!

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #2881, opened 2019-12-25
Sort type map in `_kernel.pyx`

niboshi:
`items()` is not guaranteed to be in any order (at least before Python 3.7).
This value is used as a cache key, so it should be sorted.


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #2883, opened 2019-12-25
Fix `test_helper.py` for NumPy 1.18

niboshi:
Error message was changed in 1.18.
I think error message can be simply ignored.


emcastillo:
Jenkins, test this please

PR #2884, opened 2019-12-25
Fix negative value test in `test_helper.py`

niboshi:
`assertRaisesRegex` was wrapping a function decorated with `for_unsigned_dtypes` and `for_signed_dtypes`.
In this scheme, if an error was raised in any single one of the combinations of dtypes, the test would pass, which would be unintended.


takagi:
Would rebase this on the current master?

kmaehashi:
@takagi could you rebase and submit PR?

takagi:
I take this over as #3579.

PR #2886, opened 2019-12-25
Fix imag for 0-size array

### LINKED ISSUE #2885: Complex imag causes error for 0-size array
`ndarray.imag` causes `AssertionError` if the array is 0-sized.

Version: current master (2aec8a78761cdcc88541bf1f5e40e17840d3eb84)
### Code:
```py
>>> import cupy
>>> a = cupy.zeros((0,), dtype='complex64')
>>> a.imag
  File "cupy/core/core.pyx", line 1095, in cupy.core.core.ndarray.imag.__get__
    return _math._ndarray_imag_getter(self)
  File "cupy/core/_routines_math.pyx", line 52, in cupy.core._routines_math._ndarray_imag_getter
    memptr=self.data + self.dtype.itemsize // 2,
  File "cupy/cuda/memory.pyx", line 320, in cupy.cuda.memory.MemoryPointer.__add__
    assert self.ptr != 0 or offset == 0
AssertionError
>
```

niboshi:
Fixes #2885

### DIFF
--- a/cupy/core/_routines_math.pyx
+++ b/cupy/core/_routines_math.pyx
@@ -47,8 +47,11 @@ cdef ndarray _ndarray_real_setter(ndarray self, value):
 
 cdef ndarray _ndarray_imag_getter(ndarray self):
     if self.dtype.kind == 'c':
+        dtype = get_dtype(self.dtype.char.lower())
+        if self.size == 0:
+            return ndarray(shape=self._shape, dtype=dtype)
         view = ndarray(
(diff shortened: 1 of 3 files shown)


niboshi:
If #2870 is merged earlier, its TODO can be addressed.

niboshi:
>If #2870 is merged earlier, its TODO can be addressed.

Done. PTAL

niboshi:
Your intention is to make the resulting array alway a view of the original array, right?
Fixed as you suggestion, thanks!

toslunar:
Jenkins, test this please.

PR #2893, opened 2019-12-26
[RFC] Apply synchronization detection in unit tests

niboshi:
Addresses https://github.com/cupy/cupy/pull/2819#issuecomment-568971301

I'm not sure if we should fix all the tests and apply `allow_synchronize=False` by default, because there are so many tests that were caught.
I made this PR at least for record.


niboshi:
Previous log of test failures:
https://jenkins.preferred.jp/job/chainer/job/cupy_pr/798/

I can't list all the tests here because that would exceed the GitHub limit, but there were 4,801 failures.

Issue #2896, opened 2019-12-27
More appropriate name for cupyx.time.repeat?


niboshi:
I think `cupyx.time.repeat` (#2740) is not necessarily the best name.
 
Random thoughs:
* Repeating is not the main goal of the function.
* There may be other metrics than time.
* There is `cupy.prof`.

leofang:
How about `cupyx.time.benchmark`?

kmaehashi:
#5031
> We originally named `cupyx.time.repeat` as it is similar to `timeit.repeat`, but later we noticed that the semantics is slightly different between them. The latter measures the time taken to run given code for *number* times and repeat that measurement for *repeat* times. This is one of the reason why we marked it as experimental.
> 
> However, it is important to provide a stable API for benchmarking / profiling. To remove the experimental flag from the `cupyx.time.repeat`, we should consider a good name for it.
> 
> Besides that, it would be beneficial if we can provide a wall-clock timer function with CUDA synchronization that can be passed as `timer` argument of `timeit.repeat`.

kmaehashi:
Closed via  #5940.

PR #2897, opened 2019-12-27
Make `cupyx.time.repeat` experimental

niboshi:
There might be a room to discuss its name.

#2896


emcastillo:
Jenkins, test this please

PR #2898, opened 2019-12-27
Declare device synchronization at `runtime.free()`

niboshi:
`cudaFree` (but not `cudaMalloc`) causes device synchronization.

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -11,12 +11,13 @@ import threading
 import warnings
 import weakref
 
+from cupy.cuda import runtime
+from cupy.core import syncdetect
+
 from fastrlock cimport rlock
 from libc.stdint cimport int8_t
 from libc.stdint cimport intptr_t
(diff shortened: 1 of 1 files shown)


emcastillo:
I am looking about cudaMalloc being asynchronous but I can't seem to find a reliable source.
Do you have any pointer to this?
The PR looks good to me.

niboshi:
@emcastillo I confirmed empirically. It could be dependent on devices?  I'll try to find a reliable source...

emcastillo:
Yeah, I saw a post in [stackoverflow](https://stackoverflow.com/questions/36001364/has-cudamalloc-changed-to-be-asynchronous) saying the same and the behavior was previously synchronous, but no word on documentation 😢.

emcastillo:
After the discussion today this approach seems valid.
Malloc synchronization is not documented but current versions of CUDA do not do it.
However, this behavior may change.

emcastillo:
Jenkins, test this please

PR #2644, opened 2019-11-15
Support multi-GPU `cupy.cuda.cufft.Plan1d`

leofang:
***--- PR description updated, please read ---*** 

This PR adds initial support for multi-GPU cuFFT. I only target `Plan1d` here because it's easier to test and to understand the corner cases. This PR makes the necessary preparation for future extension. For example, I will support `PlanNd` in the next PR.

In short, this PR allows multi-GPU cuFFT plans to be used with both high- and low-level API calls:
```python
import cupy as cp


a = cp.random.random(1024).astype(cp.complex64)
a += 1j*cp.random.random(1024)

cp.fft.config.use_multi_gpus = True  # flag added in this PR
cp.fft.config.set_cufft_gpus(2)      # handle added in this PR
out = cp.fft.fft(a)                  # this just works
```
```python
import cupy as cp


batch = 7
length = 512

# use 2 GPUs
a = cp.random.random((batch, length)) +1j*cp.random.random((batch, length))
a = a.astype(cp.complex64)
plan = cp.cuda.cufft.Plan1d(length, cp.cuda.cufft.CUFFT_C2C, batch,
                            use_multi_gpus=True, devices=2)
out = plan.get_output_array(a)
plan.fft(a, out, cp.cuda.cufft.CUFFT_FORWARD)

# use 1 GPU
out2 = cp.fft.fft(a)

# results should be identical 
assert cp.allclose(out, out2)
```
The second example is nice because `Plan1d` can actually be retrieved by `cupyx.scipy.fftpack.get_fft_plan()`.

Multi-GPU cuFFT are beneficial for at least two reasons:
1. Bypassing memory limitation: every GPU has limited memory, and even worse in order to run `cupy.fft.fft()`, for example, it would need additional buffer space. So it turns out one cannot even FFT a 8G array on a P100 with 16G memory. Having extra workhorses would alleviate this limitation.
2. Reasonable performance: while [it's not guaranteed to be faster using multiple GPUs](https://docs.nvidia.com/cuda/cufft/index.html#multiple-GPU-cufft-transforms), for users who have NVLink interconnects, the bandwidth is large enough that performance is acceptable for a huge array. It seems that NVLinks can support more simultaneous async transfers than allowed by the engine counts (which shows up in `nvprof`/`nvvp` as `Memcpy PtoP` --- on a DGX-2 I found overlapped memcpy among 16 GPUs is possible with this PR!)

[...]

PR #2899, opened 2019-12-27
Readability fix in `memory.pyx`

niboshi:
Converts private free functions to instance (or class) methods of relevant classes.

PR #2901, opened 2020-01-01
Ignore error when peer access is already enabled 

leofang:
Cherry-pick from #2644. This PR fixes the error when peer access is enabled by some external libraries (cuFFT in the case of #2644). The `CUDARuntimeError` should be silently ignored in this case.


niboshi:
LGTM
Jenkins, test this please

niboshi:
As #2899 has just been merged, could you resolve conflict?

leofang:
@niboshi Done, PTAL 🙂

niboshi:
Thanks!
Jenkins, test this please

PR #2076, opened 2019-03-03
Fix `true_divide` with dtype argument

okuta:
Fix #2059

PR #2860, opened 2019-12-23
Add CUB reduction support to `mean`

grlee77:
as mentioned in #2725, this PR enables a CUB code path for mean computation

PR #2903, opened 2020-01-03
Fix handling of dtypes in cupy.mean

grlee77:
The first commit here fixes a bug in CUB-based mean that was introduced in #2860. When the input had an integer dtype, the in-place addition with a float would fail with a `TypeError`.

The second two commits fix both the CUB code path and the non-CUB one to return the requested dtype. In current master, mean would fail when dtype was not a floating (or complex) type as there is not case for that defined in the `_mean_core` ufunc. NumPy does allow such cases. The newly added tests verify consistent behavior. They pass when I tried it locally for both `CUB_DISABLED=0` and `CUB_DISABLED=1`.

### DIFF
--- a/cupy/core/_routines_statistics.pyx
+++ b/cupy/core/_routines_statistics.pyx
@@ -58,7 +58,8 @@ cdef ndarray _ndarray_mean(ndarray self, axis, dtype, out, keepdims):
     dtype_sum = dtype_out = dtype
     if dtype is None:
         if self.dtype.kind in 'iub':
-            dtype_out = dtype_sum = numpy.float64
+            dtype_out = numpy.float64
+            dtype_sum = numpy.float64
         elif self.dtype.char == 'e':
             dtype_sum = numpy.float32
             dtype_out = numpy.float16
--- a/cupy/core/_routines_statistics.pyx
+++ b/cupy/core/_routines_statistics.pyx
@@ -55,18 +55,31 @@ cdef ndarray _ndarray_argmin(ndarray self, axis, out, dtype, keepdims):
 
 
 cdef ndarray _ndarray_mean(ndarray self, axis, dtype, out, keepdims):
+    dtype_sum = dtype_out = dtype
+    if dtype is None:
+        if self.dtype.kind in 'iub':
+            dtype_out = dtype_sum = numpy.float64
+        elif self.dtype.char == 'e':
+            dtype_sum = numpy.float32
+            dtype_out = numpy.float16
+    elif numpy.dtype(dtype).kind in 'iub':
+        # output will be the requested type, but compute the mean using float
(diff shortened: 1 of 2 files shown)


grlee77:
Complex-valued input with a real-valued `dtype` is still not handled correctly here, although I suspect that case would be rare. NumPy does allow it, but you will get a `ComplexWarning` about discarding the imaginary part.

tests/cupy_tests/statics_tests/test_meanvar.py
```diff
@@ -88,11 +88,24 @@ def test_external_mean_axis(self, xp, dtype):
         a = testing.shaped_arange((2, 3, 4), xp, dtype)
         return xp.mean(a, axis=1)
 
+    @testing.for_all_dtypes(no_complex=True)
     @testing.numpy_cupy_allclose()
-    def test_mean_all_dtype(self, xp):
-        a = xp.full((2, 3, 4), 123456789, dtype=numpy.int64)
+    def test_mean_real_dtypes_as_float(self, xp, dtype):
+        a = xp.full((2, 3, 4), 123456789, dtype=dtype)
         return xp.mean(a, dtype=numpy.float64)
```
niboshi: This is a test specifically for a bug fix #2076. It's better not to generalize it.
(I think the test name is inappropriate.)

niboshi:
It looks the first and the second issues are independent. Is that correct?
How about separating a PR?

niboshi:
I suggested that because the first problem looks easier to review, but there might be a room for consideration for the second one.

grlee77:
Okay, I will split just the in-place division out into a separate PR. That one is a clear bug. The rest is a question of how closely we want to try to emulate the NumPy behavior.

grlee77:
This one should now be reviewed after #2943 (which is just the first commit from this PR).

niboshi:
Sorry for delay. Could you resolve conflict?

cupy/core/_routines_statistics.pyx
```diff
@@ -55,18 +55,31 @@ cdef ndarray _ndarray_argmin(ndarray self, axis, out, dtype, keepdims):
 
 
 cdef ndarray _ndarray_mean(ndarray self, axis, dtype, out, keepdims):
+    dtype_sum = dtype_out = dtype
+    if dtype is None:
+        if self.dtype.kind in 'iub':
+            dtype_out = dtype_sum = numpy.float64
```
niboshi: Could you separate the assignments to two lines?

grlee77: okay

niboshi:
Thank you! Jenkins, test this please

niboshi:
:+1:

PR #2906, opened 2020-01-06
Fix `AssertionError` for `numpy>=1.18`

emcastillo:
New version of NumPy changed the error message for this test
[Example](https://jenkins.preferred.jp/job/chainer/job/cupy_pr/811/TEST=cupy-py3,label=mn1-p100/testReport/tests.cupy_tests.testing_tests.test_helper/TestIgnoreOfNegativeValueDifferenceOnCpuAndGpu/test_correct_failure/)

Creates a new test case only for 1.18 and restricts the older one.


niboshi:
Duplicate: #2883

PR #2909, opened 2020-01-06
Fix logic to check explicit `size` argument in `ElementwiseKernel`

### LINKED ISSUE #2908: Incorrect logic to check explicit size argument in ElementwiseKernel
Version: current master (0af45a409d002ff8a413f83df1b4df5e6d12792f)

### Code
```py
import cupy
kernel = cupy.ElementwiseKernel(
    'T x, T a', 'T out',
    '''
    out = x + a
    ''')
x = cupy.array([1, 2], dtype='float32')
a1 = cupy.array([3, 4], dtype='float32')
a2 = 9
kernel(x, a1, size=2)  # A (a1 is an array): causes an error
#kernel(x, a2, size=2)  # B (a2 is a scalar): no error
```

A causes an error: `ValueError: Specified 'size' can be used only if all of the ndarray are 'raw'.`,
while B doesn't.
I think B should cause an error as well.

Relevant code: https://github.com/cupy/cupy/blob/0af45a409d002ff8a413f83df1b4df5e6d12792f/cupy/core/_kernel.pyx#L343

niboshi:
Fixes #2908

### DIFF
--- a/tests/cupy_tests/core_tests/test_userkernel.py
+++ b/tests/cupy_tests/core_tests/test_userkernel.py
@@ -177,6 +177,22 @@ class TestElementwiseKernelSize(unittest.TestCase):
         with self.raises_size_required():
             kernel4(7, self.arr1)
 
+    def test_size_determined_by_output(self):
+        # All the input args are unsized, but the size can be determined by the
+        # output arg. size argument is not allowed.
+
+        # Raw input
+        kernel1 = self.create_kernel((True,), (False,))
+        kernel1(self.arr1, self.arr2)
+        with self.raises_size_not_allowed():
+            kernel1(self.arr1, self.arr2, size=2)
+
+        # Scalar input
+        kernel2 = self.create_kernel((False,), (False,))
+        kernel2(self.arr1, self.arr2)
+        with self.raises_size_not_allowed():
+            kernel2(7, self.arr2, size=2)
+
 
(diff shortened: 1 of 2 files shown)


tests/cupy_tests/core_tests/test_userkernel.py
```diff
@@ -67,6 +68,132 @@ def test_numpy_scalar(self, dtype):
...
+        kernel2(self.arr1, self.arr2, size=2)
+        with self.raises_size_required():
+            kernel2(self.arr1, self.arr2)
+
+    def test_all_nonraws(self):
+        # All arrays are not raw -> size not allowed
+        kernel1 = self.create_kernel((False, False), (False,))
+        with self.raises_size_not_allowed():
+            kernel1(self.arr1, self.arr2, size=2)
+        kernel2 = self.create_kernel((False, False), (True,))
```
emcastillo: Could these multiple kernel calls in all the tests be parametrized?

niboshi: Parameterization would also be pretty complicated.
It can be surely done, but I'm not sure it's worth the effort.

emcastillo: I agree with your answer.

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

PR #2731, opened 2019-11-28
Optimize kernel launch parameters

niboshi:
A prototype implementation for automatic tuning of kernel launch parameters.

Usage:
```py
with cupyx.optimize(key='some key'):
    ...
```

Current implementation is very basic: Only `block_size` of `ReductionKernel` is tuned per a single `optimize` context.

PR #2914, opened 2020-01-07
Allow `ElementwiseKernel` to set the block_size

emcastillo:
Orthogonal to #2731

Right now the `ElementwiseKernel` class launches all kernels with the same block size which is not easily tunable.

This PR replaces the hardcoded value with a kwarg for the `__call__` method.

### DIFF
--- a/cupy/core/_kernel.pyx
+++ b/cupy/core/_kernel.pyx
@@ -568,6 +568,8 @@ cdef class ElementwiseKernel:
                 This parameter must be specified if and only if all ndarrays
                 are `raw` and the range size cannot be determined
                 automatically.
+            block_size (int): Number of threads per block. By default, the
+                value is set to 128.
 
         Returns:
             If ``no_return`` has not set, arrays are returned according to the
@@ -583,6 +585,7 @@ cdef class ElementwiseKernel:
         size = -1
         size = kwargs.pop('size', -1)
         stream = kwargs.pop('stream', None)
+        block_size = kwargs.pop('block_size', 128)
         if len(kwargs):
             raise TypeError('Wrong arguments %s' % kwargs)
 
@@ -635,7 +638,7 @@ cdef class ElementwiseKernel:
         args_info = _get_args_info(inout_args)
         kern = self._get_elementwise_kernel(dev_id, args_info, types)
         kern.linear_launch(indexer.size, inout_args, shared_mem=0,
-                           block_max_size=128, stream=stream)
+                           block_max_size=block_size, stream=stream)
         return ret
 
     cpdef tuple _decide_params_type(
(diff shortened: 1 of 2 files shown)


niboshi:
@asi1024 I feel it's good to have this option. Do you have any thoughts?

cupy/core/_kernel.pyx
```diff
@@ -568,6 +568,8 @@ cdef class ElementwiseKernel:
                 This parameter must be specified if and only if all ndarrays
                 are `raw` and the range size cannot be determined
                 automatically.
+            block_size (int): Number of threads per block. By default, the 
+                value is set to 128
```
niboshi:
```suggestion
                value is set to 128.
```

Please fix flake8 errors as well.

emcastillo:
PTAL.
I really need to configure my vim autoflake plugin for cython 😅

asi1024:
I agree to support this option too.

niboshi:
@emcastillo Can you add some test?
I think it's difficult to test the block size is actually changed, so it's enough to check the argument is accepted.

emcastillo:
Yes, I was actually having doubts on how to test this, so I decided not to do it.

emcastillo:
I added a test that tries to set an invalid block size and relies on the exception. I think this should be enough. PTAL

niboshi:
Could you add a successful case?

niboshi:
Perhaps with `block_size=1`.

emcastillo:
Yes, but is there any way to know that it was successfully executed with such a block size?

niboshi:
@emcastillo 
You can simply check it doesn't raise any error and the outcome is correct.
As I said above, you don't have to check the actual block size.

emcastillo:
PTAL

niboshi:
@emcastillo 
Please check the resulted array.

emcastillo:
Of course, I look like an idiot now 😂

tests/cupy_tests/core_tests/test_userkernel.py
```diff
@@ -82,3 +83,21 @@ def test_scalar(self, xp, dtype):
...
+            kernel(x, 1, block_size=0)
+
+    @testing.numpy_cupy_array_equal()
+    def test_block_size(self, xp):
+        x = testing.shaped_arange((2, 3, 4), xp, xp.float32)
+        if xp is numpy:
+            y = numpy.array(1).astype(xp.float32)
+            return x + y
+        kernel = cupy.ElementwiseKernel('T x, T y', 'T z', 'z = x + y')
+        return kernel(x, 1, block_size=1)
```
niboshi: As the objective of this test is not for comparing with numpy, instead of using `numpy_cupy_array_equal`, how about using `testing.assert_array_equal`?
```py
y = kernel(x, 1, block_size=1)
testing.assert_array_equal(y, x + 1)
```

emcastillo: Done, thanks!

niboshi:
Thanks.
Jenkins, test this please

niboshi:
LGTM

PR #2917, opened 2020-01-07
Clean up `_scalar.pyx`

niboshi:
Cleans up `_scalar.pyx/pxd`.

* Previously, `convert_scalar()` returned different types of objects depending on a flag. This is confusing because the return type is obscure.
* Unifies `CScalar`'s various creation functions into static methods `CScalar.from_xxx()`.

### DIFF
--- a/cupy/core/_scalar.pxd
+++ b/cupy/core/_scalar.pxd
@@ -13,12 +13,23 @@ cdef class CScalar(CPointer):
         char kind
         int8_t size
 
+    @staticmethod
+    cdef CScalar from_int32(int32_t value)
+
+    @staticmethod
+    cdef CScalar from_numpy_scalar_with_dtype(object x, object dtype)
+
+    @staticmethod
+    cdef CScalar _from_python_scalar(object x)
+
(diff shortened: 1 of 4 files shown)


niboshi:
You might as well review #2918 altogether as it's based on this PR.

cupy/core/_scalar.pxd
```diff
@@ -13,12 +13,23 @@ cdef class CScalar(CPointer):
         char kind
         int8_t size
 
+    @staticmethod
+    cdef CScalar from_int32(int32_t value)
+
+    @staticmethod
+    cdef CScalar from_numpy_scalar_with_dtype(object x, object dtype)
```
toslunar: Note:
These methods are still considered as private because the class `CScalar` is in a private module and Cython "cannot call a static method on an instance variable".
https://github.com/cython/cython/blob/3dc2b9dfc23638fbef2558d619709b5235d5df08/Cython/Compiler/ExprNodes.py#L5619-L5624

toslunar:
Jenkins, test this please.

Issue #2905, opened 2020-01-05
`cupy.cumsum()` much slower than `cupy.core._routines_math.scan()`?


leofang:
I noticed that there are two functions in CuPy that can do prefix sum, so out of curiosity I did a quick benchmark, and I found `cumsum()` is much slower than the internal `scan()` (it's a `cdef` function wrapped by `_scan_for_test()`):
```python
>>> import cupy as cp
>>> import cupyx
>>> c = cp.random.random(1000000)
>>> print(cupyx.time.repeat(cp.cumsum, (c,)))
cumsum              :   468.171 us   +/-66.449 (min:  403.769 / max: 1040.217) us    493.367 us   +/-55.708 (min:  450.144 / max: 1056.448) us
>>> print(cupyx.time.repeat(cp.core._routines_math._scan_for_test, (c,)))
_scan_for_test      :    98.772 us   +/- 5.683 (min:   94.778 / max:  198.381) us    125.730 us   +/- 5.498 (min:  121.152 / max:  219.488) us
>>> cp.allclose(cp.core._routines_math._scan_for_test(c), cp.cumsum(c))
array(True)
```
I wonder if we should just use the implementation of `scan()` in `cumsum()` when the input array is 1D (or raveled)?

emcastillo:
I've been looking at the code.
It actually make sense to do it that way, but that's an ugly hack and I think that we should look into why is cumsum being that slow and try to fix that.

emcastillo:
I just have a PR that solves this without using scan, let me post it :)

emcastillo:
There is a problem with the synchronization and data races.

leofang:
Update: just did a quick test against the master (with #2907 and #2919 included)
```python
>>> import cupy as cp
>>> import cupyx
>>>
>>> c = cp.random.random(1000000)
>>> cp.cuda.cub_enabled=False
>>> print(cupyx.time.repeat(cp.cumsum, (c,)))  # from #2907
cumsum              :    37.220 us   +/-10.946 (min:   34.066 / max:  380.719) us    136.430 us   +/- 9.255 (min:  127.168 / max:  406.944) us
>>> print(cupyx.time.repeat(cp.core._routines_math._scan_for_test, (c,)))
_scan_for_test      :    22.048 us   +/- 3.469 (min:   20.648 / max:   75.092) us    100.439 us   +/- 9.159 (min:   94.528 / max:  148.224) us
[...]

PR #2907, opened 2020-01-06
Improve performance of `cumsum` and `cumprod`

emcastillo:
Solves #2905 

Uses the fast scan available in _math_routines and adds new kernels for batched cumsum/cumprod

Matrix size : (550, 100, 72, 50)

Previous to this PR:
```
cumsum   axis=None  :   228.304 us   +/-23.550 (min:  217.939 / max:  358.161) us  80835.688 us   +/-37.779 (min:80798.241 / max:81046.272) us
cumsum   axis=0     :   241.658 us   +/-11.215 (min:  234.609 / max:  315.020) us  84766.405 us   +/- 9.216 (min:84760.162 / max:84825.439) us
cumsum   axis=1     :   261.572 us   +/- 8.596 (min:  255.286 / max:  308.073) us  85690.466 us   +/- 6.717 (min:85684.509 / max:85727.554) us
cumsum   axis=2     :   261.069 us   +/- 6.102 (min:  255.105 / max:  292.426) us  85336.803 us   +/-49.660 (min:85323.997 / max:85682.014) us
cumsum   axis=3     :   258.408 us   +/-48.437 (min:  239.055 / max:  570.139) us  345277.179 us   +/-337.118 (min:344687.134 / max:346113.464) us
```

This PR:
```
cumsum   axis=None  :    65.940 us   +/-14.030 (min:   60.568 / max:  148.783) us  13706.375 us   +/-27.500 (min:13665.888 / max:13767.488) us
cumsum   axis=0     :    71.601 us   +/-20.592 (min:   63.724 / max:  185.169) us  53530.026 us   +/-17.431 (min:53509.281 / max:53601.856) us
cumsum   axis=1     :    64.266 us   +/- 3.725 (min:   61.088 / max:   83.882) us  37469.175 us   +/- 7.742 (min:37457.344 / max:37487.873) us
cumsum   axis=2     :    64.903 us   +/- 4.328 (min:   61.789 / max:   86.990) us  41359.191 us   +/-61.012 (min:41322.079 / max:41601.791) us
cumsum   axis=3     :    46.353 us   +/- 4.449 (min:   43.463 / max:   62.435) us  38549.706 us   +/- 9.629 (min:38543.297 / max:38612.896) us
```

PR #2562, opened 2019-10-23
Refactor CUB to support an explicit `axis` argument; Fix alignments for Thrust's complex types

leofang:
Part of #2519.

This PR does three things:
1. Refactor `cupy/cuda/cub.pyx` and `cupy/cuda/cupy_cub.*` to eliminate a lot of redundant lines for making the No.2 change below much easier;
2. Support `sum(axis=...)`, `min(axis=...)`, and `max(axis=...)` using CUB's **device segmented reduce** API. Because this API can only reduce over contiguous segments, any reduction with non-contiguous axes (e.g., `axis=(0, 2)` for `ndim=3`) is still deferred to the old reduction kernel. 
3. Enforce the alignments for Thrust's complex types to ensure performance

(I tried making a device-device copy to transpose the array at commit 31304f4 for non-contiguous cases, but the performance was just killed by copy. I can post an `nvvp` screenshot if anyone is interested.)

~~Note that for complex numbers with an explicit `axis`, the CUB performance may not be always better than the old reduction, so I added a warning. I believe this is because the hard-coded `DeviceReducePolicy` in `cub/device/dispatch/dispatch_reduce.cuh` saturated the shared memory and caused too many memory transactions when a contiguous block has too many elements, but I don't know how to specialize it for complex numbers with a smaller block size and/or fewer items per thread, so let us leave a performance-tuning PR for the future.~~ 

***UPDATE***: this suboptimal performance is fixed by properly aligning the data (see No.3 above).

attn: @grlee77, @anaruse

PR #2919, opened 2020-01-07
Support CUB prefix sum & product

leofang:
In the demo below, CUB gets about a 4x speed-up on a P100:
```python
>>> import cupy as cp
>>> import cupyx
>>> 
>>> a = cp.random.random(1000000)
>>>
>>> cp.cuda.cub_enabled  =False
>>> print(cupyx.time.repeat(cp.cumprod, (a, )))
cumprod             :   428.775 us   +/-57.870 (min:  396.988 / max:  976.234) us    471.108 us   +/-45.524 (min:  449.888 / max:  992.736) us
>>> cp.cuda.cub_enabled = True
>>> print(cupyx.time.repeat(cp.cumprod, (a, )))
cumprod             :   103.053 us   +/-19.460 (min:   84.242 / max:  229.079) us    132.358 us   +/-17.695 (min:  113.888 / max:  250.048) us
>>>
>>> cp.cuda.cub_enabled = False
>>> print(cupyx.time.repeat(cp.cumsum, (a, )))
cumsum              :   428.253 us   +/-55.802 (min:  398.817 / max:  978.976) us    470.163 us   +/-43.695 (min:  450.464 / max:  995.872) us
>>> cp.cuda.cub_enabled = True
>>> print(cupyx.time.repeat(cp.cumsum, (a, )))
cumsum              :   103.998 us   +/-17.858 (min:   86.243 / max:  234.290) us    133.392 us   +/-16.486 (min:  115.808 / max:  256.320) us
```

TODO:
- [ ] compare performance with `cupy.core._routines_math.scan()` (Rel: #2905)
- [ ] add a CUB call path to `cupy.core._routines_math.scan()`
- [x] check if CUB handles `NaN`s correctly
- [ ] add tests (**Question:** to here or to #2598?)

The first two action items are for speeding up certain fancy indexing cases.

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -286,61 +286,35 @@ def device_csrmv(int n_rows, int n_cols, int nnz, ndarray values,
     return y
 
 
-def device_scan(ndarray x, int op, dtype=None, ndarray out=None):
-    cdef ndarray y
+def device_scan(ndarray x, int op):
     cdef memory.MemoryPointer ws
     cdef int dtype_id, x_size
-    cdef tuple shape
     cdef size_t ws_size
     cdef void *x_ptr
-    cdef void *y_ptr
     cdef void *ws_ptr
     cdef Stream_t s
 
-    if out is not None and out.shape != x.shape:
-        raise ValueError('shape mismatch')
     if op != CUPY_CUB_CUMSUM and op != CUPY_CUB_CUMPROD:
         raise ValueError('only CUPY_CUB_CUMSUM and CUPY_CUB_CUMPROD '
                          'are supported.')
 
-    # determine dtype
-    x_dtype = x.dtype
-    if dtype is None:
-        dtype = x_dtype
-    # TODO(leofang): how to determine system's default integer type?
-    if dtype in (numpy.int8, numpy.int16, numpy.int32):
-        dtype = numpy.int64
-    elif dtype in (numpy.uint8, numpy.uint16, numpy.uint32):
-        dtype = numpy.uint64
-    else:
-        dtype = x_dtype
-
     # determine shape: x is either 1D (with axis=None,0) or ND but ravelled.
     x_size = <int>x.size
-    shape = (x_size,)
-
-    x = x.astype(dtype).reshape(shape)
-    if out is not None:
-        out = out.astype(dtype).reshape(shape)
-        y = out
-    else:
-        y = ndarray(shape, dtype)
-
     if x_size == 0:
-        return y
+        return x
 
     x = _internal_ascontiguousarray(x)
     x_ptr = <void *>x.data.ptr
-    y_ptr = <void *>y.data.ptr
     s = <Stream_t>stream.get_current_stream_ptr()
-    dtype_id = _get_dtype_id(dtype)
-    ws_size = cub_device_scan_get_workspace_size(x_ptr, y_ptr, x_size, s,
+    dtype_id = _get_dtype_id(x.dtype)
+    ws_size = cub_device_scan_get_workspace_size(x_ptr, x_ptr, x_size, s,
                                                  op, dtype_id)
     ws = memory.alloc(ws_size)
     ws_ptr = <void *>ws.ptr
     with nogil:
-        cub_device_scan(ws_ptr, ws_size, x_ptr, y_ptr, x_size, s, op, dtype_id)
-    return y
+        # the scan is in-place
+        cub_device_scan(ws_ptr, ws_size, x_ptr, x_ptr, x_size, s, op, dtype_id)
+    return x
 
 
 cdef bint _cub_device_segmented_reduce_axis_compatible(
@@ -445,33 +419,19 @@ def cub_reduction(arr, op, axis=None, dtype=None, out=None, keepdims=False):
     return None
 
 
-def can_use_device_scan(x_dtype, dtype=None):
-    # cub_device_scan seems buggy for complex128:
-    # https://github.com/cupy/cupy/pull/2919#issuecomment-574633590
-    if dtype is not None:
-        return (dtype in CUB_support_dtype and dtype != numpy.complex128)
-    return (x_dtype in CUB_support_dtype and x_dtype != numpy.complex128)
-
-
-def cub_scan(arr, op, axis=None, dtype=None, out=None):
-    """Perform a prefix scan using CUB.
+def cub_scan(arr, op):
+    """Perform an (in-place) prefix scan using CUB.
 
     If the specified scan is not possible, None is returned.
     """
-    if axis is not None:
-        if sequence.PySequence_Check(axis):
-            raise TypeError(
-                "'tuple' object cannot be interpreted as an integer")
-        if not (-arr.ndim <= axis < arr.ndim):
-            raise _AxisError('axis(={}) out of bounds'.format(axis))
-        if arr.ndim != 1 or axis != 0:
-            return None
-
     if op < CUPY_CUB_CUMSUM or op > CUPY_CUB_CUMPROD:
         return None
 
-    if can_use_device_scan(arr.dtype, dtype):
-        return device_scan(arr, op, dtype, out)
+    # cub_device_scan seems buggy for complex128:
+    # https://github.com/cupy/cupy/pull/2919#issuecomment-574633590
+    x_dtype = arr.dtype
+    if (x_dtype in CUB_support_dtype and x_dtype != numpy.complex128):
+        return device_scan(arr, op)
 
     return None
 
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -290,6 +290,7 @@ def device_scan(ndarray x, int op, dtype=None, ndarray out=None):
     cdef ndarray y
     cdef memory.MemoryPointer ws
     cdef int dtype_id, x_size
+    cdef tuple shape
     cdef size_t ws_size
     cdef void *x_ptr
     cdef void *y_ptr
@@ -302,6 +303,7 @@ def device_scan(ndarray x, int op, dtype=None, ndarray out=None):
         raise ValueError('only CUPY_CUB_CUMSUM and CUPY_CUB_CUMPROD '
                          'are supported.')
 
+    # determine dtype
     x_dtype = x.dtype
     if dtype is None:
         dtype = x_dtype
@@ -312,23 +314,26 @@ def device_scan(ndarray x, int op, dtype=None, ndarray out=None):
         dtype = numpy.uint64
     else:
         dtype = x_dtype
-    x = x.astype(dtype)
-    if out is not None:
-        y = out.astype(dtype)
-    else:
-        y = ndarray(x.shape, dtype)
 
-    if x.size == 0:
-        if out is not None:
-            out = out.reshape((0,))
-            y = out
-        else:
-            y = ndarray((0,), dtype=dtype)
+    # determine shape: x is either 1D (with axis=None,0) or ND but ravelled.
+    x_size = <int>x.size
+    if x.ndim != 1:
+        shape = (x_size,)
+    else:
+        shape = x.shape
+
+    x = x.astype(dtype).reshape(shape)
+    if out is not None:
+        out = out.astype(dtype).reshape(shape)
+        y = out
+    else:
+        y = ndarray(shape, dtype)
+
+    if x_size == 0:
         return y
 
     x = _internal_ascontiguousarray(x)
     x_ptr = <void *>x.data.ptr
-    x_size = <int>x.size
     y_ptr = <void *>y.data.ptr
     s = <Stream_t>stream.get_current_stream_ptr()
     dtype_id = _get_dtype_id(dtype)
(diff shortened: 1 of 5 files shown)


grlee77:
Nice. I tried this out and the only problem I encountered was in the shape of the returned result when `axis=None` and the input is nD. (The output should be a ravelled array)

Simple example with CUB enabled/disabled (the disabled behavior matches NumPy):

```Python
import cupy as cp
import numpy as np

a = cp.testing.shaped_arange((4, 5), cp, float)
cp.cumsum(a)
# array([[  1.,   3.,   6.,  10.,  15.],
#        [ 21.,  28.,  36.,  45.,  55.],
#        [ 66.,  78.,  91., 105., 120.],
#        [136., 153., 171., 190., 210.]])

cp.cuda.cub_enabled = False
cp.cumsum(a)
# array([  1.,   3.,   6.,  10.,  15.,  21.,  28.,  36.,  45.,  55.,  66.,
#         78.,  91., 105., 120., 136., 153., 171., 190., 210.])

**Comment by leofang** [comment]
Oops, thanks @grlee77! I read the spec wrong...

**Comment by emcastillo** [comment]
I was actually working on this :D.
I let you take care of it better :).
One of the things I saw is that CUB scanning does not support axis selection, so implementing that correctly may be a bit trickier.

**Comment by leofang** [comment]
Oh sorry for duplicate work @emcastillo. Which branch is it? I could cherry pick your code.
> One of the things I saw is that CUB scanning does not support axis selection, so implementing that correctly may be a bit trickier.

I don’t think CUB can support `axis != None`...

**Comment by emcastillo** [comment]
No worries, your PR is going to be better than mine so do the cub part and I will take care of improving the non-cub kernels.

**Comment by leofang** [comment]
@emcastillo This PR should be considered jointly with your #2907. Looks like `cupy.core._routines_math.scan()` is really good. It is slower than CUB only when there are >= 10^6 elements (for `float64` in this case):
```python
>>> a = cp.random.random(1000)
>>> print(cupyx.time.repeat(cp.cuda.cub.device_scan, (a, 5)))
device_scan         :    98.163 us   +/-16.937 (min:   76.599 / max:  222.032) us    106.869 us   +/-18.320 (min:   77.120 / max:  236.768) us
>>> print(cupyx.time.repeat(cp.core._routines_math._scan_for_test, (a, )))
_scan_for_test      :    67.176 us   +/- 9.349 (min:   49.020 / max:  147.861) us     76.206 us   +/-10.508 (min:   56.064 / max:  163.200) us
>>> 
>>> a = cp.random.random(100000)
>>> print(cupyx.time.repeat(cp.cuda.cub.device_scan, (a, 5)))
device_scan         :    97.013 us   +/-17.426 (min:   76.346 / max:  215.663) us    105.617 us   +/-18.885 (min:   83.296 / max:  300.320) us
>>> print(cupyx.time.repeat(cp.core._routines_math._scan_for_test, (a, )))
_scan_for_test      :    67.560 us   +/- 9.815 (min:   49.441 / max:  181.684) us     76.555 us   +/-11.008 (min:   56.416 / max:  198.016) us
>>> 
>>> a = cp.random.random(1000000)
>>> print(cupyx.time.repeat(cp.cuda.cub.device_scan, (a, 5)))
device_scan         :    94.037 us   +/-17.865 (min:   74.958 / max:  278.390) us    124.621 us   +/-16.371 (min:  106.752 / max:  301.504) us
>>> print(cupyx.time.repeat(cp.core._routines_math._scan_for_test, (a, )))
_scan_for_test      :    90.543 us   +/-17.130 (min:   70.443 / max:  201.541) us    123.582 us   +/-10.805 (min:  107.584 / max:  223.040) us
>>> 
>>> a = cp.random.random(10000000)
>>> print(cupyx.time.repeat(cp.cuda.cub.device_scan, (a, 5)))
device_scan         :    82.406 us   +/-12.235 (min:   74.969 / max:  204.807) us    650.291 us   +/- 6.245 (min:  644.096 / max:  714.784) us
>>> print(cupyx.time.repeat(cp.core._routines_math._scan_for_test, (a, )))
_scan_for_test      :    78.777 us   +/- 9.734 (min:   72.235 / max:  196.004) us    822.750 us   +/- 8.690 (min:  816.992 / max:  901.280) us
```
So, if `cupy.core._routines_math.scan()` is tweaked a bit (simply by changing the accumulator `+=` to `*=` I guess), perhaps in addition to `cumsum()` it can also do `cumprod()`, and we either wouldn't need CUB at all, or would switch to CUB only when the array is excessively large? 

(Actually I already cheated a bit in the above comparison by using `cp.cuda.cub.device_scan()` instead of `cp.cuda.cub.cub_scan()`: the latter adds about 8 us on my machine due to extra checks...)

leofang:
> we either wouldn't need CUB at all, or would switch to CUB only when the array is excessively large?

We could also let users decide if they want to use CUB scan, as usual by toggling `cp.cuda.cub_enabled`.

emcastillo:
I wouldn't set sub-optimal implementations by default

leofang:
Another thing: CUB `cumsum` and `cumprod` seems to be buggy for complex numbers. I am not sure why, but this is deterministically reproducible. The cumulation stopped at the 449-th element of the result:
```python
>>> import cupy as cp
>>> a = cp.random.random(500) + 1j*cp.random.random(500)
>>> cp.cuda.cub_enabled = True
>>> b = cp.cumsum(a)  # wrong
>>> cp.cuda.cub_enabled = False
>>> c = cp.cumsum(a)  # correct
>>> cp.allclose(b[0:447],c[0:447])
array(True)
>>> cp.allclose(b[0:448],c[0:448])
array(True)
>>> cp.allclose(b[0:449],c[0:449])
array(False)
>>> b[448] == a[448]
array(True)
>>> c[448] == a[448]
array(False)
```
Looking into this...

leofang:
> Another thing: CUB `cumsum` and `cumprod` seems to be buggy for complex numbers. 

Two observations:
1. This bug is `complex128`-only; for `complex64` arrays CUB works fine, so I suspect it's due to CUB's internal handling of data size exceeding 64 bits.
2. Adjusting the items per thread in this line
https://github.com/NVlabs/cub/blob/c3cceac115c072fb63df1836ff46d8c60d9eb304/cub/device/dispatch/dispatch_scan.cuh#L177
will cause the bug appear in different elements (I changed from 15 to 12 and the bug started in the element No.3xx instead of 448).

@emcastillo Due to the above observations, I tend to think this is a bug in CUB (which I have not identified yet), and so I will disable CUB prefix scan for `complex128` and proceed.

emcastillo:
I have been working on adapting the fast kernel in scan to do the batched cumsum and cumprod.
It's been a nightmare but i am pretty close to the end :D

leofang:
Nice. We can then do a benchmark to decide if this PR is to be kept or closed.

cupy/cuda/cub.pyx
```diff
@@ -372,8 +372,7 @@ def can_use_device_segmented_reduce(int op, x_dtype, Py_ssize_t ndim,
                                                         order)
 
 
-cdef _cub_reduce_dtype_compatible(x_dtype, int op, dtype=None,
-                                  bint segmented=False):
+cdef _cub_reduce_dtype_compatible(x_dtype, int op, dtype=None):
```
leofang: This is to correct a mistake I made in #2562.

cupy/math/sumprod.py
```diff
@@ -147,8 +152,15 @@ def _proc_as_batch(proc, x, axis):
...
-def _cum_core(a, axis, dtype, out, kern, batch_kern):
+def _cum_core(a, axis, dtype, out, kern, batch_kern, *, op=None):
     a = cupy.asarray(a)
+
+    if cupy.cuda.cub_enabled:
+        # result will be None if the scan is not compatible with CUB
+        result = cub.cub_scan(a, op, axis, dtype, out)
+        if result is not None:
+            return result
+
```
emcastillo: Does this work for inplace?

leofang:
This is a bit more complicated than I previously thought; even the original CuPy implementation might not be fully compliant. 

Because the NumPy doc is not very clear, I took a look at the NumPy's implementation `PyUFunc_GenericReduction()` and `PyUFunc_Accumulate()`, and this is my current understanding for the interplay between the `dtype` and `out` arguments in `cumsum()`/`cumprod()` (please correct me if I am wrong):
- if `dtype` is `None`:
  - if `out` is `None`:
    - `input.dtype` is used for both intermediate calculation and final output (with the caveat that integer types need to be promoted)
  - else:
    - `out.dype` is used for both intermediate calculation and final output
- else:
  - if `out` is `None`:
    - `dtype` is used for both intermediate calculation and final output
  - else:
    - `dtype` is used for intermediate calculation, and `out.dtype` is used for final output

Currently CuPy's `_cum_core` only takes care of the first branch above. @emcastillo should I follow the current CuPy behavior as is, or should I make the CUB scan fully NumPy-compliant?

I suspect we might have the same compliance issue in other reduction routines (not sure), but that should be addressed in a separate PR if any...

cupy/math/sumprod.py
leofang: It might, let me look into this.

emcastillo:
I think that behavior should be fixed in another PR.

leofang:
Sounds good. I'll then avoid dealing with this and let the current `_cum_core()` handle it. Then, in the next PR we just fix `_cum_core()`.

leofang:
@emcastillo Thanks for the heads-up! I guess you are right: CUB supports in-place scan. This simplifies things quite a bit.

I verified this support in a few levels:
1. Test it directly with the new change 79780e1
2. In [Thrust's documentation](https://docs.nvidia.com/cuda/thrust/index.html#prefix-sums) in-place scans are used as an example. Since CUB is one of the Thrust's backends, it should also work.
3. By inspecting CUB's code: The `ConsumeTile()` implementation in `cub/cub/agent/agent_scan.cuh` seems to permit in-place operation.

cupy/math/sumprod.py
leofang: See https://github.com/cupy/cupy/pull/2919#issuecomment-575968811

grlee77:
Hi @leofang. The dtype behavior you describe above (https://github.com/cupy/cupy/pull/2919#issuecomment-575705775) is also what I recently attempted to implement for cupy.mean in #2903. Seeing now that this is more common across reduction functions, then perhaps the logic should move out into a separate helper function.

leofang:
Thanks for bringing it up, Gregory 🙂 I just took a quick look at it. While it'd be nice to have a separate helper as you said, the rules there are slightly different from https://github.com/cupy/cupy/pull/2919#issuecomment-575705775. Maybe we need a switch or flag in the helper to determine the desired behavior? 

Anyway, if you can make that helper in your PR, @emcastillo and I will be able to follow in #2907 and here. We can move the discussion there. Thanks!

leofang:
@emcastillo I suggest we consider the PRs in the following order: #2919 (this PR) -> #2907 for two reasons.

First, I do not deal with the complexity of `out` and `dtype`, as outlined in https://github.com/cupy/cupy/pull/2919#issuecomment-575705775, and simply let the caller functions  (`cumsum` and `cumprod`) handle it, whereas in #2907 the caller code is refactored. 

Second, while Gregory's suggestion on a separate helper is tempting, I don't see immediately how such helper can be applicable since the rules vary across different functions.

If we want to make the implementation more NumPy-compliant, we can make another PR to address https://github.com/cupy/cupy/pull/2919#issuecomment-575705775 after both are merged, as you suggested earlier. 

If you agree with my suggestion, the only question I have for you is:
> - [ ] add tests (**Question:** to here or to #2598?)

Thanks 🙂

emcastillo:
Lets do that

leofang:
Cool, if we don’t need to add tests here, this PR is ready.

emcastillo:
Lets add tests in #2598 thanks!

cupy/cuda/cub.pyx
```diff
@@ -282,6 +286,37 @@ def device_csrmv(int n_rows, int n_cols, int nnz, ndarray values,
...
 
+def device_scan(ndarray x, int op):
+    cdef memory.MemoryPointer ws
+    cdef int dtype_id, x_size
+    cdef size_t ws_size
+    cdef void *x_ptr
+    cdef void *ws_ptr
+    cdef Stream_t s
+
+    if op != CUPY_CUB_CUMSUM and op != CUPY_CUB_CUMPROD:
```
emcastillo: Can you rewrite this kind of checks as
`if op not in (CUPY_CUB_CUMSUM, CUPY_CUB_CUMPROD):`

Thanks

leofang: No problem. This is done in 02059fb. Just curious: is this for readability or performance?

emcastillo: readability

leofang:
I fixed the compilation error. Sorry for my stupid mistake 😅

cupy/cuda/cub.pyx
```diff
@@ -385,6 +421,23 @@ def cub_reduction(arr, op, axis=None, dtype=None, out=None, keepdims=False):
     return None
 
 
+def cub_scan(arr, op):
+    """Perform an (in-place) prefix scan using CUB.
+
+    If the specified scan is not possible, None is returned.
+    """
+    if op < CUPY_CUB_CUMSUM or op > CUPY_CUB_CUMPROD:
```
emcastillo: `if op not in (CUPY_CUB_CUMSUM, CUPY_CUB_CUMPROD):`

emcastillo: I will fix it later in my pr

emcastillo:
Jenkins, test this please

cupy/cuda/cub.pyx
leofang: Oh, sorry for overlooking this @emcastillo 😓

leofang:
Thanks @emcastillo!

PR #2732, opened 2019-11-28
Merge reduction implementations

niboshi:
Merges the logic of `ReductionKernel` and `simple_reduction_function`.

PR #2920, opened 2020-01-07
Merge elementwise implementations

niboshi:
Elementwise version of #2732.
Merges the logic of `ElementwiseKernel` and `ufunc`.


niboshi:
@emcastillo 
I think refactoring on elementwise-kernels is finished. May I hear a brief comment on the last commit (b746a0e791d765fabe5b7ddf36c14e8a1164a040)?
Some code is still duplicated (e.g. `_preprocess_args`) because it's also used from `reduction.pyx`.
If basic idea is OK, I'll try to update `reduction.pyx` too.
Reconsideration may happen at that moment.

emcastillo:
Let me take a look

emcastillo:
At first glance, the PR changes look fine and consistent.
 But one of the things I am worried about is that with `ufunc` and `elementwise` being merged, it may complicate adding support for [generalized functions](https://docs.scipy.org/doc/numpy-1.13.0/reference/c-api.generalized-ufuncs.html) since these can do reductions as well.

In numpy ufuncs are a special case of the generalized ufunc. (described by just the signature).
So I think that having a separated interface for the ufunc and argument checkings, signatures etc decoupled from the Elementwise kernel may keep things simpler. However, my knowledge of all these internals is limited and I am not sure of to what extent they can be a problem. @toslunar @asi1024 any comments on this?.

niboshi:
@emcastillo 
I agree that element kernels and reduction kernels can be unified.
However, as it's difficult to do that all at once, you can consider this PR as a first step toward that.
(currently I don't have a concrete idea how they all will be unified though.)

>So I think that having a separated interface for the ufunc and argument checkings, signatures etc decoupled from the Elementwise kernel may keep things simpler.

I don't understand this point.
My hope is all the kernel implementations (`ufunc`, `ElementwiseKernel`, `_SimpleReductionKernel`, `ReductionKernel`) should be able to be generalized to some extent.
Could you elaborate why you think `ufunc` and `ElementwiseKernel` should be separate?

emcastillo:
If reduction and elementwise are going to be unified then there is no problem at all 😄

emcastillo:
This is a long one, so I just want to be sure that I am getting everything right. 
What you have done here is:
1) Rewrite the kernel argument managing logic and abstract it into several classes instead of raw tuples and module level functions.
2) Create a base _AbstractElementwiseKernel that essentially groups all the ufunc related `_kernel` module functions.

I think it might be better to split this in two PRs, one with just `_ArgKind`, `_ArgInfo` and `_TypeMap` and another one with the rest of changes. But if you prefer to keep them as one PR I won't be against it.

niboshi:
@emcastillo 
Thank you for a comment.
I think your suggestion is very right. I'll separate 1. as another PR while this PR is kept (I mean this PR will depend on it).

emcastillo:
Thanks!

niboshi:
Btw, `_Args` class will be classified in 2., as it's abstraction between `ufunc` and `ElementwiseKernel`.

asi1024:
@niboshi Could you resolve conflicts?

niboshi:
@asi1024 I'm planning to rebase after #2918.

niboshi:
After some internal discussion, we have decided to include scalar refactoring (#2918) in this PR, because it's difficult to discuss separately.

I'm planning to introduce `Arg` class to wrap arguments and provide uniform interface across types/kernel implementations. `Arg.get_pointer()` which returns `CPointer` is one example.

My current plan (although subject to change) is:

* Implement `Arg` in `function.pyx`. Note that `Function.__call__` has to accept `Arg`s as well as raw ndarrays and scalars.
* Merge `_ArgInfo` into `Arg`. Note that `_ArgInfo` was meant to be hashable but `Arg` is not. So, I am considering adding something like `Arg.get_immutable_key()`.

niboshi:
Reduction is still wip, but elementwise tests pass.

asi1024:
Jenkins, test this please.

cupy/core/_kernel.pyx
```diff
@@ -187,7 +187,7 @@ cdef class _Args:
         for obj in objs:
             arg = Arg.from_obj(obj)
             # Check device of an ndarray
-            if arg.is_ndarray():
+            if isinstance(obj, ndarray):
```
niboshi: Please do not revert this line.
The purpose of `is_ndarray` is to avoid repeated `isinstance`.

cupy/core/_reduction.pyx
```diff
@@ -299,7 +299,7 @@ cdef class _AbstractReductionKernel:
                 IndexerArg(in_shape),
                 IndexerArg(out_shape),
                 # block_stride is passed as the last argument.
-                #_scalar.CScalar.from_int32(block_stride),
+                # _scalar.CScalar.from_int32(block_stride),
```
niboshi: This comment can simply be removed.

cupy/core/_kernel.pyx
asi1024: When `obj` is a cuda_array_interface object, `arg.is_ndarray()` will be `Ture` but this condition should be evaluated to `False`.

niboshi: Do you mean that because the line below would not work?
Then, let's change the line instead.

```diff
-                _check_array_device_id(obj, device_id)
+                _check_array_device_id(arg.obj, device_id)
```

Basically, `_cuda_array_interface__` handling should be confined in `Arg.from_obj` and no other code should  care about it.

cupy/core/_reduction.pyx
```diff
@@ -284,21 +284,29 @@ cdef class _AbstractReductionKernel:
...
             + out_args
             + [
-                _carray.Indexer(in_shape),
-                _carray.Indexer(out_shape),
+                IndexerArg(in_shape),
+                IndexerArg(out_shape),
                 # block_stride is passed as the last argument.
-                _scalar.CScalar.from_int32(block_stride),
+                # _scalar.CScalar.from_int32(block_stride),
+                Arg.from_obj(numpy.int32(block_stride)),
```
niboshi: Please note that this fix is just tentative, because it might have performance implication.
Perhaps you should either
* Make a measurement and confirm it's negligible (I think it's unlikely)
* Make a fast shortcut constructor of `Arg` that does not require `numpy.int32` (like `CScalar.from_int32`).
* Or perhaps treat this argument (and indexers) specially, instead of just the same way as array/scalar arguments. I think this option is desirable because it doesn't make sense to include their `get_immutable_key()`s in the cache key. But it would require a bit further design reconsideration.

asi1024: I left a TODO comment. I'll address your comment in another PR.

cupy/core/_kernel.pyx
asi1024: Fixed.

asi1024:
Jenkins, test this please.

cupy/core/_reduction.pyx
niboshi: OK. Please note that that means this PR introduces performance degradation.

asi1024:
(Discussed offline with @okuta)
This PR conflicts with #2702 and merged it first, but performance improvement of this kind should be applied after this PR is merged. So, I decided to revert #2702 in this PR temporarily and send the performance improvement PR later.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

cupy/cuda/function.pyx
```diff
@@ -244,3 +261,194 @@ cdef class LinkState:
...
+        if hasattr(obj, '__cuda_array_interface__'):
+            # __cuda_array_interface__
+            obj = core._convert_object_with_cuda_array_interface(obj)
+            arg = NdarrayArg(obj, c_contiguous=obj.flags.c_contiguous)
+            return arg
+        if _scalar.is_scalar(obj):
+            # scalar
+            arg = ScalarArg(obj)
+            return arg
+        if isinstance(obj, TextureObject):
```
leofang: Thanks, @asi1024! Could you also include `MemoryPointer` in this category? I have an ongoing PR that would benefit having this. (I could also add this separately later.)
```suggestion
        if isinstance(obj, (TextureObject, MemoryPointer)):
```
(need to import `MemoryPointer` first).

asi1024: I'll fix it after your PR is merged. If this PR is merged before yours, please fix this line in your PR.

leofang: OK no worries 👍 I am still preparing it...

asi1024: Thanks for your notification!

emcastillo:
Benchmarked ufuncs,

Consistent increase in CPU time from 10us to 20us
This PR
```
time_sqrt            - case        100                    :    CPU:   20.238 us   +/- 0.263 (min:   19.956 / max:   20.900) us     GPU:   24.029 us   +/- 0.731 (min:   23.264 / max:   25.504) us
```
Master
```
time_sqrt            - case        100                    :    CPU:   11.371 us   +/- 0.247 (min:   11.143 / max:   11.853) us     GPU:   14.662 us   +/- 1.109 (min:   13.984 / max:   17.888) us
```

This time diff holds for all ufuncs, elementwise kernels, reduction and raw kernels.

kmaehashi:
Reverted in #3296 for performance issue.

PR #2921, opened 2020-01-07
Fix `cupy.cuda.cub.device_segmented_reduce()` not being used

leofang:
...I guess we forgot it back in #2725, @grlee77? 

It'd be really nice if all these CUB bug fixes could be avoided by having CIs catching them in the first place...

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -98,7 +98,7 @@ cpdef Py_ssize_t _preprocess_array(ndarray arr, tuple reduce_axis,
 
 def device_reduce(ndarray x, int op, tuple out_axis, out=None,
                   bint keepdims=False):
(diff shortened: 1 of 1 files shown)


leofang:
This bug is present in both the `v7` and `master` branches, but not in the v7.0.0 release.

leofang:
> Also +1 on having CUB testing on CI, but I don't know that in this case it would have helped? I think the results were still correct, it was just not using CUB in some cases where it was possible.

Right. I was thinking testing `cupy.cuda.cub.can_use_*()`, but that would not help. We could pick a few known cases in which CUB outperforms, measure the timings, and report failure if speedup is not observed. This was part of original intension in #2598 but later abandoned.

emcastillo:
Jenkins, test this please

PR #2923, opened 2020-01-08
Fix complex type checks in `_correlate_or_convolve`

### LINKED ISSUE #2916: Windows: module 'numpy' has no attribute 'complex256'
In file: [cupy/cupyx/scipy/ndimage/filters.py] (https://github.com/cupy/cupy/blob/master/cupyx/scipy/ndimage/filters.py)

Line 83:  ` if input.dtype in (numpy.complex64, numpy.complex128, numpy.complex256):`

There is a check for numpy.complex256. On Windows there is no numpy.complex256, so this line leads to the following error:

module 'numpy' has no attribute 'complex256'

a simple solution could be to use the function [numpy.iscomplexobj
](https://docs.scipy.org/doc/numpy/reference/generated/numpy.iscomplexobj.html#numpy.iscomplexobj) or (maybe faster) to stringify the dtype of the array and check for substring "complex"

emcastillo:
Fixes #2916

### DIFF
--- a/cupyx/scipy/ndimage/filters.py
+++ b/cupyx/scipy/ndimage/filters.py
@@ -1,5 +1,3 @@
-import numpy
-
 import cupy
 from cupy import util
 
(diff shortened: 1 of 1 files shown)


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #2925, opened 2020-01-08
Support grid synchronization in `RawKernel` and `RawModule`

asi1024:
Support `cooperative_groups` in `RawKernel`.

### DIFF
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -68,7 +68,10 @@ cdef class RawKernel:
                 bytes.
 
         """
-        self.kernel(grid, block, args, **kwargs)
+        self.kernel(
+            grid, block, args,
+            enable_cooperative_groups=self.enable_cooperative_groups,
+            **kwargs)
 
     @property
     def kernel(self):
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -237,7 +237,7 @@ cdef class RawModule:
         Each kernel in ``RawModule`` possesses independent function attributes.
     """
     def __init__(self, *, code=None, path=None, options=(), backend='nvrtc',
-                 translate_cucomplex=False):
+                 translate_cucomplex=False, enable_cooperative_groups=False):
         if (code is None) == (path is None):
             raise TypeError(
                 'Exactly one of `code` and `path` keyword arguments must be '
@@ -251,11 +251,13 @@ cdef class RawModule:
 
         self.code = code
         self.cubin_path = path
+        self.enable_cooperative_groups = enable_cooperative_groups
 
         if self.code is not None:
             self.module = cupy.core.core.compile_with_cache(
                 code, options, prepend_cupy_headers=False, backend=backend,
-                translate_cucomplex=translate_cucomplex)
+                translate_cucomplex=translate_cucomplex,
+                enable_cooperative_groups=self.enable_cooperative_groups)
             self.options = options
             self.backend = backend
             self.translate_cucomplex = translate_cucomplex
@@ -280,8 +282,10 @@ cdef class RawModule:
         if name in self.kernels:
             return self.kernels[name]
         else:
-            ker = RawKernel(None, name, self.options, self.backend,
-                            translate_cucomplex=self.translate_cucomplex)
+            ker = RawKernel(
+                None, name, self.options, self.backend,
+                translate_cucomplex=self.translate_cucomplex,
+                enable_cooperative_groups=self.enable_cooperative_groups)
             ker._kernel = self.module.get_function(name)
             self.kernels[name] = ker
             return ker
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -33,7 +33,7 @@ cdef class RawKernel:
     """
 
     def __init__(self, code, name, options=(), backend='nvrtc', *,
-                 translate_cucomplex=False):
+                 translate_cucomplex=False, grid_sync=False):
         if isinstance(code, six.binary_type):
             code = code.decode('UTF-8')
         if isinstance(name, six.binary_type):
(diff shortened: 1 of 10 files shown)


cupy/core/raw.pyx
```diff
@@ -33,7 +33,7 @@ cdef class RawKernel:
     """
 
     def __init__(self, code, name, options=(), backend='nvrtc', *,
-                 translate_cucomplex=False):
+                 translate_cucomplex=False, grid_sync=False):
```
niboshi: I think grid synchronization is only a part of `cooperative_groups`. How about something like `enable_cooperative_groups`?

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -519,3 +536,23 @@ def test_cuDoubleComplex(self):
...
+
+class TestRawGridSync(unittest.TestCase):
+
+    def setUp(self):
+        global _test_cache_dir
+        _test_cache_dir = tempfile.mkdtemp()
+        os.environ['CUPY_CACHE_DIR'] = _test_cache_dir
+
+        self.kern_grid_sync = cupy.RawKernel(
+            _test_grid_sync, 'test_grid_sync',
```
niboshi: As `_test_grid_sync` is used only once, it's better to write here directly.

cupy/cuda/compiler.py
```diff
@@ -276,18 +276,26 @@ def get_cache_dir():
 _empty_file_preprocess_cache = {}
 
 
-def compile_with_cache(source, options=(), arch=None, cache_dir=None,
-                       extra_source=None, backend='nvrtc'):
+def compile_with_cache(
+        source, options=(), arch=None, cache_dir=None, extra_source=None,
+        backend='nvrtc', grid_sync=False):
```
niboshi: Isn't this function public? It may be better to make the argument keyword-only.

cupy/core/raw.pyx
asi1024: Fixed.

tests/cupy_tests/core_tests/test_raw.py
asi1024: Fixed.

cupy/cuda/compiler.py
asi1024: Fixed.

cupy/cuda/compiler.py
```diff
@@ -276,18 +276,26 @@ def get_cache_dir():
...
+
+    if enable_cooperative_groups and backend != 'nvcc':
+        raise ValueError(
+            'Cooperative groups is supported only in NVCC backend.')
+
     if runtime.is_hip:
-        return _compile_with_cache_hipcc(source, options, arch, cache_dir,
-                                         extra_source)
+        return _compile_with_cache_hipcc(
+            source, options, arch, cache_dir, extra_source)
```
niboshi: How about raising an error in this case?

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -519,3 +519,34 @@ def test_cuDoubleComplex(self):
...
         ker((grid,), (block,), (a, out_down))
         assert (out_down == a.astype(cupy.complex64)).all()
+
+
+class TestRawGridSync(unittest.TestCase):
+
+    def setUp(self):
+        global _test_cache_dir
+        _test_cache_dir = tempfile.mkdtemp()
+        os.environ['CUPY_CACHE_DIR'] = _test_cache_dir
```
niboshi: Please write the corresponding `tearDown` logic.

cupy/cuda/compiler.py
asi1024: Fixed.

tests/cupy_tests/core_tests/test_raw.py
asi1024: Fixed.

asi1024:
PTAL.

niboshi:
Thanks
Jenkins, test this please

leofang:
Would this be applicable to RawModule as well?

asi1024:
@leofang I have not confirmed yet, but maybe yes.

asi1024:
Supported also in `RawModule`. PTAL.

leofang:
LGTM! Thanks for checking, @asi1024!

cupy/core/raw.pyx
```diff
@@ -30,10 +30,12 @@ cdef class RawKernel:
             `cuComplex.h` or not. If set to ``True``, any code that uses the
             functions from `cuComplex.h` will be translated to its Thrust
             counterpart. Defaults to ``False``.
+        enable_cooperative_groups (bool): Whether the CUDA source requires to
+            use cooperative groups.
```
niboshi: Taking it literally, you may think it must be `False` if you don't use `cooperative_groups`.
Although a bit redundant, how about something like this?

"Whether to enable cooperative groups in the CUDA source. If set to `True`, compile options are configured properly so that cooperative groups can be used from the CUDA source."

cupy/core/raw.pyx
```diff
@@ -229,12 +234,14 @@ cdef class RawModule:
...
```
niboshi: ditto.

niboshi:
@leofang 
Thank you for the comment!

cupy/core/raw.pyx
asi1024: Fixed.

cupy/core/raw.pyx
asi1024: Fixed.

niboshi:
Thanks.
Jenkins, test this please

niboshi:
Please take a look at the error.

asi1024:
@niboshi I confirmed the test failure with large `n` locally and fixed it. PTAL.

niboshi:
Jenkins, test this please

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
@niboshi Fixed test failures.

niboshi:
Thanks.
Jenkins, test this please

niboshi:
👍

PR #2932, opened 2020-01-09
Improve error message for wrong number of arguments in elementwise kernels

niboshi:
Adds expected and actual numbers.
```
TypeError: Wrong number of arguments for 'cupy_add'. It must be either 2 or 3 (with outputs), but given 4.
```


emcastillo:
Jenkins, test this please

PR #2941, opened 2020-01-09
Fix `ParameterInfo` as a cache key

niboshi:
`__hash__` and `__eq__`, required to be suitable as a cache key, were missing in `ParameterInfo`.

You can check the behavior by the following code.
```py
import cupy
a = cupy.core._kernel.ParameterInfo('raw T x', False)
b = cupy.core._kernel.ParameterInfo('raw T x', False)
print(hash(a))
print(hash(b))
print(a == b)
```

After the fix:
```
2788533504290071466
2788533504290071466
True
```

### DIFF
--- a/cupy/core/_kernel.pyx
+++ b/cupy/core/_kernel.pyx
@@ -273,6 +273,22 @@ cdef class ParameterInfo:
             else:
                 raise Exception('Unknown keyword "%s"' % i)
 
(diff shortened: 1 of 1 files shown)


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #2942, opened 2020-01-09
Use asynchronous copy in `cupy.copyto`

### LINKED ISSUE #2159: [BUG] Incorrect behavior of asynchronous D2D copy
`cupy.copyto` misses to apply user specified stream.
Please see following mini repro.

* Conditions
  - CuPy 5.4.0 and 6.0.0rc1
  - Ubuntu 16.04
  - CUDA 10.1
* Code to reproduce
```python
import cupy as cp

def main():
  SIZE = 250000000
  a = cp.empty((1, SIZE), cp.float32)
  b = cp.empty((1, SIZE), cp.float32)
  my_stream = cp.cuda.stream.Stream()

  my_stream.use()
  cp.copyto(a, b) # <- it will use default stream but it should use `my_stream`
  my_stream.synchronize()

  my_stream.use()
  a[:,:] = b # <- use `my_stream`
  my_stream.synchronize()

if __name__ == '__main__':
  main()
```
[nvvp log](https://drive.google.com/file/d/1cc2bkKVTBk60pQVXW0log7BsLK6-RFY2/view?usp=sharing)

takagi:
Part of #2159.

`cupy.copyto` currently uses synchronous DtoD copy in one of its branches, eventually using the default CUDA stream. This PR fixes it to use asynchronous DtoD copy with the current CUDA stream that CuPy manages.

### DIFF
--- a/cupy/manipulation/basic.py
+++ b/cupy/manipulation/basic.py
@@ -55,7 +55,7 @@ def copyto(dst, src, casting='same_kind', where=None):
 
     if where is None:
         if _can_memcpy(dst, src):
-            dst.data.copy_from(src.data, src.nbytes)
+            dst.data.copy_from_async(src.data, src.nbytes)
         else:
             device = dst.device
             with device:


niboshi:
Thanks for the PR.
Does it fix `cupy.dot` (as reported at https://github.com/cupy/cupy/issues/2159#issuecomment-484938869) as well?

niboshi:
And why "no-compat"?

takagi:
> Does it fix cupy.dot (as reported at #2159 (comment)) as well?

No. Let me check it as another problem.

> And why "no-compat"?

It's because that I suppose if a function performs synchronously or asynchronously is contained in its specification. It is not the case here?

takagi:
The H2D copy of `cupy.dot` reported at https://github.com/cupy/cupy/issues/2159#issuecomment-484938869 is triggered by initializing CUBLAS. I suppose it's fairly tiny, ~1us or so, and is not an essential issue to address.

niboshi:
>> Does it fix cupy.dot (as reported at #2159 (comment)) as well?
>
>No. Let me check it as another problem.

OK, then this PR should not close #2159.

>> And why "no-compat"?
>
>It's because that I suppose if a function performs synchronously or asynchronously is contained in its specification. It is not the case here?

I don't think so. It only affects performance and it won't be imcompatible.

niboshi:
>The H2D copy of `cupy.dot` reported at https://github.com/cupy/cupy/issues/2159#issuecomment-484938869 is triggered by initializing CUBLAS. I suppose it's fairly tiny, ~1us or so, and is not an essential issue to address.

Is it CPU time?
If host-device synchronization occurs, it's more problematic than any CPU time.
I don't think that has to be done in this PR though.

niboshi:
Jenkins, test this please

niboshi:
👍

takagi:
> Is it CPU time?
No, it's GPU time that I collected with nvprof and visualized with nvvp. Let me supply its screen shot.

takagi:
> I don't think so. It only affects performance and it won't be incompatible.

Okay!

PR #2943, opened 2020-01-09
Avoid invalid in-place division in CUB-based mean

grlee77:
This bug fix was split out from PR #2903 for easier review. For integer dtypes it is not possible to perform in-place division by a float.

### DIFF
--- a/cupy/core/_routines_statistics.pyx
+++ b/cupy/core/_routines_statistics.pyx
@@ -59,7 +59,10 @@ cdef ndarray _ndarray_mean(ndarray self, axis, dtype, out, keepdims):
         result = cub.cub_reduction(self, cub.CUPY_CUB_SUM, axis, dtype, out,
                                    keepdims)
         if result is not None:
-            result /= (self.size / result.size)
+            if result.dtype.kind in ['f', 'c']:
+                result /= (self.size / result.size)
+            else:
+                result = result / (self.size / result.size)
             return result
     return _mean(self, axis=axis, dtype=dtype, out=out, keepdims=keepdims)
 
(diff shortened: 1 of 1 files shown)


niboshi:
Thanks!

cupy/core/_routines_statistics.pyx
```diff
@@ -59,7 +59,10 @@ cdef ndarray _ndarray_mean(ndarray self, axis, dtype, out, keepdims):
         result = cub.cub_reduction(self, cub.CUPY_CUB_SUM, axis, dtype, out,
                                    keepdims)
         if result is not None:
-            result /= (self.size / result.size)
+            if result.dtype.kind in ['f', 'c']:
+                result /= (self.size / result.size)
+            else:
+                result = result / (self.size / result.size)
```
niboshi: I think if `out` is given, the first branch should be taken.

(More precisely, `/=` does not work for integer arrays. I think we need another branch `if out is not None and result.dtype.kind in ('i', 'u')`  to use `//=`.)

cupy/core/_routines_statistics.pyx
```diff
@@ -59,7 +59,10 @@ cdef ndarray _ndarray_mean(ndarray self, axis, dtype, out, keepdims):
         result = cub.cub_reduction(self, cub.CUPY_CUB_SUM, axis, dtype, out,
                                    keepdims)
         if result is not None:
-            result /= (self.size / result.size)
+            if result.dtype.kind in ['f', 'c']:
```
niboshi: A nitpick, but a tuple is preferable (slightly faster).
```suggestion
            if result.dtype.kind in ('f', 'c'):
```

cupy/core/_routines_statistics.pyx
grlee77: Looking into [how NumPy handles it](https://github.com/numpy/numpy/blob/c4d5e12879c19c2655b5620fa1c18242b14d5141/numpy/core/_methods.py#L149-L154), there is basically just one case, but it uses `casting='unsafe'` to handle the issue.

I think this would be the equivalent here:
```Python
navg = self.size // result.size
result = cupy.true_divide(result, navg , out=result, casting='unsafe')
```

niboshi: That sounds better 👍
(I don't think you have to reassign `result` though)

niboshi:
Thanks.
Jenkins, test this please

niboshi:
👍

PR #2948, opened 2020-01-10
Update cuTENSOR version in docs

toslunar:
Follow-up #2709.  To be backported to v7 (but not to v6).


niboshi:
Jenkins, test this please

niboshi:
👍

PR #2951, opened 2020-01-13
Import `using_allocator` in `cupy.cuda`

jakirkham:
cc @emcastillo

### DIFF
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -101,6 +101,27 @@ from cupy.cuda.stream import Stream  # NOQA
 
 
 @contextlib.contextmanager
+def using_allocator(allocator=None):
+    """Sets a thread-local allocator for GPU memory inside
+       context manager
+
+    Args:
+        allocator (function): CuPy memory allocator. It must have the same
+            interface as the :func:`cupy.cuda.alloc` function, which takes the
+            buffer size as an argument and returns the device buffer of that
+            size. When ``None`` is specified, raw memory allocator will be
+            used (i.e., memory pool is disabled).
+
+    Note:
+        This wraps an internal version of this function to provide a
+        `contextmanager` as `contextmanger` decoration doesn't behave
+        well in Cython.
+    """
+    for y in memory._using_allocator(allocator):
+        yield y
+
+
+@contextlib.contextmanager
 def profile():
     """Enable CUDA profiling during with statement.
 
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -87,7 +87,6 @@ from cupy.cuda.memory import MemoryPointer  # NOQA
 from cupy.cuda.memory import MemoryPool  # NOQA
 from cupy.cuda.memory import set_allocator  # NOQA
 from cupy.cuda.memory import get_allocator  # NOQA
-from cupy.cuda.memory import using_allocator  # NOQA
 from cupy.cuda.memory import UnownedMemory  # NOQA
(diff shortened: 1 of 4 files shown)


emcastillo:
depends on #2952

niboshi:
Why does it depend on #2952?

emcastillo:
They share a commit

niboshi:
It seems they don't share a commit and I think we can merge this PR and discard the other one.

emcastillo:
`Document `using_allocator` function` this commit is duplicated, although it has a different hash
I agree with closing #2952 and merging this one instead

jakirkham:
PR ( https://github.com/cupy/cupy/pull/2952 ) could be backported to `v7`. This PR cannot be backported.

niboshi:
>`Document using_allocator function` commit is duplicated in both, although it has a different hash.

And the content is different, too.

@jakirkham 
I think it's opposite: this PR can be backported and #2952 shouldn't.
`cupy.cuda.memory.using_allocator` is a "wrong" signature and users shouldn't use it.
Fortunately this signature is not documented yet. If it became documented in v7 (by backporting #2952) and this PR would be merged (to v8), v7 and v8 would be incompatible.

This PR breaks existing code that use `cupy.cuda.memory.using_allocator`, but we can't avoid that anyway. Maybe we can say it's their responsibility because it's not documented.

jakirkham:
Ok happy to defer 🙂

emcastillo:
I just got the flow of the two prs😅
sorry for the missunderstanding

jakirkham:
No worries. Sorry for not being clearer.

emcastillo:
Might be the sphinx error related to the function being a context manager?

jakirkham:
It looks like Travis CI is using Python 3.5.6. Is that needed or can we update to a newer Python version?

jakirkham:
My guess is using `contextmanager` in Cython results in something that Sphinx is unable to parse. This issue ( https://github.com/cython/cython/issues/1750 ) and [this blogpost]( https://opendreamkit.org/2017/06/09/CythonSphinx/ ) seem relevant here. What do we think about just moving the context manager into a pure Python file?

emcastillo:
That approach should be fine.

jakirkham:
Have pushed a few changes. These rewrite `using_allocator` to use only `get_allocator` and `set_allocator` (to make it possible to move the function). Then have pulled this code into `cupy.cuda`'s `__init__.py`. This seems to fix the Sphinx error. However as I've rewritten the function, it would be good to run this on Jenkins and also get a review. Please let me know what you think 🙂

emcastillo:
Jenkins, test this please

jakirkham:
Had missed updating the tests. This seems to fix them except for one test that is checking for a `ValueError` to be raised by `set_allocator` when called within a `using_allocator` context. Will try to work out some changes to fix this.

jakirkham:
Ok, simplified things a fair bit. Expect this should work better. Could use another run on Jenkins when you have a chance @emcastillo 🙂

jakirkham:
Should add this worked for me locally as well. 😄

emcastillo:
Jenkins, test this please

jakirkham:
The Jenkins status says the build failed, but if I look at the dashboard everything is passing. Am I missing something?

Screen Shot 2020-02-17 at 7 38 23 PM

emcastillo:
Jenkins went nuts, lets try it again.
Jenkins, test this please.

jakirkham:
No worries. Thanks for the help @emcastillo! 😄 Would also be interested to get your take on this approach 😉

PR #2965, opened 2020-01-15
cupy_setup_build: Do not require libcuda at build time

andersk:
libcuda is supplied by the nvidia kernel driver.  It’s not used at build time; it’s linked dynamically at runtime (as it must be, otherwise you would have to recompile every CUDA application when you upgrade the nvidia kernel driver).

This allows building CuPy on a system or container that doesn’t have the nvidia driver, and then using it on a system that does.  This use case is important to at least NixOS, which isolates builds from the
host system.


andersk:
This may need some more thought.

PR #1398, opened 2018-06-20
Implement RawKernel

kmaehashi:
~`SimpleKernel`~ `RawKernel` is intended to expose public API to bring in raw CUDA kernel code to CuPy world.
This is one of the frequently requested feature (#1258, #1242, etc.).

Currently it is just a thin wrapper around `cupy.core.compile_with_cache` with automatic per-device memoize. We'd like to discuss the design of SimpleKernel.

After that, I should:

- [x] write documentation
- [x] add tests

PR #2966, opened 2020-01-15
Allow to pass a `kernel` argument to `RawKernel`

emcastillo:
Similar approach to #2784 
prevents to instantiate a `RawKernel` object without code, or an actual kernel obtained from the compilation backend.

Thanks to @niboshi

### DIFF
--- a/cupy/cuda/function.pyx
+++ b/cupy/cuda/function.pyx
@@ -155,7 +155,11 @@ cdef _launch(intptr_t func, Py_ssize_t grid0, int grid1, int grid2,
 
 cdef class Function:
 
-    """CUDA kernel function."""
+    """CUDA kernel function.
+
+    Class dealing with the execution of a compiled CUDA function
+    and management of runtime attributes.
+    """
 
     def __init__(self, Module module, str funcname,
                  bint enable_cooperative_groups):
--- a/cupy/cuda/function.pyx
+++ b/cupy/cuda/function.pyx
@@ -157,12 +157,14 @@ cdef class Function:
 
     """CUDA kernel function."""
 
-    def __init__(self, Module module, str funcname):
+    def __init__(self, Module module, str funcname,
+                 bint enable_cooperative_groups):
         self.module = module  # to keep module loaded
         self.ptr = driver.moduleGetFunction(module.ptr, funcname)
+        self.enable_cooperative_groups = enable_cooperative_groups
 
     def __call__(self, tuple grid, tuple block, args, size_t shared_mem=0,
-                 stream=None, enable_cooperative_groups=False):
+                 stream=None):
         grid = (grid + (1, 1))[:3]
         block = (block + (1, 1))[:3]
         s = _get_stream(stream)
@@ -170,7 +172,7 @@ cdef class Function:
             self.ptr,
             max(1, grid[0]), max(1, grid[1]), max(1, grid[2]),
             max(1, block[0]), max(1, block[1]), max(1, block[2]),
-            args, shared_mem, s, enable_cooperative_groups)
+            args, shared_mem, s, self.enable_cooperative_groups)
 
     cpdef linear_launch(self, size_t size, args, size_t shared_mem=0,
                         size_t block_max_size=128, stream=None):
@@ -305,8 +307,9 @@ cdef class Module:
 
     """CUDA kernel module."""
 
-    def __init__(self):
+    def __init__(self, enable_cooperative_groups):
         self.ptr = 0
+        self.enable_cooperative_groups = enable_cooperative_groups
 
     def __dealloc__(self):
         if self.ptr:
@@ -331,7 +334,7 @@ cdef class Module:
     cpdef get_function(self, name):
         if isinstance(name, bytes):
             name = name.decode()
-        return Function(self, name)
+        return Function(self, name, self.enable_cooperative_groups)
 
     cpdef get_texref(self, name):
         if isinstance(name, bytes):
--- a/cupy/cuda/function.pyx
+++ b/cupy/cuda/function.pyx
@@ -182,6 +182,124 @@ cdef class Function:
         _launch(self.ptr,
                 gridx, 1, 1, blockx, 1, 1, args, shared_mem, s)
 
+    @property
+    def attributes(self):
+        """Returns a dictionary containing runtime kernel attributes. This is
+        a read-only property; to overwrite the attributes, use
+
+        .. code-block:: python
+
+            kernel = RawKernel(...)  # arguments omitted
+            kernel.max_dynamic_shared_size_bytes = ...
+            kernel.preferred_shared_memory_carveout = ...
+
+        Note that the two attributes shown in the above example are the only
+        two currently settable in CUDA.
+
+        Any attribute not existing in the present CUDA toolkit version will
+        have the value -1.
+
+        Returns:
+            dict: A dictionary containing the kernel's attributes.
+        """
+        cdef dict attrs = {}
+        cdef list keys = ['max_threads_per_block', 'shared_size_bytes',
+                          'const_size_bytes', 'local_size_bytes',
+                          'num_regs', 'ptx_version', 'binary_version',
+                          'cache_mode_ca', 'max_dynamic_shared_size_bytes',
+                          'preferred_shared_memory_carveout']
+        for attr in keys:
+            attrs[attr] = getattr(self, attr)
+        return attrs
+
+    @property
+    def max_threads_per_block(self):
+        """The maximum number of threads per block that can successfully
+        launch the function on the device.
+        """
+        attr = driver.CU_FUNC_ATTRIBUTE_MAX_THREADS_PER_BLOCK
+        return driver.funcGetAttribute(attr, self.ptr)
+
+    @property
+    def shared_size_bytes(self):
+        """The size in bytes of the statically-allocated shared memory
+        used by the function. This is separate from any dynamically-allocated
+        shared memory, which must be specified when the function is called.
+        """
+        attr = driver.CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES
+        return driver.funcGetAttribute(attr, self.ptr)
+
+    @property
+    def const_size_bytes(self):
+        """The size in bytes of constant memory used by the function."""
+        attr = driver.CU_FUNC_ATTRIBUTE_CONST_SIZE_BYTES
+        return driver.funcGetAttribute(attr, self.ptr)
+
+    @property
+    def local_size_bytes(self):
+        """The size in bytes of local memory used by the function."""
+        attr = driver.CU_FUNC_ATTRIBUTE_LOCAL_SIZE_BYTES
+        return driver.funcGetAttribute(attr, self.ptr)
+
+    @property
+    def num_regs(self):
+        """The number of registers used by the function."""
+        attr = driver.CU_FUNC_ATTRIBUTE_NUM_REGS
+        return driver.funcGetAttribute(attr, self.ptr)
+
+    @property
+    def ptx_version(self):
+        """The PTX virtual architecture version that was used during
+        compilation, in the format: 10*major + minor.
+        """
+        attr = driver.CU_FUNC_ATTRIBUTE_PTX_VERSION
+        return driver.funcGetAttribute(attr, self.ptr)
+
+    @property
+    def binary_version(self):
+        """The binary architecture version that was used during compilation,
+        in the format: 10*major + minor.
+        """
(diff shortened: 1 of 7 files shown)


leofang:
Looks nice. I have two questions:
1. > or an actual kernel obtained from the compilation backend.

what does this mean?

2. I thought `cupy.cuda.function.Function` is a private API (which was presumably why we had `RawKernel` on top?). Is it necessary to document the `kernel` argument?

emcastillo:
You are right, should we leave the arguments docstring untouched and describe this in a separate comment?

niboshi:
About 2., I thought it's a public class (there's a shorter alias `cupy.cuda.Function`), but I'm not so sure. It's not documented.

emcastillo:
@leofang I changed the description of the `kernel` arg, since it seems to be public let me know what you think

niboshi:
With `kernel` given, all other arguments except `enable_cooperative_groups` are meaningless as well.
I now feel it's not optimal in tems of modularity to add `kernel` argument to `RawKernel`.

emcastillo:
I see the point, an approach to tackle this could be to have some factory method for compiling the raw kernel itself while the actual kernel class just takes care of execution parameters.
Right now  `RawKernel` is in charge of calling the compiler for the code and the actual execution. These two responsibilities could be separated.

niboshi:
>I see the point, an approach to tackle this could be to have some factory method for compiling the raw kernel itself while the actual kernel class just takes care of execution parameters.

I'm afraid that's not possible without breaking compatibility because `RawKernel.__init__` already has arguments for compilation.
(IMO `__init__` exactly defines the role of a class in general. In this point of view, `RawKernel` has been defined to exist for compilation (and additionally, for execution)).
 
I feel it's worth exploring the possibility to separate the roles.

niboshi:
Maybe `cupy.cuda.Function` could be the class for execution and `RawModule.get_function` could simply return `Function`?
`RawKernel`'s properties such as `max_threads_per_block` could be implemented in `Function` as well.

emcastillo:
> Maybe cupy.cuda.Function could be the class for execution and RawModule.get_function could simply return Function?
RawKernel's properties such as max_threads_per_block could be implemented in Function as well.

This sounds a good behavior that maybe `ElementwiseKernel` and `ReductionKernel` could benefit from as they use the `cuda.Function` to hold the actual kernels.

leofang:
> an approach to tackle this could be to have some factory method for compiling the raw kernel itself while the actual kernel class just takes care of execution parameters.

I think `_get_raw_kernel()` was meant to be the factory method?

> Maybe `cupy.cuda.Function` could be the class for execution and `RawModule.get_function` could simply return `Function`?
> `RawKernel`'s properties such as `max_threads_per_block` could be implemented in `Function` as well.

This is interesting. As long as the following semantics is maintained:
```python
mod = cp.RawModule(...)
ker = mod.get_function('the_kernel')
ker. max_dynamic_shared_size_bytes = 1024  # set size
ker(a)  # run the kernel with array a
```
I think the backward compatibility of `RawModule` will be intact. Users won't notice `ker` becomes a `cuda.Function` instead of a `RawKernel`.

But, one small drawback I see for separating the roles is that `RawKernel` still needs to have all the properties it currently has (which will wrap on top of `cuda.Function`'s counterparts). In addition, `RawKernel.attributes` should continue to return a dict, and we don't need `cuda.Function.attributes`.

niboshi:
>But, one small drawback I see for separating the roles is that `RawKernel` still needs to have all the properties it currently has (which will wrap on top of `cuda.Function`'s counterparts).

I feel it's natural to keep such features in`RawKernel` (as a convenient addition to the "main" role).

>In addition, `RawKernel.attributes` should continue to return a dict, and we don't need `cuda.Function.attributes`.

I don't see why we don't need `Function.attributes`.
If users could be using it via `RawKernel`, should `Function` also need it for their existing code to work?

e.g.
```py
mod = cp.RawModule(...)
ker = mod.get_function('the_kernel')
ker.attributes  # Would need `Function.attributes`
```

leofang:
> If users could be using it via `RawKernel`, should `Function` also need it for their existing code to work?

@niboshi I guess you're right. I was thinking the current design, sorry. 

But I still think there would be some code repetition for the properties. What if we implement a `@classmethod` in `RawKernel` that takes a `cuda.Function` instance and warps it as a `RawKernel`? This would not only circumvent the problem of setting additional argument in the constructor, but also avoid both the splitting effort and code repetition.

niboshi:
>But I still think there would be some code repetition for the kernel properties. What if we implement a `@classmethod` in `RawKernel` that takes a `cuda.Function` instance and returns a `RawKernel`? This would not only circumvent the problem of adding additional argument to the constructor, but also avoid both the splitting effort and code repetition.

The classmethod would have to call `__init__` anyway, with all the arguments ignored. This is not clean.

Separation of roles is the most important point here IMO.
I'd rather move all the properties to `Function` to avoid repetition and keep responsibilities clear (but that would break compatibility of course).

leofang:
> The classmethod would have to call `__init__` anyway, with all the arguments ignored. This is not clean.

Putting the goal to accomplish (initiating a `RawKernel` with a `cuda.Function` instance) aside, which I am not sure why it is necessary to be a public API, this PR basically asks for an alternative constructor. From this point of view, refactoring the current implementation is just an overkill, which I think would make the abstraction less clean than it currently is. 

For alternative/overloaded constructors (or factory functions), Cython provides an example for how to do this: https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#instantiation-from-existing-c-c-pointers, which I followed in `cupy.cuda.nccl.NcclCommunicator`. There is no need to call `__init__`, just `__new__`: https://github.com/cupy/cupy/blob/a9e5ff3cb0e25891e0588b7206ec04cf9a78dac0/cupy/cuda/nccl.pyx#L306-L311 One minor thing to note, though, is that `@classmethod` in Cython, if I recall correctly, was not implemented correctly and led to compile time errors, which was why I used `@staticmethod` instead. 

> Separation of roles is the most important point here IMO.
> I'd rather move all the properties to `Function` to avoid repetition and keep responsibilities clear (but that would break compatibility of course).

"Keeping responsibilities clear" is somehow ambiguous to me. There can be many things that would be natural to tie together in a class. I don't see why it is not the case here.

niboshi:
Clearly `RawKernel` was originally intended to be a facility to compile a given CUDA source (#1398).
It was not introduced to be a `Function` wrapper.
The interface of `RawKernel.__init__` clearly reflects that intention.

Introducing a shortcut constructor that circumvent the primary role of the class, would make the role ambiguous and diverged from the original intention.

niboshi:
If we were to redefine the role of `RawKernel` to be "a `Function` wrapper", there would be some inconsistencies.
* Interface of `RawKernel.__init__` would be erratic.
* The word `Raw` would not make sense. I guess it refers to a "*raw* CUDA source".
* How would it differ from `Function`?

emcastillo:
IMO, I would keep `RawKernel` as a class whose main responsibility is to compile the kernel. Then I would just move all the CUDA attributes accessing code to the `Function` class. (Here it is possible to do a nocompat change, so the user would access the attributes through the `raw_kernel.function.num_regs` instead of the previous 
`raw_kernel.num_regs`. However, it is possible to keep the accessors in `RawKernel` as wrappers over `Function`

`RawModule.get_function` should now return `Function` instead of `RawKernels` and with the above change, compatibility is retained as the `Function` object is callable and all the properties can be retrieved.

emcastillo:
I just pushed a preliminary version of my proposal, needs work (specifically with the cooperative groups as there is no easy way to propagate it directly to the function since it only gets instantiated in module, which doesn't feel natural to have such a parameter)

niboshi:
@emcastillo Can you continue toward that direction?

emcastillo:
I just submit a complete implementation.

Notes:
I deprecated the setter/getter methods in `RawKernel` and moved them to `Function`.
However my deprecation warnings or comments might not be compliant?

`enable_cooperative_groups` is now a parameter of `Module` as it is needed inside `Function`, but `Function` is never instantiated directly from `RawKernel` or `RawModule`. An alternative to this is to explicitly set it to `False` in the `Function.__init__` method and change the value from `RawKernel` or `RawModule` after getting the `Function` object.

leofang:
> Notes:
I deprecated the setter/getter methods in RawKernel and moved them to Function.
However my deprecation warnings or comments might not be compliant?

If I may suggest, I would not deprecate setter/getter in `RawKernel`, and rather simply set to/get from the corresponding attribute of `Function`. Making them a thin wrapper, in other words. The reason is two-fold: 
1. we let `Function` float around in users' code, but we don't want them to notice this subtlety and would like them to think they are accessing `RawKernel`'s properties. 
2. `Function` is not documented and our life would be much simpler if it remains so, otherwise we will have to explain this in https://docs-cupy.chainer.org/en/stable/tutorial/kernel.html#raw-kernels.

niboshi:
> 1. we let `Function` float around in users' code, but we don't want them to notice this subtlety and would like them to think they are accessing `RawKernel`'s properties.

The decision above (https://github.com/cupy/cupy/pull/2966#issuecomment-576513413) means we make `Function` public and documented. I think there's no problem letting users notice the `Function` class. 

> 2. `Function` is not documented and our life would be much simpler if it remains so, otherwise we will have to explain this in https://docs-cupy.chainer.org/en/stable/tutorial/kernel.html#raw-kernels, where the kernel attributes are documented.

Well, we will instead have to move the mention of the properties to `Function`'s documentation, if they're going to be deprecated in `RawKernel`.

emcastillo:
I am not sure up to what extent this will affect user codes.

I am fine with both. I don't think that accessing execution time properties via function is a big deal.
But as you said we have to document it clearly.

emcastillo:
Added a first draft for the docs.
As the `cupy.cuda.Function` class is not intended to be directly instantiated by users, it is a bit tricky to document.

Another concern, is that do we only expose `Function` for `RawKernel`? `ElementwiseKernel` uses `Function` too, but it does not save it in an attribute. I am not sure if we should unify this in all the kernel classes so there is some consistency.

niboshi:
@emcastillo 

>Another concern, is that do we only expose `Function` for `RawKernel`? `ElementwiseKernel` uses `Function` too, but it does not save it in an attribute. I am not sure if we should unify this in all the kernel classes so there is some consistency.

It's not possible for `ElementwiseKernel` because `Function` can only be generated when `__call__` is called, which is not the case in `RawKernel`.

cupy/cuda/function.pyx
```diff
@@ -155,7 +155,12 @@ cdef _launch(intptr_t func, Py_ssize_t grid0, int grid1, int grid2,
 
 cdef class Function:
 
-    """CUDA kernel function."""
+    """Class in charge of the execution of a compiled CUDA kernel function.
+
+    This class is used to access the runtime attributes of kernels compiled
+    via the :class:`~RawKernel`. It is mean to be accessed through
+    :attr:`~RawKernel.function`.
```
niboshi: This is not the primary goal of this class. I think we just don't need to mention this.

emcastillo: ok!

emcastillo:
> It's not possible for ElementwiseKernel because Function can only be generated when __call__ is called, which is not the case in RawKernel.

Exactly, but it feels strange that for a `RawKernel` you can access those attributes and for an `ElementwiseKernel` you can't.

cupy/cuda/function.pyx
```diff
@@ -155,7 +155,8 @@ cdef _launch(intptr_t func, Py_ssize_t grid0, int grid1, int grid2,
 
 cdef class Function:
 
-    """CUDA kernel function."""
+    """Class in charge of the execution of a compiled CUDA kernel function.
+    """
```
niboshi: Do you have to change the summary line?
Execution is not the only function of this class.
As we're adding kernel attributes, you may as well use this class to simply retrieve those attributes.

I think the previous expression represents exactly what this class is.

niboshi:
>> It's not possible for ElementwiseKernel because Function can only be generated when __call__ is called, which is not the case in RawKernel.
>
>Exactly, but it feels strange that for a `RawKernel` you can access those attributes and for an `ElementwiseKernel` you can't. Anyway, it is not a problem of this PR.

I don't think it's strange, because it's the fundamental difference of those classes.
`RawKernel` can generate `Function` statically, and `ElementwiseKernel` generates it dynamically.
Simply that's reflected to the API.

asi1024:
Could you resolve conflicts?

emcastillo:
After the refactoring, the issue that this PR was addressing is gone, so we can close it.

PR #355, opened 2017-08-02
Avoid importing NumPy docstring

niboshi:
**(I have changed the main objective of this PR)**

Curerntly NumPy alias functions use NumPy docstring in documentation.
That incurs warnings because intersphinx environment is different between CuPy and NumPy.
This PR avoids this problem. (Rel. #442)

**(original motivation)**
[source] links for NumPy alias functions are not correctly shown.
#354 fixes the problem by hiding the links, but it might look strange for some users.
This PR is to enable the links anyway for such functions.

~(Currently only applied to `cupy.isscalar`. If this implementation is acceptable, I will apply to all such alias functions)~

PR #2974, opened 2020-01-17
Rename arg of `isscalar`

toslunar:


toslunar:
It's possible that
```python
from numpy import isscalar  # NOQA
```
is sufficient.

takagi:
pfnCI, test this please.

takagi:
As the current `cupy.isscalar` definition was introduced at #355 replacing with

```Python
from numpy import isscalar  # NOQA
```

for some reason, I suppose leaving the definition would be fine.

takagi:
LGTM! Waiting for CI finish.

niboshi:
This is nocompat, right?

takagi:
Ah... yes.

PR #2980, opened 2020-01-20
Remove dependency to six

niboshi:


cupy/cuda/memory.pyx
```diff
@@ -657,17 +657,15 @@ cdef class PooledMemory(BaseMemory):
                 device_id = self.device_id
                 pmem_id = id(self)
 
-                # avoid six for performance
-                hooks_values = hooks.values()
-                for hook in hooks_values:
+                for hook in hooks.values:
```
toslunar:
```suggestion
                for hook in hooks.values():
```

niboshi: Thanks. Force-pushed.

toslunar:
Jenkins, test this please.

leofang:
This is great! Definitely the right thing to do to reduce code complexity now that we're on PY3 only. Thanks @niboshi.

PR #2938, opened 2020-01-09
Drop NumPy<1.15 support

niboshi:
Following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

Fixes #2900.

PR #2983, opened 2020-01-20
Added `out` parameter to `cupy.concatenate` and `cupy.stack`

emcastillo:
Numpy supports `out` for both `stack` and `concatenate`.
This PR matches the behavior

https://docs.scipy.org/doc/numpy/reference/generated/numpy.stack.html
https://docs.scipy.org/doc/numpy/reference/generated/numpy.concatenate.html

### DIFF
--- a/cupy/core/_routines_manipulation.pyx
+++ b/cupy/core/_routines_manipulation.pyx
@@ -531,7 +531,7 @@ cpdef ndarray _repeat(ndarray a, repeats, axis=None):
     return ret
 
 
-cpdef ndarray concatenate_method(tup, int axis):
+cpdef ndarray concatenate_method(tup, int axis, ndarray out=None):
     cdef int ndim, a_ndim
     cdef int i
     cdef ndarray a
@@ -577,15 +577,32 @@ cpdef ndarray concatenate_method(tup, int axis):
     if not have_same_types:
         dtype = functools.reduce(numpy.promote_types,
                                  set([a.dtype for a in arrays]))
-    return _concatenate(arrays, axis, tuple(shape), dtype)
+
+    shape_t = tuple(shape)
+    if out is None:
+        out = ndarray(shape_t, dtype=dtype)
+    else:
+        if len(out.shape) != len(shape_t):
+            raise ValueError('Output array has wrong dimensionality')
+        if out.shape != shape_t:
+            raise ValueError('Output array is the wrong shape')
+        if out.dtype.kind != dtype.kind:
+            raise TypeError('Cannot cast scalar from dtype(\'{}\')'
+                            ' to dtype(\'{}\') according to the'
+                            ' rule \'same_kind\''.format(dtype, out.dtype))
+
+    return _concatenate(arrays, axis, shape_t, out)
 
 
-cpdef ndarray _concatenate(list arrays, Py_ssize_t axis, tuple shape, dtype):
-    cdef ndarray a, ret
+cpdef ndarray _concatenate(
+        list arrays, Py_ssize_t axis, tuple shape, ndarray out):
+    cdef ndarray a
     cdef Py_ssize_t i, aw, itemsize, axis_size
     cdef bint all_same_type, same_shape_and_contiguous
     # If arrays are large, Issuing each copy method is efficient.
     cdef Py_ssize_t threshold_size = 2 * 1024 * 1024
+
+    dtype = out.dtype
 
     if len(arrays) > 8:
         all_same_type = True
@@ -604,17 +621,16 @@ cpdef ndarray _concatenate(list arrays, Py_ssize_t axis, tuple shape, dtype):
 
         if all_same_type and total_bytes < threshold_size * len(arrays):
             return _concatenate_single_kernel(
-                arrays, axis, shape, dtype, same_shape_and_contiguous)
+                arrays, axis, shape, dtype, same_shape_and_contiguous, out)
 
-    ret = ndarray(shape, dtype=dtype)
     i = 0
     slice_list = [slice(None)] * len(shape)
     for a in arrays:
         aw = a._shape[axis]
         slice_list[axis] = slice(i, i + aw)
-        elementwise_copy(a, _indexing._simple_getitem(ret, slice_list))
+        elementwise_copy(a, _indexing._simple_getitem(out, slice_list))
         i += aw
-    return ret
+    return out
 
 
 cpdef Py_ssize_t size(ndarray a, axis=None) except? -1:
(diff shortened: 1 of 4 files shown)


cupy/core/_routines_manipulation.pyx
```diff
@@ -693,8 +705,8 @@ cdef _normalize_axis_tuple(axis, Py_ssize_t ndim,
 
 cdef ndarray _concatenate_single_kernel(
         list arrays, Py_ssize_t axis, tuple shape, dtype,
-        bint same_shape_and_contiguous):
-    cdef ndarray a, x, ret
+        bint same_shape_and_contiguous, out=None):
```
niboshi: As `_concatenate_single_kernel` is an internal function and called from only one place, I don't think `out` needs a default value.

cupy/core/_routines_manipulation.pyx
```diff
@@ -559,18 +559,31 @@ cpdef ndarray concatenate_method(tup, int axis):
...
         dtype = functools.reduce(numpy.promote_types,
                                  set([a.dtype for a in arrays]))
-    return _concatenate(arrays, axis, tuple(shape), dtype)
+    return _concatenate(arrays, axis, tuple(shape), dtype, out)
 
 
-cpdef ndarray _concatenate(list arrays, Py_ssize_t axis, tuple shape, dtype):
-    cdef ndarray a, ret
+cpdef ndarray _concatenate(
+        list arrays, Py_ssize_t axis, tuple shape, dtype, ndarray out=None):
```
niboshi: Ditto. `out` doesn't need a default value?

cupy/core/_routines_manipulation.pyx
```diff
@@ -706,11 +718,12 @@ cdef ndarray _concatenate_single_kernel(
         ptrs[i] = a.data.ptr
     x = core.array(ptrs)
 
-    ret = core.ndarray(shape, dtype=dtype)
+    if out is None:
+        out = core.ndarray(shape, dtype=dtype)
```
niboshi: I think `out` always has a value.
How about `assert out is not None` instead (at the beginning of the function)?

emcastillo:
Force pushed, PTAL.😊
Thanks!

cupy/core/_routines_manipulation.pyx
```diff
@@ -559,18 +559,31 @@ cpdef ndarray concatenate_method(tup, int axis):
...
+        if len(out.shape) != len(shape):
+            raise ValueError('Output array has wrong dimensionality')
+        if out.shape != shape:
+            raise ValueError('Output array is the wrong shape')
+        if out.dtype.kind != dtype.kind:
+            raise TypeError('Cannot cast scalar from dtype(\'{}\')'
+                            ' to dtype(\'{}\') according to the'
+                            ' rule \'same_kind\''.format(dtype, out.dtype))
+
+    return _concatenate(arrays, axis, tuple(shape), dtype, out)
```
niboshi: I think `tuple(shape)` should be done earlier, maybe just after the `for` loop above.

(I don't know whether/how `out.shape != shape` works for a `vector`. I guess it's converted to a `tuple` there implicitly.)

cupy/core/_routines_manipulation.pyx
```diff
@@ -559,18 +559,31 @@ cpdef ndarray concatenate_method(tup, int axis):
...
+        list arrays, Py_ssize_t axis, tuple shape, dtype, ndarray out):
+    cdef ndarray a
     cdef Py_ssize_t i, aw, itemsize, axis_size
     cdef bint all_same_type, same_shape_and_contiguous
     # If arrays are large, Issuing each copy method is efficient.
     cdef Py_ssize_t threshold_size = 2 * 1024 * 1024
 
     if len(arrays) > 8:
-        all_same_type = True
+        all_same_type = out.dtype == arrays[0].dtype
```
niboshi: Don't you need to check all the arrays, instead of just the first one?

emcastillo: i compare the out with the first one to verify the out and the first array share the same type, otherwise that flag will be set to false and the slow path is executed.
The comparison with the rest of arrays is done in the for loop below.
I can move the comparison of the out array type after this loop, maybe thats more readable?

niboshi: Ah, I didn't check the code below.

I think what to be done is to remove `dtype` argument from `_concatenate()`, use `dtype = out.dtype` and keep this line intact?

emcastillo: I like that approach!

emcastillo: I like that approach!

emcastillo:
Done! PTAL🤓

tests/cupy_tests/manipulation_tests/test_join.py
```diff
@@ -138,6 +138,49 @@ def test_concatenate_wrong_shape(self):
...
+    @testing.numpy_cupy_array_equal()
+    def test_concatenate_out_same_kind(self, xp):
+        a = testing.shaped_arange((3, 4), xp, xp.float64)
+        b = testing.shaped_reverse_arange((3, 4), xp, xp.float64)
+        c = testing.shaped_arange((3, 4), xp, xp.float64)
+        out = xp.zeros((3, 12), dtype=xp.float32)
+        xp.concatenate((a, b, c), axis=1, out=out)
+        return out
+
+    @testing.numpy_cupy_raises()
```
niboshi: Let's spell out the exception type.

tests/cupy_tests/manipulation_tests/test_join.py
```diff
@@ -138,6 +138,49 @@ def test_concatenate_wrong_shape(self):
...
+    def test_concatenate_out_invalid_shape(self, xp):
+        a = testing.shaped_arange((3, 4), xp, xp.float64)
+        b = testing.shaped_reverse_arange((3, 4), xp, xp.float64)
+        c = testing.shaped_arange((3, 4), xp, xp.float64)
+        out = xp.zeros((4, 10), dtype=xp.float64)
+        xp.concatenate((a, b, c), axis=1, out=out)
+
+    @testing.numpy_cupy_raises()
```
niboshi: Ditto.

tests/cupy_tests/manipulation_tests/test_join.py
```diff
@@ -138,6 +138,49 @@ def test_concatenate_wrong_shape(self):
...
+    def test_concatenate_out_invalid_shape_2(self, xp):
+        a = testing.shaped_arange((3, 4), xp, xp.float64)
+        b = testing.shaped_reverse_arange((3, 4), xp, xp.float64)
+        c = testing.shaped_arange((3, 4), xp, xp.float64)
+        out = xp.zeros((2, 2, 10), dtype=xp.float64)
+        xp.concatenate((a, b, c), axis=1, out=out)
+
+    @testing.numpy_cupy_raises()
```
niboshi: Ditto.

niboshi:
Let's write the same test for `stack` as well.

emcastillo:
as `stack` just calls `concatenate` I thought such a test is redundant, but I will add it.

niboshi:
That's about implementation, but we need to test the behavior.

emcastillo:
Added the tests and the missing `accept_error` in other decorators

niboshi:
LGTM
Jenkins, test this please

emcastillo:
`module 'numpy' has no attribute 'stack'`
Older numpy versions 😂

niboshi:
Maybe wait for #2938?
(Everything we're adding in this PR are supported as of NumPy 1.15).

emcastillo:
Exactly, I will wait then :)

niboshi:
https://github.com/chainer/chainer-test/pull/553 was merged.
Jenkins, test this please

emcastillo:
```
2 sha256:611578fa1b763fb9f2ed1c630c56c7f55a3be9eafbda32c679acbdbede754a41
12:20:54 + pip install --user -e cupy/
12:20:54 Obtaining file:///work/cupy
12:20:59 Requirement already satisfied: numpy>=1.9.0 in /usr/local/lib/python3.5/dist-packages (from cupy==7.0.0) (1.9.3)
```
😱

emcastillo:
Screwed the branch when doing other things, restored

niboshi:
Jenkins, test this please

niboshi:
https://github.com/chainer/chainer-test/pull/556 was merged.

Jenkins, test this please

niboshi:
Jenkins, test this please

emcastillo:
Still failing due to old numpy

emcastillo:
This should work now.
rebased, PTAL

niboshi:
Jenkins, test this please

emcastillo:
Error was due to a missing decorator, fixed and tested it locally

niboshi:
Jenkins, test this please

niboshi:
:+1:

PR #2982, opened 2020-01-20
Add `cupy.conjugate` and make `cupy.conj` its alias

toslunar:
Fix #2967.

PR #2987, opened 2020-01-21
Add decorator to parameterize xp

toslunar:
From the discussion in #2871.


niboshi:
How about `for_numpy_cupy`? I think "run" is not suitable for a test.

toslunar:
I agree that `run` does not suit here.  Thanks.

toslunar:
This PR is blocking #2982.  I made the decorator private in order to discuss its name later.

niboshi:
@toslunar #2982 could proceed without depending on this PR.

niboshi:
For now, we decided to use `for` loop for this parameterization. #3099

Issue #2991, opened 2020-01-21
Support NumPy 1.18


toslunar:
- non-compatible changes
  - [x] #2852 [`np.can_cast(np.uint64, np.timedelta64, casting='safe')` is now `False`](https://numpy.org/devdocs/release/1.18.0-notes.html#np-can-cast-np-uint64-np-timedelta64-casting-safe-is-now-false)
  - [x] #2883 error message in `numpy.testing`
  - [x] #2974 `isscalar(num)` -> `isscalar(element)`
  - [x] #2975 `step` returned by `linspace(..., retstep=True)`
- [expired deprecations](https://numpy.org/devdocs/release/1.18.0-notes.html#expired-deprecations)
  - [x] #2992 "The deprecation of `expand_dims` out-of-range axes in 1.13.0 has expired. ([gh-14051](https://github.com/numpy/numpy/pull/14051))"
- [ ] #3005 cupy documentation

PR #2992, opened 2020-01-21
Allow multiple axes in `expand_dims`

toslunar:
Follow numpy/numpy#14051.

Part of #2991.


cupy/core/__init__.py
```diff
@@ -13,6 +13,7 @@
 from cupy.core._routines_manipulation import broadcast  # NOQA
 from cupy.core._routines_manipulation import broadcast_to  # NOQA
 from cupy.core._routines_manipulation import concatenate_method  # NOQA
+from cupy.core._routines_manipulation import _expand_dims  # NOQA
```
niboshi: I think this alias is not necessary.
You can directly `from cupy.core import _routines_manipulation as _manipulation`.

niboshi:
LGTM (just a quick look)

toslunar:
I'll add a test for `axis=()`.

toslunar:
Done.

emcastillo:
Jenkins, test this please

PR #2996, opened 2020-01-23
Fix GMM example for matplotlib 3

toslunar:
On the GMM (Gaussian Mixture Model) example, fix the following.
- Use `scipy.stats` to compute bivariate normal.  `matplotlab.mlab.bivariate_normal` is deprecated since matplotlib 2.2 and is removed since 3.0
- Surpress the warning message:
  > 'c' argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with 'x' & 'y'.  Please use a 2-D array with a single row if you really want to specify the same RGB or RGBA value for all points.

```
% python gmm.py --output-image=tmp.png
Running CPU...
train_accuracy : 92.143100
test_accuracy : 92.000000
 CPU :  1.087882 sec
Running GPU...
train_accuracy : 92.030900
test_accuracy : 92.000000
 GPU :  0.651976 sec
% imgcat tmp.png
```

### DIFF
--- a/examples/gmm/gmm.py
+++ b/examples/gmm/gmm.py
@@ -97,7 +97,7 @@ def draw(X, pred, means, covariances, output):
         labels = X[pred == i]
         if xp is cupy:
             labels = labels.get()
-        plt.scatter(labels[:, 0], labels[:, 1], c=np.random.rand(3))
+        plt.scatter(labels[:, 0], labels[:, 1], c=np.random.rand(1, 3))
     if xp is cupy:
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, test this please.

toslunar:
pfnCI, test this please.

kmaehashi:
I locally confirmed that the example now works with matplotlib 3.1.0.

PR #2787, opened 2019-12-09
Add nogil to CUB

y1r:
This PR adds `nogil` flag to CUB C interface.

PR #2997, opened 2020-01-24
[backport] Fix an issue using non-existing attribute in cub.pyx

leofang:
Backport of #2985. 

While the problematic commit f7320d2 does not show up in the `v7` branch, for some reason (likely due to a backport?) the bug still appears in `v7`. The difference between this PR and #2985 is that `nogil` is not present in `v7`, so we don't need to explicitly cast `op` to `int` in the CUB calls.

cc: @emcastillo @anaruse

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -96,7 +96,7 @@ cpdef Py_ssize_t _preprocess_array(ndarray arr, tuple reduce_axis,
     return contiguous_size
 
 
-def device_reduce(ndarray x, int op, tuple out_axis, out=None,
+def device_reduce(ndarray x, op, tuple out_axis, out=None,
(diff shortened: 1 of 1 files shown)


leofang:
Please forgive and allow me to whine a bit here: We're gonna chase after CUB bugs all the time if it's not tested in CI...😢

emcastillo:
We are working on the CUB merge so it will be solved eventually. But the required CI changes and other tasks are holding us back for a while.

emcastillo:
Lets backport #2787 first and then rebase

leofang:
Sounds good, @emcastillo. In that case, should I just close this PR and let you backport #2985 directly?

emcastillo:
Maybe that would be better. Thanks!

PR #2752, opened 2019-12-04
[v6] Fix axes handling in `_fftn`

toslunar:
From #2355, this PR backports the fixes in `cupy/fft/fft.py`.

The n-dim FFT implementation `_fftn` should correctly handle
- `axes=None` regardless of whether `shape` is specified, and
- unsorted `axes`

PR #2355, opened 2019-08-01
Add scipy.fft to cupyx

peterbell10:
SciPy version 1.4 will have a new submodule [`scipy.fft`] with an interface very similar to `numpy.fft`. This adds a subset of that interface to `cupyx.scipy` and also adds [`uarray`]-based backend support allowing cupy FFTs to be called through the `scipy.fft` interface e.g.
```python
import cupy
import cupyx.scipy.fft as cu_fft
import scipy.fft

x = cupy.arange(10)
scipy.fft.fft(x)  # ValueError: object __array__ method not producing an array

with scipy.fft.set_backend(cu_fft):
    scipy.fft.fft(x)  # Calls into cupyx, returning a cupy array
```

Tests are currently failing but at least some of the failures seem to be bugs in `cupy.fft`. I need to investigate further.

[`scipy.fft`]: https://scipy.github.io/devdocs/fft.html
[`uarray`]: https://uarray.readthedocs.io/en/latest/libauthor_docs.html

PR #2998, opened 2020-01-26
Add a keyword-only `plan` argument to `cupyx.scipy.fft.*`

### LINKED ISSUE #2743: Mark explicit `plan` in `cupyx.scipy.fftpack.*` a stable API?
The explicit `plan` argument in `cupyx.scipy.fftpack.*` and the `get_fft_plan()` function there were added in v6. I have used it extensively in production since then, and I also heard positive feedbacks (via private communication). The plans (`Plan1d` or `PlanNd`) returned by the `get_fft_plan()` handle can be reused either explicitly as an argument, or implicitly as a context manager. In other words, the plans can be used
- implicitly in `cupy.fft.fft.*` (and later `cupyx.scipy.fft.*`)
- explicitly in `cupyx.scipy.fftpack.*`

Given where we are now, and that the `cupyx.scipy.fft` API will be added soon (#2355) which do not take a `plan` as argument, I wonder if this nice feature can be removed from the "experimental" status and become a stable API. Thanks.

leofang:
Closes #2743. Following scipy/scipy#11408 (to appear in SciPy v1.5.0). ref: scipy/scipy#10302.

This PR includes three changes:
1. A bulk of work for adding the `plan` arg. The added tests largely follow `fftpack_tests/test_fft.py`, with the style modified to match that in `fft_tests/test_fft.py`
2. Fix a bug in `cupyx.scipy.fftpack.get_fft_plan()` following #2752 (1c6174d, **should be backported**)
3. Fix missing `axes` arguments in `TestRfft2` (4ef81b1, **should be backported**)

No backward compatibility is broken in this PR.

TODO:
1. ~~Update the doc~~ (**EDIT:** done)
2. Discussion?


leofang:
After this PR, I would like to make `cupyx.scipy.fft` tested against SciPy 1.4.0+ instead of NumPy unless it's absent.

asi1024:
Thank you for the PR! I agree to add `plan` keyword-only argument, but I want to support it as an experimental feature until SciPy v1.5.0 is released. Could you raise a warning if non-`None` `plan` is given?

leofang:
That’s a good point, @asi1024. I would add a version check in `__ua_function__` and raise warning if <1.5.0. This is because when using as is, this module should behave similar to `cupyx.scipy.fftpack` and it’s ok to have `plan != None`, but when using as a scipy backend, it would matter if scipy does not have the plan argument added. This check-and-warn will still be needed even when 1.5.0 is out. (And an error would follow after the warning if `plan` is not accepted as a legit kwarg.) 

What do you think?

leofang:
> I would add a version check in `__ua_function__` and raise warning if <1.5.0. 

This is added in 3227ec2. I would test this warning in the next PR mentioned earlier, as currently `set_backend()` is not tested.

asi1024:
Even if `scipy.fft` supported plan, it will not accept CuPy's plan. So, I think that we don't have to worried about that.

leofang:
Oh, IIUC we better do. Consider the (slightly tweaked) example from the original PR (#2355): 
```python
import cupy
import cupyx.scipy.fft as cu_fft
import scipy.fft

x = cupy.random.random(10)
with scipy.fft.set_backend(cu_fft):
    y = scipy.fft.fft(x)  # Calls into cupyx, returning a cupy array
```
Under the hood the [uarray dispatch](https://uarray.org/en/latest/libauthor_docs.html) kicks in so that arguments passed to `scipy.fft.fft()` are delegated to `cupyx.scipy.fft.fft()`. 

Now, with `plan` added:
```python
import cupy
import cupyx.scipy.fft as cu_fft
import scipy.fft

x = cupy.random.random(10).astype(cupy.complex64)
plan = cupyx.scipy.fftpack.get_fft_plan(x)
with scipy.fft.set_backend(cu_fft):
    y = scipy.fft.fft(x, plan=plan)
```
This will fail with SciPy v1.4.0 but pass with v1.5.0. This is because the check (`plan != None`) is done [at the pocketfft level](https://github.com/scipy/scipy/blob/832bdf3b0378f05bd0f74ff5f00464524729c6d8/scipy/fft/_pocketfft/basic.py#L16-L18), so once the backend is switched from pocketfft to cupy, scipy will be happy.

asi1024:
This case does not seem to need to be considered because `plan` is created outside of `with scipy.fft.set_backend(cu_fft)` scope.

asi1024:
Jenkins, test this please.

leofang:
> This case does not seem to need to be considered because `plan` is created outside of `with scipy.fft.set_backend(cu_fft)` scope.

I don't think it matters. The docstring from scipy says: `This argument is reserved for passing in a precomputed plan provided by downstream FFT vendors.`

asi1024:
CuPy follows only the latest version of NumPy/SciPy about compatibility, so we can remove the experimental warning after SciPy 1.5 is released. Sorry that we have not yet written about it at development policies in Wiki.

leofang:
Thanks, @asi1024. I am a bit confused, what else do I need to do here?

asi1024:
The code fixes in `cupyx/scipy/fftpack/fft.py` look unrelated to this PR. Could you explain why they are needed?

leofang:
> The code fixes in `cupyx/scipy/fftpack/fft.py` look unrelated to this PR. Could you explain why they are needed?

I suppose you're referring to 1c6174d? It's because `get_fft_plan()` is used in the tests in `tests/cupyx_tests/scipy_tests/fft_tests/test_fft.py`, but there the assumption was that the bug for the output shape mismatching the input shape was fixed by #2752, whereas the same bug also appeared in `get_fft_plan()` but we failed to catch. Without fixing it the tests won't pass.

asi1024:
I understand! Thanks!

asi1024:
LGTM! Thank you for the PR!

leofang:
Thank you, @asi1024 🙂 I’ll send a second PR to test against scipy.fft soon.

PR #3009, opened 2020-01-29
Use `size_t nbytes` in `__cuda_array_interface__`

jakirkham:


jakirkham:
cc @kkraus14 @shwina @leofang

leofang:
> We're actively hitting this issue in a cuDF example and this fix looks good to me

Interesting. Do you have a pointer to the issue?

kkraus14:
> Interesting. Do you have a pointer to the issue?

Yep, you can see the test failure here: https://gpuci.gpuopenanalytics.com/job/docker/job/tests/job/docker-test-notebooks/CUDA_VERSION=10.0,LINUX_VERSION=ubuntu16.04,PYTHON_VERSION=3.6/lastCompletedBuild/testReport/(root)/cudf/notebooks_numba_cuDF_integration/

jakirkham:
@kmaehashi, would it be possible for someone to look at this? It's causing us a bit of pain atm. Am planning to include it as a patch in the conda-forge packages, but would like to make sure this seems reasonable first. So a quick review should be sufficient for us near term 🙂

emcastillo:
Jenkins, test this please

jakirkham:
Thanks for the review @emcastillo! 😄

<hr>

> Sounds like we should add a similar test to cupy!

Agreed, @leofang. Where would you suggest we add it? 🙂

jakirkham:
Submitted a backport to `v6` with PR ( https://github.com/cupy/cupy/pull/3014 ).

leofang:
> > Sounds like we should add a similar test to cupy!
> 
> Agreed, @leofang. Where would you suggest we add it? 🙂

See #3297.

PR #3011, opened 2020-01-30
Fix `fill_diagonal`

### LINKED ISSUE #2970: Columnar matrix slice doesn't work properly with ravel
When creating a view into a 2d array using `[:, n:]` notation, the resulting view doesn't work properly with `ravel` or functions that depend on it.

```
import cupy as cp
a = cp.zeros((3, 3))
b = a[:, 1:]
b.ravel()[0] = 1
print(b)
b[0, 0] = 2
print(a)
```

``` output
[[0. 0.]
 [0. 0.]
 [0. 0.]]
[[0. 2. 0.]
 [0. 0. 0.]
 [0. 0. 0.]]
```

As you can see, `b[0, 0]` still works as expected, but the array produced by `ravel` is a copy. This is not the case for the array produced with a row-wise slice:

```
c = a[1:, :]
c.ravel()[0] = 3
print(a)
```

``` output
array([[0., 2., 0.],
       [3., 0., 0.],
       [0., 0., 0.]])
```

This affects me because I'm trying to fill out a tridiagonal matrix using `cupy`, but the upper diagonal can't be properly referenced:
```
d = cp.zeros((3, 3))
cp.fill_diagonal(d, 3)
cp.fill_diagonal(d[1:], 2)
cp.fill_diagonal(d[:, 1:], 1)
print(d)
```

``` output
[[3. 0. 0.]
 [2. 3. 0.]
 [0. 2. 3.]]
```

```
In [182]: cp.show_config()
CuPy Version          : 7.1.0
CUDA Build Version    : 10000
CUDA Driver Version   : 10000
CUDA Runtime Version  : 10000
cuBLAS Version        : 10000
cuFFT Version         : 10000
cuRAND Version        : 10000
cuSOLVER Version      : (10, 0, 0)
cuSPARSE Version      : 10000
NVRTC Version         : (10, 0)
cuDNN Build Version   : 7605
cuDNN Version         : 7600
NCCL Build Version    : 2406
NCCL Runtime Version  : 2406
```

takagi:
Fix #2970 .

`cupy.fill_diagonal()` is currently implemented with `cupy.ndarray.ravel()` as opposed to NumPy that implements `numpy.fill_diagonal()` with `numpy.ndarray.flat`, that keeps off properly filling the value in the main diagonal of an array.

This PR fixes the problem with providing the counterpart of `numpy.flatiter` to touch an array by means of the flatten index space with 1D basic slicing.

### DIFF
--- a/cupy/indexing/insert.py
+++ b/cupy/indexing/insert.py
@@ -104,6 +104,7 @@ def _fill_diagonal_kernel(a, val):
     return __fill_diagonal_kernel(
         _scalar.get_typename(a.dtype), a.ndim, val.ndim)
 
+
 def fill_diagonal(a, val, wrap=False):
     """Fills the main diagonal of the given array of any dimensionality.
 
@@ -153,7 +154,7 @@ def fill_diagonal(a, val, wrap=False):
             raise ValueError(
                 'Array device must be same as the current '
                 'device: array device = %d while current = %d'
-                % (arr.data.device_id, device_id))
+                % (arr.data.device_id, dev_id))
 
     fill_diagonal_kernel = _fill_diagonal_kernel(a, val)
 
--- a/cupy/indexing/insert.py
+++ b/cupy/indexing/insert.py
@@ -129,7 +129,7 @@ def fill_diagonal(a, val, wrap=False):
     # The followings are imported from the original numpy
     if a.ndim < 2:
         raise ValueError('array must be at least 2-d')
-    end = None
+    end = a.size
     if a.ndim == 2:
         step = a.shape[1] + 1
         if not wrap:
--- a/cupy/indexing/insert.py
+++ b/cupy/indexing/insert.py
@@ -3,6 +3,7 @@ import string
 import numpy
 
 import cupy
+from cupy import util
 from cupy.core import _carray
 from cupy.core import _scalar
 from cupy.cuda import device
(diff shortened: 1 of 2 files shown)


cupy/indexing/insert.py
```diff
@@ -107,6 +131,25 @@ def fill_diagonal(a, val, wrap=False):
...
+    for arr in [a, val]:
+        if arr.data.device_id != dev_id:
+            raise ValueError(
+                'Array device must be same as the current '
+                'device: array device = %d while current = %d'
+                % (arr.data.device_id, device_id))
+
+    code = _fill_diagonal_template.substitute(
+        type=_scalar.get_typename(a.dtype), a_ndim=a.ndim, val_ndim=val.ndim)
+    fill_diagonal_kernel = cupy.RawKernel(code, 'cupy_fill_diagonal')
```
emcastillo: This should be memoized for performance

takagi:
I removed the draft and will add some tests.

cupy/indexing/insert.py
takagi: I made it memoized!

cupy/indexing/insert.py
```diff
@@ -107,6 +139,24 @@ def fill_diagonal(a, val, wrap=False):
...
         step = 1 + numpy.cumprod(a.shape[:-1]).sum()
 
-    # Since the current cupy does not support a.flat,
-    # we use a.ravel() instead of a.flat
-    a.ravel()[:end:step] = val
+    if not isinstance(val, cupy.ndarray):
+        val = cupy.array(val, dtype=a.dtype)
+    else:
+        val = val.astype(a.dtype, copy=False)
+
```
emcastillo: why not `val = cupy.asarray(val, dtype=a.dtype)`?

emcastillo: Ah, it seems that this is valid in numpy numpy.fill_diagonal(a, [2,3]) ...

maybe just using `cupy.asarray(val)` if val is not a single value and then passing it by value in any other case.

takagi: Yes, I followed NumPy here. And I fixed it to remove the branch!

cupy/indexing/insert.py
emcastillo: awesome!!!

cupy/indexing/insert.py
emcastillo: Thanks!!! 😄

emcastillo:
Jenkins, test this please

emcastillo:
Please fix test

emcastillo:
Jenkins, test this please

PR #3018, opened 2020-02-02
Fix `cupy.random.multivariate_normal`

espg:
Multivariate normal doesn't seem to work... not sure if it's an issue with svd in cuda, or in how the svd is applied in python. Regardless, this PR uses cholesky decomposition, and seems quite fast. Using cholesky also fixes the issue for all versions of cuda where cholesky is working... which is all of them that I have tried.

Removes experimental label.

### DIFF
--- a/cupy/random/distributions.py
+++ b/cupy/random/distributions.py
@@ -439,7 +439,7 @@ def negative_binomial(n, p, size=None, dtype=int):
 
 
 def multivariate_normal(mean, cov, size=None, check_valid='ignore',
-                        tol=1e-08, dtype=float):
+                        tol=1e-08, method='cholesky', dtype=float):
     """Multivariate normal distribution.
 
     Returns an array of samples drawn from the multivariate normal
@@ -462,6 +462,13 @@ def multivariate_normal(mean, cov, size=None, check_valid='ignore',
             matrix is not positive semidefinite.
         tol (float): Tolerance when checking the singular values in
             covariance matrix.
+        method : { 'cholesky', 'eigh', 'svd'}, optional
+            The cov input is used to compute a factor matrix A such that
+            ``A @ A.T = cov``. This argument is used to select the method
+            used to compute the factor matrix A. The default method 'svd' is
+            the fastest, while 'svd' is the slowest but more robust than
+            the fastest method. The method `eigh` uses eigen decomposition to
+            compute A and is faster than svd but slower than cholesky.
         dtype: Data type specifier. Only :class:`numpy.float32` and
             :class:`numpy.float64` types are allowed.
 
--- a/cupy/random/distributions.py
+++ b/cupy/random/distributions.py
@@ -475,6 +475,18 @@ def multivariate_normal(mean, cov, size=None, check_valid='ignore',
     Returns:
         cupy.ndarray: Samples drawn from the multivariate normal distribution.
 
+    .. note:: Default `method` is set to fastest, 'cholesky'. Cholesky
+        decomposition in CuPy will fail silently if the input covariance matrix
+        is not positive definite and give invalid results, unlike in numpy,
+        where an invalid covariance matrix will raise an exception. Setting
+        `check_valid` to 'raise' will replicate numpy behavior by checking
+        the input, but will also force device synchronization. If validity of
+        input is unknown, setting `method` to 'einh' or 'svd' and
+        `check_valid` to 'warn' will use cholesky decomposition for positive
+        definite matrices, and fallback to the specified `method` for other
+        matrices (i.e., positive semi-definite), and will warn if
+        decomposition is suspect.
+
     .. seealso:: :meth:`numpy.random.multivariate_normal
                  <numpy.random.mtrand.RandomState.multivariate_normal>`
 
--- a/cupy/random/distributions.py
+++ b/cupy/random/distributions.py
@@ -438,9 +438,8 @@ def negative_binomial(n, p, size=None, dtype=int):
     return rs.negative_binomial(n, p, size=size, dtype=dtype)
 
 
-def multivariate_normal(mean, cov, size=None, check_valid='ignore', tol=1e-8,
-                        dtype=float):
-    """(experimental) Multivariate normal distribution.
+def multivariate_normal(mean, cov, size=None, dtype=float):
+    """Multivariate normal distribution.
 
     Returns an array of samples drawn from the multivariate normal
     distribution. Its probability density function is defined as
@@ -458,15 +457,18 @@ def multivariate_normal(mean, cov, size=None, check_valid='ignore', tol=1e-8,
             symmetric and positive-semidefinite for proper sampling.
         size (int or tuple of ints): The shape of the array. If ``None``, a
             zero-dimensional array is generated.
-        check_valid ('warn', 'raise', 'ignore'): Behavior when the covariance
-            matrix is not positive semidefinite.
-        tol (float): Tolerance when checking the singular values in
-            covariance matrix.
         dtype: Data type specifier. Only :class:`numpy.float32` and
             :class:`numpy.float64` types are allowed.
 
     Returns:
         cupy.ndarray: Samples drawn from the multivariate normal distribution.
+
+    .. warning::
+    This function calls one or more cuSOLVER routine(s) which may yield
+    invalid results if input conditions are not met.
+    To detect these invalid results, you can set the `linalg`
+    configuration to a value that is not `ignore` in
+    :func:`cupyx.errstate` or :func:`cupyx.seterr`.
 
     .. seealso:: :meth:`numpy.random.multivariate_normal
                  <numpy.random.mtrand.RandomState.multivariate_normal>`
--- a/cupy/random/distributions.py
+++ b/cupy/random/distributions.py
@@ -437,7 +437,8 @@ def negative_binomial(n, p, size=None, dtype=int):
     return rs.negative_binomial(n, p, size=size, dtype=dtype)
 
 
-def multivariate_normal(mean, cov, size=None, dtype=float):
+def multivariate_normal(mean, cov, check_valid='ignore', tol=1e-08,
+                        size=None, dtype=float):
     """Multivariate normal distribution.
 
     Returns an array of samples drawn from the multivariate normal
@@ -456,25 +457,22 @@ def multivariate_normal(mean, cov, size=None, dtype=float):
             symmetric and positive-semidefinite for proper sampling.
         size (int or tuple of ints): The shape of the array. If ``None``, a
             zero-dimensional array is generated.
+        check_valid ('warn', 'raise', 'ignore'): Behavior when the covariance
+            matrix is not positive semidefinite.
+        tol (float): Tolerance when checking the singular values in
+            covariance matrix.
         dtype: Data type specifier. Only :class:`numpy.float32` and
             :class:`numpy.float64` types are allowed.
 
     Returns:
         cupy.ndarray: Samples drawn from the multivariate normal distribution.
 
-    .. warning::
-        This function calls one or more cuSOLVER routine(s) which may yield
-        invalid results if input conditions are not met.
-        To detect these invalid results, you can set the `linalg`
-        configuration to a value that is not `ignore` in
-        :func:`cupyx.errstate` or :func:`cupyx.seterr`.
-
     .. seealso:: :meth:`numpy.random.multivariate_normal
                  <numpy.random.mtrand.RandomState.multivariate_normal>`
 
     """
     rs = generator.get_random_state()
-    x = rs.multivariate_normal(mean, cov, size, dtype)
+    x = rs.multivariate_normal(mean, cov, check_valid, tol, size, dtype)
     return x
 
 
--- a/cupy/random/distributions.py
+++ b/cupy/random/distributions.py
@@ -465,8 +465,8 @@ def multivariate_normal(mean, cov, size=None, check_valid='ignore',
         method : { 'cholesky', 'eigh', 'svd'}, optional
             The cov input is used to compute a factor matrix A such that
             ``A @ A.T = cov``. This argument is used to select the method
-            used to compute the factor matrix A. The default method 'svd' is
-            the fastest, while 'svd' is the slowest but more robust than
+            used to compute the factor matrix A. The default method 'cholesky'
+            is the fastest, while 'svd' is the slowest but more robust than
             the fastest method. The method `eigh` uses eigen decomposition to
             compute A and is faster than svd but slower than cholesky.
         dtype: Data type specifier. Only :class:`numpy.float32` and
@@ -475,17 +475,17 @@ def multivariate_normal(mean, cov, size=None, check_valid='ignore',
     Returns:
         cupy.ndarray: Samples drawn from the multivariate normal distribution.
 
-    .. note:: Default `method` is set to fastest, 'cholesky'. Cholesky
-        decomposition in CuPy will fail silently if the input covariance matrix
-        is not positive definite and give invalid results, unlike in numpy,
-        where an invalid covariance matrix will raise an exception. Setting
-        `check_valid` to 'raise' will replicate numpy behavior by checking
-        the input, but will also force device synchronization. If validity of
-        input is unknown, setting `method` to 'einh' or 'svd' and
-        `check_valid` to 'warn' will use cholesky decomposition for positive
-        definite matrices, and fallback to the specified `method` for other
-        matrices (i.e., positive semi-definite), and will warn if
-        decomposition is suspect.
+    .. note:: Default `method` is set to fastest, 'cholesky', unlike numpy
+        which defaults to 'svd'. Cholesky decomposition in CuPy will fail
+        silently if the input covariance matrix is not positive definite and
+        give invalid results, unlike in numpy, where an invalid covariance
+        matrix will raise an exception. Setting `check_valid` to 'raise' will
+        replicate numpy behavior by checking the input, but will also force
+        device synchronization. If validity of input is unknown, setting
+        `method` to 'einh' or 'svd' and `check_valid` to 'warn' will use
+        cholesky decomposition for positive definite matrices, and fallback to
+        the specified `method` for other matrices (i.e., not positive
+        semi-definite), and will warn if decomposition is suspect.
 
     .. seealso:: :meth:`numpy.random.multivariate_normal
                  <numpy.random.mtrand.RandomState.multivariate_normal>`
--- a/cupy/random/distributions.py
+++ b/cupy/random/distributions.py
@@ -437,8 +437,8 @@ def negative_binomial(n, p, size=None, dtype=int):
     return rs.negative_binomial(n, p, size=size, dtype=dtype)
 
 
-def multivariate_normal(mean, cov, check_valid='ignore', tol=1e-08,
-                        size=None, dtype=float):
+def multivariate_normal(mean, cov, size=None, check_valid='ignore',
+                        tol=1e-08, dtype=float):
     """Multivariate normal distribution.
 
     Returns an array of samples drawn from the multivariate normal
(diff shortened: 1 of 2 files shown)


cupy/random/distributions.py
```diff
@@ -438,9 +437,8 @@ def negative_binomial(n, p, size=None, dtype=int):
     return rs.negative_binomial(n, p, size=size, dtype=dtype)
 
 
-def multivariate_normal(mean, cov, size=None, check_valid='ignore', tol=1e-8,
```
emcastillo: the `check_valid` and `tol` args are needed to keep compatibility with numpy

https://docs.scipy.org/doc/numpy/reference/random/generated/numpy.random.Generator.multivariate_normal.html?highlight=multivariate_normal#numpy.random.Generator.multivariate_normal

emcastillo:
Checks should not be removed, please try to apply them to your implementation.

cupy/random/distributions.py
espg: this should be fixed now. Since the `check_valid` option is in place, I removed the other warning... in numpy, cholesky will fail non-silently on a non-semipositive definite matrix, so `check_valid` doesn't have much meaning there. I'm not sure of the behavior for cupy, but the warning attached to cholesky about conditions needing to be met makes me think that the routine doesn't explicitly check the state of the matrix with regards to semipositive definiteness (hence the option of `cupyx.errstate()` or `cupyx.seterr()` to explicitly warn on that decomposition). The way things are set now, is that there should be a warning regardless of the state of `check_valid` if `cupyx.errstate()` or `cupyx.seterr()` is invoked; if they are not, and `check_valid` is set to anything other than 'ignore', an extra eigh call will test that the covariance matrix input supplied is valid. Cholesky should always be faster than svd, and eigh is slightly slower than cholesky...but since this wasn't working earlier, even if `check_valid` is used, the routine will now at least run and complete.

espg:
I've tested this, and it looks like it finally handles requests for draws of varying sizes correctly; a 9000 square covariance matrix returns a vector of 9000 values if `size` isn't included; if a hundred samples are requested, the return is an array of shape 100 by 9000, such that each row is a sample draw. I also tried the `check_valid` flag... it works (does not crash), but I was wrong about how much slower eigh is, so it's slow. I can pull 50000 draws of a multivariate distribution described by a 9000 square covariance matrix in under a second if `check_valid` is on the default value of 'ignore'; if `check_valid` is enabled, it adds 44 seconds =/

niboshi:
I agree experimental could be removed, but let's not do that in this PR.
Each PR should have only one clear topic.

cupy/random/distributions.py
emcastillo: Yes, I agree with what you say and totally understand it. 😄
However, using cholesky inside this function is an implementation detail and the interface exposed to the users should not deviate from the numpy one.
For example cupy fails silently with cholesky
```
>>> a=cupy.arange(9).reshape(3,3)
>>> cupy.linalg.cholesky(a)
array([[nan,  0.,  0.],
       [inf, nan,  0.],
       [inf, nan, nan]])
```
We should try to deal with cholesky error states inside this function and then adapt them to what the interface requires, even if its redundant as we strive towards compatibility.
Sorry to keep asking for changes 😔

espg:
ok, used `cupyx.seterr()` to catch invalid inputs... I think the behavior is very similar to numpy, since the `seterr()` exception is just an imported numpy LinAlgError. I added a print statement to give a bit more context about not being positive definite... the underlying cuSOLVER error just gives the row of the error I think, instead of the cause.

cupy/random/generator.py
```diff
@@ -341,29 +349,30 @@ def multivariate_normal(self, mean, cov, size=None, check_valid='ignore',
...
+                                  RuntimeWarning)
                 else:
-                    raise ValueError(
-                        'covariance is not symmetric positive-semidefinite.')
+                    raise ValueError("covariance is not " +
+                                     "positive-semidefinite.")
 
-        x = cupy.dot(x, cupy.sqrt(s)[:, None] * v)
+        x = self.standard_normal(size=shape, dtype=dtype)
+        seterr(linalg='raise')
```
emcastillo: This will force device synchronization every time,
how about `seterr(linalg=check_valid)` I think it is not as precise as numpy but is ok just to document this.
@niboshi what are your thoughts?

emcastillo: Also, please use `cupyx.errstate` since its a context manager and will restore the configuration to what it was before.
`with cupyx.errstate(linalg=check_valid`

espg: Ok, I use the context manager now. I think that we want `linalg='raise'` instead of `linalg=check_state` though; otherwise the cholesky decomposition example that you gave with `a=cupy.arange(9).reshape(3,3)` will not raise an exception to mirror numpy behavior on equivalent input. I checked, and `linalg` does revert back to previous state after the function call. Being able to use check tolerance is probably unnecessary besides for not breaking code, but maybe it still has utility in some cases? If `check_valid` is used, it certainly will verify that the covariance matrix is valid by an independent call to `einh`, so it works as advertised... I just can't think of a case where cholesky doesn't fail but the covariance matrix is invalid, but perhaps there is a corner case that I don't know of.

espg: I guess the issue is that `cupyx.errstate` only has `ignore` and `raise` and there is no `warn` implemented; so do you want to be conservative and `raise` more than numpy does? Or ignore and **not** have run time warnings for invalid input, even though numpy will warn regardless of `check_valid`? I think raising is reasonable, but I'm curious what behavior you want to target...

emcastillo: Thanks for the changes!
Yes, I understand that. However, in the current state, this function will raise an exception even if `check_valid` is set to ignore.
The problem here is that `errstate(linalg='raise')` forces device synchronization thus degrading performance, so we should avoid it in the case of `check_valid='ignore'`.

I am in favor of emulating the warn behavior by using raise and catching the exception.
I am sorry that my explanations are a bit confusing and you have to keep working on it, but we are close to finish it 😅

espg:
Ok, I think this is good now. It's way faster to catch the `LinAlgError` than to check validity via `einh`...but `einh` decomposition works in some cases when cholesky decomposition does not. Here's how the current behavior is set:

1. `check_valid='ignore'` (default): Assumes that the user is giving a covariance matrix that is known to be valid, or has set `cupyx.errstate` independently to catch errors. No check is performed, decomposition is done fast via cholesky.

2. `check_valid='warn'` : Cholesky is faster than einh, and gives valid results with an appropriate covariance matrix; context manager sets `cupyx.errstate` to raise, and catches the exception. If there is no exception, great, we are done. If there is an exception, a warning is printed that the matrix is not Positive Definite, and then the covariance matrix is decomposed via eigenvalue decomposition. The submatrix `s` is then checked against `tol` to assess validity; if the assessment fails, a warning is generated that the covariance matrix is not positive-semidefinite. Regardless of if `s` fails, values are returned using the eigenvalue decomposition. A well behaved covariance matrix should return almost as fast when `check_valid='ignore'` since there is just a sync call and covariance is still decomposed by cholesky. A less well behaved covariance structure will do a best attempt at random draws via eigenvalue decomposition, and will warn if the output is possibly invalid.

3. `check_valid='raise'` : Cholesky with context manager set to raise; if error is raised, advises that setting `check_valid='warn'` will be successful if the covariance is positive-semidefinite.

espg:
Slightly modified error and warning messages... here's the behavior:

```
In [1]: import cupy as cp                                                                                                                                        

In [2]: a = cp.arange(9).reshape(3,3)                                                                                                                            

In [3]: cp.random.multivariate_normal(a[:,0],a,size=(10))                                                                                                        
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/distributions.py:475: FutureWarning: cupy.random.multivariate_normal is experimental. The interface can change in the future.
  util.experimental('cupy.random.multivariate_normal')
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/generator.py:336: FutureWarning: cupy.random.RandomState.multivariate_normal is experimental. The interface can change in the future.
  util.experimental('cupy.random.RandomState.multivariate_normal')
Out[3]: 
array([[ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan]])

In [4]: cp.random.multivariate_normal(a[:,0],a,size=(10),check_valid='warn')                                                                                     
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/distributions.py:475: FutureWarning: cupy.random.multivariate_normal is experimental. The interface can change in the future.
  util.experimental('cupy.random.multivariate_normal')
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/generator.py:336: FutureWarning: cupy.random.RandomState.multivariate_normal is experimental. The interface can change in the future.
  util.experimental('cupy.random.RandomState.multivariate_normal')
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/generator.py:374: RuntimeWarning: covariance is not positive-semidefinite, output may be invalid.
  "invalid.", RuntimeWarning)
Out[4]: 
array([[ 1.01712647,  1.30131418,  3.82760142],
       [-2.95376009,  0.44119319,  7.32323061],
       [-0.84087281,  3.68106004,  7.9623521 ],
       [ 1.30269436,  4.12290396,  4.95423207],
       [ 2.3360361 ,  5.51396429,  8.14745924],
       [-2.39305151, -2.6835412 , -3.97150818],
       [-0.95064085,  2.43928865,  4.86455833],
       [-3.50082292,  2.04941799,  5.13111369],
       [-2.72635825,  0.56307246,  0.84395636],
       [-1.70118353,  2.54796699,  6.43011132]])

In [5]: cp.random.multivariate_normal(a[:,0],a,size=(10),check_valid='raise')                                                                                    
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/distributions.py:475: FutureWarning: cupy.random.multivariate_normal is experimental. The interface can change in the future.
  util.experimental('cupy.random.multivariate_normal')
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/generator.py:336: FutureWarning: cupy.random.RandomState.multivariate_normal is experimental. The interface can change in the future.
  util.experimental('cupy.random.RandomState.multivariate_normal')
LinAlgError                               Traceback (most recent call last)
/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/generator.py in multivariate_normal(self, mean, cov, size, check_valid, tol, dtype)
    377                 try:
--> 378                     decomp = cupy.linalg.cholesky(cov)
    379                 except LinAlgError:

/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/linalg/decomposition.py in cholesky(a)
    153     cupy.linalg.util._check_cusolver_dev_info_if_synchronization_allowed(
--> 154         potrf, dev_info)
    155 

/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/linalg/util.py in _check_cusolver_dev_info_if_synchronization_allowed(routine, dev_info)
     46             ' to the cuSOLVER documentation.'.format(
---> 47                 routine.__name__, dev_info_host))
     48 

LinAlgError: Error reported by dpotrf in cuSOLVER. devInfo = 2. Please refer to the cuSOLVER documentation.

During handling of the above exception, another exception occurred:

LinAlgError                               Traceback (most recent call last)
<ipython-input-5-09b2141ce44d> in <module>
----> 1 cp.random.multivariate_normal(a[:,0],a,size=(10),check_valid='raise')

/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/distributions.py in multivariate_normal(mean, cov, size, check_valid, tol, dtype)
    475     util.experimental('cupy.random.multivariate_normal')
    476     rs = generator.get_random_state()
--> 477     x = rs.multivariate_normal(mean, cov, size, check_valid, tol, dtype)
    478     return x
    479 

/home/jupyter/magma/min_test/_build/pip_packages/lib/python3.6/site-packages/cupy/random/generator.py in multivariate_normal(self, mean, cov, size, check_valid, tol, dtype)
    380                     raise LinAlgError("Matrix is not positive definite; if " +
    381                                       "matrix is positive-semidefinite, set" +
--> 382                                       "'check_valid' to 'warn'")
    383 
    384         x = self.standard_normal(size=shape, dtype=dtype)

LinAlgError: Matrix is not positive definite; if matrix is positive-semidefinite, set'check_valid' to 'warn'
```

cupy/random/generator.py
espg: @emcastillo How is the behavior now? Does it match what you would like?

emcastillo: Yes, it looks better now!
Thanks a lot for the changes

emcastillo:
Why is the result different when is set to ignore?
```
In [3]: cp.random.multivariate_normal(a[:,0],a,size=(10))    
array([[ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan],
       [ 0., nan, nan]])

In [4]: cp.random.multivariate_normal(a[:,0],a,size=(10),check_valid='warn')                                                                                     
array([[ 1.01712647,  1.30131418,  3.82760142],
       [-2.95376009,  0.44119319,  7.32323061],
       [-0.84087281,  3.68106004,  7.9623521 ],
       [ 1.30269436,  4.12290396,  4.95423207],
       [ 2.3360361 ,  5.51396429,  8.14745924],
       [-2.39305151, -2.6835412 , -3.97150818],
       [-0.95064085,  2.43928865,  4.86455833],
       [-3.50082292,  2.04941799,  5.13111369],
       [-2.72635825,  0.56307246,  0.84395636],
       [-1.70118353,  2.54796699,  6.43011132]])
```

emcastillo:
Now every case gives nan
```
>>> cp.random.multivariate_normal(a[:,0],a,size=(10))
array([[nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan]])
>>> cp.random.multivariate_normal(a[:,0],a,size=(10),check_valid='warn')
/home/ecastill/em-cupy/cupy/random/generator.py:386: RuntimeWarning: covariance is not positive-semidefinite, output *is* invalid.
  "invalid.", RuntimeWarning)
array([[nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan]])
```

espg:
...what do you think the behavior should be? =) The covariance matrix is invalid and the default is ignore... So it is ignored. When set to warn, it warns. We could force an exception to be raised fairly easily... that would perhaps mirror numpy? Even if the numpy version is set to ignore, I think cholesky decomposition will still fail and raise an exception with an invalid matrix (i.e., you cannot get an invalid result-- either it is valid or it catches an exception regardless of raise, warn, or ignore). 

Is that what you want for cupy? To enforce validity checks for all random draws that use cholesky? Warn, ignore, and raise would still have meaning for svd and einh, but would effectively be ignored for cholesky...

emcastillo:
I was expecting a behavior similar to NumPy, despite the covariance matrix being invalid it is returning actual numeric values. Whether they make sense or not is a different story ...

```
>>> import numpy as np
>>> a = np.arange(9).reshape(3,3)
>>> np.random.multivariate_normal(a[:,0],a,size=(10), check_valid='ignore')
array([[ 0.24316235,  4.6162679 ,  8.9893734 ],
       [ 1.5980354 ,  3.53620681,  5.47437821],
       [-1.87523916,  1.30797166,  4.49118246],
       [-0.31591572,  1.56867447,  3.45326467],
       [ 0.26105545,  2.43474871,  4.60844208],
       [-1.11015744,  0.68545803,  2.48107357],
       [-0.60454643,  1.49703322,  3.59861298],
       [ 3.51204825,  7.0907185 , 10.66938878],
       [-0.17049917,  4.59663895,  9.36377707],
       [ 1.60576213,  4.41777115,  7.2297802 ]])
>>> np.random.multivariate_normal(a[:,0],a,size=(10))
__main__:1: RuntimeWarning: covariance is not positive-semidefinite.
array([[-0.45499166,  3.61370574,  7.68240308],
       [ 1.33893504,  3.73527316,  6.13161125],
       [-3.22579971, -0.51345463,  2.19889041],
       [ 1.97965279,  3.66275213,  5.34585149],
       [ 4.56059964,  7.08999223,  9.61938485],
       [-1.07868833,  1.67256142,  4.42381115],
       [ 1.76454191,  5.29102727,  8.81751258],
       [-2.49652308, -0.21727617,  2.06197072],
       [-1.39390224,  1.51902721,  4.43195671],
       [ 1.12090272,  2.55613096,  3.99135922]])
```

Should NumPy output be considered garbage? or does it have an actual use.

espg:
I'd say the numpy results are probably garbage given the covariance matrix...

Also, the numpy version has the method set to use svd as the default instead of cholesky-- different philosophy, tries to give results for the broadest number of cases by default, instead of the fastest by default =P Since people using cupy are using for speed, using the fastest by default i thought made more sense... anyway, if you use the method keyword of cholesky on numpy, I'm sure you'll get an error or nans too.

There is a good compromise I think though-- we keep svd as the default to match numpy. Since cholesky is so much faster and gives identical results if the covariance matrix is sane, i wrote the warn cases to try with cholesky and only bother checking tolerances if a linear algebra exception gets raised... I.e., if there's no exception, it just uses cholesky and quickly returns; if it's a difficult covariance, they cholesky fails quickly and then you fall back to a best effort with either einh or svd to get a result and check the robustness-- so a fast result for simple cases, and a slow and maybe suspect result for hard cases all using the same keyword argument

espg:
The only problem with that is the ignore case-- the above logic will run fast for raise, fast for warn, and fast for keyword method of cholesky... But it will run slower for ignore and svd in combination; which is of course the numpy default >.<

Unless we ignore the ignore keyword and still do a check using cholesky?? I dunno what the best solution is or how exact you want to match numpy...we could also default to warn instead of ignore which would give good default speed

emcastillo:
Awesome,

```
>>> from numpy.random import Generator, PCG64
>>> rg = Generator(PCG64())
>>> rg..multivariate_normal(a[:,0],a,size=(10), check_valid='ignore', method='cholesky')
```
Even with this it fails 😂

Now everything is crystal clear!

Can you add a `.. note::` in the docstring talking about this and the difference with numpy?
Once that is done, this is good to go.
``` 
.. note:: Text here
    ident 4 spaces for remaining text.
```

espg:
Sure-- it will be tomorrow though, I only have my phone now and not my computer >.>

I take it you want to enforce failure for cholesky regardless of warn statement to match numpy behavior? And then use the fast hybrid cholesky precheck/results with svd fallback as the the default (i.e., svd + warn for cupy, instead of svd + ignore for numpy)?

emcastillo:
No, let's not enforce it to avoid device synchronization.
As long as its documented it is fine.

espg:
All set!

emcastillo:
Jenkins, test this please

emcastillo:
Please check test failures, thanks!

espg:
fixed now?

emcastillo:
Jenkins, test this please

espg:
Jenkins, test this please

espg:
guess I can't call jenkins...

The last failure was due to a single extra/wrong keyword (shape instead of size... and neither is actually needed since it is the positional first argument).

This should be fixed (no errors on multidimensional sample requests), but I won't know for sure till the test are rerun :-/

emcastillo:
Jenkins, test this please

PR #3020, opened 2020-02-03
Fix empty vector access

### LINKED ISSUE #3016: CuPy crashes on anything that just isn't memory allocation
**Conditions**

```
CuPy Version          : 7.1.1
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSPARSE Version      : 10301
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2506
NCCL Runtime Version  : 2506
```
Running on Fedora 30

---------------

Built and installed CuPy in a virtual environment with:
```
export CUDA=/usr/local/cuda
export PATH="$CUDA/bin:$PATH"
export LD_LIBRARY_PATH="$CUDA/lib64:$LD_LIBRARY_PATH"
export CUDA_INC_DIR="$CUDA/include"
export NVCC="nvcc -ccbin /usr/bin/cuda-gcc"
pip install cupy
```
Then ran the following code from the tutorial:
```
import cupy as cp
x_gpu = cp.array([1, 2, 3]) # any allocation method works fine, can see data allocated on GPU
l2_gpu = cp.linalg.norm(x_gpu) # any operation, even multiply by 2 or astype()
```
Crashes with:
```
/usr/include/c++/9/bits/stl_vector.h:1042: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = const char*; _Alloc = std::allocator<const char*>; std::vector<_Tp, _Alloc>::reference = const char*&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
Aborted (core dumped)
```

I used valgrind to get the stack trace since that error simply states an assert in stl::vector<> failed. A few more levels of the stack trace reveal:
```
==25539==    by 0x56CAD70A: operator[] (stl_vector.h:1042)
==25539==    by 0x56CAD70A: __pyx_f_4cupy_4cuda_5nvrtc_createProgram(_object*, _object*, _object*, _object*, int) (nvrtc.cpp:2311)
==25539==    by 0x56CADA89: __pyx_pf_4cupy_4cuda_5nvrtc_4createProgram (nvrtc.cpp:2473)
==25539==    by 0x56CADA89: __pyx_pw_4cupy_4cuda_5nvrtc_5createProgram(_object*, _object*, _object*) (nvrtc.cpp:2455)
```

So it is in the `createProgram()` Cython function when it uses the `std::vector<>::operator[]()` function. There are only two places in that function that ends up applying to, `&(header_vec[0])` and `&(include_name_vec[0])`. It appears that the vectors are empty and the STL library has an assert that checks for that (although I thought it wasn't with `[]` but was supposed to with `at()`). I did confirm that `cupy.cuda.compiler._NVRTCProgram.__init__()` (the only place `createProgram()` is called) is indeed calling it with an empty tuple for `headers` and `include_names` (which are the default values for those arguments).

Doing a  bit of research into this area, it seems that the macro `_GLIBCXX_ASSERTIONS` is being defined at some point (possibly as a default in GCC for Fedora but I can't find it in the output of `/usr/bin/cuda-g++ -dM -E -x c++ - < /dev/null`) which causes the `LIKELY()` in the `std::vector<>::operator[]` to become an assert. It seems that there are three solutions:

- Have the end-user manually undefine this macro, but seeing that I don't know what is adding it that may be difficult.
- Undefine it within the Cython source
- Don't use `&(x[0])` construct without first checking size, could do `&(x[0]) if x.size() else NULL` instead
  - If you are willing to limit to C++11 only, you can use the `std::vector<>::data()` method to get this pointer as well
  - In these particular cases a simple `malloc()`/`free()` would actually work quite well and wouldn't incur the potential overhead of dynamic resizing

My hasty search shows that this pattern shows up 3 times, all in cupy/cuda/nvrtc.pyx. According to the documentation, `nvrtcCreateProgram()` does support `NULL` for those arguments specifically (when the length is 0, which it would be in that case).

emcastillo:
Closes #3016

In some linux distributions accessing &(array[0]) can trigger an assertion in libstdc++
thanks to @coderforlife for providing details on this and how to fix it.

### DIFF
--- a/cupy/cuda/nvrtc.pyx
+++ b/cupy/cuda/nvrtc.pyx
@@ -85,8 +85,8 @@ cpdef intptr_t createProgram(unicode src, unicode name, headers,
 
     with nogil:
         status = nvrtcCreateProgram(
-            &prog, src_ptr, name_ptr, num_headers, &(header_vec[0]),
-            &(include_name_vec[0]))
+            &prog, src_ptr, name_ptr, num_headers, header_vec.data(),
+            include_name_vec.data())
     check_status(status)
     return <intptr_t>prog
 
@@ -107,7 +107,7 @@ cpdef compileProgram(intptr_t prog, options):
 
     with nogil:
         status = nvrtcCompileProgram(<Program>prog, option_num,
-                                     &(option_vec[0]))
+                                     option_vec.data())
     check_status(status)
 
 


toslunar:
Jenkins, test this please.

toslunar:
Should the title be "Fix empty vector access", because it's about `std::vector`?

emcastillo:
Good catch

PR #3022, opened 2020-02-03
Use `.data()` for `std::vector`

toslunar:
Similar to #3020, but I don't know if the input vectors can be empty.

The remaining ones are not of type `std::vector`.
```
% git grep '&.*\[0]'
cupy/cuda/cufft.pyx:        cdef int* shape_ptr = &shape_arr[0]
cupy/cuda/cufft.pyx:            inembed_ptr = &inembed_arr[0]
cupy/cuda/cufft.pyx:            onembed_ptr = &onembed_arr[0]
cupyx/scipy/sparse/compressed.py:        'if (ind == minor) atomicAdd(&answer[0], d);',
cupyx/scipy/sparse/compressed.py:          atomicAdd(&answer_real[0], real);
cupyx/scipy/sparse/compressed.py:          atomicAdd(&answer_imag[0], imag);
```

### DIFF
--- a/cupy/cuda/cudnn.pyx
+++ b/cupy/cuda/cudnn.pyx
@@ -1086,7 +1086,7 @@ cpdef getFilterNdDescriptor(size_t wDesc, int nbDimsRequested):
 
     status = cudnnGetFilterNdDescriptor_v4(
         <FilterDescriptor>wDesc, nbDimsRequested, &dataType,
-        &format, &nbDims, &filterDimA[0])
+        &format, &nbDims, filterDimA.data())
     check_status(status)
     return dataType, format, nbDims, tuple(filterDimA)
 
@@ -1176,7 +1176,7 @@ cpdef findConvolutionForwardAlgorithm(
     status = cudnnFindConvolutionForwardAlgorithm(
         <Handle> handle, <TensorDescriptor>xDesc, <FilterDescriptor>wDesc,
         <ConvolutionDescriptor>convDesc, <TensorDescriptor>yDesc,
-        requestedAlgoCount, &returnedAlgoCount, &perfResults[0])
+        requestedAlgoCount, &returnedAlgoCount, perfResults.data())
     check_status(status)
     perfResults.resize(returnedAlgoCount)
     return perfResults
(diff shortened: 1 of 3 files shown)


emcastillo:
./cupy/cudnn.pyx:913:80: E501 line too long (81 > 79 characters)
./cupy/cudnn.pyx:936:80: E501 line too long (81 > 79 characters)
😅

niboshi:
The value of `vector::data()` is unspecified if the vector is empty.
Are these fixes guaranteed to work even in that case? (#3020 too)

emcastillo:
vector::data() returns null for empty vectors, but maybe this is libstcd++ implementation specific and not in the standard?

emcastillo:
https://stackoverflow.com/questions/25419851/what-should-stdvectordata-return-if-the-vector-is-empty

emcastillo:
Do we know if these variables can be empty?

toslunar:
I've not yet.

kmaehashi:
@emcastillo will check if these variables can be empty.

emcastillo:
For the cudnn is safe since all the calls in cudnn return an error if `requestedAlgoCount < 1` so we can assume that arrays are never empty in that case.
I will keep checking the others and update this comment accordingly

cupy/cuda/cudnn.pyx
```diff
@@ -1487,7 +1487,7 @@ cpdef list getConvolutionBackwardDataAlgorithm_v7(
         <Handle>handle, <FilterDescriptor>filterDesc,
         <TensorDescriptor>diffDesc, <ConvolutionDescriptor>convDesc,
         <TensorDescriptor>gradDesc, requestedAlgoCount,
-        &returnedAlgoCount, &perfResults[0])
+        &returnedAlgoCount, perfResults.data())
     check_status(status)
     perfResults.resize(returnedAlgoCount)
     return [CuDNNAlgoPerf(p.algo, p.status, p.time, p.memory,
```
emcastillo, line 1493: All these are safe

cupy/cuda/function.pyx
```diff
@@ -146,11 +146,11 @@ cdef _launch(intptr_t func, Py_ssize_t grid0, int grid1, int grid2,
     if enable_cooperative_groups:
         driver.launchCooperativeKernel(
             func, <int>grid0, grid1, grid2, <int>block0, block1, block2,
-            <int>shared_mem, stream, <intptr_t>&(kargs[0]))
+            <int>shared_mem, stream, <intptr_t>kargs.data())
     else:
         driver.launchKernel(
             func, <int>grid0, grid1, grid2, <int>block0, block1, block2,
-            <int>shared_mem, stream, <intptr_t>&(kargs[0]), <intptr_t>0)
```
emcastillo, line 153: Function args will never be empty, double-checked.

emcastillo:
Jenkins, test this please

cupy/cuda/function.pyx
leofang: Just curious, even when we use a RawKernel like this?
```cuda
__global__ void f() { /* ... */ }
```

emcastillo: I was discussing with @asi1024 such use case the other day.
Is there any case that someone would want to do this?
We couldn't find any.

leofang: I can only think of very pathological cases of no practical uses 😂 I think it's safe to merge this!

PR #2888, opened 2019-12-26
Make test helper decorators pdb-friendly

niboshi:
This fix makes `pytest --pdb` inspect the traceback of the actual error raised in the test, instead of the traceback of `self.fail()`.

If both cupy and numpy raised errors, `--pdb` stops at the cupy error.

PR #3024, opened 2020-02-04
Make `testing.parameterize` pdb-friendly

niboshi:
Similarly to #2888


emcastillo:
Jenkiins, test this please

PR #3025, opened 2020-02-04
Check size before accesing empty vectors data ptr

emcastillo:
Fixes #3020 as the `vector::data` return value is undefined for empty vectors

### DIFF
--- a/cupy/cuda/nvrtc.pyx
+++ b/cupy/cuda/nvrtc.pyx
@@ -78,15 +78,20 @@ cpdef intptr_t createProgram(unicode src, unicode name, headers,
     cdef int num_headers = len(headers)
     cdef vector.vector[const char*] header_vec
     cdef vector.vector[const char*] include_name_vec
+    cdef const char** header_vec_ptr = NULL
+    cdef const char** include_name_vec_ptr = NULL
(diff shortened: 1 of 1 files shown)


cupy/cuda/nvrtc.pyx
```diff
@@ -78,15 +78,21 @@ cpdef intptr_t createProgram(unicode src, unicode name, headers,
...
+    cdef const char** header_vec_ptr = NULL
+    cdef const char** include_name_vec_ptr = NULL
+    if header_vec.size() > 0:
+        header_vec_ptr = &(header_vec[0])
+    if include_name_vec.size() > 0:
+        include_name_vec_ptr = &(include_name_vec[0])
     for i in headers:
         header_vec.push_back(<const char*>i)
     for i in include_names:
         include_name_vec.push_back(<const char*>i)
```
niboshi: Vectors are fed here. You should check `len(headers)` instead.

(Btw it seems it is implicitly assumed `len(headers) == len(include_names)`. It should be asserted, whether or not in this PR)

niboshi:
Jenkins, test this please

niboshi:
:+1:

PR #3026, opened 2020-02-04
Use `cupy.random` in kmeans example

niboshi:
`numpy.random.choice` is the biggest bottleneck in this example.
With this fix, "GPU" time reduced to about a half in my environment.


emcastillo:
Jenkins, test this please

PR #3028, opened 2020-02-05
Fix nvcc command lookup

niboshi:
Takes over #3023

In some environment, `nvcc` is not necessarilly looked up by`PATH`.

### DIFF
--- a/cupy/cuda/_environment.py
+++ b/cupy/cuda/_environment.py
@@ -1,0 +1,56 @@
+import os
+import shutil
+
+
+_cuda_path = ()
+_nvcc_path = ()
+
+
+def get_cuda_path():
+    # Returns the CUDA installation path or None if not found.
+    global _cuda_path
+    if _cuda_path == ():
+        _cuda_path = _get_cuda_path()
+    return _cuda_path
+
+
+def get_nvcc_path():
+    # Returns the path to the nvcc command or None if not found.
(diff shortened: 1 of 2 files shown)


niboshi:
I used the logic of the functions with the same name in `install/build.py` but rewrote entirely.

cupy/cuda/_environment.py
```diff
@@ -0,0 +1,56 @@
+import os
+import shutil
+
+
+_cuda_path = ()
+_nvcc_path = ()
+
+
+def get_cuda_path():
```
asi1024: How about naming like as `get_cuda_path_with_cache`?

niboshi: I think caching is an internal behavior and it's not important for the caller.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

leofang:
Hi @niboshi, could you briefly comment what's the difference between `cupy/cuda/_environment.py` and `cupy/_environment.py`? Both files have `get_cuda_path()`.

niboshi:
Oh, I didn't notice there's one in `cupy/_environment.py`.
I think they better fit `cupy/cuda/_environment.py` because they're related to CUDA environment (especially considering the possibility to make them public in the future).

PR #3030, opened 2020-02-05
Enable CUB support in CI

emcastillo:
Create a separate configuration for pfnCI.

There might be problems with `git clone --recursive` once #2584 is merged.

Since some routines have cub implementations and these are not explicitly tested in #2598, this CI config will run all the tests with cub enabled while the main tests will be run with cub disabled to test both paths.

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -167,3 +167,25 @@ configs {
     command: "bash .pfnci/script.sh py37.amd"
   }
 }
+
+# Use CUB
+configs {
+  key: "cupy.py37.cub"
+  value {
+    requirement {
+      cpu: 8
+      memory: 24
+      disk: 10
+      gpu: 2
+    }
+    time_limit {
+      seconds: 1800
+    }
+    checkout_strategy {
+      include_dot_git: true
+    }
+    environment_variables { key: "GPU" value: "2",
+                            key: "CUB_DISABLED" value:"0"}
+    command: "bash .pfnci/script.sh py37"
+  }
+}
--- a/.pfnci/run.sh
+++ b/.pfnci/run.sh
@@ -77,7 +77,10 @@ test_py37() {
   # TODO(niboshi): Allow option pass-through (https://github.com/chainer/xpytest/issues/14)
   export PYTEST_ADDOPTS=-rfEX
   # TODO(imos): Enable xpytest to support python_files setting in setup.cfg.
-  OMP_NUM_THREADS=1 xpytest "${xpytest_args[@]}" \
+  # OMP_NUM_THREADS=1 xpytest "${xpytest_args[@]}" \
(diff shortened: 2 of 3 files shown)


emcastillo:
flexCI, test this please

emcastillo:
flexCI test this please

niboshi:
Do we need to create a new project in https://ci.preferred.jp/ ?

emcastillo:
I created one, but doesn't seem to link it 😅

.pfnci/config.pbtxt
```diff
@@ -167,3 +167,25 @@ configs {
...
+      gpu: 2
+    }
+    time_limit {
+      seconds: 1800
+    }
+    checkout_strategy {
+      include_dot_git: true
+    }
+    environment_variables { key: "GPU" value: "2",
+                            key: "CUB_DISABLED" value:"0"}
```
leofang: You probably already knew: `CUB_DISABLED` will have no effect once #2584 is merged, so this key can be removed afterwards. (Keeping it here has no harm though.)

emcastillo: I actually think is better to keep it around, because it will simplify CI.

niboshi:
For extenals, this is the new CI project being tested (GitHub integration is not configured yet).
https://ci.preferred.jp/cupy.py37.cub/

It seems the job failed but there's no information about the error...

niboshi:
@emcastillo 

https://github.com/cupy/cupy/blob/71307a56759f83986093baafdbe1e5ea9391eef9/.pfnci/script.sh#L24

Can you change this line to `set -eux` for inspection purpose?

emcastillo:
Added and launched a new job

niboshi:
Doesn't seem to work :/

emcastillo:
Replaced xpytest with pytest just to see what happens

.pfnci/run.sh
```diff
@@ -77,7 +77,10 @@ test_py37() {
   # TODO(niboshi): Allow option pass-through (https://github.com/chainer/xpytest/issues/14)
   export PYTEST_ADDOPTS=-rfEX
   # TODO(imos): Enable xpytest to support python_files setting in setup.cfg.
-  OMP_NUM_THREADS=1 xpytest "${xpytest_args[@]}" \
+  # OMP_NUM_THREADS=1 xpytest "${xpytest_args[@]}" \
+  #     '/cupy/tests/**/test_*.py' \
+  #     && :
+  python -m pytest -m 'not slow' \
```
niboshi:
```
/src/.pfnci/run.sh: line 83: python: command not found
```
Maybe you should use `python3.7`?

emcastillo:
This PR is taken over by https://github.com/chainer/chainer-test/pull/561
due to the issues with cupy in pfnCI.

PR #3014, opened 2020-01-31
[backport] Use `size_t nbytes` in `__cuda_array_interface__`

jakirkham:
Backport of https://github.com/cupy/cupy/pull/3009

PR #3034, opened 2020-02-05
Fix: n-dimensional FFTs must preserve array contiguity when copying a view

### LINKED ISSUE #3033: bug: cannot apply pre-planned n-dimensional FFTs to Fortran-ordered views
If `cupyx.scipy.fftpack.get_fft_plan` is used to create a plan from a Fortran-ordered array and this plan is later applied to a Fortran-ordered **view**, the result will either fail with an unexpected shape or potentially give incorrect results.

This is due to an internal copy of the view that does not respected contiguity (PR forthcoming).

The issue does not apply to 1D transforms, only to those use PlanNd via `_exec_fftn`.

**Conditions**

CuPy version: 7.0.0
CUDA Build Version    : 10010
CUDA Driver Version   : 10010
CUDA Runtime Version  : 10010
cuBLAS Version        : 10201
cuFFT Version         : 10101

**Reproduce**

The following code illustrates the problem. The issue is caused by an internal copy in `_exec_fftn` which does not respect array contiguity.

```Python
import cupy
from cupy import testing
import cupyx.scipy.fftpack

shape = (32, 16, 4)
dtype = cupy.complex64
data_order = 'F'

a = testing.shaped_random(shape, cupy, dtype)
if data_order == 'F':
    a = cupy.asfortranarray(a)
    sl = [Ellipsis, 0]
else:
    sl = [0, Ellipsis]

# transform a contiguous view without pre-planning
view = a[sl]
expected = cupyx.scipy.fftpack.fftn(view)

# create plan and then apply it to a contiguous view
plan = cupyx.scipy.fftpack.get_fft_plan(view)
with plan:
    out = cupyx.scipy.fftpack.fftn(view)
testing.assert_allclose(expected, out)
```

The above should work, but instead fails with:

`ValueError: The CUFFT plan and a.shape do not match: plan.shape = (16, 32), expected_shape=(32, 16), a.shape = (32, 16)`

### LINKED ISSUE #3079: `overwrite_x=True` in cupyx.scipy.fft* doesn't do in-place transform for contiguous views
Reproducer:
```python
>>> import cupy as cp
>>> import cupyx.scipy.fftpack as cufft  # can also use cupyx.scipy.fft
>>> a = cp.random.random((64, 128, 128))+1j*cp.random.random((64, 128, 128))
>>> a.data.ptr
140130136358912
>>> b = a[:16, ...]  # create a C-contiguous view
>>> b.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
>>> b.data.ptr  # this is the same as a's
140130136358912
>>> b = cufft.fftn(b, overwrite_x=True)
>>> b.data.ptr  # expected to unchanged
140130041987072
```
This behavior is not expected, since the FFT of a contiguous view should (and can) be done in-place. (BTW, it doesn't matter whether a precomputed plan is used or not.) The underlying reason for the unexpected behavior is, curiously, the same for #3033: 
> This is due to an internal copy of the view that does not respected contiguity

This behavior is also against what we suggested in the doc (in a loose sense):
https://github.com/cupy/cupy/blob/016e4ce34723d44a7af613a76acf0baba53b6d3d/docs/source/reference/fftpack.rst#L36 
So we fix either the doc or the code. My preference is the latter, as it doesn't seem to be hard at first glance.

grlee77:
closes #3033
closes #3079

A concrete example of the issue is given in #3033. This PR fixes that use case, adding a corresponding test case over both C and Fortran ordered views.

Please also backport this fix to the ~~7.x and 6.x branches.~~ 7.x branch.

### DIFF
--- a/cupy/fft/fft.py
+++ b/cupy/fft/fft.py
@@ -356,7 +356,7 @@ def _exec_fftn(a, direction, value_type, norm, axes, overwrite_x,
     if fft_type not in [cufft.CUFFT_C2C, cufft.CUFFT_Z2Z]:
         raise NotImplementedError('Only C2C and Z2Z are supported.')
 
-    if a.base is not None:
+    if a.base is not None and not a.flags.forc:
         a = a.copy(order='A')
 
     if a.flags.c_contiguous:
--- a/cupy/fft/fft.py
+++ b/cupy/fft/fft.py
@@ -356,9 +356,6 @@ def _exec_fftn(a, direction, value_type, norm, axes, overwrite_x,
     if fft_type not in [cufft.CUFFT_C2C, cufft.CUFFT_Z2Z]:
         raise NotImplementedError('Only C2C and Z2Z are supported.')
 
-    if a.base is not None and not a.flags.forc:
-        a = a.copy(order='A')
-
     if a.flags.c_contiguous:
         order = 'C'
     elif a.flags.f_contiguous:
--- a/cupy/fft/fft.py
+++ b/cupy/fft/fft.py
@@ -357,7 +357,7 @@ def _exec_fftn(a, direction, value_type, norm, axes, overwrite_x,
         raise NotImplementedError('Only C2C and Z2Z are supported.')
 
     if a.base is not None:
-        a = a.copy()
+        a = a.copy(order='A')
 
     if a.flags.c_contiguous:
         order = 'C'
--- a/tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
+++ b/tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
@@ -606,3 +606,18 @@ class TestFftnView(unittest.TestCase):
         with plan:
             out = cupyx.scipy.fftpack.fftn(view)
         testing.assert_allclose(expected, out)
+
+    @testing.for_complex_dtypes()
+    def test_overwrite_x_with_contiguous_view(self, dtype):
+        # Test case for: https://github.com/cupy/cupy/issues/3079
+        a = testing.shaped_random(self.shape, cupy, dtype)
+        if self.data_order == 'C':
+            # C-contiguous view
+            b = a[:a.shape[0] // 2, ...]
+        else:
+            # F-contiguous view
+            a = cupy.asfortranarray(a)
+            b = a[..., :a.shape[-1] // 2]
+        b_ptr = b.data.ptr
+        out = cupyx.scipy.fftpack.fftn(b, overwrite_x=True)
+        assert out.data.ptr == b_ptr
--- a/tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
+++ b/tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
@@ -588,8 +588,8 @@ class TestRfft(unittest.TestCase):
 class TestFftnView(unittest.TestCase):
 
     @testing.for_complex_dtypes()
-    def test_fortran_ordered_view(self, dtype):
-        # test case for: https://github.com/cupy/cupy/issues/3033
+    def test_contiguous_view(self, dtype):
+        # Fortran-ordered case tests: https://github.com/cupy/cupy/issues/3079
         a = testing.shaped_random(self.shape, cupy, dtype)
         if self.data_order == 'F':
             a = cupy.asfortranarray(a)
@@ -603,6 +603,25 @@ class TestFftnView(unittest.TestCase):
 
         # create plan and then apply it to a contiguous view
         plan = cupyx.scipy.fftpack.get_fft_plan(view)
+        with plan:
+            out = cupyx.scipy.fftpack.fftn(view)
+        testing.assert_allclose(expected, out)
+
+    @testing.for_complex_dtypes()
+    def test_noncontiguous_view(self, dtype):
+        a = testing.shaped_random(self.shape, cupy, dtype)
+        if self.data_order == 'F':
+            a = cupy.asfortranarray(a)
+            sl = [Ellipsis, slice(None, None, 2)]
+        else:
+            sl = [slice(None, None, 2), Ellipsis]
+
+        # transform a non-contiguous view without pre-planning
+        view = a[sl]
+        expected = cupyx.scipy.fftpack.fftn(view)
+
+        # create plan and then apply it to a non-contiguous view
+        plan = cupyx.scipy.fftpack.get_fft_plan(view.copy())
         with plan:
             out = cupyx.scipy.fftpack.fftn(view)
         testing.assert_allclose(expected, out)
--- a/tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
+++ b/tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
@@ -1,6 +1,7 @@
 import unittest
 import pytest
 
+import cupy
 from cupy import testing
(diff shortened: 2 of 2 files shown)


leofang:
Hi Gregory,

> Please also backport this fix to the 7.x and 6.x branches.

I think they [stop making new v6 releases](https://github.com/cupy/cupy/pull/3014#issuecomment-580652425). However, if you and your collaborators install CuPy from conda-forge, we can apply the patch to the v6 feedstock for you.

cupy/fft/fft.py
```diff
@@ -357,7 +357,7 @@ def _exec_fftn(a, direction, value_type, norm, axes, overwrite_x,
         raise NotImplementedError('Only C2C and Z2Z are supported.')
 
     if a.base is not None:
-        a = a.copy()
+        a = a.copy(order='A')
```
leofang: This is a neat fix! 👍

tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
```diff
@@ -578,3 +579,30 @@ def test_irfft_overwrite(self, xp, scp, dtype):
         x = testing.shaped_random(self.shape, xp, dtype)
         return scp.fftpack.irfft(x, n=self.n, axis=self.axis,
                                  overwrite_x=True)
+
+
+@testing.parameterize(
+    {'shape': (32, 16, 4), 'data_order': 'F'},
+    {'shape': (4, 32, 16), 'data_order': 'C'},
+)
+class TestFftnView(unittest.TestCase):
```
leofang: I suppose the test is added in this file because it's for where `get_fft_plan()` resides?

grlee77: yes. I started out with it in the core cupy fft tests, but moved it here since fftpack contained the planning functionality.

grlee77:
> I think they stop making new v6 releases. However, if you and your collaborators install CuPy from conda-forge, we can apply the patch to the v6 feedstock for you.

Oh, okay. Just 7.x is fine for my purposes.

cupy/fft/fft.py
```diff
@@ -357,7 +357,7 @@ def _exec_fftn(a, direction, value_type, norm, axes, overwrite_x,
         raise NotImplementedError('Only C2C and Z2Z are supported.')
 
     if a.base is not None:
```
leofang: I confirmed this patch can fix both #3033 and #3079 and also pass all tests:
```suggestion
    if a.base is not None \
       and not a.flags.c_contiguous and not a.flags.f_contiguous:
```
Should we do this @grlee77?

grlee77: Yeah, I think so. By the way, there is also `a.flags.forc` which should be True if it is either C or Fortan contiguous.

grlee77:
```Python
if a.base is not None and not a.flags.forc:
```

leofang: Oh that's better! I didn't notice this flag...

grlee77: This change has now been made along with a test case adapted from your example in #3079

grlee77:
ping.

I would like to get this fix merged, as I have a downstream project currently requiring a workaround for the issue.

leofang:
cc @toslunar @emcastillo

toslunar:
I'll review the PR soon.

tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py
```diff
@@ -578,3 +579,45 @@ def test_irfft_overwrite(self, xp, scp, dtype):
...
+
+    @testing.for_complex_dtypes()
+    def test_fortran_ordered_view(self, dtype):
```
toslunar:
```suggestion
    def test_contiguous_view(self, dtype):
```
And mention that the F-contiguous case was the issue.

cupy/fft/fft.py
```diff
@@ -356,8 +356,8 @@ def _exec_fftn(a, direction, value_type, norm, axes, overwrite_x,
     if fft_type not in [cufft.CUFFT_C2C, cufft.CUFFT_Z2Z]:
         raise NotImplementedError('Only C2C and Z2Z are supported.')
 
-    if a.base is not None:
-        a = a.copy()
+    if a.base is not None and not a.flags.forc:
+        a = a.copy(order='A')
```
toslunar: The change to the line is essential because the default `order='K'` may leave `a` non-contiguous.

(Optional) Could you also test non-contiguous view to detect it?

grlee77: I added a `test_noncontiguous_view`. Please see if that addresses the case you were thinking of

grlee77:
Thanks for reviewing @toslunar. Please see the latest commit and run CI if it looks good to you

toslunar:
Jenkins, test this please.

cupy/fft/fft.py
toslunar: It seems `a.copy(...)` is just verbose.

- Only `_fftn` calls `_exec_fftn` and C- or F-contiguous `a` seems passed.
- In my environment all the tests pass without these two lines.

grlee77: > In my environment all the tests pass without these two lines.

You mean ALL of the fftpack tests or just the newly added tests? For me the newly added tests pass without those two lines, but many other fftpack tests fail.

Specifically, running `pytest tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py` gives `152 failed, 653 passed `. Similarly for the `cupy.fft` and `cupyx.scipy.fft` tests.

All cases pass for me with those two lines present.

toslunar: In my environment below, `pytest tests/cupyx_tests/scipy_tests/fftpack_tests/test_fftpack.py` passed without the lines on `pull/3034/merge`.  With `assert a.flags.forc`, the result was the same .

CUDA 9.0 may be too old.  With `assert False` here, the run gave `9 failed, 796 passed`.  Only 3 existing tests (in `TestFft2_param_0_{axes=None, s=None, shape=(3, 4)}`) failed.

```
CuPy Version          : 8.0.0b1
CUDA Build Version    : 9000
CUDA Driver Version   : 9000
CUDA Runtime Version  : 9000
cuBLAS Version        : 9000
cuFFT Version         : 9000
cuRAND Version        : 9000
cuSOLVER Version      : (9, 0, 0)
cuSPARSE Version      : 9000
NVRTC Version         : (9, 0)
cuDNN Build Version   : 7300
cuDNN Version         : 7300
NCCL Build Version    : 2005
```

toslunar: I reproduced `152 failed, 653 passed` after removing `a = a.copy()` in **`_exec_fft`**, where the copy is necessary (possibly because of the late `swapaxes` in the current implementation).

grlee77: Okay, I did a clean rebuild of the branch and now all tests are passing without those two lines. Apologies for the confusion!

Indeed `_fftn` always enforces contiguity so it does not necessarily need to also be checked with `_exec_fftn`.

I now get all fft-related tests passing (aside from a few known irfftn failures due to a CUFFT bug in CUDA 10.2).

toslunar:
Jenkins, test this please.

cupy/fft/fft.py
leofang: Oh sorry I failed to recognize this when reviewing! 22cede9 is a much better fix.

PR #3036, opened 2020-02-06
fix behavior of result_type when 0-dimensional arrays are involved

### LINKED ISSUE #3035: subtle difference in behavior of cupy.result_type vs numpy.result_type
`cupy.result_type` seems to handle zero-dimensional arrays differently than numpy.

```Python

import numpy 
import cupy

numpy.result_type(numpy.asarray([10, 20], dtype=numpy.float32), numpy.asarray(0.0))
# gives: dtype('float32')

cupy.result_type(cupy.asarray([10, 20], dtype=cupy.float32), cupy.asarray(0.0))
# gives dtype('float64')
```

If the 0-dimensional array is replaced by a scalar then the result matches numpy
```Python
cupy.result_type(cupy.asarray([10, 20], dtype=cupy.float32), 0.0)
# gives: dtype('float32')
```

** Conditions:**

CuPy Version          : 7.0.0
CUDA Build Version    : 10010
CUDA Driver Version   : 10010
CUDA Runtime Version  : 10010
cuBLAS Version        : 10201
cuFFT Version         : 10101
cuRAND Version        : 10101
cuSPARSE Version      : 10300
cuDNN Build Version   : 7402
cuDNN Version         : 7402
NCCL Build Version    : None
NCCL Runtime Version  : None

grlee77:
closes #3035

This fixes the different behaviour for 0-dimensional arrays in `result_type` by generating 0-dimensional NumPy arrays of the corresponding `dtype` instead of passing the `dtype` directly.

The `'zero_dim'` test cases added in 8845706 fail on current master, but should pass with the change introduced in cadf203.

(please also backport to the 7.x branch)

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -389,8 +389,17 @@ def result_type(*arrays_and_dtypes):
 
     .. seealso:: :func:`numpy.result_type`
     """
-    dtypes = [a.dtype if isinstance(a, cupy.ndarray)
-              else a for a in arrays_and_dtypes]
+    dtypes = []
+    for a in arrays_and_dtypes:
+        if isinstance(a, cupy.ndarray):
+            if a.ndim == 0:
+                # NumPy handles 0-dimensional arrays differently
+                dtypes.append(numpy.asarray(0, dtype=a.dtype))
+            else:
+                dtypes.append(a.dtype)
+        else:
+            dtypes.append(a)
+
     return numpy.result_type(*dtypes)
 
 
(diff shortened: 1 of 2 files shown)


niboshi:
@grlee77 
CC: @asi1024 

Sorry, this is a known incompatibility documented here: https://docs-cupy.chainer.org/en/stable/reference/difference.html#type-promotion
This won't be fixed, as it's impossible to implement NumPy's behavior in integer case (see the example indicated in the documentation).

grlee77:
Ah, I see that the behavior is a bit more subtle than accounted for here in that it also depends on the value. Yeah, I guess in that case we would have to have device->host transfer here, which does not seem reasonable.

Thanks for pointing out the example. I will close this and the corresponding issue.

PR #3041, opened 2020-02-06
Some fixes to `.pfnci/script.sh`

niboshi:
- Update v6 to v7
- Output error message in get_base_branch()
- Remove 'run' in get_base_branch() as it should not be skipped


asi1024:
LGTM.

PR #3048, opened 2020-02-07
Require `scipy` in `test_gmm`

toslunar:
This amends #2996 for test environments without `scipy` (but with `matplotlib`).


emcastillo:
Jenkins, test this please

PR #3049, opened 2020-02-07
Add mpi4py examples

leofang:
Refs:
1. https://docs-cupy.chainer.org/en/latest/reference/interoperability.html#mpi4py
2. https://mpi4py.readthedocs.io/en/latest/tutorial.html#cuda-aware-mpi-python-gpu-arrays 

I didn't add tests for the examples, because it'd require the CI to build CUDA-aware MPI and mpi4py, which would be too much work.

~~I think `single_device.py` could be removed as it is almost a duplicate of `multiple_devices.py` except for the device selection. Let me know how you think.~~

Thanks.


examples/cuda_aware_mpi/multiple_devices.py
```diff
@@ -0,0 +1,43 @@
...
+
+comm = MPI.COMM_WORLD
+rank = comm.Get_rank()
+size = comm.Get_size()
+if size != 2:
+    raise RuntimeError("run this script with 2 processes: mpiexec -n 2 ...")
+
+# select device based on (local) MPI rank
+cupy.cuda.Device(rank).use()
+
```
emcastillo: I have my doubts this will work, if you launch 128 processes, the rank is between 0-128
you need to calculate a local rank which is obtained with either env variables if you pass the corresponding options to mpirun
``` python
# OpenMPI:
local_rank = os.getenv("OMPI_COMM_WORLD_LOCAL_RANK"));

# MVPAICH2
local_rank = os.getenv("MV2_COMM_WORLD_LOCAL_RANK"));
```
Another option is to tweak process placement in mpirun and just do `local_rank = rank % num_devices`

emcastillo:
You can just delete single device. is a sub-case

examples/cuda_aware_mpi/multiple_devices.py
leofang: For general applicability, you're absolutely right. In fact, this is an awfully annoying issue I'd rather not dive into, so I limited the total pool size to `2`, and assumed all processes are launched on the same node. 

No all MPI implementations provide such env variables, and AFAIK in the past MPI codes on CPU must determine their own local ranks in some way if MPI-library-agnostic is needed. Open MPI and MVAPICH2 happen to 1. provide such env variables, and 2. implement CUDA-awareness earlier than their peers. But with other MPI libraries catching up (like Spectrum MPI), I don't think there is any guarantee...So your option 1 would limit the choices of MPI libraries (I don't think it's a bad thing, though). 

As for your option 2, `local_rank = rank % num_devices` will not work if your 128 processes are scattered into multiple nodes, not to mention the possibility that each node could have a variable number of GPUs. It requires some effort to get the process-node-device(-core-thread-whatnot) mapping right.

Another even more awful issue is that the initialization order of MPI and CUDA context also matters, occasionally leading to puzzling bugs. It's implementation dependent, and I don't think we can cover all these in a few short examples...

emcastillo: I know 😁.
MPI+multi device is horrible, usually the user is responsible to deal with it using placement commands in mpirun,  SLURM or the appropriate batch system.

There are options to place consecutive ranks in the same node (thats why the % device_count is generally used in this setting). But I don't think we should support all the possible use cases for such an example. Just limiting it to OpenMPI or MVAPICH covers the majority of use cases.

leofang: I thought about it overnight, and I think it's better to just leave a comment on obtaining local ranks using the env variables. 

The reason is that it's too implementation-specific and that it requires some knowledge to get CUDA-aware MPI work. The burden shouldn't be on us. In addition, even MVAPICH2 may not be CUDA-aware by default. One either needs to build from source with `--enable-cuda`, or installs MVAPICH2-GDR. ([I've requested the MVAPICH2 core devs to provide a runtime check for CUDA-awareness for a long time](http://mailman.cse.ohio-state.edu/pipermail/mvapich-discuss/2020-January/006962.html), but they never reply...😞) 

In any case, there are substantial prerequisites to set up the environment, and providing educational information here is kinda overkill and redundant. (We don't even do that in mpi4py's website.)

examples/cuda_aware_mpi/multiple_devices.py
```diff
@@ -14,6 +14,9 @@
 size = comm.Get_size()
 if size != 2:
     raise RuntimeError("run this script with 2 processes: mpiexec -n 2 ...")
+device_count = cupy.cuda.runtime.getDeviceCount()
+if size > device_count:
+    raise RuntimeError("need {} GPUs, found {}".format(size, device_count))
```
emcastillo: Error message is not clear.
Lets say something like
"This example only works with a number of processes equal to the number of GPUs in the local rank"

leofang: Can we just say "this script requires 2 GPUs"?

leofang:
Another question: Should I rename the path to something like `examples/interoperability/mpi4py_multiple_devices.py`? This way in the future we could add to here any files involving 3rd party libraries.

examples/cuda_aware_mpi/multiple_devices.py
emcastillo: ok

leofang:
Done. PTAL, thanks.

examples/interoperability/mpi4py_multiple_devices.py
```diff
@@ -0,0 +1,53 @@
...
+from mpi4py import MPI
+
+
+comm = MPI.COMM_WORLD
+rank = comm.Get_rank()
+size = comm.Get_size()
+if size != 2:
+    raise RuntimeError("run this script with 2 processes: mpiexec -n 2 ...")
+device_count = cupy.cuda.runtime.getDeviceCount()
+if device_count < 2:
```
emcastillo:
```suggestion
if device_count != 2:
```

emcastillo:
Jenkins, test this please

leofang:
Thanks, @emcastillo 🙂

PR #3052, opened 2020-02-07
Update installation guide for conda-forge

### LINKED ISSUE #3006: Update installation guide for conda-forge
@jakirkham: Does this part still make sense for installing and running cupy from conda-forge?  

https://github.com/cupy/cupy/blob/0876ef72c01c16ed4906a110eb87c88975da28c9/docs/source/install.rst#L327-L341

leofang:
Close #3006. 

@jakirkham, two quick questions for you: Do you know who manages [the `cupy` package in the `rapidsai` channel](https://anaconda.org/rapidsai/cupy), and whether `cupy` from `conda-forge` is compatible with other GPU packages from `rapidsai`? Thanks.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -117,6 +117,12 @@ the installation of a particular CUDA version (say 10.0) for driver compatibilit
 
     Currently cuTENSOR is not yet available on ``conda-forge``.
 
+.. note::
+
+    If you encounter any problem with CuPy from ``conda-forge``, please feel free to report to `cupy-feedstock 
+    <https://github.com/conda-forge/cupy-feedstock/issues>`_, and we will help investigate if it is just a packaging
+    issue in ``conda-forge``'s recipe or a real issue in CuPy.
+
 
 Install CuPy from Source
 ------------------------
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -98,6 +98,26 @@ Please make sure that only one CuPy package (``cupy`` or ``cupy-cudaXX`` where X
   $ pip freeze | grep cupy
 
 
+Install CuPy from conda-forge
+-----------------------------
+
+Conda/Anaconda is a cross-platform package management solution widely used in scientific computing and other fields.
+The above ``pip install`` instruction is compatible with ``conda`` environments. Alternatively, for Linux 64 systems
+once the CUDA driver is correctly set up, you can install CuPy from the ``conda-forge`` channel::
+
+    $ conda install -c conda-forge cupy
+
+and ``conda`` will install pre-built CuPy and most of the optional dependencies for you, including CUDA runtime libraries
+(``cudatoolkit``), NCCL, and cuDNN. It is not necessary to install CUDA Toolkit in advance. If you need to enforce
+the installation of a particular CUDA version (say 10.0) for driver compatibility, you can do::
+
+    $ conda install -c conda-forge cupy cudatoolkit=10.0
+
+.. note::
+
+    Currently cuTENSOR is not yet available on ``conda-forge``.
+
+
 Install CuPy from Source
 ------------------------
 
@@ -165,6 +185,10 @@ Use pip to uninstall CuPy::
 .. note::
 
    If you are using a wheel, ``cupy`` shall be replaced with ``cupy-cudaXX`` (where XX is a CUDA version number).
+
+.. note::
+
+   If CuPy is installed via ``conda``, please do ``conda uninstall cupy`` instead.
 
 
 Upgrade CuPy
@@ -242,7 +266,7 @@ If you need to pass environment variable (e.g., ``CUDA_PATH``), you need to spec
 
 If you are using certain versions of conda, it may fail to build CuPy with error ``g++: error: unrecognized command line option ‘-R’``.
 This is due to a bug in conda (see `conda/conda#6030 <https://github.com/conda/conda/issues/6030>`_ for details).
-If you encounter this problem, please downgrade or upgrade it.
+If you encounter this problem, please upgrade your conda.
 
 .. _install_cudnn:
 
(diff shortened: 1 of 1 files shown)


jakirkham:
This looks great Leo! Thanks for writing this up 😄

Should we add something about where the recipe lives and/or raising issues about the Conda package on the feedstock?

RAPIDS added the package before it was added to `conda-forge`. Though it uses an older version of the same recipe. FWIW RAPIDS packages (like `cudf` and `cuml`) are requiring a new enough version of `cupy` that they just get the package from `conda-forge` and have been doing so for a while. Does that help? Please let me know if you have any more questions 🙂

leofang:
Thanks, John!
 
> Should we add something about where the recipe lives and/or raising issues about the Conda package on the feedstock?

Sure, PTAL at  6f56b41.
 
> Does that help? Please let me know if you have any more questions 🙂

Yes, very helpful! And yes, I have one more question: Out of curiosity, I was reading CF meeting minutes and [found this](https://conda-forge.org/docs/minutes/2020-01-22.html#cuda): 

> Question: are we ready to advertise this? If so should we Tweet and do a NumFOCUS update about it?

So, are we ready to advertise this? 🤣

docs/source/install.rst
```diff
@@ -98,6 +98,32 @@ Please make sure that only one CuPy package (``cupy`` or ``cupy-cudaXX`` where X
...
+
+    Currently cuTENSOR is not yet available on ``conda-forge``.
+
+.. note::
+
+    If you encounter any problem with CuPy from ``conda-forge``, please feel free to report to `cupy-feedstock 
+    <https://github.com/conda-forge/cupy-feedstock/issues>`_, and we will help investigate if it is just a packaging
+    issue in ``conda-forge``'s recipe or a real issue in CuPy.
+
+
```
leofang: @jakirkham, another question: Do we want to comment on `nvcc`'s availability? Something along this line:

"If you did not install CUDA Toolkit yourselves, the ``nvcc`` compiler might not be available. The ``cudatoolkit`` package does not have ``nvcc`` included."

leofang:
btw, this PR should be backported to v7, as we started packaging CuPy before the EOF of v6.

leofang:
Ping @jakirkham.

docs/source/install.rst
jakirkham: Sure we could do that. It might also make sense to include this info in conda-forge and link out to it. WDYT? 🙂

leofang:
```suggestion
.. note::

    If you did not install CUDA Toolkit yourselves, the ``nvcc`` compiler might not be available. 
    The ``cudatoolkit`` package from Anaconda does not have ``nvcc`` included.

```

leofang:
Thanks @jakirkham! 

> We could add a note on the compiler if that makes sense.

OK I just added the note. The reason I was asking is that `RawKernel` and `RawModule` support swapping the backend between `nvrtc` (default) and `nvcc`. I am not sure how likely this note might help when conda users encounter this rare situation (`nvcc` not existing), but it doesn't hurt to mention. 

> Also would be ok leaving this to the conda-forge docs to expand on. 

Is the whole CUDA toolchain documented now? Last time I checked there's nothing...We could remove this quick note once conda-forge has everything documented.

jakirkham:
Ah yeah that makes sense. Agree it should be documented somehow then. 🙂

Not yet no. Am planning to do this soon though 😉

leofang:
> Not yet no. Am planning to do this soon though 😉

Sounds good. Ping me if you have a PR for that, John!

leofang:
Is it possible to get this doc update merged? It seems to confuse some users already: https://stackoverflow.com/q/60534075/2344149

michaelklachko:
So, to clarify this change, let's say I have system-wide CUDA installed on my Ubuntu server, and I just installed pytorch with cuda, which installed cudatoolkit. If I want to install CuPy and to have it work nicely with Pytorch (so that I can pass arrays from Pytorch to CuPy and back), I will do:
1. `conda install cupy`
2. `conda install nvcc`

Questions:
Will it use existing conda cudatoolkit installed by Pytorch, or system-wide CUDA? 
What about CuDNN, which is already present in system-wide CUDA, and also separately installed by Pytorch, but not present in `conda list`? Should I do `conda install cudnn` or not?

leofang:
First, you don't have to install `nvcc` if you don't want to. Second, please install CuPy from the conda-forge channel as instructed; the anaconda channel is very outdated.

If your pytorch is installed via conda, all of its dependencies are managed by conda. (More on this below.) This means, in particular, that all linkings are done with conda's libraries, so in this case the system CUDA is not looked up, and conda's `cudatoolkit` will be used. 

(The only exception for this is when CUDA is looked up at runtime via dlopen or ctypes. Numba and Open MPI are two such "notorious" examples. In such cases, it's possible (depending on the search logic) that the system CUDA is used.)

Finally, if your pytorch is from the anaconda channel, cuDNN is also managed by conda and thus visible to conda-forge's CuPy. (So when you do `conda install pytorch` you should see a separate `cudnn` being installed.) But, it seems if you install pytorch from the pytorch channel, cuDNN is bundled with it. In that case, when installing CuPy it will install another copy of cudnn. I didn't test, but the linkage should protect you even in such a case. The most important thing is that they still see the same `cudatoolkit`.

@jakirkham Am I making sense?

michaelklachko:
I see, thanks. Also, the link to cuTensor is broken, it should be https://developer.nvidia.com/cutensor

Should I install cuTensor is I want to perform unpackbits, bitwise shifts, boolean masking, and sum of masked elements on tensors? What do I need to make these operations as fast as possible?

leofang:
Thanks, Michael. Looks like the link was not updated when v1.0 was mentioned in #2948. Will update it shortly.

As for your question, I am not familiar with cuTENSOR but I doubt it'd help. It's mainly for element-wise, reduction, and contraction operations. Perhaps other CuPy functionalities like fancy indexing could already meet your need (provided that you can do a similar job with NumPy)? In any case, it's out of scope of this PR, and I'd recommend you to open an issue so that others can find your question and help. 🙂

kmaehashi:
pfnCI, test this please.

leofang:
Thanks, @kmaehashi!

PR #3055, opened 2020-02-10
Remove chainercv CI configs

niboshi:


kmaehashi:
Update webhook configuration:
* `cupy.wheel.nightly` will be called only when pushing to v7 branch.
* `cupy.py2.cv.gpu` and `cupy.py2.cv.examples` will never be called anymore.

`cupy.py3.cv.gpu` and `cupy.py3.cv.examples` webhooks cannot be removed as we still need to test v7 branch with ChainerCV. Fix in #3411.

PR #3060, opened 2020-02-10
add cupy.ndim

grlee77:
This is a pretty trivial function, but was missing from the CuPy API.  Please advise on whether this is wanted and if it should be in the Cython code as done here or if we should just define it in pure Python elsewhere.

Note that I left the determination for objects without an ndim attribute up to `numpy.asarray(a).ndim` rather than `cupy.asarray(a).ndim` to avoid a host->device transfer and make sure behavior is equivalent to NumPy for objects.  e.g. `assert cupy.ndim(None) == 0`

Adding this also makes it possible for `numpy.ndim` to return the dimension of a CuPy array via the `__array_function__` protocol. (e.g. `assert np.ndim(cupy.ones((16, 16))) == 2`)

Let me know if this looks likely to be accepted and I will add some test cases.

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -72,6 +72,7 @@ from cupy import lib  # NOQA
 
 # import class and function
 from cupy.core import ndarray  # NOQA
+from cupy.core import ndim  # NOQA
 from cupy.core import ufunc  # NOQA
 
 
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -697,10 +697,6 @@ def ndim(a):
     Returns:
         (int): The number of dimensions in `a`.
 
-    Note:
-        This function is consistent with Numpy when handling ndarrays, scalars
-        and simple iterables. It does not handle more complex general cases
-        such as iterables where individual elements are cupy.ndarrays.
     """
     try:
         return a.ndim
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -72,7 +72,6 @@ from cupy import lib  # NOQA
 
 # import class and function
 from cupy.core import ndarray  # NOQA
-from cupy.core import ndim  # NOQA
 from cupy.core import ufunc  # NOQA
 
 
@@ -686,6 +685,28 @@ from cupy.statistics.histogram import histogram  # NOQA
 # Undocumented functions
 # -----------------------------------------------------------------------------
 from cupy.core import size  # NOQA
+
+
+def ndim(a):
+    """Returns the number of dimensions of an array.
+
+    Args:
+        a (array-like): If it is not already an `cupy.ndarray`, a conversion
+            via :func:`numpy.asarray` is attempted.
+
+    Returns:
+        (int): The number of dimensions in `a`.
+
+    Note:
+        This function is consistent with Numpy when handling ndarrays, scalars
+        and simple iterables. It does not handle more complex general cases
+        such as iterables where individual elements are cupy.ndarrays.
+    """
+    try:
+        return a.ndim
+    except AttributeError:
+        return numpy.asarray(a).ndim
+
 
 # -----------------------------------------------------------------------------
 # CuPy specific functions
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -701,7 +701,7 @@ def ndim(a):
     try:
         return a.ndim
     except AttributeError:
-        return numpy.asarray(a).ndim
+        return numpy.ndim(a)
 
 
 # -----------------------------------------------------------------------------
--- a/tests/cupy_tests/test_ndim.py
+++ b/tests/cupy_tests/test_ndim.py
@@ -1,0 +1,79 @@
+import numpy
+import unittest
+
+import cupy
+from cupy import testing
+
+
+@testing.gpu
+class TestNdim(unittest.TestCase):
+
+    @testing.numpy_cupy_equal()
+    def test_ndim_ndarray1d(self, xp):
+        return xp.ndim(xp.arange(5))
+
+    @testing.numpy_cupy_equal()
+    def test_ndim_ndarray2d(self, xp):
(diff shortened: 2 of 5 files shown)


niboshi:
Thank you for the PR.
It looks `numpy.ndim` is not documented, but I think it's ok to have it in `cupy`, as it's necessary for `np.ndim` users as you pointed out.

As for where to implement it, I don't think it doesn't have to be written in Cython.
Could you move the implementation to `cupy/__init__.py`, and also write some tests?

Also, this function does not support a complicated data structure such as lists of `cupy.ndarray`s. I don't think we have to support such inputs at least in the first version (it would be quite complicated), but could you add some statement in the docstring to clarify that?

Thank you!

grlee77:
Okay, I will move it to pure python in `cupy/__init__.py`. Apparently this function is a relic of the old Numeric library. I did not realize it was undocumented, but as far as I can tell there has not been discussion of removing it from the public API. It was mentioned as an alternative to the recently removed `np.rank` in the [Numpy 1.18 release notes](https://numpy.org/devdocs/release/1.18.0-notes.html#expired-deprecations):

I don't think it is super widely used, but I have seen it used in a few limited places within both SciPy and Scikit-Image. It also occasionally used internally within NumPy itself.

grlee77:
> Also, this function does not support a complicated data structure such as lists of cupy.ndarray

Somewhat surprisingly, this case does seem to work: 

```Python
import cupy
import numpy

mixed_list = [cupy.arange(8), numpy.arange(8)]
assert cupy.ndim(mixed_list) == 2

obj_array = numpy.asarray(mixed_list)
print(f"shape = {obj_array.shape}")
print(obj_array)
```
gives
```
shape = (2, 8)
[[array(0) array(1) array(2) array(3) array(4) array(5) array(6) array(7)]
 [0 1 2 3 4 5 6 7]]
```
So the CuPy array got converted to an NumPy object array where each element is a 0-dimensional cupy array.

niboshi:
Oh, thank you for testing!
But I think it's not desirable to depend on that.

1. As you said, this creates as many `cupy.ndarray`s as elements in the original arrays. This overhead is significant for a simple feature like `ndim()`.
2. Maybe NumPy does not intend `ndim` to be caculated correctly for a `numpy.ndarray` with `cupy.ndarray`s as elements. Any subtle change to NumPy's internal code could potentially break the result of `cupy.ndim()`.

But it's not trivial to check and forbid such inputs, either...

niboshi:
I thought we don't have to support such inputs in the initial implementation, but maybe it's easier to support it rather than trying to forbid it?

grlee77:
> Oh, thank you for testing!
> But I think it's not desirable to depend on that.

In the last commit I removed the test cases that involved lists of cupy and/or numpy arrays so we are not locked into a specific behavior in that case. I am not really keen on trying to inspect and explicitly prevent such cases, though.

niboshi:
So you're not going to implement the real support either?
Then, please
* Remove fallback to `numpy.asarray`
* Revise the docstring to make it clear that this function expects only `cupy.ndarray`
* Perhaps add a TODO comment, too

cupy/__init__.py
```diff
@@ -686,6 +686,24 @@ def isscalar(element):
...
+            via :func:`numpy.asarray` is attempted.
+
+    Returns:
+        (int): The number of dimensions in `a`.
+
+    """
+    try:
+        return a.ndim
+    except AttributeError:
+        return numpy.asarray(a).ndim
```
leofang, line 704: How about this:
```suggestion
        return numpy.ndim(a)
```
This fallback will cover Python scalars and sequences as well.

leofang:
@kmaehashi @grlee77 Is it possible to revisit this PR? I think it's useful (see e.g. #3542).

cupy/__init__.py
grlee77: Yeah, I think that makes sense and covers most use cases.

Either approach still has issues with some types of input, though. For example, if the user were to input a list of large CuPy arrays as in the following:

```Python
np.ndim([cp.ones((1024, 1024))] * 4)
``` 

This takes several seconds to run!

I got a bit confused about what the desired behavior was in these edge cases and it stalled there.

leofang:
@kmaehashi @grlee77 For compatibility with NEP-18 I'd like to see this PR merged. Let us worry about performance and corner cases after the rc1 release, ok?

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

tests/cupy_tests/test_ndim.py
```diff
@@ -0,0 +1,67 @@
...
+@testing.gpu
+class TestNdim(unittest.TestCase):
+
+    @testing.numpy_cupy_equal()
+    def test_ndim_ndarray1d(self, xp):
+        return xp.ndim(xp.arange(5))
+
+    @testing.numpy_cupy_equal()
+    def test_ndim_ndarray2d(self, xp):
+        assert xp.ndim(xp.ones((2, 4))) == 2
```
leofang, line 17: I think with `numpy_cupy_equal` we should return instead of asserting. I can fix this later.

kmaehashi:
pfnCI, test this please.

kmaehashi:
We revisited this PR and agreed to add this. (c.f. `cupy.shape` https://github.com/cupy/cupy/pull/3229)
We can later enhance these functions to support list of cupy arrays; we already have a similar routine around here: https://github.com/cupy/cupy/blob/v8.0.0b5/cupy/core/core.pyx#L2069

leofang:
Thank you, @kmaehashi @grlee77! I will send a follow-up PR to fix the test style later.

tests/cupy_tests/test_ndim.py
leofang: See #3890.

Issue #1961, opened 2019-01-17
Support parallel compile in installing time


okuta:
Please read [this](https://stackoverflow.com/questions/38013863/how-to-perform-cython-files-compilation-in-parallel/40533012#40533012).

leofang:
It'd be great to have this ability. Building CuPy from source takes so long and it's not good for both developers and CI infrastructures.

kmaehashi:
It's worth trying, but the merit of parallelization may be limited by Thrust module.
In my environment, full build took 548.98s (thrust module = 373.36s, all the rest = 175.62s).

leofang:
Both the Thrust and CUB modules will take the majority of time due to template instantiation...I wonder if we can assign two independent processes to handle them, or even apply some tricks to speed it up.

kmaehashi:
I guess you are already aware of, but using `ccache` significantly improves build time.
https://docs-cupy.chainer.org/en/latest/contribution.html#how-to-run-tests

leofang:
Just saw cython/cython#3187. It seems starting Cython 0.29.14 it is possible to build Cython modules in parallel, but I have not tested this capability yet.

leofang:
> Just saw [cython/cython#3187](https://github.com/cython/cython/pull/3187). It seems starting Cython 0.29.14 it is possible to build Cython modules in parallel, but I have not tested this capability yet.

It seems Pandas tested this capability in their CI and found improvement: pandas-dev/pandas#30214.

leofang:
The parallel build mentioned above is really fast! https://github.com/cupy/cupy/issues/3600#issuecomment-663702728

PR #3067, opened 2020-02-12
Support CUB `prod`

leofang:
Both full and segmented reductions are supported.

TODO: 
- [x] ~~Add tests?~~ **EDIT:** Done in #2598
- [ ] Post timing?

### DIFF
--- a/cupy/core/_routines_math.pyx
+++ b/cupy/core/_routines_math.pyx
@@ -75,6 +75,12 @@ cdef ndarray _ndarray_imag_setter(ndarray self, value):
 
 
 cdef ndarray _ndarray_prod(ndarray self, axis, dtype, out, keepdims):
+    if cupy.cuda.cub_enabled:
+        # result will be None if the reduction is not compatible with CUB
+        result = cub.cub_reduction(self, cub.CUPY_CUB_PROD, axis, dtype, out,
+                                   keepdims)
+        if result is not None:
+            return result
     if dtype is None:
         return _prod_auto_dtype(self, axis, dtype, out, keepdims)
     else:
--- a/cupy/cuda/cub.pxd
+++ b/cupy/cuda/cub.pxd
@@ -6,3 +6,4 @@ cpdef enum cupy_cub_op:
     CUPY_CUB_ARGMAX = 4
     CUPY_CUB_CUMSUM = 5
     CUPY_CUB_CUMPROD = 6
+    CUPY_CUB_PROD = 7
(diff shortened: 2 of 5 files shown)


leofang:
@emcastillo: Your comments https://github.com/cupy/cupy/pull/2919#discussion_r375113326 and https://github.com/cupy/cupy/pull/2919#discussion_r375604770 are addressed here.

cupy/core/_routines_math.pyx
```diff
@@ -75,6 +75,12 @@ cdef ndarray _ndarray_imag_setter(ndarray self, value):
 
 
 cdef ndarray _ndarray_prod(ndarray self, axis, dtype, out, keepdims):
+    if cupy.cuda.cub_enabled:
+        # result will be None if the reduction is not compatible with CUB
+        result = cub.cub_reduction(self, cub.CUPY_CUB_PROD, axis, dtype, out,
+                                   keepdims)
```
emcastillo: Is the dtype promoted in cub?

leofang: No, it uses the same rule as in CUB-backed `sum()`, see the function `cupy.cuda.cub. _cub_reduce_dtype_compatible()`. If the dtype promotion is needed, currently we fall back to the old kernels (without CUB).

I am not sure if we want to handle type promotion within the CUB module. Since we don't write our own kernels in this case, to promote dtype an extra copy needs to be done before passing it to the CUB kernel. I think this was probably why we took the current approach.

emcastillo: in numpy and cupy, the current prod kernel returns an `int64` array if the input is `int32`.

So with cub we can't replicate this behavior, right?

leofang: No, we can’t do it with the CUB kernel, so in this case CuPy’s prod ufunc will be invoked to do the type conversion (`i->l`):
https://github.com/cupy/cupy/blob/de3d32536c39eefea1805f4a7a76a14356eb3bff/cupy/core/_routines_math.pyx#L576

emcastillo:
Do you know how much does build time increase before-after this PR?.

leofang:
Interesting question. Let me test it. 

Regardless of how much longer it becomes after merging this PR, I think this build-time issue (see my https://github.com/cupy/cupy/issues/1961#issuecomment-552718175) will be important to investigate for utilizing CIs efficiently.

leofang:
Seems acceptable?

```bash
time pip install -v -e .  # only the two touched modules get rebuilt
```
Result: 
Before: 5m34.901s
After: 6m10.939s

Note that I didn't limit `--generate-code` for nvcc so I compiled for all archs.

emcastillo:
How long does a non-cub build takes in your system?

leofang:
You meant building the entire project, not just the touched modules?

emcastillo:
ah! sorry i thought it was a complete build.
Then, it is fine.

emcastillo:
Jenkins, test this please

PR #3068, opened 2020-02-13
Properly decrement total bytes in memory pool

### LINKED ISSUE #3008: MemoryPool().free_all_bytes() doesn't change the value of total_bytes()
* Conditions
  - OS/Platform: Window 10
  - CuPy Version          : 7.1.1
  - CUDA Root             : C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0
  - CUDA Build Version    : 10000
  - CUDA Driver Version   : 10020
  - CUDA Runtime Version  : 10000
  - cuBLAS Version        : 10000
  - cuFFT Version         : 10000
  - cuRAND Version        : 10000
  - cuSOLVER Version      : (10, 0, 0)
  - cuSPARSE Version      : 10000
  - NVRTC Version         : (10, 0)
  - cuDNN Build Version   : 7605
  - cuDNN Version         : 7605
  - NCCL Build Version    : None
  - NCCL Runtime Version  : None
* Code to reproduce
  just as the code in https://docs-cupy.chainer.org/en/latest/reference/memory.html, 
  after a=None and mempool.free_all_blocks(), mempool.total_bytes() is still 512.
  The real memory usage in GPU is indeed freed, but the value of total_bytes() is not changed.
  This will make an influence on further malloc action when a non-zero limit is set. The actual memory is enough but the allocator thought total_bytes_limit is reached.

takagi:
Fix #3008 .

This PR the memory pool to properly decrement its total bytes when free memory blocks are released back.

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1114,6 +1114,7 @@ cdef class SingleDeviceMemoryPool:
         cdef set free_list, keep_list
         cdef vector.vector[size_t] new_index
         cdef size_t index
+        cdef size_t size_to_free = 0
 
         if stream_ptr not in self._arenas:
             return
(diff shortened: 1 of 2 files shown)


niboshi:
Thanks
Jenkins, test please

niboshi:
Jenkins, test this please

niboshi:
Thanks! :+1:

PR #2730, opened 2019-11-28
Allow install without thrust

niboshi:
It takes so much time to build cupy from source with thrust enabled, which is not needed in most daily development.

With this PR, thrust can be disabled by setting an environment variable `CUPY_SETUP_ENABLE_THRUST=0`.

Build time:
* before: 9m17s
* thrust disabled: 2m49s

PR #3071, opened 2020-02-13
Fix ROCm build errors

kmaehashi:
* Functions added in `cupy_cuda.h` and `cupy_cuda_common.h`  was missing in its HIP counterpart
* Do not use CUDA-specific libraries (cudnn, etc) in HIP build (regression in https://github.com/cupy/cupy/pull/2730)


emcastillo:
Jenkins, test this please

Issue #2716, opened 2019-11-26
Use `intptr_t` instead of `size_t ` for handle types


niboshi:
Related: #1913

Handles are pointers. `intptr_t` should be used as a representative type.

* [ ] cublas
* [ ] cusolver  #2718
* [ ] cusparse
* [ ] cudnn

anaruse:
Would it be better to use `intptr_t` in #2709 as well?

niboshi:
Yes, but there may be a room for discussion on how to define these types: https://github.com/cupy/cupy/pull/2718#issuecomment-558891688

jakirkham:
Should it be `intptr_t` or `uintptr_t`?

niboshi:
It shouldn't matter much, but I think `intptr_t` is more natural because Python `int` is a signed integer.

PR #3080, opened 2020-02-16
Use `intptr_t` instead of `size_t` for cuBLAS handles

Harshan01:
Modified `size_t` with `intptr_t` for cuBLAS as a part of #2716


Harshan01:
Closing this PR as a PR was created for cuBLAS and cuSPARSE combined

PR #3081, opened 2020-02-16
Use `intptr_t` instead of `size_t` for cuSPARSE and cuBLAS handles

Harshan01:
Modified `size_t` with `intptr_t` for cuSPARSE and cuBLAS as a part of #2716


emcastillo:
Jenkins, test this please

PR #3082, opened 2020-02-16
Use `intptr_t` for cuDNN handles

Harshan01:
Modified `size_t` with `intptr_t` for cuDNN as a part of #2716


emcastillo:
Jenkins, test this please

PR #3085, opened 2020-02-17
Add mean for sparse matrices

wongalvis14:
Equivalent to https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.sparse.spmatrix.mean.html#scipy.sparse.spmatrix.mean

### DIFF
--- a/cupyx/scipy/sparse/data.py
+++ b/cupyx/scipy/sparse/data.py
@@ -99,7 +99,7 @@ class _data_matrix(base.spmatrix):
         if axis == 0 or axis == -2:
             data = cupy.true_divide(data, nRow)
             return self._with_data(data).sum(axis, dtype, out)
-        
+
         else:
             data = cupy.true_divide(data, nCol)
             return self._with_data(data).sum(axis, dtype, out).flatten()
--- a/cupyx/scipy/sparse/data.py
+++ b/cupyx/scipy/sparse/data.py
@@ -73,7 +73,7 @@ class _data_matrix(base.spmatrix):
         """
         return cupy.count_nonzero(self.data)
 
-    def mean(self, axis=None):
+    def mean(self, axis=None, dtype=None, out=None):
         """Compute the arithmetic mean along the specified axis.
 
         Args:
@@ -88,18 +88,21 @@ class _data_matrix(base.spmatrix):
            :meth:`scipy.sparse.spmatrix.mean`
 
         """
-        nRow = self.shape[0]
-        nCol = self.shape[1]
+        validateaxis(axis)
+        nRow, nCol = self.shape
         data = self.data.copy()
+
         if not axis:
             data = cupy.true_divide(data, nRow * nCol)
-            return self._with_data(data).sum(axis)
+            return self._with_data(data).sum(axis, dtype, out)
+
         if axis == 0 or axis == -2:
             data = cupy.true_divide(data, nRow)
-            return self._with_data(data).sum(axis)
+            return self._with_data(data).sum(axis, dtype, out)
+        
         else:
             data = cupy.true_divide(data, nCol)
-            return self._with_data(data).sum(axis).flatten()
+            return self._with_data(data).sum(axis, dtype, out).flatten()
 
     def power(self, n, dtype=None):
         """Elementwise power function.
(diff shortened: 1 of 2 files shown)


wongalvis14:
@toslunar Since the function uses `sum` which handles `dtype` and `out` just before returning, is it sufficient to leave the error testing to `sum`? (and perhaps with a comment?)

cupyx/scipy/sparse/data.py
```diff
@@ -73,6 +73,37 @@ def count_nonzero(self):
...
+        Returns:
+            cupy.ndarray: Summed array.
+
+        .. seealso::
+           :meth:`scipy.sparse.spmatrix.mean`
+
+        """
+        validateaxis(axis)
+        nRow, nCol = self.shape
+        data = self.data.copy()
```
toslunar:
```python
n = ...
data = cupy.true_divide(self.data, n, dtype=dtype)
return self._with_data(data).sum(axis, dtype, out)
```
looks good.  (It seems the `_with_data` method can receive `data` of different `dtype`.)

wongalvis14: I just followed what `power` did. Do I need to change my code?

wongalvis14:
@toslunar Sorry, I'm not sure if I get what you meant. 
Are you saying that lines like `data = cupy.true_divide(data, nRow)` will result in precision / type errors because the second argument is an integer? What comment should I add? Thanks.

cupyx/scipy/sparse/data.py
toslunar: In `power` copies the input and overwrites the new array.  To follow it, `data /= n` after the copy.  Python 2 support has been dropped.

wongalvis14: `data` is copied and `self._with_data(data)` is used to sum, `_with_data` returns a new matrix right? Do you mean I should replace `true_divide` with `/=`?

wongalvis14: @toslunar What is the implication of no Python 2 support?

toslunar: > `data` is copied and `self._with_data(data)` is used to sum, `_with_data` returns a new matrix right?

Right.  There's only a performance issue that the original code allocates memory twice.

> Do you mean I should replace `true_divide` with `/=`?

Yes, (but there's another way).  It reduce the number of memory allocations.  `data /= n` (or `divide(data, n, out=data)`) is different from `data = data / n` (or `data = divide(data, n)`).
- If you choose `data /= n`, cast `data` to `dtype` first.
- What I said is `data = cupy.true_divide(self.data, n, dtype=dtype)` seems OK because `self.data` is kept.

> What is the implication of no Python 2 support?

`divide` or `/` can be used in place of `true_divide`.

wongalvis14: @toslunar It seems that the tests are passing after the latest commit because `sum` outputs dtype correctly

cupyx/scipy/sparse/data.py
```diff
@@ -73,6 +73,35 @@ def count_nonzero(self):
...
+        n = 1
+
+        if not axis:
+            n = nRow * nCol
+        if axis == 0 or axis == -2:
+            n = nRow
+        else:
+            n = nCol
+
+        return self._with_data(data / n).sum(axis, dtype, out).flatten()
```
toslunar:
```suggestion
        return self._with_data(data / n).sum(axis, dtype, out)
```

- `flatten` changes `ndim` if `axis=None`.
- It seems `flatten` is verbose otherwise.

wongalvis14: The current mean for ndarray returns a row for any axis, however if I don't add flatten, this function will return a column if the axis is 1.

toslunar:
Jenkins, test this please.

cupyx/scipy/sparse/data.py
```diff
@@ -73,6 +73,35 @@ def count_nonzero(self):
...
+
+        """
+        validateaxis(axis)
+        nRow, nCol = self.shape
+        data = self.data.copy()
+        n = 1
+
+        if not axis:
+            n = nRow * nCol
+        if axis == 0 or axis == -2:
```
toslunar:
```suggestion
        if axis is None:
            n = nRow * nCol
        elif axis in (0, -2):
```

kmaehashi:
@wongalvis14 Kindly ping.

kmaehashi:
I asked @toslunar to takeover this PR in #3333. Credit goes to @wongalvis14. Thanks!

PR #3087, opened 2020-02-18
[backport] Import `using_allocator` in `cupy.cuda`

emcastillo:
Backport of #2951

### DIFF
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -101,6 +101,27 @@ from cupy.cuda.stream import Stream  # NOQA
 
 
 @contextlib.contextmanager
+def using_allocator(allocator=None):
+    """Sets a thread-local allocator for GPU memory inside
+       context manager
+
+    Args:
+        allocator (function): CuPy memory allocator. It must have the same
+            interface as the :func:`cupy.cuda.alloc` function, which takes the
+            buffer size as an argument and returns the device buffer of that
+            size. When ``None`` is specified, raw memory allocator will be
+            used (i.e., memory pool is disabled).
+
+    Note:
+        This wraps an internal version of this function to provide a
+        `contextmanager` as `contextmanger` decoration doesn't behave
+        well in Cython.
+    """
+    for y in memory._using_allocator(allocator):
+        yield y
+
+
+@contextlib.contextmanager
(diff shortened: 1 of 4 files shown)


emcastillo:

niboshi:
"no-compat" PRs cannot be backported.

We assumed this fix as not no-compat. (https://github.com/cupy/cupy/pull/2951#issuecomment-574123088)

cupy/cuda/memory.pyx
```diff
@@ -581,8 +580,7 @@ cpdef get_allocator():
         return allocator
 
 
-@contextlib.contextmanager
-def using_allocator(allocator=None):
+def _using_allocator(allocator=None):
```
niboshi: Why not just move this function to `cuda/__init__.py`?

niboshi:
I'll make a PR for the above comment.

cupy/cuda/memory.pyx
jakirkham: I tried that, but it relies on implementation details only accessible in that module. Eventually I gave up and left it there adding a wrapper in `cuda/__init__.py`. After all my main interest was just getting it in the docs 😉

That said, if you are able to find a way to fix this, that would be great. 😄

cupy/cuda/__init__.py
```diff
@@ -100,6 +100,27 @@ def is_available():
...
+        allocator (function): CuPy memory allocator. It must have the same
+            interface as the :func:`cupy.cuda.alloc` function, which takes the
+            buffer size as an argument and returns the device buffer of that
+            size. When ``None`` is specified, raw memory allocator will be
+            used (i.e., memory pool is disabled).
+
+    Note:
+        This wraps an internal version of this function to provide a
+        `contextmanager` as `contextmanger` decoration doesn't behave
+        well in Cython.
```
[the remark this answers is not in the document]
jakirkham: That seems fair. Thanks for the feedback 🙂

cupy/cuda/memory.pyx
niboshi: Thank you for a comment!
I sent a fix as #3094.
I'm not 100% sure this is version is better, but please take a look.

PR #2600, opened 2019-11-04
Add fp16 support of CUB

y1r:
Currently, the CUB binding of CuPy doesn't support fp16 but CUB supports it.

I implemented fp16 support of CUB binding by using `float16` class on `carray.cuh`.
I am not sure what class I should use for this implementation, so please give me comments.

PR #3089, opened 2020-02-18
Fix compiler bug when building `cupy.cuda.cub` with CUDA < 9.2 

leofang:
This PR (hopefully) fixes a (yet another) CUB bug, introduced in #2600. We didn't notice that `__half2float` is accessible on host only starting CUDA 9.2 😞 Appveyor was able to catch it, though, see https://github.com/cupy/cupy/pull/2584#issuecomment-587303111. 

cc: @y1r @emcastillo @niboshi


emcastillo:
Jenkins, test this please

PR #3094, opened 2020-02-20
Minor fix to `using_allocator`

niboshi:
Follow up https://github.com/cupy/cupy/pull/3087#pullrequestreview-360104578

I feel this version (introducing `memory._set_thread_local_allocator()`) is better than `for` loop trick, but not 100% confident.

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -541,6 +541,13 @@ cpdef MemoryPointer malloc_managed(size_t size):
 cdef object _current_allocator = _malloc
 cdef object _thread_local = threading.local()
 
+
+def _set_thread_local_allocator(allocator):
+    previous_allocator = getattr(_thread_local, 'allocator', None)
+    _thread_local.allocator = allocator
+    return previous_allocator
+
+
 cpdef MemoryPointer alloc(size):
     """Calls the current allocator.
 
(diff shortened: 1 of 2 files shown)


cupy/cuda/memory.pyx
```diff
@@ -541,6 +541,13 @@ cpdef MemoryPointer malloc_managed(size_t size):
 cdef object _current_allocator = _malloc
 cdef object _thread_local = threading.local()
 
+
+def _set_thread_local_allocator(allocator):
+    previous_allocator = getattr(_thread_local, 'allocator', None)
```
jakirkham: Would it make sense to split this part out into a `_get_thread_local_allocator`?

niboshi: Thanks! I agree that's better. Fixed in 4d27577a0ea1b680ae30c2b688906348e6c57c11
Also rewrote using `except AttributeError` as it's faster than `getattr`.

cupy/cuda/memory.pyx
```diff
@@ -541,6 +541,19 @@ cpdef MemoryPointer malloc_managed(size_t size):
 cdef object _current_allocator = _malloc
 cdef object _thread_local = threading.local()
 
+
+def _get_thread_local_allocator():
+    try:
+        allocator = _thread_local.allocator
+    except AttributeError:
+        allocator = _thread_local.allocator = None
```
jakirkham: Should this be modifying `_thread_local.allocator`?

niboshi: Yes, semantically the default value of `_thread_local.allocator` should be `None`.
Assiging this at the first access saves the `AttributeError` in the second and later time.

emcastillo:
Jenkins, test this please

cupy/cuda/memory.pyx
jakirkham: Yeah it feels a little weird, but I understand why it's there. Keeping it that way makes sense.

PR #3096, opened 2020-02-20
Remove `IndexOrValueError`

asi1024:
`numpy.AxisError` was introduced from NumPy 1.13.0.
We decided to drop NumPy<1.15 support from CuPy v8. (#2938)


niboshi:
Jenkins, test this please

PR #3100, opened 2020-02-20
Fix `cub_reduction` for `CUPY_CUB_MIN` and float16 arrays

### LINKED ISSUE #3097: `cupy.min` returns wrong result when CUB is enabled
`cupy.min` returns a wrong value when CUB is enabled for float16 inputs.

Reproducer:
```py
import cupy

print(cupy.array([1, 2, 3], dtype='e').min())
print(cupy.array([1, 2, 3], dtype='e').max())
print(cupy.array([1, 2, 3], dtype='f').min())
print(cupy.array([1, 2, 3], dtype='f').max())
```

Result:
```
65500.0
3.0
1.0
3.0
```

asi1024:
Fix #3097.


niboshi:
Jenkins test this please

PR #2461, opened 2019-09-12
support array-like start/stop and add axis argument to linspace

grlee77:
This supports array-like inputs and an `axis` argument in linspace as implemented in NumPy 1.16+.

closes #2446 

~I have marked this as WIP due to what is currently relatively slow performance for the array-like code path.~ (**edit:** slow for small `num`. For large enough `num`, cupy should always be faster)

A basic translation of the code from numpy 1.17 ends up being pretty slow for small arrays. It can be faster to just run `numpy.linspace` and transfer the result to the GPU in the case of array-like `start` and/or `stop` (as currently done in #2399).

For now, I check if both start & stop are scalars and run the current implementation as a fast path. This preserves performance for this common case, but results in some code duplication.

PR #3101, opened 2020-02-20
cupy.pad: update to use cupy.linspace instead of numpy.linspace internally

grlee77:
Now that #2461 was merged, use of `numpy.linspace` within `cupy.pad` can be avoided


niboshi:

PR #1746, opened 2018-10-25
add PlanNd class for faster 2D and 3D FFTs

grlee77:
This PR is based on top of #1745, so that one should be reviewed first.

For n-dimensional data where 2 or 3 axes are transformed, `cufftMakePlanMany` is faster than applying separable 1D plans. This PR implements this capability. This benefits of this were previously discussed in #1669.

The primary downside of an n-dimensional plan vs separable 1d plans is increased GPU memory usage for the plan itself. For this reason, I made a global configuration flag at `cupy.fft.config.enabled_nd_planning` that can be set False to disable the n-dimensional planning.

The functions that benefit from this PR are:
```
cupy.fft.fft2
cupy.fft.ifft2
cupy.fft.fftn
cupy.fft.ifftn
cupyx.scipy.fftpack.fft2
cupyx.scipy.fftpack.ifft2
cupyx.scipy.fftpack.fftn
cupyx.scipy.fftpack.ifftn
```

This PR does not involve any changes to the CuPy public API aside from the addition of the `cupy.fft.config` module. 

Caching of FFT plans can also provide an additional large performance boost, but that can be the topic of a separate PR.

PR #3102, opened 2020-02-20
Add R2C/C2R support to `cupy.cuda.cufft.PlanNd`

### LINKED ISSUE #1936: [Feature Request]_fftn for R2C
I'm working on the implementation of STFT, and I think `cuFFTPlanMany` is a good API to implement it. Fortunately, in `cupy.cuda.cufft.PlanNd` is already implemented as the corresponding API, but in `cupy.fft.fft.get_cufft_plan_nd` only allows C2C and Z2Z.

Is there any reason for it? Thanks!

### LINKED ISSUE #3263: cufftPlanMany() argument support in cupyx.scipy.fftpack.get_fft_plan
I'm looking for the ability to do multiple overlapped FFTs on a 1D array of data. With cufftPlanMany() function in cuFFT I can set the istride/ostride and idist/odist arguments to accomplish this. I can also set the type to R2C, C2R, C2C (and other datatype equivalents).

I appreciate that cupyx.scipy.fftpack.get_fft_plan gives me the ability to set a plan prior to running multiple FFTs. However, it doesn't seem to give the same functionality as cufftPlanMany does. It is also currently restricted to C2C or Z2Z. Can get_fft_plan be expanded to give the user the same options as cufftPlanMany offers?

leofang:
Close #1936. Close #3263.

All tests passed locally  (P100 + CUDA 9.2). Note that
1. Passing an R2C/C2R plan from `get_fft_plan()` to functions in `cupyx.scipy.fftpack` or `cupyx.scipy.fft` is yet to be supported, most likely in a separate PR.
2. Setting `overwrite_x=True` for R2C/C2R transforms does not lead to in-place transform (I think it's too much work, but I could be wrong). The flag is silently ignored to ensure correctness for now.

~~TODO:~~
1. ~~Performance comparison~~ 
2. ~~More code cleanup~~

Thanks to @grlee77 for pointing me to some old cuFFT resources discussing how to handle F-order arrays (leofang/cupy#7). 🙂


cupy/cuda/cufft.pyx
```diff
@@ -215,29 +217,29 @@ class Plan1d(object):
...
         cdef size_t work_size
         cdef int ndim, i
-        cdef int[:] shape_arr = numpy.asarray(shape, dtype=numpy.intc)
-        cdef int[:] inembed_arr
-        cdef int[:] onembed_arr
-        cdef int* shape_ptr = &shape_arr[0]
+        cdef vector.vector[int] shape_arr = shape
+        cdef vector.vector[int] inembed_arr
+        cdef vector.vector[int] onembed_arr
+        cdef int* shape_ptr = shape_arr.data()
```
leofang: Using `std::vector` over numpy array gives 10x to 20x speedup on my machine.

leofang:
A quick benchmark for full-array R2C/C2R:
```python
>>> import cupy as cp
>>> import cupyx
>>>
>>> ca = cp.random.random((512, 512, 512))
>>> cp.fft.config.enable_nd_planning = True
>>> print(cupyx.time.repeat(cp.fft.rfftn, (ca, ), n=50))
rfftn               :   641.977 us   +/-53.088 (min:  528.481 / max:  717.539) us  17608.998 us   +/-67.957 (min:17450.815 / max:17750.336) us
>>> cp.fft.config.enable_nd_planning = False
>>> print(cupyx.time.repeat(cp.fft.rfftn, (ca, ), n=50))
rfftn               :  1582.029 us   +/-149.967 (min: 1483.209 / max: 1956.003) us  43954.073 us   +/-42.902 (min:43911.072 / max:44044.319) us
>>>
>>> cb = cp.random.random((512, 512, 512)).astype(cp.complex128)
>>> cp.fft.config.enable_nd_planning = True
>>> print(cupyx.time.repeat(cp.fft.irfftn, (cb, ), n=50))
irfftn              :   682.115 us   +/-35.178 (min:  640.845 / max:  823.347) us  69353.386 us   +/-69.217 (min:69193.855 / max:69561.440) us
>>> cp.fft.config.enable_nd_planning = False
>>> print(cupyx.time.repeat(cp.fft.irfftn, (cb, ), n=50))
irfftn              :  1748.892 us   +/-42.444 (min: 1711.711 / max: 1858.933) us  141214.269 us   +/-15.825 (min:141180.191 / max:141247.650) us
```

leofang:
@asi1024 @grlee77 The methods `check_output_array` in `Plan1d`/`PlanNd` was added back in #1746 but was never called in any call path. Any reason we wanna keep it? Can I remove them in this PR?

grlee77:
> @asi1024 @grlee77 The methods check_output_array in Plan1d/PlanNd was added back in #1746 but was never called in any call path. Any reason we wanna keep it? Can I remove them in this PR?

What do you mean? I think it is called in both `_exec_fft` and `_exec_fftn`, see:
https://github.com/cupy/cupy/blob/de3d32536c39eefea1805f4a7a76a14356eb3bff/cupy/fft/fft.py#L398

leofang:
Yes, but the line you quoted never gets executed. It's because while `_exec_fft` and `_exec_fftn` accept an `out` argument, there's no way to pass in `out` from any public API.

leofang:
Please do not merge. I am investigating some issues with F-order arrays.

leofang:
I think I hit this cuFFT implementation detail that led to a bunch of errors from the (updated) test suite:
> Note: Complex-to-real (C2R) transforms accept complex-Hermitian input, which requires the 0th element (and the N/2-th input if N is even) to be real-valued, i.e. its imaginary part should be zero. Otherwise, the behavior of the transform is undefined.

(from [cuFFT doc](https://docs.nvidia.com/cuda/cufft/index.html#fft-types)). My PR is correctly done in the sense that `irfftn(rfftn(a)) = a`, but apparently whatever happens in the intermediate step (the result of `rfftn`, and thus the acceptable input form for `irfftn`) is up to the FFT implementation, and cuFFT apparently takes a different approach with 2D transform + shape reversal (i.e. F order):
```python
a_np = np.asfortranarray(np.random.random((3,4)))
b_np1 = np.fft.rfftn(a_np)
b_np2 = np.fft.irfftn(b_np1)
print(np.allclose(a_np, b_np2))   # -> True

a_cp = cp.asfortranarray(cp.asarray(a_np))
b_cp1 = cp.fft.rfftn(a_cp)
b_cp2 = cp.fft.irfftn(b_cp1)
print(cp.allclose(a_cp, b_cp2))   # -> True

print(cp.allclose(a_np, a_cp))    # -> True
print(cp.allclose(b_np1, b_cp1))  # -> False (!); expected to be True in the tests
print(cp.allclose(b_np2, b_cp2))  # -> True
```
(For comparison, if the input is in C order, or for 3D transform (say `shape=(2,3,4)`) + F order, then it's all `True`.)

@asi1024 @grlee77 What do you think if I only add this PlanNd feature for R2C/C2R with C order  input (i.e. let `_fftn` handle this case), and fall back to `_fft` for F order input?

leofang:
> or for 3D transform (say `shape=(2,3,4)`) + F order, then it's all `True`.

I want to add that even for 3D transform, if the input of `irfftn` were not generate from the output of `rfftn` (say we stuff some random array as done in the test suite), then its result would not match the NumPy equivalent (with the same input). cuFFT's C2R transform seems to make a very strong assumption on the input.

grlee77:
> Yes, but the line you quoted never gets executed. It's because while _exec_fft and _exec_fftn accept an out argument, there's no way to pass in out from any public API.

Okay, right. I guess I had added that argument in my initial PlanNd PR, but then we never ended up using it. In that case I am fine with removing both that `out` argument and the `check_output_array` method.

It seems that the needed checks on the input array are being done within the body of `_exec_fftn`. I suppose those could potentially go in a method of `PlanNd`, but I'm not sure it would be preferable.

leofang:
I have a second thought and think I will leave them for now. They might be useful when adding multi-GPU support (not sure, because N-D multi-GPU cuFFT seems so complicated; I haven't figured it out yet).

leofang:
I avoided doing N-D R2C/C2R transforms for F-order arrays. All FFT tests passed locally. PTAL.

leofang:
It seems we don't have to consider in-place `rfftn`/`irfftn` at all -- Scipy doesn't support it either: 
https://github.com/scipy/scipy/blob/89d5349b2fb070c4349a3b5b54d39be68148e778/scipy/fft/_pocketfft/basic.py#L224

leofang:
Hi @asi1024, it seems we have a real need (#3263) for this PR. Any chance this PR can be reviewed sooner? Thanks.

leofang:
Fix a "defect" in the tests in 40a9174: It's funny that `axes` are parametrized but never used in those tests! 🤣

leofang:
On GTX 2080 Ti + CUDA 10.0:
```python
>>> import cupy as cp
>>>
>>> a = cp.random.random((512, 512, 512)) + 1j*cp.random.random((512, 512, 512))
>>>
>>> cp.fft.config.enable_nd_planning = False  # without PlanNd
>>> print(repeat(cp.fft.rfftn, (a,), n=20))
rfftn               :   850.499 us   +/-39.071 (min:  817.100 / max: 1007.183) us  85474.685 us   +/-1226.721 (min:82027.298 / max:87257.919) us
>>> cp.fft.config.enable_nd_planning = True   # with PlanNd
>>> print(repeat(cp.fft.rfftn, (a,), n=20))
rfftn               :   338.843 us   +/-28.027 (min:  310.853 / max:  431.270) us  43563.003 us   +/-11.552 (min:43547.295 / max:43592.194) us

File: cupyx/scipy/fft/fft.py [hunk 2]
```diff
@@ -6,7 +6,7 @@
 import cupy
 from cupy.cuda import cufft
 from cupy.fft.fft import (_fft, _default_fft_func, hfft as _hfft,
-                          ihfft as _ihfft)
+                          ihfft as _ihfft, _size_last_transform_axis)
 from cupy.fft.fft import fftshift, ifftshift, fftfreq, rfftfreq
```
asi1024, line 10: Sorry for my overlook before but importing non-module symbols is prohibited.

leofang: @asi1024 Would you like me to fix all imports in this file, or just my addition?

leofang: I think it was done like this for a reason: you can't import the module and then access the private functions without first exposing them in `cupy/fft/fft/__init__.py`...

leofang: Any comment, @asi1024?

asi1024: Ah, I see. Thanks!

asi1024:
Jenkins, test this please.

leofang:
Thanks, @asi1024! I fixed the errors. PTAL.

There're two errors: First, `s` and `axes` should be sequences, which is addressed in 8894af7. Second, TestFft2 is fixed for the corner case `axes=()`. The fix in 2d3ecf5 follows that of TestFftn, but I have no idea why we missed it in #3264 🤬 @toslunar, do you know why?

leofang:
@asi1024 I forgot about the cuFFT bug 🤦🏻‍♂️ It's now fixed in b65c532. Could you rerun the CIs please?

asi1024:
Jenkins, test this please.

asi1024:
LGTM! Thanks for your contribution!

leofang:
Thanks a lot, @asi1024 @grlee77!

Issue #2224, opened 2019-05-30
Add CuPy compress


jakirkham:
Would be useful to have an [`compress`]( https://docs.scipy.org/doc/numpy/reference/generated/numpy.compress.html#numpy.compress ) implementation for CuPy arrays.

andy6975:
Is this issue still open? I would like to work on this. Thanks!

jakirkham:
I think you can just give it a go. Not aware of anyone else working on it. Please let us know if you need anything. 🙂

andy6975:
Thank you for the response. If you don't mind, can I ask you for a headstart? I am new to cupy.

grlee77:
Implementation can sometimes be tricky because you often need some familiarity with NumPy/C to determine what is going on there and then some familiarity with CUDA/C++ to be able to implement an equivalent function for CuPy.  

In the case of `compress`, you could check whether maybe it can be implemented by adapting an existing function like `_take` as defined in [cupy/core/_routines_indexing.pyx](https://github.com/cupy/cupy/blob/master/cupy/core/_routines_indexing.pyx#L592). (NumPy's compress ultimately calls [PyArray_Compress](https://github.com/numpy/numpy/blob/cacdf265755ffae9d5c07d35fafa3d1366e342b8/numpy/core/src/multiarray/item_selection.c#L1936))

An easier, pure Python route is available in cases where a function can be defined in terms of other functions already implemented in CuPy. In the case of `compress`, it seems like the following the following might work:

```Python
def compress((condition, a, axis=None, out=None):
    # TODO: replace with more efficient kernel-based implementation
    return cupy.take(a, cupy.where(condition)[0], axis, out)
```

This is not computationally optimal as we have to call both `cupy.where` and `cupy.take` instead of just a single underlying CUDA kernel, but it is better than nothing. I think a PR of that sort could potentially be accepted if good tests are added to make sure the functionality matches what is provided in NumPy.

Tests would need to be added to the existing file `tests/cupy_tests/core_tests/test_ndarray_indexing.py`.

grlee77:
Hey @rgommers: my prior comment here reminded me of a discussion you started (maybe on numpy-dev?) about what defining certain core functions that are most important within the NumPy API.

It seems like `compress` (and `extract`) is an example of a non-core function that can be implemented purely in terms of other existing NumPy functions (despite having a dedicated C implementation in NumPy itself). I just thought I would ping you in case you are working on tabulating these somewhere.

[...]

PR #3103, opened 2020-02-20
Add `compress`

Harshan01:
I tried to mimic numpy as much as possible. I observed that `PyArray_Compress` in numpy internally calls `nonzero` and `take`. So, I tried to do it in the same way instead of making a separate kernel. I would be happy to receive suggestions on how to implement the kernel for `compress`, if it is required to do so.

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -436,6 +436,7 @@ from cupy.indexing.generate import unravel_index  # NOQA
 
 from cupy.indexing.indexing import choose  # NOQA
 from cupy.indexing.indexing import diagonal  # NOQA
+from cupy.indexing.indexing import compress  # NOQA
 from cupy.indexing.indexing import take  # NOQA
 from cupy.indexing.indexing import take_along_axis  # NOQA
 
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -435,8 +435,8 @@ from cupy.indexing.generate import r_  # NOQA
 from cupy.indexing.generate import unravel_index  # NOQA
 
 from cupy.indexing.indexing import choose  # NOQA
-from cupy.indexing.indexing import diagonal  # NOQA
 from cupy.indexing.indexing import compress  # NOQA
+from cupy.indexing.indexing import diagonal  # NOQA
 from cupy.indexing.indexing import take  # NOQA
 from cupy.indexing.indexing import take_along_axis  # NOQA
 
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -152,6 +152,30 @@ cdef ndarray _ndarray_choose(ndarray self, choices, out, mode):
     return out
 
 
+cdef ndarray _ndarray_compress(ndarray self, condition, axis, out):
+    a = self
+
+    if numpy.isscalar(condition):
+        raise ValueError('condition must be a 1-d array')
+    
+    if not isinstance(condition, ndarray):
+        condition = core.array(condition, dtype=int)
+        if condition.ndim != 1:
+            raise ValueError('condition must be a 1-d array')
+    
+    res = _ndarray_nonzero(condition)
+    
+    if out is None:
+        out = ndarray(out_shape, dtype=a.dtype)
+    else:
+        if out.dtype != a.dtype:
+            raise TypeError('Output dtype mismatch')
+        if out.shape != out_shape:
+            raise ValueError('Output shape mismatch')
+    
+    return _ndarray_take(a, res[0], axis, out)
+
+
 cdef ndarray _ndarray_diagonal(ndarray self, offset, axis1, axis2):
     return _diagonal(self, offset, axis1, axis2)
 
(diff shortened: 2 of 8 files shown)


Harshan01:
This PR is for the feature request Issue #2224

cupy/__init__.py
```diff
@@ -436,6 +436,7 @@ def result_type(*arrays_and_dtypes):
 
 from cupy.indexing.indexing import choose  # NOQA
 from cupy.indexing.indexing import diagonal  # NOQA
+from cupy.indexing.indexing import compress  # NOQA
```
niboshi: Please swap them to sort it in alphabetical order.

cupy/core/_routines_indexing.pyx
```diff
@@ -152,6 +152,22 @@ cdef ndarray _ndarray_choose(ndarray self, choices, out, mode):
...
+cdef ndarray _ndarray_compress(ndarray self, condition, axis, out):
+    a = self
+
+    if numpy.isscalar(condition):
+        raise ValueError('condition must be a 1-d array')
+
+    if not isinstance(condition, ndarray):
+        condition = core.array(condition, dtype=int)
+        if condition.ndim != 1:
+            raise ValueError('condition must be a 1-d array')
```
niboshi: We do not accept numpy arrays as argument in general.
I think we can just ensure `isinstance(condition, ndarray)`.

cupy/core/_routines_indexing.pyx
```diff
@@ -152,6 +152,22 @@ cdef ndarray _ndarray_choose(ndarray self, choices, out, mode):
...
+
+    res = _ndarray_nonzero(condition)
```
niboshi: Let's leave a comment to note that this line causes host-device synchronization.
```suggestion
    res = _ndarray_nonzero(condition)  # synchronize
```

cupy/indexing/indexing.py
```diff
@@ -76,7 +76,28 @@ def choose(a, choices, out=None, mode='raise'):
     return a.choose(choices, out, mode)
 
 
-# TODO(okuta): Implement compress
+def compress(condition, a, axis=None, out=None):
+    """Return selected slices of an array along given axis.
```
niboshi: Let's document that this function may synchronize, similarly to:

https://github.com/cupy/cupy/blob/1dc5e6deb8faa8157551d46ecbf5de5ede58a69d/cupy/core/core.pyx#L712-L714

niboshi:
```suggestion
    """Returns selected slices of an array along given axis.
```

cupy/core/core.pyx
```diff
@@ -724,7 +724,15 @@ cdef class ndarray:
 
         return _indexing._ndarray_nonzero(self)
 
-    # TODO(okuta): Implement compress
+    cpdef ndarray compress(self, condition, axis=None, out=None):
+        """Return selected slices of this array along given axis.
```
niboshi:
```suggestion
        """Returns selected slices of this array along given axis.
```

cupy/indexing/indexing.py
```diff
@@ -76,7 +76,28 @@ def choose(a, choices, out=None, mode='raise'):
...
+        axis (int): Axis along which to take slices. If None (default), work
+            on the flattened array.
+        out (cupy.ndarray): Output array. If provided, it should be of
+            appropriate shape and dtype.
+
+    Returns:
+        cupy.ndarray: A copy of a without the slices along axis for which
+            condition is false.
+
+    . seealso:: :func:`numpy.compress`
```
niboshi:
```suggestion
    .. seealso:: :func:`numpy.compress`
```

niboshi:
Please test some additional configurations:
* `condition` is not `bool`. Maybe use `cupy.testing.for_int_dtypes` to parameterize its dtype?
* `condition` is an empty 1-dim array (`cupy.ndarray([])`).

cupy/core/_routines_indexing.pyx
Harshan01: I am not sure if I understand what you said. Doesn't `ndarray` mean only cupy ndarrays.

niboshi: Yes, that's correct. `ndarray` is `cupy.ndarray` in this context.

Harshan01: If we just assert that instance is of `ndarray`, then we might not be able to take a python list of booleans as input. That's why I thought I might convert it to cupy ndarray (which also has a side-effect of converting numpy arrays also to cupy ndarrays). Otherwise, I would need to check manually for a list of booleans.

niboshi: We do not accept list of booleans as well.

niboshi: Sorry, as `take` accepts both lists and NumPy arrays as `indices`, I think `condition` can support them as well.

niboshi:
Thanks. Jenkins, test this please

niboshi:
Forgot one thing, could you add a documentation entry to `docs/source/reference/indexing.rst`?

Harshan01:
@niboshi I have made the requested changes. Can you please review them?

niboshi:
Thanks. Jenkins, test this please

niboshi:
👍

PR #3098, opened 2020-02-20
Deprecate `testing.numpy_cupy_raises`

asi1024:

PR #3104, opened 2020-02-20
add ravel_multi_index

grlee77:
This PR adds `cupy.ravel_multi_index`

### DIFF
--- a/cupy/indexing/generate.py
+++ b/cupy/indexing/generate.py
@@ -1,5 +1,7 @@
 # class s_(object):
 
+import numbers
+
 import numpy
 
 import cupy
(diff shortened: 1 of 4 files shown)


cupy/indexing/generate.py
```diff
@@ -244,7 +248,105 @@ def ix_(*args):
...
+    array([31, 41, 13])
+    >>> cupy.ravel_multi_index(cupy.asarray([[3,6,6],[4,5,1]]), (4,6),
+    ...                        mode='clip')
+    array([22, 23, 19])
+    >>> cupy.ravel_multi_index(cupy.asarray([[3,6,6],[4,5,1]]), (4,4),
+    ...                        mode=('clip', 'wrap'))
+    array([12, 13, 13])
+    >>> cupy.ravel_multi_index(cupy.asarray((3,1,4,1)), (6,7,8,9))
+    array(1621)
+
```
emcastillo: Please add a note like in the other functions that may `synchronize` the device

cupy/indexing/generate.py
```diff
@@ -244,7 +248,105 @@ def ix_(*args):
         out.append(new)
     return tuple(out)
 
-# TODO(okuta): Implement ravel_multi_index
+
+def ravel_multi_index(multi_index, dims, mode='raise', order='C'):
```
emcastillo: The default mode shouldn't be raise but `wrap` to avoid hidden device synchronization.

https://docs-cupy.chainer.org/en/stable/reference/difference.html

cupy/indexing/generate.py
```diff
@@ -244,7 +248,105 @@ def ix_(*args):
...
+            s = s * dims[i + 1]
+            ravel_strides[i] = s
+    elif order == "F":
+        for i in range(1, ndim):
+            s = s * dims[i - 1]
+            ravel_strides[i] = s
+    else:
+        raise TypeError("order not understood")
+
+    raveled_indices = 0
```
emcastillo: Instead of setting it to 0, I feel its better to just allocate a `cupy.array` here rather than changing the type of this variable later inside the loop.

cupy/indexing/generate.py
grlee77: Please check now (I think this can only occur when mode='raise'?)

emcastillo:
Jenkins, test this please

grlee77:
I noticed #3098 after submitting this. Should I be refactoring tests to not use `numpy_cupy_raises`? If so I can either do it now or as a follow-up PR.

grlee77:
The latest commit should have fixed the single CI failure, please retry

emcastillo:
> I noticed #3098 after submitting this. Should I be refactoring tests to not use `numpy_cupy_raises`? If so I can either do it now or as a follow-up PR.

Should be ok!

emcastillo:
Jenkins, test this please

Issue #2223, opened 2019-05-30
Add CuPy extract


jakirkham:
Would be useful to have an [`extract`]( https://docs.scipy.org/doc/numpy/reference/generated/numpy.extract.html#numpy.extract ) implementation for CuPy arrays.

andy6975:
Is this issue open? I would like to work on this. Thanks!

jakirkham:
Yes, please feel free 🙂

Harshan01:
Hi, is the issue still open? I would like to work on `extract`, if it is open.

jakirkham:
Yeah, that would be great if you could 🙂

Harshan01:
I have made a PR for `extract`. #3109 
The details of the implementation are mentioned in the PR.
I would be happy to receive suggestions on improving it.

jakirkham:
Awesome, thanks Harshan! 😀

asi1024:
Merged #3109.

PR #3109, opened 2020-02-21
Add `extract`

Harshan01:
This PR is for the feature request Issue #2223

### DIFF
--- a/cupy/indexing/indexing.py
+++ b/cupy/indexing/indexing.py
@@ -112,7 +112,7 @@ def extract(condition, a):
     ``arr[condition]``.
 
     Args:
-        condition (cupy.ndarray): An array whose nonzero or True entries
+        condition (int or array_like): An array whose nonzero or True entries
             indicate the elements of array to extract.
         a (cupy.ndarray): Input array of the same size as condition.
 
--- a/cupy/indexing/indexing.py
+++ b/cupy/indexing/indexing.py
@@ -134,7 +134,7 @@ def extract(condition, a):
 
     a = a.ravel()
     condition = condition.ravel()
-    
+
     return a.take(condition.nonzero()[0])
 
 
--- a/cupy/indexing/indexing.py
+++ b/cupy/indexing/indexing.py
@@ -104,4 +104,38 @@ def diagonal(a, offset=0, axis1=0, axis2=1):
     return a.diagonal(offset, axis1, axis2)
 
 
+def extract(condition, a):
+    """Return the elements of an array that satisfy some condition.
+
+    This is equivalent to ``np.compress(ravel(condition), ravel(arr))``.
+    If ``condition`` is boolean, ``np.extract`` is equivalent to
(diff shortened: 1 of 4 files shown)


Harshan01:
Though numpy mentions that the `condition` should be of the same size as the input array, it works for all sizes which are less than that of the array. So, I am not sure if tests have to be added for empty array cases and cases where `condition` is of smaller size than input array.

Harshan01:
Presently, I made sure that the input array has to be `cupy.ndarray`. I would like to receive suggestions if the function should also accept python list and numpy array as input array.

tests/cupy_tests/indexing_tests/test_indexing.py
```diff
@@ -93,6 +93,21 @@ def test_diagonal_invalid2(self, xp):
...
+                      [True, False, True]])
+        return xp.extract(b, a)
+
+    @testing.for_signed_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_extract_no_bool(self, xp, dtype):
+        a = testing.shaped_arange((3, 3), xp)
+        b = xp.array([[1, 0, 1], [0, 1, 0], [1, 0, 1]], dtype=dtype)
+        return xp.extract(b, a)
+
```
asi1024: Please add tests for `a.shape != b.shape` and `a.size != b.size`

cupy/indexing/indexing.py
```diff
@@ -104,4 +104,38 @@ def diagonal(a, offset=0, axis1=0, axis2=1):
...
+            This function may synchronize the device.
+
+    .. seealso:: :func:`numpy.extract`
+    """
+
+    if not isinstance(a, cupy.ndarray):
+        raise TypeError('extract requires input array to be cupy.ndarray')
+
+    if not isinstance(condition, cupy.ndarray):
+        condition = cupy.array(condition)
```
asi1024: The above docstring says that the argument `condition` is required to be of `cupy.ndarray`.

jakirkham: Well NumPy also allows this.

```python
In [1]: import numpy                                                            

In [2]: a = numpy.arange(6).reshape(2, 3)                                       

In [3]: a                                                                       
Out[3]: 
array([[0, 1, 2],
       [3, 4, 5]])

In [4]: numpy.extract([[True, False, True], 
   ...:                [False, True, True]], 
   ...:               a)                                                        
Out[4]: array([0, 2, 4, 5])
```

Though perhaps the question is how do we specify something that is `ndarray`-like in a way that is compatible with CuPy's documentation practices 🙂

Harshan01: I too had the same doubt. There was no mention of `array_like` in CuPy's documentations.

Harshan01:
@asi1024 I have made the requested changes. Can you please review them?

asi1024:
Jenkins, test this please.

asi1024:

PR #3117, opened 2020-02-24
Type `x_size` as `size_t`

jakirkham:
cc @leofang

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -119,8 +119,8 @@ def device_reduce(ndarray x, op, tuple out_axis, out=None,
                   bint keepdims=False):
     cdef ndarray y
     cdef memory.MemoryPointer ws
-    cdef int dtype_id, ndim_out, kv_bytes, x_size, op_code
-    cdef size_t ws_size
+    cdef int dtype_id, ndim_out, kv_bytes, op_code
+    cdef size_t x_size, ws_size
     cdef void *x_ptr
     cdef void *y_ptr
     cdef void *ws_ptr
@@ -155,7 +155,7 @@ def device_reduce(ndarray x, op, tuple out_axis, out=None,
     y_ptr = <void *>y.data.ptr
     dtype_id = _get_dtype_id(x.dtype)
     s = <Stream_t>stream.get_current_stream_ptr()
-    x_size = <int>x.size
+    x_size = <size_t>x.size
     ws_size = cub_device_reduce_get_workspace_size(x_ptr, y_ptr, x.size, s,
                                                    op, dtype_id)
     ws = memory.alloc(ws_size)
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -156,7 +156,7 @@ def device_reduce(ndarray x, op, tuple out_axis, out=None,
     dtype_id = _get_dtype_id(x.dtype)
     s = <Stream_t>stream.get_current_stream_ptr()
     x_size = <size_t>x.size
-    ws_size = cub_device_reduce_get_workspace_size(x_ptr, y_ptr, x.size, s,
+    ws_size = cub_device_reduce_get_workspace_size(x_ptr, y_ptr, x_size, s,
                                                    op, dtype_id)
     ws = memory.alloc(ws_size)
     ws_ptr = <void *>ws.ptr


jakirkham:
On the one hand this seems to be the right thing to do with `ndarray.size`. On the other hand it appears CUB actually limits itself types `num_items` as `int`. Not sure how we handle this.

@anaruse, do you have any thoughts on this? 🙂

leofang:
The CUB API requires it to be int...There’s nothing we can do unless CUB fixes this in the next version (I do agree it’s an awful choice).

niboshi:
`x.size` is `Py_ssize_t`, which is a signed integer, whereas `size_t` is unsigned.

Besides whether `cub_device_reduce_get_workspace_size` should accept `int` or not, given it does, I think it's reasonable to use `int` in our code.

jakirkham:
AIUI `Py_ssize_t` is really intended for indexing (hence why it signed). Here we are trying to get the size of an object, which is only positive. Also this aligns with the other code getting sizes here and using the same type.

leofang:
Reference for `num_item` being `int` in CUB API:
http://nvlabs.github.io/cub/structcub_1_1_device_reduce.html#ab7f21e8255eb842aaf74305975ae607f
It doesn’t matter how we wrap it in C/Cython. If you examine the CUB implementation you’ll find `int` everywhere.

jakirkham:
Yeah I agree. That seems like the real issue. Am less clear on how we solve that ATM.

jakirkham:
Looks like issue ( https://github.com/NVlabs/cub/issues/129 ) is relevant.

leofang:
GJ for finding the ref, John! I just replied you there. 

I read the discussion and agree that the decision to take int make sense, though. I did see significant register pressure when profiling CUB kernels, in particular for complex128. The performance would get much worse if we use size_t by default.

jakirkham:
Thanks Leo!

That's a good point. So how do we address this? Should we breaking up application of this kernel onto batches of data to work on or something else? Or is this something that gets solved by other things on-top of CuPy (like Dask for example)?

leofang:
I don't have good suggestion for this case...We can either do batching like you said in the Python space, or write a custom wrapper on top of `DispatchReduce::Dispatch()` that accepts `size_t` num_items. The cost for the former is multiple kernel launches, and for the latter one suboptimal kernel. Only benchmarks can tell us which one might be better (probably on a case-by-case basis?)

Another thing that worries me for the latter (I started thinking about this issue when @emcastillo asked https://github.com/cupy/cupy/pull/3067#discussion_r382429860) is that currently compiling `cupy.cuda.cub` takes excessive time (almost 10 mins on my machine, comparable to `cupy.cuda.thrust`). It's because we specialized a whole lot of CUB templates to be able to tackle different dtypes. It seems the latter will double the number of template specializations, and thus compilation time. Would be a decision up to the core devs I guess, because the CIs would be slowed down quite a lot.

BTW, I only focused on full array reduction so far and have not thought about partial reduction (using `device_reduce_segmented`).

Finally, just FYI, I was wondering how CuPy's own reduction kernel handles this oversize issue, and I found there're two call paths: For `num_items > sizeof(int)`, the indexing is done here https://github.com/cupy/cupy/blob/32718607a7808ec6bc3a24cf9231a9351f8fc95e/cupy/core/include/cupy/carray.cuh#L287
and  for `num_items < sizeof(int)` it's here
https://github.com/cupy/cupy/blob/32718607a7808ec6bc3a24cf9231a9351f8fc95e/cupy/core/include/cupy/carray.cuh#L301
(This is possible only because the index is stored as `ptrdiff_t`...Not saying we owe to use or need the `CIndexer` class there though.)

jakirkham:
Sorry for the slow reply Leo. 😞

That's a really nice write-up! 😄

So what do you think we should do at this point? My guess is this PR isn't needed (or at least not without some other changes). Should we close it out and file an issue? Maybe ask Akira what we should do next? These are all just thoughts. Happy to go another way 🙂

niboshi:
Closing this PR and opening an issue sounds good :+1:

leofang:
Yeah a new issue would be good. 

btw, Emilio asked me to measure the compile time increase for the CUB module. For the case of #3067, the increase isn't as drastic as I argued. But there we didn't have to double the number of template specializations, so perhaps it's not too surprising.

jakirkham:
(I am kind of tempted to steal your write-up and stick it in a new issue 😂)

leofang:
Yeah go ahead!

emcastillo:
@jakirkham can you close this PR and create an issue instead? Thanks! 😇

emcastillo:
Closing this PR, let's discuss this in #3309

jakirkham:
Sorry for the delay! Thanks for opening that issue @emcastillo! 😄

PR #3118, opened 2020-02-25
Use larger type to represent index range in `cupy.take`

### LINKED ISSUE #3017: cupy.take() does not function correctly when indices are unsigned integers or negative
```
CuPy Version          : 7.1.1
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSPARSE Version      : 10301
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2506
NCCL Runtime Version  : 2506
```

**Code to reproduce**
```
a = cp.ones(256)
indices = cp.array([150], dtype=cp.uint8)
a.take(indices)
```
This should result in an array with the value `1` in it (since the original `a` was only ones) however this gets `0` or other random values instead. Seems to be treated it as the value -106 instead of 150 allowing one to read data outside of the array (also note that apparently negative indices are also allowed and also grab garbage data).

takagi:
Fix #3017 .

`cupy.take()` currently returns unexpected results when the data size of an array to take is the same with `n + 1`, where `n` is the maximum integer the type of indices can represent. For example, `n + 1 = 256` for the indices of type `uint8`.

This PR fixes the issue to use larger type to represent index range in the kernel of `cupy.take`.


tests/cupy_tests/indexing_tests/test_indexing.py
```diff
@@ -29,6 +31,15 @@ def test_take_no_axis(self, xp):
         b = xp.array([[10, 5], [3, 20]])
         return a.take(b)
 
+    # see cupy#3017
+    @testing.for_dtypes('bBhH')
+    @testing.numpy_cupy_array_equal()
+    def test_take_index_range(self, xp, dtype):
+        iinfo = numpy.iinfo(dtype)
+        a = testing.shaped_arange((iinfo.max + 1,), xp)
+        b = testing.shaped_random((1,), xp, dtype=dtype, scale=iinfo.max)
```
niboshi: Let's use a concrete value instead of `shaped_random`.

tests/cupy_tests/indexing_tests/test_indexing.py
```diff
@@ -29,6 +31,15 @@ def test_take_no_axis(self, xp):
         b = xp.array([[10, 5], [3, 20]])
         return a.take(b)
 
+    # see cupy#3017
+    @testing.for_dtypes('bBhH')
```
niboshi: `for_int_dtypes`?
Char codes should in general be avoided because they depend on environment.

tests/cupy_tests/indexing_tests/test_indexing.py
```diff
@@ -29,6 +31,15 @@ def test_take_no_axis(self, xp):
...
+    @testing.numpy_cupy_array_equal()
+    def test_take_index_range(self, xp, dtype):
```
niboshi: Please include "overflow" or something in the test name.

tests/cupy_tests/indexing_tests/test_indexing.py
takagi: Here I didn't use `for_int_dtypes` because I thought that it allocated too much memory for larger integer types, but, yes, now I think I can broadcast a tiny array to test this.

tests/cupy_tests/indexing_tests/test_indexing.py
takagi: Fixed!

tests/cupy_tests/indexing_tests/test_indexing.py
takagi: Fixed!

tests/cupy_tests/indexing_tests/test_indexing.py
takagi: I fixed the test to use `for_int_dtypes` with broadcasting arrays to keep off actual memory allocation. Tests with 64-bit integer types are skipped because of the error with too large dimensions. Would you take a look at it again?

emcastillo:
Please, can you resolve the conflict? thanks!

takagi:
I've resolved the conflict!

emcastillo:
Jenkins, test this please

emcastillo:
Mmm, how long does this test take in your environment?

takagi:
Around 25 sec for `python -m pytest tests/cupy_tests/indexing_tests/test_indexing.py` .

takagi:
NumPy seems to take much time for the test code, maybe strangely  allocating `(iinfo.max+1,)`-shaped contiguous array in `a.take(b)`:
```
numpy: 15553.754[ms]
cupy: 0.030[ms]
```
```Python
import time
from statistics import mean
import numpy
import cupy

n = 3
dtype = numpy.uint32
for xp in [numpy, cupy]:
    ts = []
    for i in range(n + 1):
	iinfo = numpy.iinfo(dtype)
	a = xp.broadcast_to(xp.ones(1), (iinfo.max + 1,))
	b = xp.array([0], dtype=dtype)
	t = time.time()
	a.take(b)
	if xp is cupy:
            cupy.cuda.runtime.deviceSynchronize()
	elapsed_time = (time.time() - t) * 1000
	if i > 0:
            ts = ts + [elapsed_time]
    print('{}: {:.3f}[ms]'.format(xp.__name__, mean(ts)))
```

takagi:
Can I skip the tests for `numpy.int32` and `numpy.uint32` ? Does it make sense?

emcastillo:
Yes, let's skip it.

takagi:
I've modified to skip them. Would you check them as well as the comment?

tests/cupy_tests/indexing_tests/test_indexing.py
```diff
@@ -31,6 +32,22 @@ def test_take_no_axis(self, xp):
...
 
+    # see cupy#3017
+    @testing.for_int_dtypes(no_bool=True)
+    @testing.numpy_cupy_array_equal()
+    def test_take_index_range_overflow(self, xp, dtype):
+        # Skip for too large dimensions
+        if dtype in (numpy.int64, numpy.uint64):
+            return xp.array([])
+        # Skip because NumPy actually allocates a contiguous array in the
+        # `take` below to require much time.
```
emcastillo, line 43:
```suggestion
        # `take` below that takes a long time.
```

takagi: Thanks!

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

PR #3120, opened 2020-02-25
Add `bitwise_not` as alias to `invert`

Harshan01:
Added alias for `invert`. Modified docstring of `invert` to include `bitwise_not` as a note.


niboshi:
Thank you for a PR!
Could you add a test?
I think `assert xp.bitwise_not is xp.invert` for both `numpy` and `cupy` as `xp` should suffice.

niboshi:

niboshi:
:+1:

PR #3122, opened 2020-02-25
Rewrite manipulation tests not to use `numpy_cupy_raises`

asi1024:
Ready for #3098.

PR #3123, opened 2020-02-25
Fix `reshape` to raise `ValueError` for order 'K'

asi1024:
Split from #3122.


niboshi:
Jenkins, test this please

PR #3124, opened 2020-02-25
Histogram update: support range, weights and density

grlee77:
**Description**
This PR updates the `histogram` function to support the `range`, `weights` and `density` kwargs.

It should be easiest to review the 3 commits sequentially as each of them deals with a separate kwarg.

**Details**
The helpers `_get_outer_edges`, `_get_bin_edges` and `_ravel_and_check_weights` are closely based on their NumPy implementations (except that they do not automatically convert array-likes to `ndarray`).

NumPy itself has a fast code path for uniform bins that loops over blocks of ~65k elements using `bincount` and then a second path for non-uniform bins that uses `searchsorted`. I have not made any changes related to those here, but left the current implementation based on `ElementwiseKernel`.

**One question on API:**
In NumPy there is also a `normed` argument, but its use is deprecated, so I did not add it here. My question is whether the argument should still be in the function signature and just raise a `NotImplementedError` if the user tries to use it?

**Potential follow-ups:**
- add `histogramdd` (I have an implementation for this, and can make a follow-up PR)
- add various other bin selectors to `cupy.histogram`. I haven't worked on this but it looks like it would be easy to add.

### DIFF
--- a/cupy/statistics/histogram.py
+++ b/cupy/statistics/histogram.py
@@ -93,7 +93,8 @@ def _get_outer_edges(a, range):
                 'max must be larger than min in range parameter.')
         if not (numpy.isfinite(first_edge) and numpy.isfinite(last_edge)):
             raise ValueError(
-                "supplied range of [{}, {}] is not finite".format(first_edge, last_edge))
+                "supplied range of [{}, {}] is not finite".format(
+                    first_edge, last_edge))
     elif a.size == 0:
         first_edge = 0.0
         last_edge = 1.0
--- a/cupy/statistics/histogram.py
+++ b/cupy/statistics/histogram.py
@@ -34,7 +34,7 @@ _histogram_kernel = core.ElementwiseKernel(
     preamble=_preamble)
 
 
-def histogram(x, bins=10):
+def histogram(x, bins=10, density=False):
     """Computes the histogram of a set of data.
 
     Args:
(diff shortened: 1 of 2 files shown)


cupy/statistics/histogram.py
```diff
@@ -79,6 +81,10 @@ def histogram(x, bins=10):
 
     y = cupy.zeros(bins.size - 1, dtype='l')
     _histogram_kernel(x, bins, bins.size, y)
+
+    if density:
+        db = cupy.array(cupy.diff(bins), float)
```
emcastillo: I think `float` is not specific enough.
dtype='float64' better?

cupy/statistics/histogram.py
```diff
@@ -34,14 +35,111 @@
...
+def histogram(x, bins=10, range=None, density=False):
     """Computes the histogram of a set of data.
 
     Args:
         x (cupy.ndarray): Input array.
         bins (int or cupy.ndarray): If ``bins`` is an int, it represents the
             number of bins. If ``bins`` is an :class:`~cupy.ndarray`, it
             represents a bin edges.
+        range (2-tuple of float, optional): The lower and upper range of the
+            bins.  If not provided, range is simply ``(a.min(), a.max())``.
```
emcastillo:
```suggestion
            bins.  If not provided, range is simply ``(x.min(), x.max())``.
```

cupy/statistics/histogram.py
```diff
@@ -159,10 +208,38 @@ def histogram(x, bins=10, range=None, density=False):
         # TODO(unno): comparison between complex numbers is not implemented
         raise NotImplementedError('complex number is not supported')
 
+    if not isinstance(x, cupy.ndarray):
+        raise ValueError("x must be a cupy.ndarray")
+
+    x, weights = _ravel_and_check_weights(x, weights)
     bin_edges, uniform_bins = _get_bin_edges(x, bins, range)
```
emcastillo:
```suggestion
    bin_edges, _ = _get_bin_edges(x, bins, range)
```
uniform_bins seems to be unused?

cupy/statistics/histogram.py
grlee77: Thanks. It is a separate question whether the first argument's name should be changed from `x` to `a` to match NumPy.

cupy/statistics/histogram.py
grlee77: I have now  removed the variable from _get_bin_edges itself.

emcastillo:
Jenkins, test this please

PR #3126, opened 2020-02-26
Add `PythonFunctionAllocator`

emcastillo:
This allows cupy to perform memory allocation from a given python interface.
I.e. if the pytorch allocator is exposed in python, it can be used through this new `PythonFunctionAllocator` class.

The difference with `CFunctionAllocator` is that the later uses function pointers that are not easily callable from python and limiting the interoperability with other tools.

### DIFF
--- a/cupy/cuda/memory.pxd
+++ b/cupy/cuda/memory.pxd
@@ -88,16 +88,13 @@ cdef class CFunctionAllocator:
 cdef class PythonFunctionAllocatorMemory(BaseMemory):
 
     cdef:
-        object _param
         object _free_func
 
 
 cdef class PythonFunctionAllocator:
 
     cdef:
-        object _param
         object _malloc_func
         object _free_func
-        object _owner
 
     cpdef MemoryPointer malloc(self, size_t size)
--- a/cupy/cuda/memory.pxd
+++ b/cupy/cuda/memory.pxd
@@ -82,3 +82,22 @@ cdef class CFunctionAllocator:
         object _owner
 
     cpdef MemoryPointer malloc(self, size_t size)
+
+
+@cython.no_gc
+cdef class PythonFunctionAllocatorMemory(BaseMemory):
+
+    cdef:
+        object _param
+        object _free_func
+
+
+cdef class PythonFunctionAllocator:
+
+    cdef:
+        object _param
+        object _malloc_func
+        object _free_func
+        object _owner
+
+    cpdef MemoryPointer malloc(self, size_t size)
--- a/cupy/cuda/memory.pxd
+++ b/cupy/cuda/memory.pxd
@@ -84,7 +84,6 @@ cdef class CFunctionAllocator:
     cpdef MemoryPointer malloc(self, size_t size)
 
 
-@cython.no_gc
 cdef class PythonFunctionAllocatorMemory(BaseMemory):
 
     cdef:
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1470,6 +1470,10 @@ cdef class PythonFunctionAllocator:
     ``free(int, int)`` with no return, taking the pointer to the
     allocated memory and the device id on which the memory was allocated.
 
+    If the external memory management supports asynchronous operations,
+    the current CuPy stream can be retrieved inside ``malloc_func`` and
+    ``free_func`` by calling :func:`cupy.cuda.get_current_stream()`.
+
     Args:
         malloc_func (function): *malloc* function to be called.
         free_func (function): *free* function to be called.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1472,7 +1472,8 @@ cdef class PythonFunctionAllocator:
 
     If the external memory management supports asynchronous operations,
     the current CuPy stream can be retrieved inside ``malloc_func`` and
-    ``free_func`` by calling :func:`cupy.cuda.get_current_stream()`.
+    ``free_func`` by calling :func:`cupy.cuda.get_current_stream()`. To
+    use external streams, wrap them with :func:`cupy.cuda.ExternalStream`.
 
     Args:
         malloc_func (function): *malloc* function to be called.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1439,59 +1439,49 @@ cdef class CFunctionAllocator:
 @cython.no_gc
 cdef class PythonFunctionAllocatorMemory(BaseMemory):
 
-    def __init__(self, size_t size, object param,
-                 malloc_func, free_func,
+    def __init__(self, size_t size, malloc_func, free_func,
                  int device_id):
-        self._param = param
         self._free_func = free_func
         self.device_id = device_id
         self.size = size
         self.ptr = 0
         if size > 0:
-            self.ptr = malloc_func(param, size, device_id)
+            self.ptr = malloc_func(size, device_id)
 
     def __dealloc__(self):
         if self.ptr:
-            self._free_func(self._param, self.ptr, self.device_id)
+            self._free_func(self.ptr, self.device_id)
 
 
 cdef class PythonFunctionAllocator:
 
     """Allocator with python functions to perform memory allocation.
 
-    This allocator keeps a *param* object along with functions
-    corresponding to *malloc* and *free*, delegating the actual allocation to
-    external sources while only handling the timing of the resource allocation
-    and deallocation.
+    This allocator keeps functions corresponding to *malloc* and *free*,
+    delegating the actual allocation to external sources while only
+    handling the timing of the resource allocation and deallocation.
 
-    *malloc* should follow the signature ``malloc(object, int, int) -> int``
+    *malloc* should follow the signature ``malloc(int, int) -> int``
     returning the pointer to the allocated memory given the *param* object,
     the number of bytes to allocate and the device id on which the
     allocation should take place.
 
     Similarly, *free* should follow the signature
-    ``free(object, int, int)`` with no return, taking the *param* object,
-    the pointer to the allocated memory and the device id on which the
-    memory was allocated.
+    ``free(int, int)`` with no return, taking the pointer to the
+    allocated memory and the device id on which the memory was allocated.
 
     Args:
-        param (object): *param* object to be passed on allocation and deletion.
         malloc_func (function): *malloc* function to be called.
         free_func (function): *free* function to be called.
-        owner (object): Reference to the owner object to keep the param and
-            the functions alive.
 
     """
 
-    def __init__(self, object param, malloc_func,
-                 free_func, object owner):
-        self._param = param
+    def __init__(self, malloc_func, free_func):
         self._malloc_func = malloc_func
         self._free_func = free_func
-        self._owner = owner
 
     cpdef MemoryPointer malloc(self, size_t size):
         mem = PythonFunctionAllocatorMemory(
-            size, self._param, self._malloc_func,
+            size, self._malloc_func,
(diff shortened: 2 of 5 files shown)


cupy/cuda/memory.pxd
```diff
@@ -82,3 +82,21 @@ cdef class CFunctionAllocator:
...
+    cdef:
+        object _param
+        object _free_func
+
+
+cdef class PythonFunctionAllocator:
+
+    cdef:
+        object _param
+        object _malloc_func
```
emcastillo: not really sure about these types

niboshi:
As python callables can capture anything, you don't need `param`.
Similarly, `owner` is not necessary, as you're not dealing with any C objects.

leofang:
cc: @jakirkham @kkraus14 @gmarkall in case you're interested.

emcastillo:
```numpy.logaddexp(1j, 1j)
TypeError: ufunc 'logaddexp' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''```

If it's not supported by Numpy, are tests passing?

**Comment by kkraus14** [comment]
@leofang thanks for the ping

@emcastillo Any thoughts about exposing an optional cuda stream here? RMM has the ability to asynchronously allocate memory on a given stream and wondering if it would be applicable here.

**Comment by emcastillo** [comment]
> @emcastillo Any thoughts about exposing an optional cuda stream here? RMM has the ability to asynchronously allocate memory on a given stream and wondering if it would be applicable here.

The allocator object can get a stream from cupy and use it for the actual memory allocation management. Could you describe a use-case where this is not enough? Thanks for the feedback!

**Comment by kkraus14** [comment]
> The allocator object can get a stream from cupy and use it for the actual memory allocation management. Could you describe a use-case where this is not enough? Thanks for the feedback!

I wasn't clear on if this lived underneath or above the Stream management. If the idea is to inherit from the existing Stream management in CuPy then this looks good to me. Would definitely welcome review from @jakirkham and @gmarkall though 😄

**Comment by leofang** [comment]
> The allocator object can get a stream from cupy and use it for the actual memory allocation management.

@emcastillo How is this done? Can you elaborate? Did you mean the allocator object needs to look up the stream from CuPy's thread-local storage?

**Comment by emcastillo** [comment]
I am having a similar discussion on the pytorch side.
What I thought is that the python allocator itself, gets the current stream, or creates a new one, and it can pass the stream handle to the underlying allocator. It should be the responsibility of the PythonAllocator code to do this. But maybe I am missing something here.

**Comment by niboshi** [comment]
>I am having a similar discussion on the pytorch side.

Can you provide a link?

**Comment by leofang** [comment]
> What I thought is that the python allocator itself, gets the current stream, or creates a new one, and it can pass the stream handle to the underlying allocator. It should be the responsibility of the PythonAllocator code to do this.

So do I get it right that the provided functions `malloc_func` and `free_func` should handle streams?

**Comment by emcastillo** [comment]
The link was in the PR references
https://github.com/pytorch/pytorch/pull/33860

> So do I get you right that the provided functions malloc_func and free_func should handle streams?
PyTorch uses a stream aware allocator so we need to pass a stream to the malloc & free.
Right now it is hidden in the implementation as raw_alloc just selects the current stream.
Pytorch also has a pool of streams so the management is a bit trickier than I thought.

I think that the cupy actual stream can be retrieved inside the python allocator function, and it should be passed to the pytorch allocator. For other allocators that do not require a stream this shouldn't be a problem.

**Comment by leofang** [comment]
> However, I think we need a mechanism for converting a cupy stream to a pytorch stream and the opposite. We may allocate memory inside cupy using the torch allocator, but doing computations in both cupy and torch with different streams it's not safe. Probably retrieving the pytorch stream and creating a cupy stream from the pointer should be an easy solution too.

This is closely related to ongoing discussions, scattered in the following threads (oops😅): 
- numba/numba#4797 (I showed how to do a CuPy->Numba stream conversion)
- numba/numba#4933 (interoperability with TensorFlow which also has its own stream computation) 
- numba/numba#5162 (is it possible to solve stream issues via `__cuda_array_interface__`)

@jakirkham and @gmarkall should have valuable comments on this.

**Comment by leofang** [comment]
Just FYI: @gmarkall has an ongoing effort on a memory manager plugin for Numba: https://github.com/gmarkall/nbep-7. There the stream needs to be passed explicitly to the allocation and deallocation routines. The async behavior is not the focus, though.

File: cupy/cuda/memory.pyx [hunk 2]
```diff
@@ -1434,3 +1434,54 @@ cdef class CFunctionAllocator:
...
+
+    *malloc* should follow the signature ``malloc(int, int) -> int``
+    returning the pointer to the allocated memory given the *param* object,
+    the number of bytes to allocate and the device id on which the
+    allocation should take place.
+
+    Similarly, *free* should follow the signature
+    ``free(int, int)`` with no return, taking the pointer to the
+    allocated memory and the device id on which the memory was allocated.
+
```
leofang: I wonder if this could make it clearer?
```suggestion

    If the external memory management supports asynchronous operations,
    the current CuPy stream can be retrieved inside ``malloc_func`` and
    ``free_func`` by calling :func:`cupy.cuda.get_current_stream()`.

```

emcastillo: Thanks!!

cupy/cuda/memory.pxd
leofang: I think they're fine.

cupy/cuda/memory.pyx
```diff
@@ -1434,3 +1434,58 @@ cdef class CFunctionAllocator:
...
+    If the external memory management supports asynchronous operations,
+    the current CuPy stream can be retrieved inside ``malloc_func`` and
+    ``free_func`` by calling :func:`cupy.cuda.get_current_stream()`.
```
leofang: Not sure if I'm being chattering 🤣 
```suggestion
    ``free_func`` by calling :func:`cupy.cuda.get_current_stream()`. To
    use external streams, wrap them with :func:`cupy.cuda.ExternalStream`.
```

gmarkall:
Thanks for mentioning this @leofang, and apologies for the delay in my comment - I had missed the notification. I'm thinking about how this could be used to implement allocation using the [RAPIDS Memory Manager (RMM)](https://github.com/rapidsai/rmm). RMM provides a `DeviceBuffer` object, which allocates the memory on initialization and frees the memory when it is deleted. I can see a couple of ways one might use the `PythonFunctionAllocator` interface to support these objects, but I'm not too sure which way you'd intend the interface to be used:

- An implementation of a `malloc_func` for RMM which keeps the active `DeviceBuffer` objects somewhere to keep them alive whilst they're in use (e.g. in a dict), and the corresponding `free_func` deletes them from the dict.
- Alternatively, is it expected that one might implement a custom `PythonFunctionAllocator` class? If so, for RMM the `malloc` method could perhaps return something `PythonFunctionAllocatorMemory`-like object which holds a reference to the `DeviceBuffer`, so that when the object is deallocated the `DeviceBuffer` is cleaned up too. The `_free_func` would probably do nothing in this case.

Do you have a recommendation of which you intended / makes more sense?

> Just FYI: @gmarkall has an ongoing effort on a memory manager plugin for Numba: https://github.com/gmarkall/nbep-7. There the stream needs to be passed explicitly to the allocation and deallocation routines. The async behavior is not the focus, though.

Numba doesn't have anything at present that would allow an allocation to take place asynchronously, even using its internal memory management - the `stream` parameters existed already in Numba in some places in its memory allocation code, but were unused. So, I removed the `stream` parameters from the NBEP 7 PR ( https://gmarkall.github.io/numba-nbep7-docs/proposals/external-memory-management.html / https://github.com/numba/numba/pull/5328 ) because they did nothing but misleadingly imply the possibility of async behaviour. I was planning to iterate on the specification once the current version has been accepted, and once there's an implementation of a plugin or two to learn from before adding async behaviour.

As a side comment, I'm not sure there's an analagous concept in Numba of the current stream - device arrays have a default stream on which async operations occur if a stream is not otherwise specified, but there's no general "current" stream in Numba (I don't think this is an issue in any way, but thought the observation may be of interest).

leofang:
Thanks for your nice, detailed comments, Graham! 🙂 I'll revisit the bulk of it later, and for now let me quickly comment on the current stream thing. 

Yeah, it's a very interesting concept in CuPy that does not exist in Numba. (I guess PyTorch has it, too? Not sure...) It's made possible because of the need of using streams as a context manager:
```python
stream = cp.cuda.stream()
with stream:
    # do some cupy operations on stream
# and we are back to the default/null stream
```
Internally, the current stream is stored in the thread-local storage for lookup:
https://github.com/cupy/cupy/blob/81118799de3ae1df8f3eca9855d414bf1abd45d8/cupy/cuda/stream.pyx#L49-L56
(One can either retrieve the `stream` object or its raw pointer, see the code snippets around the linked code, but you get the point.) I found this feature quite handy. 

In addition to streams, I think in CuPy devices can also be context managers.

gmarkall:
> Yeah, it's a very interesting concept in CuPy that does not exist in Numba. (I guess PyTorch has it, too? Not sure...) It's made possible because of the need of using streams as a context manager:

Thanks for the explanation! Maybe this would be an interesting idea for Numba too.

leofang:
Hi @gmarkall, sorry that I forgot to revisit your comment:

> I can see a couple of ways one might use the `PythonFunctionAllocator` interface to support these objects, but I'm not too sure which way you'd intend the interface to be used:
> 
> * An implementation of a `malloc_func` for RMM which keeps the active `DeviceBuffer` objects somewhere to keep them alive whilst they're in use (e.g. in a dict), and the corresponding `free_func` deletes them from the dict.
> * Alternatively, is it expected that one might implement a custom `PythonFunctionAllocator` class? If so, for RMM the `malloc` method could perhaps return something `PythonFunctionAllocatorMemory`-like object which holds a reference to the `DeviceBuffer`, so that when the object is deallocated the `DeviceBuffer` is cleaned up too. The `_free_func` would probably do nothing in this case.
> 
> Do you have a recommendation of which you intended / makes more sense?

@emcastillo might have a better answer since he pushed for the necessary changes in CuPy and PyTorch. But to me it sounds like if you need to do some bookkeeping, a mixture of your two solutions can be used: define a new class that holds all necessary references for maintaining the lifetime, and that has the malloc and free methods. See how Emilio did it in the added test with a hypothetical `PythonAllocator` object.

leofang:
Jenkins, test this please

cupy/cuda/memory.pxd
```diff
@@ -82,3 +82,19 @@ cdef class CFunctionAllocator:
         object _owner
 
     cpdef MemoryPointer malloc(self, size_t size)
+
+
+@cython.no_gc
```
kmaehashi, line 87: I think `no_gc` is not safe when we keep a reference to an arbitrary object.

kmaehashi:
LGTM except comments!

cupy/cuda/memory.pxd
emcastillo, line 87:
```suggestion
```

cupy/cuda/memory.pyx
```diff
@@ -1434,3 +1434,59 @@ cdef class CFunctionAllocator:
         mem = CFunctionAllocatorMemory(size, self._param, self._malloc_func,
                                        self._free_func, device.get_device_id())
         return MemoryPointer(mem, 0)
+
+
+@cython.no_gc
```
emcastillo, line 1439:
```suggestion
```

emcastillo:
removed the nogc, PTAL

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

PR #3128, opened 2020-02-27
use `time.process_time` instead of `time.clock`

### LINKED ISSUE #3059: time.clock was removed in Python 3.8
https://github.com/python/cpython/pull/13270 . Given that the project is Python 3 only it will be better to use the alternatives. There is https://github.com/cupy/cupy/issues/2599 for Python 3.8 but not sure why it didn't error out.

https://github.com/cupy/cupy/blob/8d923719ced31fdff4a42702d395ec87a7e98045/cupy/random/generator.py#L681

rushabh-v:
closes #3059

### DIFF
--- a/cupy/random/generator.py
+++ b/cupy/random/generator.py
@@ -685,7 +685,7 @@ class RandomState(object):
                 seed_str = binascii.hexlify(os.urandom(8))
                 seed = int(seed_str, 16)
             except NotImplementedError:
-                seed = (time.clock() * 1000000) % _UINT64_MAX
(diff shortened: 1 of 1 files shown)


niboshi:
`time.process_time` is local to the curren process. I think it's not appropriate for a seed.
How about `time.time`?

rushabh-v:
can you take a look back, please @niboshi ?

niboshi:
Thanks, looks good to me.

takagi:
pfnCI, test this please.

Issue #2717, opened 2019-11-26
Document strides incompatibility with NumPy


niboshi:
We do not guarantee compatibility of strides returned by CuPy functions. That should be documented.

kmaehashi:
Fixed via #7344

PR #3135, opened 2020-02-28
add argwhere function

### LINKED ISSUE #3114: [FEA] Implement cupy.argwhere
A CuPy equivalent of numpy.argwhere would be a nice addition to the library. This has been mentioned in #2783 , but it's likely worth breaking out into a separate issue for tracking purposes.

rushabh-v:
closes #3114

### DIFF
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -233,8 +233,8 @@ def argwhere(a):
     if a.ndim == 0:
         a = cupy.shape_base.atleast_1d(a)
         # then remove the added dimension
-        return argwhere(a)[:,:0]
-    a = cupy.asarray(cupy.nonzero(a), dtype = int)
+        return argwhere(a)[:, :0]
+    a = cupy.asarray(cupy.nonzero(a), dtype=int)
     return cupy.transpose(a)
 
 
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -212,6 +212,32 @@ def where(condition, x=None, y=None):
     return _where_ufunc(condition.astype('?'), x, y)
 
 
+def argwhere(a):
+    """Return the indices of the elements that are non-zero.
+
+    Returns a (N, ndim) dimantional array containing the
+    indices of the non-zero elements. Where `N` is number of
+    non-zero elements and `ndim` is dimention of the given array.
+
+    Args:
+        a (cupy.ndarray): array
+
+    Returns:
+        cupy.array: Indices of elements that are non-zero.
+
+    .. seealso:: :func:`numpy.argwhere`
+
+    """
+    util.check_array(a, arg_name='a')
+    # promote 0d to 1d because argwhere doesn't work well with 0d
+    if a.ndim == 0:
+        a = cupy.shape_base.atleast_1d(a)
+        # then remove the added dimension
+        return argwhere(a)[:,:0]
+    a = cupy.asarray(cupy.nonzero(a), dtype = int)
+    return cupy.transpose(a)
+
+
 # This is to allow using the same kernels for all dtypes, ints & floats
 # as nan is a special case
 _preamble = '''
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -234,7 +234,7 @@ def argwhere(a):
     util.check_array(a, arg_name='a')
     if a.ndim == 0:
         warnings.warn(
-            'calling nonzero on 0d arrays is deprecated',
+            'calling argwhere on 0d arrays is deprecated',
             DeprecationWarning)
     return _ndarray_argwhere(a)
 
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -3,6 +3,7 @@ from cupy import core
 from cupy.core import fusion
 from cupy import util
 
+from  cupy.core._routines_indexing import _ndarray_argwhere
 from cupy.core import _routines_statistics as _statistics
 
 
@@ -229,7 +230,7 @@ def argwhere(a):
 
     """
     util.check_array(a, arg_name='a')
-    return a.argwhere()
+    return _ndarray_argwhere(a)
 
 
 # This is to allow using the same kernels for all dtypes, ints & floats
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -3,8 +3,10 @@ from cupy import core
 from cupy.core import fusion
 from cupy import util
 
-from  cupy.core._routines_indexing import _ndarray_argwhere
+from cupy.core._routines_indexing import _ndarray_argwhere
 from cupy.core import _routines_statistics as _statistics
+
+import warnings
 
 
 def argmax(a, axis=None, dtype=None, out=None, keepdims=False):
@@ -230,6 +232,10 @@ def argwhere(a):
 
     """
     util.check_array(a, arg_name='a')
+    if a.ndim == 0:
+        warnings.warn(
+            'calling nonzero on 0d arrays is deprecated',
+            DeprecationWarning)
     return _ndarray_argwhere(a)
 
 
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -229,13 +229,7 @@ def argwhere(a):
 
     """
     util.check_array(a, arg_name='a')
-    # promote 0d to 1d because argwhere doesn't work well with 0d
-    if a.ndim == 0:
-        a = cupy.shape_base.atleast_1d(a)
-        # then remove the added dimension
-        return argwhere(a)[:, :0]
-    a = cupy.asarray(cupy.nonzero(a), dtype=int)
-    return cupy.transpose(a)
+    return a.argwhere()
 
 
 # This is to allow using the same kernels for all dtypes, ints & floats
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -3,7 +3,7 @@ from cupy import core
 from cupy.core import fusion
 from cupy import util
 
-from cupy.core._routines_indexing import _ndarray_argwhere
+from cupy.core import _routines_indexing as _indexing
 from cupy.core import _routines_statistics as _statistics
 
 import warnings
@@ -226,7 +226,7 @@ def argwhere(a):
         a (cupy.ndarray): array
 
     Returns:
-        cupy.array: Indices of elements that are non-zero.
+        cupy.ndarray: Indices of elements that are non-zero.
 
     .. seealso:: :func:`numpy.argwhere`
 
@@ -236,7 +236,7 @@ def argwhere(a):
         warnings.warn(
             'calling argwhere on 0d arrays is deprecated',
             DeprecationWarning)
-    return _ndarray_argwhere(a)
+    return _indexing._ndarray_argwhere(a)
 
 
 # This is to allow using the same kernels for all dtypes, ints & floats
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -49,6 +49,16 @@ cdef _ndarray_setitem(ndarray self, slices, value):
 
 
 cdef tuple _ndarray_nonzero(ndarray self):
+    cdef int ndim
+    cdef ndarray dst = _ndarray_argwhere(self)
+    ndim = len(dst)
+    if ndim > 1:
+        return tuple([dst[i] for i in range(ndim)])
+    elif ndim == 1:
+        return (dst,)
+
+
+cdef ndarray _ndarray_argwhere(ndarray self):
     cdef Py_ssize_t count_nonzero
     cdef int ndim
     dtype = numpy.int64
@@ -62,18 +72,18 @@ cdef tuple _ndarray_nonzero(ndarray self):
         count_nonzero = int(scan_index[-1])  # synchronize!
     ndim = max(<int>self._shape.size(), 1)
     if count_nonzero == 0:
-        return (ndarray((0,), dtype=dtype),) * ndim
+        return ndarray((0, ndim), dtype=dtype)
 
     if ndim <= 1:
-        dst = ndarray((count_nonzero,), dtype=dtype)
+        dst = ndarray((count_nonzero, 1), dtype=dtype)
         _nonzero_kernel_1d(nonzero, scan_index, dst)
-        return dst,
+        return dst
     else:
         nonzero.shape = self.shape
         scan_index.shape = self.shape
-        dst = ndarray((ndim, count_nonzero), dtype=dtype)
+        dst = ndarray((count_nonzero, ndim), dtype=dtype)
         _nonzero_kernel(nonzero, scan_index, dst)
-        return tuple([dst[i] for i in range(ndim)])
+        return dst
 
 
 cdef _ndarray_scatter_add(ndarray self, slices, value):
@@ -352,7 +362,7 @@ cdef ndarray _simple_getitem(ndarray a, list slice_list):
 
 _nonzero_kernel_1d = ElementwiseKernel(
     'T src, S index', 'raw S dst',
-    'if (src != 0) dst[index - 1] = i',
+    'if (src != 0) dst[index - 1][0] = i',
     'nonzero_kernel_1d')
 
 
@@ -362,7 +372,7 @@ _nonzero_kernel = ElementwiseKernel(
     if (src != 0){
         for(int j = 0; j < _ind.ndim; j++){
             ptrdiff_t ind[] = {j, index - 1};
-            dst[ind] = _ind.get()[j];
+            dst[:, ind] = _ind.get()[:, j];
         }
     }''',
     'nonzero_kernel',
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -355,19 +355,13 @@ cdef ndarray _simple_getitem(ndarray a, list slice_list):
     return v
 
 
-_nonzero_kernel_1d = ElementwiseKernel(
-    'T src, S index', 'raw S dst',
-    'if (src != 0) dst[index - 1][0] = i',
-    'nonzero_kernel_1d')
-
-
 _nonzero_kernel = ElementwiseKernel(
     'T src, S index', 'raw S dst',
     '''
     if (src != 0){
         for(int j = 0; j < _ind.ndim; j++){
-            ptrdiff_t ind[] = {j, index - 1};
-            dst[:, ind] = _ind.get()[:, j];
+            ptrdiff_t ind[] = {index - 1, j};
+            dst[ind] = _ind.get()[j];
         }
     }''',
     'nonzero_kernel',
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -74,11 +74,16 @@ cpdef ndarray _ndarray_argwhere(ndarray self):
     if count_nonzero == 0:
         return ndarray((0, ndim), dtype=dtype)
 
-    nonzero.shape = self.shape
-    scan_index.shape = self.shape
-    dst = ndarray((count_nonzero, ndim), dtype=dtype)
-    _nonzero_kernel(nonzero, scan_index, dst)
-    return dst
+    if ndim <= 1:
+        dst = ndarray((count_nonzero, 1), dtype=dtype)
+        _nonzero_kernel_1d(nonzero, scan_index, dst)
+        return dst
+    else:
+        nonzero.shape = self.shape
+        scan_index.shape = self.shape
+        dst = ndarray((count_nonzero, ndim), dtype=dtype)
+        _nonzero_kernel(nonzero, scan_index, dst)
+        return dst
 
 
 cdef _ndarray_scatter_add(ndarray self, slices, value):
@@ -353,6 +358,12 @@ cdef ndarray _simple_getitem(ndarray a, list slice_list):
     # TODO(niboshi): Confirm update_x_contiguity flags
     v._set_shape_and_strides(shape, strides, True, True)
     return v
+
+
+_nonzero_kernel_1d = ElementwiseKernel(
+    'T src, S index', 'raw S dst',
+    'if (src != 0) dst[index - 1] = i',
+    'nonzero_kernel_1d')
 
 
 _nonzero_kernel = ElementwiseKernel(
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -74,16 +74,11 @@ cdef ndarray _ndarray_argwhere(ndarray self):
     if count_nonzero == 0:
         return ndarray((0, ndim), dtype=dtype)
 
-    if ndim <= 1:
-        dst = ndarray((count_nonzero, 1), dtype=dtype)
-        _nonzero_kernel_1d(nonzero, scan_index, dst)
-        return dst
-    else:
-        nonzero.shape = self.shape
-        scan_index.shape = self.shape
-        dst = ndarray((count_nonzero, ndim), dtype=dtype)
-        _nonzero_kernel(nonzero, scan_index, dst)
-        return dst
+    nonzero.shape = self.shape
+    scan_index.shape = self.shape
+    dst = ndarray((count_nonzero, ndim), dtype=dtype)
+    _nonzero_kernel(nonzero, scan_index, dst)
+    return dst
 
 
 cdef _ndarray_scatter_add(ndarray self, slices, value):
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -54,7 +54,7 @@ cdef tuple _ndarray_nonzero(ndarray self):
     ndim = len(dst)
     if ndim > 1:
         return tuple([dst[:, i] for i in range(ndim)])
-    elif ndim == 1:
+    else:
         return (dst,)
 
 
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -53,12 +53,12 @@ cdef tuple _ndarray_nonzero(ndarray self):
     cdef ndarray dst = _ndarray_argwhere(self)
     ndim = len(dst)
     if ndim > 1:
-        return tuple([dst[i] for i in range(ndim)])
+        return tuple([dst[:, i] for i in range(ndim)])
     elif ndim == 1:
         return (dst,)
 
 
-cdef ndarray _ndarray_argwhere(ndarray self):
+cpdef ndarray _ndarray_argwhere(ndarray self):
     cdef Py_ssize_t count_nonzero
     cdef int ndim
     dtype = numpy.int64
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -51,11 +51,11 @@ cdef _ndarray_setitem(ndarray self, slices, value):
 cdef tuple _ndarray_nonzero(ndarray self):
     cdef int ndim
     cdef ndarray dst = _ndarray_argwhere(self)
-    ndim = len(dst)
-    if ndim > 1:
+    ndim = self.ndim
+    if ndim >= 1:
         return tuple([dst[:, i] for i in range(ndim)])
     else:
-        return (dst,)
+        return ndarray((0,), dtype=numpy.int64),
 
 
 cpdef ndarray _ndarray_argwhere(ndarray self):
--- a/tests/cupy_tests/sorting_tests/test_search.py
+++ b/tests/cupy_tests/sorting_tests/test_search.py
@@ -309,6 +309,7 @@ class TestArgwhereZeroDimension(unittest.TestCase):
         array = xp.array(self.array, dtype=dtype)
         return xp.nonzero(array)
 
+
 @testing.gpu
 class TestNanArgMin(unittest.TestCase):
 
--- a/tests/cupy_tests/sorting_tests/test_search.py
+++ b/tests/cupy_tests/sorting_tests/test_search.py
@@ -296,18 +296,14 @@ class TestArgwhere(unittest.TestCase):
 
 
 @testing.parameterize(
-    {'array': numpy.array(0)},
-    {'array': numpy.array(1)},
+    {'array': cupy.array(1)},
 )
 @testing.gpu
-@testing.with_requires('numpy>=1.17.0')
 class TestArgwhereZeroDimension(unittest.TestCase):
 
-    @testing.for_all_dtypes()
-    @testing.numpy_cupy_raises(accept_error=DeprecationWarning)
-    def test_argwhere(self, xp, dtype):
-        array = xp.array(self.array, dtype=dtype)
-        return xp.nonzero(array)
+    def test_argwhere(self):
+        with testing.assert_warns(DeprecationWarning):
+            return cupy.nonzero(self.array)
 
 
 @testing.gpu
--- a/tests/cupy_tests/sorting_tests/test_search.py
+++ b/tests/cupy_tests/sorting_tests/test_search.py
@@ -295,6 +295,20 @@ class TestArgwhere(unittest.TestCase):
         return xp.argwhere(array)
 
 
+@testing.parameterize(
+    {'array': numpy.array(0)},
+    {'array': numpy.array(1)},
+)
+@testing.gpu
+@testing.with_requires('numpy>=1.17.0')
+class TestArgwhereZeroDimension(unittest.TestCase):
+
+    @testing.for_all_dtypes()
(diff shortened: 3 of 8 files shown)


rushabh-v:
I will add the tests soon.

rushabh-v:
I have run `pytest` on the new tests. And 100% of the tests are passing.

jakirkham:
Thanks for working on this @rushabh-v! 😄

@kmaehashi, would be possible to get a review of this? 😉

beckernick:
💯 thanks for working on this @rushabh-v . If you have the bandwidth, it would be great to include a quick and dirty performance benchmark of this compared to `numpy.argwhere` on a few different size/dimension arrays.

cupy/_sorting/search.py
```diff
@@ -212,6 +212,32 @@ def where(condition, x=None, y=None):
...
+    .. seealso:: :func:`numpy.argwhere`
+
+    """
+    util.check_array(a, arg_name='a')
+    # promote 0d to 1d because argwhere doesn't work well with 0d
+    if a.ndim == 0:
+        a = cupy.shape_base.atleast_1d(a)
+        # then remove the added dimension
+        return argwhere(a)[:, :0]
+    a = cupy.asarray(cupy.nonzero(a), dtype=int)
```
jakirkham: Actually I'd be tempted to do the opposite. Namely pull the `nonzero` code over here and then rewrite `nonzero` in terms of `argwhere`. Here's [how we did this in Dask for example]( https://github.com/dask/dask/blob/dfcc09c6eb48d59e391e593408c12181f6550f31/dask/array/routines.py#L1244-L1249 ).

cupy/_sorting/search.py
```diff
@@ -212,6 +212,32 @@ def where(condition, x=None, y=None):
...
+
+    .. seealso:: :func:`numpy.argwhere`
+
+    """
+    util.check_array(a, arg_name='a')
+    # promote 0d to 1d because argwhere doesn't work well with 0d
+    if a.ndim == 0:
+        a = cupy.shape_base.atleast_1d(a)
+        # then remove the added dimension
+        return argwhere(a)[:, :0]
```
beckernick: This code doesn't look to be tested currently, since there are no scalars as a test case. I'd suggest adding a test case or disallowing 0d arrays and raising a warning

beckernick: I'm also not sure if this code will work. Is `atleast_1d` in this namespace?

cupy/_sorting/search.py
asi1024: Current implementation causes redundant memory copy. How about using `_nonzero_kernel` in `_routines_indexing.pyx` directly?

tests/cupy_tests/sorting_tests/test_search.py
```diff
@@ -278,6 +278,23 @@ def test_flatnonzero(self, xp, dtype):
...
+    {'array': numpy.random.randn(3, 2, 4)},
+    {'array': numpy.empty((0,))},
+    {'array': numpy.empty((0, 2))},
+    {'array': numpy.empty((0, 2, 0))},
+)
+@testing.gpu
+class TestArgwhere(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_list_equal()
```
asi1024: The return value is not a list of arrays, but a single array.
```suggestion
    @testing.numpy_cupy_array_equal()
```

cupy/_sorting/search.py
rushabh-v: Do you mean I should implement `argwhere` in `_routines_indexing.pyx` as `nonzero` is implemented there?

rushabh-v: Do you mean implementing `argwhere` in  `_routines_indexing.pyx` like `nonzero` is implemented there?

rushabh-v: And I am sorry for the late response @jakirkham

cupy/_sorting/search.py
rushabh-v: Yes, nonzero also warns for 0d arrays. So IMO going with the same convention would be better.

cupy/_sorting/search.py
jakirkham: No worries. Not late at all 😄

Yeah that's my thinking 🙂

Would be tempted to just rename [`_ndarray_nonzero`]( https://github.com/cupy/cupy/blob/7cf5fef1f1f568b04180b20b789d2a08fdd2b9f0/cupy/core/_routines_indexing.pyx#L51 ) to `_ndarray_argwhere` and tweak [the return to be `dst`]( https://github.com/cupy/cupy/blob/7cf5fef1f1f568b04180b20b789d2a08fdd2b9f0/cupy/core/_routines_indexing.pyx#L76 ). Could still keep `_ndarray_nonzero` around as a wrapper that just calls `_ndarray_argwhere`. Should be straightforward to use `_ndarray_argwhere` here. If that makes sense 😉

Akifumi may have more thoughts 🙂

cupy/core/_routines_indexing.pyx
```diff
@@ -49,6 +49,16 @@ cdef _ndarray_setitem(ndarray self, slices, value):
 
 
 cdef tuple _ndarray_nonzero(ndarray self):
+    cdef int ndim
+    ndim = max(<int>self._shape.size(), 1)
```
jakirkham: Think it's just this

```suggestion
    ndim = self.ndim
```

jakirkham: Alternatively this could just be `len(dst)` since this should already be baked into the result.

jakirkham:
This is looking pretty good! Thanks for all of the work here @rushabh-v! 😄

rushabh-v:
I am still working on it. The tests are failing.

cupy/core/_routines_indexing.pyx
```diff
@@ -62,7 +72,7 @@ cdef tuple _ndarray_nonzero(ndarray self):
         count_nonzero = int(scan_index[-1])  # synchronize!
     ndim = max(<int>self._shape.size(), 1)
     if count_nonzero == 0:
-        return (ndarray((0,), dtype=dtype),) * ndim
+        return ndarray((0,ndim), dtype=dtype)
```
jakirkham: CI is failing due to a linter error here. Would suggest adding a space. Should fix it 🙂

```suggestion
        return ndarray((0, ndim), dtype=dtype)
```

jakirkham:
FWIW if you are comfortable [allowing edits from maintainers]( https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests ), there should be a dialog that pops up on each of these suggested changes that allows you to accept them. Might be easier than making them by hand. Just a suggestion 🙂

cupy/_sorting/search.py
```diff
@@ -212,6 +212,26 @@ def where(condition, x=None, y=None):
...
+        a (cupy.ndarray): array
+
+    Returns:
+        cupy.array: Indices of elements that are non-zero.
+
+    .. seealso:: :func:`numpy.argwhere`
+
+    """
+    util.check_array(a, arg_name='a')
+    return a.argwhere()
```
jakirkham: There's not a method for this one. Would just use the function `_ndarray_argwhere` directly.

```suggestion
    return _ndarray_argwhere(a)
```

<br>
<br>

Would need this `import` at the top of the file.

```python
from  cupy.core._routines_indexing import _ndarray_argwhere
````

File: cupy/core/_routines_indexing.pyx [hunk 7]
```diff
@@ -49,6 +49,16 @@ cdef _ndarray_setitem(ndarray self, slices, value):
 
 
 cdef tuple _ndarray_nonzero(ndarray self):
+    cdef int ndim
+    dst = _ndarray_argwhere(self)
```
jakirkham: Since we know the type, this should generate more efficient C code in the end.

```suggestion
    cdef ndarray dst = _ndarray_argwhere(self)
```

cupy/_sorting/search.py
rushabh-v: I have declared `cpdef ndarray argwhere(self)` inside `ndarray` in `core.pxd` there.

cupy/core/_routines_indexing.pyx
rushabh-v: By failures I meant the pytest failures. The results for numpy and cupy are different.

rushabh-v: below is traceback to one test failure,

```
self = <cupy.testing._bundle.TestArgwhere_param_0_{array=array([0, 0, 1, 1, 1, 1, 0, 1, 1, 0, 1, 0, 1, 1, 0, 0, 0, 0, 1, 1])} testMethod=test_argwhere>
args = (), kw = {'dtype': <class 'numpy.float64'>}
cupy_result = (array([ 2,  3,  4,  5,  7,  8, 10, 12, 13, 18, 19]),)
cupy_error = None, cupy_tb = None
numpy_result = (array([[ 2],
       [ 3],
       [ 4],
       [ 5],
       [ 7],
       [ 8],
       [10],
       [12],
       [13],
       [18],
       [19]]),)
```

Can you take a look please? @jakirkham

rushabh-v: I even have defined dst as `ndarray(count_nonzero, 1)`. Can't figure out why the resulting dims are (1, count_nonzero). And same is with the array haveing `ndim > 1`. the resulting arrays are transpose of the expected arrays.

cupy/_sorting/search.py
rushabh-v: and it works also when I run pytest.

cupy/core/_routines_indexing.pyx
jakirkham: Yeah I think that is expected. This is what they say about it in the NumPy docs

> `np.argwhere(a)` is the same as `np.transpose(np.nonzero(a))`.

ref: https://docs.scipy.org/doc/numpy/reference/generated/numpy.argwhere.html

jakirkham: The kernel in CuPy may be tailored for the `np.nonzero` case. It might be possible to tweak it to align more closely to `np.argwhere`.

cupy/_sorting/search.py
jakirkham: Yeah I think the point here is that `argwhere` is not a method on NumPy arrays. As CuPy likes to stay closely aligned with NumPy, they probably would prefer to not have this method. Maybe you could make it private if it is easier for you to implement. Though using the kernel directly seems like the safest bet 🙂

cupy/core/_routines_indexing.pyx
rushabh-v: done

tests/cupy_tests/sorting_tests/test_search.py
rushabh-v: Done

cupy/core/_routines_indexing.pyx
```diff
@@ -49,6 +49,16 @@ cdef _ndarray_setitem(ndarray self, slices, value):
 
 
 cdef tuple _ndarray_nonzero(ndarray self):
+    cdef int ndim
+    cdef ndarray dst = _ndarray_argwhere(self)
+    ndim = len(dst)
+    if ndim > 1:
+        return tuple([dst[i] for i in range(ndim)])
```
jakirkham: As `argwhere` is transposed relative to `nonzero`, you may want this to index along the right dimension. Perhaps fixes some of the testing issues you are seeing 🙂

```suggestion
        return tuple([dst[:, i] for i in range(ndim)])
```

cupy/core/_routines_indexing.pyx
jakirkham: Maybe [this comment]( https://github.com/cupy/cupy/pull/3135#discussion_r388183512 ) also helps? 🙂

rushabh-v: (just a note: above failure was of `argwhere` not `nonzero`.)

cupy/core/_routines_indexing.pyx
rushabh-v: The `nonzero` is working perfectly in the case of `ndim > 1`,  because the argwhere is returning the transposed version. the traceback I commented above was of `argwhere`.

cupy/_sorting/search.py
rushabh-v: Okay, I will make that change soon : )

cupy/core/_routines_indexing.pyx
```diff
@@ -62,18 +72,18 @@ cdef tuple _ndarray_nonzero(ndarray self):
         count_nonzero = int(scan_index[-1])  # synchronize!
     ndim = max(<int>self._shape.size(), 1)
     if count_nonzero == 0:
-        return (ndarray((0,), dtype=dtype),) * ndim
+        return ndarray((0, ndim), dtype=dtype)
 
     if ndim <= 1:
-        dst = ndarray((count_nonzero,), dtype=dtype)
+        dst = ndarray((count_nonzero, 1), dtype=dtype)
         _nonzero_kernel_1d(nonzero, scan_index, dst)
```
jakirkham: There seems to be a special case kernel for the 1-D case. Maybe we don't need this any more and can drop it? Possibly related to the dimensionality differences between CuPy and NumPy identified.

cupy/core/_routines_indexing.pyx
jakirkham: Thanks for the clarification. Maybe we are not seeing that issue given the way the CuPy kernel is constructed.

As to `nonzero`, maybe this is related to 1-D special casing as outlined in [this comment]( https://github.com/cupy/cupy/pull/3135#discussion_r388194084 )?

cupy/core/_routines_indexing.pyx
```diff
@@ -362,7 +367,7 @@ _nonzero_kernel = ElementwiseKernel(
     if (src != 0){
         for(int j = 0; j < _ind.ndim; j++){
             ptrdiff_t ind[] = {j, index - 1};
-            dst[ind] = _ind.get()[j];
+            dst[:, ind] = _ind.get()[:, j];
```
jakirkham: Maybe we need to flip these too?

```suggestion
            ptrdiff_t ind[] = {index - 1, j};
            dst[ind] = _ind.get()[j];
```

cupy/core/_routines_indexing.pyx
```diff
@@ -352,7 +357,7 @@ cdef ndarray _simple_getitem(ndarray a, list slice_list):
 
 _nonzero_kernel_1d = ElementwiseKernel(
     'T src, S index', 'raw S dst',
-    'if (src != 0) dst[index - 1] = i',
+    'if (src != 0) dst[index - 1][0] = i',
     'nonzero_kernel_1d')
```
jakirkham: Think we can drop this as well since it is no longer used

cupy/core/_routines_indexing.pyx
jakirkham: This may now need to be updated since the kernel has changed. Though would be curious to hear what you find from your testing 🙂

cupy/_sorting/search.py
```diff
@@ -3,6 +3,7 @@
 from cupy.core import fusion
 from cupy import util
 
+from  cupy.core._routines_indexing import _ndarray_argwhere
```
jakirkham: Small style fix from flake8 on Travis CI

```suggestion
from cupy.core._routines_indexing import _ndarray_argwhere
```

jakirkham:
Thanks again for all of the work here, Rushabh! 😄 This is coming along nicely. I think we are getting close to completing this. Probably a few small fixes depending on how testing goes.

cupy/_sorting/search.py
rushabh-v: Oh yes. But on testing and in `continuous-integration/travis-ci/pr` also there is an `ImportError` for this? I am trying to find out the mistake.

jakirkham:
@leofang, would you have a chance to take a look at this (no worries if not)? 🙂

cupy/_sorting/search.py
asi1024: Please add a test for 0-dim input.

cupy/core/_routines_indexing.pyx
```diff
@@ -350,18 +355,12 @@ cdef ndarray _simple_getitem(ndarray a, list slice_list):
     return v
 
 
-_nonzero_kernel_1d = ElementwiseKernel(
```
asi1024: Why did you remove this kernel? Please compare the performance for 1-dim array inputs.

rushabh-v: Okay, I did a quick benchmark for 1d array and the output is as below,
```python
>>>a = cupy.random.rand(1000000)
>>>print(repeat(cupy.argwhere, (a,)))
argwhere            :   324.478 us   +/-46.795 (min:  291.071 / max: 2166.469) us    414.731 us   +/-46.216 (min:  382.560 / max: 2251.872) us
```

leofang: @rushabh-v 
1. You should test `nonzero`, not `argwhere`, because you deleted the kernel and replaced with a new implementation.
2. You should also compare with the performance before this PR, and see if your changes improve it or not.

cupy/core/_routines_indexing.pyx
```diff
@@ -350,18 +355,12 @@ cdef ndarray _simple_getitem(ndarray a, list slice_list):
...
-    'nonzero_kernel_1d')
-
-
 _nonzero_kernel = ElementwiseKernel(
     'T src, S index', 'raw S dst',
     '''
     if (src != 0){
         for(int j = 0; j < _ind.ndim; j++){
-            ptrdiff_t ind[] = {j, index - 1};
+            ptrdiff_t ind[] = {index - 1, j};
```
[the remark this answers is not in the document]
asi1024: Please check the strides of the return values.
```py
>>> import numpy, cupy
>>>
>>> for xp in (numpy, cupy):
...     x = xp.arange(12).reshape(3, 4)
...     print(xp.argwhere(x).strides)
...
(8, 88)
(16, 8)
```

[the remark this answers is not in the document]
asi1024: Hmm, NumPy seems to use different implementation in `numpy.argwhere` and `numpy.nonzero`.
```py
>>> for xp in (numpy, cupy):
...     x = xp.arange(12).reshape(3, 4)
...     print(xp.nonzero(x)[0].strides)
...
(16,)
(16,)
```
This PR fixes the strides of return value of `numpy.nonzero`, so it is OK to keep the current implementaion. However, we may fix later for strides of `numpy.argwhere`

cupy/core/_routines_indexing.pyx
jakirkham: We were thinking this was redundant in the `argwhere` case as the result is always a 2-D array and it seemed like we were running into some test failures with it. Did not realize there was a notable performance increase in performance for that kernel. If that's true, agree we should add it back

To Leo's points above, would add one more:

3. Include more `0`s in the data. Maybe `a = (cupy.random.rand(1000000) < 0.5)`?

leofang: Oh I see, sorry about the noise. Didn't follow the discussion from start. But it'd still be good to have benchmarks done.

rushabh-v: Okay, so the performance on this patch is:
```python
>>>a = cupy.random.rand(1000000)
>>>print(repeat(cupy.nonzero, (a < 0.5,)))
nonzero             :   419.186 us   +/-23.982 (min:  401.881 / max:  960.754) us    549.686 us   +/-22.333 (min:  530.432 / max: 1090.560) us
```

And before this PR it was,
```python
>>>a = cupy.random.rand(1000000)
>>>print(repeat(cupy.nonzero, (a < 0.5,)))
nonzero             :   282.070 us   +/-46.222 (min:  251.944 / max: 1066.658) us    362.395 us   +/-45.803 (min:  333.696 / max: 1143.040) us
```

jakirkham: Ah ok! So it sounds like we do need the 1-D case for performance reasons.

rushabh-v: Oh yes, I'll be adding that back soon.

jakirkham: Sorry for the back-and-forth on this 😞

rushabh-v: No worries : )
Performance is the most important thing.

rushabh-v: I have added `_nonzero_kernel_1d` back. And it is passing all the tests.
But I don't know why while running `cupyx.time.repeat` on it for 1d array, it raises,
```python
CompileException: /content/cupy/cupy/core/include/cupy/carray.cuh(285): warning: statement is unreachable
          detected during instantiation of "void CIndexer<_ndim>::set(ptrdiff_t) [with _ndim=1]" 
/tmp/tmpt6doxcch/ecc71d65296e1d43cf7830da23d75472_2.cubin.cu(12): here

/tmp/tmpt6doxcch/ecc71d65296e1d43cf7830da23d75472_2.cubin.cu(15): error: expression must have pointer-to-object type

1 error detected in the compilation of "/tmp/tmpt6doxcch/ecc71d65296e1d43cf7830da23d75472_2.cubin.cu".
```

rushabh-v: (Note: I was running it on colab.)

rushabh-v: Oh figured it out,
It's kind of an `IndexError`.

```python
a = cupy.random.rand(1000000)
print(repeat(cupy.nonzero, (a<0.5,))) # raises error
print(repeat(cupy.nonzero, ((a<0.5).reshape(1, 1000000),))) # does not raise error
```

rushabh-v: Okay, So I have added the `_nonzero_kernel_1d` back. And it's passing all the tests.

And I also did the benchmark on it, and it's output is as follows,
```python
>>>a = cupy.random.rand(1000000)
>>>print(repeat(cupy.nonzero, (a<0.5,)))
nonzero             :   237.272 us   +/-52.330 (min:  207.846 / max: 2926.066) us    269.696 us   +/-51.127 (min:  237.248 / max: 2938.528) us
```

jakirkham: Thanks for the update Rushabh! 😄

So it looks like with these changes we *may* even be seeing a speedup, but it's difficult to say with the uncertainty in the measurement. Does that sound right?

rushabh-v: Actually, I am testing it on Google Colab(by building a development-env there as we do locally). So, if I am not wrong, it might be an uncertainty from the cloud side. Does that make sense?

rushabh-v: should do the benchmark once again?

jakirkham: If you have IPython or Jupyter, you could use `%timeit`, which would be good for measuring this sort of thing.

rushabh-v: No actually I don't have Nvidia GPU. That's why I am using colab for testing and for doing the benchmarks.

jakirkham: Right I think Colab has those things though. Or at least their implementation of a Jupyter Notebook. Would give it a try and see if it works 😉

grlee77: > If you have IPython or Jupyter, you could use %timeit, which would be good for measuring this sort of thing.

It looks like he has been using `cupyx.time.repeat` above? That function forces device synchronization to avoid artificially low timings that can occur without it. So, I wouldn't recommend switching from `repeat` to `%timeit`.

I think @rushabh-v was indicating that there may be uncertainty on the hardware side. I don't think Colab will necessarily have assigned identical hardware if those timings were run across different sessions.

leofang: Please don't do `%timeit` for any GPU code unless you're certain a synchronization will be done at the end of your code snippet. This is one of the most common pitfalls for pure Python/IPython/Jupyter users.

leofang: Let me benchmark locally and see if I can get consistent results.

jakirkham: Thanks for the feedback Gregory! 😄

Good to know. Wouldn't adding `cupy.cuda.Stream().synchronize()` be sufficient? Or is there more that `repeat` is doing.

The main thing `%timeit` would help with is providing options to gather a sufficient number of runs to really cutdown on standard deviation. Does `repeat` have an option for this?

Should add I don't know that much about the setup on Colab. So may be missing other concerns with benchmarking there.

grlee77: > The main thing %timeit would help with is providing options to gather a sufficient number of runs to really cutdown on standard deviation. Does repeat have an option for this?

Here is the function signature: https://github.com/cupy/cupy/blob/d58bbb0db60dfed1614e6e173927f0fd05f292ee/cupyx/time.py#L43

The first `n_warmup` runs are not included in the reported time and then `n` repetitions are measured. GPU times are recorded using CUDA stream events, but I think that is less important than making sure synchronization is performed.

The other advantage over `%timeit` is that the first set of numbers in the printouts gives indicates the time spent on the CPU side while the second, longer set of numbers includes the GPU kernel time as well. I think the output could be better labeled to indicate that, though. As is you kind of have to read through the code to know what is what in that display.

leofang:
Got a slight performance degradation:
```python
>>> import cupy as cp
>>> a = cp.random.rand(1000000)
>>> print(repeat(cp.nonzero, (a < 0.5,)))
```
Before this PR:
```
nonzero             :   231.787 us   +/-36.985 (min:  205.528 / max:  513.012) us    263.744 us   +/-35.361 (min:  238.016 / max:  534.624) us
```
After:
```
nonzero             :   244.592 us   +/-36.502 (min:  215.925 / max:  536.494) us    273.460 us   +/-34.160 (min:  246.336 / max:  575.584) us
```
For the moment I am not that concerned, but I do want to take another look at the changes later today. I am puzzled by the changes in the output data layout...

cupy/core/_routines_indexing.pyx
leofang: I wonder how to do a benchmark on Colab that involves rebuilding the library? Ideally we should stay in the same Colab session for a fair comparison, but I know very little about Colab 😛

rushabh-v: > I think the output could be better labeled to indicate that, though. As is you kind of have to read through the code to know what is what in that display.

I would like to open a PR for that.

grlee77: > I would like to open a PR for that.

Sounds good to me. I think allowing `kwargs` in addition to `args` would also be useful.

cupy/core/_routines_indexing.pyx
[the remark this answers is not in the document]
leofang: @asi1024 Could you take another look at this contiguity issue since @rushabh-v reverted his change on the `nonzero` kernel? I think this is the reason I saw the performance degradation (https://github.com/cupy/cupy/pull/3135#issuecomment-595891437).

tests/cupy_tests/sorting_tests/test_search.py
```diff
@@ -278,6 +278,38 @@ def test_flatnonzero(self, xp, dtype):
...
+@testing.parameterize(
+    {'array': numpy.array(0)},
+    {'array': numpy.array(1)},
+)
+@testing.gpu
+@testing.with_requires('numpy>=1.17.0')
+class TestArgwhereZeroDimension(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_raises(accept_error=DeprecationWarning)
```
leofang: I don't think this is OK. Warnings are not exceptions. Please change this line and line 256 (yeah it was not correctly done in the codebase) and use the context manager `with testing.assert_warns(DeprecationWarning)` instead.

tests/cupy_tests/sorting_tests/test_search.py
```diff
@@ -278,6 +278,38 @@ def test_flatnonzero(self, xp, dtype):
...
+        array = xp.array(self.array, dtype=dtype)
+        return xp.argwhere(array)
+
+
+@testing.parameterize(
+    {'array': numpy.array(0)},
+    {'array': numpy.array(1)},
+)
+@testing.gpu
+@testing.with_requires('numpy>=1.17.0')
```
leofang: Is this because the `DeprecationWarning` is issued since NumPy >= 1.17.0?

cupy/core/_routines_indexing.pyx
jakirkham: Thanks Gregory! That makes sense. So maybe we should be bumping `n` to a larger value?

tests/cupy_tests/sorting_tests/test_search.py
```diff
@@ -278,6 +278,38 @@ def test_flatnonzero(self, xp, dtype):
...
+    {'array': numpy.empty((0, 2, 0))},
+)
+@testing.gpu
+class TestArgwhere(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_argwhere(self, xp, dtype):
+        array = xp.array(self.array, dtype=dtype)
+        return xp.argwhere(array)
```
leofang: Not sure what happened with this test script, but I don't think the test style here is compliant with other tests. Usually we parametrize over shapes, and then initialize the input on CPU (numpy) or GPU (cupy) with the parameters to avoid explicit data transfer. For random input and other kinds of inputs, we can have different test methods (say `test_argwhere_random`, `test_argwhere_empty`, or just `test_argwhere_1`, `..._2`, etc).

For this case, though, it is @asi1024's call. We might need a separate PR for improving the style?

cupy/core/_routines_indexing.pyx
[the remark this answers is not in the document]
leofang: btw, for `argwhere`, note that the NumPy output is F-contiguous. I guess this is reason for the stride discrepancy.

leofang:
I left https://github.com/cupy/cupy/pull/3135#discussion_r389067918 and https://github.com/cupy/cupy/pull/3135#discussion_r389077690 on the performance and stride issues.

tests/cupy_tests/sorting_tests/test_search.py
rushabh-v: Oh, that's my mistake. I just confirmed that NumPy does not raise any warning for `0d` on `argwhere`. it just raises that warning in the case of `nonzero`. 
But @beckernick suggested(https://github.com/cupy/cupy/pull/3135#discussion_r386174537) adding this warning to `argwhere` as well. Should we remove the warning as NumPy does not warn in that case? If not so, then I can edit the above test to check if only CuPy raises this warning or not.

tests/cupy_tests/sorting_tests/test_search.py
rushabh-v: Done

tests/cupy_tests/sorting_tests/test_search.py
asi1024: Yes, we prefer to parametrize over shapes, but I think we don't need to resolve it in this PR.

cupy/_sorting/search.py
```diff
@@ -212,6 +215,30 @@ def where(condition, x=None, y=None):
...
+
+    Returns a (N, ndim) dimantional array containing the
+    indices of the non-zero elements. Where `N` is number of
+    non-zero elements and `ndim` is dimention of the given array.
+
+    Args:
+        a (cupy.ndarray): array
+
+    Returns:
+        cupy.array: Indices of elements that are non-zero.
```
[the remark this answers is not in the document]
rushabh-v: Done

asi1024:
Jenkins, test this please.

jakirkham:
> I am puzzled by the changes in the output data layout...

@leofang `nonzero` is effectively transposed relative to `argwhere`. Here's an example to provide more context:

```python
In [1]: import numpy as np                                                      

In [2]: a = np.array([[0, 1, 0], 
   ...:               [2, 0, 3], 
   ...:               [0, 4, 0]], dtype=float)                                  

In [3]: a.nonzero()                                                             
Out[3]: (array([0, 1, 1, 2]), array([1, 0, 2, 1]))

In [4]: np.array(a.nonzero())                                                   
Out[4]: 
array([[0, 1, 1, 2],
       [1, 0, 2, 1]])

In [5]: np.argwhere(a)                                                          
Out[5]: 
array([[0, 1],
       [1, 0],
       [1, 2],
       [2, 1]])
```

asi1024:
LGTM. Thank you for the PR!

rushabh-v:
Thanks everybody!

I've learnt many new things about performance and all with this PR :D

leofang:
Thanks!

cupy/core/_routines_indexing.pyx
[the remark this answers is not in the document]
leofang: I just recalled that CuPy does not guarantee returning identical strides as in NumPy (#2717), so there's nothing to be fixed!

PR #3139, opened 2020-03-01
Use elementwise kernel in `cupy.fill_diagonal`

takagi:
Follows #3011 .

This PR reimplement `fill_diagonal` using `ElementwiseKernel` with manual indexing its `raw` argument specifier brings.

At first I thought that I had to make some modification on `ElementwiseKernel` to provide an indexer for each raw argument, but I've found that just `CArray::operator[](ptrdiff_t i)` performs the necessary index computation here so I don't need to touch `ElementwiseKernel`.


emcastillo:
Can't beliebe it was this easy 😂

emcastillo:
Jenkins, test this please

takagi:
Yeah, it's quite awesome.

PR #3140, opened 2020-03-01
add support for 0 sized matrices in svd

### LINKED ISSUE #2371: Incorrect behavior with empty arrays for `linalg.qr` and `linalg.svd`
`qr` and `svd` routines raise `CUSOLVERError: CUSOLVER_STATUS_INVALID_VALUE` for zero-sized input arrays.
NumPy supports this kind of input for `qr`, `svd` since 1.16
https://github.com/numpy/numpy/blob/master/doc/release/1.16.0-notes.rst#linalglstsq-linalgqr-and-linalgsvd-now-work-with-empty-arrays
The problem is that when calling cuSOLVER routines leading dimension is passed as `input.shape[0]` but it should be `max(1, input.shape[0])`.

rushabh-v:
closes #2371

### DIFF
--- a/cupy/linalg/decomposition.py
+++ b/cupy/linalg/decomposition.py
@@ -364,7 +364,7 @@ def svd(a, full_matrices=True, compute_uv=True):
             if full_matrices:
                 return cupy.eye(m, m), cupy.array([]), cupy.empty((0, 0))
             else:
-                return cupy.empty(m, 0), cupy.array([]), cupy.empty((0, 0))
+                return cupy.empty((m, 0)), cupy.array([]), cupy.empty((0, 0))
         else:
             return cupy.array([])
 
--- a/cupy/linalg/decomposition.py
+++ b/cupy/linalg/decomposition.py
@@ -362,11 +362,11 @@ def svd(a, full_matrices=True, compute_uv=True):
     if mn == 0:
         if compute_uv:
             if full_matrices:
-                return cupy.eye(m, m), cupy.array([]), cupy.empty((0, 0))
+                return cupy.eye(m, m), cupy.empty((0,)), cupy.empty((0, 0))
             else:
-                return cupy.empty((m, 0)), cupy.array([]), cupy.empty((0, 0))
+                return cupy.empty((m, 0)), cupy.empty((0,)), cupy.empty((0, 0))
         else:
-            return cupy.array([])
+            return cupy.empty((0,))
 
     if compute_uv:
         if full_matrices:
--- a/cupy/linalg/decomposition.py
+++ b/cupy/linalg/decomposition.py
@@ -357,7 +357,16 @@ def svd(a, full_matrices=True, compute_uv=True):
         m, n = a.shape
         x = a.transpose().astype(a_dtype, order='C', copy=True)
         trans_flag = True
+
     mn = min(m, n)
+    if mn == 0:
+        if compute_uv:
+            if full_matrices:
+                return cupy.eye(m, m), cupy.array([]), cupy.empty((0, 0))
+            else:
+                return cupy.empty(m, 0), cupy.array([]), cupy.empty((0, 0))
+        else:
+            return cupy.array([])
 
     if compute_uv:
         if full_matrices:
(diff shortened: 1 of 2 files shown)


rushabh-v:
I have tested it. The tests are passing.

cupy/linalg/decomposition.py
```diff
@@ -357,7 +357,16 @@ def svd(a, full_matrices=True, compute_uv=True):
...
         x = a.transpose().astype(a_dtype, order='C', copy=True)
         trans_flag = True
+
     mn = min(m, n)
+    if mn == 0:
+        if compute_uv:
+            if full_matrices:
+                return cupy.eye(m, m), cupy.array([]), cupy.empty((0, 0))
+            else:
+                return cupy.empty((m, 0)), cupy.array([]), cupy.empty((0, 0))
```
asi1024: This code path looks not tested.

cupy/linalg/decomposition.py
```diff
@@ -357,7 +357,16 @@ def svd(a, full_matrices=True, compute_uv=True):
...
+        else:
+            return cupy.array([])
```
asi1024: Ditto.

cupy/linalg/decomposition.py
```diff
@@ -357,7 +357,16 @@ def svd(a, full_matrices=True, compute_uv=True):
         m, n = a.shape
         x = a.transpose().astype(a_dtype, order='C', copy=True)
         trans_flag = True
+
     mn = min(m, n)
+    if mn == 0:
+        if compute_uv:
+            if full_matrices:
+                return cupy.eye(m, m), cupy.array([]), cupy.empty((0, 0))
```
asi1024: How about writing as follows for the consistency?
```suggestion
                return cupy.eye(m, m), cupy.empty((0,)), cupy.empty((0, 0))
```

cupy/linalg/decomposition.py
rushabh-v: wouldn't it be directly tested by the testing.product as it's `'full_matrices': [True, False]` [here](https://github.com/cupy/cupy/blob/86eab8f0ebc124d9e6b746efa0d635d6f37062ae/tests/cupy_tests/linalg_tests/test_decomposition.py#L116) and `self.full_metrices` has been passed [here](https://github.com/cupy/cupy/blob/86eab8f0ebc124d9e6b746efa0d635d6f37062ae/tests/cupy_tests/linalg_tests/test_decomposition.py#L135)?

cupy/linalg/decomposition.py
rushabh-v: I added that case now.

cupy/linalg/decomposition.py
rushabh-v: Done

tests/cupy_tests/linalg_tests/test_decomposition.py
```diff
@@ -203,3 +203,16 @@ def test_svd_no_uv(self):
...
         self.check_rank2(cupy.random.randn(1, 2, 3, 4).astype(numpy.float64))
+
+    @testing.with_requires('numpy>=1.16')
+    def test_empty_array(self):
+        self.check_usv((0, 3))
+        self.check_usv((3, 0))
+        self.check_usv((1, 0))
+
+    @testing.with_requires('numpy>=1.16')
+    @testing.numpy_cupy_allclose(atol=1e-4)
```
asi1024: In this test, the result arrays can be compared without tolerances.
```suggestion
    @testing.numpy_cupy_array_equal()
```

cupy/linalg/decomposition.py
asi1024: I confirmed that this line is covered. Thanks!

tests/cupy_tests/linalg_tests/test_decomposition.py
rushabh-v: I have made that change.

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

PR #3141, opened 2020-03-02
Add `cupy.cuda.ExternalStream`

emcastillo:
Related to #3126 and https://github.com/pytorch/pytorch/pull/33860

This allows using external streams in CuPy by creating a `ExternalStream` object from a given `cudaStream_t` pointer.

The main use-case for this is to allow interface cupy with other frameworks such as numba or PyTorch. In PyTorch the streams are managed in a pool so although we can trick and reassign the`.ptr` attribute of a CuPy stream to use a Torch managed stream, the RAII interface can delete this stream thus making this approach not safe.

The approach is similar to `cupy.cuda.BaseMemory` and `cupy.cuda.Memory`

TODO: add more tests

cc @leofang

### DIFF
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -98,6 +98,7 @@ from cupy.cuda.stream import Event  # NOQA
 from cupy.cuda.stream import get_current_stream  # NOQA
 from cupy.cuda.stream import get_elapsed_time  # NOQA
 from cupy.cuda.stream import Stream  # NOQA
+from cupy.cuda.stream import ExternalStream  # NOQA
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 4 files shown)


leofang:
One more stupid question: On the event of deleting an `ExternalStream`, don't we need to check and remove the ptr from the thread-local storage (similar to what's done for `Stream`)?

emcastillo:
I think that it should be the user responsibility to set the correct stream before explicitly deleting the current one.

jakirkham:
cc @kkraus14 @gmarkall @jrhemstad

asi1024:
Jenkins, test this please.

leofang:
This is irrelevant of this PR, just for my curiosity: any reason BaseStream/Stream/ExternalStream can’t be cdef classes?

emcastillo:
I think it was because Cython problems with context managers and destructors.

leofang:
Ah right, thanks for reminder 🤦

asi1024:
LGTM.

PR #2759, opened 2019-12-05
Fix slow import of cupy

cgohlke:
Delay import of pkg_resources

On my Windows workstation, `import cupy` takes 3.5 s, about 3 s of which are spent in `import pkg_resources`.

Importing `pkg_resources` is known to be excruciatingly slow. See https://github.com/pypa/setuptools/issues/510

PR #3145, opened 2020-03-03
Remove `cupy/cuda/_environment.py`

leofang:
There are two `_environment.py` in the codebase that implement `get_cuda_path()`, and this is quite confusing and unnecessary. I propose to merge them.

In addition, note that `get_cuda_path()` in `cupy/cuda/_environment.py` is *never* used in the codebase -- currently `cupy.cuda.get_cuda_path()` is just an alias of `cupy._environment.get_cuda_path()`. So this is another reason to do this.

What I did is to move all functions in `cupy/cuda/_environment.py` to `cupy/_environment.py`, with minor changes that I'll comment in the diff shortly.

Ideally we should do this the other way around (that is, remove `cupy/_environment.py`), but it is not possible due to the intertwined dependency between `cupy.cuda` and `cupy.core`. 

I am not 100% confident this will work given the complex dependency graph; I only tested `tests/cupy_tests/core_tests/test_raw.py`, so it'd be nice to kick off the CI first.

Thanks.


cupy/_environment.py
```diff
@@ -2,35 +2,71 @@
...
 """
 
+import distutils
 import os
 import os.path
+import shutil
 
 
 _cuda_path = None
+_nvcc_path = None
```
leofang: In `cupy/cuda/_environment.py` both were initialized to `()`, which doesn't make much sense.

cupy/_environment.py
```diff
@@ -2,35 +2,71 @@
...
+        return '/usr/local/cuda'
+
+    return None
+
+
+def _get_nvcc_path():
+    # Honor the "NVCC" env var
+    nvcc_path = os.environ.get('NVCC', None)
+    if nvcc_path is not None:
+        return distutils.util.split_quoted(nvcc_path)
```
leofang: This is an extra piece I added. Not sure if this is acceptable, but would be nice for users to be able to overwrite, given that `RawKernel` can use `nvcc` as the alternative backend.

cupy/_environment.py
```diff
@@ -2,35 +2,71 @@
 This file must not depend on any other CuPy modules.
 """
 
+import distutils
```
leofang: I remember there's a PR fixing the import speed by removing `distutils` (or some other libraries), but I can't remember nor find it. Import time is also a concern in my PR.

emcastillo: Maybe this one? https://github.com/cupy/cupy/pull/2759

leofang: That's it! Thanks, Emilio. I wonder if I should just do our own string splitting here and avoid importing distutils?

emcastillo: If it's easier it should be preferable.

leofang: I think `split_quoted()` is an overkill here. It doesn't do input verification for us, and we may as well simply call `split()` to get a list of strings. With a valid input, it'd just do the work; with an invalid input, the command from either `split_quoted()` or `split()` would fail when the subprocess is invoked.

cupy/cuda/_environment.py
```diff
@@ -1,56 +0,0 @@
...
-        return '/usr/local/cuda'
-
-    return None
-
-
-def _get_nvcc_path():
-    # Directly lookup PATH
-    nvcc_path = shutil.which('nvcc')
-    if nvcc_path is not None:
-        return nvcc_path
```
leofang: This part is removed when copying to `cupy/_environment.py` because it's done in the call to `get_cuda_path()` right below.

leofang:
Done. PTAL.

emcastillo:
Jenkins, test this please

leofang:
Ah, let me rebase after breakfast.😅

cupy/cuda/compiler.py
```diff
@@ -151,7 +153,11 @@ def compile_using_nvcc(source, options=(), arch=None,
         assert not separate_compilation
 
     arch_str = '-gencode=arch=compute_{cc},code=sm_{cc}'.format(cc=arch)
-    cmd = [_environment.get_nvcc_path(), arch_str]
+    _nvcc = get_nvcc_path()
+    if isinstance(_nvcc, str):
+        cmd = [get_nvcc_path(), arch_str]
+    else:  # _nvcc is a list (from the env var NVCC)
+        cmd = _nvcc.append(arch_str)
```
leofang: I am a stupid dumbass...🤦🏻‍♂️

leofang:
Had to rebase to fix the import path in the test added in #3201. PTAL.

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

leofang:
c23302e should fix it. The problem is that the tests need to be updated due to the code fix, but not in the fix itself.

emcastillo:
Jenkins, test this please

leofang:
Thanks again!

cupy/_environment.py
leofang: Ouch, that does make sense: it distinguishes between the two states "uninitialized" and "not found". Will send a PR later...😞

leofang: Fixed in #3413.

PR #3148, opened 2020-03-03
Fix CUB-based `cupy.prod` for half precision

leofang:
Follow-up of #3067, in which I made yet another CUB bug 😞 Fortunately the tests in #2598 were able to catch it...

For reduction operations we need a start point, and for `prod` it's 1. On the device, `static_cast<__half>(1)` gives 1 in half precision as expected, but on the host it gives 0 (mdfk!) A correct behavior on host is important for CUB, because it doesn't directly launch a device kernel.

The reason is, quite annoyingly, that a conversion constructor is only available for device functions:
https://github.com/cupy/cupy/blob/106943a0006d27a2c4c164c4799bafcba1428160/cupy/core/include/cupy/_cuda/cuda-10.0/cuda_fp16.hpp#L184
and I have no idea where the zero comes from for host functions...Likely it just returns an uninitialized value?

The current workaround here is quite brute-force: we rely on the fact that a floating-point literal can be converted to other types and precisions correctly. (We might want a better solution here, preferably making an exception only for `__half`. But it's midnight here and I'm basically shutdown...😴 @emcastillo Any suggestion?)

### DIFF
--- a/cupy/cuda/cupy_cub.cu
+++ b/cupy/cuda/cupy_cub.cu
@@ -469,8 +469,10 @@ struct _cub_reduce_prod {
         int num_items, cudaStream_t s)
     {
         _multiply product_op;
+        // the init value is cast from 1.0f because on host __half can only be
+        // initialized by float or double; static_cast<__half>(1) = 0 on host.
         DeviceReduce::Reduce(workspace, workspace_size, static_cast<T*>(x),
-            static_cast<T*>(y), num_items, product_op, 1.0, s);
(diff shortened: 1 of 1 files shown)


leofang:
2955136 seems to be overkill...PTAL

emcastillo:
I think the previous approach with a comment describing it was better.

leofang:
Thanks, Emilio! Done.

emcastillo:
Jenkins, test this please

PR #3149, opened 2020-03-03
`__bytes__` for `cupy.ndarray`

### LINKED ISSUE #3007: Implementing `__bytes__` on `ndarray`s
It would be useful to have an implementation of `__bytes__` for `ndarray` analogous to NumPy's. Ideally this would avoid an intermediate copy to a NumPy array (by using things like `cuMemcpyDtoH`).

Harshan01:
Initial PR for `__bytes__` for `cupy.ndarray`

Resolves #3007

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -144,6 +144,9 @@ cdef class ndarray:
         self._set_contiguous_strides(itemsize, c_order)
         self.data = memory.alloc(self.size * itemsize)
 
+    def __bytes__(self):
+        return self.tobytes()
+
     @property
     def __cuda_array_interface__(self):
         desc = {
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -144,9 +144,6 @@ cdef class ndarray:
         self._set_contiguous_strides(itemsize, c_order)
         self.data = memory.alloc(self.size * itemsize)
 
-    def __bytes__(self):
-        return self.tobytes()
-
     @property
     def __cuda_array_interface__(self):
         desc = {
@@ -1385,6 +1382,9 @@ cdef class ndarray:
 
     def __hex__(self):
         return hex(self.get())
+
+    def __bytes__(self):
+        return self.tobytes()
 
     # String representations:
 
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1384,7 +1384,7 @@ cdef class ndarray:
         return hex(self.get())
 
     def __bytes__(self):
-        return self.tobytes()
+        return bytes(self.get())
 
     # String representations:
 
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -466,3 +466,11 @@ class TestSize(unittest.TestCase):
     def test_size_zero_dim_array_with_axis(self, xp):
         x = testing.shaped_arange((), xp, numpy.int32)
         return xp.size(x, 0)
+
+
+@testing.gpu
+class TestPythonInterface(unittest.TestCase):
+
+    def test_bytes_tobytes(self):
+        x = testing.shaped_arange((3, 4, 5))
+        assert x.tobytes() == bytes(x)
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -492,5 +492,5 @@ class TestPythonInterface(unittest.TestCase):
     @testing.for_all_dtypes()
     @testing.numpy_cupy_equal()
     def test_bytes_tobytes_scalar(self, xp, dtype):
-        x = xp.asscalar(xp.array([3], dtype))
+        x = xp.array([3], dtype).item()
         return bytes(x)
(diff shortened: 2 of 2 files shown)


emcastillo:
Hi, thanks for your contribution.
NumPy does not define a `__bytes__` operation.  `tobytes` seems to provide this functionality.
As `tobytes` is already present in CuPy, I don't think this change is necessary.

Harshan01:
I understand `__bytes__` is redundant to `tobytes`. But, `tobytes` cannot resolve the problem in the linked issue. Is there any other way to hook `tobytes` to Python so that `bytes(arr)` works?

emcastillo:
Yes, sorry it was my mistake.
Can you add a test?

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -466,3 +466,11 @@ def test_size_axis_too_small(self, xp):
...
         x = testing.shaped_arange((), xp, numpy.int32)
         return xp.size(x, 0)
+
+
+@testing.gpu
+class TestPythonInterface(unittest.TestCase):
+
+    def test_bytes_tobytes(self):
+        x = testing.shaped_arange((3, 4, 5))
+        assert x.tobytes() == bytes(x)
```
emcastillo: We would like to test that both numpy and cupy arrays yield the same values. I.e. compare the output of bytes(numpy_array) to bytes(cupy_array)

This should be done for an empty array, an scalar, and a several elements array.

Harshan01:
Tests for `empty` arrays with non-zero size and scalars are added.
However, `empty` arrays with 0 size is not yet added because of the problem as mentioned in the linked issue.

Harshan01:
As mentioned in the issue (numpy/numpy#15698), the `bytes` and `tobytes` discrepancy arises in numpy only for a specific case of empty dtype tuple and `int` dtype. I would like to receive your suggestion on how to carry forward with the PR.

cupy/core/core.pyx
```diff
@@ -1383,6 +1383,9 @@ cdef class ndarray:
     def __hex__(self):
         return hex(self.get())
 
+    def __bytes__(self):
+        return self.tobytes()
```
jakirkham: Does this work?

```suggestion
        return bytes(self.get())
```

Harshan01: I haven't tried it yet. I will make the changes and test it by tomorrow.

Harshan01: `bytes(self.get())` would probably work as per numpy as `self.get()` returns a numpy array. This would mean that corresponding tests for `numpy` and `cupy` will match, but not for `tobytes` and `bytes`

Harshan01:
Changes are made to match `cupy` output with `numpy`'s.

cupy/core/core.pyx
jakirkham: Yeah that's ok I think. Getting `bytes` and `tobytes` to align sounds like a NumPy issue. Once they fix that, it will automatically be fixed in CuPy 🙂

jakirkham:
Thanks Harshan! 😄

emcastillo:
Awesome work here, thanks!

jakirkham:
Could we please run this on CI? I think we are ready 🙂

emcastillo:
Jenkins, test this please

emcastillo:
Fails with python 3.5 
```
message = "Results are not equal:\ncupy: None\nnumpy: b'\\x00\\x00\\x00'"

            @functools.wraps(impl)
            def test_func(self, *args, **kw):
                # Run cupy and numpy
                (
                    cupy_result, cupy_error, cupy_tb,
                    numpy_result, numpy_error, numpy_tb) = (
                        _call_func_numpy_cupy(
                            self, impl, args, kw, name, sp_name, scipy_name))
    
                if cupy_result != numpy_result:
                    message = '''Results are not equal:
    cupy: %s
    numpy: %s''' % (str(cupy_result), str(numpy_result))
>                   raise AssertionError(message)
E                   AssertionError: Results are not equal:
E                   cupy: None
E                   numpy: b'\x00\x00\x00'
```

Harshan01:
@emcastillo I learnt that `xp.asscalar` shouldn't have worked in all cases. Can you share your thoughts on why it worked for some cases. BTW, I fixed the test to use `.item()` instead of `xp.asscalar()`. Can you please review it once.

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -492,5 +492,5 @@ def test_bytes_tobytes_empty2(self, xp, dtype):
     @testing.for_all_dtypes()
     @testing.numpy_cupy_equal()
     def test_bytes_tobytes_scalar(self, xp, dtype):
-        x = xp.asscalar(xp.array([3], dtype))
+        x = xp.array([3], dtype).item()
```
emcastillo:
```DeprecationWarning: np.asscalar(a) is deprecated since NumPy v1.16, use a.item() instead
2```

So this is right

emcastillo:
Jenkins, test this please

jakirkham:
Anything else needed here? 😉

emcastillo:
Sorry, felt out of my radar!
Thanks for the remainder

jakirkham:
No worries. Thanks for the review 😄

Issue #3110, opened 2020-02-22
Converting sparse matrix errors


jakirkham:
Am seeing the following error when trying to convert a `csc_matrix` to a `coo_matrix`. Though also see some variation of this when trying to go to a `csr_matrix` from a `csc_matrix` and other similar conversions.

* Conditions (you can just paste the output of `python -c 'import cupy; cupy.show_config()'`)

```
CuPy Version          : 7.2.0
CUDA Root             : /datasets/jkirkham/miniconda/envs/rapids13dev
CUDA Build Version    : 10010
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10010
cuBLAS Version        : 10201
cuFFT Version         : 10101
cuRAND Version        : 10101
cuSOLVER Version      : (10, 2, 0)
cuSPARSE Version      : 10300
NVRTC Version         : (10, 1)
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2406
NCCL Runtime Version  : 2507
```

* Code to reproduce

```python
In [1]: import cupy                                                             

In [2]: a = cupy.array([[0, 1, 0], 
   ...:                 [2, 0, 3], 
   ...:                 [0, 4, 0]], dtype=float)                                

In [3]: a2 = cupy.sparse.csc_matrix(a)                                          

In [4]: a3 = a2.tocoo()                                                         
```

* Error messages, stack traces, or logs

```python
---------------------------------------------------------------------------
CUDADriverError                           Traceback (most recent call last)
<ipython-input-4-14bf9613c858> in <module>
----> 1 a3 = a2.tocoo()

/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/cupyx/scipy/sparse/csc.py in tocoo(self, copy)
    179 
    180         """
--> 181         return self.T.tocoo(copy).T
    182 
    183     def tocsc(self, copy=None):

/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/cupyx/scipy/sparse/csr.py in tocoo(self, copy)
    239             indices = self.indices
    240 
--> 241         return cusparse.csr2coo(self, data, indices)
    242 
    243     def tocsc(self, copy=False):

/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/cupy/cusparse.py in csr2coo(x, data, indices)
    631     # data and indices did not need to be copied already
[...]

Issue #3092, opened 2020-02-18
Printing result from `.max()` errors


jakirkham:
Ran the following Python script using CuPy 7.2.0 and got an error. Note the `print` is what causes the error.

```python
import cupy

a = cupy.arange(5)
r = a.max()
print(r)
```

Exception:

```python
Traceback (most recent call last):
  File "test_cupy_repr.py", line 5, in <module>
    print(r)
  File "cupy/core/core.pyx", line 1366, in cupy.core.core.ndarray.__str__
  File "cupy/core/core.pyx", line 1456, in cupy.core.core.ndarray.get
  File "cupy/core/core.pyx", line 1462, in cupy.core.core.ndarray.get
  File "cupy/cuda/memory.pyx", line 449, in cupy.cuda.memory.MemoryPointer.copy_to_host
  File "cupy/cuda/runtime.pyx", line 378, in cupy.cuda.runtime.memcpy
  File "cupy/cuda/runtime.pyx", line 201, in cupy.cuda.runtime.check_status
cupy.cuda.runtime.CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered
```

Additional exception (during cleanup):

```python
Traceback (most recent call last):
  File "cupy/cuda/driver.pyx", line 247, in cupy.cuda.driver.moduleUnload
  File "cupy/cuda/driver.pyx", line 118, in cupy.cuda.driver.check_status
TypeError: 'NoneType' object is not callable
Exception ignored in: 'cupy.cuda.function.Module.__dealloc__'
Traceback (most recent call last):
  File "cupy/cuda/driver.pyx", line 247, in cupy.cuda.driver.moduleUnload
  File "cupy/cuda/driver.pyx", line 118, in cupy.cuda.driver.check_status
TypeError: 'NoneType' object is not callable
```

If it is a bug report, please include the following information:

* Conditions (you can just paste the output of `python -c 'import cupy; cupy.show_config()'`)
  - CuPy version: 7.2.0
  - OS/Platform: Ubuntu 16.04.5 LTS (Xenial)
  - CUDA version: 10.1.243
  - cuDNN/NCCL version (if applicable): 7.6.5/2.5.7.1
* Code to reproduce: (see above)
* Error messages, stack traces, or logs: (see above)

jakirkham:
cc @JohnZed

JohnZed:
@jakirkham note that if I build from source, I cannot repro this. It happens when I use the conda-forge conda package though.

[...]

PR #2455, opened 2019-09-10
Add appropriate pointer types in Cython to the contribute guide

leofang:
Followup of #1952. Fixes #1913 further. 

This PR is encouraged in https://github.com/cupy/cupy/issues/1913#issuecomment-529298038. The main change is that most of the types in `cupy.cuda.driver` (such as `Function` and `Stream`) are pointers, and thus should be represented by `intptr_t`, see https://github.com/cupy/cupy/issues/1913#issuecomment-529262513. Once it's fixed, the changes should propagate outward to other modules.

***Challenges***: 
1. ~~In `cupy.cuda.memory` it seems streams are used in the memory mappings. I'm trying to figure out which `size_t` is meant to present a stream pointer there...~~
2. It's likely this PR would still not be enough to clear up everything. I expect an incremental approach is more appropriate.

If Jenkins can be kicked off to test the current PR, it'd be great!

Thanks.

PR #3152, opened 2020-03-04
Readd `ndarray` type to `y`

jakirkham:
Closes https://github.com/cupy/cupy/issues/3092
Closes https://github.com/cupy/cupy/issues/3110

cc @leofang

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -98,6 +98,7 @@ cpdef Py_ssize_t _preprocess_array(ndarray arr, tuple reduce_axis,
 
 def device_reduce(ndarray x, op, tuple out_axis, out=None,
                   bint keepdims=False):
+    cdef ndarray y
     cdef memory.MemoryPointer ws
     cdef int dtype_id, ndim_out, kv_bytes, x_size, _op
     cdef size_t ws_size


leofang:
I think you can add that this PR closes #3092 and #3110.

leofang:
This PR  is tested locally btw.

jakirkham:
Ah ok. So this worked for you? 🙂

leofang:
Yes!

jakirkham:
Thanks Leo!

Yeah I'm also a bit fuzzy on why this helps, but that doesn't mean much (given I'm not as familiar with the codebase as all of you 🙂).

emcastillo:
Jenkins, test this please

leofang:
@jakirkham This is the very same error that I made multiple times in the past! The offending line is 
```python
y_ptr = <void *>y.data.ptr
``` 
If `y` is a Python object (since `cdef` was removed; can be confirmed by inspecting the generated cpp file; Cython is to be blamed -- it should deduce `y`'s type!), then this cast is just giving us a pointer to the Python object `y.data.ptr`, not the actual pointer to the data.

Refs: 
1. https://github.com/cupy/cupy/pull/2455#discussion_r325234488
2. https://github.com/cupy/cupy/blob/43c3c21e1fea5f0fc8a449aa97f10227c8ae6f91/cupy/cuda/texture.pyx#L328-L329

jakirkham:
Thanks @emcastillo! 😀 Sorry for the silly typo (autocomplete 😞)

That makes sense! Thanks for the clarification Leo! 😀

Issue #3061, opened 2020-02-10
cupy.sparse.MatDescriptor is not pickleable


cjnolet:
I am encountering a segfault when freeing a CuPy sparse `csr_matrix` which was computed from a Dask array. I have an intuition that Dask might not be the cause of this problem. From a simple inspection of properties like `__cuda_array_interface__` and the allocator on the underlying index/data arrays, I don't see any obvious indications of a double free.

I am also able to successfully pickle and unpickle a CuPy `csr_matrix` and was able to free the resulting unpickled matrix without a segfault. 

Also, the strangest part about this error is that it doesn't happen when the contents of `build_arr` is run sequentially in the calling function. That is, I only get this segfault when returning a Dask.Array backed by sparse CuPy arrays from a function and calling "compute" in the calling function. 

```
from dask.distributed import Client
from dask_cuda import LocalCUDACluster
import cupy as cp
import dask.array

cluster = LocalCUDACluster()
client = Client(cluster)

x = cp.sparse.random(1000, 10).astype(cp.float32).tocsr()
cp.cuda.Stream.null.synchronize()

def build_arr(x):
  f = client.scatter(x)
  ret = dask.array.from_delayed(f, shape=x.shape,
                                                   meta=cp.sparse.csr_matrix(cp.zeros(1), dtype=x.dtype))
  return ret

ret = build_arr(x)
x_ret = ret.compute()
print(str(x_ret))
del x_ret
del ret
```

Here's the output and exception:

```
>>> print(str(x_ret))
  (3, 0)	0.93203676
  (5, 3)	0.816514
  (22, 0)	0.7288292
  (23, 5)	0.1614296
  (56, 2)	0.10485178
  (76, 7)	0.74339426
  (100, 5)	0.58028
  (111, 2)	0.85776955
  (129, 5)	0.6123046
  (133, 4)	0.60502696
  (140, 4)	0.74886125
  (151, 9)	0.315743
  (165, 2)	0.2791235
  (182, 5)	0.9034685
  (183, 4)	0.5153093
  (202, 3)	0.08361749
  (206, 1)	0.8262547
  (221, 0)	0.743914
  (223, 0)	0.6860539
  (252, 4)	0.15654124
  (256, 8)	0.4413018
  (274, 0)	0.72435445
  (274, 2)	0.56952196
  (286, 3)	0.17318754
  (290, 4)	0.1044073
  :	:
  (775, 2)	0.19066067
  (775, 5)	0.2838834
[...]

PR #3157, opened 2020-03-05
Allow `MatDescriptor` to be pickle-able

jakirkham:
Fixes https://github.com/cupy/cupy/issues/3061

### DIFF
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_coo.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_coo.py
@@ -156,8 +156,8 @@ class TestCooMatrix(unittest.TestCase):
         assert n.shape == m.shape
         return n
 
-    def test_pickle_roundtrip(self, xp, sp):
-        s = _make(xp, sp, self.dtype)
+    def test_pickle_roundtrip(self):
+        s = _make(cupy, sparse, self.dtype)
         s2 = pickle.loads(pickle.dumps(s))
         assert s._descr.descriptor != s2._descr.descriptor
         assert s.shape == s2.shape
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_coo.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_coo.py
@@ -162,7 +162,7 @@ class TestCooMatrix(unittest.TestCase):
         assert s.shape == s2.shape
         assert s.dtype == s2.dtype
         if scipy_available:
-            assert (s.get() != s2.get()).count_nonzero()
+            assert (s.get() != s2.get()).count_nonzero() == 0
 
     def test_shape(self):
         self.assertEqual(self.m.shape, (3, 4))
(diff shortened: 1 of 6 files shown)


jakirkham:
(note to self: still needs tests)

jakirkham:
@emcastillo, could you or someone else please help us figure out where to put a test for this? Am not seeing any tests for `MatDescriptor` currently

jakirkham:
FWIW I pushed some tests for pickling sparse matrices, which should exercise this a bit.

jakirkham:
Also may need some pointers on how testing is working here. I think I got the idea, but there were some differences between test modules. So could use some pointers from someone more familiar 🙂

leofang:
John, if your new tests pass they should be fine. (I didn't review very carefully!) They are moving away from using `assertEqual()` to plain `assert` statements, see https://docs-cupy.chainer.org/en/stable/contribution.html#how-to-write-tests, so for now the test suite has a mixed old- and new-style tests.

jakirkham:
Thanks Leo! 😄

That's good to hear. `assert` feels more friendly 😉 

Yeah was noticing that some of these use a `_make` function to get a sparse array. So tried to use that. Not sure if this is preferred or not.

The other thing is some tests seem to have arguments for `xp` and `sp`. Tried to support these, but not sure if it was supported in all modules or if there was anything extra I should be doing to make sure they work.

emcastillo:
I think that current tests are fine

emcastillo:
Jenkins, test this please

tests/cupyx_tests/scipy_tests/sparse_tests/test_coo.py
```diff
@@ -155,6 +156,14 @@ def test_init_copy_other_scipy_sparse(self, xp, sp):
         assert n.shape == m.shape
         return n
 
+    def test_pickle_roundtrip(self, xp, sp):
+        s = _make(xp, sp, self.dtype)
```
leofang: I think you just need this? 
```suggestion
    def test_pickle_roundtrip(self):
        s = _make(cupy, sparse, self.dtype)
```
(Likewise for all roundtrip tests.)

jakirkham: Huh interesting. I thought this would be tested with NumPy/SciPy first and then CuPy second. Is there a way for me to get that behavior here?

leofang: You can do a quick loop like this: https://github.com/cupy/cupy/blob/7c96b16fe26277f383ebde190cc36d5518940565/tests/cupyx_tests/scipy_tests/sparse_tests/test_coo.py#L491-L492
But what's the point? Don't you just wanna test if CuPy sparse matrices are picklable? You can't compare CuPy's pickle with SciPy's, after all.

jakirkham: Good point.

jakirkham: Fixed in the latest commit 🙂

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

jakirkham:
Sorry missed one thing (now fixed). 😅 Think that is the last thing. 🍀

jakirkham:
Just as an FYI the Jenkins CI status is for the commit before the latest one. So this needs a retest 🙂

emcastillo:
Jenkins, test this please

jakirkham:
Thanks for the reviews! 😄

jakirkham:
@kmaehashi, would it be ok to backport this?

jakirkham:
Backporting with PR ( https://github.com/cupy/cupy/pull/3771 ).

Issue #3158, opened 2020-03-05
Constructing CuPy dia_matrix from SciPy dia_matrix fails


jakirkham:
Following the docs regarding [converting SciPy sparse matrices to CuPy sparse matrices]( https://docs-cupy.chainer.org/en/stable/reference/sparse.html#conversion-to-from-scipy-sparse-matrices ). While it works in other cases, it appears this doesn't work with `dia_matrix`.

```python
In [1]: import numpy                                                                                                                                   

In [2]: import scipy.sparse                                                                                                                            

In [3]: import cupy.sparse                                                                                                                             

In [4]: a_host = numpy.array([[0, 1, 0], 
   ...:                       [2, 0, 3], 
   ...:                       [0, 4, 0]], dtype=float)                                                                                                 

In [5]: asp_host = scipy.sparse.dia_matrix(a_host)                                                                                                     

In [6]: asp_dev = cupy.sparse.dia_matrix(asp_host)                                                                                                     
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-8498d6de7c98> in <module>
----> 1 asp_dev = cupy.sparse.dia_matrix(asp_host)

/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.6/site-packages/cupyx/scipy/sparse/dia.py in __init__(self, arg1, shape, dtype, copy)
     41         else:
     42             raise ValueError(
---> 43                 'unrecognized form for dia_matrix constructor')
     44 
     45         data = cupy.array(data, dtype=dtype, copy=copy)

ValueError: unrecognized form for dia_matrix constructor
```

If it is a bug report, please include the following information:

* Conditions (you can just paste the output of `python -c 'import cupy; cupy.show_config()'`)

```
$ python -c 'import cupy; cupy.show_config()'
CuPy Version          : 7.2.0
CUDA Root             : /datasets/jkirkham/miniconda/envs/rapids13dev
CUDA Build Version    : 10010
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10010
cuBLAS Version        : 10201
cuFFT Version         : 10101
cuRAND Version        : 10101
cuSOLVER Version      : (10, 2, 0)
cuSPARSE Version      : 10300
NVRTC Version         : (10, 1)
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2406
NCCL Runtime Version  : 2507
```

PR #3160, opened 2020-03-05
Allow `dia_matrix` creation from SciPy equivalent

jakirkham:
Fixes https://github.com/cupy/cupy/issues/3158

### DIFF
--- a/cupyx/scipy/sparse/dia.py
+++ b/cupyx/scipy/sparse/dia.py
@@ -34,10 +34,11 @@ class dia_matrix(data._data_matrix):
 
     def __init__(self, arg1, shape=None, dtype=None, copy=False):
         if _scipy_available and scipy.sparse.issparse(arg1):
-            data = arg1.data
-            offsets = arg1.offsets
-            shape = arg1.shape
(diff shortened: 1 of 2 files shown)


jakirkham:
Have pushed some tests. Would it be possible to get this to run on CI @emcastillo? 🙂 Could also use a review if someone is free 😉

cupyx/scipy/sparse/dia.py
```diff
@@ -33,7 +33,13 @@ class dia_matrix(data._data_matrix):
     format = 'dia'
 
     def __init__(self, arg1, shape=None, dtype=None, copy=False):
-        if isinstance(arg1, tuple):
+        if _scipy_available and scipy.sparse.issparse(arg1):
```
emcastillo: Shouldn't check that `arg1` is also stored in dia format?

jakirkham: Good point! Thanks 😄

Should be fixed in the latest commit 🙂

emcastillo:
Jenkins, test this please

PR #3162, opened 2020-03-06
Remove python 2.7 builds

emcastillo:
Remove python2.7 builds from pfnCI.

File names and config file keys should be changed too, but it requires changes to CI configuration.
Kept them now for compatibility

Thanks to @toslunar and @Hakuyume for finding it.


asi1024:
LGTM.

asi1024:
I noticed after merged that the dockerfile should be renamed.

emcastillo:
I agree but it requires changes to CI config 😢

PR #3165, opened 2020-03-06
Implement `cupy.flatiter`

takagi:
Follows #3139 .

I will post several PRs to implement `cupy.flatiter`. It will have similar interface to `cupy.ndarray` as `numpy.flatiter` does to `numpy.ndarray`. For the first, this PR implements its subscript operators so that we can replace `cupy.fill_diagonal`’s implementation using it.

I suppose that `cupy.flatiter` would not provide its iteration interface, as for-loop or the `next` method,  in contrast to `numpy.fatiter` because iteration is essentially serial concept and does not suit to GPU’s pararell computation model.

### DIFF
--- a/cupy/indexing/iterate.py
+++ b/cupy/indexing/iterate.py
@@ -1,0 +1,97 @@
+import numpy
+
+import cupy
+from cupy import core
+from cupy.core import internal
+
+
+class flatiter():
+
+    def __init__(self, a):
+        self._base = a
+
+    def __setitem__(self, ind, value):
+        if ind is Ellipsis:
+            self[:] = value
+            return
+
+        if isinstance(ind, tuple):
+            raise IndexError('unsupported iterator index')
+
+        if isinstance(ind, bool):
+            raise IndexError('unsupported iterator index')
+
+        if numpy.isscalar(ind):
+            ind = int(ind)
+            base = self._base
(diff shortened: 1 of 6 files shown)


takagi:
@emcastillo Please let me ask your advice about `cupy.flatiter`'s iteration interface later.

emcastillo:
As long as the iterator returns `cupy.ndarray` objects I think it makes sense.
Returning the actual values will synchronize the device in every iteration and lose flexibility

takagi:
Sure, I will work as such in another PR!

cupy/indexing/iterate.py
```diff
@@ -0,0 +1,139 @@
...
+                size = size // s
+                indices.append(ind // size)
+                ind %= size
+            base[tuple(indices)] = value
+            return
+
+        if isinstance(ind, slice):
+            base = self._base
+            s = internal.complete_slice(ind, base.size)
+            s_start = s.start
```
emcastillo: nit: use `size = s.stop - s.start` instead of tracking both

takagi: Fixed!

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

PR #3176, opened 2020-03-09
Fix `Event.__del__` behavior on shutdown`

### LINKED ISSUE #3173: map_reduce.py example: AttributeError: 'NoneType' object has no attribute 'eventDestroy'
When running examples/streams/map_reduce.py, I get the following:
```
elapsed time 0.12984609603881836
total bytes 167778304
total bytes 167778304
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC0572E8>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC057278>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC057208>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC057198>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC057128>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC0570B8>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC057048>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC04DEF0>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC04DF28>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
Exception ignored in: <bound method Event.__del__ of <cupy.cuda.stream.Event object at 0x00000180AC057358>>
  File "cupy\cuda\stream.pyx", line 103, in cupy.cuda.stream.Event.__del__
AttributeError: 'NoneType' object has no attribute 'eventDestroy'
```

**_Conditions:_**
CuPy Version          : 7.2.0
CUDA Root             : C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0
CUDA Build Version    : 10010
CUDA Driver Version   : 10010
CUDA Runtime Version  : 10010
cuBLAS Version        : 10200
cuFFT Version         : 10010
cuRAND Version        : 10010
cuSOLVER Version      : (10, 1, 0)
cuSPARSE Version      : 10010
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : None
NCCL Runtime Version  : None

I've had the same results on two different computers (Windows and Linux, both cupy 7.2.0)

**_Code to reproduce_**
[cupy/examples/streams/map_reduce.py](https://github.com/cupy/cupy/blob/master/examples/stream/map_reduce.py) - no changes

emcastillo:
Continuation of #2809 
Closes #3173

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -97,7 +97,9 @@ class Event(object):
                 (interprocess and runtime.eventInterprocess))
         self.ptr = runtime.eventCreateWithFlags(flag)
 
-    def __del__(self):
+    def __del__(self, is_shutting_down=util.is_shutting_down):
+        if is_shutting_down():
+            return
         if self.ptr:
             runtime.eventDestroy(self.ptr)
 


asi1024:
Could you add a test?

emcastillo:
I have no idea how to test this 😄

leofang:
As far as I can tell, there is no way to test shutdown bugs in automated unit tests.

jakirkham:
At first I would assume this is impossible (if not very hard) to test, but looking at [`is_shutting_down`'s behavior]( https://github.com/cupy/cupy/blob/79812ef247a241d569f983403f6a53b07f01a654/cupy/util.pyx#L230 ) makes me think otherwise. Would it be possible to mock the shutting down process by subbing out this function and having it return `True`?

emcastillo:
>> At first I would assume this is impossible (if not very hard) to test, but looking at is_shutting_down's behavior makes me think otherwise. Would it be possible to mock the shutting down process by subbing out this function and having it return True?

I thought about this, but then the test doesn't make much sense IMHO.

jakirkham:
What if we nullified the pointer in the other case? That way we check if the pointer here is non-null. WDYT?

emcastillo:
I think it is a good idea. I will implement it.
However, I don't feel like we are testing the real use case (interpreter shutdown).

jakirkham:
Sounds good 🙂

Yeah that's fair. I don't think that should block this PR. Though it would be good if we can find a solution for this. Future work though 😉

emcastillo:
Ok, there are not Event tests in cupy.
Maybe all this should be done in a separated PR.

leofang:
I was wrong. We could perhaps test it like this: run a subprocess with the command `python -c "import cupy; cupy.cuda.stream()"` (or create an event) and check if anything is printed to stdout/stderr.

asi1024:
I expected to mock `is_shutting_down`, but I agree to do it in a separated PR.

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

PR #3183, opened 2020-03-10
Remove unused, deprecated fields from `cudaPointerAttributes`

### LINKED ISSUE #2768: Deprecate warning for some Python and CUDA codes

leofang:
Close #2768.

This PR does what I suggested in https://github.com/cupy/cupy/issues/2768#issuecomment-596320247:
> As far as I can tell, these attributes are never used in the codebase. How about just removing them entirely so as to suppress the deprecation warning? We're gonna remove them when they are removed from CUDA anyway.

### DIFF
--- a/cupy/cuda/cupy_cuda_common.h
+++ b/cupy/cuda/cupy_cuda_common.h
@@ -77,8 +77,6 @@ struct cudaPointerAttributes{
     int device;
     void* devicePointer;
     void* hostPointer;
-    int isManaged;
-    int memoryType;
 };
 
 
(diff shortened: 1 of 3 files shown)


emcastillo:
Jenkins, test this please

leofang:
Thanks, Emilio!

kmaehashi:
Backported via #3720, removing `to-be-backported` to avoid bot comments.

Issue #3111, opened 2020-02-22
Implemention of `cupyx.scipy.ndimage.filters`


coderforlife:
I have implemented nearly all of `cupyx.scipy.ndimage.filters` matching `scipy.ndimage.filters` exactly. However, before I finish it I would like to confirm that I am developing it correctly. I have the file currently in a gist: https://gist.github.com/coderforlife/d953303da4bb7d8d28e49a568cb107b2

This works towards https://github.com/cupy/cupy/issues/2099. One of their requests was one of the missing features though.

Current Lacking Features:
- Does not have the implementations of `generic_filter()` or `generic_filter1d()`.
- I want to make some changes to how rank filters work (includes `rank_filter`, `median_filter`, `percentile_filter`) based on the answers below.
- There are no tests.
- There are no function docs (but most would just refer to the scipy docs).

Changes From Current `cupyx.scipy.ndimage.filters`:
- Implements tons of new functions (and they re-use slightly adjusted code that was already there).
- `convolution()` and `correlate()` were adjusted to support non-contiguous input arrays without copying them (weights are still forced to be contiguous, but they are expected to be quite small).
- `convolution()` and `correlate()` are slightly (~5%) faster now.
- The `ElementwiseKernel`s generated are now more runtime adjustable.

Open Questions:
- The `ElementwiseKernel`s are only used internally and are never returned. Should they have as much as possible compiled into them? The current version of the code compiles the number of dimensions, input shape, weights shape, border mode, constant value, and origin values. My version moves input shape (and actually strides), constant value, and origin values to be runtime values. Is it better to have more compile time values and thus need to compile new `ElementwiseKernel`s more frequently or have more things moved to runtime values and need fewer compiles?
- To implement `generic_filter()` and `generic_filter1d()` I think they should take a `ReductionKernel` (or a Python function that can be fused into a `ReductionKernel`). However, Cuda does not let you call functions in other modules. So to accomplish this I was thinking about re-creating the reduction kernel code in the convolution kernel like at https://github.com/cupy/cupy/blob/de3d32536c39eefea1805f4a7a76a14356eb3bff/cupy/core/_reduction.pyx#L36 however it would actually be simpler since the entire reduction would happen in a single thread and not need synchronization. Is this a reasonable approach?
- Is it okay that numpy is used to create some of the filters internally? It is several orders of magnitude faster than doing it in cupy, even at the large kernels of 25.
- Is the quality of the code okay?

Thanks for your consideration.

[...]

Issue #2099, opened 2019-03-14
Adding equivalent filters from scipy.ndimage.filters


jakirkham:
Would be great to support some common filters from [`scipy.ndimage.filters`]( https://docs.scipy.org/doc/scipy-1.0.0/reference/ndimage.html#module-scipy.ndimage.filters ). Having `convolve`, `correlate` and `generic_filter` would be great and allow users to build up many common convolutions with relatively little work using a familiar API. There are others that would be very nice as well.

kmaehashi:
Thanks for the feature requests.
We always welcome contributions for such missing features!
https://docs-cupy.chainer.org/en/latest/reference/comparison.html#multidimensional-image-processing

grlee77:
Hi @jakirkham. Do you have an immediate need for this or a particular use case in mind?

I recently implemented the equivalent of `ndi.convolve` (built using a GPU-based implementation of  `scipy.signal.upfirdn(..., up=1, down=1)` under the hood) that supports float32, float64, complex64 and complex128. I haven't tried to make a PR for it here or posted it publicly, but can share it if you are interested in collaborating on it.

I implemented `scipy.signal.upfirdn` using CuPy and added support for various signal extension modes at the boundary (not currently available in SciPy's `upfirdn`). In my case, I started implementing a library expanding the functionality in PyWavelets. The version of `upfirdn` I have currently only supports downsampled convolution with `up=1, down=M` or upsampled convolution with `up=M, down=1` as those were the only cases I needed for M-band wavelet filterbanks. With a bit more effort, it could be updated to allow the general setting of any positive integer `up`, `down` and become suitable for a PR here.

jakirkham:
Not immediate, but it is something I can see use cases coming up for.

That does sound interesting. Yeah, if you would be willing to share, it would be great to discuss.

Ah ok. Thanks for the context. Is this in PyWavelets currently or in a library built on top of it?

jakirkham:
Just as another thought, it might be worth looking at how [`chainer.functions.convolution_nd`]( https://docs.chainer.org/en/stable/reference/generated/chainer.functions.convolution_nd.html ) is implemented. Likely much of that code could be reused here as well. Perhaps even simplified a bit as this has fewer features.

grlee77:
Yeah, that would be a good place to look for general n-dimensional convolution. 

[...]

PR #3184, opened 2020-03-11
Implementation of ndimage filters

coderforlife:
So far this PR includes improved `correlate` and `convolve` functions (in terms of speed, uses similar technique to #3179) along with implementations of `correlate1d` and `convolve1d` along with tests for them. The underlying kernel creation has been generalized so that it can be used to implement all of the other filters and those will be progressively added to this PR (I have an implementation of them but haven't tested them yet).

This works to address #2099 and #3111.

Current status of tests is that it is passing 7120 tests (in test_filters.py) and failing 8 (they are all with the 1d functions along axis 0 using mode=mirror with 4D images but no other axis or mode or less-dimension image).


jakirkham:
I wonder if we should exposed the 1-D convolve as `cupy.convolve` (analogous to [`numpy.convolve`]( https://docs.scipy.org/doc/numpy/reference/generated/numpy.convolve.html )). Thoughts? 🙂

coderforlife:
@jakirkham It is possible, a quick idea of how to do it would be. Note I have never used `numpy.convolve` or `numpy.correlate` so I don't know details like what do they do with >1D data or how some mode center the arrays. However, it gives the basic idea.

```
def convolve(a, v, mode='full'):
    if mode not in ('full', 'same', 'valid'): raise ValueError('mode') 

    # Make both 1D arrays and make sure `a` is the larger array
    a, v = a.ravel(), v.ravel()
    if len(a) == 0: raise ValueError('a cannot be empty')
    if len(v) == 0: raise ValueError('v cannot be empty')
    if len(v) > len(a): a, v = v, a

    # Deal with full padding of data
    if mode == 'full':
        pad = len(v) - 1
        a = cupy.pad(a, (pad // 2, (pad + 1) // 2)
    
    # Perform convolution
    out = cupyx.scipy.ndimage.convolve1d(a, v, mode='constant')

    # Return result
    if mode == 'valid':
        remove = 1 - len(v)
        out = out[remove//2 : -(remove+1)//2]
    return out
```

The downsides of this compared to `numpy.convolve()`:

- The default `mode` of `'full'` requires padding the data (and thus copying the data).
- The `ndimage.convolve1d()` function assumes that one of `a` or `v` is relatively small. If you were to use this for a cross-correlation of a large dataset (and thus `len(a) == len(v)`), it would likely have low performance.

coderforlife:
The 8 tests that are not passing are actually apparently a bug in Scipy. I have submitted a bug report over there: https://github.com/scipy/scipy/issues/11661. It does not handle `mirror` correctly for length-1 dimensions with the 1D functions.

awthomp:
@coderforlife -- I'm interested in CuPy's support/implementation of `numpy.convolve` and `numpy.correlate` for [cusignal](https://github.com/rapidsai/cusignal). I'd be more than happy to test and guide your implementation.

>  I don't know details like what do they do with >1D data or how some mode center the arrays
For NumPy's convolve/correlate functions, only 1D arrays are supported. This is actually probably the motivation for creating 2D support in ndimage. As you mentioned in your post, the key component here is how padding is handled in the implementation; and I see that as the big difference between doing a 1D `ndimage.convolve` or `ndimage.correlate` as well.

I took a quick-ish spin of `ndimage.correlate` and noticed that it doesn't do the 'sliding window' thing that numpy.convolve and numpy.correlate do. This 'sliding dot product' is the key difference between the ndimage and standard numpy implementation, IMO.

Anyway, as you may know, `numpy.correlate` and `numpy.convolve` are basically the same algorithm. For convolve, the filter is mirrored left-to-right before beginning the sliding dot product. To step through an example:

```python
a = [1, 2, 3]
b = [0, 1, 0.5]
y = np.convolve(a, b, mode="full")
```

Yields:
`array([0. , 1. , 2.5, 4. , 1.5])`

Convolution/correlation is a point by point overlapping and sliding dot product that starts with a mirrored version of the filter (b) and the data (a). The last element in the mirrored filter is overlapped with the first element of the data, and the dot product is taken as the filter moves element by element over the data. To be explicit:

mirrored_b (denoted by `m_b`), used for calculation = `[0.5, 1, 0]`

The convolution works like:

```python
y[0] = m_b[2] * a[0] = 0 * 1 = 0
y[1] = m_b[2] * a[1] + m_b[1] * a[0] = 0 * 2 + 1 * 1 = 1
y[2] = m_b[2] * a[2] + m_b[1] * a[1] + m_b[0] * a[0] = 0 * 3 + 1 * 2 + 0.5 * 1 = 2.5
y[3] = m_b[1] * a[2] + m_b[0] * a[1] = 1 * 3 + 0.5 * 2 = 4
y[4] = m_b[0] * a[2] = 0.5 * 3 = 1.5
```

Here's a visualization of what's going on too: https://en.wikipedia.org/wiki/Convolution#/media/File:Convolution_of_box_signal_with_itself2.gif

Again, correlation is very similar, but you don't have to flip the filter component (`b`, in this case).

coderforlife:
@awthomp I know that convolution and correlation are essentially the same thing with mirrored 'kernels', 'weights', or 'filter' depending on what you call the smaller of the two arrays. `ndimage` uses that fact. Also, the `ndimage` algorithm does use overlap-add. The `ElementwiseKernel` does the following for each element: sum every neighbourhood value multiplied by the corresponding weight. This is exactly what you describe. It may not look like it is doing this since it isn't "sliding" across but instead doing it per-element.

For the 1D case, it does this along an axis instead of in all dimensions. Unlike `numpy.convolve/correlate` it handles the edges differently.

The following code demonstrates the equality:
```
a = np.arange(24)
b = np.ones(3)
(np.convolve(a, b, mode="same") == ndi.convolve1d(a, b, mode="constant")).all()
```
where `ndi` is `scipy.ndimage`. The only other reasonable way to do a convolution is through FFT and it definitely doesn't do that.

I have corrected the code from above to properly align the output with what `numpy.convolve` generates and forbid non-1D arrays (and one other bug I had), and `cupy.convolve` could be implemented as:

```
def convolve(a, v, mode='full'):
    if mode not in ('full', 'same', 'valid'): raise ValueError('mode') 

    # Check both arrays and make sure `a` is the larger array
    if a.ndim != 1: raise ValueError('a must be 1D')
    if v.ndim != 1: raise ValueError('v must be 1D')
    if len(a) == 0: raise ValueError('a cannot be empty')
    if len(v) == 0: raise ValueError('v cannot be empty')
    if len(v) > len(a): a, v = v, a

    # Deal with full padding of data
    if mode == 'full':
        pad = len(v) - 1
        a = cupy.pad(a, ((pad+1)//2, pad//2), 'constant')
    
    # Perform convolution
    out = cupyx.scipy.ndimage.convolve1d(a, v, mode='constant')

    # Return result
    if mode == 'valid':
        remove = len(v) - 1
        out = out[remove//2 : -(remove+1)//2]
    return out
```

This should handle all cases but may not be the most efficient for all cases. In all likelihood, `correlate` will be implemented like the above and then `convolve` will be implemented in terms of `correlate`.

Implementing `scipy.signal.convolve/correlate` is a whole different can of worms. There is also [`fftconvolve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.fftconvolve.html#scipy.signal.fftconvolve) and [`oaconvovle`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.oaconvolve.html#scipy.signal.oaconvolve) (oa = 'overlap-add' which is what we are using here). Both of those functions say they are much faster than [`convolve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve.html#scipy.signal.convolve) but `convolve` just uses those functions... they do confirm that the OA method is only efficient when `len(a)` >> `len(b)` or vice-versa. There is also  [`choose_conv_method`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.choose_conv_method.html#scipy.signal.choose_conv_method) for determining which to call.

awthomp:
Hi @coderforlife -- We've actually taken care of the [correlate](https://github.com/rapidsai/cusignal/blob/d64df2fc2c40bbdba4aad34641460176552cd5f6/python/cusignal/signaltools.py#L93)/[convolve ](https://github.com/rapidsai/cusignal/blob/d64df2fc2c40bbdba4aad34641460176552cd5f6/python/cusignal/signaltools.py#L645)in cusignal. For large arrays/filters, we're getting a 1-2 order of magnitude perf gain over Scipy Signal, but -- as you can see -- we default to numpy (rather than invoking an fft) for small array sizes. This isn't ideal.

Much of cusignal is based on changing SciPy Signal Numpy calls for CuPy; when there's not support (like correlate2d or upfirdn), we wrote custom Numba CUDA kernels and are now exploring Raw CuPy Modules.

Usage examples are here: https://github.com/rapidsai/cusignal/blob/d64df2fc2c40bbdba4aad34641460176552cd5f6/notebooks/api_guide/signaltools_examples.ipynb

grlee77:
Hey @awthomp, I had implemented `scipy.signal.upfirdn` a little over a year ago (prior to discovering your version in cusignal). At that time I had also reused it to implement `numpy.convolve` and `numpy.correlate` (I had added a couple arguments to control the extent and origin in upfirdn). The implementation of upfirdn differs from cusignal's in that it uses RawKernel rather than Numba. I have finally also made this version public (see https://github.com/cupy/cupy/issues/2099#issuecomment-598381598).

awthomp:
@grlee77 Awesome! Looking forward to seeing your work. In the meantime, we *also* implemented the 1D double precision `upfirdn` in cusignal earlier this week (https://github.com/rapidsai/cusignal/pull/25). @mnicely is currently looking at optimization of it with this WIP PR: https://github.com/rapidsai/cusignal/pull/31. We'd love your eyes on, if you have an opportunity and desire to contribute.

cupyx/scipy/ndimage/filters.py
```diff
@@ -373,10 +313,21 @@ def _get_nd_kernel(name, pre, found, post, mode, wshape, int_type,
...
     for j in range(ndim):
-        boundary = _generate_boundary_condition_ops(mode, 'ix_{}'.format(j),
-                                                    'xsize_{}'.format(j))
+        if wshape[j] == 1:
+            loops.append('{{ {type} ix_{j} = ind_{j} * xstride_{j};'.
+                         format(j=j, type=int_type))
+        else:
+            boundary = _generate_boundary_condition_ops(
+                mode, 'ix_{}'.format(j), 'xsize_{}'.format(j))
         loops.append('''
```
grlee77:
```suggestion
            loops.append('''
```

grlee77: Thanks for refactoring to only avoid duplicate 1d and nd code paths!

I can confirm good performance of the refactored kernel. The only problem I encountered was the indentation error above.
The 1Dd case now performs approximately the same as what I had in `cupyimg`. 3D cases I tested are up to 25% faster for this kernel.

cupyx/scipy/ndimage/filters.py
```diff
@@ -76,40 +164,105 @@ def _get_output(output, input, shape=None):
...
 
 
-def _correlate_or_convolve(input, weights, output, mode, cval, origin,
-                           convolution):
-    if input.dtype.kind == 'c':
-        raise TypeError('Complex type not supported.')
-    if not hasattr(origin, '__getitem__'):
-        origin = [origin, ] * input.ndim
+def _fix_sequence_arg(arg, ndim, name, conv=lambda x: x):
+    if hasattr(arg, '__iter__') and not isinstance(arg, str):
```
emcastillo: we favor try/except AttributeError over `hasattr`

cupyx/scipy/ndimage/filters.py
emcastillo: please address this!
thanks 🙂

cupyx/scipy/ndimage/filters.py
```diff
@@ -76,40 +164,105 @@ def _get_output(output, input, shape=None):
...
+        axis += ndim
+    if axis < 0 or axis >= ndim:
+        raise ValueError('invalid axis')
+    return axis
+
+
+def _convert_1d_args(ndim, weights, origin, axis):
+    if weights.ndim != 1 or weights.size < 1:
+        raise RuntimeError('incorrect filter size')
+    axis = _check_axis(axis, ndim)
```
emcastillo: use `cupy.util._normalize_axis_index` as it's essentially the same code

grlee77:
> The 8 tests that are not passing are actually apparently a bug in Scipy. I have submitted a bug report over there: scipy/scipy#11661. It does not handle mirror correctly for length-1 dimensions with the 1D functions.

A fix for this will be included in SciPy 1.5 (scipy/scipy#11683). You can avoid the failing test cases by decorating those cases with:
```Python
@testing.with_requires('scipy>=1.5.0')
```

cupyx/scipy/ndimage/filters.py
```diff
@@ -62,6 +63,93 @@ def convolve(input, weights, output=None, mode='reflect', cval=0.0, origin=0):
                                   True)
 
 
+def correlate1d(input, weights, axis=-1, output=None, mode="reflect", cval=0.0,
+                origin=0):
+    """One-dimensional correlate.
+    The array is correlated with the given kernel.
```
emcastillo: Please add blank lines between the summary and the description as in the other CuPy docstrings

cupyx/scipy/ndimage/filters.py
```diff
@@ -62,6 +63,93 @@ def convolve(input, weights, output=None, mode='reflect', cval=0.0, origin=0):
...
+            (``'reflect'``, ``'constant'``, ``'nearest'``, ``'mirror'``,
+            ``'wrap'``). Default is ``'reflect'``.
+        cval (scalar): Value to fill past edges of input if mode is
+            ``'constant'``. Default is ``0.0``.
+        origin (int): The origin parameter controls the placement of the
+            filter, relative to the center of the current element of the
+            input. Default is ``0``.
+    Returns:
+        cupy.ndarray: The result of the 1D correlation.
+    .. seealso:: :func:`scipy.ndimage.correlate1d`
```
emcastillo: Blank line here before the `seealso` too

cupyx/scipy/ndimage/filters.py
```diff
@@ -62,6 +63,93 @@ def convolve(input, weights, output=None, mode='reflect', cval=0.0, origin=0):
...
+    """
+    weights, origins = _convert_1d_args(input.ndim, weights, origin, axis)
+    return _correlate_or_convolve(input, weights, output, mode, cval, origins,
+                                  False)
+
+
+def convolve1d(input, weights, axis=-1, output=None, mode="reflect", cval=0.0,
+               origin=0):
+    """One-dimensional convolution.
+    The array is convolved with the given kernel.
```
emcastillo: Ditto

jakirkham:
cc @anaruse (for awareness)

emcastillo:
Hello @coderforlife!

Do you plan to keep working on this PR or should an interested person take over it.
Thanks!

pentschev:
Just a quick update on this: as per our discussion earlier this week, I've asked internally at NVIDIA and @mnicely will take a look at this PR next week to figure if this is something we can pick up.

mnicely:
@leofang @emcastillo 
Can you catch me up on what's left for this PR? 
Is it just fixing those 8 failed tests? Or does the code need to be optimized?
Is there a timetable you're trying to meet?

leofang:
Hi @grlee77 @coderforlife: could you help address @mnicely's questions on this PR? I didn't follow the discussion here (but I better do so) 😅

As for the time table, I think @kmaehashi @emcastillo can answer. Since v8.0.0b3 was pinned yesterday, the next deadline (v8.0.0b4) would be in June according to https://github.com/chainer/chainer/issues/8560.

leofang:
I think it'd be nice to resolve the conflicts first, and then address the pending reviews.

grlee77:
> Is it just fixing those 8 failed tests? Or does the code need to be optimized?

I think it is mostly just skipping those failing tests for SciPy<1.5. When I had tested this code is a bit more efficient than the existing implementation for convolve/correlate.

The main things remaining are adding all of the other specific filters, such as `gaussian_filter`, `uniform_filter`, etc. Those can be built easily just using these functions. I think those can wait for a follow-up PR, though and don't have to be in this one.

I actually have almost all of these already implemented in [cupyimg](https://github.com/mritools/cupyimg), but there are a couple of issues there.
1.) I have tests, but many are just copies of the SciPy ones and are not in the CuPy style.
2.) Those implementations have a `dtype_mode` keyword-only argument that doesn't exist in SciPy. Namely, SciPy always does the convolutions in double precision, but I had added an option to allow single-precision computations instead if desired. I'm not sure if the CuPy team would allow this when porting here.
3.) The versions in that repository also allow complex-valued inputs while SciPy and the implementations in this PR currently do not (I do plan to add support for that upstream in SciPy, but it is not there yet).

jakirkham:
> As for the time table, I think @kmaehashi @emcastillo can answer. Since v8.0.0b3 was pinned yesterday, the next deadline (v8.0.0b4) would be in June according to [chainer/chainer#8560](https://github.com/chainer/chainer/issues/8560).

We probably need a better idea of what June means here. For example June 1st is Monday, which isn't a lot of time (though likely is not what is intended give the date for v8.0.0b4). Asked in comment ( https://github.com/chainer/chainer/issues/8560#issuecomment-635543853 ) for clarification.

mnicely:
@leofang I reiteration @jakirkham questions about June.

I'll give it a try next week. The near-term goal will be to get errors fixed so the code can be merged. Based on @grlee77's comments, this should be straight-forward.

emcastillo:
Thanks for the follow up guys!

mnicely:
@leofang @emcastillo @grlee77 
What do you suggest the best way to pull this PR and get started? (Basic GH help)
1. Create fork from CuPy, create branch, copy @coderforlife's changes into my branch then submit another PR?
2. Create fork from https://github.com/coderforlife/cupy/tree/ndimage_filters, submit changes to that branch which update the current PR (if permissions are set)

leofang:
I think it'd depend on how fast @coderforlife sees the notifications and act 😅

grlee77:
> What do you suggest the best way to pull this PR and get started? (Basic GH help)

I would recommend creating a new branch based off of this one using something like

The first two lines checks out this existing branch and the last line creates a new branch starting from this one.
```
git remote add coderforlife git@github.com:coderforlife/cupy
git checkout coderforlife/ndimage_filters
git checkout -b ndimage_filters_v2
```
You can then push this branch to your CuPy fork. Assuming you are using `origin` as the name for your fork this would be:
```
git push origin ndimage_filters_v2
```
You can then create a new PR based off of your new `ndimage_filters_v2` branch and add your new commits there. That will preserve the commit history and credit to @coderforlife.

cupyx/scipy/ndimage/filters.py
mnicely: @emcastillo Are you referring to function `_check_axis`? To be replaced by `cupy.util._normalize_axis_index`

cupyx/scipy/ndimage/filters.py
mnicely: @emcastillo Can you point me to some example code you want me to use?

Just curious why y'all prefer try/except over `hasattr`?
https://stackoverflow.com/questions/903130/hasattr-vs-try-except-block-to-deal-with-non-existent-attributes

mnicely:
# Question

I've started getting spun up on this PR and implementing @emcastillo comments above. Initial testing produced a lot of error with multi-device. See file 
[ndimage_filter_initial_test.txt](https://github.com/cupy/cupy/files/4711665/ndimage_filter_initial_test.txt)

I'm tried merging master into my ndimage_filter_v2 branch, but now I'm even more errors. 
[ndimage_filter_masterMerge_test.txt](https://github.com/cupy/cupy/files/4712097/ndimage_filter_masterMerge_test.txt)

Any advice? How do I rerun the test to see the original 8 error?

cupyx/scipy/ndimage/filters.py
emcastillo: Sorry, yes, we use try/ except AttributeError because it is faster.
If you think that in this routine speed doesn't matter you can leave it as is 🙂

cupyx/scipy/ndimage/filters.py
emcastillo: Yes! if possible :)

cupyx/scipy/ndimage/filters.py
jakirkham: I wonder if it would be cleaner and faster to check if `arg` is not a `str` and then just do `arg = iter(arg)`.

mnicely: So I tried swapping
```python
if hasattr(arg, '__iter__') and not isinstance(arg, str):
```
with 
```python
if not isinstance(arg, str):
    try:
        iter(arg)
    except AttributeError:
        return NotImplemented
```
Now most/all tests are failing. Are these not equivalent?

cupyx/scipy/ndimage/filters.py
mnicely: Completed in #3390

cupyx/scipy/ndimage/filters.py
mnicely: Completed in #3390

cupyx/scipy/ndimage/filters.py
mnicely: Completed in #3390

cupyx/scipy/ndimage/filters.py
mnicely: Completed in #3390

cupyx/scipy/ndimage/filters.py
mnicely: Completed in #3390

cupyx/scipy/ndimage/filters.py
emcastillo: The exception for an invalid iter with `iter` is
```
TypeError: 'object' object is not iterable
```
😅 this might be confusing

cupyx/scipy/ndimage/filters.py
emcastillo: Thanks!

coderforlife:
Hello. I know that I completely disappeared from this. When the pandemic struck my job skyrocketed in the amount of work and I had time for nothing else. Things are significantly calmer now and I have time to pick this back up.

I would be happy to integrate the few changes suggested above:

- indentation problem which I remember fixing, at least locally, right as I has to drop everything
- flagging some tests for only scipy 1.15 and higher
- optionally an hasattr/AttributeError switch out)
- fixing another other merge issues

leofang:
Hi @coderforlife Welcome back! Not sure if you have noticed, @mnicely from NVIDIA has taken over your PR and is continuing it in #3390. Perhaps you better wait for his reply and coordinate with him...

coderforlife:
Fixed most issues brought up here and in #3390 I believe except the following:

1. Testing the speed with using `CArray` again now that it is smarter
2. A few more tests should be added for some edge cases

jakirkham:
No worries. We were just trying to make sure we got your improvements integrated 🙂

Glad to have you back 😄

grlee77:
I didn't review again recently, but if it is not there already, can you add a test case for Fortran-ordered and/or non-contiguous inputs to make sure it has correct behavior for these inputs?

One option is to always cast the input to C-contiguous so CArray indexing will be fast. On the other hand, there may be memory-limited cases where filtering non-contiguous inputs is desired and allowing it would help avoid a copy even if performance is a bit worse. What do you think?

coderforlife:
I re-implemented it to use `CArray` instead of hacking around it and it doesn't work so well... avoiding `CArray` is still faster (much faster in some cases). Below is a bunch of results of running various images and kernels with Scipy, the `CArray` avoiding code (current PR), and the code I wrote to use `CArray`. I also check that all of the results are equal to the results of Scipy.

```
----- 4000x4000 image, 3x3 kernel -----
scipy:        202 ms ± 192 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 670 µs ± 281 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
using CArray: 749 µs ± 682 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
----- 4000x4000 image, 5x5 kernel -----
scipy:        413 ms ± 708 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 1.02 ms ± 174 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
using CArray: 1.31 ms ± 471 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
----- 4000x4000 image, 7x7 kernel -----
scipy:        740 ms ± 1.48 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 1.49 ms ± 371 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
using CArray: 2.03 ms ± 114 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
----- 4000x4000 image, 3x3 kernel [Fortran] -----
scipy:        237 ms ± 123 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 1.69 ms ± 83.2 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
using CArray: 1.7 ms ± 73.6 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
----- 4000x4000 image, 5x5 kernel [Fortran] -----
scipy:        482 ms ± 467 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 4.25 ms ± 221 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
using CArray: 4.27 ms ± 1.76 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
----- 4000x4000 image, 7x7 kernel [Fortran] -----
scipy:        817 ms ± 1.04 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 8.13 ms ± 1.02 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
using CArray: 8.11 ms ± 655 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
----- 300x300x300 image, 3x3x3 kernel -----
scipy:        740 ms ± 445 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 1.96 ms ± 256 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
using CArray: 2.76 ms ± 455 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)
----- 300x300x300 image, 3x3x3 kernel [Fortran] -----
scipy:        1.4 s ± 6.74 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
avoid CArray: 43.1 ms ± 10.9 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
using CArray: 127 ms ± 33.2 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

In summary, 2D C-ordered data is still 10-30% faster avoiding `CArray`, 2D Fortran-ordered data ties, 3D C-ordered data is 30% faster avoiding `CArray`, and 3D Fortran-ordered data is 65% faster.

emcastillo:
ok! lets forget about carray for now 😅

coderforlife:
While testing, I found and fixed a bug with the minimum and maximum filters that caused an issue when a custom footprint was used that had a `False`/`0` at the origin. I am working on cleaning up the tests now and will commit the completed version tomorrow.

coderforlife:
I was thinking about why `CArray` is still slower, I came up with the following:

- `CArray` insists on using 64-bit integers for indices even for small arrays, my code "adapts" and uses `int` when the number of elements is <2<sup>31</sup> and `size_t` otherwise (note that it isn't actually just size that matters, it really is the difference in position between the first and last element, which can be larger than size for non-contiguous arrays)
- `CArray` has to do a multiplication for each dimension for every single indexing operation, however in the code above each dimension only does it's multiplication once for each change and then is saved. This would never be possible with `CArray` since it has to assume random-access while we have regular forward-only iteration over the data (and we know when we are going to jump much further forward).

emcastillo:
Hi,
Thanks for the comments on `CArray` issues.

> CArray insists on using 64-bit integers for indices even for small arrays, my code "adapts" and uses int when the number of elements is <2^31 and size_t otherwise (note that it isn't actually just size that matters, it really is the difference in position between the first and last element, which can be larger than size for non-contiguous arrays)

Yes, using int64 indexes for arrays which don't need large indexing is overkill due to the size of the array. Maybe we can come up with some kind of dynamic dispatcher for this issue. I will try to work on this

coderforlife:
Is there a way to use a combination of `@testing.numpy_cupy_raises` and `@testing.numpy_cupy_allclose` on the same function? Basically, the test passes if both raise similar exceptions OR if the results are all close.

emcastillo:
The ``@testing.numpy_cupy_allclose(accept_error=...)` decorator should do the trick :)

coderforlife:
@emcastillo Great! I can't believe it is that simple...

I have committed the working version including a major overhaul to the testing to allow more edge cases to be easily tested and for additional filters to be added in the future (in a separate PR).

I have run tests against with scipy 1.5.0rc2 (and also 1.4.1).

The Travis build did report some whitespace issues (I used pylint which doesn't report those issues apparently). I will fix them but at that point everything should be good.

coderforlife:
> However, if the array is c_contiguous, the [] operator should not perform any operation other than the indexing itself. Are the arrays c_contiguous in this case?

The weights array is forced to be C-contiguous always since those are assumed to be relatively small. The input data is however used as-is (at least for me this can be gigabytes in size and thus a copy is not reasonable). In the timing data I gave before, the non-Fortran ones were C-contiguous. The "indexing" isn't just "indexing" though. If I do `x[inds] ` where `inds` is an array for the indices then it has to do 1 multiplication per dimension plus several additions during every index operation. Since the convolution kernel code knows it is predictably forward-moving it only does the multiplication when it actually changes a dimension. This likely doesn't make a huge difference, especially if the indices were all 32-bits when possible, but it does likely contribute a little.

emcastillo:
I see!
Thanks a lot for the detailed explanation. I am currently working on a PR to use 32bits when the total size of the array is less than 4GB. But I won't ask you to use CArray in this PR because the performance won't improve at all. (It will be a little worse)

coderforlife:
It is only like 3 lines of code that need to change to switch over to `CArray`. Can be done at a later time. I did switch the weights array over to using `CArray` already.

For determining if 32-bit or 64-bit ints should be used:
```
    nbytes = sum((x-1)*strd for x, strd in zip(input.shape, input.strides))
    nbytes += input.dtype.itemsize
    int_type = 'int' if nbytes < (1 << 31) else 'ptrdiff_t'
```

The reason this doesn't just use `nbytes = input.nbytes` is for non-contiguous arrays: you need to determine the largest linear index possible (i.e. after multiplying by all of the strides).

Just realized however that negative strides are possible, so maybe it needs to be `abs(nbytes) < (1 << 31)`.

coderforlife:
Actually, it turns out negative strides are a bit more complicated. An over-estimate can be obtained by taking the absolute value of each stride while summing (and no need to absolute the final result). This will get it right when all strides are negative and will overestimate the value when there are mixed negative and positive strides (at least overestimates are way better than underestimates).

To do it correctly, I think you would have to do the following with mixed sign strides:

```python
neg = sum((x-1)*max(-stride, 0) for x, stride in zip(a.shape, a.strides))
pos = sum((x-1)*max(stride, 0) for x, stride in zip(a.shape, a.strides))
nbytes = max(neg, pos) + a.dtype.itemsize
```

coderforlife:
One other thing to point out that you may already have thought of is that the kernel that is using the `CArray` also needs to be aware of the types to use. There will need to be a type like `X_CArray::index_t` for the `X` parameter and so forth.

coderforlife:
@emcastillo Should I commit a version that has comments in the `_generate_nd_kernel()` function on what lines need to change and to what to switch it to using `CArray`?

emcastillo:
Probably using CArray won't be as fast as this so it's fine.
I will review the PR during the day. Thanks!

emcastillo:
@grlee77 can you give it another look? :)
Some of the details escape my comprehension so a bit of help would be really appreciated.
Thanks!

grlee77:
> Some of the details escape my comprehension so a bit of help would be really appreciated.
Thanks!

Yes, I can look again.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_filters.py
```diff
@@ -12,150 +11,335 @@
...
+            'filter': ['minimum_filter', 'maximum_filter'],
+            'footprint': [False, True],
+        }),
+
+        # Mode-specific params
+        testing.product({
+            **COMMON_PARAMS,
+            'mode': ['reflect'],
+            # With reflect test some of the other parameters as well
+            'origin': [0, 1, None],
```
grlee77, line 139: I suggest restricting shape for reflect only in order to cut down on the number of combinations that need to be tested. Just testing the other shapes for the rest of the modes should be good enough.

(This change reduced the number of tests from 7696 to 4500 when I tried it).

```suggestion
            'origin': [0, 1, None],
            'shape': [(4, 5)],
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_filters.py
```diff
@@ -12,150 +11,335 @@
...
+    that uses the footprint. These methods allow testing across multiple
+    filter types much more easily.
+    """
+
+    # default param values if not provided
+    filter = 'convolve'
+    shape = (4, 5)
+    ksize = 3
+    dtype = numpy.float64
+    footprint = True
```
grlee77, line 29: I had previously suggested potentially testing inputs with order='F' to make sure everything is okay for inputs that don't have C order. To do that we could add an order parameter here that defaults to 'C' and then after `shaped_random` call:

```Python
if self.order == 'F':
    arr = xp.asfortranarray(arr)
```

cupyx/scipy/ndimage/filters.py
```diff
@@ -78,38 +345,123 @@ def _get_output(output, input, shape=None):
...
+    The weights are the only optional part and can be passed as None and then
+    one less argument is passed to the kernel. If the output is given as None
+    then it will be allocated in this function.
+
+    This function deals with making sure that the weights are contiguous and
+    float64 or bool*, that the output is allocated and appriopate shaped. This
+    also deals with the situation that the input and output arrays overlap in
+    memory.
+
+    * weights is always casted to float64 or bool in order to get an output
```
grlee77, line 446:
```suggestion
    * weights is always cast to float64 or bool in order to get an output
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_filters.py
coderforlife: 'reflect' was being used to test all the combinations, the origin ones actually really put the multiple dimensions to the test. I would be more comfortable dropping some of the shapes from the other modes.

coderforlife: I would be happier with reducing the `dtype` list 1 line later since that is tested similarly later. It is also longer so will reduce the tests even more.

coderforlife: When removing the `dtype` there and also adding a F-ordered tests later it is down to 4944 (with 1000 skipped due to redundancies).

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

coderforlife:
It looks like I forgot to decorate the `TestFortranOrder` tests with `@testing.with_requires('scipy')`. Sorry about that. I just committed that fix.

However, the vast majority of the failures were from the morphology code.

coderforlife:
Looks good (except morphology tests, someone else needs to fix that). I am curious about the coverage test, is it being run without scipy? It says none of my code (except the function declarations and imports) is covered...

emcastillo:
Do you happen to know why morphology tests are failing?

coderforlife:
No idea. Do they use the filter functions?

coderforlife:
`_normalize_sequence` was renamed to `_fix_sequence_arg` - I had no idea someone else was using these "protected" functions... just needs a third argument now for the name of the argument (used in the exception).

`_min_or_max_filter` doesn't have the `structure` argument anymore, I will have to figure that one out (it seems to be specifically created for the morphology functions). Additionally, the final argument is not a bool anymore but one of two strings ('min' or 'max').

So 2 easy things to fix and 1 thing I will have to look into, could be medium or hard.

emcastillo:
If you want you can restore the old functions and someone can fix them later in a follow-up PR

coderforlife:
I have fixed all of the issues. One other issue that didn't come up in the tests is that their tests used `pytest.skip()` which caused issues when running the tests with `unittest` which is recommended in the "contribution guide". So I switched it to `raise unittest.SkipTest(...)`.

emcastillo:
Awesome work and thanks once again for patiently addressing our comments!🙇‍♂️

emcastillo:
Jenkins, test this please

coderforlife:
It's okay - I already have the next PR planned and it will hopefully go smoother. It is added all of the other basic filters on top of this core set (`uniform_filter1d`, `uniform_filter`, `gaussian_filter1d`, `gaussian_filter`, `prewitt`, `sobel`, `generic_laplace`, `laplace`, `gaussian_laplace`, `generic_gradient_magnitude`, `gaussian_gradient_magnitude`, `rank_filter`, `median_filter`, and `percentile_filter`). I have them mostly written (just need to update for recent changes this core code). However, no tests yet for them.

I am thinking to do them in two batches: non-linear filters (the last 3 in that list) since they actually have a bit of new cuda code with them, then do all of those linear filters (which all just call the `convolve` or `convolve1d` functions).

I have also started working on `generic_filter` but that is going to be a bit different since it has to parse another reduction kernel to build it. I have it working with a few simple cases of reduction kernels and fused functions, but it needs a bit more work. Of course, it can't quite match the scipy function (which takes any Python function and LowLevelCallbacks which wrap Cython or C functions), but I think taking fuseable/fused Python functions and ReductionKernels would match as close as possible.

emcastillo:
wow, thats awesome!

This PR will be merged after the release today, right now code is frozen.

jakirkham:
This is great! Thanks everyone for pushing this forward 😄

PR #3189, opened 2020-03-11
Add `array_equal`

### LINKED ISSUE #3125: implementing array_equal() for CuPy
I was trying to implement array_equal() for Cupy and was thinking along these lines

```py
def array_equal(a1, a2):
    try:
        a1, a2 = cupy.asarray(a1), cupy.asarray(a2)
    except Exception:
        return False
    if a1.shape != a2.shape:
        return False
    return bool(cupy.asarray(a1 == a2).all())
```

Should I move forward with this ? I would be grateful if someone could guide me.

rushabh-v:
closes #3125

### DIFF
--- a/cupy/logic/comparison.py
+++ b/cupy/logic/comparison.py
@@ -38,6 +38,28 @@ _is_close_complex = core.create_ufunc(
 )
 
 
+def array_equal(a, b):
+    """Returns True if two arrays are element-wise exactly equal.
+
+    Args:
+        a (cupy.ndarray): Input array to compare.
+        b (cupy.ndarray): Input array to compare.
+
+    Returns:
+        cupy.ndarray: A boolean 0-dim array.
+            If its value is ``True``, two arrays are element-wise equal.
+
+    .. seealso:: :func:`numpy.array_equal`
+
+    """
+    a = cupy.asanyarray(a)
+    b = cupy.asanyarray(b)
+
+    if a.shape != b.shape:
+        return cupy.array(False)
+    return bool((a==b).all())
+
+
 def allclose(a, b, rtol=1.e-5, atol=1.e-8, equal_nan=False):
     """Returns True if two arrays are element-wise equal within a tolerance.
 
--- a/cupy/logic/comparison.py
+++ b/cupy/logic/comparison.py
@@ -38,12 +38,12 @@ _is_close_complex = core.create_ufunc(
 )
 
 
-def array_equal(a, b):
-    """Returns True if two arrays are element-wise exactly equal.
+def array_equal(a1, a2):
+    """Returns ``True`` if two arrays are element-wise exactly equal.
 
     Args:
-        a (cupy.ndarray): Input array to compare.
-        b (cupy.ndarray): Input array to compare.
+        a1 (cupy.ndarray): Input array to compare.
+        a2 (cupy.ndarray): Input array to compare.
 
     Returns:
         cupy.ndarray: A boolean 0-dim array.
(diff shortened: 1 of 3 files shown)


cupy/logic/comparison.py
```diff
@@ -38,6 +38,28 @@
 )
 
 
+def array_equal(a, b):
```
asi1024: For compatibility to `numpy.array_equal`
```suggestion
def array_equal(a1, a2):
```

cupy/logic/comparison.py
```diff
@@ -38,6 +38,28 @@
...
+    """Returns True if two arrays are element-wise exactly equal.
```
asi1024:
```suggestion
    """Returns ``True`` if two arrays are element-wise exactly equal.
```

cupy/logic/comparison.py
```diff
@@ -38,6 +38,28 @@
...
+
+    Returns:
+        cupy.ndarray: A boolean 0-dim array.
+            If its value is ``True``, two arrays are element-wise equal.
+
+    .. seealso:: :func:`numpy.array_equal`
+
+    """
+    a = cupy.asanyarray(a)
+    b = cupy.asanyarray(b)
```
asi1024: The inputs are assumed to be `cupy.ndarray`, so this typecast is not needed.

cupy/logic/comparison.py
```diff
@@ -38,6 +38,28 @@
...
+
+    if a.shape != b.shape:
+        return cupy.array(False)
+    return bool((a == b).all())
```
asi1024: The return value should be 0-dim ndarray.
```suggestion
    return (a == b).all()
```

tests/cupy_tests/logic_tests/test_comparison.py
```diff
@@ -74,6 +74,27 @@ def test_binary_array_pyscalar(self, xp, dtype):
...
+        a = xp.array([1, 2, 3, 4])
+        b = xp.array([1, 2, 3, 4])
+        return xp.allclose(a, b)
+
+    @testing.numpy_cupy_equal()
+    def test_array_equal_diff_length(self, xp):
+        a = xp.array([1, 2, 3, 4])
+        b = xp.array([1, 2, 3])
+        return xp.allclose(a, b)
+
```
asi1024: Please add tests check
- broadcasts are not allowed
- comparison with ndarrays of different dtypes is allowed

leofang: Why is allclose called instead of array_equal?

rushabh-v: Oh, I actually copy pasted the tests from `allclose` for my convenience and mistakenly left that call unchanged. I think I should've checked it..

I'll edit that soon.

cupy/logic/comparison.py
rushabh-v: Actually `(a == b).all()` returns `array(True)` or `array(False)`. And `numpy.array_equal` returns only `bool`s. So I thought they would not be comparable. 

Should I change that?

rushabh-v: I'll make other changes soon :)

leofang: Yes, you need to do this to follow CuPy's convention. This incompatibility is documented here: https://docs-cupy.chainer.org/en/stable/reference/difference.html#zero-dimensional-array

leofang: (Note that `all()` *is* a kind of reduction operation.)

rushabh-v: Thanks for providing with the documentation. 
I think I should read more and more developer guidelines before GSoC period for better understanding.

rushabh-v: I was a bit busier today.
Will make all the changes soon.

rushabh-v:
have made all the changes. And all the tests are passing.
Can you take a look, please? @asi1024

asi1024:
@rushabh-v Could you check that broadcasts are not allowed?

rushabh-v:
Added that test. Can you review it again, please? @asi1024

rushabh-v:
I've just checked whether `Numpy` and `CuPy` results same or not for the broadcast case, because `numpy.array_equal` also doesn't allow broadcasts.

asi1024:
Jenkins, test this please.

asi1024:
LGTM. Thanks for your PR!

PR #3190, opened 2020-03-12
Implement `gcd` and `lcm`

niteya-shah:
This PR adds the functions gcd and lcm for cupy.

~Currently lcm is using gcd in the kernel itself. I want to test more to see if this is necessary or calling it externally provides the same performance.~
Testing showed that the performance is the same so the implementation is sufficient

### DIFF
--- a/cupy/math/rational.py
+++ b/cupy/math/rational.py
@@ -1,5 +1,14 @@
 from cupy import core
 
+
+def _negative_gcd_error():
+    raise TypeError("gcd cannot be computed with boolean arrays")
+
+
+def _negative_lcm_error():
+    raise TypeError("lcm cannot be computed with boolean arrays")
+
+
 _gcd_preamble = '''
 template <typename T> inline __device__ T gcd(T in0, T in1) {
   T r;
@@ -16,7 +25,8 @@ template <typename T> inline __device__ T gcd(T in0, T in1) {
 
 gcd = core.create_ufunc(
     'cupy_gcd',
-    ('bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
+    (('??->?', _negative_gcd_error),
+     'bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
      'LL->L', 'qq->q', 'QQ->Q'),
     'out0 = gcd(in0, in1)',
     preamble=_gcd_preamble,
@@ -40,7 +50,8 @@ template <typename T> inline __device__ T lcm(T in0, T in1) {
 
 lcm = core.create_ufunc(
     'cupy_lcm',
-    ('bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
+    (('??->?', _negative_lcm_error),
+     'bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
      'LL->L', 'qq->q', 'QQ->Q'),
     'out0 = lcm(in0, in1)',
     preamble=_lcm_preamble,
--- a/cupy/math/rational.py
+++ b/cupy/math/rational.py
@@ -18,7 +18,7 @@ inline __device__ T gcd(T in0, T in1) {
 
 gcd = core.create_ufunc(
     'cupy_gcd',
-    ('??->?', 'bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
+    ('bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
      'LL->L', 'qq->q', 'QQ->Q'),
     'out0 = gcd(in0, in1)',
     preamble=_gcd_preamble,
@@ -49,13 +49,13 @@ inline __device__ T lcm(T in0, T in1) {
         if(r == 0)
             return 0;
         else
-            return (in0 * in1)/r;
+            return in0/r * in1;
     }
 '''
 
 lcm = core.create_ufunc(
     'cupy_lcm',
-    ('??->?', 'bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
+    ('bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
      'LL->L', 'qq->q', 'QQ->Q'),
     'out0 = lcm(in0, in1)',
     preamble=_lcm_preamble,
--- a/cupy/math/rational.py
+++ b/cupy/math/rational.py
@@ -1,0 +1,66 @@
+from cupy import core
+
+_gcd_preamble = '''
+template <typename T>
+inline __device__ T gcd(T in0, T in1) {
+    T r;
+    while(in1!=0){
+      r = in0 % in1;
+      in0 = in1;
+      in1 = r;
+    }
+    if(in0 < 0)
+        return - in0;
+    else
+        return in0;
+    }
+'''
+
+_gcd = core.create_ufunc(
+    'cupy_gcd',
+    ('??->?', 'bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
+     'LL->L', 'qq->q', 'QQ->Q'),
+    'out0 = gcd(in0, in1)',
+    preamble=_gcd_preamble,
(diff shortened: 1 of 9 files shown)


tests/cupy_tests/math_tests/test_rational.py
```diff
@@ -0,0 +1,38 @@
...
+class TestRational(unittest.TestCase):
+
+    @testing.for_int_dtypes(no_bool=True)
+    @testing.numpy_cupy_array_equal()
+    def test_gcd_basic(self, xp, dtype):
+        a = testing.shaped_random((6, 6), xp, dtype)
+        b = testing.shaped_random((6, 6), xp, dtype)
+        return xp.gcd(a, b)
+
+    @testing.for_float_dtypes()
```
asi1024: Add tests for Boolean and complex dtypes.
```suggestion
    @testing.for_dtypes(['?', 'e', 'f', 'd', 'F', 'D'])
```

tests/cupy_tests/math_tests/test_rational.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@testing.gpu
+class TestRational(unittest.TestCase):
+
+    @testing.for_int_dtypes(no_bool=True)
+    @testing.numpy_cupy_array_equal()
+    def test_gcd_basic(self, xp, dtype):
+        a = testing.shaped_random((6, 6), xp, dtype)
+        b = testing.shaped_random((6, 6), xp, dtype)
```
asi1024: Feed different seeds to `a` and `b`.
```py
a = testing.shaped_random((6, 6), xp, dtype, seed=0)
b = testing.shaped_random((6, 6), xp, dtype, seed=1)
```

cupy/math/rational.py
```diff
@@ -0,0 +1,66 @@
...
+        return in0;
+    }
+
+template <typename T>
+inline __device__ T lcm(T in0, T in1) {
+        T r = gcd(in0, in1);
+        if(r == 0)
+            return 0;
+        else
+            return (in0 * in1)/r;
```
asi1024: Avoid integer overflow for large inputs.
```suggestion
            return in0 / r * in1;
```

niteya-shah: Done

tests/cupy_tests/math_tests/test_rational.py
niteya-shah: Done

tests/cupy_tests/math_tests/test_rational.py
niteya-shah: Done

niteya-shah:
While implementing this I came across some strange behaviour,
```
simple_ufunc = cupy.core.create_ufunc(
    'simple_ufunc',
    ('b->b', 'h->h', 'i->i', 'l->l'),
    'out0 = in0')

a = cupy.array([True, False, False])
b = cupy.array([False, False, True])
simple_ufunc(b)
```
The above works even though we never specify bool in the accepted types list, instead it permits it by typecasting it to the first type in the list(if we had written, h->h before b->b it typecasts to int16 instead of int8). Is this a bug or expected behaviour?

asi1024:
This is the expected behavior.
`create_ufunc` is undocumented, but for example `cupy.subtract` raises `TypeError` for bool inputs. Please check its implementation.

niteya-shah:
The performance benchmark for gcd and lcm are as follows
gcd

> size = 100000
> numpy
> gcd                 : 13360.778 us   +/-687.684 (min:12873.837 / max:16644.847) us
> cupy
> gcd                 :    11.630 us   +/- 0.938 (min:   11.008 / max:   17.261) us    348.258 us   +/- 2.270 (min:  344.192 / max:  356.928) us
> 

> size = 1000000
> numpy
> gcd                 : 131831.420 us   +/-1681.092 (min:129864.742 / max:137775.571) us
> cupy
> gcd                 :    23.921 us   +/-22.992 (min:   12.101 / max:  148.400) us   3245.719 us   +/-20.154 (min: 3233.440 / max: 3344.928) us

> size = 10000000
> numpy
> gcd                 : 1327643.797 us   +/-16034.114 (min:1319980.853 / max:1446820.613) us
> cupy
> gcd                 :    22.656 us   +/-16.292 (min:   13.221 / max:   73.323) us  32120.730 us   +/-18.294 (min:32102.047 / max:32217.663) us

and for lcm

> size = 100000
> numpy
> lcm                 : 14097.452 us   +/-669.122 (min:13646.137 / max:17272.843) us
> cupy
> lcm                 :    11.926 us   +/- 3.329 (min:   10.634 / max:   32.183) us    366.765 us   +/- 3.625 (min:  362.912 / max:  386.304) us

> size = 1000000
> numpy
> lcm                 : 140687.834 us   +/-4234.880 (min:137613.297 / max:154011.181) us
> cupy
> lcm                 :    17.377 us   +/-13.136 (min:   12.605 / max:   88.989) us   3433.353 us   +/-12.278 (min: 3425.824 / max: 3500.032) us

> size = 10000000
> numpy
> lcm                 : 1395407.041 us   +/-11416.222 (min:1389399.977 / max:1476812.210) us
> cupy
> lcm                 :    21.779 us   +/-10.210 (min:   13.778 / max:   71.256) us  34041.828 us   +/-10.758 (min:34025.921 / max:34093.697) us

This averages at a 40 times speed up for both lcm and gcd compared to numpy

niteya-shah:
@asi1024 can you please review ?

tests/cupy_tests/math_tests/test_rational.py
asi1024: Add a test for bool dtype.

tests/cupy_tests/math_tests/test_rational.py
```diff
@@ -0,0 +1,38 @@
...
+        b = cupy.random.randint(-10, 10, size=(10, 10)).astype(dtype)
+        with pytest.raises(TypeError):
+            cupy.gcd(a, b)
+
+    @testing.for_int_dtypes(no_bool=True)
+    @testing.numpy_cupy_array_equal()
+    def test_lcm_basic(self, xp, dtype):
+        a = testing.shaped_random((6, 6), xp, dtype, seed=2)
+        b = testing.shaped_random((6, 6), xp, dtype, seed=3)
+        return xp.lcm(a, b)
```
asi1024: `lcm(0, 0)` is not tested in this seed. Why not avoid using `shaped_random` and write a test case explicitly?.

tests/cupy_tests/math_tests/test_rational.py
niteya-shah: The test for bool fails as mentioned below, even when I specifically add 'no' casting, this fails.
example
```
import numpy
import cupy

for xp in [cupy, numpy]:
    print(xp.__name__)
    a = xp.array([True, False, False])
    b = xp.array([True, False, True])
    print(xp.logaddexp(a, b, casting='no'))
```

> cupy
> [1.693  0.6934 1.313 ]
> numpy
> Traceback (most recent call last):
>   File "/D/work/Python/trial2.py", line 9, in <module>
>     print(xp.logaddexp(a, b, casting='no'))
> TypeError: ufunc 'logaddexp' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''no''

Can you please tell me how else I should add this test because the proposed way typecasts it to int8 and the doesnt give the desired result.

tests/cupy_tests/math_tests/test_rational.py
niteya-shah: I have added additional test cases that cover boundary cases -> (0,0), negative and positive, and negative and negative

tests/cupy_tests/math_tests/test_rational.py
asi1024: Please refer this line: https://github.com/cupy/cupy/blob/master/cupy/core/_routines_math.pyx#L744

tests/cupy_tests/math_tests/test_rational.py
asi1024: What is the role of `test_lcm_basic`? I think it is better to write all the test cases explicitly.

tests/cupy_tests/math_tests/test_rational.py
niteya-shah: Thanks a lot, I have modified the ufunc and added the test case.

tests/cupy_tests/math_tests/test_rational.py
niteya-shah: I have added the test cases with all values explicitly written and have tried to choose values that represent all the boundary cases.

asi1024:
Jenkins, test this please.

asi1024:

PR #3191, opened 2020-03-13
Add check of `c_contiguous` when indexing `CArray`

emcastillo:
Originated at #3179 

When using raw arrays in a Elementwise kernel, their dimensions are not reduced even though they are contiguous. This is because the user may expect to address them using their real shape.
However, if these arrays are c_contiguous they can be indexed linearly, avoiding several expensive operations in 64-bit to transform a single integer index to a `ndim` `Indexer`.

Results on convolve function for x_size = (4000, 4000) and w_size of (3, 3) in a P100 are:

Master before #3179
```
convolve            :    6640.630 us
```
This PR
Before template
```
convolve            :    2222.367 us
```
After template
```
convolve            :    1570.488
```

Master after #3179
```
convolve            :    1571.202 us
```
#3179 addresses this issue directly in the `convolve` function. So results are expected to be better, as there is still some overhead on the indexing of this PR (A function call and a comparison). But this PR is applied generally so other routines can see performance benefits from it. If we templatize the parameter the performance gains are exactly in the range of #3179.
But this will increase the number of compiled kernels as a drawback.

Thanks to @anaruse and @coderforlife for their ideas.

### DIFF
--- a/cupy/core/_carray.pxd
+++ b/cupy/core/_carray.pxd
@@ -10,6 +10,7 @@ DEF MAX_NDIM = 25
 cdef struct _CArray:
     void* data
     Py_ssize_t size
+    Py_ssize_t c_contiguous
     Py_ssize_t shape_and_strides[MAX_NDIM * 2]
 
 
@@ -22,7 +23,8 @@ cdef class CArray(function.CPointer):
     cdef void init(
         self, void* data_ptr, Py_ssize_t data_size,
         const vector.vector[Py_ssize_t]& shape,
-        const vector.vector[Py_ssize_t]& strides)
+        const vector.vector[Py_ssize_t]& strides,
+        Py_ssize_t c_contiguous)
 
 
 cdef struct _CIndexer:
--- a/cupy/core/_carray.pxd
+++ b/cupy/core/_carray.pxd
@@ -10,7 +10,6 @@ DEF MAX_NDIM = 25
 cdef struct _CArray:
     void* data
     Py_ssize_t size
-    Py_ssize_t c_contiguous
     Py_ssize_t shape_and_strides[MAX_NDIM * 2]
 
 
@@ -23,8 +22,7 @@ cdef class CArray(function.CPointer):
     cdef void init(
         self, void* data_ptr, Py_ssize_t data_size,
         const vector.vector[Py_ssize_t]& shape,
-        const vector.vector[Py_ssize_t]& strides,
-        Py_ssize_t c_contiguous)
+        const vector.vector[Py_ssize_t]& strides)
 
 
 cdef struct _CIndexer:
--- a/cupy/core/_routines_sorting.pyx
+++ b/cupy/core/_routines_sorting.pyx
@@ -266,7 +266,7 @@ def _partition_kernel(dtype):
     dtype = _get_typename(dtype)
     source = string.Template('''
     template<typename T>
-    __device__ void bitonic_sort_step(CArray<T, 1> a,
+    __device__ void bitonic_sort_step(CArray<T, 1, true> a,
             ptrdiff_t x, ptrdiff_t y, int i, ptrdiff_t s, ptrdiff_t w) {
         for (ptrdiff_t j = i; j < (y - x) / 2; j += 32) {
             ptrdiff_t n = j + (j & -w);
@@ -281,7 +281,7 @@ def _partition_kernel(dtype):
     // Sort a[x:y].
     template<typename T>
     __device__ void bitonic_sort(
-            CArray<T, 1> a, ptrdiff_t x, ptrdiff_t y, int i) {
+            CArray<T, 1, true> a, ptrdiff_t x, ptrdiff_t y, int i) {
         for (ptrdiff_t s = 2; s <= y - x; s *= 2) {
             for (ptrdiff_t w = s / 2; w >= 1; w /= 2) {
                 bitonic_sort_step<T>(a, x, y, i, s, w);
@@ -292,7 +292,8 @@ def _partition_kernel(dtype):
     // Merge first k elements and the next 32 times t elements.
     template<typename T>
     __device__ void merge(
-            CArray<T, 1> a, int k, int i, ptrdiff_t x, ptrdiff_t z, int u) {
+            CArray<T, 1, true> a,
+            int k, int i, ptrdiff_t x, ptrdiff_t z, int u) {
         for (int s = i; s < u; s += 32) {
             if (a[x + k - s - 1] > a[z + s]) {
                 T tmp = a[x + k - s - 1];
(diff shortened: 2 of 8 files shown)


cupy/core/_carray.pxd
```diff
@@ -1,4 +1,5 @@
 cimport cython  # NOQA
+from libcpp cimport bool
```
leofang: Interesting...Didn't know `libcpp.bool`🙂 Usually we just use the Cython built-in `bint`. While I like `bool` better, I wonder if it'd conflict with Python's `bool`, which is still respected in Cython? Just curious.

emcastillo: I am thinking to actually change it to `bint` or rename the import with `as cbool`

cupy/core/include/cupy/carray.cuh
```diff
@@ -179,6 +179,7 @@ public:
   static const int ndim = _ndim;
 private:
   T* data_;
+  bool c_contiguous;
```
leofang: I wonder if data alignment in `CArray` matters or not? Could you try moving  this flag to the end of private data members and see if timing changes?

emcastillo: let me give it a try again but I had problems with it before

emcastillo: You're the boss dude.
 down to `1747.367`

emcastillo: changed types to enforce alignment, if defined after the shape & strides array probably the value is not written to the same memory location due to some memory layout differences between nvrtc & gcc. I am not 100% sure of this but I can't find another explanation.

cupy/core/_carray.pxd
```diff
@@ -10,6 +10,7 @@ DEF MAX_NDIM = 25
 cdef struct _CArray:
     void* data
     Py_ssize_t size
+    Py_ssize_t c_contiguous
```
emcastillo: using ssize_t to enforce correct alignment. 
Maybe add a comment?

cupy/core/include/cupy/carray.cuh
```diff
@@ -180,6 +180,7 @@ public:
 private:
   T* data_;
   ptrdiff_t size_;
+  ptrdiff_t c_contiguous;
```
asi1024: How about passing `c_contiguous` as a template parameter?

emcastillo: Great idea!!

leofang: Oh, I happen to be looking into the internals, so while waiting for thoughts from @emcastillo, I can share my objection: If `c_contiguous` is a template parameter instead of a class member, you'd be re-compiling kernels for C-contiguous inputs, reducing the chances of reusing existing kernels. This is a pure waste, since all the kernels do not have any specialization for such cases.

leofang: oops...we replied in sync 😛

emcastillo: I don't think it will create that much additional kernels. 
I guess I can try to run the test-suite and measure the number of created kernels in both cases. something like an extra 15~20% should be admissible I guess

leofang: Well while this increase not a big deal, I wanna say most of our test cases only use C-contiguous inputs, so it could be possible to have <10% increase 😂

emcastillo: Kernels with several parameters can create a combinatorial explosion, so we can end up with one or two orders of magnitude more kernels than the original numbers. I see your point.

emcastillo: I tried the template, the speedup is noticeable
```
1570.192
```
Exactly the same as using #3179

asi1024: Yes, the condition in [L219](https://github.com/cupy/cupy/pull/3191/files#diff-8086e46284b5ae516b8e5a426470dac6R219) can be decided at compile-time, so it is good for compile-time optimization.

emcastillo: Do you think that the total increase in the number of compiled kernels can be a burden?

asi1024: Currently, `CArray` takes `ndim` as a template parameter, but I feel it is not a burden. So I guess that `c_contiguous` will not be so too.

emcastillo: pushed the changes, PTAL

cupy/core/include/cupy/carray.cuh
leofang: I have no idea why we have the alignment issue here...Looks like the struct should be 8 byte aligned (due to `ptrdiff_t`)? Sorry I didn't help 😅

cupy/core/include/cupy/carray.cuh
leofang: Thanks for the checks, Emilio!

leofang:
One more stupid question: I suppose the name "CArray" implies that F-contiguous arrays are discriminated?

emcastillo:
yay, no one likes fortran arrays

asi1024:
Jenkins, test this please.

emcastillo:
After fixing the issues, there are 2 tests that keep failing.
However, the failure is a bit tricky.
```
tests.cupy_tests.core_tests.test_ndarray_adv_indexing.TestArrayAdvancedIndexingGetitemPerm_param_43_{indexes=([1, 2], [1, 0], slice(0, 3, 2)), shape=(4, 4, 4)}::test_adv_getitem
tests.cupy_tests.core_tests.test_ndarray_adv_indexing.TestArrayAdvancedIndexingGetitemPerm_param_45_{indexes=([1, 0], [1, 2], slice(0, 3, 2)), shape=(4, 4, 4)}::test_adv_getitem
```

If we only run these two tests, they never fail, but if they are run with other tests, they do.
If we change the parameters order and make this test to execute on the first place, then everything works. There seems to be some kind of issue in the advanced indexing kernels maybe.

cupy/core/_routines_sorting.pyx
```diff
@@ -266,7 +266,7 @@ def _partition_kernel(dtype):
     dtype = _get_typename(dtype)
     source = string.Template('''
     template<typename T>
-    __device__ void bitonic_sort_step(CArray<T, 1> a,
+    __device__ void bitonic_sort_step(CArray<T, 1, true> a,
```
asi1024: Is the input array always contiguous?

emcastillo: Yes, sorry for not adding a comment.
Input is raveled here

https://github.com/cupy/cupy/blob/99e928685f15eecba38a1b1aac51c94d330cb6a0/cupy/core/_routines_sorting.pyx#L185

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

Issue #3144, opened 2020-03-03
Configurable kernel_cache directory


jakirkham:
It would be nice to make the `kernel_cache` directory location configurable. This is handy in situations where the user's home directory might live on NFS (very slow for IO), when a container is used (might lack permissions for the home directory), etc. In these cases the user could set another location with an environment variable.

leofang:
Doesn’t `CUPY_CACHE_DIR` work?
https://docs-cupy.chainer.org/en/stable/reference/environment.html

jakirkham:
Oops 😬 Only a failure to read it seem 😞 Thanks Leo! 😄

leofang:
No problem, John! If it doesn’t work we’ll need to fix it. We also need to set this in our HPC environments.

jakirkham:
As you mention setting this on HPC, I'm curious what you choose for this. Do you use node local scratch or something visible to multiple nodes (like an NFS mounted directory)?

leofang:
For us it depends on the MPI pool size. If it's not too large, then my preference is NFS to maximize reusability (and a bit easier to set up). If using many nodes, then local scratch may be necessary (NFS is super slow and one of ours is particularly unstable 😞).

But, this is just a minor concern to me, as our bottleneck is usually something else (like `@numba.cuda.jit` or HDF5 I/O). So usually we forget about this issue 😅 

Another thing I occasionally do is to clean up the `kernel_cache` directory. I hope for faster lookup speed, but I never seriously measure it. It's more like a personal hygiene habit 😛

jakirkham:
Thanks for the info! 😄 Yeah NFS is tricky. 😞

I guess Numba has [caching too]( http://numba.pydata.org/numba-doc/latest/user/jit.html#cache ). Have you experimented with that at all? 😉

What are the slow parts of HDF5 you are running into? 🙂

Yeah that's an interesting point. I guess it is somewhat a function of the underlying filesystem.

Suppose if we got really interested in this, we could imagine using a sharded key-value database for faster retrieval. This is probably not hard to do. Just not sure of the performance benefits. 🤔

leofang:
@jakirkham:

> I guess Numba has caching too. Have you experimented with that at all? 😉

Not for `@numba.cuda.jit`. For the CPU jit (`@numba.jit`) it's pretty sophisticated and mature now: you can do either caching or ahead-of-time (AOT) compiling (via `pycc`). None of these exists for CUDA kernels so far... 

[...]

PR #3200, opened 2020-03-17
Make `RawKernel` and `RawModule` aware of CUDA context

### LINKED ISSUE #3193: Excessive overhead for CuPy raw kernel launch.
CuPy Version          : 7.2.0
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSPARSE Version      : 10301
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2406
NCCL Runtime Version  : 2507

numba_vs_cupy

I have kernels, one Numba and one CuPy raw kernel. They are nearly identical and the CuPy raw kernel is 2x faster. But the overhead of launching _cuModuleLoadData_ and _cuModuleUnload_ make the overall execution of the CuPy function 2x **slower** than Numba. I've attached a screenshot from Nsight Systems. 

Also, code can be found https://github.com/rapidsai/cusignal/pull/31

leofang:
Close #3193.

In #3193, I found the case from @mnicely hit this subtle **bug**: If a `RawKernel`/`RawModule` is compiled for one device, it cannot be reused on another device. This PR fixes that by detecting current device and recompiling when necessary. (See the added tests for how to reproduce the bug --- the current master branch, or v7, would fail in these tests.)

Usually, this issue is circumvented by using the `cupy.util.memoize(for_each_device=True)` decorator. However, because we can't use that for cdef classes afaik, the internal caching with device ID being part of the key is necessary. 

I also took this chance to enforce some typing that we didn't do before moving to PY3-only. If it's considered unnecessary, let me know and I'll revert. I don't think this would cause any backward incompatibility though.

May be in conflict with #2966? cc: @emcastillo

### DIFF
--- a/cupy/core/raw.pxd
+++ b/cupy/core/raw.pxd
@@ -4,10 +4,10 @@ cdef class RawKernel:
         readonly str code
         readonly str name
         readonly tuple options
-        object _kernel
         readonly str backend
-        bint translate_cucomplex
         readonly bint enable_cooperative_groups
+        bint translate_cucomplex
+        dict kernels
 
 
 cdef class RawModule:
@@ -16,8 +16,11 @@ cdef class RawModule:
         readonly str code
         readonly str cubin_path
         readonly tuple options
-        dict kernels
         readonly str backend
-        object module
-        bint translate_cucomplex
         readonly bint enable_cooperative_groups
+        bint translate_cucomplex
+        dict kernels
+        dict modules
+
+        _load_from_path(self, str)
+        _compile_from_source(self, str, tuple, str, bint, bint)
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -1,7 +1,8 @@
 import cupy
 from cupy import util
 from cupy.cuda cimport driver
-from cupy.cuda.function cimport Module
+from cupy.cuda cimport runtime
+from cupy.cuda.function cimport Function, Module
 
 
 cdef class RawKernel:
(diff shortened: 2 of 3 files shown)


leofang:
I am wondering it we could just reload the existing cubin to another context and avoid recompiling...

leofang:
Sorry for thinking out loud 😅

> I am wondering it we could just reload the existing cubin to another context and avoid recompiling...

Ah, I suppose this is what's done internally in `cupy/cuda/compiler.py`, since the code is the same. The only price to pay for switching context is one call to `cuModuleLoad` for reloading.

leofang:
@asi1024 I think this bug is related our discussion starting https://github.com/cupy/cupy/issues/3144#issuecomment-597464683.

emcastillo:
I discussed offline with @asi1024 and we believe it is better to get the current device in the constructor and raise an exception if the kernel is called from other device.

leofang:
OK fine with me. Whatever behavior should be clearly documented.

Please feel free to close this PR when the accepted solution is merged.

emcastillo:
Also, another alternative is to have a RawKernel and RawModule factory method in python where you can use the `cupy.util.memoize(for_each_device=True)` decorator

leofang:
Just for the record: the memoization approach taken in this PR is inspired by that of ufuncs', see https://github.com/cupy/cupy/blob/766dfd04703da1d9a971d1124951c24b91bfbc33/cupy/core/_kernel.pyx#L1060-L1072

PR #3201, opened 2020-03-18
Make `RawKernel` and `RawModule` aware of CUDA context (alt)

### LINKED ISSUE #3193: Excessive overhead for CuPy raw kernel launch.

leofang:
Close #3193. Close #3200. 

For motivation see the initial PR #3200. This is an alternative attempt (didn't want to force push there).

After offline discussion with @emcastillo, I am convinced to try and utilize the internal per-device cache (`cupy.util.memoize`), so we don't have to do the cache by ourselves in this module. This should greatly simplifies the code.

This PR has the same expected behavior as the previous one: The same `RawKernel`/`RawModule` instance would be usable on any device. Recompiling (and caching) is done internally when necessary.

No backward compatibility is broken in either PR: Note that if the context is not switched, for `RawKernel`, the actual compilation happens at first launch, whereas for `RawModule`, the compilation happens when creating the instance. Both PRs only addresses the context switching bug.

Tests are simply taken from #3200.

### DIFF
--- a/tests/cupy_tests/core_tests/test_raw.py
+++ b/tests/cupy_tests/core_tests/test_raw.py
@@ -11,9 +11,10 @@ from cupy.cuda import compiler
 
 _test_source1 = r'''
 extern "C" __global__
-void test_sum(const float* x1, const float* x2, float* y) {
+void test_sum(const float* x1, const float* x2, float* y, unsigned int N) {
     int tid = blockDim.x * blockIdx.x + threadIdx.x;
-    y[tid] = x1[tid] + x2[tid];
+    if (tid < N)
+        y[tid] = x1[tid] + x2[tid];
 }
 '''
 
@@ -103,6 +104,16 @@ __global__ void test_kernel(float *arr, int N, int inner_blk)
         test_kernel_inner<<<1, inner_blk>>>(arr+tid*inner_blk, inner_blk);
 }
 
+}
+'''
+
+# to generate cubin/ptx
+_test_source5 = r'''
+extern "C" __global__
+void test_div(const float* x1, const float* x2, float* y, unsigned int N) {
+    int tid = blockDim.x * blockIdx.x + threadIdx.x;
+    if (tid < N)
+        y[tid] = x1[tid] / (x2[tid] + 1.0);
 }
 '''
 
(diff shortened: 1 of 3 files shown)


leofang:
I found in my prior PRs I didn't add any test for loading cubin/ptx...My bad 😞 This implementation would fail with loading. Let me add a few tests and then fix it.

leofang:
> I found in my prior PRs on RawModule I didn't add any test for loading cubin/ptx...My bad 😞 This implementation would fail with loading. Let me add a few tests and then fix it.

This is addressed in 2b659e6.

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -11,9 +11,10 @@
 
 _test_source1 = r'''
 extern "C" __global__
-void test_sum(const float* x1, const float* x2, float* y) {
+void test_sum(const float* x1, const float* x2, float* y, unsigned int N) {
     int tid = blockDim.x * blockIdx.x + threadIdx.x;
-    y[tid] = x1[tid] + x2[tid];
+    if (tid < N)
+        y[tid] = x1[tid] + x2[tid];
```
leofang: This fixes a bug silently ignored in by the driver API...

emcastillo:
Jenkins, test this please

leofang:
Thanks, Emilio!

emcastillo:
Yes,
probably just having a fixed size array with the number of devices should be ok and much faster

leofang:
> Yes,
> probably just having a fixed size array with the number of devices should be ok and much faster

@emcastillo Is this a reply to https://github.com/cupy/cupy/issues/3326#issuecomment-628188846? 😂

carterbox:
Has this fix been backported to CuPy 7.x?

leofang:
Hi Daniel, I don't think so. This PR was not considered a bug fix, so it was not backported. Is there a need?

carterbox:
I wouldn't say I have a need. I was just curious. I can wait for CuPy 8.

Issue #2635, opened 2019-11-13
Using RMM with CuPy


jakirkham:
As RMM is the memory manager used in the RAPIDS stack, it might be interesting to explore what would be needed to use RMM with CuPy. This would allow users working with RAPIDS and CuPy to ensure all memory allocations are managed by the same memory manager (RMM) and come from the same memory pool. Not only can this be useful at a low-level by consolidating memory management, but it can be useful at a high-level by tools tracking memory management (like dashboards, etc.).

On the flip side, it could be interesting to use CuPy's memory manager for allocating objects outside of CuPy. This could be useful for developers debugging different kinds of problems. It could also be helpful to users that are trying to see how their performance is impacted by different kinds of memory managers. Additionally this could be helpful in improving different memory managers by learning about the design choices each makes.

kmaehashi:
Basically it sounds feasible as CuPy supports replacing allocator (https://docs-cupy.chainer.org/en/latest/reference/memory.html#changing-memory-pool) and `rmm` provides Python interface for allocating memory.
It's nice if RMM (or cuDF?) provides CuPy memory pool support.

kkraus14:
@kmaehashi We're going to explore this when we get some free cycles. The initial thought is that when someone imports RMM we'll check if cupy is importable, and if it is we'll set the cupy allocator to use an RMM defined allocator.

Do you foresee any issues with this or situations where the allocator will get overridden?

kmaehashi:
As library imports are often implicit to end-users, I'd rather prefer an explicit API to override the memory pool, e.g., `rmm.override_cupy_memory_pool()`.
(e.g. if user imports a library that has strong dependency to cupy and weak (optional) dependency to RMM, users may want to use CuPy memory allocator.)
Automatic override in higher layer application like cuDF sounds fine for me.

jakirkham:
If we are trying to constrain the allocator used based on context, we may consider using the `set_allocator` context manager ( https://github.com/cupy/cupy/issues/2625 ).

kkraus14:
> As library imports are often implicit to end-users, I'd rather prefer an explicit API to override the memory pool, e.g., `rmm.override_cupy_memory_pool()`.

My experience with this is that many users won't look at the docs / find that API and then will have conflicting memory pools. In order to avoid this with using Numba we ended up re-exposing large portions of the user facing Numba API related to device array construction, and even then we still get plenty of users creating a combination of RMM backed and cudaMalloc backed device arrays.

[...]

PR #3212, opened 2020-03-20
Expose CuPy memory pool to C/C++ space

### LINKED ISSUE #2710: Expose CuPy memory pool to C/C++ space?
I found the following snippet (L15-L40) in `cupy.cuda.thrust`: 
https://github.com/cupy/cupy/blob/4586fedac8254dcd71d97617f2f95f98254dbbcc/cupy/cuda/thrust.pyx#L15-L40

I think this is an important low-level API for external C/C++/Cython code (such as Thrust and CUB) to use CuPy's memory pool. Is it possible to move them to, say, `cupy/cuda/memory.pxd` so that it can be reused?

leofang:
Close #2710. 

This PR does 5 things:
1. Move all memory-exposure functions from `cupy.cuda.thrust` to `cupy.cuda.memory` as it seems far more proper there. Only `cupy_allocator` in `cupy/cuda/cupy_thrust.cu` is not moved, as it's needed by Thrust for template specialization.
2. Due to 1, the `public` keyword is replaced by `api` so that any C/C++ functions not linked to `cupy.cuda.memory` can also use those functions. For more info see the Cython instruction https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#c-api-declarations.
3. (**UPDATED**) Provide public C API in `cupy_memory.h` for any C/C++ applications to access CuPy's memory pool:
```C
cupy_allocator_handle* get_cupy_allocator_handle();
void destroy_cupy_allocator_handle(cupy_allocator_handle* ptr);
void* cupy_malloc(cupy_allocator_handle* handle, size_t size);
void cupy_free(cupy_allocator_handle* handle, void* ptr);
```
A public API is necessary (design subject to discussion) in order to avoid the nuisance of Python init/de-init issues. Since this implementation is simple, it can be made header-only to avoid unnecessary linking to CuPy shared libraries, which could be annoying.

4. Simplify a bit the APIs in `cupy/cuda/cupy_thrust.cu` to reflect the change, as it doesn't make sense to pass a Python object to C/C++ just for bookkeeping. Also, the type of stream pointers is changed from `size_t` to `intptr_t` (cf #2455).
5. Add tests to ensure the C/C++ memory API works.

Question: 
- Should `cupy_memory.h` be placed in `cupy/cuda/` or `cupy/core/include/cupy/`? I have a feeling that I will need it in the latter...
- ~~Clean separation of C/C++ interface? (See https://github.com/cupy/cupy/pull/3212#issuecomment-601902102) But Cython code cannot be compiled by a pure C compiler, so half of this effort would be hard to test against...~~
- Do we wanna document this?


leofang:
I think I need some help. I want to make `cupy_memory.h` a header-only implementation, so that any 3rd party can just include it without worrying the linker problem. However, I am not sure how I can cleanly separate the C interface from its C++ counterpart, so that for C applications they just need a C compiler. Any advice is more than welcome!

leofang:
I removed the C++ interface (`cupy_device_allocator`) in b68fd78 to make the implementation cleaner. This seems sufficient.

leofang:
@jakirkham @kkraus14 Given your experience in RMM, I wonder if you could comment on the approach taken here, in particular the new header `cupy_memory.h`. 🙏

The gotcha here is that CuPy's memory pool is implemented in Cython, so to access it one has to initialize the Python interpreter (if hasn't done so already); but given the intent to interact with CuPy, I don't think this would be an issue.

cupy/cuda/memory.pyx
```diff
@@ -1429,3 +1429,25 @@ cdef class CFunctionAllocator:
...
+
+###############################################################################
+# Expose CuPy memory pool to C/C++ (migrated from cupy/cuda/thrust.pyx)       #
+###############################################################################
+
+# This is for keeping track of memory exposed to C/C++.
+# (Key, Value): (ptr (of type intptr_t), MemoryPointer)
+# Items are removed when the C/C++ function calls cupy_c_free
+cdef dict _ext_mem_ptrs = {}
+
```
emcastillo: Is this solely for debugging purposes?

leofang: No. It's needed. If nothing holds a ref to the `MemoryPointer`, it'd get freed right after being allocated in C. This is the original approach: https://github.com/cupy/cupy/blob/bccda26ee25f48d79a7f99b259aedea3388d3323/cupy/cuda/thrust.pyx#L19-L24
but it's too bulky and unnecessary...

emcastillo: Ah, you're right, sorry I just woke up 😅, please document it.

leofang: Yeah looking into how to get C functions documented in Sphinx...

emcastillo: Just a comment in the code should be ok!

leofang: I think I can make pretty-looking Sphinx doc for the C API, is this OK to you? See below.
螢幕快照 2020-03-23 上午1 01 06
(It does not contain the entire doc strings, just wanna quickly flash you what I mean.🙂)

kkraus14:
> @jakirkham @kkraus14 Given your experience in RMM, I wonder if you could comment on the approach taken here, in particular the new header `cupy_memory.h`. 🙏
> 
> The gotcha here is that CuPy's memory pool is implemented in Cython, so to access it one has to initialize the Python interpreter (if hasn't done so already); but given the intent to interact with CuPy, I don't think this would be an issue.

I would also call @harrism and @jrhemstad to attention as they have been the main contributors to RMM's C++ APIs and implementations. 😄

harrism:
One thing I will point out is that RMM's allocators are all derived from a `device_memory_resource` base class that follows the C++17 std::pmr::memory_resource (Polymorphic Memory Resource). https://en.cppreference.com/w/cpp/memory/memory_resource

The thing to notice there is that the `memory_resource::deallocate()` function takes not just a pointer, but the size. All RMM `deallocate` functions going forward do the same. If copy's `free` function only takes a pointer, it will be difficult to correctly interface with RMM.

jrhemstad:
> If copy's free function only takes a pointer, it will be difficult to correctly interface with RMM.

A resource is free to ignore the `size` parameter passed to `deallocate`, just like `cuda_memory_resource`.

docs/source/reference/cuda.rst
```diff
@@ -35,6 +35,46 @@ Memory management
...
+    This function destroys the handle ``ptr`` to CuPy's device memory allocator.
+    Any device memory allocated from the handle should be freed before destroying
+    it.
+
+    Internally, it checks whether ``Py_Initialize()`` was called when the handle
+    was created, and calls ``Py_Finalize()`` if so.
+
+.. c:function:: void* cupy_malloc(cupy_allocator_handle* handle, size_t size)
+
+    This function returns a device pointer to the memory of ``size`` bytes
```
jrhemstad: Alignment?

cupy/cuda/cupy_memory.h
```diff
@@ -0,0 +1,63 @@
...
+#include "memory_api.h"  // generated by Cython from cupy.cuda.memory
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+  *********************** C public API ************************
+*/
+typedef struct cupy_allocator_handle_t cupy_allocator_handle;
```
jrhemstad: Should the header be documented?

leofang:
@harrism @jrhemstad Thanks for quick feedbacks! Allow me to respond to a few common issues in one place. 

A number of constraints apply here, IIUC, due to CuPy's Cython implementation of its memory management:
1. Regardless of whether functions and classes written in Cython are `cdef`'d or not, the manipulation must be done with a Python interpreter alive. This is [Cython's limitation](https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#c-api-declarations) as it relies on Python's module import and ref count mechanisms. 

2. CuPy has a mechanism to switch the per-thread allocator in Python (which you probably knew already as it's used in RMM 🙂), and the current allocator is respected by the C API. 

So, in theory we can even do dumb things like: Ask CuPy to use RMM allocator -> Expose CuPy's pool by the C API here -> Have another Python C module (say cuDF?) to access the pool. The net effect is both cuDF and CuPy use RMM memory, but of course in practice we don't do it in this way...

To Mark's comment:
> All RMM deallocate functions do the same. If copy's free function only takes a pointer, it will be difficult to correctly interface with RMM.

Yeah I noticed that from Thrust too. But I have two questions for this extra argument:
1. What is the expected behavior if the passed-in `size` does not match the actual size pointed by the device pointer/used during allocation? This behavior seems to be unspecified in C++17?
2. What do you mean "to interface with RMM"? Doesn't RMM already do its own memory management? 

Here the exposed memory is backed by a Python object, so like Jake pointed out I can make the size accepted, but it'd just be ignored. Also, to match the signature of `std::pmr::memory_resource` you need another wrapper on top anyway, as the first argument here is `handle`.

To Jake's comment:
> The requirement of having the Python Interpreter initialized would make it pretty difficult to ever use this generically in RMM.

Not sure what you meant by "use it in RMM" (see above). But if you're building a Python C/C++ module that is eventually imported in Python, this would not be a big issue. As for pure C/C++ code not involving Python --- Why would you want to interface with CuPy at all...

> Since the logic is in Cython, would it be that difficult to move it to native C++?

Yes, it requires the entire `cupy.cuda.memory` module to be rewritten in C++ AFAIK unfortunately. 

> since cupy uses Thrust heavily, you may be interested to know that Thrust added some of it's own pool allocators

Oh, CuPy does not use Thrust heavily at all. Only Thrust's sorting functions are used (see the touched functions in this PR). One can even set `CUPY_SETUP_ENABLE_THRUST=0` to not build `cupy.cuda.thrust`, and CuPy can be run if the sorting functionalities are not needed. In fact, I'd like to look into if this module can be replaced by CUB at some point.

Anyway, thanks for the pointer Jake. It'd be interesting to initialize a Thrust pool, and then warp it using CuPy's `CFunctionAllocator` (undocumented, why?): https://github.com/cupy/cupy/blob/ff18a56f23058a947e5ceab74995deda8514b2f1/cupy/cuda/memory.pyx#L1393

docs/source/reference/cuda.rst
leofang: CuPy hard-codes it to 512 bytes IIUC: https://github.com/cupy/cupy/blob/ff18a56f23058a947e5ceab74995deda8514b2f1/cupy/cuda/memory.pyx#L679-L681

cupy/cuda/cupy_memory.h
leofang: Oh yes, see 5a4aadc. Pending @emcastillo's feedback on the style and appearance...

docs/source/reference/cuda.rst
jrhemstad: Note that `cudaMalloc` is aligned to (at least) 256B, not 512.

I think it should just be documented here as well.

jrhemstad:
> What is the expected behavior if the passed-in size does not match the actual size pointed by the device pointer/used during allocation? This behavior seems to be unspecified in C++17?

Yep, exactly. It's undefined behavior, i.e., bad, don't do it. 

`std::pmr::memory_resource::deallocate` requires the size be exactly the same as what was passed into `allocate`.

docs/source/reference/cuda.rst
leofang: Jake, thanks for the inside info! I don't think this is ever documented publicly, is it? I also spent some time trying to verify this a while ago but in vain, so I came up [a theory](https://gist.github.com/sonots/41daaa6432b1c8b27ef782cd14064269#gistcomment-3014146) for why this is the case...

jrhemstad: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses

> Any address of a variable residing in global memory or returned by one of the memory allocation routines from the driver or runtime API is always aligned to at least 256 bytes.

leofang: Jake, are you able to verify this comment? https://gist.github.com/sonots/41daaa6432b1c8b27ef782cd14064269#gistcomment-2127860

jrhemstad: The requirement is for _at least_ 256B. It is unsafe to rely on it being anything larger.

jakirkham:
(It sounds like you are in Mark and Jake's capable hands. Though please let us know if you need anything else 🙂)

harrism:
Please note @jrhemstad edited one of his important comments above.

Aside: it may be helpful to have an example of *why* C++ PMR's require the size be passed to `deallocate()`. The first time I saw it, I was surprised and couldn't see why it was required -- after all  C++ `delete` operator and C's `free()` do not take a size. 

The reason is because of polymorphism and composability, which is the power of PMRs. For example, I built a [`hybrid_memory_resource`](https://github.com/rapidsai/rmm/blob/branch-0.13/include/rmm/mr/device/hybrid_memory_resource.hpp) in RMM which simply has a size threshold, and pointers two upstream memory_resource objects. Any allocation / deallocation uses the specified size to select between the two upstream memory_resources. The same upstream resource must be used for both allocation and deallocation of the pointer, so the size must be known at both allocation and deallocation time. Now, the memory_resource could store a map of pointers to sizes, but that would result in duplication of that mapping in every memory_resource implementation, or require tracking it in the base class, which would be overkill for resources that don't need it (e.g. the ones that just pass through to malloc/cudaMalloc.)

leofang:
Thanks all for comments!!! Before I get back to you on these, I wanna digress to a different question before I forgot: Is it possible to explicitly pass in a pre-allocated work area and a stream pointer to Thrust routines (as in CUB)? This way we could avoid writing any custom memory allocator or execution policy (I know RMM did this).

jrhemstad:
> Thanks all for comments!!! Before I get back to you on these, I wanna digress to a different question before I forgot: Is it possible to explicitly pass in a pre-allocated work area and a stream pointer to Thrust routines (as in CUB)? This way we could avoid writing any custom memory allocator or execution policy (I know RMM did this).

Not like CUB, no. You could fake this by creating a Thrust allocator facade to some pre-allocated memory, but this is basically just a poor man's pool allocator.

cupy/cuda/cupy_thrust.cu
```diff
@@ -17,29 +18,31 @@ typedef hipStream_t cudaStream_t;
...
+    }
+
+    ~cupy_allocator() {
+        destroy_cupy_allocator_handle(pool);
+    }
 
-    char *allocate(std::ptrdiff_t num_bytes) {
-        return cupy_malloc(memory, num_bytes);
+    char* allocate(size_t num_bytes) {
+        return (char*)cupy_malloc(pool, num_bytes);
```
emcastillo: I think that calling directly `cupy_malloc` is not safe.
Probably we should add a new `cupy_malloc(pool, size, stream)` to allocate memory on a given stream, this should make it versatile.

leofang: Hi Emilio, "not safe" because of the lack of a stream? I thought that none of our allocation routines accept a stream pointer, and that it's acquired internally by looking up the thread-local storage?
https://github.com/cupy/cupy/blob/ff18a56f23058a947e5ceab74995deda8514b2f1/cupy/cuda/memory.pyx#L940
Note that the original code has no stream argument either.

In any case, it's not possible to declare `cupy_malloc(pool, size, stream)` and then wrap it with `allocate(size, stream)` on top, because the signature of `allocate` is required by Thrust to be one or two arguments. In the latter case, it's for alignment AFAIK. If we need the current stream, we should look it up in the Cython space, not in here. (And this circles back to my point above.)

emcastillo: The stream that your external library is using might be different from the current cupy one.

leofang: Ah, I see what you're saying. OK, for `cupy/cuda/cupy_thrust.cu` there's nothing to fix, because we pass in the current stream from CuPy. So your point is on the general applicability of the new C API. Let me think about it. Thanks for bringing this up, Emilio!

leofang: Let me know if I miss something here. There're 4 possibilities during allocation:
1. External library is not using any stream: The arena would be recorded on the null stream.
2. External library is using its own stream (perhaps in the C/C++ space): Same as in 1, the arena is recorded on the null stream, since CuPy has no way to know the external stream.
3. External library is using a CuPy stream: In this case, the CuPy stream can be picked up internally.
4. External library is using its own stream (either in C/C++ or Python) *and* a CuPy stream in Python: The arena is recorded on the CuPy stream. 

Synchronization issues aside, it seems to me if during deallocation the stream usage remains unchanged (i.e., using the same stream as in allocation), we will be able to locate the correct arena?

cupy/cuda/memory.pyx
emcastillo: LGTM!

leofang:
Sorry for my delay in responding some very interesting feedbacks from Jake and Mark. Thanks @harrism for reminding me @jrhemstad's edits.

> In short, there's no problem here.
> This is why you shouldn't let C++ people look at Python code :)

Yeah thanks Jake! Should have been more clear in my description...😅 Also thanks to @jakirkham for making a better explanation.

> For example, I built a hybrid_memory_resource in RMM which simply has a size threshold, and pointers two upstream memory_resource objects.

Thanks for the pointer, Mark! With a working example it's so much easier to understand why this is necessary. I was just digging back RMM's issue tracker, and found you folks have already discussed the very same issue here: rapidsai/rmm#302. 

> Not like CUB, no. You could fake this by creating a Thrust allocator facade to some pre-allocated memory and pass it as a custom exec policy, but this is basically just a poor man's pool allocator.

Yeah this is what I figured...Just wanna get an authentic verification. Thanks. 

The problem with a poorman's allocator is we need to know enough the internal of the routines, so we won't be far off from the actually needed size. I think the current approach in CuPy is already the best compromise.

emcastillo:
@pentschev are you guys interested in this feature?

jakirkham:
> @pentschev are you guys interested in this feature?

(asking internally)

pentschev:
I think we were primarily curious about the intended usage for this PR, cc @kkraus14 who first raised that question internally.

I'm not aware of any use cases for this currently, as in RAPIDS we primarily use RMM as the memory manager for all CUDA-related libraries, including CuPy. That's not to say that this isn't useful though. I bet if there's someone who knows of a use case for this PR in RAPIDS, that person would be @kkraus14 .

emcastillo:
Thanks for the reply!
Would it be interesting for you guys to have some interfaces to use the RMM allocator inside cupy? or did you already solved this.

pentschev:
We solved this from the RMM side: https://github.com/rapidsai/rmm/blob/branch-0.14/python/rmm/rmm.py#L242-L262. It would probably be useful to document in CuPy that RMM can also be used in conjunction with it, any thoughts on this @emcastillo ?

jakirkham:
> Would it be interesting for you guys to have some interfaces to use the RMM allocator inside cupy? or did you already solved this.

As Peter said, this was solved already. I think Kenichi helped review.

Previously I think we discussed going the other way as well. Namely using CuPy's allocator within RMM ( https://github.com/cupy/cupy/issues/2635 ). This might still be interesting for comparative purposes. Though RMM has since added several different allocators. So we might already have plenty to do just figuring out how to combine RMM's allocators together in different ways.

leofang:
Thanks Peter and John for your comments. Indeed, if we want a universal solution supporting throughout the Python GPU stack, RMM seems far more suitable than this. As I mentioned to Emilio offline, I didn't realize CuPy's default memory pool is tied to the Python interpreter so much when I started cleaning up the Thrust module (with hindsight it is obvious...😅)

I am sensing that even we expose our pool as is done here, it is of little use outside the CuPy space, and the team might not appreciate a controversial PR just for cleaning things up. For this reason, I'm closing this PR. Feel free to revive it if you can think of other reasons.

Issue #3221, opened 2020-03-24
PyPy Support?


jakirkham:
Does CuPy support PyPy? Are there any changes users need to make to ensure a build works with PyPy? If this isn't supported, is there a sense of how much work would be required here?

leofang:
ref: https://github.com/conda-forge/cupy-feedstock/issues/43#issuecomment-603032749

`import cupy` is OK, but there's a `CUDA_ERROR_INVALID_IMAGE` driver error when trying to run CuPy with PyPy.

leofang:
I think this issue can be closed. I tested the backport PR #3227 with v7.3.0 and it worked (importable, runnable, and passed a few randomly picked tests). 

I did see the test `TestCumprod.test_cumprod_huge_array` failed, though. I got `OutOfMemoryError` for this. But we should open a separate issue to follow up. Most likely it involves how PyPy does garbage collection, which I assume must be quite different from CPython's.

jakirkham:
Yeah to summarize it sounds like CuPy supports PyPy. We may still need to iron out some issues though that are TBD. Does that sound right?

leofang:
Yes! 🙂

PR #3225, opened 2020-03-25
Fix byte buffer handling to support PyPy

okuta:
This PR solves conda-forge/cupy-feedstock#43 (#3221).

I tested on following environment.
```
Python 3.6.9 (1608da62bfc7, Dec 23 2019, 10:50:04)
[PyPy 7.3.0 with GCC 7.3.1 20180303 (Red Hat 7.3.1-5)] on linux
```

### DIFF
--- a/cupy/cuda/nvrtc.pyx
+++ b/cupy/cuda/nvrtc.pyx
@@ -120,37 +120,29 @@ cpdef compileProgram(intptr_t prog, options):
 
 cpdef unicode getPTX(intptr_t prog):
     cdef size_t ptxSizeRet
-    cdef bytes ptx
-    cdef char* ptx_ptr
+    cdef vector.vector[char] ptx
     with nogil:
         status = nvrtcGetPTXSize(<Program>prog, &ptxSizeRet)
     check_status(status)
-    ptx = b' ' * ptxSizeRet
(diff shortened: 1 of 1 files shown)


emcastillo:
cc @leofang @jakirkham

emcastillo:
Jenkins, test this please

jakirkham:
Thanks Ryosuke! 😀

leofang:
Thanks for the quick action, @okuta @emcastillo! Question: isn’t the usage of `&vector[0]` against #3020?

toslunar:
> Question: isn’t the usage of `&vector[0]` against #3020?

Lengths here seem at least 1 due to the trailing NULL.  #3022 is a proposal to avoid `&vector[0]` always as a coding rule.

emcastillo:
`&vector[0]` should be avoided in cases that arrays can be empty, but I think it is fine if we know that the array always have values.

okuta:
> Lengths here seem at least 1 due to the trailing NULL. #3022 is a proposal to avoid &vector[0] always as a coding rule.
I think that `ptxSizeRet` >= 1, because it contains NULL character at the end.

leofang:
I see, thanks! 🙂

PR #3228, opened 2020-03-26
Align vector access with #3020 #3022

emcastillo:
As per the discussion in #3225 

Can be closed if not necessary


takagi:
LGTM except the conflicting. Can I fix it as it's so trivial?

emcastillo:
This fix is only valid as long as we are sure that the buffer is not empty.

.data() in an empty vector is not  defined 😭

takagi:
I've rebased it on the current master and made a change that explicitly checks empty sizes. I think it makes sense here. Would you take a look?

emcastillo:
LGTM

leofang:
Tests weren't run...?

emcastillo:
😂　thanks @leofang !

PR #3229, opened 2020-03-26
Add `cupy.shape`

toslunar:


leofang:
Interesting. Didn't know this wasn't implemented. Would this allow `__array_function__` to take effect, so that `numpy.shape(cupy_array)` works?

toslunar:
> Would this allow `__array_function__` to take effect, so that `numpy.shape(cupy_array)` works?

Yes.
https://github.com/cupy/cupy/blob/766dfd04703da1d9a971d1124951c24b91bfbc33/cupy/core/core.pyx#L1351-L1367

leofang:
Thanks, @toslunar! I was a bit confused by the logic of `__array_function__` there; looks like if a name is in the `cupy` module, the dispatch will happen?

takagi:
LGTM! Waiting for auto-merge.

takagi:
pfnCI, test this please.

PR #3230, opened 2020-03-26
Fix coding style in Cython files

okuta:
`ignore = W503,W504` is not necessary to format Ctyhon files.

### DIFF
--- a/cupy/core/_reduction.pyx
+++ b/cupy/core/_reduction.pyx
@@ -180,7 +180,7 @@ cdef Py_ssize_t _get_contiguous_size(
     return contiguous_size
 
 
-cpdef (Py_ssize_t, Py_ssize_t, Py_ssize_t) _get_block_specs(  # NOQA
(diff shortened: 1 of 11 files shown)


cupy/core/_reduction.pyx
```diff
@@ -180,7 +180,7 @@ cdef Py_ssize_t _get_contiguous_size(
     return contiguous_size
 
 
-cpdef (Py_ssize_t, Py_ssize_t, Py_ssize_t) _get_block_specs(  # NOQA
+cpdef(Py_ssize_t, Py_ssize_t, Py_ssize_t) _get_block_specs(  # NOQA
```
leofang: Is this a typo? I think we need the space...

okuta: Thanks, I fixed.

toslunar:
PEP 8 recommends line breaks *before* binary operators, while `W504` is off in the `flake8` defaults.

okuta:
OK, Under the current rules, there are differences between Python and Cython. 
Please let me know if the rules are decided.

PR #3236, opened 2020-03-30
Refactor `__array_function__`

toslunar:
In the implementation of `ndarray.__array_function__`, this change
- removes the antipattern "`getattr` after `if hasattr`", and
- merges the `except AttributeError` logic.


asi1024:
Jenkins, test this please.

PR #3237, opened 2020-03-30
Make nvrtc `getPTX` use `bytes` instead of `unicode`

emcastillo:
This is the first step to add support for hiprtc.

In some environments, hiprtc byte encoding when returning the compiled binary is not utf-8, so we can't rely on it.

This PR just deals with the intermediate code by keeping it as bytes. Thus making it encoding agnostic.


kmaehashi:
Could you rebase?

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
@kmaehashi @emcastillo PTAL

kmaehashi:
LGTM!

leofang:
Thanks @emcastillo @kmaehashi!

PR #3238, opened 2020-03-30
Add hiprtc support

emcastillo:
Merge after #3237 

Allows using hiprtc

Thanks to @okuta for nvrtc->hiprtc bindings


emcastillo:
Probably 
https://github.com/cupy/cupy/blob/c3eb998edeb806583b1a7f84b129449796726b87/cupy/cuda/compiler.py#L120

Needs some tunning

cupy/cuda/compiler.py
```diff
@@ -538,7 +538,7 @@ def _convert_to_hip_source(source):
 
 
 def _compile_with_cache_hipcc(source, options, arch, cache_dir, extra_source,
```
leofang: Could you rename this function to, say, `_compile_with_cache_hip` to be in line with the naming convention of `_compile_with_cache_cuda`, given that it now has dual roles?

cupy/cuda/compiler.py
```diff
@@ -582,9 +582,12 @@ def _compile_with_cache_hipcc(source, options, arch, cache_dir, extra_source,
                 mod.load(binary)
                 return mod
 
-    # TODO(leofang): catch HIPCCException and convert it to CompileException
-    # with backend='hipcc'
-    binary = _hipcc(source, options, arch)
+    if backend == 'nvrtc':
```
leofang: 1. Ditto for the backend name:
```suggestion
    if backend == 'hiprtc':
```
2. Could you leave a short comment here to explain `compile_using_nvrtc` calls hiprtc for HIP builds?

cupy/cuda/cupy_rtc.h
```diff
@@ -0,0 +1,15 @@
...
+
+#if CUPY_USE_HIP
+
+#include "cupy_hiprtc.h"
+
+#else
+
+#include "cupy_nvrtc.h"
+
+#endif
```
leofang: I guess the empty lines were reserved for inserting more functions in the future? Would be nice to remove them, though.

cupy/cuda/compiler.py
```diff
@@ -538,7 +538,7 @@ def _convert_to_hip_source(source):
 
 
 def _compile_with_cache_hipcc(source, options, arch, cache_dir, extra_source,
-                              use_converter=True):
+                              backend='nvrtc', use_converter=True):
```
leofang: Wouldn't it be nice to have a separate backend named "hiprtc"? The backend name shows up in the `CompileException` I think.
```suggestion
                              backend='hiprtc', use_converter=True):
```

emcastillo: The problem I see here that I am not sure how to solve it in a clean way is that what would be the default value now.

Something like `"rtc"`, should we convert the nvrtc string to hiprtc before calling `compile_with_hipcc`?

leofang: Maybe I missed something obvious, but shouldn't [my other comment](https://github.com/cupy/cupy/pull/3238#discussion_r400583951) addresses this?

emcastillo: I am not sure I am following you, The backend default name is 'nvrtc', and so is the default value for the `backend` argument in the top-level function.
So if we change that for hip, we need to either do an 
```python
if backend == 'nvrtc'
   backend = 'hiprtc'
```
Or use an agnostic backend name for both such as `rtc`.

leofang: Ah, let me guess what could be confusing. How about overwriting the backend in `compile_with_cache`? So, it'd be something like this:
```python
def compile_with_cache(..., backend='nvrtc', ...):
    ...
    if runtime.is_hip:
        backend = 'hiprtc' if backend == 'nvrtc' else 'hipcc'
        # name changed from "_compile_with_cache_hipcc" to "_compile_with_cache_hip"
        return _compile_with_cache_hip(..., backend=backend, ...)
    else:
        return _compile_with_cache_cuda(..., backend=backend, ...)
```

emcastillo: I am not a big fan of doing this kind of things, but I'll add it since changing the global backend name leads to no-compat.

leofang: Yeah I see your point. But I guess ad-hoc hacks like this would be inevitable unless we split the HIP support from `cupy.cuda`...

leofang:
I'd like to have this PR also blocked by #3823.

emcastillo:
rebased, Jenkins test this please

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo:
PTAL

kmaehashi:
LGTM!

PR #3244, opened 2020-03-31
Performance boost: CUB-backed `_SimpleReductionKernel`

leofang:
~~Blocked by #2584.~~ ~~Blocked by #3253.~~ ~~Blocked by #3262.~~

~~## Important note~~
~~**DO NOT REVIEW the code yet!**~~

~~To core devs: I need a quick turnaround feedback first before investing time in polishing this PR. Please read the PR description below. The current implementation is admittedly not perfect, and there're a number of things to be addressed/fixed (see [TODO](#todo)) in order to make code review more comfortable.~~

## Motivation
With the `cupy.cuda.cub` module we were able to speed up a few commonly encountered reduction operations (`sum`, `min`, `max`, etc), but it would be nice to speed up ***all*** of the reduction operations currently existing in (or to be added to) CuPy (see the list [here](https://gist.github.com/leofang/b466291ea822dcb6eafbf85512315a06#file-test_cub_reduction_kernel4-py-L13-L16)). 

However, it is very tedious to add the desired support; we've seen how error-prone it was, and how many additional patches are needed to make it work. Even if we can do these for all reduction functions one by one, the maintenance is going to be difficult. Therefore, it is preferable, if possible, to build on top of the existing infrastructure for code generation in runtime.

## What is being done
This PR achieves the goal by providing an alternative reduction kernel implementation based on `cub::BlockReduce`. **All reduction functions returned by `create_reduction_func`**, which is in turn based on `_SimpleReductionKernel`, **will be benefited if the reduced axes are contiguous** (can be in either C or F order). The net effect is as if `cub::DeviceReduce` or `cub::DeviceSegmentedReduce` were used, depending on whether it's a full or partial reduction.

## What is *not* included
- Generalization to arbitrary `ReductionKernel`: I think it'll have to wait until this PR is merged, as I don't understand a few things there. I will likely need to discuss with the core devs offline.
- Support NVRTC: currently this PR requires using `nvcc` (see below) due to NVlabs/cub#131, but I think as suggested there this issue can be circumvented by using [jitify](https://github.com/NVIDIA/jitify).
- Reduction over non-contiguous arrays: I don't think it's possible with CUB...

## Implementation detail & limitations
~~Will add this later.~~ See https://github.com/cupy/cupy/pull/3244#issuecomment-608097646.

## Performance benchmarks
Using [this script](https://gist.github.com/leofang/b466291ea822dcb6eafbf85512315a06) in which all reduction functions in CuPy (searched via grep `create_reduction_func`) are tested and benchmarked, I see that in most cases, especially a full reduction, a significant boost is achieved. 

For CUDA 9.2 + P100, this is [what I got](https://gist.github.com/leofang/b466291ea822dcb6eafbf85512315a06#gistcomment-3235183). 
For CUDA 10.0 + GTX 2080 Ti, this is [what I got](https://gist.github.com/leofang/b466291ea822dcb6eafbf85512315a06#gistcomment-3236216).
**(UPDATE)**: https://github.com/cupy/cupy/pull/3244#issuecomment-639332979

## TODO
- [x] Merge with the current master: The conflict is due to #2920, which I need more time to understand, so I won't proceed further unless I get a green light from the core devs.
- [x] Merge `cupy/core/_cub_simple_reduction.py` back to `cupy/core/_reduction.pyx`. (I added the kernel in a .py file for faster dev & test cycles.)
- [ ] Discuss how to enable/disable this feature to make it easier to both users and our CIs. Toggling on/off this feature is currently done by the new flag `cupy.core.cub_block_reduction_enabled`, but frankly I am not a fan of it.
- [ ] Analyze why in certain cases CUB does not win
- [x] Provide some architecture tuning parameters such as `ITEMS_PER_THREAD`, `BLOCK_SIZE`, etc. This is how it's done in CUB, but perhaps we can defer this to the next PR.
- [x] Clean up all the mess and polish the code.
- [x] ~~(Optional) Force-push to clean up the commit history~~ (need to delete 3e8eb62)


leofang:
@emcastillo Sorry for my slow addition...Here it is.

## Implementation detail & limitations
I hijack `_AbstractReductionKernel._call()` and split the code path into two. If the conditions for using CUB is met, I then pass most of the generated strings (`pre_map_expr`, `post_map_expr`, `reduce_type`, etc) to the string template of the CUB-based kernel for runtime code generation.
(Some strings like `params` are totally different from the current implementation, see below.)

To compile the code, note that NVRTC does not support CUB natively (NVlabs/cub#131), but fortunately we can call the NVCC backend. Will address this issue in a future PR using jitify.

Because we only invoke the CUB kernel for reductions over contiguous axes, it allows a number of simplifications. First, the stride sizes are no longer needed, and instead, for partial reduction like `(a, b, c) -> (a,)`, we just need to know the “segment” size = `b * c` (following the terminology in CUB). I decide to pass in this info as a kernel argument, so that the generated kernel can be reused for different segment sizes. This also means we don’t need `CArray` and `CIndexer` at all, so I just pass the pointers to the input and output arrays to the kernel. In this case, each block handles one segment, so the grid size is simply `a`.

Second, for full reduction like `(a, b, c) -> ()`, the performance of the above “one-pass” strategy is very bad, as only one block is used! So, in this case we need a “two-pass” reduction -- in the first pass, we assume a fair share among certain number of blocks, say `N`, so that each block handles a segment of size `(a * b * c) / N`. Then, in the second pass we invoke 1 block of `N` threads to handle the single remaining segment and get the final result.

The most challenging part is to allocate a work space to hold the intermediate result between the two passes. Long story in short: we don’t know in advance what’s `sizeof(reduce_type)`. Currently I try to be on the safe side and allocate `32 * N` bytes for this purpose. This seems to work for all reductions in CuPy. Anyway, a number of C macros are defined for this purpose such that the string template can handle all cases (one-pass, the first pass, and the second pass).

In any case, items per thread and block size are two compile-time constants in order to specialize a CUB kernel. In a future PR they should be tuned for different architectures, following more or less what CUB is doing for device policies. Also, I realized the strategies taken here are more or less following the internal implementations of `cub::DeviceReduce` and `cub::DeviceSegmentedReduce`. This is why I said
> The net effect is as if `cub::DeviceReduce` or `cub::DeviceSegmentedReduce` were used, depending on whether it’s a full or partial reduction.

One thing I’m considering: there’re some hacks done to make `mean()` work, which requires to know the input and output array sizes. Currently I hard-coded them in the generated kernel, but perhaps passing them to the kernel as arguments can further increase the kernel reusability.

emcastillo:
Related to #2796

cupy/core/_reduction.pyx
```diff
@@ -107,6 +110,203 @@ extern "C" __global__ void ${name}(${params}) {
...
+        name, block_size, items_per_thread,
+        reduce_type, params, arginfos, identity,
+        pre_map_expr, reduce_expr, post_map_expr,
+        type_map, preamble, options)
+
+    # CUB is not natively supported by NVRTC (NVlabs/cub#131), so use nvcc
+    # instead. For this, we have to explicitly spell out the default values for
+    # arch, cachd, and prepend_cupy_headers to bypass cdef/cpdef limitation...
+    # TODO(leofang): investigate Jitify for using NVRTC (also NVlabs/cub#171)
+    module = compile_with_cache(module_code, options, arch=None, cachd_dir=None,
```
emcastillo, line 132: I would use some kind of context manager here to change the backend temporarily rather than to directly call the compiler in _reduction.pyx

leofang: I am not aware of any utility we have to switch the backend in a context. `compile_with_cache` is what's used in `_reduction.pyx`. Since this is meant to be private API, I think we wanna reduce overheads as much as possible. We can revisit this in the next PR when jitify is used to support NVRTC.

A related issue is what'd happen in an HIP build. I think this feature would need to be disabled until we (I?) figure out how to use hipCUB...

leofang:
Documenting progress for myself: Fixed a few memory corruptions, and I hope this is the only one left. Will resume when I have time...
```python
import cupy as cp
import numpy as np


def test_sum_all2(dtype):
    a_np = cp.testing.shaped_arange((20, 30, 40), np, dtype=dtype)
    out_np= a_np.sum()

    a_cp = cp.testing.shaped_arange((20, 30, 40), cp, dtype=dtype)
    out_cp= a_cp.sum()
    print(dtype, out_np, out_cp, cp.allclose(out_np, out_cp))


test_sum_all2(np.float64)
test_sum_all2(np.float32)
test_sum_all2(np.float16)
test_sum_all2(np.int8)    # <--- buggy, why?
```
The reason I believe it's also a memory corruption is that if I just run the last line (`test_sum_all2(np.int8)`) without others, the result is correct.

leofang:
> The reason I believe it's also a memory corruption is that if I just run the last line (`test_sum_all2(np.int8)`) without others, the result is correct.

Fixed in 6885655!

leofang:
All tests in `tests/cupy_tests` passed! I will proceed to clean up/refactor/support Optuna.

cupy/core/_reduction.pyx
```diff
@@ -407,29 +896,104 @@ cdef class _AbstractReductionKernel:
             best.user_attrs['block_stride'],
             best.params['out_block_num'])
 
+    def _get_cub_optimized_params(
```
leofang, line 899: @asi1024 Continuing our discussion on the optuna optimization from [here](https://github.com/cupy/cupy/pull/2731#discussion_r435337564), could you take a look at my implementation here? For CUB-based kernels I currently allow two parameters to be tuned: `block_size` and `items_per_thread`. On my local tests I do see Optuna could find better parameters, and I will post the benchmark later.

cupy/core/_reduction.pyx
```diff
@@ -407,29 +896,104 @@ cdef class _AbstractReductionKernel:
...
+            block_stride = block_size * items_per_thread
+            cub_params = (
+                True, items_per_thread, contiguous_size, full_reduction)
+            self._launch(
+                out_block_num, block_size, block_stride, in_args, out_args,
+                in_shape, out_shape, type_map, map_expr, reduce_expr,
+                post_map_expr, reduce_type, stream, params, cub_params)
+
+        def suggest_func(trial):
+            block_size_log = trial.suggest_int('block_size_log', 5, 10)
```
leofang, line 933: Hi @asi1024 One more question: When I changed the upper bound to `10` here (so that there could be 2^10=1024 threads in a block), in an Optuna context I would occasionally see `CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES`. Can't Optuna handle such a failure gracefully? I would expect Optuna is smart enough to avoid trying this parameter and find the next.

leofang:
@emcastillo I am going to make this PR unblocked by #2584 so that it can be queued for review sooner.

leofang:
@emcastillo This is ready for review! a34c060 make it decoupled from #2584, so nothing blocks this PR now. Two env vars are used:
1. `CUPY_CUB_PATH` is respected at runtime to discover the CUB location.
2. `CUPY_CUB_BLOCK_REDUCTION_DISABLED` (analogous to `CUB_DISABLED`) to toggle this feature on and off (default is off). This can be used in your CI.

Below I give a performance benchmark script and result. The results with Optuna turned on are labeled with `opt`. General remarks:
1. Without Optuna: CUB block reduction (this PR) generally performs much better than the CuPy reduction, occasionally even better than CUB device reduction (from `cupy.cuda.cub`) if it's available (most of time CUB block is slightly slower than CUB device).
2. With Optuna: CUB block reduction can (but not always) be further optimized, so at least it's safe to claim the optimized context implementation is working, and the optimized CUB block reduction always wins the optimized CuPy reduction.

<details><summary>script: (click me)</summary>

```python
import sys
import warnings
warnings.filterwarnings("ignore", category=FutureWarning)

import numpy as np
import cupy as cp
from cupyx.time import repeat
from cupyx import optimizing


shape = (256, 512, 512)
dtype = cp.float32

a = cp.random.random(shape).astype(dtype)
a_np = cp.asnumpy(a)
CUB_supported = ['sum', 'prod', 'min', 'max', 'argmin', 'argmax']
REST = ['amin', 'amax', 'nanmin', 'nanmax', 'nanargmin', 'nanargmax',
        'mean', 'nanmean', 
        'var', 'nanvar', 'nansum', 'nanprod',
        'all', 'any', 'count_nonzero']

for reduce_func in CUB_supported + REST:
    for axis in [(2,), (1, 2), (0, 1, 2)]: # [(0,1,2)]:  
        print("testing", reduce_func, "with axis = ", axis, '...')
        func = getattr(cp, reduce_func)

        # get numpy answer for comparison
        if reduce_func not in ('argmin', 'argmax', 'nanargmin', 'nanargmax'):
            ans = getattr(np, reduce_func)(a_np, axis)
        elif len(axis) == 1:
            ans = getattr(np, reduce_func)(a_np, axis[0])
        else:
            ans = None

        cp.cuda.cub_enabled = False
        cp.core.cub_block_reduction_enabled = False
        data = repeat(func, (a, axis), n_repeat=100)
        results = [data._to_str_per_item('GPU', data.gpu_times), data._to_str_per_item('CPU', data.cpu_times)]
        print('{:<15s} (old kernel):{}'.format(reduce_func, ' '.join(results)))
        b = func(a, axis)
        
        cp.cuda.cub_enabled = False
        cp.core.cub_block_reduction_enabled = False
        with optimizing.optimize(key=None):
            data = repeat(func, (a, axis), n_repeat=100)
            results = [data._to_str_per_item('GPU', data.gpu_times), data._to_str_per_item('CPU', data.cpu_times)]
        print('{:<11s} (old opt kernel):{}'.format(reduce_func, ' '.join(results)))
        f = func(a, axis)

        if reduce_func in CUB_supported:
            cp.cuda.cub_enabled = True
            cp.core.cub_block_reduction_enabled = False
            data = repeat(func, (a, axis), n_repeat=100)
            results = [data._to_str_per_item('GPU', data.gpu_times), data._to_str_per_item('CPU', data.cpu_times)]
            print('{:<15s} (CUB device):{}'.format(reduce_func, ' '.join(results)))
            c = func(a, axis)
        else:
            print('{:<15s} (CUB device):{}'.format(reduce_func, '    (CUB device-wide reduction not available)'))
            c = None

        cp.cuda.cub_enabled = False
        cp.core.cub_block_reduction_enabled = True
        data = repeat(func, (a, axis), n_repeat=100)
        results = [data._to_str_per_item('GPU', data.gpu_times), data._to_str_per_item('CPU', data.cpu_times)]
        print('{:<15s} (CUB blocks):{}'.format(reduce_func, ' '.join(results)))
        d = func(a, axis)

        cp.cuda.cub_enabled = False
        cp.core.cub_block_reduction_enabled = True
        with optimizing.optimize(key=None):
            data = repeat(func, (a, axis), n_repeat=100)
            e = func(a, axis)
        results = [data._to_str_per_item('GPU', data.gpu_times), data._to_str_per_item('CPU', data.cpu_times)]
        print('{:<11s} (CUB opt blocks):{}'.format(reduce_func, ' '.join(results)))

        try:
            cp.cuda.cub_enabled = False
            cp.core.cub_block_reduction_enabled = False
            if ans is not None:
                assert cp.allclose(ans, b, rtol=1E-4)
                if c is not None:
                    assert cp.allclose(ans, c, rtol=1E-4)
                assert cp.allclose(ans, d, rtol=1E-4)
                assert cp.allclose(ans, e, rtol=1E-4)
                assert cp.allclose(ans, f, rtol=1E-4)
        except AssertionError:
            print("Result not match! (function: {}, axis: {})".format(reduce_func, axis), file=sys.stderr)
            print(ans, b, d, e)
            raise
        finally:
            print()
```
</p>
</details>

<details><summary>result: (click me)</summary>

[bench_test_cub_reduction_kernel5-3.txt](https://github.com/cupy/cupy/files/4734770/bench_test_cub_reduction_kernel5-3.txt)
</p>
</details>

I think the existing test suite should provide enough coverage, but if you wish I can try adding a few simple tests targeting this feature.

cupy/core/_reduction.pyx
```diff
@@ -514,7 +1112,7 @@ cdef class _SimpleReductionKernel(_AbstractReductionKernel):
         reduce_dims = True
         return self._call(
             in_args, out_args,
-            arr._shape, axis, dtype, keepdims, reduce_dims, dev_id, None)
+            arr._shape, axis, dtype, keepdims, reduce_dims, dev_id, None, True)
```
emcastillo, line 1115: I am not sure if a default `True` for this is good.

emcastillo:
I would specialize the `_AbstractKernel` into `_CubSimpleAbstractReductionKernel` so that we have a separation of concerns

cupy/core/_reduction.pyx
emcastillo: The optimization looks good, but can they be merged somehow?
Maybe, it would be good to have the `use_cub` as a hyperparameter itself?

leofang:
> I would specialize the `_AbstractKernel` into `_CubSimpleAbstractReductionKernel` so that we have a separation of concerns

That would make the code cleaner indeed, but I don't see how it could be done given the goal is to hijack `create_reduction_func()`. We need a lot of preparation (and code petition) before we can decide if CUB is usable or not. Any thoughts?

emcastillo:
Yeah, I see that, I wonder if it would be ok moving some of it around using a Factory pattern.

The Factory just does all that common preparation and then spits out a CUB or an old style kernel

leofang:
> The Factory just does all that common preparation and then spits out a CUB or an old style kernel

Isn't `_AbstractReductionKernel` already the best factory for us? 😇 

Actually, I just recalled that in the very beginning I wanted to support `ReductionKernel` too, so I thought the base class is the best place to modify. But then I had trouble in understanding `ReductionKernel` (for example, why it could take multiple inputs and outputs? Which input is reduced to which output? How does it work with `@cupy.fuse`? etc), so I put that ambitious plan on hold and focused on `_SimpleReductionKernel`. So I'd think any decision made here should have this possible extension in mind.

cupy/core/_reduction.pyx
asi1024: Addressed this comment in PR #3410.

leofang: Ah, thanks for quick action @asi1024! So it's not Optuna's fault, but just how `_optimize` is implemented. Glad to know.

emcastillo:
> How does it work with @cupy.fuse

This might be actually an issue here

leofang:
Right, I worried about that too, but luckily so far no problem (I ran the entire test suite locally several times). I guess it's because `@cupy.fuse` uses `ReductionKernel` instead of `_SimpleReductionKernel`, so it didn't even bother to try CUB (I set `try_use_cub=False` in the former).

cupy/core/_reduction.pyx
leofang: > The optimization looks good, but can they be merged somehow?

I tried, but I realized the number of lines was not reduced, mainly because the bulk of the two `_get*optimized_params` functions, such as `target_func`, `suggest_func`, and `optimize_impl`, are different. Would it help if I pull them out to the module level, combine `_get*optimized_params`, and select which to call based on `use_cub`?

> Maybe, it would be good to have the `use_cub` as a hyperparameter itself?

This is interesting. I haven't thought about it. In my benchmark CUB is a clear winner, but for smaller problem sizes it might help (not sure). But, the hyperparameter would have to be `try_use_cub` instead of `use_cub`, because otherwise some extra work needs to be done for switching back and forth between the two implementations. I suggest we figure out how to deal with your factory function idea first, and then revisit this.

leofang:
Hi @asi1024, with #3410 and the following patch I still see the traceback associated with `cupy.cuda.driver.CUDADriverError` being printed out, but the optimization continues without disruption. Is this expected?
```diff
diff --git a/cupy/core/_reduction.pyx b/cupy/core/_reduction.pyx
index e932f2e74..c20612398 100644
--- a/cupy/core/_reduction.pyx
+++ b/cupy/core/_reduction.pyx
@@ -933,6 +933,8 @@ cdef class _AbstractReductionKernel:
             self, optimize_config, in_args, out_args, in_shape, out_shape,
             type_map, map_expr, reduce_expr, post_map_expr, reduce_type,
             stream, full_reduction, out_block_num, contiguous_size, params):
+        from cupy.cuda.driver import CUDADriverError
+
         out_size = internal.prod(out_shape)
         in_args = [_optimizer_copy_arg(a) for a in in_args]
         out_args = [_optimizer_copy_arg(a) for a in out_args]
@@ -951,7 +953,7 @@ cdef class _AbstractReductionKernel:
                 post_map_expr, reduce_type, stream, params, cub_params)
 
         def suggest_func(trial):
-            block_size_log = trial.suggest_int('block_size_log', 5, 9)
+            block_size_log = trial.suggest_int('block_size_log', 5, 10)
             block_size = 2 ** block_size_log
             items_per_thread = trial.suggest_int(
                 'items_per_thread', 2, 32, step=2)
@@ -962,10 +964,10 @@ cdef class _AbstractReductionKernel:
         optimize_impl = optimize_config.optimize_impl
         best = optimize_impl(
             optimize_config, target_func, suggest_func,
-            default_best={
-                'block_size_log': default_block_size_log,
-                'items_per_thread': default_items_per_thread,
-            })
+            default_best={'block_size_log': default_block_size_log,
+                          'items_per_thread': default_items_per_thread,},
+            # CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES might happen
+            ignore_error=(CUDADriverError,))
 
         return best.params['items_per_thread'], best.user_attrs['block_size']
 
```

asi1024:
Hmm, it's unexpected behavior. Could you show me a reproducer?

leofang:
I simply set up this PR with the above patch, set the two env vars in https://github.com/cupy/cupy/pull/3244#issuecomment-639332979, and then run the script in the comment. Not sure if there's a better, shorter, and deterministic reproducer I can provide. Perhaps I can rerun and post the full output log?

asi1024:
@leofang I agree with the support of CUB-backed _SimpleReduction to unify some internal implementations (e.g., allocations of output memory, decisions of dtypes).
However, I don't want to heavily mix the implementation for cub-support in `_reduciton.pyx`. We will also support cuTENSOR or other external libraries in reduction functions in the future, so I want to keep extensibility for multi-backends.
Could you split the implementations into another file `_cub_reduction.pyx`?

leofang:
Hi @asi1024 

> Could you split the implementations into another file `_cub_reduction.pyx`?

I thought about this too, after @emcastillo suggested the possibility of making a factory method. However, I really don't see how I can do this. I still think the only way to support CUB (or any other backends) is to intercept `_AbstractReductionKernel._call()`. Any suggestions? Perhaps I just call a function defined in `_cub_reduction.pyx` from `_AbstractReductionKernel._call()`?

asi1024:
I think it's OK just to call a function defined in `_cud_reduction.pyx` in this PR.

leofang:
@asi1024 I am working on the change you suggested. Although I have not finished this migration yet, I already foresee a problem. 

I was able to move all `_*cub*()` helper functions to `_cub_reduction.pyx`, but not the logic in `_AbstractReductionKernel._call()`. The reason is the number of lines in `_call()` would not decrease: I need to pass in a whole lot of arguments to `_cub_reduction.pyx`, get them processed, and return a bunch of parameters back to `_call()` for kernel launch. I think this change only makes it harder to maintain. May I suggest keeping `_call()` as is?

emcastillo:
Is it not possible to add a `_AbstractCUBReductionKernel` that inherits `_AbstractCUBReductionKernel` and overrides whats needed?

leofang:
Did you mean to inherit `_AbstractReductionKernel` and create a `_SimpleCUBReductionKernel`? Yes, it's always possible. The problem is we must hijack and split the code path somewhere, so the problem is three-fold:
1. If we do any alternative approach (other than the current implementation), where to split? 
2. At the split point, the code would be a bit messy (due to the need to test some conditions to decide which path). Can you accept that?
3. Would the proposal be flexible enough that we can easily extend it later (see https://github.com/cupy/cupy/pull/3244#issuecomment-640392142).

My concern for No. 1 is that the split point may have to happen as early as `create_reduction_func` is called if we do this inheritance, but we won't be prepared sufficiently to make any judgement (for using CUB or not) at that point. I don't see where in the existing code path the `_SimpleCUBReductionKernel` can enter . 

@asi1024 @emcastillo I think in the above discussions we are kinda mix these issues together. I will wait for a consensus to be reached and proceed to change.

emcastillo:
I think that all the points that you mention make complete sense.

I will merge the PR in its current state and then try to re-factor it myself while working with @asi1024 in an extensible design.

Sorry for kept asking changes.

emcastillo:
Jenkins, test this please

leofang:
That was quick! Thanks a lot @emcastillo! Yeah no worries, those are perfectly legit requests. I just don't have good ideas (for now). We can work together to improve the designs. 

I will also in the meantime start looking into hijacking `create_ufunc` (don't be mad at me 🙇‍♂️) and see if CUB helps (I hope it does!)

cupy/_environment.py
```diff
@@ -60,6 +69,24 @@ def _get_nvcc_path():
     return shutil.which('nvcc', path=os.path.join(cuda_path, 'bin'))
 
 
+def _get_cub_path():
+    # runtime discovery of CUB headers
+    cuda_path = get_cuda_path()
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+
+    if 'CUPY_CUB_PATH' in os.environ:
+        _cub_path = os.environ['CUPY_CUB_PATH']
```
asi1024, line 78: What is the difference between `CUB_PATH` and `CUPY_CUB_PATH`?

asi1024:
@leofang What is the difference between `CUB_PATH` and `CUPY_CUB_PATH`?

leofang:
@asi1024 Thanks for asking, I should have remember to leave a note 😅

Currently: 
- `CUB_PATH` is used at build time to build `cupy.cuda.cub`
- `CUPY_CUB_PATH` is used at runtime to support this PR

After #2584 is merged:
-  `CUB_PATH`  is used at build time to build `cupy.cuda.cub`, but will be deprecated
- `CUPY_CUB_PATH` is used at build time to build `cupy.cuda.cub` and at runtime to support this PR

For the purpose of supporting Conda-Forge, it's best to have all package-specific env vars to be prefixed by the package name (i.e., `CUPY_`) whenever possible, so that we can make better distinction and reduce interference among packages.

asi1024:
@leofang Thanks!

jakirkham:
Do you have some recent benchmarks with these changes Leo?

leofang:
Hi John, see https://github.com/cupy/cupy/pull/3244#issuecomment-639332979 (sorry GitHub folded the long thread...) We might get some slight improvement in master, because after posting that result I also did a few cosmetic changes (loop unroll, inline, etc). If you need I can rerun the script.

asi1024:
Can we remove these lines for each routine now?
https://github.com/cupy/cupy/blob/57235e7b203dd3c666762fdcbc4812606a18b896/cupy/core/_routines_math.pyx#L94-L99

leofang:
Not really, no. Yeah it would be great if these branches (which means all reduction routines from `cupy.cuda.cub`!) could be eliminated, but AFAIK they need to stay for now for a few reasons:
1. It seems when they are usable, they are usually the fastest, although there are exceptions;
2. There is a possibility that `nvcc` is not in `PATH` (say using conda as in #3403, or on some HPC computing nodes to which the full compiler toolchain is not visible) so this PR would not be usable:
https://github.com/cupy/cupy/blob/9c60736ee212b956b2eb0e5337972276ad75a631/cupy/core/_reduction.pyx#L350-L352
I would suggest to first merge #2584 and then evaluate.

emcastillo:
Exactly, the cub reductions in some routines are special cases in cub that are (or should be) more optimized than the general reduction algorithm.

leofang:
Ah, forgot to add -- perhaps it is obvious -- that the current implementation ensures the calling precedence is 1. routines from `cupy.cuda.cub`, 2. this PR, 3. CuPy's original reduction implementation.

PR #3250, opened 2020-04-02
Implement `flatiter.base` property

takagi:
This PR implements `flatiter.base` proeprty that is a reference to the array that is iterated over.

### DIFF
--- a/tests/cupy_tests/indexing_tests/test_iterate.py
+++ b/tests/cupy_tests/indexing_tests/test_iterate.py
@@ -8,6 +8,14 @@ import cupy
 from cupy import testing
 
 
+@testing.gpu
+class TestFlatiter(unittest.TestCase):
+
+    def test_base(self):
+        a = cupy.zeros((2, 3, 4))
+        assert(a.flat.base is a)
+
+
 @testing.parameterize(
     {'shape': (2, 3, 4), 'index': Ellipsis},
     {'shape': (2, 3, 4), 'index': 0},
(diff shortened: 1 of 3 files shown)


tests/cupy_tests/indexing_tests/test_iterate.py
```diff
@@ -8,6 +8,14 @@
 from cupy import testing
 
 
+@testing.gpu
+class TestFlatiter(unittest.TestCase):
+
+    def test_base(self):
+        a = cupy.zeros((2, 3, 4))
+        assert(a.flat.base is a)
```
toslunar, line 16:
```suggestion
        assert a.flat.base is a
```

takagi: Fixed!

toslunar:
Jenkins, test this please.

takagi:
The CI fails due to a sphinx warning: `docstring of cupy.ndarray::more than one target found for cross-reference 'base': cupy.ndarray.base, cupy.flatiter.base`

Using the latest sphinx `2.4.4` does not raise the warning on my local, so updating the sphinx version on chainer-test is an option.

As another, we can suppress the warning by adding `suppress_warnings = ['ref:python']` to `cupy/docs/source/conf.py`, which is introduced at https://github.com/sphinx-doc/sphinx/pull/3869 . It seems that `'ref:python'` has quite general name but takes effect only for the warning above.

```
$ cd sphinx-1.8.5/sphinx
$ grep -r "subtype='python'" .
./domains/python.py:                           type='ref', subtype='python', location=node)
```

I suppose that the latter is enough here. Does it makes sense?

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

takagi:
Waiting https://github.com/chainer/chainer-test/pull/575 is merged.

toslunar:
Jenkins, test this please.

PR #3251, opened 2020-04-02
Implement `flatiter.__len__()` special method.

takagi:
Follows #3250 .

This PR implements `flatiter.__len__()` special method.
https://github.com/takagi/cupy/compare/flatiter-base...flatiter-len

PR #3252, opened 2020-04-02
Implement `flatiter.__next__()` special method

takagi:
Follows #3251 .

This PR implements `flatiter.__next__()` special method.
https://github.com/takagi/cupy/compare/flatiter-len...flatiter-next

Note that `flatiter.__next__()` returns an instance of `cupy.ndarray`, as opposed to NumPy's counterpart that returns an element of the iterator's base array, to keep off device synchronization.


emcastillo:
Jenkins, test this please

PR #3259, opened 2020-04-03
Add cupyx.scipy.ndimage sum, mean,standard deviation and variance

niteya-shah:
This PR adds the following routines for cupyx.scipy.ndimage

- sum
- mean
- variance

Currently this method uses Elementwise Kernels,I tried to use Reduction Kernel but couldnt get it to work, if possible can someone recommend a way to use Reduction Kernels.

This PR is still a work in progress.
Todo list

- [x] Fix variance in accordance to alpha values
- [x] Performance Testing
- [x] Add tests 
- [x] Fix Documentation 
- [x] Add more robust type checking

### DIFF
--- a/cupyx/scipy/ndimage/measurements.py
+++ b/cupyx/scipy/ndimage/measurements.py
@@ -322,13 +322,13 @@ def variance(input, labels=None, index=None):
         use_kern = True
 
     if labels is None:
-        return input.var().astype(cupy.float64)
+        return input.var().astype(cupy.float64, copy=False)
 
     if not isinstance(labels, cupy.ndarray):
         raise TypeError('label must be cupy.ndarray')
 
     if index is None:
-        return (input[labels != 0]).var().astype(cupy.float64)
+        return (input[labels != 0]).var().astype(cupy.float64, copy=False)
 
     input, labels = cupy.broadcast_arrays(input, labels)
 
@@ -336,7 +336,8 @@ def variance(input, labels=None, index=None):
         if not isinstance(index, int):
             raise TypeError('index must be cupy.ndarray or a scalar int')
         else:
-            return (input[labels == index]).var().astype(cupy.float64)
+            return (input[labels == index]).var().astype(cupy.float64,
+                                                         copy=False)
 
     mean_val, count = _mean_driver(input, labels, index, True, use_kern)
     if use_kern:
--- a/cupyx/scipy/ndimage/measurements.py
+++ b/cupyx/scipy/ndimage/measurements.py
@@ -1,25 +1,24 @@
 import numpy
 
 import cupy
-
-import string
+from cupy import util
 
 
 def label(input, structure=None, output=None):
-    """Labels features in an array
+    """Labels features in an array.
 
     Args:
         input (cupy.ndarray): The input array.
-        structure (array_like or None): A structuring element that defines \
-            feature connections. ```structure``` must be centersymmetric. If \
-            None, structure is automatically generated with a squared \
+        structure (array_like or None): A structuring element that defines
+            feature connections. ```structure``` must be centersymmetric. If
+            None, structure is automatically generated with a squared
             connectivity equal to one.
-        output (cupy.ndarray, dtype or None): The array in which to place the \
+        output (cupy.ndarray, dtype or None): The array in which to place the
             output.
-
     Returns:
-        label (cupy.ndarray): An integer array where each unique feature in \
-            ```input``` has a unique label in the array.
+        label (cupy.ndarray): An integer array where each unique feature in
+        ```input``` has a unique label in the array.
+
         num_features (int): Number of features found.
 
     .. warning::
@@ -204,7 +203,8 @@ def _kernel_finalize():
         'cupyx_nd_label_finalize')
 
 
-def _variance_kernel():
+@util.memoize()
+def _ndimage_variance_kernel():
     input_arguments = 'T image, R label, raw X index, int32 size, raw Z mean'
     output_arguments = 'raw Z ret'
     _kernel_template = """
@@ -219,40 +219,53 @@ def _variance_kernel():
                                   _kernel_template)
 
 
-def _stats(input, labels=None, index=None, mean=False):
-
+@util.memoize()
+def _ndimage_sum_kernel():
     input_arguments = 'T image, R label, raw X index, int32 size'
-    _kernel_template = string.Template("""
+    output_arguments = 'raw Y ret'
+    _kernel_template = """
     for (ptrdiff_t j = 0; j < size; j++) {
       if (label == index[j]) {
         atomicAdd(&ret[j], image);
-        $mean
         break;
       }
     }
-    """)
-    if not mean:
-        output_arguments = 'raw Y ret'
-        _kernel_template = _kernel_template.substitute(mean="")
-        ret = cupy.zeros_like(index).astype(input.dtype)
-    else:
-        output_arguments = 'raw Y ret, raw int32 count'
-        _kernel_template = _kernel_template.substitute(
-            mean="atomicAdd(&count[j], 1);")
-        ret = cupy.zeros_like(index).astype(input.dtype)
-        count = cupy.zeros_like(index).astype(cupy.int32)
+    """
+    return cupy.ElementwiseKernel(input_arguments, output_arguments,
+                                  _kernel_template)
 
-    _kernel = cupy.ElementwiseKernel(input_arguments, output_arguments,
-                                     _kernel_template, "smvkernel")
 
+@util.memoize()
+def _ndimage_mean_kernel():
+    input_arguments = 'T image, R label, raw X index, int32 size'
+    output_arguments = 'raw Y ret, raw int32 count'
+    _kernel_template = """
+    for (ptrdiff_t j = 0; j < size; j++) {
+      if (label == index[j]) {
+        atomicAdd(&ret[j], image);
+        atomicAdd(&count[j], 1);
+        break;
+      }
+    }
+    """
+    return cupy.ElementwiseKernel(input_arguments, output_arguments,
+                                  _kernel_template)
+
+
+def _stats(input, labels=None, index=None, mean=False):
     if not mean:
+        _kernel = _ndimage_sum_kernel()
+        ret = cupy.zeros_like(index, dtype=input.dtype)
         return _kernel(input, labels, index, index.size, ret)
     else:
+        _kernel = _ndimage_mean_kernel()
+        ret = cupy.zeros_like(index, input.dtype)
+        count = cupy.zeros_like(index, dtype=cupy.int32)
         return _kernel(input, labels, index, index.size, ret, count)
 
 
 def variance(input, labels=None, index=None):
-    """Calculate the variance of the values of an n-D image array, optionally \
+    """Calculate the variance of the values of an n-D image array, optionally
     at specified sub-regions.
 
     Args:
@@ -263,17 +276,23 @@ def variance(input, labels=None, index=None):
             (default), all values where `labels` is non-zero are used.
 
     Returns:
-       variance (cupy.ndarray): Values of variance, for each sub-region if \
-            `labels` and `index` are specified.
+        variance (cupy.ndarray): Values of variance, for each sub-region if
+        `labels` and `index` are specified.
 
 
     .. seealso:: :func:`scipy.ndimage.variance`
     """
     if not isinstance(input, cupy.ndarray):
         raise TypeError('input must be cupy.ndarray')
-    if input.dtype.char in 'FD':
-        raise TypeError('Complex type not supported')
-
+    # There is constraints on types because atomicAdd() in CUDA 7.5
+    # only supports int32, uint32, uint64, and float32.
+    if not issubclass(input.dtype.type,
+                      (cupy.int32, cupy.float16, cupy.float32,
+                       cupy.float64, cupy.uint32, cupy.uint64,
+                       cupy.ulonglong)):
+        raise TypeError(
+            'cupyx.scipy.ndimage.variance only supports int32, float16, '
+            'float32, float64, uint32, uint64, as data type')
     if labels is None:
         return input.var()
 
@@ -281,22 +300,25 @@ def variance(input, labels=None, index=None):
         raise TypeError('label must be cupy.ndarray')
 
     if index is None:
-        index = cupy.unique(labels)
+        return (input[labels != 0]).var()
 
     input, labels = cupy.broadcast_arrays(input, labels)
 
     if not isinstance(index, cupy.ndarray):
-        raise TypeError('index must be cupy.ndarray')
+        if not isinstance(index, int):
+            raise TypeError('index must be cupy.ndarray or a scalar int')
+        else:
+            index = cupy.asarray(index)
 
     sum, count = _stats(input, labels, index, mean=True)
     mean = sum/count
-    ret = cupy.zeros_like(index).astype(mean.dtype)
-    return _variance_kernel()(input, labels, index,
-                              index.size, mean, ret)/count
+    ret = cupy.zeros_like(index, dtype=mean.dtype)
+    return _ndimage_variance_kernel()(input, labels, index,
+                                      index.size, mean, ret)/count
 
 
 def sum(input, labels=None, index=None):
-    """Calculate the sum of the values of an n-D image array, optionally \
+    """Calculate the sum of the values of an n-D image array, optionally
        at specified sub-regions.
 
     Args:
@@ -307,16 +329,23 @@ def sum(input, labels=None, index=None):
             (default), all values where `labels` is non-zero are used.
 
     Returns:
-       sum (cupy.ndarray): sum of values, for each sub-region if \
-            `labels` and `index` are specified.
+       sum (cupy.ndarray): sum of values, for each sub-region if
+       `labels` and `index` are specified.
 
 
     .. seealso:: :func:`scipy.ndimage.sum`
     """
     if not isinstance(input, cupy.ndarray):
         raise TypeError('input must be cupy.ndarray')
-    if input.dtype.char in 'FD':
-        raise TypeError('Complex type not supported')
+    # There is constraints on types because atomicAdd() in CUDA 7.5
+    # only supports int32, uint32, uint64, and float32.
+    if not issubclass(input.dtype.type,
+                      (cupy.int32, cupy.float16, cupy.float32,
+                       cupy.float64, cupy.uint32, cupy.uint64,
+                       cupy.ulonglong)):
+        raise TypeError(
+            'cupyx.scipy.ndimage.sum only supports int32, float16, '
+            'float32, float64, uint32, uint64, as data type')
 
     if labels is None:
         return input.sum()
@@ -325,18 +354,21 @@ def sum(input, labels=None, index=None):
         raise TypeError('label must be cupy.ndarray')
 
     if index is None:
-        index = cupy.unique(labels)
+        return input.sum()
 
     input, labels = cupy.broadcast_arrays(input, labels)
 
     if not isinstance(index, cupy.ndarray):
-        raise TypeError('index must be cupy.ndarray')
+        if not isinstance(index, int):
+            raise TypeError('index must be cupy.ndarray or a scalar int')
+        else:
+            index = cupy.asarray(index)
 
     return _stats(input, labels, index, False)
 
 
 def mean(input, labels=None, index=None):
-    """Calculate the mean of the values of an n-D image array, optionally \
+    """Calculate the mean of the values of an n-D image array, optionally
        at specified sub-regions.
 
     Args:
@@ -347,16 +379,23 @@ def mean(input, labels=None, index=None):
             (default), all values where `labels` is non-zero are used.
 
     Returns:
-        mean (cupy.ndarray): mean of values, for each sub-region if \
-            `labels` and `index` are specified.
+        mean (cupy.ndarray): mean of values, for each sub-region if
+        `labels` and `index` are specified.
 
 
     .. seealso:: :func:`scipy.ndimage.mean`
     """
     if not isinstance(input, cupy.ndarray):
         raise TypeError('input must be cupy.ndarray')
-    if input.dtype.char in 'FD':
-        raise TypeError('Complex type not supported')
+    # There is constraints on types because atomicAdd() in CUDA 7.5
+    # only supports int32, uint32, uint64, and float32.
+    if not issubclass(input.dtype.type,
+                      (cupy.int32, cupy.float16, cupy.float32,
+                       cupy.float64, cupy.uint32, cupy.uint64,
+                       cupy.ulonglong)):
+        raise TypeError(
+            'cupyx.scipy.ndimage.mean only supports int32, float16, '
+            'float32, float64, uint32, uint64, as data type')
 
     if labels is None:
         return input.sum()/input.size
@@ -365,12 +404,35 @@ def mean(input, labels=None, index=None):
         raise TypeError('label must be cupy.ndarray')
 
     if index is None:
-        index = cupy.unique(labels)
+        return (input[labels != 0]).mean()
 
     input, labels = cupy.broadcast_arrays(input, labels)
 
     if not isinstance(index, cupy.ndarray):
-        raise TypeError('index must be cupy.ndarray')
+        if not isinstance(index, int):
+            raise TypeError('index must be cupy.ndarray or a scalar int')
+        else:
+            index = cupy.asarray(index)
 
     sum, count = _stats(input, labels, index, True)
     return sum/count
+
+
+def standard_deviation(input, labels=None, index=None):
+    """Calculate the standard deviation of the values of an n-D image array,
+    optionally at specified sub-regions.
+
+    Args:
+        input (cupy.ndarray) :Nd-image data to process.
+        labels (cupy.ndarray or None): Labels defining sub-regions in `input`.
+            If not None, must be same shape as `input`.
+        index (cupy.ndarray or None): `labels` to include in output.  If None
+            (default), all values where `labels` is non-zero are used.
+
+    Returns:
+        standard_deviation (cupy.ndarray): standard deviation of values, for
+        each sub-region if `labels` and `index` are specified.
+
+    .. seealso:: :func:`scipy.ndimage.standard_deviation`
+    """
+    return cupy.sqrt(variance(input, labels, index))
--- a/cupyx/scipy/ndimage/measurements.py
+++ b/cupyx/scipy/ndimage/measurements.py
@@ -307,14 +307,24 @@ def variance(input, labels=None, index=None):
     if not isinstance(input, cupy.ndarray):
         raise TypeError('input must be cupy.ndarray')
 
+    if issubclass(input.dtype.type, (cupy.bool, cupy.complex64,
+                                     cupy.complex128)):
+        raise TypeError("cupyx.scipy.ndimage.variance doesn't support %{}"
+                        "".format(input.dtype.type))
+
+    use_kern = False
     # There is constraints on types because atomicAdd() in CUDA 7.5
     # only supports int32, uint32, uint64, and float32.
     if not issubclass(input.dtype.type,
-                      (cupy.int32, cupy.float16, cupy.float32, cupy.float64,
-                       cupy.uint32, cupy.uint64, cupy.ulonglong)):
-        raise TypeError(
-            'cupyx.scipy.ndimage.variance only supports int32, float16, '
-            'float32, float64, uint32, uint64, as data type')
+                      (cupy.int32, cupy.float16, cupy.float32,
+                       cupy.float64, cupy.uint32, cupy.uint64,
+                       cupy.ulonglong)):
+        warnings.warn(
+            'Using the slower implmentation as '
+            'cupyx.scipy.ndimage.sum supports int32, float16, '
+            'float32, float64, uint32, uint64 as data types'
+            'for the fast implmentation', util.PerformanceWarning)
+        use_kern = True
 
     if labels is None:
         return input.var()
@@ -333,7 +343,11 @@ def variance(input, labels=None, index=None):
         else:
             return (input[labels == index]).var()
 
-    mean_val, count = _mean_driver(input, labels, index, True)
+    mean_val, count = _mean_driver(input, labels, index, True, use_kern)
+    if use_kern:
+        return cupy.where(labels == index[..., None],
+                          cupy.square(input - mean_val[..., None]),
+                          0).sum(axis=(1,)) / count
     out = cupy.zeros_like(index, dtype=mean_val.dtype)
     return _ndimage_variance_kernel(input, labels, index, index.size, mean_val,
                                     out)/count
--- a/cupyx/scipy/ndimage/measurements.py
+++ b/cupyx/scipy/ndimage/measurements.py
@@ -10,15 +10,15 @@ def label(input, structure=None, output=None):
 
     Args:
         input (cupy.ndarray): The input array.
-        structure (array_like or None): A structuring element that defines
-            feature connections. ```structure``` must be centersymmetric. If
-            None, structure is automatically generated with a squared
+        structure (array_like or None): A structuring element that defines \
+            feature connections. ```structure``` must be centersymmetric. If \
+            None, structure is automatically generated with a squared \
             connectivity equal to one.
-        output (cupy.ndarray, dtype or None): The array in which to place the
+        output (cupy.ndarray, dtype or None): The array in which to place the \
             output.
 
     Returns:
-        label (cupy.ndarray): An integer array where each unique feature in
+        label (cupy.ndarray): An integer array where each unique feature in \
             ```input``` has a unique label in the array.
         num_features (int): Number of features found.
 
@@ -252,8 +252,8 @@ def _stats(input, labels=None, index=None, mean=False):
 
 
 def variance(input, labels=None, index=None):
-    """Calculate the variance of the values of an n-D image array, optionally
-       at specified sub-regions.
+    """Calculate the variance of the values of an n-D image array, optionally \
+    at specified sub-regions.
 
     Args:
         input (cupy.ndarray) :Nd-image data to process.
@@ -263,7 +263,7 @@ def variance(input, labels=None, index=None):
             (default), all values where `labels` is non-zero are used.
 
     Returns:
-       variance (cupy.ndarray): Values of variance, for each sub-region if
+       variance (cupy.ndarray): Values of variance, for each sub-region if \
             `labels` and `index` are specified.
 
 
@@ -275,7 +275,7 @@ def variance(input, labels=None, index=None):
         raise TypeError('Complex type not supported')
 
     if labels is None:
-        return input.sum()
+        return input.var()
 
     if not isinstance(labels, cupy.ndarray):
         raise TypeError('label must be cupy.ndarray')
@@ -291,11 +291,12 @@ def variance(input, labels=None, index=None):
     sum, count = _stats(input, labels, index, mean=True)
     mean = sum/count
     ret = cupy.zeros_like(index).astype(mean.dtype)
-    return _variance_kernel()(input, labels, index, index.size, mean, ret)
+    return _variance_kernel()(input, labels, index,
+                              index.size, mean, ret)/count
 
 
 def sum(input, labels=None, index=None):
-    """Calculate the sum of the values of an n-D image array, optionally
+    """Calculate the sum of the values of an n-D image array, optionally \
        at specified sub-regions.
 
     Args:
@@ -306,7 +307,7 @@ def sum(input, labels=None, index=None):
             (default), all values where `labels` is non-zero are used.
 
     Returns:
-       sum (cupy.ndarray): sum of values, for each sub-region if
+       sum (cupy.ndarray): sum of values, for each sub-region if \
             `labels` and `index` are specified.
 
 
@@ -335,7 +336,7 @@ def sum(input, labels=None, index=None):
 
 
 def mean(input, labels=None, index=None):
-    """Calculate the mean of the values of an n-D image array, optionally
+    """Calculate the mean of the values of an n-D image array, optionally \
        at specified sub-regions.
 
     Args:
@@ -346,7 +347,7 @@ def mean(input, labels=None, index=None):
             (default), all values where `labels` is non-zero are used.
 
     Returns:
-        mean (cupy.ndarray): mean of values, for each sub-region if
+        mean (cupy.ndarray): mean of values, for each sub-region if \
             `labels` and `index` are specified.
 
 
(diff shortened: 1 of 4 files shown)


emcastillo:
Can you post an example of your reduction kernels so we can see why they didn't work?

niteya-shah:
@emcastillo I couldnt write one because the strategy used in cupy Reduction Kernel relies on interleaved addressing strategies to reduce, but ndimage.sum and its variants, the values are to be selectively summed by a label, which cant summed using the reduction kernel, I dont think we can add values to certain location using our current reduction strategy.( Please correct me If i am wrong anywhere)

I tried to split the difference where in the inital step I create multiple boxes of index shape and fill them according to a split in the labels and then sum them, as a way to reduce the serialization of atomicAdd. A implementation would go like this
```

def sum(input, labels=None, index=None):

    input_arguments = 'T image, R label, raw X index, int32 size, int32 value'
    _kernel_template = string.Template("""
    for (ptrdiff_t j = 0; j < size; j++) {
      if (label == index[j]) {
        int block = i/value;
        atomicAdd(&ret[j + block * size], image);
        break;
      }
    }
    """)
    output_arguments = 'raw Y ret'
    _kernel_template = _kernel_template.substitute(mean="")
    boxes = 100
    ret = cupy.zeros((index.size * boxes), dtype=input.dtype)
    value = image.size/boxes
    _kernel = cupy.ElementwiseKernel(input_arguments, output_arguments,
                                     _kernel_template, "sumkernel")

    result = _kernel(input, labels, index, index.size, value, ret)
    return result.reshape(boxes, index.size).sum(axis = 0)
 
```

However even this method was too slow compared to ndimage.sum using numpy.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -2,21 +2,23 @@
...
 
 def label(input, structure=None, output=None):
     """Labels features in an array
 
     Args:
         input (cupy.ndarray): The input array.
-        structure (array_like or None): A structuring element that defines
-            feature connections. ```structure``` must be centersymmetric. If
-            None, structure is automatically generated with a squared
+        structure (array_like or None): A structuring element that defines \
```
emcastillo, line 13: Remove the `\`

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +202,175 @@ def _kernel_finalize():
         y[i] = j + 1;
         ''',
         'cupyx_nd_label_finalize')
+
+
+def _variance_kernel():
```
emcastillo, line 207: memoize this function

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +202,175 @@ def _kernel_finalize():
...
+    }
+    """
+    return cupy.ElementwiseKernel(input_arguments, output_arguments,
+                                  _kernel_template)
+
+
+def _stats(input, labels=None, index=None, mean=False):
+
+    input_arguments = 'T image, R label, raw X index, int32 size'
+    _kernel_template = string.Template("""
```
emcastillo, line 225: Isolate the kernel creation and memoize it

cupyx/scipy/ndimage/measurements.py
niteya-shah: Done

cupyx/scipy/ndimage/measurements.py
niteya-shah: Done

cupyx/scipy/ndimage/measurements.py
niteya-shah: Done

niteya-shah:
@emcastillo I added tests and tried to improve the kernel but I think this task is largely serial and I wasn't able to improve it significantly. 
I also added stge function standard deviation as it was a simple sqrt of variance.

takagi:
As for reduction algorithm, the following code does computation equivalent to `scipy.ndimage.sum`:

```Python
import numpy
import cupy
image = cupy.arange(100, dtype=numpy.float32)
labels = cupy.random.choice(3, 100) + 1
index = cupy.array([1, 2, 3])
cupy.where(labels == index[..., None], image, 0).sum(axis=(1,))
```

Additionally, using `cupy.fuse` may combine `cupy.where` and `cupy.sum` to reduce the number of kernel invocation times to just one.

grlee77:
@takagi
That is a neat, concise implementation. I think it is a reasonable choice when the number of labels to measure is small, but it is not memory efficient when there are a larger number of labels. (the size of some temporary arrays in the input/output of `where` are `image.size * len(index)`)

I tested the following n-d generalization:
```Python
def sum_v2(image, labels, index):
    matched = labels == index.reshape((-1,) + (1,) * image.ndim)
    sum_axes = tuple(range(1, 1 + image.ndim))
    return cupy.where(matched, image, 0).sum(axis=sum_axes, dtype=cupy.intp)
```

For shape (256, 256, 256), it was indeed faster for 2, 4 or 8 labels, but became substantially slower for larger numbers of labels.

labels | sum_v2 | This PR
-----------|-----------|-----------
2 | 14.0 (ms) | 27.0 (ms)
4 | 16.1 (ms) | 27.0 (ms)
8 | 20.2 (ms) | 27.0 (ms)
16 | 28.3 (ms) | 19.6 (ms)
32 | 46.2 (ms) | 15.1 (ms)
64 | 83.5 (ms) | 20.5 (ms)

I didn't try it, but I suppose one could run `sum_v2` in batches of a few labels at a time rather than on the full set at once to help avoid excessive memory overhead, but I suspect the performance may still be worse for large numbers of labels.

takagi:
@grlee77 Thanks for pointing it out and it makes sense. I tried to implement the same thing with `ReductionKernel` and `cupy.fuse` to intend eliminating the use of the temporary array but both made worse performance even with lesser numbers of labels.

BTW, I made a similar performance test as @grlee77 's including `scipy`.

|labels|sum_v2|This PR|scipy|
|---|---|---|---|
| 2 |17.2(ms) |39.6(ms) |140.0(ms) |
| 4 |18.4(ms) |39.5(ms) |132.9(ms) |
| 8 |21.1(ms) |39.5(ms) |132.1(ms) |
| 16 |26.3(ms) |28.9(ms) |132.4(ms) |
| 32 |36.3(ms) |22.3(ms) |132.1(ms) |
| 64 |56.5(ms) |18.4(ms) |131.9(ms) |

niteya-shah:
Thank you so much @takagi  and @grlee77 for your help and insight. I ran the tests myself using sum_v2  and ran out of memory as predicted. I then implemented the batched version as suggested by @grlee77 .   

```python
def sum_v3(image, label, index, batch_size = 4):
    out = cupy.empty_like(index, dtype=image.dtype)
    for i in range(0, index.size, batch_size):
        matched = label == index[i:i + batch_size].reshape((-1,) + (1,) * image.ndim)
        sum_axes = tuple(range(1, 1 + image.ndim))
        out[i:i + batch_size] = cupy.where(matched, image, 0).sum(axis=sum_axes)
    return out 
```

(256, 256, 256),  batch_size = 4

| label | scipy(ms)  | This PR(ms)  | sum_v2(ms) | sum_v3(ms) |
|-------|--------|----------|--------|--------|
| 2     | 103.15 | 18363.52 | 33.82  | 32.58 |
| 4     | 90.99  | 6955.65  | 45.58  | 44.31 |
| 8     | 89.80  | 3189.20  | 69.34  |  88.50 |
| 16   | 89.37  | 1264.81  | |  176.50 |
| 32   | 90.21  |  535.65   | |  352.76 |
| 64   | 89.38  |  172.86   | |  705.50  |
| 128 | 90.91  | 85.94      | |  1410.12 |

(For reference my GPU is a 970MX)

Since I cannot compute the values for sum_v2 for larger label sizes, someone else could perhaps try and compare the performance difference that arises due to batching it.

takagi:
`atomicAdd` might be badly expensive on 970MX? According to the results on @grlee77 's and my environments, the implementation of this PR might be acceptable.  And a possible improvement may be that a thread computes multiple elements, for example 16 elements, sequentially on registers to get subtotals for each index, aggregating the subtotals into the output array with `atomicAdd`. Let me ask for the core members.

takagi:
We talked about this PR and it would be reasonable to use the both algorithms, elementwise one and reduction, switching them by the number of labels empirically given.  How do you think?

niteya-shah:
I think that would be the best solution too, Since my GPU isn't very good with atomicAdd, I started testing my work on Google [Colab](https://colab.research.google.com/drive/1X_o0WsuTmNR6tGX-MxNhlb-migSyyqnk).
The results for varying sizes and labels are as follows
size : (16, 16, 16)
label_size | This PR |  scipy |  sum_v2 | sum_v3 | sum_v4
--- | --- | --- | --- | --- | ---
2 | 0.05 | 0.07 | 0.32 | 0.35 | 0.09 | 
4 | 0.05 | 0.05 | 0.24 | 0.31 | 0.09 | 
8 | 0.05 | 0.05 | 0.35 | 0.59 | 0.09 | 
16 | 0.06 | 0.06 | 0.24 | 1.11 | 0.10 | 
32 | 0.05 | 0.05 | 0.34 | 2.25 | 0.09 | 
64 | 0.05 | 0.05 | 0.25 | 4.36 | 0.09 | 


size : (32, 32, 32)
label_size | This PR |  scipy |  sum_v2 | sum_v3 | sum_v4
--- | --- | --- | --- | --- | ---
2 | 0.12 | 0.23 | 0.27 | 0.32 | 0.09 | 
4 | 0.12 | 0.20 | 0.27 | 0.32 | 0.09 | 
8 | 0.12 | 0.22 | 0.27 | 0.59 | 0.09 | 
16 | 0.11 | 0.19 | 0.28 | 1.12 | 0.09 | 
32 | 0.09 | 0.26 | 0.31 | 2.32 | 0.10 | 
64 | 0.09 | 0.20 | 0.44 | 4.33 | 0.11 | 


size : (64, 64, 64)
label_size | This PR |  scipy |  sum_v2 | sum_v3 | sum_v4
--- | --- | --- | --- | --- | ---
2 | 0.67 | 1.61 | 0.57 | 0.63 | 0.36 | 
4 | 0.67 | 1.56 | 0.65 | 0.65 | 0.25 | 
8 | 0.56 | 1.34 | 0.67 | 1.22 | 0.20 | 
16 | 0.49 | 1.38 | 0.87 | 2.44 | 0.20 | 
32 | 0.40 | 1.38 | 1.22 | 4.61 | 0.25 | 
64 | 0.26 | 1.40 | 2.04 | 9.20 | 0.33 | 


size : (128, 128, 128)
label_size | This PR |  scipy |  sum_v2 | sum_v3 | sum_v4
--- | --- | --- | --- | --- | ---
2 | 4.99 | 13.52 | 3.12 | 3.17 | 4.38 | 
4 | 4.99 | 11.20 | 3.47 | 3.52 | 3.63 | 
8 | 3.82 | 10.71 | 4.21 | 6.97 | 2.96 | 
16 | 2.99 | 10.81 | 5.61 | 13.84 | 2.31 | 
32 | 2.36 | 10.86 | 8.53 | 27.55 | 1.64 | 
64 | 1.47 | 10.81 | 15.01 | 54.99 | 2.06 | 


size : (256, 256, 256)
label_size | This PR |  scipy |  sum_v2 | sum_v3 | sum_v4
--- | --- | --- | --- | --- | ---
2 | 39.50 | 123.68 | 23.68 | 23.57 | 38.89 | 
4 | 39.49 | 116.53 | 26.50 | 26.36 | 38.15 | 
8 | 28.77 | 115.36 | 32.17 | 52.62 | 28.21 | 
16 | 22.00 | 115.15 | 43.61 | 105.14 | 21.39 | 
32 | 17.87 | 115.07 | 66.90 | 209.94 | 17.31 | 
64 | 11.00 | 114.70 | 118.48 | 419.40 | 15.63 | 

In my opinium sum_v2 requires a lot of memory, so we should only consider sum_v, sum_v4 which is a binned method that I proposed earlier doesent seem to outperform the current method.
There is consistency of results for different sizes and hence the size of the image can be ignored.
In my opinion, for label sizes less than 4, we use sum_v3 and for greater than that we use this PR(sum_v1).
Please do correct me if I am wrong anywhere as I have very little experience with heuristic solutions.

takagi:
Thanks for the detail investigation on colab, and switching the algorithms with the label size of 4 sounds good to me! The best candidate of the label size would be 4 or 8, which may vary on environments. I respect your opinion with Tesla P100. Then, would you apply this result to the code of this PR?

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,282 @@ def _kernel_finalize():
...
+        mean_axes = tuple(range(1, 1 + input.ndim))
+        count[i:i + batch_size] = matched.sum(axis=mean_axes)
+        out[i:i + batch_size] = cupy.where(matched, input, 0).sum(
+            axis=mean_axes)
+    if return_count:
+        return out/count, count
+    return out/count
+
+
+def _mean_driver(input, labels, index, return_count=False, use_kern=False):
```
niteya-shah, line 274: After performance testing this kernel it lagged behind the Elementwise Kernel implmentation for all cases,yet I want to keep it because it permits all the all the dtypes that are not supported by the Elementwise kernel because of restrictions in atomicAdd.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,282 @@ def _kernel_finalize():
...
+    """
+    if not isinstance(input, cupy.ndarray):
+        raise TypeError('input must be cupy.ndarray')
+
+    # There is constraints on types because atomicAdd() in CUDA 7.5
+    # only supports int32, uint32, uint64, and float32.
+    if not issubclass(input.dtype.type,
+                      (cupy.int32, cupy.float16, cupy.float32, cupy.float64,
+                       cupy.uint32, cupy.uint64, cupy.ulonglong)):
+        raise TypeError(
```
niteya-shah, line 315: Some help in this would be appriciated as I could not come up with a implmentation of variance that used matching that was clean and didn't require me to completely re-implmenent a new version of _ndarray_var where I could set alpha.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,282 @@ def _kernel_finalize():
...
+            input.dtype.type))
+
+    use_kern = False
+    # There is constraints on types because atomicAdd() in CUDA 7.5
+    # only supports int32, uint32, uint64, and float32.
+    if not issubclass(input.dtype.type,
+                      (cupy.int32, cupy.float16, cupy.float32,
+                       cupy.float64, cupy.uint32, cupy.uint64,
+                       cupy.ulonglong)):
+        warnings.warn(
```
niteya-shah, line 375: Because of the matching kernel, sum and mean both now support all dtypes other than complex64, complex128 and bool because scipy doesnt support them. If a operation uses this without the performance threshold a warning is thrown currently(should it ? ).

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,282 @@ def _kernel_finalize():
...
+    for (ptrdiff_t j = 0; j < size; j++) {
+      if (labels == index[j]) {
+        atomicAdd(&out[j], input);
+        break;
+      }
+    }
+    """)
+
+
+def _ndimage_sum_kernel_2(input, labels, index, out, batch_size=4):
```
niteya-shah, line 234: Should I use the un-batched version here as I dont think people will change internal mechanics of cupy and the externally for the unbatched version outperformed it? ( I kept it there to ensure reduced memory usage but for the current  results, the unbatched version doesnt seem to provide any benefits.)

takagi: Yes, I suppose the un-batched version is simpler and enough here. Others can follow the discussion on this PR later.

takagi: Yes, I suppose the un-batched version is simpler and enough here. Others can follow the discussion on this PR later.

niteya-shah:
@takagi any update on this?

cupyx/scipy/ndimage/measurements.py
takagi: I think it is fine to throw this warning as users would think they change the dtype or not.

cupyx/scipy/ndimage/measurements.py
takagi: Maybe the batched version is helpful in case `use_kern=True`?

cupyx/scipy/ndimage/measurements.py
takagi: Like this for 1D?
```Python
mean_val, count = _mean_driver(input, labels, index, True)
cupy.where(labels == index[..., None], cupy.square(input - mean_val[..., None]), 0).sum(axis=(1,)) / count
```

Maybe we can use `ravel()` for inputs more than 1D.

takagi:
Sorry for being late. I left some comments and will review further.

cupyx/scipy/ndimage/measurements.py
niteya-shah: I dont think this can work, for e.g. the input can be of shape of 256, but the mean_val is going to be of shape index, and it makes no sense for them to be able to be broadcastable together or the subtraction to be logically correct.

cupyx/scipy/ndimage/measurements.py
niteya-shah: yeah I kinda forgot about the memory issues of the unbatched version when people have to relly on use_kern=True, the batched vesion should stay as agreed upon before.

cupyx/scipy/ndimage/measurements.py
niteya-shah: The ravel wasn't necessary as it was able to internally broadcast them properly.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+
+    input, labels = cupy.broadcast_arrays(input, labels)
+
+    if not isinstance(index, cupy.ndarray):
+        if not isinstance(index, int):
+            raise TypeError('index must be cupy.ndarray or a scalar int')
+        else:
+            return (input[labels == index]).sum()
+
+    out = cupy.zeros_like(index, dtype=input.dtype)
```
takagi, line 414: What dtype of data does `scipy.ndimage.sum` return becase if, for example, `input.dtype` is `numpy.int8`, `out` would easily overflow? FYI, see: https://github.com/cupy/cupy/blob/f7fc9b9df43ec31918f90b1a34747d0630b531dc/cupy/core/_routines_math.pyx#L545

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+            (-1,) + (1,) * input.ndim)
+        sum_axes = tuple(range(1, 1 + input.ndim))
+        out[i:i + batch_size] = cupy.where(matched, input, 0).sum(
+            axis=sum_axes)
+    return out
+
+
+_ndimage_mean_kernel = cupy.ElementwiseKernel(
+    'T input, R labels, raw X index, int32 size',
+    'raw Y out, raw int32 count',
```
takagi, line 246: `int32` may be too small when, for example, computing an input with its shape `(4096, 4096, 4096)`.
```suggestion
    'raw Y out, raw uint64 count',
```

cupyx/scipy/ndimage/measurements.py
niteya-shah: Yeah, it seems the that scipy.ndimage.sum returns only int64 and float64, I will make the necessary changes to only return those types.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+        atomicAdd(&count[j], 1);
+        break;
+      }
+    }
+    """)
+
+
+def _ndimage_mean_kernel_2(input, labels, index, batch_size=4,
+                           return_count=False):
+    out = cupy.empty_like(index, dtype=input.dtype)
```
takagi, line 260: I suppose `out` would be expected to have floating point dtype even when `input`'s dtype is integral.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+    count = cupy.empty_like(index, dtype=cupy.int32)
```
takagi, line 261: How about using larger dtype for `count`?
```suggestion
    count = cupy.empty_like(index, dtype=cupy.uint64)
```

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+
+    mean_val, count = _mean_driver(input, labels, index, True, use_kern)
+    if use_kern:
+        new_axis = (..., *(cupy.newaxis for _ in range(input.ndim)))
+        return cupy.where(labels[None, ...] == index[new_axis],
+                          cupy.square(input - mean_val[new_axis]),
+                          0).sum(tuple(range(1, input.ndim + 1))) / count
+    out = cupy.zeros_like(index, dtype=mean_val.dtype)
+    return _ndimage_variance_kernel(input, labels, index, index.size, mean_val,
+                                    out)/count
```
takagi, line 354: nit: Just for the style guideline.
```suggestion
                                    out) / count
```

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+
+
+def sum(input, labels=None, index=None):
+    """Calculate the sum of the values of an n-D image array, optionally
```
takagi, line 358: ditto.
```suggestion
    """Calculates the sum of the values of an n-D image array, optionally
```

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+       at specified sub-regions.
+
+    Args:
+        input (cupy.ndarray) :Nd-image data to process.
```
takagi, line 362:
```suggestion
        input (cupy.ndarray): Nd-image data to process.
```

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+        labels (cupy.ndarray or None): Labels defining sub-regions in `input`.
+            If not None, must be same shape as `input`.
+        index (cupy.ndarray or None): `labels` to include in output.  If None
+            (default), all values where `labels` is non-zero are used.
+
+    Returns:
+       sum (cupy.ndarray): sum of values, for each sub-region if
+       `labels` and `index` are specified.
+
```
takagi, line 371:
```suggestion
```

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+
+    .. seealso:: :func:`scipy.ndimage.sum`
+    """
+    if not isinstance(input, cupy.ndarray):
+        raise TypeError('input must be cupy.ndarray')
+
+    if issubclass(input.dtype.type, (cupy.bool, cupy.complex64,
```
takagi, line 378: ditto (`in` operator is enough).

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+                                     cupy.complex128)):
+        raise TypeError("cupyx.scipy.ndimage.sum doesnt support %{}".format(
+            input.dtype.type))
+
+    use_kern = False
+    # There is constraints on types because atomicAdd() in CUDA 7.5
+    # only supports int32, uint32, uint64, and float32.
```
takagi, line 385: ditto (comment in general would be enough).

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+    if not issubclass(input.dtype.type,
```
takagi, line 386: ditto ('in' operator).

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+        use_kern = True
+
+    if labels is None:
+        return input.sum()
+
+    if not isinstance(labels, cupy.ndarray):
+        raise TypeError('label must be cupy.ndarray')
+
+    if index is None:
+        return input.sum()
```
takagi, line 404: `(input[labels != 0]).sum()`?

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,332 @@ def test_label_swirl(self, xp, scp):
...
         return labels
+
+
+@testing.gpu
+@testing.with_requires('scipy')
+class TestNdimageSum(unittest.TestCase):
+
+    @testing.for_dtypes([cupy.int32, cupy.float32, cupy.float64, cupy.uint32,
+                         cupy.uint64, cupy.ulonglong, cupy.int64, cupy.uint8,
+                         cupy.uint16])
```
takagi, line 109:
```suggestion
    @testing.for_all_dtypes(no_bool=True, no_complex=True)
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,332 @@ def test_label_swirl(self, xp, scp):
...
+    @testing.numpy_cupy_array_almost_equal(scipy_name='scp')
```
takagi, line 110: You may use `@testing.numpy_cupy_array_equal` after changing the output dtype of `sum`, removing `astype` in the return statement.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -200,3 +203,297 @@ def _kernel_finalize():
...
+        mean (cupy.ndarray): mean of values, for each sub-region if
+        `labels` and `index` are specified.
+
+
+    .. seealso:: :func:`scipy.ndimage.mean`
+    """
+    if not isinstance(input, cupy.ndarray):
+        raise TypeError('input must be cupy.ndarray')
+
+    if issubclass(input.dtype.type, (cupy.bool, cupy.complex64,
```
takagi, line 443: ditto (`in` operator and `cupy.bool_`)

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,332 @@ def test_label_swirl(self, xp, scp):
...
+    def test_ndimage_sum_scalar_index(self, xp, scp, dtype):
+        image = xp.arange(100, dtype=dtype)
+        label = testing.shaped_random((100,), xp, dtype=xp.int32, scale=3) + 1
+        return scp.ndimage.sum(image, label, 1).astype(dtype)
+
+    @testing.for_dtypes([cupy.complex64, cupy.complex128])
```
takagi, line 143: Would you add `cupy.bool_` to the test?

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,332 @@ def test_label_swirl(self, xp, scp):
...
+    def test_ndimage_sum_wrong_dtype(self, dtype):
+        image = cupy.arange(100, dtype=dtype)
+        label = cupy.random.randint(1, 3, dtype=cupy.int32)
```
takagi, line 146: This doesn't draw three.
```suggestion
        label = cupy.random.randint(1, 4, dtype=cupy.int32)
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,332 @@ def test_label_swirl(self, xp, scp):
...
+        index = cupy.array([1, 2, 3])
+        with pytest.raises(TypeError):
+            cupyx.scipy.ndimage.sum(image, label, index)
+
+    def test_ndimage_sum_wrong_label_size(self):
```
takagi, line 151:
```suggestion
    def test_ndimage_sum_wrong_label_shape(self):
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,332 @@ def test_label_swirl(self, xp, scp):
...
+            cupyx.scipy.ndimage.sum(image, label, index)
+
+    def test_ndimage_sum_wrong_index_type(self):
+        image = cupy.arange(100, dtype=cupy.int32)
+        label = cupy.random.randint(1, 3, dtype=cupy.int32, size=100)
+        index = [1, 2, 3]
+        with pytest.raises(TypeError):
+            cupyx.scipy.ndimage.sum(image, label, index)
+
+    @testing.numpy_cupy_array_almost_equal(scipy_name='scp')
```
takagi, line 179:
```suggestion
    @testing.numpy_cupy_array_equal(scipy_name='scp')
```

takagi:
Would you also add tests for inputs that have dimensions more than one?

niteya-shah:
@takagi sorry for being late, I have made the changes, can you check them?

takagi:
Thanks, let me see the changes shortly!

cupyx/scipy/ndimage/measurements.py
```diff
@@ -219,8 +219,8 @@ def _kernel_finalize():
 
 
 _ndimage_sum_kernel = cupy.ElementwiseKernel(
-    'T input, R labels, raw X index, int32 size',
-    'raw Y out',
+    'T input, R labels, raw X index, uint64 size',
+    'raw float64 out',
```
takagi, line 223: Would you use the output dtype of `uint64` for integer inputs, as well as the same dtype with `input` for floating-point inputs as `scipy.ndimage.sum` does so.

niteya-shah: So i just want to confirm what I understand from this, it should be something like
```python
if issubtype(input.dtype, cupy.integer):
    out_type = cupy.uint64
else:
    out_type = input.dtype
```
because the following code 
```python
import numpy as np
import scipy.ndimage as ndn
input = np.array([1, 2, 3])
labels = np.array([1, 1, 0])
index = np.array([1, 0])
assert(input.dtype == np.int64)
assert(labels.dtype == np.int64)
assert(labels.dtype == np.int64)
out = ndn.sum(input, labels, index)
print(out.dtype)

```
gives

> float64
And from what I understood atleast this will be non-compatible. Please correct me if I misunderstood your request.

takagi: Thanks to confirm. You're right, and what I saw might be the output dtype of `ndn.sum(input)`. Sorry, you don't need to make any change.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -206,8 +206,8 @@ def _kernel_finalize():
 
 
 _ndimage_variance_kernel = cupy.ElementwiseKernel(
-    'T input, R labels, raw X index, int32 size, raw Z mean',
-    'raw Z out',
+    'T input, R labels, raw X index, uint64 size, raw float64 mean',
```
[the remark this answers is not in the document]
takagi: Let me drop this request.

cupyx/scipy/ndimage/measurements.py
```diff
@@ -219,8 +219,8 @@ def _kernel_finalize():
 
 
 _ndimage_sum_kernel = cupy.ElementwiseKernel(
-    'T input, R labels, raw X index, int32 size',
-    'raw Y out',
+    'T input, R labels, raw X index, uint64 size',
```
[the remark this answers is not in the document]
takagi: Let me drop this request.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,339 @@ def test_label_swirl(self, xp, scp):
...
+    @testing.numpy_cupy_array_equal(scipy_name='scp')
+    def test_ndimage_sum_scalar_index(self, xp, scp, dtype):
+        image = xp.arange(100, dtype=dtype)
+        label = testing.shaped_random((100,), xp, dtype=xp.int32, scale=3) + 1
+        return scp.ndimage.sum(image, label, 1)
+
+    @testing.for_dtypes([cupy.bool_, cupy.complex64, cupy.complex128])
+    def test_ndimage_sum_wrong_dtype(self, dtype):
+        image = cupy.arange(100).astype(dtype)
+        label = cupy.random.randint(1, 4, dtype=cupy.int32)
```
takagi, line 146: Please use a valid shape for `label` even if it is not used in this test.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,339 @@ def test_label_swirl(self, xp, scp):
...
+
+    @testing.for_all_dtypes(no_bool=True, no_complex=True)
+    @testing.numpy_cupy_array_equal(scipy_name='scp')
+    def test_ndimage_sum_scalar_index(self, xp, scp, dtype):
+        image = xp.arange(100, dtype=dtype)
+        label = testing.shaped_random((100,), xp, dtype=xp.int32, scale=3) + 1
+        return scp.ndimage.sum(image, label, 1)
+
+    @testing.for_dtypes([cupy.bool_, cupy.complex64, cupy.complex128])
+    def test_ndimage_sum_wrong_dtype(self, dtype):
```
takagi, line 144:
```suggestion
    def test_ndimage_sum_wrong_image_dtype(self, dtype):
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,339 @@ def test_label_swirl(self, xp, scp):
...
+
+    @testing.for_all_dtypes(no_bool=True, no_complex=True)
+    @testing.numpy_cupy_array_almost_equal(scipy_name='scp')
+    def test_ndimage_variance_multi_dim(self, xp, scp, dtype):
+        image = xp.arange(512, dtype=dtype).reshape(8, 8, 8)
+        label = testing.shaped_random((8, 8, 8), xp, dtype=xp.int32, scale=3)
+        index = xp.array([0, 1, 2])
+        return scp.ndimage.variance(image, label, index)
+
+    @testing.for_all_dtypes(no_bool=True, no_complex=True, no_float16=True)
```
takagi, line 207: Can't we test `float16` inputs?

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -108,16 +108,16 @@ class TestNdimageSum(unittest.TestCase):
     @testing.numpy_cupy_array_equal(scipy_name='scp')
     def test_ndimage_sum(self, xp, scp, dtype):
         image = xp.arange(100, dtype=dtype)
-        label = testing.shaped_random((100,), xp, dtype=xp.int32, scale=3) + 1
-        index = xp.array([1, 2, 3])
+        label = testing.shaped_random((100,), xp, dtype=xp.int32, scale=4)
+        index = xp.array([0, 1, 2, 3])
```
takagi, line 112: Please remove 0 from `index`

takagi:
Would you parameterize the ndimage operations using `@testing.parameterize` so that we don't need to repeat similar hand-written tests? You can get the operations by `getattr(scp.ndimage, op)` where `scp` is a module name and `op` a string like `'sum'` or `'variance'`.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
niteya-shah: This throws an error because of difference in accuracy.

takagi: Thanks, it would be fine, or how about using smaller values for `float16` as:
```Python
if dtype == cupy.float16:
    image = testing.shaped_random((100,), xp, dtype=cupy.float16)
else:
    image = xp.arange(100, dtype=dtype)
```

niteya-shah: Thanks for the suggestion, unfortunately for variance and standard deviation, it fails. I tried to even reduce the number of elements for both arrange and random but it seems that both  lead to decimal differences at the 2nd decimal point. I dont think checking at 2nd decimal point is a good idea so perhaps we could make a test group for only float16 with looser checking?
Do you have any better suggestion for dealing with this ?

takagi: I guess intermediate value is computed in `float16`. Would you make it done in `float64`, not just casting final output values?

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -96,3 +98,89 @@ def test_label_swirl(self, xp, scp):
         x = xp.array(x)
         labels, num_features = scp.ndimage.label(x)
         return labels
+
+
+@testing.gpu
+@testing.parameterize({'op': 'sum'}, {'op': 'mean'}, {'op': 'variance'},
+                      {'op': 'standard_deviation'})
```
takagi, line 105: You can use `testing.product` helper.
```suggestion
@testing.parameterize(*testing.product({
    'op': ['sum', 'mean', 'variance', 'standard_deviation'],
}))
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
niteya-shah: I did some testing and got the following results,
```python
from scipy.ndimage import mean as mean_n
import numpy as np
import cupy as cp

arr = np.asarray(np.arange(100)/2, dtype = np.float16)

print("simple no casted mean ",arr.mean())
print("scipy ndimage mean",mean_n(arr))
print("pre-casted numpy mean", arr.astype(np.float64).mean())
print("casted after mean", arr.mean().astype(np.float64))
print("cast in mean func",arr.mean(dtype=np.float64))

t = [i/2 for i in list(range(100))]
print("sanity check", sum(t)/len(t))

```

> simple no casted mean  24.75
> scipy ndimage mean 24.76
> pre-casted numpy mean 24.75
> casted after mean 24.75
> cast in mean func 24.75
> sanity check 24.75

The scipy mean failes to accurately calculate values for float16 for such small numbers even till the second decimal place, and since this mean value is propogated to variance and standard deviation, I dont think any test I write will pass unless and until I set a very low threshold. Please tell me how I should proceed.

takagi: Thanks, I think your result makes sense. Hmm, then would you split the test methods for `float16` with supplying low threshold as `test_ndimage_only_input_fp16`? I'll take a look into scipy's implementation, or numpy's `bincount`.

niteya-shah:
@takagi I have added the float16 tests, with the reduced thresholds.

takagi:
pfnCI, test this please.

takagi:
Let me merge this PR to the master. Thanks for your effort! I'll make some cosmetic changes later, which are not essential for your contribution.

PR #3260, opened 2020-04-04
Fix the internal streams in multi-GPU Plan1d 

leofang:
Follow-up of #2644. I overlooked an undocumented CUDA behavior that for `cudaMemcpyAsync`/`cudaMemcpyPeerAsync`, the `stream` argument should reside on the **source** device (see, e.g., [here](https://stackoverflow.com/a/22792494/2344149); I also confirmed it with someone from NVIDIA). If this condition is not met, then CUDA would neglect the input `stream` and instead use its internal stream to do the job (can be checked on nvvp). 

If we don't want to rely on CUDA doing the job for us and prefer to be explicit, it means: 
- for the `'scatter'` operation: all streams should be on the same device as is the source
- for the `'gather'` operation: each device needs one stream. 

This PR addresses this distinction.

### DIFF
--- a/cupy/cuda/cufft.pyx
+++ b/cupy/cuda/cufft.pyx
@@ -203,9 +203,9 @@ cdef _reorder_buffers(Handle plan, intptr_t xtArr, list xtArr_buffer):
     _XtFree(temp_xtArr)
 
 
+# This is meant to replace cufftXtMalloc().
 # We need to manage the buffers ourselves in order to 1. avoid excessive,
 # uncessary memory usage, and 2. use CuPy's memory pool.
-# This is meant to replace cufftXtMalloc().
 cdef _XtMalloc(list gpus, list sizes, XtSubFormat fmt):
     cdef XtArrayDesc* xtArr_desc
     cdef XtArray* xtArr
@@ -213,8 +213,8 @@ cdef _XtMalloc(list gpus, list sizes, XtSubFormat fmt):
     cdef int i, nGPUs
     cdef size_t size
 
-    assert len(gpus) == len(sizes)
     nGPUs = len(gpus)
+    assert nGPUs == len(sizes)
     xtArr_desc = <XtArrayDesc*>PyMem_Malloc(sizeof(XtArrayDesc))
     xtArr = <XtArray*>PyMem_Malloc(sizeof(XtArray))
     c_memset(xtArr_desc, 0, sizeof(XtArrayDesc))
@@ -290,8 +290,8 @@ class Plan1d(object):
             with nogil:
                 result = cufftMakePlan1d(plan, nx, <Type>fft_type, batch,
                                          &work_size)
-
         check_result(result)
+
         work_area = memory.alloc(work_size)
         ptr = work_area.ptr
         with nogil:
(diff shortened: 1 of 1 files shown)


cupy/cuda/cufft.pyx
```diff
@@ -524,27 +552,31 @@ class Plan1d(object):
             share = self.batch_share
 
             if action == 'scatter':
-                if isinstance(a, cupy.ndarray):
-                    for i in range(nGPUs):
-                        count = int(share[i] * self.nx)
+                # if isinstance(b, cupy.ndarray):
```
asi1024, line 555: Please remove this comment and fix indentation.

cupy/cuda/cufft.pyx
```diff
@@ -554,30 +586,29 @@ class Plan1d(object):
                 #             CUFFT_COPY_HOST_TO_DEVICE)
                 #     check_result(result)
             elif action == 'gather':
-                if isinstance(a, cupy.ndarray):
+                # if isinstance(b, cupy.ndarray):
```
asi1024, line 589: Ditto.

cupy/cuda/cufft.pyx
leofang: Done.

cupy/cuda/cufft.pyx
leofang: Done.

asi1024:
Jenkins, test this please.

asi1024:
@leofang Could you check CI failures?

leofang:
Hi @asi1024, I can't reproduce it locally. I suspected it has something to do with recent changes in  the cuFFT module, so I rebased and force-pushed. Could you retry the CIs please?

asi1024:
Jenkins, test this please.

leofang:
@asi1024 Could you retry and see if the same tests continue to fail?

leofang:
I might have some clues. If I'm right, the tests should fail randomly. I need a confirmation.

asi1024:
Jenkins, test this please.

asi1024:
I re-triggered the CIs.
- https://jenkins.preferred.jp/job/chainer/job/cupy_pr/1235/
- https://jenkins.preferred.jp/job/chainer/job/cupy_pr/1245/

leofang:
> I re-triggered the CIs.
> 
> * https://jenkins.preferred.jp/job/chainer/job/cupy_pr/1235/
> * https://jenkins.preferred.jp/job/chainer/job/cupy_pr/1245/

Wow the tests went wild! Thanks for confirming for me, @asi1024. I just pushed a fix. Could you please run the CI twice to ensure no test would fail randomly?

I think the reason is the stream-wait-event should respect the order of participating GPUs.

leofang:
> Could you please run the CI twice to ensure no test would fail randomly?

Ping @asi1024

asi1024:
Jenkins, test this please.

leofang:
Thanks. Let me investigate...😞

leofang:
I am able to reproduce this random failure by stress-testing my machine (I ran two test suites simultaneously). Looks like a synchronization problem?

leofang:
@asi1024 I fixed it.

asi1024:
Jenkins, test this please.

asi1024:
Please let me run the CI twice to ensure no test would fail stochastically.
Jenkins, test this please.

leofang:
Yay!

asi1024:
LGTM!

leofang:
Thanks for help, @asi1024

PR #3261, opened 2020-04-05
implement putmask function

rushabh-v:
I just found one more important function which I accidentally missed to add in my GSoC proposal, So, I have implemented it here.


cupy/manipulation/basic.py
```diff
@@ -69,3 +69,70 @@ def _can_memcpy(dst, src):
...
+    array([  0,   1, -33, -44, -33])
+
+    .. seealso:: :func:`numpy.putmask`
+
+    """
+    if not a.shape == mask.shape:
+        raise ValueError('mask and data must be the same size')
+
+    if a.shape == values.shape:
+        putmask_kernel(mask.astype(a.dtype), values, a)
```
asi1024, line 120:
```suggestion
        a[mask] = value
```

cupy/manipulation/basic.py
```diff
@@ -69,3 +69,70 @@ def _can_memcpy(dst, src):
...
+        if a.ndim > 1:
+            a = a.flatten()
+            mask = mask.flatten()
+        if values.ndim > 1:
+            values = values.flatten()
+
+        len_vals = len(values)
+        n = 0
+
+        for i in range(len(a)):
```
asi1024, line 133: Please parallelize this loop.

tests/cupy_tests/manipulation_tests/test_basic.py
```diff
@@ -125,3 +126,44 @@ def test_copyto_where(self, xp, dtype):
...
         return dst
+
+
+@testing.parameterize(
+    *testing.product(
+        {'shape': [(0,), (1,), (2, 3), (2, 3, 4)]}))
+@testing.gpu
+class TestPutmask(unittest.TestCase):
+
+    @testing.for_all_dtypes(no_complex=True, no_bool=True)
```
asi1024, line 137:
```suggestion
    @testing.for_all_dtypes()
```

tests/cupy_tests/manipulation_tests/test_basic.py
```diff
@@ -125,3 +126,44 @@ def test_copyto_where(self, xp, dtype):
...
+        xp.putmask(a, a > 2, values)
+        return a
+
+
+@testing.gpu
+class TestPutmaskRaises(unittest.TestCase):
+
+    def test_putmask(self):
+        for xp in (numpy, cupy):
+            a = cupy.array([1, 2, 3])
```
asi1024, line 166:
```suggestion
            a = xp.array([1, 2, 3])
```

tests/cupy_tests/manipulation_tests/test_basic.py
```diff
@@ -125,3 +126,44 @@ def test_copyto_where(self, xp, dtype):
...
+
+
+@testing.parameterize(
+    *testing.product(
+        {'shape': [(0,), (1,), (2, 3), (2, 3, 4)],
+         'values_shape': [(2,), (3, 1), (5,)]}))
+@testing.gpu
+class TestPutmaskNonEqual(unittest.TestCase):
+
+    @testing.for_all_dtypes(no_complex=True, no_bool=True)
```
asi1024, line 152:
```suggestion
    @testing.for_all_dtypes()
```

cupy/manipulation/basic.py
rushabh-v: I think this loop can not be parallelized because decisions to be made in every iteration depends on previous iterations. I tried but couldn't figure out a way to make an indexed mask without looping. Do you have any other plans?

cupy/manipulation/basic.py
rushabh-v: Oh, yes that's quite better.

tests/cupy_tests/manipulation_tests/test_basic.py
rushabh-v: The operation(power) used for the array `values` in this test can not be performed on `bool`s. That's why I am going to write a separate test for `bool`s.

tests/cupy_tests/manipulation_tests/test_basic.py
rushabh-v: Oh, I missed it...

tests/cupy_tests/manipulation_tests/test_basic.py
rushabh-v: ditto

cupy/manipulation/basic.py
asi1024: `n` can be calculated by `cupy.cumsum(mask)`

tests/cupy_tests/manipulation_tests/test_basic.py
asi1024: So, do not use power and modulo in this test.

cupy/manipulation/basic.py
```diff
@@ -116,23 +116,21 @@ def putmask(a, mask, values):
...
 
-        len_vals = len(values)
-        n = 0
-
-        for i in range(len(a)):
-            if mask[i]:
-                a[i] = values[n % len_vals]
-                n += 1
-
+        mask = cupy.cumsum(mask.astype(int))
```
asi1024, line 134:
```suggestion
        mask = cupy.cumsum(mask, dtype=numpy.int64)
```

rushabh-v:
I have parallelize that loop, @asi1024 . We don't need to use even `cumsum` here. Actually I misunderstood the numpy's functionality before because of its confusing example in their docstring. I have also updated our docstring's example to not be confusing.

tests/cupy_tests/manipulation_tests/test_basic.py
```diff
@@ -125,3 +126,47 @@ def test_copyto_where(self, xp, dtype):
...
+        {'shape': [(0,), (1,), (2, 3), (2, 3, 4)]}))
+@testing.gpu
+class TestPutmask(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_allclose()
+    def test_putmask(self, xp, dtype):
+        a = testing.shaped_arange(self.shape, xp, dtype=dtype)
+        mask = testing.shaped_random(self.shape, xp, dtype=bool)
+        values = testing.shaped_random(self.shape, xp, dtype=dtype)
```
asi1024, line 142: Feed different seeds.
```suggestion
        a = testing.shaped_arange(self.shape, xp, dtype=dtype, seed=0)
        mask = testing.shaped_random(self.shape, xp, dtype=bool, seed=1)
        values = testing.shaped_random(self.shape, xp, dtype=dtype, seed=2)
```

tests/cupy_tests/manipulation_tests/test_basic.py
```diff
@@ -125,3 +126,47 @@ def test_copyto_where(self, xp, dtype):
...
+        xp.putmask(a, mask, values)
```
asi1024, line 143:
```suggestion
        ret = xp.putmask(a, mask, values)
        assert ret is None
```

cupy/manipulation/basic.py
```diff
@@ -69,3 +69,61 @@ def _can_memcpy(dst, src):
...
+        values = values.astype(a.dtype, casting='safe')
+
+    if a.shape == values.shape:
+        a[mask] = values[mask]
+
+    else:
+        if values.ndim > 1:
+            values = values.flatten()
+
+        putmask_kernel(mask.astype(int), values, len(values), a)
```
asi1024, line 129: Please use numpy dtype for the readability.
```suggestion
        putmask_kernel(mask.astype(numpy.int64), values, len(values), a)
```

cupy/manipulation/basic.py
```diff
@@ -69,3 +69,61 @@ def _can_memcpy(dst, src):
     f_contiguous = dst.flags.f_contiguous and src.flags.f_contiguous
     return (c_contiguous or f_contiguous) and dst.dtype == src.dtype and \
         dst.size == src.size
+
+
+putmask_kernel = core._kernel.ElementwiseKernel(
+    'Q mask, raw T values, uint64 len_vals', 'T out',
+    '''
+    if (mask) out = values[(i % len_vals)];
```
asi1024, line 77:
```suggestion
    if (mask) out = values[i % len_vals];
```

cupy/manipulation/basic.py
```diff
@@ -69,3 +69,61 @@ def _can_memcpy(dst, src):
     f_contiguous = dst.flags.f_contiguous and src.flags.f_contiguous
     return (c_contiguous or f_contiguous) and dst.dtype == src.dtype and \
         dst.size == src.size
+
+
+putmask_kernel = core._kernel.ElementwiseKernel(
```
asi1024, line 74: Please add single underscore prefix as it is private function.
```suggestion
_putmask_kernel = core._kernel.ElementwiseKernel(
```

asi1024:
Please add tests for the combination of dtypes of `a` and `value`.

rushabh-v:
I added the test for different test. But I didn't know that safe casting is not supported yet in cupy. And because of that, the new test fails for `a` and `values` with different dtypes. 

What should we do in this case for compatibility with numpy?

rushabh-v:
IMO, potential options are to make casting `unsafe` OR to restrict different `dtypes` for `a` and `values` and raise an error if they are different.

cupy/manipulation/basic.py
```diff
@@ -69,3 +69,61 @@ def _can_memcpy(dst, src):
...
+    input values.
+
+    Sets ``a.flat[n] = values[n]`` for each n where ``mask.flat[n]==True``.
+    If `values` is not the same size as `a` and `mask` then it will repeat.
+
+    Args
+        a (cupy.ndarray): Target array.
+        mask (cupy.ndarray): Boolean mask array. It has to be
+            the same shape as `a`.
+        values (cupy.ndarray): Values to put into `a` where `mask` is True.
```
asi1024, line 95: For a frequent case, please support scalar `values`.
```suggestion
        values (cupy.ndarray or scalar): Values to put into `a` where `mask` is True.
```

rushabh-v: Okay, will do that. But in case you haven't noticed my latest comments above, can you please respond to them? @asi1024 
So that I can commit both the changes together. See https://github.com/cupy/cupy/pull/3261#issuecomment-610927701 and https://github.com/cupy/cupy/pull/3261#issuecomment-610931251

asi1024: How about using `numpy.can_cast`?

rushabh-v: I have already used that in testing. But the tests are failing for even safe casting from numpy.int16 to numpy.int64. It says `casting is not supported yet`.

asi1024: So, how about using unsafe casting, checking types with `numpy.can_cast` and raising errors if needed?

asi1024: BTW, I think you can implement it without this typecast. Please avoid it for the performance.

rushabh-v: >BTW, I think you can implement it without this typecast. Please avoid it for the performance.

Okay, so I am removing that casting and the test with different dtypes.

asi1024: No, don't remove the tests.

rushabh-v: But they will fail on the kernel if the dtypes of `a` and `values` are different.

asi1024: Please fix to pass tests without typecasting. If you cannot, it's OK to just leave a TODO.

rushabh-v: Can you take a look now?

asi1024: I mean, please keep the compatibility to NumPy without typecasting.

rushabh-v: okay, then would casting the data inside the Kernel perform better?

rushabh-v: >okay, then would casting the data inside the Kernel perform better? I mean I have seen this somewhere in the library.

I did this. Can you please take a look and tell me whether it is a good way or not?

cupy/manipulation/basic.py
```diff
@@ -69,3 +70,69 @@ def _can_memcpy(dst, src):
...
+    if not a.shape == mask.shape:
+        raise ValueError('mask and data must be the same size')
+
+    if cupy.isscalar(values):
+        a[mask] = values
+
+    else:
+        if not numpy.can_cast(values.dtype, a.dtype):
+            raise TypeError("Cannot cast array data from dtype('float64') to "
+                            "dtype('int64') according to the rule 'safe'")
```
asi1024, line 129:
```suggestion
            raise TypeError(
                'Cannot cast array data from {} to {} according to the rule \'safe\''.format(
                    values.dtype, a.dtype)
```

cupy/manipulation/basic.py
```diff
@@ -69,3 +70,69 @@ def _can_memcpy(dst, src):
...
+    .. seealso:: :func:`numpy.putmask`
+
+    """
+    # TODO: support different dtypes for a and values without typecasting
+
+    if not a.shape == mask.shape:
+        raise ValueError('mask and data must be the same size')
+
+    if cupy.isscalar(values):
+        a[mask] = values
```
asi1024, line 124: Could you add tests for this case?

cupy/manipulation/basic.py
```diff
@@ -69,3 +70,69 @@ def _can_memcpy(dst, src):
...
+            raise TypeError("Cannot cast array data from dtype('float64') to "
+                            "dtype('int64') according to the rule 'safe'")
+
+        if a.shape == values.shape:
+            a[mask] = values[mask]
+
+        else:
+            if values.ndim > 1:
+                values = values.flatten()
+            _putmask_kernel(mask.astype(numpy.int64),
```
asi1024, line 137: Why is this typecast needed?

cupy/manipulation/basic.py
```diff
@@ -69,3 +70,69 @@ def _can_memcpy(dst, src):
...
+        if not numpy.can_cast(values.dtype, a.dtype):
+            raise TypeError("Cannot cast array data from dtype('float64') to "
+                            "dtype('int64') according to the rule 'safe'")
+
+        if a.shape == values.shape:
+            a[mask] = values[mask]
+
+        else:
+            if values.ndim > 1:
+                values = values.flatten()
```
asi1024, line 136: For the performance.
```suggestion
            values = values.ravel()
```

cupy/manipulation/basic.py
```diff
@@ -69,3 +70,69 @@ def _can_memcpy(dst, src):
...
+    # TODO: support different dtypes for a and values without typecasting
+
+    if not a.shape == mask.shape:
+        raise ValueError('mask and data must be the same size')
+
+    if cupy.isscalar(values):
+        a[mask] = values
+
+    else:
+        if not numpy.can_cast(values.dtype, a.dtype):
```
asi1024, line 127: Why not use `elif`?

cupy/manipulation/basic.py
rushabh-v: I have already added that test.

cupy/manipulation/basic.py
rushabh-v: because in the kernel it needed it to be `int` in the if condition. But maybe we can also typecast it there in the kernel directly. I am trying that out.

rushabh-v: oh, when I actually first tried passing `bool` values to the kernel I got an error(I don't remember what but something related to dtypes). And I thought we have to pass ints to the kernels instead of bools. But I just tried again and confirmed that we can pass bools directly also.

rushabh-v:
Added the checks for the arguments to be of valid types with the last commit and have made all other requested changes too.

cupy/manipulation/basic.py
```diff
@@ -69,3 +70,74 @@ def _can_memcpy(dst, src):
...
+    elif not numpy.can_cast(values.dtype, a.dtype):
+        raise TypeError('Cannot cast array data from'
+                        ' {} to {} according to the rule \'safe\''
+                        .format(values.dtype, a.dtype))
+
+    elif a.shape == values.shape:
+        a[mask] = values[mask]
+
+    else:
+        if values.ndim > 1:
```
asi1024, line 141: This condition is needed?
```suggestion
        if values.ndim > 1:
```

cupy/manipulation/basic.py
asi1024: Thanks!

cupy/manipulation/basic.py
asi1024: Please check `mask.dtype.kind` and raise an error if needed.

asi1024: Could you add tests for non-bool mask?

cupy/manipulation/basic.py
rushabh-v: removed it!

rushabh-v:
Added support and test for mask of every dtype. NumPy also supports mask of every kind of dtype.

rushabh-v:
Could you please take a look @takagi ?

takagi:
Yes, sorry for being late.

takagi:
CuPy's file structure follows NumPy's reference, where `numpy.putmask` is listed on its indexing routines:
https://numpy.org/doc/stable/reference/routines.indexing.html#inserting-data-into-arrays

Would you move your change into `cupy/indexing/insert.py` ?

rushabh-v:
Thanks for the review.
I have made that change. Can you please take a look back?

cupy/indexing/insert.py
```diff
@@ -66,7 +67,77 @@ def put(a, ind, v, mode='wrap'):
     a.put(ind, v, mode=mode)
 
 
-# TODO(okuta): Implement putmask
+_putmask_kernel = core._kernel.ElementwiseKernel(
```
takagi, line 70: Would you remove a private submodule name?
```suggestion
_putmask_kernel = core.ElementwiseKernel(
```

cupy/indexing/insert.py
```diff
@@ -66,7 +67,77 @@ def put(a, ind, v, mode='wrap'):
...
+    '''
+    if (mask) out = (T) values[i % len_vals];
+    ''',
+    'putmask_kernel'
+)
+
+
+def putmask(a, mask, values):
+    """
+    Changes elements of an array inplace, based on conditional mask and
```
takagi, line 81:
```suggestion
    Changes elements of an array inplace, based on a conditional mask and
```

cupy/indexing/insert.py
```diff
@@ -66,7 +67,77 @@ def put(a, ind, v, mode='wrap'):
...
+    input values.
+
+    Sets ``a.flat[n] = values[n]`` for each n where ``mask.flat[n]==True``.
+    If `values` is not the same size as `a` and `mask` then it will repeat.
+
+    Args
```
takagi, line 87: This colon is necessary: https://google.github.io/styleguide/pyguide.html#doc-function-args
```suggestion
    Args:
```

cupy/indexing/insert.py
```diff
@@ -66,7 +67,77 @@ def put(a, ind, v, mode='wrap'):
...
+    >>> cupy.putmask(x, x>2, [-33, -44])
+    >>> x
+    array([  0,  1,  2, -44, -33, -44])
+
+    .. seealso:: :func:`numpy.putmask`
+
+    """
+
+    if not isinstance(a, cupy.ndarray):
+        raise ValueError('`a` should be of type cupy.ndarray')
```
takagi, line 115: Would you raise `TypeError` here?

cupy/indexing/insert.py
```diff
@@ -66,7 +67,77 @@ def put(a, ind, v, mode='wrap'):
...
+    if not isinstance(mask, cupy.ndarray):
+        raise ValueError('`mask` should be of type cupy.ndarray')
```
takagi, line 117: ditto(TypeError)

cupy/indexing/insert.py
```diff
@@ -66,7 +67,77 @@ def put(a, ind, v, mode='wrap'):
...
+    if not (cupy.isscalar(values) or isinstance(values, cupy.ndarray)):
+        raise ValueError('`values` should be of type cupy.ndarray')
```
takagi, line 119: ditto(TypeError)

takagi:
pfnCI, test this please.

rushabh-v:
I have made the changes you requested @takagi

takagi:
Thanks, would you add `putmask` to the following file to generate its reference?
https://github.com/rushabh-v/cupy/blob/putmask/docs/source/reference/indexing.rst

takagi:
Also, I've found that `cupy.putmask` does not work as `numpy.putmask` in the following case, would you fix it?

```Python
>>> import numpy
>>> import cupy
>>> a = numpy.array([1, 2, 3, 4])
>>> m = numpy.array([0, 1, 2, 3])
>>> numpy.putmask(a, m, 0)
>>> a
array([1, 0, 0, 0])
>>> a = cupy.array([1, 2, 3, 4])
>>> m = cupy.array([0, 1, 2, 3])
>>> cupy.putmask(a, m, 0)
>>> a
array([0, 0, 0, 0])
```

rushabh-v:
>Also, I've found that cupy.putmask does not work as numpy.putmask in the following case, would you fix it?

Fixed it and also added a test for that! 
And added `cupy.putmask` to `indexing.rst` too.

takagi:
pfnCI, test this please.

rushabh-v:
There was a problem in the example in the docstring which was causing the failure in Jenkins build. I have fixed it. It should pass all the checks now.

takagi:
pfnCI, test this please.

rushabh-v:
Oh, I am sorry. There was a spacing mistake between the elements of the output array in the example. I have fixed it now!

rushabh-v:
I am wondering if I can run these checks locally or not!

asi1024:
You can check it locally with `make doctest` in `docs` directory.

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -115,6 +115,91 @@ def test_put_mode_error(self, dtype):
...
 
 
+@testing.parameterize(
+    *testing.product(
+        {'shape': [(0,), (1,), (2, 3), (2, 3, 4)]}))
+@testing.gpu
+class TestPutmaskEqual(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_allclose()
```
takagi, line 125:
```suggestion
    @testing.numpy_cupy_array_equal()
```

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -115,6 +115,91 @@ def test_put_mode_error(self, dtype):
...
+    def test_putmask(self, xp, dtype):
+        a = testing.shaped_random(self.shape, xp, dtype=dtype, seed=0)
+        mask = testing.shaped_random(self.shape, xp, dtype=bool, seed=1)
```
takagi, line 128:
```suggestion
        mask = testing.shaped_random(self.shape, xp, dtype=numpy.bool_, seed=1)
```

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -115,6 +115,91 @@ def test_put_mode_error(self, dtype):
...
+
+@testing.parameterize(
+    *testing.product(
+        {'shape': [(0,), (1,), (2, 3), (2, 3, 4)],
+         'values_shape': [(2,), (3, 1), (5,)]}))
+@testing.gpu
+class TestPutmaskNonEqual(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_allclose()
```
takagi, line 143:
```suggestion
    @testing.numpy_cupy_array_equal()
```

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -115,6 +115,91 @@ def test_put_mode_error(self, dtype):
...
+    def test_putmask(self, xp, dtype):
+        a = testing.shaped_random(self.shape, xp, dtype=dtype, seed=3)
+        mask = testing.shaped_random(self.shape, xp, dtype=bool, seed=4)
```
takagi, line 146:
```suggestion
        mask = testing.shaped_random(self.shape, xp, dtype=numpy.bool_, seed=4)
```

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -115,6 +115,91 @@ def test_put_mode_error(self, dtype):
                 xp.put(a, inds, vals, mode='unknown')
 
 
+@testing.parameterize(
+    *testing.product(
+        {'shape': [(0,), (1,), (2, 3), (2, 3, 4)]}))
+@testing.gpu
+class TestPutmaskEqual(unittest.TestCase):
```
takagi, line 122: How about `TestPutmaskSameShape`?

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -115,6 +115,91 @@ def test_put_mode_error(self, dtype):
...
+        assert ret is None
+        return a
+
+
+@testing.parameterize(
+    *testing.product(
+        {'shape': [(0,), (1,), (2, 3), (2, 3, 4)],
+         'values_shape': [(2,), (3, 1), (5,)]}))
+@testing.gpu
+class TestPutmaskNonEqual(unittest.TestCase):
```
takagi, line 140: How about `TestPutmaskDifferentShapes`?

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -115,6 +115,91 @@ def test_put_mode_error(self, dtype):
...
+
+
+class TestPutmaskDifferentDtypes(unittest.TestCase):
+
+    @testing.for_all_dtypes_combination(names=['a_dtype', 'val_dtype'])
+    def test_putmask_differnt_dtypes_raises(self, a_dtype, val_dtype):
+        shape = (2, 3)
+        for xp in (numpy, cupy):
+            a = testing.shaped_random(shape, xp, dtype=a_dtype)
+            mask = testing.shaped_random(shape, xp, dtype=bool)
```
takagi, line 186:
```suggestion
            mask = testing.shaped_random(shape, xp, dtype=numpy.bool_)
```

takagi:
pfnCI, test this please.

takagi:
Thanks!

PR #3262, opened 2020-04-06
Fix `_reduce_dims` call in reduction

### LINKED ISSUE #3258: Reduction fails with `CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES` in some inputs
Reduction routines become to fail in some cases after #2920 was merged.
Reproducer:
```py
import cupy

shape = (2, 2, 2, 2, 2, 2, 2)
axis = (1, 3, 4, 6)
a = cupy.zeros(shape, dtype=cupy.float32)
c = cupy.sum(a, axis=axis, keepdims=True)
```
Traceback of the error:
```
  File "a.py", line 5, in <module>
    c = cupy.sum(a, axis=axis, keepdims=True)
  File "/home/imanishi/cupy1/cupy/math/sumprod.py", line 34, in sum
    return a.sum(axis, dtype, out, keepdims)
  File "cupy/core/core.pyx", line 859, in cupy.core.core.ndarray.sum
    cpdef ndarray sum(self, axis=None, dtype=None, out=None, keepdims=False):
  File "cupy/core/core.pyx", line 867, in cupy.core.core.ndarray.sum
    return _math._ndarray_sum(self, axis, dtype, out, keepdims)
  File "cupy/core/_routines_math.pyx", line 101, in cupy.core._routines_math._ndarray_sum
    return _sum_auto_dtype(self, axis, dtype, out, keepdims)
  File "cupy/core/_reduction.pyx", line 409, in cupy.core._reduction._SimpleReductionKernel.__call__
    return self._call(
  File "cupy/core/_reduction.pyx", line 329, in cupy.core._reduction._AbstractReductionKernel._call
    func.linear_launch(
  File "cupy/cuda/function.pyx", line 187, in cupy.cuda.function.Function.linear_launch
    _launch(
  File "cupy/cuda/function.pyx", line 154, in cupy.cuda.function._launch
    driver.launchKernel(
  File "cupy/cuda/driver.pyx", line 299, in cupy.cuda.driver.launchKernel
    check_status(status)
  File "cupy/cuda/driver.pyx", line 124, in cupy.cuda.driver.check_status
    raise CUDADriverError(status)
cupy.cuda.driver.CUDADriverError: CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES: too many resources requested for launch
```

emcastillo:
Closes #3258

When calling `_reduce_dims`, the out_args tuple was empty, so dimensions were never reduced.
`_get_out_args` is the function that initializes the actual arrays if the `out` parameter is an empty list, and it wasn't called until later.

This PR calls `_get_out_args` before and then retrieves its `ndarray` objects, so that dimensions can be effectively reduced. Then it recreated the output as the dimensionality might have changed.

Another way to solve this could be to make `_reduce_dims` allow a list of `Arg` objects as the parameter and modify it accordingly.

`ElementwiseKernel` should be check to verify if this problem also happens there.

### DIFF
--- a/cupy/core/_reduction.pyx
+++ b/cupy/core/_reduction.pyx
@@ -264,7 +264,8 @@ cdef class _AbstractReductionKernel:
         out_shape = _get_out_shape(a_shape, reduce_axis, out_axis, keepdims)
         # Needs to create the empty arrays first and access them for the
         # _reduce_dims to work.
-        out_args_ = self._get_out_args(out_args, out_types, out_shape)
+        out_args_ = [Arg.from_obj(a) for a in out_args]
+        out_args_ = self._get_out_args(out_args_, out_types, out_shape)
         out_arrays_ = [(<Arg>a).obj for a in out_args_]
 
         if self.identity == '':
--- a/cupy/core/_reduction.pyx
+++ b/cupy/core/_reduction.pyx
@@ -262,6 +262,10 @@ cdef class _AbstractReductionKernel:
             out_axis = _sort_axis(out_axis, strides)
 
         out_shape = _get_out_shape(a_shape, reduce_axis, out_axis, keepdims)
+        # Needs to create the empty arrays first and access them for the
+        # _reduce_dims to work.
+        out_args_ = self._get_out_args(out_args, out_types, out_shape)
+        out_arrays_ = [(<Arg>a).obj for a in out_args_]
 
         if self.identity == '':
             for axis in reduce_axis:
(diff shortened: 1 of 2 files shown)


emcastillo:
Fixed, PTAL

cupy/core/_reduction.pyx
```diff
@@ -262,32 +262,35 @@ cdef class _AbstractReductionKernel:
             out_axis = _sort_axis(out_axis, strides)
 
         out_shape = _get_out_shape(a_shape, reduce_axis, out_axis, keepdims)
+        # Needs to create the empty arrays first and access them for the
+        # _reduce_dims to work.
+        out_args_ = self._get_out_args(out_args, out_types, out_shape)
```
emcastillo, line 267: This is where the actual out arrays are instantiated when `out=None`

asi1024:
Could you add a test for this case?

emcastillo:
The behavior is not exposed externally, so It's hard to test.
The only approach I can think is to add the case that failed, (long number of dims)

asi1024:
It can be tested with a mock test, but I think it is OK to do it in another PR.

asi1024:
Jenkins, test this please.

emcastillo:
Sorry, I just pushed a test based on the original error now 😆
I verified that this test fails in the master branch.

asi1024:
Jenkins, test this please.

emcastillo:
pfnCI is appearing but Jenkins is missing from the checks 😂

asi1024:
Jenkins, test this please.

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

emcastillo:
Fixed, PTAL

asi1024:
Jenkins, test this please.

kmaehashi:
pfnCI, test this please.

PR #2746, opened 2019-12-02
reduce overhead in calls to multi-dimensional FFTs.

grlee77:
This PR refactors some of the axis handling logic into a `_prep_fftn_axes` helper function which can be memoized, reducing FFT call overhead. Overall this PR also results in a small net reduction in lines of code. The determination of the output array dtype was also memoized here.

Prior to this PR, on a somewhat slow older PC, a call to `_default_fft_func` (called by all non-1d functions to determine whether to use 1D or nD planning) took approximately 40 microseconds. After the refactoring here, it takes ~1 microsecond. Prior to this PR, the overhead could be a substantial fraction of the total `fftn` time for smaller FFTs (e.g. 512 x 512) when using pre-planning + plan context manager.

The changes here are within `cupy.fft.fft` and so will also benefit `cupyx.scipy.fftpack` and upcoming `cupyx.scipy.fft` as well.

PR #3264, opened 2020-04-07
Raise `IndexError` for R2C/C2R FFT with `axes=()`

### LINKED ISSUE #3240: Real FFTs with empty `axes` should fail
The shape of the output of `fft.(i)rfft*` depends on the last axis.
```
>>> numpy.fft.rfftn(numpy.arange(3), axes=())
  File "<__array_function__ internals>", line 5, in rfftn
  File "/path/to/python3.8/site-packages/numpy-1.18.2-py3.8-linux-x86_64.egg/numpy/fft/_pocketfft.py", line 1124, in rfftn
    a = rfft(a, s[-1], axes[-1], norm)
IndexError: list index out of range
>>> cupy.fft.rfftn(cupy.arange(3), axes=())
array([0, 1, 2])
```

leofang:
Close #3240. This makes the behavior compatible with NumPy.

### DIFF
--- a/cupy/fft/fft.py
+++ b/cupy/fft/fft.py
@@ -166,7 +166,7 @@ def _fft(a, s, axes, norm, direction, value_type='C2C', overwrite_x=False,
         axes = [i for i in range(-dim, 0)]
     else:
         axes = tuple(axes)
-    if len(axes) == 0:
+    if not axes:
         if value_type == 'C2C':
             return a
         else:
@@ -422,7 +422,7 @@ def _fftn(a, s, axes, norm, direction, value_type='C2C', order='A', plan=None,
                          % norm)
 
     axes, axes_sorted = _prep_fftn_axes(a.ndim, s, axes)
-    if len(axes_sorted) == 0:
+    if not axes_sorted:
         if value_type == 'C2C':
             return a
         else:
--- a/cupy/fft/fft.py
+++ b/cupy/fft/fft.py
@@ -166,8 +166,11 @@ def _fft(a, s, axes, norm, direction, value_type='C2C', overwrite_x=False,
         axes = [i for i in range(-dim, 0)]
     else:
         axes = tuple(axes)
-    if not axes:
-        return a
+    if len(axes) == 0:
+        if value_type == 'C2C':
+            return a
+        else:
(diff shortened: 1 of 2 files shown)


grlee77:
Agreed that this matches NumPy behavior.

I'm not sure if the difference between C2C and R2C/C2R cases was intentional or an oversight on the NumPy side, though?

leofang:
My understanding for R2C/C2R is the last axis plays a special role, whereas for C2C all axes are treated on equal footing, so if we can't define clearly which axis is the last, it'd be a problem. So it's intentional I guess.

cupy/fft/fft.py
```diff
@@ -419,8 +422,11 @@ def _fftn(a, s, axes, norm, direction, value_type='C2C', order='A', plan=None,
                          % norm)
 
     axes, axes_sorted = _prep_fftn_axes(a.ndim, s, axes)
-    if not axes_sorted:
-        return a
+    if len(axes_sorted) == 0:
```
toslunar, line 425: ditto

cupy/fft/fft.py
```diff
@@ -166,8 +166,11 @@ def _fft(a, s, axes, norm, direction, value_type='C2C', overwrite_x=False,
         axes = [i for i in range(-dim, 0)]
     else:
         axes = tuple(axes)
-    if not axes:
-        return a
+    if len(axes) == 0:
```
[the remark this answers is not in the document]
leofang:
```suggestion
    if not axes:
```

cupy/fft/fft.py
leofang:
```suggestion
    if not axes_sorted:
```

leofang:
Thanks, @toslunar. That's a good suggestion on PEP8. I just fixed it. PTAL.

toslunar:
LGTM

Jenkins, test this please.

toslunar:
The cherry-pick conflicted with #2746, which is only in `master` (v8).  I'm checking if the issue exists in `v7`.

toslunar:
`v7` raises `IndexError: tuple index out of range` or `IndexError: list index out of range`.

Issue #3108, opened 2020-02-21
Include CUB info in show_config


jakirkham:
It would be useful to include info about the CUB build in `cupy.show_config()`. In particular was it enabled and what version/commit were used if so.

leofang:
I thought about this too, but it doesn't seem possible to fetch the CUB version because there's none from the source...

jakirkham:
Interesting. So they don’t define this in a `#define` somewhere?

leofang:
Didn’t find it. Only git tags AFAIK...I’ll take a look how PyTorch handles this issue later.

jakirkham:
No worries. Happy to poke around this as well. Just figured it might be handy when debugging issues 🙂

leofang:
PyTorch takes effectively the same approach as in CuPy: if the CUB headers are found, it uses them. Otherwise, it uses CUB from submodule. John, perhaps it's time for you to poke NVIDIA people? 😉 

(BTW, they've stopped updating CUB for quite a while...I am a bit concerned...)

leofang:
I think for the time being we can just use the git tag (currently we're using v1.8.0) as the source of version string.

leofang:
@jakirkham @emcastillo Catching up with this issue. Apparently, NVIDIA has added a version string to CUB starting v1.9.9 (not released yet), see this commit: https://github.com/thrust/cub/commit/629f01ec7b4f660d293899ab84680fb7819ece42. (Note CUB is taken over from NVLabs to Thrust.) So, if the bundled version is v1.9.9+, getting a version string from the source code wouldn't be a problem. We won't need to parse the git tag records.

emcastillo:
We can then show in the cub version field `< 1.9.9` or the exact one

PR #3271, opened 2020-04-09
Show versions of cub and cutensor on `cupy.show_config`

emcastillo:
Cub has no way to retrieve the version number and it is not written in any header that we can parse  :( so it shows `Enabled` instead ...

### DIFF
--- a/cupyx/runtime.py
+++ b/cupyx/runtime.py
@@ -14,6 +14,16 @@ try:
 except ImportError:
     nccl = None
 
+try:
+    import cupy.cuda.cub as cub
+except ImportError:
+    cub = None
+
+try:
+    import cupy.cuda.cutensor as cutensor
+except ImportError:
+    cutensor = None
+
 
 def _eval_or_error(func, errors):
     # Evaluates `func` and return the result.
@@ -134,6 +144,14 @@ class _RuntimeInfo(object):
                 nccl_runtime_version = '(unknown)'
             self.nccl_runtime_version = nccl_runtime_version
 
+        if cub is not None:
+            # There is no API in CUB to retrieve the current version
+            # We show if its enabled or disabled
+            self.cub_version = 'Enabled'
+
+        if cutensor is not None:
+            self.cutensor_version = cutensor.get_version()
+
     def __str__(self):
         records = [
             ('CuPy Version', self.cupy_version),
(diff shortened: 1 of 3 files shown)


cupyx/runtime.py
```diff
@@ -134,6 +144,14 @@ def __init__(self):
                 nccl_runtime_version = '(unknown)'
             self.nccl_runtime_version = nccl_runtime_version
 
+        if cub is not None:
```
asi1024, line 147: How about using `cupy.cuda.cub_enabled` simply?

emcastillo: It was to maintain consistency with the above checks 😅

leofang:
I did have a plan to parse the git tag for CUB (see https://github.com/cupy/cupy/issues/3108#issuecomment-590962232), but for the time being let's do this.

cupyx/runtime.py
leofang: Yeah I think here we want to know if the module is built, not if it is enabled.

emcastillo: I didn‘t thought about it that way but makes complete sense

leofang:
@emcastillo Not directly related to this PR but I wonder if you could add `CUTENSOR_PATH` to here: https://github.com/cupy/cupy/blob/f7fc9b9df43ec31918f90b1a34747d0630b531dc/cupy_setup_build.py#L342-L344
I found it easier to debug. (If you want you could also add `CUB_PATH`, although it'd be ignored after #2584 is merged.)

kmaehashi:
pfnCI, test this please.

emcastillo:
Worked fine

```
14:33:46 CuPy Version          : 8.0.0b2
14:33:46 CUDA Root             : /usr/local/cuda
14:33:46 CUDA Build Version    : 10000
14:33:46 CUDA Driver Version   : 10020
14:33:46 CUDA Runtime Version  : 10000
14:33:46 cuBLAS Version        : 10000
14:33:46 cuFFT Version         : 10000
14:33:46 cuRAND Version        : 10000
14:33:46 cuSOLVER Version      : (10, 0, 0)
14:33:46 cuSPARSE Version      : 10000
14:33:46 NVRTC Version         : (10, 0)
14:33:46 cuDNN Build Version   : 7500
14:33:46 cuDNN Version         : 7500
14:33:46 NCCL Build Version    : 2402
14:33:46 NCCL Runtime Version  : 2402
14:33:46 CUB Version           : None
14:33:46 cuTENSOR Version      : 10001
```
Detected cutensor in the Jenkins job for cutensor :)

PR #3272, opened 2020-04-09
Use sphinx 2.x on Read the Docs

takagi:
Rel #3250 and https://github.com/chainer/chainer-test/issues/571 .

This PR changes the requirements file to specify sphinx 2.x to use on Read the Docs.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
+sphinx<3
 Cython==0.28.3
 numpy >=1.15, <1.16
 scipy >=1.1, <1.2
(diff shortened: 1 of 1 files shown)


docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
+sphinx<3
```
emcastillo, line 1: I think we need to specify a minimum version, otherwise the current readthedocs one is <3 so it won't be upgraded.
Also I think its better to stick to a single one, such as sphinx==2.0.1.
Thanks!

takagi: Okay!

emcastillo:
Jenkins, test this please

I verified this worked in read-the-docs by creating a new branch in the repo and making it active in rtd.

PR #3273, opened 2020-04-09
[WIP] Use fuse to improve norm 

okuta:


asi1024:
How about using ReductionKernel?

asi1024:
The current implementation creates kernels for each combination of `axis` and `keepdims`, but `cupy.ReductionKernel` does not.

leofang:
I know very little about `@fuse()`, but I think if you provide a fast path using `create_reduction_func()`, #3244 should provide a very good performance.

okuta:
Thanks!  #3244 is a better Implementation than this.

PR #3280, opened 2020-04-14
ENH: support additional dtypes and axis sequences in cupy.median

grlee77:
~As in NumPy, bool gets converted to double~

~float16 inputs are temporarily converted to float32 for compatibility with `partition`~

~For complex dtypes, `sort` is used because `partition` lacks complex support.~
#3286 removed the need for the above workarounds.

**edit**: 
I updated this PR with enhancements for axis support as well. Let me know if you would prefer to keep that in a separate PR instead.
- Support for axis as a sequence of ints has been added
- when `axis` is out of range, `numpy.AxisError` is now returned instead of `IndexError`.

### DIFF
--- a/cupy/core/_routines_statistics.pyx
+++ b/cupy/core/_routines_statistics.pyx
@@ -313,6 +313,9 @@ cdef _nanargmax_func = create_reduction_func(
 cpdef ndarray _median(
         ndarray a, axis, out, overwrite_input, keepdims):
 
+    if a.dtype.char == '?':
+        a = a.astype(cupy.float64)
+
     keep_ndim = a.ndim
 
     if axis is None:
@@ -325,19 +328,30 @@ cpdef ndarray _median(
     else:
         kth = [(sz - 1) // 2]
 
-    if overwrite_input:
+    if a.dtype.char == 'e':
+        # partition doesn't support float16, so temporarily convert to float32
+        part = a.astype(cupy.float32)
+    elif overwrite_input:
         part = a
     else:
         part = a.copy()
 
     if axis is None:
         part = part.ravel()
-        part.partition(kth)
+        if part.dtype.kind == 'c':
+            part.sort()
+        else:
+            part.partition(kth)
     else:
-        part.partition(kth, axis=axis)
+        if part.dtype.kind == 'c':
+            part.sort(axis=axis)
+        else:
+            part.partition(kth, axis=axis)
 
     if part.shape == ():
         return part
+    if a.dtype.char == 'e':
+        part = part.astype(cupy.float16)
     if axis is None:
         axis = 0
 
--- a/cupy/core/_routines_statistics.pyx
+++ b/cupy/core/_routines_statistics.pyx
@@ -326,19 +326,21 @@ cpdef ndarray _median(
         # cupy.sort and cupy.partition only support integer axis, so move
         # all reduced dimensions to the end and reshape them into a single
         # reduction axis.
-        reduce_axis, out_axis = _reduction._get_axis(axis, ndim)
+        reduce_axis, out_axis = _reduction._get_axis(axis, keep_ndim)
         out_shape = _reduction._get_out_shape(a.shape, reduce_axis, out_axis,
                                               keepdims)
         a = a.transpose(out_axis + reduce_axis)
         sort_shape = tuple([a.shape[n] for n in range(len(out_axis))]) + (-1,)
         a = a.reshape(sort_shape)
         if not a.flags.c_contiguous:
-            a = cp.ascontiguousarray(a)
+            a = cupy.ascontiguousarray(a)
         axis = -1
 
     if axis is None:
         sz = a.size
     else:
+        if axis < -keep_ndim or axis >= keep_ndim:
+            raise numpy.AxisError('Axis overrun')
         sz = a.shape[axis]
     if sz % 2 == 0:
         szh = sz // 2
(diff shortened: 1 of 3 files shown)


cupy/core/_routines_statistics.pyx
```diff
@@ -325,19 +328,30 @@ cpdef ndarray _median(
     else:
         kth = [(sz - 1) // 2]
 
-    if overwrite_input:
+    if a.dtype.char == 'e':
+        # partition doesn't support float16, so temporarily convert to float32
+        part = a.astype(cupy.float32)
```
emcastillo, line 333: how hard would be to modify partition to support float16?

grlee77: Currently the failure for float16 is in `thrust.sort` which gets called here under certain conditions:
https://github.com/cupy/cupy/blob/49fc35379f17b056a0f311992a85ccb639a5953c/cupy/core/_routines_sorting.pyx#L178-L180

I added a new commit here that moves the float32 conversion into this code branch of `partition` instead of `median`. I didn't try looking at the code in thrust itself.

emcastillo: Thanks!!

tests/cupy_tests/padding_tests/test_pad.py
```diff
@@ -10,7 +10,8 @@
 
 @testing.parameterize(
     *testing.product({
-        'array': [numpy.arange(6).reshape([2, 3])],
```
emcastillo, line 13: I think you mixed two PRs?😅

grlee77: right. I was going to enable `mode='median'` in cupy.pad and accidentally checked that in here.

tests/cupy_tests/padding_tests/test_pad.py
```diff
@@ -10,7 +10,8 @@
...
+        'array': [numpy.arange(6).reshape([2, 3]),
+                  numpy.arange(8000).reshape(4000, 2)],
```
grlee77, line 14:
```suggestion
        'array': [numpy.arange(6).reshape([2, 3])],
```

grlee77:
I rebased the last commit to get rid of the unintential cupy.pad changes

leofang:
I am looking into whether thrust sort can support fp16 and bool or not.

leofang:
We need to modify `operator<` in `cupy_fp16.hpp` so that it can handle comparisons with NaN. Can I touch CuPy's copy of this header?

leofang:
NVM. Thrust provides an overloaded version that accepts a comparison operator.

tests/cupy_tests/padding_tests/test_pad.py
emcastillo: 😁

leofang:
@grlee77 @emcastillo I have figured out how to support fp16 in `sort()` (and bool seems trivial), would you like to wait for my PR?

grlee77:
> @grlee77 @emcastillo I have figured out how to support fp16 in sort() (and bool seems trivial), would you like to wait for my PR (#3286)?

Yeah, that sounds fine. It looks like the dtype workarounds here will become unnecessary after that is merged.

emcastillo:
I'd like to wait for #3286 if possible 😇

leofang:
#3286 is merged

grlee77:
Okay, I rebased this after #3286 was merged. Now the dtype support is trivial and only involves enabling the test cases!

The main new feature in this PR is now the axis support.

cupy/core/_routines_statistics.pyx
```diff
@@ -315,9 +318,26 @@ cpdef ndarray _median(
...
+        a = a.reshape(sort_shape)
+        if not a.flags.c_contiguous:
+            a = cupy.ascontiguousarray(a)
+        axis = -1
+
     if axis is None:
         sz = a.size
     else:
+        if axis < -keep_ndim or axis >= keep_ndim:
+            raise numpy.AxisError('Axis overrun')
```
emcastillo, line 340: nit: how about raising a cupy exception instead of numpy.
If AxisError is not available, ValueError should do it.

grlee77: There are several other `.pyx` files in `cupy/core` (e.g. `core/_routines_sorting.pyx`) that do raise `numpy.AxisError`, so I was following that example. (`cupy.AxisError` does not seem to exist)

Did you mean to try something like:
```Python
try:
    raise numpy.AxisError('Axis overrun')
except AttributeError:
    raise ValueError('Axis overrun')
```
or just to raise `ValueError` in the first place?

leofang: There used to be an `_AxisError` in `cupy/core/_errors.py`, but #3096 removed it. I don't get why.

grlee77: >  I don't get why.

I think it was just deemed unnecessary now that CuPy no longer supports older NumPy predating `numpy.AxisError`

leofang: Sorry perhaps I'm being stupid 😅 Doesn't NumPy still have `numpy.AxisError` in newer versions?

grlee77: Yes. It was introduced in 1.13. That is why I think it seems a bit odd to add a try/except like above at this point.

Should CuPy define a `cupy.AxisError` in its API? (perhaps as a subclass of numpy.AxisError). Adding that would best be addressed in a separate PR, though.

emcastillo: If other files raise it then I am ok with it.
Sorry for not verifying.

We should re-add the exception and homogenize everything in a different PR.

Thanks!

emcastillo:
Jenkins, test this please

Issue #3277, opened 2020-04-13
pad is slow than zeros and put values with slice


yxdragon:
Hi, firstly I 's like to tell you we build a cnn framework based on pure numpy-interface [planer](https://github.com/Image-Py/planer) (with out any kernel gpu code and jit code). So we try to replace numpy with cupy to get a better performance. (now it is very ok, 1-3 times slower than torch, some models even faster than torch).

We try to count every where 's time cost, adjust the code to get the best performance. And we found some function is not as fast as we think. We will report them one by one later, here is the first one.

1. for the "zeros edge mode", we use pad_img = np.zeros(xxx) to add a margin, then use pad_img[margin:-margin,....] = img to put values in. It is several times faster than pad.

2. for the "nearest edge mode", we use concatenate([img[:left], img, img[right:]]), then concatenate([img[:top], img, img[bot:]]), even with 2 concatenate, it is also faster then pad.

So we think pad is not well optimized. Some report would comes later, Thanks!

grlee77:
Hi @yxdragon. Can you confirm which version of CuPy you are using? There were some improvements to pad in #2399, but if you are using 7.0 or newer you should already have those. 

During pad refactoring in NumPy, I had suggested a fast path for constant padding at https://github.com/lagru/numpy/pull/3, but this did not end up in the final PR there. We could potentially revisit that here.

grlee77:
For me, if images are relatively large, (i.e. 2048x2048) the current `cupy.pad` gives equivalent performance to the simple method you mentioned, but there is substantial overhead in the case of small arrays.

I think the main issue here is likely that, although #2399 did improve things by removing duplicate allocations/concatenations, there is still overhead from per-axis kernel launches that could be avoided by further refactoring.

PR #3281, opened 2020-04-14
improve efficiency of cupy.pad for some simple cases

grlee77:
This should mostly resolve the zero-padding performance issue reported in #3277.

The change to `_as_pairs` reduces argument handling time for the `constant_values`, `end_values` and  `stat_length` kwargs when a scalar is provided.

### DIFF
--- a/cupy/padding/pad.py
+++ b/cupy/padding/pad.py
@@ -366,6 +366,10 @@ def _as_pairs(x, ndim, as_index=False):
         # Pass through None as a special case, otherwise cupy.round(x) fails
         # with an AttributeError
         return ((None, None),) * ndim
+    elif isinstance(x, numbers.Number):
+        if as_index:
+            x = round(x)
+        return ((x, x),) * ndim
 
     x = numpy.array(x)
     if as_index:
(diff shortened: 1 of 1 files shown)


grlee77:
Using `mode='constant'` with `'constant_values=0'` is 3-4 times faster for small arrays (e.g. shape (64, 64)) and approximately the same as before for large arrays.

For float32 inputs on a GTX 1080Ti I measured: 

size | acceleration factor (this PR vs. master)
------| ------------------------
(64, 64) | 3.27
(256, 256) | 3.59
(1024, 1024) | 2.31
(2048, 2048) | 0.99
(4096, 4096) | 1.04

grlee77:
benchmarks above correspond to commit 2c84fbd.

takagi:
pfnCI, test this please.

takagi:
Jenkins CI fails because an unsupported keyword argument does not raise an exception when it goes into the fast path.  Would you move it after the unsupported keyword arguments check? I suppose moving it does not make significant performance effect.

grlee77:
@takagi. I have moved it (and removed one duplicate line). Please retry on CI

takagi:
Thanks! pfnCI, test this please.

PR #3282, opened 2020-04-14
Bug: Add check for method in Randomstate seed

### LINKED ISSUE #3275: Unable to create RandomState with MT19937
Hello,

I work with two versions of scripts that do some math operations. I have a CPU version that works with numpy and a GPU one that works with cupy. I wanted to validate that the code will return the same results when it has the same config. In my code I sometimes need to use some random distributions, therefore ideally I want to set up numpy and cupy in the same way so far as RandomState goes. This is where I encountered a problem - it seems I'm unable to set it up in the same way (same RandomState setup). I wanted to use the MT19937 random state for this comparison. However, on cupy I'm getting the following error when creating the random state :
```
cp.random.RandomState(method=cp.cuda.curand.CURAND_RNG_PSEUDO_MT19937)
  File "/home/uname/anaconda3/envs/sai/lib/python3.7/site-packages/cupy/random/generator.py", line 56, in __init__
    self.seed(seed)
  File "/home/uname/anaconda3/envs/sai/lib/python3.7/site-packages/cupy/random/generator.py", line 697, in seed
    curand.setGeneratorOffset(self._generator, 0)
  File "cupy/cuda/curand.pyx", line 123, in cupy.cuda.curand.setGeneratorOffset
  File "cupy/cuda/curand.pyx", line 125, in cupy.cuda.curand.setGeneratorOffset
  File "cupy/cuda/curand.pyx", line 86, in cupy.cuda.curand.check_status
cupy.cuda.curand.CURANDError: CURAND_STATUS_TYPE_ERROR
```

My setup is the following (output from `python -c 'import cupy; cupy.show_config()'`):
```
CuPy Version          : 7.3.0  
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSPARSE Version      : 10301
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2506
NCCL Runtime Version  : 2506
```
I'm using an Nvidia GTX 980 Ti GPU.

PS: Also I should point out that you should probably update your documentation, since the actual RandomState method enumeration: https://github.com/cupy/cupy/blob/v7.3.0/cupy/cuda/curand.pxd
does not match the method enumeration mentioned in the docs: https://docs-cupy.chainer.org/en/stable/reference/generated/cupy.random.RandomState.html#cupy.random.RandomState

niteya-shah:
Fixes #3275


cupy/random/generator.py
```diff
@@ -739,7 +740,9 @@ def seed(self, seed=None):
                     numpy.asarray(seed).astype(numpy.uint64, casting='safe'))
 
         curand.setPseudoRandomGeneratorSeed(self._generator, seed)
-        curand.setGeneratorOffset(self._generator, 0)
+        if(self.method != curand.CURAND_RNG_PSEUDO_MT19937 and
+           self.method != curand.CURAND_RNG_PSEUDO_MTGP32):
```
emcastillo, line 744: nit here:
```python
if (self.method not in (curand.CURAND_RNG_PSEUDO_MT19937, 
        curand.CURAND_RNG_PSEUDO_MTGP32):
```

niteya-shah: Done. Should I also add a test case as there wasn't one that tests all the different methods available?

emcastillo: would be great to have

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):
LGTM!

Issue #3232, opened 2020-03-28
Lexsort output is incorrect


MrArsGravis:
The output of cupy.lexsort is incorrect. I can't see any pattern as to what is wrong with the output.

### Code to reproduce
```
import cupy
import numpy

# numpy output is correct:
a = (numpy.random.random((10,2)) * 10).astype(int)  # random ints between 0 and 9
print(a[numpy.lexsort(a.T[::-1])])                  # sorted by first column, not last

# cupy output is unsorted:
b = cupy.array(a)
print(b[cupy.lexsort(b.T[::-1])])
```
The same thing happens for floats, so that's not the reason.

### Conditions
(output of `python -c 'import cupy; cupy.show_config()'`):

CuPy Version          : 7.1.1
CUDA Root             : /opt/cuda
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSOLVER Version      : (10, 3, 0)
cuSPARSE Version      : 10301
NVRTC Version         : (10, 2)
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2507
NCCL Runtime Version  : 2507

takagi:
Thanks for the report, I can reproduce the result.

leofang:
Not sure if this is relevant, but lexsort would fail these two tests:
```diff
diff --git a/tests/cupy_tests/sorting_tests/test_sort.py b/tests/cupy_tests/sorting_tests/test_sort.py
index b00856866..5b0c67d58 100644
--- a/tests/cupy_tests/sorting_tests/test_sort.py
+++ b/tests/cupy_tests/sorting_tests/test_sort.py
@@ -202,6 +202,20 @@ class TestLexsort(unittest.TestCase):
         with self.assertRaises(TypeError):
             return cupy.lexsort(a)
 
+    @testing.for_dtypes('fdFD')
+    @testing.numpy_cupy_allclose()
+    def test_nan2(self, xp, dtype):
+        a = testing.shaped_random((2, 10), xp, dtype)
+        a[0, 2] = a[1, 6] = xp.nan
+        return xp.lexsort(a)
+
+    @testing.for_dtypes('fdFD')
+    @testing.numpy_cupy_allclose()
+    def test_nan3(self, xp, dtype):
+        a = testing.shaped_random((2, 10), xp, dtype)
[...]

PR #2745, opened 2019-12-02
Support sorting complex arrays

leofang:
Support `sort()`/`msort()`, `lexsort()` and `argsort()` on `complex64` and `complex128` arrays. I am making the build time of the `cupy.cuda.thrust` module even longer! 🤣

Supporting complex numbers in the `thrust` module  is trickier than in the `cub` module, because it can only depend on `cuComplex.h`. I couldn't get the complex headers from either CuPy or Thrust to work. The former compiles but segfaults at import time (likely due to the namespace collision), and the latter gives me "no matching `thrust::complex<T>::real` found" error even after I added specializations of `operator<()` and `real()` (a77d1bf contains leftovers of failed attempts). Perhaps enabling CUB's `sort` support would circumvent this problem, but it should be a separate PR. 

NumPy has a special rule for sorting complex NaNs, see https://docs.scipy.org/doc/numpy/reference/generated/numpy.sort.html. It is taken care of in this PR.

PR #3286, opened 2020-04-15
Support all dtypes in every sorting function in `cupy.cuda.thrust`

leofang:
All kinds of sort functions work with all dtypes now!

~~Ongoing, **DO NOT MERGE!**~~
- ~~msort~~
- ~~lexsort~~
- ~~remove redundant tests~~
- ~~clean up~~
- ~~test for performance regression~~ LGTM

### DIFF
--- a/cupy/core/_routines_sorting.pyx
+++ b/cupy/core/_routines_sorting.pyx
@@ -122,10 +122,6 @@ cdef _ndarray_partition(ndarray self, kth, int axis):
 
     """
 
-    if self.dtype.kind == 'c':
-        raise NotImplementedError('Sorting arrays with dtype \'{}\' is '
-                                  'not supported'.format(self.dtype))
-
     cdef int ndim = self._shape.size()
     cdef Py_ssize_t k, max_k, length, s, sz, t
     cdef ndarray data
@@ -282,7 +278,7 @@ def _partition_kernel(dtype):
             CArray<T, 1, true> a, ptrdiff_t x, ptrdiff_t y, int i) {
         for (ptrdiff_t s = 2; s <= y - x; s *= 2) {
             for (ptrdiff_t w = s / 2; w >= 1; w /= 2) {
-                bitonic_sort_step<T>(a, x, y, i, s, w);
+                bitonic_sort_step< T >(a, x, y, i, s, w);
             }
         }
     }
@@ -303,7 +299,7 @@ def _partition_kernel(dtype):
         // After merge step, the first k elements are already bitonic.
         // Therefore, we do not need to fully sort.
         for (int w = k / 2; w >= 1; w /= 2) {
-            bitonic_sort_step<T>(a, x, k + x, i, k, w);
+            bitonic_sort_step< T >(a, x, k + x, i, k, w);
         }
     }
 
@@ -323,7 +319,7 @@ def _partition_kernel(dtype):
         int id = i % 32;
         int x = 0;
 
-        bitonic_sort<${dtype}>(a, z, k + z, id);
+        bitonic_sort< ${dtype} >(a, z, k + z, id);
         ptrdiff_t j;
         for (j = k + id + z; j < m - (m - z) % 32; j += 32) {
             if (a[j] < a[k - 1 + z]) {
@@ -340,8 +336,8 @@ def _partition_kernel(dtype):
     #else
             if (__any(x >= t)) {
     #endif
-                bitonic_sort<${dtype}>(a, k + z, 32 * t + k + z, id);
-                merge<${dtype}>(a, k, id, z, k + z, min(k, 32 * t));
+                bitonic_sort< ${dtype} >(a, k + z, 32 * t + k + z, id);
+                merge< ${dtype} >(a, k, id, z, k + z, min(k, 32 * t));
                 x = 0;
             }
         }
@@ -353,8 +349,8 @@ def _partition_kernel(dtype):
 
         // Finally, we merge the first k elements and the remainders to be
         // stored.
-        bitonic_sort<${dtype}>(a, k + z, 32 * t + k + z, id);
-        merge<${dtype}>(a, k, id, z, k + z, min(k, 32 * t));
+        bitonic_sort< ${dtype} >(a, k + z, 32 * t + k + z, id);
+        merge< ${dtype} >(a, k, id, z, k + z, min(k, 32 * t));
     }
 
     __global__ void ${merge_kernel}(
@@ -364,7 +360,7 @@ def _partition_kernel(dtype):
         ptrdiff_t z = i / 32 * 2 * s * n / sz;
         ptrdiff_t m = (i / 32 * 2 + 1) * s * n / sz;
         int id = i % 32;
-        merge<${dtype}>(a, k, id, z, m, k);
+        merge< ${dtype} >(a, k, id, z, m, k);
     }
     }
     ''').substitute(name=name, merge_kernel=merge_kernel, dtype=dtype)
--- a/cupy/cuda/common.pxd
+++ b/cupy/cuda/common.pxd
@@ -11,3 +11,6 @@ cdef extern from '../cuda/cupy_common.h':  # thru parent to import in core
     ctypedef unsigned long long cpy_ulong
     ctypedef float cpy_float
     ctypedef double cpy_double
+    ctypedef struct cpy_complex64 'cuComplex'
+    ctypedef struct cpy_complex128 'cuDoubleComplex'
+    ctypedef struct cpy_half '__half'
--- a/cupy/cuda/common.pxd
+++ b/cupy/cuda/common.pxd
@@ -1,5 +1,7 @@
 # distutils: language = c++
 
+from libcpp cimport bool as bool_t
+
 cdef extern from '../cuda/cupy_common.h':  # thru parent to import in core
     ctypedef char cpy_byte
     ctypedef unsigned char cpy_ubyte
@@ -13,4 +15,5 @@ cdef extern from '../cuda/cupy_common.h':  # thru parent to import in core
     ctypedef double cpy_double
     ctypedef struct cpy_complex64 'cuComplex'
     ctypedef struct cpy_complex128 'cuDoubleComplex'
+    ctypedef bool_t cpy_bool 'bool'
     ctypedef struct cpy_half '__half'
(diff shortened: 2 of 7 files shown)


leofang:
Will think about how to fix Travis tomorrow...😞 Likely I forgot a `CUPY_NO_CUDA` somewhere...

emcastillo:
Jenkins, test this please.

CIs kicked to verify the implementation does not error.

leofang:
Sorry, force pushed to fix the build error (still need to fix Travis). 😅 @emcastillo could you rerun CI?

cupy/cuda/thrust.pyx
```diff
@@ -8,7 +8,9 @@ cimport cython  # NOQA
 from libcpp cimport vector
 
 from cupy.cuda cimport common
+from cupy.cuda cimport device
 from cupy.cuda cimport memory
+from cupy.cuda cimport runtime
```
grlee77, line 13: This cimport of runtime seems to be what is causing the failure when building with `--cupy-no-cuda`.

grlee77, line 13:
```suggestion
from cupy.cuda import runtime
```

grlee77: I don't fully understand why, but switching from `cimport` to `import` here resolved the issue for me.

leofang: Thanks, Greg! It's indeed confusing...By inspecting the imported C headers (`cupy_common.h` etc) I also thought the problem was this line, so I dug into the C preprocessor output, and oddly all macro guards were set up right but the includes just didn't get expanded: 
```C++
...
# 658 "cupy/cuda/thrust.cpp" 2
# 1 "cupy/cuda/../cuda/cupy_cuda.h" 1
# 54 "cupy/cuda/../cuda/cupy_cuda.h"
# 1 "cupy/cuda/../cuda/cupy_cuda_common.h" 1  // <- not OK: nothing expanded!
# 55 "cupy/cuda/../cuda/cupy_cuda.h" 2  // <- OK, cupy_cuComplex.h is included earlier



# 57 "cupy/cuda/../cuda/cupy_cuda.h"
extern "C" {

bool hip_environment = false;






CUresult cuGetErrorName(...) {
    return CUDA_SUCCESS;
}
...
```

I also found that the order of `cimport common` and `cimport runtime` matters. By switching them I got different errors. Usually this means there's some circular import at the C level, but I don't immediately see where the circularity is...

leofang: The guard name `INCLUDE_GUARD_CUPY_CUDA_COMMON_H` is reused in two headers 🤬

leofang: e673795 fixed it.

leofang:
@takagi e673795 fixed the build error with or without `CUPY_NO_CUDA`. I will start looking into possible performance implications. In the meanwhile, could you kick off the CI to see if I introduced bugs? (I only tested `test_sort.py` locally.) Thanks!

cupy/core/_routines_sorting.pyx
```diff
@@ -282,7 +278,7 @@ def _partition_kernel(dtype):
             CArray<T, 1, true> a, ptrdiff_t x, ptrdiff_t y, int i) {
         for (ptrdiff_t s = 2; s <= y - x; s *= 2) {
             for (ptrdiff_t w = s / 2; w >= 1; w /= 2) {
-                bitonic_sort_step<T>(a, x, y, i, s, w);
+                bitonic_sort_step< T >(a, x, y, i, s, w);
```
emcastillo, line 281: Any reason behind this changes?😇

cupy/cuda/common.pxd
```diff
@@ -1,5 +1,7 @@
 # distutils: language = c++
 
+from libcpp cimport bool as bool_t
```
emcastillo, line 3: Nit: I am not sure if `bool_t` is used in other places of the code base or `bint` is favored over it, but we should stick to one convention if possible.

cupy/cuda/thrust.pyx
emcastillo: I am amazed it managed to compile until now, lol

emcastillo:
Jenkins, test this please

cupy/core/_routines_sorting.pyx
leofang: This is to specialize for `bitonic_sort_step< complex<float> >` and friend. Without the spacing  the compilation would fail (in a way that the test suite couldn't report).

cupy/cuda/common.pxd
leofang: Yeah I know I said `bint` is favored in Cython 😅, but this is for template specialization for C++ `bool`. I am not sure if `bint` is always translated to `bool`, so tried to be on the safe side. Perhaps I could rename it to `cpp_bool` for style consistency? See
https://github.com/cupy/cupy/blob/0c841101dbbc0afc59f7dd340d8ba66a39795146/cupy/core/internal.pxd#L1

leofang:
> I am amazed it managed to compile until now, lol

If you meant the bug caused by the repetitive guard name, yeah it's surprising until you figure out `cupy_common.h` and `common.pxd` is only included in the thrust module...

One thing we could do next, if `cupy_common.h` is meant for type aliasing, is to move this part to it: https://github.com/cupy/cupy/blob/9d8dca77507e47bee62a2ee0ba1da4f67874e3be/cupy/cuda/cupy_cub.h#L4-L16
This will make the alias more reusable across modules.

OK I also need to fix Appveyor now 😞 I suspect we should not specialize for `__half` on host...Will wait for the CI to finish and push a test commit.

leofang:
before:
- compile time:
real	1m48.061s
user	1m44.241s
sys	0m3.883s
(`cupy/core/_routines_sorting.pyx` and `cupy/cuda/thrust.pyx` only; set `CUPY_NVCC_GENERATE_CODE=arch=compute_75,code=sm_75`)
- performance:
```python
>>> import cupy as cp
>>>
>>> # 1M elements
>>> def f():
...     a = cp.random.random((1000000,))
...     a.sort()
... 
>>> print(repeat(f, ()))
f                   :    CPU:   91.497 us   +/- 5.006 (min:   87.961 / max:  223.452) us     GPU:  961.748 us   +/-20.848 (min:  948.576 / max: 1155.328) us
>>> 
>>> # 10M elements
>>> def f():
...     a = cp.random.random((10000000,))
...     a.sort()
...     del a
... 
>>> print(repeat(f, (), n=100))
f                   :    CPU:   94.498 us   +/- 4.014 (min:   90.030 / max:  112.589) us     GPU: 6645.967 us   +/- 7.638 (min: 6629.408 / max: 6671.296) us
```

This PR:
- compile time: 
real	2m4.182s
user	2m0.274s
sys	0m3.941s
- performance:
```python
>>> import cupy as cp
>>>
>>> # 1M elements
>>> def f():
...     a = cp.random.random((1000000,))
...     a.sort()
... 
>>> print(repeat(f, ()))
f                   :    CPU:   92.628 us   +/- 4.465 (min:   88.703 / max:  208.141) us     GPU:  965.287 us   +/-18.920 (min:  951.872 / max: 1152.832) us
>>>
>>> # 10M elements
>>> def f():
...     a = cp.random.random((10000000,))
...     a.sort()
...     del a
... 
>>> print(repeat(f, (), n=100))
f                   :    CPU:   94.090 us   +/- 5.719 (min:   89.509 / max:  121.101) us     GPU: 6631.661 us   +/- 8.929 (min: 6612.512 / max: 6660.576) us
```

leofang:
@emcastillo @takagi I feel the right way to fix the compatibility with older CUDA and to increase reusability is to avoid exposing `__half` to the Cython space. I think it can be done by moving this part
https://github.com/cupy/cupy/blob/e339e1f5513afc0922b406e6e721c86d77d42a23/cupy/cuda/cupy_cub.cu#L407-L436
to `cupy_common.h`, and do the type dispatch in the C++ space. May I proceed with this?

leofang:
...I feel I might have missed something obvious? How is fp16 supported in CuPy?

leofang:
I worked around the issue of exposing `__half` to Cython, as I don't wanna do refactoring in this PR 😅 Also, I added a few more tests with the presence of NaN, which should have been added in my earlier PR #2745.

tests/cupy_tests/sorting_tests/test_sort.py
```diff
@@ -190,17 +188,23 @@ def test_lexsort_three_or_more_dim(self):
...
 
-    @testing.for_dtypes([numpy.float16, numpy.bool_])
-    def test_lexsort_unsupported_dtype(self, dtype):
-        a = testing.shaped_random((2, 10), cupy, dtype)
-        with self.assertRaises(TypeError):
-            return cupy.lexsort(a)
+    # Test NaN ordering
+    # TODO(leofang): test two more cases after #3232 is fixed
+    # 1. a[1, 2] = a[1, 6] = xp.nan
+    # 2. a[0, 2] = a[1, 6] = xp.nan
```
leofang, line 200: Ref: https://github.com/cupy/cupy/issues/3232#issuecomment-615016490

leofang:
Appveyor is happy!

takagi:
pfnCI, test this please.

cupy/cuda/common.pxd
```diff
@@ -11,3 +13,6 @@ cdef extern from '../cuda/cupy_common.h':  # thru parent to import in core
     ctypedef unsigned long long cpy_ulong
     ctypedef float cpy_float
     ctypedef double cpy_double
+    ctypedef struct cpy_complex64 'cuComplex'
+    ctypedef struct cpy_complex128 'cuDoubleComplex'
+    ctypedef bool_t cpy_bool 'bool'
```
takagi, line 18: Adding `cdef struct` and `ctypedef` as following would be better?

```Python
cdef extern from '../cuda/cupy_common.h':  # thru parent to import in core
    cdef struct cuComplex
    cdef struct cuDoubleComplex
    ...
    ctypedef cuComplex cpy_complex64
    ctypedef cuDoubleComplex cpy_complex128
    ctypedef bool_t cpy_bool
```

takagi:
LGTM except for a comment above!

cupy/cuda/common.pxd
leofang: Thanks, @takagi! Question: Aren't cuComplex and cuDoubleComplex imported from cupy_common.h? I think based on Cython's doc this usage is legit (see the "or" part in the 3rd row):
https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#styles-of-struct-union-and-enum-declaration

takagi: I thought C preprocesser would finally expand `#include cupy_cuComplex.h` to the declarations of `cuComplex` and `cuDoubleComplex`, so I could import those here. But, yes, now I understand the current code is valid. You don't need to make a change.

takagi:
Thanks!

leofang:
Thanks, @takagi 🙂

leofang:
btw this will go into v8.0.0b2, right?

PR #3290, opened 2020-04-17
Expose cuda IPC runtime calls

emcastillo:
This PR adds support for the `cudaIpc` calls in the API
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html

A simple example of how to use them:

```python
import multiprocessing as mp

import cupy


def child(handler, shape, dtype):
    import cupy

    data_ptr = cupy.cuda.runtime.ipcOpenMemHandle(handler)
    buff = {"data": (int(data_ptr), False), "shape": shape, "typestr": dtype}

    class capsule(object):
        pass

    cap = capsule()
    cap.__cuda_array_interface__ = buff
    a = cupy.asanyarray(cap)
    print(a)


if __name__ == "__main__":
    mp.set_start_method("spawn")
    a = cupy.arange(10)
    handler = cupy.cuda.runtime.ipcGetMemHandle(a.data.ptr)
    p = mp.Process(target=child, args=(handler, a.shape, str(a.dtype)))
    p.start()
    p.join()
```
The example is based in one from https://github.com/pytorch/pytorch/issues/36594
TODO: add tests?

### DIFF
--- a/cupy_backends/cuda/api/runtime.pxd
+++ b/cupy_backends/cuda/api/runtime.pxd
@@ -95,6 +95,13 @@ cdef extern from *:
         # TODO(leofang): support mipmap?
 
 
+    ctypedef struct IpcMemHandle 'cudaIpcMemHandle_t':
+        unsigned char[64] reserved
+
+    ctypedef struct IpcEventHandle 'cudaIpcEventHandle_t':
+        unsigned char[64] reserved
+
+
 ###############################################################################
 # Enum
 ###############################################################################
@@ -108,6 +115,8 @@ cpdef enum:
 
     cudaMemoryTypeHost = 1
     cudaMemoryTypeDevice = 2
+
+    cudaIpcMemLazyEnablePeerAccess = 1
 
     cudaMemAttachGlobal = 1
     cudaMemAttachHost = 2
--- a/cupy_backends/cuda/api/runtime.pxd
+++ b/cupy_backends/cuda/api/runtime.pxd
@@ -94,7 +94,6 @@ cdef extern from *:
         unsigned int maxAnisotropy
         # TODO(leofang): support mipmap?
 
-
     ctypedef struct IpcMemHandle 'cudaIpcMemHandle_t':
         unsigned char[64] reserved
 
--- a/cupy_backends/cuda/api/runtime.pyx
+++ b/cupy_backends/cuda/api/runtime.pyx
@@ -95,6 +95,14 @@ cdef extern from '../cupy_cuda.h' nogil:
     int cudaDeviceGetLimit(size_t* value, Limit limit)
     int cudaDeviceSetLimit(Limit limit, size_t value)
 
+    # IPC
+    int cudaIpcCloseMemHandle(void* devPtr)
+    int cudaIpcGetEventHandle(IpcEventHandle* handle, driver.Event event)
+    int cudaIpcGetMemHandle(IpcMemHandle*, void* devPtr)
+    int cudaIpcOpenEventHandle(driver.Event* event, IpcEventHandle handle)
+    int cudaIpcOpenMemHandle(void** devPtr, IpcMemHandle handle,
+                             unsigned int  flags)
+
     # Memory management
     int cudaMalloc(void** devPtr, size_t size)
     int cudaMallocManaged(void** devPtr, size_t size, unsigned int flags)
@@ -325,6 +333,57 @@ cpdef size_t deviceGetLimit(int limit) except? -1:
 cpdef deviceSetLimit(int limit, size_t value):
     status = cudaDeviceSetLimit(<Limit>limit, value)
     check_status(status)
+
+
+###############################################################################
+# IPC operations
+###############################################################################
+cpdef ipcCloseMemHandle(intptr_t devPtr):
+    status = cudaIpcCloseMemHandle(<void*>devPtr)
+    check_status(status)
+
+
+cpdef ipcGetEventHandle(intptr_t event):
+    cdef IpcEventHandle handle
+    status = cudaIpcGetEventHandle(&handle, <driver.Event>event)
+    check_status(status)
+    # We need to do this due to a bug in Cython that
+    # cuts out the 0 bytes in an array of chars when
+    # constructing the python object
+    # resulting in different sizes assignment errors
+    # when recreating the struct from the python
+    # array of bytes
+    reserved = [<unsigned char>handle.reserved[i] for i in range(64)]
+    return bytes(reserved)
+
+cpdef ipcGetMemHandle(intptr_t devPtr):
+    cdef IpcMemHandle handle
+    status = cudaIpcGetMemHandle(&handle, <void*>devPtr)
+    check_status(status)
+    # We need to do this due to a bug in Cython that
+    # when converting an array of chars in C to a python object
+    # it discards the data after the first 0 value
+    # resulting in a loss of data, as this is not a string
+    # but a buffer of bytes
+    reserved = [<unsigned char>handle.reserved[i] for i in range(64)]
+    return bytes(reserved)
+
+cpdef ipcOpenEventHandle(bytes handle):
+    cdef driver.Event event
+    cdef IpcEventHandle handle_
+    handle_.reserved = handle
+    status = cudaIpcOpenEventHandle(&event, handle_)
+    check_status(status)
+    return <intptr_t>event
+
+cpdef ipcOpenMemHandle(bytes handle,
+                       unsigned int flags=cudaIpcMemLazyEnablePeerAccess):
+    cdef void* devPtr
+    cdef IpcMemHandle handle_
+    handle_.reserved = handle
+    status = cudaIpcOpenMemHandle(&devPtr, handle_, flags)
+    check_status(status)
+    return <intptr_t>devPtr
 
 
 ###############################################################################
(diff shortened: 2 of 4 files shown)


cupy/cuda/driver.pxd
```diff
@@ -23,7 +23,10 @@ cdef extern from *:
     ctypedef int CUfunction_attribute 'CUfunction_attribute'
 
     ctypedef size_t(*CUoccupancyB2DSize)(int)
-
+    ctypedef struct IpcMemHandle 'cudaIpcMemHandle_t':
+        unsigned char[64] reserved
+    ctypedef struct IpcEventHandle 'cudaIpcEventHandle_t':
+        unsigned char[64] reserved
```
leofang, line 29: 1. Shouldn't these two belong to `runtime.pxd`?
2. Do you have references to the struct declarations? Oddly the official doc is empty for these two 🧐
3. Need an empty line below

cupy/cuda/runtime.pyx
```diff
@@ -306,6 +314,61 @@ cpdef deviceEnablePeerAccess(int peerDevice):
...
+    cdef driver.IpcEventHandle handle
+    status = cudaIpcGetEventHandle(&handle, <driver.Event>event)
+    check_status(status)
+    # We need to do this due to a bug in Cython that
+    # cuts out the 0 bytes in an array of chars when
+    # constructing the python object
+    # resulting in different sizes assignment errors
+    # when recreating the struct from the python 
+    # array of bytes
+    reserved = [<unsigned char>handle.reserved[i] for i in range(64)]
```
leofang, line 335: Just curious, do you have a pointer to the Cython issue?

cupy/cuda/runtime.pyx
```diff
@@ -306,6 +314,61 @@ cpdef deviceEnablePeerAccess(int peerDevice):
...
+cpdef ipcGetMemHandle(intptr_t devPtr):
+    cdef driver.IpcMemHandle handle
+    status = cudaIpcGetMemHandle(&handle, <void*>devPtr)
+    check_status(status)
+    # We need to do this due to a bug in Cython that
+    # when converting an array of chars in C to a python object
+    # it discards the data after the first 0 value
+    # resulting in a loss of data, as this is not a string
+    # but a buffer of bytes
+    reserved = [<unsigned char>handle.reserved[i] for i in range(64)]
```
leofang, line 348: ditto

cupy/cuda/driver.pxd
emcastillo: I didn't found a source with the fields, so I looked into the cuda headers myself, as the structure is passed by value to some calls instead of having a pointer.
its defined with only an array of 64 chars

cupy/cuda/runtime.pyx
emcastillo: Nope, I guess its because cython treats char* arrays as strings instead of buffers of bytes, so it automatically throws away all the data after the first 0 when converting it to a python object.

cupy/cuda/cupy_cuda.h
```diff
@@ -269,6 +269,26 @@ cudaError_t cudaDeviceEnablePeerAccess(...) {
     return cudaSuccess;
 }
```
leofang, line 271: Need to also add stubs for `cudaIpcMemHandle_t` and `cudaIpcEventHandle_t` here.

cupy/cuda/runtime.pyx
```diff
@@ -306,6 +314,61 @@ cpdef deviceEnablePeerAccess(int peerDevice):
...
+    return bytes(reserved)
+
+
+cpdef ipcOpenEventHandle(bytes handle):
+    cdef driver.Event event
+    cdef IpcEventHandle handle_
+    handle_.reserved = handle
+    status = cudaIpcOpenEventHandle(&event, handle)
```
leofang, line 356:
```suggestion
    status = cudaIpcOpenEventHandle(&event, handle_)
```

cupy/cuda/cupy_cuda.h
leofang: (may be also `cudaIpcMemLazyEnablePeerAccess`?)

cupy/cuda/driver.pxd
leofang: oh oh I see. This is what I found in `/usr/local/cuda/include/driver_types.h`:
```C
/**
 * CUDA IPC Handle Size
 */
#define CUDA_IPC_HANDLE_SIZE 64

/**
 * CUDA IPC event handle
 */
typedef __device_builtin__ struct __device_builtin__ cudaIpcEventHandle_st
{
    char reserved[CUDA_IPC_HANDLE_SIZE];
}cudaIpcEventHandle_t;

/**
 * CUDA IPC memory handle
 */
typedef __device_builtin__ struct __device_builtin__ cudaIpcMemHandle_st
{
    char reserved[CUDA_IPC_HANDLE_SIZE];
}cudaIpcMemHandle_t;
```
Seems like we can just use `char` instead of `unsigned char`?

emcastillo: If we use char, when we convert to bytes it complains because some of the values are < 0, so I just went ahead by making it unsigned char in the definition. 
I am fine with doing explicit castings but I thought this was more readable and shouldn't affect binary compat.

cupy/cuda/runtime.pyx
leofang: You're right...Cython has a horribly long doc on string processing which mentioned this: 
> The above will therefore only work correctly for C strings that do not contain null bytes.

https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#passing-byte-strings

emcastillo:
Travis checks passed! Thanks @leofang 😃

cupy/cuda/runtime.pyx
emcastillo: I discovered it the hard way 😇

leofang: ...and I would never google for it until someone told me this fact 😅

cupy/cuda/driver.pxd
leofang: I see, I didn't know negative values could be a problem for `bytes`...

cupy/cuda/runtime.pyx
leofang: By the way, despite a Python list (`reserved = [...]`) is used to hold the 64 bytes, it's actually as performant as a C++ vector. I was very surprised to learn it...

emcastillo: Do you have some link to this? 🙂

leofang: No, I tested it myself:
```cython
# distutils: language = c++

from libcpp cimport vector
from libc.string cimport memcpy

cdef int i
cdef char[64] data = [<char>i for i in range(64)]


cpdef f():
    cdef vector.vector[char] reserved
    cdef int i
    reserved.resize(64)

    memcpy(reserved.data(), data, 64)
    return bytes(reserved)


cpdef f2():
    cdef vector.vector[char] reserved = vector.vector[char](64)
    cdef int i

    memcpy(reserved.data(), data, 64)
    return bytes(reserved)


cpdef g():
    cdef list reserved = [<char>(data[i]) for i in range(64)]
    return bytes(reserved)
```
```python
In [1]: from test_vector_to_bytes import f, f2, g                                                                                                                      

In [2]: %timeit f()                                                                                                                                                    
831 ns ± 1.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit f2()                                                                                                                                                   
846 ns ± 0.859 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit g()                                                                                                                                                    
817 ns ± 2.66 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: f() == f2() == g()                                                                                                                                             
Out[5]: True
```

cupy/cuda/cupy_cuda.h
```diff
@@ -269,6 +269,35 @@ cudaError_t cudaDeviceEnablePeerAccess(...) {
...
 
+// IPC operations
+typedef struct {
+    unsigned char reserved[64];
+} cudaIpcMemHandle_t;
+
+// IPC operations
+typedef struct {
+    unsigned char reserved[64];
+} cudaIpcEventHandle_t;
```
leofang, line 280: Could you check if the structs should be put here or in `cupy_cuda_common.h`?

leofang:
@emcastillo I might find some use cases for this feature. Any chance we can get it merged? 😇

kmaehashi:
Let's merge after the conflict has been resolved.

emcastillo:
Rebased

kmaehashi:
pfnCI, test this please.

cupy_backends/cuda/api/runtime.pxd
```diff
@@ -95,6 +95,13 @@ cdef extern from *:
         # TODO(leofang): support mipmap?
```
leofang, line 97:
```suggestion
```

leofang:
I confirm that the script in https://github.com/cupy/cupy/pull/3290#issue-404898127 still works out of box, and that `cuda-memcheck` reports no memory leak.

leofang:
Jenkins, test this please

leofang:
@kmaehashi For some reason Appveyor did not run, but this time both `cupy-doc` and `chainer-doc` are happy (they were the only two failures in the previous run), so I think it's OK?

leofang:
Appveyor passed! @kmaehashi did you re-trigger it? 😇 Thanks!

PR #3291, opened 2020-04-19
Code fix + bug fix for `cupy.cuda.thrust`

### LINKED ISSUE #3232: Lexsort output is incorrect

leofang:
Close #3232, in which two different bugs were reported. Follow-up of #2745 & #3286.  

This PR does two things:
1. code fix: most pointers passed from Cython are of type `intptr_t` (#2455). Shouldn't have to cast them to `size_t`.
2. bug fix: #3232 has two bugs reported:
    - `cupy::thrust::_lexsort()` assumes a C-contiguous input but we didn't check it
    - lack of NaN handling for real numbers 

Context for the 2nd bug: In #2745 (#3286) when sorting complex (half precision) numbers was supported, NaN handling was added, but I forgot to extend it to single and double precisions 😅 Now this is done in a unified fashion, similar to how it's handled in `cupy/cuda/cupy_cub.cu`. (Perhaps I will do more refactoring later to reduce code repetition.) The master branch fails most of the added tests (for NaN and non-C-contiguous arrays), which this PR passes.

### DIFF
--- a/cupy/_sorting/sort.py
+++ b/cupy/_sorting/sort.py
@@ -73,6 +73,10 @@ def lexsort(keys):
         raise NotImplementedError('Keys with the rank of three or more is not '
                                   'supported in lexsort')
 
+    # thrust.lexsort() assumes a C-contiguous array
+    if not keys.flags.c_contiguous:
+        keys = keys.copy('C')
+
     idx_array = cupy.ndarray(keys._shape[1:], dtype=numpy.intp)
     k = keys._shape[0]
     n = keys._shape[1]
@@ -120,7 +124,6 @@ def msort(a):
 
     """
 
-    # TODO(takagi): Support float16 and bool.
     return sort(a, axis=0)
 
 
--- a/cupy/cuda/cupy_thrust.cu
+++ b/cupy/cuda/cupy_thrust.cu
@@ -135,10 +135,10 @@ __host__ __device__ __forceinline__ bool _real_less(const T& lhs, const T& rhs) 
 template <typename T>
 __host__ __device__ __forceinline__ bool _tuple_real_less(const tuple<size_t, T>& lhs,
                                                           const tuple<size_t, T>& rhs) {
-    const size_t& lhs_k = lhs.get<0>();
-    const size_t& rhs_k = rhs.get<0>();
-    const T& lhs_v = lhs.get<1>();
-    const T& rhs_v = rhs.get<1>();
+    const size_t& lhs_k = lhs.template get<0>();
+    const size_t& rhs_k = rhs.template get<0>();
+    const T& lhs_v = lhs.template get<1>();
+    const T& rhs_v = rhs.template get<1>();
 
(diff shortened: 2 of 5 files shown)


leofang:
@takagi The PR description is updated. PTAL.

takagi:
Thanks! pfnCI, test this please.

takagi:
Jenkins CI fails because of strange compiler errors in cupy_thrust.cu: https://jenkins.preferred.jp/job/chainer/job/cupy_pr/1212/TEST=chainer-py3,label=mn1-p100/console

Would you take a look into it?

leofang:
@takagi Hmm, what's the gcc and CUDA versions in chainer-py3?

leofang:
It's CUDA 10.1. I guess gcc is 7? Not sure...

Anyway, I guess it's because `tuple.get<i>()` is a dependent name, see "disambiguator for dependent names" [here](http://en.cppreference.com/w/cpp/language/dependent_name). We could either try `thrust::get` or add a `typename`.

leofang:
But then I don't know why other CIs and my local build passed...😞

takagi:
`chainer-py3` uses CUDA 10.1 and `ubuntu:18.04` docker image as its base. Other test patterns use different base image and CUDA combinations, which may be the reason other CIs pass. I'll check into the detail.

https://github.com/chainer/chainer-test/blob/master/run_test.py#L108

leofang:
Thanks, @takagi! In the meanwhile, could you give 556e486 a try?

takagi:
pfnCI, test this please.

leofang:
`ubuntu:18.04` is used in several CIs, so looks like Thrust that comes with CUDA 10.1 caused the issue...It's now fixed in 556e486.

takagi:
Thanks, LGTM!

leofang:
Thanks for quick review, @takagi :)

takagi:
I'm working on backporting this PR.

PR #3197, opened 2020-03-16
20200316

Fokatu:
Sorry. Wrong operation

PR #2432, opened 2019-08-27
Support texture memory in `RawKernel`

leofang:
Closes #1811, closes #2278. Although this PR already works (see an example at the bottom), I mark it as WIP for gathering any early feedbacks and comments. ~~Also, I have not yet added tests and docs.~~ Any help for polishing this PR is very welcome and appreciated!

## Goal
To provide basic support of texture memory (via "CUDA arrays" or even plain CuPy arrays) in `RawKernel`. 

## Implementation decisions and notes
- CUDA arrays (`cudaArray_t`) are supported in the new class `cupy.cuda.texture.CUDAArray`. 1D, 2D, 3D arrays are all supported. Data can be copied from either CuPy or NumPy arrays to `CUDAArray` and vice versa, if the datatype and shape match between source and destination.

For the time being I have no thoughts how `CUDAArray` can fit better with the existing CuPy ndarrays. It'd be awesome if CuPy ndarrays can be backed by texture memory (which in turn is backed by `CUDAArray`) for readonly purposes, but I don't know how to achieve that. The main technical obstacle is that `cudaArray_t` is an opaque data type, and it's not possible to do simple pointer arithmetics as we do with linear memories (allocated through `cudaMalloc()` etc). Another thing is that I don't know CuPy internal well enough 😛 

- The texture object runtime API is exposed. The end product is the `cupy.cuda.texture.TextureObject` class, which can be used as an argument of `RawKernel`.

- `cudaResourceTypeArray` is tested (see below), meaning `CUDAArray` can back a `TextureObject` as expected.

- mipmap is ***not*** supported in this PR.

- The user interface mostly follows its CUDA counterpart. Not sure if a better, perhaps more Pythonic public API is possible.

- ~~The texture reference API is marked deprecated in the CUDA Toolkit docs, so I didn't add them.~~

But if we end up supporting it, we should use the driver API instead of the runtime API. The reason is that `cuModuleGetTexRef()`, a critical function for supporting texture reference, handles `CUtexref*`, which doesn't seem possible to be converted to runtime's `textureReference*` (at least it's not documented). <----***UPDATE:*** I confirm that they are indeed incompatible.

- ***UPDATE:*** The texture reference driver API is exposed. The end product is the `cupy.cuda.texture.TextureReference` class, which can bind a CUDA array or a CuPy array to a texture reference returned by `cupy.RawModule.get_texref()`.

## Example (updated)
The following example illustrates what this PR enables:
```python
import cupy as cp


source=r'''
extern "C"{
[...]

PR #3294, opened 2020-04-20
Support surface memory in `RawKernel`

### LINKED ISSUE #2624: CuPy surface support
I saw that Cupy have added support to texture memory. 
Is it possiable to add surface support, such that one can read and write a cudaArray at the same time.

leofang:
Takeover from #3197. **Main credit goes to @Fokatu**. Close #2624. 

This is a follow-up of #2432. The surface object API is supported. Although I don't think there're many users who need this capability (compared to texture memory), @Fokatu has shown it's simple to support this in CuPy with little maintenance burden, so I think it's still worth adding. 

With this support, CuPy is now a full (and far better!) replacement for PyCUDA.

### DIFF
--- a/cupy/cuda/runtime.pxd
+++ b/cupy/cuda/runtime.pxd
@@ -21,6 +21,7 @@ cdef extern from *:
         int x, y, z, w
         ChannelFormatKind f
     ctypedef uintmax_t TextureObject 'cudaTextureObject_t'
+    ctypedef uintmax_t SurfaceObject 'cudaSurfaceObject_t'
     ctypedef int ResourceType 'cudaResourceType'
     ctypedef int TextureAddressMode 'cudaTextureAddressMode'
     ctypedef int TextureFilterMode 'cudaTextureFilterMode'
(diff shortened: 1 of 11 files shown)


cupy/cuda/runtime.pxd
```diff
@@ -249,6 +250,13 @@ cpdef enum:
     cudaChannelFormatKindFloat = 2
     cudaChannelFormatKindNone = 3
 
+    # CUDA array flags
+    cudaArrayDefault = 0
+    # cudaArrayLayered = 1
+    cudaArraySurfaceLoadStore = 2
```
leofang, line 256: `cuda.runtime.cudaArrayDefault` was actually referred in `CUDAarray`'s docstring but we never added it, plus `cudaArraySurfaceLoadStore` is needed for surface memory, so I add them here.

leofang:
Hi @Fokatu, sorry I had to squash your original commits to a33e9ab 😅 I wasn't able to resolve conflicts when applying your patch onto master as is, plus the original commit messages were not meaningful...

emcastillo:
Jenkins, test this please

leofang:
Thanks!

PR #3310, opened 2020-04-27
Alias of shape type

asi1024:
Declare `shape_t` and `strides_t` as aliases of `vector.vector[Py_ssize_t]` for the readability.


emcastillo:
Jenkins, test this please

PR #3311, opened 2020-04-27
Fix scipy version check in cupyx.scipy.fft

grlee77:
It seems that the current code in `cupyx/scipy/fft/fft.py` that checks for SciPy >= 1.5.0 will always evaluate to False. 

It seems `from _scipy_fft._lib._pep440 import Version` will always fail because the `Version` function lives in  `scipy._lib._pep440`. Also, should we really be importing anything from a private `_lib` folder? (We could use an alternative like `numpy.lib.NumpyVersion` or distutils `LooseVersion` instead)

@peterbell10, can you verify if this fix looks correct to you?

### DIFF
--- a/cupyx/scipy/fft/fft.py
+++ b/cupyx/scipy/fft/fft.py
@@ -17,7 +17,9 @@ __all__ = ['fft', 'ifft', 'fft2', 'ifft2', 'fftn', 'ifftn',
            'fftshift', 'ifftshift', 'fftfreq', 'rfftfreq',
            'get_fft_plan']
 
+_scipy_150 = False
 try:
+    import scipy
     import scipy.fft as _scipy_fft
 except ImportError:
     class _DummyModule:
@@ -25,15 +27,11 @@ except ImportError:
             return None
 
     _scipy_fft = _DummyModule()
-
-try:
-    from _scipy_fft._lib._pep440 import Version
-except ImportError:
-    # either _scipy_fft is _DummyModule, or scipy < 1.4.0
-    _scipy_150 = False
 else:
-    _scipy_150 = Version(_scipy_fft.version.version) >= Version('1.5.0')
+    from numpy.lib import NumpyVersion as Version
+    _scipy_150 = Version(scipy.__version__) >= Version('1.5.0')
     del Version
+    del scipy
 
 # Backend support for scipy.fft
 


peterbell10:
This looks like it should work. Though I agree that `scipy._lib` is a private subpackage so CuPy shouldn't be using it. Looks like this was added by @leofang in #2998.

leofang:
Good catch, thanks @grlee77. I don't remember why I used `_scipy_fft._lib._pep440`, now that I read it again it's clearly a typo...😢

Thanks for the alternative solutions. `numpy.lib.NumpyVersion` seems legit. `LooseVersion` does not, because it doesn't support dev versions like `'1.5.0.dev0+91b4020'`. Would you like to switch to that?

Just curious, though, I remember I read it somewhere that modules under `scipy._lib` is considered stable. Is it not the case? I couldn't locate a pointer to this statement for now, though.

peterbell10:
`scipy._lib` isn't listed as public in [the docs](https://docs.scipy.org/doc/scipy-1.4.1/reference/api.html#api-definition). Maybe you're thinking of `numpy.lib`?

leofang:
Yeah I don't remember...Anyway, @peterbell10 I just recalled your earlier PR (scipy/scipy#11290). I thought it meant `scipy._lib._pep440.Version` is preferred over `NumpyVersion`? Or it doesn't matter to CuPy?

peterbell10:
`NumpyVersion` doesn't implement all of PEP440 but does include the subset used by NumPy and SciPy which is what matters here. The main motivation for removing `NumpyVersion` from SciPy is just avoiding having two ways to do the same thing.

grlee77:
Thanks Peter. I have updated to just use `NumpyVersion` here.

cupyx/scipy/fft/fft.py
```diff
@@ -25,15 +26,15 @@ def __getattr__(self, name):
...
-try:
-    from _scipy_fft._lib._pep440 import Version
-except ImportError:
-    # either _scipy_fft is _DummyModule, or scipy < 1.4.0
-    _scipy_150 = False
 else:
-    _scipy_150 = Version(_scipy_fft.version.version) >= Version('1.5.0')
-    del Version
+    try:
+        from numpy.lib import NumpyVersion as Version
```
toslunar, line 31: nit: The line could be outside of the `try` block.  How about the following to handle both `_scipy_fft` and `_scipy_150`?
```python
try:
    import scipy
    import scipy.fft as _scipy_fft
except ImportError:
    ...
else:
    from numpy.lib import NumpyVersion as Version
    ...
```

grlee77: done. please take a look again

kmaehashi:
pfnCI, test this please.

kmaehashi:
Backport is not needed as this flag was introduced in v8.0.0a1 (#2998)

PR #3147, opened 2020-03-03
Fix gencode example in the doc

leofang:

Issue #3075, opened 2020-02-13
Configuration missing for cupy.cuda.cub extension


benjha:
Hi CuPy team,

I am compiling CuPy for ppc64le, using
```
python setup.py build
```
After installing CUB in CONDA_PREFIX  path, module configurations looked ok:

<details><summary> Output </summary>
<p>

```
Options: {'package_name': 'cupy', 'long_description': None, 'wheel_libs': [], 'wheel_includes': [], 'no_rpath': False, 'profile': False, 'linetrace': False, 'annotate': False, 'no_cuda': False, 'use_hip': False}

-------- Configuring Module: cuda --------
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++

-------- Configuring Module: cusolver --------
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++

-------- Configuring Module: cudnn --------
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++

-------- Configuring Module: nccl --------
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++

-------- Configuring Module: nvtx --------
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++

-------- Configuring Module: thrust --------
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
[...]

PR #3319, opened 2020-05-02
Support C++ template code in `cupy.RawModule`

### LINKED ISSUE #3185: Allow C++ Templating Functionality
Would it be possible to allow C++ Templating functionality to raw kernels?

Rationale:
Instead of having multiple functions based on input type, there would be one templated global function.

leofang:
Close #3185. 

This feature is enabled using NVRTC; I am not sure how it can work using NVCC...

The following example modified from https://github.com/cupy/cupy/issues/3185#issuecomment-622539640 works (**UPDATE**):
```python
import cupy as cp


cuda_code = '''
template<typename T>
__global__ void f(T* input) {
  unsigned int x = blockIdx.x * blockDim.x + threadIdx.x;
  input[x] += 1;
}
'''

specializations=('f<int>', 'f<float>')
mod = cp.RawModule(code=cuda_code, options=('--std=c++11',),
                   specializations=specializations)
f_int = mod.get_function(specializations[0])
f_float = mod.get_function(specializations[1])

in_int = cp.zeros(10, dtype=cp.int32)
in_float = cp.zeros(10, dtype=cp.float32)

f_int((1,), (10,), (in_int,))
f_float((1,), (10,), (in_float,))

assert (in_int == 1).all()
assert (in_float == 1.0).all()
```

@mnicely Is this something you expected?


leofang:
Document here in case I forgot: I don't wanna support `RawKernel` because it makes no sense to write a template code but use it for only one specialization. Conversely, if we have N>1 specializations, using `RawKernel` would force N compilations. `RawModule` was created exactly for avoiding this.

leofang:
Oh, I am convinced that this could be a handy feature. Note that it works even for non-template kernels which are not declared with `extern "C"`:
```python
import cupy as cp


cuda_code = '''
__global__ void f(float* in_arr) {
  unsigned int x = blockIdx.x * blockDim.x + threadIdx.x;
  in_arr[x] += 1;
}
'''

specializations=('f',)
mod = cp.RawModule(code=cuda_code, options=('--std=c++11',),
                   specializations=specializations)
name = mod.get_mangled_name('f')
print("mangled name:", name)
ker = mod.get_function(name)
in_arr = cp.zeros(10, dtype=cp.float32)
ker((1,), (10,), (in_arr,))
assert (in_arr == 1).all()
```
Output:
```bash
mangled name: _Z1fPf
```

mnicely:
@leofang This all looks great! I'm going to give a try through the week and give you updates on usability and performance.

mnicely:
Do I understand this code correctly. If I pass a tuple of multiple types specializations=('f<int>', 'f<float>', 'f<double>'), they are all compiled on first pass?

leofang:
> Do I understand this code correctly. If I pass a tuple of multiple types specializations=('f', 'f', 'f'), they are all compiled on first pass?

That's right, all specializations are compiled at once, which is exactly what NVRTC does.

leofang:
Added a few more tests. One of the new tests `test_context_switch_RawModule5` is subject to the same caveat as are its peers (#3320), but running the CI against the entire test suite should be just fine.

mnicely:
@leofang What is the best way to pull in these changes and test them? Sorry for the basic GH question.

leofang:
No problem, thanks for testing for me. 🙂 Assuming you already have a local build of cupy, you can do:
1. `wget https://github.com/cupy/cupy/pull/3319.diff` (or use `.patch`)
2. `git apply 3319.diff` (or `git am 3319.patch` if you download the patch)
3. rebuild

Let me know if it works or not.

mnicely:
I'm getting the following error building
```bash
    /usr/local/cuda/bin/nvcc -D_FORCE_INLINES=1 -I/usr/local/cuda/include -I/home/mnicely/anaconda3/envs/cupy/include/python3.8 -c cupy/cuda/cupy_cub.cu -o build/temp.linux-x86_64-3.8/cupy/cuda/cupy_cub.o --generate-code=arch=compute_30,code=compute_30 --generate-code=arch=compute_50,code=compute_50 --generate-code=arch=compute_60,code=sm_60 --generate-code=arch=compute_61,code=sm_61 --generate-code=arch=compute_70,code=sm_70 --generate-code=arch=compute_75,code=sm_75 --generate-code=arch=compute_70,code=compute_70 -O2 --compiler-options="-fPIC" --std=c++11
    cupy/cuda/cupy_cub.cu:1:10: fatal error: cupy/complex.cuh: No such file or directory
     #include <cupy/complex.cuh>
              ^~~~~~~~~~~~~~~~~~
    compilation terminated.
    error: command '/usr/local/cuda/bin/nvcc' failed with exit status 1
ERROR: Command errored out with exit status 1: /home/mnicely/anaconda3/envs/cupy/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/mnicely/workStuff/cupy/setup.py'"'"'; __file__='"'"'/home/mnicely/workStuff/cupy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps Check the logs for full command output.
```

Had to delete my cub directory to get successful build

leofang:
@mnicely, sorry that you hit a family of existing CUB bugs (https://github.com/cupy/cupy/issues/3075#issuecomment-5872746900). You need to set an env var `CUB_PATH` pointing to your CUB dir. The PR addressing this is still pending unfortunately, and before it is resolved I am under the impression that the CUB things are to be kept undocumented. 😅

asi1024:
The return value of `get_mangled_name` looks only used for the argument of the method `get_function`. How about supporting the method like as `get_function_from_specialization` as a shortcut of `mod.get_function(mod.get_mangled_name(...))` instead of `get_mangled_name`?

leofang:
> The return value of `get_mangled_name` looks only used for the argument of the method `get_function`. 

It is correct.

> How about supporting the method like as `get_function_from_specialization` as a shortcut of `mod.get_function(mod.get_mangled_name(...))` instead of `get_mangled_name`?

I feel the name `get_function_from_specialization` is too lengthy...If `get_mangled_name` is considered redundant, I would suggest simply combining it with `get_function` such that `get_function('func<double>')` works.

mnicely:
Just curious is it possible to pass in a particular compute capability to speed up build time?

leofang:
Hi @mnicely You can set `CUPY_NVCC_GENERATE_CODE`, see https://docs-cupy.chainer.org/en/stable/reference/environment.html.

mnicely:
It looks like `CUPY_NVCC_GENERATE_CODE=compute_75,compute_75` failed.
```bash
ccache nvcc -D_FORCE_INLINES=1 -I/usr/local/cuda/include -I/home/belt/workStuff/rapids/cupy/install/../cupy/core/include -I/usr/local/cuda/include/cub -I/home/belt/anaconda3/include/python3.7m -c cupy/cuda/cupy_cub.cu -o build/temp.linux-x86_64-3.7/cupy/cuda/cupy_cub.o --generate-code=compute_75,sm_75 -O2 --compiler-options="-fPIC" --std=c++11
    nvcc fatal   : 'compute_75' is not in 'keyword=value' format
    error: command 'ccache' failed with exit status 1
```

But `CUPY_NVCC_GENERATE_CODE=arch=compute_75,code=compute_75`
worked.

leofang:
Yes, I always set `CUPY_NVCC_GENERATE_CODE=arch=compute_75,code=sm_75`. I think the option is passed to `nvcc -arch ...`.

mnicely:
Should the documentation be updated to include ‘arch=‘ and ‘code=‘?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

mnicely:
Hi @leofang, I noticed in ‘specializations’ you specified the kernel name (‘f’) and multiple data types. What about the case where there are multiple kernels in the Raw Module. Do we just add those combinations of kernel and type in the same tuple? 

Finally have my environment setup to try tomorrow. In my simple use case I have two kernel in one module and two data types.

leofang:
> What about the case where there are multiple kernels in the Raw Module. Do we just add those combinations of kernel and type in the same tuple?

Yes, this is expected in NVRTC, so nothing special is done on the Python side (other than requiring the kernel "list" to be a Python `tuple` instead of `list`, because the cache requires keys to be hashable). This example works:
```python
import cupy as cp


cuda_code = '''
template<typename T>
__global__ void f(T* input) {
  unsigned int x = blockIdx.x * blockDim.x + threadIdx.x;
  input[x] += 1;
}

template<typename T>
__global__ void g(T* input) {
  unsigned int x = blockIdx.x * blockDim.x + threadIdx.x;
  input[x] -= 1;
}
'''

specializations=('f<int>', 'f<float>', 'g<double>')
mod = cp.RawModule(code=cuda_code, options=('--std=c++11',),
                   specializations=specializations)
names = [mod.get_mangled_name(ker) for ker in specializations]
f_int = mod.get_function(names[0])
f_float = mod.get_function(names[1])
g_double = mod.get_function(names[2])

in_int = cp.zeros(10, dtype=cp.int32)
in_float = cp.zeros(10, dtype=cp.float32)
in_double = cp.ones(10, dtype=cp.float64)

f_int((1,), (10,), (in_int,))
f_float((1,), (10,), (in_float,))
g_double((1,), (10,), (in_double,))

assert (in_int == 1).all()
assert (in_float == 1.0).all()
assert (in_double == 0.0).all()
```

mnicely:
Chipping away at testing. Trying to test on a V100 and I'm getting the following warning.
```bash
<frozen importlib._bootstrap>:219: RuntimeWarning: cupy.cuda.function.Module size changed, may indicate binary incompatibility. Expected 32 from C header, got 40 from PyObject
```
Should this be a concern?

leofang:
No, it's harmless, just an annoying Cython issue. It's because the C struct size for `Module` defined in `cupy/cuda/function.pxd` increases, but some Cython modules do not see it. If you remove all Cython-generated .c/.cpp/.so files and rebuild, it'd go away. 

pointers:
- https://github.com/numpy/numpy/issues/11788#issuecomment-423078169
- https://stackoverflow.com/a/40846742/2344149
- https://morepypy.blogspot.com/2018/04/how-to-ignore-annoying-cython-warnings.html

leofang:
By the way,

>> Yes, I always set `CUPY_NVCC_GENERATE_CODE=arch=compute_75,code=sm_75`. I think the option is passed to `nvcc --generate-code=...`. Looks like the doc should be fixed.
>
> Should the documentation be updated to include ‘arch=‘ and ‘code=‘?

Sorry, I totally forgot that I already fixed it in #3147 😅, but you need to switch to the "latest" tag to see it:
https://docs-cupy.chainer.org/en/latest/reference/environment.html

mnicely:
> By the way,
> 
> > > Yes, I always set `CUPY_NVCC_GENERATE_CODE=arch=compute_75,code=sm_75`. I think the option is passed to `nvcc --generate-code=...`. Looks like the doc should be fixed.
> > 
> > 
> > Should the documentation be updated to include ‘arch=‘ and ‘code=‘?
> 
> Sorry, I totally forgot that I already fixed it in #3147 , but you need to switch to the "latest" tag to see it:
> https://docs-cupy.chainer.org/en/latest/reference/environment.html

Thanks @leofang !

mnicely:
@leofang This works like a charm!

mnicely:
I'm trying to plan integration into cuSignal. Do you know when 8.0 will be release publicly?

leofang:
> Do you know when 8.0 will be release publicly?

This is a question for PFN folks: @kmaehashi @emcastillo

leofang:
Ping @asi1024 again to address https://github.com/cupy/cupy/pull/3319#issuecomment-625257789 🙂

asi1024:
@leofang I agree to support it in `get_function`.

kmaehashi:
> This is a question for PFN folks: @kmaehashi @emcastillo @asi1024

Added our current plan here: https://github.com/chainer/chainer/issues/8560

leofang:
Thanks @asi1024! I added the support to `get_function` in 25ba186. PTAL.

Thanks for sharing the timeline, @kmaehashi! It's very useful to know in advance so that we can plan accordingly 🙂

cupy/core/raw.pyx
```diff
@@ -205,16 +209,21 @@ cdef class RawKernel:
...
     assert (code is None) != (path is None)
     mod = _get_raw_module(code, path, options, backend,
-                          translate_cucomplex, enable_cooperative_groups)
+                          translate_cucomplex, enable_cooperative_groups,
+                          specializations)
     ker = mod.get_function(name)
     return ker
 
 
+cdef object _cpp_template_ker_name = re.compile(r'\s*'.join(['<', '.+', '>']))
```
leofang, line 224: Would be nice to have someone proficient in regex to take a look (I know very little 😅). The goal is to tell if a string contains a pair of `<` and `>`, such as `func<int>` and `func2<complex<double>>`.

asi1024: This implementation cannot support [this use case](https://github.com/cupy/cupy/pull/3319#issuecomment-623054903).

leofang: That's OK because we then would not look up the (mangled) name from the specialization dict. `test_raw.py` passed locally, which does contain that case.

leofang: Sorry I get it now. So I should simply just try retrieving the mangled name?

leofang: See 1b5de18.

leofang:
@asi1024 I thought about our discussion on merging `get_mangled_name` with `get_function`, and I think it'd lead to name collision for use cases like https://github.com/cupy/cupy/pull/3319#issuecomment-623054903:
```CUDA
template <typename T>
__global__ my_func(T* input, ...) { ... }

// note no extern "C" is used
__global__ my_func(float* input, ...) { ... }
```
So perhaps we should keep `get_mangled_name`?

leofang:
> So perhaps we should keep `get_mangled_name`?

I guess we're good? NVRTC does not allow this pathological use case anyway:
```
cupy.cuda.compiler.CompileException: __nv_name_map(2): error: cannot determine which instance of overloaded function "my_func" is intended
```

leofang:
@asi1024, changes up to 1389186 are ready for review, thanks.

cupy/core/core.pyx
```diff
@@ -1852,7 +1852,8 @@ cpdef function.Module compile_with_cache(
 
     return cuda.compile_with_cache(
         source, options, arch, cachd_dir, extra_source, backend,
-        enable_cooperative_groups=enable_cooperative_groups)
+        enable_cooperative_groups=enable_cooperative_groups,
+        specializations=specializations)
```
asi1024, line 1856: `specializations` feels not good for [this use case](https://github.com/cupy/cupy/pull/3319#issuecomment-623054903). How about using `linkage` for example?

cupy/core/raw.pyx
```diff
@@ -250,17 +252,40 @@ cdef class RawModule:
...
     def __init__(self, *, str code=None, str path=None, tuple options=(),
                  str backend='nvrtc', bint translate_cucomplex=False,
-                 bint enable_cooperative_groups=False):
+                 bint enable_cooperative_groups=False,
+                 specializations=None):
         if (code is None) == (path is None):
             raise TypeError(
                 'Exactly one of `code` and `path` keyword arguments must be '
                 'given.')
+        if specializations:
```
asi1024, line 273: For empty dict input.
```suggestion
        if specializations is not None:
```

cupy/core/raw.pxd
```diff
@@ -7,6 +7,7 @@ cdef class RawKernel:
         readonly tuple options
         readonly str backend
         readonly bint enable_cooperative_groups
+        tuple specializations
```
asi1024, line 10: Could you rename to `_specializations` if this attribute is only for internal implementation?

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -665,6 +740,64 @@ def test_context_switch_RawModule4(self):
...
+        if self.backend == 'nvcc':
+            self.skipTest('nvcc does not support template specialization')
+
+        # compile code
+        specializations = ['my_sqrt<unsigned int>']
+        mod = cupy.RawModule(code=test_cxx_template, options=('--std=c++11',),
+                             specializations=specializations)
+
+        # switch device
+        cupy.cuda.runtime.setDevice(1)
```
asi1024, line 756: Could you use `with cupy.cuda.Device(1)` to switch the current device? It automatically restores the current device at the end of the context.

cupy/core/core.pyx
leofang: Thanks, @asi1024. I have two questions:
1. I think it's more than the name `linkage` would imply. How about `name_expressions` following the NVRTC documentation?
2. Would you like this to be changed from all the way up (that is, the argument of RawModule)?

asi1024: 1. `name_expressions` looks good :+1:
2. Yes, all `specializations` should be changed.

cupy/core/raw.pxd
leofang: I think it's already because it's a `cdef` attribute. It's not visible in the Python space, so let me save one key stroke 😛

leofang:
@asi1024 PTAL.

cupy/core/core.pyx
mnicely: LGTM

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

leofang:
Thanks, @asi1024! Also thanks @mnicely for letting me know that the NVRTC API supports this!

PR #3321, opened 2020-05-04
Ensure runtime context on a per-device basis

### LINKED ISSUE #3320: Runtime context is not ensured on a per-device basis
Reproducer: running the whole test suite passed
```bash
pytest tests/cupy_tests/core_tests/test_raw.py
```
But running a particular test did not:
```bash
pytest tests/cupy_tests/core_tests/test_raw.py -k test_context_switch_RawModule1  # or test_context_switch_RawModule3
```
The reason is that the function `runtime._ensure_context()` is not done on a per-device basis, and in these tests in which some driver calls followed immediately after switching the device, the function mistakenly thought that the context was fully established on the new device, but it was not (only initialized, but not fully established). 

The pytest error is a bit obscure if running in the above way. Running the test code outside the test framework shows `cupy.cuda.driver.CUDADriverError: CUDA_ERROR_CONTEXT_IS_DESTROYED: context is destroyed`. 

<details><summary>This patch fixes it:</summary>

```diff
diff --git a/cupy/cuda/runtime.pyx b/cupy/cuda/runtime.pyx
index 7d0dd5a8f..351543670 100644
--- a/cupy/cuda/runtime.pyx
+++ b/cupy/cuda/runtime.pyx
@@ -35,10 +35,11 @@ cdef object _thread_local = threading.local()
 
 cdef class _ThreadLocal:
 
-    cdef bint context_initialized
+    cdef list context_initialized
 
     def __init__(self):
-        self.context_initialized = False
+        cdef int i
+        self.context_initialized = [False for i in range(getDeviceCount())]
 
     @staticmethod
     cdef _ThreadLocal get():
@@ -642,10 +643,11 @@ cdef _ensure_context():
     See discussion on https://github.com/cupy/cupy/issues/72 for details.
     """
     tls = _ThreadLocal.get()
-    if not tls.context_initialized:
+    cdef int dev = getDevice()
+    if not tls.context_initialized[dev]:
         # Call Runtime API to establish context on this host thread.
         memGetInfo()
-        tls.context_initialized = True
+        tls.context_initialized[dev] = True
 
 
 ##############################################################################
```

</details>

~~Will make a PR shortly.~~  **UPDATE:** See #3321.

leofang:
Close #3320.

If we need a test for this, we'll need to run it as a subprocess, because somewhere else in the test suite we'll establish contexts for all devices, and this bug won't show up (which is why we didn't see it until now). Let me know if a test is desired.

I think the overhead of extra runtime API calls like `getDevice()` and `getDeviceCount()` are very low, so performance impact, if any, can only come from the usage of a Python list. I could try to do some quick benchmarks if desired, but I think this bug definitely needs to be fixed one way or another.

### DIFF
--- a/cupy/cuda/runtime.pyx
+++ b/cupy/cuda/runtime.pyx
@@ -35,10 +35,11 @@ cdef object _thread_local = threading.local()
 
 cdef class _ThreadLocal:
 
-    cdef bint context_initialized
+    cdef list context_initialized
 
     def __init__(self):
-        self.context_initialized = False
+        cdef int i
+        self.context_initialized = [False for i in range(getDeviceCount())]
 
     @staticmethod
     cdef _ThreadLocal get():
@@ -642,10 +643,11 @@ cdef _ensure_context():
     See discussion on https://github.com/cupy/cupy/issues/72 for details.
     """
     tls = _ThreadLocal.get()
-    if not tls.context_initialized:
+    cdef int dev = getDevice()
+    if not tls.context_initialized[dev]:
         # Call Runtime API to establish context on this host thread.
         memGetInfo()
-        tls.context_initialized = True
+        tls.context_initialized[dev] = True
 
 
 ##############################################################################


leofang:
> I think the overhead of extra runtime API calls like `getDevice()` and `getDeviceCount()` are very low, so performance impact, if any, can only come from the usage of a Python list. I could try to do some quick benchmarks if desired, but I think this bug definitely needs to be fixed one way or another.

Added a snippet 
```python
cpdef test_ensure_context():
    _ensure_context()
```
to `cupy/cuda/runtime.pyx` and tested in ipython with `%timeit` magic, I got:

Before:
```bash
135 ns ± 0.386 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
This PR (with Python list):
```bash                                                                                                               
193 ns ± 0.152 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
Replace Python list by std::vector of bools:
```bash
179 ns ± 0.255 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
If desired I can switch to the vector version.

(Note the timings are obtained after warm-up, since the first shot is much longer.)

emcastillo:
Jenkins, test this please

leofang:
Thanks, Emilio!

PR #3329, opened 2020-05-11
Add `cupy.piecewise`

Dahlia-Chehata:


Dahlia-Chehata:
| n | numpy | cupy |
| :---: | :---: | :---: |
| 100 | 195.004 us | 413.825 us |
| 1000 | 1924.999 us | 1338.802 us |
| 10000 | 18665.776 us | 10674.284 us |
| 100000 | 186747.146 us | 103642.666 us |
| 1000000 |1930592.689 us | 1049231.745 us |

Dahlia-Chehata:
Is there a way to support callable functions with kernels ? 
to parallelize this loop :
` if any(isinstance(item, Callable) for item in funclist):`
           `raise ValueError('Callable functions are not supported') `

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,80 @@
...
+
+    .. warning::
+
+        This function currently doesn't support callable functions
+        args and kw parameters are not supported
+
+    .. seealso:: :func:`numpy.piecewise`
+    """
+
+    if any(isinstance(item, Callable) for item in funclist):
```
asi1024, line 48: How about using `callable`?
```suggestion
    if any(callable(item) for item in funclist):
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,80 @@
...
+        raise ValueError('Callable functions are not supported')
```
asi1024, line 49: Please raise `NotImplementedError`.
```suggestion
        raise NotImplementedError('Callable functions are not supported currently')
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,76 @@
...
+
+_piecewise_krnl = ReductionKernel(
+    'S x1, T x2',
+    'U y',
+    'x1 ? x2 : NULL',
+    'b == NULL? a : b',
+    'y = a',
+    'NULL',
+    'piecewise'
+)
```
asi1024, line 15: `ReductionKernel` performs well for a large size reduction axis.
However, int the case of this routine, the parameter `condlist` is given as a list of ndarrays or scalars -- its size is equal to the reduction size -- and [concatenating them](https://github.com/cupy/cupy/pull/3329/files#diff-26cc69efdac81a9dee1d8f0ebc5d0356R54) cannot be parallelized. So this function cannot work with good performance for a large `condlist` size even if we implement it with `ReductionKernel`.
How about using `ElementwiseKernel` in a loop simply? If we do so, we don't have to concatenate `condlist`, and we don't need to allocate additional memory space.

Dahlia-Chehata:
| n | numpy | cupy |
| :---: | :---: | :---: |
| 100 | 207.422 | 385.469 |
| 1000 | 2002.841 | 1064.974 |
| 10000 | 19664.703 | 7700.720 |
| 100000 | 195738.325 | 73730.930 |
| 1000000 |1968100.123 | 732554.892 |

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,76 @@
...
+    scalar = 0
+    if cupy.isscalar(condlist):
+        scalar = 1
+    if cupy.isscalar(condlist) or (
+            (not isinstance(condlist[0], (list, ndarray))) and x.ndim != 0):
+        condlist = [condlist]
+    condlist = cupy.array(condlist, dtype=bool)
+    condlen = len(condlist)
+    funclen = len(funclist)
+    if condlen + 1 == funclen:  # o.w
```
asi1024, line 57: How about using `funclen[-1]` as the default value of `y` simply?

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,76 @@
...
+    funclist = cupy.asarray(funclist)
+    if scalar:
+        funclist = funclist[condlist]
+        condlist = cupy.ones(shape=(1, x.size), dtype=bool)
+
+    y = cupy.zeros(x.shape, x.dtype)
+    if not x.ndim:
+        _piecewise_krnl(condlist, funclist, y)
+    else:
+        for i in range(x.size):
```
asi1024, line 74: This loop can be parallelized.

tests/cupy_tests/functional_tests/test_piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+
+import pytest
+
+import cupy
+from cupy import testing
+
+
+class TestPiecewise(unittest.TestCase):
+
+    @testing.for_all_dtypes(no_bool=True, no_complex=True)
```
asi1024, line 11: Please test also in bool and complex inputs.
```suggestion
    @testing.for_all_dtypes()
```

tests/cupy_tests/functional_tests/test_piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+    @testing.numpy_cupy_array_equal()
+    def test_linespace(self, xp, dtype):
+        x = xp.linspace(-2.5, 2.5, 6, dtype=dtype)
+        condlist = [x < 0, x >= 0]
```
asi1024, line 15: All of `condlist` in your tests are distinct covers (Exact one condition is true.) Please test other cases.

asi1024:
@Dahlia-Chehata Could you show us the benchmark script?

cupy/functional/piecewise.py
Dahlia-Chehata: If I use ElementwiseKernel and eliminated this loop. How the shapes [(funclist.size , x.size),  (funclist,), (x.size,)] corresponding to condlist, funclist and y are going to be broadcast ? 
Either funclist or y will need to be broadcast first before using the kernel (i.e tile for example). And y won't be 1D in this case right ?

Dahlia-Chehata:
> @Dahlia-Chehata Could you show us the benchmark script?

``` 
def compare(n):
    def generator(n, p=50):
        x = np.linspace(-50, 50, p)
        funclist = [random.randint(1000) for x in range(1, n)]
        condlist=[]
        for i in range(len(funclist)):
            condlist.append([bool(getrandbits(1))for j in range (len(x))])
        return x, condlist, funclist
    x, condlist, funclist = generator(n)
    np_ret = repeat(np.piecewise, (x, condlist, funclist), n=10, n_warmup=10)
    cp_ret = repeat(cp.piecewise, (x, condlist, funclist), n=10, n_warmup=10)
    print("NumPy: ", np_ret.to_str(show_gpu=True))
    print("CuPy:", cp_ret.to_str(show_gpu=True)) 
```

cupy/functional/piecewise.py
asi1024: ElementwiseKernel supports broadcast just like as ufuncs.

> How the shapes [(funclist.size , x.size), (funclist,), (x.size,)] corresponding to condlist, funclist and y are going to be broadcast ?

In this case, you can operate by `elementwise_func(x, funclist[...,None], funclist)`.

Dahlia-Chehata:
> @Dahlia-Chehata Could you show us the benchmark script?

``` 
def compare(n):
    def generator(n, p=50):
        x = np.linspace(-50, 50, p)
        funclist = [random.randint(1000) for x in range(1, n)]
        condlist=[]
        for i in range(len(funclist)):
            condlist.append([bool(getrandbits(1))for j in range (len(x))])
        return x, funclist, condlist
    x1, f1, c1 = generator(n,np)
    x2, f2, c2 = generator(n,cp)
    np_ret = repeat(np.piecewise, (x1, c1, f1), n=10, n_warmup=10)
    cp_ret = repeat(cp.piecewise, (x2, c2, f2), n=10, n_warmup=10)
    print("NumPy: ", np_ret.to_str(show_gpu=True))
    print("CuPy:", cp_ret.to_str(show_gpu=True)) 
```

tests/cupy_tests/functional_tests/test_piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+        condlist = True
+        funclist = [-10, 10]
+        return xp.piecewise(x, condlist, funclist)
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_otherwise_condition1(self, xp, dtype):
+        x = xp.linspace(-2, 4, 12, dtype=dtype)
+        condlist = [x < 0, x >= 0, x != 3, x == 10]
+        funclist = [-1, 0, 2, 3, -5]
```
asi1024, line 40: In this test, `-5` is not chosen as the return value.

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+
+_piecewise_krnl = core.ElementwiseKernel(
+    'U condlist, raw S funclist, int64 xsiz',
+    'raw I prev, raw T y',
+    '''
+        if (condlist && prev[i % xsiz] < i)
+            prev[i % xsiz] = i;
+        __syncthreads();
+        if (prev[i % xsiz] == i)
+            y[i % xsiz] = funclist[i / xsiz];
```
asi1024, line 13: The results are undefined when multiple conditions are satisfied. Please avoid such test cases.

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+            funclist: list of scalar functions.
+
+        Returns:
+            cupy.ndarray: the result of calling the functions in funclist
+                on portions of x defined by condlist.
+
+        .. warning::
+
+            This function currently doesn't support callable functions
+            args and kw parameters are not supported
```
asi1024, line 38:
```suggestion
            This function currently doesn't support callable functions,
            args and kw parameters.
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+        ''',
+    'piecewise_kernel'
+)
+
+
+def piecewise(x, condlist, funclist):
+    """Evaluate a piecewise-defined function.
+
+        Args:
+            x: input domain
```
asi1024, line 23: Write the type of arguments.
```suggestion
            x (cupy.ndarray): input domain
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+    funclen = len(funclist)
+    if condlen + 1 == funclen:  # o.w
+        y = cupy.full(shape=x.size, fill_value=funclist[-1], dtype=x.dtype)
+        funclist = funclist[:-1]
+        funclen -= 1
+    elif condlen != funclen:
+        raise ValueError('with {} condition(s), either {} or {} functions'
+                         ' are expected'.format(condlen, condlen, condlen + 1))
+    else:
+        y = cupy.zeros(shape=x.shape, dtype=x.dtype)
```
asi1024, line 66: How about writing as the following order for the readability?
```
if funclen == codnlen:
    ...
elif funclen == condlen + 1:
    ...
else:
    raise ValueError(....)
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+            args and kw parameters are not supported
+
+        .. seealso:: :func:`numpy.piecewise`
+        """
+
+    if any(callable(item) for item in funclist):
+        raise NotImplementedError(
+            'Callable functions are not supported currently')
+    if cupy.isscalar(x):
+        x = cupy.asarray(x)
```
asi1024, line 47: We don't need to support scalar inputs because it will not work with good performance.

tests/cupy_tests/functional_tests/test_piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+
+import cupy
+from cupy import testing
+
+
+class TestPiecewise(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_linespace(self, xp, dtype):
```
asi1024, line 13:
```suggestion
    def test_piecewise(self, xp, dtype):
```

tests/cupy_tests/functional_tests/test_piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+        x = xp.linspace(2.5, 12.5, 6, dtype=dtype)
+        condlist = [x < 0, x >= 0, x < 5, x >= 1.5]
+        funclist = [-1, 1, 2, 5]
+        return xp.piecewise(x, condlist, funclist)
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_scalar_value(self, xp, dtype):
```
asi1024, line 21:
```suggestion
    def test_scalar_input(self, xp, dtype):
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+        funclen -= 1
+    elif condlen != funclen:
+        raise ValueError('with {} condition(s), either {} or {} functions'
+                         ' are expected'.format(condlen, condlen, condlen + 1))
+    else:
+        y = cupy.zeros(shape=x.shape, dtype=x.dtype)
+    funclist = cupy.asarray(funclist)
+    if scalar:
+        funclist = funclist[condlist]
+        condlist = cupy.ones(shape=x.size, dtype=bool)
```
asi1024, line 70: Could you explain why these lines are needed?

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,73 @@
...
+            (not isinstance(condlist[0], (list, cupy.ndarray)))
+            and x.ndim != 0):
+        condlist = [condlist]
+    condlist = cupy.array(condlist, dtype=bool)
+    condlen = len(condlist)
+    funclen = len(funclist)
+    if condlen + 1 == funclen:  # o.w
+        y = cupy.full(shape=x.size, fill_value=funclist[-1], dtype=x.dtype)
+        funclist = funclist[:-1]
+        funclen -= 1
```
asi1024, line 61: `funclen` is unused in the following lines.

tests/cupy_tests/functional_tests/test_piecewise.py
```diff
@@ -0,0 +1,73 @@
+import unittest
+
+import pytest
+
+import cupy
+from cupy import testing
+
+
+class TestPiecewise(unittest.TestCase):
```
asi1024, line 9: Please add tests also for non-1dim inputs.

cupy/functional/piecewise.py
Dahlia-Chehata: Could you explain more, please?  You mean the kernel or the test cases. I ensured that when multiple conditions are met, only the latest is considered. So the values are overwritten.

asi1024: Multiple threads will read/write the same memory space `prev[i % xsize]` before syncthreads.

Dahlia-Chehata: I get that multiple threads can't write the same prev[i % xsize], but about reading the same memory memory space ? I mean will replacing syncthreads inside the condition `if (condlist && prev[i % xsiz] < i)` before the writing part resolve this issue ?

Dahlia-Chehata: What about this ?
``` 
        if (condlist){
          __syncthreads();
          if (prev[i % xsiz] < i)
            prev[i % xsiz] = i;
          __syncthreads();
          if (prev[i % xsiz] == i) 
            y[i % xsiz] = funclist[i / xsiz];
        }
```

cupy/functional/piecewise.py
Dahlia-Chehata: Should I raise error or not handle it at all (and eliminate its corresponding test) ?

cupy/functional/piecewise.py
asi1024: It looks to cause thread divergence.

Dahlia-Chehata: Ok, regarding the original kernel, will atomic operations (reads/ writes or workarounds using cuda) resolve the race condition issue? If not, could you provide suggestions to search/read about ?

asi1024: The atomic operations also cannot resolve the race condition because `__syncthreads` cannot synchronize two threads belonging to different blocks.
I think it's OK to give up parallelizing in the condlist-axis as I commented [here](https://github.com/cupy/cupy/pull/3329#discussion_r423589567).

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,68 @@
...
+    'piecewise_kernel'
+)
+
+
+def piecewise(x, condlist, funclist):
+    """Evaluate a piecewise-defined function.
+
+        Args:
+            x (cupy.ndarray): input domain
+            condlist (cupy.ndarray or list):
```
asi1024, line 18:
```suggestion
            condlist (list of cupy.ndarray or bool scalars):
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,68 @@
...
+                conditions list ( boolean arrays or boolean scalars).
```
asi1024, line 19: This line doesn't look needed.
```suggestion
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,68 @@
...
+                Each boolean array/ scalar corresponds to a function
+                in funclist. Length of funclist is equal to that
+                of condlist. If one extra function is given, it is used
+                as otherwise condition
+            funclist (cupy.ndarray or list): list of scalar functions.
```
asi1024, line 24: Functions are not supported currently.

cupy/functional/piecewise.py
Dahlia-Chehata: Sorry I overlooked this one :bow:

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,68 @@
...
+
+        .. seealso:: :func:`numpy.piecewise`
+        """
+
+    if any(callable(item) for item in funclist):
+        raise NotImplementedError(
+            'Callable functions are not supported currently')
+    if cupy.isscalar(condlist):
+        condlist = cupy.full(shape=x.shape, fill_value=condlist)
+    if not isinstance(condlist[0], (list, cupy.ndarray)) and x.ndim != 0:
```
asi1024: Could you explain why the codition `x.ndim != 0` is needed?
Please add the test for 0-dim input.

cupy/functional/piecewise.py
Dahlia-Chehata: Scalar ones are.. right ?

cupy/functional/piecewise.py
Dahlia-Chehata: needed for scalar input

asi1024: Sorry, I noticed now this line is already outdated :bow:

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,59 @@
...
+            args and kw parameters.
+
+        .. seealso:: :func:`numpy.piecewise`
+        """
+
+    if any(callable(item) for item in funclist):
+        raise NotImplementedError(
+            'Callable functions are not supported currently')
+    if cupy.isscalar(condlist) or (not isinstance(
+            condlist[0], (list, cupy.ndarray)) and x.ndim != 0):
```
asi1024, line 42: I'm wondering if this condition is correct. Please add tests for non-list cupy.ndarray.

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,59 @@
...
+            condlist (cupy.ndarray or list):
+                conditions list ( boolean arrays or boolean scalars).
+                Each boolean array/ scalar corresponds to a function
+                in funclist. Length of funclist is equal to that
+                of condlist. If one extra function is given, it is used
+                as otherwise condition
+            funclist (cupy.ndarray or list): list of scalar functions.
+
+        Returns:
+            cupy.ndarray: the result of calling the functions in funclist
```
asi1024, line 27: `The scalar values in funclist...` is preferred.

cupy/functional/piecewise.py
asi1024: Yes. `list of scalars` are right.

cupy/functional/piecewise.py
asi1024: So, please remove this line.

asi1024: Ah, now it is used for the argument of `range`. It's OK to leave.

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,58 @@
...
+def piecewise(x, condlist, funclist):
+    """Evaluate a piecewise-defined function.
+
+        Args:
+            x (cupy.ndarray): input domain
+            condlist (list of cupy.ndarray or bool scalars):
+                Each boolean array/ scalar corresponds to a function
+                in funclist. Length of funclist is equal to that
+                of condlist. If one extra function is given, it is used
+                as otherwise condition
```
asi1024, line 22: The sentence `Length of funclist is equal to that of condlist. If one extra function is given, it is used as otherwise condition` is for funclist?

Dahlia-Chehata: If one extra function is given, it is used as the default value when the otherwise condition is met

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,58 @@
...
+    if condlen == funclen:
+        y = cupy.zeros(shape=x.shape, dtype=x.dtype)
+    elif condlen + 1 == funclen:  # o.w
+        y = cupy.full(shape=x.shape, fill_value=funclist[-1], dtype=x.dtype)
+        funclist = funclist[:-1]
+        funclen -= 1
+    else:
+        raise ValueError('with {} condition(s), either {} or {} functions'
+                         ' are expected'.format(condlen, condlen, condlen + 1))
+    funclist = cupy.asarray(funclist)
```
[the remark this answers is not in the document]
Dahlia-Chehata: I will need this line if I am to use where in order to change funclist to x.dtype ?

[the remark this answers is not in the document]
asi1024: Please cast the type of `funclist` without using additional GPU memory space.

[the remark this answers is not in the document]
asi1024: Could you add tests for different data types of `x` and `funclist`?
The current implementation seems to raise different types of errors if `x` of integer type and `funclist` of complex type are given.

cupy/functional/piecewise.py
asi1024: Could you add a test for `not isinstance(condlist[0], (list, cupy.ndarray)) and x.ndim != 0)` condition?

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,48 @@
...
+
+            This function currently doesn't support callable functions,
+            args and kw parameters.
+
+        .. seealso:: :func:`numpy.piecewise`
+        """
+
+    if any(callable(item) for item in funclist):
+        raise NotImplementedError(
+            'Callable functions are not supported currently')
```
asi1024, line 30: How about moving this check into the for-loop [here](https://github.com/cupy/cupy/pull/3329/files#diff-26cc69efdac81a9dee1d8f0ebc5d0356R46-R47)?

cupy/functional/piecewise.py
asi1024: You use two kinds of conditions here: `cupy.isscalar(...)` and `isinstance(..., (list, cupy.ndarray)`. What is the difference between them?

cupy/functional/piecewise.py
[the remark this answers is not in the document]
Dahlia-Chehata: > The current implementation seems to raise errors of different types if `x` of integer type and `funclist` of complex type are given.

Yes, this test fails but so is numpy's version

cupy/functional/piecewise.py
Dahlia-Chehata: > Could you add a test for `not isinstance(condlist[0], (list, cupy.ndarray)) and x.ndim != 0)` condition?

``` test_scalar_condition ``` already tests that

File: cupy/functional/piecewise.py (hunk 26 above)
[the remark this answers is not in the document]
Comment (asi1024) [reply cupy/functional/piecewise.py]: So please fix to raise the error of the same type as `numpy.piecewise` raises.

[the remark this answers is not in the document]
Comment (Dahlia-Chehata) [reply cupy/functional/piecewise.py]: > Please cast the type of `funclist` without using additional GPU memory space.

I am allowing funclist to be either a list or cupy.array. So Should I only allow arrays and change the tests as well ? And for casting, use astype ?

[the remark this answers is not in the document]
Comment (Dahlia-Chehata) [reply cupy/functional/piecewise.py]: Testing for callable functions  or mixed function types is not possible when funclist is array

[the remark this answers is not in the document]
Comment (Dahlia-Chehata) [reply cupy/functional/piecewise.py]: Moving the loop ( that raises error if callable fns) is also dependent on when the cast is done or else testing for callable functions or mixed function types will fail. Also casting each function when passed as argument to kernel ( with no additional GPU memory space) fails because of mismatched data types when converting from input dtype to python type

File: cupy/functional/piecewise.py (hunk 23 above)
Comment (Dahlia-Chehata) [reply cupy/functional/piecewise.py]: ``` cupy.isscalar(condlist) ``` is used when for scalar condition as in ```test_scalar_condition``` 
```not isinstance(condlist[0], (list, cupy.ndarray)) and x.ndim != 0) ``` is used for multidimensional array input with condition list of size equal to shape[0] of input as in ``` test_piecewise_diff_shapes1, 2, 3, 4 ```

Comment (Dahlia-Chehata) [reply cupy/functional/piecewise.py]: I responded by sending the last fix but forgot to explain it here . Sorry for that :bow:

Comment (asi1024) [reply cupy/functional/piecewise.py]: Why did you write as `not isinstance(condlist[0], (list, cupy.ndarray))`? What is the difference from `cupy.isscalar(condlist[0])`?

File: cupy/functional/piecewise.py [hunk 28]
```diff
@@ -0,0 +1,58 @@
...
+    elif condlen + 1 == funclen:  # o.w
+        y = cupy.full(shape=x.shape, fill_value=funclist[-1], dtype=x.dtype)
+        funclist = funclist[:-1]
+        funclen -= 1
+    else:
+        raise ValueError('with {} condition(s), either {} or {} functions'
+                         ' are expected'.format(condlen, condlen, condlen + 1))
+    funclist = cupy.asarray(funclist)
+    for i in range(funclen):
+        _piecewise_krnl(condlist[i], funclist[i], y)
```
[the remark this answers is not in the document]
asi1024: `y = cupy.where(condlist[i], funclist[i], y)` requires memory allocation. Please fix to `cupy.where(condlist[i], funclist[i], y, y)` to reuse memory space.

cupy/functional/piecewise.py
Dahlia-Chehata: cupy.isscalar(condlist[0]) ≠ not isinstance(condlist[0], (list, cupy.ndarray)) because of tuple cases.
```
x = np.array([[0,1,4],[1,2,5],[3,4,3],[3,4,9]])
condlist = [(True, False, False), (False, True, True), (False, False, False), (False, True, True)]
>>> np.isscalar(condlist[0]))
False
>>> not isinstance(condlist[0], (list, np.ndarray)))
True
```
But I still need to handle it and add a test case

cupy/functional/piecewise.py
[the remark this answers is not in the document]
Dahlia-Chehata: Can cupy.where() take 4 arguments ?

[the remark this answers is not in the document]
asi1024: If so, how about using `cupy._sorting.search._where_ufunc`?

cupy/functional/piecewise.py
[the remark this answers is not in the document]
asi1024: Why do you want to cast `funclist` to `x.dtype`?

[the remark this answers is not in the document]
Dahlia-Chehata: If not, tests will fail 

Consider

>> x = [True, False, True]
>> condlist = [[True, False, False],  [True, False, True],  [True, True, False]]
>> funclist = [1, 0, 3]
>> np.piecewise(x, condlist, funclist)
[True, True, False]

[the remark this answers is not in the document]
asi1024: I see. Please cast with `.astype(x.dtype, copy=False)` to avoid unnecessary memory allocation.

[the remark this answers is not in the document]
Dahlia-Chehata: It won't work with tuple. I had to use asarray() for tuple cases before astype()

Dahlia-Chehata:
@asi1024 Can you please review ?

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,84 @@
...
+        if callable(funclist[-1]):
+            raise NotImplementedError(
+                'Callable functions are not supported currently')
+        y = cupy.full(shape=x.shape,
+                      fill_value=x.dtype.type(funclist[-1]), dtype=x.dtype)
+        funclist = funclist[:-1]
+    else:
+        raise ValueError('with {} condition(s), either {} or {} functions'
+                         ' are expected'.format(condlen, condlen, condlen + 1))
+    output = y.T if diffshape == 1 else y
```
asi1024, line 78: Where is this behavior documented in NumPy's documentation?

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,84 @@
...
+    """Evaluate a piecewise-defined function.
+
+        Args:
+            x (cupy.ndarray): input domain
+            condlist (list of cupy.ndarray or bool scalars):
+                Each boolean array/ scalar corresponds to a function
+                in funclist. Length of funclist is equal to that of
+                condlist. If one extra function is given, it is used
+                as the default value when the otherwise condition is met
+            funclist (cupy.ndarray or list): list of scalar functions.
```
asi1024, line 23: NumPy does not guarantee support for `numpy.ndarray`.
You can support for `cupy.ndarray`, but we don't want to document it for compatibility.
```suggestion
            funclist (list of scalars): list of scalar functions.
```

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,84 @@
...
+        condlist = [condlist]
+    elif isinstance(condlist[0], tuple) and x.ndim != 0:
+        diffshape = 2
+    elif cupy.isscalar(condlist[0]) and x.ndim != 0:
+        diffshape = 1
+    if not isinstance(condlist[0], cupy.ndarray):
+        condlist = cupy.asarray(condlist)
+    condlen = len(condlist)
+    if diffshape:
+        if x.ndim == condlist.ndim:
```
asi1024, line 47: Please check for `x.ndim != condlist.ndim` case.

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,84 @@
...
+            for i in range(x.ndim - 1, -1, -1):
+                if condlist.shape[i] != x.shape[i]:
```
asi1024, line 49: How about writing as `condlist.shape != x.shape`?

cupy/functional/piecewise.py
Dahlia-Chehata: It is not documented. It is used for shapes broadcasting purposes.
Consider
>>  x = xp.array(shape=(3,4))
>> condlist = [True, False, True]
>>  funclist = [1, 2]
How do you suggest to handle this case otherwise?

cupy/functional/piecewise.py
Dahlia-Chehata: Do you mean eliminate the loop and not mention the dimension in the error msg ?

cupy/functional/piecewise.py
Dahlia-Chehata: Isn't it already checked in this else ?
``` 
else:
    raise IndexError('boolean index did not match indexed array along '
    'dimension 0; dimension is {} but corresponding '
    'boolean dimension is {}'
    .format(x.shape[0], condlen))
```
I mean for example the first case will pass and the second will raise the error
```
x = testing.shaped_arange((2, 3), xp, dtype)
condlist = [False, True]
funclist = [1, 2]
xp.piecewise(x, condlist, funclist)
```
```
x = testing.shaped_arange((2, 3), xp, dtype)
condlist = [False, True, True]
funclist = [1, 2]
xp.piecewise(x, condlist, funclist)
```

cupy/functional/piecewise.py
asi1024: This feels complex to read. How about reshaping `condition` for the broadcast?

cupy/functional/piecewise.py
asi1024: It seems enough to show the shapes `condition.shape` and `x.shape`

Dahlia-Chehata:
@asi1024 Could you help with travis ? It seems like a common issue

cupy/functional/piecewise.py
asi1024: Could you explain why should we check the shape equality only when `x.ndim == condlist.ndim`?

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,83 @@
...
+
+        .. seealso:: :func:`numpy.piecewise`
+        """
+    diffshape = 0
+    if cupy.isscalar(condlist):
+        condlist = [condlist]
+    elif isinstance(condlist[0], tuple) and x.ndim != 0:
+        diffshape = 2
+    elif cupy.isscalar(condlist[0]) and x.ndim != 0:
+        diffshape = 1
```
asi1024, line 42: Please enhance the coverage of all types of given `condlist`. For example, `IndexError` will be raised when `condlist` is 0-dim ndarray.

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,83 @@
...
+        funclist = funclist[:-1]
+    else:
+        raise ValueError('with {} condition(s), either {} or {} functions'
+                         ' are expected'.format(condlen, condlen, condlen + 1))
+    for condition, func in zip(condlist, funclist):
+        if callable(func):
+            raise NotImplementedError(
+                'Callable functions are not supported currently')
+        if diffshape == 1:
+            condition = condition.reshape((-1, 1))
```
asi1024, line 81: Move this line to [L42](https://github.com/cupy/cupy/pull/3329/files#diff-26cc69efdac81a9dee1d8f0ebc5d0356R42) for the simplification.

cupy/functional/piecewise.py
```diff
@@ -0,0 +1,83 @@
...
+                             'dimension 0; dimension is {} but corresponding '
+                             'boolean dimension is {}'
+                             .format(x.shape[0], condlen))
+        condlist = [condlist]
+    if isinstance(funclist, cupy.ndarray):
+        funclist = funclist.tolist()
+    funclen = len(funclist)
+    if condlen == funclen:
+        y = cupy.zeros(shape=x.shape, dtype=x.dtype)
+    elif condlen + 1 == funclen:  # o.w
```
asi1024, line 66: Avoid unnecessary or trivial comments.

cupy/functional/piecewise.py
asi1024: One more question. Why you don't check the shape when `diffshape == 0`?

cupy/functional/piecewise.py
Dahlia-Chehata: line 42 ? first I reshape condition not condlist, second condlist can be a list at this line.

cupy/functional/piecewise.py
Dahlia-Chehata: > Could you explain why should we check the shape equality only when `x.ndim == condlist.ndim`?

I thought I explained it by the above examples. When ``` x.ndim == condlist.ndim ```, we should check for the shapes. By not satisfying (x.dim == conlist.ndim) condition, only 2 options are possible now, either ``` condlen == x.shape[0] ``` (and in this case ``` condlen ``` should be adjusted to 1 to skip the error in L75 or simply to mean that a special broadcasting will be used for this case) or raise the error because of index mismatch along dimension 0 as it can be seen from the examples I explained [here](https://github.com/cupy/cupy/pull/3329#discussion_r436331763)
Does this answer the question ?

asi1024: Why do you check the shape equality only when `x.ndim == condlist.ndim`, and don't check when `x.ndim != condlist.ndim`?

Dahlia-Chehata: > One more question. Why you don't check the shape when `diffshape == 0`?

diffshape flag (as by its name) already targets special cases that need a different broadcasting ( list of scalars with multidimensional input or list of tuples/ nested tuples with multidimensional input). Handling these special cases require a pre-step before resuming the normal behavior for all (diffshape == 1, 2 or 0). This extra step can be seen by entering ```if diffshape: ``` condition. After that, the default behaviour is applied for different inputs regardless of diffshape value ( that can be seen from line 62, and of course with the exception of transposing condlist in case of diffshape == 1 later on...)

Dahlia-Chehata: > Why do you check the shape equality only when `x.ndim == condlist.ndim`, and don't check when `x.ndim != condlist.ndim`?

because I don't need the shapes information when x.ndim !=  condlist.ndim
Either condlen == x.shape[0] and here special broadcasting is used 
``` 
x = testing.shaped_arange((2, 3), xp, dtype)
condlist = [False, True]
funclist = [1, 2]
xp.piecewise(x, condlist, funclist) 
```
or raise error because of mismatch along dimension 0
``` x = testing.shaped_arange((2, 3), xp, dtype)
condlist = [False, True, True]
funclist = [1, 2]
xp.piecewise(x, condlist, funclist)
```
I only care about the shape along dimension 0 when x.ndim !=  condlist.ndim

cupy/functional/piecewise.py
Dahlia-Chehata: Should I move it to L76 ?

asi1024: No. L42 is correct. Of course, the condition should be converted to ndarray before reshaping.

Dahlia-Chehata: if converted to ndarray, won't it affect the conditions from L37 to 42 ?

cupy/functional/piecewise.py
asi1024: I found some different behaviors between this implementation and NumPy's one.
- Only cupy raises `ValueError`.
```py
    @testing.for_all_dtypes()
    def test_broadcast(self, xp, dtype):
        x = testing.shaped_random(shape=(2, 3, 5), xp=xp, dtype=dtype)
        condlist = [[[True, False, False], [True, True, True]]]
        funclist = [-1, 2]
        return xp.piecewise(x, condlist, funclist)
```
- Only numpy raises error.
```py
    @testing.for_all_dtypes()
    def test_broadcast(self, xp, dtype):
        x = testing.shaped_random(shape=(2, 3, 5), xp=xp, dtype=dtype)
        condlist = [[[True, False, False, False, False]], [[True, True, True, True, True]]]
        funclist = [-1, 2]
        return xp.piecewise(x, condlist, funclist)
```

Dahlia-Chehata: I will look into it

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #3330, opened 2020-05-11
Support multiple architectures in `CUPY_NVCC_GENERATE_CODE`

### LINKED ISSUE #3317: feature request:  add ability to pass multiple arches using CUPY_NVCC_GENERATE_CODE
It would be very helpful to change this line
https://github.com/cupy/cupy/blob/2804af926cead8d2e80665283473a71212e824b0/cupy_setup_build.py#L765
to use list comprehension in order to pass multiple architectures on the command line.  This is useful for building the list of not included architectures from Jetson such as 53,62,and 72.

Something like this with + sign delimited works great
`return ['--generate-code={}'.format(val) for val in envcfg.split('+')]`

and allows me to build with CUPY_NVCC_GENERATE_CODE="arch=compute_62,code=sm_62+arch=compute_72,code=sm_72" and everything builds great.

leofang:
Close #3317.

### DIFF
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -762,7 +762,8 @@ def _nvcc_gencode_options(cuda_version):
 
     envcfg = os.getenv('CUPY_NVCC_GENERATE_CODE', None)
     if envcfg:
-        return ['--generate-code={}'.format(envcfg)]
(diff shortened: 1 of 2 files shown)


emcastillo:
Jenkins, test this please

PR #2330, opened 2019-07-24
[WIP] `cupy.around` behaves differently from NumPy for EVEN_NUMBER+0.5

hyabe:
`cupy.around` is incompatible with `numpy.around` such as `cupy.around(2.5) == numpy.around(2.5)` gives `False` (tested on CuPy 7.0.0b1 vs. NumPy 1.16.4).
This PR fixes the problem, and also solves #2276.

We may need additional fix and/or discussion because there still be incompatible case(s) even with this PR:
 * `cupy.around(2**63, -1)` differs from NumPy, the latter does *not* round the last digit,
 * corner cases like `cupy.array(255, cupy.uint8).round(-1)` are not exhaustively tested.

PR #3335, opened 2020-05-13
`cupy.around` behaves differently from NumPy for EVEN_NUMBER+0.5

### LINKED ISSUE #2276: `around` result is different compared to numpy
```
>>> np.__version__
'1.16.4'
>>> np.around([2.5])
array([2.])
>>> cupy.__version__
'7.0.0a1'
>>> cupy.around([2.5])
array([3.])
```
NumPy seems to round to even.

emcastillo:
Take over #2330
Closes #2276

--
`cupy.around` is incompatible with `numpy.around` such as `cupy.around(2.5) == numpy.around(2.5)` gives `False` (tested on CuPy 7.0.0b1 vs. NumPy 1.16.4).
This PR fixes the problem, and also solves #2276.

We may need additional fix and/or discussion because there still be incompatible case(s) even with this PR:
 * `cupy.around(2**63, -1)` differs from NumPy, the latter does *not* round the last digit,
 * corner cases like `cupy.array(255, cupy.uint8).round(-1)` are not exhaustively tested.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1902,11 +1902,11 @@ template<typename T> __device__ T pow10(long long n){
 
 cdef _round_float = '''
 if (in1 == 0) {
-    out0 = round(in0);
+    out0 = rint(in0);
 } else {
     double x;
     x = pow10<double>(abs(in1));  // TODO(okuta): Move before loop
-    out0 = in1 < 0 ? round(in0 / x) * x : round(in0 * x) / x;
+    out0 = in1 < 0 ? rint(in0 / x) * x : rint(in0 * x) / x;
 }'''
 
 cdef _round_complex = '''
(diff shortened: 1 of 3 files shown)


emcastillo:
Following up the original discussion, comment by @grlee77 

>> Would it make sense to split out just the switch to using rint in _round_float and _round_complex into a separate PR? I think that one definitely helps improve compatibility with NumPy and merging it should be non-controversial.

Here do you mean to templatize the rounding function so it can be selected depending on the routine to be used?
In think we can do it in this PR instead of opening a new one :)

take-cheeze:
Jenkins, test this please!

grlee77:
> Here do you mean to templatize the rounding function so it can be selected depending on the routine to be used?
In think we can do it in this PR instead of opening a new one :)

I just meant making the change to `rint`alone without also applying the fix for large integers. I am fine with having them both together though. I think most users are unlikely to encounter the edge cases where the result may not exactly match NumPy.

Thanks for reviving this PR!

PR #3342, opened 2020-05-14
Make `testing.multi_gpu()` respect device counts

leofang:
On systems with 1 GPU but  `CUPY_TEST_GPU_LIMIT` is unset, or with multiple GPUs but some of them are made invisible (by setting `CUDA_VISIBLE_DEVICES`), tests decorated by `@testing.multi_gpu()` might fail.

### DIFF
--- a/cupy/testing/attr.py
+++ b/cupy/testing/attr.py
@@ -1,6 +1,8 @@
 import os
 import unittest
 
+import cupy
+
 
 try:
     import pytest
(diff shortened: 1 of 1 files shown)


kmaehashi:
The current spec is by design. This is because it is difficult to notice that tests are silently skipped, especially in CIs (when only 1 GPU is available due to configuration errors or unexpected GPU fault)

leofang:
I see, so you're saying it is by design to fail instead of skipped?

kmaehashi:
I don't want them to fail, but the problem is that skipped tests are treated as if it succeeded (because we don't see the detailed logs of CI if it succeeds 😅). Maybe we can improve `reason` message to tell contributors to set `CUPY_TEST_GPU_LIMIT` appropriately?

leofang:
> Maybe we can improve `reason` message to tell contributors to set `CUPY_TEST_GPU_LIMIT` appropriately?

I am not sure how this is possible. If `CUPY_TEST_GPU_LIMIT` is unset, the test is not skip, so `reason` would not be printed. 

Since my concern was already clarified, how about I just close this PR?

PR #2277, opened 2019-06-26
Make SVD overwrite temporary array x

danielhanchen:
@pentschev 
Regarding our convo.
SVD can now auto overwrite the temporary array x. Can improve performance possibly (haven't tested yet).
Should support slightly larger data sizes also.

PR #3347, opened 2020-05-17
Reduce memory usage of `cupy.linalg.svd`

### LINKED ISSUE #3327: Reduce memory usage of `cupy.linalg.svd`
As @danielhanchen tried in #2277, `cupy.svd` may be improved to reduce its memory usage. See #2277 for its detail.

takagi:
Fix #3327 .

This PR reimplements #2277 to reduce the memory usage of `cupy.linalg.svd` with following CuPy style guideline and having less changes.

Note that the implementation of `cupy.linalg.svd` expects its input `M x N` is C-order and actually computes svd of its Fortran-order transpose `N x M` by means of cuSolver.

### DIFF
--- a/cupy/linalg/decomposition.py
+++ b/cupy/linalg/decomposition.py
@@ -375,21 +375,22 @@ def svd(a, full_matrices=True, compute_uv=True):
     if compute_uv:
         if full_matrices:
             u = cupy.empty((m, m), dtype=a_dtype)
-            vt = cupy.empty((n, n), dtype=a_dtype)
+            vt = x
+            job_u = ord('A')
+            job_vt = ord('O')
         else:
-            u = cupy.empty((mn, m), dtype=a_dtype)
+            u = x
             vt = cupy.empty((mn, n), dtype=a_dtype)
+            job_u = ord('O')
+            job_vt = ord('S')
         u_ptr, vt_ptr = u.data.ptr, vt.data.ptr
     else:
         u_ptr, vt_ptr = 0, 0  # Use nullptr
+        job_u = ord('N')
+        job_vt = ord('N')
     s = cupy.empty(mn, dtype=s_dtype)
     handle = device.get_cusolver_handle()
     dev_info = cupy.empty(1, dtype=numpy.int32)
-
-    if compute_uv:
-        job = ord('A') if full_matrices else ord('S')
-    else:
-        job = ord('N')
 
     if a_dtype == 'f':
         gesvd = cusolver.sgesvd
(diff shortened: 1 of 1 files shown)


takagi:
It looks that the implementation of #2277 may make C-order to F-order copy and return F-order outputs in some cases, while the master always makes C-order to C-order copy and ensures to return C-order outputs. Let me check the change of #2277 makes sense.

emcastillo:
One thing is that in CuPy we don't ensure order of return types.
@asi1024 may confirm this 😇

takagi:
Ah, maybe C-order to F-order copy is similarly efficient with C-order to C-order copy?

takagi:
I've confirmed that it's yes.
```
c_to_c             :    CPU:   14.367 us   +/- 2.350 (min:   13.372 / max:  237.836) us     GPU: 1984.150 us   +/- 2.451 (min: 1981.888 / max: 2216.160) us
c_to_f             :    CPU:   14.484 us   +/- 0.700 (min:   13.432 / max:   24.075) us     GPU: 1984.240 us   +/- 0.779 (min: 1981.888 / max: 1995.136) us
```

emcastillo:
Great!!

takagi:
Sorry, I made mistake. I time it again. 😢

takagi:
No surprising 😅

```
c_to_c              :    CPU:   15.887 us   +/- 4.227 (min:   14.611 / max:  235.036) us     GPU: 1985.645 us   +/- 4.336 (min: 1983.328 / max: 2209.856) us
c_to_f              :    CPU:   16.712 us   +/- 3.623 (min:   15.408 / max:  226.488) us     GPU:10678.005 us   +/-12.046 (min:10634.464 / max:10898.976) us
```

```Python
import cupy
import cupyx.time

def c_to_c():
    a = cupy.empty((10000, 10000), order='C')
    return a.astype(a.dtype, order='C')

def c_to_f():
    a = cupy.empty((10000, 10000), order='C')
    return a.astype(a.dtype, order='F')

print(cupyx.time.repeat(c_to_c))
print(cupyx.time.repeat(c_to_f))
```

takagi:
Now this is ready for review. This change reuses intermediate `x` to reduce memory usage, while the output arrays may not be contiguous.

leofang:
I think if you are struggling with the CPU overheads in allocating temp workspace using `cupy.empty` etc, one way to get around is to replace all allocations with a plain `memory.alloc()`. It is a bit faster.

emcastillo:
> the CPU overheads in allocating temp workspace using `cupy.empty` etc, one way to get around is to replace all allocations with a plain `memory.alloc()`. It is a bit faster.

As long as it is documented in the code I am fine with this

takagi:
> I think if you are struggling with the CPU overheads in allocating temp workspace using cupy.empty etc, one way to get around is to replace all allocations with a plain memory.alloc(). It is a bit faster.

If you mean https://github.com/cupy/cupy/pull/3347#issuecomment-630055795, please scroll to right to see GPU time?

leofang:
Oh I see. I didn't scroll 😅 I simply saw you mentioned "...reduce `cupy.empty` calls..." and wondered.

cupy/linalg/decomposition.py
```diff
@@ -371,22 +371,23 @@ def svd(a, full_matrices=True, compute_uv=True):
     if compute_uv:
         if full_matrices:
             u = cupy.empty((m, m), dtype=a_dtype)
-            vt = cupy.empty((n, n), dtype=a_dtype)
+            vt = x
```
toslunar, line 374: I expect we can use the slice view
```suggestion
            vt = x[:, :n]
```
, which keeps `vt.data.ptr == x.data.ptr`, to eliminate the slices in Lines 416, 418.

takagi: Yes!

takagi:
Let me remove unnecessary `min(m, n)` in another PR.

toslunar:
Jenkins, test this please.

PR #3348, opened 2020-05-17
Add `cupy.sort_complex`

Dahlia-Chehata:


cupy/_sorting/sort.py
```diff
@@ -127,7 +127,25 @@ def msort(a):
...
+
+    Returns:
+        cupy.ndarray: sorted complex array.
+
+    .. seealso:: :func:`numpy.sort_complex`
+
+    """
+    if a.dtype.char == 'e':
+        a = a.astype('D')
+    a.sort()
```
asi1024, line 145: This line may change the input array. If it is intended behavior, please check also the input array after `xp.sort_complex` is called by using `testing.numpy_cupy_array_list_equal`.

Dahlia-Chehata: @asi1024 can you please review ?

tests/cupy_tests/sorting_tests/test_sort.py
```diff
@@ -416,6 +416,35 @@ def test_msort_multi_dim(self, xp, dtype):
...
+        for xp in (numpy, cupy):
+            a = testing.shaped_random((), xp)
+            with pytest.raises(numpy.AxisError):
+                xp.sort_complex(a)
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_list_equal()
+    def test_sort_complex_1dim(self, xp, dtype):
+        a = testing.shaped_random((100,), xp, dtype)
+        return xp.sort_complex(a)
```
asi1024, line 432: https://github.com/cupy/cupy/pull/3348/files#r429978698 
```suggestion
        return [a, xp.sort_complex(a)]
```

Dahlia-Chehata:
```
 if a.dtype.char == 'e':
     a = a.astype('D') 
```
These 2 lines were intended to support fp 16 (or I can suppress it from ``` testing.for_dtypes() ```) The suggested change ( return [a, xp.sort_complex(a)]) fails with all tests whether I keep or remove these 2 lines.

asi1024: So please fix your implementation to pass this check.

leofang: Hi @Dahlia-Chehata I've enabled `float16` sort early on, so if NumPy requires the output dtype to be `float64`, you should sort first and then cast for better performance.

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

PR #3354, opened 2020-05-19
Fix `linalg.svd` for 0-sized matrices

### LINKED ISSUE #2371: Incorrect behavior with empty arrays for `linalg.qr` and `linalg.svd`

toslunar:
Fix #2371.

### DIFF
--- a/cupy/linalg/decomposition.py
+++ b/cupy/linalg/decomposition.py
@@ -349,6 +349,20 @@ def svd(a, full_matrices=True, compute_uv=True):
     # Remark 4: Remark 2 is removed since cuda 8.0 (new!)
     n, m = a.shape
 
+    mn = min(m, n)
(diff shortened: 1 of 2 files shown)


takagi:
pfnCI, test this please.

takagi:
LGTM!

PR #3363, opened 2020-05-26
Add tests for optimizer for kernel launch parameters

asi1024:
Tests for #2731.
These tests will be skipped in our CI currently.

### DIFF
--- a/tests/cupyx_tests/test_optimize.py
+++ b/tests/cupyx_tests/test_optimize.py
@@ -40,7 +40,7 @@ class TestOptimize(unittest.TestCase):
             my_sum = cupy.ReductionKernel(
                 'T x', 'T out', 'x', 'a + b', 'out = a', '0', 'my_sum')
             my_sum_ = cupy.ReductionKernel(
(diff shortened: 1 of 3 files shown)


tests/cupyx_tests/test_optimize.py
```diff
@@ -0,0 +1,95 @@
...
+                assert optimize_impl.call_count == 2
+                my_sum(x_, axis=1)
+                assert optimize_impl.call_count == 2
+                my_sum(y, axis=1)
+                assert optimize_impl.call_count == 3
+                my_sum(z, axis=1)
+                assert optimize_impl.call_count == 4
+                my_sum_(x, axis=1)
+                assert optimize_impl.call_count == 5
+
```
emcastillo, line 66: the optimizer only increases the number of calls for different parameters?

emcastillo:
Jenkins, test this please

tests/cupyx_tests/test_optimize.py
asi1024: `call_count` is the total number of all function calls. So, `_optimize` is called at most once for each reduction routine operation.

Issue #3292, opened 2020-04-19
Issue adding `-arch` to RawModule options


mnicely:
```bash
CuPy Version          : 7.3.0
CUDA Root             : /usr/local/cuda
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSOLVER Version      : (10, 3, 0)
cuSPARSE Version      : 10301
NVRTC Version         : (10, 2)
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2406
NCCL Runtime Version  : 2507
```

When I try and add `"--gpu-architecture=compute_X0"`, where **X** is a compute major version, I get the following message. The whole command looks like `module2 = cp.RawModule(
            code=src, options=("-std=c++11", "-use_fast_math","--gpu-architecture=compute_30",)`

```bash
Traceback (most recent call last):
  File "/home/belt/anaconda3/envs/cusignal/lib/python3.8/site-packages/cupy/cuda/compiler.py", line 438, in compile
    nvrtc.compileProgram(self.ptr, options)
  File "cupy/cuda/nvrtc.pyx", line 101, in cupy.cuda.nvrtc.compileProgram
  File "cupy/cuda/nvrtc.pyx", line 111, in cupy.cuda.nvrtc.compileProgram
  File "cupy/cuda/nvrtc.pyx", line 56, in cupy.cuda.nvrtc.check_status
cupy.cuda.nvrtc.NVRTCError: NVRTC_ERROR_INVALID_OPTION (5)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/belt/anaconda3/envs/cusignal/lib/python3.8/site-packages/cusignal-0.14.0a0+179.g639fccf.dirty-py3.8.egg/cusignal/__init__.py", line 15, in <module>
    from cusignal.signaltools import (
  File "/home/belt/anaconda3/envs/cusignal/lib/python3.8/site-packages/cusignal-0.14.0a0+179.g639fccf.dirty-py3.8.egg/cusignal/signaltools.py", line 54, in <module>
    from ._upfirdn import upfirdn, _output_len
[...]

Issue #3301, opened 2020-04-22
Drop support for CUDA 8.0 and 9.1 in v8


kmaehashi:
As users of CUDA 8.0 and 9.1 are decreasing, we are planning to drop support for these CUDA versions.

leofang:
Note: This is in line with conda-forge's CUDA support. Currently we only build for CUDA 9.2+.

leofang:
Hi @gmarkall I read somewhere that Numba still supports CUDA 8.0. Am I right? Would this plan cause negative impacts?

kmaehashi:
We are going to drop support for CUDA 8.0 / 9.1 starting in CuPy v8.0.0rc1.

gmarkall:
> Hi @gmarkall I read somewhere that Numba still supports CUDA 8.0. Am I right? Would this plan cause negative impacts?

It only supports it because it has not been inconvenient to - I believe we could drop support for 8.0 without causing issues.

leofang:
Thanks, @gmarkall!

PR #3366, opened 2020-05-26
Get arch per device and support CUDA 9.2+ 

leofang:
I noticed the max compute capability used in the NVRTC support is slightly outdated, so I updated the values according to https://en.wikipedia.org/wiki/CUDA#GPUs_supported; I wasn't able to find a single table listing all these from CUDA's documentation, so it'd be worth double checking. 

Also, the global cache `_nvrtc_max_compute_capability` is replaced by a per-device cache to deal with heterogeneous environments (see https://github.com/cupy/cupy/issues/3292#issuecomment-616267757).

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -56,15 +56,18 @@ def _get_nvrtc_version():
 def _get_arch():
     # See Supported Compile Options section of NVRTC User Guide for
     # the maximum value allowed for `--gpu-architecture`.
+    # Note that the second-to-max values are used, as the max usually
+    # corresponds to Tegra/Jetson devices
+    # TODO(leofang): figure out a way to support Tegra/Jetson?
     major, minor = _get_nvrtc_version()
     if major < 9:
         # CUDA 8.0
-        _nvrtc_max_compute_capability = '62'
-    elif major < 10:
-        # CUDA 9.x
-        _nvrtc_max_compute_capability = '72'
+        _nvrtc_max_compute_capability = '52'
+    elif major < 10 or (major == 10 and minor == 0):
+        # CUDA 9.x / 10.0
+        _nvrtc_max_compute_capability = '70'
     else:
-        # CUDA 10.x
+        # CUDA 10.1 / 10.2
         _nvrtc_max_compute_capability = '75'
 
     return min(device.Device().compute_capability,
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -14,7 +14,6 @@ from cupy.cuda import runtime
 from cupy import util
 
 _nvrtc_version = None
-_nvrtc_max_compute_capability = None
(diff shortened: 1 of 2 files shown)


leofang:
https://docs.nvidia.com/cuda/archive/8.0/nvrtc/index.html#group__options

https://docs.nvidia.com/cuda/archive/9.0/nvrtc/index.html#group__options
https://docs.nvidia.com/cuda/archive/9.1/nvrtc/index.html#group__options
https://docs.nvidia.com/cuda/archive/9.2/nvrtc/index.html#group__options

https://docs.nvidia.com/cuda/archive/10.0/nvrtc/index.html#group__options
https://docs.nvidia.com/cuda/archive/10.1/nvrtc/index.html#group__options
https://docs.nvidia.com/cuda/nvrtc/index.html#group__options (currently points to 10.2)

leofang:
Ah, I think we could simplify the changes a bit if 8.0 is to be dropped (#3301).

kmaehashi:
Thank you!
You may have already noticed, but `compute_72` is used for Xavier so we want to avoid using '75' as the default. I think it's better to maintain a list of known Tegra CCs (72, 62, 53) and use it if the device returns it.

kmaehashi:
You can check this page for the full list of CCs.
https://developer.nvidia.com/cuda-gpus

leofang:
> I think it's better to maintain a list of known Tegra CCs (72, 62, 53) and use it if the device returns it.

Good point! Will do.

> You can check this page for the full list of CCs.
> https://developer.nvidia.com/cuda-gpus

Oh great! I've seen this long time ago but totally forgot about it...Thanks!

leofang:
@kmaehashi PTAL

kmaehashi:
pfnCI, test this please.

leofang:
Not sure the root of the errors...Let me take a look.

leofang:
Looks like the mock device has a bug? `arch = device.Device().compute_capability` returned `'60'` but it is supposed to be mocked to `'62'`.

https://jenkins.preferred.jp/job/chainer/job/cupy_pr/1301/TEST=cupy-py3,label=mn1-p100/testReport/junit/tests.cupy_tests.cuda_tests.test_compiler/TestNvrtcArch/test_get_arch_cuda9/

leofang:
Ah, the return value of `_get_arch` is already cached when we run other tests...Let me fix it.

leofang:
@kmaehashi e5e3360 should work

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

leofang:
Some network issue?

kmaehashi:
Let me try again...
pfnCI, test this please.

kmaehashi:
LGTM!
(ignoring failing test known as flaky: https://github.com/chainer/chainer/issues/8551)

leofang:
Thanks!

PR #3371, opened 2020-05-27
Add `cupy.convolve`

Dahlia-Chehata:


Dahlia-Chehata:
@asi1024 this is a primary version that still needs modifications. Could you please check the overall approach before resuming ?

leofang:
It was suggested in the discussion of #3184 that `cupy.convolve` could be implemented on top of what's done there. Is the plan changed now?

grlee77:
> It was suggested in the discussion of #3184 that cupy.convolve could be implemented on top of what's done there. Is the plan changed now?

The biggest issue with that is probably that SciPy's `convolve` always uses double precision internally and then casts the result to the dtype of the input.

NumPy instead determines appropriate dtypes based on both the input and filter and then performs the convolution using that dtype. 

Below are examples in how this leads to different behavior for the two.

case 1: same result, but different dtype
```
>>> np.convolve(np.arange(8, dtype=np.int16), -np.ones(3, dtype=np.uint16))
array([ 0,  1,  3,  6,  9, 12, 15, 18, 13,  7], dtype=int32)

>>>scipy.ndimage.convolve(np.arange(8, dtype=np.int16), np.ones(3, dtype=np.uint16))
array([ 0,  1,  3,  6,  9, 12, 15, 18, 13,  7], dtype=int16)
```

case 2: SciPy->floating point values truncated to the input dtype
```
np.convolve(np.arange(8, dtype=np.int64), np.full((3,), 3.1, dtype=np.float32))
>>> array([ 0.        ,  3.0999999 ,  9.29999971, 18.59999943, 27.89999914,
            37.19999886, 46.49999857, 55.79999828, 40.29999876, 21.69999933])

scipy.ndimage.convolve(np.arange(8, dtype=np.int64), np.full((3,), 3.1, dtype=np.float32))
>>> array([ 3,  9, 18, 27, 37, 46, 55, 61])
```

case 3: casting negative values to uint16 in SciPy causes unexpected result
```
np.convolve(np.arange(8, dtype=np.uint16), -np.ones(3, dtype=np.int16))
>>> array([  0,  -1,  -3,  -6,  -9, -12, -15, -18, -13,  -7], dtype=int32)

scipy.ndimage.convolve(np.arange(8, dtype=np.uint16), -np.ones(3, dtype=np.int16))
>>> array([65535, 65533, 65530, 65527, 65524, 65521, 65518, 65516], dtype=uint16)
```

case 4: float16 not supported in SciPy
```
np.convolve(np.arange(8, dtype=np.float16), np.ones(3, dtype=np.float16))
>>> array([ 0,  1,  3,  6,  9, 12, 15, 18, 13,  7], dtype=float16)

scipy.ndimage.convolve(np.arange(8, dtype=np.float16), np.ones(3, dtype=np.float16))
>>> RuntimeError: array dtype not supported
```

asi1024:
@leofang @grlee77 Another implementation seems to be needed for SciPy's `convolve`.

asi1024:
@Dahlia-Chehata Does it work for non-contiguous input?

Dahlia-Chehata:
> @Dahlia-Chehata Does it work for non-contiguous input?

Yes, but the input  should be converted to contiguous form first

asi1024:
The current implementation does not support complex dtype inputs.

asi1024:
Now I doubt you need to use `RawKernel`. How about using `ReductionKernel`?

Dahlia-Chehata:
> Now I doubt you need to use `RawKernel`. How about using `ReductionKernel`?

The thing is that I need to use dynamic parallelism. The child kernel is invoked with different values for nTPB or number of blocks each time. These values depend on the parent's ` tid`. So can this be achieved using `ReductionKernel`?

asi1024:
In my recognition, dynamic parallelism is often used to speed up recursive operations and does not so good for a simple loop. Actually, the algorithm of your implementation is essentially equivalent to the matrix dot, so the classic matrix multiplication algorithm seems better for the performance.

asi1024:
The current implementation does not pass the following test.
```py
    def test_large(self, xp):
        a = testing.shaped_arange((4000,), xp, numpy.float64)
        b = testing.shaped_arange((4000,), xp, numpy.float64)
        return xp.correlate(a, b, mode=self.mode)
```

Dahlia-Chehata:
> The current implementation does not pass the following test.
> 
> ```python
>     @testing.numpy_cupy_array_equal()
>     def test_large(self, xp):
>         a = testing.shaped_arange((4000,), xp, numpy.float64)
>         b = testing.shaped_arange((4000,), xp, numpy.float64)
>         return xp.correlate(a, b, mode=self.mode)
> ```

Yes, I know. There are some issues I am still resolving as for example checking if 
```
 if (tid < len){
}
```

Dahlia-Chehata:
> In my recognition, dynamic parallelism is often used to speed up recursive operations and does not so good for a simple loop. Actually, the algorithm of your implementation is essentially equivalent to the matrix dot, so the classic matrix multiplication algorithm seems better for the performance.

So what is your suggestion ?
Using Reductionkernel for dot product or (maybe it is an existing ufunc.... will look into it) and call it inside a for loop for each of 3 cases: left, middle and right ?

asi1024:
Using `ReductionKernel`.

Dahlia-Chehata:
I know it is early to check the performance (since I am still  fixing some bugs and have not pushed the latest fix yet) but I wanted to check whether the performance will be really worse than numpy's

| n | numpy | cupy |
| :---: | :---: | :---: |
| 100 | 10.331 | 149.945 |
| 1000 | 585.134 | 208.380 |
| 10000 | 59404.040 |  266.231 |
| 100000 | 5892395.947 | 280.500 |

coderforlife:
@Dahlia-Chehata are those numbers times or speeds? Looks like maybe times. What units? Is that `n` for a cross-correlation?

Dahlia-Chehata:
> @Dahlia-Chehata are those numbers times or speeds? Looks like maybe times. What units? Is that `n` for a cross-correlation?

time in us ( didn't change a thing in `cupyx.time.repeat`)
Yes, n is the size of 1D inputs of correlation

cupy/statistics/correlation.py
```diff
@@ -48,7 +59,83 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
...
+    output = cupy.zeros(length, dtype)
+    a1 = a1.astype(dtype, copy=False)
+    a2 = a2.astype(dtype, copy=False)
+    start = j1 = 0
+    j2 = left
+    n -= left
+    for i in range(left):
+        dot_kernel(a1[:n], a2[j2:], output[i])
+        n += 1
+        j2 -= 1
```
Dahlia-Chehata, line 113: @asi1024  Can this be parallelized ?
I understand that Reduction kernel parallelizes along reduction-axis and also elementwise one. But I am asking about sliding window with each iteration. I thought about unrolling the sliding window over 1D input in a 2D form but this seems inefficient. (Which made me use dynamic parallelism in the first place ...)

Dahlia-Chehata:
`cupy.fft.fft` have different precisions than` scipy.signal.fft` or `numpy.fft.fft` which causes erroneous results with large size in `_fftconvolve1d`.

cupyx/scipy/signal/signaltools.py
```diff
@@ -0,0 +1,78 @@
...
+        nor multidimensional inputs.
+
+    .. seealso:: :func:`scipy.signal.choose_conv_method`
+
+    """
+    if in1.ndim != 1 or in2.ndim != 1:
+        raise NotImplementedError('Only 1d inputs are supported currently')
+
+    if any([_numeric_arrays([x], kinds='ui') for x in [in1, in2]]):
+        max_value = int(cupy.abs(in1).max()) * int(cupy.abs(in2).max())
```
asi1024, line 36: This line requires host-device synchronization, so it causes performance degradation.
How about using always `direct` mode in integer cases?

cupyx/scipy/signal/signaltools.py
```diff
@@ -0,0 +1,78 @@
...
+        max_value *= int(min(in1.size, in2.size))
+        if max_value > 2 ** cupy.finfo('float').nmant - 1:
+            return 'direct'
+
+    if _numeric_arrays([in1, in2], kinds='b'):
+        return 'direct'
+
+    if _numeric_arrays([in1, in2]):
```
asi1024, line 44: `_numeric_arrays([in1, in2])` always returns `True` because `cupy.ndarray` supports only for numeric elements.

cupyx/scipy/signal/signaltools.py
```diff
@@ -0,0 +1,78 @@
...
+            return 'fft'
+
+    return 'direct'
+
+
+def _fftconv_faster(x, h, mode):
+    fft_ops, direct_ops = _conv_ops(x.size, h.size, mode)
+    offset = -1e-3
+    constants = {
+        "valid": (1.89095737e-9, 2.1364985e-10, offset),
```
asi1024, line 55: Please use single quote for the consistency.
```suggestion
        'valid': (1.89095737e-9, 2.1364985e-10, offset),
```

cupyx/scipy/signal/signaltools.py
```diff
@@ -0,0 +1,78 @@
...
+        return 'direct'
+
+    if _numeric_arrays([in1, in2]):
+        if _fftconv_faster(in1, in2, mode):
+            return 'fft'
+
+    return 'direct'
+
+
+def _fftconv_faster(x, h, mode):
```
asi1024, line 51: Could you explain what does this function do? If this is from SciPy's implementation, please add a reference.

cupyx/scipy/signal/signaltools.py
```diff
@@ -0,0 +1,78 @@
...
+    elif mode == "valid":
+        direct_ops = (siz2 - siz1 + 1) * siz1 \
+            if siz2 >= siz1 else (siz1 - siz2 + 1) * siz2
+    elif mode == "same":
+        direct_ops = siz1 * siz2 if siz1 < siz2\
+            else siz1 * siz2 - (siz2 / 2) * ((siz2 + 1) / 2)
+    else:
+        raise ValueError('Unsupported mode')
+    N = siz1 + siz2 - 1
+    fft_ops = 3 * N * cupy.log(N)
```
asi1024, line 77: Please avoid using a cupy routine for 1-size array for the performance.
```suggestion
    fft_ops = 3 * N * math.log(N)
```

cupyx/scipy/signal/signaltools.py
```diff
@@ -0,0 +1,78 @@
...
+    }
+    O_fft, O_direct, O_offset = constants[mode]
+    return O_fft * fft_ops < O_direct * direct_ops + O_offset
+
+
+def _conv_ops(siz1, siz2, mode):
+    if mode == "full":
+        direct_ops = siz1 * siz2
+    elif mode == "valid":
+        direct_ops = (siz2 - siz1 + 1) * siz1 \
```
asi1024, line 69: Do not use \ for line continuation. (See [our development policies](https://github.com/cupy/cupy/wiki/Development-policies))

cupyx/scipy/signal/signaltools.py
```diff
@@ -0,0 +1,78 @@
...
+def _numeric_arrays(arrays, kinds='buifc'):
+    if type(arrays) == cupy.ndarray:
+        return arrays.dtype.kind in kinds
+    for array in arrays:
+        if array.dtype.kind not in kinds:
+            return False
+    return True
+
+
+def choose_conv_method(in1, in2, mode='full'):
```
asi1024, line 13: Please add tests for `scipy.signal.choose_conv_method`.

cupy/statistics/correlation.py
asi1024: Create two arrays of shape `(out_size, max(a1.size, a2.size))`, and reduce with the second axis.

cupy/statistics/correlation.py
```diff
@@ -48,7 +59,83 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
...
+        if inverted:
+            out = out[::-1]
+    elif method == 'fft':
+        out = cupy.math.misc._fftconvolve1d(a, conj_v, mode)
+    else:
+        raise ValueError('Unsupported method')
+    return out
+
+
+def _dot_correlate(a1, a2, mode):
```
asi1024, line 93: Why `dot_correlate` is placed in `cupy.statistics.correlation` while `_fftconvolve1d` is in `cupy.math.misc`? How about moving the implementations in `math.misc` with renaming to `_dot_convolve`?

Dahlia-Chehata: `dot_correlate` is placed  in `cupy.statistics.correlation` because `correlate` inputs are passed directly to it with reversing one of the inputs which is needed when used from `convolve`.
The same goes for `_fftconvolve1d`

cupyx/scipy/signal/signaltools.py
Dahlia-Chehata: Yes, it is an internal function in SciPy but I limited its functionality to 1D inputs.
Will add some reference

cupy/statistics/correlation.py
Dahlia-Chehata: @asi1024 Should I still move `_dot_correlate` to `misc` as `_dot_convolve` ?

asi1024:
This PR is blocked by #3464.

cupy/statistics/correlation.py
Dahlia-Chehata: I don't think I get this. This is what I intended to do
``` python
    n1 = a1.size
    n2 = a2.size
    start = 0
    for i in range(left):
        dot_kernel(a1[:n2 - left + i], a2[left - i:], output[i])
    start += left
    A1 = _rolling_window(a1, n2)
    A2 = cupy.repeat(a2[None, :], n1 - n2 + 1, axis=0)
    dot_kernel(A1, A2, output[start:(start + n1 - n2 + 1)], axis=1)
    start += n1 - n2 + 1
    for i in range(right):
        dot_kernel(a1[n1 - n2 + 1 + i:], a2[:n2 - 1 - i], output[start + i])
```
where n1  >=  n2 and `_rolling_window` is

```python
def _rolling_window(a, window):
    shape = a.shape[:-1] + (a.shape[-1] - window + 1, window)
    strides = a.strides + (a.strides[-1],)
    return cupy.lib.stride_tricks.as_strided(a, shape=shape, strides=strides)
```

Dahlia-Chehata:
So what's the plan https://github.com/cupy/cupy/pull/3464#discussion_r443613883 ?

Dahlia-Chehata:
CuPy and SciPy have differences in `fft` computations which propagate to the final result of _fftconvolve1d, specially for large sized inputs.
For example:

```
a = testing.shaped_arange((10000,), xp, xp.float32)
b = testing.shaped_arange((100,), xp, xp.float32)
shape = a.size + b.size - 1
fshape = scipy.fft.next_fast_len(shape, True)
a = scp.fft.rfft(a, fshape)
b = scp.fft.rfft(b, fshape)
out = scp.fft.irfft(a * b, fshape)
```
 `out` results for scipy are different from those of cupyx.scipy

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+        raise ValueError('Unsupported method')
+    return out
+
+
+def _fftconvolve1d(a1, a2, mode='full'):
+
+    inverted = 0
+    if a2.size > a1.size:
+        a1, a2 = a2, a1
+        inverted = not a2.size % 2
```
Dahlia-Chehata, line 64: L61-L64 are used for correlation function and has no impact on convolution

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+
+
+def _dot_convolve(a1, a2, mode):
+    inverted = 0
+    dtype = cupy.result_type(a1, a2)
+    if a1.size == 0 or a2.size == 0:
+        raise ValueError('Array arguments cannot be empty')
+    if a1.size < a2.size:
+        a1, a2 = a2, a1
+        inverted = 1
```
Dahlia-Chehata, line 106: Ditto

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+    inverted = 0
+    if a2.size > a1.size:
+        a1, a2 = a2, a1
+        inverted = not a2.size % 2
+    siz1 = a1.size
+    siz2 = a2.size
+    shape = siz1 + siz2 - 1
+    complex = a1.dtype.kind == 'c' or a2.dtype.kind == 'c'
+    # TODO: Replace with cupyx.scipy.fft.next_fast_len
+    fshape = scipy.fft.next_fast_len(shape, not complex)
```
asi1024, line 70: Don't use scipy's function because SciPy is not in the install requirements of CuPy.

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+        newsize = siz1 - siz2 + 1
+        start = (out.size - newsize) / 2
+        end = start + newsize
+        out = out[start: end]
+    else:
+        if mode != 'full':
+            raise ValueError('acceptable mode flags are `valid`,'
+                             ' `same`, or `full`.')
+    result_type = cupy.result_type(a1, a2)
+    if result_type.kind in {'u', 'i'}:
```
asi1024, line 94: `set` is not fast for small case.
```suggestion
    if result_type.kind in 'iu':
```

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+        start = (out.size - siz1) / 2 + inverted
+        end = start + siz1
+        out = out[start: end]
+    elif mode == 'valid':
+        newsize = siz1 - siz2 + 1
+        start = (out.size - newsize) / 2
+        end = start + newsize
+        out = out[start: end]
+    else:
+        if mode != 'full':
```
asi1024, line 90:
```py
    elif mode != 'full':
```

cupy/math/misc.py
asi1024: BTW, this performance improvement should be implemented in `scipy.fft`, so it is not needed here.

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+def _fftconvolve1d(a1, a2, mode='full'):
+
+    inverted = 0
+    if a2.size > a1.size:
+        a1, a2 = a2, a1
+        inverted = not a2.size % 2
+    siz1 = a1.size
+    siz2 = a2.size
+    shape = siz1 + siz2 - 1
+    complex = a1.dtype.kind == 'c' or a2.dtype.kind == 'c'
```
asi1024, line 68: Please rename.
```suggestion
    is_complex = a1.dtype.kind == 'c' or a2.dtype.kind == 'c'
```

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+    # TODO: Replace with cupyx.scipy.fft.next_fast_len
+    fshape = scipy.fft.next_fast_len(shape, not complex)
+    if not complex:
+        fa1 = fft.rfft(a1, fshape)
+        fa2 = fft.rfft(a2, fshape)
+        out = fft.irfft(fa1 * fa2, fshape)
+    else:
+        fa1 = fft.fft(a1, fshape)
+        fa2 = fft.fft(a2, fshape)
+        out = fft.ifft(fa1 * fa2, fshape)
```
asi1024, line 78: For the simplification.
```suggestion
    if is_complex:
        fa1 = fft.fft(a1, fshape)
        fa2 = fft.fft(a2, fshape)
        out = fft.ifft(fa1 * fa2, fshape)
    else:
        fa1 = fft.rfft(a1, fshape)
        fa2 = fft.rfft(a2, fshape)
        out = fft.irfft(fa1 * fa2, fshape)
```

cupy/math/misc.py
```diff
@@ -1,8 +1,169 @@
...
+    elif method == 'fft':
+        out = _fftconvolve1d(a, v, mode)
+    else:
+        raise ValueError('Unsupported method')
+    return out
+
+
+def _fftconvolve1d(a1, a2, mode='full'):
+
+    inverted = 0
```
asi1024, line 61: Please rename to `is_inverted` and handle as a Boolean value.
```suggestion
    is_inverted = False
```

cupy/math/misc.py
asi1024, line 64: For the readability.
```suggestion
        inverted = a2.size % 2 == 1
```

cupy/math/misc.py
Dahlia-Chehata: Eliminating this part requires to change `rtol` to `1e-2` for `test_convolve_inverted_odd_len`

cupy/math/misc.py
```diff
@@ -1,8 +1,122 @@
...
+    if method == 'direct':
+        _, out = _dot_convolve(a, v, mode)
+    elif method == 'fft':
+        out = _fft_convolve(a, v, mode)
+    else:
+        raise ValueError('Unsupported method')
+    return out
+
+
+def _fft_convolve(a1, a2, mode='full'):
```
asi1024, line 56: Default value is not needed here.
```suggestion
def _fft_convolve(a1, a2, mode):
```

asi1024:
Jenkins, test this please.

asi1024:
Could you remove this line?
https://github.com/cupy/cupy/blob/8f2b9fbc19f465fdcfe12b311536807bbeb6370d/tests/cupyx_tests/fallback_mode_tests/test_notifications.py#L45
The test `TestNotificationModes` seems to assume that `cupy.convolve` is not implemented.

asi1024:
Jenkins, test this please.

tests/cupy_tests/math_tests/test_misc.py
```diff
@@ -210,3 +210,70 @@ def test_nan_to_num_nan(self):
...
+
+    @testing.for_all_dtypes(no_float16=True)
+    @testing.numpy_cupy_allclose(rtol=1e-3)
+    def test_convolve_large_non_contiguous(self, xp, dtype):
+        a = testing.shaped_arange((10000,), xp, dtype)
+        b = testing.shaped_arange((100,), xp, dtype)
+        return xp.convolve(a[200::], b[10::70], mode=self.mode)
+
+    @testing.for_all_dtypes_combination(
+        names=['dtype1', 'dtype2'], no_float16=True, no_complex=True)
```
asi1024, line 252: Don't skip float16 and complex type cases for the dtype combination check.
```suggestion
        names=['dtype1', 'dtype2'])
```

ke-zhang-rd:
This looks nice but why only addresses 1d not Nd? 

I notice [numpy.convolve](https://numpy.org/doc/stable/reference/generated/numpy.convolve.html) focus on 1d but I believe that is because of numpy and scipy negotiation which let them focus separately implement `convolve` and scipy implements Nd convolve. While cupy is oligopoly in this domain, I feel cupy need to cover Nd.

asi1024:
@ke-zhang-rd CuPy aims to support NumPy compatible interface, so we won't support Nd array inputs in `cupy.convolve`. We will support it in `cupyx.scipy.signal.convolve` in the future.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

ke-zhang-rd:
> @ke-zhang-rd CuPy aims to support NumPy compatible interface, so we won't support Nd array inputs in `cupy.convolve`. We will support it in `cupyx.scipy.signal.convolve` in the future.

@asi1024 Ok, is there anyone implementing that? If not, I could help.

coderforlife:
@ke-zhang-rd See #3526 which shows that this particular implementation of `convolve` is extremely slow, especially for smaller inputs. It may be a better approach to utilize the much faster `cupyx.scipy.ndimage.convolve` function (or at least it's approach) to implement `cupyx.scipy.signal.convolve`. In fact, if you are willing to copy the input array when `mode='full'` (by using the `cupy.pad` function), there is already quick-and-dirty implementations of `cupyx.scipy.signal.convolve` using `cupyx.scipy.ndimage.convolve`.

I do have plans on making `cupyx.scipy.ndimage.convolve` usable for `mode='full'` directly and thus you can use it for those other implementations.

asi1024:
@coderforlife CuPy aims to handle large inputs, so if you want to talk about performance, you should compare on large inputs. (It is better to have good performance even with small inputs of course! :smile:)
Please note that It is hard to say which one is better depending on the shape of inputs.

script:
```py
import numpy
import scipy.ndimage

import cupy
import cupyx
import cupyx.scipy.ndimage

a_shapes = [10, 2000000]
v_shapes = [2000000, 10]

for a_shape, v_shape in zip(a_shapes, v_shapes):
    a = cupy.arange(a_shape, dtype=numpy.int32)
    v = cupy.arange(v_shape, dtype=numpy.int32)

    perf = cupyx.time.repeat(
        cupy.convolve, (a, v, 'same'),
        max_duration=1)
    print(perf)
    perf = cupyx.time.repeat(
        cupyx.scipy.ndimage.convolve1d, (a, v), {'mode': 'constant'},
        max_duration=1)
    print(perf)
```

Results:
```
convolve            :    CPU:   92.901 us   +/-40.204 (min:   86.524 / max: 1544.412) us     GPU-0:  731.759 us   +/-46.062 (min:  711.104 / max: 2168.384) us
convolve1d          :    CPU:  133.567 us   +/-115.531 (min:   58.519 / max:  333.251) us     GPU-0:266243.317 us   +/-120.219 (min:266173.096 / max:266451.538) us
convolve            :    CPU:   89.753 us   +/-11.573 (min:   86.483 / max:  509.197) us     GPU-0:  715.934 us   +/-11.555 (min:  712.544 / max: 1135.040) us
convolve1d          :    CPU:   49.686 us   +/- 3.497 (min:   47.859 / max:  328.524) us     GPU-0:   87.957 us   +/- 3.338 (min:   84.608 / max:  363.008) us
```

coderforlife:
@asi1024 The `scipy.ndimage` functions intentionally do not swap `a` or `v` to force `a` being larger like `cupy.convolve` does. This is required for `scipy` compatibility. However, if `cupy.convolve` was made as a wrapper of them, then this simple statement could be added to the wrapper: `if a.size < v.size: a,v = v,a`. This line already exists in `cupy.convolve` in fact. Thus, the 360x slower case would not be seen ever.

See the basic wrapper at https://github.com/cupy/cupy/issues/3414#issuecomment-644967792 and test against that. That wrapper does support `'full'` and `'valid'` modes as well, but `'full'` is done in such a way it will incur a penalty - I am currently making a version of the `convolve` function based on the `scipy.ndimage` functions that will natively support `'full'` and `'valid'` and thus incur no penalties for those modes. It currently works for the 1D `cupy.convolve` function but needs some fixes to work for the nd `cupyx.scipy.signal.convovle` functions.

coderforlife:
@asi1024 And yes, I regular test on large arrays. A basic 4MP image is twice as large as the size you used.

The only times when `cupy.convolve` is good competition is when `a.size` is very close to `v.size` and they are fairly large.

leofang:
I don't speak for @asi1024, but the objection for which I (and @ke-zhang-rd who is working on a FFT-based `cupyx.scipy.signal.convovle`) would have is to your [earlier statement](https://github.com/cupy/cupy/pull/3371#issuecomment-652559445)
> See #3526 which shows that this particular implementation of convolve is extremely slow, especially for smaller inputs. 

Calling it "extremely slow" is very unfair to @Dahlia-Chehata's contribution. The point is CuPy focuses the performance for large inputs, and there requires some different, specialized considerations and approaches for small inputs. Often times, for the input sizes you tested in #3526 I'd just recommend to use SciPy directly, as GPUs won't win CPUs overwhelmingly and SciPy has been well-optimized.

PR #3357, opened 2020-05-20
Support `max_duration` argument in `cupyx.time.repeat`

asi1024:

PR #3375, opened 2020-05-27
Fix `time.repeat` to accumulate duration after GPU synchronization

asi1024:
Fix a bug in #3357.


emcastillo:
Jenkins, test this please

PR #3379, opened 2020-05-28
Fix windows

emcastillo:
When building v7 with cuda 8 in windows:
```
 fatal error C1128: number of 
sections exceeded object file format limit: compile with /bigobj
```

This just adds bigobj to all the builds in windows for v7, the only drawback for this option seems to be that older linkers can't consume the generated objects.
https://stackoverflow.com/questions/15110580/penalty-of-the-msvs-compiler-flag-bigobj


kmaehashi:
pfnCI, test this please.

emcastillo:
？

PR #3386, opened 2020-05-31
Move CUDA low-level API to `cupy_cuda` namespace

### LINKED ISSUE #3385: Move CUDA Low Level API to other namespace
The `cupy.cuda` namespace is confusing.

1. `cupy.cuda` is not in NumPy
2.  Mixing Low Level API and CuPy original wrapper

Especially solving the second problem simplifies CuPy development.
(The first problem is a bit difficult because it requires design discussion.)

The implementation of some features on CuPy does not provide CUDA's Low Level API. Therefore, I think it is better to start with partial separation first.

With this change, CuPy may be able to provide Cython CUDA API in the future #130.

okuta:
Fix #3385


kmaehashi:
Note: new package `cupy_cuda` is used instead of `cupyx` to avoid circular references between `cupyx` and `cupy`.

kmaehashi:
Things to discuss:

## Module hierarchy and naming

Currently we have both low-level (e.g., `cupy.cuda.cutensor`) and high-level (e.g., `cupy.cuda.memory`) APIs in `cupy.cuda`.
We need to put these two under `cupy_cuda`, with appropriate naming and module hierarchy.

## Default Stream

Currently the "default" stream are automatically set in low-level APIs (e.g., `cupy.cuda.cudnn.convolutionForward`), but it would be better to provide a way to call low-level APIs without setting streams.

## Backward compatibility

`cupy.cuda` is widely used in user projects so it we should not break backward compatibility (I guess implementation in this PR satisfies this).

emcastillo:
Let me take over this and rebase changes, I will push directly to your branch, thanks!

emcastillo:
Fixed the compilation errors and some skipped tests

emcastillo:
I will kick the CIs just to check all works.  Not merging yet.
Jenkins, test this please

emcastillo:
Jenkins, test this please.

okuta:
Jenkins, test this please

okuta:
Jenkins, test this please

okuta:
Jenkins, test this please

kmaehashi:
As per the offline discussions:
* we may have different backends in future
* CUDA runtime/driver needs to be separated from other toolkit libraries

Design:

* `cupy_backends/cuda`  … code that uses CUDA API (including runtime/driver/library)
* `cupy_backends/cuda/api` … thin wrappers for CUDA runtime/driver APIs
* `cupy_backends/cuda/libs`  … thin wrappers for CUDA library API (cuBLAS, cuSparse, cuTensor, cuDNN, CUB, etc.)

okuta:
Jenkins, test this please

okuta:
jenkins, test this please.

leofang:
Hi @okuta, I need to understand the changes here better. Why do we have `get_current_stream_ptr`, for example, in both `cupy.cuda.stream` and `cupy_backends.cuda.stream`?

okuta:
Thank you for pointing out. This is an intermediate measure for moving APIs.
I think it is appropriate to combine streams into one in the current directory structure.

okuta:
I sent new PR #3575

leofang:
I think it's #3584 🙂

Another question for @emcastillo: I noticed this PR is labeled no-compat, but I don't see why.

emcastillo:
The import paths changed, so people directly using the cuda wrapping code will have to change their codes

leofang:
But isn't everything moved to `cupy_backends` is imported back to `cupy.cuda`?

okuta:
Some functions have a dependency on ndarray.
The modules containing them cannot be moved. The other ones can be moved.
- device
- compiler
- etc.


If there is something I want to change at the same time, I want to do it at this timing.

emcastillo:
> But isn't everything moved to cupy_backends is imported back to cupy.cuda?

As far as I remember, the intention is to remove them after a few iterations.

PR #3392, opened 2020-06-02
[WIP] CUDA JIT

asi1024:


leofang:
Very interesting!!!

cc: @jakirkham @mnicely @gmarkall @sklam @beckermr

emcastillo:
I was curious about how to deal with variable types using ast and found the following

https://stackoverflow.com/questions/8425371/how-to-get-a-type-of-the-variable-from-the-pythons-ast

This is just for reference and to spark discussion

takagi:
I suppose that an option is we can wrap `ast` nodes to have type variables, and infer the types of those nodes:

```
class Constant:
    def __init__(self, node):
        self.node = node           # An instance of ast.Constant
        self.type_var = TypeVar()  # An type variable to infer
```

Fortunately, as the target language is CUDA C, we don't need to think of the dynamic feature of Python.

leofang:
Question: Is it possible to provide an API to dump generated CUDA source code? Something like this:
```python
@jit.cuda_function()
def kernel(x, y):
     tid = jit.blockIdx.x * jit.blockDim.x + jit.threadIdx.x
     x[tid] += y

code = kernel.dump_source([('int', 'int'])  # specify that x and y are int*; could also be done in @jit.cuda_function?
print(code)
```
Expected output:
```
__global__ void kernel(int* x, int* y) {
    int tid = blockIdx.x * blockDim.x + threadIdx.x;
    x[tid] += y;
}
```
This could be useful for debugging as well as other purposes, like cuFFT callback (#4141).

asi1024:
@leofang Yes, we are planning to support it!

asi1024:
Moved the task list to #4290.

PR #3393, opened 2020-06-02
Reduce `numpy.ndarray` creation in cuTENSOR operation preparation

asi1024:
This PR avoids creating `numpy.ndarray` for each operation.

TODO:
- [x] Fix documentations
- [x] Fix examples
- [x] Enhance test coverage

### DIFF
--- a/cupy/cutensor.py
+++ b/cupy/cutensor.py
@@ -34,6 +34,7 @@ class Mode(object):
 
     def __init__(self, mode):
         self.mode = numpy.array(mode, dtype=numpy.int32)
+        assert self.mode.ndim == 1
 
     @property
     def ndim(self):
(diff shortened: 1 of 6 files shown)


anaruse:
I think it's good, but could you change the cuTensor examples in https://github.com/cupy/cupy/tree/master/examples/cutensor to also use `create_mode()`?

asi1024:
I re-fixed to allow tuple of string as a `mode` for compatibility (addf884), but I'll change the examples because I want to recommend to create `Mode` for the CPU performance.

asi1024:
@anaruse @kmaehashi Updated documentations and examples. PTAL.

kmaehashi:
pfnCI, test this please.

PR #3394, opened 2020-06-02
Fix cuTENSOR memoization

asi1024:


cupy/cutensor.py
```diff
@@ -30,13 +29,11 @@ def __del__(self, is_shutting_down=util.is_shutting_down):
             self.value = None
 
 
+@util.memoize(for_each_device=True)
 def get_handle():
-    dev = device.get_device_id()
-    if dev not in _handles:
-        handle = cutensor.Handle()
-        cutensor.init(handle)
-        _handles[dev] = handle
```
emcastillo, line 38: should we delete the `_handles` declaration in the top of the file?

asi1024: Fixed.

PR #3404, opened 2020-06-07
Add `polyvander`

Dahlia-Chehata:
cupy.polynomial.polynomial.polyvander


Dahlia-Chehata:
for degree = 100
| n | numpy | cupy |
| :---: | :---: | :---: |
| 100 | 159.712 | 1469.048 |
| 1000 | 268.832 | 1410.134 |
| 10000 | 1671.103 |  1344.060 |
| 100000 | 28912.964 | 1381.289 |
| 1000000 | 434710.022 | 1418.192 |

Dahlia-Chehata:
Complex64 have different precision than numpy's so I suppressed complex tests (for now)

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,27 @@
...
+        x (cupy.ndarray): array of points
+        deg (int): degree of the resulting matrix.
+
+    Returns:
+        cupy.ndarray: The Vandermonde matrix
+
+    .. seealso:: :func:`numpy.polynomial.polynomial.polyvander`
+
+    """
+    deg.__index__()
```
asi1024, line 17: What is this?

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,27 @@
...
+    if deg < 0:
+        raise ValueError('deg must be non-negative')
+    if not isinstance(x, cupy.ndarray) or x.ndim != 1:
```
asi1024, line 20: Is this condition needed?

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,27 @@
...
+        x = cupy.array(x, copy=False, ndmin=1)
+    x = x + 0.0
```
asi1024, line 22: What is this?

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,27 @@
...
+    dims = (deg + 1,) + x.shape
+    v = cupy.ones(dims, dtype=x.dtype)
+    for i in range(1, deg + 1):
+        cupy.multiply(v[i-1], x, v[i])
```
asi1024, line 26: How about using `cupy.cumsum` to parallelize these operations?

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,27 @@
...
+    .. seealso:: :func:`numpy.polynomial.polynomial.polyvander`
+
+    """
+    deg.__index__()
+    if deg < 0:
+        raise ValueError('deg must be non-negative')
+    if not isinstance(x, cupy.ndarray) or x.ndim != 1:
+        x = cupy.array(x, copy=False, ndmin=1)
+    x = x + 0.0
+    dims = (deg + 1,) + x.shape
```
asi1024, line 23: `shape` is preferred.
```suggestion
    shape = (deg + 1,) + x.shape
```

cupy/polynomial/polynomial.py
Dahlia-Chehata: for float with integral value case

cupy/polynomial/polynomial.py
Dahlia-Chehata: which one :sweat_smile: ?

Dahlia-Chehata: if you mean the second condition, yes it is needed because tests fail for float64 due to precision disparity if not used

cupy/polynomial/polynomial.py
Dahlia-Chehata: All types must have floating point entries in the output matrix

cupy/polynomial/polynomial.py
Dahlia-Chehata: ``` v  = cupy.cumprod(cupy.multiply(v[1:], x), axis = 0) ``` ?

cupy/polynomial/polynomial.py
asi1024: Both.
The former: This input array is expected to be a cupy.ndarray.
The latter: Please use `testing.numpy_cupy_allclose` to check array equality with tolerances.

cupy/polynomial/polynomial.py
asi1024: Please cast with `x.astype(dtype, copy=False)`.

cupy/polynomial/polynomial.py
asi1024: For `float` type `deg` case?

Dahlia-Chehata: yes

cupy/polynomial/polynomial.py
asi1024: Ah, `x ** cupy.arange(deg)` seems to be better.

cupy/polynomial/polynomial.py
asi1024: Hmm, `float` object has no attribute `__index__`. What do you mean?

Dahlia-Chehata: I meant **`operator.index(a)`** which is equivalent to `a.__index__()` acoording to [this](https://docs.python.org/3/library/operator.html#operator.index)

Dahlia-Chehata: Will I need to implement _deprecate_as_int() as described [here](https://fossies.org/linux/numpy/numpy/polynomial/polyutils.py) L766

cupy/polynomial/polynomial.py
Dahlia-Chehata: Why the first condition is not necessary. Yes, the input array is expected to be ndarray. What does this have to do with checking if degree is negative I didn't get that?

Dahlia-Chehata:
``` if  not x.ndim:
       x = cupy.array(x, copy=False, ndmin=1)
```
This condition is needed for 0dim or scalar input

asi1024: `not isinstance(x, cupy.ndarray)` is always `False`, so it is not necessary.

cupy/polynomial/polynomial.py
Dahlia-Chehata: casting with `x.astype(dtype, copy=False)` where dtype = `float` or `numpy.float32` or `numpy.float64` causes changes in floating point precision. Should I check whether the input is 32-bit or 64-bit float ?
I used `+ 0.0` for easy promotion between different precisions

cupy/polynomial/polynomial.py
Dahlia-Chehata: I have 2 questions regarding the current implementation that I need a little help with:
1) The precision for larger size of x is different than  numpy's. Why can be the reason for this ?   
 _( I had to change some values as in `test_polyvander ` to pass the tests until finding the reason for these differences)_ 

2) The precision for complex types is always different (with the loop approach or the suggested [one](https://github.com/cupy/cupy/pull/3404#discussion_r437325708)). Why ?

( This causes tests' failure even with using `testing.numpy_cupy_allclose()` )

cupy/polynomial/polynomial.py
Dahlia-Chehata: I have already put it in `polyutils` to have the same behaviour as numpy. If it is not needed, we can support integers and raise `TypeError `otherwise while eliminating  `_deprecate_as_int()`

cupy/polynomial/polynomial.py
asi1024: The result of the computation in GPU is not guaranteed to be equal to that of CPU. For example, `powf` causes 9 ulp error in the worst case according to the [CUDA documentation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#standard-functions).

cupy/polynomial/polyutils.py
```diff
@@ -0,0 +1,24 @@
...
+def _deprecate_as_int(x, desc):
+    try:
+        return operator.index(x)
+    except TypeError as e:
+        try:
+            ix = int(x)
+        except TypeError:
+            pass
+        else:
+            if ix == x:
```
asi1024, line 14: Please add a test for this case. (e.g., `deg = 3.0`)

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,25 @@
...
+
+    """
+    deg = cupy.polynomial.polyutils._deprecate_as_int(deg, 'deg')
+    if deg < 0:
+        raise ValueError('degree must be non-negative')
+    if x.ndim == 0:
+        x = cupy.expand_dims(x, axis=0)
+    x = cupy.expand_dims(x, axis=-1)
+    x = x + 0.0
+    v = x ** cupy.arange(deg + 1)
```
asi1024, line 24: The dtype of `cupy.arange(deg + 1)` is `int64`, so the dtype of `v` becomes always `float64` or `complex128`.
```suggestion
    dtype = numpy.float64 if x.dtype.kind in 'biu' else x.dtype
    v = x ** cupy.arange(deg + 1, dtype)
```

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,25 @@
...
+    return cupy.asarray(v, x.dtype, order='F')
```
asi1024, line 25: The return value is not always F-contiguous.
```py
>>> numpy.polynomial.polynomial.polyvander(numpy.ones((4, 5, 6)), 4).flags.f_contiguous
False
>>> numpy.polynomial.polynomial.polyvander(numpy.ones((4, 5, 6)), 4).strides
(240, 48, 8, 960)
```

asi1024: How about writing as follows?
```py
out = x ** cupy.arange(deg + 1, dtype=dtype).reshape((-1,) + (1,) * x.ndim)
return cupy.moveaxis(out, 0, -1)
```

cupy/polynomial/polynomial.py
```diff
@@ -0,0 +1,25 @@
...
+        cupy.ndarray: The Vandermonde matrix
+
+    .. seealso:: :func:`numpy.polynomial.polynomial.polyvander`
+
+    """
+    deg = cupy.polynomial.polyutils._deprecate_as_int(deg, 'deg')
+    if deg < 0:
+        raise ValueError('degree must be non-negative')
+    if x.ndim == 0:
+        x = cupy.expand_dims(x, axis=0)
```
asi1024, line 21: For the simplification.
```suggestion
        x = x.ravel()
```

asi1024:
Could you resolve conflicts?

cupy/polynomial/polynomial.py
asi1024: This typecast is not needed now.

asi1024:
Jenkins, test this please.

tests/cupy_tests/polynomial_tests/test_polynomial.py
```diff
@@ -10,6 +10,63 @@
...
+
+    @testing.for_all_dtypes()
+    def test_polyvander_non_integral_float_degree(self, dtype):
+        for xp in (numpy, cupy):
+            a = testing.shaped_random((10,), xp, dtype)
+            with pytest.raises(TypeError):
+                xp.polynomial.polynomial.polyvander(a, 2.6)
+
+    @testing.for_all_dtypes()
+    def test_polyvander_integral_float_degree(self, dtype):
```
asi1024, line 46: Please add `@testing.with_requires('numpy>=1.17')`.

tests/cupy_tests/polynomial_tests/test_polynomial.py
```diff
@@ -10,6 +10,63 @@
...
+
+    @testing.for_all_dtypes()
+    def test_polyvander_negative_degree(self, dtype):
+        for xp in (numpy, cupy):
+            a = testing.shaped_random((10,), xp, dtype)
+            with pytest.raises(ValueError):
+                xp.polynomial.polynomial.polyvander(a, -3)
+
+    @testing.for_all_dtypes()
+    def test_polyvander_non_integral_float_degree(self, dtype):
```
asi1024, line 39: Ditto.

asi1024:
Jenkins, test this please.

asi1024:
LGTM. Thank you for your contribution!

PR #3410, opened 2020-06-08
Support `ignore_error` in kernel optimization

asi1024:
Address https://github.com/cupy/cupy/pull/3244#discussion_r435475481.

### DIFF
--- a/cupyx/optimizing/_optimize.py
+++ b/cupyx/optimizing/_optimize.py
@@ -1,4 +1,5 @@
 import contextlib
+import math
 
 import optuna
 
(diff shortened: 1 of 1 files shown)


emcastillo:
Jenkins, test this please

leofang:
Is it possible to pass in `ignore_error` from `cupyx.optimizing.optimize`?

leofang:
Would be nice if we could do this:
```python
with optimizing.optimize(key=None, ignore_error=(some_exception, )):
    # do something
```
which is not currently possible I think?

asi1024:
It is possible to pass in `ignore_error` only from internal implementation.
In reduction case for example, it is passed from the following place:
https://github.com/cupy/cupy/blob/bedd67790e33812cdd3a4b4a8e0c841248e2cdaf/cupy/core/_reduction.pyx#L397-L404

asi1024:
I think `ignore_error` parameters don't need to be exposed to the users in most situations.

leofang:
Yeah that's [what I figured](https://github.com/cupy/cupy/pull/3244#issuecomment-641479088) 😇 Just wanna confirm with you I understand it correctly 🙏

Issue #3316, opened 2020-05-01
AttributeError: 'NoneType' object has no attribute 'record'


mnicely:
CuPy Version          : 7.3.0
CUDA Root             : /usr/local/cuda
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSOLVER Version      : (10, 3, 0)
cuSPARSE Version      : 10301
NVRTC Version         : (10, 2)
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2406
NCCL Runtime Version  : 2507

If I try to run the following command, I get the following error.
`cp.asarray(z[cp.newaxis, :, :])`

```bash
Traceback (most recent call last):
  File "kalman.py", line 125, in <module>
    z = cp.asarray(z[cp.newaxis, :, :])
  File "/home/belt/anaconda3/envs/cusignal/lib/python3.8/site-packages/cupy/creation/from_data.py", line 68, in asarray
    return core.array(a, dtype, False, order)
  File "cupy/core/core.pyx", line 1785, in cupy.core.core.array
  File "cupy/core/core.pyx", line 1862, in cupy.core.core.array
  File "cupy/core/core.pyx", line 1950, in cupy.core.core._send_object_to_gpu
AttributeError: 'NoneType' object has no attribute 'record'
```

What's interesting is that it only happens after a run a CuPy kernel. One which doesn't even access variable z.

leofang:
What's the shape of `z`?

mnicely:
2 x 1 array
`z = np.array([i, i], dtype=dt).T`

mnicely:
I thought I got around but I'm hitting it again. I'm trying to run a loop and it crashes on the first `cp.asarray` of the second loop (that uses CuPy raw kernels).

See my latest script.
[kalman_test_script.txt](https://github.com/cupy/cupy/files/4611552/kalman_test_script.txt)

takagi:
I'll try to reproduce the error with installing cuSignal.

takagi:
@mnicely I couldn't find `cusignal.KalmanFilter` when I tried to run `kalman_test_script.txt`. Do you know something about it? I installed cuSignal following its install guide with conda.

```Python
$ python kalman_test_script.py
num_points 4096
iterations 1
use_numba True
data type <class 'numpy.float32'>
[...]

PR #3423, opened 2020-06-10
Properly reset current stream in case null stream is destroyed

takagi:
Rel. #3316.

When a CuPy stream object set to the current stream is about to be destroyed, the current stream is automatically replaced with `Stream.null`. However, it does not work when the stream is a user-defined null stream, causing the error reported in #3316.

This PR fixes the problem to properly reset the current stream even if a user-defined null stream is used.

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -276,12 +276,16 @@ class Stream(BaseStream):
         cdef intptr_t current_ptr
         if is_shutting_down():
             return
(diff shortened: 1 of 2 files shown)


leofang:
FYI @mnicely

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

mnicely:
Thanks @takagi !

PR #3424, opened 2020-06-11
Use `find_packages` in `setup.py`

asi1024:


kmaehashi:
pfnCI, test this please.

PR #3419, opened 2020-06-09
Support boolean in `cupyx.scipy.ndimage` stats functions

takagi:
Follows #3259.

This PR fixes `cupyx.scipy.ndimage` stats functions to support `cupy.bool_`.

PR #3425, opened 2020-06-11
Fix `cupyx.scipy.ndimage.sum` taking zero-dimensional input

takagi:
Follows #3259. Merge #3419 first.

This PR fixes the case `cupyx.scipy.ndimage.sum` takes a zero-dimensional input.


kmaehashi:
pfnCI, test this please.

PR #2698, opened 2019-11-24
CUB-based CSR sparse matrix vector multiply

grlee77:
I was curious whether CUB would also improve the performance of sparse matrix multiplication. It seems that for < ~ 1,000,000 non-zero entries it is faster. At sizes larger than that, it is basically tied with cuSPARSE. A bit disappointing that the >3 fold gains at smaller sizes don't hold up for the larger problems of interest, although would be curious if this is also true for higher-end NVIDIA cards. (I tested with a GTX 1080 Ti)

The main limitation is that CUB only has this one sparse matrix function so it is limited to CSR only. Aside from performance, I think the other feature it has over cuSPARSE is that it can be used for additional dtypes such as `int` that are not supported by cuSPARSE. This can be done with the undocumented `device_spmv` function here, but the end-user interface is via the existing `__mul__` method of `csr_matrx` and is thus limited to floating point types.

I will post a benchmark below.

PR #3428, opened 2020-06-11
Test CUB SpMV

leofang:
This tests against the CUB SpMV support added in #2698. Style in line with the existing tests and #2598. Not sure why I didn't add this in #2598...

@grlee77 Could you take a look?


emcastillo:
Jenkins, test this please

emcastillo:
The test actually doesn't ensure that cub is being used for the spmv, but I don't think there is an easy way to do that, so LGTM

leofang:
> The test actually doesn't ensure that cub is being used for the spmv

Thanks, @emcastillo! 

That's right, but I think it's the same for all existing CUB tests...That's why in the beginning of #2598 I suggested to monitor the performance change (under the assumption that CUB is always faster). Not sure whether the backend mechanism suggested in #3445 could help for this purpose?

leofang:
> The test actually doesn't ensure that cub is being used for the spmv

Sorry @emcastillo, I was being silly. Mock tests are exactly for this purpose! I will write mock tests to cover the whole `cupy.cuda.cub` module.

Issue #2365, opened 2019-08-07
possible bug in cupyx sparse matrix transpose/multiplication


lastephey:
Hello cupy(x) developers,

I would like to report what looks like a bug in the cupyx sparse matrix library. I am running on a Skylake cpu and a Tesla V100 gpu on a SUSE-ish Linux system. CuPy version: 6.0.0. CUDA version 10.1.168. I am running inside a conda environment which you can find in my requirements.txt file: 

[requirements.txt](https://github.com/cupy/cupy/files/3474978/requirements.txt)

Here is a reproducer in which I am trying to compare the same operations on the cpu and gpu and checking at every step that they are the same. Finally in the last step the `np.allclose()` check fails. The `iCov_max_diff` value is much larger than machine precision which suggests there is a bug. 

```
import numpy as np
import scipy.sparse
import cupy as cp
import cupyx as cpx

m = n = 1000

np.random.seed(1)

#create random data to use in W
random_data = np.random.random(m)

#create A random sparse matrix
A_cpu = scipy.sparse.random(m, n, format='csr', random_state=42)
A_gpu = cpx.scipy.sparse.csr_matrix(A_cpu)
#yank gpu back and compare
A_yank = A_gpu.get()
assert np.allclose(A_cpu.todense(),A_yank.todense())

#create W random sparse matrix
W_cpu = scipy.sparse.spdiags(data=random_data, diags=[0,], m=m, n=n)
W_gpu = cpx.scipy.sparse.spdiags(data=random_data, diags=[0,], m=m, n=n)
#yank gpu back and compare
W_yank = W_gpu.get()
assert np.allclose(W_cpu.todense(),W_yank.todense())

#see how the dot products go
W_dot_A_cpu = W_cpu.dot(A_cpu)
W_dot_A_gpu = W_gpu.dot(A_gpu)
#yank gpu back and compare
W_dot_A_yank = W_dot_A_gpu.get()
assert np.allclose(W_dot_A_cpu.todense(), W_dot_A_yank.todense())

#check the transpose
A_trans_cpu = A_cpu.T
A_trans_gpu = A_gpu.T
#yank gpu back and compare
A_trans_yank = A_trans_gpu.get() #use get bc its a sparse object
assert np.allclose(A_trans_cpu.todense(), A_trans_yank.todense())

#okay now inverse covariance (where things go wrong)
iCov_cpu = A_cpu.T.dot(W_dot_A_cpu)
[...]

PR #3430, opened 2020-06-11
Speed up CSR SpMV by orders of magnitude

### LINKED ISSUE #2664: Sparse matrix multiply orders of magnitude slower than PyTorch
I have been doing some profiling comparing cupy's sparse matrix `dot()` with PyTorch's `mm()` and I'm getting some very surprising results. I noticed that while cupy is using cusparse, PyTorch appears to be using only thrust/cub to do the multiply. 

Strangely, I'm finding that it's the cusparse `csrgemm_kernel` from cupy that is dominating the runtime of the multiply, taking 40ms. The end-to-end matrix multiply is taking upwards of 70ms for cupy. The PyTorch implementation is taking a little over 2ms to run the multiply end-to-end. 

I'm attaching the [nvvp file](https://github.com/cupy/cupy/files/3870573/cupy_vs_pytorch.nvvp.zip) with my profiling results. I'm pretty surprised. The algorithm being profiled is a simple Multinomial Naive Bayes using the 20-newsgroups dataset from scikit-learn. I trained both algorithms twice in order to eliminate jit and cuda context creation from the comparison.

```
>>> cupy.show_config()
CuPy Version          : 6.5.0
CUDA Build Version    : 10000
CUDA Driver Version   : 10000
CUDA Runtime Version  : 10000
cuDNN Build Version   : 7600
cuDNN Version         : 7600
NCCL Build Version    : 2406
NCCL Runtime Version  : 2406
```

### LINKED ISSUE #2790: slow SpMV
Hello, 
I mention this issue earlier in #2365 as performance issue; the bug was fixed but performance is still an issue. When I perform a sparse matrix vector multiply using CSR format, it is over 200 times slower the first time it is computed and it is always  over 200 times slower when multiplying a sparse matrix in CSC format with a vector from the left hand side. Internally cuda should perform the same operation, and there should be no need to re-sort the entries twice. 

The attached sparse_test.py performs the same computation in different ways, compares with scipy and csc/csr formats, using a sparse matrix with dimension (10k by 10k) 10M non-zero elements giving the following output:

all cpu-gpu close?: True True
all lm/rm close?  : True True
all 1st/2nd close?: True True
time scipy     lm/rm                            :0.00446,0.00461
**wall time cupy lm (1st and 2nd time) :0.0686,  0.00024
wall time cupy rm (1st and 2nd time)         :0.0594,  0.061
cuda time cupy lm (1st and 2nd time):0.0594, 0.000229
cuda time cupy lm (1st and 2nd time):0.0593, 0.061**

The performance issue goes away if modify  cupyx/scipy/sparse/coo.py as attached by setting _has_canonical_format=True after sorting. I am not sure if it has other unintended consequences, but I have not observed any issues in my code. 

[sparse_test.py.txt](https://github.com/cupy/cupy/files/3947424/sparse_test.py.txt)
[coo.py.txt](https://github.com/cupy/cupy/files/3947330/coo.py.txt)

-------------
configuration: ubuntu 18.04, cuda 10.1, rtx-titan gpu

CuPy Version          : 7.0.0
CUDA Root             : /usr/local/cuda-10.1
CUDA Build Version    : 10010
CUDA Driver Version   : 10010
CUDA Runtime Version  : 10010
cuBLAS Version        : 10201
cuFFT Version         : 10101
cuRAND Version        : 10101
cuSPARSE Version      : 10300
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2402
NCCL Runtime Version  : 2402

leofang:
**Main credit goes to @smarkesini (from his #2790).**

Close #2790. Close #2664.

I don't know why this simple but great change was buried unnoticed 😢 Explained by @smarkesini in https://github.com/cupy/cupy/issues/2365#issuecomment-532133094: 
> I suspect this issue has something to do with '_has_canonical_format' not being set even after calling coosort. There are several places where to set this flag, and I am not sure if the following is the best place, anyway I was focusing on csr and modified the last 3 lines in cupyx/scipy/sparse/coo.py 'tocsr' function as described below and got better performance for my problem (SpMV with a csr matrix ), >100x speedup on the first iteration.

**UPDATE:** While performance improvement was the original concern in this PR, note that this is also a compatibility bug -- In SciPy it is guaranteed that a conversion from COO to CSR/CSC will always lead to a canonical format, because the duplicates are summed and the indices are sorted.

~~and in https://github.com/cupy/cupy/issues/2365#issuecomment-532434034:~~
> ~~following the definition in there [coo2csr](https://docs.nvidia.com/cuda/cusparse/index.html#coo2csr), this function converts the array containing the uncompressed row indices (corresponding to COO format) into an array of **compressed row pointers (corresponding to CSR format)**. In the [cuda definition](https://docs.nvidia.com/cuda/cusparse/index.html#csr-format): "Sparse matrices in CSR format are assumed to be stored in row-major CSR format, in other words, the index arrays are first sorted by row indices and then within the same row by column indices. It is assumed that each pair of row and column indices appears only once."~~

As for performance boost, as mentioned in the discussion there the speed-up is multiple orders of magnitude, which I verified. I'll double check my benchmark script and post the result below.

### DIFF
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -896,6 +896,25 @@ def csr2csc(x):
         (data, indices, indptr), shape=x.shape)
 
 
+def csc2csr(x):
+    handle = device.get_cusparse_handle()
+    m, n = x.shape
+    nnz = x.nnz
+    data = cupy.empty(nnz, x.dtype)
+    indptr = cupy.empty(m + 1, 'i')
+    indices = cupy.empty(nnz, 'i')
+
+    _call_cusparse(
+        'csr2csc', x.dtype,
+        handle, n, m, nnz, x.data.data.ptr,
+        x.indptr.data.ptr, x.indices.data.ptr,
+        data.data.ptr, indices.data.ptr, indptr.data.ptr,
+        cusparse.CUSPARSE_ACTION_NUMERIC,
+        cusparse.CUSPARSE_INDEX_BASE_ZERO)
+    return cupyx.scipy.sparse.csr_matrix(
+        (data, indices, indptr), shape=x.shape)
+
+
 def dense2csc(x):
     """Converts a dense matrix in CSC format.
 
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -896,6 +896,30 @@ def csr2csc(x):
         (data, indices, indptr), shape=x.shape)
 
 
+def csc2coo(x, data, indices):
+    """Converts a CSC-matrix to COO format.
+
+    Args:
+        x (cupyx.scipy.sparse.csc_matrix): A matrix to be converted.
+        data (cupy.ndarray): A data array for converted data.
+        indices (cupy.ndarray): An index array for converted data.
+
+    Returns:
+        cupyx.scipy.sparse.coo_matrix: A converted matrix.
+
+    """
+    handle = device.get_cusparse_handle()
+    n = x.shape[1]
+    nnz = len(x.data)
+    col = cupy.empty(nnz, 'i')
+    cusparse.xcsr2coo(
+        handle, x.indptr.data.ptr, nnz, n, col.data.ptr,
+        cusparse.CUSPARSE_INDEX_BASE_ZERO)
+    # data and indices did not need to be copied already
+    return cupyx.scipy.sparse.coo_matrix(
+        (data, (indices, col)), shape=x.shape)
+
+
 def csc2csr(x):
     handle = device.get_cusparse_handle()
     m, n = x.shape
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -827,6 +827,8 @@ def coosort(x, sort_by='r'):
         'gthr', x.dtype,
         handle, nnz, data_orig.data.ptr, x.data.data.ptr,
         P.data.ptr, cusparse.CUSPARSE_INDEX_BASE_ZERO)
+    if sort_by == 'c':  # coo is sorted by row first
+        x._has_canonical_format = False
 
 
 def coo2csr(x):
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -793,11 +793,12 @@ def cscsort(x):
         P.data.ptr, cusparse.CUSPARSE_INDEX_BASE_ZERO)
 
 
-def coosort(x):
+def coosort(x, sort_by='r'):
     """Sorts indices of COO-matrix in place.
 
     Args:
         x (cupyx.scipy.sparse.coo_matrix): A sparse matrix to sort.
+        sort_by (str): Sort the indices by row ('r', default) or column ('c').
 
     """
     nnz = x.nnz
@@ -812,9 +813,16 @@ def coosort(x):
     P = cupy.empty(nnz, 'i')
     data_orig = x.data.copy()
     cusparse.createIdentityPermutation(handle, nnz, P.data.ptr)
-    cusparse.xcoosortByRow(
-        handle, m, n, nnz, x.row.data.ptr, x.col.data.ptr,
-        P.data.ptr, buf.data.ptr)
+    if sort_by == 'r':
+        cusparse.xcoosortByRow(
+            handle, m, n, nnz, x.row.data.ptr, x.col.data.ptr,
+            P.data.ptr, buf.data.ptr)
+    elif sort_by == 'c':
+        cusparse.xcoosortByColumn(
+            handle, m, n, nnz, x.row.data.ptr, x.col.data.ptr,
+            P.data.ptr, buf.data.ptr)
+    else:
+        raise ValueError("sort_by must be either 'r' or 'c'")
     _call_cusparse(
         'gthr', x.dtype,
         handle, nnz, data_orig.data.ptr, x.data.data.ptr,
@@ -830,6 +838,17 @@ def coo2csr(x):
         indptr.data.ptr, cusparse.CUSPARSE_INDEX_BASE_ZERO)
     return cupyx.scipy.sparse.csr.csr_matrix(
         (x.data, x.col, indptr), shape=x.shape)
+
+
+def coo2csc(x):
+    handle = device.get_cusparse_handle()
+    n = x.shape[1]
+    indptr = cupy.empty(n + 1, 'i')
+    cusparse.xcoo2csr(
+        handle, x.col.data.ptr, x.nnz, n,
+        indptr.data.ptr, cusparse.CUSPARSE_INDEX_BASE_ZERO)
+    return cupyx.scipy.sparse.csc.csc_matrix(
+        (x.data, x.row, indptr), shape=x.shape)
 
 
 def csr2coo(x, data, indices):
--- a/cupyx/scipy/sparse/coo.py
+++ b/cupyx/scipy/sparse/coo.py
@@ -215,7 +215,7 @@ class coo_matrix(sparse_data._data_matrix):
         if self.data.size == 0:
             self._has_canonical_format = True
             return
-        keys = cupy.stack([self.row, self.col])
+        keys = cupy.stack([self.col, self.row])
         order = cupy.lexsort(keys)
         src_data = self.data[order]
         src_row = self.row[order]
--- a/cupyx/scipy/sparse/coo.py
+++ b/cupyx/scipy/sparse/coo.py
@@ -322,11 +322,8 @@ class coo_matrix(sparse_data._data_matrix):
         """
         if self.nnz == 0:
             return csc.csc_matrix(self.shape, dtype=self.dtype)
-        # copy is ignored because both sum_duplicates and coosort change
-        # the underlying data
-        if copy:
-            import warnings
-            warnings.warn("copy=True is not supported")
+        # copy is silently ignored (in line with SciPy) because both
+        # sum_duplicates and coosort change the underlying data
         self.sum_duplicates()
         x = self.copy()
         cusparse.coosort(x, 'c')
@@ -348,11 +345,8 @@ class coo_matrix(sparse_data._data_matrix):
         """
         if self.nnz == 0:
             return csr.csr_matrix(self.shape, dtype=self.dtype)
-        # copy is ignored because both sum_duplicates and coosort change
-        # the underlying data
-        if copy:
-            import warnings
-            warnings.warn("copy=True is not supported")
+        # copy is silently ignored (in line with SciPy) because both
+        # sum_duplicates and coosort change the underlying data
         self.sum_duplicates()
         x = self.copy()
         cusparse.coosort(x, 'r')
--- a/cupyx/scipy/sparse/coo.py
+++ b/cupyx/scipy/sparse/coo.py
@@ -8,6 +8,7 @@ except ImportError:
 import cupy
 from cupy import cusparse
 from cupyx.scipy.sparse import base
+from cupyx.scipy.sparse import csc
 from cupyx.scipy.sparse import csr
 from cupyx.scipy.sparse import data as sparse_data
 from cupyx.scipy.sparse import util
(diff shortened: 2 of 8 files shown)


leofang:
TODO: compare with CUB SpMV.

leofang:
Benchmarking: (look at the last line in the output)
<details><summary>script: (click me)</summary>

```python
import numpy as np
import cupy as cp
import cupyx
from cupyx.time import repeat
import scipy.sparse


# switch CUB on and off here
cp.cuda.cub_enabled = True

# matrix dimensions
M = np.int64(1e6)
N = M
nnz = M * 100


Krow=np.random.randint(0, high=M, size=nnz//2, dtype='l')
Kcol=np.random.randint(0, high=M, size=nnz//2, dtype='l')
# make duplicates
Krow=np.concatenate((Krow, Krow))
Kcol=np.concatenate((Kcol, Kcol))

Kval=np.random.rand(nnz).astype('float32')

# use scipy
S=scipy.sparse.csr_matrix((Kval.ravel(),    (Krow.ravel(), Kcol.ravel())), shape=(M,N))
ST=scipy.sparse.csc_matrix((Kval.ravel(),   (Kcol.ravel(), Krow.ravel())), shape=(N,M))

# vector to multiply
x=np.random.rand(M).astype('float32')

#print("left multiply (CSR)")
Sx = S * x
time_scipy = repeat(S.__mul__, (x,), n_repeat=100).cpu_times.mean()

#print("right multiply (CSC)")
xS = x * ST
time_scipyrm = repeat(x.__mul__, (ST,), n_repeat=100).cpu_times.mean()

print("allclose CPU:", np.allclose(Sx, xS)) #, "time:", time_scipy, time_scipyrm)

# use cupy
Krow=cp.asarray(Krow)
Kcol=cp.asarray(Kcol)
Kval=cp.asarray(Kval)
x   =cp.asarray(x)

S=cupyx.scipy.sparse.coo_matrix((Kval.ravel(),(Krow.ravel(), Kcol.ravel())), shape=(M,N))
S=cupyx.scipy.sparse.csr_matrix(S)
ST=cupyx.scipy.sparse.coo_matrix((Kval.ravel(), (Kcol.ravel(), Krow.ravel())), shape=(N,M))
ST=cupyx.scipy.sparse.csc_matrix(ST)

#print("left multiply by csr")
cSx = S * x
time_cupy1 = repeat(S.__mul__, (x,), n_repeat=100).gpu_times.mean()

#print("right multiply by csc")
cxS = x * ST
time_cupy1rm = repeat(x.__mul__, (ST,), n_repeat=100).gpu_times.mean()

xS=cp.asarray(xS)
Sx=cp.asarray(Sx)

print("allclose GPU:", cp.allclose(cSx, cxS))
print("allclose GPU-CPU: (rm) ", cp.allclose(cSx, xS),'(lm)', cp.allclose(cSx, Sx))


print("time scipy: lm:", time_scipy, "rm:", time_scipyrm)
print("time cupy: lm:", time_cupy1, "rm:", time_cupy1rm)
```

</p></details>


Before this PR:
```bash
time cupy: lm: 0.0012003951942920685 rm: 1.6158148022460939    # CUB is on
time cupy: lm: 0.0012251673579216002 rm: 1.8342621813964843    # CUB is off
```
With this PR (**>1300x speedup!**):
```bash
time cupy: lm: 0.0011969430339336395 rm: 0.0012273990404605866    # CUB is on
time cupy: lm: 0.0012167888033390046 rm: 0.0012395257604122165    # CUB is off
```

@grlee77 You might be interested.

leofang:
See the benchmark with this PR added against CUB here: https://github.com/cupy/cupy/pull/2698#issuecomment-642813596.

cupyx/scipy/sparse/coo.py
```diff
@@ -215,7 +216,7 @@ def sum_duplicates(self):
         if self.data.size == 0:
             self._has_canonical_format = True
             return
-        keys = cupy.stack([self.col, self.row])
+        keys = cupy.stack([self.row, self.col])
```
grlee77, line 218: Are you sure about this change? The description you quoted states:
"the index arrays are first sorted by row indices and then within the same row by column indices"

Since lexsort uses the last row as the primary sort key, didn't the already existing code match that description?

In either case, it seems it can be verified empirically against SciPy. If all test cases pass either way it seems that would indicate a hole in our test coverage!

grlee77:
Wow, not sure how this got overlooked. That is an amazing improvement for very little change!

cupyx/scipy/sparse/coo.py
leofang: I need your help, Gregory. I've been looking into this for a few hours, and I am super confused now. 

It seems the change at this line neither affects speed nor correctness, at least with my test script. (And FYI, the whole test suite `tests/cupyx_tests/scipy_tests/sparse_tests/` passed either way!) The only thing that matters (for speed) is to set `_has_canonical_format=True`, which makes sense because when calling `sum_duplicates()` both CSC & CSR converts to COO (that's why changes in `coo.py` affects CSR), and this flag would provide a fast path. Does it mean `keys`, `lexsort`, etc here are simply dead code? I am lost...

grlee77: I haven't looked at it closely, but was just wandering what motivated the change. Ca n try to take a look tomorrow

emcastillo:
I don't know why I didn't look this in depth when the issue was created.
Sorry about it @smarkesini and thanks @leofang for bringing this back!

cupyx/scipy/sparse/coo.py
leofang: I now incline that this was simply an overlook of previous PR and that we don't need to change this line. After `coo.sum_duplicates()` and `cusparse.coosort()`, `_has_canonical_format` should always be `True`. The only concern is that `cusparse.coosort()` sorts rows by default, and for `tocsc()` it doesn't sound right, so I will address both issues.

smarkesini: The csc format is stored in column-major format, whereas csr is in row-major, 
so I think a CSC matrix is stored internally as a transpose of a CSR matrix.

cupy/cusparse.py
```diff
@@ -832,6 +840,17 @@ def coo2csr(x):
         (x.data, x.col, indptr), shape=x.shape)
 
 
+def coo2csc(x):
+    handle = device.get_cusparse_handle()
+    n = x.shape[1]
+    indptr = cupy.empty(n + 1, 'i')
+    cusparse.xcoo2csr(
+        handle, x.col.data.ptr, x.nnz, n,
+        indptr.data.ptr, cusparse.CUSPARSE_INDEX_BASE_ZERO)
```
leofang, line 849: This works for coo2csc according to the documentation of `cusparse<t>coo2csr()`:
> It can also be used to convert the array containing the uncompressed column indices (corresponding to COO format) into an array of column pointers (corresponding to CSC format).

leofang:
I made the conversion from COO to CSC/CSR symmetric. This requires `coosort` to sort by row for CSR and by column for CSC, respectively.

cupyx/scipy/sparse/coo.py
leofang: CSR/CSC matrix internally stores 3 arrays: `data` of size `nnz`, `indices` of size `nnz`, and `indptr` of size `m+1`/`n+1`. This is the same for both cuSPASE and SciPy. A transpose operation would need to update indices` and `indptr`.

cupyx/scipy/sparse/csr.py
```diff
@@ -295,8 +295,10 @@ def transpose(self, axes=None, copy=False):
                 'swapping dimensions is the only logical permutation.')
 
         shape = self.shape[1], self.shape[0]
-        return csc.csc_matrix(
+        trans = csc.csc_matrix(
             (self.data, self.indices, self.indptr), shape=shape, copy=copy)
+        trans._has_canonical_format = self._has_canonical_format
+        return trans
```
leofang, line 301: I am still concerned in this transpose operation. If [I understand it correctly](https://github.com/cupy/cupy/pull/3430#discussion_r439567412), `self.indices` and `self.indptr` should be updated during transpose. Am I missing something?

leofang:
Hi @emcastillo @grlee77 @smarkesini @cjnolet This is ready for review. The performance improvement is same as [reported earlier](https://github.com/cupy/cupy/pull/3430#issuecomment-642805414). I left [one question](https://github.com/cupy/cupy/pull/3430#discussion_r439569215), though, which doesn't seem to be relevant to this PR (not sure why).

cupyx/scipy/sparse/csr.py
smarkesini: I think for the transpose operation as used in e.g. csrmv in cupy/cusparse.py ...cusparse.spMV(), you  compute  the multiplication directly with the transposed matrix using a _transpose_flag - without actually computing the transpose.

leofang: It seems SciPy does this too. I don't understand...

leofang: We left comments simultaneously, @smarkesini! It's good to know `_transpose_flag` -- I wasn't aware of that. But the `transpose` method here is for high-level operations, say, `b=a.T`.

cupyx/scipy/sparse/coo.py
leofang: Revisiting this line, interestingly I found SciPy actually does what @smarkesini did (sort by col first):
https://github.com/scipy/scipy/blob/adc4f4f7bab120ccfab9383aba272954a0a12fb0/scipy/sparse/coo.py#L543. I wonder why no test suite has caught this deviation so far (and how to write one)...

cupyx/scipy/sparse/csr.py
smarkesini: Generally the sparse matrix operations follow the [sparse BLAS](https://math.nist.gov/spblas/) conventions, with transpose operations computed internally, so to get b=a.T you can just update the flag without sorting everything again; then as long as any operation applied to b is aware of the flag everything works.

leofang: I dug deeper into SciPy, and it turns out that for CSR:
1. `tocsc` will ensure the indices to be sorted, but the outcome is not necessary in the canonical form (i.e., can have repeated indices)
2. `transpose` will ensure nothing.

This observation also holds vice versa for CSC. So, the change here in `csr.py` is not necessary, and `csc.py` will need to be fixed.

leofang: > so to get b=a.T you can just update the flag without sorting everything again; then as long as any operation applied to b is aware of the flag everything works

Right, that's what I expected, but I dunno where this flag is...😅

leofang: I guess `shape` might be used for this purpose...?

cupy/cusparse.py
```diff
@@ -875,6 +896,25 @@ def csr2csc(x):
...
+    indptr = cupy.empty(m + 1, 'i')
+    indices = cupy.empty(nnz, 'i')
+
+    _call_cusparse(
+        'csr2csc', x.dtype,
+        handle, n, m, nnz, x.data.data.ptr,
+        x.indptr.data.ptr, x.indices.data.ptr,
+        data.data.ptr, indices.data.ptr, indptr.data.ptr,
+        cusparse.CUSPARSE_ACTION_NUMERIC,
+        cusparse.CUSPARSE_INDEX_BASE_ZERO)
```
leofang, line 913: ...and this works for csc2csr according to the documentation of `cusparse<t>csr2csc()`:
> Notice that this routine can also be used to convert a matrix in CSC format into a matrix in CSR format.

leofang:
I think this can be treated as a compatibility bug and get tested. Will add tests later.

leofang:
Done.

emcastillo:
Jenkins, test this please

leofang:
Thanks all! 🙏

cupyx/scipy/sparse/coo.py
```diff
@@ -319,7 +320,16 @@ def tocsc(self, copy=False):
             cupyx.scipy.sparse.csc_matrix: Converted matrix.
 
         """
-        return self.T.tocsr().T
```
leofang, line 322: Leaving a note to my future self: This change is in line with SciPy that the conversion functions should be made symmetric (in this case `tocsr` vs `tocsc`) whenever a low-level library call is available to carry out the task. In this case, this change avoids two (unnecessary) constructor calls from `.T`.

PR #3432, opened 2020-06-12
Add `cupy.shares_memory`

asi1024:


asi1024:
Maybe we can use `searchsorted` for the performance.

takagi:
Would you add it to the reference `misc.rst`?

takagi:
pfnCI, test this please.

asi1024:
pfnCI, test this please.

takagi:
Interesting solution, LGTM!

leofang:
> Interesting solution

This is a very interesting approach indeed. Out of curiosity I dug into NumPy's implementation, and found it is a non-trivial problem: https://github.com/numpy/numpy/blob/master/numpy/core/src/common/mem_overlap.c

asi1024:
My solution is simpler, but NumPy's solution seems to work with better memory/time complexity.

PR #3416, opened 2020-06-09
Fix sphinx version for travis

emcastillo:
Travis started installing sphinx 3.1.0,
set it to 3.0.4 while we look for the issue 

https://travis-ci.org/github/cupy/cupy/builds/696302151

PR #3435, opened 2020-06-14
Only avoid Sphinx 3.1.0

leofang:
Follow up of #3416. 3.1.1 is just released and they claimed the bug we reported is resolved in 3.1.1 (sphinx-doc/sphinx#7809). Let's see if Travis complains.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ requirements = {
         'theano',
     ],
     'docs': [
-        'sphinx==3.0.4',
+        'sphinx!=3.1.0',  # due to a 3.1.0 bug, see sphinx-doc/sphinx#7809
         'sphinx_rtd_theme',
     ],
     'travis': [


leofang:
OK, Travis uses 3.1.1 and is happy with it!

emcastillo:
Thanks for all the effort and for reporting this to the Sphinx community plus working with them to solve this!.

Regarding the PR itself, I am wondering if we should use the same versions for the Jenkins cupy-docs test (currently 2.0.1) and travis. 
Ideally, we should fix the code in cupy for sphinx 3.1.1 to work in Jenkins, but I don't remember what is the issue. @takagi @kmaehashi what are your thoughts regarding this?

leofang:
Haha no worries. All I did was to report to upstream. Takeshi from Sphinx did all the heavy lifting. It seems there's a class of autodoc bugs, which could be resolved in one PR.

Yeah I dunno why Jenkins needs `sphinx==2.0.1`. Would be nice to sort this out.

leofang:
Jenkins, test this please

kmaehashi:
I've refactored doc-related requirements in #4630.

leofang:
Nice, thanks to follow up!

PR #3438, opened 2020-06-15
Deprecate `testing.empty`

asi1024:
The behavior of `cupy.testing.empty` is so different from `xp.empty` that I feel confusing. So I want to deprecate this feature and encourage to use `xp.zeros((0,))` instead.


kmaehashi:
pfnCI, test this please.

kmaehashi:
Agreed to remove as it is undocumented: `cupy.testing.empty`

PR #3442, opened 2020-06-16
Implement `flatiter.copy()`

takagi:
Follows #3165.

This PR implements `flatiter.copy()` that returns a copy of the iterator as a 1-D array.


emcastillo:
Jenkins, test this please

PR #3443, opened 2020-06-16
Refactor CUB-backed `_SimpleReductionKernel`

asi1024:
Follows #3244.
Moved the implementations to `_cub_reduction.pyx`

### DIFF
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -470,3 +470,45 @@ cdef inline _cub_two_pass_launch(
 
     # Launch the kernel
     func.linear_launch(gridx, inout_args, 0, blockx, stream)
+
+
+cdef _launch_cub(
+        self, out_block_num, block_size, block_stride,
+        in_args, out_args, in_shape, out_shape, type_map,
+        map_expr, reduce_expr, post_map_expr, reduce_type,
+        stream, params, cub_params):
+    cdef bint full_reduction, use_cub = cub_params[0]
+    cdef Py_ssize_t contiguous_size, items_per_thread
+    cdef function.Function func
+
+    # Kernel arguments passed to the __global__ function.
+    assert use_cub
+    items_per_thread = cub_params[1]
+    contiguous_size = cub_params[2]
+    full_reduction = cub_params[3]
+
+    if full_reduction:
+        _cub_two_pass_launch(
+            self.name, block_size, contiguous_size, items_per_thread,
+            reduce_type, params, in_args, out_args, self.identity,
+            map_expr, reduce_expr, post_map_expr,
+            type_map, self._input_expr, self._output_expr,
+            self._preamble, (), stream)
+        return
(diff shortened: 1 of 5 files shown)


leofang:
I will take a closer look (and also test this locally) later this week, but 👍 and many thanks for the nice refactoring, @asi1024! I was only able to do your first step (7781e0d) locally, and then I was stuck and had no idea how to proceed 😂

From what I skimmed over, it seems the secret is to make a separate cache `_SimpleCubReductionKernel_get_cached_function`, but before reaching to that point the call path is still going through `_SimpleReductionKernel`, right?

asi1024:
`_SimpleCubReductionKernel_get_cached_function` is called from `_SimpleReductionKernel._call` via `_cub_reduction._try_to_call_cub_reduction`, but I don't feel it is confusing.

leofang:
Thanks, @asi1024! No, I don't think it's confusing, just wanna confirm my understanding.

emcastillo:
The PR is fine.

However, there is a concern that we just discussed.
When cub is enabled, and cub reduction can't be done, all these checks seem to add non-negligible overhead when falling back to cupy reduction. We should measure the overhead and find a way to reduce it if it is significant.

However, this does not affect builds without cub as the checks will not be done.

cupy/core/_cub_reduction.pyx
```diff
@@ -0,0 +1,648 @@
...
+        if in_arr.dtype.kind == 'c':
+            return None
+        # cannot cast float16 to complex
+        if in_arr.dtype.char == 'e' and out_arr.dtype.kind == 'c':
+            return None
+
+    # check reduction axes, if not contiguous then fall back to old kernel
+    reduce_axis = tuple(sorted(reduce_axis))
+    out_axis = tuple(sorted(out_axis))
+    if in_arr.flags.f_contiguous:
```
emcastillo, line 293: This check can be performed earlier to improve cpu time if we have to fall back to cupy reduction.

also use `arr._f_contiguous`, `arr._c_contiguous` since accessing `flags` interacts with the python runtime.

asi1024: Fixed. PTAL.

emcastillo: Thanks!

emcastillo:
Jenkins, test this please

asi1024:
Jenkins, test this please

leofang:
> When cub is enabled, and cub reduction can't be done, all these checks seem to add non-negligible overhead when falling back to cupy reduction. We should measure the overhead and find a way to reduce it if it is significant.

According my benchmark in #3244 the overhead seems to be low, but yes it's still an overhead if falling back to CuPy.

By the way, I am a bit worried that the CI is not yet set up to test this (right?) I trust both of you that #3244 and this PR are tested thoroughly in a local env 😇 I'm stretched super thin this week so I'll have to revisit this later than I promised...Sorry.

PR #3444, opened 2020-06-17
Detect and show Thrust build version

leofang:
Companion PR of 03f4964 from #2584. Starting CUB & Thrust 1.9.9 they are version-locked with each other (in theory, explained below), so this could help detect/debug if there's any incompatibility issue.

* Thrust and CUB have a circular dependency. However, if we use CUB alone, there's a good chance it's decoupled from Thrust, unless iterators are used (see https://github.com/NVlabs/cub/issues/182#issuecomment-631272979). I tested with CuPy that the latest CUB 1.9.10 works with an old Thrust from CUDA 10.0, so I don't think there's any concern here, and this is just for richer config output.

### DIFF
--- a/cupy/cuda/cupy_thrust.h
+++ b/cupy/cuda/cupy_thrust.h
@@ -2,6 +2,7 @@
 #define INCLUDE_GUARD_CUPY_CUDA_THRUST_H
 
 #ifndef CUPY_NO_CUDA
+#include <thrust/version.h>  // for THRUST_VERSION
 
 namespace cupy {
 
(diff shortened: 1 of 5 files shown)


takagi:
pfnCI, test this please.

takagi:
Would you rebase this on the current master as we need #3469 to pass CI to follow the latest SciPy release?

kmaehashi:
pfnCI, test this please.

takagi:
LGTM!

leofang:
Thanks, @takagi @kmaehashi!

PR #3448, opened 2020-06-17
Index `CArray` using 32 bit indexes

emcastillo:
This uses 32 bit computations for the indexes when the number of bytes the array occupies is less than 4GB.

I benchmarked this in v7 with convolve since it was the only routing I was aware of being heavily affected by the indexing. In v8 this was solved.

Time without this change was 6349 us for gpu time vs 4091 after the change.

In other benchmarks now in v8, for non-contiguous reduction (`sum`)

Master:
```
- case head-float64           :    CPU:   23.730 us   +/- 1.094 (min:   22.743 / max:   26.625) us     GPU:  380.163 us   +/- 2.300 (min:  376.512 / max:  384.928) us
- case head-complex128    :    CPU:   24.185 us   +/- 1.336 (min:   22.938 / max:   26.518) us     GPU:  548.589 us   +/- 2.547 (min:  545.792 / max:  554.528) us
- case middle-float64         :    CPU:   26.980 us   +/-64.709 (min:   22.439 / max:  240.793) us     GPU:37630.074 us   +/-60.518 (min:37598.751 / max:37809.696) us
- case middle-complex128 :    CPU:   24.330 us   +/- 2.191 (min:   22.782 / max:   30.316) us     GPU:38163.952 us   +/-12.996 (min:38146.591 / max:38182.495) us
- case tail-float64               :    CPU:   25.171 us   +/- 2.381 (min:   22.843 / max:   31.151) us     GPU: 2047.776 us   +/- 4.583 (min: 2038.272 / max: 2057.120) us
- case tail-complex128       :    CPU:   25.782 us   +/-87.280 (min:   23.851 / max:  317.408) us     GPU: 2274.586 us   +/-87.979 (min: 2237.792 / max: 2537.888) us
- case out-float64               :    CPU:   25.216 us   +/- 2.509 (min:   23.527 / max:   32.315) us     GPU:18356.816 us   +/-14.704 (min:18334.047 / max:18389.696) us
- case out-complex128       :    CPU:   26.449 us   +/- 5.183 (min:   23.546 / max:   41.604) us     GPU:19880.867 us   +/- 8.269 (min:19868.544 / max:19892.544) us
```

This PR
```
- case head-float64           : CPU:   23.795 us   +/- 1.384 (min:   22.646 / max:   27.534) us     GPU:  353.030 us   +/- 1.890 (min:  350.464 / max:  357.440) us
- case head-complex128   : CPU:   24.769 us   +/- 1.989 (min:   23.436 / max:   30.367) us     GPU:  528.640 us   +/- 2.268 (min:  526.528 / max:  534.432) us
- case middle-float64        : CPU:   24.757 us   +/- 2.055 (min:   22.894 / max:   29.420) us     GPU:37740.697 us   +/-15.101 (min:37716.606 / max:37760.834) us
- case middle-complex128: CPU:   25.171 us   +/- 2.768 (min:   23.408 / max:   33.205) us     GPU:38205.062 us   +/-19.121 (min:38172.031 / max:38233.887) us
- case tail-float64 :    CPU:   25.404 us   +/- 2.603 (min:   23.235 / max:   31.579) us     GPU: 1974.010 us   +/- 4.241 (min: 1969.216 / max: 1984.160) us
- case tail-complex128 :    CPU:   24.865 us   +/- 2.442 (min:   22.909 / max:   31.644) us     GPU: 2198.266 us   +/- 4.752 (min: 2192.064 / max: 2208.480) us
- case out-float64 :    CPU:   24.971 us   +/- 1.281 (min:   23.972 / max:   28.579) us     GPU:17490.906 us   +/-23.365 (min:17450.527 / max:17530.272) us
- case out-complex128  :    CPU:   24.976 us   +/- 1.465 (min:   23.482 / max:   28.748) us     GPU:19376.500 us   +/-64.284 (min:19290.592 / max:19511.072) us
```

There are improvements close to 10% in this benchmark. For kernels doing heavy indexing the benefits should be much more noticeable

### DIFF
--- a/cupy/core/core.pxd
+++ b/cupy/core/core.pxd
@@ -15,6 +15,8 @@ cdef class ndarray:
         public strides_t _strides
         readonly bint _c_contiguous
         readonly bint _f_contiguous
+        # To do fast indexing in the CArray class
+        readonly bint _index_32_bits
         readonly object dtype
         readonly memory.MemoryPointer data
         # TODO(niboshi): Return arbitrary owner object as `base` if the
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -135,9 +135,11 @@ cdef class ndarray:
         # data
         if memptr is None:
             self.data = memory.alloc(self.size * itemsize)
+            self._index_32_bits = (self.size * itemsize) <= (1 << 31)
         else:
             self.data = memptr
-        self._index_32_bits = (self.size * itemsize) <= (1 << 31)
+            bound = cupy.core._memory_range.get_bound(self)
+            self._index_32_bits = bound[1] - bound[0] <= (1 << 31)
 
     cdef _init_fast(self, const shape_t& shape, dtype, bint c_order):
         """ For internal ndarray creation. """
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1686,7 +1686,7 @@ cdef inline _carray.CArray _CArray_from_ndarray(ndarray arr):
     # Note that this function cannot be defined in _carray.pxd because that
     # would cause cyclic cimport dependencies.
     cdef _carray.CArray carr = _carray.CArray.__new__(_carray.CArray)
-    carr.init(<void*>arr.data.ptr, arr.size, arr._shape, arr._strides)
+    carr.init(<void*>arr.data.ptr, arr.size, arr.nbytes, arr._shape, arr._strides)
     return carr
 
 
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -137,6 +137,7 @@ cdef class ndarray:
             self.data = memory.alloc(self.size * itemsize)
         else:
             self.data = memptr
+        self._index_32_bits = (self.size * itemsize) < (1 << 32)
 
     cdef _init_fast(self, const shape_t& shape, dtype, bint c_order):
         """ For internal ndarray creation. """
@@ -144,6 +145,7 @@ cdef class ndarray:
         self.dtype, itemsize = _dtype.get_dtype_with_itemsize(dtype)
         self._set_contiguous_strides(itemsize, c_order)
         self.data = memory.alloc(self.size * itemsize)
+        self._index_32_bits = (self.size * itemsize) < (1 << 32)
 
     @property
     def __cuda_array_interface__(self):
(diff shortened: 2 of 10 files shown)


emcastillo:
The code in the reduction routines may exceed the L2 cache size so we should templatize it.
(@asi1024 realized this)

cupy/core/core.pxd
```diff
@@ -15,6 +15,8 @@ cdef class ndarray:
         public strides_t _strides
         readonly bint _c_contiguous
         readonly bint _f_contiguous
+        # To do fast indexing in the CArray class
+        readonly bint _index_32_bits
```
emcastillo, line 19: This is to avoid accessing python runtime when creating the ArgInfo's for the kernel, otherwise this greatly affects CPU time on kernel launch.

cupy/core/include/cupy/carray.cuh
```diff
@@ -215,27 +216,42 @@ public:
...
+    // kernels
     const char* ptr = reinterpret_cast<const char*>(data_);
-    for (int dim = ndim; --dim > 0; ) {
-      ptr += static_cast<ptrdiff_t>(strides_[dim]) * (i % shape_[dim]);
-      i /= shape_[dim];
-    }
-    if (ndim > 0) {
-      ptr += static_cast<ptrdiff_t>(strides_[0]) * i;
+    if (use_32bit_indexing) {
+        int i = static_cast<int>(idx);
```
asi1024, line 230: Please use `unsigned int`.

asi1024: Ah, `strides_[dim]` may be a negaitive value. `int` is correct.

cupy/core/core.pyx
```diff
@@ -137,13 +137,15 @@ cdef class ndarray:
             self.data = memory.alloc(self.size * itemsize)
         else:
             self.data = memptr
+        self._index_32_bits = (self.size * itemsize) <= (1 << 31)
```
asi1024, line 140: This condition is wrong in non-contiguous case. How is the following one?
```py
if self.data.mem.size <= (1 << 31):
    self._index_32_bits = True  # Fast path
else:
    bound = _memory_range._get_bound(array)
    self._index_32_bits = bound.second - bound.first < (1 << 31)
```

emcastillo: We just discussed that this case only applies to views, but if a view associated array is  > 4GB the view will not be indexable with 32 bits indexing, (this will make it slower only).
So the current implementation is correct.

emcastillo: This fix is needed when we provide the array with an external pointer and custom strides, there is the possibility of the strides to be bigger than 4GB

emcastillo:
PTAL!

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #3449, opened 2020-06-17
Fix `cupy.power(0j, 0j)`

asi1024:
```py
>>> import numpy, cupy
>>> numpy.power(0j, 0j)
(1+0j)
>>> cupy.power(0j, 0j)
array(nan+nanj)
```

### DIFF
--- a/cupy/core/_routines_math.pyx
+++ b/cupy/core/_routines_math.pyx
@@ -781,6 +781,11 @@ inline __device__ T integral_power(T in0, T in1) {
     }
     return out0;
 }
+
+template <typename T>
+inline __device__ T complex_power(T in0, T in1) {
+    return in1 == T(0) ? T(1): pow(in0, in1);
+}
 '''
 
 _power = create_ufunc(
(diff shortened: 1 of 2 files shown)


leofang:
Do you think it's a defect in the Thrust complex implementation?

asi1024:
I don't think it's a bug in the Thurst complex implementation. This fix is for the compatibility with NumPy's behavior.

emcastillo:
Thrust implements `pow(a,b)` as `exp(ln(a)*b)` and log 0 is not defined so that is where nan comes.

asi1024:
Yes. https://github.com/cupy/cupy/blob/e4a1fd49ae3199f2ef3c54c71c32cbbe83168129/cupy/core/include/cupy/complex/cpow.h#L7-L11

emcastillo:
Jenkins, test this please

asi1024:
@emcastillo Fixed for CI failure. PTAL.

emcastillo:
Jenkins, test this please

leofang:
@asi1024 @emcastillo I think we could also fix the `pow()` implementation. On GCC 7.5.0 this works as expected:
```c++
#include <stdio.h>
#include <complex>


using std::complex;

int main() {
    complex<double> a (0., 0.);
    complex<double> b = pow(a, 0);
    printf("%f, %f\n", b.real(), b.imag());
    return 0;
}
```
Compile with `-std=c++11` I get:
```bash
1.000000, 0.000000
```
According to C++ standard, `the value
returned for pow(0,0) is implementation-defined.` So we don't have to follow that of Thrust. Just my two cents 😉

leofang:
Sorry, NVM my comment...

leofang:
> Thrust implements `pow(a,b)` as `exp(ln(a)*b)` 

Fun fact: Thrust does so because C++ standard requires it (§ 26.4.8, N3690): 
> The complex power of base x raised to the y-th power, defined as exp(y*log(x)).

PR #3453, opened 2020-06-18
Remove wrong tests in `test_raw.py`

asi1024:
These tests are not guaranteed to work when the compute capability of Device(0) exceeds that of Device(1).

### DIFF
--- a/tests/cupy_tests/core_tests/test_raw.py
+++ b/tests/cupy_tests/core_tests/test_raw.py
@@ -746,6 +746,46 @@ class TestRaw(unittest.TestCase):
             assert cupy.allclose(y, x1 + x2)
 
     @testing.multi_gpu(2)
+    def test_context_switch_RawModule3(self):
+        # run test_load_cubin() on another device
+        # generate cubin in the temp dir and load it on device 0
+
+        device0 = cupy.cuda.Device(0)
+        device1 = cupy.cuda.Device(1)
+        if device0.compute_capability != device1.compute_capability:
+            raise pytest.skip()
+
+        with device0:
+            file_path = self._generate_file('cubin')
+            mod = cupy.RawModule(path=file_path, backend=self.backend)
+
+        # in this test, reloading happens at get_function()
+        with device1:
+            ker = mod.get_function('test_div')
+            x1, x2, y = self._helper(ker, cupy.float32)
+            assert cupy.allclose(y, x1 / (x2 + 1.0))
+
+    @testing.multi_gpu(2)
+    def test_context_switch_RawModule4(self):
+        # run test_load_cubin() on another device
+        # generate cubin in the temp dir and load it on device 0
+
+        device0 = cupy.cuda.Device(0)
+        device1 = cupy.cuda.Device(1)
+        if device0.compute_capability != device1.compute_capability:
+            raise pytest.skip()
+
+        with device0:
+            file_path = self._generate_file('cubin')
+            mod = cupy.RawModule(path=file_path, backend=self.backend)
+            ker = mod.get_function('test_div')
+
+        # in this test, reloading happens at kernel launch
+        with device1:
+            x1, x2, y = self._helper(ker, cupy.float32)
+            assert cupy.allclose(y, x1 / (x2 + 1.0))
+
+    @testing.multi_gpu(2)
     def test_context_switch_RawModule5(self):
         # run test_template_specialization() on another device
         # in this test, re-compiling happens at get_function()
--- a/tests/cupy_tests/core_tests/test_raw.py
+++ b/tests/cupy_tests/core_tests/test_raw.py
@@ -746,32 +746,6 @@ class TestRaw(unittest.TestCase):
             assert cupy.allclose(y, x1 + x2)
 
     @testing.multi_gpu(2)
-    def test_context_switch_RawModule3(self):
-        # run test_load_cubin() on another device
-        # generate cubin in the temp dir and load it on device 0
-        file_path = self._generate_file('cubin')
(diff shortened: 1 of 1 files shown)


leofang:
Do you mean they'd fail unconditionally, or they'd always pass and not catch any problem? They were added to address context switch and I confirmed the tests would fail in a homogeneous environment. How about just adding a condition testing if dev0's CC is higher than that of dev1 (and skip if `True`) instead of removing them completely?

emcastillo:
They fail when devices of different compute capabilities are used simultaneously but #3366 should fix this issue?

asi1024:
> They fail when devices of different compute capabilities are used simultaneously but #3366 should fix this issue?

@emcastillo These tests try to compile in Device(0) and execute it in Device(1), so they don't work even if #3366 is applied.

> How about just adding a condition testing if dev0's CC is higher than that of dev1 (and skip if `True`) instead of removing them completely?

@leofang Do you want to guarantee that these tests work well in case of `Device(0).compute_capability >= Device(1).compte_capability`? It might be better that we raise an error to avoid bugs difficult to reproduce if the compile-time device id and runtime device id are different.

asi1024:
@leofang If you don't want to prohibit different device ids for this low-level feature, I will restore these tests and fix to add a skip condition.

emcastillo:
Hi,

We discussed out other alternatives such as detecting the architectures of the gpus in the system and compile the kernels for these architectures.
However, we don't know how often is the case of mixed compute capability devices  and the conditions which happen in these tests. So maybe it is not worth it and just deleting or skipping this test is enough when cc of devices is different

leofang:
Thanks @asi1024 and @emcastillo, 
> They fail when devices of different compute capabilities are used simultaneously

Sorry I'm not thinking clearly now and need your help to clarify. Do I understand you correctly that the two tests failed for CC (dev0) != CC (dev1) because the function loaded from the cubin generated with CC (dev 0) had no way to be recompiled for CC (dev 1)?

emcastillo:
Exactly!

leofang:
OK slowly catching up 😅 So for this consideration:
> other alternatives such as detecting the architectures of the gpus in the system and compile the kernels for these architectures.

recompiling would only work for ptx code (and CUDA driver would take care of it so we're good), not for cubin. Now I'm scratching my head thinking what to do with cubins...

Before we figure it out, I agree with your conclusion that
> skipping this test is enough when cc of devices is different

emcastillo:
Can we detect when a cubin was built for a different cc?
Just raising a comprehensible error in such a case would be fine, but 99.9% of the users won't fiddle with this stuff.

leofang:
I'm thinking about that too. I tried compiling a cubin with `sm_35` and loading it on my GPU with `sm_75`, and the error I got is `cupy.cuda.driver.CUDADriverError: CUDA_ERROR_INVALID_SOURCE: device kernel image is invalid`. Seems too general to extract useful info? 

I can look into the APIs and see if there's any diagnosis tool available without resorting to a subprocess call to nvcc and its friends...

leofang:
For the tools, I think unfortunately calling `cuobjdump` is the cheapest solution I can find: 
```bash
$ cuobjdump -lelf test_cubin_cc.cubin 
ELF file    1: test_cubin_cc.sm_35.cubin
```

On the other hand, despite lack of detailed explanation on the CUDA Toolkit Doc, it seems `CUDA_ERROR_INVALID_SOURCE` is the error designated for cubin version mismatching. I guess a try-except block enclosing the `_launch()` calls in `cupy/cuda/function.pyx` should be sufficient to provide a comprehensible error?

emcastillo:
Let's just skip or delete the tests, the use case is marginal and the solutions add more issues than they solve.

emcastillo:
Jenkins, test this please

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -749,11 +749,18 @@ def test_context_switch_RawModule2(self):
     def test_context_switch_RawModule3(self):
         # run test_load_cubin() on another device
         # generate cubin in the temp dir and load it on device 0
-        file_path = self._generate_file('cubin')
-        mod = cupy.RawModule(path=file_path, backend=self.backend)
+
+        device0 = cupy.cuda.Device(0)
+        device1 = cupy.cuda.Device(1)
+        if device0.compute_capability != device1.compute_capability:
+            raise pytest.skip()
```
leofang, line 756: Just a nit: could you do `self.skipTest()` to be compatible with `unittest`?

leofang:
I guess something is wrong in the failing CI's scipy? #3405 also failed in the same tests.

emcastillo:
Today scipy 1.5.0 was released, so probably is because of it.

tests/cupy_tests/core_tests/test_raw.py
emcastillo: there are some other places using `pytest.skip` instead `skipTest`, I will merge this as-is and then change all of them to obbey
https://docs-cupy.chainer.org/en/latest/contribution.html#how-to-run-tests

emcastillo:
Test failures are not related, merging.

tests/cupy_tests/core_tests/test_raw.py
leofang: Right, no worries. I think it can wait. I remember there was a discussion on which way to skip is better, but I couldn't recall where it is off top of my head.

leofang:
I think a less costly way to address different CCs is to examine the filename extension in `RawModule` and to test CCs only when cubins are loaded (ptx would be fine). We used to do this check but it was removed for some reason I don't remember.

PR #3461, opened 2020-06-19
Merge `CUPY_CUB_BLOCK_REDUCTION_DISABLED` and `CUB_DISABLED`

asi1024:
Part of #3445.
- Merges `CUPY_CUB_BLOCK_REDUCTION_DISABLED` and `CUB_DISABLED` into `CUPY_CUB_ENABLED`.
- Supports interfaces to switch to use each backend, or define their priorities.
  - Environment variable: `CUPY_BACKENDS`.
  - `cupy.core.set_reduction_backends` and `cupy.core.routine_backends` (These functions are for debug use. We will leave them undocumented.)

PR #3462, opened 2020-06-19
Add tests for `_cub_reduction`

leofang:
Follow-up of #3244. ~~Blocked by #3461.~~

This PR adds tests that should have been added in #3244. The main goals are:
1. Test that `MemoryPointer` can be used as parameters for kernel launch 
2. Test that `_can_use_cub_block_reduction` works as expected

TODO: 
- [x] Better coverage
- [x] Add PR description
- [x] Update the tests when the backend selection mechanism is settled


leofang:
@emcastillo This PR is unblocked and ready for review.

leofang:
Jenkins, test this plz

leofang:
> Jenkins, test this plz

Apparently Jenkins doesn't understand "plz" 😂
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
@emcastillo @asi1024 I added a few tests in a02d765 to ensure optuna can be used for reductions based on `create_reduction_func()`. PTAL.

leofang:
Jenkins, test this please

leofang:
I made both cub and nvcc paths cached, at the price of adding a few access functions for unit tests.

cupy/testing/helper.py
```diff
@@ -1263,14 +1262,28 @@ def test_values(self):
...
 
     def __init__(self, mock_mod, **kwargs):
+        """A handy wrapper for unittest.mock to check if a function is called.
+
+        This class should be used as a context manager.
+
+        Args:
+            mock_mod (str): the function to be mocked.
+            is_called (int or bool): the number of times the mocked function
+                should be called. Default is ``1``.
```
emcastillo, line 1272: Instead of having this as both `int` and `bool` how about adding a mutually exclusive
`times_called` argument?

cupy/core/_cub_reduction.pyx
```diff
@@ -641,3 +642,22 @@ cdef bint _try_to_call_cub_reduction(
         stream, params, cub_params)
 
     return True
+
+
+# only for testing
+def _get_cub_path():
+    return _cub_path
```
emcastillo, line 649: can't these be accessed from somewhere else, or just call `_environment` in the tests?

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -714,6 +725,19 @@ def test_template_failure(self):
                            match='named symbol not found'):
             mod.get_function('my_sqrt<double>')
 
+    def test_raw_pointer(self):
+        mod = cupy.RawModule(code=test_cast, backend=self.backend)
```
emcastillo, line 729: I think you are not ensuring that `MemoryPointer` is being used.

tests/cupyx_tests/test_optimize.py
```diff
@@ -120,3 +121,60 @@ def test_optimize_pickle(self):
...
+        _accelerator.set_routine_accelerators(self.old_routines)
+        _accelerator.set_reduction_accelerators(self.old_reductions)
+
+    def test_optimize1(self):
+        # Ensure the optimizer is run 3 times for all backends.
+        func = 'cupyx.optimizing._optimize._optimize'
+        is_called = 3
+
+        # Setting "wraps" is necessary to avoid compilation errors.
+        with testing.AssertFunctionIsCalled(
```
emcastillo, line 156: why not directly use `mock.patch` like the examples above?

cupy/core/_cub_reduction.pyx
leofang: No, unfortunately. They are initialized at the module import time, and the subsequent `_environment.get_*_path()` calls won't change them. I tried making them cpdef'd, but realized I hit a Cython limitation that led to this:
https://github.com/cupy/cupy/blob/78a83d71ed2518e7920c2bacd4e9c36b8381b45a/cupy/core/_routines_statistics.pyx#L584-L589

cupy/testing/helper.py
leofang: Sounds good.

tests/cupy_tests/core_tests/test_raw.py
leofang: What do you mean? `arr` here is a `MemoryPointer` instance. Perhaps I should rename it?

tests/cupyx_tests/test_optimize.py
leofang: Because I like my `AssertFunctionIsCalled` class 😁 Another reason is that I tested the aggregated counts to avoid calling `assert` multiple times. Let me know your preference.

cupy/testing/helper.py
emcastillo: you can also just replace `is_called` and fix the tests using it

cupy/core/_cub_reduction.pyx
emcastillo: Would it be possible to patch the global vars with mock to change the return values?
I don't think that having these setters around is a good idea ...

tests/cupy_tests/core_tests/test_raw.py
emcastillo: Sorry I was confused for a bit and thought you were directly passing the array to the kernel instead of the pointer.
LGTM

tests/cupyx_tests/test_optimize.py
emcastillo: I think this is fine, I was just curious 😁

cupy/core/_cub_reduction.pyx
leofang: For mock to work, they need to be Python objects, not cdef'd or cpdef'd. In the previous version I simply called `_environment.get_*_path()` every time they are needed so that mock tests work, but I am not a fan either...

Knowing that when nvcc or cub is absent we won't call the cub kernel, how about we just remove the two tests along with the setters?

cupy/testing/helper.py
leofang: Yeah it's easier...

cupy/core/_cub_reduction.pyx
emcastillo: I think its fine to remove those

leofang:
Done. I added two skip conditions in `test_cub_condition.py` for nvcc and cub. PTAL.

emcastillo:
Jenkins, test this please

PR #3478, opened 2020-06-24
Use SCM version to indicate development versions

kmaehashi:
When version is v8.0.0b3:

```
$ python setup.py --version
8.0.0b4.dev393+g6ddaac33e
```


leofang:
Just curious, do you know what's the difference between SCM and [versioneer](https://github.com/warner/python-versioneer)'s  approaches?

leofang:
I found some projects are moving away from versioneer to SCM: https://github.com/pytroll/satpy/issues/678

kmaehashi:
Actually I haven't ever used both 😅
AFAIK versioneer is feature-richer than setuptools_scm but needs to be included in the repository and seems not maintained well recently.  setuptools_scm is pypa's project (ie it's semi-official) and seems enough for us.

leofang:
Yeah the major concern for versioneer is it's no longer actively maintained... On the other hand, a valid argument is if it's included in the repo and it's working fine, it should not be maintained by upstream 😂 Anyway SCM is new and I'm happy there's a semi-official solution that could replace versioneer.

kmaehashi:
TODO: think about release sdist/wheel which is built before tagging at repo. Maybe add `--cupy-version x.y.z` option to override SCM version?

leofang:
I am curious, does SCM respect `cupy/_version.py`?

emcastillo (approved):

emcastillo:
Let's fix the conflict :(

emcastillo:
Jenkins, test this please

emcastillo:
```
raise RuntimeError('version information is not fouond: %s' % setup_path)
```

emcastillo:
seems like rtd is broken :(
```
ImportError: 
Error importing CuPy: version information could not be found.

It is likely that you are trying to import cupy from its source directory without building it;
please exit the cupy source tree, and relaunch your python interpreter from there.
```

emcastillo:
It seems to be working after adding the missing `setuptools_scm` package to the requirements for regular pip install and read the docs.

- I am not sure if `setuptools_scm` should be an install dependence, as sdist and the wheels will use the already bundled `VERSION` file. However it is needed for development and CIs.

Should we just do like we did with Cython, not add it as a dependence for other than CIs and modify the error message to let developers know that they need to install it.?

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

kmaehashi:
Blocked by : https://github.com/chainer/chainer-test/pull/636

emcastillo:
Jenkins, test this please

emcastillo (approved):

VERSION
```diff
@@ -0,0 +1 @@
+v9.0.0b2
```
leofang, line 1: This should be bumped.

leofang (changes requested):

leofang:
Looks like Conda-Forge already supports SCM: https://github.com/conda-forge/conda-forge.github.io/pull/1274

Issue #3482, opened 2020-06-24
Functionality to print nvcc and nvrtc output


mnicely:
```bash
CuPy Version          : 8.0.0b3
CUDA Root             : /usr/local/cuda
CUDA Build Version    : 10020
CUDA Driver Version   : 11000
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSOLVER Version      : (10, 3, 0)
cuSPARSE Version      : 10301
NVRTC Version         : (10, 2)
cuDNN Build Version   : None
cuDNN Version         : None
NCCL Build Version    : None
NCCL Runtime Version  : None
CUB Version           : Enabled
cuTENSOR Version      : None
```

I'm trying to get some register spillage information of some CuPy raw kernels. Unfortunately, the compiler fails on `-Xptxas`.
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#phase-options

```python
module = cp.RawModule(
    code=cuda_code,
    backend='nvcc',
    options=("-std=c++11", "-use_fast_math", "-lineinfo", '-Xptxas -warn-spills'),
)
```

leofang:
Just a thought: try splitting the string into two:
```python
    options=(..., '-Xptxas', '-warn-spills'),
```
I am guess `subprocess` is the unhappy guy.

mnicely:
That removed error! :smile:
But no output :confused: 
Added `-v` to try and force verbose, but no luck.

```python
module = cp.RawModule(
    code=cuda_code,
    backend='nvcc',
    options=("-std=c++11", "-use_fast_math", "-lineinfo", '-Xptxas', '-v -warn-spills'),
)
```

leofang:
Is the output written to stderr?

mnicely:
So I think I need to output from below. If I understand the correctly to output should be in stdout.
```python
def _run_nvcc(cmd, cwd):
    try:
        return subprocess.check_output(cmd, cwd=cwd, stderr=subprocess.STDOUT)
```

I'm rebuilding CuPy with this to see if I can see stdout. I'll know in a few minutes.
``` python
dump = _run_nvcc(cmd, root_dir)
print(dump)
```

leofang:
[...]

PR #3485, opened 2020-06-24
Added functionality to print nvcc and nvrtc output

mnicely:
### Info
This PR pertains to #3482. Currently, no output from the nvcc compiler is displayed to stdout. 

The intent of this implementation is to add a `verbose` flag to `RawModule` and `RawKernel` to be able to print nvcc output.

Initial push only modifies `cupy/cuda/compiler.py`.

@leofang I think the parent function is `compile_with_cache` (in `cupy/cuda/compiler.py`). But I haven't quite figured out exactly which functions that call it should be modified to accept the new flag.

```python
def compile_with_cache(
        source, options=(), arch=None, cache_dir=None, extra_source=None,
        backend='nvrtc', *, enable_cooperative_groups=False,
        name_expressions=None, verbose=False):
```

The actual print is performed here.
```python
def _run_nvcc(cmd, cwd, verbose):
    try:
        stdout = subprocess.check_output(cmd, cwd=cwd, stderr=subprocess.STDOUT,
                                         universal_newlines=True)
        if verbose is True:
            print(stdout)
        return stdout
```

### DIFF
--- a/cupy/core/raw.pxd
+++ b/cupy/core/raw.pxd
@@ -7,7 +7,7 @@ cdef class RawKernel:
         readonly tuple options
         readonly str backend
         readonly bint enable_cooperative_groups
-        readonly str log_stream
+        readonly log_stream
         tuple name_expressions
         bint translate_cucomplex
         list _kernel_cache
@@ -22,5 +22,5 @@ cdef class RawModule:
         readonly str backend
         readonly bint enable_cooperative_groups
         readonly tuple name_expressions
-        readonly str log_stream
+        readonly log_stream
         bint translate_cucomplex
--- a/cupy/core/raw.pxd
+++ b/cupy/core/raw.pxd
@@ -7,7 +7,7 @@ cdef class RawKernel:
         readonly tuple options
         readonly str backend
         readonly bint enable_cooperative_groups
-        readonly log_stream
+        object log_stream
         tuple name_expressions
         bint translate_cucomplex
         list _kernel_cache
@@ -22,5 +22,5 @@ cdef class RawModule:
         readonly str backend
         readonly bint enable_cooperative_groups
         readonly tuple name_expressions
-        readonly log_stream
+        object log_stream
         bint translate_cucomplex
--- a/cupy/core/raw.pxd
+++ b/cupy/core/raw.pxd
@@ -7,6 +7,7 @@ cdef class RawKernel:
         readonly tuple options
         readonly str backend
         readonly bint enable_cooperative_groups
+        readonly str log_stream
         tuple name_expressions
         bint translate_cucomplex
         list _kernel_cache
@@ -21,4 +22,5 @@ cdef class RawModule:
         readonly str backend
         readonly bint enable_cooperative_groups
         readonly tuple name_expressions
+        readonly str log_stream
         bint translate_cucomplex
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -35,7 +35,10 @@ cdef class RawKernel:
             ``cuLaunchCooperativeKernel`` so that cooperative groups can be
             used from the CUDA source.
             This feature is only supported in CUDA 9 or later.
-        log_stream (str): TODO
+        log_stream (str): Whether the output of nvcc should be displayed in
+            'stdout' or stored to a 'file'. For example, `log_stream('file.txt')`
+            will write to output to file.txt. This variable is not used by 
+            nvrtc.
     """
 
     def __init__(self, str code, str name, tuple options=(),
@@ -260,7 +263,10 @@ cdef class RawModule:
             the template kernel ``func1<T>`` and non-template kernel ``func2``.
             Strings in this tuple must then be passed, one at a time, to
             :meth:`get_function` to retrieve the corresponding kernel.
-        log_stream (str): TODO
+        log_stream (str): Whether the output of nvcc should be displayed in
+            'stdout' or stored to a 'file'. For example, `log_stream('file.txt')`
+            will write to output to file.txt. This variable is not used by 
+            nvrtc.
 
     .. note::
         Each kernel in ``RawModule`` possesses independent function attributes.
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -36,8 +36,8 @@ cdef class RawKernel:
             used from the CUDA source.
             This feature is only supported in CUDA 9 or later.
         log_stream (str): Whether the output of nvcc should be displayed in
-            'stdout' or stored to a 'file'. For example, `log_stream('file.txt')`
-            will write to output to file.txt.
+            'stdout' or stored to a 'file'. For example, ``log_stream='./file.txt'``
+            will write to output to ``file.txt`` in the current directory.
     """
 
     def __init__(self, str code, str name, tuple options=(),
@@ -263,8 +263,8 @@ cdef class RawModule:
             Strings in this tuple must then be passed, one at a time, to
             :meth:`get_function` to retrieve the corresponding kernel.
         log_stream (str): Whether the output of nvcc should be displayed in
-            'stdout' or stored to a 'file'. For example, `log_stream('file.txt')`
-            will write to output to file.txt.
+            'stdout' or stored to a 'file'. For example, ``log_stream='./file.txt'``
+            will write to output to ``file.txt`` in the current directory.
 
     .. note::
         Each kernel in ``RawModule`` possesses independent function attributes.
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -35,11 +35,12 @@ cdef class RawKernel:
             ``cuLaunchCooperativeKernel`` so that cooperative groups can be
             used from the CUDA source.
             This feature is only supported in CUDA 9 or later.
+        log_stream (str): TODO
     """
 
     def __init__(self, str code, str name, tuple options=(),
                  str backend='nvrtc', *, bint translate_cucomplex=False,
-                 bint enable_cooperative_groups=False):
+                 bint enable_cooperative_groups=False, str log_stream=None):
 
         self.code = code
         self.name = name
@@ -47,6 +48,7 @@ cdef class RawKernel:
         self.backend = backend
         self.translate_cucomplex = translate_cucomplex
         self.enable_cooperative_groups = enable_cooperative_groups
+        self.log_stream = log_stream
 
         # only used when RawKernels are produced from RawModule
         self.file_path = None  # for cubin/ptx
@@ -94,7 +96,7 @@ cdef class RawKernel:
             mod = _get_raw_module(
                 self.code, self.file_path, self.options, self.backend,
                 self.translate_cucomplex, self.enable_cooperative_groups,
-                self.name_expressions)
+                self.name_expressions, self.log_stream)
             ker = mod.get_function(self.name)
             self._kernel_cache[dev] = ker
         return ker
@@ -258,6 +260,7 @@ cdef class RawModule:
             the template kernel ``func1<T>`` and non-template kernel ``func2``.
             Strings in this tuple must then be passed, one at a time, to
             :meth:`get_function` to retrieve the corresponding kernel.
+        log_stream (str): TODO
 
     .. note::
         Each kernel in ``RawModule`` possesses independent function attributes.
@@ -265,7 +268,7 @@ cdef class RawModule:
     def __init__(self, *, str code=None, str path=None, tuple options=(),
                  str backend='nvrtc', bint translate_cucomplex=False,
                  bint enable_cooperative_groups=False,
-                 name_expressions=None):
+                 tuple name_expressions=None, str log_stream=None):
         if (code is None) == (path is None):
             raise TypeError(
                 'Exactly one of `code` and `path` keyword arguments must be '
@@ -290,6 +293,7 @@ cdef class RawModule:
         self.code = code
         self.file_path = path
         self.enable_cooperative_groups = enable_cooperative_groups
+        self.log_stream = log_stream
 
         if self.code is not None:
             self.options = options
@@ -312,7 +316,7 @@ cdef class RawModule:
         mod = _get_raw_module(
             self.code, self.file_path, self.options, self.backend,
             self.translate_cucomplex, self.enable_cooperative_groups,
-            self.name_expressions)
+            self.name_expressions, self.log_stream)
         return mod
 
     def get_function(self, str name):
@@ -366,7 +370,8 @@ cdef class RawModule:
         ker = RawKernel(
             self.code, name, self.options, self.backend,
             translate_cucomplex=self.translate_cucomplex,
-            enable_cooperative_groups=self.enable_cooperative_groups)
+            enable_cooperative_groups=self.enable_cooperative_groups
+            log_stream=self.log_stream)
         # for lookup in case we loaded from cubin/ptx
         ker.file_path = self.file_path
         # for lookup in case we specialize a template
@@ -423,14 +428,15 @@ cdef class RawModule:
 def _get_raw_module(str code, str path, tuple options=(), str backend='nvrtc',
                     bint translate_cucomplex=False,
                     bint enable_cooperative_groups=False,
-                    tuple name_expressions=None):
+                    tuple name_expressions=None, str log_stream=None):
     cdef Module mod
     if code is not None:
         mod = cupy.core.core.compile_with_cache(
             code, options, prepend_cupy_headers=False, backend=backend,
             translate_cucomplex=translate_cucomplex,
             enable_cooperative_groups=enable_cooperative_groups,
-            name_expressions=name_expressions)
+            name_expressions=name_expressions,
+            log_stream=log_stream)
     elif path is not None:
         mod = Module()
         mod.load_file(path)
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -43,7 +43,7 @@ cdef class RawKernel:
 
     def __init__(self, str code, str name, tuple options=(),
                  str backend='nvrtc', *, bint translate_cucomplex=False,
-                 bint enable_cooperative_groups=False, str log_stream=None):
+                 bint enable_cooperative_groups=False, log_stream=None):
 
         self.code = code
         self.name = name
@@ -274,7 +274,7 @@ cdef class RawModule:
     def __init__(self, *, str code=None, str path=None, tuple options=(),
                  str backend='nvrtc', bint translate_cucomplex=False,
                  bint enable_cooperative_groups=False,
-                 name_expressions=None, str log_stream=None):
+                 name_expressions=None, log_stream=None):
         if (code is None) == (path is None):
             raise TypeError(
                 'Exactly one of `code` and `path` keyword arguments must be '
@@ -434,7 +434,7 @@ cdef class RawModule:
 def _get_raw_module(str code, str path, tuple options=(), str backend='nvrtc',
                     bint translate_cucomplex=False,
                     bint enable_cooperative_groups=False,
-                    tuple name_expressions=None, str log_stream=None):
+                    tuple name_expressions=None, log_stream=None):
     cdef Module mod
     if code is not None:
         mod = cupy.core.core.compile_with_cache(
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -274,7 +274,7 @@ cdef class RawModule:
     def __init__(self, *, str code=None, str path=None, tuple options=(),
                  str backend='nvrtc', bint translate_cucomplex=False,
                  bint enable_cooperative_groups=False,
-                 name_expressions=None, str log_stream=None):
+                 tuple name_expressions=None, str log_stream=None):
         if (code is None) == (path is None):
             raise TypeError(
                 'Exactly one of `code` and `path` keyword arguments must be '
@@ -376,7 +376,7 @@ cdef class RawModule:
         ker = RawKernel(
             self.code, name, self.options, self.backend,
             translate_cucomplex=self.translate_cucomplex,
-            enable_cooperative_groups=self.enable_cooperative_groups
+            enable_cooperative_groups=self.enable_cooperative_groups,
             log_stream=self.log_stream)
         # for lookup in case we loaded from cubin/ptx
         ker.file_path = self.file_path
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -37,8 +37,7 @@ cdef class RawKernel:
             This feature is only supported in CUDA 9 or later.
         log_stream (str): Whether the output of nvcc should be displayed in
             'stdout' or stored to a 'file'. For example, `log_stream('file.txt')`
-            will write to output to file.txt. This variable is not used by 
-            nvrtc.
+            will write to output to file.txt.
     """
 
     def __init__(self, str code, str name, tuple options=(),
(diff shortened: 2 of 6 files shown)


leofang:
Thanks, Matt! I think this is more for library developers like us than for general users. I'd prefer reading an env var (say, `CUPY_SHOW_COMPILER_WARNING`) to passing in an argument from outside through several layers (there are several env vars used in this module: https://docs-cupy.chainer.org/en/latest/reference/environment.html). Someone from the CuPy team will make the final decision. (Currently the codebase is frozen and they'll make a beta release this week.)

mnicely:
No problem! That sounds a lot better actually :smile: 
Let me know if you need anything else from me.

emcastillo:
Hi, we were discussing about this pr.

While the env var idea is good, it will affect all the kernels in a script, so I think that having a kwarg is fine as we might be interested only in specific kernels messages.

However, instead of verbose and calling print, I would add a `log_stream` parameter that allows to write to a stream, either a file or stdout. That should make it cleaner and versatile. If `log_stream is None` then no output is recorded :).

Thanks!

mnicely:
@emcastillo I made changes per your comments. Let me know what you think.

I still don't know where exactly `compile_with_cache` gets called to propagate the required changes.

leofang:
@emcastillo So do you want `log_stream` to be passed in all the way from `RawKernel`/`RawModule` (which is the only way this variable is intended to be used I guess)?

@mnicely `RawKernel`/`RawModule` call `compile_with_cache()` from `cupy/core/core.pyx`, which calls `compile_with_cache()` from `cupy/cuda/compiler.py` 😂 It's better to append `log_stream=None` to their argument lists.

mnicely:
@leofang @emcastillo Thanks for the help! I think I'm a lot closer to want you want with the latest push. I'm getting the following error in the build process and currently investigating :smile: 

```bash
    [18/40] Cythonizing cupy/core/fusion.pyx
    [19/40] Cythonizing cupy/core/internal.pyx
    [20/40] Cythonizing cupy/core/raw.pyx
ERROR: Command errored out with exit status 1: /home/belt/anaconda3/envs/lmc/bin/python3.8 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/belt/workStuff/cupy/setup.py'"'"'; __file__='"'"'/home/belt/workStuff/cupy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps Check the logs for full command output.
Exception information:
  File "/home/belt/anaconda3/envs/lmc/lib/python3.8/site-packages/pip/_internal/cli/base_command.py", line 188, in _main
    status = self.run(options, args)
  File "/home/belt/anaconda3/envs/lmc/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 185, in wrapper
    return func(self, options, args)
  File "/home/belt/anaconda3/envs/lmc/lib/python3.8/site-packages/pip/_internal/commands/install.py", line 398, in run
    installed = install_given_reqs(
  File "/home/belt/anaconda3/envs/lmc/lib/python3.8/site-packages/pip/_internal/req/__init__.py", line 67, in install_given_reqs
    requirement.install(
  File "/home/belt/anaconda3/envs/lmc/lib/python3.8/site-packages/pip/_internal/req/req_install.py", line 780, in install
    install_editable_legacy(
  File "/home/belt/anaconda3/envs/lmc/lib/python3.8/site-packages/pip/_internal/operations/install/editable_legacy.py", line 49, in install_editable
    call_subprocess(
  File "/home/belt/anaconda3/envs/lmc/lib/python3.8/site-packages/pip/_internal/utils/subprocess.py", line 241, in call_subprocess
    raise InstallationError(exc_msg)
pip._internal.exceptions.InstallationError: Command errored out with exit status 1: /home/belt/anaconda3/envs/lmc/bin/python3.8 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/belt/workStuff/cupy/setup.py'"'"'; __file__='"'"'/home/belt/workStuff/cupy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps Check the logs for full command output.
Removed build tracker: '/tmp/pip-req-tracker-7x3yy_wp'
```

I definitely have a greater appreciation for the amazing API y'all have built!! :confetti_ball:

cupy/core/raw.pyx
```diff
@@ -258,14 +263,18 @@ cdef class RawModule:
...
+            nvrtc.
 
     .. note::
         Each kernel in ``RawModule`` possesses independent function attributes.
     """
     def __init__(self, *, str code=None, str path=None, tuple options=(),
                  str backend='nvrtc', bint translate_cucomplex=False,
                  bint enable_cooperative_groups=False,
-                 name_expressions=None):
+                 tuple name_expressions=None, str log_stream=None):
```
leofang, line 277: We made it possible to accept any iterable (say, `list`) by not typing it:
```suggestion
                 name_expressions=None, str log_stream=None):
```

cupy/cuda/compiler.py
```diff
@@ -25,9 +25,19 @@ class NVCCException(Exception):
...
+        stdout = subprocess.check_output(cmd, cwd=cwd,
+                                         stderr=subprocess.STDOUT,
+                                         universal_newlines=True)
+        if log_stream is None:
+            pass
+        elif log_stream == 'stdout':
+            print(stdout)
+        else:
+            with open(log_stream, 'w') as f:
+                f.write(stdout)
```
leofang, line 39:
```suggestion
        if log_stream == 'stdout':
            print(stdout)
        elif log_stream is not None:
              with open(log_stream, 'w') as f:
                  f.write(stdout)

File: cupy/cuda/compiler.py [hunk 3]
```diff
@@ -331,11 +341,11 @@ def _compile_with_cache_cuda(
     if _get_bool_env_variable('CUPY_CUDA_COMPILE_WITH_DEBUG', False):
         options += ('--device-debug', '--generate-line-info')
 
-    env = (arch, options, _get_nvrtc_version(), backend)
+    env = (arch, options, _get_nvrtc_version(), backend, log_stream)
```
leofang, line 344: I'd not change this line. This is for generating a key, but `log_stream` would not affect compilation results, so it won't help here.
```suggestion
    env = (arch, options, _get_nvrtc_version(), backend)
```

cupy/cuda/compiler.py
```diff
@@ -378,8 +388,9 @@ def _compile_with_cache_cuda(
         mod._set_mapping(mapping)
     elif backend == 'nvcc':
         rdc = _is_cudadevrt_needed(options)
-        cubin = compile_using_nvcc(source, options, arch, name + '.cu',
-                                   code_type='cubin', separate_compilation=rdc)
+        cubin = compile_using_nvcc(source, options, log_stream, arch,
+                                   name + '.cu', code_type='cubin',
+                                   separate_compilation=rdc)
```
leofang, line 393: The order is wrong:
```suggestion
        cubin = compile_using_nvcc(source, options, arch, name + '.cu',
                                   code_type='cubin', separate_compilation=rdc,
                                   log_stream=log_stream)
```

cupy/cuda/function.pyx
```diff
@@ -144,7 +144,7 @@ cdef inline size_t _get_stream(stream) except *:
 cdef _launch(intptr_t func, Py_ssize_t grid0, int grid1, int grid2,
              Py_ssize_t block0, int block1, int block2,
              args, Py_ssize_t shared_mem, size_t stream,
-             bint enable_cooperative_groups=False):
+             bint enable_cooperative_groups=False, str log_stream=None):
```
leofang, line 147: I don't think you need to touch this file, because by the time `launch` is called, the compilation is already done, and there's no place where `log_stream` can be used.
```suggestion
             bint enable_cooperative_groups=False):
```

cupy/cuda/function.pyx
```diff
@@ -175,15 +175,15 @@ cdef class Function:
         self.ptr = driver.moduleGetFunction(module.ptr, funcname)
 
     def __call__(self, tuple grid, tuple block, args, size_t shared_mem=0,
-                 stream=None, enable_cooperative_groups=False):
+                 stream=None, enable_cooperative_groups=False, log_stream=None):
```
leofang, line 178: ditto
```suggestion
                 stream=None, enable_cooperative_groups=False):
```

cupy/cuda/function.pyx
```diff
@@ -175,15 +175,15 @@ cdef class Function:
...
         grid = (grid + (1, 1))[:3]
         block = (block + (1, 1))[:3]
         s = _get_stream(stream)
         _launch(
             self.ptr,
             max(1, grid[0]), max(1, grid[1]), max(1, grid[2]),
             max(1, block[0]), max(1, block[1]), max(1, block[2]),
-            args, shared_mem, s, enable_cooperative_groups)
+            args, shared_mem, s, enable_cooperative_groups, log_stream)
```
leofang, line 186:
```suggestion
            args, shared_mem, s, enable_cooperative_groups)
```

mnicely:
Thanks @leofang! Still having a failure on `[20/40] Cythonizing cupy/core/raw.pyx` so I'm still investigating :confused:

cupy/core/raw.pyx
```diff
@@ -366,7 +376,8 @@ cdef class RawModule:
         ker = RawKernel(
             self.code, name, self.options, self.backend,
             translate_cucomplex=self.translate_cucomplex,
-            enable_cooperative_groups=self.enable_cooperative_groups)
+            enable_cooperative_groups=self.enable_cooperative_groups
```
leofang, line 379: Caught an obvious error 😂, not sure if this is the only one...
```suggestion
            enable_cooperative_groups=self.enable_cooperative_groups,
```

mnicely: HA! Thanks, I had just found it :joy:

mnicely: Thanks for taking to time to help me! Going to do some more testing before my next push.

leofang: Whenever I touch `raw.pyx` and/or `compiler.py` I'd always run `purest tests/cupy_tests/core_tests/test_raw.py` as a quick check to ensure I don't mess up 😁

Right, I wonder about NVRTC. Perhaps the warning log is generated in the log, fetched by `nvrtcGetProgramLog`? Then we will need to examine the log when `log_stream is not None`?

mnicely:
I got everything to build and it seems to be working correctly with `nvcc`.

I attempted to add log_stream to `nvrtc` but I'm not too familiar with the output.

`compiler.py` at line 156
```python
        try:
            ptx, mapping = prog.compile(options)
            if log_stream == 'stdout':
                print(ptx)
                print()
                print(mapping)
            elif log_stream is not None:
                with open(log_stream, 'w') as f:
                    f.write(ptx)
                    f.write("\n")
                    f.write(mapping)
```

`compiler.py` at line 271
```python
            result, _ = prog.compile(options)
            if log_stream == 'stdout':
                print(result)
            elif log_stream is not None:
                with open(log_stream, 'w') as f:
                    f.write(result)
```

cupy/core/raw.pyx
mnicely: All tests failed. :facepalm: 
I'll take a look at it in the morning.

mnicely:
All tests are passing now :smile: 

```bash
(lmc) belt@orion:~/workStuff/cupy$ pytest tests/cupy_tests/core_tests/test_raw.py
==================================================================================================================================================================================== test session starts ====================================================================================================================================================================================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/belt/workStuff/cupy, inifile: setup.cfg
collected 56 items                                                                                                                                                                                                                                                                                                                                                                          

tests/cupy_tests/core_tests/test_raw.py .................................ss..............s......                                                                                                                                                                                                                                                                                      [100%]

============================================================================================================================================================================== 53 passed, 3 skipped in 14.32s ===============================================================================================================================================================================
```

leofang:
Hi @mnicely I think for NVRTC the current handing isn't correct. The printed out things are simply `ptx` code and the mapping from template name to mangled name.

What you need instead is to pass `log_stream` to `prog.compile()` (there are two places):
https://github.com/cupy/cupy/blob/c8f03d978aa917d97e8c53542a3c8b30d7c5529c/cupy/cuda/compiler.py#L148
Then, in the `compile` method in `_NVRTCProgram`, you print out `log = nvrtc.getProgramLog(self.ptr)` before returning. Note that currently this log is accessed only when an exception is raised:
https://github.com/cupy/cupy/blob/c8f03d978aa917d97e8c53542a3c8b30d7c5529c/cupy/cuda/compiler.py#L480

cupy/core/raw.pyx
```diff
@@ -258,14 +263,18 @@ cdef class RawModule:
...
+            nvrtc.
 
     .. note::
         Each kernel in ``RawModule`` possesses independent function attributes.
     """
     def __init__(self, *, str code=None, str path=None, tuple options=(),
                  str backend='nvrtc', bint translate_cucomplex=False,
                  bint enable_cooperative_groups=False,
-                 name_expressions=None):
+                 name_expressions=None, str log_stream=None):
```
emcastillo, line 277: backend='nvrtc' and log_stream != None are not compatible, we should add a check

leofang: I think Matt will try to support both 😁

mnicely:
@emcastillo @leofang 
Thanks to all leo's help I believe this ticket is working as expected. And passing all tests.

cupy/cuda/compiler.py
```diff
@@ -25,18 +25,25 @@ class NVCCException(Exception):
...
     except subprocess.CalledProcessError as e:
         msg = ('`nvcc` command returns non-zero exit status. \n'
                'command: {0}\n'
                'return-code: {1}\n'
                'stdout/stderr: \n'
                '{2}'.format(e.cmd,
                             e.returncode,
-                            e.output.decode(encoding='UTF-8',
-                                            errors='replace')))
+                            e.output))
```
leofang, line 46: Could you comment why we need this change?

cupy/core/raw.pyx
```diff
@@ -35,18 +35,22 @@ cdef class RawKernel:
             ``cuLaunchCooperativeKernel`` so that cooperative groups can be
             used from the CUDA source.
             This feature is only supported in CUDA 9 or later.
+        log_stream (str): Whether the output of nvcc should be displayed in
+            'stdout' or stored to a 'file'. For example, `log_stream('file.txt')`
+            will write to output to file.txt.
```
leofang, line 40:
```suggestion
            'stdout' or stored to a 'file'. For example, ``log_stream='./file.txt'``
            will write to output to ``file.txt`` in the current directory.
```

cupy/core/raw.pyx
```diff
@@ -258,14 +262,17 @@ cdef class RawModule:
             the template kernel ``func1<T>`` and non-template kernel ``func2``.
             Strings in this tuple must then be passed, one at a time, to
             :meth:`get_function` to retrieve the corresponding kernel.
+        log_stream (str): Whether the output of nvcc should be displayed in
+            'stdout' or stored to a 'file'. For example, `log_stream('file.txt')`
+            will write to output to file.txt.
```
leofang, line 267:
```suggestion
            'stdout' or stored to a 'file'. For example, ``log_stream='./file.txt'``
            will write to output to ``file.txt`` in the current directory.
```

cupy/cuda/compiler.py
```diff
@@ -244,13 +251,15 @@ def compile_using_nvcc(source, options=(), arch=None,
             assert False, code_type
 
 
-def _preprocess(source, options, arch, backend):
+def _preprocess(source, options, arch, backend, log_stream):
```
leofang, line 254: I am sorry, Matt, this is my bad. I looked again and no longer think we need `log_stream` in this `_preprocess()` function. The purpose of this function is to check the compiler config by compiling an empty source code.

leofang: By not touching this function, it also ensures we don't write to the same file twice (although it has no harm).

cupy/cuda/compiler.py
mnicely: Ahh, yes. By adding `universal_newlines=True` to `check_output` we are creating a string. Therefore there is no need to `decode`. 

Here is the error without the code change
```bash
                   '{2}'.format(e.cmd,
                                e.returncode,
>                               e.output.decode(encoding='UTF-8',
                                                errors='replace')))
E           AttributeError: 'str' object has no attribute 'decode'
```

mnicely:
Okay, addressed latest issues. 

Now there is just one issue I'm seeing. 
When I run a script multiple times with `nvcc`, I'm seeing an output each execution.

But when I do the same thing with 'nvrtc`, I'm only seeing output on first execution.
The only way to force another output is to modify the RawModule. 
I'm guessing it's being cached, but is it doing so between multiple scripts launchs?

cupy/cuda/compiler.py
```diff
@@ -250,7 +257,9 @@ def _preprocess(source, options, arch, backend):
 
         prog = _NVRTCProgram(source, '')
         try:
+
             result, _ = prog.compile(options)
+
```
leofang, line 262: Just a nitpick: could you remove the empty lines?

cupy/cuda/compiler.py
leofang: Ah, good catch! Thanks Matt.

leofang:
btw I noticed you left a comment on outputs for repeated compilations. Is that still a concern? If you don't clean the cached cubins in `~/.cupy/kernel_cache/`, I'd expect the compilation only happens once (so one output).

cupy/cuda/compiler.py
mnicely: Fixed.

mnicely:
Okay, all good :smile: 

> btw I noticed you left a comment on outputs for repeated compilations. Is that still a concern? If you don't clean the cached cubins in `~/.cupy/kernel_cache/`, I'd expect the compilation only happens once (so one output).

It was a mistake on my part :smile: 
All looks good now.

leofang:
btw I think the title can be changed as both nvcc and nvrtx are supported now

cupy/cuda/compiler.py
```diff
@@ -25,18 +25,25 @@ class NVCCException(Exception):
...
 
 
-def _run_nvcc(cmd, cwd):
+def _run_nvcc(cmd, cwd, log_stream=None):
     try:
-        return subprocess.check_output(cmd, cwd=cwd, stderr=subprocess.STDOUT)
+        stdout = subprocess.check_output(cmd, cwd=cwd,
+                                         stderr=subprocess.STDOUT,
+                                         universal_newlines=True)
+        if log_stream == 'stdout':
```
emcastillo, line 33: The idea is to use `sys.stdout` value as the `log_stream` parameter so we can avoid explicitly calling `print` 😃

cupy/cuda/compiler.py
```diff
@@ -475,6 +485,11 @@ def compile(self, options=()):
                 mapping = {}
                 for ker in self.name_expressions:
                     mapping[ker] = nvrtc.getLoweredName(self.ptr, ker)
+            if log_stream == 'stdout':
+                print(nvrtc.getProgramLog(self.ptr))
```
emcastillo, line 489: use the stream directly instead of calling print.

cupy/cuda/compiler.py
mnicely: @emcastillo Are you saying to replace `print` with

```python
sys.stdout = stdout
```

cupy/cuda/compiler.py
mnicely: Same here?

```python
sys.stdout = nvrtc.getProgramLog(self.ptr)

File: cupy/cuda/compiler.py (hunk 15 above)
Comment (emcastillo) [reply cupy/cuda/compiler.py:33]: no, what I mean is to pass `sys.stdout` as the `log_stream` for the kernel.

```python
RawKernel(*args, log_stream=sys.stdout)
```
So inside this code we only do
```
log_stream.write(log)
```
The user should be responsible for opening the stream

leofang:
@mnicely If you want I can help finish the last mile.

mnicely:
@leofang @emcastillo. Yes. I’ll finish this in the next 48 hours. I’m out of town for my nephew’s 7th bday. With limited computer access. Sorry for the delay!!

leofang:
No worries, Matt! Was just checking in 😅 Enjoy your family time!

cupy/cuda/compiler.py
mnicely: @emcastillo Check it out now. I think I understand what you're saying, and yes, it is much simpler. :smile:

cupy/core/raw.pxd
```diff
@@ -7,7 +7,7 @@ cdef class RawKernel:
         readonly tuple options
         readonly str backend
         readonly bint enable_cooperative_groups
-        readonly str log_stream
+        readonly log_stream
```
leofang, line 10: Just a nitpick, I no longer think it's useful to expose this to end users:
```suggestion 
        object log_stream
```

cupy/core/raw.pxd
```diff
@@ -22,5 +22,5 @@ cdef class RawModule:
         readonly str backend
         readonly bint enable_cooperative_groups
         readonly tuple name_expressions
-        readonly str log_stream
+        readonly log_stream
```
leofang, line 25: ditto
```suggestion
        object log_stream
```

cupy/cuda/compiler.py
```diff
@@ -1,5 +1,6 @@
 import hashlib
 import math
+import io
```
leofang, line 3: This is not needed.
```suggestion
```

cupy/cuda/compiler.py
```diff
@@ -25,18 +26,22 @@ class NVCCException(Exception):
     pass
 
 
-def _run_nvcc(cmd, cwd):
+def _run_nvcc(cmd, cwd, log_stream=None):
     try:
-        return subprocess.check_output(cmd, cwd=cwd, stderr=subprocess.STDOUT)
+        log = subprocess.check_output(cmd, cwd=cwd,
+                                         stderr=subprocess.STDOUT,
+                                         universal_newlines=True)
```
leofang, line 33: Flake8:
```suggestion
                                      stderr=subprocess.STDOUT,
                                      universal_newlines=True)
```

cupy/core/raw.pyx
```diff
@@ -35,18 +35,23 @@ cdef class RawKernel:
             ``cuLaunchCooperativeKernel`` so that cooperative groups can be
             used from the CUDA source.
             This feature is only supported in CUDA 9 or later.
+        log_stream (str): Whether the output of nvcc should be displayed in
+            'stdout' or stored to a 'file'. For example,
+            ``log_stream='./file.txt'`` will write to output to
+            ``file.txt`` in the current directory.
```
leofang, line 41: Ouch...Sorry I missed this! How about:
```
log_stream (object): Pass either ``sys.stdout`` or a file object to which the compiler output will be written.
```

cupy/core/raw.pyx
```diff
@@ -258,14 +263,18 @@ cdef class RawModule:
             the template kernel ``func1<T>`` and non-template kernel ``func2``.
             Strings in this tuple must then be passed, one at a time, to
             :meth:`get_function` to retrieve the corresponding kernel.
+        log_stream (str): Whether the output of nvcc should be displayed in
+            'stdout' or stored to a 'file'. For example,
+            ``log_stream='./file.txt'`` will write to output to
+            ``file.txt`` in the current directory.
```
leofang, line 269: Ditto...

cupy/core/raw.pyx
mnicely: Good catch!

emcastillo:
Jenkins, test this please

Issue #3115, opened 2020-02-23
'csc_matrix' object does not support item assignment


surajp92:
I am using sparse matrices using CuPy. If I define the csc_matrix using scipy, I can assign the matrix after its definition. But, I am not able to do it with CuPy. Below is my code:

```
data = np.ones((3,4))
diags = np.array([-1,0,1])
M= spdiags(data, diags, 4, 4)
M = sp.csc_matrix(M)
print(M[0,0])
M[0,0] = 5
print(M[0,0])
```
I get 1.0, 5.0 as the output. If I try to do same thing in CuPy I get an error.
```
import cupy as cp
import cupyx as cpx
datac = cp.ones((3,4))
diagsc = cp.array(diags)
Mc = cpx.scipy.sparse.spdiags(datac, diagsc, 4, 4)
Mc = cpx.scipy.sparse.csc_matrix(Mc)

Mc = cpx.scipy.sparse.csc_matrix(Mc)
print(Mc.get()[0,0])
Mc[0,0] = 5
print(Mc[0,0])
```
I get 1.0 as th eoutput and then the error: 'csc_matrix' object does not support item assignment.

Am I doing something wrong?

Also, the assignment operation with csc_matrix is expensive in SciPy. But, I can use lil_matrix in scipy which is much faster. Does CuPy have a sparse lil_matirx option?

Thank you.

YonatanGM:
I am having the same problem. Item assignment doesn't work with 'csc_matrix' object.  
A = cpx.scipy.sparse.csc_matrix(A)
 A[0,0] = 1, 

gives 'csc_matrix' object does not support item assignment. This is not an issue with the scipy version ..

MikGen:
Same here. I was transferring my code into GPU with cupy. While CPU's scipy worked well, similar GPU code fails with this error

jakirkham:
@MikGen, would recommend raising a new issue with the minimal amount of code needed to reproduce the error on the latest version of CuPy 🙂

PR #3486, opened 2020-06-24
[REVIEW] Compressed Sparse Inner Indexing

### LINKED ISSUE #2360: Array indexing of sparse matrices
After talking to @quasiben I've been looking into using CuPy as a shim to support sparse matrices in PyTorch (the current support is [not great](https://github.com/pytorch/pytorch/issues/9674)).

The gist is that I can fit a sparse matrix into GPU memory but not a dense one. So, the idea is to use a CuPy sparse matrix and a custom DataLoader: I should be able to quickly train a model by indexing into the CuPy matrix, densifying a batch of data, and doing a zero-copy conversion to a tensor to feed into my model.

Unfortunately it appears that CuPy's sparse matrices don't support array indexing, which would be the natural way to do this (in the sense that PyTorch's DataLoader class returns a list of integers to index into your data). There's definitely a way to get this working with single indices and `vstack` but I thought I'd raise an issue because scipy does support this and it'd be a lot cleaner for me.

Thanks!

CuPy Version          : 6.2.0
CUDA Build Version    : 9000
CUDA Driver Version   : 9000
CUDA Runtime Version  : 9000
cuDNN Build Version   : 7402
cuDNN Version         : 7402
NCCL Build Version    : 2307
NCCL Runtime Version  : 2307
* Code to reproduce
```python
scipy_sparse = scipy.sparse.random(100, 100).tocsr()
scipy_sparse[[0,1],:] # this works for csr matrices

cupy_sparse = cupy.sparse.csr_matrix(scipy_sparse)
cupy_sparse[[0,1],:] # error :(
```
```
<ipython-input-70-425afa4195a2> in <module>
      3 
      4 cupy_sparse = cupy.sparse.csr_matrix(scipy_sparse)
----> 5 cupy_sparse[[0,1],:]

/opt/conda/lib/python3.7/site-packages/cupyx/scipy/sparse/compressed.py in __getitem__(self, slices)
    247                 return self._get_major_slice(major)
    248 
--> 249         raise ValueError('unsupported indexing')
    250 
    251     def _get_single(self, major, minor):

ValueError: unsupported indexing
```

### LINKED ISSUE #3178: Sparse arrays do not support all of Scipy's slicing options
I encountered this behavior in the most recent (CuPy 8.0.0 at the time of writing this) branch when attempting to call `dask.array.from_array()` with a `csr_matrix`. It appears that Dask is attempting to do some slicing that is supported in Scipy's `csr_matrix` but is not supported by CuPy. 

The following is a reproducible example:

```
>>> import cupy as cp
>>> import dask.array
>>> 
>>> a = cp.sparse.random(100, 100, format='csr')
>>> b = dask.array.from_array(a)
>>> b.compute()
  File "/share/workspace/dask/dask/base.py", line 165, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/share/workspace/dask/dask/base.py", line 436, in compute
    results = schedule(dsk, keys, **kwargs)
  File "/share/workspace/dask/dask/threaded.py", line 81, in get
    **kwargs
  File "/share/workspace/dask/dask/local.py", line 486, in get_async
    raise_exception(exc, tb)
  File "/share/workspace/dask/dask/local.py", line 316, in reraise
    raise exc
  File "/share/workspace/dask/dask/local.py", line 222, in execute_task
    result = _execute_task(task, data)
  File "/share/workspace/dask/dask/core.py", line 119, in _execute_task
    return func(*args2)
  File "/share/workspace/dask/dask/array/core.py", line 104, in getter
    c = a[b]
  File "/share/workspace/cupy/cupyx/scipy/sparse/compressed.py", line 509, in __getitem__
    raise ValueError('unsupported indexing')
ValueError: unsupported indexing
```

Here's the same code using Scipy:

```
>>> import scipy
>>> 
>>> a = scipy.sparse.random(100, 100, format='csr')
>>> b = dask.array.from_array(a)
>>> b.compute()
array(<100x100 sparse matrix of type '<class 'numpy.float64'>'
	with 100 stored elements in Compressed Sparse Row format>, dtype=object)
```

When using a CuPy `coo_matrix`:

```
>>> a = cp.sparse.random(100, 100, format='coo')
>>> b = dask.array.from_array(a)
>>> b.compute()
  File "/share/workspace/dask/dask/base.py", line 165, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/share/workspace/dask/dask/base.py", line 436, in compute
    results = schedule(dsk, keys, **kwargs)
  File "/share/workspace/dask/dask/threaded.py", line 81, in get
    **kwargs
  File "/share/workspace/dask/dask/local.py", line 486, in get_async
    raise_exception(exc, tb)
  File "/share/workspace/dask/dask/local.py", line 316, in reraise
    raise exc
  File "/share/workspace/dask/dask/local.py", line 222, in execute_task
    result = _execute_task(task, data)
  File "/share/workspace/dask/dask/core.py", line 119, in _execute_task
    return func(*args2)
  File "/share/workspace/dask/dask/array/core.py", line 104, in getter
    c = a[b]
TypeError: 'coo_matrix' object is not subscriptable
```

As expected, the CuPy `csc_matrix` yields the same exception as CSR:

```
>>> a = cp.sparse.random(100, 100, format='csc')
>>> b = dask.array.from_array(a)
>>> b.compute()
  File "/share/workspace/dask/dask/base.py", line 165, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/share/workspace/dask/dask/base.py", line 436, in compute
    results = schedule(dsk, keys, **kwargs)
  File "/share/workspace/dask/dask/threaded.py", line 81, in get
    **kwargs
  File "/share/workspace/dask/dask/local.py", line 486, in get_async
    raise_exception(exc, tb)
  File "/share/workspace/dask/dask/local.py", line 316, in reraise
    raise exc
  File "/share/workspace/dask/dask/local.py", line 222, in execute_task
    result = _execute_task(task, data)
  File "/share/workspace/dask/dask/core.py", line 119, in _execute_task
    return func(*args2)
  File "/share/workspace/dask/dask/array/core.py", line 104, in getter
    c = a[b]
  File "/share/workspace/cupy/cupyx/scipy/sparse/compressed.py", line 509, in __getitem__
    raise ValueError('unsupported indexing')
ValueError: unsupported indexing
```

cjnolet:
This PR adds compressed sparse support for column slicing and fancy (array-based) indexing. 

Closes #2360 
Closes #3178  

Items left to todo:
- [x] Update documentation
- [x] Convert required C-based routines from Scipy to CuPy RawKernels
- [x] Plug into existing `__getitem__` in compressed.py
- [x] Add tests comparing scipy to cupy
- [x] Add tests for graceful failure conditions
- [x] Profiling

The bulk of the changes in this PR are in cupy.scipy.sparse.index and parallelize single-threaded scipy.sparse c code either over the indptr array (rows of CSR / cols of csc) or the output indices in the case of array-based indexing. Parallelism was made more granular where possible. 

Parallelizing over the indptr has the additional benefit that the cuda kernel code remains very similar to the scipy’s single-threaded c-code. However, it also creates warp divergence when the number of indices in each indptr entry are non-uniform. It also eliminates the ability for coalesced reads and writes. Still, I think this is a great starting point and it provides a lot of highly important features.


emcastillo:
Thanks for the awesome work on a long-standing one!

emcastillo:
It would be awesome if you could give a try to #3115

cjnolet:
@leofang @emcastillo 

For the minor slicing, it sounds like the most direct approach would be to construct the output index array in one pass, and then use a second pass to fill in the corresponding minor slices and values. So far CSR, we would do one pass to compute the output `indptr` array then map one thread per block to fill in each of the `indices` and `values`. 

CSC would be the same, except the column `inptr` would be constructed first and the row indices and values constructed in the second pass. 

The "fancy indexing" just seems like a single pass to populate the elements (one thread per index). This same kernel should also be able to take care of assignment, if I'm thinking about this correctly. 

Does this sound like a reasonable solution?

cjnolet:
@leofang @emcastillo, I think once the fancy indexing is implemented for `__getitem__`, it won't be too hard to implement setting. I'd love to get both in release 8.0.0. I'll see we are with the slicing/fancy indexing in July.

cjnolet:
I still need to add docs and some tests to validate graceful handling of failure conditions.

The current implementation supports `float32` and I'm not sure the best way to add support for `float64`. I have not been able to get templates working with `cupy.RawKernel` and I'd prefer not to duplicate all the kernel strings.

cjnolet:
Disregard my previous question about float64 support. I'm going to use `RawModule` to template the kernel.

leofang:
Yes, sounds like a good target for using `name_expressions` in `RawModule`.

cjnolet:
A quick grep seems to indicate that `cupyx/scipy/sparse/index.py` (introduced in this PR) is the only file in the codebase that creates instances of `RawModule` outside of the tests and I think this might be causing the failures in CI.

leofang:
Hi @cjnolet 

> A quick grep seems to indicate that `cupyx/scipy/sparse/index.py` (introduced in this PR) is the only file in the codebase that creates instances of `RawModule` outside of the tests and I think this might be causing the failures in CI.

I think #3534 will fix it!

cjnolet:
Thanks @leofang! It looks like your PR works because CI is passing here.

cjnolet:
I've done a little bit of profiling-
- Slicing and indexing with a single integer get up to an order of magnitude faster than Scipy as the data sizes increase. 
- Fancy indexing, on the other hand, seems to be hit or miss and the kernels created in this PR are not even in the list of kernels spending the most the time in compute. The calls to cub::min and cub::max seem to be spending in upwards of 25-33% of the total time in compute, dominating the other kernels. I'm guessing this is because of the atomics needed for the allreduce? Unfortunately, these 2 kernels are currently killing the performance of fancy indexing.

leofang:
I think you have the module `cupy.cuda.cub` built? How much time would min and max take if you set `cupy.cuda.cub_enabled = False`? CUB is known performant for large size reductions, but for your problem size I am less certain.

leofang:
Another question, @cjnolet, is that given your updated PR description, I take it that this PR introduces a code structure similar to that of SciPy? If so, I think it'd make reviewing (not saying I am qualified to do this, but I'd give it a shot!) much easier. I find that being able to compare the implementations of `scipy.sparse` and `cupyx.scipy.sparse` side by side is a great plus. I did this when preparing a few sparse-related PR to spot problems.

cjnolet:
@leofang, I've attached an `nvvp` session where I generate an array of 10k random integers and use it for a major index query of CSR & CSC matrices, all of size 10kx10k w/ 7% density. According to NVTX, CuPy consistently looks to be at least an order of magnitude faster than Scipy. An index of 100 elements looks to demonstrate some speedup depending on which path is used, but the speedup is more consistent starting around 1k.

[fancy_indexing_prof.nvvp.zip](https://github.com/cupy/cupy/files/4867510/fancy_indexing_prof.nvvp.zip)

leofang:
> I've attached an `nvvp` session where I generate an array of 10k random integers and use it for a major index query of CSR & CSC matrices, all of size 10kx10k w/ 7% density.

Thanks, @cjnolet! This is with `cupy.cuda.cub_enabled = False`, right? I see that now `cupy_max`/`cupy_min` took the majority of time (33%) instead of `cub::max`/`cub::min` (and this portion of time seems to be only slightly higher than CUB).

Just a thought. In `index.py` when you check the index bound, what if you sort first and then take the leftmost and rightmost values?

cjnolet:
@leofang, it was a long day yesterday and I realized right before I called it a night that the `min()`/`max()` calls taking 33% of the time in compute were from the creation of the test data (`cupy.sparse.random()`). Adjusting for that, it looks like the remaining time is spent doing memcopies between host and device and running the custom indexing kernels. 

When the number of columns is significantly high, I'm seeing CSC doing an order of magnitude better on GPU than CSR, even for major indexing of size=100. CSR is slightly slower than CPU in these cases (by a few ms on a 1kx100k matrix). On the same matrix, I'm seeing 2x+ speedup from CSR when the index size=1k while CSC is still a little over an order of magnitude better. 

When the number of rows is significantly high, CSC does worse than CPU by almost 1.7x while CSR is an order or magnitude or more better for an index size=100. Increasing the index size=1k, closes the gap quite a bit for CSC, however CPU is still faster by a few ms while CSR is still an order of magnitude or more faster.

When slice(None) is used for the major index and fancy indexing for the minor index, both CSR & CSC are running 2 orders of magnitude faster on the GPU for 100kx1k and 1 order of magnitude faster for CSR in the 1kx100k matrix with 2 orders of magnitude for the CSC. The complex types seem to consistently run 3 orders of magnitude faster for both matrix sizes. Also, the index sizes between 100 and 1k seem to have less an of an impact when fancy indexing is used for the minor index.

When we use fancy indexing for both major and minor indexes, the GPU is consistently faster by 2-3 orders of magnitude, both for index sizes of 100 and 1k. These results don't seem affected at all by whether the data is tall/skinny or short/wide. Overall, I'm really happy with these results.

cjnolet:
It looks like boolean indexing is just faster all around on GPU: 1-2 orders of magnitude for both tall skinny and short wide for major, minor, and major/minor.

cupyx/scipy/sparse/index.py
```diff
@@ -0,0 +1,854 @@
...
+           (n_row, col_offsets, Ap, Aj, Bp))
+
+
+def csr_column_index1(n_idx, col_idxs, n_row, n_col,
+                      indptr, indices, offsets, new_indptr):
+    bin_col_offsets(n_idx, col_idxs, offsets)
+    csr_column_index1_degree(n_row, offsets, indptr, indices, new_indptr)
+
+    cupy.cumsum(offsets, out=offsets)
+    cupy.cumsum(new_indptr, out=new_indptr)
```
asi1024, line 108: Could you write a comment and rearrange the arguments? I feel them a little complex to understand whether each parameter is updated or not. How about allocating the memory of `offsets` and `res_indptr` here as return values?
```suggestion
def csr_column_index1(n_idx, col_idxs, n_row, indptr, indices):
    offsets = cupy.zeros(n_col, indices.dtype)
    res_indptr = cupy.zeros_like(indptr)
    bin_col_offsets(n_idx, col_idxs, offsets)
    csr_column_index1_degree(n_row, offsets, indptr, indices, new_indptr)
    cupy.cumsum(offsets, out=offsets)
    cupy.cumsum(new_indptr, out=new_indptr)
    return offsets, new_indptr
```

cupyx/scipy/sparse/index.py
```diff
@@ -0,0 +1,854 @@
...
+
+def bin_col_offsets(n_idx, col_ids, col_offsets, tpb=32):
+    grid = math.ceil(n_idx / tpb)
+    kernel = bin_col_offsets_ker.get_function(
+        bin_col_offsets_ker_types[int32_dtype])
+    kernel((grid,), (tpb,), (n_idx, col_ids, col_offsets))
+
+
+csr_column_index1_ker_types = {int32_dtype: 'csr_column_index1_ker<int>'}
+csr_column_index1_ker = core.RawModule(code="""
```
asi1024, line 68: It looks that this kernel can be more parallelized by using advanced indexing and cumsum.

cjnolet: I’d like to try this but I’m having trouble thinking through how advanced indexing can be used efficiently here. Can you elaborate on your thoughts here?

cupyx/scipy/sparse/index.py
cjnolet: I updated all the functions in this file that are called by `compressed.py` to allocate an return, rather than accept empty arrays and modify them. Please let me know if you'd still like me to document these.

asi1024: Also `n_idx` can be removed because it is equal to `indices.size`.

cupyx/scipy/sparse/index.py
asi1024: For example, this kernel can be written as follows:
```py
def _csr_column_index1_degree(n_row, col_offsets, Ap, Aj, Bp):
    col_sum = cupy.zeros((Aj.size + 1,), dtype=col_offsets.dtype)
    cupy.cumsum(col_offsets[Aj], out=col_sum[1:])
    Bp = col_sum[Ap]
    Bp[1:] -= Bp[:-1]
    return Bp
```

cupyx/scipy/sparse/index.py
```diff
@@ -0,0 +1,865 @@
...
+        if(jj < n_idx) {
+            const I j = col_idxs[jj];
+            atomicAdd(col_offsets+j, 1);
+        }
+    }
+""", options=_module_options, name_expressions=tuple(
+    _bin_col_offsets_ker_types.values()))
+
+
+def _bin_col_offsets(n_idx, col_ids, col_offsets, tpb=32):
```
asi1024, line 60: It is equivalent to `cupyx.scatter_add(col_offsets, col_ids, 1)`.

cupyx/scipy/sparse/index.py
```diff
@@ -0,0 +1,865 @@
...
+    grid = math.ceil(n_idx / tpb)
+    kernel = _bin_col_offsets_ker.get_function(
+        _bin_col_offsets_ker_types[_int32_dtype])
+    kernel((grid,), (tpb,), (n_idx, col_ids, col_offsets))
+
+
+def _csr_column_index1_indptr(col_offsets, Ap, Aj):
```
asi1024, line 67: Perhaps this function can be replaced with `cupy.searchsorted`. How do you think?

leofang:
Oops sorry for my poor PR description in #3534...😅 @asi1024 Could you reopen this, please?

cupyx/scipy/sparse/index.py
cjnolet: I'm happy to give it a try, though I'm having trouble seeing it would be used to directly compute the output indptr. Can you share more detail on how this would look?

asi1024: I don't know the alternative implementation now, but anyway we should find a better algorithm to avoid `n_col` size array creation in L60 at least. `n_col` value tends to be huge.
https://github.com/cupy/cupy/pull/3486/files#diff-f723ef8036e0c67c8733d4eb7f7b6d87R60

asi1024: I think it's also OK to compute the result with the dense format and convert it to the sparse format. The number of nonzero values in the result cannot be found without host-device synchronization, and all values may nonzero in worst case.

cjnolet: In that worst case, a sparse format is going to add an additional array of size `n_cols` if we convert to a sparse format. I'm wondering if we could use the sparsity (number of unique col idxs vs n_col) to determine whether it's worthwhile to convert to sparse.

For example, I don't believe anything is lost with the current algorithm if the number of unique col_idxs is very close to n_cols. However, I agree that having it in dense form sounds l like a total waste if `col_offsets` is extremely sparse. Maybe we could choose a threshold and use the current code path if the ratio of unique col_idxs to n_col is something like 90% or more?

asi1024: I mean that we cannot create `new_data` smaller than the size of indices (for example, in sliceXarray indexing case, it is the product of the size of the slice and the size of array) without looking the actual data in the input arrays. Of course, we can create `new_data` smaller than `n_col` size if the size of indices is smaller than `n_col`.

cjnolet: Spent some time on this one today to see where I could get. I was able to build the `column_offsets` such that it only contains the counts for the unique (and sorted) elements in `col_idxs`. One option for populating the output indices and data arrays could be to filter the input indices up front for only the matching columns, change them to their indices in the `col_idxs`, and use the smaller `col_offsets` array to determine the repeating values. This is proving to be pretty challenging to do with only a couple operations though and would require at least 1 additional copy of the input indices array.

I'm also thinking we should consider parallelizing the updates over the `col_idx` array in when it is much larger than the number of rows and number of columns.

cjnolet: @asi1024 I tried several different ideas for shrinking the `col_offsets` array but they all seemed to result in having several copies of the column indices and/or the index array. The big challenge there is the need to support duplicate columns in the index. I'm at a loss. 

I've been trying to keep the performance to match the constant-time lookup that we get by keeping `col_offsets` dense. Alternatively, I suppose in the case where N is very very large, we could resort to a binary search in the `csr_column_index2` kernel. I'd also be happy to try any ideas you might have.

cjnolet: I managed to find a solution that at least passes all the tests. It still needs profiling and further optimization but thus far it requires the following additional memory:
1. A mask of the input indices array (size input nnz).
2. An array of size n_indices containing the occurrence frequencies for each column in the index
3. An array that maps the columns in the mask to the array of frequencies. 

The array in 3 is necessary because it's possible the index returned from `cupy.unique` is not monotonically increasing order. Next week I'll see if I can find a way to remove this.

cjnolet:
I captured a preliminary benchmark last night w/ the recent changes 
[preliminary_slicing_benchmark.txt](https://github.com/cupy/cupy/files/4906930/preliminary_slicing_benchmark.txt)

cjnolet:
@asi1024, this is ready for another look. I've documented the main functions in `index.py` for clarity even though they aren't intended to be user-facing.

cupyx/scipy/sparse/index.py
```diff
@@ -0,0 +1,1095 @@
...
+        out_order : new array where each element points to
+                    its next potential duplicate in the index
+    """
+    grid = math.ceil(col_order.size-1 / tpb)
+    kernel = _csr_column_index2_order_ker.get_function(
+        _csr_column_index2_order_types[col_order.dtype]
+    )
+    out_order = cupy.empty_like(col_order)
+    kernel((grid,), (tpb,), (col_order, col_order.size-1, out_order))
+
```
asi1024, line 84: To keep the maintainability, could you avoid defining custom kernels if not needed? I think most of the kernels in this file can be replaced with existing kernels.
```suggestion
    out_order = cupy.empty_like(col_order)
    out_order[col_order[:-1]] = col_order[1:]
```

cjnolet:
@asi1024, 

I managed to replace a few more kernels in `index.py`. There are 3 main custom kernels remaining, which I have converted to `ElementwiseKernel`:
- fancy row indexing
- fancy column indexing
- arrayXarray

I haven't found a way to completely remove these kernels because of the need to support duplicates in the given indexes. If you have any ideas, I'd be happy to try them.

pentschev:
@asi1024 do you think it would be possible to do a good thorough review on this PR so we can work on any necessary changes as quickly as possible to ensure it makes the cut into RC1?

leofang:
I am simply curious if the CI would complain. Let's take a look.
Jenkins, test this please

leofang:
@cjnolet I notice the test style in `test_csc.py`/`test_csr.py` is not in line with CuPy's style. Could you convert them?

cjnolet:
@leofang, does this look better?

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -274,6 +274,9 @@ def test_toarray(self):
 
     def test_pickle_roundtrip(self):
         s = _make(cupy, sparse, self.dtype)
+
+        print(str(pickle.dumps(s)))
+
```
leofang, line 279: Remove this.

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -1619,6 +1618,80 @@ def test_getitem_ellipsis(self, xp, sp):
...
+        X = cupy.random.random((N, N))
+        X[X > 0.7] = 0
+        Xcsr = cupy.sparse.csr_matrix(X)
+
+        for i in range(N):
+            arr_row = X[i:i + 1, :]
+            csr_row = Xcsr.getrow(i)
+
+            cupy.testing.assert_array_almost_equal(arr_row, csr_row.toarray())
+            numpy.testing.assert_(type(csr_row) is cupy.sparse.csr_matrix)
```
leofang, line 1678: I didn't test it, but this should work. 
```python
    def test_getrow(self):
        # This test is adapted from Scipy's CSR tests
        N = 10
        X = testing.shaped_random((N, N), cupy)
        X[X > 0.7] = 0
        Xcsr = sparse.csr_matrix(X)
 
        for i in range(N):
            arr_row = X[i:i+1, :]
            csr_row = Xcsr.getrow(i)
            assert sparse.isspmatrix_csr(csr_row)
            assert (arr_row == csr_row.toarray()).all()
```

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -1619,6 +1618,80 @@ def test_getitem_ellipsis(self, xp, sp):
...
+
+    def test_getcol(self):
```
leofang, line 1680: See the suggestion above.

tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
```diff
@@ -1469,9 +1469,79 @@ def test_getitem_slice_negative(self, xp, sp):
...
+    @testing.numpy_cupy_allclose(sp_name='sp')
+    def test_getitem_rowslice_negative_start_step(self, xp, sp):
+        # This test is adapted from Scipy's CSC tests
+        return _make(xp, sp, self.dtype)[slice(-2, 1, -2)]
+
+    def test_getitem_bool_indexing(self):
+        # This test is adapted from Scipy's CSC tests
+        sp_data = scipy.sparse.csc_matrix([[0, 1, 2], [3, 4, 5], [6, 7, 8]],
+                                          dtype=self.dtype)
+        data = cupy.sparse.csc_matrix(sp_data)
```
leofang, line 1496:
```suggestion
        data = sparse.csc_matrix(sp_data)
```

tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
```diff
@@ -1469,9 +1469,79 @@ def test_getitem_slice_negative(self, xp, sp):
...
+        slice_array1 = data[array_indices1].toarray()
+        slice_list2 = data[list_indices2]
+        slice_array2 = data[array_indices2]
+        slice_list3 = data[list_indices3]
+        slice_array3 = data[array_indices3]
+        assert (slice_list1 == slice_array1).all()
+        assert (slice_list2 == slice_array2).all()
+        assert (slice_list3 == slice_array3).all()
+
+    def test_getrow(self):
```
leofang, line 1515: See my suggestion in `test_csr.py`.

tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
```diff
@@ -1469,9 +1469,79 @@ def test_getitem_slice_negative(self, xp, sp):
...
+        Xcsc = cupy.sparse.csc_matrix(X)
+
+        for i in range(N):
+            arr_row = X[i:i + 1, :]
+            csc_row = Xcsc.getrow(i)
+
+            cupy.testing.assert_array_almost_equal(arr_row, csc_row.toarray())
+            numpy.testing.assert_(type(csc_row) is cupy.sparse.csr_matrix)
+
+    def test_getcol(self):
```
leofang, line 1531: See my suggestion in `test_csr.py`.

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -1619,6 +1618,80 @@ def test_getitem_ellipsis(self, xp, sp):
...
+    @testing.numpy_cupy_allclose(sp_name='sp')
+    def test_getitem_rowslice_negative_start_step(self, xp, sp):
+        # This test is adapted from Scipy
+        return _make(xp, sp, self.dtype)[slice(-2, 1, -2)]
+
+    def test_getitem_bool_indexing(self):
+        # This test is adapted from Scipy's CSR tests
+        sp_data = scipy.sparse.csr_matrix([[0, 1, 2], [3, 4, 5], [6, 7, 8]],
+                                          dtype=self.dtype)
+        data = cupy.sparse.csr_matrix(sp_data)
```
leofang, line 1645:
```suggestion
        data = sparse.csr_matrix(sp_data)
```

tests/cupyx_tests/scipy_tests/sparse_tests/test_index.py
```diff
@@ -0,0 +1,234 @@
...
+        else:
+            maj_h = maj
+
+        if isinstance(min, cupy.ndarray):
+            min_h = min.get()
+        else:
+            min_h = min
+
+        if min is not None:
+
```
leofang, line 40: Remove empty line
```suggestion
```

tests/cupyx_tests/scipy_tests/sparse_tests/test_index.py
```diff
@@ -0,0 +1,234 @@
...
+            expected.sort_indices()
+
+            cupy.testing.assert_array_equal(
+                actual.indptr, expected.indptr)
+            cupy.testing.assert_array_equal(
+                actual.indices, expected.indices)
+            cupy.testing.assert_array_equal(
+                actual.data, expected.data)
+        else:
+
```
leofang, line 64:
```suggestion
```

tests/cupyx_tests/scipy_tests/sparse_tests/test_index.py
```diff
@@ -0,0 +1,234 @@
...
+@testing.parameterize(*testing.product({
+    'format': ['csr', 'csc'],
+    'density': [0.1, 0.4, 0.9],
+    'dtype': ['float32', 'float64', 'complex64', 'complex128'],
+    'n_rows': [25, 150],
+    'n_cols': [25, 150]
+}))
+@testing.with_requires('scipy')
+@testing.gpu
+class TestIndexing(unittest.TestCase):
```
leofang, line 18: Since this `_run()` function spreads multiple lines, I'd just comment instead of writing suggestions 😅 Generally, `cupy.testing` will do the device-to-host transfer for you, so you don't have to call `get()` or do sync yourselves.

cjnolet:
Thank you @leofang, for the review! I've made the requested changes. I'm certainly grateful if you are interested in giving the implementation a look as well. 

@asi1024 @emcastillo any chance you have had the opportunity to take a look at this PR?

asi1024:
@cjnolet This PR adds several types of slicing/indexing, but we cannot merge this PR until all reviews of them are completed. How about splitting this PR into several PRs?

leofang:
I'll try to do another pass in the next few days.

cjnolet:
@asi1024, I have opened the following 3 PRs in addition to this one: #3657, #3658, #3639. 

#3657 Includes the main scaffolding ported from Scipy but raises a NotImplementedError for all features not related to slicing. 
#3658 Includes the changes from the previous PR but includes sparse row indexing
#3659 Includes the changes from #3658, but includes sparse column indexing. 

This PR includes all the changes from #3659 and adds inner indexing (`arrayXarray`). 

I built the new PRs incrementally in this way in hopes the deltas could be more easily reviewed. I will keep rolling the changes backwards into this PR so that they can be merged one after the other. 

Will this approach work for you?

asi1024:
@cjnolet Thanks!

jakirkham:
pfnCI, test this please

cjnolet:
@asi1024, this is the last of the sparse `__getitem__` PRs! (pending follow-on improvements/optimizations).  I just merged in `master` and added the {cupy|numpy}.ndarray variations to the tests. This is ready for review!

pentschev:
pfnCI, test this please.

tests/cupyx_tests/scipy_tests/sparse_tests/test_index.py
```diff
@@ -71,7 +71,7 @@ def _run(self, maj, min=None, flip_for_csc=True,
                 actual.data, expected.data)
         else:
             testing.assert_array_equal(
-                actual, expected)
```
asi1024, line 74: Why is this change needed?

cjnolet: Scipy returns a `numpy.Matrix` object when inner indexing is used. I didn't port that object over to Cupy so we just return an `ndarray`. It took an extra conversion to compare the two directly.

cjnolet: `numpy.matrix` also looks deprecated: https://numpy.org/doc/stable/reference/generated/numpy.matrix.html

asi1024: I see. It seems OK to convert to ndarray with `asarray`.
Why do we need to compare after converting to 1d-array?

cjnolet: I have not figured out how to get output from inner indexing that are not 1d arrays and I wasn't able to find much for examples in the Scipy tests. I've expanded the dims for the Cupy output to make a more direct comparison. What do you think? Would you rather we do a `numpy.asarray(expected).squeeze()` or call `cupy.expand_dims(val, 0)` before returning the result?

jakirkham:
pfnCI, test this please

asi1024:
LGTM.

PR #3488, opened 2020-06-24
Add support for cusolverDn<t>syevj and cusolverDn<t>syevjBatched

dmargala:
This PR adds support for the cuSolver eigenvalue decomposition via Jacobi method and the batched version as well:

 * syevj: https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-syevj 
 * syevjBatched: https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-syevjbatch

Note from the CUDA documentation:

> syevj has the same functionality as syevd. The difference is that syevd uses QR algorithm and syevj uses Jacobi method. The parallelism of Jacobi method gives GPU better performance on small and medium size matrices.


kmaehashi:
cc/ @toslunar

takagi:
pfnCI, test this please.

takagi:
Would you update the following part of `eigh`'s docstring?
```
    .. note::

       Currently only 2-D matrix is supported.
```

cupy/cusolver.py
```diff
@@ -254,3 +256,203 @@ def gesvda(a, compute_uv=True):
...
+
+    Args:
+        a (cupy.ndarray): A symmetric 2-D square matrix ``(M, M)`` or a batch
+            of symmetric 2-D square matrices ``(N, M, M)``.
+        UPLO (str): Select from ``'L'`` or ``'U'``. It specifies which
+            part of ``a`` is used. ``'L'`` uses the lower triangular part of
+            ``a``, and ``'U'`` uses the upper triangular part of ``a``.
+        with_eigen_vector (bool): Indicates whether or not eigenvectors
+            are computed.
+    Returns:
```
takagi, line 275:
```suggestion

    Returns:
```

cupy/cusolver.py
```diff
@@ -254,3 +256,203 @@ def gesvda(a, compute_uv=True):
...
+        batch_size)
+    cupy.linalg.util._check_cusolver_dev_info_if_synchronization_allowed(
+        syevjBatched, dev_info)
+
+    cusolver.destroyGesvdjInfo(params)
+
+    w = w.astype(ret_w_dtype, copy=False).swapaxes(-2, -1)
+    v = v.astype(ret_v_dtype, copy=False).swapaxes(-2, -1)
+
+    return w, v
```
takagi, line 458: How about returning only w if with_eigen_vector=False?

dmargala:
@takagi thanks for the suggestions! Let me know if there's anything else I can do.

cupy/linalg/eigenvalue.py
```diff
@@ -121,7 +121,10 @@ def eigh(a, UPLO='L'):
 
     .. seealso:: :func:`numpy.linalg.eigh`
     """
-    return _syevd(a, UPLO, True)
+    if a.ndim == 3:
+        return cupy.cusolver._syevj_batched(a, UPLO, True)
```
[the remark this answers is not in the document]
dmargala: There is an issue with removing the `cupy.` prefix here since `_syevj_batched` is implemented in `cupy.cusolver` and the `cusolver` reference in this file is to `cupy.cuda.cusolver`. Do you have a suggestion for resolving this issue?

cupy/linalg/eigenvalue.py
```diff
@@ -121,7 +120,10 @@ def eigh(a, UPLO='L'):
 
     .. seealso:: :func:`numpy.linalg.eigh`
     """
-    return _syevd(a, UPLO, True)
+    if a.ndim == 3:
+        return cusolver._syevj_batched(a, UPLO, True)
```
yoshipon, line 124: Thank you for your PR.
Could you change `eigh` to support arbitrary `a.ndim > 2` cases like as follows?:
```suggestion
    if a.ndim > 2:
        *batch_shape, m, n = a.shape
        w, v = cusolver._syevj_batched(a.reshape(-1, m, n), UPLO, True)
        return w.reshape(*batch_shape, m), v.reshape(a.shape)
```

cupy/linalg/eigenvalue.py
```diff
@@ -156,4 +156,7 @@ def eigvalsh(a, UPLO='L'):
 
     .. seealso:: :func:`numpy.linalg.eigvalsh`
     """
-    return _syevd(a, UPLO, False)[0]
+    if a.ndim == 3:
+        return cusolver._syevj_batched(a, UPLO, False)[0]
```
yoshipon, line 160: This part could also be:
```suggestion
    if a.ndim > 2:
        *batch_shape, m, n = a.shape
        w, _ = cusolver._syevj_batched(a.reshape(-1, m, n), UPLO, False)
        return w.reshape(*batch_shape, m)
```

cupy/linalg/eigenvalue.py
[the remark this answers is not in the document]
takagi: Ah, okay. We may use an alias to import, but I think your original lines are enough.

takagi:
pfnCI, test this please.

cupy/linalg/eigenvalue.py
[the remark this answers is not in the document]
dmargala: Okay, I committed this suggestion before realizing the issue. I will push a fix shortly.

takagi:
@dmargala How do you think about @yoshipon 's above comments to generalize the batch support for inputs with more than three dimensions? I think it's okay to merge this PR as it is and address the generalization as another PR.

takagi:
pfnCI, test this please.

yoshipon:
Yeah, I'll make another PR if @dmargala prefers.

dmargala:
I added support for arbitrary batch dimensions as @yoshipon requested.

It looks like there are some merge conflicts now. Is there a preference for using rebase vs merge in order to resolve those?

yoshipon:
@dmargala Thank you very much! Your updates are much better than my suggestions :)

cupy/cusolver.py
```diff
@@ -254,3 +256,213 @@ def gesvda(a, compute_uv=True):
...
+
+def syevj(a, UPLO='L', with_eigen_vector=True):
+    """Eigenvalue decomposition of symmetric matrix using cusolverDn<t>syevj().
+
+    Computes eigenvalues ``w`` and (optionally) eigenvectors ``v`` of a complex
+    Hermitian or a real symmetric matrix.
+
+    Args:
+        a (cupy.ndarray): A symmetric 2-D square matrix ``(M, M)`` or a batch
+            of symmetric 2-D square matrices ``(..., M, M)``.
```
toslunar, line 269: Could you add the support for `M == 0` or mention that `M >= 1` is required?  NumPy supports empty matrices while LAPACK interface does not.  Try to be consistent in `cupy/cusolver.py`.

For `M == 0`, the line `a = a.reshape(-1, m, lda)` has to be amended, too.

dmargala: That's a good point. It looks like `numpy.linalg.eigh` indirectly requires `M >= 1` by requiring that the last two dimensions of `a` to be equal. I don't know if it's possible to create a square two dimensional array with shape `(M, M)` where `M == 0` (the closest I can think of is `numpy.array([[]])` which has shape `(1, 0)`).

```python
In [1]: import numpy as np

In [2]: np.linalg.eigh([])

    205         if a.ndim < 2:
    206             raise LinAlgError('%d-dimensional array given. Array must be '
--> 207                     'at least two-dimensional' % a.ndim)
    208

LinAlgError: 1-dimensional array given. Array must be at least two-dimensional


In [3]: np.linalg.eigh([[]])

    211         m, n = a.shape[-2:]
    212         if m != n:
--> 213             raise LinAlgError('Last 2 dimensions of the array must be square')
    214

LinAlgError: Last 2 dimensions of the array must be square
```

I can add similar checks to `cupy.linalg.eigh` and `cupy.linalg.eigvalsh`.

Does that sound good?

yoshipon: We can make such an ndarray by:
```python
In [4]: np.linalg.eigh(np.zeros([0, 0]))                                                                                                                                                                           
Out[4]: (array([], dtype=float64), array([], shape=(0, 0), dtype=float64))                                                    
```

FYI.

dmargala: Ah, thanks! I didn't know that :)

dmargala: Okay, I pushed a fix for the offending line that was mentioned. Unfortunately, I will not have access to a GPU system to test locally until next Wednesday.

takagi:
As for rebasing or merging, we don't have a preference.

takagi:
I'll kick the CI after you resolve the conflict.

takagi:
pfnCI, test this please.

PR #3500, opened 2020-06-25
ndimage rank-based filters

coderforlife:
This implements the rank-based filters (median, rank, and percentile filters).

These all use the same logic (in `_rank_filter`) and utilize the same underlying CUDA kernel as the other ndimage filters
just with some adjustments that the `_generate_nd_kernel()` function already allowed for. The biggest piece of CUDA code is the sorting function required for these filters. It chooses either selection or shell sort to be used to establish the rankings. CUB or similar sorts are not used since the sort is done per-kernel instead of with dynamic parallelism (there is 1 small sort to be done for every pixel in an image, should not parallelize within a single sort but instead parallelize across the sorts).

The cutoff of 225 (15x15), for which selection sort is used below and shell sort is used above, was chosen based on my hardware (Titan V) using uint8 images and definitely made a large different in speed by being able to use both. Selection sort is up to 2.5x faster below and shell sort is progressively faster above, tested up to 3025-sized (55x55) footprint and it was 9x faster then. Near the cutoff point it matters much less which is chosen since they are fairly close. Thus, its okay if that number is a little off.

### DIFF
--- a/cupyx/scipy/ndimage/__init__.py
+++ b/cupyx/scipy/ndimage/__init__.py
@@ -6,6 +6,9 @@ from cupyx.scipy.ndimage.filters import minimum_filter  # NOQA
 from cupyx.scipy.ndimage.filters import maximum_filter  # NOQA
 from cupyx.scipy.ndimage.filters import minimum_filter1d  # NOQA
 from cupyx.scipy.ndimage.filters import maximum_filter1d  # NOQA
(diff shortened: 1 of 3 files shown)


coderforlife:
This is working towards #3111.

emcastillo:
The PR looks good to me, I will merge it and send a follow-up on some minor cosmetic issues so we don't keep feature PRs stalled

emcastillo:
Jenkins, test this please

Issue #3408, opened 2020-06-08
Support Jitify ?


leofang:
This is more for my own little curiosity than a feature request, but I thought it's worth discussing.

It is not possible to compile Thrust/CUB code in NVRTC because many system headers are missing, which was why @emcastillo struggled in #2629 and we still have to keep a copy of our own Thrust complex headers. [Jitify](https://github.com/NVIDIA/jitify) claims to provide an easy solution to address this.

I have been looking into jitify's implementation lately, and I think the key idea is to:
1. Recursively find out what header (and its absolute path) should be included (by repeating compiling -> inspecting error log)
2. Fix the headers: for example, provide some stubs to patch a few C++ `std` functions or system headers so that they can work as `__device__` functions, remove static variables, fix `pragma`, etc

Since we have most of the toolchain in place (compile/launch/CUDA API/etc), especially #3319 is now merged, I think it is not super difficult to address this. I can think of two design choices:
1. Wrap jitify in `cupy.cuda.jitify`: 
Ideally, we still use our NVRTC wrapper to do compilation, and only take all these lookup implementations as-is so we don't reinvent the wheel. Unfortunately it is currently not possible, as many logics are baked in jitify's public APIs, and it requires some refactoring on the their side for us to reuse.
2. Implement the compile-error search loop by ourselves in `cupy/cuda/compiler.py`: 
This would require kinda reimplementing what's done in jitify, but only for a small part (I hope). The pro is jitify doesn't need to be a dependency and we just reuse what's already available (very important for code maintenance), and the cons are that it takes more effort and that it needs to copy a significant amount of code from jitify (mainly those code patches).

I think a few RAPIDS libraries like cuDF use jitify. I will dig into how they use it...

emcastillo:
cub complex headers are pretty much stable, as they are just a struct and a couple of functions to manipulate the members that are very unlikely to change. I am not sure if doing all this is better than just keep the headers the way they are.

leofang:
> cub complex headers are pretty much stable, as they are just a struct and a couple of functions to manipulate the members that are very unlikely to change. I am not sure if doing all this is better than just keep the headers the way they are.

[...]

PR #3505, opened 2020-06-29
ndimage common linear filters

coderforlife:
This adds implementations for 11 linear filters to `cupyx.scipy.ndimage.filters`. The following are nearly identical to scipy's code, just using the cupy's `convolve`/`convolve1d`/`correlate`/`correlate1d` as appropriate:

- `uniform_filter1d`
- `gaussian_filter1d`
- `generic_laplace`
- `laplace`
- `gaussian_laplace`
- `generic_gradient_magnitude`
- `gaussian_gradient_magnitude`

The following use an additional function making it easier to add any separable linear filters, but otherwise perform very closely to the original scipy's functions:

- `uniform_filter`
- `gaussian_filter`
- `prewitt`
- `sobel`

The function for handle separable filters also now handles the min/max filters when separable (in fact, the code was taken from there to make it work for any set of 1D filters to create an ND filter).

This code also introduces a new piece to the CUDA code: a templated `cast` function to add automatic casting from floats to unsigned ints in a compatible way to scipy. See https://github.com/cupy/cupy/issues/3111#issuecomment-650183301 for details discussing their implementation.

Some outstanding issues with the current code:

- Does not pass all tests. Apparently the min/max 1D filters utilize different casting behaviors than all of the other filters. I think those are the only tests failing (24 out of 4786 tests) and only in very specific cases.
- As suggested in https://github.com/cupy/cupy/issues/3111#issuecomment-650196144 we might want to add a note to the documentation that identical outputs cannot be guaranteed when the output is of integral data type due to floating point rounding issues since all intermediate results are floating-points.
- @emcastillo has suggested there are some minor edits from the previous PR that should instead be integrated here

### DIFF
--- a/cupyx/scipy/ndimage/filters.py
+++ b/cupyx/scipy/ndimage/filters.py
@@ -26,9 +26,10 @@ def correlate(input, weights, output=None, mode='reflect', cval=0.0, origin=0):
         cupy.ndarray: The result of correlate.
 
     .. seealso:: :func:`scipy.ndimage.correlate`
-    .. note:: When the output data type is integral (or when no output is
-    provided and the input is integral) the results may not perfectly match the
-    results from SciPy due to floating-point rounding of intermediate results.
+    .. note::
+        When the output data type is integral (or when no output is provided
+        and input is integral) the results may not perfectly match the results
+        from SciPy due to floating-point rounding of intermediate results.
     """
     return _correlate_or_convolve(input, weights, output, mode, cval, origin)
 
@@ -58,9 +59,10 @@ def convolve(input, weights, output=None, mode='reflect', cval=0.0, origin=0):
         cupy.ndarray: The result of convolution.
 
     .. seealso:: :func:`scipy.ndimage.convolve`
-    .. note:: When the output data type is integral (or when no output is
-    provided and the input is integral) the results may not perfectly match the
-    results from SciPy due to floating-point rounding of intermediate results.
+    .. note::
+        When the output data type is integral (or when no output is provided
+        and input is integral) the results may not perfectly match the results
+        from SciPy due to floating-point rounding of intermediate results.
     """
     return _correlate_or_convolve(input, weights, output, mode, cval, origin,
                                   True)
@@ -89,9 +91,10 @@ def correlate1d(input, weights, axis=-1, output=None, mode="reflect", cval=0.0,
     Returns:
         cupy.ndarray: The result of the 1D correlation.
     .. seealso:: :func:`scipy.ndimage.correlate1d`
-    .. note:: When the output data type is integral (or when no output is
-    provided and the input is integral) the results may not perfectly match the
-    results from SciPy due to floating-point rounding of intermediate results.
+    .. note::
+        When the output data type is integral (or when no output is provided
+        and input is integral) the results may not perfectly match the results
+        from SciPy due to floating-point rounding of intermediate results.
     """
     weights, origins = _convert_1d_args(input.ndim, weights, origin, axis)
     return _correlate_or_convolve(input, weights, output, mode, cval, origins)
@@ -120,9 +123,10 @@ def convolve1d(input, weights, axis=-1, output=None, mode="reflect", cval=0.0,
     Returns:
         cupy.ndarray: The result of the 1D convolution.
     .. seealso:: :func:`scipy.ndimage.convolve1d`
-    .. note:: When the output data type is integral (or when no output is
-    provided and the input is integral) the results may not perfectly match the
-    results from SciPy due to floating-point rounding of intermediate results.
+    .. note::
+        When the output data type is integral (or when no output is provided
+        and input is integral) the results may not perfectly match the results
+        from SciPy due to floating-point rounding of intermediate results.
     """
     weights = weights[::-1]
     origin = -origin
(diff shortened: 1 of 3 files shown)


cupyx/scipy/ndimage/filters.py
```diff
@@ -720,8 +1133,52 @@ def _generate_boundary_condition_ops(mode, ix, xsize):
     return ops
 
 
+_CAST_FUNCTION = """
+// Implements a casting function to make it compatible with scipy
+// Use like cast<to_type>(value)
+// It's actually really simple - most of this is <type_traits>
```
leofang, line 1139: @emcastillo What is being added here is precisely a kind of stubs that jitify would add (#3408), see:
https://github.com/NVIDIA/jitify/blob/2486be5500e06676dd3b2ec8a52dd1a529998fe3/jitify.hpp#L1651

@coderforlife I wonder how did you figure out the minimum stubs needed here to make it work? By trial and error?

grlee77: This looks like a nice, modern C++ approach! It avoids adding an additional argument indicating unsigned status to `_get_correlate_kernel` as would have been required for my simple suggestion in https://github.com/cupy/cupy/issues/3111#issuecomment-650183301

coderforlife: @leofang I first wrote it in regular C++ using a regular C++ compiler, making sure it worked, while importing the regular `<type_traits>`. I was quite surprised that `<type_traits>` was not available with NVRTC. I found that CUDA includes this file (in `<cuda/std/type_traits>` which mostly just includes `"cuda/std/detail/libcxx/include/type_traits"`). However, all of those functions would need `__device__` added to them to be usable in NVRTC.

So, I worked backwards. I knew exactly what I needed and looked for them in the CUDA `type_traits` file, combining them with the possible implementations listed on cppreference.com (such as [this](https://en.cppreference.com/w/cpp/types/is_signed#Possible_implementation)) as I thought was appropriate. I did that until I had everything I needed and nothing extra.

Some of them do not quite meet the official standards since I knew I was allowed to cut corners, like knowing every type would always be arithmetic, so skipped that check in `is_signed`.

coderforlife: @grlee77 Thanks! I was much happier when I could actually include the C++ header `<type_traits>`... although that might be solved by what leofang is talking about.

coderforlife:
I have completed making the `minimum_filter1d` and `maximum_filter1d` functions consistent with scipy when there is an out-of-bounds `cval` with `mode='constant'`.

@grlee77 Should all functions have the extra comment about accuracy added to them? How does this sound as a comment?

> When the `output` data type is integral (or when no `output` is provided and the `input` is integral) the results may not perfectly match the results from SciPy due to floating-point rounding of intermediate results.

As soon as I add those in I will remove the draft status.

coderforlife:
@grlee77 Can you confirm that statement as good? Which functions (or all) should it be added to?

grlee77:
> @grlee77 Can you confirm that statement as good? Which functions (or all) should it be added to?

That statement looks good to me. I think it could be within the `Notes` section for all of the convolution-based filters, but that is a decent amount of duplication, so you may want to wait and see what the core CuPy team prefers.

We could alternatively (or additionally) add a section for ndimage differences to the following page:
https://docs-cupy.chainer.org/en/stable/reference/difference.html

emcastillo:
> Apparently the min/max 1D filters utilize different casting behaviors than all of the other filters. I think those are the only tests failing (24 out of 4786 tests) and only in very specific cases

If you can detect those, It is fine to skip them, or raise a warning in the routine.

coderforlife:
@emcastillo Most recent commit fixes that.

coderforlife:
@emcastillo Are there are things that need to be changed? In the previous commit you mentioned that there were other small things to fix. Also, where should the comment go about accuracy?

emcastillo:
Nope, it's just minor fixes to the docstrings, I will do it once it is merged :). So you don't have to lose time with this minor stuff.

For the accuracy a `.. note::` or `.. warning::` in the docstring should be enough for sphinx to reflect it.

https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

coderforlife:
One note about this is that there are now only 2 functions missing from `scipy.ndimage.filters`: `generic_filter` and `generic_filter1d`. I have a working version of `generic_filter` that can accept `ReductionKernel` objects and old-style fused functions the represent `ReductionKernel`s (and if given a callable tries to fuse it). However, the new-style fused functions introduced `FusedKernel` and I have not made that one work yet. Once I get `FusedKernel` working, it will hopefully be easy to get `generic_filter1d` working (however, oddly, that 1d function is not the same as all of the other 1d filters so there may be some work there). In my tests, the cupy generic filter is 2-3 orders of magnitudes faster than scipy with Python functions and 1 order of magnitude faster than when using scipy with numba functions.

emcastillo:
@coderforlife is this ready to be reviewed & merged? or is there anything still left :).

Thanks!

coderforlife:
Nothing more for this PR. Next PR will be more complex.

emcastillo:
Jenkins, test this please

coderforlife:
I don't understand. It says cupy-py3-cutensor,mn1-p100 failed, but if I go into it it says not tests failed.

emcastillo:
Tests are fine, it was the coverage report that got a connection error.

emcastillo:
Jenkins, test this please

leofang:
> I don't understand. It says cupy-py3-cutensor,mn1-p100 failed, but if I go into it it says not tests failed.

@coderforlife After you click a CI like `cupy-py3-cutensor`, you can click "Console Output" on the left to see the full log.

coderforlife:
@emcastillo A few questions about style for my next commit:

* The ndimage/filters.py is getting very long (1350 lines currently). To add `generic_filter` it is another 500 lines. Should we start breaking that file up? Maybe all of the public and simple private functions stay in filters.py while the other code moves to like _filters_internal.py?
* Where should I put small ideas on possible future improvements? For example, 1D filters can be sped up by using the trick in scipy to detect symmetry. However, that needs a whole brand-new CUDA kernel I think. Another idea was to use a shared variable for the weights. I tested this (see https://github.com/cupy/cupy/issues/3111#issuecomment-598548602) but I don't know how to predict when it will be helpful and when it won't be.

Thanks.

PR #3508, opened 2020-06-29
Implement `flatiter.__iter__()`

takagi:
This PR implements `flatiter.__iter__()` special method.

### DIFF
--- a/tests/cupy_tests/indexing_tests/test_iterate.py
+++ b/tests/cupy_tests/indexing_tests/test_iterate.py
@@ -15,6 +15,10 @@ class TestFlatiter(unittest.TestCase):
         a = cupy.zeros((2, 3, 4))
         assert a.flat.base is a
 
+    def test_iter(self):
+        it = cupy.zeros((2, 3, 4)).flat
+        assert it.__iter__() is it
+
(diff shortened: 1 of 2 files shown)


tests/cupy_tests/indexing_tests/test_iterate.py
```diff
@@ -15,6 +15,10 @@ def test_base(self):
         a = cupy.zeros((2, 3, 4))
         assert a.flat.base is a
 
+    def test_iter(self):
+        it = cupy.zeros((2, 3, 4)).flat
+        assert it.__iter__() is it
```
asi1024, line 20: How about writing as follows to check the compatibility to NumPy?
```py
for xp in (numpy, cupy):
    it = xp.zeros((2, 3, 4)).flat
    assert iter(it) is it
```
I think the other tests should be written as this format too, but It's OK to fix in another PR.

takagi: I've fixed it as such and, okay, I will change the other tests too!

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #3509, opened 2020-06-29
Implement `has_sorted_indices`, `has_canonical_format`, `sort(ed)_indices()` for sparse matrices

leofang:
This PR brings the current implementation to full compliance with SciPy, but I don't know what's the best way to measure performance impacts...I measured SpMV and do not see slowdown with this PR. Not sure what else I should look into, but I think for compatibility this PR is good to go.

Notes:
1. The property code is largely borrowed from SciPy so that it's easier to maintain and stay compliant.
2. The two `has_*` properties are mainly for compressed matrices (i.e. CSC/CSR); for COO, it doesn't have `has_sorted_indices`, and its `has_canonical_format` is set during creation or when `sum_duplicate()` is called. 
Importantly, COO does not validate this property when it is accessed. 
3. The internal flags `_has_*` are not directly accessed anywhere in the codebase; they are maintained by the corresponding properties.  
4. Whenever an operation does not guarantee the status of these properties, they should not be overwritten manually. This is in line with SciPy.

TODO:
- [x] Expand PR description
- [x] Add tests?
- [x] Clean up
- [ ] ~~Measure performance changes (how?)~~

### DIFF
--- a/cupyx/scipy/sparse/compressed.py
+++ b/cupyx/scipy/sparse/compressed.py
@@ -681,8 +681,11 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
     def sorted_indices(self):
         """Return a copy of this matrix with sorted indices
 
-        Taken from SciPy as is.
+        .. warning::
+            Calling this function might synchronize the device.
+
         """
+        # Taken from SciPy as is.
         A = self.copy()
         A.sort_indices()
         return A
@@ -693,6 +696,15 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
         raise NotImplementedError
 
     def sum_duplicates(self):
+        """Eliminate duplicate matrix entries by adding them together.
+
+        .. note::
+            This is an *in place* operation.
+
+        .. warning::
+            Calling this function might synchronize the device.
+
+        """
         if self.has_canonical_format:
             return
         # TODO(leofang): add a kernel for compressed sparse matrices without
--- a/cupyx/scipy/sparse/compressed.py
+++ b/cupyx/scipy/sparse/compressed.py
@@ -311,7 +311,6 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
             if shape is None:
                 shape = arg1.shape
 
-            has_canonical_format = x.has_canonical_format
         elif util.isshape(arg1):
             m, n = arg1
             m, n = int(m), int(n)
@@ -321,7 +320,6 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
             # shape and copy argument is ignored
             shape = (m, n)
             copy = False
-            has_canonical_format = True
 
         elif scipy_available and scipy.sparse.issparse(arg1):
             # Convert scipy.sparse to cupyx.scipy.sparse
@@ -333,7 +331,6 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
 
             if shape is None:
                 shape = arg1.shape
-            has_canonical_format = x.has_canonical_format
 
         elif isinstance(arg1, tuple) and len(arg1) == 3:
             data, indices, indptr = arg1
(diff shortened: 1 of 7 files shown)


leofang:
Done. PTAL.

takagi:
pfnCI, test this please.

cupyx/scipy/sparse/compressed.py
```diff
@@ -558,10 +582,71 @@ def _get_major_slice(self, major):
...
+        if self.data.size == 0:
+            self._has_canonical_format = True
+        # check to see if result was cached
+        elif not getattr(self, '_has_sorted_indices', True):
+            # not sorted => not canonical
+            self._has_canonical_format = False
+        elif not hasattr(self, '_has_canonical_format'):
+            is_canonical = self._has_canonical_format_kern(
+                self.indptr, self.indices, size=self.indptr.size-1)
+            self._has_canonical_format = bool(is_canonical.all())
```
takagi, line 611: This line synchronizes device to host. Please let me make sure that you think this sync is acceptable?

leofang: Yes, that's right. It's necessary because internally we need this flag on the host to do control flow. But, the cost is most likely one-time thing (the first time this flag is encountered).

leofang:
@cjnolet If you have time, could you evaluate if and how this PR would have performance impact?

cjnolet:
@leofang, 

Sure. I think I had already copied the implementation from Scipy into #3533 just to get something working but I can merge yours into that branch next week and do a little profiling.

cupyx/scipy/sparse/compressed.py
takagi: Thanks. I think it's fine, then would you document about the synchronization as https://github.com/cupy/cupy/pull/2798?

leofang: Ah, yes, thanks for reminder!

leofang: @takagi I noticed that even though these attributes have docstrings, in actual usages like creating a `csr_matrix` and then calling `help(b.has_canonical_format`, for example, SciPy does not display it properly. I think it's because when subclassing the `_cs_matrix` (in SciPy's case), the docstrings should also be explicitly copied as follows:
```python
has_canonical_format.__doc__ = _cs_matrix.has_canonical_format.__doc__`
```
but I suspect they didn't do it because executing this line might trigger the actual computation. Should I simply document this in `docs/source/reference/sparse.rst ` instead?

leofang: I think for the purpose of getting it documented on the website, Sphinx has done the job correctly. So I modified the getter docstrings. Let's see if it's OK.

leofang: Yup: https://cupy--3509.org.readthedocs.build/en/3509/reference/generated/cupyx.scipy.sparse.csr_matrix.html#cupyx.scipy.sparse.csr_matrix.has_canonical_format

takagi: Thanks! It's completely okay.

takagi:
pfnCI, test this please.

cupyx/scipy/sparse/compressed.py
takagi: Can I ask you to fix the docstrings of `sum_duplicates()` and `sort(ed)_indices()` methods, which use the properties and might also synchronize the device after this PR is merged?

leofang: Definitely!

takagi:
pfnCI, test this please.

takagi:
LGTM! I'm going to merge this PR, and please post a new issue if it has a significant performance impact.

leofang:
Thanks!

PR #2741, opened 2019-12-02
Update thrust::complex headers with a bug fix

leofang:
Closes #2629. Changes (see the discussion starting https://github.com/cupy/cupy/issues/2629#issuecomment-558448086):
1. Add a `__host__` qualifier to all functions
2. Sync a few function implementations with the upstream counterparts
3. Adjust some style and format
4. Fix a bug: https://github.com/cupy/cupy/pull/2741#discussion_r353360865

Issue #2629, opened 2019-11-11
Sync the headers in `cupy/core/include/cupy/complex/` with upstream?


leofang:
Recently the bundled Thrust complex headers have been modified on a per-PR basis for the following reasons:
1. Performance impact: #2562 (fixed alignment)
2. Missing `__host__` qualifiers: #2538
3. Missing support: #2520 

All of these changes are more or less in sync with the latest Thrust codes. I suggest that these headers can be brought up to date with the upstream. The incentive is that one might discover another one-liner change like in #2562 that boosts the performance considerably 🙂

emcastillo:
All these changes are already present in the current thrust?

leofang:
> All these changes are already present in the current thrust?

Yes 😄 For example:

>> 1. Performance impact: #2562 (fixed alignment)

The alignment is done via a specialized data holder `complex_storage`:
https://github.com/thrust/thrust/blob/324243f6bb70687aeaeb2419193a335648c5869d/thrust/complex.h#L454

>> 2. Missing `__host__` qualifiers: #2538

All Thrust complex functions have a `__host__` qualifier, as it is meant to be interchangeable with `std::complex` on the host:
https://github.com/thrust/thrust/blob/324243f6bb70687aeaeb2419193a335648c5869d/thrust/complex.h#L140-L142

>> 3. Missing support: #2520

`real()` and `imag()` are now supported in Thrust:
https://github.com/thrust/thrust/blob/324243f6bb70687aeaeb2419193a335648c5869d/thrust/complex.h#L384-L407

Now, my question: Since Thrust comes with the CUDA toolkit now, is there a reason we can't use it and get rid of the bundled version?

kmaehashi:
> Now, my question: Since Thrust comes with the CUDA toolkit now, is there a reason we can't use it and get rid of the bundled version?

If I recall correctly Thrust bundled with CUDA Toolkit uses `__host__` code so NVRTC cannot compile it.

leofang:
> If I recall correctly Thrust bundled with CUDA Toolkit uses `__host__` code so NVRTC cannot compile it.

As long as the full qualifier is `__host__ __device__`, as in both #2538 and the Thrust, NVRTC will not complain: https://docs.nvidia.com/cuda/nvrtc/index.html#execution-space

emcastillo:
The minimum cuda version we support is 8, I guess we have to check the status of thrust headers on that version

emcastillo:
[...]

Issue #2516, opened 2019-10-02
ReductionKernel in linear algebra 


smarkesini:
Hi, 
cupy linalg norm does not use  reduction kernels, and makes copies of the input variable to compute intermediate results (abs, x*x).  The performance improvement with reduction kernel is not negligible, see below timing for a random vector of 1e8 float32 or complex64 elements, using cupy version '7.0.0b4', tesla k80, using attached code. 

 Is there a reason why reductionkernels are not used? would it be worth modifying these functions in cupy? 

The code also computes two outputs (inner product and norm squared of two difference vectors) in one kernel; this may be a useful example for others. 

Cheers, S. 

[reduce_multi_example.py.txt](https://github.com/cupy/cupy/files/3682603/reduce_multi_example.py.txt)

----

----------norm----------------
norm no kernel 1:8164.666	 t:0.664940
norm no kernel 2:8164.666	 t:0.000183
norm      kernel 1:8164.665	 t:0.017999
norm      kernel 2:8164.665	 t:0.000068

two methods for complex (abs(z)*abs(z)) or real(z*conj(z)) with no kernel:
----------norm-complex-----------
norm no kernel 1:11546.567	 t:0.215003
norm no kernel 2:11546.567	 t:0.000094
norm no kernel 1:11546.567	 t:0.432096
norm no kernel 2:11546.567	 t:0.002641
norm    kernel 1:11546.567	 t:0.000048
norm    kernel 2:11546.567	 t:0.000038

inner product of two difference vectors and norm squared computed simultaneously 

---------inner and norm2----------------
inner no kernel 1: -1508.135, norm2: 33331526.00, 	time:0.5372284
inner no kernel 2: -1508.135, norm2: 33331526.00, 	time:0.0003056
inner      kernel 1: -1508.137,  norm2: 33331528.00, 	time:0.0007456
inner      kernel 2: -1508.137,  norm2: 33331528.00, 	time:0.0000562

grlee77:
Hi @smarchesini, I also happened to be looking at this last Friday and had come up with a similar solution.

See the most recent commits in:
https://github.com/grlee77/cupy/commits/fast_norm

I used the lower level `create_reduction_func` instead of `ReductionKernel` to better handle complex dtypes (largely inspired by code written by @asi1024 adding complex dtype support for `np.var` in #2484).

For reference, the custom kernel generation code I came up with was:
```Python
_norm_preamble = '''
[...]

PR #3520, opened 2020-07-01
Small fixes for CUB block reduction kernels

leofang:
1. Remove all of the type constraints: I remember I set the limitations due to some errors when running the full test suite, but I could no longer reproduce it (with the latest master). @grlee77's [new `norm` kernels](https://github.com/cupy/cupy/issues/2516#issuecomment-538043247) might also need the support for complex numbers.
2. Add a possible exception to the optimizer: during the optuna optimization `CUDADriverError` could be raised due to out of resource. This was first observed in https://github.com/cupy/cupy/pull/3244#issuecomment-641479088, and I thought by constraining the search range it'd be remedied, but today I encountered it a few more times for different tasks, so apparently this is necessary. 
After adding this, I see that the error is gracefully handled:
```bash
[I 2020-06-30 22:29:07,612] Finished trial#1 with value: inf with parameters: {'block_size_log': 9, 'items_per_thread': 28}. Best is trial#0 with value: 0.0029116286219972552.
``` 
3. Allow compiler exceptions to propagate upward.
4. **(UPDATE)** Make `complex<T>` (almost) obey the rule of three (to fix fp16 -> complex conversion): This is basically a follow-up of #2629 and #2741. It turns out that by ensuring the rule of three (except for the destructor, which is trivial), we get the `float16` -> `complex<T>` conversion for free (through C++ implicit conversion fp16->fp32->complex<float>) without additional change. I should have done this when working on #2741...😢 Note the changes are in line with the Thrust implementation. 
5. **(UPDATE)** Fix tests

### DIFF
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -15,6 +15,7 @@ import math
 import string
 from cupy import _environment
 from cupy.core._kernel import _get_param_info
+from cupy.cuda import driver
 from cupy import util
 
 
@@ -279,18 +280,6 @@ cpdef inline tuple _can_use_cub_block_reduction(
     if axis_permutes_cub != tuple(range(in_arr.ndim)):
         return None
 
-    # To support generic reductions, note that some NumPy casting rules
-    # are not applicable in the C++ space (unless we tweak the type
-    # definitions). To circumvent this, we fall back to the old kernel.
-    # TODO(leofang): can we relax this?
-    if in_arr.dtype.kind != out_arr.dtype.kind:
-        # cannot cast complex to anything else
-        if in_arr.dtype.kind == 'c':
-            return None
-        # cannot cast float16 to complex
-        if in_arr.dtype.char == 'e' and out_arr.dtype.kind == 'c':
-            return None
-
     # full-reduction of N-D array: need to invoke the kernel twice
     cdef bint full_reduction = True if len(out_axis) == 0 else False
 
@@ -370,13 +359,13 @@ cdef _scalar.CScalar _cub_convert_to_c_scalar(
         return _scalar.CScalar.from_int32(value)
 
 
-cdef inline _cub_two_pass_launch(
+cdef inline void _cub_two_pass_launch(
         str name, Py_ssize_t block_size, Py_ssize_t segment_size,
         Py_ssize_t items_per_thread, str reduce_type, tuple params,
         list in_args, list out_args,
         str identity, str pre_map_expr, str reduce_expr, str post_map_expr,
         _kernel._TypeMap type_map, str input_expr, str output_expr,
-        str preamble, tuple options, stream):
+        str preamble, tuple options, stream) except *:
     '''
     Notes:
     1. Two-pass reduction: the first pass distributes an even share over
@@ -472,11 +461,11 @@ cdef inline _cub_two_pass_launch(
     func.linear_launch(gridx, inout_args, 0, blockx, stream)
 
 
-cdef inline _launch_cub(
+cdef inline void _launch_cub(
         self, out_block_num, block_size, block_stride,
         in_args, out_args, in_shape, out_shape, type_map,
         map_expr, reduce_expr, post_map_expr, reduce_type,
-        stream, params, cub_params):
+        stream, params, cub_params) except *:
     cdef bint full_reduction
     cdef Py_ssize_t contiguous_size, items_per_thread
     cdef function.Function func
@@ -537,7 +526,7 @@ def _get_cub_optimized_params(
             post_map_expr, reduce_type, stream, params, cub_params)
 
     def suggest_func(trial):
-        block_size_log = trial.suggest_int('block_size_log', 5, 9)
+        block_size_log = trial.suggest_int('block_size_log', 5, 10)
         block_size = 2 ** block_size_log
         items_per_thread = trial.suggest_int(
             'items_per_thread', 2, 32, step=2)
@@ -545,13 +534,14 @@ def _get_cub_optimized_params(
         trial.set_user_attr('block_size', block_size)
         return block_size, items_per_thread
 
+    # CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES is a possible error
     optimize_impl = optimize_config.optimize_impl
     best = optimize_impl(
         optimize_config, target_func, suggest_func,
         default_best={
             'block_size_log': default_block_size_log,
             'items_per_thread': default_items_per_thread,
-        })
+        }, ignore_error=(driver.CUDADriverError,))
 
     return best.params['items_per_thread'], best.user_attrs['block_size']
 
@@ -561,7 +551,7 @@ cdef bint _try_to_call_cub_reduction(
         stream, optimize_context, tuple key,
         map_expr, reduce_expr, post_map_expr, reduce_type, type_map,
         tuple reduce_axis, tuple out_axis, const shape_t& out_shape,
-        ndarray ret):
+        ndarray ret) except *:
     """Try to use cub.
 
     Updates `ret` and returns a boolean value whether cub is used.
--- a/cupy/core/_cub_reduction.pxd
+++ b/cupy/core/_cub_reduction.pxd
@@ -7,4 +7,4 @@ cdef bint _try_to_call_cub_reduction(
     stream, optimize_context, tuple key,
     map_expr, reduce_expr, post_map_expr, reduce_type, type_map,
     tuple reduce_axis, tuple out_axis, const shape_t& out_shape,
-    ndarray ret)
+    ndarray ret) except *
--- a/cupy/core/include/cupy/complex/complex.h
+++ b/cupy/core/include/cupy/complex/complex.h
@@ -75,12 +75,29 @@ struct complex {
 
   /* --- Constructors --- */
 
+  /*! Construct a complex number with an imaginary part of 0.
+   *
+   *  \param re The real part of the number.
+   */
+  inline __host__ __device__ complex(const T& re);
+
   /*! Construct a complex number from its real and imaginary parts.
    *
    *  \param re The real part of the number.
    *  \param im The imaginary part of the number.
    */
-  inline __host__ __device__ complex(const T& re = T(), const T& im = T());
+  inline __host__ __device__ complex(const T& re, const T& im);
+
+  /*! Default construct a complex number.
+   */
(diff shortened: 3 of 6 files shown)


leofang:
I guess the CI is not yet set up to test CUB kernels? @asi1024 Any chance you have resource to test this PR locally for me? I checked by running `pytest tests/cupy_tests/ -x` and encountered no error. I don't remember what were the failed tests that motivated me to add the type checks...😅

cupy/core/_cub_reduction.pyx
```diff
@@ -542,21 +530,23 @@ def _get_cub_optimized_params(
...
+        block_size_log = trial.suggest_int('block_size_log', 5, 10)
         block_size = 2 ** block_size_log
         items_per_thread = trial.suggest_int(
             'items_per_thread', 2, 32, step=2)
 
         trial.set_user_attr('block_size', block_size)
         return block_size, items_per_thread
 
+    # CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES is a possible error
+    from cupy.cuda import driver
```
asi1024, line 542: Why not import `driver` at the top of this file?

asi1024:
We will merge this PR after the CI is set up (https://github.com/chainer/chainer-test/pull/582 and https://github.com/cupy/cupy/pull/3461 are required)

cupy/core/_cub_reduction.pyx
leofang: This way we don't import until the optimizer is on? I can change that.

asi1024: `cupy.cuda.driver` will be imported anywhere from CuPy codebase, so we don't need to take care of it.

leofang: Thanks, changed in aecb521.

cupy/core/_cub_reduction.pyx
```diff
@@ -542,21 +531,22 @@ def _get_cub_optimized_params(
...
 
+    # CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES is a possible error
     optimize_impl = optimize_config.optimize_impl
     best = optimize_impl(
         optimize_config, target_func, suggest_func,
         default_best={
             'block_size_log': default_block_size_log,
             'items_per_thread': default_items_per_thread,
-        })
+        }, ignore_error=(driver.CUDADriverError,))
```
leofang, line 549: I wish we have a better way to filter out the particular error...`CUDADriverError` seems too general, although I can't think of other possibilities for how it can be thrown without out of resources.

leofang:
> We will merge this PR after the CI is set up (https://github.com/chainer/chainer-test/pull/582 and https://github.com/cupy/cupy/pull/3461 are required)

I think #2584 too?

leofang:
All the blockers are merged. Let's see what happens...
Jenkins, test this please

leofang:
Looks like the failures are due to no conversion from float16 to complex64 exists (which would make sense), but I can't reproduce it locally...@asi1024 any thoughts?

leofang:
I can reproduce it now. Working on a C++ solution. But, why doesn't the old implementation have this issue?

leofang:
> Working on a C++ solution.

@asi1024 I made this work by pulling out some of the stubs from here to a standalone header in `cupy/core/include/cupy/`:
https://github.com/cupy/cupy/blob/89cb2dfc2ccb9b3b429623c88c7ccd5d9a720278/cupyx/scipy/ndimage/filters.py#L1206-L1234
Then, I define a new constructor for complex numbers. Would you like me to push for you to take a look?

leofang:
A few chosen tests passed locally, let's see if the CI complains...
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Interesting...
- `cupy-py3-cutensor`: passed
- `cupy-py3-cub`: only one error (can't reproduce it locally)
- `cupy-py3` & `cupy-py35`: A bunch of errors happened in `test_filter.py` (can't reproduce either) and presumably timed out after 2 hr.

Main concerns:
- The CIs ran super long. Could it be the compile time is increased substantially with the header changes?
- The failures might be compiler-dependent (but because of the timeouts, I can't investigate based on the output logs).

I am thinking perhaps we should still forbid fp16 -> complex conversions for the time being, and move the header changes to a separate PR.

leofang:
Test again to see if slowness persists...

Jenkins, test this please

leofang:
My observation in https://github.com/cupy/cupy/pull/3520#issuecomment-656813192 persists. I will first do this:
> I am thinking perhaps we should still forbid fp16 -> complex conversions for the time being

But, instead of moving the header changes to a separate PR, I think we can make the new complex constructor optional, and turn it on only when CUB kernels are in use.

leofang:
This should now finish within 30 mins

Jenkins, test this please

leofang:
...Any chance the CI uses `ccache` to accelerate compilation? If so, the CI timings will not be reliable to detect compilation slowdown...

leofang:
> * `cupy-py3` & `cupy-py35`: A bunch of errors happened in `test_filter.py` (can't reproduce either) and presumably timed out after 2 hr.

I can reproduce now the failures in `test_filters.py` after clean-ups. It was simply due to 6ecf113 used `cupy` namespace, but in a later change it should have been modified to `cupy::details`. Will force-push and rerun tests.

leofang:
Jenkins, test this please

leofang:
I finally figured out what happened here...😞 CuPy has a C++ type literally called `float16`, which is the underly workhorse in the CUDA C++ space. I've worked with the raw `__half` for far too long that I kept treating `float16` as a synonym of `__half` even when I stared at the line `typedef float16 (...)`, so my previous effort was mistakenly on converting `__half` to `complex<T>`, whereas what we really need is `float16` to `complex<T>`.

Now that I discover the problem, I decided to pull it out from `carray.cuh` to `float16.cuh`, so that the definition can be reused. It's still included back to `carray.cuh`, so this change would not be noticed.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
OK so we know this works, but I wouldn't mind if we decide to roll back to a simpler version with less refactoring. I still don't understand why the original reduction kernel does not encounter this issue (https://github.com/cupy/cupy/pull/3520#issuecomment-656381146), though.

leofang:
Using type conversion operators seems to require less code change (no need to touch the complex and type_traits headers):
```diff
diff --git a/cupy/core/_cub_reduction.pyx b/cupy/core/_cub_reduction.pyx
index ffbc80a83..ff86f8fec 100644
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -29,7 +29,7 @@ cdef function.Function _create_cub_reduction_function(
 
     # 1. static_assert needs at least C++11 in NVRTC
     # 2. turn on fp16 -> complex conversion by defining the macro
-    options += ('--std=c++11', '-DCUPY_CUB_BLOCK_REDUCTION')
+    options += ('--std=c++11', '-DCUPY_CUB_BLOCK_REDUCTION_v2')
 
     # TODO(leofang): try splitting the for-loop into full tiles and partial
     # tiles to utilize LoadDirectBlockedVectorized? See, for example,
diff --git a/cupy/core/include/cupy/float16.cuh b/cupy/core/include/cupy/float16.cuh
index 3eed3722c..380387da0 100644
--- a/cupy/core/include/cupy/float16.cuh
+++ b/cupy/core/include/cupy/float16.cuh
@@ -1,4 +1,6 @@
 #pragma once
+#include <cupy/complex.cuh>
+
 
 // math
 #ifndef M_PI
@@ -60,6 +62,14 @@ public:
   explicit __device__ float16(const __half_raw &v): data_(v) {}
 
   __device__ operator float() const {return float(data_);}
+  #ifdef CUPY_CUB_BLOCK_REDUCTION_v2
+  __device__ float16::operator complex<float>() const {
+      return complex<float>(float(data_), 0);
+  }
+  __device__ float16::operator complex<double>() const {
+      return complex<double>(float(data_), 0);
+  }
+  #endif
 
   static const unsigned short nan = 0x7e00u;

```

asi1024:
Could you create a new PR for the refactoring?

leofang:
@asi1024 Could you help with https://github.com/cupy/cupy/pull/3520#issuecomment-657349107? I am thinking maybe we don't need a big refactoring.

leofang:
> @asi1024 Could you help with [#3520 (comment)](https://github.com/cupy/cupy/pull/3520#issuecomment-657349107)? (That is, why we didn't have the issue of converting fp16 to complex before.) I am thinking maybe we don't need a big refactoring.

I figured it out. It's because there's a `static_cast` in CuPy's kernels, allowing implicit conversions, whereas CUB does not have it. We'll need https://github.com/cupy/cupy/pull/3520#issuecomment-657355870.

leofang:
After the CI passes, I will factor out the refactoring part to a separate PR. We possibly don't need type traits for now...

Jenkins, test this please

leofang:
> After the CI passes, I will factor out the refactoring part to a separate PR. We possibly don't need type traits for now...

The refactoring PR is #3612. I will rebase once that one is merged.

leofang:
Test if the problem is due to https://github.com/cupy/cupy/pull/3612#issuecomment-659972674...

Jenkins, test this please

leofang:
I think we're close!

Jenkins, test this please

leofang:
OK so the non-compliance with the rule of three (https://github.com/cupy/cupy/pull/3612#issuecomment-659972674) is indeed the issue, among with other minor bugs. It seems we don't need to do any refactoring. Let me try to clean things up, perhaps with a rebase.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
`cupy-py35` timed out, but the rest is good.
Jenkins, test this please

leofang:
`cupy-py3` has a server connection issue (again), but none of them failed! I'll run the CI one more time, and then rebase and squash to clean up.
Jenkins, test this please

leofang:
Rebased. 

Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
@asi1024 I think this is ready. PR description updated too. PTAL.

asi1024:
LGTM!

leofang:
Thanks!

PR #3525, opened 2020-07-01
Add `cupy.correlate`

Dahlia-Chehata:


tests/cupy_tests/math_tests/test_misc.py
```diff
@@ -215,8 +215,8 @@ def test_nan_to_num_inf_nan(self):
 @testing.gpu
 @testing.parameterize(*testing.product({
     'mode': ['valid', 'same', 'full'],
-    'shape1': [(), (5,), (6,), (20,), (21,)],
-    'shape2': [(), (5,), (6,), (20,), (21,)],
+    'shape1': [(), (10,), (5,), (11,), (10,), (11,), (10,)],
+    'shape2': [(), (5,), (5,), (21,), (21,), (20,), (20,)],
```
asi1024, line 219: Why did you fix so?

Dahlia-Chehata: there are 3 (practically 6) cases that I wanted to test:
1 - first input has larger size
2 - second input has larger size ( the inverted case with the 4 combinations of even/odd lengths)
3 - same length

cupy/math/misc.py
```diff
@@ -55,8 +55,10 @@ def convolve(a, v, mode='full'):
 
 def _fft_convolve(a1, a2, mode):
 
+    offset = 0
     if a1.size < a2.size:
         a1, a2 = a2, a1
+        offset = a2.size % 2 == 0
```
asi1024, line 61: `1 - a2.size % 2` feels better because `a2.size % 2 == 0` is of boolean type.
```suggestion
        offset = 1 - a2.size % 2
```

cupy/statistics/correlation.py
```diff
@@ -48,7 +49,33 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
...
+    Returns:
+        cupy.ndarray: Discrete cross-correlation of a and v.
+
+    .. seealso:: :func:`numpy.correlate`
+
+    """
+    if a.ndim != 1 or v.ndim != 1:
+        raise ValueError('object too deep for desired array')
+    conj_v = cupy.conj(v[::-1])
+    method = cupyx.scipy.signal.choose_conv_method(a, conj_v, mode)
```
asi1024, line 69: `cupyx.scipy.signal.choose_conv_method` does not choose from the values in the input array, so no need to apply `conj`.
```suggestion
    method = cupyx.scipy.signal.choose_conv_method(a, v, mode)
```

cupy/statistics/correlation.py
```diff
@@ -48,7 +49,33 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
...
+    if method == 'direct':
+        if cupy.iscomplexobj(v):
+            v = cupy.conj(v)
```
asi1024, line 72: `v is v.conj` is `True` when `v` is not complex type array.
```suggestion
        v = v.conj()
```

tests/cupy_tests/math_tests/test_misc.py
asi1024: `testing.prodcut` checks in all combinations of parameters (for example, checked in 75 cases in the parameter of the previous version). So you don't need to fix it.
On the contrary, your test parameters don't include `a1.size > a2.size and a2.size % 2 == 0` case.

cupy/statistics/correlation.py
```diff
@@ -48,7 +49,32 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
...
+    """
+    if a.ndim != 1 or v.ndim != 1:
+        raise ValueError('object too deep for desired array')
+    method = cupyx.scipy.signal.choose_conv_method(a, v, mode)
+    if method == 'direct':
+        v = v.conj()
+        out = cupy.math.misc._dot_convolve(a, v[::-1], mode)
+    elif method == 'fft':
+        conj_v = cupy.conj(v[::-1])
+        out = cupy.math.misc._fft_convolve(a, conj_v, mode)
```
asi1024, line 74: For the simplification.
```suggestion
    if method == 'direct':
        out = cupy.math.misc._dot_convolve(a, v.conj()[::-1], mode)
    elif method == 'fft':
        out = cupy.math.misc._fft_convolve(a, v.conj()[::-1], mode)
```

cupy/statistics/correlation.py
```diff
@@ -48,7 +49,32 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
...
+
+    Returns:
+        cupy.ndarray: Discrete cross-correlation of a and v.
+
+    .. seealso:: :func:`numpy.correlate`
+
+    """
+    if a.ndim != 1 or v.ndim != 1:
+        raise ValueError('object too deep for desired array')
+    method = cupyx.scipy.signal.choose_conv_method(a, v, mode)
```
asi1024, line 68: Could you write a comment why we use `v` instead of `v.conj()[::-1]` here?
https://github.com/cupy/cupy/pull/3525#discussion_r448350880

cupy/statistics/correlation.py
```diff
@@ -48,7 +49,32 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
...
+    """
+    if a.ndim != 1 or v.ndim != 1:
+        raise ValueError('object too deep for desired array')
+    # choose_conv_method does not choose from the values in
+    # the input array, so no need to apply conj.
+    method = cupyx.scipy.signal.choose_conv_method(a, v, mode)
+    if method == 'direct':
+        out = cupy.math.misc._dot_convolve(a, v.conj()[::-1], mode)
+    elif method == 'fft':
+        out = cupy.math.misc._fft_convolve(a, v.conj()[::-1], mode)
```
grlee77, line 74: Perhaps it is just a micro-optimization, but for real arrays you can avoid the call to `conj`

```suggestion
    if v.dtype.kind == 'c':
        v = v.conj()
    if method == 'direct':
        out = cupy.math.misc._dot_convolve(a, v[::-1], mode)
    elif method == 'fft':
        out = cupy.math.misc._fft_convolve(a, v[::-1], mode)
```

asi1024: `conj` does not copy the memory space for real arrays, so this fix seems not needed.

asi1024:
Jenkins, test this please.

Dahlia-Chehata:
According to CI, numpy didn't raise ValueError when the input is empty. 

`>>>` ` np.correlate(np.array([]), np.array([]))`

  File "<__array_function__ internals>", line 5, in correlate
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/numpy/core/numeric.py", line 713, in correlate
    return multiarray.correlate2(a, v, mode)
ValueError: first array argument cannot be empty

cupy/math/misc.py
```diff
@@ -93,10 +95,10 @@ def _dot_convolve(a1, a2, mode):
     if a1.size == 0 or a2.size == 0:
         raise ValueError('Array arguments cannot be empty')
```
Dahlia-Chehata, line 96: Can I move this check to `correlate()` ? Already `convolve()` checks for empty inputs and also in order to keep consistency between `_fft_convolve` and `_dot_convolve`

asi1024:
@Dahlia-Chehata So please check the return value of `numpy.correlate` and `cupy.correlate` for empty input.

cupy/math/misc.py
asi1024: I agree your suggestion!

asi1024:
Jenkins, test this please.

Dahlia-Chehata:
How about removing empty test ?

numpy raises ValueError for empty inputs so is the current implementation but CI says otherwise https://github.com/cupy/cupy/pull/3525#issuecomment-652926971.
And the return value for empty inputs fail with 2 modes

asi1024:
If NumPy's behavior for the empty input depends on its version, please follow the newest stable version and add `@testing.with_requires('numpy>=...')` decorator to the test. If not, please fix your implementation.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #3528, opened 2020-07-02
Add `cupyx.scipy.sparse.kron()`

leofang:
The implementation and test cases are largely borrowed from SciPy, with minor tweaks. 

I actually implemented a correct `ElementwiseKernel` to construct the arrays `row`, `col`, and `data` at once (see 7135ec9) instead of following SciPy, but in my local tests I found it's slower, so I abandoned it. I guess it's because the memory access pattern isn't optimal. 

Moreover, the real bottleneck in this function is the three `coo_matrix()` calls. Together they took about 90% of time, so how the rest is implemented really doesn't matter much. I am not sure if this is a defect or something expected, will look into it, but it shouldn't be a blocker.

### DIFF
--- a/cupy/sparse/__init__.py
+++ b/cupy/sparse/__init__.py
@@ -24,7 +24,7 @@ from cupyx.scipy.sparse.construct import vstack  # NOQA
 # TODO(unno): implement dok_matrix
 # TODO(unno): implement lil_matrix
 
-# TODO(unno): implement kron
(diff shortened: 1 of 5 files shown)


takagi:
pfnCI, test this please.

takagi:
LGTM!

leofang:
Thanks!

PR #3533, opened 2020-07-02
[REVIEW] Compressed Sparse __setitem__

### LINKED ISSUE #2676: Implement column subsetting for `cupyx.scipy.sparse.csr.csr_matrix`
E.g. `xs[:, 1:3]` should work when `xs` is a `cupyx.scipy.sparse.csr.csr_matrix`. Note that row subsetting does work in this case.

This came up when trying to run Scanpy with CuPy and Dask: https://github.com/theislab/scanpy/issues/921.

### LINKED ISSUE #2677: Implement boolean column subsetting for `cupyx.scipy.sparse.csr.csr_matrix`
E.g. `xs[:, subset]`, where subset is e.g. `np.array([True, False, True, False, True])` and `xs` is a `cupyx.scipy.sparse.csr.csr_matrix`. Note that row subsetting does not work in this case.

This came up when trying to run Scanpy with CuPy and Dask: https://github.com/theislab/scanpy/issues/921.

### LINKED ISSUE #3115: 'csc_matrix' object does not support item assignment

cjnolet:
Closes #3115 
Closes #2676
Closes #2677

This PR builds upon #3486, also porting over over from Scipy the functions necessary to set values in both major and minor axes, using integers, slices, and arrays.


cjnolet:
Some light profiling is showing this is consistently the bottleneck in just about all the tests I've done so far: https://github.com/cupy/cupy/pull/3533/files#diff-91edca6879a46de14928e1d9107fd8adR896.

cjnolet:
@asi1024, 

With the 8.0.0 RC coming up, I'm really feeling pressure to get this PR in, along with #3486.

pentschev:
@asi1024 same here, do you think it would be possible to do a good thorough review on this PR so we can work on any necessary changes as quickly as possible to ensure it makes the cut into RC1?

cjnolet:
I'm looking suggestion anyone might have on the best way to parallelize this for loop from the `_insert_many` function. Profiling is showing it to be the largest bottleneck so far.

```
        # Collate old and new in chunks by major index
        indices_parts = []
        data_parts = []
        ui, ui_indptr = cupy.unique(i, return_index=True)

        # This part is modified slightly from the Scipy counterpart
        # because Cupy doesn't not have an `append` function for
        # ndarrays.
        to_add = cupy.array([j.size], ui_indptr.dtype)
        ui_indptr = cupy.hstack([ui_indptr, to_add])

        new_nnzs = cupy.diff(ui_indptr)

        prev = 0
        for c, (ii, js, je) in enumerate(zip(ui, ui_indptr, ui_indptr[1:])):

            # append old entries for each row
            start = self.indptr[prev]
            stop = self.indptr[ii]
            indices_parts.append(self.indices[start:stop])
            data_parts.append(self.data[start:stop])

            # handle duplicate j: keep last setting
            uj, uj_indptr = cupy.unique(j[js:je][::-1], return_index=True)
            if len(uj) == je - js:
                indices_parts.append(j[js:je])
                data_parts.append(x[js:je])
            else:
                indices_parts.append(j[js:je][::-1][uj_indptr])
                data_parts.append(x[js:je][::-1][uj_indptr])
                new_nnzs[c] = len(uj)

            prev = ii
```

cjnolet:
Now that I think about it, I might be able to combine i and j into a single `int64` array where the high order bits are i and the low order bits are j. Then we have all the info we need in a single array and we can call `unique` once to get all the  unique columns per row and then do a `scatter_add` on those.

cjnolet:
With the changes described above, `_insert_many()` is now orders of magnitude faster thank Scipy's `_insert_many()`. Most of the pytests are now faster end-to-end than Scipy but there are some cases where we are slower, all appearing to be completely unrelated to `_insert_many()`. I'm going to do more profiling of the algorithm end-to-end to better isolate those bottlenecks.

cjnolet:
@asi1024, I have opened the following 3 PRs in addition to this one: #3657, #3658, #3639. 

#3657 Includes the main scaffolding ported from Scipy but raises a NotImplementedError for all features not related to slicing. 
#3658 Includes the changes from the previous PR but includes sparse row indexing
#3659 Includes the changes from #3658, but includes sparse column indexing. 

This PR includes all the changes from #3659 and adds inner indexing (`arrayXarray`). 

I built the new PRs incrementally in this way in hopes the deltas could be more easily reviewed. I will keep rolling the changes backwards into this PR so that they can be merged one after the other. 

Will this approach work for you?

cjnolet:
@leofang, if you get a moment, I'm hoping to get your thoughts on https://github.com/cupy/cupy/pull/3533/files#diff-91edca6879a46de14928e1d9107fd8adR920 (in case it doesn't expand for you it's `cupyx/scipy/sparse/compressed.py` line 920). 

Packing 32-bit elements into a 64-bit to sort them together is a common trick but it won't work when the inputs are 64-bit indices. I just got to thinking about it with the knowledge that we are working towards support for 64-bit indices. I just haven't been able to think a better alternative to parallelize insertions. We could do this for now and raise an error if the insertions are being done on a sparse array with 64-bit index arrays, then add a better solution later when we find one, perhaps. I'm really looking for your opinion on this one.

leofang:
Hi @cjnolet Sorry for my silence. I just took a quick look at `_insert_many()`. The bit packing trick is very fancy, but I wonder if a `lexsort` could be easier? Maybe I don't get the gist here...?
```python
        keys = cupy.stack([j, i])
        order = cupy.lexsort(keys)
        new_i = i[order]
        new_j = j[order]
```
But I am not sure yet how to pick out the unique values.

cjnolet:
Thanks @leofang! This works. I just needed to write a small kernel sets a filtering mast for removing all duplicates but the last one (the ones with the highest lexsort indices). This is great, because now it can fully support 64-bit indices.

emcastillo:
Amazing work! thanks a lot 😃

emcastillo:
btw, you have just included all the docs in the PR 😂

cjnolet:
@emcastillo, I figured it was only a matter of time before I did that! I'll remove them from the history.

cjnolet:
I have one more thing to do on this PR and I think it'll be ready for review: 

Currently, the sort is taking most of the insertion time. Some of the recent changes enable us to now construct the output indices and data arrays in place. We can do this since the elements to construct the output arrays are already in sorted order so we no longer need 2 separate kernels for populating the existing and new values. We should be able execute a single kernel to populate the values in sorted order, thus removing the need for expensive sort.

cjnolet:
@emcastillo, this should be ready to begin reviewing. Now that we have the scaffolding for the sparse slicing in place, I don't believe this is being blocked anymore, unless you prefer to wait for all the sparse advanced indexing PRs first.

cjnolet:
Here's the script I've been using for benchmarking the sparse setitem:

```python
import unittest

import cupy

from cupy import sparse
from cupy import testing

import pytest


    'format': ['csr', 'csc'],
    'density': [0.5],
    'dtype': ['float32', 'float64', 'complex64', 'complex128'],
    'n_rows': [15000],
    'n_cols': [15000]
}))
@testing.with_requires('scipy')
class TestSetitemIndexing(unittest.TestCase):

    def _run(self, maj, min=None, data=5):

        for i in range(2):
            a = cupy.sparse.random(self.n_rows, self.n_cols,
                                   format=self.format,
                                   density=self.density)

            # sparse.random doesn't support complex types
            # so we need to cast
            a = a.astype(self.dtype)

            if isinstance(maj, cupy.ndarray):
                maj_h = maj.get()
            else:
                maj_h = maj

            if isinstance(min, cupy.ndarray):
                min_h = min.get()
            else:
                min_h = min

            import time

            if min is not None:
                expected = a.get()

                cupy.cuda.Stream.null.synchronize()
                cpu_time = time.time()
                expected[maj_h, min_h] = data
                cpu_stop = time.time() - cpu_time

                actual = a

                gpu_time = time.time()
                cupy.cuda.Stream.null.synchronize()
                actual[maj, min] = data
                cupy.cuda.Stream.null.synchronize()
                gpu_stop = time.time() - gpu_time
            else:
                expected = a.get()

                cupy.cuda.Stream.null.synchronize()
                cpu_time = time.time()
                expected[maj_h] = data
                cpu_stop = time.time() - cpu_time

                actual = a

                gpu_time = time.time()
                cupy.cuda.Stream.null.synchronize()
                actual[maj] = data
                cupy.cuda.Stream.null.synchronize()
                gpu_stop = time.time() - gpu_time

        print("maj=%s, min=%s, format=%s, cpu_time=%s, gpu_time=%s"
              % (maj, min, self.format, cpu_stop, gpu_stop))

        if cupy.sparse.isspmatrix(actual):
            actual.sort_indices()
            expected.sort_indices()

            cupy.testing.assert_array_equal(
                actual.indptr, expected.indptr)
            cupy.testing.assert_array_equal(
                actual.indices, expected.indices)
            cupy.testing.assert_array_equal(
                actual.data, expected.data)

        else:

            cupy.testing.assert_array_equal(
                actual.ravel(), cupy.array(expected).ravel())

    def test_major_slice(self):
        self._run(slice(5, 10000), data=5)
        self._run(slice(5, 4), data=5)
        self._run(slice(4, 5, 2), data=5)
        self._run(slice(5, 4, -2), data=5)
        self._run(slice(2, 4), slice(0, 2), [[4], [1]])
        self._run(slice(2, 4), slice(0, 2), [[4, 5], [6, 7]])
        self._run(slice(2, 4), 0, [[4], [6]])

        self._run(slice(5, 9))
        self._run(slice(9, 5))

    def test_major_all(self):
        self._run(slice(None))

    def test_major_scalar(self):
        self._run(10)

    def test_major_fancy(self):
        self._run([1, 5, 4])
        self._run([10, 2])
        self._run([2])

    def test_major_slice_minor_slice(self):
        self._run(slice(1, 5), slice(1, 5))

    def test_major_slice_minor_all(self):
        self._run(slice(1, 5), slice(None))
        self._run(slice(5, 1), slice(None))

    def test_major_slice_minor_scalar(self):
        self._run(slice(1, 5), 5)
        self._run(slice(5, 1), 5)
        self._run(slice(5, 1, -1), 5)

    def test_major_slice_minor_fancy(self):
        self._run(slice(1, 10, 2), [1, 5, 4])

    def test_major_scalar_minor_slice(self):
        self._run(5, slice(1, 5))

    def test_major_scalar_minor_all(self):
        self._run(5, slice(None))

    def test_major_scalar_minor_scalar(self):
        self._run(5, 5)

    def test_major_scalar_minor_fancy(self):
        self._run(5, [1, 5, 4])

    def test_major_all_minor_all(self):
        self._run(slice(None), slice(None))

    def test_major_all_minor_fancy(self):
        self._run([1, 2, 3, 4, 1, 6, 1, 8, 9], [1, 5, 2, 3, 4, 5, 4, 1, 5])
        self._run(slice(None), [0, 3, 4, 1, 1, 5, 5, 2, 3, 4, 5, 4, 1, 5])

    def test_major_fancy_minor_fancy(self):
        self._run([1, 5, 4], [1, 5, 4])
        self._run([2, 0, 10], [9, 2, 1])
        self._run([2, 0], [2, 1])

    def test_major_fancy_minor_all(self):
        self._run([1, 5, 4], slice(None))

    def test_major_fancy_minor_scalar(self):
        self._run([1, 5, 4], 5)

    def test_major_fancy_minor_slice(self):
        self._run([1, 5, 4], slice(1, 5))
        self._run([1, 5, 4], slice(5, 1, -1))

    def test_major_bool_fancy(self):
        rand_bool = cupy.random.random(self.n_rows).astype(cupy.bool)
        self._run(rand_bool)

    def test_major_slice_with_step(self):

        # positive step
        self._run(slice(1, 10, 2))
        self._run(slice(2, 10, 5))
        self._run(slice(0, 10, 10))

        self._run(slice(1, None, 2))
        self._run(slice(2, None, 5))
        self._run(slice(0, None,  10))

        # negative step
        self._run(slice(10, 1, -2))
        self._run(slice(10, 2, -5))
        self._run(slice(10, 0, -10))

        self._run(slice(10, None, -2))
        self._run(slice(10, None, -5))
        self._run(slice(10, None, -10))

    def test_major_slice_with_step_minor_slice_with_step(self):

        # positive step
        self._run(slice(1, 10, 2), slice(1, 10, 2))
        self._run(slice(2, 10, 5), slice(2, 10, 5))
        self._run(slice(0, 10, 10), slice(0, 10, 10))

        # negative step
        self._run(slice(10, 1, 2), slice(10, 1, 2))
        self._run(slice(10, 2, 5), slice(10, 2, 5))
        self._run(slice(10, 0, 10), slice(10, 0, 10))

    def test_major_slice_with_step_minor_all(self):

        # positive step
        self._run(slice(1, 10, 2), slice(None))
        self._run(slice(2, 10, 5), slice(None))
        self._run(slice(0, 10, 10), slice(None))

        # negative step
        self._run(slice(10, 1, 2), slice(None))
        self._run(slice(10, 2, 5), slice(None))
        self._run(slice(10, 0, 10), slice(None))

    def test_fancy_setting(self):
        self._run(0, 0, 5)
        self._run([0, 5, 10, 2], 0, [1, 2, 3, 2])

        self._run([[True], [False], [True]], data=5)
        self._run([[True], [False], [False], [True], [True], [True]], data=5)
        self._run([True, False, False, True, True, True], data=5)
```

I'm running it now and will post the results.

cjnolet:
@asi1024, here are the benchmark numbers for a 1500x1500 sparse matrix @ 50% density. As usual, I executed this benchmark on a Threadripper 16-core w/ a GV100 GPU. 

[setitem_benchmark_08020.zip](https://github.com/cupy/cupy/files/5104659/setitem_benchmark_08020.zip)

cupyx/scipy/sparse/_index.py
```diff
@@ -358,6 +469,58 @@ def __getitem__(self, key):
...
+            if row.ndim == 1:
+                row = row[:, None]
+        else:
+            col = cupy.atleast_1d(col)
+
+        i, j = cupy.broadcast_arrays(row, col)
+        if i.shape != j.shape:
+            raise IndexError('number of row and column indices differ')
+
+        from .base import isspmatrix
```
asi1024, line 499: Please import at top-level.

cupyx/scipy/sparse/compressed.py
```diff
@@ -327,6 +333,99 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
         diff = diff_out;
         ''', 'has_canonical_format')
 
+    _insert_many_populate_arrays = core.ElementwiseKernel(
```
asi1024, line 336: How about moving this kernel to `_index.py`?

cupyx/scipy/sparse/index.py
```diff
@@ -0,0 +1,767 @@
+"""Indexing mixin for sparse matrix classes.
```
asi1024, line 1: Please remove this file.

cupyx/scipy/sparse/compressed.py
```diff
@@ -660,6 +796,321 @@ def _major_slice(self, idx, copy=False):
...
+        #
+        i = cupy.array(i, dtype=self.indptr.dtype,
+                       copy=True, ndmin=1).ravel()
+        j = cupy.array(j, dtype=self.indices.dtype,
+                       copy=True, ndmin=1).ravel()
+        # check_bounds(i, M)
+        # check_bounds(j, N)
+        return i, j, M, N
+
+    @cupy.prof.TimeRangeDecorator(message="_set_many", color_id=9)
```
asi1024, line 880: Please remove this decorator.

cupyx/scipy/sparse/compressed.py
```diff
@@ -660,6 +796,321 @@ def _major_slice(self, idx, copy=False):
...
+        if not broadcast:
+            values = values[:i.size]
+
+        self[i, j] = values
+
+    def _prepare_indices(self, i, j):
+        M, N = self._swap(*self.shape)
+
+        # TODO(cjnolet) Verify whether this is needed.
+        # def check_bounds(indices, bound):
```
cjnolet, line 862: @asi1024, what are your thoughts on this one?

cupyx/scipy/sparse/compressed.py
```diff
@@ -660,6 +796,321 @@ def _major_slice(self, idx, copy=False):
...
+        # check index and data arrays
+        if (self.indices.size != self.data.size):
+            raise ValueError("indices and data should have the same size")
+        if (int(self.indptr[-1]) > self.indices.size):
+            raise ValueError("Last value of index pointer should be less than "
+                             "the size of index and data arrays")
+
+        self.prune()
+
+        # TODO(cjnolet) Verify whether this is needed
```
cjnolet, line 1100: @asi1024, also wondering about this one. Should we keep this here?

cupyx/scipy/sparse/compressed.py
```diff
@@ -673,6 +702,297 @@ def _major_slice(self, idx, copy=False):
...
+        row_counts = cupy.zeros(ui_indptr.size-1, dtype=idx_dtype)
+        cupyx.scatter_add(
+            row_counts, cupy.searchsorted(rows, indptr_inserts), 1)
+
+        self._perform_insert(indices_inserts, data_inserts,
+                             rows, row_counts, idx_dtype)
+
+        self.check_format(full_check=False)
+
+    def check_format(self, full_check=True):
```
asi1024, line 936: Could you split `check_format` and `prune` into another PR?

cjnolet:
pfnCI, test this please

cjnolet:
pfnCI, test this please

kmaehashi:
Blocked by #3826

cjnolet:
@asi1024 looking more closely at this PR, I'm not sure how much it is actually blocked by #3826. 

We were told this would be included in the 8.1.0 release and back-ported to 8.0.0. Is it possible to start a review?

cjnolet:
pfnCI, test this please

cjnolet:
pfnCI, test this pleast

jakirkham:
Odd the bot says this failed, but the checks below show this is passing. @asi1024 do you know what might be going on here? 🙂

kmaehashi:
pfnCI, test this please.

cupyx/scipy/sparse/_index.py
```diff
@@ -127,6 +127,214 @@ def _csr_row_index(rows, Ap, Aj, Ax):
...
+"""
+_csr_sample_offsets_ker = core.ElementwiseKernel(
+    '''I n_row, I n_col, raw I Ap, raw I Aj, raw I Bi, raw I Bj''',
+    'raw I Bp, raw bool dupl', '''
+    const I j = Bi[i]; // sample row
+    const I k = Bj[i]; // sample column
+    const I row_start = Ap[j];
+    const I row_end   = Ap[j+1];
+    I offset = -1;
+    for(I jj = row_start; jj < row_end; jj++)
```
asi1024, line 302: Why not avoid this loop by using `sum_duplicated`?

tests/cupyx_tests/scipy_tests/sparse_tests/test_index.py
```diff
@@ -10,14 +10,218 @@
...
+        self._run(10, 24, 5)
+
+    def test_major_scalar_minor_fancy(self):
+        self._run(5, [1, 5, 4])
+
+    def test_major_all_minor_all(self):
+        self._run(slice(None), slice(None))
+
+    def test_major_all_minor_fancy(self):
+        self._run([1, 2, 3, 4, 1, 6, 1, 8, 9], [1, 5, 2, 3, 4, 5, 4, 1, 5])
```
asi1024, line 127: Please check also for numpy.ndarray, cupy.ndarray inputs.

cupyx/scipy/sparse/_index.py
cjnolet: I tried to avoid sorting the entire array unless absolutely necessary (e.g., when a duplicate is actually found). One of the reasons for constructing the final array in sorted order during `insert_many` is that it was much less expensive than adding the insertions after the existing values and sorting.

cjnolet: And I just realized that some of my more recent changes already assume sorted order, so it needs to be done anyways.

cjnolet: The only way I can think of to avoid this loop is to convert the `indptr` to a COO row array, which would use more memory but would make the offset finding more parallel. Without knowing the offsets of the indices we are setting, it's not obvious to me how to set these values. Please let me know what I can do to fix this.

cjnolet: I was able to remove the inner loop iterations, at least. When you get a moment, can you PTAL?

asi1024: How about creating a sparse array of [nnz of initial array] + [size of Bx] size, and then apply `sum_duplicated` to it? I think all loops in element-wise kernels.

asi1024:
Could you fix Travis failures?

cjnolet:
pfnCI, test this please

asi1024:
How about this comment? https://github.com/cupy/cupy/pull/3533#discussion_r486794594

cupyx/scipy/sparse/_index.py
asi1024: `sum_duplicates` may not be needed. It seems OK to just return the array of nnz of initial array + size of Bx size.

cjnolet:
Thinking through this a little more, what do you think about mapping the rows to warps, 1 row per warp, and setting the warp size based on the average degree of the CSR? 

That will allow us to increase the parallelism over each column while avoiding the additional array of size nnz. What do you think? 

I'm really okay with either solution. The one just described is going to require a RawKernel though, I think.

cjnolet:
@asi1024, I removed `_csr_sample_offsets` and updated the code to use `_csr_sample_values`. Can you PTAL when you get a moment?

cjnolet:
pfnCI, test this please

cjnolet:
pfnCI, test this please

cjnolet:
@asi1024 @

cjnolet:
pfnCI, test this please

cjnolet:
pfnCI, test this please

asi1024:
@cjnolet Sorry for my late response.

> Thinking through this a little more, what do you think about mapping the rows to warps, 1 row per warp, and setting the warp size based on the average degree of the CSR?
>
> That will allow us to increase the parallelism over each column while avoiding the additional array of size nnz. What do you think?
>
> I'm really okay with either solution. The one just described is going to require a RawKernel though, I think.

Now I have another idea that does not require an additional array. Can I show it in another PR?
I want to improve the parallelism of `_get_arrayXarray` before the improve of setiem, because `_get_arrayXarray` is called in setiem code path. However, we should not leave this PR unmerged for a long time.

So I plan to create another PR for the improvement after this PR is merged.

tests/cupyx_tests/scipy_tests/sparse_tests/test_index.py
```diff
@@ -9,17 +9,258 @@
...
+@testing.parameterize(*testing.product({
+    'format': ['csr', 'csc'],
+    'density': [0.9],
+    'dtype': ['float32', 'float64', 'complex64', 'complex128'],
+    'n_rows': [25, 150],
+    'n_cols': [25, 150]
+}))
+@testing.with_requires('scipy>=1.4.0')
+@testing.gpu
+class TestSetitemIndexing(unittest.TestCase):
```
asi1024, line 41: Could you add tests for the case of
- Boolean mask of numpy ndarray
- numpy/cupy ndarray `data` which is set to sparse array
- numpy scalar and cupy 0-dim `data` which is set to sparse array

asi1024 (commented):
LGTM excepts the code coverage.

cjnolet:
pfnCI, test this please

tests/cupyx_tests/scipy_tests/sparse_tests/test_index.py
asi1024: Please add numpy scalar and cupy 0-dim data cases.

cjnolet:
pfnCI, test this please

cjnolet:
pfnCI, test this please

cjnolet:
pfnCI, test this please

cjnolet:
pfnCI, test this please

cjnolet:
Completely unrelated test failure. That's strange.

cjnolet:
pfnCI, test this please

asi1024:
LGTM! Thanks!

PR #3534, opened 2020-07-02
Avoid early compilation when initializing a RawModule instance

leofang:
Fix https://github.com/cupy/cupy/pull/3486#issuecomment-653145880. There, the CI error occurred because `RawModule` compiles at initialization, not when `get_function()` is called, so in a Read the Docs environment where nvcc is not available, the build fails. Now it's fixed by ~~checking if the build environment has CUDA or not~~ (**UPDATE**) deferring the compilation to when something needs to be fetched.

~~In fact, making this `no_cuda` flag (and potentially a few other ones) available could help some other Cython modules to conditionally compile.~~

cc: @cjnolet

### DIFF
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -628,9 +628,6 @@ def parse_args():
     }
     if check_readthedocs_environment():
         arg_options['no_cuda'] = True
-    # make some compile-time expressions available to Cython codes;
-    # currently we do not expose any, and leave this for future extension
-    arg_options['compile_time_env'] = {}
     return arg_options
 
 
@@ -731,7 +728,7 @@ def cythonize(extensions, arg_options):
     # Embed signatures for Sphinx documentation.
     directives['embedsignature'] = True
 
-    cythonize_option_keys = ('annotate', 'compile_time_env')
+    cythonize_option_keys = ('annotate',)
     cythonize_options = {key: arg_options[key]
                          for key in cythonize_option_keys}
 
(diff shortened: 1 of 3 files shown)


leofang:
Will do a few cosmetic changes.

leofang:
Done.

asi1024:
How about stopping compiling the kernel at `__init__` for the consistency with the behavior of `RawKernel`?

leofang:
Good point. If no compatible is acceptable then sounds good. So after the change the compilation will happen when any method is called.

leofang:
@asi1024 Done. PR title is also changed to reflect the update.

cupy_setup_build.py
```diff
@@ -728,7 +731,7 @@ def cythonize(extensions, arg_options):
     # Embed signatures for Sphinx documentation.
     directives['embedsignature'] = True
 
-    cythonize_option_keys = ('annotate',)
+    cythonize_option_keys = ('annotate', 'compile_time_env')
```
asi1024, line 734: This fix looks not needed anymore.

leofang: OK I'll take it off from this PR and revive it in another (was planning to touch the nccl module using this capability).

leofang:
> Recompile does not happen in this test now.

Good catch. I fixed it.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

Issue #3289, opened 2020-04-16
missing constructor for `scipy.sparse.csr_matrix`


csukuangfj:
<https://docs-cupy.chainer.org/en/stable/reference/sparse.html#sparse-matrices>
says

> CuPy supports sparse matrices using cuSPARSE. **These matrices have the same interfaces of SciPy’s sparse matrices.**

**However**, the constructor
> csr_matrix((data, (row_ind, col_ind)), [shape=(M, N)])

from <https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html#scipy-sparse-csr-matrix>

![Screen Shot 2020-04-16 at 18 36 47](https://user-images.githubusercontent.com/5284924/79446710-434adb00-8011-11ea-8242-ca7f7153e3a5.png)

is **missing** in `cupyx.scipy.sparse.csr_matrix` in
https://docs-cupy.chainer.org/en/stable/reference/generated/cupyx.scipy.sparse.csr_matrix.html#cupyx-scipy-sparse-csr-matrix

![Screen Shot 2020-04-16 at 18 37 47](https://user-images.githubusercontent.com/5284924/79446777-64abc700-8011-11ea-8690-ef2943ab5b33.png)

leofang:
To add another example: the constructor `coo_matrix(D)` (initializing a COO matrix with a dense matrix) is also missing.

PR #3536, opened 2020-07-03
Fix sparse min/max to return sparse matrix

### LINKED ISSUE #3504: Sparse min/max have less SciPy-compatibility
The outptus of sparse min/max are less compatible to those of Scipy.

```Python
import numpy
import scipy

import cupy
from cupy import testing
import cupyx.scipy

def _make(xp, sp):
    m = testing.shaped_random((10, 20), xp=xp, dtype='f', scale=1)
    m[m < 0.95] = 0
    return sp.sparse.csr_matrix(m)

for (xp, sp) in [(numpy, scipy), (cupy, cupyx.scipy)]:
    m = _make(xp, sp).min(axis=0)
    print(m.dtype, m.shape, type(m))
```

**Output**
```Python
float32 (1, 20) <class 'scipy.sparse.coo.coo_matrix'>
float64 (20,) <class 'cupy.core.core.ndarray'>
```

takagi:
Close #3504.

This PR fixes sparse min/max to return a value as a sparse matrix. I will parameterize the test in another PR.


cupyx/scipy/sparse/data.py
```diff
@@ -139,15 +140,28 @@ def _min_or_max_axis(self, axis, min_or_max, sum_duplicates, nonzero):
...
         if sum_duplicates:
             mat.sum_duplicates()
 
         # Do the reudction
         value = mat._minor_reduce(min_or_max, axis, nonzero)
+        major_index = cupy.arange(M)
 
-        return value
+        mask = value != 0
+        major_index = cupy.compress(mask, major_index)
```
toslunar, line 154: OK to leave the following issues because the PR is a bugfix.

- For efficiency, it is possible that `mat._minor_reduce` should compute the `major_index` and the compressed `value`.
- `coo_matrix(D)` could be implemented, which is discussed at https://github.com/cupy/cupy/issues/3289#issuecomment-653194583

tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
```diff
@@ -915,10 +915,9 @@ def test_min_sparse_axis_0(self):
                                        cupy.array(dm_data.indptr)),
                                       shape=(10, 20))
 
-        da_cupy_values = cupy.asnumpy(cp_matrix.min(axis=0))
-        da_scipy_values = numpy.array(dm_data.min(axis=0).todense().ravel())
-        da_scipy_values = da_scipy_values[0, :]
-        assert numpy.array_equal(da_cupy_values, da_scipy_values)
+        da_cupy_values = cp_matrix.min(axis=0)
+        da_scipy_values = dm_data.min(axis=0)
+        testing.assert_array_equal(da_scipy_values.A, da_scipy_values.A)
```
toslunar, line 920: `testing.assert_array_equal` does not support sparse, while `testing.numpy_cupy_array_equal` and `testing.numpy_cupy_allclose` do.

takagi: I make the tests use `testing.numpy_cupy_arrray_equal`. Would you check them? To test `nonzero` parameter, which SciPy does not have yet https://github.com/scipy/scipy/pull/11899 and will be marked as experimental in https://github.com/cupy/cupy/pull/3583, I supply a bit messy SciPy inputs.

toslunar:
Jenkins, test this please.

cupyx/scipy/sparse/data.py
takagi: I'll revisit these points in another PR!

takagi:
Thanks! I'll fix the SciPy input to make it work in case `axis=None` and `nonzero=True`.

toslunar:
Jenkins, test this please.

PR #3537, opened 2020-07-03
Fix shape of return value of sparse argmin/argmax

takagi:
Follows #3536.

This PR fixes the shape of the return value of sparse argmin/argmax to follow SciPy. I will parameterize the test in another PR.


takagi:
Close this as obsolete.

PR #3538, opened 2020-07-03
Enhance boolean indexing

### LINKED ISSUE #627: IndexError when we use indexing with combination of int and boolean.
```python
a = numpy.ones((1,1))
a[0, True] = 2   # array([[ 2.]])
b = cupy.ones((1,1))
b[0, True] = 2  # Index Error
```

More practically, In the case when we want to apply a mask to a specific channel of an image:
```python
img = xp.random.random((3, 224, 224))
mask = img[:, :, 0] > 0.5
img[mask, 0] = 0. 
```
Index error is raised when `xp == cupy`.

okuta:
Fix #627
We can use
```
import cupy
b = cupy.ones((1,1))
b[0, True] = 2
```


emcastillo:
Jenkins, test this please

PR #3540, opened 2020-07-03
Improve scan operation

okuta:
This PR reduces CUDA kernel launch count.

### DIFF
--- a/cupy/core/_routines_indexing.pyx
+++ b/cupy/core/_routines_indexing.pyx
@@ -64,6 +64,7 @@ cpdef ndarray _ndarray_argwhere(ndarray self):
     cdef Py_ssize_t count_nonzero
     cdef int ndim
     cdef ndarray nonzero
+    numpy_int64 = numpy.int64
     if self.size == 0:
         count_nonzero = 0
     else:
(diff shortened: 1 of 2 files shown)


emcastillo:
Thanks!

cupy/core/_routines_indexing.pyx
```diff
@@ -63,14 +63,18 @@ cdef tuple _ndarray_nonzero(ndarray self):
...
-        nonzero = cupy.core.not_equal(r, 0, ndarray(r.shape, dtype))
-        del r
-        scan_index = _math.scan(nonzero, op=_math.scan_op.SCAN_SUM)
+        if self.dtype == numpy.bool_:
+            nonzero = self.ravel()
+        else:
+            nonzero = cupy.core.not_equal(self, 0)
+            nonzero = nonzero.ravel()
+        scan_index = _math.scan(nonzero, op=_math.scan_op.SCAN_SUM,
+                                dtype=dtype)
```
leofang, line 77: Do we always have to use `int64` as the dtype here?

okuta: Thanks, I fixed.

okuta:
jenkins, test this please.

cupy/core/_routines_indexing.pyx
emcastillo: I think @leofang meant that we could use int32 for arrays with less elements than 2**31-1

okuta: Because numpy returns `numpy.int64` index array.

leofang: I see. But, it seems scan is just an intermediate step, and the actual returned array is computed in `_nonzero_kernel*`? Would it be possible to defer the generation of an int64 array to there?

leofang: (Not saying it'd be faster! Just curious...Feel free to ignore me 😅)

okuta: Ah, I understand this function only returns `dst`

okuta:
jenkins, test this please.

leofang:
Cool!

PR #3547, opened 2020-07-05
Add `cupy.poly`

Dahlia-Chehata:
blocked by #3828


cupy/lib/polynomial.pyx
```diff
@@ -4,6 +4,47 @@ import cupy
...
+    else:
+        raise ValueError('Input must be 1d or non-empty square 2d array.')
+
+    if seq.size == 0:
+        return 1.0
+
+    out = cupy.ones((1,), seq.dtype)
+    seq = cupy.column_stack((cupy.ones(seq.size, seq.dtype), -seq))
+    for zero in seq:
+        out = cupy.convolve(out, zero, 'full')
```
asi1024, line 44: Please parallelize this loop. We can achieve this computation with O(log n) time FFT call.

Dahlia-Chehata: The best algorithm I know for coefficients computation using FFT is a recursive O(n log^2 n) time one. Is that what you mean or do you have a better approach?

asi1024: Yes.

cupy/lib/polynomial.pyx
```diff
@@ -4,6 +4,16 @@ import cupy
...
 
 
+def _fft_poly(seq):
+    n = seq.shape[0]
+    if n == 1:
+        return seq[0]
+    if n == 2:
+        return cupy.math.misc._fft_convolve(seq[0], seq[1], 'full')
+    return cupy.math.misc._fft_convolve(_fft_poly(seq[:n/2]),
+                                        _fft_poly(seq[n/2:]), 'full')
```
asi1024, line 14: This implementation calls FFT O(n) times, but we can achieve O(log n) times if we support Nd convolve.

Dahlia-Chehata: You suggest supporting scipy.signal.convolve or extending the functionality of cupy.convolve ?

asi1024: OK!

kmaehashi:
Related discussion: https://github.com/cupy/cupy/pull/3371#issuecomment-654464151

cupy/lib/polynomial.pyx
Dahlia-Chehata:
```
def _fft_poly(seq):
    n = seq.shape[0]
    if n == 1:
        return seq[0]
    if n == 2:
        return cupy.math.misc._fft_convolve(seq[0], seq[1], 'full')
    return signal.convolve(_fft_poly(seq[:n / 2]), _fft_poly(seq[n / 2:]), 'full', 'direct')
```
I'm not really sure if that's what you meant or not. If so, I want to clarify that
1. complex dtype give different results in comparison with numpy's
2. Only `direct` method is currently supported for `cupyx.scipy.signal.convolve` that's why I have to specify the method.

asi1024: No, it requires O(n) time FFT kernel call.

asi1024:
@Dahlia-Chehata Could you add TODO comment for the issue https://github.com/cupy/cupy/pull/3547#discussion_r449964352, or support 2d inputs in `cupyx.scipy.signal.convolve` and use it?

asi1024:
@Dahlia-Chehata Can I takeover the PR?

PR #3548, opened 2020-07-05
Add `cupy.polyadd`

Dahlia-Chehata:


Dahlia-Chehata:
Should ` __radd__` be eliminated ? If not, in which cases is it called ?
I didn't add tests for it.

cupy/lib/polynomial.pyx
```diff
@@ -4,6 +4,37 @@ import cupy
...
+        a2 = a2.coeffs
+        truepoly = True
+    a1 = cupy.atleast_1d(a1)
+    a2 = cupy.atleast_1d(a2)
+    if a1.shape[0] < a2.shape[0]:
+        a1, a2 = a2, a1
+    a2 = cupy.pad(a2, (a1.shape[0] - a2.shape[0], 0))
+    val = a1 + a2
+    if truepoly:
+        val = poly1d(val)
```
asi1024, line 34: `trimseq` in `poly1d` instantiation requires device synchronization. So, using `polyadd` for `poly1d` inputs is much slower than for `ndarray`. Why not implement lazy evaluation for `coeffs` and `order`?

cupy/lib/polynomial.pyx
```diff
@@ -4,6 +4,37 @@ import cupy
...
+        truepoly = True
+    if isinstance(a2, poly1d):
+        a2 = a2.coeffs
+        truepoly = True
+    a1 = cupy.atleast_1d(a1)
+    a2 = cupy.atleast_1d(a2)
+    if a1.shape[0] < a2.shape[0]:
+        a1, a2 = a2, a1
+    a2 = cupy.pad(a2, (a1.shape[0] - a2.shape[0], 0))
+    val = a1 + a2
```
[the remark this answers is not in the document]
Dahlia-Chehata: "+=" fails with `TestPolyaddNdimShapeCombination` tests, `test_polyadd_diff_types_array` and `test_polyadd_diff_types_poly1d`

cupy/lib/polynomial.pyx
Dahlia-Chehata: Can you elaborate a bit?

asi1024: Could you split the PR for this fix?

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
asi1024: If so how about custom `ElementwiseKernel` to reduce memory usage?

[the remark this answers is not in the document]
Dahlia-Chehata: `test_polyadd_diff_types_array` and  `test_polyadd_diff_types_poly1d` can pass using `result_type`  without needing a kernel. As for `TestPolyaddNdimShapeCombination` tests, I don't think a kernel will help. Please correct me if I'm wrong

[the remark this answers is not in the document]
asi1024: Please avoid allocating memory for padded `a2`.

cupy/lib/polynomial.pyx
asi1024: Please fix not to avoid synchronize devices.

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
asi1024: Please address this comment.

cupy/lib/polynomial.pyx
Dahlia-Chehata: Please elaborate on your comment. I don't think I get your point.

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
Dahlia-Chehata: I suggest removing the multidimensional inputs support ( **tests** to be precise) for 2 reasons:

1- Even if we guarantee the same behaviour as Numpy's, we won't be able to document it. Numpy states clearly that _Each input must be either a poly1d object or a 1D sequence of polynomial coefficients_ https://numpy.org/doc/stable/reference/generated/numpy.polyadd.html
2- It will solve the problem of memory allocation with no complex implementation. And `+=` can easily be used.

@asi1024 What do you think ?

cupy/lib/polynomial.pyx
asi1024: Please fix to access `_coeffs` and fix tests to check `polyadd` does not synchronize devices.

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
asi1024: It's OK to remove multidimensional input support.

tests/cupy_tests/lib_tests/test_polynomial.py
```diff
@@ -272,3 +304,97 @@ def test_poly1d_ne1(self, xp, dtype):
...
+
+@testing.gpu
+class TestPolyadd(unittest.TestCase):
+
+    @testing.numpy_cupy_array_equal()
+    def test_polyadd_list(self, xp):
+        a = [1, 2, 3]
+        b = [4, 2]
+        with cupyx.allow_synchronize(False):
+            return xp.polyadd(a, b)
```
asi1024, line 333: `polyadd` should not accept Python primitive list to avoid implicit conversion.

tests/cupy_tests/lib_tests/test_polynomial.py
```diff
@@ -272,3 +304,97 @@ def test_poly1d_ne1(self, xp, dtype):
...
+    @testing.numpy_cupy_array_equal()
+    def test_polyadd(self, xp, dtype):
+        a = testing.shaped_arange(self.shape1, xp, dtype)
+        b = testing.shaped_arange(self.shape2, xp, dtype)
+        with cupyx.allow_synchronize(False):
+            return xp.polyadd(a, b)
+
+
+@testing.gpu
+class TestPolyadd(unittest.TestCase):
```
asi1024, line 326: Why did you split the tests into `TestPoly1dArithmetic` and `TestPolyAdd`?

cupy/lib/polynomial.pyx
```diff
@@ -4,6 +4,38 @@ import cupy
...
+    Returns:
+        cupy.ndarray or cupy.poly1d: The sum of the inputs.
+
+    .. seealso:: :func:`numpy.polyadd`
+
+    """
+    truepoly = False
+    if isinstance(a1, poly1d):
+        a1 = a1._coeffs
+        truepoly = True
```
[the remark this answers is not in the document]
Dahlia-Chehata: What about `truepoly `. Should I let `_get_coeffs(x)` return it too ? or evaluate it in polyadd 
truepoly = isinstance(a1, poly1d) or isinstance(a2, poly1d)

tests/cupy_tests/lib_tests/test_polynomial.py
Dahlia-Chehata: I didn't split tests. `TestPoly1dArithmetic` class tests __add__  and `TestPolyAdd` class tests `polyadd`. I know that they are the same in the `addition` case apart from the fact that I call `polyadd` directly or use `+`. Would you prefer to eliminate `TestPoly1dArithmetic` ?

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
Dahlia-Chehata: I don't think we need this  `return cupy.full((1,), x)` since `cupy.atleast_1d()` already exists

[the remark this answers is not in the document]
asi1024: I pushed a commit for the refactoring. Please take a look.

tests/cupy_tests/lib_tests/test_polynomial.py
asi1024: OK. Let's leave `TestPoly1dArithmetic` without eliminating.

asi1024: For the refactoring of the tests, let's organize which tests are needed.
- The type of inputs/outputs
  - 16 types of input combination: (poly1d, array, numpy scalar, or python scalar) v.s. (same as on the left)
  - should be placed on `TestPoly1dArithmetic`.
- Device synchronization
  - should be checked at the same time as the type checking.
- Shape combinations
- Dtype combinations
- Leading zeros
  - I think we don't need to check it because `polyadd` does not trim the leading zeros. It is already tested in `TestPoly1d`.

Dahlia-Chehata: Concerning the 16 combinations, two of them fail

1. numpy scalar + `poly1d` because of Implicit conversion to NumPy array
2. array + `poly1d` because `poly1d` type is not supported in `ndarray.__add__`

Dahlia-Chehata: `test_polyadd_ndim` is not in the list. So will be eliminated right?

asi1024: Yes, it's OK to remove because NumPy does not guarantee to raise an error for such inputs.

tests/cupy_tests/lib_tests/test_polynomial.py
```diff
@@ -231,6 +235,123 @@ def test_poly1d_div_scalar(self, xp, dtype):
...
+            return dtype(5) + dtype(10).item()
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_poly1d_add_numpy_scalar_array(self, xp, dtype):
+        a = testing.shaped_arange((5,), xp, dtype)
+        with cupyx.allow_synchronize(False):
+            return dtype(5) + a
+
+    @testing.for_all_dtypes(no_bool=True)
```
asi1024, line 257: Why `no_bool` is needed?
```suggestion
    @testing.for_all_dtypes(no_bool=True)
```

Dahlia-Chehata: The test fails for numpy.bool_

cupy/lib/polynomial.pyx
```diff
@@ -123,13 +126,11 @@ cdef class poly1d:
...
-    # TODO(Dahlia-Chehata): implement using polyadd
     def __add__(self, other):
-        raise NotImplementedError
-
-    # TODO(Dahlia-Chehata): implement using polyadd
-    def __radd__(self, other):
-        raise NotImplementedError
+        if isinstance(self, numpy.generic):
+            # for the case: numpy scalar + poly1d
+            return self + other._coeffs
```
Dahlia-Chehata, line 132:
```suggestion
            return self + other.coeffs
```
To allow `bool` types, device synchronization is to be disabled

asi1024: Why device synchronization is required only for bool type inputs?

Dahlia-Chehata: It is not only for bool inputs. I meant to reply for why no_bool is needed in this particular test which BTW was irrelevant for other types. I have changed it. PTAL

Dahlia-Chehata:
@asi1024 PTAL

cupy/lib/polynomial.pyx
```diff
@@ -123,13 +126,11 @@ cdef class poly1d:
...
-    # TODO(Dahlia-Chehata): implement using polyadd
     def __add__(self, other):
-        raise NotImplementedError
-
-    # TODO(Dahlia-Chehata): implement using polyadd
-    def __radd__(self, other):
-        raise NotImplementedError
+        if isinstance(self, numpy.generic):
+            # for the case: numpy scalar + poly1d
+            raise TypeError('Invalid inputs')
```
asi1024, line 132: Write a more clear message.

Dahlia-Chehata: How about 'Numpy scalar and poly1d addition is not supported' ?

cupy/lib/polynomial.pyx
```diff
@@ -123,13 +126,12 @@ cdef class poly1d:
             return poly1d(self.coeffs * other)
         raise NotImplementedError
 
-    # TODO(Dahlia-Chehata): implement using polyadd
     def __add__(self, other):
```
asi1024, line 129: If a user-defined array object whose `__array_priority__` is greater than 90, its `__radd__` should be called. Could you fix like as [`cupy.ndarray.__add__` is implemented](https://github.com/cupy/cupy/blob/v7/cupy/core/core.pyx#L965), and add a test?

Dahlia-Chehata: Won't that affect `test_poly1d_add_poly1d_array` ?

asi1024: Yes. it affects.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

cupy/testing/helper.py
```diff
@@ -287,9 +288,16 @@ def _convert_output_to_ndarray(c_out, n_out, sp_name):
...
+        # poly1d output case.
+        assert c_out.variable == n_out.variable
+        return c_out.coeffs, n_out.coeffs
     if isinstance(c_out, numpy.generic) and isinstance(n_out, numpy.generic):
-        # scalar output case.
+        # numpy scalar output case.
         return c_out, n_out
+    if numpy.isscalar(c_out) and numpy.isscalar(n_out):
+        # python scalar output case.
+        return cupy.array(c_out), numpy.array(c_out)
```
takagi, line 300: typo?
```suggestion
        return cupy.array(c_out), numpy.array(n_out)
```

asi1024: Good catch! Thanks!

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

Issue #3522, opened 2020-07-01
Support batched slogdet


yoshipon:
Thank you for developing CuPy, which helps my signal processing research very much.

The current `slogdet` is implemented to compute one-by-one for each matrix.
I already have a personal code that calculates `slogdet` in a batched manner by using `sgetr*Batched`.
Can I make PR about this extension?

yoshipon:
I'm sorry, I was looking old version (7.x) of CuPy.
I realized that the current version already has a batched `slogdet` with cuSolver instead of `sgetr*Batched`

The current implementation, however, seems only supports matrices with real numbers.
`sgetr*Batched` can handle complex numbers in addition to real numbers.

emcastillo:
Feel free to send a PR using your approach for complex numbers 😄

yoshipon:
Thank you :)

I also found that the problem is only in the `slogdet` function and `_lu_factor`, which is called in the function, already supports complex numbers with `sgetr*Batched`.
I'll try to make a PR to fix it.

Issue #589, opened 2017-10-08
__array__ and __array_wrap__


kohr-h:
A question about the Numpy array interface. `cupy.ndarray` implements `__array__`, but  it doesn't work since `np.asarray(cupy_array)` raises an exception:
```
ValueError: object __array__ method not producing an array
```
So unless you have other plans for this method, it should be changed to return a Numpy array.

Do you want a PR? I'd also like to have the `__array_wrap_` method, so I'd implement that one as well.

niboshi:
I'm sorry for the late reply.

I agree with you that it would be useful if CuPy arrays worked just like NumPy arrays. However, currently we intentionally do not allow that because that would incur hidden performance hit due to transfer between device and host.

As this feature has been requested from multiple users, however, we're currently considering some other way to accomplish this, by using configuration context (like `chainer.using_config` in Chainer), for example.

kohr-h:
Thanks for your reply. So I read too much into the `TODO` comment in the file :-)

Anyway, for now I intend to use CuPy as a drop-in replacement of NumPy, independent of Chainer. There it would be good to have interoperability with NumPy arrays since it would make a number of implementations simpler. Avoiding data transfers between host and device would then be the job of the calling code.

Would there be a way to achieve this without involving Chainer? For example, by switching the feature on by default and disabling it by default in Chainer, with a config option there to enable it again.

niboshi:
I understand your point. But if NumPy and CuPy arrays were implicitly convertible, users would have to be very careful about which arrays they are dealing with. If a single NumPy array were involved in a series of CuPy operations by mistake, it would result in synchronization of all operations without notice. It would be very difficult to find the bottleneck.

As for the alternative method, I didn't mean CuPy would depend on Chainer. I meant introducing similar configuration mechanism used in Chainer.

kohr-h:
It seems I understood your comment wrong. A config option handling *as it is done in Chainer* would be perfectly fine.

I also agree that for your primary use case in Chainer, hidden bottlenecks would be very bad, and that you want to avoid automatic CPU<->GPU transfers. I'll have to think a bit more about my use case, and if I should also avoid mixing up CPU and GPU memory/code.

For the record: Since Numpy 1.13 the `__array_ufunc__` interface can be used to intercept ufunc calls like `np.sin` or `np.add.reduce`. In particular one can set `__array_ufunc__ = None` to signal that it is not supposed to be implemented.

okuta:
[...]

PR #3551, opened 2020-07-05
Support batched slogdet with complex numbers

yoshipon:
Related to #3522, I added support for `cupy.linalg.slogdet` to handle complex numbers.
This PR also includes a bug fix on `cupy.linalg.decomposition._lu_factor` that incorrectly handled complex ndarrays with cuSolver.

### DIFF
--- a/cupy/linalg/norms.py
+++ b/cupy/linalg/norms.py
@@ -241,12 +241,7 @@ def slogdet(a):
     util._assert_nd_squareness(a)
 
     dtype = numpy.promote_types(a.dtype.char, 'f')
-    if dtype == numpy.complex64:
-        real_dtype = numpy.dtype("f")
-    elif dtype == numpy.complex128:
-        real_dtype = numpy.dtype("d")
-    else:
-        real_dtype = dtype
+    real_dtype = numpy.dtype(dtype.char.lower())
 
     if dtype not in (numpy.float32, numpy.float64,
                      numpy.complex64, numpy.complex128):
--- a/cupy/linalg/norms.py
+++ b/cupy/linalg/norms.py
@@ -275,17 +275,14 @@ def slogdet(a):
 
     logdet = cupy.log(cupy.abs(diag)).sum(axis=-1)
 
+    # ipiv is 1-origin
+    non_zero = cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1)
     if dtype.kind == "f":
-        # ipiv is 1-origin
-        non_zero = (cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1) +
-                    cupy.count_nonzero(diag < 0, axis=-1))
-    
-        # Note: sign == -1 ** (non_zero % 2)
-        sign = (non_zero % 2) * -2 + 1
-    else:  # for complex numbers
-        non_zero = cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1)
-        
-        sign = (non_zero % 2) * -2 + 1
+        non_zero += cupy.count_nonzero(diag < 0, axis=-1)
+
+    # Note: sign == -1 ** (non_zero % 2)
+    sign = (non_zero % 2) * -2 + 1
+    if dtype.kind == "c":
         sign = sign * cupy.prod(diag, axis=-1) / cupy.exp(logdet)
 
     singular = dev_info > 0
--- a/cupy/linalg/norms.py
+++ b/cupy/linalg/norms.py
@@ -282,7 +282,7 @@ def slogdet(a):
     # Note: sign == -1 ** (non_zero % 2)
     sign = (non_zero % 2) * -2 + 1
     if dtype.kind == "c":
-        sign = sign * cupy.prod(diag, axis=-1) / cupy.exp(logdet)
+        sign = sign * cupy.prod(diag / cupy.abs(diag), axis=-1)
 
     singular = dev_info > 0
     return (
--- a/cupy/linalg/norms.py
+++ b/cupy/linalg/norms.py
@@ -241,11 +241,16 @@ def slogdet(a):
     util._assert_nd_squareness(a)
 
     dtype = numpy.promote_types(a.dtype.char, 'f')
-    real_dtype = dtype
+    print(type(dtype))
+    if dtype == numpy.complex64:
+        real_dtype = numpy.dtype("f")
+    elif dtype == numpy.complex128:
+        real_dtype = numpy.dtype("d")
+    else:
+        real_dtype = dtype
 
-    # TODO(kataoka): support complex types
-    if dtype not in (numpy.float32, numpy.float64):
-        msg = ('dtype must be float32 or float64'
+    if dtype not in ("f", "d", "F", "D"):
+        msg = ('dtype must be float32, float64, complex64, or complex128'
                ' (actual: {})'.format(a.dtype))
         raise ValueError(msg)
 
@@ -268,13 +273,20 @@ def slogdet(a):
 
     diag = cupy.diagonal(lu, axis1=-2, axis2=-1)
 
-    # ipiv is 1-origin
-    non_zero = (cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1) +
-                cupy.count_nonzero(diag < 0, axis=-1))
+    logdet = cupy.log(cupy.abs(diag)).sum(axis=-1)
 
-    # Note: sign == -1 ** (non_zero % 2)
-    sign = (non_zero % 2) * -2 + 1
-    logdet = cupy.log(abs(diag)).sum(axis=-1)
+    if dtype.kind == "f":
+        # ipiv is 1-origin
+        non_zero = (cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1) +
+                    cupy.count_nonzero(diag < 0, axis=-1))
+    
+        # Note: sign == -1 ** (non_zero % 2)
+        sign = (non_zero % 2) * -2 + 1
+    else:  # for complex numbers
+        non_zero = cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1)
+        
+        sign = (non_zero % 2) * -2 + 1
+        sign = sign * cupy.prod(diag, axis=-1) / cupy.exp(logdet)
 
     singular = dev_info > 0
     return (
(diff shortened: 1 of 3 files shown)


emcastillo:
Can you measure performance before and after the change?
Thank you!

cupy/linalg/norms.py
```diff
@@ -268,13 +272,17 @@ def slogdet(a):
...
-                cupy.count_nonzero(diag < 0, axis=-1))
+    non_zero = cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1)
+    if dtype.kind == "f":
+        non_zero += cupy.count_nonzero(diag < 0, axis=-1)
 
     # Note: sign == -1 ** (non_zero % 2)
     sign = (non_zero % 2) * -2 + 1
-    logdet = cupy.log(abs(diag)).sum(axis=-1)
+    if dtype.kind == "c":
+        sign = sign * cupy.prod(diag, axis=-1) / cupy.exp(logdet)
```
emcastillo, line 285: Changing `logdet` and `sign` to use

```
det = cupy.abs(diag).prod(axis=-1)
sign = sign * cupy.prod(diag, axis=-1) / det
logdet = cupy.log(det)
```

Results in faster execution times :)

emcastillo: However,  `cupy.prod(diag, axis=-1)` might produce several precision errors. We should look for a way to make the `sign = sign * cupy.prod(diag, axis=-1) / det` expression more numerically stable

emcastillo, line 285: `cupy.prod(diag, axis=-1)` here might produce several numerical errors for large matrices resulting in inf or error. I wonder if just `cupy.prod(cupy.sign(diag), axis=-1))` is better and allows you to get rid of the `cupy.exp` division.

yoshipon:
Thank you for your comments and nice suggestion m(_ _)m
I replaced with `cupy.prod(diag / cupy.abs(diag), axis=-1)` because cupy's (and also numpy's) sign function does not support complex numbers.

> You will need to tweak 0 values somehow :)

Actually, this is already done at the last of 'slogdet' function by evaluating whether each matrix is singular or not.
https://github.com/cupy/cupy/blob/a810f1aa908748d030da60946e11106a1f440f3f/cupy/linalg/norms.py#L289

emcastillo:
Great fix! Thanks a lot 😃

emcastillo:
Jenkins, test this please

yoshipon:
@emcastillo If you have any idea about the failure of the Jenkins CI test, could you give me some advice?

I'm now confused because the test seems to succeed in my local environment as follows:
```
$ python -m pytest tests/cupy_tests/linalg_tests/test_norms.py
========================================================= test session starts =========================================================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: [modified: path/to/workspace]/cupy, inifile: setup.cfg
collected 237 items                                                                                                                   

tests/cupy_tests/linalg_tests/test_norms.py ................................................................................... [ 35%]
............................................................................................................................... [ 88%]
...........................                                                                                                     [100%]

========================================================== warnings summary ===========================================================
tests/cupy_tests/linalg_tests/test_norms.py: 14 tests with warnings
  [modified: /path/to/anaconda]/envs/dev-cupy/lib/python3.8/site-packages/numpy/linalg/linalg.py:2514: RuntimeWarning: divide by zero encountered in power
    absx **= ord

tests/cupy_tests/linalg_tests/test_norms.py: 14 tests with warnings
  [modified: /path/to/anaconda]/envs/dev-cupy/lib/python3.8/site-packages/numpy/linalg/linalg.py:2514: RuntimeWarning: divide by zero encountered in reciprocal
    absx **= ord

-- Docs: https://docs.pytest.org/en/latest/warnings.html
================================================== 237 passed, 28 warnings in 3.51s ===================================================
```
I confirmed that the source code is the version on the branch `dev-slogdet`.

cupy/linalg/norms.py
```diff
@@ -268,13 +272,17 @@ def slogdet(a):
...
-                cupy.count_nonzero(diag < 0, axis=-1))
+    non_zero = cupy.count_nonzero(ipiv != cupy.arange(1, n + 1), axis=-1)
+    if dtype.kind == "f":
+        non_zero += cupy.count_nonzero(diag < 0, axis=-1)
 
     # Note: sign == -1 ** (non_zero % 2)
     sign = (non_zero % 2) * -2 + 1
-    logdet = cupy.log(abs(diag)).sum(axis=-1)
+    if dtype.kind == "c":
+        sign = sign * cupy.prod(diag / cupy.abs(diag), axis=-1)
```
leofang, line 285: You can also do this in-place:
```suggestion
        sign *= cupy.prod(diag / cupy.abs(diag), axis=-1)
```

cupy/linalg/norms.py
```diff
@@ -241,11 +241,15 @@ def slogdet(a):
     util._assert_nd_squareness(a)
 
     dtype = numpy.promote_types(a.dtype.char, 'f')
-    real_dtype = dtype
+    if dtype == numpy.complex64:
+        real_dtype = numpy.dtype("f")
```
leofang, line 245: I suppose if you have already known the dtype, just use the explicit form:
```suggestion
        real_dtype = numpy.float32
```

cupy/linalg/norms.py
```diff
@@ -241,11 +241,15 @@ def slogdet(a):
...
+    elif dtype == numpy.complex128:
+        real_dtype = numpy.dtype("d")
```
leofang, line 247: ditto
```suggestion
        real_dtype = numpy.float64
```

cupy/linalg/norms.py
```diff
@@ -241,11 +241,15 @@ def slogdet(a):
...
+    else:
+        real_dtype = dtype
 
-    # TODO(kataoka): support complex types
-    if dtype not in (numpy.float32, numpy.float64):
-        msg = ('dtype must be float32 or float64'
+    if dtype not in ("f", "d", "F", "D"):
```
leofang, line 251: Isn't `dtype` a dtype object instead of `str`?!
```suggestion
    if dtype not in (numpy.float32, numpy.float64,
                     numpy.complex64, numpy.complex128):
```

cupy/linalg/norms.py
yoshipon: Thank you for your suggestions! :)

Unfortunately, this operation cannot be done because the `cupy.prod(...)` has complex numbers and the original `sign` contains real numbers.

cupy/linalg/norms.py
asi1024: How about writing as follows?
```py
real_dtype = numpy.dtype(dtype.char.lower())
```

leofang: Ah, it's better. No more branching!

cupy/linalg/norms.py
leofang: I see, thanks!

emcastillo:
The jenkins errors

```
00:05:18 E   AssertionError: Only cupy raises error
00:05:18 E   
00:05:18 E     File "/work/cupy/cupy/testing/helper.py", line 23, in _call_func
00:05:18 E       result = impl(self, *args, **kw)
00:05:18 E     File "/work/cupy/tests/cupy_tests/linalg_tests/test_norms.py", line 179, in test_slogdet_3
00:05:18 E       return xp.array([sign, logdet], dtype)
00:05:18 E     File "/work/cupy/cupy/creation/from_data.py", line 43, in array
00:05:18 E       return core.array(obj, dtype, copy, order, subok, ndmin)
00:05:18 E     File "cupy/core/core.pyx", line 1986, in cupy.core.core.array
00:05:18 E       cpdef ndarray array(obj, dtype=None, bint copy=True, order='K',
00:05:18 E     File "cupy/core/core.pyx", line 2065, in cupy.core.core.array
00:05:18 E       return _send_object_to_gpu(obj, dtype, order, ndmin)
00:05:18 E     File "cupy/core/core.pyx", line 2132, in cupy.core.core._send_object_to_gpu
00:05:18 E       a_cpu = numpy.array(obj, dtype=dtype, copy=False, order=order,
00:05:18 E     File "cupy/core/core.pyx", line 1171, in cupy.core.core.ndarray.__array__
00:05:18 E       raise TypeError(
00:05:18 E   Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.
```

emcastillo:
I tracked this error to be caused by numpy 1.19.

cupy/linalg/norms.py
yoshipon: Thank you!

yoshipon:
@emcastillo Thank you for tracking.
I also reproduced this error by installing `numpy=1.19.0`.

The reason was that the dtype of `sign` and `logdet` were different when the tested dtype was complex.
I fixed the bug by putting `logdet.astype(sign.dtype)`.
https://github.com/cupy/cupy/blob/29fcbcdd364d20533974131529944cffc4b0b8d4/tests/cupy_tests/linalg_tests/test_norms.py#L172

emcastillo:
Jenkins, test this please

leofang:
Do I get it right that NumPy 1.19 changed the returned types?

yoshipon:
I'm not sure, but this might be related to https://github.com/cupy/cupy/issues/589#issuecomment-644616962.

asi1024:
`cupy.array` seems to unexpected raise `TypeError` when the list of `cupy.ndarray` of different dtypes.
```
>>> numpy.array([cupy.zeros((2, 3), dtype='f'), cupy.zeros((2, 3), dtype='d')])
  File "cupy/core/core.pyx", line 1171, in cupy.core.core.ndarray.__array__
    raise TypeError(
```

yoshipon:
This seems correct because cupy does not allow the implicit conversion of `cupy.ndarray` to `numpy.ndarray`.
The main problem seems that the `cupy.array` function does not regard the list of `cupy.ndarray`s as a type of `cupy.ndarray`.
    if isinstance(obj, ndarray):

yoshipon:
@asi1024 This seems no problem because cupy does not allow the implicit conversion of `cupy.ndarray` to `numpy.ndarray`.

The main problem seems that the `cupy.array` function regards the list of `cupy.ndarray`s as something like `numpy.ndarray`:
https://github.com/cupy/cupy/blob/989ef4cd8889ea5309af288beba6814584969930/cupy/core/core.pyx#L2065

I'm not sure this is the correct behavior or not.

asi1024:
BTW, the tests don't check the dtypes of the return values. Fixed in #3577.

yoshipon:
Oh, thank you very much.

Issue #3556, opened 2020-07-06
`cupy.fft.irfft` is slow


asi1024:
`cupy.fft.irfft` is extremely slow and causes unexpected device synchronization.
```
fft     :    CPU:  511.504 us   +/-182.837 (min:  466.546 / max: 6098.848) us     GPU-0:  914.825 us   +/-181.750 (min:  868.352 / max: 6491.872) us
ifft    :    CPU:  516.756 us   +/-89.689 (min:  499.988 / max: 3025.299) us     GPU-0:  993.364 us   +/-83.566 (min:  970.336 / max: 3351.712) us
rfft    :    CPU:  473.139 us   +/-12.516 (min:  462.383 / max:  624.113) us     GPU-0:  711.520 us   +/-12.456 (min:  699.104 / max:  857.184) us
irfft   :    CPU:483281.932 us   +/-5010.430 (min:479695.227 / max:490367.569) us     GPU-0:483321.136 us   +/-5021.194 (min:479726.959 / max:490421.997) us
```

Reproducer:
```py
import cupy
import cupyx


a = cupy.random.rand(5000000).astype('f')

funcs = [
    cupy.fft.fft,
    cupy.fft.ifft,
    cupy.fft.rfft,
    cupy.fft.irfft,
]

for f in funcs:
    perf = cupyx.time.repeat(f, (a,), n_warmup=5, max_duration=1)
    print(perf)
```
cc: @leofang

emcastillo:
Sync is caused by
https://github.com/cupy/cupy/blob/master/cupy/fft/fft.py#L103

Only for irfft

emcastillo:
This line is synching the device
https://github.com/cupy/cupy/blob/master/cupy/cuda/cufft.pyx#L283-L284

It just calls 
https://docs.nvidia.com/cuda/cufft/index.html#function-cufftmakeplan1d with type as `CUFFT_C2R`
@anaruse @leofang is this expected behavior? plan creation can synchronize the device in this situation?

grlee77:
I think this is just caused by the chosen size of 5,000,000 being particularly unfavorable for `irfft`. Size 5,000,000 would result from a forward `rfft` of a signal of size 9,999,999 which is similarly fairly slow (~100 ms when I timed it).

If I take the irfft of a signal of size 5,000,001 or 2,500,001 for example, I get times around 1 ms which is more in line with the results of the other cases above.

grlee77:
[...]

PR #3560, opened 2020-07-07
Improve `_fft_convolve`

asi1024:
Rel. #3556.
Blocked by #3571.

Previous:
```
convolve            :    CPU: 1598.203 us   +/-196.056 (min: 1553.705 / max: 5673.330) us     GPU-0: 1607.828 us   +/-198.819 (min: 1562.784 / max: 5725.760) us
convolve            :    CPU: 7661.681 us   +/-683.048 (min: 7492.259 / max:14718.648) us     GPU-0: 7673.338 us   +/-684.819 (min: 7503.136 / max:14733.984) us
convolve            :    CPU: 7556.341 us   +/-45.591 (min: 7495.102 / max: 7804.286) us     GPU-0: 7567.647 us   +/-45.571 (min: 7506.176 / max: 7815.456) us
convolve            :    CPU: 9845.364 us   +/-53.787 (min: 9777.471 / max:10081.744) us     GPU-0: 9856.696 us   +/-53.819 (min: 9788.608 / max:10093.184) us
```
This PR (https://github.com/cupy/cupy/pull/3560/commits/b570a128a6bf1a76136fedc5ba2b7abcc763f43e):
```
convolve            :    CPU: 1428.156 us   +/-157.454 (min: 1391.433 / max: 5490.528) us     GPU-0: 1435.945 us   +/-158.645 (min: 1399.072 / max: 5529.664) us
convolve            :    CPU: 1511.592 us   +/-144.757 (min: 1472.319 / max: 4345.249) us     GPU-0: 1519.854 us   +/-147.089 (min: 1480.000 / max: 4380.480) us
convolve            :    CPU: 1505.722 us   +/-32.617 (min: 1474.274 / max: 1744.067) us     GPU-0: 1513.800 us   +/-32.601 (min: 1482.432 / max: 1752.128) us
convolve            :    CPU: 1509.725 us   +/-72.535 (min: 1474.584 / max: 3164.543) us     GPU-0: 1518.026 us   +/-75.407 (min: 1482.656 / max: 3254.400) us
```

benchmark script:
```py
import numpy

import cupy
import cupyx

for a_len in [10, 200000]:
    for v_len in [10, 200000]:
        a = cupy.random.rand(a_len)
        v = cupy.random.rand(v_len)
        perf = cupyx.time.repeat(
            cupy.convolve, (a, v, 'same'),
            n_warmup=5, max_duration=1)
        print(perf)
```


emcastillo:
Jenkins, test this please

emcastillo:
The PR pads the size to be the closest power of 2 so the fft algorithms can work more efficiently.

asi1024:
Jenkins, test this please.

cupy/math/misc.py
```diff
@@ -65,13 +65,13 @@ def _fft_convolve(a1, a2, mode):
 
     dtype = cupy.result_type(a1, a2)
```
leofang, line 66: What if we upcast `a1` and `a2` to the same `dtype` before calling fft?

leofang: I guess it would not matter, as only the non-contiguous test failed...

asi1024:
`cupy-py35` which uses cuFFT 10.1.2 has passed, and `cupy-py3` which uses cuFFT 10.0.0 has failed.

grlee77:
@emcastillo 
Good idea to update the size to a favorable one.

Note, that you don't necessarily need to increase to a power of two, but could instead go to the next smallest number that has only 2, 3, 5, and 7 as its prime factors (ref: https://docs.nvidia.com/cuda/cufft/index.html#accuracy-and-performance). 

SciPy has a utility function for this ([scipy.fft.next_fast_len](https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.next_fast_len.html)), which has a docstring example where the next power of two makes the signal unnecessarily large. pocketfft's prime factors are slightly different from cuFFT's factors: they are {2, 3, 5, 7, 11} for complex-to-complex transforms, but {2, 3, 5} for transforms involving real input or output. The latter would work fine for cuFFT (just neglecting some potential cases involving factors of 7)

Specifically, if scipy >= 1.4 is available, one can call

```Python
fast_size = scipy.fft.next_fast_len(size, real=True)
```
For older SciPy, FFTPACK only supported factors (2, 3, 5) so
```Python
fast_size = scipy.fftpack.next_fast_len(size)
```
would work instead.

If you really wanted to make a cuFFT-dedicated C++ function also supporting factor 7 you could start from the C++ ones in pocketfft as here:
https://github.com/scipy/scipy/blob/3f11e32f3d07e6035b46f77eafec4359875a0416/scipy/fft/_pocketfft/pocketfft_hdronly.h#L411-L436

leofang:
> If you really wanted to make a cuFFT-dedicated C++ function also supporting factor 7 you could start from the C++ ones in pocketfft 

I once thought about adding `cupyx.scipy.fft*.next_fast_len()`, but I didn't have strong motivation (until now seeing this). But, due to the difference between pocketfft and cuFFT, the returned values won't be identical. Not sure if this is acceptable.

grlee77:
> But, due to the difference between pocketfft and cuFFT, the returned values won't be identical. Not sure if this is acceptable.

I would personally find it acceptable, as it is conceptually the same: "find me the next smallest length that will not have poor performance in the underlying FFT library". Particularly in the context of a round-trip transform as for FFT-based convolution, I don't think the user will care what intermediate size was used internally.

leofang:
Hi @grlee77 You know I almost always agree with you on performance considerations 😁 I was refreshing my memory on SciPy's implementation: https://github.com/scipy/scipy/blob/3f11e32f3d07e6035b46f77eafec4359875a0416/scipy/fft/_helper.py#L6
and your suggestion

> (we can always add a note: "Due to differences in the underlying FFT libraries used by CuPy and SciPy, the length returned will not always be identical to scipy.fft.next_fast_len.")

brings up a question: with the recent backend support in `scipy.fft`, perhaps `scipy.fft.next_fast_len()` should also be registered with a backend? Ideally, we want this usage
```python
import scipy.fft as spfft
import cupyx.scipy.fft as cufft

spfft.set_global_backend(cufft)
n = spfft.next_fast_len(100)
```
to correctly return the number for cuFFT, not for pocketfft.

leofang:
@asi1024 @grlee77 I don't think I fully understand this, but here you go:
```python 
cpdef size_t next_fast_len(size_t n):
    """2^a * 3^b * 5^c * 7^d for cuFFT."""

    if n <= 8:
        return n

    cdef size_t b = 1, c = 1, d = 1, best = 2 * n

    while d < best:
        c = d
        while c < best:
            b = c
            while b < n:
                b *= 2
            while True:
                if b < n:
                    b *= 3
                elif b > n:
                    if b < best:
                        best = b
                    if b & 1:
                        break
                    b >>= 1
                else:
                    return n
            c *= 5
        d *= 7
    return best
```

grlee77:
> @asi1024 @grlee77 How about this? This is basically what SciPy does, tailored to primes (2, 3, 5, 7).

Yes, this looks correct. I empirically verified via comparison to `scipy.fft.next_fast_len` by checking that when there are factors of 11 the scipy result is smaller than this one, otherwise they are the same. details below (with `prime_factors` function taken from this [stackoverflow post](https://stackoverflow.com/questions/15347174/python-finding-prime-factors))

<details>

```Python
def prime_factors(n):
    i = 2
    factors = []
    while i * i <= n:
        if n % i:
            i += 1
        else:
            n //= i
            factors.append(i)
    if n > 1:
        factors.append(n)
    return factors


def next_fast_len(n):
    """2^a * 3^b * 5^c * 7^d for cuFFT."""
    if n <= 8:
        return n
    b = 1
    c = 1
    d = 1
    best = 2 * n
    while d < best:
        c = d
        while c < best:
            b = c
            while b < n:
                b *= 2
            while True:
                if b < n:
                    b *= 3
                elif b > n:
                    if b < best:
                        best = b
                    if b & 1:
                        break
                    b >>= 1
                else:
                    return n
            c *= 5
        d *= 7
    return best


for n in range(1024):
    n_fast_scipy = scipy.fft.next_fast_len(a)
    n_fast = next_fast_len(a)
    if 11 in prime_factors(n_fast_scipy):
        assert n_fast_scipy < n_fast
    else:
        assert n_fast_scipy == n_fast
```

</details>

leofang:
Awesome! Thanks for confirming, @grlee77. 

@asi1024, would you like to include `next_fast_len()` in this PR, or should I send a new one? Note that it better be implemented in C or Cython, because it'd be way faster:
```python
In [3]: %timeit nfl_cython(100000000000)                                                                                                                             
489 ns ± 6.36 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit nfl_python(100000000000)                                                                                                                               
53.4 µs ± 787 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [6]: %timeit nfl_cython(1000000)                                                                                                                                    
176 ns ± 0.897 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: %timeit nfl_python(1000000)                                                                                                                                    
16.2 µs ± 152 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

asi1024:
PTAL.

leofang:
@asi1024 Could you rerun your benchmark and see it's better? 😇

asi1024:
Should we fix to 1. skip tests if `cupy.cuda.cufft.getVersion() == 10000` or 2. relax tolerance to pass with any version of cuFFT?

leofang:
I am fine either way, was just wondering why we didn't see this issue in individual fft tests. 

@anaruse Do you happen to know if there's known precision issues for 1D FFT in CUDA 10.0?

asi1024:
Updated performance numbers in the PR description!

emcastillo:
I measured the dot convolution and the fft depending on the inputs a and b sizes. What I got is this

image

Y axis is a size, X axis is b size.
In green you can see the results when direct convolution is faster than fft.
This is measured in a V100 GPU so the threshold is likely to change depending on the GPU.

emcastillo:
Jenkins, test this please

leofang:
@asi1024 @grlee77 I am wondering for `C2C`  if we could further speed up this by reusing a `Plan1d` for the three fft calls

emcastillo:
Probably yes, most of the time is CPU bounded due to the plan creation. Can you do some test?

Thanks!

Issue #3545, opened 2020-07-05
Feature request: iterative solvers, bicgstab using cusparse wrappers


sunilkpai:
I would like to make a feature request or contribute to making an preconditioned iterative solver for bicgstab using cusparse wrappers (under `cupyx.scipy.sparse.linalg.bicgstab`), which is essential for large electromagnetic simulations I would like to work on. The routine would look a lot like [this](https://docs.nvidia.com/cuda/incomplete-lu-cholesky/index.html). Preconditioning may be hard to generalize to fall under the scipy `LinearOperator` template, but at the very least I can include ILU preconditioning. Let me know if this is reasonable and I can get started on a PR.

takagi:
Let me close https://github.com/cupy/cupy/pull/3569 and leave a comment to take over implementing cupyx.scipy.sparse.linalg.bicgstab. We have on-going implementation in https://github.com/cupy/cupy/pull/3569 thanks to @sunilkpai.

smarkesini:
This landed: `cupyx.scipy.sparse.linalg.bicgstab` was added in #9824 (merged 2026-04-23) and is in v14.1.0 and v14.2.0, with the scipy signature (`x0`, `rtol`, `atol`, `maxiter`, `M`, `callback`), so a preconditioner goes in as `M` (any `LinearOperator`, including an ILU from `spilu(...).solve`). The PR did not link this issue, which is presumably why it is still open. Suggest closing as done.

Issue #2749, opened 2019-12-03
Feature Request: Sparse ILU Factorization


hakeemo:
Are there plans in the near future to include Sparse ILU Factorization routines?
If this is a good first issue, I would be interested in contributing.

kmaehashi:
I think no one's working on it. We welcome your contributions 😄

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.spilu.html

PR #3569, opened 2020-07-08
[WIP] Feature/add bicgstab

sunilkpai:
Addresses #2749 and my feature request #3545 .

**Good news**
I got bicgstab  (unpreconditioned) to work and have added several cusparse routines to the library, but I am still working on this pull request (need to make several tests... ). I will need to take a small break for about a week or so. In the meantime, I would like to ensure that what I currently have implementation-wise is generally in line with what is required since there are a lot of changes in this PR. For a decently sized unpreconditioned 2d EM problem, the number of iterations is sometimes more sometimes less, and the speed per iteration is almost 5x faster on GPU. 

**Bad news**
I am also struggling to get the analysis/solve phases to work in cupy at the moment and I wonder if this has to do with an error of mine or something deep within CuPy (I get the error `CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered`) at [line 280 of solve.py](https://github.com/sunilkpai/cupy/blob/feature/add-bicgstab/cupyx/scipy/sparse/linalg/solve.py#L280). Generally, it would be good to know if what I am doing is in line with what is expected of cupy. Any help to debug this issue would be appreciated if you have the time!

### DIFF
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -99,8 +99,8 @@ _available_cusparse_version = {
     'csrilu02_analysis': (8000, None),
     'csrsv2_solve': (8000, None),
     'csrsv2_analysis': (8000, None),
-    'spmv': (10200, None),  # accuracy bugs in cuSparse 10.3.0
-    'spmm': (10301, None),
+    'spmv': (10200, None),
+    'spmm': (10301, None),  # accuracy bugs in cuSparse 10.3.0
     'csr2dense': (8000, None),
     'csc2dense': (8000, None),
     'csrsort': (8000, None),
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -411,56 +411,6 @@ def csrmm2(a, b, c=None, alpha=1.0, beta=0.0, transa=False, transb=False):
     return c
 
 
-def csrmv(a, x, y=None, alpha=1, beta=0, transa=False):
-    """Matrix-vector product for a CSR-matrix and a dense vector.
-
-    .. math::
-
-       y = \\alpha * o_a(A) x + \\beta y,
-
-    where :math:`o_a` is a transpose function when ``transa`` is ``True`` and
-    is an identity function otherwise.
-
-    Args:
-        a (cupy.cusparse.csr_matrix): Matrix A.
-        x (cupy.ndarray): Vector x.
-        y (cupy.ndarray or None): Vector y. It must be F-contiguous.
-        alpha (float): Coefficient for x.
-        beta (float): Coefficient for y.
-        transa (bool): If ``True``, transpose of ``A`` is used.
-
-    Returns:
-        cupy.ndarray: Calculated ``y``.
-
-    """
-    if not check_availability('csrsv'):
-        raise RuntimeError('csrsv is not available.')
-
-    assert y is None or y.flags.f_contiguous
-
-    a_shape = a.shape if not transa else a.shape[::-1]
-    if a_shape[1] != len(x):
-        raise ValueError('dimension mismatch')
-
-    handle = device.get_cusparse_handle()
-    m, n = a_shape
-    a, x, y = _cast_common_type(a, x, y)
-    dtype = a.dtype
-    if y is None:
-        y = cupy.zeros(m, dtype)
-    alpha = numpy.array(alpha, dtype).ctypes
-    beta = numpy.array(beta, dtype).ctypes
-
-    _call_cusparse(
-        'csrsv', dtype,
-        handle, _transpose_flag(transa),
-        a.shape[0], a.shape[1], a.nnz, alpha.data, a._descr.descriptor,
-        a.data.data.ptr, a.indptr.data.ptr, a.indices.data.ptr,
-        x.data.ptr, beta.data, y.data.ptr)
-
-    return y
-
-
 def csrgeam(a, b, alpha=1, beta=1):
     """Matrix-matrix addition.
 
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -37,6 +37,12 @@ class MatDescriptor(object):
     def set_mat_index_base(self, base):
         cusparse.setMatIndexBase(self.descriptor, base)
 
+    def set_mat_diag_type(self, typ):
+        cusparse.setMatDiagType(self.descriptor, typ)
+
+    def set_mat_fill_mode(self, mode):
+        cusparse.setMatFillMode(self.descriptor, mode)
+
 
 def _cast_common_type(*xs):
     dtypes = [x.dtype for x in xs if x is not None]
@@ -89,8 +95,12 @@ _available_cusparse_version = {
     'csrgeam2': (9020, None),
     'csrgemm': (8000, 11000),
     'csrgemm2': (8000, None),
+    'csrilu02': (8000, None),
+    'csrilu02_analysis': (8000, None),
+    'csrsv2_solve': (8000, None),
+    'csrsv2_analysis': (8000, None),
     'spmv': (10200, None),
-    'spmm': (10301, None),  # accuracy bugs in cuSparse 10.3.0
+    'spmm': (10301, None),
     'csr2dense': (8000, None),
     'csc2dense': (8000, None),
     'csrsort': (8000, None),
@@ -104,7 +114,7 @@ _available_cusparse_version = {
     'csc2csrEx2': (10200, None),  # the entity is csr2cscEx2
     'dense2csc': (8000, None),
     'dense2csr': (8000, None),
-    'csr2csr_compress': (8000, None),
+    'csr2csr_compress': (8000, None)
 }
 
 
@@ -399,6 +409,56 @@ def csrmm2(a, b, c=None, alpha=1.0, beta=0.0, transa=False, transb=False):
         a.indptr.data.ptr, a.indices.data.ptr,
         b.data.ptr, ldb, beta.data, c.data.ptr, ldc)
     return c
+
+
+def csrmv(a, x, y=None, alpha=1, beta=0, transa=False):
+    """Matrix-vector product for a CSR-matrix and a dense vector.
+
+    .. math::
+
+       y = \\alpha * o_a(A) x + \\beta y,
+
+    where :math:`o_a` is a transpose function when ``transa`` is ``True`` and
+    is an identity function otherwise.
+
+    Args:
+        a (cupy.cusparse.csr_matrix): Matrix A.
+        x (cupy.ndarray): Vector x.
+        y (cupy.ndarray or None): Vector y. It must be F-contiguous.
+        alpha (float): Coefficient for x.
+        beta (float): Coefficient for y.
+        transa (bool): If ``True``, transpose of ``A`` is used.
+
+    Returns:
+        cupy.ndarray: Calculated ``y``.
+
+    """
+    if not check_availability('csrsv'):
+        raise RuntimeError('csrsv is not available.')
+
+    assert y is None or y.flags.f_contiguous
+
+    a_shape = a.shape if not transa else a.shape[::-1]
+    if a_shape[1] != len(x):
+        raise ValueError('dimension mismatch')
+
+    handle = device.get_cusparse_handle()
+    m, n = a_shape
+    a, x, y = _cast_common_type(a, x, y)
+    dtype = a.dtype
+    if y is None:
+        y = cupy.zeros(m, dtype)
+    alpha = numpy.array(alpha, dtype).ctypes
+    beta = numpy.array(beta, dtype).ctypes
+
+    _call_cusparse(
+        'csrsv', dtype,
+        handle, _transpose_flag(transa),
+        a.shape[0], a.shape[1], a.nnz, alpha.data, a._descr.descriptor,
+        a.data.data.ptr, a.indptr.data.ptr, a.indices.data.ptr,
+        x.data.ptr, beta.data, y.data.ptr)
+
+    return y
 
 
 def csrgeam(a, b, alpha=1, beta=1):
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -99,7 +99,7 @@ _available_cusparse_version = {
     'csrilu02_analysis': (8000, None),
     'csrsv2_solve': (8000, None),
     'csrsv2_analysis': (8000, None),
-    'spmv': (10200, None),
+    'spmv': (10200, None),  # accuracy bugs in cuSparse 10.3.0
     'spmm': (10301, None),
     'csr2dense': (8000, None),
     'csc2dense': (8000, None),
@@ -114,7 +114,7 @@ _available_cusparse_version = {
     'csc2csrEx2': (10200, None),  # the entity is csr2cscEx2
     'dense2csc': (8000, None),
     'dense2csr': (8000, None),
-    'csr2csr_compress': (8000, None)
+    'csr2csr_compress': (8000, None),
 }
 
 
--- a/cupyx/scipy/sparse/linalg/solve.py
+++ b/cupyx/scipy/sparse/linalg/solve.py
@@ -307,6 +307,8 @@ def spilu(A, enable_boost: bool = True, tol=None, boost_val=None):
     out = cupy.empty(n, dtype=dtype)
 
     def M(x):
+        if x.ndim != 1 or len(x) != n:
+            raise ValueError('b must be 1-d array whose size is same as A')
         cupy.cusparse._call_cusparse('csrsv2_solve', dtype, handle, trans_L,
                                      *A_tuple_a(descr_L), info_L, x.data.ptr,
                                      y.data.ptr, policy_L, buff.data.ptr)
--- a/cupyx/scipy/sparse/linalg/solve.py
+++ b/cupyx/scipy/sparse/linalg/solve.py
@@ -68,7 +68,7 @@ def lsqr(A, b):
     return ret
 
 
-def bicgstab(A, b, x0=None, M=None, tol=1e-5, maxit=numpy.inf, callback=None):
+def bicgstab(A, b, x0=None, M=None, tol=1e-5, maxiter=numpy.inf, callback=None, atol='legacy'):
     """Solves linear system using an iterative solver BiCGSTAB.
 
     See https://docs.nvidia.com/cuda/cusparse/#csrilu02_solve and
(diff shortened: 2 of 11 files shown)


cupy/cuda/cupy_cusparse.h
```diff
@@ -411,7 +411,51 @@ cusparseStatus_t cusparseCsr2cscEx2(...) {
 // Functions deleted in cuSparse 11.0
 
 // cuSPARSE Level2 Function
-cusparseStatus_t cusparseScsrmv(...) {
```
leofang, line 414: This line is needed. Don't remove it.

cupy/cusparse.py
```diff
@@ -89,8 +95,12 @@ def _dtype_to_DataType(dtype):
     'csrgeam2': (9020, None),
     'csrgemm': (8000, 11000),
     'csrgemm2': (8000, None),
+    'csrilu02': (8000, None),
+    'csrilu02_analysis': (8000, None),
+    'csrsv2_solve': (8000, None),
+    'csrsv2_analysis': (8000, None),
     'spmv': (10200, None),
-    'spmm': (10301, None),  # accuracy bugs in cuSparse 10.3.0
```
leofang, line 93: Don't remove comments. (Maybe you were thinking there's no CUDA 10.3? Note that CUDA libraries follow different version semantics than the Toolkit itself.)

cupy/cusparse.py
```diff
@@ -104,7 +114,7 @@ def _dtype_to_DataType(dtype):
     'csc2csrEx2': (10200, None),  # the entity is csr2cscEx2
     'dense2csc': (8000, None),
     'dense2csr': (8000, None),
-    'csr2csr_compress': (8000, None),
```
leofang, line 107: It doesn't hurt to keep the comma.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -1,9 +1,9 @@
-import numpy
+import numpy as np
```
leofang, line 1: CuPy does not accept abbreviated library names, see [this section](https://docs-cupy.chainer.org/en/latest/contribution.html#coding-guidelines):
> we restrict the usage of shortcut symbols in our code base.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -1,9 +1,9 @@
...
 
-import cupy
-from cupy.cuda import cusolver
-from cupy.cuda import device
+import cupy as cp
+from cupy.cuda import cusolver, device, cusparse
```
leofang, line 4: ditto

cupyx/scipy/sparse/linalg/solve.py
leofang: Another CuPy style guide: https://github.com/cupy/cupy/wiki/Development-policies

cupy/cusparse.py
sunilkpai: This was an accident, sorry!

cupy/cuda/cupy_cusparse.h
sunilkpai: This was also an accident!

sunilkpai:
@toslunar any reason the travis build fails but not appveyor (still wip btw)?

leofang:
> any reason the travis build fails but not appveyor (still wip btw)?

@sunilkpai There're still stubs like `cusparseCreateCsrsv2Info` missing in `cupy_backends/cuda/libs/cupy_cusparse.h`.

leofang:
@anaruse Wrappers for more cuSPARSE functionalities are being added in this PR. Would be great if you could take a look 🙂

sunilkpai:
@anaruse I am actually having some unexpected behavior related to cuSPARSE (my cuda version is 10.2 i believe), which I am using for the equivalent scipy implementation of `spilu`. Here is the code to reproduce from my fork:

```
from cupyx.scipy.sparse.linalg import spilu
from cupyx.scipy.sparse.linalg.solve import _destroy_ilu
import cupy as cp
import scipy as sp
import numpy as np

N = 40
data = np.ones((3, N))
data[0, :] = 2
data[1, :] = -1
data[2, :] = -1
A = sp.sparse.spdiags(data, [0, -1, 1], N, N, format='csr')
A.sort_indices()
M, info = spilu(cp.sparse.csr_matrix(A))

b = np.random.randn(N)
x = M(cp.asarray(b))

_destroy_ilu(info) # hangs
a = x.get() # hangs when previous line is commented out
```

If you look at my implementation, you'll see that I have used the `csrsv2` and `csrilu02` routines to implement spilu, which is similar to what is typically recommended for ILU preconditioning I believe. But I am not sure I am handling memory correctly, resulting in this bug.

sunilkpai:
@leofang Since `bicgstab` does not require preconditioner definition, I wanted to see if we might at least get the general implementation into `cupy` first and handle these issues in a separate PR? If so, I can delete my current `spilu` implementation for now, as I think it will take some time for us to get it right, and keep all of the wrappers as they are! The other tests pass. Let me know if this is fine with you!

cupy/cusparse.py
```diff
@@ -89,8 +95,12 @@ def _dtype_to_DataType(dtype):
...
     'csrgemm': (8000, 11000),
     'csrgemm2': (8000, None),
-    'spmv': (10200, None),
-    'spmm': (10301, None),  # accuracy bugs in cuSparse 10.3.0
+    'csrilu02': (8000, None),
+    'csrilu02_analysis': (8000, None),
+    'csrsv2_solve': (8000, None),
+    'csrsv2_analysis': (8000, None),
+    'spmv': (10200, None),  # accuracy bugs in cuSparse 10.3.0
+    'spmm': (10301, None),
```
anaruse, line 103: Please do not change the location of the comment.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,237 @@ def lsqr(A, b):
...
+    Returns:
+        callable:
+            A function x = M(b) that approximates a linear solution for Ax = b
+        tuple:
+            the info objects that need to be
+
+    .. seealso:: :func:`scipy.sparse.linalg.spilu`
+    """
+    # support float32, float64, complex64, and complex128
+    handle = device.get_cusolver_sp_handle()
```
anaruse, line 206: You need to use `device.get_cusparse_handle()` as cuSparse is used in this routine..

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,237 @@ def lsqr(A, b):
...
+
+    # define trans
+    trans_L = cusparse.CUSPARSE_OPERATION_NON_TRANSPOSE
+    trans_U = cusparse.CUSPARSE_OPERATION_NON_TRANSPOSE
+
+    # create descriptions
+    descr_M = cupy.cusparse.MatDescriptor.create()
+    descr_L = cupy.cusparse.MatDescriptor.create()
+    descr_U = cupy.cusparse.MatDescriptor.create()
+    descr_M.set_mat_index_base(cusparse.CUSPARSE_INDEX_BASE_ONE)
```
anaruse, line 242: This must be 'cusparse.CUSPARSE_INDEX_BASE_ZERO'.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,237 @@ def lsqr(A, b):
...
+    descr_M.set_mat_type(cusparse.CUSPARSE_MATRIX_TYPE_GENERAL)
+    descr_L.set_mat_index_base(cusparse.CUSPARSE_INDEX_BASE_ONE)
```
anaruse, line 244: This must be 'cusparse.CUSPARSE_INDEX_BASE_ZERO' too.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,237 @@ def lsqr(A, b):
...
+    descr_L.set_mat_type(cusparse.CUSPARSE_MATRIX_TYPE_GENERAL)
+    descr_L.set_mat_fill_mode(cusparse.CUSPARSE_FILL_MODE_LOWER)
+    descr_L.set_mat_diag_type(cusparse.CUSPARSE_DIAG_TYPE_UNIT)
+    descr_U.set_mat_index_base(cusparse.CUSPARSE_INDEX_BASE_ONE)
```
anaruse, line 248: This must be 'cusparse.CUSPARSE_INDEX_BASE_ZERO' too.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,237 @@ def lsqr(A, b):
...
+                                     out.data.ptr, policy_U, buff.data.ptr)
+        return out
+
+    return M, (info_M, info_U, info_L)
+
+
+def _destroy_ilu(info_list):
+    if info_list is not None:
+        for info in info_list:
+            cusparse.destroyCsrilu02Info(info)
```
anaruse, line 298: The type of `info` might be `Csrsv2Info`. When it is `Csrsv2Info`, `cusparse.destroyCsrsv2Info()` needs to be used as destructor.

anaruse:
> @leofang Since `bicgstab` does not require preconditioner definition, I wanted to see if we might at least get the general implementation into `cupy` first and handle these issues in a separate PR? If so, I can delete my current `spilu` implementation for now, as I think it will take some time for us to get it right, and keep all of the wrappers as they are! The other tests pass. Let me know if this is fine with you!

The problem with `spliu` will probably be solved by the above fix, so please check.

sunilkpai:
Thanks @anaruse, that fixed my issue! I will make some tests now to compare with `scipy`.

sunilkpai:
@leofang @anaruse  I'm running into some testing trouble... any ideas on what went wrong (see above issue)?

sunilkpai:
@leofang I think I'm ready for another PR review once I resolve a few hanging threads: having to destroy info (see `_destroy_ilu`) on the end-user side seems like a nuisance, scipy's and cuda's implementation of `spilu` are different, but the small cases match since they presumably do a full `splu`, maybe I should add tests for `bicgstab` algorithm breakdown. Would like your comment on these!

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,266 @@ def lsqr(A, b):
...
+    Find the solution to a large, sparse, linear system of equations.
+    The function approximates ``Ax = b`` using BiCGSTAB iterative method.
+
+    Args:
+        A (cupy.ndarray or cupyx.scipy.sparse.csr_matrix): The input matrix
+            with dimension ``(N, N)``
+        b (cupy.ndarray): Right-hand side vector.
+        x0 (cupy.ndarray): Initial guess (defaults to zeros)
+        M (tuple(callable, optional(info))): GPU preconditioner function
+            and info object for preconditioner (to destroy after running)
```
leofang, line 87: I don't understand how to make this work. I think we'll need
1. A docstring `.. notes::` providing an example for nontrivial `M`
2. Test coverage (currently lacking)

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,266 @@ def lsqr(A, b):
...
+        if omega == 0:
+            info = -11
+            break
+
+        x += alpha * phat + omega * shat
+        r = s - omega * t
+
+        residual = cupy.linalg.norm(r)
+        if callback is not None:
+            callback(x, i, r)
```
leofang, line 178: See my question in the docstring.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,266 @@ def lsqr(A, b):
...
+        b (cupy.ndarray): Right-hand side vector.
+        x0 (cupy.ndarray): Initial guess (defaults to zeros)
+        M (tuple(callable, optional(info))): GPU preconditioner function
+            and info object for preconditioner (to destroy after running)
+        tol (float): Tolerance for convergence, exits when
+                     residual norm(r) < tol * norm(b)
+        maxiter (int): maximum number of iterations (defaults to `numpy.inf`)
+        callback (callable): a callback function to call at each iteration
+                            (unlike scipy, it accepts iteration and
+                            residual as well as current vector `x`)
```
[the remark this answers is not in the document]
sunilkpai: The rationale is that getting residual information may be useful for debugging and it might take longer since most of the iteration when unpreconditioned is probably the sparse matvecs which we also need to get the residual... But I'm fine sticking with `scipy`'s convention if you like!

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,266 @@ def lsqr(A, b):
...
+
+    # Calculate initial residual
+    x0 = x0 if x0 is not None else cupy.zeros(n, dtype=dtype)
+    r = cupy.cusparse.spmv(A, x0, b, -1, 1)
+    p, q = cupy.zeros_like(r), cupy.zeros_like(r)
+    r_tilde = r.copy()
+
+    # Main BiCGSTAB loop
+
+    i = 0
```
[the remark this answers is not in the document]
sunilkpai: Yep

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,266 @@ def lsqr(A, b):
...
+
+        residual = cupy.linalg.norm(r)
+        if callback is not None:
+            callback(x, i, r)
+        if residual < tol or residual < atol:
+            info = 0
+            break
+        info += 1
+
+    _destroy_ilu(M_info)
```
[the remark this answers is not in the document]
sunilkpai: I was thinking this function will be useful for destroying `spilu` info in a more general context (in that case, I would remove the underscore), but if you think it isn't necessary (since `spilu` is meant to be primarily a preconditioner), I can inline it.

cupyx/scipy/sparse/linalg/solve.py
```diff
@@ -65,3 +66,266 @@ def lsqr(A, b):
...
+        if rho == 0:
+            info = -10
+            break
+        beta = rho / prev_rho * alpha / omega
+        p = r + beta * (p - omega * q)
+
+        # Alpha step
+        phat = M(p)  # apply preconditioner
+
+        q = cupy.cusparse.spmv(A, phat)
```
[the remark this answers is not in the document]
sunilkpai: I have not tried that! Wasn't aware of that shortcut actually!

cupyx/scipy/sparse/linalg/solve.py
sunilkpai: Yes, I agree I should add some tests for this. The `spilu` preconditioner is actually one example. I can write a docstring with a spilu example and make a test for that. Would that work for you?

leofang:
@sunilkpai (Sorry I am outside but recalled something I wanted to say.) Could you post a few benchmarks for the two linalg functions against SciPy? Preferably with a larger problem size. For the implementation it's unclear if the current approach would win. Thanks!

sunilkpai:
@leofang That sounds like a good idea! I did find significant improvement in my `cupy` unpreconditioned bicgstab implementation versus `scipy` for my problems (a 3d FDFD simulation) and in fact I use this `cupy` implementation exclusively now. I found, though that cuda's `spilu` didn't work for my problem (ran into some numerical issues) so I could not try that. I am not ready to share that specific simulation yet, but I could create a gist testing a 2d FDFD problem from ceviche (see [ceviche](https://github.com/fancompute/ceviche)).

cupyx/scipy/sparse/linalg/solve.py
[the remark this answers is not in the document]
leofang: Actually your line here *is* the shortcut, so I was wondering if it's because you detected suboptimal performance 😄

cupyx/scipy/sparse/linalg/solve.py
leofang: Sounds good. This would then becomes an integration test, which is also good. Are they other possible preconditioners either existing in the codebase or can be made up quickly?

sunilkpai: The Jacobi one is simple as well but it requires implementing `A.diagonal()` which I think is not yet implemented.

leofang:
I think for our purpose here it's fine to benchmark against SciPy by simply making up A and b and solve for Ax+b using `bicgstab` without using a full simulation (but personally I am interested in such numbers too).

>  I found, though that cuda's `spilu` didn't work for my problem (ran into some numerical issues) so I could not try that.

This is worrying. Do you have a minimal reproducer that can demonstrate the problem? Maybe @anaruse can look into that and see if it's a real issue on cuSPARSE, or we simply messed up somewhere?

sunilkpai:
> I think for our purpose here it's fine to benchmark against SciPy by simply making up A and b and solve for Ax+b using `bicgstab` without using a full simulation (but personally I am interested in such numbers too).

Yeah, this is just a very convenient way to use `bicgstab` for a "reasonable" problem that I know has a clean solution, which is why I thought of this. It's worthwhile noting there are cases where `bicgstab` runs into problems (even observed in `scipy`) for random matrices. This is actually part of my issue with `jax`, for which I am implementing a matrix-optional version of iterative solves: (see [jax](https://github.com/google/jax/pull/3796)).

> This is worrying. Do you have a minimal reproducer that can demonstrate the problem? Maybe @anaruse can look into that and see if it's a real issue on cuSPARSE, or we simply messed up somewhere?

Yeah I can try and make a more minimal example for sure. Actually, `scipy` had some weird memory error for that particular example and so I couldn't even check it properly.

kmaehashi:
@toslunar Could you take a brief look on this PR?

sunilkpai:
Hey all, apologies for the delay on this PR. It will be another week before I can get back to this!

ma-sadeghi:
Do you have a rough estimate on when this PR will be ready? Thanks

sunilkpai:
@ma-sadeghi i unfortunately have been quite delayed in this as I am a PhD student and got wrapped up in another more urgent project. I need to figure out still what benchmarks to include for bicgstab. I can promise that I can get to this again in a week or so. Thanks for being patient.

malbergo:
Is there any update on this? I can't tell where things left off from the discussion.

sunilkpai:
I apologize. I decided to add this feature to jax and lost the momentum on this one but I still think it is worth doing. I probably won't have too much time to finish this so feel free to give it a crack!

takagi:
Okay, I'll take this over. Thanks for your effort, @sunilkpai!

takagi:
Let me close this PR to have a fresh issue as #7570.

PR #3577, opened 2020-07-08
Fix `slogdet` tests to check dtypes of return values

asi1024:
Fixes to check that the dtypes of return values of `cupy.linalg.slogdet`are equal to those of `numpy.linalg.slogdet`.


emcastillo:
I wonder if its better to change the decorator to compare a tuple of arrays?

asi1024:
We should support the new feature `testing.numpy_cupy_array_list_allclose`?

emcastillo:
I think we can modify `@testing.numpy_cupy_allclose`
to compare a single array or a tuple of arrays instead of creating a new one 🙂

emcastillo:
Jenkins, test this please

PR #3579, opened 2020-07-09
Fix negative value test in test_helper

takagi:
Taking over #2884 with making appropriate changes.

`raises` was wrapping a function decorated with `for_unsigned_dtypes` and
`for_signed_dtypes`. In this scheme, if an error was raised in any of
the combinations of dtypes, the test would pass, which would be unintended.


emcastillo:
Jenkins, test this please

PR #3582, opened 2020-07-09
Deprecate `numpy_cupy_array_list_equal`

emcastillo:
`numpy_cupy_array_equal` can compare tuples or list of arrays making `numpy_cupy_array_list_equal` redundant.

`numpy_cupy_array_equal(type_check=False)` achieves the same functionality

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -492,8 +492,7 @@ def numpy_cupy_array_equal(err_msg='', verbose=True, name='xp',
 
 
 def numpy_cupy_array_list_equal(
-        err_msg='', verbose=True, name='xp',
-        sp_name=None, scipy_name=None, type_check=True):
+        err_msg='', verbose=True, name='xp', sp_name=None, scipy_name=None):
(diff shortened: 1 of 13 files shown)


cupy/testing/helper.py
```diff
@@ -492,7 +492,8 @@ def check_func(x, y):
 
 
 def numpy_cupy_array_list_equal(
-        err_msg='', verbose=True, name='xp', sp_name=None, scipy_name=None):
+        err_msg='', verbose=True, name='xp',
+        sp_name=None, scipy_name=None, type_check=True):
```
toslunar, line 496: IMO, the deprecation of the function does not need the change of its interface while it may be reasonable to add `type_check` argument for consistency.

We may write `type_check = False` inside the function to clarify the difference between the function and `numpy_cupy_array_equal`.

emcastillo: Yeah, you are absolutely right.

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

PR #2711, opened 2019-11-25
Added min, max, argmin, argmax to sparse csr and csc matrices

dloney:
This adds min, max, argmin, and argmax functionality to cupy sparse csr and csc matrix formats. Format differences from scipy are to improve computational performance by eliminating duplicate summation and coo matrix construction time.

PR #3583, opened 2020-07-09
Mark `nonzero` parameters experimental in sparse min/max

### LINKED ISSUE #3535: Difference from SciPy about sparse min/max `sum_duplicates` parameter
I'm not sure its background, but currently, sparse min/max have `sum_duplicates` parameter and its default value is `False`. However, their SciPy counterparts don't have the parameter and always sum duplicate entries as `sum_duplicates=True`.

It would be better to change its default value to `False`.

takagi:
~~Close #3535.~~

~~Currently, sparse min/max/argmin/argmax have `sum_duplicates` and `nonzero` parameters. However, the SciPy counterparts do not have them and always conduct as `sum_duplicates=True` and `nonzero=False`. This PR removes the parameters from the methods to make compliance with SciPy.~~

Merge #3676 first.

Currently, sparse min/max have `nonzero` parameter that indicates that the operations should take care of explicitly stored elements only in a sparse matrix. However, the parameter is now proposed in https://github.com/scipy/scipy/pull/11899 and is not merged to the SciPy master nor released in its stable release. This PR makes it a keyword-only parameter and marks as experimental for future possible changes.

### DIFF
--- a/cupyx/scipy/sparse/data.py
+++ b/cupyx/scipy/sparse/data.py
@@ -137,18 +137,17 @@ class _minmax_mixin(object):
 
     """
 
-    def _min_or_max_axis(self, axis, min_or_max, sum_duplicates, nonzero):
+    def _min_or_max_axis(self, axis, min_or_max, explicit):
         N = self.shape[axis]
         if N == 0:
             raise ValueError("zero-size array to reduction operation")
         M = self.shape[1 - axis]
 
         mat = self.tocsc() if axis == 0 else self.tocsr()
-        if sum_duplicates:
-            mat.sum_duplicates()
+        mat.sum_duplicates()
 
         # Do the reudction
-        value = mat._minor_reduce(min_or_max, axis, nonzero)
+        value = mat._minor_reduce(min_or_max, axis, explicit)
         major_index = cupy.arange(M)
 
         mask = value != 0
@@ -164,7 +163,7 @@ class _minmax_mixin(object):
                 (value, (major_index, cupy.zeros(len(value)))),
                 dtype=self.dtype, shape=(M, 1))
 
-    def _min_or_max(self, axis, out, min_or_max, sum_duplicates, non_zero):
+    def _min_or_max(self, axis, out, min_or_max, explicit):
         if out is not None:
             raise ValueError(("Sparse matrices do not support "
                               "an 'out' parameter."))
@@ -193,20 +192,17 @@ class _minmax_mixin(object):
             return m
 
         if axis == 0 or axis == 1:
-            return self._min_or_max_axis(axis, min_or_max, sum_duplicates,
-                                         non_zero)
+            return self._min_or_max_axis(axis, min_or_max, explicit)
         else:
             raise ValueError("axis out of range")
 
-    def _arg_min_or_max_axis(self, axis, op, sum_duplicates):
+    def _arg_min_or_max_axis(self, axis, op):
         if self.shape[axis] == 0:
             raise ValueError("Can't apply the operation along a zero-sized "
                              "dimension.")
 
         mat = self.tocsc() if axis == 0 else self.tocsr()
-
-        if sum_duplicates:
-            mat.sum_duplicates()
+        mat.sum_duplicates()
 
         # Do the reudction
         value = mat._arg_minor_reduce(op, axis)
@@ -216,7 +212,7 @@ class _minmax_mixin(object):
         else:
             return value[:, None]
 
-    def _arg_min_or_max(self, axis, out, op, compare, sum_duplicates):
+    def _arg_min_or_max(self, axis, out, op, compare):
         if out is not None:
             raise ValueError("Sparse matrices do not support "
                              "an 'out' parameter.")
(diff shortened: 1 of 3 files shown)


dloney:
I would say toggling the default behavior from False to True to be consistent with scipy would be good. The reason I introduced the flag is as a performance enhancement to skip the sum_duplicates operation if the user knows their matrix is canonical. Changing the default flag will make it conform with the default scipy behavior.

For #3583, there's an open pull request on scipy that will implement the explicit nonzero features. I wouldn't remove the functionality from cupy for the sake of standardizing to scipy since it's a useful feature when explicit zeros are included in the sparse matrix. The default behavior for the non_zero flag is the same as standard scipy. The scipy pull request uses the keyword "explicit" rather than "nonzero" which might be a good change to make.

leofang:
I think it's best to stay complaint with the current SciPy, not the SciPy to be, as it could take months to get the PR merged in the upstream. This would also ensure compliance with CuPy's own dev policy.

leofang:
ref: https://github.com/scipy/scipy/pull/11899 (repost here for easier reference)

dloney:
I like the suggestion from @grlee77 in #2711 to keep the functionality but shift it to use keywords. That would keep the same behavior as scipy as well as the explicit zero functionality. 

It's important to keep the functionality available since it provides an order of magnitude speedup when there are explicit zeros in the the sparse matrix.

takagi:
@dloney, I agree with you that the explicit zero functionality is useful. Now I think of when and how to introduce it to CuPy's **stable** version.

We have our [API Compatibility Policy](https://docs-cupy.chainer.org/en/stable/compatibility.html), where we can make changes that break the backward compatibility only in major updates. Incompatible changes here include those as small as modifying parameter names. The next major update (8.0.0) is planned on July 30, which is soon and the upstream won't be fixed before the date.

I suppose we have the following options:
- Introducing to 8.0.0 with making it a keyword-only parameter (@grlee77 's suggestion)
- Introducing to 8.0.0 with making it keyword-only and marking as [cupy.util.experimental](https://github.com/cupy/cupy/blob/92ab7b427691dd0dd7172b562d814d640a54eb7c/cupy/util.pyx#LC119) until SciPy supports it
- Removing from 8.0.0 and introducing to (maybe) 9.0.0 after SciPy supports it (@leofang 's suggestion)

I prefer the third option as @leofang says for making things cleaner, but I think the second is acceptable. How do you think?

leofang:
Marking it experimental would not be bad as long as it helps!

dloney:
I'd be happy marking it as experimental. The core functionality probably shouldn't have the experimental flag applied to it though since it replicates what's already available in scipy. 

I'm not familiar with the cupy implementation of the experimental flag . Would we be able to invoke the experimental flag only when the keyword argument is enabled?  Looking it over quick, it looks like it would be adding an if statement with additional import were the keyword enabled.

grlee77:
> I prefer the third option as @leofang says for making things cleaner, but I think the second is acceptable. How do you think?

I am not using it personally, but given that it has already been implemented and it seems at least some are finding it useful, I would lean toward leaving it in, but marking it as experimental

takagi:
> Would we be able to invoke the experimental flag only when the keyword argument is enabled?

Yes, we can call the flag only when the keyword argument is used. The flag is implemented as a function instead of a decorator for such a case.

Okay, then I'll change this PR as such. Thanks, guys!

takagi:
I fixed this PR to make the parameters keyword-only and to mark them experimental.

Now please let me make sure about the `sum_duplicates` parameter, @dloney. `sum_duplicates` method returns immediately when a sparse matrix is already canonical. Do we still need this?

https://github.com/cupy/cupy/blob/725c350884142c85c78df13a111f29b7c508540e/cupyx/scipy/sparse/compressed.py#L708-L709

dloney:
No, I don't believe so especially given the proposed scipy format.

dloney:
No, I don't believe so especially given the proposed scipy format. Do we also want to change the parameter name from nonzero to explcit to match scipy?

takagi:
I've extracted removing `sum_duplicate` parameter to #3676 to make the changes in PRs smaller.

cupyx/scipy/sparse/data.py
```diff
@@ -298,11 +297,10 @@ def min(self, axis=None, out=None, sum_duplicates=False, nonzero=False):
...
                 the default value, as this argument is not used.
-            sum_duplicates (bool): Flag to indicate that duplicate elements
-                should be combined prior to the operation
-            nonzero (bool): Return the minimum nonzero value and ignore all
-                zero entries. If the dimension has no nonzero values, a zero is
-                then returned to indicate that it is the only available value.
+            explicit (bool): Return the minimum value explicitly specified and
+                ignore all implicit zero entries. If the dimension has no
+                explicit values, a zero is then returned to indicate that it is
+                the only implicit value.
```
emcastillo, line 303: Let's add a note about explicit being experimental

takagi: Added!

emcastillo:
Jenkins, test this please

PR #3584, opened 2020-07-09
Move stream to `cupy_backends`

okuta:
~This PR relates #3385 and #3386.~
This PR relates #3597 .


leofang:
Sorry, I think I didn't get the point 😅 Why is `cupy/cuda/stream.pyx` removed instead of `cupy_backends/cuda/stream.pyx`? I thought you wanna keep CuPy's wrappers (in this case, `Stream` and `Event`) in the `cupy.cuda` namespace?

emcastillo:
I think the same as @leofang 😅

kmaehashi:
We discussed this today and the team is thinking that it's better to keep the current implementation. https://github.com/cupy/cupy/pull/4445#issuecomment-747245162

PR #3587, opened 2020-07-10
Reuse FFT plan for convolve/correlate

leofang:
Follow-up of https://github.com/cupy/cupy/pull/3560#issuecomment-656301909. Tests passed locally.

### DIFF
--- a/cupy/math/misc.py
+++ b/cupy/math/misc.py
@@ -1,3 +1,5 @@
+import contextlib
+
 import cupy
 import cupyx
 import cupyx.scipy.fft
@@ -5,6 +7,7 @@ import cupyx.scipy.fft
 from cupy import core
 from cupy.core import _routines_math as _math
 from cupy.core import fusion
+from cupy.cuda import cufft
 from cupy.lib import stride_tricks
 
 
@@ -63,15 +66,28 @@ def _fft_convolve(a1, a2, mode):
 
     if a1.dtype.kind == 'c' or a2.dtype.kind == 'c':
         fft, ifft = cupy.fft.fft, cupy.fft.ifft
+        is_c2c = True
     else:
         fft, ifft = cupy.fft.rfft, cupy.fft.irfft
+        is_c2c = False
 
     dtype = cupy.result_type(a1, a2)
     n1, n2 = a1.size, a2.size
     out_size = cupyx.scipy.fft.next_fast_len(n1 + n2 - 1)
-    fa1 = fft(a1, out_size)
-    fa2 = fft(a2, out_size)
-    out = ifft(fa1 * fa2, out_size)
+    if is_c2c:
+        # skip calling get_fft_plan() as we know the args exactly
+        fft_plan = cufft.Plan1d(
+            out_size,
+            cufft.CUFFT_C2C if dtype == cupy.complex64 else cufft.CUFFT_Z2Z,
+            1)
+    else:
+        # this is a no-op context manager
+        # TODO(leofang): use contextlib.nullcontext() for PY37+?
+        fft_plan = contextlib.suppress()
+    with fft_plan:
+        fa1 = fft(a1, out_size)
+        fa2 = fft(a2, out_size)
+        out = ifft(fa1 * fa2, out_size)
 
     if mode == 'full':
         start, end = 0, n1 + n2 - 1
--- a/cupy/math/misc.py
+++ b/cupy/math/misc.py
@@ -74,19 +74,25 @@ def _fft_convolve(a1, a2, mode):
     dtype = cupy.result_type(a1, a2)
     n1, n2 = a1.size, a2.size
     out_size = cupyx.scipy.fft.next_fast_len(n1 + n2 - 1)
+    # skip calling get_fft_plan() as we know the args exactly
     if is_c2c:
-        # skip calling get_fft_plan() as we know the args exactly
         fft_plan = cufft.Plan1d(
             out_size,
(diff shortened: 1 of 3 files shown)


leofang:
@emcastillo Could you post the script you used to generate https://github.com/cupy/cupy/pull/3560#issuecomment-655941892, or could you run it for me?

leofang:
Both complex and real inputs are supported! For the latter, it makes sense to only reuse the forward plan; the backward one is run only once, so we may just let `cupy.fft.irfft` handle it.

emcastillo:
Let me crunch the numbers

emcastillo:
becomes 200us faster in all cases. drops to total time of 1200 us.

leofang:
If we will have a plan cache soon, the only advantage of this PR would be to avoid accessing the cache.

emcastillo:
Jenkins, test this please

leofang:
Hmm...The failures actually make sense. `result_type(float16, complex128)` returns `complex128`, but internally a `float16` array is converted to a `complex64` one before FFT runs. Not sure how my local tests and the other CIs got around this issue...

emcastillo:
Yes, I was a bit puzzled

emcastillo:
I think the problem may be when the two input arrays are not of the same type?

leofang:
What's the NumPy version in `cupy-py35`?

emcastillo:
1.15.4

leofang:
Holy *!@#, `cupy-py35` is the only one not on CUDA 10.0, and so the only one actually ran these tests...

leofang:
I can locally reproduce the error by running this snippet:
```python
import cupy as xp
from cupy import testing

dtype1 = xp.complex128
dtype2 = xp.float16

a = testing.shaped_random((200,), xp, dtype1)
b = testing.shaped_random((100,), xp, dtype2)
out = xp.convolve(a, b, mode='full')
print(out, out.dtype)
```

leofang:
@emcastillo Fixing this requires addressing #3603 first I am afraid...

leofang:
I wil find a way to work around #3603.

leofang:
I worked around the dtype issue (#3603) under the assumption that reusing the plan gains more, and the bonus coming with this change is that the precision issue that we previously encountered is gone, so the tests no longer need to be skipped in CUDA 10.0. 

@emcastillo We need to confirm if we still get the 200us reduction in time, otherwise we should just abandon this and start working on a plan cache.

leofang:
Jenkins, test this please

cupy/math/misc.py
```diff
@@ -84,12 +111,12 @@ def _fft_convolve(a1, a2, mode):
         raise ValueError(
             'acceptable mode flags are `valid`, `same`, or `full`.')
 
-    out = out[start: end]
+    out = out[start:end]
 
-    if dtype.kind in 'iu':
+    if out.dtype.kind in 'iu':
```
leofang, line 116: This is a bug fix.

cupy/math/misc.py
```diff
@@ -63,15 +66,39 @@ def _fft_convolve(a1, a2, mode):
...
+        is_c2c = True
     else:
         fft, ifft = cupy.fft.rfft, cupy.fft.irfft
+        is_c2c = False
+
+    out_dtype = cupy.result_type(a1, a2)
+
+    # hack to work around NumPy/CuPy FFT dtype incompatibility
+    if out_dtype == cupy.float16:
+        dtype = cupy.float32
```
emcastillo, line 78: We don't really want to compute a fp16 fft in fp32 if possible 😅

emcastillo:
I think the problem is that you are using a complex plan for a non-complex datatype,
you should only reuse plans if both datatypes are the same

cupy/math/misc.py
leofang: What do you mean? cuFFT does not support fp16, and we internally cast fp16 to fp32 before executing FFT. This is just to do the cast step earlier, and also to generate a correct plan.

leofang:
> you should only reuse plans if both datatypes are the same

Or when we know both dtypes would be cast to the same dtype when calling fft.

cupy/math/misc.py
emcastillo: Ok! I was not understanding it completely so it's my fault :).

I verified what you said. Could you please add a comment detailing this?
Cufft does not support fp16 so we cast it to fp32 here instead of letting cupy.fft do it to create a correct plan.

emcastillo:
Let's just add the comment and this will be good to go :)

leofang:
Jenkins, test this please

cupy/math/misc.py
```diff
@@ -1,10 +1,14 @@
...
+
 import cupy
 import cupyx
 import cupyx.scipy.fft
 
 from cupy import core
 from cupy.core import _routines_math as _math
 from cupy.core import fusion
+from cupy.cuda import cufft
+from cupy.fft.fft import _output_dtype
```
leofang, line 11: Importing in this way is unavoidable because helper functions are not exposed to `cupy/fft/__init__.py`.

cupy/math/misc.py
leofang: Thanks! I added it. I also call the internal fft helper to determine the type. Seems less error prone to me.

leofang:
On my system I saved 400us CPU time for two complex128 inputs. GPU time fluctuates a lot from shot to shot, so I couldn't tell the impact, but I think it's a good sign.

PR #3590, opened 2020-07-10
Add `cupy.polymul`

Dahlia-Chehata:


Dahlia-Chehata:
Waiting for `polyadd` to apply the same changes before review

asi1024:
Please resolve the conflict.

cupy/lib/polynomial.pyx
```diff
@@ -131,11 +152,13 @@ cdef class poly1d:
...
 
-    # TODO(Dahlia-Chehata): implement using polymul
     def __pow__(self, val, modulo):
         if not cupy.isscalar(val) or int(val) != val or val < 0:
             raise ValueError('Power to non-negative integers only.')
-        raise NotImplementedError
+        out = 1
+        for _ in range(val):
+            out = polymul(self, out)
+        return out
```
asi1024, line 161: Please split into another PR.

asi1024:
Please resolve conflicts.

Dahlia-Chehata:
What is the best way to check for devices synchronization for polyadd & polysub only without code repetition?

cupy/lib/_routines_poly.py
```diff
@@ -13,8 +16,11 @@ def _get_coeffs(x):
 
     def wrapper(*args):
         coeffs = [_get_coeffs(x) for x in args]
-        out = func(*coeffs)
-
+        if func.__name__ == 'polymul':
+            out = func(*coeffs)
+        else:
+            with cupyx.allow_synchronize(False):
+                out = func(*coeffs)
```
Dahlia-Chehata: I decided not to continue with this to keep implementation and testing separated

tests/cupy_tests/lib_tests/test_polynomial.py
```diff
@@ -380,46 +400,55 @@ def test_poly1d_ne2(self, xp, dtype):
...
+        out = func(a, b)
         assert out.variable == 'x'
         return out
+
+
+@testing.gpu
+class TestPolymul(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    def test_polymul_ndim(self, dtype):
```
asi1024, line 449: Why not check it also for `polyadd` and `polysub`?

Dahlia-Chehata: https://github.com/cupy/cupy/pull/3548#discussion_r454143443

cupy/core/core.pyx
```diff
@@ -1005,6 +1005,8 @@ cdef class ndarray:
             return _math._subtract(x, y)
 
     def __mul__(x, y):
+        if isinstance(y, cupy.poly1d):
+            y = y.coeffs
```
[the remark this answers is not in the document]
Dahlia-Chehata: 1) we cared about devices synchronization issue in polyadd & polysub. But since there is no way to avoid it in polymul, so why not support them? Actually why not supporting them in polyadd & polysub too since we already have room to allow devices synchronization for special cases
2) The tests are more neat in this case for polymul without separating the 16 combinations.

So do you still want to remove it and separate the tests?

tests/cupy_tests/lib_tests/test_polynomial.py
asi1024: I see, but does `numpy.polymul` guarantee to raise an error for such inputs?

cupy/core/core.pyx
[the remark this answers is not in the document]
asi1024: Supporting only `cupy.ndarray * cupy.poly1d` is confusing for the users.

tests/cupy_tests/lib_tests/test_polynomial.py
Dahlia-Chehata: Yes because `numpy.polymul` calls `poly1d().coeffs` on its inputs first, wich raises `ValueError` for multidimensional inputs.

asi1024: Is it documented?

Dahlia-Chehata: It is not stated that it raises error for ndim inputs. But it says that the input must be poly1d object or 1D sequence of polynomial coefficients

asi1024: The documentation of `numpy.polyadd` also says that `Each input must be either a poly1d object or a 1D sequence of polynomial coefficients` (https://numpy.org/doc/stable/reference/generated/numpy.polyadd.html), but `numpy.polyadd` does not always raise an error for non-1dim inputs.

Dahlia-Chehata: `polyadd` raises error for most of the cases. Only few cases (technically speaking where broadcasting is possible), it does not. The same applies for polysub. It is not clearly stated in any of the 3 documentations that it raises error for these cases. But polymul's implementation in numpy does guarantee raising error for ndim inputs. So what do you suggest now? Removing the test or keep it?

asi1024: How they are implemented internally does not become the reason for the difference in the behavior of `numpy.polyadd` and `numpy.polymul`.

asi1024: I think it's both OK to keep the test or remove it. However, if you want to keep the test, please fix all routines to raise `ValueError` for all non-1d inputs.

cupy/lib/_routines_poly.py
```diff
@@ -40,6 +42,8 @@ def polyadd(a1, a2):
     .. seealso:: :func:`numpy.polyadd`
 
     """
+    if a1.ndim != 1 or a2.ndim != 1:
+        raise ValueError('Multidimensional inputs are not supported')
```
asi1024, line 46: How about moving it to `_wraps_polyroutine`?

cupy/lib/_routines_poly.py
```diff
@@ -5,6 +5,8 @@
 
 def _wraps_polyroutine(func):
     def _get_coeffs(x):
+        if func.__name__ == 'polymul':
+            return cupy.poly1d(x).coeffs
```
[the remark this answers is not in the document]
asi1024: Could you address this comment?

tests/cupy_tests/lib_tests/test_polynomial.py
```diff
@@ -380,46 +387,52 @@ def test_poly1d_ne2(self, xp, dtype):
...
+        return func(a, b)
 
 
 @testing.gpu
 @testing.parameterize(*testing.product({
-    'fname': ['polyadd', 'polysub'],
+    'fname': ['polyadd', 'polysub', 'polymul'],
 }))
-class TestPolyArithmeticDiffTypes(unittest.TestCase):
+class TestPolyArithmetic(unittest.TestCase):
```
asi1024, line 409: Please restore the name to `TestPoly1dArithmeticDiffTypes` and split the multi-dimension tests.

cupy/lib/_routines_poly.py
```diff
@@ -71,3 +76,26 @@ def polysub(a1, a2):
...
+
+    .. seealso:: :func:`numpy.polymul`
+
+    """
+    a1 = cupy.trim_zeros(a1, trim='f')
+    a2 = cupy.trim_zeros(a2, trim='f')
+    if a1.size == 0:
+        a1 = cupy.array([0.])
+    if a2.size == 0:
+        a2 = cupy.array([0.])
```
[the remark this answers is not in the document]
Dahlia-Chehata: tests fail with this fix

cupy/lib/polynomial.pyx
```diff
@@ -129,11 +129,20 @@ cdef class poly1d:
     def __pos__(self):
         return self
 
-    # TODO(Dahlia-Chehata): use polymul for non-scalars
     def __mul__(self, other):
+        if _should_use_rop(self, other):
+            return other.__rmul__(self)
         if cupy.isscalar(other):
+            # case: poly1d * python scalar
             return poly1d(self.coeffs * other)
```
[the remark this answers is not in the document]
Dahlia-Chehata: It is needed for the commented case

cupy/lib/polynomial.pyx
```diff
@@ -129,11 +129,20 @@ cdef class poly1d:
...
-        raise NotImplementedError
+        if isinstance(self, numpy.generic):
+            # for consistency: numpy scalar * poly1d
+            raise TypeError('Numpy scalar and poly1d multiplication'
+                            ' is not supported currently.')
+        if cupy.isscalar(self) and isinstance(other, poly1d):
+            # case: python scalar * poly1d
+            self = other._coeffs.dtype.type(self)
```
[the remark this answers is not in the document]
Dahlia-Chehata: Ditto

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
asi1024: Why? the input array will be trimmed in polymul routine.

cupy/lib/_routines_poly.py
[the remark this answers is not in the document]
asi1024: Sorry, the following suggestion is correct.
```suggestion
    if a1.size == 0 or a2.size == 0:
        size = max(a1.size, a2.size, 1)
        return `cupy.zeros((size,))`
```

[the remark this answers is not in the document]
asi1024: Ah..., the new suggestion seems to fail in case of complex dtype input... It's OK to leave the current implementation.

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
Dahlia-Chehata: There is a difference in the result dtype in case of  poly1d * python scalar. `convolve ` in polymul uses `result_type` for both of its inputs. To work around this, the result need to be casted to numpy.float64 in case when python scalar is a rhs operand but casting can be bothersome if the original result is of complex type. Another solution is to handle this specifically for polymul function in the wrapper itself which you didn't favour before

cupy/lib/polynomial.pyx
```diff
@@ -129,11 +129,20 @@ cdef class poly1d:
...
-    # TODO(Dahlia-Chehata): use polymul for non-scalars
     def __mul__(self, other):
+        if _should_use_rop(self, other):
+            return other.__rmul__(self)
         if cupy.isscalar(other):
+            # case: poly1d * python scalar
             return poly1d(self.coeffs * other)
-        raise NotImplementedError
+        if isinstance(self, numpy.generic):
+            # for consistency: numpy scalar * poly1d
```
asi1024, line 139: If you want to comment about the "consistency", more detail should be written. This comment does not mention about what is the consistency.

cupy/lib/polynomial.pyx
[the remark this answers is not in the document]
asi1024: If so, you should write a comment about the background on this workaround.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #3612, opened 2020-07-15
[MNT] Make float16 a standalone header

leofang:
This is to support #3520, in which I found CUB BlockLoad needs a fp16 to complex conversion. @asi1024 suggested we should make a separate PR for the refactoring.  

The type_traits header is not needed in the present approach, so I do not include it here. We will revisit it when addressing #3408.


leofang:
Jenkins, test this please

cupy/core/include/cupy/float16.cuh
```diff
@@ -0,0 +1,182 @@
...
+#ifdef CUPY_CUB_BLOCK_REDUCTION
+  // these type conversion functions are needed for CUB load to be able
+  // to do an implicit conversion
+  __device__ operator complex<float>() const {
+      return complex<float>(float(data_), 0);
+  }
+  __device__ operator complex<double>() const {
+      return complex<double>(double(data_), 0);
+  }
+#endif
```
leofang, line 76: This is the only main change. It is required for #3520.

leofang: The reason I wrap them with a macro is because in other kernels, such as `cast_copy` here:
https://github.com/cupy/cupy/blob/92dffe9533294430159673dbf9875f3ca6b384f4/cupy/core/_ufuncs.py#L4-L13
there will be more than one candidates for possible specializations. If needed I can try addressing this minor issue to remove the macro.

cupy/core/include/cupy/float16.cuh
```diff
@@ -0,0 +1,182 @@
...
+
+#ifdef __HIPCC__
+
+#include <hip/hip_fp16.h>
+
+#elif __CUDACC_VER_MAJOR__ >= 9
+
+#include <cuda_fp16.h>
+
+#else  // #if __CUDACC_VER_MAJOR__ >= 9
```
asi1024, line 20: note: This part will be able to remove after https://github.com/cupy/cupy/issues/3301 is resolved.

cupy/core/include/cupy/float16.cuh
asi1024: How about defining it even when `CUPY_CUB_BLOCK_REDUCTION` is not set? This cast seems useful for general use-cases.

cupy/core/include/cupy/float16.cuh
leofang: Yeah would be great. We have this macro check in several places.

cupy/core/include/cupy/float16.cuh
leofang: Yes it should be. But, during my tests I found tests could randomly fail (in the sense it was reproducible only by luck), complaining that `cast_copy` has multiple constructor candidates, see the error log below. 

I enforce a `cast_copy` specialization for fp16 -> complex in f4c527a, but I am not sure whether there could be other issues or not. Since it is not always reproducible, I decide to push it to the CI, and retry a few times.
```
NVRTC compilation error: /home/leofang/cupy/cupy/core/include/cupy/carray.cuh(154): warning: statement is unreachable
          detected during instantiation of "void CIndexer<_ndim>::set(ptrdiff_t) [with _ndim=1]" 
/tmp/tmpikvm9878/03295f9e3745ca3612dfad58a5580446_2.cubin.cu(22): here

/tmp/tmpikvm9878/03295f9e3745ca3612dfad58a5580446_2.cubin.cu(11): error: more than one instance of constructor "thrust::complex<T>::complex [with T=float]" matches the argument list:
            function "thrust::complex<T>::complex(const T &, const T &) [with T=float]"
            function "thrust::complex<T>::complex(const thrust::complex<float> &) [with T=float]"
            argument types are: (const in0_type)
          detected during instantiation of "void cast_copy(const U &, T &) [with T=out0_type, U=in0_type]" 
(25): here

1 error detected in the compilation of "/tmp/tmpikvm9878/03295f9e3745ca3612dfad58a5580446_2.cubin.cu".

-----
Name: /tmp/tmpikvm9878/03295f9e3745ca3612dfad58a5580446_2.cubin.cu
Options: -I/home/leofang/cupy/cupy/core/include -I /home/leofang/cupy/cupy/core/include/cupy/_cuda/cuda-10.0 -I /usr/local/cuda/include -ftz=true -arch=compute_70
CUDA source:
01 #include <cupy/complex.cuh>
02 #include <cupy/float16.cuh>
03 #include <cupy/carray.cuh>
04 #include <cupy/atomics.cuh>
05 
06     typedef float16 in0_type;
07 typedef complex<float> out0_type;
08 
09 
10 template<typename T, typename U>
11 __device__ void cast_copy(const U& x, T& y) {y = T(x);}
12 template<typename T, typename U>
13 __device__ void cast_copy(const complex<U>& x, complex<T>& y) {
14     y = complex<T>(x);
15 }
16 template<typename T, typename U>
17 __device__ void cast_copy(const complex<U>& x, T& y) {y = T(x.real());}
18 
19     extern "C" __global__ void cupy_copy__float16_complex64(const CArray<float16, 1, 1, 1> _raw_in0, CArray<complex<float>, 1, 1, 1> _raw_out0, CIndexer<1> _ind) {
20       ;
21       CUPY_FOR(i, _ind.size()) {
22         _ind.set(i);
23         const in0_type in0(_raw_in0[_ind.get()]);
24 out0_type &out0 = _raw_out0[_ind.get()];
25 cast_copy(in0, out0);
26       }
27       ;
28     }
29 
-----
```

leofang:
As mentioned in https://github.com/cupy/cupy/pull/3612#discussion_r456023183, I'm going to kick off the CI for multiple times. ***DO NOT MERGE!***

Jenkins, test this please

cupy/core/include/cupy/float16.cuh
leofang: And another thing very odd is that by not clearing out the kernel caches, I expect tests to either always pass or always fail, as the compiled cubins should be reused. However, it doesn't seem the case: sometimes the same test could pass in the first few runs, but then fail once, and then pass again. No idea what's going on...😞

leofang:
Let me take a look later. I hope to find a universal solution. If not, let's revert.

leofang:
Jenkins, test this please

leofang:
CuPy's complex headers do not satisfy the rule of three (`operator=()` is missing)...🤦🏻‍♂️

leofang:
Closing this PR as it turned out https://github.com/cupy/cupy/pull/3612#issuecomment-659972674 was the missing key change, see the [updated description](https://github.com/cupy/cupy/pull/3520#issue-442411445) in #3520.

PR #3614, opened 2020-07-15
Ndimage generic filters

### LINKED ISSUE #2099: Adding equivalent filters from scipy.ndimage.filters

### LINKED ISSUE #3111: Implemention of `cupyx.scipy.ndimage.filters`

coderforlife:
This is the final PR to close #2099 and close #3111.

Additionally, the filter code was split into multiple files as it was getting too unwieldy for a single file.

Some notes:

- The `generic_filter()` function cannot exactly match in Cupy and Scipy. Scipy's accepts any Python function with a single `ndarray` parameter and single scalar return value or a wrapped C function with a specific signature. The Cupy version accepts any *fuseable* Python function with the same restrictions (`ndarray` param, scalar return), a `RawKernel` with almost the same C signature (just missing return value) as with Scipy. It also accepts `ReductionKernel` with 1 input and 1 output.
  - Support for new fuseable functions is hackish, the `RawKernel` and `ReductionKernel` are significantly more 'legit'
- The `generic_filter1d()` function matches with Scipy even less. It only accepts a `RawKernel` with a similar C signature (just missing return value). It does not accept any Python functions. This is primarily due to the very odd way that Scipy defined the function.
  - Additionally due to their odd way Scipy defined the function it isn't as parallelizable and likely has significant performance issues compared to all of the other filter functions

Still in draft state since I would like:

- Feedback on the design. Some of this is a bit "radical" maybe and perhaps some people more familiar with Cupy and CUDA can help me straighten it out.
- There are no formal tests yet.

Some informal tests which help you gauge speed-up and my results are available at https://gist.github.com/coderforlife/e3a5fffa17be71c2f97779e0c41fb5a1

Speed Summary
--------------
1000x1000 input with 3x3, 15x15, and 25x25 filter sizes.

Using raw and reduction kernels are the fastest.

Fused functions are 1.25x to 7x slower (that is 25% to 600% slower) than directly writing the kernels - they get much closer with larger filter sizes. 

Numba functions used with Scipy are 6x to 26x slower than fused functions, generally losing more ground as the filter size goes up.

Pure Python functions used with Scipy are then 10x to 200x slower than the numba functions with scipy.

Ultimately, if you can hand-write your CUDA functions expect 200x-10000x speedup over Scipy with Python function. If your function is fuseable you only get a small amount of loss and you get to write it as Python!

I have done personal tests with `generic_filter1d`, and for larger sizes it does improve upon Scipy, but not nearly as much. I am mainly thinking of that implementation simply to make the library complete (it is still okay, just not stellar).

### DIFF
--- a/cupyx/scipy/ndimage/_filters_core.py
+++ b/cupyx/scipy/ndimage/_filters_core.py
@@ -1,0 +1,366 @@
+import cupy
+
+
+def _get_output(output, input, shape=None):
+    if shape is None:
+        shape = input.shape
+    if isinstance(output, cupy.ndarray):
+        if output.shape != tuple(shape):
+            raise ValueError('output shape is not correct')
+    else:
+        dtype = output
+        if dtype is None:
+            dtype = input.dtype
+        output = cupy.zeros(shape, dtype)
+    return output
+
+
+def _fix_sequence_arg(arg, ndim, name, conv=lambda x: x):
+    if isinstance(arg, str):
+        return [conv(arg)] * ndim
+    try:
+        arg = iter(arg)
+    except TypeError:
+        return [conv(arg)] * ndim
+    lst = [conv(x) for x in arg]
+    if len(lst) != ndim:
+        msg = "{} must have length equal to input rank".format(name)
+        raise RuntimeError(msg)
+    return lst
+
+
+def _check_origin(origin, width):
+    origin = int(origin)
+    if (width // 2 + origin < 0) or (width // 2 + origin >= width):
+        raise ValueError('invalid origin')
+    return origin
+
+
+def _check_mode(mode):
+    if mode not in ('reflect', 'constant', 'nearest', 'mirror', 'wrap'):
+        msg = 'boundary mode not supported (actual: {})'.format(mode)
+        raise RuntimeError(msg)
+    return mode
+
+
+def _get_inttype(input):
+    # The integer type to use for indices in the input array
+    # The indices actually use byte positions and we can't just use
+    # input.nbytes since that won't tell us the number of bytes between the
+    # first and last elements when the array is non-contiguous
+    nbytes = sum((x-1)*abs(stride) for x, stride in
+                 zip(input.shape, input.strides)) + input.dtype.itemsize
+    return 'int' if nbytes < (1 << 31) else 'ptrdiff_t'
+
+
+def _check_size_footprint_structure(ndim, size, footprint, structure,
+                                    stacklevel=3, force_footprint=False):
+    if structure is None and footprint is None:
+        if size is None:
+            raise RuntimeError("no footprint or filter size provided")
+        sizes = _fix_sequence_arg(size, ndim, 'size', int)
+        if force_footprint:
+            return None, cupy.ones(sizes, bool), None
+        return sizes, None, None
+    if size is not None:
+        import warnings
+        warnings.warn("ignoring size because {} is set".format(
+            'structure' if footprint is None else 'footprint'),
+            UserWarning, stacklevel=stacklevel+1)
+
+    if footprint is not None:
+        footprint = cupy.array(footprint, bool, True, 'C')
+        if not footprint.any():
+            raise ValueError("all-zero footprint is not supported")
+
+    if structure is None:
+        if not force_footprint and footprint.all():
+            if footprint.ndim != ndim:
+                raise RuntimeError("size must have length equal to input rank")
+            return footprint.shape, None, None
+        return None, footprint, None
+
+    structure = cupy.ascontiguousarray(structure)
+    if footprint is None:
+        footprint = cupy.ones(structure.shape, bool)
+    return None, footprint, structure
+
+
+def _convert_1d_args(ndim, weights, origin, axis):
+    if weights.ndim != 1 or weights.size < 1:
+        raise RuntimeError('incorrect filter size')
+    axis = cupy.util._normalize_axis_index(axis, ndim)
+    wshape = [1]*ndim
+    wshape[axis] = weights.size
+    weights = weights.reshape(wshape)
+    origins = [0]*ndim
+    origins[axis] = _check_origin(origin, weights.size)
+    return weights, tuple(origins)
+
+
+def _check_nd_args(input, weights, mode, origin, wghts_name='filter weights'):
+    if input.dtype.kind == 'c':
+        raise TypeError('Complex type not supported')
+    _check_mode(mode)
+    # Weights must always be 2 GiB or less
+    if weights.nbytes > (1 << 31):
+        raise RuntimeError('weights must be 2 GiB or less, use FFTs instead')
+    weight_dims = [x for x in weights.shape if x != 0]
+    if len(weight_dims) != input.ndim:
+        raise RuntimeError('{} array has incorrect shape'.format(wghts_name))
+    origins = _fix_sequence_arg(origin, len(weight_dims), 'origin', int)
+    for origin, width in zip(origins, weight_dims):
+        _check_origin(origin, width)
+    return tuple(origins), _get_inttype(input)
+
+
+def _run_1d_filters(filters, input, args, output, mode, cval, origin=0):
+    """
+    Runs a series of 1D filters forming an nd filter. The filters must be a
+    list of callables that take input, arg, axis, output, mode, cval, origin.
+    The args is a list of values that are passed for the arg value to the
+    filter. Individual filters can be None causing that axis to be skipped.
+    """
+    output_orig = output
+    output = _get_output(output, input)
+    modes = _fix_sequence_arg(mode, input.ndim, 'mode', _check_mode)
+    origins = _fix_sequence_arg(origin, input.ndim, 'origin', int)
+    n_filters = sum(filter is not None for filter in filters)
+    if n_filters == 0:
+        output[...] = input[...]
+        return output
+    # We can't operate in-place efficiently, so use a 2-buffer system
+    temp = _get_output(output.dtype, input) if n_filters > 1 else None
+    first = True
+    iterator = zip(filters, args, modes, origins)
+    for axis, (fltr, arg, mode, origin) in enumerate(iterator):
+        if fltr is None:
+            continue
+        fltr(input, arg, axis, output, mode, cval, origin)
+        input, output = output, temp if first else input
+    if isinstance(output_orig, cupy.ndarray) and input is not output_orig:
+        output_orig[...] = input
+        input = output_orig
+    return input
+
+
+def _call_kernel(kernel, input, weights, output, structure=None,
+                 weights_dtype=cupy.float64, structure_dtype=cupy.float64):
+    """
+    Calls a constructed ElementwiseKernel. The kernel must take an input image,
+    an optional array of weights, an optional array for the structure, and an
+    output array.
+
+    weights and structure can be given as None (structure defaults to None) in
+    which case they are not passed to the kernel at all. If the output is given
+    as None then it will be allocated in this function.
+
+    This function deals with making sure that the weights and structure are
+    contiguous and float64 (or bool for weights that are footprints)*, that the
+    output is allocated and appriopately shaped. This also deals with the
+    situation that the input and output arrays overlap in memory.
+
+    * weights is always cast to float64 or bool in order to get an output
+    compatible with SciPy, though float32 might be sufficient when input dtype
+    is low precision. If weights_dtype is passed as weights.dtype then no
+    dtype conversion will occur. The input and output are never converted.
+    """
+    args = [input]
+    if weights is not None:
+        weights = cupy.ascontiguousarray(weights, weights_dtype)
+        args.append(weights)
+    if structure is not None:
+        structure = cupy.ascontiguousarray(structure, structure_dtype)
+        args.append(structure)
+    output = _get_output(output, input)
+    needs_temp = cupy.shares_memory(output, input, 'MAY_SHARE_BOUNDS')
+    if needs_temp:
+        output, temp = _get_output(output.dtype, input), output
+    args.append(output)
+    kernel(*args)
+    if needs_temp:
+        temp[...] = output[...]
+        output = temp
+    return output
+
+
+def _generate_boundary_condition_ops(mode, ix, xsize):
+    if mode == 'reflect':
+        ops = '''
+        if ({ix} < 0) {{
+            {ix} = - 1 -{ix};
+        }}
+        {ix} %= {xsize} * 2;
+        {ix} = min({ix}, 2 * {xsize} - 1 - {ix});'''.format(ix=ix, xsize=xsize)
+    elif mode == 'mirror':
+        ops = '''
+        if ({xsize} == 1) {{
+            {ix} = 0;
+        }} else {{
+            if ({ix} < 0) {{
+                {ix} = -{ix};
+            }}
+            {ix} = 1 + ({ix} - 1) % (({xsize} - 1) * 2);
+            {ix} = min({ix}, 2 * {xsize} - 2 - {ix});
+        }}'''.format(ix=ix, xsize=xsize)
+    elif mode == 'nearest':
+        ops = '''
+        {ix} = min(max({ix}, 0), {xsize} - 1);'''.format(ix=ix, xsize=xsize)
+    elif mode == 'wrap':
+        ops = '''
+        {ix} %= {xsize};
+        if ({ix} < 0) {{
+            {ix} += {xsize};
+        }}'''.format(ix=ix, xsize=xsize)
+    elif mode == 'constant':
+        ops = '''
+        if ({ix} >= {xsize}) {{
+            {ix} = -1;
+        }}'''.format(ix=ix, xsize=xsize)
+    return ops
+
+
+_CAST_FUNCTION = """
+// Implements a casting function to make it compatible with scipy
+// Use like cast<to_type>(value)
+// It's actually really simple - most of this is <type_traits>
+
+// Small bit of <type_traits> which cannot be imported in NVRTC
+// Requires compiling with --std=c++11 or higher
+template<bool B, class T=void> struct enable_if {};
+template<class T> struct enable_if<true, T> { typedef T type; };
+template<class T> struct remove_const          { typedef T type; };
+template<class T> struct remove_const<const T> { typedef T type; };
+template<class T> struct remove_volatile             { typedef T type; };
+template<class T> struct remove_volatile<volatile T> { typedef T type; };
+template<class T> struct remove_cv {
+  typedef typename remove_volatile<typename remove_const<T>::type>::type type;
+};
+template<class T, T v>
+struct integral_constant { static constexpr T value = v; };
+typedef integral_constant<bool, true> true_type;
+typedef integral_constant<bool, false> false_type;
+template<class T> struct __is_fp : public false_type {};
+template<>        struct __is_fp<float16> : public true_type {};
+template<>        struct __is_fp<float> : public true_type {};
+template<>        struct __is_fp<double> : public true_type {};
+template<>        struct __is_fp<long double> : public true_type {};
+template<class T> struct is_floating_point
+    : public __is_fp<typename remove_cv<T>::type> {};
+template<class T> struct is_signed : integral_constant<bool, (T)(-1)<0> {};
+
+template <typename B, typename A>
+__device__
+typename enable_if<!is_floating_point<A>::value||is_signed<B>::value, B>::type
+cast(A a) { return (B)a; }
+
+template <typename B, typename A>
+__device__
+typename enable_if<is_floating_point<A>::value&&!is_signed<B>::value, B>::type
+cast(A a) { return (a >= 0) ? (B)a : -(B)(-a); }
+
+
+"""
+
+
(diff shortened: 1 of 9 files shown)


coderforlife:
I have now added the tests but would still like feedback on the design.

coderforlife:
@emcastillo Can you give any feedback on the design of these two new functions?

emcastillo:
I'll try to do the first pass soon, sorry for the delay

coderforlife:
Great!

This is only 2 "public" functions. The one function does have 3 major components to it, and if I had left off fused kernels it would have cut the total code in about half or more. It's just hard since I have a ton of code ready to commit, and there are various inter-dependencies. However, in this case it clearly could have been committed without generic_filter1d and with only support for raw (and maybe reduction) kernels to get it started.

leofang:
cc: @grlee77

cupyx/scipy/ndimage/filters.py
```diff
@@ -969,381 +993,147 @@ def _get_rank_kernel(filter_size, rank, mode, wshape, origins, cval, int_type):
...
-        args.append(weights)
-    if structure is not None:
-        structure = cupy.ascontiguousarray(structure, structure_dtype)
-        args.append(structure)
+    # This filter is very different than all other filters (including
+    # generic_filter and all 1d filters) and it has a customized solution.
+    # It is also likely fairly terrible, but only so much can be done when
+    # matching the scipy interface of having the sub-kernel work on entire
+    # lines of data.
+    from cupyx.scipy.ndimage._filters_generic import _get_generic_filter1d
```
grlee77, line 1120: I think the CuPy style will require moving this to the top of the file and using

`from cupyx.scipy.ndimage import _filters_generic`
and then calling via `_filters_generic._get_generic_filter1d`.

cupyx/scipy/ndimage/filters.py
```diff
@@ -1,6 +1,12 @@
 import cupy
 
 
+from cupyx.scipy.ndimage._filters_core import (
```
grlee77, line 4: as explained in more detail above, this will need to be:

```Python
from cupyx.scipy.ndimage import _filters_core
```

cupyx/scipy/ndimage/filters.py
```diff
@@ -969,381 +993,147 @@ def _get_rank_kernel(filter_size, rank, mode, wshape, origins, cval, int_type):
...
+
+          * resolve to a :class:`cupy.ReductionKernel` that meets the
+            requirements above
+          * resolve to a `FusedKernel` that takes 1 array input argument and
+            returns 1 'scalar' output
+
+    .. seealso:: :func:`scipy.ndimage.generic_filter`
     """
-    output_orig = output
+    from cupyx.scipy.ndimage._filters_generic import (
```
grlee77, line 1055: I think the CuPy style may require moving this to the top of the file and using

`from cupyx.scipy.ndimage import _filters_generic`
and then calling via `_filters_generic._get_generic_filter_raw`, etc.

grlee77:
As a note to other reviewers:

The new code that needs to be reviewed is everything in `_filters_generic.py` and the two functions `generic_filter` and `generic_filter1d` within `filters.py`.

I think all code in `_filters_core.py` is just pre-existing code that was split out from `filters.py` and has not been otherwise modified in this PR.

I support the general approach of splitting across these files as I think it is a logical organization and is preferable to one massive file. 

I will try to revisit this PR again over the next couple of days.

cupyx/scipy/ndimage/_filters_generic.py
```diff
@@ -0,0 +1,538 @@
+import cupy
+
+
+from cupyx.scipy.ndimage._filters_core import _generate_nd_kernel
```
grlee77, line 4:
```suggestion
from cupyx.scipy.ndimage import _filters_core
```


based on the [CuPy style guide](https://docs.cupy.dev/en/v4.5.0/contribution.html#coding-guidelines)
```
    Relative imports are prohibited. [H304]
    Importing non-module symbols is prohibited.
    Import statements must be organized into three parts: standard libraries, third-party libraries, and internal imports. [H306]
```

cupyx/scipy/ndimage/_filters_core.py
```diff
@@ -0,0 +1,368 @@
...
+        raise ValueError('invalid origin')
+    return origin
+
+
+def _origins_to_offsets(origins, w_shape):
+    return tuple(x//2+o for x, o in zip(w_shape, origins))
+
+
+def _check_mode(mode):
+    if mode not in {'reflect', 'constant', 'nearest', 'mirror', 'wrap'}:
```
leofang, line 44: I think the style preference is to use tuple; the uniqueness is already guaranteed:
```suggestion
    if mode not in ('reflect', 'constant', 'nearest', 'mirror', 'wrap'):
```

cupyx/scipy/ndimage/_filters_core.py
```diff
@@ -0,0 +1,368 @@
...
+    return weights, tuple(origins)
+
+
+def _check_nd_args(input, weights, mode, origin, wghts_name='filter weights'):
+    if input.dtype.kind == 'c':
+        raise TypeError('Complex type not supported')
+    _check_mode(mode)
+    # Weights must always be less than 2 GiB
+    if weights.nbytes >= (1 << 31):
+        raise RuntimeError('weights must be 2 GiB or less, use FFTs instead')
```
leofang, line 111: Just curious, is it a simple rule of thumb, or is it something well-known to the image processing community?

cupyx/scipy/ndimage/_filters_core.py
```diff
@@ -0,0 +1,368 @@
...
+        return output
+    # We can't operate in-place efficiently, so use a 2-buffer system
+    temp = _get_output(output.dtype, input) if n_filters > 1 else None
+    first = True
+    iterator = zip(filters, args, modes, origins)
+    for axis, (fltr, arg, mode, origin) in enumerate(iterator):
+        if fltr is None:
+            continue
+        fltr(input, arg, axis, output, mode, cval, origin)
+        input, output = output, temp if first else input
```
leofang, line 144: Isn't `first` always `True`?

cupyx/scipy/ndimage/_filters_core.py
```diff
@@ -0,0 +1,368 @@
...
+    The args is a list of values that are passed for the arg value to the
+    filter. Individual filters can be None causing that axis to be skipped.
+    """
+    output_orig = output
+    output = _get_output(output, input)
+    modes = _fix_sequence_arg(mode, input.ndim, 'mode', _check_mode)
+    origins = _fix_sequence_arg(origin, input.ndim, 'origin', int)
+    n_filters = sum(filter is not None for filter in filters)
+    if n_filters == 0:
+        output[...] = input[...]
```
leofang, line 134: We likely can just do this?
```suggestion
        output[...] = input
```

cupyx/scipy/ndimage/_filters_generic.py
```diff
@@ -0,0 +1,538 @@
...
+                       '{}*{}'.format(prod, itemsize))
+        prod *= x
+    return tuple(strides[::-1])
+
+
+def _prod(array):
+    prod = 1
+    for x in array:
+        prod *= x
+    return prod
```
leofang, line 399: There's `internal.prod`.

cupyx/scipy/ndimage/_filters_core.py
grlee77: I think this was just a hard requirement because the kernels currently use 32-bit ints for indexing the weights. The cases where the `ndimage` filters perform well are when the `weights` kernel is much smaller than the image itself. That is a common usage scenario, so this is a reasonable assumption. For cases where the assumption does not hold, one should probably be using FFT-based filtering instead.

coderforlife:
grlee77 is correct in that much of this is simply from the breaking up of the monolithic file into a few separate files.

@leofang agreed that some of these seem more general than image processing, but I have no idea where they should go. Also, I am unaware of what other internal functions are out there. I have found some of them, but cupy is quite large... any that you see (like the `internal.prod` you pointed out) would be great. Anything you think should be moved to a more general location please inform me what and where.

@leofang which reduction kernel? I don't have any reduction kernels... unless you mean the `RawKernel` that takes the specified axis from `n` long, increases it to `n+f+1` temporarily within the kernel, and then reduces it back to `n` long, but that reduction is an outside kernel. The only other kernel I have written is the `ElementwiseKernel` which is the major workhorse here, which along the way applies a reduction kernel, but that kernel has to be integrated into the `ElementwiseKernel`. The main parallelization is across 'pixels' (*samples* in the signal processing world) and not within the reduction needed for each pixel. Although, I would be willing to try to see if those reductions could be done with CUB, at that point I would assume that the GPU is already full taxed with dealing with each individual pixel.

cupyx/scipy/ndimage/_filters_core.py
coderforlife: grlee77 is correct, this was an optimization made for the by far most common (possibly always) situation that grlee77 points out. This does only apply to weights (the smaller of the two inputs), the other input is only restricted by available memory. The optimization probably doesn't have a huge impact on speed, so I could lift it if desired.

cupyx/scipy/ndimage/_filters_core.py
coderforlife: Copy-paste error (didn't copy the last line from another place...). Thanks for catching it!

cupyx/scipy/ndimage/_filters_core.py
coderforlife: Probably, pretty sure that line came straight from scipy.

coderforlife:
Some functions that may be more general:
* `_fix_sequence_args` - this is a fairly general function to be used whenever a function accepts a scalar or iterable of scalars
* `_get_inttype` - discussed with someone (I think @emcastillo) about using this in the `CArray` implementation for deciding what type to choose for indices
* a bunch of `_filters_generic` could be refactored out to provide general support for ripping apart kernels and converting them to C/C++ code for other kernels to be able to reuse, thinking of something like fuse functions being able to integrate from arbitrary (non-raw) kernels
* `_get_output` - in scipy this is actually shared by all of `ndimage`, it is likely that the other `ndimage` files have their own version
* `_check_mode` - same as above
* `_check_origin` - same as above, only needed in filters and morphology though

coderforlife:
@emcastillo (or maybe someone else, I just think he has had a bunch to do with `CArray` and `CIndexer`): is there a reason `CArray` and `CIndexer` have no constructor in C++?

cupyx/scipy/ndimage/_filters_core.py
leofang: Ah, I see. I missed the 32-bit indexing thing. Thanks for jumping in @grlee77 @coderforlife 🙂

cupyx/scipy/ndimage/_filters_core.py
```diff
@@ -0,0 +1,378 @@
+import warnings
+
+import cupy
+
+
+def _get_output(output, input, shape=None):
```
emcastillo, line 6: I think we can use `zeros_like`instead of this function
we can override shape and dtype

https://docs.cupy.dev/en/stable/reference/generated/cupy.zeros_like.html

cupyx/scipy/ndimage/_filters_generic.py
```diff
@@ -0,0 +1,533 @@
...
+        # We will assume that it has the correct API
+        return f
+    elif isinstance(f, cupy.ReductionKernel):
+        if f.nin != 1 or f.nout != 1:
+            raise TypeError('ReductionKernel must have 1 input and output')
+        return f
+    elif isinstance(f, cupy.ElementwiseKernel):
+        # special error message for ElementwiseKernels
+        raise TypeError('only ReductionKernel allowed (not ElementwiseKernel)')
+    elif isinstance(f, cupy.core._fusion_kernel.FusedKernel):
```
emcastillo, line 41: I think we can remove fused kernels from this PR and add a follow-up or just for this feature.
I think there is a lot to discuss with the fusion that might need changes in the fusion interface too.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_filters.py
```diff
@@ -253,6 +256,157 @@ def test_filter(self, xp, scp):
...
+    for (int i = 0; i < filter_size; ++i) { ss += x[i]*x[i]; }
+    y[0] = ss/filter_size;
+}''', 'rms')
+rms_red = cupy.ReductionKernel('X x', 'Y y', 'x*x',
+                               'a + b', 'y = a/_in_ind.size()', '0', 'rms')
+
+
+def rms_fuse_wrapper(filter_size):
+    def rms_fuse(x):
+        return (x*x).sum()/filter_size
```
emcastillo, line 272: Just formatting, please add a space between the operator and the operands.

emcastillo:
Left some comments on my first pass.

The code for supporting fusion kernels is a bit rough, but most probably this is due to the way that fusion is being performed now. I think that to fix this we need to change the fusion interface itself?

Besides that, I don't have any major concerns with the PR. Maybe `_broadcast` or `_transpose` could be moved to `cupy.core` or `cupy._kernel`.

Regarding why CArray and CIndexers has no constructors. I have currently no idea about this, let me ask the team and I will update the reply

cupyx/scipy/ndimage/_filters_generic.py
emcastillo: We have been discussing here that some of the Fusion internal details are private and the only class that is accessible when a user creates a fused kernel is `cupy.core.fusion.Fusion` so this might help a bit to simplify the fusion handling

cupyx/scipy/ndimage/_filters_generic.py
```diff
@@ -0,0 +1,533 @@
...
+    elif callable(f):
+        kernel = cupy.fuse(f)
+    else:
+        raise TypeError('bad function type')
+    return _get_sub_kernel(kernel, filter_size, in_dtype)
+
+
+def _get_fused_kernel(func, filter_size, in_dtype):
+    # This may have to call the fused function so that it generates the
+    # reduction kernel for our datatype
```
emcastillo, line 66: I think that these functions can be joined and moved to cupy/core/_fusion* as they are quite generic.

cupyx/scipy/ndimage/_filters_core.py
coderforlife: Not exactly. This function, if given an array for `output` returns that array, not a newly allocated array. However, the `else` can probably be replaced with `cupy.zeros_like(input, dtype=output, shape=shape)`.

cupyx/scipy/ndimage/_filters_generic.py
coderforlife: Okay, will do that tomorrow.

cupyx/scipy/ndimage/_filters_generic.py
coderforlife: Yeah, basically getting the underlying callable for a given set of inputs (preferably with just shapes/dtypes of inputs, but understandable if it has to be the actual inputs themselves) but without actually calling it. It would be very nice if that were available on fused functions.

coderforlife:
Yeah, I know the fusion kernel stuff is rough, but I was trying to support Python functions. I will remove it for now and we can discuss elsewhere what the interface would need to be.

The functions `_indexing`, `_broadcast_to`, and `_transpose` are only used for fusion kernels so they are actually about to be temporarily removed. However, the goal of all 3 of them is to do those 3 operations on the shape/strides of an array but without needing an actual array, just the shape/strides. In fact, I copied most of the the code from the real versions of those functions, just without the final creation of an array view or error-checking (I removed error-checking since it will have already occurred through the real functions at an earlier point). So maybe those functions could just be refactored to have a version that works on shapes/strides and then a version that calls the shape/stride version and just creates the view from it?

Right now you do the memory layout for `CArray` and `CIndexer` in Python. The code I have does it in C, just annoying to set each element in an array that is being casted from an object to some random int/pointer type... I think to actually make it useful from the C++ side you would need to use initializer lists, but not sure.

emcastillo:
Thanks for the reply!
Seems reasonable to me.

Regarding the `CArray`, you can submit a PR to initialize these in C++ if you like and we can discuss about this in that PR.
Currently we don't have any strong argument to support the way we do it, so if your approach makes life easier we are open to change it.

emcastillo:
Can you mark it as ready for review?
I would like to merge this today if I don't found issues with it :)

coderforlife:
Forgot to do that after removing fusion code... thanks!

emcastillo:
Jenkins, test this please

coderforlife:
It appears as though a bunch of the functions I moved into _filters_core are used by other modules in ndimage. I think these should actually be moved into common file in ndimage. Scipy has `scipy.ndimage_ni_support` for this and most of these functions are there. How about we call it `cupyx.scipy.ndimage._util`?

Functions that I saw had this issue:
 * `_fix_sequence_arg`
 * `_generate_boundary_condition_ops`

Additional functions that I think should be shared:
 * `_get_output`
 * `_check_origin`
 * `_check_mode`
 * `_get_inttype`
 * `_generate_indices_ops`

emcastillo:
`cupyx.scipy.ndimage._util` sounds good to me.

coderforlife:
Code has been moved around. I have run all of the ndimage tests now and they all passed so let's hope they all pass here as well.

emcastillo:
Jenkins, test this please

GenevieveBuckley:
I notice the function signature for `cupyx.scipy.ndimage.generic_filter` is identical to `scipy.ndimage.generic_filter`, except for the inclusion of `extra_arguments=()` and `extra_keywords=None` (which aren't exposed or available in the cupy version).

I assume they're not there because it was hard to do, perhaps @coderforlife can share some insight?

Context: @jakirkham and I have been adding support for cupy arrays to dask-image, and currently we always try to pass along `extra_arguments` and `extra_keywords` and that's how I noticed this.

coderforlife:
@GenevieveBuckley The problem is that the "function" given to `cupyx.scipy.ndimage.generic_filter` is actually a CUDA function. This function definitely won't support `extra_keywords`. It would be possible to support `extra_arguments` although the argument types would definitely be limited and it would likely still be very tricky/difficult to implement.

GenevieveBuckley:
> 
> 
> @GenevieveBuckley The problem is that the "function" given to `cupyx.scipy.ndimage.generic_filter` is actually a CUDA function. This function definitely won't support `extra_keywords`. It would be possible to support `extra_arguments` although the argument types would definitely be limited and it would likely still be very tricky/difficult to implement.

Right, I think our best option is what we're doing now, which is to ignore any `extra_keywords`/`extra_arguments` if we're dispatching to cupy. It is *possible* that a user might become confused, and we might think about adding some clarification to our own docs if we find people are running into it (I suspect this might not be a very commonly used feature though) 

Thanks @coderforlife!

PR #3617, opened 2020-07-15
Fix float16 ndarray input in histogram with CUB

okuta:
In my environment, histogram test is failed with float16.
I use GeForce GTX Titan X (sm_52). 

```
% pytest tests/cupy_tests/statics_tests/test_histogram.py -s
============================================================================================== test session starts ==============================================================================================
platform linux -- Python 3.7.6, pytest-4.1.1, py-1.8.1, pluggy-0.13.1
rootdir: /home/okuta/env/chainer37/cupy, inifile: setup.cfg
collected 88 items

tests/cupy_tests/statics_tests/test_histogram.py .....................................................................................dtype is <class 'numpy.float16'>
F..

=================================================================================================== FAILURES ====================================================================================================
________________________________________________________________________________________ TestCubHistogram.test_histogram ________________________________________________________________________________________
cupy/testing/helper.py:659: in test_func
    impl(self, *args, **kw)
cupy/testing/helper.py:253: in test_func
    check_func(cupy_r, numpy_r)
cupy/testing/helper.py:488: in check_func
    array.assert_array_equal(x, y, err_msg, verbose, strides_check)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), y = array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1]), err_msg = '', verbose = True, strides_check = False

    def assert_array_equal(x, y, err_msg='', verbose=True, strides_check=False):
        """Raises an AssertionError if two array_like objects are not equal.

        Args:
             x(numpy.ndarray or cupy.ndarray): The actual object to check.
             y(numpy.ndarray or cupy.ndarray): The desired, expected object.
             strides_check(bool): If ``True``, consistency of strides is also
                 checked.
             err_msg(str): The error message to be printed in case of failure.
             verbose(bool): If ``True``, the conflicting values
                 are appended to the error message.

        .. seealso:: :func:`numpy.testing.assert_array_equal`
        """
        numpy.testing.assert_array_equal(
            cupy.asnumpy(x), cupy.asnumpy(y), err_msg=err_msg,
>           verbose=verbose)
E       AssertionError:
E       Arrays are not equal
E
E       Mismatched elements: 10 / 10 (100%)
E       Max absolute difference: 1
E       Max relative difference: 1.
E        x: array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
E        y: array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1])

cupy/testing/array.py:93: AssertionError
====================================================================================== 1 failed, 87 passed in 1.94 seconds ======================================================================================```

### DIFF
--- a/cupy/statistics/histogram.py
+++ b/cupy/statistics/histogram.py
@@ -6,6 +6,7 @@ import numpy
 import cupy
 from cupy import core
 from cupy.core import _accelerator
+from cupy.cuda import device
 # TODO(leofang): always import cub when hipCUB is supported
 if not cupy.cuda.runtime.is_hip:
     from cupy.cuda import cub
@@ -223,7 +224,8 @@ def histogram(x, bins=10, range=None, weights=None, density=False):
                     bin_type = numpy.float
                 else:
                     # TODO(okuta): support float16
-                    if bin_type == numpy.float16:
+                    if (bin_type == numpy.float16 and
+                            int(device.get_compute_capability()) < 53):
                         bin_type = numpy.float32
                     x = x.astype(bin_type, copy=False)
                 acc_bin_edge = bin_edges.astype(bin_type, copy=True)
@@ -264,7 +266,7 @@ def histogram(x, bins=10, range=None, weights=None, density=False):
 
     if density:
         db = cupy.array(cupy.diff(bin_edges), cupy.float64)
-        return y/db/y.sum(), bin_edges
+        return y / db / y.sum(), bin_edges
     return y, bin_edges
 
 
--- a/cupy/statistics/histogram.py
+++ b/cupy/statistics/histogram.py
@@ -217,29 +217,24 @@ def histogram(x, bins=10, range=None, weights=None, density=False):
                     and x.size <= 0x7fffffff and bin_edges.size <= 0x7fffffff):
                 # Need to ensure the dtype of bin_edges as it's needed for both
                 # the CUB call and the correction later
-                if isinstance(bins, cupy.ndarray):
-                    bin_type = cupy.result_type(bin_edges, x)
-                    if cupy.issubdtype(bin_type, cupy.integer):
-                        bin_type = cupy.result_type(bin_type, float)
-                    bin_edges = bin_edges.astype(bin_type, copy=False)
+                assert isinstance(bin_edges, cupy.ndarray)
+                bin_type = numpy.result_type(bin_edges.dtype, x.dtype)
+                if numpy.issubdtype(x.dtype, numpy.integer):
+                    bin_type = numpy.float
+                else:
+                    # TODO(okuta): support float16
+                    if bin_type == numpy.float16:
+                        bin_type = numpy.float32
+                    x = x.astype(bin_type, copy=False)
+                acc_bin_edge = bin_edges.astype(bin_type, copy=True)
                 # CUB's upper bin boundary is exclusive for all bins, including
                 # the last bin, so we must shift it to comply with NumPy
                 if x.dtype.kind in 'ui':
-                    bin_edges[-1] += 1
+                    acc_bin_edge[-1] += 1
                 elif x.dtype.kind == 'f':
-                    old_edge = bin_edges[-1].copy()
-                    bin_edges[-1] = cupy.nextafter(bin_edges[-1],
-                                                   bin_edges[-1] + 1)
-                y = cub.device_histogram(x, bin_edges, y)
-                # shift the uppermost edge back
-                if x.dtype.kind in 'ui':
-                    bin_edges[-1] -= 1
-                elif x.dtype.kind == 'f':
-                    bin_edges[-1] = old_edge
-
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by leofang** [comment]
Context: https://github.com/cupy/cupy/pull/3542#issuecomment-658931822

**Comment by leofang** [comment]
@okuta Perhaps we should close this one and move to #3618?

**Comment by okuta** [comment]
OK, please.

**Comment by okuta** [comment]
#3618 does not sole float16 problem. But, please review #3618 first.
There may be a problem on the cub side.

**Comment by leofang** [comment]
@okuta I figured it out. Half precision is supported starting sm_53, so it was not built for your device. We now have a check function `cupy.cuda.common_is_fp16_supported()`, could you call it in the `for accelerator in ...` loop and do something like this?
```python
if x.dtype == numpy.float16 and not cupy.cuda.common_is_fp16_supported():
    continue
```
This will make it fall back to CuPy's implementation.

takagi:
I will revisit this after #3618 is merged.

leofang:
By the way, in @okuta's case you should have encountered a harder error:
https://github.com/cupy/cupy/blob/aee774b0b848a7bd9b08c8ae8130640307fd08ff/cupy/core/include/cupy/type_dispatcher.cuh#L55-L56
I am guessing the reason you didn't see it is because we also wrote `!defined(__CUDA_ARCH__)` above (it was added in https://github.com/cupy/cupy/pull/2600#issuecomment-549230769), so when the dispatcher (running on the host) can still send you to a no-op __half kernel.

@okuta Could you confirm you would get `runtime_error` by removing `!defined(__CUDA_ARCH__)`? If so, we should also fix it...(I don't have good idea for the moment, though.)

okuta:
I already tried it. But, `__CUDA_ARCH__` is not defined in host compiling mode...

kmaehashi:
@takagi Could you takeover this PR?

leofang:
I can help if you need. 🙂

okuta:
I got same error with GeForce RTX 2080 (master branch).

leofang:
@okuta Any chance you could try rebuilding your CuPy? We've seen in the past that a dirty build could lead to random failures in the CUB module that are impossible to debug...If this passes with a clean state, all we need is to add a `cub_histogram` function to `cupy.cuda.cub` that provides a dtype guard, similar to `cub_reduction` and `cub_scan` there.

leofang:
In particular,

> I got same error with GeForce RTX 2080 (master branch).

This should not happen. The CI covers it well with sm > 53.

okuta:
Sorry, late reply.
I added CC checking code.

cupy/statistics/histogram.py
```diff
@@ -217,29 +218,25 @@ def histogram(x, bins=10, range=None, weights=None, density=False):
...
-                        bin_type = cupy.result_type(bin_type, float)
-                    bin_edges = bin_edges.astype(bin_type, copy=False)
+                assert isinstance(bin_edges, cupy.ndarray)
+                bin_type = numpy.result_type(bin_edges.dtype, x.dtype)
+                if numpy.issubdtype(x.dtype, numpy.integer):
+                    bin_type = numpy.float
+                else:
+                    # TODO(okuta): support float16
+                    if (bin_type == numpy.float16 and
+                            int(device.get_compute_capability()) < 53):
```
leofang, line 228: Nitpick: Could you call `cupy.cuda.common._is_fp16_supported()` instead?

cupy/statistics/histogram.py
```diff
@@ -217,29 +218,25 @@ def histogram(x, bins=10, range=None, weights=None, density=False):
...
+                        bin_type = numpy.float32
+                    x = x.astype(bin_type, copy=False)
```
leofang, line 230: Is this for `x.dtype == numpy.float16`? If so, it should be indented following the line above. Otherwise, I don't see why we need a type conversion here 😅

cupy/statistics/histogram.py
okuta: Thanks! I will use this.

cupy/statistics/histogram.py
okuta: I will add a corresponding test case.

okuta:
I fixed.

leofang:
Jenkins, test this please

takagi:
Let me merge this to the master!

PR #3626, opened 2020-07-20
Add `__name__` to custom kernels

### LINKED ISSUE #3527: timing kernels with cupyx.time.repeat requires specifying a name
I have a question about convenience when using `cupyx.time.repeat` with kernels such as a `ReductionKernel` or `ElementwiseKernel`. Unless the user specifies something like `name=kernel.name` as an argument to `repeat`, it will fail when trying to access a non-existant `__name__` attribute.

Having to specify the name manually is a little inconvenient. What do you think about any of the following solutions?:

1.) try falling back to a `.name` attribute
2.) fall back to a default name (e.g.  `name=''"` or `name="unknown"`)
3.) add a `__name__` attribute to `_AbstractReductionKernel`, `ElementwiseKernel`, etc. 

minimal example:

```Python
import cupy
from cupyx.time import repeat

x = cupy.arange(8)
e = cupy.core.ElementwiseKernel('X x', 'Y y', '2 * x', 'doubler')
repeat(e, (x,))
```

```Python

  File "<ipython-input-11-2afc789db431>", line 6, in <module>
    repeat(e, (x,))

  File "/home/lee8rx/my_git/pyir/cupy/cupyx/time.py", line 53, in repeat
    name = func.__name__

AttributeError: 'cupy.core._kernel.ElementwiseKernel' object has no attribute '__name__'
```

emcastillo:
Closes #3527

### DIFF
--- a/cupy/core/_kernel.pyx
+++ b/cupy/core/_kernel.pyx
@@ -684,7 +684,8 @@ cdef class ElementwiseKernel:
         readonly Py_ssize_t nargs
         readonly tuple params
         readonly object operation
-        readonly object name
+        readonly str name
+        readonly str __name__
         readonly bint reduce_dims
         readonly object preamble
         readonly bint no_return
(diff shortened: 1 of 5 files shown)


kmaehashi:
pfnCI, test this please.

PR #3633, opened 2020-07-20
Fix sparse argmin/argmax return shape

takagi:
Merge #3638 first.

This PR fixes sparse argmin/argmax to return ndarrays that have the same shape as their SciPy counterparts. Additionally, I've parameterized the test to make it simpler.


takagi:
Close this for the duplicate fix.

takagi:
Close this as I reposted #3639.

PR #3638, opened 2020-07-21
Add tests for sparse max with `axis=None`

takagi:
Rel #3515. Merge #3536 first.

This PR adds missing tests for sparse max with `axis=None`.

PR #3497, opened 2020-06-25
Support negative `axis` in sparse min/max/argmin/argmax

takagi:
Close #3474.

~~This PR fixes sparse `min` and `max` to take a negative `axis`. I'll parameterize its test in another PR.~~

Merge #3656 first. This PR fixes sparse min/max/argmin/argmax to take a negative `axis`.

PR #3639, opened 2020-07-21
Fix sparse argmin/argmax return shape

takagi:
Merge #3638 first.

This PR fixes the return shapes of sparse argmin/argmax.

### DIFF
--- a/cupyx/scipy/sparse/data.py
+++ b/cupyx/scipy/sparse/data.py
@@ -211,7 +211,10 @@ class _minmax_mixin(object):
         # Do the reudction
         value = mat._arg_minor_reduce(op, axis)
 
-        return value
+        if axis == 0:
+            return value[None, :]
+        else:
+            return value[:, None]
 
     def _arg_min_or_max(self, axis, out, op, compare, sum_duplicates):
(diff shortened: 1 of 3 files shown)


cupyx/scipy/sparse/data.py
```diff
@@ -211,7 +211,10 @@ def _arg_min_or_max_axis(self, axis, op, sum_duplicates):
         # Do the reudction
         value = mat._arg_minor_reduce(op, axis)
 
-        return value
+        if axis == 0:
```
leofang, line 214: Note to myself: `_arg_min_or_max()` says `axis` is not `None` at this point, and currently we don't support negative axis until #3497, so it can only be 0 or 1 here.

leofang:
Jenkins, test this please

toslunar:
The Commits tab of GitHub (`git diff master...sparse-argmin-argmax-shape`) shows three commits, while #3638 has been merged to `master`.  I'll `git merge master` on the branch for clarity.

toslunar:
LGTM, too.  Thanks, @leofang.

Jenkins, test this please.

Issue #3631, opened 2020-07-20
Drop support for SciPy 1.2 or earlier


kmaehashi:
We should consider dropping support for older SciPy like we did for NumPy.

- [ ] CI https://github.com/chainer/chainer-test/pull/600
- [x] Clarify the supported version in docs

kmaehashi:
In v8, we are going to drop support for SciPy 1.12 or earlier.

There's no policy like NEP 29 in SciPy, so we picked the version considering the release date.

* SciPy 1.12 released on Dec 18, 2018
* NumPy 1.16 released on Jan 14, 2019 (minimum version we support in v8 (https://github.com/cupy/cupy/issues/3643), following NEP 29)
* SciPy 1.13 released on May 18, 2019
* SciPy 1.14 released on Dec 17, 2019

kmaehashi:
Closed via https://github.com/chainer/chainer-test/pull/629

PR #3641, opened 2020-07-21
Adding most of scipy.linalg.special_matrices.

coderforlife:
This adds 17 out of 20 of the functions in `scipy.linalg.special_matrices`. The only missing ones are currently implemented in scipy with `scipy.special.comb` which is currently not implemented in cupy (and would also run into integer overflow issues on the GPU quickly as well). Additionally, the `tri` function actually just calls the already efficiently implemented `cupy.tri` function.

Will be adding tests shortly, however, what is the earliest still-supported scipy version? These functions were added later:
* 0.8.0 - `circulant`, `hadamard`, `leslie`, `companion`
* 0.10.0 - `hilbert`
* 0.14.0 - `dft`
* 1.3.0 - `fiedler`, `fiedler_companion`
* 1.5.0 - `convolution_matrix`


coderforlife:
Looking through the tests, the requirement "scipy>=" is used at earliest with 0.19.0 so I added testing rules for only the final 3 special matrix functions.

grlee77:
I don't see a specific SciPy version listed as supported in the docs, but I think scipy 0.18 is the oldest version tested on CI:
https://github.com/chainer/chainer-test/blob/master/run_multi_test.py#L22

kmaehashi:
Could you split the PR for each function?

Regarding SciPy version support, we discussed and planning to only support 1.3+. (#3631)

coderforlife:
@kmaehashi I can split it up a bit, but some of them are related to each other so I would probably still keep them together (like `tri`/`tril`/triu` and `toeplitz`/`circulant`/`hankel`). However, currently the tests are currently a bit intertwined.

How should I proceed? Do I make 15 different branches and submit 15 PRs right now for all of them, or do I do it one PR at a time and can reuse the same branch? Will the merging handle it just fine if they are done as 15 separate branches? It seems like it will think they should all be the "first" function in the file.

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,551 @@
...
+
+
+def tri(N, M=None, k=0, dtype=None):
+    """ Construct (N, M) matrix filled with ones at and below the kth diagonal.
+    The matrix has A[i,j] == 1 for i <= j + k
+
+    Args:
+        N (int): The size of the first dimension of the matrix.
+        M (int, optional): The size of the second dimension of the matrix. If
+            `M` is None, `M = N` is assumed.
```
asi1024, line 18: Please use double-backquotes for the variables. Could you check locally the documentation builds correctly?
```suggestion
            ``M`` is None, ``M = N`` is assumed.
```

leofang: One of the CI (`docs/readthedocs.org:cupy`) builds the doc for every PR. You can also browse the rendered outcome there.

asi1024:
Could you update `docs/source/reference/linalg.rst`?

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,552 @@
...
+    .. seealso:: :func:`cupyx.scipy.linalg.fielder_companion`
+    .. seealso:: :func:`scipy.linalg.companion`
+    """
+    n = a.size
+    if a.ndim != 1:
+        raise ValueError("`a` must be one-dimensional.")
+    if n < 2:
+        raise ValueError("The length of `a` must be at least 2.")
+    if a[0] == 0:
+        raise ValueError("The first coefficient in `a` must not be zero.")
```
asi1024, line 329: This check requires a device-to-host synchronization. CuPy does not raise an error in such a situation. For example, `cupy.linalg.inv` does not raise `LinAlgError` for singular matrix input.

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,552 @@
...
+    Args:
+        n (int): The size of the array to create.
+
+    Returns:
+        cupy.ndarray: The Hilbert matrix.
+
+    .. seealso:: :func:`scipy.linalg.hilbert`
+    """
+    values = cupy.arange(1, 2*n, dtype=float)
+    cupy.divide(1.0, values, values)
```
asi1024, line 379: Why not use `cupy.reciprocal`?
```suggestion
    cupy.reciprocal(values, out=values)
```

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,552 @@
...
+
+    Args:
+        n (int): The size of the array to create.
+
+    Returns:
+        cupy.ndarray: The Hilbert matrix.
+
+    .. seealso:: :func:`scipy.linalg.hilbert`
+    """
+    values = cupy.arange(1, 2*n, dtype=float)
```
asi1024, line 378: Please use `cupy.float64` for the readability.
```suggestion
    values = cupy.arange(1, 2*n, dtype=numpy.float64)
```

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,552 @@
...
+    Args:
+        a (cupy.ndarray): 1-D array of polynomial coefficients. The length of
+            ``a`` must be at least two, and ``a[0]`` must not be zero.
+
+    Returns:
+        (cupy.ndarray): The first row of the output is ``-a[1:]/a[0]``, and the
+            first sub-diagonal is all ones. The data-type of the array is the
+            same as the data-type of ``-a[1:]/a[0]``.
+
+    .. seealso:: :func:`cupyx.scipy.linalg.fielder_companion`
```
asi1024, line 320: Fix typo.
```suggestion
    .. seealso:: :func:`cupyx.scipy.linalg.fiedler_companion`
```

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,552 @@
...
+        cupy.ndarray: the symmetric Fiedler matrix
+
+    .. seealso:: :func:`cupyx.scipy.linalg.circulant`
+    .. seealso:: :func:`cupyx.scipy.linalg.toeplitz`
+    .. seealso:: :func:`scipy.linalg.fiedler`
+    """
+    if a.ndim != 1:
+        raise ValueError("Input 'a' must be a 1D array.")
+    if a.size == 0:
+        return cupy.zeros(())
```
asi1024, line 448: Please add a test for this case.

cupyx/scipy/linalg/special_matrices.py
coderforlife: Didn't know it existed! Thanks!

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,552 @@
...
+    lg2 = 0 if n < 1 else (int(n).bit_length() - 1)
+    if 2 ** lg2 != n:
+        raise ValueError("n must be an positive a power of 2 integer")
+
+    # Sylvester's construction
+    H = cupy.ones((1, 1), dtype=dtype)
+    for _ in range(0, lg2):
+        n = H.shape[0]
+        H = cupy.tile(H, (2, 2))
+        H[n:, n:] *= -1
```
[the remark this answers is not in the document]
coderforlife: Tried figuring it out but couldn't do it easily. Likely for each value in the matrix you would have to figure out the number of times it would be negated. The lower-right corner is negated each iteration (after tiling).

cupyx/scipy/linalg/special_matrices.py
```diff
@@ -0,0 +1,552 @@
...
+        dtype (dtype, optional): The data type of the array to be constructed.
+
+    Returns:
+        (cupy.ndarray): The Hadamard matrix.
+
+    .. seealso:: :func:`scipy.linalg.hadamard`
+    """
+    lg2 = 0 if n < 1 else (int(n).bit_length() - 1)
+    if 2 ** lg2 != n:
+        raise ValueError("n must be an positive a power of 2 integer")
```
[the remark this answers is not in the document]
coderforlife: Fixed all quotes.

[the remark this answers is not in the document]
coderforlife: It was tested (with `n` of 10 and 25). I added a test for args `6, float64` as well.

cupyx/scipy/linalg/special_matrices.py
coderforlife: Added and fixed this line to return the correct value.

cupyx/scipy/linalg/special_matrices.py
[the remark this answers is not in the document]
coderforlife: I actually found a clever way to do this one very quickly with a raw kernel. Each value is positive or negative one. The negative values are when the number of bits in `i & j` is odd where `i` is the row and `j` is the column (starting both at 0). After a bit more testing I will commit that version as well.

coderforlife:
Using a custom an `ElementwiseKernel` for `hadamard` has great results (old is the version with a loop and new is the `ElementwiseKernel` version):

`n` | scipy (µs) | old (µs) | new (µs)
---|---|---|---
256 | 151 | 306 | 12.9
512 | 517 | 339 | 15.6
1024 | 3600 | 391 | 25.5
2048 | 24600 | 521 | 65.4
4096 | 120 ms | 994 | 224
8192 | 513 ms | 2880 | 861
16384 | 2.08 s | 10.4 ms | 3.4 ms

The cupy function is now faster than scipy for all values of `n` except 1 (but previously it was only faster `n>=512`).

asi1024:
@coderforlife Could you fix for Travis failure?

coderforlife:
Fixed documentation issues.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #3642, opened 2020-07-22
Fix missing imports and `cupy.show_config`

emcastillo:
After creating `cuda_backends` some imports where not changed and `cupy.show_config` wasn't able to detect cudnn, cutensor and other libraries

### DIFF
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -52,6 +52,12 @@ except ImportError:
     nccl_enabled = False
 
 try:
+    from cupy_backends.cuda.libs import cudnn  # NOQA
+    cudnn_enabled = True
+except ImportError:
+    cudnn_enabled = False
+
+try:
     from cupy_backends.cuda.libs import cutensor  # NOQA
     cutensor_enabled = True
 except ImportError:
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -52,12 +52,6 @@ except ImportError:
     nccl_enabled = False
 
 try:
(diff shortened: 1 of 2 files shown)


cupyx/runtime.py
```diff
@@ -10,12 +10,12 @@
...
 
 try:
-    import cupy.cuda.cudnn as cudnn
+    import cupy_backends.cuda.libs.cudnn as cudnn
 except ImportError:
     cudnn = None
 
 try:
-    import cupy.cuda.nccl as nccl
+    import cupy_backends.cuda.libs.nccl as nccl
```
leofang, line 18: I thought `nccl` is still in `cupy.cuda`?

emcastillo:
Updated, PTAL

leofang:
Jenkins, test this please

cupy/cuda/__init__.py
```diff
@@ -51,6 +51,12 @@
 except ImportError:
     nccl_enabled = False
 
+try:
+    from cupy_backends.cuda.libs import cudnn  # NOQA
```
kmaehashi, line 55: This is not needed as cudnn is not exposed here.
https://github.com/cupy/cupy/blob/v8.0.0b4/cupy/cuda/__init__.py

emcastillo: Previous versions had `import cupy.cuda.cudnn as cudnn` as a valid path.
I think we should add a `cupy.cuda.cudnn` file that imports `cupy_backends.cuda.libs.cudnn` not to break compatibility?

leofang:
Jenkins, test this please

PR #1889, opened 2018-12-12
Add RawKernel.compile() method

leofang:
As discussed in #1657, a `.compile()` method is desired. I implemented it and added two tests.

A few words on this implementation: while it functions as requested, upon reflection I feel the `compile()` method should belong to the module level instead of the `RawKernel` class. The later implies that a CUDA kernel, instead of a module, is returned and used. However, I don't see any other way to do so without breaking backward compatibility, so this is a design choice I leave for you to worry in the future. Should you have any comments or suggestions, I'm more than happy to discuss and make changes. 

Thanks.

PR #3644, opened 2020-07-22
Add a `compile` method for `RawKernel` and `RawModule`

### LINKED ISSUE #3554: Improve printing compiler log to stream
Follows #3485 

This PR introduced a mechanism to use a user opened stream for saving compiler warnings when compiling RawKernels.

However, the stream is passed on the `__init__` method but is not employed until the code is compiled, which can be pretty late in the execution.

We should discuss a way to effectively manage streams in this situation.

emcastillo:
Closes #3554 

Takes over #1889 (thanks @leofang!)
Tried to add tests, but it is hard to assert wether the compiler is called or not using mock due to cython issues and memorization issues.

### DIFF
--- a/cupy/core/raw.pyx
+++ b/cupy/core/raw.pyx
@@ -35,13 +35,11 @@ cdef class RawKernel:
             ``cuLaunchCooperativeKernel`` so that cooperative groups can be
             used from the CUDA source.
             This feature is only supported in CUDA 9 or later.
-        log_stream (object): Pass either ``sys.stdout`` or a file object to
-            which the compiler output will be written.
     """
 
     def __init__(self, str code, str name, tuple options=(),
                  str backend='nvrtc', *, bint translate_cucomplex=False,
-                 bint enable_cooperative_groups=False, log_stream=None):
+                 bint enable_cooperative_groups=False):
 
         self.code = code
         self.name = name
@@ -49,7 +47,6 @@ cdef class RawKernel:
         self.backend = backend
         self.translate_cucomplex = translate_cucomplex
         self.enable_cooperative_groups = enable_cooperative_groups
-        self.log_stream = log_stream
 
         # only used when RawKernels are produced from RawModule
         self.file_path = None  # for cubin/ptx
@@ -83,6 +80,9 @@ cdef class RawKernel:
 
     @property
     def kernel(self):
+        return self._kernel()
+
+    def _kernel(self, log_stream=None):
         # The kernel is cached, so on the device where this has been called,
         # we would just look up from the cache, and do recompiling only when
         # switching to a different device
@@ -100,7 +100,7 @@ cdef class RawKernel:
             mod = _get_raw_module(
                 self.code, self.file_path, self.options, self.backend,
                 self.translate_cucomplex, self.enable_cooperative_groups,
-                self.name_expressions, self.log_stream)
+                self.name_expressions, log_stream)
             ker = mod.get_function(self.name)
             self._kernel_cache[dev] = ker
         return ker
@@ -223,6 +223,21 @@ cdef class RawKernel:
         attr = driver.CU_FUNC_ATTRIBUTE_PREFERRED_SHARED_MEMORY_CARVEOUT
         driver.funcSetAttribute(self.kernel.ptr, attr, fraction)
 
+    def compile(self, log_stream=None):
+        """Compile the current kernel.
+
+        In general, you don't have to call this method;
+        kernels are compiled implicitly on the first call.
+
+        Args:
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written.
+                Defaults to ``None``.
+        """
+        # Flush the cache when compilation is explicitly requested
+        self._kernel_cache = [None] * runtime.getDeviceCount()
+        self._kernel(log_stream=log_stream)
+
 
 cdef class RawModule:
     """User-defined custom module.
@@ -245,7 +260,7 @@ cdef class RawModule:
         options (tuple of str): Compiler options passed to the backend (NVRTC
             or NVCC). For details, see
             https://docs.nvidia.com/cuda/nvrtc/index.html#group__options or
-            https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#command-option-description
+            https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#command-option-description.
         backend (str): Either `nvrtc` or `nvcc`. Defaults to `nvrtc`
         translate_cucomplex (bool): Whether the CUDA source includes the header
             `cuComplex.h` or not. If set to ``True``, any code that uses the
@@ -263,8 +278,6 @@ cdef class RawModule:
             the template kernel ``func1<T>`` and non-template kernel ``func2``.
             Strings in this tuple must then be passed, one at a time, to
             :meth:`get_function` to retrieve the corresponding kernel.
-        log_stream (object): Pass either ``sys.stdout`` or a file object to
-            which the compiler output will be written.
 
     .. note::
         Each kernel in ``RawModule`` possesses independent function attributes.
@@ -278,7 +291,7 @@ cdef class RawModule:
     def __init__(self, *, str code=None, str path=None, tuple options=(),
                  str backend='nvrtc', bint translate_cucomplex=False,
                  bint enable_cooperative_groups=False,
-                 name_expressions=None, log_stream=None):
+                 name_expressions=None):
         if (code is None) == (path is None):
             raise TypeError(
                 'Exactly one of `code` and `path` keyword arguments must be '
@@ -303,7 +316,6 @@ cdef class RawModule:
         self.code = code
         self.file_path = path
         self.enable_cooperative_groups = enable_cooperative_groups
-        self.log_stream = log_stream
 
         if self.code is not None:
             self.options = options
@@ -316,15 +328,37 @@ cdef class RawModule:
 
     @property
     def module(self):
+        return self._module()
+
+    def _module(self, log_stream=None):
         # The module is cached, so on the device where this has been called,
         # we would just look up from the cache, and do recompiling only when
         # switching to a different device
         cdef Module mod
+
         mod = _get_raw_module(
             self.code, self.file_path, self.options, self.backend,
             self.translate_cucomplex, self.enable_cooperative_groups,
-            self.name_expressions, self.log_stream)
+            self.name_expressions, log_stream)
         return mod
+
+    def compile(self, log_stream=None):
+        """Compile the current module.
+
+        In general, you don't have to call this method;
+        kernels are compiled implicitly on the first call.
+
+        Args:
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written.
+                Defaults to ``None``.
+
+        .. note::
+            Calling :meth:`compile` will reset the internal state of
+            a :class:`RawKernel`.
+
+        """
+        self._module(log_stream)
 
     def get_function(self, str name):
         """Retrieve a CUDA kernel by its name from the module.
@@ -377,8 +411,8 @@ cdef class RawModule:
         ker = RawKernel(
             self.code, name, self.options, self.backend,
             translate_cucomplex=self.translate_cucomplex,
-            enable_cooperative_groups=self.enable_cooperative_groups,
-            log_stream=self.log_stream)
+            enable_cooperative_groups=self.enable_cooperative_groups)
+
         # for lookup in case we loaded from cubin/ptx
         ker.file_path = self.file_path
         # for lookup in case we specialize a template
@@ -432,10 +466,11 @@ cdef class RawModule:
 
 
 @cupy.util.memoize(for_each_device=True)
-def _get_raw_module(str code, str path, tuple options=(), str backend='nvrtc',
-                    bint translate_cucomplex=False,
-                    bint enable_cooperative_groups=False,
-                    tuple name_expressions=None, log_stream=None):
+def _get_raw_module(str code, str path, tuple options, str backend,
+                    bint translate_cucomplex,
+                    bint enable_cooperative_groups,
+                    tuple name_expressions,
+                    object log_stream):
     cdef Module mod
     if code is not None:
         mod = cupy.core.core.compile_with_cache(
--- a/tests/cupy_tests/core_tests/test_raw.py
+++ b/tests/cupy_tests/core_tests/test_raw.py
@@ -1,4 +1,5 @@
 import contextlib
+import io
 import os
 import sys
 import tempfile
@@ -21,6 +22,16 @@ void test_sum(const float* x1, const float* x2, float* y, unsigned int N) {
     int tid = blockDim.x * blockIdx.x + threadIdx.x;
     if (tid < N)
         y[tid] = x1[tid] + x2[tid];
+}
+'''
+
+_test_compile_src = r'''
+extern "C" __global__
+void test_op(const float* x1, const float* x2, float* y, unsigned int N) {
+    int tid = blockDim.x * blockIdx.x + threadIdx.x;
+    int j;  // To generate a warning to appear in the log stream
+    if (tid < N)
+        y[tid] = x1[tid] OP x2[tid];
 }
 '''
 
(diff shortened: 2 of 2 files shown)


cupy/core/raw.pyx
```diff
@@ -223,6 +228,26 @@ cdef class RawKernel:
...
+        Compiles the kernels in code, and returns a Module instance. If the
+        arguments are not given, the corresponding attributes of the RawKernel
+        instance that were given during initialization are used.
+
+        Args:
+            options (tuple of str): Compiler options passed to the backend (NVRTC
+                or NVCC). For details, see
+                https://docs.nvidia.com/cuda/nvrtc/index.html#group__options or
+                https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#command-option-description
+            backend (str): Either `nvrtc` or `nvcc`. Defaults to `RawKernel.backend`
```
leofang, line 243: Oops, I think we did it wrong before...Could you correct it here as well as in other places?
```suggestion
            backend (str): Either ``'nvrtc'`` or ``'nvcc'``. Defaults to
                :attr:`RawKernel.backend`.
```

cupy/core/raw.pyx
```diff
@@ -316,16 +338,43 @@ cdef class RawModule:
...
         # switching to a different device
         cdef Module mod
+
+        if options is None:
+            options = self.options
+        if backend is None:
+            backend = self.backend
+
         mod = _get_raw_module(
             self.code, self.file_path, self.options, self.backend,
```
leofang, line 355: Same here:
```suggestion
            self.code, self.file_path, options, backend,
```

cupy/core/raw.pyx
```diff
@@ -316,16 +338,43 @@ cdef class RawModule:
...
             self.translate_cucomplex, self.enable_cooperative_groups,
-            self.name_expressions, self.log_stream)
+            self.name_expressions, log_stream)
         return mod
 
+    def compile(self, options=None, backend=None, log_stream=None):
+        """ Compile the current module
```
leofang, line 361:
```suggestion
        """Compile the current module.
```

cupy/core/raw.pyx
```diff
@@ -316,16 +338,43 @@ cdef class RawModule:
...
+        Compiles the kernels in code, and returns a Module instance. If the
+        arguments are not given, the corresponding attributes of the RawKernel
+        instance that were given during initialization are used.
+
+        Args:
+            options (tuple of str): Compiler options passed to the backend (NVRTC
+                or NVCC). For details, see
+                https://docs.nvidia.com/cuda/nvrtc/index.html#group__options or
+                https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#command-option-description
+            backend (str): Either `nvrtc` or `nvcc`. Defaults to `RawKernel.backend`
```
leofang, line 372: ditto

cupy/core/raw.pyx
```diff
@@ -223,6 +228,26 @@ cdef class RawKernel:
...
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written.
+        """
+        return self._kernel(
+            options=options, backend=backend,
+            log_stream=log_stream)
```
leofang, line 249: I think it can simply be like this (as you did for RawModule below):
```suggestion
        return self._kernel(options, backend, log_stream)
```

cupy/core/raw.pyx
```diff
@@ -83,12 +80,20 @@ cdef class RawKernel:
 
     @property
     def kernel(self):
+        return self._kernel()
+
+    def _kernel(self, options=None, backend=None, log_stream=None):
```
leofang, line 85: I think it's OK to cdef this:
```suggestion
    cdef Function _kernel(self, options=None, backend=None, log_stream=None):
```
(btw, Cython doc says `cdef inline` won't be faster for class cdef methods, so it's ok to not inline.)

cupy/core/raw.pyx
```diff
@@ -316,16 +338,43 @@ cdef class RawModule:
 
     @property
     def module(self):
+          return self._module()
+
+    def _module(self, options=None, backend=None, log_stream=None):
```
leofang, line 343: ditto
```suggestion
    cdef Module _module(self, options=None, backend=None, log_stream=None):
```

cupy/core/raw.pyx
```diff
@@ -223,6 +228,26 @@ cdef class RawKernel:
...
+        instance that were given during initialization are used.
+
+        Args:
+            options (tuple of str): Compiler options passed to the backend (NVRTC
+                or NVCC). For details, see
+                https://docs.nvidia.com/cuda/nvrtc/index.html#group__options or
+                https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#command-option-description
+            backend (str): Either `nvrtc` or `nvcc`. Defaults to `RawKernel.backend`
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written.
```
leofang, line 245: Maybe we should mention the default is `log_stream=None`?

cupy/core/raw.pyx
```diff
@@ -316,16 +338,43 @@ cdef class RawModule:
...
```
leofang, line 374: ditto

leofang:
> 2\. In either approach (return or not), the non-default options in `compile()` are gonna have a very bad interplay with the cached kernels/modules compiled with default or different options, as they are not used as part of the key (which is just the device id `dev`). I don't have a good suggestion here...The most trivial thing is to enforce `_kernel()`/`_module()` to be called only once per device, but I am not sure if that's the intent.

Ah, another simple way out is to support `compile()` in `RawModule` only.

leofang:
cc: @mnicely (for awareness)

leofang:
> Tried to add tests, but it is hard to assert wether the compiler is called or not using mock due to cython issues and memorization issues.

@emcastillo As a workaround, you could set different macros in options to compile different version of kernels, and see if the resulting kernel can do what that macro asks for. #1889 has a rough version of this.

emcastillo:
The module `get_function` just recompiles a function everything is accessed, so it's not easy to test.

emcastillo:
Fixed now, everything the options or backend changes, the object stored ones does so accordingly

leofang:
> > 2. In either approach (return or not), the non-default options in `compile()` are gonna have a very bad interplay with the cached kernels/modules compiled with default or different options, as they are not used as part of the key (which is just the device id `dev`). I don't have a good suggestion here...The most trivial thing is to enforce `_kernel()`/`_module()` to be called only once per device, but I am not sure if that's the intent.
> 
> Ah, another simple way out is to support `compile()` in `RawModule` only.

This is being discussed offline with @emcastillo.

leofang:
Jenkins, test this please

cupy/core/raw.pyx
```diff
@@ -223,6 +234,27 @@ cdef class RawKernel:
...
+                (NVRTC or NVCC). For details, see
+                https://docs.nvidia.com/cuda/nvrtc/index.html#group__options or
+                https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#command-option-description  # noqa
+            backend (str): Either ``'nvrtc'`` or ``'nvcc'``. Defaults to
+                attr::`RawKernel.backend`.
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written. Defaults to `None`.
+        """
+        # Flush the cache when compilation is explicitly requested
+        self._kernel_cache = []
```
leofang, line 255: Oops!
```suggestion
        self._kernel_cache = [None] * runtime.getDeviceCount()
```

emcastillo:
Jenkins, test this please

leofang:
Jenkins, test this please

cupy/core/raw.pyx
```diff
@@ -432,10 +499,11 @@ cdef class RawModule:
...
 @cupy.util.memoize(for_each_device=True)
-def _get_raw_module(str code, str path, tuple options=(), str backend='nvrtc',
-                    bint translate_cucomplex=False,
-                    bint enable_cooperative_groups=False,
-                    tuple name_expressions=None, log_stream=None):
+def _get_raw_module(str code, str path, tuple options, str backend,
+                    bint translate_cucomplex,
+                    bint enable_cooperative_groups,
+                    tuple name_expressions,
+                    object log_stream):
```
leofang, line 506: Note: I didn't realize until now that `log_stream` becomes part of the cache key, which is a little odd since it doesn't affect the generated kernels. However, I don't see any real harm here, as in most use cases I expect `log_stream=None`.

leofang:
@kmaehashi PTAL.

cupy/core/raw.pyx
```diff
@@ -223,6 +234,27 @@ cdef class RawKernel:
         attr = driver.CU_FUNC_ATTRIBUTE_PREFERRED_SHARED_MEMORY_CARVEOUT
         driver.funcSetAttribute(self.kernel.ptr, attr, fraction)
 
+    def compile(self, options=None, backend=None, log_stream=None):
```
kmaehashi, line 237: Are these options (`options`, `backend`) needed? I think just recreating the RawKernel is enough for users.

leofang: I think the idea is to compile (or force recompile) an existing `RawKernel` instance so that we can support a development cycle in a console like this:
```python
ker = RawKernel(...)  
ker.compile(options=some_options, log_stream=sys.stdout)
ker((16,), (128,), args)   # test
ker.compile(options=some_other_options, log_stream=sys.stdout)
ker((16,), (128,), args)   # test

**Comment by emcastillo** [comment]
PTAL, removed `options` and `backend` args from compile.
If we want to change it from the cmdline, just changing the `kernel.options = '...'` attr before calling compile should suffice.

File: cupy/core/raw.pyx [hunk 13]
```diff
@@ -83,6 +80,9 @@ cdef class RawKernel:
 
     @property
     def kernel(self):
+        return self._kernel()
+
+    def _kernel(self, log_stream=None):
```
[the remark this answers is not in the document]
emcastillo: when adding the def to the pxd, I have to import Function and Module and then Cython crashes ...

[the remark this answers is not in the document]
leofang: Ouch...Looks like we hit another cyclic import...Alright sorry for bad suggestion 😢

cupy/core/raw.pyx
```diff
@@ -223,6 +223,21 @@ cdef class RawKernel:
         attr = driver.CU_FUNC_ATTRIBUTE_PREFERRED_SHARED_MEMORY_CARVEOUT
         driver.funcSetAttribute(self.kernel.ptr, attr, fraction)
 
+    def compile(self, log_stream=None):
+        """Compile the current kernel.
+
+        Compiles the kernels in code. If the
+        arguments are not given, the corresponding attributes of the RawKernel
+        instance that were given during initialization are used.
```
leofang, line 231: I think the docstring is outdated. How about just removing them?
```suggestion
```

cupy/core/raw.pyx
```diff
@@ -223,6 +223,21 @@ cdef class RawKernel:
...
+
+        Args:
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written. Defaults to `None`.
```
leofang, line 235:
```suggestion
                which the compiler output will be written. Defaults to ``None``.
```

cupy/core/raw.pyx
```diff
@@ -316,16 +328,40 @@ cdef class RawModule:
...
+
+    def _module(self, log_stream=None):
         # The module is cached, so on the device where this has been called,
         # we would just look up from the cache, and do recompiling only when
         # switching to a different device
         cdef Module mod
+
+        # Change the options to reflect
+        # the actual options the module was compiled with.
+
```
leofang, line 341: ditto
```suggestion
```

cupy/core/raw.pyx
```diff
@@ -316,16 +328,40 @@ cdef class RawModule:
...
-            self.name_expressions, self.log_stream)
+            self.name_expressions, log_stream)
         return mod
 
+    def compile(self, log_stream=None):
+        """Compile the current module.
+
+        Compiles the kernels in code. If the
+        arguments are not given, the corresponding attributes of the RawKernel
+        instance that were given during initialization are used.
```
leofang, line 353: ditto
```suggestion
```

cupy/core/raw.pyx
```diff
@@ -316,16 +328,40 @@ cdef class RawModule:
...
+
+        Args:
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written. Defaults to `None`.
```
leofang, line 357:
```suggestion
                which the compiler output will be written. Defaults to ``None``.
```

cupy/core/raw.pyx
```diff
@@ -316,16 +328,40 @@ cdef class RawModule:
...
+
+        .. note::
+            Calling `compile` will reset the internal state of a ``RawKernel``.
```
leofang, line 360:
```suggestion
            Calling :meth:`compile` will reset the internal state of a :class:`RawKernel`.
```

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -375,6 +384,40 @@ def _helper(self, kernel, dtype):
...
+            _test_compile_src, 'test_op',
+            options=('-DOP=+',),
+            backend=self.backend)
+        kern.compile()
+        x1, x2, y = self._helper(kern, cupy.float32)
+        assert cupy.allclose(y, x1 + x2)
+        kern = cupy.RawKernel(
+            _test_compile_src, 'test_op',
+            options=('-DOP=*',),
+            backend=self.backend)
```
leofang, line 398: I'd prefer to see this tested:
```suggestion
        kern.options = ('-DOP=*',)
```

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -375,6 +384,40 @@ def _helper(self, kernel, dtype):
...
+            backend=self.backend,
+            options=('-DOP=+',))
+        module.compile()
+        kern = module.get_function('test_op')
+        x1, x2, y = self._helper(kern, cupy.float32)
+        assert cupy.allclose(y, x1 + x2)
+        module = cupy.RawModule(
+            code=_test_compile_src,
+            backend=self.backend,
+            options=('-DOP=*',))
```
leofang, line 415: ditto
```suggestion
        module.options = ('-DOP=*',)
```

emcastillo:
Now should be ok! sorry 😅

leofang:
LGTM! 

Jenkins, test this please

cupy/core/raw.pyx
```diff
@@ -223,6 +223,18 @@ cdef class RawKernel:
         attr = driver.CU_FUNC_ATTRIBUTE_PREFERRED_SHARED_MEMORY_CARVEOUT
         driver.funcSetAttribute(self.kernel.ptr, attr, fraction)
 
+    def compile(self, log_stream=None):
+        """Compile the current kernel.
```
kmaehashi, line 227: Could you add docs like this?
```
In general, you don't have to call this method; kernels are compiled implicitly on the first call.
```

kmaehashi (changes requested):

cupy/core/raw.pyx
```diff
@@ -316,16 +325,35 @@ cdef class RawModule:
...
+        """Compile the current module.
+
+        Args:
+            log_stream (object): Pass either ``sys.stdout`` or a file object to
+                which the compiler output will be written.
+                Defaults to ``None``.
+
+        .. note::
+            Calling :meth:`compile` will reset the internal state of
+            a :class:``RawKernel``.
```
leofang, line 352: This rendered incorrectly on RTD. It should be single ticks, not double.
```suggestion
            a :class:`RawKernel`.
```

kmaehashi:
pfnCI, test this please.

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -375,6 +384,40 @@ def _helper(self, kernel, dtype):
...
+        module.compile()
```
kmaehashi, line 416: Could you add test cases to call `compile()` with log stream specified?

emcastillo:
PTAL

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -922,6 +933,42 @@ def test_context_switch_RawModule6(self):
...
+
+    def test_compile_kernel(self):
+        kern = cupy.RawKernel(
+            _test_compile_src, 'test_op',
+            options=('-DOP=+',),
+            backend='nvcc')
+        log = io.StringIO()
+        with use_temporary_cache_dir():
+            kern.compile(log_stream=log)
+        assert 'warning' in log.getvalue()
```
leofang, line 954: "warning" for what?

leofang:
Jenkins, test this please

tests/cupy_tests/core_tests/test_raw.py
emcastillo: ununsed variable :p

PR #3645, opened 2020-07-22
Adding scipy.signal functions that are simple wrappers of ndimage functions

coderforlife:
I have implemented most of `scipy.signal.signaltools`. This is the first batch that should be the most straightforward since it is just utilizing the ndimage functions which includes:
 * `wiener`
 * `order_filter`
 * `med_filt`
 * `med_filt2d`
 * `sepfir2d` (which is actually in bsplines.py)

Next batch will be some of the utility functions and functions like `hilbert`. After that will be `convolve` and friends (but those are dependent on #3614). After that, there will be `lfilter` and `sosfilt` and friends. By then I may have the splines completed as well.

Currently there are no formal tests for these but those are coming shortly.

### DIFF
--- a/cupyx/__init__.py
+++ b/cupyx/__init__.py
@@ -1,4 +1,3 @@
-# "NOQA" to suppress flake8 warning
 from cupyx.rsqrt import rsqrt  # NOQA
 from cupyx.runtime import get_runtime_info  # NOQA
 from cupyx.scatter import scatter_add  # NOQA
@@ -13,8 +12,6 @@ from cupyx import optimizing  # NOQA
 from cupyx._ufunc_config import errstate  # NOQA
 from cupyx._ufunc_config import geterr  # NOQA
 from cupyx._ufunc_config import seterr  # NOQA
-
-from cupyx.scipy import signal  # NOQA
 
 from cupy.core.syncdetect import allow_synchronize  # NOQA
 from cupy.core.syncdetect import DeviceSynchronized  # NOQA
--- a/cupyx/__init__.py
+++ b/cupyx/__init__.py
@@ -1,3 +1,4 @@
+# "NOQA" to suppress flake8 warning
 from cupyx.rsqrt import rsqrt  # NOQA
 from cupyx.runtime import get_runtime_info  # NOQA
 from cupyx.scatter import scatter_add  # NOQA
--- a/cupyx/scipy/signal/bsplines.py
+++ b/cupyx/scipy/signal/bsplines.py
@@ -19,6 +19,11 @@ def sepfir2d(input, hrow, hcol):
 
     .. seealso:: :func:`scipy.signal.sepfir2d`
     """
+    if input.dtype.kind != 'f':
+        input = input.astype(float)
+    hrow = hrow.astype(input.dtype, copy=False)
+    hcol = hcol.astype(input.dtype, copy=False)
     filters = (hcol[::-1], hrow[::-1])
+    origins = (0 if x.size % 2 else -1 for x in filters)
     return cupyx.scipy.ndimage.filters._run_1d_correlates(
-        input, (0, 1), lambda i: filters[i], None, 'reflect', 0.0)
+        input, (0, 1), lambda i: filters[i], None, 'reflect', 0.0, origins)
--- a/cupyx/scipy/signal/bsplines.py
+++ b/cupyx/scipy/signal/bsplines.py
@@ -1,3 +1,4 @@
+import cupy
 import cupyx.scipy.ndimage
 
 
@@ -19,11 +20,21 @@ def sepfir2d(input, hrow, hcol):
 
     .. seealso:: :func:`scipy.signal.sepfir2d`
     """
-    if input.dtype.kind != 'f':
-        input = input.astype(float)
-    hrow = hrow.astype(input.dtype, copy=False)
-    hcol = hcol.astype(input.dtype, copy=False)
+    dtype = input.dtype
+    if dtype.kind == 'c':
+        # TODO: adding support for complex types requires ndimage filters
+        # to support complex types (which they could easily if not for the
+        # scipy compatibility requirement of forbidding complex and using
+        # float64 intermediates)
+        raise TypeError("complex types not currently supported")
+    if dtype == cupy.float32 or dtype.itemsize <= 2:
+        dtype = cupy.float32
+    else:
+        dtype = cupy.float64
+    input = input.astype(dtype, copy=False)
+    hrow = hrow.astype(dtype, copy=False)
+    hcol = hcol.astype(dtype, copy=False)
     filters = (hcol[::-1], hrow[::-1])
-    origins = (0 if x.size % 2 else -1 for x in filters)
+    origins = [0 if x.size % 2 else -1 for x in filters]
     return cupyx.scipy.ndimage.filters._run_1d_correlates(
         input, (0, 1), lambda i: filters[i], None, 'reflect', 0.0, origins)
--- a/cupyx/scipy/signal/bsplines.py
+++ b/cupyx/scipy/signal/bsplines.py
@@ -1,3 +1,6 @@
+import cupyx.scipy.ndimage
+
+
 def sepfir2d(input, hrow, hcol):
     """Convolve with a 2-D separable FIR filter.
 
@@ -16,7 +19,6 @@ def sepfir2d(input, hrow, hcol):
 
     .. seealso:: :func:`scipy.signal.sepfir2d`
     """
-    from cupyx.scipy.ndimage.filters import _run_1d_correlates
     filters = (hcol[::-1], hrow[::-1])
-    return _run_1d_correlates(input, (0, 1), lambda i: filters[i],
-                              None, 'reflect', 0.0)
+    return cupyx.scipy.ndimage.filters._run_1d_correlates(
+        input, (0, 1), lambda i: filters[i], None, 'reflect', 0.0)
--- a/cupyx/scipy/signal/bsplines.py
+++ b/cupyx/scipy/signal/bsplines.py
@@ -1,0 +1,22 @@
+def sepfir2d(input, hrow, hcol):
+    """Convolve with a 2-D separable FIR filter.
+
+    Convolve the rank-2 input array with the separable filter defined by the
+    rank-1 arrays hrow, and hcol. Mirror symmetric boundary conditions are
+    assumed. This function can be used to find an image given its B-spline
+    representation.
+
+    Args:
+        input (cupy.ndarray): The input signal
+        hrow (cupy.ndarray): Row direction filter
+        hcol (cupy.ndarray): Column direction filter
+
+    Returns:
+        cupy.ndarray: The filtered signal
+
+    .. seealso:: :func:`scipy.signal.sepfir2d`
+    """
+    from cupyx.scipy.ndimage.filters import _run_1d_correlates
+    filters = (hcol[::-1], hrow[::-1])
+    return _run_1d_correlates(input, (0, 1), lambda i: filters[i],
+                              None, 'reflect', 0.0)
--- a/cupyx/scipy/signal/signaltools.py
+++ b/cupyx/scipy/signal/signaltools.py
@@ -1,3 +1,5 @@
+import cupy
+
 
 def choose_conv_method(in1, in2, mode='full'):
     """Find the fastest convolution/correlation method.
@@ -9,7 +11,7 @@ def choose_conv_method(in1, in2, mode='full'):
 
     Returns:
         str: A string indicating which convolution method is fastest,
-        either ``direct`` or ``fft1``.
+        either ``direct`` or ``fft``.
 
     .. warning::
         This function currently doesn't support measure option,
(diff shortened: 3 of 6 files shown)


leofang:
Tagging signal processing experts @mnicely @awthomp

awthomp:
Thanks for the ping, @leofang!

@coderforlife - please check out cuSignal. We already have support for many of these functions (lfilter on GPU is close to impossible, though!) and would love to have you contribute. 

https://github.com/rapidsai/cusignal

coderforlife:
I know those are there, but I thought this scipy was supposed to be near duplication of scipy and the what would be (or is) in a scikit would be in cuSignal, cuImage, or similar. Missing them here is a bit confusing.

Also, the functions in this commit are essentially freebies since they just call the ndimage functions after doing a small amount with the arguments (like checking).

I have been struggling with lfilter as well... best parallelization so far is with each signal/channel. But I have a few ideas, although they may not pan out.

awthomp:
> I know those are there, but I thought this scipy was supposed to be near duplication of scipy and the what would be (or is) in a scikit would be in cuSignal, cuImage, or similar. Missing them here is a bit confusing.
> 
> Also, the functions in this commit are essentially freebies since they just call the ndimage functions after doing a small amount with the arguments (like checking).
> 
> I have been struggling with lfilter as well... best parallelization so far is with each signal/channel. But I have a few ideas, although they may not pan out.

I'll default to whatever the core CuPy devs want. cuSignal heavily depends on this wonderful library, so even wrapping functions core to CuPy within cuSignal is totally fine with me.

As for the heavy-weight signal processing functions, like `lfilter` and `sosfilt`, in particular, I think it does make more sense for us to include these in cuSignal, but I'll admit I'm biased.

Here's our work with `lfilter`: https://github.com/rapidsai/cusignal/pull/78. We found there's a data dependency in the DF-II implementation, and our way around this was to have a single thread CUDA kernel and hope to get E2E acceleration. This didn't work out, and we deprecated our work here: https://github.com/rapidsai/cusignal/pull/93

Here's our `sosfilt` implementation: https://github.com/rapidsai/cusignal/pull/99

grlee77:
> Here's our work with lfilter: rapidsai/cusignal#78. We found there's a data dependency in the DF-II implementation, and our way around this was to have a single thread CUDA kernel and hope to get E2E acceleration. This didn't work out, and we deprecated our work here: rapidsai/cusignal#93

I implemented the specific IIR filters for the `spline_prefilter` and `spline_prefilter1d` functions from `scipy.ndimage.interpolation`. It will be slow in 1D, but I found that for `ndim >= 2`, it was possible to run 1 line of data per GPU thread and get reasonable acceleration. Here is a quick benchmark example.

```Python
import cupy as cp
import numpy as np
import scipy.ndimage as ndi

d = cp.cuda.Device()
a = np.random.randn(4096 * 4096).astype(np.float32)
a_gpu = cp.asarray(a)

%timeit ndi.spline_filter1d(a)
%timeit ndi.spline_filter1d(ag); d.synchronize()
```
Here the CPU time was: `186 ms ± 716 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)`
and the GPU time was: `6.9 ms ± 24.2 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)`

If I instead ravel the same arrays so that they are 1D, the CPU time becomes `237 ms` and GPU becomes MUCH longer at `3750 ms ` as it can no longer split up the work line-by-line on the GPU.

For reference: The [spline kernel code is here](https://github.com/mritools/cupyimg/blob/master/cupyimg/scipy/ndimage/_kernels/spline.py) and the configuration of RawKernel and block/grid is [here](https://github.com/mritools/cupyimg/blob/607553a8d30beb552fa74831ecae35d9d252bcf9/cupyimg/scipy/ndimage/interpolation.py#L243-L263).  I think the kernel is basically the same type of code as `lfilter`, just with hard-coded `a, b` corresponding to various spline orders and some specific boundary handling cases that aren't present in `lfilter`.

coderforlife:
@grlee77 @awthomp There are ways to make even IIR filters across single signals parallel. I am trying them out now to see if there is an improvement in speed (they require more computations, but in turn allow parallelization).

coderforlife:
@grlee77 My currently version only parallelizes across the signals like you suggest.

@awthomp I looked through your code and I was starting to approach something similar with a few differences. It seems as though the `x*b` can be done completely parallel for each sample (assuming `a_len` is less than `x_len`). That's over half of the work for each sample done in parallel. Those results are stored in `out`. Then scan through and subtract off the `a*y`. You are doing more work overall, but it seems as though it might make up for it. I am going to try a version like that right now.

coderforlife:
@awthomp I have had some success with this! By having the `x*b` portion completely parallelized and then doing `y*a` in parallel over the signals/channels I see a 50% to 85% speed improvement over my old method that only parallelized over signals.

I have now created 3 versions:

1. My original version which operated in parallel across signals but not within the samples for a single signal, may not have been well-optimized in other ways.
2. My second version where I made an `ElementwiseKernel` to do the `x*b` step completely in parallel and then a `RawKernel` to do the `y*a` in parallel across signals. I never completed this version so it doesn't support `zi`/`zf`. It also has a bug right now and doesn't generate the correct output.
3. While making v2 I realized that the `ElementwiseKernel` was really just doing a convolution and I should reuse my already-made `convolve` function! The third version uses `convolve` for the `x*b` step and then the same `RawKernel` from v2 (with some minor modifications to support `zi`/`zf`).

I tested with 2^20 elements and started at having an equal number of signals and samples (both 1024) and then reduced signals while increasing the samples until getting to only 1 signal with 2^20 samples. Here is the timing data (in ms, from a Titan V GPU) with percent of time of v1.

  | Original Version | Version 2 |   | Version 3 |  
-- | -- | -- | -- | -- | --
1024x1024 | 1.68 | 0.986 | 59% | 0.699 | 42%
512x2048 | 3.13 | 1.79 | 57% | 1.2 | 38%
256x4096 | 6.11 | 3.4 | 56% | 2.28 | 37%
128x8192 | 12 | 6.62 | 55% | 4.24 | 35%
64x16384 | 21.1 | 6.83 | 32% | 5.72 | 27%
32x32768 | 40.3 | 9.73 | 24% | 10.4 | 26%
16x65536 | 72.6 | 13.8 | 19% | 16.4 | 23%
8x131072 | 142 | 23.3 | 16% | 31.8 | 22%
4x262144 | 276 | 43.9 | 16% | 61.4 | 22%
2x524288 | 542 | 81.9 | 15% | 101 | 19%
1x1048576 | 352 | 156 | 44% | 156 | 44%

Overall, v2 and v3 are between 15% and 44% of the original's time. There is a strange behavior in the v2 is faster near the bottom and v3 is faster near the top. I am really curious why this is (it also has a bug so maybe those timings aren't good anyways). Also, everyone sucks at 1x however v1 actually is faster at 1x than 2x.

BTW, all of these version support ND data in the same way the scipy does (one axis is considered the samples, all other axes are signals).

Finally, I tried against your implementation (only the last row up there since yours only supports 1D), and yours takes 738 ms vs 352 (v1) or 156 (v2 and v3). I have no idea why yours is so much slower. But, for me, I haven't tried adding `__restrict__` to my pointers yet... maybe I will get some improvement!

coderforlife:
@awthomp Another benefit of the new approach is that `b` and `a` can be different lengths without needing to be 0-padded since each is used completely separately.

awthomp:
@coderforlife so happy to see you making progress. I’ll be in and out over the next two weeks for vacation. These optimizations were all due to the great work from @mnicely, and I’m sure he’ll be able to help.

coderforlife:
I think we should move this discussion out of this PR since it has nothing to do with these first few filters.

For now, just need approval of current 5 filters which all just utilize the `ndimage.filters` functions to do nearly all of the work.

coderforlife:
Seems as though the signal processing experts are okay with this. It really is just mapping some `signal` functions to their equivalents in `ndimage` (`wiener` is a bit more complex, but only slightly).

cupyx/__init__.py
```diff
@@ -1,4 +1,3 @@
-# "NOQA" to suppress flake8 warning
```
takagi, line 1: Would you leave this comment if there is no special reason here?

cupyx/scipy/signal/bsplines.py
```diff
@@ -0,0 +1,29 @@
...
+        input (cupy.ndarray): The input signal
+        hrow (cupy.ndarray): Row direction filter
+        hcol (cupy.ndarray): Column direction filter
+
+    Returns:
+        cupy.ndarray: The filtered signal
+
+    .. seealso:: :func:`scipy.signal.sepfir2d`
+    """
+    if input.dtype.kind != 'f':
```
takagi, line 22: I think it is okay to leave complex dtypes not supported in this PR.  Would you raise a `TypeError` in case a complex ndarray is supplied?

cupyx/scipy/signal/bsplines.py
```diff
@@ -0,0 +1,29 @@
...
+        input = input.astype(float)
+    hrow = hrow.astype(input.dtype, copy=False)
+    hcol = hcol.astype(input.dtype, copy=False)
+    filters = (hcol[::-1], hrow[::-1])
+    origins = (0 if x.size % 2 else -1 for x in filters)
```
takagi, line 27: I think a list is enough compared with a generator here as `filters` has only two elements. How about this?
```suggestion
    origins = [0 if x.size % 2 else -1 for x in filters]
```

tests/cupyx_tests/scipy_tests/signal_tests/test_bsplines.py
```diff
@@ -0,0 +1,27 @@
...
+    import scipy.signal  # NOQA
+except ImportError:
+    pass
+
+
+@testing.parameterize(*testing.product({
+    'input': [(256, 256), (4, 512), (512, 3)],
+    'hrow': [1, 3, 4],
+    'hcol': [1, 3, 4],
+    'dtype': ['int32', 'float32', 'float64'],
```
takagi, line 17: nit: Would you use data types instead of string literals here?
```suggestion
    'dtype': [cupy.int32, cupy.float32, cupy.float64],
```

tests/cupyx_tests/scipy_tests/signal_tests/test_bsplines.py
```diff
@@ -0,0 +1,27 @@
+import unittest
+
+from cupy import testing
+
+import cupyx.scipy.signal  # NOQA
+
+try:
+    import scipy.signal  # NOQA
+except ImportError:
+    pass
```
takagi, line 10: Would you remove these unused lines in this PR?

cupyx/scipy/signal/signaltools.py
```diff
@@ -38,3 +45,146 @@ def _fftconv_faster(x, h, mode):
...
+    Returns:
+        cupy.ndarray: An array the same size as input containing the median
+            filtered result.
+    See also
+    --------
+    .. seealso:: :func:`cupyx.scipy.ndimage.median_filter`
+    .. seealso:: :func:`cupyx.scipy.signal.medfilt`
+    .. seealso:: :func:`scipy.signal.medfilt2d`
+    """
+    # Scipy's version only supports uint8, float32, and float64
```
takagi, line 175: Would you raise an error if a ndarray of other dtypes is supplied as SciPy does?

tests/cupyx_tests/scipy_tests/signal_tests/test_bsplines.py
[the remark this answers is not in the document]
coderforlife: Removing these will cause *all* tests to fail because they won't be able to use `cupyx.scipy.signal` or `scipy.signal` (used as `scp.signal` later with `scp` either `cupyx.scipy` or `scipy`).

tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
```diff
@@ -3,9 +3,15 @@
...
 import cupy
-import cupyx
 from cupy import testing
 
+import cupyx.scipy.signal  # NOQA
+
+try:
+    import scipy.signal  # NOQA
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
coderforlife: Removing these will cause *all* tests to fail because they won't be able to use `cupyx.scipy.signal` or `scipy.signal` (used as `scp.signal` later with `scp` either `cupyx.scipy` or `scipy`).

coderforlife:
No problem, I understand it must be busy with the recent releases. I fixed all of the issues except the ones where if I remove those lines everything fails. Also, I am going to try right now to add support for complex dtypes (support actually already exists, it just isn't directly accessible).

coderforlife:
So I was able to fairly easily add complex support to both `signal.bsplines.sepfir2` and `signal.signaltools.wiener` however it required duplicating code from `ndimage.filters` and `ndimage._filters_core`. I am unsure if it wouldn't be better to simply add an extra bool argument to several `ndimage.filters` functions that simply allows turning off the requirement of intermediate results must be `float64` (that is to reproduce scipy's behavior and prevents complex numbers from working). I think that discussion can happen elsewhere though and for now we can just avoid complex in this PR.

tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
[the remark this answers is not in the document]
takagi: I see. `@testing.numpy_cupy_allclose` imports just `scipy` and `cupyx.scipy` packages only, and `scipy.signal` and `cupyx.scipy.signal` can't be seen.

takagi:
Thanks for updating the changes, and it is fine to leave complex unsupported here.

takagi:
pfnCI, test this please.

takagi:
LGTM!

leofang:
@coderforlife We're seeing irrelevant PRs (#3730, #3735) failing at `tests/cupyx_tests/scipy_tests/signal_tests/test_bsplines.py`, but only on one of the CIs. Do you happen to know if it's due to the different SciPy version that it used?

coderforlife:
It turns out that there is a bug in scipy. Way down in the source code for `sepfir2d` I found this comment: `/* h must be odd length */`. However, scipy never enforces this or has this in the public documentation. You will notice that all of the failing ones have even `hcol` and/or `hrow`.

For the moment I have submitted the issue at scipy: https://github.com/scipy/scipy/issues/12691. I think in the mean time we should add a check in `cupyx.scipy.signal.bsplines` for even-length `hrow` and `hcol` that raises an exception and remove even-length tests.

leofang:
Great, thanks for pinpointing the issue so quickly, @coderforlife! I am wondering if the errors could happen randomly? It's not always reproducible on the failing CI (for example, this PR went in just fine, and a few other PRs too).

coderforlife:
Well scipy's `sepfir2d` is reading out-of-bounds memory when the length of `hrow` or `hcol` is even so yes, it has random effects since whatever the memory contents is that it ends up reading.

leofang:
Yeah can't reproduced it locally...

coderforlife:
Different memory allocator or something.

coderforlife:
See https://github.com/cupy/cupy/pull/3750.

PR #3651, opened 2020-07-23
MAINT: Use builtins directly

larsoner:
Avoid warnings with latest `numpy` of the form:
```
E   DeprecationWarning: `np.complex` is a deprecated alias for the builtin `complex`. Use `complex` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.complex_` here.
```

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -226,13 +226,16 @@ from numpy import complex128  # NOQA
 # Built-in Python types
 # -----------------------------------------------------------------------------
 
-int = __builtins__['int']  # NOQA
+# After NumPy 1.20 is released, CuPy should mimic the DeprecationWarning
+# behavior for these types
 
-bool = __builtins__['bool']  # NOQA
+from builtins import int  # NOQA
 
-float = __builtins__['float']  # NOQA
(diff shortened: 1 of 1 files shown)


leofang:
Thanks, Eric. Maybe the core devs have different opinions, so you should wait for further reply. I am just curious if we could instead do a `catch_warnings()` here, as the idea is to alias "whatever NumPy uses". Not saying these are likely to be changed again on the NumPy side, but if they do, it would still be bulletproof. Just my two cents...

kmaehashi:
Thanks for the fix! As for now, I think the proposed change is reasonable. (Although I prefer `from builtins import int` style.)

After NumPy 1.20 release, CuPy should mimic DeprecationWarning behavior. (I think that will be after CuPy v8 release)
https://github.com/cupy/cupy/wiki/Development-policies#numpyscipy-compatibility

larsoner:
> Although I prefer from builtins import int style

Changed to use this and added a comment that the types should be deprecated after NumPy releases 1.20

kmaehashi:
LGTM!
pfnCI, test this please.

PR #3653, opened 2020-07-23
improve efficiency of cupy.fft.fftfreq and rfftfreq

grlee77:
This is a minor change to reduce the number of elementwise division kernels called in `fftfreq` and `rfftfreq`.

For `fftfreq` with n=16,  it reduces execution time from ~75 us to ~60 us.
For `fftfreq` with n = 10,000,000 it reduces execution time from ~1.4 ms to 1.0 ms.

### DIFF
--- a/cupy/fft/fft.py
+++ b/cupy/fft/fft.py
@@ -896,7 +896,7 @@ def fftfreq(n, d=1.0):
     .. seealso:: :func:`numpy.fft.fftfreq`
     """
     return cupy.hstack((cupy.arange(0, (n - 1) // 2 + 1, dtype=np.float64),
-                        cupy.arange(-(n // 2), 0, dtype=np.float64))) / n / d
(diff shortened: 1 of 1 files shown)


leofang:
Interesting...I wonder if doing only one multiplication (`* (1.0 / (n * d))` would further speed it up?

grlee77:
You can generate the frequencies with a float step size so that there are no elementwise multiplications or divisions, but that results in subtly differences in the frequencies due to limited floating point precision (e.g. ~1e-17 instead of 0 at the center frequency, etc. which I think is undesirable).

I don't think its worth spending a lot of time refining it as I doubt this function is a real bottleneck in anybody's code. I just happened to be looking at it when working on #3654, which calls this function once per array axis.

leofang:
Jenkins, test this please

asi1024:
LGTM.

PR #3654, opened 2020-07-23
add cupyx.scipy.ndimage.fourier_shift, fourier_gaussian, fourier_uniform

grlee77:
This adds the `fourier_shift` function from ndimage. These functions take a Fourier-domain image and apply the complex exponentials that would be needed to give a (circular) shift in the spatial domain. (i.e. based on the [Fourier shift theorem](https://www.dsprelated.com/freebooks/sasp/Shift_Theorem.html)).

It seems like we might already have a utility like `_reshape_nd` here somewhere else, but I'm not sure. If desired we could move it to `cupy.core.internal`.

`_get_output_fourier_complex` was copied from SciPy

The SciPy implementation is in C and could be emulated via an `ElementwiseKernel`. This separable implementation only has to take the complex exponential of a 1d array along each axis rather than at every n-d element, though, so I think it might be faster.

### DIFF
--- a/cupyx/scipy/ndimage/fourier.py
+++ b/cupyx/scipy/ndimage/fourier.py
@@ -47,13 +47,7 @@ def fourier_shift(input, shift, n=-1, axis=-1, output=None):
             If given, the result of shifting the input is placed in this array.
 
     Returns:
-        output (ndarray or None): The shifted output (in the Fourier domain).
-
-    Returns
-    -------
-    fourier_shift : ndarray
-        The shifted input.
-
+        output (ndarray): The shifted output (in the Fourier domain).
     """
     ndim = input.ndim
     output = _get_output_fourier_complex(output, input)
--- a/cupyx/scipy/ndimage/fourier.py
+++ b/cupyx/scipy/ndimage/fourier.py
@@ -1,5 +1,7 @@
-import cupy
 import numpy as np
+
+import cupy
+from cupyx.scipy.ndimage import filters
 
 
 def _get_output_fourier_complex(output, input):
@@ -19,13 +21,6 @@ def _get_output_fourier_complex(output, input):
 
 def _reshape_nd(arr, ndim, axis):
     """Promote a 1d array to ndim with non-singleton size along axis."""
-    if arr.ndim != 1:
-        raise ValueError("expected a 1d array")
-    if axis < -ndim or axis > ndim - 1:
-        raise ValueError("invalid axis")
-    if ndim < 1:
-        raise ValueError("ndim must be >= 1")
-    axis = axis % ndim
     nd_shape = (1,) * axis + (arr.size,) + (1,) * (ndim - axis - 1)
     return arr.reshape(nd_shape)
 
@@ -60,25 +55,16 @@ def fourier_shift(input, shift, n=-1, axis=-1, output=None):
         The shifted input.
 
     """
+    ndim = input.ndim
     output = _get_output_fourier_complex(output, input)
+    axis = cupy.util._normalize_axis_index(axis, ndim)
+    shifts = filters._fix_sequence_arg(shift, ndim, 'shift')
+
     output[...] = input
-
-    ndim = output.ndim
-    if axis < -ndim or axis >= ndim:
-        raise ValueError("invalid axis")
-    axis = axis % ndim
-
-    if np.isscalar(shift):
-        shift = (shift,) * ndim
-    elif len(shift) != ndim:
-        raise ValueError("number of shifts must match input.ndim")
-
-    for kk in range(ndim):
-        ax_size = output.shape[kk]
-        shiftk = shift[kk]
+    for ax, (shiftk, ax_size) in enumerate(zip(shifts, output.shape)):
         if shiftk == 0:
             continue
-        if kk == axis and n > 0:
+        if ax == axis and n > 0:
             # cp.fft.rfftfreq(ax_size) * (-2j * np.pi * shiftk *  ax_size / n)
             arr = cupy.arange(ax_size, dtype=output.dtype)
             arr *= -2j * np.pi * shiftk / n
@@ -88,7 +74,7 @@ def fourier_shift(input, shift, n=-1, axis=-1, output=None):
         cupy.exp(arr, out=arr)
 
         # reshape for broadcasting
-        arr = _reshape_nd(arr, ndim=ndim, axis=kk)
+        arr = _reshape_nd(arr, ndim=ndim, axis=ax)
         output *= arr
 
     return output
--- a/cupyx/scipy/ndimage/fourier.py
+++ b/cupyx/scipy/ndimage/fourier.py
@@ -4,6 +4,23 @@ import cupy
 from cupyx.scipy.ndimage import filters
 
 
+def _get_output_fourier(output, input):
+    if output is None:
+        if input.dtype.type in [cupy.complex64, cupy.complex128,
+                                cupy.float32]:
+            output = cupy.zeros(input.shape, dtype=input.dtype)
+        else:
+            output = cupy.zeros(input.shape, dtype=cupy.float64)
+    elif type(output) is type:
+        if output not in [cupy.complex64, cupy.complex128,
+                          cupy.float32, cupy.float64]:
+            raise RuntimeError("output type not supported")
+        output = cupy.zeros(input.shape, dtype=output)
+    elif output.shape != input.shape:
+        raise RuntimeError("output shape not correct")
+    return output
+
+
 def _get_output_fourier_complex(output, input):
     if output is None:
         if input.dtype.type in [cupy.complex64, cupy.complex128]:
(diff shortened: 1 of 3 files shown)


grlee77:
Benchmarking indicates that it is up to ~200 times faster than SciPy for sufficiently large inputs.


```Python
import numpy as np
import cupy as cp
import scipy.ndimage as ndi

print(" case | timing | acceleration")
print("------|--------|-------------")
for dtype in [np.complex64, np.complex128]:
    for shape in [(16, 16), (512, 512), (4096, 4096), (256, 256, 256), (64, 64, 64, 64)]:

        a = cp.random.randn(*shape).astype(dtype)
        a = a + 1j * cp.random.randn(*shape).astype(dtype)
        a_cpu = a.get()

        perf_cpu = repeat(
            ndi.fourier_shift, (a_cpu, 5), n_warmup=1, n_repeat=10
        )
        # print(f"shape={shape, np.dtype(dtype).name},  CPU | {perf_cpu}")

        perf_gpu = repeat(fourier_shift, (a, 5), n_warmup=20, n_repeat=100)
        accel = perf_cpu.cpu_times.mean() / perf_gpu.gpu_times.mean()
        print(f"shape={shape, np.dtype(dtype).name},  GPU | {perf_gpu} | {accel}")
```
 case | timing | acceleration
------|--------|-------------
shape=((16, 16), 'complex64'),  GPU | fourier_shift       :    CPU:  259.434 us   +/- 9.128 (min:  248.098 / max:  289.571) us     GPU-0:  265.398 us   +/- 9.740 (min:  253.056 / max:  295.584) us | 0.119
shape=((512, 512), 'complex64'),  GPU | fourier_shift       :    CPU:  242.909 us   +/-10.986 (min:  232.220 / max:  289.472) us     GPU-0:  260.337 us   +/-10.977 (min:  246.784 / max:  307.136) us | 33.431
shape=((4096, 4096), 'complex64'),  GPU | fourier_shift       :    CPU:  247.149 us   +/- 8.721 (min:  238.362 / max:  304.858) us     GPU-0: 2707.958 us   +/- 2.065 (min: 2704.256 / max: 2720.768) us | 207.857
shape=((256, 256, 256), 'complex64'),  GPU | fourier_shift       :    CPU:  361.868 us   +/-14.145 (min:  348.473 / max:  414.055) us     GPU-0: 4156.544 us   +/- 2.868 (min: 4153.280 / max: 4177.472) us | 153.500
shape=((64, 64, 64, 64), 'complex64'),  GPU | fourier_shift       :    CPU:  462.723 us   +/-20.095 (min:  448.648 / max:  588.778) us     GPU-0: 6233.556 us   +/-190.690 (min: 6143.744 / max: 6674.432) us | 130.562
shape=((16, 16), 'complex128'),  GPU | fourier_shift       :    CPU:  231.031 us   +/- 5.431 (min:  225.306 / max:  249.774) us     GPU-0:  235.647 us   +/- 5.492 (min:  229.792 / max:  254.912) us | 0.118
shape=((512, 512), 'complex128'),  GPU | fourier_shift       :    CPU:  256.580 us   +/-26.733 (min:  244.917 / max:  434.671) us     GPU-0:  277.703 us   +/-26.485 (min:  266.080 / max:  455.680) us | 30.315
shape=((4096, 4096), 'complex128'),  GPU | fourier_shift       :    CPU:  249.217 us   +/- 9.349 (min:  240.328 / max:  284.779) us     GPU-0: 4819.422 us   +/- 5.673 (min: 4810.752 / max: 4831.200) us | 119.193
shape=((256, 256, 256), 'complex128'),  GPU | fourier_shift       :    CPU:  367.907 us   +/-16.750 (min:  347.834 / max:  480.342) us     GPU-0: 6462.972 us   +/- 6.579 (min: 6453.248 / max: 6496.032) us | 100.121
shape=((64, 64, 64, 64), 'complex128'),  GPU | fourier_shift       :    CPU:  464.110 us   +/-21.806 (min:  447.258 / max:  567.377) us     GPU-0: 9009.900 us   +/- 2.527 (min: 9006.080 / max: 9023.488) us | 93.633

grlee77:
cc @coderforlife for awareness: I had already done this function for `cupyimg` as it is used within scikit-image functions, but I have not worked on anything else from the `ndimage.fourier` submodule.

coderforlife:
@grlee77 I have never actually used any of the `ndimage.fourier` functions. I have looked at them, and many of them look to mostly jus tpre-compute some values (like exponentials for Guassians) and then do an element-wise multiplication. I have always just done the multiplication myself.

grlee77:
I went ahead and added `fourier_gaussian` and `fourier_uniform` as well. They are also separable and thus can be implemented very similarly to the `fourier_shift` case.

The test structure is nearly the same too. Mainly it is just the function name and the second argument name/range that is different across these three functions.

coderforlife:
These all seem way simpler than the scipy implementations (which I always thought was overkill).

cupyx/scipy/ndimage/fourier.py
```diff
@@ -0,0 +1,194 @@
...
+            output = cupy.zeros(input.shape, dtype=input.dtype)
+        else:
+            output = cupy.zeros(input.shape, dtype=cupy.complex128)
+    elif type(output) is type:
+        if output not in [cupy.complex64, cupy.complex128]:
+            raise RuntimeError("output type not supported")
+        output = cupy.zeros(input.shape, dtype=output)
+    elif output.shape != input.shape:
+        raise RuntimeError("output shape not correct")
+    return output
```
leofang, line 36: I think you can combine the two functions?
```python
def _get_output_fourier(output, input, complex_only=False):
      types = [cupy.complex64, cupy.complex128]
      if not complex_only:
          types += [cupy.float32, cupy.float64]
     
      if output is None:
          if input.dtype in types:
              output = cupy.zeros(input.shape, dtype=input.dtype)
          else:
              output = cupy.zeros(input.shape, dtype=types[-1])
      elif type(output) is type:
          if output not in types:
              raise RuntimeError("output type not supported")
          output = cupy.zeros(input.shape, dtype=output)
      elif output.shape != input.shape:
          raise RuntimeError("output shape not correct")
      return output
```

cupyx/scipy/ndimage/fourier.py
```diff
@@ -0,0 +1,194 @@
+import numpy as np
```
leofang, line 1: `import numpy`

cupyx/scipy/ndimage/fourier.py
grlee77: Yeah, I can refactor to combine some things.

grlee77:
> These all seem way simpler than the scipy implementations (which I always thought was overkill).

Yeah, actually these implementations when using NumPy instead of CuPy are faster than SciPy's in most cases. SciPy's C implementation wins in 1D, but for 2D/3D this separable approach can be 3-5 times faster.

I think the SciPy approach is more justified for the elliptical filter case (not implemented in this PR). It is the only case that is non-separable along the axes (you have to compute the Euclidean distance from the f=0 coordinate at each frequency location).

coderforlife:
FTI, `_fix_sequence_arg` is now in `cupyx.scipy.ndimage._util` and not `cupyx.scipy.ndimage.filters`.

takagi:
pfnCI, test this please.

cupyx/scipy/ndimage/fourier.py
```diff
@@ -0,0 +1,194 @@
...
+    for ax, (shiftk, ax_size) in enumerate(zip(shifts, output.shape)):
+        if shiftk == 0:
+            continue
+        if ax == axis and n > 0:
+            # cp.fft.rfftfreq(ax_size) * (-2j * np.pi * shiftk *  ax_size / n)
+            arr = cupy.arange(ax_size, dtype=output.dtype)
+            arr *= -2j * np.pi * shiftk / n
+        else:
+            arr = cupy.fft.fftfreq(ax_size)
+            arr = arr * (-2j * np.pi * shiftk)
```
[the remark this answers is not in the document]
grlee77: In-place won't work here because `arr` is real-valued prior to the multiplication, but complex afterwards.

We can make it slightly more efficient if we duplicate some internals of `fftfreq` rather than calling `fftfreq`, but I thought that was probably not worth the added complexity. I did do that for the real-valued case above where `rfftfreq` basically just corresponds to `arange`, so wasn't complicated to replace.

cupyx/scipy/ndimage/fourier.py
grlee77: Thanks. I have now made the suggested change

grlee77:
This should now be ready to test

takagi:
pfnCI, test this please.

takagi:
Thanks! Merging will be done after the CI passes.

grlee77:
I see the reason for the failure. The tests as written require scipy >= 1.4 to run due to use of the newer `scipy.fft` module. We should be able to fall back to `scipy.fftpack` instead for older SciPy.

grlee77:
Hopefully the most recent commit will change the previous CI failure. I didn't see a really easy way to run the real-valued tests on older scipy because the behavior of the real-valued FFTs in scipy.fft behave like numpy.fft and not scipy.fftpack. We also can't use numpy.fft instead of scipy.fft because it always returns double precision results while scipy's does not.

takagi:
pfnCI, test this please.

leofang:
pfnCI, test this please.

PR #3655, opened 2020-07-23
support order argument in cupy.ones, cupy.full and cupy.eye

### LINKED ISSUE #3632: Cupy.full does not have order parameter
Hello, 

I noticed that Cupy.full does not have an order parameter, this both exists in Numpy.full and all the other Cupy.full like functions like full_like, ones_like and zeros_like. Is there any reason for that? I'm trying to use CuPy in a project as a drop in replacement and this errors out. 

Thank you!

grlee77:
closes #3632

`empty` and `zeros` already supported this, but for some reason these two functions did not

### DIFF
--- a/cupy/creation/basic.py
+++ b/cupy/creation/basic.py
@@ -130,7 +130,7 @@ def identity(n, dtype=float):
     return eye(n, dtype=dtype)
 
 
-def ones(shape, dtype=float):
+def ones(shape, dtype=float, order='C'):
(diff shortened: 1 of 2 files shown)


grlee77:
Yes, I will add eye as well. Good catch!

leofang:
Jenkins, test this please

leofang:
@takagi I think this is ready to go.

takagi:
LGTM!

PR #3656, opened 2020-07-24
Parameterize sparse min/max/argmin/argmax tests

takagi:
Merge #3639 first.

This PR parameterizes the tests of sparse min/max/argmin/argmax. As argmin/argmax currently do not work with `axis=None`, I make the corresponding tests skip here and will post another PR to fix it.


leofang:
How about merging with master?

toslunar:
Merged `master`.

Jenkins, test this please.

PR #3664, opened 2020-07-27
[MNT] Expose accelerator internal API to one level up

leofang:
We already exposed the set methods to `cupy.core`. Might as well expose the get methods.


leofang:
Jenkins, test this please

leofang:
Thanks Emilio!

PR #3665, opened 2020-07-27
List all documented APIs to `__all__`

### LINKED ISSUE #3420: Make undocumented APIs unvisible to users

takagi:
Close #3420.

(The description will be described later.)

I need to have the following two steps:
- Define `__all__` variable for each modules
- Import `*` from its submodules for each package

I will use importing `*` from submodules instead of defining `__all__` for packages; otherwise, we have to repeatedly define a name in a module on packages that contain the module directly or indirectly. That means we can not partially expose submodules of a package, i.g. `cupy.core.internal` which is imported into `cupy.core` package but not into `cupy` package.

Is it enough to define `__all__` in the topmost packages only? If so, we can get the same thing with the current `cupy/__init__.py, modifying it to hide unnecessary names as standard libraries.

### DIFF
--- a/cupy/_sorting/__init__.py
+++ b/cupy/_sorting/__init__.py
@@ -5,3 +5,7 @@
 from cupy._sorting import count  # NOQA
 from cupy._sorting import search  # NOQA
 from cupy._sorting import sort  # NOQA
+
+from cupy._sorting.count import *  # NOQA
+from cupy._sorting.search import *  # NOQA
+from cupy._sorting.sort import *  # NOQA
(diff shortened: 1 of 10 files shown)


takagi:
Note about `__all__` variable:
- `__all__` determines public names defined in a module when it is imported by `from ... import *` statement.
- It is optional to define `__all__`. If it is not defined, the set of public names of a module includes all names found in the module's namespace which do not begin with an underscore character.

takagi:
Close this PR as discussed on #3420.

PR #3676, opened 2020-07-28
Remove `sum_duplicate` parameter in sparse min/max/argmin/argmax

### LINKED ISSUE #3535: Difference from SciPy about sparse min/max `sum_duplicates` parameter

takagi:
Close #3535. Extracted from #3583. Merge #3656 first.

Currently, sparse min/max/argmin/argmax have `sum_duplicate` parameter that, when `sum_duplicate=False` is supplied, indicates to skip combining duplicate elements prior to the operation. However, the implementation of `sum_duplicate` method immediately returns if a sparse matrix is already canonically formatted so we don't need to specify `sum_duplicate=False` explicitly. This PR removes the parameter.

### DIFF
--- a/cupyx/scipy/sparse/data.py
+++ b/cupyx/scipy/sparse/data.py
@@ -137,15 +137,14 @@ class _minmax_mixin(object):
 
     """
 
-    def _min_or_max_axis(self, axis, min_or_max, sum_duplicates, nonzero):
+    def _min_or_max_axis(self, axis, min_or_max, nonzero):
         N = self.shape[axis]
         if N == 0:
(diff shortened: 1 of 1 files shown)


kmaehashi:
This is no-compat (but not to be in upgrade guide since this changes the interface introduced during v8 beta)

toslunar:
Jenkins, test this please.

PR #3679, opened 2020-07-28
Work around a known CUB SpMV bug

leofang:
**UPDATE 2**: This bug should have very minimal to zero impact to us, as our default is to allocate memory more than needed, see the discussion in the replies below. But, for the sake of safety and correctness, we still opt in this workaround. 

---

I was browsing the CUB issue tracker and noticed an out-of-bound access bug was reported in NVlabs/cub#139 and NVlabs/cub#190. However, strangely it does ***not*** affect us:
- The reported test cases works correctly in CuPy;
- `cuda-memcheck pytest tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py -k TestCubSpmvBug` pass on my system
- **UPDATE**: this bug appears only if CuPy's memory pool is disabled, see the discussion starting https://github.com/cupy/cupy/pull/3679#issuecomment-665141174

I added the two reproducers to see if the CI can reproduce it. @grlee77 Any chance you know why we're safe?


leofang:
Jenkins, test this please

leofang:
@anaruse Maybe you also have some insight for this issue?

leofang:
I am guessing the vectors in the two reproducers are simply not long enough to hit the alignment boundary. (We always allocate more to ensure alignment.) Will try to probe this limit tomorrow...

leofang:
> I am guessing the vectors in the two reproducers are simply not long enough to hit the alignment boundary. (We always allocate more than needed to ensure alignment.) Will try to probe this limit tomorrow...

Bingo. 

```python
# test_cub_bug.py
import cupy
from cupy import testing
from cupyx.scipy import sparse as sp


xp = cupy
cupy.cuda.set_allocator(None)  # <-- key

N = 31

indptr = xp.arange(N+1, dtype=xp.int32)
indices = xp.arange(N, dtype=xp.int32)
data = xp.random.random(N, dtype=xp.float64)
m = sp.csr_matrix((data, indices, indptr), shape=(N, N))
x = xp.ones((N,), dtype=xp.float64)
assert m.indptr.data.mem.size == m.indptr.size * m.indptr.dtype.itemsize
assert x.data.mem.size == x.size * x.dtype.itemsize
result = m * x
print(result)
```
Run with `cuda-memcheck python test_cub_bug.py`, I reproduced the error (`Address 0x7f556b000080 is out of bounds`):
```
========= CUDA-MEMCHECK
========= Invalid __global__ read of size 4
=========     at 0x00000870 in void cub::DeviceSpmvKernel<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, cub::ReduceByKeyScanTileState<double, int, bool=1>, double, int, int2, bool=0, bool=0>(cub::SpmvParams<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, double>, int*, cub::KeyValuePair<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, cub::SpmvParams>*, int, int, int)
=========     by thread (31,0,0) in block (0,0,0)
=========     Address 0x7f556b000080 is out of bounds
=========     Device Frame:void cub::DeviceSpmvKernel<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, cub::ReduceByKeyScanTileState<double, int, bool=1>, double, int, int2, bool=0, bool=0>(cub::SpmvParams<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, double>, int*, cub::KeyValuePair<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, cub::SpmvParams>*, int, int, int) (void cub::DeviceSpmvKernel<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, cub::ReduceByKeyScanTileState<double, int, bool=1>, double, int, int2, bool=0, bool=0>(cub::SpmvParams<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, double>, int*, cub::KeyValuePair<cub::DispatchSpmv<double, int>::PtxSpmvPolicyT, cub::SpmvParams>*, int, int, int) : 0x870)
=========     Host Frame:/usr/local/cuda/lib64/libcuda.so.1 (cuLaunchKernel + 0x34e) [0x2d728e]
=========     Host Frame:/usr/local/cuda/lib64/libcudart.so.10.0 [0x1a23d]
=========     Host Frame:/usr/local/cuda/lib64/libcudart.so.10.0 [0x1a2c7]
=========     Host Frame:/usr/local/cuda/lib64/libcudart.so.10.0 (cudaLaunchKernel + 0x225) [0x4e3c5]
=========     Host Frame:/home/leofang/cupy/cupy/cuda/cub.cpython-37m-x86_64-linux-gnu.so [0x5a557]
=========     Host Frame:/home/leofang/cupy/cupy/cuda/cub.cpython-37m-x86_64-linux-gnu.so (_ZN3cub10DeviceSpmv5CsrMVIdEE9cudaErrorPvRmPT_PiS7_S6_S6_iiiP11CUstream_stb + 0x563) [0xeae43]
=========     Host Frame:/home/leofang/cupy/cupy/cuda/cub.cpython-37m-x86_64-linux-gnu.so [0x3e29e]
=========     Host Frame:python (_PyMethodDef_RawFastCallKeywords + 0x264) [0x168a04]
=========     Host Frame:python (_PyCFunction_FastCallKeywords + 0x21) [0x168b21]
=========     Host Frame:python (_PyEval_EvalFrameDefault + 0x501e) [0x1d4d5e]
=========     Host Frame:python (_PyFunction_FastCallDict + 0x10b) [0x11579b]
=========     Host Frame:python (_PyObject_FastCall_Prepend + 0x65) [0x135275]
=========     Host Frame:python [0x16623c]
=========     Host Frame:python [0x20afdf]
=========     Host Frame:python (PyNumber_Multiply + 0xb8) [0x132278]
=========     Host Frame:python (_PyEval_EvalFrameDefault + 0x17e5) [0x1d1525]
=========     Host Frame:python (_PyEval_EvalCodeWithName + 0x2f9) [0x114729]
=========     Host Frame:python (PyEval_EvalCodeEx + 0x44) [0x115654]
=========     Host Frame:python (PyEval_EvalCode + 0x1c) [0x11567c]
=========     Host Frame:python [0x22bcb4]
=========     Host Frame:python (PyRun_FileExFlags + 0xa1) [0x236191]
=========     Host Frame:python (PyRun_SimpleFileExFlags + 0x1c3) [0x236383]
=========     Host Frame:python [0x237475]
=========     Host Frame:python (_Py_UnixMain + 0x3c) [0x23759c]
=========     Host Frame:/lib/x86_64-linux-gnu/libc.so.6 (__libc_start_main + 0xe7) [0x21b97]
=========     Host Frame:python [0x1dfb50]

(...the rest is irrelevant.)
```
It seems we are protected by CuPy's memory pool, though, which always pads allocated memory to ensure alignment; I can't reproduce this without disabling the pool. In addition, reproducing this requires carefully crafted array sizes.

cc @grlee77 @anaruse

leofang:
These all reproduce the bug if the memory pool is disabled.
```python
# case 2 (NVlabs/cub190)
indptr = xp.array([0, 2, 5, 7, 10, 14, 17, 19, 22, 24],
                  dtype=cupy.int32)
indices = xp.array([1, 3, 0, 2, 4, 1, 5, 0,
                    4, 6, 1, 3, 5, 7, 2, 4,
                    8, 3, 7, 4, 6, 8, 5, 7], dtype=cupy.int32)
data = xp.ones((indices.size,), dtype=cupy.float64)
m = sp.csr_matrix((data, indices, indptr), shape=(9, 9))
x = xp.ones((9,), dtype=cupy.float64)
```
```python
# case 3 (NVlabs/cub#139) 
indptr = xp.arange(6, dtype=cupy.int32)
indices = xp.arange(5, dtype=cupy.int32)
data = xp.ones((indices.size,), dtype=cupy.float64)
m = sp.csr_matrix((data, indices, indptr), shape=(5, 5))
x = xp.ones((5,), dtype=cupy.float64)
```

asi1024:
I think the test is not needed because we don't want to guarantee that CUB has a bug.
Why not add a  workaround in SpMV implementation for these kinds of cases?

leofang:
> I think the test is not needed because we don't want to guarantee that CUB has a bug.

Yes, so I labeled `DO NOT MERGE` 🙂 

Actually this brings up an interesting question I've been thinking recently, perhaps @asi1024 @kmaehashi you have discussed internally? Since we have our own copy of CUB, can we also manage our own CUB patches? There're a few bugs that would be nice to address, such as this SpMV bug and the oversize histogram https://github.com/thrust/cub/pull/38. I am under the impression that the Thrust/CUB team has very little bandwidth to address the issues that we're interested in (someone from NVIDIA confirmed this), for example this SpMV thing has been dragging over a year, so perhaps we better fix it ourselves.

The cons I can think of:
1. Maintenance overhead
2. CUDA 11 users can no longer use the latest version they have

> Why not add a workaround in SpMV implementation for these kinds of cases?

I will confirm if zero-padding would indeed trigger this. If so, we just need to check whether the padding is zero (`indptr.data.mem.size >= (indptr.size - 1) * indptr.dtype.itemsize`), and avoid calling CUB if it is. We can also check the current allocator, but I think it's kinda hacky. What do you think @asi1024?

leofang:
> Actually this brings up an interesting question I've been thinking recently

I opened an issue (#3723) for this discussion.

leofang:
@asi1024 PTAL.

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

leofang:
Thanks!

PR #3681, opened 2020-07-28
Re-add the `cudnn` and `profiler` `cupy.cuda` modules

emcastillo:
They were deleted from the namespace with the `cupy_backends` change

### DIFF
--- a/cupy/cuda/cudnn.py
+++ b/cupy/cuda/cudnn.py
@@ -1,0 +1,1 @@
+from cupy_backends.cuda.libs.cudnn import *  # NOQA
--- a/cupy/cuda/profiler.py
+++ b/cupy/cuda/profiler.py
@@ -1,0 +1,1 @@
+from cupy_backends.cuda.libs.profiler import *  # NOQA


leofang:
I thought `profiler` was imported in `__init__.py`? 
https://github.com/cupy/cupy/blob/e738ace716c6e099318ce102a693119c108b8d92/cupy/cuda/__init__.py#L19
And can't we do the same thing to `cudnn`?

kmaehashi:
This PR intends to make `import cupy.cuda.cudnn` possible which was working in v8.0.0b4.

kmaehashi:
There files have been removed from cupy/cuda in `cupy_backends` refactoring:

```
-cublas.pyx
-cudnn.pyx
-curand.pyx
-cusolver.pyx
-cusparse.pyx
-cutensor.pyx
-driver.pyx
-nvrtc.pyx
-profiler.pyx
-runtime.pyx
```

However providing wrapper (`... import *`) for all these sounds overkill. My idea is to only provide wrapper for these 3 modules:

- documented namespaces (namespaces used in [documented APIs](https://docs.cupy.dev/en/latest/reference/cuda.html)): profiler, runtime
- undocumented namespace, but often directly imported for historical reasons (Chainer): cudnn

kmaehashi:
pfnCI, test this please.

PR #3682, opened 2020-07-28
Adding fused kernel support to `generic_filter`.

coderforlife:
This is to added fused function support to `cupyx.scipy.ndimage.generic_filter`. This was determined to be "a bit rough" previously and it still is (since there haven't been any changes). This PR is being submitted to be a place for discussion about how to do it properly. Some things already discussed:

- The functions `_indexing`, `_broadcast_to`, and `_transpose` are actually just the functions https://github.com/cupy/cupy/blob/b367bfacea289009414e88b2b58a6e3243683afe/cupy/core/_routines_indexing.pyx#L340 https://github.com/cupy/cupy/blob/2804af926cead8d2e80665283473a71212e824b0/cupy/core/_routines_manipulation.pyx#L415 https://github.com/cupy/cupy/blob/2804af926cead8d2e80665283473a71212e824b0/cupy/core/_routines_manipulation.pyx#L321 but only operating on the shapes/slices and not the arrays (and thus don't create views). Also, all error checking was removed since it would happen elsewhere (additionally, it seems that I took more liberty with `_indexing` and may have taken from a few other sources to complete it). The is that maybe those internal functions could be broken into two parts: one that deals with just taking shape/slices and returning shape/slices and then another which calls that one and makes the view. Both of these functions would then be internally exposed.
- The functions `_get_fused_kernel` and `_get_new_fused_kernel` should be moved in with the fused functions. Basically, the idea is to acquire the underlying kernel that would be used for a particular set of inputs but not actually run the kernel. Preferably they could be called without actual arrays but simply with shape/dtype tuples, but that is an "extra". The functions right now *do* cause the kernel to be run but that is because they had no other way of building the cache.
- Most of the other code is to take a `FusionKernel` and turn it into a *serial* function that can be called. Since the `generic_filter` is already providing the massive parallelism, the kernel being called should only be serial. The question is what can be added to the `FusionKernel` to make this process easier and less rough.


asi1024:
Could you elaborate a bit? Do you mean that you want some kind of interface in `FusedKernel`?

coderforlife:
The very high-level goal here would be to support fuseable Python functions being called from other CUDA kernels. For the immediate term we want to be able to support calling fuseable Python functions from the `ndimage.filter` kernel so that generic filters are possible without the user writing CUDA code for them.

It seems reasonable that a fused function would have a method to get the `FusedKernel` or `Fusion` for a given set of arguments without actually invoking the kernel (and without needing to muck around in the cache data structure like the PR code currently does).

Then, once you have a `FusedKernel` or `Fusion` value, you could call a method that would return the complete generated CUDA code for it (preferably with a flag to suppress parallelism) and the name of the function that needs to be called. If not able to do all of this, then adding some helper methods to make the code in this PR less dependent on the internals of these classes would be great.

asi1024:
A `function` parameter taken by `scipy.ndimage.generic_filters` should be a function to apply at each element, but `FusedKernel` represents a combined function of array operations. So, `FusedKernel` is not so useful for this feature.

https://github.com/cupy/cupy/pull/3392 may be more related to this PR. The details of this design are still under discussion, but I think at least it will support an easy way to write a device function.

asi1024 (commented):
This PR uses private methods of Fusion class and replaces CUDA code string. It is hard to maintain, so we would rather not accept this PR unless the users have strong use cases. 
I feel `new_fusion.Fusion` will be rarely needed to support. How do you think?

asi1024:
If you still want to support the new fused kernel, we prefer to add an interface for `generic_filter` in fused kernel to avoid hacky string conversion!

leofang:
> If you still want to support the new fused kernel, we prefer to add an interface for `generic_filter` in fused kernel to avoid hacky string conversion.

I support this! With hindsight the support for `RawKernel` and `ReductionKernel` in `generic_filter` was also kinda hacky -- these kernels can't be a fully functional one that can be used standalone, but must obey certain syntax requirement tailored for the need of `generic_filter`. It's probably a better idea -- if no such interface exists -- to ask users to insert a plain code string, similar to what's done with [cuFFT callbacks](https://docs.cupy.dev/en/latest/reference/generated/cupy.fft.config.set_cufft_callbacks.html#cupy.fft.config.set_cufft_callbacks) (they already need to do this anyway, with an additional step to wrap it with `RawKernel`/`ReductionKernel`).

kmaehashi:
Let me close this one for now, as it is inactive for a while and some additional work is necessary to merge this feature. Feel free to reopen if you have time to restart working on this one.

PR #3683, opened 2020-07-29
Adding C++ constructors for CArray and CIndexer

coderforlife:
Currently the `CArray` and `CIndexer` types are "constructed" by allocating the memory and filling it with the values from Python. Something similar can be done in C++ but it is annoying. For example: https://github.com/cupy/cupy/blob/452ec5684226543b38cdcaaa52a5c4a356421586/cupyx/scipy/ndimage/_filters_generic.py#L134-L155 That doesn't even include 2 helper functions `_contig_strides` and `_assign_array`.

This PR adds both basic constructors that just take pointers and nicer constructors that take initializer lists (if compiled with `-std=c++11` or better) to `CArray` and `CIndexer`. Adding these constructors have no impact on other functions and those classes behave just as they always did. This greatly simplifies the C++ code required to create objects of these types to these possibilities:

```C++
int shape[3] = {3, 4, 5};
int strides[3] = {0, 0, 40};
CArray<T, 3> arr(data, shape, strides);
// or if C++11 is available:
CArray<T, 3> arr(data, {3, 4, 5}, {0, 0, 40});
// or if C-contiguous:
CArray<T, 3, true> arr(data, {3, 4, 5}); // strides are optional if C-contiguous
// and with CIndexer as well:
CIndexer<3> ind(shape); // indices are optional, default is all 0s
// or with C++11
CIndexer<3> ind({3, 4, 5});
// or even simpler:
CIndexer<3> ind = {3, 4, 5};
```

This PR adds one other little feature: indexing `CArray` with an initializer list (with `-std=c++11` or better): `arr[{0, 1, 2}];`.

I have done basic testing with `nvcc` and `nvrtc` compiling with and without `-std=c++11` utilizing `RawKernel` and `ElementwiseKernel`. Will add unit tests once it is determined that this concept is appropriate.


emcastillo:
The main use case for this is to create Custom indexers and arrays inside of `RawKernels` so we can do stride tricks inside the kernels itself. I think this is right?

I am ok with the idea and the implementation. However, I a don't know how many codes can benefit from this besides the generic ndimage filters. Probably in such cases reshaping the arrays before calling the actual kernel should be enough for most of the cases?

I agree with merging this, I just want to completely understand the implications. 😅

coderforlife:
That is one use of it. In generic ndimage filters it is actually so that other kernels can be called from the main kernel. All of the kernels generated through `ElementwiseKernel`, `ReductionKernel`, and `FusionKernel` take these objects so if calling them C++ these objects need to be constructed.

As a side note I have been calling these functions directly and *not* with dynamic parallelism. But I am unsure if I should be using dynamic parallelism for them eventually (although I believe that would end up hurting in many cases).

emcastillo:
Sorry for the delay,
Yeah, that sounds totally legit, I think it is better to have this PR merged.

emcastillo:
Jenkins, test this please

emcastillo (approved):

leofang:
Jenkins, test this please

leofang:
Timed out. Retry...

leofang:
Jenkins, test this please

coderforlife:
Seems that a default ctor needs to be added for the classes to support the old way of using them.

leofang:
I think it's best to follow the rule of three (or five). It took me a good week to realize. Glad that the error message in your case is so clear 😂 I got indeterministic errors in one of the PRs...

If we are solely on C++11 or above it could be as simple as adding `= default`, but too bad we can't do that yet...

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

coderforlife:
Not quite sure why those a failing. It doesn't look like its during build but instead during tests and in each of the failures it is a different test (one in signaltools and one in morphology). While those are both things that I have interacted with, neither of them use directly allocation of CArray or CIndexer.

leofang:
This is likely just a simple timeout. It happens when any of the core headers is touched, as every kernel needs to be recompiled (usually ccache helps reduce this cost greatly if recompilation is not needed).

Jenkins, test this please

emcastillo:
The testing time is excessive, I want to check locally whats going on

cupy/core/include/cupy/carray.cuh
```diff
@@ -1,5 +1,20 @@
 #pragma once
 
+#if __cplusplus >= 201103
```
leofang, line 3: This won't work for MSVC. You need something like this:
https://github.com/cupy/cupy/blob/b1b215037c29cf92a0c12811ead22730917630cf/cupy/core/include/cupy/complex/complex.h#L91

leofang (changes requested):

cupy/core/include/cupy/carray.cuh
leofang: IIRC MSVC has full C++11 support starting vs2015, but my memory could be a bit vague.

coderforlife: According to https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-160 you are right (with a few odd-ball exceptions that don't apply here).

emcastillo:
Jenkins, test this please

emcastillo:
I measured locally and found no performance degradation.
Tests will take a long time because all the kernels will be rebuilt, but that is expected.

emcastillo (approved):

coderforlife:
Problem seems to be that Jitiffy now adds `std:enable_if` when you include `<type_traits>` which wasn't available when I started this. So I think it should just be `#include <type_traits>`... It is interesting that it is only trigger in particular setups (and not the setup on my own machine so I can't test it directly).

The two warnings about `warning: dynamic initialization in unreachable code` are because of the `if (c_contiguous) { ... }`. Apparently there is no real good way to get rid of it (you can specialize the entire class based on `c_contiguous`, create additional functions, or suppress the warning like is done in the Eigen library).

coderforlife:
I have fixed the compile-time errors (and warnings) and dealt with duplicate code in the carray and cindexer.

leofang (changes requested):
Hi @coderforlife Could you revert the last commit (c5753ac) please? The identifiers there are semantically correct, and the warnings are harmless. It's just that older NVCC/NVRTC messed up so they appear; they've been fixed recently. (I am on my cell so take my words for now, and I will try to dig out where I found this.) Also, this has no problem with HIPCC.

Another thing, I would suggest @emcastillo to run this change through the Windows test branch #4409 if possible, and see if there could be any potential issues on Windows.

coderforlife:
Reverted changes to complex.h - I figured there could be an objection to that one!

@emcastillo mind running jenkins and the Windows one?

emcastillo:
Jenkins, test this please

coderforlife:
Got killed with SIGTERM, guessing that means it timed out...

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

cupy/core/include/cupy/carray.cuh
```diff
@@ -228,32 +324,19 @@ public:
...
-        ptr += static_cast<ptrdiff_t>(strides_[dim]) * (i % shape_[dim]);
-        i /= shape_[dim];
-      }
-      if (ndim > 0) {
-        ptr += static_cast<ptrdiff_t>(strides_[0]) * i;
+          index_t shape_dim = static_cast<index_t>(shape_[dim]);
+          ptr += static_cast<index_t>(strides_[dim]) * (i % shape_dim);
+          i /= shape_dim;
       }
+      ptr += static_cast<index_t>(strides_[0]) * i;
```
leofang, line 337: Question: We don't need to check if `ndim>0` here or not because in that case `strides_[0]` is 0?

coderforlife:
No, because `ndim=0` is a specialized template class that does everything different. That code can never be reached with `ndim=0`.

leofang:
Ah right, thanks for reminding me @coderforlife! Nice work.

PR #3689, opened 2020-07-29
Fix rpath for wheel build

kmaehashi:
Fix the issue that rpath generation was expecting that all modules that use shared library are under `cupy.*`.
This is needed for `cuda_backends`.


emcastillo:
Jenkins, test this please

PR #3690, opened 2020-07-30
[WIP] Fix Windows build failure of cuSparse generic API

kmaehashi:
Fix build failure using CUDA 10.1 / 10.2 on Windows.

I retested #3491 and found that:

* CuPy + cuSPARSE 10.2.0 (= CUDA 10.1.168 = Update 1) cannot be built on Windows.
* CuPy + cuSPARSE 10.3.0 (= CUDA 10.1.243 = Update 2) can be built on Windows, but actually it is not working.

So we should provide generic API support in Windows for cuSPARSE 11.1.0 (= CUDA 11.0 stable) or later.

ref: #3491


emcastillo:
Jenkins, test this please

PR #3691, opened 2020-07-30
Fix broken message format

kmaehashi:
(this is not a release blocker as this is "should not happen" route)


takagi:
pfnCI, test this please.

takagi:
LGTM!

PR #3702, opened 2020-08-01
[WIP] Check not to always raise accepted errors for all combinations

asi1024:


cupy/testing/helper.py
```diff
@@ -1021,14 +1037,21 @@ def for_orders(orders, name='order'):
...
+                    ret = impl(self, *args, **kw)
+                    if ret in (True, None):
+                        exists_true_case = True
                 except Exception:
                     print(name, 'is', order)
                     raise
 
+            if not exists_true_case:
+                raise AssertionError(
+                    'An accepted error was raised in all orders. '
```
emcastillo, line 1053: Just a nitpick
Maybe "all combinations" instead of "all orders"?

Feel free to ignore this.

emcastillo:
This PR is to detect situations when all the cases of a test raise an error. 
There are cases when `accept_error` is set to a certain type of Exception, but a bug in the routine may cause it to launch that same exception due to a different error.
If this happens for all cases, this change can help to identify this. I support this to be added.

@leofang @kmaehashi

asi1024:
The current implementation reraises the AssertionError if all cases raise the same error.

kmaehashi:
Let me close this one as it is inactive for a long time. Let's revisit when the need arises.

PR #3706, opened 2020-08-03
Remove unnecessary imports in `cupy._sorting`

takagi:
Rel #3420.

This PR removes unnecessary imports from `cupy._sorting` submodule.

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -64,7 +64,6 @@ from cupy import manipulation  # NOQA
 from cupy import padding  # NOQA
 from cupy import polynomial  # NOQA
 from cupy import random  # NOQA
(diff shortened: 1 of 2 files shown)


kmaehashi:
These must be fixed?

```
_misc/memory_ranges.py:from cupy._sorting import search
manipulation/basic.py:from cupy._sorting import search
```

kmaehashi:
pfnCI, test this please.

takagi:
I suppose we should leave the lines as they are, but you mean?

kmaehashi:
Ah, ok I misunderstood that `from cupy._sorting import search` won't work after this fix. Locally confirmed it's still working. LGTM!

PR #3707, opened 2020-08-03
Rename `cupy.binary` submodule to `cupy._binary`

takagi:
Rel #3420.

This PR renames `cupy.binary` submodule to `cupy._binary` to indicate it is CuPy's private submodule and not intended to be used by users.


emcastillo:
I think this is not currently conflicting with Open PRs?

emcastillo:
Jenkins, test this please

PR #3709, opened 2020-08-03
Remove unnecessary import of `cupy.fft`

takagi:
Rel #3420.

This PR removes unnecessary import of `cupy.fft`, which is naturally imported without the line.


takagi:
Let me close this PR as the `numpy.fft` is available via `import numpy; numpy.fft` or `from numpy import *; fft`.

PR #3718, opened 2020-08-04
Handle `__cuda_array_interface__` in `asnumpy`

jakirkham:
If the object given to `asnumpy` supports `__cuda_array_interface__`, then coerce it to a CuPy `ndarray` and use `.get(...)` to transfer it back to host (as would be done with a CuPy `ndarray`).

### DIFF
--- a/tests/cupy_tests/creation_tests/test_from_data.py
+++ b/tests/cupy_tests/creation_tests/test_from_data.py
@@ -503,7 +503,7 @@ class TestCudaArrayInterface(unittest.TestCase):
     @testing.for_all_dtypes()
     def test_asnumpy(self, dtype):
         a = testing.shaped_arange((2, 3, 4), cupy, dtype)
-        b = DummyObjectWithCudaArrayInterface(a, self.ver, self.strides))
+        b = DummyObjectWithCudaArrayInterface(a, self.ver, self.strides)
         a_cpu = cupy.asnumpy(a)
         b_cpu = cupy.asnumpy(b)
         testing.assert_array_equal(a_cpu, b_cpu)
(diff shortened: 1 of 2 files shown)


pentschev:
@jakirkham maybe it would be worth adding a test? We probably don't want to depend on a 3rd-party library, but we could create a new class that would just implement the `__cuda_array_interface__`.

leofang:
One place the test can go is `TestCudaArrayInterface` in `tests/cupy_tests/creation_tests/test_from_data.py`, but I suppose you could also add it to where `asnumpy` is tested.

kmaehashi:
Thanks! Could you tell us the use-case for this feature? The code itself LGTM.

jakirkham:
Thanks all! Was looking for where to add a test 🙂

Was using it with RMM `DeviceBuffer`s and cuDF `Buffer`s, both of which support `__cuda_array_interface__`, with some success. This just made that code a bit more succinct and seemed to perform reasonably well.

tests/cupy_tests/creation_tests/test_from_data.py
```diff
@@ -500,6 +500,14 @@ def test_with_zero_size_array(self, dtype):
         assert a.data.ptr == 0
         assert a.size == 0
 
+    @testing.for_all_dtypes()
+    def test_asnumpy(self, dtype):
+        a = testing.shaped_arange((2, 3, 4), cupy, dtype)
+        b = DummyObjectWithCudaArrayInterface(a, self.ver, self.strides))
```
leofang, line 506: I guess this syntax error made flake8 nuts?
```suggestion
        b = DummyObjectWithCudaArrayInterface(a, self.ver, self.strides)
```

jakirkham: Thanks Leo! 😂

leofang:
Thanks, John. Travis is happy now!

Jenkins, test this please

PR #3730, opened 2020-08-06
Add a cuFFT plan cache

### LINKED ISSUE #3588: Reuse cufft plan objects
Reusing a plan object have a significant impact in the complete fft performance. I measured a ~500us time drop.

We should try to create a `PlanAllocator` or a pool of plans to reuse objects according to the requested size.

(Plans allocate gpu memory so we have to be careful when reusing them).

CC @leofang @asi1024 @grlee77

leofang:
***UPDATE***: Close #3588.

This PR implements a least recently used (LRU) cache for cuFFT plans. The implementation is done in Cython to minimize the Python overhead; yet, I still use cdef classes (instead of pointers to structs) to avoid managing memory myself, and cdef'ing as much as I can. 

Properties of this cache:
- **Per-thread, per-device**
- The "size" of the cache can be set by both the number of plans *and* the amount of (GPU) memory used by the plans (i.e., the work areas)
- Enabled by default (with size = 16, which I picked ungroundedly)
- Good performance:
  - Greatly reduced the CPU overhead of plan allocation, especially with non-prime lengths (#3556): https://github.com/cupy/cupy/pull/3730#issuecomment-670061799
  - Fast access time (< 1 us get/set): https://github.com/cupy/cupy/pull/3730#issuecomment-671725253
  - Bonus: a bit of GPU time reduction (as certain plan allocations would launch a few kernels): https://github.com/cupy/cupy/pull/3730#issuecomment-670061799
- Adding and removing a multi-GPU plan is done collectively among all participating GPUs' caches to respect the memory limitation.
- A few handles are exposed to `cupy.fft.config`:
  - Documented: `get_plan_cache()`, `show_plan_cache_info()`
  - Undocumented: the five APIs from scipy/scipy#12512
  - I do not expect they need to be used, though
  - **Question: How can I generate a doc page for `PlanCache` without explicitly referencing it in the autosummary?**

What is **NOT** done in this PR (see the discussions in the replies below):
- Lock stream on to work area  
- Manage a work area pool by the cache

I think it's out of scope, requires careful planning, and the performance gain, if any, is unknown.

---

~~Work in progress. Description to follow. All tests passed locally.~~
~~Aim to address #3588 and follow scipy/scipy#12512.~~

TODO:
- [ ] **Decide to what extent we would like to expose the cache to end users** (the public API from scipy/scipy#12512 is a different thing)
- [x] Per-device cache
- [x] Merge `add_multi_gpu_plan()` and `__setitem__()`
- [x] Add PR description 
- [x] Add tests
- [x] Add comments to code
- [x] Mark the public API from scipy/scipy#12512 experimental
- [ ] ~~Add `get_fft_plan` outcomes to the cache?~~


leofang:
Jenkins, test this please

leofang:
Stupid me...🤦🏻‍♂️

leofang:
***HUGE*** reduction in CPU time (and some improvements in GPU time as well):
```python
import cupy as cp
from cupyx.time import repeat

# choosing non-prime numbers could make plan generation time longer?
a = cp.random.random((198, 597, 418)).astype(cp.complex128)

# typical workload of ours
def fft_roundtrip(a, axis=None):
    out = cp.fft.fftn(a, axes=axis)
    out = cp.fft.ifftn(out, axes=axis)
    return out

n_repeat = 10

print("with cache:")
print(repeat(fft_roundtrip, (a, (1,2)), n_repeat=n_repeat))
print(repeat(fft_roundtrip, (a, (0,1,2)), n_repeat=n_repeat))
print(cp.fft.cache.plan_cache, '\n\n')

cp.fft.cache.plan_cache.set_size(0)  # disable the cache
print("without cache:")
print(repeat(fft_roundtrip, (a, (1,2)), n_repeat=n_repeat))
print(repeat(fft_roundtrip, (a, (0,1,2)), n_repeat=n_repeat))
print(cp.fft.cache.plan_cache)
```
output:
```
with cache:
/home/leofang/cupy/cupyx/time.py:56: FutureWarning: cupyx.time.repeat is experimental. The interface can change in the future.
  util.experimental('cupyx.time.repeat')
fft_roundtrip       :    CPU:  251.862 us   +/-45.510 (min:  221.953 / max:  380.593) us     GPU-0:294705.273 us   +/-66.591 (min:294637.634 / max:294866.943) us
fft_roundtrip       :    CPU:  256.680 us   +/-42.280 (min:  230.437 / max:  381.849) us     GPU-0:462417.114 us   +/-671.659 (min:460680.573 / max:462825.470) us
-------------- cuFFT plan cache --------------
cache enabled? True
current / max size: 2 / 16 (counts)
current / max memsize: 7141533696 / (unlimited) (bytes)

cached plans (least used first):
key: ((597, 418), (597, 418), 1, 249546, (597, 418), 1, 249546, 105, 198, 'C', 2, None), plan type: PlanNd, memory usage: 3873374208
key: ((198, 597, 418), (198, 597, 418), 1, 1, (198, 597, 418), 1, 1, 105, 1, 'C', 2, None), plan type: PlanNd, memory usage: 3268159488 


without cache:
fft_roundtrip       :    CPU:297780.415 us   +/-304.553 (min:297310.701 / max:298156.878) us     GPU-0:297769.476 us   +/-301.531 (min:297322.083 / max:298143.738) us
fft_roundtrip       :    CPU:466226.026 us   +/-3733.645 (min:464298.991 / max:476941.334) us     GPU-0:466205.548 us   +/-3734.419 (min:464284.668 / max:476923.767) us
-------------- cuFFT plan cache --------------
cache enabled? False
current / max size: 0 / 0 (counts)
current / max memsize: 0 / (unlimited) (bytes)

cached plans (least used first):
```
cc: @peterbell10 @grlee77 @mnicely @emcastillo

leofang:
With the cache enabled (whose size is defaulted to 16), it saves 25% end-to-end runtime for running `pytest tests/cupyx_tests/scipy_tests/fft_tests/test_fft.py`.

leofang:
Jenkins, test this please

mnicely:
> _**HUGE**_ reduction in CPU time (and some improvements in GPU time as well):
> 
> ```python
> import cupy as cp
> from cupyx.time import repeat
> 
> # choosing non-prime numbers could make plan generation time longer?
> a = cp.random.random((198, 597, 418)).astype(cp.complex128)
> 
> # typical workload of ours
> def fft_roundtrip(a, axis=None):
>     out = cp.fft.fftn(a, axes=axis)
>     out = cp.fft.ifftn(out, axes=axis)
>     return out
> 
> n_repeat = 10
> 
> print("with cache:")
> print(repeat(fft_roundtrip, (a, (1,2)), n_repeat=n_repeat))
> print(repeat(fft_roundtrip, (a, (0,1,2)), n_repeat=n_repeat))
> print(cp.fft.cache.plan_cache, '\n\n')
> 
> cp.fft.cache.plan_cache.set_size(0)  # disable the cache
> print("without cache:")
> print(repeat(fft_roundtrip, (a, (1,2)), n_repeat=n_repeat))
> print(repeat(fft_roundtrip, (a, (0,1,2)), n_repeat=n_repeat))
> print(cp.fft.cache.plan_cache)
> ```
> 
> output:
> 
> ```
> with cache:
> /home/leofang/cupy/cupyx/time.py:56: FutureWarning: cupyx.time.repeat is experimental. The interface can change in the future.
>   util.experimental('cupyx.time.repeat')
> fft_roundtrip       :    CPU:  251.862 us   +/-45.510 (min:  221.953 / max:  380.593) us     GPU-0:294705.273 us   +/-66.591 (min:294637.634 / max:294866.943) us
> fft_roundtrip       :    CPU:  256.680 us   +/-42.280 (min:  230.437 / max:  381.849) us     GPU-0:462417.114 us   +/-671.659 (min:460680.573 / max:462825.470) us
> -------------- cuFFT plan cache --------------
> cache enabled? True
> current / max size: 2 / 16 (counts)
> current / max memsize: 7141533696 / (unlimited) (bytes)
> 
> cached plans (least used first):
> key: ((597, 418), (597, 418), 1, 249546, (597, 418), 1, 249546, 105, 198, 'C', 2, None), plan type: PlanNd, memory usage: 3873374208
> key: ((198, 597, 418), (198, 597, 418), 1, 1, (198, 597, 418), 1, 1, 105, 1, 'C', 2, None), plan type: PlanNd, memory usage: 3268159488 
> 
> 
> without cache:
> fft_roundtrip       :    CPU:297780.415 us   +/-304.553 (min:297310.701 / max:298156.878) us     GPU-0:297769.476 us   +/-301.531 (min:297322.083 / max:298143.738) us
> fft_roundtrip       :    CPU:466226.026 us   +/-3733.645 (min:464298.991 / max:476941.334) us     GPU-0:466205.548 us   +/-3734.419 (min:464284.668 / max:476923.767) us
> -------------- cuFFT plan cache --------------
> cache enabled? False
> current / max size: 0 / 0 (counts)
> current / max memsize: 0 / (unlimited) (bytes)
> 
> cached plans (least used first):
> ```
> 
> cc: @peterbell10 @grlee77 @mnicely @emcastillo

@awthomp

cupy/fft/fft.py
```diff
@@ -97,16 +98,29 @@ def _exec_fft(a, direction, value_type, norm, axis, overwrite_x,
...
         if plan is None:
             plan = curr_plan
         else:
             raise RuntimeError('Use the cuFFT plan either as a context manager'
                                ' or as an argument.')
+
     if plan is None:
         devices = None if not config.use_multi_gpus else config._devices
-        plan = cufft.Plan1d(out_size, fft_type, batch, devices=devices)
+        # TODO(leofang): do we need to add the current stream to keys?
```
grlee77, line 116: It looks like currently, the plans are associated with the default stream on creation, but when calling the FFT with a plan provided, the plan will be associated wit the current stream. 

https://github.com/cupy/cupy/blob/97573d737beefb7a9ae9f81c294a005ed475bf4f/cupy/cuda/cufft.pyx#L782-L783

In that case, it seems like we shouldn't need to cache the stream as the plan will be reassociated with whatever stream is currently active. Is there a scenario where one wants to do FFTs from separate streams on a single device? If not, then I don't think we would need to redesign anything.

cupy/fft/fft.py
```diff
@@ -373,18 +387,15 @@ def _get_cufft_plan_nd(shape, fft_type, axes=None, order='C', out_size=None):
...
-                        ostride=ostride,
-                        odist=odist,
-                        fft_type=fft_type,
-                        batch=nbatch,
-                        order=order,
-                        last_axis=fft_axes[-1],
-                        last_size=out_size)
+    keys = (plan_dimensions, inembed, istride,
+            idist, onembed, ostride, odist,
+            fft_type, nbatch, order, fft_axes[-1], out_size)
```
grlee77, line 392: shouldn't `cupy.cuda.device.get_device_id()` also be in the list of keys here? (and in the single-GPU case for 1D plans above?)

cupy/fft/fft.py
leofang: > Is there a scenario where one wants to do FFTs from separate streams on a single device?

I feel this is an uncharted territory as running the same FFT plan on multiple streams might drive the plan's internal state and work area nuts. The cuFFT documentation doesn't say anything about this, only that overlapping plan executions with data transfer is possible. I don't even know how we can set up a "stream mutex" to prevent the same plan from being executed on multiple streams, and I'd rather put this burden to end users...

cupy/fft/fft.py
leofang: I think we need a per-thread, per-device cache. I will do the per-device part.

leofang:
@grlee77 @peterbell10 You folks must have much more experience in caching than I do. What do we need to add to the test suite for the plan cache? @peterbell10 If you have time to add tests to scipy/scipy#12512, it'd be lovely as I can steal things from there 😇

grlee77:
Hi @leofang. Thanks for working on this!
I like that you can control based on the total memory size. I had tried that previously with a basic pure-python implementation, just storing the cache in a Python dict, but had never gotten around to finishing it. This cython-based LRU cache implementation should be efficient and looks good to me.

The other idea I wanted to look into, but never tried was to modify the `Plan1d` and `PlanNd` to store the `work_size` (which they already compute as a temporary), e.g.
https://github.com/cupy/cupy/blob/97573d737beefb7a9ae9f81c294a005ed475bf4f/cupy/cuda/cufft.pyx#L752-L759
Then we could have an method to dynamically allocate and deallocate the work area as needed (allowing many plans to be stored without substantial memory overhead). Obviously this would only make sense if the memory allocation is not a substantial portion of the required planning overhead. As far as I can tell the work area (`plan.work_area.mem.size`) is equal in size to the data itself. For data that is a substantial fraction of the GPU memory this can severely limit the number of plans that can be stored. If implemented, the default behavior should be as it is currently for performance reasons, but it seems like it could be of interest in memory-limited scenarios.

cupy/fft/fft.py
grlee77: > running the same FFT plan on multiple streams

I am not suggesting one would ever use the same plan simultaneously from separate streams, as I don't see how that could possibly work. It seems like a user could potentially try and do that though with the current design, but agree that it is probably a rare scenario and we don't need to try and guard against it here.  

I was asking more if there is a usage scenario for adding the stream attribute on plan creation (and thus also to the plan cache keys), so that multiple plans that are identical aside from stream could be used simultaneously? I agree that we don't need to add this now even if that is a plausible scenario.

grlee77:
> What do we need to add to the test suite for the plan cache? 

I don't have any fancy ideas, just basic stuff like verifying that the cache size and memory limits add up to the expected value for some simple 1d and nd plan cases.

Probably also a couple things to test the LRU aspect: e.g.
   - limit to 2 plans, adding a 3rd plan results in the first plan that was added being removed
   - adding a plan that already exists in the cache moves it to the "most recent" position

cupy/fft/fft.py
grlee77: You mean each device will have it's own independent cache? I think that would be best as I don't think it would make sense to combine memory from separate devices when computing the current cache memory usage.

cupy/fft/fft.py
leofang: Hi Gregory, yes I think we are talking about different scenarios:
1. Same transform, **multiple** plan handles, running in parallel on multiple streams (but each handle on a unique stream),
2. Same transform, **same** plan handle, running in parallel on multiple streams,

by "plan handle" I mean `cufftHandle`. 

I agree both scenarios are complex that they should be considered in a separate PR, but just out of curiosity I re-read the doc, and I feel there might be a solution to address **both** scenarios. Quoting [Streamed cuFFT Transforms](https://docs.nvidia.com/cuda/archive/10.2/cufft/index.html#streamed-cufft-transforms):
> Please note that in order to overlap plans using single plan handle user needs to manage work area buffers. Each concurrent plan execution needs it's exclusive work area. Work area can be set by cufftSetWorkArea function.

If I parse this sentence correctly, it is saying that the situation (2) is possible, as long as the plan is executing on distinct pairs of `(stream, work_area)`. What do you think? I feel it's too good to be true, for example when Run 1 is being executed in the midway and we set the work area and stream for the same plan handle for Run 2, wouldn't Run 1 write to the work area of Run 2 and make everything scrambled?

If an internal lock mechanism makes this possible, then in principle the situation (1) is also fixed (by reusing the same handle) as we still need to allocated multiple work areas anyway, although getting distinct handles is easier to manage.

cupy/fft/fft.py
leofang: Done in 737d0ab.

leofang:
@pentschev Just wanna confirm: A per-thread, per-device cache will make dask-cuda happy, right?

leofang:
Oh shoot, it looks like for multi-gpu plans we need to add a node to every device's cache...Is it worth the effort? 😂

pentschev:
> @pentschev Just wanna confirm: A per-thread, per-device cache will make dask-cuda happy, right?

Generally speaking, I'd say yes. Each worker in dask-cuda uses a single-thread and single-GPU, so that should be fine.

> Oh shoot, it looks like for multi-gpu plans we need to add a node to every device's cache...Is it worth the effort? 😂

I'm not sure what do you mean by node. I believe in that scenario you're not talking about dask-cuda though, right?

leofang:
> Generally speaking, I'd say yes. Each worker in dask-cuda uses a single-thread and single-GPU, so that should be fine.

Great, thanks Peter!

> I'm not sure what do you mean by node. I believe in that scenario you're not talking about dask-cuda though, right?

Sorry, I was just thinking out loud 😅 By node I meant a node in my LRU cache here.

leofang:
OK, now we have a per-thread, per-device cache. All tests passed locally. Once the CI is happy, I'll add LRU-specific tests (https://github.com/cupy/cupy/pull/3730#issuecomment-670170866) and update the PR description to make the review easier.

Jenkins, test this please

leofang:
Jenkins was dead. Let me retry...

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
I think the failing `bsplines` tests are irrelevant. All plan cache tests passed locally (but one failed on the CI which has 8 GPUs because I (again) misunderstood the meaning of `@testing.multi_gpu(N)`; it should be fixed now).

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
< 1 us insertion & fetch times in average with a size-100 cache:
```python
import cupy as cp
import numpy as np
from cupyx.scipy.fftpack import get_fft_plan
from time import perf_counter


cache = cp.fft.config.get_plan_cache()
cache.set_size(100)

a = cp.random.random(1000000)
plan = get_fft_plan(a)
runs = 100

settime = 0.0
for i in range(runs):
    t1 = perf_counter()
    cache[(i,)] = plan
    t2 = perf_counter()
    settime += t2 - t1
print("insert time:", settime/runs*1000000, 'us')

gettime = 0.0
keys = np.arange(runs)
np.random.shuffle(keys)
for key in keys:
    key = (key,)
    t1 = perf_counter()
    plan4 = cache[key]
    t2 = perf_counter()
    gettime += t2 - t1
print("retrive time:", gettime/runs*1000000, 'us')

# check key is correct
iterator = iter(cache)
keys = keys[::-1]
for key in keys:
    assert key == next(iterator)[0]
```

output:
```
$ python test_fft_cache_1_pretty.py
insert time: 0.5785771645605564 us
retrive time: 0.36641256883740425 us
```

leofang:
This PR is ready to review! I updated the PR description as promised. I also left a question:

> * Question: How can I generate a doc page for `PlanCache` without explicitly referencing it in the autosummary?

leofang:
Jenkins, test this please

leofang:
cc: @asi1024 @grlee77 @peterbell10 @peterbell10 @mnicely @awthomp

cupy/fft/cache.pyx
```diff
@@ -0,0 +1,599 @@
...
+    """A per-thread, per-device, least recently used (LRU) cache for cuFFT
+    plans.
+
+    Args:
+        size (int): The number of plans that the cache can accommodate. The
+            default is 16. Setting this to ``-1`` will make this limit ignored.
+        memsize (int): The amount of GPU memory, in bytes, that the plans in
+            the cache will use for their work areas. Default is ``-1``, meaning
+            it is unlimited.
+        dev (int): The ID of the device for which the cache targets.
```
grlee77, line 149:
```suggestion
        dev (int): The ID of the device that the cache targets.
```

cupy/cuda/cufft.pyx
```diff
@@ -327,7 +338,7 @@ class Plan1d(object):
         if fft_type != CUFFT_C2C and fft_type != CUFFT_Z2Z:
             raise ValueError('Currently for multiple GPUs only C2C and Z2Z are'
                              ' supported.')
-        if isinstance(devices, list):
+        if isinstance(devices, (tuple, list)):
```
grlee77, line 341: This could be `collections.abc.Iterable` to match the ValueError message below, but I think it is probably best to keep it as-is so we don't allow iterables like `cupy.ndarray`.

cupy/fft/cache.pyx
```diff
@@ -0,0 +1,599 @@
...
+               from cupy.cuda import Device
+               from cupy.fft.config import get_plan_cache
+
+               # get the cache for Device n
+               with Device(n):
+                   cache = get_plan_cache()
+                   cache.set_size(0)  # disable the cache
+
+           In particular, the cache for Device n should be manaipulated under
+           the Device n's context.
```
grlee77, line 172:
```suggestion
           In particular, the cache for device n should be manipulated under
           the device n's context.
```

cupy/fft/cache.pyx
```diff
@@ -0,0 +1,599 @@
...
+
+        3. This class is thread-safe since by default it is created on a
+           per-thread basis. When starting a new thread a new cache is not
```
grlee77, line 175:
```suggestion
           per-thread basis. When starting a new thread, a new cache is not
```

cupy/fft/cache.pyx
```diff
@@ -0,0 +1,599 @@
...
+           initialized until :func:`~cupy.fft.config.get_plan_cache` is
+           called or when the constructor is manually invoked.
+
+        4. For multi-GPU plans, the plan will be added to all participating
+           GPUs' caches. Upon removal (by any of the caches), the plan will
+           be removed from all participating GPU's caches.
```
grlee77, line 181:
```suggestion
        4. For multi-GPU plans, the plan will be added to each participating
           GPU's cache. Upon removal (by any of the caches), the plan will
           be removed from each participating GPU's cache.
```

cupy/fft/cache.pyx
```diff
@@ -0,0 +1,599 @@
...
+
+# TODO(leofang): remove experimental warning when scipy/scipy#12512 is merged
+cpdef clear_plan_cache():
+    util.experimental('cupy.fft.cache.clear_plan_cache')
+    cdef PlanCache cache = get_plan_cache()
+    cache.clear()
+
+
+cpdef show_plan_cache_info():
+    """Show all of the plan caches's info on this thread.
```
grlee77, line 580:
```suggestion
    """Show all of the plan caches' info on this thread.
```

cupyx/scipy/fftpack/fft.py
```diff
@@ -112,20 +113,31 @@ def get_fft_plan(a, shape=None, axes=None, value_type='C2C'):
         raise ValueError('C2R/R2C PlanNd for F-order arrays is not supported')
 
     # generate plan
+    # (load from cache if it exists, otherwise create one but not cache it)
```
grlee77, line 116:
```suggestion
    # (load from cache if it exists, otherwise create one but don't cache it)
```

cupyx/scipy/fftpack/fft.py
```diff
@@ -112,20 +113,31 @@ def get_fft_plan(a, shape=None, axes=None, value_type='C2C'):
...
     if n > 1:  # ND transform
         out_size = _get_fftn_out_size(
             shape, transformed_shape, axes[-1], value_type)
+        # _get_cufft_plan_nd handles the interaction with plan cache
         plan = _get_cufft_plan_nd(
-            shape, fft_type, axes=axes, order=order, out_size=out_size)
+            shape, fft_type, axes=axes, order=order, out_size=out_size,
+            to_cache=False)
```
grlee77, line 123: Is the assumption here that the reason to call `get_fft_plan` is when you want to handle the plans yourself and so it should not be cached? I think that sounds reasonable.

cupy/fft/fft.py
dicta: There's also a third case here: Same transform, same plan handle, same stream, with multiple FFT launches being able to be overlapped just like any other kernel launch. That's the case that the quoted documentation section is primarily addressing. 

The following common pattern in user code works pretty well:
- create plan, once
- call cufftSetWorkArea, exec FFT N times
- call cudaStreamSynchronize() to wait until all of the FFTs are complete.

I haven't personally tested with setting a separate stream on each invocation, but it is very possible that this would work.

Furthermore, The majority of the memory being used by an allocated plan is the work area that gets allocated by default (when cufftSetAutoAllocation is true). I think it would be a better implementation for the cache if this memory was not allocated at plan creation time -- i.e, always call `cufftSetAutoAllocation(handle, false)` when generating plans, then explicitly assign a work area for each fft() invocation using a normal cupy memory pool as you describe above. The messy part of this is that you do need an event/callback to handle releasing the work area once the FFT completes.

Probably a separate PR to clean this up? But it's pretty important for a cache implementation, with large batch FFTs I've worked on systems where this work area ended up being hundreds of megabytes of GPU memory.

cupyx/scipy/fftpack/fft.py
leofang: That's right. Originally I made it cached here, but I changed my mind as the intention of `get_fft_plan()` was to allow users to manage plans themselves.

cupy/fft/cache.pyx
```diff
@@ -0,0 +1,600 @@
...
+        2. This class can be instantiated by users, but it is discouraged.
+           Instead, we expect the following canonical usage pattern to
+           retrieve a handle to the cache through :func:`get_plan_cache`:
+
+           .. code-block:: python
+
+               from cupy.cuda import Device
+               from cupy.fft.config import get_plan_cache
+
+               # get the cache for Device n
```
leofang, line 166:
```suggestion
               # get the cache for device n
```

awthomp:
> > _**HUGE**_ reduction in CPU time (and some improvements in GPU time as well):
> > ```python
> > import cupy as cp
> > from cupyx.time import repeat
> > 
> > # choosing non-prime numbers could make plan generation time longer?
> > a = cp.random.random((198, 597, 418)).astype(cp.complex128)
> > 
> > # typical workload of ours
> > def fft_roundtrip(a, axis=None):
> >     out = cp.fft.fftn(a, axes=axis)
> >     out = cp.fft.ifftn(out, axes=axis)
> >     return out
> > 
> > n_repeat = 10
> > 
> > print("with cache:")
> > print(repeat(fft_roundtrip, (a, (1,2)), n_repeat=n_repeat))
> > print(repeat(fft_roundtrip, (a, (0,1,2)), n_repeat=n_repeat))
> > print(cp.fft.cache.plan_cache, '\n\n')
> > 
> > cp.fft.cache.plan_cache.set_size(0)  # disable the cache
> > print("without cache:")
> > print(repeat(fft_roundtrip, (a, (1,2)), n_repeat=n_repeat))
> > print(repeat(fft_roundtrip, (a, (0,1,2)), n_repeat=n_repeat))
> > print(cp.fft.cache.plan_cache)
> > ```
> > 
> > 
> > output:
> > ```
> > with cache:
> > /home/leofang/cupy/cupyx/time.py:56: FutureWarning: cupyx.time.repeat is experimental. The interface can change in the future.
> >   util.experimental('cupyx.time.repeat')
> > fft_roundtrip       :    CPU:  251.862 us   +/-45.510 (min:  221.953 / max:  380.593) us     GPU-0:294705.273 us   +/-66.591 (min:294637.634 / max:294866.943) us
> > fft_roundtrip       :    CPU:  256.680 us   +/-42.280 (min:  230.437 / max:  381.849) us     GPU-0:462417.114 us   +/-671.659 (min:460680.573 / max:462825.470) us
> > -------------- cuFFT plan cache --------------
> > cache enabled? True
> > current / max size: 2 / 16 (counts)
> > current / max memsize: 7141533696 / (unlimited) (bytes)
> > 
> > cached plans (least used first):
> > key: ((597, 418), (597, 418), 1, 249546, (597, 418), 1, 249546, 105, 198, 'C', 2, None), plan type: PlanNd, memory usage: 3873374208
> > key: ((198, 597, 418), (198, 597, 418), 1, 1, (198, 597, 418), 1, 1, 105, 1, 'C', 2, None), plan type: PlanNd, memory usage: 3268159488 
> > 
> > 
> > without cache:
> > fft_roundtrip       :    CPU:297780.415 us   +/-304.553 (min:297310.701 / max:298156.878) us     GPU-0:297769.476 us   +/-301.531 (min:297322.083 / max:298143.738) us
> > fft_roundtrip       :    CPU:466226.026 us   +/-3733.645 (min:464298.991 / max:476941.334) us     GPU-0:466205.548 us   +/-3734.419 (min:464284.668 / max:476923.767) us
> > -------------- cuFFT plan cache --------------
> > cache enabled? False
> > current / max size: 0 / 0 (counts)
> > current / max memsize: 0 / (unlimited) (bytes)
> > 
> > cached plans (least used first):
> > ```
> > 
> > 
> > cc: @peterbell10 @grlee77 @mnicely @emcastillo
> 
> @awthomp

Awesome work @leofang! Looking forward to playing more with this PR; this is huge for online signal processing tasks!

cupy/fft/fft.py
leofang: Hi @dicta Thanks for your comments. I have a few questions:

> There's also a third case here: Same transform, same plan handle, same stream, with multiple FFT launches being able to be overlapped just like any other kernel launch. 

If multiple such FFTs are launched on the same stream, how can they be overlapped with each other? Did you mean to overlap FFT with other user kernels and/or data transfer? Isn't that already possible / being done?

> The following common pattern in user code works pretty well:
> 
> * create plan, once
> * call cufftSetWorkArea, exec FFT N times
> * call cudaStreamSynchronize() to wait until all of the FFTs are complete.

On the same stream, yes it will, and it is what will be enabled either by the cache. (FYI, it's already possible by explicitly reusing a plan returned by `get_fft_plan()`.) 

On different streams, this is the Scenario 2 I mentioned above, and I feel we will need x unique buffers for x overlapped streams (x <= N).

> I haven't personally tested with setting a separate stream on each invocation, but it is very possible that this would work.

I guess so too.

> Furthermore, The majority of the memory being used by an allocated plan is the work area that gets allocated by default (when cufftSetAutoAllocation is true). I think it would be a better implementation for the cache if this memory was not allocated at plan creation time -- i.e, always call `cufftSetAutoAllocation(handle, false)` when generating plans, then explicitly assign a work area for each fft() invocation using a normal cupy memory pool as you describe above. The messy part of this is that you do need an event/callback to handle releasing the work area once the FFT completes.

This memory pool thing is actually what @grlee77 mentioned in his https://github.com/cupy/cupy/pull/3730#issuecomment-670154850. But I think it's too messy and hard to make it right. You need to have a mutex protecting every unique pairs of `(stream, buffer)` and also a callback. I don't think it's worth the effort. 

Furthermore, CuPy already has its own memory pool, and we should just reuse it instead of layering on top. The lines of code in this simple module will increase significantly, and it's hard to predict/optimize the mempool for every use cases.

Finally, with the recent pains we are experiencing with newer cuFFT (search "fft" in our issues/PRs), I'd rather not count on cuFFT behaving stably predictively, which we'll need for such a mempool. The cuFFT team never provides a strong guarantee for cuFFT's behaviors, nor do they do a good job in documenting the changes where there's one.

> Probably a separate PR to clean this up? 

Clean up what?

> But it's pretty important for a cache implementation, with large batch FFTs I've worked on systems where this work area ended up being hundreds of megabytes of GPU memory.

That's one of the main scenarios I have in mind that a plan cache will be super useful! Another main issue is there's significant slowdown at plan creation time for certain FFTs (for example, #3556).

leofang:
Test the new memsize test.

Jenkins, test this please

leofang:
Let me try if CUDA 11 complains.

Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
@asi1024 PTAL.

cupy/fft/cache.pyx
```diff
@@ -0,0 +1,600 @@
...
+    # 0: disable cache
+    cdef Py_ssize_t size
+
+    # current number of cached plans
+    cdef Py_ssize_t curr_size
+
+    # total amount of memory for all of the cached plans
+    # -1: unlimited/ignored, cache size is restricted by "size"
+    # 0: disable cache
+    cdef Py_ssize_t memsize
```
peterbell10, line 198: This has a different meaning in the SciPy PR. There it is the max size of a _single_ plan, rather than all plans together. So, for example if you have one really large transform it will evict the entire cache here, even if the rest are small and frequently used transforms. Although having limits for both could make sense.

leofang: Thanks for the heads-up, Peter! I think in the pocketfft case the plan sizes are much smaller and predictable, so a per-plan cap makes better sense. The problem with cuFFT is that we could have super large plans (see https://github.com/cupy/cupy/issues/3777#issuecomment-676561525), so I think it's better to have a cap for all plans, otherwise we might run into many "small" plans, each is say 1 GB, eating up all of the GPU memory.

leofang: But we do need to revise some of the APIs currently exposed to `cupy.fft.config` when working on the SciPy compilance (next PR). Currently they are marked experimental so it's OK to leave them as is.

peterbell10: Well the absolute upper bound is is `cache_size * maxmemsize`. Although your thread-local caches does complicate that quite a bit.

leofang: Yeah, and we also have multi-GPU plans that make it even more complicated 😉 Plus, `cache_size` could be unlimited (=-1).

peterbell10: Well yes, an unlimited `cache_size` doesn't make sense without a total memory limit.

mnicely:
Hi @leofang! What's the ETA on this merge? We're excited to begin testing with cuSignal.
I'm noticing large gaps before my FFT calls with `fftpack.fft()`. I'm betting it's the plan creation. I'd like to use this instead of creating my own cache. :smile:

leofang:
Hi @mnicely! I hope this can get in soon. 🙂

@kmaehashi @asi1024 Please backport this PR.

leofang:
Jenkins, test this please

leofang:
Hi @asi1024 Following #3420 I made the cache module private. PTAL.

asi1024:
Why not use `cupy._util.memoize` or `functools.lru_cache`?

leofang:
If you read my PR description, you'd see they probably won't give us what we need (high performance and dealing with all complexities here). Plus I am the only one so far who has time to experiment with this. The need of having a working cache has been dragging for far too long and let's end it for good.

asi1024:
Why fft plan needs so fast cache? `cupy._util.memoize` may cost several us, but it is ignorable compared to other operations.

leofang:
Fine. Do it yourself.

mnicely:
@leofang What exactly happened??? 

Are fft plan caches not being implemented?

grlee77:
@leofang, please don't get frustrated. Let's continue the discussion here as I think the work you have done so far is valuable and I can tell a lot of work went into this PR.

re: @asi1024
> Why not use cupy._util.memoize or functools.lru_cache?

A useful feature added here is being able to control the maximum size of the cache in terms of total GPU memory usage. I think `functools.lru_cache` only allows controlling the number of items cached.

The only argument I see in favor of `memoize` or `lru_cache` is lower maintenance cost (and lower development cost, but in this case the development of the custom cache has already been completed!). Given that the custom cache here is pretty extensively tested, I am not too worried that it has bugs and will need substantial revisions going forward.

awthomp:
> @leofang, please don't get frustrated. Let's continue the discussion here as I think the work you have done so far is valuable and I can tell a lot of work went into this PR.
> 
> re: @asi1024
> 
> > Why not use cupy._util.memoize or functools.lru_cache?
> 
> A useful feature added here is being able to control the maximum size of the cache in terms of total GPU memory usage. I think `functools.lru_cache` only allows controlling the number of items cached.
> 
> The only argument I see in favor of `memoize` or `lru_cache` is lower maintenance cost (and lower development cost, but in this case the development of the custom cache has already been completed!). Given that the custom cache here is pretty extensively tested, I am not too worried that it has bugs and will need substantial revisions going forward.

@mnicely and I have been closely tracking this PR, as having fast caching features for FFT plans is critical for online signal processing applications and completely applicable to our work on cusignal. The work you've done here has been fantastic, @leofang, and I'm very excited about the simple API.

Let us know how we can help.

emcastillo:
The whole team thinks that this PR is great and we really want to merge it! thanks for all the hard work you put in this @leofang

VolkerH:
I came across this PR last week using a google search (just recent user of cupy, not a team member) and really hope any remaining issues can be addressed and this PR makes it into the code base.

leofang:
Hi @asi1024 Sorry for my outburst. I took a few days off and am now ready to revisit this. I should not have replied at 5am when I was already in a very bad state. I misinterpreted your question as a demand to overturn the current approach. I should have asked for your clarification either here or offline. It wasn't professional...Though I do hope such design discussions could happen earlier. Being able to receive early feedbacks is one of the advantages to develop in the open. 

Also, thanks @grlee77 @mnicely @awthomp @kmaehashi @emcastillo @VolkerH for your kind replies and supports.

So, yes, as @grlee77 helped clarified, `functools.lru_cache` is limited if we want to impose extra constraints such as memory usage. In particular, given that `functools.lru_cache` is implemented as a function decorator (so is `cupy.memoize`) instead of an inheritable class, I do not really have a clear picture how to reuse it while dealing with all these complexities (`Plan1d` vs `PlanNd`, single-GPU vs multi-GPU, per thread & per device cache, etc) that require special logics for insertion and removal; if it were inheritable I could overwrite/wrap some of its methods. (I actually peeked [its internal](https://github.com/python/cpython/blob/d09bead93990eed26ecb8fcd02a8a3f6e8fa73b7/Lib/functools.py#L434-L645) while developing my cache here to gain my confidence.)

`cupy.memoize` has an additional obstacle that I do not know how to remove cached plans from it if needed. Perhaps it's simply because I didn't give enough thoughts to it.

I chose to implement it in Cython because I wanted to minimize the overhead (and looks like we made it), though I didn't benchmark how large it could be if the cache were implemented in Python. I think once cuFFT improves, or when running on AMD GPUs (https://github.com/cupy/cupy/pull/3896#issuecomment-691819683), the CPU overhead could be a potential concern, so I'd rather make this the least thing to consider.

Let me know if you have any questions or thoughts!

leofang:
Jenkins, test this please

cupy/fft/_cache.pyx
```diff
@@ -0,0 +1,612 @@
...
+
+    # data
+    cdef readonly tuple key
+    cdef readonly object plan
+    cdef readonly Py_ssize_t memsize
+    cdef readonly list gpus
+
+    # link
+    cdef _Node prev
+    cdef _Node next
```
asi1024, line 76: It causes a circular reference, so some memory spaces may not be free unexpectedly.
How about using a singly-linked list instead of the current double-linked list? It may cost more time to eject a node from the linked list, but it seems to be small enough compared to the time to make a new plan.

asi1024: Perhaps using Python's primitive list may be simpler than a singly-linked list.

leofang: Hi @asi1024, could you help me understand better about the circular reference issue? Is it because when ejecting a node, I didn't clear out the node's `prev` and `next`?

asi1024: If `_LinkedList` is deleted for some reason (not when ejecting a node), its nodes will not be deleted immediately because they have circular references.
One solution for this issue is changing the data structure to a singly-linked list by deleting a pointer (maybe `next`?). It costs linear time to eject an element because we should trace the linked list from the other side, but I think it is ignorable compared to the cpu-time to make a new plan.

asi1024:
This caching mechanism looks useful also for other purposes, for example cuTENSOR descriptors and other objects that have GPU memory spaces!

cupy/fft/_cache.pyx
leofang: It sounds like we could also just add a destructor for `_LinkedList`, walk through the nodes, and delete them in order?

asi1024: I think both of them work correctly. I prefer singly-linked list or Python's primitive list for maintainability, but adding a destructor is also OK if the difference from the current implementation is small.

asi1024: Sorry, now I notice that I was mentioned at issue #3999.

leofang: Yeah, looks like we can't do `__del__` here. I added a `clear()` routine and invoked it manually when needed, see 8d0705c.

asi1024:
Jenkins, test this please.

cupy/fft/_cache.pyx
leofang: @asi1024 Give me some time to improve the `clear()` routine. A colleague of mine reminded me that `weakref` is exactly for such purposes: https://docs.python.org/3/library/weakref.html#comparing-finalizers-with-del-methods

leofang:
Jenkins, test this please

asi1024:
LGTM!

leofang:
Thanks a lot, @asi1024 and everyone!

> This caching mechanism looks useful also for other purposes, for example cuTENSOR descriptors and other objects that have GPU memory spaces!

Interesting! Do cuTENSOR descriptors use a lot of memory? If so in the future we might consider making this cache to work globally, i.e., store all kinds of plans (cuFFT, cuTENSOR, etc). This also makes it easier to combat with OOM, as we can just clear the cache at once. 

Note though a cache like this could also be useful when the object itself takes little memory but a long time to generate (cuFFT plans have both such weaknesses).

leofang:
Just a follow-up: It seems implementing the cache in Cython is further justified: @mnicely showed in https://github.com/rapidsai/cusignal/pull/254 that  the performance difference with a Python-based cache can be as high as 1.6x, which is very surprising!

PR #3562, opened 2020-07-07
[MNT] Clean up CUB-related stuff

leofang:
~~Blocked by #3461.~~

This PR does three things:
1. Remove the temporary env var `CUPY_CUB_PATH` and related codes from the codebase (#3445)
2. Set a size limit for `cupy.cuda.cub.cub_reduction()` until we have a better way to address #3309. 
3. `cdef`'d a few internal functions

PR #3740, opened 2020-08-07
Fix `can_use_device_segmented_reduce()` for incompatible axes

### LINKED ISSUE #3738: AssertionError occurs for some reductions when CUB is enabled
I get some test suite failures in `cupyimg` when `CUPY_ACCELERATORS="cub,cutensor"` environment variable is set.

CuPy Version          : 8.0.0b5
CUDA Root             : /usr/local/cuda-10.2
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSPARSE Version      : 10301
Thrust Version        : 100907
CUB Build Version     : 100800
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2604
NCCL Runtime Version  : 2604
cuTENSOR Version      : 10001

* Code to reproduce

```Python
import cupy as cp
x = cp.arange(2*4*3).reshape((2, 4, 3), order='F')
x.max(-1)  # or x.max(1), but not for x.max(0) which is the contiguous axis
```

This gives the correct result, but prints:
```
AssertionError

Exception ignored in: 'cupy.cuda.cub._preprocess_array'
AssertionError
```

leofang:
Close #3738.

There are two errors in this function:
1. #3562 mistakenly assumed that it's safe to call `_preprocess_array()` right away, but that function assumes `reduce_axis` and `out_axis` being legitimate. 
2. If an exception is raised (in this case, `_preprocess_array()` raised an unexpected `AssertionError`), we should propagate it.

cc: @grlee77

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -92,7 +92,7 @@ cdef tuple _get_output_shape(ndarray arr, tuple out_axis, bint keepdims):
 
 
 cpdef Py_ssize_t _preprocess_array(tuple arr_shape, tuple reduce_axis,
-                                   tuple out_axis, str order):
+                                   tuple out_axis, str order) except -1:
     '''
     This function more or less follows the logic of _get_permuted_args() in
     reduction.pxi. The input array arr is C- or F- contiguous along axis.
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -395,12 +395,13 @@ cdef bint can_use_device_segmented_reduce(
         dtype=None, str order='C'):
     if not _cub_reduce_dtype_compatible(x.dtype, op, dtype):
         return False
+    if not _cub_device_segmented_reduce_axis_compatible(
+            reduce_axis, x.ndim, order):
+        return False
+    # until we resolve cupy/cupy#3309
     cdef Py_ssize_t contiguous_size = _preprocess_array(
         x.shape, reduce_axis, out_axis, order)
-    return (_cub_device_segmented_reduce_axis_compatible(
-        reduce_axis, x.ndim, order) and
-        # until we resolve cupy/cupy#3309
-        contiguous_size <= 0x7fffffff)
+    return contiguous_size <= 0x7fffffff
 
 
 cdef _cub_support_dtype(bint sum_mode, int dev_id):


leofang:
Jenkins, test this please

grlee77:
Thanks. I am taking a look. 

I noticed the docstring to `_preprocess_array` has reference to some outdated file/function names if you want to clean that up while you are modifying this.

grlee77:
It seems to me that within `_preprocess_array`, the axis permutes logic should be:

```Python
    if order == 'C':
        axis_permutes = out_axis + reduce_axis
    elif order == 'F':
        axis_permutes = reduce_axis[::-1] + out_axis[::-1]
```

or 

```Python
    axis_permutes = out_axis + reduce_axis
    if order == 'F':
        axis_permutes = axis_permutes[::-1]
```

leofang:
> Thanks. I am taking a look.

Cool, thanks!

> I noticed the docstring to `_preprocess_array` has reference to some outdated file/function names if you want to clean that up while you are modifying this.

Will do. I think it's simple enough that we can just remove the docstring...At least the reference part.

> It seems to me that within `_preprocess_array`, the axis permutes logic should be:

`reduce_axis` and `out_axis` are returned by `_get_axis()`, and I think they're sorted there, so shouldn't need to reverse.

I can't think straight as it's late now...Will revisit this tomorrow and understand the CI errors.

grlee77:
> reduce_axis and out_axis are returned by _get_axis(), and I think they're sorted there, so shouldn't need to reverse.

> I can't think straight as it's late now...Will revisit this tomorrow and understand the CI errors.

Yeah, nevermind that previous suggestion. I think it was right as is.

grlee77:
Okay, I see what the CI issue is now. It is actually two-fold:
1.) logic in the test cases to determine `o_axis` is only correct when axes in `r_axis` are non-negative.
2.) `_can_use_cub_block_reduction` also will only sort the axes correctly if they are nonnegative. Under normal usage (i.e. not during these tests), the call to `_get_axis` in `_AbstractReductionKernel` takes the modulus of the axis so that negative ones never get passed to this function. The test cases, would have to do that on their own, though.

see suggested fix in leofang/cupy#11

grlee77:
It looks like `_get_axis` always returns sorted tuples, so I think the tuple and sorted calls could be removed from the following line to reduce overhead a bit:
https://github.com/cupy/cupy/blob/19073f6b9fb54ca7e45d033b92b1ab5e993605a1/cupy/core/_cub_reduction.pyx#L265-L268

asi1024:
@leofang Could you split the code change of `for_contiguous_axes` into another PR? This change seems to affect tests other than cub tests.

leofang:
@asi1024 Sounds good. I will split the test helper change along with @grlee77's suggestions (thanks!) to another PR.

leofang:
Jenkins, test this please

leofang:
> I will split the test helper change along with @grlee77's suggestions (thanks!) to another PR.

This is in #3745.

leofang:
Jenkins, test this please.

asi1024:
LGTM.

PR #3745, opened 2020-08-07
Extend test helper coverage; simplify CUB axis test

leofang:
Split from #3740. This PR does two things:
1. Cover more valid contiguous axes for the tests: now negative axes are also tested. 
2. Improve `_can_use_cub_block_reduction()`: the tuples `out_axis` and `reduce_axis` there are returned by `_get_axis()`, which makes the axes always sorted and positive, so we can simply join them as we do in `cupy.cuda.cub._preprocess_array()`. But, this also means that for its tests we need to convert the negative axes to positive ones. Thanks @grlee77 for suggestions!

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -1077,8 +1077,10 @@ def for_contiguous_axes(name='axis'):
     .. note::
         1. Adapted from tests/cupy_tests/fft_tests/test_fft.py.
         2. Example: for ``shape = (1, 2, 3)``, the tested axes are
-            ``[(2,), (1, 2), (0, 1, 2)]`` for the C order, and
-            ``[(0,), (0, 1), (0, 1, 2)]`` for the F order.
+            ``[(0, 1, 2), (-3, -2, -1), (1, 2), (-2, -1), (2,), (-1,)]`` for
+            the C order, and
+            ``[(0,), (-3,), (0, 1), (-3, -2), (0, 1, 2), (-3, -2, -1)]`` for
+            the F order.
     '''
     def decorator(impl):
         @functools.wraps(impl)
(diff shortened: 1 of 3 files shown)


grlee77:
The two files aside from the CUB tests where this `for_contiguous_axes` decorator is used are:
`cupy_tests/math_tests/test_sumprod.py`
`cupy_tests/core_tests/test_ndarray_reduction.py`

Those both pass without modification when I tested locally.

leofang:
Jenkins, test this please.

leofang:
Jenkins, test this please

cupy/testing/helper.py
```diff
@@ -1098,6 +1100,9 @@ def test_func(self, *args, **kw):
                 try:
                     kw[name] = a
                     impl(self, *args, **kw)
+                    a = tuple(i - ndim for i in a)
```
asi1024, line 1103: Conventionally we write as follows because building tuple from a list is faster than from a generator.
```suggestion
                    a = tuple([i - ndim for i in a])
```

leofang:
Jenkins, test this please

asi1024:
The decorator `for_contiguous_axes` seems for a low-level checks, but negative axes are removed by normalizing at preprocessing. Why not manually convert axes as `axis = [a - ndim for a in axis]` in each test?

leofang:
Hi @asi1024 there are many tests using this decorator and I hope to test them all. Like you said, in many cases it should be normalized during preprocessing, and adding negative axes to the decorator will ensure the normalization happens as expected.

asi1024:
We don't want to mix parameters for interface-level checks in to that low-level checks.

asi1024:
Do you mean that `for_contiguous_axes` is only for CUB tests?

leofang:
Currently yes!

asi1024:
@leofang Sorry for my late response.
If this PR is only for the CUB test, could you make a fix in the CUB test code?

leofang:
Sorry for dropping the ball, @asi1024. What fix did you have in mind?

asi1024:
Sorry for my late response.
Could you add `negative_axis` parameter to `testing.product` in `TestSimpleCubReductionKernelContiguity`?

leofang:
Closing it as the main fix to `_can_use_cub_block_reduction` is in. The test improvement no longer seems important to me, but I am happy to reopen if there's a need.

PR #3750, opened 2020-08-10
Adding a workaround for even-length inputs to scipy.signal.sepfir2d

coderforlife:
test_bsplines.py has been failing randomly due to a bug in scipy where out-of-bounds memory is being read under certain circumstance.

A bug report has been submitted to scipy: https://github.com/scipy/scipy/issues/12691https://github.com/scipy/scipy/issues/12691

See https://github.com/cupy/cupy/pull/3645#issuecomment-671149939 for some other details.

### DIFF
--- a/cupyx/scipy/signal/bsplines.py
+++ b/cupyx/scipy/signal/bsplines.py
@@ -10,6 +10,8 @@ def sepfir2d(input, hrow, hcol):
     assumed. This function can be used to find an image given its B-spline
     representation.
 
+    The arguments `hrow` and `hcol` must be 1-dimensional and of off length.
+
     Args:
         input (cupy.ndarray): The input signal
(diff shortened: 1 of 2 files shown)


leofang:
Could it be that SciPy pads the array somewhere to make it of odd length? On the one hand it's nice that you already made even length work and we could probably just keep it while waiting for SciPy to fix, on the other hand since we can't test against SciPy for even lengths it's indeed nerve wracking...

leofang:
Let me kick off the CI while waiting for the core devs' opinions...
Jenkins, test this please

coderforlife:
No, they do not pad, they read random memory.

takagi:
I think it is okay to make it accept odd-length `hrow` and `hcol`.

takagi:
LGTM!

PR #3751, opened 2020-08-10
Use single-quotes instead of double-quotes

asi1024:
https://github.com/cupy/cupy/wiki/Development-policies#general-coding-standards


emcastillo:
Jenkins, test this please

kmaehashi:
We discussed this and agreed fixing it incrementally rather than doing it in one PR.

Issue #3757, opened 2020-08-11
CUDA 11 Test: `cupy_tests.fft_tests.test_fft.TestMultiGpu`


kmaehashi:
```
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_2_{n=None, norm=None, shape=(64,)}::test_fft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_2_{n=None, norm=None, shape=(64,)}::test_ifft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_6_{n=None, norm='ortho', shape=(64,)}::test_fft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_6_{n=None, norm='ortho', shape=(64,)}::test_ifft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_24_{n=64, norm=None, shape=(0,)}::test_fft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_24_{n=64, norm=None, shape=(0,)}::test_ifft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_26_{n=64, norm=None, shape=(64,)}::test_fft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_26_{n=64, norm=None, shape=(64,)}::test_ifft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_28_{n=64, norm='ortho', shape=(0,)}::test_fft - AssertionError: Parameterized test failed.
FAILED tests/cupy_tests/fft_tests/test_fft.py::TestMultiGpuFft_param_28_{n=64, norm='ortho', shape=(0,)}::test_ifft - AssertionError: Parameterized test failed.
[...]

PR #3766, opened 2020-08-11
Proposal: Support transforming NumPy arrays with multi-GPU `Plan1d`

### LINKED ISSUE #1429: How to leverage memory of multiple GPUs (with NCCL) using unified memory
Using `set_allocator(MemoryPool(malloc_managed).malloc)`, the "unified memory" seems to allocate/use CPU and (one) GPU memory, not memory of multiple GPUs available.

How can I use memory of multiple GPUs available on the system, possibly with NCCL?
I have tested on a system with four P100 GPUs.

My use-case is - I have large 3D arrays that don't fit into single GPU memory, and I'd like to do some operations (e.g., FFT/IFFT) on these arrays. For example,
```python
import cupy as cp

...

# A and B are large 3D numpy arrays, e.g., 4096x4096x100
arrayA = cp.array(A)
arrayB = cp.array(B)

output = cp.fft.ifftn(cp.multiply(cp.fft.fft(arrayA), arrayB)).real
```

### LINKED ISSUE #2742: Discussion on multi-GPU support for cuFFT and cuSOLVER
Original context: See #2644.

Some of CUDA libraries like cuFFT and cuSOLVER provide native multi-GPU support. However, there is no guarantee that multiple GPUs will outperform single GPU; on the contrary, for cuFFT I found in https://github.com/cupy/cupy/pull/2644#issuecomment-558835428 that any workload that can be handled in one GPU should ***not*** be processed using multi GPUs; the extra data transfer just kills the performance (even if NVLinks are present). I suspect the same holds for cuSOLVER.

Since CuPy operations incurs additional memory overhead, I think it is clear that multi-GPU is useful only when the workload is too large to fit in one GPU. For such excessive workloads, it's better/unavoidable to start from NumPy arrays directly, such as the "NumPy-in, NumPy-out" feature currently implemented in #2644. **This thread is for discussions regarding this feature and other possible alternative strategies.**

leofang:
Follow-up of #2644. Close #1429, close #2742, and address https://github.com/cupy/cupy/issues/2742#issuecomment-608951598.

This PR supports `Plan1d.fft(a, b, direction)` with `a` and `b` being NumPy arrays. The main purpose is to transform a massive dataset that single GPUs cannot hold (which to me is one of the biggest motivations to do multi-GPU FFTs). 

No real code is added in this PR, only tests --- all of the work was done in #2644, but the NumPy support at both low- and high- levels were commented out in 3e2b7d0 and 705fe1a, respectively, and this PR restores the former so that advanced users can allocate and use `Plan1d` if needed. **This PR does not affect any high-level APIs** such as `cupy.fft.fft()`.

I also made a few variables cdef'd, but this is just a very minor code fix.

### DIFF
--- a/cupy/cuda/cufft.pyx
+++ b/cupy/cuda/cufft.pyx
@@ -142,13 +142,13 @@ class CuFFTError(RuntimeError):
 
 
 @cython.profile(False)
-cpdef inline check_result(int result):
+cpdef inline void check_result(int result) except *:
     if result != 0:
         raise CuFFTError(result)
 
 
 cpdef size_t getVersion() except? -1:
-    cdef int version
+    cdef int version, result
     result = cufftGetVersion(&version)
     check_result(result)
     return version
@@ -251,6 +251,7 @@ class Plan1d(object):
                  devices=None, out=None):
         cdef Handle plan
         cdef bint use_multi_gpus = 0 if devices is None else 1
+        cdef int result
 
         self.plan = 0
         self.xtArr = <intptr_t>0  # pointer to metadata for multi-GPU buffer
@@ -466,8 +467,10 @@ class Plan1d(object):
             execZ2Z(plan, a.data.ptr, out.data.ptr, direction)
         elif self.fft_type == CUFFT_D2Z:
             execD2Z(plan, a.data.ptr, out.data.ptr)
-        else:
+        elif self.fft_type == CUFFT_Z2D:
             execZ2D(plan, a.data.ptr, out.data.ptr)
+        else:
+            raise ValueError
 
     def _multi_gpu_setup_buffer(self, a):
         cdef XtArrayDesc* xtArr_desc
@@ -482,9 +485,7 @@ class Plan1d(object):
         # uncessary memory usage. Note that these buffers are used for in-place
         # transforms, and are re-used (lifetime tied to the plan).
 
-        # TODO(leofang): revisit this when NumPy support is added
-        # if isinstance(a, cupy.ndarray) or isinstance(a, numpy.ndarray):
-        if isinstance(a, cupy.ndarray):
+        if isinstance(a, cupy.ndarray) or isinstance(a, numpy.ndarray):
             if self.xtArr == 0 and self.xtArr_buffer is None:
                 nGPUs = len(self.gpus)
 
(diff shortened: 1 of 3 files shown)


leofang:
Jenkins, test this please

leofang:
08cbd2d is added to follow #3775 and circumvent the CUDA 11 bug, making this PR robust against CUDA 11.

@emcastillo @kmaehashi I now think this PR should be included in v8 for two reasons:
1. This usage follows the "simpleCUFFT_MGPU" demo from the CUDA samples
2. This helps diagnose the bug (#3757)

emcastillo:
Offline discussion with @leofang 

> Users need to literally call Plan1d.fft(a, b, dir) , not any high-level api, so it won’t affect the 1:1 mapping or __array_function__  dispatch

We are not modifying `cupy.fft` to accept numpy arrays, but the plan itself.

Pitch to allow this change:

> Think about where this massive data could come from? for us it’s usually experimental data (images, etc) loaded from disk, and another library such as h5py already returns a numpy array
so you’re suggesting to do another copy from host (numpy) to host/device (managed memory), and then do FFT
If the data is big enough, we are bounded by host memory and we may not even be able to do a copy.


Personally, as here the user needs to explicitly call fft over the plan, I think allowing host memory buffers here (or numpy arrays) makes sense as the plan is not part of numpy or cupy.
@asi1024 what do you think?

leofang:
a9e5f50 is an extra safety taken to prevent numpy arrays from being passed to high-level functions (`cupy.fft.fft` etc). For most cases, it also helps error out early (otherwise an exception is not thrown until the plan is actually executed).

leofang:
Jenkins, test this please

leofang:
Oops, sorry, made the entrypoint wrong. Should be fixed now.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
RTD seems misbehaving, let me power-cycle to reboot it.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

emcastillo (approved):

emcastillo:
Jenkins, test this please

leofang:
Thank you, @emcastillo!

PR #3776, opened 2020-08-13
Rename `cupy.cudnn` submodule to `cupy._cudnn`

takagi:
Rel #3420.

This PR renames `cupy.cudnn` submodule to `cupy._cudnn` to indicate it is CuPy's private submodule and not intended to be used by users.


takagi:
Close as discussed off line.

PR #3703, opened 2020-08-01
Add `cupy.roots` for Hermitian or symmetric matrix

Dahlia-Chehata:

PR #3780, opened 2020-08-13
Add `cupy.polydiv`

Dahlia-Chehata:
This PR is an implementation of `numpy.polydiv`. 
1- I don't think adding a workaround to support `bool` inputs is necessary since polynomials are supposed to be numerical values.  I followed what was agreed on here https://github.com/cupy/cupy/pull/3703#discussion_r464436905. So I needed to add `no bool=True` in all polyarithmetic tests to avoid tests separation. Nonetheless, some `polydiv` specific tests had to be separated. LMK if this needs to be changed.
2- The current behavior of `numpy.polydiv` allows zero division. I didn't support it in this implementation and raised `ZeroDivisionError()`. If this is not OK, I can raise `NotImplementedError` instead.
3- `numpy.polydiv` has some bug which prevents promotion to `poly1d` when necessary. So I was not able to test related cases. I added a TODO line with the issue link https://github.com/numpy/numpy/issues/17076
4- I don't think the loop can be avoided. So is there a way to parallelize it?


cupy/lib/_routines_poly.py
```diff
@@ -103,6 +103,66 @@ def polymul(a1, a2):
...
+            u[i: i + len2 - 1] -= v * u[i - 1]
+
+        quot = u[: dlen + 1] / scale
+        rem = u[dlen + 1:]
+
+    mask = cupy.allclose(rem, 0, rtol=1e-05, atol=1e-05)
+    rem[mask] = 0
+    rem = cupy.trim_zeros(rem, trim='f')
+    if rem.size == 0:
+        rem = cupy.array([0.])
```
Dahlia-Chehata, line 161: L157 - L161 can be replaced with #3793

Dahlia-Chehata:
@asi1024 PTAL

asi1024:
In typical cases, the size of `u` is much larger than `v`, so I don't want a loop of `u.size - v.size` size. How about calculating the results from `roots(v)` in such cases?

Dahlia-Chehata:
@asi1024 Could you elaborate a bit? Do you mean a synthetic division for each monic polynomial in `cupy.roots(v)`?

I have thought of a division based on the fft multiplication of the dividend with the divisor's reciprocal. I can try it out if you think it's a good idea.

asi1024:
@Dahlia-Chehata For `v.size == 2` case, the return value is calculated by using `polyval` without for-loop.

For `v.size > 2` case, the return value is calculated with a for-loop of `v.degree` iteration by factorizing `v` by using `cupy.roots(v)` and calling `polyval` `v.degree` times.
However, currently `cupy.roots` requires symmetric input for `v.degree >= 2` case, so it's OK to just left a TODO comment.

asi1024:
@Dahlia-Chehata Can I takeover the PR?

asi1024:
Closing the PR as it has been WIP for a long time.
Please feel free to reopen if you want to continue.

PR #3451, opened 2020-06-18
Fix `TypeError` in parameterize test catching `CUDADriverError`

asi1024:
Fix #3308.

PR #3785, opened 2020-08-14
Define base class of CUDA errors

asi1024:
Reverts #3451.


kmaehashi:
I think the current assumption in `cupy.testing.parameterze` that exceptions are always constructable with one string argument is wrong (e.g., `scipy.sparse.linalg.ArpackNoConvergence`) and should be fixed first.

I understand the motivation and the design proposed in this PR is cleaner, but I feel the complexity (inheritance) added by this fix is somewhat cost-ineffective.

kmaehashi:
Discussion moved to #3915

PR #3807, opened 2020-08-17
Fix a broken document

asi1024:
https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.linalg.solve_triangular.html


kmaehashi:
Checked HTML preview and looks good to me.
pfnCI, test this please.

PR #3811, opened 2020-08-18
Add some of `cupy.polynomial.chebyshev` arithmetic

Dahlia-Chehata:
This PR adds the following arithmetic interfaces of the `chebyshev` series:
`chebadd`, `chebsub`, `chebmul`, `chebmulx`, `chebpow`
The tests are intertwined so I think no need for separation.


cupy/polynomial/polyutils.py
```diff
@@ -75,3 +75,24 @@ def as_series(alist, trim=True):
     dtype = cupy.common_type(*arrays)
     ret = [a.astype(dtype, copy=False) for a in arrays]
     return ret
+
+
+def _add(c1, c2):
```
asi1024, line 80: Why is this function placed in this file? This function is used only from `chebyshev.py`?

asi1024: Anyway, this implementation looks duplicated with polyadd.

Dahlia-Chehata: > Why is this function placed in this file? This function is used only from `chebyshev.py`?

`Hermite`, `HermiteE`, `Laguerre`, `Legendre` and `Power` series uses the same `_add` and `_sub` 

> this implementation looks duplicated with polyadd.

Yes but with the reverse order of the Poly1d convention, instead of reversing twice using `polyadd/sub`

Dahlia-Chehata:
@asi1024 PTAL

cupy/polynomial/polyutils.py
```diff
@@ -108,3 +108,24 @@ def trimcoef(c, tol=0):
...
         return cupy.zeros_like(c[:1])
     return c[: ind]
+
+
+def _add(c1, c2):
+    """Helper function used in ``<type>add`` functions.
+    """
+    if c1.size > c2.size:
+        c1[:c2.size] += c2
+        return c1
```
asi1024, line 118: The input array should not be overwritten.

Dahlia-Chehata: The inputs won't be overwritten because of `_wraps_chebroutine` wrapper (which is called before these internal operations are called). I think all polynomial series (power, chebyshev...) will need their own preprocessing before calling `_add`/`_sub` so probably no overwriting will occur when reusing them for other series as well, right?

asi1024: So, do you mean that the wrapper copies the inputs before calling this function? If so, in which line of your code?

cupy/polynomial/polyutils.py
asi1024: How about reversing the inputs before the function call?

asi1024:
Closing the PR as it has been WIP for a long time.
Please feel free to reopen if you want to continue.

PR #3816, opened 2020-08-19
Move binary-op kernel definitions to _routines_logic.pyx

asi1024:


cupy/core/_routines_binary.pyx
```diff
@@ -0,0 +1,96 @@
+from cupy.core._kernel import create_ufunc
+from cupy.core._reduction import create_reduction_func
+
+from cupy.core.core cimport ndarray
```
takagi, line 4: These imports are needed?

takagi:
pfnCI, test this please.

PR #3828, opened 2020-08-21
Adding cupyx.scipy.signal.fftconvolve

coderforlife:
This adds `cupyx.scipy.signal.fftconvolve` along with support for `method='fft'` in `cupyx.scipy.signal.convolve` and `cupyx.scipy.signal.correlate`. This function fully supports n-dimensional data.

This PR does not include adjustments to `choose_conv_method` or `_fftconv_faster` to intelligently choose between the two methods, that will be done in a future PR.


cupyx/scipy/signal/_signaltools_core.py
```diff
@@ -95,3 +98,87 @@ def _inputs_swap_needed(mode, shape1, shape2, axes=None):
...
+    # See documentation in scipy.fft._helper._init_nd_shape_and_axes
+    # except shape argument is always None and doesn't return new shape
+    if axes is None:
+        axes = range(x.ndim)
+        shape = x.shape
+    else:
+        # Get the list of axes
+        if isinstance(axes, numbers.Number):
+            axes = (axes,)
+        axes = [cupy.util._normalize_axis_index(a, x.ndim) for a in axes]
```
asi1024, line 136: Why not use `cupy._util._normalize_axis_indices` isntead?

cupyx/scipy/signal/_signaltools_core.py
```diff
@@ -95,3 +98,87 @@ def _inputs_swap_needed(mode, shape1, shape2, axes=None):
...
+    axes = _init_nd_and_axes(in1, axes)
+    # Length-1 axes can rely on broadcasting rules, no fft needed
+    axes = [a for a in axes if s1[a] != 1 and s2[a] != 1]
+    if sorted_axes:
+        axes.sort()
+
+    # Check that unused axes are either 1 (broadcast) or the same length
+    if not all(s1[a] == s2[a] or s1[a] == 1 or s2[a] == 1
+               for a in range(in1.ndim) if a not in axes):
+        raise ValueError("incompatible shapes for in1 and in2:"
```
asi1024, line 115: Please use single quotes.

cupyx/scipy/signal/_signaltools_core.py
```diff
@@ -95,3 +98,87 @@ def _inputs_swap_needed(mode, shape1, shape2, axes=None):
...
+        axes = [cupy.util._normalize_axis_index(a, x.ndim) for a in axes]
+
+        # Check the axes
+        if not len(axes):
+            raise ValueError("when provided, axes cannot be empty")
+        if len(set(axes)) != len(axes):
+            raise ValueError("all axes must be unique")
+
+        # Get the resulting shape
+        shape = (x.shape[a] for a in axes)
```
asi1024, line 145: Please convert generator to a tuple.
```suggestion
        shape = tuple([x.shape[a] for a in axes])
```

tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
```diff
@@ -55,6 +55,59 @@ def test_correlate_float16(self, xp, scp, dtype=cupy.float16):
         return self._filter('correlate', dtype, xp, scp)
 
 
+@testing.parameterize(*testing.product({
+    'in1': [(10,), (5, 10), (10, 3), (3, 4, 10)],
```
asi1024, line 59: How about renaming to `size1`? In `_filter` method, `self.in1` is an integer that represents the input size, and `in1` is `xp.ndarray`.

cupyx/scipy/signal/_signaltools_core.py
```diff
@@ -95,3 +98,87 @@ def _inputs_swap_needed(mode, shape1, shape2, axes=None):
...
+
+    # Check the resulting shape
+    if any(s < 1 for s in shape):
+        raise ValueError(
+            "invalid number of data points ({0}) specified".format(shape))
+
+    return axes
+
+
+def _freq_domain_conv(in1, in2, axes, shape, calc_fast_len=False):
```
asi1024, line 155: `calc_fast_len` is always `True`.

cupyx/scipy/signal/_signaltools_core.py
coderforlife: I was unaware of its existence, will add soon.

cupyx/scipy/signal/_signaltools_core.py
coderforlife: However, I have also written up `oaconvolve` this does pass `calc_fast_len` as `False`. I could remove it for now just to add it back when `oaconvolve` is added.

cupyx/scipy/signal/_signaltools_core.py
coderforlife: It turns out `_normalize_axis_indices` always sorts the axes but here the axes can only be sorted when used by `oaconvolve`.

cupyx/scipy/signal/_signaltools_core.py
```diff
@@ -92,6 +95,81 @@ def _inputs_swap_needed(mode, shape1, shape2, axes=None):
...
+        in1, in2 = in2, in1
+
+    return in1, in2, axes
+
+
+def _init_nd_and_axes(x, axes):
+    # See documentation in scipy.fft._helper._init_nd_shape_and_axes
+    # except shape argument is always None and doesn't return new shape
+    if axes is None:
+        axes = range(x.ndim)
```
asi1024, line 130:
```suggestion
        axes = list(range(x.ndim))
```

cupyx/scipy/signal/_signaltools_core.py
```diff
@@ -92,6 +95,81 @@ def _inputs_swap_needed(mode, shape1, shape2, axes=None):
...
+    else:
+        # Get the list of axes
+        if isinstance(axes, numbers.Number):
```
asi1024, line 133: The error message "`axes`  is not iterable" will be shown for a non-number object, but I feel it is confusing.
```suggestion
        if not isinstance(axes, collections.abc.Iterable):
```

cupyx/scipy/signal/_signaltools_core.py
```diff
@@ -92,6 +95,81 @@ def _inputs_swap_needed(mode, shape1, shape2, axes=None):
...
+    # Length-1 axes can rely on broadcasting rules, no fft needed
+    axes = [a for a in axes if s1[a] != 1 and s2[a] != 1]
+    if sorted_axes:
+        axes.sort()
+
+    # Check that unused axes are either 1 (broadcast) or the same length
+    if not all(s1[a] == s2[a] or s1[a] == 1 or s2[a] == 1
+               for a in range(in1.ndim) if a not in axes):
+        raise ValueError('incompatible shapes for in1 and in2:'
+                         ' {} and {}'.format(s1, s2))
```
asi1024, line 116: How about writing as follows for the readability?
```py
    for ax, (dim1, dim2) in enumerate(zip(in1, in2)):
        if ax in axes and not (dim1 == dim2 or dim1 == 1 or dim2 == 1):
            raise ValueError(...)
```

cupyx/scipy/signal/_signaltools_core.py
asi1024: How about adding `sort_axes` option to `_normalize_axis_indices`?

coderforlife: Will do!

asi1024:
Jenkins, test this please.

asi1024:
@coderforlife Could you check the Jenkins failure?

coderforlife:
It looks like there are 2 tests that fail, both when using `float32` and the actual and desired outputs are just barely missing the tolerances. I will relax the tolerances for `float32`s for this test.

asi1024:
@coderforlife Any updates?

asi1024:
Jenkins, test this please.

coderforlife:
I forgot which values were off by and by how much and Jenkins info was removed due to being too long ago. Was able to use the newest results and hopefully will pass this one last time. Thanks!

leofang:
> I forgot which values were off by and by how much and Jenkins info was removed due to being too long ago. Was able to use the newest results and hopefully will pass this one last time. Thanks!

Hi @coderforlife, in case you don't know already, or if needed, it is now possible to set tolerance per dtype: https://github.com/cupy/cupy/pull/4269.

coderforlife:
I should the reset the tolerances so it is more strict for the float64, don't run the code through Jenkins yet, I will update soon.

coderforlife:
It should be better now! Can we run it on Jenkins (which needs different tolerances than on my machine in some cases so may fail even though it is working on my own machine). Thanks!

leofang:
Happy to help! (I haven't read the changes as I am not in a right condition.)

Jenkins, test this please

coderforlife:
One question about the next step when I add `oaconvolve`: in `cupy.core.internal._normalize_axis_indices` it only accepts `tuple`s as sequences. Is there any reason to not also accept `list`s and possibly any other iterables? All that would be require would be to change the `elif` condition on line 424 to something like `not isinstance(axes, (tuple, list))` or `not isinstance(axes, collections.abc.Iterable)` or possibly something a bit more complex like:

```python
else:
    try:
        iter(axes)
    except TypeError:
        axes = axes,
```

asi1024:
> Is there any reason to not also accept lists and possibly any other iterables?

It is because numpy does not support non-tuple axes. More specifically, for 2-dim `x` of `numpy.ndarray`, `numpy.sum(x, axis=[1, 2])` and `numpy.sum(x, axis=range(2))` raises `TypeError`.

asi1024:
LGTM!

coderforlife:
> It is because numpy does not support non-tuple axes. More specifically, for 2-dim `x` of `numpy.ndarray`, `numpy.sum(x, axis=[1, 2])` and `numpy.sum(x, axis=range(2))` raises `TypeError`.

Okay! Thanks, fixed the `oaconvolve` code itself to generate tuples for axes.

PR #3731, opened 2020-08-06
Implement library preloading for wheels

kmaehashi:
This implements #3724.

* Download the latest supported cuDNN by:

```
python -m cupyx.tools.install_library --library cudnn --cuda 11.0
```

Libraries are extracted to `~/.cupy/cuda_lib/11.0/cudnn/8.0.2/` (actual shared libraries are placed like `~/.cupy/cuda_lib/11.0/cudnn/8.0.2/lib64/libcudnn.so.8.0.2`)

* Warn when importing cuDNN if the package is installed via wheels (`cupy/_wheel.json` exists) and library could not be loaded.

This preloading is only performed in wheel packages (from PyPI).

TODO:
- [x] Test if it works with Windows
- [x] Add test for the download tool
- [ ] Support to add `cupy/_wheel.json` in cupy-release-tools (*will be in separate PR to cupy-release-tools*)

PR #3846, opened 2020-08-24
Fix test to run without cuDNN

kmaehashi:
Follow-up #3731.

The test fails to run in environment without cuDNN.


kmaehashi:
pfnCI, test this please.

emcastillo (approved):

PR #3848, opened 2020-08-24
Copy input by default for C2R transform

### LINKED ISSUE #3770: Make a copy for input when calling FFT on CUDA 10.1+
#3763 adds a warning that inputs may be modified (because cuFFT performs in-place operation), but it's better to make a copy for input before calling cuFFT.

* Make a copy by default.
* Provide an option to avoid copy, e.g., `cupy.fft(..., *, overwrite_x=False)` for users needing performance. (using keyword argument as this is not numpy-compatible)

This feature is not to be backported to v7.

kmaehashi:
Closes #3770

### DIFF
--- a/cupy/fft/fft.py
+++ b/cupy/fft/fft.py
@@ -86,6 +86,11 @@ def _exec_fft(a, direction, value_type, norm, axis, overwrite_x,
 
     if a.base is not None or not a.flags.c_contiguous:
         a = a.copy()
+    elif (value_type == 'C2R' and not overwrite_x and
+            10010 <= cupy.cuda.runtime.runtimeGetVersion()):
+        # The input array may be modified in CUDA 10.1 and above.
+        # See #3763 for the discussion.
+        a = a.copy()
 
     n = a.shape[-1]
     if n < 1:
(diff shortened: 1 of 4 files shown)


kmaehashi:
pfnCI, test this please.

leofang:
I think we also need to copy when C2R PlanNd is in use, just in case. We can do it in `_fftn` around here by checking `value_type`:
https://github.com/cupy/cupy/blob/3c4f481c2eca55045119bc2b96b3335deea2e6da/cupy/fft/fft.py#L509-L512
(`_exec_fftn` assumes the array is already C- or F- contiguous.)

toslunar:
> I think we also need to copy when C2R PlanNd is in use, just in case. We can do it in `_fftn` around here by checking `value_type`:
> 
> https://github.com/cupy/cupy/blob/3c4f481c2eca55045119bc2b96b3335deea2e6da/cupy/fft/fft.py#L509-L512
> 
> (`_exec_fftn` assumes the array is already C- or F- contiguous.)

I agree.

BTW, should we move the 4 lines to `_exec_fftn` for consistency with `_exec_fft`?

kmaehashi:
Thanks for comments! Will fix fftn as well.

Discussed offline, we think it's better to avoid adding numpy-incompatible `overwrite_a` option to NumPy fft. We suggest users who need such performance boost to use scipy interface instead.

leofang:
> We suggest users who need such performance boost to use scipy interface instead.

That's what I learned from @toslunar @asi1024 🙂

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

leofang:
Jenkins, test this please

toslunar (approved):

Issue #3821, opened 2020-08-20
Circular imports in `cupyx.scipy.signal.convolve`


Dahlia-Chehata:
I am unable to import `cupyx.scipy.signal.convolve` because cupyx.signal was removed in #3645.
I got the following output when I tried ``` from cupyx.scipy.signal import convolve```
```
File "/home/.../cupy/cupyx/scipy/ndimage/_filters_core.py", line 105, in <module>
    weights_dtype=cupy.float64, structure_dtype=cupy.float64):
AttributeError: module 'cupy' has no attribute 'float64'
```
Rel #3748, #3743
blocks #3547

asi1024:
Will be resolved in #3854.

PR #3854, opened 2020-08-25
Resolve circular imports between `cupy` and `cupyx.scipy`

asi1024:
Resolve https://github.com/cupy/cupy/issues/3821.
Current CuPy's codebase does not have any circular imports, but it occurs if we try importing `cupyx.scipy.***` from `cupy/***/***.py`.


leofang:
Jenkins, test this please

PR #3839, opened 2020-08-24
Deprecate `cupy.sparse` package

takagi:
Close #3773.

This PR raises a deprecation warning for `cupy.sparse` package as discussed in #3773.

PR #3856, opened 2020-08-25
Fix `cupy.sparse.*` deprecation

kmaehashi:
Follow-up #3839

### DIFF
--- a/cupy/sparse/__init__.py
+++ b/cupy/sparse/__init__.py
@@ -16,6 +16,8 @@ if (3, 7) <= sys.version_info:
             msg = 'cupy.sparse is deprecated. Use cupyx.scipy.sparse instead.'
             warnings.warn(msg, DeprecationWarning)
             return getattr(cupyx.scipy.sparse, name)
+        raise AttributeError(
+            "module 'cupyx.scipy.sparse' has no attribute '{}'".format(name))
 else:
     from cupyx.scipy.sparse.base import issparse  # NOQA
     from cupyx.scipy.sparse.base import isspmatrix  # NOQA
--- a/cupy/sparse/__init__.py
+++ b/cupy/sparse/__init__.py
@@ -19,46 +19,4 @@ if (3, 7) <= sys.version_info:
         raise AttributeError(
             "module 'cupyx.scipy.sparse' has no attribute '{}'".format(name))
 else:
-    from cupyx.scipy.sparse.base import issparse  # NOQA
-    from cupyx.scipy.sparse.base import isspmatrix  # NOQA
-    from cupyx.scipy.sparse.base import spmatrix  # NOQA
-    from cupyx.scipy.sparse.coo import coo_matrix  # NOQA
-    from cupyx.scipy.sparse.coo import isspmatrix_coo  # NOQA
-    from cupyx.scipy.sparse.csc import csc_matrix  # NOQA
-    from cupyx.scipy.sparse.csc import isspmatrix_csc  # NOQA
-    from cupyx.scipy.sparse.csr import csr_matrix  # NOQA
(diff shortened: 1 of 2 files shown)


kmaehashi:
pfnCI, test this please.

cupy/sparse/linalg/__init__.py
```diff
@@ -1,5 +1,18 @@
...
 
-# "NOQA" to suppress flake8 warning
-from cupyx.scipy.sparse.linalg.solve import lsqr  # NOQA
+import cupyx.scipy.sparse.linalg
+
+
+if (3, 7) <= sys.version_info:
+    def __getattr__(name):
+        if hasattr(cupyx.scipy.sparse.linalg, name):
+            warnings.warn(
```
kmaehashi, line 10: FYI: users will see this message (maybe only) when `from cupy.sparse.linalg import lsqr`.

cupy/sparse/__init__.py
```diff
@@ -16,47 +16,7 @@ def __getattr__(name):
             msg = 'cupy.sparse is deprecated. Use cupyx.scipy.sparse instead.'
             warnings.warn(msg, DeprecationWarning)
             return getattr(cupyx.scipy.sparse, name)
+        raise AttributeError(
+            "module 'cupyx.scipy.sparse' has no attribute '{}'".format(name))
```
toslunar, line 20: It could be done by
```python
try:
    attr = getattr(cupyx.scipy.sparse, name)
except AttributeError:
    raise
warnings.warn(msg, DeprecationWarning)
return attr
```

I suppose the intention is below for the better compatibility
```suggestion
            "module 'cupy.sparse' has no attribute '{}'".format(name))
```

kmaehashi: Hmm, I think we can't do both. If we just `raise` as your suggested, the error message says `module 'cupyx.scipy.sparse' has no attribute 'xyz'`

toslunar: And how about using `!r` format?

Combining my three suggestions,
```python
try:
    attr = getattr(cupyx.scipy.sparse, name)
except AttributeError:
    raise AttributeError("module 'cupy.sparse' has no attribute {!r}".format(name))
warnings.warn(msg, DeprecationWarning)
return attr
```

kmaehashi: Ah, ok your point was looking up attribute twice is costly.
I'll fix in that way.

kmaehashi:
```py
    def __getattr__(name):
        try:
            attr = getattr(cupyx.scipy.sparse, name)
        except AttributeError as e:
            raise AttributeError(
                "module 'cupy.sparse' has no attribute {!r}".format(
                    name)) from e
        warnings.warn(
            'cupy.sparse is deprecated. Use cupyx.scipy.sparse instead.',
            DeprecationWarning)
        return attr
```

Result:

```
>>> cupy.sparse.foo
  File "/home/kenichi/Development/cupy/cupy/sparse/__init__.py", line 16, in __getattr__
    attr = getattr(cupyx.scipy.sparse, name)
AttributeError: module 'cupyx.scipy.sparse' has no attribute 'foo'

The above exception was the direct cause of the following exception:

  File "/home/kenichi/Development/cupy/cupy/sparse/__init__.py", line 20, in __getattr__
    name)) from e
AttributeError: module 'cupy.sparse' has no attribute 'foo'
```

toslunar:
pfnCI, test this please.

toslunar (approved):

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

PR #3768, opened 2020-08-12
Show numpy and scipy versions in `show_config`

emcastillo:
Currently `cupy.show_config` only shows cuda related info.
In order to easily lookup for errors in CI and locally, printing numpy and scipy versions as well can be helpful.

Blocked by #3858 and #3877

PR #3858, opened 2020-08-25
Add `cudaGetDeviceProperties`

emcastillo:
This call was missing and it is pretty useful to get relevant info.

I didn't fill all the fields because there are too many and they change by CUDA release, so we have to be careful with this.
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g1bf9d625a931d657e08db2b4391170f0

Depends on #3877 

TODO: Check HIP support


leofang:
We use `cudaDeviceGetAttribute`. It's different?!

kmaehashi:
Its' different. By using `cudaGetDeviceProperties ` we can display human-readable device names in `cupy.show_config()`. (want to use this in #3768)

emcastillo:
The struct contents change on different cuda versions:

Cuda 9.0:
https://docs.nvidia.com/cuda/archive/9.0/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp
Cuda 10.x: (more fields)
https://docs.nvidia.com/cuda/archive/10.0/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp
Cuda 11.0: (Adds `reservedSharedMemPerBlock`)
https://docs.nvidia.com/cuda/archive/11.0/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp

leofang:
They didn't mess up the struct layout and simply added new fields, right? In that case, perhaps we should define a custom struct in `cupy_cuda_common.h` with CUDA-version-dependent fields, and typedef to it?

asi1024:
How about supporting only for the newest CUDA version for now?

emcastillo:
Indeed,
lets do it for 10.2 and 11

leofang:
> They didn't mess up the struct layout and simply appended new fields, right? 

I checked 10.0 and 11.0 and found it's not the case. They changed the memory layout of the struct (by inserting a new field `persistingL2CacheMaxSize` right after `l2CacheSize` in the middle). 

But, if the goal is to report a few common properties that exist in all versions, I think when you do `cdef extern from ...` you don't have to know the full struct layout and can just import what you need. I have a vague impression that Cython's doc mentioned this.

cupy_backends/cuda/api/runtime.pxd
```diff
@@ -100,6 +99,321 @@ cdef extern from *:
     ctypedef struct IpcEventHandle 'cudaIpcEventHandle_t':
         unsigned char[64] reserved
 
+    ctypedef struct cudaUUID 'cudaUUID_t':
+        char bytes[16]
+
+    IF CUDA_VERSION >= 11000:
+        # We can't use IF in the middle of structs declaration
```
emcastillo, line 106: We could use a single struct here, since Cython ignores the fields that are not present in the associated C one.
Also the order of the fields is not important, they don't have to match the C ones.

I decided to add all three structs to make this more explicit, but I can change it

emcastillo:
PTAL!

cupy_backends/cuda/api/runtime.pxd
asi1024: > We could use a single struct here, since Cython ignores the fields that are not present in the associated C one.
Also the order of the fields is not important, they don't have to match the C ones.

I didn't know that!

asi1024:
Jenkins, test this please.

emcastillo:
Jenkins, test this please

asi1024:
@emcastillo The cython compilation failure in cupy.py3.amd seems related to this PR.

emcastillo:
Jenkins, test this please

cupy_backends/cuda/api/runtime.pxd
leofang: > since Cython ignores the fields that are not present in the associated C one.

This is new to me too! I wonder how did they convince the compiler to pass...?

emcastillo: Probably it's just wrapping the structure in another one and just re-mapping the fields? I didn't check the generated code

asi1024:
LGTM.

PR #3861, opened 2020-08-25
overwrite intermediate arrays in FFT

toslunar:
- maintain `readonly` flag of the current array `a`. The flag is
  turned off after a copy or a non-overwriting step.
- `ascontiguousarray` is moved from `_fftn` to `_exec_fftn` for
  consistency with `_exec_fft`.
- internal `out` arg is removed because the previous implementation
  seems buggy if `overwrite_x and out is not None`.


leofang:
> * `ascontiguousarray` is moved from `_fftn` to `_exec_fftn` for
>   consistency with `_exec_fft`.

I guess this makes sense. IIRC `_exec_fftn` requires a contiguous input.

> * internal `out` arg is removed because the previous implementation
>   seems buggy if `overwrite_x and out is not None`.

It was kept for a possible future extension that users can pass in an existing buffer. But, given that 1. no one is working on this capability (at least I'm not, perhaps @grlee77 has a local branch for this?) and 2. it could be error-prone with certain (and likely more) cuFFT transforms that silently do things in place, I think it's fine to remove it.

The code seems to be fine but I only skimmed quickly. I'll try to do a more in-depth review next week if it's not urgent.

grlee77:
I am also fine with removing the internal `out` argument.

asi1024:
@toslunar How about checking if `a` shares the same memory space with the input array with `may_share_memory`, instead of `readonly` flag? I think it will reduce code diffs.

kmaehashi:

PR #3867, opened 2020-08-26
Expose submodules in `cupy.testing`

takagi:
Rel #3420.

This PR exposes `cupy.testing`'s  missing submodules. Based on this, I will properly hide all of its internal submodules as `cupy.testing.attr`.


takagi:
Let me close this PR as discussed offline that we don't need to expose these names.

Issue #3403, opened 2020-06-05
Import fails although CUDA is present


daniel1v:
In Python <3.8 it is possible to use the cupy-cudaXXX inside a conda environment in combination with cudatoolkit from anaconda repositories. The cuda binaries are installed into the conda environment and no CUDA_PATH has to be set.

Under Python 3.8 the following lines make the import fail, although CUDA is present:

https://github.com/cupy/cupy/blob/1654d38ca733e75f1bbae89dabb3d912d17ab4f7/cupy/__init__.py#L11-L12

Uncommenting these lines make the import work again and cupy works also as expected. So my question: Is it really necessary to enforce users to set the CUDA_PATH variable? And wouldn't it be better to explicitely check if the needed dll is present rather than the environment variable?

leofang:
Just a quick question before I lay out my thoughts: You don't have nvcc in your `PATH`, do you?

daniel1v:
@leofang Thank you for your reply. No, nvcc is not on my PATH, as it is not included in the cudatoolkit anaconda package.

kmaehashi:
This feature request makes sense as the current cupy works without installing `nvcc`. I think we can rewrite:

1. Call `_setup_win32_dll_directory`. Return false if CUDA path cannot be detected. (no error raised here)
2. Import cupy.core. If it fails and 1. returned false, show an error message.

I also think it is better to document is when CUDA development components (e.g., nvcc and CUDA header files, which are not a part of CUDA Runtime components) are required:
* When building cupy from source
* When using `nvcc` as a compiler backend
* When using rdc option

leofang:
Just to complement @kmaehashi since he's ahead of me 😂: In this case the failure is due to the  absence of both `CUDA_PATH` and `nvcc`, so `get_cuda_path()` returns `None` and in PY38 it errors out. Ideally, `get_cuda_path()` should return the path of CUDA libraries to which CuPy is currently linked. Unfortunately in Python there's no simply, reliable, and cross-platform solution to achieve this AFAIK, which answers (admittedly unsatisfactorily) your questions.

The workaround we took for conda-forge's cupy package is to set `CUDA_PATH` when a conda environment is activated. But we do not (yet) provide a Windows build...

leofang:
By the way,

> * When using rdc option

I think technically `cudadevrt` is part of CUDA runtime?

kmaehashi:
[...]

PR #3880, opened 2020-08-26
[WIP] Fix wheel not loading bundled cuDNN on Windows

kmaehashi:
After `cupy_backends` separation cuDNN bundled in wheels are not loaded on Windows.

The following description only applies for Windows. (Linux behavior is different).

Currently we are copying two DLLs in wheels:

- `nvToolsExt64_1.dll` (used from `cupy.cuda.nvtx`)
- `cudnn64_7.dll` (used from `cupy_backends.cuda.libs.cudnn`)

In wheels, DLLs are placed under `cupy/cuda/` directory. This is because DLLs are looked up from the same directory as the Python extension module requiring that DLL. This assumption was OK until `cupy_backends` separation.

This PR changes the path to place shared library to `cupy/.data/lib` (which is the same as Linux), and add that directory to `PATH` (for Python 3.7 or earlier) or DLL lookup directory `os.add_dll_directory` (for Python 3.8+).

TODO: I'm now testing the behavior on Windows.

### DIFF
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -135,14 +135,43 @@ def _get_cub_path():
 
 
 def _setup_win32_dll_directory():
-    # Setup DLL directory to load CUDA Toolkit libs on Windows & Python 3.8+.
-    if sys.platform.startswith('win32') and (3, 8) <= sys.version_info:
+    # Setup DLL directory to load CUDA Toolkit libs and shared libraries
+    # added during the build process.
+    if sys.platform.startswith('win32'):
+        # Path to the CUDA Toolkit binaries
         cuda_path = get_cuda_path()
-        if cuda_path is None:
-            raise RuntimeError('CUDA path could not be detected.')
-        cuda_bin_path = os.path.join(cuda_path, 'bin')
-        _log('Adding DLL search path: {}'.format(cuda_bin_path))
-        os.add_dll_directory(cuda_bin_path)
+        if cuda_path is not None:
+            cuda_bin_path = os.path.join(cuda_path, 'bin')
+        else:
+            cuda_bin_path = None
+            warnings.warn(
+                'CUDA path could not be detected.'
+                ' Set CUDA_PATH environment variable if CuPy fails to load.')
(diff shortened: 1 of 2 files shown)


cupy/_environment.py
```diff
@@ -135,14 +135,43 @@ def _get_cub_path():
...
-        if cuda_path is None:
-            raise RuntimeError('CUDA path could not be detected.')
-        cuda_bin_path = os.path.join(cuda_path, 'bin')
-        _log('Adding DLL search path: {}'.format(cuda_bin_path))
-        os.add_dll_directory(cuda_bin_path)
+        if cuda_path is not None:
+            cuda_bin_path = os.path.join(cuda_path, 'bin')
+        else:
+            cuda_bin_path = None
+            warnings.warn(
```
emcastillo, line 147: is it safe to change from error to warning?

cupy_setup_build.py
```diff
@@ -654,64 +654,44 @@ def get_long_description():
 def prepare_wheel_libs():
     """Prepare shared libraries and include files for wheels.
 
-    On Windows, DLLs will be placed under `cupy/cuda`.
-    On other platforms, shared libraries are placed under `cupy/.data/lib` and
-    RUNPATH will be set to this directory later.
+    Shared libraries are placed under `cupy/.data/lib` and
+    RUNPATH will be set to this directory later (Linux only).
```
emcastillo, line 658: Windows is the same path?

emcastillo:
Jenkins, test this please

cupy/_environment.py
kmaehashi: This mitigates the error anaconda users will face when Python 3.8+. https://github.com/cupy/cupy/issues/3403

cupy_setup_build.py
kmaehashi: Here I meant `RUNPATH will be set to this directory` is not applicable for Windows.

emcastillo:
Jenkins, test this please

PR #3732, opened 2020-08-06
Deprecate `*_enabled` flags under `cupy.cuda`

kmaehashi:
Currently when using `*_enabled` flags, users should write:

```py
import cupy
if cupy.cuda.cutensor_enabled:
    from cupy.cuda import cutensor as cuda_cutensor
    from cupy import cutensor

def foo():
    if cupy.cuda.cutensor_enabled:
        # do cutensor things
        ...
```

I think this is better:

```py
# This import is always safe.
# Dummy object is returned when cuTENSOR is unavailable.
from cupy.cuda import cutensor


def foo():
    if cupy.cuda.cutensor.available:
        # do cutensor things
        pass
```

This is related to #3731.
In this way, we can use `__getattr__` in dummy objects to warn users if the library is not installed.
(note: module-level `__getattr__` is another choice, but only available in Python 3.7 or later)

I would like to hear opinions before completing this PR.

PR #3883, opened 2020-08-26
Fix not to use cupy.cuda.* from cupy codebase

kmaehashi:
In #3732, `cupyx/runtime.py` was fixed to use `cupy.cuda.cudnn` but this was incorrect.
`cupyx/runtime.py` is automatically imported when doing `import cupy`, so preloading warnings for cuDNN will be displayed just by doing `import cupy` if cudnn is unavailable.

### DIFF
--- a/cupy/cuda/cudnn.py
+++ b/cupy/cuda/cudnn.py
@@ -1,3 +1,9 @@
+"""
+cuDNN Wrapper
+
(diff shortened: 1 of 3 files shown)


emcastillo (approved):

emcastillo:
Jenkins, test this please

kmaehashi:
pfnCI, test this please.

Issue #3609, opened 2020-07-14
Document that contiguity match is not guaranteed


kmaehashi:
Let's add it to this docs:

https://docs-cupy.chainer.org/en/latest/reference/difference.html

Current policy: We try to reproduce NumPy's contiguity behavior for the output array when it is "reasonably" possible (e.g., we won't do `cupy.asfortranarray` just to align the spec with NumPy).

leofang:
Is #2717 related?

kmaehashi:
Good catch!

kmaehashi:
Fixed via #7344

PR #3897, opened 2020-08-29
Check format of sparse matrix in `numpy_cupy_array_equal`

asi1024:


kmaehashi:
From offline discussion: In general, it is better to align the output class (csr, csc, coo, ...) of the returned value with SciPy. However, it should be considered carefully if there is a performance impact. Same policy as contiguity compatibility (#3609).

takagi:
pfnCI, test this please.

takagi:
LGTM! Waiting for CI finish.

PR #3916, opened 2020-09-01
Fix `cupy.testing.parameterize` to reraise error of AssertionError type

### LINKED ISSUE #3915: `cupy.testing.parameterze` does not work when Exception requiring arguments more than one
Currently `cupy.testing.parameterze` assumes that exceptions are always constructible with one string argument, but this is not always true (e.g., `scipy.sparse.linalg.ArpackNoConvergence`).

- Fix the `cupy.testing.parameterze` to work with exceptions whose constructor require more than one argument, OR
- Rewrite it as a wrapper for `pytest.mark.parameterize`

cc/ @asi1024 @toslunar

asi1024:
Fix #3915.

### DIFF
--- a/cupy/testing/parameterized.py
+++ b/cupy/testing/parameterized.py
@@ -6,8 +6,6 @@ import typing as tp  # NOQA
 import unittest
 
 from cupy.testing import _bundle
-from cupy.cuda import cufft
-from cupy.cuda import driver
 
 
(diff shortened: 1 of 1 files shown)


toslunar:
Thanks @asi1024 for the early fix.  Now #3915 is closed.

PR #2079, opened 2019-03-03
Allow copying in the format `cupy_array[:] = numpy_array`

pentschev:
The __setitem__ implementation from cupy.ndarray checks for an empty slice and
if the value being passed is an instance of numpy.ndarray to make a copy of it.
That can is a very useful feature in circumstances where we want to create a
copy of a numpy.ndarray without requiring mechanisms to identify the type of an
array as a CuPy array.

This resolves #593.

PR #3779, opened 2020-08-13
Rename `cupy.util` submodule to `cupy._util`

takagi:
Rel #3420.

This PR renames `cupy.util` submodule to `cupy._util` to indicate it is CuPy's private submodule and not intended to be used by users.

PR #3928, opened 2020-09-03
Handle transfer to cupy view

kalvdans:
When trying out host-to-device transfer with the syntax introduced in #2079, `cupy_array[:] = numpy_array`, I noticed that it gives the wrong result if cupy_array is a non-contiguous view of a contiguous cupy array. We use views to support pitched device memory.

This pull request makes it raise a ValueError instead. In future, we should check if the strides looks like pitched memory and use cudaMemcpy2D in that case.


cupy/core/core.pyx
```diff
@@ -1320,7 +1320,9 @@ cdef class ndarray:
                 and slices == slice(None, None, None)
                 and isinstance(value, numpy.ndarray)
         ):
-            if self.dtype == value.dtype and self.shape == value.shape:
+            if (self.dtype == value.dtype
+                    and self.shape == value.shape
+                    and (self.flags.f_contiguous or self.flags.c_contiguous)):
```
leofang, line 1325: I think this will disallow copies between two (possibly non-contiguous) views that have the same strides, which is likely a regression. Looks to me that the assumption in a few lines below is wrong; it should be:
```python
if self.strides == value.strides:
    ptr = ctypes.c_void_p(value.__array_interface__['data'][0])
elif self._f_contiguous or self._c_contiguous:  # note we should use the C flags for performance
    # set up ptr and friends
else:
    raise

# do copy
```

kalvdans: I don't see how you can copy into a non-contiguous memory area with a single cudaMemcpy().

leofang: That is a good point! Then I think we should not need to check the strides below, and always require the arrays being contiguous.

kalvdans: Allowing the source array to be non-contigous was an [explicit goal](https://github.com/cupy/cupy/pull/2079#issuecomment-486167028) in #2079, so I think we should continue to handle it by rearranging the source data on the CPU side before the cudaMemcpy.

emcastillo:
As a side note for the future, one of the things I am concerned is that cuda only offers pitched memory copies for 2 and 3d (I might be wrong here).
But the general Nd case will require us to do multiple memory copies.

kalvdans:
I think we should error out if it requires more than one call to cudaMemcpy. Then it might be more beneficial to do a device allocation + a linear host-to-device copy + unpacking on the device side by explicitly doing `cupy_array[:] = cp.asarray(numpy_array)`

kalvdans:
Read the Docs build failure seems unrelated to this pull request, reported at https://github.com/readthedocs/readthedocs.org/issues/7445

cupy/core/core.pyx
leofang: The source array can be non-contiguous as long as we do this rearrangement on CPU like you said, but the destination must be contiguous. Sorry for not being clear.

kalvdans: I could eliminate the strides test, and always do a `ravel()`. This is okay since `ravel()` will not do anything if the array is already in the desired form. I don't know if this is the transform you had in mind @leofang . See https://github.com/cupy/cupy/pull/3928/commits/477785846566ab864e4d9488ce77f967bfcc6406

cupy/core/core.pyx
```diff
@@ -1320,13 +1320,12 @@ cdef class ndarray:
...
-                if self.strides == value.strides:
-                    ptr = ctypes.c_void_p(value.__array_interface__['data'][0])
-                else:
-                    order = 'F' if self.flags.f_contiguous else 'C'
-                    tmp = value.ravel(order)
-                    ptr = ctypes.c_void_p(tmp.__array_interface__['data'][0])
+            if (self.dtype == value.dtype
+                    and self.shape == value.shape
+                    and (self.flags.f_contiguous or self.flags.c_contiguous)):
+                order = 'F' if self.flags.f_contiguous else 'C'
```
leofang, line 1326: Use CuPy's internal flags:
```suggestion
                    and (self._f_contiguous or self._c_contiguous)):
                order = 'F' if self._f_contiguous else 'C'
```

tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
```diff
@@ -180,11 +180,10 @@ def test_isinstance_numpy_copy(self, xp, dtype, order):
...
     def test_isinstance_numpy_copy_wrong_dtype(self):
-        for xp in (numpy, cupy):
-            a = numpy.arange(100, dtype=numpy.float64).reshape(10, 10)
-            b = cupy.empty(a.shape, dtype=numpy.int32)
-            with pytest.raises(ValueError):
-                b[:] = a
+        a = numpy.arange(100, dtype=numpy.float64).reshape(10, 10)
+        b = cupy.empty(a.shape, dtype=numpy.int32)
+        with pytest.raises(ValueError):
+            b[:] = a
```
leofang, line 186: Good catch for fixing this! I am not sure why the CI does not catch this...Perhaps we don't have `ENABLE_SLICE_COPY` set up?

leofang:
I think the RTD CI is down for a few days, so don't worry about it.

tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
kalvdans: This is no functional change; it was just that the same exact test were run twice in a row, since `xp` was not used inside the for loop.

cupy/core/core.pyx
kalvdans: What is the benefit of using internal flags? (First time I see the "Suggested change" feature of github, really cool!)

tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
leofang: Oh I misread the line `b = ...` 😅 Thanks!

cupy/core/core.pyx
leofang: Performance. It's faster to access the C attributes than the Python attributes, so for array metadata like contiguity and shape we use internal C attributes whenever possible in the Cython space.

kalvdans: Smart! Applied. Thanks for the tip and the review so far!

leofang (changes requested):
I think you need to merge with (or rebase onto) master to fix Travis's error. The module `util` was renamed to `_util` (#3779).

kalvdans:
> I think you need to merge with (or rebase onto) master to fix Travis's error. The module `util` was renamed to `_util` (#3779).

Thanks for the heads up. What I don't understand though, is how travis could have gotten that change in, when it was testing my branch. Let me try the travis commands locally.

kalvdans:
> What I don't understand though, is how travis could have gotten that change in, when it was testing my branch.

I found the cause now. Travis doesn't test my branch head but instead does `git fetch origin +refs/pull/3928/merge` which is a preview of what a merge with master would look like. I shall do the merge then :)

leofang:
Jenkins, test this please

emcastillo (approved):

PR #3932, opened 2020-09-03
Remove "min" from shown statistics of `cupyx.time.repeat`

asi1024:
We rarely refer to the minimum execution time.


leofang:
I personally found it useful.

kmaehashi:
I also think `min` is useful.
It's also recommended to see min in timeit docs (although the context is CPU here): https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat

PR #3948, opened 2020-09-04
Fix `_compressed_sparse_matrix._minor_slice` for step > 1 case

asi1024:
Fix `_minor_slice` and enhance test coverage.

Issue #3513, opened 2020-06-29
Support sparse matrices with more than 2^31-1 elements


leofang:
Currently there are a number of places in the codebase where the dtype for `indptr` and `indices` are hardcoded to `'i'` (i.e., `int32`), for example: 
https://github.com/cupy/cupy/blob/22e270f303603848584974d1fbee5a38c9a5c428/cupyx/scipy/sparse/compressed.py#L330-L331
But they shouldn't be. In SciPy, they use a helper function `get_index_dtype()` to determine this:
https://github.com/scipy/scipy/blob/f4b5605031f738bc87ae4e193d614d525b98ffba/scipy/sparse/compressed.py#L65-L71
I think this change shouldn't be hard, but it is worth a standalone PR with sufficient test coverage added. 

For the short term, I'd say let's encourage PR authors to take this into account, so we can make the changes gradually. In particular, when a new `ElementwiseKernel` or `ReductionKernel` is added, we just need to use template types `T` as oppose to hardcoded `int32`.

takagi:
It seems that the newer cuSPARSE Generic API supports 64bit indices (and the legacy API we currently use only supports 32bit indices).

https://docs.nvidia.com/cuda/cusparse/index.html#cusparse-generic-api-reference

takagi:
cuSPARSE Generic API is starting with CUDA 10.1.
https://docs.nvidia.com/cuda/archive/10.1/cusparse/index.html#cusparse-generic-api-reference

takagi:
CuPy already has cuSPARSE Generic API supports, #3129 and #3242, awesome @anaruse!

leofang:
That's right, so using 64-bit indices becomes possible!

leofang:
When working on #4778 I noticed one obstacle: In some index calculations we need `scatter_add` to handle `int64`, but it relies on CUDA's builtin `atomicAdd`, which does not have a `int64` specialization. Don't see an easy workaround ftm.

ref: https://github.com/cupy/cupy/pull/4778/commits/54227888221679ece62ec70273fcfdada281202d

kmaehashi:
I noticed docs say several cuSPARSE APIs will be removed in the next major release. e.g.

https://docs.nvidia.com/cuda/cusparse/index.html#csrgemm2

```
11.2. cusparse<t>csrgemm2() [DEPRECATED]
[[DEPRECATED]] use cusparseSpGEMM() instead. The routine will be removed in the next major release
```

[...]

PR #3949, opened 2020-09-04
Specify dtype of indptr and indices as int32 explicitly

asi1024:
`indptr` and `indices` are always int32 dtype.


leofang:
If this is not fixing any bug, I'd suggest to avoid it. We need to support large sparse matrices (#3513), so I've advised to leave the type generic in new PRs so that it's less work for whoever picking up this conversion challenge 😅

asi1024:
I see! Now I agree to leave dtypes of indptr and indices generic.
I closed this PR and moved some fixes to #3948.

leofang:
Thanks 🙏

PR #3950, opened 2020-09-04
ROCm: Support more hipBLAS/rocBLAS and rocSOLVER functions

leofang:
- Most BLAS calls no longer raise `HIPBLAS_STATUS_NOT_SUPPORTED` (unless HIP does not support it)
- Reduced the number of failures in tests/cupy_tests/linalg_tests/ by about 10 (ex: `TestDet`, `TestSlogdet`, etc are now error-free)

~~I think it'd take some efforts to make most tests pass, as we don't yet support rocsolver, but I consider this a good progress and should not be blocked.~~

~~TODO: run a few more tests, perhaps manually, to see if rocBLAS is well-behaving or not.~~

**UPDATE**: It turns out that cuBLAS and cuSOLVER map the BLAS/LAPACK interface poorly. For example, in CUDA `getri` (batched version) is in cuBLAS, and there is no non-batched version, but in ROCm all of them (batched/non-batched) are in rocSOLVER (which correctly follows LAPACK). Another example is `getrf`: in CUDA the batched version is in cuBLAS, but the non-batched version is in cuSOLVER, whereas in ROCm all of them are in rocSOLVER (again correctly following LAPACK). 

Given this, and that rocSOLVER now simply uses rocBLAS's handle (at least as of ROCm 3.5.0), I adjust the headers' contents to follow ROCm's convention, and dump all wrappers to the same place.

Now relevant tests are either passed, or return `rocblas_status_not_implemented`/`HIPBLAS_STATUS_NOT_SUPPORTED` instead of a crappy output with silent success. This compares the failure counts before and after this PR:
```
Before:
tests/cupy_tests/linalg_tests/
164 failed
tests/cupyx_tests/linalg_tests/
20 failed
tests/cupyx_tests/scipy_tests/linalg_tests/
114 failed

After:
tests/cupy_tests/linalg_tests/
138 failed
tests/cupyx_tests/linalg_tests/
20 failed
tests/cupyx_tests/scipy_tests/linalg_tests/
0 failed
```
A lot of failing tests are now due to the lack of support of complex numbers, see rocSOLVER's chart:
https://rocsolver.readthedocs.io/en/latest/userguide_intro.html#brief-description-and-functionality


emcastillo:
btw, AMD folks recommended us to use MAGMA for lapack support

leofang:
I was looking at the rocsolver APIs and found some of them have quite similar signatures with cusolver. Will try to look into that and see how far we can reach. If possible I'd rather not add magma as a dependency. In any case, that would be a separate PR from this one.

leofang:
Jenkins, test this please

leofang:
Blocked by #3936.

leofang:
@emcastillo I have updated the PR description. I think a lot of these work now! My question is: Do we want to skip the tests in which we know ROCm is not working yet (for example due to lack of complex number support), or you think it's fine to leave as is?

leofang:
Jenkins, test this please

emcastillo (approved):

PR #3961, opened 2020-09-08
[WIP] ROCm: support some rocSOLVER functions

leofang:
Blocked by #3950. Address https://github.com/cupy/cupy/pull/3950#issuecomment-688453442. Description to be added.

TODO:
- [ ] More wrappers
- [ ] Ensure the workspaces are zero-size to avoid any potentially uninitialized values
- [x] Handle return type (need a new dict for cusolverStatus_t)
- [ ] TEST!!!! (Currently I only test that it builds and is importable, but not its correctness.)


leofang:
Looks like cuBLAS and cuSOLVER map the BLAS/LAPACK interface poorly. For example, in CUDA `getri` (batched version) is in cuBLAS, and there is no non-batched version, but in ROCm all of them (batched/non-batched) are in rocSOLVER (which correctly follows LAPACK). Another example is `getrf`: in CUDA the batched version is in cuBLAS, but the non-batched version is in cuSOLVER, whereas in ROCm all of them are in rocSOLVER (again correctly following LAPACK). I got to say NVIDIA is not making sense here...

Given this, and that rocSOLVER now simply uses rocBLAS's handle, I will have to adjust the headers' contents to follow ROCm's convention. I think all wrappers should be dumped to the same place.

leofang:
@kmaehashi I just discussed offline with @emcastillo and I will move everything to #3950, since on ROCm it's hard to separate the rocBLAS and rocSOLVER supports. For example, `cublasSgetriBatched` should wrap `rocsolver_sgetri_batched`.

PR #3958, opened 2020-09-07
Refactor sparse indexing tests

asi1024:
Blocked by #3944.

PR #3975, opened 2020-09-09
Fix `sparse.__getitem__` not to return view of input

asi1024:
Blocked by #3958

### DIFF
--- a/cupyx/scipy/sparse/_index.py
+++ b/cupyx/scipy/sparse/_index.py
@@ -22,20 +22,20 @@ _int_scalar_types = (int, numpy.integer, numpy.int_)
 _bool_scalar_types = (bool, numpy.bool, numpy.bool_)
 
 
-def _get_csr_submatrix_major_axis(Ap, Aj, Ax, start, stop):
+def _get_csr_submatrix_major_axis(Ax, Aj, Ap, start, stop):
     """Return a submatrix of the input sparse matrix by slicing major axis.
 
     Args:
-        Ap (cupy.ndarray): indptr array from input sparse matrix
-        Aj (cupy.ndarray): indices array from input sparse matrix
         Ax (cupy.ndarray): data array from input sparse matrix
+        Aj (cupy.ndarray): indices array from input sparse matrix
+        Ap (cupy.ndarray): indptr array from input sparse matrix
         start (int): starting index of major axis
         stop (int): ending index of major axis
 
     Returns:
-        Bp (cupy.ndarray): indptr array of output sparse matrix
-        Bj (cupy.ndarray): indices array of output sparse matrix
         Bx (cupy.ndarray): data array of output sparse matrix
+        Bj (cupy.ndarray): indices array of output sparse matrix
+        Bp (cupy.ndarray): indptr array of output sparse matrix
 
     """
     Ap = Ap[start:stop + 1]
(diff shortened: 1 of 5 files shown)


cupyx/scipy/sparse/compressed.py
```diff
@@ -550,113 +548,49 @@ def _minor_index_fancy(self, idx):
...
+        start, stop, step = idx.indices(N)
 
-        if idx == slice(None):
+        if start == 0 and stop == N and step == 1:
             return self.copy() if copy else self
 
-        M, N = self._swap(*self.shape)
-        start, stop, step = idx.indices(M)
-        M = len(range(start, stop, step))
+        N = max(0, math.ceil((stop - start) / step))
```
toslunar, line 582: `len(range(start, stop, step))` was clear and may be fast enough.  Could you explain the change?

With CPython, `length` is computed on the creation of a range object, and it uses an integer division.  The equivalent formula in Python seems
```python
if step > 0:
    lo, hi = start, stop
else:
    step = -step
    assert step
    lo, hi = stop, start
return ((hi - lo - 1) // step) + 1
```
https://github.com/python/cpython/blob/master/Objects/rangeobject.c

cupyx/scipy/sparse/_index.py
```diff
@@ -22,20 +22,20 @@
 _bool_scalar_types = (bool, numpy.bool, numpy.bool_)
 
 
-def _get_csr_submatrix_major_axis(Ap, Aj, Ax, start, stop):
+def _get_csr_submatrix_major_axis(Ax, Aj, Ap, start, stop):
```
toslunar, line 25: Looks good because it is the same order as in the `scipy.sparse.csr` doc.

toslunar (commented):
The new checks LGTM.

cupyx/scipy/sparse/compressed.py
asi1024: I misunderstood and thought `len(range(start, stop, step))` computes by iterating range object. I will revert this fix.

asi1024:
@toslunar PTAL.

toslunar (approved):
LGTM

Could you make the PR title look like a bugfix?

toslunar:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins server looks failing...

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

cupyx/scipy/sparse/csc.py
```diff
@@ -305,7 +305,7 @@ def getrow(self, i):
         Returns:
             cupyx.scipy.sparse.csc_matrix: Sparse matrix with single row
         """
-        return self._minor_slice(slice(i, i + 1), copy=True)
+        return self.tocsr()._major_slice(slice(i, i + 1), copy=True)
```
toslunar, line 308: Is it faster than `self._minor_slice(slice(i, i + 1), copy=True).tocsr()`?

BTW, I was surprised at output types:

| | `getrow` | `getcol` |
| - | - | - |
| `csr_matrix` | `csr_matrix` | `csr_matrix` |
| `csc_matrix` | `csr_matrix` | `csc_matrix` |

cupyx/scipy/sparse/compressed.py
```diff
@@ -558,9 +558,9 @@ def _major_slice(self, idx, copy=False):
         M = len(range(start, stop, step))
         new_shape = self._swap(M, N)
 
-        if M == 0 or self.nnz == 0:
-            return self.__class__(new_shape)
         if step == 1:
+            if M == 0 or self.nnz == 0:
+                return self.__class__(new_shape, dtype=self.dtype)
```
toslunar, line 563: Could you explain why the shortcut requires `step == 1`?

cupyx/scipy/sparse/csc.py
asi1024: I think `_major_slice` is faster than `_minor_slice`, but not compared the performance yet, I'll show the performance numbers later.

cupyx/scipy/sparse/compressed.py
asi1024: I fixed it for the dtype of the return value. SciPy's sparse array returns a sparse matrix array of `self.dtype` type if `step == 1`, but returns float64 otherwise...

cupyx/scipy/sparse/csc.py
asi1024: `_major_slice` is faster than `_minor_slice` actually. However, I compared the performance and turned out that converting to csr format after slicing is faster. I fixed in b73ca10.

toslunar (approved):

toslunar:
Jenkins, test this please.

PR #3992, opened 2020-09-15
Avoid using CUlinkState unless absolutely necessary

leofang:
I noticed `ls.add_ptr_data()` took about 20% of time in `_compile_with_cache_cuda()`, but it's actually not needed unless we are doing separate compilation and need to link with other input files/data. Otherwise, we can just pass the generated PTX to `cuModuleLoad()` which should work just fine.


leofang:
Jenkins, test this please

takagi:
LGTM!

PR #4008, opened 2020-09-17
Add `apply_along_axis`

grlee77:
This PR implements `apply_along_axis` which is a utility to repeatedly apply a 1d function along a given axis, looping over all other axes.

This function is slightly simplified from [NumPy's](https://github.com/numpy/numpy/blob/c6853c7c27bb8352ab498848439d4fee9eb79a33/numpy/lib/shape_base.py#L268) because it doesn't support a separate `matrix` class and doesn't try to preserve array subclasses like `numpy.ma.core.MaskedArray` via `__array_wrap__` (neither masked arrays or `__array_wrap__` are currently implemented by CuPy)

### DIFF
--- a/cupy/lib/shape_base.py
+++ b/cupy/lib/shape_base.py
@@ -48,7 +48,7 @@ def apply_along_axis(func1d, axis, arr, *args, **kwargs):
     # remove the requested axis, and add the new ones on the end.
     # laid out so that each write is contiguous.
     # for a tuple index inds, buff[inds] = func1d(inarr_view[inds])
-    buff = cupy.zeros(inarr_view.shape[:-1] + res.shape, res.dtype)
+    buff = cupy.empty(inarr_view.shape[:-1] + res.shape, res.dtype)
 
     # permutation of axes such that out = buff.transpose(buff_permute)
     buff_dims = list(range(buff.ndim))
--- a/cupy/lib/shape_base.py
+++ b/cupy/lib/shape_base.py
@@ -50,18 +50,13 @@ def apply_along_axis(func1d, axis, arr, *args, **kwargs):
     # for a tuple index inds, buff[inds] = func1d(inarr_view[inds])
     buff = cupy.empty(inarr_view.shape[:-1] + res.shape, res.dtype)
 
-    # permutation of axes such that out = buff.transpose(buff_permute)
-    buff_dims = list(range(buff.ndim))
-    buff_permute = (
-        buff_dims[0:axis] +
-        buff_dims[buff.ndim - res.ndim:buff.ndim] +
-        buff_dims[axis:buff.ndim - res.ndim]
-    )
-
     # save the first result, then compute and save all remaining results
     buff[ind0] = res
     for ind in inds:
         buff[ind] = func1d(inarr_view[ind], *args, **kwargs)
 
-    # finally, rotate the inserted axes back to where they belong
-    return cupy.transpose(buff, buff_permute)
+    # restore the inserted axes back to where they belong
+    for i in range(res.ndim):
+        buff = cupy.moveaxis(buff, -1, axis)
+
+    return buff
--- a/cupy/lib/shape_base.py
+++ b/cupy/lib/shape_base.py
@@ -1,0 +1,72 @@
+from numpy.lib import index_tricks
+
+import cupy
+from cupy import _util
+
+
+def apply_along_axis(func1d, axis, arr, *args, **kwargs):
+    """Apply a function to 1-D slices along the given axis.
+
+    Args:
+        func1d (function (M,) -> (Nj...)): This function should accept 1-D
+            arrays. It is applied to 1-D slices of ``arr`` along the specified
+            axis. It must return a 1-D ``cupy.ndarray``.
+        axis (integer): Axis along which ``arr`` is sliced.
+        arr (cupy.ndarray (Ni..., M, Nk...)): Input array.
+        args: Additional arguments for `f`unc1d``.
+        kwargs: Additional keyword arguments for ``func1d``.
+
+    Returns:
+        cupy.ndarray: The output array. The shape of ``out`` is identical to
+            the shape of ``arr``, except along the ``axis`` dimension. This
+            axis is removed, and replaced with new dimensions equal to the
+            shape of the return value of ``func1d``. So if ``func1d`` returns a
+            scalar ``out`` will have one fewer dimensions than ``arr``.
+
+    .. seealso:: :func:`numpy.apply_over_axes`
+    """
+    ndim = arr.ndim
+    axis = _util._normalize_axis_index(axis, ndim)
+
+    # arr, with the iteration axis at the end
+    in_dims = list(range(ndim))
+    inarr_view = cupy.transpose(
+        arr, in_dims[:axis] + in_dims[axis + 1:] + [axis]
+    )
+
+    # compute indices for the iteration axes, and append a trailing ellipsis to
+    # prevent 0d arrays decaying to scalars
+    inds = index_tricks.ndindex(inarr_view.shape[:-1])
+    inds = (ind + (Ellipsis,) for ind in inds)
+
+    # invoke the function on the first item
+    try:
+        ind0 = next(inds)
+    except StopIteration:
+        raise ValueError(
+            'Cannot apply_along_axis when any iteration dimensions are 0'
+        )
+    # cupy.asarray needed in case func1d returns a scalar
+    res = cupy.asarray(func1d(inarr_view[ind0], *args, **kwargs))
+
+    # build a buffer for storing evaluations of func1d.
+    # remove the requested axis, and add the new ones on the end.
+    # laid out so that each write is contiguous.
+    # for a tuple index inds, buff[inds] = func1d(inarr_view[inds])
+    buff = cupy.zeros(inarr_view.shape[:-1] + res.shape, res.dtype)
+
+    # permutation of axes such that out = buff.transpose(buff_permute)
+    buff_dims = list(range(buff.ndim))
+    buff_permute = (
+        buff_dims[0:axis] +
+        buff_dims[buff.ndim - res.ndim:buff.ndim] +
+        buff_dims[axis:buff.ndim - res.ndim]
+    )
+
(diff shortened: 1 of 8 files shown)


grlee77:
In most cases, for ufuncs that take an axis argument it will be preferable to use that rather than this function. However, here is a demo of a case where reduction along a non-contiguous axis via `cp.apply_along_axis` is currently faster than using the axis argument for `cp.mean` (with CUB enabled):

```Python
import cupy as cp
from cupyx.time import repeat

a = cp.random.randn(1000000, 16)
perf1 = repeat(cp.mean, (a, 0), n_warmup=20, n_repeat=20)
perf2 = repeat(cp.apply_along_axis, (cp.mean, 0, a), n_warmup=20, n_repeat=20)

print("Result for cp.mean:")
print(perf1)

print("Result for cp.apply_along_axis:")
print(perf2)
```

```
Result for cp.mean:
mean                :    CPU:   31.015 us   +/-13.272 (min:   21.701 / max:   81.904) us     GPU-0:15901.830 us   +/-213.947 (min:15736.736 / max:16391.808) us
Result for cp.apply_along_axis:
apply_along_axis    :    CPU:  532.148 us   +/-26.029 (min:  504.228 / max:  612.164) us     GPU-0:10424.643 us   +/- 8.711 (min:10407.936 / max:10451.968) us
```

On the contrary, if I create a with shape (16, 1000000) and take the reduction along axis 1 (contiguous), the computation is much faster and the built-in `cupy.mean` comes out twice as fast. 

```Python
a = cp.random.randn(16, 1000000)
perf1 = repeat(cp.mean, (a, 1), n_warmup=20, n_repeat=20)
perf2 = repeat(cp.apply_along_axis, (cp.mean, 1, a), n_warmup=20, n_repeat=20)

print("Result for cp.mean:")
print(perf1)

print("Result for cp.apply_along_axis:")
print(perf2)
```

```
Result for cp.mean:
mean                :    CPU:   67.724 us   +/-23.307 (min:   51.729 / max:  147.081) us     GPU-0:  394.077 us   +/-19.734 (min:  380.032 / max:  465.088) us
Result for cp.apply_along_axis:
apply_along_axis    :    CPU:  793.639 us   +/-105.092 (min:  723.818 / max: 1096.841) us     GPU-0:  800.146 us   +/-105.610 (min:  730.112 / max: 1104.896) us
```

cupy/lib/shape_base.py
```diff
@@ -0,0 +1,72 @@
...
+    .. seealso:: :func:`numpy.apply_over_axes`
+    """
+    ndim = arr.ndim
+    axis = _util._normalize_axis_index(axis, ndim)
+
+    # arr, with the iteration axis at the end
+    in_dims = list(range(ndim))
+    inarr_view = cupy.transpose(
+        arr, in_dims[:axis] + in_dims[axis + 1:] + [axis]
+    )
```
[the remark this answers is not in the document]
grlee77: sure. I found that moveaxis took > 3 microseconds while the above takes around 2.5, but this is not a very consequential difference.

cupy/lib/shape_base.py
```diff
@@ -0,0 +1,72 @@
...
+def apply_along_axis(func1d, axis, arr, *args, **kwargs):
+    """Apply a function to 1-D slices along the given axis.
+
+    Args:
+        func1d (function (M,) -> (Nj...)): This function should accept 1-D
+            arrays. It is applied to 1-D slices of ``arr`` along the specified
+            axis. It must return a 1-D ``cupy.ndarray``.
+        axis (integer): Axis along which ``arr`` is sliced.
+        arr (cupy.ndarray (Ni..., M, Nk...)): Input array.
+        args: Additional arguments for `f`unc1d``.
```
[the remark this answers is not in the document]
grlee77: done

grlee77:
If the new scalar moveaxis function is not desired let me know and I will refactor to only include the minor Cython improvements to the existing moveaxis function.

This PR reduces the time for a call like `cp.moveaxis(arr, 0, 1)` from just over 3 to around 1.8 us (or 1.5 us if source and destination axis are the same). I found four additional functions that use moveaxis with scalar axes only and may also benefit from this:  `linspace`, `pad`, `polynomial` and `linalg.product`.

grlee77:
The `_normalize_axis_index` helper in `_routines_manipulation` is a `cdef` copy of the one defined in `cupy._util.pyx`. I'm not sure if the one in `_util.pyx` should now just call the one from `_routines_manipulation` internally?

leofang:
> The `_normalize_axis_index` helper in `_routines_manipulation` is a `cdef` copy of the one defined in `cupy/_util.pyx`. I'm not sure if the one in `_util.pyx` should now just call the one from `_routines_manipulation` internally?

Just my two cents, you should wait for @asi1024's reply: The helper should be moved to `cupy/core/internal.pyx`.

cupy/lib/shape_base.py
```diff
@@ -0,0 +1,67 @@
...
+            'Cannot apply_along_axis when any iteration dimensions are 0'
+        )
+    # cupy.asarray needed in case func1d returns a scalar
+    res = cupy.asarray(func1d(inarr_view[ind0], *args, **kwargs))
+
+    # build a buffer for storing evaluations of func1d.
+    # remove the requested axis, and add the new ones on the end.
+    # laid out so that each write is contiguous.
+    # for a tuple index inds, buff[inds] = func1d(inarr_view[inds])
+    buff = cupy.zeros(inarr_view.shape[:-1] + res.shape, res.dtype)
```
leofang, line 51: Just a nitpick: If all axes are full covered in the for loop below, perhaps we just need `cupy.empty` here to avoid launching a kernel?

cupy/lib/shape_base.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # permutation of axes such that out = buff.transpose(buff_permute)
+    buff_dims = list(range(buff.ndim))
+    buff_permute = (
+        buff_dims[0:axis] +
+        buff_dims[buff.ndim - res.ndim:buff.ndim] +
```
leofang, line 57: Stupid question: What does this indexing mean? It seems to work but I didn't know that an `int` subtracts a `slice` can work.

grlee77: I think maybe the whitespace is misleading you. It is equivalent to:

```Python
buff_dims[(buff.ndim - res.ndim):buff.ndim]
``` 
or 
```Python
buff_dims[slice(buff.ndim - res.ndim, buff.ndim)]
```

grlee77:
> Just my two cents, you should wait for @asi1024's reply: The helper should be moved to `cupy/core/internal.pyx`

That seems like a reasonable location. If we do that then the existing `_normalize_axis_tuple` should also probably be moved there.

The other style I see is in `_routines.sorting.pyx`, the functions `_ndarray_sort`, `_ndarray_argsort`, `_ndarray_partition`, `_ndarray_argpartition` all just use duplicate in-line code like the following:
https://github.com/cupy/cupy/blob/dd2fb09f63fce8b7ddb58deb4b91f0b3669d1a0d/cupy/core/_routines_sorting.pyx#L34-L37
I am not sure if that helps avoid any function call overhead or if those should also just call this utility.

A couple of other places in the `core` module (`_fusion_trace.pyx` and `_routines_math.pyx`), currently use the version from `cupy/_util.pyx`.

I think my preference would be to move to internal as you suggest and just use this version for everything within core.

cupy/lib/shape_base.py
grlee77: Yes, they are fully covered so `empty` should be better.

cupy/lib/shape_base.py
leofang: Ah indeed I parsed the operator precedence wrong 😂 Thanks for clarifying Gregory!

cupy/lib/shape_base.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # invoke the function on the first item
+    try:
+        ind0 = next(inds)
+    except StopIteration:
+        raise ValueError(
+            'Cannot apply_along_axis when any iteration dimensions are 0'
+        )
+    # cupy.asarray needed in case func1d returns a scalar
+    res = cupy.asarray(func1d(inarr_view[ind0], *args, **kwargs))
```
asi1024, line 45: If `func1d` returns a `numpy.ndarray`, it will be converted to cupy.ndarray in this line. However, `ValueError` will be raised in L64 only if the length of the reduction axis is greater than 1.
How about checking if the return value of `func1d` is scalar or cupy.ndarray value?

cupy/lib/shape_base.py
```diff
@@ -0,0 +1,67 @@
...
+        buff_dims[axis:buff.ndim - res.ndim]
+    )
+
+    # save the first result, then compute and save all remaining results
+    buff[ind0] = res
+    for ind in inds:
+        buff[ind] = func1d(inarr_view[ind], *args, **kwargs)
+
+    # finally, rotate the inserted axes back to where they belong
+    return cupy.transpose(buff, buff_permute)
```
asi1024, line 67: Just a nitpicks but how about applying `moveaxis`  several times as follows for the simplification?
```suggestion
    for i in range(res.ndim):
        buf = cupy.moveaxis(buf, -1, axis)
    return buf
```
This implementation has a for-loop, but I think `res.ndim` will be not large in most cases.

asi1024:
I agree to move `_normalize_axis_{index, indices}` to `cupy/core/internal.pyx` too! I will work on it after this PR is merged.

cupy/lib/shape_base.py
grlee77: Nice. Yes, I think `res.ndim` will most commonly be 1.

cupy/lib/shape_base.py
grlee77: See if the change in 3f6ec11 matches what your were suggesting? Now if `func1d` returns a non-scalar numpy array, a value ValueError (`"non-scalar numpy.ndarray cannot be used for fill"`) will occur when trying to assign the NumPy result to `buff`.

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

PR #4013, opened 2020-09-21
Improve the plan cache documentation

leofang:
Follow up of #3730. This PR moves `cupy.fft._cache.PlanCache` from the auto summary to a separate page, as it is meant to be a private class (for now). But, this change introduces a warning in Sphinx, which I am not quite sure how to fix, although it's certainly harmless (and intended):
```
...
pickling environment... done
checking consistency... /home/leofang/dev/cupy_rocm3.5.0/docs/source/reference/generated/cupy.fft._cache.PlanCache.rst: WARNING: document isn't included in any toctree
done
preparing documents... done
...
```
In addition, I also fixed a few hyperlink issues in the docstrings.

### DIFF
--- a/cupy/fft/_cache.pyx
+++ b/cupy/fft/_cache.pyx
@@ -192,7 +192,8 @@ cdef class PlanCache:
 
         2. This class can be instantiated by users, but it is discouraged.
            Instead, we expect the following canonical usage pattern to
-           retrieve a handle to the cache through :func:`get_plan_cache`:
+           retrieve a handle to the cache through
+           :func:`~cupy.fft.config.get_plan_cache`:
 
            .. code-block:: python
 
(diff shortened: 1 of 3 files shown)


leofang:
Seems like the Sphinx warning is safely ignored in the CIs.

leofang:
Jenkins, test this please

asi1024:
`cupy.fft._cache` is a private module, so we don't want to document `PlanCache` if the users will not call `PlanCache.__init__`. How about removing `PlanCache` documentation or moving them to `get_fft_plan`?

leofang:
Yeah it is a somewhat interesting situation. We don't recommend users to initialize their own `PlanCache` as discussed in #3730 IIRC, but if they call `cupy.fft.config.get_plan_cache()`, for example, they will get the `PlanCache` instance that we create. So in this regard I think it's better to document it so that advanced users can adjust the cache settings, but in a way that the doc page does not appear in TOC or autosummary.

The original purpose of adding `cupy.fft._cache.PlanCache` to autosummary was to force generating the doc page, because we have `.. seealso`in the public helpers `cupy.fft.config.get_plan_cache()` and `cupy.fft.config.show_plan_cache_info()`  pointing to it. Without having an `cupy.fft._cache.PlanCache` entry floating somewhere, the doc page would not be generated, and the hyperlinks would have no effect. So this PR simply moves the entry to a different place.

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

leofang:
Thanks!

leofang:
@asi1024 Forgot to ask: Could you backport this PR?

PR #4015, opened 2020-09-21
Fix managed memory leak

leofang:
Looks like the managed memory is not freed even after `free_all_blocks()` is called, see https://stackoverflow.com/q/63998616/2344149. I checked this PR is in effect by monitoring `top` and `nvidia-smi`.


leofang:
Jenkins, test this please

kmaehashi:
LGTM!

PR #4018, opened 2020-09-23
Improve `cupy.in1d`

mrkwjc:
New implementation is based on numpy code. This is much much faster and less less memory hungry. Call to `broadcast_to` is avoided now. Additionally `assume_unique` keyword is handled.


leofang:
Very interesting! Could you share insights on why the NumPy approach works better (which is not obvious to me), and also post performance benchmarks that you did before and with this PR?

mrkwjc:
Numpy uses approach with sorting input arrays, whereas cupy tries direct comparison by broadcasting inputs. Broadcasting is generally ok, it creates strided views on inputs, but the next comparison line tries to allocate MxN boolean array, where M, N are lengths of inputs. This is a recipe for out of memory error:

```python
In [11]: import cupy

In [12]: a = cupy.random.randint(0, 1000000, size=500000)

In [13]: b = cupy.random.randint(0, 1000000, size=100000)

In [14]: cupy.in1d(a, b)
OutOfMemoryError                          Traceback (most recent call last)
<ipython-input-14-e28ce4babb52> in <module>
----> 1 cupy.in1d(a, b)

~/miniconda3/lib/python3.8/site-packages/cupy/logic/truth.py in in1d(ar1, ar2, assume_unique, invert)
     92     ar1_broadcast = cupy.broadcast_to(ar1[..., cupy.newaxis], shape)
     93     ar2_broadcast = cupy.broadcast_to(ar2, shape)
---> 94     count = (ar1_broadcast == ar2_broadcast).sum(axis=1)
     95     if invert:
     96         return count == 0

cupy/core/core.pyx in cupy.core.core.ndarray.__richcmp__()

cupy/core/_kernel.pyx in cupy.core._kernel.ufunc.__call__()

cupy/core/_kernel.pyx in cupy.core._kernel._get_out_args()

cupy/core/core.pyx in cupy.core.core.ndarray.__init__()

cupy/cuda/memory.pyx in cupy.cuda.memory.alloc()

cupy/cuda/memory.pyx in cupy.cuda.memory.MemoryPool.malloc()

cupy/cuda/memory.pyx in cupy.cuda.memory.MemoryPool.malloc()

cupy/cuda/memory.pyx in cupy.cuda.memory.SingleDeviceMemoryPool.malloc()

cupy/cuda/memory.pyx in cupy.cuda.memory.SingleDeviceMemoryPool._malloc()

cupy/cuda/memory.pyx in cupy.cuda.memory._try_malloc()

OutOfMemoryError: Out of memory allocating 50,000,000,000 bytes (allocated so far: 18,521,088 bytes).
```
Nothing like this happens with numpy's implementation, also for many times bigger inputs. In terms of speed current in1d is poor for larger inputs:
```python
In [15]: a = cupy.random.randint(0, 100000, size=50000)

In [16]: b = cupy.random.randint(0, 100000, size=50000)

In [17]: %timeit cupy.in1d(a, b)
45.7 ms ± 37.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [18]: %timeit in1d(a, b)  # Numpy-like implementation
1.78 ms ± 72.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
For smaller arrays `cupy.in1d` is faster:
```python
In [19]: b = cupy.random.randint(0, 100000, size=5000)

In [20]: a = cupy.random.randint(0, 100000, size=5000)

In [21]: timeit cupy.in1d(a, b)
490 µs ± 112 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [22]: %timeit in1d(a, b)  # Numpy-like implementation
1.71 ms ± 59.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
So maybe for small inputs, here some heuristics have to be used, the current approach should still be used?

asi1024:
@mrkwjc Thank you for the PR!
The memory usage explosion in `ar1_broadcast == ar2_broadcast` seems to be avoidable by using `cupy.ReductionKernel`. I will compare the performance numbers between this PR and ReductionKernel version.

asi1024:
I confirmed that this PR improves GPU performance, but causes host/device synchronization.
```
master    :    CPU:   31.325 us   +/-37.739 (min:   27.504 / max:  662.737) us     GPU-0:16218.380 us   +/-52.691 (min:16190.464 / max:16840.704) us
this PR   :    CPU: 1228.504 us   +/-82.605 (min: 1207.951 / max: 5493.522) us     GPU-0: 1366.806 us   +/-80.935 (min: 1344.512 / max: 5547.008) us
```

cupy/logic/truth.py
```diff
@@ -82,23 +84,41 @@ def in1d(ar1, ar2, assume_unique=False, invert=False):
...
+        bool_ar = (sar[1:] == sar[:-1])
+    flag = cupy.concatenate((bool_ar, cupy.asarray([invert])))
+    ret = cupy.empty(ar.shape, dtype=cupy.bool_)
+    ret[order] = flag
+
+    if assume_unique:
+        return ret[:len(ar1)]
     else:
-        return count > 0
+        return ret[rev_idx]
```
asi1024, line 121: How about using `cupy.searchsorted`? This implementation is faster than this PR and does not cause any CPU performance degradation.
```
    ar2 = cupy.sort(ar2)
    v1 = cupy.searchsorted(ar2, ar1, 'left')
    v2 = cupy.searchsorted(ar2, ar1, 'right')
    return v1 == v2 if invert else v1 != v2
```

```
in1d                :    CPU:  158.895 us   +/-19.169 (min:  155.257 / max: 1831.730) us     GPU-0:  362.021 us   +/-18.003 (min:  357.376 / max: 1875.968) us
```

cupy/logic/truth.py
```diff
@@ -82,23 +84,41 @@ def in1d(ar1, ar2, assume_unique=False, invert=False):
...
     ar2 = ar2.ravel()
-    if ar1.size == 0 or ar2.size == 0:
+
+    # Seems this conditions works nicely
+    if len(ar2) < 10 * len(ar1) ** 0.145:
         if invert:
-            return cupy.ones(ar1.shape, dtype=cupy.bool_)
+            mask = cupy.ones(len(ar1), dtype=cupy.bool_)
+            for a in ar2:
+                mask &= (ar1 != a)
```
asi1024, line 95: How about using `cupy.all` to parallelize this for-loop?

cupy/logic/truth.py
mrkwjc: This is a brilliant piece. To be applied. With this I would ignore the workarounds for `if len(ar2) < 10 * len(ar1) ** 0.145.

mrkwjc: I pushed new `in1d` with this code. My humble investigations on my laptop P1000 show that this is much faster then cupy if only `N = ar1.size = ar2.size > ~300`. If `N < 300` the worst case is that this is two times slower. Didn't tested other devices and didn't tested the case when `ar1.size != ar2.size`.

asi1024:
Jenkins, test this please.

asi1024:
LGTM. Thank you for the PR!

cupy/logic/truth.py
vsraptor: > How about using `cupy.searchsorted`? This implementation is faster than this PR and does not cause any CPU performance degradation.
> 
> ```
>     ar2 = cupy.sort(ar2)
>     v1 = cupy.searchsorted(ar2, ar1, 'left')
>     v2 = cupy.searchsorted(ar2, ar1, 'right')
>     return v1 == v2 if invert else v1 != v2
> ```
> 
> ```
> in1d                :    CPU:  158.895 us   +/-19.169 (min:  155.257 / max: 1831.730) us     GPU-0:  362.021 us   +/-18.003 (min:  357.376 / max: 1875.968) us
> ```

.isin() on my laptop on CPU is **6 times slower** than stock numpy on (200000,100) array..
```

def in1d(ar1, ar2, assume_unique=False, invert=False):

	ar1 = ar1.ravel()
	ar2 = ar2.ravel()
	if ar1.size == 0 or ar2.size == 0:
		if invert:
			return np.ones(ar1.shape, dtype=np.bool)
		else:
			return np.zeros(ar1.shape, dtype=np.bool)
	ar2 = np.sort(ar2)
	v1 = np.searchsorted(ar2, ar1, 'left')
	v2 = np.searchsorted(ar2, ar1, 'right')
	return v1 == v2 if invert else v1 != v2

def isin(element, test_elements, assume_unique=False, invert=False):
	return in1d(element, test_elements, assume_unique=assume_unique, invert=invert).reshape(element.shape)
```

PR #3765, opened 2020-08-11
Use cuTENSOR in `cupy.prod`, `cupy.max`, `cupy.min`, `cupy.ptp` and `cupy.mean`

asi1024:
Blocked by #3700 and #3732.

PR #4024, opened 2020-09-25
Fix potential segfault when reduction axis is empty

### LINKED ISSUE #4023: segfault with cutensor reduction over empty axis
### Conditions

I am building from current master
```
CuPy Version          : 8.0.0rc1
CUDA Build Version    : 11000
CUDA Driver Version   : 11010
CUDA Runtime Version  : 11010
cuBLAS Version        : 11201
cuFFT Version         : 10300
cuRAND Version        : 10202
cuSPARSE Version      : 11200
NVRTC Version         : (11, 0)
Thrust Version        : 100909
CUB Build Version     : 100910
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2708
NCCL Runtime Version  : 2708
cuTENSOR Version      : 10200
```

## Code to reproduce

The following causes a segfault if cutensor is enabled via cupy.core.set_routine_accelerators([2])`

```Python
cupy.arange(8, dtype=float).sum(())
```

For CUB via `cupy.core.set_routine_accelerators([1])` it does not crash, but does give the following warning
```
Exception ignored in: 'cupy.cuda.cub._cub_device_segmented_reduce_axis_compatible'
IndexError: tuple index out of range
```

There is no warning or crash if both accelerators are disabled

grlee77:
closes #4023

~TODO: The test cases here no longer segfault, but many of the Fortran-contiguous cases do not pass. This seems to be a separate bug somewhere in reductions on F-ordered arrays when CUB is disabled.~

After a clean rebuild, the only difference is that sometimes unaccelerated or cuTENSOR reductions reductions return a result with different contiguity. The actual values seem to be fine and the tests here all pass for me now.

### DIFF
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -265,13 +265,13 @@ cpdef inline tuple _can_use_cub_block_reduction(
 
     # check reduction axes, if not contiguous then fall back to old kernel
     if in_arr._f_contiguous:
-        order = 'f'
+        order = 'F'
         if not cub._cub_device_segmented_reduce_axis_compatible(
                 reduce_axis, in_arr.ndim, order):
             return None
         axis_permutes_cub = tuple(sorted(reduce_axis) + sorted(out_axis))
     elif in_arr._c_contiguous:
-        order = 'c'
+        order = 'C'
         if not cub._cub_device_segmented_reduce_axis_compatible(
                 reduce_axis, in_arr.ndim, order):
             return None
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -271,6 +271,10 @@ cpdef inline tuple _can_use_cub_block_reduction(
             return None
         axis_permutes_cub = tuple(sorted(reduce_axis) + sorted(out_axis))
     elif in_arr._c_contiguous:
+        order = 'c'
+        if not cub._cub_device_segmented_reduce_axis_compatible(
+                reduce_axis, in_arr.ndim, order):
+            return None
         axis_permutes_cub = tuple(sorted(out_axis) + sorted(reduce_axis))
     else:
         return None
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -263,6 +263,8 @@ cpdef inline tuple _can_use_cub_block_reduction(
 
     # check reduction axes, if not contiguous then fall back to old kernel
     if in_arr._f_contiguous:
+        if reduce_axis is () and not in_arr._c_contiguous:
+            return None
         axis_permutes_cub = tuple(sorted(reduce_axis) + sorted(out_axis))
     elif in_arr._c_contiguous:
         axis_permutes_cub = tuple(sorted(out_axis) + sorted(reduce_axis))
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -7,6 +7,7 @@ from cupy.core.core cimport _internal_asfortranarray
 from cupy.core.core cimport compile_with_cache
 from cupy.core.core cimport ndarray
 from cupy.core cimport internal
+from cupy.cuda cimport cub
 from cupy.cuda cimport function
 from cupy.cuda cimport memory
 from cupy_backends.cuda.api cimport runtime
@@ -247,6 +248,7 @@ cpdef inline tuple _can_use_cub_block_reduction(
     cdef tuple axis_permutes_cub
     cdef ndarray in_arr, out_arr
     cdef Py_ssize_t contiguous_size = 1
+    cdef str order
 
     # detect whether CUB headers exists somewhere:
     if _cub_path is None:
@@ -263,7 +265,9 @@ cpdef inline tuple _can_use_cub_block_reduction(
 
     # check reduction axes, if not contiguous then fall back to old kernel
     if in_arr._f_contiguous:
-        if reduce_axis is () and not in_arr._c_contiguous:
+        order = 'f'
+        if not cub._cub_device_segmented_reduce_axis_compatible(
+                reduce_axis, in_arr.ndim, order):
             return None
         axis_permutes_cub = tuple(sorted(reduce_axis) + sorted(out_axis))
     elif in_arr._c_contiguous:
--- a/tests/cupy_tests/math_tests/test_sumprod.py
+++ b/tests/cupy_tests/math_tests/test_sumprod.py
@@ -5,7 +5,6 @@ import pytest
 
 import cupy
 import cupy._util
-from cupy.core import _accelerator
 from cupy import testing
 
 
@@ -210,11 +209,11 @@ class TestSumprod(unittest.TestCase):
 class TestCubReduction(unittest.TestCase):
 
     def setUp(self):
-        self.old_accelerators = _accelerator.get_routine_accelerators()
-        _accelerator.set_routine_accelerators(['cub'])
+        self.old_accelerators = cupy.core.get_routine_accelerators()
+        cupy.core.set_routine_accelerators(['cub'])
 
     def tearDown(self):
-        _accelerator.set_routine_accelerators(self.old_accelerators)
+        cupy.core.set_routine_accelerators(self.old_accelerators)
 
     @testing.for_contiguous_axes()
     # sum supports less dtypes; don't test float16 as it's not as accurate?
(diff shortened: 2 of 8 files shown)


grlee77:
Actually the tests here apply to reductions, but the original issue was related to math routines such as `sum`. The underlying cause of the segfault is the same, though

asi1024:
@grlee77 Could you fix for Travis failure?

grlee77:
Also, I see that cuTENSOR max/min tests should probably be removed unless #3765 is merged first. I think currently those cases will also just be using the default, unaccelerated implementation.

grlee77:
I can look at adding tests for sum to `tests/cupy_tests/math_tests/test_sumprod.py` to more closely match the original issue that was reported.

grlee77:
@asi1024: I fixed the Travis failure and updated the test cases. Please take another look

In some test cases, array contiguity didn't match SciPy and I had to set `contiguous_check=False` in the `numpy_cupy_allclose` decorator, but that is a separate issue from the one this PR is trying to resolve. I don't know that we typically guarantee the same contiguity for the result in any case?

asi1024:
@grlee77 We guarantee the same contiguity in most cases, but not for all cases. I think it is OK to set `contiguous_check=False` in this situation.

asi1024:
Jenkins, test this please.

grlee77:
I thought all of these tests were passing locally when I uploaded this, but I can reproduce these failures running locally with 
`CUPY_ACCELERATORS="cub" pytest tests/cupy_tests/core_tests/test_ndarray_reduction.py -v`

I will fix it locally and then ping again.

grlee77:
Some tests fail locally if:
`CUPY_ACCELERATORS="cub"` or `CUPY_ACCELERATORS="cub,cutensor"`

But all pass if:
`CUPY_ACCELERATORS=""` or `CUPY_ACCELERATORS="cutensor"`

I was a bit surprised by this given that `TestCubReduction`'s `setUp` and `tearDown` methods should be setting the accelerators independently of what the environment is. However these methods are only calling `set_routine_accelerators` while it seems that the setting of `set_reduction_accelerators` also matters when it falls back to `_amax` or `_amin`.

grlee77:
Okay, this should be working now for all environment variable settings and is ready to test on CI again.

Apparently there needed to be a stricter axis check criteria within `_can_use_cub_block_reduction` function within `_cub.reduction.pyx` to prevent passing incompatible Fortran-ordered arrays. I reused the one from `cupy.cuda.cub.pyx` here.

@leofang, can you double-check this? Without the change in f16a2e7 there were still failures in `TestUnacceleratedReduction` when the **routine accelerators** were disabled, but the environment variable included CUB in the default **reduction accelerators**.

cupy/core/_cub_reduction.pyx
```diff
@@ -263,6 +265,10 @@ cpdef inline tuple _can_use_cub_block_reduction(
 
     # check reduction axes, if not contiguous then fall back to old kernel
     if in_arr._f_contiguous:
+        order = 'f'
+        if not cub._cub_device_segmented_reduce_axis_compatible(
+                reduce_axis, in_arr.ndim, order):
+            return None
```
leofang, line 271: Looks like the correct thing to do, thanks Gregory! Perhaps we should just call `_cub_device_segmented_reduce_axis_compatible()` for both C and F orders, and then construct `axis_permutes_cub`?

cupy/core/_cub_reduction.pyx
```diff
@@ -7,6 +7,7 @@ from cupy.core.core cimport _internal_asfortranarray
 from cupy.core.core cimport compile_with_cache
 from cupy.core.core cimport ndarray
 from cupy.core cimport internal
+from cupy.cuda cimport cub
```
leofang, line 10: I am glad I made things there cdef'd...

tests/cupy_tests/math_tests/test_sumprod.py
```diff
@@ -319,6 +336,57 @@ def _mitigate_cumprod(self, xp, dtype, result):
...
+        a = testing.shaped_random(self.shape, xp, dtype)
+        if self.order in ('c', 'C'):
+            a = xp.ascontiguousarray(a)
+        elif self.order in ('f', 'F'):
+            a = xp.asfortranarray(a)
+
+        if xp is numpy:
+            return a.sum(axis=axis)
+
+        # xp is cupy, first ensure we really use CUB
```
leofang, line 370: nitpick: CUB -> cuTENSOR?

leofang:
LGTM except for a few harmless nitpicks. Let's get it tested in case we wanna backport it.

Jenkins, test this please

kmaehashi:
Thank you! We're updating chainer-test so please wait until https://github.com/chainer/chainer-test/pull/607 got merged.

cupy/core/_cub_reduction.pyx
grlee77: I have now added this call to the C-contiguous case as well

leofang: ?? Did you push @grlee77?

grlee77: whoops. I did it, but forgot to commit the change. It is there now in f83a0ca

leofang (approved):
LGTM! Thanks, @grlee77.

asi1024:
Jenkins, test this please.

grlee77:
The failures on CI were because of a mistake using lowercase characters for order.
I fixed those, but this is not quite ready to test again. 

Now that CUB is not always avoided outright in the reduction kernel cases, there are a few failures again on multidimensional arrays when order='F'. These occur when the tests are run with "cub" in CUPY_ACCELERATORS. I cannot look into it right now, but will check back later today. 

To reproduce:
`CUPY_ACCELERATORS="cub" pytest tests/cupy_tests/core_tests/test_ndarray_reduction.py -v -k test_unaccel`


```
===================================================================================== short test summary info ======================================================================================
FAILED tests/cupy_tests/core_tests/test_ndarray_reduction.py::TestUnacceleratedReduction_param_6_{order='F', shape=(10, 20, 30)}::test_unaccelerated_max - AssertionError: Parameterized test fai...
FAILED tests/cupy_tests/core_tests/test_ndarray_reduction.py::TestUnacceleratedReduction_param_6_{order='F', shape=(10, 20, 30)}::test_unaccelerated_min - AssertionError: Parameterized test fai...
FAILED tests/cupy_tests/core_tests/test_ndarray_reduction.py::TestUnacceleratedReduction_param_7_{order='F', shape=(10, 20, 30, 40)}::test_unaccelerated_max - AssertionError: Parameterized test...
FAILED tests/cupy_tests/core_tests/test_ndarray_reduction.py::TestUnacceleratedReduction_param_7_{order='F', shape=(10, 20, 30, 40)}::test_unaccelerated_min - AssertionError: Parameterized test...
```

leofang:
Hi @grlee77, this is indeed very confusing...😞 My current understanding is that the current implementation gives a wrong answer when dim > 2 as you discovered and banned in 69d3806. (To be precise, some elements are correct but some are wrong, and not just misplacement but totally off.) 

But, if I comment out this line
https://github.com/cupy/cupy/blob/f5417c40e6b704f96cd1d2d40b779d2a9381ec7e/cupy/core/_cub_reduction.pyx#L569-L570
Then the outcome will be correct up to a transpose or rearrangement. I could take a deeper look to understand what happened between my original PR and the current status that led to this wrong behavior. I don't think it was like this before, so thanks for catching it!

grlee77:
> My current understanding is that the current implementation gives a wrong answer when dim > 2 as you discovered and banned in 69d3806

Yes, I was not able to resolve the issue quickly when I looked at it yesterday, so I just put this commit to avoid the problematic case. It is a separate issue than the one this PR was intended to address, it is just that the new test cases here revealed it. 

One option is to merge this PR with the workaround above and open a separate issue regarding fixing this case.

grlee77:
I had to rebase to fix a merge conflict, but have not otherwise changed anything here since yesterday

leofang:
> One option is to merge this PR with the workaround above and open a separate issue regarding fixing this case.

Yeah let's do this.

By the way, for `TestUnacceleratedReduction` could you also set reduction accelerators to empty in `setUp`? This will provide the full control without any interference with `CUPY_ACCELERATORS`.

grlee77:
> By the way, for TestUnacceleratedReduction could you also set reduction accelerators to empty in setUp?

Done. This should be ready for CI testing again

leofang:
Jenkins, test this please

leofang:
@grlee77 I fixed the CUB bug in #4062. It should be merged after yours here.

leofang (approved):
LGTM on the CUB related changes.

asi1024:
Jenkins, test this please.

asi1024:
LGTM! Thanks!

leofang:
Thanks, @grlee77 @asi1024!

PR #4036, opened 2020-09-29
Use `__dealloc__` instead of `__del__` for cdef class

### LINKED ISSUE #3999: cdef classes should not define `__del__()`?
When investigating cdef classes' deallocation behavior, I noticed Cython's documentation has a small footnote on this:
> Note There is no `__del__()` method for extension types.

So what it really means, IIUC, is that we can define `__del__()` and the compilation will pass, but it just won't be invoked. For example, the print statement in the snippet below is never executed, even if `gc.collect()` is called:
```python
cdef class TEST:
    def __init__(self, x):
        self.x = x

    def __del__(self):
        print("oh no")
```
So any C pointer deallocation should be done through `__dealloc__()` instead.

@asi1024 Could you confirm this behavior? I think in one of the cythonizing cuTENSORPRs we left `__del__()` untouched in a few cdef classes:
https://github.com/cupy/cupy/blob/faea9d9553e9d1e9cd8ada8ec201d24713e51b41/cupy_backends/cuda/libs/cutensor.pyx#L159-L161

kmaehashi:
Fixes #3999

### DIFF
--- a/cupy_backends/cuda/libs/cutensor.pyx
+++ b/cupy_backends/cuda/libs/cutensor.pyx
@@ -156,7 +156,7 @@ cdef class Handle:
     def __init__(self):
         self._ptr = PyMem_Malloc(sizeof(cutensorHandle_t))
 
-    def __del__(self):
+    def __dealloc__(self):
         PyMem_Free(self._ptr)
         self._ptr = NULL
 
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, test this please.

asi1024:
LGTM.

leofang:
Thanks!

PR #4058, opened 2020-10-02
allow non-contiguous array input to binary morphology functions

grlee77:
Similar to other ndimage functions, the input does not have to be C contiguous for the binary erosion kernel. Test cases were added to verify this.

A typo in spelling `cupy.asfortranarray` was fixed in the `mask` code path as well.


emcastillo (approved):

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

Issue #4061, opened 2020-10-04
argmax and argmin return wrong results for F-order arrays


leofang:
On CuPy master:
```python
import cupy as cp
import numpy as np


cp.core.set_routine_accelerators([])
cp.core.set_reduction_accelerators([])

a = cp.asarray([[3, 1], [4, 2]], order='F')
b = cp.asnumpy(a)
print("cp.argmax:", a.argmax(), "  np.argmax:", b.argmax())
print("cp.argmin:", a.argmin(), "  np.argmin:", b.argmin())
```
output:
```
cp.argmax: 1   np.argmax: 2
cp.argmin: 2   np.argmin: 1
```
This bug affects both CuPy's original implementation as well as the CUB block reduction (in `cupy.core._cub_reduction`), but not the CUB device reduction (`cupy.cuda.cub`). My guess is for Fortran arrays we calculated the array index (`_J`) wrong...

kmaehashi:
Thanks for reporting!
@asi1024 Could you look into this in detail?

PR #4062, opened 2020-10-04
Fix CUB block reduction for F-order arrays with ndim > 2

leofang:
Follow-up of ~~and blocked by~~ #4024. This is to fix the bug caught in https://github.com/cupy/cupy/pull/4024#issuecomment-701370441. 

I don't know why we didn't catch this, but I think the fix (~~2ea4798~~ 81af347, which is simple) was the original intention; that is, we want the strides etc are made such that we view the underlying data correctly. I expanded the existing CUB tests to cover both device and block reductions (~~dbf5065~~ 35a2669).

This PR correctly does NOT address the `argmin`/`argmax` bug (#4061). Will need to look into that later...


tests/cupy_tests/math_tests/test_sumprod.py
```diff
@@ -5,6 +5,7 @@
 
 import cupy
 import cupy._util
+import cupy.core._accelerator as _acc
```
grlee77, line 8:
```suggestion
import cupy.core._accelerator as _acc
from cupy.core import _cub_reduction
```

grlee77:
> I don't know why we didn't catch this, but I think the fix (2ea4798, which is simple) was the original intention;

Thanks, I figured it was probably something small like this, but looked at it for a little while without seeing where the issue was.

tests/cupy_tests/math_tests/test_sumprod.py
leofang: Thanks!

leofang:
#4024 was in, and I rebased. PTAL.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
@asi1024 @grlee77 I fixed two bugs in 544d9e1:
- The tests for block reduction were skipped due to a typo
- `reduce_axis` and `out_axis` are not sorted for F-order arrays when passed to `_can_use_cub_block_reduction()`: We assumed they are from `_get_axis()`, but the subsequent call to `_sort_axis()` makes them unsorted (due the strides in the F order). 

This is now finally ready! PTAL.

grlee77 (approved):
Looks good. Thanks @leofang!

leofang:
Jenkins, test this please

asi1024:
Jenkins, test this please.

asi1024:
LGTM!

leofang:
Thanks all!

PR #4094, opened 2020-10-07
Make parameterized dtype test skip by `pytest.skip`

takagi:
This PR makes parameterized dtype test possible to skip a test for a specific dtype by `pytest.skip`. 

Once `unittest` was used as CuPy's unit test framework, and we could skip it by `raise unittest.SkipTest()`. It seems that, however, the code was not properly updated when we replaced `unittest` with `pytest` and we need to return a dummy array for that purpose.

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -195,6 +195,7 @@ numpy
 def _signed_counterpart(dtype):
     return numpy.dtype(numpy.dtype(dtype).char.lower()).type
 
+
 def _make_positive_mask(self, impl, args, kw, name, sp_name, scipy_name):
     # Returns a mask of output arrays that indicates valid elements for
     # comparison. See the comment at the caller.
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -718,10 +718,11 @@ def for_dtypes(dtypes, name='dtype'):
     by passing the each element of ``dtypes`` to the named
     argument.
     """
+    check_available()
+
     def decorator(impl):
         @functools.wraps(impl)
         def test_func(self, *args, **kw):
-            check_available()
             for dtype in dtypes:
                 try:
                     kw[name] = numpy.dtype(dtype).type
@@ -937,6 +938,7 @@ def for_dtypes_combination(types, names=('dtype',), full=None):
     If the value is set to ``'1'``, it behaves as if ``full=True``, and
     otherwise ``full=False``.
     """
+    check_available()
 
     types = list(types)
 
@@ -964,7 +966,6 @@ def for_dtypes_combination(types, names=('dtype',), full=None):
     def decorator(impl):
         @functools.wraps(impl)
         def test_func(self, *args, **kw):
-            check_available()
             for dtypes in combination:
                 kw_copy = kw.copy()
                 kw_copy.update(dtypes)
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -8,7 +8,12 @@ import unittest
 from unittest import mock
 import warnings
 
-import _pytest
+try:
+    import _pytest.outcomes
+    _error = None
+except ImportError as e:
+    _error = e
+
 import numpy
 
 import cupy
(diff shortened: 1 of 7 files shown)


asi1024:
@takagi Could you check flake8 failure?

takagi:
Resolved the conflict!

asi1024:
Jenkins, test this please.

toslunar (commented):
How about testing the feature at `TestSkip` and `TestSkipFail` in `tests/cupy_tests/testing_tests/test_helper.py`?

takagi:
Sure!

cupy/testing/helper.py
```diff
@@ -8,6 +8,7 @@
 from unittest import mock
 import warnings
 
+import _pytest
```
toslunar, line 11: `_pytest.outcomes` is not available unless there is `import pytest`.
```suggestion
import _pytest.outcomes
```

toslunar (changes requested):

cupy/testing/helper.py
toslunar: `import cupy`, which makes `cupy.testing.helper` available for some reason, should not fail with newer PyTest (than the version of the test requirement: `pytest<4.2`).  It would be safer either
- `import _pytest.outcomes` inside `for_dtypes`, or
- `try: import _pytest.outcomes except: pass`.

However it does not seem an immediate issue, because I confirmed that `_pytest.outcomes` is still available in `pytest==6.1.1`.

toslunar: Sorry.  First of all, `pytest` does not have to be installed.  Could you follow `cupy.testing.attr`?
https://github.com/cupy/cupy/blob/0a8b1104aae25ee5973c126f3634b54a1bf9622e/cupy/testing/attr.py#L5-L21

takagi: I fixed as such. Would you take a look again?

toslunar (commented):
I'm fine with removing `unittest.skipTest` from CuPy's test codes.

@asi1024 Is it `no-compat`?

asi1024:
I think we should support both `raise unittest.SkipTest()` and `pytest.skip()` for compatibility.

cupy/testing/helper.py
```diff
@@ -703,11 +721,12 @@ def for_dtypes(dtypes, name='dtype'):
     def decorator(impl):
         @functools.wraps(impl)
         def test_func(self, *args, **kw):
+            check_available()
```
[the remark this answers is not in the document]
takagi: It definitely makes sense. I'll fix it.

takagi:
> I think we should support both raise unittest.SkipTest() and pytest.skip() for compatibility.

Okay, then I'll fix this PR not to remove `unittest.SkipTest` handlers.

cupy/testing/helper.py
```diff
@@ -8,6 +8,12 @@
 from unittest import mock
 import warnings
 
+try:
+    import _pytest.outcomes
+    _error = None
+except ImportError as e:
+    _error = e
```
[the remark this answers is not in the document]
takagi: Sounds good. I didn't know `except` clause evaluates the second token...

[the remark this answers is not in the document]
takagi: I use tuple as list is not allowed.

toslunar (approved):

toslunar:
Jenkins, test this please.

takagi:
I've fixed numpy/cupy comparison test decorators to support `pytest.skip()` as well, which was the cause of the CI failure.

asi1024:
Jenkins, test this please.

asi1024:
@takagi Could you check Jenkins failure?

takagi:
It looks strange that `cupy-py3` only fails...

takagi:
I've found that the combination test chose `uint16` and `float32`, which return `True` for `np.can_cast`, while making positive mask updated `uint16` to `intp`, aka. `int64`, which returns `False` for `np.can_cast` with `float32` and caused the CI failure.

asi1024:
Jenkins, test this please.

cupy/testing/helper.py
```diff
@@ -17,13 +17,27 @@
...
+    exceptions = Exception,
+    if _is_pytest_available:
+        exceptions += _pytest.outcomes.Skipped,
+
     try:
         result = impl(self, *args, **kw)
         error = None
         tb = None
-    except Exception as e:
+    except exceptions as e:
```
toslunar, line 40: The change LGTM.  It's better than `except BaseException`.

toslunar (approved):

asi1024:
LGTM!

PR #3857, opened 2020-08-25
Remove broken support for macOS

leofang:
It has been broken for quite some time (ex: https://github.com/cupy/cupy/issues/952#issuecomment-491635935), and CUDA has dropped support for Mac, so there's no reason to keep them around.

PR #4099, opened 2020-10-08
ROCm: Support RCCL

### LINKED ISSUE #3927: ROCm: Support RCCL
RCCL is the ROCm's counterpart for NCCL:
https://github.com/ROCmSoftwarePlatform/rccl

leofang:
Close #3927. All tests in `tests/cupy_tests/cuda_tests/test_nccl.py` passed locally, but I only have 1 GPU, so we might need further tests to ensure it's working with multi-GPUs.

This works because RCCL uses exactly the same API names as in NCCL, so we don't even need a shim layer to wrap the names!

### DIFF
--- a/install/build.py
+++ b/install/build.py
@@ -155,7 +155,10 @@ def get_compiler_setting(use_hip):
         extra_compile_args.append('-std=c++11')
 
     if PLATFORM_DARWIN:
-        library_dirs.append('/usr/local/cuda/lib')
+        if cuda_path:
+            library_dirs.append('/usr/local/cuda/lib')
+        elif rocm_path:
+            library_dirs.append('/opt/rocm/lib')
 
     if PLATFORM_WIN32:
         nvtoolsext_path = os.environ.get('NVTOOLSEXT_PATH', '')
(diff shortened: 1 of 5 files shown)


leofang:
cc: @emcastillo

leofang:
Jenkins, test this please

install/build.py
```diff
@@ -155,7 +155,10 @@ def get_compiler_setting(use_hip):
         extra_compile_args.append('-std=c++11')
 
     if PLATFORM_DARWIN:
-        library_dirs.append('/usr/local/cuda/lib')
+        if cuda_path:
+            library_dirs.append('/usr/local/cuda/lib')
+        elif rocm_path:
+            library_dirs.append('/opt/rocm/lib')
```
leofang, line 161: Oops, it's Darwin...I think these lines will be removed in #3857.

leofang:
Fixed a bug in f174fc2 that I wasn't able to catch because I have a CUDA installation on the development machine.

leofang:
Jenkins, test this please

install/build.py
```diff
@@ -404,22 +407,29 @@ def check_nccl_version(compiler, settings):
...
+                            #include <rccl.h>
+                            #endif
+                            #include <stdio.h>
+                            #ifdef NCCL_MAJOR
+                            #ifndef NCCL_VERSION_CODE
+                            #  define NCCL_VERSION_CODE \
+                            (NCCL_MAJOR * 1000 + NCCL_MINOR * 100 + NCCL_PATCH)
+                            #endif
+                            #else
+                            #  define NCCL_VERSION_CODE 0
```
emcastillo, line 424: with rccl we will have the 0?

leofang: No, because `NCCL_VERSION_CODE`, `NCCL_MAJOR`, etc are all defined in `rccl.h`.

emcastillo: LOL they are defined as `NCCL` instead of `RCCL`?

leofang: That's right....

emcastillo (approved):

PR #4110, opened 2020-10-11
ROCm: Build on latest ROCm  

leofang:
The symlinks `libhiprtc.so` and `libhip_hcc.so` are gone starting ROCm 3.7.0. We should just use the actual shared library that they point to. I tested locally that this is backward compatible with 3.5.0.

This PR also changes the Debian repo to the latest one so that the pfnCI can always target the latest ROCm release.


leofang:
CI passed, nice. We should mark this as a bug fix and backport it.

leofang:
Jenkins, test this please

emcastillo (approved):

emcastillo (approved):

PR #4135, opened 2020-10-14
Add `cupy.vectorize`

asi1024:
First step of #3392.

This PR adds `cupy.vectorize`.
Currently, only primitive arithmetic operations are supported, but I'm planning to cover more features in the future.

Future works:
- `vectorize.__init__` options: (`otypes`, `excluded`, ... etc.)
- Compile-time constants
- python math library / cupy ufunc function calls.
- Control statements (for, while, ... etc.)

### DIFF
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -292,7 +292,7 @@ def _transpile_expr(expr, env):
         return _emit_cuda_object_from_constants(expr.value, env)
     if isinstance(expr, ast.Num):
         # Deprecated since py3.8
-        return _emit_cuda_object_from_constants(expr.n)
+        return _emit_cuda_object_from_constants(expr.n, env)
     if isinstance(expr, ast.Subscript):
         # # TODO(asi1024): Fix.
         # value = _transpile_expr(expr.value, env)
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -1,0 +1,323 @@
+import ast
+import inspect
+
+import numpy
+
+from cupyx.jit import _types
+from cupyx.jit import _typerules
+
+
+def transpile(func, attributes, mode, in_types, ret_type):
+    """Transpile the target function
+    Args:
+        func (function): Target function.
+        attributes (list of str): Attributes of the generated CUDA function.
+        mode ('numpy' or 'cuda'): The rule for typecast.
+        in_types (list of _types.TypeBase): Types of the arguments.
+        ret_type (_types.TypeBase or None): Type of the return value.
+    """
+
+    if not callable(func):
+        raise ValueError('`func` must be a callable object.')
+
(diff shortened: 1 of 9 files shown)


tests/cupy_tests/functional_tests/test_vectorize.py
```diff
@@ -0,0 +1,256 @@
...
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_allclose()
+    def test_vectorize_augassign(self, xp, dtype):
+        def my_augassign(x):
+            x += x
+            return x + x
+
+        f = xp.vectorize(my_augassign)
+        x = testing.shaped_random((20, 30), xp, dtype, seed=1)
+        return f(x)
```
emcastillo, line 256: How about testing functions with multiple operations and constants?
`return x+x*x+2` or something similar so we can get an ast with a bit of depth :)

cupyx/jit/_compile.py
```diff
@@ -0,0 +1,323 @@
...
+from cupyx.jit import _typerules
+
+
+def transpile(func, attributes, mode, in_types, ret_type):
+    """Transpile the target function
+    Args:
+        func (function): Target function.
+        attributes (list of str): Attributes of the generated CUDA function.
+        mode ('numpy' or 'jit'): The rule for typecast.
+        in_types (list of _types.TypeBase): Types of the arguments.
```
emcastillo, line 16: I wonder what happens if we have constants in the middle of the function such as `2`. How do we know if its an int32 or int64? I guess that the safest way is to declare everything as int64 to avoid overflow unless there are some kind of type hints

asi1024: Current implementation has 2 types of typerule: `numpy`-mode and `cuda`-mode, and `numpy`-mode is chosen in `cupy.vectorize`. https://github.com/cupy/cupy/blob/13b5bb57c341c8dc4733b6e1c675f462de303d6a/cupyx/jit/_typerules.py#L82-L102
In `numpy`-mode, `int` is always handled as `int64` type. In `cuda`-mode, handled as `int32` only if the value is small enough.

emcastillo: cool!

leofang: This is very nice. Numba defaults to 64 bits whenever possible, causing a lot of performance issues that need to be worked around manually by users, and IIUC they have no way to improve it as it wasn't part of original design consideration. It's critical to take this into account in the early stage as done here.

emcastillo:
Jenkins, test this please

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

emcastillo (approved):
LGTM

emcastillo:
We should add docs to this, but let's merge it now

PR #4143, opened 2020-10-16
Fix `generate_matrix` for numpy<1.16

toslunar:


leofang:
I thought we've already been on NumPy>=1.16 as per NEP-29?
https://numpy.org/neps/nep-0029-deprecation_policy.html

emcastillo:
We discussed at the daily that it might be better to drop old numpy versions in the combination test
https://github.com/chainer/chainer-test/pull/608

PR #4144, opened 2020-10-16
DOC: Add note about starting nvprof with profiling off

carterbox:
It wasn't immediately obvious to me that I needed to start `nvprof` with profiling turned off in order only profile only the code inside the profiling context, so I added this note to the documentation to help Python users who may not be familiar with profiling compiled languages.


kmaehashi:
Thanks! LGTM 👍🏼 
pfnCI, test this please.

PR #4149, opened 2020-10-19
Fix tests that have side effects

takagi:
Rel #4138.

Currently, `NotificationTestBase` keeps the value of `_ufunc_config._config` and restores it on tearing down. However, the value of `_config` is an instance that is destructively changed and we need to have another object to properly restore its original contents.


kmaehashi:
pfnCI, test this please.

PR #4150, opened 2020-10-19
Fix cupyx.seterr() when `linalg` not supplied

takagi:
Rel #4138.

Currently, `cupyx.seterr()` unintentionally set `None` to `_config.linalg` when `linalg` argument is not supplied. This PR fixes it to leave as it is in such a case.

### DIFF
--- a/tests/cupyx_tests/test_cupyx.py
+++ b/tests/cupyx_tests/test_cupyx.py
@@ -13,10 +13,9 @@ class TestErrState(unittest.TestCase):
         orig = cupyx.geterr()
         with cupyx.errstate(divide=self.divide):
             state = cupyx.geterr()
-            assert state['divide'] == self.divide
-            for key in state:
-                if key != 'divide':
-                    assert state[key] == orig[key]
+            assert state.pop('divide') == self.divide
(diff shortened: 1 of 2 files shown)


tests/cupyx_tests/test_cupyx.py
```diff
@@ -10,9 +10,13 @@
 class TestErrState(unittest.TestCase):
 
     def test_errstate(self):
+        orig = cupyx.geterr()
         with cupyx.errstate(divide=self.divide):
             state = cupyx.geterr()
             assert state['divide'] == self.divide
+            for key in state:
+                if key != 'divide':
+                    assert state[key] == orig[key]
```
asi1024, line 19: How about simplifying as follows?
```suggestion
            assert state.pop('divide') == self.divide
            assert state == orig
```

takagi: Looks good. I add `orig.pop('divide')` as `orig` also has `divide` key and we need to remove it to assert.

kmaehashi:
pfnCI, test this please.

asi1024:
LGTM.

PR #4151, opened 2020-10-19
fix broadcasting behavior in ndimage.measurements functions

grlee77:
The `cupyx.scipy.ndimage.measurements` functions `sum`, `mean`, `variance` and `standard_deviation` currently do not broadcast the labels argument properly in all cases.

The call to `cupy.broadcast_arrays` just needs to precede any calls like `input[labels > 0]` or `input[labels == index]` as proposed here.

The test cases added here fail on master, but pass after the changes in this PR.


emcastillo (approved):

emcastillo:
Jenkins, test this please

PR #4157, opened 2020-10-20
fix: respect user-supplied output array in all binary morphology functions

grlee77:
This PR fixes a one line bug fix in `binary_erosion`. The current code was causing the output to always be stored in a new array even when the user supplied an output array. Test cases using a user-provided array were added.


emcastillo (approved):

emcastillo:
Jenkins, test this please

Issue #4165, opened 2020-10-22
`nonzero` is wrong for nonzero 0d-array


toslunar:
```
>>> import numpy, cupy
>>> for xp in [numpy, cupy]: print(xp.__name__, xp.__version__, xp.array(3, 'i').nonzero())
...
<stdin>:1: DeprecationWarning: Calling nonzero on 0d arrays is deprecated, as it behaves surprisingly. Use `atleast_1d(arr).nonzero()` if the old behavior was intended.
numpy 1.19.2 (array([0]),)
<stdin>:1: DeprecationWarning: calling nonzero on 0d arrays is deprecated
cupy 8.0.0rc1 (array([], dtype=int64),)
```

https://github.com/cupy/cupy/blob/a8dfcd66d89c8e66a60e4b7272f95a15c26fc907/cupy/core/_routines_indexing.pyx#L58-L61

PR #4169, opened 2020-10-22
Use `testing.assert_warns`

toslunar:
- It is not very good idea to assume `filterwarnings` in `[tool:pytest]` section contains `error::DeprecationWarning`.  See chainer/chainer#7824 for the previous discussion.
- It seems better to check results even if they are warned.  In fact, 46e5443be2c6de60a15c31f6b4e46866335a9a2a could have prevented the regression #4165.


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

Issue #4163, opened 2020-10-22
Add `cupy.random.BitGenerator` and `cupy.random.Generator`


emcastillo:
In order to match NumPy's new random API, these mechanisms need to be added.

There are two main issues that I am currently working on:
1st, curand, and NumPy random generators are not equivalent.
NumPy supports
- PCG64 (PCG XSL RR 128/64) 128 bit state/ 64 bit output
- MT19937
- Philox 4x64
- SFC64

And the only one available in curand is mt19937.
Moreover, in the current `cupy.random.RandomState`, when distributions are generated, in some of them the specified curand generator is ignored, and a custom xor based random bits generator is used in the device code. 

To draw samples from a distribution, most of the kernels rely on a sample-rejection mechanism, so in order to consistently use the user-specified random generator inside these kernels, curand generators need to be device callable. But curand does not work with NVRTC (@anaruse any hint about this?). We need to create an additional .cu file with the distributions and mechanism to draw samples from either curand generators or custom ones inside the kernels generating the distributions.

toslunar:
See also #1431.

leofang:
Cool, didn't know there's device API!

PR #4177, opened 2020-10-26
Add NumPy v1.17 new random API with `Generator`

emcastillo:
This is still on-going

Due to some issues when generating the distributions in #4163, the added bit generators will use only the curand device API

I have spent a great deal of time trying to figure out how to correctly templatize the code, some ideas will be greatly appreciated. The idea is that we want multiple curand generators (or custom ones) to be supported in the kernels that implement the distributions. The generators are usually specified in the `BitGenerator` in python so we need to get that type info down to the Cython level.

TODO
- HIP
- Add minimum set of distributions (normal, uniform, exponential)
- Support all device curand generators
- Tests


emcastillo:
I tried to design some templatized mechanism for the kernels to allow any type of generator obbeying the API,
However there are several limitations that made this approach not feasible at the first try.

The idea was that the `BitGenerators` in python have a method that returns a wrapped instance of their C generator.
But at the time of calling the templatized module, the wrapped instance is always a PyObj because we can't determine the type at compile time since this is doing several kinds of polimorphism. So in the end the correct template can't be invoked unless the object is explicitly casted to the correct type, which is something I wanted to avoid.

cupy_setup_build.py
```diff
@@ -302,6 +301,20 @@
         })
 
 
+MODULES.append({
+    'name': 'random',
+    'file': [
+        ('cupy.random._bit_generators', ['cupy/random/cupy_distributions.cu']),
```
kmaehashi, line 307: How about adding this to `cuda` module? (around line 99)

emcastillo: Yes, this is all WIP so I will be changing a lot of places where things are declared :)

emcastillo:
Added an `integers` naive implementation that takes 25 us to generate 10.000 elements compared to the 500 us of the current `randint`

cupy/random/cupy_distributions.cu
```diff
@@ -0,0 +1,211 @@
...
+
+struct beta_functor {
+    template<typename... Args>
+    __device__ double operator () (Args&&... args) {
+        return rk_beta(args...);
+    }
+};
+
+// There are several errors when trying to do this a full template
+struct exponential_functor {
```
emcastillo, line 169: Tried to create a single template for the functor, but since it is being used inside other templates there were a lot of conflicts with argument types and other stuff so I gave up.

cupy/random/cupy_distributions.cu
```diff
@@ -0,0 +1,211 @@
...
+        random.init_state(id, param);
+        ((R*) out)[id] = func(&random, args...);
+    }
+    return;
+}
+
+void interval_32(intptr_t param, int mx, int mask, void* out, ssize_t size) {
+    int tpb = 256;
+    int bpg =  (size + tpb - 1) / tpb;
+    execute_dist<interval_32_functor, curand_pseudorand_state, int32_t><<<bpg, tpb>>>(param, out, size, mx, mask);
```
emcastillo, line 191: These are entry points from cython, I am still wondering how to generalize them for different random generators without using an enum and tons of ifs. But with templates, I couldn't get Cython to get the right C object to call the correct template. Any clue about this @asi1024 @leofang ?

cupy/random/cupy_distributions.cu
```diff
@@ -0,0 +1,211 @@
...
+// There are several errors when trying to do this a full template
+struct exponential_functor {
+    template<typename... Args>
+    __device__ double operator () (Args&&... args) {
+        return rk_standard_exponential(args...);
+    }
+};
+
+template<typename F, typename T, typename R, typename... Args>
+__global__ void execute_dist(intptr_t param, void* out, ssize_t size, Args... args) {
```
emcastillo, line 177: since all the distributions use the same kernel, this avoids the need to write redundant kernels, however If there was an easy way to templatize the functors too, it would be awesome.

emcastillo:
Btw, randint with this is super fast, around 26 us to generate 10.000 random integers compared to 500 us of current cupy implementation

cupy/random/cupy_distributions.cu
emcastillo: There is the issue that the memory pointer is in the device so we need to do stuff like `cudaMemcpyFromSymbol` which is not versatile at all

cupy/random/cupy_distributions.cu
```diff
@@ -0,0 +1,252 @@
...
+        F func;
+        out_ptr[id] = func(&random, std::forward<Args>(args)...);
+    }
+    return;
+}
+
+template <typename F, typename R>
+struct kernel_launcher {
+    template<typename T, typename... Args>
+    void operator()(int bpg, int tpb, cudaStream_t stream, Args&&... args) {
```
emcastillo, line 213: we can't have kernels as function members, so we have to resort to dirty tricks like this ...

cupy/random/cupy_distributions.cu
```diff
@@ -0,0 +1,252 @@
...
+        return curand_normal(_state);
+    }
+};
+
+
+// This design is the same as the dtypes one
+template <typename F, typename... Ts>
+void generator_dispatcher(int generator_id, F f, int bpg, int tpb, cudaStream_t stream, Ts&&... args) {
+   switch(generator_id) {
+       case CURAND_XOR_WOW: return f.template operator()<curand_xor_state>(bpg, tpb, stream, std::forward<Ts>(args)...);
```
emcastillo, line 50: Data comes from python, so we lose it in the inheritance hierarchy of BitGenerator :( and this makes imposible to directly instatiate a template. We have to use an enum like this.

cupy/random/cupy_distributions.cu
```diff
@@ -0,0 +1,261 @@
...
+        }
+    } else {
+        Ga = rk_standard_gamma(state, a);
+        Gb = rk_standard_gamma(state, b);
+        return Ga/(Ga + Gb);
+    }
+}
+
+
+__device__ uint32_t rk_interval_32(rk_state* state, int mx, int mask) {
```
emcastillo, line 159: fix int types here

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+        cdef int generator = self._bit_generator.generator
+
+        cdef bsize = self._bit_generator.state_size()
+        if bsize == 0:
+            func(state, y_ptr, out.size, strm, *args)
+        else:
+            chunks = (out.size + bsize - 1) // bsize
+            for i in range(chunks):
+                y_ptr = <intptr_t>out[i*bsize:].data.ptr
+                func(generator, state, y_ptr, bsize, strm, *args)
```
leofang, line 183: What happens for the last chunk that might be smaller than `bsize`?

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+        cdef intptr_t strm = stream.get_current_stream_ptr()
+        state_ptr = self._bit_generator.state()
+        cdef state = <intptr_t>state_ptr
+        cdef y_ptr = <intptr_t>out.data.ptr
+        cdef ssize_t size = out.size
+        cdef int generator = self._bit_generator.generator
+
+        cdef bsize = self._bit_generator.state_size()
+        if bsize == 0:
+            func(state, y_ptr, out.size, strm, *args)
```
leofang, line 178: Why don't we need `generator` as the first argument here?

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+
+cdef extern from 'cupy_distributions.cuh' nogil:
+    cppclass curandState:
+        pass
+    cppclass curandStateMRG32k3a:
+        pass
+    cppclass curandStatePhilox4_32_10_t:
+        pass
+
+    cdef enum _RandGenerators '::RandGenerators':
```
leofang, line 21: Do we need the scope resolution operator?!
```suggestion
    cdef enum _RandGenerators 'RandGenerators':
```

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+        CURAND_XOR_WOW
+        CURAND_MRG32k3a
+        CURAND_PHILOX_4x32_10
+
+    void init_curand_generator(int generator, intptr_t state_ptr, uint64_t seed, ssize_t size, intptr_t stream);
+    void interval_32(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream, int mx, int mask);
+    void interval_64(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream, uint64_t mx, uint64_t mask);
+    void beta(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream, double a, double b);
+    void exponential(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream);
```
leofang, line 30: I think you don't need the tailing semicolons.

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+        else:
+            raise ValueError(
+                'high - low must be within uint64 range (actual: {})'.format(diff))
+
+        y = ndarray(size if size is not None else (), dtype)
+
+        if dtype is numpy.uint32:
+           self._launch_distribution_kernel(interval_32, y, diff, mask)
+        else:
+           self._launch_distribution_kernel(interval_64, y, diff, mask)
```
leofang, line 135: Nitpick: Seems like `interval_32`/`interval_64` are passed around as Python objects, despite that they're in fact C functions. If this is valid, perhaps we can just merge this part with the previous dtype check:
```python
if diff <= _UINT32_MAX:
    dtype = numpy.uint32
    func = interval_32
elif diff <= _UINT64_MAX:
    dtype = numpy.uint64
    func = interval_64
else:
    raise ...

y = ndarray(...)
self._launch_distribution_kernel(func, y, diff, mask)
```

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+            y = out
+        return y
+
+    def _launch_distribution_kernel(self, func, out, *args):
+        # The generator might only have state for a few number of threads,
+        # what we do is to split the array filling in several chunks that are
+        # generated sequentially using the same state
+        cdef intptr_t strm = stream.get_current_stream_ptr()
+        state_ptr = self._bit_generator.state()
+        cdef state = <intptr_t>state_ptr
```
leofang, line 171: nitpick:
```suggestion
        cdef intptr_t state = <intptr_t>self._bit_generator.state()
```

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+    void interval_32(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream, int mx, int mask);
+    void interval_64(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream, uint64_t mx, uint64_t mask);
+    void beta(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream, double a, double b);
+    void exponential(int generator, intptr_t state, intptr_t out, ssize_t size, intptr_t stream);
+
+_UINT32_MAX = 0xffffffff
+_UINT64_MAX = 0xffffffffffffffff
+
+
+class BitGenerator:
```
leofang, line 36: Maybe we can also hide this class (rename to `_BitGenerator`)?

leofang: Looks like NumPy exposes this object, but given that it's a base class which is useless without subclassing, I wonder how useful it is...

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+        return y
+
+    def _launch_distribution_kernel(self, func, out, *args):
+        # The generator might only have state for a few number of threads,
+        # what we do is to split the array filling in several chunks that are
+        # generated sequentially using the same state
+        cdef intptr_t strm = stream.get_current_stream_ptr()
+        state_ptr = self._bit_generator.state()
+        cdef state = <intptr_t>state_ptr
+        cdef y_ptr = <intptr_t>out.data.ptr
```
leofang, line 172: `cdef intptr_t y_ptr = ...`

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+        cdef ssize_t size = out.size
+        cdef int generator = self._bit_generator.generator
+
+        cdef bsize = self._bit_generator.state_size()
```
leofang, line 176: What's the expected type of `bsize`?

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,184 @@
...
+        return self._size
+
+    def _type_size(self):
+        return 0 
+
+class XORWOW(_cuRANDGenerator):
+    generator = CURAND_XOR_WOW  # Use The Enum
+
+    def _type_size(self):
+        return sizeof(curandState)
```
leofang, line 93: I am a bit confused: what's the difference between `curandState` and `curandState_t`?

leofang (commented):
This looks good! Just did a quick pass and left some 1st round comments.

cupy/random/_bit_generator.pyx
emcastillo: LOL

emcastillo: I am wondering why it worked....

cupy/random/_bit_generator.pyx
emcastillo: good catch!

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,186 @@
+import threading
```
leofang, line 1: I think this should fix Travis error
```suggestion
# distutils: language = c++

import threading
```

leofang (changes requested):

cupy/random/_bit_generator.pyx
emcastillo: `ssize_t`

cupy/random/_bit_generator.pyx
emcastillo: no one knows, they are used interchangeably in the documentation, but we should stick to only one, will fix

cupy/random/_bit_generator.pyx
leofang: Looks like Travis has a hard time understanding `func = void (*)(...)` 😂

leofang:
87e959a fixed the C++ error, but we still have errors due to using variadic arguments (https://github.com/cupy/cupy/pull/4177#discussion_r515746465).

emcastillo:
Would like some feedback on the new design :)

tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -1206,214 +257,7 @@ def test_randint_int64_1(self):
...
-class TestTriangular(RandomGeneratorTestCase):
-
-    target_method = 'triangular'
-
-    def test_triangular(self):
-        self.generate(
-            left=self.left, mode=self.mode, right=self.right, size=(3, 2))
-
-
+"""
```
emcastillo, line 260: will remove this

leofang (changes requested):
Very quick first-pass scan, mostly on style. Will take a deeper look later.

I don't fully understand the design decision, likely because I'm not familiar with either NumPy's or cuRAND's API, will look into them to understand things better.

I noticed some classes in `.pyx` are plain Python classes. Given the simplicity of data management here, I am not too worried about not turning them into cdef classes. But it'd be good to have a benchmark done. Is it possible to compare random number generation throughput before and after this PR?

Based on offline discussion with @emcastillo we think it's straightforward to add HIP support. Given the massive amount of changes in this PR I suggest we defer it (I could try to handle it in a separate PR).

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,185 @@
...
+
+    def _state_size(self):
+        """Maximum number of samples that can be generated at once
+        """
+        return 0
+
+    def _check_device(self):
+        if cupy.cuda.Device().id != self._current_device:
+            raise RuntimeError(
+                'This Generator state is allocated in a different device')
```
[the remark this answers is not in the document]
emcastillo: This should error, users shouldn't be able to use a generator with its state allocated in one device, in another device.

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,185 @@
...
+
+
+class _cuRANDGenerator(BitGenerator):
+    # Size is the number of threads that will be initialized
+    def __init__(self, seed=None, *, size=10000*256):
+        super().__init__(seed)
+        # Raw kernel has problems with integers with the 64th bit set
```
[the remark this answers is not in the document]
emcastillo: When passing a 64-bit unsigned integer as a parameter to a raw kernel, it is cast somewhere in the code to a signed one using `numpy.int64`, if that integer has its 64-th bit set, then an Out of range error is raised 😔

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,185 @@
...
+    Args:
+        seed (int, array_like[ints], numpy.random.SeedSequence, optional):
+            A seed to initialize the `BitGenerator`. If None, then fresh,
+            unpredictable entropy will be pulled from the OS. If an ``int`` or
+            ``array_like[ints]`` is passed, then it will be passed to
+            ~`numpy.random.SeedSequence` to derive the initial `BitGenerator`
+            state. One may also pass in a `SeedSequence` instance.
+    """
+    def __init__(self, seed=None):
+        self.lock = threading.Lock()
```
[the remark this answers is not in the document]
emcastillo: Yeah, it is actually documented in the numpy API, but I believe we don't need it, so I can get rid of it.

cupy/random/_distributions_module.py
```diff
@@ -0,0 +1,255 @@
...
+    __device__ virtual double rk_double() {
+        return  0.0;
+    }
+    __device__ virtual double rk_normal() {
+        return  0.0;
+    }
+};
+template<typename CURAND_TYPE>
+struct curand_pseudo_state: rk_state {
+    // Valid for  XORWOW and MRG32k3a
```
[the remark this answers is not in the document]
emcastillo: good catch

cupy/random/_distributions_module.py
```diff
@@ -0,0 +1,255 @@
...
+    CURAND_TYPE* _state;
+    int _id;
+    __device__ curand_pseudo_state(int id, intptr_t state) {
+        _state = reinterpret_cast<CURAND_TYPE*>(state) + id;
+        _id = id;
+    }
+    __device__ virtual uint32_t rk_int() {
```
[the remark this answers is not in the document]
emcastillo: thanks!, will do this :)

cupy/random/_bit_generator.pyx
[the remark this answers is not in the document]
asi1024: Maybe `self.lock` should be used in `random_raw` of subclasses.

cupy/random/_distributions_module.py
```diff
@@ -0,0 +1,259 @@
...
+    }
+    return sampled;
+}
+
+// There are several errors when trying to do this a full template
+// THIS CAN BE A PYTHON TEMPLATE
+struct exponential_functor {
+    template<typename... Args>
+    __device__ double operator () (Args&&... args) {
+        return rk_standard_exponential(std::forward<Args>(args)...);
```
asi1024, line 157: No need to forward because `args` are rvalue reference.
```suggestion
        return rk_standard_exponential(args...);
```

cupy/random/cupy_distributions.cu
asi1024: Is this issue already resolved?

emcastillo: yes, by using jitify we could solve this :)

asi1024:
Jenkins, test this please.

cupy/random/_bit_generator.pyx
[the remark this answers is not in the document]
asi1024: The user's custom generator class may require `self.lock`, so let's leave this interface!

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

leofang:
Let me know if you need me to jump on the HIP support @emcastillo!

emcastillo:
Jenkins, test this please

leofang:
So there're a couple of issues in this PR presently that took me one full day to sort out... Three different but related problems:
1. Jitify could be deceived by NVRTC and locates a wrong header due to bad NVRTC behavior [*]
2. cuRAND headers are not (and cannot be) distributed
3. Jitify does not work on HIP due to hipRTC issues

I suggest we instead use NVCC to compile the cuRAND device kernels:
   - Pros: avoid Jitify issues as it always finds the correct header; works for HIP (by using HIPCC)
   - Cons: might have slightly increased overhead; does not work if CuPy is installed via Conda-Forge *and* there is no local, matching CUDA installation that provides either NVCC or the headers at runtime; unclear if this works on Windows or not

I've been thinking to package cuRAND headers for Jitify+NVRTC, like what we've done for CUB and Jitify, but I don't think EULA permits it...

[*] I encountered a very subtle bug that leads to `cudaErrorIllegalAddress` when testing the MRG32k3a bit generator. It turns out I got mismatching sizes for the struct `curandStateMRG32k3a` on device (72) and on host (48). I identify it has the same root cause as in NVIDIA/jitify#81. 

Here's what happens, bear with me: First it's about my dev env. I have multiple CUDA installed in my env, 10.0 via package manager and other versions via runfile. Ubuntu is very annoying that it scatters CUDA’s stuff all over the place, so for example the 10.0 headers are put in `/usr/include` instead of the standard `/usr/local/cuda-10.0/include`. I've been doing development in 10.2 lately (installed in the default path `/usr/local/cuda-10.2/`), and it usually works just fine. 

Second, the struct members of `curandStateMRG32k3a` have changed between 10.0 and 10.2. Usually this is not a concern as they're implementation details that end users need not to care. 

But, when compiling the device kernel using Jitify,  NVRTC will by default search `/usr/include` (a behavior out of users' control, see https://github.com/NVIDIA/jitify/issues/81#issuecomment-726818514) so it sees the header from 10.0, which has struct size 72 for `curandStateMRG32k3a`. But when we got `sizeof(curandStateMRG32k3a)` from the Cython module, it was correctly compiled on host against CUDA 10.2, so I got the size 48 which is in fact correct.

For ordinary users with a single CUDA installation, the above bug would not happen and Jitify+NVRTC seems to be just fine, but for advanced users or developers it would be very hard to debug if it happens.

emcastillo:
The main problem that we have with nvcc is that it is not installed in some environments.
I think that hipcc for hip its fine, but for nvidia, we should add a kernel at startup that retrieves the correct struct sizes from the device.

leofang:
Right, as I said conda environments would not have nvcc available. But it's not just about the struct size, it's about the whole header (including the device kernel implementations) is off. Maybe if we consistently use the "wrong header" it would work out fine?

emcastillo:
The alternative is to go back to the dispatch templates 😂

leofang:
If it's possible then yes, definitely an option. What was the reason we chose Jitify instead of the dispatcher approach?

emcastillo:
Much more cleaner code, and no need to have a `module` for random

emcastillo:
the lack of redistributable curand headers motivated us to restore the old module-based approach.
@asi1024 @leofang PTAL!

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,183 @@
...
+        # TODO(ecastill) port SeedSequence
+        if isinstance(seed, numpy.random.SeedSequence):
+            self._seed_seq = seed
+        else:
+            if isinstance(seed, cupy.ndarray):
+                seed = cupy.asnumpy(seed)
+            self._seed_seq = numpy.random.SeedSequence(seed)
+        dev = cupy.cuda.Device()
```
asi1024, line 52: How about using `cupy.cuda.get_device_id` to avoid instantiating `Device` object?

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,183 @@
...
+        self._current_device = dev.id
```
asi1024, line 53: I feel `self._device_id` is better for readability.

cupy/random/_bit_generator.pyx
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, seed=None, *, size=1000*256):
+        super().__init__(seed)
+        # Raw kernel has problems with integers with the 64th bit set
+        self._seed = self._seed_seq.generate_state(1, numpy.uint32)[0]
+        self._size = size
+        cdef uint64_t b_size = self._type_size() * size
+        self._state = cupy.zeros(b_size, dtype=numpy.int8)
+        ptr = self._state.data.ptr
+        cdef intptr_t state_ptr = <intptr_t>ptr
+        cdef uint64_t c_seed = <uint64_t>self._seed
```
asi1024, line 81: Unused variables.

cupy/random/_generator.pyx
```diff
@@ -0,0 +1,226 @@
...
+            raise ValueError(
+                f'high - low must be within uint64 range (actual: {diff})')
+
+        y = ndarray(size if size is not None else (), pdtype)
+
+        if dtype is numpy.uint32:
+            _launch_dist(self.bit_generator, interval_32, y, (diff, mask))
+        else:
+            _launch_dist(self.bit_generator, interval_64, y, (diff, mask))
+        return (lo + y).astype(dtype)
```
asi1024, line 122: Could you use `cupy.add` to reduce memory usage?
```suggestion
        return cupy.add(y, lo, dtype=dtype)
```

cupy/random/_generator.pyx
```diff
@@ -0,0 +1,226 @@
...
+            cupy.ndarray: Samples drawn from the beta distribution.
+
+        .. seealso::
+            :meth:`numpy.random.Generator.beta
+            <numpy.random.generator.Generator.beta>`
+        """
+        cdef ndarray y
+        y = ndarray(size if size is not None else (), numpy.float64)
+        _launch_dist(self.bit_generator, beta, y, (a, b))
+        return y.astype(dtype)
```
asi1024, line 151:
```suggestion
        return y.astype(dtype, copy=False)
```

cupy/random/_generator.pyx
```diff
@@ -0,0 +1,226 @@
...
+
+        if method == 'zig':
+            raise NotImplementedError('Ziggurat method is not supported')
+
+        y = ndarray(size if size is not None else (), numpy.float64)
+        _launch_dist(self.bit_generator, exponential, y, ())
+        if out is not None:
+            out[...] = y
+            y = out
+        return y.astype(dtype)
```
asi1024, line 190: Can this method return the same object as `out` option if typecast is not needed?
```suggestion
        return y.astype(dtype, copy=False)
```

cupy/random/_generator.pyx
```diff
@@ -0,0 +1,226 @@
...
+
+
+def init_curand(generator, state, seed, size):
+    cdef int generator_enum = <int> generator
+    cdef intptr_t state_ptr = <intptr_t>state
+    cdef uint64_t c_seed = <uint64_t>seed
+    cdef uint64_t c_size = <uint64_t>size
+    cdef intptr_t strm = stream.get_current_stream_ptr()
+    init_curand_generator(generator_enum, state_ptr, c_seed, c_size, strm)
```
asi1024, line 199: How about the following one for the simplification?
```suggestion
    init_curand_generator(
        <int>generator,
        <intptr_t>state,
        <uint64_t>c_seed,
        <uint64_t>c_size,
        <intptr_t>stream.get_current_stream_ptr()
    )
```

cupy/random/__init__.py
```diff
@@ -14,6 +16,33 @@ def bytes(length):
...
 
+def default_rng(seed=None):  # NOQA  avoid redefinition of seed
+    """Construct a new Generator with the default BitGenerator (XORWOW).
+
+    Args:
+        seed (int, array_like[ints], numpy.random.SeedSequence, cupy.random.BitGenerator, cupy.random.Generator, optional):
+            A seed to initialize the :class:`cupy.random.BitGenerator`. If 
+            ``None``, then unpredictable entropy will be pulled from the OS.
+            If an ``int`` or ``array_like[ints]`` is passed, then it will be
+            passed to :class:`numpy.random.SeedSequence` to detive the initial
```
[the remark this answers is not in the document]
emcastillo: I am wondering if we should implement `SeedSequence` in cupy as well. What do you think?

cupy/random/_bit_generator.pyx
emcastillo: Good catch! thanks

cupy/random/_generator.pyx
emcastillo: Let me check what numpy does

cupy/random/_generator.pyx
emcastillo: Yes, much cleaner thanks!, I (mistakenly) thought what Cython required an explicit declaration with cdef, but seems that this is enough :)

leofang (commented):
Could you add a comment to explain why this is disabled for HIP? Perhaps you can add the link to https://github.com/ROCm-Developer-Tools/HIP/issues/2186.

cupy/random/_generator.pyx
```diff
@@ -148,7 +148,8 @@ class Generator:
         cdef ndarray y
         y = ndarray(size if size is not None else (), numpy.float64)
         _launch_dist(self.bit_generator, beta, y, (a, b))
-        return y.astype(dtype)
+        to_return = y
```
asi1024, line 151:
```suggestion
```

emcastillo:
PTAL!

asi1024:
Jenkins, test this please.

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

asi1024:
LGTM!

leofang:
Awesome work @emcastillo!!!

PR #4192, opened 2020-10-28
Add pytest backend implementation of `testing.parameterize`

toslunar:
A step to migrate to PyTest-style tests.  

The PR adds `pytest.mark.parametrize` implementation of `cupy.testing.parameterize`.  The pytest implementation is opt-in.  For now, the trigger is to stop inheriting `unittest.TestCase`.

`pytest.mark.parametrize` is fast because it does not generate classes/functions.  With `pytest==4.1.1`, `pytest test_arithmetic.py` was `8432 passed in 27.47 seconds` and will be `8432 passed in 18.72 seconds`.  The difference reflects the speed-up of the collection time (`pytest --collect-only`) 14s -> 3s.

`cupy.testing.helper` decorators (e.g. `numpy_cupy_allclose`, `for_dtypes`) will support non-`unittest.TestCase` test collection.  The change required a hack because the decorators change function signatures, which pytest does not expect.  Other `cupy.testing` functionalities (e.g. `repeat`, `retry`, `fix_random`) still assume `unittest.TestCase` (e.g. `unittest.TestSuite`, `unittest.setUp`), and they are out of scope of the PR.


cupy/testing/_pytest_impl.py
```diff
@@ -0,0 +1,60 @@
...
+
+Reason: {}: {}'''.format(__name__, type(_error).__name__, _error))
+
+
+if is_available():
+
+    class _TestingParameterizeMixin:
+
+        def __repr__(self):
+            return "<{}  parameter: {}>".format(
```
asi1024, line 29: Please use single quotes instead.

asi1024 (commented):
LGTM except a nitpick.

cupy/testing/_pytest_impl.py
toslunar: Sure.  Done in a6e02a63c16c8e63e74923efb4bca84ef62a227b.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #4198, opened 2020-10-28
Fix ndarray binary operations with non-CuPy arrays to return `NotImplemented`

### LINKED ISSUE #4118: Raise `NotImplemented` in binary operations with non-CuPy arrays
Binary operations in CuPy currently raise `TypeError` when `other` is a non-CuPy array, see below:

```python
In [1]: import cupy as cp, dask.array as da

In [2]: a = cp.arange(5)

In [3]: b = da.ones_like(a, shape=(5,))

In [4]: b - a
Out[4]: dask.array<sub, shape=(5,), dtype=int64, chunksize=(5,), chunktype=cupy.ndarray>

In [5]: a - b
<ipython-input-5-09bd029d0285> in <module>
----> 1 a - b

cupy/core/core.pyx in cupy.core.core.ndarray.__sub__()

cupy/core/_kernel.pyx in cupy.core._kernel.ufunc.__call__()

cupy/core/_kernel.pyx in cupy.core._kernel._preprocess_args()

TypeError: Unsupported type <class 'dask.array.core.Array'>
```

It would be useful to instead raise a `NotImplemented` exception as it would allow the `other` array to try to compute that operation, see [`NotImplemented` docs](https://docs.python.org/3/library/constants.html#NotImplemented) for details.

Any thoughts on allowing this in CuPy?

asi1024:
Fix #4118.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -994,92 +994,92 @@ cdef class ndarray:
 
     def __add__(x, y):
         if _should_use_rop(x, y):
-            return y.__radd__(x)
+            return NotImplemented
         else:
             return _math._add(x, y)
 
     def __sub__(x, y):
         if _should_use_rop(x, y):
-            return y.__rsub__(x)
+            return NotImplemented
         else:
             return _math._subtract(x, y)
 
     def __mul__(x, y):
         if _should_use_rop(x, y):
-            return y.__rmul__(x)
+            return NotImplemented
         else:
             return _math._multiply(x, y)
 
     def __matmul__(x, y):
         if _should_use_rop(x, y):
-            return y.__rmatmul__(x)
+            return NotImplemented
         else:
             return _linalg.matmul(x, y)
 
     def __div__(x, y):
         if _should_use_rop(x, y):
-            return y.__rdiv__(x)
+            return NotImplemented
         else:
             return _math._divide(x, y)
 
     def __truediv__(x, y):
         if _should_use_rop(x, y):
-            return y.__rtruediv__(x)
+            return NotImplemented
         else:
             return _math._true_divide(x, y)
 
     def __floordiv__(x, y):
         if _should_use_rop(x, y):
-            return y.__rfloordiv__(x)
+            return NotImplemented
         else:
             return _math._floor_divide(x, y)
 
     def __mod__(x, y):
         if _should_use_rop(x, y):
-            return y.__rmod__(x)
+            return NotImplemented
         else:
             return _math._remainder(x, y)
 
     def __divmod__(x, y):
         if _should_use_rop(x, y):
-            return y.__rdivmod__(x)
+            return NotImplemented
         else:
             return divmod(x, y)
 
     def __pow__(x, y, modulo):
         # Note that we ignore the modulo argument as well as NumPy.
         if _should_use_rop(x, y):
-            return y.__rpow__(x)
+            return NotImplemented
         else:
             return _math._power(x, y)
 
     def __lshift__(x, y):
         if _should_use_rop(x, y):
-            return y.__rlshift__(x)
+            return NotImplemented
         else:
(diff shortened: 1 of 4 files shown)


cupy/core/core.pyx
```diff
@@ -1723,6 +1721,7 @@ cpdef strides_t _get_strides_for_order_K(ndarray x, dtype, shape=None):
 
 
 _HANDLED_TYPES = (ndarray, numpy.ndarray)
+_KNOWN_TYPES = (ndarray, numpy.generic, bool, int, float, complex)
```
jakirkham, line 1724: Would suggest using [`numbers.Number`]( https://docs.python.org/3/library/numbers.html ) for these.

```python
In [1]: import numbers

In [2]: isinstance(True, numbers.Number)
Out[2]: True

In [3]: isinstance(1, numbers.Number)
Out[3]: True

In [4]: isinstance(1.5, numbers.Number)
Out[4]: True

In [5]: isinstance(1.5 + 1j, numbers.Number)
```

```suggestion
_KNOWN_TYPES = (ndarray, numpy.generic, numbers.Number)
```

jakirkham: I think we can also drop `numpy.generic`, but will defer to @seberg

```python
In [1]: import numbers

In [2]: import numpy as np

In [3]: isinstance(np.float32(1), np.generic)
Out[3]: True

In [4]: isinstance(np.float32(1), numbers.Number)
Out[4]: True
```

```suggestion
_KNOWN_TYPES = (ndarray, numbers.Number)
```

seberg: I am not sure any of these actually matter, since they all do not implement `__array_ufunc__` anyway.  One trick numpy does is to skip the `hasattr` check on builtin python types as a micro-optimization. I am not sure how much that actually helps.

jakirkham: Right I think that's what is happening here as well. Also don't know how much it help

seberg: Fair point, in which case you should swap the check and put `hasattr` second?  If this is about overhead/performance, I am not sure that `numbers.Number` is all that great though, I do not think `isinstance` checks are any faster than `hasattr` for ABCs.

asi1024: The overhead seems small ignorable, but I agree with your refactoring suggestions!

takagi:
pfnCI, test this please.

takagi:
Would you check the CI failure? It's about binary operations of sparse matrices.

cupy/core/core.pyx
```diff
@@ -55,12 +56,10 @@ from cupy_backends.cuda.libs cimport cublas
 
 @cython.profile(False)
 cdef inline _should_use_rop(x, y):
-    xp = getattr(x, '__array_priority__', 0)
-    yp = getattr(y, '__array_priority__', 0)
-    return xp < yp and not isinstance(y, ndarray)
+    return not isinstance(y, _KNOWN_TYPES) and hasattr(y, '__array_ufunc__')
```
toslunar, line 59: > Would you check the CI failure? It's about binary operations of sparse matrices.

`cupy.ndarray op cupyx.sparse.spmatrix` should use rop.  `spmatrix` does not have `__array_ufunc__`.

toslunar:
pfnCI, test this please.

toslunar:
pfnCI, test this please.

toslunar:
pfnCI, test this please.

asi1024:
Rebased onto `master`.
Jenkins, test this please.

asi1024:
@takagi PTAL!

asi1024:
@takagi Could you take another look?

takagi:
Sorry, I dropped the mention. Let me read it through shortly!

takagi:
pfnCI, test this please.

takagi:
LGTM!

PR #4213, opened 2020-10-29
Add `SparseEfficiencyWarning` and warn inefficient comparison

toslunar:
`==`, `<=`, and `>=` are inefficient for sparse matrices (because `a.nnz + b.nnz + result.nnz >= row * col`).  This PR makes CuPy warn the inefficiency like SciPy does.

`cupyx.scipy.sparse.SparseEfficiencyWarning` is `scipy.sparse.SparseEfficiencyWarning` if SciPy is available.  Otherwise, CuPy defines a new warning class for it.  It is similar to the case with `numpy.AxisError` until #3096.
https://github.com/cupy/cupy/blob/25b68333154c5aa59750fb6a83ef12f7dc0552f7/cupy/core/_errors.py#L4-L5


kmaehashi:
LGTM!
pfnCI, test this please.

kmaehashi:
LGTM! Test failures are not related to this PR.

PR #875, opened 2018-01-11
Fix to preserve dtype of input array in cupy.linalg.norm

kmaehashi:
In NumPy 1.14, np.linalg.norm has been changed to preserve dtype of input array.
However, as fix made in NumPy 1.14 was incomplete, there are some cases (e.g., when arbitrary order is specified) that dtype is not preserved.
The issue is recognized by the team and the additional fix is ongoing.

* https://github.com/numpy/numpy/issues/10364
* https://github.com/numpy/numpy/pull/10368

This PR makes CuPy behavior consistent with NumPy 1.14.1 (assuming the above PR got merged and released).

We need to discuss if we should separate tests for `<1.14`, `==1.14`, '>1.14'.
(Maybe we can ignore `==1.14` ... but in that case, can we say that we support NumPy 1.14?)

PR #4227, opened 2020-11-04
Fix output dtype of `linalg.norm`

toslunar:
ref: #875


emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4228, opened 2020-11-05
Support Jitify

### LINKED ISSUE #3408: Support Jitify ?

leofang:
Close #3408.

For the jitify idea, please refer to the original issue (https://github.com/cupy/cupy/issues/3408#issue-634131790). Here I adopt the first choice:
> 1. Wrap jitify in `cupy.cuda.jitify`: Ideally, we still use our NVRTC wrapper to do compilation, and only take all these lookup implementations as-is so we don't reinvent the wheel. Unfortunately it is currently not possible, as many logics are baked in jitify's public APIs, and it requires some refactoring on the their side for us to reuse.

It turns out to be possible if we use Jitify's internal function, which performs this compile-fail-search loop for us. Once all headers are identified and collected, we pass them to our own NVRTC workflow and compile as usual. As a result, the amount of code additions/changes is kept very minimal 🙂

With the support of Jitify, we can achieve the following two things: 
1. The CUB block reduction kernel (from `cupy.core._cub_reduction`) can now be compiled using Jitify! It saves us a subprocess call to `nvcc`, and all compilation is done within the same process.
2. `RawKernel`/`RawModule` can now use Jitify to assist compiling C++ kernels by setting `jitify=True` (default is `False` to reduce overhead). In particular, it works with the `name_expressions` option (for template specialization, see #3319) seamlessly as expected. 

***Warning***: certain Thrust functionalities might still be unsupported due to a Jitify bug: https://github.com/NVIDIA/jitify/issues/81. I hope it's a easy upstream fix.

This unfortunately does not support HIP without modifying the Jitify header.

TODO:
- [x] Support RawKernel/RawModule
- [x] Add more tests
- [x] Clean up + refactoring
- [x] bdist/wheel content
- [x] Performance benchmark: https://github.com/cupy/cupy/pull/4228#issuecomment-723748948
- [x] ~~HIP?~~ (~~**UPDATE**: not possible~~ **UPDATE2**: possible if hipRTC is fixed)
- [ ] Discussions: Mark as experimental? Lock submodule commit?


leofang:
The PR description is updated.

leofang:
Hi @mnicely Would this be useful to you?

leofang:
Tested compiling a CUB block reduction kernel. Jitify's first compilation is slower than simply calling `nvcc` (the old approach), but the subsequent encountering by reading from disk is faster:
```python
# test.py

import cupy as cp
import numpy as np
import time

cp.core._accelerator.set_routine_accelerators([])
cp.core._accelerator.set_reduction_accelerators(['cub'])

shape = (512, 512, 512)
axes=[(2,)]

a = cp.random.random(shape, dtype=cp.float32)
a_np = cp.asnumpy(a)

for axis in axes:
    t1 = time.time()
    out_cp = cp.sum(a, axis)
    t2 = time.time()
    print(t2 - t1)  # check CPU time
    out_np = np.sum(a_np, axis)
    if not np.allclose(cp.asnumpy(out_cp), out_np, rtol=1E-4):
        print("cupy:", out_cp)
        print("numpy:", out_np)
```
and repeat these commands to time it
```
$ rm ~/.cupy/kernel_cache/*.cubin
$ python test.py  # compilation happens
$ python test.py  # just read cubin from disk without actual compiling
```
I got the following rough timings:

time  | 1st call | 2nd and onward calls
--|--|--
NVCC | 0.641 | 0.187
Jitify+NVRTC | 5.51 | 0.07

So looks like changing to Jitify is acceptable if the kernel is expected to be reused.

leofang:
Jenkins, test this please

leofang:
This is very strange. Can't reproduce any of the errors locally...🙃 Could that be a transient thing?

Jenkins, test this please

leofang:
No idea why Jitify on the CI cannot locate where `cupy_fp16.h` is. Obviously the include path is passed correctly. Give it one more shot.

Jenkins, test this please

leofang:
I think the previous problem is fixed. We no longer see weird errors, just CI timeout, which is expected. May need one or two more retries to complete the full test.

Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
cupy-py3-cutensor passed:

https://jenkins.preferred.jp/job/chainer/job/cupy_pr/2454/TEST=cupy-py3-cutensor,label=mn1-p100/

leofang:
Two more tests to go!

Jenkins, test this please

mnicely:
> Hi @mnicely Would this be useful to you?

Hi @leofang. I think it might be! I’ve had to switch gears to work a new task. When I get time I think about valid scenarios in cuSignal.

leofang:
Converted to draft mode to investigate if it's possible to accelerate Jitify's slow search loop.

leofang:
In 3a7a6a9 we keep the headers found by Jitify to help it search faster. This reduces the runtime dramatically (previously each Jitify fire would search from scratch), ex: `test_sumprod.py` takes 760.19s -> 145.64s. (Cached cubins are removed to ensure a meaningful timing.)

leofang:
> In [3a7a6a9](https://github.com/cupy/cupy/commit/3a7a6a9bf86d7bbc8c1a6078152343f5908e5fb3) we keep the headers found by Jitify to help it search faster. This reduces the runtime dramatically (previously each Jitify fire-up would search from scratch), ex: `test_sumprod.py` takes 760.19s -> 145.64s. (Cached cubins are removed prior to tests to ensure a meaningful timing.)

Following https://github.com/cupy/cupy/pull/4228#issuecomment-723272460, we can check the effect of header cache by expanding the test script there to force compiling two different CUB kernels:
```python
# test2.py

import cupy as cp
import numpy as np
import time

cp.core._accelerator.set_routine_accelerators([])
cp.core._accelerator.set_reduction_accelerators(['cub'])

shape = (512, 512, 512)
axes=[(2,), (1, 2)]  # this leads two different CUB kernels compiled

a = cp.random.random(shape, dtype=cp.float32)
a_np = cp.asnumpy(a)

for axis in axes:
    t1 = time.time()
    out_cp = cp.sum(a, axis)
    t2 = time.time()
    print(t2 - t1)  # check CPU time
    out_np = np.sum(a_np, axis)
    if not np.allclose(cp.asnumpy(out_cp), out_np, rtol=1E-4):
        print("cupy:", out_cp)
        print("numpy:", out_np)
```
and the outputs are
time | 1st run | 2nd and onward runs
-----|--------|---------------------
NVCC | 0.602, 0.418 | 0.195, 0.0008
Jitify+NVRTC (3a7a6a9) | 4.272, 0.356 | 0.0715, 0.0010

So clearly the cached headers help Jitify dramatically (note that the time compiling the 2nd kernel is faster than nvcc). The time compiling the 1st kernel is also reduced a bit, as we explicitly pass in CuPy headers to initialize Jitify.

leofang:
I think this is ready for review.

Jenkins, test this please

cupy_setup_build.py
```diff
@@ -558,8 +575,6 @@ def make_extensions(options, compiler, use_cython):
         link_args = s.setdefault('extra_link_args', [])
 
         if module['name'] == 'cusolver':
-            compile_args = s.setdefault('extra_compile_args', [])
-            link_args = s.setdefault('extra_link_args', [])
```
leofang, line 562: Not sure these two lines are here, but they're redundant as we just did this 3 lines above.

emcastillo:
Does this work for windows?

leofang:
According to Jitify's README.md, they say it does.

emcastillo (approved):

emcastillo:
Jenkins, test this please

leofang:
Thanks!

PR #4229, opened 2020-11-05
Skip `scipy.sparse.linalg.eighs` tests for CUDA 9.2

emcastillo:
In the cupy daily combination tests, the recent PR #4138 makes use of a call that returns nan arrays in CUDA 9.2

With the provided [t.npy.zip](https://github.com/cupy/cupy/files/5492771/t.npy.zip), cusolver syevd fails in CUDA 9.2
```python
import cupy
a = cupy.load('t.npy')
cupy.linalg.eigh(a)  # nans
```

We will skip these tests since there is no easy way to fix this in our side, as this is an older cusolver internal problem.
@anaruse any thoughts about this? thanks!

### DIFF
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
@@ -117,6 +117,9 @@ class TestVectorNorm(unittest.TestCase):
     'return_eigenvectors': [True, False],
 }))
 @unittest.skipUnless(scipy_available, 'requires scipy')
+@unittest.skipUnless(
+    cupy.cuda.runtime.runtimeGetVersion() > 9200, 'requires CUDA > 9.2'
+)
 class TestEigsh(unittest.TestCase):
     n = 30
     density = 0.33


emcastillo:
We need to document this behavior. @toslunar, @kmaehashi 
Should we add a note or warning in the `cupy.linalg.eigh` docstring?

toslunar:
Could you run the following to check if cuSOLVER reports an error?
```python
import cupy, cupyx
a = cupy.load('t.npy')
with cupyx.errstate(linalg='raise'):
    cupy.linalg.eigh(a)  # nans
```

emcastillo:
No error is reported 😢

anaruse:
So the failure is occurring here? Perhaps there is a bug in CUDA 9.2's cuSOLVER.

https://github.com/cupy/cupy/blob/786c303540fbcd81ff11667ac23ede67cad104c9/cupyx/scipy/sparse/linalg/_eigen.py#L134-L150

One workaround is to use `numpy.linalg.eigh` here instead of `cupy.linalg.eigh`. This would require data transfer between CPU and GPU, but since the matrix size processed here is very small, there would be little or no performance loss.

emcastillo:
The bug is indeed in cuSOLVER, let me discuss with the team the possible solution :).

Thanls!

anaruse:
@emcastillo , Could you try https://github.com/cupy/cupy/pull/4231 on CUDA 9.2?

emcastillo:
Error was fixed in #4231, closing this

PR #4233, opened 2020-11-06
Refactor `AssertFunctionIsCalled`

### LINKED ISSUE #4232: `AssertFunctionIsCalled` fails to unpatch target
`AssertFunctionIsCalled` does not exit `mock.patch` if the assertion fails:
https://github.com/cupy/cupy/blob/c0c99108b78f4480b17cdf22ba3e903eb65f3646/cupy/testing/helper.py#L1441-L1444

As a consequence, CI reports weird failures.

Example: https://jenkins.preferred.jp/job/chainer/job/daily_master_cupy-shuffled/378/ID=20,label=mn1-p100/consoleFull

- The first failure is `TestCubReduction_param_5_{backend='device', order='F', shape=(10, 20)}.test_cub_max` on `assert self.handle.call_count == int(self.times_called)` in `AssertFunctionIsCalled.__exit__`.
- The second failure is `TestCubReduction_param_5_{backend='device', order='F', shape=(10, 20)}.test_cub_min` on `assert cupy_r.shape == numpy_r.shape`, where `cupy_r.shape == ()`, `numpy_r.shape == (20,)`.

toslunar:
Fixed issues:
- assertion failure leaves the target unpatched
- `times_called` is also passed to `mock.patch`

(close #4232)

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -1414,6 +1414,15 @@ class NumpyAliasValuesTestBase(NumpyAliasTestBase):
         assert self.cupy_func(*self.args) == self.numpy_func(*self.args)
 
 
+@contextlib.contextmanager
+def _assert_function_is_called(*args, times_called=1, **kwargs):
+    with mock.patch(*args, **kwargs) as handle:
+        assert handle.call_count == 0
+        yield handle
+        assert handle.call_count == int(times_called)
+        del handle
+
+
 class AssertFunctionIsCalled:
 
     def __init__(self, mock_mod, **kwargs):
@@ -1427,18 +1436,10 @@ class AssertFunctionIsCalled:
                 called. Default is ``1``.
 
         """
-
-        self.patch = mock.patch(mock_mod, **kwargs)
-
-        times_called = kwargs.get('times_called')
-        self.times_called = times_called if times_called is not None else 1
+        self._ctx = _assert_function_is_called(mock_mod, **kwargs)
 
     def __enter__(self):
-        self.handle = self.patch.__enter__()
-        assert self.handle.call_count == 0
-        return self.handle
+        return self._ctx.__enter__()
 
     def __exit__(self, exc_type, exc_value, traceback):
-        assert self.handle.call_count == int(self.times_called)
-        del self.handle
-        return self.patch.__exit__(exc_type, exc_value, traceback)
+        return self._ctx.__exit__(exc_type, exc_value, traceback)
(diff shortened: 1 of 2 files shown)


toslunar:
I am fine with
```python
@contextlib.contextmanager
def assert_function_is_called(*args, times_called=1, **kwargs):
    with mock.patch(*args, **kwargs) as handle:
        yield
        assert handle.call_count == times_called

# TODO: remove this alias
AssertFunctionIsCalled = assert_function_is_called
```

@leofang Do you want to keep any of the following?
- `with AssertFunctionIsCalled(...) as handle`
- forbidding usage as a function decorator
- supporting bool `times_called`

leofang:
Sorry for my late reply @toslunar!

With regard to the failures that `AssertFunctionIsCalled` captured, as I mentioned in the original issue #4232 they should be investigated.

With regard to the assertion error in `AssertFunctionIsCalled` that causes issues, I think either the current PR or or the shorter design https://github.com/cupy/cupy/pull/4233#issuecomment-722990369 might still have similar issues if the `assert` is done in the `with mock.patch ...` context (not 100% sure, though). Perhaps we could remove one indentation level and pull it out of the context?

The original design was to make it work with `mock.patch` properly. There're a few tricky cases where
- we need to mock objects from Cython
- we need actually run the underlying code despite it's being mocked

I remember I tried a `contextmanager` approach but I don't remember why I abandoned it.

If the refactored design can handle all cases properly **and** also fix the assert issue (make the detected error propagate up) I don't have strong opinions for the 3 listed questions. Please feel free to adjust as you see fit 🙂

toslunar:
Thanks @leofang.  I'll write tests for the tricky cases.

To assert `call_count` even after the inner code raises,
```python
@contextlib.contextmanager
def assert_function_is_called(*args, times_called=1, **kwargs):
    with mock.patch(*args, **kwargs) as handle:
        try:
            yield
        finally:
            assert handle.call_count == times_called
```
is required. However, I'd rather not support it because `with assert_function_is_called(): with pytest.raises():` looks better than the other order `with pytest.raises(): with assert_function_is_called():`.

BTW, try-except like the following seems the way to fix the issue in the style of custom `__exit__`.
```python
try:
    assert ...
finally:
    self.patch.__exit__(...)
```

takagi:
pfnCI, test this please.

takagi:
See also: https://github.com/cupy/cupy/issues/4232#issuecomment-724474182

takagi:
LGTM!

leofang:
Thank you, @toslunar!

PR #3896, opened 2020-08-28
ROCm: Support rocFFT/hipFFT

leofang:
~~Blocked by #3823.~~ Close #3892.

~~This builds and many tests pass, but some are failing.~~

***UPDATE***: major changes
- Support hipFFT
    - Split the code paths for `CUPY_USE_HIP` and `CUPY_NO_CUDA` for functions supported by HIP
    - Work around several (likely) bugs I found, in particular `PlanNd` cannot be used for C2R transforms (due to https://github.com/cupy/cupy/pull/3896#issuecomment-683797721); see the added comments for further detail.
- Cythonize `Plan1d` and `PlanNd`
    - this is necessary because in cuFFT the plan handle is an `int`, but in `hipFFT` it is a pointer to struct...Now we unify the code path by passing an `intptr_t` around
    - the code path splitting requires a compile time constant passed to Cython by setuptools
    - `__del__` is changed to `__dealloc__` (#3999)

PR #4234, opened 2020-11-06
Use `.imag = 0` at hipFFT workaround

toslunar:
Nitpick at #3896.

### DIFF
--- a/cupy/fft/_fft.py
+++ b/cupy/fft/_fft.py
@@ -108,11 +108,11 @@ def _exec_fft(a, direction, value_type, norm, axis, overwrite_x,
     # while hipFFT handles it badly in both Plan1d and PlanNd, so we must
     # do the correction ourselves to ensure the condition is met.
     if cupy.cuda.runtime.is_hip and value_type == 'C2R':
-        a[..., 0] = a[..., 0].real + 0j
+        a[..., 0].imag = 0
         if out_size is None:
-            a[..., -1] = a[..., -1].real + 0j
+            a[..., -1].imag = 0
         elif out_size % 2 == 0:
-            a[..., out_size // 2] = a[...,  out_size // 2].real + 0j
+            a[..., out_size // 2].imag = 0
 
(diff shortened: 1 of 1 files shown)


leofang:
Hi @toslunar Thanks for the code fix, but unfortunately there is a reason I didn't do this. For real arrays (either NumPy or CuPy) we can't overwrite their imaginary components:
```python
>>> import cupy as cp
>>> a = cp.arange(33, dtype=cp.float32)
>>> a.imag = 0
  File "cupy/core/core.pyx", line 1143, in cupy.core.core.ndarray.imag.__set__
    _math._ndarray_imag_setter(self, value)
  File "cupy/core/_routines_math.pyx", line 83, in cupy.core._routines_math._ndarray_imag_setter
    raise TypeError('cupy.ndarray does not have imaginary part to set')
TypeError: cupy.ndarray does not have imaginary part to set
>>> 
>>> import numpy as np
>>> a = np.arange(33, dtype=cp.float32)
>>> a.imag = 0
TypeError: array does not have imaginary part to set
```
And when we arrive here there's no guarantee that the input would be complex. Perhaps you could add a comment instead?

toslunar:
After `a = _convert_dtype(a, value_type)` with `value_type='C2R'` (or `'C2C'`), we have complex `a.dtype`.  Then, its `imag` attribute is writable.
```
>>> a = np.arange(3, dtype=np.complex64)
>>> a.imag = 1
>>> a
array([0.+1.j, 1.+1.j, 2.+1.j], dtype=complex64)
```

emcastillo:
Jenkins, test this please

leofang:
Let me test this on my AMD build later today.

leofang (approved):
Thank you for your explanation, @toslunar! You're absolutely right. I don't know what I was thinking 😅 Good catch! All FFT tests passed locally on a Radeon VII +  ROCm 3.5.0.

emcastillo (approved):

PR #4242, opened 2020-11-09
Detect the CC of the device when building

emcastillo:
When building CuPy, most of the time the users do not specify the CC of their devices and end up building all of it, this is especially true when they just do `pip install cupy`. This PR detects the CCs at setup time and tailors the build unless specified otherwise (useful when building wheels).

TODO:
- Add a define with the ccs CuPy was built, so we can prevent future errors. (User installing a new GPU in his prebuilt CuPy install).

### DIFF
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -464,6 +464,8 @@ def preconfigure_modules(compiler, settings):
             # Skip checking other modules when CUDA is unavailable.
             if module['name'] == 'cuda':
                 break
+    # Get a list of the CC of the devices connected to this node
+    build.check_compute_capabilities(compiler, settings)
 
     if len(ret) != len(MODULES):
         if 'cuda' in ret:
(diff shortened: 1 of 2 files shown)


kmaehashi:
I think some users rely on the current behavior (e.g. HPC cluster users where build and runtime environment is different). I prefer "safe by default, opt-in for performance" basis here, because users who trapped in this issue are almost impossible to know what's happening (maybe they will see an error message like `PTX incompatible` only when using thrust/cub).

However allowing users to build CuPy based on visible devices sounds nice. How about doing this when `CUPY_NVCC_GENERATE_CODE=current` is set?

emcastillo:
Fixed to use `CUPY_NVCC_GENERATE_CODE=current`

kmaehashi:
LGTM! Could you add `CUPY_NVCC_GENERATE_CODE=current` to docs?

kmaehashi:
pfnCI, test this please.

PR #4244, opened 2020-11-09
Fix `numpy_cupy_equal` for case that both numpy cupy raise errors

asi1024:
Blocked by https://github.com/cupy/cupy/pull/4252.

Fixes `testing.numpy_cupy_equal` decorator for the following situation.

```py
class TestX(unittest.TestCase):
    @testing.numpy_cupy_equal()
    def test_x(self, xp):
        raise ValueError
```

The current master branch passes the above test, even if `numpy` and `cupy` raise errors of different types.

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -661,6 +661,12 @@ def numpy_cupy_equal(name='xp', sp_name=None, scipy_name=None):
                     _call_func_numpy_cupy(
                         self, impl, args, kw, name, sp_name, scipy_name))
 
+            if cupy_error or numpy_error:
+                _check_cupy_numpy_error(
+                    self, cupy_error, cupy_tb, numpy_error, numpy_tb,
+                    accept_error=False)
+                return
+
             if cupy_result != numpy_result:
                 message = '''Results are not equal:
 cupy: %s
--- a/tests/cupy_tests/sorting_tests/test_sort.py
+++ b/tests/cupy_tests/sorting_tests/test_sort.py
@@ -624,9 +624,9 @@ class TestArgpartition(unittest.TestCase):
         a = testing.shaped_random((10,), xp, dtype, 100)
         kth = 2
         idx = self.argpartition(a, kth)
-        self.assertTrue((a[idx[:kth]] < a[idx[kth]]).all())
-        self.assertTrue((a[idx[kth]] < a[idx[kth + 1:]]).all())
-        return idx[kth]
+        self.assertTrue((a[idx[:kth]] <= a[idx[kth]]).all())
+        self.assertTrue((a[idx[kth]] <= a[idx[kth + 1:]]).all())
+        return a[idx[kth]]
 
     # TODO(leofang): test all dtypes -- this workaround needs to be kept,
     # likely due to #3287? Need investigation.


toslunar (commented):
What was the problem with `_check_cupy_numpy_error` (51fde7ac2f9411f7ebcd77e8d614fbebc18b220c)?

asi1024:
@toslunar `_check_cupy_numpy_error` fails at the following line, if both `numpy` and `cupy` raise no errors.
https://github.com/cupy/cupy/blob/51fde7ac2f9411f7ebcd77e8d614fbebc18b220c/cupy/testing/helper.py#L155

toslunar:
Thanks.  I overlooked `if cupy_error or numpy_error:`.  Many decorators use `_make_decorator`, and its error check is
https://github.com/cupy/cupy/blob/c42e571418841346589ce94d5c13b274dc19ec73/cupy/testing/helper.py#L236-L241

toslunar:
By the way, should `numpy_cupy_equal` check that the results are not `ndarray`s?  For example, `assert cupy.array([3.0]) == numpy.array(3)` passes (because `bool` applies to one-element array) but shapes and dtypes are different.

cupy/testing/helper.py
```diff
@@ -635,6 +636,14 @@ def check_func(x, y):
                            accept_error, sp_name, scipy_name)
 
 
+def _has_ndarray(x):
+    if isinstance(x, (numpy.ndarray, cupy.ndarray)):
+        return True
+    if isinstance(x, collections.abc.Iterable):
```
toslunar, line 642:
```suggestion
    if isinstance(x, collections.abc.Sequence):
```
`Sequence` looks better because iterators can be consumed.

Use `(tuple, list)` to be compatible with other decorators.

cupy/testing/helper.py
```diff
@@ -661,6 +670,17 @@ def test_func(self, *args, **kw):
...
+            if cupy_error or numpy_error:
+                _check_cupy_numpy_error(
+                    self, cupy_error, cupy_tb, numpy_error, numpy_tb,
+                    accept_error=False)
+                return
+
+            if _has_ndarray(cupy_result) or _has_ndarray(numpy_result):
+                raise AssertionError(
+                    'ndarrays are not allowed as return values. '
+                    'Please use `testing.numpy_cupy_arra_equal` instead.')
```
toslunar, line 682:
```suggestion
                    'Please use `testing.numpy_cupy_array_equal` instead.')
```

asi1024:
Rebased.

toslunar (approved):
LGTM

I'd test the decorator for the cases that `_has_ndarray` return True, but OK to leave it now because the fix is high priority.

toslunar:
Jenkins, test this please.

toslunar (changes requested):
> By the way, should `numpy_cupy_equal` check that the results are not `ndarray`s? For example, `assert cupy.array([3.0]) == numpy.array(3)` passes (because `bool` applies to one-element array) but shapes and dtypes are different.

Could you make a new PR for this fix?  Many tests with `numpy_cupy_equal` return ()-shaped arrays.  It may be intentional because `numpy_cupy_equal` essentially asserts the size of result arrays is 1 unlike `numpy_cupy_array_equal`.

toslunar (approved):

toslunar:
Jenkins, test this please.

toslunar:
The PR found other failures :+1:
```
tests.cupy_tests.sorting_tests.test_sort.TestArgpartition_param_0_{external=False}.test_argpartition_one_dim
tests.cupy_tests.sorting_tests.test_sort.TestArgpartition_param_1_{external=True}.test_argpartition_one_dim
```

Fix (or temporarily mark xfail) them, first.

toslunar:
Jenkins, test this please.

PR #4252, opened 2020-11-10
Fix tests of `__bytes__`

toslunar:
- `bytes` of uninitialized arrays should not be tested.  Changed them to have size = 0.
- `ndarray.item` returns python scalar.  Test `()`-shaped arrays instead.

### DIFF
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -499,7 +499,11 @@ class TestPythonInterface(unittest.TestCase):
         x = xp.empty((3, 0, 4), dtype)
         return bytes(x)
 
-    @testing.for_all_dtypes()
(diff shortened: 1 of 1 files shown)


toslunar:
To be concrete,
```python
x = xp.array([3], dtype).item()
return bytes(x)
```
calls `bytes(int)`: bytes object of size (= 3) given by the parameter.

```
>>> help(bytes)
Help on class bytes in module builtins:

class bytes(object)
 |  bytes(iterable_of_ints) -> bytes
 |  bytes(string, encoding[, errors]) -> bytes
 |  bytes(bytes_or_buffer) -> immutable copy of bytes_or_buffer
 |  bytes(int) -> bytes object of size given by the parameter initialized with null bytes
 |  bytes() -> empty bytes object
```

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #4261, opened 2020-11-11
Use GitHub Actions

asi1024:
Behaves as the same as https://github.com/cupy/cupy/blob/master/.travis.yml.
https://github.com/asi1024/cupy/runs/1384140459?check_suite_focus=true


kmaehashi:
LGTM!

PR #4267, opened 2020-11-11
Reduce the file size of cuFFT callback modules

leofang:
Follow-up of #4141. Many thanks to @maxpkatz for suggesting [`nvprune`](https://docs.nvidia.com/cuda/cuda-binary-utilities/index.html#nvprune) 🙏

With this PR, the file size of each Python share library generated at runtime is greatly reduced from 159M to 55M on my system (CUDA 10.2 + sm75). This is made possible by using `nvprune` to slim the size of `libcufft_static.a` for the target arch. The slimmed static library is cached in the same directory as the Python modules for easy reuse.

`nvprune` is made as a soft requirement: According to @maxpkatz, `nvprune` will always be present iff `nvcc` is (which is a hard requirement for the callback support), so in most scenarios this would just work. In the extremest cases where `nvprune` is not found, we fall back to the old behavior (linking against the full `libcufft_static.a`).

### DIFF
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -57,8 +57,12 @@ cdef str _python_include = sysconfig.get_path('include')
 cdef list _nvcc = []
 cdef str _cuda_path = get_cuda_path()
 cdef str _cuda_include = None  # workaround for Read the Docs...
+cdef str _nvprune = None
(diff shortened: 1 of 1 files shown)


leofang:
Jenkins, test this please

emcastillo (approved):

PR #4269, opened 2020-11-11
`testing.numpy_cupy_allclose` with per-dtype tolerance

### LINKED ISSUE #4220: Proposal: Tolerance per dtype for `cupy.testing.numpy_cupy_allclose`
Currently, `cupy.testing.numpy_cupy_allclose` takes only a tolerance value, so we need to split a test or make an imperative call of `cupy.testing.assert_allclose` to supply different tolerance values per dtype.

Here proposed is making `cupy.testing.numpy_cupy_allclose` take association from a dtype to a tolerance value something as following:

```Python
@testing.numpy_cupy_allclose(rtol={np.float32: 1e-4, np.float64: 1e-12,
                                   np.complex64: 1e-4, np.complex128: 1e-12,
                                   'default': 1e-7},
                             atol={np.float32: 1e-4, np.float64: 1e-12,
                                   np.complex64: 1e-4, np.complex128: 1e-12,
                                   'default': 1e-7},
                             sp_name='sp')
```

Actual representation would have room for discussion. If the idea looks good to you, I will implement this feature.

takagi:
Closes #4220.

This PR makes `testing.numpy_cupy_allclose` allow tolerance per dtype.

### DIFF
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -386,8 +386,13 @@ def numpy_cupy_allclose(rtol=1e-7, atol=0, err_msg='', verbose=True,
     """Decorator that checks NumPy results and CuPy ones are close.
 
     Args:
-         rtol(float): Relative tolerance.
-         atol(float): Absolute tolerance.
+         rtol(float or dict): Relative tolerance. Besides a float value, a
+             dictionary that maps a dtypes to a float value can be supplied to
+             adjust tolerance per dtype. If the dictionary has ``'default'``
+             string as its key, its value is used as the default tolerance in
+             case any dtype keys do not match.
+         atol(float or dict): Absolute tolerance. Besides a float value, a
+             dictionary can be supplied as ``rtol``.
          err_msg(str): The error message to be printed in case of failure.
          verbose(bool): If ``True``, the conflicting values are
              appended to the error message.
--- a/cupy/testing/helper.py
+++ b/cupy/testing/helper.py
@@ -450,7 +450,7 @@ def numpy_cupy_allclose(rtol=1e-7, atol=0, err_msg='', verbose=True,
     # tolerance associations.
     if not type_check:
         if isinstance(rtol, dict) or isinstance(atol, dict):
-            raise TypeError('When `type_ckeck` is `False`, `rtol` and `atol` '
+            raise TypeError('When `type_check` is `False`, `rtol` and `atol` '
                             'must be supplied as float.')
 
     def check_func(c, n):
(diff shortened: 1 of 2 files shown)


cupy/testing/helper.py
```diff
@@ -352,15 +352,47 @@ def _convert_output_to_ndarray(c_out, n_out, sp_name, check_sparse_format):
             type(c_out), type(n_out)))
 
 
+def _resolve_tolerance(type_check, result, rtol, atol):
+    def _resolve(dtype, tol):
+        if isinstance(tol, dict):
+            tol1 = tol.get(dtype.type)
```
asi1024, line 358: If `tol` is specified with a wrong format, for example `{'float16': 1e-3}`, the default tolerance will be chosen and may pass the wrong test unexpectedly. Did you have some ideas to avoid this case?

asi1024 (changes requested):
Could you add xfail tests?

cupy/testing/helper.py
takagi: Definitely. I think an option is using an association list `[(<dtype>, <tolerance>)]` instead of a dictionary, where `<dtype>` is an object such that `numpy.dtype(<dtype>) == {numpy,cupy}_result.dtype` is satisfied. This way we can use `[(numpy.float16, 1e-3)]`, `[('float16', 1e-3)]` or `[('f', 1e-3)]` as well. How do you think?

asi1024: Hmm, I feel the `dict` is a more natural interface. How about checking if all keys are of `type` type or `default` during the preprocess?

takagi: Sounds good. I'll fix the code as such.

takagi: I've made a commit, PTAL!

cupy/testing/helper.py
```diff
@@ -352,15 +352,63 @@ def _convert_output_to_ndarray(c_out, n_out, sp_name, check_sparse_format):
...
+        else:
+            return tol
+
+    # When `type_check` is `False`, cupy result and numpy result may have
+    # different dtypes so we can not determine the dtype to use from the
+    # tolerance associations.
+    if not type_check:
+        if isinstance(rtol, dict) or isinstance(atol, dict):
+            raise TypeError('When `type_ckeck` is `False`, `rtol` and `atol` '
+                            'must be supplied as float.')
```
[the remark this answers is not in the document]
takagi: Sure!

asi1024:
Jenkins, test this please.

asi1024 (approved):

asi1024:
LGTM!

PR #4276, opened 2020-11-12
Fix possible redefinition of "-ccbin" in `cupy.fft._callback`

leofang:
refs:
- https://jenkins.preferred.jp/job/chainer/job/daily_master_cupy-shuffled/ID=18,label=mn1-p100/383/console
- `environment.sh:        export NVCC="${NVCC} --compiler-bindir gcc-6"`

cc: @emcastillo

### DIFF
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -95,11 +95,12 @@ cdef inline void _set_nvcc_path() except*:
         nvcc = get_nvcc_path()
         if nvcc is not None:
             _nvcc = nvcc.split(' ')
-            # if the host compiler is set in NVCC, we force it to align with
+            # if the host compiler is set in NVCC, we use it to compile the
+            # Python module; if the host compiler is not set, then we use
             # the one reported by sysconfig
             for i, f in enumerate(_nvcc):
                 if f in ('-ccbin', '--compiler-bindir'):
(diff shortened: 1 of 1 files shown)


leofang:
Jenkins, test this please

cupy/fft/_callback.pyx
```diff
@@ -96,6 +75,38 @@ cdef class _ThreadLocal:
...
+    cdef int i
+
+    global _nvcc
+    if _nvcc == []:
+        nvcc = get_nvcc_path()
+        if nvcc is not None:
+            _nvcc = nvcc.split(' ')
+            # if the host compiler is set in NVCC, we force it to align with
+            # the one reported by sysconfig
+            for i, f in enumerate(_nvcc):
```
emcastillo, line 100: why don't we use the one that is provided in nvcc?

leofang: My original thought was I wanna use the compiler Python says it'd use for building extensions, but now think it also makes sense to do the other way around. Will change.

leofang: Changed in 30a620d. @emcastillo PTAL.

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4278, opened 2020-11-12
add scipy.fft module to the API comparison table

grlee77:
This PR adds the `scipy.fft` module to the [API Comparison Table](https://docs.cupy.dev/en/stable/reference/comparison.html).

Doing this requires SciPy >= 1.4 when building the docs. Given that `requirements.txt` only specifies SciPy >=1.1, I added a separate requirements_docs.txt for readthedocs. Does that seem reasonable? (We can go back to a single `requirements.txt` again once the minimum SciPy is >= 1.4).

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 sphinx==2.0.1
 Cython==0.28.3
 numpy >=1.15
-scipy >=1.1
+scipy >=1.4
--- a/docs/requirements_docs.txt
+++ b/docs/requirements_docs.txt
@@ -1,0 +1,4 @@
+sphinx==2.0.1
+Cython==0.28.3
+numpy >=1.15
+scipy >=1.4
--- a/docs/requirements_docs.txt
+++ b/docs/requirements_docs.txt
@@ -1,4 +1,0 @@
-sphinx==2.0.1
-Cython==0.28.3
-numpy >=1.15
-scipy >=1.4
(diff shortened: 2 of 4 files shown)


toslunar:
It seems OK to edit `docs/requirements.txt` for CuPy docs.

Reviewers should check the file to be edited.  Maybe `setup.py`.
https://github.com/cupy/cupy/blob/b4a296f68bb78f57048c2c3288d13c46482e9034/setup.py#L60-L63

kmaehashi:
Thanks!
pfnCI, test this please.

kmaehashi:
The comparison table is rendered correctly. https://cupy--4278.org.readthedocs.build/en/4278/reference/comparison.html#id1
Although there're several left undocumented (missing links)...

grlee77:
> Although there're several left undocumented (missing links)...

It looks like the ones missing links are all `fftshift`/`fftfreq` functions that are just imported from the `cupy.fft` module into the `scipy.fft` namespace. I assume this is why they are being treated differently and don't get a documentation link? 

In terms of functions that are missing altogether
- `get_workers`/`set_workers` is not applicable for CUPY (these control CPU threads for MKL, FFTW, etc.)
- all `*_backend` functions should not be implemented in CuPy (they are used by SciPy to implement its uarray backend system. CuPy users would be calling these functions from SciPy to control the backend)
- all DCT and DST (real-to-real FFTs) functions have not been implemented here, although I do have code in a branch at `cupyimg` that implements these (for the type II and III variants) and can make a PR at some point.
- `hfft2`/`ihfft2` and `hfftn`/`ihfftn` have not been implemented yet

toslunar (approved):

Issue #4105, opened 2020-10-09
Feature request: Support cuFFT callback routines


leofang:
This is something that @maxpkatz and I tried a while ago but we never figured it out. It would be nice to support the cuFFT callback routines in CuPy:
https://docs.nvidia.com/cuda/cufft/index.html#callback-routines
https://developer.nvidia.com/blog/cuda-pro-tip-use-cufft-callbacks-custom-data-processing/
The idea is to reduce access to global memory by fusing both elementwise pre- and post- processing kernels with the FFT operations to increase the performance.

The problem we encountered was that it requires linking to cuFFT's static library instead of shared library, but if we do this by changing this line
https://github.com/cupy/cupy/blob/5dcd1a53c2c6667b6cfafc132402554f8e5d3299/cupy_setup_build.py#L141
to ~~`'cufftw'`~~ `'cufft_static'` as instructed by the [cuFFT documentation](https://docs.nvidia.com/cuda/cufft/index.html#static-library) and rebuild CuPy (to be precise, the `cupy.cuda.cufft` module), we encounter "undefined symbols" errors when importing CuPy. Those symbols are for the callback support and are leaked from ~~`libcufftw.a`~~ `libcufft_static.a` into the module object. 

**I can work out this support myself, but I need someone to help me resolve this issue.**

cc: @bjoernenders @smarkesini

leofang:
If this works, my idea was that users can write `cupy.RawKernel`s for pre- and post- processings, respectively, and then pass them to the callback APIs to tie with an existing cuFFT plan. This might not be fully compatible with the high-level APIs like `cupy.ftt`, but at least makes it exposed to the Python space for advanced users to try.

mnicely:
@leofang If I'm reading the docs right, `cufftw` is for FFTW user that want to port code.

To utilize callbacks you have to link `cufft_static` and `culibos`. Unfortunately, the library also has to be device linked with `-dc`.

From blog: https://developer.nvidia.com/blog/cuda-pro-tip-use-cufft-callbacks-custom-data-processing/

The cuFFT callback feature is available in the statically linked cuFFT library only, currently only on 64-bit Linux operating systems. Callbacks therefore require us to compile the code as relocatable device code using the --device-c (or short -dc) compile flag and to link it against the static cuFFT library with -lcufft_static. Important: If you forget to compile as relocatable device code your application may quietly ignore callbacks and produce wrong results!

[...]

PR #4283, opened 2020-11-15
Fix undefined symbols in `cupy.fft._callback`

leofang:
Fix https://github.com/cupy/cupy/issues/4105#issuecomment-727633420. This reverts commit 30a620db38889692694745f5e738e86f080fe524. It seems when linking all callback objects, a C++ compiler must be used (ex: g++ instead of gcc) in order to resolve undefined symbols (ex: `__cxa_pure_virtual`).

cc: @llukas @mnicely @emcastillo

### DIFF
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -95,12 +95,11 @@ cdef inline void _set_nvcc_path() except*:
         nvcc = get_nvcc_path()
         if nvcc is not None:
             _nvcc = nvcc.split(' ')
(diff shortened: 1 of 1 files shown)


emcastillo:
Jenkins, test this please

emcastillo (approved):

leofang:
Looks like we finally fixed it! No more callback-related errors:
https://jenkins.preferred.jp/job/chainer/job/daily_master_cupy-shuffled/387

PR #4289, opened 2020-11-16
Support CUB-backed `cupy.ReductionKernel` and `@cupy.fuse`

### LINKED ISSUE #3983: Enhancement: enable CUB backend for ReductionKernel
Currently the CUB reduction is disabled for `ReductionKernel` (by the last `False` argument):
https://github.com/cupy/cupy/blob/c1bf1aab6eb79a1cfcd5bb374460082d8eb2f3ca/cupy/core/_reduction.pyx#L724-L727
But we should revisit this at some point.

leofang:
Close #3983. Follow-up of #3244. Currently this only supports 1 in + 1 out for efficiently accessing the global memory. N in + 1 out could be revisited in the future; IIUC `ReductionKernel` does not support multiple outputs.

New changes:
- `cupy.ReductionKernel` can be backed by CUB (from `cupy.core._cub_reduction`)
- Fusion (`@cupy.fuse`) with reduction can also use CUB now; this again assumes 1 in + 1 out, which holds IIUC
- No user code change is required other than the need of setting `CUPY_ACCELERATORS=cub`.

Taking the example code from the tutorial:
```python
import cupy as cp
import numpy as np
from cupyx.time import repeat


l2norm_kernel = cp.ReductionKernel(
    'T x',  # input params
    'T y',  # output params
    'x * x',  # map
    'a + b',  # reduce
    'y = sqrt(a)',  # post-reduction map
    '0',  # identity value
    'l2norm'  # kernel name
)

x_cp = cp.random.random(1000000, dtype=cp.float64)

cp.core.set_reduction_accelerators([])
print(repeat(l2norm_kernel, (x_cp,), n_repeat=100))
cp.core.set_reduction_accelerators(['cub'])
print(repeat(l2norm_kernel, (x_cp,), n_repeat=100))

y_cp = l2norm_kernel(x_cp)
x_np = cp.asnumpy(x_cp)
y_np = np.sqrt(np.sum(x_np**2))
print(cp.allclose(y_cp, y_np))
```
Output:
```
l2norm              :    CPU:   12.386 us   +/- 0.652 (min:   11.617 / max:   15.534) us     GPU-0:  701.201 us   +/- 1.012 (min:  699.232 / max:  705.760) us
l2norm              :    CPU:   24.327 us   +/- 0.496 (min:   23.527 / max:   27.277) us     GPU-0:   54.900 us   +/- 0.478 (min:   54.208 / max:   57.376) us
True
```

***UPDATE***: fusion example
```python
import cupy as cp
import numpy as np
from cupyx.time import repeat

cp.core._accelerator.set_routine_accelerators([])

shape = (512, 512, 512)

a = cp.random.random(shape, dtype=cp.float32)
a_np = cp.asnumpy(a)


def my_op(x, sum_axes):
    # make the fuse function internal so that sum_axes can be varied
    @cp.fuse()
    def _my_op(x):
        y = cp.sin(x)
        y = cp.sum(y**2, axis=sum_axes)
        return y
    return _my_op(x)


for sum_axes in [(2,), (1,2), (0,1,2)]:
    print(f"axes={sum_axes}")
    cp.core._accelerator.set_reduction_accelerators([])
    print(repeat(my_op, (a, sum_axes), n_repeat=100))
    cp.core._accelerator.set_reduction_accelerators(['cub'])
    print(repeat(my_op, (a, sum_axes), n_repeat=100))
    out_cp = my_op(a, sum_axes)
    out_np = np.sum(np.sin(a_np)**2, axis=sum_axes)
    if not np.allclose(cp.asnumpy(out_cp), out_np, rtol=1E-4):
        print("cupy:", out_cp)
        print("numpy:", out_np)
    print()
```
Output:
```
axes=(2,)
my_op               :    CPU:  219.712 us   +/-10.059 (min:  213.884 / max:  303.909) us     GPU-0: 4580.685 us   +/-459.435 (min: 4371.840 / max: 5942.080) us
my_op               :    CPU:  220.783 us   +/- 3.813 (min:  216.017 / max:  237.544) us     GPU-0: 1743.457 us   +/-69.232 (min: 1566.688 / max: 1823.744) us

axes=(1, 2)
my_op               :    CPU:  217.272 us   +/- 4.960 (min:  211.869 / max:  243.187) us     GPU-0: 2183.468 us   +/-23.620 (min: 2097.472 / max: 2217.088) us
my_op               :    CPU:  220.258 us   +/- 3.579 (min:  216.138 / max:  239.502) us     GPU-0: 1948.207 us   +/- 7.690 (min: 1930.272 / max: 1974.240) us

axes=(0, 1, 2)
my_op               :    CPU:  224.835 us   +/- 5.922 (min:  214.166 / max:  253.125) us     GPU-0:138661.313 us   +/-26.385 (min:138596.954 / max:138738.693) us
my_op               :    CPU:  233.950 us   +/- 5.803 (min:  228.622 / max:  261.253) us     GPU-0: 1842.682 us   +/-46.423 (min: 1708.256 / max: 1912.768) us
```

### DIFF
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -43,12 +43,6 @@ cdef function.Function _create_cub_reduction_function(
 #        # from ReductionKernel (names supplied by the user)
 #        in_var = params[0].name
 #        out_var = params[1].name
-    cdef str old_in0 = params[0].name, old_out0 = params[1].name
-    cdef str args = _get_cub_kernel_params(params, arginfos)
-    print("old:", old_in0, old_out0)
-    pre_map_expr = pre_map_expr.replace(old_in0, 'in0')
-    post_map_expr = post_map_expr.replace(old_out0, 'out0')
-    print(pre_map_expr, '\n', post_map_expr)
 
     # static_assert needs at least C++11 in NVRTC
     options += ('--std=c++11',)
@@ -223,7 +217,7 @@ __global__ void ${name}(${params}) {
         block_size=block_size,
         items_per_thread=items_per_thread,
         reduce_type=reduce_type,
-        params=args,
+        params=_get_cub_kernel_params(params, arginfos),
 #        _raw_in0=in_var,
 #        _raw_out0=out_var,
         identity=identity,
@@ -649,15 +643,11 @@ cdef bint _try_to_call_cub_reduction(
     if can_use_cub is None:
         return False
 
+    print("\n\n\npreparing CUB...\n\n\n")
     axis_permutes, contiguous_size, full_reduction = can_use_cub
 
     in_shape = _reduction._set_permuted_args(
         in_args, axis_permutes, a_shape, self.in_params)
-
-    # When coming from ReductionKernel, type_map does not contain the expected
-    # names (type_in0_raw and type_out0_raw), so we must append them
-    type_map._pairs += (('type_in0_raw', in_args[0].dtype.type),
-                        ('type_out0_raw', out_args[0].dtype.type),)
 
     if in_args[0]._f_contiguous:
         ret._set_contiguous_strides(ret.dtype.itemsize, False)
@@ -682,6 +672,24 @@ cdef bint _try_to_call_cub_reduction(
               + _get_param_info(
                   size_type + '_segment_size', not full_reduction)
               + _get_param_info(size_type + '_array_size', True))
+
+    # HACK for ReductionKernel:
+    # 1. input/output arguments might not be named as in0/out0
+    # 2. pre-/post- maps might not contain in0/out0
+    # 3. type_map does not contain the expected names (type_in0_raw and
+    #    type_out0_raw)
+    cdef str old_in0 = params[0].name, old_out0 = params[1].name
+    if old_in0 != 'in0' or old_out0 != 'out0':
+        type_map._pairs += (('type_in0_raw', in_args[0].dtype.type),
+                            ('type_out0_raw', out_args[0].dtype.type),)
+        print("old:", old_in0, old_out0)
+        map_expr = map_expr.replace(old_in0, 'in0')
+        post_map_expr = post_map_expr.replace(old_out0, 'out0')
+        print(map_expr, '\n', post_map_expr)
+        # avoid overwriting self's attributes
+        params = (_get_param_info('T in0', True)
+                  + _get_param_info('T out0', False)
+                  + params[2:])
 
     # Calculate the reduction block dimensions.
     optimize_context = _optimize_config.get_current_context()
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -25,6 +25,7 @@ cdef function.Function _create_cub_reduction_function(
         reduce_type, params, arginfos, identity,
         pre_map_expr, reduce_expr, post_map_expr,
         _kernel._TypeMap type_map, preamble, options):
+    print(type_map)
     # A (incomplete) list of internal variables:
     # _J            : the index of an element in the array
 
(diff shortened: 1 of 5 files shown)


leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
4 more errors to tackle...Looks like I need to look into how fusion works 😩

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
OK I hope I fixed fusion. The PR description is updated. PTAL. I think we already have enough test coverage so we don't need new ones?

cupy/core/_cub_reduction.pyx
```diff
@@ -50,9 +50,9 @@ cdef function.Function _create_cub_reduction_function(
 
     cdef str module_code = _get_cub_header_include()
     module_code += '''
+${type_preamble}
 ${preamble}
 
-${type_preamble}
```
leofang, line 55: This is necessary for fusion to work, as it ensures the input/output types are defined before being used in the `preamble` that is emitted by fusion.

cupy/core/fusion.pyx
```diff
@@ -798,7 +801,7 @@ class _FusionHistory(object):
         else:
             _, reduce_expr, postmap_expr, reduce_ctype = self.reduce_op.routine
             if reduce_ctype is None:
-                reduce_ctype = 'type_in0_raw'
+                reduce_ctype = 'type_out0_raw'
```
leofang, line 804: This does not make any semantics difference if CUB is not in use (note that in fusion `type_in0_raw` and `type_out0_raw` are always identical, for a reason I don't understand yet), but if CUB is turned on then we must follow `ReductionKernel`'s expectation that if the reduction type is not specified, we use the output dtype.

@asi1024 It seems you were the last one touching this part of code. Any chance you could recall why `type_in0_raw ` was used here?

cupy/core/fusion.pyx
```diff
@@ -815,8 +818,10 @@ class _FusionHistory(object):
                                      for s, t in zip(out_cvars, out_params))
 
             submodule_code += self._emit_premap_code(in_params, operation)
-            submodule_code += 'typedef {} type_in0_raw;\n'.format(
-                postmap_ctype)
+            use_cub = ACCELERATOR_CUB in _accelerator._reduction_accelerators
+            if not use_cub:
+                submodule_code += 'typedef {} type_in0_raw;\n'.format(
+                    postmap_ctype)
```
leofang, line 824: In general CUB expects a different `type_in0_raw`, so without doing this there'd be a redeclaration error. I consider this an imperfect workaround, but it leads to the smallest amount of code change IMHO.

cupy/core/_cub_reduction.pyx
leofang: btw this also follows the same order in CuPy's original implementation (`cupy/core/_reduction.pyx`).

cupy/core/fusion.pyx
asi1024: Sorry, that is still under investigation, but I'm now wondering that `reduce_ctype` should be neither `type_in0_raw` nor `type_out0_raw`. `reduce_ctype` is the type of data after pre-map and before post-map. It may be a different type from both input type and output type.

leofang: That's right. If `reduce_ctype` is given, we need to use it (ex: `max`/`min`); if not given, we use the output dtype (ex: `sum`/`prod`). My change would make fusion in line with everywhere in the codebase.

leofang: In the fusion case, though, I think it's expected that the type before and after postmap is identical, another part I don't understand but verify via tests.

emcastillo:
Jenkins, test this please

cupy/core/_reduction.pyx
```diff
@@ -506,7 +506,8 @@ cdef class _SimpleReductionKernel(_AbstractReductionKernel):
 
     cdef:
         readonly _kernel._Ops _ops
-        readonly _preamble
+        readonly str preamble
+        readonly str _preamble
```
leofang, line 510: @emcastillo Let me try merging these two to `preamble`. The underscored version was used in fusion, but in ReductionKernels the one without the underscore is used.

emcastillo: no prob! thanks

leofang:
Jenkins, test this please

cupy/core/_reduction.pyx
leofang: Done in 514042f.

leofang:
btw I added a quick benchmark for CUB-backed fusion to the PR description.

emcastillo (approved):

leofang:
Thanks @emcastillo @asi1024!

Issue #4237, opened 2020-11-07
Add batch cholesky solve


dmargala:
I have a batch cholesky solver implementation that uses `cupy.cuda.cusolver.dpotrfBatched` and `cupy.cuda.cusolver.dpotrsBatched`. It's very similar to `cupyx.linalg._solve._batched_invh`, replacing the identity matrix used on the rhs of Ax = b with an argument for the user to supply a batch of vectors b.

I see a pretty good speed up using batch cholesky solve for my use. Would it be useful to add this to the public api somewhere?

Batch cholesky factorization is currently available via `cupy.linalg.cholesky` (`cupy.linalg._decomposition._potrf_batched`) but it doesn't seem like there is a nice public option for the cholesky solve step.

I checked for a cholesky solver in the numpy API but did not find one. scipy has `scipy.linalg.cho_factor` and `scipy.linalg.cho_solve`.

kmaehashi:
Thanks! Support for `cho_solve`  is currently ongoing here: https://github.com/cupy/cupy/pull/4172

PR #4172, opened 2020-10-22
[WIP] add cho_solve 

daxiongshu:
This PR is to add `cho_solve` like [`scipy.linalg.cho_solve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cho_solve.html). It is based on `cupy.linalg.cholesky` and existing `cusolver` APIs.

PR #4291, opened 2020-11-17
add batched cholesky solver

dmargala:
This PR adds a public function `cupyx.linalg.cholesky_solve` (#4237) that computes the solution to a system of linear equations A * X = B for symmetric/Hermitian positive-definite matrices. Similar to the LAPACK xPOSV routines, it performs factorization and then uses the factored form to solve the system of equations. This PR is mostly a refactor of `cupyx.linalg.invh`, which is implemented as the special case A * X = I. 

There is related work in #4172 to implement `cupyx.scipy.linalg.cho_factor` and `cupyx.scipy.linalg.cho_solve` which are wrappers for the non-batched cholesky factorization step (xPOTRF) and cholesky solve step (xPOTRS) and match the scipy api. 

It might make more sense to separate the factorization and solution steps here too but since this was mostly just a refactor of `cupyx.linalg.invh`, I didn't want to change too much.


anaruse:
Thanks for the PR to generalize `cupyx.linalg.invh` to cholesky solver.

It's a bit annoying that there is no function in `numpy.linalg` or `scipy.linalg` that corresponds to the cholesky solver, but since it it the LAPACK posv equivalent, I thinks it is better to put it in `cupyx.lapack` as `posv`.

For reference, there are `gesv` and `gels` in `cupyx.lapack` now.

dmargala:
Thanks for the suggestion, I moved the cholesky solver to `cupyx.linalg.posv`.

It looks like `scipy.linalg.solve(..., assume_a='pos', ...)` calls `posv` here: https://github.com/scipy/scipy/blob/v1.5.4/scipy/linalg/basic.py#L243

In general though, it doesn't seem like the scipy api supports batch inputs which is really what I am trying to accomplish here.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi:
LGTM!

PR #4296, opened 2020-11-17
Make `CUPY_CUB_VERSION_CODE` defined only for the CUB module

leofang:
Just a nitpick...


kmaehashi:
Is there any harm if `CUPY_CUB_VERSION_CODE` is defined in all modules?
I feel controlling the visibility of all `define_macros` for each module depending on whether it is used or not is a bit cumbersome.

leofang:
No harm at all, it's just a nitpick 😅 I don't think I made it right in my first pass, and I agree it's a bit cumbersome, so let's close it! Thanks for replying early @kmaehashi!

PR #4292, opened 2020-11-17
Use `assert` statement instead of `self.assert*` methods

toslunar:

PR #4308, opened 2020-11-20
`DeprecationWaring` on truth value on empty array

### LINKED ISSUE #4249: Deprecate truth value of empty array
> Truth testing of an empty array is deprecated. To check if an array is not empty, use array.size > 0.

https://numpy.org/doc/stable/release/1.14.0-notes.html#deprecations

```
$ python -qWerror
>>> import cupy
>>> bool(cupy.array([]))
False
>>> import numpy
>>> bool(numpy.array([]))
DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
```

takagi:
Close #4249.

This PR makes an empty array warn `DeprecationWarning` for its truth value as NumPy does.

### DIFF
--- a/tests/cupy_tests/core_tests/test_ndarray_unary_op.py
+++ b/tests/cupy_tests/core_tests/test_ndarray_unary_op.py
@@ -13,7 +13,8 @@ class TestArrayBoolOp(unittest.TestCase):
 
     @testing.for_all_dtypes()
     def test_bool_empty(self, dtype):
-        assert not bool(cupy.array((), dtype=dtype))
+        with testing.assert_warns(DeprecationWarning):
+            assert not bool(cupy.array((), dtype=dtype))
 
     def test_bool_scalar_bool(self):
         assert bool(cupy.array(True, dtype=numpy.bool))
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -964,6 +964,10 @@ cdef class ndarray:
 
     def __nonzero__(self):
         if self.size == 0:
+            msg = ('The truth value of an empty array is ambiguous. Returning '
+                   'False, but in future this will result in an error. Use '
+                   '`array.size > 0` to check that an array is not empty.')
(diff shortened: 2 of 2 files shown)


toslunar (changes requested):
Could you add `testing.assert_warns` to the `bool` tests in `tests/cupy_tests/core_tests/test_ndarray_unary_op.py`?

takagi:
Sure!

tests/cupy_tests/core_tests/test_ndarray_unary_op.py
```diff
@@ -13,7 +13,8 @@ class TestArrayBoolOp(unittest.TestCase):
 
     @testing.for_all_dtypes()
     def test_bool_empty(self, dtype):
-        assert not bool(cupy.array((), dtype=dtype))
+        with testing.assert_warns(DeprecationWarning):
+            assert not bool(cupy.array((), dtype=dtype))
```
toslunar, line 17: #4292 replaced `assertTrue` and `assertFalse` with the equivalent checks, but the intention of the original tests were probably `assert bool(...) is True` or `is False`.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue

toslunar (changes requested):

tests/cupy_tests/core_tests/test_ndarray_unary_op.py
toslunar: But it is irrelevant to the addition of `DeprecationWarning`, which might not be backported.  I'll fix it in another PR.

toslunar (approved):

toslunar:
Jenkins, test this please.

tests/cupy_tests/core_tests/test_ndarray_unary_op.py
takagi: Yeah, I also thought of that.

PR #4307, opened 2020-11-20
Refactor ndimage_tests

toslunar:
Requires #4304.

PR #4311, opened 2020-11-21
cupyx.ndimage.measurements: add center_of_mass, histogram, labeled_comprehension

grlee77:
This PR adds three functions to `cupyx.scipy.ndimage.measurements`

- `center_of_mass`
- `labeled_comprehension`  (applies an arbitrary function over each labeled region)
- `histogram` (implemented via `labeled_comprehension`)

One difference in behavior was made to avoid the need for object arrays (in this case, an arrays where each element is an array). This is used for histogram which calls `labeled_comprehension` using `dtype=object` so that each element of the object array is the histogram over a particular label. For the CuPy implementation, a list of arrays is returned instead of an object array.

If the return value is a scalar in the scipy implementation, a 0-dimensional device array is returned instead here to avoid device->host transfers as much as possible.

I also suppressed all instances of `cupy._util.PerformanceWarning` occuring in the tests of this module that existed prior to this PR.

### DIFF
--- a/cupyx/scipy/ndimage/__init__.py
+++ b/cupyx/scipy/ndimage/__init__.py
@@ -46,6 +46,7 @@ from cupyx.scipy.ndimage.measurements import minimum_position  # NOQA
 from cupyx.scipy.ndimage.measurements import maximum_position  # NOQA
 from cupyx.scipy.ndimage.measurements import median  # NOQA
 from cupyx.scipy.ndimage.measurements import extrema  # NOQA
+from cupyx.scipy.ndimage.measurements import center_of_mass  # NOQA
 
 from cupyx.scipy.ndimage.morphology import generate_binary_structure  # NOQA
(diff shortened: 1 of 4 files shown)


emcastillo:
sorry to ask this, but could you please align the tests to those of https://github.com/cupy/cupy/pull/4307?

Thanks!

grlee77:
> sorry to ask this, but could you please align the tests to those of #4307?

Sure. I will do it for the recent `fourier_ellipsoid` PR as well.

grlee77:
rebased to fix merge conflict. should be ready for testing

emcastillo (approved):
LGTM

emcastillo:
TODO: Add docs

emcastillo:
Jenkins, test this please

grlee77:
> TODO: Add docs

I have already added this one to the API docs.

Issue #4247, opened 2020-11-09
allow single precision operation in the cupyx.scipy.ndimage module


grlee77:
## allow single-precision computations in cupyx.scipy.ndimage

I opened this PR to discuss the potential for performing computations in single precision in `cupyx.scipy.ndimage` functions when the input is single precision. Currently, the internal `weights` arrays used in the convolution function, etc. are always `np.float64`, which does match the internal use of double precision in `scipy.ndimage`. However, to improve performance and reduce memory usage, it would be preferable to use single precision for the computations when possible.

A similar exception was made previously for `cupy.fft` where inputs are NOT automatically promoted to double precision as in NumPy.

I see three ways this can potentially be done (I am willing to make the necessary changes, but need to know what approach would be preferred).

- automatically promote types to at least float32 via: `cp.promote_types(image, cupy.float32)` instead of just using `cupy.float64` everywhere  (e.g. for the `weights` arrays in convolution)
- leave the default internal floating point type float64, unless a boolean keyword-only argument (e.g. `allow_float32`) is specified
- add a global cupy configuration variable to control whether single-precision computation is allowed within the cupyx.scipy.ndimage module

### related issue: complex dtype support
A separate consideration is adding support for complex-valued inputs. I have open PRs for this in SciPy for the filtering and interpolation functions (scipy/scipy#12725 and scipy/scipy#12812). Hopefully those will be merged for SciPy 1.6.0 and then we can revisit adding complex support here.

cc: @jakirkham, @quasiben

grlee77:
There was some related previous discussion during the review of #2483.

grlee77:
I just realized that I had actually implemented the `allow_float32` keyword-only argument in #4145 which was just merged. If one of the other approaches above is preferred instead then we should change that prior to making a new release that includes this.

jakirkham:
If we are able to use the `dtype`s of the array(s) provided, that would be good.

Do you know if there's an equivalent issue raised for SciPy?

grlee77:
I don't think there is an equivalent issue related to single-precision open on SciPy. For SciPy itself, I think the argument is less compelling because the amount of work required would be substantial and the performance benefits of single precision may not be substantial. 

[...]

PR #4145, opened 2020-10-18
add spline_filter1d and spline_filter to cupyx.scipy.ndimage.interpolation

grlee77:
### Overview

This adds the functions `spline_filter1d` and `spline_filter`. These are used for the prefiltering stage for all interpolation functions in `scipy.ndimage` when order >= 2. I will make a follow-up PR later with the actual spline interpolation support.

### Additional Details

The `ElementwiseKernel` approach used for convolutions in `cupyx.scipy.ndimage.filters` is not applicable for these spline prefilters because they are infinite impulse response (IIR) filters. Instead, a `RawKernel` is used that follows the SciPy implementation closely. The implementation was originally only for batched operation along the last axis (commit c5ab0f5), but I later refactored it (8f6223b) to use strided operation instead as that ends up being faster.

For the strided operation, I borrowed the `row` function and some stride calculation code from @coderforlife's [gist for cupyx.scipy.signal.lfilter](https://gist.github.com/coderforlife/17bf471a519a7d092e82f95f10e8f6e4), so I added a co-author credit to @coderforlife here.

The only departures from the SciPy API should be as follows:
1.) There is an additional keyword-only argument `allow_float32` that allows the kernels to operate in single precision. This defaults to False so that the behavior matches SciPy's double-precision implementation. There is performance and memory benefit to allowing single precision, so I would like to provide the option (I would actually like to do that more broadly in cupyx.scipy.ndimage and can open a separate issue regarding whether to do this and whether it should be enabled by a keyword-only argument or perhaps instead by a global/environment variable).

2.) This function allows operation on complex-valued images while SciPy currently does not. I do have an open PR at SciPy to support this (scipy/scipy#12812), but it has not yet been reviewed. If this is acceptable, I can expand the tests here to include complex-valued inputs as well.

The changes made in 347c8c6 truncate the terms z**N for values of N less than the full array size in order to reduce computation time (here, `z`, are the filter poles which are typically < 0.5 so that z**N rapidly approaches zero for large N). I may make a corresponding change upstream in SciPy to improve the efficiency there as well.

### Benchmark Results
<details>
 <summary>Benchmark Script</summary>

```Python
import numpy as np
import scipy.ndimage as ndi

import cupy as cp
import cupyx.scipy.ndimage as cndi
from cupyx.time import repeat

print("dtype | shape | order | SciPy Time | CuPy Time | Acceleration")
print("------|-------|-------|------------|-----------|-------------")
[...]

PR #4314, opened 2020-11-23
revise dtype handling in cupyx.scipy.ndimage.spline_filter

grlee77:
This PR is related to #4247, but deals only with the recently implemented `spline_filter` functions from #4145.

The first commit here removes the `allow_float32` keyword-only argument which is not present in scikit-image and has not yet appeared in a released CuPy version. ~It also changes the default `output=np.float64` to `output=None` which will autoselect a floating point precision for the output. This allows computations to run in single precision by default on float16, float32 or complex64 inputs. In other words, the functionality previously provided by `allow_float32=True` becomes the default.~ **update**: restored default output to float64 for SciPy consistency. internal dtype behaves as if allow_float32=True.

If this type of single precision dtype handling seems okay, I would like to extend it to other functions in the ndimage module in a follow-up PR.

The second commit here adds tests for complex-valued inputs. These were previously supported, but were untested. Complex support was only recently merged upstream in SciPy and will be present in the upcoming 1.6.0 release, so for now a separate test class is used to allow this to also be tested on older SciPy versions.

### DIFF
--- a/cupyx/scipy/ndimage/interpolation.py
+++ b/cupyx/scipy/ndimage/interpolation.py
@@ -37,9 +37,8 @@ def _check_parameter(func_name, order, mode):
 def _get_spline_output(input, output):
     """Create workspace array, temp, and the final dtype for the output.
 
-    Differs from SciPy by not always forcing promotion to double precision.
-    The default output dtype will be single precision for single precision
-    inputs.
+    Differs from SciPy by not always forcing the internal floating point dtype
+    to be double precision.
     """
     complex_data = input.dtype.kind == "c"
     if complex_data:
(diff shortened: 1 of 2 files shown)


emcastillo:
For this PR we have decided to change the 
`allow_float32=False` kwarg to `use_intermediate_fp32=True` to set this by default.

Eventually we want to have a common interface to address all these cases using a defined `COMPUTE_MODE` in an env var instead of just adding kwargs :).

How do you think?

grlee77:
I can change the kwarg name and default to `use_intermediate_fp32=True`. 

However, I am interested in extending this to most other functions in `ndimage` (aside from exceptions like binary morphology functions where it would not apply). It doesn't seem like it makes sense to proceed to add this same kwarg to all functions in `ndimage.filters`, `ndimage.interpolation`, etc. if the preferred solution is to move away from this and use a `COMPUTE_MODE` environment variable. Why not just start with the environment variable approach now? Otherwise it seems that we eventually have to go back and deprecate any unwanted keyword arguments.

grlee77:
I reverted the change to `output=None` as a default. Now the output dtype will be consistent with SciPy, with the only difference being that intermediate computations can use float32 internally when the input is single precision.

emcastillo (approved):

emcastillo:
Jenkins, test this please

PR #4320, opened 2020-11-24
Make names of test classes start with `Test`

toslunar:
This coding rule will be required in order to stop inheriting `unittest.TestCase`. https://docs.pytest.org/en/stable/reference.html#confval-python_classes

- This PR overlaps with 61fe60fd479a684ff01c3f699b5d40f125c38044 in #4307.
- This PR does not fix that `cupy_tests/testing_tests/test_condition.py::MockUnitTest::runTest` is collected.

### DIFF
--- a/tests/cupy_tests/core_tests/fusion_tests/test_example.py
+++ b/tests/cupy_tests/core_tests/fusion_tests/test_example.py
@@ -6,7 +6,7 @@ from cupy_tests.core_tests.fusion_tests import fusion_utils
 
 @testing.gpu
 @testing.slow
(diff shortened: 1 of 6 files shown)


toslunar:
To confirm,

`tests/conftest.py`
```python
def pytest_collection_finish(session):
    for item in session.items:
        if type(item).__name__ == 'TestCaseFunction':
            unittestcase = item.listchain()[-2]
            if not unittestcase.name.startswith('Test'):
                print(f'bad name: {item.nodeid}')
```

takagi:
pfnCI, test this please.

takagi:
LGTM!

PR #4326, opened 2020-11-25
Add import test for ROCm

kmaehashi:


toslunar (approved):

toslunar:
OK to skip other CIs.

leofang:
Cool, didn't know it would work without GPU!

PR #4328, opened 2020-11-25
Fix Travis CI warning

toslunar:
```
root: deprecated key sudo (The key `sudo` has no effect anymore.)
```


toslunar:
#4373 will close the issue :smile:

PR #4337, opened 2020-11-25
[DISCUSSION] Should the previous stream in use be remembered?

leofang:
Not sure if this PR counts as a bug fix or an enhancement, but this would definitely lead to a backward incompatible behavior, although it's a minor one.

Consider the following (arguably pathological) snippet:
```python
import cupy as cp


s1 = cp.cuda.Stream()
s2 = cp.cuda.Stream()
print("s1:", s1.ptr)
print("s2:", s2.ptr)

print("current:", cp.cuda.get_current_stream().ptr)
print("using s1...")
s1.use()
print("current:", cp.cuda.get_current_stream().ptr)
print("using s2...")
s2.use()
print("current:", cp.cuda.get_current_stream().ptr)
print("deleting s2...")
del s2
print("current:", cp.cuda.get_current_stream().ptr)
```
Output:
```
s1: 93876262641696
s2: 93876262476608
current: 0
using s1...
current: 93876262641696
using s2...
current: 93876262476608
deleting s2...
current: 0      # <--- not s1
```
Notice in the end `s1` is not restored, but it's not very intuitive to me. If a stream is in use but gets deleted by accident, I'd expect either of the following behaviors:
1. a `RuntimeError` is raised;
2. restoring to the previous stream in use, if any

However, option 1 is not possible because we can't raise in `__del__`, so this PR implements option 2.

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -15,6 +15,10 @@ cdef class _ThreadLocal:
     cdef object current_stream_ref
     cdef list prev_stream_ref_stack
 
+    def __init__(self):
+        self.current_stream_ref = None
+        self.prev_stream_ref_stack = []
+
     @staticmethod
     cdef _ThreadLocal get():
         try:
@@ -174,8 +178,6 @@ class BaseStream(object):
 
     def __enter__(self):
         tls = _ThreadLocal.get()
-        if tls.prev_stream_ref_stack is None:
-            tls.prev_stream_ref_stack = []
         prev_stream_ref = tls.get_current_stream_ref()
         tls.prev_stream_ref_stack.append(prev_stream_ref)
         tls.set_current_stream(self)
@@ -185,7 +187,6 @@ class BaseStream(object):
         tls = _ThreadLocal.get()
         prev_stream_ref = tls.prev_stream_ref_stack.pop()
         tls.set_current_stream_ref(prev_stream_ref)
-        pass
 
     def use(self):
         """Makes this stream current.
@@ -193,6 +194,8 @@ class BaseStream(object):
         If you want to switch a stream temporarily, use the *with* statement.
         """
         tls = _ThreadLocal.get()
+        prev_stream_ref = tls.get_current_stream_ref()
+        tls.prev_stream_ref_stack.append(prev_stream_ref)
         tls.set_current_stream(self)
         return self
 
@@ -291,7 +294,8 @@ class Stream(BaseStream):
         if self.ptr:
             current_ptr = <intptr_t>tls.get_current_stream_ptr()
             if <intptr_t>self.ptr == current_ptr:
-                tls.set_current_stream(self.null)
+                prev_stream_ref = tls.prev_stream_ref_stack.pop()
+                tls.set_current_stream_ref(prev_stream_ref)
             runtime.streamDestroy(self.ptr)
         else:
             current_stream = tls.get_current_stream()


leofang:
I noticed CuPy seems to make a strong assumption on the current behavior. It's even tested:
https://github.com/cupy/cupy/blob/d8f8a31393f5c7d3993cce951fcd29886b02c458/tests/cupy_tests/cuda_tests/test_stream.py#L22-L31
Looks like I should close this PR?

leofang:
nvm, the tested behavior only has 1 stream in use, but "nested streams" as discussed here are not tested.

pentschev:
> Notice in the end `s1` is not restored, but it's not very intuitive to me. If a stream is in use but gets deleted by accident, I'd expect either of the following behaviors:
> 
> 1. a `RuntimeError` is raised;
> 2. restoring to the previous stream in use, if any
> 
> However, option 1 is not possible because we can't raise in `__del__`, so this PR implements option 2.

I have a different opinion of what should happen, I think if a stream is in use and gets deleted it should just be synchronized and you're done with it, no errors need to be raised.

As to what stream to fallback, I don't think restoring the previous stream makes sense, but rather going back to the default stream should be the case. The reason for that is that if you're using several streams it becomes a very tricky thing to keep track what stream is in use after you delete several previously used stream, from a developer's perspective. IOW, `stream.use()` statement seems to me that we're setting the _current_ behavior, meaning we don't do any bookkeeping, and if for whatever reason `use()` is not valid anymore (e.g., due to the stream being deleted), we roll back to the default global behavior of CuPy.

leofang:
> I have a different opinion of what should happen, I think if a stream is in use and gets deleted it should just be synchronized and you're done with it, no errors need to be raised.

I hope you meant synchronized by the user who deletes it, right? We shouldn't do the synchronization internally, or there'd be tons of syncs when streams go out of scope...

> As to what stream to fallback, I don't think restoring the previous stream makes sense, but rather going back to the default stream should be the case. The reason for that is that if you're using several streams it becomes a very tricky thing to keep track what stream is in use after you delete several previously used stream, from a developer's perspective. 

Yes, I think bookkeeping could be an issue despite we do establish a fallback rule here.

> IOW, `stream.use()` statement seems to me that we're setting the _current_ behavior, meaning we don't do any bookkeeping, and if for whatever reason `use()` is not valid anymore (e.g., due to the stream being deleted), we roll back to the default global behavior of CuPy.

On the one hand I quite agree with this rationale, so I could just simply close this PR (or at most add another test for deleting a stream after multiple `stream.use()` calls) 😁 On the other hand this brings up a question when I first started thinking about PTDS (#4322): What happens if `s1` is `stream.ptds` there, and we delete `s2`? Do we fall back to `stream.ptds` or `stream.null`?

pentschev:
> I hope you meant synchronized by the user who deletes it, right? We shouldn't do the synchronization internally, or there'd be tons of syncs when streams go out of scope...

But how do you ensure the stream is synchronized after it goes out-of-scope? Do you need to enforce a default stream/device synchronization?

> Yes, I think bookkeeping could be an issue despite we do establish a fallback rule here.

Do you mean we already have a fallback rule or it would be an issue even if we decide on one?

> On the other hand this brings up a question when I first started thinking about PTDS (#4322): What happens if `s1` is `stream.ptds` here, and we delete `s2`? Do we fall back to `stream.ptds` or `stream.null`?

That's a good point, I think we fall back to `stream.ptds`, although we can try to generalize that better. We could perhaps continue fallback to stream `0`, and that would mean the legacy stream if PTDS is off, otherwise evaluate https://github.com/cupy/cupy/pull/4322/files#diff-c7ca9ae40b6014a65e9d4420604c0e83adb8aadd77011ce04fd2026d74f43a01R59-R60 to true. This isn't a "clean" solution, but we can think of a cleaner way of achieving the same. In either case, the effect would be the same as falling back to `stream.ptds` when it's on, or `stream.null` when it's off.

leofang:
> > I hope you meant synchronized by the user who deletes it, right? We shouldn't do the synchronization internally, or there'd be tons of syncs when streams go out of scope...
> 
> But how do you ensure the stream is synchronized after it goes out-of-scope? Do you need to enforce a default stream/device synchronization?

I was trying to say that
1. We should not call `runtime.streamSynchronize()` inside `Stream.__del__()`, which would lead to unexpected, likely redundant, syncs.
2. If a user chooses to explicitly delete a stream, it's his/her responsibility to sync the stream before doing so.

What do you think?

> > Yes, I think bookkeeping could be an issue despite we do establish a fallback rule here.
> 
> Do you mean we already have a fallback rule or it would be an issue even if we decide on one?

Yes, for using streams as a context manager there is already a rule implemented using a simple stream stack (`prev_stream_ref_stack` in the code), and with this PR I make the rule also applicable when `stream.use()` is called. So now when a new stream is made in use (regardless of how) we append the current stream to the end of the stack to keep track the fallback order. When deleting a stream, we fallback to the last stream in the stack.

> > On the other hand this brings up a question when I first started thinking about PTDS (#4322): What happens if `s1` is `stream.ptds` here, and we delete `s2`? Do we fall back to `stream.ptds` or `stream.null`?
> 
> That's a good point, I think we fall back to `stream.ptds`, although we can try to generalize that better. We could perhaps continue fallback to stream `0`, and that would mean the legacy stream if PTDS is off, otherwise evaluate https://github.com/cupy/cupy/pull/4322/files#diff-c7ca9ae40b6014a65e9d4420604c0e83adb8aadd77011ce04fd2026d74f43a01R59-R60 to true. This isn't a "clean" solution, but we can think of a cleaner way of achieving the same. In either case, the effect would be the same as falling back to `stream.ptds` when it's on, or `stream.null` when it's off.

This is what I had in mind too 👍 But I think the discussion starting at https://github.com/cupy/cupy/pull/4322#discussion_r530074974 might make this implementation cleaner: If we establish the concept that `Stream.null` refers to "CuPy's default stream" (which was the case without PTDS) then the fallback behavior is trivial. Not sure if we should continue the discussion on this point here or in #4322, though.

pentschev:
> I was trying to say that
> 
> 1. We should not call `runtime.streamSynchronize()` inside `Stream.__del__()`, which would lead to unexpected, likely redundant, syncs.
> 2. If a user chooses to explicitly delete a stream, it's his/her responsibility to sync the stream before doing so.
> 
> What do you think?

The CUDA developer in me agrees with you, but I'm not 100% sure my inner-Python developer does. I'm fully onboard with the idea of not enforcing synchronization, as long as people are fine with that too -- and of course documenting that well.

> Yes, for using streams as a context manager there is already a rule implemented using a simple stream stack (`prev_stream_ref_stack` in the code), and with this PR I make the rule also applicable when `stream.use()` is called. So now when a new stream is made in use (regardless of how) we append the current stream to the end of the stack to keep track the fallback order. When deleting a stream, we fallback to the last stream in the stack.

Thanks for clarifying, I really didn't know about this. I do not oppose to this, although I think from a programming perspective I still think this has the potential to increase complexity and user error more than the alternative for just falling back to the default.

> This is what I had in mind too 👍 But I think the discussion starting at [#4322 (comment)](https://github.com/cupy/cupy/pull/4322#discussion_r530074974) might make this implementation cleaner: If we establish the concept that `Stream.null` refers to "CuPy's default stream" (which was the case without PTDS) then the fallback behavior is trivial. Not sure if we should continue the discussion on this point here or in #4322, though.

I think we should first decide on the proper behavior in https://github.com/cupy/cupy/pull/4322#discussion_r530074974 first, I agree that the final solution here depends on how we handle that ultimately.

kmaehashi:
@leofang Sorry I failed to show a direction on this. I think `stream.use()` API without handling a stack can be useful in some usecases, e.g.,

```py
while True:
    data = get_data_from_queue()
    stream1.use()
    # some op
    stream2.use()
    # some op
```

Also, the current semantics of `Stream().use()` is aligned with `Device().use()`, so I think it's better to keep the current behavior.

leofang:
Thanks, @kmaehashi. OK I'll then continue the work in #5172. @pentschev thanks for the nice discussion!

Issue #4148, opened 2020-10-19
RTD environment is very outdated


leofang:
I am seeing a very strange issue like `NameError: name '__file__' is not defined` in RTD, but it's been fixed upstream for quite sometime (https://github.com/cython/cython/issues/1035#issuecomment-327210233). I think Cython 0.28 is too outdated:

https://github.com/cupy/cupy/blob/d2c5d51c222166b7b012b7beba4b6653cccbac89/docs/requirements.txt#L2

@kmaehashi Can I updated this requirement?

leofang:
ref: https://readthedocs.org/projects/cupy/builds/12131511/

leofang:
Closed by #4630.

PR #3835, opened 2020-08-23
ROCm: Fix bugs and test suites to make ROCm/HIP happy - Part 2

leofang:
~~Blocked by #3786.~~ ~~Details to follow.~~

**UPDATE**
- Add default complex<T> constructors: without them Clang would complain under C++11 mode (default for ROCm); this also brings us more in sync with upstream (Thrust)
- Fix Thrust support: we unified the type dispatcher (#3546) but forgot to include HIP's fp16 header and set it enabled, so the entire dispatcher was malfunctioning on ROCm...(oops!)
- Handle `HIP_VISIBLE_DEVICES` properly
- Fix segfault in `deviceGetAttribute()`
- Disable managed memory for ROCm/HIP: it's simply not supported
- Disable adding callbacks to default stream (https://github.com/ROCm-Developer-Tools/HIP/issues/2119)
- Fix `deviceGetByPCIBusId()`
- Fix a few other tests

PR #4338, opened 2020-11-25
Add `cudaLaunchHostFunc`

leofang:
According to the [CUDA Runtime API documentation](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g74aa9f4b1c2f12d994bf13876a5a2498), `cudaStreamAddCallback` will be deprecated in favor of this new function, added since CUDA 10.0:
> This function is slated for eventual deprecation and removal. If you do not require the callback to execute in case of a device error, consider using `cudaLaunchHostFunc`. Additionally, this function is not supported with `cudaStreamBeginCapture` and `cudaStreamEndCapture`, unlike `cudaLaunchHostFunc`.

TODO:
- [x] Skip tests for CUDA 9.x
- [x] Add stubs

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -215,15 +215,11 @@ class BaseStream(object):
             arg (object): Argument to the callback.
 
         .. note::
-            Whenever possible, using the :meth:`launch_host_func` method
+            Whenever possible, use the :meth:`launch_host_func` method
             instead of this one, as it may be deprecated and removed from
             CUDA at some point.
 
         """
-        if runtime._is_hip_environment and self.ptr == 0:
-            raise RuntimeError('HIP does not allow adding callbacks to the '
-                               'default (null) stream')
-
         def f(stream, status, dummy):
             callback(self, status, arg)
 
@@ -238,7 +234,7 @@ class BaseStream(object):
             arg (object): Argument to the callback.
 
         .. note::
-            Whenever possible, using this method is recommended over
+            Whenever possible, this method is recommended over
             :meth:`add_callback`, which may be deprecated and removed from
             CUDA at some point.
 
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -224,6 +224,20 @@ class BaseStream(object):
 
         runtime.streamAddCallback(self.ptr, f, 0)
 
+    def launch_host_func(self, callback, arg):
+        """Launch a callback on host when all queued work is done.
+
+        Args:
+            callback (function): Callback function. It must take only one
+                argument (user data object), and returns nothing.
+            arg (object): Argument to the callback.
+
+        """
+        def f(dummy):
+            callback(arg)
+
+        runtime.launchHostFunc(self.ptr, f, 0)
+
     def record(self, event=None):
         """Records an event on the stream.
 
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -214,6 +214,11 @@ class BaseStream(object):
                 object), and returns nothing.
             arg (object): Argument to the callback.
 
+        .. note::
+            Whenever possible, using the :meth:`launch_host_func` method
+            instead of this one, as it may be deprecated and removed from
+            CUDA at some point.
+
         """
         if runtime._is_hip_environment and self.ptr == 0:
             raise RuntimeError('HIP does not allow adding callbacks to the '
@@ -231,6 +236,16 @@ class BaseStream(object):
             callback (function): Callback function. It must take only one
                 argument (user data object), and returns nothing.
             arg (object): Argument to the callback.
+
+        .. note::
+            Whenever possible, using this method is recommended over
+            :meth:`add_callback`, which may be deprecated and removed from
+            CUDA at some point.
+
+        .. seealso:: `cudaLaunchHostFunc()`_
+
+        .. _cudaLaunchHostFunc():
+            https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EXECUTION.html#group__CUDART__EXECUTION_1g05841eaa5f90f27124241baafb3e856f
 
         """
         def f(dummy):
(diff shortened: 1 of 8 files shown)


tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -93,11 +110,7 @@ def test_get_and_add_callback(self):
         N = 100
         cupy_arrays = [testing.shaped_random((2, 3)) for _ in range(N)]
 
-        if not cuda.runtime.is_hip:
-            stream = cuda.Stream.null
-        else:
-            # adding callbacks to the null stream in HIP would segfault...
-            stream = cuda.Stream()
+        stream = self.stream
```
leofang, line 113: This fixes the mistake I made in #3835 🤦🏻‍♂️

pentschev (commented):
This looks good to me overall, I have a few questions only, but thanks for working on it @leofang !

tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -59,6 +59,25 @@ def test_get_and_add_callback(self):
...
+
+        stream = cuda.Stream.null
+
+        out = []
+        for i in range(N):
+            numpy_array = cupy_arrays[i].get(stream=stream)
+            stream.launch_host_func(
+                lambda t: out.append(t[0]), (i, numpy_array))
+
+        stream.synchronize()
```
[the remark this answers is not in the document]
leofang: This is in fact done in the other test added below. It's just that the stream pointer is wrapped by `ExternalStream` there, but it should achieve what you had in mind.

cupy/cuda/stream.pyx
```diff
@@ -214,16 +214,41 @@ class BaseStream(object):
...
             callback(self, status, arg)
 
         runtime.streamAddCallback(self.ptr, f, 0)
 
+    def launch_host_func(self, callback, arg):
+        """Launch a callback on host when all queued work is done.
+
+        Args:
+            callback (function): Callback function. It must take only one
+                argument (user data object), and returns nothing.
```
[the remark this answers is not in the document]
leofang: Ah, I am following the requirement from `cudaLaunchHostFunc` that the host function should only have 1 argument instead of 3. 

For the test we would like to do there, we could either use `add_callback` + a 3-arg callback, or use this new function `launch_host_func` + a 1-arg callback, with the stream added as part of the arg:
```python
    data = []
    data.append(stream)
    stream.launch_host_func(callback, data)
```
and check the stream pointer in the callback function.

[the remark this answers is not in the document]
pentschev: Got it, I missed the fact that's how `cudaLaunchHostFunc` should behave out of my ignorance of that function, thanks for clarifying!

tests/cupy_tests/cuda_tests/test_stream.py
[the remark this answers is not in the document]
pentschev: Ah right, I should have checked more than what GH showed to see how the stream was created, thanks for pointing that out and sorry for the false alarm.

pentschev (approved):
LGTM, thanks @leofang !

cupy/cuda/stream.pyx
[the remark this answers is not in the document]
leofang: No worries, Peter! I didn't know this function before either. I became aware of it only because your PTDS PR made me wonder if it's possible to test it programmatically other than eyeballing `nvvp`/`nsys`, and I started browsing the Runtime API doc 😄

tests/cupy_tests/cuda_tests/test_stream.py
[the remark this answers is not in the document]
leofang: Thanks for checking! It's always good to have extra pairs of eyes 🙏

takagi:
pfnCI, test this please.

cupy_backends/cuda/api/runtime.pyx
```diff
@@ -809,6 +823,21 @@ cpdef streamAddCallback(intptr_t stream, callback, intptr_t arg,
     check_status(status)
 
 
+cpdef launchHostFunc(intptr_t stream, callback, intptr_t arg):
+    if _is_hip_environment:
+        raise RuntimeError('This feature is not supported on HIP')
+    if CUDA_VERSION < 10000:
+        raise RuntimeError('This feature is only supported on CUDA 10.0+')
```
leofang, line 830: @takagi by inspecting the generated C file I noticed Cython (at least 0.29.21 that I'm using) has a nice property that for the `if` conditions that can be determined at compile time, Cython will be smart enough to determine the rest of code is dead and eliminate it! (So for RTD and HIP, this function actually stops right after raising.) As a result, it's fine even if we don't add any stubs (for example I forgot to add stubs for CUDA 9.2 😂) Can we rely on this behavior?

takagi: Sounds so nice! Would you check if it works with Cython 0.28.0 as, currently, CuPy requires Cython 0.28.0 or later to build it from its source?

leofang: Hi @takagi, I just checked Cython 0.28.0 will also eliminate the dead code! This is the warning thrown during cythonizing (in both versions): 
```
[50/59] Cythonizing cupy_backends/cuda/api/runtime.pyx
warning: cupy_backends/cuda/api/runtime.pyx:832:16: Unreachable code
```
Though I think 0.28 is too outdated and should be avoided as much as possible (#4148).

leofang:
Jenkins, test this please

cupy_backends/cuda/api/runtime.pyx
takagi: Can I ask you to post a new issue so other core members can find that?

leofang: No problem, @takagi, see #4393.

takagi: Thanks!

takagi:
LGTM!

leofang:
Thanks!

leofang:
Thanks @takagi @pentschev!

PR #4340, opened 2020-11-26
Drop support for Python 3.5 in CuPy v9

kmaehashi:


emcastillo:
Jenkins, test this please

leofang:
Does this mean we can start using f-strings in the codebase now? 😇

kmaehashi:
Yes, but I'd suggest avoiding it if the code needs to-be-backported, otherwise we need to edit manually before backport.

Issue #4290, opened 2020-11-17
Support `cupyx.jit` for custom kernel definition


asi1024:
Task list:
- [x] Add `cupy.vectorize` #4135
- [ ] Support compile-time constants #4241
- [ ] Support function calls (math library functions, cupy ufuncs)
- [ ] Explicit typecasts
- [ ] for-loops
- [ ] raw-kernel from python function
- [ ] ndarray/indexer arguments
- [ ] syncthreads, shared_memory, threadIdx
- [ ] *args, **kwargs, default arguments, keyword-only arguments
- [ ] recursive function
- [ ] lambda function
- [ ] lambda function
- [ ] decorators
- [ ] Python classes, class methods

leofang:
Is it possible to prioritize for "raw cuda kernel code output"? Might be useful for early adopters 🙂

leofang:
btw it'd be great in the future to allow passing compiler options to `cupyx.jit.rawkernel`, similar to what's already possible with `RawKernel` and `RawModule`. I think there are real interests from performance analysis perspective (with `-use_fast_math`, `-lineinfo`, `-v -warn-spills`, etc, see https://github.com/cupy/cupy/issues/3185#issuecomment-649064011 for example).

kmaehashi:
This is nice-to-have: it would be great if source Python code can be embedded as comment in generated CUDA code so that users can easily debug casting issues etc. with #5017 API.

PR #4347, opened 2020-11-26
Support ufunc call in CuPy JIT

asi1024:
Part of #4290.


emcastillo:
I am a bit concerned with the ufunc being a `__global__` function in the case of most of the declared ufuncs in cupy.
Although in CuPy `__calling__` a global func, from inside another `__global__` func is working. It is technically not allowed in CUDA. We should resort to the `<<<...>>>` syntax. @leofang any thoughts here?

asi1024:
This PR supports `ufunc` only for scalar operation.

CuPy JIT is currently used only from `cupy.vectorize`, and we don't need to consider supporting for ndarray input. Also, I have no plan for ndarray input support in the future also for CuPy JIT raw kernel definition.

leofang:
Not sure if I got the gist here. Does this PR allow the JIT to emit something equivalent of this?
```C++
template<typename T>
__global__ void my_ufunc(const T& a, T& b) {
    b += a;
}

__global__ void my_jit_ker(const float* a, float* b, unsigned int N) {
    unsigned int tid = ...
    if (tid < N) {
        my_ufunc(a[tid], b[tid]);
    }
}
```
If so, the concern would be the call to `my_ufunc` is illegal as it lacks of the launch syntax `<<< >>>`. NVRTC happens to allow this and it seems to work fine, but both NVCC and HIPRTC will fail to compile it.

asi1024:
@leofang No, ufunc is represented with device function!

asi1024:
Now ufunc call is supported.
In this PR, ufunc call is represented with device function in the generic case, but in special cases, inline expansion is enabled for the readability of generated CUDA code.

For example, `cupy.add` uses inline expansion.
```py
__device__ double f(double x, double y) {
  return (x + y);
}
```

On the other hand, `cupy.remainder` does not use inline expansion because `in0` occurs twice.
```py
__device__ double cupy_remainder_float64(T0 in0, T1 in1) {
    return in0 - _floor_divide(in0, in1) * in1;
}

__device__ double f(double x, double y) {
  return cupy_remainder_float64(x, y);
}
```

asi1024:
PTAL.

cupyx/jit/_compile.py
```diff
@@ -172,19 +191,43 @@ def _eval_operand(op, args, env, dtype=None):
...
+                expr = expr.replace(f'in{i}', x.code)
+            expr = '(' + expr.replace('out0_type', str(out_type)) + ')'
+            env.preambles.add(ufunc._preamble)
+        else:
+            template_typenames = ', '.join([
+                f'typename T{i}' for i in range(ufunc.nin)])
+            ufunc_name = f'{ufunc.name}_{str(numpy.dtype(op.out_types[0]))}'
+            params = ', '.join([f'T{i} in{i}' for i in range(ufunc.nin)])
+            ufunc_code = f"""template <{template_typenames}>
+__device__ {out_type} {ufunc_name}({params}) {{
```
leofang, line 221: Do we wanna inline it?
```suggestion
__device__ __forceinline__ {out_type} {ufunc_name}({params}) {{
```

cupyx/jit/_compile.py
```diff
@@ -346,9 +401,23 @@ def _transpile_expr(expr, env):
...
+}
+
+
+def _astype_scalar(x, ctype, unsafe=False):
+    from_t = x.ctype.dtype
+    to_t = ctype.dtype
+    if from_t == to_t:
         return x
+    if not unsafe:
+        if _scalar_kind_score[from_t.kind] > _scalar_kind_score[to_t.kind]:
```
leofang, line 419: Would it be better to just rely on `cp.can_cast` to avoid corner cases?

cupyx/jit/_compile.py
asi1024: The inline expansion I commented on https://github.com/cupy/cupy/pull/4347#issuecomment-766612103 is for the readability of generated code, not for the performance improvement. Do you think `__forceinline__` is needed for the performance?

emcastillo: I wonder the real effect of this.

leofang: Let's go without it and trust the compiler's heuristics! 🙂

emcastillo:
Jenkins, test this please

emcastillo (approved):

asi1024:
Jenkins, test this please.

emcastillo:
Jenkins, test this please

asi1024:
Jenkins, test this please.

emcastillo (approved):

emcastillo:
Seems like some very light precision issues

asi1024:
Jenkins, test this please.

PR #4355, opened 2020-11-27
Apply fixes to the FFT callback module

### LINKED ISSUE #4352: import cupy 9.0.0a2 doesn't work on Windows 10
Thank you for using and contributing to CuPy!

If your issue is a **request for support in using CuPy**, please post it on [Stack Overflow](https://stackoverflow.com/questions/tagged/cupy) or Google Groups ([en](https://groups.google.com/forum/#!forum/cupy) / [ja](https://groups.google.com/forum/#!forum/cupy-ja)).
Developer team members are monitoring questions on these channels.

If it is a bug report, please include the following information:

  - CuPy version **9.0.0a2**
  - OS/Platform: **Windows 10**
  - CUDA version: **10.2**
  - cuDNN/NCCL version (if applicable)
* Code to reproduce
```python
>>> import cupy
```
```bash
  File "C:\Users\Amin\anaconda3\envs\py38\lib\site-packages\cupy\__init__.py", line 57, in <module>
    from cupy import fft  # NOQA
  File "C:\Users\Amin\anaconda3\envs\py38\lib\site-packages\cupy\fft\__init__.py", line 1, in <module>
    from cupy.fft._fft import fft  # NOQA
  File "C:\Users\Amin\anaconda3\envs\py38\lib\site-packages\cupy\fft\_fft.py", line 9, in <module>
    from cupy.fft import config
  File "C:\Users\Amin\anaconda3\envs\py38\lib\site-packages\cupy\fft\config.py", line 13, in <module>
    from cupy.fft._callback import get_current_callback_manager  # NOQA
  File "cupy\fft\_callback.pyx", line 35, in init cupy.fft._callback
AttributeError: 'NoneType' object has no attribute 'split'
```

leofang:
Close #4352. This PR adds a few extra safes:
1. For Windows, we do not build the module `cupy.fft._callback` as it cannot be used on Windows anyway 
2. We defer the initialization of the module-level variables, so even if there's any error it won't happen at import time
3. We honor the `CXX` environment variable if `sysconfig` cannot find `CXX` (I don't understand how's that possible, but at least it's the root cause of the reported error).

### DIFF
--- a/cupy/fft/config.py
+++ b/cupy/fft/config.py
@@ -9,9 +9,12 @@ from cupy.fft._cache import get_plan_cache_max_memsize  # NOQA
 from cupy.fft._cache import set_plan_cache_max_memsize  # NOQA
 from cupy.fft._cache import show_plan_cache_info  # NOQA
 
-# expose callback handles to this module
-from cupy.fft._callback import get_current_callback_manager  # NOQA
-from cupy.fft._callback import set_cufft_callbacks  # NOQA
+# on Linux, expose callback handles to this module
+import sys
+if sys.platform.startswith('linux'):
+    from cupy.fft._callback import get_current_callback_manager  # NOQA
+    from cupy.fft._callback import set_cufft_callbacks  # NOQA
+del sys
 
 
 enable_nd_planning = True
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -32,26 +32,22 @@ from cupy.cuda.cufft import getVersion as get_cufft_version
 
 
 # information needed for building an external module
-cdef list _cc = sysconfig.get_config_var('CXX').split(' ')
-cdef str _python_include = sysconfig.get_path('include')
+cdef list _cc = []
+cdef str _python_include = None
 cdef list _nvcc = []
-cdef str _cuda_path = get_cuda_path()
-cdef str _cuda_include = None  # workaround for Read the Docs...
+cdef str _cuda_path = None
+cdef str _cuda_include = None
 cdef str _nvprune = None
-if _cuda_path is not None:
-    _cuda_include = _cuda_path + '/include/'
-    _nvprune = os.path.join(_cuda_path, 'bin/nvprune')
-    if not os.path.isfile(_nvprune):
-        _nvprune = None
-cdef str _build_ver = str(get_build_version())
-cdef int _cufft_ver = get_cufft_version()
+cdef str _build_ver = None
+cdef int _cufft_ver = 0
 cdef str _cupy_root = None
 cdef str _cupy_include = None
 cdef str _source_dir = None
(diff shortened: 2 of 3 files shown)


leofang:
Jenkins, test this please

cupy/fft/config.py
```diff
@@ -9,9 +9,12 @@
 from cupy.fft._cache import set_plan_cache_max_memsize  # NOQA
 from cupy.fft._cache import show_plan_cache_info  # NOQA
 
-# expose callback handles to this module
-from cupy.fft._callback import get_current_callback_manager  # NOQA
-from cupy.fft._callback import set_cufft_callbacks  # NOQA
+# on Linux, expose callback handles to this module
+import sys
+if sys.platform.startswith('linux'):
```
emcastillo, line 14: I wonder if this will work with WSL?

emcastillo: This is defined the same in https://github.com/cupy/cupy/blob/master/install/build.py#L14 so it should work fine.

leofang: > I wonder if this will work with WSL?

Good question, but I do not know 😂 @mnicely @llukas Does cuFFT callback work with WSL?

leofang:
Regardless whether cuFFT callback works with WSL or not, I think we should disable it until someone requests for it (and can verify it really works). I don't think any of the CIs runs on Windows?

emcastillo:
Let's merge this to fix the builds,

I have in my checklist to test cupy in wsl, but the needed kernel is only in the insider's program which I am not registered in.

emcastillo (approved):

mnicely:
> > > I wonder if this will work with WSL?
> > 
> > 
> > Good question, but I do not know joy @mnicely @llukas Does cuFFT callback work with WSL?
> 
> Regardless whether cuFFT callback works with WSL or not, I think we should disable it until someone requests for it (and can verify it really works). I don't think any of the CIs runs on Windows?

Although I haven't tested it, I can't see a reason why it shouldn't work :smile:

leofang:
> > > > I wonder if this will work with WSL?
> > > 
> > > 
> > > Good question, but I do not know joy @mnicely @llukas Does cuFFT callback work with WSL?
> > 
> > 
> > Regardless whether cuFFT callback works with WSL or not, I think we should disable it until someone requests for it (and can verify it really works). I don't think any of the CIs runs on Windows?
> 
> Although I haven't tested it, I can't see a reason why it shouldn't work 😄

A little update. I forgot that WSL is a variant of Ubuntu, so an upshot is things would just work there. If we detect a WSL environment following the approaches below
- https://github.com/microsoft/WSL/issues/423
- https://github.com/microsoft/WSL/issues/4071
- https://www.scivision.dev/python-detect-wsl/

Then we probably could make callbacks work. But, until someone comes and requests this support let's stop here due to lack of development envs and CIs 🙂

PR #4357, opened 2020-11-27
Update CUDA Array Interface to v3 - Part 1

leofang:
~~This PR is blocked by #4322 because we need to know how to handle per-thread default streams (PTDS).~~

**UPDATE**: This PR excludes PTDS from consideration as it's currently not supported in the codebase, see https://github.com/cupy/cupy/pull/4357#issuecomment-735852102. In the coming "Part 2" PR it'll be properly addressed.

I am propagating the upstream change in the CAI protocol https://github.com/numba/numba/pull/5162 to CuPy. The most notable change in the update to v3 is the requirement of **stream synchronization**. CAI v3 specifies that
- Producer should include the stream pointer (whenever applicable) on which the data pointed by CAI can be safely operated 
- Consumer by default must synchronize the stream, if given, at the interoperating point
- Either Producer or Consumer can provide options to Users to overwrite this behavior (the protocol does ***not*** specify how this option should be implemented)

For the detailed definitions (Producer, Consumer, User, etc), see the [CAI v3 documentation](https://numba.readthedocs.io/en/latest/cuda/cuda_array_interface.html); note the capitalized nouns follow the CAI's definitions. 

Accompanying numba/numba#5162 Numba introduces a new environment variable `NUMBA_CUDA_ARRAY_INTERFACE_SYNC` to avoid syncing when acting as a Consumer. Following this path, I introduce two environment variables for advanced users to overwrite the sync behaviors, both of which are defaulted to 1 to make it compliant with CAI v3. However, due to the mismatched concepts on "default streams" between Numba and CuPy, the effect is slightly different. 

In a nutshell, they allows us to fully restore the old (status quo) behavior, *as if the v3 update does not exist*. Specifically, 
- As a Producer, CuPy would not export any stream if `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_STREAM` is set to 0
- As a Consumer, CuPy would not synchronize over any external streams provided through CAI if `CUPY_CUDA_ARRAY_INTERFACE_SYNC` is set to 0

This should make it as performant and give Users the full control as before (if so desired), while taking care the need of certain libraries (notably, mpi4py) in which none of CUDA API is accessible and thus the required synchronization cannot be performed. 

The v3 protocol also made it clear about Users' responsibility of maintaining the lifetime of GPU arrays and streams for the purpose of utilizing CAI, so we can safely assume any given external stream is valid.

**UPDATE 2**: To make it easier to review, note the 3 test files touched in this PR examine different (though arguably overlapping) aspects of the CAI according to my interpretation:
- `tests/cupy_tests/core_tests/test_ndarray.py`: Check CuPy's behavior as a Producer
- `tests/cupy_tests/creation_tests/test_from_data.py`: Check CuPy's behavior as a Consumer
- `tests/cupy_tests/core_tests/test_ndarray_cuda_array_interface.py`: Ensure various operations are correctly done when CAI is in play

cc: @jakirkham @pentschev


kmaehashi:
We're thinking of backporting this to v8 to allow consuming CAI v3 safely in CuPy v8, but excluding PTDS support.

pentschev:
> We're thinking of backporting this to v8 to allow consuming CAI v3 safely in CuPy v8, but excluding PTDS support.

Sounds reasonable to me, @leofang I'm fine if you want to push this forward before we merge PTDS.

leofang:
Sounds good. So perhaps this PR should split into two: the first one handles CAI v3 without PTDS (as @kmaehashi suggested) and will be backported (hopefully it's straightforward), and the second handles PTDS once #4322 is settled and will *not* be backported. 

btw, note that the earliest Numba that supports CAI v3 is expected to be released early next year: https://github.com/numba/numba/pull/5162#pullrequestreview-539317820.

cupy/core/core.pyx
```diff
@@ -2363,6 +2378,14 @@ cpdef ndarray _convert_object_with_cuda_array_interface(a):
...
     mem = memory_module.UnownedMemory(ptr, nbytes, a, dev_id)
     memptr = memory.MemoryPointer(mem, 0)
+    # the v3 protocol requires an immediate synchronization, unless
+    # 1. the stream is not set (ex: from v0 ~ v2) or is None
+    # 2. users explicitly overwrite this requirement
+    stream_ptr = desc.get('stream')
+    if stream_ptr is not None:
+        if _util.CUDA_ARRAY_INTERFACE_SYNC:
+            stream = cuda.stream.ExternalStream(stream_ptr)
+            stream.synchronize()
```
leofang, line 2388: This is done mainly to emphasize that `stream_ptr` is likely coming from external libraries. If desired I can simplify it to `runtime.streamSynchronize(stream_ptr)`, which should still allow mock tests I think.

leofang:
@kmaehashi @asi1024 @pentschev @jakirkham @kkraus14 @gmarkall I think this is ready. The PR description is updated. PTAL.

leofang:
Jenkins, test this please

cupy/core/core.pyx
```diff
@@ -167,6 +167,21 @@ cdef class ndarray:
...
+                desc['stream'] = 1  # TODO(leofang): use runtime.streamLegacy
+            else:
+                desc['stream'] = stream_ptr
+        else:
+            # Old behavior (prior to CAI v3): stream sync is explicitly handled
+            # by users. To restore it, we do not export any stream if
+            # CUPY_CUDA_ARRAY_INTERFACE_SYNC is set to 0 (so that other
+            # participating libraries lacking a finer control over sync
+            # behavior can avoid syncing).
+            desc['stream'] = None
```
kkraus14, line 184: If this is set to `None` then it's implied that `self.data.ptr` can be used on any stream safely. I'm not 100% familiar with cupy's execution patterns, but would we need to synchronize here to guarantee that?

cupy/core/core.pyx
```diff
@@ -2363,6 +2378,14 @@ cpdef ndarray _convert_object_with_cuda_array_interface(a):
         dev_id = device.get_device_id()
     mem = memory_module.UnownedMemory(ptr, nbytes, a, dev_id)
     memptr = memory.MemoryPointer(mem, 0)
+    # the v3 protocol requires an immediate synchronization, unless
+    # 1. the stream is not set (ex: from v0 ~ v2) or is None
+    # 2. users explicitly overwrite this requirement
+    stream_ptr = desc.get('stream')
+    if stream_ptr is not None:
```
asi1024, line 2385: Do we need to consider the case of `stream1` = 1 or 2?

cupy/cuda/stream.pyx
```diff
@@ -187,6 +187,9 @@ class BaseStream(object):
         tls.set_current_stream_ref(prev_stream_ref)
         pass
 
+    def __repr__(self):
+        return 'Stream {}'.format(self.ptr)
```
asi1024, line 191: How about enclosing in angle brackets?
```suggestion
        return '<{} {}>'.format(type(self).__name__, self.ptr)
```

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -223,52 +224,95 @@ class TestNdarrayCudaInterface(unittest.TestCase):
...
         assert not iface['data'][1]
-        assert iface['version'] == 2
         assert iface['strides'] is None
         assert iface['descr'] == [('', '<f8')]
+        assert iface['stream'] == 1
+
+
+# TODO(leofang): test PTDS
+@testing.parameterize(*testing.product({
+    'stream': (cuda.Stream.null, cuda.Stream()),
```
asi1024, line 280: In the current implementation, `runtime.streamCreate()` looks called during test case collection. Could you move it to `setUp`?

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -223,52 +224,95 @@ class TestNdarrayCudaInterface(unittest.TestCase):
...
+        assert iface['version'] == 3  # bump this when the protocol is updated!
+        assert (set(iface.keys()) ==
+                set(['shape', 'typestr', 'data', 'version', 'descr',
+                     'stream', 'strides']))
+        assert iface['shape'] == (10,)
+        assert iface['typestr'] == '<f8'
+        assert isinstance(iface['data'], tuple)
+        assert len(iface['data']) == 2
+        assert iface['data'][0] == arr.data.ptr
+        assert not iface['data'][1]
```
asi1024, line 306: How about writing as `assert iface['data'] == (arr.data.ptr, False)`?

cupy/core/core.pyx
asi1024: Yes, it can be simplified to `runtime.streamSynchronize(stream_ptr)`, I think!

cupy/core/core.pyx
leofang: If a user sets `CUPY_CUDA_ARRAY_INTERFACE_SYNC`, IMHO he/she should get the full control over the synchronization. First, a nice thing CuPy has, unlike the notorious TensorFlow for example, is there's no internal streams floating around, so all CuPy streams are explicitly managed by users, and without specifying any stream the default in CuPy is `cupy.cuda.Stream.null` (up to this PR; see the ongoing discussion in #4322).

Moreover, an ability I'd like to have is a way to "forget" the previous stream. Suppose I "launder" the array over 3 libraries (say, CuPy -> mpi4py -> Numba), since mpi4py does not explicitly handle any CUDA stuff (it's delegated to the MPI library), by the time a CuPy array arrives at Numba it still holds the stream from CuPy if we didn't empty it, and Numba could synchronize over it, which is semantically incorrect. What should have been done at the transition points (by the user) is simply
1. before entering mpi4py: to sync the stream (if any),
2. before entering Numba: do either nothing (for a blocking MPI call) or call `MPI.Wait()` (for a non-blocking MPI call),

all of which need to be performed explicitly by the user. Then, the stream in CAI is irrelevant in this picture.

cupy/core/core.pyx
leofang: Strictly speaking, no. We should sync any given stream unless this behavior is explicitly turned off by users. The issue here is at this point we as Consumer could be on a different stream, so we must ensure the data is ready for subsequent operations. From this point of view, 1 or 2 are no different from other streams. 

In previous commits I tried to compare the current stream (in CuPy) with the external stream (from CAI), and decide whether to sync or not, but I felt it's error-prone so I removed it.

I think the CAI v3 protocol gave a [very complex example](https://numba.readthedocs.io/en/latest/cuda/cuda_array_interface.html#interface-requirements) (need to scroll down a bit) that could ruin any attempt of sync policy. For example, a foreign library (Producer) could operate the data on multiple streams, and export a totally different one which internally waits on all aforementioned streams. Then the Consumer must sync if correctness is the priority (or if the Producer's internal is unknown to Users and thus impossible to control).

A potential micro-optimization for the external stream being 1 (`cudaStreamLegacy`) is to rely on its implicit synchronization and don't sync. But again, if we're on a non-blocking stream we're doomed. Am I making sense?

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -223,52 +224,95 @@ class TestNdarrayCudaInterface(unittest.TestCase):
     def test_cuda_array_interface(self):
         arr = cupy.zeros(shape=(2, 3), dtype=cupy.float64)
         iface = arr.__cuda_array_interface__
+        assert iface['version'] == 3  # bump this when the protocol is updated!
```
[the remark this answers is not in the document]
leofang: This is for the future us in case v4 is out 😂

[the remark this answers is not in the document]
leofang: But, on a second thought it's fine to remove it. If we bump the version, this test will fail so that we know it's time to bump.

cupy/cuda/stream.pyx
leofang: Thanks! This is nicer.

tests/cupy_tests/core_tests/test_ndarray.py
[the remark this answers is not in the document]
pentschev: Ohhh, I get it now, it was meant as an alert for future, I thought it was some lingering comment from when you were still working on the PR. 😄 

Feel free to ignore my comment(s) then or remove the comment, no strong opinions either way from my part.

cupy/core/core.pyx
kkraus14: This is then violating the specification: https://github.com/numba/numba/blob/master/docs/source/cuda/cuda_array_interface.rst#interface-requirements

> - No synchronization is required on the part of the Consumer.
> - The Consumer may enqueue operations on the underlying data immediately on any stream.

leofang: Hi @kkraus14 Recall that we agreed upon a few "backdoor" clauses 😁 For example, what's implemented here is to provide options for flexibility:
> Optionally, to facilitate the User relaxing conformance to synchronization semantics:
> - Producers may provide a configuration option to always set stream to None.
> - Consumers may provide a configuration option to ignore the value of stream and act as if it were None or not provided.

See https://numba.readthedocs.io/en/latest/cuda/cuda_array_interface.html#interface-requirements (need to scroll down a bit).

For now, this PR uses `CUPY_CUDA_ARRAY_INTERFACE_SYNC` to control both Producer and Consumer behaviors. Perhaps I could introduce another env var, say, `CUPY_CUDA_ARRAY_INTERFACE_NO_STREAM_EXPORT` for controlling Producer (and let the other env var controlling Consumer)? Would this be better?

Finally, let me reiterate that without Users explicitly setting the env var(s), my PR is strictly compliant with the v3 protocol by default. Without making sure compliance our time invested in the Numba CAI v3 PR would a waste, after all 🙂

kkraus14: Sounds good, thanks for the explanation!

cupy/core/core.pyx
leofang: Hi @asi1024 Looks like I can't mock `cpdef` functions...😢 Should I leave this as is?

tests/cupy_tests/core_tests/test_ndarray.py
leofang: Done in ad95878.

cupy/core/core.pyx
leofang: Done in 56ede6d. I used a positive term `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_STREAM` as I want both env vars to default to 1.

leofang:
Jenkins, test this please

cupy/core/core.pyx
```diff
@@ -157,7 +157,7 @@ cdef class ndarray:
             'shape': self.shape,
             'typestr': self.dtype.str,
             'descr': self.dtype.descr,
-            'version': 2,
+            'version': 3,
```
asi1024, line 160: When `CUDA_ARRAY_INTERFACE_EXPORT_STREAM` is disabled, the `version` should be 2?

cupy/core/core.pyx
asi1024: I feel `CUPY_CUDA_ARRAY_INTERFACE_VERSION` (defaults to `3`) is better. How do you think?

leofang: Hi @asi1024 For this and your other comment, I have a mixed feeling. On the one hand I totally get that you want a flag to say we're complaint up to v3. On the other hand, the expectation for `!=3` is unclear to me (for example, do we error out when a foreign library is on v3 but we're set to v2, or vice versa?), and it's a little less intuitive for Users to know the meaning of `CUPY_CUDA_ARRAY_INTERFACE_VERSION` vs `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_STREAM`/`CUPY_CUDA_ARRAY_INTERFACE_SYNC`. Finally, do we use the same env var to control the behavior of both Producer and Consumer?

Note all the fields added since v0 are optional, so from v0 to v3 didn't really change much, except that the safety of operation is thoroughly considered. Also, AFAIK no other participating libraries check the `version` field.

From my standpoint as long as both of them can be set to opt out by Users, I don't have strong opinion. But I do want to make sure that everyone here is happy with this PR, and that we don't make mistakes that we have to revisit later, as this PR will be backported to a stable release (v8.3?) and we better not mess up 🙂

asi1024: May `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_VERSION` be much better?

cupy/core/core.pyx
asi1024: One workaround I know is defining `def` function just calling `runtime.streamSynchronize` here, but I'm wondering if it causes performance degradation. I think it's OK to leave this as is without mock tests.

cupy/core/core.pyx
asi1024: I think it's OK to synchronize on default and let users decide that synchronization as the current implementation does. :+1:

How about on `cuda.stream.ExternalStream(stream_ptr)` for the case of `stream` = 1 or 2?

cupy/core/core.pyx
leofang: Sounds good, thanks.

cupy/core/core.pyx
leofang: Thanks, @asi1024!

> How about on `cuda.stream.ExternalStream(stream_ptr)` for the case of `stream` = 1 or 2?

Could you clarify on this? Isn't this already done?

cupy/core/core.pyx
leofang: Hi @asi1024 Do you mean we just rename `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_STREAM` to `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_VERSION`, or you have something more in mind?

asi1024: Yes! I mean just renaming the variable.

leofang: Ahhh OK I totally misunderstood 😅 Thanks for bearing with me...

cupy/core/core.pyx
asi1024: Sorry for not explaining enough. In my understanding, More additional handling is needed for these cases, and you are planning to make another PR for complete support. Is it correct?
If so, could you fix to raise `NotImplementedError` in this PR, and restore in the other PR?

leofang: No, actually this is enough. The point here is whatever pointer here, if given, will be synchronized.

cupy/core/core.pyx
leofang: @asi1024 Sorry I am confused again...What would be the allowed values for `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_VERSION`? Only 2 and 3 (default)? Do we also need to fix the value for the key `version`?

cupy/core/core.pyx
asi1024: Sorry, I was misunderstanding something.

cupy/core/core.pyx
asi1024: Yes, only 2 and 3 are supported, and an error will be raised otherwise. The key `version` also need to be fixed.

leofang: Thanks, @asi1024, I updated it. PTAL.

asi1024:
Jenkins, test this please.

asi1024:
LGTM! Thanks!

PR #4362, opened 2020-11-30
Add FlexCI for Windows

kmaehashi:
The first attempt to run CI on Windows.

- [x] reuse cupy kernel cache

### DIFF
--- a/.pfnci/trim_cupy_kernel_cache.py
+++ b/.pfnci/trim_cupy_kernel_cache.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+
+"""
+A tool to clean-up unused CuPy kernel caches.
+
+The expiry is the length to invalidate unused CuPy cache, in seconds.
+For example, setting the expiry to 3,600 will trim all kernel caches
+not being used within an hour.
+
+Note that this code relies on atime support of the filesystem.
+"""
+
+import argparse
+import datetime
+import glob
+import os
+import sys
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--expiry', type=int, required=True)
+    parser.add_argument('--rm', default=False, action='store_true')
+    options = parser.parse_args()
+
+    expiry = datetime.datetime.now() - datetime.timedelta(
+                seconds=options.expiry)
+    sys.stderr.write('Looking for cache unused since {}\n'.format(expiry))
+    expiry_ts = expiry.timestamp()
+
+    total_size = 0
+    trim_size = 0
+    for f in glob.glob(os.path.expanduser('~/.cupy/kernel_cache/*.cubin')):
+        stat = os.lstat(f)
+        total_size += stat.st_size
+        if stat.st_atime < expiry_ts:
+            print(f)
+            trim_size += stat.st_size
+            if options.rm:
+                os.unlink(f)
+    sys.stderr.write('Total: {} bytes\n'.format(total_size))
+    sys.stderr.write('Expired: {} bytes\n'.format(trim_size))
+
+
+if __name__ == '__main__':
+    main()
--- a/.pfnci/trim_cupy_kernel_cache.py
+++ b/.pfnci/trim_cupy_kernel_cache.py
@@ -14,6 +14,7 @@ Note that this code relies on atime support of the filesystem.
 import argparse
 import datetime
 import glob
+import itertools
 import os
 import sys
 
@@ -36,7 +37,9 @@ def main():
             os.path.expanduser('~/.cupy/kernel_cache/'))
     _log('Looking for cache files under {}...'.format(cache_dir))
     records = []
-    for f in glob.glob(os.path.join(cache_dir, '*.cubin')):
+    for f in itertools.chain(
+            glob.iglob(os.path.join(cache_dir, '*.cubin')),
+            glob.iglob(os.path.join(cache_dir, '*.hsaco'))):
         stat = os.lstat(f)
         records.append((stat.st_atime, stat.st_size, f))
     records.sort(reverse=True)  # newest cache first
--- a/.pfnci/trim_cupy_kernel_cache.py
+++ b/.pfnci/trim_cupy_kernel_cache.py
@@ -1,11 +1,12 @@
 #!/usr/bin/env python
 
 """
-A tool to clean-up unused CuPy kernel caches.
+A tool to clean-up least-recently-used CuPy kernel caches.
 
 The expiry is the length to invalidate unused CuPy cache, in seconds.
 For example, setting the expiry to 3,600 will trim all kernel caches
-not being used within an hour.
+not being used within an hour. The maximum number of files or maximum total
+size can also be specified to cap the amount of caches.
 
 Note that this code relies on atime support of the filesystem.
 """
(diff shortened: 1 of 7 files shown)


kmaehashi:
pfnCI, test this please.

leofang:
oops, looks like there's some issue in compiling reduction kernels? 😬 Would be great to have access to the error logs...

leofang:
> oops, looks like there's some issue in compiling reduction kernels? 😬 Would be great to have access to the error logs...

Thanks, @kmaehashi! I was able to extract just enough log, see #4383. The errors in histogram tests seem to have a different origin, probably due to `long long` having different widths on Windows vs Linux.

kmaehashi:
> = 793 failed, 26715 passed, 43281 skipped, 13 xfailed, 1 xpassed, 178 warnings in 2529.09 seconds =

list of failing tests: https://gist.github.com/kmaehashi/eafed20b94025431fa53d97689bdc888/raw/a8926d3ea4216372053098d59e9ae6e00612c9d0/full_log2_list.txt
full log: https://gist.github.com/kmaehashi/eafed20b94025431fa53d97689bdc888/raw/a8926d3ea4216372053098d59e9ae6e00612c9d0/gistfile1.txt

leofang:
Nice. The callback failures should be trivial. I have a branch locally. A large number of failures are CUB block kernel related (including fusion), to be addressed in #4383. The raw tests are puzzling, will take a look later.

leofang:
@kmaehashi Could you try applying this patch and see if we could eliminate the huge number of issues directly or indirectly involving the CUB block reduction kernels (#4383)?
```diff
diff --git a/cupy/core/_cub_reduction.pyx b/cupy/core/_cub_reduction.pyx
index c9d855e1b..6954bd448 100644
--- a/cupy/core/_cub_reduction.pyx
+++ b/cupy/core/_cub_reduction.pyx
@@ -38,11 +38,8 @@ cdef function.Function _create_cub_reduction_function(
         options += ('-I' + _rocm_path + '/include', '-O2')
         backend = 'nvcc'  # this is confusing...
     else:
-        # use jitify + nvrtc
-        # TODO(leofang): how about simply specifying jitify=True when calling
-        # compile_with_cache()?
-        options += ('-DCUPY_USE_JITIFY',)
-        backend = 'nvrtc'
+        options += ('-O2',)
+        backend = 'nvcc'
 
     # TODO(leofang): try splitting the for-loop into full tiles and partial
     # tiles to utilize LoadDirectBlockedVectorized? See, for example,
```

leofang:
All test failures in `test_raw.py` are due to `nvcc fatal   : Cannot find compiler 'cl.exe' in PATH`, so perhaps the simple fix is to set up `PATH` properly?

kmaehashi:
#4397 should fix the `cl.exe` issue.

@leofang Thanks! Tested the patch on #4397 branch and confirmed 17 failures disappeared.

leofang:
> Thanks! Tested the patch on #4397 branch and confirmed 17 failures disappeared.

Cool! So it sounds like a bad interplay among Jitify, NVRTC, and Windows? I took some time earlier to inspect the code but still not sure whom to blame, so perhaps we could apply the patch as a Windows-only workaround for now, and I can continue investigating #4383? 

btw #4385 is in, so we should see another batch of errors gone 🦾

leofang:
btw the patch only killed 17 errors? Would expect more, like ~60 🤔

kmaehashi:
Ah sorry, forgot to write that I only tested parts that seem related.
```
python -m pytest  tests/cupy_tests/core_tests/test_cub_reduction.py tests/cupy_tests/core_tests/test_reduction.py tests/cupy_tests/core_tests/test_ndarray_reduction.py
```
Will run the full test again and share the log.

leofang:
Aha, that makes sense 👍 Thanks.

kmaehashi:
Posted the current results to #4409.

kmaehashi:
We will be merging this PR to enable Windows tests now, although test failures are remaining in Windows. We will work on fixing the issue incrementally.
I've renamed `cupy.win.cuda100` to `cupy.experimental.win.cuda100` to indicate that the test may fail.

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

.pfnci/trim_cupy_kernel_cache.py
```diff
@@ -0,0 +1,82 @@
...
+    parser.add_argument('--expiry', type=int, default=None)
+    parser.add_argument('--max-files', type=int, default=None)
+    parser.add_argument('--max-size', type=int, default=None)
+    parser.add_argument('--print', default=False, action='store_true')
+    parser.add_argument('--rm', default=False, action='store_true')
+    options = parser.parse_args()
+
+    cache_dir = os.environ.get(
+            'CUPY_CACHE_DIR',
+            os.path.expanduser('~/.cupy/kernel_cache/'))
```
emcastillo, line 36: Is this path safe?

.pfnci/trim_cupy_kernel_cache.py
```diff
@@ -0,0 +1,82 @@
...
+    _log('Looking for cache files under {}...'.format(cache_dir))
+    records = []
+    for f in glob.glob(os.path.join(cache_dir, '*.cubin')):
```
emcastillo, line 39: amd uses `*hsaco` files, maybe it is also worth cleaning them

.pfnci/trim_cupy_kernel_cache.py
kmaehashi: Yes, it's the same as we do in cupy: https://github.com/cupy/cupy/blob/065adda84f751f69593540baf7eaccd0c5350e9d/cupy/cuda/compiler.py#L375-L379

.pfnci/trim_cupy_kernel_cache.py
kmaehashi: Nice catch!

kmaehashi:
pfnCI, test this please.

.pfnci/windows/_flexci.ps1
```diff
@@ -0,0 +1,43 @@
...
+        $pydir = "Python38"
+    } elseif ($version -eq "3.9") {
+        $pydir = "Python39"
+    } else {
+        throw "Unsupported Python version: $version"
+    }
+    $Env:PATH = "C:\Development\Python\$pydir;C:\Development\Python\$pydir\Scripts;" + $Env:PATH
+}
+
+function ActivateCUDA($version) {
```
emcastillo, line 18: I guess this is just for dealing with all the configs in the windows image and will ease a lot of the backport effort.
So I don't mind having unsupported versions in here.

kmaehashi:
Test failure of `cupy.experimental.win.cuda100` is expected (we will fix these tests in upcoming PRs)

.pfnci/windows/test.ps1
```diff
@@ -0,0 +1,110 @@
...
+    if ($use_cache) {
+        DownloadCache
+    }
+    echo "Running test..."
+    python -m pytest -rfEX $Env:PYTEST_OPTS tests > cupy_test_log.txt
+    if (-not $?) {
+        $test_retval = $LastExitCode
+    }
+    if ($use_cache) {
+        UploadCache
```
emcastillo, line 92: It looks like a PR with errors will upload their invalid kernels here, forcing everyone to recompile?
I think we need to check for `test_retval` before uploading the cache.

I wonder if it's better to run the tests only once a day and then upload the cache from the master branch, rather than having all the tests uploading their cache.

emcastillo: We can just address this in another PR, lets leave it as TODO :)

emcastillo (approved):

leofang:
Awesome work @kmaehashi! Question: Looks like the log is stored as a build artifact. Is it something that can be downloaded by anyone?

kmaehashi:
Yes, logs are visible for everyone. This is needed because FlexCI omits the stdout/stderr when the output is huge.

leofang:
Ahhhh sorry I was being silly...Didn't notice the logs are pointed to by urls 🤦🏻‍♂️

PR #4363, opened 2020-11-30
Fix dtypes in `cupy.linalg`

### LINKED ISSUE #4309: `cupy.linalg.inv` returns wrong value on integer input
```
>>> cupy.linalg.inv(cupy.array([[1, 2], [3, 4]]))
array([[-inf,  inf],
       [ inf, -inf]])
>>> numpy.linalg.inv(numpy.array([[1, 2], [3, 4]]))
array([[-2. ,  1. ],
       [ 1.5, -0.5]])
>>> cupy.__version__
'9.0.0a1'
```

toslunar:
- Support integer inputs in `cupy.linalg` (close #4309).
- Add `cupy.linalg._util.common_type`: Get dtypes for computation and for NumPy-compatible output.


toslunar:
This PR is WIP because only `_solve.py` in `cupy/linalg` has been fixed.

toslunar:
TODO: Use `cupy.common_type` for consistency in CuPy.

- `cupy.common_type` is equivalent with `numpy.common_type`.  Then, by default, integer types should use the double precision.
- It's another topic how to configure the default precision for integers in `cupy.common_type`.

toslunar:
`numpy.common_type` rejects `numpy.bool_`, but `numpy.linalg.linalg._commonType` accepts it.
```
>>> numpy.common_type(numpy.array(True))
TypeError: can't get common type for non-numeric array
```

```
>>> x = ~numpy.identity(3, numpy.bool_); x
array([[False,  True,  True],
       [ True, False,  True],
       [ True,  True, False]])
>>> numpy.linalg.matrix_rank(x)
3
```

emcastillo:
Some quick thoughts

> cupy.common_type is equivalent with numpy.common_type. Then, by default, integer types should use the double precision.

I agree with this, our current default precision is float64 unless it is explicitly set.

> It's another topic how to configure the default precision for integers in cupy.common_type.

I wonder if we can add an env var here? `CUPY_DEFAULT_PRECISION`

cupy/linalg/_util.py
```diff
@@ -1,10 +1,17 @@
...
 from numpy import linalg
 
 import cupy
+import cupy._util
 from cupy import core
 import cupyx
 
 
+_default_precision = os.getenv('CUPY_DEFAULT_PRECISION')
+
```
emcastillo, line 13: Lets add this to the docs :)

emcastillo (dismissed):

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

toslunar:
I'll fix the test failures (`cupy_tests/test_cusolver.py::TestSyevj`).

toslunar:
Fixed.

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4365, opened 2020-11-30
Fix `rcond` arg of `linalg.lstsq`

toslunar:
- The default is `rcond='warn'`, which fallbacks to `rcond=-1`.
- Negative `rcond` should be treated as the machine precision.
- NumPy's future default `rcond=None` is max(m, n) times machine precision.

### DIFF
--- a/cupy/linalg/_solve.py
+++ b/cupy/linalg/_solve.py
@@ -251,6 +251,9 @@ def lstsq(a, b, rcond='warn'):
 
     if rcond is None:
         rcond = numpy.finfo(s.dtype).eps * max(m, n)
+    elif rcond <= 0 or rcond >= 1:
+        # some doc of gelss/gelsd says "rcond < 0", but it's not true!
+        rcond = numpy.finfo(s.dtype).eps
 
     # number of singular values and matrix rank
     cutoff = rcond * s.max()
--- a/cupy/linalg/_solve.py
+++ b/cupy/linalg/_solve.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy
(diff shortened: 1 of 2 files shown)


takagi:
pfnCI, test this please.

cupy/linalg/_solve.py
```diff
@@ -223,6 +225,18 @@ def lstsq(a, b, rcond=1e-15):
...
+        warnings.warn(
+            '`rcond` parameter will change to the default of '
+            'machine precision times ``max(M, N)`` where M and N '
+            'are the input matrix dimensions.\n'
+            'To use the future default and silence this warning '
+            'we advise to pass `rcond=None`.\n'
+            'Using the old CuPy default: `rcond=1e-15`, which is '
+            'neither the same as the old NumPy default (`rcond=-1`).',
+            FutureWarning)
+        rcond = 1e-15
```
toslunar, line 238: The different default value is considered as a bug.
```suggestion
        rcond = -1
```

Then, the function should also implement the following spec of `*gelss`:

> If RCOND < 0, machine precision is used instead.

toslunar: It seems a reference implementation uses `if rcond >= 0 or rcond <= 1`.  I found the following code in `dlalsd`.
```fortran
       IF( ( rcond.LE.zero ) .OR. ( rcond.GE.one ) ) THEN
          rcnd = eps
       ELSE
          rcnd = rcond
       END IF
```

takagi:
pfnCI, test this please.

takagi:
LGTM! Waiting for CI pass.

PR #4370, opened 2020-11-30
add cupy.quantile

grlee77:
This is a simple PR to add `cupy.quantile` which is the same as `cupy.percentile` but with q in the range [0, 1] rather than [0, 100].

This just reuses the existing `cupy.percentile` implementation, so there is not much new to review. The test cases are also copies of the tests for percentile, but using floating point `q` in the range [0, 1].

`numpy.quantile` was introduced in NumPy 1.15 which I think CuPy now requires, so I didn't do any `with_requires` checks in the tests.

**One area to consider**
I added range checks on `q` as in NumPy, but these require device synchronization, so it may not be desirable. If we don't want this, I can just remove use of `_quantile_is_valid` and the two `*_out_of_range_q` tests.

### DIFF
--- a/cupy/_statistics/order.py
+++ b/cupy/_statistics/order.py
@@ -306,8 +306,8 @@ def quantile(a, q, axis=None, out=None, interpolation='linear',
     """Computes the q-th quantile of the data along the specified axis.
 
     Args:
-        a (cupy.ndarray): Array for which to compute percentiles.
-        q (float, tuple of floats or cupy.ndarray): Percentiles to compute
+        a (cupy.ndarray): Array for which to compute quantiles.
+        q (float, tuple of floats or cupy.ndarray): Quantiles to compute
             in the range between 0 and 1 inclusive.
         axis (int or tuple of ints): Along which axis or axes to compute the
             quantiles. The flattened array is used by default.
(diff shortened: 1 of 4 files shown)


docs/source/reference/statistics.rst
```diff
@@ -1,7 +1,7 @@
 Statistical Functions
 =====================
 
-.. https://docs.scipy.org/doc/scipy/reference/stats.html
+.. https://numpy.org/doc/stable/reference/routines.statistics.html
```
toslunar, line 4: The change to this line will be backported.

cupy/_statistics/order.py
```diff
@@ -286,3 +261,72 @@ def percentile(a, q, axis=None, out=None, interpolation='linear',
...
+
+
+def quantile(a, q, axis=None, out=None, interpolation='linear',
+             keepdims=False):
+    """Computes the q-th quantile of the data along the specified axis.
+
+    Args:
+        a (cupy.ndarray): Array for which to compute percentiles.
+        q (float, tuple of floats or cupy.ndarray): Percentiles to compute
+            in the range between 0 and 1 inclusive.
```
toslunar, line 311:
```suggestion
        a (cupy.ndarray): Array for which to compute quantiles.
        q (float, tuple of floats or cupy.ndarray): Quantiles to compute
            in the range between 0 and 1 inclusive.
```

toslunar (approved):
LGTM

toslunar:
pfnCI, test this please.

PR #4379, opened 2020-12-01
[backport] Fix cufft callback manager error in Windows

kmaehashi:
Backport of #4377

PR #4380, opened 2020-12-01
[backport] Fix cufft callback manager error in Windows

emcastillo:


emcastillo:
Seems it was created in #4379

PR #4395, opened 2020-12-03
Generating CUDA-related library FFI from its headers

### LINKED ISSUE #4358: Generating CUDA-related library FFI from its headers
Rel #3513.

Currently, the FFI for CUDA-related library is hand-written in the files as `cusparse.pyx`, `cusparse.pxd` and `cupy_cusparse.h` for cuSPARSE. It may be time to generate it from its C header files automatically. The process consists of two parts: parsing and generating.

As for parsing, any reasonably matured C parser in any language would be fine. For example, the following gist is the AST of `cusparse.h` parsed with [pycparser](https://github.com/eliben/pycparser) library and dumped in JSON. I think it has enough information for the purpose here.

https://gist.github.com/takagi/21ae00f3a896abd50d4b5d3dcdbeb426

As for generating, we have to give some knowledge to the translating script:
- Which entries to generate and which not to
- Which to .pxd files and which to .pyx files
- Translation rule for names and types
- Stubs for RTD and HIP
- Compatibility with older CUDA Toolkit versions
- Need mixing with hand-written code?
- If possible, also generating the bottommost wrapper layers would be fine

Additionally, we consider the way to include it into our development process:
- Where to place and when to run the script
- How to track the headers of older CUDA Toolkit versions

takagi:
Close #4358.

This PR adds a script to generate CUDA-related library FFI and wrappers from its headers.

Where to have the script in CuPy repository is TBD.


takagi:
This change has a problem that it needs all CUDA Toolkit versions that CuPy supports for generating the compatibility headers, which is especially painful for contributors who make some changes to this generator when they regenerate the outcomes.

https://gitub.com/cupy/cupy/pull/4606 tried to generate the files on build-time for the CUDA Toolkit version for that CuPy is to be built, but it also has a problem that it makes CuPy build depends on pycparser. 

An idea is generating Cython files for each CUDA Toolkit version, document, and possibly ROCm, putting them in the repo statically, and selecting ones on build-time for environment for which CuPy is to be built.

```
cupy_backends/.../libs/11.2/cublas.pyx
                            cublas.pxd
                       11.3/cublas.pyx
                            cublas.pxd
                       stub/cublas.pyx
                       rocm/cublas.pyx
                            cublas.pxd
```

PR #4397, opened 2020-12-03
Inherit environment variable and detect cl.exe automatically

kmaehashi:
* NVCC refers to several environment variables, so it should be inherited from the environment.
* On Windows, it's better to append path to the compiler to allow using nvcc backend without frustration.

See comments in the code for details.

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -1,3 +1,4 @@
+import copy
 import hashlib
 import math
 import os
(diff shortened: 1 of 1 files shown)


leofang:
LGTM. Do I understand it correctly that if `cl.exe` is not found, we rely on the fact that an error would be raised when `nvcc` is invoked?

kmaehashi:
Yes, in that case users will see `nvcc fatal   : Cannot find compiler 'cl.exe' in PATH` error, as far as I tested. But I'm not 100% sure because it's not precisely documented how NVCC locates the compiler on Windows...

leofang (approved):

kmaehashi:
(For later reference) When `TMP`/`TEMP` is not set, `nvcc.exe` dies with exit status 2, without any messages at all.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

kmaehashi:
memorandum (for future reference): when `TMP`/`TEMP` is not set, nvcc.exe dies with exit status 2 without any output.

emcastillo (approved):

PR #4399, opened 2020-12-03
Fix `poly1d`

toslunar:
Do not "maintain" poly1d

The semantics of NumPy's poly1d (legacy polynomial class) is not maintained.
Therefore `__array_priority__` should be removed.


asi1024:
note: https://numpy.org/doc/stable/reference/routines.polynomials.html

asi1024:
LGTM.

asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #4400, opened 2020-12-03
ndimage: support all interpolation boundary modes

grlee77:
## Overview

This PR adds support for additional boundary modes in all interpolation functions. It is based on the most recent bug fixes and enhancements in this area for SciPy 1.6.0 (~12 issues in SciPy were closed: see meta-issue scipy/scipy#12773).

The changes in this PR correspond to those from scipy/scipy#12767 and scipy/scipy#12776 and include:
- enables the previously disabled 'reflect' and 'wrap' modes ('reflect' behavior was fixed in SciPy 1.6. 'wrap' remains as before, but a new 'grid-wrap' mode is probably what most users will want)
- new modes: 'grid-wrap' and 'grid-constant' (also 'grid-mirror' as a synonym for 'reflect')

The easiest way to understand the various modes is probably via the new figures introduced to SciPy's ndimage tutorial. The main doc site doesn't show these yet, but here is documentation artifact link from one of the CI runs:
https://22174-1460385-gh.circle-artifacts.com/0/html-scipyorg/tutorial/ndimage.html#interpolation-boundary-handling

There is also a text-based explanation in the SciPy `mode` docstrings for each individual function. I have not reproduced that here, but can if desired. CuPy users should also be to reach those via our cross-linking to SciPy in the `seealso` sections of our docstrings.
https://22174-1460385-gh.circle-artifacts.com/0/html-scipyorg/generated/scipy.ndimage.affine_transform.html#scipy-ndimage-affine-transform

image from the tutorial reproduced here (Ignore order > 1 for now. Orders 2-5 will be added in a follow-up PR):
plot_boundary_modes

## Additional Details

Many of the new tests require SciPy >=1.6.0 which should have a release candidate any day now. I have tested locally vs. SciPy master.

The kernel generation code would be somewhat simplified if we dropped support for the legacy `'wrap'` mode and only support the newer `'grid-wrap'` mode which is probably what most users would want (see figure above).

### DIFF
--- a/cupyx/scipy/ndimage/_filters_core.py
+++ b/cupyx/scipy/ndimage/_filters_core.py
@@ -303,7 +303,7 @@ def _generate_nd_kernel(name, pre, found, post, mode, w_shape, int_type,
                ws_init=ws_init, ws_pre=ws_pre, ws_post=ws_post,
                loops='\n'.join(loops), found=found, end_loops='}'*ndim)
 
-    mode_str = mode.replace('-', '_') # avoid potential hyphen in kernel name
+    mode_str = mode.replace('-', '_')  # avoid potential hyphen in kernel name
     name = 'cupy_ndimage_{}_{}d_{}_w{}'.format(
         name, ndim, mode_str, '_'.join(['{}'.format(x) for x in w_shape]))
     if all_weights_nonzero:
--- a/cupyx/scipy/ndimage/_filters_core.py
+++ b/cupyx/scipy/ndimage/_filters_core.py
@@ -82,6 +82,8 @@ def _run_1d_filters(filters, input, args, output, mode, cval, origin=0):
     output = _util._get_output(output, input)
     modes = _util._fix_sequence_arg(mode, input.ndim, 'mode',
                                     _util._check_mode)
+    # for filters, "wrap" is a synonym for "grid-wrap".
+    modes = ['grid-wrap' if m == 'wrap' else m for m in modes]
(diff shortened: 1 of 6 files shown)


takagi:
pfnCI, test this please.

takagi:
Thanks for the PR! I think of merging this with `cupy.experimental` mark as the upstream SciPy 1.6 is not released yet.

https://github.com/cupy/cupy/wiki/Development-policies#numpyscipy-compatibility

BTW, do you have information about when SciPy 1.6 is released?

grlee77:
BTW, do you have information about when SciPy 1.6 is released?

A release candidate should be out any day now. I would guess early Jan for the final release, but don't have a specific date. The original plan was to release in December, but things have gone a bit slower than anticipated, partly due to having to move a lot of CI off of Travis after their billing policy changes.

takagi:
Okay, thanks for the information! CuPy v9 is planned to release on next April, so SciPy 1.6 is likely released before that.

takagi:
Would you mark the functions with `cupy._util.experimental` when the new modes are passed until SciPy 1.6 is released? I suppose in `_check_parameter` and `spline_filter1d` would be fine as each function goes them through.

takagi:
LGTM except for that!

takagi:
Thanks! Let me fix the arguments of the marks a bit to make them properly output messages.

takagi:
pfnCI, test this please.

takagi:
Waiting for CI pass.

PR #4401, opened 2020-12-03
ndimage: add grid_mode option to zoom

grlee77:
blocked by #4400 (This PR builds on that one. Only the final "ENH: implement grid_mode parameter for zoom" commit here is new)

This PR implements zooming in a manner consistent with scikit-image's `resize` and OpenCV's `cv2.resize`. See further explanation in the corresponding SciPy PR scipy/scipy#13095 (to be released in SciPy 1.6).

Basically the issue relates to whether coordinate (0, 0) is at the center or corner of a pixel. In the kernel, this translates into computing coordinates via: 
`c = zoom * (in_coord + 0.5) - 0.5` (for grid_mode=True)
 vs.
`c = zoom * in_coord` (for grid_mode=False)

### DIFF
--- a/cupyx/scipy/ndimage/_filters_core.py
+++ b/cupyx/scipy/ndimage/_filters_core.py
@@ -303,7 +303,7 @@ def _generate_nd_kernel(name, pre, found, post, mode, w_shape, int_type,
                ws_init=ws_init, ws_pre=ws_pre, ws_post=ws_post,
                loops='\n'.join(loops), found=found, end_loops='}'*ndim)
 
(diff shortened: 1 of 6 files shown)


takagi:
pfnCI, test this please.

takagi:
As well as #4400, would you mark `cupy.experimental` if `grid_mode` is supplied by the user. LGTM except for that.

grlee77:
Thanks for reviewing @takagi, I have added the experimental marker here and in #4400

takagi:
pfnCI, test this please.

takagi:
Thanks! I'll mrege this with following #4400 .

PR #4402, opened 2020-12-04
ndimage: implement higher order spline interpolation

### LINKED ISSUE #1523: Higher order interpolation modes in future (v5.0) map_coordinates function
I realize I am jumping the gun here but I noticed that in the future CuPy v5.0 you aim to implement scipy's map_coordinate function. I am very excited for this as my code spends a lot of its time in that function at present. 

I noticed that you also state that spline orders higher than 1 (linear) will not be supported initially but you aim to implement these higher orders at a later stage. Just putting my hand up now to show that at least one person is interested in this functionality. This library has already given 5-6 fold performance boosts on my ffts. So thanks very much!

grlee77:
closes #1523

blocked by #4400 and #4401

### Overview

This PR implements spline interpolation orders 2-5. Spline interpolation was substantially improved in SciPy 1.6, with all boundary modes now being handled correctly. Only the `mode='mirror'` cases can be compared tested against older SciPy.

**To decide**: How do we want to handle the order argument to these functions. Should we change the default back from None to 3 so that it matches SciPy? For now, I have left it as it was in CuPy 8.0.

### Performance

I will post some benchmark results when I get a chance. This is still fast relative to SciPy, but the acceleration is worse than that seen for orders 0 and 1. I would like to see if there is a way to refactor this a bit to reduce register usage, as I suspect that is part of the reason for the performance difference.

### DIFF
--- a/cupyx/scipy/ndimage/_interp_kernels.py
+++ b/cupyx/scipy/ndimage/_interp_kernels.py
@@ -3,6 +3,8 @@ import numpy
 import cupy
 import cupy.core.internal
 
+from cupyx.scipy.ndimage import _spline_prefilter_core
+from cupyx.scipy.ndimage import _spline_kernel_weights
 from cupyx.scipy.ndimage import _util
 
 math_constants_preamble = r'''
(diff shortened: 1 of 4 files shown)


takagi:
Thanks! As for the default value of `order`, yes, it should be changed to align with SciPy. I put `no-compat` label.

takagi:
pfnCI, test this please.

grlee77:
I have changed the default order to 3 as in SciPy and updated the docstrings accordingly.

Please try running the tests again (I fixed a missing argument that was due to a rebasing mistake). All interpolation tests are passing locally.

takagi:
pfnCI, test this please.

takagi:
Thanks! Waiting for CI pass and will be merged soon.

Issue #4406, opened 2020-12-04
Support FFT with more than 2^31 elements


leofang:
This is a parallel issue of #3513 but for FFT functions. cuFFT provides API to do this, so I think it's possible.

Work plan:
- [ ] Wrap low-level cuFFT APIs: We only need `cufftXtMakePlanMany` and `cufftXtExec`; this will also be used as part of #3370, the main driver for me to look into this
- [ ] Native support in high-level APIs like `cupy.fft.fftn`: We should determine which API to use depending on the transform size. For 32 bit indices the current `PlanNd` should cover well, and we only need to switch to the new `cufftXt` APIs when needed.
- [ ] Support callback, multi-GPU transforms, etc: low priority

leofang:
Motivation:
```python
>>> import numpy as np
>>> a_np = np.random.random((2**31,)).astype(np.float32)
>>> a_np += np.random.random((2**31,)).astype(np.float32)
>>> a_np.dtype
dtype('float32')
>>> 
>>> import cupy as cp
>>> a = cp.asarray(a_np)
>>> b = cp.fft.fft(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/leofang/dev/cupy_cuda111/cupy/fft/_fft.py", line 669, in fft
    return _fft(a, (n,), (axis,), norm, cupy.cuda.cufft.CUFFT_FORWARD)
  File "/home/leofang/dev/cupy_cuda111/cupy/fft/_fft.py", line 238, in _fft
    a = _fft_c2c(a, direction, norm, axes, overwrite_x, plan=plan)
  File "/home/leofang/dev/cupy_cuda111/cupy/fft/_fft.py", line 206, in _fft_c2c
    a = _exec_fft(a, direction, 'C2C', norm, axis, overwrite_x, plan=plan)
  File "/home/leofang/dev/cupy_cuda111/cupy/fft/_fft.py", line 154, in _exec_fft
    plan = cufft.Plan1d(out_size, fft_type, batch, devices=devices)
  File "cupy/cuda/cufft.pyx", line 269, in cupy.cuda.cufft.Plan1d.__init__
    def __init__(self, int nx, int fft_type, int batch, *,
OverflowError: value too large to convert to int
```

PR #4407, opened 2020-12-04
Add `cufftXtMakePlanMany` and `cufftXtExec`

leofang:
Part of #4406 and #3370. ~~Work in progress. Will update the PR description later.~~

**UPDATE**: This PR wraps `cufftXtMakePlanMany` and `cufftXtExec` in a new kind of cuFFT plan `XtPlanNd`, so that it can be constructed manually. The purpose of supporting `cufftXtMakePlanMany` is twofold:
1. Support 64-bit indexing (#4406)
2. Support exotic types such as complex32 (#3370) and possibly bf16.

See the discussion in this PR (and also in #4406) for why we don't integrate it with the existing `PlanNd`. In the next PR I'll try to integrate `XtPlanNd` with the high-level APIs to address #4406.

This PR does not work on HIP.

**===End of UPDATE===**

A very preliminary test shows it's possible to do *half-precision* complex FFT (with time reduced by a half)! 
```python
import cupy as cp
import numpy as np
from cupyx.time import repeat


shape = (1024, 256, 256)
dtypes = (cp.complex64, cp.complex128, 'E')  # 'E' = cp.complex32

for t in dtypes:
    if t == 'E':
        dtype = cp.float16
        old_shape = shape
        shape = (shape[0], shape[1], 2*shape[2])  # complex32 has two fp16
    else:
        dtype = t
    idtype = odtype = edtype = np.dtype(t) if t != 'E' else t
    a = cp.random.random(shape).astype(dtype)
    out = cp.empty_like(a)
    if t == 'E':
        shape = old_shape
    plan = cp.cuda.cufft.XtPlanNd(shape[1:],
                                  shape[1:], 1, shape[1]*shape[2], idtype,
                                  shape[1:], 1, shape[1]*shape[2], odtype,
                                  shape[0], edtype,
                                  order='C', last_axis=-1, last_size=None)
    print(repeat(plan.fft, (a, out, cp.cuda.cufft.CUFFT_FORWARD), n_repeat=100))
    plan.fft(a, out, cp.cuda.cufft.CUFFT_FORWARD)
    if t != 'E':
        out_np = np.fft.fftn(cp.asnumpy(a), axes=(-2,-1))
        #print(t, 'ok' if cp.allclose(out, out_np, rtol=1E-3) else 'not ok')
    else:
        a_np = cp.asnumpy(a).astype(np.float32)  # upcast
        a_np = a_np.view(np.complex64)
        out_np = np.fft.fftn(a_np, axes=(-2,-1))
        out_np = np.ascontiguousarray(out_np).astype(np.complex64)  # downcast
        out_np = out_np.view(np.float32)
        out_np = out_np.astype(np.float16)
        ##print(t, 'ok' if cp.allclose(out, out_np, atol=1E-2) else 'not ok')
        # don't worry about accruacy for now, as we probably lost a lot during casting
        print(t, 'ok' if cp.mean(cp.abs(out - cp.asarray(out_np))) < 0.1 else 'not ok')
```
Output (CUDA 10.2 + GTX 2080 Ti):
```
fft                 :    CPU:    5.013 us   +/- 0.214 (min:    4.811 / max:    6.541) us     GPU-0: 4290.127 us   +/-18.198 (min: 4240.768 / max: 4329.792) us
fft                 :    CPU:    5.066 us   +/- 0.243 (min:    4.805 / max:    6.668) us     GPU-0:17339.548 us   +/-27.822 (min:17324.127 / max:17452.127) us
fft                 :    CPU:    4.930 us   +/- 0.197 (min:    4.679 / max:    6.180) us     GPU-0: 2192.620 us   +/- 9.862 (min: 2165.984 / max: 2204.448) us
E ok
```

cc: @carterbox @grlee77


carterbox:
```
GTX 1050 Ti
Compute Capability: 61

<class 'numpy.complex128'>
CPU:        27.964 us   +/-15.416 (min:   12.679 / max:   66.799) us 
GPU-0: 098,351.780 us   +/-18.624 (min:98321.053 / max:98400.253) us

<class 'numpy.complex64'>
CPU:        14.488 us   +/- 7.084 (min:   11.268 / max:   54.220) us
GPU-0: 022,074.731 us   +/-45.726 (min:22031.136 / max:22353.920) us

<class 'numpy.complex32'>
CPU:        29.874 us   +/-12.739 (min:   14.001 / max:   80.358) us
GPU-0: 182,748.333 us   +/-13.548 (min:182727.615 / max:182788.101) us
```

leofang:
Thanks for quick tests, @carterbox! Which CUDA version are you on? I tested it on CUDA 10.2 + GTX 2080 Ti.

carterbox:
```
Quadro P4000
Compute Capability: 61
CUDA Version: 11.0

<class 'numpy.complex128'>
CPU:       11.993 us   +/-  9.071 (min:    8.901 / max:   77.487) us
GPU-0: 44,388.763 us   +/-342.440 (min:44291.073 / max:46947.712) us

<class 'numpy.complex64'>
CPU:       10.256 us   +/-  2.494 (min:    8.689 / max:   28.082) us
GPU-0: 10,651.921 us   +/- 20.315 (min:10634.816 / max:10778.272) us

<class 'numpy.complex32'>
CPU:       11.335 us   +/-  6.400 (min:    9.096 / max:   66.276) us
GPU-0: 82,250.319 us   +/- 70.232 (min:82195.068 / max:82420.227) us
```

carterbox:
It seems that 16-bit floating point was added in Pascal (6.x), but from my tests they don't perform well until Turing (7.5).

The speed-up for our FFT test is only 2x compared with 32-bit floats whereas the speed-up to 32-bit from 64-bit floats is 4x.

carterbox:
I've only read the introduction of [this paper by Ho and Wong (2017)](https://www.comp.nus.edu.sg/~wongwf/papers/hpec17.pdf). It seems that 2x is the maximum expected speed-up and it requires that pairs of 16-bit floats are processed with the same instruction because 16-bit floats are processed with the existing 32-bit compute units. This is very amenable to complex types whose operations are inherently paired.

However, it does not explain why I observed slow-downs on Pascal. Maybe there is a missing compile flag?

leofang:
CUDA 10.2 + V100 (CC 7.0):
```
fft (complex128)   :    CPU:    6.660 us   +/- 0.238 (min:    6.340 / max:    8.180) us     GPU-0: 7260.723 us   +/-41.022 (min: 7254.016 / max: 7668.736) us
fft (complex64)    :    CPU:    6.555 us   +/- 0.236 (min:    6.160 / max:    7.890) us     GPU-0: 3639.777 us   +/- 2.703 (min: 3634.176 / max: 3647.488) us
fft (complex32)    :    CPU:    6.445 us   +/- 0.223 (min:    6.190 / max:    8.060) us     GPU-0: 2087.393 us   +/- 3.978 (min: 2077.696 / max: 2098.176) us
```

leofang:
No, I don't think there's any compile flag missing. Everything used here is just host API, and we simply link the Python module to cuFFT. I guess it's more like Pascal is not well optimized (despite it has the capability for a potential 2x speedup), but I need to check further. Note that on V100 the speedup is less than 2x, but it's probably because the workload is too small compared to the hardware capacity (ex, I got 2x, not 4x, for complex128 -> complex64).

leofang:
@takagi Question: performance issue aside (which I already bugged an NVIDIA friend to look into), is the current design acceptable? As mentioned in #4406, I prefer to switch to use `cufftXtMakePlanMany` only as needed (ex: dealing with oversize arrays or exotic dtypes), as it's unclear to me if it'd incur additional overhead compared to using plans from `cufftMakePlanMany`. Therefore, it's isolated in a new `XtPlanNd` cdef class, with slightly different constructor arguments.

If the design is OK, I'll add a few tests for `XtPlanNd`. We currently don't test `Plan1d` or `PlanNd` directly, but it's because they are fully covered by the high-level tests. Given that it'd take a while to fully resolve #4406, I'd like to have a few simple tests around. 

Thanks!

leofang:
@carterbox: @maxpkatz figured it out the reason for the poor performance you observed: Unlike CC 6.0 (ex: P100), CC 6.1/6.2 do not have accelerated fp16 hardware support. According to the specs I found online for 1080 Ti and Quadro P4000 (oddly I can't find any detailed spec on NVIDIA's website...😞), they have 1:64 ratio to fp32, so your tests actually indicated that the problem size saturated your GPUs for fp16, as the slowdowns were far less than 64.

1080 Ti: https://www.techpowerup.com/gpu-specs/geforce-gtx-1080-ti.c2877
P4000: https://www.techpowerup.com/gpu-specs/quadro-p4000.c2930

carterbox:
Thanks for linking that database; I wouldn't have thought to look there. I'm also confused about why NVidia would release hardware that doesn't do fp16 at least as fast fp32 i.e. (1:1)

leofang:
> This is also a good resource because fp16 support isn't getting better with all newer cards. i.e. the 3090 is only 1:1 whereas the 2080 is 2:1.

This is an interesting question and I don't know 😄 But if the throughput is not improved, at least we'd save some memory...

takagi:
Sorry, I dropped your comment, @leofang. Yes, using `cufftMakePlanMany` only when it is needed also looks reasonable to me.

leofang:
> Yes, using `cufftMakePlanMany` only when it is needed also looks reasonable to me.

Thanks, @takagi! I suppose you meant `cufftXtMakePlanMany` 😁 OK, I'll then proceed to add a few simple tests.

leofang:
@takagi Tests added, PTAL.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

takagi:
LGTM!

PR #4416, opened 2020-12-06
Disentangle HIP from CUDA in the build script

### LINKED ISSUE #4415: Bug: HIP builds implicitly assume CUDA is present
In https://github.com/cupy/cupy/issues/4218#issuecomment-737071882 @ipe-zhangyz reported that `import cupy` still fails on both v8.2 and master despite the CUB issue was already fixed in #4327 (a backport of the support added in #4027).

It turns out that it has to do with the build script. We borrow most of the build infrastructure from CUDA, but one of the assumptions is that CUDA Toolkit is present, which both my develop env and the pfnCI do have (but @ipe-zhangyz's does not), so we failed to detect it until someone who has a "clean" environment came and tested it.

leofang:
Close #4415. Thanks to @ipe-zhangyz for reporting and helping debug the issue 🙏

**UPDATE**: This PR addresses the following things:
- Ensure pfnCI has a pure ROCm environment (by removing all CUDA related stuff)
- Print HIPCC path at build time, if found
- Ensure wherever NVCC is asked, HIPCC can also be used if `use_hip` is set
- Add a version check for HIP, though currently we only detect its version at build time but do not burn it into any module (we should revisit this in a later PR)
- Don't check compute capability on HIP
- Ensure Windows does not have ROCm support
- Remove the internal helper `_format_cuda_version()` as it's just a alias of `str()`

### DIFF
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -496,7 +496,8 @@ def preconfigure_modules(compiler, settings):
                 break
 
     # Get a list of the CC of the devices connected to this node
-    build.check_compute_capabilities(compiler, settings)
+    if not use_hip:
+        build.check_compute_capabilities(compiler, settings)
 
     if len(ret) != len(MODULES):
         if 'cuda' in ret:
@@ -1010,9 +1011,6 @@ class _MSVCCompiler(msvccompiler.MSVCCompiler):
                     include_dirs=None, debug=0, extra_preargs=None,
                     extra_postargs=None, depends=None):
         # Compile CUDA C files, mainly derived from UnixCCompiler._compile().
-        if use_hip:
-            raise RuntimeError('ROCm is not supported on Windows')
-
         macros, objects, extra_postargs, pp_opts, _build = \
             self._setup_compile(output_dir, macros, include_dirs, sources,
                                 depends, extra_postargs)
@@ -1038,6 +1036,9 @@ class _MSVCCompiler(msvccompiler.MSVCCompiler):
 
     def compile(self, sources, **kwargs):
         # Split CUDA C sources and others.
+        if use_hip:
+            raise RuntimeError('ROCm is not supported on Windows')
+
         cu_sources = []
         other_sources = []
         for source in sources:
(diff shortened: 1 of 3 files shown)


leofang:
@ipe-zhangyz I still need to figure out a few things, but I think this should fix it for you. Could you give it a try please?

ipe-zhangyz:
@leofang , thanks for your excellent work! I pulled the _fix_rocm_build_ branch from your repository and built it in my environment. Now all of my tests have passed on ROCm 3.9. Thank you, and waiting your patch to be merged in a release version!

leofang:
Awesome, thanks for confirming @ipe-zhangyz!

leofang:
@emcastillo I think this is a bug fix and so should be backported, thanks!

emcastillo:
Jenkins, test this please

cupy_setup_build.py
```diff
@@ -464,7 +470,8 @@ def preconfigure_modules(compiler, settings):
             # Fail on per-library condition check (version requirements etc.)
             installed = True
             errmsg = ['The library is installed but not supported.']
-        elif module['name'] in ('thrust', 'cub') and nvcc_path is None:
+        elif (module['name'] in ('thrust', 'cub')
+                and (nvcc_path is None and hipcc_path is None)):
             installed = True
             errmsg = ['nvcc command could not be found in PATH.',
```
emcastillo, line 476: let's change the error message too, if people is installing this without cuda, an error about nvcc can be confusing.

install/build.py
```diff
@@ -228,21 +230,21 @@ def _match_output_lines(output_lines, regexs):
...
     """
     global _compiler_base_options
     if _compiler_base_options is None:
-        _compiler_base_options = _get_compiler_base_options()
+        _compiler_base_options = _get_compiler_base_options(use_hip)
     return _compiler_base_options
 
 
-def _get_compiler_base_options():
+def _get_compiler_base_options(use_hip=False):
```
emcastillo, line 243: how about passing the path as `compiler_path` here instead of just the flag,

emcastillo (commented):
There are some nitpicks but in general, I think it's all fine and I have no major concerns. But I would like to avoid `nvcc` mentions in places that can lead to user confusion

leofang:
Good point! I fixed in 2b4814f.

leofang:
Jenkins, test this please

emcastillo (approved):

PR #4369, opened 2020-11-30
add cupyx.scipy.stats.entropy

grlee77:
This PR adds the `scipy.stats` function, `entropy`. This implementation relies on the already existing `cupyx.scipy.special.entr` and `cupyx.scipy.special.rel_entr` functions and so should be simple to review.

There is not currently a `cupyx.scipy.stats` module, so this PR has added it. However, I do not currently have plans to implement other functions within the `stats` module. This function is used within a scikit-image function we created a CUDA equivalent for in `cupyimg`. If there is not an interest in having stats functions within CuPy itself, we can just keep this implementation there.

## Benchmark with pk input only
shape | GPU time (ms) | CPU time (ms) | acceleration
------|---------------|---------------|-------------
(256,) | 0.045 +/- 0.002 | 0.019 +/- 0.001 | 0.420
(512, 512) | 0.117 +/- 0.002 | 5.334 +/- 0.067 | 45.745
(2048, 4096) | 3.266 +/- 0.213 | 181.113 +/- 1.445 | 55.447
(192, 192, 192) | 1.597 +/- 0.153 | 162.830 +/- 1.174 | 101.973

## Benchmark with pk and qk inputs
shape | GPU time (ms) | CPU time (ms) | acceleration
------|---------------|---------------|-------------
(256,) | 0.075 +/- 0.009 | 0.022 +/- 0.003 | 0.295
(512, 512) | 0.162 +/- 0.003 | 2.807 +/- 0.100 | 17.349
(2048, 4096) | 4.317 +/- 0.266 | 100.531 +/- 1.208 | 23.288
(192, 192, 192) | 1.985 +/- 0.002 | 96.379 +/- 0.544 | 48.560

<details><summary>benchmark script</summary>

import numpy
import scipy.stats

import cupy
import cupyx.scipy.stats
from cupyx.time import repeat

print("## Benchmark with pk input only")
print("shape | GPU time (ms) | CPU time (ms) | acceleration")
print("------|---------------|---------------|-------------")
for shape in [(256,), (512, 512), (2048, 4096), (192, 192, 192)]:
    pg = cupy.testing.shaped_random(shape, dtype=numpy.float32)
    perf_gpu = repeat(cupyx.scipy.stats.entropy, (pg,), n_warmup=20, n_repeat=100)
    gpu_times = perf_gpu.gpu_times * 1000

    p = cupy.asnumpy(pg)
    perf_cpu = repeat(scipy.stats.entropy, (p,), n_warmup=1, n_repeat=20)
[...]

PR #4424, opened 2020-12-08
Slight fix on `entropy` test

takagi:
This PR follows #4369 to slightly fix its test.

### DIFF
--- a/tests/cupyx_tests/scipy_tests/stats_tests/test_distributions.py
+++ b/tests/cupyx_tests/scipy_tests/stats_tests/test_distributions.py
@@ -94,7 +94,7 @@ class TestEntropyBasic(unittest.TestCase):
     })
 ))
 @testing.gpu
-@testing.with_requires('scipy')
+@testing.with_requires('scipy>=1.4.0')
 class TestEntropy(unittest.TestCase):
 
     def _entropy(self, xp, scp, dtype, shape, use_qk, base, axis, normalize):
--- a/tests/cupyx_tests/scipy_tests/stats_tests/test_distributions.py
+++ b/tests/cupyx_tests/scipy_tests/stats_tests/test_distributions.py
@@ -124,11 +124,12 @@ class TestEntropy(unittest.TestCase):
         return res
 
     @testing.for_all_dtypes(no_complex=True)
-    @testing.numpy_cupy_allclose(rtol={cupy.float16: 1e-3,
-                                       cupy.float32: 1e-6,
+    @testing.numpy_cupy_allclose(rtol={cupy.float32: 1e-6,
                                        'default': 1e-15},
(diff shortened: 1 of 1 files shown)


tests/cupyx_tests/scipy_tests/stats_tests/test_distributions.py
```diff
@@ -124,11 +124,12 @@ def _entropy(self, xp, scp, dtype, shape, use_qk, base, axis, normalize):
...
 
     @testing.for_all_dtypes(no_complex=True)
-    @testing.numpy_cupy_allclose(rtol={cupy.float16: 1e-3,
-                                       cupy.float32: 1e-6,
+    @testing.numpy_cupy_allclose(rtol={cupy.float32: 1e-6,
                                        'default': 1e-15},
                                  scipy_name='scp')
     def test_entropy(self, xp, scp, dtype):
+        # `entropy` returns float32 value for float16 input so float32
+        # tolerance is used for it.
```
emcastillo, line 132: Are the intermediate computations done in float32?
If that is the case this is fine. But if all the steps are done in float16, even though the output is float32, there will be precision losses. So the original tolerance should be kept.

takagi:
Yes, float16 input is `astype`ed to float32 at the very early point of
`entropy` (at least on CuPy.)

https://github.com/cupy/cupy/blob/3ce7aafc824863b35021862316d2b1b9b1a90736/cupyx/scipy/stats/distributions.py#L46

But the problem is that per-dtype tolerance uses the dtype of the returned
ndarray, so it can not determine that the input was float16 or floar32 when
the function returns float32 value for both float16 and float32 inputs...


2020年12月9日(水) 12:32 emcastillo <notifications@github.com>:

takagi:
Maybe possible to make `numpy_cupy_allclose` to sniff `dtype` when it is
used with `for_dtypes` family at the same time?

takagi:
Shuffle test failed as `axis` parameter is supported since scipy==1.4.0.

tests/cupyx_tests/scipy_tests/stats_tests/test_distributions.py
grlee77: In SciPy, float16 inputs do not seem to be explicitly cast to float32, so the normalization step that divides by a sum over the array will use float16. Thus, the normalization step here should be more accurate than in SciPy. The later use of the `scipy.special` functions does result in things becoming float32, though (I don't think any functions in `scipy.special` support half precision)

emcastillo:
Jenkins, test this please

emcastillo:
Seems like it didn't go well 😁

toslunar:
Could you merge #4481 first?

emcastillo:
Jenkins, test this please

PR #4427, opened 2020-12-10
[DO NOT MERGE] Replace CUDA-related library wrappers with automatically generated ones

takagi:
Follows #4395.

This PR runs CI to check if `cusparse.pyx` automatically generated in #4395 works properly.

### DIFF
--- a/cupy_backends/cuda/libs/cusparse.pxd
+++ b/cupy_backends/cuda/libs/cusparse.pxd
@@ -1,131 +1,197 @@
+# This code was automatically generated. Do not modify it directly.
 
 from libc.stdint cimport intptr_t, int64_t
 
+
+########################################
+# Opaque pointers
+
 cdef extern from *:
-    ctypedef int IndexBase 'cusparseIndexBase_t'
-    ctypedef int Status 'cusparseStatus_t'
-
-    ctypedef void* Handle 'cusparseHandle_t'
-
-    ctypedef void* MatDescr 'cusparseMatDescr_t'
-
-    ctypedef int Direction 'cusparseDirection_t'
-
-    ctypedef int MatrixType 'cusparseMatrixType_t'
-    ctypedef int FillMode 'cusparseFillMode_t'
-    ctypedef int DiagType 'cusparseDiagType_t'
-
-    ctypedef int Operation 'cusparseOperation_t'
-
-    ctypedef int PointerMode 'cusparsePointerMode_t'
-
-    ctypedef int Action 'cusparseAction_t'
-    ctypedef int AlgMode 'cusparseAlgMode_t'
-
-    ctypedef void* cusparseHandle_t
-    ctypedef void* cusparseMatDescr_t
-    ctypedef void* csrsv2Info_t
-    ctypedef void* csrsm2Info_t
-    ctypedef void* csric02Info_t
-    ctypedef void* bsric02Info_t
-    ctypedef void* csrilu02Info_t
-    ctypedef void* bsrilu02Info_t
-    ctypedef void* csrgemm2Info_t
-
-    # Declarations for cuSparse generic API
-    ctypedef int cusparseStatus_t
-    ctypedef int cusparseDirection_t
-    ctypedef int cusparseSolvePolicy_t
-
-    ctypedef int IndexType 'cusparseIndexType_t'
-    ctypedef int Format 'cusparseFormat_t'
-    ctypedef int Order 'cusparseOrder_t'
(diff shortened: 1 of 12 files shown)


takagi:
pfnCI, test this please.

cupy_backends/cuda/libs/cusparse.pyx
```diff
@@ -1356,341 +380,179 @@ cdef dict STATUS = {
...
-cpdef int getVersion(intptr_t handle) except? -1:
-    cdef int version
-    status = cusparseGetVersion(<Handle>handle, &version)
-    check_status(status)
-    return version
-
-
-def get_build_version():
-    return CUSPARSE_VERSION
+    def __init__(self, int batchCount, int64_t batchStride)
```
leofang, line 487: Missed a colon (`:`) in the methods.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

cupy_backends/cuda/libs/cusparse.pyx
```diff
@@ -1357,340 +381,189 @@ cdef dict STATUS = {
...
 cdef class SpVecAttributes:
 
-    def __init__(self, int64_t size, int64_t nnz,
-                 intptr_t idx, intptr_t values,
-                 IndexType idxType, IndexBase idxBase, DataType valueType):
+    def __init__(self, int64_t size, int64_t nnz, intptr_t indices, intptr_t values, IndexType idxType, IndexBase idxBase, DataType valueType):
         self.size = size
         self.nnz = nnz
-        self.idx = idx
+        self.indices = indices
```
takagi, line 434: Attribute names should not be changed.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

leofang:
Wow amazing, you're almost there @takagi!!!

takagi:
I've cleaned up the commits, just including the generated `cusparse.pyx` (and small change on `cusparse.pxd`) for this PR.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

cupy_backends/cuda/libs/cusparse.pxd
```diff
@@ -1,191 +1,264 @@
...
-    CUSPARSE_CSR2CSC_ALG2 = 2  # low memory requirement, non-deterministc
+########################################
+# Auxiliary structures
 
 cdef class SpVecAttributes:
     cdef:
         public int64_t size
         public int64_t nnz
-        public intptr_t idx
+        public intptr_t indices
```
takagi, line 194: Attribute names should not be changed?

takagi:
pfnCI, test this please.

takagi:
The CI configurations with older CUDA fail because of things introduced in newer CUDA. That with CUDA 11.1 (cupy-py36) succeeds.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi:
Keeping cuTENSOR files as they are because it does not have so many APIs and most of them are already implemented. Although I still have a bunch of TODOs, the CI has passed with the generated files for cuBLAS, cuSPARSE, and cuTENSOR now. (I'm still not sure I can go this way about HIP)

leofang:
> (I'm still not sure I can go this way reasonably with HIP)

Sorry I dropped the ball. HIP is a bit complicated AFAIK. At the C level we need to:
- Compare contents between CUDA and the corresponding HIP headers
- Check since which HIP version an API becomes available (optional; I did this manually simply because I have one machine still on ROCm 3.5.0)
- Generate wrappers for the CUDA APIs:
This works most straightforwardly for Runtime APIs by just replacing the prefix `cuda` to `hip`. However, in many places type casting and some extra handling are required. It could be that the enums and constants differ between the two platforms, or due to weird constness mismatching. The worst place is `cusolver.pyx`, because there's no hipSOLVER and I had to wrap rocSOLVER directly. It still works, but the wrapper is not as trivial as in other places (like cuBLAS -> hipBLAS).

Then, at the Cython level we need to raise if an API is not supported by HIP, handle errors and attribute mappings, etc.

takagi:
Superseded by #5876.

PR #4433, opened 2020-12-12
Support texture object in `cupy.ElementwiseKernel`

### LINKED ISSUE #2550: Support texture objects in `ElementwiseKernel`?
This feature is requested by @the-lay in https://github.com/cupy/cupy/pull/2432#issuecomment-543193359. I'll briefly comment below.

leofang:
Close #2550.

This PR allows users to pass a `cupy.cuda.texture.TextureObject` to `cupy.ElementwiseKernel`. The texture coordinates can be computed from the loop indexing variable `i`, see the added tests (thanks to @the-lay for pointing out this possibility which I failed to see!). 

This PR has a particular potential benefit in that since a lot of ndimage kernels are written in `cupy.ElementwiseKernel`, if someone is willing to spend time on adding texture-utlized versions of them, it is possible to use the texture hardware to accelerate some of the computations (ex: [interpolation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#texture-fetching)), which GPUs are known very good since this is one of GPUs' primary applications.

Notes: 
- Texture objects cannot be part of output
- Texture references are not supported since they are deprecated

TODO:
- [x] Where to document this?

### DIFF
--- a/cupy/core/_scalar.pyx
+++ b/cupy/core/_scalar.pyx
@@ -85,6 +85,9 @@ cdef _setup_type_dict():
         _typenames[t] = _typenames_base[d]
         k = ord(d.kind)
         _dtype_kind_size_dict[t] = (k, d.itemsize)
+    # CUDA types
+    for t in ('cudaTextureObject_t',):
+        _typenames[t] = t
 
 
 _setup_type_dict()
--- a/tests/cupy_tests/core_tests/test_userkernel.py
+++ b/tests/cupy_tests/core_tests/test_userkernel.py
@@ -286,7 +286,7 @@ class TestElementwiseKernelTexture(unittest.TestCase):
             'T y',
             '''
             T temp = tex1D<T>(texObj,
-                              T(i)
+                              float(i)
                               );
             y = temp + x;
             ''', name='test_tex1D')
@@ -297,8 +297,8 @@ class TestElementwiseKernelTexture(unittest.TestCase):
             'T y',
             '''
             T temp = tex2D<T>(texObj,
-                              (T)(i % width),
-                              (T)(i / width)
+                              (float)(i % width),
+                              (float)(i / width)
                               );
             y = temp + x;
             ''', name='test_tex2D')
@@ -309,9 +309,9 @@ class TestElementwiseKernelTexture(unittest.TestCase):
             'T y',
             '''
             T temp = tex3D<T>(texObj,
-                              (T)((i % (width * height)) % width),
-                              (T)((i % (width * height)) / width),
-                              (T)((i / (width * height)))
+                              (float)((i % (width * height)) % width),
+                              (float)((i % (width * height)) / width),
+                              (float)((i / (width * height)))
                               );
             y = temp + x;
             ''', name='test_tex3D')
--- a/tests/cupy_tests/core_tests/test_userkernel.py
+++ b/tests/cupy_tests/core_tests/test_userkernel.py
@@ -5,6 +5,10 @@ import pytest
 
 import cupy
 from cupy import testing
+from cupy.cuda import runtime
+from cupy.cuda.texture import (ChannelFormatDescriptor, CUDAarray,
+                               ResourceDescriptor, TextureDescriptor,
+                               TextureObject,)
 
 
 class TestUserkernel(unittest.TestCase):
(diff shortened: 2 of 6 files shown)


leofang:
cc: @grlee77 @coderforlife Perhaps this might be interesting to you? 😛

leofang:
Jenkins, test this please

grlee77:
I think having textures as a possibility is nice.

For `ndimage`, the general kernels we have now can support double precision and complex values as well as various boundary modes not supported by texture objects. For single precision or integer-valued inputs with order = 0 or 1 and when the data has `ndim <= 3` textures are a possibility, but it will probably be more difficult to directly test against SciPy (For example, I think the texture built-in linear interpolation mode has lower numerical accuracy). 

I suppose you could still use our coordinate boundary mapping functions with textures, but it would negate some of the efficiency vs. using the built in texture boundary handling.

tests/cupy_tests/core_tests/test_userkernel.py
```diff
@@ -237,3 +241,102 @@ def test_block_size(self):
...
+
+    def _prep_kernel3D(self):
+        return cupy.ElementwiseKernel(
+            'T x, U texObj, uint64 width, uint64 height',
+            'T y',
+            '''
+            T temp = tex3D<T>(texObj,
+                              (T)((i % (width * height)) % width),
+                              (T)((i % (width * height)) / width),
+                              (T)((i / (width * height)))
```
grlee77, line 314:
```suggestion
                              (float)((i % (width * height)) % width),
                              (float)((i % (width * height)) / width),
                              (float)((i / (width * height)))
```

I think these coordinates all need to be `(float)` rather than `(T)`. In these examples your input happens to be float, so it works, but in general textures with integer types could be used.

(same for 1D and 2D cases above)

tests/cupy_tests/core_tests/test_userkernel.py
```diff
@@ -237,3 +241,102 @@ def test_block_size(self):
...
+            ''', name='test_tex1D')
+
+    def _prep_kernel2D(self):
+        return cupy.ElementwiseKernel(
+            'T x, U texObj, uint64 width',
+            'T y',
+            '''
+            T temp = tex2D<T>(texObj,
+                              (T)(i % width),
+                              (T)(i / width)
```
grlee77, line 301:
```suggestion
                              (float)(i % width),
                              (float)(i / width)
```

the-lay:
Thank you for your work and happy to see you can now use textures in element wise kernel - I had to implement it myself with RawKernel. 

My main usecase was 3D affine transformations with different interpolation types (https://github.com/the-lay/voltools).
I will see if I can rewrite and avoid RawKernel in my voltools. Thanks again!

On a side note, do you think such texture powered affine transformations might be useful to have in cupy? Maybe as a cupyx module. During new year holidays I have some time and could merge some of my existing code maybe.

tests/cupy_tests/core_tests/test_userkernel.py
```diff
@@ -237,3 +241,102 @@ def test_block_size(self):
...
+        # create a texture object
+        return TextureObject(res, tex)
+
+    def _prep_kernel1D(self):
+        return cupy.ElementwiseKernel(
+            'T x, U texObj',
+            'T y',
+            '''
+            T temp = tex1D<T>(texObj,
+                              T(i)
```
grlee77, line 289:
```suggestion
                              float(i)
```

texture index needs to be a float regardless of the type, T.

the-lay:
> (For example, I think the texture built-in linear interpolation mode has lower numerical accuracy). 

Small anecdote, we were worried about that float32 texture precision was not enough and did benchmarks for 3D GPU/CPU interpolations. To our surprise, we found that [cubic bspline interpolation on GPU](http://www.dannyruijters.nl/cubicinterpolation/) performed on-par and even better than CPU interpolations that we've tried.

grlee77:
> My main usecase was 3D affine transformations with different interpolation types (https://github.com/the-lay/voltools).
I will see if I can rewrite and avoid RawKernel in my voltools. Thanks again!

I had not seen this package (but was aware of the Danny Ruijters texture-based cubic interpolation code). It would be interesting to compare with the implementation proposed in #4402. The kernels for order > 2 in that PR are substantially faster in 2D than 3D, I think because the register usage starts becoming excessive in higher dimensions.

grlee77:
**Regarding accuracy:**  SciPy's own spline implementation uses prefilters and is pretty closely based on Thévenaz et. al., but had long had a number of bugs in boundary handling for modes other than 'mirror'. These bugs should be resolved as of the upcoming 1.6 release (1.6.0rc1 was just released). The implementation in #4402 is basically exactly like what is in SciPy 1.6, just adapted for the GPU. It does not make use of texture memory, though. 

I also recently looked up OpenCV's linear interpolation code when a colleague was getting an unexpected result and determined that the interpolation there rounds the coordinate position to something like the nearest 1/32 of a pixel so that a lookup table can be used. That is probably adequate for most applications, but there is no equivalent rounding in the SciPy coordinate positions.

emcastillo:
I kinda like this

tests/cupy_tests/core_tests/test_userkernel.py
leofang: Thanks Gregory! Yes, they need to be `float`. I was being sloppy 😅

leofang:
Thanks for the sparkling discussion @grlee77 @the-lay! This is indeed my main motivation for this PR: to make it possible for interested people to think about potential performance gains that were previously hard/not possible to reach. 

Yes, I suppose given that the hardware-accelerated capability is limited, it can at most cover a subset of what general purpose tools like `scipy.ndimage` or `cupyx.scipy.ndimage` can deliver. I think handling inputs with `float16`, `float32`, or `complex64` (by using two `float32` channels), and integer inputs <= 32 bit width could be possible, but as Gregory pointed out it's harder to directly test against SciPy's outcomes.

> On a side note, do you think such texture powered affine transformations might be useful to have in cupy? Maybe as a cupyx module. During new year holidays I have some time and could merge some of my existing code maybe.

This could be an interesting investigation and I am sure some people would love to know more and/or use it, @the-lay. It would be great if you could open a new issue so that interested people and CuPy core devs can discuss there. 

Frankly the original intention of my texture support PRs was to investigate if reading from texture could be faster than from global memory, which was important for many of our applications (we didn't need the additional hardware capabilities like texture filtering or boundary modes, so in this regard we were kinda not fully utilizing the GPU's power). However, my (very preliminary) studies indicated that it's only as fast as global on high-end GPUs like P100 or V100, so I turned my attention away. I would be more than happy to be proven wrong if anyone here has the opposite experience 🙂

coderforlife:
It seems that the only texture functions available are for fetching with interpolation and gradients and, as stated above,  those don't seem to align with scipy ndimage too well and are apparently a bit slower in many cases than global memory.

the-lay:
I've been taught that the texture memory should be faster for neighbouring reads, but in practice I've never seen a big difference between that and general memory reads from coalesced data. So indeed, the single biggest benefit of texture memory is hardware ("~free") interpolation. 

I've done [benchmarks](https://github.com/the-lay/voltools#benchmark) comparing 3D affine transformation using GPU texture memory with scipy CPU implementation, and the difference is massive as expected. I assume that there still is difference between transformation with cupy-fied scipy and transformation with texture memory. As @grlee77 mentioned, it would be very interesting to measure it and I will do that during the new year holidays when I have some free time.

emcastillo:
> I've been taught that the texture memory should be faster for neighbouring reads, but in practice I've never seen a big difference between that and general memory reads from coalesced data. So indeed, the single biggest benefit of texture memory is hardware (~"free") interpolation.

Yeah, apparently it works well for 2D data since it follows some kind of space-filling curve for the data layout. Maybe @anaruse can shed some light here :)

anaruse:
On older generation GPUs, the use of texture memory actually had a significant effect. However, since modern GPUs have larger L1 cache and shorter latency, the opportunity to improve performance by using texture memory is, honestly speaking, decreasing. Therefore, I personally do not recommend the use of texture memory for the past few years, except when interpolation is necessary.

cupy/core/_scalar.pyx
```diff
@@ -85,6 +85,9 @@ cdef _setup_type_dict():
         _typenames[t] = _typenames_base[d]
         k = ord(d.kind)
         _dtype_kind_size_dict[t] = (k, d.itemsize)
+    # CUDA types
+    for t in ('cudaTextureObject_t',):
+        _typenames[t] = t
```
leofang, line 90: Leaving a note before I forgot what's going on here: the key could perhaps be renamed to a different one from the value (`cudaTextureObject_t`) used in actual code -- say, for example, `CUPY_TexObj` -- if we wanna emphasize it's an internal type used by CuPy.

emcastillo (approved):

emcastillo:
Thanks for the great work!

emcastillo:
Jenkins, test this please

leofang:
Thanks everyone for great discussion and comments! 🙂

Issue #3597, opened 2020-07-13
The role of `cupy_backends.cuda` namespace


okuta:
`cupy_backends.cuda` is a useful namespace. Wouldn't it be a good idea to cut out features that could be used in other libraries? #3584 is an example implementation of this.

How about including the following items in `cupy_backends.cuda`?
- An object with a life cycle. It hides the create-destroy method call. For example, Stream/Event/Memory etc.
- Thin Pythonic wrappers. [Like this.](https://github.com/NVIDIA/pynvrtc/tree/master/pynvrtc 
) For example, compiler/profiler etc.

I'd like to get opinions.

This PR relates #3385 .

leofang:
Hi @okuta just a few initial thoughts. But first I'd like to follow up your https://github.com/cupy/cupy/issues/3385#issuecomment-657358847: 

> However, I gave up moving the whole `cupy.cuda` because `cupy.cuda` has a dependency on `cupy.core.ndarray`. A circular reference exists when `cupy.cuda` is moved to another namespace (e.g., `cupy_XXXX`).

Can't we use `cupy.ndarray` instead of doing `from cupy.core cimport ndarray` to break this circular dependency?

Now, for the `cupy_backends.cuda` namespace I'd say having CUDA low-level/library APIs exposed to Python, as done in #3386, already constituted the first step toward my earlier https://github.com/cupy/cupy/issues/3385#issuecomment-643561809. 

The second step in my imagination, however, is to actually decouple it from CuPy, exactly opposite to what's being considered here. The reason is not all Python CUDA libraries want CuPy as a build-time dependency (they may want to be tightly integrated with CuPy at runtime via `import cupy`, `__cuda_array_interface__`, or other means), but all of them need a CUDA wrapper, so that they can build a customized Python wrapper around it since they might have a different policy for the object lifetimes. In other words, they may not need the high-level wrappers provided by CuPy, so from this viewpoint they should stay in `cupy.cuda`, not `cupy_backends.cuda`. 

For those libraries that include CuPy as one of the dependencies, it doesn't matter in which namespace these wrappers would appear, so it's fine either way. (For my own use cases, I don't have strong opinions, but I might have a second thought later, so when I do I will revisit here 😁)

But, the decoupling that I suggest above will be a giant leap from the status quo, so without strong incentives from external library developers I think we should not do it.

Tagging a few people to solicit comments: @seibert @gmarkall @jakirkham @kkraus14 @mnicely

leofang:
[...]

PR #4445, opened 2020-12-16
Merge `cupy_backends.cuda.stream` into `cupy.cuda.stream`

leofang:
Blocked by #4322 due to potential conflicts. I can resolve it once that one is merged.

Close #3584.

This PR goes in the opposite direction of #3584. I think it is clear that one way or another we need to resolve the code fragmentation when splitting out the code to `cupy_backends`, but this PR makes better sense to me (sorry, @okuta!) since unlike other modules in `cupy_backends` the stream module does not wrap any low-level C APIs; instead, it provides CuPy-specific, high-level functionalities that we normally put in `cupy.cuda`.

The other advantage of this PR over #3584 is this has much less changes in the lines of code.


leofang:
Jenkins, test this please

kmaehashi:
Thank you for raising a PR, and sorry for leaving #3584 without updates.
We discussed this today and thinking that the current implementation is best-balanced between convenience (setting "default stream" automatically instead) and module design.
I understand your point that there is a code fragmentation regarding streams, but as this pull-request introduces a circular reference to `cupy` module which is somewhat the different direction of the initial intension to separate `cupy_backends`.

leofang:
Thanks for quick reply @kmaehashi!

> but this pull-request introduces a circular reference to `cupy` module, which is somewhat the different direction of the initial intension to separate `cupy_backends`.

That was my concern too, but if you take a closer look at this PR, you'll see it's actually not "more" circular, as `cupy.cuda.stream` only imports `cupy_backends.cuda.api.runtime` and nothing else. The import order can be checked by `python -vv -c "import cupy"` or `python -X importtime -c "import cupy"` (Python 3.7+). 

Below is the first few imports involving the modules in `cupy` with this PR:
```
...
import 'cupy._environment' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb6301dc810>
import 'cupy._version' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb63016d8d0>
import 'copy' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb63018d850>
import 'tempfile' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb63018df50>
import 'cupy_backends' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb6301a53d0>
import 'cupy_backends.cuda' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb6301a55d0>
import 'cupy_backends.cuda.api' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb6301a5ad0>
import 'cupy_backends.cuda.api.runtime' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb6301a5dd0>
import 'atexit' # <class '_frozen_importlib.BuiltinImporter'>
import 'cupy._util' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb611c55f90>
import 'cupy.core.syncdetect' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb611c55a10>
import 'cupy_backends.cuda.libs' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb611c55ed0>
import 'cupy.cuda.stream' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb611c64bd0>
import 'cupy_backends.cuda.libs.cublas' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb611c5fcd0>
import 'cupy_backends.cuda.libs.cusolver' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb611c685d0>
import 'cupy_backends.cuda.libs.cusparse' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb611c73e10>
import 'cupy.cuda.device' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb63019f850>
import 'fastrlock' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb60619bf90>
import 'fastrlock.rlock' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb6061a5210>
import 'cupy.cuda.memory_hook' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb6061a57d0>
import 'gc' # <class '_frozen_importlib.BuiltinImporter'>
import 'cupy.cuda.memory' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb6061999d0>
...
```
Note where `cupy.cuda.stream` is imported. In comparison, the current master is just as "circular" as the present PR:
```
...
import 'cupy._environment' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa1102b86d0>
import 'cupy._version' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa110249790>
import 'copy' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa110269750>
import 'tempfile' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa110269f90>
import 'cupy_backends' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa110281290>
import 'cupy_backends.cuda' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa110281490>
import 'cupy_backends.cuda.api' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa110281a10>
import 'cupy_backends.cuda.api.runtime' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa110281d10>
import 'atexit' # <class '_frozen_importlib.BuiltinImporter'>
import 'cupy._util' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0f1d31e90>
import 'cupy.core.syncdetect' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa0f1d31910>
import 'cupy_backends.cuda.libs' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa0f1d31dd0>
import 'cupy_backends.cuda.stream' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0f1d40a90>
import 'cupy_backends.cuda.libs.cublas' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0f1d3bb90>
import 'cupy_backends.cuda.libs.cusolver' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0f1d40b10>
import 'cupy_backends.cuda.libs.cusparse' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0f1d4e410>
import 'cupy.cuda.device' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa11027b710>
import 'fastrlock' # <_frozen_importlib_external.SourceFileLoader object at 0x7fa0e61e9590>
import 'fastrlock.rlock' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0e61e97d0>
import 'cupy.cuda.memory_hook' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0e61e9d50>
import 'cupy.cuda.stream' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa0e61e9f10>
import 'gc' # <class '_frozen_importlib.BuiltinImporter'>
import 'cupy.cuda.memory' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fa11026d050>
...
```
Given the above observation and the fact that this passes the CI, I suppose there's nothing harmful to worry about?

kmaehashi:
Ah, yes, I agree that there should be no problem in terms of the import order of Python modules. 
Our concern is that this change adds a dependency to `cupy` module into `cupy_backends` module, which will make  `cupy_backends` separation which may happen in future (e.g., maintaining it as a separate package) harder.
I'm sorry my wording was ambiguous.

leofang:
Ah...Good point. I missed that, sorry 😅 Let me think about it.

leofang:
@kmaehashi I see the problem now. I think you're referring to the possibilities being discussed in #3597. The problem is for the CUDA library wrappers in `cupy_backends`, there's a (perhaps CuPy-specific) concept of current streams, and the assumption is the pointer to the current stream can be retrieved from the thread-local storage. I think when we split out `cupy_backends` this is not necessarily applicable (or preferred) to other libraries that potentially can adopt it, and some housekeeping changes must be done on our side first. 

My proposal for such housekeeping is to pass the stream pointer explicitly to all library calls. So for example in `cupy_backends.cuda.libs.cublas` instead of this signature
```python
cpdef isamax(intptr_t handle, int n, size_t x, int incx, size_t result)
```
we use this 
```python
cpdef isamax(intptr_t handle, int n, size_t x, int incx, size_t result, intptr_t stream)
```
and use `setStream()` (note: no underscore!) there to set it. This is obviously some work to do, but hopefully with @takagi's #4395 this can be done in a semi-auto fashion, and then we worry about how to make the necessary changes in the Python space. 

This is the only way IMHO that splitting `cupy_backends` could make sense, otherwise I think it's still tied to CuPy too strongly and unlikely to split, and merging this PR would serve CuPy better. What do you think?

leofang:
@kmaehashi WDYT for https://github.com/cupy/cupy/pull/4445#issuecomment-748192120?

leofang:
I am not sure if this is still a good idea with CUDA Python as an alternative backend. Let's close it for now.

PR #4453, opened 2020-12-17
Eliminate read past end of array in percentile_weightnening

### LINKED ISSUE #4451: Read past end of array in percentile kernel can result in incorrect 100th quantile value
* Description:
The [percentile_weightnening kernel](https://github.com/cupy/cupy/blob/master/cupy/_statistics/order.py#L242-L252) in `cupy.percentile` will read one element past the end of the input array if it is asked to calculate the 100th quantile. If a NaN happens to be stored at that device memory location, the reported 100th quantile value will be NaN rather than the largest value in the array. Because that element is multiplied by zero (`weight_above`), this issue _only_ manifests when that value past the end of the array is a NaN.

* Conditions
  - CuPy version: 8.1.0 (installed via conda)
  - OS/Platform: Ubuntu 20.04
  - CUDA version: 10.1
Reviewing the source code, it seems clear that this should not be system-dependent, and I have encountered this bug across a range of OSes and CUDA versions in CI

* Code to reproduce (SEE COMMENT BELOW FOR BETTER REPRODUCER)
This is somewhat difficult to consistently reproduce without modifying cuPy code. One has to simply be unlucky about the value that happens to be at the memory address just past the end of the input array. If we add the following lines to order.py at [line 241](https://github.com/cupy/cupy/blob/master/cupy/_statistics/order.py#L241) just before the invocation of percentile_weightnening, however, we can force the issue:

```python
# Allocate new array with spot for NaN at the end
ap_new = cupy.zeros((ap.shape[0] + 1,), dtype=ap.dtype)
# Put NaN at end of array
ap_new[-1] = cupy.nan
# Set rest of array to the original values
ap_new[:-1] = ap
# Set the input array to percentile_weightnening to the original values (without NaN)
ap = ap_new[:-1]
```
Note that we are still passing the same values into the kernel; we have merely ensured that the memory just past the end of the input array contains a NaN.

With this modification in place, the following will consistently reproduce the issue:
```python
import cupy
quantiles = cupy.array([0, 50, 100], dtype='float32')
arr = cupy.array([0, 1, 2, 3, 4], dtype='float32')
print(cupy.percentile(arr, quantiles))
# Expected output: [0, 2, 4]
# Actual output: [0, 2, nan]
```

Without modifying the cupy code, we have to be unlucky to encounter this bug. I first discovered this issue due to an intermittent failure in a cuML unit test as described [here](https://github.com/rapidsai/cuml/issues/2933#issuecomment-733285578).

Running cuda-memcheck shows that an invalid read is taking place in `percentile_weightnening`.
```
========= Invalid __global__ read of size 4                                                                                                                                                                     
=========     at 0x00000190 in percentile_weightnening                                                                                                                                                          
=========     by thread (5,0,0) in block (0,0,0)                                                                                                                                                                
=========     Address 0x7faa0480abd0 is out of bounds                                                                                                                                                           
=========     Device Frame:percentile_weightnening (percentile_weightnening : 0x190)  
```
We can also confirm that the array index in [this line](https://github.com/cupy/cupy/blob/master/cupy/_statistics/order.py#L250) can reach a value equal to the size of the array by adding a `printf` inside the kernel.

wphicks:
Because the existing percentile_weightnening kernel reads past the end of the input array, the 100th quantile returned by `cupy.percentile` can (incorrectly) be a NaN if that extra read returns a NaN as described in #4451. This PR ensures that the kernel does not read past the end of the array by passing in the size of the array and ensuring that the index is at most `size - 1`.

While the fix is straightforward, the associated test is somewhat trickier. The unit test provided as part of this PR allocates a large array of NaNs, deletes that array, and then calls `cupy.percentile` in order to maximize the likelihood that arrays newly-allocated during the computation will have a NaN in device memory just past their end.

I'm currently leaving this PR as WIP because I would appreciate feedback on the included test. While this test consistently fails for me without the described fix, I don't have a good theoretical justification that this would correctly fail every time. More specifically, the size of the allocated array of NaNs is somewhat arbitrary. 1000 produced consistent failures on my system, but 100 did not.

At a higher level, is this unit test worth having? It demonstrates the very specific issue in the initial `percentile_weightnening` kernel, but `cuda-memcheck` is probably better suited to finding this class of errors than trying to guard against them in tests.

Resolve #4451

### DIFF
--- a/tests/cupy_tests/statistics_tests/test_order.py
+++ b/tests/cupy_tests/statistics_tests/test_order.py
@@ -127,7 +127,7 @@ class TestOrder(unittest.TestCase):
 
         # Check that percentile still returns non-NaN results
         a = testing.shaped_random((5,), cupy, dtype)
-        q = cupy.array((0,100,), dtype=dtype)
+        q = cupy.array((0, 100), dtype=dtype)
         percentiles = cupy.percentile(a, q, axis=None, interpolation='linear')
         assert not cupy.any(cupy.isnan(percentiles))
 
--- a/tests/cupy_tests/statistics_tests/test_order.py
+++ b/tests/cupy_tests/statistics_tests/test_order.py
@@ -6,7 +6,8 @@ import pytest
 
 import cupy
 import cupy.core._accelerator as _acc
-from cupy import cuda, testing
+from cupy import cuda
+from cupy import testing
 
 
 _all_interpolations = (
--- a/tests/cupy_tests/statistics_tests/test_order.py
+++ b/tests/cupy_tests/statistics_tests/test_order.py
@@ -1,4 +1,3 @@
-import gc
 import unittest
 import warnings
 
@@ -123,6 +122,7 @@ class TestOrder(unittest.TestCase):
         # Create an allocator that guarantees array allocated in
         # cupy.percentile call will be followed by a NaN
         original_allocator = cuda.get_allocator()
+
         def controlled_allocator(size):
             memptr = original_allocator(size)
             base_size = memptr.mem.size
@@ -139,7 +139,8 @@ class TestOrder(unittest.TestCase):
 
         cuda.set_allocator(controlled_allocator)
         try:
-            percentiles = cupy.percentile(a, q, axis=None, interpolation='linear')
+            percentiles = cupy.percentile(a, q, axis=None,
+                                          interpolation='linear')
         finally:
             cuda.set_allocator(original_allocator)
 
--- a/tests/cupy_tests/statistics_tests/test_order.py
+++ b/tests/cupy_tests/statistics_tests/test_order.py
@@ -126,7 +126,7 @@ class TestOrder(unittest.TestCase):
         def controlled_allocator(size):
             memptr = original_allocator(size)
             base_size = memptr.mem.size
-            assert base_size == 512
+            assert base_size % 512 == 0
             item_size = dtype().itemsize
             shape = (base_size // item_size,)
             x = cupy.ndarray(memptr=memptr, shape=shape, dtype=dtype)
--- a/tests/cupy_tests/statistics_tests/test_order.py
+++ b/tests/cupy_tests/statistics_tests/test_order.py
@@ -1,3 +1,4 @@
+import gc
 import unittest
 import warnings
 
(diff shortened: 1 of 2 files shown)


takagi:
pfnCI, test this please.

takagi:
Thanks for your quick fix! As you mark this WIP, the unit test implicitly depends on the behavior of CuPy's memory pool facility. I think It is acceptable, but please let me think of a clearer way to do that just a while.

wphicks:
Happy to help! Thank you for giving the unit test some thought. I've also asked the rest of my team to consider other possibilities for that test, and I'm sure they'll post here if they think of a better alternative.

takagi:
The following code should always reproduce the issue. Here, I hook CuPy's default allocator to view the memory block it allocates as float32, filling its contents with `cupy.nan` before it returns the memory pointer to the block. Please not that CuPy's allocator takes a memory block of a multiple of 512 bytes, so the ndarray `ap` [here](https://github.com/cupy/cupy/blob/bcbaf919c49942a146597f281c4a15f2836b08d8/cupy/_statistics/order.py#L200) is always followed by another 4 bytes filled with NaN out of its shape. Would you check if it works?

```Python
import cupy
from cupy import cuda

a = cupy.arange(5, dtype='f')
q = cupy.array([0, 50, 100], dtype='f')

orig = cuda.get_allocator()
def alloc(size):
    memptr = orig(size)
    base_size = memptr.mem.size
    assert base_size == 512
    item_size = cupy.float32().itemsize
    shape = base_size // item_size,
    x = cupy.ndarray(memptr=memptr, shape=shape, dtype='f')
    x.fill(cupy.nan)
    return memptr

cuda.set_allocator(alloc)
try:
    cupy.percentile(a, q)
finally:
    cupy.cuda.set_allocator(orig)
```

wphicks:
Brilliant! That works perfectly. I've just updated the unit test to make use of this. I've modified it very slightly to generalize across the various dtypes we use in the test; let me know if you see any issues with those tweaks.

I'm going to go ahead and pull off the WIP since your updated test seems entirely robust. Thanks very much for the help!

tests/cupy_tests/statistics_tests/test_order.py
```diff
@@ -6,7 +6,7 @@
 
 import cupy
 import cupy.core._accelerator as _acc
-from cupy import testing
+from cupy import cuda, testing
```
takagi, line 9: Would you split this line to import a module per line to follow our guideline that requires it in general?
```suggestion
from cupy import cuda
from cupy import testing
```

tests/cupy_tests/statistics_tests/test_order.py
```diff
@@ -117,6 +117,35 @@ def test_percentile_unexpected_interpolation(self, dtype):
             with pytest.raises(ValueError):
                 xp.percentile(a, q, axis=-1, interpolation='deadbeef')
 
+    @testing.for_float_dtypes()
```
takagi, line 120: Would you add a comment to refer to GitHub to show why this test is provided?
```suggestion
    # See gh-4453
    @testing.for_float_dtypes()
```

takagi (changes requested):
Thanks for checking it and updating the unit test! The change is clearly fine. Would you make some minor changes as commented? After that, I will kick the CI again and merge this change to the master.

wphicks:
Thanks very much! Updated accordingly.

takagi (approved):

takagi:
pfnCI, test this please.

takagi:
LGTM, thanks! The change will be merged into the master after the CI passes.

PR #4454, opened 2020-12-17
[WIP] Add `cupy.complex32`

leofang:
First step toward #3370. Related to #4407. **DECISION NEEDED BEFORE ANY CODE REVIEW**

This is a working prototype to demonstrate that we can define a new Python type (`cupy.complex32`) and the associated NumPy dtype (`np.dtype(cupy.complex32)`) so that this type object can be used to hint that the input array consists of half-precision complex numbers, to be used internally wherever its siblings `cupy.complex64` and `cupy.complex128` are applicable. The subsequent GPU operations is out of scope in this PR and will be addressed when this is merged.

Currently it works for:
- Type identification
- Correct inheritance (ex: `cupy.complex32` inherits from `numpy.complexfloating`, same as other complex types)
- Correct itemsize (4 bytes) for the purpose of array allocation and indexing
- Correct type kind (`'c'` for complex) 
- Meaningful type character (uppercase `'E'` for 2 float16 as complex, as oppose to lowercase `'e'` for 1 float16 as real)
- 4-byte alignment
- Very limited functionalities (see below)

***Critical decision to make before moving on***: I incline to consider it acceptable that it has very limited functionality when used as a dtype. In particular, I do not wish to support all kinds of CPU ufuncs or array funcs (as defined in [this section](https://numpy.org/doc/1.19/reference/c-api/types-and-structures.html#c.PyArray_ArrFuncs)) when manipulated as a NumPy complex32 array, because it would then be more suitable for a PR to NumPy, not to CuPy, and as far as I know NumPy folks do not want to add more new built-in types (numpy/numpy#14753), nor do I have the bandwidth to do so. I think having this type object defined is already enough for GPU needs. If it's acceptable then I will work on polishing this.

### DIFF
--- a/note_for_complex32
+++ b/note_for_complex32
@@ -1,0 +1,15 @@
+Follow this file
+https://github.com/numpy/numpy/blob/master/numpy/core/src/umath/_rational_tests.c.src
+to define a customized Python type complex32 and then dtype
+
+All numpy built-in types are defined in this file:
+https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/arraytypes.c.src
+
+Refs:
+https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/_multiarray_tests.c.src
+https://docs.python.org/3/extending/newtypes_tutorial.html
+https://github.com/rainwoodman/gaepsi/blob/860b662d7a4f99acdfd44c0862cf93b4d5bc777a/gaepsi/compiledbase/npydtype.pxd
+https://github.com/gidden/cyclopts/blob/e346b1721c8d8722af2862823844ab2e7864141b/cyclopts/dtypes.pyx
+https://github.com/sglyon/SGLcapstone/blob/3c15048ca01b53cc302d83fa72345de8a85ef54c/code/xdress/hsfpy/stlcontainers.pyx
+
+
--- a/test_dtype.py
+++ b/test_dtype.py
@@ -1,0 +1,3 @@
+import test_dtype
+test_dtype.init()
+type(test_dtype.complex32)
--- a/test_dtype.py
+++ b/test_dtype.py
@@ -1,5 +1,23 @@
 import test_dtype
+
+
 test_dtype.init()
 print(type(test_dtype.complex32))
 print(test_dtype.complex32)
 print(test_dtype.complex32_dtype)
+
+
+x = test_dtype.complex32_dtype
+assert x.base.__repr__() == "dtype('complex32')"
+assert x.byteorder == '='
+assert x.char == 'E'
+assert x.descr == [('', '<c4')]
+assert x.fields is None
+assert not x.hasobject
+assert x.isnative
+assert x.itemsize == 4
+assert x.kind == 'c'
+assert x.name == str(x) == 'complex32'
+assert x.ndim == 0
+assert x.str == '<c4'
+assert str(x.type).endswith(".complex32'>")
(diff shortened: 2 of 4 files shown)


leofang:
The discussion is ongoing in https://github.com/numpy/numpy/issues/14753!

emcastillo:
We have been talking offline, maybe it's better to postpone this until NumPy implements complex32.

leofang:
So let me quickly summarize the discussion in https://github.com/numpy/numpy/issues/14753 so far. It's doable and the Numpy core devs generally support it. We agree that this dtype will be called `numpy.complex32` with the single character representation `'E'`, as chosen in https://github.com/cupy/cupy/pull/4407. NumPy's new dtype system will be finalized hopefully in a few months, and the questions left are how to add `complex32` to their type hierarchy, and whether or not to make it as "odd" as `float16` (see the discussion therein to know what they meant by "odd"; as far as CuPy is concerned, it's just another dtype).

leofang:
I'd like to note, though, that if CuPy would just alias `numpy.complex32`, it means this feature won't be fully supported in at least 1 year or more until the new NumPy version (the earliest one in which `complex32` is added) becomes our lowest supported NumPy...

emcastillo:
> I'd like to note, though, that if CuPy would just alias numpy.complex32, it means this feature won't be available in CuPy for at least 1 year or more, until the new NumPy version (the earliest one in which complex32 is added) becomes our lowest supported NumPy...

We discussed this, we can implement it once it is released in NumPy master branch, as we always support the latest released NumPy version during development. However, we will need to disable it for older NumPy versions.

leofang:
Let me close this to clean up my PR queue 😁 I'll leave the branch alive for reference. This need is already tracked in #3370.

leofang:
Hmmm I can't reopen because the target branch (`master`) is gone. @kmaehashi or @emcastillo, can you try changing the target branch to `main`? It seems don't have access to do that. If it doesn't work, I'll reopen a PR. Thanks!

leofang:
Sorry for delay. @seberg has kindly taken a look at this PR, and agreed I was on the right track. What I'll do is to clean up/polish the code, hook it up with the build system, and add tests. I'll ping Sebastian and the team once it's ready.

kmaehashi:
Milestone removed.

leofang:
Yeah unfortunately I can't make it for v13.

jakirkham:
It looks like this is in the [`v14` milestone]( https://github.com/cupy/cupy/milestone/146 )

Though most of the CuPy 14 items are under [the `v14.0.0a1` milestone]( https://github.com/cupy/cupy/milestone/147 )

Should we move this under the `v14.0.0a1` milestone? Or do we think this won't make CuPy 14, in which case maybe the milestone should be dropped?

leofang:
I am trying to find someone from NVIDIA to take over this PR and include it in ml_dtypes (https://github.com/jax-ml/ml_dtypes/issues/110#issuecomment-2166815801). I doubt this can be done in v14.0.0.a1, but might be a stretching goal for the v14 a2 or b1 releases, so I think keeping it in the v14 milestone sounds appropriate.

leofang:
> I am trying to find someone from NVIDIA to take over this PR and include it in ml_dtypes ([jax-ml/ml_dtypes#110 (comment)](https://github.com/jax-ml/ml_dtypes/issues/110#issuecomment-2166815801)). 

@seberg will have something ready to upstream soon (https://github.com/jax-ml/ml_dtypes/issues/110#issuecomment-3723167695), but it may not land by v14.0.0. In the meanwhile, the scalar infra is being set up in #9494 and once `complex32` is ready, it should be possible to extend the infra to accept it. This PR is moot, then, with all these ongoing works!

PR #4458, opened 2020-12-18
Fix `__cuda_array_interface__` for HIP

emcastillo:
We were passing 1, or 2 to the cuda stream synchronize and this is not yet supported in HIP

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -2471,6 +2471,11 @@ cpdef ndarray _convert_object_with_cuda_array_interface(a):
     # 1. the stream is not set (ex: from v0 ~ v2) or is None
     # 2. users explicitly overwrite this requirement
     stream_ptr = desc.get('stream')
+    if stream_ptr == 1:
+        stream_ptr = cuda.cupy.cuda.Stream.null.ptr
+    if stream_ptr == 2:
+        # TODO(ecastill) -> fix after merge #4322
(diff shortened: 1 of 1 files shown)


kmaehashi:
I'm wonder if we have to expect ROCm pointers passed via `__cuda_array_stream__` (it can be "This CuPy is built for HIP" error instead?), but anyway I think this fix makes sense.

I locally confirmed sync with the null pointer stream is ok.

```
cupy.cuda.ExternalStream(0).synchronize()  # -> ok
cupy.cuda.ExternalStream(1).synchronize()  # -> segfault
cupy.cuda.ExternalStream(2).synchronize()  # -> segfault
```

pfnCI, test this please.

leofang:
Ah...I feel like I've seen this issue somewhere. Thanks for catching it!

leofang:
@kmaehashi @emcastillo On a second thought, perhaps we should skip all CUDA Array Interface related tests on HIP, as there's no guarantee that it should work at all. There's a stalling discussion on generalizing CUDA Array Interface to HIP (https://github.com/numba/numba/issues/4566), but since Numba does not support ROCm 3.x (https://github.com/numba/numba/issues/6196) I think it'd take quite some time to sort everything out...

kmaehashi:
@leofang Thanks for the pointers! I talked with @emcastillo and thinking it's better to raise an error when `__cuda_array_interface__` is used in CuPy built for HIP and skip tests.

Issue #3833, opened 2020-08-23
ROCm 3.5.0 test: TestCArray32BitBoundary


leofang:
When working on fixing the tests for ROCm 3.5.0 (ex: #3823), I noticed one weird error:
```
$ pytest tests/cupy_tests/core_tests/test_carray.py 
========================================================================= test session starts =========================================================================
platform linux -- Python 3.7.8, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /home/leo/dev/cupy_rocm35, configfile: setup.cfg
collected 11 items                                                                                                                                                    

tests/cupy_tests/core_tests/test_carray.py ....F......                                                                                                          [100%]

============================================================================== FAILURES ===============================================================================
_______________________________________________________ TestCArray32BitBoundary_param_4_{size=4294967296}.test ________________________________________________________

self = <cupy.testing._bundle.TestCArray32BitBoundary_param_4_{size=4294967296} testMethod=test>

    def test(self):
        # Elementwise
        a = cupy.full((1, self.size), 7, dtype=cupy.int8)
        # Reduction
        result = a.sum(axis=0, dtype=cupy.int8)
>       assert result.sum() == self.size * 7
E       AssertionError: assert array(-283467841536) == (4294967296 * 7)
E        +  where array(-283467841536) = <built-in method sum of cupy.core.core.ndarray object at 0x7fdaf0c12e10>()
E        +    where <built-in method sum of cupy.core.core.ndarray object at 0x7fdaf0c12e10> = array([-66, -66, -66, ..., -66, -66, -66], dtype=int8).sum
E        +  and   4294967296 = <cupy.testing._bundle.TestCArray32BitBoundary_param_4_{size=4294967296} testMethod=test>.size

tests/cupy_tests/core_tests/test_carray.py:75: AssertionError
======================================================================= short test summary info =======================================================================
FAILED tests/cupy_tests/core_tests/test_carray.py::TestCArray32BitBoundary_param_4_{size=4294967296}::test - AssertionError: assert array(-283467841536) == (4294967...
==================================================================== 1 failed, 10 passed in 38.88s ====================================================================
```
I don't understand why it would fail when the array size is precisely 2**32 but fine with larger or smaller sizes than that.

Not sure if this is relevant, but this bool test fails if array size > 2**31:

<details><summary>(click here for a long output)</summary>
<p>

```python
>>> import cupy
>>> a = cupy.full((1, 2**32), 7, dtype=cupy.int8)
[...]

PR #4461, opened 2020-12-18
Skip large size test for hip

emcastillo:
There is a bug in HIP/Rocm in which grids with total size of 2**32 do not launch threads at all.

This can be reproduced with 

```python

import cupy as cp


size = 2**32

add_kernel = cp.RawKernel(r'''
extern "C" __global__
void my_add(char* y) {
    unsigned long long block_dim = blockDim.x;
    unsigned long long block_idx = blockIdx.x;
    unsigned long long tid = block_dim * block_idx + threadIdx.x;
    y[tid] = 24;
}
''', 'my_add')

y = cp.empty(size, dtype=cp.int8)
add_kernel((size//256,), (256,), (y,))  # grid, block and arguments
cp.cuda.stream.get_current_stream().synchronize()
print(y)
```

I also created a C reproducer to report this to AMD folks.

```c
#include <hip/hip_runtime.h>
#include <stdio.h>
#include <stdlib.h>
#include <iostream>

using namespace std;

__global__ void helloworld(char* out)
{
	uint64_t bdim = hipBlockDim_x;
	uint64_t bidx = hipBlockIdx_x;
	uint64_t bsize = bdim * bidx;
        uint64_t num = hipThreadIdx_x + bsize;
	out[num] = 23;
}

int main(int argc, char* argv[])
{

    //uint64_t size = 4294967296 - 1024; // 2^32 - WORKS
    uint64_t size = 4294967296; // 2^32 - FAILS, No elements sets at all between 2^32 and 2^32 + 1023
    // uint64_t size = 4294967296 + 1024; // 2^32 +1024 - FAILS, Only sets the first 1024 elements.
    char* output = new char[size];
    char* outputBuffer;
    hipMalloc((void**)&outputBuffer, (size) * sizeof(char));
    hipLaunchKernelGGL(helloworld,
    size/256,
    256,
    0, 0,
    outputBuffer);
    hipMemcpy(output, outputBuffer, (size) * sizeof(char), hipMemcpyDeviceToHost);
    hipMemcpy(output, outputBuffer, (size) * sizeof(char), hipMemcpyDeviceToHost);
    cout << (int)output[0]<<"  "<<(int)output[size-1] << endl;
    hipFree(outputBuffer);
    delete[] output;
    return 0;
}

### DIFF
--- a/tests/cupy_tests/core_tests/test_carray.py
+++ b/tests/cupy_tests/core_tests/test_carray.py
@@ -67,6 +67,8 @@ class TestCArray32BitBoundary(unittest.TestCase):
         # Free huge memory for slow test
         cupy.get_default_memory_pool().free_all_blocks()
 
+    # HIP is known to fail with sizes > 2**32-1024
+    @unittest.skipIf(cupy.cuda.runtime.is_hip, 'HIP does not support this')
     def test(self):
         # Elementwise
         a = cupy.full((1, self.size), 7, dtype=cupy.int8)


leofang:
WTF....So does this also affect the bug in #3833? Should we also add a check in the `_launch()` function in `cupy/cuda/function.pyx ` to prevent an oversize grid from being launched throughout the codebase?

kmaehashi:
pfnCI, test this please.

kmaehashi:
LGTM!

kmaehashi:
We'll check if this issue is resolved in ROCm 4.0.

kmaehashi:
This issue is still reproducible with ROCm 4.0.

amathews-amd:
Hi @emcastillo, @leofang 
I saw the conversation at https://github.com/ROCm-Developer-Tools/HIP/issues/2253
The check in the hip code is at: 
https://github.com/ROCm-Developer-Tools/HIP/blob/main/rocclr/hip_module.cpp#L359 
std::numeric_limits<uint32_t>::max() = 4294967295

If you run your code with AMD_LOG_LEVEL=3, you should see the failure.

`int main(int argc, char* argv[])
{
        int device_count = 0;
        hipGetDeviceCount(&device_count);

        std::cout<< "Found " << device_count << " HIP devices." << std::endl;
        for( int i = 0; i < device_count; i++ )
        {
                std::cout << "Device " << i << std::endl;
                hipDeviceProp_t props;
                hipGetDeviceProperties(&props, i);
                std::cout << " Name: " << props.name << std::endl;
                std::cout << " Max Grid Size X: " << props.maxGridSize[0] << std::endl;
                std::cout << " Max Grid Size Y: " << props.maxGridSize[1] << std::endl;
                std::cout << " Max Grid Size Z: " << props.maxGridSize[2] << std::endl;
        }
       return 0; 
}
`

should return 2**31-1. 

I believe this is same from NVIDIA ? https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities 

In any case, I think you should be able to check this within CuPy before you launch the kernel, and may have special handling if such sizes are applicable to the CuPy platform ?

leofang:
Hello @amathews-amd, Please kindly check again the HIP issue that you linked to (and in particular my comment https://github.com/ROCm-Developer-Tools/HIP/issues/2253#issuecomment-811957200 there). It's not about max grid size (which is 2^31-1 on both NVIDIA and AMD GPU without doubt), but about the array size; the description in https://github.com/cupy/cupy/pull/4461#issue-542348493 is a bit misleading, but the reproducer is correct: If you check the C reproducer carefully, you'll see that the grid size is far less than 2^31-1, but because of the reason discussed in the HIP issue, it'd still fail on AMD GPU (while working correctly on NVIDIA GPU), which led to this PR to disable the failing test only on HIP.

amathews-amd:
# /opt/rocm/bin/rocminfo  | grep "Grid Max"
  Grid Max Size:           4294967295(0xffffffff)
2 ** 32 = 4294967296
So, max supported is (2**32 -1 ).

If you code the reproducer from @emcastillo to produce 4294967295 threads, it works (atleast on my machine). 
`
    uint64_t blocks_per_grid = 16843009;
    uint64_t threads_per_block = 255;
    uint64_t size = blocks_per_grid  * threads_per_block; // 4294967295
`

leofang:
Hi @amathews-amd What do you see if you set threads_per_block = 256 as in the reproducer?

Let me ask this way. In the reproducer we launched the kernel with 1D grid size = 2^32/256 = 16777216 and 1D block size = 256. Therefore, the grid size is smaller than MaxGridDimX = 2^31-1 = 2147483647, and the block size is smaller than MaxBlockDimX = 1024. Why would this configuration fail? This code runs perfectly fine in CUDA (output is `23  23`), and according to the specs returned by `hipDeviceGetAttribute()` it should also be valid in HIP.

Another thing is the kernel launch should not return success if the kernel config is considered invalid. We should not do the check ourselves; instead the kernel launcher API should do it for us and we simply check the return value. Here the kernel launch fails silently without any error, but if the config is invalid it shouldn't even launch. (This is a minor issue compared to the above one, though.)

amathews-amd:
> In the reproducer we launched the kernel with 1D grid size = 2^32/256 = 16777216 and 1D block size = 256. Therefore, the grid size is smaller than MaxGridDimX = 2^31-1 = 2147483647, and the block size is smaller than MaxBlockDimX = 1024. Why would this configuration fail? This code runs perfectly fine in CUDA (output is `23 23`), and according to the specs returned by `hipDeviceGetAttribute()` it should also be valid in HIP.
Because of this: https://github.com/ROCm-Developer-Tools/HIP/blob/main/rocclr/hip_module.cpp#L359 
and this:
 $ /opt/rocm/bin/rocminfo | grep "Grid Max"
Grid Max Size: 4294967295(0xffffffff)

> We should not do the check ourselves; instead the kernel launcher API should do it for us and we simply check the return value.

Regarding kernel launches not returning success, its the same with CUDA (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#error-checking). Have you tried  hipGetLastError ? https://rocmdocs.amd.com/en/latest/Programming_Guides/CUDAAPIHIPTEXTURE.html

leofang:
@amathews-amd Let's move the discussion to https://github.com/ROCm-Developer-Tools/HIP/issues/2253. See my reply  https://github.com/ROCm-Developer-Tools/HIP/issues/2253#issuecomment-887987272 there.

PR #4463, opened 2020-12-18
Update to use the new CArray and CIndexer constructors

coderforlife:
This updates the ndimage scipy generic filters to use the new `CArray` and `CIndexer` constructors instead of manipulating their memory which was quite hackish.


emcastillo:
Jenkins, test this please

emcastillo:
Thanks a lot! I couldn't wrap my head to do this when I wrote that code 😂

Issue #4450, opened 2020-12-16
implement cupyx.scipy.special.lambertw


brandondube:
This function is missing in the scipy compatible routines.  It would be nice to see it added.

[Scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.lambertw.html)

[cupyx special page](https://docs.cupy.dev/en/stable/reference/special.html)

izaid:
Indeed, part of the inspiration for doing this is to create a minimal, header-only C++ library for special functions. Right now, it is part of SciPy, but that may change in the future. These are functions that are meant to just be fed into `cp.ElementwiseKernel` or something similar. It should already be possible to provide `lambertw` this way, and we're certainly keen on feedback.

PR #4468, opened 2020-12-21
Implementing cupyx.scipy.signal.oaconvolve

coderforlife:
This implements `cupyx.scipy.signal.oaconvolve` which uses the recently merged `cupyx.scipy.signal.fftconvolve` to do most of the work by breaking the larger data into smaller blocks to make the FFTs faster. Nearly all of the other logic comes from scipy with some simplifications.

A custom version of scipy's `special.lambertw` function is provided that only works for the situation needed to determine the block size (called `_optimal_oa_block_size`). This function is only called once per call to `oaconvolve` and takes a 1-2 microseconds in my tests, if necessary this can be moved to Cython which is 10x faster.

### DIFF
--- a/cupyx/scipy/signal/signaltools.py
+++ b/cupyx/scipy/signal/signaltools.py
@@ -220,7 +220,7 @@ def oaconvolve(in1, in2, mode="full", axes=None):
         in2 (cupy.ndarray): Second input. Should have the same number of
             dimensions as ``in1``.
         mode (str): Indicates the size of the output:
-            
+
             - ``'full'``: output is the full discrete linear \
                           cross-correlation (default)
             - ``'valid'``: output consists only of those elements that do \
--- a/cupyx/scipy/signal/signaltools.py
+++ b/cupyx/scipy/signal/signaltools.py
@@ -258,7 +258,6 @@ def oaconvolve(in1, in2, mode="full", axes=None):
 
     # Fall back to fftconvolve if there is only one block in every dimension
     if in1_step == s1 and in2_step == s2:
-        print(axes)
         return fftconvolve(in1, in2, mode=mode, axes=axes)
 
     # Pad and reshape the inputs for overlapping and adding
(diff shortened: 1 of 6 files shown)


coderforlife:
Additionally, if #4450 is addressed, `cupyx.scipy.special.lambertw` could be used instead, but it will unlikely be faster than this custom version which has parts optimized for the hardcoded values of `z` and `k`.

takagi:
pfnCI, test this please.

cupyx/scipy/signal/signaltools.py
```diff
@@ -201,6 +202,106 @@ def choose_conv_method(in1, in2, mode='full'):
...
+    # Calculate the block sizes for the output, steps, first and second inputs.
+    # It is simpler to calculate them all together than doing them in separate
+    # loops due to all the special cases that need to be handled.
+    optimal_sizes = (_st_core._calc_oa_lens(s1[i], s2[i]) if i in axes else
+                     (-1, -1, s1[i], s2[i]) for i in range(in1.ndim))
+    block_size, overlaps, in1_step, in2_step = zip(*optimal_sizes)
+
+    # Fall back to fftconvolve if there is only one block in every dimension
+    if in1_step == s1 and in2_step == s2:
+        print(axes)
```
takagi, line 261: To  be removed?

takagi:
Would you add `oaconvolve` to `docs/source/reference/signal.rst` so we can generate its reference page?

takagi:
LGTM except for the points!

cupyx/scipy/signal/signaltools.py
coderforlife: Sorry! Debug print still around... Thanks for the catch.

coderforlife:
Should now be ready to go! Thanks for the input!

leofang:
Jenkins, test this please

takagi:
LGTM, thanks!

PR #4472, opened 2020-12-21
Fix raw kernel test

kmaehashi:
This is to fix Windows test failure

### DIFF
--- a/tests/cupy_tests/core_tests/test_raw.py
+++ b/tests/cupy_tests/core_tests/test_raw.py
@@ -2,7 +2,6 @@ import contextlib
 import io
 import os
 import pickle
-import string
 import subprocess
 import sys
 import tempfile
@@ -1121,15 +1120,21 @@ class TestRawGridSync(unittest.TestCase):
 
 _test_script = r'''
 import pickle
+import sys
+
 import cupy as cp
+
 
 N = 100
 a = cp.random.random(N, dtype=cp.float32)
 b = cp.random.random(N, dtype=cp.float32)
 c = cp.empty_like(a)
-with open('${temp_dir}' + '/TestRawPicklable', 'rb') as f:
+with open('raw.pkl', 'rb') as f:
     ker = pickle.load(f)
-${fetch_ker}
+
+if len(sys.argv) == 2:
+    ker = ker.get_function(sys.argv[1])
+
 ker((1,), (100,), (a, b, c, N))
 assert cp.allclose(a + b, c)
 assert ker.enable_cooperative_groups
@@ -1143,6 +1148,9 @@ assert ker.enable_cooperative_groups
     'compile': (False, True),
     'raw': ('ker', 'mod', 'mod_ker'),
 }))
+@unittest.skipUnless(
+    60 <= int(cupy.cuda.device.get_compute_capability()),
+    'Requires compute capability 6.0 or later')
 @unittest.skipIf(cupy.cuda.runtime.is_hip,
                  'HIP does not support enable_cooperative_groups')
 class TestRawPicklable(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, test this please.

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -1143,6 +1148,9 @@ def test_grid_sync_rawmodule(self):
     'compile': (False, True),
     'raw': ('ker', 'mod', 'mod_ker'),
 }))
+@unittest.skipUnless(
+    60 <= int(cupy.cuda.device.get_compute_capability()),
+    'Requires compute capability 6.0 or later')
```
leofang, line 1153: Question: what does CC have to do with pickling?

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -1190,23 +1198,17 @@ def test_raw_picklable(self):
         elif self.raw == 'mod_ker':
             # pickle the RawKernel fetched from the RawModule
             obj = self.mod.get_function('test_sum')
-        with open(self.temp_dir + '/TestRawPicklable', 'wb') as f:
+        with open(self.temp_dir + '/raw.pkl', 'wb') as f:
```
leofang, line 1201: I thought one of the issues was that we need to handle the paths in a more robust way, using `os.path.join`?

leofang: Ditto for below.

tests/cupy_tests/core_tests/test_raw.py
kmaehashi: This is because `enable_cooperative_groups=True` is used in tests. Maybe we can instead use `translate_cucomplex` attribute so that we don't have to care about CC.

tests/cupy_tests/core_tests/test_raw.py
kmaehashi: I agree that using `os.path.join` is the cleaner way but that was not a problem this time.
The reason why it's failing only in Windows is that `string` template is used to embed a file path. Windows file paths contain backslashes, so when embedding a path (or any arbitrary strings) into a string literal in Python code template, we need to escape `\` as `\\`. Obviously, we don't want to do this 😅

kmaehashi: Rename (`TestRawPicklable` -> `raw.pkl`) itself has no meaning (sorry for confusion), I just wanted to make sure that it's a pickle file.

tests/cupy_tests/core_tests/test_raw.py
leofang: Ah, that's a good point! We can mark `translate_cucomplex` as `readonly` in `raw.pxd`, then we can replace `enable_cooperative_groups` by it. (The attribute being exposed to Python was the main reason I didn't use it in the first place 😅)

tests/cupy_tests/core_tests/test_raw.py
leofang: I see, thanks for explanation! 👍

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -1190,23 +1198,17 @@ def test_raw_picklable(self):
...
         with open(self.temp_dir + '/TestRawPicklable.py', 'w') as f:
-            f.write(test_ker)
+            f.write(_test_script)
+        test_args = ['test_sum'] if self.raw == 'mod' else []
 
         # run another process to check the pickle
-        s = subprocess.run([sys.executable,
-                            self.temp_dir + '/TestRawPicklable.py'])
+        s = subprocess.run([sys.executable, 'TestRawPicklable.py'] + test_args,
+                           cwd=self.temp_dir)
```
toslunar, line 1211: Good!  Setting `cwd` makes the test more isolated, in general.

toslunar (approved):
Thanks @leofang for good review comments!

leofang:
@kmaehashi This fix should be backported to v8 too.

PR #4477, opened 2020-12-21
Exclude unsupported dtypes for TestOrderFilter

kmaehashi:
This fixes the test variation to exclude unsupported dtypes. Errors were hidden by `accept_error=ValueError` option, but should be explicitly excluded.

I found this because the following code cause hard crash on Windows + SciPy 1.5.4.

```
import numpy, scipy.signal
scipy.signal.order_filter(numpy.ones(10, dtype=numpy.float16), numpy.ones(3, dtype=numpy.bool_), 0)
```

On Linux, this raises an exception.

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
@@ -203,7 +203,7 @@ class TestWiener(unittest.TestCase):
 @testing.gpu
 @testing.with_requires('scipy')
 class TestOrderFilter(unittest.TestCase):
-    @testing.for_all_dtypes()
+    @testing.for_all_dtypes(no_float16=True, no_bool=True, no_complex=True)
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, test this please.

toslunar:
> hard crash

It seems only `float16` crashes.
https://github.com/scipy/scipy/blob/v1.5.4/scipy/signal/sigtoolsmodule.c#L833-L841
https://github.com/scipy/scipy/blob/v1.5.4/scipy/signal/sigtoolsmodule.c#L897
https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_TYPE
https://github.com/numpy/numpy/blob/v1.19.4/numpy/core/include/numpy/ndarraytypes.h#L65-L89

toslunar (approved):
LGTM because this PR corrects the comment `# for even kernels` on `accept_error=ValueError`.

Issue #4440, opened 2020-12-15
cuslover causes memory overrun while calling syevjBatched


mor2code:
cuslover.py causes memory overrun while calling syevjBatched,

according to NVidia documentation https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-syevjbatch

"an integer array of dimension batchSize. If CUSOLVER_STATUS_INVALID_VALUE is returned, info[0] = -i (less than zero) indicates i-th parameter is wrong (not counting handle). Otherwise, if info[i] = 0, the operation is successful. if info[i] = n+1, syevjBatched dose not converge on i-th matrix under given tolerance and maximum sweeps."

dev_info must be defined otherwise cuda overruns memory not allocated to it (in my case it zeroed a cupy array not related to the call, it actually wrote the results of dev_info on unrelated neighbouring in memory cupy array).

To resolve line 432 of cusolver.py need to be changed from:

dev_info = _cupy.empty((), _numpy.int32)

to:

dev_info = _cupy.empty((int(batch_size)), _cupy.int32)

attached patch

cusolver.py.patch.txt
[cusolver.py.patch.txt](https://github.com/cupy/cupy/files/5693616/cusolver.py.patch.txt)

takagi:
Thanks for reporting! It looks as the code of the non-batched version was copied there. If possible, would you post a pull request with the attached patch?

asi1024:
This issue is resolved by #4491.

PR #4491, opened 2020-12-23
fix dev info allocation

mor2code:
fix dev info allocation

### DIFF
--- a/cupy/cusolver.py
+++ b/cupy/cusolver.py
@@ -344,7 +344,7 @@ def syevj(a, UPLO='L', with_eigen_vector=True):
 
     m, lda = a.shape
     w = _cupy.empty(m, inp_w_dtype)
-    dev_info = _cupy.empty((), _numpy.int32)
+    dev_info = _cupy.empty((1,), _cupy.int32)
     handle = _device.Device().cusolver_handle
 
     if with_eigen_vector:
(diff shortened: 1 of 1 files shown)


cupy/cusolver.py
```diff
@@ -344,7 +344,7 @@ def syevj(a, UPLO='L', with_eigen_vector=True):
 
     m, lda = a.shape
     w = _cupy.empty(m, inp_w_dtype)
-    dev_info = _cupy.empty((), _numpy.int32)
+    dev_info = _cupy.empty((1,), _cupy.int32)
```
leofang, line 347: This change is not strictly necessary as it does allocate 4 bytes correctly.

leofang:
btw you can modify your PR description and add "this closes #4440".

leofang:
Jenkins, test this please

asi1024:
LGTM! Thanks!

mor2code:
as far as i understand the dev_info = _cupy.empty((), _numpy.int32),
an array with memory of size zero bytes is created rather than the required
size 1 int32 of 4 bytes dev_info.data.ptr  will be incorrect without
modification

While using the batched version of syevj it exposed the problem but it same
for the non batch version which needs sizeof 1 in32

leofang:
Hi @mor2code, I see your confusion. Note the shape `()` is not equivalent to `(0,)`: the former refers to a 0-d array (which is a scalar, backed by 4 bytes of the device memory) while the latter refers to a zero-size array (so no device memory is allocated):
```python
>>> a=cupy.empty((), cupy.int32)
>>> a
array(0, dtype=int32)
>>> a.size  # note this is not 0
1
>>> a[()]=1  # change the content
>>> a
array(1, dtype=int32)
```

PR #4494, opened 2020-12-24
[WIP] Fix cuFFT sdist issue

kmaehashi:
This fixes the sdist build error when Cython is installed in user environment.
(CuPy sdist installation does not contain pxd, so it will raise an build error when installed from sdist)

https://github.com/cupy/cupy/pull/4141#discussion_r512390571

Please review commit by commit.
* rename a file 492098dedd5bb83b06d2c362e55735617522cda5 d3a722b0776025a30593cfcbc7f486627953e77a
* avoid using cimport from other cupy modules, as pxd files are not included in sdist dbe6295f9d823a2cb1fa9c5eb8845316a8d62b4b 30dd958dbd84cfa36f8a0d7076cd9555aa3e9aab
* expose `get_current_stream_ptr` to Python to allow using it from cufft.pxi  629a9adc4962c3aac051d3b96f5df1c5a74cfc5f


kmaehashi:
pfnCI, test this please.

emcastillo (approved):

toslunar (changes requested):
> rename only: [b1ad9c9](https://github.com/cupy/cupy/commit/b1ad9c9ddc873c30f22adb78175961833d7b2213).

To let git track the "move", please do not readd `cupy/cuda/cufft.pyx` (`include "cufft.pxi"`) in a single commit.

leofang:
Sorry, I don't understand the change in this PR. 

> This fixes the sdist build error when Cython is installed in user environment.
> (CuPy sdist installation does not contain pxd, so it will raise an build error when installed from sdist)

When I run `python setup.py sdist` I see the following files are included:
```
...
cupy-9.0.0a3/cupy/cuda/cufft.cpp
cupy-9.0.0a3/cupy/cuda/cufft.pxd
cupy-9.0.0a3/cupy/cuda/cufft.pyx
...
```
and if I unpack the sdist in a fresh Python environment with Cython installed and run `pip install`, it builds fine. What am I missing here?

kmaehashi:
> and if I unpack the sdist in a fresh Python environment with Cython installed and run pip install, it builds fine. What am I missing here?

The current `cufft.pyx` cimports from `cupy_backends`, but as sdist does not contain pxd files other than `cufft.pxd`, the build should fail. I guess you have `driver.pxd` / `runtime.pxd` somewhere in your python environment?

kmaehashi:
pfnCI, test this please.

emcastillo:
Jenkins, test this please

leofang:
> > and if I unpack the sdist in a fresh Python environment with Cython installed and run pip install, it builds fine. What am I missing here?
> 
> The current `cufft.pyx` cimports from `cupy_backends`, but as sdist does not contain pxd files other than `cufft.pxd`, the build should fail. I guess you have `driver.pxd` / `runtime.pxd` somewhere in your python environment?

No I actually tried building in a fresh Python env. I think in this case the build system just picks up the .cpp file and compiles it, bypassing cythonize (same as other models). Could you confirm? The steps I took are:
- git clone a fresh copy of cupy
- `python setup.py sdist`
- uncompress the generated `.tar.gz` file somewhere
- create a new, fresh Python env (with Cython installed)
- go to the sdist source directory, and run either `pip install -v .` or `pip install -v -e .`

kmaehashi:
Thank you for details, I'm now checking again in my local environment.
FYI this was detected in https://jenkins.preferred.jp/job/chainer/job/cupy_master_cupy-install/412/ID=4,label=mn1-p100/console

leofang:
Thanks, @kmaehashi! I noticed these two lines in the error log:
```
18:18:20     Compiling cupy/cuda/cufft.pyx because it changed.
18:18:20     [1/1] Cythonizing cupy/cuda/cufft.pyx
```
Do you know how the change is detected? I suppose not by timestamp, but by, say, different Cython (or Python) versions between the env used to generate the sdist and the user env? That might explain...

kmaehashi:
I'm unsure how Cython detects the changes... anyway it's better to avoid depending on unclear Cython behavior.

I tested locally:
(1) Create sdist -> extract  -> build: no error
(2) Create sdist -> extract -> `touch cupy/cuda/cufft.pyx` -> build: error (`Compiling cupy/cuda/cufft.pyx because it changed.`)
(3) Create sdist -> extract -> downgrade Cython from 0.29.21 to 0.29.20 -> build: error  (`Compiling cupy/cuda/cufft.pyx because it depends on /home/maehashi/.pyenv/versions/local-3.8.1/lib/python3.8/site-packages/Cython/Includes/libcpp/__init__.pxd.`)

toslunar (approved):

leofang:
> I'm unsure how Cython detects the changes... anyway it's better to avoid depending on unclear Cython behavior.

I agree. It's just a bit nerve wrecking to me that we have to sacrifice cimport...🙁 Does it make sense to, for example, when building from sdist, move cufft.pxd/.pyx before building to force using cufft.cpp, and move them back?

> (1) Create sdist -> extract -> build: no error

Thanks for confirming 🙏 So I didn't do enough tests back then... 😥

kmaehashi:
> I agree. It's just a bit nerve wrecking to me that we have to sacrifice cimport...🙁 Does it make sense, for example, when building from sdist, move cufft.pxd/.pyx away before building to force using cufft.cpp, and move them back?

I think that's also feasible (although I would like to proceed this PR for now, as the release is planned for tomorrow).
Create a flag file when `sdist_with_cython` and included it in sdist; if the flag file exists do not use pyx at all.

kmaehashi:
pfnCI, test this please.

kmaehashi:
It seems the problem cannot be fixed soon. We'll be proceed with the release without this PR (considering the original PR is already in the v9a2 release and no one complaining it), and think about how we can avoid this rename.

kmaehashi:
Closing as #4619 should fix this issue.

PR #4485, opened 2020-12-23
Fixes a bunch of AMD issues

emcastillo:
Fixes Raw kernel tests and some issues with `__cuda_array_interface__`

Issue #4484, opened 2020-12-23
HIP test errors


emcastillo:
Currently core tests

- [ ] core_tests/test_ndarray.py::TestNdarrayCopy::test_copy_multi_device_non_contiguous - AssertionError:

- Happens only if run with other tests, memory corruption suspected

- [ ] core_tests/test_ndarray_elementwise_op.py::TestArrayElementwiseOp::test_broadcasted_pow - AssertionError:
- Requires adjusting tolerance
```
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       Mismatched elements: 2 / 6 (33.3%)
E       Max absolute difference: 7.6293945e-06
E       Max relative difference: 2.1192763e-07
E        x: array([[ 1.      +0.j,  2.      +0.j,  3.      +0.j],
E              [16.      +0.j, 24.999996+0.j, 35.999992+0.j]], dtype=complex64)
E        y: array([[ 1.+0.j,  2.+0.j,  3.+0.j],
E              [16.+0.j, 25.+0.j, 36.+0.j]], dtype=complex64)
```
- [ ] core_tests/test_ndarray_elementwise_op.py::TestArrayElementwiseOp::test_pow_array - AssertionError:
- Tolerance issue
```
E       AssertionError:
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       Mismatched elements: 1 / 6 (16.7%)
E       Max absolute difference: 7.6293945e-06
E       Max relative difference: 1.5258789e-07
E        x: array([[ 1.      +0.j, 32.      +0.j, 81.00001 +0.j],
E              [64.      +0.j, 24.999996+0.j,  6.      +0.j]], dtype=complex64)
E        y: array([[ 1.+0.j, 32.+0.j, 81.+0.j],
E              [64.+0.j, 25.+0.j,  6.+0.j]], dtype=complex64)
```

- [ ]  core_tests/fusion_tests/test_array.py::TestFusionArrayMethods::test_all - AssertionError:
- [ ]  core_tests/fusion_tests/test_array.py::TestFusionArrayMethods::test_max - AssertionError:
- [ ]  core_tests/fusion_tests/test_array.py::TestFusionArrayMethods::test_min - AssertionError:
- [ ]  core_tests/fusion_tests/test_array.py::TestFusionArrayMethods::test_prod - AssertionError:
- [ ]  core_tests/fusion_tests/test_array.py::TestFusionArrayMethods::test_sum - AssertionError:
[...]

PR #4504, opened 2020-12-25
Slight code improvements for the CUB reduction kernel

leofang:
Related to #4484, #4485. Tons of thanks to @emcastillo who figured out that the last segment size is corrupted in HIP due to the difference of its memory model from CUDA's.

**UPDATE:** This PR mainly does two things:
1. Instead of modifying the kernel argument directly, we copy it to a register and modify the local copy.
2. Sync threads only on HIP.


emcastillo:
Is this ready? seems to be so

leofang:
I prefer #4485 to go in first, so that I can resolve a possible conflict here.

leofang:
> I prefer #4485 to go in first, so that I can resolve a possible conflict here.

@emcastillo I resolved the conflict. PTAL.

leofang:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4509, opened 2020-12-25
Document `random.uniform` may return high

### LINKED ISSUE #4139: cupy.random.uniform(low=low, high=high) returns high
```python
import cupy as cp

cp.random.seed(seed=13)
high = 2**32 - 1
low = 2**32 - 3
size = 100000000
(cp.random.uniform(low=low, high=high, size=size) == high).sum()
```
returns `array(11)`, while `array(0)` is expected, as "Samples are drawn from a uniform distribution over the half-open interval [low, high)".

Interestingly, the documentation of standard `random.uniform` says "The end-point value b may or may not be included in the range depending on floating-point rounding in the equation `a + (b-a) * random()`".  `RandomState._scale_kernel ` seems to do the same thing, so it might be a documentation problem.

BTW, the behavior looks like it is consistent with `numpy`.

* Conditions
```
CuPy Version          : 6.0.0
CUDA Build Version    : 10000
CUDA Driver Version   : 10010
CUDA Runtime Version  : 10000
cuDNN Build Version   : 7301
cuDNN Version         : 7605
NCCL Build Version    : 1000
Linux 4.4.180-102-default 
Intel(R) Xeon(R) Gold 5118 CPU @ 2.30GHz
Tesla P100
```

toslunar:
Close #4139.

### DIFF
--- a/cupy/random/_distributions.py
+++ b/cupy/random/_distributions.py
@@ -834,7 +834,8 @@ def uniform(low=0.0, high=1.0, size=None, dtype=float):
     """Returns an array of uniformly-distributed samples over an interval.
 
     Samples are drawn from a uniform distribution over the half-open interval
-    ``[low, high)``.
+    ``[low, high)``. The samples may contain the ``high`` limit due to
+    floating-point rounding.
 
     Args:
         low (float): Lower end of the interval.


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

Issue #4387, opened 2020-12-02
`tensordot_core` raises `CUDA_ERROR_INVALID_VALUE`


kmaehashi:
```py
>>> cp.outer(cp.arange(2**23), cp.arange(3))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kenichi/Development/cupy/cupy/linalg/_product.py", line 245, in outer
    return core.tensordot_core(a, b, None, n, m, 1, ret_shape)
  File "cupy/core/_routines_linalg.pyx", line 421, in cupy.core._routines_linalg.tensordot_core
    cpdef ndarray tensordot_core(
  File "cupy/core/_routines_linalg.pyx", line 493, in cupy.core._routines_linalg.tensordot_core
    return _integral_tensordot_core(b, a, out, m, n, k, dtype, ret_shape)
  File "cupy/core/_routines_linalg.pyx", line 320, in cupy.core._routines_linalg._integral_tensordot_core
    kern(grid, block, args=args, shared_mem=shared_mem)
  File "cupy/core/raw.pyx", line 89, in cupy.core.raw.RawKernel.__call__
    self.kernel(
  File "cupy/cuda/function.pyx", line 201, in cupy.cuda.function.Function.__call__
    _launch(
  File "cupy/cuda/function.pyx", line 183, in cupy.cuda.function._launch
    driver.launchKernel(
  File "cupy_backends/cuda/api/driver.pyx", line 299, in cupy_backends.cuda.api.driver.launchKernel
    check_status(status)
  File "cupy_backends/cuda/api/driver.pyx", line 124, in cupy_backends.cuda.api.driver.check_status
    raise CUDADriverError(status)
cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_VALUE: invalid argument
```

emcastillo:
Seems like this comes from the matrix int kernel with a large grid size

leofang:
Interestingly, this works fine on HIP (even without the fix #4512) but not on CUDA. Magic! 😂

ashkspark:
I am getting a similar error when my array becomes large. I am using RTX 8000 . Here is the error:

cupy/core/_routines_linalg.pyx in cupy.core._routines_linalg.tensordot_core()

cupy/core/_routines_linalg.pyx in cupy.core._routines_linalg.tensordot_core()

[...]

PR #4217, opened 2020-11-02
ROCm: Cover more new BLAS/SOLVER functions

leofang:
Related to #4132. ~~Blocked by #4205~~. 

TODO:
- [x] Add SVD (gesvd) from ROCm 3.9.0
- [x] Add  (`*geam` and `*dgmm`) (#4263) from ROCm 3.7.0+
- [ ] ~~Update the stubs from #4073?~~
**UPDATE**: `gels` was just recently added upstream (https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/164) and is not included in any official ROCm release yet
- [x] Force push

PR #3994, opened 2020-09-16
Add a kernel for integer GEMM

emcastillo:
Fixes #3941

We rely only on cublas for the matrix multiplication and we cast the inputs to float and then back to integral types again. This causes precision issues.

This also has an speedup of 45% for a matrix product with `dtype=numpy.int32` and`shape=(5000, 5000)`

PR #4512, opened 2020-12-26
Fix integer GEMM

leofang:
Follow-up of #3994. ~~Blocked by #4217.~~

This PR fixes two things:
1. For ROCm/HIP, there is a bunch of errors in `test_product.py` (see https://github.com/cupy/cupy/issues/4484#issuecomment-751204258) due to ~~the different handlings in out-of-bound access between CUDA and HIP. A check is added on HIP.~~ a strange behavior in HIP (likely the same cause as in #4504. Fixed by syncing the block only on HIP.
2. The kernel allocates shared memory statically, not dynamically, so there's no reason to pass `shared_mem` when launching the kernel.

### DIFF
--- a/cupy/core/_routines_linalg.pyx
+++ b/cupy/core/_routines_linalg.pyx
@@ -163,6 +163,11 @@ __global__ void _tensordot_core_int_kernel(
     }
     __syncthreads();
 
+    // On CUDA, it's okay that m, n exceed matrix bounds as all work is in
+    // either registers or shared memory, and out-of-bounds rC[n][m] will
+    // not be saved later.
+    // On HIP, however, bound checks must be done to ensure correctness.
+
     for (kk = 0; kk < K - BLK_K; kk += BLK_K)
     {
         offs_dA += BLK_K * M;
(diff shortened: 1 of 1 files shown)


leofang:
Jenkins, test this please

emcastillo (approved):

leofang:
> there is a bunch of errors in `test_product.py` (see [#4484 (comment)](https://github.com/cupy/cupy/issues/4484#issuecomment-751204258)) due to the different handlings in out-of-bound access between CUDA and HIP. 

Frankly I don't understand why this isn't an issue on CUDA. @emcastillo any insight? 🙂

leofang:
I looked at the PTX code in CUDA but it wasn't obvious to me if there is any special guard: https://godbolt.org/z/ad8Pje

emcastillo:
Actually I have no idea ...

leofang:
> Actually I have no idea ...

Haha no worries. I will write to a MAGMA developer that I know and see what he know about.

leofang:
btw it's best to merge this PR after #4217 is in.

leofang:
> I will write to a MAGMA developer that I know and see what he know about. But this PR should be ready to go.

The discussion is in MAGMA's mailing list: https://groups.google.com/a/icl.utk.edu/g/magma-user/c/igc66uduTfI.

leofang:
Jenkins, test this please

emcastillo:
I wonder if this will fix https://github.com/cupy/cupy/issues/4387

leofang:
Just checked, unfortunately it's not fixed 😅 I checked the grid and block sizes and everything looked fine to me. Not sure why the driver complained about invalid arguments...

emcastillo:
Hey, is this ready for review? :)

leofang:
> Hey, is this ready for review? :)

@emcastillo I think so 🙂

leofang:
Jenkins, test this please

emcastillo (approved):

PR #4517, opened 2020-12-31
Add DLPack as submodule & Fix multiple issues

### LINKED ISSUE #3611: Segfault when DLPack tensor consumed twice
After calling `fromDlpack` on a dlpack object, the original object becomes `used_dltensor`, and can not be reused, otherwise an invalid memory access will occur.

* Code to reproduce

``` python
import cupy as cp
n = 100
X = cp.random.random((n, 2))
dlpack = X.toDlpack()

y = cp.fromDlpack(dlpack)
y = cp.fromDlpack(dlpack)  # segfault
```

segfault

### LINKED ISSUE #4513: Make DLPack a submodule?
I am looking into DLPack as we likely need to fix it for ROCm (it supports the `kDLROCM` option). One thing I noticed is that we basically reimplemented everything from the DLPack header [`dlpack.h`](https://github.com/dmlc/dlpack/blob/master/include/dlpack/dlpack.h) in `cupy/core/dlpack.pyx`. I suppose like Jitify it's also a single header thing, so we might as well make it a submodule? 

This will also prepare us for the Python Array API involvement, for which DLPack is being considered as the data exchange format (https://github.com/data-apis/consortium-feedback/issues/1; see also the [Array API Standard draft](https://data-apis.github.io/array-api/latest/design_topics/data_interchange.html)), but it currently lacks 
1. support for complex numbers (https://github.com/dmlc/dlpack/issues/50)
2. specification for synchronization semantics as in CUDA Array Interface v3 (https://github.com/dmlc/dlpack/issues/57)

so it will likely be revised in the near future. If this is acceptable then please assign it to me.

leofang:
Close #3611. Close #4513. 

~~PR description to follow.~~ Happy New Year!

**UPDATE**: As an overhaul for the `cupy.core.dlpack` module, this PR does multiple things:
- ~~Add DLPack as a git submodule~~ & include `dlpack.h` from there (see #4513 for the reason) **UPDATE**: We bundle `dlpack.h` directly without making it a submodule.
- Add two new enum values `kDLVulkan` and `kDLBfloat` newly added in upstream
- Safer PyCapsule handling by using `PyCapsule_IsValid()`
  - One could have instead called `PyCapsule_GetName()` to do less checks; however, Cython has a bug in handling the return value of type `const char*` (cython/cython#3964), so I switched to this API.
- Support `kDLROCM`. If CuPy is built against HIP, this is used instead of `kDLGPU`. This follows at least PyTorch's implementation.
- Fix segfault if a PyCapsule is consumed multiple times (#3611).
- Raise an error if the input PyCapsule is already consumed. This aligns with the intended usage (though DLPack's documentation is poor and [people are working on it](https://github.com/data-apis/array-api/pull/106)) and serves as an additional safeguard to the above fix.
- More sanity checks and bug fixes, which I'll comment inline.
- Add tests.
- Document that the current DLPack spec is ***unsafe*** with regard to CUDA/HIP streams (dmlc/dlpack#57).

### DIFF
--- a/cupy/core/dlpack.pxd
+++ b/cupy/core/dlpack.pxd
@@ -1,5 +1,3 @@
-# DLPACK_VERSION: 010
-
 from cupy.core.core cimport ndarray
 
 cpdef object toDlpack(ndarray array) except +
--- a/cupy/core/dlpack.pyx
+++ b/cupy/core/dlpack.pyx
@@ -15,7 +15,7 @@ from cupy.cuda cimport memory
 import cupy
 
 
-cdef extern from './include/cupy/dlpack/include/dlpack/dlpack.h' nogil:
+cdef extern from './include/cupy/dlpack/dlpack.h' nogil:
     cdef enum DLDeviceType:
         kDLCPU
         kDLGPU
--- a/cupy/core/dlpack.pyx
+++ b/cupy/core/dlpack.pyx
@@ -62,7 +62,7 @@ cdef void pycapsule_deleter(object dltensor):
     if cpython.PyCapsule_IsValid(dltensor, 'dltensor'):
         dlm_tensor = <DLManagedTensor*>cpython.PyCapsule_GetPointer(
             dltensor, 'dltensor')
-        deleter(dlm_tensor)
+        dlm_tensor.deleter(dlm_tensor)
 
 
 cdef void deleter(DLManagedTensor* tensor) with gil:
@@ -70,8 +70,8 @@ cdef void deleter(DLManagedTensor* tensor) with gil:
         return
     stdlib.free(tensor.dl_tensor.shape)
     cpython.Py_DECREF(<ndarray>tensor.manager_ctx)
-    stdlib.free(tensor)
     tensor.manager_ctx = NULL
+    stdlib.free(tensor)
 
 
 # The name of this function is following the framework integration guide of
@@ -171,7 +171,7 @@ cdef class DLPackMemory(memory.BaseMemory):
         cdef DLManagedTensor* dlm_tensor = self.dlm_tensor
         # dlm_tensor could be uninitialized if an error is raised in __init__
         if dlm_tensor != NULL:
-            deleter(dlm_tensor)
+            dlm_tensor.deleter(dlm_tensor)
 
 
 # The name of this function is following the framework integration guide of
--- a/cupy/core/dlpack.pyx
+++ b/cupy/core/dlpack.pyx
@@ -56,6 +56,12 @@ cdef extern from './include/cupy/dlpack/dlpack.h' nogil:
         void (*deleter)(DLManagedTensor*)  # noqa: E211
 
 
+def get_build_version():
+    # We use the commit ID here because DLPACK_VERSION is not reliable...
+    # The commit ID is from the latest change made to dlpack.h.
+    return '3efc489'
+
+
 cdef void pycapsule_deleter(object dltensor):
     cdef DLManagedTensor* dlm_tensor
     # Do not invoke the deleter on a used capsule
--- a/cupy/core/dlpack.pyx
+++ b/cupy/core/dlpack.pyx
@@ -201,6 +201,13 @@ cpdef ndarray fromDlpack(object dltensor) except +:
         from a :class:`~cupy.ndarray` to a DLPack tensor (which is encapsulated
         in a :class:`PyCapsule` object).
 
+    .. warning::
+
+        As of the DLPack v0.3 specification, it is (implicitly) assumed that
+        the user is responsible to ensure the Producer and the Consumer are
+        operating on the same stream. This requirement might be relaxed/changed
+        in a future DLPack version.
+
     .. admonition:: Example
 
         >>> import cupy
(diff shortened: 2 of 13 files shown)


cupy_setup_build.py
```diff
@@ -323,6 +322,17 @@
             'version_method': build.get_thrust_version,
         })
 
+MODULES.append({
```
emcastillo, line 325: Can we disable this for HIP? or should we do it inside the module

leofang: Hi @emcastillo, I updated the PR description. DLPack is meant to support a wide spectrum of architectures, and there's a dedicated device type called `kDLROCM`, so we can't disable it.

I isolated `cupy.core.dlpack` out because I want to add `s['depends']` to it. If I don't pull it out but instead add `s['depends']` in the loop over module files `f` when `f == `cupy.core.dlpack`, many irrelevant modules would always be rebuilt for some reason. If you know how to fix it I'd be happy to move it back under the `cuda` module (as it's really not that special) 🙂

cupy/core/dlpack.pxd
```diff
@@ -1,5 +1,3 @@
-# DLPACK_VERSION: 010
```
leofang, line 1: This value is apparently not used anywhere, but should we record this information in the module (and perhaps report it in `cupy.show_config()`?

cupy/core/dlpack.pyx
```diff
@@ -5,75 +5,73 @@ from libc.stdint cimport uint8_t
...
 
 
 cdef void deleter(DLManagedTensor* tensor) with gil:
     if tensor.manager_ctx is NULL:
         return
     stdlib.free(tensor.dl_tensor.shape)
     cpython.Py_DECREF(<ndarray>tensor.manager_ctx)
-    stdlib.free(tensor)
     tensor.manager_ctx = NULL
+    stdlib.free(tensor)
```
leofang, line 74: Fixed a technically illegal access.

cupy/core/dlpack.pyx
```diff
@@ -234,10 +257,14 @@ cpdef ndarray fromDlpack(object dltensor) except +:
     shape_vec.assign(shape, shape + ndim)
 
     if mem.dlm_tensor.dl_tensor.strides is NULL:
+        # Make sure this capsule will never be used again.
+        cpython.PyCapsule_SetName(mem.dltensor, 'used_dltensor')
```
leofang, line 261: The name change is deferred to right before `return` as errors could be raised anywhere before arriving here (or below), and technically the tensor is not consumed.

setup.cfg
```diff
@@ -1,8 +1,5 @@
-[flake8]
-exclude = .eggs,*.egg,build,docs,.git
-
-[pep8]
-exclude = .eggs,*.egg,build,.git
+[flake8]  # autopep8 would honor this setting, but not the other way around
+exclude = .eggs,*.egg,build,docs,.git,*cupy/core/include/**.py
```
leofang, line 2: This fixed a crazy bug in autopep8 that took me quite a while to understand. The long story in short is autopep8 would pick up both `flake8` and `pep8` sections here, causing the exclude patterns being overwritten in an unpredictable (to me) way. A simpler fix is to duplicate the pattern to both sections to avoid conflicts, but then it's a duplication so we may as well remove one of them 🤦🏻‍♂️

leofang:
Jenkins, test this please

cupy/core/dlpack.pyx
```diff
@@ -133,23 +137,41 @@ cdef class DLPackMemory(memory.BaseMemory):
...
-        # Make sure this capsule will never be used again.
-        cpython.PyCapsule_SetName(dltensor, 'used_dltensor')
+        self.size = dlm_tensor.dl_tensor.dtype.bits * n // 8
 
     def __dealloc__(self):
-        self.dlm_tensor.deleter(self.dlm_tensor)
+        cdef DLManagedTensor* dlm_tensor = self.dlm_tensor
+        # dlm_tensor could be uninitialized if an error is raised in __init__
+        if dlm_tensor != NULL:
+            dlm_tensor.deleter(dlm_tensor)
```
leofang, line 174: This is the real fix for the segfault: When a PyCapsule is consumed, `PyCapsule_GetPointer` would raise a `ValueError` due to the name change, and the cleanup kicks in and invokes `dlm_tensor->deleter`, but `dlm_tensor` is `NULL`...

emcastillo:
We discussed this offline. We want to merge this PR as it fixes a lot of things, but we feel that making dlpack a module itself is a bit heavy. DLPack code is very small so in the meantime just having a copy of the header in CuPy feels enough for us.
Is there any case that we miss that having it as a module would be very beneficial?

leofang:
@emcastillo Thanks. Do you mean I should just check in `dlpack.h` (like we've done for fp16 headers) without making it a submodule? I can do it, as the maintenance load is the same.

setup.cfg
leofang: This change is reverted in 47314ff since we no longer download the whole git repo of DLPack.

leofang:
Jenkins, test this please

leofang:
This is ready. PTAL.

emcastillo (approved):

leofang:
Thanks!

leofang:
@emcastillo Forgot to ask: I suppose we should backport this PR as it fixes several things? If you don't need the bundle for v8 I can send a stripped down version.

emcastillo:
Yeah, let's go with a stripped one.
Thanks!

PR #4521, opened 2021-01-05
Fix fusion tests for HIP

emcastillo:
Merge after #4485 
changes are those of commit https://github.com/cupy/cupy/commit/21fc6c83805d9d62ef721f080d93b9ead29025a1

### DIFF
--- a/cupy/_math/floating.py
+++ b/cupy/_math/floating.py
@@ -1,5 +1,6 @@
 from cupy import core
 from cupy._math import ufunc
+from cupy.cuda import runtime
 
 
 signbit = core.create_ufunc(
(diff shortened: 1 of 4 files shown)


leofang (changes requested):
Could you revert the changes in these files?
- `cupy_backends/hip/cupy_hip.h`
- `tests/cupy_tests/core_tests/fusion_tests/test_example.py`

emcastillo:
Sorry, I can't see what is wrong with `tests/cupy_tests/core_tests/fusion_tests/test_example.py`
The test needs to be skipped

leofang:
> Sorry, I can't see what is wrong with `tests/cupy_tests/core_tests/fusion_tests/test_example.py`
> The test needs to be skipped

Why? I thought you already made fusion work?

emcastillo:
HIP does not support cooperative groups for doing inter-block barriers, so in some fusion cases it will fail.
The example relies precisely on that :(

leofang:
Ohhh OK. I missed that, thanks. So do we want to raise in fusion when cooperative group is required?

emcastillo:
That's what we do!
https://github.com/cupy/cupy/blob/master/cupy/cuda/compiler.py#L414

leofang:
...Right...I really should get some more sleep 😧😵🥱

emcastillo:
I think that just showing some random cuda related API message is not helpful at all for end users, so we have to catch it in the fusion side and raise a more friendly user related error.

leofang:
@emcastillo Could you merge or rebase with master?

leofang:
Jenkins, test this please

asi1024:
LGTM!

PR #4534, opened 2021-01-08
Skip sparse svd `test_dense` for older scipy and complex64

emcastillo:
This test fails in the daily shuffles requiring a tolerance of 1e-2.
This tolerance is too low, so it's better to just skip it if scipy is < 1.4


emcastillo:
Not present in v8. Closing

PR #4536, opened 2021-01-08
Fix sparse format of `kron`

toslunar:
Follow scipy/scipy#13100

### DIFF
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_construct.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_construct.py
@@ -401,7 +401,7 @@ arrs = [[[0]],
     'arrA': arrs,
     'arrB': arrs,
 }))
-@testing.with_requires('scipy')
+@testing.with_requires('scipy>=1.6')
 class TestKron(unittest.TestCase):
 
     def _make_sp_mat(self, xp, sp, arr, dtype):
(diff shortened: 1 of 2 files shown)


toslunar:
The `cupy_pr` CI should have SciPy 1.6 tests.  Blocked by chainer/chainer-test#630.

tests/cupyx_tests/scipy_tests/sparse_tests/test_construct.py
```diff
@@ -401,7 +401,7 @@ def test_diags_non_square(self, xp, sp):
     'arrA': arrs,
     'arrB': arrs,
 }))
-@testing.with_requires('scipy')
+@testing.with_requires('scipy>=1.6')
```
leofang, line 404: Would this requirement be too strong? This method already exists before SciPy 1.6. Perhaps we can add the requirement at the method level?

emcastillo: I think adding here it is fine, there is only one test and currently, it won't pass for scipy<1.6

toslunar:
```python
if testing.scipy_satisfies('<1.6'):  # similar to testing.numpy_satisfies
    kron = kron.A
```

is an option.  But I justify the current fix as a kind of white-box test.  CuPy's `kron` does not directly refer `cupy.sparse`, i.e. the dependency `scipy>=1.6` is only for the testing purpose.

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4539, opened 2021-01-11
Add cupy.round

grlee77:
NumPy has a `round` function that is equivalent to `around`, but CuPy currently does not define this function.

When checking the proposed Array API Standard, I see that `round` rather than `around` is specified in the standard, so it would be good to support it here. See:
https://data-apis.github.io/array-api/latest/API_specification/elementwise_functions.html#round-x

### DIFF
--- a/docs/source/reference/math.rst
+++ b/docs/source/reference/math.rst
@@ -49,6 +49,7 @@ Rounding
 
    cupy.around
    cupy.round_
+   cupy.round
    cupy.rint
    cupy.fix
    cupy.floor
--- a/docs/source/reference/math.rst
+++ b/docs/source/reference/math.rst
@@ -49,7 +49,6 @@ Rounding
 
    cupy.around
    cupy.round_
-   cupy.round
    cupy.rint
    cupy.fix
    cupy.floor
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -580,6 +580,7 @@ from cupy._math.hyperbolic import sinh  # NOQA
 from cupy._math.hyperbolic import tanh  # NOQA
 
 from cupy._math.rounding import around  # NOQA
+from cupy._math.rounding import around as round # NOQA
 from cupy._math.rounding import ceil  # NOQA
 from cupy._math.rounding import fix  # NOQA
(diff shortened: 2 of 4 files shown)


cupy/_math/rounding.py
```diff
@@ -29,6 +29,11 @@ def round_(a, decimals=0, out=None):
     return around(a, decimals, out=out)
 
 
+round = around
+round.__name__ = 'round'
+round.__doc__ = around.__doc__.replace('around', 'round')
+
+
```
asi1024, line 36: In NumPy implementation, `numpy.round` is just an alias of `numpy.round_` How about defining `round = round_` here?

grlee77: Do you mean just `round = round_`, removing the `__doc__` setting? NumPy's `round` does have a docstring that is minimal and just refers the user to `around`.

grlee77: Can you make a suggestion here with exactly what you want and I can accept it?

docs/source/reference/math.rst
```diff
@@ -49,6 +49,7 @@ Rounding
 
    cupy.around
    cupy.round_
+   cupy.round
```
kmaehashi, line 52: Conventionally, functions undocumented (ie, not listed in the routines list) in NumPy are left undocumented in CuPy.

cupy/_math/rounding.py
kmaehashi: In CuPy, undocumented aliases are defined like this:

https://github.com/cupy/cupy/blob/v9.0.0b1/cupy/__init__.py#L708-L711

So just adding `from cupy._math.rounding import round_ as round` to `cupy/__init__.py` sounds enough to me.

docs/source/reference/math.rst
asi1024: @grlee77 Could you address this comment?

grlee77: I have removed it now.

asi1024:
Jenkins, test this please.

asi1024 (approved):

asi1024:
LGTM! Thanks!

PR #4543, opened 2021-01-12
Fix windows build of random module

emcastillo:
Windows does not define `ssize_t`.
Verified that the module builds and the tests work.


kmaehashi:
LGTM!
pfnCI, test this please.

PR #4544, opened 2021-01-12
Fix `__format__`

### LINKED ISSUE #4532: Unsupported f-string format
`f"{element}:.2f"` isn't a valid format string

* Code to reproduce
```python
import numpy as np
import cupy as cp

arr = [1.12345]
np_arr = np.asarray(arr)
cp_arr = cp.asarray(arr)

print(f"With 2 precisions: {np_arr[0]:.2f})                                # With 2 precisions: 1.12

print(f"With 2 precisions: {cp_arr[0]:.2f}) 

  File "er.py", line 10, in <module>
    print(f"With 2 precisions: {cp_arr[0]:.2f}")
TypeError: unsupported format string passed to cupy.core.core.ndarray.__format__
```

Conditions (you can just paste the output of `python -c 'import cupy; cupy.show_config()'`)
```console
CuPy Version          : 6.0.0
CUDA Build Version    : 10000
CUDA Driver Version   : 11010
CUDA Runtime Version  : 10000
cuDNN Build Version   : 7301
cuDNN Version         : 7605
NCCL Build Version    : 1000
```

emcastillo:
Closes #4532

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1508,6 +1508,9 @@ cdef class ndarray:
     def __str__(self):
         return str(self.get())
 
+    def __format__(self, format_spec):
+        return self.get().__format__(format_spec)
+
     # -------------------------------------------------------------------------
     # Methods outside of the ndarray main documentation
     # -------------------------------------------------------------------------
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1509,7 +1509,7 @@ cdef class ndarray:
         return str(self.get())
 
     def __format__(self, format_spec):
-        return self.get().__format__(format_spec)
+        return format(self.get(), format_spec)
 
     # -------------------------------------------------------------------------
     # Methods outside of the ndarray main documentation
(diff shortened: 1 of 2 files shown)


tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -570,6 +570,11 @@ def test_bytes_tobytes_scalar_array(self, xp, dtype):
         x = xp.array(3, dtype)
         return bytes(x)
 
+    @testing.numpy_cupy_equal()
+    def test_format(self, xp):
+        x = xp.array(1.12345)
+        return x.__format__('.2f')
```
toslunar, line 576:
```suggestion
        return format(x, '.2f')
```

like `bytes(x)` instead of `x.__bytes__()`.   FYI, `numpy.ndarray` does not have `__bytes__` but `bytes(x)` works.

cupy/core/core.pyx
```diff
@@ -1508,6 +1508,9 @@ cdef class ndarray:
     def __str__(self):
         return str(self.get())
 
+    def __format__(self, format_spec):
+        return self.get().__format__(format_spec)
```
toslunar, line 1512: For consistency around the lines, the builtin `format` looks better here, too.
```suggestion
        return format(self.get(), format_spec)
```

Do you have other pros/cons in mind?  e.g. performance

emcastillo: I think that since this implies device synchronization, there is not much to care about. Thanks!

toslunar (approved):
LGTM

toslunar:
pfnCI, test this please.

PR #1055, opened 2018-03-23
limit arch to the maximum value allowed in each NVRTC version

kmaehashi:
Fixes #1048.
See also: http://docs.nvidia.com/cuda/nvrtc/index.html#group__options

PR #4558, opened 2021-01-18
CUDA 11.1: Wrap the new `nvrtcGetCUBIN` API

leofang:
**UPDATE**: this PR only adds Cython wrappers for `nvrtcGetCUBIN`, and the migration strategy to this new API is to be discussed.

This PR changes the NVRTC default for CUDA 11.1+ from compiling PTX to compiling CUBIN, bypassing the driver JIT step when the module is loaded. This could potentially improve the overall performance (to be verified) in terms of both reducing overall time and enabling better code generation.

For CUDA 9.2 - 11.0, this PR does not do anything.

TODO:
- [x] Run some benchmarks to see if we get some speedup.

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -238,7 +238,7 @@ def compile_using_nvrtc(source, options=(), arch=None, filename='kern.cu',
             if _cuda_version >= 11010 and jitify:
                 # Convert the virtual arch to a real arch
                 options = [f'-arch=sm_{arch}' if opt.startswith('-arch=')
-                    else opt for opt in options]
+                           else opt for opt in options]
                 options = tuple(options)
             ptx, mapping = prog.compile(options, log_stream)
         except CompileException as e:
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -15,8 +15,10 @@ from cupy_backends.cuda.api import runtime
 from cupy_backends.cuda.libs import nvrtc
 from cupy import _util
 
-if not runtime.is_hip and driver.get_build_version() > 0:
-    from cupy.cuda.jitify import jitify
+if not runtime.is_hip:
+    _cuda_version = driver.get_build_version()
+    if _cuda_version > 0:
+        from cupy.cuda.jitify import jitify
 
 
 _nvrtc_version = None
@@ -115,30 +117,10 @@ def _get_nvrtc_version():
     return _nvrtc_version
 
 
-# Known archs for Tegra/Jetson/Xavier/etc
-_tegra_archs = ('53', '62', '72')
-
-
 @_util.memoize(for_each_device=True)
 def _get_arch():
-    # See Supported Compile Options section of NVRTC User Guide for
-    # the maximum value allowed for `--gpu-architecture`.
-    major, minor = _get_nvrtc_version()
-    if major < 10 or (major == 10 and minor == 0):
-        # CUDA 9.x / 10.0
-        _nvrtc_max_compute_capability = '70'
-    elif major < 11:
-        # CUDA 10.1 / 10.2
-        _nvrtc_max_compute_capability = '75'
-    else:
-        # CUDA 11.0 / 11.1
-        _nvrtc_max_compute_capability = '80'
-
     arch = device.Device().compute_capability
-    if arch in _tegra_archs:
-        return arch
-    else:
-        return min(arch, _nvrtc_max_compute_capability)
+    return arch
 
 
 def _is_cudadevrt_needed(options):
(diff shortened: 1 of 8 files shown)


leofang:
2719bff is a demo to show the minimal change needed to make this work on CUDA 11.1. But we need to retain backward compatibility for older CUDA as well as HIP, so some more work will be required.

cupy/cuda/compiler.py
```diff
@@ -115,30 +117,10 @@ def _get_nvrtc_version():
...
-    else:
-        # CUDA 11.0 / 11.1
-        _nvrtc_max_compute_capability = '80'
-
     arch = device.Device().compute_capability
-    if arch in _tegra_archs:
-        return arch
-    else:
-        return min(arch, _nvrtc_max_compute_capability)
+    return arch
```
leofang, line 123: I simplify this function by just returning the current device's CC. The current implementation was initially done by @kmaehashi in #1055 due to the NVRTC issue in CUDA 8. But since we drop CUDA 8 support, and apparently the issue is resolved in CUDA 9+ (all CCs' supported by CUDA are listed in the NVRTC option), looks like we no longer need this workaround.

Refs:
- https://docs.nvidia.com/cuda/archive/9.2/nvrtc/index.html#group__options (CUDA 9.2)
- https://docs.nvidia.com/cuda/archive/10.0/nvrtc/index.html#group__options (CUDA 10.0)
- https://docs.nvidia.com/cuda/archive/10.1/nvrtc/index.html#group__options (CUDA 10.1)
- https://docs.nvidia.com/cuda/archive/10.2/nvrtc/index.html#group__options (CUDA 10.2)
- https://docs.nvidia.com/cuda/archive/11.0/nvrtc/index.html#group__options (CUDA 11.0)
- https://docs.nvidia.com/cuda/archive/11.1.1/nvrtc/index.html#group__options (CUDA 11.1.1)

leofang:
Looks like this new feature actually increases the compilation overhead...😞 

Below I run two test files for different purposes:
- `test_raw.py`: Tests in this file do not involve the cubin cache on disk, so it gives a measure for total compilation time
- `test_fft.py`: Kernels used in this test are cached on disk, so the first run involves compiling, and the second run involves loading from disk (which in theory would involve driver JIT)

Timing | `test_raw.py` | `test_fft.py` (1st run) | `test_fft.py` (2nd run)
------|--------------|-------------|--------
before this PR | 44.8 s | 36.4 s | 10.7 s
with this PR | 54.8 s | 45.3 s | 10.8 s

So we can see 1. the compile time increases, 2. the load time does not decrease. Not sure what I'm missing to properly utilize this new feature...

leofang:
OK I see where I got it wrong. The JIT'd binaries are cached by the driver, and I forgot to disable it. By setting `CUDA_CACHE_DISABLE=1` I got a timing making better sense:

Timing | `test_raw.py` | `test_fft.py` (1st run) | `test_fft.py` (2nd run)
------|--------------|-------------|--------
before this PR | 52.4 s | 46.3 s | 20.9 s
with this PR | 55.1 s | 45.2 s | 10.8 s 

1. The increase of compile time is mild, and likely only because Jitify still generates PTX (in `test_raw.py`)
2. The driver JIT time is not negligible and was saved because CUDA did the cache for us

@takagi How about we switch to `getCUBIN()` from `getPTX()` when we detect a user sets `CUDA_CACHE_DISABLE=1`?

takagi:
I think switching when `CUDA_CACHE_DISABLE=1` is okay. BTW, CuPy's kernel cache on the disk will be changed from PTX to cubin, then what happens if the GPU on the machine is replaced with another and the cached cubin is loaded??

leofang:
Hi @takagi I am still wondering if `getCUBIN()` should always be enable on CUDA 11.1+. By doing so we shift the cache responsibility from CUDA driver to CuPy, but we already do this anyway.

For invalidating the CUBIN when using a different GPU is in use, I think it's already done when generating CuPy's cache key. I will add CUDA version to the key to further disambiguate it. (By the way, on HIP we also cache binary files (hasco).)

Let me know what you think and I'll proceed accordingly 🙂

takagi:
Thanks for explaining! I also checked CuPy's cache key and looks having enough information to invalidate the CUBIN properly.

BTW, I read through the changes again and now I think using `nvrtcGetCUBIN` only for the case `CUDA_CACHE_DISABLE=1` with adding extra branches doesn't make us much happier. How about just adding the wrappers for `nvrtcGetCUBIN` and cleaning some codes at this time?

leofang:
> BTW, I read through the changes again and now I think using `nvrtcGetCUBIN` only for the case `CUDA_CACHE_DISABLE=1` with adding extra branches doesn't make us much happier. 

Yes, this is what I was wondering too. 

> How about just adding the wrappers for `nvrtcGetCUBIN` and cleaning some codes at this time?

Did you mean we don't use `getCUBIN()` in `cupy/cuda/compiler.py` at all, or you meant something else? I am happy to do either way, just need to clarify.

takagi:
> Did you mean we don't use getCUBIN() in cupy/cuda/compiler.py at all, or you meant something else?

Yes, what I meant is the former. Of course, I always respect your highly productive works!

leofang:
Thanks, @takagi. I cleaned up `compiler.py` and `test_raw.py`, and left the Cython wrappers and `_get_arch()` untouched (https://github.com/cupy/cupy/pull/4558#discussion_r559974610). PTAL.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
CI out of disk space 😂

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

takagi:
LGTM! Waiting for CI pass.

takagi:
Thanks!

leofang:
Thanks, @takagi! I'll open an issue to discuss when to turn on `getCUBIN()`.

PR #4535, opened 2021-01-08
[backport] Import DLPack header file & Fix multiple issues

emcastillo:
Backport of #4517

PR #4559, opened 2021-01-18
Fix DLPack header version

leofang:
Follow-up of #4517. @tqchen kindly fixed this issue in upstream right after I reported it, so given that we manually bundle the header we can just apply the same fix ourselves.

This PR should be backported (after #4535).


emcastillo (approved):

emcastillo:
Jenkins, test this please

PR #4562, opened 2021-01-19
Support passing pointers as integers in memory copy routines

### LINKED ISSUE #4317: Make `ctypes.c_void_p` optional for copying to/from host memory?
Take this function as an example: https://github.com/cupy/cupy/blob/50361e0d42ca8e41e7200382c3031cfc4e201764/cupy/cuda/memory.pyx#L368-L378
I found it a bit cumbersome having to wrap a pointer by `c_void_p` first. Using integer as input is more intuitive IMHO. IIUC this is because the distinction between host and device pointers was desired: `c_void_p` is always for the former and `MemoryPointer` for the latter. (Correct me if I'm wrong!)

However, I think in all these copy functions we can just accept `int` as inputs everywhere, and pass `runtime.memcpyDefault` to let the runtime API figure out where the data is (through inspecting the pointer attributes internally).

Perhaps the above proposal is too disruptive and thus backward incompatible, so I wonder if it would be more acceptable if we accept `int` wherever `c_void_p` is expected? At least to me it's a milder change:
```python
     cpdef copy_from_host(self, mem, size_t size): 
         """Copies a memory sequence from the host memory. 
  
         Args: 
             mem (int or ctypes.c_void_p): Source memory pointer. 
             size (int): Size of the sequence in bytes. 
  
         """ 
         if size > 0:
             ptr = mem if isinstance(mem, int) else mem.value
             runtime.memcpy(self.ptr, ptr, size,
                            runtime.memcpyDefault) 
```
The benefit is that lines such as this one get cleaned up a bit:
https://github.com/cupy/cupy/blob/abbcecc59828bb381e3e6fa33889b4b1631c3a74/cupy/core/core.pyx#L1272

leofang:
Close #4317. Wrapping pointers using `ctypes.c_void_p()` is still accepted, but it's not required.

TODO:
- [x] Add test

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -7,7 +7,6 @@ import re
 import sys
 import warnings
 
-import ctypes
 import numpy
 
 import cupy
@@ -1361,7 +1360,7 @@ cdef class ndarray:
                     and (self._f_contiguous or self._c_contiguous)):
                 order = 'F' if self._f_contiguous else 'C'
                 tmp = value.ravel(order)
-                ptr = ctypes.c_void_p(tmp.__array_interface__['data'][0])
+                ptr = tmp.ctypes.data
                 stream_ptr = stream_module.get_current_stream_ptr()
                 if stream_ptr == 0:
                     self.data.copy_from_host(ptr, self.nbytes)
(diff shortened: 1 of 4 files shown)


leofang:
Jenkins, test this please

emcastillo (approved):

leofang:
Thanks, @emcastillo!

PR #4554, opened 2021-01-15
Fix return type of `polynomial.__eq__`

asi1024:
Ready for https://github.com/cupy/cupy/pull/4262.

PR #4563, opened 2021-01-21
Fix return type of `polynomial.__eq__` to `numpy.bool_`

asi1024:
#4554

### DIFF
--- a/cupy/lib/_polynomial.pyx
+++ b/cupy/lib/_polynomial.pyx
@@ -193,7 +193,7 @@ cdef class poly1d:
             raise NotImplementedError
         if self.coeffs.shape != other.coeffs.shape:
             return False
-        return numpy.bool_((self.coeffs == other.coeffs).all().item())
+        return (self.coeffs == other.coeffs).all().get()[()]
 
     def __ne__(self, other):
         if not isinstance(other, poly1d):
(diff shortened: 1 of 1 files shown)


cupy/lib/_polynomial.pyx
```diff
@@ -193,7 +193,7 @@ cdef class poly1d:
             raise NotImplementedError
         if self.coeffs.shape != other.coeffs.shape:
             return False
-        return bool((self.coeffs == other.coeffs).all())
+        return numpy.bool_((self.coeffs == other.coeffs).all().item())
```
toslunar, line 196: I suppose we don't have to convert it via Python bool.
```suggestion
        return (self.coeffs == other.coeffs).all().get()[()]
```

toslunar (commented):
Should `polynomial.__eq__` return a CPU value?  `cupy.allclose` returns `cupy.ndarray` even though `numpy.allclose` returns Python `bool`.

asi1024:
I have two reasons for returning a CPU value as a result.
1. When the size of coeffs are different, it can return `False` as a CPU value immediately. I don't want to mix CPU and GPU values in return value depending on the sizes of inputs.
2. I think we don't need to pay so much attention to device-to-host synchronization in `cupy.poly1d`, because `cupy.poly1d` already requires many device-to-host synchronization mainly due to `trim_zeros` for coeffs.

toslunar (approved):
Then, LGTM

toslunar:
pfnCI, test this please.

PR #4564, opened 2021-01-21
Fix return type of `has_sorted_indices` of sparse array

asi1024:
```py
>>> import scipy.sparse
>>> scipy.sparse.random(3, 4).tocsc().has_sorted_indices
1
>>> import cupyx.scipy.sparse
>>> cupyx.scipy.sparse.random(3, 4).tocsc().has_sorted_indices
True
```

### DIFF
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
@@ -890,10 +890,19 @@ class TestCscMatrixScipyComparison(unittest.TestCase):
         assert 2 == len(M.indices)  # unaffected content
         return M
 
+    @testing.with_requires('scipy>1.6.0')
     @testing.numpy_cupy_equal(sp_name='sp')
     def test_has_sorted_indices(self, xp, sp):
         m = self.make(xp, sp, self.dtype)
         return m.has_sorted_indices
+
+    # TODO(asi1024): Remove test after the fixed version is released.
+    # https://github.com/scipy/scipy/pull/13426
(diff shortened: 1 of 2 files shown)


toslunar:
I posted scipy/scipy#13408.

asi1024:
@toslunar Should we add a workaround in the test code instead?

toslunar:
scipy/scipy#13408 is fixed :tada:

> Should we add a workaround in the test code instead?

Yes.  Even though the fixed SciPy hasn't been released, I'd edit the test rather than changing the behavior back and forth.

asi1024:
@toslunar PTAL!

tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
```diff
@@ -890,11 +890,20 @@ def test_has_canonical_format4(self, xp, sp):
         assert 2 == len(M.indices)  # unaffected content
         return M
 
+    @testing.with_requires('scipy>1.6.0')
```
toslunar, line 893: Maybe `scipy>=1.7`?

asi1024: Won't this SciPy's bug be fixed in 1.6.1?

toslunar: You are right.  The SciPy PR has the `backport-candidate` label.

toslunar:
pfnCI, test this please.

toslunar (approved):

PR #4566, opened 2021-01-21
Check if the accepted_error was not raised from test code

asi1024:
When using `testing.numpy_cupy_**` testing helpers, both NumPy and CuPy raise the same error type if the test script has a bug, for example, the wrong number of parameters.

This PR avoids passing such a wrong test unexpectedly.

Rel. https://github.com/cupy/cupy/pull/3702


kmaehashi:
pfnCI, test this please.

kmaehashi (approved):
LGTM!

PR #4567, opened 2021-01-23
Support stream capture

leofang:
This PR adds the minimal support to use CUDA graphs. CUDA graphs can be created in two ways:
1. stream capture: any work queued on the capturing stream becomes a node in the graph, and by ending capturing a graph is returned for launching (on any stream, which can be different from the one used to capture). This is what this PR supports.
2. graph API: the nodes of a graph can be built up using the new APIs, but [there're too many of them](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH) that I think is generally hard to use in a pythonic CuPy code.

A basic usage is like this:
```python
import cupy as cp

a = cp.random.randint(0, 10, 100, dtype=np.int32)
s = cp.cuda.Stream(non_blocking=True)

with s:
    s.begin_capture()
    a += 3
    a = cp.abs(a)
    g = s.end_capture()  # work is queued, but not yet launched
g.launch()
s.synchronize()

print(a)
```
Note that the capturing stream has to be non-blocking (thus it can't be the legacy default stream), otherwise a runtime error would be raised.


emcastillo:
Support for graph API?

leofang:
Right, but only very minimal ones, just enough for stream capturing to work. I have no interest (for now) to support the full graph APIs (for manually constructing the graphs). I'll explain further in the PR description.

I've done some quick tests and it seems to provide a little performance improvements as promised, but I need to check more on its robustness.

leofang:
This wouldn't work for certain CuPy functionalities such as `nonzero`/`argwhere`/etc, in which some synchronization calls are made. For example, this is a common idiom:
```python
if int(output[-1]) > 0:  # synchronize!
    # do something
```
Internally, if on a stream the copy is done asynchronously, so it can be captured (i.e. delayed). But the `int` conversion is a Python (host) operation that cannot be delayed, so the value could be incorrect. This is a very minimal reproducer:
```python
import cupy as cp
import numpy as np


a = cp.random.randint(0, 10, 100, dtype=np.int32)
buf = np.empty((1,), dtype=np.int32)
s = cp.cuda.Stream(non_blocking=True)

with s:
    s.begin_capture()
    a[-1].data.copy_to_host_async(buf.ctypes.data, 4, s)  # only async copy, no host operation
    last = int(a[-1])  # async copy + host operation 
    g = s.end_capture()
g.launch()
s.synchronize()

assert buf[0] == a[-1]  # this is ok
assert last == a[-1]  # this fails
```

I am looking into possible workarounds, such as wrapping all such host operations right after device-host transfers in `cudaLaunchHostFunc` (#4338) so that they are captured as part of the graph, but it doesn't seem trivial and may require a substantial amount of work. Another option is to delegate the burden to users, but I'd like to know we've exhausted all possible means before considering it.

leofang:
> I am looking into possible workarounds, such as wrapping all such host operations right after device-host transfers in `cudaLaunchHostFunc` (#4338) so that they are captured as part of the graph, but it doesn't seem trivial and may require a substantial amount of work.

I looked into it and it's not possible to work around using `cudaLaunchHostFunc`, as it explicitly forbid *any* CUDA API calls, not even the async ones.

leofang:
I took a cursory look at PyTorch's implementation and found that they only test simple stream capture cases as I did in this PR. @emcastillo Do you know if PyTorch has a strong guarantee that internally there wouldn't be any implicit synchronization so that they can safely adopt this feature, or they just don't care? 😂

leofang:
When testing stream capture with the SVD loop (#4628) I noticed cuBLAS and cuSOLVER do not play nicely with it...An error is raised at the end of capture. Even a trivial code like this fails:
```python
with s:
    s.begin_capture()
    c = cp.matmul(a, b)
    g = s.end_capture()
```
It looks like this works best only when simple operations or custom kernels are invoked, and should not be used with library calls (but, strangely, cuFFT works fine!)

leofang:
@pentschev @maxpkatz @anaruse Any chance you guys have insight on the interoperability between CUDA stream capture and CUDA libraries (see https://github.com/cupy/cupy/pull/4567#issuecomment-776895530 above)? 🙂

pentschev:
I don't know, but asked a question about that internally @leofang , will let you know when I hear back.

leofang:
Thanks, @pentschev. I feel I must have missed something obvious. At least [the cuBLAS doc](https://docs.nvidia.com/cuda/cublas/index.html#CUDA-graphs) says it's safe to capture streams. I will also keep looking into it...Would be great to have a C example to compare.

leofang:
1. Unlike cuBLAS, cuFFT, and cuSPARSE, the cuSOLVER doc has no mention of stream capture, so I suspect it just doesn't work. For the former 3 libraries, a subset of functionality work with stream capture.
2. After merging with master and rebuild this branch, the tiny `matmul` example above (which is based on `gemm` from cuBLAS) works with stream capture 🎉 So either I had a dirty build, or I messed up my branch somehow...

@pentschev Would be great to confirm No.1 😅

pentschev:
> Unlike cuBLAS, cuFFT, and cuSPARSE, the cuSOLVER doc has no mention of stream capture, so I suspect it just doesn't work. For the former 3 libraries, a subset of functionality work with stream capture.

That's correct, cuSOLVER, cuRAND and NPP do _not_ currently support stream capture, as confirmed by @mnicely . Other libraries may have only partial support for stream capture as well.

leofang:
Thanks for confirming, Peter!

pentschev:
From the docs, what I understand is that the latter will _not_ execute the graph. However, I agree it's not clear when it will be executed then, so I queried internally, will reply here when I hear back.

pentschev:
@leofang the reply I got is this:

> When you launch it with `cudaGraphLaunch()`, `cudaGraphUpload()` lets you do some of the work ahead of time.  This allows you to achieve faster launch but may be slower overall.  Some applications can take advantage of pipelining so the increased overhead is worth it.

I followed up to question whether you still have to execute it with `cudaGraphLaunch` after `cudaGraphUpload`, and the answer was positive.

leofang:
Ahh, thanks a lot @pentschev! So am I understanding you correctly that `cudaGraphUpload` is an optional step between the end of capture and `cudaGraphLaunch`?

pentschev:
That's my understanding as well, you can opt to just use `cudaGraphLaunch` by itself.

leofang:
@emcastillo @kmaehashi What are the thoughts from the team regarding this stream capture PR? Given the above discussions it's clear that it would only work with a subset of CUDA/CuPy features, but can still be useful if there're many custom Elementwise/Reduction/Raw Kernels in a workflow. Should I continue polishing it (likely with adding big runtime warnings and/or marking it experimental etc), or would you like to take more time for discussing the design?

johnzlli-tencent:
Is this feature still in developing？

leofang:
I can quickly resolve conflicts and prepare this PR such that it can be reviewed and merged, however given the above discussions it is clear that stream capturing is more limited than I expected and is thus not as powerful/intuitive as it should have been. For library/framework maintainers like the CuPy team, it is not easy to explain what can be captured and what can't, so I guess it's part of the reason we did not pursue it actively. 

@johnzlli-tencent would you mind sharing how you intended to use this feature?

johnzlli-tencent:
> I can quickly resolve conflicts and prepare this PR such that it can be reviewed and merged, however given the above discussions it is clear that stream capturing is more limited than I expected and is thus not as powerful/intuitive as it should have been. For library/framework maintainers like the CuPy team, it is not easy to explain what can be captured and what can't, so I guess it's part of the reason we did not pursue it actively.
> 
> @johnzlli-tencent would you mind sharing how you intended to use this feature?

We have many small operators writing in cupy. And we want to speed up the whole pipeline by multiple streams. But this idea is not work. Because our operators is too small, and the bottleneck seems to be the latency of cuda kernel launch. We made a test in c++ which use stream capture mechanism to define a graph and then launch the graph. In this case, even all kernels are still small, but there is significant improvement with multiple streams.

kmaehashi:
> What are the thoughts from the team regarding this stream capture PR? Given the above discussions it's clear that stream capture (or more generally CUDA graphs) would only work with a subset of CUDA/CuPy features, but can still be useful if there're many custom Elementwise/Reduction/Raw Kernels in a workflow.

I think we can introduce the stream capturing by fixing these functions to raise an exception if the current stream is capturing:
* memory copy is requested (both D2H and H2D, but not D2D)
* `cupy_backends.cuda.apis.*.setStream()` (for libraries not supporting CUDA graph)

leofang:
Thanks, @kmaehashi. Quick question:
> * `cupy_backends.cuda.apis.*.setStream()` (for libraries not supporting CUDA graph)

Is the purpose to make the stream-capture support more stringent than it should be (which I am fine with)? For example, what if some library APIs begin to support capture in a later CUDA release? I am asking because I noticed a previously failing example (https://github.com/cupy/cupy/pull/4567#issuecomment-776895530) is now passing (on CUDA 11.4). 

I doubt we wanna keep an updated list of (un)supported APIs, so I am fine with making only a limited subset of CuPy APIs supported. Just wanna make sure we are on the same page 🙂

leofang:
Let me do a test drive...

/test mini

kmaehashi:
> > * `cupy_backends.cuda.apis.*.setStream()` (for libraries not supporting CUDA graph)
> 
> Is the purpose to make the stream-capture support more stringent than it should be (which I am fine with)?

Yes, I'm thinking it is better to make it fail-safe.

> For example, what if some library APIs begin to support capture in a later CUDA release?

We can add a branch using e.g. `curandGetVersion()` after we confirm that the library supports CUDA graph.

> I doubt we wanna keep an updated list of (un)supported APIs, so I am fine with making only a limited subset of CuPy APIs supported.

Agree. It is difficult to list-up CuPy APIs supporting CUDA graph, because some routines may compute in a different ways based on conditions (e.g. routines may use cuTENSOR is used when enabled)

leofang:
The CUDA Python CI failures are irrelevant (they are due to the usage of the latest NumPy in the CI, which is 1.22 that just came out a few days ago).

leofang:
@kmaehashi @emcastillo Disabling stream capture for library calls is done in efe82d9. One minor note:

> > For example, what if some library APIs begin to support capture in a later CUDA release?
> 
> We can add a branch using e.g. `curandGetVersion()` after we confirm that the library supports CUDA graph.

I found the support conditions vary across libraries. For example, in cuBLAS one of the requirements involves the pointer mode being set to "device". In cuSPARSE, the routines supporting stream capture require the usage of cudaMallocAsync. In short, the checks are more involved than merely calling `*GetVersion()`. After this PR is merged I will create a meta issue to track and discuss how to support them.

leofang:
One more note: each `is_capturing()` call has an overhead ~0.5 us, not sure if it's acceptable...
```python
>>> with s:
...     s.begin_capture()
...     t1 = time.perf_counter()
...     for i in range(10000):
...             status = s.is_capturing()
...     t2 = time.perf_counter()
...     g = s.end_capture()
... 
>>> (t2-t1)/10000*1E9  # s -> ns
512.0798945426941
```

kmaehashi:
> One more note: each is_capturing() call has an overhead ~0.5 us, not sure if it's acceptable...(I think it's ok.)

I think it's ok. Maybe we can later add a fastpath if needed (skip calling `is_capturing` for `stream == 0 or stream == 2`)

emcastillo:
for some strange reason cuda python tests fail :)

leofang:
@emcastillo see https://github.com/cupy/cupy/pull/4567#issuecomment-1003783206.

leofang:
Thanks for help, @pentschev @emcastillo @kmaehashi!

cupy/cuda/stream.pyx
```diff
@@ -318,6 +319,103 @@ class _BaseStream:
...
+
+        Returns:
+            bool:
+                If the capturing status is successfully queried, the returned
+                value indicates the capturing status. An exception could be
+                raised if such a query is illegal, please refer to the CUDA
+                Programming Guide for detail.
+
+        """
+        # TODO(leofang): is it better to be a property?
```
leofang, line 411: Point for future discussion (cf. `Stream.done`).

PR #4552, opened 2021-01-13
ndimage: coordinate rounding fix for order=0 interpolation

grlee77:
Fix rounding behavior for `cupyx.scipy.ndimage.interpolation` functions with order=0.

closes #4550

reference line from SciPy source: 
https://github.com/scipy/scipy/blob/255d49cd63d4341c4d95b28ee8272657bf5e8e42/scipy/ndimage/src/ni_interpolation.c#L479

PR #4568, opened 2021-01-23
Fix cupyx.scipy.ndimage.zoom for outputs of size 1

grlee77:
This PR fixes a bug in `cupyx.scipy.ndimage.zoom` when the output has any dimension of size 1 and `grid_mode=True`. The purpose of the modified code section was just to prevent any potential divide by zero when calculating the zoom factors.

This PR is built on top of #4552, otherwise the order=0 test cases added here would fail.

### DIFF
--- a/cupyx/scipy/ndimage/interpolation.py
+++ b/cupyx/scipy/ndimage/interpolation.py
@@ -705,11 +705,10 @@ def zoom(input, zoom, output=None, order=3, mode='constant', cval=0.0,
 
         zoom = []
         for in_size, out_size in zip(input.shape, output_shape):
-            if out_size > 1:
-                if grid_mode:
-                    zoom.append(in_size / out_size)
-                else:
-                    zoom.append((in_size - 1) / (out_size - 1))
+            if grid_mode and out_size > 0:
+                zoom.append(in_size / out_size)
+            elif out_size > 1:
+                zoom.append((in_size - 1) / (out_size - 1))
             else:
                 zoom.append(0)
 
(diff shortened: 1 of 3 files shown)


emcastillo:
Jenkins, test this please

emcastillo (approved):

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
```diff
@@ -117,6 +117,29 @@ def test_map_coordinates_int(self, xp, scp, dtype):
         return out
 
 
+@testing.parameterize(*testing.product({
+    'order': [0, 1, 2, 3, 4, 5],
+    'mode': ['constant', 'nearest', 'mirror'] + scipy16_modes,
```
grlee77, line 122: probably have to drop the `scipy16_modes` here and the `_condition_scipy_version_skip` below for the 8.x branch

(and the same for `TestZoomOutputSize1` too)

grlee77: The comment above was intended to apply to a backport PR (if one is planned)

emcastillo:
v8 does not support grid_mode so this PR does not need to be backported.

cupyx/scipy/ndimage/interpolation.py
```diff
@@ -705,11 +705,10 @@ def zoom(input, zoom, output=None, order=3, mode='constant', cval=0.0,
 
         zoom = []
         for in_size, out_size in zip(input.shape, output_shape):
-            if out_size > 1:
-                if grid_mode:
-                    zoom.append(in_size / out_size)
-                else:
-                    zoom.append((in_size - 1) / (out_size - 1))
+            if grid_mode and out_size > 0:
```
hadipash, line 708: Why is it necessary to specify `out_size > 0` if technically any output size should be higher than 0?

PR #4574, opened 2021-01-27
Fix windows build

emcastillo:
`ssize_t` was in the .cu file instead of the .cuh


kmaehashi:
pfnCI, test this please.

PR #4594, opened 2021-02-01
Fix device property default name

leofang:
This bug shadowed another bug being investigated: https://github.com/conda-forge/cupy-feedstock/issues/98.


kmaehashi:
pfnCI, test this please.

kmaehashi:
Test failures are unrleated.

PR #4605, opened 2021-02-02
Stop using deprecated type aliases

toslunar:
`np.float`, `np.int`, `np.bool` are deprecated since NumPy 1.20.


kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

PR #4619, opened 2021-02-04
Always run cythonize on sdist installation

### LINKED ISSUE #4597: BUG: CUDA_VERSION is set to 0 in sdist
I was investigating the bug reported in https://github.com/conda-forge/cupy-feedstock/issues/98, and it turns out the reason is on Conda-Forge we built CuPy from sdist, but the source has `CUDA_VERSION` inlined to 0:
```c++
// cupy_backends/cuda/api/driver.cpp

/* ... */

static PyObject *__pyx_pf_13cupy_backends_4cuda_3api_6driver_2get_build_version(CYTHON_UNUSED PyObject *__pyx_self) {
  PyObject *__pyx_r = NULL;
  __Pyx_RefNannyDeclarations
  __Pyx_RefNannySetupContext("get_build_version", 0);

  /* "cupy_backends/cuda/api/driver.pyx":141
 * 
 * def get_build_version():
 *     return CUDA_VERSION             # <<<<<<<<<<<<<<
 * 
 * 
 */
  __Pyx_XDECREF(__pyx_r);
  __Pyx_INCREF(__pyx_int_0);
  __pyx_r = __pyx_int_0;
  goto __pyx_L0;

  /* "cupy_backends/cuda/api/driver.pyx":140
 * ###############################################################################
 * 
 * def get_build_version():             # <<<<<<<<<<<<<<
 *     return CUDA_VERSION
 * 
 */

  /* function exit code */
  __pyx_L0:;
  __Pyx_XGIVEREF(__pyx_r);
  __Pyx_RefNannyFinishContext();
  return __pyx_r;
}

/* ... */

static CYTHON_SMALL_CODE int __Pyx_InitGlobals(void) {
  if (__Pyx_InitStrings(__pyx_string_tab) < 0) __PYX_ERR(0, 1, __pyx_L1_error);
  __pyx_int_0 = PyInt_FromLong(0); if (unlikely(!__pyx_int_0)) __PYX_ERR(0, 1, __pyx_L1_error)
  return 0;
  __pyx_L1_error:;
  return -1;
}
```
I don't have a good solution or patch for now, as it seems to imply sdist should be generated per CUDA version. Perhaps I am overlooking some obvious fix?

cc: @jakirkham @kmaehashi

kmaehashi:
This PR removes C++ sources from sdist and changes the CuPy installer to always run cythonize at user-side.
Cython will automatically (temporarily) be downloaded by setuptools via `setup_requires` configuration.

The primary motivation for this is to use compile-time constants in the code appropriately (closes #4597).
Besides that, cythonize at user-side  is also needed for #4395 which generates pyx during the installation.

### DIFF
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -13,7 +13,6 @@ import sys
 import pkg_resources
 import setuptools
 from setuptools.command import build_ext
-from setuptools.command import sdist
 
 from install import build
 from install.build import PLATFORM_LINUX
@@ -1092,18 +1091,6 @@ class _MSVCCompiler(msvccompiler.MSVCCompiler):
 
         # Return compiled object filenames.
         return other_objects + cu_objects
-
-
-class sdist_with_cython(sdist.sdist):
-
-    """Custom `sdist` command with cyhonizing."""
-
-    def __init__(self, *args, **kwargs):
-        if not cython_available:
-            raise RuntimeError('Cython is required to make sdist.')
-        ext_modules = get_ext_modules(True)  # get .pyx modules
-        cythonize(ext_modules, cupy_setup_options)
-        sdist.sdist.__init__(self, *args, **kwargs)
 
 
 class custom_build_ext(build_ext.build_ext):
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -807,7 +807,8 @@ def cythonize(extensions, arg_options):
     cython_version = pkg_resources.parse_version(Cython.__version__)
     if (cython_version < required_cython_version or
             cython_version in ignore_cython_versions):
-        raise AssertionError('Unsupported Cython version: {}'.format(cython_version))
+        raise AssertionError(
+            'Unsupported Cython version: {}'.format(cython_version))
 
     directive_keys = ('linetrace', 'profile')
     directives = {key: arg_options[key] for key in directive_keys}
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -807,7 +807,7 @@ def cythonize(extensions, arg_options):
     cython_version = pkg_resources.parse_version(Cython.__version__)
     if (cython_version < required_cython_version or
             cython_version in ignore_cython_versions):
-        raise AssertionError('Unsupported Cython version')
+        raise AssertionError('Unsupported Cython version: {}'.format(cython_version))
 
     directive_keys = ('linetrace', 'profile')
     directives = {key: arg_options[key] for key in directive_keys}
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -19,6 +19,8 @@ from install.build import PLATFORM_LINUX
 from install.build import PLATFORM_WIN32
 
 
+# Cython requirements (minimum version and versions known to be broken).
+# Note: this must be in sync with setup_requires defined in setup.py.
 required_cython_version = pkg_resources.parse_version('0.28.0')
 ignore_cython_versions = [
 ]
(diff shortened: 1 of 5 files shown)


kmaehashi:
Ready for review, updated the description.
pfnCI, test this please.

cupy_setup_build.py
```diff
@@ -801,18 +795,16 @@ def prepare_wheel_libs():
...
     cython_version = pkg_resources.parse_version(Cython.__version__)
-    cython_available = (
-        cython_version >= required_cython_version and
-        cython_version not in ignore_cython_versions)
-except ImportError:
-    cython_available = False
-
+    if (cython_version < required_cython_version or
+            cython_version in ignore_cython_versions):
+        raise AssertionError('Unsupported Cython version')
```
leofang, line 806: nitpick:
```suggestion
        raise AssertionError('Unsupported Cython version: {}'.format(cython_version))
```

cupy_setup_build.py
```diff
@@ -364,14 +365,7 @@ def module_extension_name(file):
...
     base = path.join(*pyx.split('.'))
-    if use_cython:
-        pyx = base + '.pyx'
-        if not os.path.exists(pyx):
-            use_cython = False
-            print(
-                'NOTICE: Skipping cythonize as {} does not exist.'.format(pyx))
-    if not use_cython:
-        pyx = base + '.cpp'
+    pyx = base + ('.pyx' if use_cython else '.cpp')
```
leofang, line 368: I thought `.cpp` is removed?

kmaehashi: This is a hacky part (it's better to do refactoring). When doing `pip install cupy`:
1. setup.py: use `use_cython=False` to create a list of extension CPP sources generated by Cython (`*.cpp`)
2. setup.py: pass it to `setuptools.setup()` (at the bottom of the script)
3. setuptools: install Cython
5. cupy_setup_build.py: use `use_cython=True` to create a list of extension PYX sources to list up pyx files and  cythonize them (e.g., generate `core.cpp` from `core.pyx`).
6. setuptools: build the resulting CPP files, using the list generated in 1.

So `use_cython` is still needed here.

leofang:
Jenkins, test this please

jakirkham:
Instead of using `setup_requires` what do you think about using `pyproject.toml` from [PEP 517]( https://www.python.org/dev/peps/pep-0517/ )? This fixed some warts with `setup_requires` IIRC

kmaehashi:
@jakirkham Thanks for the nice suggestion! Moved the dependencies to `pyproject.yaml`.

We should be aware that pip v19.0 (released on 2019-01-22) is the first version that supports `pyproject.yaml`.
I believe most users are OK (we suggest upgrading pip in the installation guide), but we can add `setup_requires` to `setup.py` as a fallback for old pip users anytime.

kmaehashi:
pfnCI, test this please.

jakirkham:
That's a good point Kenichi 🙂

It looks like Setuptools version [36.6.0]( https://github.com/pypa/setuptools/blob/e1ffc2abbae4f2aa78dd09ee9827d754b7702b7b/CHANGES.rst#v3660 ) first added support for PEP 517. We could just check if Setuptools is older than that. If it is, we can add `setup_requires`.

kmaehashi:
I see, so should we check versions of both setuptools and pip?

I think it's also ok to always specify `setup_requires` if we have concern 😃  There's no harm for new pip users.

kmaehashi:
pfnCI, test this please.

kmaehashi:
Blocked by https://github.com/chainer/chainer-test/pull/634

takagi:
The change looks good to me! Waiting for the blocker merged.

emcastillo:
Jenkins, test this please

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
@kmaehashi will this be backported for fixing #4597 in v8?

emcastillo:
let me backport it

PR #3247, opened 2020-04-01
Support cusolverDn<t>gesvdjBatched

toslunar:
Rework of #2337 on #3192

- The second commit is optional.
- The third commit can be done in a separate PR.

Issue #3062, opened 2020-02-11
cupy.linalg.{svd, pinv} should support broadcasting


carterbox:
Problem
-----------
The NumPy implementation of [pinv](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.pinv.html) supports broadcasting, but the CuPy implementation does not.

Conditions
--------------
```
CuPy Version          : 7.0.0
CUDA Root             : /local/cuda-10.0
CUDA Build Version    : 10010
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10010
cuBLAS Version        : 10201
cuFFT Version         : 10101
cuRAND Version        : 10101
cuSOLVER Version      : (10, 2, 0)
cuSPARSE Version      : 10300
NVRTC Version         : (10, 1)
cuDNN Build Version   : 7604
cuDNN Version         : 7605
NCCL Build Version    : 2406
NCCL Runtime Version  : 2506
```
Reproduce
--------------
```python
import numpy as np
import cupy as cp

A = np.random.rand(3, 4, 5)
xA = cp.array(A)
np.linalg.pinv(A)
cp.linalg.pinv(xA)
```
```
Traceback (most recent call last):
  File "pinv.py", line 7, in <module>
    cp.linalg.pinv(xA)
  File "/home/beams/DCHING/miniconda3/envs/tike/lib/python3.7/site-packages/cupy/linalg/solve.py", line 421, in pinv
    u, s, vt = decomposition.svd(a.conj(), full_matrices=False)
  File "/home/beams/DCHING/miniconda3/envs/tike/lib/python3.7/site-packages/cupy/linalg/decomposition.py", line 245, in svd
    util._assert_rank2(a)
  File "/home/beams/DCHING/miniconda3/envs/tike/lib/python3.7/site-packages/cupy/linalg/util.py", line 20, in _assert_rank2
    'two-dimensional'.format(a.ndim))
numpy.linalg.LinAlgError: 3-dimensional array given. Array must be two-dimensional
```
Solution
-----------
Change CuPy implementation to allow broadcasting on N+2 dimensional arrays. This may also require updating `cupy.linalg.svd` because `pinv` relies on `svd` and NumPy allows broadcasting for that function too.

carterbox:
Broadcasting has been enabled for some `linalg` functions since [`numpy=1.8`](https://numpy.org/doc/1.18/reference/routines.linalg.html#module-numpy.linalg)

carterbox:
[...]

PR #2337, opened 2019-07-25
Support batched SVD

toslunar:

PR #4628, opened 2021-02-05
Support batched SVD

### LINKED ISSUE #3470: Support batched SVD (`cupy.linalg.svd`)
https://github.com/cupy/cupy/pull/2337#issuecomment-630577091
> It is supposed that now batched cupy.cusolver.gesvdj that uses cusolverDn<t>gesvdjBatched is merged in #3247 and remaining is to fix cupy.linalg.svd to support batched svd using cupy.cusolver layer.

leofang:
Close #3470. Part of #3062. Another shot based on #3247 ~~to avoid manual looping.~~ 
**UPDATE**: See the below discussions in this PR.

Prepare for adopting Array API (https://github.com/data-apis/array-api/pull/114).

Code path divergence:
- On CUDA: Call gesvdj if m,n <= 32, else call gesvd in a C loop
- On ROCm: Call gesvd as it has no limitation in the input size

TODO:
- [x] Support integers
- [x] Support zero-size (empty) input
- [x] Docstring
- [x] Test on HIP
- [x] Benchmark


leofang:
...`*gesvdjBatched()` only support a stack of matrices with dimension < (32, 32) on either side... @toslunar Looks like we have to do a manual looping anyway for general cases?

leofang:
I should have read carefully the discussion in #2337 first 😅 Looks like I was duplicating the work...

leofang:
> ...`*gesvdjBatched()` only support a stack of matrices with dimension < (32, 32) on either side... 

I am testing `rocsolver_<t>gesvd_batched` and it doesn't seem to have this limitation, so I'll try to test & support it in this PR.

cupy/cusolver.py
```diff
@@ -190,17 +200,19 @@ def _gesvdj_batched(a, full_matrices, compute_uv, overwrite_a):
     u = _cupy.empty((batch_size, m, ldu), dtype=a.dtype).swapaxes(-2, -1)
     v = _cupy.empty((batch_size, n, ldv), dtype=a.dtype).swapaxes(-2, -1)
     params = _cusolver.createGesvdjInfo()
-    lwork = helper(handle, jobz, m, n, a.data.ptr, lda, s.data.ptr,
+    lwork = helper(handle, jobz, m, n, ap.data.ptr, lda, s.data.ptr,
                    u.data.ptr, ldu, v.data.ptr, ldv, params, batch_size)
     work = _cupy.empty(lwork, dtype=a.dtype)
-    info = _cupy.empty(1, dtype=_numpy.int32)
-    solver(handle, jobz, m, n, a.data.ptr, lda, s.data.ptr,
+    info = _cupy.empty(batch_size, dtype=_numpy.int32)
```
leofang, line 206: This is a bug: both cuSOLVER and rocSOLVER need the `info` array to be of the batch size. I thought we fixed it earlier...?

leofang:
Everything works on the HIP side, but I noticed something odd: If I run the test it hangs at `test_svd_rank4*` (with all other tests pass), but if I filter out the dtype and run them one by one, nothing would go wrong. I can't reproduce this outside the test environment. Still looking into it...

leofang:
Another oddity is on HIP we have to call `rocsolver_<t>gesvd_batched`. Manual looping over `rocsolver_<t>gesvd` is much slower, and, with a bit larger batch size (~64?), it takes forever to run. Not sure if it's related to the hanging issue mentioned above or not.

leofang:
On CUDA: the manual looping follows what's done in JAX: https://github.com/google/jax/blob/8bf3f032989caddf05c702acc1ff5353abbe72d2/jaxlib/cusolver.cc#L937-L947
It is made possible here by cythonizing `cupy/cusolver.py` so that we can speed up the loop using Cython's constructs. But the caveat here is that by calling the `<t>gesvd()` wrapper we release and acquire the gil in every iteration, which isn't optimal.

leofang:
> It is made possible here by cythonizing `cupy/cusolver.py` so that we can speed up the loop using Cython's constructs. 

btw The loop is implemented in (cythonized) `cupy/cusolver.py` instead of `cupy_backends/cuda/libs/cusolver.pyx` because I don't wanna interfere with @takagi's FFI works (#4395/#4606/#4427). My understanding is all lib wrappers there will be autogen'd, so it doesn't seem wise to add a custom loop there.

leofang:
(Test run)

Jenkins, test this please

leofang:
Cannot locally reproduce the failed SVD tests, but other tests should be fixed now.

Jenkins, test this please

leofang:
>> Everything works on the HIP side, but I noticed something odd: If I run the test it hangs at `test_svd_rank4*` (with all other tests pass), but if I filter out the dtype and run them one by one, nothing would go wrong. I can't reproduce this outside the test environment. Still looking into it...
>
> Another oddity is on HIP we have to call `rocsolver_<t>gesvd_batched`. Manual looping over `rocsolver_<t>gesvd` is much slower, and, with a bit larger batch size (~64?), it takes forever to run. Not sure if it's related to the hanging issue mentioned above or not.

Turns out both issues have the same root cause: Looping in Python (which was done for comparing the result in the tests) is awfully slow on HIP for some reason. It looks like hanging because the loop is too slow; when checking with `rocm-smi` it shows 100% GPU usage rate for over an hour 😂 The resolution was to reduce the test problem size so it could complete in a reasonable time on HIP (entire `test_decomposition.py` now takes ~2 mins).

leofang:
> > It is made possible here by cythonizing `cupy/cusolver.py` so that we can speed up the loop using Cython's constructs. But the caveat here is that by calling the <t>gesvd() wrapper we release and acquire the gil in every iteration, which isn't optimal.
> 
> btw The loop is implemented in (cythonized) `cupy/cusolver.py` instead of `cupy_backends/cuda/libs/cusolver.pyx` because I don't wanna interfere with @takagi's FFI works (#4395/#4606/#4427). My understanding is all lib wrappers there will be autogen'd, so it doesn't seem wise to add a custom loop there.

This C loop is moved to `cupy_lapack.h` so that 1. gil can be released throughout the entire loop, 2. code duplication is minimized by using C++ template + type traits to select which cuSOLVER API to call. On the Cython side, given it's now "type  neutral" one can just use a function pointer. I think after a few iterations this is a much better design now 🙂

leofang:
Just a very quick benchmark: Compared to a naive Python implementation (looping over `cp.linalg.svd`), the C-loop-based batched SVD is about 2-10 ms faster per call, translating to an overhead ~50 us per matrix.  
```
svd                 :    CPU:955688.987 us   +/-286.070 (min:955402.440 / max:956859.979) us     GPU-0:955817.447 us   +/-284.945 (min:955536.133 / max:956985.352) us
loop_svd            :    CPU:965675.944 us   +/-407.225 (min:965178.026 / max:966321.092) us     GPU-0:965688.611 us   +/-407.043 (min:965195.496 / max:966334.534) us
cpu time diff: 9986.957034561761 us, gpu time diff: 9871.163940429862 us
```
```python
import cupy as cp
from cupyx.time import repeat


#shape = (64, 128, 128)
shape = (256, 64, 64)
a = cp.random.random(shape)

def loop_svd(x, compute_uv=True):
    batch_size = x.shape[0]
    u_t = []
    s_t = []
    v_t = []
    for i in range(batch_size):
        out = cp.linalg.svd(x[i], compute_uv=compute_uv)
        if compute_uv:
            u, s, v = out
            u_t.append(u)
            s_t.append(s)
            v_t.append(v)
        else:
            s_t.append(out)
    if compute_uv:
        return cp.stack(u_t, axis=0), cp.stack(s_t, axis=0), cp.stack(v_t, axis=0)
    else:
        return cp.stack(s_t, axis=0)

svd = repeat(cp.linalg.svd, (a,), n_repeat=20)
naive_svd = repeat(loop_svd, (a,), n_repeat=20)
print(svd)
print(naive_svd)
print(f'cpu time diff: {(naive_svd.cpu_times.mean() - svd.cpu_times.mean()) * 1000000} us, '
      f'gpu time diff: {(naive_svd.gpu_times.mean() - svd.gpu_times.mean()) * 1000000} us')
```

leofang:
Jenkins, test this please

leofang:
@toslunar Looks like there's some randomness in the test failures -- each run has different tests failing. Do you have suggestions for how to locally reproduce the failures deterministically?

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
I run `cupy-py3` locally through `cuda-memcheck`. Looks like a cuSOLVER bug to me:
```
$ cuda-memcheck python -m pytest tests/cupy_tests/linalg_tests/test_decomposition.py -k svd

(...)

.............========= Invalid __global__ write of size 16
=========     at 0x00000060 in void larf_minus_tau<double2>(double2 const *, double2*)
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x7f520fb4d200 is out of bounds
=========     Device Frame:void larf_minus_tau<double2>(double2 const *, double2*) (void larf_minus_tau<double2>(double2 const *, double2*) : 0x60)
=========     Host Frame:/usr/lib/x86_64-linux-gnu/libcuda.so.1 (cuLaunchKernel + 0x2b8) [0x222dc8]
=========     Host Frame:/usr/local/cuda/lib64/libcusolver.so.10.0 [0x100c822]
=========     Host Frame:/usr/local/cuda/lib64/libcusolver.so.10.0 [0x100ca17]
=========     Host Frame:/usr/local/cuda/lib64/libcusolver.so.10.0 [0x1040dd5]
=========     Host Frame:/usr/local/cuda/lib64/libcusolver.so.10.0 [0x1a460f]
=========     Host Frame:/usr/local/cuda/lib64/libcusolver.so.10.0 [0x1b4476]
=========     Host Frame:/usr/local/cuda/lib64/libcusolver.so.10.0 [0x181cfc]
=========     Host Frame:/usr/local/cuda/lib64/libcusolver.so.10.0 (cusolverDnZgesvd + 0x6ec) [0x18523c]
=========     Host Frame:/work/cupy/cupy/cusolver.cpython-38-x86_64-linux-gnu.so (_Z10gesvd_loopI7double2Eilcciilillililili + 0x12c) [0x62e6c]
=========     Host Frame:/work/cupy/cupy/cusolver.cpython-38-x86_64-linux-gnu.so [0x5f1ee]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (PyCFunction_Call + 0xf5) [0xa2525]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyObject_MakeTpCall + 0x9e) [0x9f7fe]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x6d43) [0x74063]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x6549) [0x73869]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x5d3d) [0x7305d]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (PyVectorcall_Call + 0x6f) [0xa200f]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x1609) [0x6e929]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x5d87) [0x730a7]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 [0x6c0ff]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (PyVectorcall_Call + 0x6f) [0xa200f]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x1609) [0x6e929]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x6ca6) [0x73fc6]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 [0x6c0ff]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 [0xa3c45]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x6549) [0x73869]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 [0x6c0ff]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x5d87) [0x730a7]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 [0xa3b37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (PyVectorcall_Call + 0x6f) [0xa200f]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x1609) [0x6e929]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyObject_FastCallDict + 0x55) [0xa1d95]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyObject_Call_Prepend + 0xcc) [0xa1f6c]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 [0x11c169]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyObject_MakeTpCall + 0x9e) [0x9f7fe]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalFrameDefault + 0x6d43) [0x74063]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyEval_EvalCodeWithName + 0xb67) [0x1a0f37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (_PyFunction_Vectorcall + 0x92) [0x9f602]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 [0xa3b37]
=========     Host Frame:/opt/pyenv/versions/3.8.0/lib/libpython3.8.so.1.0 (PyVectorcall_Call + 0x6f) [0xa200f]
=========
FFFFFFFFFFFFFFFFFFF                                                                            [100%]

============================================================================== FAILURES ===============================================================================
______________________________________________________ TestSVD_param_0_{full_matrices=True}.test_svd_rank4_loop _______________________________________________________
cupy/testing/condition.py:79: in wrapper
    fail()
cupy/testing/condition.py:50: in fail
    instance.fail(msg)
E   AssertionError: 
E   Fail: 1, Success: 0
E   
E   The first error message:
E     File "/work/cupy/cupy/testing/condition.py", line 59, in <lambda>
E       lambda: f(ins, *args[1:], **kwargs),
E     File "/work/cupy/tests/cupy_tests/linalg_tests/test_decomposition.py", line 310, in test_svd_rank4_loop
E       self.check_usv((3, 2, 64, 64))
E     File "/work/cupy/cupy/testing/helper.py", line 825, in test_func
E       impl(self, *args, **kw)
E     File "/work/cupy/tests/cupy_tests/linalg_tests/test_decomposition.py", line 150, in check_usv
E       result_gpu = cupy.linalg.svd(a_gpu, full_matrices=self.full_matrices)
E     File "/work/cupy/cupy/linalg/_decomposition.py", line 465, in svd
E       return _svd_batched(a, a_dtype, full_matrices, compute_uv)
E     File "/work/cupy/cupy/linalg/_decomposition.py", line 401, in _svd_batched
E       out = _gesvd_batched(a, a_dtype, full_matrices, compute_uv, False)
E     File "cupy/cusolver.pyx", line 262, in cupy.cusolver._gesvd_batched
E       cpdef _gesvd_batched(a, a_dtype, full_matrices, compute_uv, overwrite_a):
E     File "cupy/cusolver.pyx", line 351, in cupy.cusolver._gesvd_batched
E       raise _cusolver.CUSOLVERError(status)
E   cupy_backends.cuda.libs.cusolver.CUSOLVERError: CUSOLVER_STATUS_INTERNAL_ERROR
```
I can also reproduce some test failures, which are not exactly identical to those seen in the CI but also fail at gesvd loop tests.

leofang:
Looks like the failing kernel `larf_minus_tau` is one of the known bugs, but the documentation said `gesvd` shouldn't be affected, not sure what's wrong...
> This bug affects cuSOLVER dense linear algebra functionality in CUDA 10.0, and is fixed in CUDA 10.1. An internal routine (LARFT) in cuSOLVER dense linear algebra has a race condition if the user stream is not null. Although this routine is not in the cuSOLVER public API, it affects all routines based on householder reflection, including ORGBR, ORMTR, ORGTR, ORMQR, ORMQL, ORGQR AND SYEVD, SYGVD.
The following routines are not affected:
GEQRF and GESVD.
Also, GESVDJ and SYEVDJ are not affected because they are based-on the Jacobi method.

https://docs.nvidia.com/cuda/archive/10.1/cuda-toolkit-release-notes/index.html#unique_1334956937

Let me try skipping the test on CUDA 10.0.

leofang:
Jenkins, test this please

emcastillo:
@anaruse can you please take a look at this? 😇

cupy/linalg/_decomposition.py
```diff
@@ -352,6 +354,63 @@ def qr(a, mode='reduced'):
...
+                vt = cupy.empty(batch_shape + (k, m), dtype=a_dtype)
+            return u, s, vt
+        else:
+            return s
+    elif m == 0 or n == 0:
+        s = cupy.empty(batch_shape + (0,), s_dtype)
+        if compute_uv:
+            if full_matrices:
+                u = [cupy.eye(n, dtype=a_dtype) for _ in range(batch_size)]
+                u = cupy.stack(u, axis=0).reshape(*batch_shape, n, n)
```
toslunar, line 382: How about
```suggestion
                u = cupy.ascontiguousarray(cupy.broadcast_to(
                    cupy.identity(n, a_dtype),
                    batch_shape + (n, n)))
```
or
```suggestion
                u = cupy.empty(batch_shape + (n, n), a_dtype)
                u[...] = cupy.identity(n, a_dtype)
```

My thought on other possible implementations:

- `cupy.stack([cupy.eye(n, dtype=a_dtype)] * batch_size, axis=0).reshape(*batch_shape, n, n)` reduces the number of calls of `eye`, but still creates a list.
- `cupy.eye(n, dtype=a_dtype)[None].repeat(batch_size, axis=0).reshape(*batch_shape, n, n)` might not be fast because `repeat` supports
  ```
  >>> cupy.repeat(cupy.arange(3), [3, 2, 1])
  array([0, 0, 0, 1, 1, 2])
  ```
- `cupy.tile(cupy.eye(n, dtype=a_dtype)[None], (batch_size, 1, 1)).reshape(*batch_shape, n, n)` is somewhat verbose.

toslunar (changes requested):

leofang:
Check again 

Jenkins, test this please

cupy_backends/cupy_lapack.h
```diff
@@ -14,4 +16,61 @@
...
+    for (int i=0; i<batch_size; i++) {
+        // setting rwork to NULL as we don't need it
+        status = func(
+            reinterpret_cast<cusolverDnHandle_t>(handle), jobu, jobvt, m, n, A, lda,
+            S, U, ldu, VT, ldvt, Work, buffersize, NULL, devInfo);
+        if (status != 0) break;
+        A += m * n;
+        S += k;
+        U += m * m;
+        VT += n * n;
```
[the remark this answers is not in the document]
leofang: Yes, when the wrapper is called this is ensured. How about removing `lda` etc front he signature and just using plain `m`, `n`?

cupy/linalg/_decomposition.py
leofang: I think your second suggestion of using `cupy.identity` seems nice, I'll test it locally. Note that this is a very special (that is, silly) call path that in reality would be unlikely to hit, and I don't think performance matters here...

leofang: Updated in 5a0376a.

cupy/linalg/_decomposition.py
```diff
@@ -352,6 +354,63 @@ def qr(a, mode='reduced'):
...
+    if runtime.is_hip or (m <= 32 and n <= 32):
+        # copy is done in _gesvdj_batched, so let's try not to do it here
+        a = a.astype(a_dtype, order='C', copy=False)
+        out = _gesvdj_batched(a, full_matrices, compute_uv, False)
+    else:
+        # manually loop over cusolverDn<t>gesvd()
+        # copy (via possible type casting) is done in _gesvd_batched
+        out = _gesvd_batched(a, a_dtype, full_matrices, compute_uv, False)
+    if compute_uv:
+        u, s, v = out
```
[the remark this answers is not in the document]
leofang: Good catch, @toslunar! There was actually a bug related to this, which I fixed in 62e11ac. Yes, it returns `V` instead of `V^H` (transpose conjugate).

cupy_backends/cupy_lapack.h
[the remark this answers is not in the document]
leofang: > How about removing `lda` etc from the signature and just using plain `m`, `n`?

Done in ce32d03.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

toslunar (approved):
LGTM

toslunar:
The failed test on Jenkins is known to be flaky (#4673).

leofang:
Thanks, @toslunar!

cupy/cusolver.py
leofang: Backported this fix to v8 (#4747).

PR #4631, opened 2021-02-05
Update install requirements

kmaehashi:
- Update NumPy requirement. (Actually this had to be done when we decided to drop NumPy 1.16 or earlier in v9 releases)
- Add `all` extra specifier to install all optional dependencies

c.f. https://docs.cupy.dev/en/latest/install.html#python-dependencies

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,9 +28,14 @@ requirements = {
         'fastrlock>=0.3',
     ],
     'install': [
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, test this please.

asi1024:
LGTM.

PR #4636, opened 2021-02-06
Bump cuDNN to v8.1.0

takagi:


cupyx/tools/install_library.py
```diff
@@ -50,17 +50,17 @@ def _make_cudnn_record(
...
+    '11.0', '8.1.0',
+    'cudnn-11.2-linux-x64-v8.1.0.77.tgz',
+    'cudnn-11.2-windows-x64-v8.1.0.77.zip'))
 _cudnn_records.append(_make_cudnn_record(
-    '10.2', '8.0.5',
-    'cudnn-10.2-linux-x64-v8.0.5.39.tgz',
-    'cudnn-10.2-windows10-x64-v8.0.5.39.zip'))
+    '10.2', '8.1.0',
+    'cudnn-10.2-linux-x64-v8.1.0.77.tgz',
+    'cudnn-10.2-windows-x64-v8.1.0.77.zip'))
```
kmaehashi, line 63: The filename seems`cudnn-10.2-windows10-x64-v8.1.0.77.zip`?

takagi: Fixed, thanks!

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

install/build.py
```diff
@@ -15,7 +15,7 @@
 
 minimum_cuda_version = 9020
 minimum_cudnn_version = 7600
-maximum_cudnn_version = 8099
+maximum_cudnn_version = 8199
```
kmaehashi, line 18: It's ok to proceed to merge, but I think we can reconsider whether this check is still needed or not.

takagi: Let me check why the check is/was needed.

takagi: You mean only for maximum, or both minimum and maximum?

kmaehashi: I meant maximum only. We don't check max version for CUDA, so I felt it is unnatural to limit cudnn.

takagi: Makes sense. I will make another PR to remove it!

PR #4271, opened 2020-11-12
ROCm: Fix filters in `cupyx.scipy.ndimage`

leofang:
Close #4268. Close #4270.

Two changes:
1. Add `cupy/math_constants.h` for both CUDA and HIP
2. Convert a `__global__` RawKernel to a `__device__` function in the ndimage generic filters.

This simple fix makes a handful of tests in `tests/cupyx_tests/scipy_tests/ndimage_tests/test_filters.py` pass (ROCm 3.5.0 + Radeon VII). The failing ones are irrelevant of #4268 or #4270, and will be documented in a separate issue (#4272).

PR #4565, opened 2021-01-21
Remove dependency on CUDA's math_constants.h

leofang:
Fix #4414.

This PR makes CuPy's own copy of `math_constants.h` a standalone header, so that users who are running CuPy without a full CUDA installation (ex: Conda-Forge or CUDA runtime docker) can still JIT compile codes. Unneeded constants are removed to avoid potential EULA violation 😞

cc: @grlee77

PR #4642, opened 2021-02-08
[backport] ROCm: Fix filters in `cupyx.scipy.ndimage` - Part 1

kmaehashi:
Backport of #4271

### DIFF
--- a/cupy/core/include/cupy/math_constants.h
+++ b/cupy/core/include/cupy/math_constants.h
@@ -1,0 +1,110 @@
+#ifndef CUPY_MATH_CONSTANTS_H
+#define CUPY_MATH_CONSTANTS_H
+
+#ifdef __CUDA_ARCH__
+
+#include <math_constants.h>
+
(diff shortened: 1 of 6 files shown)


kmaehashi:

cupy/core/include/cupy/math_constants.h
```diff
@@ -0,0 +1,110 @@
+#ifndef CUPY_MATH_CONSTANTS_H
```
grlee77, line 1: The backport should probably use the more minimal version of this file that was introduced for #4565. 

#4565 also helped avoid a patch during the conda package build

emcastillo: Yeah, we need to backport that one first :)

leofang: Good catch, @grlee77!

leofang: Perhaps this PR can be backported first, and then followed by backporting #4565?

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4633, opened 2021-02-05
Use `CodeBlock` to generate human-readable code in JIT

asi1024:
This PR fixes JIT code with `CodeBlock` object, which we implemented in CuPy fusion to generate human-readable CUDA code.

PR #4646, opened 2021-02-09
Support if-statement in CuPy JIT

asi1024:
Blocked by #4633.

### DIFF
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -185,15 +185,16 @@ def _call_ufunc(ufunc, args, dtype, env):
     if len(args) != ufunc.nin:
         raise ValueError('invalid number of arguments')
 
-    args = [_to_cuda_object(x, env) for x in args]
-
+    in_types = []
     for x in args:
-        if not isinstance(x.ctype, _types.Scalar):
-            raise NotImplementedError
+        if is_constants([x]):
+            t = _typerules.get_ctype_from_scalar(env.mode, x.obj).dtype
+        else:
+            t = x.ctype.dtype
+        in_types.append(t)
 
-    in_types = tuple([x.ctype.dtype for x in args])
     if dtype is None:
-        op = ufunc._ops._guess_routine_from_in_types(in_types)
+        op = ufunc._ops._guess_routine_from_in_types(tuple(in_types))
     else:
         op = ufunc._ops._guess_routine_from_dtype(dtype)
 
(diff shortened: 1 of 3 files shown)


cupyx/jit/_compile.py
```diff
@@ -285,15 +297,22 @@ def _transpile_stmt(stmt, env):
...
         raise NotImplementedError('Not implemented.')
     if isinstance(stmt, ast.If):
-        raise NotImplementedError('Not implemented.')
+        condition = _transpile_expr(stmt.test, env)
+        if is_constants([condition]):
+            stmts = stmt.body if condition.obj else stmt.orelse
+            return _transpile_stmts(stmts, env)
+        head = f'if ({condition.code})'
+        then_body = _transpile_stmts(stmt.body, env)
+        else_body = _transpile_stmts(stmt.orelse, env)
```
leofang, line 314: How about an `if ... elif ... elif ... else` block? Does `elif` get handled here?

By the way, I noticed Cython translates such a block to a C switch when the condition is simple enough (ex: condition is a runtime constant). Do you think it's possible to do so here? Just curious 🙂

asi1024: Python's `ast` module handles `if ... elif ... else ...` equivalent to `if ... else { if ... else ... }`.
For example, in the test case [test_elif](https://github.com/cupy/cupy/pull/4646/files#diff-e0187cde046aeb239283aa518edbbac8aae3c09193ab03ccc82d2c607e79f1f3R324-R332), cupy.vectorize generates the following cuda code.
```cpp
__device__ long long cupy_remainder_int64(T0 in0, T1 in1) {
    return (in0 - _floor_divide(in0, in1) * in1) * (in1 != 0);
}

__device__ long long func_if(long long x) {
  long long y;
  y = 0ll;
  if ((cupy_remainder_int64(x, 2ll) == 0ll)) {
    y = x;
  }
  else {
    if ((cupy_remainder_int64(x, 3ll) == 0ll)) {
      y = (-x);
    }
    else {
    }
  }
  return y;
}
```

asi1024:
Rebased!

cupyx/jit/_typerules.py
```diff
@@ -132,4 +136,29 @@ def get_ctype_from_scalar(mode, x):
...
             return _types.Scalar(numpy.complex64)
 
-    raise NotImplementedError(f'{x} is not supported as a constant.')
+    raise NotImplementedError(f'{x} is not scalar object.')
+
+
+_suffix_literals_dict = {
+    numpy.dtype('float64'): '',
+    numpy.dtype('float32'): 'f',
+    numpy.dtype('int64'): 'll',
```
emcastillo, line 145: I am not sure, but could this cause problems in windows?
I think it should be ok, but double check is not bad. @leofang

leofang: Prefer q over l whenever possible.

cupyx/jit/_compile.py
leofang: Cool, didn't know the ast's behavior. Thanks!

emcastillo (commented):
LGTM!!
The only thing is the dtype concern

cupyx/jit/_typerules.py
asi1024: Fixed in b84eb5f!

emcastillo:
Jenkins, test this please.

Awesome work!

emcastillo (approved):

Issue #4132, opened 2020-10-14
ROCm 3.5+ support roadmap


emcastillo:
Tracking issue for ROCm tasks before v9
@leofang 

- rocSOLVER/ hipMAGMA

leofang:
Library support:
- sparse: #4101 
- dnn: #4100 

Improvement:
- move from hipBLAS to rocBLAS (#3989)
- SVD (available since ROCm 3.8.0?)

Known bugs:
- #4078 
- #3841 
- #3834 
- #3833

leofang:
Looks like ROCm 3.9.0 is out...😞😞😞

leofang:
> Looks like ROCm 3.9.0 is out...😞😞😞

It seems CuPy master builds just fine on 3.9.0: https://ci.preferred.jp/cupy.py3.amd/64299/

leofang:
btw I noticed the cythonized .c & .cpp codes could be very different between CUDA and HIP builds, so if uploading sdist to PyPI is on the roadmap, this issue must be addressed (ex: can we upload two sets of sdist, one for CUDA and the other for HIP?)

leofang:
ROCm 3.10:
<img width="384" alt="rocsolverAPI" src="https://user-images.githubusercontent.com/5534781/101657748-3d778f00-3a12-11eb-8665-88bdc0545318.PNG">
https://rocmdocs.amd.com/en/latest/Current_Release_Notes/Current-Release-Notes.html#new-rocsolver-apis

kmaehashi:
https://github.com/cupy/cupy/pull/4461

> grids with total size of 2**32 do not launch threads at all.

this is not fixed as of ROCm 3.10/4.0.0.

kmaehashi:
We have to take care of branching based on CUDA versions.
e.g. https://github.com/cupy/cupy/blob/65ca0818b8c922e52adf2c5788226316849c60d7/cupy/core/core.pyx#L1971-L1977

leofang:
> We have to take care of branching based on CUDA versions.

@kmaehashi This will be handled in #4102.

kmaehashi:
> > We have to take care of branching based on CUDA versions.

> @kmaehashi This will be handled in #4102.

Ah, nice! 👍🏼

leofang:
[...]

PR #4651, opened 2021-02-09
ROCm: disable PTDS

leofang:
Follow-up of #4322.

PTDS does not work on HIP and leads to segfault when going out of scope (https://github.com/cupy/cupy/issues/4132#issuecomment-775734406). This PR disables it on HIP:
- Raise if `ptds=True` is set in the constructor
- Avoid creating the `cupy.cuda.Stream.ptds` singeleton
- Skip PTDS tests


leofang:
Jenkins, test this please

emcastillo (approved):
Thanks!

Issue #4540, opened 2021-01-12
ROCm: Error in hipBLAS dgmm 


leofang:
From https://github.com/cupy/cupy/pull/4217#issuecomment-749783900:
```
$ pytest tests/cupyx_tests/test_cupyx.py tests/cupy_tests/test_cublas.py tests/cupy_tests/test_cusolver.py 
...
======================================================================= short test summary info =======================================================================
FAILED tests/cupy_tests/test_cublas.py::TestDgmm_param_5_{ordera='C', orderc='F', shape=(9, 10), side='R'}::test_dgmm_incx_minus_one - AssertionError: 
FAILED tests/cupy_tests/test_cublas.py::TestDgmm_param_7_{ordera='C', orderc='F', shape=(10, 9), side='R'}::test_dgmm_incx_minus_one - AssertionError: 
FAILED tests/cupy_tests/test_cublas.py::TestDgmm_param_12_{ordera='F', orderc='F', shape=(9, 10), side='L'}::test_dgmm_incx_minus_one - AssertionError: 
FAILED tests/cupy_tests/test_cublas.py::TestDgmm_param_14_{ordera='F', orderc='F', shape=(10, 9), side='L'}::test_dgmm_incx_minus_one - AssertionError: 
============================================================ 4 failed, 2434 passed, 640 skipped in 10.53s =============================================================
```
It seems the trigger condition is
- mode=0
- incx=-1

I suspect this is a hipBLAS bug.

leofang:
cc: @amathews-amd

AdrianAbeyta:
M=rows, N=cols seems reversed for ROCm compared with CUDA; Internal Ticket: SWDEV-309804

PR #4652, opened 2021-02-10
Skip failing tests for `cupy.cublas` under HIP

takagi:
Rel #4132, #4540.

### DIFF
--- a/tests/cupy_tests/test_cublas.py
+++ b/tests/cupy_tests/test_cublas.py
@@ -453,11 +453,22 @@ class TestDgmm(unittest.TestCase):
         cublas.dgmm(self.side, self.a, self.x, out=self.a)
         cupy.testing.assert_allclose(self.a, ref, rtol=self.tol, atol=self.tol)
 
-    @pytest.mark.skipif(cupy.cuda.runtime.is_hip,
-                        reason="HIP fails for some condition")
+    _dgmm_incx_minus_one_hip_skip_condition = [
+        ('C', 'F', (9, 10), 'R'),
+        ('C', 'F', (10, 9), 'R'),
+        ('F', 'F', (9, 10), 'L'),
+        ('F', 'F', (10, 9), 'L'),
+    ]
+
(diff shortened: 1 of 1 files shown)


kmaehashi:
Could you elaborate the failing conditions more? In general simple `cublas.dgmm` case should work with ROCm too.

leofang:
Skip or xfail?

tests/cupy_tests/test_cublas.py
```diff
@@ -453,12 +453,23 @@ def test_dgmm_inplace(self, dtype):
         cublas.dgmm(self.side, self.a, self.x, out=self.a)
         cupy.testing.assert_allclose(self.a, ref, rtol=self.tol, atol=self.tol)
 
-    @pytest.mark.skipif(cupy.cuda.runtime.is_hip,
-                        reason="HIP fails for some condition")
+    _dgmm_incx_minus_one_hip_skip_condition = [
+        ('C', 'F', (9, 10), 'R'),
+        ('C', 'F', (10, 9), 'R'),
+        ('F', 'F', (9, 10), 'L'),
+        ('F', 'F', (10, 9), 'L'),
```
emcastillo, line 460: Seems like it fails only when c is 'F' order is involved regardless of the other parameters?

leofang: When I dug into how the API was called, the condition was (#4540)
- mode=0
- incx=-1

Not sure how this maps to that c is in  the `F` order.

leofang:
btw It'd be great to keep a list of all HIP-xfail tests somewhere so that we can revisit them.

tests/cupy_tests/test_cublas.py
takagi: There are tests that pass even when c is 'F'. I can't find an apparent rule there.

emcastillo (approved):
LGTM

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

PR #4654, opened 2021-02-10
Xfail test for cupy._indexing.generate under HIP

takagi:
Rel #4132.

This test strangely fails. Sometimes an additional statement gets the fail away.


leofang:
This is part of https://github.com/cupy/cupy/issues/4484#issuecomment-767067322.

emcastillo:
If the test fails intermitently, there might be a memory corruption happening in another tests or some issue in synchronization.

emcastillo (approved):

emcastillo:
Jenkins, test this please

PR #4648, opened 2021-02-09
Fix typecast from complex to bool in JIT

asi1024:
https://jenkins.preferred.jp/job/chainer/job/cupy_pr/2972/TEST=cupy-py36,label=mn1-p100/testReport/junit/tests.cupy_tests.functional_tests.test_vectorize/TestVectorizeExprs/test_vectorize_typecast/

PR #4659, opened 2021-02-10
Update CUDA Array Interface to v3 - Part 2

leofang:
Follow up of #4357 and #4322. This PR should ***not*** be backported.

Now that we support PTDS (#4322), we should detect if it's in use before exporting (context: https://github.com/cupy/cupy/pull/4357#issuecomment-735852102). The test suite is updated accordingly. One of the main focus is to ensure the tests pass regardless of the value of the env var `CUPY_CUDA_PER_THREAD_DEFAULT_STREAM` (verified locally). 

This PR also adds an extra safe for HIP following #4651: On HIP, `CUPY_CUDA_PER_THREAD_DEFAULT_STREAM` should be ignored as PTDS is not supported.

cc: @pentschev

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -49,6 +49,7 @@ from cupy.cuda cimport function
 from cupy.cuda cimport pinned_memory
 from cupy.cuda cimport memory
 from cupy.cuda cimport stream as stream_module
+from cupy_backends.cuda cimport stream as _stream_module
 from cupy_backends.cuda.api cimport runtime
 from cupy_backends.cuda.libs cimport cublas
 
@@ -191,10 +192,12 @@ cdef class ndarray:
 
         if ver == 3:
             stream_ptr = stream_module.get_current_stream_ptr()
-            # TODO(leofang): check if we're using PTDS
             # CAI v3 says setting the stream field to 0 is disallowed
             if stream_ptr == 0:
-                stream_ptr = 1  # TODO(leofang): use runtime.streamLegacy
+                if _stream_module.is_ptds_enabled():
+                    stream_ptr = runtime.streamPerThread
+                else:
+                    stream_ptr = runtime.streamLegacy
             desc['stream'] = stream_ptr
         elif ver == 2:
             # Old behavior (prior to CAI v3): stream sync is explicitly handled
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -194,10 +194,7 @@ cdef class ndarray:
             stream_ptr = stream_module.get_current_stream_ptr()
             # CAI v3 says setting the stream field to 0 is disallowed
             if stream_ptr == 0:
-                if _stream_module.is_ptds_enabled():
-                    stream_ptr = runtime.streamPerThread
-                else:
-                    stream_ptr = runtime.streamLegacy
+                stream_ptr = _stream_module.get_default_stream_ptr()
             desc['stream'] = stream_ptr
         elif ver == 2:
             # Old behavior (prior to CAI v3): stream sync is explicitly handled
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -218,7 +218,7 @@ class TestNdarrayCudaInterface(unittest.TestCase):
         assert not iface['data'][1]
         assert iface['descr'] == [('', '<f8')]
         assert iface['strides'] is None
-        assert iface['stream'] == 2 if stream_module.is_ptds_enabled() else 1
+        assert iface['stream'] == stream_module.get_default_stream_ptr()
 
     def test_cuda_array_interface_view(self):
         arr = cupy.zeros(shape=(10, 20), dtype=cupy.float64)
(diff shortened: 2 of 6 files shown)


leofang:
Jenkins, test this please

leofang:
Looks like the failure will be fixed in #4648.

pentschev (approved):
Thanks @leofang for this PR, I completely forgot about `__cuda_array_interface__` when I submitted the PTDS PR. This looks good already to me! 😄

leofang:
Thanks @pentschev! No worries, this was on my todo list anyway. Was just waiting for PTDS to go in.

leofang:
Jenkins, test this please

emcastillo:
Jenkins, test this please

cupy/core/core.pyx
```diff
@@ -191,10 +192,12 @@ cdef class ndarray:
 
         if ver == 3:
             stream_ptr = stream_module.get_current_stream_ptr()
-            # TODO(leofang): check if we're using PTDS
             # CAI v3 says setting the stream field to 0 is disallowed
             if stream_ptr == 0:
-                stream_ptr = 1  # TODO(leofang): use runtime.streamLegacy
+                if _stream_module.is_ptds_enabled():
```
emcastillo, line 197: Can we add a function to `cupy_backends.cuda.stream` that gives us the current default stream?
This if is being repeated a lot all over the code and I believe that this would simplify it.

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -217,7 +218,7 @@ def test_cuda_array_interface(self):
         assert not iface['data'][1]
         assert iface['descr'] == [('', '<f8')]
         assert iface['strides'] is None
-        assert iface['stream'] == 1
+        assert iface['stream'] == 2 if stream_module.is_ptds_enabled() else 1
```
emcastillo, line 221: using the above proposed function cleans this code

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -295,7 +297,13 @@ def test_cuda_array_interface_stream(self):
         assert iface['descr'] == [('', '<f8')]
         assert iface['strides'] is None
         if self.ver == 3:
-            assert iface['stream'] == 1 if stream.ptr == 0 else stream.ptr
+            if stream.ptr == 0:
+                if stream_module.is_ptds_enabled():
```
emcastillo, line 301: Same, this could be simplified
```
if stream.ptr == 0:
    assert iface['stream'] == current_default_stream()
```

tests/cupy_tests/creation_tests/test_from_data.py
```diff
@@ -594,9 +594,10 @@ def __cuda_array_interface__(self):
         # synchronization is done via calling a cpdef function, which cannot
         # be mock-tested.
         if self.stream is not None:
-            # TODO(leofang): how about PTDS?
             if self.stream is cuda.Stream.null:
-                desc['stream'] = 1  # TODO(leofang): use runtime.streamLegacy
+                desc['stream'] = cuda.runtime.streamLegacy
+            elif (not cuda.runtime.is_hip) and self.stream is cuda.Stream.ptds:
+                desc['stream'] = cuda.runtime.streamPerThread
```
emcastillo, line 600: This can be cleaned up too

emcastillo (commented):
I don't have any issues with the proposed solution.
Just some cosmetic touches to simplify the code.

cupy/core/core.pyx
leofang: Sounds good. I call it `get_default_stream_ptr()` because I want to avoid using "current", which could be confusing if the default stream is not current.

leofang: Done in 7a938a8.

tests/cupy_tests/core_tests/test_ndarray.py
leofang: Done in 7a938a8.

tests/cupy_tests/core_tests/test_ndarray.py
leofang: Done in 7a938a8.

tests/cupy_tests/creation_tests/test_from_data.py
leofang: No, not here, because here we need to get the pointer from `self.stream`.

cupy/core/core.pyx
leofang: @emcastillo btw this function would never return `0` (the default null stream), only `1` or `2`.

leofang:
@emcastillo PTAL.

Jenkins, test this please

cupy/core/core.pyx
pentschev: Didn't we roll back to `0` because of the NCCL bug with `cudaStreamLegacy`/`cudaStreamPerThread`?

leofang: I think for now this function is used only to deal with CUDA Array Interface, which does not allow `0`. Should I add a comment to the function?

pentschev: I think I forgot about this with CUDA Array Interface, and probably unnecessary to add a comment. I think this is a very edgy case that is already fixed upstream anyway.

emcastillo (approved):

emcastillo:
Much cleaner, Thanks!!

PR #4664, opened 2021-02-13
Add diagonal and setdiag methods for COO sparse matrices

grlee77:
The functions added here are direct copies of the SciPy methods, only switching out `np`->`cupy`.

`diagonal` worked okay prior to this PR via the base class `diagonal ` method which first converted to CSR. 

The `TestCooMatrixDiagonal` class here is modification of the existing `TestCsrMatrixDiagonal` class. The only methods that needed to be modified for the COO case were `_make_matrix` and `_test_setdiag`.

### DIFF
--- a/cupyx/scipy/sparse/coo.py
+++ b/cupyx/scipy/sparse/coo.py
@@ -183,6 +183,34 @@ class coo_matrix(sparse_data._data_matrix):
                 (data, (self.row, self.col)), shape=self.shape,
                 dtype=data.dtype)
 
+    def diagonal(self, k=0):
+        """Returns the k-th diagonal of the matrix.
+
+        Args:
+            k (int, optional): Which diagonal to get, corresponding to elements
+            a[i, i+k]. Default: 0 (the main diagonal).
+
+        Returns:
+            cupy.ndarray : The k-th diagonal.
+        """
+        rows, cols = self.shape
+        if k <= -rows or k >= cols:
+            return cupy.empty(0, dtype=self.data.dtype)
+        diag = cupy.zeros(min(rows + min(k, 0), cols - max(k, 0)),
+                        dtype=self.dtype)
+        diag_mask = (self.row + k) == self.col
+
+        if self.has_canonical_format:
+            row = self.row[diag_mask]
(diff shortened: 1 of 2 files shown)


grlee77:
`diagonal` benchmark script

```Python
density = 0.25
shape = (8192, 8192)
dtype = np.float32
a = testing.shaped_random(shape, cupy, dtype=dtype)
mask = testing.shaped_random(shape, cupy, dtype='f', scale=1.0)
a[mask > density] = 0
cupyx_a = sparse.coo_matrix(a)

def get_diag_old(a):  # convert to CSR as in the base class diagonal method
    return a.tocsr().diagonal()

def get_diag_new(a):
    return a.diagonal()

perf_old = repeat(get_diag_old, (cupyx_a,), n_warmup=5, n_repeat=10)
print(perf_old)

perf_new = repeat(get_diag_new, (cupyx_a,), n_warmup=100, n_repeat=500)
print(perf_new)

````

```
get_diag_old        :    CPU:152304.076 us   +/-201.652 (min:151929.896 / max:152518.376) us     GPU-0:200185.504 us   +/-186.176 (min:199851.013 / max:200420.197) us

get_diag_new        :    CPU: 2244.524 us   +/-18.557 (min: 2223.562 / max: 2453.090) us     GPU-0: 2369.057 us   +/-14.685 (min: 2353.760 / max: 2573.312) us
```

The diagonal method introduced here takes about 2.5 ms for the matrix in the demo above vs. 200 ms for the one involving a COO->CSR conversion. The CSR diagonal method itself is very fast (~0.04 ms), so it is the conversion from COO->CSR that accounts for nearly all of the time in the old approach.

leofang:
> it is the conversion from COO->CSR that accounts for nearly all of the time in the old approach

Right, I've seen that in some sparse PRs too. It'd be great to avoid such conversions if possible, but it seems to make both implementation and maintenance harder...

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi (approved):
LGTM! Super sorry for leaving this so long...

PR #4669, opened 2021-02-15
Fix definition order in preamble

takagi:
Rel #4132 and #4484.

This PR fixes the order of definitions of `gcd` and `lcm` in the preamble as the HIP compiler requires.


emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4672, opened 2021-02-16
Support assignment of compile-time constants in CuPy JIT

asi1024:
~Depends on #4646~

### DIFF
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -88,8 +88,8 @@ class Environment:
 
     Attributes:
         mode ('numpy' or 'cuda'): The rule for typecast.
-        globals (dict): The dictionary with keys as the variable names and
-            the values as the data stored at the global scopes.
+        consts (dict): The dictionary with keys as the variable names and
+            the values as the data that is determined at compile-time.
         params (dict): The dictionary of function arguments with keys as
             the variable names and the values as the CudaObject.
         locals (dict): The dictionary with keys as the variable names and the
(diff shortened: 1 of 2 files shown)


cupyx/jit/_compile.py
```diff
@@ -244,14 +244,14 @@ def _call_ufunc(ufunc, args, dtype, env):
     raise NotImplementedError(f'ufunc `{ufunc.name}` is not supported.')
 
 
-def _transpile_stmts(stmts, env):
+def _transpile_stmts(stmts, is_toplevel, env):
```
emcastillo, line 247: I am not sure but will it make sense to use a counter for the current level that you are transpiling?
I wonder if there are situations where declaring a variable inside an if/for or a block and accessing it outside can be detected this way.

emcastillo: I think we can have the variables declared in the `environment` with the scope level they are accessible,
Maybe using a stack of dicts, and looking through it?, then the level is just the length of the stack in the env.

asi1024: All local variables are declared at the top of the function.
https://github.com/cupy/cupy/blob/b5df3aa4d09b54bb45c82c184a88bfb93dd7cfd9/cupyx/jit/_compile.py#L169-L171
I think we would rather not support the scope level of local variable, because Python itself does not support it.

asi1024: When a non-constant value is assigned to a variable, the current compiler implementation emits a CUDA code that has a local variable declaration at the top of the function. I didn't adopt the scope of the local variable in generated code because Python does not have it.
We can assign constant variables also in if/for statement body if we introduce scope level in the compiler, but I personally don't want to adopt a different rule from non-constant variables.

emcastillo: Seems fair to me :).

emcastillo:
Jenkins, test this please

emcastillo (approved):

asi1024:
Jenkins, test this please.

Issue #4549, opened 2021-01-12
Numpy dispatch of matmul doesn't work 


ravwojdyla:
* Conditions
OS                           : Linux-4.19.112+-x86_64-with-Ubuntu-18.04-bionic
CuPy Version                 : 8.3.0
NumPy Version                : 1.19.5
SciPy Version                : 1.4.1
Cython Build Version         : 0.29.21
CUDA Root                    : /usr/local/cuda
CUDA Build Version           : 10010
CUDA Driver Version          : 10010
CUDA Runtime Version         : 10010
cuBLAS Version               : 10201
cuFFT Version                : 10101
cuRAND Version               : 10101
cuSOLVER Version             : (10, 2, 0)
cuSPARSE Version             : 10300
NVRTC Version                : (10, 1)
Thrust Version               : 100906
CUB Build Version            : 100800
cuDNN Build Version          : None
cuDNN Version                : None
NCCL Build Version           : 2708
NCCL Runtime Version         : 2708
cuTENSOR Version             : None
Device 0 Name                : Tesla T4
Device 0 Compute Capability  : 75
* Code to reproduce
```python
import numpy as np
import cupy as cp

a = np.random.randn(100, 100)
a_cpu = np.asarray(a)
a_gpu = cp.asarray(a)

# numpy on CPU arrays works:
np.matmul(a_cpu, a_cpu)
# direct cupy matmul works:
cp.matmul(a_gpu, a_gpu)
# some ufunc work fine:
np.dot(a_gpu, a_gpu)

# np dispatch of matmul doesn't work:
np.matmul(a_gpu, a_gpu)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-fecacee25691> in <module>()
----> 1 np.matmul(a_gpu, a_gpu)

TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'matmul'>, '__call__', array([[-1.41473897, -0.81697767,  0.93869847, ..., -0.27356864,
        -0.18850912,  1.03690492],
       [ 1.30736764, -1.17802181, -0.53378097, ..., -1.37164886,
        -1.40653998, -0.69411561],
       [-0.54080081,  0.63547369, -0.22747362, ..., -1.54336986,
        -0.2687723 ,  0.04278493],
       ...,
       [ 0.31192026,  0.82086513, -0.00734687, ...,  0.17736056,
        -1.13869101, -0.61541535],
[...]

Issue #2552, opened 2019-10-18
Support generalized universal functions


emcastillo:
Cupy does not offer generalized universal functions.
I would like to discuss how should we proceed and implement this feature.

https://docs.scipy.org/doc/numpy/reference/c-api.generalized-ufuncs.html

TODOs

- [ ] Parse gufuncs signatures to retrieve the core dimensions.
- [ ] Modify `core.pyx::ndarray::__array_ufunc__` to remove the signature check.
- [ ] Given the signature, modify `_kernel.pyx::ufunc::__call__` to call the 
       function sub-arraywise. Ufunc class relies on elementwise kernels when calling `_get_ufunc_kernel`. GUfuncs can also call kernels operating on regular vectors.
- [ ] Convert cupy inner1d, matmat, vecmat, matvec, matmul, outer_inner to actual gufuncs
- [ ] Add features such as `axes`,  `order`, etc.

List of numpy gufuncs (thanks to @asi1024)
```
(n?,k),(k,m?)->(n?,m?)
         matmul
(m,m)->(),()
         slogdet
(m,m)->()
         det
(m,m)->(m),(m,m)
         eigh_lo
         eigh_up
         eig
(m,m)->(m)
         eigvalsh_up
         eigvalsh_lo
         eigvals
(m,m),(m,n)->(m,n)
         solve
(m,m),(m)->(m)
         solve1
(m, m)->(m, m)
         inv
(m,m)->(m,m)
         cholesky_lo
(m,n)->(m)
         svd_m
(m,n)->(n)
         svd_n
(m,n)->(m,m),(m),(m,n)
         svd_m_s
(m,n)->(m,n),(n),(n,n)
         svd_n_s
(m,n)->(m,m),(m),(n,n)
         svd_m_f
(m,n)->(m,m),(n),(n,n)
         svd_n_f
(m,n),(m,nrhs),()->(n,nrhs),(nrhs),(),(m)
         lstsq_m
(m,n),(m,nrhs),()->(n,nrhs),(nrhs),(),(n)
         lstsq_n
```

leofang:
FYI, Numba already supports this, see http://numba.pydata.org/numba-doc/latest/cuda/ufunc.html. But I am not sure if their generalized ufuncs support CuPy arrays (via `__cuda_array_interface__`). Should be quickly verifiable.

emcastillo:
Dask have some facilities too https://docs.dask.org/en/latest/array-gufunc.html

[...]

PR #4675, opened 2021-02-17
Support for Generalized Universal Functions

emcastillo:
This is a big nasty one.

Currently wip.

Adds a GUFunc class (all the axes and batching stuff has been copied straight from dask).

There are some heavy incompatibilities here with things such as `matmul` since all the batching is done in the relevant cublas routine, we should avoid doing repeated calls at any cost. So for some gufuncs we need to skip several of these steps.

Also, we need to find a way to blend this with the regular `UniversalFunction`, probably wrapping it and doing a straight call should be enough, but we will provide additional support for currently non-supported kwargs such as `where`.

This needs to be cythonized too.

Fixes:
#4549 
#2552

### DIFF
--- a/cupy/core/_gufuncs.py
+++ b/cupy/core/_gufuncs.py
@@ -1,0 +1,482 @@
+import numpy
+import re
+
+import cupy
+
+import cupy.core._routines_manipulation as _manipulation
+from cupy.core import internal
+
+# Signature parsing code and dimension accessing has been borrowed
+# from dask
+# https://github.com/dask/dask/blob/61b578f5a3ad88cbc6a8b9a73ce08c551bd969fa/dask/array/gufunc.py#L12-L55
+_DIMENSION_NAME = r'\w+\?*'
+_CORE_DIMENSION_LIST = '(?:{0:}(?:,{0:})*,?)?'.format(_DIMENSION_NAME)
+_ARGUMENT = r'\({}\)'.format(_CORE_DIMENSION_LIST)
+_INPUT_ARGUMENTS = '(?:{0:}(?:,{0:})*,?)?'.format(_ARGUMENT)
+_OUTPUT_ARGUMENTS = '{0:}(?:,{0:})*'.format(
+    _ARGUMENT
+)  # Use `'{0:}(?:,{0:})*,?'` if gufunc-
+# signature should be allowed for length 1 tuple returns
+_SIGNATURE = '^{0:}->{1:}$'.format(_INPUT_ARGUMENTS, _OUTPUT_ARGUMENTS)
+
+
+def _parse_gufunc_signature(signature):
+    # The code has been modifyed from dask to support optional dimensions
+    if not isinstance(signature, str):
+        raise TypeError('Signature is not a string')
+
+    if signature == '' or signature is None:
+        raise ValueError('Signature cannot be empty')
+
+    signature = signature.replace(' ', '')
+    if not re.match(_SIGNATURE, signature):
+        raise ValueError('Not a valid gufunc signature: {}'.format(signature))
+    in_txt, out_txt = signature.split('->')
+    ins = [tuple(x.split(',')) if x != '' else ()
+           for x in in_txt[1:-1].split('),(')]
+    outs = [tuple(y.split(',')) if y != '' else ()
+            for y in out_txt[1:-1].split('),(')]
+    # TODO(ecastill) multiple output support
+    if len(outs) > 1:
+        raise ValueError('Currently more than 1 output is not supported')
+    outs = outs[0] if ((len(outs) == 1) and (out_txt[-1] != ',')) else outs
+    return ins, outs
+
+
+def _validate_normalize_axes(
+    axes, axis, keepdims, input_coredimss, output_coredimss
+):
+    # This code credit goes to Dask
+    # https://github.com/dask/dask/blob/61b578f5a3ad88cbc6a8b9a73ce08c551bd969fa/dask/array/gufunc.py#L58-L172
+    nin = len(input_coredimss)
+    nout = (
+        1 if not isinstance(output_coredimss, list) else len(output_coredimss)
+    )
+
+    if axes is not None and axis is not None:
+        raise ValueError(
+            'Only one of `axis` or `axes` keyword arguments should be given')
+    if axes and not isinstance(axes, list):
+        raise ValueError('`axes` has to be of type list')
+
+    output_coredimss = output_coredimss if nout > 1 else [output_coredimss]
+    filtered_core_dims = list(filter(len, input_coredimss))
+    nr_outputs_with_coredims = len(
+        [True for x in output_coredimss if len(x) > 0])
+
+    if keepdims:
+        if nr_outputs_with_coredims > 0:
+            raise ValueError('`keepdims` can only be used for scalar outputs')
+        output_coredimss = len(output_coredimss) * [filtered_core_dims[0]]
+
+    core_dims = input_coredimss + output_coredimss
+    if axis is not None:
+        if not isinstance(axis, int):
+            raise ValueError('`axis` argument has to be an integer value')
+        if filtered_core_dims:
+            cd0 = filtered_core_dims[0]
+            if len(cd0) != 1:
+                raise ValueError(
+                    '`axis` can be used only, if one core dimension is present'
+                )
+            for cd in filtered_core_dims:
+                if cd0 != cd:
+                    raise ValueError(
+                        'To use `axis`, all core dimensions have to be equal'
+                    )
+
+    # Expand defaults or axis
+    if axes is None:
+        if axis is not None:
+            axes = [(axis,) if cd else tuple() for cd in core_dims]
+        else:
+            axes = [tuple(range(-len(icd), 0)) for icd in core_dims]
+
+    axes = [(a,) if isinstance(a, int) else a for a in axes]
+
+    if (
+        (nr_outputs_with_coredims == 0)
+        and (nin != len(axes))
+        and (nin + nout != len(axes))
+    ) or ((nr_outputs_with_coredims > 0) and (nin + nout != len(axes))):
+        raise ValueError(
+            'The number of `axes` entries is not equal the number'
+            ' of input and output arguments')
+
+    # Treat outputs
+    output_axes = axes[nin:]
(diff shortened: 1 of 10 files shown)


jakirkham:
cc @seberg @pentschev (in case this is of interest 🙂)

leofang:
Since @seberg its pinged, I'd like to know whether https://github.com/cupy/cupy/issues/4549 (which this PR is trying to fix) should be considered as a NumPy bug or not. NumPy made the gufuncs not dispatchable via `__array_function__` IMHO, see https://github.com/cupy/cupy/issues/4549#issuecomment-759922990.

seberg:
I am not sure I understand. Since `__array_ufunc__` takes complete control, it doesn't matter whether the cupy version of `matmul` is implemented as a ufunc?  IIRC the `__array_function__` NEP basically says, `__array_ufunc__` is preferable, and any function may be moved to that (i.e. be converted into a ufunc if applicable).

I admit it is a bit weird that some cases will actually dispatch to a ufunc from within a non-ufunc wrapped with `__array_function__`

emcastillo:
Hi!
in CuPy right now `matmul` is a regular function instead of a ufunc and does not support most of the ufunc kwargs, so we are not dispatching properly. The issue is entirely in CuPy function dispatcher. 
This PR is adding a wrapper to be used by matmul and other ufuncs so that now dispatch will behave properly, and we also get arguments such as `axes` and others in routines that did not support them before.

emcastillo:
Btw, although the code is 60% straight from dask I added support for stuff such as optional dimensions in signatures `(m?, n), (n, k?) -> (m?, k?)`, `out` kwarg, and now I am working on supporting dtypes and type signatures. So hopefully we will have a numpy 1:1 GUFunc API.

How should I reference Dask here? All the matrix transposition credits goes to them, so I don't know if I have to reproduce their license in the code.

pentschev:
> How should I reference Dask here? All the matrix transposition credits goes to them, so I don't know if I have to reproduce their license in the code.

Usually Dask itself adds a comment to where it was taken from with a GitHub permalink (see https://github.com/dask/dask/blob/91cee5b7944e47959f475eaebc42dc75be14bc21/dask/array/numpy_compat.py#L167-L208 for reference), so maybe do the same here?

emcastillo:
PTAL,
Still some work left for future PRs.

Cythonize -> better performance
Improve test coverage
Support multiple outputs
Support dtype, order and the rest of kwargs.

BTW, `matmul` dispatch works now, so Dask will not complain :)

cupy/core/_gufuncs.py
```diff
@@ -0,0 +1,500 @@
...
+_OUTPUT_ARGUMENTS = '{0:}(?:,{0:})*'.format(
+    _ARGUMENT
+)  # Use `'{0:}(?:,{0:})*,?'` if gufunc-
+# signature should be allowed for length 1 tuple returns
+_SIGNATURE = '^{0:}->{1:}$'.format(_INPUT_ARGUMENTS, _OUTPUT_ARGUMENTS)
+
+
+def _parse_gufunc_signature(signature):
+    # The code has been modifyed from dask to support optional dimensions
+    if type(signature) != str:
```
leofang, line 24: nitpick:
```suggestion
    if not isinstance(signature, str):
```

cupy/core/_gufuncs.py
```diff
@@ -0,0 +1,500 @@
...
+                    raise ValueError(
+                        'To use `keepdims`, all core dimensions'
+                        ' have to be equal')
+            iax0 = input_axes[0]
+            output_axes = [iax0 for _ in output_coredimss]
+
+    return input_axes, output_axes
+
+
+def _broadcast_shapes(shapes):
```
leofang, line 142: This is exactly a dup of this one:
https://github.com/cupy/cupy/blob/cbbb13acfabc7cc8978c25cea56682b7ef398bd2/cupy/core/_fusion_trace.pyx#L99

leofang (changes requested):
I haven't finished a full pass and will do so asap 😅 Found two minor issues so far.

cupy/core/_routines_linalg.pyx
```diff
@@ -931,3 +932,5 @@ cpdef ndarray matmul(ndarray a, ndarray b, ndarray out=None):
         ret = ndarray(out_shape, ret_dtype)
         elementwise_copy(out, ret)
         return ret
+
+matmul = _GUFunc(_matmul, '(n?,k),(k,m?)->(n?,m?)', supports_batched=True)
```
asi1024, line 936: Could you move this line to `cupy/linalg/_product.py`?

cupy/core/_gufuncs.py
```diff
@@ -0,0 +1,500 @@
...
+                    shape = leaf_arr.shape
+                    # This line deletes the dimensions that were not present
+                    # in the input
+                    core_shape = shape[-len(ocd):]
+                    core_shape = tuple([
+                        d for d, n in zip(core_shape, ocd) if n not in m_dims])
+                    shape = shape[:-len(ocd)] + core_shape
+                    leaf_arr = leaf_arr.reshape(shape)
+                leaf_arrs.append(leaf_arr)
+        return (*leaf_arrs,) if self._nout else leaf_arrs[0]
```
asi1024, line 500: For the readabiltiy.
```suggestion
        return tuple(leaf_arrs) if self._nout else leaf_arrs[0]
```

cupy/core/_gufuncs.py
```diff
@@ -0,0 +1,500 @@
...
+    def _transpose_element(self, arg, iax, shape):
+        iax = tuple(a if a < 0 else a - len(shape) for a in iax)
+        tidc = (
+            tuple(i for i in range(
+                -len(shape) + 0, 0) if i not in iax) + iax
+        )
+        return arg.transpose(tidc)
+
+    def _get_args_transposed(self, args, input_axes, input_coredimss,
+                             outs, output_axes, output_coredimss):
```
asi1024, line 259: Could you remove `input_coredimss` and `output_coredimss` argument and use `self.*` instead?

cupy/core/_gufuncs.py
```diff
@@ -0,0 +1,500 @@
...
+                    max_loopdims - n, max_loopdims)
+            )
+            for n in num_loopdims
+        ]
+        input_dimss = [li + c for li, c in zip(
+            loop_input_dimss, input_coredimss)]
+
+        loop_output_dims = (
+            max(loop_input_dimss, key=len) if loop_input_dimss else tuple()
+        )
```
asi1024, line 337:
```suggestion
        loop_output_dims = max(loop_input_dimss, key=len, default=())
```

cupy/core/_gufuncs.py
```diff
@@ -0,0 +1,500 @@
...
+        out_shape = tuple([dimsizess[od][0] for od in loop_output_dims]
+                          + [dimsizess[od][0] for od in output_coredimss])
+        if outs is None:
+            outs = (cupy.zeros(out_shape, dtype=ret_dtype),)
+        elif outs[0].shape != out_shape:
+            raise ValueError(f'Invalid shape for out {outs[0].shape}'
+                             f' needs {out_shape}')
+        self._apply_func_to_inputs(0, dimsizess, loop_output_dims, args, outs)
+
+        tmp = outs
```
asi1024, line 457:
```suggestion
```

cupy/core/_gufuncs.py
```diff
@@ -0,0 +1,500 @@
...
+        args, dimsizess, loop_output_dims, outs, m_dims = self._get_args_transposed(  # NOQA
+            args, input_axes, input_coredimss,
+            outs, output_axes, output_coredimss)
+
+        # The output shape varies depending on optional dims or not
+        # TODO(ecastill) this only works for one out argument
+        out_shape = tuple([dimsizess[od][0] for od in loop_output_dims]
+                          + [dimsizess[od][0] for od in output_coredimss])
+        if outs is None:
+            outs = (cupy.zeros(out_shape, dtype=ret_dtype),)
```
asi1024, line 451: Don't need to initialize with zeros.
```suggestion
            outs = (cupy.empty(out_shape, dtype=ret_dtype),)
```

asi1024: The size of `outs` should be `self._nout`?

emcastillo: I think this is for the case where a gufunc returns two arrays, currently we only support returning one array.

emcastillo:
PTAL!

cupy/core/core.pyx
```diff
@@ -1091,7 +1091,7 @@ cdef class ndarray:
         if not isinstance(y, ndarray) and _should_use_rop(x, y):
             return NotImplemented
         else:
-            return _linalg.matmul(x, y)
+            return cupy.core._routines_linalg._matmul(x, y)
```
asi1024, line 1094:
```suggestion
            return cupy.linalg._product.matmul(x, y)
```

asi1024:
@emcastillo Could you merge the master branch to pass CI?

cupy/core/core.pyx
```diff
@@ -1091,7 +1091,7 @@ cdef class ndarray:
         if not isinstance(y, ndarray) and _should_use_rop(x, y):
             return NotImplemented
         else:
-            return _linalg.matmul(x, y)
+            return cupy.linalg._product.matmul(x, y)
```
leofang, line 1094: Silly question: can't we wrap `_linalg._matmul()` with gufunc only when dispatching is required (i.e. inside `__array_ufunc__`)?

leofang: btw the comment above (line 1090) can be removed.

emcastillo: If we only wrap on dispatching we lose the ability of supporting axis and the other gufunc keywords in the regular matmul.
Also it would pollute the dispatcher by needing to check if the function is matmul or not.

emcastillo:
rebased :)

asi1024:
Jenkins, test this please.

cupy/core/core.pyx
leofang: Thanks, Emilio!

leofang:
Jenkins, test this please

asi1024:
LGTM!

asi1024 (approved):

PR #4677, opened 2021-02-17
Support while/for-statement in CuPy JIT

asi1024:
Part of #4290.

~Depends on #4672.~

### DIFF
--- a/tests/cupy_tests/functional_tests/test_vectorize.py
+++ b/tests/cupy_tests/functional_tests/test_vectorize.py
@@ -379,6 +379,19 @@ class TestVectorizeStmts(unittest.TestCase):
         x = xp.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
         return f(x)
 
+    @testing.numpy_cupy_array_equal()
+    def test_while(self, xp):
+        def func_while(x):
+            y = 0
+            while x > 0:
+                y += x
+                x -= 1
+            return y
+
+        f = xp.vectorize(func_while)
+        x = xp.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+        return f(x)
+
 
 class _MyClass:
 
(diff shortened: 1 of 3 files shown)


tests/cupy_tests/functional_tests/test_vectorize.py
```diff
@@ -379,6 +379,79 @@ def func_if(x):
...
+
+        f = xp.vectorize(func_for)
+        x = xp.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+        return f(x)
+
+    @testing.numpy_cupy_array_equal()
+    def test_for_range_step(self, xp):
+        def func_for(x, y, z):
+            res = 0
+            for i in range(x, y, z):
```
emcastillo, line 422: Can we add a test that checks for compound expressions inside range?
```
for i in range(x, some_function()+1):
```

leofang (commented):
Question: For the `while` loop, what happens if the condition can only be determined by a device synchronization? (Say, `while x>0:` with `x` being a 0d array.)

asi1024:
@leofang Device synchronization is out of support scope of `cupy.vectorize`. `vectorize` is for element-wise calculation. The input arguments are copied to each local variable at the top of the global function, and the results are written back from the local variable to the global memory at the end of the function.
We will support RawKernel construction from CuPy JIT in the near future, and device synchronization will be supported in it.

asi1024:
Found a bug and fixed it in 6cdb394.

tests/cupy_tests/functional_tests/test_vectorize.py
asi1024: Added a test: 4124f3d

emcastillo (approved):

emcastillo:
Jenkins, test this please

emcastillo:
CI unrelated

PR #4682, opened 2021-02-18
Revert `_get_arch`

takagi:
This PR partially reverts the change made in https://github.com/cupy/cupy/pull/4558, where `_get_arch` was simplified.

`_get_arch` was originally for the case that nvrtc supported only lower compute capability than nvcc, but at the same time, it also worked for the situation that the compute capability that the CUDA Toolkit version supported was lower than the GPU did.

The situation may have to be always an error as, for example, one should not expect to run Ampere on CUDA 10.x, but we keep working user's environment here. (Actually, we have to build CUDA 9.0 wheels on Tesla T4 (CC = 7.5))

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -117,10 +117,30 @@ def _get_nvrtc_version():
     return _nvrtc_version
 
 
+# Known archs for Tegra/Jetson/Xavier/etc
+_tegra_archs = ('53', '62', '72')
+
+
 @_util.memoize(for_each_device=True)
 def _get_arch():
+    # See Supported Compile Options section of NVRTC User Guide for
+    # the maximum value allowed for `--gpu-architecture`.
+    major, minor = _get_nvrtc_version()
+    if major < 10 or (major == 10 and minor == 0):
+        # CUDA 9.x / 10.0
+        _nvrtc_max_compute_capability = '70'
+    elif major < 11:
+        # CUDA 10.1 / 10.2
+        _nvrtc_max_compute_capability = '75'
+    else:
+        # CUDA 11.0 / 11.1
+        _nvrtc_max_compute_capability = '80'
+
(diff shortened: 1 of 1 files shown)


kmaehashi:
LGTM!
pfnCI, test this please.

leofang:
Thanks @takagi.

leofang:
@toslunar @kmaehashi The failing tests are also frequently seen when testing #4628 (thus I suspect it might have something to do with #4684):
```
tests.cupy_tests.random_tests.test_generator.TestInterval.test_goodness_of_fit
tests.cupy_tests.random_tests.test_generator.TestChoiceChi.test_goodness_of_fit
tests.cupy_tests.random_tests.test_generator.TestChoiceChi.test_goodness_of_fit_2
tests.cupy_tests.random_tests.test_sample.TestRandint2.test_goodness_of_fit
tests.cupy_tests.random_tests.test_sample.TestRandint2.test_goodness_of_fit_2
tests.cupy_tests.random_tests.test_sample.TestRandomIntegers2.test_goodness_of_fit
tests.cupy_tests.random_tests.test_sample.TestRandomIntegers2.test_goodness_of_fit_2
```
https://jenkins.preferred.jp/job/chainer/job/cupy_pr/3058/

kmaehashi:
Let me merge this as 
As for the test failures, we're discussing reverting batched SVD PR to isolate the issue. (as the next release date is close)

leofang:
> As for the test failures, we're discussing reverting batched SVD PR to isolate the issue. (as the next release date is close)

@kmaehashi All failures are from the  `test_svd_rank<N>_loop` tests. Perhaps it's easier to just disable this call path https://github.com/cupy/cupy/blob/1976a612043e63d149b9ed08e622220f54ed7f3d/cupy/linalg/_decomposition.py#L398-L402
and also those tests?

PR #4709, opened 2021-02-23
Bump CUDA version for Docker images

### LINKED ISSUE #4593: Excessive malloc time on Nvidia Ampere Arch
We recently upgraded our servers from V100s to A100s and encountered a lengthy initialization cupy.cuda.runtime.malloc issue. Narrowing down the problem we were able to reproduce it with:

`docker run -it --rm cupy/cupy:v8.0.0 python3 -c 'import time, cupy; start=time.time(); ptr=cupy.cuda.runtime.malloc(1); end=time.time(); print(end-start); cupy.cuda.runtime.free(ptr)'`

This took 67 seconds to malloc a single byte. However, successive mallocs within the same process took fractional seconds:

`docker run -it --rm cupy/cupy:v8.0.0 python3 -c 'import time, cupy; start=time.time(); ptr=cupy.cuda.runtime.malloc(1); end=time.time(); print(end-start); cupy.cuda.runtime.free(ptr); start=time.time(); ptr=cupy.cuda.runtime.malloc(1); end=time.time(); print(end-start); cupy.cuda.runtime.free(ptr)'`

Prints 67 seconds followed by 0.15 seconds. We put the V100s back into the same server and ran the same docker command and it resulted in 0.15s always. We have also tried to malloc different amounts, but whether it is 1 byte or multiple gigs, the first initialization is always roughly 67 seconds. Most of our processes do not last that long so the initial delay is doubling or tripling our runtimes.

We have contacted Nvidia and they were able to reproduce the same lengthy cupy.cuda.runtime.malloc on Ampere A100 and RTX A6000, with no such delay on Volta and Turing (they tested cuPy on A100, RTX A6000, GV100, & T4).

We plan on additional testing when time permits, but were wondering if this is something already known by cuPy (a quick search for Ampere and/or malloc delay returned nothing), and if there are additional commands or configurations we can try to help debug this problem once we have the hardware setup again.

kmaehashi:
Closes #4593.

We may bump to more recent one in v9, if the newer CUDA version is released before CuPy v9rc release.

### DIFF
--- a/docker/python3/Dockerfile
+++ b/docker/python3/Dockerfile
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:10.2-cudnn7-devel
+FROM nvidia/cuda:11.1-cudnn7-devel
 
 RUN apt-get update -y && \
     apt-get install -y --no-install-recommends \
@@ -9,4 +9,4 @@ RUN apt-get update -y && \
     rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
 
 RUN pip3 install --no-cache-dir -U install setuptools pip
-RUN pip3 install --no-cache-dir cupy-cuda102==9.0.0b2 scipy optuna
+RUN pip3 install --no-cache-dir cupy-cuda111==9.0.0b2 scipy optuna


emcastillo (approved):
LGTM
no need to trigger CIs

PR #4720, opened 2021-02-23
Remove unneeded environment variable from ROCm install guide

kmaehashi:


leofang:
Is it because we set it here?
https://github.com/cupy/cupy/blob/0d103ab79708d620642d0aa13253b211ef219d51/cupy_setup_build.py#L597

kmaehashi:
I guess the documentation was wrong from the start.
It seems there are no modules using `__HIP_PLATFORM_HCC__` as an environment variable. Moreover `export __HIP_PLATFORM_HCC__` (without value) does nothing.

leofang:
Yeah I didn't notice it's not recommended to be set to 1, and indeed it's only needed by compilers (as a compile time flag) to select the code path.

emcastillo (approved):

emcastillo:
Change looks correct in the docs. Merge without jenkins

PR #4721, opened 2021-02-23
Document that only x86_64 wheels are provided

kmaehashi:
https://github.com/cupy/cupy/issues/4663#issuecomment-781807982


emcastillo (approved):

emcastillo:
Change is ok in the RTD, merging without jenkins.

PR #4713, opened 2021-02-23
Fix thrust compilation in MSVC 14 for CUDA 11.2

emcastillo:
The headers changed to have `constexpr`

PR #4730, opened 2021-02-24
 Missing `constexpr` in `cupy_thrust.cu`

emcastillo:
Windows ate it, but linux doesn't.
also when declaring some functions as constexpr other stuff started to blow up.
I built it locally and confirmed that now cuda 11.2 can be built.


emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

kmaehashi:
For record: this is a follow-up for #4713

PR #4690, opened 2021-02-19
Fix random errors in batched SVD

leofang:
Close #4684. Close #4687.

Fix 3 bugs:
- The workspace size is not large enough
- The pointer address should be computed differently based on `full_matrices`
- Force C++11 compilation to work with older gcc (like gcc 5)

Remove the skip on CUDA 10.0 to force the CI to test it out, as gesvd shouldn't be affected by the known cuSOLVER bug.

PR #4731, opened 2021-02-24
Improve batched SVD

leofang:
Follow-up of https://github.com/cupy/cupy/pull/4690#discussion_r579984220. Overwrites the intermediate array to reduce memory footprint, which seems to also perform faster.

### DIFF
--- a/cupy/cusolver.pyx
+++ b/cupy/cusolver.pyx
@@ -304,12 +304,13 @@ cpdef _gesvd_batched(a, a_dtype, full_matrices, compute_uv, overwrite_a):
             vt = x[..., :n]
             job_u = b'A'
             job_vt = b'O'
+            u_ptr, vt_ptr = u.data.ptr, 0
         else:
             u = x
             vt = _ndarray_init((batch_size, k, n), a_dtype)
             job_u = b'O'
             job_vt = b'S'
-        u_ptr, vt_ptr = u.data.ptr, vt.data.ptr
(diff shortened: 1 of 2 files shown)


leofang:
Will experiment with a small pool of `(handle, stream)` and see if it improves performance or not.

leofang:
> Will experiment with a small pool of `(handle, stream)` and see if it improves performance or not.

My preliminary conclusion is it does not help. In each `gesvd` call there are many small kernels launched and a few host-device transfers involved, so all these bring in non-negligible overheads that cannot be eliminated by simple stream parallelism. Only when cuSOLVER supports stream capture (#4567) would these be mitigated I am afraid...

toslunar:
link to the issue (on non-batched SVD): #3327

leofang:
Before this PR: 
```
svd                 :    CPU:967495.508 us   +/-387.404 (min:967270.049 / max:968551.172) us     GPU-0:967618.262 us   +/-388.642 (min:967400.391 / max:968675.598) us
```
after this PR:
```
svd                 :    CPU:957821.612 us   +/-3866.952 (min:953878.185 / max:962332.340) us     GPU-0:957949.573 us   +/-3870.807 (min:954001.221 / max:962460.022) us
```
input array: `cp.random.random((256, 64, 64))`

leofang:
This is ready.

toslunar:
pfnCI, test this please.

toslunar (approved):

leofang:
Thanks, @toslunar!

Issue #130, opened 2017-06-12
Cython-level interface


jmschrei:
Howdy!

I'm currently trying to add CuPy support to one of my packages, and it would be very useful if there was a cython level interface. Since I'd mostly like to replace BLAS with GPU support, I was wondering if there were any plans to support a cython level API similar to the API that scipy supports for BLAS. Thanks!

jekbradbury:
This is the low-level Cython cuBLAS wrapper https://github.com/cupy/cupy/blob/master/cupy/cuda/cublas.pyx -- it’s mostly used towards the bottom of https://github.com/cupy/cupy/blob/master/cupy/core/core.pyx to implement `cupy.dot`, `cupy.tensordot`, and `cupy.matmul` but I think it’s still exported.

yuyu2172:
@jmschrei 

As @jekbradbury pointed out, cuBLAS wrapper functions are exported at Cython level.
However, many of the NumPy compatible functions are either not `cpdef` functions or `cpdef` functions not exported outside.

I played around a bit, and it was relatively simple to add `cpdef` functions to `pxd` files. By doing that, projects outside of CuPy can call them at Cython level.
I created a simple demo project that demonstrates such usage.
https://github.com/yuyu2172/cupy-cython

jmschrei:
Howdy

Thanks for the responses. I've been out for a bit, but I'm looking into this more. When I said I wanted a cython function, I meant that I was interested in a drop-in replacement for dgemm. Essentially, this is what my code looks like now in a function that has the GIL released:

```python
if GPU[0] == 1:
	with gil:
		x = ndarray_wrap_cpointer(X, n*d).reshape(n, d)
		dot_ndarray = cupy.dot(x, self.inv_cov)
		dot = <double*> (<numpy.ndarray> dot_ndarray).data
else:
	dot = <double*> calloc(n*d, sizeof(double))
	mdot(X, self._inv_cov, dot, n, d, d)
```

What I would ideally like is something like:

```python
if GPU[0] == 1:
    gpu_mdot(X, self.inv_cov, dot, n, d, d)
else:
    mdot(X, self.inv_cov, dot, n, d, d)
```

In my code, mdot is a wrapper for dgemm, and I'd like for gpu_mdot to also be some wrapper of cublasDgemm. I'm unsure what to pass in for the handle though. Can you point m to the appropriate usage of that?

okuta:
This issue relates #79.
I agree about open Cython-level interface for users.
[...]

Issue #79, opened 2017-05-31
Install .pxd files


kmaehashi:
[CuPy Overview](http://docs.cupy.chainer.org/en/latest/overview.html) says:

> CuPy also includes following features for performance:
> * Customizable memory allocator, and a simple memory pool as an example

When writing a custom memory allocator from scratch, I want to reuse or inherit some classes in CuPy (``cupy.cuda.memory.Memory``, ``cupy.cuda.memory.MemoryPointer``).  However ``*.pxd`` files are not installed along with the module by ``pip install cupy``, so cupy modules cannot be cimported from my Cython module.  Currently I have to copy ``*.pxd`` files from CuPy source tree to my project and specify the path to ``include_path`` option when building my module using Cython.

So it would be nice if CuPy installs ``*.pxd`` files to make it cimport-able from other packages.  I think this is possible by using ``package_data`` option as mentioned in the Cython docs:

* http://docs.cython.org/en/latest/src/userguide/sharing_declarations.html#search-paths-for-definition-files
* https://stackoverflow.com/questions/36853975/can-i-pip-install-a-cython-module-and-make-its-pxds-available-for-cimport

okuta:
Thank you for your comment.
Current CuPy does not provide Cyhton interface ( `*.pxd` file).
Please implement by Python. Maybe, it is enough in most case.

Do you have some special reason to use Cython?

kmaehashi:
I'm using Cython for performance.  Internal testing (micro benchmark to malloc/free memory blocks) shows that my memory allocator implemented in Cython was 5x faster than one implemented in pure-python.

Just to make sure, this issue is not urgent for me as I have a workaround as noted above.

okuta:
I close this issue. We will discuss this issue in #130.

PR #4734, opened 2021-02-24
Remove `pyproject.toml`

### LINKED ISSUE #4727: `pip install` completely ignores existing source builds and installed dependencies
I suspect this has to do with #4619. 

I am on the latest master, and now every time I call `pip install -v -e .` two things happens:
1. These packages keeps being reinstalled despite I already have them in my env: setuptools, wheel, Cython, fastrlock
2. All modules are re-cythonized and recompiled from scratch, despite they've been built and nothing has changed 

I will try to build in a fresh env to see if something is wrong with my current env. But it's better to be confirmed independently.

cc: @kmaehashi @emcastillo

kmaehashi:
Closes #4727

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ for submodule in ('cupy/core/include/cupy/cub/',
 
 
 requirements = {
-    # TODO(kmaehashi): migrate to pyproject.yaml (see #4727, #4619)
+    # TODO(kmaehashi): migrate to pyproject.toml (see #4727, #4619)
     'setup': [
         'Cython>=0.28.0',
         'fastrlock>=0.5',
--- a/setup.py
+++ b/setup.py
@@ -24,8 +24,7 @@ for submodule in ('cupy/core/include/cupy/cub/',
 
 
 requirements = {
-    # setup_requires remains here for pip v18 or earlier.
-    # Keep in sync with pyproject.yaml.
+    # TODO(kmaehashi): migrate to pyproject.yaml (see #4727, #4619)
     'setup': [
         'Cython>=0.28.0',
         'fastrlock>=0.5',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,0 @@
-[build-system]
-requires = ["setuptools", "wheel", "Cython>=0.28.0", "fastrlock>=0.5"]


setup.py
```diff
@@ -24,8 +24,7 @@
 
 
 requirements = {
-    # setup_requires remains here for pip v18 or earlier.
-    # Keep in sync with pyproject.yaml.
+    # TODO(kmaehashi): migrate to pyproject.yaml (see #4727, #4619)
```
toslunar, line 27:
```suggestion
    # TODO(kmaehashi): migrate to pyproject.toml (see #4727, #4619)
```

kmaehashi: Nice catch!

toslunar (approved):

toslunar:
pfnCI, test this please.

leofang:
Thanks for handling my last minute request @kmaehashi 🙏

leofang:
FFT callback tests failed?! Is it new?

kmaehashi:
`cupy-py36` now changed to use CUDA 11.2 https://github.com/chainer/chainer-test/pull/635 so that can be a reason...

leofang:
Hmmm I thought I checked with 11.2. Let me recheck.

leofang:
Tests passed fine locally, but I'm in a dev env. Based on the error log it looks like some Cython files are missing so cythonize fails. I saw other CI runs also failed with callbacks, so it's probably not due to removing `pyproject.toml`.

kmaehashi:
The reason for these failures is coming from the design of cufft callback. CuPy installation does not contain pxd files although `cufft.pxd` is `cimport`ing `memory.pxd`, that's why Cython errors are raised.
Previously, chainer-test was testing against CuPy installed via `pip install -e` , so all pxd files were visible. However, https://github.com/chainer/chainer-test/pull/634 changed tests to runs against CuPy installed to site-packages, so pxd files are now invisible from the test script.

CUDA 11.2 is not related to these failure. For CUDA 11.1 or earlier, no cythonize is running on Jenkins because `callback_cache` is already generated.

I think it's better to skip all cufft callback tests for this week's release, and discuss what we can do to support cufft callback for the March release.

leofang:
Thanks, that makes sense, though I was under the impression that we included *.pxd/*.pyx in both sdist and wheel, so shouldn't it just work? I could be wrong though.

kmaehashi:
> we included *.pxd/*.pyx in both sdist and wheel, so shouldn't it just work?

Technically, yes. But once we do that people will start using Cython API from external libraries (e.g. #79, #130), in that case, we have to be very very careful to avoid breaking the compatibility because Cython API breakage will cause the library import to fail immediately. Maintaining Cython API is much costly compared to Python API. That's why we don't expose Cython level API since the initial release.

leofang:
> Maintaining Cython API is much costly compared to Python API. That's why we don't expose Cython level API since the initial release.

I agree. What if we explicitly state in the docs etc that the Cython API is meant to be private and we do not guarantee stability across even minor releases? I think it'd also help address general questions.

I am testing this locally. I think it's easiest if we could include the following `.pxd` files:
```
    'cupy_backends/cuda/api/__init__.pxd',
    'cupy_backends/cuda/api/driver.pxd',
    'cupy_backends/cuda/api/runtime.pxd',
    'cupy/cuda/__init__.pxd',
    'cupy/cuda/memory.pxd',
    'cupy/cuda/stream.pxd',
```

kmaehashi:
@toslunar Could you proceed to merge this PR? The test failures are irrelevant.

PR #4735, opened 2021-02-24
Disable some `constexpr` only for windows

emcastillo:
In linux having all with `constexpr` works (otherwise, it fails)
In windows, if these are declared as `constexpr` compilation fails :D.

Will revisit after release.


kmaehashi:
pfnCI, test this please.

kmaehashi:
Merging as test failures are unrelated and Windows build succeeded.

kmaehashi:
For record: this is a follow-up for #4713 and #4730

PR #4739, opened 2021-02-24
Add missing attrs

leofang:
Fix a likely merge error.

PR #4743, opened 2021-02-24
Fix cuFFT callback compilations

leofang:
Address https://github.com/cupy/cupy/pull/4734#issuecomment-785392652.

This is proposal v1, and the solution is to include a few more `.pxd` files in sdist/wheel needed for compiling the callback modules.

Proposal v2 is to turn any `cimport` to `import` in the codepath, so that we only compile `cufft.pyx` and use only Python functions. I will try to put up a PR asap, but if this one is acceptable than please let me know.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,13 +94,20 @@ tests_require = requirements['test']
 # - Files only needed in sdist should be added to `MANIFEST.in`.
 # - The following glob (`**`) ignores items starting with `.`.
 cupy_package_data = [
+    'cupy_backends/cuda/api/__init__.pxd',  # for cuFFT callback
+    'cupy_backends/cuda/api/driver.pxd',  # for cuFFT callback
+    'cupy_backends/cuda/api/runtime.pxd',  # for cuFFT callback
     'cupy/cuda/cupy_thrust.cu',
     'cupy/cuda/cupy_cub.cu',
+    'cupy/cuda/__init__.pxd',  # for cuFFT callback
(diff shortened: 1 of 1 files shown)


leofang:
Jenkins, test this please

leofang:
@kmaehashi Failed tests are irrelevant of callbacks and will be fixed in #4739 💪

emcastillo:
Jenkins, test this please
(just wanna see all ok in Jenkins)

leofang:
I don't think we need to backport this as cufft callbacks are v9 only...?

emcastillo:
> Proposal v2 is to turn any cimport to import in the codepath, so that we only compile cufft.pyx and use only Python functions. I will try to put up a PR asap, but if this one is acceptable than please let me know.

I believe that if there is not a significant performance impact, this second approach is better.

emcastillo:
We decided to skip the tests until we can find a clean solution #4745

kmaehashi:
@leofang Thanks for the fix, but I think we need more discussion if we include pxd. Considering this is an already-released feature, not a regression in this release, we decided to skip tests for now.

> Proposal v2 is to turn any cimport to import in the codepath, so that we only compile cufft.pyx and use only Python functions. I will try to put up a PR asap, but if this one is acceptable than please let me know.

Downsides of this idea is that, in addition to performance impact, this we have to copy some definitions (enum etc.) in `cufft.pxd` to `cufft.pyx`.

Let's discuss further in #4746 🙏🏼

PR #4772, opened 2021-02-28
Support int argument for Dirichlet shape

aryamccarthy:
Currently, cupy's Dirichlet distribution does not support an integer argument a a shape. Only tuples are allowed; ints lead to a `TypeError`. This differs from the behavior of numpy and is not a necessary restriction. It also does not match the [documented parameters](https://docs.cupy.dev/en/stable/reference/generated/cupy.random.dirichlet.html).

The issue has been [detailed on Stack Overflow recently](https://stackoverflow.com/a/66405505/7802200). A workaround is for the user to provide the size as a tuple, but they shouldn't have to.

I address this by turning the provided argument into a list. (If it's already a list or tuple, this won't affect anything.) I provide a test case as well.

Note that cupy's multivariate normal distribution supports both integer and tuple arguments.

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -143,7 +143,7 @@ class RandomState(object):
         if size is None:
             size = alpha.shape
         elif isinstance(size, (int, cupy.integer)):
-            size = [size]
+            size = [size] + alpha.shape
         else:
             size += alpha.shape
         y = cupy.empty(shape=size, dtype=dtype)
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -143,6 +143,7 @@ class RandomState(object):
         if size is None:
             size = alpha.shape
         else:
+            size = list(size)  # In case size is an int
             size += alpha.shape
         y = cupy.empty(shape=size, dtype=dtype)
         _kernels.standard_gamma_kernel(alpha, self._rk_seed, y)
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -143,7 +143,7 @@ class RandomState(object):
         if size is None:
             size = alpha.shape
         elif isinstance(size, (int, cupy.integer)):
-            size = [size] + alpha.shape
+            size = (size,) + alpha.shape
         else:
             size += alpha.shape
         y = cupy.empty(shape=size, dtype=dtype)
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -142,8 +142,9 @@ class RandomState(object):
         alpha = cupy.asarray(alpha)
         if size is None:
             size = alpha.shape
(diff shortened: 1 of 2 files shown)


DimitrisMel (dismissed):

cupy/random/_generator.py
```diff
@@ -143,6 +143,7 @@ def dirichlet(self, alpha, size=None, dtype=float):
         if size is None:
             size = alpha.shape
         else:
+            size = list(size)  # In case size is an int
```
toslunar, line 146: `list(5)` causes `TypeError: 'int' object is not iterable`.  How about aligning with this?

https://github.com/cupy/cupy/blob/79085bcdb5ebfecf906a8bc4453dfa0f85bbddb4/cupy/random/_generator.py#L660-L661

```suggestion
            if isinstance(size, int):
                size = size,
```

toslunar (changes requested):

cupy/random/_generator.py
aryamccarthy: Can do even better by matching multivariate_normal, supporting both `int` and `cupy.integer`. Just pushed that fix.

Why isn't my test case being run, to catch this error?

cupy/random/_generator.py
```diff
@@ -142,6 +142,8 @@ def dirichlet(self, alpha, size=None, dtype=float):
         alpha = cupy.asarray(alpha)
         if size is None:
             size = alpha.shape
+        elif isinstance(size, (int, cupy.integer)):
+            size = [size]
```
toslunar, line 146:
```suggestion
            size = [size] + alpha.shape
```

toslunar (approved):

toslunar:
pfnCI, test this please.

cupy/random/_generator.py
leofang: > Why isn't my test case being run, to catch this error?

It's best to test locally. Tests are not run until a reviewer launches CuPy's CI. The GH action etc just test for building and formatting.

cupy/random/_generator.py
```diff
@@ -142,6 +142,8 @@ def dirichlet(self, alpha, size=None, dtype=float):
         alpha = cupy.asarray(alpha)
         if size is None:
             size = alpha.shape
+        elif isinstance(size, (int, cupy.integer)):
+            size = [size] + alpha.shape
```
toslunar, line 146:
```suggestion
            size = (size,) + alpha.shape
```

toslunar:
pfnCI, test this please.

toslunar:
pfnCI, test this please.

toslunar:
pfnCI, test this please.

toslunar (approved):

toslunar:
Thanks, @aryamccarthy, @DimitrisMel.

Issue #3684, opened 2020-07-29
Kernel Occupancy Functions


coderforlife:
CUDA 6.5 introduced several useful functions for computing optimal occupancy. See https://developer.nvidia.com/blog/cuda-pro-tip-occupancy-api-simplifies-launch-configuration/ for details. But as a first pass it would be nice if `RawKernel`s provided access to these functions (like all of the wrappers around `funcGetAttribute` are there, in fact these use those attributes plus device attributes to compute their values):

* `cudaOccupancyMaxActiveBlocksPerMultiprocessor(int* numBlocks, T func, int blockSize, size_t dynamicSMemSize)` (+`WithFlags` version)
* `cudaOccupancyMaxPotentialBlockSize(int* minGridSize, int* blockSize, T func, size_t dynamicSMemSize=0, int blockSizeLimit=0)` (+`WithFlags` version)
* `cudaOccupancyMaxPotentialBlockSizeVariableSMem(int* minGridSize, int* blockSize, T func, UnaryFunction blockSizeToDynamicSMemSize, int blockSizeLimit=0)` (+`WithFlags` version)
* `cudaOccupancyAvailableDynamicSMemPerBlock(size_t* dynamicSmemSize, T func, int  numBlocks, int blockSize)`

Several of these could be combined into a single Python method that determines which one based on the parameters. Additionally they could be used to provide decent default values for `RawKernels` (and many other kernels probably). The only flag for any of these is `cudaOccupancyDisableCachingOverride`.

For `RawKernel` I also imagine a cleaner interface for calling them. I like the below method because it reflects the C++ way of calling a kernel (just uses `[ ]` instead of `<<< >>>`).

```python
kernel[grid, block](arg1, arg2, arg3)  # grid and block can be int OR tuple
kernel[n](arg1, arg2, arg3)  # compute grid and block based on the total number and occupancy
# This is a bit weird, but you could specify x,y,z values using slices:
kernel[5:5:5, 3:3:3](arg1, arg2, arg3)
kernel[5:5](arg1, arg2, arg3)  # computes grid and block based on the total number and occupancy
# the last one is not possible without slices since tuples are already used for the first style
```

The original `kernel(grid, block, args)` interface could be kept for backwards compatibility and the new one could just call it:

```python
def __getitem__(self, gb):
    if isinstance(gb, tuple):
        grid, block = gb
        if isinstance(grid, slice):
            grid = _convert_slice(grid)
        elif not isinstance(grid, tuple):
[...]

PR #4774, opened 2021-03-01
Add CuPy JIT Kernel definition

asi1024:
Part of #4290.

This PR supports a raw kernel interface like as following.
```py
from cupyx import jit

@jit.rawkernel()
def f(x, y, z, n):
    tid = jit.threadIdx.x + jit.blockIdx.x * jit.blockDim.x
    ntid = jit.blockDim.x * jit.gridDim.x
    for i in range(tid, n, ntid):
        z[i] = x[i] + y[i]

n = numpy.uint32(1024)
x = cupy.arange(n)
y = cupy.arange(n)
z = cupy.empty((n,), dtype='l')
f[16, 16](x, y, z, n)
```

### DIFF
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -312,11 +312,15 @@ def _transpile_stmt(stmt, is_toplevel, env):
     if isinstance(stmt, ast.Assign):
         if len(stmt.targets) != 1:
             raise NotImplementedError('Not implemented.')
-        target = stmt.targets[0]
-        if not isinstance(target, ast.Name):
-            raise NotImplementedError('Tuple is not supported.')
-        name = target.id
+
         value = _transpile_expr(stmt.value, env)
+        target = stmt.targets[0]
+
+        if not isinstance(target, ast.Name):
+            target = _transpile_expr(target, env)
+            return [f'{target.code} = {value.code};']
+
+        name = target.id
 
         if is_constants([value]):
             if not isinstance(value.obj, _typeclasses):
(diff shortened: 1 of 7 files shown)


cupyx/jit/_compile.py
```diff
@@ -519,12 +547,21 @@ def _transpile_expr(expr, env):
...
+        index = _transpile_expr(expr.slice, env)
+        if is_constants([value, index]):
+            return Constant(value[index])
+        value = _to_cuda_object(value, env)
+        index = _to_cuda_object(index, env)
+
+        if not isinstance(value.ctype, _types.Array):
+            raise ValueError(f'{value.code} must be Array type.')
+        if value.ctype.ndim != 1:
+            raise NotImplementedError('Not implemented for ndim > 1.')
```
asi1024, line 560: Currently, only 1-dim arrays are supported, but we are planning to support also multi-dimension arrays in the next PR.

tests/cupyx_tests/jit_tests/test_raw.py
```diff
@@ -0,0 +1,56 @@
...
+        @jit.rawkernel()
+        def f(x, y, size):
+            tid = jit.threadIdx.x + jit.blockDim.x * jit.blockIdx.x
+            ntid = jit.blockDim.x * jit.gridDim.x
+            for i in range(tid, size, ntid):
+                y[i] = x[i]
+
+        x = testing.shaped_random((1024,), dtype=numpy.int32, seed=0)
+        y = testing.shaped_random((1024,), dtype=numpy.int32, seed=1)
+        f[5, 6](x, y, numpy.uint32(1024))
```
asi1024, line 55: `tid` and `ntid` is of `uint32` datatype and `range` requires arguments of the same datatype. So, we need to cast `size` to uint32 here. Is it inconvenient?

leofang:
Question: is there a plan to support dynamic allocated shared memory & explicit stream argument? Like
```python
f[16, 16, 256, s](x, y, z, n)  # allocating 256 bytes of shared memory, run the kernel on stream s
```
This is more or less in line with #3684.

tests/cupyx_tests/jit_tests/test_raw.py
emcastillo: I think its reasonable.

tests/cupyx_tests/jit_tests/test_raw.py
```diff
@@ -0,0 +1,56 @@
+import unittest
+import numpy
+
+import cupy
+from cupyx import jit
+from cupy import testing
+
+
+class TestRaw(unittest.TestCase):
```
emcastillo, line 9: let's remove unittest as discussed yesterday and it is good to go :)

emcastillo:
Jenkins, test this please

emcastillo (approved):

emcastillo:
Jenkins, test this please

jakirkham:
Very cool! Thanks for working on this. Looking forward to trying CuPy JIT 😄

asi1024:
@leofang Yes, we are planning to support dynamic allocated shared memory and explicit stream argument, but their interface is still under discussion.

PR #4792, opened 2021-03-02
Remove solved xfail mark for msort

takagi:
This PR removes an `xfail` mark as it is already fixed and leaves XPASS on pytest output.


takagi:
pfnCI, test this please.

emcastillo (approved):

emcastillo:
CIs unrelated

PR #4805, opened 2021-03-03
Windows: Fix csc/csr argmax/argmin tests

leofang:
CuPy always returns in `int64`, but on Windows the default int type is `int32`.

```
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_2_{axis=-2, dense=False}::test_argmax
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_2_{axis=-2, dense=False}::test_argmin
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_3_{axis=-2, dense=True}::test_argmax
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_3_{axis=-2, dense=True}::test_argmin
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_4_{axis=-1, dense=False}::test_argmax
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_4_{axis=-1, dense=False}::test_argmin
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_5_{axis=-1, dense=True}::test_argmax
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_5_{axis=-1, dense=True}::test_argmin
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_6_{axis=0, dense=False}::test_argmax
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_6_{axis=0, dense=False}::test_argmin
FAILED tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py::TestCscMatrixScipyCompressedMinMax_param_7_{axis=0, dense=True}::test_argmax
[...]

PR #4813, opened 2021-03-04
Improve/fix csc/csr argmax/argmin 

leofang:
Addressed issues:
1. For argmin/argmax, the precision of input floats does not matter since they can be compared just fine, so it doesn't make sense to always cast to `double` (`float64`).
2. However, the output int type is platform dependent (`int64` on Linux, `int32` on Windows), so we want to fix it properly to eliminate the Windows errors (see #4805).

The solution to fix both issues at once is to do template specializations using `RawModule`. This is both an improvement ***and*** a bug fix.

Close #4805.

### DIFF
--- a/cupyx/scipy/sparse/compressed.py
+++ b/cupyx/scipy/sparse/compressed.py
@@ -1052,21 +1052,16 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
             if ufunc == cupy.argmax:
                 ker_name = 'max' + ker_name
                 ker = self._max_arg_reduction_mod.get_function(ker_name)
-                ker((self.shape[0],), (1,),
-                    (self.data, self.indices,
-                     self.indptr[:len(self.indptr) - 1],
-                     self.indptr[1:], cupy.int64(self.shape[1]),
-                     value))
-            if ufunc == cupy.argmin:
+            elif ufunc == cupy.argmin:
                 ker_name = 'min' + ker_name
                 ker = self._min_arg_reduction_mod.get_function(ker_name)
-                ker((self.shape[0],), (1,),
-                    (self.data, self.indices,
-                     self.indptr[:len(self.indptr) - 1],
-                     self.indptr[1:], cupy.int64(self.shape[1]),
-                     value))
+            ker((self.shape[0],), (1,),
+                (self.data, self.indices,
+                 self.indptr[:len(self.indptr) - 1],
+                 self.indptr[1:], cupy.int64(self.shape[1]),
+                 value))
 
-        if axis == 0:
+        elif axis == 0:
             # Create the vector to hold output
             # Note: it's important to set "int" here, following what SciPy
             # does, as the outcome dtype is platform dependent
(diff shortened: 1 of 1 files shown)


leofang:
Jenkins, test this please

leofang:
This is ready!

cupyx/scipy/sparse/compressed.py
```diff
@@ -1026,40 +1041,54 @@ def _arg_minor_reduce(self, ufunc, axis):
...
                      self.indptr[:len(self.indptr) - 1],
                      self.indptr[1:], cupy.int64(self.shape[0]),
                      value))
             if ufunc == cupy.argmin:
-                self._min_arg_reduction_kern(
-                    (self.shape[1],), (1,),
-                    (self.data.astype(cupy.float64), self.indices,
+                ker_name = 'min' + ker_name
+                ker = self._min_arg_reduction_mod.get_function(ker_name)
+                ker((self.shape[1],), (1,),
```
emcastillo, line 1090: these calls are duplicated, we can just move this one out of the if and delete the above one :)

cupyx/scipy/sparse/compressed.py
```diff
@@ -1026,40 +1041,54 @@ def _arg_minor_reduce(self, ufunc, axis):
...
                      self.indptr[:len(self.indptr) - 1],
                      self.indptr[1:], cupy.int64(self.shape[1]),
                      value))
             if ufunc == cupy.argmin:
-                self._min_arg_reduction_kern(
-                    (self.shape[0],), (1,),
-                    (self.data.astype(cupy.float64), self.indices,
+                ker_name = 'min' + ker_name
+                ker = self._min_arg_reduction_mod.get_function(ker_name)
+                ker((self.shape[0],), (1,),
```
emcastillo, line 1063: dupped calls :), lets take the last one out.

cupyx/scipy/sparse/compressed.py
```diff
@@ -175,10 +176,15 @@ class _compressed_sparse_matrix(sparse_data._data_matrix,
...
-        void max_arg_reduction(double* data, int* indices, int* x, int* y,
-                               int length, long long* z) {
+    # For _max_arg_reduction_mod and _min_arg_reduction_mod below, we pick
+    # the right template specialization according to input dtypes at runtime.
+    # The distinction in int types (T2) is important for portability in OS.
+
+    _max_arg_reduction_mod = core.RawModule(
+        code=r'''
+        template<typename T1, typename T2> __global__ void
+        max_arg_reduction(T1* data, int* indices, int* x, int* y,
```
emcastillo, line 186: Optional: these two kernels are essentially the same but with just a single operator changing.
Can we templatize them? even just a single python f-string template is fine :).

You can ignore this, or do it in another PR if you want!

Thanks

emcastillo (changes requested):
Check comments please

cupyx/scipy/sparse/compressed.py
leofang: Good catch! Done in c7a03be.

cupyx/scipy/sparse/compressed.py
leofang: Done in c7a03be.

cupyx/scipy/sparse/compressed.py
leofang: Yeah I wanted to do this too but for some reason I forgot. Done in 82a6dc0.

leofang:
Jenkins, test this please

leofang:
@emcastillo Windows is still happy 😆

cupyx/scipy/sparse/compressed.py
```diff
@@ -1026,42 +992,46 @@ def _arg_minor_reduce(self, ufunc, axis):
...
+            ker((self.shape[0],), (1,),
+                (self.data, self.indices,
+                 self.indptr[:len(self.indptr) - 1],
+                 self.indptr[1:], cupy.int64(self.shape[1]),
+                 value))
+
+        elif axis == 0:
             # Create the vector to hold output
-            value = cupy.zeros(self.shape[1]).astype(cupy.int64)
+            # Note: it's important to set "int" here, following what SciPy
```
emcastillo, line 1017: Sorry that I did not notice before but these two braces are exactly the same, but one uses `shape[0]` and the other `shape[1]` 😄

leofang: Do you mean you wanna remove the repeated code?

emcastillo: I will do it in another PR. lets get this in

emcastillo (approved):

cupyx/scipy/sparse/compressed.py
leofang: oh oh ok, I can do it too, just wasn't sure what you meant 😅

leofang:
btw @emcastillo I suppose we should backport this to fix v8?

PR #4815, opened 2021-03-05
Complete build only when all of the essential modules are available

### LINKED ISSUE #4750: Fail build if essential module cannot be built
Currently, when CuPy failed to configure a module other than `cuda`, build succeeds without building the module.
If non-optional (essential) modules cannot be built, the build should fail with an error.

Optional modules are nccl, cudnn, nvtx
Non-optional modules are others e.g., random, thrust, cub, dlpack, ...

takagi:
Fix #4750.

Currently, CuPy completes its build even if an essential module is failed to be built. This PR fixes the build process to complete only when we can successfully build all of the required modules. See #4750 for an exhaustive list of modules CuPy depends on mandatorily or optionally.

### DIFF
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -97,6 +97,7 @@ if use_hip:
     MODULES.append({
         # TODO(leofang): call this "rocm" or "hip" to avoid confusion?
         'name': 'cuda',
+        'required': True,
         'file': cuda_files + [
             'cupy.cuda.nvtx',
             'cupy_backends.cuda.libs.cusolver',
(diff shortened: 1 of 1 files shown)


cupy_setup_build.py
```diff
@@ -214,6 +217,7 @@
 
     MODULES.append({
         'name': 'cutensor',
+        'required': True,
```
kmaehashi, line 220: cuTENSOR is an optional. (sorry I failed to list it up in my original issue :bow:_

kmaehashi (changes requested):

cupy_setup_build.py
takagi: Fixed!

kmaehashi (approved):
LGTM!

kmaehashi:
pfnCI, test this please.

PR #4797, opened 2021-03-03
Support `norm='forward'`/`'backward'` in `cupy.fft` functions

leofang:
Take over #3318 with permission from @mattvend.

Close #944. Close #3318.

This PR follows the upstream change included since NumPy 1.20 (numpy/numpy#16476).

PR #4816, opened 2021-03-05
Support `norm='forward'`/`'backward'` in `cupyx.scipy.fft` functions

leofang:
~~Blocked by #4797.~~

This PR follows the upstream change scipy/scipy#12532 included since SciPy 1.6.0.

The support is done in #4797, so really here we only need to update the docstrings (c8c0ecb) and expand tests (e35a940).

The total number of tests inflates to about 10k if none is skipped, but the total runtime is less than 1 min, so I consider it reasonable (cc: @toslunar).


toslunar:
Merged #4797.

leofang:
Jenkins, test this please

toslunar (approved):
LGTM

PR #4829, opened 2021-03-06
Ability to pass structured data types by value as kernel parameters

### LINKED ISSUE #3359: Is there a way to pass a structured buffer to a custom kernel, like arrow cuda buffer or flatbuffers buffer
This isn't actually an issue, just a question and maybe a feature request or at least documentation request, as I didn't find other ways to ask a question, like chat or group. 

I'm looking for a way / example to pass a structured byte buffer to a custom kernel. 

I didn't find any documentation or examples of how to use low level cuda api, for example 

cupy.cuda.MemoryPointer with a custom kernel, ie pass it as an argument, create it from the host and/or device buffer?

I'm looking for an integration with an Arrow and/or Flatbuffers or alike offset based flat structures.

### LINKED ISSUE #4828: Ability to pass structured types by value as kernel parameters
Currently, it is not possible to pass structures **by value** as kernel arguments. Only some fixed size numpy dtype arguments have been hardcoded yet in `cupy/cuda/function.pyx`. This implies that a cupy kernel launch only succeeds with numpy.generic arguments which are scalar types with itemsize 1, 2, 4, 8 or 16 bytes. This allows us to pass directly by value np.int8 up to np.complex128 numpy data types as kernel arguments. I would like to extend pass-by-value to any numpy dtype.

I am using RawKernels extensively and I would like to be able to pass cuda builtin vector types such as `float3` as kernel parameters. This would make multidimensional code in physics much clearer for example (ie. replace many scalar kernel arguments by vectorized arguments or static arrays of scalars). Depending on device compute capability we can now pass up to 4kB of parameters to cuda kernels, and doing it in 16 bytes chunks is not very efficient.

The definition of the structure layout using numpy.dtype should be the responsibility of the user. This off course implies defining the struct with the same offsets, alignment and itemsize between host and device. `pyopencl` and `pycuda` offer utility functions to convert a device side struct to a numpy dtype with matching layout, but this feature would be overkill for now.

Please note that **this feature is not related to general structured array support in cupy**. It is however already possible to pass a pointer to any array of structures to a RawKernel: simply cast any numpy array to np.int8 data type and let cupy handle np.int8 cuda arrays. Within the kernel, the pointer will be cast to the kernel argument type anyway. This off course only works if host and device structures layout match exactly.

Here is a simple example to illustrate what I would be able to do:
```python
import numpy as np
import cupy as cp

code = r"""
    template <typename T>
    __global__ void affect(T* const __restrict__ output,
                         const T arg0, const T arg1,
                         const size_t nelements) {
        ptrdiff_t i = blockIdx.x * blockDim.x + threadIdx.x;
        if (i >= nelements)
            return;
        output[i] = (i%2 == 0 ? arg0 : arg1);
    }
"""
module = cp.RawModule(code=code,
                      name_expressions=('affect<float3>',),
                      options=('-std=c++11',))
kernel = module.get_function('affect<float3>')

float3 = np.dtype( { 'names': ['x', 'y', 'z'],
                     'formats': [np.float32]*3 } )

arg0 = np.random.rand(3).astype(np.float32).view(float3)
arg1 = np.random.rand(3).astype(np.float32).view(float3)

N = 512
h_output = np.empty(dtype=float3, shape=N)
d_output = cp.asarray(h_output.view(np.float32))

grid = (N,)
block = (256,)
args = (d_output, arg0, arg1, np.uint64(N))
kernel(grid, block, args)

h_output[0::2] = arg0
h_output[1::2] = arg1

np.testing.assert_array_equal(h_output,
                              cp.asnumpy(d_output).view(float3))
```

keckj:
This PR is directly related to issue #4828.

Currently, we can only pass by value the following numpy items as kernel parameters: np.int8, np.int16, np.int32, np.int64, np.uint8, np.uint16, np.uint32, np.uint64, np.float16, np.float32, np.float64, np.complex64 and np.comples128 and other numpy builtin datatypes if their itemsize is either 1, 2, 4, 8 or 16 bytes.

This patch aims to enable **pass by value** for any structured data type built with numpy.dtype as kernel parameter.
It enhances current kernel launch capabilities of cupy without changing its interface. It is the responsibility of the user to define a numpy dtype that matches exactly device structure layout.

It provides the ability to:
- pass cuda builtin vector types (char2, double3, int4) as kernel parameters.
- pass "static arrays" of parameters such has `struct double32 { double[32] val };` and other packed structures.
- define and use any cuda composite type as kernel parameter (including structure of structures) such as
 ```cuda
   struct custom_user_struct {
    int4 a;
    char b;
    double c[2];
    short1 d;
    unsigned long long int e[3];
  };
```
Implementation status:
- [x] Patch `cupy/cuda/function.pyx` to enable pass by value to any dtype
- [x] Add one test for each use case in ` tests/cupy_tests/core_tests/test_function.py`: builtin vector type, static array and custom struct
- [x] Document this under `User-Defined Kernels` section
- [x] Throw an error when numpy array argument is not of size 1 (this would be error-prone)
- [x] Add some examples of complex structs 

close #3359, close #4828

### CODE REVIEW

**Comment by emcastillo** [comment]
This is neat! thanks for the contribution :)

**Comment by keckj** [comment]
Everything has been implemented, including tests and documentation.
The cupy kernel lauch interface should not be affected at all by these changes.
This PR closes issue #4828.

**Comment by leofang** [comment]
Thanks for your nice contribution, @keckj! I've skimmed over your PR and am happy to help review, but before I find time to do that, could you apply this patch please? Your PR do not work without it on ROCm due to a bug in hipRTC...
```diff
diff --git a/tests/cupy_tests/core_tests/test_function.py b/tests/cupy_tests/core_tests/test_function.py
index 5f8d71004..f2970e5f2 100644
--- a/tests/cupy_tests/core_tests/test_function.py
+++ b/tests/cupy_tests/core_tests/test_function.py
@@ -4,12 +4,16 @@ import pytest
 import numpy
 
 import cupy
+from cupy.core import core
 from cupy.cuda import compiler
+from cupy.cuda import runtime
 from cupy import testing
 
 
 def _compile_func(kernel_name, code):
-    mod = compiler.compile_with_cache(code)
+    # workaround for hipRTC
+    extra_source = core._get_header_source() if runtime.is_hip else None
+    mod = compiler.compile_with_cache(code, extra_source=extra_source)
     return mod.get_function(kernel_name)
 
 
```

leofang:
Also, you can add this phrase "`close #3359, close #4828`" to your PR description so that these two issues get automatically closed upon merging of this PR 🙂

keckj:
Thanks for this quick feedback, the patch has been applied and the description has been changed to incorporate issues. I just began to use the CuPy module instead of pyCUDA, and I am not yet aware of cupy best practices such as complex numbers handling. Could you check that my additions to the documentation make sense ?

cupy/cuda/function.pyx
```diff
@@ -26,51 +26,13 @@ cdef class CPointer:
...
-    cdef:
-        double complex val
-
-    def __init__(self, double complex v):
-        self.val = v
-        self.ptr = <void*>&self.val
+        int8_t[:] val
 
+    def __init__(self, a):
+        self.val = numpy.atleast_1d(a).ravel().view(numpy.int8)
```
emcastillo, line 34: We are a bit concerned about this,
we want to have kernel launch as fast as possible and calling several full python numpy routine for every argument that we pass can be overkill and bump the launch time from a few microseconds to several hundreds.

I still have to measure this, I expect to do it tomorrow. In the case that the overhead is significant, the solution here will be to leave the old values as they are and just add this one when it is necessary. But this will need some reworking to deal with the array pointer instead of all these calls.

keckj: I am not an expert in cython at all and I don't think val is required here. I think you can directly use `val.ctypes.data`  without defining the memory view but I dit not manage to cast it correctly with the `<void*>` syntax.

On my side `python3 -m timeit -s 'import numpy as np; a=np.float64(12);' -- 'np.atleast_1d(a).ravel().view(np.int8)'` gives me 2.03 usec per loop which is effectively a cost that we should get rid off.

docs/source/tutorial/kernel.rst
```diff
@@ -263,16 +263,128 @@ Accessing texture (surface) memory in :class:`~cupy.RawKernel` is supported via
...
++-----------------+-----------------------------------------------+------------------+
+
+The CUDA standard guarantees that the size of fundamental types on the host and device always match.
+The itemsize of ``size_t``, ``ptrdiff_t``, ``intptr_t``, ``uintptr_t``, 
+``long``, ``signed long`` and ``unsigned long`` are thus platform dependent. 
+You can also pass any CUDA vector builtins such as ``float3`` or any other user defined structure 
+as kernel arguments provided it matches device side kernel parameter type, see section :ref:`custom_user_structs`.
+
+.. note::
+    To use ``cuFloatComplex`` and ``cuDoubleComplex`` in your CUDA kernel, you need to include the header ``cuComplex.h``. You may also need to pass `translate_cucomplex=True` to :class:`~cupy.RawKernel` or :class:`~cupy.RawModule`.
```
leofang, line 325: I would suggest to remove this for the following reasons:
1. For code that uses `cuFloatComplex` etc, users should already know that `cuComplex.h` needs to be included, so here is nothing new.
2. This is already mentioned above around line 233. Looks like largely duplicate to me

docs/source/tutorial/kernel.rst
```diff
@@ -263,16 +263,128 @@ Accessing texture (surface) memory in :class:`~cupy.RawKernel` is supported via
...
+
+.. note::
+    To use ``complex<float>`` and ``complex<double>`` in your CUDA kernel, you simply need to include the header ``<cupy/complex.cuh>``. This is the recommended way to get complex support in :class:`~cupy.RawKernel`.
```
leofang, line 328: I would also suggest to remove this part as it's duplicate.

docs/source/tutorial/kernel.rst
```diff
@@ -263,16 +263,128 @@ Accessing texture (surface) memory in :class:`~cupy.RawKernel` is supported via
...
+
+.. note::
+    To be able to directly use ``std::complex<float>`` and ``std::complex<double>``, you need to include the header ``<cuda/std/complex.h>`` which is provided by `NVIDIA C++ Standard Library` version 1.4.0 and above. Currently **this header is not shipped with CUDA toolkit 11.2**, see this `link <https://nvidia.github.io/libcudacxx/releases.html>`_ to get more information.
```
leofang, line 331: IMHO I would suggest to remove this part for the following reasons:
1. CuPy currently has not included any integration test with libcu++, and while we are keen to the `std::complex` support (due to https://github.com/cupy/cupy/issues/3370) it is better to not recommend this usage until we have a good test coverage. 
2. This is oversimplified: Users not only have to include the header, but also have to set the include path via `options=('-I/path/to/libcu++/header', ...)` as libcu++ is not available in any standard (CUDA or CuPy) path, and perhaps some linker options as well (that I don't know).
3. libcu++'s `std::complex` support was buggy on NVRTC, which was just fixed very recently (https://github.com/NVIDIA/libcudacxx/pull/109), available since 1.4.1. I

In short, I think the current status of libcu++ is more for adventurous users, not average ones. For the former group they don't count on our tutorial here, and for the latter they'd just get confused.

docs/source/tutorial/kernel.rst
```diff
@@ -263,16 +263,128 @@ Accessing texture (surface) memory in :class:`~cupy.RawKernel` is supported via
...
+    >>> import numpy as np
+    >>> float5x5 = np.dtype({'names': ['dummy'], 'formats': [(np.float32,(5,5))]}) 
+    >>> arg = np.random.rand(25).astype(np.float32).view(float5x5)
+    >>> print(arg.itemsize)
+    100
+
+Here ``arg`` represents a 100 bytes scalar (i.e. a numpy array of size 1)
+that can be passed by value to any kernel.
+Upper bound for total kernel parameters size is 256 bytes for compute capability `1.x` (via shared memory)
+and 4kB for compute capability `2.x` and higher (via constant memory).
```
leofang, line 380: CC 1.x is no longer supported by any supported CUDA versions in CuPy, so it's safe to just mention the 4kB limit here. Also, it might be good to add a link to where this info (4kB dedicated constant memory) is available. Note I mention "dedicated" because it's not the normal constant memory space (with qualifier `__constant__`), so it could be a little confusing to just call it constant memory.

docs/source/tutorial/kernel.rst
```diff
@@ -263,16 +263,128 @@ Accessing texture (surface) memory in :class:`~cupy.RawKernel` is supported via
...
+
+Composite types can be built recursively using numpy dtype `offsets` and `itemsize` capabilities,
+see ``cupy/tests/cupy_tests/core_tests/test_function.py`` for examples of advanced usage where structures offsets are determined by calling a cuda kernel.
```
leofang, line 383: I wonder if it's better to showcase a few examples in [`cupy/examples`](https://github.com/cupy/cupy/tree/master/examples) instead of referring users to the test suite (although I also did this at times), which contains some noise from the test framework. But we can also do this in a separate PR IMHO.

leofang (commented):
Sorry that I couldn't reply sooner as promised. I only reviewed the changes in the code and tutorial in detail. I can take a deeper look at the tests later.

keckj:
I fixed the documentation and added back the check for scalar arguments.
I also added three examples in `cupy/examples/custom_struct`.

cupy/cuda/function.pyx
```diff
@@ -90,6 +90,11 @@ cdef class CIntptr(CPointer):
         self.ptr = <void*>&self.val
 
 
+cdef class CNumpyArray(CPointer):
+    def __init__(self, a):
+        self.ptr = <void*><size_t>a.__array_interface__['data'][0]
```
leofang, line 95: I think it's better to hold the reference to the numpy array, in case it goes out of scope (say we pass in a temporary numpy array). Something like
```suggestion
cdef class CNumpyArray(CPointer):
    cdef:
        object val

    def __init__(self, a):
        self.val = a
        self.ptr = <void*><intptr_t>a.ctypes.data
```

leofang: Also, accessing `a.ctypes.data` should be slightly faster.

leofang (changes requested):
Sorry, I have to run and will try to resume tonight, but feel I better post a few drop-by comments before I forget...

Could you fix the signature in line 102 for us? It should be
```python
cdef inline CPointer _pointer(x) except*:
```
This is our oversight: potentially a `TypeError` could be raised (before your great addition), but without the `except*` clause it'd be swallowed silently. Now with your change, it becomes more important in order to show the warning you wrote. You can try verifying this locally.

Thanks a lot! And great to see examples added! Will take a closer look...

cupy/cuda/function.pyx
keckj: (Tests are running, patch will be pushed soon)

1. I fixed the signature. The type error was however already raised before this change with the message (this is checked in `tests/cupy_tests/core_tests/test_function.py:106`). Removing the context manager will trigger the new type error.
2. Hold a reference to the ndarray: done
3. About the performance, `__array_interface__` seems to do slightly better than `ctypes`:

**Array interface:**
```
python3 -m timeit -s 'import numpy as np; a=np.empty((8,8), dtype=np.float32);' -n 100000 -- 'a.__array_interface__["data"][0]'
100000 loops, best of 3: 0.738 usec per loop
```

**Ctypes interface**
```
python3 -m timeit -s 'import numpy as np; a=np.empty((8,8), dtype=np.float32);' -n 100000 -- 'a.ctypes.data'
100000 loops, best of 3: 0.846 usec per loop
```

The only possible improvement would be to use directly `numpy.ndarray.data` (0.165 usec) which returns a memory view. The problem is that I did not manage to obtain a pointer from this python memory view without first declaring a cython memory view with cdef. I did not manage to cdef a generic memory view object (both in dimension and in type) so my initial solution was to:
- viewing the array as a one dimensional array with `numpy.ndarray.ravel`
- viewing it as a "compile time" known type (like `np.uint8`) with `numpy.ndarray.view`
as it can be seen in my first commit.

leofang: > 1. The signature cannot be fixed with your change, cython build fails with
>    `Exception clause not allowed for function returning Python object`.

Thank you, @keckj. Apology, I shouldn't have rushed into comments right before leaving from work. `CPointer` is a `cdef class`, which is indeed a Python object, so there's no problem in propagating any exception.

Re: `__array_interface__`, thanks for testing it again. I just recalled that you already tested it earlier (https://github.com/cupy/cupy/pull/4829#discussion_r589392268) 😅 I think it's fine to leave it as is. Any other option is more complex than this or `a.ctypes.data`.

leofang: > According to my very limited cython knowledge you cannot cdef a generic memory view object (both in dimension and in type)

I think you are right on this. AFAIK Cython's memoryview is only limited to primitive (C) types, so is Python's built-in `array` module.

leofang (changes requested):
Thanks a lot, @keckj! The doc looks great, so are the added examples. Very illustrative and pedagogical. 

I left some comments, which are perhaps a bit annoying but mostly just nitpicks. I also I have one more request: Could you follow `tests/example_tests/test_gemm.py ` and add a new test file, say, `tests/example_tests/test_custom_struct.py` to get your 3 examples tested? I verified locally that it works in an ROCm environment, so the rest can be taken care by the CI (which only runs on CUDA ftm) once it's in.

Thanks again. IMHO this is a great feature that future improvement and enhancements can be built upon, so it's better to get it solidly done.

docs/source/tutorial/kernel.rst
```diff
@@ -263,16 +263,122 @@ Accessing texture (surface) memory in :class:`~cupy.RawKernel` is supported via
...
+named members you can directly define such numpy datatype as the following:
+
+.. doctest::
+
+    >>> import numpy as np
+    >>> names = ['x', 'y', 'z']
+    >>> types = [np.float32]*3
+    >>> float3 = np.dtype({'names': names, 'formats': types})
+    >>> arg = np.random.rand(3).astype(np.float32).view(float3)
+    >>> print(arg)
```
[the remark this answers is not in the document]
leofang: Oh, I think this is better: allocate `arg` using some deterministic function, like `arg = np.arange(...)`, then we can do doc test safely.

docs/source/tutorial/kernel.rst
```diff
@@ -263,16 +263,122 @@ Accessing texture (surface) memory in :class:`~cupy.RawKernel` is supported via
...
+    [(0.9940819, 0.62873816, 0.8953669)]
+    >>> arg['x'] = 42.0
+    >>> print(arg)
```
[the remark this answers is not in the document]
leofang: See https://github.com/cupy/cupy/pull/4829#discussion_r598199158.

examples/custom_struct/builtin_vectors.py
```diff
@@ -0,0 +1,48 @@
...
+    # The kernel computes out = lhs+rhs where lhs and rhs are double3 vectors.
+    # lhs is an array of N such vectors and rhs is double3 kernel parameter.
+
+    lhs = cupy.random.rand(3*N, dtype=numpy.float64).reshape(N, 3)
+    rhs = numpy.random.rand(3).astype(numpy.float64)
+    out = cupy.empty_like(lhs)
+
+    kernel = cupy.RawKernel(code, 'sum_kernel')
+    args = (lhs, rhs.view(double3), out)
+    kernel((N,), (N,), args)
```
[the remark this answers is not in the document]
keckj: Oh my god you are right, I just came back to using CUDA after four years of OpenCL where the first argument was the grid size and not the number of blocks !

keckj:
Thanks for this in depth feedback ! I added a test for the examples and applied all of your suggestions. Hopefully I will give you less work for the next pull request :) 

I think this PR opens the way for further improvements:
1. Simplify custom user type creation and type registration (as done in PyOpenCL for example).
2. Official support for cupy arrays with user defined types (at least for RawKernels).

Currently, in my examples, I think there is no guarantee that the arrays buffer allocated by the default cupy memory pools meet device side pointer alignment requirements. It seems that memory allocation routines from the CUDA driver or runtime API is always aligned to at least 256 bytes. This will need to be checked for further developments.

leofang (approved):
LGTM! Thanks! Let's give the CI a shot while waiting for @emcastillo to take another look at the changes.

leofang:
Jenkins, test this please

examples/custom_struct/packed_matrix.py
```diff
@@ -0,0 +1,91 @@
...
+                             Matrix<T>* out) {
+  int i = threadIdx.x;
+  out[i] = A[i] * B[i] + C;
+}
+'''
+
+
+def main():
+    N = 8
+    module = cupy.RawModule(code=code, options=('-std=c++03',),
```
leofang, line 57: Apparently we hit the same issue in two different PRs...See https://github.com/cupy/cupy/pull/4922#issuecomment-803718555.
```suggestion
    module = cupy.RawModule(code=code, options=('-std=c++11',),
```

leofang (changes requested):
CI fails due to using C++03.

examples/custom_struct/packed_matrix.py
keckj: I was not aware that C++03 was CUDA 11 specific, hope CUDA 12 will bring support for C++98 !

leofang:
Jenkins, test this please

leofang:
Windows build failed:
https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/70247/cupy_test_log.txt

emcastillo:
```
____________________ TestCustomStruct.test_complex_struct _____________________

self = <example_tests.test_custom_struct.TestCustomStruct testMethod=test_complex_struct>

    def test_complex_struct(self):
        output = example_test.run_example('custom_struct/complex_struct.py')
>       assert re.match(
            r"Overall structure itemsize: \d+ bytes\n"
            r"Structure members itemsize: \[(\s*\d+){5}]\n"
            r"Structure members offsets: \[(\s*\d+){5}]\n"
            r"Complex structure value:\n"
            r"\s+\[.*\]\n"
            r"Kernel output matches expected value.",
            output.decode('utf-8'),
        )
E       AssertionError: assert None
E        +  where None = <function match at 0x000001E346772378>('Overall structure itemsize: \\d+ bytes\\nStructure members itemsize: \\[(\\s*\\d+){5}]\\nStructure members offsets: \\[(\\s*\\d+){5}]\\nComplex structure value:\\n\\s+\\[.*\\]\\nKernel output matches expected value.', 'Overall structure itemsize: 80 bytes\r\nStructure members itemsize: [16  1 16  2 24]\r\nStructure members offsets: [ ... structure value:\r\n  [((0, 1, 2, 3), (4,), (5., 6.), (7,), (8, 9, 10))]\r\nKernel output matches expected value.\r\n')
E        +    where <function match at 0x000001E346772378> = re.match
E        +    and   'Overall structure itemsize: 80 bytes\r\nStructure members itemsize: [16  1 16  2 24]\r\nStructure members offsets: [ ... structure value:\r\n  [((0, 1, 2, 3), (4,), (5., 6.), (7,), (8, 9, 10))]\r\nKernel output matches expected value.\r\n' = <built-in method decode of bytes object at 0x000001E420499D80>('utf-8')
E        +      where <built-in method decode of bytes object at 0x000001E420499D80> = b'Overall structure itemsize: 80 bytes\r\nStructure members itemsize: [16  1 16  2 24]\r\nStructure members offsets: [... structure value:\r\n  [((0, 1, 2, 3), (4,), (5., 6.), (7,), (8, 9, 10))]\r\nKernel output matches expected value.\r\n'.decode

tests\example_tests\test_custom_struct.py:26: AssertionError
_____________________ TestCustomStruct.test_packed_matrix _____________________

self = <example_tests.test_custom_struct.TestCustomStruct testMethod=test_packed_matrix>

    def test_packed_matrix(self):
        output = example_test.run_example('custom_struct/packed_matrix.py')
>       assert re.match(
            r"Kernel output matches expected value for type 'float'.\n"
            r"Kernel output matches expected value for type 'double'.",
            output.decode('utf-8'),
        )
E       AssertionError: assert None
E        +  where None = <function match at 0x000001E346772378>("Kernel output matches expected value for type 'float'.\\nKernel output matches expected value for type 'double'.", "Kernel output matches expected value for type 'float'.\r\nKernel output matches expected value for type 'double'.\r\n")
E        +    where <function match at 0x000001E346772378> = re.match
E        +    and   "Kernel output matches expected value for type 'float'.\r\nKernel output matches expected value for type 'double'.\r\n" = <built-in method decode of bytes object at 0x000001E442A4CB60>('utf-8')
E        +      where <built-in method decode of bytes object at 0x000001E442A4CB60> = b"Kernel output matches expected value for type 'float'.\r\nKernel output matches expected value for type 'double'.\r\n".decode

tests\example_tests\test_custom_struct.py:18: AssertionError
```

leofang:
Forgot to reply:

> Currently, in my examples, I think there is no guarantee that the arrays buffer allocated by the default cupy memory pools meet device side pointer alignment requirements. It seems that memory allocation routines from the CUDA driver or runtime API is always aligned to at least 256 bytes. This will need to be checked for further developments.

I am not sure what's been referred to here. If you're asking about CuPy's device mempools, it always returns memory in chunks of size 512 bytes:
https://github.com/cupy/cupy/blob/6badbe0e61d4b604cf3eac39e423f9d7ed22248d/cupy/cuda/memory.pyx#L809-L811
But I thought it's irrelevant for the purpose of constructing custom structs for passing by value? Maybe I missed something?

keckj:
> But I thought it's irrelevant for the purpose of constructing custom structs for passing by value? Maybe I missed something?

For passing struct by value I agree this is irrelevant. 

However, In some examples I cast `__global__ char*` pointers to `__global__ custom_struct*` pointers. Those `custom_struct` pointers should be aligned at most on the biggest contained scalar type (I imagine this is 8B in Cuda). I just expect undefined behavior on user structs that are artificially aligned on more than 256B (or 512B when using CuPy memory pools). I think such extreme alignment requirements can only be achieved with `alignas` and `__align__` directives.

The following example works on my machine [test_alignment.py](https://gist.github.com/keckj/1af7a5cc776187958428f4edbaa4f87e), but the compiler may optimize the copy assuming 4096B alignment for T2 and T3.

leofang:
Jenkins, test this please

leofang:
I still don't think the 512B chunk size matters here. The compiler can freely make its own padding and alignment decisions, but in any case we will be retrieving the correct offsets, wouldn't we? As long as the memory allocated from either raw malloc or CuPy's mempool is enough to accommodate a custom struct (so that 512*N >= `sizeof(custom_struct)`) we should be ok.

emcastillo:
Jenkins, test this please

emcastillo (dismissed):

tests/cupy_tests/core_tests/test_function.py
```diff
@@ -1,14 +1,19 @@
 import unittest
+import pytest
 
 import numpy
 
 import cupy
+from cupy.core import core
```
leofang, line 7:
```suggestion
from cupy._core import core
```

leofang (changes requested):
@keckj Please rebase onto or merge with the latest master. The `cupy.core` module is renamed to `cupy._core` (as it's meant to be private).

leofang:
Jenkins, test this please

emcastillo (approved):

PR #4830, opened 2021-03-07
Fixed choice function for 0 samples from 0 candidates

### LINKED ISSUE #4511: `random.choice`: 0 samples from 0 candidates
Inconsistency in
```
>>> numpy.random.choice(0, size=(3, 0, 7))
array([], shape=(3, 0, 7), dtype=int64)
>>> numpy.random.choice([], size=(3, 0, 7))
array([], shape=(3, 0, 7), dtype=float64)
```

vs

```
>>> cupy.random.choice(0, size=(3, 0, 7))
ValueError: a must be greater than 0
>>> cupy.random.choice([], size=(3, 0, 7))
ValueError: a must be non-empty
```

Note that
```
>>> numpy.random.choice(0)
ValueError: a must be greater than 0 unless no samples are taken
```

povinsahu1909:
Fix #4511

Current version
```python
>>> cupy.random.choice(0, size=(3, 0, 7))
ValueError: a must be greater than 0
>>> cupy.random.choice([], size=(3, 0, 7))
ValueError: a must be non-empty
```
After changes
```python
>>> cupy.random.choice(0, size=(3, 0, 7))
array([], shape=(3, 0, 7), dtype=int64)
>>> cupy.random.choice([], size=(3, 0, 7))
array([], shape=(3, 0, 7), dtype=float64)

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -1021,16 +1021,13 @@ class RandomState(object):
             raise NotImplementedError
         if isinstance(a, int):
             a_size = a
-            if a_size <= 0:
-                raise ValueError('a must be greater than 0')
+            if a_size < 0:
+                raise ValueError('a must be greater than or equal to 0')
         else:
             a = cupy.array(a, copy=False)
             if a.ndim != 1:
                 raise ValueError('a must be 1-dimensional or an integer')
-            else:
-                a_size = len(a)
-                if a_size == 0:
-                    raise ValueError('a must be non-empty')
+            a_size = len(a)
 
         if p is not None:
             p = cupy.array(p)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cupy/random/_generator.py [hunk 1]
```diff
@@ -1072,6 +1072,8 @@ def choice(self, a, size=None, replace=True, p=None):
             if not isinstance(shape, int):
                 index = cupy.reshape(index, shape)
         else:
+            if a_size == 0:
+                a_size = 1
```
toslunar, line 1076: This workaround does not seem needed if `randint` method is fixed, too.  Let's fix it in another PR.

Could you comment here that ``# TODO: (#4511) Fix `randint` instead``?

toslunar (changes requested):
Thanks for contributing to CuPy.  The logic of the fix LGTM.

Could you add testcases for the issue (to https://github.com/cupy/cupy/blob/4e63310017851789a9a8787b8b11ffb62bcebe03/tests/cupy_tests/random_tests/test_generator.py#L1003-L1008)?  The two cases for `a` seems sufficient: `0` and `array([])`.

`test_bound` in the class can be skipped (``self.skipTest('no bound check for empty `random.choice`')``) if `size_ == (0,)`.

toslunar (approved):
LGTM

PR #4842, opened 2021-03-09
Avoid evaluating PTDS environment variable everytime

kmaehashi:
`get_current_stream_ptr` is one of the frequently called function (every kernel launch / malloc), so let's avoid the overhead.


emcastillo:
Jenkins, test this please

emcastillo (dismissed):

emcastillo (approved):

PR #4843, opened 2021-03-09
Fix broken Stream docs

kmaehashi:
Fix parameters list in https://docs.cupy.dev/en/latest/reference/generated/cupy.cuda.Stream.html


leofang (approved):
Good catch!

emcastillo (approved):

PR #4848, opened 2021-03-09
Add NCCL to library installer

kmaehashi:
Depends on ~https://github.com/cupy/cupy/pull/4912~ (merged) and #4852

PR #4852, opened 2021-03-09
Improve cuTENSOR installer

kmaehashi:
* Use `lib` instead of `lib64` (keep dir name in original archive)
  ~Depends on https://github.com/cupy/cupy/pull/4912~ merged
* Preserve license file
* A bit cleanup

### DIFF
--- a/cupyx/tools/install_library.py
+++ b/cupyx/tools/install_library.py
@@ -165,12 +165,17 @@ Remove the directory first if you want to reinstall.'''.format(destination))
         if library == 'cudnn':
             shutil.move(os.path.join(tmpdir, 'cuda'), destination)
         elif library == 'cutensor':
-            include = os.path.join(destination, 'include')
-            lib = os.path.join(destination, 'lib64')
-            shutil.move(os.path.join(tmpdir, 'libcutensor/include'), include)
-            if cuda.startswith('11'):
+            if cuda.startswith('11.'):
                 cuda = '11'
-            shutil.move(os.path.join(tmpdir, 'libcutensor/lib', cuda), lib)
+            shutil.move(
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, test this please.

kmaehashi:
#4912 merged, ready for review.
/test

emcastillo (approved):

cupyx/tools/install_library.py
```diff
@@ -165,12 +165,17 @@ def install_lib(cuda, prefix, library):
...
         elif library == 'cutensor':
-            include = os.path.join(destination, 'include')
-            lib = os.path.join(destination, 'lib64')
-            shutil.move(os.path.join(tmpdir, 'libcutensor/include'), include)
-            if cuda.startswith('11'):
+            if cuda.startswith('11.'):
                 cuda = '11'
-            shutil.move(os.path.join(tmpdir, 'libcutensor/lib', cuda), lib)
+            shutil.move(
+                os.path.join(tmpdir, 'libcutensor', 'include'),
```
emcastillo, line 171: Supper stupid and irrelevant.
Is there any strong reason not to use `Path` https://docs.python.org/3/library/pathlib.html
now that we only support python 3? I feel it is more readable than `os.path.join`.
But we don't need to change it now!

asi1024: Don't need to change it now, but I prefer `pathlib` too.

kmaehashi: Thanks for the suggestion! Agree with that, I'll rewrite it pathlib after merging this one and #4848.

asi1024 (approved):

Issue #4821, opened 2021-03-05
CuPy memory leak? (Outstanding consumption.)


willemavjc:
Hello.

I am wondering how is managed the memory through CuPy and if there could be an issue in the way it is handled.
Let me explain:

I was curious about CuPy and decided to compare it against NumPy via a very simple code that can be find below.
While it works like a charm on NumPy - so on CPU and standard RAM - it totally crashes with CuPy: `cupy.cuda.memory.OutOfMemoryError` is thrown.

```
import numpy as np
import  cupy as cp
import  time

# Reminder: The ** operator means a square operation i.e. 10^9 = 1,000,000,000.
P = 9
N = 10**P

print(f"Scenario 1: Summing ranges of {N} random integers between 0 and 9.")
print("")

Lx = np.random.randint(0, 10, size=N)
print(Lx.nbytes/1024/1024)

T0 = time.perf_counter()
np.sum(Lx)
T1 = time.perf_counter()

del Lx

print("CPU computation done in {:5,.0f} ms.".format((T1-T0)*1000))

Lx = cp.random.randint(0, 10, size=N)
print(Lx.nbytes/1024/1024)

T0 = time.perf_counter()
cp.sum(Lx)
T1 = time.perf_counter()

del Lx

print("GPU computation done in {:5,.0f} ms.".format((T1-T0)*1000))
```

This small code gives the following output:

```
Scenario 1: Summing ranges of 1000000000 random integers between 0 and 1.

3814.697265625
CPU computation done in   585 ms.
Traceback (most recent call last):
  File "C:\APPLIS\nvidia\samples\operations.py", line 23, in <module>
    Lx = cp.random.randint(0, 10, size=N)
  File "C:\APPLIS\python\lib\site-packages\cupy\random\_sample.py", line 110, in randint
    return rs.randint(low, high, size, dtype)
  File "C:\APPLIS\python\lib\site-packages\cupy\random\_generator.py", line 1161, in randint
    x = self._interval(diff, size).astype(dtype, copy=False)
  File "C:\APPLIS\python\lib\site-packages\cupy\random\_generator.py", line 709, in _interval
    ret[n - n_rem:n - n_rem + n_succ] = sample[success][:n_succ]
  File "cupy\core\core.pyx", line 1228, in cupy.core.core.ndarray.__getitem__
  File "cupy\core\_routines_indexing.pyx", line 39, in cupy.core._routines_indexing._ndarray_getitem
[...]

Issue #4080, opened 2020-10-06
Documentation lacking on managed memory


niboshi:
Hello.
I found [the documentation of memory management](https://docs.cupy.dev/en/v8.0.0/reference/memory.html) lacks a mention about managed memory (unified memory).
I couldn't figure out how to allocate an `ndarray` with managed memory without reading the code.
Hopefully it can be improved.

Btw I feel this documentation focuses too much on memory pools.
Memory pools are just instances of allocators after all (in my understanding). 
Hopefully that will be fixed as well.

kmaehashi:
Sharing a convenient snippet for future readers:

```
# Enable unified memory for all CuPy allocations
cupy.cuda.set_allocator(cupy.cuda.MemoryPool(cupy.cuda.malloc_managed).malloc)
```

Better to add this to the docs. I also think the class diagram is needed to help users understand the relationship and inheritance between these classes:
https://docs.cupy.dev/en/stable/reference/cuda.html#memory-management

leofang:
Didn't we already document how to use the mempool for managed/pinned memory?
https://docs.cupy.dev/en/stable/user_guide/memory.html#changing-memory-pool
Buy I agree having a visual aid is useful for users.

cc: @jakirkham for vis

kmaehashi:
> Didn't we already document how to use the mempool for managed/pinned memory?
> https://docs.cupy.dev/en/stable/user_guide/memory.html#changing-memory-pool

Ah right, I was searching the doc with "unified" and only seeing API references 😅 
https://docs.cupy.dev/en/stable/search.html?q=unified

leofang:
We should mention "a.k.a. unified memory" 😂

PR #4857, opened 2021-03-10
Revert memory back to reference

emcastillo:
We discussed that the memory explanation should be in reference instead of tutorial.


leofang:
I moved it to tutorial in #4537 because it was harder to find when it was in reference (ex: https://github.com/cupy/cupy/issues/4080, #4821). Also, the presentation there is more like a tutorial than a pure API reference to me. Perhaps we could at least add a link to this page somewhere, say in the landing page ("Overview"), so that it's easier for new comers to know it exists?

kmaehashi:
I asked @emcastillo to move it back to the reference 😃 

> Also, the presentation style there is more like a tutorial than a pure API reference to me. Perhaps we could at least add a link to this page somewhere, say in the landing page ("Overview"), so that it's easier for new comers to know it exists?

I talked with @emcastillo and agreed on this point. Let's add a simple description and a link to memory management page in the "Basics of CuPy" page.

Maybe in future, it's better to add a new top-level section like "Using CuPy", which includes topics like memory pools, interoperability, how to profile, etc.

emcastillo:
Pushed a link from the tutorial to the reference.

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

leofang:
Thanks guys!

Issue #4476, opened 2020-12-21
cp.roll cannot take array as shift argument


thomasaarholt:
I was taking the output of cp.unravel_index, which returns an array, and feeding it to cp.roll, which shifts an array. This raises the error below, but works when the shift argument is a tuple. It applies to 1D shifts as well.

It looks like perhaps the `numpy.broadcast(shift, axis)` in the error below should be `cupy.broadcast(shift, axis)` instead?

```python
>>> arr = cp.array([1,2,3,4]).reshape((2,2))
>>> shift = cp.array([1,1]) # shift (1,1) works
>>> cp.roll(arr, shift)

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-158-005527abcbc1> in <module>
      1 arr = cp.array([1,2,3,4])
      2 shift = cp.array([1])
----> 3 cp.roll(arr, shift)

~\Miniconda3\envs\py\lib\site-packages\cupy\_manipulation\rearrange.py in roll(a, shift, axis)
     99     """
    100     if axis is None:
--> 101         return roll(a.ravel(), shift, 0).reshape(a.shape)
    102     else:
    103         axis = _reduction._get_axis(axis, a.ndim)[0]

~\Miniconda3\envs\py\lib\site-packages\cupy\_manipulation\rearrange.py in roll(a, shift, axis)
    103         axis = _reduction._get_axis(axis, a.ndim)[0]
    104 
--> 105         broadcasted = numpy.broadcast(shift, axis)
    106         if broadcasted.nd > 1:
    107             raise ValueError(

cupy\core\core.pyx in cupy.core.core.ndarray.__array__()

TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.
```

CuPy Version                 : 8.2.0
NumPy Version                : 1.19.4
SciPy Version                : 1.5.3
Cython Build Version         : 0.29.21
CUDA Root                    : C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.1
CUDA Build Version           : 11010
CUDA Driver Version          : 11020
CUDA Runtime Version         : 11010
cuBLAS Version               : 11201
cuFFT Version                : 10300
cuRAND Version               : 10202
cuSOLVER Version             : (11, 0, 0)
cuSPARSE Version             : 11200
NVRTC Version                : (11, 1)
Thrust Version               : 100910
CUB Build Version            : 100910
[...]

PR #4864, opened 2021-03-10
Fix broadcasting error in cupy roll if shift argument is an array

awthomp:
Closes https://github.com/cupy/cupy/issues/4476

CuPy's roll function relies on Numpy's broadcasting function. This is all fine and good if the shift parameter is either a scalar or tuple, but if the shift parameter is an array, the user receives a cupy/numpy conversion error. Since CuPy's broadcast function doesn't yield an iterable object, we can't simply replace `numpy.broadcast` with `cupy.broadcast`.

I've submitted a PR to check if the shift argument is a `cupy.ndarray` object. If so, we migrate to the CPU and let `numpy.broadcast` do its thing with a numpy shift array.

The following code now works as expected:

```python
import cupy as cp
a = cp.arange(12).reshape(4,3)
shift = cp.array([1,2,3,4])
rolled = cp.roll(a, shift)
```

### DIFF
--- a/cupy/_manipulation/rearrange.py
+++ b/cupy/_manipulation/rearrange.py
@@ -1,4 +1,5 @@
 import itertools
+import warnings
 
 import numpy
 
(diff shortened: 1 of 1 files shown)


thomasaarholt:
Can I suggest adding a warning? That'll A) make it clear to users that it is *preferable* to pass a `numpy` array rather than a `cupy` one, and B) it'll serve as a reminder that this could be improved in the future.

asi1024:
@awthomp Thank you for your PR!
We are considering another implementation that does not require device-to-host data transfer now (#4884). Do you have any concerns?

awthomp:
Hey @asi1024. No concerns here! This PR was always meant to be a stop-gap, so I totally approve of a more efficient implementation.

asi1024:
#4884 is merged!

PR #4870, opened 2021-03-11
Add APIs for creating NumPy arrays backed by pinned memory

### LINKED ISSUE #3625: Pinned memory allocation returns odd size
It looks like CuPy allocates more bytes than expected when calling `cupy.cuda.alloc_pinned_memory`. Any ideas why that might be?

```python
In [1]: import numpy 
   ...: import cupy                                                             

In [2]: a = cupy.arange(1_000_000)                                              

In [3]: b = numpy.asarray(cupy.cuda.alloc_pinned_memory(a.nbytes)).view(a.dtype)
   ...:                                                                         

In [4]: b.nbytes                                                                
Out[4]: 8388608

In [5]: a.nbytes                                                                
Out[5]: 8000000
```

  - CuPy version: 7.6.0
  - OS/Platform: Ubuntu 18.04.4 LTS
  - CUDA version: 10.2
  - cuDNN/NCCL version (if applicable): 7.6.5/2.5.7.1

```$ python -c 'import cupy; cupy.show_config()'
CuPy Version          : 7.6.0
CUDA Root             : /datasets/jkirkham/miniconda/envs/rapids15dev
CUDA Build Version    : 10020
CUDA Driver Version   : 10020
CUDA Runtime Version  : 10020
cuBLAS Version        : 10202
cuFFT Version         : 10102
cuRAND Version        : 10102
cuSPARSE Version      : 10301
cuDNN Build Version   : 7605
cuDNN Version         : 7605
NCCL Build Version    : 2406
NCCL Runtime Version  : 2507
CUB Version           : None
cuTENSOR Version      : None
```

leofang:
Close #3625. 

This PR adds a few convenience functions to the `cupyx` namespace for allocating NumPy arrays backed by pinned/pagelocked memory, see the discussion around https://github.com/cupy/cupy/issues/3625#issuecomment-704384736. Currently the provided APIs are as follows:
- `cupyx.empty_pinned()`
- `cupyx.empty_like_pinned()`
- `cupyx.zeros_pinned()`
- `cupyx.zeros_like_pinned()`

with exactly the semantics as their NumPy counterparts.

Ideally, we can do better than this if we use NumPy's memory hooks to always allocate NumPy arrays from pinned memory (https://github.com/numpy/numpy/issues/17467), however it's still a WIP, so we can only revisit it when it's in.

TODO:
- [x] Decide if the name/API is appropriate or not
- [x] Add docstrings 
- [x] Fix order = 'K'
- [x] Add tests


leofang:
Help needed: I am writing tests and I noticed `@testing.numpy_cupy_array_equal()` does not work for this PR, because we are comparing two `numpy.ndarray`s which breaks the assumption of this test utility. Any advice?

leofang:
> Help needed: I am writing tests and I noticed `@testing.numpy_cupy_array_equal()` does not work for this PR, because we are comparing two `numpy.ndarray`s which breaks the assumption of this test utility. Any advice?

So I wrote a decorator `@numpy_cupyx_array_equal()` to address this need. The tests added in the new `tests/cupyx_tests/test_pinned_array.py` largely follows `tests/cupy_tests/creation_tests/test_basic.py` for the four added functions, with this decorator change and some other necessary but minor changes.

leofang:
Flake8 error will be fixed in #4888.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

cupyx/_pinned_array.py
```diff
@@ -0,0 +1,157 @@
...
+        shape = a.shape
+    elif isinstance(shape, int):
+        shape = (shape,)
+    else:
+        shape = tuple(shape)
+    return shape
+
+
+def _update_order_char(x, order_char):
+    # This is a pure Python version of cupy.core.core._update_order_char()
```
leofang, line 19: We could relax the typing in `cupy.core.core._update_order_char()` so that it can be used here. However, doing so increases the overhead a little bit, as we have to access Python attributes (`x.flags.f_contiguous`) instead of C attributes (`x._f_contiguous`) which is absent in `numpy.ndarray`:
|  | time |
|--- | --- |
| original `cupy.core.core._update_order_char()` | 0.13 us |
| remove the `ndarray` typing from `x` (but still kept in Cython) | 0.48 us |
| full Python implementation (as done above) | 0.90 us |

Let me know if this overhead is acceptable for the purpose of reducing code duplication.

kmaehashi: How about changing the original `_update_order_char` function to take `bool c_contiguous, bool f_continguous`  instead of `ndarray`?

cupyx/_pinned_array.py
```diff
@@ -0,0 +1,157 @@
...
+    elif order_char == ord('K'):
+        if x.flags.f_contiguous:
+            order_char = ord('F')
+        elif x.flags.c_contiguous:
+            order_char = ord('C')
+    return order_char
+
+
+def empty_pinned(shape, dtype=float, order='C'):
+    """Returns an array without initializing the elements.
```
kmaehashi, line 35:
```suggestion
    """Returns a NumPy ndarray without initializing the elements.
```

cupyx/_pinned_array.py
```diff
@@ -0,0 +1,157 @@
...
+
+    This is a convenience function which is just :func:`numpy.empty`,
+    except that the underlying memory is allocated from CuPy's pinned
+    memory pool.
```
kmaehashi, line 39: Users may disable the memory pool, so:

```suggestion
    except that the underlying memory is allocated as a pinned memory.
```

cupyx/_pinned_array.py
```diff
@@ -0,0 +1,157 @@
...
+    order, strides, _ = _new_like_order_and_strides(
+        a, dtype, order, shape, get_memptr=False, get_char=_update_order_char)
+    nbytes = internal.prod(shape) * numpy.dtype(dtype).itemsize
+    mem = cuda.alloc_pinned_memory(nbytes)
+    out = numpy.ndarray(shape, dtype=dtype, buffer=mem,
+                        strides=strides, order=order)
+    return out
+
+
+def zeros_pinned(shape, dtype=float, order='C'):
```
kmaehashi, line 104: I discussed with the dev members and wondering `zeros_*` APIs are really needed or not, because users can simply zero-fill the arrays returned by `empty_*`.

kmaehashi (changes requested):

cupyx/_pinned_array.py
leofang: I don't have strong opinion here. Let me know if you'd like me to remove zeros*.

I simply followed the NumPy API, which has zeros*, though I do agree when I first learned this I also thought users could just zero it out by themselves.

By the way, I just found PyCUDA also provides both empty* and zeros*:
https://documen.tician.de/pycuda/driver.html?highlight=page%20locked#pagelocked-allocation

cupyx/_pinned_array.py
leofang: Thanks, done in 1bc4e3c with minor changes.

cupyx/_pinned_array.py
leofang: Done in 1bc4e3c with minor changes.

cupyx/_pinned_array.py
leofang: Good point, @kmaehashi! I should have thought about it...Done in b833b24. No performance regression is observed, with a caveat that in Python the (likely small) overhead is swept into function parameter preparation (`c_contiguous` and `f_contiguous`).

leofang:
Note to self:
1. Today I took some time looking into the possibility of `numpy_array[:] = cupy_array`, the opposite direction of  the slice copy enabled by `CUPY_EXPERIMENTAL_SLICE_COPY` (#2079). The motivation is if `numpy_array` can be backed by pinned memory then this could be a cheaper operation than before. However, this doesn't seem possible without implementing `__array__`, which we forbid.
2. However, it is possible to support `a_np_on_pinned = cp.asnumpy(a_cp, out=a_np_on_pinned)` and a few other copy methods that internally allocate pinned memory. I will try to make a PR for these once this one is in.

leofang:
@kmaehashi This is ready (but see https://github.com/cupy/cupy/pull/4870#discussion_r595790284).

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

jakirkham:
Jenkins, test this please

leofang:
@kmaehashi Errors are due to OSError: [Errno 28] No space left on device.

kmaehashi:
I've freed up some space now.

pfnCI, test this please.

leofang:
hmmm...I wonder why Jenkins still shows a red cross despite chainer ci says success?

kmaehashi:
It seems Jenkins github plugin was using deprecated github APIs and now they're removed. I bumped the plugin version so it should work now.

jakirkham:
Jenkins, test this please

leofang:
Resolved conflicts.

Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
@kmaehashi as mentioned I'd love to have this PR backported 🙏

leofang:
Jenkins, test this please

kmaehashi (approved):
LGTM!

cupyx/_pinned_array.py
kmaehashi: Thanks for the PyCUDA link, sounds reasonable to provide all of them.

leofang:
Thanks, @kmaehashi and @jakirkham!

Issue #4850, opened 2021-03-09
Do not bundle cuDNN / NCCL for all wheel packages starting in CuPy v9


kmaehashi:
We have been distributing binary packages (`cupy-cudaXXX`) via PyPI. However recently the size of GPU-related packages (including CuPy) is starting to cause a problem on PyPI (see [discussions on Python forum](https://discuss.python.org/t/what-to-do-about-gpus-and-the-built-distributions-that-support-them/7125)).

We take this problem seriously, and to help PyPI ecosystem healthy, we are planning to stop bundling cuDNN / NCCL shared libraries from all wheels, starting in v9 releases.
Note that cuDNN / NCCL are still supported and enabled in wheels, but users who want to use these features need to install the library via `python -m cupyx.tools.install_library` command.

----

FYI, here are our past efforts to reduce the package size on PyPI:

* CuPy v7: Do not bundle cuDNN v8.x (#3724)
* CuPy v8: Remove outdated pre-release wheels (#4360)

pentschev:
As we discussed earlier today on our call, could we extend this effort to conda packages, effectively making cuDNN an optional install rather than a requirement?

leofang:
> As we discussed earlier today on our call, could we extend this effort to conda packages, effectively making cuDNN an optional install rather than a requirement?

@pentschev This is totally doable (cc: @jakirkham). Do you mind opening up an issue in https://github.com/conda-forge/cupy-feedstock so we can track this need there?

leofang:
> This is totally doable

Sorry, I spoke too fast. There is a tiny bit of work that needs to be done in CuPy, namely how to detect the absence of optional dependencies (cudnn/nccl/cutensor) and ask users to install them through Conda. We already have a mechanism to recommend doing `python -m cupyx.tools.install_library` as needed, so the best case scenario is we modify it to accommodate Conda.

jakirkham:
So what happens when `nccl` is not present? It raises an error and asks the user to install?

kmaehashi:
Currently, wheels use library metadata (in JSON) which is generated and bundled during the wheel build process.
https://github.com/cupy/cupy/blob/v9.0.0b3/cupy/_environment.py#L22-L48

Warnings are displayed using the metadata as follows:
https://github.com/cupy/cupy/blob/v9.0.0b3/cupy/cuda/cudnn.py#L9-L14
https://github.com/cupy/cupy/blob/v9.0.0b3/cupy/_environment.py#L330-L341

[...]

PR #4873, opened 2021-03-11
Support optional dependencies from Conda-Forge

leofang:
Blocked by ~~#4848, #4919, #4920~~.

This is a fix related to #4850 to get Conda-Forge supported.

I expect to generate `cupy/.data/_wheel.json` in the Conda recipe, with a top-level entry `packaging` set to `conda`, so it's not touched in this PR.

TODO:
- [x] Update installation guide.


leofang:
cc: @jakirkham

emcastillo:
cc @pentschev

leofang:
Flake8 error will be fixed in #4888.

cupy/cuda/__init__.py
```diff
@@ -53,13 +54,15 @@ def __init__(self, name):
 
 try:
     from cupy.cuda import nccl  # NOQA
-except ImportError:
+except ImportError as e:
```
kmaehashi, line 57: To support preloading we need to separate this code into another file like done in `cudnn.py`, and remove import from `__init__.py`. Otherwise, warnings will be shown every time running `import cupy` when nccl/cutensor is unavailable.

Strictly speaking this is backward incompatible as this breaks user code (e.g., `import cupy.cuda; print(cupy.cuda.nccl)`), but I think it's acceptable.

cupyx/tools/install_library.py
```diff
@@ -35,6 +35,7 @@ def _make_cudnn_record(
         cuda_version, public_version, filename_linux, filename_windows):
     major_version = public_version.split('.')[0]
     return {
+        'packaging': 'pip',
```
kmaehashi, line 38: `_wheel.json` is not directly generated from this metadata, so no change is needed in installers.

cupyx/tools/install_library.py
```diff
@@ -92,6 +93,7 @@ def _make_cutensor_url(public_version, filename):
 def _make_cutensor_record(
         cuda_version, public_version, filename_linux, filename_windows):
     return {
+        'packaging': 'pip',
```
kmaehashi, line 96: ditto

cupy/cuda/__init__.py
```diff
@@ -53,13 +54,15 @@ def __init__(self, name):
...
     nccl = _UnavailableModule('cupy.cuda.nccl')
+    _preload_warning('nccl', e)
 
 try:
     from cupy_backends.cuda.libs import cutensor
-except ImportError:
+except ImportError as e:
```
kmaehashi, line 63: ditto

cupy/_environment.py
```diff
@@ -330,12 +340,23 @@ def _get_preload_logs():
...
+'''
+        if config['packaging'] == 'pip':
+            msg += '''
   $ python -m cupyx.tools.install_library --library {lib} --cuda {cuda}
-'''.format(lib=lib, exc_type=type(exc).__name__, exc=str(exc),
-           cuda=config['cuda']))
+'''
+        elif config['packaging'] == 'conda':
+            msg += '''
+  $ conda install -c conda-forge {lib}
```
kmaehashi, line 356: Question: Is it command enough to pick a correct library version? (`cudatoolkit=x.y` should be specified in the same command line as in the install guide?)

kmaehashi (changes requested):

cupy/cuda/__init__.py
leofang: You are right, I was thinking to always warn if libs are not found at CuPy import time, but perhaps it's too much noise.

@kmaehashi in this case, let me send a separate PR moving nccl to cupy_backends (which is something long overdue anyway), and then add cupy/cuda/nccl.py to enable this.

cupyx/tools/install_library.py
leofang: @kmaehashi But I thought you generate the json file by running this script?

cupy/cuda/__init__.py
kmaehashi: Separate PR sounds good, thank you!

cupy/_environment.py
leofang: So the scenario I expect is like this: Users have successfully installed CuPy with the correct pick of cudatoolkit, but do not install the optional libs. Then at runtime we recommend users to do so if needed. In this case, when doing conda install, conda will pick the right libs that work with the (already installed) cudatoolkit version, so we don't need to add cudatoolkit to the install command explicitly. (Doing so has a small chance to reinstall it if for some reason we bump the build number in the cudatoolkit feedstock, which is no harm but a bit wasting bandwidth and time.)

cupyx/tools/install_library.py
kmaehashi: Currently, the release tool runs the installer with `--action dump` option to extract info from this installer, then generate JSON using that info:
https://github.com/cupy/cupy-release-tools/blob/c9ad0ddb8cc4a5f20d062073c27e77d5748e4deb/dist.py#L404-L413

So installer does not need to have `packaging` field.

cupy/_environment.py
kmaehashi: Thanks for clarification! So conda will pick a correct package for  `cupy` + `cudatoolkit` currently installed in the environment.

cupyx/tools/install_library.py
leofang: Ahh ok, didn't realize it's a two-step process. I thought `--action dump` is enough...OK, then I'll remove the changes in the installer, and let you handle `'packaging': 'pip'` in the release tool (which I kinda lost track 😅).

cupy/_environment.py
leofang: That's right!

cupyx/tools/install_library.py
kmaehashi: Yes, I'm working on it now.  > 'packaging': 'pip' in the release tool

cupy/cuda/__init__.py
leofang: See #4919. @kmaehashi I suppose we also need a `cupy/cuda/cutensor.py`?

kmaehashi: Yes!

leofang: > I suppose we also need a `cupy/cuda/cutensor.py`?

See #4920.

leofang: Both #4919 and #4920 are in!

cupyx/tools/install_library.py
leofang: Changes in `install_library.py` are reverted.

kmaehashi:
LGTM :+1:

cupy/_environment.py
```diff
@@ -338,12 +348,23 @@ def _get_preload_logs():
...
+'''
+        else:
+            assert False
+        msg.format(lib=lib, exc_type=type(exc).__name__, exc=str(exc),
```
kmaehashi, line 368:
```suggestion
        msg  = msg.format(lib=lib, exc_type=type(exc).__name__, exc=str(exc),
```

kmaehashi:
pfnCI, test this please.

kmaehashi:
Ready for review?

leofang:
Yes, thank you @kmaehashi 🙂

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

leofang:
@kmaehashi The AMD CI has passed long ago, but the status is not updated here, so automerge does not kick in.

jakirkham:
Thanks Leo and everyone for the reviews! 😄

PR #4884, opened 2021-03-14
Support `cupy.ndarray` type `shift` in `cupy.roll`

asi1024:
Another version of #4864. (cc: @awthomp)

### DIFF
--- a/cupy/_manipulation/rearrange.py
+++ b/cupy/_manipulation/rearrange.py
@@ -99,6 +99,28 @@ def roll(a, shift, axis=None):
     """
     if axis is None:
         return roll(a.ravel(), shift, 0).reshape(a.shape)
+    elif isinstance(shift, cupy.ndarray):
+        shift = shift.ravel()
+        axes = _reduction._get_axis(axis, a.ndim)[0]
+        n_axes = max(len(axes), shift.size)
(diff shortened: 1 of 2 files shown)


cupy/_manipulation/rearrange.py
```diff
@@ -99,6 +99,28 @@ def roll(a, shift, axis=None):
...
+        n_axes = max(len(axes), shift.size)
+        axes = numpy.broadcast_to(axes, (n_axes,))
+        shift = cupy.broadcast_to(shift, (n_axes,))
+
+        # TODO(asi1024): Improve after issue #4799 is resolved.
+        indices = []
+        for ax in range(a.ndim):
+            ind_shape = [1] * a.ndim
+            ind_shape[ax] = a.shape[ax]
+            indices.append(cupy.arange(a.shape[ax]).reshape(ind_shape))
```
asi1024, line 114: This implementation requires GPU memory of `sum(a.shape)` size. It is almost ignorable for multi-dimensional input, but not for 1-dim input. ElementwiseKernel can reduce the memory usage, but increases the code size.

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #4890, opened 2021-03-15
Support tuple in CuPy JIT

asi1024:


cupy/core/include/cupy/tuple/tuple.h
```diff
@@ -0,0 +1,961 @@
+/*
```
emcastillo, line 1: Maybe using jitify instead of forking the header works?

emcastillo: However, I am not sure if jitify worked with amd

leofang: No, it doesn't work, we checked together.

leofang: Well, another route for AMD is to compile using hipcc instead of hiprtc if tuple is needed. I didn't test it though.

asi1024:
PTAL!

cupy/core/core.pyx
```diff
@@ -1865,6 +1865,7 @@ cdef list _cupy_header_list = [
     'cupy/complex.cuh',
     'cupy/carray.cuh',
     'cupy/atomics.cuh',
+    'cupy/tuple.cuh',
```
leofang, line 1868: Do we have to include this for every kernel?

tests/cupy_tests/functional_tests/test_vectorize.py
```diff
@@ -477,6 +477,29 @@ def func_for(x):
...
+            a, b = z
+            return a * a + b
+
+        f = xp.vectorize(func_tuple)
+        x = xp.array([0, 1, 2, 3, 4])
+        y = xp.array([5, 6, 7, 8, 9])
+        return f(x, y)
+
+    @testing.numpy_cupy_array_equal()
+    def test_return_tuple(self, xp):
```
leofang, line 494: I suppose this test also implies the following operation will work fine as a `vectorize` target?
```python
>>> x = 5
>>> y = 3
>>> a, b = y-2, x-3
>>> print(a, b)
1 2
```

leofang (commented):
Left a few quick questions 🙂 btw, could you add a short note to `cupy/core/include/cupy/README.md` to explain the source of the bundled headers? While it's not as big as the full `complex` support, it's still helpful for unforeseen future needs.

emcastillo:
Jenkins, test this please

tests/cupy_tests/functional_tests/test_vectorize.py
asi1024: Yes, the test of a tuple of expressions is covered in `test_return_tuple`.

emcastillo (dismissed):

cupy/core/core.pyx
asi1024: No, but it will not break compatibility.

emcastillo:
Jenkins, test this please

asi1024:
Jenkins, test this please

cupy/core/core.pyx
leofang: Initially I was only concerned about compilation time (which could be lengthened a bit), but now I am also concerned about potential conflict with Jitify:
https://github.com/NVIDIA/jitify/blob/60e9e7231fe7529ecfc76fa97b9c02b3033f4157/jitify.hpp#L2380
although it would only happen if `#include <tuple>` is done and `using std::tuple` is in use.

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo (approved):

emcastillo:
Jenkins, test this please

PR #4894, opened 2021-03-16
Fix uniform random generation interval

emcastillo:


cupy/random/cupy_distributions.cu
```diff
@@ -37,8 +37,8 @@ struct curand_pseudo_state: rk_state {
     __device__ virtual double rk_double() {
         // Curand returns (0, 1] while the functions
         // below rely on [0, 1)
-        double r = curand_uniform(_state) - 1e-12;
-        if (r < 0.0) { 
+        double r = curand_uniform(_state);
+        if (((int) r) == 1) {
```
toslunar, line 41: How about `r == 1.0` or `r >= 1.0`?

emcastillo: Is `r==1.0` safe?

emcastillo: I’ve been wondering about this and you are right.

If the result r from curand is 1.0, but r == 1.0 is false due to fp issues then, r is either > or < than 1.0 so if r < 1.0 then
we have sampled in the [0, 1) interval.
But if r >= 1.0 then we are including a value out of the domain so it is the condition we should look for.

Thanks!

toslunar (approved):

toslunar:
pfnCI, test this please.

PR #4905, opened 2021-03-18
Add gamma distributions to random API

emcastillo:


cupy/random/_generator_api.pyx
```diff
@@ -195,6 +222,40 @@ class Generator:
...
+            raise ValueError('shape < 0')
+
+        if out is not None:
+            self._check_output_array(dtype, size, out)
+
+        y = ndarray(size if size is not None else (), numpy.float64)
+        _launch_dist(self.bit_generator, standard_gamma, y, (shape,))
+        if out is not None:
+            out[...] = y
+            y = out
```
leofang, line 253: Silly question: Could you remind me why here and elsewhere in this file we had to create a new array `y` first, and then copy the outcome to `out` if provided? Can't we pass `out` to `_launch_dist` directly if it's not `None`, saving one copy?

emcastillo: this is a very good question.
Arrays are generated using fp64 due to curand limitations, so we cant write them directly to other types.
But I agree that if the type of out is the same type than the generated array we should pass out directly.
I will fix this once all the random outgoing prs are merged.

emcastillo:
Ready! PTAL

cupy/random/_generator_api.pyx
emcastillo: Fixed :P

cupy/random/_generator_api.pyx
```diff
@@ -219,6 +222,49 @@ class Generator:
...
+            if out.dtype.char == 'd':
+                y = out
+
+        if numpy.dtype(dtype).char not in ('f', 'd'):
+            raise TypeError(
+                f'Unsupported dtype {y.dtype.name} for standard_normal')
+
+        y = ndarray(size if size is not None else (), numpy.float64)
+        _launch_dist(self.bit_generator, standard_gamma, y, (shape,))
+        if out is not None and y is not out:
```
leofang, line 260: I think at this point this is safe:
```suggestion
        if y is not out:
```

cupy/random/_generator_api.pyx
```diff
@@ -219,6 +222,49 @@ class Generator:
...
+        if shape < 0.0:
+            raise ValueError('shape < 0')
+
+        if out is None or out.dtype.char != 'd':
+            y = ndarray(size if size is not None else (), numpy.float64)
+
+        if out is not None:
+            self._check_output_array(dtype, size, out)
+            if out.dtype.char == 'd':
+                y = out
```
leofang, line 252: How about this?
```suggestion
        y = None
        
        if out is not None:
            self._check_output_array(dtype, size, out)
            if out.dtype.char == 'd':
                y = out
        if y is None:
             y = ndarray(size if size is not None else (), numpy.float64)
```
Then you can remove line 258, which mistakenly allocates a new `y` regardless of `out`.

cupy/random/_generator_api.pyx
emcastillo: It was not 😂, 
`None` is not `y`, so we try to copy stuff in `None` and the tests failed. I thought so at the beginning too :)

leofang: Ah ok 😂

cupy/random/_generator_api.pyx
emcastillo: Ah, good catch! will update it tomorrow :)

emcastillo:
Rebased, PTAL

cupy/random/_generator_api.pyx
```diff
@@ -285,6 +288,49 @@ class Generator:
 
         return y
 
+    def gamma(self, shape, scale=1.0, size=None):
+        """Returns an array of samples drawn from a gamma distribution.
+
+        .. seealso::
+            - :func:`cupy.random.gamma` for full documentation
+            - :meth:`numpy.random.RandomState.gamma`
```
toslunar, line 296: - OK to link to `cupy.random.gamma` for the explanation of the distribution.
- `numpy.random.Generator.gamma` seems more suitable than `numpy.random.RandomState.gamma`.

cupy/random/_generator_api.pyx
```diff
@@ -285,6 +288,49 @@ class Generator:
...
+        """Returns an array of samples drawn from a standard gamma distribution.
+
+        .. seealso::
+            - :func:`cupy.random.standard_gamma` for full documentation
+            - :meth:`numpy.random.RandomState.standard_gamma`
+        """
+        cdef ndarray y
+
+        if shape < 0.0:
+            raise ValueError('shape < 0')
```
toslunar, line 310: Will the method support `shape: Union[float, cupy.ndarray]`?

cupy/random/_generator_api.pyx
emcastillo: Yes, I wanted to fix all docstrings in another PR but I will fix this one in here

cupy/random/_generator_api.pyx
emcastillo: This is a good question, I think it should not support it because it will be enforcing device synchronization.

emcastillo: I have double checked and it also acceps an array of floats, so I have to think about this check.

emcastillo:
Even the old api in cupy is not conformant to numpy. This needs more work

cupy/random/cupy_distributions.cu
```diff
@@ -255,14 +256,46 @@ struct standard_normal_float_functor {
...
+    using RetType = T;
+    static __device__ RetType get(T value, int id) {
+        return value;
+    }
+   
+};
+
 template<typename F, typename T, typename R, typename... Args>
 __global__ void execute_dist(intptr_t state, intptr_t out, ssize_t size, Args... args) {
     int id = threadIdx.x + blockIdx.x * blockDim.x;
```
asi1024, line 293: Unrelated to this PR, but `id` should be of type int64?

cupy/random/cupy_distributions.cu
```diff
@@ -273,19 +273,23 @@ struct pointer_fnct {};
 
 template<typename T>
 struct pointer_fnct<true, T> {
-    using RetType = typename std::remove_pointer<T>::type;
-    static __device__ RetType get(T value, int id) {
-        return value[id];
+    static __device__ double get(int64_t *value, int id) {
```
emcastillo, line 276: This trick is nice!
I was just planning on going all the way with Carray + Cindexer

asi1024: I think that using `CArray` is difficult because it has `int ndim` template parameter.

emcastillo: Yes, I did a try by setting ndim always to 1 and making it contiguous, similarly to how the arrays are set to 1d before calling the elementwise function, but I thought it will be too much boilerplate code.

emcastillo:
PTAL,
we had to replicate some functionality of elementwise kernel by using C++ templates to support array arguments
Will fix the current distributions in a different PR

cupy/random/cupy_distributions.cu
```diff
@@ -255,14 +256,50 @@ struct standard_normal_float_functor {
...
+
 template<typename F, typename T, typename R, typename... Args>
 __global__ void execute_dist(intptr_t state, intptr_t out, ssize_t size, Args... args) {
     int id = threadIdx.x + blockIdx.x * blockDim.x;
     R* out_ptr = reinterpret_cast<R*>(out);
     if (id < size) {
         T random(id, state);
         F func;
-        out_ptr[id] = func(&random, std::forward<Args>(args)...);
+        out_ptr[id] = func(&random, (pointer_fnct<std::is_pointer<Args>::value, Args>::get(args, id))...);
```
leofang, line 302: Silly question: If we check `Args` with `std::is_pointer`, doesn't it mean `Args` is unpack with only 1 type? Why don't we just make it, say,
```c++
template<typename F, typename T, typename R, typename X>
```
and avoid packing/unpacking?

leofang (commented):
LGTM except for 1 silly question

leofang:
Let's get it tested

Jenkins, test this please

cupy/random/cupy_distributions.cu
emcastillo: The function that applies this is also a variadic template, so what we do is inspect all args and when we have a pointer, it means is a cupy array, so we do the elementwise access in compile time :)

toslunar (changes requested):
- Could you test array inputs like this?  https://github.com/cupy/cupy/blob/4469fae998df33c72ff40ef954cb08b8f0004b18/tests/cupy_tests/random_tests/test_distributions.py#L459-L466
- TODO: Require `out` is C-contiguous after NumPy 1.21 is released. numpy/numpy#18731

cupy/random/cupy_distributions.cu
leofang: Ah OK. I didn't understand why the template type `Args` has to be variadic, but now I see `execute_dist` is called a few lines below with forwarded variadic arguments. Thanks.

cupy/random/cupy_distributions.cu
```diff
@@ -2,6 +2,7 @@
 #include <stdexcept>
 #include <utility>
 #include <iostream>
+#include <type_traits>
```
leofang, line 5: For `int64_t`:
```suggestion
#include <type_traits>
#include <cstdint>
```

leofang (changes requested):
Let's see if this would fix Windows build error.

emcastillo:
Jenkins, test this please

emcastillo:
chekcing to see if windows ci is fixed

emcastillo:
windows error
```
    cupy/random/cupy_distributions.cu(388): error: expression must be a pointer to a complete object type
              detected during:
                instantiation of "std::enable_if<std::is_pointer<T>::value, double>::type get_index(T, int) [with T=int64_t *]"
    (403): here
                instantiation of "void execute_dist<F,T,R,Args...>(intptr_t, intptr_t, ssize_t, Args...) [with F=standard_gamma_functor, T=curand_pseudo_state<curandState>, R=double, Args=<int64_t *>]"
    (416): here
                instantiation of "void kernel_launcher<F, R>::operator()<T,Args...>(Args &&...) [with F=standard_gamma_functor, R=double, T=curand_pseudo_state<curandState>, Args=<intptr_t &, intptr_t &, intptr_t &, int64_t *>]"
    (65): here
                instantiation of "void generator_dispatcher(int, F, Ts &&...) [with F=kernel_launcher<standard_gamma_functor, double>, Ts=<intptr_t &, intptr_t &, ssize_t &, int64_t *>]"
    (471): here

    1 error detected in the compilation of "C:/Windows/TEMP/tmpxft_00001254_00000000-10_cupy_distributions.cpp1.ii".

**Comment by emcastillo** [comment]
Jenkins, test this please

**Comment by emcastillo** [comment]
Jenkins, test this please

**Comment by emcastillo** [comment]
Jenkins, test this please

**Comment by emcastillo** [comment]
Jenkins, test this please

**Comment by emcastillo** [comment]
Jenkins, test this please

**Comment by emcastillo** [comment]
Jenkins, test this please

**Comment by emcastillo** [comment]
@toslunar I think I addressed all your comments 🙇‍♂️

File: cupy/random/_generator_api.pyx [hunk 10]
```diff
@@ -285,6 +293,107 @@ class Generator:
...
+            shape = shape.astype('d', copy=False)
+
+        if size is None:
+            size = shape.shape
+        else:
+            # if shape.shape != size
+            # we need to obtain a contiguous
+            # broadcasted view of shape
+            # if shape < size or the corresponding
+            # first slice of shape that matches size
```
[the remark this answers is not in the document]
emcastillo: Checked for both size and shape and rearranged the order of some statements to make it easier

emcastillo:
further clean up, PTAL

tests/cupy_tests/random_tests/test_generator_api.py
```diff
@@ -239,6 +239,10 @@ def invalid_contiguity(self, **kwargs):
         with pytest.raises(ValueError):
             self.generate(size=(3, 2), out=out, **kwargs)
 
+        out = cupy.zeros((4, 6), order='F', dtype=cupy.float64)
+        with pytest.raises(ValueError):
+            self.generate(size=(4, 6), out=out, **kwargs)
```
toslunar, line 244: F-contiguous `out` is allowed except for `standard_gamma`.

cupy/random/_generator_api.pyx
```diff
@@ -285,6 +293,107 @@ class Generator:
...
+            elif shape.size < t_size:
+                # We have checked before that the two arrays
+                # are broadcastable
+                # Array must be C-Contiguous
+                shape = cupy.broadcast_to(shape, size)
+
+        if out is not None:
+            self._check_output_array(dtype, size, out)
+            if out.dtype.char == 'd' and out.flags.c_contiguous:
+                y = out
```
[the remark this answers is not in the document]
toslunar: Could you add `TestStandardGammaEmpty.test_empty_out` with assertions:
```
assert y.shape == ...
assert y is out
```

emcastillo:
Comments addressed! thanks :)

cupy/random/_generator_api.pyx
```diff
@@ -351,6 +360,98 @@ class Generator:
...
+                gamma distribution.  Must be non-negative.
+                Default equals to 1
+            size (int or tuple of ints): The shape of the array.
+                If ``None``, a zero-dimensional array is generated.
+
+        .. seealso::
+            - :func:`cupy.random.gamma` for full documentation
+            - :meth:`numpy.random.Generator.gamma`
+        """
+        return self.standard_gamma(shape, size) * scale
```
toslunar, line 379:
```suggestion
        if size is None:
            size = cupy.broadcast(shape, scale).shape
        y = self.standard_gamma(shape, size)
        y *= scale
        return y
```

- `y *= scale` detects the bug below and might improve the performance. 
- `size` has to be inferred here if `scale` is an array.

```
for xp in (numpy, cupy):
    print(xp.__name__, xp.random.default_rng(0).gamma(2, scale=xp.ones(5)))
```
results
```
numpy [1.83431007 2.63769797 1.06635807 3.98054168 0.91323087]
cupy [0.77175678 0.77175678 0.77175678 0.77175678 0.77175678]
```

toslunar (changes requested):
Can I assume CuPy continues allowing `float32` computation (and `dtype` argument) in all the random operations?  In NumPy, only a few methods of the new `Generator` (and the old `randint`) support `dtype`.

emcastillo:
I think the default is `float64`,and it is used when dtype is not specified In the functions.

cupy/random/_generator_api.pyx
```diff
@@ -351,6 +360,102 @@ class Generator:
...
+        # first slice of shape that matches size
+        t_size = numpy.prod(size)
+
+        if shape.size > t_size:
+            shape = shape.ravel()[:t_size].reshape(size)
+
+        y = None
+        if out is not None:
+            self._check_output_array(dtype, size, out, True)
+            if out.dtype.char == 'd' and out.flags.c_contiguous:
```
toslunar, line 433:
```suggestion
            if out.dtype.char == 'd':
```

C-contiguity is checked by `_check_output_array`.

cupy/random/cupy_distributions.cu
```diff
@@ -362,14 +364,43 @@ struct standard_normal_float_functor {
...
+    template<typename... Args>
+    __device__ double operator () (Args&&... args) {
+        return rk_standard_gamma(args...);
+    }
+};
+
+// The following templates are used to unwrap arrays into an elementwise
+// approach
+// When a pointer is present in the variadic Args, it will be replaced by
+// the value of pointe[thread_id]
```
toslunar, line 378:
```suggestion
// the value of pointer[thread_id]
```

cupy/random/cupy_distributions.cu
```diff
@@ -362,14 +364,43 @@ struct standard_normal_float_functor {
...
+// There are several errors when trying to do this a full template
+struct standard_gamma_functor {
+    template<typename... Args>
+    __device__ double operator () (Args&&... args) {
+        return rk_standard_gamma(args...);
+    }
+};
+
+// The following templates are used to unwrap arrays into an elementwise
+// approach
```
toslunar, line 376: Could you mention that the wrapper is `_array_data` in `cupy/random/_generator_api.pyx`?

cupy/random/_generator_api.pyx
```diff
@@ -351,6 +360,102 @@ class Generator:
...
+            # Check if size is broadcastable to shape
+            # but size determines the output
+            shape = shape.astype('d', copy=False)
+
+        if size is not None and not isinstance(size, tuple):
+            size = (size,)
+        elif size is None:
+            size = shape.shape if out is None else out.shape
+
+        internal._broadcast_shapes([size, shape.shape])
```
toslunar, line 422: `shape.shape` is checked later at `shape = cupy.broadcast_to(shape, y.shape)`

toslunar (changes requested):
LGTM except for minor bugfixes and code-fixes.

cupy/random/_generator_api.pyx
toslunar: I found `cupy.random.default_rng(0).standard_gamma(-1e-15)` hangs.

How about adding
```
    } else if (shape <= 0.0) {
        return 0.0
```
to `rk_standard_gamma`?

emcastillo: sgtm!

cupy/random/_generator_api.pyx
```diff
@@ -351,6 +360,102 @@ class Generator:
...
+
+        if y is None:
+            y = ndarray(size if size is not None else (), numpy.float64)
+
+        if numpy.dtype(dtype).char not in ('f', 'd'):
+            raise TypeError(
+                f'Unsupported dtype {y.dtype.name} for standard_gamma')
+
+        if numpy.prod(size) == 0:
+            return y
```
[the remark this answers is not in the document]
emcastillo: nice! since the shape is unrolled, the kernel will be launched with 0 threads so this early return is not needed :)

cupy/random/_generator_api.pyx
```diff
@@ -351,6 +360,102 @@ class Generator:
...
+        elif size is None:
+            size = shape.shape if out is None else out.shape
+
+        internal._broadcast_shapes([size, shape.shape])
+
+        # first slice of shape that matches size
+        t_size = numpy.prod(size)
+
+        if shape.size > t_size:
+            shape = shape.ravel()[:t_size].reshape(size)
```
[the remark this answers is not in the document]
emcastillo: this seems to be fixed in a new numpy version

numpy 1.18.0
```
>>> numpy.random.default_rng(0).standard_gamma(numpy.ones((2, 3, 4)), size=(3, 4))
array([[6.79931904e-01, 1.01959710e+00, 1.98066626e-02, 2.26932668e-03],
       [5.50342873e-01, 1.62994043e+00, 6.73582953e-01, 7.55301358e-01],
       [2.81678598e+00, 6.05775308e+00, 3.28642826e+00, 1.28775033e-03]])
```

I originally tried to mimic this one :D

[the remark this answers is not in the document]
emcastillo: now since we broadcast the shape and size together, we don't need that unrolling code, so we can just delete it :)

cupy/random/_generator_api.pyx
emcastillo: I think I did this to raise some errors in the same order as numpy, but I can't remember which. Anyway I think it's better to delete it :)

toslunar (approved):

toslunar:
Jenkins, test this please.

PR #4907, opened 2021-03-19
Refactor random tests

emcastillo:
The old and new API shares a lot of methods with subtle differences in the API.

This PR tries to abstract all the infrastructure for testing the old `RandomState` so that they can be directly applied to both
`RandomState` and `Generator`


emcastillo:
rebased!

tests/cupy_tests/random_tests/common_distributions.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def get_random_state(self, xp, seed):
+        pass
+
+    def set_rs_seed(self, seed):
+        pass
+
+    def setUp(self):
+        self.__seed = testing.generate_seed()
+        self.rs = self.get_random_state(cupy, self.__seed)
```
toslunar, line 49: - Could you comment `self.rs` will be a (new or old) generator object.
- I'd rather not mixing `rs` and `random_state`.  My suggestion is to name them `rng`.

toslunar (commented):
- Could you exclude the commit bad132b17a7832da2738549e82caa64a1ec33ff8?
- Parameter sets (i.e. args of `testing.parameterize`) are still repeated, but the new tests are better than before.

tests/cupy_tests/random_tests/common_distributions.py
emcastillo: done! abstracted parameter sets too

toslunar: How about renaming methods?
- `get_random_state` -> `get_rng`
- `set_rs_seed` -> `set_rng_seed` (or `reseed_rng`)

emcastillo: done, sorry for overlooking it

toslunar (approved):
LGTM

toslunar:
Jenkins, test this please.

PR #4908, opened 2021-03-19
Use NVCC `--threads` option when building CuPy

emcastillo:


kmaehashi:
> Add support in nvcc backend for using parallel compilation at runtime too.

I think this does not make sense as compilation only runs for a single architecture at runtime.

https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-guiding-compiler-driver-threads
> This option can be used to improve the compilation speed when compiling for multiple architectures.

emcastillo:
Yes, I was thinking of the case when we have different gpus in the system, but when we compile a kernel we select the arch of the device that is going to execute it, so as you said it does not apply :)

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

PR #4910, opened 2021-03-19
Update wheel list in README

kmaehashi:


emcastillo (approved):
LGTM, readme changes not affected by CIs,
Doesn't need backport as v8 has no ROCm release

PR #4913, opened 2021-03-19
Add names for workflows

kmaehashi:
Use human readable name instead of full path.


emcastillo (approved):

emcastillo:
This will require a manual backport due to backport workflow not being present in v8

Issue #3677, opened 2020-07-28
Remove code for cuDNN v6 or earlier and NCCL v1


kmaehashi:

leofang:
Update: We have dropped cuDNN (https://github.com/cupy/cupy/pull/9326).

PR #4919, opened 2021-03-19
Move the NCCL module to `cupy_backends.cuda.libs`

leofang:
Prepare for #4873.

The change is a three-step one:
1. Move and break the header `cupy/cuda/cupy_nccl.h` into 3 pieces (for CUDA, HIP, and RTD)
2. Move `cupy/cuda/nccl.{pxd, pyx}` to `cupy_backends/cuda/libs`
3. Add `cupy/cuda/nccl.py` to import the moved module


cupy/cuda/nccl.py
```diff
@@ -0,0 +1,14 @@
+"""
+NCCL Wrapper
+
+Use `cupy_backends.cuda.libs.nccl` directly in CuPy codebase.
```
kmaehashi, line 4: `cupyx/_runtime.py` uses `cupy.cuda.nccl`, so it should be updated.

leofang: Good catch, thanks! Fixed in 437ee18.

kmaehashi:
pfnCI, test this please.

leofang:
Note: we still need to do some cleanup for the NCCL codes (ex: #3677), but I prefer to not touch it in this PR. It's not urgent I think.

leofang:
Hmm...I thought I fixed it. Let me check.

leofang:
The fix 67289a7 is due to https://github.com/cupy/cupy/pull/4920#discussion_r597843135.

leofang:
Jenkins, test this please

cupy/cuda/__init__.py
```diff
@@ -51,10 +51,7 @@ def __init__(self, name):
 except ImportError:
     thrust = _UnavailableModule('cupy.cuda.thrust')
 
-try:
-    from cupy.cuda import nccl  # NOQA
-except ImportError:
-    nccl = _UnavailableModule('cupy.cuda.nccl')
+from cupy.cuda import nccl  # NOQA
```
kmaehashi, line 54: This line needs to be removed.

leofang: Fixed in 9e8192b.

leofang:
Jenkins, test this please

tests/cupy_tests/cuda_tests/test_nccl.py
```diff
@@ -19,60 +20,60 @@ class TestNCCL(unittest.TestCase):
 
     @testing.gpu
     def test_single_proc_ring(self):
-        id = cuda.nccl.get_unique_id()
-        comm = cuda.nccl.NcclCommunicator(1, id, 0)
+        id = nccl.get_unique_id()
```
kmaehashi, line 23: FYI (no need to revert this), you didn't have to change `cuda.nccl` to `nccl` just by adding `import cupy.cuda.nccl` 😄

kmaehashi (approved):

leofang:
Thanks!

PR #4920, opened 2021-03-19
Add `cupy/cuda/cutensor.py`

leofang:
Prepare for #4873. ~~Blocked by #4919.~~

The change starts at 1d55ca52.

### DIFF
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -51,8 +51,6 @@ try:
 except ImportError:
     thrust = _UnavailableModule('cupy.cuda.thrust')
 
-from cupy.cuda import cutensor  # NOQA
-
 
 def __getattr__(key):
     # `*_enabled` flags are kept for backward compatibility.
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -51,10 +51,7 @@ try:
 except ImportError:
     thrust = _UnavailableModule('cupy.cuda.thrust')
 
-try:
-    from cupy_backends.cuda.libs import cutensor
-except ImportError:
-    cutensor = _UnavailableModule('cupy.cuda.cutensor')
+from cupy.cuda import cutensor  # NOQA
 
 
 def __getattr__(key):
@@ -68,7 +65,7 @@ This flag always returns True as cuSOLVER is always available in CUDA 8.0 or lat
             ''', DeprecationWarning)  # NOQA
         return True
 
-    for mod in [nvtx, thrust, cub, cutensor]:
+    for mod in [nvtx, thrust, cub]:
         flag = '{}_enabled'.format(mod.__name__.split('.')[-1])
         if key == flag:
             warnings.warn('''
--- a/cupy/core/_routines_math.pyx
+++ b/cupy/core/_routines_math.pyx
@@ -21,10 +21,10 @@ from cupy.cuda cimport memory
 
 from cupy.cuda import cub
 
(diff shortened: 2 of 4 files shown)


cupy/cuda/__init__.py
```diff
@@ -51,15 +51,7 @@ def __init__(self, name):
...
-try:
-    from cupy.cuda import nccl  # NOQA
-except ImportError:
-    nccl = _UnavailableModule('cupy.cuda.nccl')
-
-try:
-    from cupy_backends.cuda.libs import cutensor
-except ImportError:
-    cutensor = _UnavailableModule('cupy.cuda.cutensor')
+from cupy.cuda import cutensor  # NOQA
```
leofang, line 54: (sorry, repost to make the discussion appear in the PR.)

Looks like there are a number of places in the codebase expecting to do `if cupy.cuda.cutensor.available: ...`, so I had to import `cutensor` here, making the treatment a bit asymmetric w.r.t. other two optional dependencies (cudnn and nccl). @kmaehashi WDYT?

kmaehashi: How about rewriting `if cupy.cuda.cutensor.available:` with this?

```py
try:
    import cupy_backends.cuda.libs.cutensor
except ImportError:
    pass
```

leofang: Wouldn't that make the `available` flag entirely useless?

leofang: Perhaps this is better?
```python
from cupy.cuda import cutensor
if cutensor.available:
    ...
```

kmaehashi: I think `available` flag is still useful for end-users. e.g, `if cupy.cuda.cutensor.available: import cupy.cutensor`

leofang: thanks, done in a18fe9e.

leofang: So, without importing the wrapper modules in `cupy/cuda/__init__.py`, end users would see `AttributeError` when doing `if cupy.cuda.cutensor.available: ...`. I think they still need to be imported...

leofang: Sorry, @kmaehashi I think I confused myself 😂 Let me summarize:
1. If `nccl`/`cutensor` is imported in `cupy/cuda/__init__.py`:
    - We restore the backward compatibility (as nothing is lost), in particular users can still write `if cupy.cuda.cutensor.available: ...`
    - but we bring in warnings at cupy import time if the library is not installed
2. If `nccl`/`cutensor` is NOT imported in `cupy/cuda/__init__.py`:
    - We lose backward compatibility, so user codes like `if cupy.cuda.cutensor.available: ...` will stop working
    - warnings are not raised until the module is actually needed

I suppose Option No.2 is what you are looking for?

kmaehashi: Yes, Option 2 is what I meant.  User code `import cupy.cuda; if cupy.cuda.cutensor.available` will stop working, and users need to rewrite it as `import cupy.cuda.cutensor; if cupy.cuda.cutensor.available`. `import cupy.cuda.cutensor` should generate a warning, but `import cupy.cuda` should not.

leofang: Thanks for bearing with me @kmaehashi. Will fix the other PR shortly.

kmaehashi:
Merged #4919!

leofang:
Thanks, @kmaehashi! Done.

Jenkins, test this please

kmaehashi (approved):

PR #4264, opened 2020-11-11
Replace stubs in ndimage by Jitify

leofang:
Follow-up of #4228. We no longer need customized C++ std code now that we have Jitify.

TODO:
- [x] Test HIP

PR #4922, opened 2021-03-19
Fix C++ compilation error

### LINKED ISSUE #4875: [BUG] NVRTC does not support <type_traits> out of box
On the current master, this snippet
```python
import cupy as cp

code = r'''
#include <cupy/carray.cuh>

extern "C" __global__ void fff(float* a) {

   int tid = threadIdx.x;
   a[tid] += 1;
}
'''

mod = cp.RawModule(code=code, options=('-std=c++11',))
ker = mod.get_function("fff")
```
raises an exception:
```
cupy.cuda.compiler.CompileException: /home/leofang/dev/cupy_cuda112/cupy/core/include/cupy/carray.cuh(4): catastrophic error: cannot open source file "type_traits"
```
This is a regression due to #3683 adding this line:
https://github.com/cupy/cupy/blob/fa21dd5d43ea93a48ccfa3bd796da101cf4c1c7c/cupy/core/include/cupy/carray.cuh#L4
NVRTC does not support most of the C++ std headers, which is why we need Jitify. I don't know how come all tests managed to pass so far...🤔

leofang:
Close #4875.

As mentioned in the linked issue, NVRTC by default does not support many C++ features, including the offending `type_traits` header. Since it is best to not revert a feature we already supported (#3683), I choose to add a workaround patch, which in this case is still manageable though unfortunately against my earlier intention to avoid custom C++ `std` stubs (ex: #4264).

The added test ensures an error like the one reported in #4875 can be detected by the CI and prevents a regression from happening again.

cc: @takagi @coderforlife @grlee77

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1858,16 +1858,19 @@ cdef bint _is_hip = runtime._is_hip_environment
 cdef int _cuda_runtime_version = -1
 cdef str _cuda_path = ''  # '' for uninitialized, None for non-existing
 
-cdef list _cupy_header_list = [
+cdef list cupy_header_list = [
     'cupy/complex.cuh',
     'cupy/carray.cuh',
     'cupy/atomics.cuh',
 ]
 if _is_hip:
-    _cupy_header_list.append('cupy/math_constants.h')
+    cupy_header_list.append('cupy/math_constants.h')
+
+# expose to Python for unit testing
+_cupy_header_list = cupy_header_list
 
 cdef str _cupy_header = ''.join(
-    ['#include <%s>\n' % i for i in _cupy_header_list])
+    ['#include <%s>\n' % i for i in cupy_header_list])
 
 # This is indirect include header list.
 # These header files are subject to a hash key.
@@ -1917,7 +1920,7 @@ cpdef str _get_header_source():
     if _header_source is None or not _header_source_map:
         source = []
         base_path = _get_header_dir_path()
-        for file_path in _cupy_extra_header_list + _cupy_header_list:
+        for file_path in _cupy_extra_header_list + cupy_header_list:
             header_path = os.path.join(base_path, file_path)
             with open(header_path) as header_file:
                 header = header_file.read()
@@ -1957,7 +1960,7 @@ cpdef function.Module compile_with_cache(
         log_stream=None, bint jitify=False):
     if translate_cucomplex:
         source = _translate_cucomplex_to_thrust(source)
-        _cupy_header_list.append('cupy/cuComplex_bridge.h')
+        cupy_header_list.append('cupy/cuComplex_bridge.h')
         prepend_cupy_headers = True
 
     if prepend_cupy_headers:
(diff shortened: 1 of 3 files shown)


coderforlife:
This is a much better approach to this. That std::enable_if is also used at https://github.com/cupy/cupy/blob/a0aa435b9459c2d5b9a1266edc6f670a1c903e34/cupyx/scipy/ndimage/_filters_core.py#L164-L170.

leofang:
@coderforlife Do you think I should also substitute `std::enable_if` in `cupy/cupyx/scipy/ndimage/_filters_core.py`?

leofang:
Jenkins, test this please

cupy/core/core.pyx
```diff
@@ -1858,16 +1858,19 @@ cdef bint _is_hip = runtime._is_hip_environment
 cdef int _cuda_runtime_version = -1
 cdef str _cuda_path = ''  # '' for uninitialized, None for non-existing
 
-cdef list _cupy_header_list = [
+cdef list cupy_header_list = [
```
leofang, line 1861: This is a `cdef` variable defined in `.pyx`. I don't think we need to mark it private (by prepending an underscore), as it already is. No functions (either in Cython or Python) outside this module can access it.

tests/cupy_tests/core_tests/test_core.py
```diff
@@ -1,11 +1,13 @@
 import unittest
+import sys
 
 import numpy
 import pytest
 
 import cupy
 from cupy.core import core
 from cupy import testing
+from .test_raw import use_temporary_cache_dir
```
takagi, line 10: Would you use an absolute module path?

takagi (changes requested):
LGTM except for one guideline issue!

tests/cupy_tests/core_tests/test_core.py
leofang: Done in 00d3594.

leofang:
Jenkins, test this please

tests/cupy_tests/core_tests/test_core.py
```diff
@@ -1,11 +1,13 @@
 import unittest
+import sys
 
 import numpy
 import pytest
 
 import cupy
 from cupy.core import core
 from cupy import testing
+from tests.cupy_tests.core_tests import test_raw
```
kmaehashi, line 10:
```suggestion
from cupy_tests.core_tests import test_raw
```

leofang: Thanks, @kmaehashi. I feel absolute importing from other test modules is shaky...

leofang:
Jenkins, test this please

takagi:
Jenkins, test this please

takagi (approved):

leofang:
@takagi NVRTC in CUDA < 11.1 does not support `--std=c++03`, so I took it away. On CUDA 11.2, I tested that it works.

leofang:
Jenkins, test this please

takagi (approved):

takagi:
Thanks! Will be merged soon once it passes.

PR #4927, opened 2021-03-20
Add poisson distribution to random API

povinsahu1909:


cupy/random/cupy_distributions.cu
```diff
@@ -164,6 +164,61 @@ __device__ double rk_beta(rk_state* state, double a, double b) {
...
+            continue;
+        }
+        if ((log(V) + log(invalpha) - log(a/(us*us)+b)) <=
+            (-lam + k*loglam - loggam(k+1))) {
+            return k;
+        }
+    }
+}
+
+__device__ long rk_poisson(rk_state *state, double lam) {
```
emcastillo, line 212: is the return type long correct here? I think it should be double

povinsahu1909: I use long because according to the NumPy docs 'output is limited to the range of the C int64 type' and also the current implementation of poisson distribution in RandomState module uses long as a return type.

leofang: If NumPy is right on the `int64` return type, then this must be `long long`, not `long`, because on Windows a `long` is 32 bits, not 64. See LP64 vs LLP64: https://docs.nvidia.com/cuda/nvrtc/index.html#integer-size.

leofang: (I didn't read the code here, just wanna point out `long` is not portable. So don't mind me if this is irrelevant of the implementation here. It's just that I spent quite some time fixing Windows errors, so wanna make sure we get it right from now on 😁)

emcastillo: let's use `int64_t`

emcastillo: I just confirmed that the datatype in windows is `int64`

povinsahu1909:
While going through cuRAND API reference guide, I saw a function for generating value from a poisson distribution similar to what used in the current implementation. Can we use this function for our implementation?

https://docs.nvidia.com/cuda/curand/device-api-overview.html#distributions

emcastillo:
the curand function returns a single 32 bit integer while we have the int64 range, so it's better to use the current one to ensure compatibility with numpy.

emcastillo:
Sorry, can you rebase to fix the conflicts? thanks!

emcastillo (dismissed):
Great work! thanks!

emcastillo:
Jenkins, test this please

emcastillo:
```
15:30:53     cupy/random/cupy_distributions.cu(209): warning: expression has no effect
15:30:53 
15:30:53     cupy/random/cupy_distributions.cu(221): error: identifier "loggam" is undefined
15:30:53 
```

povinsahu1909:
Oops! I'll fix this.

emcastillo (approved):

emcastillo:
Jenkins, test this please

povinsahu1909:
I'll try to fix this and will update this tomorrow

povinsahu1909:
Apologies for the delay in updating the PR.

emcastillo:
Jenkins, test this please

emcastillo:
cupy/random/cupy_distributions.cu(199): error: identifier "M_PI" is undefined

povinsahu1909:
> cupy/random/cupy_distributions.cu(199): error: identifier "M_PI" is undefined

I'm not aware of this error. Can you please guide me on why this error occurred, as this math constant was used in RandomState class? 

I searched for the possible cause of this error and found [this](https://stackoverflow.com/questions/56319494/nvcc-compilation-errors-with-m-pi-and-or). Is this applies here?

emcastillo:
The jenkins failure is a flaky one, ignore it.

The `M_PI` seems to be a weird windows issue:
https://stackoverflow.com/questions/56319494/nvcc-compilation-errors-with-m-pi-and-or

adding `-D_USE_MATH_DEFINES` around here 
https://github.com/cupy/cupy/blob/master/cupy_setup_build.py#L655

as 
```python
      if module['name'] == 'random':
         if compiler.compiler_type == 'msvc':
                compile_args.append('-D_USE_MATH_DEFINES')
```
should fix it

emcastillo:
Jenkins, test this please

leofang:
Windows build still fails. I think @povinsahu1909 you should also add that macro definition to this line:
https://github.com/cupy/cupy/blob/cc6b3e583071d668ed1f3fc6e3f305526f2c54dd/cupy_setup_build.py#L1087
so that nvcc picks it up during the 2-pass compilation (host and device codes).

leofang:
Jenkins, test this please

emcastillo (approved):

PR #4929, opened 2021-03-21
Fix to work with scheduled FlexCI job

kmaehashi:
FlexCI does not define `FLEXCI_BRANCH` when run as scheduled job.


emcastillo (approved):

emcastillo:
dont need jenkins test

PR #4853, opened 2021-03-09
Fix cuFFT callback compilations - v2

leofang:
Close #4746.

This is an alternative of #4743 (proposal v2 mentioned there). The solution here is to avoid cimporting any CuPy modules so that we don't expose Cython-level APIs. The incurred overhead is OK ~~(which I will post later)~~: about +/- 5 us (yes, it can be faster in some cases, which I didn't expect), see https://github.com/cupy/cupy/pull/4853#issuecomment-794903196.

TODO:
- [x] Fix https://github.com/cupy/cupy/issues/4746#issuecomment-785543235

PR #4088, opened 2020-10-07
Clean up `cupy_backends`

leofang:
Follow up of #3616. 

***UPDATE***: See https://github.com/cupy/cupy/pull/4088#issuecomment-745784398. Ignore the original PR description below.

----

~~This PR is to ensure compliance of the proposed directory structure (https://github.com/cupy/cupy/pull/3616#issuecomment-699662624), which was not fully honored. Specifically, instead of doing an import like this~~
```python
from cupy_backends.cuda.api import runtime
```
~~we now do~~
```python
from cupy_backends.api import runtime  # "cuda" is gone
```
~~making it neutral of the backend provider~~

PR #3685, opened 2020-07-29
Fix broken imports for `cupy.cuda.*`

kmaehashi:
Supersedes #3685 

This fixes the issue that `import cupy.cuda.{runtime,profiler,cudnn}` is not working after cupy_backends refactoring.
discussion: https://github.com/cupy/cupy/pull/3681#issuecomment-665093779

(although the code freeze period is already over, we will include this fix to v8b5 as this will break user code)

PR #4930, opened 2021-03-21
Move the NVTX module to `cupy_backends.cuda.libs`

leofang:
Follow-up of #4919, #4920. 

This PR moves the `nvtx` module out of `cupy.cuda`, which makes better sense now that we put all CUDA modules that link to vendor libraries under `cuda_backends/cuda/{api, libs}`. The rest of CUDA modules in `cupy.cuda` are either CuPy-specific, or do not link to a particular CUDA component (ex: Thrust/CUB/Jitify are all header-only); the `cufft` module is the only exception due to the need of supporting FFT callbacks (#4853).

This also makes sense for rocTX, for which we already put the stub in `cupy_backends/hip/cupy_roctx.h` (#4088).

This PR does not break backward compatibility.


leofang:
Not sure what caused this error: 
https://github.com/cupy/cupy/pull/4930/checks?check_run_id=2158104374#step:7:881

leofang:
I took a closer look and still don't understand the error. Like `runtime`, `nvtx` is now also imported from `cupy_backends`, but Sphinx has no problem in documenting the functions from the `runtime` module.

leofang:
> I took a closer look and still don't understand the error. Like `runtime`, `nvtx` is now also imported from `cupy_backends`, but Sphinx has no problem in documenting the functions from the `runtime` module.

Oh... I got it, there's a `cupy/cuda/runtime.py` just to make Sphinx happy...#3685

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

emcastillo (approved):

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

PR #4932, opened 2021-03-21
Document that cuDNN and NCCL are no longer included

### LINKED ISSUE #4850: Do not bundle cuDNN / NCCL for all wheel packages starting in CuPy v9

kmaehashi:
Closes #4850
Blocked by https://github.com/cupy/cupy-release-tools/pull/81, #4848, #4852

### DIFF
--- a/docs/source/upgrade.rst
+++ b/docs/source/upgrade.rst
@@ -30,6 +30,11 @@ Dropping Support of Python 3.5
 
 Python 3.5 is no longer supported in CuPy v9.
 
+NCCL and cuDNN No Longer Included in Wheels
+-------------------------------------------
+
+NCCL and cuDNN shared libraires are no longer included in wheels (see `#4850 <https://github.com/cupy/cupy/issues/4850>`_ for discussions). 
+You can manually install them after installing wheel; see :doc:`install/install` for details.
 
 CuPy v8
 =======
--- a/docs/source/upgrade.rst
+++ b/docs/source/upgrade.rst
@@ -34,7 +34,7 @@ NCCL and cuDNN No Longer Included in Wheels
 -------------------------------------------
 
 NCCL and cuDNN shared libraires are no longer included in wheels (see `#4850 <https://github.com/cupy/cupy/issues/4850>`_ for discussions). 
-You can manually install them after installing wheel; see :doc:`install/install` for details.
+You can manually install them after installing wheel if you don't have a previous installation; see :doc:`install/install` for details.
 
 CuPy v8
 =======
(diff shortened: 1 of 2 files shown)


leofang:
If a user already installed a compatible library say NCCL, and it's discoverable through LD_LIBRARY_PATH, I suppose everything would work out of box (i.e., no need to run installer)?

docs/source/install/install.rst
```diff
@@ -91,13 +91,10 @@ Package names are different depending on your CUDA Toolkit version.
 
 .. note::
 
-   Wheel packages are built with NCCL (Linux only) and cuDNN support enabled.
+   To enable features provided by additional CUDA libraries (cuTENSOR / NCCL / cuDNN), you need to install them manually.
+   If you installed CuPy via wheels, you can use the installer command below to setup these libraries.
```
emcastillo, line 95:
```suggestion
   If you installed CuPy via wheels, you can use the installer command below to setup these libraries in case you don't have a previous installation.
```

docs/source/upgrade.rst
```diff
@@ -30,6 +30,11 @@ Dropping Support of Python 3.5
 
 Python 3.5 is no longer supported in CuPy v9.
 
+NCCL and cuDNN No Longer Included in Wheels
+-------------------------------------------
+
+NCCL and cuDNN shared libraires are no longer included in wheels (see `#4850 <https://github.com/cupy/cupy/issues/4850>`_ for discussions). 
+You can manually install them after installing wheel; see :doc:`install/install` for details.
```
emcastillo, line 37:
```suggestion
You can manually install them after installing wheel if you don't have a previous installation; see :doc:`install/install` for details.
```

kmaehashi:
Thanks, applied suggestions!
pfnCI, test this please.

kmaehashi:
Rebased!
pfnCI, test this please.

emcastillo (approved):

PR #4939, opened 2021-03-22
Support tuple indexing in CuPy JIT

asi1024:
Depends on #4890

### DIFF
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -546,18 +546,47 @@ def _transpile_expr_internal(expr, env):
     if isinstance(expr, ast.Subscript):
         value = _transpile_expr(expr.value, env)
         index = _transpile_expr(expr.slice, env)
-        if is_constants([value, index]):
-            return Constant(value[index])
-        value = _to_cuda_object(value, env)
-        index = _to_cuda_object(index, env)
 
-        if not isinstance(value.ctype, _types.Array):
-            raise ValueError(f'{value.code} must be Array type.')
-        if value.ctype.ndim != 1:
-            raise NotImplementedError('Not implemented for ndim > 1.')
-        return CudaObject(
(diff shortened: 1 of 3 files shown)


asi1024:
@emcastillo Rebased. PTAL!

emcastillo (dismissed):

cupyx/jit/_compile.py
```diff
@@ -546,18 +546,47 @@ def _transpile_expr_internal(expr, env):
...
+                    f'{value.code}[{index.code}]',
+                    _types.Scalar(value.ctype.dtype))
+            if isinstance(index.ctype, _types.Tuple):
+                if ndim != len(index.ctype.types):
+                    raise IndexError(f'The size of index must be {ndim}')
+                for t in index.ctype.types:
+                    if not isinstance(t, _types.Scalar):
+                        raise TypeError('Array indices must be scalar.')
+                    if t.dtype.kind not in 'iu':
+                        raise TypeError('Array indices must be integer.')
```
emcastillo, line 581: I wonder if an empty tuple would hard error before arriving here.

emcastillo:
Jenkins, test this please

cupyx/jit/_compile.py
asi1024: Empty tuples will be caught here:
https://github.com/cupy/cupy/pull/4939/files#diff-92c76e1dfae3dd9e376fac22b852d0deb6cea6d0affa13bda3ac7d2ee33ce17dR575-R576
, but indexing of 0-dim array is not tested. I will add the test in this PR.

asi1024: I noticed 0-dim array does not have `_indexing` member function. Fixed in cbf02f3!

emcastillo (approved):

emcastillo:
Jenkins, test this please

PR #4942, opened 2021-03-22
Added the implementation of method Hyperbolic Sine  Function (sinhm) for Matrix

ameyachawla99:
Hi , I am Ameya Chawla currently pursuing my B.Tech in Computer Science and Engineering .
I read about your organization in GSOC 2021 and looking forward to work with your organization .

Added the implementation of method "Hyperbolic Sine  Function"(sinhm)
as mentioned in the comparison document
https://docs.cupy.dev/en/stable/reference/comparison.html

PR #4944, opened 2021-03-22
Implemented the Hyperbolic Sine Function for matrix 

ameyachawla99:
Hi , I am Ameya Chawla currently pursuing my B.Tech in Computer Science and Engineering .
I read about your organization in GSOC 2021 and looking forward to work with your organization .

Added the implementation of method "Hyperbolic Sine Function"(sinhm)
as mentioned in the comparison document

As discussed in #4942 i have added code to  my code to matfuncs.py
https://docs.cupy.dev/en/stable/reference/comparison.html

### DIFF
--- a/cupyx/scipy/linalg/matfuncs.py
+++ b/cupyx/scipy/linalg/matfuncs.py
@@ -1,0 +1,9 @@
+from numpy import exp
+
+
+def sinhm(arr):
+    """Returns hyperbolic sine of the given square matrix.
+Args: arr: square matrix .
+Returns: Hyperbolic sine of given square matrix as input.
+    """
+    return 0.5*(exp(arr)-exp(-1*arr))
--- a/cupyx/scipy/linalg/matfuncs.py
+++ b/cupyx/scipy/linalg/matfuncs.py
@@ -4,12 +4,13 @@ from numpy import exp
 def sinhm(arr):
     """Returns hyperbolic sine of the given square matrix.
 
-    Args:. 
+Args:. 
         arr(cupy.ndarray) : Square matrix whose hyperbolic .
         sine has to be calculated .
 
-    Returns: 
-    
+Returns: 
+
         (cupy.ndarray):  Hyperbolic sine of given square matrix as input.
     """
     return 0.5 * (exp(arr) - exp(-1 * arr))
+
--- a/cupyx/scipy/linalg/matfuncs.py
+++ b/cupyx/scipy/linalg/matfuncs.py
@@ -3,10 +3,11 @@ from numpy import exp
 
 def sinhm(arr):
     """Returns hyperbolic sine of the given square matrix.
-    
-    Args: 
-        arr(cupy.ndarray) : Square matrix whose hyperbolic sine has to be calculated .
-
+.    
+    Args:. 
+        arr(cupy.ndarray) : Square matrix whose hyperbolic .
+        sine has to be calculated .
+.
     Returns: 
     
         (cupy.ndarray):  Hyperbolic sine of given square matrix as input.
--- a/cupyx/scipy/linalg/matfuncs.py
+++ b/cupyx/scipy/linalg/matfuncs.py
@@ -13,7 +13,7 @@ def sinhm(arr):
 
                         Hyperbolic sine of given square matrix as input.
 
-    ..seealso:: :func: 'scipy.linlag.matfuncs.py'                
+    ..seealso:: :func: 'scipy.linlag.matfuncs.py'           
     """
 
     arr = numpy.asanyarray(arr)
--- a/cupyx/scipy/linalg/matfuncs.py
+++ b/cupyx/scipy/linalg/matfuncs.py
@@ -3,7 +3,12 @@ from numpy import exp
 
 def sinhm(arr):
     """Returns hyperbolic sine of the given square matrix.
-Args: arr: square matrix .
-Returns: Hyperbolic sine of given square matrix as input.
+    
+    Args: 
+    arr(cupy.ndarray) : Square matrix whose hyperbolic sine has to be calculated .
+
+    Returns: 
+    
+    (cupy.ndarray):  Hyperbolic sine of given square matrix as input.
     """
     return 0.5 * (exp(arr) - exp(-1 * arr))
--- a/cupyx/scipy/linalg/matfuncs.py
+++ b/cupyx/scipy/linalg/matfuncs.py
@@ -2,13 +2,15 @@ from numpy import exp
 
 
 def sinhm(arr):
-    """Returns hyperbolic sine of the given square matrix.
+    """Hyperbolic Sine
+    Returns hyperbolic sine of the given square matrix.
 
-Args:.
+    Args:.
+
         arr(cupy.ndarray) : Square matrix whose hyperbolic .
         sine has to be calculated .
 
-Returns: .
+    Returns: .
 
         (cupy.ndarray):  Hyperbolic sine of given square matrix as input.
     """
--- a/cupyx/scipy/linalg/matfuncs.py
+++ b/cupyx/scipy/linalg/matfuncs.py
@@ -8,9 +8,8 @@ Args:.
         arr(cupy.ndarray) : Square matrix whose hyperbolic .
         sine has to be calculated .
 
-Returns: 
+Returns: .
 
         (cupy.ndarray):  Hyperbolic sine of given square matrix as input.
     """
     return 0.5 * (exp(arr) - exp(-1 * arr))
-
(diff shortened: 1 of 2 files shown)


emcastillo:
please add tests

ameyachawla99:
Where should i add test cases?

emcastillo:
In tests/cupyx_tests/scipy_tests/linalg_tests folder.

cupyx/scipy/linalg/matfuncs.py
```diff
@@ -0,0 +1,9 @@
+from numpy import exp
+
+
+def sinhm(arr):
+    """Returns hyperbolic sine of the given square matrix.
+Args: arr: square matrix .
+Returns: Hyperbolic sine of given square matrix as input.
```
emcastillo, line 7: Please, format the docstring with identation and line breaks

cupyx/scipy/linalg/matfuncs.py
```diff
@@ -0,0 +1,9 @@
+from numpy import exp
+
+
+def sinhm(arr):
+    """Returns hyperbolic sine of the given square matrix.
+Args: arr: square matrix .
+Returns: Hyperbolic sine of given square matrix as input.
+    """
+    return 0.5*(exp(arr)-exp(-1*arr))
```
emcastillo, line 9:
```suggestion
    return 0.5 * (exp(arr) - exp(-1 * arr))
```
We leave a space between operands.

cupyx/scipy/linalg/matfuncs.py
```diff
@@ -0,0 +1,15 @@
+from numpy import exp
+
+
+def sinhm(arr):
+    """Returns hyperbolic sine of the given square matrix.
+
+Args:.
```
emcastillo, line 7: Please look other cupy functions docstring and use the same format:

https://github.com/cupy/cupy/blob/master/cupyx/scipy/linalg/decomp_lu.py#L13-L35

emcastillo:
you can install flake8 locally with `pip install flake8` and check the file with `flake8 cupyx/scipy/linalg/matfuncs.py` so you don't need to push everytime to check the changes.

cupyx/scipy/linalg/matfuncs.py
ameyachawla99: I will make the necessary changes.

ameyachawla99:
I have made changes in matfuncs.py , please can your guide me how should I make the test file?

emcastillo:
look for other tests, ideally you should add tests for edge cases and common ones, empty arrays, invalid arguments, etc.

kmaehashi:
Tests reside in the directory with `test_` prefixed for the corresponding module name.
For example, tests code for `cupyx/scipy/linalg` will be in `tests/cupyx_tests/scipy_tests/linalg_tests`. 

To start reviewing, please write (at least) one most simple test case to verify your implementation is correct, and confirm that the test pass **in your local machine** (this is important!) before pushing it to the pull-request.

ameyachawla99:
I am also adding the conditions for edge cases as if wrong input given by user should produce error .

ameyachawla99:
I have added the edge cases and test file .

cupyx/scipy/linalg/matfuncs.py
```diff
@@ -0,0 +1,31 @@
...
+    Returns hyperbolic sine of the given square matrix.
+    Args:.
+        arr(cupy.ndarray) : Square matrix whose hyperbolic .
+        sine has to be calculated .
+    Returns: .
+        (cupy.ndarray):  Hyperbolic sine of given square matrix as input.
+    """
+    arr = numpy.asanyarray(arr)
+
+    "Checking whether the input is a 2D matrix or not"
```
kmaehashi, line 15: For single-line comments use `#` instead.

cupyx/scipy/linalg/matfuncs.py
```diff
@@ -0,0 +1,31 @@
+import numpy
+
+
+def sinhm(arr):
+    """Hyperbolic Sine
+    Returns hyperbolic sine of the given square matrix.
+    Args:.
```
kmaehashi, line 7: Please follow the doc style of other functions.

tests/cupyx_tests/scipy_tests/linalg_tests/test_matfuncs.py
```diff
@@ -0,0 +1,32 @@
+from matfuncs import sinhm
+import numpy
+import unittest
+
+
+class TestMatFuncsNan(unittest.TestCase):
+
+    "Testing for nan values in matrix"
```
kmaehashi, line 8: Comments for functions must be written inside the function.

tests/cupyx_tests/scipy_tests/linalg_tests/test_matfuncs.py
```diff
@@ -0,0 +1,32 @@
+from matfuncs import sinhm
+import numpy
+import unittest
+
+
+class TestMatFuncsNan(unittest.TestCase):
```
kmaehashi, line 6: It seems your test cases only cover anomaly cases, and using NumPy array as an input.

kmaehashi (changes requested):

tests/cupyx_tests/scipy_tests/linalg_tests/test_matfuncs.py
ameyachawla99: I was converting array in input as NumPy array to use functions defined for NumPy array like shape.

ameyachawla99: Which other test cases should i also add?

ameyachawla99: I have edited the files .
Please guide me for any further changes .

asi1024: @ameyachawla99 Did you check if this test passes locally?

ameyachawla99: Yes all 5 tests were passes 
i have added the screenshot of terminal .
Screenshot from 2021-03-27 12-25-38

asi1024: Why `from matfuncs import sinhm` works? `matfuncs.py` is not placed in test directory.

asi1024: Anyway, this test code looks to just call NumPy routines. Could you use `cupy.ndarray` for the input?

ameyachawla99: > Why `from matfuncs import sinhm` works? `matfuncs.py` is not placed in test directory.

While i was testing i placed matfuncs in same directory .

ameyachawla99: > call NumPy routines

Sure i will update it .

ameyachawla99: > Anyway, this test code looks to just call NumPy routines. Could you use `cupy.ndarray` for the input?

But then changes has to be made in both files as error will be caused of cupy.ndarray as not of same datatype of numpy.ndarray

ameyachawla99: @asi1024 i have made the changes in both files so that they can use cupy.ndarray instead of numpy.ndarray

asi1024: Could you add an alias `cupyx.scipy.linalg.sinhm` and use it in the test?

asi1024: SciPy's documentation says that `scipy.linalg.sinhm` uses **matrix exponentials**, but your implementation uses element-wise exponential.  https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.sinhm.htm.

ameyachawla99: Then i will first define expm which is used 

> SciPy's documentation says that `scipy.linalg.sinhm` uses **matrix exponentials**, but your implementation uses element-wise exponential. https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.sinhm.htm.

Then i will first define expm

kmaehashi: Oh, sorry I overlooked that `cupyx.scipy.sparse.linalg.expm` is not implemented yet... I should have noticed it and comment that in the first review round. You can either try implementing it in this PR or in a separate PR.

ameyachawla99: Actually it is my fault i did not understood the function properly

PR #4951, opened 2021-03-23
Declare `CArray._indexing()` only in CuPy JIT mode

asi1024:
Enclosed the `CArray._indexing` declarations with `#ifdef CUPY_JIT_MODE`, because ROCm tries to find `thrust::get` before the expansion of template member functions whereas CUDA does not.


takagi (approved):
LGTM! I've checked this change works on AMD's cluster.

takagi:
pfnCI, test this please.

PR #4955, opened 2021-03-23
First step of `mypy` introduction

asi1024:
Introduces mypy 0.931 for static type checks.

### DIFF
--- a/.github/workflows/pretest.yml
+++ b/.github/workflows/pretest.yml
@@ -50,6 +50,10 @@ jobs:
         autopep8 -r . --diff | tee check_autopep8
         test ! -s check_autopep8
 
+    - name: Static Type Checking (mypy)
+      run: |
+        mypy cupy cupyx cupy_backends
+
     - name: Test - Docs Build
       run: |
         # Test to detect invalid escape sequences in docstrings (#1619)
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,3 +37,7 @@ xfail_strict=true
 
 [metadata]
 license_files = docs/source/license.rst
+
+[mypy]
+ignore_missing_imports = True
+disable_error_code = attr-defined
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,8 @@ extras_require = {
         'flake8==3.8.4',
         'pbr==5.5.1',
         'pycodestyle==2.6.0',
+        'types-setuptools==57.4.8',
+        'mypy==0.931',
     ],
     'test': [
         # 4.2 <= pytest < 6.2 is slow collecting tests and times out on CI.
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ build
 .\#*
 .coverage
 .eggs/
+.mypy_cache
 _readthedocs_build
 /TAGS
 /docs/source/reference/**/generated
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -16,15 +16,15 @@ _environment._preload_library('nccl')  # NOQA
 
 try:
     from cupy import _core  # NOQA
-except ImportError as e:
+except ImportError as exc:
     raise ImportError(f'''
 ================================================================
 {_environment._diagnose_import_error()}
 
 Original error:
-  {type(e).__name__}: {e}
+  {type(exc).__name__}: {exc}
 ================================================================
-''') from e
+''') from exc
 
 
 from cupy import cuda  # NOQA
--- a/cupy/_core/__init__.py
+++ b/cupy/_core/__init__.py
@@ -1,3 +1,5 @@
+# mypy: ignore-errors
+
 from cupy._core import core  # NOQA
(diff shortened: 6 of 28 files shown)


leofang:
Just FYI. From the recent typing work in mpi4py (https://github.com/mpi4py/mpi4py/pull/20) I was under the impression that all mypy type annotations have to be done manually, which is a horrifying work...I hope there is a better way to do this!

kmaehashi:
Random notes:
* Cython has an open PR to support generating stubs (but seems stalled: https://github.com/cython/cython/pull/3818)
* `mypy-stubgen` may be used to generate a stub https://github.com/cython/cython/issues/3150
* We have to add `*.pyi` files to sdist/wheel distribution, once we are ready to expose it to users

asi1024:
Rebased!

setup.py
```diff
@@ -42,8 +42,10 @@
     'stylecheck': [
         'autopep8==1.5.5',
         'flake8==3.8.4',
+        'numpy==1.20.*',
```
asi1024, line 45: `mypy` seems to require also type hints of `pytest`. May I split these requirements to `typecheck`?
```py
'typecheck': [
    'mypy==0.812',
    'numpy==1.20.*'
    'pytest>=6.2',
]
```

asi1024: Had an offline discussion with @kmaehashi, and decided to install them at `pretest.yml` instead.

kmaehashi:
> Cython has an open PR to support generating stubs (but seems stalled: Add type stub generation to cython cython/cython#3818)

Cython types are not fully compatible with Python types, so by nature Cython cannot generate a "final" `pyi` file, although stub can be generated. We have to consider the alternate way.

First of all, I'd like to avoid writing `pyi` files by hand, if possible. Maintaining 3 separate files for each module (`pyx` for implementations, `pxd` for C APIs (`cdef`/`cpdef`), `pyi` for Python APIs (`def`/`cpdef`)) is terrible, especially considering our resource.

From the experience on my small project (adding type annotation to my private Cython-based project: https://github.com/kmaehashi/pypict/pull/28), I think adding type annotations to docstring and using `mypy-stubgen` to *directly* generate a "final" `pyi` file is the most reasonable idea (although I also agree that this is not the "best" solution.) This approach is similar to what mpi4py is using (https://github.com/mpi4py/mpi4py/pull/20). The difference is:

* `mpi4py` patches Cython to automatically generate signature in docstring (https://github.com/mpi4py/mpi4py/blob/8275070dc9bc2d2221012863ecba459d65742b17/conf/cyautodoc.py#L244-L246)
* `mpi4py` uses forked version of `mypy-stubgen` (https://github.com/mpi4py/mpi4py/blob/8275070dc9bc2d2221012863ecba459d65742b17/conf/mpistubgen.py)

Pros:
* Both Cython/Python signatures are defined in one place, so it is easy to maintain and review.
  Example: https://github.com/kmaehashi/pypict/blob/44c7be54d0889c5162d6cb63807dd8a8c854ef8c/pypict/capi.pyx#L182-L183
* Signatures embedded to docstrings can be used as a documentation on Sphinx's autodoc.
  (Although this is no longer a merit once https://github.com/sphinx-doc/sphinx/pull/4824 got merged.)

Cons:
* We rely on `mypy-stubgen` to generate a final `pyi` files.
  However we can fork `mypy-stubgen` when needed.
* Internal methods added by Cython also appear in the output of `mypy-stubgen` so we have to remove them manually.
  We have to hack the generated pyi files like this: https://github.com/kmaehashi/pypict/blob/master/stubgen.sh

.github/workflows/pretest.yml
```diff
@@ -50,6 +50,11 @@ jobs:
         autopep8 -r . --diff | tee check_autopep8
         test ! -s check_autopep8
 
+    - name: Static Type Checking (mypy)
+      run: |
+        pip install numpy==1.21.* mypy==0.790
```
kmaehashi, line 55: * Are there any reasons to use old versions of numpy/mypy here?
* How about adding `mypy` to `setup.py` (`test`) so that contributors can use the same mypy version as CI?

.github/workflows/pretest.yml
```diff
@@ -50,6 +50,11 @@ jobs:
...
+        mypy cupy
```
kmaehashi, line 56: I guess cupyx and cupy_backends needs to be specified?

setup.cfg
```diff
@@ -37,3 +37,7 @@ xfail_strict=true
 
 [metadata]
 license_files = docs/source/license.rst
+
+[mypy]
+ignore_missing_imports = True
+disable_error_code = attr-defined
```
kmaehashi, line 43: Note for later reference: these are needed as Cython part is not typed at all and Python code calling APIs written in Cython raises an error.

kmaehashi (commented):
Thanks, LGTM except for few nits!

kmaehashi (approved):

PR #4963, opened 2021-03-24
[DO NOT MERGE] Monitor compile cache utilization

emcastillo:
Check if cache is being utilized in jenkins and windows CI

### DIFF
--- a/cupy/_execution_stats.py
+++ b/cupy/_execution_stats.py
@@ -1,0 +1,26 @@
+import atexit
+# import os
+
+
+class RuntimeStatistics:
+
+    def __init__(self):
+        self.compile_cache_hits = 0
+        self.compile_cache_miss = 0
+
+    def show_statistics(self):
+        print('Runtime Statistics')
+        print('==================')
+        print(f'Compile Cache hits {self.compile_cache_hits}')
+        print(f'Compile Cache misses {self.compile_cache_miss}')
+        cc_hr = self.compile_cache_hits / (
+            self.compile_cache_hits + self.compile_cache_miss
+        )
+        print(f'Compile Cache hit ratio {cc_hr:.2f}')
+
+
+runtime_statistics = RuntimeStatistics()
+
+
+# if os.environ.get('CUPY_SHOW_CACHE_STATISTICS', False):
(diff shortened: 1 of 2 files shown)


emcastillo:
Jenkins, test this please

leofang:
1. The cache hit in Windows is surprisingly low...I wonder why? Could it be the hash on Windows is not deterministic?
2. Jenkins reported failure, but the GitHub UI does not show a link. @kmaehashi maybe this also relates to the GitHub API fix?

emcastillo:
Cache stats for windows,

```
Runtime Statistics
==================
Compile Cache hits 5002
Compile Cache misses 34942
Compile Cache hit ratio 0.13
```
I also suspect that the hashing in windows has something to do with this, or that we are unpacking the saved cache in the wrong folder.

kmaehashi:
I checked a cache key. https://github.com/cupy/cupy/blob/v9.0.0b3/cupy/cuda/compiler.py#L479 and noticed that the key includes the full path to the CuPy's source tree. 


```
('70', ('-IC:\\Users\\chainer\\cupy\\cupy\\core\\include', '-IC:\\Users\\chainer\\cupy\\cupy\\core\\include\\cupy\\_cuda\\cuda-10.0', '-IC:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.0\\include', '-ftz=true'), (10, 0), 'nvrtc') b'//\n// Generated by NVIDIA NVVM Compiler\n//\n// Compiler Build ID: CL-24817639\n// Cuda compilation tools, release 10.0, V10.0.130\n// Based on LLVM 3.4svn\n//\n\n.version 6.3\n.target sm_70\n.address_size 64\n\n\n\n' #include <cupy/complex.cuh>
```

The problem is that FlexCI checkouts the git repo to a different path for every launch because it includes a execution ID, e.g., `C:\Windows\TEMP\flexci\run-00145123\work\src`.

kmaehashi:
I think we can workaround this by pip installing it (ie avoid using `pip install -e`)

kmaehashi:
Windows kernel cache issue is resolved via #4970. This helped a lot diagnosing the issue, thanks!

PR #4965, opened 2021-03-24
Fix `cutensor` import in the test

leofang:
Follow-up of #4920. This is needed to enable the check `cupy.cuda.cutensor.available`. Not sure how I missed this...

ref: https://github.com/cupy/cupy/pull/4920#discussion_r597873769


kmaehashi (approved):

kmaehashi:
This will fail if only this test file is run, otherwise the test will pass because some other test (within the same process) already imports `import cupy.cuda.cutensor`.

kmaehashi:
pfnCI, test this please.

PR #4969, opened 2021-03-24
Make install_tests runnable without depending on current path

kmaehashi:
Currently install_tests expects that the current directory is at `cupy/` source root.
Due to this limitation we have to run tests separately, in two step (install_tests and other tests).

https://github.com/chainer/chainer-test/blob/master/test_cupy.sh#L28-L34

### DIFF
--- a/tests/install_tests/__init__.py
+++ b/tests/install_tests/__init__.py
@@ -1,0 +1,14 @@
+import importlib
+import os
+import sys
+
+
+def _from_install_import(name):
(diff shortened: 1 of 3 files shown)


kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

asi1024:
LGTM.

asi1024 (approved):

Issue #3222, opened 2020-03-24
Update CUDA search path to pick up `cudatoolkit` in Conda installs


jakirkham:
Based on [discussion starting here]( https://github.com/AnacondaRecipes/cudatoolkit-feedstock/pull/5#issuecomment-603415108 ), it appears `cudatoolkit` is not getting picked up. This could easily be fixed by adding `sys.prefix` to the CUDA search paths.

cc @leofang @gmarkall

jakirkham:
Copying Leo's list from the linked issue, it sounds like this is the search order:

1. `CUDA_PATH` (@jakirkham, is this set by `cudatoolkit`??)
2. the parent directory of `nvcc` (on CF this does not come with `cudatoolkit`)
3. `/usr/local/cuda` (not applicable to CF)

Would propose checking `sys.prefix` between 1 and 2. This means users can still override with `CUDA_PATH`. Though if users do not override, we check `sys.prefix` before checking `nvcc`'s location or `/usr/local/cuda`. Using the latter 2 as fallbacks.

Thoughts? 🙂

leofang:
My feeling is we could do what you suggested, but we still need to fix cupy-feedstock anyway, as `RawModule` and `RawKernel` can use `nvcc` as backend, but `nvcc` by default does not come with `cudatoolkit`. However, if we fix that, the current `get_cuda_path()` would just work (at least I hope).

leofang:
For discussions on potential changes in cupy-feedstock, I'm redirecting to conda-forge/cupy-feedstock#46.

leofang:
> However, if we fix that, the current get_cuda_path() would just work (at least I hope).

According to https://github.com/conda-forge/cupy-feedstock/issues/46#issuecomment-603611591, this issue is not fixed by making `nvcc` as a run dependency in CuPy's recipe, since Conda-forge does not redistribute the CUDA headers. So, John's suggestion on `sys.prefix` seems valid.

leofang:
@jakirkham Is there a safe way to detect if we're using conda's Python? Check `$CONDA_PREFIX`?

leofang:
NVM, @jakirkham. Ignore my `$CONDA_PREFIX` question above. I thought about it very carefully, and also inspected how Numba picks up the CUDA libs. My conclusion is that the present issue is not solvable. Please correct me if I am wrong.

First, let me recap the situation with a conda-installed CuPy:
1. All Python modules are linked to conda's `cudatoolkit`
[...]

PR #4974, opened 2021-03-25
Fix Windows dll loading for Conda

leofang:
See https://github.com/conda-forge/cupy-feedstock/pull/113#issuecomment-807010102. Without this patch, the dll loading errors at this line:
https://github.com/cupy/cupy/blob/30e004549b1862f3360b3b5922566bd6b9805bb2/cupy/_environment.py#L226
The reason is Anaconda/Conda-Forge put all dlls under the Conda env root (`$CONDA_PREFIX`), which is then assigned to `$CUDA_PATH`, so the expected dlls are just in the root dir. Also, on Windows there is no `$CONDA_PREFIX\bin` directory.

### DIFF
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -199,10 +199,15 @@ def _setup_win32_dll_directory():
     # Setup DLL directory to load CUDA Toolkit libs and shared libraries
     # added during the build process.
     if sys.platform.startswith('win32'):
+        is_conda = ((os.environ.get('CONDA_PREFIX') is not None)
+                    or (os.environ.get('CONDA_BUILD_STATE') is not None))
         # Path to the CUDA Toolkit binaries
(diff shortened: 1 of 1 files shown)


leofang:
Looks like there is `$CONDA_PREFIX\Library\bin`. Will check.

leofang:
Resolution: We agreed in https://github.com/conda-forge/cupy-feedstock/pull/114/files#r602357673 to set `CUDA_PATH` to `$CONDA_PREFIX`, so this PR is needed and should be backported. However, we should consider possible improvements that could avoid setting this, discussed in #3222.

kmaehashi:
pfnCI, test this please.

kmaehashi:
LGTM! Test failures are not related to this PR.

PR #4995, opened 2021-03-29
Fix cuSPARSELt example not to use internal function

asi1024:
Fix `examples/cusparselt/matmul.py` not to use internal function `_dtype.to_cuda_dtype`. (cc: @anaruse)
Related to #4883.

### DIFF
--- a/examples/cusparselt/matmul.py
+++ b/examples/cusparselt/matmul.py
@@ -6,8 +6,8 @@
 import cupy
 import numpy
 
+from cupy.cuda import runtime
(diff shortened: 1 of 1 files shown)


examples/cusparselt/matmul.py
```diff
@@ -6,8 +6,8 @@
 import cupy
 import numpy
 
+from cupy_backends.cuda.api import runtime
```
leofang, line 9: Maybe `from cupy.cuda import runtime` is better? I don't think we documented `cupy_backends`?

asi1024: Fixed!

asi1024:
Backport this PR after #4994 is merged.

PR #5004, opened 2021-03-30
Use async memcpy in `ndarray.copy`

### LINKED ISSUE #4988: asarray Ignores the Current Stream
Right now, when copying from one GPU to another, cupy.asarray (and ndarray.copy) execute on the default stream even if some other stream is currently in use. This is because cudaMemcpy always runs on the default stream. This can cause synchronization issues if the next operation on the current stream executes on the copied data. This shows up in a larger app I'm helping with, but we haven't yet found a small example that shows the bad synchronization. I suspect that the fix is to change https://github.com/cupy/cupy/blob/23f0b72e6b9e8c8daf34548ac6649dd164da1c45/cupy/_core/core.pyx#L518 to use copy_from_device_async, but I'm not sure if there would be other undesirable consequences from that approach. Now that we've identified the problem, we can work around this by calling copy_from_device_async directly, but this is a major pitfall for multi-device users. What would be a sane way to fix this? A big thanks to @nicelhc13 for pinning down what was happening on our end.

takagi:
Fix #4988.

This PR fixes `ndarray.copy` to use async memcpy so that it respects the current stream.

I'll try to supply an appropriate testing code.

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -515,9 +515,11 @@ cdef class ndarray:
             # it is recommended to set the current device to the device
             # where the src data is physically located.
             runtime.setDevice(self.data.device_id)
-        newarray.data.copy_from_device(x.data, x.nbytes)
-        if runtime._is_hip_environment:
-            runtime.setDevice(dev_id)
+        try:
+            newarray.data.copy_from_device_async(x.data, x.nbytes)
+        finally:
+            if runtime._is_hip_environment:
+                runtime.setDevice(dev_id)
         return newarray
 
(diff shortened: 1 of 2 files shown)


jakirkham:
cc @pentschev @leofang (for awareness)

takagi:
This PR is ready for review. Would you take a look?

kmaehashi:
pfnCI, test this please.

leofang:
The test is neat!

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #5013, opened 2021-04-01
Give gufunc a name

leofang:
Currently if we throw `_GUFunc` to `cupyx.time.repeat` we get this error:
```
AttributeError: '_GUFunc' object has no attribute '__name__'
```
So this PR fixes it. Though, I am not sure if assigning the name of the wrapped function to the `_GUFunc` object is a good idea. Suggestions welcome!

### DIFF
--- a/cupy/_core/_gufuncs.py
+++ b/cupy/_core/_gufuncs.py
@@ -177,6 +177,7 @@ class _GUFunc:
         self._func = func
         self._signature = signature
         self._default_casting = 'same_kind'
+        self.__name__ = func.__name__
 
         # The following are attributes to avoid applying certain steps
         # when wrapping cupy functions that do some of the gufunc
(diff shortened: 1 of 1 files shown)


cupy/_core/_gufuncs.py
```diff
@@ -177,6 +177,7 @@ def __init__(self, func, signature, **kwargs):
         self._func = func
         self._signature = signature
         self._default_casting = 'same_kind'
+        self.__name__ = func.__name__
```
emcastillo, line 180: How about
```
self.__name__ = kwargs.get('name', func.__name__)
```
😄

leofang: Not bad!

emcastillo:
Jenkins, test this please

leofang:
Jenkins, test this please

emcastillo (approved):

Issue #4902, opened 2021-03-18
Documentation Issues


kmaehashi:
- [ ] source link dead: https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.using_allocator.html
- [ ] some functions are undocumented (e.g. `cupy.conjugate`): https://docs.cupy.dev/en/stable/reference/comparison.html

leofang:
It'd be nice to also document `cupyx.time.repeat` and mention it in, say, the tutorial, for how to perform a fair timing.

leofang:
related: #3692

leofang:
I don't remember where we discussed this (sorry!), but we should also document common ROCm/HIP pitfalls.

leofang:
@kmaehashi I'd like to have a new page added to "Tutorial", placed between "Basics of CuPy" and "User-Defined Kernels". I am thinking to call it "Advanced Tricks of Using CuPy", in which we cover (a) a number of topics that advanced users would appreciate and (b) a number of items in which we invested efforts and would like users to try and give us feedbacks. My current list is:
- [ ] How to use streams and events (we can mention PTDS here)
- [ ] How to measure GPU time correctly (Ans: use cupyx.time.repeat)
- [ ] Advanced FFT usage (cuFFT plan as context manager, multi-GPU FFT, callbacks, etc)
- [ ] Link to "Interoperability"
- [ ] ...what else? Comments and suggestions are welcome!

I'd also like to add a few more examples for `RawModule` (in "User-Defined Kernels"), mainly to show what can be done with the new options:
- [ ] `translate_cucomplex`
- [ ] `name_expressions`
- [ ] `jitify`

I am hoping to get all these done in time before the code-freeze for v9.0.0, but we'll see....

kmaehashi:
Thanks Leo! These docs would definitely help users 😄 
I feel however these topics are more than "Tutorial" things. (that's why I reverted this: #4857)
I'm thinking it's better to rename "Tutorial" to "User Guide" following NumPy's docs: https://numpy.org/devdocs/user/index.html
Then we can move topics like Interoperability, Memory Management, API Compatibility Policy, Difference between CuPy and NumPy, License, etc. to there, instead of wildcard "API Reference" section.

Some other topics I would cover is #1566.

leofang:
> I'm thinking it's better to rename "Tutorial" to "User Guide" following NumPy's docs: https://numpy.org/devdocs/user/index.html
[...]

PR #5015, opened 2021-04-02
Document `cupyx.time.repeat`

leofang:
Address https://github.com/cupy/cupy/issues/4902#issuecomment-801662941. I think it's better to get this API documented sooner while we plan to expand on the tutorial. It is a very handy API that we keep referring users in the issue tracker to use, so I say we just get it on the docs to have something to refer to. As such, we should backport this PR to v9.

### DIFF
--- a/cupyx/time.py
+++ b/cupyx/time.py
@@ -70,7 +70,7 @@ def repeat(
     GPU time is properly recorded by synchronizing internal streams. As a
     result, to time a multi-GPU function all participating devices must be
     passed as the ``devices`` argument so that this helper knows which devices
-    to record. A simple example is given as follows::
+    to record. A simple example is given as follows:
 
     .. code-block:: py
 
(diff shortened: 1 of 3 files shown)


takagi:
Thanks! It was originally introduced as experimental but is pretty helpful and now often used.

leofang:
Right, so I do put a warning saying it is still experimental. But we should get it documented anyway, like a few other experimental APIs 🙂

UmashankarTriforce:
Thank you for this addition! The timing is impeccable :D 
I was just looking for a documentation source on `cupyx.time.repeat` to understand the function better!

leofang:
Jenkins, test this please

asi1024 (approved):

asi1024:
LGTM! Thanks!

leofang:
Thanks!

PR #5016, opened 2021-04-02
Document `cupy.cuda.runtime.getDeviceProperties`

leofang:


leofang:
I checked that the only undocumented runtime APIs are for creating/destroying texture and surface objects, but it's ok because we don't want users to access them directly anyway; those should be managed by classes in `cupy.cuda.texture`.

leofang:
Jenkins, test this please

asi1024 (approved):

asi1024:
LGTM!

PR #5017, opened 2021-04-02
CuPy JIT: Print kernel code

asi1024:
Part of #4290.

### DIFF
--- a/cupyx/jit/_interface.py
+++ b/cupyx/jit/_interface.py
@@ -1,4 +1,5 @@
 import collections
+import warnings
 
 import numpy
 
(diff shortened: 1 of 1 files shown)


leofang:
Thanks, @asi1024!

cupyx/jit/_interface.py
```diff
@@ -81,6 +84,10 @@ def __getitem__(self, grid_and_block):
             block = (block, 1, 1)
         return lambda *args, **kwargs: self(grid, block, args, **kwargs)
 
+    @property
+    def codes(self):
```
kmaehashi, line 88: * I prefer returning it as dict (like `self._cache`) instead of list.
* Let's make it explicit that this is an interface for debugging purposes. How about naming it `cached_codes` instead?

asi1024: Fixed!

leofang (approved):

emcastillo (approved):

emcastillo:
Jenkins, test this please

emcastillo:
Failure unrelated

PR #5019, opened 2021-04-02
More documentation on the supported backends

### LINKED ISSUE #2085: `max` and `min` for float32 cupy array may be slow
  - CuPy version = commit 2146ce2949c339126d2cdd8feb0af4da8313de5b
  - OS/Platform = Ubuntu 16.04/ V100
  - CUDA version = 10.0
* Code to reproduce
```python
import cupy as cp
from contextlib import contextmanager

@contextmanager
def sync_time(name):
    start = cp.cuda.Event()
    end = cp.cuda.Event()
    start.record()
    start.synchronize()
    yield
    end.record()
    end.synchronize()
    t = cp.cuda.get_elapsed_time(start,end)
    print("{} : {} ms".format(name,t))

x = cp.random.normal(size=((400, 32, 28, 28))).astype(cp.float32)

with sync_time("cupy"):
    for i in range(1000):
        x.max()

x = cp.asnumpy(x) #move to cpu

with sync_time("numpy"):
    for i in range(1000):
        x.max()
```
*  Results
```
cupy : 8457.2451171875 ms
numpy : 3005.8154296875 ms
```
In this case, Cupy is slower then Numpy.

### LINKED ISSUE #2750: `max()` and `min()` handle complex NaN differently from NumPy
Reproducer:
```python
import numpy as np
import cupy as cp


cp.cuda.cub_enabled = False  # fall back to cupy's reduction routines
a = [0.5 + 3.7j, np.complex(0.7, np.nan), np.complex(np.nan, -3.9),  np.complex(np.nan, np.nan)]

a_np = np.asarray(a)
print(a_np.max(), a_np.min())

a_cp = cp.asarray(a_np)
print(a_cp.max(), a_cp.min())
```
Outputs from NumPy and CuPy do not match:
```shell
(0.7+nanj) (0.7+nanj)
(nan-3.9j) (nan-3.9j)
```
NumPy seems to propagate the first encountered NaN, but I can't verified this rigorously. BTW, CUB handles this correctly (set `cp.cuda.cub_enabled = True`).

I can't figure out what's wrong by inspecting the `_amax` and `_amin` implementations in `cupy/core/_routines_statistics.pyx`. Seems to be an issue of how pairwise operations are actually done in the reduction kernel.

### LINKED ISSUE #3692: Document which routines support cub
Right now, we support CUB based reduction, but some routines such as max, min, sum, or scan use some CUB special functions.

We should add a note to the docstring of such functions stating that cub acceleration is available.

leofang:
Close #2750.
Close #3692.
Close #2085.

### DIFF
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -29,7 +29,9 @@ CuPy also includes the following features for performance:
 
 - User-defined elementwise CUDA kernels
 - User-defined reduction CUDA kernels
+- Just-in-time compiler converting Python functions to CUDA kernels
 - Fusing CUDA kernels to optimize user-defined calculation
+- CUB/cuTENSOR backends for reduction and other routines
 - Customizable memory allocator and memory pool
 - `cuDNN <https://developer.nvidia.com/cudnn>`_ utilities
 
--- a/docs/source/reference/difference.rst
+++ b/docs/source/reference/difference.rst
@@ -168,6 +168,4 @@ counterparts:
   (nan-3.9j) (nan-3.9j)
 
 The reason is that internally the reduction is performed in a strided fashion, thus it does not ensure a proper
-comparison order and cannot follow NumPy's rule to always propagate the first-encountered NaN. If full compliance
-for this behavior is needed, we recommend using the CUB reduction by setting ``CUPY_ACCELERATORS=cub``, see
-:ref:`here<accelerators>` for a brief introduction to the CUB backend.
+comparison order and cannot follow NumPy's rule to always propagate the first-encountered NaN.
--- a/docs/source/reference/difference.rst
+++ b/docs/source/reference/difference.rst
@@ -118,7 +118,7 @@ This may affect the dtype of its output, depending on the values of the "scalar"
 Data types
 ----------
 
-Data type of CuPy arrays cannot be non-numeric like strings and objects.
+Data type of CuPy arrays cannot be non-numeric like strings or objects.
 See :ref:`overview` for details.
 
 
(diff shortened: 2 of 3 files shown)


docs/source/overview.rst
```diff
@@ -29,7 +29,9 @@ CuPy also includes the following features for performance:
 
 - User-defined elementwise CUDA kernels
 - User-defined reduction CUDA kernels
+- Just-in-time compiler converting Python functions to CUDA kernels
```
leofang, line 32: cc: @asi1024 I think `@cupyx.jit` should be mentioned in the Overview page, but not sure if this is the best place.

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

docs/source/overview.rst
asi1024: I think it is appropriate place! Thanks!

docs/source/reference/difference.rst
```diff
@@ -149,3 +149,25 @@ full numpy arrays.
...
+  (0.7+nanj) (0.7+nanj)
+  >>>
+  >>> a_cp = cp.asarray(a_np)
+  >>> print(a_cp.max(), a_cp.min())
+  (nan-3.9j) (nan-3.9j)
+
+The reason is that internally the reduction is performed in a strided fashion, thus it does not ensure a proper
+comparison order and cannot follow NumPy's rule to always propagate the first-encountered NaN. If full compliance
+for this behavior is needed, we recommend using the CUB reduction by setting ``CUPY_ACCELERATORS=cub``, see
+:ref:`here<accelerators>` for a brief introduction to the CUB backend.
```
asi1024, line 173: Could you add a comment that the accelerators may work only for the input of good contiguity?

kmaehashi: I think it's better to avoid mentioning CUB in this context if it only works for contiguous arrays.

leofang: OK you have a point. Removed in 95a5957.

asi1024:
Jenkins, test this please.
Jenkins, test this please.

asi1024 (approved):

asi1024:
LGTM!

PR #5025, opened 2021-04-05
Document `CFunctionAllocator` and `ManagedMemory`

leofang:
I think it'd be good to expand the memory management documentation, but before that let's make them appear in the index first.

### DIFF
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -108,6 +108,7 @@ from cupy.cuda.memory import MemoryAsync  # NOQA
 from cupy.cuda.memory import MemoryPointer  # NOQA
 from cupy.cuda.memory import MemoryPool  # NOQA
 from cupy.cuda.memory import PythonFunctionAllocator  # NOQA
+from cupy.cuda.memory import CFunctionAllocator  # NOQA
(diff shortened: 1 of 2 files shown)


kmaehashi:
Do you come up with any usecase for `CFunctionAllocator`? It was originally implemented for ChainerX but I guess no one is using it now.

leofang:
For now it's mostly for completeness:
allocate memory | from CuPy via C | from CuPy via Python | from external via C | from external via Python
---|---|---|---|---
method | N/A (see #3212) | `cupy.cuda.alloc` | `cupy.cuda.CFunctionAllocator` | `cupy.cuda.PythonFunctionAllocator`

In the near future there might be C libraries wanting to share the same memory pool with CuPy (ex: nvJPEG2000, in which there is a [device allocator interface](https://docs.nvidia.com/cuda/nvjpeg2000/userguide.html#device-allocator-interface)). If "from CuPy via C" is not available, at least they can do it in the reverse direction using `CFunctionAllocator`.

kmaehashi:
Thanks for the clarification!

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

PR #5032, opened 2021-04-05
Add warnings for duplicate installation

### LINKED ISSUE #3552: Detect wrong installation (duplicate wheels) and warn
Refs: #3541

https://github.com/chainer/chainer/blob/e8283065bc706447bf4f721122fc3e49f0a06ea1/chainer/_environment_check.py#L44

kmaehashi:
Closes #3552

### DIFF
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -398,7 +398,10 @@ def _detect_duplicate_installation():
         'cupy-cuda110',
         'cupy-cuda111',
         'cupy-cuda112',
+        'cupy-cuda113',
         'cupy-rocm-4-0',
+        'cupy-rocm-4-1',
+        'cupy-rocm-4-2',
     ])
     installed = set(
         [p.project_name for p in pkg_resources.working_set])
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -404,13 +404,14 @@ def _detect_duplicate_installation():
         [p.project_name for p in pkg_resources.working_set])
     cupy_installed = known & installed
     if 1 < len(cupy_installed):
-        warnings.warn('''
+        cupy_packages_list = ', '.join(sorted(cupy_installed))
+        warnings.warn(f'''
 --------------------------------------------------------------------------------
 
   CuPy may not function correctly because multiple CuPy packages are installed
   in your environment:
 
-    {cupy_installed}
+    {cupy_packages_list}
 
   Follow these steps to resolve this issue:
 
@@ -419,10 +420,14 @@ def _detect_duplicate_installation():
 
          $ pip uninstall <package_name>
 
+      If you previously installed CuPy via conda, also run the following:
+
+         $ conda uninstall cupy
+
     2. Install the appropriate CuPy package.
        Refer to the Installation Guide for detailed instructions.
 
          https://docs.cupy.dev/en/stable/install.html
 
 --------------------------------------------------------------------------------
-'''.format(cupy_installed=', '.join(sorted(cupy_installed))))
+''')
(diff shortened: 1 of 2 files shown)


cupy/_environment.py
```diff
@@ -369,3 +371,47 @@ def _preload_warning(lib, exc):
...
+  in your environment:
+
+    {cupy_installed}
+
+  Follow these steps to resolve this issue:
+
+    1. For all packages listed above, run the following command to remove all
+       existing CuPy installations:
+
+         $ pip uninstall <package_name>
```
leofang, line 409:
```suggestion
    1. For all packages listed above, run either of the following commands to
       remove all existing CuPy installations:

         $ pip uninstall <package_name>  # for package installed via pip
         $ conda uninstall cupy          # for package installed via conda
         
       It is important to first make sure the source of previous installations.
```

leofang (commented):
I had experience that there's a pip-installed cupy and a conda-installed cupy in my env, and it's important to remove them using the correct package manager.

cupy/_environment.py
```diff
@@ -10,6 +10,8 @@
 import sys
 import warnings
 
+import pkg_resources
```
toslunar, line 13: Do you know whether the import is fast enough in the Python versions CuPy supports?

https://github.com/cupy/cupy/blob/9704573e333760ec80cb7353eef66ed94bf4afab/cupy/testing/_helper.py#L1252-L1254

cupy/_environment.py
```diff
@@ -369,3 +371,47 @@ def _preload_warning(lib, exc):
...
+        [p.project_name for p in pkg_resources.working_set])
+    cupy_installed = known & installed
+    if 1 < len(cupy_installed):
+        warnings.warn('''
+--------------------------------------------------------------------------------
+
+  CuPy may not function correctly because multiple CuPy packages are installed
+  in your environment:
+
+    {cupy_installed}
```
toslunar, line 402: Could you use an f-string instead of `format`?

toslunar (changes requested):

cupy/_environment.py
kmaehashi: Good point! This was added from the user on Windows. https://github.com/cupy/cupy/pull/2759

`pkg_resources` is part of `setuptools`, so it can be depending on the `setuptools` version too.
I'm not 100% sure about this import is performant in all environment, so maybe it's better to avoid using `pkg_resources`. I'm thinking of `importlib.metadata` instead. Unfortunately, it is only supported in Python 3.8+.

cupy/_environment.py
kmaehashi: Ok, I forgot that rule I just added... 😅

cupy/_environment.py
kmaehashi: Thanks! Added conda instructions.

cupy/_environment.py
toslunar:
```python
installed = {d.metadata['Name'] for d in importlib.metadata.distributions()}
```
or
```python
cupy_installed = [name for name in known if list(importlib.metadata.distributions(name=name))]
```
if `importlib.metadata` is used.

kmaehashi: I'm thinking of using `importlib.metadata` for Python 3.8+ and do nothing for 3.7 or earlier.
[`importlib-metadata`](https://pypi.org/project/importlib-metadata/) package is available for backport for 3.7 but we don't want to introduce additonal dependency.

kmaehashi:
Now uses `importlib.metadata` instead of `pkg_resources`. Note that this check is nop for Python 3.7 or earlier.

toslunar (approved):
LGTM

toslunar:
Jenkins, test this please.

kmaehashi:
pfnCI, test this please.

PR #5033, opened 2021-04-05
Fix code block in installation guide

leofang:
Fix an oversight of mine. This needs to be backported.


leofang:
Confirmed fixed: https://cupy--5033.org.readthedocs.build/en/5033/install.html#installing-cupy-from-conda-forge

toslunar (approved):

PR #5037, opened 2021-04-06
Add PCI Bus ID to show_config

kmaehashi:
So that users can map CUDA device ID and nvidia-smi device when multiple cards having the same name are installed.


asi1024:
Jenkins, test this please.

asi1024:
CI failure seems unrelated. (https://github.com/cupy/cupy/pull/5061)
LGTM.

asi1024 (approved):

PR #5040, opened 2021-04-06
Add debug print for flaky test

kmaehashi:
To diagnose #4673

Issue #4673, opened 2021-02-16
Flaky test `manipulation_tests.test_basic.TestBasic.test_copyto_multigpu`


emcastillo:
```
Error Message
AssertionError: 
Arrays are not equal

Mismatched elements: 24 / 24 (100%)
Max absolute difference: 23.
Max relative difference: 1.
 x: array([[[0.   , 0.   , 0.   , 1.984],
        [0.   , 0.   , 0.   , 2.   ],
        [0.   , 0.   , 0.   , 2.016]],...
 y: array([[[ 1.,  2.,  3.,  4.],
        [ 5.,  6.,  7.,  8.],
        [ 9., 10., 11., 12.]],...
```

https://jenkins.preferred.jp/job/chainer/job/cupy_pr/3016/TEST=cupy-py3-cub,label=mn1-p100/testReport/junit/tests.cupy_tests.manipulation_tests.test_basic/TestBasic/test_copyto_multigpu/

kmaehashi:
Other cases:
https://jenkins.preferred.jp/job/chainer/job/cupy_pr/TEST=cupy-py3-cub,label=mn1-p100/3005/testReport/
https://jenkins.preferred.jp/job/chainer/job/daily_master_cupy-shuffled/449/ID=14,label=mn1-p100/testReport/junit/tests.cupy_tests.manipulation_tests.test_basic/TestBasic/test_copyto_multigpu/

kmaehashi:
This test failures shows a similar behavior, although this is not a multi-gpu test.
https://jenkins.preferred.jp/job/chainer/job/cupy_pr/3044/TEST=cupy-py3,label=mn1-p100/testReport/junit/tests.cupy_tests.cuda_tests.test_texture/TestCUDAarray_param_25_%7Bdimensions=(67,%200,%200),%20dtype=int8,%20n_channels=1,%20stream=True,%20xp='cupy'%7D/test_array_gen_cpy/

leofang:
Could it be something wrong with the random numbers?

kmaehashi:
As for `test_copyto_multigpu `, I think random numbers are not related, as it just tries to copy `arange` result between devices.

takagi:
I cannot reproduce this behavior on my local 😢

leofang:
Just a thought: fixing #3980 might fix this, as the "current stream" shouldn't be the same for all devices.

leofang:
[...]

PR #5042, opened 2021-04-06
Fix `TestStream` cleanup

emcastillo:
This seems to be why the copy multigpu error is happening in the CI https://github.com/cupy/cupy/issues/4673.

In these tests, ptds is left as the default stream for the rest of the CI execution, and when the array is copied back, `asarray` ignores the ptds set https://github.com/cupy/cupy/pull/5004, and just starts copying to host before the multidevice copy is over.

If we are right, https://github.com/cupy/cupy/pull/5040 will print 2 in the current stream being used once that test fails.

#5004 fixes the root cause of the error, but this PR is needed to set the CI environment correctly.

### DIFF
--- a/tests/cupy_tests/cuda_tests/test_stream.py
+++ b/tests/cupy_tests/cuda_tests/test_stream.py
@@ -16,10 +16,15 @@ class TestStream(unittest.TestCase):
         if cuda.runtime.is_hip and self.stream_name == 'ptds':
             self.skipTest('HIP does not support PTDS')
 
+        self._prev_stream = cuda.get_current_stream()
+
         if self.stream_name == 'null':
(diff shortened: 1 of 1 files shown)


leofang:
@emcastillo If #4337 had been merged this wouldn't have been a bug that keeps haunting us. Just saying 😂

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

kmaehashi:
@cupy/code-owners Failed to backport automatically.

----

```
fPR #5042 is not merged
```

kmaehashi:
Let me test bot mode (https://github.com/cupy/backport/pull/27) of backport tool in this PR 🙇🏼 :

PR #4695, opened 2021-02-20
Update DLPack support

leofang:
~~**DO NOT MERGE** until the upstream is tagged!~~ 

The new v0.4 has the following change:
- Support complex numbers
- Rename the struct `DLContext` to `DLDevice`

Prepare for Array API adoption (https://github.com/data-apis/array-api/pull/106).

PR #5045, opened 2021-04-06
Fix DLPack lanes

leofang:
Follow-up of #4695. This PR contains two fixes:
1.  We need to set `lanes=1` for complex numbers to match the `bits` field (which was set correctly in #4695).
2. We need to reject `lanes>1` following what PyTorch has been doing; the reason our round-trip test passed for the No.1 bug is because we allowed any `lanes` but actually ignored it

related: pytorch/pytorch#55365, https://github.com/dmlc/dlpack/pull/66

cc: @emcastillo @rgommers


emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #5046, opened 2021-04-06
Disable cuFFT plan cache on CUDA 11.1

### LINKED ISSUE #5022: Issues with consecutive 2D FFT runs on the same plan
I have noticed a problem when CuPy 2D FFTs with the same plan are run consecutively. With certain shapes, they give different and incorrect results the second time. This seems to only happen when the longer dimension of the shape is odd, but I have not done an exhaustive scan. NumPy and SciPy FFT functions show the expected result, and the first run of CuPy always agrees with these. The second run is the problem. For now, I was able to work around by making sure the longer dimension was even, but I wanted to post this in case others were having the same issues. 

  - CuPy version: 8.6.0
  - OS/Platform: Windows 10
  - CUDA version: 11.2
* Code to reproduce

import numpy as np
import cupy as cp

shape = (151, 2229)
a = np.random.randn(*shape)

A1 = np.fft.rfft2(a)
A2 = np.fft.rfft2(a)

b = cp.asarray(a)
B1 = cp.fft.rfft2(b)
B2 = cp.fft.rfft2(b)
if cp.any(B1 != B2):
    print(A1[0,0], A2[0,0], B1[0,0], B2[0,0], shape)

Output from the code above: (-84.42714344719724-2.5579538487363607e-13j) (-84.42714344719724-2.5579538487363607e-13j) (-84.42714344719798+0j) (17.20028862520553-3.1086244689504383e-15j) (151, 2229)

leofang:
Close #5022.

This is a band-aid fix to avoid reusing cuFFT plans on CUDA 11.1 that introduces this regression:

> Plans with primes larger than 127 in FFT size decomposition or FFT size being a prime number bigger than 4093 do not perform calculations on second and subsequent cufftExecute* calls. Regression was introduced in cuFFT 11.1

https://docs.nvidia.com/cuda/archive/11.2.0/cuda-toolkit-release-notes/index.html#cufft-resolved-issues

Thanks @michaelmarty for reporting and @mnicely for pointing out the issue!

### DIFF
--- a/cupy/fft/_cache.pyx
+++ b/cupy/fft/_cache.pyx
@@ -1,6 +1,7 @@
 # distutils: language = c++
 
 import gc
+import warnings
 import weakref
 
 from cupy_backends.cuda.api cimport runtime
@@ -263,6 +264,11 @@ cdef class PlanCache:
     # ---------------------- Python methods ---------------------- #
 
     def __init__(self, Py_ssize_t size=16, Py_ssize_t memsize=-1, int dev=-1):
+        if runtime.runtimeGetVersion() == 11010:
+            warnings.warn('cuFFT plan cache is disabled on CUDA 11.1 due to a '
+                          'known bug, so performance may be degraded. The bug '
+                          'is fixed on CUDA 11.2+.')
+            size = 0
         self._validate_size_memsize(size, memsize)
         self._set_size_memsize(size, memsize)
         self._reset()


leofang:
I wonder why the shuffled CI does not catch this...?

leofang:
Jenkins, test this please

leofang:
> I wonder why the shuffled CI does not catch this...?

I think it's because we mainly test small-size arrays.

cupy/fft/_cache.pyx
```diff
@@ -263,6 +264,11 @@ cdef class PlanCache:
     # ---------------------- Python methods ---------------------- #
 
     def __init__(self, Py_ssize_t size=16, Py_ssize_t memsize=-1, int dev=-1):
+        if runtime.runtimeGetVersion() == 11010:
```
emcastillo, line 267: We would like to do a more strict check here, but since this happens with prime sizes, add primality tests here can be overkill.

leofang: That would incur extra overhead and actually requires some knowledge about cufft internal for a proper test, so it's easier to just kill the cache unconditionally.

emcastillo (approved):

PR #5047, opened 2021-04-06
Rename cupyx.scipy.stats.distributions to _distributions

rgommers:
`stats.distributions` is listed as public API by SciPy for historical reasons, however all of its contents are also star-imported into the `stats` namespace - which should be preferred.

`cupyx.scipy.stats` does not yet appear in https://docs.cupy.dev/en/stable/reference/scipy.html and the tests only cover `stats.entropy`, so this move should be safe.


rgommers:
This is an issue in other `cupyx.scipy` submodules as well. I intend to clean up the "private modules missing an underscore" in SciPy at some point; it'd be useful to consider doing that cleanup in CuPy now and marking everything clearly as private that way except for the `scipy.xxx` top-level submodules and the few public more deeply nested modules. The only one more than 2 levels deep that CuPy has at the moment is `sparse.linalg`.

I'd be happy to help with this.

Related: I came across this when investigating https://github.com/scipy/scipy/issues/10204, which looks at putting a dispatching layer into the `scipy.ndimage` API that is CuPy-aware.

grlee77:
Thanks @rgommers. Do you want me to make a corresponding PR to append underscores to `interpolation`, `filters`, etc. within `cupyx.scipy.ndimage`? 

I was looking at scikit-image code just now and found a few places where imports are currently done from scipy.ndimage submodules. (e.g. `from scipy.ndimage.filters import gaussian_filter`) and quite a few uses of morphology functions via, e.g. `ndi.morphology.binary_dilation` rather than just `ndi.binary_dilation`. I can make a PR to scikit-image to remove these unintended usage patterns, but I suspect they may be somewhat common in other 3rd party libraries as well. This is a bit less of an issue for CuPy than SciPy itself since many of these functions will officially appear for the first time in the pending 9.0 release and there will be less downstream code to potentially break.

rgommers:
That would be great, thanks @grlee77 

> I can make a PR to scikit-image to remove these unintended usage patterns, but I suspect they may be somewhat common in other 3rd party libraries as well

Sounds useful as well. I'm indeed not sure how quickly we can do that for `ndimage`, because the line between public and private was historically blurry. We've had a prominent "these are the public submodules" list (http://scipy.github.io/devdocs/reference/index.html#api-definition) for a long time, but as you know no one reads the docs .....

asi1024:
@rgommers Thank you for your PR!
`scipy.stats.distributions` is a private module, and also CuPy add `_` prefix to private modules and routines.

However in our [development policy](https://github.com/cupy/cupy/wiki/Development-policies#numpyscipy-compatibility), we don't apply fixes until the corresponding change is merged in NumPy/SciPy upstream branch. Actually `scipy.stats.distributions` is widely used (https://github.com/search?q=scipy.stats.distributions&type=code) in spite of it is a private module. So we are concerning that users may think it is CuPy's issue of lack of SciPy-compatibility if we apply this change before SciPy does.

rgommers:
> However in our development policy, we don't apply fixes until the corresponding change is merged in NumPy/SciPy upstream branch.

Fair enough, thanks @asi1024. What I will do is hurry up with restructuring some thing in SciPy. Right now, it's not even tested whether something stays in a private submodule, and this is exactly what happened with `stats.distributions.entropy` - it was moved out without tests failing, and I only randomly caught it: https://github.com/scipy/scipy/pull/13787#issuecomment-814268495. So now that SciPy's API mistakes are proliferating, that's extra motivation to do something about it quickly.

I'll get back to you once some changes have landed in SciPy.

kmaehashi:
Following SciPy 1.8, we've done hiding all private APIs (#6252).

rgommers:
Great, thanks @kmaehashi!

PR #5050, opened 2021-04-07
Skip cuTENSOR/cuSPARSELt installation for unsupported CUDA version

asi1024:


leofang:
#5136 may be relevant.

cupy_setup_build.py
```diff
@@ -526,6 +527,14 @@ def preconfigure_modules(compiler, settings):
             cmd = 'nvcc' if not use_hip else 'hipcc'
             errmsg = ['{} command could not be found in PATH.'.format(cmd),
                       'Check your PATH environment variable.']
+        elif ('required_cuda_version' in module
```
kmaehashi, line 530: This needs to be checked first (before `not check_library()` condition) to skip building cuTENSOR/cuSPARSELt support for unsupported CUDA versions.

cupy_setup_build.py
```diff
@@ -526,6 +527,14 @@ def preconfigure_modules(compiler, settings):
...
+              and not module['check_cuda_version'](build.get_cuda_version())):
+            cuda_version = build.get_cuda_version()
+            cuda_major = cuda_version // 1000
+            cuda_major_minor = f'{cuda_major}.{(cuda_version // 10) % 100}'
+            installed = True
+            errmsg = [
+                f'The library is not supported for CUDA {cuda_major_minor}']
```
kmaehashi, line 537:
```suggestion
                f'Not supported in CUDA {cuda_major_minor}']
```

kmaehashi (changes requested):

kmaehashi:
I think this doesn't make much sense after dropping CUDA 10.2... maybe it's ok to close? @asi1024 :bow:

PR #5060, opened 2021-04-08
Improve comments for memory and stream API usage

kmaehashi:
To prevent further error like #5004

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -417,7 +417,7 @@ cdef class MemoryPointer:
 
             This function always uses the legacy default stream and does not
             honor the current stream. Use `copy_from_device_async` instead
-            if you don't understand what you are doing.
+            if you are using streams in your code, or have PTDS enabled.
 
         """
         if size > 0:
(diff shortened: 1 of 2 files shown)


cupy/cuda/memory.pyx
```diff
@@ -413,6 +413,12 @@ cdef class MemoryPointer:
             src (cupy.cuda.MemoryPointer): Source memory pointer.
             size (int): Size of the sequence in bytes.
 
+        .. warning::
+
+            This function always uses the legacy default stream and does not
+            honor the current stream. Use `copy_from_device_async` instead
+            if you don't understand what you are doing.
```
emcastillo, line 420: I think that `copy_from_device_async` requires you to know what you are doing 😃.
Maybe we can change it to something in the lines of
`Use copy_from_device_async instead if you are using streams in your code, or have PTDS enabled`

kmaehashi: Right, updated as suggested! 👍🏼

emcastillo:
Jenkins, test this please

emcastillo (approved):

Issue #4959, opened 2021-03-24
Support ROCm 4.1


kmaehashi:

leofang:
related: https://github.com/cupy/cupy/pull/4961

leofang:
```
#warning The hipFFT version included in rocFFT is deprecated.  Please use hipFFT from the hipFFT package/repository instead.
```

kmaehashi:
ROCm v4.1 support is skipped due to the issues in ROCm (see #5067). ROCm 4.2 support is ongoing: #5213

PR #5067, opened 2021-04-08
Changes to build against future ROCm ( >4.1)

amathews-amd:
Fixes
* Build error 'explicit specialization after instantiation' (https://groups.google.com/g/cupy/c/-L8HPMTy9n8)
* rocFFT -> hipFFT https://github.com/cupy/cupy/issues/4959

### DIFF
--- a/cupy/cuda/cupy_cub.cu
+++ b/cupy/cuda/cupy_cub.cu
@@ -63,8 +63,6 @@ template <> struct NumericTraits<complex<double>> : BaseTraits<FLOATING_POINT, t
 // TODO(leofang): wrap CuPy's thrust namespace with another one (say, cupy::thrust) for safer scope resolution?
 
 namespace std {
-// TODO(amathews-amd): Check if this code can be safetly removed.
-#if HIP_VERSION < 401
 template <>
 class numeric_limits<thrust::complex<float>> {
   public:
@@ -88,7 +86,6 @@ class numeric_limits<thrust::complex<double>> {
         return thrust::complex<double>(-std::numeric_limits<double>::max(), -std::numeric_limits<double>::max());
     }
 };
-#endif
 
 // Copied from https://github.com/ROCmSoftwarePlatform/hipCUB/blob/master-rocm-3.5/hipcub/include/hipcub/backend/rocprim/device/device_reduce.hpp
 // (For some reason the specialization for __half defined in the above file does not work, so we have to go
--- a/cupy/cuda/cupy_cub.cu
+++ b/cupy/cuda/cupy_cub.cu
@@ -63,6 +63,8 @@ template <> struct NumericTraits<complex<double>> : BaseTraits<FLOATING_POINT, t
 // TODO(leofang): wrap CuPy's thrust namespace with another one (say, cupy::thrust) for safer scope resolution?
 
 namespace std {
+// TODO(amathews-amd): Check if this code can be safetly removed.
+#if HIP_VERSION < 401
 template <>
 class numeric_limits<thrust::complex<float>> {
   public:
@@ -86,6 +88,7 @@ class numeric_limits<thrust::complex<double>> {
         return thrust::complex<double>(-std::numeric_limits<double>::max(), -std::numeric_limits<double>::max());
     }
 };
+#endif
 
 // Copied from https://github.com/ROCmSoftwarePlatform/hipCUB/blob/master-rocm-3.5/hipcub/include/hipcub/backend/rocprim/device/device_reduce.hpp
 // (For some reason the specialization for __half defined in the above file does not work, so we have to go
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -116,7 +116,7 @@ if use_hip:
             'amdhip64',  # was hiprtc and hip_hcc before ROCm 3.8.0
             'hipblas',
             'hiprand',
-            'rocfft',
+            'hipfft',
             'roctx64',
             'rocblas',
             'rocsolver',
--- a/install/build.py
+++ b/install/build.py
@@ -147,6 +147,7 @@ def get_compiler_setting(use_hip):
             library_dirs.append(os.path.join(cuda_path, 'lib'))
 
     if rocm_path:
+        include_dirs.append(os.path.join(rocm_path, 'hipfft', 'include'))
         include_dirs.append(os.path.join(rocm_path, 'include'))
         include_dirs.append(os.path.join(rocm_path, 'include', 'hip'))
         include_dirs.append(os.path.join(rocm_path, 'include', 'rocrand'))
--- a/install/build.py
+++ b/install/build.py
@@ -147,7 +147,6 @@ def get_compiler_setting(use_hip):
             library_dirs.append(os.path.join(cuda_path, 'lib'))
 
     if rocm_path:
-        include_dirs.append(os.path.join(rocm_path, 'hipfft', 'include'))
         include_dirs.append(os.path.join(rocm_path, 'include'))
         include_dirs.append(os.path.join(rocm_path, 'include', 'hip'))
         include_dirs.append(os.path.join(rocm_path, 'include', 'rocrand'))


amathews-amd:
@jithunnair-amd @pruthvistony @jeffdaily @sunway513 - please review

cupy_setup_build.py
```diff
@@ -116,7 +116,7 @@
             'amdhip64',  # was hiprtc and hip_hcc before ROCm 3.8.0
             'hipblas',
             'hiprand',
-            'rocfft',
+            'hipfft',
```
leofang, line 119: This line needs to be reverted. `libraries` here indicates the names of the shared libraries that are linked by the builder, so this change is not necessary because there's no `libhipfft.so`, only `librocfft.so`.

install/build.py
```diff
@@ -147,6 +147,7 @@ def get_compiler_setting(use_hip):
             library_dirs.append(os.path.join(cuda_path, 'lib'))
 
     if rocm_path:
+        include_dirs.append(os.path.join(rocm_path, 'hipfft', 'include'))
```
leofang, line 150: This change is not needed because `hipfft.h` can be found in `/opt/rocm/include`, so the line below is enough to make it locatable.

cupy/cuda/cupy_cub.cu
```diff
@@ -63,6 +63,8 @@ template <> struct NumericTraits<complex<double>> : BaseTraits<FLOATING_POINT, t
 // TODO(leofang): wrap CuPy's thrust namespace with another one (say, cupy::thrust) for safer scope resolution?
 
 namespace std {
+// TODO(amathews-amd): Check if this code can be safetly removed.
+#if HIP_VERSION < 401
```
leofang, line 67: Could you install pytest and run this test locally to see it'd fail?
```
pytest tests/cupy_tests/core_tests/test_ndarray_reduction.py -k TestCubReduction
```

leofang (changes requested):
Thanks for looking into this for us @amathews-amd! A few changes are not necessary.

emcastillo:
Thanks for the PR!

cupy_setup_build.py
amathews-amd: See
* https://github.com/ROCmSoftwarePlatform/rocFFT/blob/develop/library/include/hipfft.h#L26
* https://github.com/ROCmSoftwarePlatform/hipFFT#transitioning-from-rocfft 

The change is part of ROCm4.1. 

Is 'HIP_VERSION' available in the build scripts already, so we can key off that ?

install/build.py
amathews-amd: See https://github.com/cupy/cupy/pull/5067#discussion_r610272175 above

cupy/cuda/cupy_cub.cu
amathews-amd: Fails with the change on ROCm4.1 (See below) . Any ideas on how to fix ? 

`
cupy_tests/core_tests/test_ndarray_reduction.py ..F...F...F...F...F...F...F...F................................. 

FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_0_{backend='device', order='C', shape=(10,)}::test_cub_min - AssertionError:
FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_1_{backend='device', order='C', shape=(10, 20)}::test_cub_min - AssertionError:
FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_2_{backend='device', order='C', shape=(10, 20, 30)}::test_cub_min - AssertionError:
FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_3_{backend='device', order='C', shape=(10, 20, 30, 40)}::test_cub_min - AssertionError:
FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_4_{backend='device', order='F', shape=(10,)}::test_cub_min - AssertionError:
FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_5_{backend='device', order='F', shape=(10, 20)}::test_cub_min - AssertionError:
FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_6_{backend='device', order='F', shape=(10, 20, 30)}::test_cub_min - AssertionError:
FAILED cupy_tests/core_tests/test_ndarray_reduction.py::TestCubReduction_param_7_{backend='device', order='F', shape=(10, 20, 30, 40)}::test_cub_min - AssertionError:
_______________________
        numpy.testing.assert_allclose(
            cupy.asnumpy(actual), cupy.asnumpy(desired),
>           rtol=rtol, atol=atol, err_msg=err_msg, verbose=verbose)
E       AssertionError:
E       Not equal to tolerance rtol=1e-05, atol=0
E
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 9.507621
E       Max relative difference: 1.
E        x: array(0.+0.j, dtype=complex64)
E        y: array(3.834415+8.700122j, dtype=complex64)
`

cupy_setup_build.py
emcastillo: did the actual .so file changed names?

leofang: @amathews-amd OK. I checked the new hipfft directory structure, and I think the packaging messed up. 

For all other hip* packages, there is a main header e.g. `/opt/rocm-4.1.0/include/hipblas.h` which symlinks to `/opt/rocm-4.1.0/hipblas/include/hipblas.h`. However, at least for Ubuntu's deb `opt/rocm-4.1.0/include` is empty (no symlink to `opt/rocm-4.1.0/hipfft/include/hipfft.h`). The same applies to shared libraries. This is something ROCm/HIP needs to fix, and let's better not make any exception for hipfft since the old way is still working (just a warning at compile time).

Also, this change would break backward compatibility. We've been able to make things work smoothly since ROCm 3.5+ (one of my development machines is still on that!) and I personally disfavor moving as fast as the upstream. Our expected users aren't moving that fast either. It's arguable that everything on HIP is still experimental, but let's do our best here 🙂

leofang: > Is 'HIP_VERSION' available in the build scripts already, so we can key off that ?

Sorry, forgot to reply on this. Yes, we do set `HIP_VERSION` at the build time, but not here but later, which is why I said it'd break the backward compatibility.

install/build.py
leofang: Please see my reply https://github.com/cupy/cupy/pull/5067/files#r610286064.

cupy/cuda/cupy_cub.cu
leofang: @amathews-amd this is why we needed these lines...😓

This is not a proper fix. It breaks NumPy's [lexical comparison rules for complex numbers](https://numpy.org/doc/stable/reference/generated/numpy.sort.html):
> The sort order for complex numbers is lexicographic. If both the real and imaginary parts are non-nan then the order is determined by the real parts except when they are equal, in which case the order is determined by the imaginary parts.

So you can see that all `max`/`min` tests are failing with complex numbers. (I think in this test we test all dtypes as I figured out a way to make CUB work for everything we do!)

No one is happy with the way NumPy handles the comparison of complex numbers, but the upstream effort to change it is stalled, so unfortunately before it lands in NumPy's main branch we must comply...

The root cause of the build failure is reported in https://github.com/ROCm-Developer-Tools/HIP/issues/2247. Basically the bug is introduced in ROCm 4.1.0 and **its impact is beyond the build error here**. Now hip-clang forbids any possibility of specializing `std::numeric_limits<T>` for non-standard `T`, which IMHO is a *series bug against the C++ standard*. The bug itself should really be fixed asap upstream...

I would suggest you to looking into workarounds for us to 1. build successfully and 2. stay at the same level of the NumPy compliance that we have strived very hard to achieve. I don't think I know enough the internal of rocPRIM so this approach was the only way I found 😅 but it has been working very well since ROCm 3.5.0...

cupy_setup_build.py
amathews-amd: @emcastillo, yes

@leofang 
Right now, things should still work the old; Just print a 'deprecated' message. I will reverse the changes in the build scripts for now.

leofang: > did the actual .so file changed names?

@emcastillo Sorry, for some reason GitHub didn't show your reply promptly. 

Since ROCm 4.1.0 there is a new `opt/rocm-4.1.0/lib/libhipfft.so` symlinking to `opt/rocm-4.1.0/hipfft/lib/libhipfft.so`.

leofang: > > did the actual .so file changed names?
> 
> @emcastillo Sorry, for some reason GitHub didn't show your reply promptly.
> 
> Since ROCm 4.1.0 there is a new `opt/rocm-4.1.0/lib/libhipfft.so` symlinking to `opt/rocm-4.1.0/hipfft/lib/libhipfft.so`.

OK so for this particular line it's ok to change for ROCm 4.1.0.

cupy/cuda/cupy_cub.cu
leofang: > 2. stay at the same level of the NumPy compliance that we have strived very hard to achieve.

It is up to the team to decide if we would allow an exception to be made. For example, we could disable the code path for complex numbers on ROCm 4.1.0 if there is really no other way to work around...

sunway513: >Now hip-clang forbids any possibility of specializing std::numeric_limits<T> for non-standard T, which IMHO is a serious bug violating the C++ standard. The bug itself should really be fixed asap upstream...

cc @yxsamliu, would you help review the reported bug and share you insights if any? Thanks..

cupy_setup_build.py
leofang: Had a quick chat offline with @kmaehashi, let's keep this so name change here as the team is interested in seeing the support for ROCm 4.1.0 to land. Backward compatibility starts with 4.1.0.

But, @amathews-amd it'd be great if you could kindly ask your team to fix the directory structure (https://github.com/cupy/cupy/pull/5067#discussion_r610286064) 🙂

cupy/cuda/cupy_cub.cu
leofang: @sunway513, thanks, but @yxsamliu already did, see https://github.com/ROCm-Developer-Tools/HIP/issues/2247#issuecomment-808507106.

sunway513: @leofang aha, thanks for the heads up and glad it's fully triaged! We'll follow up internally for resolution..

leofang: Thank you @sunway513 (and @yxsamliu for tracking down the problem)! Please keep us posted if you have a decision 🙂

leofang:
To wrap up quickly with the current discussion about hipFFT: 
1. We can keep using the old header `/opt/rocm/include/hipfft.h` (on ROCm 4.1.0, this is installed by `rocfft`, and hopefully in a future version of ROCm this becomes a working symlink to the actual header installed by `hipfft`)
2. For ROCm 4.1.0 onward, we need to link to `/opt/rocm/lib/libhipfft.so`

cupy_setup_build.py
amathews-amd: I discussed the directory structure issue internally with Emmett. 
New issue opened on rocFFT to track this : https://github.com/ROCmSoftwarePlatform/rocFFT/issues/312

leofang: Thanks @amathews-amd!

cupy/cuda/cupy_cub.cu
amathews-amd: Adding comment for note:
Solved in trunk: https://github.com/ROCm-Developer-Tools/HIP/issues/2247#issuecomment-816675887

amathews-amd:
> To wrap up quickly with the current discussion about hipFFT:
> 
> 1. We can keep using the old header `/opt/rocm/include/hipfft.h` (on ROCm 4.1.0, this is installed by `rocfft`, and hopefully in a future version of ROCm this becomes a working symlink to the actual header installed by `hipfft`)
> 2. For ROCm 4.1.0 onward, we need to link to `/opt/rocm/lib/libhipfft.so`

Reverted changes in install/build.py  and cupy/cuda/cupy_cub.cu . Code will not build on ROCm4.1 as-is.

emcastillo:
@amathews-amd how is this going 😀
Can we help you in any way?
Pretty excited to see this working!

amathews-amd:
We are waiting on the numeric_limits fix to land on ROCm.

kmaehashi:
Let me close this as the build now works with ROCm 4.2 (#5209). Thanks again @amathews-amd!

PR #5072, opened 2021-04-09
Support PTDS in CuPy memory pool

### LINKED ISSUE #5069: CuPy memory pool does not support PTDS
Currently, CuPy memory pool uses the stream pointer as a key of an arena dict.
https://github.com/cupy/cupy/blob/v9.0.0rc1/cupy/cuda/memory.pyx#L968-L970
However, when PTDS is enabled, the key will always become `2` despite that different streams are used in each thread, causing invalid reallocation across threads.

Maybe we should add some normalization to the key:
* ~When the current stream ptr is `1` (streamLegacy), use `0` as a key.~
* When the current stream ptr is `2` (streamPerThread), use `-1 * thread_id` as a key.

kmaehashi:
Closes #5069

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -275,7 +275,7 @@ cdef class _Chunk:
         mem (~cupy.cuda.Memory): The device memory buffer.
         offset (int): An offset bytes from the head of the buffer.
         size (int): Chunk size in bytes.
-        stream_ptr (intptr_t): Raw stream handle of cupy.cuda.Stream
+        stream_ident (intptr_t): Value to uniquely identify the stream.
 
     Attributes:
         mem (Memory): The device memory buffer.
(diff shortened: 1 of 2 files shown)


kmaehashi:
pfnCI, test this please.

cupy/cuda/memory.pyx
```diff
@@ -675,6 +675,21 @@ def _set_thread_local_allocator(allocator):
...
+    # stream for each thread. (#5069)
+    if stream_ptr != runtime.streamPerThread:
+        return stream_ptr
+
+    cpdef intptr_t tid
+    try:
+        tid = _thread_local._tid
+    except AttributeError:
+        _thread_local._tid_obj = tid_obj = object()
+        _thread_local._tid = tid = id(tid_obj)
```
leofang, line 689: This is a nice trick! I wonder if it would be faster than the builtin [`threading.get_ident()`](https://docs.python.org/3/library/threading.html#threading.get_ident)?

tests/cupy_tests/cuda_tests/test_memory.py
```diff
@@ -805,24 +805,42 @@ def test_thread_local_valid(self):
...
             t.start()
             t.join()
         assert not self._error
+        return main_ptr, self._ptr
+
+    def test_reuse_between_thread(self):
+        main_ptr, sub_ptr = self._reuse_between_thread(cupy.cuda.Stream.null)
+        assert main_ptr == sub_ptr
+
+    def test_reuse_between_thread_ptds(self):
```
leofang, line 841:
```suggestion
    @pytest.mark.skipif(cupy.cuda.runtime.is_hip, reason='No PTDS on HIP')
    def test_reuse_between_thread_ptds(self):
```

leofang (changes requested):
Shall we add two more `test_reuse_between_thread` tests? One on the same custom stream, and another on different streams (other than null/ptds)?

cupy/cuda/memory.pyx
kmaehashi: Yes, it's faster than `threading.get_ident()` as far as I tested. Plus, [`id`](https://docs.python.org/3/library/functions.html#id) is guaranteed to be a positive value (on CPython) whereas `get_ident` only says `nonzero integer`.

kmaehashi:
Thanks for the quick review! Added more stream tests as suggested.

leofang (approved):
LGTM!

kmaehashi:
pfnCI, test this please.

cupy/cuda/memory.pyx
```diff
@@ -675,6 +675,21 @@ def _set_thread_local_allocator(allocator):
...
+    return -tid
```
emcastillo, line 690: why is the reason because this is negative?
to differentiate pointers from ids?

leofang: Yeah, to avoid conflicts with stream pointers I think.

kmaehashi: Yes.

emcastillo (approved):

PR #5077, opened 2021-04-11
CuPy JIT: Fix range type

asi1024:
Fix the iteration dtype generated from `range(start, stop, step)`.

Current master:
- `start`, `stop` and `step` must have the same integer dtype (raises `TypeError` otherwise).

This PR
- `start`, `stop` and `step` can be different integer dtype.
- iteration dtype is inferred from `stop`.


emcastillo:
Jenkins, test this please

emcastillo (approved):

Issue #5006, opened 2021-03-30
cupy.cuda.Device()._device_stack is not thread-local


insertinterestingnamehere:
We noticed mysterious device switching behavior in a multithreaded app we're developing and have traced it back to the device stack used to make cupy.cuda.Device a context manager not being thread local. The net effect of this is that a single cupy.cuda.Device object can't safely be used as a context manager in different threads. We can work around this downstream, but this seems like a potential pitfall in the multithreaded use case.

insertinterestingnamehere:
I'm not sure I fully follow your point about matching the behavior of cudaSetDevice. Can you elaborate? The problem I'm seeing is that the context manager API built on top of cudaSetDevice breaks when multiple threads use the same Device object simultaneously.

More concretely, the issue we were seeing is something like this:
Thread A: enters device 1 from device 0
Thread B: enters device 1 from device 2
Thread A: exits to device 2 erroneously (popping the most recent entry from the non-thread-local stack)
Thread B: exits to device 1 erroneously

The expected behavior is:
Thread A: enters device 1 from device 0
Thread B: enters device 1 from device 2
Thread A: exits to device 0
Thread B: exits to device 2

I think the fix is to have thread-local copies of _device_stack associated with each Device object, not in the thread-local storage that's global to the module. I can try to throw together a PR if that'd help. I don't have a development environment set up for cupy at the moment though, so there may be some delay.

The alternative is to tell users to never use the same Device object in different threads simultaneously. That's doable, but at least in our case, we'll likely end up writing a similar thread-local storage idiom downstream to make the right thing happen.

insertinterestingnamehere:
I think a workaround downstream is to refer to devices based off of their ids and initialize a new cupy.cuda.Device object every time we want to use the context manager and just store device ids. Neither this solution nor the thread-local storage one suggested above has any significant performance impact.

kmaehashi:
Thanks for reporting @insertinterestingnamehere, I think it makes sense to keep the device ID history in TLS as the device ID is set per-thread in CUDA.

Issue #5049, opened 2021-04-07
CuPy v10 Release Schedule


asi1024:
Here is the current plan for CuPy v10 development.

## Highlights in CuPy v10

- Enhanced Multi-GPU Support
  - `cupyx.distributed`
  - Peer access support
- #4789
  - [Python array API standard](https://data-apis.org/array-api/latest/)
  - ref: https://github.com/numpy/numpy/pull/18585
  - https://github.com/cupy/cupy/pull/5698
- More cuTENSOR support
  - Implement element-wise operation using cuTENSOR

## Timeline

- [x] **v10.0.0a1 (v9.1.0)** (2021-05-27)
- [x] **v10.0.0a2 (v9.2.0)** (2021-06-24)
- [x] **v10.0.0b1 (v9.3.0)** (2021-08-05)
- [x] **v10.0.0b2 (v9.4.0)** (2021-08-26)
- [x] **v10.0.0b3 (v9.5.0)** (2021-09-30)
- [x] **v10.0.0rc1 (v9.6.0)** (2021-11-11)
- [x] **v10.0.0** (2021-12-09)

Release Blocking Issues: https://github.com/cupy/cupy/labels/blocking

Notes:

* The schedule and listed features are tentative and may subject to change depending on the progress.
* The code will be frozen around 15:00 on 2 days before the release date for QA. Dates are in JST.
* `v9.x.x` releases will only contain backported pull-requests (mainly bug-fixes) from the v9 (master) branch.

## Past Release Schedules

* v9: https://github.com/cupy/cupy/issues/3891

kmaehashi:
Due to scheduling reasons, v10.0.0b1 / v9.3.0 release date has been changed from 2021-07-29 to 2021-08-05.

kmaehashi:
As there are several remaining issues and pull-requests that we would like to include in v10.0.0rc1 release, we are updating the release schedule as follows:

* v10.0.0rc1 (v9.6.0) (~2021-10-28~) -> (2021-11-11)
* v10.0.0 (~2021-11-25~) -> (2021-12-09)

asi1024:
Released v10.0.0! https://github.com/cupy/cupy/releases/tag/v10.0.0

PR #5083, opened 2021-04-13
Make cupy.cuda.Device context manager interface thread safe

insertinterestingnamehere:
WIP. Fixes https://github.com/cupy/cupy/issues/5006.

### DIFF
--- a/cupy/cuda/device.pyx
+++ b/cupy/cuda/device.pyx
@@ -21,6 +21,28 @@ cdef dict _devices = {}
 cdef dict _compute_capabilities = {}
 
 
+cdef class _ThreadLocalStack:
+    cdef list _devices
+
+    def __init__(self):
+        self._devices = []
+
+    @staticmethod
+    cdef _ThreadLocalStack get():
+        try:
+            stack = _thread_local._device_stack
+        except AttributeError:
+            stack = _ThreadLocalStack()
+            _thread_local._device_stack = stack
+        return <_ThreadLocalStack>stack
+
+    cdef void push(self, int device_id) except *:
+        self._devices.append(device_id)
+
+    cdef int pop(self) except -1:
+        return <int>self._devices.pop()
+
+
 cpdef int get_device_id() except? -1:
     return runtime.getDevice()
 
@@ -114,8 +136,6 @@ cdef class Device:
         else:
             self.id = int(device)
 
-        self._device_stack = []
-
     @classmethod
     def from_pci_bus_id(cls, pci_bus_id):
         """Returns a new device instance based on a PCI Bus ID
@@ -138,13 +158,13 @@ cdef class Device:
 
     def __enter__(self):
         cdef int id = runtime.getDevice()
-        self._device_stack.append(id)
+        _ThreadLocalStack.get().push(id)
         if self.id != id:
             self.use()
         return self
 
     def __exit__(self, *args):
-        runtime.setDevice(self._device_stack.pop())
+        runtime.setDevice(_ThreadLocalStack.get().pop())
 
     def __repr__(self):
         return '<CUDA Device %d>' % self.id
--- a/tests/cupy_tests/cuda_tests/test_device.py
+++ b/tests/cupy_tests/cuda_tests/test_device.py
@@ -155,3 +155,40 @@ class TestDeviceHandles(unittest.TestCase):
 class TestDeviceFromPointer(unittest.TestCase):
     def test_from_pointer(self):
         assert cuda.device.from_pointer(cupy.empty(1).data.ptr).id == 0
+
+
+@testing.multi_gpu(2)
+class TestDeviceContextManager(unittest.TestCase):
+    def test_thread_safe(self):
+        dev0 = cuda.Device(0)
+        dev1 = cuda.Device(1)
+
+        t0_setup = threading.Event()
+        t1_setup = threading.Event()
+        t0_first_exit = threading.Event()
+
+        t0_exit_device = []
+        t1_exit_device = []
+
+        def t0_seq():
+            with dev0:
+                with dev0:
+                    t0_setup.set()
+                    t1_setup.wait()
+                t0_exit_device.append(cuda.Device().id)
+                t0_first_exit.set()
+
+        def t1_seq():
+            t0_setup.wait()
+            with dev1:
+                with dev0:
+                    t1_setup.set()
+                    t0_first_exit.wait()
+                t1_exit_device.append(cuda.Device().id)
+
+        t1 = threading.Thread(target=t1_seq)
+        t1.start()
+        t0_seq()
+        t1.join()
+        assert t0_exit_device[0] == 0
+        assert t1_exit_device[0] == 1


insertinterestingnamehere:
Okay, I've managed to confirm that this works locally. What else needs to be done to get this merged?

cupy/cuda/device.pyx
```diff
@@ -15,7 +15,19 @@ from cupy import _util
...
+
+class _ThreadLocalStack(threading.local):
+    def __init__(self):
+        self._device_stack = []
+
+    def push_device(self, value):
+        self._device_stack.append(value)
+
+    def pop_device(self):
+        return self._device_stack.pop()
```
leofang, line 27: Could you follow the style used elsewhere in the codebase and write it as `cdef class _ThreadLocal:` without inheriting the `thread.local` class, and convert all methods to `cdef`? See, for example, 
https://github.com/cupy/cupy/blob/596d1af53b5793d3d52994c8b493ff42be453a8d/cupy_backends/cuda/stream.pyx#L13-L30

tests/cupy_tests/cuda_tests/test_device.py
```diff
@@ -155,3 +155,39 @@ def test_cusparse_handle(self):
 class TestDeviceFromPointer(unittest.TestCase):
     def test_from_pointer(self):
         assert cuda.device.from_pointer(cupy.empty(1).data.ptr).id == 0
+
+
+class TestDeviceContextManager(unittest.TestCase):
+    def test_thread_save(self):
```
leofang, line 161:
```suggestion
@testing.multi_gpu(2)
class TestDeviceContextManager(unittest.TestCase):
    def test_thread_safe(self):
```

leofang (changes requested):
LGTM except for a style nit!

cupy/cuda/device.pyx
insertinterestingnamehere: Will do!

insertinterestingnamehere: There wasn't any state to encapsulate aside from the thread-local stack itself, so I routed it through an accessor function. That's not exactly the same as what you suggested, but it is consistent with other code in this file. LMK if the new version is okay.

tests/cupy_tests/cuda_tests/test_device.py
insertinterestingnamehere: Fixed. Thank you!

cupy/cuda/device.pyx
leofang: Hi @insertinterestingnamehere, I think you misunderstood me 😅 Could you take a look again at the `_ThreadLocal` class that I linked above? The state can be encoded as a cdef attribute (ex: `current_stream` there can be replaced by your `_device_stack`). So, your previous implementation was actually fine, just need to make it a cdef class, add a `cdef list _device_stack` attribute, etc.

insertinterestingnamehere: Oh, yah I totally misunderstood the suggestion. Thanks for clarifying! I'll switch it back on Monday.

insertinterestingnamehere: Actually Cython doesn't let us subclass from a non extension type with an extension type, so `cdef class` doesn't work here. Given that, what version do you prefer? I can switch it back to a plain Python class or just go with the accessor function as it currently is. Alternatively I can store the thread-local object as a member of an extension type and then move the accessor routine _get_handle to be an attribute on the new class instead. That's probably the best change, but it'll involve some additional refactoring outside of the code that handles the device stack. Really all three approaches work fine.

leofang: Hi @insertinterestingnamehere, please kindly read my original https://github.com/cupy/cupy/pull/5083#discussion_r615008880 again. My suggestion was to *not* inherit the `threading.local` class (for the very same reason you correctly pointed out). The `_ThreadLocal` implementation that I linked to (which is adopted in many places in the codebase) should follow closely with your original implementation. 
>  Alternatively I can store the thread-local object as a member of an extension type and then move the accessor routine _get_handle to be an attribute on the new class instead.

This is indeed the idea.

insertinterestingnamehere: Okay, got it. For some reason I had assumed it'd be best to avoid changing the `_get_handle` code at all and was trying to interpret your suggestion in that light. Sure, I can refactor the `_get_handle` code here too.

leofang: Wait, I am confused, why do we need to touch `_get_handle`?

insertinterestingnamehere: _get_handle currently routes all the thread-local data used in this file through _thread_local. By subclassing I let it continue doing that without any changes since _thread_local can still be used that way even with the added interface for device stack management. If I switch it to just being an attribute in the new class then I can't re-use the same thread-local object.

Any of the following would work: just using a different thread-local object, moving _get_handle to be a member function on the new class and using that interface instead of _get_handle (this one makes the most sense to me), or having _get_handle access the thread-local member of the new _thread_local object instead of using the _thread_local object itself. What do you prefer?

insertinterestingnamehere: The reason I tried writing _get_device_stack is because that roughly follows the idea behind _get_handle. It provides an accessor function to manage something in a module-level thread-local object.

leofang: Right, I was being silly today, sorry...I agree making both `_get_handle` and `_get_device_stack` as the methods of the new `_ThreadLocal` class makes sense.

insertinterestingnamehere: No worries. I've pushed a new version that will hopefully work well. Thanks for your input!

cupy/cuda/device.pyx
```diff
@@ -25,6 +24,43 @@ cpdef int get_device_id() except? -1:
     return runtime.getDevice()
 
 
+cdef class _ThreadLocalStack:
+
+    cdef object _local
+
+    def __init__(self):
+        self._local = threading.local()
```
leofang, line 32: This is not correct unfortunately. What's done here will create the `_ThreadLocal` instance only in the thread that initializes this module; other new threads do not have this object. 

Please follow the practice done elsewhere and define a factory method `get()` so that each thread has its own copy of `_ThreadLocal`:
https://github.com/cupy/cupy/blob/596d1af53b5793d3d52994c8b493ff42be453a8d/cupy_backends/cuda/stream.pyx#L16-L22

cupy/cuda/device.pyx
```diff
@@ -15,7 +15,6 @@ from cupy import _util
 # It is always True as cuSOLVER library is always available in CUDA 8.0+.
 cusolver_enabled = True
 
-cdef object _thread_local = threading.local()
```
leofang, line 18: This should be restored for the reason below (above?)

cupy/cuda/device.pyx
```diff
@@ -25,6 +24,43 @@ cpdef int get_device_id() except? -1:
...
+            return handle.handle
+        with device:
+            handle = create_func()
+            handles[device.id] = Handle(handle, destroy_func)
+            return handle
+
+
+cdef _ThreadLocalStack _thread_local = _ThreadLocalStack()
+
+
```
leofang, line 63: This is not needed for the reason stated above (below?)
```suggestion
```

leofang (changes requested):

cupy/cuda/device.pyx
insertinterestingnamehere: It is correct as written. The thread-local object itself is visible to all threads. Its attributes vary depending on which thread accesses them. The equivalent of the idiom you're pointing to here is handled further down where I do
```
current = getattr(self._local, "_device_stack", None)
if current is None:
    current = []
    self._local._device_stack = current
```
In your try-except example the try is to detect whether or not the attribute on the thread-local object is present. The thread-local object is present in all threads. That's why it's catching an AttributeError, not a NameError.

insertinterestingnamehere: I guess I can try to refactor to more closely mimic the idiom used in the stream file if that'd help.

cupy/cuda/device.pyx
```diff
@@ -25,6 +24,43 @@ cpdef int get_device_id() except? -1:
     return runtime.getDevice()
 
 
+cdef class _ThreadLocalStack:
```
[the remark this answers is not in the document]
insertinterestingnamehere: Sure. Will do.

cupy/cuda/device.pyx
leofang: No please don't. I really had a bad day and wasn't thinking straight. I think you are correct. Sorry for the disturbance.

cupy/cuda/device.pyx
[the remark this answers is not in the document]
leofang: Let's do this name change and I'll fire up the CI. 🙂

cupy/cuda/device.pyx
insertinterestingnamehere: Okay, if I _exactly_ follow the idiom in the other file we end up implicitly creating an unnecessary Cython vtable implicitly. I've pushed a version that does that though.

cupy/cuda/device.pyx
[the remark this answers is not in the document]
insertinterestingnamehere: Not really applicable with the new refactored version. Thanks though.

insertinterestingnamehere:
Can we get a CI run here?

cupy/cuda/device.pyx
leofang: Thanks! Sorry again for my unreasonable requests.

leofang (approved):
LGTM! I verified the test indeed fails on the current master.

leofang:
Sorry for the long wait. This week there was a code frozen for the v9.0.0 release, so the CI had to be reserved for internal tests IIUC.

Jenkins, test this please

cupy/cuda/device.pyx
insertinterestingnamehere: No problem. Thanks for taking the time to review the various versions!

insertinterestingnamehere:
Great, thanks for running the CI!

kmaehashi (approved):
LGTM!

kmaehashi:
Hi @insertinterestingnamehere, sorry to have kept you waiting due to the v9 release.
@leofang Thanks for reviewing!

leofang:
Thanks for reporting and fixing the issue @insertinterestingnamehere!

insertinterestingnamehere:
Thanks for your help reviewing! I'm unfamiliar with the CuPy release schedule. It seems like releases are pretty frequent. Would it be possible to get this into 9.1? Roughly when will that release be?

kmaehashi:
@insertinterestingnamehere In general we don't backport BC-breaking changes like this one to the stable branch, but I think  this one can be an exception because this issue can also be considered as a bug-fix (rather than enhancements). We'll merge it to v9.1 which is planned on next month (see #5049 for release plans).

insertinterestingnamehere:
Thanks! That'd be really helpful downstream. Do you want me to submit a backport PR to the v9 branch?

leofang:
@insertinterestingnamehere The bot has already done that: #5147 🙂

insertinterestingnamehere:
Cool! I didn't realize you had that automated. Thanks!

PR #5087, opened 2021-04-15
Fix poisson to support lam array

emcastillo:
We were not supporting arrays for lam parameter ...

~merge after #4907~

### DIFF
--- a/cupy/random/_generator_api.pyx
+++ b/cupy/random/_generator_api.pyx
@@ -333,10 +333,8 @@ class Generator:
                 lam_a.fill(lam)
                 lam = lam_a
             else:
-                if lam is None:
-                    raise TypeError('lam must be real number, not NoneType')
-                raise ValueError('lam is required to be a cupy.ndarray'
-                                 ' or a scalar')
+                raise TypeError('lam is required to be a cupy.ndarray'
+                                ' or a scalar')
         else:
             # Check if size is broadcastable to shape
             # but size determines the output
--- a/cupy/random/_generator_api.pyx
+++ b/cupy/random/_generator_api.pyx
@@ -333,6 +333,8 @@ class Generator:
                 lam_a.fill(lam)
                 lam = lam_a
             else:
+                if lam is None:
+                    raise TypeError('lam must be real number, not NoneType')
                 raise ValueError('lam is required to be a cupy.ndarray'
                                  ' or a scalar')
         else:
@@ -436,6 +438,8 @@ class Generator:
                 shape_a.fill(shape)
                 shape = shape_a
             else:
+                if shape is None:
+                    raise TypeError('shape must be real number, not NoneType')
                 raise ValueError('shape is required to be a cupy.ndarray'
                                  ' or a scalar')
         else:
--- a/cupy/random/_generator_api.pyx
+++ b/cupy/random/_generator_api.pyx
@@ -329,7 +329,7 @@ class Generator:
 
         if not isinstance(lam, ndarray):
             if type(lam) in (float, int):
-                lam_a = ndarray(1, numpy.float64)
+                lam_a = ndarray((), numpy.float64)
                 lam_a.fill(lam)
                 lam = lam_a
             else:
@@ -432,7 +432,7 @@ class Generator:
 
         if not isinstance(shape, ndarray):
             if type(shape) in (float, int):
-                shape_a = ndarray(1, numpy.float64)
+                shape_a = ndarray((), numpy.float64)
                 shape_a.fill(shape)
                 shape = shape_a
             else:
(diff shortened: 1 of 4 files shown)


emcastillo:
rebased, PTAL

cupy/random/_generator_api.pyx
```diff
@@ -320,8 +325,32 @@ class Generator:
             <numpy.random.generator.Generator.poisson>`
         """
         cdef ndarray y
+        cdef ndarray lam_arr
+
+        if not isinstance(lam, ndarray):
+            if type(lam) in (float, int):
+                lam_a = ndarray(1, numpy.float64)
```
toslunar, line 332:
```suggestion
                lam_a = ndarray((), numpy.float64)
```

`standard_gamma` has the same issue.  `cupy.random.default_rng(0).standard_gamma(3.7).shape` returns `(1,)` but it should be `()`.

toslunar (changes requested):

cupy/random/_generator_api.pyx
emcastillo: fixed! thanks :)

tests/cupy_tests/random_tests/test_distributions.py
```diff
@@ -526,17 +526,26 @@ def test_pareto(self, a_dtype, dtype):
...
 
     @cupy.testing.for_int_dtypes('dtype')
     @cupy.testing.for_float_dtypes('lam_dtype')
-    def test_poisson(self, lam_dtype, dtype):
+    def test_poisson_legacy(self, lam_dtype, dtype):
         self.check_distribution(_distributions.poisson, lam_dtype, dtype)
 
+    @cupy.testing.for_int_dtypes('dtype')
+    @cupy.testing.for_float_dtypes('lam_dtype')
+    def test_poisson_generator(self, lam_dtype, dtype):
```
toslunar, line 545:
```suggestion
    @cupy.testing.for_float_dtypes('lam_dtype')
    def test_poisson_generator(self, lam_dtype):
```

Remove the unused arg `dtype`.

cupy/random/_generator_api.pyx
toslunar: How about adding `shape=None` parameter to `TestDistributionsPoisson` (and in `TestDistributionsStandardGamma`)?

toslunar (changes requested):

tests/cupy_tests/random_tests/test_distributions.py
emcastillo: Thanks!

cupy/random/_generator_api.pyx
emcastillo: ok, lets add it!

emcastillo: Added for both in legacy/generator APIs, legacy throws a different error.

toslunar:
Sorry.  I meant `'shape': [..., None]` in `testing.parameterize` to pass `size=None` to distributions in the tests.

emcastillo:
Updated 😄

cupy/random/_generator_api.pyx
```diff
@@ -320,8 +325,34 @@ class Generator:
...
+        if not isinstance(lam, ndarray):
+            if type(lam) in (float, int):
+                lam_a = ndarray((), numpy.float64)
+                lam_a.fill(lam)
+                lam = lam_a
+            else:
+                if lam is None:
+                    raise TypeError('lam must be real number, not NoneType')
+                raise ValueError('lam is required to be a cupy.ndarray'
+                                 ' or a scalar')
```
toslunar, line 339: `lam < 0` raises `ValueError`.  Is there any other case with `ValueError`?

emcastillo: I think no, also we don't check the value of lam to avoid synchronization.

Should we check it if it's a regular scalar value instead of a ndarray?

toslunar:
```suggestion
                raise TypeError('lam is required to be a cupy.ndarray'
                                ' or a scalar')
```

BTW, I found a `ValueError` case, but I didn't get NumPy's intention.
```
>>> np.random.default_rng().poisson(1+0j)
TypeError: can't convert complex to float
>>> np.random.default_rng().poisson('inf')
TypeError: must be real number, not str
>>> np.random.default_rng().poisson('foo')
ValueError: could not convert string to float: 'foo'
>>> np.__version__
'1.20.2'
```

emcastillo: oh! I didn't though of testing string, I just assumed that all primitive types would result in `TypeError`
Good catch!

emcastillo: should we match this behavior too?

toslunar:
Jenkins, test this please.

toslunar (approved):

emcastillo:
I think we can merge this :D

PR #5091, opened 2021-04-16
Fix stream usage on D2D copy test under HIP

takagi:
Follows up #5004 under HIP environment, keeping Device 0's stream alive over the D2D copy as well as giving longer enough busy-wait.

### DIFF
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -204,7 +204,7 @@ class TestNdarrayCopy(unittest.TestCase):
         # Allocates a memory and launches the kernel on a device with its
         # stream.
         with cuda.Device(0):
-            with cuda.Stream() as s1:  # Keep the stream alive for HIP
+            with cuda.Stream() as s1:  # NOQA: Keep the stream alive for HIP
                 a = cupy.zeros((2,), dtype=numpy.uint64)
                 kern((1,), (1,), a)
 
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -204,7 +204,8 @@ class TestNdarrayCopy(unittest.TestCase):
         # Allocates a memory and launches the kernel on a device with its
         # stream.
         with cuda.Device(0):
-            with cuda.Stream() as s1:  # NOQA: Keep the stream alive for HIP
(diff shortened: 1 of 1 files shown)


leofang (commented):
This is concerning...So looks like a simple `with Stream():` statement is not enough to keep it alive? I feel a better fix is to change this line
https://github.com/cupy/cupy/blob/cf224075cd5430cdd692eb52872c6f9b67416095/cupy/cuda/stream.pyx#L191
to use `set_current_stream_ref()`. What do you think @takagi?

takagi:
Thoughtful... Let me run the test with it!

takagi:
Wait, let me see it further...

takagi:
Make me confirm that I added `s1` to make the stream alive over the second block that invokes D2D copy (as opposed to the first one that launches the kernel):
https://github.com/cupy/cupy/blob/bfc671e9f1d72e0be30408e1e6dcfd97a9f9004d/tests/cupy_tests/core_tests/test_ndarray.py#L214-L218

I don't think changing to `set_current_stream_ref()` there helps as `{get,set}_current_stream_ref()` are just fast paths to handle the stack of the current streams without repeatedly wrap/unwrap them by weakref.

Have I got your point?

leofang:
@takagi Thank you for explanation. I misunderstood and it's clear to me now 👍

leofang:
A follow-up question: Do you know what's special about HIP that the D2D test could not pass without this PR?

takagi:
The documentations say as below and only CUDA seems to ensure all work in the queue is completed when `cudaStreamDestroy()` is called when the device is still doing work.

**CUDA**
> In case the device is still doing work in the stream stream when cudaStreamDestroy() is called, the function will return immediately and the resources associated with stream will be released automatically once the device has completed all work in stream.

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1gfda584f1788ca983cb21c5f4d2033a62

**HIP**
> If commands are still executing on the specified stream, some may complete execution before the queue is deleted.
> 
> The queue may be destroyed while some commands are still inflight, or may wait for all commands queued to the stream before destroying it.

http://rocm-developer-tools.github.io/HIP/group__Stream.html#ga3076a3499ed2c7821311006100bb95ec

leofang:
Jenkins, test this please

emcastillo (approved):

Issue #5095, opened 2021-04-19
Support CUDA 11.3


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read [CUDA Release Notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html).
- [x] Update CUDA driver in Linux (Jenkins) / Windows (FlexCI) test infrastructure (ask @kmaehashi). -> ongoing
- [x] Update chainer-test to test the CUDA version. https://github.com/chainer/chainer-test/pull/658
- [x] Update compute capability support.
  - [x] Check [NVRTC Reference (--gpu-architecture option)](https://docs.nvidia.com/cuda/nvrtc/index.html#group__options) and change the maximum value of supported compute capability defined in [cupy/cuda/compiler.py](https://github.com/cupy/cupy/blob/master/cupy/cuda/compiler.py) (`_nvrtc_max_compute_capability`) if needed. -> *no changes needed*
  - [x] Check [NVCC Reference (--generate-code)](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation-generate-code) and change the list of architectures in [cupy_setup_build.py](https://github.com/cupy/cupy/blob/master/cupy_setup_build.py) (`_arch_list`) if needed. -> *no changes needed*
  - [X] Backport. -> *not needed*
- [x] Update bundled FP16 header files.
  - [x] Copy FP16 header files to `cupy/core/include/cupy/_cuda/`, then update `cupy/core/core.pyx` to use it. https://github.com/cupy/cupy/pull/5218
  - [x] Backport. #5232
- [x] Update library installer.
  - [x] Fix `cupyx/tools/install_library.py`. https://github.com/cupy/cupy/pull/5219
  - [x] Backport. #5235
- [x] Wheel distribution -> https://github.com/cupy/cupy-release-tools/pull/111
  - [x] Update cupy-release-tools. https://github.com/cupy/cupy-release-tools/pull/111
  - [x] Ask PyPI team to increase the limit.
- [x] Documentation
  - [x] Add new wheel package to [Installtion Guide](https://docs.cupy.dev/en/latest/install.html) and `README.md`. Update requirements in the installation guide. https://github.com/cupy/cupy/pull/5220
  - [x] Backport.
[...]

PR #5097, opened 2021-04-19
Compile with SASS for CUDA versions >= 11.1

### LINKED ISSUE #4591: Make `nvrtc.getCUBIN()` default on CUDA 11.1+?
In #4558 the new `nvrtcGetCUBIN()` API was exposed. The purpose was to avoid the driver from performing just-in-time compilation of PTX codes, which [was shown slow unless the jitted kernels are cached](https://github.com/cupy/cupy/pull/4558#issuecomment-765610834) (CUDA's default behavior). However, we decided to temporarily disable this feature in 8cb443f. This issue is for discussing the strategy to utilize this feature, which IMHO is a promising one 🙂

emcastillo:
This is needed to allow things like running code generated with cuda 11.3 and run it in an older driver although the driver is marked as  "supported" by [nvidia](https://docs.nvidia.com/cuda/nvrtc/index.html#versioning):

> the more recent NVRTC library1. However, the more recent NVRTC library may generate PTX with a version that is not accepted by the CUDA Driver API functions of an older CUDA driver, as explained in the CUDA Compatibility document. 
Some approaches to resolving this issue:
Install a more recent CUDA driver that is compatible with the CUDA toolkit containing the NVRTC library being used.
Compile directly to SASS instead of PTX with NVRTC (see CUDA Compatibility document).

If we compile CuPy with cuda 11.3 and run it with an old driver, we will get the invalid PTX errors. 
by compiling using cubins everything will work fine.

Thanks to @leofang for implementing the initial support

Closes #4591

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -120,40 +120,50 @@ def _get_nvrtc_version():
 _tegra_archs = ('53', '62', '72')
 
 
+def _get_max_compute_capability():
+    major, minor = _get_nvrtc_version()
+    if major < 10 or (major == 10 and minor == 0):
+        # CUDA 9.x / 10.0
+        nvrtc_max_compute_capability = '70'
+    elif major < 11:
+        # CUDA 10.1 / 10.2
+        nvrtc_max_compute_capability = '75'
+    elif major == 11 and minor == 0:
+        # CUDA 11.0
+        nvrtc_max_compute_capability = '80'
+    else:
+        # CUDA 11.1 / 11.2 / 11.3 / 11.4
+        nvrtc_max_compute_capability = '86'
+    return nvrtc_max_compute_capability
+
+
 @_util.memoize(for_each_device=True)
 def _get_arch():
     # See Supported Compile Options section of NVRTC User Guide for
     # the maximum value allowed for `--gpu-architecture`.
-    major, minor = _get_nvrtc_version()
-    if major < 10 or (major == 10 and minor == 0):
-        # CUDA 9.x / 10.0
-        _nvrtc_max_compute_capability = '70'
-    elif major < 11:
-        # CUDA 10.1 / 10.2
-        _nvrtc_max_compute_capability = '75'
-    elif major == 11 and minor == 0:
-        # CUDA 11.0
-        _nvrtc_max_compute_capability = '80'
-    else:
-        # CUDA 11.1 / 11.2 / 11.3 / 11.4
-        _nvrtc_max_compute_capability = '86'
+    nvrtc_max_compute_capability = _get_max_compute_capability()
 
     arch = device.Device().compute_capability
     if arch in _tegra_archs:
         return arch
     else:
-        return min(arch, _nvrtc_max_compute_capability)
+        return min(arch, nvrtc_max_compute_capability)
 
 
-def _get_arch_for_options(arch=None, jitify=False):
-    # This is needed to differentiate between
-    # compute or sm depending on the CUDA version
-    # needed to ensure backwards compatibility with nvrtc
+def _get_arch_for_options_for_nvrtc(arch=None, jitify=False):
+    # NVRTC in CUDA 11.3+ generates PTX that cannot be run an earlier driver
+    # version than the one included in the used CUDA version, as
+    # documented in:
+    # https://docs.nvidia.com/cuda/archive/11.3.0/nvrtc/index.html#versioning
+    # Here we use `-arch=sm_*` instead of `-arch=compute_*` to directly
+    # generate cubin (SASS) instead of PTX. See #5097 for details.
     if arch is None:
         arch = _get_arch()
-    if _cuda_hip_version >= 11010 and not jitify:
-        return f'-arch=sm_{arch}'
-    return f'-arch=compute_{arch}'
+    if (_cuda_hip_version >= 11010
+            and arch < _get_max_compute_capability()
+            and not jitify):
+        return f'-arch=sm_{arch}', 'nvrtc'
+    return f'-arch=compute_{arch}', 'ptx'
 
 
 def _is_cudadevrt_needed(options):
@@ -256,10 +266,11 @@ def compile_using_nvrtc(source, options=(), arch=None, filename='kern.cu',
         else:
             headers = include_names = ()
 
-        prog = _NVRTCProgram(source, cu_path, headers, include_names,
-                             name_expressions=name_expressions)
+        arch_opt, method = _get_arch_for_options_for_nvrtc(arch, jitify)
+        options += (arch_opt,)
 
-        options += (_get_arch_for_options(arch=arch, jitify=jitify),)
+        prog = _NVRTCProgram(source, cu_path, headers, include_names,
+                             name_expressions=name_expressions, method=method)
 
         try:
             if _cuda_hip_version >= 11010 and jitify:
@@ -381,7 +392,7 @@ def compile_using_nvcc(source, options=(), arch=None,
 
 def _preprocess(source, options, arch, backend):
     if backend == 'nvrtc':
-        if not runtime.is_hip and _cuda_hip_version >= 11010:
+        if _cuda_hip_version >= 11010:
             options += ('-arch=sm_{}'.format(arch),)
         else:
             options += ('-arch=compute_{}'.format(arch),)
@@ -484,8 +495,8 @@ def _compile_with_cache_cuda(
     if jitify and backend != 'nvrtc':
         raise ValueError('jitify only works with NVRTC')
 
-    env_options = options + (_get_arch_for_options(arch, jitify),)
-    env = (arch, env_options, _get_nvrtc_version(), backend)
+    env = ((arch, options, _get_nvrtc_version(), backend)
+           + _get_arch_for_options_for_nvrtc(arch, jitify))
     base = _empty_file_preprocess_cache.get(env, None)
     if base is None:
         # This is for checking NVRTC/NVCC compiler internal version
@@ -614,7 +625,7 @@ class CompileException(Exception):
 class _NVRTCProgram(object):
 
     def __init__(self, src, name='default_program', headers=(),
-                 include_names=(), name_expressions=None):
+                 include_names=(), name_expressions=None, method='ptx'):
         self.ptr = None
 
         if isinstance(src, bytes):
@@ -626,6 +637,7 @@ class _NVRTCProgram(object):
         self.name = name
         self.ptr = nvrtc.createProgram(src, name, headers, include_names)
         self.name_expressions = name_expressions
+        self.method = method
 
     def __del__(self, is_shutting_down=_util.is_shutting_down):
         if is_shutting_down():
@@ -647,9 +659,12 @@ class _NVRTCProgram(object):
             if log_stream is not None:
                 log_stream.write(nvrtc.getProgramLog(self.ptr))
             # This is to ensure backwards compatibility with nvrtc
-            if not runtime.is_hip and _cuda_hip_version >= 11010:
+            if self.method == 'cubin':
                 return nvrtc.getCUBIN(self.ptr), mapping
-            return nvrtc.getPTX(self.ptr), mapping
+            elif self.method == 'ptx':
+                return nvrtc.getPTX(self.ptr), mapping
+            else:
+                raise RuntimeError('Unknown NVRTC compile method')
         except nvrtc.NVRTCError:
             log = nvrtc.getProgramLog(self.ptr)
             raise CompileException(log, self.src, self.name, options,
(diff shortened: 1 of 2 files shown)


leofang (changes requested):
Notes: 
1. This PR is in a sense reverting https://github.com/cupy/cupy/pull/4558/commits/8cb443f7ad131635c92cde832e4b7d25900cd04f from #4558.
2. This PR will close #4591.

cupy/cuda/compiler.py
```diff
@@ -143,6 +143,17 @@ def _get_arch():
...
 
+def _get_arch_for_options(arch=None):
+    # This is needed to differentiate between
+    # compute or sm depending on the CUDA version
+    # needed to ensure backwards compatibility with nvrtc
+    if arch is None:
+        arch = _get_arch()
+    if _cuda_version >= 11010:
+        return f'-arch=sm_{arch}'
+    return f'-arch=compute_{arch}'
```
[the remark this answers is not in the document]
emcastillo: I confirmed it didn't work

emcastillo:
I have seen failures with raw_kernel jitify and the dynamic parallelism.

cupy/cuda/compiler.py
```diff
@@ -236,9 +247,7 @@ def compile_using_nvrtc(source, options=(), arch=None, filename='kern.cu',
                         cache_in_memory=False, jitify=False):
     # For hipRTC, arch is ignored
     if not runtime.is_hip:
-        if not arch:
-            arch = _get_arch()
-        options += ('-arch=compute_{}'.format(arch),)
+        options += (_get_arch_for_options(),)
```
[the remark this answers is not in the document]
emcastillo: yeah, this is a good idea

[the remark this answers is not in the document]
leofang: How about this? We first do `git revert 8cb443f` and then fix the cache key?

kmaehashi:
CUDA 11.4 with CUDA 11.3 (or earlier) driver has the same issue as  https://github.com/cupy/cupy/issues/5095#issuecomment-833003035.
Let's prioritize this PR.

emcastillo:
PTAL!

@leofang I included the revert here :)

leofang:
Took a quick look and it seems ok. Will take a second look tonight.

Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

cupy/cuda/compiler.py
```diff
@@ -368,8 +380,10 @@ def compile_using_nvcc(source, options=(), arch=None,
 
 def _preprocess(source, options, arch, backend):
     if backend == 'nvrtc':
-        options += ('-arch=compute_{}'.format(arch),)
-
+        if not runtime.is_hip and _cuda_hip_version >= 11010:
```
leofang, line 383: On HIP we don't call this function (there are `_preprocess_{hiprtc,hipcc}`) so we can simplify a bit:
```suggestion
        if _cuda_hip_version >= 11010:
```

cupy/cuda/compiler.py
```diff
@@ -263,6 +273,8 @@ def _compile(
             raise
         return ptx, mapping
 
+    options += (_get_arch_for_options(arch=arch, jitify=jitify),)
```
leofang, line 276: I don't think this should be here? (See my main comment.)

cupy/cuda/compiler.py
```diff
@@ -145,6 +145,17 @@ def _get_arch():
         return min(arch, _nvrtc_max_compute_capability)
 
 
+def _get_arch_for_options(arch=None, jitify=False):
+    # This is needed to differentiate between
+    # compute or sm depending on the CUDA version
+    # needed to ensure backwards compatibility with nvrtc
+    if arch is None:
+        arch = _get_arch()
+    if not runtime.is_hip and _cuda_hip_version >= 11010 and not jitify:
```
leofang, line 154: Let's remove the HIP check, as when this function is called we are guaranteed on CUDA:
```suggestion
    if _cuda_hip_version >= 11010 and not jitify:
```

leofang (commented):
@emcastillo I am puzzled by the last commit 0915b94. I think if we wanna invalidate users' existing kernel cache, `options` should be modified before we use it to construct `env`, so we need to call `_get_arch_for_options` a few lines *earlier*, not *later*. Am I missing anything?

cupy/cuda/compiler.py
emcastillo: the cache key is obtained in preprocess :), if we don't do it here, some compiler calls may fail. (There were some tests failing because of this in `cuda_tests/test_compiler`

emcastillo:
Jenkins, test this please

cupy/cuda/compiler.py
leofang: Is it possible to simplify it as follows?
```python
if backend == 'nvrtc':
    options += (_get_arch_for_options(arch=arch, jitify=jitify),)
```
Perhaps it'd allow you to use it as part of the cache key *and* fix the compiler error without calling `_get_arch_for_options()` twice?

emcastillo: sorry I don't follow :),
the tests in `cuda_tests/test_compiler` call `compile_using_nvrtc`, so the cache checking and all that will be bypassed. and we need to check it here. I don't currently see a way to avoid the two calls :(

leofang: Ah got it, sorry I misread. I understand it now. `_compile_with_cache_cuda()` calls `compile_using_nvrtc()`. I was hoping we just need to fix the options once in the former, and I missed that the latter is used directly in the tests.

cupy/cuda/compiler.py
```diff
@@ -145,6 +145,17 @@ def _get_arch():
         return min(arch, _nvrtc_max_compute_capability)
 
 
+def _get_arch_for_options(arch=None, jitify=False):
```
kmaehashi, line 148: Better name for this? 😃 
```suggestion
def _get_arch_option_for_nvrtc(arch=None, jitify=False):
```

cupy/cuda/compiler.py
```diff
@@ -145,6 +145,17 @@ def _get_arch():
...
+    # This is needed to differentiate between
+    # compute or sm depending on the CUDA version
+    # needed to ensure backwards compatibility with nvrtc
```
kmaehashi, line 151:
```suggestion
    # NVRTC in CUDA 11.3+ generates PTX that cannot be run with the driver
    # version earlier than the one included in that CUDA version, as
    # documented in:
    # https://docs.nvidia.com/cuda/archive/11.3.0/nvrtc/index.html#versioning
    # Here we use `-arch=sm_*` instead of `-arch=compute_*` to directly
    # generate cubin (SASS) instead of PTX. See #5097 for details.
```

cupy/cuda/compiler.py
```diff
@@ -253,7 +258,15 @@ def _compile(
 
         prog = _NVRTCProgram(source, cu_path, headers, include_names,
                              name_expressions=name_expressions)
+
+        options += (_get_arch_for_options(arch=arch, jitify=jitify),)
```
kmaehashi, line 262:
```suggestion
        options += (_get_arch_option_for_nvrtc(arch=arch, jitify=jitify),)
```

cupy/cuda/compiler.py
```diff
@@ -253,7 +258,15 @@ def _compile(
...
+
         try:
+            if _cuda_hip_version >= 11010 and jitify:
+                # Convert the virtual arch to a real arch
```
kmaehashi, line 266: Isn't this already done in `_get_arch_for_options`?

cupy/cuda/compiler.py
kmaehashi: This seems not resolved?

cupy/cuda/compiler.py
```diff
@@ -145,6 +145,17 @@ def _get_arch():
         return min(arch, _nvrtc_max_compute_capability)
 
 
+def _get_arch_for_options(arch=None, jitify=False):
+    # This is needed to differentiate between
+    # compute or sm depending on the CUDA version
+    # needed to ensure backwards compatibility with nvrtc
+    if arch is None:
+        arch = _get_arch()
```
kmaehashi, line 153: This method does not always return the real device architecture. If we hit `_nvrtc_max_compute_capability` we should use `compute_*` instead.

cupy/cuda/compiler.py
```diff
@@ -145,6 +145,17 @@ def _get_arch():
...
+    if _cuda_hip_version >= 11010 and not jitify:
+        return f'-arch=sm_{arch}'
+    return f'-arch=compute_{arch}'
```
kmaehashi, line 156: Maybe need to return the flag and use it to decide which method to call (`getCUBIN` or `getPTX`)

```suggestion
        return f'-arch=sm_{arch}', 'cubin'
    return f'-arch=compute_{arch}', 'ptx'
```

cupy/cuda/compiler.py
```diff
@@ -469,7 +484,8 @@ def _compile_with_cache_cuda(
     if jitify and backend != 'nvrtc':
         raise ValueError('jitify only works with NVRTC')
 
-    env = (arch, options, _get_nvrtc_version(), backend)
+    env_options = options + (_get_arch_for_options(arch, jitify),)
+    env = (arch, env_options, _get_nvrtc_version(), backend)
```
kmaehashi, line 488: How about just adding the option to `env`?

kmaehashi (changes requested):

cupy/cuda/compiler.py
emcastillo: this is some kind of tricky case when jitify is used.

emcastillo: maybe @leofang can explain it better :)

cupy/cuda/compiler.py
emcastillo: do you mean

```
env = (arch, options, _get_nvrtc_version(), backend) + (_get_arch_for_options(arch, jitify),)
```

cupy/cuda/compiler.py
emcastillo: oops :)

cupy/cuda/compiler.py
emcastillo: maybe we can directly return the function?
`nvrtc.getPTX` or `nvrtc.getCUBIN`

kmaehashi: That's also possible but sounds a bit hacky considering the returned value will be used as a hash key.

cupy/cuda/compiler.py
kmaehashi: Yes!

```py
env = (arch, options, _get_nvrtc_version(), backend, _get_arch_for_options(arch, jitify))
```

emcastillo:
PTAL!

kmaehashi (changes requested):

cupy/cuda/compiler.py
```diff
@@ -368,8 +392,10 @@ def compile_using_nvcc(source, options=(), arch=None,
 
 def _preprocess(source, options, arch, backend):
     if backend == 'nvrtc':
-        options += ('-arch=compute_{}'.format(arch),)
-
+        if _cuda_hip_version >= 11010:
```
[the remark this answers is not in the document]
emcastillo: this is a good question, it was needed before but by adding the `method` to `NVRTCProgram` class preprocess will now always generate `ptx` in this case, so `arch=sm_` can be avoided here.
Before this change, `compute` would be mixed with cubin and cause an error.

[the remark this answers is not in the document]
kmaehashi: Thanks for clarification!

cupy/cuda/compiler.py
```diff
@@ -251,9 +267,18 @@ def _compile(
         else:
             headers = include_names = ()
 
+        arch_opt, method = _get_arch_for_options_for_nvrtc(arch, jitify)
```
kmaehashi, line 270: Is it ok to add `arch` even in HIP?

cupy/cuda/compiler.py
emcastillo: Now, by tracking the method, we can also make it completely `jitify` independent :), so we can get rid of these hacks too.

cupy/cuda/compiler.py
emcastillo: I think this code is not executed for hip, but `compile_using_hiprtc`

kmaehashi: Seems this method is called from `_compile_with_cache_hip`: https://github.com/cupy/cupy/blob/8378fb776d98bce9f56d7aa0a6dc8e2b17631b66/cupy/cuda/compiler.py#L847-L852 😅

emcastillo: yes, I got confused because there is both `compile_with_nvcc` and `compile_with_hipcc` :)
Should be fixed now :D

kmaehashi (approved):
LGTM!

kmaehashi:
pfnCI, test this please.

leofang:
LGTM, thanks @emcastillo @kmaehashi! Question: Should we backport this PR?

PR #5107, opened 2021-04-19
Add ROCm limitations to docs

kmaehashi:
@takagi could you fill these sections?

PR #5122, opened 2021-04-21
Revise ROCm doc

leofang:
This is just a nitpick PR following #5107:
- Supporting Jitify is not possible due to hiprtc's limitation. Plus I doubt we need Jitify once hiprtc fixes its support for `-I`.
- Dynamical parallelism on ROCm is hardware-specific. There's a device attribute or property that we can query. (My recollection is gfx906 does not support it, but I could be wrong.)


kmaehashi:
Thanks!

leofang:
> @kmaehashi Failed to backport automatically.

I guess this is because #5119 just went in?

kmaehashi:
Thanks, triggered the bot by relabelling it.

Issue #5000, opened 2021-03-30
Sparse SVD fails if matrix rank is lower than the requested number of  singular values k


PavelStishenko:
* Conditions
CuPy Version                 : 9.0.0rc1
NumPy Version                : 1.20.1
SciPy Version                : 1.6.2
CUDA Build Version           : 11010
CUDA Driver Version          : 11020
CUDA Runtime Version         : 11010

* Code to reproduce
```
import cupy as cp
from cupyx.scipy.sparse.linalg import svds

a = cp.random.random((10,10))
u,s,vt = svds(a, 5)
s[0] = 0.0
b = u @ cp.diag(s) @ vt
u,s,vt = svds(b, 5)	# TypeError: 'cupy.core.core.ndarray' object cannot be interpreted as an integer
```

* Error messages, stack traces, or logs
```
Traceback (most recent call last):
  File "/home/mazay/prg/regr/cupy_svds_bug.py", line 8, in <module>
    u,s,vt = svds(b, 5)	# TypeError: 'cupy.core.core.ndarray' object cannot be interpreted as an integer
  File "/home/mazay/prg/regr/ve/lib/python3.9/site-packages/cupyx/scipy/sparse/linalg/_eigen.py", line 363, in svds
    u = _augmented_orthnormal_cols(u, k - n_large)
  File "/home/mazay/prg/regr/ve/lib/python3.9/site-packages/cupyx/scipy/sparse/linalg/_eigen.py", line 373, in _augmented_orthnormal_cols
    y = cupy.empty((m, n + n_aug), dtype=x.dtype)
  File "/home/mazay/prg/regr/ve/lib/python3.9/site-packages/cupy/_creation/basic.py", line 22, in empty
    return cupy.ndarray(shape, dtype, order=order)
  File "cupy/core/core.pyx", line 141, in cupy.core.core.ndarray.__init__
TypeError: 'cupy.core.core.ndarray' object cannot be interpreted as an integer
```

The problem is probably in `svds()` code with `n_large` variable. Currently it is not converted from `cupy`, so `k - n_large` is also a `cupy` number. If `n_large` is less than `k`, then the issue is triggered in `_augmented_orthnormal_cols`.

Mazay0:
There is a similar issue #5001 , but I think it is another bug, with similar test conditions

smarkesini:
Partly fixed, with the remainder tracked elsewhere:

[...]

PR #5140, opened 2021-04-26
Fix TypeError in `svds`

toslunar:
Fix TypeError reported at #5000.  Because of inaccurate `eigsh`, the added tests fail at `numpy_cupy_allclose`.

### DIFF
--- a/cupyx/scipy/sparse/linalg/_eigen.py
+++ b/cupyx/scipy/sparse/linalg/_eigen.py
@@ -344,7 +344,7 @@ def svds(a, k=6, *, ncv=None, tol=0, which='LM', maxiter=None,
     cond = factor[t] * numpy.finfo(t).eps
     cutoff = cond * cupy.max(w)
     above_cutoff = (w > cutoff)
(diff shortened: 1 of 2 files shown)


kmaehashi:
pfnCI, test this please.

emcastillo:
Jenkins, test this please

cupyx/scipy/sparse/linalg/_eigen.py
```diff
@@ -344,7 +344,7 @@ def svds(a, k=6, *, ncv=None, tol=0, which='LM', maxiter=None,
     cond = factor[t] * numpy.finfo(t).eps
     cutoff = cond * cupy.max(w)
     above_cutoff = (w > cutoff)
-    n_large = above_cutoff.sum()
+    n_large = above_cutoff.sum().item()
```
emcastillo, line 347: This seems to be introducing device synchronization. But I guess there is no easy workaround sot this is acceptable.

emcastillo (approved):

PR #5146, opened 2021-04-27
cuDNN v8.2.0 for library installer

takagi:

PR #5148, opened 2021-04-27
cuDNN v8.2 on documentation

takagi:


kmaehashi:
Blocked by https://github.com/chainer/chainer-test/pull/653 and https://github.com/cupy/cupy/pull/5146

PR #5155, opened 2021-04-27
Add a new argument `out` to `cupy.asnumpy()`

leofang:
Follow-up of https://github.com/cupy/cupy/pull/4870#issuecomment-802501530. This is useful for reusing an existing NumPy array backed by pinned memory (#4870).


cupy/_core/core.pyx
```diff
@@ -2281,16 +2281,21 @@ cdef ndarray _send_object_to_gpu(obj, dtype, order, Py_ssize_t ndmin):
...
     stream = stream_module.get_current_stream()
-    cdef pinned_memory.PinnedMemoryPointer mem = (
-        _alloc_async_transfer_buffer(nbytes))
-    if mem is not None:
-        src_cpu = numpy.frombuffer(mem, a_dtype, a_cpu.size)
-        src_cpu[:] = a_cpu.ravel(order)
+    cdef pinned_memory.PinnedMemoryPointer mem
+    if isinstance(a_cpu.base, pinned_memory.PinnedMemoryPointer):
+        # NumPy arrays could be backed by pinned memory, in this case we
+        # don't need to monitor the copy using the watch list
```
kmaehashi, line 2287: I guess copy is needed anyway to avoid race?

```py
arr_cpu = cupyx.empty_pinned(...)
arr_gpu = cupy.array(arr_cpu) # trigger HtoD async memcpy
arr_cpu[:] = 0  # this may run during HtoD
```

leofang: Ahhh ok, thanks for catching this @kmaehashi...But the same is also true for `asnumpy()` even before this PR:
```python
arr_gpu = ...
with stream1:
    arr_cpu = cp.asnumpy(arr_gpu)  # D2H
with stream2:
    arr_gpu[:] = 0  # can run during D2H
```
Did we have a way to prevent this from happening? Looks hard to me.

leofang: Friendly ping @kmaehashi 🙂

kmaehashi: When using streams I think users are responsible for correctly handling data dependency (not limited to data copy but also computation).

leofang: Thanks, @kmaehashi. OK if we agree that when using streams users are responsible for managing data dependency, I think the same principle also applies to your example. The H2D copy via `cupy.array()` internally calls `copy_from_host_async` (which is `cudaMemcpyAsync`), but on the default stream it's a *synchronous* call, so without using a non-default stream the subsequent CPU operation on the array will be performed after the copy is finished.

kmaehashi: > (which is cudaMemcpyAsync), but on the default stream it's a synchronous call

Thanks, @leofang, do you have any references for this behavior?

The [docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g85073372f776b4c4d5f89f7124b7bf79) says:
> cudaMemcpyAsync() is asynchronous with respect to the host, so the call may return before the copy is complete

leofang: @kmaehashi Apparently I misread it, sorry! I was checking out P.23, P.28, and P.71 in this GTC talk:
https://on-demand.gputechconf.com/gtc/2014/presentations/S4158-cuda-streams-best-practices-common-pitfalls.pdf, where it kept emphasizing on using non-default streams, but the concern there was about concurrency, not about whether the call would be synchronous or not. I was further confused by a few other searches.

I checked that it is asynchronous even on the default stream:
```python
>>> import cupy as cp
>>> import numpy as np
>>> import cupyx
>>> 
>>> a = cp.empty((4*1024**3,), dtype='B')
>>> b = np.empty((4*1024**3,), dtype='B')
>>> c = cupyx.empty_like_pinned(b)
>>> s = cp.cuda.Stream()
>>> 
>>> print(cupyx.time.repeat(a.data.copy_from_host_async, (b.ctypes.data, 4*1024**3), n_repeat=10))
copy_from_host_async:    CPU:371055.705 us   +/-180.830 (min:370816.393 / max:371481.343) us     GPU-0:371144.601 us   +/-180.666 (min:370905.243 / max:371569.885) us
>>> print(cupyx.time.repeat(a.data.copy_from_host_async, (c.ctypes.data, 4*1024**3), n_repeat=10))
copy_from_host_async:    CPU:    2.685 us   +/- 0.525 (min:    2.260 / max:    4.054) us     GPU-0:352696.140 us   +/- 5.403 (min:352688.812 / max:352707.092) us
>>> print(cupyx.time.repeat(a.data.copy_from_host_async, (b.ctypes.data, 4*1024**3, s), n_repeat=10))
copy_from_host_async:    CPU:371298.881 us   +/-174.813 (min:371003.641 / max:371556.770) us     GPU-0:371388.605 us   +/-174.693 (min:371092.926 / max:371645.325) us
>>> print(cupyx.time.repeat(a.data.copy_from_host_async, (c.ctypes.data, 4*1024**3, s), n_repeat=10))
copy_from_host_async:    CPU:    3.404 us   +/- 0.748 (min:    2.800 / max:    5.371) us     GPU-0:352719.327 us   +/- 5.015 (min:352714.111 / max:352729.980) us
```
The control is returned to the host thread when pinned memory is in use, so the CPU time is significantly less.

OK you convinced me the race is possible and unavoidable.... So sounds like I should just touch `asnumpy()` and undo the changes made in `cupy/_core/core.pyx`?

kmaehashi: Thanks for testing with code! It's always important to read docs with suspicious eyes 😄 

> just touch asnumpy() and undo the changes made in cupy/_core/core.pyx?

Yeah, sounds good to me.

kmaehashi: Maybe it's better to add a comment so that future code readers can understand why copy is needed here 😃

leofang: Done in 9d4f6e0.

kmaehashi (approved):
LGTM!

kmaehashi:
pfnCI, test this please.

leofang:
Thanks, @kmaehashi. I merged with master to fix the RTD tests.

Jenkins, test this please

PR #5156, opened 2021-04-28
Fix PTDS destruction

### LINKED ISSUE #5145: Stream destructor must take care of PTDS
Destructor does not take care of PTDS. This is an edge case. Only users creating their own `Stream(ptds=True)` instead of using singleton (`Stream.ptds`) are affected.

https://github.com/cupy/cupy/blob/v9.0.0/cupy/cuda/stream.pyx#L345-L353

```py
>>> import cupy
>>> cupy.cuda.Stream.ptds = None
Exception ignored in: <bound method Stream.__del__ of <Stream 2>>
  File "cupy/cuda/stream.pyx", line 357, in cupy.cuda.stream.Stream.__del__
    class ExternalStream(BaseStream):
  File "cupy_backends/cuda/api/runtime.pyx", line 866, in cupy_backends.cuda.api.runtime.streamDestroy
    check_status(status)
  File "cupy_backends/cuda/api/runtime.pyx", line 261, in cupy_backends.cuda.api.runtime.check_status
```

emcastillo:
Closes #5145 

The actual stream tests show this error, but due to [Python ignoring exceptions in `__del__`](http://docs.python.org/3/reference/datamodel.html#object.__del__) methods they never failed when they should have done it :).

> Warning: Due to the precarious circumstances under which __del__() methods are invoked, exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead. Also, when __del__() is invoked in response to a module being deleted (e.g., when execution of the program is done), other globals referenced by the __del__() method may already have been deleted or in the process of being torn down (e.g. the import machinery shutting down). For this reason, __del__() methods should do the absolute minimum needed to maintain external invariants. Starting with version 1.5, Python guarantees that globals whose name begins with a single underscore are deleted from their module before other globals are deleted; if no other references to such globals exist, this may help in assuring that imported modules are still available at the time when the __del__() method is called.

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -339,7 +339,7 @@ class Stream(BaseStream):
         if is_shutting_down():
             return
         tls = _ThreadLocal.get()
-        if self.ptr:
+        if self.ptr and self.ptr != runtime.streamPerThread:
             current_ptr = <intptr_t>tls.get_current_stream_ptr()
             if <intptr_t>self.ptr == current_ptr:
                 tls.set_current_stream(self.null)


cupy/cuda/stream.pyx
```diff
@@ -339,7 +339,7 @@ class Stream(BaseStream):
         if is_shutting_down():
             return
         tls = _ThreadLocal.get()
-        if self.ptr:
+        if self.ptr and self.ptr != runtime.streamPerThread:
```
leofang, line 342: Just a nitpick: Can we also include `streamLegacy` here to be future-proof? See
https://github.com/cupy/cupy/blob/8cf5983112a4e90fe5e83b8daec5e55da472f400/cupy/cuda/stream.pyx#L322-L325
```suggestion
        if self.ptr not in (0, 1, 2):
```

jakirkham:
cc @pentschev

cupy/cuda/stream.pyx
```diff
@@ -339,7 +339,7 @@ class Stream(BaseStream):
...
             current_ptr = <intptr_t>tls.get_current_stream_ptr()
             if <intptr_t>self.ptr == current_ptr:
                 tls.set_current_stream(self.null)
```
kmaehashi, line 345: Stream must be reverted to PTDS when PTDS is enabled.

kmaehashi (changes requested):

kmaehashi:
Closing as @leofang's PR #5172 includes the fix for this issue  😄

Issue #5136, opened 2021-04-23
Support cuSPARSELt v0.1.0


leofang:
See https://github.com/conda-forge/cusparselt-feedstock/pull/9.

Changes:
- It's CUDA 11.2+ only
- Cross-platform (linux64, arm, windows)

leofang:
May be related to #5050?

takagi:
I `diff`ed v0.1.0 header against v0.0.1:
- `cusparseLtMatDescriptorDestroy` is added
- `CUSPARSE_COMPUTE_TF32` and `CUSPARSE_COMPUTE_TF32_FAST` are added to `cusparseComputeType `
- `cusparseMatmulDescriptorInit` is added
- cusparseLtSpMMAPrune2 is added
- cusparseLtSpMMAPruneCheck2 is added
- cusparseLtSpMMACompressedSize2 is added
- cusparseLtSpMMACompress2 is added

takagi:
Seems `cusparseMatmulDescriptorInit ` is doubly declared in v0.1.0 header...

leofang:
> Seems `cusparseLtMatmulDescriptorInit ` is doubly declared in v0.1.0 header...

cc: @mnicely

mnicely:
Thanks @leofang, we'll look into it

leofang:
Cool, glad to know @mnicely!

PR #5149, opened 2021-04-27
Cosmetic change in cuSPARSELt stub header

takagi:
Follows up #4883.
- Removed extra indent
- Untabify

Issue #4393, opened 2020-12-02
Relying on Cython to eliminate some C stubs that were needed for version compatibility


leofang:
As suggested by @takagi I am opening this issue for sharing the information. In #4338 I found that Cython is smart enough to eliminate dead code if there's enough information provided at compile time. For example, in https://github.com/cupy/cupy/pull/4338#discussion_r533071052 I showed that the two `if-not supported-then-raise` clauses there completely eliminated the need for adding stubs for RTD / HIP / CUDA 9.2 that do not support the newly added function. This could be a way to make the codebase a lot cleaner, since in addition to adding C stubs we also need to handle the wrappers in the Cython space anyway.

emcastillo:
This looks pretty neat!

emcastillo:
We feel that this might be a bit dangerous in case this behavior is not confirmed to be kept in future versions, so we might want to check cython docs and ask them directly

PR #5158, opened 2021-04-28
Support cuSPARSELt v0.1.0

takagi:
Close https://github.com/cupy/cupy/issues/5136. Merge https://github.com/cupy/cupy/pull/5149 first. 

Note that it's CUDA 11.2+ only. Rel https://github.com/cupy/cupy/pull/5050.

I've locally confirmed that it successfully builds and cusparselt example runs.

### DIFF
--- a/cupy_backends/cuda/cupy_cusparselt.h
+++ b/cupy_backends/cuda/cupy_cusparselt.h
@@ -1,0 +1,32 @@
+#ifndef INCLUDE_GUARD_CUDA_CUPY_CUSPARSELT_H
+#define INCLUDE_GUARD_CUDA_CUPY_CUSPARSELT_H
+
+#include <library_types.h>
+#include <cusparseLt.h>
+
+#if CUSPARSELT_VERSION < 100
+// Added in cuSPARSELt 0.1.0
+
+cusparseStatus_t cusparseLtMatDescriptorDestroy(...) {
+    return CUSPARSE_STATUS_SUCCESS;
+}
+
+cusparseStatus_t cusparseLtSpMMAPrune2(...) {
+    return CUSPARSE_STATUS_SUCCESS;
+}
+
+cusparseStatus_t cusparseLtSpMMAPruneCheck2(...) {
+    return CUSPARSE_STATUS_SUCCESS;
+}
+
+cusparseStatus_t cusparseLtSpMMACompressedSize2(...) {
+    return CUSPARSE_STATUS_SUCCESS;
+}
+
+cusparseStatus_t cusparseLtSpMMACompress2(...) {
+    return CUSPARSE_STATUS_SUCCESS;
+}
+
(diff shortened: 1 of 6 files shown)


takagi:
Should we also keep successful build with v0.0.1?

kmaehashi:
> Should we also keep successful build with v0.0.1?

Yes, as we have to backport this to v9.

takagi:
`cusparsLT.h` are exactly the same in Linux and Windows packages so this change should also work on Windows.

cupy_backends/cuda/cupy_cusparselt.h
```diff
@@ -0,0 +1,32 @@
...
+#define INCLUDE_GUARD_CUDA_CUPY_CUSPARSELT_H
+
+#include <library_types.h>
+#include <cusparseLt.h>
+
+#if CUSPARSELT_VERSION < 100
+// Added in cuSPARSELt 0.1.0
+
+cusparseStatus_t cusparseLtMatDescriptorDestroy(...) {
+    return CUSPARSE_STATUS_SUCCESS;
```
leofang, line 11: This file is not for RTD/HIP stubs. I feel it may make better sense to return `CUSPARSE_STATUS_NOT_SUPPORTED` instead.

cupy_backends/cuda/libs/cusparselt.pyx
```diff
@@ -239,6 +260,12 @@ cpdef structuredDescriptorInit(Handle handle, MatDescriptor matDescr,
         <cusparseLtSparsity_t> sparsity)
     check_status(status)
 
+cpdef matDescriptorDestroy(MatDescriptor matDescr):
+    """Releases the resources used by an instance of a matrix descriptor."""
```
leofang, line 264: If cuSPARSELt v0.0.1 is to be supported, we may want to add a guard in the newly added wrappers in this file:
```python
    if CUSPARSELT_VERSION < 100:
        raise RuntimeError(...)
```
See https://github.com/cupy/cupy/issues/4393.

examples/cusparselt/matmul.py
```diff
@@ -92,5 +92,8 @@
 #
 # destroys plan and handle
 #
+cusparselt.matDescriptorDestroy(matA)
+cusparselt.matDescriptorDestroy(matB)
+cusparselt.matDescriptorDestroy(matC)
```
leofang, line 97: These would not work if we return `CUSPARSE_STATUS_NOT_SUPPORTED` for cuSPARSELt v0.0.1, so this may need to change depending on the final decision.

takagi:
Thanks @leofang!

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

kmaehashi:
pfnCI, test this please.

PR #5164, opened 2021-04-28
Fix for updated FlexCI base image

kmaehashi:
FlexCI base image no longer contains CUDA components other than `cuda-driver`.


kmaehashi:
`pfn-public-ci/cupy.py3.amd` passed.

kmaehashi:
Let me merge this as this blocks all other PRs.

Issue #5120, opened 2021-04-21
[RFC] [v10] Dropping Support for CUDA 9.2 & NCCL 2.4


kmaehashi:
In CuPy v10, we are thinking of dropping support for CUDA 9.2 and NCCL 2.4. Please leave a comment if you have any concerns.

This does not affect CuPy v9.x releases.

PR #5169, opened 2021-05-01
Add `cupyx.jit.atomic_add`

asi1024:
Part of https://github.com/cupy/cupy/issues/4290.

Example:
```py
@jit.rawkernel()
def reduction(x, y, size):
    tid = jit.blockIdx.x * jit.blockDim.x + jit.threadIdx.x
    ntid = jit.blockDim.x * jit.gridDim.x

    value = cupy.float32(0)
    for i in range(tid, size, ntid):
        value += x[i]

    smem = jit.shared_memory(cupy.float32, 1024)
    smem[jit.threadIdx.x] = value

    jit.syncthreads()

    if jit.threadIdx.x == cupy.uint32(0):
        value = cupy.float32(0)
        for i in range(jit.blockDim.x):
            value += smem[i]
        jit.atomic_add(y, 0, value)
```

Generated CUDA code:
```cpp
extern "C" __global__ void reduction(CArray<float, 1, true, true> x, CArray<float, 1, true, true> y, unsigned int size) {
  unsigned int tid;
  unsigned int ntid;
  float value;
  unsigned int i;
  __shared__ float _smem1[1024];
  float* smem;
  tid = ((blockIdx.x * blockDim.x) + threadIdx.x);
  ntid = (blockDim.x * gridDim.x);
  value = 0.0f;
  for (unsigned int __it = tid, __stop = size, __step = ntid; __it < __stop; __it += __step) {
    i = __it;
    value = (value + x[i]);
  }
  smem = _smem1;
  smem[threadIdx.x] = value;
  __syncthreads();
  if ((threadIdx.x == 0u)) {
    value = 0.0f;
    for (unsigned int __it = 0, __stop = blockDim.x, __step = 1; __step >= 0 ? __it < __stop : __it > __stop; __it += __step) {
      i = __it;
      value = (value + smem[i]);
    }
    atomicAdd(&y[0], value);
  }
  else {
  }
}
```

### DIFF
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -111,4 +111,6 @@ builtin_functions_dict = {
 
 syncthreads = SyncThreads()
 shared_memory = SharedMemory()
+
+# TODO: Add more atomic functions.
 atomic_add = AtomicOp('Add', 'iILefd')
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -113,4 +113,4 @@ syncthreads = SyncThreads()
 shared_memory = SharedMemory()
 
 # TODO: Add more atomic functions.
-atomic_add = AtomicOp('Add', 'iILefd')
+atomic_add = AtomicOp('Add', 'iILQefd')
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -1,3 +1,4 @@
+from cupy_backends.cuda.api import runtime
 from cupyx.jit import _cuda_types
 from cupyx.jit._internal_types import BuiltinFunc
 from cupyx.jit._internal_types import Data
@@ -102,6 +103,9 @@ class AtomicOp(BuiltinFunc):
         value = Data.init(value, env)
         if ctype.dtype.char not in self._dtypes:
             raise TypeError(f'`{name}` does not support {ctype.dtype} input.')
+        if ctype.dtype.char == 'e' and runtime.runtimeGetVersion() < 10000:
+            raise RuntimeError(
+                'float16 atomic operation is not supported this CUDA version.')
         return Data(f'{name}(&{target.code}, {value.code})', ctype)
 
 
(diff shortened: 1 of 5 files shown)


cupyx/jit/_builtin_funcs.py
```diff
@@ -82,9 +83,34 @@ def call_const(self, env, dtype, size):
...
+
 builtin_functions_dict = {
     range: RangeFunc(),
 }
 
 syncthreads = SyncThreads()
 shared_memory = SharedMemory()
+
+# TODO: Add more atomic functions.
+atomic_add = AtomicOp('Add', 'iILefd')
```
leofang, line 116: I think this should be fixed?
```suggestion
atomic_add = AtomicOp('Add', 'iqQefd')
```
Notes:
1. The supported atomics are
```C
int atomicAdd(int* address, int val);
unsigned int atomicAdd(unsigned int* address,
                       unsigned int val);
unsigned long long int atomicAdd(unsigned long long int* address,
                                 unsigned long long int val);
float atomicAdd(float* address, float val);
double atomicAdd(double* address, double val);
__half atomicAdd(__half *address, __half val);
```
and their bit-widths aren't affected by LP64 (linux64) vs LLP64 (win), so we should prefer `q` over `l`.

2. Just a note: by adding `'e'` (for half) here we are dropping the support for CUDA 9.2 (https://github.com/cupy/cupy/issues/5120), as this version was added since 10.0. But I think we should do it.

leofang (changes requested):
I am not sure if separating out the `_indexing` helper function is needed, which is the biggest change in this PR. Just left a quick observation.

cupyx/jit/_builtin_funcs.py
asi1024: @leofang Thanks for your review!

1. This list of dtypes is used only to check if `atomicAdd` supports the data type of input, and is NOT used to determine the dtype of output. So, I just added `Q` dtype to the list. (973d6e7)
2. I want to backport this PR to v9, so I fixed to raise `RuntimeError` for float16 input for CUDA 9.2. (a419bd5)

leofang: 1. Yes, but in this case input and output happen to have the same data type 🙂 btw, you're right to not add `q`, my mistake!
2. Sounds good!

leofang (approved):

leofang:
Jenkins, test this please

asi1024:
I will fix the Windows CI failure in #5173.

emcastillo:
Jenkins, test this please

asi1024:
Jenkins, test this please.

emcastillo (approved):

PR #5171, opened 2021-05-03
[WIP] Texture memory 2D/3D affine transformations

the-lay:
I would like to propose a new set of methods for affine transformations that make use of the texture memory. [Texture memory](https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#texture-memory) is cached, optimized for spatial locality and can use hardware filtering with `texND()`. There are three drawbacks: texture memory is read-only, supports only up to 3D and the filtering has relatively low-precision [(9-bit fixed point format with 8 bits of fractional value)](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#linear-filtering).

The closest thing to cupy-powered affine transformations right now is in `cupyx.scipy.ndimage`. More specifically, #4402 ported scipy's interpolation methods, allowing to use `map_coordinates`, `rotate`, `affine_transformation` and others. It is very useful, straightfoward and conveniently compatible with scipy. However, affine transformations with texture memory are faster (counting in texture upload) with much better performance scaling (especially noticeable in 3D) and precise enough for many use cases.

Made possible by @leofang 's texture memory PRs in #2432 and #4433.

## Benchmark
I have prepared a benchmark (runnable on the current version of cupy/cupy) to compare `cupyx.scipy.ndimage.affine_transform` and affine transformation with texture memory in 2D and 3D. Both code and raw results are under the spoilers. Benchmark was ran on a dated laptop with NVIDIA 1050Ti. Each operation is repeated 100 times with random rotation matrices and not in succession, to minimize caching effects. The results are avg, min and max time of 100 operations, in milliseconds.

<details>
  <summary>2D benchmark code</summary>

  ```python
import cupy as cp
from cupyx.scipy import ndimage
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt


def random_2d_rotation_matrix(rotation_center) -> np.ndarray:
    theta = np.radians(np.random.randint(0, 180))
    c, s = np.cos(theta), np.sin(theta)
    m = np.array([
        [1, 0, rotation_center[0]],
        [0, 1, rotation_center[1]],
        [0, 0, 1]
    ], np.float32)
    m = np.dot(m, np.array([
        [c, -s, 0],
        [s, c, 0],
        [0, 0, 1]
    ], np.float32))
    m = np.dot(m, np.array([
        [1, 0, -rotation_center[0]],
        [0, 1, -rotation_center[1]],
        [0, 0, 1]
    ], np.float32))
    return m


def create_texture(data: cp.ndarray):
    ch = cp.cuda.texture.ChannelFormatDescriptor(32, 0, 0, 0, cp.cuda.runtime.cudaChannelFormatKindFloat)
    arr = cp.cuda.texture.CUDAarray(ch, *data.shape[::-1])  # CUDAArray: last dimension=fastest changing dimension
    res = cp.cuda.texture.ResourceDescriptor(cp.cuda.runtime.cudaResourceTypeArray, cuArr=arr)
    tex = cp.cuda.texture.TextureDescriptor((cp.cuda.runtime.cudaAddressModeBorder, ) * data.ndim,
                                            cp.cuda.runtime.cudaFilterModeLinear,
                                            cp.cuda.runtime.cudaReadModeElementType)
    tex_obj = cp.cuda.texture.TextureObject(res, tex)
    arr.copy_from(data)
    return tex_obj


affine_transform_2d = cp.ElementwiseKernel(
    'U texObj, raw float32 m, uint64 width',
    'float32 transformed_image',
    '''
    float3 pixel = make_float3(
        (float)(i / width) + .5f,
        (float)(i % width) + .5f,
        1.0f
    );

    float x = dot(pixel, make_float3(m[0],  m[1],  m[2]));
    float y = dot(pixel, make_float3(m[3],  m[4],  m[5]));

    transformed_image = tex2D<float>(texObj, y, x);
    ''',
    'affine_transformation_2d',
    preamble='''
    inline __host__ __device__ float dot(float3 a, float3 b)
    {
        return a.x * b.x + a.y * b.y + a.z * b.z;
    }
    ''')

repeats = 100
table_2d = pd.DataFrame(columns=['Dims', 'Method', 'Avg time', 'Min time', 'Max time'])
dims_2d = [(5, 5), (10, 10), (15, 15), (30, 30), (50, 50), (100, 100), (300, 300),
           (500, 500), (1000, 1000), (2000, 2000), (4000, 4000), (6000, 6000)]
for dim_i, dims in enumerate(dims_2d):
    image = cp.arange(0, np.prod(dims), dtype=cp.float32).reshape(dims)
    ms = [cp.array(random_2d_rotation_matrix(np.divide(dims, 2))) for _ in range(repeats)]

    timings = {
        'ndimage0': [],
        'ndimage3': [],
        'ndimage5': [],
        'proposed': []
    }
    for m in ms:

        # cupyx.scipy.ndimage.affine_transform(order=0)
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        ref = ndimage.affine_transform(image, m, order=0)
        t_end = stream.record()
        t_end.synchronize()
        timings['ndimage0'].append(cp.cuda.get_elapsed_time(t_start, t_end))

        # cupyx.scipy.ndimage.affine_transform(order=3)
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        res1 = ndimage.affine_transform(image, m, order=3)
        t_end = stream.record()
        t_end.synchronize()
        timings['ndimage3'].append(cp.cuda.get_elapsed_time(t_start, t_end))

        # cupyx.scipy.ndimage.affine_transform(order=5)
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        res2 = ndimage.affine_transform(image, m, order=5)
        t_end = stream.record()
        t_end.synchronize()
        timings['ndimage5'].append(cp.cuda.get_elapsed_time(t_start, t_end))

        # proposed kernel
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        texture = create_texture(image)
        res3 = affine_transform_2d(texture, m, image.shape[1], size=image.size).reshape(image.shape)
        t_end = stream.record()
        t_end.synchronize()
        timings['proposed'].append(cp.cuda.get_elapsed_time(t_start, t_end))

    table_2d.loc[dim_i * 4] = [dims, 'ndimage.affine_transform(order=0)', sum(timings['ndimage0']) / repeats,
                               min(timings['ndimage0']), max(timings['ndimage0'])]
    table_2d.loc[dim_i * 4 + 1] = [dims, 'ndimage.affine_transform(order=3)', sum(timings['ndimage3']) / repeats,
                               min(timings['ndimage3']), max(timings['ndimage3'])]
    table_2d.loc[dim_i * 4 + 2] = [dims, 'ndimage.affine_transform(order=5)', sum(timings['ndimage5']) / repeats,
                               min(timings['ndimage5']), max(timings['ndimage5'])]
    table_2d.loc[dim_i * 4 + 3] = [dims, 'affine_transform_2d', sum(timings['proposed']) / repeats,
                               min(timings['proposed']), max(timings['proposed'])]

# Visualize rotation
fig, ax = plt.subplots(1, 4, sharex=True, sharey=True)
ax[0].set_title('ndimage.affine_transform(order=0)')
ax[0].imshow(ref.get())
ax[1].set_title('ndimage.affine_transform(order=3)')
ax[1].imshow(res1.get())
ax[2].set_title('ndimage.affine_transform(order=5)')
ax[2].imshow(res2.get())
ax[3].set_title('Proposed PR')
ax[3].imshow(res3.get())
plt.show()

# Print benchmark timings
print('2D benchmark results')
with pd.option_context('display.max_rows', None, 'display.max_columns', None, 'display.expand_frame_repr', False):
    print(table_2d)

  ```

</details>

<details>
  <summary>2D benchmark results</summary>

  2d_6k_example

  2d_6k_zoom_staircase_example
  <pre>
	    Dims                             Method    Avg time    Min time     Max time
0         (5, 5)  ndimage.affine_transform(order=0)    0.386053    0.262656     4.366336
1         (5, 5)  ndimage.affine_transform(order=3)    0.908673    0.651264     6.031616
2         (5, 5)  ndimage.affine_transform(order=5)    0.845332    0.652288     4.664320
3         (5, 5)                affine_transform_2d    0.142651    0.094240     1.021696
4       (10, 10)  ndimage.affine_transform(order=0)    0.342074    0.268128     1.204576
5       (10, 10)  ndimage.affine_transform(order=3)    0.834139    0.662528     1.795168
6       (10, 10)  ndimage.affine_transform(order=5)    0.834414    0.659456     2.234368
7       (10, 10)                affine_transform_2d    0.141074    0.095264     0.650464
8       (15, 15)  ndimage.affine_transform(order=0)    0.332913    0.265216     1.517376
9       (15, 15)  ndimage.affine_transform(order=3)    0.868016    0.653408     2.266464
10      (15, 15)  ndimage.affine_transform(order=5)    0.856418    0.655168     2.100000
11      (15, 15)                affine_transform_2d    0.123574    0.095008     0.240800
12      (30, 30)  ndimage.affine_transform(order=0)    0.335848    0.265248     1.559808
13      (30, 30)  ndimage.affine_transform(order=3)    0.861030    0.653888     3.524512
14      (30, 30)  ndimage.affine_transform(order=5)    0.834707    0.660032     2.796320
15      (30, 30)                affine_transform_2d    0.131306    0.095136     0.591584
16      (50, 50)  ndimage.affine_transform(order=0)    0.339147    0.265632     1.231712
17      (50, 50)  ndimage.affine_transform(order=3)    0.876927    0.658432     2.828928
18      (50, 50)  ndimage.affine_transform(order=5)    0.868408    0.668672     2.513472
19      (50, 50)                affine_transform_2d    0.133409    0.095072     0.473088
20    (100, 100)  ndimage.affine_transform(order=0)    0.356329    0.266272     2.349056
21    (100, 100)  ndimage.affine_transform(order=3)    0.910992    0.700416     2.535520
22    (100, 100)  ndimage.affine_transform(order=5)    0.905464    0.740288     3.478368
23    (100, 100)                affine_transform_2d    0.129393    0.095712     0.386816
24    (300, 300)  ndimage.affine_transform(order=0)    0.422491    0.318400     2.259968
25    (300, 300)  ndimage.affine_transform(order=3)    1.321957    1.051488     4.044448
26    (300, 300)  ndimage.affine_transform(order=5)    1.756830    1.483776     4.703232
27    (300, 300)                affine_transform_2d    0.172797    0.114688     0.415744
28    (500, 500)  ndimage.affine_transform(order=0)    0.530057    0.396288     1.527456
29    (500, 500)  ndimage.affine_transform(order=3)    2.034564    1.809408     3.554304
30    (500, 500)  ndimage.affine_transform(order=5)    3.357076    3.050496     4.954944
31    (500, 500)                affine_transform_2d    0.248367    0.152032     0.895616
32  (1000, 1000)  ndimage.affine_transform(order=0)    0.971819    0.832064     2.316000
33  (1000, 1000)  ndimage.affine_transform(order=3)    5.946925    5.627904     7.003136
34  (1000, 1000)  ndimage.affine_transform(order=5)   10.888610   10.208256    12.568352
35  (1000, 1000)                affine_transform_2d    0.817483    0.592768     7.205632
36  (2000, 2000)  ndimage.affine_transform(order=0)    2.596743    2.447104     3.602432
37  (2000, 2000)  ndimage.affine_transform(order=3)   22.038571   20.799295    24.351744
38  (2000, 2000)  ndimage.affine_transform(order=5)   41.882197   38.439713    51.206783
39  (2000, 2000)                affine_transform_2d    1.657099    1.481664     2.630656
40  (4000, 4000)  ndimage.affine_transform(order=0)   12.121363    9.301472   269.421234
41  (4000, 4000)  ndimage.affine_transform(order=3)  105.667198   92.500832  1028.231689
42  (4000, 4000)  ndimage.affine_transform(order=5)  201.624636  170.276413  1174.135742
43  (4000, 4000)                affine_transform_2d    5.292619    4.753408     6.666240
44  (6000, 6000)  ndimage.affine_transform(order=0)   28.691127   20.913889   280.324097
45  (6000, 6000)  ndimage.affine_transform(order=3)  296.954384  247.346176  1162.645508
46  (6000, 6000)  ndimage.affine_transform(order=5)  578.262190  462.934967  1489.004517
47  (6000, 6000)                affine_transform_2d   12.377960   10.293248    19.489792
  </pre>
</details>

<details>
  <summary>3D benchmark code</summary>

  ```python
import cupy as cp
from cupyx.scipy import ndimage
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt


def random_3d_rotation_matrix(rotation_center) -> np.ndarray:
    theta = np.radians(np.random.randint(0, 180))
    c, s = np.cos(theta), np.sin(theta)
    m = np.array([
        [1, 0, 0, rotation_center[0]],
        [0, 1, 0, rotation_center[1]],
        [0, 0, 1, rotation_center[2]],
        [0, 0, 0, 1]
    ], np.float32)
    m = np.dot(m, np.array([
        [1, 0, 0, 0],
        [0, c, -s, 0],
        [0, s, c, 0],
        [0, 0, 0, 1]
    ], np.float32))
    m = np.dot(m, np.array([
        [1, 0, 0, -rotation_center[0]],
        [0, 1, 0, -rotation_center[1]],
        [0, 0, 1, -rotation_center[2]],
        [0, 0, 0, 1]
    ], np.float32))
    return m


def create_texture(data: cp.ndarray):
    ch = cp.cuda.texture.ChannelFormatDescriptor(32, 0, 0, 0, cp.cuda.runtime.cudaChannelFormatKindFloat)
    arr = cp.cuda.texture.CUDAarray(ch, *data.shape[::-1])  # CUDAArray: last dimension=fastest changing dimension
    res = cp.cuda.texture.ResourceDescriptor(cp.cuda.runtime.cudaResourceTypeArray, cuArr=arr)
    tex = cp.cuda.texture.TextureDescriptor((cp.cuda.runtime.cudaAddressModeBorder, ) * data.ndim,
                                            cp.cuda.runtime.cudaFilterModeLinear,
                                            cp.cuda.runtime.cudaReadModeElementType)
    tex_obj = cp.cuda.texture.TextureObject(res, tex)
    arr.copy_from(data)
    return tex_obj


affine_transform_3d = cp.ElementwiseKernel(
    'U texObj, raw float32 m, uint64 height, uint64 width',
    'float32 transformed_volume',
    '''
    float4 voxel = make_float4(
        (float)(i / (width * height)) + .5f,
        (float)((i % (width * height)) / width) + .5f,
        (float)((i % (width * height)) % width) + .5f,
        1.0f
    );

    float x = dot(voxel, make_float4(m[0],  m[1],  m[2],  m[3]));
    float y = dot(voxel, make_float4(m[4],  m[5],  m[6],  m[7]));
    float z = dot(voxel, make_float4(m[8],  m[9],  m[10], m[11]));

    transformed_volume = tex3D<float>(texObj, z, y, x);
    ''',
    'affine_transformation_3d',
    preamble='''
    inline __host__ __device__ float dot(float4 a, float4 b)
    {
        return a.x * b.x + a.y * b.y + a.z * b.z + a.w * b.w;
    }
    ''')

repeats = 100
table_3d = pd.DataFrame(columns=['Dims', 'Method', 'Avg time', 'Min time', 'Max time'])
dims_3d = [(5, 5, 5), (10, 10, 10), (15, 15, 15), (30, 30, 30), (50, 50, 50), (100, 100, 100), (300, 300, 300)]
for dim_i, dims in enumerate(dims_3d):
    image = cp.arange(0, np.prod(dims), dtype=cp.float32).reshape(dims)
    ms = [cp.array(random_3d_rotation_matrix(np.divide(dims, 2))) for _ in range(repeats)]

    timings = {
        'ndimage0': [],
        'ndimage3': [],
        'ndimage5': [],
        'proposed': []
    }
    for m in ms:

        # cupyx.scipy.ndimage.affine_transform(order=0)
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        ref = ndimage.affine_transform(image, m, order=0)
        t_end = stream.record()
        t_end.synchronize()
        timings['ndimage0'].append(cp.cuda.get_elapsed_time(t_start, t_end))

        # cupyx.scipy.ndimage.affine_transform(order=3)
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        res1 = ndimage.affine_transform(image, m, order=3)
        t_end = stream.record()
        t_end.synchronize()
        timings['ndimage3'].append(cp.cuda.get_elapsed_time(t_start, t_end))

        # cupyx.scipy.ndimage.affine_transform(order=5)
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        res2 = ndimage.affine_transform(image, m, order=5)
        t_end = stream.record()
        t_end.synchronize()
        timings['ndimage5'].append(cp.cuda.get_elapsed_time(t_start, t_end))

        # proposed kernel
        stream = cp.cuda.Stream.null
        t_start = stream.record()
        texture = create_texture(image)
        res3 = affine_transform_3d(texture, m, image.shape[1], image.shape[2], size=image.size).reshape(image.shape)
        t_end = stream.record()
        t_end.synchronize()
        timings['proposed'].append(cp.cuda.get_elapsed_time(t_start, t_end))

    table_3d.loc[dim_i * 4] = [dims, 'ndimage.affine_transform(order=0)', sum(timings['ndimage0']) / repeats,
                               min(timings['ndimage0']), max(timings['ndimage0'])]
    table_3d.loc[dim_i * 4 + 1] = [dims, 'ndimage.affine_transform(order=3)', sum(timings['ndimage3']) / repeats,
                               min(timings['ndimage3']), max(timings['ndimage3'])]
    table_3d.loc[dim_i * 4 + 2] = [dims, 'ndimage.affine_transform(order=5)', sum(timings['ndimage5']) / repeats,
                               min(timings['ndimage5']), max(timings['ndimage5'])]
    table_3d.loc[dim_i * 4 + 3] = [dims, 'affine_transform_3d', sum(timings['proposed']) / repeats,
                               min(timings['proposed']), max(timings['proposed'])]

# Visualize rotation
fig, ax = plt.subplots(1, 4, sharex=True, sharey=True)
vol_slice = dims_3d[-1][0] // 2
ax[0].set_title('ndimage.affine_transform(order=0)')
ax[0].imshow(ref[vol_slice].get())
ax[1].set_title('ndimage.affine_transform(order=3)')
ax[1].imshow(res1[vol_slice].get())
ax[2].set_title('ndimage.affine_transform(order=5)')
ax[2].imshow(res2[vol_slice].get())
ax[3].set_title('Proposed PR')
ax[3].imshow(res3[vol_slice].get())
plt.show()

# Print benchmark timings
print('3D benchmark results')
with pd.option_context('display.max_rows', None, 'display.max_columns', None, 'display.expand_frame_repr', False):
    print(table_3d)

  ```

</details>

<details>
  <summary>3D benchmark results</summary>

  3d_300_example
  
  3d_300_zoom_staircase_example
  <pre>
	       Dims                             Method     Avg time    Min time     Max time
0         (5, 5, 5)  ndimage.affine_transform(order=0)     0.398012    0.271552     4.236064
1         (5, 5, 5)  ndimage.affine_transform(order=3)     1.165629    0.851648     6.432576
2         (5, 5, 5)  ndimage.affine_transform(order=5)     1.191894    0.868352     6.321952
3         (5, 5, 5)                affine_transform_3d     0.166360    0.101536     1.001440
4      (10, 10, 10)  ndimage.affine_transform(order=0)     0.371300    0.269504     1.397760
5      (10, 10, 10)  ndimage.affine_transform(order=3)     1.167064    0.864896     2.814368
6      (10, 10, 10)  ndimage.affine_transform(order=5)     1.167654    0.908288     3.306048
7      (10, 10, 10)                affine_transform_3d     0.147276    0.101376     0.399360
8      (15, 15, 15)  ndimage.affine_transform(order=0)     0.359232    0.268256     1.310304
9      (15, 15, 15)  ndimage.affine_transform(order=3)     1.149133    0.852864     3.437472
10     (15, 15, 15)  ndimage.affine_transform(order=5)     1.291052    0.950272     3.544064
11     (15, 15, 15)                affine_transform_3d     0.165462    0.100352     0.761728
12     (30, 30, 30)  ndimage.affine_transform(order=0)     0.391748    0.296960     1.280000
13     (30, 30, 30)  ndimage.affine_transform(order=3)     1.435411    1.176320     2.672640
14     (30, 30, 30)  ndimage.affine_transform(order=5)     2.202434    1.784416     5.009408
15     (30, 30, 30)                affine_transform_3d     0.188001    0.113632     0.898048
16     (50, 50, 50)  ndimage.affine_transform(order=0)     0.521921    0.379616     1.303040
17     (50, 50, 50)  ndimage.affine_transform(order=3)     2.509606    2.159264     4.907008
18     (50, 50, 50)  ndimage.affine_transform(order=5)     5.333314    4.808384     7.107584
19     (50, 50, 50)                affine_transform_3d     0.298158    0.159744     1.087488
20  (100, 100, 100)  ndimage.affine_transform(order=0)     1.392856    1.215488     2.672576
21  (100, 100, 100)  ndimage.affine_transform(order=3)    13.808296   13.070336    16.248833
22  (100, 100, 100)  ndimage.affine_transform(order=5)    37.516795   34.871010    44.173313
23  (100, 100, 100)                affine_transform_3d     1.023004    0.796672     2.108416
24  (300, 300, 300)  ndimage.affine_transform(order=0)    27.509351   25.227264    44.711937
25  (300, 300, 300)  ndimage.affine_transform(order=3)   356.705551  317.809601   551.584778
26  (300, 300, 300)  ndimage.affine_transform(order=5)  1025.695103  864.895996  1683.631836
27  (300, 300, 300)                affine_transform_3d    13.514331   12.266336    20.514721
  </pre>
</details>

In short, here are the graphs. Please note, the y axis is in log scale.

2d_benchmark_graph

3d_benchmark_graph

## TODO

- [x] Tests, including different data types
- ~~Add shortcut functions for affine transformations similar to how scipy does it, i.e. `rotate()`, `zoom()`/`scale()`, `shift()`/`translate()`~~
- [x] Add support to use directly `cupy.cuda.texture.TextureObject` (including linear memory texture object)
- [x] Address modes as a keyword
- [x] Add reshape functionality for transformed arrays
- ~~Textures support up to 4 channels (see `ChannelFormatDescriptor`), so the new methods should support that too~~

The feature is not finished, but **I wanted to open PR early to get feedback.**
Please let me know what you think and if any changes are necessary, thanks!


leofang:
cc: @grlee77 @coderforlife

grlee77:
This looks useful to me and the code that is here so far looks good. Do you think you will be able to compare to `scipy.ndimage` results for the tests? I think especially for iterative registration methods, having a fast linear interpolation will be very useful during the optimization process.

Do you plan to add `address_mode` as a kwarg to the top level `affine_transformation` function? I see that it is already supported in the kernel generation codes, but the value is hard-coded to `border` within the `affine_transformation` function.

jakirkham:
cc @not522

the-lay:
@grlee77 thank you for the feedback!

> Do you think you will be able to compare to `scipy.ndimage` results for the tests?
There will be difference at the borders due to the anti-aliasing (see the figures under the results spoilers), but it should be possible with `cupy.allclose` and some tolerance, for example. I will think whether there is a better way.

> I think especially for iterative registration methods, having a fast linear interpolation will be very useful during the optimization process.
To support iterative methods even further, the methods can also accept a texture object directly to minimize data transfer. That is something we do for template matching (i.e. rotating a single static object to many different orientations) in our group and the difference is noticeable. I will add this too in the next commit.

> Do you plan to add `address_mode` as a kwarg to the top level `affine_transformation` function? I see that it is already supported in the kernel generation codes, but the value is hard-coded to `border` within the `affine_transformation` function.
Yes! Forgot to add this to TODO, thank you.

cupyx/texture/_affine.py
```diff
@@ -0,0 +1,159 @@
...
+        read_mode = runtime.cudaReadModeElementType
+    elif read_mode == 'normalized_float':
+        read_mode = runtime.cudaReadModeNormalizedFloat
+    else:
+        raise ValueError(
+            'Unsupported read mode '
+            '(supported: element_type, normalized_float)')
+
+    texture_fmt = texture.ChannelFormatDescriptor(
+        data.itemsize * 8, 0, 0, 0, fmt_kind)
```
leofang, line 52: Is there a way for us to explore the multi-channel support? Just curious.

cupyx/texture/_affine.py
```diff
@@ -0,0 +1,159 @@
...
+        output = cupy.empty_like(data, data.dtype)
+    else:
+        # TODO(the-lay): check whether input data array fits into output
+        pass
+
+    kernel = _get_affine_transform_kernel(data.ndim)
+
+    kernel(texture_object, transformation_matrix, *data.shape[1:], output)
+
+    del texture_object
```
leofang, line 158: This is not needed.
```suggestion
```

cupyx/texture/_affine.py
```diff
@@ -0,0 +1,159 @@
...
+            preamble='''
+            inline __host__ __device__ float dot(float4 a, float4 b)
+            {
+                return a.x * b.x + a.y * b.y + a.z * b.z + a.w * b.w;
+            }
+            ''')
+    else:
+        raise ValueError(
+            'Texture memory affine transformation is defined '
+            'only for single channel 2D and 3D arrays')
```
leofang, line 121: Forgive my ignorance of `ndimage` APIs, I am just wondering if any of the `ndimage` APIs supports batch processing (say transforming a stack of images at once). Looks to me the current implementation only supports single-image processing.

cupyx/texture/_affine.py
```diff
@@ -0,0 +1,159 @@
...
+        runtime.cudaResourceTypeArray, cuArr=array)
+    tex_desc = texture.TextureDescriptor(
+        (address_mode, ) * data.ndim, filter_mode, read_mode)
+    tex_obj = texture.TextureObject(res_desc, tex_desc)
+    array.copy_from(data)
+
+    return tex_obj
+
+
+@cupy._util.memoize(for_each_device=True)
```
leofang, line 66: I don't think this is needed as it's done internally in `ElementwiseKernel`.
```suggestion
```

leofang: btw It'd be good to move the two `ElementwiseKernel` to the module level. They do not depend on the input parameters so there's no need to construct new kernel objects on the fly.

leofang (commented):
This is very cool, @the-lay! This use case perfectly justifies why texture memory is exposed in CuPy. I am very happy to see someone building up nice code upon my prior work!

I would like to have a bigger picture discussion. Does it make sense to split the `ndimage` code path into two, without (the existing one) and with texture (this PR), and allow users to choose based on a simple switch or a context manager without modifying code at all? For example, is this possible? 
```python
cupyx.scipy.ndimage.config.use_texture()  # <-- a switch
out = cupyx.scipy.ndimage.affine_transformation(...)  # <-- and the call uses this PR

# alternative way
with cupyx.scipy.ndimage.config.use_texture():
    out = cupyx.scipy.ndimage.affine_transformation(...)  # <-- and the call uses this PR
```
If not possible, what would be the obstacle(s)? (The `use_texture()` API is just for illustration purpose; I'm sure we can find a better name if we decide to do it.)

cupyx/texture/_affine.py
grlee77: It looks like that is already in the TODO list up top

cupyx/texture/_affine.py
grlee77: The SciPy ndimage APIs don't support batch processing (or vector-valued pixels). One has to call the function separately on each channel of a color image, for example. 

This does result in some inefficiency due to redundant computation since the interpolation weights should be the same for all channels!  I would be interested in having better support for `float3` (e.g. RGB images) and `float4` (r.g. RGBA) cases at a minimum for the existing functions in `cupyx.scipy.ndimage`. I don't recall offhand if we have any tests or examples using these types with `ElementwiseKernel`s?

cupyx/texture/_affine.py
the-lay: There is now four elementwise kernels, because I am adding support for texture objects that are bound to linear memory instead of a CUDA array (and require `texNDfetch()`, instead of `textND()`). But I agree, and will move them to module level.

cupyx/texture/_affine.py
leofang: Ah thanks Greg. I must have missed that...

cupyx/texture/_affine.py
leofang: That's nice! Any advantage for using linear memory? Is it easier to code or more flexible?

the-lay: Never tried linear memory textures before to be honest! If I understand correctly, they are a bit easier to code and slightly faster at random reading at the expense of losing spatial optimizations (Z-order curve) and hardware texture filtering (hence why they need `texNDfetch()` instead of `texND()`). I found some benchmark comparing linear and cuda array backed textures and (it seems cuda array should be generally faster](https://github.com/ricsonc/linear_vs_texture_memory_cuda).

the-lay:
> I would like to have a bigger picture discussion. Does it make sense to split the `ndimage` code path into two, without (the existing one) and with texture (this PR), and allow users to choose based on a simple switch or a context manager without modifying code at all?

While that indeed seems convenient, unfortunately texture memory operations work only for 2D or 3D (instead of being n-dimensional). Personally, I think I prefer to have texture affine transformations separate, instead of modifying `scipyx`.

emcastillo:
> Personally, I think I prefer to have texture affine transformations separate, instead of modifying scipyx.

Hey, thanks for the contribution, this looks great.

We have discussed this in the team and we would like to avoid increasing public APIs that are not scipy/numpy as much as possible. 
We think that having this in a private namespace (something in the lines of `cupyx._texture`) and allow it to be used in `cupyx.scipy.ndimage.affine_transform` with a kwarg is a good compromise.

Although this only works for 2d or 3d inputs, you can throw an exception if the kwarg is active and an input with ndim > 2 is fed.

Let us know your thoughts :)

Thanks again!

the-lay:
> We have discussed this in the team and we would like to avoid increasing public APIs that are not scipy/numpy as much as possible.
> We think that having this in a private namespace (something in the lines of `cupyx._texture`) and allow it to be used in `cupyx.scipy.ndimage.affine_transform` with a kwarg is a good compromise.

I understand, and respectfully, if providing only scipy/numpy APIs is the goal, then it is not a good idea to add texture memory affine transformations into cupy package at all. At the same time, I would be against modifying `cupyx.scipy` method signatures and adding quite complex code differences. I think having `cupyx.scipy` as close as possible to the original `scipy` methods is important.

I am perfectly okay if this PR is not something you want, no worries! I already have [a separate python package for 3D only](https://github.com/the-lay/voltools), but it is outdated so I would just continue the work there. Let me know! :)

leofang:
@the-lay I think historically CuPy has a more stricter compliance policy for the `cupy` namespace (mirroring the NumPy APIs) than for the `cupyx` namespace. The latter not only provides the SciPy APIs (with deviations!) but also custom APIs, so I think with an additional keyword-only argument to switch to a different code path such as `cupyx._texture` is permissible. 

An obvious advantage is that you can reuse the `ndimage` test suite that covers the subset of APIs supported by this PR to test the texture code path (just make the test parametrized by texture/non-texture), so we don't have to add and maintain separate tests. Also, I don't think there would be "complex code differences" as you put it. As long as we document the new argument clearly, since it's keyword-only the API's signature and default behavior are still the same, and I am sure it's a welcomed addition. 

I think I've said this earlier but I will be more than happy to repeat: I think having texture-based imaging processing capability is very cool, and I personally would love to see it provided in CuPy rather than in a separate package 🙂

emcastillo:
having this in cupy is pretty cool, so let's try to find a way where everyone is happy :)

the-lay:
If that's indeed the preferred way, I am okay with making this a keyword for `cupyx.scipy.affine_transformation`. I will send update commits this week.

leofang:
Great! Thanks @the-lay. Let us know if you need any help or guidance. I am happy to help the integration/test parts if needed.

cupyx/_texture.py
```diff
@@ -0,0 +1,170 @@
...
+        fmt_kind = runtime.cudaChannelFormatKindSigned
+    elif cupy.issubdtype(data.dtype, cupy.floating):
+        fmt_kind = runtime.cudaChannelFormatKindFloat
+    else:
+        raise ValueError(f'Unsupported data type {data.dtype}')
+
+    if address_mode == 'nearest':
+        address_mode = runtime.cudaAddressModeClamp
+    elif address_mode == 'constant':
+        address_mode = runtime.cudaAddressModeBorder
```
the-lay, line 71: I had to remove modes "wrap" and "mirror". If cudaTextureDesc::normalizedCoords is set to zero, cudaAddressModeWrap and cudaAddressModeMirror won't be supported and will be switched to cudaAddressModeClamp.

the-lay: I also renamed "border" and "clamp", to be consistent with scipy mode names.

the-lay:
@leofang thanks! I am a bit stuck with testing, could you advise?
I found `TestAffineTransform`, but I am confused with how to make it work with `texture_memory` keyword that I have added. As mentioned before, texture memory affine transform will work only for 2D and 3D data, with specific modes and order. If I just add it to the `@testing.parameterize` it wouldn't and shouldn't work for many cases.

leofang (commented):
Just a quick reply to your question:

> I found `TestAffineTransform`, but I am confused with how to make it work with `texture_memory` keyword that I have added. As mentioned before, texture memory affine transform will work only for 2D and 3D data, with specific modes and order. If I just add it to the `@testing.parameterize` it wouldn't and shouldn't work for many cases.

I think you could add a skip condition to only test valid parameter combinations (see how `pytest.skip()` is used in the file). You can also check if exceptions are correctly raised for invalid parameters.

the-lay:
Sorry for delay, had a few busy weeks.I have returned back to this but I found new obstacles. I would appreciate if someone could guide me on how to address them.

1. `tex2d` does not seem to have an overloaded variant for float16, i.e. 
```python
k = ElementwiseKernel('...', 'T output', '... output = tex2D<T>(...)')
f16 = cupy.zeros(.., dtype=cp.float16)
k(..., output=f16)
```
throws `cupy.cuda.compiler.CompileException: __nv_nvrtc_builtin_header.h(128934): error: no instance of overloaded function "tex2D" matches the argument list`. Is that expected or am I doing something wrong?

2. Texture memory supports only up to 32 bits per channel, up to 4 channels. I've implemented splitting dtypes that are >32 bits and <=128 bits into separate channels by 32 bits, reading texture as chars and then casting them back into original dtype, but it feels hacky and is probably prone to errors. Alternative would be to allow only float32 and float16 (if previous problem is resolved) dtypes.

3. The difference between scipy and texture memory results is too big, and I can't use `@testing.numpy_cupy_allclose(...)` for that. What would be another way to test this? The idea that I had is counting the number of elements mismatched, but that is not very precise...
 
4. Continuing previous problem: it seems that this testing problem at least in part it is a problem with scipy. For example, a sanity check of 90 degrees show that there is an empty column, in various modes. I will take another look at this tomorrow morning, but maybe I have to open a scipy issue about this.

<details>
  <summary>mode='constant'</summary>
  
  tex90_constant
  scipy90_constant

</details>
<details>
  <summary>mode='nearest'</summary>
  
  tex90_nearest
  scipy90_nearest


</details>

leofang:
> Sorry for delay, had a few busy weeks.

No need to say sorry @the-lay! Most of open source contributions are volunteer efforts which are based on the contributors' bandwidth and availability, we all get that! 🙂 Besides this is not blocking anything, so take your time. 

> 1. `tex2d` does not seem to have an overloaded variant for float16, i.e.
> 
> ```python
> k = ElementwiseKernel('...', 'T output', '... output = tex2D<T>(...)')
> f16 = cupy.zeros(.., dtype=cp.float16)
> k(..., output=f16)
> ```
> 
> throws `cupy.cuda.compiler.CompileException: __nv_nvrtc_builtin_header.h(128934): error: no instance of overloaded function "tex2D" matches the argument list`. Is that expected or am I doing something wrong?

Looks like this is expected, see https://forums.developer.nvidia.com/t/read-half2-directly-from-2d-texture/61561/6. The advice there is to read `float` instead and convert it to `half`. But, I think if the performance of this approach is suboptimal, it would not hurt to skip the fp16 support entirely for texture-based transforms.

> 2. Texture memory supports only up to 32 bits per channel, up to 4 channels. I've implemented splitting dtypes that are >32 bits and <=128 bits into separate channels by 32 bits, reading texture as chars and then casting them back into original dtype, but it feels hacky and is probably prone to errors. Alternative would be to allow only float32 and float16 (if previous problem is resolved) dtypes.

I think your alternative would make sense if you feel it's error prone and/or hard to maintain. But I want to make sure I understand your description correctly: we already exclude pretty much most dtypes (int/uint/bool/complex), so really the only thing we drop in the alternative is fp16, isn't it?
 
> Edit: I've removed points 3 and 4. I am now using `scipy.special.cosdg` and `scipy.special.sindg` and that fixed the precision errors. I've also found an off-by-0.5 error in my code, which is now also fixed :)

Awesome! Keep us posted.

the-lay:
> I think your alternative would make sense if you feel it's error prone and/or hard to maintain. But I want to make sure I understand your description correctly: we already exclude pretty much most dtypes (int/uint/bool/complex), so really the only thing we drop in the alternative is fp16, isn't it?

Thank you for the comments! Yes, I think supporting only fp32 is the best approach for now.
I have pushed new updates and I think the PR is more or less feature ready. Could someone review the code and request tests?

<details>
  <summary>Code example of how to use texture memory affine transform</summary>

  ```python
  
  import cupy as cp
  import numpy as np
  from cupyx.scipy.ndimage import affine_transform
  from scipy.special import sindg, cosdg
  import matplotlib.pyplot as plt
  
  
  def _2d_rotation_matrix(theta, rotation_center):
      c, s = cosdg(theta), sindg(theta)
      m = np.array([
          [1, 0, rotation_center[0]],
          [0, 1, rotation_center[1]],
          [0, 0, 1]
      ], np.float32)
      m = np.dot(m, np.array([
          [c, -s, 0],
          [s, c, 0],
          [0, 0, 1]
      ], np.float32))
      m = np.dot(m, np.array([
          [1, 0, -rotation_center[0]],
          [0, 1, -rotation_center[1]],
          [0, 0, 1]
      ], np.float32))
      return m
  
  
  data = cp.random.rand(100, 100, dtype=cp.float32)
  degrees = 33
  rot_m = _2d_rotation_matrix(degrees, np.divide(np.subtract(data.shape, 1), 2))
  rot_m = cp.array(rot_m)
  
  data_rot = affine_transform(data, rot_m)
  data_rot_pr = affine_transform(data, rot_m, texture_memory=True)
  abs_diff = cp.abs(data_rot - data_rot_pr)
  
  
  fig, ax = plt.subplots(1, 4)
  ax[0].set_title('Original data')
  ax[0].imshow(data.get())
  ax[1].set_title(f'Rotated {degrees} deg')
  ax[1].imshow(data_rot.get(), vmin=0)
  ax[2].set_title(f'Rotated {degrees} deg - TexMem PR')
  ax[2].imshow(data_rot_pr.get(), vmin=0)
  ax[3].set_title('Abs diff')
  ax[3].imshow(abs_diff.get(), vmin=0)
  plt.show()
  
  ```
  cupy_pr


</details>

leofang:
Nice! @the-lay Could you merge or rebase with master? The pre-tests are failing due to some old dependencies.

the-lay:
> Nice! @the-lay Could you merge or rebase with master? The pre-tests are failing due to some old dependencies.

Done, pre-tests seem to pass now.

kmaehashi:
Thanks for working on this @the-lay!
GitHub doesn't run GitHub workflows (pretest) for first-time contributors until any of the admin members clicks the "Approve Workflow Run" button for each commit, so there will be a time lag between push and pretest run.

FYI, you can also check coding styles locally. We are using these linters: https://github.com/cupy/cupy/blob/9188895216af5a713aa62fa2018364e42285097a/setup.py#L42-L47

leofang:
```
./tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py:1:14: E999 SyntaxError: invalid syntax
```

cupyx/_texture.py
leofang: I was refreshing my memory on the texture memory. `normalizedCoords` is not easy to use in this context, right?

cupyx/texture/_affine.py
leofang: @the-lay @grlee77  Looks like we're not going to support multi-channel texture (which I am totally fine). But, I am still curious about the use case of transforming a stack/batch of images. I think this is possible with `ndimage` API:
```python
>>> import cupy as cp
>>> import cupyx.scipy.ndimage as img
>>>
>>> a = cp.random.random((4, 16, 16))  # 4 images of size 16*16
>>> b = cp.random.random((3, 3))
>>> out = img.affine_transform(a, b)
>>> out.shape
(4, 16, 16)
``` 
but perhaps to make this work with texture memory we need to support [layered texture](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#layered-textures)? If it's the case, I am happy to play with it and see if I can create a PR once this one is in.

cupyx/scipy/ndimage/interpolation.py
```diff
@@ -309,7 +310,8 @@ def map_coordinates(input, coordinates, output=None, order=3,
 
 
 def affine_transform(input, matrix, offset=0.0, output_shape=None, output=None,
-                     order=3, mode='constant', cval=0.0, prefilter=True):
+                     order=3, mode='constant', cval=0.0, prefilter=True,
```
emcastillo, line 313: let;s mark this as purely kwarg
```suggestion
                     order=3, mode='constant', cval=0.0, prefilter=True, *,
```

cupyx/_texture.py
```diff
@@ -0,0 +1,194 @@
...
+        read_mode = runtime.cudaReadModeElementType
+    elif read_mode == 'normalized_float':
+        read_mode = runtime.cudaReadModeNormalizedFloat
+    else:
+        raise ValueError(
+            f'Unsupported read mode {read_mode} '
+            '(supported: element_type, normalized_float)')
+
+    texture_fmt = texture.ChannelFormatDescriptor(
+        data.itemsize * 8, 0, 0, 0, fmt_kind)
```
emcastillo, line 96: I am talking from ignorance here, but why are you only creating the channel format descriptor with 1d?
(Again, I am quite ignorant on this :) )

emcastillo:
This is looking great!

cupyx/_texture.py
the-lay: It should be possible to use normalized coordinates and that in turn should allow using mirror and wrap texture modes. I will take a look in the next few days, maybe it is not too much effort. Thanks for the suggestion!

cupyx/_texture.py
the-lay: According to the docs, texels are "restricted to the basic integer and single-precision floating-point types and any of the 1-, 2-, and 4-component vector types defined in Built-in Vector Types that are derived from the basic integer and single-precision floating-point types. ".

If we didn't need interpolation, we could store up to 128bits data types (32bits in each channel), read them by components and then cast back to input data type. For example, have a texture store 2 channel int32 data type (int2) and then after sampling, cast it to double with `__hiloint2double` intrinsic. However, the whole idea relies on interpolation, so we can only support data types of up to 32 bits :(

We could however support up to 4 channel images, i.e. RGB 2D images. I will try to implement this case, thanks for the suggestion!

cupyx/texture/_affine.py
the-lay: Actually I will try to implement support for up to 4 channel in the next commit, that should be doable, as long as the data type is fixed to float32.

I didn't know about layered textures, looks interesting! It seems they support only 1D and 2D, but as many channels (layers) as needed.

leofang:
@the-lay Just a drive-by comment (on my cell): We don't have to do everything in a single PR. It's fine to send follow-up PRs as you see fit. I'd say the current one is ready to go as is. (I haven't done a full review but if Emilio already did it's enough.)

emcastillo:
Jenkins, test this please

emcastillo:
```
./tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py:1:14: E999 SyntaxError: invalid syntax
Error: Process completed with exit code 1.
```

cupyx/texture/_affine.py
leofang: Ah yes, layered texture is only for 1D & 2D. But I think transforming a stack of 2D images is a common task. I will try to put together some code and test the performance when I have time.

cupyx/_texture.py
```diff
@@ -0,0 +1,194 @@
...
+    # CUDAArray: last dimension is the fastest changing dimension
+    array = texture.CUDAarray(texture_fmt, *data.shape[::-1])
```
leofang, line 98: We should check if `data` is C contiguous or not.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
```diff
@@ -272,6 +272,127 @@ def test_invalid_output_dtype(self):
...
+
+        if len(self.shape) == 2:
+            matrix = self._2d_rotation_matrix(self.theta, center)
+        elif len(self.shape) == 3:
+            matrix = self._3d_rotation_matrix(self.theta, center)
+        else:
+            return pytest.xfail('Unsupported shape')
+
+        if self.output == 'empty':
+            output = xp.empty_like(a)
```
leofang, line 374: This path is not tested. Maybe you meant this?
```suggestion
        if self.output:
            output = xp.empty_like(a)
```

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
```diff
@@ -272,6 +272,127 @@ def test_invalid_output_dtype(self):
...
+            m = cupyx.scipy.ndimage.affine_transform
+            matrix = cupy.array(matrix)
+            return m(a, matrix, output_shape=output_shape,
+                     output=output, order=self.order,
+                     mode=self.mode, texture_memory=True)
+        else:
+            m = scp.ndimage.affine_transform
+            return m(a, matrix, output_shape=output_shape,
+                     output=output, order=self.order,
+                     mode=self.mode)
```
leofang, line 394: Nitpick:
```suggestion
        if xp == cupy:
            m = cupyx.scipy.ndimage.affine_transform
            matrix = cupy.array(matrix)
            use_texture = True
        else:
            m = scp.ndimage.affine_transform
            use_texture = False
        return m(a, matrix, output_shape=output_shape,
                 output=output, order=self.order,
                 mode=self.mode, texture_memory=use_texture)
```

leofang (changes requested):
LGTM except for a few minor issues 👍

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
emcastillo: another nitpick :p

```suggestion
        use_texture = xp == cupy
        if use_texture:
            m = cupyx.scipy.ndimage.affine_transform
            matrix = cupy.array(matrix)
        else:
            m = scp.ndimage.affine_transform
```

the-lay: @emcastillo @leofang  I agree it looks better, but I think both of your options wouldn't work since `texture_memory` keyword is present only in cupy: scipy's `affine_transform` will throw an unexpected keyword argument.

leofang: Ahahaha sounds like both Emilio and I are lack of sleep....

cupyx/_texture.py
the-lay: Maybe changing `array.copy_from(data)` (few lines lower) to `array.copy_from(cupy.ascontiguousarray(data))` would be better?

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
the-lay: Thanks for spotting this! I wanted to check this combination too and just forgot. I've just pushed a new commit that uses this pathway and added a new check for output shapes to be consistent with scipy.

the-lay:
> ```
> ./tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py:1:14: E999 SyntaxError: invalid syntax
> Error: Process completed with exit code 1.
> ```

I don't understand where this error comes from. I have tried autopep8 and flake8 locally, and it doesn't show up. Does it have something to do with the configuration?

cupyx/_texture.py
leofang: Apparently that was my overlook! `copy_from` and `copy_to` should handle this internally. Let me send a PR for it so you shouldn't have to make any change here.

leofang: Tracked in #5341.

leofang:
Hmmm I can't reproduce the flake8 error either. Let's see if it's transient or not.

Jenkins, test this please

leofang:
Looks like I don't have access to trigger pretest 😅

cupyx/_texture.py
```diff
@@ -0,0 +1,197 @@
...
+        output = cupy.zeros(output_shape, dtype=dtype)
+    elif isinstance(output, (type, cupy.dtype)):
+        if output != cupy.float32:
+            raise ValueError(f'Texture memory affine transformation is '
+                             f'available only for float32 data type (not '
+                             f'{output})')
+        output = cupy.zeros(output_shape, dtype=output)
+    elif isinstance(output, cupy.ndarray):
+        if output.shape != output_shape:
+            raise ValueError(f'Output shapes do not match')
```
leofang, line 192: @the-lay check out the pretest error:
```
./cupyx/_texture.py:192:30: F541 f-string is missing placeholders
```
I can reproduce this locally.
```suggestion
            raise ValueError('Output shapes do not match')
```

leofang (changes requested):

cupyx/_texture.py
the-lay: Thanks, I forgot to run linter after the last commit!

the-lay:
Not sure if I trigger testing, but I'll try:

Jenkins, test this please

emcastillo:
sorry, you can't :)
Let me check if the pretest passes, and then I will trigger it

emcastillo:
Jenkins, test this please

emcastillo:
Jenkins, test this please

emcastillo (approved):

the-lay:
The part of build that failed is "cupy-py3-cutensor,mn1-p100" and if I understand correctly [something is wrong with the git](https://jenkins.preferred.jp/job/chainer/job/cupy_pr/3918/TEST=cupy-py3-cutensor,label=mn1-p100/console) configuration.

leofang:
Jenkins, test this please

leofang:
Nice work, @the-lay! Thanks again for your contribution!

the-lay:
Thanks for the help, everybody!

kmaehashi:
Please ignote the above error :pray:

PR #5172, opened 2021-05-03
Per device stream

### LINKED ISSUE #3980: Current streams must be managed for each device
Currently, it is the users' responsibility to switch the current stream after changing the current device.
I think it's better to let CuPy manage it.

```py
print(cupy.cuda.get_current_stream().ptr)  # 0
cupy.cuda.Stream().use()  # stream on GPU 0
print(cupy.cuda.get_current_stream().ptr)  # 55448656
cupy.cuda.Device(1).use()  # switch to GPU 1
print(cupy.cuda.get_current_stream().ptr)  # 55448656
cupy.ones(10)
-> cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_HANDLE: invalid resource handle
```

### LINKED ISSUE #5143: Stream stack does not take care of garbage-collected Stream (i.e. weakref is None)
```py
import cupy as cp
from cupy import cuda

with cuda.Stream() as s1:
    s2 = cuda.Stream()
    s2.use()
    with cuda.Stream() as s3:
        del s2
```

This fails when exiting `s3` context:

```
  File "test.py", line 8, in <module>
    del s2
  File "cupy/cuda/stream.pyx", line 197, in cupy.cuda.stream.BaseStream.__exit__
  File "cupy/cuda/stream.pyx", line 34, in cupy.cuda.stream._ThreadLocal.set_current_stream_ref
AttributeError: 'NoneType' object has no attribute 'ptr'
```

### LINKED ISSUE #5145: Stream destructor must take care of PTDS

leofang:
Close #3980. Fix #5143. Fix #5145.

For the reason in https://github.com/cupy/cupy/issues/3980#issuecomment-831487066, we call `getDevice()` every time we need to look up the device.

***UPDATE:*** major changes:
- Stream references are kept when entering the `with stream` context to avoid dead references, but not when calling `stream.use()` (which still keeps only the stream weakref)
- The stream stack now also records the *current* stream so that if a stream is deleted it can fall back correctly.

TODO:
- [x] Add tests

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -181,6 +181,15 @@ def get_elapsed_time(start_event, end_event):
     return runtime.eventElapsedTime(start_event.ptr, end_event.ptr)
 
 
+cdef void check_stream_device_match(int dev) except*:
+    cdef int curr_dev = runtime.getDevice()
+    if dev == -1:
+        dev = curr_dev
+    if dev != curr_dev:
+        raise RuntimeError(
+            f'This stream was not created on device {curr_dev}')
+
+
 class BaseStream(object):
 
     """CUDA stream.
@@ -202,8 +211,9 @@ class BaseStream(object):
     def __enter__(self):
         tls = _ThreadLocal.get()
         cdef int dev = self.dev
-        if dev == -1:
-            dev = runtime.getDevice()
+        check_stream_device_match(dev)
+        # to prevent from popping the wrong stream at exit
+        self._curr_dev = dev
         if tls.prev_stream_ref_stack[dev] is None:
             tls.prev_stream_ref_stack[dev] = []
         prev_stream_ref = tls.get_current_stream_ref(dev)
@@ -213,9 +223,8 @@ class BaseStream(object):
 
     def __exit__(self, *args):
         tls = _ThreadLocal.get()
-        cdef int dev = self.dev
-        if dev == -1:
-            dev = runtime.getDevice()
+        dev = self._curr_dev
+        self._curr_dev = -1
         prev_stream_ref = tls.prev_stream_ref_stack[dev].pop()
         tls.set_current_stream_ref(prev_stream_ref)
 
@@ -228,6 +237,7 @@ class BaseStream(object):
         If you want to switch a stream temporarily, use the *with* statement.
         """
         tls = _ThreadLocal.get()
+        check_stream_device_match(self.dev)
         tls.set_current_stream(self)
         return self
 
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -36,37 +36,32 @@ cdef class _ThreadLocal:
         if dev == -1:
             dev = runtime.getDevice()
         stream_module.set_current_stream_ptr(ptr, dev)
-        self.current_stream[dev] = <intptr_t>ptr
+        self.current_stream[dev] = ptr
         self.current_stream_ref[dev] = weakref.ref(stream)
 
     cdef set_current_stream_ref(self, stream_ref):
-        cdef intptr_t ptr = <intptr_t>stream_ref().ptr
+        cdef intptr_t ptr = <intptr_t>(stream_ref().ptr)
         cdef int dev = stream_ref().dev
         if dev == -1:
             dev = runtime.getDevice()
         stream_module.set_current_stream_ptr(ptr, dev)
-        self.current_stream[dev] = <intptr_t>ptr
+        self.current_stream[dev] = ptr
         self.current_stream_ref[dev] = stream_ref
 
     cdef get_current_stream(self, int dev=-1):
-        if dev == -1:
-            dev = runtime.getDevice()
-        if self.current_stream_ref[dev] is None:
-            if stream_module.is_ptds_enabled():
-                self.current_stream_ref[dev] = weakref.ref(Stream.ptds)
-            else:
-                self.current_stream_ref[dev] = weakref.ref(Stream.null)
-        return self.current_stream_ref[dev]()
+        return self.get_current_stream_ref(dev)()
 
     cdef get_current_stream_ref(self, int dev=-1):
         if dev == -1:
             dev = runtime.getDevice()
-        if self.current_stream_ref[dev] is None:
+        stream_ref = self.current_stream_ref[dev]
+        if stream_ref is None:
             if stream_module.is_ptds_enabled():
-                self.current_stream_ref[dev] = weakref.ref(Stream.ptds)
+                stream_ref = weakref.ref(Stream.ptds)
             else:
-                self.current_stream_ref[dev] = weakref.ref(Stream.null)
-        return self.current_stream_ref[dev]
+                stream_ref = weakref.ref(Stream.null)
+            self.current_stream_ref[dev] = stream_ref
+        return stream_ref
 
     cdef intptr_t get_current_stream_ptr(self, int dev=-1):
         # Returns the stream previously set, otherwise returns
@@ -75,7 +70,7 @@ cdef class _ThreadLocal:
         if dev == -1:
             dev = runtime.getDevice()
         cdef intptr_t curr_stream = self.current_stream[dev]
-        if stream_module.is_ptds_enabled() and curr_stream == 0:
+        if curr_stream == 0 and stream_module.is_ptds_enabled():
             return <intptr_t>runtime.streamPerThread
         return curr_stream
 
@@ -181,13 +176,15 @@ def get_elapsed_time(start_event, end_event):
     return runtime.eventElapsedTime(start_event.ptr, end_event.ptr)
 
 
-cdef void check_stream_device_match(int dev) except*:
+cdef int check_stream_device_match(int dev) except? -1:
+    """Check if the stream was created on the current device."""
     cdef int curr_dev = runtime.getDevice()
     if dev == -1:
         dev = curr_dev
     if dev != curr_dev:
         raise RuntimeError(
             f'This stream was not created on device {curr_dev}')
+    return dev
 
 
 class BaseStream(object):
@@ -206,12 +203,15 @@ class BaseStream(object):
         # This operator is implemented to compare the singleton instance
         # of null stream (Stream.null) can safely be compared with null
         # stream instance created by a user.
-        return self.ptr == other.ptr
+        if self.dev == -1 or other.dev == -1:
+            return self.ptr == other.ptr
+        else:
+            return (self.dev == other.dev and self.ptr == other.ptr)
 
     def __enter__(self):
         tls = _ThreadLocal.get()
         cdef int dev = self.dev
-        check_stream_device_match(dev)
+        dev = check_stream_device_match(dev)
         # to prevent from popping the wrong stream at exit
         self._curr_dev = dev
         if tls.prev_stream_ref_stack[dev] is None:
@@ -223,13 +223,13 @@ class BaseStream(object):
 
     def __exit__(self, *args):
         tls = _ThreadLocal.get()
-        dev = self._curr_dev
-        self._curr_dev = -1
+        cdef int dev = self._curr_dev
         prev_stream_ref = tls.prev_stream_ref_stack[dev].pop()
         tls.set_current_stream_ref(prev_stream_ref)
 
     def __repr__(self):
-        return '<{} {}>'.format(type(self).__name__, self.ptr)
+        return '<{} {} (device {})>'.format(
+            type(self).__name__, self.ptr, self.dev)
 
     def use(self):
         """Makes this stream current.
@@ -237,7 +237,8 @@ class BaseStream(object):
         If you want to switch a stream temporarily, use the *with* statement.
         """
         tls = _ThreadLocal.get()
-        check_stream_device_match(self.dev)
+        cdef int dev = self.dev
+        check_stream_device_match(dev)
         tls.set_current_stream(self)
         return self
 
@@ -380,7 +381,7 @@ class Stream(BaseStream):
             return
         tls = _ThreadLocal.get()
         if self.ptr:
-            current_ptr = <intptr_t>tls.get_current_stream_ptr()
+            current_ptr = tls.get_current_stream_ptr(self.dev)
             if <intptr_t>self.ptr == current_ptr:
                 tls.set_current_stream(self.null)
             runtime.streamDestroy(self.ptr)
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -205,11 +205,7 @@ class BaseStream(object):
         # This operator is implemented to compare the singleton instance
         # of null stream (Stream.null) can safely be compared with null
         # stream instance created by a user.
-        if self.device_id == -1 or other.device_id == -1:
-            return self.ptr == other.ptr
-        else:
-            return (self.device_id == other.device_id
-                    and self.ptr == other.ptr)
+        return self.ptr == other.ptr
 
     def __enter__(self):
         tls = _ThreadLocal.get()
@@ -376,7 +372,7 @@ class Stream(BaseStream):
         if is_shutting_down():
             return
         tls = _ThreadLocal.get()
-        if self.ptr not in (0, 1, 2):
+        if self.ptr not in (0, runtime.streamLegacy, runtime.streamPerThread):
             current_ptr = tls.get_current_stream_ptr()
             if <intptr_t>self.ptr == current_ptr:
                 tls.set_current_stream(get_default_stream())
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -12,15 +12,15 @@ cdef object _thread_local = threading.local()
 
 
 cdef class _ThreadLocal:
-    cdef list current_stream  # list of intptr_t
     cdef list current_stream_ref  # list of object
+    cdef list current_device_id_stack  # list of int
     cdef list prev_stream_ref_stack  # list of list
 
     def __init__(self):
         cdef int i, num_devices = runtime.getDeviceCount()
-        self.current_stream = [0 for i in range(num_devices)]
         self.current_stream_ref = [None for i in range(num_devices)]
         self.prev_stream_ref_stack = [None for i in range(num_devices)]
+        self.current_device_id_stack = []
 
     @staticmethod
     cdef _ThreadLocal get():
@@ -30,13 +30,26 @@ cdef class _ThreadLocal:
             tls = _thread_local.tls = _ThreadLocal()
         return <_ThreadLocal>tls
 
+    cdef void push_stream(self, stream, int dev) except *:
+        if self.prev_stream_ref_stack[dev] is None:
+            self.prev_stream_ref_stack[dev] = []
+        prev_stream_ref = self.get_current_stream_ref(dev)
+        self.prev_stream_ref_stack[dev].append(prev_stream_ref)
+        # record dev to prevent from popping the wrong stream at exit
+        self.current_device_id_stack.append(dev)
+        self.set_current_stream(stream)
+
+    cdef pop_stream(self):
+        cdef int dev = self.current_device_id_stack.pop()
+        prev_stream_ref = self.prev_stream_ref_stack[dev].pop()
+        self.set_current_stream_ref(prev_stream_ref)
+
     cdef set_current_stream(self, stream):
         cdef intptr_t ptr = <intptr_t>stream.ptr
         cdef int dev = stream.dev
         if dev == -1:
             dev = runtime.getDevice()
         stream_module.set_current_stream_ptr(ptr, dev)
-        self.current_stream[dev] = ptr
         self.current_stream_ref[dev] = weakref.ref(stream)
 
     cdef set_current_stream_ref(self, stream_ref):
@@ -45,7 +58,6 @@ cdef class _ThreadLocal:
         if dev == -1:
             dev = runtime.getDevice()
         stream_module.set_current_stream_ptr(ptr, dev)
-        self.current_stream[dev] = ptr
         self.current_stream_ref[dev] = stream_ref
 
     cdef get_current_stream(self, int dev=-1):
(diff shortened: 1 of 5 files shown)


cupy/cuda/stream.pyx
```diff
@@ -182,16 +193,18 @@ class BaseStream(object):
...
         prev_stream_ref = tls.get_current_stream_ref()
-        tls.prev_stream_ref_stack.append(prev_stream_ref)
+        tls.prev_stream_ref_stack[dev].append(prev_stream_ref)
         tls.set_current_stream(self)
         return self
 
     def __exit__(self, *args):
         tls = _ThreadLocal.get()
-        prev_stream_ref = tls.prev_stream_ref_stack.pop()
+        cdef int dev = runtime.getDevice()
```
takagi, line 206: Should we need to keep the device id on `__enter__` as the current device may be changed in the block?

leofang: I have been thinking about this too. It is more complicated than I expected. I feel it is still necessary to tie a stream to a device so we know which device the stream was created on without looking up the current device (which may not be the same as you pointed out).

leofang: Hmmm...I changed my mind and now feel this should be illegal:
```python
with cuda.Device(0):
    stream0 = cuda.Stream()
    with stream0:
        with cuda.Device(1):
            ...  # <--- stream0 does not work on device 1
```

leofang: It has to be legal so that we can switch device, create a new stream, and set it current.

cupy/cuda/stream.pyx
```diff
@@ -362,14 +402,30 @@ class ExternalStream(BaseStream):
...
-    def __init__(self, ptr):
+    def __init__(self, ptr, dev=-1):
         self.ptr = ptr
+        # It is in theory unsafe to just call runtime.getDevice() here, as the
+        # stream pointer could come from a different device (although
+        # unlikely). While we could use driver API combos cuStreamGetCtx ->
+        # cuCtxSetCurrent -> cuCtxGetDevice -> ... to retrieve the device ID
+        # associated with the stream, it is way too complicated. Let us keep
+        # this as thin as possible.
+        self.dev = dev
```
leofang, line 428: There is no single API to look up the device that the stream resides on, so for a lightweight operation I am leaving the lookup out. Please let me know if we prefer to do it rigorously.

leofang: Another note: HIP does not support `cuStreamGetCtx`.

leofang:
I feel the current implementation (up to 105503a) adds too many calls to `getDevice()` which should not be necessary. Let me see what I can do.

leofang:
Jenkins, test this please

cupy/cuda/stream.pyx
```diff
@@ -314,6 +349,8 @@ class Stream(BaseStream):
 
     Attributes:
         ~Stream.ptr (intptr_t): Raw stream handle.
+        ~Stream.dev (int): The ID of the device that the stream was created on.
+            The value ``-1`` is used for the singleton stream objects.
```
leofang, line 353: If we create a singleton object per device, the `dev` attribute could be more meaningful, saving us a few lookups. However, it'd be a breaking change which I am not sure if acceptable or not.

cupy/cuda/stream.pyx
```diff
@@ -178,32 +203,42 @@ class BaseStream(object):
...
+        else:
+            return (self.dev == other.dev and self.ptr == other.ptr)
 
     def __enter__(self):
         tls = _ThreadLocal.get()
-        if tls.prev_stream_ref_stack is None:
-            tls.prev_stream_ref_stack = []
-        prev_stream_ref = tls.get_current_stream_ref()
-        tls.prev_stream_ref_stack.append(prev_stream_ref)
+        cdef int dev = self.dev
```
kmaehashi, line 213: I think it's better to move `__enter__` (and `__exit__`) implementation to `tls` (like `tls.push_stream(self.ptr, self.dev)`) to aovid modifying TLS internal states in various places.

cupy_backends/cuda/stream.pyx
```diff
@@ -50,11 +52,12 @@ cdef intptr_t get_current_stream_ptr():
     return <intptr_t>tls.get_current_stream_ptr()
 
 
-cdef set_current_stream_ptr(intptr_t ptr):
+cdef set_current_stream_ptr(intptr_t ptr, int dev=-1):
```
kmaehashi, line 55: Could you use `device_id` instead of `dev` for consistency with `cupy.cuda.Memory*` APIs?

cupy/cuda/stream.pyx
```diff
@@ -12,9 +12,15 @@ cdef object _thread_local = threading.local()
...
-    cdef intptr_t current_stream
-    cdef object current_stream_ref
-    cdef list prev_stream_ref_stack
+    cdef list current_stream  # list of intptr_t
+    cdef list current_stream_ref  # list of object
+    cdef list prev_stream_ref_stack  # list of list
+
+    def __init__(self):
+        cdef int i, num_devices = runtime.getDeviceCount()
+        self.current_stream = [0 for i in range(num_devices)]
```
kmaehashi, line 21: This is a matter of preference but you can also write this like this:
```suggestion
        self.current_stream = [0] * len(num_devices)
```

kmaehashi:
I feel TLS implementation is a bit complicated because we need to avoid creating `weakref`s everytime setting a stream. Maybe doing `self._weakref = weakref.ref(self)` in `Stream`/`ExternalStream` constructor may simplify the code?
(This is out of the scope of the original motivation of this PR, so you can ignore this comment 😄 )

cupy_backends/cuda/stream.pyx
leofang: Great! I forgot that, thanks.

leofang:
> I feel TLS implementation is a bit complicated because we need to avoid creating `weakref`s everytime setting a stream.

@kmaehashi I don't think I get it, could you elaborate?

> (This is out of the scope of the original motivation of this PR, so you can ignore this comment 😄 )

That's OK, let me address it properly in this PR.

kmaehashi:
@leofang 
> I feel TLS implementation is a bit complicated because we need to avoid creating weakrefs everytime setting a stream. > Maybe doing self._weakref = weakref.ref(self) in Stream/ExternalStream constructor may simplify the code?

What I meant is that if we create a weakref of the stream inside `__init__` of `Stream`/`ExternalStream` (i.e., let each stream instance hold the weakref to self), we can remove `get_current_stream_ref`/`set_current_stream_ref` APIs in TLS.

cupy/cuda/stream.pyx
kmaehashi: I guess reducing `cudaGetDevice` lookup is difficult as CuPy may be used in conjunction with other libraries. For example CuPy code may be placed under `with torch.cuda.device(1):`.

leofang:
@kmaehashi Since I'm working on the stream stack, it'd be nice if the team could make a decision on #4337 🙂 Specifically, do we want to preserve the current behavior if the stream `s` is deleted after `s.used()` is called?

cupy/cuda/stream.pyx
```diff
@@ -178,32 +203,42 @@ class BaseStream(object):
...
     def __enter__(self):
         tls = _ThreadLocal.get()
-        if tls.prev_stream_ref_stack is None:
-            tls.prev_stream_ref_stack = []
-        prev_stream_ref = tls.get_current_stream_ref()
-        tls.prev_stream_ref_stack.append(prev_stream_ref)
+        cdef int dev = self.dev
+        dev = check_stream_device_match(dev)
+        # to prevent from popping the wrong stream at exit
+        self._curr_dev = dev
```
takagi, line 216: As a `BaseStream` instance may `__enter__` recursively (regardless if it makes sense), we also have to manage `_curr_dev` in the stack. Pushing `(dev, prev_stream_ref)` instead of `prev_stream_ref` only and poping it would be enough.

```Python
with cuda.Device(0):
    s = cuda.Stream()
    with s:
        with cuda.Device(1):
            with s:  # <--- s._curr_dev is replaced with 1
                ...
        ...
        # The first context `__exit__`s with popping the wrong stream
```

takagi: (maybe nitpick)

leofang: Your snippet will raise an error with this PR, because I added a `check_stream_device_match` check. However, I haven't thought about this possibility (which is clearly an abuse that should not happen...):
```python 
with cuda.Device(0):
    s = cuda.Stream()
    with s:
        with s:  # <--- s._curr_dev is washed out
            ...
        ...
```
@takagi should we worry about this?

leofang:
> it'd be nice if the team could make a decision on #4337 🙂 Specifically, do we want to preserve the current behavior if the stream `s` is deleted after `s.used()` is called?

The resolution is to keep the present behavior as is, so after `s` is deleted we fall back to the default stream.

cupy/cuda/stream.pyx
takagi: Please leave it as it is. I was trapped by a minor thing. (I initially thought of making the interpreter's context stack manage the current stream history, but given `Stream.null` and `Stream.ptds`, I gave it up as they are shared between threads and we eventually need thread-local data management unless we provide the instances (`null` and `ptds`) for each thread)

leofang:
@takagi @kmaehashi I need your input on another design decision. This time it's related to #4445 with the assumption that in the near future the two stream modules `cupy/cuda/stream.pyx` and `cupy_backends/cuda/stream.pyx` remain separate.

Currently we store the current stream pointer in the thread-local storages of both modules. I would like to reduce the repetition. Should we separate the roles of the two storages in the following fashion?
- In `cupy/cuda/stream.pyx` we store the stream stack and weak refs (see also https://github.com/cupy/cupy/pull/5172#discussion_r627048001)
- In `cupy_backends/cuda/stream.pyx` we store the current stream pointer

The only change IIUC is to remove `self.current_stream` from the the thread-local storage in `cupy/cuda/stream.pyx`. I hope I am not missing anything...

cupy/cuda/stream.pyx
leofang: I realized accessing `self._curr_dev` is unsafe either for precisely the same reason (the singletons are not thread-safe), but your suggestion 

> Pushing `(dev, prev_stream_ref)` instead of `prev_stream_ref` only and poping it would be enough.

cannot be taken as is, because we need to know `dev` to pop from the right stack. We need to keep track of two stacks to address this. Working on it...

leofang:
> The only change IIUC is to remove `self.current_stream` from the the thread-local storage in `cupy/cuda/stream.pyx`. I hope I am not missing anything...

I made the judgement call to do this in 9cdd026.

cupy/cuda/stream.pyx
leofang: Fixed in 9cdd026.

cupy_backends/cuda/stream.pyx
leofang: Done in 253510f.

cupy/cuda/stream.pyx
```diff
@@ -178,32 +208,35 @@ class BaseStream(object):
         # This operator is implemented to compare the singleton instance
         # of null stream (Stream.null) can safely be compared with null
         # stream instance created by a user.
-        return self.ptr == other.ptr
+        if self.device_id == -1 or other.device_id == -1:
+            return self.ptr == other.ptr
+        else:
+            return (self.device_id == other.device_id
+                    and self.ptr == other.ptr)
```
leofang, line 215: This is actually a bug fix that I think we should backport...For example, multiple devices can have stream 88 created, but they should not compare to equal.

leofang:
Power cycle to restart doc build tests.

leofang:
Looks like I need to merge with master...

leofang:
What's wrong with the CI...

cupy/cuda/stream.pyx
takagi: Okay, I'll make a PR that backports this part after #5172 is merged.

takagi:
This PR looks good to me now. Do you still have something to add here (i.g. #5143), @leofang?  And would you take a look again @kmaehashi?

leofang:
Yes I am still working on a fix for #5143. I lost access to my remote workstation, so it'd take some time to recover.

kmaehashi (commented):
Thanks @leofang for working on this!

cupy/cuda/stream.pyx
kmaehashi: Sorry, I couldn't get the point here. I think it's safe to assume that different streams have different pointers?

takagi: It's the case for `Stream.null` and `Stream.ptds`.

cupy/cuda/stream.pyx
```diff
@@ -24,42 +30,53 @@ cdef class _ThreadLocal:
...
+        prev_stream_ref = self.prev_stream_ref_stack[device_id].pop()
+        self.set_current_stream_ref(prev_stream_ref)
+
     cdef set_current_stream(self, stream):
         cdef intptr_t ptr = <intptr_t>stream.ptr
-        stream_module.set_current_stream_ptr(ptr)
-        self.current_stream = <intptr_t>ptr
-        self.current_stream_ref = weakref.ref(stream)
+        cdef int device_id = stream.device_id
+        if device_id == -1:
```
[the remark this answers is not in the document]
leofang: Yes I am looking into this , but i doubt it can be removed. For `null` and `ptds` singletons we still need to keep track the device IDs, as they do not carry the current device info (set to -1). But maybe I am just overlooking something.

cupy/cuda/stream.pyx
leofang: Right, because they are "global" singletons, not "per-device" singletons.

cupy/cuda/stream.pyx
```diff
@@ -24,42 +40,40 @@ cdef class _ThreadLocal:
...
         return <_ThreadLocal>tls
 
+    cdef void push_stream(self, stream, int device_id) except*:
+        assert device_id >= 0
+        self.current_stream_stack[device_id].append(stream)
+        # record device_id to prevent from popping the wrong stream at exit
+        self.current_device_id_stack.append(device_id)
+        self.set_current_stream(stream)
+
+    cdef void pop_stream(self) except*:
```
kmaehashi, line 50: How about passing a `device_id` from `__exit__`? I think `current_device_id_stack` is not needed in that case.

cupy/cuda/stream.pyx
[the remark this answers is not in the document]
kmaehashi: Thank you @leofang, I agree that it's needed for null/ptds.

cupy/cuda/stream.pyx
```diff
@@ -323,31 +356,35 @@ class Stream(BaseStream):
...
+            self.device_id = runtime.getDevice()
 
     def __del__(self, is_shutting_down=_util.is_shutting_down):
         cdef intptr_t current_ptr
         if is_shutting_down():
             return
         tls = _ThreadLocal.get()
-        if self.ptr:
-            current_ptr = <intptr_t>tls.get_current_stream_ptr()
+        if self.ptr not in (0, 1, 2):
```
kmaehashi, line 379:
```suggestion
        if self.ptr not in (0, runtime.streamLegacy, runtime.streamPerThread):
```

cupy/cuda/stream.pyx
takagi: Wait. As @kmaehashi pointed out, can `cudaStreamCreate` create the same stream pointer for different devices?

kmaehashi:
Only the bug-fix part (#5156) should be backported.

cupy/cuda/stream.pyx
kmaehashi: As for singletons, device_id is always -1 and ptr is always 1 or 2, so I think device_id comparison is not necessary.

leofang: Ohhh yeah, I got it now, sorry you're right @kmaehashi, I will revert this part.

cupy/cuda/stream.pyx
leofang: Oh...This reminds me of a question: Do we want to handle the interaction between `Stream` and `Device` objects bidirectionally? The current behavior is:
```python
>>> cp.cuda.Device(0).use()
>>> s = cp.cuda.Stream()
>>> with cp.cuda.Device(1):
...     with s:
...         print(cp.cuda.get_current_stream())  # raise as we check if stream is valid on the device
... 
  File "<stdin>", line 2, in <module>
  File "cupy/cuda/stream.pyx", line 217, in cupy.cuda.stream.BaseStream.__enter__
    device_id = check_stream_device_match(device_id)
  File "cupy/cuda/stream.pyx", line 186, in cupy.cuda.stream.check_stream_device_match
    raise RuntimeError(
RuntimeError: This stream was not created on device 1
```
and
```python
>>> cp.cuda.Device(0).use()
>>> s = cp.cuda.Stream()
>>> with s:
...     with cp.cuda.Device(1):
...         print(cp.cuda.get_current_stream())  # we don't raise here <---- should we?
... 
<Stream 0 (device -1)>
```
If we remove `current_device_id_stack` and pass `device_id` from `__exit__`, we need to get the current device, which could be different from the entrant device (if `Device(1).use()` is called above).

kmaehashi: > If we remove current_device_id_stack and pass device_id from __exit__, we need to get the current device, which could be different from the entrant device (if Device(1).use() is called above).

Ah, ok, now I understood `current_device_id_stack` is needed for streams with device_id = -1 (singletons and ExtenalStream). Thank you for the explanation @leofang!

leofang: > This reminds me of a question: Do we want to handle the interaction between `Stream` and `Device` objects bidirectionally?

Reply to self: No. See https://github.com/cupy/cupy/pull/5172#discussion_r625649470.

leofang:
I think most concerns are addressed by now, if not please let me know.

I do want to point out the performance impact of this PR. Below is a simple script to time 1. synthetic workload with streams, 2. `with stream` context manager, and 3. `stream.use()` switch:
```python
import cupy as cp
from cupyx.time import repeat


N = 32
runs = 128
streams = []
for i in range(N):
    streams.append(cp.cuda.Stream())

def stream_test1():
    result = []
    for i in range(runs):
        with streams[i % N] as s:
            a = cp.random.random((2**20,))
            b = cp.sum(a)
            result.append(b)
    for i in range(N):
        streams[i].synchronize()

def stream_test2():
    cp.cuda.Stream.null.use()
    for i in range(runs):
        with streams[i % N] as s:
            pass
    cp.cuda.Stream.null.use()

def stream_test3():
    cp.cuda.Stream.null.use()
    for i in range(runs):
        streams[i % N].use()
    cp.cuda.Stream.null.use()

print(repeat(stream_test1, n_repeat=100))
print(repeat(stream_test2, n_repeat=1000))
print(repeat(stream_test3, n_repeat=1000))
```
Before this PR (current master):
```
stream_test1        :    CPU: 9198.319 us   +/-25.499 (min: 9153.547 / max: 9265.513) us     GPU-0: 9203.953 us   +/-25.540 (min: 9158.656 / max: 9271.200) us
stream_test2        :    CPU:   67.898 us   +/- 1.714 (min:   66.936 / max:   85.735) us     GPU-0:   70.155 us   +/- 1.727 (min:   68.128 / max:   88.064) us
stream_test3        :    CPU:   33.462 us   +/- 0.653 (min:   33.110 / max:   45.828) us     GPU-0:   35.764 us   +/- 0.887 (min:   34.816 / max:   48.992) us
```
With this PR:
```
stream_test1        :    CPU: 9203.129 us   +/-28.180 (min: 9139.099 / max: 9281.473) us     GPU-0: 9208.702 us   +/-28.213 (min: 9145.504 / max: 9287.552) us
stream_test2        :    CPU:  103.754 us   +/- 1.722 (min:  102.756 / max:  124.242) us     GPU-0:  106.116 us   +/- 1.818 (min:  104.448 / max:  125.856) us
stream_test3        :    CPU:   48.798 us   +/- 1.327 (min:   44.627 / max:   64.436) us     GPU-0:   51.176 us   +/- 1.393 (min:   47.008 / max:   66.624) us
```
The performance impact is negligible with sufficient workload (`stream_test1`). However, both `stream_test2` and `stream_test3` show an additional ~50% overhead, though the overhead increase *per stream* is on the order of 1 us.

leofang:
Jenkins, test this please

takagi (approved):

kmaehashi:
I noticed that the current stream is lost when pushing a stream.

```py
>>> s1 = cupy.cuda.Stream()
>>> s2 = cupy.cuda.Stream()
>>> s2.use()
<Stream 94167172638448 (device 0)>
>>> with s1:
...     pass
... 
>>> 
>>> cupy.cuda.get_current_stream()
<Stream 0 (device -1)>  # <--- should be  s2
```

kmaehashi:
I'm looking into the issue above.

takagi:
To make it work well with `use()`, I think, we need to shift the values to push into `current_stream_stack`. That is, we have to push the old current stream here instead of the new `stream`:

https://github.com/cupy/cupy/blob/8d163896d9e7fa39eea6b6ab24b4c7b6e0d97bd6/cupy/cuda/stream.pyx#L45

as well as changing `current_stream_stack`'s initial value for each device to `[]`.

https://github.com/cupy/cupy/blob/8d163896d9e7fa39eea6b6ab24b4c7b6e0d97bd6/cupy/cuda/stream.pyx#L33

leofang:
@kmaehashi @takagi Maybe I still misunderstood you. I thought by closing #4337 we agreed to not push a stream to the stack if it's used via calling `stream.use()`? With this PR, [your snippet](https://github.com/cupy/cupy/pull/5172#issuecomment-845764323) is expected.

takagi:
I understand #4337 that `with stream:` remembers the (previous) current stream when it has been called regardless it was set by `use()` or another stream context. On the other hand, `stream.use()` doesn't remember it so doesn't push it on the stack.

leofang:
If we push the "previous" current stream to the stack (the old approach), we lose the ability to rewind to the last stream in use when the stream made in use via `.use()` is deleted (#5143). I think it's fundamentally incompatible unless we push the stream to the stack unconditionally (#4337).

takagi:
Thanks, now I got #5143. I needed to work by my hands to understand these related PRs... An extra slot for each stack level that shadows the stream context when `use()` might work, but I'm not confident without writing codes for myself.

PR #5177, opened 2021-05-06
CUDA 11.2/11.3: Support `MemoryAsyncPool` statistics and limits

leofang:
- CUDA 11.2+: `get_limit()` and `set_limit()` 
- CUDA 11.3+: `used_bytes()`, `free_bytes()`, and `total_bytes()`

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -11,9 +11,6 @@ import threading
 import warnings
 import weakref
 
-from cupy_backends.cuda.api.runtime import CUDARuntimeError
-from cupy._core import syncdetect
-
 from fastrlock cimport rlock
 from libc.stdint cimport int8_t
 from libc.stdint cimport intptr_t
@@ -22,8 +19,10 @@ from libcpp cimport algorithm
 from cupy.cuda cimport device
 from cupy.cuda cimport memory_hook
 from cupy.cuda cimport stream as stream_module
+from cupy_backends.cuda.api cimport driver
 from cupy_backends.cuda.api cimport runtime
 
+from cupy_backends.cuda.api.runtime import CUDARuntimeError
 from cupy import _util
 
 
@@ -656,7 +655,7 @@ cpdef MemoryPointer malloc_async(size_t size):
     .. seealso:: `Stream Ordered Memory Allocator`_
 
     .. _Stream Ordered Memory Allocator:
-        https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html
+        https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#stream-ordered-memory-allocator
     """
     cdef intptr_t stream_ptr
     stream_ptr = stream_module.get_current_stream_ptr()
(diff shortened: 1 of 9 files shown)


leofang:
The current behavior doesn't seem right...😂
```python
>>> import cupy as cp
>>> pool = cp.cuda.MemoryAsyncPool()
<stdin>:1: FutureWarning: cupy.cuda.MemoryAsyncPool is experimental. The interface can change in the future.
>>> pool.set_limit(size=1024)
>>> m1 = pool.malloc(12)
>>> pool.used_bytes()
512
>>> m2 = pool.malloc(12)
  File "cupy/cuda/memory.pyx", line 1625, in cupy.cuda.memory.MemoryAsyncPool.malloc
    cpdef MemoryPointer malloc(self, size_t size):
  File "cupy/cuda/memory.pyx", line 1653, in cupy.cuda.memory.MemoryAsyncPool.malloc
    raise OutOfMemoryError(size, total - size, total_bytes_limit)
cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 12 bytes (allocated so far: 33,554,932 bytes, limit set to: 1,024 bytes).
```

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
The CI failure is irrelevant and should be fixed by #5169? 
https://jenkins.preferred.jp/job/chainer/job/cupy_pr/3668/TEST=cupy-py3,label=mn1-p100/testReport/junit/tests.cupyx_tests.jit_tests.test_raw/TestRaw/test_atomic_add/
Let's retry...

Jenkins, test this please

leofang:
@takagi I think this is ready for review. Note that most of the added tests are *not* run in the CI due to insufficient driver version, so this may require a local test. 

Another note: the pool size limit is a *soft* limit, in the sense that `cudaMemPoolAttrReservedMemCurrent` (for `total_bytes()`) can grow beyond `cudaMemPoolAttrReleaseThreshold` (which I used behind `get_limit()`/`set_limit()`). The limit only guarantees to not release memory back to the OS before it's reached.

cc: @jrhemstad @shwina @harrism @pentschev for vis

leofang:
One think I noticed is that this works with driver 11.3 + cudatoolkit 11.2, but not the other way around (driver 11.2 + cudatoolkit 11.3). I don't think Forward Compatibility works with CUDA mempools. This PR is tested with both the former combination and also driver = cudatoolkit = 11.3.

takagi:
Thanks! The change looks good to me, but my environment also doesn't have a driver that supports 11.3. Please wait until it gets ready so I can test it...

leofang:
No rush, @takagi! In the meantime, feel free to test this PR locally with an old driver (<11.3) and see if I am putting reasonable constraints or not. I think `*_bytes()` APIs would not work, but `*_limit()` APIs would. (Of course it'd involve commenting out some version checks I added in various places...)

takagi (approved):

takagi:
I tested the change locally with driver 11.3 and all of its tests successfully passed!

leofang:
Glad to hear! Thanks @takagi.

PR #5189, opened 2021-05-11
[WIP] Add Dependabot configuration to auto-update docs dependencies

kmaehashi:


leofang:
FYI: NumPy folks are considering to remove dependabot 
https://github.com/numpy/numpy/issues/18977
Maybe we should see what they find...

kmaehashi:
Dependabot is disabled by default in forks. But once you opt-in with your fork, there's no way to disable it 😓 
https://github.com/twbs/bootstrap/discussions/31385

leofang:
https://github.com/dependabot/dependabot-core/issues/2198

kmaehashi:
Ok I struggled a bit but couldn't turn dependabot off in forks. Let's wait until the bug is fixed...

Issue #5165, opened 2021-04-28
Support cuTENSOR 1.3


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read Release Notes ([cuTENSOR](https://docs.nvidia.com/cuda/cutensor/index.html#changelog) / [cuDNN](https://docs.google.com/document/d/1NA1fg-Lqub44dIaArvBdKhrmw4VSYYCmP2cvd8O648k/edit#) / [NCCL](https://docs.nvidia.com/deeplearning/sdk/nccl-release-notes/index.html)).
- [x] Update chainer-test to test version. https://github.com/chainer/chainer-test/pull/655
- [x] Update the library installer.
  - [x] Fix `cupyx/tools/install_library.py`. https://github.com/cupy/cupy/pull/5192
  - [x] Backport. https://github.com/cupy/cupy/pull/5234
- [x] Documentation
  - [x] Update Requirements section in [Installtion Guide](https://docs.cupy.dev/en/latest/install.html) https://github.com/cupy/cupy/pull/5196
  - [x] Backport. https://github.com/cupy/cupy/pull/5238
- [x] Implement new features using the version, if any.

leofang:
According to the release note cuTENSOR 1.3 is binary incompatible with 1.2.x, so I suppose this support cannot be backported to v9?

xref: https://github.com/conda-forge/cutensor-feedstock/issues/11

takagi:
No API addition/removal nor its signature change.

```
$ diff 1.2.2/include/cutensor.h 1.3.0/include/cutensor.h
20,21c20,21
< #define CUTENSOR_MINOR 2 //!< cuTensor minor version.
< #define CUTENSOR_PATCH 2 //!< cuTensor patch version.
---
> #define CUTENSOR_MINOR 3 //!< cuTensor minor version.
> #define CUTENSOR_PATCH 0 //!< cuTensor patch version.
256c256
<  * |    typeA   |    typeB   |    typeC   | typeCompute |
---
>  * |    typeA   |    typeB   |    typeC   | typeScalar |
258a259,261
>  * | CUDA_R_16F | CUDA_R_16F | CUDA_R_16F | CUDA_R_32F  |
>  * | CUDA_R_16BF| CUDA_R_16BF| CUDA_R_16BF| CUDA_R_16BF |
[...]

PR #5192, opened 2021-05-12
cuTENSOR v1.3.0 for library installer

takagi:
~~We may need an option to switch cuTENSOR v1.3.0 and v1.2.2 to install if v1.3.0 support is not to be backported to v9 as v1.3.0 is not binary compatible with previous versions. https://github.com/cupy/cupy/issues/5165#issuecomment-831001486~~

Just not backporting this PR is enough to leave v9 supporting v1.2.2. No additional switch is needed and we can use v9 `install_library.py`.

### DIFF
--- a/cupyx/tools/install_library.py
+++ b/cupyx/tools/install_library.py
@@ -225,7 +225,7 @@ The current platform ({}) is not supported.'''.format(target_platform))
             shutil.move(
                 os.path.join(outdir, 'libcutensor', 'lib', cuda),
                 os.path.join(destination, 'lib'))
-            if cuda == '10.2':
+            if cuda == '10.1':
                 license = 'license.pdf'  # v1.2.2
             else:
                 license = 'license.txt'  # v1.3.0
--- a/cupyx/tools/install_library.py
+++ b/cupyx/tools/install_library.py
@@ -109,21 +109,21 @@ def _make_cutensor_record(
 
 
 _cutensor_records.append(_make_cutensor_record(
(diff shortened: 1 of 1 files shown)


takagi:
The directory structure of `lib` is changed and the license file is modified to txt file.

**v1.3.0**
```
libcutensor_1.3.0/
├── include/
├── lib/
│   ├── 10.2
│   ├── 11
│   └── 11.0
└── license.txt
```
**v1.2.2**
```
libcutensor_1.2.2/
├── include/
├── lib/
│   ├── 10.1
│   ├── 10.2
│   └── 11
└── license.pdf
```

cupyx/tools/install_library.py
```diff
@@ -217,17 +217,20 @@ def install_lib(cuda, prefix, library):
...
-            if cuda.startswith('11.'):
+            if cuda.startswith('11.') and cuda != '11.0':
                 cuda = '11'
             shutil.move(
                 os.path.join(outdir, 'libcutensor', 'include'),
                 os.path.join(destination, 'include'))
             shutil.move(
                 os.path.join(outdir, 'libcutensor', 'lib', cuda),
                 os.path.join(destination, 'lib'))
+            if cuda == '10.2':
```
kmaehashi, line 228:
```suggestion
            if cuda == '10.1':
```

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):
LGTM!

PR #5200, opened 2021-05-14
add sum_labels to cupyx.scipy.ndimage.measure

### LINKED ISSUE #5111: Rename `scipy.ndimage.sum` to `scipy.ndimage.sum_labels`
SciPy 1.6.0 changed the API: https://github.com/scipy/scipy/pull/12242

grlee77:
closes #5111

This PR renames `sum` to `sum_labels`, but leaves the old function name available as well. This follows what was done upstream in SciPy 1.6.0. I was initially going to add a warning if the old name was used, but it doesn't look like SciPy has one, so I followed the same approach here.

I think this can be backported to 9.x as well since it is not really a new feature, but an API fix.

### DIFF
--- a/cupyx/scipy/ndimage/measurements.py
+++ b/cupyx/scipy/ndimage/measurements.py
@@ -361,7 +361,7 @@ def variance(input, labels=None, index=None):
                                     out) / count
 
 
-def sum(input, labels=None, index=None):
+def sum_labels(input, labels=None, index=None):
     """Calculates the sum of the values of an n-D image array, optionally
        at specified sub-regions.
 
@@ -376,7 +376,7 @@ def sum(input, labels=None, index=None):
        sum (cupy.ndarray): sum of values, for each sub-region if
        `labels` and `index` are specified.
 
-    .. seealso:: :func:`scipy.ndimage.sum`
+    .. seealso:: :func:`scipy.ndimage.sum_labels`
     """
     if not isinstance(input, cupy.ndarray):
         raise TypeError('input must be cupy.ndarray')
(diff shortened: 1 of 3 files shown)


cupyx/scipy/ndimage/measurements.py
```diff
@@ -425,6 +425,31 @@ def sum(input, labels=None, index=None):
     return _ndimage_sum_kernel(input, labels, index, index.size, out)
 
 
+def sum(input, labels=None, index=None):
```
leofang, line 428: Maybe just a nitpick: Can we simply copy `sum_label`'s docstring? Something like
```python
sum.__doc__ = sum_labels.__doc__.replace('scipy.ndimage.sum', 'scipy.ndimage.sum_labels')
```

grlee77: There is a `Note:` section added here, though. The other option is to just omit the parameter and result descriptions and refer users to the docstring of `sum_labels` instead.

leofang: Ohhh I overlooked! OK then it should be alright...

leofang:
Jenkins, test this please

leofang:
@grlee77 We need to skip testing `sum_labels` unless SciPy >= 1.6.0.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -11,10 +11,17 @@
 import cupyx.scipy.ndimage  # NOQA
 
 try:
+    import scipy
     import scipy.ndimage  # NOQA
+    scipy_version = numpy.lib.NumpyVersion(scipy.__version__)
 except ImportError:
     pass
```
leofang, line 18: Maybe we need this? I am sure there's a less hackier solution 😅
```suggestion
    scipy_version = numpy.lib.NumpyVersion('0')
```

leofang:
Jenkins, test this please

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
grlee77: Thanks. I have added it, but with '0.0.0' since '0' is not considered a valid version string by `NumpyVersion`. I used that function because it evaluates things like `numpy.lib.NumpyVersion('1.5.0') > '1.5.0rc1'` to `True` while regular string comparison would not.  Another option is to use `packing.packing.Version` which can also handle cases like this.

I thought adding a list of function names at the top would be better than adding a version check and `pytest.skip` to each individual test method in the `TestStats` class.

leofang: > I thought adding a list of function names at the top would be better than adding a version check and `pytest.skip` to each individual test method in the `TestStats` class.

This makes sense because the `TestStats` class does not subclass `unittest.TestCase` so we can't use `setUp()` to do this for all test methods.

kmaehashi (approved):

kmaehashi:
Thanks @grlee77, LGTM!

PR #5205, opened 2021-05-18
Fix matmul for input with relaxed strides

toslunar:
`_get_stride_for_strided_batched_gemm` assumed that the input `a` has the canonical strides at least in the batch dimensions (`a.shape[:-2]`), which was not true.

This PR computes stride for `<t>gemmStridedBatched` from `a.shape[-2:]` by assuming `a` is C-contiguous in the sense of NumPy>=1.10 (see https://numpy.org/doc/stable/reference/arrays.ndarray.html#index-3 for `NPY_RELAXED_STRIDES_CHECKING`).


kmaehashi:
pfnCI, test this please.

takagi (approved):

Issue #5204, opened 2021-05-18
Need to check compute capability before using cutensor


kmaehashi:
Currently `cupy.einsum` uses cuTENSOR, but cuTENSOR code path should only be used for CC 6.0+.

https://docs.nvidia.com/cuda/cutensor/index.html#support
https://docs.nvidia.com/cuda/cutensor/release_notes.html#cutensor-v1-0-1

anaruse:
Thank you for pointing this out!  It is true that compute capability is not checked in cuTENSOR related code.
Some functions in cuSolver have compute capability restriction, so we check the compute capability before calling those functions (as shown below). I would like to add a code to `cupy/cutensor.pyx` to check compute capability in a similar way. What would you think about this? 

https://github.com/cupy/cupy/blob/9730eac593a5816fe46501f6eaedc1bc3d0d3fce/cupy/cusolver.pyx#L81-L108

kmaehashi:
Thanks! Checking restrictions in `cupy/cutensor.pyx` and calling `check_availability ` from `cupy.einsum` sounds reasonable.

By the way, I noticed that `@_util.memoize()` should be `@_util.memoize(for_each_device=True)` for these function 😅

PR #5207, opened 2021-05-18
Fix `check_availablity` for `cupy.cusolver`

kmaehashi:
Found in discusson of #5204.
Functions that return per-device results should be memorized for each device.


kmaehashi:
pfnCI, test this please.

asi1024 (approved):

PR #4961, opened 2021-03-24
Tentatively pin CI to ROCm 4.0.1

kmaehashi:
ROCm 4.1 is out but seems the build is failing. Tentatively pin to 4.0.1 until we support it.

PR #5002, opened 2021-03-30
Use THRUST_OPTIONAL_CPP11_CONSTEXPR

emcastillo:
Should fix #4984

PR #5209, opened 2021-05-19
Fix thrust compilation for ROCm 4.2.0

takagi:
This fixes the constexpr issue for building CuPy with ROCm 4.2.0, and also reverts https://github.com/cupy/cupy/pull/4961 to make it use the latest ROCm.

### DIFF
--- a/.pfnci/script.sh
+++ b/.pfnci/script.sh
@@ -68,7 +68,7 @@ main() {
       python3.7 -m pip install cython numpy
 
       wget -qO - http://repo.radeon.com/rocm/rocm.gpg.key | apt-key add -
-      echo 'deb [arch=amd64] http://repo.radeon.com/rocm/apt/4.0.1/ xenial main' | tee /etc/apt/sources.list.d/rocm.list
+      echo 'deb [arch=amd64] http://repo.radeon.com/rocm/apt/debian/ xenial main' | tee /etc/apt/sources.list.d/rocm.list
 
       # Uninstall CUDA to ensure it's a clean ROCm environment
       # https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#removing-cuda-tk-and-driver
--- a/cupy/cuda/cupy_thrust.cu
+++ b/cupy/cuda/cupy_thrust.cu
@@ -6,7 +6,7 @@
 #include <thrust/sort.h>
 #include <thrust/tuple.h>
 #include <thrust/execution_policy.h>
-#if (__CUDACC_VER_MAJOR__ >11 || (__CUDACC_VER_MAJOR__ == 11 && __CUDACC_VER_MINOR__ >= 2))
+#if (__CUDACC_VER_MAJOR__ >11 || (__CUDACC_VER_MAJOR__ == 11 && __CUDACC_VER_MINOR__ >= 2) || HIP_VERSION >= 402)
 // This is used to avoid a problem with constexpr in functions declarations introduced in
 // cuda 11.2, MSVC 15 does not fully support it so we need a dummy constexpr declaration
 // that is provided by this header. However optional.h is only available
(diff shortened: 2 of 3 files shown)


takagi:
`__CUDA_ARCH__` equivalent for ROCm/HIP?

leofang:
> Any `__CUDA_ARCH__` equivalent for ROCm/HIP?

If you meant to enclose device-only code, use __HIP_DEVICE_COMPILE__. 

However, I think rocThrust might have solved some constexpr issues following Thrust (if that's the same constexpr problem that Emilio solved before). Could you check if extending line 14 (include thrust/optional.h) to ROCm/HIP solves the issue? If so it might be a simpler solution.

takagi:
Thanks! I tried extending line 14 but it didn't solve the issue. I'll go as you mentioned on gitter.

leofang:
> Thanks! I tried extending line 14 but it didn't solve the issue. I'll go as you mentioned on gitter.

Wait, to use this we need to use `THRUST_OPTIONAL_CPP11_CONSTEXPR` instead of raw `constexpr`. I thought #5002 fixed them all and didn't realize we still have lingering `constexpr` until now...

takagi:
Hmm... the build still failes and it seems that `THRUST_OPTIONAL_CPP11_CONSTEXPR` is not expanded to `constexpr`. Would you look at the change to see if I can catch your point?
```
    cupy/cuda/cupy_thrust.cu:148:28: error: non-constexpr declaration of 'operator()' follows constexpr declaration
    bool less<complex<float>>::operator() (
                               ^
    /opt/rocm/include/thrust/functional.h:765:18: note: previous declaration is here
      constexpr bool operator()(const T &lhs, const T &rhs) const
                     ^
```

leofang:
On my cell, so obviously I do not test building it myself 😅 

I think we need to build with --std=c++14. The macro is defined to be empty for c++11 (which is the default of hipcc IIRC). I think it can be added to this line 
https://github.com/cupy/cupy/blob/3152723c6454070fbaa099d1dd6cf1c2a940653e/cupy_setup_build.py#L1064
We did the same switch for the nvcc counterpart in a few lines above.

takagi:
Oh, I didn't know that the cpp standard used is different for CUDA >= 11.2

takagi:
It works fine with `--std=c++14`!

kmaehashi:
We need to make sure that `--std=c++14` can be built on CentOS 7 which is an environment to build a wheel.

leofang:
> We need to make sure that `--std=c++14` can be built on CentOS 7 which is an environment to build a wheel.

Yes, on Conda-Forge we use cos7 for CUDA 11.0+ and some 10.x versions. It works fine.

leofang (approved):
LGTM!

leofang:
Let's give the CI a shot.

Jenkins, test this please

leofang:
AMD build passed!

emcastillo (approved):

leofang:
Should this be backported?

leofang:
We are back to the game! Thanks guys!

emcastillo:
I suspect this change is breaking windows CI
https://github.com/cupy/cupy-release-tools/pull/110

kmaehashi:
I guess this commit is not needed? 5282dafbfaa31f3489ad6bf1afb9401c14d4675c

emcastillo:
#5233  fixes both thrust and windows
constexpr does not work the same way in c++11 and 14 and mscv 15 has no full c++14 support
so we need hacks like this 😭😭

leofang:
> I guess this commit is not needed? [5282daf](https://github.com/cupy/cupy/commit/5282dafbfaa31f3489ad6bf1afb9401c14d4675c)

It's needed for ROCm.

PR #5226, opened 2021-05-20
Remove syncdetect APIs

### LINKED ISSUE #2946: Remove cupyx.allow_synchronize?
`cupyx.allow_synchronize` (#2808) was introduced originally for both unit tests and users who want to notice device synchronization in their own code.

Application in uint tests was dissmissed (#2893) because much more tests were synchronous than I had expected. 

Now I doubt the usefulness of this feature for users as well.
It's fundamentally impossible to eliminate false positives and/or false negatives.

If we took a policy for zero false positives, there would be too many false negatives which I think would make this feature useless.
For example, the documentation of [cudaMemcpy](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gc263dbe6574220cc776b45438fc351e8) says *"This function exhibits synchronous behavior for most use cases"*. If we took this policy, we wouldn't be able to consider this function synchronous, because no condition is mentioned that would make this function asynchronous.

If we took a policy to allow some false positives, this feature wouldn't be used to detect unexpected synchronization in users' code.

kmaehashi:
Closes #2946

### DIFF
--- a/cupy/_core/syncdetect.py
+++ b/cupy/_core/syncdetect.py
@@ -2,8 +2,6 @@ import contextlib
 import threading
 import warnings
 
-from cupy import _util
-
 
 _thread_local = threading.local()
 
--- a/cupy/_core/syncdetect.py
+++ b/cupy/_core/syncdetect.py
@@ -1,5 +1,6 @@
 import contextlib
 import threading
+import warnings
 
 from cupy import _util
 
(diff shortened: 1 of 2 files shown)


leofang:
This PR issues warnings. I assume it's because you wanna backport this PR, and do the actual removal in a separate one (which would not be backported)?

emcastillo:
Jenkins, test this please

kmaehashi:
> This PR issues warnings. I assume it's because you wanna backport this PR, and do the actual removal in a separate one (which would not be backported)?

Deprecation warnings will not be backported (actual removal will happen in v11 or later), as per our deprecation policy:
https://docs.cupy.dev/en/stable/user_guide/compatibility.html#deprecation-dropping-and-its-preparation

emcastillo (approved):

PR #5233, opened 2021-05-22
Fix windows constexpr

emcastillo:
constexpr does not work the same way in c++11 and 14 and mscv 15 has no full c++14 support
so we need hacks like this 😭😭


asi1024:
Jenkins, test this please.

emcastillo:
Jenkins, test this please

asi1024 (approved):

PR #5237, opened 2021-05-23
Fix per device stream

takagi:
Following up #5172.

(To be described soon)

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -12,25 +12,27 @@ cdef object _thread_local = threading.local()
 
 
 cdef class _ThreadLocal:
+    # TODO(takagi): Update the following comment
     # We keep both current_stream_ref and current_stream_stack because the
     # former is also used when calling stream.use(). This bookkeeping enables
     # correct rewinding when "with" blocks are mixed with ".use()" (though
     # this is considered an anti-pattern). As the name suggested, stream
     # lifetime is not tracked in current_stream_ref.
 
-    cdef list current_stream_ref  # list of object
     cdef list current_device_id_stack  # list of int
-    cdef list current_stream_stack  # list of list
+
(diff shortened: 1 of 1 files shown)


takagi:
I checked that this change satisfies the following cases relating to #5172.

Case 1: #5143
```Python
import gc
from cupy import cuda

with cuda.Stream() as s1:
    s2 = cuda.Stream()
    s2.use()
    with cuda.Stream() as s3:
        del s2
        gc.collect()
    assert cuda.get_current_stream() == s1
    print('Ok.')
```

Case 2: https://github.com/cupy/cupy/pull/5172#issuecomment-845764323
```Python
from cupy import cuda

s1 = cuda.Stream()
s2 = cuda.Stream()
s2.use()
with s1:
    pass
assert cuda.get_current_stream() == s2
print('Ok.')
```

Case 3: Rewind to the context stream when `use()`d is deleted
```Python
import gc
from cupy import cuda

with cuda.Stream() as s1:
    s2 = cuda.Stream()
    s2.use()
    del s2
    gc.collect()
    assert cuda.get_current_stream() == s1
    print('Ok.')
```

Case 4: ... but, it does not mean `use()` remembers the current stream when it has been called
```Python
import gc
from cupy import cuda

with cuda.Stream() as s1:
    s2 = cuda.Stream()
    s2.use()
    s3 = cuda.Stream()
    s3.use()
    del s3
    gc.collect()
    assert cuda.get_current_stream() == s1
    print('Ok.')
```

Case 5: A stream context remembers a `use()`d current stream even when another is `use()`d in it
```Python
from cupy import cuda

with cuda.Stream() as s1:
    s2 = cuda.Stream()
    s2.use()
    with cuda.Stream() as s3:
        s4 = cuda.Stream()
        s4.use()
    assert cuda.get_current_stream() == s2
    print('Ok.')
```

PR #5224, opened 2021-05-20
Test build for ROCm 4.0 and latest

kmaehashi:

PR #5248, opened 2021-05-24
Fix mergify configuration

kmaehashi:
Follow-up #5224. As ROCm CI has been renamed, mergify config needs to be updated.

Actually, Mergify respects the branch protections settigns configured in GitHub (i.e., `Required` signs of pull-request statuses), so nothing needs to be listed in the configuration.
However I'd like to keep `pretest` and `Jenkins` here for fail-safe.


emcastillo (approved):

emcastillo:
merging as it doesnt need cis

PR #5251, opened 2021-05-24
Fix to allow sharing a Stream instance between threads

kmaehashi:
Currently, CuPy uses weakref to hold a reference to the current stream.
However this does not work when there are two threads sharing the same stream instance and both thread frees the reference to it, because the dtor `__del__` only runs in one of the thread so the current stream of the other thread will not be reset.

The reproducer code is added in the test code, which fails in the current master and succeeds after this PR got merged.

This PR is backward-incompatible in terms of the life-cycle of the Stream. In the current master, the stream will be destructed when a user frees the reference:

```py
s1.use()
del s1
# s1 is destructed
```

After this PR, the stream will be freed when changing to another stream:

```py
s1.use()
del s1
# s1 will not be destructed here because TLS holds a hard reference.

Stream.null.use()  # s1 is destructed
# -- or --
with Stream() as another_stream:
    # s1 is destructed
    pass
```


kmaehashi:
pfnCI, test this please.

kmaehashi:
cc/ @leofang

kmaehashi:
pfnCI, test this please.

leofang:
I just checked that this PR does not fix https://github.com/cupy/cupy/pull/5172#issuecomment-845764323 (which I am fine with), is it intended @kmaehashi?

kmaehashi:
@leofang Sorry I forgot to comment on that issue. I understood and agree with you on this point:
> I think it's fundamentally incompatible unless we push the stream to the stack unconditionally (both with and .use())

and your design looks reasonable to me.

I talked with @takagi regarding the "shadowing with the alternate stream" idea proposed in #5237 but it seems impossible to solve the issue mentioned in this PR as it still uses weakref/`__del__` architecture.

tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -60,10 +63,12 @@ def test_eq_hip(self):
 
     def check_del(self, null, ptds):
         stream = cuda.Stream(null=null, ptds=ptds).use()
+        assert stream is cuda.get_current_stream()
         stream_ptr = stream.ptr
         x = from_data.array([1, 2, 3])
         del stream
-        assert cuda.Stream.null == cuda.get_current_stream()
+        cuda.Stream.null.use()
```
leofang, line 70: Should we test that the stream is still alive after deletion?
```suggestion
        assert stream is cuda.get_current_stream()
        cuda.Stream.null.use()
```

leofang:
Is this description still accurate if a `Stream` can be shared among threads?
https://github.com/cupy/cupy/blob/cd831e96a201cd923ab7b3aff6013307aec76d66/docs/source/user_guide/basic.rst#L127

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

takagi (dismissed):

takagi:
pfnCI, test this please.

takagi (approved):

PR #5260, opened 2021-05-25
Adding GUFunc order, dtype and casting kwarg support

emcastillo:
This mostly completes the support for GUFunc adding the same core characteristics than numpy

### DIFF
--- a/cupy/_core/_gufuncs.py
+++ b/cupy/_core/_gufuncs.py
@@ -345,8 +345,8 @@ class _GUFunc:
 
         return args, dimsizess, loop_output_dims, outs, missing_dims
 
-    def _can_cast(d1, d2, casting):
-        if casting == 'same_kind' and get_dtype(d1).kind == d2.kind:
+    def _can_cast(self, d1, d2, casting):
+        if casting == 'same_kind' and get_dtype(d1).kind == get_dtype(d2).kind:
             return True
         return numpy.can_cast(d1, d2, casting=casting)
 
@@ -409,6 +409,7 @@ class _GUFunc:
         signature = kwargs.pop('signature', None)
         casting = kwargs.pop('casting', 'same_kind')
 
+        ret_dtype = None
         if signature is None:
             if dtype is not None:
                 if type(dtype) == tuple:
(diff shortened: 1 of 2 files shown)


cupy/_core/_gufuncs.py
```diff
@@ -141,6 +142,167 @@ def _validate_normalize_axes(
...
+                    'is not consistent with general signature')
+            if len(outs) != self._nout:
+                raise ValueError(
+                    f'signature {sig} for dtypes is invalid number of inputs '
+                    'is not consistent with general signature')
+
+            ops.append(_OpsRegister._Op(ins, outs, op))
+        return ops
+
+    def _can_cast(self, d1, d2, casting):
```
asi1024, line 201: In what case does this function behave differently from `numpy.can_cast` itself?

cupy/_core/_gufuncs.py
```diff
@@ -141,6 +142,167 @@ def _validate_normalize_axes(
...
+
+    def _sig_str_to_tuple(self, sig):
+        sig = sig.replace(' ', '')
+        toks = sig.split('->')
+        if sig.startswith('->'):
+            ins = ''
+            outs = toks
+        elif sig.endswith('->'):
+            ins = toks
+            outs = ''
```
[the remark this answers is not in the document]
emcastillo: yeah, sorry about m stupidity here

cupy/_core/_gufuncs.py
emcastillo: Its the same but with a shortcut, this was taken from kernel.pyx iirc

emcastillo:
Added tests for both order and signatures and fixed some quirks with # of outputs.

PTAL!

I will add tests for the rest of features (I did it before but launched a git clean before adding the file :() and expose gufunc to the user via cupyx in a different PR.

asi1024:
Jenkins, test this please.

emcastillo:
failure is not related :)

asi1024:
I confirmed that `test_gufunc.py` ran without errors in windows CI. LGTM.

emcastillo:
Thank you!

PR #5263, opened 2021-05-26
Fix threading memory pool tests

### LINKED ISSUE #5252: Invalid use of `threading.Barrier` in test code
`threading.Barrier` needs to be instantiated.

https://github.com/cupy/cupy/blob/v9.0.0/tests/cupy_tests/cuda_tests/test_memory.py#L774-L796
https://docs.python.org/3/library/threading.html#barrier-objects

Ref: #2627

kmaehashi:
Closes #5252.


emcastillo (approved):

emcastillo:
Jenkins, test this please

PR #5267, opened 2021-05-26
Update list of wheels in README

kmaehashi:


emcastillo (approved):
LGTM!, merging as it does not require triggering CIs

PR #5274, opened 2021-05-27
Use cdef instead of cpdef where appropraite

kmaehashi:
This eliminates the following warnings:

```
cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables
```


leofang (approved):

asi1024:
Jenkins, test this please.

asi1024 (approved):

Issue #4983, opened 2021-03-26
`distutils` deprecation (Python 3.10 support)


kmaehashi:
`distutils` is deprecating in Python 3.10 so we should get ready for it.
https://www.python.org/dev/peps/pep-0632/
https://github.com/scipy/scipy/issues/13615

Thanks @leofang for letting us know 😄

leofang:
SciPy has been successfully built with the new build system Meson: https://labs.quansight.org/blog/2021/07/moving-scipy-to-meson/

kmaehashi:
Python 3.10.0 stable release is planned on 2021-10-04.

Currently, most of distutils hacks in `cupy_setup_build.py` is to run `nvcc` from within the Extension build process.
I'm currently assuming we can avoid (direct) use of distutils only by changing the build order as follows:

1. Compile `*.cu` files as CUDA object files using NVCC. This is done by our script (maybe in `custom_build_ext`).
2. Pass the above object files to `extra_objects` argument of `setuptools.Extension`.
3. Run `setuptools.setup()` and let setuptools build Python modules and link CUDA object files.

I feel using an external build system is overkill for step 1., as it is quite simple. The only thing we have to implement on our own is to compare the timestamp between `*.cu` and `*.o` to avoid building it when the file is not modified.

kmaehashi:
* Use of `distutils` has been removed from CuPy modules (`cupy`, `cupyx` and `cupy_backends`) in #6025.
* https://github.com/cupy/cupy/issues/4983#issuecomment-915357337 has been implemented in #6273.

The remaining `distutils` usages are in the installer. So far they have no replacement in setuptools.

* `install/cupy_builder/_compiler.py`: `distutils.ccompiler.gen_preprocess_options` -> Very simple logic, so I think it's better to implement it on our own.
* `install/cupy_builder/cupy_setup_build.py`: `distutils.{ccompiler,sysconfig}` -> Maybe keep it as-is...
* `install/cupy_builder/install_build.py`: `distutils.util.split_quoted` -> Rewrite with `shlex`

PR #5292, opened 2021-05-28
Support detecting cl.exe in vs2019

eternalphane:
Use `setuptools.msvc` instead of `distutils.msvc9compiler` to detect cl.exe

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -96,6 +96,7 @@ def _get_extra_path_for_msvc():
         return None
 
     from setuptools import msvc
+    import platform
     try:
         vctools = msvc.EnvironmentInfo(platform.machine().lower()).VCTools
     except distutils.errors.DistutilsPlatformError:
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -89,22 +89,23 @@ def _run_cc(cmd, cwd, backend, log_stream=None):
 
 @_util.memoize()
 def _get_extra_path_for_msvc():
-    import distutils.spawn
-    cl_exe = distutils.spawn.find_executable('cl.exe')
+    import shutil
+    cl_exe = shutil.which('cl.exe')
     if cl_exe:
         # The compiler is already on PATH, no extra path needed.
         return None
 
     from setuptools import msvc
+    from setuptools.errors import DistutilsError
     import platform
     try:
         vctools = msvc.EnvironmentInfo(platform.machine().lower()).VCTools
-    except distutils.errors.DistutilsPlatformError:
+    except DistutilsError:
         # Failed to find VC.
         return None
 
     path = vctools[-1]
-    if not distutils.spawn.find_executable('cl.exe', path):
+    if not shutil.which('cl.exe', path):
         # The compiler could not be found.
         return None
     return path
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thanks @eternalphane!

It looks nice to discover VS2019, however I'm concerning that CUDA 10.0, which is still in our supported CUDA version, does not support VS2019.
https://docs.nvidia.com/cuda/archive/10.0/cuda-installation-guide-microsoft-windows/index.html
Let me try your patch on CUDA 10.0 environment later.

leofang (changes requested):
Thanks for your patch. I do not have access to any Windows CUDA env, so I cannot verify the patch and I think @kmaehashi and others from the team will help. But, for the purpose of #4983 I think this PR is necessary for us to move away from depending `distutils` at runtime, so I am happy to see this change!

With the above background stated, @eternalphane could you try replacing all `distutils` occurrences in this file (likely only contained in the function you're touching) by their `setuptools` counterparts and help us verify the change please? I think PEP 632 should be clear in the needed replacements so it should be rather straightforward.

cupy/cuda/compiler.py
```diff
@@ -89,21 +89,23 @@ def _run_cc(cmd, cwd, backend, log_stream=None):
 
 @_util.memoize()
 def _get_extra_path_for_msvc():
-    import distutils.spawn
-    cl_exe = distutils.spawn.find_executable('cl.exe')
+    import shutil
```
kmaehashi, line 92: Could you move this import (and `import platform` too --- as they're standard modules) to the top of the file?

cupy/cuda/compiler.py
```diff
@@ -89,21 +89,23 @@ def _run_cc(cmd, cwd, backend, log_stream=None):
...
     if cl_exe:
         # The compiler is already on PATH, no extra path needed.
         return None
 
-    from distutils import msvc9compiler
-    vcvarsall_bat = msvc9compiler.find_vcvarsall(
-        msvc9compiler.get_build_version())
-    if not vcvarsall_bat:
+    from setuptools import msvc
+    from setuptools.errors import DistutilsError
```
kmaehashi, line 99: It seems `setuptools.errors` is only available in setuptools v42.0+. In my environment, the default one that comes with Python 3.8 for Windows was 41.2.0, so this line failed. Until we decide a policy on which setuptools versions to support (we should do this once we fully migrate from distutils), I'm negative to use this module. So, could you use `distutils.errors.DistutilsPlatformError` instead, as in your original code? :pray:

cupy/cuda/compiler.py
```diff
@@ -89,21 +89,23 @@ def _run_cc(cmd, cwd, backend, log_stream=None):
...
+    try:
+        vctools = msvc.EnvironmentInfo(platform.machine().lower()).VCTools
+    except DistutilsError:
         # Failed to find VC.
         return None
 
-    path = os.path.join(os.path.dirname(vcvarsall_bat), 'bin')
-    if not distutils.spawn.find_executable('cl.exe', path):
+    path = vctools[-1]
+    if not shutil.which('cl.exe', path):
```
kmaehashi, line 108:
```suggestion
    if not shutil.which('cl.exe', path=path):
```

kmaehashi (changes requested):

kmaehashi:
Could you fix the comments here as well?

https://github.com/cupy/cupy/blob/b01641dedb065f65a4d3f846cf3c6222693a8e44/cupy/cuda/compiler.py#L54-L56

`the VC++ compiler used to build Python / CuPy to the PATH` -> `the latest VC++ compiler found via setuptools`

kmaehashi:
I've tookover the work in #6025.

Issue #5280, opened 2021-05-27
Support more CUDA API in CuPy JIT


asi1024:
Part of #4290 

* Math modules: https://docs.nvidia.com/cuda/cuda-math-api/
  - [ ] math module functions
  - [x] cupy ufunc
* Atomic Functions: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions
  - [x] atomicAdd
  - [ ] atomicSub
  - [ ] atomicExch
  - [ ] atomicMin
  - [ ] atomicMax
  - [ ] atomicInc
  - [ ] atomicDec
  - [ ] atomicCAS
  - [ ] atomicAnd
  - [ ] atomicOr
  - [ ] atomicXor
* Warp intrinsics: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-shuffle-functions
  - [ ] Warp shuffle functions
  - [ ] Warp match functions
  - [ ] Warp reduce functions

leofang:
This is nice. I am interested in implementing the warp functions, the new async functions ([Asynchronous Barrier](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#aw-barrier), [Asynchronous Data Copies](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#memcpy_async)), and [alloca](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#alloca-function), though I need to check my recent work schedule and plan accordingly.

leofang:
xref: https://numba.readthedocs.io/en/stable/cuda-reference/kernel.html

leofang:
> the new async functions ([Asynchronous Barrier](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#aw-barrier), [Asynchronous Data Copies](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#memcpy_async))

I had an incorrect understanding. Looks like to support these two features, we need to support cooperative groups first. Maybe I could check out [how Numba did it](https://numba.readthedocs.io/en/stable/cuda-reference/kernel.html#cooperative-groups)...

leofang:
> CuPy JIT specific interface

The drop-in replacement of `getsource` (#5071) should also be included/documented.

leofang:
> Looks like to support these two features, we need to support cooperative groups first. 

[...]

PR #5293, opened 2021-05-28
Add `len`, `min`, `max`, `.ndim`, `.size`, `.shape` and `.strides` to jit

eternalphane:
Part of #4290 and #5280
- Built-in functions
  - [x] len
- CUDA Math API
  - [x] min
  - [x] max
- Array size functions
  - [x] `ArrayBase.ndim`
  - [x] `CArray.size`
  - [x] `CArray.shape`
  - [x] `CArray.strides`
- Misc
  - [x] support tuple indexing
  - [x] support unpacking `Ptr` (like `n_rows, n_cols = x.shape`)

### DIFF
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -53,7 +53,7 @@ class RangeFunc(BuiltinFunc):
 
 class Len(BuiltinFunc):
     def call(self, env, *args, **kwds):
-        if len(args) > 1:
+        if len(args) != 1:
             raise TypeError(f'len expects only 1 argument, got {len(args)}')
         arg = args[0]
         if isinstance(arg.ctype, _cuda_types.CArray):
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -1,11 +1,14 @@
+import cupy
+
 from cupy_backends.cuda.api import runtime
 from cupyx.jit import _cuda_types
-from cupyx.jit._cuda_typerules import implicit_conversion
 from cupyx.jit._internal_types import BuiltinFunc
 from cupyx.jit._internal_types import Data
 from cupyx.jit._internal_types import Constant
 from cupyx.jit._internal_types import Range
 from cupyx.jit import _compile
+
+from functools import reduce
 
 
 class RangeFunc(BuiltinFunc):
@@ -68,12 +71,8 @@ class Min(BuiltinFunc):
                 f'min expects at least 2 arguments, got {len(args)}')
         if kwds:
             raise TypeError('keyword arguments are not supported')
-        args = [Data.init(arg, env) for arg in args]
-        codes, ctypes = zip(*[(arg.code, arg.ctype) for arg in args])
-        code = 'min(' * (len(codes) - 1) + codes[0]
-        code += ''.join(f', {code})' for code in codes[1:])
-        ctype = implicit_conversion(*ctypes)
-        return Data(code, ctype)
+        return reduce(lambda a, b: _compile._call_ufunc(
+            cupy.minimum, (a, b), None, env), args)
 
 
 class Max(BuiltinFunc):
@@ -83,12 +82,8 @@ class Max(BuiltinFunc):
                 f'max expects at least 2 arguments, got {len(args)}')
         if kwds:
             raise TypeError('keyword arguments are not supported')
-        args = [Data.init(arg, env) for arg in args]
-        codes, ctypes = zip(*[(arg.code, arg.ctype) for arg in args])
-        code = 'max(' * (len(codes) - 1) + codes[0]
-        code += ''.join(f', {code})' for code in codes[1:])
-        ctype = implicit_conversion(*ctypes)
-        return Data(code, ctype)
+        return reduce(lambda a, b: _compile._call_ufunc(
+            cupy.maximum, (a, b), None, env), args)
 
 
 class SyncThreads(BuiltinFunc):
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -56,13 +56,9 @@ class Len(BuiltinFunc):
         if len(args) != 1:
             raise TypeError(f'len expects only 1 argument, got {len(args)}')
         arg = args[0]
-        if isinstance(arg.ctype, _cuda_types.CArray):
-            return Data(f'{arg.code}.size()', _cuda_types.PtrDiff())
-        if isinstance(arg.ctype, _cuda_types.SharedMem):
-            if arg.ctype._size is None:
-                raise TypeError('unsized shared memory is not supported')
-            return Constant(arg.ctype._size)
-        raise TypeError('len supports only array type')
+        if not isinstance(arg.ctype, _cuda_types.CArray):
+            raise TypeError('len supports only array type')
+        return Data(f'{arg.code}.size()', _cuda_types.PtrDiff())
 
 
 class Min(BuiltinFunc):
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -59,6 +59,8 @@ class Len(BuiltinFunc):
         if isinstance(arg.ctype, _cuda_types.CArray):
             return Data(f'{arg.code}.size()', _cuda_types.PtrDiff())
         if isinstance(arg.ctype, _cuda_types.SharedMem):
+            if arg.ctype._size is None:
+                raise TypeError('unsized shared memory is not supported')
             return Constant(arg.ctype._size)
         raise TypeError('len supports only array type')
 
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -1,5 +1,6 @@
 from cupy_backends.cuda.api import runtime
 from cupyx.jit import _cuda_types
+from cupyx.jit._cuda_typerules import implicit_conversion
 from cupyx.jit._internal_types import BuiltinFunc
 from cupyx.jit._internal_types import Data
 from cupyx.jit._internal_types import Constant
@@ -48,6 +49,48 @@ class RangeFunc(BuiltinFunc):
             assert False
 
         return Range(start, stop, step, ctype, step_is_positive)
+
+
+class Len(BuiltinFunc):
+    def call(self, env, *args, **kwds):
+        if len(args) > 1:
+            raise TypeError(f'len expects only 1 argument, got {len(args)}')
+        arg = args[0]
+        if isinstance(arg.ctype, _cuda_types.CArray):
+            return Data(f'{arg.code}.size()', _cuda_types.PtrDiff())
+        if isinstance(arg.ctype, _cuda_types.SharedMem):
+            return Constant(arg.ctype._size)
+        raise TypeError('len supports only array type')
+
+
+class Min(BuiltinFunc):
+    def call(self, env, *args, **kwds):
+        if len(args) < 2:
+            raise TypeError(
+                f'min expects at least 2 arguments, got {len(args)}')
+        if kwds:
+            raise TypeError('keyword arguments are not supported')
+        args = [Data.init(arg, env) for arg in args]
+        codes, ctypes = zip(*[(arg.code, arg.ctype) for arg in args])
+        code = 'min(' * (len(codes) - 1) + codes[0]
+        code += ''.join(f', {code})' for code in codes[1:])
+        ctype = implicit_conversion(*ctypes)
+        return Data(code, ctype)
+
+
+class Max(BuiltinFunc):
+    def call(self, env, *args, **kwds):
+        if len(args) < 2:
+            raise TypeError(
+                f'max expects at least 2 arguments, got {len(args)}')
+        if kwds:
+            raise TypeError('keyword arguments are not supported')
+        args = [Data.init(arg, env) for arg in args]
+        codes, ctypes = zip(*[(arg.code, arg.ctype) for arg in args])
+        code = 'max(' * (len(codes) - 1) + codes[0]
+        code += ''.join(f', {code})' for code in codes[1:])
+        ctype = implicit_conversion(*ctypes)
+        return Data(code, ctype)
 
 
 class SyncThreads(BuiltinFunc):
@@ -111,6 +154,9 @@ class AtomicOp(BuiltinFunc):
 
 builtin_functions_dict = {
     range: RangeFunc(),
+    len: Len(),
+    min: Min(),
+    max: Max(),
 }
 
 syncthreads = SyncThreads()
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -23,6 +23,13 @@ class RangeFunc(BuiltinFunc):
             raise TypeError(
                 f'range expected at most 3 argument, got {len(args)}')
 
+        if isinstance(step, Constant):
+            step_is_positive = step.obj >= 0
+        elif step.ctype.dtype.kind == 'u':
+            step_is_positive = True
+        else:
+            step_is_positive = None
+
         stop = Data.init(stop, env)
         start = Data.init(start, env)
         step = Data.init(step, env)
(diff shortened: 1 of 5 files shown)


kmaehashi:
Thank you @eternalphane, this would be a great addition to JIT!
@asi1024, can you take a look?

cupyx/jit/_builtin_funcs.py
```diff
@@ -50,6 +51,50 @@ def call(self, env, *args, **kwargs):
...
+    def call(self, env, *args, **kwds):
+        if len(args) > 1:
+            raise TypeError(f'len expects only 1 argument, got {len(args)}')
+        arg = args[0]
+        if isinstance(arg.ctype, _cuda_types.CArray):
+            return Data(f'{arg.code}.size()', _cuda_types.PtrDiff())
+        if isinstance(arg.ctype, _cuda_types.SharedMem):
+            if arg.ctype._size is None:
+                raise TypeError('unsized shared memory is not supported')
+            return Constant(arg.ctype._size)
```
asi1024, line 64: The variable of SharedMem type is not exposed to the users, because the return value of `shared_memory()` is converted to Ptr type.
https://github.com/cupy/cupy/blob/1af3300bfaa224d5f0da8f8a2a0daf91f9c406f8/cupyx/jit/_builtin_funcs.py#L123-L129

cupyx/jit/_builtin_funcs.py
```diff
@@ -50,6 +51,50 @@ def call(self, env, *args, **kwargs):
...
+        raise TypeError('len supports only array type')
+
+
+class Min(BuiltinFunc):
+    def call(self, env, *args, **kwds):
```
asi1024, line 69: How about calling `_compile._call_ufunc(cupy.minimum, args, None, env)` for simplification?

cupyx/jit/_builtin_funcs.py
```diff
@@ -50,6 +51,50 @@ def call(self, env, *args, **kwargs):
...
+            raise TypeError('keyword arguments are not supported')
+        args = [Data.init(arg, env) for arg in args]
+        codes, ctypes = zip(*[(arg.code, arg.ctype) for arg in args])
+        code = 'min(' * (len(codes) - 1) + codes[0]
+        code += ''.join(f', {code})' for code in codes[1:])
+        ctype = implicit_conversion(*ctypes)
+        return Data(code, ctype)
+
+
+class Max(BuiltinFunc):
```
asi1024, line 83: Ditto.

cupyx/jit/_compile.py
```diff
@@ -562,6 +562,21 @@ def _transpile_expr_internal(expr, env):
...
+                return _builtin_funcs.Len().call(value)
+        if isinstance(value.ctype, _cuda_types.CArray):
+            if 'size' == expr.attr:
+                return _builtin_funcs.Len().call(value)
+            if 'shape' == expr.attr:
+                ctype = _cuda_types.Ptr(_cuda_types.PtrDiff())
+                return Data(f'{value.code}.shape()', ctype)
+            if 'strides' == expr.attr:
+                ctype = _cuda_types.Ptr(_cuda_types.PtrDiff())
+                return Data(f'{value.code}.strides()', ctype)
```
asi1024, line 579: I'm wondering if it causes compilation error because `CArray.strides()` returns `const ptrdiff_t*` value.

asi1024: IMO, `Tuple(PtrDiff, ...)` is preferred as the return type because the shapes and the strides should not be rewritable. How about adding a new member function to `CArray` that returns a tuple value (like as `_indexing` function)?

asi1024 (commented):
Thanks for your PR! Could you add unit tests to `tests/cupyx_tests/jit_tests/test_raw.py` for the new features?

cupyx/jit/_builtin_funcs.py
eternalphane: You are right. I overlooked that part...

cupyx/jit/_builtin_funcs.py
eternalphane: I suppose `cupy.maximum` only support two arguments... or you mean calling it for multiple times?

asi1024: Yes! I mean calling `cupy.maximum` n-1 times.

eternalphane: So we prefer ufunc to math api?

cupyx/jit/_compile.py
eternalphane: Does this look good to you? I can't test on my machine currently...
<details>
<summary>diff</summary>

```diff
diff --git a/cupy/_core/include/cupy/carray.cuh b/cupy/_core/include/cupy/carray.cuh
index 54d0a000b..81a1f281c 100644
--- a/cupy/_core/include/cupy/carray.cuh
+++ b/cupy/_core/include/cupy/carray.cuh
@@ -8,22 +8,7 @@
 #endif

 // Basic implementation of std::type_traits
-// We use this regardless when C++ is requested, as NVRTC by default lacks many
-// C++ features like this. We need to wrap in a namespace in case Jitify kicks
-// in and/or users provide custom definitions.
-namespace cupy {
-  namespace type_traits {
-    template<bool B, class T, class F>
-    struct conditional { typedef T type; };
-    template<class T, class F>
-    struct conditional<false, T, F> { typedef F type; };
-
-    template<bool B, class T = void>
-    struct enable_if {};
-    template<class T>
-    struct enable_if<true, T> { typedef T type; };
-  }
-}
+#include <cupy/type_traits.h>

 // math
 #ifndef M_PI
@@ -201,11 +186,42 @@ __device__ int signbit(float16 x) {return x.signbit();}
 #ifdef CUPY_JIT_MODE
 #include <cupy/swap.cuh>
 #include <cupy/tuple.cuh>
+#include <cupy/utility.h>

 template <int dim>
 struct Dim {
   __device__ Dim() {}
 };
+
+namespace helper {
+
+namespace detail {
+
+template <class T, class Seq>
+struct array_tuple_;
+template <class T, size_t... Is>
+struct array_tuple_<T, cupy::utility::index_sequence<Is...>> {
+  template <size_t>
+  using elem_type = T;
+  using type = thrust::tuple<elem_type<Is>...>;
+  static const type from(const T (&array)[sizeof...(Is)]) {
+    return thrust::make_tuple(array[Is]...);
+  }
+};
+
+} // namespace detail
+
+template <class T, size_t N>
+using array_tuple = typename detail::array_tuple_<T, cupy::utility::make_index_sequence<N>>::type;
+
+template <class T, size_t N, class R = detail::array_tuple_<
+  T, cupy::utility::make_index_sequence<N>
+>>
+typename R::type as_tuple(const T (&array)[N]) {
+  return R::from(array);
+}
+
+} // namespace helper
 #endif

 template <typename T, int _ndim, bool _c_contiguous=false, bool _use_32bit_indexing=false>
@@ -221,6 +237,10 @@ private:
   ptrdiff_t shape_[ndim];
   ptrdiff_t strides_[ndim];

+#ifdef CUPY_JIT_MODE
+  using tuple_t = helper::array_tuple<ptrdiff_t, _ndim>;
+#endif // ifdef CUPY_JIT_MODE
+
 public:
   // Constructor supports pointers or initializer lists and strides is optional
   // as long as _c_contiguous=true.
@@ -361,6 +381,14 @@ public:
   __forceinline__ __device__ T& _indexing(const Tuple &idx) {
     return const_cast<T&>(const_cast<const CArray&>(*this)._indexing(idx));
   }
+
+  __forceinline__ __device__ tuple_t _shape() const {
+    return helper::as_tuple(shape_);
+  }
+
+  __forceinline__ __device__ tuple_t _strides() const {
+    return helper::as_tuple(strides_);
+  }
 #endif

   __device__ const T& operator[](ptrdiff_t idx) const {
diff --git a/cupy/_core/include/cupy/type_traits.h b/cupy/_core/include/cupy/type_traits.h
new file mode 100644
index 000000000..90adcca6f
--- /dev/null
+++ b/cupy/_core/include/cupy/type_traits.h
@@ -0,0 +1,29 @@
+/*
+ * Basic implementation of std::type_traits
+ * We use this regardless when C++ is requested, as NVRTC by default lacks many
+ * C++ features like this. We need to wrap in a namespace in case Jitify kicks
+ * in and/or users provide custom definitions.
+ */
+
+namespace cupy {
+  namespace type_traits {
+    template<bool B, class T, class F>
+    struct conditional { typedef T type; };
+    template<class T, class F>
+    struct conditional<false, T, F> { typedef F type; };
+
+    template<bool B, class T = void>
+    struct enable_if {};
+    template<class T>
+    struct enable_if<true, T> { typedef T type; };
+
+    template <class T, T v>
+    struct integral_constant {
+      static const T value = v;
+      using value_type = T;
+      using type = integral_constant;
+      const operator value_type() const noexcept { return value; }
+      const value_type operator()() const noexcept { return value; }
+    };
+  }
+}
diff --git a/cupy/_core/include/cupy/utility.h b/cupy/_core/include/cupy/utility.h
new file mode 100644
index 000000000..d450ffb23
--- /dev/null
+++ b/cupy/_core/include/cupy/utility.h
@@ -0,0 +1,69 @@
+#include <cupy/type_traits.h>
+
+namespace cupy {
+
+namespace utility {
+
+template <class T, T... Ints>
+struct integer_sequence {
+  using value_type = T;
+  static const size_t size() noexcept { return sizeof...(Ints); }
+};
+
+template <size_t... Ints>
+using index_sequence = integer_sequence<size_t, Ints...>;
+
+namespace detail {
+
+template <class Offset, class Lhs, class Rhs>
+struct merge_sequence_ {};
+template <class T, T Offset, T... Lhs, T... Rhs>
+struct merge_sequence_<
+  type_traits::integral_constant<T, Offset>,
+  integer_sequence<T, Lhs...>, integer_sequence<T, Rhs...>
+> {
+  using type = integer_sequence<T, Lhs..., (Offset + Rhs)...>;
+};
+
+template <class T, class N>
+struct make_integer_sequence_ {
+  using L = type_traits::integral_constant<T, N::value / 2>;
+  using R = type_traits::integral_constant<T, N::value - L::value>;
+  using type = typename merge_sequence_<
+    L, typename make_integer_sequence_<T, L>::type, typename make_integer_sequence_<T, R>::type
+  >::type;
+};
+template <class T>
+struct make_integer_sequence_<T, type_traits::integral_constant<T, 0>> {
+  using type = integer_sequence<T>;
+};
+template <class T>
+struct make_integer_sequence_<T, type_traits::integral_constant<T, 1>> {
+  using type = integer_sequence<T, 0>;
+};
+template <class T>
+struct make_integer_sequence_<T, type_traits::integral_constant<T, 2>> {
+  using type = integer_sequence<T, 0, 1>;
+};
+template <class T>
+struct make_integer_sequence_<T, type_traits::integral_constant<T, 3>> {
+  using type = integer_sequence<T, 0, 1, 2>;
+};
+template <class T>
+struct make_integer_sequence_<T, type_traits::integral_constant<T, 4>> {
+  using type = integer_sequence<T, 0, 1, 2, 3>;
+};
+
+} // namespace detail
+
+template <class T, T N>
+using make_integer_sequence = typename detail::make_integer_sequence_<
+  T, type_traits::integral_constant<T, N>
+>::type;
+
+template <size_t N>
+using make_index_sequence = make_integer_sequence<size_t, N>;
+
+} // namespace utility
+
+} // namespace cupy
```
</details>

leofang:
This is great! Thanks for the nice enhancements @eternalphane. I also noticed that the current JIT compiler lacks the support of accessing basic array attributes that Numba can do, so this is a good step toward a full-fledged JIT compiler.

cupyx/jit/_compile.py
asi1024: An implementation like as follows may be simpler. (This is a pseudo-code, and I have not tried compilation...)
```cpp
template <int ndim, typename... Args>
struct as_tuple {
    using ChildClass = as_tuple<ndim - 1, ptrdiff_t, Args...>;
    using T = typename ChildClass::T;

    template <typename Ints>
    static T call(Ints ints, Args... args) {
        return ChildClass::call(ints, ints[ndim - 1], args);
    }
};

template <typename... Args>
struct as_tuple<0, Args...> {
    using T = thrust::tuple<Args...>;

    template <typename Ints>
    static T call(Ints ints, Args... args) {
        return thrust::make_tuple(args...);
    }
};

...

return as_tuple<ndim>::call(shape)
```

cupyx/jit/_builtin_funcs.py
asi1024: Yes, I prefer ufunc because Python's built-in min/max behavior is more similar to ufunc than CUDA math API.
In the future, we will create a module for CUDA math API and intrinsic functions.

cupyx/jit/_compile.py
eternalphane: I just realized that though tuple ensures constancy, it also prohibits non-const indexing. So maybe the simplest way is just add const qualifier to the pointer.

asi1024: Do you mean that CuPy JIT will handle the shape as of an array type? Or internally holds the shape values as an array, and the JIT compiler handles it as a tuple of ints?

In any case, I don't want to directly feed back the compilation error from the generated CUDA code to the users for usability. CuPy's JIT compiler should detect the compilation error before feeding a generated code to the CUDA compiler.

asi1024: In addition, a shape of tuple of ints enables us to use pattern match. As a typical use case, we can write like `n, m = array.shape()`. :smiley:

eternalphane: If JIT handle the shape as tuple, then we can't write `for i in range(array.ndim): array.shape[i]`. Maybe returning a 1-dim `CArray` can cover both usecases.

eternalphane: > So you are planning to support a fixed-length array type whose length is determined at compile time?

Yes, just like `std::array`.
> It seems that we can cover these use cases by adding `size` and `writable` attributes to `Ptr` type.

That's a great idea. Thanks~

eternalphane: I added the following lines in `_transpile_stmt` before calling `_transpile_lvalue`:
https://github.com/cupy/cupy/blob/f0c999e8d22880f06e7f9bcf260209e702688be9/cupyx/jit/_compile.py#L348-L355
Maybe there's a better way to support unpacking `Ptr` without calling `thrust::make_tuple`?

asi1024: There are several concerns:
- The same expression `value.code` seems to be evaluated multiple times.
- Nested tuples are hard to support (e.g., `(n1, n2), (s1, s2) = array.shape, array.stride`)

Now I'm again feeling that it would be better to return a tuple as the shape. Since `array.ndim` is a constant, we can write `for i in range(array.ndim): array.shape[i]` by supporting compile-time loop unrolling in the future.

Anyway, the support for `.shape` and `.strides` still needs discussion will take some time. How about starting from the support for `len`, `min`, `max` and `.ndim` in this PR (It seems to be able to merge them soon! :smiley:), and continuing that discussion in another issue or PR?

eternalphane: I'll close this one and reopen another pr for `len`, `min`, `max` and `ndim`.

eternalphane: Anyway currently we don't support unpacking nested tuples. I'm thinking that instead of using `thrust::tie` on the lhs, we should call `thrust::get` on the rhs and treat each pair of elements as an assignment. Therefor `(a, b), (c, d) = (1, 2), (3, 4)` should be transpiled as `a, b = 1, 2` & `c, d = 3, 4`, and then `a = 1`, `b = 2`, `c = 3` & `d = 4`. To avoid calling rhs multiple times (if the rhs is an rvalue), maybe we should first assign it to a temporary variable and enclose the entire assignments in a block scope. If I recall correctly we always declare the variables at the beginning of the function, so the additional block scope won't introduce any side effect.

asi1024: > Anyway currently we don't support unpacking nested tuples.

Thank you for the reporting! I didn't notice that. I will fix it soon!

leofang:
@asi1024 I tried to revive this PR and work on `.shape` and `.strides` but hit a small blocker: https://github.com/NVIDIA/thrust/issues/524#issuecomment-1101762977. But, I think since CuPy vendored Thrust's tuple header, we could offer our own specialization up to NDIM=32. 
https://github.com/cupy/cupy/blob/e6a6bccb7613431436709ea56fae0571920e6ede/cupy/_core/_carray.pxd#L12
What do you think?

leofang:
Follow-up: #6668

PR #5304, opened 2021-06-01
Fix cuFFT callback for sm_61 etc

### LINKED ISSUE #5296: Get segmentation fault with FFT callback function
`pr.py`
```python
import numpy as np
import cupy as cp

code = r'''
__device__ cufftComplex CB_ConvertInput(
    void *dataIn,
    size_t offset,
    void *callerInfo,
    void *sharedPtr) {
            return make_float2(0.0, 0.1);
}

__device__ cufftCallbackLoadC d_loadCallbackPtr = CB_ConvertInput;
'''

in_arr = cp.random.rand(12, dtype=np.float32)

with cp.fft.config.set_cufft_callbacks(cb_load=code):
    out_arr = cp.fft.fft(in_arr)

print(out_arr)
```
Run with
```
CUDA_VISIBLE_DEVICES=0 python pr.py
```
Then I got
```
[1]    410392 segmentation fault (core dumped)  CUDA_VISIBLE_DEVICES=0 python pr.py
```

My system information:
```
OS                           : Linux-5.12.8-arch1-1-x86_64-with-glibc2.33
Python Version               : 3.9.5
CuPy Version                 : 9.1.0
NumPy Version                : 1.20.0
SciPy Version                : 1.6.3
Cython Build Version         : 0.29.23
Cython Runtime Version       : 0.29.23
CUDA Root                    : /opt/cuda
nvcc PATH                    : /opt/cuda/bin/nvcc
CUDA Build Version           : 11030
CUDA Driver Version          : 11030
CUDA Runtime Version         : 11030
cuBLAS Version               : 11402
cuFFT Version                : 10402
cuRAND Version               : 10204
cuSOLVER Version             : (11, 1, 1)
cuSPARSE Version             : 11500
NVRTC Version                : (11, 3)
Thrust Version               : 101100
CUB Build Version            : 101100
Jitify Build Version         : 60e9e72
cuDNN Build Version          : 8200
cuDNN Version                : 8200
NCCL Build Version           : 20906
NCCL Runtime Version         : 20906
Device 0 Name                : NVIDIA GeForce GTX 1080
Device 0 PCI Bus ID          : 0000:17:00.0
Device 1 Name                : NVIDIA Quadro K600
Device 1 Compute Capability  : 30
Device 1 PCI Bus ID          : 0000:65:00.0
```

leofang:
Fix #5296. Thanks to @szsdk for reporting. 

We need to check if `nvprune` actually succeeds or not, as the static library does not contain device codes for certain CCs:
> Please note that the cuFFT library might not contain code for certain architectures as long as there is code for a lower architecture that is binary compatibile (e.g. SM37, SM52, SM61). 
> https://docs.nvidia.com/cuda/cufft/index.html#static-library

for example see https://github.com/cupy/cupy/issues/5296#issuecomment-851713662. 

To avoid the cost of rerunning `nvprune` for a CC known to not have any device code, we cache the CC after the first validation call. An alternative approach would be to not run `nvprune` at all (save us one long run) for a list of pre-compiled CCs, but there would be a maintenance burden on us to update the list for future CUDA releases.

### DIFF
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -207,7 +207,8 @@ cdef inline str _prune(str temp_dir, str cache_dir, str _cufft_ver, str arch):
         if not os.path.isfile(cufft_lib_cached):
             p = subprocess.run([_nvprune, '-arch=sm_' + arch,
                                 cufft_lib_full, '-o', cufft_lib_temp],
-                               env=os.environ, cwd=temp_dir)
+                               env=os.environ, cwd=temp_dir,
+                               stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             p.check_returncode()
             if p.stderr:
                 # if no device code exists for sm_XX, nvprune exits normally
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -213,7 +213,7 @@ cdef inline str _prune(str temp_dir, str cache_dir, str _cufft_ver, str arch):
             if p.stderr:
                 # if no device code exists for sm_XX, nvprune exits normally
                 # with a warning printed to stderr
-                assert 'No device code' in p.stderr
+                assert b'No device code' in p.stderr
                 cufft_lib_pruned = None
                 _cufft_static_no_device_code.append(arch)
             else:
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -213,6 +213,7 @@ cdef inline str _prune(str temp_dir, str cache_dir, str _cufft_ver, str arch):
             if p.stderr:
                 # if no device code exists for sm_XX, nvprune exits normally
                 # with a warning printed to stderr
+                assert 'No device code' in p.stderr
                 cufft_lib_pruned = None
                 _cufft_static_no_device_code.append(arch)
             else:
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -39,6 +39,7 @@ cdef str _cuda_include = None
 cdef str _nvprune = None
 cdef str _build_ver = None
 cdef int _cufft_ver = 0
+cdef list _cufft_static_no_device_code = []
 cdef str _cupy_root = None
 cdef str _cupy_include = None
 cdef str _source_dir = None
@@ -196,7 +197,7 @@ cdef inline void _mod_compile(str tempdir, str mod_name, str obj_host) except*:
 cdef inline str _prune(str temp_dir, str cache_dir, str _cufft_ver, str arch):
     cdef str cufft_lib_full, cufft_lib_pruned, cufft_lib_temp, cufft_lib_cached
 
-    if _nvprune:
+    if _nvprune and arch not in _cufft_static_no_device_code:
         cufft_lib_full = os.path.join(_cuda_path, 'lib64/libcufft_static.a')
         cufft_lib_pruned = 'cufft_static_' + _cufft_ver + '_sm' + arch
         cufft_lib_temp = os.path.join(temp_dir,
@@ -208,9 +209,15 @@ cdef inline str _prune(str temp_dir, str cache_dir, str _cufft_ver, str arch):
                                 cufft_lib_full, '-o', cufft_lib_temp],
                                env=os.environ, cwd=temp_dir)
             p.check_returncode()
-            # atomic move with the destination guaranteed to be overwritten;
-            # using os.replace() is also ok here
-            os.rename(cufft_lib_temp, cufft_lib_cached)
+            if p.stderr:
+                # if no device code exists for sm_XX, nvprune exits normally
+                # with a warning printed to stderr
+                cufft_lib_pruned = None
+                _cufft_static_no_device_code.append(arch)
+            else:
+                # atomic move with the destination guaranteed to be overwritten
+                # (using os.replace() is also ok here)
+                os.rename(cufft_lib_temp, cufft_lib_cached)
     else:
         # nvprune is not found, just link against the full static lib
         cufft_lib_pruned = None


leofang:
@szsdk, any chance you can help us verify this patch? I can give you instructions for building from source if needed 🙂

szsdk:
@leofang Yeah, sure. I'd love to give some help if I could.

szsdk:
@leofang I've installed 4dc8bfbd68d48b6. But still, get the same segfault.

leofang:
> @leofang I've installed [4dc8bfb](https://github.com/cupy/cupy/commit/4dc8bfbd68d48b62bc2cf64559ebf0ee43edc899). But still, get the same segfault.

Thanks for the quick test @szsdk. Could you remove the cached files under ~/.cupy and retry?

leofang:
> > @leofang I've installed [4dc8bfb](https://github.com/cupy/cupy/commit/4dc8bfbd68d48b62bc2cf64559ebf0ee43edc899). But still, get the same segfault.
> 
> Thanks for the quick test @szsdk. Could you remove the cached files under ~/.cupy and retry?

One more note: Please make sure you do not have multiple CuPy installed in your Python environment. If you're using CuPy from the system package manager, you should uninstall it first.

szsdk:
I removed the `.cupy` folder when I did the test. There should not be multiple installations. Anyway, I am trying it again.

May I know how to improve compile speed? Now I install cupy with `pip install .`. It compiles the whole project each times and takes a long time. I am supposed that installation should be much faster in development.

leofang:
> May I know how to improve compile speed? Now I install cupy with `pip install .`. It compiles the whole project each times and takes a long time. I am supposed that installation should be much faster in development.

Thanks. You can set `CUPY_NVCC_GENERATE_CODE` to limit to your SM only, and `CUPY_NUM_BUILD_JOBS ` to increase the parallelism (has less speedup than the previous one), see https://docs.cupy.dev/en/stable/reference/environment.html.

szsdk:
This is the test code:
```python
import numpy as np
import cupy as cp

print(cp.__version__)

code = r'''
__device__ cufftComplex CB_ConvertInput(
    void *dataIn,
    size_t offset,
    void *callerInfo,
    void *sharedPtr) {
            return make_float2(0.0, 0.1);
}

__device__ cufftCallbackLoadC d_loadCallbackPtr = CB_ConvertInput;
'''

in_arr = cp.random.rand(12, dtype=np.float32)

with cp.fft.config.set_cufft_callbacks(cb_load=code):
    out_arr = cp.fft.fft(in_arr)

print(out_arr)
```
This is the output after removing `.cupy`.
```
10.0.0a1
nvprune warning : No device code that matched architecture, so stripped out all device code
[1]    1427046 segmentation fault (core dumped)  python pr.py
```

leofang:
Thanks, @szsdk. I will take a closer look tonight.

leofang:
@szsdk Could you give cf78fbf a try? I fixed a stupid mistake and am seeing interesting behavior locally: Looks like we do not have to fix the `nvcc` compile and link flags, but it's best to verify this with your 980/1080 GPUs.

On the build time: if it's a small change, only the affected module is rebuilt so should be very fast, but I usually do `pip install -v -e .` to build it in the develop mode, so I am less certain if without `-e` it'd still reuse the leftovers from the previous build or not.

szsdk:
OMG. cf78fbf86eec6a8a4bae works in my workstation. I will try it with other machines.

leofang:
Jenkins, test this please

emcastillo:
Looks good to me at first sight but I want @szsdk to confirm that it worked in all scenarios if possible :)
Thanks!

cupy/fft/_callback.pyx
```diff
@@ -206,11 +207,18 @@ cdef inline str _prune(str temp_dir, str cache_dir, str _cufft_ver, str arch):
...
+                               stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             p.check_returncode()
-            # atomic move with the destination guaranteed to be overwritten;
-            # using os.replace() is also ok here
-            os.rename(cufft_lib_temp, cufft_lib_cached)
+            if p.stderr:
+                # if no device code exists for sm_XX, nvprune exits normally
+                # with a warning printed to stderr
+                cufft_lib_pruned = None
+                _cufft_static_no_device_code.append(arch)
```
emcastillo, line 217: Can we do an explicit check of the warning message here?

szsdk:
@emcastillo 
Yes. What I have tested are
- NVIDIA 980; nvcc 11.0
- NVIDIA 1080; nvcc 11.3
- NVIDIA 1080; nvcc 11.1

cupy/fft/_callback.pyx
leofang: Yes, let me add an assert.

leofang: Added in cf6f1ab.

leofang:
Jenkins, test this please

szsdk:
@leofang I tried to test commit cf6f1abf39789f0ed46e107e807cdb866964d800 in my GTX980 workstation. But got an error.
This is my code:
```python
import cupy as cp
_load_callback_1 = r"""
__device__ cufftReal CB_ConvertInput(
    void *dataIn,
    size_t offset,
    void *callerInfo,
    void *sharedPtr
) {
    return 2.0;
}

__device__ cufftCallbackLoadR d_loadCallbackPtr = CB_ConvertInput;
"""

cp.fft.config.set_cufft_callbacks(
    cb_load=_load_callback_1,
)
```
This is the error message:
```
  File "hh.py", line 15, in <module>
    cp.fft.config.set_cufft_callbacks(
  File "cupy/fft/_callback.pyx", line 567, in cupy.fft._callback.set_cufft_callbacks.__init__
  File "cupy/fft/_callback.pyx", line 379, in cupy.fft._callback._CallbackManager.__init__
  File "cupy/fft/_callback.pyx", line 216, in cupy.fft._callback._prune
TypeError: a bytes-like object is required, not 'str'
```
But the strange thing is that my 1080 desktop works fine.

leofang:
Oops @szsdk thanks for catching it...It should now be fixed in 5002572.

szsdk:
@leofang Now it works on all my machines.

emcastillo:
Jenkins, test this please

emcastillo (approved):

leofang:
Thanks @szsdk @emcastillo!

Issue #4789, opened 2021-03-02
Adopt Python Array API standard


leofang:
This is a meta-issue tracking the adoption of Python Array API standard:
- Intro: https://data-apis.org/blog/array_api_standard_release/
- API spec: https://data-apis.github.io/array-api/latest/
- NumPy NEP-47: https://github.com/numpy/numpy/pull/18456 (rendered in https://numpy.org/neps/nep-0047-array-api-standard.html)
- Array API repo: https://github.com/data-apis/array-api

In the Data API weekly meetings, it was decided that a reference implementation will be provided to NumPy (see NEP-47), which other implementations can follow and test against. Given that CuPy is IMHO by far the most compliant library with the API standard (probably more compliant than NumPy is, given many design deviations), I think it is much easier if we just wait until the ref implementation is out, and try taking it as is.

A new array object will be added in a separate namespace, delegating real work to `cupy.ndarray` whenever appropriate. Following NEP-47 I suggest the namespace to be `cupy.array_api`, but I suppose `cupyx.array_api` is also a fine choice.
The array object will only have "dunder" methods (names with double underscores), and all APIs for manipulating this object live under the new namespace.

But in addition to the new array object and APIs added to the new namespace, similar to `numpy`/`numpy.ndarray` there are some work to be done in the `cupy` namespace as well as in `cupy.ndarray`. A list of such examples are copied from NEP-47 for convenience:
- [x] `ndarray.__dlpack__`
- [x] `ndarray.__dlpack_device__`
- [x] `ndarray.device`

In addition, since DLPack is the main data exchange mechanism, we better support
- [ ] allocate numpy arrays backed by pinned memory (`kDLCPUPinned`) (related: #3625)

There are still some issues in the standard to be ironed out (ex: linalg related APIs), but I think it's getting close to be completed.

leofang:
> it was decided that a reference implementation will be provided to NumPy 

See numpy/numpy#18585.

leofang:
> > it was decided that a reference implementation will be provided to NumPy
> 
> See [numpy/numpy#18585](https://github.com/numpy/numpy/pull/18585).

@kmaehashi Would CuPy follow exactly what NumPy does and creates a new `_array_api` namespace, or would you rather follow PyTorch and turn `cupy.ndarray` directly into an Array API compliant array object? I took the former as granted, but I think I better double check with you guys 😅

kmaehashi:
[...]

PR #5306, opened 2021-06-01
Support the new DLPack exchange protocol

leofang:
~~Blocked by the design question https://github.com/numpy/numpy/pull/19083#issuecomment-852286591.~~ 
***UPDATE***: `cupy.from_dlpack()` returns a `cupy.ndarray`, and in the future `cupy.array_api.from_dlpack()` will return an array-API-compliant object, see https://github.com/numpy/numpy/pull/19083#issuecomment-854926910.

Part of #4789. This PR implements the support for the new DLPack exchange protocol specified in the Python Array API Standard, see https://data-apis.org/array-api/latest/design_topics/data_interchange.html for detail.

This PR is a parallel of the ongoing work in NumPy (https://github.com/numpy/numpy/pull/19083). 

TODO:
- [x] Add tests
- [x] Add docs
- [ ] Support DLPack 0.5+ (separate PR)


leofang:
@kmaehashi Could you create a new label for Array API related work, say `cat:array-api`, and label this PR? I would like to keep track of relevant efforts.

tests/cupy_tests/core_tests/test_dlpack.py
```diff
@@ -6,25 +6,42 @@
...
+        array = cupy.random.randint(
+            0, 10, size=(2, 3)).astype(dtype)
+    elif cupy.issubdtype(dtype, cupy.integer):
+        array = cupy.random.randint(
+            -10, 10, size=(2, 3)).astype(dtype)
+    elif cupy.issubdtype(dtype, cupy.floating):
+        array = cupy.random.rand(
+            2, 3).astype(dtype)
+    elif cupy.issubdtype(dtype, cupy.complexfloating):
+        array = cupy.random.random((2,3)).astype(dtype)
```
leofang, line 20: This actually fixes a test bug (complex dtypes not covered) which I will backport manually.

kmaehashi:
> @kmaehashi Could you create a new label for Array API related work, say cat:array-api, and label this PR? I would like to keep track of relevant efforts.

Sure, I've made a label for it.

leofang:
> Sure, I've made a label for it.

Thanks, @kmaehashi! I think this is ready.

leofang:
Let's do a test drive.

Jenkins, test this please

cupy/_core/dlpack.pxd
```diff
@@ -2,3 +2,8 @@ from cupy._core.core cimport ndarray
 
 cpdef object toDlpack(ndarray array) except +
 cpdef ndarray fromDlpack(object dltensor) except +
+cpdef from_dlpack(array)
+
+cdef enum:
+    device_CUDA = 2  # kDLCUDA
+    device_ROCM = 10  # kDLROCM
```
kmaehashi, line 9:
```suggestion
+cdef extern from './include/cupy/dlpack/dlpack.h' nogil:
+    int device_CUDA 'kDLGPU'
+    int device_ROCM 'kDLROCM'
```

cupy/_core/dlpack.pyx
```diff
@@ -303,3 +309,58 @@ cpdef ndarray fromDlpack(object dltensor) except +:
...
+        implementing the DLPack data exchange protocol.
+
+    .. seealso::
+        `Data interchange mechanisms`_
+
+    .. _Data interchange mechanisms:
+        https://data-apis.org/array-api/latest/design_topics/data_interchange.html
+    """
+    try:
+        dev_type, dev_id = array.__dlpack_device__()
```
kmaehashi, line 343: Could you use `hasattr` instead? In general, it's better to avoid relying on exceptions in normal cases. This also avoids catching `AtttributeError` raised from inside `__dlpack_device__()`.

docs/source/reference/ext.rst
```diff
@@ -19,13 +19,15 @@ CuPy-specific functions are placed under ``cupyx`` namespace.
...
-This is a helper function for creating a :class:`cupy.ndarray` from a DLPack tensor.
+Below are helper functions for creating a :class:`cupy.ndarray` from either a DLPack tensor
+or any object supporting the DLPack data exchange protocol.
 For further detail see :ref:`dlpack`.
 
 .. autosummary::
    :toctree: generated/
 
    cupy.fromDlpack
+   cupy.from_dlpack
```
kmaehashi, line 30: Honestly speaking I don't want to document this API, as it's likely to be removed in the future & having two similar API is confusing to users. How about changing `cupy.fromDlpack` to support both raw pycapsule and `__dlpack__` protocol?
@emcastillo is working in this approach in PyTorch: https://github.com/pytorch/pytorch/pull/57110

kmaehashi (changes requested):
Thanks @leofang for working on this!

cupy/_core/dlpack.pxd
leofang: I have no idea why I didn't do this! Thanks @kmaehashi 😅

docs/source/reference/ext.rst
leofang: This API is [documented](https://data-apis.org/array-api/latest/API_specification/creation_functions.html#from-dlpack-x) as part of the Array API standard. People have invested great effort into it, so I don't think it'd go away 🙂 

I would actually suggest to deprecate `cupy.fromDlpack` as it is known unsafe w.r.t. non-default streams
https://github.com/cupy/cupy/blob/2f8552bade9f5994a9417922da272778dcf0c850/docs/source/user_guide/interoperability.rst#L296
It is functionally equivalent to `arr.__dlpack__(None)` so users do not lose anything. With more number of libraries supporting the new DLPack protocol (`__dlpack_device__` + `__dlpack__` + `from_dlpack()`) such as us, NumPy, PyTorch, RAPIDS, etc, I think it is better to encourage users to switch to the new protocol and not manage the pycapsule object by themselves.

As for making `from_dlpack()` support both pycapsule and compliant arrays, I feel it's best that we discuss it in the PyTorch PR (or the array-api repo) first, and then implement the decision. I personally think this is confusing and against the safe design of the protocol...

cupy/_core/dlpack.pyx
leofang: I think `__dlpack_device__` shouldn't raise, but ok will do.

docs/source/reference/ext.rst
kmaehashi: First of all, I really thank many people, including you, for the efforts on standardization! 😄 
My understanding is that we will introduce `cupy.array_api.from_dlpack()` at some point and then remove `cupy.from_dlpack()` and `cupy.fromDlpack()`. Is this correct?

> I would actually suggest to deprecate cupy.fromDlpack as it is known unsafe w.r.t. non-default streams

I agree with deprecating `fromDlpack()`, however as most existing producers do not support `__dlpack__` yet, we still have to provide a way to load raw PyCapsule objects. So I think supporting both raw PyCausule and `__dlpack__` in `from_dlpack` is a good tentative solution for now.

leofang: > First of all, I really thank many people, including you, for the efforts on standardization! 😄

My pleasure! I was just trying to stress that the hope is to make the DLPack protocol sustain. We can definitely mark it experimental at least in CuPy v10, given that there are still ongoing discussions (ex: https://github.com/dmlc/dlpack/issues/74), but the goal is to improve it, not to replace it by something else.

> My understanding is that we will introduce `cupy.array_api.from_dlpack()` at some point and then remove `cupy.from_dlpack()` and `cupy.fromDlpack()`. Is this correct?

It depends. There is a [discussion](https://github.com/cupy/cupy/issues/4789#issuecomment-864725951) I'd like to kick off with the team. If we follow NumPy and have two separate namespaces, `cupy.from_dlpack()` can stay (and return a `cupy.ndarray`) and a new `cupy._array_api.from_dlpack()`will be added (and return a `cupy._array_api.ndarray`). 

In any case I hope to deprecate and remove `cupy.fromDlpack()`.

> So I think supporting both raw PyCausule and `__dlpack__` in `from_dlpack` is a good tentative solution for now.

Sounds like this will be [an acceptable workaround even in PyTorch](https://github.com/pytorch/pytorch/pull/57110#issuecomment-864599710), so I will follow!

kmaehashi: > If we follow NumPy and have two separate namespaces, `cupy.from_dlpack()` can stay (and return a `cupy.ndarray`) and a new `cupy._array_api.from_dlpack()`will be added (and return a `cupy._array_api.ndarray`). 

I see, sounds nice.

> In any case I hope to deprecate and remove `cupy.fromDlpack()`.

Agree.

> Sounds like this will be [an acceptable workaround even in PyTorch](https://github.com/pytorch/pytorch/pull/57110#issuecomment-864599710), so I will follow!

:+1:

leofang: @kmaehashi I am wondering if we should deprecate `toDlpack` as well...

kmaehashi: I think `toDlpack` deprecation should happen after most consumers support new `from_dlpack` protocol.

docs/source/reference/ext.rst
```diff
@@ -19,13 +19,15 @@ CuPy-specific functions are placed under ``cupyx`` namespace.
...
 
-This is a helper function for creating a :class:`cupy.ndarray` from a DLPack tensor.
+Below are helper functions for creating a :class:`cupy.ndarray` from either a DLPack tensor
+or any object supporting the DLPack data exchange protocol.
 For further detail see :ref:`dlpack`.
 
 .. autosummary::
    :toctree: generated/
 
    cupy.fromDlpack
```
kmaehashi, line 29: Could you remove this line from the list?

kmaehashi (changes requested):
LGTM excpet for docs!

docs/source/reference/ext.rst
leofang: Thanks! Done in 918f3f7.

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

kmaehashi:
Test failures are unrelated. Thanks @leofang for working on this!

tests/cupy_tests/core_tests/test_dlpack.py
leofang: See #5425.

PR #5308, opened 2021-06-02
Temporarily remove the async pool test from `TestAllocator`

leofang:
Follow-up of #5177. When debugging for https://github.com/chainer/chainer-test/pull/658#issuecomment-853133782 I realized my changes in the `TestAllocator` class was not legitimate, because it requires the ability of creating a new memory pool, which we currently do not support for `MemoryAsyncPool`:
https://github.com/cupy/cupy/blob/29eb596738a612bd77e2852f1cf6be99eb391ec0/cupy/cuda/memory.pyx#L1607-L1609
I will add the support in a separate PR,  after which we can re-enable this test.

cc: @kmaehashi

### DIFF
--- a/tests/cupy_tests/cuda_tests/test_memory.py
+++ b/tests/cupy_tests/cuda_tests/test_memory.py
@@ -728,8 +728,11 @@ class TestMemoryPool(unittest.TestCase):
             assert 0 == self.pool.total_bytes()
 
 
+# TODO(leofang): test MemoryAsyncPool. We currently remove the test because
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, test this please.

kmaehashi (approved):
Thanks, LGTM!

leofang:
Thanks, @kmaehashi. This does not need to backport.

kmaehashi:
Ah, ok #5177 was not backported.

PR #5309, opened 2021-06-03
Don't set FFT work area when gained work area size is zero

takagi:
`hipfftMakePlan1d` can output zero for the work area size, and given a null pointer returned by the allocator that supplied zero for the size, `hipfftSetWorkArea` (internally calls `rocfft_execution_info_set_work_buffer`) returns `HIPFFT_INVALID_VALUE`.

This PR fixes `Plan1d` not to set the work area pointer if the computed work area size is zero. Also `PlanNd` has the same fix.

### DIFF
--- a/cupy/cuda/cufft.pyx
+++ b/cupy/cuda/cufft.pyx
@@ -337,13 +337,14 @@ cdef class Plan1d:
                                          &work_size)
         check_result(result)
 
-        work_area = memory.alloc(work_size)
-        ptr = <intptr_t>(work_area.ptr)
-        with nogil:
-            result = cufftSetWorkArea(plan, <void*>(ptr))
-        check_result(result)
+        if work_size > 0:
+            work_area = memory.alloc(work_size)
+            ptr = <intptr_t>(work_area.ptr)
+            with nogil:
+                result = cufftSetWorkArea(plan, <void*>(ptr))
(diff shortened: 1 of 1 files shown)


cupy/cuda/cufft.pyx
```diff
@@ -789,18 +790,18 @@ cdef class PlanNd:
...
-        check_result(result)
+        if work_size > 0:
+            # TODO: for CUDA>=9.2 could also allow setting a work area policy
+            # result = cufftXtSetWorkAreaPolicy(plan, policy, &work_size)
+            work_area = memory.alloc(work_size)
+            ptr = <intptr_t>(work_area.ptr)
+            with nogil:
+                result = cufftSetWorkArea(plan, <void*>(ptr))
+            check_result(result)
+            self.work_area = work_area
```
kmaehashi, line 801:
```suggestion
            self.work_area = work_area
        else:
            self.work_area = None        
```

emcastillo: good catch!

takagi:
I mistakenly pushed the branch to cupy repo instead of mine. Let me close this PR and post the correct one again.

kmaehashi:
Removed the branch.

takagi:
Thanks!

PR #4794, opened 2021-03-02
Windows: fix vectorize tests

leofang:
For `a = np.array([1, 2, 3, 4])`, on Linux `a.dtype == np.int64`, but on Windows `a.dtype == np.int32`.

PR #5319, opened 2021-06-07
Support `len`, `min`, `max`, `.ndim`, `.size` in jit

eternalphane:
Updated version of #5293.
- Built-in functions
  - [x] len
  - [x] min
  - [x] max
- Array size functions
  - [x] `ArrayBase.ndim`
  - [x] `CArray.size`
- Misc
  - [x] const-indexing of tuple

### DIFF
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -54,7 +54,7 @@ class RangeFunc(BuiltinFunc):
         return Range(start, stop, step, ctype, step_is_positive)
 
 
-class Len(BuiltinFunc):
+class LenFunc(BuiltinFunc):
     def call(self, env, *args, **kwds):
         if len(args) != 1:
             raise TypeError(f'len() expects only 1 argument, got {len(args)}')
@@ -69,7 +69,7 @@ class Len(BuiltinFunc):
                     _cuda_types.Scalar('q'))
 
 
-class Min(BuiltinFunc):
+class MinFunc(BuiltinFunc):
     def call(self, env, *args, **kwds):
         if len(args) < 2:
             raise TypeError(
@@ -80,7 +80,7 @@ class Min(BuiltinFunc):
             cupy.minimum, (a, b), None, env), args)
 
 
-class Max(BuiltinFunc):
+class MaxFunc(BuiltinFunc):
     def call(self, env, *args, **kwds):
         if len(args) < 2:
             raise TypeError(
@@ -152,9 +152,9 @@ class AtomicOp(BuiltinFunc):
 
 builtin_functions_dict = {
     range: RangeFunc(),
-    len: Len(),
-    min: Min(),
-    max: Max(),
+    len: LenFunc(),
+    min: MinFunc(),
+    max: MaxFunc(),
 }
 
 syncthreads = SyncThreads()
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -65,7 +65,8 @@ class Len(BuiltinFunc):
             raise TypeError('len() supports only array type')
         if not arg.ctype.ndim:
             raise TypeError('len() of unsized array')
-        return Data(f'{arg.code}.shape()[0]', _cuda_types.PtrDiff())
+        return Data(f'static_cast<long long>({arg.code}.shape()[0])',
+                    _cuda_types.Scalar('q'))
 
 
 class Min(BuiltinFunc):
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -1,3 +1,5 @@
+import cupy
+
 from cupy_backends.cuda.api import runtime
 from cupyx.jit import _cuda_types
 from cupyx.jit._internal_types import BuiltinFunc
@@ -5,6 +7,8 @@ from cupyx.jit._internal_types import Data
 from cupyx.jit._internal_types import Constant
 from cupyx.jit._internal_types import Range
 from cupyx.jit import _compile
+
+from functools import reduce
 
 
 class RangeFunc(BuiltinFunc):
@@ -22,6 +26,13 @@ class RangeFunc(BuiltinFunc):
             raise TypeError(
                 f'range expected at most 3 argument, got {len(args)}')
 
+        if isinstance(step, Constant):
+            step_is_positive = step.obj >= 0
+        elif step.ctype.dtype.kind == 'u':
+            step_is_positive = True
+        else:
+            step_is_positive = None
+
         stop = Data.init(stop, env)
         start = Data.init(start, env)
         step = Data.init(step, env)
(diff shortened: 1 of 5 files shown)


tests/cupyx_tests/jit_tests/test_raw.py
```diff
@@ -72,6 +89,20 @@ def f(x, y):
         f((1,), (1,), (x, y))
         assert bool((x == y).all())
 
+    def test_min_max(self):
```
asi1024, line 92: Could you split this test into `test_min` and `test_max`?

asi1024 (commented):
Could you add tests also for `len`, `.ndim`, `.size` and tuple-indexing?

asi1024:
Jenkins, test this please.

eternalphane:
The error in `test_raw_ndim` can be fixed by the following patch:
```diff
diff --git a/cupyx/jit/_compile.py b/cupyx/jit/_compile.py
index 836d161f4..9e42a3a29 100644
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -343,7 +343,7 @@ def _transpile_stmt(stmt, is_toplevel, env):
                 else:
                     raise TypeError(
                         'Cannot assign constant value not at top-level.')
-            value = Data.init(value, env)
+        value = Data.init(value, env)

         target = _transpile_lvalue(target, env, value.ctype)
         return [f'{target.code} = {value.code};']
```
However I can't reproduce the error in `test_raw_multidimensional_array_with_attr`...

asi1024:
In `test_raw_multidimensional_array_with_attr`, the following CUDA function is generated.

```cpp
__device__ long long cupy_remainder_int64(T0 in0, T1 in1) {
    return (in0 - _floor_divide(in0, in1) * in1) * (in1 != 0);
}

extern "C" __global__ void f(CArray<int, 2, true, true> x, CArray<int, 2, true, true> y) {
  unsigned int tid;
  unsigned int ntid;
  long long n_col;
  long long i;
  long long i_row;
  long long i_col;
  tid = (threadIdx.x + (blockDim.x * blockIdx.x));
  ntid = (blockDim.x * gridDim.x);
  n_col = (_floor_divide(x.size(), x.shape()[0]));
  for (long long __it = tid, __stop = x.size(), __step = ntid; __it < __stop; __it += __step) {
    i = __it;
    i_row = (_floor_divide(i, n_col));
    i_col = cupy_remainder_int64(i, n_col);
    y._indexing(thrust::make_tuple(i_row, i_col)) = x._indexing(thrust::make_tuple(i_row, i_col));
  }
}
```

`_floor_divide(x.size(), x.shape()[0])` calls `_floor_divide` with `(ptrdiff_t, const ptrdiff_t)` argument type, but it is not supported (https://github.com/cupy/cupy/blob/master/cupy/_core/include/cupy/carray.cuh#L597-L626)

cupyx/jit/_cuda_types.py
```diff
@@ -40,6 +40,14 @@ def __hash__(self):
         return hash(self.dtype)
 
 
+class PtrDiff(Scalar):
+    def __init__(self):
+        super().__init__('q')
+
+    def __str__(self):
+        return 'ptrdiff_t'
```
asi1024, line 48: How about using `Scalar(numpy.int64)` instead of `PtrDiff`?

cupyx/jit/_compile.py
```diff
@@ -562,6 +562,12 @@ def _transpile_expr_internal(expr, env):
         value = _transpile_expr(expr.value, env)
         if is_constants(value):
             return Constant(getattr(value.obj, expr.attr))
+        if isinstance(value.ctype, _cuda_types.ArrayBase):
+            if 'ndim' == expr.attr:
+                return Constant(value.ctype.ndim)
+        if isinstance(value.ctype, _cuda_types.CArray):
+            if 'size' == expr.attr:
+                return Data(f'{value.code}.size()', _cuda_types.PtrDiff())
```
asi1024, line 570:
```suggestion
                return Data(f'(long long)({value.code}.size())', _cuda_types.Scalar(numpy.int64))
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -33,13 +44,6 @@ def call(self, env, *args, **kwargs):
         if step.ctype.dtype.kind not in 'iu':
             raise TypeError('range supports only for integer type.')
 
-        if isinstance(step, Constant):
-            step_is_positive = step.obj >= 0
-        elif step.ctype.dtype.kind == 'u':
-            step_is_positive = True
-        else:
-            step_is_positive = None
-
```
leofang, line 42: Question: Why this block needs to be moved ahead?

cupyx/jit/_builtin_funcs.py
```diff
@@ -111,6 +151,9 @@ def call(self, env, array, index, value):
 
 builtin_functions_dict = {
     range: RangeFunc(),
+    len: Len(),
+    min: Min(),
+    max: Max(),
```
leofang, line 156: Nitpick: Can we follow the naming convention above and append `Func` for Python built-in functions? So they would read `LenFunc`, `MinFunc`, and `MaxFunc`, respectively.

cupyx/jit/_cuda_typerules.py
```diff
@@ -47,27 +48,27 @@
 _py_ops = {
     ast.And: lambda x, y: x and y,
     ast.Or: lambda x, y: x or y,
-    ast.Add: lambda x, y: x + y,
```
leofang, line 50: Using `operator` is great! 👍

cupyx/jit/_cuda_typerules.py
```diff
@@ -140,7 +141,7 @@ def get_ctype_from_scalar(mode, x):
     raise NotImplementedError(f'{x} is not scalar object.')
 
 
-_typechars = '?bBhHiIlLefdFD'
+_typechars = '?bBhHiIlLqQefdFD'
```
leofang, line 144: @asi1024 Could you confirm this is OK? I seem to remember we removed `qQ` for good reason (perhaps for Windows)?

cupyx/jit/_builtin_funcs.py
eternalphane: We should check whether `step` is `Constant` before `step = Data.init(step, env)`, since after this line it will be converted to `Data`.

cupyx/jit/_cuda_types.py
eternalphane: I think maybe we should use SFINAE and `is_signed`, `is_unsigned`, `is_floating_point` to implement `_floor_divide`, but static-casting to `long long` is indeed much simpler.😄

cupyx/jit/_cuda_typerules.py
asi1024: Maybe [this fix](https://github.com/cupy/cupy/pull/4794/files#diff-178317a5acc0687f9d0bc99003038b5b5b754a98e4fbccb8024761fc81d88a6fR122) resolved most CuPy JIT issues on Windows!

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024 (approved):

asi1024:
LGTM. Thanks!

leofang:
Nice work @eternalphane!

PR #5331, opened 2021-06-09
Add cupyx.scipy.sparse.linalg.lsmr

povinsahu1909:
This PR add `cupyx.sparse.linalg.lsmr`.


povinsahu1909:
Will fix the docs error with other changes.

takagi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
```diff
@@ -1268,3 +1268,28 @@ def test_spilu_0(self, dtype, xp, sp):
...
+    def _make_matrix(self, sp):
+        A = testing.shaped_sparse_random((50, 50), sp, density=0.2)
+        return A
+
+    def _make_normalized_vector(self, xp):
+        b = testing.shaped_random((50,), xp, scale=1)
+        return b
+
+    def _test_lsmr(self, xp, sp, a, b):
+        return sp.linalg.lsmr(a, b, damp=0.0, atol=1e-6, btol=1e-6)
```
takagi, line 1288: Missing `self.damp`?

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
```diff
@@ -1268,3 +1268,28 @@ def test_spilu_0(self, dtype, xp, sp):
         else:
             ainv = sp.linalg.spilu(a)
         return ainv.solve(b)
+
+
+@testing.parameterize(*testing.product({
+    'damp': [0, 1, 2]
+}))
+@testing.gpu
```
takagi, line 1276: Now we don't need to use `@testing.gpu` as it was originally introduced when CuPy was a part of Chainer, and does not make any use in CuPy. (Recent [change](https://github.com/cupy/cupy/wiki/Development-policies/_compare/caddfd6f010b3c93a90dd8820db9c0b7426cd86e...072ee9bf3da3d15119f4620b67cbb5aae03eaa18) in our development policy )

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
```diff
@@ -1268,3 +1268,28 @@ def test_spilu_0(self, dtype, xp, sp):
         else:
             ainv = sp.linalg.spilu(a)
         return ainv.solve(b)
+
+
+@testing.parameterize(*testing.product({
```
takagi, line 1273: Would you test each matrix format: `csr`, `csc`, `coo`?

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
povinsahu1909: Ah forgot to change this. I added this while testing something.

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
povinsahu1909: Okay, I'll add matrix format to test.

takagi:
"Returns" part of the rendered doc looks partially broken in style. Is it possible to fix them?
https://cupy--5331.org.readthedocs.build/en/5331/reference/generated/cupyx.scipy.sparse.linalg.lsmr.html

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -76,6 +78,286 @@ def lsqr(A, b):
...
+    # Main iteration loop.
+    while itn < maxiter:
+        itn = itn + 1
+
+        # Perform the next step of the bidiagonalization to obtain the
+        # next  beta, u, alpha, v.  These satisfy the relations
+        #         beta*u  =  a*v   -  alpha*u,
+        #        alpha*v  =  A'*u  -  beta*v.
+
+        u = A.matvec(v) - alpha * u
```
takagi, line 233: Would you use in-place oparators, i.g. `*=`, where we can? In-place operators on ndarray invoke in-place kernels and overwrite the contents of the left-hand side ndarray instance. On the other hand, Binary operators, i.g. `*`, on ndarray make a new ndarray instance for the result and allocate GPU memory from the memory pool.

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
```diff
@@ -1268,3 +1268,31 @@ def test_spilu_0(self, dtype, xp, sp):
...
+    def _make_matrix(self, sp):
+        A = testing.shaped_sparse_random((self.m, self.n), sp, density=0.2)
+        return A
+
+    def _make_normalized_vector(self, xp):
+        b = testing.shaped_random((self.m,), xp, scale=1)
+        return b
+
+    def _test_lsmr(self, xp, sp, a, b):
+        return sp.linalg.lsmr(a, b, damp=self.damp, atol=1e-6, btol=1e-6)
```
takagi, line 1290: It seems most of `lsmr` call finish with `istop=7` because of the small matrix size, where `maxiter=min([m,n])`. Would you explicitly give it a larger `maxiter`?

povinsahu1909: I think we should make `maxiter = min([m,n]) * 10` similar to what we have for `cg` and `gmres`.

takagi:
I tried an approach that makes only matrix and vector computation on GPU and other scalar arithmetic on CPU. I checked it passes the test.

https://github.com/takagi/cupy/commit/fca2ab7b7eb639daa873eae7b931bfaff1f136a3

The following is the result of a performance comparison based on your script, modifying density=0.01. With m=10000 and density=0.01, nnz is around 1,000,000 and I think it has a fairly reasonable performance improvement.

m|scipy|cupy
---|---|---
500|53.843938446044916|287.2928771972656
1000|130.22197265625|601.302734375
5000|4808.26650390625|2992.645263671875
10000|20126.594335937505|3263.73974609375

<details>
 <summary>script</summary>

```Python
import numpy as np
import cupy as cp
from cupyx.time import repeat
from scipy.sparse.linalg import lsmr as ls
import scipy.sparse
from cupyx.scipy.sparse import csr_matrix
from cupyx.scipy.sparse.linalg import lsmr

def get_time(perf):
    return max(cpu_time, gpu_time) * 1000

i = 0
times = []
print('m, scipy, cupy')
for m in (500, 1000, 5000, 10000):
    a = scipy.sparse.random(m, m, density=0.01).astype('f')
    b = np.random.rand(m).astype('f')
    x = ls(a, b)
    times.append(get_time(repeat(ls, (a,b), n_repeat=10, n_warmup=5)))
    a = csr_matrix(a)
    b = cp.array(b)
    y = lsmr(a, b)
    times.append(get_time(repeat(lsmr, (a,b), n_repeat=1, n_warmup=1)))
    print(m , times[i], times[i+1])
    i+=2
```
</details>

---

I also attach a screenshot of Nsight Systems report illustrating an iteration of lsmr. CUDA kernel invocations are now only for matrix and vector computations and are less in number compared to your original code. You can find the .qdrep file [here](https://drive.google.com/file/d/1Jo4OFd0IczYb7veSzJwzQTox55tXUmR2/view?usp=sharing) and open it on your local Nsight Systems to explore it.

screenshot2021-07-14 1 40 43

FYI, you can use `cupy.prof.time_range` to describe event ranges in a report.
https://github.com/takagi/cupy/blob/sparse_linalg_lsmr3_time_range/cupyx/scipy/sparse/linalg/_solve.py#L231
https://gist.github.com/unnonouno/877f314870d1e3a2f3f45d84de78d56c

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -76,6 +78,287 @@ def lsqr(A, b):
...
+    A = _interface.aslinearoperator(A)
+    b = b.squeeze()
+
+    m, n = A.shape
+    minDim = min([m, n])
+
+    if maxiter is None:
+        maxiter = minDim
+
+    u = b
```
takagi, line 163: This breaks the contents of `b` during the iterations. Would you assign a copy of `b` here?

povinsahu1909:
> I tried an approach that makes only matrix and vector computation on GPU and other scalar arithmetic on CPU. I checked it passes the test.

This approach gives huge performance improvement and it also reduced the overhead of using a raw kernel.

povinsahu1909:
There was a huge performance improvement from this approach. :astonished:. I think this approach can be used with other solvers as well as they have similar issue of invoking individual kernel.

cupyx/scipy/sparse/linalg/_solve.py
povinsahu1909: b was only used to get beta only single time and after that, there is no use of b. So I think this works well.

takagi: Now `u` refers to the same object with `b`, so when the contents of `u` is changed, that of `b` is also.

takagi: The contents of `u` is changed here: https://github.com/cupy/cupy/blob/405fe1a67a67b37e0d62acaaa84ca18b1a4e5332/cupyx/scipy/sparse/linalg/_solve.py#L170

povinsahu1909: I think as there is no use of b(except for beta which is calculated before modifying u) so making a copy of it would be needless.

takagi: `b` and `n` share the same ndarray, so modifying `u` also changes the contents of `b`. Please check if the following assertion passes.

```Python
import numpy as np
import scipy.sparse
import cupy as cp
from cupy import testing
from cupyx.scipy.sparse import csr_matrix
from cupyx.scipy.sparse.linalg import lsmr

m = 1000
a = scipy.sparse.random(m, m, density=0.001)
b = np.random.rand(m)
a = csr_matrix(a)
b = cp.array(b)
b0 = b.copy()
lsmr(a,b)
testing.assert_array_equal(b, b0)
print('pass')
```

povinsahu1909: Ok, now I got it.

takagi:
pfnCI, test this please.

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -78,6 +80,300 @@ def lsqr(A, b):
...
+            depending on atol and btol.
+        conlim (float): `lsmr` terminates if an estimate of ``cond(A)`` i.e.
+            condition number of matrix exceeds `conlim`. If `conlim` is None,
+            the default value is 1e+8.
+        maxiter (int): Maximum number of iterations.
+
+    Returns:
+        tuple:
+            - `x` (ndarray): Least-square solution returned.
+            - `istop` (int): istop gives the reason for stopping
```
takagi, line 120: Would you make the detail of `istop` description a literal block?
```suggestion
            - `istop` (int): istop gives the reason for stopping::
```

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -78,6 +80,300 @@ def lsqr(A, b):
     return ret
 
 
+def lsmr(A, b, damp=0.0, atol=1e-6, btol=1e-6, conlim=1e8, maxiter=None):
```
takagi, line 83: How about letting it accept `x0` for the initial guess?

takagi:
I think this is almost done except for some points. Let's see if CI passes.

cupyx/scipy/sparse/linalg/_solve.py
povinsahu1909: Yeah, I will add it.

takagi:
pfnCI, test this please.

takagi (approved):

povinsahu1909:
I tried running the tests multiple times on my system and I got no error. @takagi can you please review the test file again.

takagi:
I reproduced the error on my local. Looking into it.

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
```diff
@@ -1319,6 +1319,81 @@ def test_spilu_0(self, dtype, xp, sp):
...
+    def test_sparse(self, xp, sp):
+        if runtime.is_hip and self.format == 'csc':
+            pytest.xfail('may be buggy')  # trans=True
+        if (self.damp == 0 and self.x0 == 'ones' and self.n != 20):
+            raise unittest.SkipTest
+        a = self._make_matrix(xp)
+        a = sp.coo_matrix(a).asformat(self.format)
+        if self.use_linear_operator:
+            a = sp.linalg.aslinearoperator(a)
+        return self._test_lsmr(xp, sp, a)
```
takagi, line 1363: You compare each element of the tuples that SciPy's and CuPy's`lsmr` return, causing the CI to fail because of a larger error in some values in the tuples. I think comparing the least-square solution returned is enough.

```suggestion
        return self._test_lsmr(xp, sp, a)[0]
```

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
```diff
@@ -1319,6 +1319,81 @@ def test_spilu_0(self, dtype, xp, sp):
...
+
+    @testing.numpy_cupy_allclose(rtol=1e-1, atol=1e-1, sp_name='sp')
+    def test_dense(self, xp, sp):
+        if (self.damp == 0 and self.x0 == 'ones' and self.n != 20):
+            raise unittest.SkipTest
+        a = self._make_matrix(xp)
+        if self.use_linear_operator:
+            a = sp.linalg.aslinearoperator(a)
+        return self._test_lsmr(xp, sp, a)
```
takagi, line 1372: Ditto.

```suggestion
        return self._test_lsmr(xp, sp, a)[0]
```

takagi:
pfnCI, test this please.

takagi (approved):
LGTM!

PR #5295, opened 2021-05-30
Support `nan`, `posinf`, `neginf` in `cupy.nan_to_num`

eternalphane:
Fixes #4867.

PR #5337, opened 2021-06-14
Add tests for num_to_num's optional parameters

takagi:
Follows up #5295.

This PR adds tests for `nan_to_num`'s `nan`, `posinf`, and `neginf` parameters.

### DIFF
--- a/cupy/testing/_helper.py
+++ b/cupy/testing/_helper.py
@@ -1,3 +1,4 @@
+from collections import abc
 import contextlib
 import functools
 import inspect
@@ -226,7 +227,7 @@ def _make_positive_mask(self, impl, args, kw, name, sp_name, scipy_name):
 
 
 def _contains_signed_and_unsigned(kw):
-    vs = set(kw.values())
+    vs = set(filter(lambda x: isinstance(x, abc.Hashable), kw.values()))
     return any(d in vs for d in _unsigned_dtypes) and \
         any(d in vs for d in _float_dtypes + _signed_dtypes)
 
--- a/cupy/testing/_helper.py
+++ b/cupy/testing/_helper.py
@@ -1,4 +1,3 @@
-from collections import abc
 import contextlib
 import functools
 import inspect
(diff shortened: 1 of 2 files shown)


cupy/testing/_helper.py
```diff
@@ -226,7 +227,7 @@ def _make_positive_mask(self, impl, args, kw, name, sp_name, scipy_name):
 
 
 def _contains_signed_and_unsigned(kw):
-    vs = set(kw.values())
+    vs = set(filter(lambda x: isinstance(x, abc.Hashable), kw.values()))
```
toslunar, line 230: Could you add a comment to the function?  I suppose the function is intended to catch dtypes from `for_*_dtypes` decorators, and it was assumed that dtype should not be passed in other ways.

Is this better than `list(kw.values())`?  I would prefer `set` if the next line were `return vs.intersection(_unsigned_dtypes) and vs.intersection(_float_dtypes + _signed_dtypes)`.

tests/cupy_tests/math_tests/test_misc.py
```diff
@@ -212,6 +212,13 @@ def test_nan_to_num_nan(self):
     def test_nan_to_num_inf_nan(self):
         self.check_unary_inf_nan('nan_to_num')
 
+    def test_nan_to_num_nan_arg(self):
+        self.check_unary_nan('nan_to_num', kwargs={'nan': 1.0})
```
toslunar, line 216: How about
```suggestion
        self.check_unary_nan('nan_to_num', nan=1.0)
```
and
```diff
-def check_unary_nan(self, name, xp, dtype, kwargs={}
+def check_unary_nan(self, name, xp, dtype, **kwargs
```

takagi: Okay! I thought of it first and also that there are some methods, `check_unary` and `check_binary`, that take a keyword argument `no_bool` for another purpose than passing it to the function to test. But maybe overthinking.

https://github.com/cupy/cupy/blob/c93c05cb7d603a2b30d7b81cf25f31d039d35873/tests/cupy_tests/math_tests/test_misc.py#L17-L21

cupy/testing/_helper.py
takagi: I made this change because kwargs as a dict was passed in `kw` and it was not hashable. Now I'm going to follow https://github.com/cupy/cupy/pull/5337/files#r658514045, so we don't need it anymore.

takagi: (And, yes, using `list` would be enough even when we need some changes here)

toslunar (approved):

toslunar:
Jenkins, test this please.

PR #5342, opened 2021-06-15
Ensure source array is C-contiguous before copying to `CUDAArray`

### LINKED ISSUE #5341: `CUDAArray.copy_{from, to}()` implicitly assume C-contiguous inputs
A copy should be made internally if the input is not C-contiguous. I'll send a PR to fix this.

leofang:
Close #5341. For `copy_from` we can make a copy internally for handling non-contiguous inputs, but for `copy_to` we must require the input to be contiguous, otherwise it's not supported by `memcpy3DAsync`. 

The added tests fail on the current master.

cc: @the-lay


leofang:
Jenkins, test this please

emcastillo (approved):

PR #5335, opened 2021-06-14
Support warp shuffle and sync functions in CuPy JIT

leofang:
Part of #5280.

Example:
```python
import cupy as cp
from cupyx import jit

@jit.rawkernel()
def f_shfl(a, b):
    laneId = jit.threadIdx.x & 0x1f
    if laneId == 0:
        value = a
    value = jit.shfl_sync(0xffffffff, value, 0)
    b[laneId] = value


a = cp.int64(100)
b = cp.empty((32,), dtype=cp.int64)
f_shfl[1, 32](a, b)
print(b)


@jit.rawkernel()
def f_shfl_up(a):
    value = jit.shfl_up_sync(0xffffffff, a[jit.threadIdx.x], 5)
    a[jit.threadIdx.x] = value

a = cp.arange(32, dtype=cp.int64)
f_shfl_up[1, 32](a)
print(a)
```
Output:
```
[100 100 100 100 100 100 100 100 100 100 100 100 100 100 100 100 100 100
 100 100 100 100 100 100 100 100 100 100 100 100 100 100]
[ 0  1  2  3  4  0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18
 19 20 21 22 23 24 25 26]
```

TODO:
- [x] Support `__syncwarp()`
- [x] Add docs
- [x] Add tests

PR #5343, opened 2021-06-15
support compile-time for-loop unrolling

eternalphane:
This PR adds compile-time for-loop unrolling to CuPy JIT. The loop will be unrolled if both `start`, `stop`, `step` are `Constant`.

Example
```python
@jit.rawkernel()
def f(x, y):
    for i in range(5):
        y[i] = x[i]
x = cupy.ones(5, dtype=cupy.int32)
y = cupy.empty(5, dtype=cupy.int32)
f((1,), (1,), (x, y))
```
Generated kernel
```cpp
extern "C" __global__ void f(CArray<int, 1, true, true> x, CArray<int, 1, true, true> y) {
  {
    y[0] = x[0];
  }
  {
    y[1] = x[1];
  }
  {
    y[2] = x[2];
  }
  {
    y[3] = x[3];
  }
  {
    y[4] = x[4];
  }
}
```
Don't know if it's necessary to add unit test...


asi1024:
@eternalphane Thank you for your PR!
This PR always unrolls for-loops when the argument of `range` is a constant, but I am concerned that the huge size CUDA code will be generated when looping a huge number of times.

How about introducing `#pragma unroll`-like syntax in CuPy JIT likes as follows?

```py
@jit.rawkernel()
def f(x, y):
    with jit.unroll()
        for i in range(5):
            y[i] = x[i]
```

asi1024:
There are two possible types of implementation to generate CUDA codes.

1. Unrolled in CuPy JIT (The one suggested)

```cpp
extern "C" __global__ void f(CArray<int, 1, true, true> x, CArray<int, 1, true, true> y) {
  {
    y[0] = x[0];
  }
  {
    y[1] = x[1];
  }
   ...
```

2. Just use C++ `#pargma unroll`
```cpp
extern "C" __global__ void f(CArray<int, 1, true, true> x, CArray<int, 1, true, true> y) {
  int i;
  #pragma unroll
  for (i = 0; i < 5; ++i) {
    y[i] = x[i];
  }
}
```

I think both ways are acceptable, but the first one is better because the variable `i` can be a compile-time constant.

leofang:
`with jit.unroll(n)` is a nice idea. I'd prefer it to be 1:1 mapped to `#pragma unroll(n)` and let the C++ compiler make the decision for a given architecture (the compiler is free to ignore this instruction if it thinks it is not helping). It's easier to implement in CuPy JIT, and generate more readable code (which in turn is easier for debugging and reasoning). 

If we checkout out the Thrust/CUB repo, recently there are a few bug reports pointing out that aggressive unrolling (as is done in the current form e0d3f7c) actually hurts performance. Just my two cents.

kmaehashi:
Thanks for working on this @eternalphane 😄 

* I guess NVCC automatically unrolls in cases like `range(5)`: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#pragma-unroll
* +1 to use `#pragma unroll` and let the compiler do the optimization.
* I'm unsure if `with` block fits here, as the unroll is effective only for the loop right after the pragma. Maybe 1:1 mapping from `jit.unroll()` -> `#pragma unroll` and `jit.unroll(n)` -> `#pragma unroll(n)` is better?

leofang:
> * I'm unsure if `with` block fits here, as the unroll is effective only for the loop right after the pragma. Maybe 1:1 mapping from `jit.unroll()` -> `#pragma unroll` and `jit.unroll(n)` -> `#pragma unroll(n)` is better?

How about a language extension? 😂 We can add a keyword-only argument to `range`, so we
- map `for i in range(5, unroll=True)` to
```C
#pragma unroll
for (int i = 0; i < 5; i++) {
```
- map `for i in range(5, unroll=5)` to
```C
#pragma unroll(5)
for (int i = 0; i < 5; i++) {
```
- default to no `#pragma unroll` (which is the status quo)

I think this should be doable by extending the internal `Range`/`RangeFunc` objects.

kmaehashi:
That's interesting, but I heard @asi1024 has an idea of implementing a CPU emulation in the future, so I guess it's better not to diverge from Python's language spec?

asi1024:
CPU emulation also requires AST parse and cannot be performed by directly calling the target function. (For example, some threads may pass then-clause of the if statement, and other threads may pass else-clause.)

asi1024:
Even if not so, I agree not to diverge from Python's language spec. How about supporting `cupyx.jit.range` for the extension?

eternalphane:
I'm for supporting `cupyx.jit.range(unroll=n)`. Any other ideas? @asi1024 @kmaehashi  @leofang

leofang:
As long as it is clearly documented (hopefully with an example) I think it's fine. We need a documentation to explain the loop unrolling feature anyway.

eternalphane:
With `cupyx.jit.range`, the following code
```python
@jit.rawkernel()
def f(x, y):
    for i in jit.range(5):
        y[i] = x[i]
    for i in jit.range(5, unroll=True):
        y[i] = x[i]
    for i in jit.range(5, unroll=False):
        y[i] = x[i]
    for i in jit.range(5, unroll=5):
        y[i] = x[i]
x = cupy.ones(5, dtype=cupy.int32)
y = cupy.empty(5, dtype=cupy.int32)
f((1,), (1,), (x, y))
```
will be transpiled to
```cpp
extern "C" __global__ void f(CArray<int, 1, true, true> x, CArray<int, 1, true, true> y) {
  int i;
  for (int __it = 0, __stop = 5, __step = 1; __it < __stop; __it += __step) {
    i = __it;
    y[i] = x[i];
  }
  #pragma unroll
  for (int __it = 0, __stop = 5, __step = 1; __it < __stop; __it += __step) {
    i = __it;
    y[i] = x[i];
  }
  #pragma unroll(1)
  for (int __it = 0, __stop = 5, __step = 1; __it < __stop; __it += __step) {
    i = __it;
    y[i] = x[i];
  }
  #pragma unroll(5)
  for (int __it = 0, __stop = 5, __step = 1; __it < __stop; __it += __step) {
    i = __it;
    y[i] = x[i];
  }
}
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -15,6 +15,31 @@
...
+                If ``None``, leave the control of loop unrolling to compiler.
+
+        .. seealso:: `#pragma unroll`_
+
+        .. _#pragma unroll:
+            https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#pragma-unroll
+        """
+        super.__call__(self)
+
     def call(self, env, *args, **kwargs):
```
leofang, line 43: My impression is that @asi1024 prefers it to be explicit (https://github.com/cupy/cupy/pull/5335#discussion_r654654471):
```suggestion
    def call(self, env, *args, *, unroll=None):
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -15,6 +15,31 @@
...
+                If ``int`` type, add ``#pragma unroll(n)`` directive before the
+                loop, where ``n`` means the number of iterations to unroll.
+                If ``None``, leave the control of loop unrolling to compiler.
+
+        .. seealso:: `#pragma unroll`_
+
+        .. _#pragma unroll:
+            https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#pragma-unroll
+        """
+        super.__call__(self)
```
leofang, line 41: Nitpick: technically this is incorrect (see https://github.com/cupy/cupy/pull/5361):
```suggestion
        super().__call__()
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -28,6 +53,25 @@ def call(self, env, *args, **kwargs):
             raise TypeError(
                 f'range expected at most 3 argument, got {len(args)}')
 
+        unroll = kwargs.get('unroll')
```
leofang, line 56: Together with the change above, this can be removed:
```suggestion
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -28,6 +53,25 @@ def call(self, env, *args, **kwargs):
...
+                raise TypeError(
+                    'loop unrolling requires constant start, stop, step and '
+                    'unroll value')
+            unroll = unroll.obj
+            if not (isinstance(unroll, int) or isinstance(unroll, bool)):
+                raise TypeError(
+                    'unroll value expected to be of type int, '
+                    f'got {type(unroll).__name__}')
+            if unroll is False:
+                unroll = 1
```
leofang, line 69: I am not sure how useful it is to set `#pragma unroll(1)`, at least I haven't encountered such need in real life yet. Let's have @asi1024 make the final call 🙂

leofang (changes requested):

cupyx/jit/_builtin_funcs.py
eternalphane: @kmaehashi [mentions in the first bullet](https://github.com/cupy/cupy/pull/5343#issuecomment-862842340) that the CUDA doc states:
> By default, the compiler unrolls small loops with a known trip count.

I agree it's not a common use case, but `#pragma unroll(1)` explicitly disables unrolling, which is necessary for users to take full control of the unrolling behavior.

cupyx/jit/_builtin_funcs.py
eternalphane: `unroll` is already a keyword-only argument, note the `*args` before `unroll` 😄

cupyx/jit/_builtin_funcs.py
asi1024: The users can control it by specifying `unroll=1` instead.

kmaehashi: How about mapping `False` to `#pragma nounroll`?
https://github.com/NVlabs/xmp/blob/0052dbb206ef0cb2b47b0fe5d61f54e5cd2fa73c/src/include/digitmp/mul.h#L42

asi1024: I prefer using `"always"`, `"default"`, `"never"` literals instead of `True`, `None` and `False`.

kmaehashi: @asi1024 As `unroll` is intended for users understanding `#pragma unroll` behavior, I think these mappings are enough clear and straightforward, so +1 not to introduce our own semantics on the argument.
* `unroll=True`: `#pragma unroll`
* `unroll=False`:  `#pragma nounroll`
* `unroll=3`:  `#pragma unroll(3)`
* not specified (or `unroll=None`): `` (no pragma)
* others: error

asi1024:
Could you add tests?

cupyx/jit/_internal_types.py
```diff
@@ -47,12 +47,13 @@ def __repr__(self):
 
 class Range(Expr):
 
-    def __init__(self, start, stop, step, ctype, step_is_positive):
+    def __init__(self, start, stop, step, ctype, step_is_positive, unroll):
```
asi1024, line 50: This fix looks to change the behavior also when using the built-in range.

leofang:
Hi @eternalphane would you have time to finish this PR, or should one of us take over?

cupyx/jit/_builtin_funcs.py
leofang: Sorry, I was being silly. @eternalphane was right that `#pragma unroll(1)` is the correct directive to disable unrolling. This is documented in the Programming Guide. 

`#pragma nounroll` has no effect (`nounroll` should be done at the PTX level, not the C++ level). 

We can do a little experiment with https://godbolt.org/z/soanE1hhv by observing the output PTX length for different pragma setups.

PR #5383, opened 2021-06-22
JIT: Fix supported dtype of `atomic_add` on HIP

leofang:
On HIP `atomicAdd` does not support half precision floating points.

PR #5387, opened 2021-06-23
JIT: Support all atomic intrinsics

leofang:
Part of #5280 (and #4290). ~~Please merge after #5383.~~

TODO:
- [x] Check lowest supported CUDA/HIP version
- [x] Add docs


leofang:
@asi1024 Now that I think about it, I don't really understand Numba's atomic API design (which CuPy follows?) I think we should be able to follow closely with the C API and support `jit.atomic_add(array[index], value)` instead of the current approach (`jit.atomic_add(array, index, value)`), as we have `__getitem__` properly implemented in JIT. The understanding is we take the pointer to `array[index]`. Do I miss something?

leofang:
> Do I miss something?

Maybe it's not easy to construct `target = _compile._indexing(array, index, env)` when we are given only `array[index]`...

asi1024:
> Maybe it's not easy to construct target = _compile._indexing(array, index, env) when we are given only array[index]...

I think so too. It is difficult to handle pointers directly in Python syntax.

asi1024:
Jenkins, test this please.

tests/cupyx_tests/jit_tests/test_raw.py
```diff
@@ -307,6 +307,162 @@ def f(x, index, out):
...
+                jit.atomic_dec(out, tid, x[tid])
+
+        x = cupy.zeros(1024, dtype=dtype)
+        out = cupy.arange(1024, dtype=dtype)
+        f((32,), (32,), (x, out))
+        expected = x
+        self._check(out, expected)
+
+    @testing.for_dtypes('iILQ' if runtime.is_hip else 'iHILQ')
+    def test_atomic_cas(self, dtype):
```
asi1024, line 398: Could you skip uint16 test in old CUDA version?

leofang: Oops I forgot about that! Thanks @asi1024. Fixed in 3dae766.

asi1024:
Jenkins, test this please.

leofang:
Jenkins, test this please

leofang:
For the `atomicCAS` error: it looks like only sm70 (and above) supports the `unsigned short` version:
```
$ grep -R /usr/local/cuda-10.2/include/ --include=*.{h,} -e 'atomicCAS\>'
/usr/local/cuda-10.2/include/device_atomic_functions.h:__DEVICE_ATOMIC_FUNCTIONS_DECL__ int atomicCAS(int *address, int compare, int val) __DEF_IF_HOST
/usr/local/cuda-10.2/include/device_atomic_functions.h:__DEVICE_ATOMIC_FUNCTIONS_DECL__ unsigned int atomicCAS(unsigned int *address, unsigned int compare, unsigned int val) __DEF_IF_HOST
/usr/local/cuda-10.2/include/device_atomic_functions.h:__DEVICE_ATOMIC_FUNCTIONS_DECL__ unsigned long long int atomicCAS(unsigned long long int *address, unsigned long long int compare, unsigned long long int val) __DEF_IF_HOST
/usr/local/cuda-10.2/include/crt/sm_70_rt.h:__SM_70_RT_DECL__ unsigned short int atomicCAS(unsigned short int *address, unsigned short int compare, unsigned short int val) __DEF_IF_HOST
```

leofang:
@asi1024 any idea for the cause of `atomicDec` error on Windows? 😅

asi1024:
@leofang Probably because uint32 character is 'L' on windows. I shouldn't have implemented `AtomicOp` class using dtype character ...

https://gist.github.com/asi1024/60da8cf4d45db6c772ce59643a3b972e

leofang:
Ouch! But, isn't "I" representing `numpy.uintc` which in turn means uint32?

asi1024:
`dtype.name` might be better than `dtype.char`?

leofang:
It's so confusing...Why didn't `atomic_inc` fail on Windows like `atomic_dec`? They support the same type (uint32), but only the latter fails?!

leofang:
Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
>> `dtype.name` might be better than `dtype.char`?

This sounds good, but I couldn't think of a way to make use of it if the atomic functions are initialized with the dtype characters, so I added a little workaround which should hopefully deliver the desired behavior.

> It's so confusing...Why didn't `atomic_inc` fail on Windows like `atomic_dec`? They support the same type (uint32), but only the latter fails?!

Still can't figure out this...

Jenkins, test this please

asi1024:
I can't figure out why didn't atomic_inc fail too, but I think this PR is OK with the test workaround!

asi1024 (approved):

leofang:
Thanks, @asi1024!

PR #5327, opened 2021-06-08
Fix cuDNN preloading

kmaehashi:
Currently some of features in cuDNN v8 is not working when installed via wheels.

```py
import ctypes
import cupy
import cupy.cuda.cudnn as libcudnn
import cupy.cudnn
x = cupy.arange(12, dtype=cupy.float32).reshape(3, 4)

# This is implemented in cudnn_ops_infer:
cupy.cudnn.activation_forward(x, libcudnn.CUDNN_ACTIVATION_RELU)
# This is implemented in cudnn_cnn_train (error)
cupy.cudnn.create_fused_ops_plan(0)  # CUDNN_FUSED_SCALE_BIAS_ACTIVATION_CONV_BNSTATS
```

The above example fails because `cudnn_cnn_train` depends on `cudnn_ops_infer`, `cudnn_ops_train`, and `cudnn_cnn_infer`, but `cudnn_ops_train` and `cudnn_cnn_infer` are not loaded and could not be discovered automatically by cuDNN.
This issue only occurs when chain-loading dependencies (i.e., loading a split module from another split module).
`libcudnn.so.8` has a `$ORIGIN` in RUNPATH, so `libcudnn.so.8` can find split modules in the same directory and load them.

```
~/.cupy/cuda_lib/11.2/cudnn/8.2.0/lib64 % for FILE in libcudnn*8.2.0; do echo $FILE; readelf -d $FILE | grep RUNPATH; done
libcudnn_adv_infer.so.8.2.0
libcudnn_adv_train.so.8.2.0
libcudnn_cnn_infer.so.8.2.0
libcudnn_cnn_train.so.8.2.0
libcudnn_ops_infer.so.8.2.0
libcudnn_ops_train.so.8.2.0
libcudnn.so.8.2.0
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]
```

This PR fixes the preload system to load all split modules.

This changes the wheel metadata structure but does not affect conda-forge distributions.

xref: #5103

TODO:
- [x] verify this works on Windows

PR #5404, opened 2021-06-28
Fix preloading slow tests

kmaehashi:
This fixes failues in slow test.
Follow-up: https://github.com/cupy/cupy/pull/5327


kmaehashi:
pfnCI, test this please.

PR #5410, opened 2021-06-29
Fix astype from boolean

### LINKED ISSUE #4354: suprising behavior of cupy.array on numpy views with base array of a different dtype
The following example illustrates that creating CuPy array from a NumPy view where the underlying array has a different base dtype can result in unexpected behavior. Here is a minimal concrete example for a boolean view of a `uint8` array. In NumPy calling `np.array()` creates a copy. In CuPy, the created array has flags.owndata=True as expected, so it looks like it is a copy, but the underlying data still correspond to the view! This can be seen if converting back from `bool`->`uint8` via astype, values > 1 occur!

CuPy 9.0.0a2

* Code to reproduce

```Python
import numpy as np
import cupy as cp

# boolean view of a uint8 array
a = np.array([0, 3, 1, 2, 0, 1], dtype=np.uint8)
a_view = a.view(np.bool)
print(f"a_view.flags = {a_view.flags}")

# calling np.array makes a copy as expected
b = np.array(a_view)  # will make a copy!
print(f"b.flags = {b.flags}")
print(f"b.astype(int) = {b.astype(int)}\n")  # only 1 and 0 as expected

# create CuPy array from the NumPy view
gpu_a = cp.array(a_view)
print(f"gpu_a = {gpu_a}")
print(f"gpu_a.astype(int) = {gpu_a.astype(int)}")   # unexpected values of 2 and 3!
print(f"gpu_a.flags = {gpu_a.flags}")  # flags look like it is a copy (owndata=True)
```
gives
```
a_view.flags =   C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

b.flags =   C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

b.astype(int) = [0 1 1 1 0 1]

gpu_a = [False  True  True  True False  True]
gpu_a.astype(int) = [0 3 1 2 0 1]
gpu_a.flags =   C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
```

Possible Solution
=============
Possibly need a check on the input object in cp.array that looks something like the following
```Python
if isinstance(obj, numpy.ndarray) and not obj.flags.owndata and obj.base.dtype != obj.dtype:
    obj = numpy.copy(obj)
```
or if a copy is not requested then perhaps creating the cupy.ndarray from obj.base and then create a view?

takagi:
Fix #4354.

The result of `astype` from a ndarray whose dtype is boolean to another dtype is expected to be zero or one. However, its underlying representation is not necessarily zero or one (i.g. a view). And in such a case, `astype` returns a ndarray with values other than zero or one. This PR fixes `astype` from boolean ndarray to return the proper value.

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -462,6 +462,14 @@ cdef class ndarray:
                 'Casting complex values to real discards the imaginary part',
                 numpy.ComplexWarning)
             elementwise_copy(self.real, newarray)
+        elif self.dtype.kind == 'b':
+            # See #4354. The result of `astype` from a ndarray whose dtype is
+            # boolean to another dtype is expected to be zero or one. However,
+            # its underlying representation is not necessarily zero or one
+            # (i.g. a view). In such a case,`elementwise_copy` copies the data
+            # as it is, resulting in an undesirable output. To keep it off, we
+            # give a special path for a boolean ndarray.
+            cupy.not_equal(self, 0, out=newarray)
         else:
             elementwise_copy(self, newarray)
         return newarray
--- a/tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
+++ b/tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
@@ -155,6 +155,12 @@ class TestArrayCopyAndView(unittest.TestCase):
                                      xp.empty((2, 3, 2), dtype=src_dtype))
         return astype_without_warning(src, dst_dtype, order='K').strides
 
+    @testing.numpy_cupy_array_equal()
+    def test_astype_boolean_view(self, xp):
+        # See #4354
+        a = xp.array([0, 1, 2], dtype=numpy.int8).view(dtype=numpy.bool_)
+        return a.astype(numpy.int8)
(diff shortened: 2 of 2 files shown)


toslunar:
Does the change also fix automatic conversion from boolean?

```
>>> for xp in (numpy, cupy): xp.arange(3, dtype=xp.int8).view(bool) + 4.5
...
array([4.5, 5.5, 5.5])
array([4.5, 5.5, 6.5])
```

takagi:
I got the same result with this change, so it seems to be another issue.

toslunar (approved):
LGTM because fixing the `astype` issue can have higher priority than other type conversions.

toslunar:
Jenkins, test this please.

kmaehashi (approved):

PR #5429, opened 2021-07-01
Add binomial distribution to new Generator

povinsahu1909:


cupy/random/_generator_api.pyx
```diff
@@ -495,6 +498,30 @@ class Generator:
...
+        cdef ndarray p_arr
+
+        n = cupy.asarray(n)
+        p = cupy.asarray(p)
+
+        if size is None:
+            size = cupy.broadcast(n, p).shape
+
+        y = ndarray(size if size is not None else (), numpy.int64)
+
```
emcastillo, line 514: need to add several checks as in https://github.com/cupy/cupy/pull/5182

cupy/random/cupy_distributions.cu
```diff
@@ -12,6 +12,11 @@
 
 struct rk_state {
 
+#ifdef CUPY_USE_BINOMIAL
+    int has_binomial;
+    int nsave, m;
+    double psave, r, q, fm, p1, xm, xl, xr, c, laml, lamr, p2, p3, p4;
```
emcastillo, line 18: To be honest, i'd rather have this in a different struct rather than rk_state

cupy/random/cupy_distributions.cu
```diff
@@ -299,6 +305,175 @@ __device__ uint64_t rk_interval_64(rk_state* state, uint64_t  mx, uint64_t mask)
...
+        xr = state->xr;
+        c = state->c;
+        laml = state->laml;
+        lamr = state->lamr;
+        p2 = state->p2;
+        p3 = state->p3;
+        p4 = state->p4;
+    }
+  /* sigh ... */
+  Step10:
```
emcastillo, line 352: could this be rewritten to use inline functions?

emcastillo (changes requested):
Keep on the good work!
Thanks!

cupy/random/cupy_distributions.cu
povinsahu1909: AFAIK, we cannot use inline functions here as both are different variables one is a field of the struct and other is a normal variable

cupy/random/cupy_distributions.cu
povinsahu1909: Yeah writing this in a separate struct will increase readability.

cupy/random/_generator_api.pyx
povinsahu1909: Should I also add checks for the input domain like for `p >= 0 and p<=1`?

povinsahu1909: This check is missing in the distributions.

emcastillo: that check involves device synchronization so I think it is ok not do it, but leave it as a comment

also p can be an array so check the above PR to deal with it accordingly :)

cupy/random/cupy_distributions.cu
povinsahu1909: I am writing separate class for this, but while doing so I have a doubt. How to use the multiple structs in generator dispatcher.

takagi: You can use a ndarray that has the same size in bytes as the size of the other struct to keep the additional state. The ndarray's type can be `uint8`.

emcastillo: https://github.com/cupy/cupy/blob/995ce208dc475525c37b6a449e64d7e40bcb5fdc/cupy/random/_bit_generator.pyx#L77

Yes, this is what we do to allocate the state, you can do this for the binomial and pass the pointer as a parameter to the distribution

takagi:
- Added `rk_binomial_state`, which is besides `rk_state` and managed by `Generator` instead of `BitGenerator.`
- Modified `get_index` as I had new cases: taking `rk_binomial_state*` that is a pointer but not represented using `_array_data`, and returning a value of a type that is not `double.`

takagi:
@emcastillo I made some fixes on binomial distribution. Would you take a look to see if my change makes sense?

emcastillo:
> @emcastillo I made some fixes on binomial distribution. Would you take a look to see if my change makes sense?

 it looks very nice!! thanks :D

takagi:
Would you resolve the conflict and fill in the docstring to make this ready to merge, @povinsahu1909?

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi:
pfnCI, test this please.

takagi (dismissed):

takagi:
pfnCI, test this please.

takagi:
Now waiting for CI pass.

emcastillo (approved):
LGTM! Thanks a lot!!

takagi:
Thanks, @povinsahu1909!

PR #5434, opened 2021-07-02
Support CUDA 11.4 and support `compute_86`

takagi:
Blocked by https://github.com/chainer/chainer-test/pull/673.

Includes the following updates for CUDA 11.4:
- Compute capability support
- bundled FP16 header files
- library installer

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -133,7 +133,7 @@ def _get_arch():
         # CUDA 10.1 / 10.2
         _nvrtc_max_compute_capability = '75'
     else:
-        # CUDA 11.0 / 11.1 / 11.2
+        # CUDA 11.0 / 11.1 / 11.2 / 11.3 / 11.4
         _nvrtc_max_compute_capability = '80'
 
     arch = device.Device().compute_capability
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -132,9 +132,15 @@ def _get_arch():
     elif major < 11:
         # CUDA 10.1 / 10.2
         _nvrtc_max_compute_capability = '75'
-    else:
-        # CUDA 11.0 / 11.1 / 11.2 / 11.3 / 11.4
+    elif major == 11 and minor == 0:
+        # CUDA 11.0
         _nvrtc_max_compute_capability = '80'
+    elif major == 11 and minor < 4:
+        # CUDA 11.1 / 11.2 / 11.3
+        _nvrtc_max_compute_capability = '86'
+    else:
+        # CUDA 11.4
+        _nvrtc_max_compute_capability = '87'
 
     arch = device.Device().compute_capability
     if arch in _tegra_archs:
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -135,12 +135,9 @@ def _get_arch():
     elif major == 11 and minor == 0:
         # CUDA 11.0
         _nvrtc_max_compute_capability = '80'
-    elif major == 11 and minor < 4:
-        # CUDA 11.1 / 11.2 / 11.3
-        _nvrtc_max_compute_capability = '86'
     else:
-        # CUDA 11.4
-        _nvrtc_max_compute_capability = '87'
+        # CUDA 11.1 / 11.2 / 11.3 / 11.4
+        _nvrtc_max_compute_capability = '86'
 
     arch = device.Device().compute_capability
     if arch in _tegra_archs:
--- a/cupy_setup_build.py
+++ b/cupy_setup_build.py
@@ -1004,18 +1004,7 @@ def _nvcc_gencode_options(cuda_version):
         #
         #   https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation
 
-        if cuda_version >= 11040:
-            arch_list = ['compute_35',
-                         'compute_50',
-                         ('compute_60', 'sm_60'),
-                         ('compute_61', 'sm_61'),
-                         ('compute_70', 'sm_70'),
-                         ('compute_75', 'sm_75'),
-                         ('compute_80', 'sm_80'),
-                         ('compute_86', 'sm_86'),
-                         ('compute_87', 'sm_87'),
-                         'compute_87']
-        elif cuda_version >= 11010:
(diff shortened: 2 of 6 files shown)


kmaehashi:
Hmm, seems NVRTC docs is outdated.

CUDA 11.2 / 11.3
```
>>> cupy.cuda.nvrtc.getSupportedArchs()
(35, 37, 50, 52, 53, 60, 61, 62, 70, 72, 75, 80, 86)
```

CUDA 11.4
```
>>> cupy.cuda.nvrtc.getSupportedArchs()
(35, 37, 50, 52, 53, 60, 61, 62, 70, 72, 75, 80, 86, 87)
```

takagi:
I also suspect nvcc's. NVCC accepts `compute_86/sm_86` from CUDA 11.1.

kmaehashi:
> I also suspect nvcc's. NVCC accepts compute_86/sm_86 from CUDA 11.1.

I see... `compute_86` is used in the official CUDA v11.1 examples, so I think it's safe to assume it's supported since 11.1.
https://github.com/NVIDIA/cuda-samples/blob/v11.1/Samples/bf16TensorCoreGemm/bf16TensorCoreGemm_vs2019.vcxproj

cupyx/tools/install_library.py
```diff
@@ -174,6 +182,9 @@ def _make_nccl_record(
     }
 
 
+_nccl_records.append(_make_nccl_record(
```
kmaehashi, line 185: Could you use 2.10.3?

cupy_setup_build.py
```diff
@@ -1004,7 +1004,17 @@ def _nvcc_gencode_options(cuda_version):
...
+        if cuda_version >= 11040:
+            arch_list = ['compute_35',
+                         'compute_50',
+                         ('compute_60', 'sm_60'),
+                         ('compute_61', 'sm_61'),
+                         ('compute_70', 'sm_70'),
+                         ('compute_75', 'sm_75'),
+                         ('compute_80', 'sm_80'),
+                         ('compute_86', 'sm_86'),
+                         'compute_86']
```
kmaehashi, line 1016: This entry is intended as a fallback for all `compute_8X`.
```suggestion
                         'compute_80']
```

cupy/cuda/compiler.py
```diff
@@ -133,7 +133,7 @@ def _get_arch():
         # CUDA 10.1 / 10.2
         _nvrtc_max_compute_capability = '75'
     else:
-        # CUDA 11.0 / 11.1 / 11.2
+        # CUDA 11.0 / 11.1 / 11.2 / 11.3 / 11.4
         _nvrtc_max_compute_capability = '80'
```
kmaehashi, line 137: Could you allow 86 for CUDA 11.1+?

cupy_setup_build.py
```diff
@@ -1004,7 +1004,17 @@ def _nvcc_gencode_options(cuda_version):
         #
         #   https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation
 
-        if cuda_version >= 11000:
+        if cuda_version >= 11040:
```
kmaehashi, line 1007:
```suggestion
        if cuda_version >= 11010:
```

kmaehashi (changes requested):

kmaehashi:
Let me test `pfn-public-ci/cupy.linux.cuda114`.
/test

cupy_setup_build.py
takagi: I think GPU with cc=80 and cc=86 are covered by the two lines above, and the line is for forward compatibility for future cc>86. Also I think nVidia doesn’t insert a new cc number below the maximum one. Do I have some misunderstanding?

takagi:
Ah, sorry, I commited a suggested change before the test run finishes, and it cleared the CI status on github. (The test itself keeps running? https://ci.preferred.jp/cupy.linux.cuda114/74868/)

cupy_setup_build.py
kmaehashi: Ah, you are correct, that seems like a correct approach to provide forward compatibility 😄 
https://github.com/NVIDIA/cuda-samples/blob/v11.4/Samples/vectorAddMMAP/Makefile#L323-L328

cupyx/tools/install_library.py
kmaehashi: Let's do this in separate NCCL 2.10 PR 😄

cupy/cuda/compiler.py
takagi: Can we also add that of `87`? It seems that nvcc 11.4 also supports `compute_87/sm_87`.

cupy_setup_build.py
```diff
@@ -1004,7 +1004,27 @@ def _nvcc_gencode_options(cuda_version):
...
+            arch_list = ['compute_50',
```
takagi, line 1008: Kepler `compute_35` support is deprecated in CUDA 11.4.

cupy/cuda/compiler.py
kmaehashi: I think it's safer to keep `86` here (when the new GPU with CC = 90 is released users may hit this) as it's undocumented and CUDA examples don't use it.
https://github.com/NVIDIA/cuda-samples/search?l=Makefile&q=SMS

cupy_setup_build.py
```diff
@@ -1004,7 +1004,28 @@ def _nvcc_gencode_options(cuda_version):
...
+            arch_list = ['compute_35',
+                         'compute_50',
+                         ('compute_60', 'sm_60'),
+                         ('compute_61', 'sm_61'),
+                         ('compute_70', 'sm_70'),
+                         ('compute_75', 'sm_75'),
+                         ('compute_80', 'sm_80'),
+                         ('compute_86', 'sm_86'),
+                         ('compute_87', 'sm_87'),
+                         'compute_87']
```
kmaehashi, line 1017:
```suggestion
                         'compute_86']
```

kmaehashi (changes requested):

cupyx/tools/install_library.py
takagi: Sure!

takagi:
pfnCI, test this please.

kmaehashi (approved):

PR #5435, opened 2021-07-02
docs: user_guide/basic.rst: impr. device agnostic section

maxim-belkin:
Various improvements and fixes to `basic.rst` but mainly to the "Device agnostic code" section.

### DIFF
--- a/docs/source/user_guide/basic.rst
+++ b/docs/source/user_guide/basic.rst
@@ -115,8 +115,8 @@ Current Stream
 --------------
 
 Associated with the concept of current devices are *current streams*, which help avoid explicitly passing streams
-in every single operation so as to keep the APIs pythonic and user-friendly. In CuPy, any CUDA operations
-such as data transfer (see the next section) and kernel launches are enqueued onto the current stream,
+in every single operation so as to keep the APIs pythonic and user-friendly. In CuPy, all CUDA operations
+such as data transfer (see the :ref:`data-transfer-basics` section) and kernel launches are enqueued onto the current stream,
 and the queued tasks on the same stream will be executed in serial (but *asynchronously* with respect to the host).
 
 The default current stream in CuPy is CUDA's null stream (i.e., stream 0). It is also known as the *legacy*
@@ -127,6 +127,7 @@ retrieved using :func:`cupy.cuda.get_current_stream`.
 It is worth noting that CuPy's current stream is managed on a *per thread, per device* basis, meaning that on different
 Python threads or different devices the current stream (if not the null stream) can be different.
 
+.. _data-transfer-basics:
 
 Data Transfer
 -------------
@@ -180,34 +181,39 @@ We can also use :meth:`cupy.ndarray.get()`:
 Memory management
 -----------------
 
-Check :doc:`./memory` for a detailed description of how is memory managed in CuPy
+Check :doc:`./memory` for a detailed description of how memory is managed in CuPy
 using memory pools.
 
 
-How to write CPU/GPU agnostic code
+How to write device agnostic code
 ----------------------------------
 
-The compatibility of CuPy with NumPy enables us to write CPU/GPU generic code.
-It can be made easy by the :func:`cupy.get_array_module` function.
-This function returns the :mod:`numpy` or :mod:`cupy` module based on arguments.
-A CPU/GPU generic function is defined using it like follows:
+CuPy's compatibility with NumPy makes it possible to write device agnostic code.
+For this purpose, CuPy implements the :func:`cupy.get_array_module` function, that
+returns a reference to :mod:`cupy` if any of its arguments resides on a GPU
+and :mod:`numpy` otherwise.
+Here is an example of a device agnostic function that computes ``log1p``:
 
 .. doctest::
 
    >>> # Stable implementation of log(1 + exp(x))
    >>> def softplus(x):
-   ...     xp = cp.get_array_module(x)
-   ...     return xp.maximum(0, x) + xp.log1p(xp.exp(-abs(x)))
+   ...     mod = cp.get_array_module(x)
+   ...     print("Using module:", mod.__name__)
+   ...     return mod.maximum(0, x) + mod.log1p(mod.exp(-abs(x)))
 
-Sometimes, an explicit conversion to a host or device array may be required.
-:func:`cupy.asarray` and :func:`cupy.asnumpy` can be used in agnostic implementations
-to get host or device arrays from either CuPy or NumPy arrays.
+When you need to operate on arrays that reside on CPU and GPU,
+you first need to transfer them to the same device -- either CPU or GPU.
+For this purpose, CuPy implements two sister methods called :func:`cupy.asnumpy`  and
+:func:`cupy.asarray`. Here is an example that demonstrates the use of both methods:
 
 .. doctest::
 
+   >>> x_cpu = np.array([1, 2, 3])
    >>> y_cpu = np.array([4, 5, 6])
    >>> x_cpu + y_cpu
    array([5, 7, 9])
+   >>> x_gpu = cp.asarray(x_cpu)
    >>> x_gpu + y_cpu
    Traceback (most recent call last):
    ...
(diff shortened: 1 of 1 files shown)


docs/source/user_guide/basic.rst
```diff
@@ -180,34 +181,39 @@ We can also use :meth:`cupy.ndarray.get()`:
 Memory management
 -----------------
 
-Check :doc:`./memory` for a detailed description of how is memory managed in CuPy
+Check :doc:`./memory` for a detailed description of how memory is managed in CuPy
 using memory pools.
 
 
-How to write CPU/GPU agnostic code
+How to write device agnostic code
```
leofang, line 188: I'd prefer the "CPU/GPU agnostic" terminology in this section to remain unchanged. "Device agnostic" has a different meaning, usually associated with code portability (ex; vendor-neutral support such as the support of NVIDIA/AMD GPUs). We should add a section on this point in a separate PR.

docs/source/user_guide/basic.rst
```diff
@@ -180,34 +181,39 @@ We can also use :meth:`cupy.ndarray.get()`:
...
 ----------------------------------
 
-The compatibility of CuPy with NumPy enables us to write CPU/GPU generic code.
-It can be made easy by the :func:`cupy.get_array_module` function.
-This function returns the :mod:`numpy` or :mod:`cupy` module based on arguments.
-A CPU/GPU generic function is defined using it like follows:
+CuPy's compatibility with NumPy makes it possible to write device agnostic code.
+For this purpose, CuPy implements the :func:`cupy.get_array_module` function, that
```
leofang, line 192: No comma.
```suggestion
For this purpose, CuPy implements the :func:`cupy.get_array_module` function that
```

docs/source/user_guide/basic.rst
```diff
@@ -180,34 +181,39 @@ We can also use :meth:`cupy.ndarray.get()`:
...
 
 .. doctest::
 
    >>> # Stable implementation of log(1 + exp(x))
    >>> def softplus(x):
-   ...     xp = cp.get_array_module(x)
-   ...     return xp.maximum(0, x) + xp.log1p(xp.exp(-abs(x)))
+   ...     mod = cp.get_array_module(x)
+   ...     print("Using module:", mod.__name__)
+   ...     return mod.maximum(0, x) + mod.log1p(mod.exp(-abs(x)))
```
leofang, line 203: Please revert this example. `xp` is a standard usage that already prevails in the community.

docs/source/user_guide/basic.rst
```diff
@@ -180,34 +181,39 @@ We can also use :meth:`cupy.ndarray.get()`:
...
 
-Sometimes, an explicit conversion to a host or device array may be required.
-:func:`cupy.asarray` and :func:`cupy.asnumpy` can be used in agnostic implementations
-to get host or device arrays from either CuPy or NumPy arrays.
+When you need to operate on arrays that reside on CPU and GPU,
+you first need to transfer them to the same device -- either CPU or GPU.
```
leofang, line 206: The new description reads odd to me, mainly because 1. I don't see it offers better clarity compared to the old one, and 2. "CPU" usually is not considered a device. How about this?
```suggestion
When you need to manipulate CPU and GPU arrays, an explicit data
transfer may be required to move them to the same location -- either CPU or GPU.
```

docs/source/user_guide/basic.rst
```diff
@@ -180,34 +181,39 @@ We can also use :meth:`cupy.ndarray.get()`:
...
+For this purpose, CuPy implements two sister methods called :func:`cupy.asnumpy`  and
+:func:`cupy.asarray`. Here is an example that demonstrates the use of both methods:
 
 .. doctest::
 
+   >>> x_cpu = np.array([1, 2, 3])
    >>> y_cpu = np.array([4, 5, 6])
    >>> x_cpu + y_cpu
    array([5, 7, 9])
+   >>> x_gpu = cp.asarray(x_cpu)
```
leofang, line 216: Good catch! This makes the snippet self-contained.

docs/source/user_guide/basic.rst
```diff
@@ -220,3 +226,8 @@ to get host or device arrays from either CuPy or NumPy arrays.
    array([5, 7, 9])
    >>> cp.asarray(x_gpu) + cp.asarray(y_cpu)
    array([5, 7, 9])
+
+The :func:`cupy.asnumpy` method returns a NumPy array (array on the host),
+whereas :func:`cupy.asarray` method returns a CuPy array (array on the current device).
+Both of these methods act on an arbitrary source, meaning that they can be applied to any data
+on the host or device that can be converted to an array.
```
leofang, line 233: nitpick:
```suggestion
Both methods can accept arbitrary input, meaning that they can be applied to any data that
is located on either the host or device and can be converted to an array.
```

leofang (commented):
Thanks for the improvements! I left a few comments.

docs/source/user_guide/basic.rst
maxim-belkin: Thanks for the feedback/comments, @leofang. I agree with all of them except this one. I understand that saying "device agnostic" is not exactly right and that the word "device" is primarily used to mean GPUs, but I feel that saying "CPU/GPU agnostic code" is a bit misleading. The slash usually means "or", so if we consider only one part (say, CPU) -- it would become "CPU agnostic", which implies that code is agnostic of CPU it runs on (which is, probably, true, but not what this section is about). Should it be "CPU-GPU agnostic", perhaps? I'm OK leaving this "as-was" but it actually confused me when I was reading it, hence the PR.

docs/source/user_guide/basic.rst
maxim-belkin: I'm OK with reverting this, but wanted to note that this document targets those who are not part of the community yet. So, do you think a note about `xp` (in the form of a Python comment) would fit this example?

docs/source/user_guide/basic.rst
maxim-belkin: I reverted this change (and changed the next paragraph accordingly) but I suggest to take a closer look at this at some point in the future.

leofang: > The slash usually means "or", so if we consider only one part (say, CPU) -- it would become "CPU agnostic", which implies that code is agnostic of CPU it runs on (which is, probably, true, but not what this section is about). Should it be "CPU-GPU agnostic", perhaps? I'm OK leaving this "as-was" but it actually confused me when I was reading it, hence the PR.

Oh, OK, so this was what confused you. I'd say it's kinda a common term used in the field --- If you google it, you'd find way more entries using "CPU/GPU agnostic" than using "CPU-GPU agnostic", although a portion of those are focusing on CuPy/Chainer 😄 

Not that a forward slash has overloaded meanings in English --- in addition to "or", it can also be used for punctuation, indicating connection/conflict, or **"and"**. (See, I just said "CuPy/Chainer".)

docs/source/user_guide/basic.rst
leofang: > this document targets those who are not part of the community yet

I wouldn't say so. The current documentation is written under an (implicit) assumption that users are already familiar with NumPy (so `import numpy as np` shouldn't be considered foreign) and basic terminology of GPU programming (ex: users need to know the distinction between host and device, data locality, etc). This, compounded with `import cupy as cp`, renders `xp` here an easy-to-understand convention.   

IIRC there is a plan to write a new piece of docs targeting beginners, just like NumPy's (https://numpy.org/doc/stable/user/absolute_beginners.html). We could explain this convention there. (Maybe you're interested in contributing to this write-up? 😇)

leofang (approved):
LGTM! Someone from the team will take a second look. Thanks 👍

docs/source/user_guide/basic.rst
maxim-belkin: > > this document targets those who are not part of the community yet
> 
> I wouldn't say so. The current documentation is written under an (implicit) assumption that users are already familiar with NumPy (so `import numpy as np` shouldn't be considered foreign) and basic terminology of GPU programming (ex: users need to know the distinction between host and device, data locality, etc). This, compounded with `import cupy as cp`, renders `xp` here an easy-to-understand convention.

I'm sorry I wasn't clear -- I was talking about the CuPy community, not NumPy/GPU programming beginners: this is the "Basics" page of the CuPy's user's guide, so the target audience is those who are unfamiliar with CuPy. For example, I know about `numpy/np` convention and this page introduces `cupy/cp` but `xp` didn't strike me as super obvious. I think once the reader's knowledge of `cupy/cp` convention solidifies, `xp` var. name will make more sense.

kmaehashi (approved):

kmaehashi:
Thank you for improvements! LGTM 👍🏼

kmaehashi:
pfnCI, test this please.

kmaehashi:
Test failure is related and will be fixed in #5449.

PR #5426, opened 2021-07-01
Fix `unwrap` tests for v9

emcastillo:
This should fix the tests

PR #5439, opened 2021-07-05
Fix mergify to only comment when pull-request is open

kmaehashi:
Maybe `closed=false` condition is needed to avoid this: https://github.com/cupy/cupy/pull/5426#issuecomment-873896502
Just a wild guess.


emcastillo (approved):
Merging as this does not require CI

PR #5443, opened 2021-07-05
Add geometric distribution to new Generator

povinsahu1909:


cupy/random/_generator_api.pyx
```diff
@@ -262,6 +265,54 @@ class Generator:
...
+        """Geometric distribution.
+
+        Returns an array of samples drawn from the geometric distribution. Its
+        probability mass function is defined as
+
+        .. math::
+            f(x) = p(1-p)^{k-1}.
+
+        Args:
+            p (float): Success probability of the geometric distribution.
```
emcastillo, line 278: p can be an array of floats
check numpy

https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.geometric.html#numpy.random.Generator.geometric

tests/cupy_tests/random_tests/common_distributions.py
```diff
@@ -268,3 +268,23 @@ def test_gamma_2(self):
     def test_gamma_ks(self):
         self.check_ks(0.05)(
             self.shape, self.scale, size=2000)
+
+
+geometric_params = [
+    {'p': 0.5},
+    {'p': 0.1},
+    {'p': 1.0},
```
emcastillo, line 276: test also arrays please

emcastillo (commented):
test in https://github.com/cupy/cupy/blob/master/tests/cupy_tests/random_tests/test_generator.py is not refactored, should be like

https://github.com/cupy/cupy/blob/master/tests/cupy_tests/random_tests/test_generator.py#L576-L580

tests/cupy_tests/random_tests/common_distributions.py
povinsahu1909: I tried passing the array as test parameters, but as of now we have to explicitly pass p as cupy.ndarray for distributions of Generator, but in RandomState there is no such requirement. And tests fails due to broadcast error.

takagi: > And tests fails due to broadcast error.

Maybe the shape of `p` and `size` make a conflict?

cupy/random/_generator_api.pyx
```diff
@@ -262,6 +265,55 @@ class Generator:
...
+        else:
+            # Check if size is broadcastable to shape
+            # but size determines the output
+            p = p.astype('d', copy=False)
+
+        if size is not None and not isinstance(size, tuple):
+            size = (size,)
+        elif size is None:
+            size = p.shape
+        y = ndarray(size if size is not None else(), numpy.int64)
```
takagi, line 309:
```suggestion
        y = ndarray(size if size is not None else (), numpy.int64)
```

takagi:
I think this PR is basically fine except for some points left. Would you make a performance comparison against SciPy? Also, I'd like to merge this in this week if possible as we have GSoC's first evaluation next week. It's clearly plus for you if you make a merged PR before it.

tests/cupy_tests/random_tests/common_distributions.py
povinsahu1909: https://github.com/cupy/cupy/blob/d53212bb87579013ec8a1c4b8d31de8512f92b40/tests/cupy_tests/random_tests/common_distributions.py#L111-L116

Is there any specific se of defining the numpy_len = 1000? 
As this is causing the broadcast error. I replaced the above line with following and it works well for the p as ndarray, but was failing for the `lognormal` and `normal` by rejecting null hypothesis.

```python
# kwargs['size'] = numpy_len
dtype = kwargs.pop('dtype', None)
numpy_rng = self.get_rng(numpy, self.__seed)
func_numpy = getattr(numpy_rng, self.target_method)(*args, **kwargs)
vals_numpy = func_numpy
count = 1 + (numpy_len - 1) // vals_numpy.size
for _ in range(1, count):
    vals_numpy = numpy.append(vals_numpy, func_numpy)

if dtype is not None:
    vals_numpy = vals_numpy.astype(dtype, copy=False)
vals_numpy = numpy.stack(vals_numpy).ravel()
```

Note: This change was only for testing the distribution of RandomState as for Generator `p` is required to be cupy.ndarray or a scalar.

povinsahu1909:
Following is the performance comparision result measured on GTX1050
size(rows * columns) | scipy (sec) | cupy (sec)
--- | --- | ---
10000 * 1 | 0.001 | 0.000
12000 * 1 | 0.001 | 0.000
15000 * 1 | 0.001 | 0.000
20000 * 1 | 0.002 | 0.000
10000 * 1000 | 0.826 | 0.001
12000 * 1000 | 0.996 | 0.001
15000 * 1000 | 1.299 | 0.001
20000 * 1000 | 1.752 | 0.001
10000 * 2000 | 1.668 | 0.001
12000 * 2000 | 2.001 | 0.001
15000 * 2000 | 2.695 | 0.002
20000 * 2000 | 3.445 | 0.001
10000 * 3000 | 2.596 | 0.001
12000 * 3000 | 3.127 | 0.001
15000 * 3000 | 3.968 | 0.002
20000 * 3000 | 5.273 | 0.001

<details>
<summary>script</summary>


import numpy as np
import cupy as cp
import cupyx

n_repeat = 2
n_warmup = 1
crng = cp.random.default_rng()
nrng = np.random.default_rng()

def get_time(perf):

def run(p, q):
    times = []
    perf = cupyx.time.repeat(nrng.geometric,
                             (p,),
                             n_repeat=n_repeat, n_warmup=n_warmup)
    times.append(get_time(perf))
    perf = cupyx.time.repeat(crng.geometric,
                             (q,),
                             n_repeat=n_repeat, n_warmup=n_warmup)
    times.append(get_time(perf))
    return times

print('size(nrows * ncols) | scipy (sec) | cupyx (sec)')
print('---|---|---')
nrows = [10000, 12000, 15000, 20000]
ncols = [1, 1000, 2000, 3000]
for m in ncols:
    for n in nrows:
        p = np.random.random((n, m))
        q = cp.asarray(p)
        times = run(p, q)
        print('{} x {} | {:.3f} | {:.3f}'.format(n, m, times[0], times[1]))

</details>

cupy/random/_generator_api.pyx
```diff
@@ -262,6 +265,55 @@ class Generator:
...
+                p_a = ndarray((), numpy.float64)
+                p_a.fill(p)
+                p = p_a
+            else:
+                raise TypeError('p is required to be a cupy.ndarray'
+                                ' or a scalar')
+        else:
+            # Check if size is broadcastable to shape
+            # but size determines the output
+            p = p.astype('d', copy=False)
```
povinsahu1909, line 303: Can we remove this test, as due to this test input array is required to be cupy.ndarray. So something like `rng.geometric([0.2, 0.3, 0.4])` will not work, but was supported by numpy.

takagi:
pfnCI, test this please.

takagi:
Seems that cupy generates wrong numbers with larger size more than 1d. I get a similar result with another distribution, i.g. standard_normal, so it's supposed to be a general problem of the generator class.

```Python
import numpy as np
import cupy as cp
n, m = 1000, 1000
p = cp.asarray(np.random.random((n, m)))
rng = cp.random.default_rng()
print(rng.geometric(p))
```

```
[[ 1  3  3 ...  1  2  1]
 [51  1  2 ...  5  1  1]
 [ 1  1  2 ...  1  1  4]
 ...
 [ 0  0  0 ...  0  0  0]
 [ 0  0  0 ...  0  0  0]
 [ 0  0  0 ...  0  0  0]]
```

takagi:
Would you try the performance comparison again with flattened shapes `m*n` with the same conditions?

cupy/random/_generator_api.pyx
takagi: Making p take array-like other than cupy.ndarray is okay.

emcastillo: we only allow cupy arrays or scalars in order to avoid implicit synchronization.

povinsahu1909:
Performance comparison for flattened shapes
size | numpy | cupy
--- | --- | ---
10000 | 0.001 | 0.000
12000 | 0.001 | 0.000
15000 | 0.001 | 0.000
20000 | 0.002 | 0.000
10000000 | 0.803 | 0.025
12000000 | 0.981 | 0.030
15000000 | 1.205 | 0.038
20000000 | 1.603 | 0.051
24000000 | 1.918 | 0.076
30000000 | 2.411 | 0.080
36000000 | 2.941 | 0.092
40000000 | 3.310 | 0.105
45000000 | 3.678 | 0.135
60000000 | 4.931 | 0.153

<details>
<summary>script</summary>
import numpy as np
import cupy as cp
import cupyx

n_repeat = 2
n_warmup = 1
crng = cp.random.default_rng()
nrng = np.random.default_rng()

def get_time(perf):

def run(p, q):
    times = []
    perf = cupyx.time.repeat(nrng.geometric, (p,), n_repeat=n_repeat, n_warmup=n_warmup)
    times.append(get_time(perf))
    perf = cupyx.time.repeat(crng.geometric, (q,), n_repeat=n_repeat, n_warmup=n_warmup)
    times.append(get_time(perf))
    return times

print('size(nrows * ncols) | scipy (sec) | cupyx (sec)')
print('---|---|---')
nrows = [10000, 12000, 15000, 20000]
ncols = [1, 1000, 2000, 3000]
for m in ncols:
    for n in nrows:
        p = np.random.random((n*m, ))
        q = cp.asarray(p)
        times = run(p, q)
        print('{} | {:.3f} | {:.3f}'.format(n*m, times[0], times[1]))

</details>

takagi:
> Performance comparison for flattened shapes

Thanks! This result looks reasonable to me.

tests/cupy_tests/random_tests/common_distributions.py
takagi: I don't think `check_ks` for `p` of ndarray makes much sense, so we can skip it.

takagi:
pfnCI, test this please.

takagi (approved):
LGTM! Waiting for CI pass.

takagi:
Thanks, now it is merged!

PR #5444, opened 2021-07-05
Use `dtype.name` instead `dtype.char`

asi1024:
Fix to use `dtype.name` instead `dtype.char` and remove workaround from Windows behavior difference.
https://github.com/cupy/cupy/pull/5387#issuecomment-872319018


asi1024:
Jenkins, test this please.

emcastillo (approved):

PR #5449, opened 2021-07-06
Fix new DLPack protocol error messages and tests

kmaehashi:
Follow-up: #5306

* Formatted string were broken :smile:
* Fix wrong tests. I noticed that stream consistency cannot be tested unless directly calling `__dlpack__` because the current stream in consumer (`cupy.from_dlpack`) and producer (`cupy.ndarray__dlpack__`) are always the same if both ends are CuPy.
  Test failure detected in https://github.com/chainer/chainer-test/pull/673#issuecomment-874877502

### DIFF
--- a/tests/cupy_tests/core_tests/test_dlpack.py
+++ b/tests/cupy_tests/core_tests/test_dlpack.py
@@ -48,19 +48,28 @@ class TestNewDLPackConversion(unittest.TestCase):
 
     @testing.for_all_dtypes(no_bool=True)
     def test_conversion(self, dtype):
+        orig_array = _gen_array(dtype)
+        out_array = cupy.from_dlpack(orig_array)
+        testing.assert_array_equal(orig_array, out_array)
+        testing.assert_array_equal(
+            orig_array.data.ptr, out_array.data.ptr)
+
+    def test_stream(self):
         allowed_streams = ['null', True]
         if not cuda.runtime.is_hip:
             allowed_streams.append('ptds')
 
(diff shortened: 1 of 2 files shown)


kmaehashi:
pfnCI, test this please.

leofang (approved):
Good catch! Thanks @kmaehashi 🙏

tests/cupy_tests/core_tests/test_dlpack.py
```diff
@@ -48,19 +48,28 @@ def _get_stream(self, stream_name):
...
         if not cuda.runtime.is_hip:
             allowed_streams.append('ptds')
 
         # stream order is automatically established via DLPack protocol
-        for src in allowed_streams:
-            src_s = self._get_stream(src)
-            for dst in allowed_streams:
-                dst_s = self._get_stream(dst)
+        for src_s in [self._get_stream(s) for s in allowed_streams]:
+            for dst_s in [self._get_stream(s) for s in allowed_streams]:
```
toslunar, line 64: BTW, how about using `parametrize`
```
@pytest.mark.parametrize("src", ['null', 'ptds', True])
@pytest.mark.parametrize("dst", ['null', 'ptds', True])
```
and `pytest.skip` in `_get_stream`?

kmaehashi: Looks nice, but seems this doesn't work with unittest-style methods... 😞

toslunar (approved):

PR #5458, opened 2021-07-07
Make `matrix_power` support stacked matrices 

### LINKED ISSUE #5451: Documentation for `cupy.linalg.matrix_power()` has incorrect warning
In master, the documentation for `cupy.linalg.matrix_power()` states that "M must be of dtype float32 or float64.". However, the function is only implemented using `cupy.linalg.matmul()` which does support integer matrices, i.e. the following works fine:

```python
cupy.linalg.matrix_power(cupy.arange(1, 26).reshape(5, 5), 5)
```

It is possible that this was originally intended to be implemented using an eigenvalue decomposition, however, eigenvalue decomposition isn't well supported with CUDA unless the matrix is symmetric, and thus it is *not* implemented using that setup. Additionally, in the future if the eigenvalue decomposition method is implemented, the current method could still be used for integer matrices.

toslunar:
and fix #5451.

### DIFF
--- a/cupy/linalg/_product.py
+++ b/cupy/linalg/_product.py
@@ -360,8 +360,6 @@ def matrix_power(M, n):
     Returns:
         ~cupy.ndarray: Output array.
 
-    .. note:: M must be of dtype `float32` or `float64`.
-
     ..seealso:: :func:`numpy.linalg.matrix_power`
     """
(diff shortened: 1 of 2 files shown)


emcastillo (approved):
LGTM!

emcastillo:
Jenkins, test this please

PR #5334, opened 2021-06-13
Support Numba-like `jit.grid()` syntax in CuPy JIT

leofang:
This PR adds a convenience helper to make it easier to port existing Numba codes to CuPy JIT. The purpose of `jit.grid()` is to compute the absolute index of a thread in the grid. 

Note that `grid(n)` returns a tuple for n=2, 3. Calling `thrust::make_tuple()` would indeed incur a bit of overhead for such a simple operation, but to retain the Numba convention I am not sure if there's a better solution. Suggestions are more than welcome, as always!

The below snippet shows the generated CUDA code:
```python
import cupy as cp
from cupyx import jit

@jit.rawkernel()
def grid_1d():
    x = jit.grid(1)

grid_1d[1, 1]()
print(grid_1d.cached_code)


@jit.rawkernel()
def grid_2d():
    x, y = jit.grid(2)

grid_2d[1, 1]()
print(grid_2d.cached_code)


@jit.rawkernel()
def grid_3d():
    x, y, z = jit.grid(3)

grid_3d[1, 1]()
print(grid_3d.cached_code)
```
Output: 
```
extern "C" __global__ void grid_1d() {
  unsigned int x;
  x = threadIdx.x + blockIdx.x * blockDim.x;
}
extern "C" __global__ void grid_2d() {
  unsigned int x;
  unsigned int y;
  thrust::tie(x, y) = thrust::make_tuple(threadIdx.x + blockIdx.x * blockDim.x, threadIdx.y + blockIdx.y * blockDim.y);
}
extern "C" __global__ void grid_3d() {
  unsigned int x;
  unsigned int y;
  unsigned int z;
  thrust::tie(x, y, z) = thrust::make_tuple(threadIdx.x + blockIdx.x * blockDim.x, threadIdx.y + blockIdx.y * blockDim.y, threadIdx.z + blockIdx.z * blockDim.z);
}
```

TODO:
- [x] Add error messages
- [x] Add tests
- [x] Add docs

PR #5461, opened 2021-07-08
Support Numba-like `jit.gridsize()` syntax in CuPy JIT

leofang:
Follow-up of #5334.


asi1024:
Jenkins, test this please.

asi1024 (approved):

PR #5462, opened 2021-07-08
Support Numba-like `jit.laneid()` and `jit.warpsize` syntax in CuPy JIT

leofang:
Follow-up of #5334 and #5461. ~~To be rebased after #5461 is merged.~~ 

Note that unlike in Numba, `laneid()` has to be a callable. The reason is that IIUC only the `BuiltinFunc` subclasses have access to `env`, which would allow us to insert a preamble as needed. I'd be happy to learn if there's any alternative approach...


leofang:
Merged with master!

cupyx/jit/_interface.py
```diff
@@ -180,3 +180,16 @@ def __init__(self, name):
 blockDim = _Dim3('blockDim')
 blockIdx = _Dim3('blockIdx')
 gridDim = _Dim3('gridDim')
+
+warpsize = _internal_types.Data(
+    '64' if runtime.is_hip else '32', _cuda_types.uint32)
```
asi1024, line 185: `warpsize` can be defined as a constant value.
```suggestion
warpsize = numpy.int32(64'if runtime.is_hip else 32)
```

leofang: @asi1024 Interesting, thanks. But, I found that in this case we cannot overwrite `warpsize`'s `__doc__` attribute:
```python
>>> a = np.uint32(10)
>>> a.__doc__ = "hahaha"
AttributeError: 'numpy.uint32' object attribute '__doc__' is read-only
```
Moreover, this doesn't work with RTD either (tried in fbf602b):
```python
warpsize = ...
"""my docstring"""
```
Am I missing something?

asi1024: I got the point. Thanks!

asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024 (approved):

asi1024:
LGTM. Thanks!

leofang:
Thank you, @asi1024!

PR #5463, opened 2021-07-08
Assert function inside `@testing.for_*` returns None

toslunar:
For example, the following test is wrong because it just checks `TypeError` is raised for some dtype.
```python
# wrong order of decorators
@testing.numpy_cupy_allclose(..., accept_error=TypeError)
@testing.for_dtypes(...)
def test_neg(self, xp, dtype):
    return -xp.zeros((), dtype)
```


kmaehashi:

PR #5474, opened 2021-07-09
Introduce new toolset for CI

kmaehashi:
### Abstract

The new CI toolset intends to replace `chainer-test`. The new toolset intends to:

* Maximize the environment coverage
    * Currently environment configurations (versions of OS/CUDA/Python/NumPy/...) in `chainer-test` are chosen in ad-hoc manner and difficult to understand what is covered and what is not.
      The new tool generates a coverage [matrix in Markdown](https://github.com/kmaehashi/cupy/blob/add-ci-generator/.pfnci/coverage.md) so that we can confirm that all supported environments are well tested.
* Make matrix maintenance easy
    * Configuration is now defined in YAML, making maintenance much easier.
    * By utilizing binary packages (apt/yum) we no longer have to use the redist tarball URLs when adding new versions of optional libraries (cuTENSOR, cuDNN, etc.)
    * The tool checks the consistency between libraries and warns if an unsupported combination is given (e.g. NCCL 2.9 does not support CUDA 11.2) so it's easier and safer to modify the matrix.
* Make it easy to reproduce the CI environment
    * Everyone can reproduce the test environment and run unit tests by `./.pfnci/linux/run.sh cuda114 build test` (requires Docker).
      Maintainers can even pull the docker images from the GCR (docker registry), so no need to build Docker images: just `./.pfnci/linux/run.sh cuda114 test`.
    * Docker image build is much faster than `chainer-test` as the tool uses `nvidia/cuda` as base images.
    * Unlike `chainer-test`, the new tool can run unit tests against the current source tree, making try-and-error loop faster.
* Run on both FlexCI and Jenkins
    * Eventually we will migrate from Jenkins to FlexCI, except for ROCm tests as AMD GPUs are not yet supported by FlexCI which is backed by GCP.

### Migration of CUDA Test from Jenkins to FlexCI

There are several tests running on Jenkins:

1. Unit tests (master/stable + pull-request)
2. Doctests (master/stable + pull-request)
3. Example tests (master/stable + pull-request)
4. Slow tests (master/stable)
5. Head tests (master/stable)
6. Installation tests (master/stable)
7. Shuffle tests (master/stable)

The tool in this PR currently covers 1-5.

### Remaining tasks to be done in separate PR

* Add configuration for each FlexCI project defined in `matrix.yml`.
* Support Windows.
* Support CUDA Python axis.
* Support installation tests.
* Support shuffle tests.


kmaehashi:
I think this is ready for review.

Most of files added are auto-generated Dockerfiles+scripts (`.pfnci/linux/tests`) and `coverage.md` that illustrates the current test environment coverage. `generate.py` generates them from `schema.yml` and `matrix.yml`.

.pfnci/generate.py
```diff
@@ -0,0 +1,470 @@
...
+            raise AssertionError
+
+        lines += [
+            f'ARG BASE_IMAGE="{base_image}"',
+            'FROM ${BASE_IMAGE}',
+            '',
+        ]
+
+        # Install tools and additional libraries.
+        if os_name == 'ubuntu':
```
emcastillo, line 52: How about using inheritance to specify different OSs?

emcastillo: Mmm maybe it is much easier to read this way, let's forget about the above.

.pfnci/schema.yaml
```diff
@@ -0,0 +1,342 @@
...
+cython:
+  "0.29":
+    spec: "==0.29.*"
+  "pre":
+    spec: ">=0a0"  # latest, including pre-releases
+
+###############################################################################
+# Accelerators
+###############################################################################
+# Corresponds to the environment variable "CUPY_ACCELERATORS".
```
emcastillo, line 327: How about directly supporting other environment variables instead?, I think the needed change is negligible and the benefits are great.

.pfnci/schema.yaml
```diff
@@ -0,0 +1,342 @@
...
+###############################################################################
+# The optional libraries to install: NCCL, cuTENSOR, cuSPARSELt and cuDNN.
+# `spec` must be a version specifier accepted by apt and yum.
+# `cuda` lists the supported CUDA version.
+# `cuda.alias` (cuDNN only) is an alias name for package name suffix.
+nccl:
+  # https://docs.nvidia.com/deeplearning/nccl/archives/
+  # https://developer.nvidia.com/nccl/nccl-download
+  null:
+    spec: null
```
emcastillo, line 78: Is this needed? seems that all libraries define it by default, but I didn't see any explicit check against it in the code.

emcastillo (commented):
Some minor comments to be discussed, in general it looks great!!

.pfnci/schema.yaml
kmaehashi: The current design is to define a list of **all possible values** in schema so that the coverage can be illustrated in [2-d matrix](https://github.com/kmaehashi/cupy/blob/add-ci-generator/.pfnci/coverage.md).
But I understand your point that adding a new env var test axis should be easy.
How about automatically handling all `env:*` in schema keys as environment variable axis? e.g.,

```yaml
env:CUPY_ACCELERATORS:
  null:
  "cub":
  "cutensor":
  "cub,cutensor":
  "cutensor,cub":
env:CUPY_USE_CUDA_PYTHON:
  null:
  "1":
```

.pfnci/schema.yaml
kmaehashi: Good point, `spec: null` is not needed actually. Will remove.

.pfnci/schema.yaml
emcastillo: this looks cool, thanks!

.pfnci/schema.yaml
kmaehashi: I tried to remove and recalled that this is as expected 😅 
In YAML, 

```yaml
key:
  subkey: 1
key2:
```

is equivalent to

```py
{'key1': {'subkey': 1}, 'key2': None}
```

so I need to add many null(None)-checking to `generate.py` when iterating the dict above. Defining subkeys as a dict:

```yaml
key:
  subkey: 1
key2:
  subkey: null
```

makes a life lot easier, as it will be loaded as:

```py
{'key1': {'subkey': 1}, 'key2': {'subkey': None}}
```

emcastillo:
Jenkins, test this please

emcastillo (dismissed):
LGTM!

.pfnci/schema.yaml
emcastillo: understood! thanks for the explanation!

kmaehashi:
Let me disable auto-merge 😄 
Before merging I want to confirm that:
* ROCm tests run fine on Jenkins (need to check log, as currently Jenkins says "SUCCESS" even if ROCm tests are failing)
* CUDA tests with new `cuda114` docker image runs fine on FlexCI (need to trigger a test after docker image rebuild: https://ci.preferred.jp/cupy.prep-linux/79140/)

kmaehashi:
pfnCI, test this please.

kmaehashi:
The above commits fixes `rocm-4-3` test failure in Jenkins.
pfnCI, test this please.

kmaehashi:
Avoid hitting FlexCI memory limit when doing parallel docker image build. Rebuild ongoing https://ci.preferred.jp/cupy.prep-linux/79180/

kmaehashi:
Rebuild succeeded!

/test

kmaehashi:
pfnCI, run Jenkins please.

kmaehashi:
Blocked by #5757

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, run Jenkins please.

emcastillo:
Needs rebase :)

kmaehashi:
Will trigger FlexCI after docker image rebuild:  https://ci.preferred.jp/cupy.prep-linux/80247/

emcastillo (approved):
LGTM!!! Thanks :D

kmaehashi:
I think it's ok to keep v9 as-is, as we will cut v10 branch this month.

PR #5479, opened 2021-07-12
Fix random `integers`

emcastillo:
This was broken by returning invalid values due to use the wrong type in the output array

The statistical properties of the distribution were not currently tested, now they are

### DIFF
--- a/tests/cupy_tests/random_tests/test_generator_api.py
+++ b/tests/cupy_tests/random_tests/test_generator_api.py
@@ -194,35 +194,51 @@ class TestStandardNormalInvalid(InvalidOutsMixin, GeneratorTestCase):
 @testing.gpu
 @testing.fix_random()
 class TestIntegers(GeneratorTestCase):
-    # TODO(niboshi):
-    #   Test soundness of distribution.
-    #   Currently only reprocibility is checked.
-
     target_method = 'integers'
 
-    def test_randint_1(self):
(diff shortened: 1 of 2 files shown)


tests/cupy_tests/random_tests/test_generator_api.py
```diff
@@ -194,36 +194,52 @@ def test_invalid_dtypes(self):
...
 
+    @_condition.repeat_with_success_at_least(10, 3)
+    def test_integers_ks(self):
+        self.check_ks(0.05)(
+            low=100, high=1000, size=2000)
+
+    @_condition.repeat_with_success_at_least(10, 3)
+    def test_integers_ks_low(self):
+        self.check_ks(0.05)(
+            low=100, size=2000)
```
toslunar, line 231: For discrete distributions, K-S test might be unusual and `check_ks` returns inaccurate p-value.  However, adding a distribution test is a great progress and I hope the bin size (100) is sufficiently large.

toslunar (approved):

toslunar:
Jenkins, test this please.

tests/cupy_tests/random_tests/test_generator_api.py
emcastillo: yes, I actually thought of converting the distribution results to the`[0, 1)` interval for the test but I wasn't sure of it.

PR #5488, opened 2021-07-13
Use `from_dlpack` instead of fromDlpack

kmaehashi:
Follow up #5306

PR #5509, opened 2021-07-14
Add a lingering doc page for `fromDlpack()`

leofang:
Follow-up of #5488. I noticed that both the docstrings of `cp.from_dlpack()` and `cp.ndarray.toDlpack()` have a mention of the old `cp.fromDlpack()`, so perhaps before its full removal we should still keep its docstring rendered (but not indexed)?


leofang:
The rendered page is here: 
https://cupy--5509.org.readthedocs.build/en/5509/reference/generated/cupy.fromDlpack.html

kmaehashi:
How about naming the file `_deprecated.rst` instead of `_dlpack.rst` so that we can use it in similar cases in future and also make the role of the file explicit?

leofang:
Sounds good, done in d267241.

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

kmaehashi:
Merging as docs build are fine. Thanks!

Issue #5511, opened 2021-07-15
Memory type information is not available when `MemoryPool` is in use


leofang:
Consider the following snippet that uses `MemoryPool`:
```python
>>> import cupy
>>>
>>> cupy.cuda.set_allocator(cupy.cuda.MemoryPool(cupy.cuda.malloc_managed).malloc)
>>> a = cp.arange(10)
>>> a.data
<MemoryPointer 0x7f26c0000000 device=0 mem=<cupy.cuda.memory.PooledMemory object at 0x7f28da594930>>
```
The actual memory type (`ManagedMemory`) is absorbed into `PooledMemory`. But ideally it should be available somewhere for query. If `MemoryPool` is not used, it is exposed as expected:
```python
>>> cupy.cuda.set_allocator(cupy.cuda.malloc_managed)
>>> a = cp.arange(10)
>>> a.data
<MemoryPointer 0x7f26c0000200 device=0 mem=<cupy.cuda.memory.ManagedMemory object at 0x7f28da5a72a0>>
``` 

The accessibility to the type information can be useful, for example in DLPack we can set `kDLCUDAManaged` accordingly. Perhaps we could add a new attribute to `BaseMemory` and record the type name?

emcastillo:
Maybe you can use?
```
cupy_backends/cuda/api/runtime.pyx:    int cudaPointerGetAttributes(_PointerAttributes* attributes,
```

leofang:
Yeah I have to pay a look-up price anyway. I forgot about pointer attributes -- thanks Emilio!

emcastillo:
@kmaehashi suggested it :)

PR #5512, opened 2021-07-15
Update DLPack header to v0.6

leofang:


leofang:
I wanted to support detecting managed memory but it's currently blocked by #5511.

leofang:
To rebase after #5544.

leofang:
Note: This PR could break user code (see https://github.com/dmlc/dlpack/issues/74#issuecomment-888096863), so we probably should not backport it.

leofang:
Jenkins, test this please

leofang:
> Note: This PR could break user code (see [dmlc/dlpack#74 (comment)](https://github.com/dmlc/dlpack/issues/74#issuecomment-888096863)), so we probably should not backport it.

Perhaps we should add an env var, say `CUPY_DLPACK_ENABLE_MANAGED_MEMORY`, to control if managed memory should be treated differently from normal device memory?

emcastillo:
Jenkins, test this please

kmaehashi:
Maybe adding `CUPY_DLPACK_EXPORT_VERSION`, similar to `CUPY_CUDA_ARRAY_INTERFACE_EXPORT_VERSION`?

emcastillo:
Hi @leofang 
Can you add the
> Maybe adding CUPY_DLPACK_EXPORT_VERSION, similar to CUPY_CUDA_ARRAY_INTERFACE_EXPORT_VERSION?

as @kmaehashi suggested? :)
Thanks!

leofang:
Sorry I dropped the ball. Will do!

cupy/_util.pyx
```diff
@@ -26,6 +26,10 @@ CUDA_ARRAY_INTERFACE_EXPORT_VERSION = int(
     os.environ.get('CUPY_CUDA_ARRAY_INTERFACE_EXPORT_VERSION', 3))
 
 
+CUPY_DLPACK_EXPORT_VERSION = float(
```
kmaehashi, line 29: How about using `60` here, like `DLPACK_VERSION`? Using float can be problematic when comparison.

leofang: `DLPACK_VERSION` is very bad. Before v0.6, it is `050` for v0.5 and so on. The comparison would still work if users type it correctly, but I think they should not have to worry about how exactly `DLPACK_VERSION` was formatted.

leofang: (See the removed stringify code in `cupy/_core/dlpack.pyx`.)

emcastillo: I think @kmaehashi mostly meant to use `CUPY_DLPACK_EXPORT_VERSION=60` or `CUPY_DLPACK_EXPORT_VERSION=50`, just to avoid the float conversion, we decide the version values instead of using `DLPACK_VERSION` directly.

I agree with him, we should avoid `float` unless there is a strong and very good reason for it.

leofang: I mean why would users know `60` or `50` are acceptable inputs, instead of say `050`? And how about other values? I think the version string is implementation detail and users, if interpreting this env var literally, would want to set it according to [the DLPack public releases](https://github.com/dmlc/dlpack/releases), no?

PR #5513, opened 2021-07-15
Fix mergify condition

takagi:
With the change of #5439, mergify complains its configuration is invalid:

> Invalid condition 'closed=false'. Expected end of text, found '=' (at char 6), (line:1, col:7) @ pull_request_rules → item 1 → conditions → item 1

~~I can read that we have to use '-' prefix instead of '=' operator for negating a boolean condition:~~ 🤔 
https://docs.mergify.io/conditions/#grammar

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,7 +12,7 @@ pull_request_rules:
   - name: Notify conflict
     conditions:
       - conflict
-      - closed=false
+      - -closed
     actions:
       comment:
         message: This pull request is now in conflicts. Could you fix it @{{author}}? 🙏


takagi:
mergify still says the same error...

.mergify.yml
```diff
@@ -12,7 +12,7 @@ pull_request_rules:
   - name: Notify conflict
     conditions:
       - conflict
-      - closed=false
+      - -closed
```
kmaehashi, line 15: Thanks for the catch! Maybe this works?
```suggestion
    conditions:
      - and:
        - conflict
        - closed=false

takagi: oh... let me try it!

takagi:
mergify validator says `-closed` is valid and `closed=false` is invalid.
https://docs.mergify.io/configuration/#validation

Seems that `.mergify.yml` in the default branch is used.
https://docs.mergify.io/configuration/#file-used

takagi:
I've rolled back this change to `86acd63`. Would you merge this into the master once to see if it works?

PR #5538, opened 2021-07-19
Add auto notify bot for `hip` label

kmaehashi:
Automatically leave `cc/ someone` comment when specific label is set.
If we later need to support multiple labels, we can use `fromJSON` to define label-to-users mappings in JSON.

`peter-evans/*` is a third-party action, but I think it's quite stable as it's documented on GitHub's official website.
https://docs.github.com/en/actions/guides/commenting-on-an-issue-when-a-label-is-added

### DIFF
--- a/.github/workflows/auto-cc.yml
+++ b/.github/workflows/auto-cc.yml
@@ -1,0 +1,32 @@
+name: "Auto Notify"
+
+on:
+  issues:
+    types: [labeled]
+  pull_request:
+    types: [labeled]
+
+jobs:
+  auto-notify:
+    if: |
+        github.repository_owner == 'cupy' &&
+        github.event.label.name == 'hip'
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Find comment
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+            issue-number: ${{github.event.number}}
+            body-includes: "<!-- Added by CuPy Auto Notify Bot -->"
+
+      - name: Create comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+            issue-number: ${{github.event.number}}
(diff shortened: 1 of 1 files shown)


kmaehashi:
I've tested it here https://github.com/kmaehashi/cupy/pull/61 and seems working fine.

emcastillo (approved):

kmaehashi:
🤔 
```
Error: Resource not accessible by integration
Error: See this action's readme for details about this error
```

kmaehashi:
Now it works :tada:
(ignore the backport error comment by bot)

.github/workflows/auto-cc.yml
```diff
@@ -0,0 +1,32 @@
...
+        github.event.label.name == 'hip'
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Find comment
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+            issue-number: ${{github.event.number}}
+            body-includes: "<!-- Added by CuPy Auto Notify Bot -->"
```
leofang, line 22: Can't the bot create a comment directly? Looks like it first creates an empty comment (this step) and then modify it (the next step). Just curious 🙂

leofang (commented):
This is cool! Just left a quick question.

.github/workflows/auto-cc.yml
kmaehashi: This step tries to find a comment including `"<!-- Added by CuPy Auto Notify Bot -->"`, and if that comment is found, the comment will be updated in the next step. This is to avoid leaving multiple comments when doing label -> unlabel -> label.

leofang: I see, good to learn about the trick!

Issue #4493, opened 2020-12-24
CUDA_PATH v.s. ROCM_HOME


kmaehashi:
The naming convention is conflicting. Maybe better to use ROCM_PATH instead? It's used in ROCm docs (but only once)
https://rocmdocs.amd.com/en/latest/Deep_learning/Deep-learning.html?highlight=ROCM_PATH#install-rocm

leofang:
Yeah the ROCm doc is as confusing as always...To add one more confusing data point, PyTorch uses both `ROCM_HOME` and `ROCM_PATH` 😂
https://github.com/pytorch/pytorch/search?q=ROCM_HOME
https://github.com/pytorch/pytorch/search?q=ROCM_PATH

leofang:
I think we should also accept `ROCM_PATH`. Try `grep -R /opt/rocm -e 'ROCM_PATH'`.

leofang:
Looks like `hipconfig` expects `ROCM_PATH`:
```
$ hipconfig -h
usage: hipconfig [OPTIONS]
  --path,  -p        : print HIP_PATH (use env var if set, else determine from hipconfig path)
  --rocmpath,  -R    : print ROCM_PATH (use env var if set, else determine from hip path or /opt/rocm)
  ...
```

amathews-amd:
Recommend using ROCM_PATH. I only see 3 projects using ROCM_HOME, one of which is CuPy.

kmaehashi:
https://github.com/cupy/cupy/pull/9351#issuecomment-3287525167

PR #5546, opened 2021-07-20
Fix auto notify bot for issues

kmaehashi:
The bot was not working for issues...


kmaehashi:
Confirmed that the bot now works on issues.
https://github.com/cupy/cupy/issues/4493

PR #5551, opened 2021-07-21
consistent kernel names

grlee77:
CuPy tends to follow the convention of naming kernels starting with a `cupy_` (or `cupyx_`) prefix. This PR is just to make this consistent across the library. There should be no change in behavior!

There were a few instances without any name specified, but mostly these are just adding or modify the prefix for an existing name.

For the `cupyx.scipy` modules, I used more descriptive `cupyx_scipy_ndimage_`, `cupyx_scipy_sparse_`, etc. as the prefixes

### DIFF
--- a/cupyx/scipy/ndimage/_spline_prefilter_core.py
+++ b/cupyx/scipy/ndimage/_spline_prefilter_core.py
@@ -209,7 +209,7 @@ __device__ T* row(
 _batch_spline1d_strided_template = """
 extern "C" __global__
 __launch_bounds__({block_size})
-void cupyx_spline_filter(T* __restrict__ y, const idx_t* __restrict__ info) {{
+void {kernel_name}(T* __restrict__ y, const idx_t* __restrict__ info) {{
     const idx_t n_signals = info[0], n_samples = info[1],
         * __restrict__ shape = info+2;
     idx_t y_elem_stride = 1;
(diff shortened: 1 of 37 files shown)


leofang:
Jenkins, test this please

cupyx/scipy/ndimage/_spline_prefilter_core.py
```diff
@@ -250,4 +250,6 @@ def get_raw_spline1d_kernel(axis, ndim, mode, order, index_type='int',
     # generate code handling batch operation of the 1d filter
     code += _batch_spline1d_strided_template.format(ndim=ndim, axis=axis,
                                                     block_size=block_size)
-    return cupy.RawKernel(code, 'cupyx_spline_filter')
+    kernel_name = (f'cupyx_scipy_ndimage_spline_filter_{ndim}d_ord{order}_'
+                   f'axis{axis}_{mode}')
+    return cupy.RawKernel(code, kernel_name)
```
leofang, line 255: This is the most immediate error I can spot: line 212 should also be changed in order to fetch the correct kernel name:
https://github.com/cupy/cupy/blob/840e112e1dd5757eff074f1dc65dc6d950f34da5/cupyx/scipy/ndimage/_spline_prefilter_core.py#L212

grlee77: Yes, I have fixed this locally now, but other tests are still running. It was also necessary to `replace('-', '_')` in the mode name string to get a valid C++ function name for the kernel

leofang:
Jenkins, test this please

emcastillo (approved):
Thanks!!

kmaehashi:
Backported: #5625

PR #5556, opened 2021-07-24
Disable Mergify's auto-merge

kmaehashi:
We started using GitHub's native auto-merge, so let's turn off the Mergify.
Native auto-merge works well even with GitHub Actions (due to security restrictions Mergify can't do auto-merge if GitHub workflow files are updated in base branch after submitting the PR).

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,14 +1,4 @@
 pull_request_rules:
-  - name: Auto merge
-    conditions:
-      - "#changes-requested-reviews-by=0"
-      - label=st:test-and-merge
-      - status-success=Jenkins
-      - status-success=pretest  # github workflows
-    actions:
-      merge:
-        method: merge
-        strict: false  # strict mode and manually-triggered CIs do not get along
   - name: Notify conflict
     conditions:
       - conflict


.mergify.yml
```diff
@@ -1,14 +1,4 @@
 pull_request_rules:
-  - name: Auto merge
-    conditions:
-      - "#changes-requested-reviews-by=0"
```
toslunar, line 4: Instead of "request changes", [disable auto-merge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request#disabling-auto-merge) will be used (by anyone with write permission) in an emergency.

.mergify.yml
```diff
@@ -1,14 +1,4 @@
...
-      - label=st:test-and-merge
```
toslunar, line 5: This label will no longer have special meaning.

.mergify.yml
```diff
@@ -1,14 +1,4 @@
...
-      - status-success=Jenkins
-      - status-success=pretest  # github workflows
```
toslunar, line 7: The requirements will be managed by Settings -> Branches -> Branch protection rules (master or v9) -> Status checks that are required.

toslunar (approved):

PR #5561, opened 2021-07-26
Fix CUDA API call on module initialization

emcastillo:
This should reduce memory consumption when importing cupy

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1514,9 +1514,6 @@ cdef class MemoryPool:
         return mp.get_limit()
 
 
-cdef bint MemoryAsyncHasStat = (runtime.driverGetVersion() >= 11030)
-
-
 cdef class MemoryAsyncPool:
     """(Experimental) CUDA memory pool for all GPU devices on the host.
 
@@ -1573,6 +1570,9 @@ cdef class MemoryAsyncPool:
     cdef:
         # A list of cudaMemPool_t to each device's mempool
         readonly list _pools
+        bint _memoryAsyncHasStat
+
+
 
     def __init__(self, pool_handles='current'):
         _util.experimental('cupy.cuda.MemoryAsyncPool')
@@ -1580,7 +1580,7 @@ cdef class MemoryAsyncPool:
         cdef dict limit = _parse_limit_string()
         dev_counts = runtime.getDeviceCount()
         self._pools = []
-
+        self._memoryAsyncHasStat = (runtime.driverGetVersion() >= 11030)
         if (cpython.PySequence_Check(pool_handles)
                 and not isinstance(pool_handles, str)):
             # allow different kinds of handles on each device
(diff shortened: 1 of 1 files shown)


cupy/cuda/memory.pyx
```diff
@@ -1573,14 +1570,17 @@ cdef class MemoryAsyncPool:
     cdef:
         # A list of cudaMemPool_t to each device's mempool
         readonly list _pools
+        bint _memoryAsyncHasStat
+
+
```
leofang, line 1575: Just two nits:
1. We can just name it as `memoryAsyncHasStat` since it's not exposed to Python; you could also mark it `readonly` if you like.
2. The two extra lines are not needed.

leofang (approved):
Thanks a lot for catching the problem, @emcastillo! I just left two nits but this works as expected.

Before this fix (on master):
```
$ nvprof python -c "import cupy"
==31552== NVPROF is profiling process 31552, command: python -c import cupy
==31552== Profiling application: python -c import cupy
==31552== Profiling result:
No kernels were profiled.
            Type  Time(%)      Time     Calls       Avg       Min       Max  Name
      API calls:   57.85%  424.24us         1  424.24us  424.24us  424.24us  cuDeviceTotalMem
                   29.58%  216.96us       101  2.1480us     146ns  95.608us  cuDeviceGetAttribute
                    7.02%  51.494us         1  51.494us  51.494us  51.494us  cuDeviceGetName
                    4.09%  29.981us         1  29.981us  29.981us  29.981us  cuDeviceGetPCIBusId
                    0.57%  4.1960us         1  4.1960us  4.1960us  4.1960us  cudaDriverGetVersion
                    0.36%  2.6540us        14     189ns     130ns     775ns  cudaCreateChannelDesc
                    0.33%  2.4170us         3     805ns     273ns  1.7680us  cuDeviceGetCount
                    0.15%  1.1170us         2     558ns     218ns     899ns  cuDeviceGet
                    0.04%     318ns         1     318ns     318ns     318ns  cuDeviceGetUuid
```
With this fix:
```
$ nvprof python -c "import cupy"
======== Warning: No profile data collected.
```
All CUDA API calls are gone at import time.

leofang:
Maybe we should add a test that runs `nvprof python -c "import cupy"` with a subprocess to monitor if any CUDA API calls are made?

cupy/cuda/memory.pyx
```diff
@@ -1514,9 +1514,6 @@ cdef class MemoryPool:
         return mp.get_limit()
 
 
-cdef bint MemoryAsyncHasStat = (runtime.driverGetVersion() >= 11030)
```
leofang, line 1517: For the record this was my bad. I thought it's safe to query the version without initializing the CUDA context, but apparently I didn't verify this assumption...🤦🏻‍♂️

emcastillo: no worries! :)

kmaehashi (approved):

kmaehashi:
pfnCI, test this please.

emcastillo:
v9 does not have this issue

PR #5574, opened 2021-07-28
Release the GIL for several NCCL ops

hauntsaninja:
(patch authored by Greg Brockman)

### DIFF
--- a/cupy_backends/cuda/libs/nccl.pyx
+++ b/cupy_backends/cuda/libs/nccl.pyx
@@ -26,22 +26,22 @@ cdef extern from '../../cupy_nccl.h':
     ctypedef enum ncclDataType_t:
         pass
 
-    const char* ncclGetErrorString(ncclResult_t result)
-    ncclResult_t ncclGetVersion(int* version)
+    const char* ncclGetErrorString(ncclResult_t result) nogil
+    ncclResult_t ncclGetVersion(int* version) nogil
     ncclResult_t ncclCommGetAsyncError(ncclComm_t comm,
                                        ncclResult_t *asyncError) nogil
-    ncclResult_t ncclGetUniqueId(ncclUniqueId* uniqueId)
+    ncclResult_t ncclGetUniqueId(ncclUniqueId* uniqueId) nogil
     ncclResult_t ncclCommInitRank(ncclComm_t* comm, int ndev,
-                                  ncclUniqueId commId, int rank)
+                                  ncclUniqueId commId, int rank) nogil
     ncclResult_t ncclCommInitAll(ncclComm_t* comm, int ndev,
                                  const int* devlist)
     ncclResult_t ncclGroupStart() nogil
     ncclResult_t ncclGroupEnd() nogil
-    void ncclCommDestroy(ncclComm_t comm)
-    void ncclCommAbort(ncclComm_t comm)
-    ncclResult_t ncclCommCuDevice(const ncclComm_t comm, int* device)
-    ncclResult_t ncclCommUserRank(const ncclComm_t comm, int* rank)
-    ncclResult_t ncclCommCount(const ncclComm_t comm, int* count)
+    void ncclCommDestroy(ncclComm_t comm) nogil
+    void ncclCommAbort(ncclComm_t comm) nogil
+    ncclResult_t ncclCommCuDevice(const ncclComm_t comm, int* device) nogil
+    ncclResult_t ncclCommUserRank(const ncclComm_t comm, int* rank) nogil
+    ncclResult_t ncclCommCount(const ncclComm_t comm, int* count) nogil
     ncclResult_t _ncclAllReduce(const void* sendbuff, void* recvbuff,
                                 size_t count,
                                 ncclDataType_t datatype, ncclRedOp_t op,
@@ -110,7 +110,9 @@ class NcclError(RuntimeError):
 
     def __init__(self, int status):
         self.status = status
-        cdef msg = ncclGetErrorString(<ncclResult_t>status)
+        cdef const char* msg
+        with nogil:
+            msg = ncclGetErrorString(<ncclResult_t>status)
         if NCCL_VERSION_CODE < 2000:
             s = ERROR1[status]
         else:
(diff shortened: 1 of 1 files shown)


kmaehashi:
Hi @hauntsaninja, thank you for the contribution!
Overall this PR looks good to go, but want to understand the context. Did you see performance improvements by this fix?

> (patch authored by Greg Brockman)

The credit of this pull-request goes to `@gdb`? (Just wanted to make sure 😃 )

hauntsaninja:
Yup, @gdb wrote the patch and I just made some small fixups.

We were looking to optimise NCCL comm creation time. As you'd expect, without the patch, it takes about the same time to create them serially vs in threads; with the patch you get a nice speed up. Also, when creating NCCL comms in threads without the patch, it's easy to deadlock.

Here's an example script I wrote up (to be run using MPI):
```
import os
import platform
import random
import threading
import time
import sys

from cupy.cuda import Stream, get_current_stream, nccl
from mpi4py import MPI


def get_comm_groups():
    groups = []
    mpi_rank = MPI.COMM_WORLD.Get_rank()
    mpi_num_ranks = MPI.COMM_WORLD.Get_size()
    if mpi_rank == 0:
        rand = random.Random(12345)
        groups = [
            (
                nccl.get_unique_id(),
                rand.sample(range(mpi_num_ranks), rand.randrange(1, mpi_num_ranks)),
            )
            for _ in range(50)
        ]
    groups = MPI.COMM_WORLD.bcast(groups, root=0)
    return groups


def create_comm(unique_id, group, comms):
    mpi_rank = MPI.COMM_WORLD.Get_rank()
    if mpi_rank not in group:
        return

    comm = nccl.NcclCommunicator(len(group), unique_id, group.index(mpi_rank))
    comms.append(comm)


def serial_create():
    groups = get_comm_groups()

    comms = []
    for unique_id, group in groups:
        create_comm(unique_id, group, comms)
    MPI.COMM_WORLD.Barrier()


def parallel_create():
    groups = get_comm_groups()

    comms = []
    threads = [
        threading.Thread(target=create_comm, args=(unique_id, group, comms))
        for unique_id, group in groups
    ]
    # without the patch, shuffling should make us deadlock pretty consistently
    # without the patch and without the shuffle, there's no benefit to creating comms in parallel without the patch
    # with the patch, we don't deadlock and we get a speedup from creating comms in parallel
    random.shuffle(threads)
    for t in threads:
        t.start()
    for t in threads:
        t.join()
    MPI.COMM_WORLD.Barrier()


if __name__ == "__main__":
    rank = MPI.COMM_WORLD.Get_rank()
    node = platform.node()
    rank_nodes = MPI.COMM_WORLD.allgather((rank, node))
    local_rank = sum(1 for r, n in rank_nodes if r < rank and n == node)
    os.environ["CUDA_VISIBLE_DEVICES"] = str(local_rank % 8)
    MPI.COMM_WORLD.Barrier()

    start = time.time()

    if sys.argv[1] == "serial":
        serial_create()
    elif sys.argv[1] == "parallel":
        parallel_create()
    else:
        raise ValueError("Invalid argument")

    end = time.time()

    print(f"Time taken: {end - start:.2f}")
```

emcastillo:
It would be interesting to see a stack trace of where the processes are stuck.

hauntsaninja:
When you deadlock after creating NcclCommunicators in different orders on different ranks (in threads)? It's what one would expect.

Python stacktrace:
```
Thread 66880 (idle): "Thread-50"
    create_comm (cupyrepro.py:34)
    run (threading.py:865)
    _bootstrap_inner (threading.py:917)
    _bootstrap (threading.py:885)
```
C stacktrace:
```
#0  0x00007faddbdf87c7 in __libc_accept (fd=65, addr=..., len=0x7fadd29bc980) at ../sysdeps/unix/sysv/linux/accept.c:26
#1  0x00007fad7a251232 in ?? () from /usr/lib/x86_64-linux-gnu/libnccl.so.2
#2  0x00007fad7a23edcb in ?? () from /usr/lib/x86_64-linux-gnu/libnccl.so.2
#3  0x00007fad7a242460 in ?? () from /usr/lib/x86_64-linux-gnu/libnccl.so.2
#4  0x00007fad7a2430a0 in ?? () from /usr/lib/x86_64-linux-gnu/libnccl.so.2
#5  0x00007fad7a24331f in ncclCommInitRank () from /usr/lib/x86_64-linux-gnu/libnccl.so.2
#6  0x00007fad7d9449a3 in __pyx_pf_13cupy_backends_4cuda_4libs_4nccl_16NcclCommunicator_2__init__ (__pyx_v_rank=4, __pyx_v_commId=0x55a6b18db618, __pyx_v_ndev=14, 
    __pyx_v_self=0x7fad76f318f0) at cupy_backends/cuda/libs/nccl.cpp:3806
#7  __pyx_pw_13cupy_backends_4cuda_4libs_4nccl_16NcclCommunicator_3__init__ (__pyx_v_self=0x7fad76f318f0, __pyx_args=<optimized out>, __pyx_kwds=<optimized out>)
    at cupy_backends/cuda/libs/nccl.cpp:3720
#8  0x000055a6aec429b8 in type_call (kwds=0x0, args=0x7fad76f45990, type=<optimized out>) at /tmp/build/80754af9/python_1553721932202/work/Objects/typeobject.c:949
#9  _PyObject_FastCallKeywords () at /tmp/build/80754af9/python_1553721932202/work/Objects/call.c:199
#10 0x000055a6aec9e7ae in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>) at /tmp/build/80754af9/python_1553721932202/work/Python/ceval.c:4619
...
```

Because this patch changes things so that the GIL isn't held during `ncclCommInitRank`, other threads are able to progress and create their comms / comms creation actually happens in parallel and saves time.

Lmk if I was unclear about something or if the example in my previous message gives you different results :-)

emcastillo:
Nice! understood now :) thanks!!!

kmaehashi (approved):
LGTM!

kmaehashi:
pfnCI, test this please.

PR #543, opened 2017-09-27
Fix broadcast

niboshi:
Fixes #532

PR #5579, opened 2021-07-29
Fix broadcast error messages

toslunar:
Example: `ValueError: operands could not be broadcast together with shapes (2,) (3,)`

- lower case "operands"
- `' '.join` instead of `', '.join` shapes

### DIFF
--- a/cupy/_core/internal.pyx
+++ b/cupy/_core/internal.pyx
@@ -18,9 +18,6 @@ cdef extern from 'halffloat.h':
     uint32_t npy_halfbits_to_floatbits(uint16_t h)
 
 
-cdef Py_ssize_t PY_SSIZE_T_MAX = sys.maxsize
-
-
 @cython.profile(False)
 cpdef inline Py_ssize_t prod(const vector.vector[Py_ssize_t]& args):
     cdef Py_ssize_t n = 1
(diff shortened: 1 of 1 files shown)


asi1024:
Jenkins, test this please.

asi1024 (approved):

cupy/_core/internal.pyx
```diff
@@ -331,20 +328,24 @@ cdef _broadcast_core(list arrays, shape_t& shape):
...
+                continue
+            a_sh = a._shape[i - (nd - a_ndim)]
+            if a_sh == s or a_sh == 1:
+                continue
+            if s == 1:
+                s = a_sh
+                continue
             raise ValueError(
-                'shape mismatch: objects cannot be broadcast to a '
-                'single shape')
```
toslunar, line 346: FYI, the error message (added in #543) seems from [`PyArray_Broadcast`](https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_Broadcast) in NumPy C-API.
https://github.com/numpy/numpy/blob/v1.21.1/numpy/core/src/multiarray/iterators.c#L1159-L1162

asi1024:
Jenkins, test this please.

PR #5585, opened 2021-08-01
Add cupyx.scipy.sparse.linalg.minres

povinsahu1909:
This PR add MINRES solver to sparse linear algebra.

### DIFF
--- a/cupyx/scipy/sparse/linalg/_solve.py
+++ b/cupyx/scipy/sparse/linalg/_solve.py
@@ -832,7 +832,7 @@ def minres(A, b, x0=None, shift=0.0, tol=1e-5, maxiter=None,
     sn = 0
     w = cupy.zeros(n, dtype=xtype)
     w2 = cupy.zeros(n, dtype=xtype)
-    r2 = r1.copy()
+    r2 = r1
 
     while itn < maxiter:
 
@@ -847,15 +847,18 @@ def minres(A, b, x0=None, shift=0.0, tol=1e-5, maxiter=None,
             y -= (beta / oldb) * r1
 
         alpha = cupy.inner(v, y)
+        alpha = alpha.get().item()
         y -= (alpha / beta) * r2
-        r1 = r2.copy()
-        r2 = y.copy()
+        r1 = r2
+        r2 = y
         y = psolve(r2)
         oldb = beta
         beta = cupy.inner(r2, y)
+        beta = cupy.sqrt(beta)
+        beta = beta.get().item()
         if beta < 0:
             raise ValueError('non-symmetric matrix')
-        beta = cupy.sqrt(beta)
+
         tnorm2 += alpha ** 2 + oldb ** 2 + beta ** 2
 
         if itn == 1:
@@ -885,8 +888,8 @@ def minres(A, b, x0=None, shift=0.0, tol=1e-5, maxiter=None,
         # Update  x.
 
         denom = 1.0 / gamma
-        w1 = w2.copy()
-        w2 = w.copy()
+        w1 = w2
+        w2 = w
         w = (v - oldeps * w1 - delta * w2) * denom
         x += phi * w
 
@@ -900,10 +903,12 @@ def minres(A, b, x0=None, shift=0.0, tol=1e-5, maxiter=None,
 
         # Estimate various norms and test for convergence.
 
-        Anorm = cupy.sqrt(tnorm2)
+        Anorm = numpy.sqrt(tnorm2)
         ynorm = cupy.linalg.norm(x)
+        ynorm = ynorm.get().item()
         epsa = Anorm * eps
         epsx = Anorm * ynorm * eps
+        epsr = Anorm * ynorm * tol
         diag = gbar
 
         if diag == 0:
--- a/cupyx/scipy/sparse/linalg/_solve.py
+++ b/cupyx/scipy/sparse/linalg/_solve.py
@@ -744,7 +744,7 @@ def _symOrtho(a, b):
 
 
 def minres(A, b, x0=None, shift=0.0, tol=1e-5, maxiter=None,
-           M=None, callback=None, show=False, check=False):
+           M=None, callback=None, check=False):
     A, M, x, b = _make_system(A, M, x0, b)
 
     matvec = A.matvec
@@ -775,17 +775,17 @@ def minres(A, b, x0=None, shift=0.0, tol=1e-5, maxiter=None,
     if beta1 < 0:
         raise ValueError('indefinite preconditioner')
     elif beta1 == 0:
-        return (x, 0)
+        return x, 0
 
     beta1 = cupy.sqrt(beta1)
     beta1 = beta1.get().item()
-    if check:
 
-        # # see if A is symmetric
+    if check:
+        # see if A is symmetric
         if not _check_symmetric(A, Ax, x, eps):
             raise ValueError('non-symmetric matrix')
 
-        # # see if M is symmetric
+        # see if M is symmetric
         if not _check_symmetric(M, y, r1, eps):
             raise ValueError('non-symmetric preconditioner')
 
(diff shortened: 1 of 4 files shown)


povinsahu1909:
Following is the performance comparision result measured on GTX1050

size | scipy | cupy 
--- | --- | ---
100 |30.786543846130375| 947.5719909667969
500 |41.63404846191406 |1094.1433105468752
1000| 48.57564735412598| 1165.3687133789062
5000 |309.1716766357422 |1239.9229736328125
10000| 1342.04052734375 |1101.5445556640627
15000 |3401.540649414063| 1478.1609497070312

<details>
<summary> script </summary>

```python

import numpy as np
import cupy as cp
from cupyx.time import repeat
from scipy.sparse.linalg import minres as minr
import scipy.sparse 
from cupyx.scipy.sparse import csr_matrix
from cupyx.scipy.sparse.linalg import minres 

def get_time(perf):

i = 0
times = []
print('m, scipy, cupy')
for m in (100, 500, 1000, 5000, 10000, 15000, ):
    a = scipy.sparse.random(m, m, density=0.01) 
    b = np.random.rand(m)
    times.append(get_time(repeat(minr, (a,b), n_repeat=2, n_warmup=1))) 
    a = csr_matrix(a)
    b = cp.array(b) 
    times.append(get_time(repeat(minres, (a,b), n_repeat=2, n_warmup=1)))
    print(m , times[i], times[i+1])
    i+=2
```

takagi:
pfnCI, test this please.

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,238 @@ def _symOrtho(a, b):
...
+
+        y = matvec(v)
+        y -= shift * v
+
+        if itn >= 2:
+            y -= (beta / oldb) * r1
+
+        alpha = cupy.inner(v, y)
+        y -= (alpha / beta) * r2
+        r1 = r2.copy()
```
takagi, line 851: Do we need `r2` copied as `r2` is assigned to another ndarray, `y.copy()`, and the ndarray that was assigned to `r2` is not used anymore?

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,238 @@ def _symOrtho(a, b):
...
+        r2 = y.copy()
+        y = psolve(r2)
+        oldb = beta
+        beta = cupy.inner(r2, y)
```
takagi, line 855: Currently, `beta` is assigned to a ndarray on the device, causing multiple synchronizations in the following lines. How about synchronizing it to the host soon, and using the host value below?

takagi: Also, it would be better to make `beta` always keep host value for clarity.

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,238 @@ def _symOrtho(a, b):
...
+        s = 1.0 / beta
+        v = s * y
+
+        y = matvec(v)
+        y -= shift * v
+
+        if itn >= 2:
+            y -= (beta / oldb) * r1
+
+        alpha = cupy.inner(v, y)
```
takagi, line 849: Seems `alpha` is used for scalar arithmetics below, so it would be better to synchronize it to the host before those.

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,238 @@ def _symOrtho(a, b):
...
+        y -= (alpha / beta) * r2
+        r1 = r2.copy()
+        r2 = y.copy()
```
takagi, line 852: Also, do we need `y` copied here?

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,238 @@ def _symOrtho(a, b):
...
+        cs = gbar / gamma  # ck
+        sn = beta / gamma  # sk
+        phi = cs * phibar  # phik
+        phibar = sn * phibar  # phibark+1
+
+        # Update  x.
+
+        denom = 1.0 / gamma
+        w1 = w2.copy()
+        w2 = w.copy()
```
takagi, line 889: Do we need these copies?

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,238 @@ def _symOrtho(a, b):
...
+        gmax = max(gmax, gamma)
+        gmin = min(gmin, gamma)
+        z = rhs1 / gamma
+        rhs1 = rhs2 - delta * z
+        rhs2 = - epsln * z
+
+        # Estimate various norms and test for convergence.
+
+        Anorm = cupy.sqrt(tnorm2)
+        ynorm = cupy.linalg.norm(x)
```
takagi, line 904: How about making `tnorm2` always have host value, and computing it on the host?

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,238 @@ def _symOrtho(a, b):
...
+
+        gmax = max(gmax, gamma)
+        gmin = min(gmin, gamma)
+        z = rhs1 / gamma
+        rhs1 = rhs2 - delta * z
+        rhs2 = - epsln * z
+
+        # Estimate various norms and test for convergence.
+
+        Anorm = cupy.sqrt(tnorm2)
```
takagi, line 903: How about making `tnorm2` always have host value, and computing it on the host?

cupyx/scipy/sparse/linalg/_solve.py
takagi, line 904: Synchronizing `ynorm` to the host here would be better as it is used in scalar arithmetics and comparison below.

takagi:
For NumPy-CuPy mixed code like this, codes don't obviously say each computation is done on the host or the device. So it would be better for you first to think computation model, or which computation you want to do on the host and which on the device, then translating it to the code (instead of thinking of the code itself).

povinsahu1909:
>Following is the performance comparision result measured on GTX1050

There was a increase in performance after synchronization
size | scipy | cupy
--- | --- | ---
100 |29.10331249237061 |270.47564697265625
500 |42.720144271850586 |324.1162109375
1000 |48.38244819641113 |336.3395233154297
5000 |250.0668640136719 |311.1663513183594
10000 |1388.4001464843752 |374.30403137207037
15000 |3601.8338623046875 |521.3742980957032

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -739,3 +740,243 @@ def _symOrtho(a, b):
...
+        alpha = cupy.inner(v, y)
+        alpha = alpha.get().item()
+        y -= (alpha / beta) * r2
+        r1 = r2
+        r2 = y
+        y = psolve(r2)
+        oldb = beta
+        beta = cupy.inner(r2, y)
+        beta = cupy.sqrt(beta)
+        beta = beta.get().item()
```
takagi, line 858: `beta` is a scalar so faster on CPU (maybe a couple of percent gain)
```suggestion
        beta = beta.get().item()
        beta = numpy.sqrt(beta)
```

takagi:
pfnCI, test this please.

takagi:
I analyzed the change with Nsight and found no scalar arithmetic performed on the device.

takagi:
Same script at https://github.com/cupy/cupy/pull/5585#issuecomment-890582598 on A100.

m|scipy|cupy
---|---|---
100|33.018672943115234|241.35374450683594
500|37.46206283569336|279.8908386230469
1000|40.40849685668946|263.1447448730469
5000|215.61602020263672|271.46942138671875
10000|805.2312622070312|272.2580261230469
15000|3429.39111328125|323.3358459472656

takagi (approved):
LGTM!

takagi:
Thank you!

Issue #2840, opened 2019-12-19
NCCL Python Wrapper 


cjnolet:
There are a lot of use-cases for having NCCL run directly in the Python layer. It looks like the current wrapper requires the user to work at the level of ctypes pointers rather than native Python objects. 

There is user interest in being able to initialize a NCCL clique and pass cupy arrays directly to the collective comms functions. This opens up a lot of possibilities for users and I have also seen interest from the NCCL project in having such a wrapper made.

Included in this feature, should also be some well-documented examples of using the new NCCL Python API.

leofang:
It’s not that complicated. The underlying device pointer of CuPy arrays is exposed via either `arr.data.ptr` or `arr.__cuda_array_interface__['data'][0]`. Once this is known, making a Python wrapper is straightforward without using ctypes, which I hate 😉

For example, this is a thin wrapper we used in one of our projects:
```python
import numpy as np
import cupy as cp
try:
    from cupy.cuda import nccl
except ImportError:
    import sys
    print("NCCL does not exist.", file=sys.stderr)
    nccl = None
    del sys

if nccl is not None:
    def get_unique_id():
        return nccl.get_unique_id()

    class NCCLComm(nccl.NcclCommunicator):
        def __init__(self, ndev, commId, rank):
            super(NCCLComm, self).__init__(ndev, commId, rank)
    
        def get_NCCL_count_dtype(self, arr):
            if arr.dtype == cp.complex64:
                return arr.size*2, nccl.NCCL_FLOAT32
            elif arr.dtype == cp.complex128:
                return arr.size*2, nccl.NCCL_FLOAT64
            elif arr.dtype == cp.float32:
                return arr.size, nccl.NCCL_FLOAT32
            elif arr.dtype == cp.float64:
                return arr.size, nccl.NCCL_FLOAT64
            else:
                raise ValueError("This dtype is not supported by NCCL.")
    
        def Allreduce(self, send_arr, recv_arr, op=nccl.NCCL_SUM, stream=None):
            sendbuf = send_arr.__cuda_array_interface__['data'][0]
            recvbuf = recv_arr.__cuda_array_interface__['data'][0]
            count, datatype = self.get_NCCL_count_dtype(send_arr)
            if stream is None:
                stream = cp.cuda.Stream.null.ptr
            else:
                stream = stream.ptr
            super(NCCLComm, self).allReduce(sendbuf, recvbuf, count, datatype, op, stream)
    
[...]

PR #5590, opened 2021-08-03
Add `cupyx.distributed`

emcastillo:
Heavily influenced by https://github.com/cupy/cupy/issues/2840

TODO:
- [x] Implement collectives that are not in nccl using the existing ones (alltoall)
- [x] tests
- [x] docs

Are there MPI-Esche features that you guys would like this to support? 
splitting communicators, process management stuff (`ncclrun` tool analogous to `mpirun`)

cc @cjnolet @anaruse @pentschev @leofang


emcastillo:
Added additional collective calls as described in:
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/p2p.html#one-to-all-scatter

leofang (commented):
Thanks, Emilio! I have a few high-level comments:
1. `NCCLCommunicator` is already a cpdef class. Do we really need another layer on top of it? Can't we just add the new methods to it with uppercase names? This would align with mpi4py's convention (ex: `Allreduce()`, `Reduce()`, etc). The lowercase methods (ex: `allReduce()`, `reduce()`, etc) can be kept for backward compatibility. 
2. `_get_nccl_dtype_and_count` and `_get_op` should be merged. For complex numbers only `NCCL_SUM` is applicable, so we need a check somewhere.

emcastillo:
Thanks!!

> NCCLCommunicator is already a cpdef class. Do we really need another layer on top of it? Can't we just add the new methods to it with uppercase names? This would align with mpi4py's convention (ex: Allreduce(), Reduce(), etc). The lowercase methods (ex: allReduce(), reduce(), etc) can be kept for backward compatibility.

I am not sure if we should expose the low-level methods with names just changing a letter to regular users. I would like to have a clean API for this object so this is why I didn't extend the original communicator class.

> _get_nccl_dtype_and_count and _get_op should be merged. For complex numbers only NCCL_SUM is applicable, so we need a check somewhere.

This is a good point, thanks for the hint about `NCCL_SUM`, Indeed merging the two methods seems the way to go for implementing this check, but I don't feel it is clean :(.

emcastillo:
I added a simple TCPStore mechanism to broadcast data from the cpu and initialize nccl without mpi as in https://forums.developer.nvidia.com/t/nccl2-across-multiple-nodes-without-mpi/60243

still wip and yet to be tested, but this will allow us to have a `cupy.distributed` module that is similar to torch and does not require ext libraries.

cupyx/distributed/_comm.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+_nccl_ops = {'sum': nccl.NCCL_SUM,
+             'prod': nccl.NCCL_PROD,
+             'max': nccl.NCCL_MAX,
+             'min': nccl.NCCL_MIN}
+
+
+class NCCL:
+    def __init__(self, n_devices, rank):
```
leofang, line 28: Usually `rank` is from MPI or multiprocessing. If we have our own TCPStore that can bypass any of these bootstrapping mechanism, I suppose we should also come up with a way to determine `rank`?

leofang (commented):
Looks interesting, Emilio! Have a quick question for now.

cupyx/distributed/_comm.py
emcastillo: Yeah, I've been banging my head with this and an easy way to determine the rank 0 process automatically since it is the one that starts the tcpstore. But my idea required using filesystem locks, matching the hostname and doing several other tricks ... In the end I just thought of manually assigning the ranks as pytorch does :D.

cupyx/distributed/_store.py
```diff
@@ -0,0 +1,154 @@
...
+                store._current_barrier = None
+            return True
+
+    def _set_process(self, process):
+        self._process = process
+
+    def _process_request(self, c_socket):
+        with c_socket:
+            data = c_socket.recv(1024)
+            r = pickle.loads(data)(self)
```
emcastillo, line 77: Security ...

kmaehashi: As discussed offline, if we implement a server socket in a product we need to take care of security things. I think the very basic KLV binary format is sufficient as we don't have to worry about protocol versioning.

Please also add a disclaimer in docstring that:
* use of this function launches the server
* the functionality is expected to be used in a trusted cluster environment and there is no authentication systems

emcastillo:
PTAL

cupyx/distributed/_comm.py
emcastillo: I will do this in a follow up PR that boots and agent per each node in a similar way to pytorch.

tests/cupyx_tests/distributed_tests/test_comm.py
```diff
@@ -0,0 +1,286 @@
...
+            if rank == root:
+                in_array = expected
+            else:
+                in_array = cupy.zeros((2, 3, 4), dtype=dtype)
+
+            comm.broadcast(in_array, root)
+            testing.assert_allclose(in_array, expected)
+
+        self._launch_workers(N_WORKERS, run_broadcast, (0, dtype))
+        self._launch_workers(N_WORKERS, run_broadcast, (1, dtype))
```
leofang, line 54: This looks strange... Don't you already have a loop in `_launch_workers` to generate rank 0 & 1? Why do we need to make 2 calls here?

emcastillo: it launches the broadcast two times, with a different root process.
Could have parametrized it but was not in the mood I guess.

leofang: Ah OK, thanks for the explanation!

kmaehashi (changes requested):
LGTM except for nitpicks!

cupyx/distributed/_init.py
```diff
@@ -0,0 +1,88 @@
+import os
+
+from cupy_backends.cuda.libs import nccl
```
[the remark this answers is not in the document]
emcastillo: good catch! this was because `cupy.cuda.nccl` was importable in absence of nccl, but the cupy_backends one raises an error. Thanks!!

kmaehashi:
Sorry, let me takedown the previous comment. I'm now thinking it is better to use `import cupy.cuda.nccl` instead of `cupy_backends` as `import cupyx` does not import `import cupyx.distributed` automatically, and users who does `import cupyx.distributed` will surely want to use NCCL.

emcastillo:
Changed imports! PTAL!

kmaehashi:
pfnCI, test this please.

kmaehashi:
@emcastillo Could you check test failures?

emcastillo:
templatized the tests to avoid the error because we can't fork when the cuda context was created ...

tests/cupyx_tests/distributed_tests/test_comm_template.py.txt
```diff
@@ -0,0 +1,254 @@
...
+            in_array = in_array + 1
+        comm.broadcast(in_array, 0)
+        testing.assert_allclose(in_array, cupy.ones(1))
+
+    _launch_workers(N_WORKERS, run_init)
+
+
+if __name__ == '__main__':
+    # Run the templatized test
+    {name}({dtype})
```
kmaehashi, line 254: How about using command line arguments instead of templates? I'd like to avoid templates if possible, because it's fragile (e.g. if we write `print(f'{dtype}')` in this file, it break...)


```
func = globals()[sys.argv[1]]
dtype = getattr(cupy, sys.argv[2])
func(dtype)
```

In that case let's name this script `comm_runner.py` or sth (so that flake8 can be applied and pytest does not pick it up as a test script)

emcastillo: Done!!

kmaehashi (dismissed):

emcastillo:
/test mini

(docs passed!)

PR #5597, opened 2021-08-04
Mark `lsmr` tests xfail for CSR matrices on HIP

leofang:
Follow-up of #5331. Looks like both CSC and CSR matrices do not work for the moment.


takagi:
pfnCI, test this please.

takagi (approved):

PR #5600, opened 2021-08-04
Fix compute capability check

kmaehashi:
When no GPUs are available and ``CUPY_NVCC_GENERATE_CODE=current`` is set, invalid CC is printed (I got `--generate-code=arch=compute_211033746432543,code=sm_211033746432543`).


emcastillo (approved):

toslunar:
Jenkins, test this please.

PR #5607, opened 2021-08-05
Reload GPG key for Docker images

kmaehashi:
GPG expired so `apt` stopped working in already-released Docker containers.


emcastillo (approved):
LGTM

PR #5614, opened 2021-08-06
Add `batch_identity` helper

toslunar:
`linalg` functions often use stacked identity matrices.


cupy/linalg/_decomposition.py
```diff
@@ -228,11 +228,7 @@ def _qr_batched(a, mode):
             return (cupy.empty(batch_shape + (m, 0), out_dtype),
                     cupy.empty(batch_shape + (0, n), out_dtype))
         elif mode == 'complete':
-            if batch_size == 0:
-                q = cupy.empty(batch_shape + (m, m), out_dtype)
-            else:
-                q = cupy.eye(m)
-                q = cupy.stack([q for i in range(batch_size)])
```
toslunar, line 235: `dtype=out_dtype` and `.reshape` were missing.

cupy/linalg/_util.py
```diff
@@ -151,3 +151,19 @@ def _tril(x, k=0):
 def _triu(x, k=0):
     _triu_kernel(k, x)
     return x
+
+
+def batch_identity(batch_shape, n, dtype):
+    shape = batch_shape + (n, n)
+    idx = cupy.arange(n)
+    x = cupy.zeros(shape, dtype)
+    x[..., idx, idx] = 1
```
toslunar, line 160: The line can be `x.diagonal(axis1=-2, axis2=-1)[...] = 1` because CuPy doesn't make it readonly.

cupy/linalg/_decomposition.py
leofang: Good catch, thanks!

cupy/linalg/_util.py
```diff
@@ -151,3 +151,19 @@ def _tril(x, k=0):
 def _triu(x, k=0):
     _triu_kernel(k, x)
     return x
+
+
+def batch_identity(batch_shape, n, dtype):
```
leofang, line 156: I noticed `_util` is not a public module, so it's OK that we don't prefix the function name with an underscore.

cupy/linalg/_util.py
```diff
@@ -151,3 +151,19 @@ def _tril(x, k=0):
...
+    shape = batch_shape + (n, n)
+    idx = cupy.arange(n)
+    x = cupy.zeros(shape, dtype)
+    x[..., idx, idx] = 1
+    return x
+
+
+def batch_identity_like(x):
```
leofang, line 164: Should we make sure `x` is a (batch of) square matrices here?

leofang (approved):
LGTM except for one question.

toslunar:
It may be better to fix `_assert_nd_squareness` first (#5642).

cupy/linalg/_util.py
toslunar: Added a comment to the code.

asi1024:
Jenkins, test this please.

leofang (approved):

asi1024 (approved):

asi1024:
Sorry for my late response. LGTM!

PR #5575, opened 2021-07-29
Increase timeout for CUDA 11.4 tests

kmaehashi:
It seems that SASS fix (#5097) increased the compilation time.

Previously, running all unit tests without cache took:

* `CUPY_TEST_FULL_COMBINATION=1` 3h53m (https://ci.preferred.jp/cupy.linux.cuda114/75345/)
* `CUPY_TEST_FULL_COMBINATION=0` 2h8m (https://ci.preferred.jp/cupy.linux.cuda114/75362/)

Now it takes more than 4h (`CUPY_TEST_FULL_COMBINATION=1`):
https://ci.preferred.jp/cupy.linux.cuda114/75868/

PR #5550, opened 2021-07-21
Fix casts in ufunc outputs

toslunar:
This PR fixes #5527 and simplifies `elementwise_copy` (e.g. reverts #5410 except for the test).

See also #5539.

PR #5622, opened 2021-08-08
Allow to compile using PTX with an envvar

emcastillo:
Seems like SASS (CUBIN) increases compilation times in a non-negligible amount (specially when running the test-suite).

Adds `CUPY_COMPILE_WITH_PTX` env var to explicitly request the older behavior.

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -160,10 +160,8 @@ def _get_arch_for_options_for_nvrtc(arch=None):
     # generate cubin (SASS) instead of PTX. See #5097 for details.
     if arch is None:
         arch = _get_arch()
-    use_ptx = _get_bool_env_variable(
-        'CUPY_COMPILE_WITH_PTX', False)
     if (
-        not use_ptx and _cuda_hip_version >= 11010
+        not _use_ptx and _cuda_hip_version >= 11010
         and arch < _get_max_compute_capability()
     ):
         return f'-arch=sm_{arch}', 'cubin'
@@ -217,6 +215,7 @@ def _get_bool_env_variable(name, default):
         return False
 
 
+_use_ptx = _get_bool_env_variable('CUPY_COMPILE_WITH_PTX', False)
 _jitify_header_source_map_populated = False
 
 
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -160,7 +160,12 @@ def _get_arch_for_options_for_nvrtc(arch=None):
     # generate cubin (SASS) instead of PTX. See #5097 for details.
     if arch is None:
         arch = _get_arch()
-    if _cuda_hip_version >= 11010 and arch < _get_max_compute_capability():
+    use_ptx = _get_bool_env_variable(
+        'CUPY_COMPILE_WITH_PTX', False)
+    if (
+        not use_ptx and _cuda_hip_version >= 11010
+        and arch < _get_max_compute_capability()
+    ):
         return f'-arch=sm_{arch}', 'cubin'
     return f'-arch=compute_{arch}', 'ptx'
 
--- a/docs/source/reference/environment.rst
+++ b/docs/source/reference/environment.rst
@@ -118,7 +118,7 @@ Here are the environment variables that CuPy uses at runtime.
 
   By default, CuPy directly compiles kernels into SASS (CUBIN) to support `CUDA Enhanced Compatibility <https://docs.nvidia.com/deploy/cuda-compatibility/>`_
   If set to ``1``, CuPy instead compiles kernels into PTX and lets CUDA Driver assemble SASS from PTX.
-  This option is only effective for CUDA 11.1 or later; CuPy always compiles into PTX on earlier CUDA versions. Also, this option only applies when NVRTC is selected as the compilation backend.
+  This option is only effective for CUDA 11.1 or later; CuPy always compiles into PTX on earlier CUDA versions. Also, this option only applies when NVRTC is selected as the compilation backend. NVCC backend always compiles into SASS (CUBIN).
 
 CUDA Toolkit Environment Variables
   In addition to the environment variables listed above, as in any CUDA programs, all of the CUDA environment variables listed in the `CUDA Toolkit Documentation`_ will also be honored.
--- a/docs/source/reference/environment.rst
+++ b/docs/source/reference/environment.rst
@@ -116,8 +116,9 @@ Here are the environment variables that CuPy uses at runtime.
 
   Default: ``0``
 
-  If set to ``1``, CuPy will compile kernels using PTX instead of SSAS (CUBIN), PTX results in faster compilation time at the expenses of non-compatibility with drivers older than the toolkit.
-  If set to ``0`` (default), SASS (CUBIN) compilation mode is used for kernels.
+  By default, CuPy directly compiles kernels into SASS (CUBIN) to support `CUDA Enhanced Compatibility <https://docs.nvidia.com/deploy/cuda-compatibility/>`_
+  If set to ``1``, CuPy instead compiles kernels into PTX and lets CUDA Driver assemble SASS from PTX.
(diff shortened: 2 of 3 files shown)


leofang:
Hmmm this is a bit strange as I think I did some timing (around https://github.com/cupy/cupy/pull/4558#issuecomment-765610834) and concluded that getCUBIN() should be enabled by default.
1. If the generated cubins are correctly cached, we only pay the overhead once and the subsequent reuses are quite fast.
2. getCUBIN() spends time on generating PTX and SASS, same as getPTX() + cuModuleLoad(), so if we start with a fresh environment without any cache (CuPy's or CUDA driver's), as I explored in the above link we should get close time spans.
3. For correctness getCUBIN() should be _opted out_, i.e., enabled by default. Otherwise we'd keep hitting the driver bugs.

emcastillo:
this pr sets cubin by default but leaves the old behavior just in case it may be needed.
I saw that compile times in the CI almost doubled with the cubin change, but i cant recall who showed me it.

leofang:
I meant the CI should be able to take advantage of cached cubins. It would be indeed programatic if *every* CI run is lengthened, but it doesn't seem `cupy_pr` showed this trend?

kmaehashi:
I've benchmarked this branch but SASS seems to have not much impact on time.
(It can be possible that the difference will get larger when compiling more complex kernels, though)


```
% rm ~/.cupy/kernel_cache ~/.nv -rf; CUPY_COMPILE_WITH_PTX=0 python benchmark.py
/home/maehashi/Development/cupy/cupyx/time.py:115: FutureWarning: cupyx.time.repeat is experimental. The interface can change in the future.
  _util.experimental('cupyx.time.repeat')
test                :    CPU: 3432.153 us     GPU-0: 3467.072 us
% rm ~/.cupy/kernel_cache ~/.nv -rf; CUPY_COMPILE_WITH_PTX=1 python benchmark.py
/home/maehashi/Development/cupy/cupyx/time.py:115: FutureWarning: cupyx.time.repeat is experimental. The interface can change in the future.
  _util.experimental('cupyx.time.repeat')
test                :    CPU: 3320.955 us     GPU-0: 3361.504 us
```

Code:
```py
import cupy as cp
from cupyx.time import repeat


def test(count):
    for value in range(count):
        add_kernel = cp.RawKernel('''
            extern "C" __global__
            void my_add(const float* x1, const float* x2, float* y) {
                int tid = blockDim.x * blockIdx.x + threadIdx.x;
                y[tid] = x1[tid] + x2[tid] + ''' + str(value) + ''';
            }
            ''', 'my_add')
        x = cp.arange(25, dtype=cp.float32).reshape(5, 5)
        y = cp.zeros((5, 5), dtype=cp.float32)
        add_kernel((5,), (5,), (x, x, y))


print(repeat(test, (100,), n_repeat=1))
```

emcastillo:
Thanks!

I just felt we should have the option to keep the old behavior around if needed, not saying that this is explicitly for CI

leofang:
> we should have the option to keep the old behavior around if needed

Sorry I was being silly. I agree with @emcastillo that making `CUPY_COMPILE_WITH_PTX` available (but defaulted to 0) should be useful.

emcastillo:
unless i screwed it up big time (as always) 
it should be defaulted to 0 😄

kmaehashi:
To clarify: I also have no objection to introducing this switch 😃

Actually, SASS increased the test time a bit, but it is almost negligible. Running tests from scratch (without cache) took (https://github.com/cupy/cupy/pull/5575):

* Before SASS PR: 3h53m
* After SASS PR: 3h58m

I found that the recent test time increase is due to #5550. It contains new tests like this:

```
    @pytest.mark.parametrize('casting', [
        pytest.param('no', marks=pytest.mark.xfail(strict=False)),
        pytest.param('equiv', marks=pytest.mark.xfail(strict=False)),
        'safe',
        'same_kind',
    ])
    @testing.for_all_dtypes_combination(
        names=['in0_type', 'in1_type', 'out_type'])
    @testing.numpy_cupy_allclose(accept_error=TypeError)
    def test_casting_out(self, xp, in0_type, in1_type, out_type, casting):
```

This is generating `casting (4) * dtype (16 ** 3) = 16,384` parameterization. I feel this is a bit overkill.

cupy/cuda/compiler.py
```diff
@@ -160,7 +160,12 @@ def _get_arch_for_options_for_nvrtc(arch=None):
     # generate cubin (SASS) instead of PTX. See #5097 for details.
     if arch is None:
         arch = _get_arch()
-    if _cuda_hip_version >= 11010 and arch < _get_max_compute_capability():
+    use_ptx = _get_bool_env_variable(
```
kmaehashi, line 163: Could you move this code to module-level to avoid getting env var everytime?

docs/source/reference/environment.rst
```diff
@@ -112,6 +112,13 @@ Here are the environment variables that CuPy uses at runtime.
   If set to ``1``, CuPy will use the CUDA per-thread default stream, effectively causing each host thread to automatically execute in its own stream, unless the CUDA default (``null``) stream or a user-created stream is specified.
   If set to ``0`` (default), the CUDA default (``null``) stream is used, unless the per-thread default stream (``ptds``) or a user-created stream is specified.
 
+.. envvar:: CUPY_COMPILE_WITH_PTX
+
+  Default: ``0``
+
+  If set to ``1``, CuPy will compile kernels using PTX instead of SSAS (CUBIN), PTX results in faster compilation time at the expenses of non-compatibility with drivers older than the toolkit.
```
kmaehashi, line 119: This option should only be used for those who *really* need this, so I want to avoid documenting it as if it is a performance tuning parameter. How about saying like this?

```
By default, CuPy directly compiles kernels into SASS (CUBIN) to support `CUDA Enhanced Compatibility <https://docs.nvidia.com/deploy/cuda-compatibility/>`_.
If set to ``1``, CuPy instead compiles kernels into PTX and let CUDA Driver assemble SASS from PTX.
This option is only effective for CUDA 11.1 or later; it will always compiles into PTX on earlier CUDA versions.
```

Question: what happens if `backend='nvcc'`?

kmaehashi (changes requested):

docs/source/reference/environment.rst
emcastillo: This is only supported for nvrtc, I will mention that in the docs

cupy/cuda/compiler.py
emcastillo: FYI, there is a lot of code who calls this every time, so probably we should fix all of these in a separate PR.

emcastillo:
PTAL

docs/source/reference/environment.rst
```diff
@@ -116,8 +116,9 @@ Here are the environment variables that CuPy uses at runtime.
 
   Default: ``0``
 
-  If set to ``1``, CuPy will compile kernels using PTX instead of SSAS (CUBIN), PTX results in faster compilation time at the expenses of non-compatibility with drivers older than the toolkit.
-  If set to ``0`` (default), SASS (CUBIN) compilation mode is used for kernels.
+  By default, CuPy directly compiles kernels into SASS (CUBIN) to support `CUDA Enhanced Compatibility <https://docs.nvidia.com/deploy/cuda-compatibility/>`_
+  If set to ``1``, CuPy instead compiles kernels into PTX and lets CUDA Driver assemble SASS from PTX.
+  This option is only effective for CUDA 11.1 or later; CuPy always compiles into PTX on earlier CUDA versions. Also, this option only applies when NVRTC is selected as the compilation backend.
```
kmaehashi, line 121:
```suggestion
  This option is only effective for CUDA 11.1 or later; CuPy always compiles into PTX on earlier CUDA versions. Also, this option only applies when NVRTC is selected as the compilation backend. NVCC backend always compiles into SASS (CUBIN).
```

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):
LGTM!

leofang:
Thanks @emcastillo!

Issue #5557, opened 2021-07-24
import cupy eats ~1.8Gb of virtual memory for nothing in every process instance


iperov:
My app (https://github.com/iperov/DeepFaceLive) uses multiple python subprocesses (multiprocessing.Process) to do the work.
Every process which imports cupy uses 1.8Gb+ physical memory or paging file. 

```import cupy```

https://user-images.githubusercontent.com/8076202/126863232-90bd879a-dedc-48a6-94c9-bfc8a3e3ee7f.mp4

If user has no free space in paging file, MemoryError will occur.

Can you fix it?

Temporary solution is to import cupy on demand, but the code looks like crap,
and unable to use type hinting like ```img : cp.ndarray```, because it requires cupy to be imported.

```
>>> cupy.show_config()
OS                           : Windows-10-10.0.19041-SP0
CuPy Version                 : 9.0.0
NumPy Version                : 1.21.1
SciPy Version                : 1.5.4
Cython Build Version         : 0.29.23
Cython Runtime Version       : None
CUDA Root                    : D:\DevelopPPP\projects\DeepFaceLive\_internal\CUDA
CUDA Build Version           : 11010
CUDA Driver Version          : 11030
CUDA Runtime Version         : 11010
cuBLAS Version               : 11201
cuFFT Version                : 10300
cuRAND Version               : 10202
cuSOLVER Version             : (11, 0, 0)
cuSPARSE Version             : 11200
NVRTC Version                : (11, 1)
Thrust Version               : 100910
CUB Build Version            : 100910
Jitify Build Version         : 60e9e72
cuDNN Build Version          : 8101
cuDNN Version                : 8005
NCCL Build Version           : None
NCCL Runtime Version         : None
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA GeForce RTX 3090
Device 0 Compute Capability  : 86
Device 0 PCI Bus ID          : 0000:01:00.0
```

kmaehashi:
I think this issue is not specific to CuPy but may happen on many libraries using CUDA on Windows, because `multiprocessing` on Windows cannot use `fork` (copy-on-write). Please consider using `threading` if it works for your use case.

> cuDNN Version                : 8005

CuPy automatically loads cuDNN if it is installed. You may try uninstalling cuDNN to reduce the memory footprint (a bit).

iperov:
@kmaehashi

in reproduce example I don't use multiprocessing. 

[...]

PR #1781, opened 2018-11-07
[WIP] Avoid cimport-ing driver

kmaehashi:
This fix is to support environment which doesn't have CUDA Driver installed.
It is better to avoid depending on CUDA driver shared library during `import cupy` (c.f. https://github.com/cupy/cupy/issues/1073#issuecomment-436229508, #1790).

This makes possible to import CuPy without try-except block in environment without driver, e.g. in the following environment:

- Windows environment without CUDA GPU
- Linux Docker environment running without `nvidia-docker`

(Note: in both case CUDA Runtime must be installed separately unless using anaconda cupy package)

TODOs:

- [ ] avoid `except ImportError: pass`
- [ ] avoid `cimport driver_types as driver`
- [ ] avoid `cimport *`
- [ ] support cupy.show_config() to work without driver

This PR depends on #1780.

PR #5638, opened 2021-08-12
Add support for CUDA Python

kmaehashi:
This PR rewrites the CuPy backend using [CUDA Python](https://github.com/NVIDIA/cuda-python). You will need to opt-in by `export CUPY_USE_CUDA_PYTHON=1` to build CuPy using CUDA Python. By adopting CUDA Python, we can avoid directly linking to CUDA libraries at build time.

Benefits of adopting CUDA Python are:

* Build-time link to CUDA Driver (`libcuda.so`) can be removed (c.f. https://github.com/cupy/cupy/pull/1781).
* CUDA libraries can be lazy-loaded: faster and smaller memory footprint when doing `import cupy` (c.f. https://github.com/cupy/cupy/issues/5557).
* When all CUDA Toolkit libraries are supported in CUDA Python:
    * The number of wheels to release can be reduced by utilizing CUDA Forward Compatibility, i.e., one release for each CUDA major version.
    * No longer need to maintain stub headers (which is currently needed to absorb CUDA version difference and build CuPy on RTD without CUDA).
    * CuPy can be imported without CUDA.

Note that the following components call CUDA Toolkit APIs from C/C++ layer (which bypasses CUDA Python):

* `cupy._core.cub` (via CUB header): Runtime
* `cupy._core.jitify` (via Jitify header): Runtime
* `cupy.cuda.thrust` (via Thrust header): Runtime
* `cupy.random._generator_api` (via cuRAND header): Runtime, cuRAND
* `cupy_backends.cuda.libs.profiler`: Runtime
    * This module calls `cudaProfilerInitialize` but is unsupported by CUDA Python, as the API is a deprecated one

We are planning to build CuPy v10 wheels using CUDA Python once binary distributions of CUDA Python become available.

cc/ @mmccarty @vzhurba01


kmaehashi:
pfnCI, test this please.

cupy_backends/cuda/api/_driver_typedef.pxi
```diff
@@ -0,0 +1,31 @@
...
+    # For Texture Reference
+    ctypedef void* Array 'CUarray_st*'  # = cupy.cuda.runtime.Array
+    ctypedef int Array_format 'CUarray_format'
+    ctypedef struct Array_desc 'CUDA_ARRAY_DESCRIPTOR':
+        Array_format Format
+        size_t Height
+        unsigned int NumChannels
+        size_t Width
+    ctypedef int Address_mode 'CUaddress_mode'
+    ctypedef int Filter_mode 'CUfilter_mode'
```
leofang, line 31: Note to self: Remove the texture reference support as it's deprecated and unsupported by CUDA Python.

cupy_backends/cuda/api/driver.pyx
```diff
@@ -15,89 +15,23 @@ cimport cython  # NOQA
...
-        block2shmem, size_t dynamicSMemSize, int blockSizeLimit)
+IF CUPY_USE_CUDA_PYTHON:
+    from cuda.ccuda cimport *
+ELSE:
+    include '_driver_extern.pxi'
 
+cdef extern from '../../cupy_backend.h' nogil:
     # Build-time version
-    int CUDA_VERSION
     int HIP_VERSION
```
leofang, line 30: Can we leave a comment here? Something like
```suggestion
    # Note: CUDA_VERSION is defined either in CUDA Python or _driver_extern.pxi
    int HIP_VERSION
```

cupy_backends/cuda/api/driver.pyx
```diff
@@ -15,89 +15,23 @@ cimport cython  # NOQA
...
 
+# Provide access to constants from Python.
+from cupy_backends.cuda.api._driver_enum import *
```
leofang, line 33: In the case of CUDA Python, can't we utilize the enums exposed to Python from `cudapython/cuda.pyx`? Though the current approach is probably a lot easier to manage.

cupy_backends/cuda/api/driver.pyx
```diff
@@ -148,6 +82,10 @@ cpdef get_build_version():
         return 0
 
 
+cpdef bint _is_cuda_python():
+    return CUPY_USE_CUDA_PYTHON
```
leofang, line 86: Just a thought, haven't figured it out completely: Maybe this should live in the "Constants (CuPy)" section in `runtime.pyx`, sitting side by side with `_is_hip_environment`?

cupyx/_runtime.py
```diff
@@ -149,7 +149,11 @@ def __init__(self, *, full=True):
         else:
             self.nvcc_path = cupy._environment.get_hipcc_path()
 
-        self.cuda_build_version = cupy.cuda.driver.get_build_version()
+        driver_build_version = cupy.cuda.driver.get_build_version()
+        if cupy.cuda.driver._is_cuda_python():
```
leofang, line 153: Ah I see `_is_cuda_python()` is used here.

leofang (dismissed):
LGTM except for a few questions! This is great work @kmaehashi!

cupy_backends/cuda/api/_driver_typedef.pxi
kmaehashi: It's deprecated but seems `CUtexref` exists in CUDA Python. (maybe I'm missing something?)

cupy_backends/cuda/api/driver.pyx
kmaehashi: Thanks, applied!

cupy_backends/cuda/api/driver.pyx
kmaehashi: Thanks, applied!

cupy_backends/cuda/api/driver.pyx
kmaehashi: That's possible but we have to maintain `_driver_enum` for HIP anyways.

cupy_backends/cuda/api/driver.pyx
kmaehashi: I'm thinking it's better to place this here considering `get_build_version()` returns the version of CUDA Python.

kmaehashi:
pfnCI, test this please.

cupy_backends/cuda/api/_driver_typedef.pxi
leofang: nvm I read it wrong, sorry!

kmaehashi:
Added support CI with CUDA Python.
https://ci.preferred.jp/cupy.linux.cuda11x-cuda-python/77003/

kmaehashi:
pfnCI, test this please.

kmaehashi:
All tests passed :tada:

mmccarty:
Thanks everyone! I'm so excited to see the adoption of CUDA Python in CuPy!

emcastillo (approved):
LGTM!
I don't have concerns with this PR, I think it is better to merge it now and address all the issues that may happen incrementally

PR #5642, opened 2021-08-13
Fix squareness checks

toslunar:
Fix bugs caused by the fact that `_assert_nd_squareness` does not assert the inputs are at least 2d.

```
>>> for xp in (numpy, cupy):
...  try: xp.linalg.cholesky(xp.ones(()))
...  except Exception: traceback.print_exc(limit=0)
...
numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
ValueError: max() arg is an empty sequence
>>> for xp in (numpy, cupy):
...  try: xp.linalg.eigh(xp.ones(()))
...  except Exception: traceback.print_exc(limit=0)
...
numpy.linalg.LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
ValueError: Array must be at least two-dimensional
```

`scipy.sparse.csgraph`, on the other hand, should not raise `LinAlgError`.
```
>>> scipy.sparse.csgraph.connected_components(numpy.ones((3, 2)))
ValueError: graph should be a square array
```

This PR also fixes the names of the assertions (see numpy/numpy#14814).

### DIFF
--- a/cupy/linalg/_util.py
+++ b/cupy/linalg/_util.py
@@ -12,6 +12,8 @@ import cupyx
 _default_precision = os.getenv('CUPY_DEFAULT_PRECISION')
 
 
+# The helper functions raise LinAlgError if the conditions are not met.
+
 def _assert_cupy_array(*arrays):
     for a in arrays:
         if not isinstance(a, cupy._core.ndarray):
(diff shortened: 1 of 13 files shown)


cupy/linalg/_util.py
```diff
@@ -19,17 +19,26 @@ def _assert_cupy_array(*arrays):
...
 
 
-def _assert_nd_squareness(*arrays):
+def _assert_stacked_2d(*arrays):
     for a in arrays:
-        if max(a.shape[-2:]) != min(a.shape[-2:]):
+        if a.ndim != 2:
+            raise linalg.LinAlgError(
+                '{}-dimensional array given. Array must be '
+                'at least two-dimensional'.format(a.ndim))
```
leofang, line 35: `_assert_stacked_2d()` doesn't make sense to me? It rules out the possibility of having a stack of 2D matrices.

toslunar: Oops 😳

emcastillo:
Jenkins, test this please

emcastillo (approved):

PR #5599, opened 2021-08-04
Stop inheriting `unittest.TestCase` for performance

toslunar:
`cd tests; for i in *test*/*test*; do echo $i; pytest -q --collect-only $i | tail -n 1; done`

||before|after|
|-|-|-|
|cupy_tests/binary_tests|9 tests collected in 0.70s|9 tests collected in 0.75s|
|cupy_tests/core_tests|2357 tests collected in 13.82s|2357 tests collected in 8.13s|
|cupy_tests/creation_tests|415 tests collected in 2.07s|415 tests collected in 1.86s|
|cupy_tests/cuda_tests|834 tests collected in 5.95s|834 tests collected in 1.90s|
|cupy_tests/fft_tests|2711 tests collected in 1.90s|2711 tests collected in 1.87s|
|cupy_tests/functional_tests|67 tests collected in 0.75s|67 tests collected in 0.72s|
|cupy_tests/indexing_tests|384 tests collected in 2.44s|384 tests collected in 2.30s|
|cupy_tests/io_tests|37 tests collected in 0.67s|37 tests collected in 0.68s|
|cupy_tests/lib_tests|484 tests collected in 7.79s|484 tests collected in 1.41s|
|cupy_tests/linalg_tests|1327 tests collected in 8.20s|1327 tests collected in 3.74s|
|cupy_tests/logic_tests|788 tests collected in 2.40s|788 tests collected in 0.89s|
|cupy_tests/manipulation_tests|479 tests collected in 1.39s|479 tests collected in 1.43s|
|cupy_tests/math_tests|9445 tests collected in 11.24s|9445 tests collected in 5.37s|
|cupy_tests/misc_tests|22 tests collected in 0.66s|22 tests collected in 0.66s|
|cupy_tests/padding_tests|75 tests collected in 1.45s|75 tests collected in 1.29s|
|cupy_tests/polynomial_tests|55 tests collected in 0.74s|55 tests collected in 0.74s|
|cupy_tests/prof_tests|8 tests collected in 0.72s|8 tests collected in 0.69s|
|cupy_tests/random_tests|1880 tests collected in 22.85s|1880 tests collected in 6.68s|
|cupy_tests/sorting_tests|412 tests collected in 3.45s|412 tests collected in 3.48s|
|cupy_tests/statistics_tests|662 tests collected in 7.40s|662 tests collected in 1.37s|
|cupy_tests/test_cublas.py|2436 tests collected in 12.74s|2436 tests collected in 0.94s|
[...]

PR #5659, opened 2021-08-18
Eliminate random test ids

### LINKED ISSUE #5623: Avoid using random in test parameterization value
Currently, the following tests are using random values in test parameterization. I'd like to remove use of random values so that we can use pytest-xdist to run tests in parallel.

```
tests/cupy_tests/core_tests/test_ndarray_adv_indexing.py
tests/cupy_tests/core_tests/test_ndarray_scatter.py
```

When using pytest-xdist, each test process collects tests. The list of collected tests must match between all processes, but if random values are used, pytest-xdist fails with `Different tests were collected between ...` error.

toslunar:
`pytest -n 2 tests` looks OK with pytest-xdist 2.3.0

Let's close #5623, while the parameters are still randomly generated in the test collection.

Depends #5599 because
- the new option `_ids=False` is implemented only for pytest-backend `cupy.testing.parameterize`; and
- the test collection should be fast to benefit from pytest-xdist.


cupy/testing/_parameterized.py
```diff
@@ -1,19 +1,22 @@
...
+
+
 def _param_to_str(obj):
     if isinstance(obj, type):
         return obj.__name__
     elif hasattr(obj, '__name__') and isinstance(obj.__name__, str):
         # print __name__ attribute for classes, functions and modules
         return obj.__name__
-    return repr(obj)
+    return _bad_name.sub('', repr(obj))
```
toslunar, line 19: The other source of randomness is `repr` of param.  Examples:
```diff
-tests/cupyx_tests/fallback_mode_tests/test_fallback.py::TestDocs_param_3_{object=<numpy = <function eig at 0x7f197d6d8ca0>, cupy = None>}::test_docs
+tests/cupyx_tests/fallback_mode_tests/test_fallback.py::TestDocs_param_3_{object=<numpy = <function eig at 0x7f7d606f3ca0>, cupy = None>}::test_docs

-tests/cupyx_tests/fallback_mode_tests/test_notifications.py::TestNotificationModes_param_0_{func=<numpy = <function array_equiv at 0x7f197de3c700>, cupy = None>, shape=(3, 4)}::test_notification_ignore
+tests/cupyx_tests/fallback_mode_tests/test_notifications.py::TestNotificationModes_param_0_{func=<numpy = <function array_equiv at 0x7f7d680b0700>, cupy = None>, shape=(3, 4)}::test_notification_ignore

-tests/cupyx_tests/scipy_tests/ndimage_tests/test_filters.py::TestGenericFilter::test_filter[_param_0_{dtype=uint8, filter='generic_filter', footprint=False, func_or_kernel=(<cupy._core.raw.RawKernel object at 0x7f188c83ff40>, <function rms_pyfunc at 0x7f188c7d65e0>), ksize=3, mode='reflect', origin=0, output=None, shape=(4, 5)}]
+tests/cupyx_tests/scipy_tests/ndimage_tests/test_filters.py::TestGenericFilter::test_filter[_param_0_{dtype=uint8, filter='generic_filter', footprint=False, func_or_kernel=(<cupy._core.raw.RawKernel object at 0x7f7c6f881f40>, <function rms_pyfunc at 0x7f7c6f8175e0>), ksize=3, mode='reflect', origin=0, output=None, shape=(4, 5)}]
```

Maybe the 3 test files should be fixed instead.

kmaehashi: I guess this may generate duplicate test names when lambda is parameterized (e.g., `<function <lambda> at 0x7fba03090200>`) so I think fixing tests is better. Let me take care of it.

kmaehashi: #5727 #5728

kmaehashi:
Thanks! LGTM except for `_bad_name` as commented.

.pfnci/linux/tests/_build_and_test.sh
```diff
@@ -9,13 +9,14 @@ pytest_opts=(
     --timeout 300
     --maxfail 500
     --showlocals
+    --numprocesses 2
```
kmaehashi, line 12: Could you rotate `CUDA_VISIBLE_DEVICES` for each worker process, like we do in chainer-test? This should prevent OOM on GPU and improve performance.

https://github.com/pytest-dev/pytest-xdist/issues/179#issuecomment-312419876

maybe something like this:

```import collections; q = collections.deque(range(2)); q.rotate(int(1)); print(",".join(map(str, q)))```

Max number of GPUs are given as environment variable "${CUPY_TEST_GPU_LIMIT}`

Comment (toslunar) [reply .pfnci/linux/tests/_build_and_test.sh:12]: I'll try `pytest_configure_node(node)` hook.
https://github.com/pytest-dev/pytest-xdist/blob/766e67ce526c88327670e9bcd4804c335df51015/src/xdist/newhooks.py#L38-L39

Comment (kmaehashi) [reply .pfnci/linux/tests/_build_and_test.sh:12]: I'm unsure if setting `CUDA_VISIBLE_DEVICES` after loading CUDA shared libraries will work, but seems worth trying.

Comment (kmaehashi) [reply .pfnci/linux/tests/_build_and_test.sh:12]: I think overriding `CUDA_VISIBLE_DEVICES` in test code is not always a good idea, because I sometimes want to run tests on different devices by manually setting `CUDA_VISIBLE_DEVICES`.

File: tests/conftest.py [hunk 3]
```diff
@@ -27,3 +30,37 @@ def pytest_configure(config):
...
+class DeferPlugin:
+    """Simple plugin to defer pytest-xdist hook functions."""
+
+    # Edit the environment variable `CUDA_VISIBLE_DEVICES` for each session.
+    # Cannot use `pytest_configure_node` nor `pytest_testnodeready` hook,
+    # because they are called in the `master` node (process).
+    # See also https://github.com/pytest-dev/pytest-xdist/issues/179.
+    @pytest.fixture(autouse=True, scope='session')
+    def _rotate_cuda_visible_devices(self, worker_id):
+        n_gpu = os.environ.get('CUPY_TEST_GPU_LIMIT')
```
kmaehashi, line 47: Can we ensure that CUDA is not loaded (so that we can overwrite `CUDA_VISIBLE_DEVICES`)? Maybe `assert 'cupy' not in sys.modules` can be used.

toslunar: I tried
```python
        assert not {
            'cupy', 'cupy_backends', 'cupyx',
        }.intersection(sys.modules)
```
in this function, but it failed.  However, `CUDA_VISIBLE_DEVICES` looks working to me because `cupy.cuda.Device(0).attributes["PciBusId"]` returned different values.

I didn't come up with an idea for testing this in a portable manner.

tests/conftest.py
```diff
@@ -27,3 +30,37 @@ def pytest_configure(config):
...
+        if n_gpu is None:
+            return
+        n_gpu = int(n_gpu)
+
+        assert worker_id.startswith('gw')
```
kmaehashi, line 52: `worker_id` can be `master` if pytest-xdist is installed and run without `-n` option.

tests/conftest.py
```diff
@@ -27,3 +30,37 @@ def pytest_configure(config):
...
+
+    # Edit the environment variable `CUDA_VISIBLE_DEVICES` for each session.
+    # Cannot use `pytest_configure_node` nor `pytest_testnodeready` hook,
+    # because they are called in the `master` node (process).
+    # See also https://github.com/pytest-dev/pytest-xdist/issues/179.
+    @pytest.fixture(autouse=True, scope='session')
+    def _rotate_cuda_visible_devices(self, worker_id):
+        n_gpu = os.environ.get('CUPY_TEST_GPU_LIMIT')
+        if n_gpu is None:
+            return
```
kmaehashi, line 49:
```suggestion
            print('Tip: when using pytest-xdist, you can automatically rotate'
                  ' CUDA_VISIBLE_DEVICES for each test worker by setting'
                  ' CUPY_TEST_GPU_LIMIT environment variable.')
            return
```

.pfnci/linux/tests/_build_and_test.sh
kmaehashi: I now think overriding `CUDA_VISIBLE_DEVICES` in test code makes sense as developers may want to use pytest-xdist interactively. 😄

tests/conftest.py
kmaehashi: I see, thanks for testing!

kmaehashi:
pfnCI, test this please.

PR #282, opened 2017-07-11
Refactor argsort tests.

takagi:
This PR make `argsort` tests clearer in terms of parameterizing tests for `cupy.argsort` and `cupy.ndarray.argsort` before supporting sort multi rank arrays.

PR #5688, opened 2021-08-27
expm implementation for dense matrices

MrRobot2211:
This will implement the equivalent to `scipy.linalg.expm` for dense matrices.   

Implemetationss worth taking into account, both [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/linalg/expm) and [PyTorch](https://pytorch.org/docs/stable/generated/torch.matrix_exp.html) already have implementations. The former's relies on scaling and squaring Padé's aproximation while the latter uses an optimized polynomial taylor sum approximation.

There is also [this](https://github.com/pd93/CUDAMatrixExp) implementation, which is also based on Pade's aproximation.

And this reference from the implementation on [Mathematica](https://www.mathworks.com/matlabcentral/fileexchange/67668-batch-matrix-exponential-computation-with-cpu-gpu), which hints at some optimizations for hermitian matrices.

This will tackle #282, and help with https://github.com/qutip/qutip-cupy/issues/40 .


kmaehashi:
Thank you for working on this! I think `cupyx.scipy.linalg.expm` is the right place to put this function as `numpy.linalg.expm` does not exist 😄

https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.expm.html

SciPy's implementation supports both dense and sparse but it's ok to start with the dense-only implementation.

MrRobot2211:
> Thank you for working on this! I think `cupyx.scipy.linalg.expm` is the right place to put this function as `numpy.linalg.expm` does not exist 
> 
> https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.expm.html
> 
> SciPy's implementation supports both dense and sparse but it's ok to start with the dense-only implementation.

Thank you for the catch , I am editing the description and changing the locationof the files.

.gitignore
```diff
@@ -18,6 +18,7 @@ cupy.egg-info/
 dist/
 htmlcov/
 .idea/
+.vscode/
```
takagi, line 21: We do not have `.gitignore` setting for development environments. Would you remove this change from the PR? Instead, you can use `~/.config/git/ignore` to ignore vscode files in all situations.

leofang:
Hi @MrRobot2211 Any chance you can push your WIP so that I can ask someone to take over and help move it pass the finish line? As you know this is an important API for quantum computing and quantum physics, so let's not drop the ball 🙂 Thanks!

kmaehashi:
Kindly ping @MrRobot2211 😃

leofang:
So I think this repo (https://github.com/Lezcano/expm) was how PyTorch used to implement `expm`. It implements both GPU-friendly approaches (see its README and the linked papers). We could port it to CuPy and see how the performance goes, and decide if we need any dedicated CUDA kernel. Could be a good mini project 🙂

ev-br:
https://github.com/cupy/cupy/pull/7660 has an implementation based on Pade 13/13 approximation of Higham'2005

leofang:
Thanks, @ev-br, it's very encouraging to know! I suggest we close this PR due to inactivity. Once #7660 is in, we can revisit `expm()` anytime if the Taylor-expansion based approach is needed.

ev-br:
If/when this PR is closed, I'd suggest to open an enhancement issue with links to other implementations, especially Lezcano's one (https://github.com/cupy/cupy/pull/5688#issuecomment-1552067121). As a general purpose expm for potentially large matrices, it is probably way more efficient than gh-7660's version. And as a yet further enhancement, https://github.com/higham/expmv has a Taylor-expansion based expm times vector algorithm.

leofang:
@takagi shall we close this PR and follow up in #7769?

takagi:
Yes!

PR #5673, opened 2021-08-23
Fix spline filter with large array

takagi:
Close #5537. This PR fixes the spline filter with large array.

PR #5693, opened 2021-08-31
Improve performance of `TestSplineFilter1dLargeArray`

toslunar:
https://github.com/cupy/cupy/pull/5673#discussion_r698180409


takagi:
pfnCI, test this please.

takagi:
Now that I observed this test ran just in a few seconds on my local, how about removing `@testing.slow`? Very neat that the OS just maps memory address for `empty`.

takagi (approved):

PR #5698, opened 2021-09-05
Adopt the `numpy.array_api` module as `cupy.array_api`

### LINKED ISSUE #4789: Adopt Python Array API standard

leofang:
~~Very experimental attempt to see how far we can reach. **DO NOT MERGE.**~~

Close #4789.

This PR has two components:
1. It brings in the PR https://github.com/numpy/numpy/pull/18585 sent to NumPy by @asmeurer. It's mostly done as is -- check out the commit history. The only changes I made to fix the `git am` failures were those related to NumPy specifics, such as `setup.py` and `_pytesttester.py`; the relevant diffs were carefully removed from the patch to enable a simple `git am 18585.patch` call.
2. It modifies the module to tailor for CuPy --- ongoing, see the latest progress: https://github.com/leofang/cupy/compare/fb93033..d3a6f0d

cc: @rgommers @asmeurer (feel free to tag more interested people)


kmaehashi:
Thank you for working on this Leo! I was looking around the array API code in NumPy this weekend and thinking it can be as simple as replacing `import numpy as np` with `import cupy as cp`, but you did that blazing fast 😄 

Just a question: do you prefer to put them as `cupyx.array_api` over `cupy.array_api`?

leofang:
> and thinking it can be as simple as replacing `import numpy as np` with `import cupy as cp`, but you did that blazing fast 😄

No problem. The implementation status happened to be discussed in this week's meeting, so I decided to take a very quick look; after the weekend if you feel this PR needs to be expedited, someone may have to take over...

My current understanding is `numpy/np`->`cupy/cp` works to some extent (`tests/cupyx_tests/array_api_tests` already passed now), but a few things still need to be worked on:
- device support was completely left out in the NumPy implementation 
- scalar handling was unnecessarily complicated (due to NumPy's notorious scalar/array distinction)
- documentation and tests

> Just a question: do you prefer to put them as `cupyx.array_api` over `cupy.array_api`?

I put the code under `cupyx/` just because it's the convention where we add the non-NumPy compliant stuff, but if you feel strongly in either way I can certainly change. We can also import all Array API functions in `cupy/array_api/__init__.py` but still keep the implementation in `cupyx/array_api/`. Just let me know whatever you decide 🙂

leofang:
> but a few things still need to be worked on

As I spoke I found a bug in https://github.com/numpy/numpy/pull/18585...

leofang:
OK it's a good progress for today, will resume when I have time (?)
```
git clone https://github.com/data-apis/array-api-tests.git
ARRAY_API_TESTS_MODULE=cupyx.array_api pytest ./array-api-tests/
```

Outputs:
```
FAILED ../array-api-tests/array_api_tests/test_creation_functions.py::test_eye - OverflowError: Python int too large to convert to C ssize_t
FAILED ../array-api-tests/array_api_tests/test_creation_functions.py::test_linspace - hypothesis.errors.MultipleFailures: Hypothesis found 2...
FAILED ../array-api-tests/array_api_tests/test_indexing.py::test_multiaxis_indexing - hypothesis.errors.MultipleFailures: Hypothesis found 3...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_has_names[einsum] - AssertionError: cupyx.array_api is missing the linalg...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_has_names[linalg] - AttributeError: 'NoneType' object has no attribute 'r...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[__dlpack__] - AssertionError: Unexpected excepti...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[__dlpack_device__] - AssertionError: Unexpected ...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[from_dlpack] - AssertionError: Unexpected except...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[finfo] - AssertionError: Unexpected exception At...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[positive] - AssertionError: Unexpected exception...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[concat] - AssertionError: Unexpected exception T...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[argsort] - AssertionError: Unexpected exception ...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_positional_args[sort] - AssertionError: Unexpected exception Typ...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_keyword_only_args[__dlpack__] - RuntimeError: Don't know how to ...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_keyword_only_args[concat] - AssertionError: Unexpected exception...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_keyword_only_args[argsort] - AssertionError: Unexpected exceptio...
FAILED ../array-api-tests/array_api_tests/test_signatures.py::test_function_keyword_only_args[sort] - AssertionError: Unexpected exception T...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-f4] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-f8] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-i1] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-i2] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-i4] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-i8] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-u1] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-u2] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-u4] - AttributeE...
FAILED ../array-api-tests/array_api_tests/test_type_promotion.py::test_elementwise_function_one_arg_type_promotion[positive-u8] - AttributeE...
FAILED ../array-api-tests/array_api_tests/special_cases/test_logaddexp.py::test_logaddexp_special_cases_two_args_equal__notequal - Assertion...
===================================== 28 failed, 3361 passed, 24 skipped, 1 warning in 1093.47s (0:18:13) =====================================
```

leofang:
For all failures in `test_signatures.py`, I've identified the causes: 
- `einsum` is not part of Array API v1 => the test suite to be fixed
- `linalg` is not yet finalized => TODO
- DLPack support was done incorrectly => fixed in ce6d809
- `finfo.smallest` didn't consider backward compatibility +> fixed in ce6d809
- Sorting functions lack full NumPy compliance (#5701)
- `positive()` seems redundant to me => raised a discussion (https://github.com/data-apis/array-api/issues/255)
- `concat()` lacks full NumPy compliance (#5700)

@asmeurer You may want to take ce6d809 if you plan to send another round of NEP-47 changes to NumPy 🙂

asmeurer:
The dlpack support is not in NumPy yet, so it will need to be fixed once https://github.com/numpy/numpy/pull/19083 is merged. Let me know if you run into any issues with the test suite (preferably by opening an issue at https://github.com/data-apis/array-api-tests). Also note that some tests in the test suite are XFAIL for NumPy right now (see https://github.com/data-apis/array-api-tests/blob/master/.github/workflows/numpy.yml). You should also be aware that there are still some gaps in the test suite. Many functions are not tested at all yet, except for basic signature tests. However once this is merged I'll keep an eye on cupy compliance as I update the test suite so I can be aware of any issues. Is it possible to run cupy on CI, or will I need to do this manually (I do have a Nvidia GPU I can manually run cupy with)?

leofang:
Thanks for suggestions, @asmeurer!
> Is it possible to run cupy on CI, or will I need to do this manually (I do have a Nvidia GPU I can manually run cupy with)?

Whose CI are you referring to? 😅 I see that the `array-api-tests` repo uses GH runners, so it probably can't run GPU tests?

For CuPy's CI it's up to @kmaehashi to decide. We can do either of the following:
- Include the Array API tests as part of the normal test workflow: If we modify setup.py such that `pip install cupy[array-api]` pull down the latest `array-api-tests`, it should work.
- Have a dedicated CI runner (in either Jenkins or pfnCI) that only tests this support.

In any case, we should mark such tests as non-blocking (like the ROCm tests) as it's highly experimental.

cupyx/array_api/_typing.py
```diff
@@ -0,0 +1,47 @@
...
+    "Dtype",
+    "SupportsDLPack",
+    "SupportsBufferProtocol",
+    "PyCapsule",
+]
+
+from typing import Any, Sequence, Type, Union
+
+from . import (
+    Array,
```
leofang, line 24: Hi @asmeurer I am confused by these two lines. `Array` does not live in either `numpy.arrray_api` or `numpy.arrray_api._typing`, does it? So why does this import statement work?

asmeurer:
> Whose CI are you referring to? 😅 I see that the array-api-tests repo uses GH runners, so it probably can't run GPU tests?

Either one is fine with me. If it's not straightforward to run cupy on a free CI then I guess it should be done on the cupy CI.

cupyx/array_api/_typing.py
asmeurer: It looks like it doesn't work actually. I guess this file is never explicitly imported anywhere. 

```
>>> import numpy.array_api._typing
  File "./numpy/array_api/_typing.py", line 20, in <module>
    from . import (
ImportError: cannot import name 'Array' from 'numpy.array_api' (./numpy/array_api/__init__.py)
```

leofang: hmmm... @asmeurer then it looks like the NumPy code would not work with any static type analyzer?
```python
>>> import typing
>>> typing.TYPE_CHECKING
False
>>> typing.TYPE_CHECKING = True
>>> from cupyx import array_api
<stdin>:1: UserWarning: The numpy.array_api submodule is still experimental. See NEP 47.
  File "/home/leof/dev/cupy_cuda114/cupyx/array_api/__init__.py", line 133, in <module>
    from ._creation_functions import (
  File "/home/leof/dev/cupy_cuda114/cupyx/array_api/_creation_functions.py", line 7, in <module>
    from ._typing import (
  File "/home/leof/dev/cupy_cuda114/cupyx/array_api/_typing.py", line 23, in <module>
    from . import (
ImportError: cannot import name 'Array' from partially initialized module 'cupyx.array_api' (most likely due to a circular import) (/home/leof/dev/cupy_cuda114/cupyx/array_api/__init__.py)
```

leofang:
> > Whose CI are you referring to? 😅 I see that the array-api-tests repo uses GH runners, so it probably can't run GPU tests?
> 
> Either one is fine with me. If it's not straightforward to run cupy on a free CI then I guess it should be done on the cupy CI.

Free GPU CI is available, just not from GH, Azure, etc 😆 @aktech's cirun.io has been used by QuTiP folks to set up their CuPy tests (https://github.com/qutip/qutip-cupy/pull/9) so it's definitely doable. Maybe you'd be interested to set it up in `array-api-tests` for testing other GPU libraries (like PyTorch) once they catch up? Let's wait for @kmaehashi to chime in first.

asmeurer:
Let's do whatever makes the most sense. If you have the CI on CuPy, is it possible to set it up so that I get notified of any failures, in case they end up being due to a bug in the test suite on my part?

leofang:
> is it possible to set it up so that I get notified of any failures, in case they end up being due to a bug in the test suite on my part?

@asmeurer CuPy has set a simple bot that would ping specific persons if an issue/PR is tagged: https://github.com/cupy/cupy/blob/master/.github/workflows/auto-cc.yml. In this case, it means the [array-api](https://github.com/cupy/cupy/labels/array-api) label. Would this be acceptable to you? If not, I can always ping you manually 🙂

kmaehashi:
Adding a test axis for array api tests in CuPy is great, it should be as simple as `build cupy; git clone https://github.com/data-apis/array-api-tests; ARRAY_API_TESTS_MODULE=cupy pytest`. If the test failure is due to the test suite we can just create an issue.
Once array API support is merged to CuPy, it also makes sense to setup CI in array-api side so you can test pull-requests of array-api-tests suite against CuPy.

> Free GPU CI is available, just not from GH, Azure, etc 😆 @aktech's cirun.io has been used by QuTiP folks to set up their CuPy tests (qutip/qutip-cupy#9) so it's definitely doable.

My understanding is that cirun.io itself is free, but to launch a self-hosted test runner AWS/GCP/Cloudocean account is needed which is pay-as-you-go?

leofang:
> but to launch a self-hosted test runner AWS/GCP/Cloudocean account is needed which is pay-as-you-go?

Right, I keep forgetting... 🤦🏻‍♂️

leofang:
OK I've identified all the issues that led to the test failures (https://github.com/cupy/cupy/pull/5698#issuecomment-913406882). Note: NumPy and CuPy suffer from the same linespace issue (https://github.com/numpy/numpy/issues/18881), so unless NumPy has a fix out we should just skip the test as Aaron suggested.

Tomorrow, I'll add some simple docs explaining the current status, pointing users to the API spec, etc.

@kmaehashi let me know if there is any test related change I should address; otherwise, feel free to take a look anytime 🙂

asmeurer:
Great. Yes you can add me to the bot for the array-api label. 

> Adding a test axis for array api tests in CuPy is great, it should be as simple as `build cupy; git clone https://github.com/data-apis/array-api-tests; ARRAY_API_TESTS_MODULE=cupy pytest`

It should be `ARRAY_API_TESTS_MODULE=cupy.array_api`.

> OK I've identified all the issues that led to the test failures (#5698 (comment)). Note: NumPy and CuPy suffer from the same linespace issue (numpy/numpy#18881), so unless NumPy has a fix out we should just skip the test as Aaron suggested.

I saw some of the cross-referenced issues. What is the test_eye error? I can't tell from the summary what the issue is (an unfortunate feature of the test suite, which I don't know how to improve), but from the exception it's possible I have the wrong strategy input to the hypothesis test.

leofang:
> I saw some of the cross-referenced issues. What is the test_eye error? I can't tell from the summary what the issue is (an unfortunate feature of the test suite, which I don't know how to improve), but from the exception it's possible I have the wrong strategy input to the hypothesis test.

@asmeurer all failures I discovered are CuPy's fault. I haven't found any misbehavior from the test suite, though I haven't dug hard either 🙂

kmaehashi:
The PR overall looks good to me. :+1:

> I put the code under `cupyx/` just because it's the convention where we add the non-NumPy compliant stuff, but if you feel strongly in either way I can certainly change. We can also import all Array API functions in `cupy/array_api/__init__.py` but still keep the implementation in `cupyx/array_api/`. Just let me know whatever you decide 🙂

* As `numpy.array_api` is implemented I think it can also be considered as a part of NumPy-compliant API.

* I'm wondering how we catch-up with the latest NumPy fixes in `numpy.array_api` to `cupy.array_api`, as `numpy.array_api` is likely to be updated frequently. Maybe better to do `import cupy as np` and keep the remaining code using `np` to make diff comparison easier? (I agree this is not the best solution though...)

* All the unittests will fail on Py37 for `cupy.array_api`? In that case module-level `pytest.skip` is needed.

leofang:
@kmaehashi 

> * As `numpy.array_api` is implemented I think it can also be considered as a part of NumPy-compliant API.

OK, I'll move `cupyx.array_api` to `cupy.array_api`. (Change 370c7db, maybe?)

> * I'm wondering how we catch-up with the latest NumPy fixes in `numpy.array_api` to `cupy.array_api`, as `numpy.array_api` is likely to be updated frequently. Maybe better to do `import cupy as np` and keep the remaining code using `np` to make diff comparison easier? (I agree this is not the best solution though...)

I was hoping to make diff easy too. Keeping `np` as is (revert e3104ec) would certainly help, with the understanding that the docstrings would contain `numpy` too. But we can improve them (by diverging from NumPy) once the implementation is considered stable in the future. 

> * All the unittests will fail on Py37 for `cupy.array_api`? In that case module-level `pytest.skip` is needed.

Yes PY37 will fail, but not because of the tests, but because of the use of position-only arguments in the function signatures, which is a PY38+ thing. The module-level skip is already added in `tests/cupyx_tests/array_api_tests/__init__.py` and I also bumped the Python version in the pretest (`.github/workflows/pretest.yml`).

cupyx/array_api/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+- Complex number support in array API spec is planned but not yet finalized,
+  as are the fft extension and certain linear algebra functions such as eig
+  that require complex dtypes.
+
+"""
+
+import warnings
+
+warnings.warn(
+    "The numpy.array_api submodule is still experimental. See NEP 47.", stacklevel=2
```
asmeurer, line 124: Should this be `cupy.array_api`.

asmeurer:
I don't know how much cupy's testing framework diverges from NumPy's, but we never found a good way to make the tests for the numpy.array_api submodule work in Python 3.7. NumPy ended up just dropping 3.7 support before merging my PR, in part because of this issue. Just a heads up that you may run into the same issue.

cupyx/array_api/__init__.py
leofang: We can change this later according to the above discussion.

leofang:
Hi @kmaehashi this is ready. I've moved the module to under `cupy/`, and reverted most of cupy/cp occurrences.

leofang:
Jenkins, test this please

kmaehashi:
Compared diff with https://github.com/numpy/numpy/commit/098f874144161b6a49efa5108846a408ca8f39b8 and looks good to me!

kmaehashi (approved):

PR #5706, opened 2021-09-07
Improve stream mismatch error message

kmaehashi:
I felt that the current error message is a bit difficult to understand what is happening.


kmaehashi:
pfnCI, test this please.

leofang (approved):

takagi (approved):

PR #5707, opened 2021-09-07
Fix import failure when pytest namespace is available (#5703)

### LINKED ISSUE #5703: Do not fail when failed to import `pytest`
As `pytest` is a namespace package, `import pytest` does not mean that pytest is available. Specifically cupy cannot be imported if an empty directory `pytest` exists and pytest is not installed.

kmaehashi:
Closes #5703


kmaehashi:
pfnCI, test this please.

cupy/testing/_helper.py
```diff
@@ -17,17 +17,20 @@
 import cupyx
 import cupyx.scipy.sparse
 
-try:
+from cupy.testing._pytest_impl import is_available
+from cupy.testing._pytest_impl import check_available
```
kmaehashi, line 21:
```suggestion

```

kmaehashi:
pfnCI, test this please.

kmaehashi:
Seems OOM.
pfnCI, test this please.

cupy/testing/_attr.py
```diff
@@ -1,30 +1,12 @@
...
-    import pytest
-    _error = None
-except ImportError as e:
-    _error = e
-
-
-def is_available():
-    return _error is None
-
+from cupy.testing._pytest_impl import is_available, check_available
```
toslunar, line 4: `check_available` is not nice to be imported because of the error message, which starts with `f'{__name__} is not available.'`.

cupy/testing/_pytest_impl.py
```diff
@@ -4,17 +4,20 @@
...
     _error = None
 except ImportError as e:
+    pytest = None
+    _pytest = None
     _error = e
 
 
 def is_available():
-    return _error is None
+    return _error is None and hasattr(pytest, 'fixture')
```
toslunar, line 16: If we allow importing non-modules, it looks better to try-import the least requirements for each file than to import `is_available`.

```python
try:
    import pytest
    from pytest import fixture
except ImportError:
    ...
```

cupy/testing/_attr.py
kmaehashi: Good catch!

cupy/testing/_pytest_impl.py
kmaehashi: By this line I intended to make sure that `pytest` is not an empty namespace. (e.g., `hasattr(pytest, 'mark')` can also be used here)

I think `is_available()` and `check_available()` should behave consistently, so if we don't export `is_available()`, `check_available()` should be implemented in every module. And that what I want to avoid.

toslunar (approved):
LGTM

toslunar:
pfnCI, test this please.

asi1024:
pfnCI, test this please.

asi1024:
Jenkins, test this please.

asi1024 (approved):

PR #5724, opened 2021-09-09
`TestSplineFilter1dLargeArray` marked slow and reduced memory usage

takagi:
Following up #5693.

This PR marks `TestSplineFilter1dLargeArray` slow as well as reducing its memory usage. Now it uses ~2GB of memory.


kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

PR #5725, opened 2021-09-09
Update auto-notify bot for array-api label

kmaehashi:
Adding array-api notification to @asmeurer :tada:
https://github.com/cupy/cupy/pull/5698#issuecomment-915733273

The workflow is tested here: https://github.com/kmaehashi/sandbox/issues/12

### DIFF
--- a/.github/workflows/auto-cc.yml
+++ b/.github/workflows/auto-cc.yml
@@ -10,7 +10,11 @@ jobs:
   auto-notify:
     if: |
         github.repository_owner == 'cupy' &&
-        github.event.label.name == 'hip'
+        contains('|hip|array-api|', github.event.label.name)
+    env:
+      NOTIFY_TARGETS: '{"hip": "@amathews-amd", "array-api": "@asmeurer"}'
+      ISSUE_NUMBER: ${{ (github.event_name == 'issues' && github.event.issue.number) || github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-cc.yml
```diff
@@ -10,23 +10,27 @@ jobs:
   auto-notify:
     if: |
         github.repository_owner == 'cupy' &&
-        github.event.label.name == 'hip'
+        contains('|hip|array-api|', github.event.label.name)
```
kmaehashi, line 13: Unfortunately `env` is not accessible from within `if` condition.

.github/workflows/auto-cc.yml
```diff
@@ -10,23 +10,27 @@ jobs:
...
+    env:
+      NOTIFY_TARGETS: '{"hip": "@amathews-amd", "array-api": "@asmeurer"}'
```
leofang, line 15: Would this work? 😛
```suggestion
      NOTIFY_TARGETS: '{"hip": "@amathews-amd", "array-api": "@asmeurer @leofang"}'
```

kmaehashi: Sure! I thought you are watching everything 😉

kmaehashi:
Let me merge this and check it's working.

PR #5732, opened 2021-09-10
Support cuTENSOR 1.3.3

takagi:


takagi:
- The download URL is changed
- The license file is renamed from `license.txt` to `LICENSE`
- The archive directory name is renamed from `libcutensor` to like `libcutensor-linux-x86_64-1.3.3.2-archive`

kmaehashi:
LGTM. blocked by https://github.com/chainer/chainer-test/pull/685

kmaehashi:
pfnCI, test this please.

kmaehashi:
* According to the test result some URL seems invalid.
* It can be simplified once we drop CUDA 10.1 support.
* Maybe need to test whether `xz` module is available, and fallback to CLI if not available?

takagi:
pfnCI, test this please.

leofang:
> * The download URL is changed

Yes. FWIK this is part of the internal efforts that aims to help package distributors. There is a new json scheme defined for new packages, and an example tool can be found here: https://github.com/NVIDIA/build-system-archive-import-examples. (Since the example repo went public I think it's fine to mention here 🙂) Not sure if it'd help CuPy nor how stable the current scheme is, though.

takagi:
pfnCI, test this please.

kmaehashi:
I have tested with Python env without lzma (xz) support, and fixed the exception type.
c.f. https://github.com/python/cpython/blob/77a1f8d94c29da7bda4bd43e5a9c2233573d46ee/Lib/shutil.py#L1261

kmaehashi:
pfnCI, test this please.

kmaehashi (approved):

PR #5741, opened 2021-09-10
Fix backport trigger

kmaehashi:
Backport should be triggered only when the label added is `to-be-backported`.

This avoids backport to retrigger when labels other than `to-be-backported` is added to a already-closed PR with `to-be-backported` label, e.g., https://github.com/cupy/cupy/pull/5735#issuecomment-916625858


toslunar (approved):
LGTM

PR #5470, opened 2021-07-08
Fix reshape

toslunar:
Fix #5469 and some other issues.

PR #5749, opened 2021-09-13
Fix `MAX_NDIM` and add guards/tests

### LINKED ISSUE #4193: Question: maximum ndim discrepancy between CuPy and NumPy?
I noticed in CuPy we set `MAX_NDIM = 25`:
https://github.com/cupy/cupy/blob/46e5ff0bdf401ad88f404cd7cb7d26dd62b061c0/cupy/core/_carray.pxd#L10
whereas in NumPy this is set to 32 more than a decade ago:
https://github.com/numpy/numpy/blob/9563a3a63fd3dfb3b687fbabad134b5ded72bd46/numpy/core/include/numpy/ndarraytypes.h#L40

Is there a particular reason to pick 25? I was experimenting some iterator stuff and realized I need to know the largest possible `ndim` at build time. I can work around things so this is not critical at all, just simply curious.

### LINKED ISSUE #5721: Segfault when copying a non-contiguous CuPy array to host
```python
>>> import cupy as cp
>>>
>>> ptr = cp.cuda.alloc(2**27 * 8)
>>> a = cp.ndarray(shape=(2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2),\
...                strides=(128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216, 33554432, 67108864, 134217728, 536870912, 268435456, 8, 16, 32, 64),\
...                memptr=ptr)
>>> 
>>> b = cp.asnumpy(a)
Segmentation fault (core dumped)
```

It either segfaults hard, or causes `CUDA_ERROR_MISALIGNED_ADDRESS` in a real workload:
```python
  File "<string>", line 23, in <module>
  File "/home/leof/miniforge3/envs/opt_einsum_dev/lib/python3.8/site-packages/cupy/__init__.py", line 773, in asnumpy
    return a.get(stream=stream, order=order)
  File "cupy/_core/core.pyx", line 1567, in cupy._core.core.ndarray.get
  File "cupy/_core/core.pyx", line 1623, in cupy._core.core.ndarray.get
  File "cupy/_core/core.pyx", line 1625, in cupy._core.core.ndarray.get
  File "cupy/_core/core.pyx", line 2404, in cupy._core.core._internal_ascontiguousarray
  File "cupy/_core/_kernel.pyx", line 1185, in cupy._core._kernel.ufunc.__call__
  File "cupy/_core/_kernel.pyx", line 1212, in cupy._core._kernel.ufunc._get_ufunc_kernel
  File "cupy/_core/_kernel.pyx", line 952, in cupy._core._kernel._get_ufunc_kernel
  File "cupy/_core/_kernel.pyx", line 62, in cupy._core._kernel._get_simple_elementwise_kernel
  File "cupy/_core/core.pyx", line 2030, in cupy._core.core.compile_with_cache
  File "/home/leof/miniforge3/envs/opt_einsum_dev/lib/python3.8/site-packages/cupy/cuda/compiler.py", line 460, in compile_with_cache
    return _compile_with_cache_cuda(
  File "/home/leof/miniforge3/envs/opt_einsum_dev/lib/python3.8/site-packages/cupy/cuda/compiler.py", line 529, in _compile_with_cache_cuda
    mod.load(cubin)
  File "cupy/cuda/function.pyx", line 241, in cupy.cuda.function.Module.load
  File "cupy/cuda/function.pyx", line 243, in cupy.cuda.function.Module.load
  File "cupy_backends/cuda/api/driver.pyx", line 253, in cupy_backends.cuda.api.driver.moduleLoadData
  File "cupy_backends/cuda/api/driver.pyx", line 125, in cupy_backends.cuda.api.driver.check_status
cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_MISALIGNED_ADDRESS: misaligned address
  File "cupy_backends/cuda/api/driver.pyx", line 260, in cupy_backends.cuda.api.driver.moduleUnload
  File "cupy_backends/cuda/api/driver.pyx", line 125, in cupy_backends.cuda.api.driver.check_status
cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_MISALIGNED_ADDRESS: misaligned address
Exception ignored in: 'cupy.cuda.function.Module.__dealloc__'
(... repeat the above traceback N times ...)
``` 

By the way, technically this array is illegal in CuPy (`a.ndim=27` > `MAX_NDIM=25`, see #4193), but I am not sure if this is relevant.

leofang:
Fix #4193. Fix #5721.

This PR bumps `MAX_NDIM` to 32, in line with NumPy's choice. We could bump it higher, say 40 as in cuTENSOR. However, for ndarrays with ndim>32 they wouldn't be able to be copied as NumPy ndarrays, so I gave up...

I add a check in `_set_shape_and_strides`, which is called in all reshape-related APIs, to prevent unacceptable reshaping from happening.

For `CArray` and `CIndexer` I also add the same check, but the above ndarray check makes it impossible to test these checks from Python IIUC. I'd consider it an extra safety. (Note that the device codes for them actually work for all ndims, so no need to add `static_assert` there. It's just that on the host side we have to use a fixed-size struct.)


cupy/_core/_routines_manipulation.pyx
```diff
@@ -57,13 +57,7 @@ cdef _ndarray_shape_setter(ndarray self, newshape):
         newshape = (newshape,)
     shape = internal.infer_unknown_dimension(newshape, self.size)
     _get_strides_for_nocopy_reshape(self, shape, strides)
-    if strides.size() != shape.size():
-        raise AttributeError(
-            'Incompatible shape for in-place modification. Use `.reshape()` '
-            'to make a copy with the desired shape.')
```
leofang, line 63: Hi @toslunar, I noticed you touched this error message in #5470, but it seems to me there is no way to trigger this error. Could you confirm?

leofang:
Let me do a test drive.

Jenkins, test this please

cupy/_core/_routines_manipulation.pyx
toslunar: I should have added a test.

```
>>> x = cupy.zeros((2, 3), order='F')
>>> x.shape = (3, 2)
  File "cupy/_core/core.pyx", line 297, in cupy._core.core.ndarray.shape.__set__
  File "cupy/_core/_routines_manipulation.pyx", line 61, in cupy._core._routines_manipulation._ndarray_shape_setter
AttributeError: Incompatible shape for in-place modification. Use `.reshape()` to make a copy with the desired shape.
```

leofang: Thanks, @toslunar. No worries, let me add it.

leofang: Done in a0e1ae5.

cupy/_core/_carray.pxd
```diff
@@ -7,7 +7,11 @@ from cupy.cuda cimport function
 ctypedef vector.vector[Py_ssize_t] shape_t
 ctypedef vector.vector[Py_ssize_t] strides_t
 
-DEF MAX_NDIM = 25
+# this matches NPY_MAXDIMS
+DEF MAX_NDIM = 32
+cdef inline int get_max_ndim():
+    # we can't access module-level constants due to Cython's limitation...
+    return MAX_NDIM
```
kmaehashi, line 14: Not tested, but I guess this should work?
```
cdef enum: MAX_NDIM = 32
```

leofang: This works indeed! Thanks. I was scratching my head when hitting https://github.com/cython/cython/issues/4369, knowing I knew a way that it works but couldn't recall...

leofang:
Jenkins, test this please

leofang:
Not sure if the coverage test has a glitch or not, let me retry.

Jenkins, test this please

leofang:
Jenkins, test this please

leofang:
btw I tried to allow `CArray` on host to be flexible in `ndim`, however the main blocker is that if the struct size is unknown, the CUDA Driver API can't copy it to the host when launching a kernel. Any suggestion is welcome!

leofang:
@kmaehashi Can we backport this to v9 please? We need this fix to run some tests 🙂

cupy/_core/core.pyx
```diff
@@ -1799,6 +1799,11 @@ cdef class ndarray:
                                  bint update_f_contiguity):
         if shape.size() != strides.size():
             raise ValueError('len(shape) != len(strides)')
+        if shape.size() > _carray.MAX_NDIM:
+            msg = f'The input array dimension ({shape.size()}) is '
+            msg += 'beyond CuPy\'s current limitation ('
+            msg += f'{_carray.MAX_NDIM})'
+            raise ValueError(msg)
```
kmaehashi, line 1806: How about using the same error message as NumPy? 😄 

```
>>> numpy.empty((1,)*34)
ValueError: maximum supported dimension for an ndarray is 32, found 34
```

cupy/_core/_carray.pyx
```diff
@@ -10,8 +10,12 @@ cdef class CArray(function.CPointer):
 
     cdef void init(
             self, void* data_ptr, Py_ssize_t data_size,
-            const shape_t& shape, const strides_t& strides):
+            const shape_t& shape, const strides_t& strides) except*:
         cdef size_t ndim = shape.size()
+        assert ndim == strides.size()
+        if ndim > MAX_NDIM:
```
kmaehashi, line 16: I feel `assert ndim <= MAX_NDIM` is enough as these classes are private and not constructed by users.

cupy/_core/_carray.pyx
```diff
@@ -26,7 +30,11 @@ cdef class CArray(function.CPointer):
 
 cdef class CIndexer(function.CPointer):
 
-    cdef void init(self, Py_ssize_t size, const shape_t &shape):
+    cdef void init(self, Py_ssize_t size, const shape_t &shape) except*:
+        cdef size_t ndim = shape.size()
+        if ndim > MAX_NDIM:
```
kmaehashi, line 35: ditto

kmaehashi:
> Can we backport this to v9 please? We need this fix to run some tests 🙂

@leofang This change is not trivial and may affect performance for existing users, so I'm thinking it's safe to avoid backporting. Which tests require this fix? 👀

leofang:
> This change is not trivial and may affect performance for existing users, so I'm thinking it's safe to avoid backporting. Which tests require this fix? 👀

We have a project that would generate large dimension arrays and it's why I hit https://github.com/cupy/cupy/issues/5721. If it's a concern then we can just apply the patch locally; it's a bit of work but not too much.

I'm interested in the perf impacts. What kind of workload should we use to measure this? Perhaps we measure the launch speed of kernels with 
- 1 arg
- 2 args
- 3 args

(all arguments are ndarrays) and compare the timings?

cupy/_core/_carray.pyx
leofang: Good point. I don't remember what I was thinking, but indeed `CArray` and `CIndexer` are not constructed by users.

cupy/_core/core.pyx
leofang: Fixed in f534a49.

cupy/_core/_carray.pyx
leofang: Fixed in f534a49.

cupy/_core/_carray.pyx
leofang: Fixed in f534a49.

leofang:
I also fixed the large allocation problem (https://github.com/cupy/cupy/issues/5721#issuecomment-928007983) in f0bad07. This is ready.

kmaehashi:
> We have a project that would generate large dimension arrays and it's why I hit #5721. If it's a concern then we can just apply the patch locally; it's a bit of work but not too much.

Ah, ok I wasn't thinking you actually hit this limit... worth backporting in that case.

> Perhaps we measure the launch speed of kernels with
> 1 arg
> 2 args
> 3 args
> (all arguments are ndarrays) and compare the timings?

Yes, this sounds perfect.

leofang:
I ran a quick check instead of a more systematic one:
```python
def run(size):
    a = cp.random.random(size)
    b = cp.ones(size, dtype=a.dtype)
    a = cp.sin(a * b)
    b = cp.sum(a)
    return b

repeat(run, (1024**2,), n_repeat=1000)
```
The generated kernels are (unsorted in appearance order)
```
__global__ void cupy_random_x_mod_1(CArray<double, 1, 1, 1> _raw_x, CIndexer<1> _ind)
__global__ void cupy_sin__float64_float64(const CArray<double, 1, 1, 1> _raw_in0, CArray<double, 1, 1, 1> _raw_out0, CIndexer<1> _ind)
__global__ void cupy_sum(const CArray<double, 1, 1, 1> _raw_in0, CArray<double, 0, 1, 1> _raw_out0, CIndexer<1> _in_ind, CIndexer<0> _out_ind, const int _block_stride)
__global__ void cupy_multiply__float64_float64_float64(const CArray<double, 1, 1, 1> _raw_in0, const CArray<double, 1, 1, 1> _raw_in1, CArray<double, 1, 1, 1> _raw_out0, CIndexer<1> _ind)
__global__ void cupy_fill(const double x, CArray<double, 1, 1, 1> _raw_y, CIndexer<1> _ind)
```
Before and after this PR, the CPU and GPU times remain largely unaffected (50us & 996us on my system, with some small fluctuations).

PR #5754, opened 2021-09-14
Reorganize build scripts, part 4: remove global `cupy_setup_options`

kmaehashi:
context: #5722, #5730, #5743, #5745

Remove the global variable `cupy_setup_options`.
Recommends commit-by-commit review.

PR #5745, opened 2021-09-11
Reorganize build scripts, part 3: simplify `setup.py`

kmaehashi:
context: #5722, #5730, #5743

This also adds a message to take care of the source tarball is downloaded from GitHub.

Depends on ~#5743~ (merged)

PR #5743, opened 2021-09-11
Reorganize build scripts, part 2: separate modules

kmaehashi:
context: #5722, #5730

Please review commit by commit.

PR #5730, opened 2021-09-09
Reorganize build scripts, part 1

kmaehashi:
Installers are now reorganized to a package. #5722

PR #5755, opened 2021-09-14
Reorganize build scripts, part 5: remove Cython version check

kmaehashi:
context: #5722, #5730, #5743, #5745, #5754

As Cython is now in `setup_requires`, version check is no longer necessary.

Depends on #5754

### DIFF
--- a/install/cupy_builder/cupy_setup_build.py
+++ b/install/cupy_builder/cupy_setup_build.py
@@ -26,11 +26,6 @@ try:
 except Exception:
     pass
 
-# Cython requirements (minimum version and versions known to be broken).
-# Note: this must be in sync with setup_requires defined in setup.py.
-required_cython_version = pkg_resources.parse_version('0.29.22')
-ignore_cython_versions = [
-]
 use_hip = build.use_hip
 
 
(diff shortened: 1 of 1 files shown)


emcastillo:
I took the liberty of rebasing 🙇‍♂️

emcastillo:
Jenkins, test this please

install/cupy_builder/cupy_setup_build.py
```diff
@@ -512,9 +503,9 @@ def check_extensions(extensions):
             if not path.isfile(f):
                 raise RuntimeError('''\
 Missing file: {}
-Please install Cython {} or later. Please also check the version of Cython.
+Please install Cython.
 See https://docs.cupy.dev/en/stable/install.html for details.
```
emcastillo, line 507: From the docs: 
Cython 0.29.22 or later is required to build CuPy from source. It will be automatically installed during the build process if not available.


Here it doesn't specify that Cython should be less than 3.0
there is no upper bound in Cython currently? I think we should modify the docs to specify this.

emcastillo (approved):

PR #5757, opened 2021-09-15
Fix `cupyx.optimize` to save file when no optimization ran

kmaehashi:
Previously `cupyx.optimize` context manager does not save a file when no optimization run within the scope.
I think it is better to save an empty result instead to make the behavior consistent.

This PR also fixes the issue that `cupyx.optimize` tests fail when CUB acceleartor is enabled.

### DIFF
--- a/cupyx/optimizing/_optimize.py
+++ b/cupyx/optimizing/_optimize.py
@@ -101,7 +101,8 @@ The optimization results will never be stored.
 
     try:
         yield context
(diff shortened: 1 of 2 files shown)


emcastillo:
Jenkins, test this please

emcastillo (approved):
LGTM

kmaehashi:
pfnCI, run Jenkins please.

Issue #5762, opened 2021-09-20
Supporting `bitorder='little'` in `cupy.unpackbits` function as for `numpy` 


iskode:
### Description
This issue is about getting know your opinions concerning `bitorder` support in `cupy.unpackbits` as in numpy.
I've been implemented and tested it locally.

### Motivation or Use case
I'm trying to unpack bits from an array of a cudf dataframe column so as to retrieve the validation mask
for missing values. Below is a code snippet to elucidate the use case:
```
>>> import cudf
>>> import cupy as cp
>>> import numpy as np
>>> df = cudf.DataFrame({'a': np.arange(10)})
>>> df.at[[1, 9]: 'a'] = None
>>> df
      a
0     0
1  <NA>
2     2
3     3
4     4
5     5
6     6
7     7
8     8
9  <NA>
>>> valid_mask = cp.asarray(df.a._column.mask)
>>> valid_mask_np = df.a._column.mask.to_host_array()
>>> cp.unpackbits(valid_mask)[:10]
array([1, 1, 1, 1, 1, 1, 0, 1, 1, 1], dtype=uint8)
>>> np.unpackbits(valid_mask_np)[:10]
array([1, 1, 1, 1, 1, 1, 0, 1, 1, 1], dtype=uint8)
>>> np.unpackbits(valid_mask_np, bitorder='little')[:10]
array([1, 0, 1, 1, 1, 1, 1, 1, 1, 0], dtype=uint8)
```
As you can see, calling `unpackbits` from either `numpy` or `cupy` without setting `bitorder`
to `'little'` which results to `bitorder='big'`, gives the same results `array([1, 1, 1, 1, 1, 1, 0, 1, 1, 1], dtype=uint8)`.
This validity mask is not the expected one from the `df['a']` due to bit ordering. However, setting `bitorder='little'` gives the expected array as you can see the 0 value indexes correspond to NA values and 1 values to valid or non NA values. 

Can I submit a PR for implementing the `bitorder='little'` feature  ?

emcastillo:
> Can I submit a PR for implementing the bitorder='little' feature ?

Yes please!! it will be a great contribution :)

kmaehashi:
Implemented via #5765

PR #5765, opened 2021-09-21
Adding 'bitorder' support to `cupy.unpackbits` 

iskode:
This PR is about this issue: https://github.com/cupy/cupy/issues/5762

### DIFF
--- a/cupy/_binary/packing.py
+++ b/cupy/_binary/packing.py
@@ -62,7 +62,7 @@ def unpackbits(myarray, bitorder='big'):
     Args:
         myarray (cupy.ndarray): Input array.
 
-        bitorder : {'big', 'little'}, optional. Defaults to 'big'
+        bitorder (str, optional): bit order to use when unpacking the array, allowed values are `'little'` and `'big'`. Defaults to `'big'`.
 
     Returns:
         cupy.ndarray: The unpacked array.
--- a/cupy/_binary/packing.py
+++ b/cupy/_binary/packing.py
@@ -62,7 +62,9 @@ def unpackbits(myarray, bitorder='big'):
     Args:
         myarray (cupy.ndarray): Input array.
 
-        bitorder (str, optional): bit order to use when unpacking the array, allowed values are `'little'` and `'big'`. Defaults to `'big'`.
+        bitorder (str, optional): bit order to use when unpacking the array,
+
+        allowed values are `'little'` and `'big'`. Defaults to `'big'`.
 
     Returns:
         cupy.ndarray: The unpacked array.
--- a/cupy/_binary/packing.py
+++ b/cupy/_binary/packing.py
@@ -62,7 +62,7 @@ def unpackbits(myarray, bitorder='big'):
     Args:
         myarray (cupy.ndarray): Input array.
 
-        bitorder{'big', 'little'}, optional. Defaults to 'big'
+        bitorder : {'big', 'little'}, optional. Defaults to 'big'
 
     Returns:
         cupy.ndarray: The unpacked array.
@@ -73,7 +73,7 @@ def unpackbits(myarray, bitorder='big'):
         raise TypeError('Expected an input array of unsigned byte data type')
 
     if bitorder not in ['big', 'little']:
-        raise ValueError(f"bitorder must be either 'big' or 'little' not '{bitorder}'")
+        raise ValueError("bitorder must be either 'big' or 'little'")
 
     unpacked = cupy.ndarray((myarray.size * 8), dtype=cupy.uint8)
     return _unpackbits_kernel[bitorder](myarray, unpacked)
--- a/cupy/_binary/packing.py
+++ b/cupy/_binary/packing.py
@@ -42,16 +42,16 @@ def packbits(myarray):
     return _packbits_kernel(myarray, myarray.size, packed)
 
 
-_unpackbits_kernel = { 'big': _core.ElementwiseKernel(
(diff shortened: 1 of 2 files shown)


cupy/_binary/packing.py
```diff
@@ -42,21 +42,28 @@ def packbits(myarray):
...
-def unpackbits(myarray):
+def unpackbits(myarray, bitorder='big'):
     """Unpacks elements of a uint8 array into a binary-valued output array.
 
     This function currently does not support ``axis`` option.
 
     Args:
         myarray (cupy.ndarray): Input array.
 
+        bitorder : {'big', 'little'}, optional. Defaults to 'big'
```
emcastillo, line 65:
```suggestion
        bitorder (str, optional): bit order to use when unpacking the array, allowed values are `'little'` and `'big'`. Defaults to `'big'`.
```

emcastillo: minor fix to docs to keep format consistent.

tests/cupy_tests/binary_tests/test_packing.py
```diff
@@ -39,3 +40,17 @@ def test_unpackbits(self):
...
+
+    def test_pack_unpack_order(self):
+        for bo in ['big', 'little']:
+            self.check_unpackbits([], bitorder=bo)
+            self.check_unpackbits([0], bitorder=bo)
+            self.check_unpackbits([1], bitorder=bo)
+            self.check_unpackbits([255], bitorder=bo)
+            self.check_unpackbits([100, 200, 123, 213], bitorder=bo)
+        a = cupy.array([10, 20, 30], dtype=cupy.uint8)
+        pytest.raises(ValueError, cupy.unpackbits, a, bitorder='r')
```
emcastillo, line 55: can you separate these two assers into a new test?

`def test_unpack_invalid_order` or something similar :).
Thanks!

tests/cupy_tests/binary_tests/test_packing.py
```diff
@@ -39,3 +40,17 @@ def test_unpackbits(self):
...
+        pytest.raises(TypeError, cupy.unpackbits, a)
+        pytest.raises(TypeError, cupy.unpackbits, a.astype(float))
+
+    def test_pack_unpack_order(self):
+        for bo in ['big', 'little']:
+            self.check_unpackbits([], bitorder=bo)
+            self.check_unpackbits([0], bitorder=bo)
+            self.check_unpackbits([1], bitorder=bo)
+            self.check_unpackbits([255], bitorder=bo)
+            self.check_unpackbits([100, 200, 123, 213], bitorder=bo)
```
emcastillo, line 53: seems like check is not asserting that the bits where correctly unpacked. Can you add a check that verifies that the unpacking was effectively done in big or little endian?

emcastillo (commented):
In general looks great! I only have some doubts about the tests :)
Thanks again

tests/cupy_tests/binary_tests/test_packing.py
iskode: Thank you for the suggestion. I did the same for the two asserts I've added to `test_unpackbits` in order to assert the array validity.

tests/cupy_tests/binary_tests/test_packing.py
iskode: Here we're checking against numpy counterpart implementation following `test_unpackbits`
pattern. So if 'big' endian from cupy matches that of numpy then we're doing a good job.
Should I not test against numpy in the case of bitorder ?

emcastillo: ah sorry, i did not properly read the check code, then this is fine!!

cupy/_binary/packing.py
```diff
@@ -42,21 +42,28 @@ def packbits(myarray):
...
-def unpackbits(myarray):
+def unpackbits(myarray, bitorder='big'):
     """Unpacks elements of a uint8 array into a binary-valued output array.
 
     This function currently does not support ``axis`` option.
 
     Args:
         myarray (cupy.ndarray): Input array.
 
+        bitorder (str, optional): bit order to use when unpacking the array, allowed values are `'little'` and `'big'`. Defaults to `'big'`.
```
emcastillo, line 65:
```suggestion
        bitorder (str, optional): bit order to use when unpacking the array,
        allowed values are `'little'` and `'big'`. Defaults to `'big'`.
```

emcastillo: my bad :D

emcastillo:
Jenkins, test this please

emcastillo (approved):

iskode:
Hi @emcastillo ,
sorry :( about my answer concerning the clarity of 'big'/'little' endian tests. Now thinking back, I believe this test pattern is not good. We must have tests with clear expected values asserted not only against `numpy`. That means if there is a bug on numpy we should reproduce it and also it is a burden to drag this dependency as the library evolves.
Thus I'd like to refactor the whole file `test_packing` by adding tests free from numpy.
What is your take on this ?

emcastillo:
Hey, sorry for my late reply!!!

> We must have tests with clear expected values asserted not only against numpy. That means if there is a bug on numpy we should reproduce it and also it is a burden to drag this dependency as the library evolves.

Yeah, this totally makes sense, but we don't need to go that far, actually, we try to guarantee numpy compatibility so just testing against numpy is fine, we do it with all routines. Thanks!!

PR #5774, opened 2021-09-22
Add `cupy.positive()`

### LINKED ISSUE #5702: Missing `cupy.positive()`
See https://numpy.org/doc/stable/reference/generated/numpy.positive.html for `numpy.positive()`. This is basically a no-op + copying the input (subject to allowed dtypes etc). This is needed by the Array API standard.

takagi:
Close #5702.

This PR adds `cupy.positive()` as well as `ndarray.__pos__()`, which return elementwise-numerical positive.

As for boolean dtype, Array API standard says the input array should have a numeric data type, which does not include boolean.
https://data-apis.org/array-api/latest/API_specification/elementwise_functions.html#positive-x

Also, numpy raises the following exceptions, which I don't think is worth exactly mimicking in CuPy so our counterparts just raise `TypeError`.

```Python
>>> import numpy as np
>>> np.positive(np.array(True, dtype=np.bool_))
numpy.core._exceptions.UFuncTypeError: ufunc 'positive' did not contain a loop with signature matching types dtype('bool') -> dtype('bool')
>>> +np.array(True, dtype=np.bool_)
<stdin>:1: DeprecationWarning: Applying '+' to a non-numerical array is ill-defined. Returning a copy, but in the future this will error.
True
```

Regarding `__pos__`, CuPy currently returns the ndarray instance itself, however, both Array API and NumPy say the results must equal the results returned by `positive()`, so I fixed it to align the way. This is a no-compat change.
https://data-apis.org/array-api/latest/API_specification/array_object.html#pos-self

```Python
>>> import numpy as np
>>> a = np.array([1])
>>> a is +a
False
```


cupy/_core/_routines_math.pyx
```diff
@@ -902,6 +902,24 @@ _angle = create_ufunc(
     ''')
 
 
+def _positive_boolean_error():
+    raise TypeError(
```
toslunar, line 906: Unlike `negative`, it would be just `DeprecationWarning` with the current NumPy.

toslunar (changes requested):
I'd like to test `+a is not a`.  Should it be done in this PR or in an array-API PR?

takagi:
Sure! I'll do it here.

cupy/_core/_routines_math.pyx
takagi: Okay, I'll fix it so! Also, Array API layer is to check a boolean input.

tests/cupy_tests/core_tests/test_ndarray_elementwise_op.py
```diff
@@ -480,6 +480,12 @@ def test_add_array_boolarray(self):
     def test_iadd_array_boolarray(self):
         self.check_array_boolarray_op(operator.iadd)
 
+    def test_pos_boolarray(self):
+        for xp in (numpy, cupy):
+            a = xp.array(True, dtype=xp.bool_)
+            with pytest.deprecated_call():
+                assert a is not +a
```
toslunar, line 487: `tests/cupy_tests/core_tests/test_ndarray_unary_op.py::TestArrayUnaryOp` would be the correct place for the test.

- For `bool`, could you move the test to there?
- For the other dtypes, how about testing `a is not +a` in `test_pos_array`?

toslunar (changes requested):
LGTM except for test

tests/cupy_tests/core_tests/test_ndarray_elementwise_op.py
takagi: I worried about where to place the tests when I posted this PR, thanks!

takagi:
PTAL!

PR #5780, opened 2021-09-26
Release the GIL for several CUDA ops

hauntsaninja:
In a similar vein as #5574

### DIFF
--- a/cupy_backends/cuda/api/runtime.pyx
+++ b/cupy_backends/cuda/api/runtime.pyx
@@ -807,7 +807,9 @@ cpdef launchHostFunc(intptr_t stream, callback, intptr_t arg):
 
 
 cpdef streamQuery(intptr_t stream):
-    return cudaStreamQuery(<driver.Stream>stream)
+    with nogil:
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thank you for the pull-request @hauntsaninja!
I'm curious if these fixes improved performance in your usecase. Originally these functions are intentionally kept holding GIL as we assumed they return immediately. Releasing & re-acquiring GIL is not a free operation as Cython docs says:

https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#acquiring-and-releasing-the-gil

> Note that acquiring the GIL is a blocking thread-synchronising operation, and therefore potentially costly. It might not be worth releasing the GIL for minor calculations. Usually, I/O operations and substantial computations in parallel code will benefit from it.

hauntsaninja:
Thanks for pushing back! We didn't measure this one as carefully as the NCCL one (it also doesn't lead to deadlocks like that one did). Let me run a more careful benchmark on this and get back to you :-)

PR #5560, opened 2021-07-25
Add hypergeometric distribution to new Generator

povinsahu1909:

PR #5785, opened 2021-09-27
Fix hypergeometric distribution implementation to use int

takagi:
Follow up #5560.

This PR fixes the hypergeometric distribution to use int instead of double in its implementation.


emcastillo:
Jenkins, test this please

emcastillo:
CI failures seem to be real :(

PR #5794, opened 2021-09-27
Do not ignore UserWarning

### LINKED ISSUE #5790: Remove `ignore:UserWarning`
It seems it's added in #1066, but I think UserWarnings should be captured.

kmaehashi:
Closes #5790.


kmaehashi:
pfnCI, test this please.

(Let's check what will happen...)

PR #5826, opened 2021-10-02
Tentatively disable pytest-xdist

kmaehashi:
It seems that recently CI hangs for some reason.

https://ci.preferred.jp/cupy.linux.cuda114/
https://ci.preferred.jp/cupy.linux.cuda11x-cuda-python/

(Some timeout after running ``6h00m`` hard limit)

I haven't identified the root cause yet, but I'd like to disable recently-introduced pytest-xdist to isolate the issue.


emcastillo (approved):
LGTM!

Issue #5831, opened 2021-10-04
Random State returning inconsistent results 


Nanthini10:
cuPy random state is resulting in different results for length of array > 128.  See code below, with n=128 the following code returns the same shuffled indices. When n=129 it returns a different array once in the 10 runs.

See https://github.com/rapidsai/cuml/issues/4238 for more context

<details>
<summary>Environment details</summary>

```
OS                           : Linux-5.4.0-65-generic-x86_64-with-glibc2.10
Python Version               : 3.8.12
CuPy Version                 : 9.4.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 1.21.2
SciPy Version                : 1.6.0
Cython Build Version         : 0.29.24
Cython Runtime Version       : 0.29.24
CUDA Root                    : /opt/conda/envs/rapids
nvcc PATH                    : None
CUDA Build Version           : 11000
CUDA Driver Version          : 11000
CUDA Runtime Version         : 11000
cuBLAS Version               : (available)
cuFFT Version                : 10201
cuRAND Version               : 10201
cuSOLVER Version             : (10, 6, 0)
cuSPARSE Version             : (available)
NVRTC Version                : (11, 0)
Thrust Version               : 100909
CUB Build Version            : 100909
Jitify Build Version         : <unknown>
cuDNN Build Version          : None
cuDNN Version                : None
NCCL Build Version           : 21003
NCCL Runtime Version         : 21003
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : Quadro RTX 8000
Device 0 Compute Capability  : 75
Device 0 PCI Bus ID          : 0000:15:00.0
Device 1 Name                : Quadro RTX 8000
Device 1 Compute Capability  : 75
Device 1 PCI Bus ID          : 0000:2D:00.0
```
</details>

* Code to reproduce
```
n = 128 # change to 129
for _ in range(10):
    import cupy as cp
    random_state = cp.random.RandomState(seed=1)
    tmp = cp.arange(n) 
    random_state.shuffle(tmp)
    print(tmp)

```

PR #5838, opened 2021-10-05
Fix non-determinisitc behavior in `cupy.random.shuffle`

asi1024:
Close https://github.com/cupy/cupy/issues/5831. (cc: @Nanthini10)

`_cupy_permutation` kernel has a non-deterministic behavior.

There is the issue of non-determinism behavior in the kernel algorithm itself, so it cannot be solved by simply modifying the kernel implementation.
This PR will be modified to always use `argsort` for any array size.

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -1111,31 +1111,9 @@ class RandomState(object):
 
     def _permutation(self, num):
         """Returns a permuted range."""
-        sample = cupy.empty((num), dtype=numpy.int32)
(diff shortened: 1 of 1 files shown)


kmaehashi:
ping @anaruse, as the kernel was introduced for performance in #603.
(The issue #5831 is that `RandomState.shuffle()` is nondetermitisic for `128 < num` case even when seed is specified)

asi1024:
Now rewrote a new kernel for permutation algorithm.

Issue #5832, opened 2021-10-04
Enhancment: Confusion with NotImplementedError


lanttu1243:
In **cupy\random\_generator.py line 1045** instead of raising a **NotImplementedError**, a **ValuError('size cannot be None')** would make using cupy easier and less confusing. To findout what caused the problem I had to follow the instance trail to **_generator.py line 1045** to find out what was wrong instead of just getting a **ValueError size cannot be None** which would have immeadiately told me what was wrong.

kmaehashi:
Thanks for the feedback! I think your suggestion makes sense.

https://github.com/cupy/cupy/blob/v10.0.0b3/cupy/random/_generator.py#L1044-L1045

If you rae interested could you send a pull-request? 😄

PR #5841, opened 2021-10-05
Enhancement: NotImplementedError clarity

lanttu1243:
Changed ***raise NotImplementedError*** to ***raise ValueError('size cannot be none')*** in ***def choice(self, a, size=None, replace=True, p=None)*** on line ***1045***.
[https://github.com/cupy/cupy/issues/5832]


kmaehashi:
pfnCI, test this please.

PR #5842, opened 2021-10-06
Code-fix on `cupy.array()`

takagi:
This PR renames and fixes `_get_concat_shape()` called in `cupy.array()` to make what and how it does clearer.


takagi:
Let me kick the CI to see it makes no degradation.

takagi:
BTW, I suppose what we want to do here is stack rather than concatenate, if we do not have any problem doing that.

takagi:
pfnCI, test this please.

PR #5849, opened 2021-10-08
Fix ROCm tests does not export LLVM_PATH

kmaehashi:
During rebase of #5474, `ENV LLVM_PATH` was mistakenly lost from Dockerfile.
This reverts it so that ROCm 4.3 test pass.


kmaehashi:
pfnCI, test this please.

kmaehashi:
Confirmed that `HIPRTC compilation error: <built-in>:1:10: fatal error: '__clang_hip_runtime_wrapper.h' file not found` disappeared. Please ignore other Jenkins failures; will be fixed in #5851

PR #5850, opened 2021-10-08
Add test for CI generator

kmaehashi:
This makes sure that the modification to schema/matrix YAML is in sync with generated files.

Jenkins/FlexCI not required.


kmaehashi:
Running... https://github.com/kmaehashi/cupy/actions/runs/1327641112

Issue #5297, opened 2021-05-31
`compile_with_cache()` is mistakenly exposed to public?


leofang:
I don't know how it started (I am sure it's not in CuPy's doc at all) but I've seen several bug reports from the ML community calling this function (such as #5294)
https://github.com/cupy/cupy/blob/4469fae998df33c72ff40ef954cb08b8f0004b18/cupy/cuda/__init__.py#L95
Since we already have `cupy.RawModule` to meet this need (and beyond), I think we can safely remove it from `cupy/cuda/__init__.py`, perhaps with a deprecation warning.

kmaehashi:
RawKernel introduced in v5 and RawModule introduced in v7. `compile_with_cache` has never been documented API, but it's widely used by major users like Thinc and 3d-ken-burns which need `Raw*` features before we implement it.

I basically agree with the deprecation but we also need to take care of existing users. I think we need some more steps than usual deprecations, e.g.:

* DeprecationWarning (silent by default) in v10
* Warning (visible to users) in v11/v12
* Removal in v13

leofang:
@kmaehashi Shall we tag this with v10.0rc? 🙂

kmaehashi:
Yeah, let's do it in v10rc!

leofang:
Yup, let me send a follow-up PR to mention this in the v10 upgrade guide.

kmaehashi:
Added this one to v13.0.0b1.

PR #5858, opened 2021-10-08
Deprecate `cupy.cuda.compile_with_cache`

kmaehashi:
As discussed in #5297, deprecating the legacy compile API.

I've renamed the existing `compile_with_cache` function under `cupy.cuda` to `_compile_module_with_cache` to distinguish it with `cupy._core.core.compile_with_cache`.


asi1024:
LGTM.

PR #5723, opened 2021-09-09
[WIP] Use cudaRuntimeGetVersion instead of CUDA_VERSION for CUDA Python support

takagi:
Close #5696. Supersedes #5709.

PR #5861, opened 2021-10-08
Avoid using `driver.get_build_version`

kmaehashi:
Follows-up #5723.


takagi:
(Let me kick CI again as I don't have a privilege to merge when required statuses not passing.

PR #5862, opened 2021-10-08
Add `array-api-tests` in FlexCI

### LINKED ISSUE #5781: Add FlexCI config to run array api test suite

kmaehashi:
Closes #5781.

Testing:
~https://ci.preferred.jp/cupy.linux.array-api/81600/~
https://ci.preferred.jp/cupy.linux.array-api/81613/

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -255,7 +255,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 3600
+      seconds: 7200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh array-api"


kmaehashi:
I can setup a Gitter room that posts a notification when test fails in `master` branch, but it can be a bit noisy for you as the test does not pass yet. Still wondering what would be the best way...

asmeurer:
It looks like you didn't give enough time for the tests to run. If the runtime of the test suite is too long, you can lower it with the `--max-examples` flag to pytest to lower the number of hypothesis examples. The default is 100, so `--max-examples=50` will make the tests run in about half the time. Since you will be running the CI all the time, lowering the number of examples per run wouldn't be a huge deal.

PR #5865, opened 2021-10-08
Document ppc64le and aarch64 are supported on conda-forge

leofang:


leofang:
@kmaehashi btw this can be backported.

PR #5872, opened 2021-10-11
Do not use cuTENSOR unless available

kmaehashi:


kmaehashi:
pfnCI, test this please.

PR #5873, opened 2021-10-11
Show error when importing `cupy.array_api` with Python 3.7

kmaehashi:
This is to avoid this on Py37:

```
>>> import cupy.array_api
__main__:1: UserWarning: The numpy.array_api submodule is still experimental. See NEP 47.
  File "/home/maehashi/Development/cupy/cupy/array_api/__init__.py", line 133, in <module>
    from ._creation_functions import (
  File "/home/maehashi/Development/cupy/cupy/array_api/_creation_functions.py", line 43
    /,
    ^
SyntaxError: invalid syntax
```

### DIFF
--- a/cupy/array_api/__init__.py
+++ b/cupy/array_api/__init__.py
@@ -118,6 +118,13 @@ Still TODO in this module are:
 
 """
 
+# CuPy-specific: still need to support Python 3.7.
+import sys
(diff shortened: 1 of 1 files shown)


asi1024:
Jenkins, test this please.

PR #5874, opened 2021-10-11
Add CI for pylibcugraph

kmaehashi:
This PR adds FlexCI to test pylibcugraph support. The test won't succeed until #5830 got merged, but please merge this one first so that tests can be run against #5830.


kmaehashi:
Testing: https://ci.preferred.jp/cupy.linux.cuda-rapids/80723/ (this should fail as this expects `pylibcugraph`)

PR #5881, opened 2021-10-12
Copy source directory to support pip 21.3

kmaehashi:
Previously pip copied the whole source tree before building it.
However pip 21.3, released today, stopped copying it.
As the source tree is read-only mounted in the Docker container we should manually copy it.

https://github.com/pypa/pip/issues/10128


kmaehashi:
pfnCI, run Jenkins please.

kmaehashi:
Jenkins broken, will fix

PR #5899, opened 2021-10-15
FlexCI test against Python 3.10

takagi:
Update FlexCI variants to test against Python 3.10.0. Note that NumPy and Scipy do not officially support Python 3.10 yet, and this PR will be updated after their official support is released.


kmaehashi:
Optuna depends on SciPy, so it tries to build SciPy...

kmaehashi:
SciPy 1.7.2 with Py310 wheels are out 😄 
/test

kmaehashi:
pfnCI, run Jenkins please.

PR #5900, opened 2021-10-16
NEW: Add trim_mean from scipy.stats to cupyx

carterbox:
This PR aims to add a single function, trim_mean, from scipy.stats to cupyx.

It is a trivial port because it only requires swapping numpy for cupy in the import headers. However, the scipy implementation does not work on cupy arrays because the first step in the scipy implementation is to convert the input to a numpy array.

Everything (including tests) are copied almost verbatim from the scipy source.


takagi:
Thanks! Let me merge this after CI pass, then let me rewrite the test so that it is checked against SciPy's result. CuPy's test should have a consistency check against NumPy/SciPy counterparts, but I don't think you need to learn how to write that only for this PR.

takagi:
pfnCI, test this please.

takagi:
@kmaehashi Would you approve the changes for your change request?

PR #5901, opened 2021-10-16
Fix `__repr__` of mode and scalar in cuTENSOR

kmaehashi:
This fixes the issue that `__repr__` is not working.
Also removed unnecessary `(object)`.


asi1024:
Jenkins, test this please.

PR #5902, opened 2021-10-16
Support using cuTENSOR in elementwise ufuncs

### LINKED ISSUE #5776: Enable cuTENSOR in elementwise operations
This can improve performance especially when tensors of different contiguity is given to binary operators, e.g., `c_cont_arr + f_cont_arr`.

kmaehashi:
Closes #5776.

This PR also fixes the issue in `cupy.cutensor.elementwise_binary` that the wrong mode is used for `C` (the second argument).

### DIFF
--- a/cupy/cutensor.pyx
+++ b/cupy/cutensor.pyx
@@ -5,6 +5,8 @@ import cupy as _cupy
 from cupy import _util
 from cupy.cuda import device as _device
 
+cimport cython
+from libcpp cimport vector
 from libc.stdint cimport intptr_t, uint32_t, uint64_t
 from cupy._core._carray cimport shape_t
 from cupy._core.core cimport ndarray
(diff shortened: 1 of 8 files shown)


asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

asi1024:
@kmaehashi PTAL.

cupy/cutensor.pyx
```diff
@@ -803,3 +807,77 @@ def _try_reduction_routine(
         reduce_op, _get_cutensor_compute_type(compute_dtype))
 
     return out
+
+
+def _all_strides_positive(ndarray a):
+    # cuTENSOR requires each stride > 0.
+    return all(s > 0 for s in a._strides)
```
asi1024, line 814: I feel this function might be worth speeding up.
```py
@cython.profile(False)
cdef inline bint _all_strides_positive(ndarray a):
    # cuTENSOR requires each stride > 0.
    cdef const vector.vector[Py_ssize_t]& strides = a._strides
    for i in range(<Py_ssize_t>strides.size()):
        if strides[i] <= 0:
            return False
    return True
```

kmaehashi:
Thanks for completing the remaining work @asi1024! I've benchmarked the current commit and it looks good to me.

Code: https://gist.github.com/kmaehashi/c2815216ef7e98730ac092d46ce85f37
Result: 
Screen Shot 2021-11-01 at 21 41 37

As for large arrays, the performance of C/C case (i.e., c-contiguous OP c-contiguous) remains almost the same as the current ufunc implementation, and the performance of C/F case significantly improved when cuTENSOR enabled. As for small arrays, cuTENSOR is slower but I think this is negilible.

kmaehashi:
pfnCI, test this please.

PR #5853, opened 2021-10-08
Change `Device.use()` semantics to align with `Stream.use()`

kmaehashi:
Closes #5705.

PR #5904, opened 2021-10-17
Fix to push device after setDevice succeed

kmaehashi:
Follows-up #5853.

`runtime.setDevice()` may raise in `__enter__`, and in that case `__exit__` will not fire and the stack will become inconsistent.


kmaehashi:
pfnCI, test this please.

takagi:
Waiting #5891 merged to pass rapids test.

PR #5915, opened 2021-10-18
Avoid use of setDevice in CuPy codebase

kmaehashi:
Follows-up: #5913, fixes regression in #5853

`dev = getDevice(); do_something(); dev.use()` in CuPy code may cause surprise when `do_something()` contains `with device:` context manager (e.g., within another CuPy code or user callback).


kmaehashi:
pfnCI, test this please.

PR #5926, opened 2021-10-20
Fix library installer to limit architecture

kmaehashi:


emcastillo:
Jenkins, test this please

PR #2150, opened 2019-04-17
Add `strides_check` option in array testing functions

okapies:
Add `strides_check` option. Separated from #2081.

PR #5968, opened 2021-10-29
Test strides

toslunar:


toslunar:
I missed `strides_check` (#2150).

kmaehashi:
I chatted about this with @asi1024 and @emcastillo, and thinking it's better to declare that we only guarantee consistency of contiguity and not strides between NumPy and CuPy, because (1) NumPy's behavior is sometimes difficult/costly to reproduce and (2) in some cases there can be additional overhead just to align the strides.

kmaehashi:
We've revisited this one today and agreed it's difficult to fix all functions to guarantee stride compatibility with NumPy. Let me close this one for now but let's reopen when there is a necessity.

PR #5970, opened 2021-11-01
Fix `__dlpack__` protocol

emcastillo:
When doing interop tests with PyTorch two issues were found,

`__dlpack_device__` returns the cupy device object instead of the id, so PyTorch crashes.
`__dlpack__` gets the current torch stream, which defaults to 0 (legacy stream), but for some reason, CuPy seems to error on it. I think it should be accepted (cc @leofang)

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -248,10 +248,10 @@ cdef class ndarray:
                     f' 1, 2, or any larger value, but {stream} was provided')
             if curr_stream_ptr == 0:
                 warnings.warn(
-                   'Stream 0 is passed from a library that owns a tensor you '
-                   'are converting from; CuPy assumes it as a legacy default '
-                   'stream. Please report this problem to the library as this '
-                   'violates the DLPack protocol.')
+                    'Stream 0 is passed from a library that owns a tensor you '
+                    'are converting from; CuPy assumes it as a legacy default '
+                    'stream. Please report this problem to the library as this'
+                    ' violates the DLPack protocol.')
                 curr_stream_ptr = runtime.streamLegacy
         else:  # ROCm/HIP
             if stream is None:
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -247,13 +247,11 @@ cdef class ndarray:
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
                     f' 1, 2, or any larger value, but {stream} was provided')
             if curr_stream_ptr == 0:
-                if not _util.DLPACK_ACCEPT_STREAM_ZERO:
-                    raise ValueError(
-                        'stream 0 is passed by library XXX, which is not '
-                        'accepted per the DLPack protocol. Please report this '
-                        'problem to the upstream library, and in the meantime '
-                        'work around this by setting the env var '
-                        'CUPY_DLPACK_ACCEPT_STREAM_ZERO=1')
+                warnings.warn(
+                   'Stream 0 is passed from a library that owns a tensor you '
+                   'are converting from; CuPy assumes it as a legacy default '
+                   'stream. Please report this problem to the library as this '
+                   'violates the DLPack protocol.')
                 curr_stream_ptr = runtime.streamLegacy
         else:  # ROCm/HIP
             if stream is None:
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -248,8 +248,8 @@ cdef class ndarray:
                     f' 1, 2, or any larger value, but {stream} was provided')
             if stream == 0:
                 warnings.warn(
-                    'Stream 0 is passed from a library that owns a tensor you '
-                    'are converting from; CuPy assumes it as a legacy default '
+                    'Stream 0 is passed from a library that you are converting to;'
+                    ' CuPy assumes 0 as a legacy default '
                     'stream. Please report this problem to the library as this'
                     ' violates the DLPack protocol.')
                 stream = runtime.streamLegacy
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -278,7 +278,7 @@ cdef class ndarray:
             attrs = runtime.pointerGetAttributes(self.data.ptr)
             is_managed = (
                 attrs.type == runtime.memoryTypeManaged
-                and _util.DLPACK_EXPORT_VERSION >= (0, 6))
+                and _util.CUPY_DLPACK_EXPORT_VERSION >= (0, 6))
             if is_managed:
                 device_type = dlpack.managed_CUDA
             else:
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -252,6 +252,7 @@ cdef class ndarray:
                     'are converting from; CuPy assumes it as a legacy default '
                     'stream. Please report this problem to the library as this'
                     ' violates the DLPack protocol.')
+                stream = runtime.streamLegacy
             if curr_stream_ptr == 0:
                 curr_stream_ptr = runtime.streamLegacy
         else:  # ROCm/HIP
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -248,8 +248,8 @@ cdef class ndarray:
                     f' 1, 2, or any larger value, but {stream} was provided')
             if stream == 0:
                 warnings.warn(
-                    'Stream 0 is passed from a library that you are converting to;'
-                    ' CuPy assumes 0 as a legacy default '
+                    'Stream 0 is passed from a library that you are'
+                    ' converting to; CuPy assumes 0 as a legacy default '
                     'stream. Please report this problem to the library as this'
                     ' violates the DLPack protocol.')
                 stream = runtime.streamLegacy
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -246,12 +246,13 @@ cdef class ndarray:
                 raise ValueError(
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
                     f' 1, 2, or any larger value, but {stream} was provided')
-            if curr_stream_ptr == 0:
+            if stream == 0:
                 warnings.warn(
                     'Stream 0 is passed from a library that owns a tensor you '
                     'are converting from; CuPy assumes it as a legacy default '
                     'stream. Please report this problem to the library as this'
                     ' violates the DLPack protocol.')
+            if curr_stream_ptr == 0:
                 curr_stream_ptr = runtime.streamLegacy
         else:  # ROCm/HIP
             if stream is None:
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -239,9 +239,13 @@ cdef class ndarray:
             if stream is None:
                 stream = runtime.streamLegacy
             elif not isinstance(stream, int) or stream < -1:
+                # DLPack does not accept 0 as a valid stream, but there is a
+                # bug in PyTorch that exports the default stream as 0, which
+                # renders the protocol unusable, we will accept a 0 value
+                # meanwhile.
                 raise ValueError(
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
-                    f' 0, 1, 2, or any larger value, but {stream} was provided')
+                    f' 1, 2, or any larger value, but {stream} was provided')
             if curr_stream_ptr == 0:
                 curr_stream_ptr = runtime.streamLegacy
         else:  # ROCm/HIP
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -238,10 +238,10 @@ cdef class ndarray:
         if not runtime._is_hip_environment:  # CUDA
             if stream is None:
                 stream = runtime.streamLegacy
-            elif not isinstance(stream, int) or stream < -1 or stream == 0:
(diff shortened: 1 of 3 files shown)


cupy/_core/core.pyx
```diff
@@ -238,10 +238,10 @@ cdef class ndarray:
         if not runtime._is_hip_environment:  # CUDA
             if stream is None:
                 stream = runtime.streamLegacy
-            elif not isinstance(stream, int) or stream < -1 or stream == 0:
+            elif not isinstance(stream, int) or stream < -1:
                 raise ValueError(
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
-                    f' 1, 2, or any larger value, but {stream} was provided')
+                    f' 0, 1, 2, or any larger value, but {stream} was provided')
```
leofang, line 244: This one is on PyTorch. We forbad 0 due to its ambiguity:
https://data-apis.org/array-api/latest/API_specification/array_object.html?dlpack-self-stream-none#dlpack-self-stream-none

cupy/_core/core.pyx
```diff
@@ -274,7 +274,7 @@ cdef class ndarray:
                 device_type = dlpack.device_CUDA
         else:
             device_type = dlpack.device_ROCM
-        return (device_type, self.device)
+        return (device_type, self.device.id)
```
leofang, line 277: Thanks, good catch! `self.data.device_id` should also work btw.

leofang (changes requested):

cupy/_core/core.pyx
emcastillo: ok, will send pr to pytorch!

emcastillo: we will accept this value meanwhile

emcastillo: https://github.com/pytorch/pytorch/pull/67618

kmaehashi:
pfnCI, test this please.

kmaehashi (dismissed):

cupy/_core/core.pyx
leofang: Is it possible, then, to use an env var to control this, and default to not accept? Otherwise this would defeat the purpose of all the hard debates for removing 0 ...

emcastillo: right now the only framework accepting this is PyTorch and it doesn't have a way to use PTDS by default, so 99,9% of the time, we will get a 0 passed to `__dlpack__` and people will start complaining in stackflow or the issues section that the code they copypasted doesn't work 😄.

I don't mind adding the env var, but I think that for the mean time it is better to accept 0 until this is fixed in pytorch, otherwise the pytorch-cupy tandem won't work by default and that feels worse.

I agree that this is not something that should be done, but the alternatives feel worse to me ...

leofang: > for the mean time it is better to accept 0 until this is fixed in pytorch

I think it only adds one more reason to disable by default. This is a workaround, not a permanent thing. We could have something like this:
```python
if stream_ptr == 0:
    if CUPY_DLPACK_ACCEPT_STREAM_ZERO: 
        warnings.warn('stream 0 is accepted only to help the transition of external libraries, '
                      'this support will be removed in the future to comply with the DLPack protocol')
        # do the work
    else:
        raise ValueError('stream 0 is passed by library XXX, which is not accepted per the DLPack protocol. ' 
                         'Please report this problem to the upstream library, and in the meantime work around this by '
                         'setting the env var CUPY_DLPACK_ACCEPT_STREAM_ZERO=1')

Comment (emcastillo) [reply cupy/_core/core.pyx]: yeah, I just gave it another thought in the shower and reached a similar conclusion.
I think the warning is not even needed. Just hard error and tell the user to set that env var should make her/him aware

File: docs/source/reference/environment.rst [hunk 3]
```diff
@@ -104,6 +104,13 @@ Here are the environment variables that CuPy uses at runtime.
 
   This controls CuPy's DLPack support. Currently, setting a value smaller than 0.6 would disguise managed memory as normal device memory, which enables data exchanges with libraries that have not updated their DLPack support, whereas starting 0.6 CUDA managed memory can be correctly recognized as a valid device type.
 
+.. envvar:: CUPY_DLPACK_ACCEPT_STREAM_ZERO
+
+  Default: ``0``
+
+  This controls CuPy's DLPack protocol interoperability. Some external libraries may pass a current stream value of 0, which is not accepted as per the DLPack protocol specification in the `Array API standard <https://data-apis.org/array-api/latest/API_specification/array_object.html?dlpack-self-stream-none#dlpack-self-stream-none>`_
```
leofang, line 111: nit: miss period
```suggestion
  This controls CuPy's DLPack protocol interoperability. Some external libraries may pass a current stream value of 0, which is not accepted as per the DLPack protocol specification in the `Array API standard <https://data-apis.org/array-api/latest/API_specification/array_object.html?dlpack-self-stream-none#dlpack-self-stream-none>`_.
```

leofang (dismissed):
LGTM!

leofang:
Jenkins, test this please

cupy/_core/core.pyx
```diff
@@ -238,11 +238,22 @@ cdef class ndarray:
...
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
                     f' 1, 2, or any larger value, but {stream} was provided')
             if curr_stream_ptr == 0:
+                if not _util.DLPACK_ACCEPT_STREAM_ZERO:
+                    raise ValueError(
+                        'stream 0 is passed by library XXX, which is not '
+                        'accepted per the DLPack protocol. Please report this '
+                        'problem to the upstream library, and in the meantime '
+                        'work around this by setting the env var '
+                        'CUPY_DLPACK_ACCEPT_STREAM_ZERO=1')
```
leofang, line 256: Sorry, this line (249) is wrong: We should check the `stream` argument, not CuP's current stream (`curr_stream_ptr`).
```suggestion
            if stream == 0:
                if not _util.DLPACK_ACCEPT_STREAM_ZERO:
                    raise ValueError(
                        'stream 0 is passed by library XXX, which is not '
                        'accepted per the DLPack protocol. Please report this '
                        'problem to the upstream library, and in the meantime '
                        'work around this by setting the env var '
                        'CUPY_DLPACK_ACCEPT_STREAM_ZERO=1')
            if curr_stream_ptr == 0:
```

leofang (changes requested):

leofang:
Jenkins, test this please

leofang:
CUDA 11.5 CI passed right on the time limit 😂

kmaehashi:
I'm sorry for coming late to the discussion here, but I'm against introducing this env var. What we should provide is a smooth experience between libraries that just work. I don't want to say "please set this CuPy env var because PyTorch has a bug", which imposes the responsibility to users and is bad messaging for promoting a new standard.

> Otherwise this would defeat the purpose of all the hard debates for removing 0 ...

I agree on this point, but as long as all producers are implemented correctly there's should be no issue. And the only known bad implementation is PyTorch which does not use PTDS.

> This is a workaround, not a permanent thing.

We can possibly remove this workaround before the CuPy v11 final release considering PyTorch releases are out every 3 months 😃

leofang:
> I don't want to say "please set this CuPy env var because PyTorch has a bug", which imposes the responsibility to users 

I agree, but

> and is bad messaging for promoting a new standard.

> And the only known bad implementation is PyTorch which does not use PTDS.

We need to be future-proof. Other libraries will catch up and some of them might mess up like PyTorch did in the first few iterations, and we won't know if each of them is on PTDS or not. 

I'd say while we don't wanna introduce something for others' fault, we also don't want CuPy to be blamed when something hard to debug happens. Plus, the way I see it, the error message is actually doing a good thing to the community: it prompts users to report a very clear bug for upstream to fix.

kmaehashi:
(Hmm maybe I'm an enthusiastic fan of the Postel's Law 😅)

> We need to be future-proof. Other libraries will catch up and some of them might mess up like PyTorch did in the first few iterations, and we won't know if each of them is on PTDS or not.

I agree with this. However, PyTorch/CuPy integration is a "real" use-case for us that users will likely start using immediately after CuPy v10 is out. I don't want to see many users defining `CUPY_DLPACK_ACCEPT_STREAM_ZERO=1` which will be a short-lived API (i.e., likely to be removed once PyTorch is fixed). If users get used to defining it everywhere, we cannot be a "future-proof".

How about emitting a warning like this, then?

```
Stream 0 is passed from a library that owns a tensor you are converting from; CuPy assumes it as a legacy default stream.
Please report this problem to the library as this violates the DLPack protocol.
```

emcastillo:
Changed to use warning, PTAL!

kmaehashi (dismissed):
LGTM!

emcastillo:
fixed flake8 😥

kmaehashi (dismissed):

cupy/_core/core.pyx
```diff
@@ -238,11 +238,20 @@ cdef class ndarray:
...
                 raise ValueError(
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
                     f' 1, 2, or any larger value, but {stream} was provided')
             if curr_stream_ptr == 0:
+                warnings.warn(
+                    'Stream 0 is passed from a library that owns a tensor you '
+                    'are converting from; CuPy assumes it as a legacy default '
+                    'stream. Please report this problem to the library as this'
+                    ' violates the DLPack protocol.')
                 curr_stream_ptr = runtime.streamLegacy
```
leofang, line 255: Looks like the previous fix was lost...
```suggestion
            if stream == 0:
                warnings.warn(
                    'Stream 0 is passed from a library that owns a tensor you '
                    'are converting from; CuPy assumes it as a legacy default '
                    'stream. Please report this problem to the library as this'
                    ' violates the DLPack protocol.')
            if curr_stream_ptr == 0:
                curr_stream_ptr = runtime.streamLegacy
```

leofang (changes requested):

leofang:
(Sorry, I pushed to Emilio's branch because I couldn't disable auto merge)

Jenkins, test this please

leofang:
Jenkins, test this please

cupy/_core/core.pyx
```diff
@@ -238,10 +238,20 @@ cdef class ndarray:
...
+                # meanwhile.
                 raise ValueError(
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
                     f' 1, 2, or any larger value, but {stream} was provided')
+            if stream == 0:
+                warnings.warn(
+                    'Stream 0 is passed from a library that owns a tensor you '
+                    'are converting from; CuPy assumes it as a legacy default '
+                    'stream. Please report this problem to the library as this'
+                    ' violates the DLPack protocol.')
```
kmaehashi, line 254:
```suggestion
                    ' violates the DLPack protocol.')
                stream = runtime.streamLegacy
```

kmaehashi (changes requested):
Maybe this is needed?

cupy/_core/core.pyx
leofang: Yes, we missed that! Thanks, Kenichi-san.

kmaehashi:
pfnCI, run Jenkins please.

cupy/_core/core.pyx
```diff
@@ -238,10 +238,21 @@ cdef class ndarray:
...
+                # bug in PyTorch that exports the default stream as 0, which
+                # renders the protocol unusable, we will accept a 0 value
+                # meanwhile.
                 raise ValueError(
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
                     f' 1, 2, or any larger value, but {stream} was provided')
+            if stream == 0:
+                warnings.warn(
+                    'Stream 0 is passed from a library that owns a tensor you '
+                    'are converting from; CuPy assumes it as a legacy default '
```
kmaehashi, line 252: My bad, the warning message was not correct... :bow:

```suggestion
                    'Stream 0 is passed from a library that you are converting to;'
                    ' CuPy assumes 0 as a legacy default '
```

kmaehashi:
Confirmed that warning is displayed as expected. (as reviewed the message needs to be fixed)

```py
>>> torch.utils.dlpack.from_dlpack(cupy.arange(10))
/home/maehashi/.pyenv/versions/local-3.9.5/lib/python3.9/site-packages/torch/utils/dlpack.py:61: UserWarning: Stream 0 is passed from a library that owns a tensor you are converting from; CuPy assumes it as a legacy default stream. Please report this problem to the library as this violates the DLPack protocol.
  dlpack = ext_tensor.__dlpack__(stream=stream.cuda_stream)
tensor([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], device='cuda:0')
```

torch: 1.10.0+cu113

emcastillo:
updated the warning message

cupy/_core/core.pyx
```diff
@@ -238,10 +238,21 @@ cdef class ndarray:
...
+                # bug in PyTorch that exports the default stream as 0, which
+                # renders the protocol unusable, we will accept a 0 value
+                # meanwhile.
                 raise ValueError(
                     f'On CUDA, the valid stream for the DLPack protocol is -1,'
                     f' 1, 2, or any larger value, but {stream} was provided')
+            if stream == 0:
+                warnings.warn(
+                    'Stream 0 is passed from a library that you are converting to;'
+                    ' CuPy assumes 0 as a legacy default '
```
kmaehashi, line 252: flake8
```suggestion
                    'Stream 0 is passed from a library that you are'
                    ' converting to; CuPy assumes 0 as a legacy default '
```

PR #5981, opened 2021-11-01
Use index_t for faster address calculation

okuta:
Use `index_t` to calculate the address difference.
NVRTC generate faster code when `_use_32bit_indexing` is `True`


toslunar (approved):
LGTM

toslunar:
pfnCI, test this please.

PR #5994, opened 2021-11-04
Update upgrade guide for v10

### LINKED ISSUE #5818: Summarize `no-compat` PRs in v10 and update upgrade guide

takagi:
Close #5818.

### DIFF
--- a/docs/source/upgrade.rst
+++ b/docs/source/upgrade.rst
@@ -112,7 +112,7 @@ A new module :mod:`cupyx.profiler` is added to host all profiling related APIs i
 
 The old routines are deprecated.
 
-:func:`cupy.ndarray.__pos__` behaves samely as :func:`cupy.positive` instead of returning ``self``.
+:func:`cupy.ndarray.__pos__` now returns the copy (samely as :func:`cupy.positive`) instead of returning ``self``.
 
 Deprecated APIs may be removed in the future CuPy releases.
 
--- a/docs/source/upgrade.rst
+++ b/docs/source/upgrade.rst
@@ -28,8 +28,24 @@ Dropping NumPy 1.17 Support
 
 NumPy 1.17 is no longer supported.
 
-Changes in :class:`cupy.cuda.Stream` and :class:`cupy.cuda.Device` Behavior
----------------------------------------------------------------------------
+Change in :class:`cupy.cuda.Device` Behavior
+--------------------------------------------
+
+Current device set via ``use()`` will not be restored when exiting ``with`` block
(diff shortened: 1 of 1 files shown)


docs/source/upgrade.rst
```diff
@@ -13,18 +13,23 @@ Dropping CUDA 9.2 / 10.0 / 10.1 Support
...
 ---------------------------
 
 Python 3.6 is no longer supported.
 
-Changes in :class:`cupy.cuda.Stream` Behavior
----------------------------------------------
+Dropping NumPy 1.17 Support
+---------------------------
+
+NumPy 1.17 is not longer supported.
```
kmaehashi, line 29:
```suggestion
NumPy 1.17 is no longer supported.
```

docs/source/upgrade.rst
```diff
@@ -13,18 +13,23 @@ Dropping CUDA 9.2 / 10.0 / 10.1 Support
...
+
+Changes in :class:`cupy.cuda.Stream` and :class:`cupy.cuda.Device` Behavior
```
kmaehashi, line 31: Could you split the Device and Stream section?
Device things are more important for users (compared to Stream things) so I don't want to mix them.

```suggestion
Changes in :class:`cupy.cuda.Device` and :class:`cupy.cuda.Stream` Behavior
```

docs/source/upgrade.rst
```diff
@@ -91,8 +109,16 @@ A new module :mod:`cupyx.profiler` is added to host all profiling related APIs i
 
 The old routines are deprecated.
 
+:func:`cupy.cuda.Device.use` now returns ``self`` to get aligned with :func:`cupy.cuda.Stream.use()`.
```
kmaehashi, line 112: I think non-breaking changes need not be covered here. (otherwise, we have to cover most of the changes 😄 )

docs/source/upgrade.rst
```diff
@@ -91,8 +109,16 @@ A new module :mod:`cupyx.profiler` is added to host all profiling related APIs i
...
+
+:func:`cupy.ndarray.__pos__` behaves samely as :func:`cupy.positive` instead of returning ``self``.
```
kmaehashi, line 114:
```suggestion
:func:`cupy.ndarray.__pos__` now returns the copy (samely as :func:`cupy.positive`) instead of returning ``self``.
```

kmaehashi (changes requested):
Thanks!

docs/source/upgrade.rst
takagi: Okay! I didn't think well that the change does not affect users' existing codes...

takagi:
PTAL!

kmaehashi (approved):
Thanks!

kmaehashi:
Will merge after pre-review and RTD pass.

PR #5997, opened 2021-11-04
Support CUDA 11.5 on documents

takagi:


kmaehashi:
Confirmed rendered correctly on RTD.

PR #6007, opened 2021-11-05
Fix cuTENSOR installation on Windows

kmaehashi:
Follows-up #5996.
Detected by CI failure: https://ci.preferred.jp/cupy-release-tools.win/83625/

Confirmed `python install_library.py --library cutensor --cuda 11.4` runs on WIndows.

PR #6009, opened 2021-11-05
Fix archived directory name for cuTENOR in library installer

takagi:


takagi:
Duplicate fix as #6007.

PR #6012, opened 2021-11-05
Add universal CuPy package

kmaehashi:
This is a RFC PR to introduce universal CuPy package.

CuPy have been providing separate binary packages for each CUDA/ROCm version, but this is making downstream packages difficult to declare a dependency to CuPy.
This PR aims to improve the situation by introducing a new meta PyPI package named `cupy-wheel`, that dynamically declares dependency to one of the "real" CuPy wheel (`cupy-cudaXXX` or `cupy-rocm-X-X`) by looking into CUDA/ROCm installed in the users' environment.
Downstream packages can declare dependency (in `install_requires` or `requiements.txt`) like `cupy-wheel==9.*`.

### DIFF
--- a/install/universal_pkg/setup.py
+++ b/install/universal_pkg/setup.py
@@ -1,5 +1,6 @@
 import ctypes
 import pkg_resources
+import os
 import sys
 from typing import List, Optional
 
@@ -214,14 +215,19 @@ def infer_best_package() -> str:
 #
 
 def main():
-    package = infer_best_package()
-    requires = f'{package}=={VERSION}'
-    _log(f'Installing package: {requires}')
+    if os.environ.get('CUPY_UNIVERSAL_PKG_BUILD', None) is None:
+        package = infer_best_package()
+        requires = f'{package}=={VERSION}'
+        _log(f'Installing package: {requires}')
+        install_requires = [requires]
+    else:
+        _log('Building cupy-wheel package for release.')
+        install_requires = []
 
     setup(
         name='cupy-wheel',
         version=VERSION,
-        install_requires=[requires],
+        install_requires=install_requires,
     )
 
 
--- a/install/universal_pkg/setup.py
+++ b/install/universal_pkg/setup.py
@@ -88,7 +88,7 @@ def _get_cuda_version() -> Optional[int]:
     if sys.platform == 'linux':
         libnames = ['libcudart.so']
     elif sys.platform == 'win32':
-        libnames = ['cudart64.dll']
+        libnames = ['cudart64_110.dll', 'cudart64_102.dll']
     else:
         _log(f'CUDA detection unsupported on platform: {sys.platform}')
         return None
--- a/install/universal_pkg/setup.py
+++ b/install/universal_pkg/setup.py
@@ -214,7 +214,7 @@ def infer_best_package() -> str:
 # Entrypoint
 #
 
-def main():
+def main() -> None:
     if os.environ.get('CUPY_UNIVERSAL_PKG_BUILD', None) is None:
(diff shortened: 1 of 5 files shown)


leofang:
cc: @jakirkham, I don't think setting up `install_requires`/`python_requires`/etc would have any impact to conda recipes, but I'd like to confirm 🙂

jakirkham:
Maybe I'm missing something here (and please feel free to correct me if so), but we are not creating the `cupy-wheel` package for Conda as well, right?

leofang:
No, just wanna make sure the stuff added here would not interfere with conda 🙂

jakirkham:
Thanks for checking Leo 🙂 Don't think so

kmaehashi:
This `setup.py` script is intended for a separate `cupy-wheel` PyPI package, so this does not affect conda-forge distribution 😄

install/universal_pkg/setup.py
```diff
@@ -0,0 +1,226 @@
...
+
+    if sys.platform == 'linux':
+        libnames = ['libcudart.so']
+    elif sys.platform == 'win32':
+        libnames = ['cudart64.dll']
+    else:
+        _log(f'CUDA detection unsupported on platform: {sys.platform}')
+        return None
+    _log(f'Trying to detect CUDA version from candidates: {libnames}')
+    version = _get_version_from_library(libnames, 'cudaRuntimeGetVersion')
```
leofang, line 94: A subtlety here: This relies on the installing machine to have a GPU. For a reason I can't recall now, we chose to look up `CUDA_VERSION` instead, but with hindsight it created more problems that it solved, so this is better 👍 (Also as a reminder to self that I need to fix it...)
https://github.com/NVIDIA/cuQuantum/blob/8bd2f77bb6a03ba49b69e8706c9d5fd94752b95c/python/setup.py#L107-L108

leofang (commented):
This would be nice to have for cuQuantum Python to depend on CuPy. 

My question: would `cupy-wheel` be distributed only as sdist in order to do the runtime env check?

kmaehashi:
Thanks for the feedback Leo, we are targeting this to be released in the next month's release : https://github.com/cupy/cupy/issues/6246

> My question: would `cupy-wheel` be distributed only as sdist in order to do the runtime env check?

Yes.

emcastillo (approved):

emcastillo:
Fails for 10.2 in windows

```
    def test_get_cuda_version():
>       assert setup._get_cuda_version() == cupy.cuda.runtime.runtimeGetVersion()
E       AssertionError: assert None == 10020
```

emcastillo:
Fails for 10.2 in windows

```
    def test_get_cuda_version():
>       assert setup._get_cuda_version() == cupy.cuda.runtime.runtimeGetVersion()
E       AssertionError: assert None == 10020
```

install/universal_pkg/setup.py
```diff
@@ -0,0 +1,235 @@
...
+    """Returns the library version from list of candidate libraries."""
+
+    for libname in libnames:
+        try:
+            _log(f'Looking for library: {libname}')
+            runtime_so = ctypes.CDLL(libname)
+            break
+        except Exception as e:
+            _log(f'Failed to open {libname}: {e}')
+            return None
```
emcastillo, line 64:
```suggestion
            return None
```

This will cause the first one failing to return an error.
If we pass multiple I think we don't want this.

kmaehashi: Thanks!

leofang:
Seems to be working well 🎉 🎉 🎉

PR #6016, opened 2021-11-05
CI: Fix ccache not working

kmaehashi:
This was a regression in #5881.

When src directory is copied, compiler command line always include a random path, preventing ccache to reuse the cache.

This PR sets the source root as a ccache basedir to let ccache omit the dir from hash calculation.

https://ccache.dev/manual/4.4.2.html#_configuration_options

> base_dir (CCACHE_BASEDIR)
> ...
> ccache will rewrite absolute paths into paths relative to the current working directory, but only absolute paths that begin with base_dir. Cache results can then be shared for compilations in different directories even if the project uses absolute paths in the compiler command line.


toslunar (approved):

toslunar:
Let's merge this now (without tests), because the change would not be destructive and the CI queues are long.

PR #5988, opened 2021-11-02
Use SHA1 hash for kernel cache key to support Linux in FIPS-compliant mode

kmaehashi:
Closes #5959.

I chose SHA1 based on the quick micro benchmark:

```py
>>> a = ('*' * 4096).encode()
>>> timeit('hashlib.md5(a).hexdigest()', globals=globals())
6.068204950541258
>>> timeit('hashlib.sha1(a).hexdigest()', globals=globals())
4.463794406503439
>>> timeit('hashlib.sha256(a).hexdigest()', globals=globals())
9.106148925609887
```

Under Python 3.6.8 on CentOS 8 stream with FIPS-compilant mode enabled:

```
>>> hashlib.md5(b"foo").hexdigest()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS

>>> hashlib.md5(b"foo", usedforsecurity=False).hexdigest()
'acbd18db4cc2f85cedef654fccc4a4d8'

>>> hashlib.sha1(b"foo").hexdigest()
'0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33'

>>> hashlib.sha1(b"foo", usedforsecurity=False).hexdigest()
'0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33'
```

(seems `usedforsecurity` is available in any versions of Python installed via RHEL/CentOS 7/8 RPM)

PR #6023, opened 2021-11-06
Fix hash length for SHA1

kmaehashi:
Regression in #5988. Kernel cache was not used at all because hexdigest length changed from 32 (MD5) to 40 (SHA1).
This PR removes the magic number.

When writing a kernel cache file, CuPy prepends a checksum (hexdigest hash string) before the actual cubin data. However, the checksum length was hard-coded to 32, causing checksum validation to always fail.

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -271,6 +271,9 @@ def _hash_hexdigest(value):
     return hashobj.hexdigest()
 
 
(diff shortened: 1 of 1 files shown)


kmaehashi:
pfnCI, run Jenkins please.

kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, run Jenkins please.

emcastillo (dismissed):
Good catch!

PR #6027, opened 2021-11-07
CI: Use `ccache` in Pre-review Test

kmaehashi:
This reduces the build time from 3~4 minutes to ~1 minute.

First run: https://github.com/cupy/cupy/runs/4130095842 (3m9s)
Second run: https://github.com/cupy/cupy/runs/4130153202 (46s)


emcastillo (approved):
LGTM!

emcastillo:
Merging as we don't require CIs

Issue #5837, opened 2021-10-05
Support Python 3.10


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read Release Notes
- [x] Ask fastrlock to release wheels (e.g., https://github.com/scoder/fastrlock/issues/11)
- [x] Update FlexCI variants to use test against new Python version. (note: Some Python libraries may not be supported with the newer versions of Python. In that case you should also limit the library version.) https://github.com/cupy/cupy/pull/5899
- [x] Wheel distribution
  - [x] Update cupy-release-tools. https://github.com/cupy/cupy-release-tools/pull/193
  - [x] Add new Python matrix to the Jenkins job configuration (other_cupy_release_builder).
- [x] Documentation
  - [x] Update Requirements section in [Installation Guide](https://docs.cupy.dev/en/latest/install.html) #6038 
- [x] Update CuPy codebase.
  - [x] Update package Classifier in `setup.py`. #6038
  - [ ] Backport. (no backport this time -- v9 only supports Python ~3.9)

Related task: #4983 (distutils)

kmaehashi:
Let's proceed with this after NumPy / SciPy wheels with the official Py 3.10 support is out. Milestone set to v11 (will be backported to v10.)

dvolgyes:
As far as I see CuPy is not yet supporting python 3.10, and Scipy 1.7.3 is out with python 3.10 support since late November, so shouldn't be this ticket open?

kmaehashi:
@dvolgyes Python 3.10 support has been added in v10.0.0. Let us know if you have any issues.

dvolgyes:
I am sorry, I overlooked that. I will test it out.

PR #4711, opened 2021-02-23
Update Python / NumPy / SciPy requirements

kmaehashi:
Closes #4095

PR #6038, opened 2021-11-07
Add Python 3.10.0 to support list

kmaehashi:
Part of #5837


toslunar (approved):
LGTM

Note: #4711 added Python 3.9 to the lists.

PR #6039, opened 2021-11-07
CI: allow running tests selectively

kmaehashi:
This PR adds a feature to run FlexCI tests selectively, based on the tags assigned to each FlexCI project.

For example:

* `/test mini` to run tests tagged with `mini`
* `/test example,rapids` to run tests tagged with `example` or `rapids`

`/test` runs all tests regardless of tags.

We may also need to discuss review criteria (e.g., what kind of PRs can be merged without "all" tests).

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -60,6 +60,7 @@ configs {
       seconds: 21600
     }
     environment_variables { key: "GPU" value: "2" }
+    environment_variables { key: "TAGS" value: "all,mini" }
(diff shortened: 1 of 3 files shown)


emcastillo:
doing `/test cupy.linux.cuda102` will trigger this one only?

Thanks

kmaehashi:
> doing /test cupy.linux.cuda102 will trigger this one only?

No, currently tag-based filter only, butI can add that if it's needed 😃

emcastillo (approved):
LGTM!

PR #6045, opened 2021-11-08
Support cuSPARSELt 0.2.0

takagi:


cupy_backends/cuda/libs/cusparselt.pyx
```diff
@@ -19,18 +19,20 @@ cdef extern from '../../cupy_cusparselt.h' nogil:
 
     # Opaque Data Structures
     ctypedef struct cusparseLtHandle_t 'cusparseLtHandle_t':
-        uint8_t data[1024]
+        uint8_t data[11024]
```
takagi, line 22: The size of opaque structures has increased. Should we make a branch to keep the original value for <0.2.0?

leofang: I think you should just do something like this:
```python
ctypedef struct cusparseLtHandle_t 'cusparseLtHandle_t':
    pass
```
since the `data` field is not supposed to be used. But, while this would make the wrapper cleaner, you still need some `IF/ELSE` macros if you wanna support multiple ABI incompatible versions...

kmaehashi: I think we don't have to keep supporting cuSPARSElt 0.1.0 or earlier, right?

* Let's use `pass` as Leo suggested.
* Currently wheels are built without cuSPARSELt , ABI compatibility needs not to be cared for.
* Please update FlexCI matrix so that cuSPARSELt 0.2.0 can be tested with this PR.

takagi: Thanks! I'm going to fix it. (As for the data sizes, Cython itself does not need the sizes of the structure, but we still need them here to allocate space for the structures as `Handle` and `MatDescriptor` do.)

leofang: IIUC we just need to expose the struct names (cusparseLtHandle_t etc) and not their members to Cython if they are unused. It's enough to make sizeof(cusparseLtHandle_t) return the correct size.

takagi: Sorry, I must not omit `pass` to make Cython recognize them complete types. 😢

leofang: I see, thanks for the update, lesson learned 😓

kmaehashi:
Let's merge this after RC, then backport to v10. I'll include `We are going to change the minimum required cuSPARSELt version to v0.2.0 in v10 final release.` in rc release notes.

toslunar:
The CI config LGTM.

/test

kmaehashi:
Merging, Jenkins failures unrelated.

PR #6015, opened 2021-11-05
Fix `testing.multi_gpu` to add pytest marker

kmaehashi:
Closes #5912.

Now tests can be run selectively using `pytest -m pytest -m "multi_gpu"` or `pytest -m pytest -m "not multi_gpu"`.

PR #6050, opened 2021-11-08
CI: use separate project for multi-GPU tests

kmaehashi:
Recently we are hitting GPUs quota in FlexCI GCP and causing the delay in CI pipeline.

I'm thinking of separating multipe GPU tests to separate FlexCI project.
Currently unit tests use only one GPU except for few tests marked `multi_gpu`, but we are reserving 2 GPUs for whole time.
This should also reduce the total cost.

This PR depends on `multi_gpu` marker added in #6015.


toslunar:
Besides #6098, `cupyx_tests/distributed_tests/test_store.py::TestTCPStore` is failing, but I've not reproduced it (by `CUDA_VISIBLE_DEVICES=0`).

kmaehashi:
Tried to fix in #6114.
/test mini

kmaehashi:
Test passed 😄

kmaehashi:
Test failures are known ones @takagi is currently working on  so I think it's ok to merge this one.

PR #6056, opened 2021-11-09
Fix notifier to work on Python 3.6

kmaehashi:


kmaehashi:
Flake8 passed, let me merge and test it in FlexCI :bow:

PR #6066, opened 2021-11-11
Fix CI result notification message format

kmaehashi:
The current message was too verbose...


kmaehashi:
Merging to improve the Slack notification message, let's discuss if needed 😃

PR #6072, opened 2021-11-11
Exclude `kernel_version` from comparison table

kmaehashi:
~This hides several APIs that are unexpectedly exposed, by using `__all__`.~

~https://gist.github.com/kmaehashi/8e89e6cfd863a4bc0746b90c3467b795~

~Depends on #6071~

See discussion below.

### DIFF
--- a/docs/source/_comparison_generator.py
+++ b/docs/source/_comparison_generator.py
@@ -8,8 +8,9 @@ _footnotes = {}
 
 
 def _get_functions(obj, exclude=None):
+    children = obj.__all__ if hasattr(obj, '__all__') else dir(obj)
     return set([
-        n for n, target in [(n, getattr(obj, n)) for n in dir(obj)]
+        n for n, target in [(n, getattr(obj, n)) for n in children]
         if (
             # not in exclude list
             (exclude is None or n not in exclude)
--- a/docs/source/_comparison_generator.py
+++ b/docs/source/_comparison_generator.py
@@ -8,9 +8,8 @@ _footnotes = {}
 
 
 def _get_functions(obj, exclude=None):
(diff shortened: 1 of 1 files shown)


docs/source/_comparison_generator.py
```diff
@@ -8,8 +8,9 @@
 
 
 def _get_functions(obj, exclude=None):
+    children = obj.__all__ if hasattr(obj, '__all__') else dir(obj)
```
toslunar, line 11: `__all__` can contain "private" names that start with `_`.  How about moving the condition `not n.startswith('_')` here? (For example, which I admit not really a good example, `'__version__' in numpy.__all__`).

I'm not sure if default `__all__` contains submodule names, but it looks always fine to exclude submodules from the comparison table.

```suggestion
    try:
        children = obj.__all__
    except AttributeError:
        children = [n for n in dir(obj) if not n.startswith('_')]
```

toslunar (changes requested):
I'd like to preserve `abs`, `max`, `min`, and `round`.  NumPy excludes the names in builtins from `numpy.__all__`.  They are shorthand of `absolute`, `amax`, `amin`, and `round_`, respectively.
```
>>> {n: type(v).__name__ for n in sorted(set(dir(np)).difference(np.__all__)) if not n.startswith('_') and not inspect.ismodule(v := getattr(np, n))}
{'Tester': 'type', 'abs': 'ufunc', 'kernel_version': 'tuple', 'max': 'function', 'min': 'function', 'numarray': 'str', 'oldnumeric': 'str', 'round': 'function', 'test': 'PytestTester', 'use_hugepage': 'int'}
>>> np.abs is np.absolute
True
```

kmaehashi:
Makes sense, fixed 😃

toslunar (approved):

PR #6074, opened 2021-11-11
Borrow more non-GPU APIs from NumPy

kmaehashi:


cupy/__init__.py
```diff
@@ -309,6 +309,10 @@ def is_available():
 from cupy._manipulation.rearrange import roll  # NOQA
 from cupy._manipulation.rearrange import rot90  # NOQA
 
+# Borrowed from NumPy
+if hasattr(_numpy, 'broadcast_shapes'):  # NumPy 1.20
+    from numpy import broadcast_shapes  # NOQA
```
toslunar, line 314: We have `cupy._core.internal._broadcast_shapes`.  Our implementation could be removed once NumPy<1.20 is dropped and we check the performance of `numpy.broadcast_shapes`.

toslunar (approved):

PR #6075, opened 2021-11-11
Add more aliases for compatibility with NumPy

kmaehashi:


toslunar (commented):
Aliases are different functions in NumPy (for docstrings, it says).  But I'd rather not pay the additional maintenance cost for `cupy.all is not cupy.alltrue` etc.  What do you think?

https://github.com/numpy/numpy/blob/v1.21.4/numpy/core/fromnumeric.py#L3727-L3728

kmaehashi:
> Aliases are different functions in NumPy (for docstrings, it says). But I'd rather not pay the additional maintenance cost for cupy.all is not cupy.alltrue etc. What do you think?

Agree. And all 4 functions added in this PR are undocumented in NumPy so I don't think we have to take care of docstrings. 😄

toslunar (approved):

PR #6076, opened 2021-11-11
Import more dtype aliases from NumPy

kmaehashi:


toslunar (approved):
LGTM

cupy/__init__.py
```diff
@@ -398,6 +374,11 @@ def result_type(*arrays_and_dtypes):
 from numpy import mintypecode  # NOQA
 from numpy import sctype2char  # NOQA
 from numpy import typename  # NOQA
+from numpy import maximum_sctype  # NOQA
```
toslunar, line 377:
```
>>> numpy.maximum_sctype(float)
<class 'numpy.float128'>
```
in my environment. Should `cupy.maximum_sctype(float)` return `float64`?  (But it'll be surprising for `cupy.maximum_sctype(numpy.float128)` to return `float64`.)

kmaehashi: Thansk! Removed ones that cannot be imported directly and added to #6078.

toslunar (approved):
LGTM

The PR title would be "Import more dtype aliases from NumPy".

kmaehashi:
Fixed!

PR #6077, opened 2021-11-11
Borrow indexing APIs from NumPy

kmaehashi:


cupy/__init__.py
```diff
@@ -436,6 +436,14 @@ def result_type(*arrays_and_dtypes):
 
 from cupy._indexing.iterate import flatiter  # NOQA
 
+# Borrowed from NumPy
+from numpy import get_array_wrap  # NOQA
+from numpy import index_exp  # NOQA
+from numpy import ndindex
+from numpy import s_  # NOQA
+from numpy import mask_indices  # NOQA
+from numpy import tril_indices  # NOQA
```
toslunar, line 445: `mask_indices` and `tril_indices` return tuple of arrays.  Maybe GPU arrays are expected.

kmaehashi: Thanks! Yes that was the thing what I was not confident and that's why this is marked WIP. Will fix tomorrow!

kmaehashi:
```suggestion
```

cupy/__init__.py
```diff
@@ -436,6 +436,12 @@ def result_type(*arrays_and_dtypes):
 
 from cupy._indexing.iterate import flatiter  # NOQA
 
+# Borrowed from NumPy
+from numpy import get_array_wrap  # NOQA
+from numpy import index_exp  # NOQA
+from numpy import ndindex
```
kmaehashi, line 442:
```suggestion
from numpy import ndindex  # NOQA
```

toslunar (approved):

cupy/__init__.py
kmaehashi: Added them to #6078

PR #6082, opened 2021-11-12
Add polynomial modules to comparison table

toslunar:


kmaehashi:
Thanks! Confirmed RTD rendered fine :smile:
https://cupy--6082.org.readthedocs.build/en/6082/reference/comparison.html#polynomials

PR #6088, opened 2021-11-13
[DO NOT MERGE] Add wrappers for `nvPTXCompiler` APIs

leofang:


leofang:
@emcastillo @kmaehashi This PR was done a while ago and refreshed as a parallel effort with [RAPIDS/ptxcompiler](https://github.com/rapidsai/ptxcompiler) for Numba. 

I marked this PR as "do not merge" because of the build/distribution requirement. If nvPTXCompiler is built with CTK `11.x`, it is backward compatible with all CTK `11.y` (`y<=x`) seen at runtime but not `y>x` due to [the PTX ISA versioning being bumped for every 11.y release](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes). So if we wanna merge this PR we need to apply either of the two suggestions:
1. Always build this module using the latest CTK available at the time (might need nontrivial changes to the existing build pipeline)
2. Build this module for every CTK >= 11.1, and add a runtime check to `nvPTXCompiler.create()` to ensure it's not used with a newer CTK.

To make it compatible with the way conda-forge utilizes CUDA Enhanced Compatibility, Option 2 is needed.

emcastillo:
Sorry for the long time to check this,

I think 2 is reasonable since we always build wheels for every minor version (we want to change this once CUDA python is mature enough). However I understand that in Conda, updating CTK may not trigger a CuPy update and could break the stuff, and this makes the check needed.

kmaehashi:
> To make it compatible with the way conda-forge utilizes CUDA Enhanced Compatibility, Option 2 is needed.

What CUDA version is used to build CuPy packages on conda-forge?

leofang:
To be clear I added this just for completeness and fun/curiosity. I currently do not have use cases and the RAPIDS team has moved on with their own wrapper 🙂

> What CUDA version is used to build CuPy packages on conda-forge?

On conda-forge, CuPy 10.x is built with CUDA 10.2, 11.0, 11.1 and 11.2. The 11.2-built package has a relaxed pin on `cudatoolkit` such that any `cudatoolkit` 11.x >= 11.2,<12 can be used at runtime; other packages are pinned to the build-time CTK version. 

This would make the wrapper suffer from the lack of support of newer PTX (from 11.3+). We've seen some other issues (ex: https://github.com/conda-forge/nccl-feedstock/issues/62) that may eventually prompt the bump of `nvcc` version on conda-forge (say to 11.6). We can also consider dropping some CTK versions.

leofang:
Closing this PR as CUDA 12 offers nvJitLink that supersedes nvPTXCompiler.
https://developer.nvidia.com/blog/cuda-12-0-compiler-support-for-runtime-lto-using-nvjitlink-library/

PR #6117, opened 2021-11-18
Improve FlexCI test scripts

kmaehashi:
* Retry gsutil download as cache download may fail during other test uploading cache.
* Do `gcloud auth` outside of `run.sh` script to avoid configuring twice. (this is harmless but verbose)
* Improve message (show size of cache file, make docker push quiet).


kmaehashi:
Confirmed the output, let me merge to test it in branch.

PR #6125, opened 2021-11-18
Define HOME in Dockerfile

kmaehashi:
It is difficult to debug CI interactively using shell mode (`./.pfnci/linux/run.sh cuda114 shell`) because ``HOME`` environment variable is not set.


kmaehashi:
Moved to #9308

PR #6127, opened 2021-11-18
Support all advanced indexing

toslunar:
In `__getitem__`, `__setitem__`, and scatter ops, allow mask index to be combined with other masks or arrays.


toslunar:
Shape of masks should be checked.  For example, `cupy.arange(12).reshape(3, 4)[[1, 0], [True, True]]` should raise `IndexError`.
```
>>> numpy.arange(12).reshape(3, 4)[[1, 0], [True, True]]
IndexError: boolean index did not match indexed array along dimension 1; dimension is 4 but corresponding boolean dimension is 2
```

asi1024:
Jenkins, test this please.

toslunar:
Updated the branch because it seems CI config has been updated.

asi1024 (approved):

PR #6128, opened 2021-11-18
Fix cuSPARSELt version in manually crafted Dockerfiles

kmaehashi:
Follows-up: #6045


kmaehashi:
Let me merge to prevent further branch test failures. These are not covered by CI.

PR #6131, opened 2021-11-19
Add `cupy.array_api.linalg`

leofang:
Applied diff from numpy/numpy#19980.


leofang:
@asmeurer Please help verify this is the last piece we need to catch up with NumPy 🙂

asmeurer:
I don't know enough about cupy to do a heavy review here. Most of the linalg tests are implemented in the test suite, however. I would say the main thing to check is `solve`, as I had to copy and modify the internal NumPy implementation to work around the stacked vectors issue.

leofang:
Thanks for heads-up @asmeurer! I fixed `solve()` as CuPy cannot reuse the NumPy private routines.

kmaehashi (approved):

PR #6142, opened 2021-11-21
Update footnotes in comparison table

kmaehashi:


docs/source/_comparison_generator.py
```diff
@@ -118,6 +118,7 @@ def _section(
 _float_error_state = (
     'Floating point error handling depends on CPU-specific features'
     ' which is not available in GPU.')
+_byte_order = 'Not supported as GPUs only support little-endian computations.'
```
emcastillo, line 121:
```suggestion
_byte_order = 'Not supported as GPUs only support little-endian byte-encoding.'
```

emcastillo (approved):

emcastillo:
Merging as docs rendered successfully

https://cupy--6142.org.readthedocs.build/en/6142/reference/comparison.html

PR #6144, opened 2021-11-22
CI: Need to update CUDA driver in cuda115.multi

kmaehashi:


kmaehashi:
Merging, confirmed `pfn-public-ci/cupy.linux.cuda115.multi` passed!

PR #6149, opened 2021-11-22
Remove `__all__` from `cupyx/scipy/*`

### LINKED ISSUE #6070: Remove `__all__` in `cupyx.scipy.*` modules
`__all__` is not used in CuPy codebase.

Ref: #6071

Yutaro-Sanada:
fixes #6070 

- ```__all__``` is not used. Then delete it.

### DIFF
--- a/cupyx/scipy/fft/__init__.py
+++ b/cupyx/scipy/fft/__init__.py
@@ -1,5 +1,12 @@
 # flake8: NOQA
 from cupyx.scipy.fft._fft import (
+    fft, ifft, fft2, ifft2, fftn, ifftn,
+    rfft, irfft, rfft2, irfft2, rfftn, irfftn,
+    hfft, ihfft, hfft2, ihfft2, hfftn, ihfftn,
+    fftshift, ifftshift, fftfreq, rfftfreq,
+    get_fft_plan
+)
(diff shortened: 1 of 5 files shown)


kmaehashi:
Thanks @Yutaro-Sanada! Could you update `cupyx/scipy/linalg/__init__.py`, so that the set of exposed APIs (i.e., functions exposed via `from cupyx.scipy.linalg import *`) are not changed before/after the `__all__` removal? Same for `cupyx/scipy/fft/__init__.py`.

kmaehashi:
Thanks for the fix! Now that `__all__` is removed, the following lines must be removed to avoid exposing unexpected modules (e.g., `cupyx.scipy.fft.np` should not be visible).

`cupyx/scipy/fft/__init__.py`:

```py
# flake8: NOQA
from cupyx.scipy.fft._fft import *
```

`cupyx/scipy/linalg/__init__.py`: 

```py
from cupyx.scipy.linalg.special_matrices import *  # NOQA
```

Could you remove them and instead explicitly import public APIs?

Yutaro-Sanada:
@kmaehashi Sorry for late. I fixed, then please review it.

kmaehashi (approved):
Thanks! LGTM :+1:

PR #6152, opened 2021-11-23
delete from os import path

### LINKED ISSUE #3885: Eliminate `from os import path` from `cupy_setup_build.py`
Let's use `os.path` instead.

Yutaro-Sanada:
fixes #3885
 
- change ```from os import path``` to ```os.path```

### DIFF
--- a/install/cupy_builder/cupy_setup_build.py
+++ b/install/cupy_builder/cupy_setup_build.py
@@ -7,7 +7,6 @@ from distutils import msvccompiler
 from distutils import sysconfig
 from distutils import unixccompiler
 import os
-from os import path
 import shutil
 import sys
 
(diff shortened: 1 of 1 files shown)


emcastillo:
Thank you!
Instead of using the archaic `os.path`, how about moving to `pathlib`?
Some places in the code base are already using it :)

kmaehashi:
I think `distutils` does not expect path-like objects for the most part, so if we use `pathlib` in these scripts we'll need to do cast everywhere. I think this PR resolves the original issue #3885 and it's fine to merge as-is 😃

emcastillo:
Good catch! thanks!

emcastillo:
Jenkins, test this please

emcastillo (approved):

Issue #6078, opened 2021-11-11
[Tracker] Implement all `numpy.*` APIs in CuPy


kmaehashi:
Implement GPU version of `numpy.*` functions in `cupy.*` namespace.

This is a tracker issue that lists the remaining `numpy.*` APIs (see also: [comparison table](https://docs.cupy.dev/en/latest/reference/comparison.html#module-level)). I've categorized them based on difficulty (comments welcomed :smile:) so that new contributors can pick the right task. Your contribution is highly welcomed and appreciated!

### Very Easy

- [ ] `numpy.asfarray`
- [ ] `numpy.byte_bounds`
- [ ] `numpy.format_float_positional` (just call NumPy function like `cupy.array2string`)
- [ ] `numpy.format_float_scientific` (just call NumPy function like `cupy.array2string`)
- [ ] `numpy.loads`
- [ ] `numpy.ndfromtxt`

### Easy

- [ ] `numpy.array_equiv`
- [ ] `numpy.asarray_chkfinite` (note: likely to require synchronization)
- [ ] `numpy.fabs`
- [ ] `numpy.float_power`
- [ ] `numpy.heaviside`
- [ ] `numpy.isneginf`
- [ ] `numpy.isposinf`
- [ ] `numpy.ndenumerate`
- [ ] `numpy.real_if_close` (note: likely to require synchronization)
- [ ] `numpy.tril_indices_from` (just wrap NumPy function)

These are also considered easy tasks but relatively low priority:

- [ ] `numpy.deprecate`
- [ ] `numpy.deprecate_with_doc`
- [ ] `numpy.info`
- [ ] `numpy.lookfor`
- [ ] `numpy.source`

### Medium

- [ ] `numpy.apply_over_axes`
- [ ] `numpy.block`
- [ ] `numpy.cast` (note: undocumented API in `numpy/core/numerictypes.py`)
- [ ] `numpy.delete`
- [ ] `numpy.einsum_path`
- [ ] `numpy.frompyfunc` (maybe just call GUFunc?)
- [ ] `numpy.geomspace`
- [ ] `numpy.insert`
- [ ] `numpy.nested_iters`
- [ ] `numpy.nditer`
- [ ] `numpy.put_along_axis`
- [ ] `numpy.row_stack`
- [ ] `numpy.set_numeric_ops`
- [ ] `numpy.spacing`
- [ ] `numpy.vander`

### Medium to High

- [ ] `numpy.histogram_bin_edges`
[...]

PR #6156, opened 2021-11-24
Add `cupy.mask_indices`

shwina:
As part of https://github.com/cupy/cupy/issues/6078, adding the `mask_indices` API.

### DIFF
--- a/cupy/_indexing/generate.py
+++ b/cupy/_indexing/generate.py
@@ -426,6 +426,43 @@ def unravel_index(indices, dims, order='C'):
     return tuple(unraveled_coords)
 
 
+def mask_indices(n, mask_func, k=0):
+    """
+    Return the indices to access (n, n) arrays, given a masking function.
+
+    Assume `mask_func` is a function that, for a square array a of
+    size ``(n, n)`` with a possible offset argument `k`, when called
+    as ``mask_func(a, k)`` returns a new array with zeros in certain
+    locations (functions like `triu` or `tril` do precisely
+    this). Then this function returns the indices where the non-zero
+    values would be located.
+
+    Args:
+        n (int): The returned indices will be valid to access arrays
+            of shape (n, n).
+        mask_func (callable): A function whose call signature is
+            similar to that of `triu`, `tril`.  That is,
+            ``mask_func(x, k)`` returns a boolean array, shaped like
+            `x`.  `k` is an optional argument to the function.
+        k (scalar): An optional argument which is passed through to
+            `mask_func`. Functions like `triu`, `tril` take a second
+            argument that is interpreted as an offset.
+
+    Returns:
+        tuple of arrays: The `n` arrays of indices corresponding to
+        the locations where ``mask_func(np.ones((n, n)), k)`` is
+        True.
+
+    .. warning::
+
+        This function may synchronize the device.
+
+    .. seealso:: :func:`numpy.mask_indices`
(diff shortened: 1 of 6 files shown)


cupy/_sorting/search.py
```diff
@@ -139,6 +139,41 @@ def nonzero(a):
     _util.check_array(a, arg_name='a')
     return a.nonzero()
 
+def mask_indices(n, mask_func, k=0):
```
shwina, line 142: Is this the right module for this function? I placed it here because of the similarities to  `nonzero`.

leofang: Hi @shwina maybe this is the file to go:
https://github.com/cupy/cupy/blob/c0c5f9e4b7e8693ce6a435276d28e21ae3a55bf5/cupy/_indexing/generate.py#L435

shwina: 👍 Thanks!

tests/cupy_tests/indexing_tests/test_generate.py
```diff
@@ -289,3 +289,22 @@ def test_invalid_mode(self, dtype):
             a = tuple([xp.arange(min(dims), dtype=dtype) for d in dims])
             with pytest.raises(ValueError):
                 xp.ravel_multi_index(a, dims, mode='invalid')
+
+
+@testing.gpu
+class TestMaskIndices:
+
+    @testing.numpy_cupy_array_equal()
+    def test_mask_indices(self, xp):
```
shwina, line 298: Note: I tried to parametrize this test over all dtypes, but I believe `cupyx.sparse` (which is  used to generate the data for this test) supports only float and complex types. Is it worth parametrizing over dtypes and generating the input data in a different way?

cupy/_indexing/generate.py
```diff
@@ -426,6 +426,43 @@ def unravel_index(indices, dims, order='C'):
...
+        the locations where ``mask_func(np.ones((n, n)), k)`` is
+        True.
+
+    .. warning::
+
+        This function may synchronize the device.
+
+    .. seealso:: :func:`numpy.mask_indices`
+    """
+    a = cupy.ones((n, n))  # TODO what dtype to use here?
```
shwina, line 462: For memory savings, should we use `int8` or `bool` here?

takagi:
Would you also add `mask_indices` to `docs/source/reference/indexing.rst` so that the  document of `mask_indices` is generated?

cupy/_indexing/generate.py
takagi: Yes, `int8` suits here.

takagi (commented):
Thanks for your pull request! The code is almost fine, and would you check some comments?

tests/cupy_tests/indexing_tests/test_generate.py
shwina: Thanks! I used your idea  of a boolean array as a multiplier to generate a square random matrix of 50% density (exactly  what we were doing before)

shwina:
Thanks for reviewing @takagi! I've addressed your comments. Please let me know if you think this PR needs anything else!

takagi (approved):

takagi:
Thanks! Let me push a fix to your repo as it is a minor change, and now I've kicked the CI. Your PR is going to be merged soon after CI passes.

Issue #6139, opened 2021-11-21
Compilation time is inconsistent between different environments


mlazzarin:
### Description

I'm working on a project that uses CuPy to accelerate quantum computing simulations.
We employ CuPy with custom CUDA kernels loaded with RawModule.
We are trying to benchmark our code to assess the JIT approach with respect to alternatives.
During benchmarks, we found out that the compilation times are inconsistent between different versions of CuPy and CUDA.

For example, I've written a gist https://gist.github.com/mlazzarin/1e2128e90d78c4cb1a220075f64bc297 that loads our custom kernels with ``cp.RawModule`` and compile them with ``.compile()`` method.
I tried to run such example with different versions of CuPy and CUDA toolkit:

| Environment                                                                             | Compilation time |
| -------------------------------------------------------------------------------- | ---------------------- |
| System installation of CUDA 11.5, ``cupy-cuda115`` from pip | ~ 3.2 s                 |
| ``cudatoolkit=11.5.0`` and ``cupy=9.6.0`` from conda-forge    | ~ 3.2 s                 |
| ``cudatoolkit=11.4.2`` and ``cupy=9.6.0`` from conda-forge    | ~ 3.2 s                 |
| ``cudatoolkit=11.3.1`` and ``cupy=9.6.0`` from conda-forge    | ~ 3.2 s                 |
| ``cudatoolkit=11.2.2`` and ``cupy=9.6.0`` from conda-forge    | ~ 3.4 s                 |
| ``cudatoolkit=11.1.1`` and ``cupy=9.6.0`` from conda-forge    | ~ 1.6 s                 |
| ``cudatoolkit=11.5.0`` and ``cupy=9.5.0`` from conda-forge    | ~ 2.1 s  (3.2 s first exec) |
| ``cudatoolkit=11.4.2`` and ``cupy=9.5.0`` from conda-forge    | ~ 2.1 s  (3.2 s first exec) |
| ``cudatoolkit=11.3.1`` and ``cupy=9.5.0`` from conda-forge    | ~ 2.1 s  (3.2 s first exec) |
| ``cudatoolkit=11.2.2`` and ``cupy=9.5.0`` from conda-forge    | ~ 2.3 s  (3.5 s first exec) |
| ``cudatoolkit=11.1.1`` and ``cupy=9.5.0`` from conda-forge    | ~ 1.0 s  (1.7 s first exec) |

[...]

Issue #4926, opened 2021-03-20
Improve file-based kernel cache


kmaehashi:
* Often `~/.cupy/kernel_cache` inode becomes very big. I think it's better to introduce a hierarchy e.g. using the first 2 chars from the cache key: `~/.cupy/ab/abXXXXXXXXX....cubin`.
* Avoid appending MD5 to a file. The reason this is introduced was to support Windows that needs special care to do atomic rename (https://github.com/cupy/cupy/pull/37/files).
  `os.replace` is in Python 3 so we can just use it now.
  (FYI, on Windows `os.replace` uses `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` flag which can be considered atomic in our usecase; there is a long thread in the [forum](https://social.microsoft.com/Forums/partner/ja-JP/449bb49d-8acc-48dc-a46f-0760ceddbfc3/movefileexmovefilereplaceexisting-ntfs-same-volume-atomic?forum=windowssdk).)
* Include some hint info to filename to manage cache files easily. (maybe NVRTC version and compute capability)

leofang:
> * Often `~/.cupy/kernel_cache` inode becomes very big.

I've been wondering about this. I would be interested in exploring solutions like periodically invalidating outdated caches (any update of CUDA or CuPy or CC versions could trigger this). We could also do it a bit aggressively, say purging caches older than a month. So far I've been cleaning this manually, and it would be great if there's a mechanism doing a better job than II do...😄 

The hierarchy approach is not mutually exclusive to this auto invalidation, nor does it address the inode inflation problem entirely.

kmaehashi:
How about adding this tool https://github.com/cupy/cupy/blob/master/.pfnci/trim_cupy_kernel_cache.py to `cupyx/tools`?

> Include some hint info to filename to manage cache files easily. (maybe NVRTC version and compute capability)

Also including a CuPy's major version ('8', '9', etc.) to filename may help managing caches.

leofang:
> How about adding this tool https://github.com/cupy/cupy/blob/master/.pfnci/trim_cupy_kernel_cache.py to `cupyx/tools`?

Yeah, something like this would likely help. Do users need to run this manually? Is it cross platform (i.e. support Windows)?

kmaehashi:
> Do users need to run this manually?

Yes, users need to manually run the tool. Once we have a hint tag in filename we can add some more convenient filters (e.g. purge all caches for non-current CUDA versions).
[...]

PR #6159, opened 2021-11-24
Add caching for nvrtc kernels with name_expressions

keckj:
This is PR to enable caching RawKernel/RawModule compiled with name expressions.

With this patch, by using #6139 gist as a benchmark, I get the following results (Ubuntu 20.04, Cuda 11.4):
- 2.7s on the first run
- 0.23s on the following runs (2.7s without this patch)

Currently, this uses restricted pickling to dump name_expressions along with the binary. It could also be extended to HIP.

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -277,6 +277,7 @@ def _hash_hexdigest(value):
 
 
 _hash_length = len(_hash_hexdigest(b''))  # 40 for SHA1
+_pickle_protocol = pickle.HIGHEST_PROTOCOL  # depends on actual python version
 
 
 class RestrictedUnpickler(pickle.Unpickler):
@@ -284,8 +285,8 @@ class RestrictedUnpickler(pickle.Unpickler):
 
     def find_class(self, module, name):
         """Only allow a restricted set of classes."""
-        if module == __name__ and name == 'CachedCudaModule':
-            return CachedCudaModule
+        if module == __name__ and name == 'CachedModule':
+            return CachedModule
         msg = "'{}.{}' is forbidden and cannot be unpickled."
         raise pickle.UnpicklingError(msg.format(module, name))
 
@@ -295,22 +296,20 @@ class RestrictedUnpickler(pickle.Unpickler):
         return cls(io.BytesIO(s)).load()
 
 
-class CachedCudaModule:
-    __slots__ = ['backend', 'cubin', 'mapping']
+class CachedModule:
+    __slots__ = ['backend', 'binary', 'mapping']
 
-    _pickle_protocol = 4  # first introduced in Python 3.4
-
-    def __init__(self, backend, cubin, mapping):
+    def __init__(self, backend, binary, mapping):
         self.backend = backend
-        self.cubin = cubin
+        self.binary = binary
         self.mapping = mapping
 
     @classmethod
-    def dump_module(cls, path, backend, cubin, mapping):
-        """Build and dumps a CachedCudaModule object and checksum to path."""
-        python_obj = cls(backend, cubin, mapping)
+    def dump_module(cls, path, backend, binary, mapping):
+        """Build and dumps a CachedModule object and checksum to path."""
+        python_obj = cls(backend, binary, mapping)
 
-        data_bytes = pickle.dumps(python_obj, protocol=cls._pickle_protocol)
+        data_bytes = pickle.dumps(python_obj, protocol=_pickle_protocol)
         data_hash = _hash_hexdigest(data_bytes).encode('ascii')
         assert len(data_hash) == _hash_length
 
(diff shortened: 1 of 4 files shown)


cupy/cuda/compiler.py
```diff
@@ -276,6 +278,93 @@ def _hash_hexdigest(value):
...
+        data_hash = _hash_hexdigest(data_bytes).encode('ascii')
+        assert len(data_hash) == _hash_length
+
+        directory = os.path.dirname(path)
+        if not os.path.isdir(directory):
+            os.makedirs(directory, exist_ok=True)
+
+        with open(path, 'wb') as f:
+            f.write(data_hash)
+            f.write(data_bytes)
```
leofang, line 322: This needs to follow the original approach (write to a temporary location and then move). The rationale there was that in HPC environments there could be thousands of processes writing to the same file (if they all need the same kernel). Writing directly to the destination would significantly increases the chances of file corruption, effectively leading to cache invalidation.

In fact, to support cuFFT callbacks we've adopted atomic moves using `os.rename()` (`os.replace()` is also OK), but it's only atomic on Linux. @kmaehashi maybe we can use it here too (once the temporary location is restored as an extra safety layer)?

leofang (commented):
Thanks, @keckj! I skimmed over it and looks good. Let me know when you are done and I can review again. Left one issue I found.

cupy/cuda/compiler.py
```diff
@@ -276,6 +278,93 @@ def _hash_hexdigest(value):
...
+    @classmethod
+    def loads(cls, s):
+        """Helper function analogous to pickle.loads()."""
+        return cls(io.BytesIO(s)).load()
+
+
+class CachedCudaModule:
+    __slots__ = ['backend', 'cubin', 'mapping']
+
+    _pickle_protocol = 4  # first introduced in Python 3.4
```
leofang, line 300: One more thing: Could you make the default to 5 for Python 3.8+? I guess we would likely drop Python 3.7 support for CuPy v11 as NumPy has dropped it, so it'd be nice to adopt the newer (and more performant) protocol.

But, to do so the value of `_pickle_protocol` must be added as part of the file hash to avoid issues when multiple Python environments exist locally.

cupy/cuda/compiler.py
kmaehashi: Use of `os.replace()` (but not `os.rename()`) sounds good to me. xref #4926

cupy/cuda/compiler.py
```diff
@@ -531,36 +629,27 @@ def _compile_with_cache_cuda(
         base = _preprocess('', options, arch, backend)
         _empty_file_preprocess_cache[env] = base
 
-    key_src = '%s %s %s %s' % (env, base, source, extra_source)
-    key_src = key_src.encode('utf-8')
-    name = _hash_hexdigest(key_src) + '.cubin'
-
-    mod = function.Module()
+    key = (env, base, source, extra_source, _pickle_protocol)
+    key_str = ('{} '*len(key)).format(*key)
```
kmaehashi, line 633: Let's use faster method :)
```suggestion
    key = (env, base, source, extra_source, str(_pickle_protocol))
    key_str = ' '.join(key)
```

cupy/cuda/compiler.py
```diff
@@ -274,6 +276,102 @@ def _hash_hexdigest(value):
...
+
+        # We use a temp file to handle concurrent processes caching
+        # to the same path at the same time.
+        with tempfile.NamedTemporaryFile(
+                dir=directory, delete=False, mode='wb') as tf:
+            tf.write(data_hash)
+            tf.write(data_bytes)
+            temp_path = tf.name
+
+        # os.replace is not atomic operation, so it could result in a
```
kmaehashi, line 327: `os.replace` is an atomic operation, so checksum in cache file is no longer needed 😄  (see #4926)

kmaehashi (changes requested):

leofang (changes requested):

cupy/cuda/compiler.py
leofang: Yup when I left the comment to @keckj I forgot about this. @kmaehashi, do you suggest to revise the comment here, or to remove the hash prefix entirely?

kmaehashi: I think it's ok to entirerly remove hash prefix.

kmaehashi:
This would be really helpful for `name_expression` users, thank you for your work @keckj!

cupy/cuda/compiler.py
keckj: I removed the hash. What should happen if mod.load(cached_module.binary) fails ?

kmaehashi: Thanks! I think it's safe to assume that the file is not broken.

kmaehashi (dismissed):
Thanks, LGTM!

keckj:
Also, I just realized that current caching key only depends on raw cuda source given by the user (and not full cuda source after running the preprocessor). This means that if an included header is modified somewhere on the system, cupy may return a cached module that does not correspond to actual sources.

leofang:
I think that's fine because the header contents are part of the hash for creating the cubin name, but @kmaehashi can confirm.

kmaehashi:
@keckj is correct, headers bundled with CuPy are included to cache key, but other headers `#include`d by raw kernels are not. NVRTC does not provide a way to run preprocess-only so it is difficult to solve this issue...

kmaehashi:
Hmm Windows CI failing at `os.replace` under high load. I will take a look more later.

```
PermissionError: [WinError 5] Access is denied: 'C:\\Windows\\system32\\config\\systemprofile\\.cupy\\kernel_cache\\tmp1tpbhrds' -> 'C:\\Windows\\system32\\config\\systemprofile/.cupy/kernel_cache\\3872abba2747ec448c3c517594d652f6237da675.cubin'
```

keckj:
NVRTC still do not provide built-in option to run only the C preprocessor, we still have the problem of included headers source change for custom user kernels caching.

Issue #6151, opened 2021-11-22
Use JIT compiler to back `cupy.fromfunction()`


leofang:
### Description

#5925 added a bunch of NumPy-compliant functions, including `cupy.fromfunction()`. However, it is suboptimal as `numpy.fromfunction` is called internally. 

It'd be nice to take the same JIT capability behind `cupy.vectorize()` and make it work with `fromfunction`. However, a common usage of `fromfunction` [seems to involve lambda functions](https://numpy.org/doc/stable/reference/generated/numpy.fromfunction.html), which CuPy does not support yet (#4290). Not sure if this is a blocker, though.

leofang:
A potential tradeoff here is compilation speed vs data transfer speed.

asi1024:
CuPy JIT needs to support device function call firstly.

leofang:
> However, we cannot (maybe will not also in the future) support lambda functions as the `inspect.getsource` cannot get lambda function as the target function.

Thanks, @asi1024. Wasn't it partially worked around by #5071?

kmaehashi:
> However, we cannot (maybe will not also in the future) support lambda functions as the inspect.getsource cannot get lambda function as the target function.

How about walking through the AST generated from `inspect.getsource` and using the first lambda found? (I think it's acceptable to raise an error when two lambdas are found in the AST).

asi1024:
@leofang @kmaehashi 
Yes, it was partially worked around by #5071. `inspect.get_source` gets the source line that includes the lambda function, so the users need to feed a better get_source function that finds the lambda function from the line, and set the function to `jit._getsource`.
I agree to support a better `get_source` function that finds the lambda function in a heuristic way as the default function of `jit._getsource` in CuPy.

asi1024:
I am planning the support of lambda function input in `cupy.vectorize`. as the first step :smiley:

PR #6170, opened 2021-11-26
Support lambda function in `cupy.vectorize`

asi1024:
https://github.com/cupy/cupy/issues/6151#issuecomment-979891362


cupyx/jit/_preprocess.py
```diff
@@ -0,0 +1,47 @@
+def _strip_lambda(source):
```
kmaehashi, line 1: * How about walking an AST as suggested in my original comment, instead of relying on heuristics?

```py
def get_lambda_node(x):
    lines = linecache.getlines(inspect.getsourcefile(x))
    _, lineno = inspect.getsourcelines(x)
    lambdas = [n for n in ast.walk(ast.parse(''.join(lines))) if isinstance(n, ast.Lambda) and n.lineno == lineno]
    if len(lambdas) != 1: raise RuntimeError
    return lambdas[0]
```

* Maybe better to memorize the result?

cupyx/jit/_compile.py
```diff
@@ -67,11 +68,16 @@ def transpile(func, attributes, mode, in_types, ret_type):
...
-
+    func_name = func.__name__
     attributes = ' '.join(attributes)
     source = jit._getsource_func(func)
+
+    if func_name == '<lambda>':
+        # Program transformation from lambda function to normal function
+        func_name = '_lambda_func'
+        args, source = _preprocess._strip_lambda(source)
+        source = _preprocess._make_function_str(func_name, args, source)
```
kmaehashi, line 79: How about transforming `ast.Lambda` to `ast.FunctionDef` directly, instead of string manipulation?

```py
ast.FunctionDef(func_name, lambda_node.args, [ast.Return(lambda_node.body)], [])
```

tests/cupy_tests/functional_tests/test_vectorize.py
```diff
@@ -615,3 +615,19 @@ def my_func(x1, x2, x3):
...
+    def test_vectorize_lambda(self, xp):
+        f = xp.vectorize(lambda a, b, c: a + b * c)
+        x1 = testing.shaped_random((20, 30), xp, numpy.int64, seed=1)
+        x2 = testing.shaped_random((20, 30), xp, numpy.int64, seed=2)
+        x3 = testing.shaped_random((20, 30), xp, numpy.int64, seed=3)
+        return f(x1, x2, x3)
+
+    def test_vectorize_lambda_xfail(self):
+        functions = [lambda a, b: a + b, lambda a, b: a * b]
+        f = cupy.vectorize(functions[0])
```
leofang, line 629: Why do we expect an error with a single lambda function?

asi1024: In this testcase, the result of `inspect.getsource` is `functions = [lambda a, b: a + b, lambda a, b: a * b]`, not `lambda a, b: a + b`. So we cannot detect which lambda function is the target.

leofang: Ahh interesting, thanks for explanation @asi1024!

cupyx/jit/_compile.py
asi1024: Fixed in 3af94fc.

cupyx/jit/_compile.py
```diff
@@ -54,6 +55,53 @@ def new_func(node, *args, **kwargs):
...
+        filename = inspect.getsourcefile(func)
+        with open(filename) as f:
+            full_source = f.read()
+        source, start_line = inspect.getsourcelines(func)
+        end_line = start_line + len(source)
+        source = ''.join(source)
+
+    tree = ast.parse(full_source)
+
+    def search(node, instance):
```
kmaehashi, line 77: Let's use `ast.walk` instead :)

cupyx/jit/_compile.py
```diff
@@ -54,6 +55,53 @@ def new_func(node, *args, **kwargs):
...
+    # Parses function object into ast.FunctionDef object.
+    if not callable(func):
+        raise ValueError('`func` must be a callable object.')
+
+    if jit._getsource_func is not None:
+        full_source = jit._getsource_func(func)
+        start_line, end_line = 0, math.inf
+        source = full_source
+    else:
+        filename = inspect.getsourcefile(func)
```
kmaehashi, line 68: Needs error handling:
* TypeError may be raised here.
* filename can be None if func is C extension function.

cupyx/jit/_compile.py
```diff
@@ -54,6 +55,53 @@ def new_func(node, *args, **kwargs):
...
+        with open(filename) as f:
+            full_source = f.read()
+        source, start_line = inspect.getsourcelines(func)
+        end_line = start_line + len(source)
+        source = ''.join(source)
+
+    tree = ast.parse(full_source)
+
+    def search(node, instance):
+        if isinstance(node, instance) and start_line <= node.lineno < end_line:
```
kmaehashi, line 78:
```suggestion
        if isinstance(node, instance) and (start_line is None or start_line == node.lineno):
```

cupyx/jit/_compile.py
```diff
@@ -54,6 +55,53 @@ def new_func(node, *args, **kwargs):
...
 
 
+def _parse_function_object(func):
+    # Parses function object into ast.FunctionDef object.
+    if not callable(func):
+        raise ValueError('`func` must be a callable object.')
+
+    if jit._getsource_func is not None:
+        full_source = jit._getsource_func(func)
+        start_line, end_line = 0, math.inf
```
kmaehashi, line 65:
```suggestion
        start_line = None
```

cupyx/jit/_compile.py
```diff
@@ -54,6 +55,53 @@ def new_func(node, *args, **kwargs):
...
+        source = full_source
+    else:
+        filename = inspect.getsourcefile(func)
+        with open(filename) as f:
+            full_source = f.read()
+        source, start_line = inspect.getsourcelines(func)
+        end_line = start_line + len(source)
```
kmaehashi, line 72:
```suggestion
```

kmaehashi (changes requested):
Great, LGTM except for nits!

cupyx/jit/_compile.py
asi1024: When a single line decorator is attached to the target function, the value of node.line_no will be `start_line + 1`.

kmaehashi: Ah I see! Thanks for explanation 😄

cupyx/jit/_compile.py
asi1024: Do you mean that we need to fallback to inspect.getsource?

kmaehashi: No, I meant that proper error handling is needed here. 😄 
```suggestion
        try:
            filename = inspect.getsourcefile(func)
        except TypeError:
            filename = None
        if filename is None:
            raise ValueError(f'JIT needs access to Python source for {func}'
                             'but could not be located')
```

cupyx/jit/_compile.py
```diff
@@ -54,6 +55,44 @@ def new_func(node, *args, **kwargs):
...
+        source = ''.join(source)
+
+    tree = ast.parse(full_source)
+
+    def get_node(instance):
+        nodes = [node for node in ast.walk(tree)
+                 if isinstance(node, instance)
+                 and start_line <= node.lineno < end_line]
+        if len(nodes) > 1:
+            raise RuntimeError('Parse error: multiple function is found.')
```
kmaehashi, line 82:
```suggestion
            raise ValueError('Multiple callables are found near the'
                             f' definition of {func}, and JIT could not'
                             ' identify the source code for it.')
```

asi1024:
Fixed! PTAL!

kmaehashi (approved):
LGTM!

kmaehashi:
Tests seems failing, could you check? @asi1024

cupyx/jit/_compile.py
```diff
@@ -60,6 +60,19 @@ def _parse_function_object(func):
...
         raise ValueError('`func` must be a callable object.')
 
+    if func.__name__ != '<lambda>':
+        if jit._getsource_func is None:
+            lines = inspect.getsource(func).split('\n')
+            num_indent = len(lines[0]) - len(lines[0].lstrip())
+            source = '\n'.join([
+                line.replace(' ' * num_indent, '', 1) for line in lines])
+        else:
+            source = jit.getsource(func)
```
kmaehashi, line 70:
```suggestion
            source = jit._getsource_func(func)
```

kmaehashi (approved):

PR #6172, opened 2021-11-29
Handles infinities of the same sign in `logaddexp` and `logaddexp2`

### LINKED ISSUE #5716: Incorrect output of `cupy.logaddexp()`
For this case, mathematically we should get `inf`, but CuPy returns `nan`:
```python
>>> np.logaddexp(np.inf, np.inf)
inf
>>>
>>> cp.logaddexp(np.inf, np.inf)
array(nan)

```
The reason is `in0-in1` gives `nan` when both are `inf`, and it propagates all the way out:
https://github.com/cupy/cupy/blob/4469fae998df33c72ff40ef954cb08b8f0004b18/cupy/_math/explog.py#L73

takagi:
Fix #5716.


toslunar (approved):
LGTM

PR #6174, opened 2021-11-29
CI: Update limits to reduce cache size

kmaehashi:
Further limit cache size to reduce upload/download/creation/extraction time.

* Default limit of ccache is 5GB but it was too large to keep CuPy build cache.
* Also adjust CuPy kernel cache size limit.


emcastillo (approved):
LGTM!

PR #6179, opened 2021-11-30
Remove memory copy in matmul

okuta:
The current implementation incurs extra memory consumption and memory copy in matmul operation. This PR solves it.

### DIFF
--- a/cupy/_core/_routines_linalg.pyx
+++ b/cupy/_core/_routines_linalg.pyx
@@ -726,7 +726,15 @@ cpdef ndarray matmul(ndarray a, ndarray b, ndarray out=None):
 
     ndim = max(orig_a_ndim, orig_b_ndim)
     if ndim <= 2:
-        return dot(a, b, out)
+        if out is None or out._c_contiguous:
+            return dot(a, b, out)
+        ret_dtype = numpy.promote_types(a.dtype, b.dtype)
+        if out._c_contiguous and ret_dtype == out.dtype:
+            return dot(a, b, out)
+        c = _ndarray_init(out._shape, dtype=ret_dtype)
+        dot(a, b, c)
+        elementwise_copy(c, out)
+        return out
 
     orig_a = a
     orig_b = b
--- a/cupy/_core/_routines_linalg.pyx
+++ b/cupy/_core/_routines_linalg.pyx
@@ -449,7 +449,7 @@ cpdef ndarray tensordot_core(
         out = _ndarray_init(ret_shape, dtype)
     else:
         if out.dtype != dtype:
-            out = _ndarray_init(ret_shape, dtype)
+            raise NotImplementedError("The out array dtype is mismatched")
     cdef int ace
     if m == 1 and n == 1:
         for ace in _accelerator._routine_accelerators:
@@ -712,10 +712,6 @@ cpdef ndarray matmul(ndarray a, ndarray b, ndarray out=None):
 
     """
 
-    if out is not None:
-        raise NotImplementedError('The out array as input is currently not '
-                                  'supported')
-
     cdef Py_ssize_t i, n, m, ka, kb, a_sh, b_sh, c_sh
     cdef Py_ssize_t batchCount, a_part_outshape, b_part_outshape
     cdef int orig_a_ndim, orig_b_ndim, a_ndim, b_ndim, ndim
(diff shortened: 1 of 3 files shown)


cupy/_core/_routines_linalg.pyx
```diff
@@ -817,10 +813,26 @@ cpdef ndarray matmul(ndarray a, ndarray b, ndarray out=None):
...
+        else:
+            out.fill(0)
+            return out
 
-    out = ndarray(out_shape, dtype=dtype)
+    ret = None
+    if out is None:
+        out = ndarray(out_shape, dtype=dtype)
+    elif not out.flags.c_contiguous:
+        raise ValueError('Output array must be C-contiguous')
```
toslunar, line 830: NumPy accepts non-contiguous `out`.  Copy to `out` later.

cupy/_core/_routines_linalg.pyx
```diff
@@ -817,10 +813,26 @@ cpdef ndarray matmul(ndarray a, ndarray b, ndarray out=None):
...
+    elif out.dtype != ret_dtype:
+        raise ValueError('Output array has incorrect dtype')
```
toslunar, line 832: NumPy accepts other dtypes, while `casting` option is checked.

cupy/_core/_routines_linalg.pyx
```diff
@@ -449,7 +449,7 @@ cpdef ndarray tensordot_core(
         out = _ndarray_init(ret_shape, dtype)
     else:
         if out.dtype != dtype:
-            out = _ndarray_init(ret_shape, dtype)
+            raise NotImplementedError("The out array dtype is mismatched")
```
toslunar, line 452: I agree that the current implementation, which ignores `out`, seems wrong.  I'm not sure if the change is better than commenting `# TODO: Fix to write to out`.

cupy/linalg/_product.py
```diff
@@ -10,7 +10,8 @@
 from cupy.linalg import _util
 
 _gu_func_matmul = _GUFunc(
-    _core.matmul, '(n?,k),(k,m?)->(n?,m?)', supports_batched=True)
+    _core.matmul, '(n?,k),(k,m?)->(n?,m?)', supports_batched=True,
+    supports_out=True)
```
toslunar, line 14: Yes. This is necessary to eliminate copy operations.  The general `out` support in `cupy._core._gufuncs._GUFunc` cannot know C-contiguous output is assumed at the cublas call in `cupy._core.matmul`.

toslunar (changes requested):

cupy/linalg/_product.py
okuta: I think that there is no problem in normal usage. Do you think you need to take any measures?

cupy/_core/_routines_linalg.pyx
```diff
@@ -817,10 +814,26 @@ cpdef ndarray matmul(ndarray a, ndarray b, ndarray out=None):
...
+    ret = None
+    if out is None:
+        out = ndarray(out_shape, dtype=dtype)
+    elif not out.flags.c_contiguous:
+        raise NotImplementedError('Output array must be C-contiguous')
+    elif out.dtype != ret_dtype:
+        raise NotImplementedError('Output array has incorrect dtype')
+    elif dtype != ret_dtype:
+        ret = out
+        out = ndarray(out_shape, dtype=dtype)
```
toslunar, line 836:
```suggestion
    if out is not None and out.dtype == dtype and out.flags.c_contiguous:
        ret = None
    else:
        if out is None and dtype != ret_dtype:
            ret = ndarray(out_shape, dtype=ret_dtype)
        else:
            ret = out
        out = ndarray(out_shape, dtype=dtype)
```

Moreover, renaming (`out`, `ret`) to (`c`, `out`) may make the code more readable, i.e.
```python
    if out is not None and out.dtype == dtype and out.flags.c_contiguous:
        c = out
    else:
        c = ndarray(out_shape, dtype=dtype)
        if out is None:
            if dtype == ret_dtype:
                out = c
            else:
                out = ndarray(out_shape, dtype=ret_dtype)
...
    if out is not c:
        elementwise_copy(c, out)
```

toslunar (changes requested):

cupy/_core/_routines_linalg.pyx
okuta: OK!

toslunar (dismissed):
LGTM

cupy/_core/_routines_linalg.pyx
```diff
@@ -726,7 +726,15 @@ cpdef ndarray matmul(ndarray a, ndarray b, ndarray out=None):
 
     ndim = max(orig_a_ndim, orig_b_ndim)
     if ndim <= 2:
-        return dot(a, b, out)
+        if out is None or out._c_contiguous:
```
toslunar, line 729:
```suggestion
        if out is None:
```
because the condition `out._c_contiguous and ret_dtype == out.dtype` below is correct for `out is not None` cases.

toslunar (changes requested):

toslunar (approved):

cupy/_core/_routines_linalg.pyx
jakirkham: FWIW NumPy's `matmul` will just error in this case.

```python
import numpy as np

a = np.random.random((2, 3))
b = np.random.random((3, 2))
c = np.empty((2, 2), dtype=int)

np.matmul(a, b, out=c)
```

```python
UFuncTypeError                            Traceback (most recent call last)
<ipython-input-7-a4f34170f335> in <module>
      5 c = np.empty((2, 2), dtype=int)
      6 
----> 7 np.matmul(a, b, out=c)

UFuncTypeError: Cannot cast ufunc 'matmul' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

<br>

Note: `UFuncTypeError ` is [just a `TypeError` subclass]( https://github.com/numpy/numpy/blob/f4353323d0be0c3b3c7d0c752d6c135bc6d36465/numpy/core/_exceptions.py#L32 )

PR #6181, opened 2021-11-30
Improve copy and assign operation

okuta:
`__getitem__` method has some overhead. And, It finally calls `elemntwise_copy`.


toslunar (commented):
`dst[...] = src` and `cupy.copyto(dst, src, casting='unsafe')` utilize `dst.data.copy_from_async` if memcpy is possible.  I suppose a faster `cupy.copyto` (without shape/casting checks) is what we need.

okuta:
OK, I will rewrite.

okuta:
I checked `copyto`.  Basically, copying for `out` does not cause non-contiguous cases. So I think there is no problem with `elementwise_copy`.

What do you think?

cupyx/scipy/ndimage/morphology.py
```diff
@@ -257,7 +258,7 @@ def _binary_erosion(input, structure, iterations, mask, output, border_value,
                 break
         output = tmp_out
     if temp_needed:
-        temp[...] = output
+        _core.elemntwise_copy(output, temp)
```
grlee77, line 261:
```suggestion
        _core.elementwise_copy(output, temp)
```

okuta: Thanks

takagi (approved):

PR #6188, opened 2021-12-01
Deprecate MachAr (support NumPy 1.22)

kmaehashi:
Test with NumPy 1.22.0rc1 failed:
https://ci.preferred.jp/cupy.linux.cuda-head/87655/

This is because `np.MachAr` has been deprecated (https://github.com/numpy/numpy/pull/20201)


toslunar (approved):
LGTM

toslunar:
OK to backport this because v10 also requires Python >= 3.7.

PR #6218, opened 2021-12-07
Docs: CentOS installation from source

ptim0626:
A quick fix on a typo in the nvcc compiler flag and the `gcc` executable, as there is no `gcc-7` or similar in CentOS, unlike Ubuntu which all get symlinked. After installing the `devtoolset-7-gcc-c++` and activating those binaries, calling `gcc` will use the `gcc` at version 7 for compilation.


kmaehashi:
Thanks, LGTM!

PR #6222, opened 2021-12-08
Support comparison to `None` and `Ellipsis`

### LINKED ISSUE #5715: `cupy.ndarray` cannot be compared with ellipsis
Not sure if it's due to #4198. The `__richcmp__` method does not handle trivial comparison with Python objects correctly:
```python
>>> np.empty(3) == ...
array([False, False, False])
>>>
>>> cp.empty(3) == ...
  File "cupy/_core/core.pyx", line 1094, in cupy._core.core.ndarray.__richcmp__
    return numpy.equal(self, other)
TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'equal'>, '__call__', array([0., 0., 0.]), Ellipsis): 'ndarray', 'ellipsis'
```
It seems NumPy internally calls `PyObject_RichCompare` to handle this. Perhaps before delegating to `numpy.equal()` (and friends) we should handle this edge case first?

takagi:
Fix #5715.

This PR implements comparisons to Python's built-in constants, `None` and `Ellipsis`. It returns a ndarray filled with `True` or `False` with the same shape as the ndarray as if the operation is performed elementwise.

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -1112,6 +1112,12 @@ cdef class ndarray:
                 return _logic._ndarray_greater(self, other)
             if op == 5:
                 return _logic._ndarray_greater_equal(self, other)
+        elif other is None or other is Ellipsis:
+            if op == 2:
+                return cupy.zeros(self._shape, dtype=cupy.bool_)
+            if op == 3:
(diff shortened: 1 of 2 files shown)


toslunar (approved):
LGTM

PR #6257, opened 2021-12-20
Fixing index calculation for random constructor

WiseroOrb:
The previous algorithm had numerical issues with certain values, like m = 3150
Now it uses directly the integer division, without doing magic tricks


cupyx/scipy/sparse/construct.py
```diff
@@ -365,7 +365,7 @@ def random(m, n, density=0.01, format='coo', dtype=None,
         data_rvs = random_state.rand
 
     ind = random_state.choice(mn, size=k, replace=False)
-    j = cupy.floor(ind * (1. / m)).astype('i')
+    j = ind//m
```
emcastillo, line 368:
```suggestion
    j = ind // m
```

emcastillo (approved):

PR #6260, opened 2021-12-21
Tentatively pin to `setuptools<60` in Windows CI

kmaehashi:
`setuptools==60.*` and `numpy.distutils` from `numpy==1.21.*` conflicts on Windows, as setuptools now uses vendored distutils by default.

https://setuptools.pypa.io/en/latest/history.html#v60-0-0


asi1024 (approved):

kmaehashi:
This pin was removed in #6273

PR #6269, opened 2021-12-27
CI: allow discarding docker image cache manually

kmaehashi:
This fix allows regenerating CI docker images manually.
The script can be triggered via FlexCI Web UI <https://ci.preferred.jp/cupy.prep-linux/>.

Currently we pin dependent library versions using wildcard:
https://github.com/cupy/cupy/blob/master/.pfnci/linux/tests/cuda-head.Dockerfile#L26

However as CI docker images are cached without expiry, tests won't be able to newer versions of these libraries (e.g., newer RC releases of NumPy/SciPy).

### DIFF
--- a/.pfnci/linux/run.sh
+++ b/.pfnci/linux/run.sh
@@ -52,7 +52,7 @@ main() {
   cache_archive="linux-${TARGET}-${base_branch}.tar.gz"
   cache_gcs_dir="${CACHE_GCS_DIR:-gs://tmp-asia-pfn-public-ci/cupy-ci/cache}"
 
-  if [[ "${DOCKER_IMAGE_CACHE}" = "0" ]]; then
+  if [[ "${DOCKER_IMAGE_CACHE:-1}" = "0" ]]; then
     docker_cache_from=""
   fi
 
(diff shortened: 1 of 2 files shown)


.pfnci/linux/run.sh
```diff
@@ -46,9 +48,14 @@ main() {
   repo_root="$(cd "$(dirname "${BASH_SOURCE}")/../.."; pwd)"
   base_branch="$(cat "${repo_root}/.pfnci/BRANCH")"
   docker_image="${DOCKER_IMAGE:-asia.gcr.io/pfn-public-ci/cupy-ci}:${TARGET}-${base_branch}"
+  docker_cache_from="${docker_image}"
   cache_archive="linux-${TARGET}-${base_branch}.tar.gz"
   cache_gcs_dir="${CACHE_GCS_DIR:-gs://tmp-asia-pfn-public-ci/cupy-ci/cache}"
 
+  if [[ "${DOCKER_IMAGE_CACHE}" = "0" ]]; then
```
kmaehashi, line 55:
```suggestion
  if [[ "${DOCKER_IMAGE_CACHE:-1}" = "0" ]]; then
```

emcastillo (approved):

Issue #5722, opened 2021-09-09
Reorganizing build scripts


kmaehashi:
`cupy_setup_build.py` is now more than 1200 lines. There are lots of `if`s, global variables, and hacks around.
Before doing large changes like #4983 I'd like to do code fixes:

1. Reorganize them into a python package (`cupy_builder`) and split files.
1. Avoid using global variables to improve readability/maintainability/testability.

With this refactoring, I'm also expecting to:
* Make module configuration results cache-able (like autoconf `./configure`) to speed up try-and-error of CuPy build.
* Avoid running module configuration two times when doing `pip install -e .`. (this internally runs `python setup.py egg_info; python setup.py develop` so configuration is happening two times)

The first step would be to reorganize it into a package with only changing imports.
I'm thinking of changing the current file structure:

```
/setup.py
/cupy_setup_build.py
/install/build.py
/install/utils.py
```

as:

```
/setup.py
/install/cupy_builder/__init__.py
/install/cupy_builder/cupy_setup_build.py
/install/cupy_builder/install_build.py
/install/cupy_builder/install_utils.py
```

then I can split files under `cupy_builder` package into optimal modules.

leofang:
> * Avoid running module configuration two times when doing `pip install -e .`. (this internally runs `python setup.py egg_info; python setup.py develop` so configuration is happening two times)

Question: How is this avoided? Is there a flag from setuptools to set?

kmaehashi:
> Question: How is this avoided? Is there a flag from setuptools to set?

https://github.com/cupy/cupy/blob/485257874104e3efc5e71778afbff8cb0ef27fb9/setup.py#L85
This line runs configuration, but the list of extensions are not needed for generating `egg_info`. So I think we can just skip it if `sys.argv[1] == 'egg_info'`.

PR #6272, opened 2021-12-27
Reorganize build scripts, part 6: separate compilers and command

kmaehashi:
Context: #5722

Please review commit-by-commit.


toslunar (commented):
LGTM except for the flake8 failure

kmaehashi:
I thought this PR is just a code-fix, but it seems distutils import order is causing trouble on Windows.
Rather than debugging distutils internal details, I'll merge this PR with #6273 that eliminates the distutils monkeypatching, which should resolve the CI failure.

PR #3256, opened 2020-04-02
Rewrite tests not use `numpy_cupy_raises`

asi1024:
Ready for #3098.

PR #6275, opened 2021-12-27
Add `asarray_chkfinite` API.

khushi-411:
Hi, Team!
**Aim of the PR:** To implement `asarray_chkfinite` API.

It follows https://github.com/cupy/cupy/issues/6078#issue-1051191869.

Looking forward to your viewpoints. Thanks!

cc: @kmaehashi (:


cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,31 @@
 from cupy import _core
 
 
+def asarray_chkfinite(a, dtype=None, order=None):
+    """
+    Convert the given input to an array,
+    and raises an error if the input is NaNs or Infs.
```
toslunar, line 8:
```suggestion
    and raises an error if the input contains NaNs or Infs.
```

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,31 @@
...
+
+    Args:
+        a: array like.
+        dtype: data type, optional
+        order: {'C', 'F'}, optional
```
toslunar, line 13: It seems 'A' and 'K' are also supported.
https://numpy.org/doc/stable/reference/generated/numpy.asarray_chkfinite.html

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,31 @@
...
+
+    Returns:
+        out (cupy.ndarray): Array interpretation of `a`.
```
toslunar, line 16: Google style cannot name the returned value.
```suggestion
        cupy.ndarray: Array interpretation of `a`.
```

toslunar (changes requested):

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,13 @@
 @testing.gpu
 class TestKind(unittest.TestCase):
 
+    @testing.for_all_dtypes()
+    def test_asarray_chkfinite(self, dtype):
```
leofang, line 14: Could you add another test with a non-finite-value array? I am not sure if bool/int/complex types make sense here, so worth a look.

cupy/_manipulation/kind.py
khushi-411: Noted (:

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Hi, @leofang! 
I added the tests for the non-finite-value array, and all the tests are passed in the CI. Are more tests required?
> I am not sure if bool/int/complex types make sense here, so worth a look.

So, why did you think that was worth checking?
I checked for `numpy.asarray_chkfinite`. The following were the outputs: (which are as expected)

#### For bool
```
In [12]: a = numpy.asarray_chkfinite([numpy.inf, 0., -numpy.inf], dtype=bool)

In [13]: a
Out[13]: array([ True, False,  True])
```
#### For int
```
In [14]: a = numpy.asarray_chkfinite([numpy.inf, 0., -numpy.inf], dtype=int)
OverflowError                             Traceback (most recent call last)
<ipython-input-14-d6d91c379cab> in <module>
----> 1 a = numpy.asarray_chkfinite([numpy.inf, 0., -numpy.inf], dtype=int)

~/anaconda3/lib/python3.8/site-packages/numpy/lib/function_base.py in asarray_chkfinite(a, dtype, order)
    481 
    482     """
--> 483     a = asarray(a, dtype=dtype, order=order)
    484     if a.dtype.char in typecodes['AllFloat'] and not np.isfinite(a).all():
    485         raise ValueError(

~/anaconda3/lib/python3.8/site-packages/numpy/core/_asarray.py in asarray(a, dtype, order)
     81 
     82     """
---> 83     return array(a, dtype, copy=False, order=order)
     84 
     85 

OverflowError: cannot convert float infinity to integer
```
#### For float
```
In [15]: a = numpy.asarray_chkfinite([numpy.inf, 0., -numpy.inf], dtype=float)
<ipython-input-15-15536ad3bf83> in <module>
----> 1 a = numpy.asarray_chkfinite([numpy.inf, 0., -numpy.inf], dtype=float)

~/anaconda3/lib/python3.8/site-packages/numpy/lib/function_base.py in asarray_chkfinite(a, dtype, order)
    483     a = asarray(a, dtype=dtype, order=order)
    484     if a.dtype.char in typecodes['AllFloat'] and not np.isfinite(a).all():
--> 485         raise ValueError(
    486             "array must not contain infs or NaNs")
    487     return a

ValueError: array must not contain infs or NaNs
```
#### For Complex
```
a = numpy.asarray_chkfinite([numpy.inf, 0., -numpy.inf], dtype=complex)
<ipython-input-16-14e1c1a2ac61> in <module>
----> 1 a = numpy.asarray_chkfinite([numpy.inf, 0., -numpy.inf], dtype=complex)

~/anaconda3/lib/python3.8/site-packages/numpy/lib/function_base.py in asarray_chkfinite(a, dtype, order)
    483     a = asarray(a, dtype=dtype, order=order)
    484     if a.dtype.char in typecodes['AllFloat'] and not np.isfinite(a).all():
--> 485         raise ValueError(
    486             "array must not contain infs or NaNs")
    487     return a

ValueError: array must not contain infs or NaNs
```

khushi-411:
Hi, @leofang! Hi, @toslunar!
Thank you so much for reviewing up the PR! I made all the suggested changes. Please let me know if anything else is required. I would be happy to learn (:

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,20 @@
...
+        a = [0, 4, 0, 5]
+        a_gpu = cupy.asarray_chkfinite(a, dtype=dtype)
+        a_cpu = numpy.asarray_chkfinite(a, dtype=dtype)
+        assert a_gpu.dtype == a_cpu.dtype
+
+    @testing.for_all_dtypes()
+    def test_asarray_chkfinite_non_finite_vals(self, dtype):
+        a = [-numpy.inf, 0., numpy.inf]
+        a_gpu = cupy.asarray_chkfinite(a, dtype=dtype)
```
oke-aditya, line 23: I believe this is what you need @khushi-411 

```suggestion
        self.assertRaises(ValueError, cupy.asarray_chkfinite, a, dtype):
```

This will check that the expected ValueError is raised when you are passing an array with infinite values.

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,20 @@
...
+        a_cpu = numpy.asarray_chkfinite(a, dtype=dtype)
```
oke-aditya, line 24: I think this too will raise an error and you won't get anything in `a_cpu`. As nothing would be returned.
I'm not sure you need to test this or even write this.

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,20 @@
...
+        assert a_gpu.dtype == a_cpu.dtype
```
oke-aditya, line 25: Hence from above comment checking the dtype wouldn't make sense.

oke-aditya (commented):
Little help as you asked offline. 

To other reviewers: -
@khushi-411 pinged me offline and asked for a small look. I thought I could help a bit.
I haven't contributed to CuPy but have experience in PyTorch

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Yes, `a_cpu` too will raise value error.

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Thanks a lot, @oke-aditya, for taking a look (: It worked.
But I am not too sure if we need this test or not.

oke-aditya: I believe you need the test. It is recommended to test the valueErrors as this is an expected behivour.

Any case the maintainers will tell you :)

tests/cupy_tests/manipulation_tests/test_kind.py
leofang: > and all the tests are passed in the CI

Just FYI: actual GPU tests are not yet run, the green CIs are just preliminary checks (formatting, compilation, etc).

> > I am not sure if bool/int/complex types make sense here, so worth a look.
> 
> So, why did you think that was worth checking? 

Because experience showed that these are often corner cases in newly added APIs that require special treatments. It'd be nice to take care of them in one shot.

> I checked for `numpy.asarray_chkfinite`. The following were the outputs: (which are as expected)

Thank you, @khushi-411. Looks like the current implementation should work fine. I'll make a few minor, style suggestions shortly.

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,19 @@
 @testing.gpu
 class TestKind(unittest.TestCase):
 
+    @testing.for_all_dtypes()
+    def test_asarray_chkfinite(self, dtype):
```
leofang, line 14: There are testing utilities that can help compare returned NumPy & CuPy arrays, parametrize array orders, etc. It looks like this file was rarely touched so it didn't get updated to use those, but it's never too late to modernize it 🙂
```suggestion
    @testing.for_orders('CFAK')
    @testing.for_all_dtypes()
    @testing.numpy_cupy_array_equal()
    def test_asarray_chkfinite(self, xp, dtype, order):
```

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,19 @@
...
+        a = [0, 4, 0, 5]
+        a_gpu = cupy.asarray_chkfinite(a, dtype=dtype)
+        a_cpu = numpy.asarray_chkfinite(a, dtype=dtype)
+        assert a_gpu.dtype == a_cpu.dtype
```
leofang, line 18: Using the testing helpers (see above) allows you to write a lot less boilerplate codes and increase test coverages (after the return a lot of things are checked).
```suggestion
        return xp.asarray_chkfinite(a, dtype=dtype, order=order)
```

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,19 @@
...
+
+    @unittest.expectedFailure
+    @testing.for_all_dtypes()
+    def test_asarray_chkfinite_non_finite_vals(self, dtype):
```
leofang, line 22: Ditto. Also, we are phasing out the usage of `unittest`, and using pytest is preferred:  
```suggestion
    @testing.for_orders('CFAK')
    @testing.for_all_dtypes()
    def test_asarray_chkfinite_non_finite_vals(self, dtype, order):
```

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,19 @@
...
+        a = [-numpy.inf, 0., numpy.inf]
```
leofang, line 23: Nitpick: Ideally we also want to include NaN, though it's probably shadowed by the first non-finite entry... 
```suggestion
        a = [-numpy.inf, 0., numpy.inf, numpy.nan]
```

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,19 @@
...
+        self.assertRaises(ValueError, cupy.asarray_chkfinite, a, dtype=dtype)
```
leofang, line 24: See https://github.com/cupy/cupy/pull/3256 for how to check exceptions.
```suggestion
        for xp in (numpy, cupy):
            with pytest.raises(ValueError):
                xp.asarray_chkfinite(a, dtype=dtype, order=order)
```

leofang (changes requested):
As mentioned earlier, suggesting a few style changes to be in line with current practice.

tests/cupy_tests/manipulation_tests/test_kind.py
leofang: Please see my review comments below.

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: > actual GPU tests are not yet run

Oo, that's the reason (: 
> It'd be nice to take care of them in one shot.

Ok, I understood. I'll try to figure out those cases the next time.

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: > t looks like this file was rarely touched so it didn't get updated to use those, but it's never too late to modernize it :slightly_smiling_face: 

I totally agree with you and it seems Interesting. Shall I update other parts too? 
In this file, I think the following test cases can be updated:
- [test_asfortranarray](https://github.com/cupy/cupy/blob/0757aa31a10785e6b728ad465724e97a83945b73/tests/cupy_tests/manipulation_tests/test_kind.py#L21): All asfortranarray tests.
- [test_require_C_and_F_flags](https://github.com/cupy/cupy/blob/0757aa31a10785e6b728ad465724e97a83945b73/tests/cupy_tests/manipulation_tests/test_kind.py#L88)

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Got it! It's cooler than the previous approach.

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Sure, I changed it!

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: > See #3256 for how to check exceptions.

Thanks for sharing the informative PR. I read that thoroughly (: 
I was wondering if a bit of enhancement can be made in code. For example, here, arrays are created inside the `for-loops`; that is not needed as it is independent of `xp`.
```
def test_dim_mismatch_ellipsis2(self):
        for xp in (numpy, cupy):
            a = xp.arange(12).reshape(2, 3, 2)
                xp.einsum('i...,...i', a, a)
```
We can modify it to:
```
def test_dim_mismatch_ellipsis2(self):
        a = xp.arange(12).reshape(2, 3, 2)
        for xp in (numpy, cupy):
                xp.einsum('i...,...i', a, a)
```
#### Gains
- Will take lesser computation time.

What do you think? Thanks for reading it (:

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,31 @@
 from cupy import _core
 
 
+def asarray_chkfinite(a, dtype=None, order=None):
+    """
+    Convert the given input to an array,
+    and raises an error if the input contains NaNs or Infs.
```
leofang, line 8: CuPy's convention is to start the short description right after the triple quotes. Also, correct the verb to be consistent:
```suggestion
    """Converts the given input to an array,
    and raises an error if the input contains NaNs or Infs.
```

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,31 @@
...
+
+    Args:
+        a: array like.
+        dtype: data type, optional
+        order: {'C', 'F', 'A', 'K'}, optional
+
+    Returns:
+        cupy.ndarray: Array interpretation of `a`.
```
leofang, line 16:
```suggestion
        cupy.ndarray: An array on the current device.
```

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,21 @@
...
+    @testing.for_orders('CFAK')
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_asarray_chkfinite(self, xp, dtype, order):
+        a = [0, 4, 0, 5]
+        return xp.asarray_chkfinite(a, dtype=dtype, order=order)
+
+    @testing.for_orders('CFAK')
+    @testing.for_all_dtypes()
+    def test_asarray_chkfinite_non_finite_vals(self, xp, dtype, order):
```
leofang, line 22: We don't use the decorator `numpy_cupy_array_equal()` here, so `xp` is not needed:
```suggestion
    def test_asarray_chkfinite_non_finite_vals(self, dtype, order):
```

leofang (changes requested):
Thanks, @khushi-411, just left a few minor suggestions. 

btw in case you don't know already: you can apply the suggested changes _at once_ easily via the GH web UI by going to the "Files changed" tab -> selecting "Add suggestions in batch" -> "Commit changes". Then, to check it locally you can just do `git pull` on the current branch.

tests/cupy_tests/manipulation_tests/test_kind.py
leofang: Feel free to work on it! But I'd suggest to create a separate PR so that each PR's scope is well defined.

tests/cupy_tests/manipulation_tests/test_kind.py
leofang: No, `xp` is the loop variable, so it can't be pulled out of the loop.

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Yeah, sure. Thanks!!

leofang (dismissed):
LGTM! Thanks for contributing, @khushi-411.

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Oops, agreed. I was walking without looking into the other side :(

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,21 @@
...
+    def test_asarray_chkfinite(self, xp, dtype, order):
+        a = [0, 4, 0, 5]
+        return xp.asarray_chkfinite(a, dtype=dtype, order=order)
+
+    @testing.for_orders('CFAK')
+    @testing.for_all_dtypes()
+    def test_asarray_chkfinite_non_finite_vals(self, dtype, order):
+        a = [-numpy.inf, 0., numpy.inf, numpy.nan]
+        for xp in (numpy, cupy):
+            with pytest.raises(ValueError):
```
oke-aditya, line 25: Hey Khushi @khushi-411 
I believe you can check the error message as well. 


```
with pytest.raises(ValueError, match='Expected error message'):
```

This is a good practice and ensures exact error is raised. You can use Regex etc but passing a string works well too.

khushi-411: Hi, @oke-aditya!
Oo yes! Thanks for pointing out (: I'll keep a note!

leofang: Hmmm I missed your [previous observation](https://github.com/cupy/cupy/pull/6275#discussion_r775804617) and so the CI failed, let's fix it...
```suggestion
            if xp.issubdtype(dtype, xp.integer):
                error = OverflowError
            else:
                error = ValueError
            with pytest.raises(error):
```

khushi-411: Here, we set `OverflowError` for all the dtypes lower/equal than integer type and the rest raises the `ValueError`.
I want to know more about the fixes you made and the error raised. The error message in the CI is:
```
=================================== FAILURES =================================== |  
-- | --
00:48:25.963796 STDOUT 2232] | _______________ TestKind.test_asarray_chkfinite_non_finite_vals ________________ |  
00:48:25.963800 STDOUT 2232] |   |  
00:48:25.963892 STDOUT 2232] | self = <cupy_tests.manipulation_tests.test_kind.TestKind testMethod=test_asarray_chkfinite_non_finite_vals> |  
00:48:25.963896 STDOUT 2232] | dtype = <class 'numpy.int8'>, order = 'C' |  
00:48:25.963902 STDOUT 2232] |   |  
00:48:25.963940 STDOUT 2232] | @testing.for_orders('CFAK') |  
00:48:25.963943 STDOUT 2232] | @testing.for_all_dtypes() |  
00:48:25.963950 STDOUT 2232] | def test_asarray_chkfinite_non_finite_vals(self, dtype, order): |  
00:48:25.964010 STDOUT 2232] | a = [-numpy.inf, 0., numpy.inf, numpy.nan] |  
00:48:25.964015 STDOUT 2232] | for xp in (numpy, cupy): |  
00:48:25.964020 STDOUT 2232] | with pytest.raises(ValueError): |  
00:48:25.964056 STDOUT 2232] | >               xp.asarray_chkfinite(a, dtype=dtype, order=order) |  
00:48:25.964058 STDOUT 2232] |   |  
00:48:25.964065 STDOUT 2232] | a          = [-inf, 0.0, inf, nan] |  
00:48:25.964113 STDOUT 2232] | dtype      = <class 'numpy.int8'> |  
00:48:25.964118 STDOUT 2232] | order      = 'C' |  
00:48:25.964126 STDOUT 2232] | self       = <cupy_tests.manipulation_tests.test_kind.TestKind testMethod=test_asarray_chkfinite_non_finite_vals> |  
00:48:25.964190 STDOUT 2232] | xp         = <module 'numpy' from '/opt/pyenv/versions/3.9.6/lib/python3.9/site-packages/numpy/__init__.py'>
```
Thank you!!

cc: @leofang

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,30 @@
...
+    and raises an error if the input contains NaNs or Infs.
+
+    Args:
+        a: array like.
+        dtype: data type, optional
+        order: {'C', 'F', 'A', 'K'}, optional
+
+    Returns:
+        cupy.ndarray: An array on the current device.
+
```
kmaehashi, line 16:
```suggestion
   .. note::
      This function performs device synchronization when ``a`` is a
      ``cupy.ndarray``.
```

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,30 @@
...
+    .. seealso:: :func:`numpy.asarray_chkfinite`
+
+    """
+
+    a = cupy.asarray(a, dtype=dtype, order=order)
```
kmaehashi, line 21: To avoid synchronization as much as possible, I think it's better to do inf check on CPU.

```py
if isinstance(a, cupy.ndarray):
    if not cupy.isfinite(a).all():
        raise ValueError(...)
else:
    a = cupy.asarray(np.asarray_chkfinite(a, dtype=dtype, order=order), dtype=dtype, order=order))
return a
```

kmaehashi (changes requested):

tests/cupy_tests/manipulation_tests/test_kind.py
leofang: In part of this test we convert infs/nans to ints. But they are not compatible and an `OverflowError` would be raised if `dtype` is of any integer type, so we need to branch out to let pytest catch different errors.

Looks like when `dtype` is bool nothing would be raised, so this test still failed. Let me suggest another approach shortly. (You can see why I said we need to check those corner cases 🙂)

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -10,6 +10,25 @@
...
 
+    @testing.for_orders('CFAK')
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_asarray_chkfinite(self, xp, dtype, order):
+        a = [0, 4, 0, 5]
+        return xp.asarray_chkfinite(a, dtype=dtype, order=order)
+
+    @testing.for_orders('CFAK')
+    @testing.for_all_dtypes()
```
leofang, line 21: It does not make sense to test bool with nan/inf, so we simply disable it. (Alternatively, we can use `contextlib.nullcontext` for bool, which is available since Python 3.7, and a `pytest.raise` context manager for other dtypes. I don't have strong opinions here.) 
```suggestion
    @testing.for_all_dtypes(no_bool=True)
```

leofang (changes requested):

tests/cupy_tests/manipulation_tests/test_kind.py
khushi-411: Oo, that's the reason, Got It! Thanks!
> You can see why I said we need to check those corner cases 🙂

Yes, it's very much interesting 🙂

cupy/_manipulation/kind.py
leofang: @kmaehashi Does it matter to avoid synchronization here if an exception is going to be raised anyway? Wouldn't it be better to raise sooner (by doing the check on GPU)? Put it differently, if the check passes (no exception) this takes longer than the original implementation.

kmaehashi: Sorry my proposal was inaccurate, this is what I intended:
```py
if isinstance(a, cupy.ndarray):
    if not cupy.isfinite(a).all():
        raise ValueError(...)
else:
    a = np.asarray_chkfinite(a, dtype=dtype, order=order)
return cupy.asarray(a, dtype=dtype, order=order)
```

When an array is `cupy.ndarray`, there's no difference between the original implementation and my proposal. When an array is CPU array-like, the original implementation copies the array to GPU, checks for inf on GPU, and copies the result back to CPU. My proposal changes this to check for inf on CPU and (if no exceptions are raised) copy the array it to GPU.

khushi-411: From the above discussion:
#### Original Implementation for CPU array
- Copies the arrays from CPU to GPU
- Perform Operations on GPU
- Copies back to CPU
#### Your Implementation for CPU array
- Performs operations on CPU
- Copies to GPU (if there are no infs and nans)
#### As per @leofang's comment
- If checks are passed, the implementation takes a longer time.
- GPU error's will be faster

@leofang, @kmaehashi, I have a small doubt. Like in the original implementation, we are making two implicit switches (CPU->GPU->CPU), while by your proposal, we are making a single switch (CPU->GPU). I am not too sure, but don't you think these *implicit* switches will affect the model's performance? Which factor to consider prior?

Thanks for the discussions! (:

leofang: > When an array is `cupy.ndarray`, there's no difference between the original implementation and my proposal.

Yes.

> When an array is CPU array-like, the original implementation copies the array to GPU, checks for inf on GPU, and copies the result back to CPU. My proposal changes this to check for inf on CPU and (if no exceptions are raised) copy the array it to GPU.

Depends on our expectation for the likelihood of raising an error (due to either inf or nan) for this routine, the decision could differ. @kmaehashi has implicitly assumed that it's more likely to raise, whereas I expected in most cases the check will pass. In the former case, it's OK to do the check on host (which is slower) since the subsequent control flow will be returned to host, while in the latter case, it's cheaper to run the check on device (faster) and proceed to the next GPU calculation right away if there's no error.

kmaehashi: Hmm I can be missing something.

> it's cheaper to run the check on device (faster) and proceed to the next GPU calculation right away

The evaluation of `if not cupy.isfinite(a).all()` always requires a copy of 0-dim bool array from device to host, so synchronization happens in any case (whether the inf check pass or not).

> I expected in most cases the check will pass

I also expect this :)

leofang: Yes, but If the check runs much faster on GPU than on CPU, then it's probably ok to pay the sync cost 🙂

khushi-411: cc: @kmaehashi @leofang

As per the above discussion, I updated the function considering speed as a priority.
Please let me know if anything else is required. I'll be happy to work on it.

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,32 @@
...
+    Args:
+        a: array like.
+        dtype: data type, optional
+        order: {'C', 'F', 'A', 'K'}, optional
+
+    Returns:
+        cupy.ndarray: An array on the current device.
+    .. note::
+        This function performs device synchronization when ``a`` is a
+        ``cupy.ndarray``.
```
kmaehashi, line 18:
```suggestion

    .. note::
        This function performs device synchronization.
```

kmaehashi (changes requested):

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,31 @@
...
+    and raises an error if the input contains NaNs or Infs.
+
+    Args:
+        a: array like.
+        dtype: data type, optional
+        order: {'C', 'F', 'A', 'K'}, optional
+
+    Returns:
+        cupy.ndarray: An array on the current device.
+    .. note::
```
kmaehashi, line 16:
```suggestion

    .. note::
```

cupy/_manipulation/kind.py
```diff
@@ -2,6 +2,31 @@
...
+        This function performs device synchronization.
+
+    .. seealso:: :func:`numpy.asarray_chkfinite`
+
+    """
+
+    a = cupy.asarray(a, dtype=dtype, order=order)
+    if not cupy.isfinite(a).all():
+        raise ValueError(
+            "array must not contain Infs or NaNs.")
```
kmaehashi, line 26:
```suggestion
            "array must not contain Infs or NaNs")
```

kmaehashi (approved):

khushi-411:
Hi, @kmaehashi, @leofang, @toslunar, @oke-aditya!
Thank you so much for reviewing! :)

kmaehashi:
@toslunar Could you approve if it's ok to you? 😄

leofang (approved):

khushi-411:
Hi, @toslunar! A gentle ping for your approval :)

toslunar (approved):
LGTM

I'm sorry for keeping you waiting.

leofang:
Thanks @khushi-411 and all!

PR #6279, opened 2021-12-29
Add `cupy.vander` API.

tom24d:
Ref: #6078 

Hi team,
I have implemented `vander` API.

cc @kmaehashi 

Thanks!

### DIFF
--- a/cupy/_creation/matrix.py
+++ b/cupy/_creation/matrix.py
@@ -141,7 +141,39 @@ def triu(m, k=0):
     return cupy.where(mask, m.dtype.type(0), m)
 
 
-# TODO(okuta): Implement vander
+def vander(x, N=None, increasing=False):
+    """Returns a Vandermonde matrix.
+
+    Args:
+        x (array-like): 1-D array or array-like object.
+        N (int, optional): Number of columns in the output.
+            ``N = len(x)`` by default.
+        increasing (bool, optional): Order of the powers of the columns.
+            If True, the powers increase from right to left,
+            if False (the default) they are reversed.
+
+    Returns:
+        cupy.ndarray: A Vandermonde matrix.
+
+    .. seealso:: :func:`numpy.vander`
+
+    """
+    x = cupy.asarray(x)
+    if x.ndim != 1:
+        raise ValueError("x must be a one-dimensional array or sequence.")
+    if N is None:
+        N = len(x)
+
+    v = cupy.empty((len(x), N), dtype=numpy.promote_types(x.dtype, int))
+    tmp = v[:, ::-1] if not increasing else v
+
+    if N > 0:
+        tmp[:, 0] = 1
+    if N > 1:
+        tmp[:, 1:] = x[:, None]
+        cupy.cumprod(tmp[:, 1:], out=tmp[:, 1:], axis=1)
+
+    return v
 
 
 # TODO(okuta): Implement mat
--- a/cupy/_creation/matrix.py
+++ b/cupy/_creation/matrix.py
@@ -167,11 +167,7 @@ def vander(x, N=None, increasing=False):
     v = cupy.empty((len(x), N), dtype=numpy.promote_types(x.dtype, int))
     tmp = v[:, ::-1] if not increasing else v
 
-    if N > 0:
-        tmp[:, 0] = 1
-    if N > 1:
-        tmp[:, 1:] = x[:, None]
-        cupy.cumprod(tmp[:, 1:], out=tmp[:, 1:], axis=1)
+    cupy.power(x.reshape(-1, 1), cupy.arange(N), out=tmp)
 
     return v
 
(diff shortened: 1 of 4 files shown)


tests/cupy_tests/creation_tests/test_matrix.py
```diff
@@ -179,3 +179,46 @@ def test_triu_nega(self, xp, dtype):
...
+
+    def test_vander_scalar(self):
+        for xp in (numpy, cupy):
+            with pytest.raises(ValueError):
+                xp.vander(1, N=self.N, increasing=self.increasing)
+
+    def test_vander_0dim(self):
+        for xp in (numpy, cupy):
+            with pytest.raises(ValueError):
+                a = xp.zeros(())
```
toslunar, line 217: nitpick: To clarify `vander` is tested,
```suggestion
            a = xp.zeros(())
            with pytest.raises(ValueError):
```

tests/cupy_tests/creation_tests/test_matrix.py
```diff
@@ -179,3 +179,46 @@ def test_triu_nega(self, xp, dtype):
...
+                xp.vander(a, N=self.N, increasing=self.increasing)
+
+    def test_vander_2dim(self):
+        for xp in (numpy, cupy):
+            with pytest.raises(ValueError):
+                m = xp.zeros((2, 2))
```
toslunar, line 223: ditto
```suggestion
            m = xp.zeros((2, 2))
            with pytest.raises(ValueError):
```

toslunar (changes requested):

tom24d:
@toslunar Thanks for the review! PTAL.

cupy/_creation/matrix.py
```diff
@@ -141,7 +141,39 @@ def triu(m, k=0):
...
+        N = len(x)
+
+    v = cupy.empty((len(x), N), dtype=numpy.promote_types(x.dtype, int))
+    tmp = v[:, ::-1] if not increasing else v
+
+    if N > 0:
+        tmp[:, 0] = 1
+    if N > 1:
+        tmp[:, 1:] = x[:, None]
+        cupy.cumprod(tmp[:, 1:], out=tmp[:, 1:], axis=1)
```
asi1024, line 174:
```suggestion
    cupy.power(x.reshape(-1, 1), cupy.arange(N), out=tmp)
```
How about using `cupy.power` instead for performance?

<details>

Benchmark script:
```py
import cupy
import cupyx.profiler

x = cupy.arange(1000, dtype='float32')

perf = cupyx.profiler.benchmark(cupy.vander, (x, 1000))
print(perf)
```

Result (original)
```
vander              :    CPU:  118.208 us   +/- 2.582 (min:  115.581 / max:  214.753) us     GPU-0:  208.787 us   +/-21.077 (min:  199.680 / max:  329.728) us
```

Result (suggestion):
```
vander              :    CPU:   49.705 us   +/-19.058 (min:   46.526 / max: 1923.795) us     GPU-0:   76.910 us   +/-20.474 (min:   64.512 / max: 1950.720) us
```

tom24d: That's cool! I'm surprised that the proposed impl is faster in cupy while cumulative prod is preferred in numpy( https://github.com/numpy/numpy/commit/30aeb2ee59d8b52bdd1ace33ac3d9b4e45c17b32) ...😯

toslunar (dismissed):
LGTM!

asi1024:
Could you use `testing.numpy_cupy_allclose`?

asi1024:
@tom24d LGTM! Thanks!

asi1024 (approved):

cupy/_creation/matrix.py
```diff
@@ -141,7 +141,35 @@ def triu(m, k=0):
...
+            If True, the powers increase from right to left,
+            if False (the default) they are reversed.
+
+    Returns:
+        cupy.ndarray: A Vandermonde matrix.
+
+    .. seealso:: :func:`numpy.vander`
+
+    """
+    x = cupy.asarray(x)
```
khushi-411, line 161: Hi, @toslunar, @asi1024, @tom24d,
I was very interested in looking into the PR and took a glance at it. I wonder if we can modify this statement to
```
if not isinstance(x, cupy.ndarray):
    raise TypeError('`x` should be of type cupy.ndarray')
```
To avoid host-to-device sync as mentioned in https://github.com/cupy/cupy/pull/6280#discussion_r779972513. Benefits of it:
- Performance of the model will be affected if we apply internal synchronization without the user's knowledge.

Credits: @takagi! Thanks to you (:

asi1024: @khushi-411 In our policy, the features classified as [Array creation routines](https://numpy.org/doc/stable/reference/routines.array-creation.html) (e.g., `vander`, `asarray`, `diag`, ...) should allow all ArrayLike inputs.  The details of the policy are still under discussion, but we are planning to update the [Development policies](https://github.com/cupy/cupy/wiki/Development-policies) after they are finalized!

khushi-411: Oo, got it! Thanks for it (: @asi1024!
(I would love to read the updated policies :))

Issue #3972, opened 2020-09-08
Add support for subclassing cupy.ndarray


mrkwjc:
Hi!

Many projects (including mine) use subclasses of numpy.ndarray, like this:
```python
class Field(np.ndarray):
    def __new__(cls, data, prefix=1):
        data = np.asarray(data)
        field = data.view(cls)
        field.prefix = prefix
        return field

    def __array_finalize__(self, field):
        if field is None:
            return
        self.prefix = getattr(field, 'prefix', 1)
```
If one tries to use cupy.ndarray as a numpy replacement, then at creation of the `Field` instance an error is raised:
```python
AttributeError: 'cupy.core.core.ndarray' object has no attribute 'prefix'
```
Are there plans to support in cupy subclassing similar to [numpy approach](https://numpy.org/doc/stable/user/basics.subclassing.html)? Maybe some workarounds are possible currently?

Regards,
Marek

leofang:
What if you follow https://numpy.org/devdocs/user/basics.subclassing.html#simple-example-adding-an-extra-attribute-to-ndarray and do `field = super(Field, cls).__new__(...)` in `__new__`?

In general, though, my impression is even in NumPy subclassing `ndarray` is considered a poor practice despite they provide the instruction. In your case it might also prevent you from using CuPy as a drop-in replacement (though not impossible).

mrkwjc:
Thanks for the answer. I redefined Field this way:
```python
class Field(np.ndarray):
    def __new__(cls, shape, dtype=float, buffer=None, offset=0,
                strides=None, order=None, prefix=1):
        field = super().__new__(cls, data.shape, dtype, buffer, offset,
                                strides, order)
        field.prefix = prefix
        return field

    def __array_finalize__(self, field):
        if field is None:
            return
        self.prefix = getattr(field, 'prefix', 1)
```
and then i succesfully created an instance with:
```python
f = Field(shape=(2, 2))
```
But the initialization with prefix gives an error:
```python
f = Field(shape=(2, 2), prefix=2)
TypeError: __init__() got an unexpected keyword argument 'prefix'
```
It seems this method is not a real option because i cannot initialize prefix at instance creation. I don't see the way to initialize also the array entries...

leofang:
[...]

PR #6280, opened 2021-12-29
Add `cupy.ediff1d` API.

khushi-411:
Hi, Team!
The PR **aim** to add `cupy.ediff1d` API. It aims to resolve one part to https://github.com/cupy/cupy/issues/6078#issue-1051191869.

Looking forward to your viewpoints. Thanks!

cc: @kmaehashi

### DIFF
--- a/cupy/_math/sumprod.py
+++ b/cupy/_math/sumprod.py
@@ -540,7 +540,8 @@ def ediff1d(arr, to_end=None, to_begin=None):
     if to_begin is None:
         l_begin = 0
     else:
-        to_begin = cupy.array(to_begin)
+        if not isinstance(to_begin, cupy.ndarray):
+            raise TypeError('`to_begin` should be of type cupy.ndarray')
         if not cupy.can_cast(to_begin, dtype_req, casting="same_kind"):
             raise TypeError("dtype of `to_begin` must be compatible "
                             "with input `arr` under the `same_kind` rule.")
@@ -551,7 +552,8 @@ def ediff1d(arr, to_end=None, to_begin=None):
     if to_end is None:
         l_end = 0
     else:
-        to_end = cupy.array(to_end)
+        if not isinstance(to_end, cupy.ndarray):
+            raise TypeError('`to_end` should be of type cupy.ndarray')
         if not cupy.can_cast(to_end, dtype_req, casting="same_kind"):
             raise TypeError("dtype of `to_end` must be compatible "
                             "with input `arr` under the `same_kind` rule.")
--- a/cupy/_math/sumprod.py
+++ b/cupy/_math/sumprod.py
@@ -507,7 +507,65 @@ def gradient(f, *varargs, axis=None, edge_order=1):
         return outvals
 
 
-# TODO(okuta): Implement ediff1d
+def ediff1d(arr, to_end=None, to_begin=None):
+    """
+    Calculates the difference between consecutive elements of an array.
+
+    Args:
+        arr (cupy.ndarray): Input array.
+        to_end (cupy.ndarray, optional): Numbers to append at the end
+            of the returend differences.
+        to_begin (cupy.ndarray, optional): Numbers to prepend at the
+            beginning of the returned differences.
+
+    Returns:
+        cupy.ndarray: New array consisting differences among succeeding
+        elements.
+
+    .. seealso:: :func:`numpy.ediff1d`
+    """
+    # to flattened array.
+    arr = cupy.asanyarray(arr).ravel()
+
+    # to ensure the dtype of the output array is same as that of input.
+    dtype_req = arr.dtype
+
+    # if none optional cases are given
+    if to_begin is None and to_end is None:
+        return arr[1:] - arr[:-1]
+
+    if to_begin is None:
+        l_begin = 0
+    else:
+        to_begin = cupy.asanyarray(to_begin)
+        if not cupy.can_cast(to_begin, dtype_req, casting="same_kind"):
+            raise TypeError("dtype of `to_begin` must be compatible "
+                            "with input `arr` under the `same_kind` rule.")
+
+        to_begin = to_begin.ravel()
+        l_begin = len(to_begin)
+
+    if to_end is None:
+        l_end = 0
+    else:
+        to_end = cupy.asanyarray(to_end)
+        if not cupy.can_cast(to_end, dtype_req, castinf="same_kind"):
+            raise TypeError("dtype of `to_end` must be compatible "
+                            "with input `arr` under the `same_kind` rule.")
+
+        to_end = to_end.ravel()
+        l_end = len(to_end)
+
+    # calulating using in place operation
+    l_diff = max(len(arr) - 1, 0)
+    result = cupy.empty(l_diff + l_begin + l_end, dtype=arr.dtype)
+    result = arr.__array_wrap__(result)
+    if l_begin > 0:
+        result[:l_begin] = to_begin
+    if l_end > 0:
+        result[l_begin + l_diff:] = to_end
+    cupy.subtract(arr[1:], arr[:-1], result[l_begin:l_begin + l_diff])
+    return result
 
 
 # TODO(okuta): Implement cross
--- a/cupy/_math/sumprod.py
+++ b/cupy/_math/sumprod.py
@@ -559,7 +559,7 @@ def ediff1d(arr, to_end=None, to_begin=None):
     # calulating using in place operation
     l_diff = max(len(arr) - 1, 0)
     result = cupy.empty(l_diff + l_begin + l_end, dtype=arr.dtype)
-    #result = arr.__array_wrap__(result)
+    # result = arr.__array_wrap__(result)
     if l_begin > 0:
         result[:l_begin] = to_begin
     if l_end > 0:
--- a/cupy/_math/sumprod.py
+++ b/cupy/_math/sumprod.py
@@ -549,7 +549,7 @@ def ediff1d(arr, to_end=None, to_begin=None):
         l_end = 0
     else:
         to_end = cupy.asanyarray(to_end)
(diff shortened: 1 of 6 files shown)


cupy/_math/sumprod.py
```diff
@@ -507,7 +507,65 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+            raise TypeError("dtype of `to_end` must be compatible "
+                            "with input `arr` under the `same_kind` rule.")
+
+        to_end = to_end.ravel()
+        l_end = len(to_end)
+
+    # calulating using in place operation
+    l_diff = max(len(arr) - 1, 0)
+    result = cupy.empty(l_diff + l_begin + l_end, dtype=arr.dtype)
+    # result = arr.__array_wrap__(result)
```
khushi-411, line 562: As CuPy does not support `__array_wrap__` protocol. Other way can be using `__array_ufunc__` protocol. Will it be wiser to add that here?

cupy/_math/sumprod.py
```diff
@@ -507,7 +507,65 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+            beginning of the returned differences.
+
+    Returns:
+        cupy.ndarray: New array consisting differences among succeeding
+        elements.
+
+    .. seealso:: :func:`numpy.ediff1d`
+    """
+    # to flattened array.
+    arr = cupy.asanyarray(arr).ravel()
```
takagi, line 528: To avoid implicit host-to-device sync, CuPy has a design policy that it accepts only `cupy.ndarray` where NumPy takes an arraylike. Would you remove `cupy.asanyarray`, then add a type check for `arr` instead?

cupy/_math/sumprod.py
```diff
@@ -507,7 +507,65 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+    dtype_req = arr.dtype
+
+    # if none optional cases are given
+    if to_begin is None and to_end is None:
+        return arr[1:] - arr[:-1]
+
+    if to_begin is None:
+        l_begin = 0
+    else:
+        to_begin = cupy.asanyarray(to_begin)
```
takagi, line 540: ditto (a type check as `arr`)

cupy/_math/sumprod.py
```diff
@@ -507,7 +507,65 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+            raise TypeError("dtype of `to_begin` must be compatible "
+                            "with input `arr` under the `same_kind` rule.")
+
+        to_begin = to_begin.ravel()
+        l_begin = len(to_begin)
+
+    if to_end is None:
+        l_end = 0
+    else:
+        to_end = cupy.asanyarray(to_end)
```
takagi, line 551: ditto (a type check as `arr`)

cupy/_math/sumprod.py
takagi: We can skip calling `__array_wrap__` here as CuPy does not support subclassing `cupy.ndarray` yet, which may be considered in #3972. Would you leave a comment for now?

```suggestion
    # result = arr.__array_wrap__(result)  # CuPy does not support subclassing a ndarray
```

takagi (changes requested):
Almost fine, and please see the comments?

cupy/_math/sumprod.py
khushi-411: Added!
Thanks for sharing the ongoing issue.

cupy/_math/sumprod.py
khushi-411: Yes, sure! Added. Thanks for the concept.

cupy/_math/sumprod.py
```diff
@@ -507,7 +507,66 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+            beginning of the returned differences.
+
+    Returns:
+        cupy.ndarray: New array consisting differences among succeeding
+        elements.
+
+    .. seealso:: :func:`numpy.ediff1d`
+    """
+    # to flattened array.
+    arr = cupy.array(arr).ravel()
```
takagi, line 528: `cupy.array()` works as same as `cupy.asanyarray()`, and it can also bring host-to-device sync. How about the following (also `to_begin` and `to_end`)?
```suggestion
    if not isinstance(arr, cupy.ndarray):
        raise TypeError('`arr` should be of type cupy.ndarray')
    arr = arr.ravel()
```

khushi-411: I think it's a more thoughtful approach to do so. Thanks for it! Just curious to know why do we need to avoid host-to-device sync here?

Also, I tried for `to_begin` and `to_end` too. I got the following results.
#### Code
```
if to_begin is None:
        l_begin = 0
    else:
        if not isinstance(to_begin, numpy.ndarray):
            raise TypeError('`to_begin` should be of type numpy.ndarray')
        if not cupy.can_cast(to_begin, dtype_req, casting="same_kind"):
            raise TypeError("dtype of `to_begin` must be compatible "
                            "with input `arr` under the `same_kind` rule.")

        to_begin = to_begin.ravel()
        l_begin = len(to_begin)
```
#### Result
```
tests/cupy_tests/math_tests/test_sumprod.py:1066: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cupy/_math/sumprod.py:578: in ediff1d
    result[:l_begin] = to_begin
cupy/_core/core.pyx:1498: in cupy._core.core.ndarray.__setitem__
    _indexing._ndarray_setitem(self, slices, value)
cupy/_core/_routines_indexing.pyx:51: in cupy._core._routines_indexing._ndarray_setitem
    _scatter_op(self, slices, value, 'update')
cupy/_core/_routines_indexing.pyx:987: in cupy._core._routines_indexing._scatter_op
    y.fill(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   raise ValueError(
E   AssertionError: Only cupy raises error
E   
E     File "/home/khushi/Documents/cupy/cupy/testing/_loops.py", line 43, in _call_func
E       result = impl(*args, **kw)
E     File "/home/khushi/Documents/cupy/tests/cupy_tests/math_tests/test_sumprod.py", line 1066, in test_ediff1d_to_begin1
E       return xp.ediff1d(a, to_begin=numpy.array([0], dtype=dtype))
E     File "/home/khushi/Documents/cupy/cupy/_math/sumprod.py", line 578, in ediff1d
E       result[:l_begin] = to_begin
E     File "cupy/_core/core.pyx", line 1498, in cupy._core.core.ndarray.__setitem__
E       _indexing._ndarray_setitem(self, slices, value)
E     File "cupy/_core/_routines_indexing.pyx", line 51, in cupy._core._routines_indexing._ndarray_setitem
E       _scatter_op(self, slices, value, 'update')
E     File "cupy/_core/_routines_indexing.pyx", line 987, in cupy._core._routines_indexing._scatter_op
E       y.fill(value)
E     File "cupy/_core/core.pyx", line 676, in cupy._core.core.ndarray.fill
E       raise ValueError(
E   ValueError: non-scalar numpy.ndarray cannot be used to fill

cupy/_core/core.pyx:676: AssertionError
----------------------------------------- Captured stdout call -----------------------------------------
dtype is <class 'numpy.float64'>
```
#### My viewpoints
- We cannot use `cupy.ndarray` type in `to_begin` and `to_end` to `isinstance` for the reason that CuPy's `ndarray` does not support `subclassing` as of now.
- For `numpy.ndarray`, it throws the mentioned error as we cannot access the CPU memory. Credits: [StackOverflow](https://stackoverflow.com/questions/68634058)
- Since, we are passing `numpy.ndarray` as input for `to_begin` and `to_end`, which runs on CPU. We need to transfer the data to GPU. Henceforth, we need host-to-device sync.

Do these sound good to you? Will you please let me know if I am getting something wrong?

Thanks a lot for reviewing the PR and for all valuable suggestions (:

cc: @takagi

takagi: Host-to-device synchronization blocks the execution until CUDA kernels that have been launched before calling the sync complete their work. If `cupy.ediff1d` implicitly synchronized and the user was not aware of it, the performance of the system could be unintentionally incurred. So we have a design policy that we require users to explicitly manage host-to-device synchronization.

For checking `to_begin` and `to_end`, I don't think we need CuPy to support subclassing and the following would be fine. Can I have why you think so?

```Python
if not isinstance(to_begin, cupy.ndarray):
    raise TypeError('`to_begin` should be of type cupy.ndarray')
```

khushi-411: >  If cupy.ediff1d implicitly synchronized and the user was not aware of it, the performance of the system could be unintentionally incurred. So we have a design policy that we require users to explicitly manage host-to-device synchronization.

This is fascinating!

khushi-411: > For checking to_begin and to_end, I don't think we need CuPy to support subclassing and the following would be fine. Can I have why you think so?

Yes, you are absolutely correct. They ain't related to subclassing. I was wrong. I revisited the concept and felt so. Thanks for it!

cc: @takagi

takagi (approved):

takagi:
LGTM!

takagi:
Let me kick the CI again as #6303 is now merged, which fixes cuda115's CI failure.

khushi-411:
Thank you very much for reviewing, @takagi!

> Let me kick the CI again as #6303 is now merged, which fixes cuda115's CI failure.

Oo, that's the reason. Interesting! Thanks for sharing the PR.

PR #6281, opened 2021-12-30
Fix cache key for github actions

kmaehashi:
I noticed that we should update `actions/cache` key for every commit, otherwise cahce will never be updated...

https://github.com/cupy/cupy/runs/4648731921?check_suite_focus=true
> Post Load ccache
> Post job cleanup.
> Cache hit occurred on the primary key ccache, not saving cache.

This PR changes to use the commit hash as a primary key, and falls back to the base branch's commit hash for pull-requests.


toslunar (approved):
LGTM

PR #6285, opened 2022-01-03
Make streams hashable

### LINKED ISSUE #6243: Make CuPy stream hashable?
### Description

This observation was made internally by @mtjrider:
```python
>>> import torch
>>> s=torch.cuda.Stream()
>>> hash(s)
-181027509999151647
>>>
>>> import cupy as cp
>>> s = cp.cuda.Stream()
>>> hash(s)
TypeError: unhashable type: 'Stream'
```
It would be nice to make it hashable because, for example, then streams can be used as keys for a mapping to parallel workloads, which is currently not possible with CuPy streams:
```python
>>> s1 = cp.cuda.Stream()
>>> s2 = cp.cuda.Stream()
>>> map = {s1: 1, s2: 2}
TypeError: unhashable type: 'Stream'
```

### Additional Information

_No response_

leofang:
Close #6243. Because `stream_ptr` is an int, `hash(stream_ptr)` is equal to `stream_ptr` and there is no point to do the former. This actually allows a very simple implementation as `__eq__` does not need to be modified at all.

### DIFF
--- a/cupy/cuda/stream.pyx
+++ b/cupy/cuda/stream.pyx
@@ -225,6 +225,9 @@ class _BaseStream:
         return '<{} {} (device {})>'.format(
             type(self).__name__, self.ptr, self.device_id)
 
+    def __hash__(self):
+        return self.ptr
+
     def use(self):
         """Makes this stream current.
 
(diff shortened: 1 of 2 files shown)


emcastillo (approved):
LGTM!

PR #6287, opened 2022-01-04
Check isinstance before comparison in ``__eq__``

kmaehashi:
Follows-up: #6285

This fixes the following failure (https://github.com/cupy/cupy/issues/6243#issuecomment-994672793):

```py
>>> x = {cupy.cuda.Stream.null: 100, 0: 200}
  File "cupy/cuda/stream.pyx", line 209, in cupy.cuda.stream._BaseStream.__eq__
    return self.ptr == other.ptr
AttributeError: 'int' object has no attribute 'ptr'
```


leofang:
Ah, good catch, thanks!

emcastillo (approved):

PR #5687, opened 2021-08-27
add additional special functions (spherical harmonics, Legendre, Gamma functions)

grlee77:
closes #5684

Marked as draft because:

~- a couple of test cases related to `gammaincinv` and `gammainccinv` are failing (unexpected NaN values)~ Fixed by 
17c75f4  
~- we may need to expand the tests a bit more.~ -> new `sph_harm` and `log1p` test cases have been added
- I wanted to think about how to potentially refactor. For example, notice that the ufunc code template for `sph_harm` includes the template strings for `lpmv` which in turn includes other functions, etc.. Thus, there is duplicate C code across the CUDA kernels, but I'm not sure of the easiest way to avoid that. 

I reorganized the `__init__.py` a bit to group related functions together (as in the SciPy docs). The new functions added in this PR are:
`sph_harm`, `log1p`, `gamminc`, `gammincinv`, `gammincc`, `gammainccinv`, `gammasgn`, `lpmv` and `poch`.

The additional functions are here because they were needed within the implementation of `sph_harm`.

PR #6300, opened 2022-01-11
add logit, expit and log_expit to cupyx.scipy.special

grlee77:
This PR adds three very simple statistical functions to `cupyx.scipy.special`. This PR is built on top of #5687 to avoid conflicts with the changes to `__init__.py` made there. Only the last commit here is new.

cc @IvanYashchuk

### DIFF
--- a/cupyx/scipy/special/_statistics.py
+++ b/cupyx/scipy/special/_statistics.py
@@ -88,7 +88,7 @@ log_expit = _core.create_ufunc(
     Returns:
         cupy.ndarray: values of log(expit(x))
 
-    .. seealso:: :data:`scipy.special.expit`
+    .. seealso:: :data:`scipy.special.log_expit`
 
     .. note::
         The function is mathematically equivalent to ``log(expit(x))``, but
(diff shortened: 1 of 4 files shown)


grlee77:
SciPy [defines these functions](https://github.com/scipy/scipy/blob/d5497bdc3c1a5ce1432c8ca6f1ab92c87937f74d/scipy/special/functions.json#L567-L572) for single ('f'), double ('d') and long double ('g') dtypes while the implementation here is for single or double only.

grlee77:
rebased after #5687 was merged. should be much easier to review now

cupyx/scipy/special/_statistics.py
```diff
@@ -1,6 +1,102 @@
...
+    preamble=log_expit_definition,
+    doc='''Logarithm of the logistic sigmoid function.
+
+    Args:
+        x (cupy.ndarray): input data (must be real)
+
+    Returns:
+        cupy.ndarray: values of log(expit(x))
+
+    .. seealso:: :data:`scipy.special.expit`
```
kmaehashi, line 91:
```suggestion
    .. seealso:: :data:`scipy.special.log_expit`
```

kmaehashi (approved):

PR #6301, opened 2022-01-11
add xlogy and xlog1py to cupyx.scipy.special

grlee77:
This PR adds a couple of simple convenience functions to `cupyx.scipy.special`. This PR is built on top of #5687 to avoid conflicts with the changes to `__init__.py` made there. Only the last commit here is new.

cc @IvanYashchuk 

There is one deviation from SciPy in that the templated implementation here also has a single precision code path, while SciPy itself always promotes to double (see [here](https://github.com/scipy/scipy/blob/main/scipy/special/_xlogy.pxd) where `number_t` is a fused type made up of `double` and `complex<double>`). When testing locally the float32 code path was 4-5 times faster on large arrays.

### DIFF
--- a/cupyx/scipy/special/_xlogy.py
+++ b/cupyx/scipy/special/_xlogy.py
@@ -1,0 +1,68 @@
+from cupy import _core
+
+
+# Note: complex-valued isnan, log and log1p are all defined in
+#       cupy/_core/include/cupy/complex.cuh
+xlogy_definition = """
+
+template <typename T>
+static __device__ T xlogy(T x, T y) {
+    if ((x == (T)0.0) && ~isnan(y)) {
+        return (T)0.0;
+    } else {
+        return x * log(y);
+    }
+}
+
+"""
+
+
+# Note: SciPy only defines dd->d and DD->D
+xlogy = _core.create_ufunc(
+    'cupy_xlogy',
+    ('ee->f', 'ff->f', 'dd->d', 'FF->F', 'DD->D'),
+    'out0 = out0_type(xlogy(in0, in1));',
+    preamble=xlogy_definition,
+    doc='''Compute ``x*log(y)`` so that the result is 0 if ``x = 0``.
+
+    Args:
+        x (cupy.ndarray): input data
+
+    Returns:
+        cupy.ndarray: values of ``x * log(y)``
+
+    .. seealso:: :data:`scipy.special.xlogy`
+
+    ''')
+
+
+xlog1py_definition = """
+
+template <typename T>
+static __device__ T xlog1py(T x, T y) {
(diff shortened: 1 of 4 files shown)


grlee77:
rebased after #5687 was merged. should be much easier to review now

tests/cupyx_tests/scipy_tests/special_tests/test_basic.py
```diff
@@ -81,3 +82,36 @@ def test_log1p_complex(self):
         # complex-valued log1p not yet implemented
         with pytest.raises(TypeError):
             cupyx.scipy.special.log1p(0 + 0j)
+
+    @pytest.mark.parametrize("function", ["xlogy", "xlog1py"])
+    @testing.for_float_dtypes()
```
asi1024, line 87: How about testing also for complex dtypes?
```suggestion
    @testing.for_dtypes('efdFD')
```

tests/cupyx_tests/scipy_tests/special_tests/test_basic.py
```diff
@@ -81,3 +82,36 @@ def test_log1p_complex(self):
...
+        # only test with values > 0 to avoid NaNs
+        x = xp.linspace(-100, 100, 1000, dtype=dtype)
+        y = xp.linspace(0.001, 100, 1000, dtype=dtype)
+        if x.dtype.kind == 'c':
+            x -= 1j * x
+            y += 1j * y
+        return getattr(scp.special, function)(x, y)
+
+    @pytest.mark.parametrize("function", ["xlogy", "xlog1py"])
+    @testing.for_float_dtypes()
```
asi1024, line 100:
```suggestion
    @testing.for_dtypes('efdFD')
```

cupyx/scipy/special/_xlogy.py
```diff
@@ -0,0 +1,68 @@
+from cupy import _core
+
+
+# Note: complex-valued isnan, log and log1p are all defined in
+#       cupy/_core/include/cupy/complex.cuh
+xlogy_definition = """
+
+template <typename T>
+static __device__ T xlogy(T x, T y) {
+    if ((x == (T)0.0) && ~isnan(y)) {
```
asi1024, line 10:
```suggestion
    if ((x == (T)0.0) && !isnan(y)) {
```

tests/cupyx_tests/scipy_tests/special_tests/test_basic.py
grlee77: yes, this was my intention! (I already had complex values in the data generation below and had tested it locally in a console)

asi1024 (approved):

asi1024:
LGTM! Thanks!

PR #6305, opened 2022-01-11
Add `tril_indices` and `tril_indices_from` API.

khushi-411:
Hi, Team!
The PR **aims** to add `cupy.tril_indices` and `cupy.tril_indices_from` API. It follows #6078.

Looking forward to your viewpoints :)
Thanks!

cc: @kmaehashi @takagi @leofang @asi1024  @toslunar

### DIFF
--- a/cupy/_indexing/generate.py
+++ b/cupy/_indexing/generate.py
@@ -469,13 +469,11 @@ def mask_indices(n, mask_func, k=0):
 # TODO(okuta): Implement diag_indices_from
 
 
-# TODO(okuta): Implement mask_indices
-
-
 def tril_indices(n, k=0, m=None):
     """Returns the indices of the lower triangular matrix.
-    Here, the first row contains row coordinates of all indices and
-    the second row contains column coordinates.
+    Here, the first group of elements contains row coordinates
+    of all indices and the second group of elements
+    contains column coordinates.
 
     Parameters
     ----------
@@ -488,7 +486,7 @@ def tril_indices(n, k=0, m=None):
         below it and `k > 0` is above.
     m : int, optional
         The column dimension of the arrays for which the
-        returned arrays will be valid. By default, `m = 0`.
+        returned arrays will be valid. By default, `m = n`.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


khushi-411:
Hmm, the CI didn't pass since these functions are experimental. Any tips to resolve this for CuPy?
Thanks!

cupy/_indexing/generate.py
```diff
@@ -472,10 +472,63 @@ def mask_indices(n, mask_func, k=0):
...
+    n : int
+        The row dimension of the arrays for which the returned
+        indices will be valid.
+    k : int, optional
+        Diagonal above which to zero elements. `k = 0`
+        (the default) is the main diagonal, `k < 0` is
+        below it and `k > 0` is above.
+    m : int, optional
+        The column dimension of the arrays for which the
+        returned arrays will be valid. By default, `m = 0`.
```
takagi, line 491: Be default, `m` is taken to equal to `n`?

cupy/_indexing/generate.py
```diff
@@ -472,10 +472,63 @@ def mask_indices(n, mask_func, k=0):
 # TODO(okuta): Implement mask_indices
 
 
-# TODO(okuta): Implement tril_indices
+def tril_indices(n, k=0, m=None):
+    """Returns the indices of the lower triangular matrix.
+    Here, the first row contains row coordinates of all indices and
+    the second row contains column coordinates.
```
takagi, line 478: I read "row" is used for another meaning in this sentence. How about "element" to indicate members of a tuple?

```suggestion
    Here, the first element contains row coordinates of all indices and
    the second element contains column coordinates.
```

takagi (changes requested):
Thanks! Almost fine, would you see some minor comments?

cupy/_indexing/generate.py
khushi-411: Sure, if you say. But I think it may create confusion, as `element` refers to a single term. What about `group of elements` instead of that?

cupy/_indexing/generate.py
khushi-411: Oops, typo. Thanks!

cupy/_indexing/generate.py
takagi: Perfect.

takagi (approved):

khushi-411:
Thanks for reviewing, @takagi! :smile:

takagi:
Thanks!

PR #6315, opened 2022-01-14
Use CUDA's log1p for cupyx.scipy.special.log1p

grlee77:
This PR addresses the following post-merge review comment: https://github.com/cupy/cupy/pull/5687#pullrequestreview-852497622

Thanks, @asi1024!

I benchmarked calling log1p on a vector with 10,000,000 elements both before and after this PR and performance was very similar either way. 

One question: 
It is a little suprising to me that float32 ends up running 6-7 times faster than float64 even though I there is no float32-specific code path.  Do any of you have insight into why this is? Does CuPy somehow use `log1pf` instead of `log1p` when the input is single precision? (I think I did find that some functions like `log` have complex support via Thrust, but wasn't sure if this also extends to single vs. double precision)

### DIFF
--- a/cupyx/scipy/special/_basic.py
+++ b/cupyx/scipy/special/_basic.py
@@ -1,0 +1,19 @@
+from cupy import _core
+
+
+log1p = _core.create_ufunc(
+    "cupyx_scipy_log1p",
+    ("f->f", "d->d"),
+    "out0 = out0_type(log1p(in0));",
+    doc="""Elementwise function for scipy.special.log1p
+
+    Calculates log(1 + x) for use when `x` is near zero.
+
+    Notes
+    -----
+    This implementation currently does not support complex-valued `x`.
+
+    .. seealso:: :meth:`scipy.special.log1p`
+
+    """,
+)
--- a/cupyx/scipy/special/_basic.py
+++ b/cupyx/scipy/special/_basic.py
@@ -1,10 +1,13 @@
 from cupy import _core
 
 
+# Note: cast complex<single> to complex<double> or tests fail tolerance
 log1p = _core.create_ufunc(
-    "cupyx_scipy_log1p",
-    ("f->f", "d->d"),
-    "out0 = out0_type(log1p(in0));",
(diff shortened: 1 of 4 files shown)


cupyx/scipy/special/_basic.py
```diff
@@ -0,0 +1,19 @@
+from cupy import _core
+
+
+log1p = _core.create_ufunc(
+    "cupyx_scipy_log1p",
+    ("f->f", "d->d"),
+    "out0 = out0_type(log1p(in0));",
```
asi1024, line 7: We can support also complex type inputs.
https://github.com/cupy/cupy/blob/master/cupy/_core/include/cupy/complex.cuh#L43

asi1024:
```suggestion
    (('f->f', 'out0 = log1pf(in0)'), "d->d', 'F->F', 'D->D'),
    "out0 = log1p(in0);",
```
How about using `log1pf` for float32 inputs?

grlee77: Thanks, I have updated to use that.

For complex-valued inputs, this does work okay, but the implementation in SciPy is not simply `log(x + 1)`. For reference the `clog1p` implementation that gets used in the complex case is  [defined here](https://github.com/scipy/scipy/blob/8f640fd997e0386bdd4525a03869c45c726ac3cc/scipy/special/_cunity.pxd#L25-L66). There is even one branch there that even uses long double precision. 

With a simple `F->F` as suggested above, the logspace tests were failing for me with `complex64` outside of `logspace(-2.5, 2.5)` whereas with double precision it passes until close to `logspace(-12, 12)`. For that reason, I currently have it cast to complex128 for complex64 inputs. I guess `log1pf` is doing something more numerically accurate than `logf(1 + x)` internally so the `float32` case didn't have the same accuracy issue.

asi1024 (approved):

asi1024:
@grlee77 LGTM. Thanks!

PR #6317, opened 2022-01-14
Add special functions from the CUDA Math API

grlee77:
This is a small PR adding simple ufunc wrappers for a handful of other CUDA Math API functions that overlap with functions in `cupyx.scipy.special`.

One question:

- For sinc: `scipy.special.sinc` **is** `numpy.sinc`. This function promotes any integer dtype to double while the implementation here does not. For example, for `cupy.sinc` as used here, providing `cp.int8` input will result in a `cp.float16` array being returned. Should we make a wrapper that promotes any ints to `cp.float64` for consistency with SciPy?


grlee77:
I guess the since question is not specific to the special namespace. In general it is `cupy.sinc` that behaves differently from `numpy.sinc` in regards to integer type promotions.

grlee77:
Please test again (the prior failure in `test_yn` should now be fixed).

asi1024 (approved):

asi1024:
LGTM! Thanks!

PR #6288, opened 2022-01-04
add discrete cosine and sine transforms to cupyx.scipy.fft

grlee77:
This PR implements type II and III DCT and DST transforms. Since the corresponding real-to-real transforms are not implemented in cuFFT, the implementations here rely on manipulating the output of standard FFTs of equivalent length (see the references at the top of `_realtransforms.py` in this PR). This is a bit less efficient than if cuFFT supplied these transforms directly, but will still be fast relative to a CPU-based implementation. 

cc @IvanYashchuk for awareness

Issue #5437, opened 2021-07-05
[RFC] Use NumPy-style instead of Google-style for docstrings


kmaehashi:
Currently, there is no policy on CuPy docstring styles. The majority of docstrings use Google style, but [some](https://github.com/cupy/cupy/blob/v10.0.0a2/cupy/lib/stride_tricks.py#L4-L34) uses NumPy style. 
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#google-vs-numpy

I think it's better to have a policy on it, and proposing to use NumPy style over Google style.
Pros of NumPy styles over Google style are:
* Supports multiple return arguments (mentioned by @toslunar in #5433).
* Less indentation level, i.e., less lines. It's easy to write and maintain especially when the docstring is long.
* NumPy/SciPy uses it 😄, i.e., we can copy-paste from the upstream docstring when implementing a new function.

kmaehashi:
Seems no objections 😄 
I've added the following to the [Development Policy](https://github.com/cupy/cupy/wiki/Development-policies)

> Docstrings
> Use NumPy-style documentation. Currently, most functions are using Google-style but we will gradually migrate to the NumPy-style. (see https://github.com/cupy/cupy/issues/5437)

PR #6318, opened 2022-01-14
Add beta functions to `cupyx.scipy.special`

grlee77:
This PR adds four related functions from cupyx.scipy.special:  `beta`, `betaln`, `betainc` and `betaincinv`.
It also adds `ndtri` since that is just a simple call to CUDAs `normcdfinv`.

As with the spherical harmonics PR, the majority of the lines here involve porting Cephes C library code to CUDA kernels. This was done with minimal changes, just commenting out some calls to `sf_error`, swapping NYP_INF->CUDART_INFINITY, etc. and fixing up indentation a bit.

I refactored a few of the functions to remove a handful of `goto` statements, but the `incbi` implementation relies more strongly on goto and I didn't try to change that here.

I initially had tried using a simpler `lgam_sgn` in the first commit, but that version did not pass the test suite over the full range tested here.


cupyx/scipy/special/_beta.py
```diff
@@ -0,0 +1,1113 @@
...
+
+__device__ static double beta_negint(int a, double b)
+{
+    int sgn;
+    if (b == (int)b && 1 - a - b > 0) {
+        sgn = ((int)b % 2 == 0) ? 1 : -1;
+        return sgn * beta(1 - a - b, b);
+    }
+    else {
+        // sf_error("lbeta", SF_ERROR_OVERFLOW, NULL);
```
emcastillo, line 219: nitpick here, can we remove unused comments?
Thanks

cupyx/scipy/special/_beta.py
```diff
@@ -0,0 +1,1113 @@
...
+    if (fabs(y) > MAXGAM || fabs(a) > MAXGAM || fabs(b) > MAXGAM) {
+        int sgngam;
+        y = lgam_sgn(y, &sgngam);
+        sign *= sgngam;     /* keep track of the sign */
+        y = lgam_sgn(b, &sgngam) - y;
+        sign *= sgngam;
+        y = lgam_sgn(a, &sgngam) + y;
+        sign *= sgngam;
+        if (y > MAXLOG) {
+            // goto overflow;
```
emcastillo, line 273: ditto

cupyx/scipy/special/_beta.py
```diff
@@ -0,0 +1,1113 @@
...
+        y = a / y;
+        y *= b;
+    }
+
+    return (y);
+
+// overflow:
+//     sf_error("beta", SF_ERROR_OVERFLOW, NULL);
+//     return (sign * CUDART_INF);
+}
```
emcastillo, line 301: ditto

cupyx/scipy/special/_beta.py
```diff
@@ -0,0 +1,1113 @@
...
+
+__device__ static double lbeta_negint(int a, double b)
+{
+    double r;
+    if (b == (int)b && 1 - a - b > 0) {
+        r = lbeta(1 - a - b, b);
+        return r;
+    }
+    else {
+        // sf_error("lbeta", SF_ERROR_OVERFLOW, NULL);
```
emcastillo, line 322: ditto

cupyx/scipy/special/_beta.py
```diff
@@ -0,0 +1,1113 @@
...
+        polevl_definition +
+        p1evl_definition +
+        lgam_sgn_definition +
+        lbeta_symp_definition +
+        beta_definition
+    ),
+    doc="""Beta function.
+
+    Parameters
+    ----------
```
emcastillo, line 425: Let's match docstring format to CuPy style :)

emcastillo (approved):
I only have two minor issues
 - Commented code
 - Docstrings format

The rest is great :). Thanks!

cupyx/scipy/special/_beta.py
grlee77: I do see that I have an inconsistent mixture of both in the same docstring here where `.. seealso:: `` from the CuPy style was used when the rest of the docstring is in NumPy style, so I definitely need to make it self-consistent!

However, based on these comments, I thought NumPy style was being encouraged now? 
https://github.com/cupy/cupy/issues/5437#issuecomment-879148251
https://github.com/cupy/cupy/pull/6288#issuecomment-1010609903

For this PR it is not difficult to update to either style, please just confirm the preference so I can make any future PRs in the preferred style.

emcastillo: Sorry for the late reply.

Oops, I completely overlooked this. Then it is ok to use NumPy style!

Thanks for the pointer and sorry for the confusion

grlee77:
Interesting, the failure here looks like it is a timeout during compilation: https://ci.preferred.jp/cupy.linux.cuda102/93451/#L4617. I wonder what is making the compilation so long in this case (`betaincinv` is does involve several functions, but it doesn't seem THAT large).

grlee77:
One oddity of that particular function is all of the `goto` statements that I had not refactored. Not sure if those can increase compilation time (perhaps it makes optimization passes more difficult?)

grlee77:
I do not observe the extended compilation time locally. After clearining the kernel cache and running the following set of all `betaincinv` tests:

`pytest tests/cupyx_tests/scipy_tests/special_tests/test_beta.py -v -k incinv`

completed in 34.6 seconds (1.8 seconds once the kernels were cached).

The above were on a linux system with CUDA 11.2, so not the same CUDA version as the failed test.

grlee77:
Maybe just try running the tests again and see if this failure was a fluke?

grlee77:
Hmmm, it failed in the same way again (only for CUDA 10.2 on linux). I don't have a CUDA 10.2 install to debug with. Should we mark this as XFail or are there other ideas?

kmaehashi:
Let me run the full test to try isolate the issue.

/test full

grlee77:
Okay, I think I can probably fix this by specifying `__noinline__` on some more of the functions. Some specific `.cubin` files here end up quite large (500k-1MB). The CUDA source is still less than 1000 lines, but ends up generating PTX code that is > 60,000 lines which is quite large! 

The same is true for the `gammainccinv` function merged previously. I tried experimenting locally with adding `__noinline__` to more functions and was able to reduce the cubin size for that by a factor of about 8. I will try now to see how aggressive I need to be with that and if it causes much performance impact.

grlee77:
Okay, locally the `.cubin` files related to beta functions all come out to <100 kB now and compilation completes in about half the time when testing locally. 

Please try testing again

grlee77:
```Python
import cupy as cp
from cupy import testing
from cupyx.scipy import special
from cupyx.time import repeat

a = testing.shaped_arange((1, 1000, 1), cp, cp.float32)
b = testing.shaped_arange((1000, 1, 1), cp, cp.float32)
x = cp.asarray([0, 0.25, 0.5, 0.75, 1], dtype=cp.float32).reshape(1, 1, 5)

perf = repeat(special.betaincinv, (a, b, x), n_warmup=10, n_repeat=10)
print(perf.gpu_times.mean())
```

The above example gave me 318-320 ms both before and after the change here (on a GTX 1080 Ti). The same thing in SciPy is 3472 ms.

emcastillo (approved):

PR #6321, opened 2022-01-16
avoid DeprecationWarning from SciPy 1.8 (`cupyx.scipy.sparse`)

grlee77:
This small change is to avoid a new `DeprecationWarning` I noticed when using SciPy 1.8.0rc2

```
DeprecationWarning: Please use `spmatrix` from the `scipy.sparse` namespace, the `scipy.sparse.base` namespace is deprecated.
  from scipy.sparse.base import spmatrix
```

### DIFF
--- a/cupyx/scipy/sparse/_index.py
+++ b/cupyx/scipy/sparse/_index.py
@@ -14,7 +14,7 @@ from cupy.cuda import runtime
 import numpy
 
 try:
-    import scipy
+    import scipy.sparse
     scipy_available = True
 except ImportError:
     scipy_available = False
@@ -564,7 +564,7 @@ class IndexMixin(object):
 
 def _try_is_scipy_spmatrix(index):
     if scipy_available:
-        return isinstance(index, scipy.sparse.base.spmatrix)
+        return isinstance(index, scipy.sparse.spmatrix)
     return False
 
 


cupyx/scipy/sparse/_index.py
```diff
@@ -14,7 +14,7 @@
 import numpy
 
 try:
-    import scipy
+    import scipy.sparse
```
grlee77, line 17: This change is because `scipy.sparse` is not imported when top-level `import scipy` is done. I think it happens to work as it was because `base.py` does call `import scipy.sparse`.

cupyx/scipy/sparse/_index.py
```diff
@@ -564,7 +564,7 @@ def _set_arrayXarray_sparse(self, row, col, x):
 
 def _try_is_scipy_spmatrix(index):
     if scipy_available:
-        return isinstance(index, scipy.sparse.base.spmatrix)
+        return isinstance(index, scipy.sparse.spmatrix)
```
grlee77, line 567: removing `.base` here is what fixes the `DeprecationWarning`

takagi (approved):

takagi:
Would you merge this PR as the CI failure is not related to this change, @kmaehashi? (I'll look into TestLOBPCG's failure aside)

takagi:
Thanks! @grlee77 @kmaehashi

PR #6254, opened 2021-12-17
Add array_equiv API.

khushi-411:
Hi, CuPy Team,
The **aim** of the PR is to add the `array_equiv` API. It follows up https://github.com/cupy/cupy/issues/6078#issue-1051191869.

Looking forward to your viewpoints. Thanks!

PR #6333, opened 2022-01-17
Use `get_include` instead of `array_equiv` for fallback test

kmaehashi:
`array_equiv` is going to be added in #6254.
Use `get_include`, which is unlikely to be added in CuPy, for fallback test.


toslunar (approved):
LGTM

PR #6337, opened 2022-01-17
Fix cuDNN installer not working

kmaehashi:
Follows-up: #6314
Slow test failed: https://ci.preferred.jp/cupy.linux.cuda-slow/91281/#L4316

Directory structure in cuDNN archive has been changed and installer was not working.


takagi (approved):
LGTM!

PR #6336, opened 2022-01-17
CI: use CUDA docker images for CUDA Python CI

kmaehashi:
This avoids using hand-crafted Dockerfile for CUDA Python CI for maintainability.

PR #6338, opened 2022-01-18
[backport] CI: use CUDA docker images for CUDA Python CI

kmaehashi:
Backport of #6336 by @kmaehashi


kmaehashi:
Difference from #6336: cutensor 1.3 is used instead of 1.4.

PR #6342, opened 2022-01-19
Fix for cuDNN directory structure in Windows

kmaehashi:
Bug added in #6337.

In cuDNN Windows archive,
* `lib` directory only contains static libraries
* `bin` directory contains dynamic libraries (DLL) which are what we need


asi1024 (approved):

kmaehashi:
pfnCI, run Jenkins please.

Issue #6319, opened 2022-01-15
Implement `__repr__` for `cupyx.profiler._time._PerfCaseResult`


kmaehashi:
### Description

Let's implement `__repr__` for interactive benchmarking.

```
>>> cupyx.profiler.benchmark(lambda: numpy.asarray_chkfinite(z), n_repeat=10, n_warmup=0)
<cupyx.profiler._time._PerfCaseResult object at 0x7f3025808c40>
```

Subhraneel77:
Can I Give It A Try?

kmaehashi:
@Subhraneel77 Sure!

HallowedDust5:
I want to contribute - is there any format that the returned string should have?

kmaehashi:
Fixed in #6617

PR #6378, opened 2022-01-25
Add __str__ implementation for cupyx.profiler._time._PerfCaseResult

HallowedDust5:
This is for issue #6319. This is my first open source contribution, so sorry if this isn't what was needed.


kmaehashi:
Hi, thanks for the contribution @HallowedDust5!
I overlooked that `__str__` is already implemented in #6319 😅 
As described in the original issue, the motivation is to provide a string representation in the REPL environment, so I think just renaming `__str__` to `__repr__` looks enough here.

PR #6382, opened 2022-01-27
Fix type annotations in installer

kmaehashi:
Adds mypy configuration at `installer/mypy.ini`.

### DIFF
--- a/install/cupy_builder/install_build.py
+++ b/install/cupy_builder/install_build.py
@@ -231,7 +231,7 @@ def _match_output_lines(output_lines, regexs):
     return None
 
 
-def get_compiler_base_options(compiler_path) -> List[str]:
+def get_compiler_base_options(compiler_path: List[str]) -> List[str]:
     """Returns base options for nvcc compiler.
 
     """
@@ -321,7 +321,7 @@ def check_cuda_version(compiler, settings):
     return True
 
 
-def get_cuda_version(formatted=False) -> int:
+def get_cuda_version(formatted: bool = False) -> int:
     """Return CUDA Toolkit version cached in check_cuda_version()."""
     global _cuda_version
     if _cuda_version is None:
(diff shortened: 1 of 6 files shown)


install/cupy_builder/install_build.py
```diff
@@ -228,7 +231,7 @@ def _match_output_lines(output_lines, regexs):
     return None
 
 
-def get_compiler_base_options(compiler_path):
+def get_compiler_base_options(compiler_path) -> List[str]:
```
emcastillo, line 234:
```suggestion
def get_compiler_base_options(compiler_path: str) -> List[str]:
```

install/cupy_builder/install_build.py
```diff
@@ -318,7 +321,7 @@ def check_cuda_version(compiler, settings):
     return True
 
 
-def get_cuda_version(formatted=False):
+def get_cuda_version(formatted=False) -> int:
```
emcastillo, line 324:
```suggestion
def get_cuda_version(formatted: bool = False) -> int:
```

install/cupy_builder/install_build.py
```diff
@@ -356,7 +359,7 @@ def check_hip_version(compiler, settings):
     return True
 
 
-def get_hip_version(formatted=False):
+def get_hip_version(formatted=False) -> int:
```
emcastillo, line 362:
```suggestion
def get_hip_version(formatted: bool = False) -> int:
```

install/cupy_builder/install_build.py
```diff
@@ -400,7 +403,7 @@ def check_compute_capabilities(compiler, settings):
     return True
 
 
-def get_compute_capabilities(formatted=False):
+def get_compute_capabilities(formatted=False) -> Set[int]:
```
emcastillo, line 406:
```suggestion
def get_compute_capabilities(formatted: bool = False) -> Set[int]:
```

emcastillo (commented):
Some types were missing in function args 😁

install/cupy_builder/install_build.py
kmaehashi, line 234: Thanks, but surprisingly this is a list 😇 

```suggestion
def get_compiler_base_options(compiler_path: List[str]) -> List[str]:
```

(That's why we need a refactoring in installer...)

emcastillo (approved):

PR #6389, opened 2022-01-27
Eliminate unnecessary configuration pass in setup

kmaehashi:
This PR reduces the installation time by making the module configuration (`get_ext_modules`) happen only once during the setup process.

Currently, module configuration runs three times when doing `pip install -e .`.
`pip` launches `setup.py` two times:

* `setup.py egg_info` (configuration happens at [setup.py](https://github.com/cupy/cupy/blob/v11.0.0a1/setup.py#L85))
* `setup.py develop` (configuration happens at [setup.py](https://github.com/cupy/cupy/blob/v11.0.0a1/setup.py#L85) & [custom_build_ext](https://github.com/cupy/cupy/blob/v11.0.0a1/install/cupy_builder/cupy_setup_build.py#L796))

However, the first process is only for generating `egg-info` (to list setup requirements etc.), and does not need module configuration (https://github.com/cupy/cupy/issues/5722#issuecomment-1004566726). The second process can also be improved to run configuration only once. This PR fixes the both.

In my environment, `pip install -e .` overhead (i.e., installation time without any changes to the source tree) reduced from 22 sec to 12 sec.

Depends on ~#6382~ (merged)

### DIFF
--- a/install/cupy_builder/_command.py
+++ b/install/cupy_builder/_command.py
@@ -91,16 +91,14 @@ class custom_build_ext(setuptools.command.build_ext.build_ext):  # type: ignore[
         print('Cythonizing...')
         cythonize(self.extensions, cupy_builder.get_context())
 
-        # Rewrite source filenames of each extension from "*.pyx" to "*.cpp".
+        # Change an extension in each source filenames from "*.pyx" to "*.cpp".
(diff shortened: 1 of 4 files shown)


emcastillo (approved):
Nice!　I didn`t know `splitext` 😃

PR #6390, opened 2022-01-27
Remove `CUPY_SETUP_ENABLE_THRUST=0` environment variable

kmaehashi:
Bug added in #5743.
(No one complained... anyone still need this?)

Update: as discussed below this PR removes `CUPY_SETUP_ENABLE_THRUST=0` which was not working since #5743.


emcastillo:
I think we can remove this already, the parallel build and cc selection achieves the same result in development environments.

https://github.com/cupy/cupy/pull/2730

emcastillo (approved):

emcastillo (approved):

PR #6391, opened 2022-01-27
Remove global `use_hip` flag in setup

kmaehashi:


install/cupy_builder/install_build.py
```diff
@@ -23,8 +23,6 @@
 _rocm_path = 'NOT_INITIALIZED'
 _compiler_base_options = None
 
-use_hip = bool(int(os.environ.get('CUPY_INSTALL_USE_HIP', '0')))
```
leofang, line 26: Just curious, this env var is no longer needed?

emcastillo: It is present in [`context`](https://github.com/cupy/cupy/blob/master/install/cupy_builder/_context.py#L22-L23)

emcastillo:
Seems we need to fix tests :)
```
@pytest.mark.skipif(build.use_hip, reason='For CUDA environment') |  
-- | --
00:16:41.305516 STDOUT 1592] | E   AttributeError: module 'cupy_builder.install_build' has no attribute 'use_hip'
```

install/cupy_builder/install_build.py
kmaehashi: The code was under refactoring and there were two flags for `CUPY_INSTALL_USE_HIP` (here and context).

leofang: I see, I missed that!

emcastillo (approved):

PR #6401, opened 2022-01-30
Added dtype, fweights, aweights to cupy.cov

BasLaa:
**Note**
I have not been able to get the unit tests running on my computer, so adding unit tests and performing them still needs to be done. Apologies for this.

The arguments dtype, fweights and aweights have been added to the cupy.cov method. Furthermore, the dtype argument has been added to the corrcoef method. The ordering of arguments follows the numpy implementation

### DIFF
--- a/cupy/_statistics/correlation.py
+++ b/cupy/_statistics/correlation.py
@@ -78,7 +78,8 @@ def correlate(a, v, mode='valid'):
     return out
 
 
-def cov(a, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=None, *, dtype=None):
+def cov(a, y=None, rowvar=True, bias=False, ddof=None,
+        fweights=None, aweights=None, *, dtype=None):
     """Returns the covariance matrix of an array.
 
     This function currently does not support ``fweights`` and ``aweights``
@@ -100,9 +101,9 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=N
         fweights (cupy.ndarray, int): 1-D array of integer frequency weights.
             the number of times each observation vector should be repeated.
         aweights (cupy.ndarray): 1-D array of observation vector weights.
-            These relative weights are typically large for observations 
-            considered "important" and smaller for observations considered 
-            less "important". If ``ddof=0`` the array of weights can be used 
+            These relative weights are typically large for observations
+            considered "important" and smaller for observations considered
+            less "important". If ``ddof=0`` the array of weights can be used
             to assign probabilities to observation vectors.
         dtype: Data type specifier. By default, the return data-type will have
             at least `numpy.float64` precision.
@@ -125,8 +126,10 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=N
         else:
             if y.ndim > 2:
                 raise ValueError('y must be <= 2-d')
-            dtype = functools.reduce(numpy.promote_types,
-                                    (a.dtype, y.dtype, numpy.float64))
+            dtype = functools.reduce(
+                numpy.promote_types,
+                (a.dtype, y.dtype, numpy.float64)
+            )
 
     X = cupy.array(a, ndmin=2, dtype=dtype)
     if not rowvar and X.shape[0] != 1:
@@ -158,7 +161,7 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=N
             raise ValueError(
                 "fweights cannot be negative")
         w = fweights
-    
+
     if aweights is not None:
         aweights = cupy.asarray(aweights, dtype=float)
         if aweights.ndim > 1:
--- a/cupy/_statistics/correlation.py
+++ b/cupy/_statistics/correlation.py
@@ -109,8 +109,8 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None,
             considered "important" and smaller for observations considered
             less "important". If ``ddof=0`` the array of weights can be used
             to assign probabilities to observation vectors.
-            It is required that fweights >= 0. However, the function will not
-            error when fweights < 0 for performance reasons.
+            It is required that aweights >= 0. However, the function will not
+            error when aweights < 0 for performance reasons.
         dtype: Data type specifier. By default, the return data-type will have
             at least `numpy.float64` precision.
 
--- a/cupy/_statistics/correlation.py
+++ b/cupy/_statistics/correlation.py
@@ -19,6 +19,8 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None, *, dtype=None):
             transposed.
         bias (None): Has no effect, do not use.
         ddof (None): Has no effect, do not use.
+        dtype: Data type specifier. By default, the return data-type will have
+            at least `numpy.float64` precision.
 
     Returns:
         cupy.ndarray: The Pearson product-moment correlation coefficients of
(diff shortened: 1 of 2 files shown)


cupy/_statistics/correlation.py
```diff
@@ -31,7 +31,7 @@ def corrcoef(a, y=None, rowvar=True, bias=None, ddof=None):
         warnings.warn('bias and ddof have no effect and are deprecated',
```
emcastillo, line 31: Please update the docstring

cupy/_statistics/correlation.py
```diff
@@ -131,13 +145,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
     if ddof is None:
         ddof = 0 if bias else 1
 
-    fact = X.shape[1] - ddof
+    w = None
+    if fweights is not None:
+        fweights = cupy.asarray(fweights, dtype=float)
```
emcastillo, line 150: Let's add some typechecks here, fweights should only be `int` or `cupy.ndarray`.
Other types are forbidden.

cupy/_statistics/correlation.py
```diff
@@ -131,13 +145,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
...
+        if not cupy.all(fweights == cupy.around(fweights)):
```
emcastillo, line 151: This causes device synchronization. It is better to check just the type info, `fweights.dtype.char in 'bBhHiIlLqQ'`

cupy/_statistics/correlation.py
```diff
@@ -131,13 +145,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
...
+        if fweights.shape[0] != X.shape[1]:
+            raise RuntimeError(
+                "incompatible numbers of samples and fweights")
+        if any(fweights < 0):
+            raise ValueError(
+                "fweights cannot be negative")
+        w = fweights
+
+    if aweights is not None:
+        aweights = cupy.asarray(aweights, dtype=float)
```
emcastillo, line 166: ditto

emcastillo:
If you write a test we can run it in the CI or I can run it for you If needed.

BasLaa:
Got it, will send you an update asap.

cupy/_statistics/correlation.py
BasLaa: Which part exactly causes device sync?

cupy/_statistics/correlation.py
BasLaa: There is a typecheck in the next line though, `if not cupy.all(fweights == cupy.around(fweights)):`, which I will replace with `fweights.dtype.char in 'bBhHiIlLqQ'`. Is that sufficient for typecheck or should there be something before line 150 here?

BasLaa: Although now that I think about it, line 150 sets the dtype to float, so `fweights.dtype.char in 'bBhHiIlLqQ'` would not work after that anymore. How can I best structure this?

emcastillo: do the check before, and then with `astype` cast it to float if needed. Probably the numpy casting rules will do it for you if you use it later.

cupy/_statistics/correlation.py
```diff
@@ -131,13 +145,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
...
+        if not cupy.all(fweights == cupy.around(fweights)):
+            raise TypeError(
+                "fweights must be integer")
+        if fweights.ndim > 1:
+            raise RuntimeError(
+                "cannot handle multidimensional fweights")
+        if fweights.shape[0] != X.shape[1]:
+            raise RuntimeError(
+                "incompatible numbers of samples and fweights")
+        if any(fweights < 0):
```
emcastillo, line 160: This also causes device synchronization. 
You apply an operation `<` to all the elements of the array in the GPU, and then move the result to the CPU to execute the python `any` to drive that `if`.

To avoid this we are flexible and allow negative results although it can produce errors, just a note in the docstring saying that the function does not error with negative values for performance should be enough.

BasLaa: This would also apply to `if any(aweights < 0)` then.

BasLaa:
@emcastillo could you have a look? I'm not sure how to implement the testing for errors, as it currently uses a test function checking for ValueError. Should I make equivalent functions for other errors?

cupy/_statistics/correlation.py
```diff
@@ -131,13 +151,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
     if ddof is None:
         ddof = 0 if bias else 1
 
-    fact = X.shape[1] - ddof
+    w = None
+    if fweights is not None:
+        if not type(fweights) == cupy.ndarray:
```
emcastillo, line 156:
```suggestion
        if not isinstance(fweights, cupy.ndarray)
```

cupy/_statistics/correlation.py
```diff
@@ -131,13 +151,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
...
+        if fweights.ndim > 1:
+            raise RuntimeError(
+                "cannot handle multidimensional fweights")
+        if fweights.shape[0] != X.shape[1]:
+            raise RuntimeError(
+                "incompatible numbers of samples and fweights")
+        w = fweights
+
+    if aweights is not None:
+        if not type(aweights) == cupy.ndarray:
```
emcastillo, line 172: ditto

cupy/_statistics/correlation.py
```diff
@@ -131,13 +151,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
...
+
+    # Determine the normalization
+    if w is None:
+        fact = X.shape[1] - ddof
+    elif ddof == 0:
+        fact = w_sum
+    elif aweights is None:
+        fact = w_sum - ddof
+    else:
+        fact = w_sum - ddof*sum(w*aweights)/w_sum
```
emcastillo, line 198:
```suggestion
        fact = w_sum - ddof * sum(w * weights) / w_sum
```
formatting

cupy/_statistics/correlation.py
```diff
@@ -131,13 +151,62 @@ def cov(a, y=None, rowvar=True, bias=False, ddof=None):
...
         warnings.warn('Degrees of freedom <= 0 for slice',
                       RuntimeWarning, stacklevel=2)
         fact = 0.0
 
     X -= X.mean(axis=1)[:, None]
-    out = X.dot(X.T.conj()) * (1 / cupy.float64(fact))
+    if w is None:
+        X_T = X.T
+    else:
+        X_T = (X*w).T
```
emcastillo, line 209:
```suggestion
        X_T = (X * w).T
```

emcastillo:
if errors happens in both, numpy and cupy is ok, the decorators will compare both.
Let me run the tests before committing the suggestions

tests/cupy_tests/statistics_tests/test_correlation.py
```diff
@@ -36,6 +36,13 @@ def test_corrcoef_rowvar(self, xp, dtype):
         y = testing.shaped_arange((2, 3), xp, dtype)
         return xp.corrcoef(a, y=y, rowvar=False)
 
+    @testing.for_all_dtypes()
```
emcastillo, line 39:
```suggestion
    @testing.with_requires('numpy>=1.20')
    @testing.for_all_dtypes()
```

tests/cupy_tests/statistics_tests/test_correlation.py
```diff
@@ -50,25 +57,31 @@ def generate_input(self, a_shape, y_shape, xp, dtype):
     @testing.for_all_dtypes()
```
emcastillo, line 57:
```suggestion
    @testing.with_requires('numpy>=1.20')
    @testing.for_all_dtypes()
```

emcastillo:
Sorry to keep asking,

Tests require numpy >= 1.20, since we test with many different numpy versions, we don't want to skip tests for all combinations.

Can you decouple the new tests from the old ones and name them something like `test_name_with_fweights`
and add `@testing.with_requires('numpy>=1.20')` to the ones using `dtype`?

emcastillo:
Closing, feel free to reopen

PR #6406, opened 2022-01-31
Fix `cupyx.ndimage.spline_filter1d` for HIP

emcastillo:
Fixes an issue with some `cupyx.scipy.ndimage` routines that relied in `spline_filter` for input preprocessing in HIP environments.

This issue is similar to #4504, some of the values are not being updated in memory unless thread synchronization happens.

cc @grlee77 for future reference :)

### DIFF
--- a/cupyx/scipy/ndimage/_spline_prefilter_core.py
+++ b/cupyx/scipy/ndimage/_spline_prefilter_core.py
@@ -169,7 +169,9 @@ def _get_spline1d_code(mode, poles, n_boundary):
         for (i = 1; i < n; ++i) {{
             c[i * element_stride] += z * c[(i - 1) * element_stride];
         }}''')
-
+        code.append('''
+        __syncthreads();
+        ''')
         # initialize and apply the anti-causal filter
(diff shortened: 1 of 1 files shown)


cupyx/scipy/ndimage/_spline_prefilter_core.py
```diff
@@ -169,7 +169,9 @@ def _get_spline1d_code(mode, poles, n_boundary):
         for (i = 1; i < n; ++i) {{
             c[i * element_stride] += z * c[(i - 1) * element_stride];
         }}''')
-
+        code.append('''
+        __syncthreads();
+        ''')
```
grlee77, line 174: should this have an ifdef here as in #4504?

```suggestion
        code.append('''
        #ifdef __HIP_DEVICE_COMPILE__
        __syncthreads();
        #endif
        ''')
```

emcastillo: Thank you!

asi1024 (approved):

PR #6408, opened 2022-02-01
Fix `cupy.nan_to_num`

### LINKED ISSUE #6400: cupy.nan_to_num does not accept cupy.nan
### Description

After upgrading to `cupy` v10, some calls like `cupy.nan_to_num(cupy.nan)` are failing at [this line](https://github.com/cupy/cupy/pull/5870/files#diff-c8d13a3653ab59bc655bb3f6bf2e749e8a475a2fb06133bf39fcd7f1319585fcR381).

`cupy.nan` is a `float` and so does not have a `.dtype` attribute.

I guess that this is desired as it removes an implicit memory transfer, however, this is valid in `numpy`.

If this is intentional I can work around it, but it seems counter-intuitive to me.

Thanks!

### To Reproduce

```py
>>> import cupy
>>> cupy.nan_to_num(cupy.nan)
  File "/home/colm.talbot/.conda/envs/dev-39/lib/python3.9/site-packages/cupy/_math/misc.py", line 383, in nan_to_num
    dtype = x.dtype
AttributeError: 'float' object has no attribute 'dtype'
>>> cupy.nan_to_num(0)
  File "/home/colm.talbot/.conda/envs/dev-39/lib/python3.9/site-packages/cupy/_math/misc.py", line 383, in nan_to_num
    dtype = x.dtype
AttributeError: 'int' object has no attribute 'dtype'
```

by contrast, this is the `numpy` behaviour

```py
>>> import numpy
>>> numpy.nan_to_num(numpy.nan)
0.0
>>> numpy.nan_to_num(0)
0
```

the old behaviour

```py
>>> import cupy
>>> cupy.nan_to_num(cupy.asanyarray(cupy.nan))
array(0.)
```

### Installation

### Environment

```
OS                           : Linux-3.10.0-1160.49.1.el7.x86_64-x86_64-with-glibc2.17
Python Version               : 3.9.9
CuPy Version                 : 10.1.0
NumPy Version                : 1.22.1
Cython Build Version         : 0.29.26
CUDA Build Version           : 11020
CUDA Driver Version          : 11050
CUDA Runtime Version         : 11060
cuFFT Version                : 10600
cuRAND Version               : 10209
cuSOLVER Version             : (11, 3, 2)
NVRTC Version                : (11, 6)
Thrust Version               : 101000
CUB Build Version            : 101000
Jitify Build Version         : 7ed41e6
cuDNN Build Version          : 8201
cuDNN Version                : 8301
cuTENSOR Version             : 10303
Device 0 Name                : NVIDIA A100-PCIE-40GB
Device 0 Compute Capability  : 80
Device 0 PCI Bus ID          : 0000:C1:00.0
Device 1 Name                : NVIDIA GeForce RTX 3080
Device 1 Compute Capability  : 86
Device 1 PCI Bus ID          : 0000:01:00.0```


### Additional Information

_No response_

### PR DESCRIPTION
**Pull request by takagi**
Close #6400.

### DIFF
--- a/cupy/_math/misc.py
+++ b/cupy/_math/misc.py
@@ -388,11 +388,14 @@ def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None):
     .. seealso:: :func:`numpy.nan_to_num`
 
     """
-    dtype = x.dtype
+    if not isinstance(x, cupy.ndarray):
+        out = cupy.full((), x)
+    else:
+        out = cupy.empty_like(x) if copy else x
+    dtype = out.dtype
     nan = _check_nan_inf(nan, dtype)
     posinf = _check_nan_inf(posinf, dtype, False)
     neginf = _check_nan_inf(neginf, dtype, True)
-    out = None if copy else x
     return _nan_to_num(x, nan, posinf, neginf, out=out)
 
 
--- a/tests/cupy_tests/math_tests/test_misc.py
+++ b/tests/cupy_tests/math_tests/test_misc.py
@@ -1,5 +1,4 @@
 import sys
-import unittest
 
 import numpy
 import pytest
@@ -8,8 +7,7 @@ import cupy
 from cupy import testing
 
 
-@testing.gpu
-class TestMisc(unittest.TestCase):
+class TestMisc:
 
     @testing.for_all_dtypes()
     @testing.numpy_cupy_allclose(atol=1e-5)
@@ -87,8 +85,8 @@ class TestMisc(unittest.TestCase):
                      dtype=dtype)
         return getattr(xp, name)(a, b)
 
-    @unittest.skipIf(
-        sys.platform == 'win32', 'dtype problem on Windows')
+    @pytest.mark.skipIf(
+        sys.platform == 'win32', reason='dtype problem on Windows')
     @testing.for_all_dtypes(no_complex=True)
     @testing.numpy_cupy_array_equal()
     def test_clip1(self, xp, dtype):
@@ -120,8 +118,8 @@ class TestMisc(unittest.TestCase):
             with pytest.raises(ValueError):
                 a.clip(None, None)
 
-    @unittest.skipIf(
-        sys.platform == 'win32', 'dtype problem on Windows')
+    @pytest.mark.skipIf(
+        sys.platform == 'win32', reason='dtype problem on Windows')
     @testing.for_all_dtypes(no_complex=True)
     @testing.numpy_cupy_array_equal()
     def test_external_clip1(self, xp, dtype):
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: cupy/_math/misc.py [hunk 1]
```diff
@@ -388,12 +384,21 @@ def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None):
     .. seealso:: :func:`numpy.nan_to_num`
 
     """
-    dtype = x.dtype
-    nan = _check_nan_inf(nan, dtype)
-    posinf = _check_nan_inf(posinf, dtype, False)
-    neginf = _check_nan_inf(neginf, dtype, True)
+    if copy and isinstance(x, cupy.ndarray):
+        # Passing nan, posinf and neginf values with the same kind as x so
+        # that the ufunc works on the same dtype with x.
```
takagi, line 389: See: https://github.com/cupy/cupy/blob/e0fe38a8d6a68e53e76c9d99969154d583b89599/cupy/_core/_kernel.pyx#L978

toslunar (commented):
- Since `nan_to_num` does not have to be a ufunc, we can early-return on `x.dtype.kind in 'biu'`. In fact, for those dtypes, `nan_to_num(x, nan=object())` succeeds in NumPy.
- The correct broadcast here is to broadcast to `x.shape`, but the current fix may return larger shape (if copy=True).  I suggest `out = cupy.empty_like(x) if copy else x` rather than asserting the other args are scalars.  NumPy allows array-likes for the args `nan`, `posinf`, `neginf`, while the doc says "int, float, optional".

cupy/_math/misc.py
```diff
@@ -388,12 +384,22 @@ def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None):
...
+            return x.copy() if copy else x
+
+    posinf_supplied = posinf is not None
+    posinf = posinf if posinf is not None else 0
+    neginf_supplied = neginf is not None
+    neginf = neginf if neginf is not None else 0
+    if isinstance(x, cupy.ndarray):
+        out = cupy.empty_like(x) if copy else x
+    else:
+        out = cupy.full((), x)
```
toslunar, line 400: If `out` is computed earlier, we can use `out.dtype`, `cupy.finfo(dtype).min`, and etc.

takagi: All right! Let me make the change based on the original code, force-pushing.

toslunar (approved):
LGTM

The performance improvement can be done in another PR.  As in your previous branch, `_check_nan_inf` could be replaced by some faster code.  For example, it works without `asanyarray`.
https://github.com/cupy/cupy/compare/4466e6d90d2df135f5838ae0d61dece8f8333b2d

PR #6413, opened 2022-02-01
Allow `__cuda_array_interface__` in HIP

emcastillo:
This should fix

```
cupy_tests/lib_tests/test_polynomial.py::TestPoly1dPolynomialArithmetic::test_poly1d_arithmetic[_param_8_{func=<lambda>, type_l='ndarray', type_r='poly1d'}]
cupy_tests/lib_tests/test_polynomial.py::TestPoly1dPolynomialArithmetic::test_poly1d_arithmetic[_param_20_{func=<lambda>, type_l='ndarray', type_r='poly1d'}]
cupy_tests/lib_tests/test_polynomial.py::TestPoly1dPolynomialArithmetic::test_poly1d_arithmetic[_param_32_{func=<lambda>, type_l='ndarray', type_r='poly1d'}]
```

tests in hip, we can actually use this but we have to resort to v2 to avoid issues with stream sync.

TODO: run tests with hip

PR #6414, opened 2022-02-02
Add `__cupy_get_ndarray__` dunder method to transform objects to arrays'

emcastillo:
Alternative to #6413

in HIP `__cuda_array_interface__` is not well defined, so it causes some functionality that relies on it such as polynomial API to fail where it shouldn't because it is being used as a hack to transform a regular object into a cupy ndarray in a really convoluted way.
We add a method to transform objects into ndarrays explicitly, avoiding the obscurity of the previous approach.

This PR fixes

```
cupy_tests/lib_tests/test_polynomial.py::TestPoly1dPolynomialArithmetic::test_poly1d_arithmetic[_param_8_{func=<lambda>, type_l='ndarray', type_r='poly1d'}]
cupy_tests/lib_tests/test_polynomial.py::TestPoly1dPolynomialArithmetic::test_poly1d_arithmetic[_param_20_{func=<lambda>, type_l='ndarray', type_r='poly1d'}]
cupy_tests/lib_tests/test_polynomial.py::TestPoly1dPolynomialArithmetic::test_poly1d_arithmetic[_param_32_{func=<lambda>, type_l='ndarray', type_r='poly1d'}]
```

### DIFF
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -19,7 +19,6 @@ from cupy.cuda cimport texture
 from cupy._core cimport _accelerator
 from cupy._core cimport _carray
 from cupy._core cimport _scalar
-from cupy._core.dlpack cimport from_dlpack
 from cupy._core._dtype cimport get_dtype
 from cupy._core._memory_range cimport may_share_bounds
 from cupy._core._scalar import get_typename as _get_typename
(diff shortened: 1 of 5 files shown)


leofang (commented):
I like it a lot more! Only nitpick: can we check `__dlpack__` before `__cuda_array_interface__`?

kmaehashi:
This will allow things like `cupy.sum(torch_tensor)` and wondering this is against the DLPack design that requires users to explicitly use `from_dlpack`.

emcastillo:
changed the pr to not use dlpack and add a cupy specific private protocol to achieve the same.

kmaehashi (approved):
LGTM!

PR #6420, opened 2022-02-03
Drop support for ROCm 4.0

emcastillo:
Since the test segfaults at the middle, lots of tests fails, and ROCm 4.0 is buggy and mostly discouraged, we should stop wasting resources on it :).


cupy/_environment.py
```diff
@@ -421,7 +421,6 @@ def _detect_duplicate_installation():
         'cupy-cuda113',
         'cupy-cuda114',
         'cupy-cuda115',
-        'cupy-rocm-4-0',
```
kmaehashi, line 424: Please keep all packages here :)

kmaehashi:
LGTM!
Could you also update list of supported versions? https://docs.cupy.dev/en/latest/install.html#id1

PR #6438, opened 2022-02-07
mypy: array_api

asi1024:
Depends on #6486.

### DIFF
--- a/cupy/array_api/_array_object.py
+++ b/cupy/array_api/_array_object.py
@@ -1,5 +1,3 @@
-# mypy: ignore-errors
-
 """
 Wrapper class around the ndarray object for the array API standard.
 
@@ -40,6 +38,7 @@ if TYPE_CHECKING:
 
 import cupy as np
 from cupy.cuda import Device as _Device
+from cupy.cuda import stream as stream_module
 from cupy_backends.cuda.api import runtime
 
 from cupy import array_api
@@ -65,7 +64,7 @@ class Array:
     # Use a custom constructor instead of __init__, as manually initializing
     # this class is not supported API.
     @classmethod
-    def _new(cls, x, /):
+    def _new(cls, x: Union[np.ndarray, np.generic], /) -> Array:
         """
         This is a private method for initializing the array API Array
         object.
@@ -164,7 +163,7 @@ class Array:
         return other
 
     # Helper function to match the type promotion rules in the spec
-    def _promote_scalar(self, scalar):
+    def _promote_scalar(self, scalar: Union[bool, int, float]) -> Array:
         """
         Returns a promoted version of a Python scalar appropriate for use with
         operations on self.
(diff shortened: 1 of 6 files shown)


kmaehashi:
Should we upstream this instead of fixing in CuPy?

toslunar (approved):
LGTM

leofang:
cc: @asmeurer @BvB93

cupy/array_api/_data_type_functions.py
```diff
@@ -96,7 +94,7 @@ def finfo(type: Union[Dtype, Array], /) -> finfo_object:
 
     See its docstring for more information.
     """
-    fi = np.finfo(type)
+    fi = np.finfo(type)  # type: ignore
```
BvB93, line 97: The errors for `finfo` and `iinfo` are more legit though. The issue stems from the `Array.dtype` property being annotated as [returning a union](https://github.com/cupy/cupy/blob/ec6d2c9a98440b19a133868576fb299972c21f6c/cupy/array_api/_typing.py#L56-L67), which _always_ includes both floating- and integer-types while the `iinfo`/`finfo` constructors only accept one category or the other. 
This is a demonstration why it's usually a bad idea to return these so-called "unsafe" unions, and I would recommend to simply switch the dtype parameter over to `Any`:

``` diff
diff --git a/cupy/array_api/_array_object.py b/cupy/array_api/_array_object.py
index c2ae0545d..d713e927a 100644
--- a/cupy/array_api/_array_object.py
+++ b/cupy/array_api/_array_object.py
@@ -1047,7 +1047,7 @@ class Array:
             return Array._new(arr)

     @property
-    def dtype(self) -> Dtype:
+    def dtype(self) -> np.dtype[Any]:
         """
         Array API compatible wrapper for :py:meth:`np.ndarray.dtype <numpy.ndarray.dtype>`.
```

cupy/array_api/_array_object.py
```diff
@@ -442,7 +441,7 @@ def __dlpack_device__(self: Array, /) -> Tuple[IntEnum, int]:
         """
         return self._array.__dlpack_device__()
 
-    def __eq__(self: Array, other: Union[int, float, bool, Array], /) -> Array:
+    def __eq__(self: Array, other: Union[int, float, bool, Array], /) -> Array:  # type: ignore
```
BvB93, line 444: Yeah, this is an unfortunate side-effect of being incompatible with the (somewhat overly strict) [`object.__eq__`/`__ne__`](https://github.com/python/typeshed/blob/9687d53b65d055a990f0187b7baa1137cb7c717b/stdlib/builtins.pyi#L102-L103) annotations. Nothing to do about this though except adding a `# type: ignore` comment or complain a bit more upstream...

BvB93 (commented):
Thanks for the ping @leofang!

PR #6439, opened 2022-02-08
Make `cupy.var` stable

### LINKED ISSUE #6425: Large array std is different from numpy (CPU)
I am trying to understand this behavior. When my array gets larger in size, the calculation for std is different from numpy even for obvious or basic array of ones. In fact, the difference between cupy std and numpy std exacerbates for arrays with completely random values. 

```python
import numpy as np
import cupy as cp

cp.cuda.Device(0).use()
x = np.ones((13000000, 269)).astype('f4')
x_cp = cp.array(x)

vcpu = x.std()
vgpu = x_cp.std()
vcpu,vgpu

>>> (0.0, array(0.9225852, dtype=float32))
```

Python version: 3.6
CUDA version: 10.2 & 11.1
cupy version: 9.6.0
OS: RedHat 7.4

emcastillo:
Fixes #6425 

For some reason, the calculated mean of a large array inside `var` routine may have some marginal error that is accumulated when calculating the variance leading to an incorrect result.

### DIFF
--- a/cupy/_core/_routines_statistics.pyx
+++ b/cupy/_core/_routines_statistics.pyx
@@ -559,7 +559,12 @@ cdef ndarray _var(
     # on why NaN is the result.
     div = max(items - ddof, 0)
     alpha = 1. / div if div != 0 else nan
-
+    # when using very large arrays, the mean calculation with low precision
+    # can lead to incorrect results when calculating the std. 
+    # Just by accomulating the (x_mean - x) we can notice the error.
+    # see #6425
+    if items > 10 ** 7:
+        dtype_mean = 'float64'
     arrmean = a.mean(axis=axis, dtype=dtype_mean, out=None, keepdims=True)
 
     if out is None:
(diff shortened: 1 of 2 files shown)


cupy/_core/_routines_statistics.pyx
```diff
@@ -559,7 +559,12 @@ cdef ndarray _var(
     # on why NaN is the result.
     div = max(items - ddof, 0)
     alpha = 1. / div if div != 0 else nan
-
+    # when using very large arrays, the mean calculation with low precision
+    # can lead to incorrect results when calculating the std.
```
emcastillo, line 563:
```suggestion
    # can lead to incorrect results when calculating the std.
```

emcastillo:
this error is also gone if using cub/cutensor

emcastillo:
The real error is regular CuPy reduction kernels with # of elements > 2.6 * 10 ^ 8 and `keepdims=True` and `dtype=cupy.float32`

```
x = cupy.ones(3 * 10**8, dtype=cupy.float32)
x.sum(keepdims=True) -> [2.6843546e+08]
x.sum(keepdims=False) -> 300000000.0
```

emcastillo:
Fixed by #6457

PR #6443, opened 2022-02-08
Revert "Merge pull request #6440 from kmaehashi/fix-scipy-intersphinx"

takagi:
SciPy's reverence looks back. https://docs.scipy.org/doc/scipy/reference


takagi:
Still fails as we can see https://docs.scipy.org/doc/scipy-1.8.0/html-scipyorg/objects.inv, but https://docs.scipy.org/doc/scipy/objects.inv just redirects to https://docs.scipy.org/doc/scipy-1.8.0/html-scipyorg/reference/ .

takagi:
See https://github.com/scipy/scipy/issues/15545.

takagi:
Let me withdraw this.

PR #6460, opened 2022-02-11
Add uarray backend support for `scipy.linalg`

Smit-create:
The `uarray` support for `scipy.linalg` is implemented in https://github.com/scipy/scipy/pull/15610 

This PR aims to add `cupy` backend support for `scipy.linalg`

### DIFF
--- a/cupyx/scipy/linalg/_special_matrices.py
+++ b/cupyx/scipy/linalg/_special_matrices.py
@@ -511,7 +511,8 @@ def fiedler_companion(a):
     return c
 
 
-@_uarray.implements('convolution_matrix')
+# TODO: uncomment the following line when scipy >= 1.5.0 is supported in tests
+# @_uarray.implements('convolution_matrix')
 def convolution_matrix(a, n, mode='full'):
     """Construct a convolution matrix.
 


Smit-create:
I think the PR is ready for 1st review.

tests/cupyx_tests/scipy_tests/linalg_tests/test_decomp_lu.py
```diff
@@ -5,8 +5,17 @@
 import cupy
 from cupy import testing
 import cupyx.scipy.linalg
-if cupyx.scipy._scipy_available:
-    import scipy.linalg
+try:
+    import scipy.linalg as scipy_linalg
+except ImportError:
+    scipy_linalg = None
```
toslunar, line 11: Please keep `scipy.linalg`.  In CuPy's tests, `scipy` or `cupyx.scipy` is often provided by `testing.numpy_cupy_*` and `import scipy.linalg` is required just because `import scipy` does not import the submodule.
```suggestion
try:
    import scipy.linalg
except ImportError:
    pass
```

toslunar (changes requested):

Smit-create:
The tests fail because `convolution_matrix` was introduced in `scipy` 1.5.0. See https://github.com/scipy/scipy/pull/12135

rgommers:
Hi @toslunar, all. This quick merge surprised me a little, I thought that would wait till after SciPy changes were merged. The `scipy.linalg` PR and the scope and usage of Uarray in SciPy (and scikit-learn / scikit-image) are still under discussion. I will try to write a larger summary asap.

This code does look good and is nicely concise, and doesn't break anywhere. However, it's very well possible that changes need to be made. We don't want to be in a position where we can't make those changes anymore because the CuPy backend is already merged. So I just want to check: do you want to keep these and make whatever matching changes are needed later, or do you want to revert this now and keep an open PR around? 

Also for the record: these changes, and corresponding Colab notebook that @Smit-create posted in one of his other PRs, are super useful to get a better sense for how everything fits together and easily play with CuPy + SciPy (+ scikits). That and the quick reviews here are much appreciated!

toslunar:
Hi @rgommers.

I should have made it clear that the feature is experimental.  I'm thinking of adding the following comment to `cupyx/scipy/linalg/__init__.py` so that SciPy can discuss its implementation without worrying about CuPy.
```diff
+# uarray backend support (NEP 31)
+# The uarray feature for scipy.linalg is experimental.
+# The interface can change in the future.
 from cupyx.scipy.linalg._uarray import __ua_convert__  # NOQA
 from cupyx.scipy.linalg._uarray import __ua_domain__  # NOQA
 from cupyx.scipy.linalg._uarray import __ua_function__  # NOQA
```

I'd like to keep the implementation in CuPy for a while, in order to suggest changes in CuPy-side implementation details (to resolve the postponed issue, for example).
```python
# TODO: uncomment the following line when scipy >= 1.5.0 is supported in tests
# @_uarray.implements('convolution_matrix')
```

kmaehashi:
We usually merge this kind of change after SciPy supports the corresponding feature. With that said, considering this PR itself is harmless and does not break any existing things, I think keeping this PR merged is acceptable. As this PR is only in the master (v11) branch and will not be backported to the stable v10 series, we have time to apply any changes or revert until the time of shipping v11 release-candidate, if needed.

PR #6461, opened 2022-02-11
cupyx.scipy.special: add statistical distributions

grlee77:
This PR adds many functions related to statistical distributions. There are many new functions here, but each is quite small (often <10 lines of GPU kernel code) because they mostly call previously implemented gamma and beta functions. They also often share a common testing logic. It was mainly implementing the test cases that was time consuming.

I have grouped these together in their own `_stats_distributions.py` file and moved the existing normal distribution functions into that file as well. I tried to have common logic where possible in the test classes, but there are occasionally subtle differences in dtype behavior (e.g. differences on behavior/warnings around unsafe casting of double->int32). The `ufunc.types` match and are in the same order as those in SciPy to ensure the same type casting behavior.

attn: @IvanYashchuk

### DIFF
--- a/tests/cupyx_tests/scipy_tests/special_tests/test_statistics.py
+++ b/tests/cupyx_tests/scipy_tests/special_tests/test_statistics.py
@@ -143,19 +143,19 @@ class TestFusionSpecial(_TestBase):
 
 class _TestDistributionsBase:
 
-    def _test_scalar(self, function, args, expected, rtol=None):
+    def _test_scalar(self, function, args, expected, rtol=1e-12, atol=1e-12):
         special = cupyx.scipy.special
         function = getattr(special, function)
-        if rtol is None:
-            testing.assert_array_equal(function(*args), expected)
-        else:
-            testing.assert_allclose(function(*args), expected, rtol=rtol)
+        testing.assert_allclose(function(*args), expected, rtol=rtol,
+                                atol=atol)
 
 
 @testing.gpu
 @testing.with_requires('scipy')
 class TestTwoArgumentDistribution(_TestDistributionsBase):
 
+    @pytest.mark.skipif(cp.cuda.runtime.is_hip,
+                        reason="avoid failures observed on HIP")
     @pytest.mark.parametrize('function', ['chdtr', 'chdtrc', 'chdtri',
                                           'pdtr', 'pdtrc', 'pdtri'])
     @testing.for_float_dtypes()
@@ -187,18 +187,18 @@ class TestTwoArgumentDistribution(_TestDistributionsBase):
         return func(k, y)
 
     @pytest.mark.parametrize(
-        'function, args, expected, rtol',
-        [('chdtr', (1, 0), 0.0, None),
-         ('chdtr', (0.7, cupy.inf), 1.0, None),
-         ('chdtr', (0.6, 3), 0.957890536704110, 1e-12),
-         ('chdtrc', (1, 0), 1.0, None),
-         ('chdtrc', (0.6, 3), 1 - 0.957890536704110, 1e-12),
-         ('chdtri', (1, 1), 0.0, None),
-         ('chdtri', (0.6, 1 - 0.957890536704110), 3, 1e-12),
+        'function, args, expected',
+        [('chdtr', (1, 0), 0.0),
+         ('chdtr', (0.7, cupy.inf), 1.0),
+         ('chdtr', (0.6, 3), 0.957890536704110),
+         ('chdtrc', (1, 0), 1.0),
+         ('chdtrc', (0.6, 3), 1 - 0.957890536704110),
+         ('chdtri', (1, 1), 0.0),
+         ('chdtri', (0.6, 1 - 0.957890536704110), 3),
          ]
     )
-    def test_scalar(self, function, args, expected, rtol):
-        self._test_scalar(function, args, expected, rtol)
+    def test_scalar(self, function, args, expected):
+        self._test_scalar(function, args, expected)
 
 
 @testing.gpu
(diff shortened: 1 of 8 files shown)


emcastillo:
Thanks @grlee77! looking awesome as always!

grlee77:
I opened https://github.com/scipy/scipy/issues/15596 upstream to discuss what seems a bit inconsistent in use of `DeprecationWarning` vs. `RuntimeWarnings` for some of the unsafe floating point to int casting.

emcastillo:
Seems that tests in windows fail, windows CI uses scipy 1.6, may it be related? 
for some strange reason, the stacktrace is broken and we get infinite recursion on it 😥

Also ROCm seems to have failures too

tests/cupyx_tests/scipy_tests/special_tests/test_statistics.py
```diff
@@ -141,3 +139,158 @@ def check_unary_lower_precision(self, name, xp, scp, dtype):
...
+    @testing.for_signed_dtypes(name='int_dtype')
+    @testing.numpy_cupy_allclose(atol=1e-5, rtol=1e-5, scipy_name='scp')
+    def test_binomdist_linspace(self, xp, scp, function, dtype, int_dtype):
+        import scipy.special  # NOQA
+
+        if xp.dtype(int_dtype) not in [xp.int32, xp.int64]:
+            if xp.dtype(dtype) != xp.float64:
+                # Skip cases deprecated in SciPy 1.7+ via this Cython code:
+                # https://github.com/scipy/scipy/blob/cdb9b034d46c7ba0cacf65a9b2848c5d49c286c4/scipy/special/_legacy.pxd#L39-L43  # NOQA
+                # It causes infinite recursion in numpy_cupy_allclose
```
emcastillo, line 245: This seems to be not solved in windows ci with scipy 1.6 :)

grlee77: I did also find a change in behavior for SciPy < 1.5, so I added a `with_requires('scipy>=1.5.0')` on this test case. Given that the failure occurred for SciPy 1.6.3 that should be unrelated, though.

I did fix a difference in the order of type casting cases in the ufunc in 18f1436, so maybe worth trying again? Doesn't seem likely that this would be Windows-specific though.

grlee77: The actual issue may be related to `long` being 32-bit instead of 64-bit on that platform? That might cause a different type cast to occur there than on the linux system I have been testing with locally.

grlee77: I just verified on a Windows laptop that `np.dtype(int).itemsize` is 4 and not 8, so I think character code 'l' as in the ufunc type codes is also 32-bit there which is probably the source of the difference. I have attempted a fix in 7a3652a, but couldn't test it locally as there is no CUDA-capable GPU in that laptop.

emcastillo: Thanks for the effort in reproducing this!!! 😃
I think in this case you can use `numpy.dtype('q')` that gives the same int64 type in linux and windows

grlee77: I look at the following to see what types SciPy has defined:
```Python
from scipy import special

print(special.bdtr.types)
```
which gives `['fff->f', 'dld->d', 'ddd->d']` which is where the `l` type is coming in. I assume that on Windows this means a 64-bit int will go to the `ddd->d` path instead `dld->d` since 'l' is 32-bit on that system?

emcastillo: I think so, but shouldn't scipy ensure that `dld` is also executed in windows?

grlee77:
> /test mini

Unfortunately this still failed. I updated it again. Can you try one more time? (As long as `dtype` is float64 none of the `int_dtypes` raise a DeprecationWarning on Windows in my local testing)

grlee77:
Okay, maybe I already triggered the restart by pasting the command above (wasn't sure if only maintainers could initiate the tests?)

emcastillo:
@grlee77 sorry for asking you this,

I have seen some issues in ROCm. Essentially

`TestTwoArgumentDistribution.test_linspace_broadcast` this test should be skipped for ROCm

`TestThreeArgumentDistributions.test_scalar` we should change the tolerances here.
```
    Mismatched elements: 1 / 1 (100%)
19:33:07 E       Max absolute difference: 1.11022302e-16
19:33:07 E       Max relative difference: 2.22044605e-16
19:33:07 E        x: array(0.5)
```

Besides that, once windows things are ok this should be good to go!

grlee77:
> TestTwoArgumentDistribution.test_linspace_broadcast this test should be skipped for ROCm

I have added the skip. Should we only skip a subset of the parameterized functions (which ones failed?). I am curious if it is just a tolerance issue or if there is something we should try to address.

> TestThreeArgumentDistributions.test_scalar we should change the tolerances here.

okay, I will just set atol and rtol to 1e-12 on all of these scalar tests

> Besides that, once windows things are ok this should be good to go!

I think that may finally be fixed now in the latest commit. I wrote the following script to verify no warning was raised on Windows or linux and print the dtypes that won't be skipped on each:

```Python
**import numpy as np
from scipy import special

for dtype in [np.float16, np.float32, np.float64]:
    for int_dtype in [np.int8, np.int16, np.int32, np.int64,
                      np.uint8, np.uint16, np.uint32, np.uint64]:
        safe_cast = np.result_type(int_dtype, 'l') == np.dtype('l')
        safe_cast &= np.result_type(int_dtype, dtype) == np.float64
        if not safe_cast:
            continue
        print(f"dtype={np.dtype(dtype).name}, int_dtype={np.dtype(int_dtype).name}")
        k = np.asarray([2, 3, 4], dtype=dtype)
        n = np.asarray([2, 3, 4], dtype=int_dtype)
        p = np.asarray([0.2, 0.4, 0.6], dtype=dtype)
        special.bdtr(k, n, p)
        special.bdtrc(k, n, p)
        special.bdtri(k, n, p)
```

on linux this gives:
```
dtype=float16, int_dtype=int32
dtype=float16, int_dtype=int64
dtype=float16, int_dtype=uint32
dtype=float32, int_dtype=int32
dtype=float32, int_dtype=int64
dtype=float32, int_dtype=uint32
dtype=float64, int_dtype=int8
dtype=float64, int_dtype=int16
dtype=float64, int_dtype=int32
dtype=float64, int_dtype=int64
dtype=float64, int_dtype=uint8
dtype=float64, int_dtype=uint16
dtype=float64, int_dtype=uint32
```

while on Windows it is a smaller set:
```
dtype=float16, int_dtype=int32
dtype=float32, int_dtype=int32
dtype=float64, int_dtype=int8
dtype=float64, int_dtype=int16
dtype=float64, int_dtype=int32
dtype=float64, int_dtype=uint8
dtype=float64, int_dtype=uint16
```

Last time around I had only enforced dtype=float64 which was not sufficient on windows to avoid the `ddd->d` code path that raises SystemError/DeprecationWarning on SciPy.

emcastillo:
Thanks @grlee77 hope it goes through this time!!

emcastillo:
hey @grlee77 since this is mostly ready I took the liberty of debugging rocm issues and push to your branch.
Thanks!!

emcastillo (approved):
LGTM

emcastillo:
Glad to see this merged! thanks a lot @grlee77 ☺️

Issue #6455, opened 2022-02-10
Support Optuna 3.0


kmaehashi:
Optuna 3.0 alpha release is out. There is a deprecation that affects our code:

```
E       FutureWarning: IntUniformDistribution has been deprecated in v3.0.0. This feature will be removed in v6.0.0. See https://github.com/optuna/optuna/releases/tag/v3.0.0. Use :class:`~optuna.distributions.IntDistribution` instead.
```

PR #6470, opened 2022-02-16
Ignore warnings from Optuna 3.0 pre-releases

kmaehashi:
Part of #6455

It seems that Optuna 3.0.0a1/a2 has an issue that emits FutureWarning when using `Trial.suggest_int`.
Let's ignore them until the issue is fixed in Optuna side.

https://github.com/optuna/optuna/releases/tag/v3.0.0-a1

> v3.0.0-a1 contains several changes for this project and you will temporarily see UserWarning when you call Trial.suggest_int and Trial.suggest_float. We apologize for the inconvenience and the warning will be removed from the next release.


kmaehashi:
Let me merge this as the head test passed.

PR #6308, opened 2022-01-12
added `cupy.format_float_positional`

amanchhaparia:
Hello,
I recently came across the **CUPY** and found it interesting. I am excited and want to contribute to the cupy. 

I have implement the `cupy.format_float_positional` as requested in #6078,
looking for a review and further guidance.

Thank you.

PR #6474, opened 2022-02-16
Add `cupy.format_float_scientific` API

pri1311:
Hello!

This PR adds a version of `format_float_scientific` as requested in #6078 . This is my first contribution here, I hope it's the expected format and documented.


cupy/_io/formatting.py
```diff
@@ -63,3 +63,14 @@ def format_float_positional(x, *args, **kwargs):
 
     """
     return numpy.format_float_positional(cupy.asnumpy(x), *args, **kwargs)
+
+
+def format_float_scientific(x, *args, **kwargs):
```
takagi, line 68: Would you make it explicitly have the same parameters as those of `numpy.format_float_scientific`? I guess you followed the current implementation of `cupy.format_float_positional`, but I think it also should be fixed to have explicit parameters as its numpy counterpart.

takagi:
Thanks for your contribution! It's almost fine except for one I commented above.

cupy/_io/formatting.py
kmaehashi: That's true in general, but in this case, I think it's ok to keep as-is.

When I reviewed `format_float_positional` in #6308 I asked to use `*args, **kwargs` because several options are only available in recent NumPy versions (e.g., `min_digits` added in NumPy 1.21.0), and we don't want to pay the cost to fill the gap between NumPy versions by ourself considering the importance of these "print" features.

takagi: Makes sense. Then let me kick the CI to merge this soon!

takagi (approved):

takagi:
@kmaehashi, would you merge this if Jenkins looks okay?

Issue #6491, opened 2022-02-22
Improve Contribution Guide


kmaehashi:
### Description

As new contributors are increasing, I think it's better to consider revising the [Contribution Guide](https://docs.cupy.dev/en/latest/contribution.html). There are several things to improve in my mind:

* Cover the basic design rules of CuPy
  * Provides the same signature as NumPy/SciPy API
  * Code must be GPU optimized (parallelized); in most cases not just copy-paste of NumPy code
  * Avoid host/device data transfer and synchronization as much as possible
* Shorten the length of the documentation
  * "Issue/PR Labels" and "How to File an Issue" can be removed as we now have issue templates.
  * "Development Cycle" describes the release cycle and most part is not related to contribution. Move it to the User Guide.
* Cover techniques when writing a test.
  * Use `shaped_random` for generating input
  * Cover combination of shape and type variations
  * Test must (in general) check equality of result with NumPy, using appropriate test helpers
* Merge https://github.com/cupy/cupy/wiki/Development-policies

PR #6498, opened 2022-02-22
Moved development cycle from contributing to user guidelines.

Onkar627:
Issue #6491


kmaehashi:
Thanks for the PR, but I think we also have to update the whole content as it's a bit outdated. I'd suggest taking another task until you get familiar with the dev cycle and contribution steps in our project 😄

Onkar627:
> Thanks for the PR, but I think we also have to update the whole content as it's a bit outdated. I'd suggest taking another task until you get familiar with the dev cycle and contribution steps in our project 😄

Yes sure😃

PR #6510, opened 2022-02-25
Skip appending `--compiler-bindir` if `cl.exe` is already on `PATH`

kmaehashi:
This avoids appending a `--compiler-bindir` if `cl.exe` is already on PATH.

The issue was discovered in conda-forge build:
https://github.com/conda-forge/cupy-feedstock/pull/165#issuecomment-1050536590

conda-forge build env uses nvcc-feedstock that comes with the [wrapper batch file](https://github.com/conda-forge/nvcc-feedstock/blob/master/recipe/windows/nvcc_windows.bat) that prepends `-ccbin "%CXX%"`, and this batch file is used as NVCC command. As CuPy can't tell whether NVCC command is a wrapper or a real executable, this PR just skips appending `--compiler-bindir` and let NVCC find `cl.exe` if it's already on PATH.

cc/ @leofang

### DIFF
--- a/install/cupy_builder/_command.py
+++ b/install/cupy_builder/_command.py
@@ -102,6 +102,8 @@ class custom_build_ext(setuptools.command.build_ext.build_ext):  # type: ignore[
             for src in ext.sources:
                 if not os.path.isfile(src):
                     raise RuntimeError(f'Fatal error: missing file: {src}')
+
+        print('Building extensions...')
         super().build_extensions()
 
     def build_extension(self, ext: setuptools.Extension) -> None:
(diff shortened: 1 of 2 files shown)


leofang:
Thanks for quickly coming up with a patch, @kmaehashi! I was thinking to try something different but this approach should be just fine. Let me try this patch in conda-forge shortly.

Question:
> As CuPy can't tell whether NVCC command is a wrapper or a real executable, this PR just skips appending --compiler-bindir and let NVCC find cl.exe if it's already on PATH.

Didn't CuPy use to honor the env var `NVCC`? Is it gone now?
https://github.com/conda-forge/cupy-feedstock/blob/6dbba68264306820f27aa04cf3fab9d161b216c0/recipe/meta.yaml#L30-L32

leofang:
Looks like it works!

emcastillo (approved):

PR #6513, opened 2022-02-26
CI: Implement Docker caching for CuPy release workflow

HarshCasper:
## Description

This PR implements Docker caching for CuPy release workflow.

Relevant Docs: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#local-cache

### Before caching ([Logs](https://github.com/cupy/cupy/runs/5343897816?check_suite_focus=true))

image

#### After caching ([Logs](https://github.com/cupy/cupy/runs/5343946966?check_suite_focus=true))

image

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,6 +20,16 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Cache Docker layers
+      uses: actions/cache@v2
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-buildx-${{ github.sha }}
+        restore-keys: ${{ runner.os }}-buildx-
+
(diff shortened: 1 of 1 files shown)


kmaehashi:
Hi @HarshCasper, thanks for the PR!
I'm wondering if caching is needed here because (1) this workflow only runs when making a release (i.e. once a month) and (2) the base CUDA Docker images are updated frequently and we would like to use the latest one every time when making a release, so it is unlikely to hit the cache.

kmaehashi:
Let me close this as I think caching is not needed for the above reasons. Thanks anyways! 😃

HarshCasper:
@kmaehashi No worries! I believe this would have been helpful in the case of PRs as well. I will find something else to work on. Thanks!

kmaehashi:
> I believe this would have been helpful in the case of PRs as well.

Actually, the workflow runs in PRs as well but only for PRs modifying `docker/**` and `.github/workflows/docker.yml`, which is almost not happening except for the release.

Issue #6459, opened 2022-02-10
Support ROCm 5.0


takagi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read [ROCm Release Notes](https://rocmdocs.amd.com/en/latest/Current_Release_Notes/Current-Release-Notes.html).
- [x] Add FlexCI configuration for the ROCm version. #6466 #6496
- Wheel distribution
  - [x] Update cupy-release-tools. https://github.com/cupy/cupy-release-tools/pull/237
  - [x] Add the package to [duplicate detection](https://github.com/cupy/cupy/blob/master/cupy/_environment.py). #6466
  - [x] Backport. #6496
- Documentation
  - [x] Add new wheel package to [Installation Guide](https://docs.cupy.dev/en/latest/install.html) and `README.md`.  Update requirements in the installation guide. #6530
  - [x] Backport. #6607
  - [x] Add new wheel package to the [website](https://cupy.dev). https://github.com/cupy/cupy.dev/pull/54
- [x] Implement or create an issue to support new features, if applicable.

takagi:
The half float representation issue which we were facing on ROCm 4.5 looks gone. With the change in #6466, I'm running CuPy's test locally.

takagi:
ROCm 5.0.1 is out.

takagi:
Looks cumsum doesn't work well and causes some of the CI errors including `test_dlpack.py`:
```
$ python -c 'import cupy; print(cupy.cumsum(cupy.array([0,0])))'
[                  0 1304722626565153334]
```

takagi:
The way discussed in #4366 goes wrong on ROCm 5.0?

PR #6524, opened 2022-03-03
Fix for supporting ROCm 5.0

takagi:
Rel #6459. Merge #6517 first.

A number of fixes for making CuPy work healthy with ROCm 5.0.

### DIFF
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -4,6 +4,7 @@ import unittest
 import numpy
 import pytest
 
+from cupy_backends.cuda.api import runtime
 from cupy_backends.cuda import stream as stream_module
 import cupy
 from cupy import _util
@@ -163,6 +164,8 @@ class TestNdarrayDeepCopy(unittest.TestCase):
 
 
 _test_copy_multi_device_with_stream_src = r'''
+#include <ctime>
+
 extern "C" __global__
 void wait_and_write(long long *x) {
   clock_t start = clock();
@@ -180,8 +183,7 @@ void wait_and_write(long long *x) {
 '''
 
 
-@testing.gpu
-class TestNdarrayCopy(unittest.TestCase):
+class TestNdarrayCopy:
 
     @testing.multi_gpu(2)
     @testing.for_orders('CFA')
@@ -197,11 +199,14 @@ class TestNdarrayCopy(unittest.TestCase):
     def test_copy_multi_device_non_contiguous_K(self):
         arr = _core.ndarray((20,))[::2]
         with cuda.Device(1):
-            with self.assertRaises(NotImplementedError):
+            with pytest.raises(NotImplementedError):
                 arr.copy('K')
 
     # See cupy/cupy#5004
     @testing.multi_gpu(2)
+    @pytest.mark.xfail(
+        runtime.is_hip,
+        reason='ROCm may work differently in async D2D copy with streams')
     def test_copy_multi_device_with_stream(self):
         # Kernel that takes long enough then finally writes values.
         kern = cupy.RawKernel(
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -4,6 +4,7 @@ import unittest
 import numpy
 import pytest
 
+from cupy_backends.cuda.api import driver
 from cupy_backends.cuda.api import runtime
 from cupy_backends.cuda import stream as stream_module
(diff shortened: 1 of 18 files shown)


takagi:
Now ready to review.

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -197,11 +199,14 @@ def test_copy_multi_device_non_contiguous(self, order):
...
         with cuda.Device(1):
-            with self.assertRaises(NotImplementedError):
+            with pytest.raises(NotImplementedError):
                 arr.copy('K')
 
     # See cupy/cupy#5004
     @testing.multi_gpu(2)
+    @pytest.mark.xfail(
+        runtime.is_hip,
+        reason='ROCm may work differently in async D2D copy with streams')
```
kmaehashi, line 209: Question: Is this new in ROCm 5.0?

tests/cupy_tests/cuda_tests/test_memory.py
```diff
@@ -79,7 +79,15 @@ def check(self, device_id):
         if self.specify_device_id:
             kwargs = {'device_id': device_id}
 
-        unowned_mem = memory.UnownedMemory(*args, **kwargs)
+        if cupy.cuda.runtime.is_hip and self.allocator is memory._malloc:
+            # In ROCm, it seems that `hipPointerGetAttributes()`, which is
```
kmaehashi, line 83: Question: is this new in ROCm 5.0?

kmaehashi (commented):
Great! LGTM except for a few questions.

tests/cupy_tests/core_tests/test_ndarray.py
takagi: It must be no. I get failed in ROCm 4.3 on my local (with different values compared to that of ROCm 5.0). How about the recent result on Jenkins ROCm tests?

tests/cupy_tests/cuda_tests/test_memory.py
takagi: This test also fails in ROCm 4.3 on my local.

tests/cupy_tests/core_tests/test_ndarray.py
kmaehashi: Ah, Jenkins is not failing because it only has one GPU so this test was not running...

kmaehashi (approved):

kmaehashi:
@takagi Could you take a look at CUDA failures? :pray:

takagi:
Let me kick the CI.

takagi:
@kmaehashi, can you see the situation of cupy.win.cuda116 that runs so long...?

kmaehashi:
Oops I overlooked a test failure in Jenkins... (can be worked on in a separate PR)


```
11:31:51 FAILED cupyx_tests/scipy_tests/sparse_tests/test_linalg.py::TestLOBPCG::test_maxit_None
...
11:31:51 = 1 failed, 88968 passed, 7328 skipped, 1005 deselected, 4336 xfailed, 160 xpassed, 1733 warnings in 3519.20s (0:58:39) =
```

takagi:
oh... I can't reproduce the failure on my local with ROCm 4.3/5.0 or CUDA 11.5. Can I have the test condition on Jenkins?

kmaehashi:
Maybe the problem is that the test env is still ROCm 5.0.0?
https://github.com/cupy/cupy/blob/master/.pfnci/linux/tests/rocm-5-0.Dockerfile

full log: [consoleText.txt](https://github.com/cupy/cupy/files/8376274/consoleText.txt)

takagi:
Seems the direct cause is that the solver's iteration has finished in fewer times than expected. After I look into how the test works, I've found that CuPy's implementation returns wierd eigenvalues `[9.3633120e+07 1.6104681e+00 8.8367283e-02 1.4951467e-03]` at least in the case of `test_maxit_None` as opposed to SciPy's `[-1.        -1.9999943 -3.000001  -4.000016 ]`. I'm going to see if it reproduces on ROCm 4.3.

takagi:
omg, it works well on ROCm 4.3 with the eigenvalues `[-1.        -2.0000021 -3.0000026 -3.9999971]`. Let me mark the test as XFAIL in ROCm 5.0+ for this release and post an issue.

takagi:
> Maybe the problem is that the test env is still ROCm 5.0.0?

I don't think the patch version 5.0.x is related as far as I read the release notes, but I'll also confirm it.

takagi:
5.0.0 and 5.0.2 make no difference.

PR #6528, opened 2022-03-03
Add cupyx.scipy.special loggamma, multigammaln and fast Hankel transforms

grlee77:
This PR adds a few things to `cupyx.scipy.special` and implements a couple of the missing `cupyx.scipy.fft` functions that required `loggamma` for their implementation. New things are:

- added complex-valued support to `cupyx.scipy.special.gamma`
- added new special functions: `rgamma`, `loggamma` and `multigammaln`.
- added forward and inverse Fast Hankel Transform (aka [FFTLog](https://jila.colorado.edu/~ajsh/FFTLog/)). These transforms were introduced in SciPy 1.7.

I did not implement `scipy.fft.fhtoffset` as it only takes scalar inputs and returns a scalar `offset` for use with `fht` or `ifht`. It calls `loggamma` internally only with scalar input so any GPU implementation is going to be slower due to higher overhead in that case. We could import the SciPy version into the namespace when SciPy is present or just copy the SciPy implementation here if desired. Let me know what is preferred.

There is no upstream uarray multimethod for `scipy.fft.fht` or `scipy.fft.ifht` so I need to get a PR merged there for SciPy 1.9 before we can run the backend tests for those two on CI. It does work in local testing with a development version of SciPy, though.

cc @IvanYashchuk


grlee77:
As far as I could tell, FFTLog seems to be used mainly in cosmology. I also see a different variant and a 2D version at https://github.com/xfangcosmo that uses (complex-valued) `gamma` rather than `loggamma`.

pinging @xfangcosmo in case a GPU-accelerated variant of these is of interest. It looks like the [2D-FFTLog code](https://github.com/xfangcosmo/2DFFTLog/blob/master/python/twobessel.py) could be adapted to use CuPy pretty easily using the complex-valued `gamma` from this PR along with `cupy.fft.rfft2` and `cupy.fft.irfft2`. Out of curiousity, what are the typical sizes of the transforms being performed in this application?

xfangcosmo:
> 

Hi Gregory, my application involves transforming a 1024x1024 symmetric matrix. Also in actual application, I don't use the simplest 2D-FFTLog routine, but the "bin-averaged" version.

grlee77:
I was able to reproduce the failure locally when testing with SciPy 1.4. The version skip here resolves it.

tests/cupyx_tests/scipy_tests/special_tests/test_beta.py
```diff
@@ -98,20 +103,20 @@ def test_arange(self, xp, scp, dtype, function):
...
+    #     import scipy.special  # NOQA
+
+    #     func = getattr(scp.special, function)
+    #     # TODO: Some choices of start/stop value can give mismatched location
+    #     #       of +inf or -inf values.
+    #     tmp = xp.linspace(-20, 21, 10, dtype=dtype)
+    #     a = tmp[:, xp.newaxis, xp.newaxis]
+    #     b = tmp[xp.newaxis, :, xp.newaxis]
+    #     x = xp.linspace(0, 1, 5, dtype=dtype)[xp.newaxis, xp.newaxis, :]
+    #     return func(a, b, x)
```
grlee77, line 119: Looks like the last run passed. Is the plan to uncomment this one again or restrict the HIP version here as well?

emcastillo: this one was unrelated, just a leftover of my test combining runs :D

emcastillo (approved):

PR #6538, opened 2022-03-09
CI: Trigger `push` event of FlexCI via GitHub Actions

kmaehashi:
We have been triggering FlexCI via `push` webhook to run tests against merged commit. However, it is becoming costly to maintain a webhook URL because we have to embed a comma-separated list of FlexCI projects to trigger in the URL, and the list is so long.
This PR lets GitHub Actions trigger FlexCI on push event in addition to `issue_comment` event (`/test`).

(note: after merging this PR we should remove webhook configuration)

### DIFF
--- a/.github/workflows/flexci.yml
+++ b/.github/workflows/flexci.yml
@@ -2,7 +2,7 @@ name: "FlexCI"
 
 on:
   push:
-    branches: ["master", "v*"]
+    branches: ["master", "v[0-9]+"]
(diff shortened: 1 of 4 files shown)


kmaehashi:
It now looks working as long as I tested in my repo: https://github.com/kmaehashi/cupy/runs/5548866360?check_suite_focus=true#step:5:1
(the error is because that my repo doesn't have tokens set)

PR #6545, opened 2022-03-14
Fix array creation shape

### LINKED ISSUE #5503: Inconsistent behavior between numpy and cupy when handling shapes in array creation
Thank you for using and contributing to CuPy!

If your issue is a **request for support in using CuPy**, please post it on [Stack Overflow](https://stackoverflow.com/questions/tagged/cupy) or Google Groups ([en](https://groups.google.com/forum/#!forum/cupy) / [ja](https://groups.google.com/forum/#!forum/cupy-ja)).
Developer team members are monitoring questions on these channels.

If it is a bug report, please include the following information:

When using numpty integer types as datatype for shape information, cupy behaves differently than numpy. For example using np.int32 in shape info causes cupy to fail whereas numpy accepts it. I think cupy could be more generous of what datatypes it accepts. Reproducing steps below.

System info:
OS                           : Linux-4.4.0-119-generic-x86_64-with-glibc2.23
Python Version               : 3.9.5
CuPy Version                 : 9.2.0
NumPy Version                : 1.20.2
SciPy Version                : 1.6.3
Cython Build Version         : 0.29.23
CUDA Root                    : /home/cuda/miniconda3/envs/connectomes
CUDA Build Version           : 11000
CUDA Driver Version          : 11040
CUDA Runtime Version         : 11000
cuFFT Version                : 10201
cuRAND Version               : 10201
cuSOLVER Version             : (10, 6, 0)
NVRTC Version                : (11, 0)
Thrust Version               : 100909
CUB Build Version            : 100909
NCCL Build Version           : 20909
NCCL Runtime Version         : 2803
cuTENSOR Version             : 10300
Device 0 Name                : Tesla V100-DGXS-16GB
Device 0 PCI Bus ID          : 0000:07:00.0
Device 1 Name                : Tesla V100-DGXS-16GB
Device 1 Compute Capability  : 70
Device 1 PCI Bus ID          : 0000:08:00.0
Device 2 Name                : Tesla V100-DGXS-16GB
Device 2 Compute Capability  : 70
Device 2 PCI Bus ID          : 0000:0E:00.0
Device 3 Name                : Tesla V100-DGXS-16GB
Device 3 Compute Capability  : 70
Device 3 PCI Bus ID          : 0000:0F:00.0

reproducer with output:

```>>> import numpy as np
>>> import cupy as cp
>>> np.zeros(np.int32(10))
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
>>> cp.zeros(np.int32(10))
  File "/home/cuda/miniconda3/envs/connectomes/lib/python3.9/site-packages/cupy/_creation/basic.py", line 209, in zeros
    a = cupy.ndarray(shape, dtype, order=order)
  File "cupy/_core/core.pyx", line 128, in cupy._core.core.ndarray.__init__
  File "cupy/_core/internal.pyx", line 62, in cupy._core.internal.get_size
ValueError: size should be None, collections.abc.Sequence, or int```

### PR DESCRIPTION
**Pull request by takagi**
Fix #5503.

This PR fixes array creation to be able to take values as NumPy ints and NumPy zero-dim arrays for `shape`.

### DIFF
--- a/cupy/_core/internal.pyx
+++ b/cupy/_core/internal.pyx
@@ -8,6 +8,7 @@ from libc.stdint cimport uint32_t
 import sys
 import warnings
 
+import operator
 import numpy
 
 from cupy._core.core cimport ndarray
@@ -53,10 +54,18 @@ cpdef inline tuple get_size(object size):
         )
         return ()
     if cpython.PySequence_Check(size):
-        return tuple(size)
-    if isinstance(size, int):
-        return size,
-    raise ValueError('size should be None, collections.abc.Sequence, or int')
+        try:
+            return tuple(size)
+        except TypeError:
+            # A numpy.ndarray unconditionally succeeds in PySequence_Check as
+            # it implements __getitem__, but zero-dim one is an unsized object
+            # and fails to make a tuple from it.
+            pass
+    try:
+        return operator.index(size),
+    except TypeError as e:
+        raise ValueError(
+            'size should be None, collections.abc.Sequence, or int') from e
 
 
 @cython.profile(False)
--- a/cupy/_core/internal.pyx
+++ b/cupy/_core/internal.pyx
@@ -8,7 +8,6 @@ from libc.stdint cimport uint32_t
 import sys
 import warnings
 
-import operator
 import numpy
 
 from cupy._core.core cimport ndarray
@@ -45,6 +44,18 @@ cpdef inline bint is_in(const vector.vector[Py_ssize_t]& args, Py_ssize_t x):
 
 
 @cython.profile(False)
+cpdef inline void _check_not_bool(object x) except *:
+    if isinstance(x, (bool, numpy.bool_)):
+        raise TypeError('an integer is required')
+    if (
+        isinstance(x, numpy.ndarray)
+        and x.shape == ()
+        and x.dtype == numpy.bool_
+    ):
+        raise TypeError('an integer is required')
+
+
+@cython.profile(False)
 cpdef inline tuple get_size(object size):
     if size is None:
         warnings.warn(
@@ -54,18 +65,19 @@ cpdef inline tuple get_size(object size):
         )
         return ()
     if cpython.PySequence_Check(size):
+        # A numpy.ndarray unconditionally succeeds in PySequence_Check as
+        # it implements __getitem__, but zero-dim one is an unsized object
+        # and fails to make a tuple from it.
         try:
-            return tuple(size)
+            ret = tuple(size)
         except TypeError:
-            # A numpy.ndarray unconditionally succeeds in PySequence_Check as
-            # it implements __getitem__, but zero-dim one is an unsized object
-            # and fails to make a tuple from it.
-            pass
-    try:
-        return operator.index(size),
-    except TypeError as e:
-        raise ValueError(
-            'size should be None, collections.abc.Sequence, or int') from e
+            _check_not_bool(size)
+            return size,
+        for x in ret:
+            _check_not_bool(x)
+        return ret
+    _check_not_bool(size)
+    return size,
 
 
 @cython.profile(False)
--- a/tests/cupy_tests/core_tests/test_internal.py
+++ b/tests/cupy_tests/core_tests/test_internal.py
@@ -1,6 +1,7 @@
 import math
 import unittest
 
+import numpy
 import pytest
 
 from cupy._core import internal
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: cupy/_core/internal.pyx [hunk 1]
```diff
@@ -53,10 +54,18 @@ cpdef inline tuple get_size(object size):
...
+            return tuple(size)
+        except TypeError:
+            # A numpy.ndarray unconditionally succeeds in PySequence_Check as
+            # it implements __getitem__, but zero-dim one is an unsized object
+            # and fails to make a tuple from it.
+            pass
+    try:
+        return operator.index(size),
+    except TypeError as e:
+        raise ValueError(
```
takagi, line 67: Should we also fix to just propagate `TypeError` instead of raising a `ValueError`?
```Python
>>> import numpy
>>> numpy.zeros('a')
TypeError: 'str' object cannot be interpreted as an integer
```

asi1024: I agree to propagate `TypeError` which `operator.index` raises.

asi1024:
@takagi I am wondering if `internal.get_size` should check the input type.
If it should check, we should check all items of `size` when `cpython.PySequence_Check(size)` is `True`. If not, we can return `size,` immediately without raising an error when `cpython.PySequence_Check(size)` is `False`

cupy/_core/internal.pyx
```diff
@@ -53,10 +54,18 @@ cpdef inline tuple get_size(object size):
...
-    raise ValueError('size should be None, collections.abc.Sequence, or int')
+        try:
+            return tuple(size)
+        except TypeError:
+            # A numpy.ndarray unconditionally succeeds in PySequence_Check as
+            # it implements __getitem__, but zero-dim one is an unsized object
+            # and fails to make a tuple from it.
+            pass
+    try:
+        return operator.index(size),
```
toslunar, line 65: I suppose CuPy must reject `bool` and `np.bool_`.  Are they rejected somewhere else?
https://github.com/numpy/numpy/blob/2e2c4c20c745781719006e6110b6f1236c4e51f9/numpy/core/src/multiarray/conversion_utils.c#L918-L926

cupy/_core/internal.pyx
```diff
@@ -53,10 +54,18 @@ cpdef inline tuple get_size(object size):
         )
         return ()
     if cpython.PySequence_Check(size):
-        return tuple(size)
-    if isinstance(size, int):
-        return size,
-    raise ValueError('size should be None, collections.abc.Sequence, or int')
+        try:
+            return tuple(size)
```
toslunar, line 58: Should we test each element of `size`?  The fix might be out of scope of the PR, but it might help refactoring.

toslunar (commented):
In NumPy,
- `PyArray_IntpConverter` parses `shape` https://github.com/numpy/numpy/blob/2e2c4c20c745781719006e6110b6f1236c4e51f9/numpy/core/src/multiarray/multiarraymodule.c#L1945 and
- `PyArray_PyIntAsIntp` parses each value in shape.

takagi:
Thanks! I traced the code, and as far as the way `internal.get_size` is currently used, its result is to be pushed or converted to a `vector.vector[Py_ssize_t]`, so Cython is responsible to interpret its elements as integers even if `get_size` doesn't check the type of its argument. I've removed checking types from this PR except for `bool` and `np.bool_`, which are not rejected elsewhere.

```
$ find cupy -name '*.py' -or -name '*.pyx' | xargs grep get_size
cupy/_core/internal.pyx:cpdef inline tuple get_size(object size):
cupy/_core/__init__.py:from cupy._core.internal import get_size  # NOQA
cupy/_core/core.pyx:        cdef tuple s = internal.get_size(shape)
cupy/random/_generator.py:        size = _core.get_size(size)
cupy/fft/_cache.pyx:    cpdef Py_ssize_t get_size(self):
cupy/fft/_cache.pyx:    return cache.get_size()
```

cupy/_core/internal.pyx
takagi: Makes sense. Seems they are not rejected anywhere.

tests/cupy_tests/core_tests/test_internal.py
```diff
@@ -49,9 +50,27 @@ def test_tuple(self):
...
+            internal.get_size(value)
+        with pytest.raises(TypeError):
+            internal.get_size((value, value))
+
     def test_float(self):
-        with pytest.raises(ValueError):
-            internal.get_size(1.0)
+        # `internal.get_size` is not responsible to interpret values as
+        # integers.
+        assert internal.get_size(1.0) == (1.0,)
```
asi1024, line 73: If `internal.get_size` is expected not to raise errors, we should add a test checks if `cupy.ndarray(shape=[1.0, 2.0])` raises TypeError?

takagi: Sure!

kmaehashi:
Test seems passed.

asi1024 (approved):

Issue #4897, opened 2021-03-17
Use CUB by default


emcastillo:
It feels a bit weird that since cub is bundled in cupy and performs quite fast in the majority of cases, it is not the default choice for at least routines such as sum or mission-critical ones.

Are there any issues that prevent us from using CUB by default, at least in routines?

I was discussing offline with @kmaehashi and maybe reproducibility of results with cub is not ensured, but I am not familiar with cub enough to assert on this. Maybe @leofang or @asi1024 can shed some light here.

Thanks!

leofang:
I don't remember in which issue/PR we decided to not make it default, but IIRC the concern was not about CUB's reproducibility, but about users code might not be perfectly reproducible after they upgrade CuPy if CuPy update the default backend.

AFAIK, CUB claims deterministic outcomes, so I wouldn't worry about it.

leofang:
But, be aware of https://github.com/cupy/cupy/issues/3822#issuecomment-782607637, and potential issues linked in #3723.

emcastillo:
I think that we should select a series of routines where CUB is enabled by default.
Others could be left as experimental.

leofang:
That's what I thought too. CUB reduction is definitely fine. I think it's probably safe to default everything except CUB SpMV.

leofang:
> That's what I thought too. CUB reduction is definitely fine. 

btw if we default to this, we fix #2750 for free, as CUB reduction is not strided. Though it's not a big deal. 

> I think that we should select a series of routines where CUB is enabled by default.
> Others could be left as experimental.

We'll need to document whatever we decide (#3692).

kmaehashi:
I think `atomicAdd` is used in CUB ops which does not have reproducibility.

kmaehashi:
Discussed internally:

* We are using `atomicAdd` in several locations (not limited to CUB) and not guaranteeing reproducibility overall. Considering the merit (esp performance improvements in reduction) it is worth doing enabling CUB by default.
* However as this will change the behavior of many reduction operations, we should explicitly add documentation for users who needs reproducibility (i.e., `export CUPY_ACCELERATORS=""` to disable CUB)

duncanriach:
> I think `atomicAdd` is used in CUB ops which does not have reproducibility.

[...]

PR #6549, opened 2022-03-15
Use CUB by default

asi1024:
https://github.com/cupy/cupy/issues/4897#issuecomment-1067850147


kmaehashi:
/test mini

* Will you do docs in separate PR or in this PR?
* We will need to add test variant for `CUPY_ACCELERATOR=""` case.

cupy/_core/_accelerator.pyx
```diff
@@ -43,7 +43,7 @@ def get_routine_accelerators():
 
 
 cdef _set_default_accelerators():
-    cdef str b, accelerator_names = os.getenv('CUPY_ACCELERATORS', '')
+    cdef str b, accelerator_names = os.getenv('CUPY_ACCELERATORS', 'cub')
```
asi1024, line 46:
```suggestion
    cdef str b, accelerator_names = os.getenv('CUPY_ACCELERATORS', '' if runtime.is_hip else 'cub')
```

asi1024:
@kmaehashi Updated. PTAL!

docs/source/reference/environment.rst
```diff
@@ -71,7 +71,7 @@ Here are the environment variables that CuPy uses at runtime.
 
 .. envvar:: CUPY_ACCELERATORS
 
-  Default: ``""`` (no accelerators)
+  Default: ``"cub"`` (In ROCm HIP environment, the default value is ``""``. i.e., no accelerator is used.)
```
kmaehashi, line 74:
```suggestion
  Default: ``"cub"`` (In ROCm HIP environment, the default value is ``""``. i.e., no accelerators are used.)
```

docs/source/user_guide/performance.rst
```diff
@@ -106,6 +106,9 @@ contiguous axes.
 In any case, we recommend to perform some benchmarks to determine whether CUB/cuTENSOR offers
 better performance or not.
 
+.. note::
+   CuPy v11 and above uses CUB by default. If you sets ``CUPY_ACCELERATORS=""`` it will be disabled.
```
kmaehashi, line 110:
```suggestion
   CuPy v11 and above uses CUB by default. To turn it off, you need to explicitly specify the environment variable ``CUPY_ACCELERATORS=""``.
```

kmaehashi (commented):
Thanks, made tiny docs suggestions. Could you fix doctest failure?

kmaehashi:
> We will need to add test variant for CUPY_ACCELERATOR="" case.

Did it in: https://github.com/cupy/cupy/pull/6580

kmaehashi:
> Could you fix doctest failure? https://ci.preferred.jp/r/job/97008

* Add that this difference does not apply when CUB is enabled (CuPy v11+)
  https://docs.cupy.dev/en/latest/user_guide/difference.html#nan-not-a-number-handling
* Use `env:CUPY_ACCELERATORS: ""` instead of `env:CUPY_ACCELERATORS: null` so that doctests run with CUB disabled: https://github.com/cupy/cupy/blob/42718a96a1de4c9d1007624add585eda7239350f/.pfnci/matrix.yaml#L339-L362

kmaehashi (dismissed):

kmaehashi:
Ah wait, this is not done yet 😅 

> Add that this difference does not apply when CUB is enabled (CuPy v11+)
> https://docs.cupy.dev/en/latest/user_guide/difference.html#nan-not-a-number-handling

leofang (approved):
LGTM!

docs/source/user_guide/difference.rst
```diff
@@ -178,3 +178,4 @@ counterparts:
 
 The reason is that internally the reduction is performed in a strided fashion, thus it does not ensure a proper
 comparison order and cannot follow NumPy's rule to always propagate the first-encountered NaN.
+Note that this difference does not apply when CUB is enabled (CuPy v11+)
```
kmaehashi, line 181:
```suggestion
Note that this difference does not apply when CUB is enabled (which is the default for CuPy v11 or later.)
```

PR #6555, opened 2022-03-16
CI: Fix event name in dispatcher

kmaehashi:


kmaehashi:
cc/ @emcastillo 
Let me self merge to test now.

PR #6570, opened 2022-03-20
eigsh: Prefer gemv over gemm

peterbell10:
Instead of doing a `gemm` where one of the matrices is a column vector, we can just call `gemv`. This doesn't matter too much on CUDA but for ROCm this performs significantly better.


IvanYashchuk:
For reference, here's a bug report reported to rocBLAS: https://github.com/ROCmSoftwarePlatform/rocBLAS/issues/1238.

asi1024 (approved):

PR #6582, opened 2022-03-23
add `cupy.setxor1d` api

pri1311:
Linked Issue: #6078 

Hey! I have implememeted numpy.setxor1d but seems like when you provide a multi-dim array with assume_unique as True, numpy raises an error, since their code concatenates array as - np.concatenate((ar1,ar2)) and not np.concatenate((ar1,ar2), axis=None).
However, when assume_unique is false, the array is flattened as it uses the numpy.unique function which returns a flattened array. Do we replicate the same behaviour in CuPy or allow multi-dim arrays even when assume_unique is set to true?

Attaching a screenshot below to explain the above:
159218884-0c0ac806-b821-477b-b68a-bec7ef48077e

Open discussion in numpy - https://github.com/numpy/numpy/issues/14670
Currently, the code I have written does not throw errors for multi-dim arrays, regardless of where `assume_unique` is True or False. However, tests currently don't cover that, since I can't compare it to numpy results. Should I add some manual-comparison tests? Please let me know your opinion.

### DIFF
--- a/cupy/_logic/truth.py
+++ b/cupy/_logic/truth.py
@@ -255,6 +255,44 @@ def setdiff1d(ar1, ar2, assume_unique=False):
     return ar1[in1d(ar1, ar2, assume_unique=True, invert=True)]
 
 
+def setxor1d(ar1, ar2, assume_unique=False):
+    """Find the set exclusive-or of two arrays.
+
+    Parameters
+    ----------
+    ar1, ar2 : cupy.ndarray
+        Input arrays. They are flattend if they are not already 1-D.
+    assume_unique : bool
+        By default, False, i.e. input arrays are not unique.
+        If True, input arrays are assumed to be unique. This can
+        speed up the calculation.
+
+    Returns
+    -------
+    setxor1d : cupy.ndarray
+        Return the sorted, unique values that are in only one
+        (not both) of the input arrays.
+
+    See Also
+    --------
+    numpy.setxor1d
+
+    """
+    if not assume_unique:
+        ar1 = cupy.unique(ar1)
+        ar2 = cupy.unique(ar2)
+
+    aux = cupy.concatenate((ar1, ar2), axis=None)
+    if aux.size == 0:
+        return aux
+
+    aux.sort()
+    flag = aux
+    flag = cupy.concatenate((cupy.asanyarray([True]), cupy.asanyarray(
+        aux[1:] != aux[:-1]), cupy.asanyarray([True])), axis=None)
+    return aux[flag[1:] & flag[:-1]]
+
+
 def union1d(arr1, arr2):
     """Find the union of two arrays.
 
--- a/cupy/_logic/truth.py
+++ b/cupy/_logic/truth.py
@@ -305,7 +305,8 @@ def setxor1d(ar1, ar2, assume_unique=False):
         'setxorkernel'
     )
 
-    return aux[setxorkernel(aux, aux.size, cupy.zeros(aux.size, dtype=cupy.bool8))]
+    return aux[setxorkernel(aux, aux.size,
+                            cupy.zeros(aux.size, dtype=cupy.bool8))]
 
 
 def union1d(arr1, arr2):
--- a/cupy/_logic/truth.py
+++ b/cupy/_logic/truth.py
@@ -4,7 +4,7 @@ from cupy._core import _fusion_thread_local
 from cupy import _util
 
 
-setxorkernel = cupy._core.ElementwiseKernel(
+_setxorkernel = cupy._core.ElementwiseKernel(
     'raw T X, int64 len',
     'bool z',
     'z = (i == 0 || X[i] != X[i-1]) && (i == len - 1 || X[i] != X[i+1])',
--- a/cupy/_logic/truth.py
+++ b/cupy/_logic/truth.py
@@ -287,10 +287,25 @@ def setxor1d(ar1, ar2, assume_unique=False):
         return aux
 
     aux.sort()
-    flag = aux
-    flag = cupy.concatenate((cupy.asanyarray([True]), cupy.asanyarray(
-        aux[1:] != aux[:-1]), cupy.asanyarray([True])), axis=None)
-    return aux[flag[1:] & flag[:-1]]
+
+    setxorkernel = cupy.ElementwiseKernel(
+        'raw T X, int32 len',
+        'bool z',
+        '''
+            if (i == 0) {
+                z = (X[0] != X[1]);
+            }
+            else if (i == len - 1) {
+                z = (X[i] != X[i-1]);
+            }
+            else {
+                z = X[i] != X[i-1] && X[i] != X[i+1];
+            }
+        ''',
+        'setxorkernel'
+    )
+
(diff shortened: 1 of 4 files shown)


emcastillo:
We align to the latest supported numpy version in the master branch. If NumPy errors, CuPy should error too unless there is a good reason that prevents this (device synchronization, checks using divergent threads, etc). In such case it should be clearly stated in the docstring.

pri1311:
> We align to the latest supported numpy version in the master branch. If NumPy errors, CuPy should error too

Alright, thanks. Will make the required changes.

asi1024:
An algorithm using `in1d` may be faster. Could you compare the performance?

asi1024:
@pri1311 Could you try `cupyx.profiler.benchmark`?

cupy/_logic/truth.py
```diff
@@ -255,6 +255,44 @@ def setdiff1d(ar1, ar2, assume_unique=False):
...
+
+    aux = cupy.concatenate((ar1, ar2), axis=None)
+    if aux.size == 0:
+        return aux
+
+    aux.sort()
+    flag = aux
+    flag = cupy.concatenate((cupy.asanyarray([True]), cupy.asanyarray(
+        aux[1:] != aux[:-1]), cupy.asanyarray([True])), axis=None)
+    return aux[flag[1:] & flag[:-1]]
```
asi1024, line 293: By using `ElementwiseKernel`, we can compute `flag[1:] & flag[:-1]` without allocating additional memories for temporary values.

pri1311:
> @pri1311 Could you try `cupyx.profiler.benchmark`?

Aah yes, on it.

pri1311:
image
These are the current stats, for array sizes
x: (3,2) y: (2,5)
x: (30,20) y: (20,50)
x: (300,200) y: (200,500)

cupy/_logic/truth.py
asi1024: @pri1311 How about this comment?

pri1311: Hey @asi1024! Sorry for the delay, was in between my exams, will get this done asap.

pri1311:
I replaced the following two lines
```
flag = cupy.concatenate((cupy.asanyarray([True]), cupy.asanyarray(aux[1:] != aux[:-1]), cupy.asanyarray([True])))
return aux[flag[1:] & flag[:-1]]
```
with
```
flag = cupy.concatenate((cupy.asanyarray([True]), boolcompare(aux[1:], aux[:-1]), cupy.asanyarray([True])))
return aux[xorkernel(flag[1:],flag[:-1])]
```

```
boolcompare = cupy.ElementwiseKernel(
    'T x, T y',
    'bool z',
    'z = x != y',
    'boolcompare' 
)
```
and this is the change in performance. 
image

I'll try to make further changes so as to avoid using the variable flag altogether by tonight. But would like to know if this is better than before or I could've done better?

Array sizes used for comparison:
```
x = cupy.random.rand(300,200)
y = cupy.random.rand(200,500)
```
Apologies for the delay.
Thank you!

pri1311:
Yes working on that

pri1311:
```
def setxor1dusingkernel(ar1, ar2, assume_unique=False):
    if not assume_unique:
        ar1 = cupy.unique(ar1)
        ar2 = cupy.unique(ar2)

    aux = cupy.concatenate((ar1, ar2), axis=None)
    if aux.size == 0:
        return aux

    aux.sort()

    setxorkernel = cupy.ElementwiseKernel(
        'raw T X, int32 len',
        'T z',
        '''
            if (i == 0) {
                z = (X[0] != X[1]);
            }
            else if (i == len - 1) {
                z = (X[i] != X[i-1]); 
            }
            else {
                z = X[i] != X[i-1] && X[i] != X[i+1];
            }
        ''',
        'setxorkernel' 
    )
    return setxorkernel(aux, aux.size, aux)
```
How about this?
Results:::
image

For array sizes
```
x = cupy.random.rand(300,200)
y = cupy.random.rand(200,500)
```
```
x = cupy.random.rand(3000,2000)
y = cupy.random.rand(2000,5000)
```

pri1311:
Hey, a few tests are failing (and I understand why) let me just make that correction, and will commit the changes. For some reason, the bug didn't show up when I was manually running the function from a separate file.

pri1311:
Hey @asi1024, apologies for the delay, had my final exams going on. Have added the final changes from my side. Can you review them once?

cupy/_logic/truth.py
```diff
@@ -255,6 +255,60 @@ def setdiff1d(ar1, ar2, assume_unique=False):
...
+        ar2 = cupy.unique(ar2)
+
+    aux = cupy.concatenate((ar1, ar2), axis=None)
+    if aux.size == 0:
+        return aux
+
+    aux.sort()
+
+    setxorkernel = cupy.ElementwiseKernel(
+        'raw T X, int32 len',
```
asi1024, line 292:
```suggestion
        'raw T X, int64 len',
```

cupy/_logic/truth.py
```diff
@@ -255,6 +255,60 @@ def setdiff1d(ar1, ar2, assume_unique=False):
...
+        '''
+            if (i == 0) {
+                z = (X[0] != X[1]);
+            }
+            else if (i == len - 1) {
+                z = (X[i] != X[i-1]);
+            }
+            else {
+                z = X[i] != X[i-1] && X[i] != X[i+1];
+            }
```
asi1024, line 303: How about simplifying as follows?
```suggestion
        z = (i == 0 || X[i] != X[i-1]) && (i == len - 1 || X[i] != X[i+1])
```

cupy/_logic/truth.py
```diff
@@ -255,6 +255,60 @@ def setdiff1d(ar1, ar2, assume_unique=False):
...
+        ''',
+        'setxorkernel'
+    )
+
+    return aux[setxorkernel(aux, aux.size,
+                            cupy.zeros(aux.size, dtype=cupy.bool8))]
```
asi1024, line 309:
```suggestion
                            cupy.zeros(aux.size, dtype=cupy.bool_))]
```

cupy/_logic/truth.py
```diff
@@ -255,6 +255,60 @@ def setdiff1d(ar1, ar2, assume_unique=False):
...
+        ar1 = cupy.unique(ar1)
+        ar2 = cupy.unique(ar2)
+
+    aux = cupy.concatenate((ar1, ar2), axis=None)
+    if aux.size == 0:
+        return aux
+
+    aux.sort()
+
+    setxorkernel = cupy.ElementwiseKernel(
```
asi1024, line 291: Could you define this kernel outside the scope of this function?

cupy/_logic/truth.py
```diff
@@ -4,6 +4,14 @@
 from cupy import _util
 
 
+setxorkernel = cupy._core.ElementwiseKernel(
```
asi1024, line 7: Could you prefix the kernel name with an underscore?

```suggestion
_setxorkernel = cupy._core.ElementwiseKernel(
```

asi1024 (approved):

asi1024:
LGTM! Thank you for your PR!

Issue #6577, opened 2022-03-23
`cupy.full` makes casting in `safe` rule


takagi:
### Description

Rel #6181.

As fixed in #6481, both `numpy.fill` and `cupy.fill` make casting in `safe` rule. However, `numpy.full` make casting in `unsafe` rule unlike `numpy.fill` does, and #6481 brought a bug that `cupy.full` works in `safe` rule as it is currently implemented using `cupy.fill`.

### To Reproduce

```py
>>> import numpy as np
>>> import cupy as cp
>>> np.full((), np.array(1, dtype=np.float64), dtype=np.int64)
array(1)
>>> cp.full((), cp.array(1, dtype=cp.float64), dtype=cp.int64)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ext-mtakagi/cupy.cuda-11.6/cupy/_creation/basic.py", line 276, in full
    a.fill(fill_value)
  File "cupy/_core/core.pyx", line 665, in cupy._core.core.ndarray.fill
    cpdef fill(self, value):
  File "cupy/_core/core.pyx", line 679, in cupy._core.core.ndarray.fill
    raise TypeError(
TypeError: Cannot cast scalar from dtype('float64') to dtype('int64') according to the rule 'safe'
```

### Installation

Built from GitHub source

### Environment

```
# Paste the output here
```

### Additional Information

`numpy.fill` and `cupy.fill` make casting in `safe` rule:
```py
>>> import numpy as np
>>> import cupy as cp
>>> a = np.array(0, dtype=np.int64)
>>> a.fill(np.array(1, dtype=np.float64))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Cannot cast scalar from dtype('float64') to dtype('int64') according to the rule 'safe'
>>> a = cp.array(0, dtype=cp.int64)
>>> a = a.fill(cp.array(1, dtype=cp.float64))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy/_core/core.pyx", line 665, in cupy._core.core.ndarray.fill
    cpdef fill(self, value):
  File "cupy/_core/core.pyx", line 679, in cupy._core.core.ndarray.fill
    raise TypeError(
TypeError: Cannot cast scalar from dtype('float64') to dtype('int64') according to the rule 'safe'
```

PR #6584, opened 2022-03-24
Fix `cupy.copyto` to take NumPy array scalars

takagi:
Rel #6577.

This PR fixes `cupy.copyto` to be able to take NumPy array scalars as its `src` data. I want to change this in advance to fixing #6577 in order to implement `cupy.full` using `cupy.copyto` as `numpy.full` does.

### DIFF
--- a/cupy/_manipulation/basic.py
+++ b/cupy/_manipulation/basic.py
@@ -37,7 +37,7 @@ def copyto(dst, src, casting='same_kind', where=None):
         src_dtype = numpy.dtype(type(src))
         can_cast = numpy.can_cast(src, dst.dtype, casting)
     elif isinstance(src, numpy.ndarray):
-        if src.shape != ():
+        if src.size != 1:
             raise ValueError(
                 'non-scalar numpy.ndarray cannot be used for copyto')
         src_dtype = src.dtype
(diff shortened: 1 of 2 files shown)


cupy/_manipulation/basic.py
```diff
@@ -35,6 +36,14 @@ def copyto(dst, src, casting='same_kind', where=None):
     if src_is_python_scalar:
         src_dtype = numpy.dtype(type(src))
         can_cast = numpy.can_cast(src, dst.dtype, casting)
+    elif isinstance(src, numpy.ndarray):
+        if src.shape != ():
```
asi1024, line 40: ndarray.item works when src.size == 1?

takagi: Yes, passing no arguments to ndrray.item is for that purpose. (Or you say another thing?)

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.item.html
> none: in this case, the method only works for arrays with one element (a.size == 1), which element is copied into a standard Python scalar object and returned.

asi1024: How about changing this condition to `src.size == 1`? We can support also `(1,)` shape arrays.

takagi: Sure!

asi1024 (approved):

asi1024:
LGTM.

PR #6588, opened 2022-03-25
JIT: CPU simulator

asi1024:
TODO:
- [ ] builtin functions
- [ ] device function call
- [ ] where inputs share the same memory
- [ ] tests

```py
import numpy
import cupy
from cupyx import jit


@jit.rawkernel()
def f(x, y):
    tid = jit.threadIdx.x + jit.blockDim.x * jit.blockIdx.x
    y[tid] = x[tid]


x = numpy.random.randint(100, size=30)
y = numpy.random.randint(100, size=30)
f.simulate((5,), (6,), (x, y))
assert bool((x == y).all())
```

### DIFF
--- a/cupyx/jit/_simulate.py
+++ b/cupyx/jit/_simulate.py
@@ -1,0 +1,85 @@
+import concurrent.futures
+import itertools
+import threading
+
+import numpy
+
+import cupy
+
+
+_thread_local = threading.local()
+_lock = threading.Lock()
+
+
+class DataRaceError(RuntimeError):
+    pass
+
+
+class ThreadDivergenceError(RuntimeError):
+    pass
+
+
+class _Global:
+
+    def __init__(self, array, read, write):
+        self._array = array
+        self._read = read
+        self._write = write
+
+    def __repr__(self, array):
+        return repr(array)
+
+    def __getitem__(self, item):
+        with _lock:
+            out = self._array[item]
+            if out.size == 1:
+                self._read[item] = True
+                if self._write[item]:
+                    raise DataRaceError('read/write data race.')
+                return out
+            return _Global(out, self._read[item], self._write[item])
+
+    def __setitem__(self, item, value):
+        with _lock:
+            if self._array[item].size != 1:
+                raise TypeError('only size-1 arrays can be set.')
+            if self._read[item]:
+                raise DataRaceError('read/write data race.')
+            if self._write[item]:
+                raise DataRaceError('write/write data race.')
+            self._array[item] = True
+            self._array[item] = value
+
+
+def _run_func(func, args, grid_dim, block_idx, block_dim, thread_idx):
+    _thread_local.gridDim = grid_dim
+    _thread_local.blockIdx = block_idx
+    _thread_local.blockDim = block_dim
+    _thread_local.threadIdx = thread_idx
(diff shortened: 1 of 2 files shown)


kmaehashi:
Discussed offline: Launching host threads of the same number as CUDA threads could be a problematic. How about running them sequentially and use yield when syncthread is needed?

cupyx/jit/_simulate.py
```diff
@@ -0,0 +1,85 @@
...
+def simulate(func, args, grid_dim, block_dim):
+    new_args = []
+    for x in args:
+        if isinstance(x, (numpy.ndarray, cupy.ndarray)):
+            read = numpy.zeros(x.shape, dtype=numpy.bool_)
+            write = numpy.zeros(x.shape, dtype=numpy.bool_)
+            x = _Global(x, read, write)
+        new_args.append(x)
+    n_workers = numpy.prod(grid_dim) * numpy.prod(block_dim)
+    with concurrent.futures.ThreadPoolExecutor(n_workers) as executor:
```
emcastillo, line 75: GIL is going to make this sequential anyway, so how about just looping and not using threads or futures at all?

Also, utilities for executing a single thread, or a single block to check indexes for arrays access could be handy here.

asi1024:
> Discussed offline: Launching host threads of the same number as CUDA threads could be a problematic. How about running them sequentially and use yield when syncthread is needed?

@kmaehashi @emcastillo 
I've thought about it after the discussion, and it seems that this idea will require changing the syntax to `yield from jit.syncthreads()`.
Now, I want to change the syntax to simply write `yield "<label>"` as a syncthreads, and deprecate `jit.synchtreads`. How do you think about it?

takagi:
Is it too much to implement another naive compiler that transpiles CuPy JIT to Python and evaluate its generated result with `exec` so that we can have arbitrary semantics on `jit.syncthreads()`?

kmaehashi:
Yeah @emcastillo and I were thinking of the similar idea, e.g., parse the user function and rewrite `syncthreads` etc to `await` or `yield` and execute it.

emcastillo:
The thing is like we have to take into account block & grid sync (cooperative groups?).
Probably mem fences are Ok since we don't have real parallel execution but the barriers will force us to have live "threads" holding their own local contexts.
We discussed this and maybe `asyncio` is the cleanest way to go. Barriers are not directly available but they can be easily implemented.

asi1024:
How about simulating on AST? For example, when using warp shuffle operations, we want to synchronize the threads for each operation, but I think that multi-threading or yield cannot deal with it well.

emcastillo:
might be a good idea

kmaehashi:
Offline discussion:
We would like to avoid both:
* launching thousands of host threads
* reinventing an interpreter that executes AST
So rewriting the target function with `asyncio` sounds the way to go. This should be doable by  by rewriting the target function `def` to `async def`, and rewrite `syncthreads` / ooperative groups / etc. with `await from`, all via AST.

asi1024:
@kmaehashi I've considered how to implement it, but I think it would be difficult to support a CPU simulator without reinventing an interpreter that executes AST while ensuring perfect synchronization for identical warps.
Closing this PR until we find a suitable implementation policy.

PR #6591, opened 2022-03-28
Fix mempool tests

### LINKED ISSUE #5349: MemoryAsyncPool test issues (tracker issue)
* #5308
* https://github.com/cupy/cupy/pull/5350

leofang:
Close #5349. ~~Blocked by #6590.~~

Although it is still unclear why the errors are not reproducible when running `test_memory.py` alone, the fix in this PR is necessary as the same mempool is reused across tests.


leofang:
7 failures with commit 89bd548:
https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/97860/log.txt

leofang:
2 failures with 1481550:
https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/98038/log.txt

leofang:
(Please merge after #6590.)

leofang:
OK I think we should be good, but plz run a bigger test coverage to make sure 🙂

leofang:
It seems the failed tests are related to https://github.com/cupy/cupy/pull/6624.

leofang:
https://github.com/cupy/cupy/pull/6646 is in, let's try again.

/test mini

leofang:
ah I was being silly... I ran the full tests before...

leofang:
@kmaehashi this is ready 🙂

kmaehashi:
Merging as test failures are unrelated.

leofang:
Thanks, @kmaehashi! To reply your question:

> Let me check if my understanding is correct for sure: `MemoryAsyncPool()` is a mempool backed by `handle=='current'`

Correct.

> (which is by default the "default" memory pool for the device, managed by CUDA runtime). 

Nit: by default it's using the "current" device mempool (as set by `cudaDeviceSetMemPool`, hence the name "current"), but if this API is not called anywhere, it uses the default mempool.

> So memory allocated in `MemoryAsyncPool` will not be freed even after `MemoryAsyncPool` instance is destroyed.

Correct.

PR #6597, opened 2022-03-29
add `cupy.insert` api

pri1311:
Linked Issue: #6078 

Hey! Have added the function for `numpy.insert` however, since subclassing is not yet supported in `CuPy`, it's not an exact replication of the `NumPy` function.

Looking forward to hearing from you :)


tests/cupy_tests/lib_tests/test_function_base.py
```diff
@@ -0,0 +1,76 @@
...
+    @testing.numpy_cupy_array_equal()
+    def test_insert_multi_dim_axis_none(self, xp, dtype):
+        a = xp.array([[1, 1], [2, 2], [3, 3]], dtype)
+        return xp.insert(a, 1, 5)
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_insert_multi_dim_with_axis(self, xp, dtype):
+        a = xp.array([[1, 1], [2, 2], [3, 3]], dtype)
+        return xp.insert(a, 1, 5, axis=1)
```
emcastillo, line 76: please test slices and invalid cases for multi dim too.
Thanks

cupy/lib/_function_base.py
```diff
@@ -0,0 +1,122 @@
...
+        axis = internal._normalize_axis_index(axis, ndim)
+    slobj = [slice(None)]*ndim
+    N = arr.shape[axis]
+    newshape = list(arr.shape)
+
+    if isinstance(obj, slice):
+        indices = cupy.arange(*obj.indices(N), dtype=cupy.intp)
+    else:
+        # need to copy obj, because indices will be changed in-place
+        indices = cupy.array(obj)
```
emcastillo, line 61: Here you need to be careful, creating a cupy array from a single integer can be expensive. It is faster to just pass the raw integer to the Cuda kernel since there are buffers for this.

cupy/lib/_function_base.py
```diff
@@ -0,0 +1,122 @@
...
+    arrorder = 'F' if arr.flags.fnc else 'C'
+    if axis is None:
+        if ndim != 1:
+            arr = arr.ravel()
+        # needed for np.matrix, which is still not 1d after being ravelled
+        ndim = arr.ndim
+        axis = ndim - 1
+    else:
+        axis = internal._normalize_axis_index(axis, ndim)
+    slobj = [slice(None)]*ndim
```
emcastillo, line 53: make the style through the code consistent please.
spaces should be added between operators

cupy/lib/_function_base.py
```diff
@@ -0,0 +1,122 @@
...
+    elif indices.size == 0 and not isinstance(obj, cupy.ndarray):
+        indices = indices.astype(cupy.intp)
+
+    indices[indices < 0] += N
+
+    numnew = len(indices)
+    order = indices.argsort()
+    indices[order] += cupy.arange(numnew)
+
+    newshape[axis] += numnew
```
emcastillo, line 111: variable naming should be consistent, separate words with `_` instead of mixing styles.

emcastillo:
Thanks for the PR.
Can you take some performance numbers for small, medium, and large matrices and compare them with numpy.
I feel that this approach may be slow.
You can use `cupyx.time.repeat` for this.

Also, I feel that using `RawKernel` and `CIndexer` can simplify the code a lot and make it much faster, but this is just a thought out of the wild.

takagi:
CuPy's file structure basically follows NumPy's API reference (although there may be some historical exceptions):
https://numpy.org/doc/stable/reference/routines.array-manipulation.html#adding-and-removing-elements

Would you move your implementation here?
https://github.com/cupy/cupy/blob/ec6d2c9a98440b19a133868576fb299972c21f6c/cupy/_manipulation/add_remove.py#L10

takagi:
(I mistakenly merged master https://github.com/cupy/cupy/pull/6597/commits/176279d11b6a386cb4b284b9a90aef74a03b84cd. Please don't care)

pri1311:
> Also, I feel that using RawKernel and CIndexer can simplify the code a lot and make it much faster, but this is just a thought out of the wild.

I was working on this, however I am little confused about the use of `CIndexer`. Any place(some implementation probably) where I could look up it's use to understand better? Apologies if it's a trivial doubt.
Cause I did use `cupy.profile.benchmark`, and the current code is definitely very slow.

emcastillo:
Yeah, `CIndexer` is only used in [here](https://github.com/cupy/cupy/blob/master/cupyx/scipy/ndimage/_filters_generic.py#L89-L90)
And it may not be easy to understand ...

For a bit of more information look:
https://docs.cupy.dev/en/stable/user_guide/kernel.html

And using `CArray` instead of `CIndexer` may be easier to understand. Check routines like
https://github.com/cupy/cupy/blob/master/cupy/_core/_routines_math.pyx#L492-L602

kmaehashi:
Let me close this one for now as it is in WIP for a while. If anyone is interested in taking over this PR please leave a comment.

PR #6617, opened 2022-04-03
Implemented a __repr__ for cupyx.profiler._time._PerfCaseResult

davidegavio:


.gitignore
```diff
@@ -24,3 +24,4 @@ htmlcov/
 .pytest_cache/
 cupy/.data/
 cupy/cuda/thrust.h
+.history
```
emcastillo, line 27: Please revert changes to this file.

cupyx/profiler/_time.py
```diff
@@ -26,6 +26,14 @@ def __init__(self, name, ts, devices):
         self._ts = ts
         self._devices = devices
 
+    def __repr__(self) -> str:
+        """ Returns a string representation of the object.
+
+        Returns:
+            str: A string representation of the object.
+        """
+        return f'name: {self.name}, ts: {self._ts}, devices: {self._devices}'
```
emcastillo, line 35: Unfortunately this does not give any information of the GPU time/ CPU time etc. Ideally we would like `__repr__` to just call `to_string`

kmaehashi:
Closes https://github.com/cupy/cupy/issues/6319

cupyx/profiler/_time.py
davidegavio: The call to `to_str(self, show_gpu=False)` method must be with `show_gpu=False` or `True`?

emcastillo:
Please remove `.gitignore` changes

emcastillo (approved):

PR #6619, opened 2022-04-04
JIT: Avoid calling default constructor if possible

asi1024:
Rel https://github.com/cupy/cupy/issues/5280#issuecomment-1087142600.

Combines variable declaration and initialization if the variable is first used at the top level.

cc: @leofang

### DIFF
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -163,7 +163,9 @@ class SharedMemory(BuiltinFunc):
         child_type = _cuda_types.Scalar(dtype)
         while env[name] is not None:
             name = env.get_fresh_variable_name(prefix='_smem')  # retry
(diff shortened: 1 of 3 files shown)


leofang:
Thanks for coming up the solution in such a short period @asi1024! I am sure this nice improvement would also help when we support Python classes 🙂 Will help review later tonight.

leofang (approved):
LGTM! I guess we don't need tests as #6620 would not work without this PR, but if there is a way to test this it'd be nice to have.

emcastillo:
Right now, I think we don't have constructors in the JIT that would use a default constructor, so the only thing I can think of for testing this is to parse the generated cuda code. Once #6620 is in, we will have the testing for free so I wouldn't care too much.

asi1024:
@emcastillo All tests have passed!

emcastillo (approved):
LGTM!

PR #6575, opened 2022-03-22
Avoid using the same kernel from different devices in JIT

asi1024:
Fix #5795 and #6574.

Adds `device_id` to the cache key.

PR #6620, opened 2022-04-04
Support cooperative group in JIT compiler

leofang:
~~Blocked by #6619.~~

This PR adds basic cooperative group support to JIT. Specifically, two [implicit groups](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#group-types-implicit-cg) are added:
- thread block group (constructed via `jit.cg.this_thread_block()`)
- grid group (constructed via `jit.cg.this_grid()`)

There's another implicit group (multi grid group) but not added, as `cudaLaunchCooperativeKernelMultiDevice` is deprecated since CUDA 11.3.

The `cg` namespace naming convention follows Numba's (which in turn follows CUDA's guide). 

To support this PR, a few changes are made
- Add the helper `method_as_builtin_func()` to turn class methods into CUDA functions
  - This would open up the possibility to support jitting Python classes in the future
  - I am not sure if this is the best approach, but this is the only approach I could think of to make it work...
- Register `Dim3` as a CUDA type
- Auto-detect if cooperative launch is needed
  - the argument `enable_cooperative_groups` is removed from `_JitRawKernel.__call__` because it makes no sense: Unlike `RawKernel`/`RawModule` in which case we can't detect if cooperative launch is needed and so have to give users full control, here we can totally handle it for users (and, conversely, if users already use cooperative group, it's an extra step for them to set the flag)
- Use Numba's Sphinx index for cross reference


cupyx/jit/_interface.py
```diff
@@ -138,7 +138,7 @@ def cached_codes(self):
             warnings.warn(
                 'No codes are cached because compilation is deferred until '
                 'the first function call.')
-        return [result.code for result in self._cached_codes]
+        return dict([(k, v.code) for k, v in self._cached_codes.items()])
```
leofang, line 141: This fixes a bug introduced in #6575 IIUC. I can backport the fix to v10 if needed.

leofang:
Looks like most of the grid group methods already existed back in CUDA 10.2, but no proper PTX code could be generated thus leading to a driver error at the module load time. I'll just disable the grid group support entirely on CUDA 10.2.

leofang:
@asi1024 this should be ready 🙂

cupyx/jit/_interface.py
asi1024: Thanks for the reporting! Could you split the fix into another PR?

cupyx/jit/_compile.py
```diff
@@ -138,6 +142,10 @@ def __init__(self):
         self.codes = []
         # (function, in_types) => Optional(function_name, return_type)
         self.device_function = {}
+        # whether to use cooperative launch
+        self.enable_cg = False
+        # whether to include cooperative_groups.h
+        self.cg_include = False
```
asi1024, line 148: :+1:

cupyx/jit/_interface.py
leofang: Will do!

leofang: See #6639.

cupyx/jit/_internal_types.py
```diff
@@ -73,3 +73,17 @@ def __init__(self):
 
     def __call__(self):
         raise RuntimeError('Cannot call this function from Python layer.')
+
+
+def method_as_builtin_func(instance_name, method):
```
asi1024, line 78: We are planning to implement `Callable` type as a subclass of `_cuda_types.TypeBase` and may erase this function in a process of a cosmetic change in the future.

docs/source/reference/cooperative_groups.rst
```diff
@@ -0,0 +1,11 @@
+:orphan:
+
+
+Cooperative groups
```
asi1024, line 4: How about adding them in `kernel.rst`?

cupyx/jit/_internal_types.py
leofang: Sounds good! Will BuiltinFunc be removed too? I think in this case it is still useful to use BuiltinFunc because they are from CUDA. (btw, this function could have been made as a classmethod... Look a bit nicer that way, so I will change it regardless of the plan for Callable.)

docs/source/reference/cooperative_groups.rst
leofang: Yes it makes sense too because the returned groups are public API. Will change.

leofang: Done in [ce85e2d](https://github.com/cupy/cupy/pull/6620/commits/ce85e2d7164471bac43d60f0f5b15b2151a3aa0e).

cupyx/jit/_internal_types.py
leofang: Done in 6eef9f3.

docs/source/reference/cooperative_groups.rst
leofang: I noticed that the class autosummary template would expose the parent class's methods (in this case, `declvar` and `assign`, see [here](https://cupy--6620.org.readthedocs.build/en/6620/reference/generated/cupyx.jit.cg._ThreadBlockGroup.html#cupyx.jit.cg._ThreadBlockGroup)), while Sphinx's `autoclass` by default turns it off. I feel it's a bit distracting...

leofang: dba542457e may be better?

cupyx/jit/_internal_types.py
asi1024: BuiltinFunc will not be removed

docs/source/reference/cooperative_groups.rst
asi1024: Thanks!

asi1024 (approved):

asi1024:
@leofang LGTM! Thanks!

leofang:
thx @asi1024!

docs/source/reference/cooperative_groups.rst
leofang: @asi1024 sorry, on a second thought `autoclass` doesn't look nice either... https://docs.cupy.dev/en/latest/reference/kernel.html#cooperative-groups
It expands all members on the otherwise index-only page. Time to reconsider making an orphan page for redirection?

leofang:
btw @asi1024 this should be tagged with v11.0.0b2?

PR #6622, opened 2022-04-05
Add test to ensure that nvrtc/nvcc pass for all supported compute capabilities

kmaehashi:
cc/ @takagi

### DIFF
--- a/tests/cupy_tests/core_tests/test_include.py
+++ b/tests/cupy_tests/core_tests/test_include.py
@@ -1,0 +1,79 @@
+import os
+from unittest import mock
+
+import cupy
+
+import pytest
+
+
+_code_base = '''
+#include <cuda_fp16.h>
+
+#include <cupy/carray.cuh>
+#include <cupy/complex.cuh>
+#include <cupy/cuComplex_bridge.h>
+#include <cupy/math_constants.h>
+#include <cupy/atomics.cuh>
+#include <cupy/hip_workaround.cuh>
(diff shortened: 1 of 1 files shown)


tests/cupy_tests/core_tests/test_include.py
```diff
@@ -0,0 +1,79 @@
...
+#include <cuda_fp16.h>
+
+#include <cupy/carray.cuh>
+#include <cupy/complex.cuh>
+#include <cupy/cuComplex_bridge.h>
+#include <cupy/math_constants.h>
+#include <cupy/atomics.cuh>
+#include <cupy/hip_workaround.cuh>
+#include <cupy/swap.cuh>
+#include <cupy/tuple.cuh>
```
leofang, line 19: Can we use Python to loop over the folder and pick up all headers programmatically? This way we would not miss any future additions.

kmaehashi: Yeah I tried it first but there is a dependency between headers 😞
(`atomics.cuh` requires `carray.cuh`)

leofang: I see. It might be a good idea to defer this when re-considering #3612.

leofang: (The context is one of the `atomicAdd` involves `float16`.)

kmaehashi:
merging to #6624

PR #6628, opened 2022-04-06
Add support for sparse matrices in `cupyx.distributed`

emcastillo:
There are several difficult things going on here that need to be discussed.

A proper way to send the shape/sizes of the internal matrix arrays, right now is done in nccl and this ends up syncing the device. Doing through the backing TCP store would be also pretty slow and will require modifying the store too for cases when a process is accessing elements that are not yet set.

cc @leofang do you have some thoughts on this?

Thanks!

### DIFF
--- a/cupyx/distributed/_nccl_comm.py
+++ b/cupyx/distributed/_nccl_comm.py
@@ -2,6 +2,7 @@ import cupy
 from cupy.cuda import nccl
 from cupyx.distributed import _store
 from cupyx.distributed._comm import _Backend
+from cupyx.scipy.sparse import issparse
 
 
 try:
@@ -125,6 +126,12 @@ class NCCLBackend(_Backend):
                 'Only nccl.SUM is supported for complex arrays')
         return _nccl_ops[op]
 
+    def _dispatch_arg_type(self, function, args):
+        comm_class = _DenseNCCLCommunicator
+        if issparse(args[0]):
+            pass
+        getattr(comm_class, function)(self, *args)
+
     def all_reduce(self, in_array, out_array, op='sum', stream=None):
         """Performs an all reduce operation.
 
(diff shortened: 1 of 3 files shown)


leofang:
I am wondering a few possibilities:
- Use a communication framework --- MPI, nvshmem, UCX, etc --- for CPU data like shapes and sizes
   - Maybe a good idea to pursue, since we still need one of them to bootstrap the multi processes?  
- For all participating sparse arrays, we preallocate device memory to store these metadata
   - we still need to sync (D2H) in order to create dense arrays for internal data

emcastillo:
> Use a communication framework --- MPI, nvshmem, UCX, etc --- for CPU data like shapes and sizes
Yeah, this is what I wanted to avoid though, I didn't want to have an external library for this if it could be avoided.

> For all participating sparse arrays, we preallocate device memory to store these metadata
This sounds nice, but yeah, we still need to do the dth transfer

emcastillo:
@leofang I was thinking and how about this approach?

We make mpi4py optional here, and if it is not installed we do all the exchange using nccl with device sync and we show a warning. Otherwise, we do it using mpi and hopefully overlapping the sending with nccl :D.

leofang:
That sounds reasonable. I have a stupid question, though: why didn't we run into this issue for dense arrays? The recv side still needs to know the shape from the send side, right?

emcastillo:
Yeah, with dense arrays we know the shapes before hand so there is no problem in populating these. but with sparse arrays, even though we know the global shape, we don't know the internal arrays details so we need to send the sizes beforehand

emcastillo:
Depends on #6658

I added the capability of using MPI for the sparse arrays exchange.

This is a small snipper to test send/recv

```python
import cupyx
import cupyx.distributed
from cupyx.scipy import sparse
import cupy
from mpi4py import MPI

comm = MPI.COMM_WORLD
rank = comm.Get_rank()

dev = cupy.cuda.Device(rank)
dev.use()
comm = cupyx.distributed.init_process_group(2, rank)


def _make(dtype):
    data = cupy.array([0, 1, 2, 3], dtype)
    indices = cupy.array([0, 1, 3, 2], 'i')
    indptr = cupy.array([0, 2, 3, 4], 'i')
    return sparse.csr_matrix((data, indices, indptr), shape=(3, 4))


def _make_empty(dtype):
    data = cupy.array([0], dtype)
    indices = cupy.array([0], 'i')
    indptr = cupy.array([0], 'i')
    return sparse.csr_matrix((data, indices, indptr), shape=(0, 0))


if rank == 0:
    matrix = _make('d')
    comm.send(matrix, 1)
    print('sent', matrix)
else:
    matrix = _make_empty('d')
    # We provide an empty sparse matrix to the routine
    # so it knows the sparse format, and it can just swap the
    # internal state
    comm.recv(matrix, 0)
    print('recv', matrix)
```


I need to do big tests changes for sparse arrays to work. Will do it soon!

kmaehashi (approved):

PR #6637, opened 2022-04-08
Bump Dockerfile version to v10.3.1

emcastillo:


kmaehashi:
Thanks! Actually I've already done this (hot-fix releases will be cut from `hotfix-v10.3.1` branched from `v10.3.0` tag) 😄 
https://github.com/cupy/cupy/tree/hotfix-v10.3.1

PR #6640, opened 2022-04-10
JIT: Add more cooperative group APIs

leofang:
~~Blocked by #6620. PR description to be updated.~~

This PR expands the cooperative group support (#6620). 4 more APIs are added:
- `cg.sync()`
- `cg.memcpy_async()`
- `cg.wait()`
- `cg.wait_prior()`

In order to utilize the optimization for certain alignments, I also added an extra argument in
- `cg.memcpy_async()`
- `shared_memory()`

to statically declare the arguments' alignment (in bytes).


cupyx/jit/cg.py
```diff
@@ -284,5 +305,130 @@ def call_const(self, env):
...
+                The behavior is undefined if the guarantee is not held.
+
+        .. seealso:: `cg::memcpy_sync`_
+
+        .. _cg::memcpy_sync:
+            https://docs.nvidia.com/cuda/archive/11.6.0/cuda-c-programming-guide/index.html#collectives-cg-memcpy-async
+        """
+        super().__call__()
+
+    def call(self, env, group, dst, dst_idx, src, src_idx, size, *,
```
leofang, line 354: Note: this signature is borrowing from the atomic functions. It seems each pointer address argument in C maps to two Python arguments (`array` & `index`). 

cc: @gmarkall in case Numba is planning on expanding the support for cooperative groups 🙂 (CuPy borrowed the atomic function signature from Numba, and it'd be nice to keep the signature consistent.)

gmarkall: Thanks for the CC - my plan for extending Numba support for this is to just enable directly calling the C++ equivalents from Numba kernels once this is merged: https://github.com/numba/numba/pull/7621

leofang: I see, so the planned approach is to ask Numba users to write CG-related codes directly in C++, and then link to a Numba kernel?

gmarkall: Ah, sorry - I didn't quite represent my intention properly - I'd write some C shim functions that call the C++ versions (similar to https://github.com/numba/numba-examples/pull/40/files#diff-e494235cb05f7f751143c66858983a68f1b7630a7a186154de92f213d80179a0) then have that shim automatically linked in by Numba if the user calls them. They'd write their CG-related code in Numba, but internally it would be using the C++ functions underneath it. 

We'd have to map pointer address arguments to the base and offset in Python for these as well (similar to what I'm doing for `curand_init` here: https://github.com/numba/numba-examples/pull/40/files#diff-e494235cb05f7f751143c66858983a68f1b7630a7a186154de92f213d80179a0R23, called like https://github.com/numba/numba-examples/pull/40/files#diff-14c8f7166d95bffa81bda8cdc5b386c5d47737132f63794019e386814464d4d3R32 from Python.

leofang:
@asi1024 This is ready for review, but I can't access Jenkins so I don't know what's the error there...

cupyx/jit/cg.py
```diff
@@ -284,5 +308,139 @@ def call_const(self, env):
...
+            group: a valid cooperative group
+
+        .. seealso:: `cg::sync`_
+
+        .. _cg::sync:
+            https://docs.nvidia.com/cuda/archive/11.6.0/cuda-c-programming-guide/index.html#collectives-cg-sync
+        """
+        super().__call__()
+
+    def call(self, env, group):
```
asi1024, line 326: How about checking `isinstance(group.ctype, _ThreadGroup)` here?

cupyx/jit/cg.py
```diff
@@ -284,5 +308,139 @@ def call_const(self, env):
...
+        if _runtime.runtimeGetVersion() < 11000:
+            raise RuntimeError("not supported in CUDA < 11.0")
+        _check_cg_include(env, target='memcpy_async')
+
+        dst = _Data.init(dst, env)
+        src = _Data.init(src, env)
+        for arr in (dst, src):
+            if not isinstance(
+                    arr.ctype, (_cuda_types.CArray, _cuda_types.Ptr)):
+                print(arr, arr.ctype)
```
asi1024, line 371:
```suggestion
```

asi1024:
Jenkins failure looks unrelated to this PR.

cupyx/jit/_cuda_types.py
```diff
@@ -86,17 +86,26 @@ def __hash__(self):
...
+            code = f'__align__({self._alignment})'
+        else:
+            code = ''
         if self._size is None:
-            return f'extern __shared__ {self.child_type} {x}[]'
-        return f'__shared__ {self.child_type} {x}[{self._size}]'
+            code = f'extern {code} __shared__ {self.child_type} {x}[]'
+        else:
+            code = f'{code} __shared__ {self.child_type} {x}[{self._size}]'
+        return code
```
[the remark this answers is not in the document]
leofang: Hi @asi1024 there is a subtle difference in the current implementation: `size` is a positional argument (either `None` or an integer), while `alignment` is a kw argument. The intention is for users to always specify `size`, but not `alignment` unless deemed necessary (ex: in the case of calling `memcpy_async`).

cupyx/jit/cg.py
```diff
@@ -1,11 +1,41 @@
...
 from cupyx.jit._internal_types import Data as _Data
 
 
 # public interface of this module
-__all__ = ['this_grid', 'this_thread_block']
+__all__ = ['this_grid', 'this_thread_block',
+           'sync', 'wait', 'wait_prior', 'memcpy_async']
+
+
+def _check_cg_include(env, target='cg'):
```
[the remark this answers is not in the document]
leofang: Will do.

cupyx/jit/cg.py
leofang: Good catch..

leofang: Fixed in 6e759600.

cupyx/jit/cg.py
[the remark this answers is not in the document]
leofang: In 6e759600 I combined them into `_check_include` and further reduced the code duplication. Let me know if this works 🙂

cupyx/jit/cg.py
```diff
@@ -1,11 +1,41 @@
...
+    if target == 'cg':
+        included = env.generated.include_cg
+        if included is False:
```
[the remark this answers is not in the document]
leofang: Done in 6e759600.

cupyx/jit/cg.py
leofang: Done in 6e759600.

asi1024:
LGTM! Thanks!

asi1024 (approved):

PR #6646, opened 2022-04-12
Fix `TestIncludesCompileCUDA` for HEAD tests

emcastillo:


kmaehashi:
Thanks! CUDA 11.2 seems also failing when compiling for Tegra CCs on non-Tegra.
I think it's better to do `set(_get_cuda_archs()) - set(cupy.cuda.compiler._tegra_archs)` for CUDA 10.2~11.2.

kmaehashi:
Thanks!!

PR #6648, opened 2022-04-12
JIT: fix function signatures in the docs

### LINKED ISSUE #5390: JIT: Obscure function signature in the documentation
For example: 
https://docs.cupy.dev/en/latest/reference/generated/cupyx.jit.shfl_up_sync.html
截圖 2021-06-23 下午12 33 32
The function signature here reveals no information other than the internal object `WarpShuffleOp` (which we don't want to expose). Ideally Sphinx would take the function signature of `__call__` and display it, but for some reason it is not working properly...This happens in most of objects under `cupyx.jit`.

leofang:
Close #5390. 

Still not perfect but a lot better than before. It'd be nice to backport this.

before | after 
-------|------
截圖 2022-04-12 上午3 02 39 | 截圖 2022-04-12 上午3 02 49

before | after 
-------|------
截圖 2022-04-12 上午3 03 36 | 截圖 2022-04-12 上午3 03 47

### DIFF
--- a/docs/source/reference/kernel.rst
+++ b/docs/source/reference/kernel.rst
@@ -14,7 +14,7 @@ Custom kernels
 JIT kernel definition
 ---------------------
 
-Supported Python built-in functions include: :func:`range`, :func:`len`, :func:`max`, :func:`min`
+Supported Python built-in functions include: :obj:`range`, :func:`len`, :func:`max`, :func:`min`
 
 .. autosummary::
    :toctree: generated/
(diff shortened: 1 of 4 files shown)


docs/source/reference/kernel.rst
```diff
@@ -14,6 +14,8 @@ Custom kernels
 JIT kernel definition
 ---------------------
 
+Supported Python built-in functions include: :obj:`range`, :func:`len`, :func:`max`, :func:`min`
```
leofang, line 17: Note: ":class:`range`" does not work, `range` does not work, either.

asi1024 (approved):

asi1024:
LGTM. Thanks!

PR #6228, opened 2021-12-09
Change a parameter name in `percentile` and `quantile` to support NumPy 1.22

takagi:
Partially fix #6198.

`percentile` and `quantile`'s `interpolation` parameter is renamed to `method` in NumPy 1.22, and some new interpolation methods are introduced as well. In this PR, I cover the change in the parameter name only to support NumPy 1.22, and will open another issue to implement the newly introduced methods.

PR #6658, opened 2022-04-18
Use MPI for management in `cupyx.distributed`

emcastillo:
This is needed by #6228
We will use MPI to efficiently exchange the sizes of sparse matrices because doing it with NCCL causes device synchronization and the TCP store will kill performance when used heavily.

### DIFF
--- a/cupyx/distributed/_nccl_comm.py
+++ b/cupyx/distributed/_nccl_comm.py
@@ -4,6 +4,13 @@ from cupyx.distributed import _store
 from cupyx.distributed._comm import _Backend
 
 
+try:
+    from mpi4py import MPI
+    _mpi_available = True
+except ImportError:
+    _mpi_available = False
+
+
 if nccl.available:
     # types are not compliant with windows on long/int32 issue
     # but nccl does not support windows so we don't care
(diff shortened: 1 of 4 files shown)


cupyx/distributed/_nccl_comm.py
```diff
@@ -44,25 +51,49 @@ class NCCLBackend(_Backend):
...
+            # make them positive and send them as bytes to the proxy store
+            shifted_nccl_id = bytes([b + 128 for b in nccl_id])
+            self._store_proxy['nccl_id'] = shifted_nccl_id
             self._store_proxy.barrier()
         else:
             self._store_proxy.barrier()
             nccl_id = self._store_proxy['nccl_id']
-        # Initialize devices
-        nccl_id = tuple([int(b) - 128 for b in nccl_id])
+            nccl_id = tuple([int(b) - 128 for b in nccl_id])
```
leofang, line 96: `nccl_id` is gone for rank 0?

tests/cupyx_tests/distributed_tests/comm_runner.py
```diff
@@ -46,91 +46,118 @@ def run_broadcast(rank, n_workers, root, dtype):
...
-    _launch_workers(N_WORKERS, run_broadcast, (0, dtype))
-    _launch_workers(N_WORKERS, run_broadcast, (1, dtype))
+    if use_mpi:
+        from mpi4py import MPI
+        # This process was run with mpiexec
+        run_broadcast(MPI.COMM_WORLD.Get_rank(), 0, dtype, False)
+        run_broadcast(MPI.COMM_WORLD.Get_rank(), 1, dtype, False)
+    else:
+        _launch_workers(run_broadcast, (0, dtype))
+        _launch_workers(run_broadcast, (1, dtype))
```
leofang, line 56: Can't we do this inside `_launch_workers` to avoid code dup/branching? Something like
```
def _launch_workers(func, args, ..., use_mpi=False):
    if use_mpi:
        from mpi4py import MPI
        func(...)
    else:
        # the old _launch_workers impl
```

leofang, line 56: Can't we do this inside `_launch_workers` to avoid code dup/branching? Something like
```python
def _launch_workers(func, args, ..., use_mpi=False):
    if use_mpi:
        from mpi4py import MPI
        func(...)
    else:
        # the old _launch_workers impl
```

emcastillo: I tried, but the code was hard to follow and the only repeated lines are 2, 
```
nccl_id = nccl.get_unique_id()
...
self._comm = nccl.NcclCommunicator(n_devices, nccl_id, rank)
```
and for the older case since we need to manipulate the nccl_id it becomes quite of spaghetti code with ifs ...
I think this approach is much simpler to read

kmaehashi (approved):

PR #6668, opened 2022-04-20
JIT: Support `.shape` and `.strides`

leofang:
Based on the discussion between @eternalphane and @asi1024 in #5293.

It'd take some effort to support arrays of `ndim>10` (due to a Thrust limitation, see https://github.com/cupy/cupy/pull/5293#issuecomment-1101766792), but at least with this PR we unblock the most common use cases.

### DIFF
--- a/cupy/_core/include/cupy/carray.cuh
+++ b/cupy/_core/include/cupy/carray.cuh
@@ -326,6 +326,16 @@ public:
     return strides_;
   }
 
+#ifdef CUPY_JIT_MODE
+  __device__ typename cupy::as_tuple<_ndim>::type get_shape() const {
+    return cupy::as_tuple<_ndim>::call(shape_);
+  }
+
+  __device__ typename cupy::as_tuple<_ndim>::type get_strides() const {
+    return cupy::as_tuple<_ndim>::call(strides_);
+  }
+#endif  // CUPY_JIT_MODE
+
 #if __cplusplus >= 201103 || (defined(_MSC_VER) && _MSC_VER >= 1900)
   template <typename Int>
   __device__ T& operator[](const std::initializer_list<Int> idx_) {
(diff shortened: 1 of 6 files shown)


cupyx/jit/_compile.py
```diff
@@ -768,7 +772,8 @@ def _indexing(array, index, env):
     if isinstance(array.ctype, _cuda_types.Tuple):
         if is_constants(index):
             i = index.obj
-            return Data(f'thrust::get<{i}>({array.code})', array.types[i])
+            t = array.ctype.types[i]
+            return Data(f'thrust::get<{i}>({array.code})', t)
```
leofang, line 776: This is a bug fix which I am not sure if we should backport, because apparently it's not caught by the test suite.

cupy/_core/include/cupy/carray.cuh
```diff
@@ -326,6 +326,16 @@ public:
     return strides_;
   }
 
+#ifdef CUPY_JIT_MODE
+  __device__ typename cupy::as_tuple<_ndim, ptrdiff_t>::type get_shape() const {
```
leofang, line 330: Note: The `as_tuple` implementation by @asi1024 is extremely clever, but it wouldn't be needed if we move JIT from C++11 to C++14, where `std::index_sequence` (and its relatives) and `auto` as the return type  would just work...😉

asi1024: I am concerned that C++14 will work with old CUDA/HIP versions. I will check if CI passes with c++14 option in another PR.

leofang: Thanks, @asi1024. I think C++11 is ok for now, we don't have to bump it right now, but it'd be good to explore a timeline for it 🙂

leofang:
hmmm Jenkins failed...?

asi1024:
Jenkins failure looks unrelated to this PR.

asi1024 (approved):

asi1024:
LGTM. Thanks!

leofang:
Thank you, @asi1024 (and, of course, @eternalphane) 🙂

PR #6676, opened 2022-04-25
[v10] Skip `ndimage.filter` tests for ROCm 4.0

emcastillo:


kmaehashi:
@emcastillo Seems closed unexpectedly?

Issue #6656, opened 2022-04-14
Support cuTENSOR 1.5.0


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read Release Notes ([cuTENSOR](https://docs.nvidia.com/cuda/cutensor/index.html#changelog) / [cuSPARSELt](https://docs.nvidia.com/cuda/cusparselt/release_notes.html) / [cuDNN](https://docs.nvidia.com/deeplearning/cudnn/release-notes/index.html) / [NCCL](https://docs.nvidia.com/deeplearning/sdk/nccl-release-notes/index.html)).
- FlexCI
  - [x] Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml` #6665
  - [x] Backport. #6722
- Library Installer
  - [x] Update `cupyx/tools/install_library.py`. #6665
  - [x] Backport. #6722
- Documentation
  - [x] Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) #6665
  - [x] Backport. #6722
- [x] Implement new features using the version, if any. #6665

leofang:
I think https://github.com/cupy/cupy/pull/6665 closes this 🙂

PR #6677, opened 2022-04-25
Support cuTensorNet (from cuQuantum) as an `einsum` backend

### LINKED ISSUE #6261: RFC: Support performant `einsum()` and `einsum_path()` using cuQuantum
### Description

Related to #6078.

The new [cuQuantum SDK](https://developer.nvidia.com/cuquantum-sdk) provides two libraries, one of which is [cuTensorNet](https://docs.nvidia.com/cuda/cuquantum/cutensornet/index.html) that accelerates tensor network contraction (backed by cuTENSOR). Specifically, cuTensorNet currently handles two major challenges from problems that have a huge Einsum expression:
1. Finding the optimal contraction path
2. Executing the actual pairwise contraction for a given path in an optimal fashion

Those two functionalities map very nicely to NumPy's `einsum_path()` and `einsum()` APIs, respectively. As a result, in [cuQuantum Python](https://docs.nvidia.com/cuda/cuquantum/python/), the Python binding for cuQuantum libraries, we provide pythonic APIs [`cuquantum.einsum_path()`](https://docs.nvidia.com/cuda/cuquantum/python/api/generated/cuquantum.einsum_path.html) and [`cuquantum.einsum()`](https://docs.nvidia.com/cuda/cuquantum/python/api/generated/cuquantum.einsum.html) among other things. We currently provide two conda packages, `cuquantum` and `cuquantum-python`, and a pip wheel release is in our near-term roadmap.

I would like to propose using cuQuantum Python to back CuPy's `einsum_path()` and `einsum()` when available. Performance improvement data can be shared if needed. However, there are a few issues to be addressed:
- Because CuPy is currently a required dependency of cuQuantum Python, we need to break any potential circular import.
- For the initial release, `cuquantum.einsum_path()` and `cuquantum.einsum()` are not yet a 100% drop-in replacement of their NumPy counterparts (we mark most optional arguments as unsupported).
- Directly using our drop-in replacement APIs has a small performance penalty (due to the library handle not reused).
- We currently only support classical einsum; ellipsis, broadcasting, etc are not yet supported 

Below is my 4-step proposal, involving a small refactoring of the existing codebase:
1. Add `cupy.einsum_path()` as a fallback: The functionalities are already in `cupy/linalg/_einsum.py`, we just need to put everything together.
2. Introduce `ACCELERATOR_CUQUANTUM` as a optional routine accelerator backend: This way, we allow users to set `CUPY_ACCELERATORS=cub,cutensor,cuquantum`. This is desired to be able to test things separately (ex: [see here](https://github.com/NVIDIA/cuQuantum/blob/4e63ab07d0d1bce2fb53e8eebcb3818d9444dbc5/python/tests/cuquantum_tests/cutensornet_tests/__init__.py#L4-L10)).
3. Use the core functionalities of cuTensorNet to back `cupy.einsum_path()` and `cupy.einsum()`: I don't think I'd like to use `cuquantum.einsum_path()` and `cuquantum.einsum()` directly due to the aforementioned reasons, so most likely we'll need to add some helper functions to check
   - If `ACCELERATOR_CUQUANTUM` is requested
   - If `cuquantum` can be imported
   - If all user inputs can be handled by cuTensorNet
   - If so, forward to cuQuantum Python, otherwise fall back 
4. Add some mock tests to ensure the backend can be used when available

### Additional Information

_No response_

leofang:
This PR
- closes #6261 
- supports using cuTensorNet as an `einsum` backend for
   - path optimization (by setting `optimize="cutensornet"`)
   - tensor contraction
- adds `cutensornet` as a valid CuPy accelerator
- fixes the `einsum` docstring
   - Commits f92be944cd864d3635f89d7a0e0b219ad8b41a47 and 18d32e632d587279ce1a9d4c121e3358d96fd4ce can be backported
- requires cuquantum-python >= 22.03 and cuTENSOR >= 1.5.0 (#6656)

In terms of implementation, for both path finding and actual contraction I simply hijack the `einsum` call and delegate to ``cuquantum.contract()`` (cuQuantum Python's public APIs that use cuTensorNet). The reason includes:
- the way that NumPy/CuPy internally computes the input/output subscripts makes it very hard to just intercept at [the pairwise contraction loop](https://github.com/cupy/cupy/blob/ef19f7107ecac0ad2f3533e44e0b8cbbb20f0b29/cupy/linalg/_einsum.py#L641) (as in the cuTENSOR's case)
- In cuQuantum Python we made significant efforts to parse/validate the inputs and can generate better error messages if an error occurs, so it'd be a waste to do input verification twice (in CuPy and then in cuQuantum Python) instead of reusing it
- cuQuantum Python supports enormously large inputs that are otherwise incomprehensible to NumPy/CuPy due to the way the unicode string or interleaved inputs are parsed. (Those cases usually require calling [`opt_einsum.contract()`](https://dgasmith.github.io/opt_einsum/api_reference/#opt_einsumcontract) + NumPy/CuPy as its backend.) 

Note that I did not do any refactoring as proposed in #6261 to support `cupy.einsum_path()` for reasons discussed later in that issue. I will send a follow-up PR to add it.


leofang:
Let me do a test drive (should be a no-op before a CI with cuQuantum installed is set up 🙂).

/test mini

cupy/linalg/_einsum.py
```diff
@@ -444,29 +445,54 @@ def _tuple_sorted_by_0(zs):
 
 
 def einsum(*operands, **kwargs):
-    """einsum(subscripts, *operands, dtype=False)
+    """einsum(subscripts, *operands, dtype=None, optimize=False)
```
takagi, line 448: How about making `optimize` a keyword-only argument? (maybe `dtype` too, as `numpy.einsum` has `out` followed by it. though it can be another issue)

takagi: ah, it's in docstring. let me withdraw this comment.

leofang: Oh yes, this first line is needed to overwrite the function signature in Sphinx. This is a fix that can be backported (see the PR description) because the current docstring isn't accurate.

takagi: I see, thanks!

takagi:
I locally tried the cuTensorNet-based einsum, and it did work. Super cool! Should we have some CI condition that tests `_einsum_cutn` (possibly in another PR), @kmaehashi?

takagi:
Installing cuQuantum Python also installs libcutensor.so which is independent of one CuPy sees to have `cutensor` accelerator. I think this fact would make no harm.

takagi (dismissed):
LGTM!

leofang:
> Should we have some CI condition that tests `_einsum_cutn` (possibly in another PR)

That'd be great, but I'm not sure if you'd wanna run the full test suite or just target `einsum`.

> Installing cuQuantum Python also installs libcutensor.so which is independent of one CuPy sees to have `cutensor` accelerator. I think this fact would make no harm.

Yes. That's why it'd be nice to get #6656 in, otherwise depending on the import order we could be loading a different cuTENSOR (I hit this issue during local development). With cuTENSOR 1.5.0+ it should not matter which gets loaded first.

takagi:
CI cuda102 complains:
```
#6 48.07 E: Unable to locate package libcudnn7
#6 48.07 E: Unable to locate package libcudnn7-dev
```

leofang:
> Maybe removed from Nvidia's apt repository?

Fixed by #6708. 

/test mini

leofang:
Oh I was being silly and forgot to merge with the master branch...

takagi:
Now that cuTENSOR 1.5 support is marged to master, I've enabled auto-merge. The backport follows soon.

leofang:
Thanks, @takagi!

Issue #6611, opened 2022-03-31
CuPy int8 matmul takes much longer time than float32 and float16 on GTX 3090


ziyangjoy:
Hi,

I am trying to use CuPy to implement some low precision algorithm, so I need int8 matrix multiplications in the implementation. However, I find that the running time of int8 multiplications is much longer than float16 and float32 on GTX 3090. 

The cupyx.profiler.benchmark is used to record the GPU time. For the matrix multiplication of dimensions 3000 * 3000 and 3000 * 3000, int8 takes 6645us, float32 takes 2792us, and float16 takes 964us on GTX 3090. It looks like the float32 and float16 times are normal, but the int8 time is much longer than the float ones. 

The following are the codes used to test running times. 

```
import numpy as np
import cupy as cp
from cupyx.time import repeat

a = cp.random.randint(-30,high=30,size=[3000,3000],dtype = cp.int8)
b = cp.random.randint(-30,high=30,size=[3000,3000],dtype = cp.int8)


f = lambda a,b: cp.matmul(a,b)


print('matmul int8 results')
print(repeat(f, (a, b), n_repeat=100),'\n')



a = a.astype(cp.float32)
b = b.astype(cp.float32)


print('matmul float32 results')
print(repeat(f, (a, b), n_repeat=100),'\n')


a = a.astype(cp.float16)
b = b.astype(cp.float16)


print('matmul float16 results')
print(repeat(f, (a, b), n_repeat=100),'\n')
```

emcastillo:
We use a custom kernel extracted from MAGMA for integer matrix multiplication that's probably the reason for low performance as it is mostly optimized for different types. A solution for this would be to try [cutlass](https://github.com/NVIDIA/cutlass).

https://github.com/cupy/cupy/blob/master/cupy/_core/_routines_linalg.pyx#L96-L287

We would welcome a PR that optimizes this.

ziyangjoy:
> We use a custom kernel extracted from MAGMA for integer matrix multiplication that's probably the reason for low performance as it is mostly optimized for different types. A solution for this would be to try [cutlass](https://github.com/NVIDIA/cutlass).
> 
> https://github.com/cupy/cupy/blob/master/cupy/_core/_routines_linalg.pyx#L96-L287
> 
> We would welcome a PR that optimizes this.

Thanks a lot for your explanation! I will check the suggested solution there.

andoorve:
[...]

PR #6703, opened 2022-05-06
Improved integer matrix multiplication performance by modifying tuning parameters

andoorve:
This pull request close to doubles performance compared to the baseline, and is comparable with the volta sgemm kernel fp32 performance on RTX 2070s.


andoorve:
Should improve performance on #6611

leofang:
This is interesting, thanks for the tuning work! Is the value determined based on some heuristics? Could you share the perf numbers before/after the change?

andoorve:
Sure @leofang, here are the numbers before and after (Platform: RTX 2070 Super, Ryzen 3700x) on the script in #6611:

Before: GPU-0:19393.219 us   +/-283.384 (min:19037.249 / max:19775.520) us
After: GPU-0: 9992.426 us   +/-277.752 (min: 9629.440 / max:10451.392) us
Improvement: ~1.94x

The rough heuristic I tried to use was maximizing SMEM usage/minimizing memory accesses + minimizing uncoalesced accesses *(Note that this kernel is templated for different dtypes, which may cause uncoalesced accesses with smaller dtypes that would not be apparent with larger dtypes)*. However, even some parameters that should be better based on this heuristic lead to worse performance, so it seems the heuristic is not super accurate, and the kernel is very sensitive to changes in the tuning parameters. 

Profiling suggests that compared with the equivalent fp32 volta sgemm the effective memory throughput is much worse, which is another area to look at.

However, compared to the original kernel, SM and FMA pipeline utilization is much higher, probably indicating the larger sizes we are working with using these tuning parameters compensates/overlaps with memory transfers making it less of an issue. This is further indicated by the higher % of eligible warps for the new tuning parameters.

emcastillo:
Thanks for all the effort in fine tuning this!
Just glazed through the numbers and they make complete sense for modern architectures.

I originally just ripped the magma kernel without considering anything else as the main goal was to solve numerical issues because we were converting the arrays to fp since cublas does not support integers.

leofang:
Looks like there are some numerical issues. IIRC those tuning parameters are constrained. Maybe this patch broke some assumptions.

andoorve:
I’ll have to take a closer look to see what’s causing this.

emcastillo:
Thanks! The kernel has some fishy stuff where out-of-bound accesses are happening shouldn't affect but who knows ...
We are willing to use a different kernel or switch to cutlass if possible. So feel free to do the thing that you feel is easier. :)

andoorve:
So I took a look at why it wasn't working again after a few days. It seems like the constraints to ensure correctness are that, at least from a rough look, the following:

$$\mathtt{dim_x \cdot dim_y = blk_m \cdot blk_k = blk_n \cdot blk_k = dim_{xa} \cdot dim_{ya} = dim_{xb} \cdot dim_{yb}}$$

Under these constraints from whatever quick testing I've done the original parameters seem to work best. Hence, using cutlass or a different kernel may be a better option as @emcastillo suggested.

andoorve:
Another quick update. According to the Cutlass [README](https://github.com/NVIDIA/cutlass) the integer types supported for multiplication are 32 bit, 64 bit, and 1 byte booleans. From the below lines it seems like cupy supports a much wider range of integer types. Hence, I'm not sure if cutlass will work, at least not directly. @emcastillo did you have any thoughts https://github.com/cupy/cupy/blob/1c55be4179e93470afe994daa864b568ee6de7a5/cupy/_core/_routines_linalg.pyx#L291-L301

andoorve:
So, I had a chance to take a quick look again, and improve the tuning slightly. The LSU utilization goes down 15% from around 85% to 60%. The improvement is below.

Before: 19561.722 us   +/-1014.614 (min:17894.272 / max:21265.888) us
After: 15132.362 us   +/-2005.728 (min:13217.984 / max:18647.297) us
Improvement: 1.3x

So, there is a clear improvement, but it is not significant when compared to the FP32 and FP16 performance. The kernel may need to be rewritten to support better load store semantics, but hopefully this tuning improvement is still useful until more effort can be invested in this. @emcastillo @leofang

emcastillo:
Thanks a lot for trying this again!! Hope this time the tests won't have problems :)

emcastillo (approved):

PR #6704, opened 2022-05-08
Fix memory leak in the FFT plan cache during multi-threading

### LINKED ISSUE #6355: Multithreaded cufft memory leak
### Description

The cufft plan cache does not appear to deallocate GPU memory during thread clean up (GC).  If I launch `cp.fft.fft` on a thread and then exit/join the thread, there is residual memory still allocated on the GPU.

### To Reproduce

```py
from concurrent.futures import ThreadPoolExecutor
import cupy as cp

def task():
    data = cp.ones(2**20)
    return cp.fft.fft(data)

while True:
    with ThreadPoolExecutor(max_workers=3) as executor:
        tasks = [executor.submit(task) for _ in range(10)]
        
    cp.get_default_memory_pool().free_all_blocks()
    cp.get_default_pinned_memory_pool().free_all_blocks()
    cp.fft.config.get_plan_cache().clear()
    
    input("Check nvidia-smi memory usage... Press any key to run again")

```

### Installation

### Environment

```
OS                           : Linux-3.10.0-1160.36.2.el7.x86_64-x86_64-with-glibc2.29
Python Version               : 3.8.10
CuPy Version                 : 10.0.0
NumPy Version                : 1.21.5
SciPy Version                : 1.7.3
Cython Build Version         : 0.29.24
CUDA Build Version           : 11040
CUDA Driver Version          : 11040
CUDA Runtime Version         : 11040
cuFFT Version                : 10502
cuRAND Version               : 10205
cuSOLVER Version             : (11, 2, 0)
NVRTC Version                : (11, 4)
Thrust Version               : 101201
CUB Build Version            : 101201
Jitify Build Version         : 60e9e72
NCCL Build Version           : 21104
NCCL Runtime Version         : 21104
Device 0 Name                : Quadro RTX 5000

```

### Additional Information

Runtime is within docker image `nvidia/cuda:11.4.2-runtime-ubuntu20.04`

leofang:
Close #6355.

### DIFF
--- a/cupy/fft/_cache.pyx
+++ b/cupy/fft/_cache.pyx
@@ -274,6 +274,9 @@ cdef class PlanCache:
         self._reset()
         self.dev = dev if dev != -1 else runtime.getDevice()
 
+    def __dealloc__(self):
+        self._cleanup()
+
     def __getitem__(self, tuple key):
         # no-op if cache is disabled
         if not self.is_enabled:
@@ -281,8 +284,6 @@ cdef class PlanCache:
             return
 
         cdef _Node node
-        cdef int dev
-        cdef PlanCache cache
         cdef list gpus
 
         node = self.cache.get(key)


leofang:
Notes that I've taken when researching the solution:
- Thread local storages should automatically get cleaned up when a thread is joined; users need not worry about it
- The root cause here likely involves some reference cycles (why?) preventing the garbage collector from doing its job
  - This patch would break the cycle
- Python methods should not be called in `__dealloc__`, but apparently `cdef` methods are fine
- `__dealloc__` holds the GIL so it's safe to manipulate Python objects

kmaehashi (approved):

PR #6707, opened 2022-05-10
Fix bad appearance of the CG docs

leofang:
Follow-up of https://github.com/cupy/cupy/pull/6620#discussion_r848458223. This PR fixes the bad looking of the docs (see [here](https://docs.cupy.dev/en/latest/reference/kernel.html#cooperative-groups)).

~~The solution here is to not expand the classes' docs in the API list, but rather to create an orphan page _and_ hyperlink to it. The orphan page is needed because we want to use `autoclass` instead of `autosummary` to avoid documenting parent attributes/methods. This is a mix of the two previously tried approaches in #6620.~~


leofang:
hmm I wonder why the CI failed but my local build didn't...

leofang:
Alright, we need to revert to the old approach to make it pretty... 😞

leofang:
@asi1024 WDYT?

asi1024:
@leofang Sorry for my late response.
I am concerned that if the Cooperative Groups section is separated from the JIT section on different pages, it will be hard to search for features.
How about splitting the whole JIT section including the Cooperative Groups section from the Custom Kernel page?

leofang:
But as long as the group classes are rendered on the same page as other functions, we always have the poor appearance that the fonts of the class methods are larger than those of the standalone functions, moving out the JIT section does not help here 😞

leofang:
Thanks, @kmaehashi. Let's close this one!

PR #6712, opened 2022-05-11
Document the returned benchmark object

leofang:
Make it a bit easier to understand what to do with `cupyx.profiler.benchmark()`.

### DIFF
--- a/cupyx/profiler/_time.py
+++ b/cupyx/profiler/_time.py
@@ -112,7 +112,8 @@ def benchmark(
             the timing test. If not given, the current device is used.
 
     Returns:
-        :class:`_PerfCaseResult`: an object collecting all test results.
+        :class:`~cupyx.profiler._time._PerfCaseResult`:
+            an object collecting all test results.
 
     """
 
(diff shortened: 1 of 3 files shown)


cupyx/profiler/_time.py
```diff
@@ -112,7 +112,8 @@ def f(a, b):
             the timing test. If not given, the current device is used.
 
     Returns:
-        :class:`_PerfCaseResult`: an object collecting all test results.
+        :class:`~cupyx.profiler._time._PerfCaseResult`:
```
emcastillo, line 115: If we are documenting this, should we remove the underscore?

leofang: It's a good question... For now I am taking the view that the objects which 
- users should not construct themselves, but 
- can be returned by public APIs 

should keep the underscore prefix. This unifies the treatment for the cooperative group APIs.

emcastillo: Seems reasonable!

emcastillo (approved):

leofang:
@kmaehashi do we still have time to backport this? 🙂

kmaehashi:
Looks better to backport 👍🏼

leofang:
Thanks!

PR #6716, opened 2022-05-13
Support cupy.ndarray subclassing - Part 1 - Direct constructor call

takagi:
Merge #6718 first.

This PR is a part of supporting `cupy.narray` subclassing. I implement Direct constructor call here.
- Define `cupy._core.core._ndarray` that wraps cdef'ed `cupy._core.core.ndarray`
- Replace codes where directly instantiating ndarrays
- Fix parts where an object is compared to `cupy._core.core.ndarray`
- Add a test


takagi:
I'm making renames from `ndarray` to `ndarray_base` and `_ndarray` to `ndarray`, but perhaps it won't be in time for beta3 release as it has a lot of lines changed. In such a case, please consider including commits so far to beta3, which passed the CI.

kmaehashi (approved):

PR #6718, opened 2022-05-13
Use exposed name in user guide

takagi:
Rel #6716. This PR removes an unnecessarily used internal name in the user guide, which caused a `cuda-example` CI failure in #6716.


emcastillo:
blocked by #6716

takagi:
This is blocking #6716.

emcastillo (approved):

PR #6720, opened 2022-05-13
Support cupy.ndarray subclassing - Part 2 - View casting

takagi:
Merge #6746 first.

This PR is a part of supporting `cupy.ndarray` subclassing. Following #6716, I implement View casting.

### DIFF
--- a/cupy/_core/core.pxd
+++ b/cupy/_core/core.pxd
@@ -108,7 +108,7 @@ cpdef _ndarray_base array(
     obj, dtype=*, bint copy=*, order=*, bint subok=*, Py_ssize_t ndmin=*)
 cpdef _ndarray_base _convert_object_with_cuda_array_interface(a)
 
-cdef _ndarray_base _ndarray_init(const shape_t& shape, dtype)
+cdef _ndarray_base _ndarray_init(subtype, const shape_t& shape, dtype, obj)
 
 cdef _ndarray_base _create_ndarray_from_shape_strides(
     const shape_t& shape, const strides_t& strides, dtype)
--- a/cupy/_core/core.pxd
+++ b/cupy/_core/core.pxd
@@ -33,7 +33,7 @@ cdef class _ndarray_base:
     cpdef _ndarray_base astype(
         self, dtype, order=*, casting=*, subok=*, copy=*)
     cpdef _ndarray_base copy(self, order=*)
-    cpdef _ndarray_base view(self, dtype=*, array_class=*)
+    cpdef _ndarray_base _view_impl(self, dtype=*, array_class=*)
     cpdef fill(self, value)
     cpdef _ndarray_base swapaxes(self, Py_ssize_t axis1, Py_ssize_t axis2)
     cpdef _ndarray_base flatten(self, order=*)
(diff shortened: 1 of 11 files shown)


cupy/_core/core.pyx
```diff
@@ -589,7 +615,7 @@ cdef class ndarray:
             newarray.data.copy_from_device_async(x.data, x.nbytes)
         return newarray
 
-    cpdef ndarray view(self, dtype=None):
+    cpdef ndarray view(self, dtype=None, typ=None):
```
emcastillo, line 618:
```suggestion
    cpdef ndarray view(self, dtype=None, type=None):
```

cupy/_core/core.pxd
```diff
@@ -32,7 +32,7 @@ cdef class ndarray:
     cpdef bytes dumps(self)
     cpdef ndarray astype(self, dtype, order=*, casting=*, subok=*, copy=*)
     cpdef ndarray copy(self, order=*)
-    cpdef ndarray view(self, dtype=*)
+    cpdef ndarray view(self, dtype=*, typ=*)
```
takagi, line 35: Seem that Cython does not allow using `type` in an argument list. Also, `numpy.ndarray.view` has the signature `ndarray.view([dtype][, type])`, so I suppose we should have (at most)two positional arguments for `cupy.ndarray.view` instead of keyword args.

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.view.html

cupy/_core/core.pyx
```diff
@@ -615,7 +615,7 @@ cdef class ndarray:
...
```
takagi, line 618: ditto (core.pxd)

cupy/_core/core.pxd
emcastillo: Thanks for the clarification, I thought it was a typo at first.
How about naming it `array_class`? feels more descriptive :)

takagi: Renamed!

takagi:
PTAL!

kmaehashi:
Could you merge master and push?

cupy/_core/core.pyx
```diff
@@ -620,7 +620,7 @@ cdef class _ndarray_base:
             newarray.data.copy_from_device_async(x.data, x.nbytes)
         return newarray
 
-    cpdef _ndarray_base view(self, dtype=None):
+    cpdef _ndarray_base view(self, dtype=None, array_class=None):
```
kmaehashi, line 623: This function should accept all of these:

```
a.view()
a.view(some_dtype)
a.view(dtype=some_dtype)
a.view(ndarray_subclass)
a.view(type=ndarray_subclass)  # <---
```

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.view.html#numpy.ndarray.view

I think we will need `**kwargs` here...

takagi: Thanks! I've added another `def view` method around the original `cdef`d one, as it seems that Cython does not take `*args` and `**kwargs` for `cdef`d methods.

.pyx codes that called the original `cdef view` can be also fixed to call `cdef _view_impl` instead of the new `def view`, but I'll make this change in another PR as we have tens of such lines and the next release is approaching.

```
$ find . -name '*.pyx' | xargs grep 'view(' | wc -l
50
```

takagi: I found we can provide a wrapper of `view` method in `ndarray` class, which makes things easier. We can still use `cpdef view` in Cython files with `_ndarray_base` objects. [0b1343f](https://github.com/cupy/cupy/pull/6720/commits/0b1343f31460b7806b93af10bf625f5ec24da88e)

kmaehashi (approved):

PR #6723, opened 2022-05-16
Add `cupy.einsum_path`

leofang:
Part of #6078. 

For the previous discussion about the `einsum_path` refactoring, please see #6261 and #6677.

This PR adds the missing `cupy.einsum_path` function for NumPy compatibility. In order to avoid the infinite `__array_function__` dispatching, I followed the advice from upstream (see https://github.com/numpy/numpy/issues/21379) and created dummy NumPy arrays (of zero size) to reuse `numpy.einsum_path()`. The path computation only cares about the array dimension and shape; actual data, memory layout, or dtype are irrelevant. 

cuTensorNet support will be added in a separate PR.

**Question for the team**: Should I also remove all related code copied from NumPy (specifically, `cupy/linalg/_einsum_opt.py`) as part of this PR, since they are redundant now?

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -467,6 +467,7 @@ from numpy import set_string_function  # NOQA
 # Linear algebra
 # -----------------------------------------------------------------------------
 from cupy.linalg._einsum import einsum  # NOQA
+from cupy.linalg._einsum import einsum_path  # NOQA
 
 from cupy.linalg._product import cross  # NOQA
 from cupy.linalg._product import dot  # NOQA
--- a/cupy/linalg/_einsum.py
+++ b/cupy/linalg/_einsum.py
@@ -4,6 +4,8 @@ import operator
 import string
 import warnings
 
+import numpy
+
 import cupy
 from cupy._core import _accelerator
 from cupy import _util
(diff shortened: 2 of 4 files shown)


leofang:
Interesting, Dask already follows the same approach:
https://github.com/dask/dask/blob/cccb9d8d8e33a891396b1275c2448c352ef40c27/dask/array/einsumfuncs.py#L219

kmaehashi:
> Question for the team: Should I also remove all related code copied from NumPy (specifically, cupy/linalg/_einsum_opt.py) as part of this PR, since they are redundant now?

I think it's safe to remove it, do you have any comments @toslunar?

toslunar:
Calling `numpy.einsum_path` can eliminate the copied code (`cupy/linalg/_einsum_opt.py`).

My concern is about the performance. `cupy.einsum` will have to parse args twice, while `numpy.einsum` parses them only once by using hidden `einsum_call=True` option of `numpy.einsum_path`. In my opinion,
- Keeping the copied code is better than using NumPy's hidden option because the latter may break suddenly.
- Calling public `numpy.einsum_path` seems a tradeoff between cleanness and speed.  It's great if the performance drop is minor.

leofang:
Thanks guys! I'll do some quick checks for the potential perf hit, but for now I'm thinking two ways out:
- Use the hidden `einsum_call` option: It's been there since `einsum_path` was added (2015) and for the same reason (don't parse twice) I don't think NumPy would ever consider removing it. AFAIK the NumPy `einsum` related code has been stable for quite some time.
- Refactor `einsum` such that internally we let NumPy do input verification by delegating to `numpy.einsum_path` (and also get a path back if succeeds): I think this is feasible and has the additional benefit that we will always generate the same errors, but I need to experiment it a bit after I do the perf check.

leofang:
I did some quick checks on the overhead of double parsing and it's a bit large (50-80us). The way I checked is to simply replace [this block](https://github.com/cupy/cupy/blob/a42c458bdfcbcc510dcf4dfc417e0c593304de0c/cupy/linalg/_einsum.py#L640-L665) by `einsum_path`, so it's arguably not the most optimal approach. 

However, what's concerning is before arriving this block, CuPy did several simplifications that mutate the inputs, which makes reusing `einsum_path` very challenging. As a result I'd suggest we defer the refactoring to later, and focus on making this API available. WDYT @toslunar?

toslunar:
> However, what's concerning is before arriving this block, CuPy did several simplifications that mutate the inputs, which makes reusing `einsum_path` very challenging. As a result I'd suggest we defer the refactoring to later, and focus on making this API available. WDYT @toslunar?

I'd like to postpone the refactoring, too.  It looks OK to me to make `cupy.einsum_path` return the same result as `numpy.einsum_path` does.

An issue already CuPy has is that a path computed by `cupy.einsum` is sometimes different than NumPy's.  (For example, the simplifications you mentioned include internal trace, therefore `cupy.einsum` does not reproduce the bug numpy/numpy#20962.)  It's not a big issue, but it might be going to be more visible from users in the sense that
```python
path = cupy.einsum_path(*operands, optimize=True)[0]
cupy.einsum(*operands, optimize=path)
```
is different than
```python
cupy.einsum(*operands, optimize=True)
```
After the PR is merged, I'm willing to make these code equivalent by sharing some implementations between `cupy.einsum` and `cupy.einsum_path`, which is a kind of opposite of eliminating `_einsum_opt.py`.

leofang:
@toslunar Hmmm... good point... I am concerned now, because it means this branch
https://github.com/cupy/cupy/blob/292bb3c80d6631b429bb6adfdf434ec718e4920d/cupy/linalg/_einsum.py#L646-L647
is already technically incorrect (irrelevant of the present PR). The user-provided path must be mutated accordingly before it can be used in the pairwise contraction loop below, isn't it?

toslunar:
Technically yes.  It seems similar to the question "should `cupy.einsum` support genuine `optimize=False` einsum?"

`cupy.einsum`'s actual path is a subdivision of the given path (if we ignore the execution order of unary ops and binary ops). In other words, cupy follows users' intention and may optimize further.  For the `Qij,Qkk->ij` example, the (unique) path `[(0, 1)]` is interpreted like `[(1,), (0, 1)]`.

Drawbacks of this further optimization are
- extra memory, which is as small as the order of inputs, outputs, or intermediate results in the original path
- accuracy or reproducibility, for example
```
>>> for xp in [numpy, cupy]: xp.einsum('i,j->', xp.array([1e308, 1e308]), xp.array([1e-8, 1e-9]))
2.2e+300
array(inf)
```

leofang:
Let me close this one as I don't have time to address the challenges discussed above.

PR #6725, opened 2022-05-17
Fix batched matmul for integral numbers

### LINKED ISSUE #6715: Wrong overflow with matmul of uint16 arrays
### Description

The following arrays give an apparently incorrect overflowed result with matmul. 

```py
>>> import cupy as cp
>>> x = cp.asarray([51307], dtype=cp.uint16)
>>> y = cp.asarray([[[327]]], dtype=cp.uint16)
>>> x
array([51307], dtype=uint16)
>>> y
array([[[327]]], dtype=uint16)
>>> x @ y # This is giving a different answer
array([[172]], dtype=uint16)
>>> x*y
array([[[173]]], dtype=uint16)
>>> x @ y[0]
array([173], dtype=uint16)
>>> x * y[0]
array([[173]], dtype=uint16)
```

The error only occurs when the second array has 3 dimensions. Presumably cupy uses different code paths for stacked vs. non-stacked linear algebra functions. Note that `51307*327 % 2**16` is indeed `173`.

I'm unclear whether this behavior is actually a bug or if this is something that is to be expected from overflowing integers. 

I identified this with the array API test suite, but it is unrelated to the array API module. The array API doesn't specify overflow behavior, but it would be simpler in the test suite if it could just ignore overflowing. 

### To Reproduce

```py
# Write the code here
```

### Installation

### Environment

```
OS                           : Linux-5.13.0-40-generic-x86_64-with-glibc2.34
Python Version               : 3.10.4
CuPy Version                 : 10.4.0
NumPy Version                : 1.22.3
SciPy Version                : 1.8.0
Cython Build Version         : 0.29.28
CUDA Root                    : /home/aaronmeurer/anaconda3/envs/array-api
CUDA Build Version           : 11020
CUDA Driver Version          : 11060
CUDA Runtime Version         : 11060
cuFFT Version                : 10600
cuRAND Version               : 10209
cuSOLVER Version             : (11, 3, 2)
NVRTC Version                : (11, 6)
Thrust Version               : 101000
CUB Build Version            : 101000
Jitify Build Version         : 70f5331
Device 0 Name                : NVIDIA GeForce RTX 2080
```

### Additional Information

_No response_

takagi:
Close #6715.

(UPDATED)This PR fixes batched matmul for integral numbers by implementing custom kernels on top of non-batched one.


emcastillo:
Does this work? there is an specific kernel used for integer matrix multiplication so cublas shouldn't be used.
If this is working there is something wrong in the cublas/kernel selection 😂

takagi:
Yes, but I think the kernel doesn't work as batched one. Implementing another for batched gemm is the right way.

https://github.com/cupy/cupy/blob/8968a3e04cfa84bd6d22f33c63e65b9fbd24d041/cupy/_core/_routines_linalg.pyx#L103

takagi:
Ah, we can compute batched gemm almost with the kernel in parallel by blocks of threads. I'm going to implement it soon. Thanks for your pointing!

takagi:
Seems that the CI error is caused by SciPy 1.8.1. (Can be reproduced locally)

takagi:
Comparison between master and this PR, with int32 matrix batches. I get 10~20%  improvement in efficiency with this PR. BTW, as touched in #6703, I also observed that batched matmul on the same dtype size (float32) matrices runs ~30% faster with cublas routines. Maybe they're well tuned for each Nvidia device.

shape|master[ms]|#6725[ms]|numpy[ms]
---|---|---|---
(16, 256, 256)| 0.15575040131807327| 0.14878720194101336| 183.2779342651367
(32, 256, 256)| 0.24320000112056736| 0.21473280340433124| 453.9682891845704
(64, 256, 256)| 0.43950079977512363| 0.3778559982776642| 716.5655822753906
(128, 256, 256)| 0.8155135989189148| 0.6744063913822175| 2409.152319335937
(256, 256, 256)| 1.5748095870018006| 1.2906496286392215| 4853.44130859375

cupy/_core/_routines_linalg.pyx
```diff
@@ -333,6 +390,51 @@ cdef ndarray _integral_tensordot_core(
...
+        ndarray a, ndarray b, ndarray out, Py_ssize_t m, Py_ssize_t n,
+        Py_ssize_t k, str dtype, Py_ssize_t batch_count):
+
+    config, dim_x, dim_y, blk_m, blk_n = _integral_tensordot_core_config()
+    kern = _tensordot_core_int_kernel_batched(config, dtype)
+    block = (dim_x, dim_y, 1)
+    matPtrA = _mat_ptrs(a)
+    matPtrB = _mat_ptrs(b)
+    matPtrOut = _mat_ptrs(out)
+    max_batch_count = 65000
```
takagi, line 403: This number is borrowed from MAGMA.

emcastillo:
The kernel that we are using could be much faster, there is a PR open trying to tune it, but seems difficult.

emcastillo:
Seems like hip doesn't like this kernel

takagi:
Ah, it's because of the name expressions problem on HIP.

takagi:
Seems that hiprtc doesn't permit calling `__global__` function from another `__global__`.

leofang:
> Seems that hiprtc doesn't permit calling `__global__` function from another `__global__`.

Right, on the CUDA side it's called dynamic parallelism l, which requires separate compilation, but back then it was not supported on HIP https://github.com/cupy/cupy/blob/3ffd5f359246d9afd9fd51be3b8bd191d777ef81/cupy/cuda/compiler.py#L835-L836 not sure about now.

takagi:
> Right, on the CUDA side it's called dynamic parallelism l, which requires separate compilation, but back then it was not supported on HIP

Thanks! Got it. Seems that the situation is still not changed.

takagi:
PTAL!

emcastillo (approved):
LGTM!!

PR #6736, opened 2022-05-23
Fix to return correct CUDA version when in CUDA Python mode

kmaehashi:
This will allow using `runtime.runtimeGetVersion()` in all locations where branch based on the current CUDA version (to test feature availability) is needed.

### DIFF
--- a/tests/cupy_tests/cuda_tests/test_runtime.py
+++ b/tests/cupy_tests/cuda_tests/test_runtime.py
@@ -46,6 +46,8 @@ class TestMemPool:
         runtime.memPoolDestroy(pool)
 
 
+@pytest.mark.skipif(runtime.is_hip,
+                    reason='This assumption is correct only in CUDA')
 def test_assumed_runtime_version():
     # When CUDA Python is enabled, CuPy calculates the CUDA runtime version
     # from NVRTC version. This test ensures that the assumption is correct
--- a/tests/cupy_tests/cuda_tests/test_runtime.py
+++ b/tests/cupy_tests/cuda_tests/test_runtime.py
@@ -4,6 +4,7 @@ import pytest
 
 import cupy
 from cupy.cuda import driver
+from cupy.cuda import nvrtc
 from cupy.cuda import runtime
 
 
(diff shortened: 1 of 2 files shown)


leofang (approved):

emcastillo (approved):

kmaehashi:
This is primarily needed for #6730 so not backporting to avoid breaking existing behavior.

tests/cupy_tests/cuda_tests/test_runtime.py
```diff
@@ -43,3 +44,12 @@ def test_mallocFromPoolAsync(self):
         assert ptr > 0
         runtime.freeAsync(ptr, s.ptr)
         runtime.memPoolDestroy(pool)
+
+
+def test_assumed_runtime_version():
```
emcastillo, line 49: This test fails on hip

emcastillo (changes requested):

tests/cupy_tests/cuda_tests/test_runtime.py
emcastillo:
```
12:50:19     def test_assumed_runtime_version():
12:50:19         # When CUDA Python is enabled, CuPy calculates the CUDA runtime version
12:50:19         # from NVRTC version. This test ensures that the assumption is correct
12:50:19         # by running the same logic in non-CUDA Python environment.
12:50:19         # When this fails, `runtime.runtimeGetVersion()` logic needs to be fixed.
12:50:19         (major, minor) = nvrtc.getVersion()
12:50:19 >       assert runtime.runtimeGetVersion() == major * 1000 + minor * 10
12:50:19 E       assert 3275 == ((9 * 1000) + (0 * 10))
12:50:19 E        +  where 3275 = <built-in function runtimeGetVersion>()
12:50:19 E        +    where <built-in function runtimeGetVersion> = runtime.runtimeGetVersion
12:50:19 
12:50:19 major      = 9
12:50:19 minor      = 0
```

kmaehashi: Ah, right, will fix shortly.

kmaehashi:
Tests passed, could you approve? @emcastillo :)

emcastillo (approved):

PR #6731, opened 2022-05-21
Support testing CUDA 11.6+ in FlexCI

kmaehashi:
CUDA 11.6 issue on GCP seems resolved.
https://cloud.google.com/compute/docs/troubleshooting/known-issues

Testing: https://ci.preferred.jp/cupy.linux.cuda116/101922/ (seems not working, investigating...)

PR #6738, opened 2022-05-23
Fix GPG key issue in FlexCI base image

kmaehashi:
Follows-up #6731


kmaehashi:
Testing at: https://ci.preferred.jp/cupy.linux.cuda116/102104/

emcastillo (approved):

PR #6754, opened 2022-05-26
Fix `cupy.ndarray` document relating to subclass support

takagi:
This PR fixes `cupy.ndarray`'s documentation that #6716 has broken.


takagi:
Let me include this change into #6746 as sphinx seem not to identify `cupy.ndarray` as expected without the renaming to be handled in #6746. It still emits 'alias of...' here.

PR #6765, opened 2022-06-02
[WIP] Add sparse `all_to_all` & `gather_reduce`

emcastillo:


cupyx/distributed/_nccl_comm.py
```diff
@@ -779,4 +794,22 @@ def gather(cls, comm, in_array, out_array, root=0, stream=None):
     @classmethod
     def all_to_all(cls, comm, in_array, out_array, stream=None):
```
asi1024, line 795: How about repeating all_gather simply?

asi1024, line 795: How about repeating all_gather simply?

asi1024 (approved):

PR #6768, opened 2022-06-03
Add CUDA 11.7 on documents

takagi:


kmaehashi:
Skip CI as docs only fix.

PR #6775, opened 2022-06-08
Subclass cupy.ndarray subclassing - Part 3 - New from template (ufunc)

takagi:
Merge #6720 first.

This PR is a part of supporting `cupy.ndarray` subclassing, and support instantiating subclass via ufunc, including two paths: the original one and cutensor one.


kmaehashi (approved):
LGTM!

PR #5648, opened 2021-08-16
Add Dirichlet distribution to Generator

povinsahu1909:
This PR adds Dirichlet distribution to Generator.

Following is the performance comparison on GTX1050.
(The comparison was done only for flattened shape as numpy does not support multi-dimension alpha as input)

size | numpy | cupy 
---|---|---
10000 | 0.001 | 0.001
12000 | 0.002 | 0.001
15000 | 0.002 | 0.001
25000 | 0.002 | 0.001
10000000 | 0.562 | 0.305
12000000 | 0.665 | 0.367
15000000 | 0.830 | 0.459
20000000 | 1.106 | 0.623
24000000 | 1.327 | 0.736
25000000 | 1.381 | 0.764
30000000 | 1.785 | 0.962
36000000 | 2.018 | 1.110
45000000 | 2.639 | 1.439
50000000 | 2.798 | 1.579
75000000 | 6.607 | 2.456

<details>
<summary>script</summary>

```python
import numpy as np
import cupy as cp
import cupyx

n_repeat = 2
n_warmup = 1
crng = cp.random.default_rng()
nrng = np.random.default_rng()

def get_time(perf):
    cpu_time = perf.cpu_times.mean()
    gpu_time = perf.gpu_times.mean()
    return max(cpu_time, gpu_time)

def run(p, q):
    times = []

    perf = cupyx.time.repeat(nrng.dirichlet, (p,),
                            n_repeat=n_repeat, n_warmup=n_warmup)
    times.append(get_time(perf))
    perf = cupyx.time.repeat(crng.dirichlet, (q,),
                            n_repeat=n_repeat, n_warmup=n_warmup)
    times.append(get_time(perf))
    return times


print('size | numpy | cupy ')
print('---|---|---')
nrows = [10000, 12000, 15000, 25000]
ncols = [1, 1000, 2000, 3000]
for m in ncols:
    for n in nrows:
        p = np.random.random( n*m )
        q = cp.asarray(p)
        times = run(p, q)
        print('{} | {:.3f} | {:.3f}'.format(n * m, times[0], times[1]))
```

PR #6776, opened 2022-06-09
[special] Add `cupyx.scipy.special.log_ndtr`

khushi-411:
Hi,
The PR adds `cupyx.scipy.special.log_ndtr`. Thanks! :)

### DIFF
--- a/cupyx/scipy/special/_stats_distributions.py
+++ b/cupyx/scipy/special/_stats_distributions.py
@@ -41,9 +41,9 @@ static __device__ T log_ndtr(T x)
 {
     T t = x * NPY_SQRT1_2;
     if (x < -1.0) {
-        return log(erfcx(-t) / 2) - t * t;
+        return logf(erfcxf(-t) / 2) - t * t;
     } else {
-        return log1p(-erfc(t) / 2);
+        return log1pf(-erfcf(t) / 2);
     }
 }
 
--- a/cupyx/scipy/special/_stats_distributions.py
+++ b/cupyx/scipy/special/_stats_distributions.py
@@ -41,9 +41,9 @@ static __device__ T log_ndtr(T x)
 {
     T t = x * NPY_SQRT1_2;
     if (x < -1.0) {
-        return logf(erfcxf(-t) / 2) - t * t;
+        return log(erfcx(-t) / 2) - t * t;
     } else {
-        return log1pf(-erfcf(t) / 2);
+        return log1p(-erfc(t) / 2);
     }
 }
 
@@ -52,7 +52,7 @@ static __device__ T log_ndtr(T x)
 
 log_ndtr = _core.create_ufunc(
     'cupyx_scipy_special_log_ndtr',
-    (('f->f', 'out0 = log_ndtr(in0)'), 'd->d'),
+    (('f->f', 'out0 = log_ndtrf(in0)'), 'd->d'),
     'out0 = log_ndtr(in0)',
     preamble=log_ndtr_definition,
     doc="""Logarithm of Gaussian cumulative distribution function.
--- a/cupyx/scipy/special/_stats_distributions.py
+++ b/cupyx/scipy/special/_stats_distributions.py
@@ -36,10 +36,9 @@ log_ndtr_definition = """
 
 #define NPY_SQRT1_2   0.707106781186547524400844362104849039  /* 1/sqrt(2) */
 
-template <typename T>
-static __device__ double log_ndtr(T x)
+static __device__ double log_ndtr(double x)
 {
-    T t = x * NPY_SQRT1_2;
+    double t = x * NPY_SQRT1_2;
     if (x < -1.0) {
         return log(erfcx(-t) / 2) - t * t;
     } else {
@@ -47,10 +46,9 @@ static __device__ double log_ndtr(T x)
     }
 }
 
-template <typename T>
-static __device__ float log_ndtrf(T x)
+static __device__ float log_ndtrf(float x)
 {
-    T t = x * NPY_SQRT1_2;
+    float t = x * NPY_SQRT1_2;
     if (x < -1.0) {
         return logf(erfcxf(-t) / 2) - t * t;
     } else {
--- a/cupyx/scipy/special/_stats_distributions.py
+++ b/cupyx/scipy/special/_stats_distributions.py
@@ -37,13 +37,24 @@ log_ndtr_definition = """
 #define NPY_SQRT1_2   0.707106781186547524400844362104849039  /* 1/sqrt(2) */
 
 template <typename T>
-static __device__ T log_ndtr(T x)
+static __device__ double log_ndtr(T x)
 {
     T t = x * NPY_SQRT1_2;
     if (x < -1.0) {
         return log(erfcx(-t) / 2) - t * t;
     } else {
         return log1p(-erfc(t) / 2);
+    }
+}
+
+template <typename T>
+static __device__ float log_ndtrf(T x)
+{
+    T t = x * NPY_SQRT1_2;
+    if (x < -1.0) {
+        return logf(erfcxf(-t) / 2) - t * t;
+    } else {
+        return log1pf(-erfcf(t) / 2);
     }
 }
 
--- a/cupyx/scipy/special/_stats_distributions.py
+++ b/cupyx/scipy/special/_stats_distributions.py
@@ -32,6 +32,53 @@ ndtr = _core.create_ufunc(
     ''')
 
 
+log_ndtr_definition = """
+
+#define NPY_SQRT1_2   0.707106781186547524400844362104849039  /* 1/sqrt(2) */
+
+template <typename T>
+static __device__ T log_ndtr(T x)
+{
+    T t = x * NPY_SQRT1_2;
+    if (x < -1.0) {
+        return log(erfcx(-t) / 2) - t * t;
+    } else {
+        return log1p(-erfc(t) / 2);
+    }
+}
+
+"""
+
+
+log_ndtr = _core.create_ufunc(
+    'cupyx_scipy_special_log_ndtr',
+    (('f->f', 'out0 = log_ndtr(in0)'), 'd->d'),
+    'out0 = log_ndtr(in0)',
+    preamble=log_ndtr_definition,
+    doc="""Logarithm of Gaussian cumulative distribution function.
+
+    Returns the log of the area under the standard Gaussian propability
+    density function.
+
+    Parameters
+    ----------
+    x : array-like
+        The input array
+
+    Returns
+    -------
+    y : cupy.ndarray
+        The value of the log of the normal cumulative distribution
+        function evaluated at x
+
+    See Also
+    --------
+    :func:`scipy.special.log_ndtr`
+
+    """,
+)
+
+
 ndtri = _core.create_ufunc(
     'cupyx_scipy_special_ndtri',
     (('f->f', 'out0 = normcdfinvf(in0)'), 'd->d'),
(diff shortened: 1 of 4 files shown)


cupyx/scipy/special/_stats_distributions.py
```diff
@@ -32,6 +32,53 @@
...
+template <typename T>
+static __device__ T log_ndtr(T x)
+{
+    T t = x * NPY_SQRT1_2;
+    if (x < -1.0) {
+        return log(erfcx(-t) / 2) - t * t;
+    } else {
+        return log1p(-erfc(t) / 2);
+    }
+}
```
khushi-411, line 48: We can change this algorithm to `log(ndtr)`. But I see scipy has used this algorithm, due to numerical precision issues. Ref: https://github.com/scipy/scipy/pull/15187. Is this okay? Thanks

khushi-411, line 48: We can change this algorithm to `log(ndtr)`. But I see scipy has used this algorithm, due to numerical precision issues. Ref: https://github.com/scipy/scipy/pull/15187. Is this okay? Thanks

tests/cupyx_tests/scipy_tests/special_tests/test_statistics.py
```diff
@@ -11,6 +11,9 @@ class _TestBase:
     def test_ndtr(self):
         self.check_unary_linspace0_1('ndtr')
 
+    def test_log_ndtr(self):
+        self.check_unary_linspace0_2('log_ndtr')
```
takagi, line 15: Maybe we can use `check_unary_linspace0_1`. (Looks like that `0_1` means the range of linspace from zero to one)

cupyx/scipy/special/_stats_distributions.py
takagi: Using this formula is okay! BTW, would you also implement single precision version using `logf`, `log1pf`, `erfcf` and `erfcxf`?

tests/cupyx_tests/scipy_tests/special_tests/test_statistics.py
khushi-411: Exactly, right! Have changed it. Thanks

takagi:
Thanks! Almost done except for a few comments!

takagi:
Would you make a performance comparison? SciPy vs. CuPy across some sizes would be nice. You can follow the scripts you can see in #4316 and #5648.

khushi-411:
Hi, here's the benchmark:

size | dtype | SciPy | CuPy
-----|-------|-------|-----
1000 | int8 | 0.026 ms | 0.018 ms
1000 | int32 | 0.028 ms | 0.020 ms
1000 | uint32 | 0.025 ms | 0.020 ms
1000 | float32 | 0.025 ms | 0.018 ms
1000 | float64 | 0.024 ms | 0.020 ms
100000 | int8 | 2.263 ms | 0.070 ms
100000 | int32 | 2.060 ms | 0.338 ms
100000 | uint32 | 2.047 ms | 0.339 ms
100000 | float32 | 2.156 ms | 0.044 ms
100000 | float64 | 2.019 ms | 0.341 ms
1000000 | int8 | 21.381 ms | 0.120 ms
1000000 | int32 | 20.709 ms | 3.293 ms
1000000 | uint32 | 20.399 ms | 3.262 ms
1000000 | float32 | 21.267 ms | 0.120 ms
1000000 | float64 | 19.619 ms | 3.265 ms

<details>
<summary>Script: Click Please</summary>

<p>

```python
import cupy
import cupyx

import scipy

from cupy import testing


n_warmup = 1
n_repeat = 10
dtype_set = ('int8', 'int32', 'uint32', 'float32', 'float64')


def get_time(pref):


def time_log_ndtr(a):
    cp_a = cupy.array(a)
    ref = scipy.special.log_ndtr(a)
    ret = cupyx.scipy.special.log_ndtr(cp_a)
    times = []
    perf = benchmark(scipy.special.log_ndtr, (a,),
    times.append(get_time(perf))
    perf = benchmark(cupyx.scipy.special.log_ndtr, (cp_a,),
    times.append(get_time(perf))
    return times


for n in n_set:
    for dtype in dtype_set:
        a = scipy.random.randint(0, 2, size=n).astype(dtype)
        times = time_log_ndtr(a)
```

</p>
</details>

Thanks! :)

cupyx/scipy/special/_stats_distributions.py
khushi-411: Have updated!

But I believe we will need to use previous test cases, due to tolerance issues. Will that be good? Also, any reason you preferred the `single precision version`?

Thanks!

takagi: > But I believe we will need to use previous test cases, due to tolerance issues. Will that be good?

Seems that `check_unary_linspace0_1`'s atol 'default' value is too strict. Would you try changing to 1e-6? (It should be also safe for other tests that use the value)

https://github.com/cupy/cupy/blob/462d14e6ebcd4c8b2f89e12b65a92c217ebbb12f/tests/cupyx_tests/scipy_tests/special_tests/test_statistics.py#L31

khushi-411: Fixed

takagi: > EDIT: Ooo, I see. Because of performance reasons!

Yes. Single-precision operations are generally faster on GPU. Can I ask you to implement both single-precision version and double-precision version, instead of providing single-precision version only? We want to use double-precision version when precision matters.

I explain some of `create_ufunc`'s parameters.

```py
log_ndtr = _core.create_ufunc(
    'cupyx_scipy_special_log_ndtr',
    (('f->f', 'out0 = log_ndtr(in0)'), 'd->d'),
    'out0 = log_ndtr(in0)',
    ...
```

The second specifies a list of input and output dtypes supported by the create ufunc. 'f->f' means that it takes a float32 (single-precision) input, then returns a float32 output. 'd->d' means that it takes a float64 (double-precision) input then returns a float64 output.

The third supplies the "default" operation performed by this ufunc. Why I say "default" is that please see the first element of the second parameter `('f->f', 'out0 = log_ndtr(in0)')`.  Here provided a tuple of strings instead of 'f->f' only. This means that, when the input's dtype if float32, the second element of the tuple `'out0 = log_ndtr(in0)'` is used instead of the default operation. It happens to be the same as the default one in your code, but you can give a single-precision version, say, `out0 = log_ndtrf(in0)` here.

khushi-411: Interesting! Thanks, for the detailed description!

tests/cupyx_tests/scipy_tests/special_tests/test_statistics.py
```diff
@@ -56,7 +59,7 @@ def check_unary_lower_precision(self, name, xp, scp, dtype):
         return self._check_unary(a, name, scp)
 
     @testing.for_dtypes(['e', 'f', 'd'])
-    @testing.numpy_cupy_allclose(atol=atol, rtol=rtol, scipy_name='scp')
+    @testing.numpy_cupy_allclose(atol=1e-6, rtol=1e-6, scipy_name='scp')
```
takagi, line 62: This change unexpectedly relaxes the tolerance for float64. Would you fix the definition of `atol` variable as follows?
```py
atol = {'default': 1e-6, cupy.float64: 1e-14}
```

khushi-411: I think we won't need this. I rechecked this, it worked with the default atol tolerance = `1e-14`.

takagi: Okay!

cupyx/scipy/special/_stats_distributions.py
```diff
@@ -52,7 +52,7 @@
 
 log_ndtr = _core.create_ufunc(
     'cupyx_scipy_special_log_ndtr',
-    (('f->f', 'out0 = log_ndtr(in0)'), 'd->d'),
+    (('f->f', 'out0 = log_ndtrf(in0)'), 'd->d'),
```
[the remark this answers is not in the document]
khushi-411: Done. Thanks!

cupyx/scipy/special/_stats_distributions.py
```diff
@@ -37,7 +37,7 @@
 #define NPY_SQRT1_2   0.707106781186547524400844362104849039  /* 1/sqrt(2) */
 
 template <typename T>
```
takagi, line 39: Now you can eliminate this template type.

cupyx/scipy/special/_stats_distributions.py
```diff
@@ -47,6 +47,17 @@
     }
 }
 
+template <typename T>
+static __device__ float log_ndtrf(T x)
```
takagi, line 51: Also `T` to `float`.

cupyx/scipy/special/_stats_distributions.py
khushi-411: Fixed

takagi:
Thanks!

khushi-411:
Thanks for your guidance!! 😊

takagi:
CI passes, thanks! Would you make a performance comparison again with the latest commit? Then, I'll merge this into the master.

khushi-411:
> Would you make a performance comparison again with the latest commit?

Sure, here we go ...

size | dtype | SciPy | CuPy
-----|-------|-------|-----
1000 | int8 | 0.026 ms | 0.018 ms
1000 | int32 | 0.026 ms | 0.020 ms
1000 | uint32 | 0.025 ms | 0.020 ms
1000 | float32 | 0.025 ms | 0.017 ms
1000 | float64 | 0.024 ms | 0.020 ms
100000 | int8 | 2.159 ms | 0.061 ms
100000 | int32 | 2.038 ms | 0.339 ms
100000 | uint32 | 2.059 ms | 0.338 ms
100000 | float32 | 2.169 ms | 0.024 ms
100000 | float64 | 2.005 ms | 0.339 ms
1000000 | int8 | 21.383 ms | 0.118 ms
1000000 | int32 | 20.215 ms | 3.258 ms
1000000 | uint32 | 20.402 ms | 3.258 ms
1000000 | float32 | 21.290 ms | 0.119 ms
1000000 | float64 | 19.824 ms | 3.261 ms

<details>
<summary>Script: Click Please</summary>

<p>

```python
import cupy
import cupyx

import scipy

from cupy import testing


n_warmup = 1
n_repeat = 10
dtype_set = ('int8', 'int32', 'uint32', 'float32', 'float64')


def get_time(pref):


def time_log_ndtr(a):
    cp_a = cupy.array(a)
    ref = scipy.special.log_ndtr(a)
    ret = cupyx.scipy.special.log_ndtr(cp_a)
    times = []
    perf = benchmark(scipy.special.log_ndtr, (a,),
    times.append(get_time(perf))
    perf = benchmark(cupyx.scipy.special.log_ndtr, (cp_a,),
    times.append(get_time(perf))
    return times


for n in n_set:
    for dtype in dtype_set:
        a = scipy.random.randint(0, 2, size=n).astype(dtype)
        times = time_log_ndtr(a)
```

</p>
</details>

Thanks! :)

takagi:
Great! 🎉  From the result of size = 1000000 where float32 is ~30 times faster than float64, I guess you're using a RTX or GTX board that has one FP64 unit for 32 FP32 units. GPUs for high-performance computing have more FP64 units (for example, one FP64 unit for two FP32 units in A100), so the difference should be much smaller.

khushi-411:
> Great! :tada:
> 
> From the result of size = 1000000 where float32 is ~30 times faster than float64, I guess you're using a RTX or GTX board that has one FP64 unit for 32 FP32 units. GPUs for high-performance computing have more FP64 units (for example, one FP64 unit for two FP32 units in A100), so the difference should be much smaller.

Yep, using GTX board! This is interesting! Wasn't knowing about it. Will take a note :) Thanks!

takagi (approved):
LGTM, thanks!

PR #6782, opened 2022-06-14
Fix docs to hide `ndarray_base`

kmaehashi:
Depends on ~#6746~ (merged)

* Declare `__init__` signature in docstring instead of modifying it in `conf.py`
* Replace `_ndarray_base` that appears in argument types of several functions


takagi (approved):
LGTM! Would you merge this as it contains only fix for docs and I've checked on RTD. It's okay. @kmaehashi

PR #6791, opened 2022-06-16
Make the warning for cupy.array_api say "cupy" instead of "numpy"

asmeurer:
There are a lot of other references to "numpy" in the module docstrings, which I did not change, but this one is confusing because it prints to the screen when `cupy.array_api` is imported.


leofang (approved):
I think we kept the "numpy" references around to make it easier to apply patches from NumPy, but I think this PR should be fine. We eventually need to deviate from NumPy when things get more stable.

kmaehashi (approved):

PR #6800, opened 2022-06-22
Add support for `cupy-cuda11x` wheel

kmaehashi:


leofang:
Question: Would this new wheel work with `cupy-wheel`?

emcastillo (approved):
LGTM!

PR #6813, opened 2022-06-27
[fix] support multiple output args in reduction kernels

khushi-411:
#### TODO
- [ ] Add tests

### DIFF
--- a/cupy/_core/_reduction.pyx
+++ b/cupy/_core/_reduction.pyx
@@ -59,7 +59,7 @@ cpdef function.Function _create_reduction_function(
 ${type_preamble}
 ${preamble}
 #define REDUCE(a, b) (${reduce_expr})
-#define POST_MAP(a) (${post_map_expr})
+#define POST_MAP(a, b) (${post_map_expr})
 #define _REDUCE(_offset) if (_tid < _offset) { \
   _type_reduce _a = _sdata[_tid], _b = _sdata[(_tid + _offset)]; \
   _sdata[_tid] = REDUCE(_a, _b); \
@@ -103,9 +103,13 @@ extern "C" __global__ void ${name}(${params}) {
       _s = _sdata[_tid];
     }
     if (_tid < _block_stride && _i < _out_ind.size()) {
-      _out_ind.set(static_cast<ptrdiff_t>(_i));
-      ${output_expr}
-      POST_MAP(_s);
+      for (ptrdiff_t _j = _i + _j_offset; _j < _out_ind.size();
+           _j += _j_stride, _J += _J_stride) {
+        _out_ind.set(_j);
+        ${output_expr}
+        _type_reduce _a = static_cast<_type_reduce>(${post_map_expr});
+        POST_MAP(_s, _a);
+      }
     }
   }
 }''').substitute(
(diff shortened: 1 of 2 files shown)


cupy/_core/_reduction.pyx
```diff
@@ -59,7 +59,7 @@ cpdef function.Function _create_reduction_function(
 ${type_preamble}
 ${preamble}
 #define REDUCE(a, b) (${reduce_expr})
-#define POST_MAP(a) (${post_map_expr})
+#define POST_MAP(a, b) (${post_map_expr})
```
takagi, line 62: You can leave POST_MAP macro's argument to be only `a` (that is, `POST_MAP(a)`) as it is the result of the reduction phase. Instead, let's modify the expansion of the macro `(${post_map_expr})` to take statements instead of an expression. In C grammar, a parenthesis `()` can only contain an expression. For multi-line macro, we usually use `do { ... } while (0)` loop;

cupy/_core/_reduction.pyx
```diff
@@ -737,16 +741,13 @@ cdef class ReductionKernel(_AbstractReductionKernel):
...
-            if self.nout != 1:
-                raise NotImplementedError('')
             if len(out_args) != 0:
                 raise ValueError("cannot specify 'out' as both "
                                  "a positional and keyword argument")
-            out_args = [out]
 
         dev_id = device.get_device_id()
         in_args = _preprocess_args(dev_id, args[:self.nin], False)
-        out_args = _preprocess_args(dev_id, out_args, False)
```
takagi, line 749: Let's keep `out_args` as `__call__` method also can take output arrays in positional arguments `args`. Instead, please assign `out` parameter to `out_args` with properly adjusted above.

cupy/_core/_reduction.pyx
khushi-411: Hi, @takagi! Thanks for the comments! I'll think from that perspective and update the PR :)

tests/cupy_tests/core_tests/test_reduction.py
```diff
@@ -189,6 +189,47 @@ def test_invalid_kernel_name(self):
...
+        else:
+            return a.sum(axis=axis, keepdims=keepdims, dtype='b')
+
+
+class ReductionKernelMultipleArgumentsTestBase(
+        AbstractReductionTestBaseMultipleArguments):
+
+    def get_sum_func(self):
+        return cupy.ReductionKernel(
+            'T x', 'T out', 'x', 'a + b', 'out = a, b', '0', 'my_sum')
```
takagi, line 220: This time I think of making ReductionKernel able to handle multiple values only in the post-reduction phase (Is it also okay for you?). So would you try to make the following ReductionKernel pass the test? The ReductionKernel  takes two arrays for output, `out0` and `out1`, then assigning them to sum of `x` as well as its negated sum. The negated sum does not have much sense, just testing purpose.

```
return cupy.ReductionKernel(
            'T x', 'T out0, T out1', 'x', 'a + b', 'out0 = a; out1 = -a', '0', 'my_sum_neg_sum')
```

takagi:
Let me close this PR to keep the PR list clearer. Feel free to reopen again, thanks!

PR #6814, opened 2022-06-28
Update tags for FlexCI projects

kmaehashi:
* Add `cuda102` etc. tags to trigger tests based on CUDA versions
* Use CUDA 11.7 (Linux) instead of CUDA 11.5 (Linux) for `mini` tag
* Fix CUDA 11.7 (Windows) missing


kmaehashi:
The new tags won't be effective unless merged 😓

Issue #6688, opened 2022-04-27
Feedback for Universal Wheel Package (`cupy-wheel`)


kmaehashi:
CuPy v10.4.0 (#6012) added an experimental meta package `cupy-wheel` on PyPI. Instead of specifying an individual wheel package name, users can now install CuPy with the single command line `pip install cupy-wheel`; the `cupy-wheel` meta-package probes the CUDA version installed in the users' environment, and automatically picks and installs the real CuPy wheel package (e.g., `cupy-cuda115`).

It is currently marked experimental. Towards the v11 final release, we are going to stabilize this feature.

Major TODOs remaining are:
- [ ] Add documentation for downstream packages to use `cupy-wheel`
- [x] Disable caching `cupy-wheel` package by `pip` to enforce running `setup.py` everytime
- [ ] Add `long_description`
- [ ] Automate the release process of the meta package

If you find difficulties using `cupy-wheel`, or have any comments or suggestions on `cupy-wheel` package, please feel free to post them here.

kmaehashi:
### GPUs are required to install `cupy-wheel`

I was trying `cupy-wheel` with several scenarios and now thinking this restriction is not negligible:
https://github.com/cupy/cupy/pull/6012#discussion_r843109969

> This relies on the installing machine to have a GPU.

The most problematic case is that CUDA driver is unavailable during `docker build`. This means `cupy-wheel` (and dependents) will not be able to be burnt in container images.

```dockerfile
FROM nvidia/cuda:11.2.2-cudnn8-devel-ubuntu20.04
RUN apt-get update && apt-get -y install python3-pip
RUN pip install cupy-wheel
```

```py
    __main__.AutoDetectionFailed: libcudart.so: <_FuncPtr object at 0x7f2b0dd3e1c0> returned error: 35
```

I come up with several alternative ways to probe CUDA version.

1. Extract from shared library name (e.g. `libcudart.so.11.6.55` -> 11.6).
    * Cons: Does not work on Windows (all CUDA 11.x runtime DLLs are named `cudart64_110.dll`)
2. Extract from development tools (`CUDA_VERSION` in `cuda.h`,  `nvcc --version`, `nvrtcGetVersion`, etc.)
    * Cons: Does not work in the environment without dev headers/tools (e.g., "base" CUDA images like `nvidia/cuda:11.6.2-base-ubuntu20.04`).
3. Extract from `$CUDA_PATH/version.json`
[...]

PR #6819, opened 2022-06-28
cupy-wheel: Use NVRTC to infer the toolkit version

kmaehashi:
Ref #6688

This PR fixes this issue described in the above PR.

> GPUs are required to install cupy-wheel

This PR also fixes the issue that ROCm 4.0/4.2 detection was not working.


leofang:
This is clever!

emcastillo (approved):

PR #6827, opened 2022-06-29
CI: allow Jenkins to be skipped

kmaehashi:


kmaehashi:
Ah, I think this isn't working...

PR #6834, opened 2022-06-30
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/auto-cc.yml
+++ b/.github/workflows/auto-cc.yml
@@ -6,8 +6,14 @@ on:
   pull_request_target:
     types: [labeled]
 
+permissions:
+  contents: read
+
 jobs:
   auto-notify:
+    permissions:
+      issues: write  # for peter-evans/create-or-update-comment to create or update comment
+      pull-requests: write  # for peter-evans/create-or-update-comment to create or update comment
     if: |
         github.repository_owner == 'cupy' &&
         contains('|hip|array-api|', github.event.label.name)
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -6,6 +6,9 @@ on:
     branches:
        - master
 
+permissions:
+  contents: read
+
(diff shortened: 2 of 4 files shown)


kmaehashi:
Thanks! Could you fix `flexci.yml` as well, so that we can change the default token permission to read-only in repository settings? The token is used only to write a status to a commit.

https://github.com/cupy/cupy/blob/1c33cb8c56f1457ddb9001e9e8c75d5714100c0d/.github/workflows/flexci_dispatcher.py#L63-L81

naveensrinivasan:
> Thanks! Could you fix `flexci.yml` as well, so that we can change the default token permission to read-only in repository settings? The token is used only to write a status to a commit.
> 
> https://github.com/cupy/cupy/blob/1c33cb8c56f1457ddb9001e9e8c75d5714100c0d/.github/workflows/flexci_dispatcher.py#L63-L81

Can that be another PR, please?

kmaehashi:
> Can that be another PR, please?

I don't see the merit of merging this one unless the default token permission cannot be changed to readonly. Please update the PR when ready!

kmaehashi:
Closing as no activity for a while but feel free to reopen when ready for another review!

PR #6844, opened 2022-06-30
Fix augassign target is evaluated twice in JIT

asi1024:
Reproducer:
```py
import cupy
import cupyx

@cupyx.jit.rawkernel(device=True)
def g(x):
    x[0] += 1
    return 1

@cupyx.jit.rawkernel()
def f(x):
    x[g(x)] += 1

x = cupy.zeros((2,))

f[1, 1](x)

print(x)
print(f.cached_code)
```

master:
```cc
[2. 1.]

__device__ int g_1(CArray<double, 1, true, true> x) {
  x[0] = (x[0] + (double)(1));
  return 1;
}
extern "C" __global__ void f(CArray<double, 1, true, true> x) {
  x[g_1(x)] = (x[g_1(x)] + (double)(1));
}
```

This PR:
```cc
[1. 1.]

__device__ int g_1(CArray<double, 1, true, true> x) {
  { double &tmp__ = x[0]; tmp__ = (tmp__ + (double)(1)); }
  return 1;
}
extern "C" __global__ void f(CArray<double, 1, true, true> x) {
  { double &tmp__ = x[g_1(x)]; tmp__ = (tmp__ + (double)(1)); }
}
```

### DIFF
--- a/cupyx/jit/_compile.py
+++ b/cupyx/jit/_compile.py
@@ -254,7 +254,11 @@ class Environment:
 
     def get_fresh_variable_name(self, prefix='', suffix=''):
         self.count += 1
-        return f'{prefix}{self.count}{suffix}'
+        name = f'{prefix}{self.count}{suffix}'
+        if self[name] is None:
+            return name
+        else:
+            return self.get_fresh_variable_name(prefix, suffix)
 
 
 def _transpile_function(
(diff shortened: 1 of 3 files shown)


kmaehashi (dismissed):
LGTM!

cupyx/jit/_compile.py
```diff
@@ -464,13 +464,18 @@ def _transpile_stmt(stmt, is_toplevel, env):
     if isinstance(stmt, ast.AugAssign):
         value = _transpile_expr(stmt.value, env)
         target = _transpile_expr(stmt.target, env)
-        assert isinstance(target, Data)
+        if not isinstance(target, Data):
+            raise TypeError(f'Cannot augassign to {target.code}')
         value = Data.init(value, env)
-        result = _eval_operand(stmt.op, (target, value), env)
+        tmp = Data('tmp__', target.ctype)
```
kmaehashi, line 470: Just curious: Are there any other variable names that are reserved by CuPy JIT?

asi1024: The users might use `tmp__` as a local variable in the python function. Fixed in [1c229fa](https://github.com/cupy/cupy/pull/6844/commits/1c229fa0e9e2f63b39376594d80ce8b71dafcb24).

kmaehashi: Looks nice!

Issue #6821, opened 2022-06-28
Follow NumPy 1.23's API changes


takagi:
https://github.com/cupy/cupy/issues/6764#issuecomment-1167538198

takagi:
I'll make this done with the highest priority.

takagi:
I checked that CuPy happens to work as NumPy 1.23 does:

```py
>>> import cupy as cp
>>> f32 = cp.float32([1, 2])
>>> cp.linalg.norm(f32, 2).dtype
dtype('float32')
>>> f16 = cp.float16([1, 2])
>>> cp.linalg.norm(f16, 2).dtype
dtype('float16')
```

https://numpy.org/devdocs/release/1.23.0-notes.html#d-np-linalg-norm-preserves-float-input-types-even-for-scalar-results

takagi:
Close as done.

PR #6848, opened 2022-07-01
Relaxed C-contiguous requirement for changing dtype of different size

takagi:
Part of #6821, following a change of NumPy 1.23.

https://numpy.org/devdocs/release/1.23.0-notes.html#changing-to-dtype-of-a-different-size-now-requires-contiguity-of-only-the-last-axis

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -679,12 +679,14 @@ cdef class _ndarray_base:
             raise ValueError(
                 'Changing the dtype of a 0d array is only supported if '
                 'the itemsize is unchanged')
-        if self._c_contiguous:
-            axis = ndim - 1
-        else:
+        axis = ndim - 1
+        if (
+            self._shape[axis] != 1
+            and self._strides[axis] != self.dtype.itemsize
+        ):
             raise ValueError(
-                'To change to a dtype of a different size, the array must '
-                'be C-contiguous')
+                'To change to a dtype of a different size, the last axis '
+                'must be contiguous')
 
         # Normalize `_strides[axis]` whenever itemsize changes
         v._strides[axis] = v_is
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -682,6 +682,7 @@ cdef class _ndarray_base:
         axis = ndim - 1
         if (
             self._shape[axis] != 1
+            and self.size != 0
             and self._strides[axis] != self.dtype.itemsize
         ):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


cupy/_core/core.pyx
```diff
@@ -679,12 +679,14 @@ cdef class _ndarray_base:
             raise ValueError(
                 'Changing the dtype of a 0d array is only supported if '
                 'the itemsize is unchanged')
-        if self._c_contiguous:
-            axis = ndim - 1
-        else:
+        axis = ndim - 1
+        if (
+            self._shape[axis] != 1
```
toslunar, line 684: The condition to ignore `strides[axis]` is `shape[axis] == 1 or size == 0`.
```suggestion
            self._shape[axis] != 1
            and self.size != 0
```

```
>>> np.ones((3, 20), np.int16)[:0, ::2].view(np.int32)
array([], shape=(0, 5), dtype=int32)
```

tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
```diff
@@ -335,6 +335,37 @@ def test_copy_host_to_device_view(self):
...
 
+class TestViewDtype:
+
+    @testing.numpy_cupy_array_equal()
+    def test_smaller_dtype_multiple(self, xp):
+        # x is non-contiguous
+        x = xp.arange(10, dtype=xp.int32)[::2]
+        with pytest.raises(ValueError):
+            x.view(xp.int16)
+        return x[:, xp.newaxis].view(xp.int16)
```
toslunar, line 346: The `shape[-1] == 1 and stride[-1] == 0 != itemsize` case is covered by this testcase.

[Optional] Add another testcase to cover other `shape[-1] == 1 and stride[-1] != itemsize case`, for example `xp.ones((3, 4), xp.int32)[:, :1:2].view(xp.int16)`.

tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
```diff
@@ -335,6 +335,37 @@ def test_copy_host_to_device_view(self):
             dev[:] = host
 
 
+class TestViewDtype:
```
toslunar, line 338: It might be a good chance to separate the `view` tests.

Could you rearrange tests in `TestArrayCopyAndView` and this `TestViewDtype` into new classes `TestArrayCopy` and `TestView`?

toslunar (changes requested):

tests/cupy_tests/core_tests/test_ndarray_copy_and_view.py
takagi: Sure!

cupy/_core/core.pyx
takagi: Thanks! Also added another test to check this.

takagi:
PTAL!

toslunar:
The new tests may need `testing.requires('numpy>=1.23')`.

/test mini

PR #6853, opened 2022-07-04
Support `equal_nan` parameter for `unique`

takagi:
Part of https://github.com/cupy/cupy/issues/6821, following a change of NumPy 1.23.

https://numpy.org/devdocs/release/1.23.0-notes.html#new-parameter-equal-nan-added-to-np-unique

Note that `numpy.unique` seems to have a bug that it returns wrong results when it takes float16 ndarrays with one or more `nan`.

### DIFF
--- a/cupy/_manipulation/add_remove.py
+++ b/cupy/_manipulation/add_remove.py
@@ -123,8 +123,17 @@ def trim_zeros(filt, trim='fb'):
     return filt[start:end]
 
 
+@_core.fusion.fuse()
+def _unique_update_mask_equal_nan(mask, x0, x1):
+    x0_isnan = cupy.isnan(x0)
+    x1_isnan = cupy.isnan(x1)
+    mask1 = cupy.logical_not(
+        cupy.logical_and(x0_isnan, x1_isnan))  # NAND across isnan
+    mask[:] = cupy.logical_and(mask, mask1)
+
+
 def unique(ar, return_index=False, return_inverse=False,
-           return_counts=False, axis=None):
+           return_counts=False, axis=None, *, equal_nan=True):
     """Find the unique elements of an array.
 
     Returns the sorted unique elements of an array. There are three optional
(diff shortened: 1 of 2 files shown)


cupy/_manipulation/add_remove.py
```diff
@@ -181,6 +192,8 @@ def unique(ar, return_index=False, return_inverse=False,
     mask = cupy.empty(aux.shape, dtype=cupy.bool_)
     mask[:1] = True
     mask[1:] = aux[1:] != aux[:-1]
+    if equal_nan:
+        _unique_update_mask_equal_nan(mask[1:], aux[1:], aux[:-1])
```
toslunar, line 196: Since nans are always after non-nan values, we don't have to use `aux[1:]`

https://numpy.org/doc/stable/reference/generated/numpy.sort.html
> In numpy versions >= 1.4.0 nan values are sorted to the end. The extended sort order is:
> > Real: [R, nan]
> > Complex: [R + Rj, R + nanj, nan + Rj, nan + nanj]

tests/cupy_tests/manipulation_tests/test_add_remove.py
```diff
@@ -148,6 +147,33 @@ def test_unique_empty_return_all(self, xp, dtype):
         return xp.unique(
             a, return_index=True, return_inverse=True, return_counts=True)
 
+    @pytest.mark.parametrize('equal_nan', [True, False])
+    @pytest.mark.parametrize('dtype', 'fdFD')
+    @testing.numpy_cupy_array_equal()
+    @testing.with_requires('numpy>=1.23')
+    def test_unique_equal_nan(self, xp, dtype, equal_nan):
+        if xp.dtype(dtype).kind == 'c':
+            # Nan and Nan+Nan*1j are collapsed when equal_nan=True
```
toslunar, line 156: We may test which nan is selected.
```
>>> np.unique([complex(np.nan, 3), complex(7, np.nan), complex(np.nan, np.nan)])
array([7.+nanj])
```

toslunar (commented):
While NumPy 1.23 just added the flexibility to set `equal_nan=False`, this PR also fixes CuPy's default behavior from `equal_nan=False` to `equal_nan=True`.

cupy/_manipulation/add_remove.py
takagi: Genius!

takagi:
https://github.com/numpy/numpy/issues/21838

toslunar (approved):
LGTM

leofang:
I can't tell what's wrong with Jenkins here, but it'd be nice to get this PR merged so that I can fire another array-api change 🙂

PR #6855, opened 2022-07-06
[stats] add `cupyx.scipy.stats.{zmap, zscore}`

khushi-411:
The PR adds `cupyx.scipy.stats.zmap` API.

#### Benchmarks
size | dtype | SciPy | CuPy
-----|-------|-------|-----
1000 | int8 | 0.151 ms | 0.835 ms
1000 | int32 | 0.105 ms | 0.710 ms
1000 | float32 | 0.096 ms | 0.743 ms
1000 | float64 | 0.098 ms | 0.682 ms
1000 | complex64 | 0.117 ms | 0.706 ms
1000 | complex128 | 0.116 ms | 0.692 ms
100000 | int8 | 0.639 ms | 0.827 ms
100000 | int32 | 0.600 ms | 0.793 ms
100000 | float32 | 0.367 ms | 0.702 ms
100000 | float64 | 0.471 ms | 0.745 ms
100000 | complex64 | 1.035 ms | 0.774 ms
100000 | complex128 | 1.102 ms | 1.113 ms
1000000 | int8 | 5.349 ms | 2.173 ms
1000000 | int32 | 5.476 ms | 2.312 ms
1000000 | float32 | 2.903 ms | 1.407 ms
1000000 | float64 | 5.100 ms | 1.755 ms
1000000 | complex64 | 9.877 ms | 1.678 ms
1000000 | complex128 | 15.183 ms | 6.931 ms

#### Updated benchmarks (after removing apply_along_axis)
size | dtype | SciPy | CuPy
-----|-------|-------|-----
1000 | int8 | 0.107 ms | 0.693 ms
1000 | int32 | 0.106 ms | 0.749 ms
1000 | float32 | 0.101 ms | 0.820 ms
1000 | float64 | 0.104 ms | 0.686 ms
1000 | complex64 | 0.117 ms | 0.831 ms
1000 | complex128 | 0.122 ms | 0.699 ms
100000 | int8 | 0.702 ms | 0.806 ms
100000 | int32 | 0.652 ms | 0.867 ms
100000 | float32 | 0.390 ms | 0.819 ms
100000 | float64 | 0.522 ms | 0.879 ms
100000 | complex64 | 1.110 ms | 0.972 ms
100000 | complex128 | 1.173 ms | 1.345 ms
1000000 | int8 | 5.794 ms | 2.139 ms
1000000 | int32 | 5.914 ms | 2.194 ms
1000000 | float32 | 3.098 ms | 1.387 ms
1000000 | float64 | 5.441 ms | 1.685 ms
1000000 | complex64 | 11.030 ms | 1.661 ms
1000000 | complex128 | 15.822 ms | 6.772 ms

<details>
<summary>Script (Click Please)</summary>

<p>

```python
import cupy
import cupyx

import scipy

from cupy import testing

import cupyx.scipy.stats

import scipy.stats


n_warmup = 1
n_repeat = 10
dtype_set = ('int8', 'int32', 'float32', 'float64', 'complex64', 'complex128')


def get_time(pref):


def time(a, b):
    cp_a = cupy.array(a)
    cp_b = cupy.array(b)
    ref = scipy.stats.zmap(a, b)
    ret = cupyx.scipy.stats.zmap(cp_a, cp_b)
    times = []
    perf = benchmark(scipy.stats.zmap, (a, b),
    times.append(get_time(perf))
    perf = benchmark(cupyx.scipy.stats.zmap, (cp_a, cp_b),
    times.append(get_time(perf))
    return times


for n in n_set:
    for dtype in dtype_set:
        a = scipy.random.randint(0, 2, size=n).astype(dtype)
        b = scipy.random.randint(0, 10, size=n).astype(dtype)
        times = time(a, b)
```

</p>
</details>


tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy
+
+import cupy
+
+from cupy import testing
+
+import cupyx
+import cupyx.scipy.stats  # NOQA
+
+import scipy.stats  # NOQA
```
takagi, line 12: nit:
```suggestion
import numpy
import pytest

import cupy
from cupy import testing
import cupyx
import cupyx.scipy.stats  # NOQA

import scipy.stats  # NOQA
```

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@testing.with_requires('scipy')
+class TestZmap:
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_zmap_1dim(self, xp, scp, dtype):
+        x = testing.shaped_random((5,), xp, dtype=dtype)
+        y = testing.shaped_random((4,), xp, dtype=dtype)
```
takagi, line 26: Would you add tests where `compare` arrays have the same dimension with `scores` as scipy's assumes as such?

It also seems true that scipy's accept `compare` with the different dimension without seeing any exceptions or warnings. So I think keeping the current tests is okay. (only multi-dim would be enough)

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
```diff
@@ -0,0 +1,83 @@
...
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_zmap_empty(self, xp, scp, dtype):
+        x = xp.array([], dtype=dtype)
+        y = xp.array([1, 3, 5], dtype=dtype)
+        return scp.stats.zmap(x, y)
+
+    @testing.for_dtypes('fdFD')
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_zmap_nan_pocily_omit(self, xp, scp, dtype):
```
takagi, line 66: Would you add a test with the `nan_policy='propagate'` (or default) with taking nan inputs?

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    a = compare
+
+    if a.size == 0:
+        return cupy.empty(a.shape)
+
+    contains_nan, nan_policy = _contains_nan(a, nan_policy)
+
+    if contains_nan and nan_policy == 'omit':
```
takagi, line 124: Seems that `_quiet_nanmean` and `_quiet_nanstd` is just to avoid making a warning for arrays that have only nan.

```py
>>> np.nanmean([np.nan, np.nan])
<stdin>:1: RuntimeWarning: Mean of empty slice
nan/
```

We may use `cupy.nanmean` and `cupy.nanstd` as they do not make a warning in such a case.

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,147 @@
...
+        if axis is None:
+            mn = _quiet_nanmean(a.ravel())
+            std = _quiet_nanstd(a.ravel(), ddof=ddof)
+            isconst = _isconst(a.ravel())
+        else:
+            mn = cupy.apply_along_axis(_quiet_nanmean, axis, a)
+            std = cupy.apply_along_axis(_quiet_nanstd, axis, a, ddof=ddof)
```
takagi, line 131: CuPy has `cupy.apply_along_axis`, but it is quite inefficient because it repeatedly launches kernels for 1d-arrays along the axis. We may simply use `cupy.nanmean` and `cupy.nanstd`.

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if a.size == 0:
+        return cupy.empty(a.shape)
+
+    contains_nan, nan_policy = _contains_nan(a, nan_policy)
+
+    if contains_nan and nan_policy == 'omit':
+        if axis is None:
+            mn = _quiet_nanmean(a.ravel())
+            std = _quiet_nanstd(a.ravel(), ddof=ddof)
```
takagi, line 127: Seems that `_quiet_nanmean` and `_quiet_nanstd` is just to avoid making a warning for arrays that have only nan.

```py
>>> np.nanmean([np.nan, np.nan])
<stdin>:1: RuntimeWarning: Mean of empty slice
nan
```

We may use `cupy.nanmean` and `cupy.nanstd` as they do not make a warning in such a case.

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,147 @@
...
+            isconst = _isconst(a.ravel())
+        else:
+            mn = cupy.apply_along_axis(_quiet_nanmean, axis, a)
+            std = cupy.apply_along_axis(_quiet_nanstd, axis, a, ddof=ddof)
+            isconst = cupy.apply_along_axis(_isconst, axis, a)
```
takagi, line 132: Is it possible to implement an equivalent thing with `cupy.apply_along_axis(_isconst, axis, a)` without using `apply_along_axis` as it is very inefficient?

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,147 @@
...
+        try:
+            contains_nan = cupy.nan in set(a.ravel())
+        except TypeError:
+            contains_nan = False
+            nan_policy = 'omit'
+            warnings.warn("The input array could not be properly "
+                          "checked for nan values. nan values "
+                          "will be ignored.", RuntimeWarning)
+
+    if contains_nan and nan_policy == 'raise':
```
takagi, line 25: We want evaluate `bool(contains_nan)` only when it is truly necessary as it causes device-to-host sync and inccurs the performance. `contains_nan` is a boolean array on GPU.

```suggestion
    if nan_policy == 'raise' and contains_nan:
```

cupyx/scipy/stats/_stats_py.py
khushi-411: Great idea! I'll take a look at it.

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if nan_policy not in policies:
+        raise ValueError("nan_policy must be one of {%s}" %
+                         ', '.join("'%s'" % s for s in policies))
+
+    try:
+        contains_nan = cupy.isnan(cupy.sum(a))
+    except TypeError:
+        try:
+            contains_nan = cupy.nan in set(a.ravel())
```
takagi, line 17: We cannot pass cupy.ndarray to make a set. When do you think the line above `contains_nan = cupy.isnan(cupy.sum(a))` raises a TypeError, and this `except` block is needed?

cupyx/scipy/stats/_stats_py.py
takagi, line 124: Can we remove this `contains_nan` as evaluating `bool(contains_nan)` makes host-to-device sync and we don't want to have it as much as possible.

takagi:
Sorry for late review. Would you check my comments? I think we have several points to improve the implementation to get better performance.

cupyx/scipy/stats/_stats_py.py
takagi: I guess this line would help you with modifying to omit nan.

https://github.com/cupy/cupy/blob/babc4994d14dd248ed94c84a921bf64613f853be/cupyx/scipy/stats/_stats_py.py#L139

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
khushi-411: Fixed

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
khushi-411: Added test. Thanks for your input :)

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
khushi-411: Yep, added. Thanks!

cupyx/scipy/stats/_stats_py.py
khushi-411: Hi, @takagi!
I've removed `apply_along_axis` and used `cupy.nanmean` and `cupy.nanstd`. But I see there is one drawback to this:
- `cupy.nanstd` that uses `cupy.nanvar` which is not implemented for complex dtypes. But we will require cases where we use nan_polity=`omit` and require the `axis` argument.

Using apply_along_axis resolves these problems because we use `mean` and `std` that support complex dtypes.

Anyways, for now, I've removed the `apply_along_axis`, considering performance as the first priority.

What would you prefer? Can we add support for complex dtypes in `nanvar`?

Thank you!

cupyx/scipy/stats/_stats_py.py
khushi-411: Ah, right! Thanks for sharing.

cupyx/scipy/stats/_stats_py.py
khushi-411: I think we don't need `contain_nan` now because `nanmean`, and `nanstd` conventionally ignore nans. That was required when we used `apply_along_axis` initially.

cupyx/scipy/stats/_stats_py.py
khushi-411: SciPy says: [Link here](https://github.com/scipy/scipy/blob/main/scipy/stats/_stats_py.py#L108-L111)
```txt
# This can happen when attempting to sum things which are not
# numbers (e.g. as in the function `mode`). Try an alternative method:
```

Shall I remove it?

khushi-411:
Thank you very much for looking into the PR! I've updated the PR according to your reviews. I had some thoughts on a few points. Could you please look into it?

cupyx/scipy/stats/_stats_py.py
takagi: Let's support complex numbers in `cupy.nanvar`. I'll put a pointer soon.

takagi: When we remove the assertion at 
https://github.com/cupy/cupy/blob/fc04c9445f488272c9a1b78999e4e0df15a721b5/cupy/_core/_routines_statistics.pyx#L698

, we'll get the following error:

```
TypeError: Wrong type ((<class 'numpy.complex64'>,)) of arguments for cupy_count_non_nan
```

Let's add complex support to `_count_non_nan`, that is, `F->q` and `D->q`

https://github.com/cupy/cupy/blob/fc04c9445f488272c9a1b78999e4e0df15a721b5/cupy/_core/_routines_statistics.pyx#L682-L685

, then we'll get the following compile error:

```
cupy.cuda.compiler.CompileException: /tmp/tmp380imuka/328e8b517679a3804b06c7d50f2bc4fa55c4a2df.cubin.cu(42): error: no operator "/" matches
these operands
            operand types are: const T / const long long

/tmp/tmp380imuka/328e8b517679a3804b06c7d50f2bc4fa55c4a2df.cubin.cu(11): error: no operator "/" matches these operands
            operand types are: T / long long
          detected during instantiation of "T nanvar_impl(S, T, long long) [with S=S, T=T]"
(42): here
```

The related complex division is defined in 

https://github.com/cupy/cupy/blob/fc04c9445f488272c9a1b78999e4e0df15a721b5/cupy/_core/include/cupy/complex/arithmetic.h#L110-L114

, where it requires that the divisor has the same type as the dividend's element type `T`. I think we can relax the constraint with another typename `S`.

Would you make another PR to support complex numbers in `cupy.nanvar`?

cupyx/scipy/stats/_stats_py.py
takagi: Thanks, let's remove it!

cupyx/scipy/stats/_stats_py.py
takagi: Okay, then let's remove it!

cupyx/scipy/stats/_stats_py.py
khushi-411: Interesting, Thanks for the clear description!

> Would you make another PR to support complex numbers in cupy.nanvar?

Definitely, I'll send a PR within some hours!

takagi: Thanks!

cupyx/scipy/stats/_stats_py.py
khushi-411: Removed it.

takagi:
any reason for test file name `test_stats_py.py` instead of `test_stats.py`?

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @testing.for_dtypes('fdFD')
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_zmap_nan_policy_omit_axis_ddof(self, xp, scp, dtype):
+        x = xp.array([4.0, 1.0, 1.0, cupy.nan], dtype=dtype)
+        y = xp.array([cupy.nan, -4.0, -1.0, -5.0], dtype=dtype)
+        return scp.stats.zmap(x, y, axis=0, ddof=1, nan_policy='omit')
+
+    @testing.for_dtypes('fdFD')
+    def test_zmap_nan_policy_raise(self, dtype):
```
takagi, line 95: Let's test that both scipy and cupy raise an error.
```py
for xp, scp in [(numpy, scipy), (cupy, cupyx.scipy)]:
    ....
        ...
```

khushi-411:
> nit: any reason for test file name `test_stats_py.py` instead of `test_stats.py`?

Oops, nice catch. I'll change it. Thanks for pointing it out :)

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
```diff
@@ -0,0 +1,178 @@
...
+        return scp.stats.zscore(x, nan_policy='omit')
+
+    @testing.for_dtypes('fdFD')
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_zmap_nan_policy_omit_axis_ddof(self, xp, scp, dtype):
+        x = xp.array([4.0, 1.0, 1.0, cupy.nan], dtype=dtype)
+        return scp.stats.zscore(x, axis=0, ddof=1, nan_policy='omit')
+
+    @testing.for_dtypes('fdFD')
+    def test_zscore_nan_policy_raise(self, dtype):
```
takagi, line 175: ditto (testing both numpy and cupy raise the same error)

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
khushi-411: Yes, sure! Update it, thanks!

tests/cupyx_tests/scipy_tests/stats_tests/test_stats_py.py
khushi-411: Done

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,143 @@
...
+    policies = ['propagate', 'raise', 'omit']
+
+    if nan_policy not in policies:
+        raise ValueError("nan_policy must be one of {%s}" %
+                         ', '.join("'%s'" % s for s in policies))
+
+    contains_nan = cupy.isnan(cupy.sum(a))
+
+    if nan_policy == 'raise' and contains_nan:
+        raise ValueError("The input contains nan values")
```
takagi, line 14: Let's compute contains_nan only when nan_policy is `raise`.

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,143 @@
...
+            isconst = _isconst(a.ravel())
+        else:
+            mn = cupy.nanmean(a, axis=axis, keepdims=True)
+            std = cupy.nanstd(a, axis=axis, keepdims=True, ddof=ddof)
+            isconst = (_first(a, axis) == a).all(axis=axis, keepdims=True)
+    else:
+        mn = a.mean(axis=axis, keepdims=True)
+        std = a.std(axis=axis, ddof=ddof, keepdims=True)
+        if axis is None:
+            isconst = (a.item(0) == a).all()
```
takagi, line 133: `ndarray.item()` transfers data from GPU to the host, then convert it to a Python scalar. We may use `a.ravel()[0]` instead to make a comparison with `a` whose data is on GPU.

takagi:
I will kick the CI once #6869 is merged.

cupyx/scipy/stats/_stats_py.py
khushi-411: Sure, I've updated it. I think we now don't need to define that in another function. Have changed it. Hope it's okay!

cupyx/scipy/stats/_stats_py.py
khushi-411: Got that! Thanks for the concept :)

cupyx/scipy/stats/_stats_py.py
```diff
@@ -0,0 +1,138 @@
...
+
+    a = compare
+
+    if a.size == 0:
+        return cupy.empty(a.shape)
+
+    if nan_policy == 'raise':
+        contains_nan = cupy.isnan(cupy.sum(a))
+
+        if contains_nan:
```
takagi, line 112:
```suggestion
        if contains_nan:  # synchronize!
```

takagi (approved):
LGTM!

takagi (dismissed):

takagi (dismissed):

takagi:
It seems that `nan_policy` was introduced in scipy 1.7, so would you mark `nan_policy` tests with `@testing.with_requires('scipy>=1.7')`?

tests/cupyx_tests/scipy_tests/stats_tests/test_stats.py
```diff
@@ -60,3 +64,173 @@ def test_propotion_too_big(self):
...
+            y = xp.array([8, -4, xp.nan, 4], dtype=dtype)
+            with pytest.raises(ValueError):
+                scp.stats.zmap(x, y, nan_policy='raise')
+
+
+@testing.with_requires('scipy')
+class TestZscore:
+
+    @testing.for_all_dtypes(no_float16=True, no_bool=True)
+    @testing.numpy_cupy_allclose(scipy_name='scp', rtol=rtol)
```
takagi, line 162: aid for some CI failures
```suggestion
    @testing.numpy_cupy_allclose(scipy_name='scp', atol=atol, rtol=rtol)
```

khushi-411: Ho updated it. Thanks for the reviews and for re-kicking the CI multiple times!

tests/cupyx_tests/scipy_tests/stats_tests/test_stats.py
```diff
@@ -60,3 +64,173 @@ def test_propotion_too_big(self):
...
+    @testing.for_all_dtypes(no_bool=True)
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_zmap_empty(self, xp, scp, dtype):
+        x = xp.array([], dtype=dtype)
+        y = xp.array([1, 3, 5], dtype=dtype)
+        return scp.stats.zmap(x, y)
+
+    @testing.for_dtypes('fdFD')
+    @testing.numpy_cupy_allclose(scipy_name='scp', atol=atol, rtol=rtol)
+    def test_zmap_nan_policy_propagate(self, xp, scp, dtype):
```
takagi, line 129: We want to skip only tests that use `nan_policy`
```suggestion
    @testing.with_requires('scipy>=1.7')
    def test_zmap_nan_policy_propagate(self, xp, scp, dtype):
```

tests/cupyx_tests/scipy_tests/stats_tests/test_stats.py
```diff
@@ -60,3 +64,173 @@ def test_propotion_too_big(self):
             a = xp.array([4, 8, 2, 0, 9, 5, 10, 1, 7, 3, 6])
             with pytest.raises(ValueError):
                 scp.stats.trim_mean(a, 0.6)
+
+
+@testing.with_requires('scipy>=1.7')
```
takagi, line 69: Please see https://github.com/cupy/cupy/pull/6855/files#r935552655.
```suggestion
@testing.with_requires('scipy')
```

tests/cupyx_tests/scipy_tests/stats_tests/test_stats.py
khushi-411: Got that!

tests/cupyx_tests/scipy_tests/stats_tests/test_stats.py
```diff
@@ -60,3 +64,181 @@ def test_propotion_too_big(self):
             a = xp.array([4, 8, 2, 0, 9, 5, 10, 1, 7, 3, 6])
             with pytest.raises(ValueError):
                 scp.stats.trim_mean(a, 0.6)
+
+
+@testing.with_requires('scipy')
+class TestZmap:
+
+    @testing.for_all_dtypes(no_bool=True)
+    @testing.numpy_cupy_allclose(scipy_name='scp')
```
takagi, line 73:
```suggestion
    @testing.numpy_cupy_allclose(scipy_name='scp', atol=atol, rtol=rtol)
```

tests/cupyx_tests/scipy_tests/stats_tests/test_stats.py
```diff
@@ -60,3 +64,181 @@ def test_propotion_too_big(self):
...
+    def test_zmap_1dim(self, xp, scp, dtype):
+        x = testing.shaped_random((10,), xp, dtype=dtype)
+        y = testing.shaped_random((8,), xp, dtype=dtype)
+        return scp.stats.zmap(x, y)
+
+    @testing.for_all_dtypes(no_bool=True)
+    @testing.numpy_cupy_allclose(scipy_name='scp')
```
takagi, line 80:
```suggestion
    @testing.numpy_cupy_allclose(scipy_name='scp', atol=atol, rtol=rtol)
```

takagi:
Would you merge this @kmaehashi ?

takagi (approved):

takagi:
Thanks, @khushi-411!

khushi-411:
Thanks for your guidance, @takagi!

PR #6871, opened 2022-07-18
Update `cupy.array_api`

leofang:
Apply patches from the following two upstream PRs:
- https://github.com/numpy/numpy/pull/21260
- https://github.com/numpy/numpy/pull/21377

After #6853 is merged I'll send another update PR.

ps: I think this PR can be backported?

### DIFF
--- a/cupy/array_api/_array_object.py
+++ b/cupy/array_api/_array_object.py
@@ -289,11 +289,11 @@ class Array:
         for i in _key:
             if isinstance(i, bool) or not (
                 isinstance(i, SupportsIndex)  # i.e. ints
+                or isinstance(i, Array)
+                or isinstance(i, np.ndarray)
                 or isinstance(i, slice)
                 or i == Ellipsis
                 or i is None
-                or isinstance(i, Array)
-                or isinstance(i, np.ndarray)
             ):
                 raise IndexError(
                     f"Single-axes index {i} has {type(i)=}, but only "
--- a/cupy/array_api/_array_object.py
+++ b/cupy/array_api/_array_object.py
@@ -29,7 +29,7 @@ from ._dtypes import (
     _dtype_categories,
 )
 
-from typing import TYPE_CHECKING, Optional, Tuple, Union, Any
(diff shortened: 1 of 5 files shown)


tests/cupy_tests/array_api_tests/test_array_object.py
```diff
@@ -70,11 +72,6 @@ def test_validate_index():
     assert_raises(IndexError, lambda: a[[0, 1]])
     assert_raises(IndexError, lambda: a[cp.array([[0, 1]])])
```
leofang, line 73: This test fails on the CI because of #6893.

leofang:
@kmaehashi do you know what's up with Jenkins? Other two CI failures are irrelevant.

leofang:
I see, thanks @kmaehashi. The array-api CI fails likely because something is wrong when checking out the test suite. The cuda117 CI failure is irrelevant.

tests/cupy_tests/array_api_tests/test_array_object.py
leofang: Fixed in 0788cfe.

cupy/array_api/_array_object.py
```diff
@@ -289,11 +289,11 @@ def _validate_index(self, key):
         for i in _key:
             if isinstance(i, bool) or not (
                 isinstance(i, SupportsIndex)  # i.e. ints
+                or isinstance(i, Array)
+                or isinstance(i, np.ndarray)
                 or isinstance(i, slice)
                 or i == Ellipsis
                 or i is None
-                or isinstance(i, Array)
-                or isinstance(i, np.ndarray)
```
leofang: @asmeurer The check of array instances after the `==` comparison is invalid, because both NumPy and CuPy would generated a list of bools, leading to `ValueError: ... Use .any() or .all()` instead. I'll send a fix to NumPy.

asi1024:
LGTM! Thanks!

asi1024 (approved):

PR #6872, opened 2022-07-18
Accept `kind` in `sort`/`argsort` and fix `cupy.array_api.{sort,argsort}` accordingly

### LINKED ISSUE #5701: Sorting functions miss arguments `kind` and `order`
Refs:
- `argsort()`:
  - NumPy: https://numpy.org/doc/stable/reference/generated/numpy.argsort.html
  - CuPy: https://docs.cupy.dev/en/stable/reference/generated/cupy.argsort.html
- `sort()`
  - NumPy: https://numpy.org/doc/stable/reference/generated/numpy.sort.html
  - CuPy: https://docs.cupy.dev/en/stable/reference/generated/cupy.sort.html

I haven't checked other sorting functions (ex: `partition()`) but they probably missed them too. Though, Array API v1 only needs the above two sorting functions.

The `kind` argument is needed by the Array API standard (#5698, #4789).

### LINKED ISSUE #6779: array_api sort() does not work
### Description

The array_api sort() is copied from the NumPy implementation, but passes through a `kind` keyword which doesn't exist in `cupy.sort`.

### To Reproduce

```py
>>> a = xp.asarray([1, 0, 2])
>>> xp.sort(a)
  File "/home/aaronmeurer/anaconda3/envs/array-api/lib/python3.10/site-packages/cupy/array_api/_sorting_functions.py", line 45, in sort
    res = np.sort(x._array, axis=axis, kind=kind)
TypeError: sort() got an unexpected keyword argument 'kind'
```

### Installation

### Environment

```
# Paste the output here
```

### Additional Information

The array API does require support for a `stable` keyword. If `cupy.sort` is always stable, it can just ignore this flag, as `stable=False` is allowed to be stable (see https://data-apis.org/array-api/latest/API_specification/generated/signatures.sorting_functions.sort.html). Otherwise, I would suggest raising NotImplementedError for `stable=True` (which is unfortunately the default).

leofang:
This is for both NumPy compliance (#5701) and Array API compliance (#6779).

What this PR does is to simply add the kw-arg `kind`, allowing passing through of the argument.

However, given that currently there's only one sorting algorithm implemented in CuPy (from Thrust), we can't support `kind` being anything other than `None` or `"stable"`. So for Array API internally we always set `kind` to `None` regardless if users set `stable` to `True` or `False`. (Confirmed during a meeting that this was the intent of the standard.)

Close #5701. Close #6779.

cc: @asmeurer

### DIFF
--- a/cupy/_sorting/sort.py
+++ b/cupy/_sorting/sort.py
@@ -4,7 +4,7 @@ import numpy
 from cupy.cuda import thrust
 
 
-def sort(a, axis=-1):
+def sort(a, axis=-1, kind=None):
     """Returns a sorted copy of an array with a stable sorting algorithm.
 
     Args:
@@ -12,6 +12,8 @@ def sort(a, axis=-1):
         axis (int or None): Axis along which to sort. Default is -1, which
             means sort along the last axis. If None is supplied, the array is
             flattened before sorting.
+        kind: Default is `None`, which is equivalent to 'stable'. Unlike in
+            NumPy any other options are not accepted here.
 
     Returns:
         cupy.ndarray: Array of the same type and shape as ``a``.
@@ -24,6 +26,8 @@ def sort(a, axis=-1):
     .. seealso:: :func:`numpy.sort`
 
     """
+    if kind is not None and kind != 'stable':
+        raise ValueError("kind can only be None or 'stable'")
     if axis is None:
         ret = a.flatten()
         axis = -1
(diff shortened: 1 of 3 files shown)


kmaehashi (approved):
LGTM!

leofang:
It seems the array-api CI has some issues...?

leofang:
This is ready.

kmaehashi (approved):

PR #6874, opened 2022-07-18
Make sure that cupy (array-api) Array objects can be composed using asarray

### LINKED ISSUE #6367: `cupy.array_api.asarray` doesn't work with list of arrays
### Description

CuPy's Array API module doesn't work with lists of arrays, while the main CuPy namespace works fine. `numpy.array_api` also works fine for this usecase.

### To Reproduce

```py
In [1]: import cupy.array_api as xp

In [2]: xp.asarray([xp.ones(5), xp.ones(5)])
<ipython-input-3-2681967f8885> in <module>
----> 1 xp.asarray([xp.ones(5), xp.ones(5)])

~\mambaforge\envs\scipy-cupy\lib\site-packages\cupy\array_api\_creation_functions.py in asarray(obj, dtype, device, copy)
     81     try:
     82         runtime.setDevice(device.id)
---> 83         res = np.asarray(obj, dtype=dtype)
     84     finally:
     85         runtime.setDevice(prev_device)

~\mambaforge\envs\scipy-cupy\lib\site-packages\cupy\_creation\from_data.py in asarray(a, dtype, order)
     74
     75     """
---> 76     return _core.array(a, dtype, False, order)
     77
     78

cupy\_core\core.pyx in cupy._core.core.array()

cupy\_core\core.pyx in cupy._core.core.array()

cupy\_core\core.pyx in cupy._core.core._array_default()

ValueError: Unsupported dtype object
```

### LINKED ISSUE #6882: Array creation functions should accept a sequence of objects that implement `__cupy_get_ndarray__`
### Description

Root cause of #6367, see https://github.com/cupy/cupy/pull/6874#issuecomment-1189740670.

### To Reproduce

_No response_

### Installation

_No response_

### Environment

```
# Paste the output here
```

### Additional Information

To address this, I believe we need to touch `cupy/_core/core.pyx`. We need to modify `_array_info_from_nested_sequence` and/or `_array_from_nested_sequence`, such that it recognizes a sequence of objects that implement `__cupy_get_ndarray__` is a valid object.

andfoy:
Fixes #6367 
Fixes #6882 

This patch will only work for Cupy and Numpy Array API objects since they define the non-standard `_array` attribute


cupy/array_api/_creation_functions.py
```diff
@@ -77,6 +77,10 @@ def asarray(
         # Give a better error message in this case. NumPy would convert this
         # to an object array. TODO: This won't handle large integers in lists.
         raise OverflowError("Integer out of bounds for array dtypes")
+    if isinstance(obj, (list, tuple)):
```
leofang, line 80: nit: we should check for sequences here, not just lists and tuples.

leofang:
cc: @asmeurer

leofang:
Sorry, I have another question... Since CuPy copied NumPy's Array API implementation, do we know why this is supported in NumPy but not in CuPy? What did we miss?

leofang:
I feel the real fix might need to go to `cupy.asarray`:
```python
>>> import numpy as np
>>> import numpy.array_api as xp
<stdin>:1: UserWarning: The numpy.array_api submodule is still experimental. See NEP 47.
>>> x = [xp.ones(5), xp.ones(5)]
>>> np.asarray(x, dtype=None)
array([[1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.]])
>>> 
>>> 
>>> import cupy as cp
>>> import cupy.array_api as xp
<stdin>:1: UserWarning: The cupy.array_api submodule is still experimental. See NEP 47.
>>> x = [xp.ones(5), xp.ones(5)]
>>> cp.asarray(x, dtype=None)
  File "/home/leof/dev/cupy_cuda114/cupy/_creation/from_data.py", line 76, in asarray
    return _core.array(a, dtype, False, order)
  File "cupy/_core/core.pyx", line 2357, in cupy._core.core.array
    cpdef _ndarray_base array(obj, dtype=None, bint copy=True, order='K',
  File "cupy/_core/core.pyx", line 2381, in cupy._core.core.array
    return _array_default(obj, dtype, order, ndmin)
  File "cupy/_core/core.pyx", line 2509, in cupy._core.core._array_default
    raise ValueError('Unsupported dtype %s' % a_cpu.dtype)
ValueError: Unsupported dtype object
```

leofang:
Ah I figured it out. NumPy has `__array__` implemented to the Array object (https://github.com/numpy/numpy/pull/20527), but we purposely replaced it with `__cupy_get_ndarray__`. So the problem seems to be with a sequence of objects the CuPy conversion does not kick in, because this works fine:
```python
>>> import cupy as cp
>>> class A:
...     def __cupy_get_ndarray__(self):
...             return cp.zeros(5)
... 
>>> 
>>> cp.asarray(A())
array([0., 0., 0., 0., 0.])
```

leofang:
My 2c here is we should aim to fix #6882, as #6367 is just the symptom, but it's up to @kmaehashi to decide 🙂

andfoy:
Thanks for the pointers! I was trying to fiddle with `core.pyx` but I didn't know the functions had to check for `__cupy_imterface__`

cupy/_core/_routines_manipulation.pyx
```diff
@@ -594,7 +594,11 @@ cpdef _ndarray_base concatenate_method(
     if dtype is not None:
         dtype = get_dtype(dtype)
 
-    arrays = list(tup)
+    arrays = []
+    for array in tup:
+        if hasattr(array, '__cupy_get_ndarray__'):
```
andfoy, line 599: Should we consider `__array__` as well?

andfoy: Any ideas on this one?

rgommers: This seems inconsistent in numpy right now:
```python
>>> import cupy as cp
>>> import numpy as np
>>> import torch
>>> import numpy.array_api

>>> t = torch.tensor([1, 2], device='cpu')  # has __array__, not buffer protocol
>>> np.array_api.asarray(t)
Array([1, 2], dtype=int64)
>>> np.array_api.asarray(x)
Array([0, 1], dtype=int64)
>>> np.array_api.asarray([x, x])  # via buffer protocol says the array_api.asarray docstring
Array([[0, 1],
       [0, 1]], dtype=int64)

>>> # The issue:
>>> np.array_api.asarray([t, t])
/home/rgommers/anaconda3/envs/cupy-torch/lib/python3.9/site-packages/numpy/array_api/_creation_functions.py:71: FutureWarning: The input object of type 'Tensor' is an array-like implementing one of the corresponding protocols (`__array__`, `__array_interface__` or `__array_struct__`); but not a sequence (or 0-D). In the future, this object will be coerced as if it was first converted using `np.array(obj)`. To retain the old behaviour, you have to either modify the type 'Tensor', or assign to an empty array created with `np.empty(correct_shape, dtype=object)`.
  res = np.asarray(obj, dtype=dtype)
/home/rgommers/anaconda3/envs/cupy-torch/lib/python3.9/site-packages/numpy/array_api/_creation_functions.py:71: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  res = np.asarray(obj, dtype=dtype)
  Input In [31] in <cell line: 1>
    np.array_api.asarray([t, t])
  File ~/anaconda3/envs/cupy-torch/lib/python3.9/site-packages/numpy/array_api/_creation_functions.py:72 in asarray
    return Array._new(res)
  File ~/anaconda3/envs/cupy-torch/lib/python3.9/site-packages/numpy/array_api/_array_object.py:80 in _new
    raise TypeError(
TypeError: The array_api namespace does not support the dtype 'object'
```

So:
-  `__array__` is used for converting a single array
- a sequence where elements support the buffer protocol are okay too
- but, a sequence where elements support `__array__` is not.

Unclear if that is on purpose. @asmeurer can you comment?

asmeurer: Does `np.asarray` work here? np.asarray just calls out to that.

rgommers: Kinda but not really:
```
>>> np.asarray([t, t])
<ipython-input-33-035f9ee92d0b>:1: FutureWarning: The input object of type 'Tensor' is an array-like implementing one of the corresponding protocols (`__array__`, `__array_interface__` or `__array_struct__`); but not a sequence (or 0-D). In the future, this object will be coerced as if it was first converted using `np.array(obj)`. To retain the old behaviour, you have to either modify the type 'Tensor', or assign to an empty array created with `np.empty(correct_shape, dtype=object)`.
  np.asarray([t, t])
<ipython-input-33-035f9ee92d0b>:1: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  np.asarray([t, t])
array([tensor([1, 2]), tensor([1, 2])], dtype=object)
```

asmeurer: Well that's all it's doing is calling np.asarray, and when it sees that the result is an object array, it hits the guard against that and gives the corresponding error. So I'd say the bug here is that np.asarray doesn't recursively call asarray on sequences. Surely this has been discussed before.

andfoy: I'm resolving this one in the meantime due a lack of clarity about this path on NumPy. Given the scope of this PR, only CuPy array-api objects will get composed

andfoy:
@kmaehashi @leofang I've just added a test, please let me know if the changes are correct

cupy/array_api/_typing.py
```diff
@@ -28,7 +28,7 @@
     Union,
     TYPE_CHECKING,
     TypeVar,
-    Protocol,
+    Protocol
```
asi1024, line 31: This code change seems unrelated to this PR.

cupy/array_api/_creation_functions.py
```diff
@@ -77,6 +77,7 @@ def asarray(
         # Give a better error message in this case. NumPy would convert this
         # to an object array. TODO: This won't handle large integers in lists.
         raise OverflowError("Integer out of bounds for array dtypes")
+
```
asi1024, line 80: Ditto.

asi1024 (approved):

asi1024:
@andfoy LGTM! Thank you for your contribution!

PR #6799, opened 2022-06-22
Utilize CUDA Enhanced Compatibility in all wrappers

kmaehashi:
This PR completes the support of CUDA Enhanced Compatibility in CuPy.

Using the mechanism introduced in #6730, this PR lazy loads all symbols added in CUDA 11.2 or later.
This should allow CuPy built with any of CUDA 11.2+ to run on any of CUDA 11.2+.

PR #6879, opened 2022-07-19
Remove proxy functions for softlink

kmaehashi:
This reduces the code needed to support new CUDA APIs.

Follows-up #6730 and #6799.


cupy_backends/cuda/_softlink.pxd
```diff
@@ -1,5 +1,7 @@
+ctypedef int (*func_ptr)(...) nogil  # NOQA
```
leofang, line 1: This is great 👍

emcastillo (approved):
Neat

PR #6849, opened 2022-07-01
[stats] add `cupyx.scipy.stats.boxcox_llf`

khushi-411:
Hi, the PR adds `cupyx.scipy.stats.boxcox_llf`!

#### Benchmarks (using cupy fusion for `log` and `sum`) :
size | dtype | SciPy | CuPy
-----|-------|-------|-----
1000 | int8 | 0.101 ms | 0.356 ms
1000 | int32 | 0.079 ms | 0.353 ms
1000 | uint32 | 0.080 ms | 0.414 ms
1000 | float32 | 0.076 ms | 0.389 ms
1000 | float64 | 0.084 ms | 0.457 ms
100000 | int8 | 4.178 ms | 0.458 ms
100000 | int32 | 2.964 ms | 0.522 ms
100000 | uint32 | 2.845 ms | 0.521 ms
100000 | float32 | 1.425 ms | 0.550 ms
100000 | float64 | 2.693 ms | 0.521 ms
1000000 | int8 | 41.181 ms | 3.824 ms
1000000 | int32 | 27.627 ms | 4.676 ms
1000000 | uint32 | 27.741 ms | 4.795 ms
1000000 | float32 | 13.221 ms | 4.771 ms
1000000 | float64 | 27.190 ms | 4.804 ms

#### Updated Benchmarks (without using cupy fusion):
size | dtype | SciPy | CuPy
-----|-------|-------|-----
1000 | int8 | 0.100 ms | 0.310 ms
1000 | int32 | 0.079 ms | 0.298 ms
1000 | uint32 | 0.079 ms | 0.370 ms
1000 | float32 | 0.072 ms | 0.413 ms
1000 | float64 | 0.077 ms | 0.331 ms
100000 | int8 | 4.397 ms | 0.453 ms
100000 | int32 | 2.930 ms | 0.514 ms
100000 | uint32 | 2.889 ms | 0.513 ms
100000 | float32 | 1.391 ms | 0.537 ms
100000 | float64 | 2.745 ms | 0.511 ms
1000000 | int8 | 44.292 ms | 3.863 ms
1000000 | int32 | 29.930 ms | 6.129 ms
1000000 | uint32 | 28.011 ms | 3.745 ms
1000000 | float32 | 14.635 ms | 4.381 ms
1000000 | float64 | 29.569 ms | 4.234 ms

<details>
<summary>Script (Click Please)</summary>

<p>

```python
import cupy
import cupyx
import cupyx.scipy.stats

import scipy

from cupy import testing
[...]

PR #6884, opened 2022-07-20
Fix `boxcox_llf`

takagi:
Following up https://github.com/cupy/cupy/pull/6849.

This PR fixes `boxcox_llf` to return a scalar array of nan instead of float nan when empty data is given as well as improving its tests on the testing tolerance and input ranges.


emcastillo (approved):
LGTM

kmaehashi:
No backport needed as #6849 is only in master.

PR #6894, opened 2022-07-25
Fix `XFAIL` for `tests/cupyx_tests/scipy_tests/sparse_tests/test_coo.py` when `scipy>=1.9.0rc2`

emcastillo:


kmaehashi:
The fix is in SciPy 1.9.0 https://github.com/scipy/scipy/pull/15235

PR #6898, opened 2022-07-25
Document matmul supports out

### LINKED ISSUE #6870: matmul with out=in doesn't raise an exception
Hello, sorry if I'm not following a template but I don't really have time. Just a quick remark to say that in the doc of cupy.matmul:

https://docs.cupy.dev/en/stable/reference/generated/cupy.matmul.html

there's a note "The out array as input is currently not supported.". Indeed I tried it and it doesn't give the right result, but I'm surprised I was able to try it without getting any exception or at least warning. I think this is a bit dangerous...

Thanks,

Camille

toslunar:
Fix #6870.


asi1024 (approved):

asi1024:
CI failure looks unrelated.

PR #6889, opened 2022-07-22
Update install instructions on README

kmaehashi:
**MERGE THIS JUST BEFORE v11 RELEASE.**
Depends on #6888.

PR #6902, opened 2022-07-26
Remove ROCm 4.2 or earlier from README

asi1024:


kmaehashi:
as discussed offline cherrypicked into #6889

PR #6905, opened 2022-07-26
Support `deg` in `cupy.angle`

emcastillo:
There are numerical issues so for negative numbers I try to avoid using fp16 type and anything promoting to fp16.
This is just to reuse the test machinery and avoid creating new tests only for this.

numyp does the full calc in fp16 resulting in 179.9 due to numerical issues. CuPy just returns 180

### DIFF
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -915,7 +915,7 @@ _angle = create_ufunc(
 
 
 _angle_deg = create_ufunc(
-    'cupy_angle',
+    'cupy_angle_deg',
     ('?->d', 'e->e', 'f->f', 'd->d',
      ('F->f', 'out0 = arg(in0) * (180.0 / M_PI)'),
      ('D->d', 'out0 = arg(in0) * (180.0 / M_PI)')),
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -914,6 +914,19 @@ _angle = create_ufunc(
     ''')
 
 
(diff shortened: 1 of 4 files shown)


cupy/_core/_routines_math.pyx
```diff
@@ -914,6 +914,19 @@ _angle = create_ufunc(
     ''')
 
 
+_angle_deg = create_ufunc(
+    'cupy_angle',
```
kmaehashi, line 918: Let's avoid giving the same kernel name as it can be confusing when profiling.
```suggestion
    'cupy_angle_deg',
```

tests/cupy_tests/math_tests/test_arithmetic.py
```diff
@@ -75,7 +85,11 @@ def test_unary(self, xp):
         arg1 = self.arg1
         if isinstance(arg1, numpy.ndarray):
             arg1 = xp.asarray(arg1)
-        y = getattr(xp, self.name)(arg1)
+
+        if self.name in ('angle'):
```
kmaehashi, line 89:
```suggestion
        if self.name in {'angle'}:
```

kmaehashi (changes requested):
LGTM except for nits!

emcastillo:
fixed!

kmaehashi (approved):

jakirkham:
Thanks all! 🙏

jakirkham:
Will this be included in CuPy 11.0.0?

kmaehashi:
> Will this be included in CuPy 11.0.0?

Yes.

Issue #2568, opened 2019-10-24
Avoid exposing functions overlapping with NCCL names


jakirkham:
Currently CuPy exposes NCCL stub functions that overlap with those that NCCL itself provides. This is done depending on whether a CUDA enabled build of CuPy was produced. Also which functions are exposed depends on which version of NCCL was used in the CuPy build. If the latest version of NCCL (supported by CuPy) was not available during the build, CuPy will expose some functions with names that overlap with those in NCCL. This could be an issue if CuPy were built against an older version of NCCL (like 2.3) and installed on a system with a newer version of NCCL (like 2.4). As a result both CuPy and NCCL will have symbols that clash. This would be detrimental to a user's program (causing crashes).

To fix this issue, it would be helpful if CuPy renamed the stub functions in [`cupy_nccl.h`]( https://github.com/cupy/cupy/blob/74ba7a94f58ffcb0a673441b0de7b8e7bfcb4fec/cupy/cuda/cupy_nccl.h ) with unique names not found in NCCL. A simple strategy might be to prefix all stub functions with  `cupy_*`. To simplify usability and maintainability, CuPy could make these stub functions either call NCCL or fallback to the some default behavior (like `return ncclSuccess`) depending on NCCL version.

Though this is merely one option. There could be other viable options. In any event it would be useful to avoid symbol clashes with NCCL.

cc @isuruf @leofang

leofang:
> This could be an issue if CuPy were built against an older version of NCCL (like 2.3) and installed on a system with a newer version of NCCL (like 2.4).

Let me fill in some context. @jakirkham correct me if I am wrong. So far CuPy always bundles a specific version of NCCL so that what CuPy was built against *is* what's actually delivered to and runs on users' systems. This saves `pip` users a lot of headache.

Now, with the Conda build https://github.com/conda-forge/staged-recipes/pull/9959 in mind, the idea is to try decoupling CuPy from NCCL, as NCCL will be offered as a separate Conda package (and is likely to be a single version used by all downstream Conda consumers), on which the new CuPy Conda package will depend. Thus, the build version is not necessary as the runtime version, although Conda folks will make sure it is as much as possible.

jakirkham:
That sounds like a good summary. Thanks for expanding that point, @leofang. 😄

> ...although Conda folks will make sure of it is as much as possible.

[...]

Issue #5649, opened 2021-08-17
Lazy-load CUDA Toolkit libraries


kmaehashi:
Currently, CUDA libraries are loaded when doing `import cupy`. But this have several issues:
* `import cupy` takes some time and memory (c.f., https://github.com/cupy/cupy/issues/5557#issuecomment-886378382)
* `import cupy` cannot be done without CUDA. CPU/GPU agnostic downstream libraries need to `try: import cupy except: ...` to support importing the library in non-CUDA environment.

By supporting [CUDA Python](https://github.com/NVIDIA/cuda-python) we can add a lazy-load layer to CuPy backends for free 😄
As the initial support PR merged (#5638), CuPy now lazy-loads CUDA driver when built with CUDA Python. We would like to support lazy-loading more libraries once CUDA Python adds support for other Toolkit libraries such as cuBLAS and cuRAND.

Related to #5641.

mmccarty:
One thing to think about is [SPEC 1](https://scientific-python.org/specs/spec-0001/) which contains a proposed standard for Lazy-loading submodules. It is based on [PEP 562](https://www.python.org/dev/peps/pep-0562/) which was introduced in Python 3.7.

kmaehashi:
Thanks for the info @mmccarty! I wasn't aware of the SPEC standardization.

One difficult thing for us is that we are using Cython so Python-level hacks may not be utilized; if we write `cimport bar` in `foo.pyx` (we often use `cimport` to avoid Python overhead), then `bar` module is automatically imported when importing `foo`.

leofang:
Another note on lazy loading. The two issues that prompted for this need should be separated:

> * `import cupy` takes some time and memory 
> * `import cupy` cannot be done without CUDA. 

The reason is that the first OOM tissue is ideally "fixed" since CUDA 11.7+ when `CUDA_MODULE_LOADING=LAZY` is set, see [the CUDA documentation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#lazy-loading). Any other solutions targeting the first issue should be considered as workarounds, as the proper fix has landed as part of the programming model.

As of CUDA 12.0, the default is still "EAGER" not "LAZY" (which needs opt-in), but I urge everyone to give it a shot and see if it helps (I believe it is). In the near future there's high chance that the default would be changed to "LAZY", unless there are unexpected blockers.

[...]

PR #6908, opened 2022-07-26
[RFC] Make `import cupy` always success

kmaehashi:
Currently `import cupy` requires GPU toolkit (CUDA/ROCm) installed.
However several people reported this is cumbersome, especially when developing or running NumPy/CuPy agnostic code under CPU-only machine.

This PR defers the notificaiton of import failure to the time of accessing CuPy features.

TODO: Also make `import cupyx` always success.


kmaehashi:
I come up with 3 approaches to fix this issue.

## 1. Discard errors when importing CUDA-linked modules and raise when CuPy features are invoked

Pros:
* Can be implemented quickly, as in this PR.

Cons:
* We will need to provide dummy (stub) classes. This may cause unexpected errors within CuPy codebase (e.g., dummy `fuse` is exposed during `import cupy`, so CuPy code cannot use `cupy.fuse` in global context).

## 2. Lazy-load all CUDA API symbols

Load CUDA symbols on the first call of the API. For example, in `runtime.pyx`, rewrite this:

```py
cdef extern from '../../cupy_backend_runtime.h' nogil:
    cudaGetDevice(int *device)

cpdef int getDevice():
    cdef int device
    status = cudaGetDevice(&device)
    check_status(status)
    return device
```

with this:

```py
cdef bool loaded
cdef func_ptr cudaGetDevice

cdef _init():
    if loaded: return
    rt = SoftLink("libcudart.so")
    cudaGetDevice = rt.get("cudaGetDevice")
    loaded = True

cpdef int getDevice():
    _load()  # NEW
    cdef int device
    status = cudaGetDevice(&device)
    check_status(status)
    return device
```

Pros:
* This also solves #5649
* This eliminates stub headers & headers for absorbing the API differences between CUDA versions
* Can easily migrate to CUDA Python later

Cons:
* We need to take care of the new state that we have never expected in our codebase, i.e. CUDA is not available

## 3. Bundle all CUDA toolkit libraries in wheels

Pros:
* CuPy will become importable and runnable without external shared library dependency.

Cons:
* Wheels will become very huge, meaning we can no longer distribute on PyPI

kmaehashi:
Considering pros and cons above, I'm now thinking the 2nd approach sounds the right way to go.

leofang:
I feel it'd be nice to coordinate. Ideally lazy-loading of all CUDA symbols should be done in CUDA Python, so CuPy doesn't have to reinvent the wheel. Now, if the blocker is CUDA Python lacks support for Math Libraries, we would like to know and figure out a path. Personally I'd rather like to see this gets done in CUDA Python properly, and then CuPy just makes CUDA Python a required dependency without worrying about it.

cc: @jakirkham @shwina @vzhurba01 @mnicely for vis

leofang:
Suggesting an alternative (as a **_workaround_** until CUDA Python supports Math Libraries, I guess?): Make a `cupy-stubs` wheel that is compiled with `READTHEDOCS=1`. Then, CPU-only users (including those who need to build the docs on RTD, see e.g. xarray-contrib/cupy-xarray#5) can just `pip install cupy-stubs` and then `import cupy`.

kmaehashi:
Another user who will benefit from this change: https://stackoverflow.com/questions/73054707/what-is-the-best-way-to-write-a-module-that-uses-cupy-if-available-and-falls-ba

> Ideally lazy-loading of all CUDA symbols should be done in CUDA Python, so CuPy doesn't have to reinvent the wheel.

I totally agree with this. In fact, this was one of the primary motivations to introduce CUDA Pythion (#5638).

> Now, if the blocker is CUDA Python lacks support for Math Libraries, we would like to know and figure out a path.

Yes, currently CuPy links against cuBLAS, cuBLASLt, cuFFT, cuRAND, cuSOLVER, and cuSPARSE. They are used in many places of CuPy codebase with `cimport` for performance, so it is difficult to do lazy-load in Python layer.

> Suggesting an alternative (as a **_workaround_** until CUDA Python supports Math Libraries, I guess?): Make a `cupy-stubs` wheel that is compiled with `READTHEDOCS=1`. Then, CPU-only users (including those who need to build the docs on RTD, see e.g. [xarray-contrib/cupy-xarray#5](https://github.com/xarray-contrib/cupy-xarray/issues/5)) can just `pip install cupy-stubs` and then `import cupy`.

This sounds like a good workaround as more packages started to depend on CuPy and they will likely need such a solution.

ideasrule:
For the case of running cupy/numpy agnostic code on a CPU-only machine, don't you want "import cupy" to fail so that you know to use numpy instead?  It would be great if cupy can automatically switch to the CPU version if a GPU isn't detected, but that's not what it currently does, and it seems like a lot of work would be necessary to implement that.

toslunar:
`cupy.is_available()` can be used.
https://github.com/cupy/cupy/blob/9608557bf2f7361940d52e8bc5db69de036689fe/cupy/__init__.py#L35-L36

kmaehashi:
If we lazy-load NCCL symbols, we can also remove NCCL symbols from `cupy_nccl.h` file and close #2568

pentschev:
Hi @kmaehashi , what is written in [item 2](https://github.com/cupy/cupy/pull/6908#issuecomment-1196459041) suggests that _all_ CUDA Toolkit libraries will be lazy-loaded when this is complete, is that correct? We've had an internal request for this in which the use case has to allow for partial installs too, that would mean the following would be required to lazy-load and not required to be installed too:

- cuda-nvrtc
- libcufft
- libcublas
- libcurand
- libcusparse
- libcusolver

Could you confirm that all the above will meet the requirement stated, and if not would it be possible to add that as well?

kmaehashi:
> Hi @kmaehashi , what is written in https://github.com/cupy/cupy/pull/6908#issuecomment-1196459041 suggests that all CUDA Toolkit libraries will be lazy-loaded when this is complete, is that correct?

Yes. To be clear, "lazy-load" still loads shared libraries at the time of `import cupy`, but using Python code (SoftLink) instead of OS's dynamic loader, to implement Python-level error handling when CUDA libraries could not be found.

After this task is complete, users can do `import cupy` and use a very limited set of features (e.g., use decorators like `@cupy.memoize` in a global context, use `cupy.ndarray` as an argument signature for mypy, etc.), without installing libraries you listed. Providing fine-grained set of features based on libraries available (e.g. disable sparse features if `libcusparse` is missing) is out of the scope of this PR, although it might be implemented in future.

pentschev:
> Yes. To be clear, "lazy-load" still loads shared libraries at the time of `import cupy`, but using Python code (SoftLink) instead of OS's dynamic loader, to implement Python-level error handling when CUDA libraries could not be found.
> 
> After this task is complete, users can do `import cupy` and use a very limited set of features (e.g., use decorators like `@cupy.memoize` in a global context, use `cupy.ndarray` as an argument signature for mypy, etc.), without installing libraries you listed.

Makes sense, thanks for confirming.

> Providing fine-grained set of features based on libraries available (e.g. disable sparse features if `libcusparse` is missing) is out of the scope of this PR, although it might be implemented in future.

When you say providing fine-grained set of/disabling features, what exactly do you mean? Hiding the relevant API, providing information to the use on what's available somehow, or something else? My understanding is that you won't indeed be able to run any of those if libraries are missing, so I guess CuPy will need anyway to fail somehow when the user attempts to call one of those functions.

kmaehashi:
> When you say providing fine-grained set of/disabling features, what exactly do you mean? Hiding the relevant API, providing information to the use on what's available somehow, or something else?

For example, making CuPy API except for `cupyx.scipy.sparse.*` available, when `libcusparse` is not installed. Currently, cuTENSOR/NCCL/cuDNN do this (https://github.com/cupy/cupy/blob/c836b9300717cd1a8e09f6db7d5b79b502c477cf/cupy/_environment.py#L381), but I don't intend to do this for CUDA Toolkit libraries; if any of CUDA toolkit libraries are missing, functionalities requiring CUDA will be disabled in CuPy.

pentschev:
Thanks @kmaehashi for the details. I think that is a reasonable solution for the time being, what would you think of raising an exception instead of (or in conjunction with) a warning? Maybe that would be better because it would allow continuing programatically, although I must admit I don't know if there would be other negative consequences to raising an exception.

kmaehashi:
Moving to #7620.

Issue #6873, opened 2022-07-18
Abandon `dtype_dispatcher` in favor of faster compilation time?


leofang:
[`dtype_dispatcher`](https://github.com/cupy/cupy/blob/fc04c9445f488272c9a1b78999e4e0df15a721b5/cupy/_core/include/cupy/type_dispatcher.cuh#L38) is a very useful utility originally borrowed from cuDF to avoid writing custom type dispatching logic. It was heavily used in our Thrust and CUB modules. However, the biggest drawback is it serializes the compilation time of all specializations, as all template instantiations are done in the same translation unit. 

cuNumeric recently faced the same issue for supporting Thrust sort. What bothers them was it significantly slowed down the developer time while still being in the rapid development. The workaround they did (https://github.com/nv-legate/cunumeric/pull/277) was to break down the function into N smaller files with C linkage. 

This is admittedly a very ugly hack, especially because CuPy's Thrust/CUB support is much more mature. However, we might want to consider a similar approach too, for the sake of reducing the compilation time. On conda-forge, the compilation on native aarch64/ppc64le CI exceeds Travis's cap (50 mins), so we can only use the emulator to build (I don't have time to explore cross-compilation), but even so we often exceed the 6-hr cap on Azure, and have to retry multiple times.

We should discuss any viable approaches, but there's no need to rush into any substantial changes. Keep in mind conda-forge is a community project, so we can't do multi-threading on the free resource there (at most 2 threads IIRC) and any change we make would likely not be very impactful there...

leofang:
@kmaehashi I have a question. I had in mind that we not only parallelize over the builds of extension modules, but also parallelize over building the `.cu` files for each data type. See [my prototype](https://github.com/leofang/cupy/tree/split_TU). However, I notice the latter is serialized by the for loop here:
https://github.com/cupy/cupy/blob/9d439227f1a53bbd9ebb16784e40bcce44bc8042/install/cupy_builder/_command.py#L51-L54
Do you think it is ok to launch `threading.Thread` here, or are there neater tricks applicable?

leofang:
I enabled a very native threading and saw tremendous speedup. The remaining question is how to avoid resource contention between the two parallelization levels.

cc: @jakirkham in case you have some thoughts on how we can reduce compilation time on conda-forge (or if it's possible at all...)

leofang:
[...]

Issue #4874, opened 2021-03-12
Using `-Xfatbin=-compress-all` for kernel compilation


jakirkham:
Not sure how relevant this is for how CuPy compiles things, but something we have been exploring in RAPIDS that may be of interest is building with `-Xfatbin=-compress-all`. In some cases this can make for quite dramatic reductions in compiled kernel sizes. For an example please see PR ( https://github.com/rapidsai/cudf/pull/7583 ). Again not sure if this is relevant for how CuPy is building things, but just wanted to mention it in case it is.

kmaehashi:
Thanks for sharing this! I tried this in locally (my environment uses `export CUPY_NVCC_GENERATE_CODE=arch=compute_60,code=sm_60;arch=compute_70,code=sm_70;arch=compute_80,code=sm_80`) and measured `du -csh cupy cupyx cupy_backends`:

* with compression: 95MB
* without compression: 102MB

In CuPy it seems the effect is limited because `nvcc` is only used for a few modules (cub/thrust/random).

Moreover, after compressing files into a wheel (ie ZIP), the **compressed** one was 262 KB larger.

leofang:
@kmaehashi Should we reopen this issue? I see that while this flag might not help reduce the wheel size, it helps reduce the total size taken by CuPy after decompression/installation: https://github.com/cupy/cupy/pull/6941#issuecomment-1204005112.

kmaehashi:
There should be some (maybe one time?) overhead when compression is enabled, so I'm wondering it's worth doing only for saving 17MB.

leofang:
It's another 15-25M for the `cub` module and I didn't check other modules. I don't see the build time increased, and the runtime overhead for decompression is negligible based on an internal discussion (though likely outdated). For reference, all RAPIDS binaries are built with this option.

Though, to be fair, it does not always help. For example, with [this feature](https://github.com/cupy/cupy/issues/6699#issuecomment-1120348863) I see `cub` inflates to 190M+, and the compression only brings it down to 170M+, not a significant change.

kmaehashi:
OK, let's try it and see how it would work :)

PR #6941, opened 2022-08-01
Split `cupy.cuda.thrust` into function-and-dtype-based translation units for parallelization

leofang:
Part of #6873.

Pros of the approach proposed in this PR:
- The parallelization over each function and dtype is possible
- Template specialization can be reduced easily if needed (no need for SFINAE hacks to disable certain dtypes)

Cons:
- The information of supported dtypes must leak into a few places
  1. The TU-generator script: This is because instead of hiding everything in the same TU, we now use a script to generate the needed TUs
  2. The implementation (`cupy/cuda/cupy_thrust.cu`): This is because we need a way to dispatch to the correct function pointer
- Future contributors will have to adopt this strategy to split the translation unit:
  - For early development, it's still possible to keep everything in `cupy/cuda/cupy_thrust.cu` until it's done

Once this PR is merged I'll apply the same strategy to `cupy.cuda.cub`.

To discuss:
- [x] Parallelization strategy to reduce resource contention
- [x] Dtype dispatch strategy to select the right function pointer
- [ ] Write docs (for developers)
- Misc:
  - [x] Handle sdist
  - [x] Detect source file changes to determine if rebuilding
- [x] Clean up
  - [ ] Rewrite the commit history?


leofang:
(updated the PR description)

leofang:
Time to cleanly re-compile `cupy.cuda.thrust` (other modules are untouched):
- On master: 10m13s
- This PR (up to 6d66e48d8): 0m41s

Test env:
- CPU: AMD Ryzen Threadripper PRO 3975WX 32-Cores
- RAM: 256 GB
- CUDA: 11.4
- `CUPY_NVCC_GENERATE_CODE`: unset (all SMs are targeted)
- `CUPY_NUM_NVCC_THREADS`: unset (2)
- `CUPY_NUM_BUILD_JOBS`: 16

leofang:
With b95877c the max number of concurrent threads is roughly `CUPY_NUM_NVCC_THREADS` * `CUPY_NUM_BUILD_JOBS`. Given that most other modules are built quickly and before `thrust`/`cub`, I feel this is the right way to reduce resource contention.

install/cupy_builder/install_utils.py
```diff
@@ -21,3 +21,19 @@ def search_on_path(filenames):
             if os.path.exists(full):
                 return os.path.abspath(full)
     return None
+
+
+def generate_translation_unit(
+        func_name, type_name, code_name, source_path) -> str:
+    with open(source_path) as f:
```
kmaehashi, line 28: Can we make these files committed instead of generating them during installation, like done in Legate? I'm wondering dynamically generating files will make users difficult to track what code is actually running.

install/cupy_builder/_command.py
```diff
@@ -28,40 +29,54 @@ def filter_files_by_extension(
...
         compiler = DeviceCompilerWin32(ctx)
     else:
         compiler = DeviceCompilerUnix(ctx)
 
     objects = []
+    futures = []
     for src in sources_cu:
         print(f'{ext.name}: Device code: {src}')
         obj_ext = 'obj' if sys.platform == 'win32' else 'o'
         # TODO(kmaehashi): embed CUDA version in path
```
kmaehashi, line 57: Future work to myself: we should embed (maybe a hash of) CUPY_NVCC_GENERATE_CODE to path as well.

install/cupy_builder/_command.py
```diff
@@ -28,40 +29,54 @@ def filter_files_by_extension(
...
-        if os.path.exists(obj) and (_get_timestamp(src) < _get_timestamp(obj)):
-            print(f'{ext.name}: Reusing cached object file: {obj}')
-        else:
-            os.makedirs(os.path.dirname(obj), exist_ok=True)
-            print(f'{ext.name}: Building: {obj}')
-            compiler.compile(obj, src, ext)
         objects.append(obj)
 
-    return sources_cpp, objects
+        # We need to look at the timestamps for the templates and other
```
kmaehashi, line 61: :+1:

install/cupy_builder/_command.py
```diff
@@ -110,7 +125,17 @@ def build_extension(self, ext: setuptools.Extension) -> None:
         ctx = cupy_builder.get_context()
 
         # Compile "*.cu" files into object files.
-        sources_cpp, extra_objects = compile_device_code(ctx, ext)
+        sources_cpp, extra_objects, futures = compile_device_code(ctx, ext)
+
+        # We may have many TUs in the queue waiting for nvcc to complete them,
+        # so for safety we should wait for the completion here
+        done, _ = concurrent.futures.wait(
```
kmaehashi, line 132: `build_extension()` will be called in parallel (`CUPY_NUM_BUILD_JOBS` threads), so this will cause `CUPY_NUM_BUILD_JOBS * CUPY_NUM_BUILD_JOBS * CUPY_NUM_NVCC_THREADS` threads to be launched at maximum. It's difficult for users to control this parallelism. How about doing nvcc compilation in `build_extensions()` (which is called before starting extension build in parallel) instead of here?

kmaehashi (commented):
Reviewed the installer part.

emcastillo:
I think the templates are fine, but I agree with @kmaehashi & I would recommend storing them in a directory and have a script to manually build and commit the generated codes to the repository instead of generating them on the fly. Similar to what we do with CI scripts.

cupy/cuda/cupy_thrust.cu
```diff
@@ -1,476 +1,62 @@
...
+    thrust_argsort_CUPY_TYPE_INT32     ,
+    thrust_argsort_CUPY_TYPE_UINT32    ,
+    thrust_argsort_CUPY_TYPE_INT64     ,
+    thrust_argsort_CUPY_TYPE_UINT64    ,
+    thrust_argsort_CUPY_TYPE_FLOAT16   ,
+    thrust_argsort_CUPY_TYPE_FLOAT32   ,
+    thrust_argsort_CUPY_TYPE_FLOAT64   ,
+    thrust_argsort_CUPY_TYPE_COMPLEX64 ,
+    thrust_argsort_CUPY_TYPE_COMPLEX128,
+    thrust_argsort_CUPY_TYPE_BOOL      ,
```
leofang, line 57: @emcastillo Here we collect an array of func ptr for runtime dispatching. I feel this is the only way, but any chance you can think of some template tricks that can achieve this (but without instantiating the template implicitly)?

install/cupy_builder/_command.py
leofang: Well, technically it's true, but in practice when we start building the `thrust`/`cub` modules all other modules are done, so when I monitored the progress it's much less (https://github.com/cupy/cupy/pull/6941#issuecomment-1200775453). That said, moving the `nvcc` part to `build_extensions()` sounds like a good idea, as it increases the parallelism (compiling all *.cu at once).

install/cupy_builder/_command.py
leofang: Yes I think this is a better solution that checking `settings['depends']`.

install/cupy_builder/install_utils.py
leofang: Done in 8173f5c.

install/cupy_builder/_command.py
leofang: Done in 8173f5c.

leofang:
(btw thanks for refactoring the build system @kmaehashi, it makes this experiment very easy to do 👍)

leofang:
@kmaehashi I have some concerns on the file size inflation when `CUPY_NVCC_GENERATE_CODE` is unset (so targeting all SMs):
- After this PR: `cupy/cuda/thrust.cpython-38-x86_64-linux-gnu.so` 124M
- Before this PR: `cupy/cuda/thrust.cpython-38-x86_64-linux-gnu.so` 57M

Similar inflation is observed with [this branch](https://github.com/leofang/cupy/tree/split_TU_cub2) applying the same strategy to `cupy.cuda.cub` (55M -> 119M). Do we know how good the compressibility is with wheels?

leofang:
> Do we know how good the compression rate is with wheels?

Taking [this wheel](https://files.pythonhosted.org/packages/24/31/359a0e6f384321f449cef0dc05cdf9b0e0426c52984c7c773f095de1f8d5/cupy_cuda11x-11.0.0-cp39-cp39-manylinux1_x86_64.whl) as example, a very rough estimation is 2.8x (81.8M -> 227.4M after decompression) and so the wheel size would be increased by ~50M in total with this PR applied to `thrust` and `cub`...

kmaehashi:
You can run `python setup.py bdist_wheel` to build a wheel, but I guess your assessment of the compression ratio is correct. Can #4874 work in this case?

leofang:
> Can #4874 work in this case?

Thanks for reminder! Adding `-Xfatbin=-compress-all` helps a lot:
- Before this PR: 57M
- With this PR, without `-Xfatbin=-compress-all`: 124M
- With this PR, with `-Xfatbin=-compress-all`: 44M **(new)**

Sounds like we can add it and proceed?

leofang:
Alright, I tested the whole change on conda-forge ([here](https://github.com/conda-forge/cupy-feedstock/pull/178)). Generally I am seeing slowdown of at least 5-10 mins on linux64 (best config: `CUPY_NUM_NVCC_THREADS` 1, `CUPY_NUM_BUILD_JOBS`: 2, likely because the free tier CI only has 2 vCPUs), and this is already native build, so the emulator builds (for aarch64/ppc64le) would only be worse. 

If the team still likes it I can clear the loose ends, but it seems splitting TUs would not help free CIs.

kmaehashi:
Thanks for testing, Leo. So compiling Thrust headers multiple times is heavier than expected... Although this PR improves the developer experience by improving build speed in multi-core env, I think it's better not to get this merged considering the situation of conda-forge build (#6873).

leofang:
> Similar inflation is observed with [this branch](https://github.com/leofang/cupy/tree/split_TU_cub2) applying the same strategy to `cupy.cuda.cub` (55M -> 119M).

For reference I pushed the only additional commit 20d6e09 here. 

> I think it's better not to get this merged considering the situation of conda-forge build (https://github.com/cupy/cupy/issues/6873).

Yes, thanks Kenichi-san. We can consider revisiting this PR once the conda-forge GPU CI is up (https://github.com/conda-forge/conda-forge.github.io/issues/1272), but right now there is no need to move forward.

PR #6942, opened 2022-08-01
Fix occasional hang in sparse distributed

### LINKED ISSUE #6916: CI: Need to install MPI for distributed tests

emcastillo:
Close #6916


emcastillo:
multigpu tests seems to be ok but jenkins!!

leofang:
btw this example can be tested now 😄
https://github.com/cupy/cupy/blob/master/examples/interoperability/mpi4py_multiple_devices.py

kmaehashi:
@emcastillo Could you briefly describe what was wrong with `cupyx/distributed/_nccl_comm.py`? :)

kmaehashi:
I've refined CI tooling part and rebased to the master (force pushed :bow:).

emcastillo:
Thanks!

The main issue was that the alltoall implementation for sparse matrices was causing nccl to hang in some cases due to the exchange of metadata interleaved with the exchange of the actual data. This was fixed once I split it into two phases.
Exchange the metadata first, then do the actual data exchange. It is weird that it didn't happen before though, since this was not an MPI test, but the new alltoall is stable enough now.

kmaehashi (approved):

PR #6953, opened 2022-08-02
Efficiency improvements in `cupyx.scipy.ndimage` utilities

grlee77:
This PR changes the `_get_output` utility to initialize output arrays to `empty` rather than `zeros` for efficiency. This utility is used by all filtering, morphology and interpolation functions in `ndimage`.

It also removes the need for a copy in the `_run_1d_filters` utility when the number of filters is even. This utility is used by functions like `gaussian_filter` and `uniform_filter` to apply a 1d filter along each axis of an array in turn.

These two changes give improvements on the order of 5-10% in some cases I tested.

A separate `_init_weights_dtype` utility was added and used in a few of the separable filters to avoid unnecessary double precision computation for the `sobel` and `prewitt` filters. On my system this gives > 2x improvement for these because previously the integer valued kernels would get promoted to `float64` even when the input was `float32`. I applied this utility to `gaussian_filter` and `uniform_filter` as well, but there the impact is much lower as the kernel values were already floating point and so were already being copied to a precision matching the input by `_get_weights_dtype` later on.


jakirkham (approved):

cupyx/scipy/ndimage/_filters.py
```diff
@@ -356,19 +361,20 @@ def gaussian_filter(input, sigma, order=0, output=None, mode="reflect",
...
             return None
-        return _gaussian_kernel1d(sigma, order, radius)
+        return _gaussian_kernel1d(sigma, order, radius, dtype=weights_dtype)
 
     return _run_1d_correlates(input, list(zip(sigmas, orders)), get, output,
                               mode, cval, 0)
 
 
-def _gaussian_kernel1d(sigma, order, radius):
+def _gaussian_kernel1d(sigma, order, radius, dtype=float):
```
takagi, line 377: nit: using numpy dtype
```suggestion
def _gaussian_kernel1d(sigma, order, radius, dtype=cupy.float64):
```

takagi (approved):
LGTM, thanks!

PR #6126, opened 2021-11-18
Docs: Update linkcode implementation

kmaehashi:
Unwrap functions wrapped by `contextmanager` to fix `[source]` deadlink, e.g.:

* https://docs.cupy.dev/en/v10.0.0rc1/reference/generated/cupy.cuda.using_allocator.html?highlight=[source]
* https://docs.cupy.dev/en/v10.0.0rc1/reference/generated/cupyx.optimizing.optimize.html?highlight=[source]

Add `[source]` link to objects defined in C-extensions, e.g.:

* https://docs.cupy.dev/en/v10.0.0rc1/reference/generated/cupy.cuda.Device.html

PR #6955, opened 2022-08-02
Fix docs build failure

kmaehashi:
This fixes the master branch doc build (LaTeX) broken.

https://readthedocs.org/projects/cupy/builds/

This `assert` check was improper as the filenames for extension objects are not existing filenames after #6126 (which manually crafts the PYX filename by removing `~.so` and appending `.pyx` from the real extension object file).
Depending on the module discovery order of Sphinx, this assert triggers and build fails.


asi1024:
doctest has passed!

PR #6960, opened 2022-08-03
Fix CUB reduction with zero-size arrays

### LINKED ISSUE #6937: CUB reduction mishandles zero-size arrays along axes with nonzero elements
### Description

Confirmed in upstream (https://github.com/numpy/numpy/issues/22063) that this is the intended behavior:
```python
import numpy as np
np.min(np.empty((0, 2)), axis=0)  # raise ValueError
np.min(np.empty((0, 2)), axis=1)  # does not raise
```
CuPy followed the same behavior, until v11.0 in which the CUB backend is turned on by default, and it raises unconditionally due to this check:
https://github.com/cupy/cupy/blob/9d439227f1a53bbd9ebb16784e40bcce44bc8042/cupy/cuda/cub.pyx#L163-L165

I am happy to help fix it. Feel free to assign it to me 🙂

### To Reproduce

_No response_

### Installation

_No response_

### Environment

_No response_

### Additional Information

cc: @shwina @jakirkham

leofang:
Close #6937.

Both `device_reduce` and `device_segmented_reduce` can now handle zero-size arrays correctly. In particular, if CUB is enabled we won't leave this case for the "default" backend to handle, so as to return the result as early as possible.

I also did a minor clean up of the tests to
1. combine two test classes into one
2.  modernize (ex: use pytest) and fix (ex: `rtol` is not needed for min/max tests, etc)

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -8,7 +8,7 @@ from libc.stdint cimport intptr_t
 from cupy_backends.cuda.api cimport runtime
 from cupy._core.core cimport _internal_ascontiguousarray
 from cupy._core.core cimport _internal_asfortranarray
-from cupy._core.internal cimport _contig_axes
+from cupy._core.internal cimport _contig_axes, is_in, prod
 from cupy.cuda cimport common
 from cupy.cuda cimport device
 from cupy.cuda cimport memory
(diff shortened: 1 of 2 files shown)


asi1024 (approved):

asi1024:
LGTM. Thanks!

PR #6961, opened 2022-08-03
Fix sparse test to follow SciPy 1.9

takagi:
https://github.com/scipy/scipy/issues/15210 is fixed in scipy==1.9.0.


takagi:
close as dup. #6894

PR #6911, opened 2022-07-26
Reorganize build scripts: define compile options declaratively

kmaehashi:
Part of #5722.

Currently compilation options for each module are adjusted in `preconfigure_modules` and `make_extensions` in adhoc way.
https://github.com/cupy/cupy/blob/master/install/cupy_builder/cupy_setup_build.py

This PR intends to allow defining options for each module in more declarative way.

PR #6975, opened 2022-08-05
Parallelize Cythonize

### LINKED ISSUE #5979: Support multi-threaded Cythonize (`nthreads` option)
### Description

In general, `cythonize` is reasonably fast as-is, but it can be slow when cross-compiling aarch64 binary on x86_64 host. I think this option can improve the build speed by utilizing multiple cores on host.

### Additional Information

_No response_

kmaehashi:
Closes #5979. Depends on ~#6911~.

This PR improves cythonize time, which takes ~0.5 minutes without parallelization.

This is implemented by passing `nthreads` to `cythonize`:
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#cythonize-arguments


leofang:
Question: Is `python setup.py build_ext -j N` expected to work (even before this PR)? I don't need this support, just curious 🙂

kmaehashi:
> Question: Is `python setup.py build_ext -j N` expected to work (even before this PR)? I don't need this support, just curious 🙂

No, because we overwrite `self.parallel` (which is set via `build_ext -j`) with our own env var to support parallel build through pip (which does not provide `-j`).

https://github.com/cupy/cupy/blob/ce80889721b77b9c2b8c02b7373f09b9b601275a/install/cupy_builder/_command.py#L77-L79

leofang:
Yup that's what I figured, thanks for confirming!

leofang:
btw this is wrong right?
https://github.com/cupy/cupy/blob/ce80889721b77b9c2b8c02b7373f09b9b601275a/install/cupy_builder/_context.py#L35-L36

kmaehashi:
Windows (or multiprocessing with spawn) didn't work with our installer which depends on many global variables so disabled.

/test

kmaehashi:
> btw this is wrong right?
> 
> https://github.com/cupy/cupy/blob/ce80889721b77b9c2b8c02b7373f09b9b601275a/install/cupy_builder/_context.py#L35-L36

Ah you're right, `annotate=True` generates HTML to investigate Python overhead in Cython code, so it's harmless but ideally should be a different option. Seems it's there since v1.0.0 😅 
https://github.com/cupy/cupy/blob/v1.0.0/cupy_setup_build.py#L296-L297

leofang:
It's amazing that a bug survived multiple refactoring!

kmaehashi:
> It's amazing that a bug survived multiple refactoring!

Refactoring shouldn't change the behavior 😄

PR #6976, opened 2022-08-05
Make sparse argmin/max return a scalar array containing the index

### LINKED ISSUE #3640: Sparse argmin/argmax return scalar values on CPU
CuPy returns zero-dimensional ndarrays for reduction functions where NumPy/SciPy return scalar values to keep off synchronizing between GPU and CPU. However, CuPy's sparse matrix argmin/argmax currently return scalar values as SciPy's counterparts, causing device to host sync.

```Python
>>> import cupy
>>> import cupyx
>>> a = cupy.array([[0,1,0],[2,0,3]], dtype='f')
>>> cupyx.scipy.sparse.csc_matrix(a).argmin()
0
```

andfoy:
Fixes #3640

The aforementioned issue only happened when `argmax/min` was called without an `axis` defined. Now `argmin/max` returns a scalar array with the index.

### DIFF
--- a/cupyx/scipy/sparse/_data.py
+++ b/cupyx/scipy/sparse/_data.py
@@ -123,16 +123,10 @@ class _data_matrix(_base.spmatrix):
 def _find_missing_index(ind, n):
     positions = cupy.arange(ind.size)
     diff = ind != positions
-    non_zero = cupy.argwhere(diff).ravel()
-
-    if non_zero.size != 0:
-        return non_zero[0]
-
-    k = ind.size + 1
-    if k < n:
-        return cupy.asarray(k)
-    else:
-        return cupy.asarray(-1)
+    return cupy.where(
+        diff.any(),
+        diff.argmax(),
+        cupy.asarray(ind.size if ind.size < n else -1))
 
 
 class _minmax_mixin(object):
--- a/cupyx/scipy/sparse/_data.py
+++ b/cupyx/scipy/sparse/_data.py
@@ -139,7 +139,7 @@ def _non_zero_cmp(mat, am, zero, m):
         zero_ind = _find_missing_index(ind, size)
         return cupy.where(
             m == zero,
-            cupy.where(cupy.greater(zero_ind, am), am, zero_ind),
+            cupy.minimum(zero_ind, am)
             zero_ind)
 
 
--- a/cupyx/scipy/sparse/_data.py
+++ b/cupyx/scipy/sparse/_data.py
@@ -1,6 +1,5 @@
-from operator import mul
-from functools import reduce
 import cupy
+import numpy as np
 from cupy._core import internal
 from cupy import _util
 from cupyx.scipy.sparse import _base
@@ -132,7 +131,7 @@ def _find_missing_index(ind, n):
 
 
 def _non_zero_cmp(mat, am, zero, m):
-    size = reduce(mul, mat.shape, 1)
+    size = np.prod(mat.shape)
     if size == mat.nnz:
         return am
     else:
--- a/cupyx/scipy/sparse/_data.py
+++ b/cupyx/scipy/sparse/_data.py
@@ -130,6 +130,19 @@ def _find_missing_index(ind, n):
         cupy.asarray(ind.size if ind.size < n else -1))
 
 
+def _non_zero_cmp(mat, am, zero, m):
+    size = math.prod(mat.shape)
+    if size == mat.nnz:
+        return am
+    else:
+        ind = mat.row * mat.shape[1] + mat.col
+        zero_ind = _find_missing_index(ind, size)
+        return cupy.where(
+            m == zero,
+            cupy.where(cupy.greater(zero_ind, am), am, zero_ind),
+            zero_ind)
+
+
 class _minmax_mixin(object):
     """Mixin for min and max methods.
     These are not implemented for dia_matrix, hence the separate class.
@@ -233,23 +246,9 @@ class _minmax_mixin(object):
                 am = op(mat.data)
                 m = mat.data[am]
 
-                zero_cmp = cupy.where(compare(m, zero), True, False)
-
-                if zero_cmp:
-                    return mat.row[am] * mat.shape[1] + mat.col[am]
-                else:
-                    size = math.prod(mat.shape)
-                    if size == mat.nnz:
-                        return am
-                    else:
-                        ind = mat.row * mat.shape[1] + mat.col
-                        zero_ind = _find_missing_index(ind, size)
-                        if m == zero:
-                            min_idx = cupy.where(
-                                cupy.greater(zero_ind, am), am, zero_ind)
-                            return min_idx
-                        else:
-                            return zero_ind
+                return cupy.where(
+                    compare(m, zero), mat.row[am] * mat.shape[1] + mat.col[am],
+                    _non_zero_cmp(mat, am, zero, m))
 
         if axis < 0:
             axis += 2
--- a/cupyx/scipy/sparse/_data.py
+++ b/cupyx/scipy/sparse/_data.py
@@ -225,7 +225,7 @@ class _minmax_mixin(object):
             if self.nnz == 0:
                 return 0
             else:
-                zero = self.dtype.type(0)
+                zero = cupy.asarray(self.dtype.type(0))
                 mat = self.tocoo()
 
                 mat.sum_duplicates()
@@ -243,7 +243,7 @@ class _minmax_mixin(object):
                         ind = mat.row * mat.shape[1] + mat.col
                         zero_ind = _find_missing_index(ind, size)
                         if m == zero:
-                            return min(zero_ind, am)
+                            return cupy.min(zero_ind, am)
                         else:
                             return zero_ind
 
--- a/cupyx/scipy/sparse/_data.py
+++ b/cupyx/scipy/sparse/_data.py
@@ -139,7 +139,7 @@ def _non_zero_cmp(mat, am, zero, m):
         zero_ind = _find_missing_index(ind, size)
         return cupy.where(
             m == zero,
-            cupy.minimum(zero_ind, am)
+            cupy.minimum(zero_ind, am),
             zero_ind)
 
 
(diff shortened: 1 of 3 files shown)


cupyx/scipy/sparse/_data.py
```diff
@@ -238,7 +238,7 @@ def _arg_min_or_max(self, axis, out, op, compare):
                 if compare(m, zero):
                     return mat.row[am] * mat.shape[1] + mat.col[am]
                 else:
-                    size = cupy.prod(mat.shape)
+                    size = cupy.prod(cupy.asarray(mat.shape))
```
asi1024, line 241: How about the following simplification?
```suggestion
                    size = mat.size
```

cupyx/scipy/sparse/_data.py
```diff
@@ -123,13 +123,13 @@ def power(self, n, dtype=None):
 def _find_missing_index(ind, n):
     for k, a in enumerate(ind):
         if k != a:
```
asi1024, line 125: It still looks that device-to-host sync is caused here.
How about comparing `ind` and `cupy.arange(ind.size)` and finding the minimum index `i` such that `ind[i] != i`, without synchronization?

cupyx/scipy/sparse/_data.py
andfoy: I think `size` for a sparse matrix refers to the number of non-zero entries, which is different from the total elements that would be on its dense version

leofang: I am not sure what @asi1024 had in mind, but I think the variable `size` is needed on CPU for conditional branches, so it does not make sense to compute and keep it on GPU. By the time we need to do `if size == mat.nnz` it's moved back to CPU again. (Though apparently the mistake was done earlier, not in this PR 😅)

asi1024: Yes, I want to keep `size` in CPU memory. How about using `numpy.prod` or `math.prod` instead?

cupyx/scipy/sparse/_data.py
```diff
@@ -121,15 +121,18 @@ def power(self, n, dtype=None):
 
 
 def _find_missing_index(ind, n):
-    for k, a in enumerate(ind):
-        if k != a:
-            return k
+    positions = cupy.arange(ind.size)
+    diff = ind != positions
+    non_zero = cupy.argwhere(diff).ravel()
```
asi1024, line 126: `cupy.argwhere` also causes device-to-host sync because the length of return values varies depending on the value of elements of the inputs.

How about the following implementation?

```py
return cupy.where(
    diff.any(),
    diff.argmax(),
    cupy.asarray(ind.size if ind.size < n else -1))
```
```

**Comment by asi1024** [comment]
@andfoy Could you remove `pytest.skip` in test_csc.py and test_csr.py?
https://github.com/cupy/cupy/blob/011a38eb1f5e8ac2a8e2e430e041e96bb6557c02/tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py#L1263-L1265

**Comment by andfoy** [comment]
Some of the errors are related to the return type mismatch between NumPy and CuPy. Since now `argmin/max` is returning an array, it voids the int return type produced by NumPy

File: tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py [hunk 4]
```diff
@@ -1260,17 +1260,11 @@ def test_max_explicit(self, xp, sp):
 
     @testing.numpy_cupy_array_equal(sp_name='sp')
     def test_argmin(self, xp, sp):
-        # TODO(takagi) Fix axis=None
-        if self.axis is None:
-            pytest.skip()
         data = self._make_data_min(xp, sp, dense=self.dense)
         return data.argmin(axis=self.axis)
```
asi1024, line 1264: Could you try the following fix to avoid a type mismatch?
```suggestion
        return xp.array(data.argmin(axis=self.axis))
```

asi1024: Could you also fix other tests?

asi1024:
Could you check CI failures?

cupyx/scipy/sparse/_data.py
```diff
@@ -238,7 +236,7 @@ def _arg_min_or_max(self, axis, out, op, compare):
                 if compare(m, zero):
                     return mat.row[am] * mat.shape[1] + mat.col[am]
                 else:
-                    size = cupy.prod(mat.shape)
+                    size = math.prod(mat.shape)
                     if size == mat.nnz:
                         return am
```
andfoy, line 241: The test failures are related to this path, which was unchanged on this PR. Unless there's a difference between `cupy.prod` and `math.prod`, I'm not sure what should be the resolution here

andfoy: @asi1024, it is not clear how the changes introduced on this PR could've modified something outside? In particular, the test failure is not related to the particular changes

asi1024: Hmm, this test failure seems to be due to a SciPy bug...

```py
>>> x = scipy.sparse.coo_matrix(([1, 0.1, 1, 1], ([0, 0, 1, 1], [0, 1, 0, 1])), shape=(2, 2)).tocsc()
>>> x.todense()
matrix([[1. , 0.1],
        [1. , 1. ]])
>>> x.argmin()
2
>>> x.todense().argmin()
1
```

Could you restore the test skip and leave a comment about it in the test?

asi1024:
There still seems to be device-to-host data synchronization in the following lines.

https://github.com/cupy/cupy/blob/87d008df13be282471b77bf2494f316d849784d5/cupyx/scipy/sparse/_data.py#L236

https://github.com/cupy/cupy/blob/87d008df13be282471b77bf2494f316d849784d5/cupyx/scipy/sparse/_data.py#L246

cupyx/scipy/sparse/_data.py
```diff
@@ -227,7 +225,7 @@ def _arg_min_or_max(self, axis, out, op, compare):
             if self.nnz == 0:
                 return 0
             else:
-                zero = self.dtype.type(0)
+                zero = cupy.asarray(self.dtype.type(0))
```
andfoy, line 228: Is this enough to prevent synchronization?

asi1024: No, the result of `compare(m, zero)` is a `cupy.ndarray`, and it will be converted to a CPU bool for use as a condition of if-statement. So, we need to branch on the GPU process by using `cupy.which`, just as we implement in `_find_missing_index`.

cupyx/scipy/sparse/_data.py
```diff
@@ -227,25 +225,29 @@ def _arg_min_or_max(self, axis, out, op, compare):
...
 
                 mat.sum_duplicates()
 
                 am = op(mat.data)
                 m = mat.data[am]
 
-                if compare(m, zero):
+                zero_cmp = cupy.where(compare(m, zero), True, False)
+
+                if zero_cmp:
```
andfoy, line 238: Is this accurate?

cupyx/scipy/sparse/_data.py
```diff
@@ -227,25 +225,29 @@ def _arg_min_or_max(self, axis, out, op, compare):
...
                     if size == mat.nnz:
                         return am
                     else:
                         ind = mat.row * mat.shape[1] + mat.col
                         zero_ind = _find_missing_index(ind, size)
                         if m == zero:
-                            return min(zero_ind, am)
+                            min_idx = cupy.where(
+                                cupy.greater(zero_ind, am), am, zero_ind)
+                            return min_idx
```
andfoy, line 250: As well this?

andfoy (commented):
@asi1024, thanks for the latest review pointers, I've just replaced all instances of comparisons to use `cupy.where` instead. Please let me know if the changes are accurate

cupyx/scipy/sparse/_data.py
asi1024: The element of `zero_cmp` is stored on GPU memory, and it is transferred to CPU memory for Python code path branching here. My suggestion is to use `cupy.where` to branch a code path in CUDA!

asi1024:
Could you check CI failure?

andfoy:
@asi1024, I think everything is ready to run on the actual CIs

andfoy:
For some reason, the Python version on the CI didn't declare `math.prod`, it was replaced by `operator.mul` used with `functools.reduce`. I think it should pass this time on the CIs

asi1024:
How about using `numpy.prod` instead?

asi1024:
The CI seems to fail due to dtype mismatch. Could you check the following test log?
https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/111699/cupy_test_log.txt

cupyx/scipy/sparse/_data.py
andfoy: @asi1024, it seems that this is the only place where a zero of a certain dtype is being defined, my guess here is that the test inputs are `int64`, which produces `int64` outputs as well. Do you want to restrict the output to `int32`?

asi1024: Do you mean this failure is due to a bug in SciPy? If so, I think it is OK to skip type check by feeding `type_check=False` to the testing decorator.
https://docs.cupy.dev/en/stable/reference/generated/cupy.testing.numpy_cupy_array_equal.html#cupy.testing.numpy_cupy_array_equal

andfoy: In Windows, calls to `argmax` on a SciPy sparse matrix return a scalar number, as opposed to an `np.array`:

```
>>> a = np.random.randint(2, size=(10, 20))
>>> x = sp.csc_matrix(a)
>>> x.dtype
dtype('int32')
>>> x.argmax(axis=None)
0
>>> x.astype(np.int64).argmax(axis=None)
0
```

andfoy: So, we can add `type_check=False`

asi1024 (approved):

asi1024:
@andfoy LGTM! Thanks for your contribution!

PR #6989, opened 2022-08-15
Don't use `__del__` in `TCPStore`

### LINKED ISSUE #6678: Exception raised in `TCPStore.__del__` upon process termination
### Description

`__del__` should not perform syncronization when called during process termination.

```
Exception ignored in: <function TCPStore.__del__ at 0x7fb939be23a0>
  File "/home/maehashi/Development/cupy/cupyx/distributed/_store.py", line 49, in __del__
  File "/home/maehashi/Development/cupy/cupyx/distributed/_store.py", line 97, in stop
  File "/home/maehashi/Development/cupy/cupyx/distributed/_store.py", line 31, in join
  File "/home/maehashi/.pyenv/versions/3.8.1/lib/python3.8/multiprocessing/connection.py", line 251, in recv
ModuleNotFoundError: import of builtins halted; None in sys.modules
```

### To Reproduce

_No response_

### Installation

_No response_

### Environment

_No response_

### Additional Information

_No response_

emcastillo:
Fixes #6678

### DIFF
--- a/cupyx/distributed/_store.py
+++ b/cupyx/distributed/_store.py
@@ -54,6 +54,10 @@ class TCPStore:
         self._lock = threading.Lock()
         self._current_barrier = None
 
+    def __del__(self):
+        if not _exit_mode:
+            self.stop()
+
     def _set_process(self, process):
         self._process = process
 
--- a/cupyx/distributed/_store.py
+++ b/cupyx/distributed/_store.py
@@ -1,3 +1,4 @@
+import atexit
 from ctypes import sizeof
 import multiprocessing
 import threading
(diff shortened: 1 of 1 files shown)


cupyx/distributed/_store.py
```diff
@@ -45,9 +54,6 @@ def __init__(self, world_size):
         self._lock = threading.Lock()
         self._current_barrier = None
 
-    def __del__(self):
-        self.stop()
```
kmaehashi, line 49: How about keeping `__del__`, and call `stop()` only when `_exit_mode` is not set? I think it's better to destroy the process once the communicator goes out of scope.

emcastillo: done :)

kmaehashi (approved):

PR #7001, opened 2022-08-17
JIT: Fix `BuiltinFunc` not to inherit `Expr` class

asi1024:
Cosmetic changes.

### DIFF
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -1,6 +1,5 @@
 import functools
-import itertools
-from typing import Any, Mapping, NoReturn, TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


leofang:
Q: With this change, is there any semantics change for `Expr` or `BuiltinFunc`?

asi1024:
`BuiltinFunc` will be handled as `Constnant`, not `Expr` after this PR, but it is just an internal change. There is no need to modify the definition of each builtin-function.

asi1024:
I have totally forgotten that a `Constant` class object of `Function` needs to have CUDA variables whose values are not determined at compile time unless `BuiltinFunc` inherits Expr. Please let me withdraw this pull request...

PR #7005, opened 2022-08-18
Improve installation command

kmaehashi:
I saw several users confusing with `cupy-cuda11x` and `cupy-cuda112...117`. This PR adds some more description of the change in CuPy v11.


leofang:
What's the plan/timeline to mention `cupy-wheel`? 🙂 I think it'd help resolve further confusion.

Issue #6229, opened 2021-12-09
Support more interpolation methods in `cupy.percentile` and `cupy.quantile`


takagi:
Since NumPy 1.22, `numpy.percentile` and `numpy.quantile` have new methods to estimate the percentile/quantile. 

https://numpy.org/devdocs/reference/generated/numpy.quantile.html
https://numpy.org/devdocs/reference/generated/numpy.percentile.html

gajendra0180:
Hey, Can you specify what actually needs to be done  here?
Thanks.

kmaehashi:
I've added the checklist 😃

Ashish9836:
@gajendra0180  Hey Gajendra! , Can i work on this issue?

gajendra0180:
Ya Sure,Go ahead.

Ashish9836:
@gajendra0180  Thanks!

pri1311:
So I was working on this, and I started off with `interpolated_inverted_cdf`. 
I added the basic indices math needed and it passed most the tests, but I keep getting `nan location mismatch error` for 2 of the test cases.
```
FAILED tests/cupy_tests/statistics_tests/test_order.py::TestQuantileMethods::test_quantile_out[interpolated_inverted_cdf]
FAILED tests/cupy_tests/statistics_tests/test_order.py::TestQuantileMethods::test_quantile_overwrite[interpolated_inverted_cdf]
```
```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-06, atol=0
E       
E       x and y nan location mismatch:
E        x: array([[[[5.650563, 5.505339, 4.453514],
E                [4.81928 , 4.400868, 5.545542]],
E       ...
E        y: array([[[[5.650563, 5.505339, 4.453514],
E                [4.81928 , 4.400868, 5.545542]],
E       ...

cupy\testing\_array.py:24: AssertionError
```

Could you help me by pointing out where I could be wrong?
```
 elif method == 'interpolated_inverted_cdf':
        indices = q * (Nx) - 1
```

takagi:
Hi @pri1311, can I see your branch?

pri1311:
@takagi 
commit - https://github.com/cupy/cupy/compare/master...pri1311:cupy:interpolation

Branch - https://github.com/pri1311/cupy/tree/interpolation

takagi:
Thanks @pri1311! Would you make a PR for your change? I'll review it next week as I have days off this week.

[...]

PR #7007, opened 2022-08-19
add support for `interpolated_inverted_cdf`

pri1311:
Note: Does not pass all tests. Created the PR for review to figure out the bug.

[linked] closes: #6229


takagi:
I reproduced the problem locally. I can't immediately find the cause and need to investigate it.

takagi:
sorry, it may take so much before I have time to see this problem in detail. would you have other issues if you're interested in?

PR #7008, opened 2022-08-22
Support cuDNN 8.5

takagi:


kmaehashi:
Could you remove 11.7~11.2 records as well? It seems the download site has some WAF enabled recently and test failing maybe due to burst HEAD requests from test code.

takagi:
(also removed cutensor's)

kmaehashi (approved):

Issue #6011, opened 2021-11-05
[RFC] Move CUDA high-level wrappers to `cupyx`?


kmaehashi:
### Description

Historically, CUDA high-level wrappers (`cublas`, `cudnn`, `cusolver`, `cusparse`, `cutensor`) are placed under `cupy.*` namespace without explicit documentation.
Recently we have been cleaning up `cupy.*` namespace to be consistent with `numpy.*`, so it would also be better to move these modules to `cupyx`.

PR #7013, opened 2022-08-26
Move `cupy.cudnn` `cupy.cusolver` `cupy.cutensor` `cupy.cusparse` to `cupyx`

emcastillo:
Close [#6011](https://github.com/cupy/cupy/issues/6011)

Should we keep aliases?


asi1024:
@emcastillo Could you split the test fixes into another PR, to check this change keeps the conventional interface?

emcastillo:
> Could you split the test fixes into another PR, to check this change keeps the conventional interface?

I feel the deprecation warning will make all the tests fails.

Should we have this PR without the warning & tests, and a follow-up one with the warning and the tests?

asi1024:
@emcastillo I thought that we should separate this PR and run CIs for each PR to ensure that this change does not break the conventional APIs. However, Now I found that it is easy to confirm it by reviewing it, so no need to separate the PR!

asi1024 (approved):

Issue #6324, opened 2022-01-17
[Tracker] Implement all `scipy.*` APIs in CuPy


kmaehashi:
Implement GPU version of `scipy.*` functions in `cupyx.scipy.*` namespace.

This is a tracker issue that summarizes the implementation status of each SciPy public modules in CuPy.
See the [comparison table](https://docs.cupy.dev/en/stable/reference/comparison.html#scipy-cupy-apis) for details.

## List of Modules

- 🥚 Clustering package (`scipy.cluster`)
    - 🥚 K-means clustering and vector quantization (`scipy.cluster.vq`) (📌 planned in #5947)
    - 🥚 Hierarchical clustering (`scipy.cluster.hierarchy`)
- 🥚 Constants (`scipy.constants`)
- :shipit: Discrete Fourier transforms (`scipy.fft`)
- :shipit: Legacy discrete Fourier transforms (`scipy.fftpack`) (📌 note: deprecated, no further development planned)
- 🥚 Integration and ODEs (`scipy.integrate`)
- 🥚 Interpolation (`scipy.interpolate`)
- 🥚 Input and output (`scipy.io`)
    - 🥚 `scipy.io.arff`
    - 🥚 `scipy.io.matlab`
    - 🥚 `scipy.io.wavfile`
- 🐥 Linear algebra (`scipy.linalg`)
    - 🥚 Low-level BLAS functions (`scipy.linalg.blas`)
    - 🥚 `scipy.linalg.cython_blas` (📌 note: unlikely supported in CuPy)
    - 🥚 Low-level LAPACK functions (`scipy.linalg.lapack`)
    - 🥚 `scipy.linalg.cython_lapack`
    - 🥚 Interpolative matrix decomposition (`scipy.linalg.interpolative`)
- 🥚 Miscellaneous routines (`scipy.misc`)
- :shipit: Multidimensional image processing (`scipy.ndimage`)
- 🥚 Orthogonal distance regression (`scipy.odr`)
- 🥚 Optimization and root finding (`scipy.optimize`) (📌 planned in #6112)
- 🥚 Signal processing (`scipy.signal`)
    - 🥚 `scipy.signal.windows`
- :shipit: Sparse matrices (`scipy.sparse`)
    - :shipit: Sparse linear algebra (`scipy.sparse.linalg`)
    - 🐥 Compressed sparse graph routines (`scipy.sparse.csgraph`) (📌 planned in #2431)
[...]

Issue #4831, opened 2021-03-08
Introduce Type Annotations in CuPy APIs


kmaehashi:
c.f. https://numpy.org/devdocs/reference/typing.html

leofang:
btw Array API (#4789) also requires typed APIs, but since it's contained in a standalone namespace, it's much easier to address that CuPy itself.

kmaehashi:
Before adding type annotations we have to define a policy on arguments (e.g., which kind of arguments should or should not accept numpy/cupy/array-like/tuple/list/...)

Here is a list of past issues regarding argument types:

https://github.com/cupy/cupy/issues/1206
https://github.com/cupy/cupy/issues/1427
https://github.com/cupy/cupy/issues/1529
https://github.com/cupy/cupy/issues/3511
https://github.com/cupy/cupy/issues/3512
https://github.com/cupy/cupy/issues/4466
https://github.com/cupy/cupy/issues/4476
https://github.com/cupy/cupy/issues/5070
https://github.com/cupy/cupy/issues/5117
#5822
#7596
#8659
#9367

kmaehashi:
I've talked with @asi1024 and I'm thinking the steps as follow:

1. Add type annotations for public APIs. Also prepare a set of python scripts calling CuPy public APIs to test the correctness of the annotations. This immedeately benefit users with typed codebase.
2. Add type annotations for all CuPy APIs, so that `mypy` passes against CuPy codebase itself.
3. Support `numpy.typing` features, e.g., make `mypy` aware of shapes and dtypes.

Each step can be done incrementally. We can also try tools listed [here](https://mypy.readthedocs.io/en/stable/existing_code.html#automate-annotation-of-legacy-code) to semi-automate the process.

For Python code, I prefer to have them defined inline rather than maintaining separate `pyi` files. We may need to switch to `pyi` files when doing step 3, but the migration should be easily doable with mypy's `stubgen` command. For Cython code, I think the only choice is to create `pyi` files and declare type annotation for public APIs.

GF-Huang:
Any progress?

asi1024:
We plan to support type annotations after introducing support for NumPy 2.0.

GF-Huang:
This [issue](https://github.com/microsoft/pyright/issues/9051#issuecomment-2368944137) may be related to this one.

jorenham:
[...]

PR #7014, opened 2022-08-26
Add `scipy.special.expi` and `scipy.special.exp1`

1MrEnot:
This PR adds support for `expi` and `exp1` special functions from scipy
Uses original scipy algorithm for both [expi](https://github.com/scipy/scipy/blob/main/scipy/special/specfun/specfun.f#L5250) and [exp1](https://github.com/scipy/scipy/blob/main/scipy/special/specfun/specfun.f#L5317), but only for real arguments.
There're two functions in one PR as `expi` uses `exp1`, if real argument is less than 0.

P.S.
Yes, it's recommended first to open an issue and discuss the design first, but i have no idea what to discuss: functions must provide the same API and return the same results as scipy ones...


1MrEnot:
Also I experimented with template implementation and faced ambigous results

On the one hand templates significantly improves performance (GPU time only):
| Function| float32 | float64|
| :---         | :---       | :---      |
| Expi| 837.124 us   +/-32.309 | 3131.325 us   +/-13.002 |
| Exp1| 791.881 us   +/-34.616 | 1773.603 us   +/-15.066 |

But on the onther hand there's small range of float16 and float32 values, for which scipy returns concrete value, but cupy implementation returns +inf. For exmaple, scipy.special.expi(90) == 1.24106E+37, but cupyx.scipy.special.expi(90) == +inf. Personaly I dont think it's not critical, as we're talikng about 1e37

cupyx/scipy/special/_exp1.py
```diff
@@ -0,0 +1,65 @@
...
+
+from cupy import _core  # NOQA
+
+
+math_constants_and_eul = """
+#include <cupy/math_constants.h>
+__constant__ double EUL = 0.5772156649015328;
+"""
+
+exp1_defenition = """
```
emcastillo, line 12:
```suggestion
exp1_definition = """
```
Super stupid nit :)

cupyx/scipy/special/_exp1.py
```diff
@@ -0,0 +1,65 @@
...
+{
+    if (x == 0) {
+        return CUDART_INF;
+    } else if (x <= 1) {
+        T e1 = 1.0;
+        T R = 1.0;
+
+        for(int k=1; k<=25; k++)
+        {
+            int den = (k+1)*(k+1);
```
emcastillo, line 24:
```suggestion
            int den = (k + 1) * (k + 1);
```
let's add spaces between operators like this :)

emcastillo (commented):
This looks really great, thanks!!
I just left some minor comments

cupyx/scipy/special/_exp1.py
1MrEnot: Done!

cupyx/scipy/special/_exp1.py
1MrEnot: Fixed, thanks!

cupyx/scipy/special/_exp1.py
```diff
@@ -0,0 +1,65 @@
...
+    if (x == 0) {
+        return CUDART_INF;
+    } else if (x <= 1) {
+        T e1 = 1.0;
+        T R = 1.0;
+
+        for(int k=1; k<=25; k++)
+        {
+            int den = (k + 1) * (k + 1);
+            R = -R*k*x/den;
```
emcastillo, line 25:
```suggestion
            R = -R * k * x / den;
```

Please keep the format consistent in all the lines & files

kmaehashi:
Thanks! cross-linking to #6324

tests/cupyx_tests/scipy_tests/special_tests/test_exp1.py
```diff
@@ -0,0 +1,41 @@
+import cupyx.scipy.special  # NOQA
+
+from cupy import testing
+from cupy.testing import numpy_cupy_allclose
+
+
+@testing.gpu
+@testing.with_requires("scipy")
+class TestExp1:
+
```
khushi-411, line 10: Hi, just a minor note. I think `@testing.gpu` decorator can be removed as it was conventionally used. Right, @emcastillo!?

tests/cupyx_tests/scipy_tests/special_tests/test_expi.py
```diff
@@ -0,0 +1,41 @@
+import cupyx.scipy.special  # NOQA
+
+from cupy import testing
+from cupy.testing import numpy_cupy_allclose
+
+
+@testing.gpu
+@testing.with_requires("scipy")
+class TestExpi:
+
```
khushi-411, line 10: ditto

tests/cupyx_tests/scipy_tests/special_tests/test_expi.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_expi_positive(self, xp, scp):
+        return scp.special.expi(4.2)
+
+    @numpy_cupy_allclose(scipy_name="scp")
+    def test_expi_large_positive(self, xp, scp):
+        return scp.special.expi(1000)
+
+    @testing.for_dtypes(["f"])
+    @numpy_cupy_allclose(scipy_name="scp", rtol=1e-5)
+    def test_expi_linspace_float32(self, xp, scp, dtype):
```
khushi-411, line 33: nit: I think you don't need to use the list argument in the `for_dtypes()` decorator...

khushi-411 (commented):
Hi, just thought to point out some minor points. It's a nice PR. Thanks!

tests/cupyx_tests/scipy_tests/special_tests/test_expi.py
1MrEnot: Yes, it's not really neccessary, added list just to supress dumb PyCharm warning.
image
Removed!
P.S. Maybe, we should add "Iterable" type annotation to `for_dtypes` function? There's an issue about type annotations #4831. Ofc, just for cuntributiors, not library users

cupyx/scipy/special/_exp1.py
1MrEnot: Oh, sorry, my bad, didnt think about it. Fixed!
Is there any guide or maybe formatter, that you would recommend for c++ code?
I used [this](https://codebeautify.org/cpp-formatter-beautifier) formatter

tests/cupyx_tests/scipy_tests/special_tests/test_exp1.py
1MrEnot: With this attrubute pytest deselects these tests if we want to run tests without GPU like this `py -m pytest tests/cupyx_tests/scipy_tests/special_tests/test_expi.py -m='not gpu'` . After I removed `@testing.gpu` the same command run these tests. As for me it seems like correct behavour, but I may be wrong

emcastillo: Yeah, we actually disabled the `not gpu` marker since merely import cupy requires a GPU, so it is deprecated now :).

1MrEnot: So, should I remove @testing.gpu and add nothing new to imports?

emcastillo: thats it!

1MrEnot: done!

Issue #7003, opened 2022-08-18
Remove use of `pkg_resources`


kmaehashi:
Use of `pkg_resources` (part of `setuptools`) is now discouraged.

https://setuptools.pypa.io/en/latest/pkg_resources.html

> Use of pkg_resources is discouraged in favor of [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html), and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Please consider using those libraries instead of pkg_resources.

jakirkham:
This appears to be a duplicate of issue: https://github.com/cupy/cupy/issues/7654

Should we consolidate these somehow?

PR #7025, opened 2022-08-30
Load cuTENSOR from wheel distribution

### LINKED ISSUE #6999: Support cuTENSOR wheel
This issue tracks two aspects of the cuTENSOR wheel:
1. The installer (`cupyx.tools.install_library`) can take advantage of the cuTENSOR wheel
   - ideally we wanna do `pip install cutensor` so that `pip` is made aware of it
   - potential issue:
     - the `cutensor` wheel is very large (~300 MB as of 1.6.0) as it contains 3 versions (10.2, 11.0, 11), we are internally discussing if we can split it
2. The runtime loader (`cupy._environment._preload_library('cutensor')`) should first check from site-packages:
   - This issue is more urgent as we (cuQuantum) noticed wheel users can end up loading two copies of `libcutensor.so`, causing potentially hard-to-debug problems.

@kmaehashi I'd like to get the second aspect addressed asap.

kmaehashi:
Part of #6999.


cupy/_environment.py
```diff
@@ -369,6 +375,23 @@ def _preload_library(lib):
                 _log(f'Library {lib} could not be preloaded: {e}')
 
 
+def _get_cutensor_from_wheel(
+        version: str, cuda: str) -> List[str]:
+    """
+    Returns the list of shared library path candidates.
+    """
+    import pkg_resources  # defer import # NOQA
```
leofang, line 383: With #2759 and #7003, perhaps we should not rely on `pkg_resources`?

cupy/_environment.py
```diff
@@ -369,6 +375,23 @@ def _preload_library(lib):
...
+    try:
+        dist = pkg_resources.get_distribution(f'cutensor=={version}.*')
+    except pkg_resources.ResolutionError as e:
+        _log(f'cuTENSOR wheel could not be loaded: {type(e).__name__}: {e}')
+        return []
+    return [os.path.join(
+                dist.module_path, 'cutensor', 'lib', cuda_ver_path,
```
leofang, line 390: Q: In the case that we split the `cutensor` wheel into per-CUDA version, would you like us to keep this directory structure?

leofang: (Another note on the directory structure is it seems most of CUDA wheels are installed to the `nvidia` folder, but `cutensor` has not followed because no one asked.)

leofang (commented):
Thanks, Kenichi-san. Left a few quick notes.

cupy/_environment.py
kmaehashi: I think it would be convenient if Python APIs to discover bundled files are added to cutensor package (`cutensor.get_include_dir()`, `cutensor.get_lib_dir()`, `cutensor.get_lib_files()` etc.)
For example, [certifi](https://pypi.org/project/certifi/), which provides a root CA certificate for TLS, contains a PEM file along with the `certifi.where()` API that returns the path to the file.  

`cutensor.__version__` will be also helpful. If this is implemented, we can check cuTENSOR version on our own, without using `pkg_resources` or `importlib.metadata`.

kmaehashi: > Q: In the case that we split the cutensor wheel into per-CUDA version, would you like us to keep this directory structure?

I don't have a specific requirement (we can just update the code 😄), but  `lib/libcutensor.so.1` sounds more natural to me than `lib/11/libcutensor.so.1`.

leofang: Thanks for feedback. I will discuss with both cuQuantum and cuTENSOR teams and investigate alternative possibilities. 

We implemented exactly what you asked for in cuQuantum (internal repo), but it lives in the `cuquantum-python` wheel, not in the `cuquantum` wheel. The reason is that there is ambiguity in the behavior of your suggested APIs, so we had to disambiguate by letting Python do the import (during which the dynamic linker would load `libcutensor.so`, `libcutensornet.so`, etc), and inspect the linker's internal state. This ensures regardless of how cuTENSOR etc are installed (wheel, conda, tarball, etc), users calling these APIs will always get the correct path.

Obviously this only works on Linux, but for cuQuantum Python it is sufficient, and we can cover cuTENSOR as well. The other drawback is obviously this capability must stay in the Python binding, not with the binary-repackaged wheel. Thoughts are more than welcome...

leofang: As a temporary solution until we land on something for the cuTENSOR wheel, I am thinking here you can just inspect `site.getsitepackages()` and `site.getusersitepackages()`. If cuTENSOR wheel is installed, it must live in one of the site-packages.

leofang: Update:
> We implemented exactly what you asked for in cuQuantum (internal repo), but it lives in the `cuquantum-python` wheel, not in the `cuquantum` wheel. 

See https://github.com/NVIDIA/cuQuantum/blob/main/python/cuquantum/__main__.py.

> The reason is that there is ambiguity in the behavior of your suggested APIs, so we had to disambiguate by letting Python do the import (during which the dynamic linker would load `libcutensor.so`, `libcutensornet.so`, etc), and inspect the linker's internal state. 

Correction: There's no ambiguity in your suggested APIs. It's just that the wheel would no longer be simple binary repackaging, but instead it needs to come with a Python script like the above one living in the same directory. Since it incurs additional resources that do not yet exist in our packaging pipeline, it'd not be addressed in the near term 😅

The ambiguity I had in mind referred to the fact that there could be multiple `libcutensor.so` living in different places loaded in the same Python process. This is possible. It depends on how different modules were built and linked. Python allows this because Python by default loads a shared library in private mode (`RTLD_LOCAL`). It's fine for the purpose of a runtime, but not OK if we want to return the right path to a build system or user. Either your suggestion or our (cuQuantum) approach is needed.

cupy/_environment.py
leofang: btw if PY37 is dropped (#7293) we can use `importlib.metadata` here.

leofang:
I can help with this one. Will send a PR targeting your branch.

leofang:
If we drop CUDA 10.2 (https://github.com/cupy/cupy/issues/7557), I hope using cuTENSOR wheels becomes the default for Linux users, and we can keep the old CuPy installer solution only for Windows users. WDYT?

kmaehashi:
> If we drop CUDA 10.2 (#7557), I hope using cuTENSOR wheels becomes the default for Linux users, and we can keep the old CuPy installer solution only for Windows users. WDYT?

Sounds nice! I'll update this PR.

kmaehashi:
I'll do the rest of the work after merging #7647

leofang:
Thanks @kmaehashi for offline testing, we've made cuTENSOR 1.7.0 wheels available on all platforms (linux64, linux-aarch64, win64). Let us know how it goes!

leofang:
btw cuTENSOR 2.0.0 wheels are also posted

cupy/_environment.py
leofang: I suppose for this case, we can just compare the strings? `"2.0.0" >= "1.7.0"` should work. It might sound fragile, but should be ok for SemVer?

kmaehashi:
Test failures are unrelated.

cupy/_environment.py
kmaehashi: Sorry I overlooked your reply :bow:
`import pkg_resources` emits DeprecationWarning so it should be avoided anyway.

Just comparing string works... until 2.10.0 is shipped 😄
```
>>> '2.10.0' > '2.9.0'
False
```

kmaehashi: https://github.com/cupy/cupy/pull/8012

PR #7040, opened 2022-09-07
Initial experimental & private `cupyx.distributed._array` implementation

emcastillo:
Adds a private experimental API `cupyx.distributed._array` that is wip and subject to change.

Currently it works with ufuncs and custom elementwise kernels but with lots of limitations.

This is only intra-node, won't support inter-node communication like dask does.


emcastillo:
PTAL!

kmaehashi (approved):

PR #7039, opened 2022-09-06
Fix `csrsm2` memory leak

leofang:
Close #6978. cc: @anaruse @anadodik 

Note that csrsm2 is deprecated and will be removed from cuSPARSE, otherwise there're some optimization opportunities that one may exploit (ex: the info struct can be used in multiple `solve` calls for the same input matrix, see the documentation) for iterative solvers.

PR #7042, opened 2022-09-07
Fix `csrilu02` memory leak

leofang:
This leak is not as serious because in the typical iterative-solve workflow it'd be only run once. Unlike in #7039, no synchronization is needed here. cc: @anaruse

### DIFF
--- a/cupy/cusparse.py
+++ b/cupy/cusparse.py
@@ -1672,6 +1672,9 @@ def csrilu02(a, level_info=False):
     try:
         check(handle, info, position.ctypes.data)
     except Exception:
+        if info != 0:
+            _cusparse.destroyCsrilu02Info(info)
+            info = 0
         raise ValueError('a({0},{0}) is missing'.format(position[0]))
 
     solve(handle, m, nnz, desc.descriptor, a.data.data.ptr,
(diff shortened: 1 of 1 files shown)


leofang:
The simplest reproducer is to trigger the code path many times:
```python
for i in range(10000):
    lu_t = linalg.spilu(A, fill_factor=1)
```
with a sparse matrix `A`.

leofang:
Hmmm it seems the mempool tests overflowed, I wonder why...

leofang:
I can reproduce the error locally with this command
```
$ pytest tests/cupy_tests/test_cusparse.py tests/cupy_tests/cuda_tests/test_memory.py
```
Need some time to look into it.

leofang:
Keeping my notes here since I need to attend to other things (cc @anaruse).
1. This overflow happens because when computing `free_bytes`, we query both `total_bytes` and `used_bytes` and compute their difference (`total_bytes` - `used_bytes`), and the former is 0 but the latter is not. I asked internally and this is not supposed to happen, so something is off.
2. I noticed #7039 is not the full fix. I monitor the `MemoryAsyncPool` usage (since cuSPARSE uses the async pool if available) and I noticed a leak of 8~32 bytes every time `csrsm2` is called. Though I think this is likely a cuSPARSE bug (I used CUDA 11.4).
3. The first overflow happens at the shutdown time of `TestMemoryAsyncPool.test_alloc`. The memory we allocate (100 bytes, but internally we make it a 512B block) is correctly returned to the pool, but `total_bytes` goes off (set to 0) here. It doesn't matter if we synchronize at any point before querying the pool statistics or not.

Issue #1644, opened 2018-09-13
Some cupy.random distribution does not support `size=None`


toslunar:
`size=None` is different from `size=()`.
```
>>> np.random.uniform([2], [[3]])
array([[2.63150331]])
>>> np.random.uniform([2], [[3]], size=())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mtrand.pyx", line 1317, in mtrand.RandomState.uniform
  File "mtrand.pyx", line 267, in mtrand.cont2_array
ValueError: size is not compatible with inputs
>>> 
>>> cupy.random.uniform([2], [[3]])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kataoka/cupy/cupy/random/distributions.py", line 454, in uniform
    return rs.uniform(low, high, size=size, dtype=dtype)
  File "/home/kataoka/cupy/cupy/random/generator.py", line 484, in uniform
    return RandomState._scale_kernel(low, high, rand)
  File "cupy/core/_kernel.pyx", line 532, in cupy.core._kernel.ElementwiseKernel.__call__
    out_args = _get_out_args_with_params(
  File "cupy/core/_kernel.pyx", line 368, in cupy.core._kernel._get_out_args_with_params
    raise ValueError('Out shape is mismatched')
ValueError: Out shape is mismatched
```

kmaehashi:
I guess not all fixed yet? @asi1024

andfoy:
@kmaehashi @asi1024, do you know which distributions are still not supporting `size=None` properly?

andfoy:
After a little bit of inspection using the following [script](https://gist.github.com/andfoy/2b52426b4e0bddcb4c264b1f1e9eb4b7), the status regarding the broadcasting support when `size=None`. Checked means that the generator supports array broadcasting on its parameters, otherwise not.

## Valid generators
- [x] `beta`
- [x] `binomial`
- [x] `chisquare`
- [x] `dirichlet`
- [x] `exponential`
- [x] `f`
- [x] `gamma`
- [x] `geometric`
- [x] `gumbel`
- [x] `hypergeometric`
- [x] `laplace`
- [x] `logistic`
- [x] `lognormal`
- [x] `logseries`
- [x] `negative_binomial`
- [x] `noncentral_chisquare`
- [x] `noncentral_f`
- [x] `normal`
- [x] `pareto`
- [x] `poisson`
- [x] `power`
[...]

PR #7048, opened 2022-09-12
Make sure weibull distribution support ndarrays

andfoy:
See https://github.com/cupy/cupy/issues/1644


tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -1090,6 +1090,9 @@ class TestWeibull(RandomGeneratorTestCase):
     def test_weibull(self):
         self.generate(a=self.a, size=(3, 2))
 
+    def test_weibull_size_none(self):
+        self.generate([[0.5, 1.0, 3.0]])
```
emcastillo, line 1094:
```suggestion
        self.generate([[0.5, 1.0, 3.0]], size=None)
```

Lets make it explicit :)

emcastillo (approved):

PR #7051, opened 2022-09-13
Allow randint to support array bounds

andfoy:
See #1644

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -662,21 +662,35 @@ class RandomState(object):
         elif isinstance(size, int):
             size = size,
 
-        if mx == 0:
-            return cupy.zeros(size, dtype=numpy.uint32)
+        if numpy.isscalar(mx):
+            if mx == 0:
+                return cupy.zeros(size, dtype=numpy.uint32)
 
-        if mx < 0:
-            raise ValueError(
-                'mx must be non-negative (actual: {})'.format(mx))
-        elif mx <= _UINT32_MAX:
-            dtype = numpy.uint32
-            upper_limit = _UINT32_MAX - (1 << 32) % (mx + 1)
-        elif mx <= _UINT64_MAX:
-            dtype = numpy.uint64
-            upper_limit = _UINT64_MAX - (1 << 64) % (mx + 1)
+            if mx < 0:
+                raise ValueError(
+                    'mx must be non-negative (actual: {})'.format(mx))
+            elif mx <= _UINT32_MAX:
+                dtype = numpy.uint32
+                upper_limit = _UINT32_MAX - (1 << 32) % (mx + 1)
+            elif mx <= _UINT64_MAX:
+                dtype = numpy.uint64
+                upper_limit = _UINT64_MAX - (1 << 64) % (mx + 1)
+            else:
+                raise ValueError(
+                    'mx must be within uint64 range (actual: {})'.format(mx))
         else:
-            raise ValueError(
-                'mx must be within uint64 range (actual: {})'.format(mx))
+            dtype = mx.dtype
+            if dtype == cupy.int32 or dtype == cupy.uint32:
+                dtype = numpy.uint32
+                mx = mx.astype(dtype)
+                upper_limit = _UINT32_MAX - (_UINT32_MAX - mx) % (mx + 1)
+            elif dtype == cupy.int64 or dtype == cupy.uint64:
+                dtype = numpy.uint64
+                mx = mx.astype(dtype)
+                upper_limit = _UINT64_MAX - (_UINT64_MAX - mx) % (mx + 1)
+            else:
+                raise ValueError(
+                    'dtype must be integer, got: {}'.format(dtype))
 
         n_sample = functools.reduce(operator.mul, size, 1)
         if n_sample == 0:
@@ -684,13 +698,17 @@ class RandomState(object):
         sample = self._curand_generate(n_sample, dtype)
 
         mx1 = mx + 1
-        if mx1 != (1 << (mx1.bit_length() - 1)):
+        if not numpy.isscalar(mx) or mx1 != (1 << (mx1.bit_length() - 1)):
             # Get index of samples that exceed the upper limit
             ng_indices = self._get_indices(sample, upper_limit, False)
             n_ng = ng_indices.size
 
+            if n_ng > 0 and not numpy.isscalar(mx):
+                upper_limit = upper_limit[ng_indices]
+
             while n_ng > 0:
-                n_supplement = max(n_ng * 2, 1024)
+                n_supplement = (max(n_ng * 2, 1024)
+                                if numpy.isscalar(mx) else upper_limit.size)
                 supplement = self._curand_generate(n_supplement, dtype)
 
                 # Get index of supplements that are within the upper limit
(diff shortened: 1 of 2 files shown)


cupy/random/_generator.py
```diff
@@ -1150,26 +1150,51 @@ def randint(self, low, high=None, size=None, dtype=int):
...
+            if high is None:
+                lo = cupy.zeros_like(low)
+                hi = low - 1
+            else:
+                lo = low
+                hi = cupy.asarray(high) - 1
+
+            if cupy.any(lo > high):
+                raise ValueError('low >= high')
+
```
takagi, line 1164: This check causes device to host sync, which may harm its performance. We could remove this check and make it the user's responsibility as `uniform` does.

cupy/random/_generator.py
```diff
@@ -1150,26 +1150,51 @@ def randint(self, low, high=None, size=None, dtype=int):
...
+            if size is None:
+                size = cupy.broadcast(lo, hi).shape
+
+            diff = hi - lo
+            total_elems = functools.reduce(operator.mul, size, 1)
+            out = cupy.empty(total_elems, dtype=low.dtype)
+            for i, d in enumerate(cupy.flatiter(diff)):
+                out[i] = self._interval(d.item(), None)
```
takagi, line 1172: How do you think if it is possible to make `_interval` work as batch to draw samples with the upperbound of `diff` that may have different value across its elements?

andfoy: Let me take a look at the interval implementation

andfoy: @takagi, if the numbers are sampled as follow, then it is possible to sample numbers with different bounds (assuming `uint32` and no overflow cases):

```python
def _interval_nd(diff):
    mx = diff.flatten()
    mx1 = mx + 1
    sample = self._curand_generate(mx.size, np.uint32)
    sample %= mx1
    return sample
```

andfoy: I have some questions regarding the overflow management. 

1. In the current implementation the following max bound (on `uint64`) is imposed: https://github.com/cupy/cupy/blob/d588608a45e14d41bda6057cdb23daef3e40ae85/cupy/random/_generator.py#L676 This works fine on Python integers, since they operate over bigint arithmetic, however, since ndarrays operate on fixed-precision numbers, then it couldn't be possible to store `1 << 64` (the total `uint64` numbers represented) on a ndarray, which would prevent the computation of the `upper_bound` on `(u)int64` inputs.
2. Since `mx1` would be an fixed-precision ndarray, then would be this check redundant since the output numbers will be on a fixed precision? https://github.com/cupy/cupy/blob/d588608a45e14d41bda6057cdb23daef3e40ae85/cupy/random/_generator.py#L687

takagi: Thanks for your points.

1. Yes, but when `mx < _UINT64_MAX`, we can use `(_UINT64_MAX - mx) % (mx + 1)` equivalent to `(1 << 64) % (mx + 1)`, which we can compute in uint64. When `mx == _UINT64_MAX`, `(1 << 64) % (mx + 1)` is clearly zero.
2. Also yes, but let's keep the branch as a fast path in case mx is a scalar. Masking is much performant when possible.

andfoy:
@takagi, I've just applied the modifications that we previously discussed, please let me know if they require further changes

cupy/random/_generator.py
```diff
@@ -662,35 +662,53 @@ def _interval(self, mx, size):
...
+            else:
+                raise ValueError(
+                    'mx must be within uint64 range (actual: {})'.format(mx))
         else:
-            raise ValueError(
-                'mx must be within uint64 range (actual: {})'.format(mx))
+            dtype = mx.dtype
+            if dtype == cupy.int32 or dtype == cupy.uint32:
+                dtype = numpy.uint32
+                mx = mx.astype(dtype)
```
takagi, line 685:
```suggestion
                mx = mx.astype(dtype, copy=False)
```

cupy/random/_generator.py
```diff
@@ -662,35 +662,53 @@ def _interval(self, mx, size):
...
+                upper_limit = _UINT32_MAX - (_UINT32_MAX - mx) % (mx + 1)
+            elif dtype == cupy.int64 or dtype == cupy.uint64:
+                dtype = numpy.uint64
+                mx = mx.astype(dtype)
```
takagi, line 689:
```suggestion
                mx = mx.astype(dtype, copy=False)
```

cupy/random/_generator.py
```diff
@@ -1150,26 +1170,46 @@ def randint(self, low, high=None, size=None, dtype=int):
...
+                hi = cupy.asarray(high) - 1
+
+            if size is None:
+                size = cupy.broadcast(lo, hi).shape
+
+            diff = hi - lo
+            total_elems = functools.reduce(operator.mul, size, 1)
+            out = self._interval(diff.flatten(), total_elems)
+            out = cupy.reshape(out, size)
+            return out + lo
```
takagi, line 1189: I get a float64 result from `cp.random.randint([0],[10])`, which comes from `out`(uint64) + `lo` (int64). We'd like to have the same dtype as numpy here. Would you take care of this?
```py
>>> cp.random.randint([0],[10])
array([5.])  # dtype=cp.float64
>>> np.random.randint([0],[10])
array([8])  # dtype=np.int64
```

It seems that numpy does the work in `numpy.random._bounded_integers._rand_int64_broadcast`. I think it's defined in the following file as `_rand_{{nptype}}_broadcast`, which is written in Cython with maybe numpy specific source code templating. Cython is a kind of compiler that translates Python-like syntax to C, which CuPy also uses.
https://github.com/numpy/numpy/blob/main/numpy/random/_bounded_integers.pyx.in

andfoy: Just for clarification purposes, would you like to mask the inputs as in: https://github.com/numpy/numpy/blob/3b8dca3d0588048c425488a5e39ceddf815d9138/numpy/random/_bounded_integers.pyx.in#L139? Wouldn't a `astype` cast fix the issues?

takagi: At least, `float32`/`float64` can not represent all values that `int32`/`int64` do. For example, `float64` has only 52 bits for fraction.

takagi: Perhaps, we can use `int64` instead of `uint64` in `_interval` if the supplied dtype is so?

andfoy:
@takagi, I've just left some dtype-related changes on the latest commit, please let me know if they work

cupy/random/_generator.py
```diff
@@ -662,53 +672,82 @@ def _interval(self, mx, size):
         elif isinstance(size, int):
             size = size,
 
-        if mx == 0:
-            return cupy.zeros(size, dtype=numpy.uint32)
+        is_mx_scalar = numpy.is_scalar(mx)
```
takagi, line 675:
```suggestion
        is_mx_scalar = numpy.isscalar(mx)
```

takagi: my bad

takagi:
I will merge this after the branch is unfrozen.

takagi (approved):

takagi:
Thanks, @andfoy!

PR #7059, opened 2022-09-22
added cupy.ndarray.searchsorted

Diwakar-Gupta:
Hello,
I recently came across the CUPY and found it interesting. I am excited and want to contribute to the cupy.

As asked in #6078, I have implemented `cupy.ndarray.searchsorted`;
However,tests still need to be written. I'll need some advice on how to do that. 

Thank you.


1MrEnot:
@Diwakar-Gupta, for tests I can suggest add your tests to existing tests on cupy.searchsorted [here](https://github.com/cupy/cupy/blob/master/tests/cupy_tests/sorting_tests/test_search.py#L632). Similar method, but it should call `y.searchsorted`, not `xp.searchsorted`.

And in general you should add documetation somewhere in `docs/source/reference/...` , but documentation about this method seems already [present](https://docs.cupy.dev/en/stable/reference/generated/cupy.searchsorted.html)

1MrEnot:
@Diwakar-Gupta, hey, your tests are correct and work nicely, but quite often in cupy tests are implemented another way.
Cupy offers drop-in replacment, which means, that cupy must have the same api as numpy as scipy. To ensure that cupy has the same api, and corresponding function return the same result at the same time there's interastructure around tests, that allows you to write test library-agnostic and then comare results
For example [this](https://github.com/cupy/cupy/blob/66b4964569a64ef0c33dbf751ce2f06a840aab99/tests/cupy_tests/sorting_tests/test_search.py#L634-L640) test.
There's an attribute `numpy_cupy_array_equal`, that runs the same test function twice: once with `numpy` provided as `xp` argument and then with `cupy` as `xp` (there's nothing special about `xp` - it's just default argument name that can be overwirtten by providing the argument to the decorator). This test function returns value (actualy twice), calculated by provided library, and then machinery behind this attribute asserts, that results are equal for both libraries

So, I think, it would be better to rewrite the test using this technique

Diwakar-Gupta:
There are more types of tests, such as `TestSearchSortedInvalid` `TestSearchSortedNanInf` and others. Do we also need to implement tests for them? @1MrEnot

1MrEnot:
I think so - it's better have more tests, than less. 
Also I did not really undersntand your idea about speeding up test. Nevertheless, it seems like these concrete tests does not requre a lot of time - there're scipy tests, that take tens of seconds per run, so even x10 speedup won't be significant. And it would be better to have the same 'style' for all tests at least in this class

cupy/_core/core.pyx
```diff
@@ -942,6 +942,18 @@ cdef class _ndarray_base:
...
         return _sorting._ndarray_argpartition(self, kth, axis)
 
+    def searchsorted(self, v, side='left', sorter=None):
+        """Finds indices where elements of v should be inserted to maintain order.
+
+        For full documentation, see :func:`cupy.searchsorted`
+
+        Returns:
+
+        .. seealso:: :func:`cupy.searchsorted`
```
kmaehashi, line 952: In CuPy, seealso section works as a link to NumPy counterpart.

```suggestion
        .. seealso:: :func:`numpy.searchsorted`
```

tests/cupy_tests/sorting_tests/test_search.py
```diff
@@ -743,3 +758,94 @@ def test_nonint_sorter(self):
             sorter = xp.array([], dtype=xp.float64)
             with pytest.raises(TypeError):
                 xp.searchsorted(bins, x, sorter=sorter)
+
+
+@testing.gpu
```
kmaehashi, line 763: This annotation is there for historical reasons, no longer needed for new tests :)
```suggestion
```

tests/cupy_tests/sorting_tests/test_search.py
```diff
@@ -743,3 +758,94 @@ def test_nonint_sorter(self):
...
+    @testing.numpy_cupy_array_equal()
+    def test_searchsorted_minf(self, xp):
+        x = testing.shaped_arange((10,), xp, xp.float64)
+        x[5] = float('-inf')
+        bins = xp.array([0, 1, 2, 4, 10])
+        y = bins.searchsorted(x, side=self.side)
+        return y,
+
+
+@testing.gpu
```
kmaehashi, line 826:
```suggestion
```

kmaehashi (changes requested):
Sorry for the delay in review, looks good except for nits!

kmaehashi (approved):

PR #7072, opened 2022-09-30
CI: tentatively fix hypothesis version

kmaehashi:
hypothesis 65.5.0 started to infer array api spec version from `__array_api_version__` attribute. https://github.com/HypothesisWorks/hypothesis/releases/tag/hypothesis-python-6.55.0
This broke the CI so until we decide the proper value for that attribute pin the hypothesis version to the previous one.

https://data-apis.org/array-api/latest/future_API_evolution.html#versioning

`numpy.array_api` doesn't seem to have the attribute either.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ extras_require = {
     'test': [
         # 4.2 <= pytest < 6.2 is slow collecting tests and times out on CI.
         'pytest>=6.2',
-        'hypothesis>=6.37.2',
+        'hypothesis>=6.37.2,<6.55.0',
     ],
 }
 tests_require = extras_require['test']


takagi:
LGTM!

leofang:
Hi @honno @asmeurer is this an expected issue?

honno:
> Hi @honno @asmeurer is this an expected issue?

Yep `__array_api_version__` was always expected in the spec but we all forgot about it haha. Wrote a clarifying PR in https://github.com/data-apis/array-api/pull/480.

honno:
Updated the test suite so it bumps Hypothesis (we need the latest version going forward) and makes libraries missing `__array_api_version__` still work https://github.com/data-apis/array-api-tests/pull/144. Made a release too.

EarlMilktea:
Any updates?

Can we safely use the latest version now?

PR #7077, opened 2022-10-03
Support `add.at`, `maximum.at`, `minimum.at`

asi1024:
Part of #7062.


asi1024:
I will split deprecations of `ndarray.scatter_{add, maxmimum, minimum}` into another PR.

emcastillo (approved):

PR #7078, opened 2022-10-03
Add pdist implementation to distance functions

andfoy:
This implementation makes use of `cdist` matrix computation, from which the upper triangular values (without the main diagonal) are taken in order to be SciPy-compatible


leofang:
cc: @cjnolet

cupyx/scipy/spatial/distance.py
```diff
@@ -494,6 +494,37 @@ def cdist(XA, XB, metric='euclidean', out=None, **kwargs):
         raise TypeError('2nd argument metric must be a string identifier')
 
 
+def pdist(X, metric='euclidean', out=None, **kwargs):
```
takagi, line 497:
```suggestion
def pdist(X, metric='euclidean', *, out=None, **kwargs):
```

tests/cupyx_tests/scipy_tests/spatial_tests/test_distance.py
```diff
@@ -62,17 +67,48 @@ def test_cdist_(self, xp, scp):
...
+    'dtype': ['float32', 'float64'],
+    'rows': [20, 100],
+    'cols': [20, 100],
+    'p': [1.0, 2.0, 3.0],
+    'order': ["C", "F"]
+}))
+@pytest.mark.skipif(cupy.cuda.runtime.is_hip, reason="tests for CUDA only")
+@pytest.mark.skipif(not scipy_available or not pylibraft_available,
+                    reason='requires scipy and pylibraft')
+class TestPdist(unittest.TestCase):
```
takagi, line 97: nit: We use pytest for our unit testing framework, so test classes do not need to inherit unittest.TestCase. (Some tests in CuPy do so just because of historical reasons)

```suggestion
class TestPdist:
```

cupyx/scipy/spatial/distance.py
```diff
@@ -494,6 +494,37 @@ def cdist(XA, XB, metric='euclidean', out=None, **kwargs):
...
+            metric documentation for a list of all possible arguments.
+            Some possible arguments:
+            p (float): The p-norm to apply for Minkowski, weighted and
+            unweighted. Default: 2.0
+
+    Returns:
+        Y (cupy.ndarray): A :math:`m` by :math:`m` distance matrix is
+            returned. For each :math:`i` and :math:`j`, the metric
+            ``dist(u=X[i], v=X[j])`` is computed and stored in the
+            :math:`ij` th entry.
```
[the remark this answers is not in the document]
takagi: FYI: You can see the documentation including changes in a PR via `docs/readthedocs.org:cupy`'s "Details" link in GitHub checks list below (maybe hidden in the scrolling view)

cupyx/scipy/spatial/distance.py
takagi: We're taking care of making CuPy functions always have exactly the same signatures against NumPy/SciPy counterparts whenever possible.

takagi:
LGTM except for some minor fixes! I also checked the test passes with pylibraft.

takagi (approved):

PR #7079, opened 2022-10-03
Conditionally change identifiers for ROCm

hubertlu-tw:
CuPy installation fails on ROCm 5.4 due to a missing identifier in `cupy/cuda/cupy_thrust.cu`. Issue is due to an update in ROCPrim API in ROCm5.4. This PR is to conditionally add ROCPrim identifier for ROCm5.4 and above.

@amathews-amd @AdrianAbeyta


kmaehashi:
Thank you, and sorry for taking long time to merge :bow:

PR #7086, opened 2022-10-05
Add 1-D BSpline to interpolate module

andfoy:


cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,420 @@
...
+
+        # Note that the normalized axis is stored in the object.
+        self.axis = axis
+        if axis != 0:
+            # roll the interpolation axis to be the first one in self.c
+            # More specifically, the target shape for self.c is (n, ...),
+            # and axis !=0 means that we have c.shape (..., n, ...)
+            #                                               ^
+            #                                              axis
+            self.c = np.moveaxis(self.c, axis, 0)
```
takagi, line 293:
```suggestion
            self.c = cp.moveaxis(self.c, axis, 0)
```

takagi:
Thanks for your commits! I'm roughly seeing your codes. I suppose you think of going to write tests after you have finished implementing the methods. If possible, would you add a test every time you've implemented a function as this PR is relatively large? It would help me to review your work here. (Anyway, it's not urgent)

andfoy:
@takagi, I'm not sure which is the Sphinx error on the CI, the docs are building correctly both on ReadTheDocs and also locally

takagi:
Let me run the test in environments including AMD's ROCm.

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,912 @@
...
+    n = t.shape[0] - k - 1
+    intervals = cupy.empty_like(xp, dtype=cupy.int_)
+
+    # Compute intervals for each value
+    interval_kernel = _get_module_func(INTERVAL_MODULE, 'find_interval', xp)
+    interval_kernel((max(1, xp.shape[0] // 128),), (min(128, xp.shape[0]),),
+                    (t, xp, intervals, k, n, extrapolate))
+
+    # Compute interpolation
+    num_c = int(np.prod(c.shape[1:]))
```
ev-br, line 246: Minor: Scipy has a small utility for this, https://github.com/scipy/scipy/blob/main/scipy/_lib/_util.py#L153 maybe use it here and elsewhere for products of shape arrays.

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,912 @@
...
+import cupy
+from cupy._core import internal
+from cupy._core._scalar import get_typename
+
+from cupyx.scipy.sparse import csr_matrix
+
+import numpy as np
+
+TYPES = ['float', 'double', 'complex<float>', 'complex<double>']
+INT_TYPES = ['int', 'long long']
```
ev-br, line 14: Why long long? (My ignorance is showing, yes.)

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,912 @@
...
+
+INTERVAL_KERNEL = r'''
+#include <cupy/complex.cuh>
+
+template<typename T>
+__global__ void find_interval(
+        const T* t, const T* x, long long* out,
```
ev-br, line 21: So, it will get templated on the TYPES above, correct?  `x` and `t` are evaluation points and knots, they are certainly real-valued only. IIRC, the scipy analog upcasts floats to doubles even.

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,912 @@
...
+        int k, int n, bool extrapolate) {
```
ev-br, line 22: For my education: this is a C++ bool, correct? What's the story of converting C++ bools to python bools here -- note that Cython has a special `bint` type to avoid a mess with C bools vs python bools.

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,912 @@
...
+        Computed values of the spline at each of the input points.
+        This argument is modified in-place.
+    """
+    n = t.shape[0] - k - 1
+    intervals = cupy.empty_like(xp, dtype=cupy.int_)
+
+    # Compute intervals for each value
+    interval_kernel = _get_module_func(INTERVAL_MODULE, 'find_interval', xp)
+    interval_kernel((max(1, xp.shape[0] // 128),), (min(128, xp.shape[0]),),
+                    (t, xp, intervals, k, n, extrapolate))
```
ev-br, line 243: Again, for my education: what is the magic constant 128 here?

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,912 @@
...
+    d_boor_kernel((max(1, x.shape[0] // 128),), (min(128, x.shape[0]),),
+                  (t, None, k, 0, x, intervals, None, bspline_basis, 0, 0))
+
+    data = cupy.zeros(x.shape[0] * (k + 1), dtype=cupy.float_)
+    design_mat_kernel = _get_module_func(
+        DESIGN_MAT_MODULE, 'compute_design_matrix', x, indices)
+    design_mat_kernel((max(1, x.shape[0] // 128),), (min(128, x.shape[0]),),
+                      (k, intervals, bspline_basis, data, indices))
+
+    return data, indices
```
ev-br, line 301: Note that if the structure inherits the scipy version, which minimizes the amount of stuff being sent between python and cython. ISTM the code will be easier to understand if indptr is constructed here as well.

See however a separate comment on this whole function.

ev-br (commented):
A scipy.interpolate dev here.

This is only a cursory review. Cannot comment on CUDA parts (will need to do my homework), so several comments on the interface and questions. 

Also it is definitely my lack of knowledge --- would it make sense to just drop the relevant part of the scipy test suite here?

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,912 @@
...
+            int_dtype = cupy.int64
+        # Preallocate indptr and indices
+        indices = cupy.empty(n * (k + 1), dtype=int_dtype)
+        indptr = cupy.arange(0, (n + 1) * (k + 1), k + 1, dtype=int_dtype)
+
+        # indptr is not passed to CUDA as it is already fully computed
+        data, indices = _make_design_matrix(
+            x, t, k, extrapolate, indices
+        )
+        return csr_matrix(
```
ev-br, line 673: This is a problem, actually. The scipy version returns a csr_array, not csr_matrix. Does cupy have csr_arrays even? If not, maybe remove this whole method until csr_arrays are implemented.

takagi: CuPy does not have csr_arrays yet.

cupyx/scipy/interpolate/_bspline.py
takagi: We rely on `numpy.bool_` to get C bools from Python bools.
https://github.com/cupy/cupy/blob/6b31424246eb7d0e27a330b075211921a24406ae/cupy/cuda/function.pyx#L141-L142

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,918 @@
...
+
+INTERVAL_KERNEL = r'''
+#include <cupy/complex.cuh>
+
+template<typename T>
+__global__ void find_interval(
+        const T* t, const T* x, long long* out,
+        int k, int n, bool extrapolate) {
+
+    int idx = blockDim.x * blockIdx.x + threadIdx.x;
```
takagi, line 24: Don't we need to check the boundary for `idx` so that it doesn't make overrun?

ev-br: Also, I *think* the binary search needs to account for multiple knots. These can occur anywhere, incl at either end of the knot array. The scipy routine uses a linear search, https://github.com/scipy/scipy/blob/v1.9.3/scipy/interpolate/_bspl.pyx#L32, which is literally translated from FITPACK, https://github.com/scipy/scipy/blob/v1.9.3/scipy/interpolate/fitpack/splev.f#L93

cupyx/scipy/interpolate/_bspline.py
takagi: `interval_kernel`'s first and second arguments are corresponding to CUDA's `<<<...>>>` syntax that specify the number of blocks per grid and the number of threads per block. There could be more optimal value other than `128` depending on GPUs, but I suppose `128` is fair for general purpose libraries as CuPy. Note that the value should be a multiple of 32 as CUDA's multiprocessors execute 32 parallel threads called warps at the same time.

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,918 @@
...
+    out : ndarray, shape (s, m)
+        Computed values of the spline at each of the input points.
+        This argument is modified in-place.
+    """
+    n = t.shape[0] - k - 1
+    intervals = cupy.empty_like(xp, dtype=cupy.int_)
+
+    # Compute intervals for each value
+    interval_kernel = _get_module_func(INTERVAL_MODULE, 'find_interval', xp)
+    interval_kernel((max(1, xp.shape[0] // 128),), (min(128, xp.shape[0]),),
```
takagi, line 248: We need two blocks for, say, 129 elements.

```suggestion
    interval_kernel(((xp.shape[0] + 128 - 1) // 128,), (min(128, xp.shape[0]),),
```

cupyx/scipy/interpolate/_bspline.py
andfoy: The linear search (after the binary one) is called once an initial interval is found: https://github.com/cupy/cupy/pull/7086/commits/9869f1842780cac21d57ba0896ea2c4fb1f182d2

cupyx/scipy/interpolate/_bspline.py
andfoy: `long long` maps to `int64` and `int` to `int32`, which can occur here: https://github.com/cupy/cupy/blob/9dac7f88fb0f22d2e29d56f70d36744c43a70a4b/cupyx/scipy/interpolate/_bspline.py#L690

takagi: (`long` is 32bit integer on Windows)

cupyx/scipy/interpolate/_bspline.py
andfoy: I have a question regarding the `threadSize`: right now is declared as `min(128, x.shape[0])`, which should work as long as `x.shape[0]` is less than 1024. Do you know if it should be 128 instead?

cupyx/scipy/interpolate/_bspline.py
```diff
@@ -0,0 +1,934 @@
...
+'''
+
+INTERVAL_MODULE = cupy.RawModule(
+    code=INTERVAL_KERNEL, options=('-std=c++11',),)
+#    name_expressions=[f'find_interval<{type_name}>' for type_name in TYPES])
+
+
+D_BOOR_KERNEL = r'''
+#include <cupy/complex.cuh>
+#include <math_constants.h>
```
takagi, line 76: Addresses Jenkins failures, this is from ROCm environment.

```suggestion
#include <cupy/math_constants.h>
```

andfoy:
These are the benchmark results that compare the current `BSpline` implementation (as of a99d344) against the SciPy one. For larger inputs, CuPy implementation presents a ~10X improvement over the Scipy one.

<details>

```python
from itertools import product

import cupy as cp
import numpy as np
from cupy import testing

from cupyx.scipy.interpolate import BSpline as CuBSpline
from scipy.interpolate import BSpline as CPUBSpline

from tqdm import tqdm


input_sizes = [5, 10, 100, 1000, 10000, 50000, 100000, 500000, 700000]
modules = [(cp, CuBSpline, 'CuPy', False), (np, CPUBSpline, 'SciPy', True)]
kwargs = [('extrapolate', [True, False, 'periodic'])]
measurement = {}


def gather_time(prof):
    return {
        'cpu': cpu_time,
        'gpu': gpu_time
    }


    n = 35
    k = 3
    t = xp.sort(testing.shaped_random((n + k + 1,), xp, dtype=np.float64))
    c = testing.shaped_random((n,), xp, dtype=np.float64)
    b = cls(t, c, k, **kwargs)
    x = testing.shaped_random(size, xp)
    return b, x


def closure(b, x):
    return b(x)


for kwarg_comb in kwargs:
    kwarg_comb = dict(kwarg_comb)
        for xp, cls, _id, _ in modules:


lines = []
for kwarg_comb in kwargs:
    lines.append(f'## `{kwarg_line}`\n')
    lines.append('| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |')
    lines.append('|:----:|:-------------:|:-------------:|:----------:|')
    for size in input_sizes:
        comp = f'{size}'
        for _, _, _id, cpu_only in modules:
            times = size_measurement[_id]
            gpu_time = times['gpu']
            cpu_time = times['cpu']
            time_info = f'{cpu_time:3f}'
            if not cpu_only:
                time_info = f'{time_info} | {gpu_time:3f}'
        lines.append(f'| {comp} |')
    lines.append('\n')

print('\n'.join(lines))
```

</details>

### `extrapolate=True`

| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |
|:----:|:-------------:|:-------------:|:----------:|
| 5 | 0.143162 | 0.159167 | 0.009024 |
| 10 | 0.141033 | 0.152604 | 0.008848 |
| 100 | 0.138681 | 0.143938 | 0.013551 |
| 1000 | 0.140014 | 0.145257 | 0.060882 |
| 10000 | 0.141720 | 0.173564 | 0.529937 |
| 50000 | 0.144045 | 0.354529 | 2.594270 |
| 100000 | 0.146247 | 0.587671 | 5.173619 |
| 500000 | 0.145439 | 2.388427 | 25.833946 |
| 700000 | 0.149244 | 3.492099 | 36.182867 |


### `extrapolate=False`

| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |
|:----:|:-------------:|:-------------:|:----------:|
| 5 | 0.139301 | 0.163633 | 0.008737 |
| 10 | 0.138752 | 0.175799 | 0.008718 |
| 100 | 0.139201 | 0.143952 | 0.012707 |
| 1000 | 0.140492 | 0.145340 | 0.052743 |
| 10000 | 0.141624 | 0.174490 | 0.457812 |
| 50000 | 0.144299 | 0.365292 | 2.243783 |
| 100000 | 0.145420 | 0.592371 | 4.479947 |
| 500000 | 0.145595 | 2.507396 | 22.361803 |
| 700000 | 0.146802 | 3.093445 | 31.318701 |


### `extrapolate=periodic`

| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |
|:----:|:-------------:|:-------------:|:----------:|
| 5 | 0.213266 | 0.244532 | 0.016144 |
| 10 | 0.212232 | 0.235847 | 0.016432 |
| 100 | 0.213510 | 0.218770 | 0.021867 |
| 1000 | 0.215104 | 0.220409 | 0.076617 |
| 10000 | 0.217468 | 0.272842 | 0.608236 |
| 50000 | 0.224108 | 0.450436 | 2.965852 |
| 100000 | 0.224882 | 0.703648 | 5.911659 |
| 500000 | 0.222896 | 2.756336 | 29.488543 |
| 700000 | 0.227213 | 3.767208 | 41.443494 |

takagi:
Looking so nice result! I'm happy with seeing it.

The Jenkins CI still fails in AMD's ROCm environments. I'll check and fix them. (I guess you don't have access to AMD GPUs, right?) Other than that, the code and tests look good to me. Do you have anything else to do in your mind?

andfoy:
Thanks for fixing the ROCm issues, at the moment, we don't have any access to AMD GPUs (that we know of)

Besides that, this PR is finally ready

rgommers:
Question: I looked at [the benchmark result docs](https://docs.cupy.dev/en/stable/reference/generated/cupyx.profiler._time._PerfCaseResult.html#cupyx.profiler._time._PerfCaseResult), but it didn't enlighten me as to what the difference is between CuPy-CPU and CuPy-GPU. Is the thing to compare to SciPy the sum of both of them, or something else?

> Thanks for fixing the ROCm issues! At the moment, we don't have any access to AMD GPUs (that we know of)

We should have at least one coming our way soon, but I haven't heard about the current status recently.

takagi:
`cupyx.profiler.benchmark` measures both CPU time and GPU time, and we can compare max(CuPy-CPU, CuPy-GPU) with SciPy. Note that because of overheads, GPU time in smaller size is likely to be inaccurate.

@andfoy, would you merge the CuPy columns into one to show max(CuPy-CPU, CuPy-GPU) as CuPy time, which would make the message of the tables clearer?

ev-br:
My layperson understanding was that indeed the wallclock time is max(CPU_time, GPU_time), and that both times are meaningful: say, CPU_time being is much larger than the GPU_time means that GPUs are underused. Is this correct?

andfoy:
The benchmark results now reflect the actual walltime

takagi:
> My layperson understanding was that indeed the wallclock time is max(CPU_time, GPU_time), and that both times are meaningful: say, CPU_time being is much larger than the GPU_time means that GPUs are underused. Is this correct?

Yes!

> The benchmark results now reflect the actual walltime

Thanks!

takagi (approved):
LGTM!

takagi:
Thanks, @andfoy!

andfoy:
Thanks a lot for the reviews! @ev-br @takagi

takagi:
@andfoy Our daily CI fails in `TestBSpline::test_design_matrix_same_as_BSpline_call` these days, raising a ValueError in both scipy and cupy. Would you take a look? Maybe the tests draw inappropriate values and say "Out of bounds".
https://ci.preferred.jp/cupy.linux.cuda-head/128581/

(please see its full log below as that on web UI is omitted)
https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/128581/log.txt

Thanks!

andfoy:
@takagi, it is now done: https://github.com/cupy/cupy/pull/7521

I forgot to add a raises check context manager, the error was expected both on SciPy and CuPy

PR #7089, opened 2022-10-06
Find wheels from pip.cupy.dev in Dockerfile

kmaehashi:
Assets section of GitHub Releases are now lazy-loaded and so pip cannot find links from there.


emcastillo (approved):
LGTM

Issue #7092, opened 2022-10-07
Support ROCm 5.3


takagi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [ ] Read [ROCm Release Notes](https://rocmdocs.amd.com/en/latest/Current_Release_Notes/Current-Release-Notes.html).
- [x] Update AMD driver in Jenkins test infrastructure (ask @kmaehashi).
- [ ] Fix code and CI to support the new vesrion.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - **Wheel Package Detection**: Add the package to the [duplicate detection](https://github.com/cupy/cupy/blob/master/cupy/_environment.py).
  - **Universal Wheel Package (`cupy-wheel`)**: Update the list of packages and detection logic in [universal_pkg](https://github.com/cupy/cupy/blob/master/install/universal_pkg/setup.py).
- [ ] Backport the above PR.
- [ ] Update Jenkins job configuration for ROCm wheel build (`other_cupy_release_builder`) (ask @kmaehashi)
- [ ] Fix `cupy-release-tools` to support the new version.
- [ ] Backport the above PR.
- [ ] Fix documentation.
  - Add new wheel package to the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) and `README.md`.
  - Update requirements in the installation guide.
- [ ] Backport the above PR.
- [ ] Add new wheel package to the [website](https://cupy.dev).
- [ ] Implement or create an issue to support new features, if applicable.

kmaehashi:
#7217

PR #7120, opened 2022-10-14
ROCm 5.3 rocprim support

amathews-amd:
Part of https://github.com/cupy/cupy/issues/7092

cc: @takagi , @AdrianAbeyta


emcastillo:
I think this is the same than #7079 ?
Thanks for the PR!

amathews-amd:
> I think this is the same than #7079 ? Thanks for the PR!

Yes, I forgot @hubertlu-tw  opened the other PR. Found that ROCm5.3 also needs this on Friday.

PR #7122, opened 2022-10-17
Fix compile error from `inf`/`nan` in cupy.fuse

### LINKED ISSUE #7118: [Discussion needed] Add ways to write `inf` and `nan` as literals in fused and user-defined kernels.
### Description

Currently, literals such as `inf` and `nan` cannot be used inside a `ElementwiseKernel` or a fused kernel. 
To test whether that's the case, we can write down the following code.
```python
@cp.fuse
def test_inf_nan(arr):
    arr[:] = cp.inf
    arr[:] = cp.nan
    arr[:] = 0
    return arr
print(test_inf_nan(cp.linspace(0,4,10)))
```
The compiler will spit out two errors regarding the fused kernel:
```
error: identifier "inf" is undefined
error: a value of type "double (*)(const char *)" cannot be used to initialize an entity of type "const double"
```
However, I am not so sure whether this is something that could be useful. Here are some of the pros and cons that I can think of:

pros:
- Functions that accept those literals will have a easier time when being fused. Such as `cupy.clip`.
- It might be helpful for compile-time-optimizations.

cons:
- The actual use case for those literals is quite scarce.
- We can always store those literals inside a variable and pass them into the kernels when calling the functions instead.

Any suggestions would be appreciated, thanks in advance!

### Additional Information

_No response_

asi1024:
Close #7118.

### DIFF
--- a/cupy/_core/fusion.pyx
+++ b/cupy/_core/fusion.pyx
@@ -119,7 +119,15 @@ class _FusionVarCUDA(object):
         elif isinstance(val, complex):
             init = '({}, {})'.format(c.real, c.imag)
         elif isinstance(val, (int, float)):
-            init = '= {}'.format(c)
+            if numpy.isnan(val):
+                init = '= __int_as_float(0x7f800001)'
+            elif numpy.isinf(val):
+                if val > 0:
+                    init = '= __int_as_float(0x7f800000)'
+                else:
+                    init = '= __int_as_float(0xff800000)'
+            else:
+                init = '= {}'.format(c)
         else:
             raise TypeError('Invalid constant type: {}'.format(type(c)))
         return 'const {} v{} {};\n'.format(ctype, self.index, init)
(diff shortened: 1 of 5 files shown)


cupy/_core/fusion.pyx
```diff
@@ -119,7 +119,15 @@ class _FusionVarCUDA(object):
         elif isinstance(val, complex):
             init = '({}, {})'.format(c.real, c.imag)
         elif isinstance(val, (int, float)):
-            init = '= {}'.format(c)
+            if numpy.isnan(val):
```
kmaehashi, line 122: (As discussed offline) there are constants defined in CUDA header file, but is not included by default so I think it's ok to remain as-is.
https://github.com/cupy/cupy/blob/master/cupy/_core/include/cupy/math_constants.h

Btw, the variable could be a double?

cupy/_core/fusion.pyx
```diff
@@ -119,7 +119,15 @@ class _FusionVarCUDA(object):
...
+                init = '= __int_as_float(0x7f800001)'
```
emcastillo, line 123: nit:
/usr/local/cuda/include/math_constants.h
seems to define `CUDART_NAN` `CUDART_NAN_F` and INF ones
This is purely a cosmetic change as these values will not change, but you can use the mnemonic if you feel in the mood :)

asi1024: Fixed!

Issue #7131, opened 2022-10-19
Support from_dlpack from numpy.NDArray


ar90n:
### Description

In [dlpack.pyx](https://github.com/cupy/cupy/blob/a96007c548204b3b1a18a88d56cd105dd68be3b7/cupy/_core/dlpack.pyx#L392), there is a TODO comment about support from_dlpack from numpy.NDArray.
Now, this [PR](https://github.com/numpy/numpy/pull/19083) about DLPack support is merged.
Is there any plan to solve this TODO?

In my environment, the following modifications work correctly.
```diff
diff --git a/cupy/_core/dlpack.pyx b/cupy/_core/dlpack.pyx
index 5860b4e1a..24d6125b0 100644
--- a/cupy/_core/dlpack.pyx
+++ b/cupy/_core/dlpack.pyx
@@ -391,7 +391,9 @@ cpdef from_dlpack(array):
     elif dev_type == <int>kDLCPU:
         # TODO(kmaehashi): Call `np.from_dlpack` when DLPack support is in:
         # https://github.com/numpy/numpy/pull/19083
-        raise ValueError('CPU arrays cannot be imported to CuPy.')
+        #raise ValueError('CPU arrays cannot be imported to CuPy.')
+        import numpy as np
+        return cupy.asarray(np.from_dlpack(array))
     else:
         # TODO(leofang): support kDLCUDAPinned etc
         dltensor = None
```

kmaehashi:
Thanks for raising this discussion! I'm the one who added that TODO comment, but now I'm wondering how we should handle `cupy.from_dlpack(cpu_tensor)`. Possible ideas are (1) return `numpy.ndarray`, (2) return `cupy.ndarray` as you suggested, or (3) raise ValueError. Returning `cupy.ndarray` makes sense at some level, but that means the copy of the input tensor is created, which is in general unexpected behavior in terms of the DLPack's design. At this point, my opinion is to raise ValueError as it is now and suggest users to use `cupy.array(np.from_dlpack(cpu_tensor))`.

ar90n:
@kmaehashi 
Thanks for your comment. It's so helpful to me. Because I'm not familiar with DLPack's design.
I understand your opinion, and I'm going to use the snippets you suggested.

PR #7133, opened 2022-10-19
Raise `TypeError` instead of `ValueError` in `cupy.from_dlpack` when CPU tensor is passed

kmaehashi:
https://data-apis.org/array-api/latest/design_topics/data_interchange.html

> If an array that is accessed via the interchange protocol lives on a device that the requesting library does not support, it is recommended to raise a TypeError.

ref. #7131

### DIFF
--- a/cupy/_core/dlpack.pyx
+++ b/cupy/_core/dlpack.pyx
@@ -389,9 +389,9 @@ cpdef from_dlpack(array):
         finally:
             cupy.cuda.runtime.setDevice(prev_device)
     elif dev_type == <int>kDLCPU:
-        # TODO(kmaehashi): Call `np.from_dlpack` when DLPack support is in:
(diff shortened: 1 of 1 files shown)


takagi (approved):

cupy/_core/dlpack.pyx
```diff
@@ -355,8 +351,12 @@ cpdef from_dlpack(array):
         are responsible to ensure data safety.
 
     .. seealso::
-        `Data interchange mechanisms`_
+        :func:`numpy.from_dlpack`,
+        `Python Specification for DLPack`_
+        `Data interchange mechanisms`_,
```
kmaehashi, line 356:
```suggestion
        `Python Specification for DLPack`_,
        `Data interchange mechanisms`_
```

takagi (approved):
LGTM!

leofang (approved):

PR #7135, opened 2022-10-19
Adding note about `nvcc` backend for using cub headers

rietmann-nv:
Ran into this and thought a little note in the docs would have saved us some trouble.

### DIFF
--- a/cupy/_core/raw.pyx
+++ b/cupy/_core/raw.pyx
@@ -26,7 +26,8 @@ cdef class RawKernel:
             or NVCC). For details, see
             https://docs.nvidia.com/cuda/nvrtc/index.html#group__options or
             https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#command-option-description
-        backend (str): Either `nvrtc` or `nvcc`. Defaults to `nvrtc`
+        backend (str): Either `nvrtc` or `nvcc`. Defaults to `nvrtc` , but if you
+            need headers like `<cub/cub.cuh>` you will need to use `nvcc`.
         translate_cucomplex (bool): Whether the CUDA source includes the header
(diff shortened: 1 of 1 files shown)


leofang (changes requested):
As discussed offline this is not needed. Including the CUB global header `cub.cuh` brings in everything, and it is expected to fail. If you only do `cub::BlockReduce`, you can just include the block reduction header, and it should work fine with NVRTC. I forget if you need to set `jitify=True` or not (**update: yes we need it**), but it definitely works because it's [how CuPy uses CUB reduction internally](https://github.com/cupy/cupy/blob/da05392f5e2c6fc45ea8282ea1b5badadd9c5132/cupy/_core/_cub_reduction.pyx#L24).

emcastillo:
> you can just include the block reduction header

Not sure if this is documented, but it would be great! I could fall into this same pitfall myself :)

leofang:
> Not sure if this is documented

I don't think so, it's too specific to CUB. Mainly, CUB device algorithms and some internal constructs cannot be compiled by NVRTC, and sometimes even Jitify cannot help. Moreover, A majority of CUB device algorithms are regularly tested with CUDA Dynamic Parallelism (CDP) -- i.e. launched on the device -- but it is only possible when compiling with NVCC. To me it seems more appropriate to add these Q&A to CUB documentation instead of CuPy.

Anyway, I've been wondering instead of adding tricks and caveats like these to the docstring if it's better to mention them here 
https://docs.cupy.dev/en/stable/user_guide/kernel.html
or here
https://docs.cupy.dev/en/stable/user_guide/performance.html
We could rename it from "Performance Best Practices" to "Best Practices" to be more general.

emcastillo:
Closing as per Leo's comments.

PR #7141, opened 2022-10-25
CI: improve use of cache in GitHub Actions

kmaehashi:
## Fix ccache not used

At some point ccache version has bumped to v4 and was not working as expected.

ccache v3 https://github.com/ccache/ccache/blob/3.7-maint/doc/MANUAL.adoc
> cache_dir (CCACHE_DIR)
> ... The default is $HOME/.ccache.

ccache v4 https://ccache.dev/manual/latest.html#_location_of_the_configuration_file
> cache_dir (CCACHE_DIR)
> ... the default is $HOME/.ccache if such a directory exists, otherwise $XDG_CACHE_HOME/ccache if XDG_CACHE_HOME is set, otherwise $HOME/Library/Caches/ccache (macOS) or $HOME/.config/ccache (other systems).

## Utilize cache generated by the previous commit

For `push` events, cache generated by the test run for previous commit hash can be utilized.


emcastillo:
Merging as only GHA is required (SciPy is down now)

PR #7117, opened 2022-10-14
Support CUDA 11.8

takagi:
#7080

PR #7153, opened 2022-10-31
Support NCCL 2.15

takagi:
Follows up #7117.


kmaehashi:
Seems the URL has changed from public version to full version `v2.14` -> `v2.15.1`

PR #7161, opened 2022-11-02
CI: Add support for the latest FlexCI Windows image

kmaehashi:
The new CI image (which is already deployed) has cuDNN in a separate path.


kmaehashi:
Self-merging as this blocks other CI for PRs.

PR #7175, opened 2022-11-07
CI: Generate coverage count just after the parameter axis in table

kmaehashi:
When reviewing matrix change of CI, it is difficult to check coverage count as the table is now large.

Preview: https://github.com/cupy/cupy/blob/ee87336918fc3ad0de6947a754de0426423c738a/.pfnci/coverage.rst


takagi (dismissed):
LGTM!

takagi (approved):
LGTM!

Issue #7186, opened 2022-11-11
Tracker: cupyx.scipy.interpolate


andfoy:
### Description

This is an issue to keep track of the current improvement work to the `cupyx.scipy.interpolate` module.  The order of the checkboxes denote the planned order and priority of implementation of each module. This plan was devised by @ev-br and is subject to change and discussion.

## First phase
- [ ] `BSpline` → https://github.com/cupy/cupy/pull/7086
- [ ] `make_interp_spline`
- [ ] `PPoly`
- [ ] `CubicHermiteSpline`: Required by `pchip/akima/CubicSpline`
- [ ] `PchipInterpolator`: aliases (`pchip`, `pchip_interpolate`)
- [ ] `Akima1DInterpolator`
- [ ] `RegularGridInterpolator`: methods `linear` and `nearest`
- [ ] `RBFInterpolator`

## Second phase (details subject to change)
- [ ] `RegularGridInterpolator`: methods `cubic`, `quintic`, `slinear`, `pchip`
- [ ] `Interpn`: alias to `RegularGridInterpolator`
- [ ] `CubicBSpline`
- [ ] `splantider`, `splder`

----

##  QHull-dependent modules
- [ ] `BPoly`
- [ ] `NdPPoly`
- [ ] `make_lsq_spline`
- [ ]  `interp1d`: @ev-br states that it could be ignored (maybe?)

## "Odds and ends"
 - [x] `BarycentricInterpolator` → https://github.com/cupy/cupy/pull/6958
 - [x] `barycentric_interpolate` → https://github.com/cupy/cupy/pull/6958
 - [x] `KroghInterpolator` → https://github.com/cupy/cupy/pull/6990
 - [x] `krogh_interpolate` → https://github.com/cupy/cupy/pull/6990
 - [ ] `approximate_taylor_polynomial`
 - [ ] `lagrange` 
 - [ ] `pade`
 
### QHull-dependent
- [ ] `CloughTocher2DInterpolator
- [ ] `LinearNDInterpolator`
- [ ] `NearestNDInterpolator`
- [ ] `griddata`: a wrapper over the above three

### FITPACK code (Black-box Fortran code)
- [ ] `BivariateSpline`
- [ ] `LSQBivariateSpline`
- [ ] `LSQSphereBivariateSpline`
- [ ] `RectBivariateSpline`
- [ ] `RectSphereBivariateSpline`
- [ ] `SmoothBivariateSpline`
[...]

PR #7190, opened 2022-11-16
Add `cupyx.scipy.interpolate.RBFInterpolator`

ev-br:
Add the RBFInterpolator, as per gh-7186.

### DIFF
--- a/cupyx/scipy/interpolate/_rbfinterp.py
+++ b/cupyx/scipy/interpolate/_rbfinterp.py
@@ -245,7 +245,6 @@ def polynomial_matrix(x, powers, out):
 #            out[i, j] = cp.prod(x[i]**powers[j])
 
 
-
 # pythran export _build_system(float[:, :],
 #                              float[:, :],
 #                              float[:],
@@ -735,17 +734,9 @@ class RBFInterpolator:
         self.epsilon = epsilon
         self.powers = powers
 
-    def _chunk_evaluator(
-            self,
-            x,
-            y,
-            shift,
-            scale,
-            coeffs,
-    ):
+    def _evaluate(self, x, y, shift, scale, coeffs):
         """
-        Evaluate the interpolation while controlling memory consumption.
-        We chunk the input if we need more memory than specified.
+        Evaluate the interpolation.
 
         Parameters
         ----------
@@ -765,20 +756,14 @@ class RBFInterpolator:
         (Q, S) float ndarray
         Interpolated array
         """
-        nx, ndim = x.shape
-        if self.neighbors is None:
-            nnei = len(y)
-        else:
-            nnei = self.neighbors
-
         vec = _build_evaluation_coefficients(
-                x,
-                y,
-                self.kernel,
-                self.epsilon,
-                self.powers,
-                shift,
-                scale)
+            x,
+            y,
+            self.kernel,
+            self.epsilon,
+            self.powers,
+            shift,
+            scale)
         out = cp.dot(vec, coeffs)
         return out
 
@@ -806,13 +791,15 @@ class RBFInterpolator:
                              f"{self.y.shape[1]}.")
 
         if self.neighbors is None:
-            out = self._chunk_evaluator(
+            out = self._evaluate(
                 x,
                 self.y,
                 self._shift,
                 self._scale,
                 self._coeffs)
         else:
+            raise NotImplementedError    # XXX: needs KDTree
+
             # Get the indices of the k nearest observation points to each
             # evaluation point.
             _, yindices = self._tree.query(x, self.neighbors)
--- a/cupyx/scipy/interpolate/_rbfinterp.py
+++ b/cupyx/scipy/interpolate/_rbfinterp.py
@@ -239,10 +239,12 @@ def polynomial_vector(x, powers, out):
 
 def kernel_matrix(x, kernel_func, out):
     """Evaluate RBFs, with centers at `x`, at `x`."""
-    for i in range(x.shape[0]):
-        for j in range(i+1):
-            out[i, j] = kernel_func(cp.linalg.norm(x[i] - x[j]))
-            out[j, i] = out[i, j]
+    delta = x[None, :, :] - x[:, None, :]
+    out[...] = kernel_func(cp.linalg.norm(delta, axis=-1))
+# The above is equivalent to the original semi-scalar version:
+#        for j in range(i+1):
+#            out[i, j] = kernel_func(cp.linalg.norm(x[i] - x[j]))
+#            out[j, i] = out[i, j]
 
 
 def polynomial_matrix(x, powers, out):
@@ -251,14 +253,6 @@ def polynomial_matrix(x, powers, out):
         for j in range(powers.shape[0]):
             out[i, j] = cp.prod(x[i]**powers[j])
 
-
-# pythran export _kernel_matrix(float[:, :], str)
-def _kernel_matrix(x, kernel):
-    """Return RBFs, with centers at `x`, evaluated at `x`."""
-    out = cp.empty((x.shape[0], x.shape[0]), dtype=float)
-    kernel_func = NAME_TO_FUNC[kernel]
-    kernel_matrix(x, kernel_func, out)
-    return out
 
 
 # pythran export _polynomial_matrix(float[:, :], int[:, :])
--- a/cupyx/scipy/interpolate/_rbfinterp.py
+++ b/cupyx/scipy/interpolate/_rbfinterp.py
@@ -722,6 +722,7 @@ class RBFInterpolator:
             self._coeffs = coeffs
 
         else:
+            raise NotImplementedError
             self._tree = KDTree(y)
 
         self.y = y
@@ -799,50 +800,6 @@ class RBFInterpolator:
                 self._coeffs)
         else:
             raise NotImplementedError    # XXX: needs KDTree
-
-            # Get the indices of the k nearest observation points to each
-            # evaluation point.
-            _, yindices = self._tree.query(x, self.neighbors)
-            if self.neighbors == 1:
-                # `KDTree` squeezes the output when neighbors=1.
-                yindices = yindices[:, None]
-
-            # Multiple evaluation points may have the same neighborhood of
-            # observation points. Make the neighborhoods unique so that we only
-            # compute the interpolation coefficients once for each
-            # neighborhood.
-            yindices = cp.sort(yindices, axis=1)
-            yindices, inv = cp.unique(yindices, return_inverse=True, axis=0)
-            # `inv` tells us which neighborhood will be used by each evaluation
-            # point. Now we find which evaluation points will be using each
-            # neighborhood.
-            xindices = [[] for _ in range(len(yindices))]
-            for i, j in enumerate(inv):
-                xindices[j].append(i)
-
-            out = cp.empty((nx, self.d.shape[1]), dtype=float)
-            for xidx, yidx in zip(xindices, yindices):
-                # `yidx` are the indices of the observations in this
-                # neighborhood. `xidx` are the indices of the evaluation points
-                # that are using this neighborhood.
-                xnbr = x[xidx]
-                ynbr = self.y[yidx]
-                dnbr = self.d[yidx]
-                snbr = self.smoothing[yidx]
-                shift, scale, coeffs = _build_and_solve_system(
-                    ynbr,
-                    dnbr,
-                    snbr,
-                    self.kernel,
-                    self.epsilon,
-                    self.powers,
-                )
-                out[xidx] = self._chunk_evaluator(
-                    xnbr,
-                    ynbr,
-                    shift,
-                    scale,
-                    coeffs)
 
         out = out.view(self.d_dtype)
         out = out.reshape((nx, ) + self.d_shape)
--- a/cupyx/scipy/interpolate/_rbfinterp.py
+++ b/cupyx/scipy/interpolate/_rbfinterp.py
@@ -151,7 +151,7 @@ def _build_system(y, d, smoothing, kernel, epsilon, powers):
     # zeros with ones.
     scale[scale == 0.0] = 1.0
 
-    yeps = y*epsilon
+    yeps = y * epsilon
     yhat = (y - shift)/scale
 
     # Transpose to make the array fortran contiguous. This is required for
--- a/cupyx/scipy/interpolate/_rbfinterp.py
+++ b/cupyx/scipy/interpolate/_rbfinterp.py
@@ -225,12 +225,6 @@ NAME_TO_FUNC = {
 }
 
 
-def polynomial_vector(x, powers, out):
-    """Evaluate monomials, with exponents from `powers`, at the point `x`."""
-    for i in range(powers.shape[0]):
-        out[i] = cp.prod(x**powers[i])
-
-
 def kernel_matrix(x, kernel_func, out):
     """Evaluate RBFs, with centers at `x`, at `x`."""
     delta = x[None, :, :] - x[:, None, :]
@@ -243,18 +237,13 @@ def kernel_matrix(x, kernel_func, out):
 
 def polynomial_matrix(x, powers, out):
     """Evaluate monomials, with exponents from `powers`, at `x`."""
-    for i in range(x.shape[0]):
-        for j in range(powers.shape[0]):
-            out[i, j] = cp.prod(x[i]**powers[j])
+    pwr = x[:, None, :] ** powers[None, :, :]
+    cp.prod(pwr, axis=-1, out=out)
+# The above is equivalent to the following loop
+#    for i in range(x.shape[0]):
+#        for j in range(powers.shape[0]):
+#            out[i, j] = cp.prod(x[i]**powers[j])
 
-
-
-# pythran export _polynomial_matrix(float[:, :], int[:, :])
-def _polynomial_matrix(x, powers):
-    """Return monomials, with exponents from `powers`, evaluated at `x`."""
-    out = cp.empty((x.shape[0], powers.shape[0]), dtype=float)
-    polynomial_matrix(x, powers, out)
-    return out
 
 
 # pythran export _build_system(float[:, :],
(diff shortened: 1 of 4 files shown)


ev-br:
What is the canonical way to detect that a linear system is rank-deficient in cuda? The scipy code calls the raw LAPACK ?gesv routine and checks the `info` output for the status: https://github.com/scipy/scipy/blob/main/scipy/interpolate/_rbfinterp.py#L113
So far I tried to to use `cupy.linalg.solve` and `cupyx.lapack.gesv` and both seem to return (or fill the second argument) with `nans` --- I can of course check for the all-nan result, or is there a better way?

ev-br:
OK, I think the first pass is here, and this is ready for a first review. Or rather, this being my first attempt at anything cupy, I'm sure there are suboptimal choices and I'd appreciate a look-through from an expert.

Quick notes:
- This simply takes the scipy implementation and slightly adapts it to cupy. Not explicit kernels, just pure python.
- Tests are also just scipy test suite, adapted.
- this PR only implements a part of the scipy functionality: it looks at the whole dataset, the `neighbors=` functionality for only looking at nearest neighbors is not implemented --- it needs scipy.spatial.KDTree or an analog, so let's push it stage two.

Several specific points I'd appreciate feedback on:

- [ ] linear algebra: scipy version calls into gesv, need to replicate it here or just keep `cupy.linalg.solve` as this PR does at the moment.
- [ ] The scipy version uses evaluation in chunks to fit a memory budget. The loop over the chunks is in pure python. Is it OK like this or what is the right pattern on GPU? 
- [ ] Perf benchmark: how to profile/measure performance.

cupyx/scipy/interpolate/_rbfinterp.py
```diff
@@ -0,0 +1,738 @@
...
+
+def linear(r):
+    return -r
+
+
+def thin_plate_spline(r):
+    if r == 0:
+        return 0.0
+    else:
+        return r**2*cp.log(r)
```
emcastillo, line 20: All these small functions will result in several kernel invocations, they could be combined into a single one for performance gains :)

There is an example on how to achieve this here

https://github.com/cupy/cupy/blob/master/cupyx/scipy/special/_stats_distributions.py#L37-L88

cupyx/scipy/interpolate/_rbfinterp.py
```diff
@@ -0,0 +1,738 @@
...
+    mins = cp.min(y, axis=0)
+    maxs = cp.max(y, axis=0)
+    shift = (maxs + mins)/2
+    scale = (maxs - mins)/2
+    # The scale may be zero if there is a single point or all the points have
+    # the same value for some dimension. Avoid division by zero by replacing
+    # zeros with ones.
+    scale[scale == 0.0] = 1.0
+
+    yeps = y*epsilon
```
emcastillo, line 154: Usually per format guidelines we space operators and variables
```
yeps = y * epsilon
```

emcastillo (commented):
This is looking great :)

emcastillo:
> What is the canonical way to detect that a linear system is rank-deficient in cuda? The scipy code calls the raw LAPACK ?gesv routine and checks the info output for the status: https://github.com/scipy/scipy/blob/main/scipy/interpolate/_rbfinterp.py#L113
So far I tried to to use cupy.linalg.solve and cupyx.lapack.gesv and both seem to return (or fill the second argument) with nans --- I can of course check for the all-nan result, or is there a better way?
EDIT: okay, have found with cupyx.errstate(linalg="raise") which seems to do it. Please let me know if this is not the way.

Error checking will synchronize the device and cause heavy performance degradation, we usually disable this by default and let the user enable it when he requires checks in cases that are not going to fail easily.
As long as it is documented is fine.

cupyx/scipy/interpolate/_rbfinterp.py
ev-br: Great, thanks for the pointer! This is insanely cool. 
It does indeed make a big difference, roughly halving the time to run the tests.

I've to admit I'm somewhat confused by the invocation, especially the what goes into which tuple arguments:

```
log_ndtr = _core.create_ufunc(
    'cupyx_scipy_special_log_ndtr',
    (('f->f', 'out0 = log_ndtrf(in0)'), 'd->d'),     # not (('f->f', 'signature'),  ('d->d', signature)) ?
    'out0 = log_ndtr(in0)',                          # this tacitly assumes the same name as the ufunc?
```

Sure the example is not hard to follow when pointed at, so this comment is a gentle request for documenting the syntax :-).

ev-br:
Review comments are addressed, tests pass locally, this PR is ready for review from my side.

cupyx/scipy/interpolate/_rbfinterp.py
ev-br: <s>halving the time to run the tests</s> Scratch that. Moving python loops to ufuncs + broadcasting makes the test run locally from 70s to 2s.  In fairness, the scipy version had loops in pythranized C++, so lifting them back to pure python was bound to be slow. So the this PR is all ufuncs now.

cupyx/scipy/interpolate/_rbfinterp.py
```diff
@@ -0,0 +1,806 @@
...
+
+import cupy as cp
+from scipy.spatial import KDTree        # FIXME
+from scipy.special import comb          # FIXME
+
+
+# Define the kernel functions.
+
+kernel_definitions = """
+static __device__ double linear(double r)
```
emcastillo, line 13: How about using C++ templates to reduce the length of the boilerplate?

cupyx/scipy/interpolate/_rbfinterp.py
```diff
@@ -0,0 +1,806 @@
...
+#        for j in range(powers.shape[0]):
+#            out[i, j] = cp.prod(x[i]**powers[j])
+
+
+# pythran export _build_system(float[:, :],
+#                              float[:, :],
+#                              float[:],
+#                              str,
+#                              float,
+#                              int[:, :])
```
emcastillo, line 253:
```suggestion
```

cupyx/scipy/interpolate/_rbfinterp.py
```diff
@@ -0,0 +1,806 @@
...
+
+    # Transpose to make the array fortran contiguous.
+    rhs = cp.empty((s, p + r), dtype=float).T
+    rhs[:p] = d
+    rhs[p:] = 0.0
+
+    return lhs, rhs, shift, scale
+
+
+# pythran export _build_evaluation_coefficients(float[:, :],
```
emcastillo, line 320: ditto

emcastillo (dismissed):
LGTM! in general, some nitpicks & please add the public functions to the docs :)

https://github.com/cupy/cupy/blob/master/docs/source/reference/scipy_interpolate.rst

emcastillo:
Could we also get performance numbers?
Comparing CPU-GPU for various sizes.
You can use https://docs.cupy.dev/en/stable/reference/generated/cupyx.profiler.benchmark.html
to get cpu/gpu times easily

emcastillo:
```
ModuleNotFoundError: No module named 'scipy.stats.qmc'
```
We need to restrict scipy versions :)

ev-br:
> How about using C++ templates to reduce the length of the boilerplate?

The syntax of `cupy._core.create_ufunc` still defeats me: https://github.com/cupy/cupy/pull/7190/commits/79400389787bea61a0076002e4541b26c7098d6a, modelled after cupyx.special._xlogy.py, fails to compile and for the life of me I don't understand what the issue is.

All other comments are addressed, will post the benchmarks shortly.

ev-br:
Here's a benchmark (based on 3e8f828427e4919bd826dcb887a5eeb77d2ab556 ), using essentially @andfoy 's script from gh-7086


<details>
<summary>Benchmarking scipt</summary>

```
from itertools import product

import cupy as cp
import numpy as np
from cupy import testing

from cupyx.scipy.interpolate import RBFInterpolator as CuRBF
from scipy.interpolate import RBFInterpolator as SRBF

from scipy.stats.qmc import Halton

from tqdm import tqdm


input_sizes = [5, 10, 100, 200, 500, 1000] #, 1000, 10000, 50000, 100000, 500000, 700000]
modules = [(cp, CuRBF, 'CuPy', False), (np, SRBF, 'SciPy', True)]
kwargs = [('n_obs', [10, 100])]
measurement = {}

def gather_time(prof):
    return {
        'cpu': cpu_time,
        'gpu': gpu_time
    }


    rng = np.random.default_rng(123)
    n_obs = kwargs.get('n_obs')
    xobs = xp.asarray(2*Halton(2, seed=rng).random(n_obs) - 1)
    yobs = xp.sum(xobs, axis=1)*xp.exp(-6*xp.sum(xobs**2, axis=1))

    rbf = cls(xobs, yobs)
    xgrid = xp.meshgrid(xp.linspace(-1, 1, size[0]),
                        xp.linspace(-1, 1, size[0]), indexing='ij')
    xgrid = xp.asarray(xgrid)
    xflat = xgrid.reshape(2, -1).T

    return rbf, xflat


def closure(b, x):
    return b(x)


for kwarg_comb in kwargs:
    kwarg_comb = dict(kwarg_comb)
        for xp, cls, _id, _ in modules:
            prof = benchmark(closure, (b, x), n_repeat=10)


lines = []
for kwarg_comb in kwargs:
    lines.append(f'## `{kwarg_line}`\n')
    lines.append('| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |')
    lines.append('|:----:|:-------------:|:-------------:|:----------:|')
    for size in input_sizes:
        comp = f'  {size}'
        for _, _, _id, cpu_only in modules:
            times = size_measurement[_id]
            gpu_time = times['gpu']
            cpu_time = times['cpu']
            time_info = f'{cpu_time:3f}'
            if not cpu_only:
                time_info = f'{time_info} | {gpu_time:3f}'
        lines.append(f'| {comp} |')
    lines.append('\n')
    print('\n'.join(lines))
    lines = []
```

</details>


Here `n_obs` is the number of observations (the data the RBF interpolates) and `size` is the size of the array for evaluating the interpolant. In scikit-learn parlance, `n_obs` would be the size of the argument of `.fit`, and `size` is the size of the argument of `.predict`. In fact, the input size in bytes is quadratic in `size` because the example is two-dimensional.




## `n_obs=10`

| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |
|:----:|:-------------:|:-------------:|:----------:|
|   5 | 0.236155 | 0.241914 | 0.012085 |
|   10 | 0.238265 | 0.243597 | 0.027106 |
|   100 | 0.239196 | 2.018858 | 4.494913 |
|   200 | 0.242961 | 5.800250 | 17.691019 |
|   500 | 0.256228 | 34.579328 | 108.704841 |
|   1000 | 0.265602 | 138.023178 | 414.705756 |


## `n_obs=100`

| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |
|:----:|:-------------:|:-------------:|:----------:|
|   5 | 0.318413 | 0.324979 | 0.068589 |
|   10 | 0.309386 | 0.316157 | 0.337352 |
|   100 | 0.245368 | 8.798675 | 32.009701 |
|   200 | 0.253618 | 32.031261 | 127.304233 |
|   500 | 0.273228 | 199.664247 | 797.655235 |
|   1000 | 0.285420 | 798.268823 | 3157.704820 |

So, at large `size` this is faster than SciPy by a factor of about three to four. For still larger `size`s, the benchmarking script starts failing with MemoryErrors which look like this:

```
  File "/home/ev-br/repos/cupy/cupyx/scipy/interpolate/_rbfinterp.py", line 782, in __call__
    out = self._evaluate(
  File "/home/ev-br/repos/cupy/cupyx/scipy/interpolate/_rbfinterp.py", line 747, in _evaluate
    vec = _build_evaluation_coefficients(
  File "/home/ev-br/repos/cupy/cupyx/scipy/interpolate/_rbfinterp.py", line 353, in _build_evaluation_coefficients
    delta = xeps[:, None, :] - yeps[None, :, :]
  File "cupy/_core/core.pyx", line 1286, in cupy._core.core._ndarray_base.__sub__
    return _math._subtract(x, y)
  File "cupy/_core/_kernel.pyx", line 1291, in cupy._core._kernel.ufunc.__call__
    out_args = _get_out_args_from_optionals(
  File "cupy/_core/_kernel.pyx", line 639, in cupy._core._kernel._get_out_args_from_optionals
    out_args[i] = _ndarray_init(
  File "cupy/_core/core.pyx", line 2792, in cupy._core.core._ndarray_init
    ret._init_fast(shape, dtype, True)
  File "cupy/_core/core.pyx", line 241, in cupy._core.core._ndarray_base._init_fast
    self.data = memory.alloc(self.size * itemsize)
  File "cupy/cuda/memory.pyx", line 742, in cupy.cuda.memory.alloc
    return get_allocator()(size)
  File "cupy/cuda/memory.pyx", line 1419, in cupy.cuda.memory.MemoryPool.malloc
    cpdef MemoryPointer malloc(self, size_t size):
  File "cupy/cuda/memory.pyx", line 1440, in cupy.cuda.memory.MemoryPool.malloc
    return mp.malloc(size)
  File "cupy/cuda/memory.pyx", line 1120, in cupy.cuda.memory.SingleDeviceMemoryPool.malloc
    return self._malloc(rounded_size)
  File "cupy/cuda/memory.pyx", line 1141, in cupy.cuda.memory.SingleDeviceMemoryPool._malloc
    mem = self._try_malloc(size)
  File "cupy/cuda/memory.pyx", line 1379, in cupy.cuda.memory.SingleDeviceMemoryPool._try_malloc
    raise OutOfMemoryError(
cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 4,000,000,000 bytes (allocated so far: 5,564,229,120 bytes).
```

This is not entirely unexpected: the current version just tries to rely on broadcasting and *that* gets memory bounds quickly.
So I guess it needs to either restore the chunked evaluation that SciPy has, or a proper CUDA kernel for essentially this:

```kernel_func(cp.linalg.norm(x[None, :, :] - x[:, None, :]  ,axis=-1))```

**Comment by ev-br** [comment]
With chunking on python level, https://github.com/cupy/cupy/pull/7190/commits/e830f0cc10042532549f931db9deb9bc8e097030, there are no more MemoryErrors are gone at an expense of a speedup w.r.t. scipy being about a factor of two:

## `n_obs=10`

| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |
|:----:|:-------------:|:-------------:|:----------:|
|   5 | 0.231712 | 0.239328 | 0.012400 |
|   10 | 0.234733 | 0.239840 | 0.027750 |
|   100 | 0.235243 | 2.020096 | 4.468586 |
|   200 | 0.246832 | 6.523296 | 17.399986 |
|   500 | 0.925231 | 34.385311 | 107.815256 |
|   1000 | 1.897384 | 138.291229 | 375.835255 |
|   2000 | 1.904943 | 552.411133 | 1126.478567 |
|   5000 | 1.942894 | 3451.459473 | 6116.871658 |


## `n_obs=100`

| Size | CuPy-CPU (ms) | CuPy-GPU (ms) | SciPy (ms) |
|:----:|:-------------:|:-------------:|:----------:|
|   5 | 0.307543 | 0.314848 | 0.067981 |
|   10 | 0.305694 | 0.312768 | 0.335694 |
|   100 | 0.511716 | 11.173408 | 31.809123 |
|   200 | 1.210014 | 31.952864 | 133.608430 |
|   500 | 5.857753 | 198.755676 | 790.110258 |
|   1000 | 11.791895 | 793.093689 | 3121.488084 |
|   2000 | 11.936967 | 3194.358643 | 8549.335506 |
|   5000 | 12.018478 | 20113.894531 | 44212.444842 |

**Comment by emcastillo** [comment]
IF I am interpreting the performance numbers right

We have that the ops are non-blocking (no device synchronization 🎉) and for mid-sizes cupy can be 3-4 times faster than SciPy, for large sizes speedup falls to 2.

However, for interpolation I think the number of samples is not very large in general (I am not sure of this, just an intuition) so probably the performance is good enough. Also getting it to be non-blocking is kinda cool!
I will check the code once more just to be sure of formatting etc and give it green light!

Thanks a lot and awesome work

**Comment by emcastillo** [comment]
Seems we have to restrict the scipy versions in the tests

```
 AttributeError: module 'scipy.interpolate' has no attribute 'RBFInterpolator'
```

**Comment by ev-br** [comment]
Okay, decorated all tests with `@testing.with_requires("scipy>=1.7.0")` in the last commit (https://github.com/cupy/cupy/pull/7190/commits/1b2a9d14908ea4e240ff4fd3746b85faf19e5232), this should be the version where RBFInterpolator was introduced IIRC. Would you restart the tests please?

And, what is the scipy version compatibility requirement actually?

**Comment by emcastillo** [comment]
```
TestRBFInterpolatorNeighborsNone.test_smoothing_misfit[quintic]



16:54:30 E       Not equal to tolerance rtol=1e-07, atol=1e-09
16:54:30 E       
16:54:30 E       Mismatched elements: 1 / 100 (1%)
16:54:30 E       Max absolute difference: 2.32355823e-09
16:54:30 E       Max relative difference: 3.83837644e-07
16:54:30 E        x: array([ 0.23218 ,  0.211929, -0.452106,  0.40035 , -0.900726,  0.333872,
16:54:30 E               0.073761,  0.211888, -0.9431  ,  0.210694, -0.238658,  0.312162,
16:54:30 E              -0.758641,  0.305938,  0.170728,  0.274621, -0.676315,  0.287145,...
16:54:30 E        y: array([ 0.23218 ,  0.211929, -0.452106,  0.40035 , -0.900726,  0.333872,
16:54:30 E               0.073761,  0.211888, -0.9431  ,  0.210694, -0.238658,  0.312162,
16:54:30 E              -0.758641,  0.305938,  0.170728,  0.274621, -0.676315,  0.287145,...
```

We may need to adjust tolerances in the tests to avoid flakiness :(

ev-br:
okay: https://github.com/cupy/cupy/pull/7190/commits/2547a4af6c0001bd78acc61a9c2cfa50c93498cb

emcastillo (approved):

PR #7193, opened 2022-11-21
add flag to cupyx.profiler.benchmark to deepcopy positional arguments before calling function under test

jimmyjamison:
This PR adds a `copy_args` flag to `cupyx.profiler.benchmark` to deep copy all positional arguments before calling the function under test.

Currently `cupyx.profiler.benchmark` will call the function under test repeatedly with the same positional arguments. I ran into an issue where I wanted to benchmark a function I wrote that modifies data in place for one of it's arguments. This function does not behave the same when passed data that has already been modified, so in order to get an accurate benchmark, the the positional arguments need to be copied before each call. This seems like a feature that could be helpful for others too.

### DIFF
--- a/cupyx/profiler/_time.py
+++ b/cupyx/profiler/_time.py
@@ -82,7 +82,8 @@ class _PerfCaseResult:
 
 def benchmark(
         func, args=(), kwargs={}, n_repeat=10000, *,
-        name=None, n_warmup=10, max_duration=_math.inf, devices=None, copy_args=False):
+        name=None, n_warmup=10, max_duration=_math.inf, devices=None,
+        copy_args=False):
     """ Timing utility for measuring time spent by both CPU and GPU.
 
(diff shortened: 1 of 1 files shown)


kmaehashi:
@jimmyjamison Thanks for the pull request, but I think this kind of issue should be workarounded by the target function. Simply deep-copying might not work well in all cases; especially it won't if the argument is a `cupy.ndarray`. Also note that Python's standard `timeit` has the same issue but not offering an option to copy args.

leofang:
You could for example pass a generator as your function argument, then deepcopy is completely not needed. Every time `benchmark()` invokes your function, the generator can give you a fresh argument.

jimmyjamison:
Thanks for taking a look at this and for the feedback! 

Sorry for the delay in getting back to you all, but I was able to work around this issue in the target function like you suggested so I am going to close out this PR without merging.

PR #7195, opened 2022-11-22
add cupyx.scipy.interpolate.make_interp_spline

ev-br:
gh-7186
Based off gh-7086, will rebase on master when that one is done.
<s>This is very WIP at the moment</s>.  This is ready for a first review.

### DIFF
--- a/cupyx/scipy/interpolate/_bspline2.py
+++ b/cupyx/scipy/interpolate/_bspline2.py
@@ -3,6 +3,7 @@ import operator
 from numpy.core.multiarray import normalize_axis_index
 
 import cupy
+from cupyx.scipy import sparse
 
 from ._bspline import _get_module_func, INTERVAL_MODULE, D_BOOR_MODULE, BSpline
 
@@ -333,5 +334,27 @@ def make_interp_spline(x, y, k=3, t=None, bc_type=None, axis=0,
     # 6. Finally, solve for the coefficients.
     from cupy.linalg import solve
     coef = solve(A, rhs)
-    coef = cupy.ascontiguousarray(coef.reshape((nt,) + y.shape[1:]))
-    return BSpline(t, coef, k)
+
+    # try the sparse solver
+    matr = BSpline.design_matrix(x, t, k)
+
+    if nleft > 0:
+        matr = sparse.vstack([sparse.csr_matrix(A[:nleft, :]),
+                              matr])
+    if nright > 0:
+        matr = sparse.vstack([matr,
+                             sparse.csr_matrix(A[nleft+len(x):, :])])
+
+    from cupyx.scipy.sparse.linalg import spsolve
+    coef2 = cupy.empty((nt, extradim), dtype=y.dtype)
+    for dim in range(extradim):
+
+        # spsolve only accepts a single r.h.s.
+        if cupy.issubdtype(rhs.dtype, cupy.complexfloating):
+            coef2[:, dim] = (spsolve(matr, rhs[:, dim].real) +
+                             spsolve(matr, rhs[:, dim].imag) * 1.j) 
+        else:
+            coef2[:, dim] = spsolve(matr, rhs[:, dim])
+
+    coef2 = cupy.ascontiguousarray(coef2.reshape((nt,) + y.shape[1:]))
+    return BSpline(t, coef2, k)
--- a/cupyx/scipy/interpolate/_bspline2.py
+++ b/cupyx/scipy/interpolate/_bspline2.py
@@ -302,7 +302,7 @@ def make_interp_spline(x, y, k=3, t=None, bc_type=None, axis=0,
         d_boor_kernel = _get_module_func(D_BOOR_MODULE, 'd_boor', dummy_c)
 
         # find the intervals for x[0] and x[-1]
-        intervals_bc = cupy.empty(2, dtype=cupy.int_)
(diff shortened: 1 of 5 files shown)


ev-br:
The basic prototype is here with tests, and the main question now is what is the recommended way to do banded matrix solve? The SciPy version directly calls to LAPACK's dgbsv, https://github.com/scipy/scipy/blob/main/scipy/interpolate/_bsplines.py#L1391
What would be the CuPy way?

ev-br:
Based on the discussion in gitter, the current version constructs the colocation matrices in the CSR format and uses the a sparse linear solver from `cupyx.scipy.sparse.spsolve`. 
The code is not very pretty and not very efficient. However, this is essentially a user-facing constructor which runs only once per dataset, and evaluations are delegated to `BSpline.__call__` . So I think there is not much need to squeeze performance from this one.

Also not that this PR does not implement periodic boundary conditions yet. Will do this in a follow-up PR, this one is large enough already.

takagi:
> Based on the discussion in gitter, the current version constructs the colocation matrices in the CSR format and uses the a sparse linear solver from cupyx.scipy.sparse.spsolve.

Thanks! Using sparse matrices sounds reasonable.

cupyx/scipy/interpolate/_bspline2.py
```diff
@@ -0,0 +1,509 @@
+import operator
+
+from numpy.core.multiarray import normalize_axis_index
+
+import cupy
+from cupyx.scipy import sparse
+from cupyx.scipy.sparse.linalg import spsolve
+
+from ._bspline import _get_module_func, INTERVAL_MODULE, D_BOOR_MODULE, BSpline
```
takagi, line 9: We don't use relative module paths, so would you rewrite this to an absolute one?

ev-br:
(Some of) CI errors needed a test skip : they were testing scipy input validation bugs which were only fixed for scipy 1.10. Could you please restart the CI?

cupyx/scipy/interpolate/_bspline2.py
ev-br: Good to know, thanks! Done in https://github.com/cupy/cupy/pull/7195/commits/b08a5989960ac77e69178fc0c66cb0f81f231cee

ev-br:
I've to admit I'm baffled by the CI failures.

- Not entirely sure what is the CI structure, what is skipped on what platform --- is to correct that the failure only occurs on windows, and on linux CI this same test passes? It definitely passes for me locally under linux.
- The failing test is not very critical, so can in principle be skipped on windows.
- However, the partial CI logs seem to indicate that something very fundamental breaks with either basic numpy array operations, or in the invocation of the `d_boor` kernel from gh-7086. Specifically, the failure is 

```
00:50:39.147845 STDOUT 4088] | rows[j, left-k:left+1] = temp[:k+1] |  
00:50:39.147872 STDOUT 4088] | File "cupy\_core\core.pyx", line 1605, in cupy._core.core._ndarray_base.__setitem__ |  
00:50:39.147872 STDOUT 4088] | File "cupy\_core\_routines_indexing.pyx", line 52, in cupy._core._routines_indexing._ndarray_setitem |  
00:50:39.147872 STDOUT 4088] | File "cupy\_core\_routines_indexing.pyx", line 1027, in cupy._core._routines_indexing._scatter_op |  
00:50:39.147872 STDOUT 4088] | File "cupy\_core\_kernel.pyx", line 1256, in cupy._core._kernel.ufunc.__call__ |  
00:50:39.147872 STDOUT 4088] | File "cupy\_core\internal.pyx", line 382, in cupy._core.internal._broadcast_core |  
00:50:39.148828 STDOUT 4088] | ValueError: operands could not be broadcast together with shapes (4,) (0,) |  
00:50:39.148828 STDOUT 4088]
```

The relevant code lines are https://github.com/cupy/cupy/pull/7195/files#diff-48c02048e681d7e67fba7739a19afcc2ae95b9d77d35403e011303d2a7f20607R473

Also note that exactly the same invocation is used in non-test code paths (multiple times) and seems to not cause CI failures.

Not sure how to debug it.

takagi:
> is correct that the failure only occurs on windows, and on linux CI this same test passes? It definitely passes for me locally under Linux.

Yes, it's possible. Let me see the failure. Also, Jenkins CI fails on ROCm environment, I'm going to check it.

takagi:
As for windows, the cause is that we use `cupy.int_` for some intervals, but the kernels expect `long long`. How about using `cupy.int64` for interval ndarrays?

takagi:
Thanks for quick fix!

takagi:
I'm just about to let you know we may skip the ROCm test as #7219.

takagi (approved):

ev-br:
The last commit here skips testing on ROC, similar to https://github.com/cupy/cupy/pull/7219

Note that it means that this whole functionality is not available on ROC! This whole function is essentially building the l.h.s. for a linear algebra problem and solving it.

takagi:
> Note that it means that this whole functionality is not available on ROC! This whole function is essentially building the l.h.s. for a linear algebra problem and solving it.

Yes, and I think rocSOLVER is still under its development so it's okay for us to wait until it implements missing cuSOLVER counterparts (at least) in this PR.

ev-br:
Thank you @takagi !

PR #7197, opened 2022-11-23
Implementing RegularGridInterpolator and interpn from scipy.interpolate

ideasrule:
This is a naive implementation where I just replaced numpy with cupy and changed a few things.  I copied over the unit tests from scipy, and all of them pass.  No optimization yet.

I implemented these features to help with issue #7186, a general effort to implement all of scipy.interpolate


cupyx/scipy/interpolate/_rgi.py
```diff
@@ -0,0 +1,347 @@
...
+                 fill_value=cp.nan):
+        if method not in self._ALL_METHODS:
+            raise ValueError("Method '%s' is not defined" % method)
+
+        self.method = method
+        self.bounds_error = bounds_error
+
+        if not hasattr(values, 'ndim'):
+            # allow reasonable duck-typed values
+            values = cp.asarray(values)
```
ev-br, line 47: I think this can be omiitted: the main reason for this in scipy was to be able to deal with out-of-memory arrays (NedCDF files, hdf5 etc). Not sure if it's valuable for CuPy.

ev-br (commented):
Looks nice! This seems to be a conversion of a slightly outdated version of RegularGridInterpolator. Would be nice to pick the recent tweaks from the main branch,  where they make sense for cupy. It has fixes for several glitches and perf improvements (np.diff, integer-valued points etc).
A benchmark would indeed be great to see to gauge the need to dedicated cuda kernels.

Note however that I'm not a CuPy dev, so this is a strictly an outsider review :-).

cupyx/scipy/interpolate/_rgi.py
```diff
@@ -0,0 +1,347 @@
...
+                                         bounds_error=bounds_error,
+                                         fill_value=fill_value)
+        return interp(xi)
+    elif method == "nearest":
+        interp = RegularGridInterpolator(points, values, method="nearest",
+                                         bounds_error=bounds_error,
+                                         fill_value=fill_value)
+        return interp(xi)
+    elif method == "splinef2d":
+        raise NotImplementedError("splinef2d not implemented yet")
```
ev-br, line 347: This I think can be removed: this method in scipy dates to before `method="cubic"` was a thing and is no longer very even in scipy, let alone in CuPy going forward.

cupyx/scipy/interpolate/_rgi.py
ideasrule: Removing it would make the interpolation fail if the inputs are lists instead of cupy arrays.  This might actually be desired behavior because the other cupy functions seem to strictly accept only cupy arrays, not array-like objects or even scalars.  Can a cupy dev let us know what the desired behavior is?

cupyx/scipy/interpolate/_rgi.py
ideasrule: Thanks!  Removed.

ideasrule:
> Looks nice! This seems to be a conversion of a slightly outdated version of RegularGridInterpolator. Would be nice to pick the recent tweaks from the main branch, where they make sense for cupy. It has fixes for several glitches and perf improvements (np.diff, integer-valued points etc). A benchmark would indeed be great to see to gauge the need to dedicated cuda kernels.
> 
> Note however that I'm not a CuPy dev, so this is a strictly an outsider review :-).

I could have sworn that I used the latest version of _rgi.py, but it turns out that I might have used the oldest.  It was actually a blessing in disguise because the newest version of _rgi.py uses Cython, which is a lot harder to port (at least for me).  I went through the commit history and applied the commits that made sense.  Could you check the new code to make sure the glitches are fixed and the perf improvements are included?

cupyx/scipy/interpolate/_rgi.py
```diff
@@ -42,47 +42,65 @@ def __init__(self, points, values, method="linear", bounds_error=True,
...
+                raise ValueError("The points in dimension %d must be "
+                                 "1-dimensional" % i)
+            if not values.shape[i] == len(p):
+                raise ValueError("There are %d points and %d values in "
+                                 "dimension %d" % (len(p), values.shape[i], i))
+
+    def _check_values(self, values):
+        if not hasattr(values, 'ndim'):
+            # allow reasonable duck-typed values
+            values = cp.asarray(values)
```
ev-br, line 67: Duck-typed values come back? Also `if hasattr(values, 'dtype')` --- the whole `_check_values` method can be simplified to first call `cp.asarray` then `if not cp.issubdtype(....)` to convert integers into floats (cannot just call asarray(..., dtype=float)) because of complex-valued values.

ev-br (commented):
Looks reasonable!

Converting cython extensions is indeed a separate (large) task.
`find_indices` can leverage the corresponding kernel from the BSpline PR, the kernel for the nearest method is probably not very difficult, the linear one is tricker and can certainly be postponed to a follow-up.

ideasrule:
I got around to benchmarking the code by setting up airspeed velocity (asv), and comparing it to scipy's asv results.  I have a laptop with a Intel Core i7-9750H CPU @ 2.60GHz, and a Nvidia GTX 2060 Mobile GPU.  Here are the results with scipy's default benchmarking parameters:

[default_results_cupy.txt](https://github.com/cupy/cupy/files/10096997/default_results_cupy.txt)
[default_results_scipy.txt](https://github.com/cupy/cupy/files/10096998/default_results_scipy.txt)

At the smallest sizes (ndim=2, max_coord_size=10, nsamples=10), cupy is much slower than scipy: 2 ms vs. 54 us for interpolation, 208 us vs. 12 us for initialization.  At the largest size (ndim=4, max_coord_size=200, nsamples=10,000), the two are comparable for interpolation (5.82 ms for cupy, 4.90 ms for scipy).

I then modified the benchmarking parameters to see what happens for very large nsamples of 100,000 and 1,000,000:

[large_results_cupy.txt](https://github.com/cupy/cupy/files/10097037/large_results_cupy.txt)
[large_results_scipy.txt](https://github.com/cupy/cupy/files/10097036/large_results_scipy.txt)
 
Here, cupy significant outperforms scipy.  Interpolation for the largest size (ndim=4, max_coord_size=200, nsamples=1,000,000) runs in 37 ms on cupy, vs. 613 ms on scipy.  (The setup times are unaltered because those depend only on ndim and max_coord_size, not on the number of samples.)  Even for the smallest size in this new benchmark (ndim=2, max_coord_size=10, nsamples=100,000), cupy is 2x faster (4.02 ms vs. 2.04 ms).

Per @ev-br's suggestion, I tried incorporating the BSpline find_intervals kernel into _find_indices, but it had a negligible effect on speed.  In hindsight, I'm not surprised: the existing _find_indices calls cupy.searchsorted, which in turn calls a kernel similar to the BSpline find_intervals kernel.  I haven't tried writing my own kernel for _evaluate_nearest, but it just calls cupy.where, which is probably as fast as CUDA code gets.

I don't know if this performance is good enough or not.  It's normal for CPU code to outperform GPU code for small data sizes, but to lose dramatically for large data sizes.  For this interpolation code, the crossover point seems to be ~20,000 samples.  @takagi @ev-br Do you think the performance is acceptable as-is, or should we try to improve it further?

cupyx/scipy/interpolate/_rgi.py
ideasrule: I could remove these lines, but then the code would fail for lists.  Is this expected behavior?  I honestly don't know.  I'm leaning towards "yes", based on the rest of the code, but I was hoping someone would chime in.

ev-br: My suggestion was to always cp.asarray.

cupyx/scipy/interpolate/_rgi.py
takagi: Basically, CuPy takes `cupy.ndarray`s where NumPy/SciPy take general array-like. We'd like to avoid implicit host-to-device sync and to make it responsibility of library uses to transfer data. (However, I know, the policy is not always alive across CuPy's APIs...)

takagi:
Thanks for the benchmark and analysis, @ideasrule! It's quite fair for CuPy that it is slower than CPU with smaller data sizes because invoking GPU kernels has some overheads. The result you got with larger data is satisfying. Anyway, let me see your implementation a bit further.

cupyx/scipy/interpolate/__init__.py
```diff
@@ -3,3 +3,4 @@
 from cupyx.scipy.interpolate._polyint import KroghInterpolator  # NOQA
 from cupyx.scipy.interpolate._polyint import barycentric_interpolate  # NOQA
 from cupyx.scipy.interpolate._polyint import krogh_interpolate  # NOQA
+from cupyx.scipy.interpolate._rgi import RegularGridInterpolator,interpn
```
takagi, line 6: Would you split this line to RegularGridInterpolator and interpn to follow our convention?

takagi:
Would you add RegularGridInterpolator and interpn to the doc?
https://github.com/cupy/cupy/blob/master/docs/source/reference/scipy_interpolate.rst

cupyx/scipy/interpolate/__init__.py
ideasrule: Done

ideasrule:
> Would you add RegularGridInterpolator and interpn to the doc? https://github.com/cupy/cupy/blob/master/docs/source/reference/scipy_interpolate.rst

Done

ideasrule:
> This pull request is now in conflicts. Could you fix it @ideasrule? pray

Fixed

cupyx/scipy/interpolate/_rgi.py
```diff
@@ -0,0 +1,551 @@
...
+        domain of the input data, a ValueError is raised.
+        If False, then `fill_value` is used.
+        Default is True.
+
+    fill_value : float or None, optional
+        The value to use for points outside of the interpolation domain.
+        If None, values outside the domain are extrapolated.
+        Default is ``cp.nan``.
+
+    Methods
```
takagi, line 105: We could make the detail refer to SciPy, which would get the doc-related CI errors away:
```
.. seealso:: :class:`scipy.interpolate.RegularGridInterpolator`
```

cupyx/scipy/interpolate/_rgi.py
```diff
@@ -0,0 +1,551 @@
...
+    References
+    ----------
+    .. [1] Python package *regulargrid* by Johannes Buchner, see
+           https://pypi.python.org/pypi/regulargrid/
+    .. [2] Wikipedia, "Trilinear interpolation",
+           https://en.wikipedia.org/wiki/Trilinear_interpolation
+    .. [3] Weiser, Alan, and Sergio E. Zarantonello. "A note on piecewise
+           linear and multilinear table interpolation in many dimensions."
+           MATH. COMPUT. 50.181 (1988): 189-196.
+           https://www.ams.org/journals/mcom/1988-50-181/S0025-5718-1988-0917826-0/S0025-5718-1988-0917826-0.pdf
```
takagi, line 225: We could make the detail refer to SciPy, which would get the doc-related CI errors away:
```
.. seealso:: :class:`scipy.interpolate.RegularGridInterpolator`
```

takagi:
Looks nice to me! I'll kick the CI once you fix the docstring as I commented.

cupyx/scipy/interpolate/_rgi.py
ideasrule: > Looks nice to me! I'll kick the CI once you fix the docstring as I commented.

I fixed the docstring warnings.  Did you specifically want the line you mentioned, or did you just want the errors fixed?

ideasrule:
@takagi Could you take a look at the pull request and merge it if it looks good, before another merge conflict pops up?  Thanks!

takagi:
Thanks for your effort, @ideasrule! Let me merge this PR as the CI issues are known and not related here.

takagi (approved):

ev-br:
Thank you @ideasrule, it's a great one indeed! 

Would you be interested in continuing, taking more pieces off the board of https://github.com/cupy/cupy/issues/7186? 

Am also curious, you mentioned needing parts of cupyx.interpolate in your downstream package, which parts your package needs and what are the uses?

Issue #6359, opened 2022-01-21
Implementation of `numpy.linalg.eigvals` equivalent


N0rbert:
### Description

NumPy has a [`numpy.linalg.eigvals`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigvals.html) function, which is often used in real-life programs.
Please provide an equivalent of it for CuPy and name it `cupy.linalg.eigvals`.

emcastillo:
CuPy relies on `cuSolver` for implementing these linear algebra routines. However, there are no routines in cuSolver for non-symmetric matrices, so this requires implementing Cuda kernels or custom routines to solve this.

leofang:
I think we might be able to steal MAGMA's kernels for this purpose 😄

emcastillo:
yeah, I njust checked and actually torch uses MAGMA for this!

IvanYashchuk:
I'm sorry to disappoint but MAGMA's eigenvalue solver for non-symmetric matrices is really bad. There's no specialized GPU-only path, it uses a mix of LAPACK calls with a tiny bit of CUDA kernels. Eigenvalues are computed using LAPACK.
See lots of `lapack77` calls in https://bitbucket.org/icl/magma/src/656d51d82375e791f3f6b20f47271cdbc41d5a8a/src/dgeev.cpp

Up to about 3k x 3k matrices, it's faster to compute on CPU.

emcastillo:
Thanks @IvanYashchuk!!

I was looking into MAGMA code and saw the very same as you, I am just thinking of naively throwing this into a QR solver despite being slow just to avoid CPU sync.

emcastillo:
@N0rbert @IvanYashchuk hi

We have discussed this internally, and some people suggest that it is better to use `svd` if possible. And if there is a case in your applications where `svd` can't be used and you really need `eigvals` no matter what.

Thanks

FedorIVS:
It seems to me that `cuSolver` now has the necessary implementations for non-symmetric matrices:

https://docs.nvidia.com/cuda/cusolver/index.html#cusolverdn-dense-lapack

If I'm correct, could this request be revisited?

Thanks!

kmaehashi:
cuSOLVER does not cover this yet. See https://docs.nvidia.com/cuda/cusolver/index.html#cusolverdnxsyevd

byron158:
[...]

PR #7204, opened 2022-11-26
Add PPoly to cupyx.scipy.interpolate

andfoy:
See #7186


cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,902 @@
...
+
+    intervals = cupy.empty(xp.shape, dtype=cupy.int64)
+    interval_kernel = INTERVAL_MODULE.get_function('find_breakpoint_position')
+    interval_kernel(((xp.shape[0] + 128 - 1) // 128,), (128,),
+                    (x, xp, intervals, extrapolate, xp.shape[0], x.shape[0],
+                     ascending))
+
+    # Compute coefficient displacement stride (in elements)
+    c_shape = cupy.asarray(c.shape, dtype=cupy.int_)
+    c_strides = cupy.asarray(c.strides, dtype=cupy.int_) // c.itemsize
```
ev-br, line 339: `int_` types caused errors in https://github.com/cupy/cupy/pull/7195#issuecomment-1343939311, maybe better to pre-emptively use int64 , as suggested above?

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -869,3 +870,33 @@ def integrate(self, a, b, extrapolate=None):
...
+            If bool, determines whether to extrapolate to out-of-bounds points
+            based on first and last intervals, or to return NaNs.
+            If 'periodic', periodic extrapolation is used. Default is True.
+        """
+        if isinstance(tck, BSpline):
+            t, c, k = tck.tck
+            if extrapolate is None:
+                extrapolate = tck.extrapolate
+        else:
+            t, c, k = tck
```
ev-br, line 894: Minor: this can be simplified now that you're creating a BSpline object anyway. (In scipy, BSplines had to be retrofitted, hence this whole stanza)

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,986 @@
...
+        implementation.
+        """
+        if not scipy_found:
+            raise NotImplementedError(
+                'At the moment there is not a GPU implementation for '
+                'solve and SciPy was not found')
+
+        cpu_pp = ScPPoly(self.c.get(), self.x.get(),
+                         self.extrapolate, self.axis)
+        solution = cpu_pp.solve(y, discontinuity, extrapolate)
```
andfoy, line 924: Since https://github.com/cupy/cupy/issues/6359 has no current solution and `solve` depends on eigenvalue solve, we decided to delegate the implementation to the SciPy one, which means that the coefficients and the breakpoints are copied to CPU. Please let me know if you'd prefer to omit this method altogether

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,990 @@
...
+        return;
+    }
+
+    double xp = *&x[idx];
+    long long interval = *&intervals[idx];
+    double breakpoint = *&breakpoints[interval];
+
+    const int num_c = *&c_dims[2];
+    const long long stride_0 = *&c_strides[0];
+    const long long stride_1 = *&c_strides[1];
```
andfoy, line 169: @takagi, I noticed that sometimes some `const` pointers mutate even though they are only accessed and not modified, this has led us to define "copies" by referencing and dereferencing pointer accesses, do you think this is a correct pattern? Or are we doing something wrong when calling the kernels?

cc @ev-br

andfoy:
These are the benchmark results that compare the current `PPoly` implementation (as of [aea3be3](https://github.com/cupy/cupy/pull/7204/commits/aea3be3b4dfdd75e2147d6cae50c649b6dad616e)) against the SciPy one. For larger inputs, CuPy implementation presents a ~30X improvement over the Scipy one.

<details>

```python
from itertools import product

import cupy as cp
import numpy as np
from cupy import testing

from cupyx.scipy.interpolate import PPoly as CuPPoly
from scipy.interpolate import PPoly as CPUPPoly

from tqdm import tqdm


input_sizes = [5, 10, 100, 1000, 5000, 10000, 50000, 100000, 500000, 700000]
modules = [(cp, CuPPoly, 'CuPy'), (np, CPUPPoly, 'SciPy')]
kwargs = [('extrapolate', [True, False, 'periodic'])]
measurement = {}


def gather_time(prof):


    order = 3
    rand_sample = testing.shaped_random((30,), xp, scale=1)
    x = xp.unique(xp.r_[0, 10 * rand_sample, 10])
    c = testing.shaped_random((order + 1, len(x) - 1, 2, 3), xp, scale=2) - 1
    b = cls(c, x, **kwargs)
    x = testing.shaped_random(size, xp)
    return b, x


def closure(b, x):
    return b(x)


for kwarg_comb in kwargs:
    kwarg_comb = dict(kwarg_comb)
        for xp, cls, _id in modules:


lines = []
for kwarg_comb in kwargs:
    lines.append('| Size | CuPy | SciPy (ms) | Speedup |')
    lines.append('|:----:|:----:|:----------:|:-------:|')
    for size in input_sizes:
        comp = f'{size}'
        times = []
        for _, _, _id in modules:
            time = size_measurement[_id]
            times.append(time)
            time_info = f'{time:3f}'
        speedup = times[1] / times[0]
        lines.append(f'| {comp} |')
    lines.append('\n')


print('\n'.join(lines))
```

</details>

### `extrapolate=True`

| Size | CuPy | SciPy (ms) | Speedup |
|:----:|:----:|:----------:|:-------:|
| 5 | 0.320933 | 0.011598 | 0.036139 |
| 10 | 0.337176 | 0.012092 | 0.035864 |
| 100 | 0.319225 | 0.017970 | 0.056293 |
| 1000 | 0.317063 | 0.086160 | 0.271744 |
| 5000 | 0.330330 | 0.387660 | 1.173552 |
| 10000 | 0.321229 | 0.762014 | 2.372179 |
| 50000 | 0.409374 | 3.769140 | 9.207074 |
| 100000 | 0.481331 | 7.516434 | 15.615940 |
| 500000 | 1.368281 | 37.582210 | 27.466735 |
| 700000 | 1.691455 | 55.106748 | 32.579493 |


### `extrapolate=False`

| Size | CuPy | SciPy (ms) | Speedup |
|:----:|:----:|:----------:|:-------:|
| 5 | 0.373357 | 0.011303 | 0.030275 |
| 10 | 0.316492 | 0.011930 | 0.037695 |
| 100 | 0.316866 | 0.018098 | 0.057114 |
| 1000 | 0.339481 | 0.085620 | 0.252207 |
| 5000 | 0.316303 | 0.387208 | 1.224170 |
| 10000 | 0.319142 | 0.761279 | 2.385392 |
| 50000 | 0.415272 | 3.765009 | 9.066375 |
| 100000 | 0.480709 | 7.515335 | 15.633852 |
| 500000 | 1.365497 | 37.558548 | 27.505412 |
| 700000 | 1.628276 | 55.119626 | 33.851515 |


### `extrapolate=periodic`

| Size | CuPy | SciPy (ms) | Speedup |
|:----:|:----:|:----------:|:-------:|
| 5 | 0.450037 | 0.018703 | 0.041558 |
| 10 | 0.389540 | 0.019247 | 0.049411 |
| 100 | 0.417161 | 0.026111 | 0.062592 |
| 1000 | 0.390178 | 0.099643 | 0.255379 |
| 5000 | 0.406433 | 0.423225 | 1.041314 |
| 10000 | 0.394978 | 0.824831 | 2.088296 |
| 50000 | 0.497388 | 4.054940 | 8.152476 |
| 100000 | 0.553875 | 8.070541 | 14.571047 |
| 500000 | 1.579735 | 40.325218 | 25.526576 |
| 700000 | 1.977624 | 59.258549 | 29.964523 |

cupyx/scipy/interpolate/_interpolate.py
takagi: Thanks for the suggestion, but I'd prefer to let this method just raise `NotImplementedError` (unconditionally if scipy is there or not) to avoid data transfer and computation on CPU that users can not appearently find.

tests/cupyx_tests/scipy_tests/interpolate_tests/test_ppoly.py
```diff
@@ -0,0 +1,748 @@
...
+
+        # c array needs two axes for the coefficients and intervals, so
+        # 0 <= axis < c.ndim-1; raise otherwise
+        instance = cls(c=c, x=x, axis=axis)
+        return instance.c, instance.x, instance.axis
+
+
+@testing.with_requires("scipy")
+class TestPPoly:
+    @testing.numpy_cupy_allclose(scipy_name='scp')
```
takagi, line 213: How about `@pytest.mark.parametrize('xp,scp', [(numpy, scipy), (cupy, cupyx.scipy)])`?

tests/cupyx_tests/scipy_tests/interpolate_tests/test_ppoly.py
```diff
@@ -0,0 +1,748 @@
...
+    def test_simple(self, xp, scp):
+        c = xp.array([[1, 4], [2, 5], [3, 6]])
+        x = xp.array([0, 0.5, 1])
+        p = scp.interpolate.PPoly(c, x)
+        testing.assert_allclose(p(0.3), 1*0.3**2 + 2*0.3 + 3)
+        testing.assert_allclose(p(0.7), 4*(0.7-0.5)**2 + 5*(0.7-0.5) + 6)
+        return True
```
takagi, line 220: Does the following work with `numpy_cupy_allclose`?
```suggestion
        return [p(0.3), p(0.7)]
```

tests/cupyx_tests/scipy_tests/interpolate_tests/test_ppoly.py
```diff
@@ -0,0 +1,748 @@
...
+        lst = []
+        for writeable in (True, False):
+            x.flags.writeable = writeable
+            f = scp.interpolate.PPoly(c, x)
+            vals = f(xnew)
+            assert xp.isfinite(vals).all()
+            lst.append(vals)
+        return vals
+
+    @testing.numpy_cupy_allclose(scipy_name='scp')
```
takagi, line 252: How about `@pytest.mark.parametrize('xp,scp', [(numpy, scipy), (cupy, cupyx.scipy)])`?

takagi:
Would you use the following or something like that in tests that want to make assertion in cupy and numpy/scipy instad of using `numpy_cupy_allclose`?

```
@pytest.mark.parametrize('xp,scp', [(numpy, scipy), (cupy, cupyx.scipy)])
```

cupyx/scipy/interpolate/_interpolate.py
takagi: Do you have the compilation failed if you just pass c_strides to `eval_poly_1`?

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,990 @@
+
+import cupy
+from cupy._core import internal  # NOQA
+from cupy._core._scalar import get_typename  # NOQA
+from cupyx.scipy import special as spec
+from ._bspline import BSpline
```
takagi, line 6: Would you use an absolute path here?

cupyx/scipy/interpolate/_interpolate.py
andfoy: Not a compilation error, but a runtime failure where the first value of `c_strides` changes to a random value.

takagi: It is supposed that it's a kind of synchronization problem and the memory area is unintentionally broken. Copying the value to register files is not the solution. Let me see it further.

tests/cupyx_tests/scipy_tests/interpolate_tests/test_ppoly.py
```diff
@@ -139,8 +158,12 @@ def test_shape_2(self, xp, scp, cls):
         assert p(0.5).shape == ()
         assert p(xp.array(0.5)).shape == ()
 
-        xxx = xp.array([[0.1, 0.2], [0.4]], dtype=object)
-        p(xxx)   # raises ValueError
+        try:
+            xxx = xp.array([[0.1, 0.2], [0.4]], dtype=object)
+            p(xxx)   # raises ValueError
+            assert False
+        except ValueError:
```
ev-br, line 165: ` pytest.raises(ValueError) `?

tests/cupyx_tests/scipy_tests/interpolate_tests/test_ppoly.py
```diff
@@ -0,0 +1,766 @@
...
+        assert_allclose(P.integrate(3.5, 5 + 12),
+                        poly_int(2) - poly_int(1) + poly_int(4) -
+                        poly_int(3.5) + 4 * period_int)
+
+        assert_allclose(P.integrate(0, -1), poly_int(2) - poly_int(3))
+        assert_allclose(P.integrate(-9, -10), poly_int(2) - poly_int(3))
+        assert_allclose(
+            P.integrate(0, -10), poly_int(2) - poly_int(3) - 3 * period_int)
+
+    # TODO: Add a solve implementation
```
ev-br, line 529: Maybe better `pytest.skip` this and other tests of roots and solve methods? While at it, maybe the the comment/skip message could be made more specific and point to the lack of non-symmetric eigenvalue solver in CUDA.

cupyx/scipy/interpolate/_interpolate.py
ev-br: I wonder if gh-7240 is related.

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,972 @@
...
+        piecewise and may have discontinuities.
+    extrapolate : bint
+        Whether to extrapolate to out-of-bounds points based on first
+        and last intervals, or to return NaNs.
+    out : ndarray, shape (r, n)
+        Value of each polynomial at each of the input points.
+        This argument is modified in-place.
+    """
+    # Determine if the breakpoints are in ascending order or descending one
+    ascending = (x[x.shape[0] - 1] >= x[0]).item()
```
takagi, line 328: We don't need `item()` and transfer the value to CPU as `asending` will be used on GPU in `interval_kernel`.

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,972 @@
...
+    interval_kernel(((xp.shape[0] + 128 - 1) // 128,), (128,),
+                    (x, xp, intervals, extrapolate, xp.shape[0], x.shape[0],
+                     ascending))
+
+    # Compute coefficient displacement stride (in elements)
+    c_shape = cupy.asarray(c.shape, dtype=cupy.int64)
+    c_strides = cupy.asarray(c.strides, dtype=cupy.int64) // c.itemsize
+
+    ppoly_kernel = _get_module_func(PPOLY_MODULE, 'eval_ppoly', c)
+    ppoly_kernel(((xp.shape[0] + 128 - 1) // 128,), (128,),
```
takagi, line 341: It looks like that `ppoly_kernel` is in parallel on `r`, but how about parallelizing it on `r * n` with which you don't need to use `c.shape` and `c.strides` on GPU and, I think, makes things easier?

cupyx/scipy/interpolate/_interpolate.py
andfoy: I tried it, but it always was false

cupyx/scipy/interpolate/_interpolate.py
andfoy: I tried to divide the kernel into block sizes of (128, 2) such that the kernel operates over both the input and `c.shape[2]`. However, this reduces the performance by 2X on the benchmark

takagi: Okay, thanks!

cupyx/scipy/interpolate/_interpolate.py
takagi: You mean when you passed `ascending` to `find_breakpoint_position ` kernel without using `item()`, the value of `asc` the kernel read was always false?

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,972 @@
...
+    out[idx] = left;
+}
+}
+'''
+
+INTERVAL_MODULE = cupy.RawModule(
+    code=INTERVAL_KERNEL, options=('-std=c++11',),)
+
+PPOLY_KERNEL = r"""
+#include <cuda_runtime.h>
```
takagi, line 85: Would you use `hip_runtime.h` instead of `cuda_runtime.h` on ROCm? You can use `cupy_backends.cuda.api.runtime.is_hip` to find if the environment is CUDA or ROCm.

takagi: ah, I'll make the fix!

cupyx/scipy/interpolate/_interpolate.py
andfoy: Yes, that was the behavior I saw

cupyx/scipy/interpolate/_interpolate.py
takagi: I checked it on my local. I passed `c_strides[0]` and `c_strides[1]` instead of `stride_0` and `stride_1` to `eval_poly_1()`, but I didn't see the behavior and the tests passed (on CUDA 11.8).

takagi: > I wonder if https://github.com/cupy/cupy/issues/7240 is related.

We suspect that the cause of gh-7240 is that scipy returns zero or nan depending on its version (right?). So I guess it is not related here.

takagi: Now I'm not confident with my [comment](https://github.com/cupy/cupy/pull/7204#discussion_r1050275618). Let us leave the lines as far as the CI passes, and revisit them when we have any problems.

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,977 @@
...
+        There are `n` polynomials in each interval.
+        Coefficient of highest order-term comes first.
+    x : ndarray, shape (m+1,)
+        Breakpoints of polynomials.
+    xp : ndarray, shape (r,)
+        Points to evaluate the piecewise polynomial at.
+    dx : int
+        Order of derivative to evaluate.  The derivative is evaluated
+        piecewise and may have discontinuities.
+    extrapolate : bint
```
takagi, line 325:
```suggestion
    extrapolate : bool
```

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,977 @@
...
+    ----------
+    c : ndarray, shape (k, m, n)
+        Coefficients local polynomials of order `k-1` in `m` intervals.
+    x : ndarray, shape (m+1,)
+        Breakpoints of polynomials
+    a : double
+        Start point of integration.
+    b : double
+        End point of integration.
+    extrapolate : bint, optional
```
takagi, line 391: ditto.
```suggestion
    extrapolate : bool, optional
```

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,977 @@
...
+    # Determine if the breakpoints are in ascending order or descending one
+    ascending = (x[x.shape[0] - 1] >= x[0]).item()
+
+    a = cupy.asarray([a], dtype=cupy.float64)
+    b = cupy.asarray([b], dtype=cupy.float64)
+    if not ascending:
+        a, b = b, a
+
+    start_interval = cupy.empty(a.shape, dtype=cupy.int_)
+    end_interval = cupy.empty(b.shape, dtype=cupy.int_)
```
takagi, line 408: The kernel takes `long long*` argument, but `cupy.int_` is 32bit on windows.
```suggestion
    start_interval = cupy.empty(a.shape, dtype=cupy.int64)
    end_interval = cupy.empty(b.shape, dtype=cupy.int64)
```

cupyx/scipy/interpolate/_interpolate.py
```diff
@@ -0,0 +1,977 @@
...
+    interval_kernel(((a.shape[0] + 128 - 1) // 128,), (128,),
+                    (x, a, start_interval, extrapolate, a.shape[0], x.shape[0],
+                     ascending))
+    interval_kernel(((b.shape[0] + 128 - 1) // 128,), (128,),
+                    (x, b, end_interval, extrapolate, b.shape[0], x.shape[0],
+                     ascending))
+
+    # Compute coefficient displacement stride (in elements)
+    c_shape = cupy.asarray(c.shape, dtype=cupy.int_)
+    c_strides = cupy.asarray(c.strides, dtype=cupy.int_) // c.itemsize
```
takagi, line 420: ditto.
```suggestion
    c_shape = cupy.asarray(c.shape, dtype=cupy.int64)
    c_strides = cupy.asarray(c.strides, dtype=cupy.int64) // c.itemsize
```

cupyx/scipy/interpolate/_interpolate.py
takagi: Thanks. I've also tried to see if we could eliminate the D2H sync. Would you try the following patch to check if it works for you?
https://gist.github.com/takagi/bfd7b5ef66ad69af617ed1ceb9ec3c0f

takagi:
The CI failure on Windows is not related to this.

cupyx/scipy/interpolate/_interpolate.py
andfoy: Thanks for the patch, I've just tested it, and it works locally (the tests pass)

takagi (approved):
LGTM! I'll merge this after the CI finishes.

takagi:
Thanks, @andfoy!

cupyx/scipy/interpolate/_interpolate.py
florianquintes: Hi, I might be wrong, but I believe `numpy.linalg.eigvals` is now implemented. Could `.solve` be implemented next?

Issue #6359 was closed as "completed" in #8854, but I couldn’t find a corresponding issue for `solve`. This would be a very useful feature, and I suspect it might have been overlooked.

This is one of my first public GitHub comments—please let me know if this is the right place to suggest this.

ev-br: Yes, now that `cupy.linalg.eigvals` is available, `PPoly.solve` can be implemented (and `cupy.roots` can be fixed, and  a route-through-numpy workaround in `cupyx.scipy.signal` can go). Interested in sending a PR for any one of these?

> This is one of my first public GitHub comments—please let me know if this is the right place to suggest this.

A closed PR is probably among least visible places for new comments! I'd expect that a new issue (or better, a new PR!) would have better visibility.

florianquintes: Thanks for the fast reply. I will open a new issue.

PR #7205, opened 2022-11-27
Adding uniform() to random generator, since numpy has this functionality

ideasrule:
Implementing this method: https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.uniform.html

### DIFF
--- a/cupy/random/_generator_api.pyx
+++ b/cupy/random/_generator_api.pyx
@@ -227,7 +227,7 @@ class Generator:
         # cython cant call astype with the default values for
         # omitted args.
         return (<object>y).astype(dtype, copy=False)
-    
+
     def integers(
             self, low, high=None, size=None,
             dtype=numpy.int64, endpoint=False):
--- a/cupy/random/_generator_api.pyx
+++ b/cupy/random/_generator_api.pyx
@@ -159,6 +159,75 @@ class Generator:
         # omitted args.
         return (<object>y).astype(dtype, copy=False)
 
+    def uniform(self, low=0.0, high=1.0, size=None, dtype=numpy.float64):
+        """
+        Draw samples from a uniform distribution.
+        Samples are uniformly distributed over the half-open interval
+        ``[low, high)`` (includes low, but excludes high).  In other words,
+        any value within the given interval is equally likely to be drawn
+        by `uniform`.
+
+        Parameters
+        ----------
+        low : float or array_like of floats, optional
+            Lower boundary of the output interval.  All values generated will
+            be greater than or equal to low.  The default value is 0.
+        high : float or array_like of floats
+            Upper boundary of the output interval.  All values generated will
+            be less than high.  The high limit may be included in the returned
+            array of floats due to floating-point rounding in the equation
+            ``low + (high-low) * random()``.  high - low must be
+            non-negative.  The default value is 1.0.
+        size : int or tuple of ints, optional
+            Output shape.  If the given shape is, e.g., ``(m, n, k)``, then
+            ``m * n * k`` samples are drawn.  If size is ``None`` (default),
+            a single value is returned if ``low`` and ``high`` are both
+            scalars.  Otherwise, ``cupy.broadcast(low, high).size`` samples are
+            drawn.
+
+        Returns
+        -------
+        out : ndarray or scalar
+            Drawn samples from the parameterized uniform distribution.
+        See Also
+        --------
+        - :meth:`numpy.random.Generator.uniform`
+        - :meth:`integers`: Discrete uniform distribution, yielding integers.
+        - :meth:`random`: Floats uniformly distributed over ``[0, 1)``.
+        """
+
+        cdef _ndarray_base y
+
+        if not isinstance(low, _ndarray_base):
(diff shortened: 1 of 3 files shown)


cupy/random/_generator_api.pyx
```diff
@@ -159,6 +159,75 @@ class Generator:
         # omitted args.
         return (<object>y).astype(dtype, copy=False)
 
+    def uniform(self, low=0.0, high=1.0, size=None, dtype=numpy.float64):
```
asi1024, line 162: How about returning `low + (high - low) * self.random()` simply?

tests/cupy_tests/random_tests/common_distributions.py
```diff
@@ -127,6 +127,23 @@ def _check_ks(
...
+
+class Uniform:
+    target_method = 'uniform'
+
+    def test_uniform(self):
+        print("testing uniform")
+        result = self.generate(self.low, self.high, self.size)
+        assert cupy.all(result >= cupy.asarray(self.low).min())
+        assert cupy.all(result < cupy.asarray(self.high).max())
+
```
asi1024, line 145: Could you add `test_uniform_ks` that checks the equality of distributions between numpy and cupy with Two-sample Kolmogorov-Smirnov Test?

tests/cupy_tests/random_tests/common_distributions.py
```diff
@@ -127,6 +127,23 @@ def _check_ks(
...
+    {'low': cupy.array([1, 2]), 'high': 3, 'size': None},
+    {'low': 20, 'high': 20.1, 'size': 1000}
+]
+
+
+class Uniform:
+    target_method = 'uniform'
+
+    def test_uniform(self):
+        print("testing uniform")
```
asi1024, line 141:
```suggestion
```

tests/cupy_tests/random_tests/common_distributions.py
```diff
@@ -127,6 +127,23 @@ def _check_ks(
             raise AssertionError(message)
 
 
+uniform_params = [
+    {'low': 1, 'high': 10.0, 'size': (3, 5)},
+    {'low': cupy.array([1, 2]), 'high': 3, 'size': None},
```
asi1024, line 132: Could you move cupy.array creation into `test_uniform` to avoid GPU memory allocation outside of the test? (see `Beta.test_beta`)
```suggestion
    {'low': [1, 2], 'high': 3, 'size': None},
```

cupy/random/_generator_api.pyx
ideasrule: That was my initial idea, but the problem is that it doesn't match numpy's behavior.  numpy allows mismatched inputs, e.g. "low" can be a float and "high" can be an array of length 10, in which case the output would be an array of length 10.

tests/cupy_tests/random_tests/common_distributions.py
ideasrule: Fixed

tests/cupy_tests/random_tests/common_distributions.py
ideasrule: Fixed

cupy/random/_generator_api.pyx
```diff
@@ -159,6 +159,75 @@ class Generator:
...
+                low = cupy.asarray(low, numpy.float64)
+            else:
+                raise TypeError('low is required to be a cupy.ndarray'
+                                ' or a scalar')
+        if not isinstance(high, _ndarray_base):
+            if type(high) in (float, int):
+                high = cupy.asarray(high, numpy.float64)
+            else:
+                raise TypeError('high is required to be a cupy.ndarray'
+                                ' or a scalar')
```
asi1024, line 212: The given `low` and `high` might be a numpy scalar type (e.g., `numpy.float64(10)`.) How about writing as follows?
```suggestion
        low = cupy.asarray(low)
        high = cupy.asarray(high)
```

cupy/random/_generator_api.pyx
```diff
@@ -159,6 +159,75 @@ class Generator:
...
+
+        if size is not None and not isinstance(size, tuple):
+            size = (size, )
+        elif size is None:
+            size = cupy.broadcast(low, high).shape
```
asi1024, line 217: `_core.ndarray()` accepts scalar integer type as `size`.
```suggestion
        if size is None:
            size = cupy.broadcast(low, high).shape
```

cupy/random/_generator_api.pyx
ideasrule: Fixed

cupy/random/_generator_api.pyx
ideasrule: Great, I made this change.  But do you know why the original numpy code doesn't do this?  numpy's ndarray also accepts scalar integer types as size.

asi1024: Not sure but maybe because of historical reasons :thinking:

asi1024 (approved):

asi1024:
@ideasrule LGTM! Thank you for your contribution!

PR #7210, opened 2022-11-29
Expose CuPy allocator

emcastillo:
This will allow to use CuPy with PyTorch pluggable allocator https://github.com/pytorch/pytorch/pull/86786

```python
import torch
import cupy

# Initialize CuPy
cupy.zeros(10)

# Use cupy allocator in torch
cupy_allocator_so = 'cupy/cuda/memory.cpython-39-x86_64-linux-gnu.so'
new_alloc = torch.cuda.memory.CUDAPluggableAllocator(cupy_allocator_so, 'cupy_malloc_ext', 'cupy_free_ext')

torch.cuda.memory.change_current_allocator(new_alloc)
print('new allocator changed')
print(torch.zeros(10, device='cuda'))
print('done')
```


leofang:
I was visionary 😄 (https://github.com/cupy/cupy/pull/3212).

leofang:
@emcastillo Should we add a simple dlopen test to ensure the symbols exist? If you change this line in you example from
```python
cupy_allocator_so = 'cupy/cuda/memory.cpython-39-x86_64-linux-gnu.so'
```
to 
```python
cupy_allocator_so  = cupy.cuda.memory.__file__
```
a test based on `ctypes.CDLL` can be done programmatically.

emcastillo:
test added! thanks @leofang :)

cupy/cuda/memory.pyx
```diff
@@ -1973,3 +1973,22 @@ cdef class PythonFunctionAllocator:
...
+cdef dict _allocations = dict()
+
+
+cdef public void* cupy_malloc_ext(
+        ssize_t size, int device, void* stream) with gil:
+    # TODO(ecastill) dynamically change stream & device?
+    cdef curr_stream = stream_module.get_current_stream().ptr
+    if <intptr_t>stream != <intptr_t>curr_stream:
+        raise RuntimeError(f'stream must be the current one')
+    mem = alloc(size)
```
leofang, line 1987: I have two conflicting concerns here:
1. Shouldn't the (user-provided) `stream` be honored? Currently this assumes users manage CUDA streams using CuPy's machinery (ex: `with stream as s:`), which may or may not be feasible when an external library is hooked with CuPy's pool.
2. On the other hand, if `stream` is not the (CuPy) current stream, `alloc()` is incorrect because it places the chunk under the wrong stream, potentially breaking the stream ordering. If 1 is addressed, we'd need to do
```python
with ExternalStream(ptr=stream):
    mem = alloc(size)
```
instead. The current constraint (that `stream` is made current) makes the situation a bit easier.

cupy/cuda/memory.pyx
```diff
@@ -1973,3 +1973,22 @@ cdef class PythonFunctionAllocator:
...
+    _allocations[mem.ptr] = mem
+    return <void *>mem.ptr
```
leofang, line 1989: This is safe because `Memory.ptr` is a cdef class attribute?

cupy/cuda/memory.pyx
emcastillo: You are right :), I just sent an attempt to do it

cupy/cuda/memory.pyx
emcastillo: I hope so!

kmaehashi (changes requested):

cupy/cuda/memory.pyx
kmaehashi: We originally thought of asking users to use [`ppe.cuda.stream()`](https://github.com/pfnet/pytorch-pfn-extras/blob/5c8be80b1f70719e501e01ef2b67f620aad3ad46/pytorch_pfn_extras/cuda/_allocator.py#L13) (which is an utility we use to switch both PyTorch/CuPy streams), but I agree that there is a case of hooking external libraries that cannot be modified. Let's honor the stream provided.

cupy/cuda/memory.pyx
kmaehashi: `alloc()` returns `MemoryPointer` which have `readonly intptr_t ptr`, so this should be OK :)

tests/cupy_tests/cuda_tests/test_memory.py
```diff
@@ -1342,3 +1342,11 @@ def test_set_limit_fraction(self):
...
         with self.assertRaises(ValueError):
             self.pool.set_limit(fraction=1.1)
+
+
+class TestExternalAllocatorSymbols:
+    def test_exported_symbols(self):
+        cupy_allocator_so = cupy.cuda.memory.__file__
+        allocator = ctypes.CDLL(cupy_allocator_so)
+        getattr(allocator, 'cupy_malloc_ext')
+        getattr(allocator, 'cupy_free_ext')
```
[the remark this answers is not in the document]
emcastillo: done!

cupy/cuda/memory.pyx
```diff
@@ -1973,3 +1973,21 @@ cdef class PythonFunctionAllocator:
...
+
+
+cdef dict _allocations = dict()
+
+
+cdef public void* cupy_malloc_ext(
+        ssize_t size, int device, void* stream) with gil:
+    # TODO(ecastill) dynamically change stream & device?
+    cdef curr_stream = stream_module.get_current_stream().ptr
+    with stream_module.ExternalStream(ptr=curr_stream):
```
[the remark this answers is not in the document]
emcastillo: added stream and device changes :)

cupy/cuda/memory.pyx
```diff
@@ -1973,3 +1973,21 @@ cdef class PythonFunctionAllocator:
             size, self._malloc_func,
             self._free_func, device.get_device_id())
         return MemoryPointer(mem, 0)
+
+
+cdef dict _allocations = dict()
+
+
+cdef public void* cupy_malloc_ext(
```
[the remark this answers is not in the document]
emcastillo: returns nullptr on exception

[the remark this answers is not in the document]
leofang: This function holds the GIL, so it's OK to raise.

cupy/cuda/memory.pyx
```diff
@@ -1973,3 +1973,28 @@ cdef class PythonFunctionAllocator:
...
+        ssize_t size, int device_id, void* stream) with gil:
+    # TODO(ecastill) dynamically change stream & device?
+    try:
+        with device.Device(device_id):
+            with stream_module.ExternalStream(ptr=<intptr_t> stream):
+                mem = alloc(size)
+    except Exception:
+        return <void*>0  # return nullptr
+    cdef intptr_t ptr = <intptr_t> mem.ptr
```
leofang, line 1990: Is this conversion necessary? I would think the Python dict converts `ptr` back to a Python int so that it can be used as a key?

cupy/cuda/memory.pyx
```diff
@@ -1973,3 +1973,28 @@ cdef class PythonFunctionAllocator:
...
+    _allocations[ptr] = mem
+    return <void*>mem.ptr
+
+
+cdef public void cupy_free_ext(
+        void* ptr, ssize_t size, int device_id, void* stream) with gil:
+    try:
+        del _allocations[<intptr_t>ptr]
+    except Exception:
+        pass   # Silently ignore non-tracked values?
```
leofang, line 2000: hmmm...

tests/cupy_tests/cuda_tests/test_memory.py
```diff
@@ -1342,3 +1342,43 @@ def test_set_limit_fraction(self):
...
+
+class TestExternalAllocator(unittest.TestCase):
+    def setUp(self):
+        self.pool = memory.MemoryPool(memory._malloc)
+        self.old_pool = cupy.get_default_memory_pool()
+        memory.set_allocator(self.pool.malloc)
+
+    def tearDown(self):
+        self.pool.free_all_blocks()
+        memory.set_allocator(self.old_pool.malloc)
```
leofang, line 1355: I think we don't wanna use `unittest` anymore. I didn't test but this should work:
```suggestion


@pytest.fixture()
def ext_pool():
    old_pool = cupy.get_default_memory_pool()
    new_pool = memory.MemoryPool(memory._malloc)
    memory.set_allocator(new_pool.malloc)
    yield new_pool
    new_pool.free_all_blocks()
    memory.set_allocator(old_pool.malloc)


class TestExternalAllocator:
```
and pass `ext_pool` to all test methods as an argument (ex: `def test_malloc(self, ext_pool):`).

cupy/cuda/memory.pyx
emcastillo: I had lots of compilation errors with the `void*` & `int` related conversions ...
I am not sure if this is strictly necessary now, but will double check

cupy/cuda/memory.pyx
emcastillo: we will be modifying pytorch allocator to retrieve the error instead of dealing with exceptions here

leofang: Great, is there a PR you can link to? 🙂

kmaehashi:

PR #7214, opened 2022-11-29
Fix cusparseSpSM compatibility

takagi:
CUDA 11.x basically have minor version compatibility. However, they have some APIs that are introduced in the middle of 11.x minor versions, which APIs do not exist in CUDA toolkit before they're introduced. In such a case we have to dlopen the APIs at runtime instead of (dynamic-)link them at build-time against the latest 11.x minor version. Otherwise, the process complains when the depending (cython-origin) shared object is loaded.

### DIFF
--- a/cupy_backends/cuda/cupy_cusparse.h
+++ b/cupy_backends/cuda/cupy_cusparse.h
@@ -645,21 +645,6 @@ cusparseStatus_t cusparseSpSM_solve(...) {
 
 #endif // #if CUSPARSE_VERSION < 11600
 
-// See cusparse.pyx
-cusparseStatus_t _cusparseSpSM_solve(cusparseHandle_t     handle,
-                                     cusparseOperation_t  opA,
-                                     cusparseOperation_t  opB,
-                                     const void*          alpha,
-                                     cusparseSpMatDescr_t matA,
-                                     cusparseDnMatDescr_t matB,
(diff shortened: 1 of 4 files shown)


kmaehashi:
Please remove these stubs as well, as this will cause duplicated symbols. LGTM otherwise!


```
cusparseStatus_t cusparseSpMatSetAttribute(...) {   
  return CUSPARSE_STATUS_SUCCESS;                   
}
```

```
cusparseStatus_t cusparseSpSM_createDescr(...) {    
  return CUSPARSE_STATUS_SUCCESS;                   
}                                                   
                                                    
cusparseStatus_t cusparseSpSM_destroyDescr(...) {   
  return CUSPARSE_STATUS_SUCCESS;                   
}                                                   
                                                    
cusparseStatus_t cusparseSpSM_bufferSize(...) {     
  return CUSPARSE_STATUS_SUCCESS;                   
}                                                   
                                                    
cusparseStatus_t cusparseSpSM_analysis(...) {       
  return CUSPARSE_STATUS_SUCCESS;                   
}                                                   
                                                    
cusparseStatus_t cusparseSpSM_solve(...) {          
  return CUSPARSE_STATUS_SUCCESS;                   
}                                                   
```

takagi:
Thanks! I've removed them.

kmaehashi (approved):

PR #7215, opened 2022-11-29
Expose splder and splantider

andfoy:
This is a minor change missing from https://github.com/cupy/cupy/pull/7086


takagi (approved):
LGTM!

PR #7218, opened 2022-12-05
Remove scipy import

takagi:
Follows up #7190, which causes CI failure in cupy.linux.cuda111 and cupy.linux.cuda111.multi

### DIFF
--- a/cupyx/scipy/interpolate/_rbfinterp.py
+++ b/cupyx/scipy/interpolate/_rbfinterp.py
@@ -4,7 +4,7 @@ import warnings
 from itertools import combinations_with_replacement
 
 import cupy as cp
-#from scipy.spatial import KDTree        # FIXME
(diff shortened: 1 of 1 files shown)


kmaehashi (approved):
LGTM!

kmaehashi:
@takagi Approved. Please merge by yourself when necessary tests pass.

kmaehashi:
@ev-br We are keeping CuPy importable without scipy installation like this:
https://github.com/cupy/cupy/blob/445169dae3d1a5e5a9e6ec9adbef441a5ea4f70f/cupyx/scipy/sparse/linalg/_solve.py#L17-L22

We are also trying to keep the code runnable without scipy installation whenever possible, with some exceptions e.g., conversion of sparse arrays from GPU to CPU (which obviously requires `scipy.sparse.*_matrix` data structure) and sparse linalg functions like `SuperLU`.

PR #7219, opened 2022-12-05
`cupyx.scipy.sparse.linalg.spsolve` : allow two-dimensional right-hand sides in `A @ X = B`

ev-br:
Make `cupyx.scipy.sparse.spsolve` loop over the columns of `b` --- this matches the behavior of the scipy equivalent, which accepts 2D `B` arrays/matrices in `A @ X = B`: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.spsolve.html

This is  a pure python loop since the underlying cusolver routine IIUC only accepts vectors. 

Several things to note:
1. This does not handle the case where A is real and b is complex. This still needs an upcast or looping on the user side 
2. This patch removes `b.ravel()` --- I tried testing with non-contiguous `b` arrays, so am *pretty sure* that it's OK, but please advise about how far one has to bend backwards to assert contiguity of inputs to cusolver routines.
3. Fortran ordering: this PR explicitly casts the result to be Fortran contiguous to match what `scipy.sparse.linalg.spsolve`  returns. Please advise.
4. Testing tolerance is pretty weak currently. Not sure how to best choose the dtype-dependent tolerance of `@assert_numpy_cupy_allclose` for a test which is additionally parametrized by dtypes.

### DIFF
--- a/cupyx/scipy/sparse/linalg/_solve.py
+++ b/cupyx/scipy/sparse/linalg/_solve.py
@@ -483,7 +483,7 @@ def spsolve(A, b):
         A (cupyx.scipy.sparse.spmatrix):
             Sparse matrix with dimension ``(M, M)``.
         b (cupy.ndarray):
-            Dense vector or matrix with dimension ``(M)`` or ``(M, 1)``.
+            Dense vector or matrix with dimension ``(M)`` or ``(M, N)``.
 
     Returns:
         cupy.ndarray:
@@ -498,7 +498,7 @@ def spsolve(A, b):
     if A.shape[0] != A.shape[1]:
         raise ValueError('A must be a square matrix (A.shape: {})'.
                          format(A.shape))
-    if not (b.ndim == 1 or (b.ndim == 2 and b.shape[1] == 1)):
+    if not (b.ndim == 1 or b.ndim == 2):
         raise ValueError('Invalid b.shape (b.shape: {})'.format(b.shape))
     if A.shape[0] != b.shape[0]:
         raise ValueError('matrix dimension mismatch (A.shape: {}, b.shape: {})'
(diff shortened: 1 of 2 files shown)


emcastillo:
>>> This patch removes b.ravel() --- I tried testing with non-contiguous b arrays, so am pretty sure that it's OK, but please advise about how far one has to bend backwards to assert contiguity of inputs to cusolver routines.

Cusolver wrapper makes it contiguous, so it is safe to remove the ravel :)
https://github.com/cupy/cupy/blob/master/cupyx/cusolver.pyx#L853

>>> Fortran ordering: this PR explicitly casts the result to be Fortran contiguous to match what scipy.sparse.linalg.spsolve returns. Please advise.
In general, we allow deviations here. In some cases, performance can be hurt by ensuring strides compatibility. If such is the case, it is ok to leave it as c-contiguous. But this should be noted in a warning in the documentation.

ev-br:
Thanks Emilio! 
So 2 and 3 are OK as is, it seems; 1 can stay for a sequel; 4 is fixed in the latest commit by splitting the test into two (a low-tech solution). So this is ready from my side.

takagi (approved):
LGTM!

takagi:
Jenkins CI fails on ROCm environments as rosSOLVER does not seem to have sparse LAPACK functions yet (`csrlsvqr` here). We can skip the test like:

https://github.com/cupy/cupy/blob/6e7831ed71d2862066c4b1c493008e2f85a88f9a/tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py#L1777

ev-br:
Test skip done in https://github.com/cupy/cupy/pull/7219/commits/52cf7ad0c3e6c3596de784800f113a2ca78f30d9, cannot test it myself though.

takagi (approved):

PR #7221, opened 2022-12-06
Use naive comb() for Python 3.7

takagi:
Follows up #7218. We fixed to use `math.comb()` instead of `scipy.special.comb()` there to remove scipy dependency, however, it requires Python 3.8 or higher.


takagi:
no need to backport

PR #7222, opened 2022-12-06
Fix interpreting Sparse init arguments

takagi:
https://ci.preferred.jp/cupy.linux.cuda11x-cuda-python/117857/


takagi:
https://github.com/cupy/cupy/pull/7221 blocks this.

emcastillo (approved):

Issue #7243, opened 2022-12-19
Support NumPy 1.24


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read Release Notes.
- [x] Fix code and CI to support the new version. (https://github.com/cupy/cupy/pull/7248)
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - **Update Requirements**: For NumPy/SciPy, bump the pinned minimum/maximum version in `setup.py` (see #4773.)
- [x] Backport the above PR. (https://github.com/cupy/cupy/pull/7423)
- [x] For NumPy update, reflect the changes in `numpy.array_api` to `cupy.array_api`. -> Split to #7250
- [x] Backport the above PR. -> Split to #7250
- [x] Fix documentation.
  - Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) (note: the baseline version must be updated only in the development branch).
  - Update `docs/requirements.txt` to use the latest versions.
- [x] Backport the above PR.
- [x] Apply fixes needed to follow the specification changes. -> Split to #7250

kmaehashi:
The following tests need to be updated.

https://ci.preferred.jp/cupy.linux.cuda-head/118940/ ([full log](https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/118940/log.txt)):


```
FAILED cupy_tests/binary_tests/test_elementwise.py::TestElementwise::test_bitwise_and
FAILED cupy_tests/binary_tests/test_elementwise.py::TestElementwise::test_bitwise_or
FAILED cupy_tests/binary_tests/test_elementwise.py::TestElementwise::test_bitwise_xor
FAILED cupy_tests/binary_tests/test_elementwise.py::TestElementwise::test_invert
FAILED cupy_tests/binary_tests/test_elementwise.py::TestElementwise::test_left_shift
FAILED cupy_tests/binary_tests/test_elementwise.py::TestElementwise::test_right_shift
FAILED cupy_tests/core_tests/test_core.py::TestMinScalarType::test_numpy_scalar
FAILED cupy_tests/core_tests/test_core.py::TestMinScalarType::test_cupy_scalar
FAILED cupy_tests/core_tests/test_core.py::TestMinScalarType::test_numpy_ndarray
[...]

PR #7245, opened 2022-12-19
Update for deprecations in NumPy 1.24

kmaehashi:
Refs #7243

https://numpy.org/devdocs/release/1.24.0-notes.html

Deprecated in NumPy 1.24:
* Deprecate msort
* np.str0 and similar are now deprecated

Expired deprecations in NumPy 1.24:
* The deprecation for the aliases np.object, np.bool, np.float, np.complex, np.str, and np.int is expired (introduces NumPy 1.20). Some of these will now give a FutureWarning in addition to raising an error since they will be mapped to the NumPy scalars in the future.

### DIFF
--- a/.pfnci/windows/test.ps1
+++ b/.pfnci/windows/test.ps1
@@ -73,6 +73,7 @@ function Main {
 
     echo "Building..."
     $build_retval = 0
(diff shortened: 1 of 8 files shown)


emcastillo (approved):
lgtm

Issue #6793, opened 2022-06-20
Wrong CUB resolution order


kmaehashi:
Currently, CUB headers bundled with CuPy is preferred over the one bundled with CUDA Toolkit. This should be reversed.

https://github.com/cupy/cupy/blob/1459a98eefb0eec2d65bedff13981b87669e7d69/cupy/_environment.py#L177-L183

leofang:
I suggest #6700 as an alternative solution. However we've internally noticed some issues when doing a naive update over the bundled CUB to v1.15+ (?) It mainly has to do with the CUB reduction kernels JIT'd at runtime. cc: @anaruse

emcastillo:
Seems like jitify doesn't support some of the features in recent CUB, so using the cudatoolkit one in runtime causes errors

```
../util_device.cuh(44): warning: atomic: [jitify] File not found
../util_device.cuh(45): warning: array: [jitify] File not found
---------------------------------------------------
--- JIT compile log for /tmp/tmphtk125s5/cca62eaaf7317b7707771c787338a68b6d3c61cb.cubin.cu ---
---------------------------------------------------
cupy/complex/complex.h(94): warning: __device__ annotation is ignored on a function("complex") that is explicitly defaulted on its first declaration

cupy/complex/complex.h(101): warning: __device__ annotation is ignored on a function("complex") that is explicitly defaulted on its first declaration

util_cpp_dialect.cuh(129): warning: unrecognized #pragma

specializations/block_reduce_raking_commutative_only.cuh(98): warning: declaration does not declare anything

../util_device.cuh(154): error: identifier "cudaSetDevice" is undefined

../util_device.cuh(160): error: identifier "cudaSetDevice" is undefined

../util_device.cuh(273): error: qualified name is not allowed

../util_device.cuh(273): error: this declaration has no storage class or type specifier

../util_device.cuh(273): error: expected a ";"

../util_device.cuh(278): error: qualified name is not allowed

../util_device.cuh(278): error: this declaration has no storage class or type specifier

../util_device.cuh(278): error: expected a ";"

../util_device.cuh(284): error: "entries_" is not a nonstatic data member or base class of class "cub::PerDeviceAttributeCache"

../util_device.cuh(301): error: identifier "entries_" is undefined

../util_device.cuh(308): error: namespace "std" has no member "memory_order_acquire"

../util_device.cuh(314): error: namespace "std" has no member "memory_order_acq_rel"

[...]

PR #7259, opened 2022-12-21
Fix race condition in Jitify

kmaehashi:
This fixes a race condition in jitify that `cupy_headers` (static variable) being modified under `with nogil:`. Reported in https://github.com/cupy/cupy/issues/6793#issuecomment-1308791091 and https://github.com/cupy/cupy/issues/6793#issuecomment-1360368533.

Reproducer

```
$ export CUPY_CACHE_IN_MEMORY=1
$ ./test.py 10
```

`test.py`:

```py
import sys
import threading
import cupy


def run():
    kern = cupy.RawKernel(code='''
#include <cupy/cuda_workaround.h>
#include <cupy/cub/cub/block/block_reduce.cuh>
#include <cupy/cub/cub/block/block_load.cuh>

extern "C" __global__ void mykernel() {}''', name='mykernel', jitify=True)
    kern.compile()

threads = []
for i in range(int(sys.argv[1])):
    threads.append(threading.Thread(target=run))

for t in threads:
    t.start()

for t in threads:
    t.join()
```

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -20,7 +20,7 @@ from cupy import _util
 
 _cuda_hip_version = driver.get_build_version()
 if not runtime.is_hip and _cuda_hip_version > 0:
-    from cupy.cuda.jitify import jitify
+    from cupy.cuda import jitify
 
 
 _nvrtc_version = None
@@ -234,11 +234,7 @@ def _jitify_prep(source, options, cu_path):
     global _jitify_header_source_map_populated
     if not _jitify_header_source_map_populated:
         from cupy._core import core
-        _jitify_header_source_map = core._get_header_source_map()
-        _jitify_header_source_map_populated = True
-    else:
-        # this is already cached at the C++ level, so don't pass in anything
-        _jitify_header_source_map = None
+        jitify._add_sources(core._get_header_source_map())
 
     # jitify requires the 1st line to be the program name
     old_source = source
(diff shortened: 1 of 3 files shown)


emcastillo (approved):

cupy/cuda/compiler.py
```diff
@@ -234,11 +234,7 @@ def _jitify_prep(source, options, cu_path):
     global _jitify_header_source_map_populated
     if not _jitify_header_source_map_populated:
         from cupy._core import core
-        _jitify_header_source_map = core._get_header_source_map()
-        _jitify_header_source_map_populated = True
-    else:
-        # this is already cached at the C++ level, so don't pass in anything
-        _jitify_header_source_map = None
+        jitify._add_sources(core._get_header_source_map())
```
leofang, line 237: It seems `_jitify_header_source_map_populated` will never be set to `True` with the current change?

leofang (commented):
Thanks, @kmaehashi, that's a good catch. I missed it 😓 

Question: If this is true
> This fixes a race condition in jitify that `cupy_headers` (static variable) being modified under `with nogil:`.

shouldn't we just run call the Jitify wrappers `with gil` and be done with it?

kmaehashi:
> shouldn't we just run call the Jitify wrappers with gil and be done with it?

This also works but `load_program` is one of the heaviest parts so I think it's better to make it run in parallel.

cupy/cuda/compiler.py
kmaehashi: Fixed! :+1:

leofang:
I see. So IIUC with this PR we have a few guards against a race:
1. `_add_sources()` in `compiler.py` is protected by GIL
2. before calling `load_program()`, each thread has its local copy of the global, static dict
3. after calling `load_program()`, the global dict is updated but protected by GIL

leofang (approved):

emcastillo:
```
ERROR cupyx_tests\scipy_tests\sparse_tests\test_csr.py::TestCsrMatrix::test_reshape_0[_param_3_{dtype=complex128}] - cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered
```
CUDA 10.2 & windows

kmaehashi:
Seems that the problem is in the master branch. I've asked @takagi to investigate.

PR #7206, opened 2022-11-28
Implement `make_interp_spline(..., bc_type="periodic")`

ev-br:
This is branched off gh-7195, will rebase after it is in. Relevant commits start from 64c80c417cc490f20e1c9e0bff22bb7552e200bf

PR #7262, opened 2022-12-22
Fix CI failures related to `cupyx.scipy.interpolate`

takagi:
Follows up #7206.

PR #7264, opened 2022-12-23
Fix int dtype

takagi:
Follows up #7206.

### DIFF
--- a/cupy/typing/_generic_alias.py
+++ b/cupy/typing/_generic_alias.py
@@ -18,8 +18,10 @@ else:
 
 if sys.version_info >= (3, 9):
     from types import GenericAlias
+elif numpy.lib.NumpyVersion(numpy.__version__) >= '1.23.0':
+    from numpy._typing import _GenericAlias as GenericAlias
 elif numpy.lib.NumpyVersion(numpy.__version__) >= '1.21.0':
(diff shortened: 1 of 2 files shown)


cupy/typing/_generic_alias.py
```diff
@@ -18,6 +18,8 @@
 
 if sys.version_info >= (3, 9):
     from types import GenericAlias
+elif numpy.lib.NumpyVersion(numpy.__version__) >= '1.23.0':
+    from numpy._typing import _GenericAlias as GenericAlias
```
takagi, line 22: Seems that `_GenericAlias` is moved to `numpy._typing` since numpy 1.23. I wonder why we first see this failure here, though.

takagi: ah, I found it in cupy.win.cuda113, which is not required in master test.

takagi:
This PR only fixes the erros on windows. The liinux CI failure should be fixed at #7262.

takagi:
I've merged these commits into #7262, and will test them there at once.

carterbox:
Could we please backport this patch to versions `11.x`?

kmaehashi:
@carterbox Thanks for the feedback, will do.

PR #7272, opened 2023-01-02
Implementing odds and ends in interpolate module

ideasrule:
Implementing approximate_taylor_polynomial, pade, lagrange.  Part of #7186

### DIFF
--- a/cupyx/scipy/interpolate/__init__.py
+++ b/cupyx/scipy/interpolate/__init__.py
@@ -3,8 +3,11 @@ from cupyx.scipy.interpolate._polyint import BarycentricInterpolator  # NOQA
 from cupyx.scipy.interpolate._polyint import KroghInterpolator  # NOQA
 from cupyx.scipy.interpolate._polyint import barycentric_interpolate  # NOQA
 from cupyx.scipy.interpolate._polyint import krogh_interpolate  # NOQA
+from cupyx.scipy.interpolate._polyint import approximate_taylor_polynomial  # NOQA
 from cupyx.scipy.interpolate._interpolate import PPoly  # NOQA
+from cupyx.scipy.interpolate._interpolate import lagrange  # NOQA
 from cupyx.scipy.interpolate._cubic import CubicHermiteSpline  # NOQA
+from cupyx.scipy.interpolate._pade import pade  # NOQA
 
 # 1-D Splines
 from cupyx.scipy.interpolate._bspline import BSpline, splantider, splder  # NOQA
--- a/cupyx/scipy/interpolate/_interpolate.py
+++ b/cupyx/scipy/interpolate/_interpolate.py
@@ -1,5 +1,5 @@
-
 import cupy
+from cupy import poly1d
 from cupy._core import internal  # NOQA
 from cupy._core._scalar import get_typename  # NOQA
 from cupy_backends.cuda.api import runtime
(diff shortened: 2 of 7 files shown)


takagi:
Would you also make a performance comparison against SciPy counterparts for this PR? We'd like you to leave a benchmark script and tables on a GH comment as andfoy did at https://github.com/cupy/cupy/pull/7204#issuecomment-1352117803. Thanks!

ideasrule:
@takagi I'm glad you asked for benchmarks, because I found out that the cupy versions I wrote of the three functions are horribly, unacceptably slow compared to the CPU versions.  The reason is because for all three functions, the inputs are by necessity small.  The documentation for lagrange says it's only stable for arrays smaller than 20 elements; approximate_taylor_polynomial creates arrays that are comparable to the order of the polynomial, so never more than a dozen or so; and pade accepts the coefficients of a polynomial, which again won't ever be longer than a dozen or so.

I decided that the fastest way to implement these functions is to make them thin wrappers around their scipy counterparts.  Because of the very small sizes of the input arrays, there is no reasonable use case in which the GPU outperforms the CPU for these functions.  Do you think this is a valid implementation, or should we instead not have these functions in cupy at all and force the user to call the scipy versions themselves?

<details>
<summary>Benchmarking script</summary>

```python
import numpy as np
import cupy as cp
import cupyx.scipy.interpolate 
import scipy.interpolate

def gather_time(prof):

    
print("Benchmarking lagrange (100 runs)")
selections = [(np, scipy.interpolate.lagrange, 'scipy'), (cp, cupyx.scipy.interpolate.lagrange, 'cupy')]
for xp, func, label in selections:
    t = xp.linspace(0, 1, 20)
    x = t**2
    y = xp.sin(t)
    time = gather_time(benchmark(func, (x, y), n_repeat=100))
    print(label, time)

print("Benchmarking approximate_taylor_polynomial (100 runs)")
selections = [(np, scipy.interpolate.approximate_taylor_polynomial, 'scipy'), (cp, cupyx.scipy.interpolate.approximate_taylor_polynomial, 'cupy')]
for xp, func, label in selections:
    time = gather_time(benchmark(func, (xp.exp, 1.2, 10, 0.5), n_repeat=100))
    print(label, time)

print("Benchmarking pade (100 runs)")
selections = [(np, scipy.interpolate.pade, 'scipy'), (cp, cupyx.scipy.interpolate.pade, 'cupy')]
for xp, func, label in selections:
    time = gather_time(benchmark(func, (xp.asarray([1.0, 1.0, 1.0/2.0, 1.0/6.0, 1.0/24.0, 1.0/120.0]), 2), n_repeat=100))
    print(label, time)
```
</details>

takagi:
Thanks! I understand your analysis and it sounds making sense. This time we'd like to have CuPy version of these functions as you pushed at first (4d80ab65bbbeff9e50268da489d015a37fb2d754). While the polynomial construction would be slower on GPU, its evaluation, or `__call__`, could be faster when large data points to evaluate are supplied.

We also know your implementation could be more optimized for GPU so we'll fix it after it is merged.

ideasrule:
> Thanks! I understand your analysis and it sounds making sense. This time we'd like to have CuPy version of these functions as you pushed at first ([4d80ab6](https://github.com/cupy/cupy/commit/4d80ab65bbbeff9e50268da489d015a37fb2d754)). While the polynomial construction would be slower on GPU, its evaluation, or `__call__`, could be faster when large data points to evaluate are supplied.

The thing is that for these functions in particular, large data points should never be supplied.  "lagrange" returns an interpolating polynomial through the given datapoints, but the documentation says "this implementation is numerically unstable. Do not expect to be able to use more than about 20 points even if they are chosen optimally."  "pade" returns an approximation to a polynomial as the ratio of two polynomials, but nobody's going to have a millionth-order polynomial, and the code will break if it gets one.  approximate_taylor_polynomial constructs a polynomial approximation to a user-provided function at a given point.  The polynomial will rarely have more than ~several coefficients, and the documentation says "the algorithm used becomes numerically unstable around order 30 even under ideal circumstances."

So there is no legitimate use case in which the GPU implementation, even if heavily optimized, is likely to be faster than the CPU implementation.  I think the two choices are to use the CPU implementation, or to not implement these functions in cupy at all.

PR #7286, opened 2023-01-09
Check if the slice does not have inhomogeneous shape before converting it to array

kmaehashi:
Part of NumPy 1.24 support #7243.

NumPy 1.24:

```pycon
>>> import numpy as np
>>> np.arange(10)[[1,2,[3]]]
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.
```

NumPy 1.23:

```pycon
>>> import numpy as np
>>> np.arange(10)[[1,2,[3]]]
<stdin>:1: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

### DIFF
--- a/cupy/_core/_routines_indexing.pyx
+++ b/cupy/_core/_routines_indexing.pyx
@@ -266,6 +266,9 @@ cpdef list _prepare_slice_list(slices):
                 # keep scalar int
                 continue
 
+        if cupy.min_scalar_type(s).char == 'O':
+            raise IndexError(
+                'arrays used as indices must be of integer (or boolean) type')
         try:
             s = core.array(s, dtype=None, copy=False)
         except ValueError:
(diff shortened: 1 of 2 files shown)


takagi (approved):
LGTM!

takagi:
`test_ndarray_adv_indexing.py` passes. Other Ci failures are not related.

kmaehashi:
Tests for v11 updated in #7287.

PR #7305, opened 2023-01-14
Add `CUPY_INCLUDE_PATH` and `CUPY_LIBRARY_PATH` env vars

### LINKED ISSUE #5751: Introduce `CUPY_INSTALL_LIBRARY_PATH` and `CUPY_INSTALL_INCLUDE_PATH`
When building from source, currently we allow users to directly specify `CUTENSOR_PATH` and `CUGRAPH_PATH` but does not support other libraries.

I think it's better to have `CUPY_INSTALL_LIBRARY_PATH` and `CUPY_INSTALL_INCLUDE_PATH` so that users don't have to specify `LDFLAGS`, `CFLAGS`, etc.

kmaehashi:
This closes #5751.
(I'm going to use this one in v11 to support building NVTX1 on CUDA 12.0 / Windows)

### DIFF
--- a/install/cupy_builder/_context.py
+++ b/install/cupy_builder/_context.py
@@ -10,6 +10,13 @@ def _get_env_bool(name: str, default: bool, env: Mapping[str, str]) -> bool:
     return env[name] != '0' if name in env else default
 
 
+def _get_env_path(name: str, env: Mapping[str, str]) -> List[str]:
+    paths = env.get(name, None)
+    if paths is None:
+        return []
+    return [x for x in paths.split(os.pathsep) if len(x) != 0]
+
+
 class Context:
     def __init__(
             self, source_root: str, *,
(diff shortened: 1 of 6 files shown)


leofang:
Q: How does this change the build process for those who need to build from source?

kmaehashi:
Nothing will change if these envs are not set. By these env vars, users can just set a list of directories to discover, instead of using platform-dependent argument env vars (`LDFLAGS=-L/path/to/cudnn/lib64` on Linux, `LINK=/LIBPATH:C:\path\to\cudnn\lib` on Windows).

kmaehashi:
As a next step I'm thinking of `python cupyx/tools/install_library.py --action env` generate outputs like

```
export CUPY_INCLUDE_PATH=~/.cupy/cuda_libs/11.8/cudnn/8.7.0/include
export CUPY_LIBRARY_PATH=~/.cupy/cuda_libs/11.8/cudnn/8.7.0/lib64
```

so that contributors can use tools managed by the installer for development (and also for wheel build).

emcastillo:
Rebased and fixed conflict!

emcastillo (approved):
I wasn't aware that `Context`  could be used outside the real building context :), sorry for the mess up

Issue #7232, opened 2022-12-09
Support CUDA 12.0


takagi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read [CUDA Release Notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html).
- [x] Update CUDA driver in FlexCI images (Linux/Windows) (ask @kmaehashi).
- [x] Fix code and CI to support the new version. (**SEE LIST BELOW**)
  - [x] **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - [x] **Compute Capability Support**: Check [NVRTC Reference (--gpu-architecture option)](https://docs.nvidia.com/cuda/nvrtc/index.html#group__options) and [NVCC Reference (--generate-code)](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation-generate-code), and if needed, change the maximum value of supported compute capability defined in [cupy/cuda/compiler.py](https://github.com/cupy/cupy/blob/master/cupy/cuda/compiler.py) (`nvrtc_max_compute_capability`) and `_cc_major_map` in `cupy/fft/_callback.pyx`. Also update the list of architectures in [_compiler.py](https://github.com/cupy/cupy/blob/master/install/cupy_builder/_compiler.py) (`arch_list`).
  - [x] **FP16 Headers**: Copy FP16 header files to `cupy/_core/include/cupy/_cuda/`, then update `cupy/_core/core.pyx` to use it.
  - [x] **Library Installer**: Update `cupyx/tools/install_library.py`.
  - [x] **Wheel Package Detection**: Add the package to the [duplicate detection](https://github.com/cupy/cupy/blob/master/cupy/_environment.py).
  - [x] **Universal Wheel Package (`cupy-wheel`)**: Update the list of packages and detection logic in [universal_pkg](https://github.com/cupy/cupy/blob/master/install/universal_pkg/setup.py)
- [x] Backport the above PR.
- [x] Fix `cupy-release-tools` to support the new version.  https://github.com/cupy/cupy-release-tools/pull/322
- [x] Backport the above PR. https://github.com/cupy/cupy-release-tools/pull/325
- [x] Fix documentation.
  - Add new wheel package to the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) and `README.md`.
[...]

PR #7308, opened 2023-01-16
Remove texture reference completely

leofang:
Related to #7232 (see https://github.com/cupy/cupy/issues/7232#issuecomment-1347611986). This PR completely removes any APIs related to texture reference, which has been marked as deprecated since forever (?) ~~and is finally removed in CUDA 12+~~.


docs/source/user_guide/kernel.rst
```diff
@@ -434,7 +434,7 @@ The instruction above for using complex numbers in :class:`~cupy.RawKernel` also
 
 For CUDA kernels that need to access global symbols, such as constant memory, the :meth:`~cupy.RawModule.get_global` method can be used, see its documentation for further detail.
 
-CuPy also supports the Texture Reference API. A handle to the texture reference in a module can be retrieved by name via :meth:`~cupy.RawModule.get_texref`. Then, you need to pass it to :class:`~cupy.cuda.texture.TextureReference`, along with a resource descriptor and texture descriptor, for binding the reference to the array. (The interface of :class:`~cupy.cuda.texture.TextureReference` is meant to mimic that of :class:`~cupy.cuda.texture.TextureObject` to help users make transition to the latter, since as of CUDA Toolkit 10.1 the former is marked as deprecated.)
+Note that the deprecated API :meth:`cupy.RawModule.get_texref` has been removed since CuPy vX.X due to the removal of texture reference support from CUDA.
```
leofang, line 437: Once the release plan is decided, the version number here should be updated.

leofang:
Sorry, I had wrong understanding. The texture reference APIs from CUDA **runtime** are removed in CUDA 12.0, but those from CUDA driver (that we use) are intact. Nevertheless, I still think it's good to be slated for removal in order to be in preparation for a future CUDA change and suppress build-time warnings. Functionality-wise, texture (and surface) objects are the way to go.

leofang:
btw @kmaehashi @emcastillo this is technically a breaking change, but I seriously doubt there's any user... Just FYI for planning

emcastillo:
yeah, we are not very sure about what to do with this and wanted to discuss it with you

leofang:
That's what I figured 🙂 My justification is that there's no public project using `get_texref()`, which is the last API in the pipeline for utilizing texture references, so I assume the impact is very small:
https://github.com/search?q=cupy+get_texref+-org%3Acupy+language%3APython&type=code&l=Python

The other justification is we've documented it since the beginning that `TextureReference` is deprecated and subject to removal. Since all capabilities are also coveted by `TextureObject` and the `TextureReference` API was designed to mimic that of `TextureObject` (on purpose), I think even if there's user code broken it is at-most a one-line fix.

emcastillo (approved):
ByeBye

PR #7315, opened 2023-01-17
Partially revert device context behavior (to v9 and before)

### LINKED ISSUE #6965: Discussion: Restore (or document how to restore) the old `Device` context manager behavior (pre-v10)?
Follow-up of #5705, #6867. 

For downstream libraries depending on CuPy, having a context manager that respects any `cudaSetDevice`/`cudaGetDevice` calls inside _or outside_ CuPy is useful. However, this is not possible since v10. 

We are investigating this following snippet as a workaround:
```python
@contextlib.contextmanager
def device_ctx(new_device_id):
    """
    old_device_id = cp.cuda.runtime.getDevice()
    try:
        if old_device_id != new_device_id:
            cp.cuda.runtime.setDevice(new_device_id)
        yield
    finally:
        # We should always restore the old device at exit.
        cp.cuda.runtime.setDevice(old_device_id)
```
For comparison this is what CuPy is currently doing:
https://github.com/cupy/cupy/blob/fc04c9445f488272c9a1b78999e4e0df15a721b5/cupy/cuda/device.pyx#L180-L191
Compare it with CuPy's ctx mgr, we solely rely on the facts that
1. CUDA device context has a per-thread state, so we don't need a thread-local stack.
2. The `device_ctx` is implemented such that it is stateless (which is critical when leaving the context). Each `device_ctx` instance keeps its own `old_device_id` thanks to the `contextmanager` decorator.

This snippet has passed CuPy's device tests. I am wondering what have we missed in CuPy earlier? I assume it is not possible for CuPy to adopt this approach given that CuPy's `Device` is stateful. If so, perhaps we can at least document it?

leofang:
Close #6965.

It's a "partial" revert because the non-reentrant property (https://github.com/cupy/cupy/issues/6965#issuecomment-1383574548) had not been emphasized before AFAIR.

### DIFF
--- a/cupy/cuda/device.pxd
+++ b/cupy/cuda/device.pxd
@@ -17,6 +17,7 @@ cdef class Handle:
 cdef class Device:
     cdef:
         public int id
+        int previous_id
         list _device_stack
 
     cpdef use(self)
--- a/cupy/cuda/device.pxd
+++ b/cupy/cuda/device.pxd
@@ -18,7 +18,6 @@ cdef class Device:
     cdef:
         public int id
         int previous_id
-        list _device_stack
 
     cpdef use(self)
     cpdef synchronize(self)
--- a/cupy/cuda/device.pyx
+++ b/cupy/cuda/device.pyx
@@ -114,8 +114,14 @@ cdef class Device:
        with Device(0):
            do_something_on_device_0()
 
-    After the *with* statement gets done, the current device is reset to the
-    original one.
+    After the ``with`` statement gets done, the current device is reset to the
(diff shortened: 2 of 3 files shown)


cupy/cuda/device.pxd
```diff
@@ -17,7 +17,7 @@ cdef class Handle:
 cdef class Device:
     cdef:
         public int id
```
leofang, line 19: Just an aside: Shouldn't this be `readonly` instead of `public`?

leofang:
Superseded by https://github.com/cupy/cupy/pull/7427.

PR #7316, opened 2023-01-17
Work around a potential OOM error raised by CUB histogram 

leofang:
Thanks @ogreen for reporting and providing the reproducer.

This PR adds a workaround to avoid CUB histogram from asking for unrealistic amount of workspace (NVIDIA/cub#613). For example, the reproducer (added as a standalone test) would ask for 91.6 GB of device memory on my system. The added workaround ensures that if such extreme cases happen, we fall back to CuPy's existing histogram implementation.

Note that the workaround is made as thin as possible, with the eventual removal in mind once it's fixed upstream. In `cub.pyx` we have never checked for unrealistic workspace requests (we always assume no OOM), and so I do not intend to make too much change, as it might need a separate PR to properly handle all CUB functions 😅

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -16,6 +16,7 @@ from cupy.cuda cimport stream
 
 import cupy
 import cupy._core as _core
+from cupy.cuda import memory as _memory
 import numpy
 
 
@@ -572,3 +573,15 @@ cpdef cub_scan(_ndarray_base arr, op):
         return device_scan(arr, op)
 
     return None
+
+
+cpdef cub_histogram(_ndarray_base x, _ndarray_base y, bins):
+    """Check if the required workspace size is too much, if not then proceed
+    to compute the histogram, otherwise return None. This is a workaround for
+    NVIDIA/cub#613.
+    """
+    try:
+        out = device_histogram(x, y, bins)
+    except _memory.OutOfMemoryError:
+        return None
+    return out
--- a/cupy/_statistics/histogram.py
+++ b/cupy/_statistics/histogram.py
@@ -246,7 +246,12 @@ def histogram(x, bins=10, range=None, weights=None, density=False):
                     acc_bin_edge[-1] = cupy.nextafter(last, last + 1)
                 if runtime.is_hip:
                     y = y.astype(cupy.uint64, copy=False)
-                y = cub.device_histogram(x, y, acc_bin_edge)
+                out = cub.cub_histogram(x, y, acc_bin_edge)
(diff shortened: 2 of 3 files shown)


cupy/cuda/cub.pyx
```diff
@@ -572,3 +573,15 @@ cpdef cub_scan(_ndarray_base arr, op):
...
+
+
+cpdef cub_histogram(_ndarray_base x, _ndarray_base y, bins):
+    """Check if the required workspace size is too much, if not then proceed
+    to compute the histogram, otherwise return None. This is a workaround for
+    NVIDIA/cub#613.
+    """
+    try:
+        out = device_histogram(x, y, bins)
+    except _memory.OutOfMemoryError:
```
emcastillo, line 585: Would it be possible to set a threshold value based on `bins` rather than just make this error?

leofang: It's possible, but how much threshold is needed?

leofang: (but it'd only reduce the chance of hitting OOM, not eliminate it)

ogreen: The threshold of `bins` will change based on the device's memory\capability making such a solution device-specific.

ogreen (approved):

ogreen:
@leofang @emcastillo , I approved the fix.
Thanks for taking care of this.

cupy/cuda/cub.pyx
kmaehashi: I'm wondering whether we should disable CUB in `cupy.histogram` until CUB issue gets fixed. It is an unpredictable behavior for users to use so much memory even if it fits in the memory.

leofang: This is a corner case in which users have huge number of bins (I have no idea why @ogreen did that TBH 😂), but for other use cases, CUB is still usable and preferred I'd say.

ogreen: I think that so long as the application doesn't crash, then based on what I have seen CUB's implementation is very efficient. However, it would be nice to have a fallback in case it fails.
FWIW, the application that I was using histogram was in the context of graphs. I was attempting to get the adjacency list size for each vertex in the graph. While this can be accomplished with atomic operations, it would seem redundant to implement it as it is exactly what a histogram does.

kmaehashi (approved):
LGTM!

leofang:
Thanks, everyone! btw should we backport this?

kmaehashi:
Sounds good :)

PR #7322, opened 2023-01-17
Small fixes and refactor of casting related things

seberg:
This is mainly from me trying to understand the machinery better.  I had larger changes, which continuously shrank again, thus i am not sure that `_raise_if_invalid_cast` is worthwhile (and I am very happy to drop it!  Or most of the rest for that matter.).

### DIFF
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -1,5 +1,6 @@
 cimport cython  # NOQA
 import numpy
+import warnings
 
 from cupy_backends.cuda.api cimport runtime
 
@@ -89,8 +90,11 @@ cpdef int to_cuda_dtype(dtype, bint is_half_allowed=False) except -1:
 cdef _numpy_can_cast = numpy.can_cast
 
 
-cpdef _raise_if_invalid_cast(from_dt, to_dt, str casting, argname="array data"):
-    """Raise an error if a cast is not valid.
+cpdef _raise_if_invalid_cast(
+    from_dt, to_dt, str casting, argname="array data"
+):
+    """Raise an error if a cast is not valid.  Also checks whether the cast
+    goes from complex to real and warns if it does.
 
     The error raised can be customized by giving `obj`.  May pass a (lambda)
     function to avoid string construction on success.
@@ -100,7 +104,14 @@ cpdef _raise_if_invalid_cast(from_dt, to_dt, str casting, argname="array data"):
     if from_dt is to_dt:
         return
 
+    to_dt = get_dtype(to_dt)  # may still be a type not a dtype instance
     if _numpy_can_cast(from_dt, to_dt, casting):
+        if casting == "unsafe" and from_dt.kind == "c" and to_dt.kind in "iuf":
+            # Complex warning, we are dropping the imagine part:
+            warnings.warn(
+                'Casting complex values to real discards the imaginary part',
+                numpy.ComplexWarning)
+
         return
 
     # Casting is not possible, raise the error
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -47,8 +47,6 @@ _init_dtype_dict()
 
 @cython.profile(False)
 cpdef get_dtype(t):
-    if type(t) is _dtype:  # Exact type check
-        return t
     ret = _dtype_dict.get(t, None)
     if ret is None:
         return _dtype(t)
@@ -57,8 +55,6 @@ cpdef get_dtype(t):
 
 @cython.profile(False)
 cpdef tuple get_dtype_with_itemsize(t):
-    if type(t) is _dtype:  # Exact type check
-        return t, t.itemsize
     ret = _dtype_dict.get(t, None)
     if ret is None:
         t = _dtype(t)
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -105,6 +105,10 @@ cpdef _raise_if_invalid_cast(
         return
 
     to_dt = get_dtype(to_dt)  # may still be a type not a dtype instance
+
+    if casting == "same_kind" and from_dt.kind == to_dt.kind:
+        # same-kind is the most common casting used and for NumPy dtypes.
+        return
     if _numpy_can_cast(from_dt, to_dt, casting):
         if casting == "unsafe" and from_dt.kind == "c" and to_dt.kind in "iuf":
             # Complex warning, we are dropping the imagine part:
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -88,3 +88,28 @@ cpdef int to_cuda_dtype(dtype, bint is_half_allowed=False) except -1:
         return runtime.CUDA_C_16F
     else:
         raise TypeError('dtype is not supported: {}'.format(dtype))
+
+
+cdef _numpy_can_cast = numpy.can_cast
+
+
+cpdef _raise_if_invalid_cast(from_dt, to_dt, str casting, argname="array data"):
+    """Raise an error if a cast is not valid.
+
+    The error raised can be customized by giving `obj`.  May pass a (lambda)
+    function to avoid string construction on success.
+    This function exists mainly to build a similar error everywhere.
+
+    """
+    if from_dt is to_dt:
+        return
+
+    if _numpy_can_cast(from_dt, to_dt, casting):
+        return
+
+    # Casting is not possible, raise the error
+    if not isinstance(argname, str):
+        argname = argname()
+    raise TypeError(
+        f'Cannot cast {argname} from {from_dt!r} to {to_dt!r} '
+        f'according to the rule \'{casting}\'')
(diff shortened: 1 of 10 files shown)


cupy/_core/_dtype.pyx
```diff
@@ -57,8 +55,6 @@ cpdef get_dtype(t):
 
 @cython.profile(False)
 cpdef tuple get_dtype_with_itemsize(t):
-    if type(t) is _dtype:  # Exact type check
-        return t, t.itemsize
```
seberg, line 61: The exact check never works, `type(t)` is always a subclass.  I doubt this function is used much with instances though, so just deleted it.

cupy/_core/_kernel.pyx
```diff
@@ -953,6 +932,10 @@ cdef str fix_cast_expr(src_type, dst_type, str expr):
         if dst_kind == 'b':
             return f'({expr}) != {_scalar.get_typename(src_type)}()'
         else:  # dst_kind in 'iuf' (int, uint, float)
+            # give ComplexWarning, as we are explictly using `.real()`
+            warnings.warn(
+                'Casting complex values to real discards the imaginary part',
+                numpy.ComplexWarning)
```
seberg, line 938: I am not sure I like this placement... But, since both the warning and the `.real()` fix are very closely related, it seemed sensible to merge the two.

tests/cupy_tests/manipulation_tests/test_join.py
```diff
@@ -238,12 +239,13 @@ def test_concatenate_dtype_invalid_out(self):
         'unsafe',
     ])
     @testing.for_all_dtypes_combination(names=['dtype1', 'dtype2'])
-    @testing.numpy_cupy_array_equal(accept_error=TypeError)
+    @testing.numpy_cupy_array_equal(
+            accept_error=(TypeError, numpy.ComplexWarning))
```
seberg, line 243: Not quite sure that raising is the right path?  But seemed like a decent way of testing the complex warning being raised.

cupy/_core/_gufuncs.py
```diff
@@ -191,15 +192,15 @@ def _process_signatures(self, signatures):
             ops.append(_OpsRegister._Op(ins, outs, op))
         return ops
 
-    def _determine_from_args(self, args, casting):
+    def _determine_from_args(self, args):
```
seberg, line 195: Forgot to comment: Casting is not relevant if we are looking for the right loop (promotion), safe-casting must be used for things to make sense.

As far as I can tell, this code is unused though.  GUfuncs just wrap the "default_func" always currently.

tests/cupy_tests/manipulation_tests/test_join.py
```diff
@@ -238,12 +239,13 @@ def test_concatenate_dtype_invalid_out(self):
...
-    @testing.numpy_cupy_array_equal(accept_error=TypeError)
+    @testing.numpy_cupy_array_equal(
+        accept_error=(TypeError, numpy.ComplexWarning))
     def test_concatenate_casting(self, xp, dtype1, dtype2, casting):
         a = testing.shaped_arange((3, 4), xp, dtype1)
         b = testing.shaped_arange((3, 4), xp, dtype1)
-        # may raise TypeError
-        return xp.concatenate((a, b), dtype=dtype2)
+        # may raise TypeError or ComplexWarning
+        return xp.concatenate((a, b), dtype=dtype2, casting=casting)
```
seberg, line 248: OK, seems moving the warning to the re-used function was the right thing.  Just confirmed that the test fails with the fix of including `casting=casting` without the other changes.

cupy/_core/_dtype.pyx
asi1024: This fast path is introduced for performance but has not worked since NumPy 1.20. I agree to delete it!

cupy/_core/_kernel.pyx
```diff
@@ -611,20 +611,6 @@ cdef list _broadcast(list args, tuple params, bint use_size, shape_t& shape):
 cdef _numpy_can_cast = numpy.can_cast
 
 
-cdef bint _can_cast(d1, d2, casting):
-    # most ufunc passes `same_kind`
-    if casting == 'same_kind' and get_dtype(d1).kind == d2.kind:
-        return True
```
asi1024, line 617: Could you add this fast path in `_raise_if_invalid_cast` for performance? This would avoid calling a NumPy function from Cython layer.

seberg: Re-added, it may indeed be worthwhile (I am not quite sure, but presumably it was there for a reason)...
The check could be refined, but probably not for here.  (Also, maybe I could get `can_cast` or so in NumPy to be so fast that this stops mattering, but it isn't there yet probably)

cupy/_core/_gufuncs.py
```diff
@@ -191,15 +192,15 @@ def _process_signatures(self, signatures):
...
         n = len(args)
         in_types = tuple(arg.dtype for arg in args)
         for op in self._ops:
             op_types = op.in_types
             for i in range(n):
                 it = in_types[i]
                 ot = op_types[i]
-                if not numpy.can_cast(it, ot, casting=casting):
+                if not numpy.can_cast(it, ot):
```
asi1024, line 203: CI failure of TestGUFuncSignatures seems to be related to this fix. This code is certainly unused from matmul, but we want to keep the support of `signatures` option.

cupy/_core/fusion.pyx
```diff
@@ -611,17 +611,11 @@ class _FusionHistory(object):
...
-                    else:
-                        raise TypeError(
-                            'output (typecode \'{}\') could not be coerced '
-                            'to provided output parameter (typecode \'{}\') '
-                            'according to the casting rule '
-                            '"same_kind"'.format(
-                                out_dtypes[i].char, out_vars[i].dtype.char))
+
+                    _raise_if_invalid_cast(
+                        out_dtypes[i], out_vars[i], 'same_kind',
```
asi1024, line 616: `out_vars[i]` is not an actual `cupy.ndarray` object, but another array object with dummy values for tracing function in `cupy.fuse`.
```suggestion
                        out_dtypes[i], out_vars[i].dtype, 'same_kind',
```

cupy/_core/fusion.pyx
```diff
@@ -611,17 +611,11 @@ class _FusionHistory(object):
                         out_var = make_fusion_var(out_var, ndim)
                         out_vars.append(out_var)
                         ret.append(out_var)
-                    elif numpy.can_cast(out_dtypes[i], out_vars[i].dtype,
-                                        'same_kind'):
-                        out_var = out_vars[i]
```
asi1024, line 616: This line is missing. How about fixing it as follows?
```py
                    if i >= len(out_vars):
                        out_var = elf._fresh_local(out_dtypes[i])
                        out_var = make_fusion_var(out_var, ndim)
                        out_vars.append(out_var)
                        ret.append(out_var)
                    else:
                        _raise_if_invalid_cast(
                            out_dtypes[i], out_vars[i].dtype, 'same_kind',
                            lambda: f'output {i}')
                        out_var = out_vars[I]
```

cupy/_core/_gufuncs.py
seberg: Arg, sorry for not realizing... I guess I can just undo the change.  There may also be a point in using `casting="safe" if casting in ["unsafe", "same_kind"] else casting` to allow fallback to default when `casting="no"` is given.  In principle, I think casting shouldn't matter, but with a fallback to `default`, I guess stricter casting does matter.

(not sure it is worth trying to figure out compared to just undoing...)

seberg: Sorry for not doing it earlier.  I just pulled it out for now.  I think the above is more correct (currently `casting="unsafe"` would just pick the first loop, which is certainly correct).

But, there was always too much in here, give me a nudge and I will make sure to follow up.

cupy/_core/_fusion_trace.pyx
```diff
@@ -369,17 +371,13 @@ class TraceImpl:
...
-                    'according to the casting rule '
-                    '"same_kind"'.format(
-                        out_dtypes[i].char, out_params[i].dtype.char))
+
+            _raise_if_invalid_cast(
+                out_dtypes[i], out_params[i].dtype, 'same_kind',
+                'output operand')
+
+            out_pvar = out_params[i]
+            ret.append(out_pvar)
```
asi1024, line 380: This line is duplicated with L366.

cupy/_core/fusion.pyx
```diff
@@ -611,18 +611,14 @@ class _FusionHistory(object):
                         out_var = make_fusion_var(out_var, ndim)
                         out_vars.append(out_var)
                         ret.append(out_var)
```
asi1024, line 613: This line should be moved after L621.

cupy/_core/_fusion_trace.pyx
asi1024: I fixed this part as the release date is scheduled soon.

asi1024 (approved):

asi1024:
@seberg Thank you for your pull request! LGTM!

seberg:
Thanks, sorry, had forgotten about this a bit, but hopefully not the last PR!

PR #7338, opened 2023-01-23
Avoid int8 overflow warning in `TestRoundHalfway`

asi1024:
Part of https://github.com/cupy/cupy/issues/7243

```
E   AssertionError: Both cupy and numpy raise exceptions
E
E   cupy
E   Traceback (most recent call last):
E     File ".../cupy/testing/_loops.py", line 45, in _call_func
E       result = impl(*args, **kw)
E     File ".../tests/cupy_tests/core_tests/test_ndarray_math.py", line 72, in test_round_halfway_int
E       a *= xp.array(scale, dtype=dtype)
E     File ".../cupy/_creation/from_data.py", line 46, in array
E       return _core.array(obj, dtype, copy, order, subok, ndmin)
E     File "cupy/_core/core.pyx", line 2390, in cupy._core.core.array
E       cpdef _ndarray_base array(obj, dtype=None, bint copy=True, order='K',
E     File "cupy/_core/core.pyx", line 2414, in cupy._core.core.array
E       return _array_default(obj, dtype, order, ndmin)
E     File "cupy/_core/core.pyx", line 2539, in cupy._core.core._array_default
E       a_cpu = numpy.array(obj, dtype=dtype, copy=False, order=order,
E   DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of 1000 to int8 will fail in the future.
E   For the old behavior, usually:
E       np.array(value).astype(dtype)`
E   will give the desired result (the cast overflows).
E
E
E   numpy
E   Traceback (most recent call last):
E     File ".../cupy/testing/_loops.py", line 45, in _call_func
E       result = impl(*args, **kw)
E     File ".../cupy1/tests/cupy_tests/core_tests/test_ndarray_math.py", line 72, in test_round_halfway_int
E       a *= xp.array(scale, dtype=dtype)
E   DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of 1000 to int8 will fail in the future.
E   For the old behavior, usually:
E       np.array(value).astype(dtype)`
E   will give the desired result (the cast overflows).
```

PR #7340, opened 2023-01-23
Fix int8 overflow in vectorize tests

asi1024:
Rel. https://github.com/cupy/cupy/pull/7338

### DIFF
--- a/tests/cupy_tests/functional_tests/test_vectorize.py
+++ b/tests/cupy_tests/functional_tests/test_vectorize.py
@@ -5,7 +5,6 @@ import pytest
 
 import cupy
 from cupy import testing
-from cupy.cuda import runtime
 
 
 class TestVectorizeOps(unittest.TestCase):
@@ -35,13 +34,13 @@ class TestVectorizeOps(unittest.TestCase):
 
         return self._run(my_add, xp, [dtype1, dtype2])
 
-    @testing.for_all_dtypes_combination(names=('dtype1', 'dtype2'))
-    @testing.numpy_cupy_array_equal(accept_error=TypeError)
-    def test_vectorize_sub(self, xp, dtype1, dtype2):
+    @testing.for_dtypes('bhilqefdFD')
+    @testing.numpy_cupy_array_equal()
+    def test_vectorize_sub(self, xp, dtype):
         def my_sub(x, y):
             return x - y
 
-        return self._run(my_sub, xp, [dtype1, dtype2])
+        return self._run(my_sub, xp, [dtype, dtype])
 
     @testing.for_all_dtypes_combination(names=('dtype1', 'dtype2'))
     @testing.numpy_cupy_allclose(rtol=1e-6)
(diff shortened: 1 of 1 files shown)


tests/cupy_tests/functional_tests/test_vectorize.py
```diff
@@ -85,7 +84,8 @@ def run_div(self, func, xp, dtypes):
         return f(x1, x2)
 
     @testing.for_all_dtypes_combination(names=('dtype1', 'dtype2'))
-    @testing.numpy_cupy_allclose(rtol=1e-6)
+    # TODO(asi1024): Fix return type.
+    @testing.numpy_cupy_allclose(rtol=1e-3, type_check=False)
```
asi1024, line 88:
```py
>>> numpy.vectorize(lambda x, y: x / y)(numpy.array([1], dtype=numpy.uint16), numpy.array([1], dtype=numpy.float32)).dtype
dtype('float64')
```

asi1024: @seberg Question: I found that the dtype of the result of true_divide of int16 scalar and float32 scalar in numpy is different between NumPy 1.24 and earlier. Is this an intentional change or a bug?

NumPy 1.23.5
```py
>>> numpy.__version__
'1.23.5'
>>> (numpy.int16(1) + numpy.float32(1)).dtype
dtype('float32')
>>> (numpy.int16(1) / numpy.float32(1)).dtype
dtype('float32')
```

NumPy 1.24.1
```py
>>> numpy.__version__
'1.24.1'
>>> (numpy.int16(1) + numpy.float32(1)).dtype
dtype('float32')
>>> (numpy.int16(1) / numpy.float32(1)).dtype
dtype('float64')
```

seberg: Arrg, no that is a subtle bug in my new scalar logic :(.  The `int / int -> float64`, so it tries converting float32 to that and succeeds, but since the other is a float64, it must not do that.

jakirkham: xref: https://github.com/numpy/numpy/pull/23079

cc @leofang (for awareness)

kmaehashi: Thanks Sebastian for fixing this quickly!

@asi1024 Let's mark the test skipif numpy==1.24.0 or 1.24.1.

tests/cupy_tests/functional_tests/test_vectorize.py
```diff
@@ -35,13 +34,13 @@ def my_add(x, y):
 
         return self._run(my_add, xp, [dtype1, dtype2])
 
-    @testing.for_all_dtypes_combination(names=('dtype1', 'dtype2'))
-    @testing.numpy_cupy_array_equal(accept_error=TypeError)
-    def test_vectorize_sub(self, xp, dtype1, dtype2):
+    @testing.for_dtypes('bhilqefdFD')
```
kmaehashi, line 37: Talked with @asi1024 offline, this shrinks the test axis (combination of different dtypes) but it is acceptable as it is tested in `test_vectorize_add`.

tests/cupy_tests/functional_tests/test_vectorize.py
```diff
@@ -624,20 +622,17 @@ def my_func(x1, x2):
...
         def my_func(x1, x2, x3):
             y = x1 + x2 * x3 ** x1
             x2 = y + x3 * x1
             return x1 + x2 + x3
 
         f = xp.vectorize(my_func)
-        x1 = testing.shaped_random((20, 30), xp, dtype, seed=1, scale=5)
-        x2 = testing.shaped_random((20, 30), xp, dtype, seed=2)
-        x3 = testing.shaped_random((20, 30), xp, dtype, seed=3)
+        x1 = testing.shaped_random((20, 30), xp, xp.float32, seed=1, scale=5)
```
kmaehashi, line 633: Question: why is this change needed?

kmaehashi (dismissed):
LGTM except for one question

tests/cupy_tests/functional_tests/test_vectorize.py
asi1024: Because this test overflows on int8 inputs.

kmaehashi: I see, then how about shrinking the data size to, say, `(12, 16)`? As `TestVectorize` is a collection of realistic use-case tests, it makes sense to me to test dtype combination here.

asi1024: Fixed in a1c8eac.

kmaehashi: Thanks! (I just noticed I misread `shaped_random` as `shaped_arange` 😅 the fix looks good anyways)

kmaehashi (approved):

PR #7345, opened 2023-01-24
Fix kernel cache not working in CUDA 12.0

kmaehashi:
We noticed that kernel caching (`~/.cupy/kernel_cache`) is not working on CUDA 12.0. This is because PTX generated by CUDA 12.0 NVRTC includes identifiers randomly generated for every invocation.

CUDA 11.6 NVRTC:
```ptx
//
// Generated by NVIDIA NVVM Compiler
//
// Compiler Build ID: CL-30794723
// Cuda compilation tools, release 11.6, V11.6.55
// Based on NVVM 7.0.1
//

.version 7.6
.target sm_61
.address_size 64
```

CUDA 12.0 NVRTC:
```ptx
//
// Generated by NVIDIA NVVM Compiler
//
// Compiler Build ID: CL-31968024
// Cuda compilation tools, release 12.0, V12.0.76
// Based on NVVM 7.0.1
//
.version 8.0
.target sm_90
.address_size 64
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9all_hostsE = 1;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_35_bitE = 2;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_37_bitE = 4;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_50_bitE = 8;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_52_bitE = 16;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_53_bitE = 32;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_60_bitE = 64;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_61_bitE = 128;
.global .align 8 .u64 _ZN53_INTERNAL_00000000_15_default_program_26edde23_8009862nv6target6detail9sm_62_bitE = 256;
[...]

PR #7347, opened 2023-01-25
Add test for kernel caching

kmaehashi:
Ensures kernel cache hit and preprocess output consistency across invocations within the process.


kmaehashi:
I noticed that these tests cannot catch #7345, as PTX generation seems consistent within the same process. With that said I still think these tests make some sense to detect future cache-related errors.

takagi (approved):
LGTM!

takagi:
Seems that we need some fix on ROCm.

kmaehashi:

PR #7367, opened 2023-02-02
Support SciPy 1.10

takagi:


tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
```diff
@@ -213,10 +213,12 @@ def test_correlate2d(self, xp, scp, dtype):
 class TestConvolve2DEdgeCase:
 
     @testing.numpy_cupy_allclose(atol=1e-5, rtol=1e-5, scipy_name='scp')
+    @testing.with_requires('scipy>=1.10')
     def test_convolve2d_1(self, xp, scp):
         # see cupy/cupy#5989
         from scipy import misc
-        ascent = misc.ascent()
+        with pytest.warns(DeprecationWarning):
+            ascent = misc.ascent()
```
asi1024, line 221: `scipy.misc` will be removed in SciPy version 1.12.0. How about using `scipy.datasets.ascent` instead?

```suggestion
        if np.lib.NumpyVersion(scipy.__version__) >= '1.10.0':
            ascent = scipy.datasets.ascent()
        else:
            ascent = scipy.misc.ascent()
```

takagi: Thanks!

takagi:
```
E           ImportError: Missing optional dependency 'pooch' required for scipy.datasets module. Please use pip or conda to install 'pooch'.
```

takagi:
@asi1024 I think we're okay to have `pooch` in the CI to use `scipy.datasets`. 
https://docs.scipy.org/doc/scipy/reference/datasets.html#how-dataset-retrieval-and-storage-works

The data of `scipy.datasets.ascent()` seems to be 262KB binary indicated here:
```
>>> scipy.datasets.ascent()
Downloading file 'ascent.dat' from 'https://raw.githubusercontent.com/scipy/dataset-ascent/main/ascent.dat' to '/home/mtakagi/.cache/scipy-data'.
array([[ 83,  83,  83, ..., 117, 117, 117],
       [ 82,  82,  83, ..., 117, 117, 117],
       [ 80,  81,  83, ..., 117, 117, 117],
       ...,
       [178, 178, 178, ...,  57,  59,  57],
       [178, 178, 178, ...,  56,  57,  57],
       [178, 178, 178, ...,  57,  57,  58]], dtype=uint8)
```

Any idea?

kmaehashi:
IMO we can just use random numbers instead of real image data for convolution tests.

takagi:
It may work, but what I want to check here is that the testcase is named EdgeCase, so I'm going to see what the data means.

takagi:
Ah, just a gray scale picture
https://docs.scipy.org/doc/scipy/reference/generated/scipy.datasets.ascent.html

takagi:
The build-cuda pre-review test failure is due to cuquantum intersphinx inventory error.

kmaehashi:
It seems some tests need to be skipped in ROCm as rocSOLVER does not implement several methods.

kmaehashi (approved):

kmaehashi:
`cupyx_tests/scipy_tests/signal_tests/test_signaltools.py` seems failing:

https://ci.preferred.jp/cupy.linux.cuda110/131069/

```
E   cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered
```

PR #7387, opened 2023-02-10
Avoid int8 overflow in core test

asi1024:
Rel. https://github.com/cupy/cupy/pull/7338


emcastillo (approved):

emcastillo:
These tests are not in v11 so no backport needed

PR #7399, opened 2023-02-13
Change implementation of fftshift and ifftshift

Nordicus:
This changes the fftshifts to generate the amount of shift for each shifted axis and then execute the shift with a single call to cupy roll. For 1D shifts (i.e., where there is a single axis being shifted, regardless of the dimension of the array) there is no effective change, but for shifts along multiple axes, this decreases the amount of working memory required. Specifically, for an array of size N, and shifting along k axes, this requires only O(N) memory instead of O(kN) memory as before.

For FFTs that take most of the GPU's memory, this can mean the difference between completing the computation and an out of memory error or having to chunk the computation.

### DIFF
--- a/cupy/fft/_fft.py
+++ b/cupy/fft/_fft.py
@@ -1079,9 +1079,7 @@ def fftshift(x, axes=None):
         axes = list(range(x.ndim))
     elif isinstance(axes, np.compat.integer_types):
         axes = (axes,)
-    for axis in axes:
-        x = cupy.roll(x, x.shape[axis] // 2, axis)
-    return x
+    return cupy.roll(x, [x.shape[axis] // 2 for axis in axes], axes)
 
 
 def ifftshift(x, axes=None):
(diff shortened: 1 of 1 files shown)


cupy/fft/_fft.py
```diff
@@ -1102,6 +1100,4 @@ def ifftshift(x, axes=None):
         axes = list(range(x.ndim))
     elif isinstance(axes, np.compat.integer_types):
         axes = (axes,)
-    for axis in axes:
-        x = cupy.roll(x, -(x.shape[axis] // 2), axis)
-    return x
+    return cupy.roll(x, [-x.shape[axis] // 2 for axis in axes], axes)
```
takagi, line 1103: Thanks for your contribution! Seems that our CI fails in ifftshift and we'd need these parentheses so we have appropriate rounding:

```suggestion
    return cupy.roll(x, [-(x.shape[axis] // 2) for axis in axes], axes)
```

Nordicus: Thanks, you're quite right. I had only properly tested this with fftshift, since I always use even sized (and in fact power of two sized) arrays in my own code. I missed the difference that the parentheses would make.

takagi (approved):
LGTM!

PR #7410, opened 2023-02-17
Add cupyx.scipy.ndimage.value_indices

grlee77:
This MR implements the ndimage function [value_indices](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.value_indices.html#scipy-ndimage-value-indices) (recently introduced in SciPy 1.10).

This function returns a dictionary where the **keys** are the unique integer values in an image and the **values** are their corresponding coordinate locations. The implementation here does not match how it is done in the SciPy C code, but gives an equivalent result. The acceleration starts to diminish as more and more labels are present ranging from ~50x for 4 labels down to ~12x for 16k labels on my system (see benchmark below).

I added an optional `adaptive_index_dtype` keyword-only argument not present in SciPy that can reduce memory usage by using integers smaller than int64 for the indices. Without it memory used for the output is ``N = image.ndim * 8 / image.dtype.itemsize`` times larger than the size of the input image. 

When `adaptive_index_dtype` is True, if the array shape fits within int8, int16 or int32 limits these smaller dtypes will be used instead leading to 2x-8x lower memory requirement for the output.

### Benchmark results 

For an image of shape (4096, 4096) with default `adaptive_index_dtype=False`
(**CPU:** Intel i9-10980XE, **GPU:** RTX A6000)

|# labels |  SciPy   | CuPy | CuPy (adaptive=True) |
|---------|----------|--------|--------|
|4        |  212 ms  | 4.55 ms| 4.14 ms |
|16       |  218 ms  | 4.19 ms| 3.87 ms |
|64       |  319 ms  | 4.39 ms| 3.86 ms |
|256      |  365 ms  | 5.05 ms| 4.64 ms |
|1024     |  467 ms  | 9.50 ms| 8.99 ms |
|4096     |  547 ms  | 22.7 ms| 22.2 ms |
|16384    |  920 ms  | 74.5 ms| 74.4 ms |

"# labels" indicates the number of unique integer values present in the image.
Cases with <= 256 unique values used an uint8 image, while the others used uint16.

There is a slight performance benefit to `adaptive_index_dtype=True`, but it is mainly of benefit in reducing memory use.


grlee77:
I guess #7367 should be merged first otherwise the new tests will just be skipped if there is no SciPy 1.10 case on CI.

grlee77:
Hi @leofang, I see this in the failed CI run:
```
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://docs.nvidia.com/cuda/cuquantum/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 403 Client Error: Forbidden for url: https://docs.nvidia.com/cuda/cuquantum/objects.inv
```
not sure if it is network issue or if the URL is outdated?

leofang:
Sorry I was out and missed the ping, Greg. It's fixed earlier this week. Thanks for letting me know anyway 🙏

grlee77:
updated, please try running the tests again @emcastillo

emcastillo (approved):

emcastillo:
Awesome as always! Thanks!!

PR #7412, opened 2023-02-17
ROCm5.3.0+ rocPrim C++14 extension requirement.

AdrianAbeyta:
ROCm5.3.0 and above use return types are that require a minimum of C++14 extension. Fix is also backwards compatible with  ROCm4.3.0.

cc: @amathews-amd @pnunna93


cupy/_core/_cub_reduction.pyx
```diff
@@ -28,9 +28,13 @@ cdef function.Function _create_cub_reduction_function(
         _kernel._TypeMap type_map, preamble, options):
     # A (incomplete) list of internal variables:
     # _J            : the index of an element in the array
-
-    # static_assert needs at least C++11 in NVRTC
-    options += ('--std=c++11',)
+
```
takagi, line 31:
```suggestion

```

AdrianAbeyta: @takagi Seems like cuda-build pre-review may have some issues? See:https://github.com/cupy/cupy/actions/runs/4220966825/jobs/7327741747#step:7:119

takagi:
The build-cuda pre-review test failure is due to cuquantum intersphinx inventory error. See #7417

cupy/_core/_cub_reduction.pyx
takagi: yeah, it's going to have a patch for now at #7417.

takagi (approved):

takagi:
Thanks, @AdrianAbeyta! I manually confirmed test_cub_reduction.py passes with this fix on ROCm 5.3+. I'm going to merge after the repo is unfrozen for the release.
(Our ROCm 5.3 CI dies on its way, so we configure it to ignore failures and just report pass)

AdrianAbeyta:
> Thanks, @AdrianAbeyta! I manually confirmed test_cub_reduction.py passes with this fix on ROCm 5.3+. I'm going to merge after the repo is unfrozen for the release. (Our ROCm 5.3 CI dies on its way, so we configure it to ignore failures and just report pass)

Thanks! @takagi

kmaehashi:
I guess it's better to backport this one? 😄

PR #7419, opened 2023-02-21
DOC: cupyx.interpolate: document limitations on ROCm

ev-br:
Splines use banded linear algebra, which we emulate with sparse solvers, and those are not available on ROCm, so document the limitation.


takagi (approved):
LGTM!

takagi:
Thanks, @ev-br!

PR #7405, opened 2023-02-15
Drop support for Python 3.7, NumPy 1.20, and SciPy 1.6

kmaehashi:
* Drop support for Python 3.7, closes #7293. Python version in ROCm docker image bumped to 3.8.
* Drop support for NumPy 1.20, closes #7249.
* Drop support for SciPy 1.6, as it requires NumPy 1.16~1.20.

PR #7426, opened 2023-02-27
Drop support for Python 3.7, NumPy 1.20, and SciPy 1.6 on document and setup.py

takagi:
#7405 has some CI issues, so we update the documentation and setup.py first for the next release.


asi1024 (approved):

kmaehashi:
`docker/rocm/Dockerfile` needs to be updated as well :bow:

kmaehashi:
Let's try to merge #7405 as discussed :pray:

Issue #5897, opened 2021-10-14
Unexpected stride when performing cupy.asarray on numpy.ndarray with leading dim 1


keckj:
### Description

When a cupy.ndarray is obtained with cupy.asarray from a numpy.ndarray that has leading ones in its shape, the resulting array exposes wrong strides. Numpy gives expected results, see example to reproduce. This could be related to issue #913. 

### To Reproduce

```py
import numpy as np
import cupy as cp

a = np.arange(4).reshape(-1, 4)
b = cp.arange(4).reshape(-1, 4)

print(a.strides)  # ok, prints (32, 8)
print(b.strides)  # ok, prints (32, 8)

print(np.asarray(a).strides) # ok, prints (32, 8)
print(cp.asarray(b).strides) # ok, prints (32, 8)

print(cp.asarray(a).strides) # ko, prints (8, 8)
print(cp.asnumpy(b).strides) # ok, prints (32, 8)

print(cp.asarray(a).reshape(-1,4).strides) # ko, prints (8, 8)
print(cp.asarray(a).reshape(4).reshape(-1,4).strides) # ok, prints (32, 8)
```

### Installation

Source (`pip install cupy`)

### Environments
```
OS                           : Linux-5.4.0-80-generic-x86_64-with-glibc2.27
Python Version               : 3.8.12
CuPy Version                 : 9.5.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 1.21.2
SciPy Version                : 1.7.1
Cython Build Version         : 0.29.23
Cython Runtime Version       : 0.29.23
CUDA Root                    : /usr/local/cuda
nvcc PATH                    : /usr/local/cuda/bin/nvcc
CUDA Build Version           : 11040
CUDA Driver Version          : 11040
CUDA Runtime Version         : 11040
cuBLAS Version               : (available)
cuFFT Version                : 10502
cuRAND Version               : 10205
cuSOLVER Version             : (11, 2, 0)
cuSPARSE Version             : (available)
NVRTC Version                : (11, 4)
Thrust Version               : 101201
CUB Build Version            : 101201
Jitify Build Version         : <unknown>
cuDNN Build Version          : None
cuDNN Version                : None
NCCL Build Version           : None
NCCL Runtime Version         : None
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA GeForce RTX 2080 Ti
Device 0 Compute Capability  : 75
[...]

PR #7438, opened 2023-02-28
BUG: Fix new strides when array is both C and F-contiguous

seberg:
In this case, NumPy will prefer the C contiguous strides, not the F contiguous ones.

I will note that I do not think users should reuly on strides if they rely on contiguity, and the old strides could be considered just as valid as the new one.

Just as NumPy the strides of an F- and C-contiguous array created as an F-contiguous one will have strides mutate on copy after this. Currently, this happens for C contiguous ones.

Closes gh-5897

### DIFF
--- a/cupy/_core/internal.pyx
+++ b/cupy/_core/internal.pyx
@@ -498,15 +498,15 @@ cpdef int _update_order_char(
         bint is_c_contiguous, bint is_f_contiguous, int order_char):
     # update order_char based on array contiguity
     if order_char == b'A':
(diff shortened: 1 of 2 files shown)


grlee77:
Ah, I missed that you already made a PR. Looks like we have exactly the same solution just with different test cases. I can close the one I opened as duplicate.

asi1024:
@seberg LGTM! Thanks!

asi1024 (approved):

PR #7439, opened 2023-02-28
Prevent contiguous 1D arrays from switching to F-order

grlee77:
This MR fixes the behavior of array creation with `order='A'` or `order='K'` when the array is both C and F contiguous (as is the case for a 1D contiguous array). In this case, 'F' order was being selected over 'C' order while in NumPy the behavior is the other way around. This MR changes the behavior to match NumPy's.

Some new test cases were added. Prior to this MR 4 of these cases would fail as follows
```
FAILED tests/cupy_tests/core_tests/test_ndarray.py::TestNdarrayInit::test_array_1d[A-True] - assert (4, 4, 4) == (12, 12, 4)
FAILED tests/cupy_tests/core_tests/test_ndarray.py::TestNdarrayInit::test_array_1d[K-True] - assert (4, 4, 4) == (12, 12, 4)
FAILED tests/cupy_tests/core_tests/test_ndarray.py::TestNdarrayInit::test_astype_1d[A-True] - assert (4, 4, 4) == (12, 12, 4)
FAILED tests/cupy_tests/core_tests/test_ndarray.py::TestNdarrayInit::test_astype_1d[K-True] - assert (4, 4, 4) == (12, 12, 4)
```

### DIFF
--- a/cupy/_core/internal.pyx
+++ b/cupy/_core/internal.pyx
@@ -498,15 +498,15 @@ cpdef int _update_order_char(
         bint is_c_contiguous, bint is_f_contiguous, int order_char):
     # update order_char based on array contiguity
     if order_char == b'A':
-        if is_f_contiguous:
+        if is_f_contiguous and not is_c_contiguous:
             order_char = b'F'
         else:
             order_char = b'C'
     elif order_char == b'K':
-        if is_f_contiguous:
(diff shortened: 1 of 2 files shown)


grlee77:
closing as duplicate with #7438

PR #7457, opened 2023-03-06
Fixup array/asarray call to prefer C order on plain NumPy arrays

seberg:
This is a small followup to also prefer C order over F when converting from a NumPy array.

I had not realized that the original issue mentioned another similar path where it may be better to prefer C order strides (just for symmetry to NumPy, since I don't think the strides were ever wrong).


asi1024:
The CI failures seem to be unrelated. LGTM!

Issue #7414, opened 2023-02-20
Implementaion of numpy.ndarray.itemset in CuPy


oishigyunyu:
### Description

Hello. 

I'd like to implement the functin of `numpy.ndarray.itemset` in CuPy #6078.

I don't know how to implement the function as a class method of `cupy.ndarray`.

From the [document of NumPy](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.itemset.html), this function was implemented as a inplace function. 
```
>>> np.random.seed(123)
>>> x = np.random.randint(9, size=(3, 3))
>>> x
array([[2, 2, 6],
       [1, 3, 6],
       [1, 0, 1]])
>>> x.itemset(4, 0)
>>> x.itemset((2, 2), 9)
>>> x
array([[2, 2, 6],
       [1, 0, 6],
       [1, 0, 9]])
```
Normally, funtion requires a variable to return values, but this method doesn't require it.

How Can I implement this function? 

Sincerly,

kmaehashi:
Hi @oishigyunyu, thanks for the interest. `cupy.ndarray` class methods are defined at: https://github.com/cupy/cupy/blob/9816e20b3915dd28bc54a7549f28f69633ceb112/cupy/_core/core.pyx#L95

oishigyunyu:
@kmaehashi 
Thank you for your suggesting.
I will work on this issue, then I will reopen the pull request.

kmaehashi:
Moving the discussion to #7569

PR #7459, opened 2023-03-07
Implement numpy.ndarray.itemset

oishigyunyu:
Hello.

I've implemented `numpy.ndarray.itemset` in CuPy referenced at issue of #7414 .

I wrote the test code and tested for 3 diffrent shepe arrays both of cupy and numpy.

Could you please check it?

Thank you.


cupy/_core/core.pyx
```diff
@@ -485,7 +485,32 @@ cdef class _ndarray_base:
         """
         return self.get().tolist()
 
-    # TODO(okuta): Implement itemset
+    
+    cpdef itemset(self, index, value):
```
kmaehashi, line 489: `itemset` should be callable with one argument (if size of self is 1). Use `*args` here.

Basically, this method must behave equivalently to the following NumPy code:
https://github.com/numpy/numpy/blob/ea0b1708b27e683201285da53c6afda24bc8ba74/numpy/core/src/multiarray/methods.c#L2985-L2987
https://github.com/numpy/numpy/blob/ea0b1708b27e683201285da53c6afda24bc8ba74/numpy/core/src/multiarray/methods.c#L768

tests/cupy_tests/indexing_tests/test_itemset.py
```diff
@@ -0,0 +1,26 @@
...
+    {'shape': (3, 256, 256), 'index': (0, 100, 200), 'value': 5},
+    {'shape': (20, 3, 128, 128), 'index': (560), 'value': 255},
+    {'shape': (20, 3, 128, 128), 'index': (2, 0, 19, 34), 'value': 5},
+)
+@testing.gpu
+class TestItemset(unittest.TestCase):
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_itemset(self, xp, dtype):
```
kmaehashi, line 23: Please cover the following cases:
* shape = 0-dim (size = 1)
* shape = 1-dim (size = 1)
* dtype of array and value are different
* exception cases

kmaehashi (changes requested):

cupy/_core/core.pyx
oishigyunyu: Thank you for checking my code.
I've forgotten the case of if size of self is 1.
I will fix this and also test code.

kmaehashi:
As discussed in #7569 this PR is open to everyone to takeover.

PR #7462, opened 2023-03-10
ENH: Explorer a path to string (and structured dtype) support for cupy

seberg:
This is a squash of multiple changes (since it was unwieldy anyway) and it is *not* a polished solution, rather *basis for discussion*.

To outline the design idea a little bit:

1. Create a C++ class to describes strings (this is very much a prototype).  This is in `numpystring.h`
2. Strings may need more casting work or work to define the correct result dtypes for example.  So, modelled on NumPy, I added a `resolve_dtypes` to the `_Op`. It is used by adding `_Op`s directly, see `_ufunc.py` or the `==` code.
3. Casting is a bit special, so I have a special path for it unfortunately, to encode that `int64 -> S` cast needs a certain length.
4. I added a refactor to auto-include the string header whenever the string is used (it may be added multiple times). (That is not strictly needed, but it allows not including it by default.  A similar thing could be done for complex now, but the change is not 100% as of now.)

There are some related smaller changes of course.  And e.g. the way that loops gets selected work well enough for strings, but has its limits (something that NumPy has a newer different mechanism for because of that).  That is probably OK for the moment, I think.

Scalar strings do not work.  Casting is only implemented for strings and for long<->string casts.  (All others will just fail compliation.)

---

Now, for me the main questions/discussion points would be about whether this type of design seems fair, and it is easier to discuss with some code :).
If it is, I will try to polish it up a bit, add scalars support and at least better errors for the casts that do not work, I suspect.

This does add the complexity of `resolve_descriptors()` and it would choose that complicated DTypes classes should have a C++ equivalent.  I don't think that is strictly necessary all to it: `routine` could become a function so that it can be adapted for the string length in theory! (but it probably helps if at least conversions work.)

I am not sure strings are worth the complexity!  OTOH, I have hope that we can create a C++ tuple-like class to describe a large class of structured DTypes (or wait for CCCL and use tuples, the main thing to work out is that in NumPy `arr = np.ones(3, "i,i,i")[["f0", "f1"]]` creates a dtype with a "hole" and that hole needs to be encoded, but I think we can do that with a simple no-op C++ object).

It might also be interesting for the longer-term future, to help set up infrastructure that might allow users to write custom kernels for custom C++ classes (which of course would need some form of "dtype" object either compatible with NumPy, or CuPy would have to wrap NumPy)?


leofang:
@seberg I guess you forgot to check in the headers? 😂

cupy/_core/include/cupy/numpystring.h
```diff
@@ -0,0 +1,150 @@
...
+    __host__ __device__ NumPyString& operator=(const NumPyString<OT, Olen> &other)
+    {
+        // NOTE: Unlike NumPy, we just cast U->S (unsafe).
+        for (int i = 0; i < this->maxlen; i++) {
+            this->data[i] = other[i];
+        }
+        return *this;
+    }
+
+    // TODO: Howto best template it for all integers?!
```
leofang, line 115: One way I can think of is to make `long` here as a template parameter T, and use std::enable_if to enforce that only integer T is allowed.

cupy/_core/include/cupy/numpystring.h
```diff
@@ -0,0 +1,150 @@
...
+        return *this;
+    }
+    __host__ __device__ NumPyString& operator=(const long long &value)
+    {
+        integer_to_string(value, maxlen_, this->data);
+        return *this;
+    }
+
+
+    // TODO: Howto best template it for all integers?!
```
leofang, line 128: I didn't try but I wonder if the same enable_if trick can be used for conversion operators too?

cupy/_core/include/cupy/numpystring.h
```diff
@@ -0,0 +1,150 @@
...
+}
+
+
+template<typename T, int maxlen_>
+class NumPyString {
+public:
+    static const int maxlen = maxlen_;
+    // TODO: 0 is possible, but C doesn't like it here (for good reasons)
+    //       there may be a way to tell C++ that empty is OK/possible?
+    T data[maxlen_ ? maxlen_ : 1];
```
leofang, line 76: Does this work? Don't we need to mark this as `__host__ __device__`?

cupy/_core/include/cupy/numpystring.h
seberg: Yeah, I need to strengthen my C++ foo in this area.  I suspect the problem I had with `enable_if` was/may have been that we probably need CCCL to have the type traits header available?

leofang: Ah, we have it here
https://github.com/cupy/cupy/blob/c2133b1b5d723893b06e0ae4feec1bfd2098618d/cupy/_core/include/cupy/carray.cuh#L10-L26
so that we don't need CCCL (for the time being). I don't recall if jitify is needed or not.

cupy/_core/_scalar.pyx
```diff
@@ -65,12 +65,37 @@ cdef object _numpy_float_ = numpy.float_
 cdef object _numpy_complex_ = numpy.complex_
 
 
-cpdef str get_typename(dtype):
+cpdef tuple get_typename_with_preamble(dtype):
```
emcastillo, line 68: Would it be possible to separate this? the preamble applies to string data types only so maybe having a separate routine for the preamble that is only called when the any of the args is a string may be better and avoid changing a lot of code in kernel.pyx

emcastillo:
Thanks for working on this!
Gave it a look and it looks promising!
I am not sure if we want to do the autodetect header, it seems to be making the kernel.pyx code more complicated than it already is and I am not sure of the CPU side overhead that this may cause. I think that just checking the types of the arguments and add the preamble if one of them is a string should keep things much more simple? (not 100% sure)

> This does add the complexity of resolve_descriptors() and it would choose that complicated DTypes classes should have a C++ equivalent. I don't think that is strictly necessary all to it: routine could become a function so that it can be adapted for the string length in theory! (but it probably helps if at least conversions work.)

With descriptors you mean the struct/class declarations in C++ right? I think we could do some kind of templating to create the C++ ones from simple python descriptions.

> I am not sure strings are worth the complexity! OTOH, I have hope that we can create a C++ tuple-like class to describe a large class of structured DTypes (or wait for CCCL and use tuples, the main thing to work out is that in NumPy arr = np.ones(3, "i,i,i")[["f0", "f2"]] creates a dtype with a "hole" and that hole needs to be encoded, but I think we can do that with a simple no-op C++ object EDIT: OK, its not that simple, since arr[["f2", "f0"]] creates a dtype with swapped fields, so they have a weird offset, which would require a more complicated solution).

Not sure if we can reuse some of the C++ side Indexer mechanics?


> It might also be interesting for the longer-term future, to help set up infrastructure that might allow users to write custom kernels for custom C++ classes (which of course would need some form of "dtype" object either compatible with NumPy, or CuPy would have to wrap NumPy)?

I think this can be done with RawKernel and an array of bytes, just passing the pointer around and casting it to the appropriate object inside the kernel is a trick we use a lot. I am not 100% sure if you mean this.

cupy/_core/_scalar.pyx
seberg: Thanks for the look, I am on a summit right now, so will dive back in more later.  The reason I did it this way was to avoid checking for string explicitly.  Also so that tagging it on for structured dtypes should not have to tuch `kernel.pyx` anymore than.  (And in theory even allowing custom ones as long as they can hook in here.)

OTOH, yeah I can probably avoid quite a bit of code churn by just special casing (I don't remember how much code churn was for this and how much for preserving the length information).

seberg: Sorry for the late followup.  The changes in `kernel.pyx` are roughly split in half, I think:
* Threading this in is one change (mostly simple changes I think)
* The other part is making sure that we don't lose the string length information.

I could try to at least refactor that into two commits for clarity.  It still seems somewhat useful to have a single point where we go from Python to C, rather than do this and then check again if any of the types in the kernel need additional preamble information?
But, I agree that I mainly think that, because I was hoping to re-use this for structured dtypes and potentially broaden it up even to allow users to hook into it.

seberg:
Sorry realized that I never answered everything above, but was wondering how I could unstick this a bit.  I can move the header/preamble things, but I am not quite sure to where since this is the place where the types are converted to a string (so it seems natural to deal with it there?).

> With descriptors you mean the struct/class declarations in C++ right

This is a NumPy C name: "descriptor" is what we call the dtype instance in C in NumPy `np.dtype("S8")`.  This can be mapped to the exact C++ type (e.g. `NumPyStrig<char, 8>`).  I sometimes use it to distinguish it from the type `NumPyString` or `type(np.dtype("S8"))` which has no string length information.

> I think this can be done with RawKernel and an array of bytes, just passing the pointer around and casting it to the appropriate object inside the kernel is a trick we use a lot. I am not 100% sure if you mean this.

What I imagined was that you would allow users to register their own C++ type (for a given NumPy extension one).  Allowing indexing/copying to work on types cuda doesn't know.
But I don't think this is important: if we have strings and structured, generalization to user defined ones might be possible, I guess it shouldn't be a priority.

seberg:
Sorry, should have come back quickly :/.  I have to follow up on some NumPy last minute things, but will prioritize making this PR run through successfully again.

Were there any smaller/larger changes you would like to see, or should we discuss those?

> One thing that @asi1024 is concerned is that for strings, when we operate them together, we need to look into the actual values of the array to determine the dtype of the result and this incurs in device synchronization.

Right, this is an unfortunate limitation also in NumPy.  In the meantime I added a way for the resolution step to deal with Python integers _only_ in NumPy.
I may look a bit into what it would take to do this here, although it may be easier to take it in multiple steps (so long you don't dislike the current setup).

But `string_arr * integer_arr` doesn't work with the ufunc setup (and neither does it in NumPy, the `char.multiply` function isn't a full ufunc).
NumPy is growing variable length string support now, which "solves" that issue, but is probably a bit trickier in general in CUDA (because of the need for additional memory management.)

wiwa:
Hi! Just jumping in here because I'd love to have support for simple fixed-width strings, or at least raw bytes.

seberg:
Sorry for the long wait.  I fixed up many of the things and this may be ready for more serious review now.
There may still be test failures (I am hoping not, but I had some failures locally and I am not sure if they are related.)

Based on the old state (first commit) it is mainly a set of fixes for reductions.  There are also a couple of string related fixes and basic tests for the things that work:
* Creating/round-tripping string arrays
* Assignments (e.g. advanced indexing).
* Casts between strings and integer<->strings

**New Changes (beyond cleaning up the old changes and new tests):**
* I supported 0-D and string scalars by avoiding conversion to scalars [here](https://github.com/cupy/cupy/pull/7462/commits/3749ac5ffb7edc3aa1b5e8c072ffa71f3df7e45b).   I am not sure that is the right approach.
* The original code allowed comparing bytes/strings (and needed the dedicated resolver for that).  NumPy actually aligns with Python to `np.str_("a") == np.bytes_("a")` giving False always.  Rather than doing that (for now), I thought it is maybe just as well to raise an error.
I kept the resolver even though one could possible eliminate it here, because it should be a useful pattern for other things (e.g. if someone wants to write `string + string` kernel).

My understanding is still that string support would remain not a main feature and expanding to some struct support is the really interesting things here.

emcastillo:
@asi1024 @leofang any comment about the scalar conversion?😇

jcrist:
Hi! If/when this lands we'd be _very_ interested in making use of this in [cuml](https://github.com/rapidsai/cuml). This is the only dtype missing that we'd need to support for common sklearn compatibility - the lack of current support means we have to either punt on this or return `cudf` types for non-numeric values. For our simple use case we'd really only need to pass around `dtype('U*')` arrays, and call `cp.unique`/`cp.take` on them.

Is this feature still possibly on the roadmap? If there's anything we can do to help get this in I'd be happy to advocate for getting some time to spend helping with this. If nothing else, having some insight into the roadmap for when/if this might land would be useful for us. Thanks!

seberg:
I think we can resurrect this (the NEP 50 work meant a lot of conflicts unfortunately, although it might also have made the diff smaller).
But while it was pretty far along, it is presumably also still non-trivial to finish off, so need to decide on priorities here with @leofang. This was a vessel for structured dtypes kernels also.

(I think unnormalized unique is probably supported here on grounds of it currently working via sorting. That should be changed for most dtypes, but it may be fine if that remains the implementation strings.)

seberg:
Closing this, gh-9650 contains the rest of the interesting ideas from here (except things that are very string specific).

PR #7474, opened 2023-03-22
add `cupyx.scipy.special.exprel`

pri1311:
Adds `cupyx.scipy.special.exprel`

Benchmarks:
size | dtype | SciPy | CuPy
-----|-------|-------|-----
1000 | float32 | 0.137 ms | 0.092 ms
1000 | float64 | 0.030 ms | 0.309 ms
100000 | float32 | 1.972 ms | 0.149 ms
100000 | float64 | 1.685 ms | 0.270 ms
1000000 | float32 | 26.247 ms | 0.414 ms
1000000 | float64 | 22.007 ms | 2.114 ms

<details>
<summary>Script</summary>

```
import cupy
import cupyx
import cupyx.scipy.stats

import scipy

import scipy.stats


n_warmup = 1
n_repeat = 10
dtype_set = ('float32', 'float64')


def get_time(pref):


def time(a):
    cp_a = cupy.array(a)
    times = []
    perf = benchmark(scipy.special.exprel, (a,),
    times.append(get_time(perf))
    perf = benchmark(cupyx.scipy.special.exprel, (cp_a,),
    times.append(get_time(perf))
    return times


for n in n_set:
    for dtype in dtype_set:
        a = scipy.random.random(size=n).astype(dtype) * 0.01
        times = time(a)
```
</details>

### DIFF
--- a/cupyx/scipy/special/__init__.py
+++ b/cupyx/scipy/special/__init__.py
@@ -101,6 +101,7 @@ from cupyx.scipy.special._basic import tandg  # NOQA
 from cupyx.scipy.special._basic import cotdg  # NOQA
 from cupyx.scipy.special._basic import log1p  # NOQA
 from cupyx.scipy.special._basic import expm1  # NOQA
+from cupyx.scipy.special._basic import exprel  # NOQA
 from cupyx.scipy.special._basic import cosm1  # NOQA
 from cupy._math.rounding import round_ as round  # NOQA
 from cupyx.scipy.special._xlogy import xlogy  # NOQA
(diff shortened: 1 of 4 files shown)


takagi (approved):
LGTM!

takagi:
Let me merge this PR as the CI failure is not related to the change here. Thanks, @pri1311!

takagi:
I missed that the test does not import SciPy, so it doesn't run. Would you import SciPy there in another pull request? As SciPy is optional for CuPy's test, it should look like this (scipy.special instead of scipy.signal). Thanks! 
https://github.com/cupy/cupy/blob/c8bcae476916985074f5dca4a9f6a5d10458fccd/tests/cupyx_tests/scipy_tests/signal_tests/test_splines.py#L10-L13

PR #7442, opened 2023-03-01
Add lfilter/IIR utilities to cupyx.scipy.signal

andfoy:
See https://github.com/cupy/cupy/issues/7403

This PR derives a parallel implementation to compute recursive filters based on https://doi.org/10.1145/3173162.3173168

Issue #7403, opened 2023-02-14
Tracker: cupyx.scipy.signal


andfoy:
## Description

This is a tracking issue regarding the current and future state of compatibility between `scipy.signal` and `cupyx.scipy.signal`. The sections denoted below are aligned with respect to SciPy's reference API: https://docs.scipy.org/doc/scipy/reference/signal.html

---

### Convolution
- [x] `convolve` → https://github.com/cupy/cupy/pull/3748
- [x] `correlate` → https://github.com/cupy/cupy/pull/3748
- [x] `fftconvolve` → https://github.com/cupy/cupy/pull/3828
- [x] `oaconvolve` → https://github.com/cupy/cupy/pull/4468
- [x] `convolve2d` → https://github.com/cupy/cupy/pull/3748
- [x] `correlate2d` → https://github.com/cupy/cupy/pull/3748
- [x] `sepfir2d` → https://github.com/cupy/cupy/pull/3645
- [x] `choose_conv_method` → https://github.com/cupy/cupy/pull/3464
- [ ] `correlation_lags`

---

### B-Splines
- [ ] `bspline`
- [ ] `cubic`
- [ ] `quadratic`
- [ ] `gauss_spline`
- [ ] `cspline1d`
- [ ] `qspline1d`
- [ ] `cspline2d`
- [ ] `qspline2d`
- [ ] `cspline1d_eval`
- [ ] `qspline1d_eval`
- [ ] `spline_filter`

---

### Filtering
- [x] `order_filter`  → https://github.com/cupy/cupy/pull/3645
- [x] `medfilt` → https://github.com/cupy/cupy/pull/3645
- [x] `medfilt2d` → https://github.com/cupy/cupy/pull/3645
- [x] `wiener` → https://github.com/cupy/cupy/pull/3645
- [ ] `symiirorder1`
- [ ] `symiirorder2`
- [ ] `lfilter`
- [ ] `lfiltic`
- [ ] `lfilter_zi`
- [ ] `filtfilt`
- [ ] `savgol_filter`
- [ ] `deconvolve`
- [ ] `sosfilt`
- [ ] `sosfilt_zi`
- [ ] `sosfiltfilt`
- [ ] `hilbert`
- [ ] `hilbert2`
- [ ] `decimate`
- [ ] `detrend`
- [ ] `resample`
- [ ] `resample_poly`
- [ ] `upfirdn` 

---

### Filter design
- [ ] `bilinear`
- [ ] `bilinear_zpk`
[...]

PR #7477, opened 2023-03-24
Add lfiltic and lfilter_zi to cupyx.scipy.signal

andfoy:
Depends on https://github.com/cupy/cupy/pull/7442

### DIFF
--- a/cupyx/scipy/signal/_signaltools.py
+++ b/cupyx/scipy/signal/_signaltools.py
@@ -723,6 +723,66 @@ def lfilter(b, a, x, axis=-1, zi=None):
         return out
 
 
+def lfiltic(b, a, y, x=None, axis=-1):
+    """
+    Construct initial conditions for lfilter given input and output vectors.
+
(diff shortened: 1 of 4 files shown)


cupyx/scipy/signal/_signaltools.py
```diff
@@ -723,6 +723,120 @@ def lfilter(b, a, x, axis=-1, zi=None):
         return out
 
 
+def lfiltic(b, a, y, x=None, axis=-1):
```
takagi, line 726: `scipy.signal.lfiltic` does not have `axis` parameter, so we don't have to have it. Any reason?

andfoy: It is to be consistent with respect to the `lfilter` signature. However, since SciPy does not support it, we can leave it commented and enable it again if SciPy supports it someday

takagi (approved):
LGTM!

takagi:
Let me merge this as the CI failures are not related. Thanks, @andfoy!

Issue #7250, opened 2022-12-19
Support new features in NumPy 1.24


kmaehashi:
Support NumPy 1.24 features. I've extracted items related to CuPy from the [Release Notes](https://numpy.org/doc/stable/release/1.24.0-notes.html):

- [ ] New attribute `symbol` added to polynomial classes
- [ ] `strict` option for `testing.assert_array_equal`
- [ ] New parameter `equal_nan` added to `np.unique`
- [ ] `casting` and `dtype` keyword arguments for `numpy.stack`
- [ ] `casting` and `dtype` keyword arguments for `numpy.vstack`
- [ ] `casting` and `dtype` keyword arguments for `numpy.hstack`
- [ ] The bit generator underlying the singleton RandomState can be changed
- [ ] `numpy.typing` protocols are now runtime checkable

Split from #7243.

PR #7481, opened 2023-03-25
Support NumPy 1.24: Adding `strict` option to `testing.assert_array_equal`

tysonwu:
Added `strict` option to the two related functions `assert_array_equal()` and `assert_array_list_equal()` conforming to the respective new feature in NumPy 1.24, discussed in #7250.

_[This is my very first PR in a big project. Let me know if I violated anything :) ]_

### DIFF
--- a/cupy/testing/_array.py
+++ b/cupy/testing/_array.py
@@ -88,8 +88,7 @@ def assert_array_equal(x, y, err_msg='', verbose=True, strict=False,
              are appended to the error message.
          strict(bool): If ``True``, raise an AssertionError when either
              the shape or the data type of the array_like objects does not
-             match. The special handling for scalars mentioned in the Notes
-             section is disabled.
+             match.
 
     .. seealso:: :func:`numpy.testing.assert_array_equal`
     """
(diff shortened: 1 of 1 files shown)


cupy/testing/_array.py
```diff
@@ -85,12 +86,15 @@ def assert_array_equal(x, y, err_msg='', verbose=True, strides_check=False):
...
+         strict(bool): If ``True``, raise an AssertionError when either
+             the shape or the data type of the array_like objects does not
+             match.
 
     .. seealso:: :func:`numpy.testing.assert_array_equal`
     """
     numpy.testing.assert_array_equal(
         cupy.asnumpy(x), cupy.asnumpy(y), err_msg=err_msg,
-        verbose=verbose)
+        verbose=verbose, strict=strict)
```
kmaehashi, line 97: This will make CuPy incompatible with NumPy 1.23 or earlier. I think it's better to receive kwargs and pass-thorugh it to `assert_array_equal`.

cupy/testing/_array.py
```diff
@@ -103,7 +107,8 @@ def assert_array_equal(x, y, err_msg='', verbose=True, strides_check=False):
             raise AssertionError('\n'.join(msg))
 
 
-def assert_array_list_equal(xlist, ylist, err_msg='', verbose=True):
+def assert_array_list_equal(xlist, ylist, err_msg='', verbose=True,
```
kmaehashi, line 110: This is not NumPy-compatible API, so I think `strict` options is not needed here.

kmaehashi (changes requested):

cupy/testing/_array.py
tysonwu: Got it. I have revert the addition of `strict` options here so this function remains untouched.

cupy/testing/_array.py
tysonwu: I have modified it so that now the function accepts kwargs. Modified docstring in `strict` option to explicitly state the required NumPy version.

kmaehashi:
Thanks, LGTM! Let me run remaining tests after the release this week.

kmaehashi (approved):

PR #7487, opened 2023-03-28
Let the user specify a starting vector for eigsh.

jmbr:
Sometimes a good initial guess for an eigenvector is known and it is advantageous to use it within the eigensolver. This patch adds the keyword argument ``v0`` to ``eigsh`` so that it behaves in the same way as in ``scipy.sparse.linalg``.

### DIFF
--- a/cupyx/scipy/sparse/linalg/_eigen.py
+++ b/cupyx/scipy/sparse/linalg/_eigen.py
@@ -85,7 +85,8 @@ def eigsh(a, k=6, *, which='LM', v0=None, ncv=None, maxiter=None,
         u = cupy.random.random((n,)).astype(a.dtype)
         V[0] = u / cublas.nrm2(u)
     else:
-        V[0] = cupy.asarray(v0)
+        u = v0
+        V[0] = v0 / cublas.nrm2(v0)
 
     # Choose Lanczos implementation, unconditionally use 'fast' for now
     upadte_impl = 'fast'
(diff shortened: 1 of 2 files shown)


cupyx/scipy/sparse/linalg/_eigen.py
```diff
@@ -79,8 +81,11 @@ def eigsh(a, k=6, *, which='LM', ncv=None, maxiter=None, tol=0,
     V = cupy.empty((ncv, n), dtype=a.dtype)
 
     # Set initial vector
-    u = cupy.random.random((n,)).astype(a.dtype)
-    V[0] = u / cublas.nrm2(u)
+    if v0 is None:
+        u = cupy.random.random((n,)).astype(a.dtype)
+        V[0] = u / cublas.nrm2(u)
+    else:
+        V[0] = cupy.asarray(v0)
```
kmaehashi, line 88: In CuPy, ndarray arguments must always be `cupy.ndarray` to avoid implicit host-to-device copy. So could you remove `cupy.asarray` here?

kmaehashi (changes requested):
Thanks, LGTM except minor comment. Could you add test code for `v0 != None` case?

jmbr:
> Thanks, LGTM except minor comment. Could you add test code for `v0 != None` case?

I have added the requested test in the last commit.

kmaehashi (approved):
Hi, sorry for the delay in review. LGTM!

jmbr:
Hi, is there a pending step on my end to get this pull request merged? Please let me know if that's the case so I can help.

kmaehashi:
Hi @jmbr, sorry for the delay! Test failures are irrelevant, so let's proceed to merge.

YanivDorGalron:
Hey, 
Thanks for the helpful discussion
In the eigsh function i can determine the number of eigenvectors to calculate. 
how can i use v0 if i have a different initial guess for each of the k eigenvectors?

PR #7490, opened 2023-03-28
Support NumPy 1.24: `dtype` and `casting` keyword arguments for `hstack`, `vstack`, `stack`

tysonwu:
Following the items listed in #7250, adding `dtype` and `casting` keyword arguments for `hstack`, `vstack`, `stack`.

The addition of these kwargs should not make CuPy incompatible with NumPy versions before 1.24, as the under-the-hood `concatenate` supports these two kwargs since NumPy 1.20.

### DIFF
--- a/cupy/_manipulation/join.py
+++ b/cupy/_manipulation/join.py
@@ -76,7 +76,7 @@ def dstack(tup):
     return concatenate([cupy.atleast_3d(m) for m in tup], 2)
 
 
-def hstack(tup):
+def hstack(tup, *, dtype=None, casting='same_kind'):
     """Stacks arrays horizontally.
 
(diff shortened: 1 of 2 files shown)


takagi:
Thanks for your work! Your implementation looks fine. Although the `stack` family just wraps `concatenate`, we'd like to have tests for the new arguments. Can I ask you to make it have such tests? You can find the current tests here:
https://github.com/cupy/cupy/blob/main/tests/cupy_tests/manipulation_tests/test_join.py

tysonwu:
Thanks for the review. I have added the tests for dtype and casting options accordingly.

takagi:
Let me merge this PR as the CI failures are not related here. Thanks, @tysonwu!

PR #7494, opened 2023-03-29
Bump version to v13.0.0a1

emcastillo:


kmaehashi:
v13.0.0a1 release will be on next month, but bumping it to clarify that main branch is for v13

PR #7453, opened 2023-03-05
Remove unused test decorators

kmaehashi:
This PR removes legacy unused test decorators: `testing.gpu` and `testing.cudnn`.

PR #7496, opened 2023-03-29
Add filtfilt to cupyx.scipy.signal

andfoy:

This is basically a line-per-line port from SciPy, the only changes are related to the initial state differences between CuPy and SciPy's `lfilter` implementation

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
@@ -564,7 +564,6 @@ class TestDeconvolve:
         return scp.signal.deconvolve(o, b)
 
 
-@testing.gpu
 @testing.with_requires('scipy')
 class TestFiltFilt:
     @pytest.mark.parametrize('size', [11, 20, 32, 51, 64, 120, 128, 250])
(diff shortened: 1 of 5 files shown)


takagi:
Seems that CI makes illegal memory access and fails on CUDA 10.2. I guess some dtype combinations make this problem. Would you try to simply let dtype parametrized instead of `for_all_dtypes_combination`?

tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
```diff
@@ -572,3 +572,81 @@ def test_deconvolve(self, order, xp, scp):
         b = testing.shaped_random((order,), xp, scale=0.3)
         o = scp.signal.convolve(x, b)
         return scp.signal.deconvolve(o, b)
+
+
+@testing.gpu
```
takagi, line 577: This decorator was here due to a historical reason and was removed in #7453.
```suggestion
```

andfoy:
@takagi, the errors are related to some Cython compilation error, do you know if this is a failure on the CIs? The tests are passing just fine in other CUDA versions, as well as my local installation

takagi:
Yes, let us merge this

takagi:
Thanks, @andfoy!

PR #7507, opened 2023-04-04
Add `cupy.scipy.linalg.bandwidth` functionality

KyanCheung:
What it says on the tin.
One thing worth noting is that since the ElementwiseKernel tests for every cell in a matrix, this method would take longer for unbanded matrices, but I doubt people would be using `bandwidth` for unbanded matrices anyways so I think it's fine. That being said, if anyone has ideas on making the code more efficient for high-bandwidth matrices that would be cool.
Linked Issue : https://github.com/cupy/cupy/issues/6324


KyanCheung:
I've added more tests, which allowed me to find that F-contiguous arrays were transposed twice by mistake. I've done a bit of a rewrite to make things a bit more clear (by introducing a new variable `A` which is C contiguous).
Also, if it's fine, I've updated `mypy` to `0.981` due to issues I had with pre-commit hooks.

emcastillo:
Sorry for the delay!, seems that the tests failures are real :(
https://ci.preferred.jp/cupy.linux.cuda121/129650/

KyanCheung:
Actually, do you recommend I add the `mypy` update into a separate PR as well? If this PR keeps getting bogged down, then at least `mypy` would get updated faster.

KyanCheung:
Hmm, just realised that this only is an issue with Windows systems, while Linux systems (like the ones I've been using) have been working fine. I'll try and get a Windows system up and running this weekend to figure out a fix to this.

KyanCheung:
Alright, fixed!

Turns out the underlying problem is that `cupy.mgrid` has dtype `int32` on Windows, but `int64` on Linux. Using a type placeholder should fix things.

emcastillo:
Test failures are unrelated

PR #7518, opened 2023-04-10
Add symiirorder2 to cupyx.scipy.signal

andfoy:
Depends on https://github.com/cupy/cupy/pull/7511


takagi (approved):
LGTM!

takagi:
Do you think it makes sense to have performance comparisons on `symiirorder1` and `symiirorder2`? I know both use `lfilter`; if possible, I'd like to have them in #7511 and #7518 for reference.

PR #7521, opened 2023-04-17
Fix TestBSpline::test_design_matrix_same_as_BSpline_call

andfoy:
This diff asserts that design matrix test raises ValueError when `extrapolate=False`


takagi (approved):
LGTM!

takagi:
Thanks, @andfoy!

PR #7357, opened 2023-01-30
Add NdPPoly to cupyx.scipy.interpolate

andfoy:
See #7186

PR #7523, opened 2023-04-18
Mark scipy required in a test

takagi:
Following up #7357.


takagi:
cuda115 test depends on #7524.

PR #7537, opened 2023-04-29
ENH: signal: add firls and freqz, freqz_zpk, sosfreqz

ev-br:
This complements `firwin`, which is available in cuSignal.

Heavy lifting is delegted to `solve` and special matrices (hankel, toeplitz), so there is not much to optimize at this level. Will need to test a bit the behavior with ill-conditioned matrices, where scipy switches from solve to lstsq.

Testing uses `freqz`, so add that as well. While at it, also add `sosfreqz` to finish up frequency response functions.

All implementations and their tests are pretty much straight vendoring scipy.signal sources, with minimal modifications. 

Two small things:

- `scipy.signal.freqz` uses an evaluation routine from `numpy.polynomial.polynomial`. This uses the Horner's scheme, unlike the old-style `cupy.poly1d` which just straight multiplies monomials. So this PR just vendors the evaluation routine. <s>If it turns out that we use a lot of new-style polynomial API, we can think of exposing it in `cupy`. For now it's juts a private routine.</s> These routines are added to `cupy.polynomial.polynomial`.
- 0D cupy arrays do not decay to python scalars in the same way numpy 0D arrays do. E.g. `operator.index(np.array(5))` succeeds but `cupy.array(5)` raises. Not sure if it's a deliberate choice or bug.  

EDIT:  depends on https://github.com/cupy/cupy/pull/7591


cupyx/scipy/signal/_fir_filter_design.py
```diff
@@ -0,0 +1,168 @@
...
+    b[1:] += m * cupy.cos(n[1:] * cupy.pi * bands) / (cupy.pi * n[1:]) ** 2
+    b = cupy.diff(b, axis=2)[:, :, 0] @ weight
+
+    # Now we can solve the equation : XXX CuPy failure modes (?)
+    with cupyx.errstate(linalg="raise"):
+        try:
+            a = solve(Q, b)
+        except LinAlgError:
+            # in case Q is rank deficient
+            a = lstsq(Q, b, rcond=None)[0]
```
ev-br, line 152: Not sure if this is a right way of detecting an ill-conditioned linear algebraic system. scipy.signal version relies on `scipy.linalg.solve` emitting a LinAlgWarning , https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html

Would appreciate a pointer @takagi

ev-br: Looking at the `scipy.linalg.solve`, it appears that it uses separate steps for LU factorization and backsubstitution [1], so that it has a chance to call lapack's `*gecon` to estimate the condition number. CuPy linalg.solve, OTOH, calls straight into `*gesv` which does not return the condition number info [2]. 

So the procedure above is not entirely correct really. However, let's look at the tests. The procedure in this PR does not pass value-to-value comparisons between scipy and cupy, but does pass the test from scipy.signal [3] locally (currently commented out).

[1] https://github.com/scipy/scipy/blob/v1.10.1/scipy/linalg/_basic.py#L40-L265
[2] https://github.com/cupy/cupy/blob/v12.0.0/cupy/linalg/_solve.py#L15
[3] https://github.com/cupy/cupy/pull/7537/files#diff-15f8bbf4433eed91ce1cc13976ead22a49f85bc207b85c04a28a896a62b49960R93

ev-br:
Also, there's a general design question for this filter design functionality. Scipy.signal versions accept array_likes (e.g. lists), while typically CuPy is more strict and requires cupy arrays. Which convention do we want to follow?

takagi:
> 0D cupy arrays do not decay to python scalars in the same way numpy 0D arrays do. E.g. operator.index(np.array(5)) succeeds but cupy.array(5) raises. Not sure if it's a deliberate choice or bug.

This behavior is CuPy's choice.

cupyx/scipy/signal/_fir_filter_design.py
takagi: Sorry for dropping the ball. `cupyx.scipy.linalg.solve` doesn't raise here with expected ill-conditioned sytem?
https://github.com/cupy/cupy/blob/60fe053011b751129c5ccc7e6828f4f23c42763e/cupyx/lapack.py#L71

takagi:
> Also, there's a general design question for this filter design functionality. Scipy.signal versions accept array_likes (e.g. lists), while typically CuPy is more strict and requires cupy arrays. Which convention do we want to follow?

We generally want to have a stricter way that requires only CuPy arrays to avoid implicit sync, however, there would be exceptional cases where it is expected that users almost always pass a list to an argument. Let's discuss it when you have such a case.

ev-br:
> We generally want to have a stricter way that requires only CuPy arrays to avoid implicit sync, however, there would be exceptional cases where it is expected that users almost always pass a list to an argument. Let's discuss it when you have such a case.

Sure, so the decision is to disallow the usage with this, unless there's a very compelling reason: 

```
In [4]: import cupyx.scipy.signal as signal

In [5]: w, h = signal.freqz([1, 2, 3], [4, 5, 6])
```

Implementation-wise, this amount to _not_ starting scipy.signal clone functions with `cupy.asarray`.

cupyx/scipy/signal/_fir_filter_design.py
ev-br: Thanks for circling back to it @takagi ! No, `solve` does not raise the error here. Looking at https://docs.nvidia.com/cuda/cusolver/#cusolverdn-dense-lapack-function-reference, ISTM that `dev_info` reports the error condition only, similar to what CPU LAPACK does. Here the scipy.signal version explicitly checks the condition number via `gecon`.
Is there an analog of  https://netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational_ga188b8d30443d14b1a3f7f8331d87ae60.html on CUDA?

takagi:
> Sure, so the decision is to disallow the usage with this, unless there's a very compelling reason

Indeed.

cupyx/scipy/signal/_fir_filter_design.py
takagi: I don't think cuSolver has `gecon` equivalent.

takagi: Can we compute the condition number using the result of `getrf` here?
https://github.com/cupy/cupy/blob/60fe053011b751129c5ccc7e6828f4f23c42763e/cupyx/lapack.py#LL69C21-L69C21

takagi: BTW, does the value-to-value comparison pass with some relaxed tolerance? Also, I think it's okay as far as the test from scipy.signal passes as it checks the condition the result should have.

ev-br: So this is a bit different, it turns out. 
There are two tests with rank-deficient LSQ problems. One indeed passes with a lower tolerance. The other runs into https://github.com/scipy/scipy/issues/18533: in a nutshell, the solution rather strongly depends on the LAPACK driver for the LSQ problem.
So the updated PR now duplicates the scipy tests and links to the scipy issue.

cupyx/scipy/signal/_filter_design.py
```diff
@@ -0,0 +1,332 @@
...
+    integer scalar.
+
+    NB: would be easier if ``operator.index(cupy.array(5))`` worked
+    (numpy.array(5) does)
+    """
+    if isinstance(x, cupy.ndarray):
+        if x.ndim == 0:
+            value = x.item()
+        else:
+            return x, False
```
takagi, line 21: `x.item()` makes device-host synchronization, so I think returning cupy.ndarray is okay.

ev-br: Sadly, a python integer is required down the line: 

```
(Pdb) start, stop, num
(0, 3.141592653589793, array(7))
(Pdb) cupy.linspace(start, stop, num)
*** TypeError: 'ndarray' object cannot be interpreted as an integer
```

cupyx/scipy/signal/_fir_filter_design.py
```diff
@@ -0,0 +1,151 @@
...
+                         "gains for %s frequencies."
+                         % (desired.size, bands.size))
+    desired.shape = (-1, 2)
+    if (cupy.diff(bands) <= 0).any() or (cupy.diff(bands[:, 0]) < 0).any():
+        raise ValueError("bands must be monotonically nondecreasing and have "
+                         "width > 0.")
+    if (bands[:-1, 1] > bands[1:, 0]).any():
+        raise ValueError("bands must not overlap.")
+    if (desired < 0).any():
+        raise ValueError("desired must be non-negative.")
```
takagi, line 91: We may have these checks responsibility of users as they brings device-host synchronization.

cupyx/scipy/signal/_fir_filter_design.py
```diff
@@ -0,0 +1,151 @@
...
+    if weight is None:
+        weight = cupy.ones(len(desired))
+    weight = cupy.asarray(weight).flatten()
+    if len(weight) != len(desired):
+        raise ValueError("weight must be the same size as the number of "
+                         "band pairs ({}).".format(len(bands)))
+    if (weight < 0).any():
+        raise ValueError("weight must be non-negative.")
```
takagi, line 99: ditto (device-host synchronization)

cupyx/scipy/signal/_filter_design.py
takagi: How about also using the following branch for 0-dim cupy.ndarray?

https://github.com/cupy/cupy/blob/483c169ed597e6a2468738b0893b8a1732aa6125/cupyx/scipy/signal/_filter_design.py#L168-L169

ev-br: That would change the interface: for `worN=asarray(5)` currently means that the frequencies are `w = linspace(0, 2*pi, 5)`, and that change would make it `w = np.asarray([5])`.

We can probably just disallow 0D arrays and make worN argument to be either a python int or a 1D array, would that work?

takagi: It's fine, thanks for the explanation!

cupyx/scipy/signal/_fir_filter_design.py
ev-br: Updated.

cupyx/scipy/signal/_fir_filter_design.py
ev-br: Updated.

takagi:
I've reran the static check.

ev-br:
Updated:

- xfailed firls tests which run into https://github.com/scipy/scipy/issues/18533
- added a test copied from scipy, which only looks at the extreme values of the filtered signal instead of a direct comparison with scipy.signal
- removed tests which depend on *ord functions. These tests are fixed up and included into https://github.com/cupy/cupy/pull/7632

This is PR is now I think ready from my side.

ev-br:
CUDA 102 and 11x CI failures are most likely the same as those dealt with gh-7627

takagi (approved):
LGTM!

takagi:
Thanks, @ev-br!

PR #7541, opened 2023-05-06
cupyx.scipy.signal: add bilinear & bilinear_zpk

ev-br:
cross-ref gh-7403

Add IIR filter converters in the TF and ZPK formats:
- [x] bilinear
- [x] lp2lp
- [x] lp2hp
- [x] lp2bs
- [x] lp2bp
- [x]  bilinear_zpk
- [x] lp2lp_zpk
- [x] lp2hp_zpk
- [x] lp2bs_zpk
- [x] lp2bp_zpk
- [x] normalize

All is pure python. Much of the content can be moved to cuda kernels, but inputs are typically very short, so not sure if this is worth it.

https://github.com/cupy/cupy/pull/7537#issuecomment-1537144791 is relevant, will need to update the PR with the outcome of that discussion.


takagi (approved):
LGTM!

takagi:
Thanks, @ev-br!

PR #7528, opened 2023-04-21
Add sosfilt to cupyx.scipy.signal 

andfoy:
See https://github.com/cupy/cupy/issues/7403

This implementation will contemplate the performance against the following baselines:

1. N sequential calls to `lfilter`, where each one of them computes each second order section.
3. N calls to convolve1d  to compute the FIR part and a extension of the `apply_iir` kernel to chained second order sections (the kernel would compute the IIR part from the FIR output in a chained fashion).
4. A generalization of `apply_iir` where the function would compute both FIR and IIR sections sequentially one after another with custom kernels for both parts.

If the third option improves the performance, then it would be possible to think about improving `lfilter` this route.

PR #7552, opened 2023-05-10
Add sosfilt_zi to cupyx.scipy.signal

andfoy:
Depends on https://github.com/cupy/cupy/pull/7528

This PR is an extension of `lfilter_zi`, which uses the same principle applied to chained second order sections.


takagi (approved):
LGTM!

takagi:
Thanks, @andfoy!

PR #7553, opened 2023-05-10
filter design "prototypes"

ev-br:
- [x] `buttap`
- [x] `cheb1ap`
- [x] `cheb2ap`
- [x] `ellipap` --- requires elliptic integrals, so add them to `cupyx.scipy.special`
- [ ] `besselap` -- will do in a follow-up PR

cupyx/scipy/special
- [x] `ellipk`
- [x] `ellipkm1`
- [x] `ellipj`

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_iir_filter_conversion.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_iir_filter_conversion.py
@@ -252,5 +252,3 @@ class TestLowLevelAP:
     @testing.numpy_cupy_allclose(scipy_name="scp")
     def test_cheb2ap(self, xp, scp):
         return scp.signal.cheb2ap(3, 1)
-
-
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_iir_filter_conversion.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_iir_filter_conversion.py
@@ -253,6 +253,6 @@ class TestLowLevelAP:
     def test_cheb2ap(self, xp, scp):
         return scp.signal.cheb2ap(3, 1)
 
-    @testing.numpy_cupy_allclose(scipy_name="scp", atol=1e-5, rtol=1e-5)
+    @testing.numpy_cupy_allclose(scipy_name="scp", atol=3e-5, rtol=3e-5)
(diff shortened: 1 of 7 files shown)


ev-br:
Would appreciate help with the CI failures. AFAIU, the issue is that this PR adds a couple of ufuncs in scipy.special with `.types = "d->d"` only, while their scipy.special analogs also have an `f->f` loop. Do I need to add a special case to https://github.com/cupy/cupy/blob/main/tests/cupyx_tests/scipy_tests/special_tests/test_ufunc_dispatch.py#L44 ?

takagi:
LGTM except for the static checks!

ev-br:
A rebase on main fixed the style checks, it seems.

takagi:
doc build does not have CUDA, so `_POW10_LOG10 = cupy.log(10)` causes an error on import time. Having numpy scalar, `numpy.log(10)`, would be enough for the precomputed value to multiply with a `cupy.ndarray` instead.

ev-br:
Thanks! using `math.log(10)` seems to have done the trick.

takagi:
> Thanks! using math.log(10) seems to have done the trick.

Sound good! FYI, scalar values are passed to GPU via cudaLaunchKernel's argument, so we don't need explicit H2D copy here.

takagi (approved):
LGTM!

tests/cupyx_tests/scipy_tests/signal_tests/test_iir_filter_conversion.py
```diff
@@ -237,3 +237,22 @@ def test_basic(self, xp, scp):
...
+
+    @testing.numpy_cupy_allclose(scipy_name="scp")
+    def test_cheb1ap(self, xp, scp):
+        return scp.signal.cheb1ap(3, 1)
+
+    @testing.numpy_cupy_allclose(scipy_name="scp")
+    def test_cheb2ap(self, xp, scp):
+        return scp.signal.cheb2ap(3, 1)
+
+    @testing.numpy_cupy_allclose(scipy_name="scp", atol=1e-5, rtol=1e-5)
```
takagi, line 256:
```suggestion
    @testing.numpy_cupy_allclose(scipy_name="scp", atol=3e-5, rtol=3e-5)
```

takagi:
Thanks, @ev-br!

Issue #7557, opened 2023-05-11
[RFC] Drop support for CUDA 10.2/11.0/11.1 in CuPy v13


kmaehashi:
The CuPy team plans to drop support for CUDA 10.2, 11.0, and 11.1 in CuPy v13, which will be released in October 2023 (#7555). The primary motivation for this change is to reduce the work necessary to publish a CuPy release. These legacy CUDA versions require separate packaging for each version as they don't support CUDA Minor Compatibility policy. Reducing the number of packages also helps keep the PyPI community healthy in terms of the disk space and bandwidth required to distribute packages.

Here are the download counts of our binary packages in the recent three months (sdist and `cupy-wheel` are excluded):

![CuPy PyPI Binary Package Downloads (Feb 2023 - Apr 2023)](https://github.com/cupy/cupy/assets/939877/2db9df60-a759-4c0f-a9f9-b378b42d12e0)

According to the statistics, 22.3% of users will be affected by this change. However, users of these CUDA versions can continue to use CuPy v12, which we believe is stable enough and will continue to be maintained until the release of the CuPy v13. The number also shows that 31.2% of users are still on CUDA 8.0~10.1, which is no longer supported, indicating they are satisfied with the legacy CuPy releases (v9 or earlier). That said, we recommend upgrading to the latest CUDA release, which provides better performance.

Please leave a comment on this issue if you have any feedback on this change.

kmaehashi:
RFC for a month, and there doesn't seem to be any problem. Let's proceed.

jakirkham:
Thanks all! 🙏

PR #7567, opened 2023-05-16
Try to fix nan value mismatches in filtfilt tests

andfoy:
After https://github.com/cupy/cupy/pull/7496, some failures started to appear in the `cuda102` CIs regarding `nan` and zeros mismatches. Since these errors do not appear (and cannot be reproduced) in CIs that run newer CUDA versions, they are most probably related to unstable input coefficients which cause `lfilter_zi` to be `nan`, which then when they are applied to a signal that is zero, it will output `nan` as opposed to zero.

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_signaltools.py
@@ -744,6 +744,13 @@ class TestFiltFilt:
         ):
             # ROCm 4.3 raises in Module.get_function()
             pytest.skip()
+        if (
+            not runtime.is_hip
+            and cupy.cuda.runtime.runtimeGetVersion() == 10020
+        ):
+            # The tests fail on CUDA 10.2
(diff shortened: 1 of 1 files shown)


takagi:
We can only merge this to v12 as CUDA 10.2 is planned to drop in v13.

takagi:
Would you move this change on top of `v12` branch?

kmaehashi:
CUDA 10.2 drop is under RFC in #7557 although no one complained so far. How about skipping these tests for CUDA 10.2 (`cupy.cuda.runtime.runtimeGetVersion() == 10020`) like we do in 
https://github.com/cupy/cupy/blob/6f135c488e87aafda0b947c5d766545a3b29ae3a/tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py#L1112-L1114, and merge this one to both main/v12 branch?

takagi:
Thanks, @andfoy!

takagi:
No need to make a backport.

PR #7574, opened 2023-05-19
Fix CUDA Python CI failure

kmaehashi:
Fixes the CI failure of CUDA Python. THe CUDA Python installed in the environment cannot be found during CuPy build because the build is done in isolated environment if `wheel` package is missing.

https://pip.pypa.io/en/stable/news/#v23-1

> When the wheel package is not installed, pip now uses the default build backend instead of setup.py install and setup.py develop for project without pyproject.toml.


emcastillo (approved):

kmaehashi:
Test passed!

PR #7578, opened 2023-05-22
Fix `_depends.json` not included in wheel

kmaehashi:
The feature added in #7279 was not working in shipped wheels as the `_depends.json` file is not included in the wheel.

c.f. https://github.com/pypa/setuptools/issues/1064

### DIFF
--- a/install/cupy_builder/_command.py
+++ b/install/cupy_builder/_command.py
@@ -189,6 +189,7 @@ class custom_build_ext(setuptools.command.build_ext.build_ext):
 
             depends_json = os.path.join(
                 self.build_lib, 'cupy', '.data', '_depends.json')
+            os.makedirs(os.path.dirname(depends_json), exist_ok=True)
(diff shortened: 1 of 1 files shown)


leofang:
Q: Is this something that we should also include in the conda package?

kmaehashi:
> Q: Is this something that we should also include in the conda package?

It's nice to have. CuPy ignores it silently if it is missing. This PR let `setup.py` generate `cupy/.data/_depends.json` under `build/libs.*` on Windows, but not sure if conda packaging will include it automatically.

kmaehashi:
Test failure is irrelevant.

takagi (approved):
LGTM!

leofang:
> but not sure if conda packaging will include it automatically.

Confirmed: https://github.com/conda-forge/cupy-feedstock/pull/218#issuecomment-1611235676

kmaehashi:
Thanks for checking, Leo!

PR #7580, opened 2023-05-22
Improve README and Installation Guide

kmaehashi:
Recently we are getting more reports of installation failures, and it looks most issues are happening with CuPy wheels installed on conda env. This PR adds the instruction for Conda users to README to promote using packages from conda-forge.

This PR also removed Coveralls badge which is out-dated, and added several wording improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,10 @@ Choose the right package for your platform.
 | CUDA 11.1             | x86_64            | `pip install cupy-cuda111`                                    |
 | CUDA 11.2 ~ 11.8      | x86_64 / aarch64  | `pip install cupy-cuda11x`                                    |
 | CUDA 12.x             | x86_64 / aarch64  | `pip install cupy-cuda12x`                                    |
-| ROCm 4.3 (*)          | x86_64            | `pip install cupy-rocm-4-3`                                   |
-| ROCm 5.0 (*)          | x86_64            | `pip install cupy-rocm-5-0`                                   |
+| ROCm 4.3 (*[experimental](https://docs.cupy.dev/en/latest/install.html#using-cupy-on-amd-gpu-experimental)*)          | x86_64            | `pip install cupy-rocm-4-3`                                   |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -51,18 +52,28 @@ Choose the right package for your platform.
...
-## Run on Docker
+### Conda
+
+Binary packages are also available for Linux and Windows on [Conda-Forge](https://anaconda.org/conda-forge/cupy).
+
+| Platform              | Architecture                | Command                                                       |
+| --------------------- | --------------------------- | ------------------------------------------------------------- |
+| CUDA                  | x86_64 / aarch64 / ppc64le  | `conda install -c conda-forge cupy`                           |
+
+If you need to use a particular CUDA version (say 11.8), you can do `conda install -c conda-forge cupy cudatoolkit=11.8`.
```
leofang, line 64: This is future proof (works now and will continue to work after CUDA 12 is up):
```suggestion
If you need to use a particular CUDA version (say 11.8), you can do `conda install -c conda-forge cupy cudatoolkit cuda-version=11.8`.
```
xref: [`cuda-version`](https://github.com/conda-forge/cuda-version-feedstock/blob/main/recipe/README.md)

leofang:
Would be nice to also copy/paste this sentence: https://github.com/cupy/cupy/blob/76906e5a2c7e3edf07af8cfb9118906cee9aa38a/docs/source/install.rst?plain=1#L144-L146
Ideally, conda package users should file bug reports or support requests to the feedstock, not here (as it'd be too noisy).

README.md
leofang: Ah it can be simplified
```suggestion
If you need to use a particular CUDA version (say 11.8), you can do `conda install -c conda-forge cupy cuda-version=11.8`.
```

kmaehashi: Thanks! Also updated the installation guide.

kmaehashi:
> Would be nice to also copy/paste this sentence:

Done!

leofang (approved):

emcastillo (approved):

PR #7594, opened 2023-05-25
update badges

12rambau:
The conda badge was not displaying on my side. I changed the link and dropped the svg extention where it was not necessary

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 
 # CuPy : NumPy & SciPy for GPU
 
-[![pypi](https://img.shields.io/pypi/v/cupy.svg)](https://pypi.python.org/pypi/cupy)
-[![Conda Version](https://img.shields.io/conda/vn/conda-forge/cupy.svg)](https://anaconda.org/conda-forge/cupy)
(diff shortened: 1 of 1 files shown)


kmaehashi:
Hi @12rambau, thanks for the PR! The change looks fine - I guess shields.io used to append `.svg` extension ([NumPy README](https://raw.githubusercontent.com/numpy/numpy/main/README.md) also uses the URL with the extension), but now it doesn't.

> The conda badge was not displaying on my side.

On which page, specifically?
https://github.com/cupy/cupy rendering fine on my side, and both of these URLs seems working:
* https://img.shields.io/conda/vn/conda-forge/cupy.svg
* https://img.shields.io/conda/vn/conda-forge/cupy

12rambau:
that was on the github page and only for the conda forge one. I happened to change them all to remain consistent.

kmaehashi:
> that was on the github page and only for the conda forge one. I happened to change them all to remain consistent.

Thanks for the explanation. That sounds like a tentative issue on GitHub's proxy (actually, images on GitHub markdown are HTTP-proxied and thus do not render these URLs directly), but anyway it's better to update the URL to the latest format generated by shields.io.

PR #7595, opened 2023-05-26
Bump versions of static checkers

kmaehashi:
This fixes the problem when applying static checkers in Python 3.10+.


takagi (approved):
LGTM!

takagi:
CI failure is not related.

PR #7597, opened 2023-05-26
Fix `cupy-wheel` package installation fails with pip 23.1+

### LINKED ISSUE #7593: `pip install cupy-wheel` - "not a valid wheel filename"
### Description

When I try to `pip install cupy-wheel` I get the error `cupy_wheel<...>.whl is not a valid wheel filename`.

### To Reproduce

```bash
pip install cupy-wheel
```
OS: Windows 10
Python: 3.8.9
pip: 23.1.2
CUDA: CUDA 11.7

### Installation

None

### Environment

Unable to install using `cupy-wheel`. Can install using `pip install cupy-cuda11x`.

### Additional Information

_No response_

kmaehashi:
Fixes #7593.

`META_VERSION` has been introduced to allow releasing a hot-fix for `cupy-wheel`, e.g., `cupy-wheel==12.1.0.post1` install `cupy-cuda12x==12.1.0`.

### DIFF
--- a/install/universal_pkg/setup.py
+++ b/install/universal_pkg/setup.py
@@ -2,12 +2,13 @@ import ctypes
 import pkg_resources
 import os
 import sys
-from typing import Dict, List, Optional
+from typing import List, Optional
 
 from setuptools import setup
 
 
 VERSION = '13.0.0a1'
+META_VERSION = VERSION
 
 # List of packages supported by this version of CuPy.
 PACKAGES = [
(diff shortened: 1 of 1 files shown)


leofang:
Thank you for the detailed investigation, @kmaehashi, this sound super annoying +1. I was hoping cupy-wheel to be the golden solution going forward... 😞

> This means that users will have to run pip cache remove cupy_wheel before re-installing cupy-wheel. 

Q: what happens with `pip install --no-cache-dir ...`? Is it not honored anymore?
> I will also explore the possibility of supporting multiple CUDA versions in a single CuPy binary package in https://github.com/cupy/cupy/pull/6908.

Regarding multi CUDA version support, I may have another suggestion, will create an issue later.

kmaehashi:
> Q: what happens with `pip install --no-cache-dir ...`? Is it not honored anymore?

This still works, although [not recommended according to the pip docs](https://pip.pypa.io/en/stable/topics/caching/#cache-management).

emcastillo (approved):

PR #7508, opened 2023-04-04
Introduce `cython-lint`

pri1311:
Closes: #7465 
Hi, PR aims to introduce cython-lint and fix the current issues that violate the lining rules.

PR #7602, opened 2023-06-02
Fix flake8-cython not working

kmaehashi:
This had to be in #7595 to enable flake8 lint against Cython.

I think these config can be removed after a certain period of time, once #7508 got merged.

### DIFF
--- a/.flake8.cython
+++ b/.flake8.cython
@@ -1,4 +1,4 @@
 [flake8]
 filename = *.pyx, *.pxd, *.pxi
 exclude = .git, .eggs, *.py
-ignore = W503,W504,E225,E226,E227,E402,E999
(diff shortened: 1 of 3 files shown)


.flake8.cython
```diff
@@ -1,4 +1,4 @@
 [flake8]
 filename = *.pyx, *.pxd, *.pxi
 exclude = .git, .eggs, *.py
-ignore = W503,W504,E225,E226,E227,E402,E999
+ignore = W503,W504,E225,E226,E227,E275,E402,E999
```
kmaehashi, line 4: E275 is to silence errors at `except?` (`missing whitespace after keyword`).

emcastillo (approved):
Pre test is passing so it should be fine!

PR #7603, opened 2023-06-02
Patch the build system to better support conda-build

### LINKED ISSUE #8124: Hardcoded lib path in build script causes error
### Description

Library path on setup_build is [hardcoded](https://github.com/cupy/cupy/blob/f215deeb7a2f3cfd046a3f3f80c59cfb4499f8b6/install/cupy_builder/cupy_setup_build.py#L263). Some systems, such as Debian, do not install nvidia libraries to /usr/lib64/ when installing `nvidia-cuda-dev` and `nvidia-cuda-toolkit` through apt:

```
$ find /usrt/ -name libcudart_static*
/usr/lib/x86_64-linux-gnu/libcudart_static.a
```

I solved this issue by doing a symlink:

```
ln -s /usr/lib/x86_64-linux-gnu/libcuda* /usr/lib64/
```

However, the correct solution for this issue would be to not hardcode paths or add conditions to it.

### To Reproduce

```
poetry add cupy
```
On Debian:
```
$ uname -a
Linux anon 6.6.9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.6.9-1 (2024-01-01) x86_64 GNU/Linux
```

### Installation

Source (`pip install cupy`)

### Environment

```
OS                           : Linux-6.6.9-amd64-x86_64-with-glibc2.37
Python Version               : 3.11.7
CuPy Version                 : 13.0.0
NumPy Version                : 1.26.3
SciPy Version                : 1.11.4
CUDA Root                    : /usr
nvcc PATH                    : /usr/bin/nvcc
CUDA Build Version           : 12020
CUDA Driver Version          : 12000
CUDA Runtime Version         : 12020 (linked to CuPy) / 12000 (locally installed)
cuFFT Version                : 11001
cuRAND Version               : 10301
cuSOLVER Version             : (11, 4, 3)
NVRTC Version                : (12, 0)
Device 0 Name                : NVIDIA GeForce GTX 1050
Device 0 PCI Bus ID          : 0000:08:00.0
```

### Additional Information

_No response_

leofang:
Patch taken from https://github.com/conda-forge/cupy-feedstock/pull/199.

Context: 
1. The conda compiler is a cross compiler by design (and it turns out nvcc is, too). It means that a conda package can be cross-compiled instead of native-compiled. In other words, the build platform is not necessary the same as the target platform. But this creates an issue in particular for CuPy, because CuPy has never promised to support build != target. In short, CuPy has an Autotool-like dependency checking system (backed by `build_shlib()` and `build_and_run()`) that essentially checks the dependencies on the build platform (say linux64), but the actual binaries on the target platform (say aarch64) cannot be linked by the cross compiler (on linux64). A linker failure is raised when attempting to do so.
2. The [CUDA 12 bringup on conda-forge](https://github.com/conda-forge/staged-recipes/issues/21382) further reinforces the need to support this use case. As mentioned, we learned that nvcc is also a cross compiler (and people urge for this capability crazily in order to reduce the package build time), and when transitioning to the new package layout we see that it's best to assume a cross compiler is in use. So the patch that was originally written [to only support cross compiling](https://github.com/conda-forge/cupy-feedstock/pull/197) (+ CUDA 11), can also be tweaked to support CUDA 12 on conda-forge.

Approach:
- This patch is turned on only during conda-build. This is possible by detecting a few env vars set by conda-build; that is, when building CuPy on CF. It's a no-op for 99% of the time.
- This patch is written to support both the old layout (CUDA 11) and new layout (CUDA 12) on conda-forge.
- This patch works with both native and cross compiling.
- ~~This patch has not been validated on Windows yet, because~~
  - ~~It makes little sense to cross compile from linux64 to win64~~
  - ~~CUDA 12 + win64 is [not supported yet](https://github.com/conda-forge/cupy-feedstock/issues/205)~~
  - **Update**: Windows support is included (https://github.com/cupy/cupy/pull/7603#issuecomment-1819330466)

Also close #8124.


install/cupy_builder/install_build.py
```diff
@@ -16,6 +18,8 @@
 from cupy_builder._context import Context
 
 
+logging.basicConfig(level=logging.DEBUG)
```
kmaehashi, line 21: Question: Is `logging` mandatory? It's better if we can avoid mutating global states because this file is called from unit tests.

install/cupy_builder/install_build.py
```diff
@@ -689,9 +693,59 @@ def get_cusparselt_version(formatted=False):
...
+    # because we create a fresh compiler instance that does not honor CFLAGS etc set
+    # in the conda-build environment.
+    include_dirs = list(include_dirs)
+    library_dirs = list(library_dirs)
+
+    if os.environ.get('CONDA_BUILD_CROSS_COMPILATION'):
+        # If we're cross compiling, we need to generate stub files that are
+        # executable in the build environment, not the target environment.
+        # This assumes, however, that the build/host environments see the same
+        # CUDA Toolkit.
```
kmaehashi, line 721: Question: Is it possible for the conda-forge build environment to run [`docker run --rm --privileged multiarch/qemu-user-static --reset -p yes`](https://github.com/multiarch/qemu-user-static) before building CuPy? This should allow the host to run aarch64/ppc64le binaries via QEMU.

This technique was originally used to build Jetson wheels on x86_64. https://github.com/cupy/cupy-release-tools/blob/d588c6f155f1b43ef983f75ffcd556fa6b8f6215/dist.py#L420-L428

install/cupy_builder/install_build.py
leofang: Yes I can hide it under the env var `CONDA_BUILD`.

install/cupy_builder/install_build.py
kmaehashi: Thanks, got it. I thought the overhead of running cross-compiled stubs on QEMU could be negligible compared to that of running all the compilation on QEMU. But anyway let's merge as-is, it's working great.

install/cupy_builder/install_build.py
kmaehashi: Thanks, then could you put it under `CONDA_BUILD` check (or just replace the logging with `print` also work?)

install/cupy_builder/install_build.py
leofang: Yeah with the 2-core Azure runners it's hard to do any optimization for compilation intensive packages... 😢 Cross compilation is also bad in that it doesn't even allow basic sanity checks (like `import cupy`) and I am not a fan either, but it gets the problem solved for now. (Actually we have a new problem in the feedstock, but I can't get to that this week...) I hope to revisit this when the CF GPU CI is up. At least we'd have a more beefy CPU.

install/cupy_builder/install_build.py
leofang: Done in commit [35b836b](https://github.com/cupy/cupy/pull/7603/commits/35b836b8ca16211b4e504f587730cc39e1e2159e).

leofang:
FYI @kmaehashi this is ready 🙂

leofang:
Update: I updated the PR with Windows support (from https://github.com/conda-forge/cupy-feedstock/pull/228).

leofang:
@kmaehashi Shall we try to get this merged (and backported) before the next release, so that we can avoid maintaining the patch in cupy-feedstock?

leofang:
Hi @kmaehashi any chance we can get this PR merged and included in v13.2.0? It's getting more and more challenging for me to maintain a big patch for conda-forge... 😅

jakirkham:
/test mini

(to include now required CUDA 12.5 checks)

kmaehashi (approved):

PR #7610, opened 2023-06-05
Fix build-cuda test `restore-keys` not working

kmaehashi:
`restore-keys` were not working since #7546... :bow:


emcastillo (approved):
Merging as this is GHA related only

Issue #7620, opened 2023-06-09
Allow `import cupy` without CUDA installation


kmaehashi:
Currently, `import cupy` requires the GPU toolkit (CUDA/ROCm) installed. However, several people reported this is cumbersome, especially when developing or running NumPy/CuPy agnostic code under a CPU-only machine.

This is a tracker issue to implement idea 2 discussed in the RFC pull-request (https://github.com/cupy/cupy/pull/6908#issuecomment-1196459041).

```[tasklist]
### Modules
- [x] NVRTC #7621
- [x] cuSOLVER (depends on cuBLAS and cuSPARSE) #7843
- [x] cuSPARSE #7847
- [x] cuFFT #7849
- [x] cuRAND #7856
- [x] cuBLAS #7921
- [x] CUB (depends on Runtime) #7850
- [x] Jitify (depends on NVRTC, Runtime, Driver) #7954
- [x] Runtime #7954
- [x] Profiler (part of Runtime) #7954
- [x] Implement `cupy.cuda.get_local_runtime_version()` to query CUDART version in user environment (#7954)
- [x] Driver #7994
- [x] Improve `cupy.show_config()` to show local runtime version #7995
- [ ] Add CI to make sure `import cupy` succeeds without CUDA Toolkit #7996
```

leofang:
Hi @kmaehashi I'd like to double check one design decision here. @pentschev told me that the current design intends to fail all GPU computation if any of the dlopen (via SoftLink) fails. This is quite surprising to me, because the way we did dlopen to support deferred import and missing libraries should work for a partial CTK. That is, if some DSOs are missing, the function pointers from those DSOs would be `None`, but that's all. The rest of the stack should still work as long as other needed DSOs are there. 
 
For example, if `libcufft.so` is missing, it's natural to expect all `cupy.fft.*` functions stop working, but other functions, say `cupy.matmul()` or `cupy.sin()`, should continue working (the former needs cuBLAS, the latter needs NVRTC). Is this understanding correct?

Asking because there are Jetson users who can only afford a partial CTK installation. Also, conda-forge's new layout naturally supports partial installation, too.

kmaehashi:
> For example, if `libcufft.so` is missing, it's natural to expect all `cupy.fft.*` functions stop working, but other functions, say `cupy.matmul()` or `cupy.sin()`, should continue working (the former needs cuBLAS, the latter needs NVRTC). Is this understanding correct?

[...]

PR #7621, opened 2023-06-09
Soft link NVRTC for `cupy_backends.cuda.libs.nvrtc`

kmaehashi:
Part of #7620.

This PR changes the `cupy_backends.cuda.libs.nvrtc` module to load NVRTC symbols dynamically at runtime, instead of build-time linking.

Pros:
* Make the module buildable & importable without NVRTC installed. (note: Jitify module still links to NVRTC, so this PR is not enough to run `import cupy` without NVRTC)
* No longer need to maintain headers in `cupy_backend`.

Cons:
* When calling NVRTC APIs, maintainers are responsible for passing the correct number/type of arguments, as the compiler can no longer detect such errors.


emcastillo:
This seems to be affecting rocm build somehow?

kmaehashi:
Will resume after #7647

leofang:
Q: This might change the design significantly, but I wonder if it's possible to import the function signatures as is, and reinterpret the fetched symbol using the function signature? That is, instead of 
```cython
     cdef F_t nvrtcCreateProgram = <F_t>_L.get('CreateProgram')
```
we do 
```cython
    ctypedef int (*createProgramFn)(  # naming convention TBD
             Program* prog, const char* src, const char* name, int numHeaders,
             const char** headers, const char** includeNames)
    cdef createProgramFn nvrtcCreateProgram = < createProgramFn>_L.get('CreateProgram')
```
Basically, we'd be trading this pro 
> No longer need to maintain headers in `cupy_backend`.

with this con
> When calling NVRTC APIs, maintainers are responsible for passing the correct number/type of arguments, as the compiler can no longer detect such errors.

The main blocker for this would be the maintainability of the Cython layer. We still have to keep the function signatures...

kmaehashi:
@leofang This is the part I was still wondering... "maintainers are responsible for passing the correct number/type of arguments" might be a bit dangerous as this immediately causes SegV on failure.

I'm now thinking that maintaining `ctypedef`s could be easier if we stop the convention of renaming typedefs (`ctypedef nvrtcProgram Program`). In that case, we can (almost) copy-paste the arguments from NVRTC header:

```c
nvrtcResult nvrtcCreateProgram(nvrtcProgram *prog,
                               const char *src,
                               const char *name,
                               int numHeaders,
                               const char * const *headers,
                               const char * const *includeNames);
```

leofang:
Sorry for some reason I missed your reply, Kenichi 😅 
> if we stop the convention of renaming typedefs

Yeah, sounds reasonable to me. I think at some point my hope was that all these are automated (say, #4395/#4427/#4606), and we can pick whatever style we like, but if it is no less work than this route, allowing easier copy/paste sounds like the right thing to do indeed.

.flake8.cython
```diff
@@ -1,4 +1,4 @@
 [flake8]
 filename = *.pyx, *.pxd, *.pxi
 exclude = .git, .eggs, *.py
-ignore = W503,W504,E225,E226,E227,E275,E402,E999
+ignore = W503,W504,E211,E225,E226,E227,E275,E402,E999
```
kmaehashi, line 4: E211 added to avoid `Whitespace before '('` error:

```
ctypedef nvrtcResult (*F_nvrtcVersion)(int *major, int *minor) nogil
                    ^
```

.pre-commit-config.yaml
```diff
@@ -36,6 +36,7 @@ repos:
     rev: v0.15.0
     hooks:
     -   id: cython-lint
+        args: ["--max-line-length", "79"]
```
kmaehashi, line 39: This is the same value as current coding standard (.flake8.cython)

cupy_backends/cuda/libs/_cnvrtc.pxi
```diff
@@ -0,0 +1,97 @@
...
+ctypedef nvrtcResult (*F_nvrtcCreateProgram)(
+    nvrtcProgram* prog, const char* src, const char* name, int numHeaders,
+    const char** headers, const char** includeNames) nogil
+cdef F_nvrtcCreateProgram nvrtcCreateProgram = <F_nvrtcCreateProgram>_L.get('CreateProgram')  # NOQA
+
+ctypedef nvrtcResult (*F_nvrtcDestroyProgram)(nvrtcProgram *prog) nogil
+cdef F_nvrtcDestroyProgram nvrtcDestroyProgram = <F_nvrtcDestroyProgram>_L.get('DestroyProgram')  # NOQA
+
+ctypedef nvrtcResult (*F_nvrtcCompileProgram)(
+    nvrtcProgram prog, int numOptions, const char** options) nogil
```
kmaehashi, line 58: `ctypedef`s can be generated by copy-pasting the header file.

[`nvrtc.h`](https://gitlab.com/nvidia/headers/cuda-individual/nvrtc/-/blob/88492df6d46cd684bd34df93b307cdc57c9c4dfe/nvrtc.h#L224-225):
```cpp
nvrtcResult nvrtcCompileProgram(nvrtcProgram prog,
                                int numOptions, const char * const *options);
```

cupy_backends/cuda/libs/_cnvrtc.pxi
```diff
@@ -0,0 +1,139 @@
...
+            else:
+                libname = 'nvrtc64_112_0.dll'
+        elif 12000 <= runtime_version < 13000:
+            # CUDA 12.x
+            if _sys.platform == 'linux':
+                libname = 'libnvrtc.so.12'
+            else:
+                libname = 'nvrtc64_120_0.dll'
+    elif CUPY_HIP_VERSION != 0:
+        runtime_version = runtime.runtimeGetVersion()
```
emcastillo, line 130: This needs an actual device when compiling. Is this behavior intended?
ROCm build only CI fails

cupy_backends/cuda/libs/_cnvrtc.pxi
```diff
@@ -0,0 +1,139 @@
...
+
+ctypedef nvrtcResult (*F_nvrtcGetNVVMSize)(nvrtcProgram prog, size_t *nvvmSizeRet) nogil  # NOQA
+cdef F_nvrtcGetNVVMSize nvrtcGetNVVMSize
+
+ctypedef nvrtcResult (*F_nvrtcGetNVVM)(nvrtcProgram prog, char *nvvm) nogil
+cdef F_nvrtcGetNVVM nvrtcGetNVVM
+
+
+cdef SoftLink _L = None
+cdef void initialize():
```
leofang, line 68: Q: Would it help to inline this?

cupy_backends/cuda/libs/_cnvrtc.pxi
kmaehashi: It's failing because ROCm build CI is doing `import cupy` but `runtime.runtimeGetVersion()` won't work without actual device. I'll fix.

cupy_backends/cuda/libs/_cnvrtc.pxi
kmaehashi: Sounds better to inline this.

cupy_backends/cuda/libs/_cnvrtc.pxi
kmaehashi: This should be a problem for CUDA 11.x as well...
https://github.com/cupy/cupy/issues/6688#issuecomment-1383216671

cupy_backends/cuda/libs/_cnvrtc.pxi
kmaehashi: Fixed!

cupy_backends/cuda/libs/_cnvrtc.pxi
kmaehashi: This should be fixed.

leofang:
LGTM!

kmaehashi:
Test passed!

leofang:
> Sorry for some reason I missed your reply, Kenichi 😅
> 
> > if we stop the convention of renaming typedefs
> 
> Yeah, sounds reasonable to me. I think at some point my hope was that all these are automated (say, #4395/#4427/#4606/#5876), and we can pick whatever style we like, but if it is no less work than this route, allowing easier copy/paste sounds like the right thing to do indeed.

PR #7622, opened 2023-06-10
Avoid overflow warnings in test_astype_strides

asi1024:


takagi (approved):
I confirmed `test_astype_strides` failures are gone. LGTM!

Issue #7544, opened 2023-05-09
TestChoiceChi::test_goodness_of_fit sensitive to data randomness


pnunna93:
### Description

The TestChoiceChi::test_goodness_of_fit (https://github.com/cupy/cupy/blob/main/tests/cupy_tests/random_tests/test_generator.py#L830) is very sensitive to data randomness, but is failing on ROCm and passing on CUDA with its current setup. 

The test is using numpy.random to generate seeds(https://github.com/cupy/cupy/blob/main/cupy/testing/_random.py#L90), which are set into cupy random generator. The cupy random generator then generates random numbers for the test. 

The seeds for cupy random generator are same on both platforms as numpy.random is hardcoded to a seed 100 (https://github.com/cupy/cupy/blob/main/cupy/testing/_random.py#L37), but the random numbers generated by cupy random generator are different on ROCm and CUDA which caused failure only on ROCm. 

The test failed on both platforms when run more than 10 times as shown below, please propose modifications to this test to make this more robust, to test the correct functionality, but insensitive to random seed.

| #   of runs | # of   failures (V100) |  # of failures (MI250X) | %Failed (V100) | %Failed (MI250X) |
|:-----------:|:---------------:|:----------------:|:-------:|:------:|
|     10     |        0        |        1        |    0    |    10   |
|     100     |        6        |         8        |    6    |    8   |
|     1000    |        43       |        40        |   4.3   |    4   |
|    10000    |       500       |        494       |    5    |  4.94  |
|    100000   |       4819      |       4880       |  4.819  |  4.88  |

### To Reproduce

Install cupy from source on relevant platform: https://docs.cupy.dev/en/stable/install.html#

Change test iterations to desired number here: https://github.com/cupy/cupy/blob/main/tests/cupy_tests/random_tests/test_generator.py#L829
For 100 iterations, @_condition.repeat(3, 10) -> @_condition.repeat(3, 100)

Replace https://github.com/cupy/cupy/blob/main/tests/cupy_tests/random_tests/test_generator.py#L836 with following two lines to skip failures and continue test for desired iterations:
import pytest_check as check
check.is_true(_hypothesis.chi_square_test(counts, expected))

Command to run the test:
CUPY_TEST_GPU_LIMIT=4  pytest -s -vvv tests/cupy_tests/random_tests/test_generator.py::TestChoiceChi::test_goodness_of_fit

### Installation

Built from GitHub source

### Environment

**MI250X(ROCm):**
[...]

PR #7626, opened 2023-06-12
Change test pass condition for TestChoiceChi::test_goodness_of_fit

pnunna93:
TestChoiceChi::test_goodness_of_fit(https://github.com/cupy/cupy/blob/main/tests/cupy_tests/random_tests/test_generator.py#L830) is sensitive to data randomness and fails on ROCm with its current setup. The test generates random numbers using cupy random generator, which is seeded with random numbers from numpy.random(https://github.com/cupy/cupy/blob/main/cupy/testing/_random.py#L90). 
The numpy.random is hardcoded to a seed of 100(https://github.com/cupy/cupy/blob/main/cupy/testing/_random.py#L37). 
The test runs for 10 test cases following the seed, but one of them fails on ROCm, which is fine based on significance level used for the test(more details in https://github.com/cupy/cupy/issues/7544). This PR changes test passing condition to let it pass on ROCm as mentioned here - https://github.com/cupy/cupy/issues/7544#issuecomment-1566616536.

### DIFF
--- a/tests/cupy_tests/random_tests/test_generator.py
+++ b/tests/cupy_tests/random_tests/test_generator.py
@@ -826,7 +826,7 @@ class TestChoiceChi(RandomGeneratorTestCase):
 
     target_method = 'choice'
 
-    @_condition.repeat(3, 10)
+    @_condition.repeat_with_success_at_least(10, 9)
     def test_goodness_of_fit(self):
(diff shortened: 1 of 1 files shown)


takagi:
Thanks, @pnunna93!

PR #7591, opened 2023-05-25
add iirfilter and related filter design functions

ev-br:
Add `iirfilter` and its matlab-styled aliases. While at it, add converters from zero-poles-gain format to transfer function and SOS formats:

- [x] iirfilter
- [x] zpk2tf
- [x] zpk2sos
.
- [x] butter
- [x] cheby1
- [x] cheby2
- [x] ellip
- [ ] bessel

What's missing / suboptimal:
- bessel filters. This requires finding roots of certain Bessel functions, will port separately.
- zpk2sos loses precision for higher order filters. This happens on CUDA for lower orders than than in NumPy, can be traced to `xp.convolve` based multiplication of polynomials. Will investigate separately.

<s>This PR is based off gh-7553, will rebase once that one is ready. For now, only commits starting from https://github.com/cupy/cupy/pull/7591/commits/fe2dfe4985ea3885f274f00d7a47eeb2af88cd51 are revelant.</s> rebased

cross-ref https://github.com/cupy/cupy/issues/7403

PR #7627, opened 2023-06-13
cupyx/signal: skip tests against older SciPy versions, adjust test tolerances

ev-br:
Older CUDA versions seem to have precision issues, had to make test tolerances *very* loose. So instead tighten tolerances, conditionally skip tests on older CUDA versions.

As suggested in https://github.com/cupy/cupy/pull/7591#issuecomment-1586561815

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_filter_design.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_filter_design.py
@@ -441,7 +441,7 @@ class TestFreqz_zpk:
             assert_array_almost_equal(h, [1])
 
 
-@testing.with_requires("scipy")
+@testing.with_requires("scipy>=1.8")
 class TestSOSFreqz:
 
     @testing.numpy_cupy_allclose(scipy_name='scp')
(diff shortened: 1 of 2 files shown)


takagi:
I found that some older SciPy returns different results from newer, so these failures may not be because of CUDA versions. For example, `TestIIRFilter::test_symmetry` passes with `atol=1e-5, rtol=1e-6` on `scipy==1.7`, but fails on `scipy==1.6`.

We test some combinations of CUDA versions and scipy (and other dependencies) versions as you can see in https://github.com/cupy/cupy/blob/main/.pfnci/coverage.rst. 
- cupy.linux.cuda113 -> scipy 1.6
- cupy.linux.cuda114 -> scipy 1.7
- cupy.linux.cuda115 -> scipy 1.8

takagi:
Would you try requiring newer scipy as `@testing.numpy_cupy_allclose(scipy_name='scp>=1.8', ...)` instead of skipping older CUDA versions? Then, I run the full test again.

ev-br:
Done!

takagi:
Let me give it some fixes, and kick the full test again.

takagi:
`pkg_resources` issue is to be resolved in #7653.

takagi:
CI failures are not related to this PR.

takagi:
Thanks, @ev-br! I slightly changed the title, and sorry for making your branch dirty.

ev-br:
Thank you @takagi!

PR #7629, opened 2023-06-14
Fix signal tests to get align with newer SciPy

takagi:
Older scipy.signal may return different results from newer.
https://ci.preferred.jp/cupy.linux.cuda112/133635/ (scipy 1.7)
https://ci.preferred.jp/cupy.linux.cuda113/133625/ (scipy 1.6)
https://ci.preferred.jp/cupy.linux.cuda114/133628/ (scipy 1.7)


takagi:
Being solved in #7627.

PR #7635, opened 2023-06-15
add kaiser{ord,_beta,_atten} functions

ev-br:

Add 

- [x] kaiserord
- [x] kaser_beta
- [x] kaiser_atten

<s>on top of gh-7537</s>

### DIFF
--- a/cupyx/scipy/signal/_fir_filter_design.py
+++ b/cupyx/scipy/signal/_fir_filter_design.py
@@ -10,6 +10,121 @@ import cupy
 __all__ = ["firls", "minimum_phase"]
 
 
+def kaiser_beta(a):
+    """Compute the Kaiser parameter `beta`, given the attenuation `a`.
+
+    Parameters
+    ----------
+    a : float
+        The desired attenuation in the stopband and maximum ripple in
+        the passband, in dB.  This should be a *positive* number.
+
+    Returns
+    -------
+    beta : float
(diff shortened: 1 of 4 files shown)


cupyx/scipy/signal/_fir_filter_design.py
```diff
@@ -9,6 +9,121 @@
...
+        # Formula for N is not valid in this range.
+        raise ValueError("Requested maximum ripple attentuation %f is too "
+                         "small for the Kaiser formula." % A)
+    beta = kaiser_beta(A)
+
+    # Kaiser's formula (as given in Oppenheim and Schafer) is for the filter
+    # order, so we have to add 1 to get the number of taps.
+    numtaps = (A - 7.95) / 2.285 / (cupy.pi * width) + 1
+
+    return int(cupy.ceil(numtaps)), beta
```
takagi, line 124: Computing this on CPU would be enough.
```suggestion
    return int(numpy.ceil(numtaps)), beta
```

takagi (approved):
LGTM! Waiting for CI finishes.

takagi:
For record: These functions take Python scalars and return them as well, where computing on CPU is enough. An option is depending on SciPy then having aliases to it, but here we're independent of SciPy and have the same ones on CuPy.

takagi (approved):
LGTM!

takagi:
CI failures are not related.

takagi:
Thanks, @ev-br!

PR #7638, opened 2023-06-16
ENH: add `cupyx.scipy.signal.minimum_phase`

ev-br:
cross-ref https://github.com/cupy/cupy/issues/7403

<s>This PR is on top of gh-7537 to minimize git conflicts  going forward.</s>

PR #7641, opened 2023-06-16
ENH: add cupyx.signal.{freqs, freqs_zpk, findfreqs}

ev-br:

As discussed on gitter, this PR delegates to `numpy.roots`  on CPU. The issue is that root-finding is typically done as the eigenvalue problem for the companion matrix, but `eigvals` is only available in MAGMA which is not a CuPy dependency.

<s>PR is on top of https://github.com/cupy/cupy/pull/7638 , will rebase</s> rebased.


ev-br:
Converting to draft as it depends on gh-7638

takagi (approved):
LGTM! CI failures are not related.

takagi:
Thanks, @ev-br!

PR #7623, opened 2023-06-10
Deprecate `cupy.round_`

asi1024:
`numpy.round_` will be deprecated in NumPy 1.25.

PR #7643, opened 2023-06-19
Fix `test_fht` not to feed `cupy.ndarray` to `scipy.fft.fhtoffset`

asi1024:
https://github.com/cupy/cupy/pull/7623


kmaehashi:
Test failures are irrelevant.

PR #7644, opened 2023-06-19
ENH: cupyx/scipy/signal: add unique_roots, invres{z}, residue{z}

ev-br:
`scipy.signal` version uses `scipy.spatial.KDTree`; here we loop over the points manually instead. The inputs are most likely to be small, so there is not much need to seriously optimize performance. 

Also add 
- [x] unique_roots
- [x] invres
- [x] invresz
- [x] residue
- [x] residuez


ev-br:
Apparently, pytest requires test file names to be unique across the project, and `cupyx_tests/scipy_tests/signal_tests/test_polyutils.py` was clashing with `tests/cupy_tests/polynomial_tests/test_polyutils.py`. So renamed the former into `test_polyutils_.py` and rebased on main.

takagi:
Seems that we've got to have `__init__.py` in the same directory with `cupyx_tests/scipy_tests/signal_tests/test_polyutils.py` to make pytest recognize it as part of a package.
https://docs.pytest.org/en/latest/explanation/pythonpath.html

I understand renaming works, but would like to prefer `test_polyutils.py`. Would you make it as such?

ev-br:
OK, done.

takagi (approved):
LGTM! CI failures are not related and is to be handled elsewhere.

takagi:
Thanks, @ev-br!

Issue #7555, opened 2023-05-11
CuPy v13 Release Plan


kmaehashi:
## Roadmap

```[tasklist]
## Planned Features

- [ ] Signal API coverage enhancement (`cupyx.scipy.signal.*`)
- [ ] `complex32` support for limited APIs (restart #4454)
- [ ] Support reduction ops in distrubuted ndarray
- [ ] Allow `import cupy` without CUDA installation (#6908)
- [ ] Migrate from Thurst/CUB to [CCCL](https://github.com/NVIDIA/cccl)
- [ ] Typed CuPy APIs (#4831)
- [ ] Memory allocator interoperability with PyTorch (to be discussed in a separate issue)
- [ ] Drop support for CUDA 10.2, 11.0, and 11.1 (to be discussed in a separate issue)
- [ ] Support Python 3.12 (both in CuPy v12 and v13)
- [ ] Drop support for Python 3.9 following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)
- [ ] Release workflow improvement: GitHub Action-based wheel build (https://github.com/cupy/cupy-release-tools/issues/272#issuecomment-1165093812)
```

```[tasklist]
## Timeline & Release Manager
- [ ] **v13.0.0a1 (v12.1.0)** (May 2023) @takagi
- [ ] **v13.0.0b1 (v12.2.0)** (July 2023) @kmaehashi
- [ ] **v13.0.0rc1 (v12.3.0)** (September 2023) @asi1024
- [ ] **v13.0.0** (October 2023) @emcastillo
```

## Notes

* As of CuPy v13, we have adjusted our release frequency to once every two months. However, mid-term or hot-fix releases may be provided depending on necessity, such as for new CUDA/Python version support or critical bug fixes.
* The schedule and planned features may be subject to change depending on the progress of the development.
* `v12.x` releases will only contain backported pull-requests (mainly bug-fixes) from the v13 (`main`) branch.

## Past Release Plans

* v9: https://github.com/cupy/cupy/issues/3891
* v10: https://github.com/cupy/cupy/issues/5049
* v11: https://github.com/cupy/cupy/issues/6246
* v12: https://github.com/cupy/cupy/issues/6866

leofang:
> Drop support for Python 3.9 following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)

I think you meant 3.8, not 3.9? 🙂

kmaehashi:
[...]

Issue #7127, opened 2022-10-17
CI: Add tests for RHEL 9 & Ubuntu 22.04


kmaehashi:
Let's add them to the CI as they are now both supported in CUDA.

PR #7647, opened 2023-06-19
Deprecate support for out-dated platforms in CuPy v13

### LINKED ISSUE #6805: Change rapidsai CI image to stable once pylibraft support is released
Due to #6690 changing the `rapidsai/rapidsai-core-dev-nightly` image to nightly, we should pin it to a stable one once all the required features are released.

### LINKED ISSUE #7557: [RFC] Drop support for CUDA 10.2/11.0/11.1 in CuPy v13

kmaehashi:
This PR drops support for:

* CUDA 11.1 or earlier (closes #7557)
* Python 3.8 (#7555)
* NumPy 1.21 (NEP 29)
* cuTENSOR 1.5 or earlier
* NCCL 2.15 or earlier
* cuDNN 8.8 or earlier
* Ubuntu 18.04 (#7127)

Also closes #6805


kmaehashi:
Updated `/ test` tags (thansks @asi1024 for pointing this out!)
`mini` now triggers CI for CUDA {11.2, 11.8, 12.0, 12.1} x {Linux, Windows}.

.pfnci/linux/tests/cuda-rapids.Dockerfile
```diff
@@ -1,5 +1,4 @@
-# TODO(kmaehashi): Fix stable image after release (#6805)
-ARG BASE_IMAGE="rapidsai/rapidsai-core-dev-nightly:22.08-cuda11.5-devel-ubuntu20.04-py3.8"
+ARG BASE_IMAGE="rapidsai/rapidsai-core-dev:23.06-cuda11.8-devel-ubuntu22.04-py3.10"
```
emcastillo, line 1: I think that this image is not working well :(

emcastillo: https://github.com/rapidsai/cugraph/pull/2866

This PR sseems to have changed the API of `weakly_connected_components`

kmaehashi, line 1: Thanks! Let's pin to the old one and fix that in a separate pull-request.

```suggestion
ARG BASE_IMAGE="rapidsai/rapidsai-core-dev:22.10-cuda11.5-devel-ubuntu20.04-py3.9"
```

PR #7649, opened 2023-06-19
Use GitHub App to update org-wide project

kmaehashi:
Another follow-up for https://github.com/cupy/cupy/pull/7631
Things were more complicated than I expected... 

Let me self-merge.
cc/ @asi1024

### DIFF
--- a/.github/workflows/pr-project.yml
+++ b/.github/workflows/pr-project.yml
@@ -12,6 +12,32 @@ jobs:
     if: github.repository_owner == 'cupy'
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


kmaehashi:
Ok, it's now working.

I used https://github.com/cupy/cupy/pull/7645 for testing.

1. I set the status to `Awaiting Author`
2. The first workflow triggered: https://github.com/cupy/cupy/actions/runs/5314648846/jobs/9622092581
3. The second workflow triggered: https://github.com/cupy/cupy/actions/runs/5314650375/jobs/9622095578
4. THe status is changed to `Needs Attention`

PR #7650, opened 2023-06-19
Do not trigger on closed pull-requests and fix to work with `issue_comment` trigger

kmaehashi:
Follow-up https://github.com/cupy/cupy/pull/7631


emcastillo:
Merging as it doesn't require CIs

emcastillo (approved):

PR #7652, opened 2023-06-20
ENH: cupyx/scipy/signal: add missing LTI format conversions

ev-br:
cross-ref gh-7403

- [x] tf2zpk
- [x] tf2sos
- [x] tf2ss
- [x] ss2tf
- [x] zpk2ss
- [x] ss2zpk
- [x] sos2tf
- [x] sos2zpk
- [x] BadCoefficients

Note that there are two utilities which delegate to NumPy on CPU, `_polyutils.roots` and `.poly`. Both rely on `np.linalg.eigvals`, which on GPU is only available from MAGMA.

<s>this is on top of gh-7644</s>

PR #7660, opened 2023-06-21
ENH: add LTI class hierarchy and lti/dlti related functions

ev-br:

- [x] StateSpace
- [x] TransferFunction
- [x] ZerosPolesGain
.
- [x] dlti
- [x] dlsim
- [x] dimpulse
- [x] dstep
- [x] dbode
- [x] dfreqresp
.
- [x] lti
- [x] lsim
- [x] impulse
- [x] step 
- [x] freqresp
- [x] bode
.
- [x] cont2discrete

Also add the matrix exponential, `cupyx.scipy.linalg.expm` (needed for lsim, cont2discrete).

### DIFF
--- a/cupyx/scipy/signal/__init__.py
+++ b/cupyx/scipy/signal/__init__.py
@@ -107,6 +107,8 @@ from cupyx.scipy.signal._peak_finding import peak_prominences  # NOQA
 from cupyx.scipy.signal._peak_finding import peak_widths  # NOQA
 
 from cupyx.scipy.signal._ltisys import lti  # NOQA
+from cupyx.scipy.signal._ltisys import freqresp  # NOQA
+
 from cupyx.scipy.signal._ltisys import dlti  # NOQA
 from cupyx.scipy.signal._ltisys import dlsim  # NOQA
 from cupyx.scipy.signal._ltisys import dstep  # NOQA
--- a/cupyx/scipy/signal/__init__.py
+++ b/cupyx/scipy/signal/__init__.py
@@ -108,6 +108,7 @@ from cupyx.scipy.signal._peak_finding import peak_widths  # NOQA
 
 from cupyx.scipy.signal._ltisys import lti  # NOQA
 from cupyx.scipy.signal._ltisys import lsim  # NOQA
+from cupyx.scipy.signal._ltisys import impulse  # NOQA
 from cupyx.scipy.signal._ltisys import freqresp  # NOQA
 from cupyx.scipy.signal._ltisys import bode  # NOQA
 
--- a/cupyx/scipy/signal/__init__.py
+++ b/cupyx/scipy/signal/__init__.py
@@ -105,3 +105,12 @@ from cupyx.scipy.signal._lti_conversion import abcd_normalize   # NOQA
 from cupyx.scipy.signal._peak_finding import find_peaks  # NOQA
 from cupyx.scipy.signal._peak_finding import peak_prominences  # NOQA
 from cupyx.scipy.signal._peak_finding import peak_widths  # NOQA
+
+from cupyx.scipy.signal._ltisys import dlsim
+from cupyx.scipy.signal._ltisys import dstep
+from cupyx.scipy.signal._ltisys import dimpulse
+from cupyx.scipy.signal._ltisys import dbode
+from cupyx.scipy.signal._ltisys import dfreqresp
+from cupyx.scipy.signal._ltisys import StateSpace
+from cupyx.scipy.signal._ltisys import TransferFunction
+from cupyx.scipy.signal._ltisys import ZeroPolesGain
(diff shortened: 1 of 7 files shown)


ev-br:
DLTI hierarchy and related functions are done here. The PR is still in a draft status, for two reasons:

- it depends on gh-7652, which brings low-level conversions between formats
- static checkers complain about missing continuous LTI functions. These two, continuous and discrete LTI hierarchies, are sufficiently intertwined, so probably best to add both in the same PR. However continuous LTI evolution, `lsim`, relies on a matrix exponential, so this would require porting `scipy.linalg.expm` to CuPy.

takagi:
We have a draft expm PR #5688, but don't hear from its author for so long, though.

ev-br:
Update: 

- Added `expm`. For the moment, only implement the Higham'2005 algorithm [1], in a simplified form:
     -  [13/13] Pade approximant only (no selection of 3-5-7-9 orders)
     - no matrix balancing (this requires LAPACK's `*gebal` which is not available on CUDA): balancing could improve numerical stability.
     - the implementation is pure python + matmuls.

- Added `lsim`, which is essentially `expm(t*A)` for `t` a linspace array. Like in scipy.signal, the computation is a python loop with a repeated application of `expm(dt*A)`. While performance is not terrible, it can probably be improved by using an Al-Mohy & Higham 2010 algorithm for a `t` array [2]. Keeping it simple for now.

- Added `impulse` and `step`, both are straighforward applications of `lsim`. Their APIs are annoyingly polymorphic (from 2000s), am keeping them compatible with scipy.signal. One annoying issue is that both use `eigvals` to estimate a default time span given the system matrix. Am delegating to `np.linalg.eigvals`. An alternative is to estimate the boundaries for the eigenvalues using Gershgorin theorems, but that would make the default behavior incompatible with scipy.signal.

[1] https://epubs.siam.org/doi/10.1137/04061101X
[2] https://eprints.maths.manchester.ac.uk/1300/1/

ev-br:
`<rant mode on>`  Have to admit I'm baffled by mypy complaining about ways to format module-level lists. What on earth is wrong with 

```
b = [0] * 14
b[0] = 1
b[1] = 2
...
```
`<rant mode off>`

Anyway, this is fixed in https://github.com/cupy/cupy/pull/7660/commits/c4004004f7d1300cde3e0ed5bc2c0c72ee82b052.

ev-br:
OK, this now implements both continuous and discrete LTI functionality. Is ready from my side, modulo the need to rebase when   gh-7652 gets in.

ev-br:
Rebased on main, so this PR is ready for review from my side.

cupyx/scipy/linalg/_matfuncs.py
```diff
@@ -35,3 +37,95 @@ def khatri_rao(a, b):
...
+    n = a.shape[0]
+
+    # try reducing the norm
+    mu = cupy.diag(a).sum() / n
+    A = a - cupy.eye(n)*mu
+
+    # scale factor
+    nrmA = cupy.linalg.norm(A, ord=1)
+
+    scale = nrmA > th13
```
takagi, line 99: Transferring the data of `nrmA` to CPU first would reduce the number of sync. I suppose, ideally, we could eliminate the sync to do these scaling and squaring later with keeping `s` on GPU, but it would include troublesome cuBLAS device API calls and HIP compatibility issues. It would be too much here.

takagi: Also, can I ask you to have `expm` in another PR to separate from this one?

ev-br: Would it help to keep `b` and scalars s and th13 on the device?

takagi: I suppose we need `s` on the host to the squaring:
```py
for _ in range(s):
    x = x @ x
```
We could do this repetitive matrix multiplication if we have `s` on the device, but it would require cuBLAS device API calls in a RawKernel.

ev-br: I see. Just confirm my understanding: changing the matmul into `cupy.matmul(x, x, out=x) ` also incurs synchronizarion, correct?

takagi: Both of the matmul and `cupy.matmul` do not have synchronization. In the current code, `if scale` where `scale` is evaluated as if `bool(scale)` and `float(nrmA)` have.

ev-br:
> have expm in another PR to separate from this one?

Done in https://github.com/cupy/cupy/pull/7744.

Meanwhile I hopefully fixed the CI errors in this PR, and will rebase this PR to remove expm-related commits when gh-7744 is decided on.

takagi:
> Done in https://github.com/cupy/cupy/pull/7744.

Thanks!

takagi (approved):
LGTM! I'll kick the CI again and then merge this once #7744 is merged.

ev-br:
Rebased on main.

ev-br:
Pushed an update to relax the contiguity checks which were failing on CI for some configurations.

takagi (approved):

takagi:
Thanks, @ev-br!

PR #7663, opened 2023-06-25
Add mixed precision (FP16) support for ROCm

jglaser:
This PR adds mixed precision support for matrix-matrix multiplies on ROCm, which is available in hipBLAS through `hipblasGemmEx`. On MI200 GPUs, this utilizes the MFMA (matrix fused-multiply add) instructions, which are intended as analogues of the NVIDIA tensor cores.

For `cupy.float16` arrays, we allow the use of the hipBLAS primitive, which was previously disabled, and fp16 values were casted to single precision, albeit issuing a warning. I am not entirely sure if the reason for this was the lack of hardware and/or API support in earlier versions of the AMD libraries, therefore a version check for hipBLAS might be required. It appears though that support has been stable since ROCm 4.2.0 (https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/CHANGELOG.md).

To verify that the code now uses `rocblas_gemm_ex` under the hood, enable rocBLAS runtime tracing with the unit test like this
``` 
ROCBLAS_LAYER=1 pytest tests/cupy_tests/math_tests/test_matmul.py -s -k TestMatmulOut
```

### DIFF
--- a/cupy/_core/_routines_linalg.pyx
+++ b/cupy/_core/_routines_linalg.pyx
@@ -646,7 +646,8 @@ cpdef _ndarray_base tensordot_core(
     if dtype == 'e':
         use_tensor_core = (not runtime._is_hip_environment and
                            _cuda_runtime_version >= 9000 and
-                           compute_capability >= 70) or (runtime._is_hip_environment)
+                           compute_capability >= 70
+                           ) or runtime._is_hip_environment
         if use_tensor_core:
             can_opt_in_tensorcore = not runtime._is_hip_environment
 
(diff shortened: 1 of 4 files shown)


takagi:
Thanks, @jglaser! Your change looks good to me. Can I ask you to fix just static-checks?

takagi:
> ROCBLAS_LAYER=1 pytest tests/cupy_tests/math_tests/test_matmul.py -s -k TestMatmulOut

I also checked `hipblasGemmEx` was called locally.

jglaser:
Looks like two CUDA related tests are still failing, can't see the test results without logging in

takagi (approved):
LGTM!

takagi:
CI failures are not related. Our CI requires logging in for security reasons.

takagi:
Thanks, @jglaser!

PR #7666, opened 2023-06-26
cupyx/scipy/signal: add place poles

ev-br:
cross-ref gh-7403

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_ltisys.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_ltisys.py
@@ -2,8 +2,9 @@ import cupy
 from cupyx.scipy.signal import abcd_normalize
 import cupyx.scipy.signal as signal
 
-from cupy import testing
+import pytest
 from pytest import raises as assert_raises
+from cupy import testing
 
 
 def assert_equal(actual, desired):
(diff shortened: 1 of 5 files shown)


tests/cupyx_tests/scipy_tests/signal_tests/test_ltisys.py
```diff
@@ -540,3 +541,318 @@ def test_second_order(self, xp, scp):
...
+             -491, -1543, -686]
+        ).reshape(6, 5)
+        P = xp.array([-25.-29.j, -25.+29.j, 31.-42.j,
+                     31.+42.j, 33.-41.j, 33.+41.j])
+
+        fsf = scp.signal.place_poles(A, B, P)
+        return fsf.computed_poles
+
+    @pytest.mark.xfail(reason="numerical stability: scipy QR vs numpy QR")
+    @testing.numpy_cupy_allclose(scipy_name='scp')
```
takagi, line 643: Does relaxing the tolerance (`rtol` and `atol` of `numpy_cupy_allclose`) remove the xfail?

ev-br:
Updated with skips of tests which seem brittle on CI (and locally). Hopefully skipping based on the scipy version would be enough.

tests/cupyx_tests/scipy_tests/signal_tests/test_ltisys.py
ev-br: Not really. This test is flaky, and failure modes depend on the scipy version. Comments in the scipy code indicate things start differing between scipy.linalg.qr and numpy.qr (?), so this whole function is a bit brittle.
So the last commit skips the test depending on the scipy version, based on what the CI seems to suggest.

ev-br:
Rebased on main and tightened up a test which was failing on CI for no good reason.

takagi:
Would you try to make it contiguous then sort?
```
E   AssertionError: Only cupy raises error
E   
E     File "/root/.local/lib/python3.9/site-packages/cupy/testing/_loops.py", line 45, in _call_func
E       result = impl(*args, **kw)
E     File "/tmp/tmp.YnnJbxVJui/tests/cupyx_tests/scipy_tests/signal_tests/test_ltisys.py", line 585, in test_real_2
E       poles.sort()
E     File "cupy/_core/core.pyx", line 854, in cupy._core.core._ndarray_base.sort
E     File "cupy/_core/core.pyx", line 872, in cupy._core.core._ndarray_base.sort
E     File "cupy/_core/_routines_sorting.pyx", line 33, in cupy._core._routines_sorting._ndarray_sort
E   NotImplementedError: Sorting non-contiguous array is not supported.
```

takagi:
Ah, seems that scipy returns results with different dtype on Windows. `cupy.linux.cuda122`'s failure is not releated, so it's okay.

ev-br:
> seems that scipy returns results with different dtype on Windows.

Grrrr. Given that the test uses `np.real_if_close`,  a minimal guess is that on windows the fp noise exceeds the default tolerance of `100 * eps`.

It's too bad that the CI logs do not  show the arrays themselves or max difference. I wonder what's the easiest way to check these values (I can push a commit with a print of course but that would need somebody to trigger the CI run).

https://github.com/cupy/cupy/pull/7666/files#diff-e20ce11193646698fc4b7ae40e1bfd240f663a2051c4e0c5520b05d6e045a4d3R584

ev-br:
We checked with @andfoy 's local windows setup (thank you Edgar!) and we cannot reproduce the dtype failure. On his local system with the matching scipy version (scipy 1.11.3), `test_real_2` both CuPy and SciPy only produces real poles:
https://gist.github.com/andfoy/5ff15c9df67c2f8d392fe03574443727.

So we cannot reproduce the CI failure and frankly I'm tempted to just xfail the test.

takagi:
Please mark xfail the tests, thanks for your effort!

takagi (approved):
LGTM!

takagi:
The CI failure is not related.

takagi:
Thanks @ev-br!

Issue #7662, opened 2023-06-23
Synchronisation NumPy View and CuPy


Electricks94:
### Description

I create a numpy array and then a view on the array with swapped axis. I have assumed, that cupy recognizes that my object is a view on data, copies the data, and creates a view with the same properties as my CPU array view.

Even more confusing is that I can create the cupy array. Swap the axis similarly to create a view on my array but then I can not use `set` or `get` anymore due to `RuntimeError: Cannot set to non-contiguous array`

### To Reproduce

```py
import cupy as cp
import numpy as np

cpu_array = np.empty((10, 20, 30), order='c')
cpu_array = cpu_array.swapaxes(1, 0)
gpu_array = cp.asarray(cpu_array)

assert gpu_array.strides == cpu_array.strides # stride missmatch

# Create new array and adapt strides manually
gpu_array = cp.asarray(cpu_array.base, order="C")
gpu_array = gpu_array.swapaxes(1, 0)
gpu_array.set(cpu_array) # Can not copy data anymore. --> non-contiguous array
```

### Installation

Wheel (`pip install cupy-***`)

### Environment

OS                           : Linux-3.10.0-1160.el7.x86_64-x86_64-with-glibc2.17
Python Version               : 3.9.5
CuPy Version                 : 12.1.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 1.21.0
SciPy Version                : 1.7.0
Cython Build Version         : 0.29.32
Cython Runtime Version       : 0.29.33
CUDA Root                    : /softs/nvidia/cuda-11.2
nvcc PATH                    : /softs/nvidia/cuda-11.2/bin/nvcc
CUDA Build Version           : 11080
CUDA Driver Version          : 12000
CUDA Runtime Version         : 11020
cuBLAS Version               : (available)
cuFFT Version                : 10401
cuRAND Version               : 10203
cuSOLVER Version             : (11, 1, 0)
cuSPARSE Version             : (available)
NVRTC Version                : (11, 2)
Thrust Version               : 101501
CUB Build Version            : 101501
Jitify Build Version         : 4a37de0
cuDNN Build Version          : 8801
cuDNN Version                : 8101
NCCL Build Version           : None
NCCL Runtime Version         : None
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA A30
Device 0 Compute Capability  : 80
Device 0 PCI Bus ID          : 0000:17:00.0
[...]

PR #1845, opened 2018-11-29
support order kwarg in asarray, asanyarray, ndarray.get, tonumpy

grlee77:
This is a small PR that continues the quest to improve `order` kwarg support in CuPy (see also #1819 and #1843).

One outstanding question is what to do for non-cupy objects that define `__cuda_array_interface__`. For now, I chose to ignore the `order` argument in that case.

PR #7668, opened 2023-06-28
Fix docstring of `asarray`

kmaehashi:
`cupy.asarray` supports order argument since #1845 but docstring was not updated.
Discovered during looking into #7662.

The default value was also wrong. `asarray` internally uses `array(a, copy=False)` which does not copy if the input is `cupy.ndarray`.

```pycon
>>> z = cupy.ones((10,10,10), order='F')
>>> cupy.asarray(z) is z
True
```


emcastillo (approved):

emcastillo:
Failures unrelated

PR #7669, opened 2023-06-28
Skip `TestLOBPCG::test_maxit_None` in CUDA 12.1.1 & cuSOVLER 11.4.5

asi1024:
Locally confirmed that the test passes with CUDA 12.1.0, and fails with CUDA 12.1.1.


asi1024:
cc: @leofang @anaruse

leofang:
Do we know what cuSOLVER routine is involved here? What's the error?

asi1024:
@leofang I haven't investigated enough to conclude the reason for the test failure, but I suspect that the test failure seems from a different result of `cupyx.scipy.linalg.lobpcg` from the error log. The failed CI log is here: https://ci.preferred.jp/cupy.linux.cuda121/134677/. I updated also the PR description!

PR #7671, opened 2023-06-28
Add KDTree to cupyx.scipy.spatial

andfoy:
This PR aims to implement a basic version of a left-balanced, dimension alternating KD-Tree based on https://arxiv.org/abs/2211.00120

KDTree is needed by some interpolation and signal APIs to generate an spatial index. This implementation relies on local, constant computations that can be obtained in a parallel fashion, it also employs `lexsort`, which is already parallel.


cupyx/scipy/spatial/_kdtree.py
```diff
@@ -0,0 +1,273 @@
...
+        but only columns that store the requested values of k are preserved.
+        This is implemented in a manner that reduces memory usage.
+
+        Examples
+        --------
+
+        >>> import cupy as cp
+        >>> from cupyx.scipy.spatial import KDTree
+        >>> x, y = cp.mgrid[0:5, 2:8]
+        >>> tree = KDTree(np.c_[x.ravel(), y.ravel()])
```
takagi, line 205:
```suggestion
        >>> tree = KDTree(cp.c_[x.ravel(), y.ravel()])
```

cupyx/scipy/spatial/_kdtree.py
```diff
@@ -0,0 +1,273 @@
...
+
+        >>> dd, ii = tree.query([[0, 0], [2.2, 2.9]], k=[1, 2])
+        >>> print(dd, ii, sep='\n')
+        [[2.         2.23606798]
+         [0.2236068  0.80622577]]
+        [[ 0  6]
+         [13 19]]
+        """
+        if self.copy_query_points:
+            if cupy.dtype(x.dtype) is not cupy.dtype(cupy.float64):
```
takagi, line 252: nit: `==` comprehensively compares dtypes.
```suggestion
            if x.dtype == cupy.float64:
```

leofang:
cc: @anaruse @cjnolet for vis

andfoy:
These are the benchmark results when a `KDTree` is built with 2D elements. Right now the main bottleneck is `cupy.lexsort`, which takes more than 95% of the computation time. 

<details>

```python
from itertools import product

import cupy as cp
import numpy as np
from cupy import testing

from cupyx.scipy.spatial import KDTree as CuKDTree
from scipy.spatial import KDTree as CPUKDTree

from tqdm import tqdm


input_sizes = [5, 10, 100, 1000, 5000, 10000, 50000, 100000, 200000, 500000,
               700000, 1_000_000] # 1_000_000, 2_000_000, 10_000_000]
modules = [(cp, CuKDTree, 'CuPy'), (np, CPUKDTree, 'SciPy')]
kwargs = [('ndims', [2])] #[1, 2, 4, 6, 10])]
measurement = {}


def gather_time(prof):


def input_creation(xp, fn, size, ndims=1, **kwargs):
    x = testing.shaped_random(size + (ndims,), xp)
    # size = size[0]
    # signs = xp.tile(xp.asarray([1, -1], dtype=xp.float32),
    #                 int(2 * np.ceil(size / 4)))
    # x = xp.arange(3, size + 3, dtype=xp.float32) * signs[:size]
    # sos = testing.shaped_random((sections, 6), xp, dtype=xp.float32, scale=0.2)
    # sos[:, 3] = 1
    return fn, (x,)


def closure(fn, sig):
    return fn(*sig)


for kwarg_comb in kwargs:
    kwarg_comb = dict(kwarg_comb)
        for xp, cls, _id in modules:


lines = []
for kwarg_comb in kwargs:
    lines.append('| Size | CuPy (ms) | SciPy (ms) | Speedup |')
    lines.append('|:----:|:---------:|:----------:|:-------:|')
    for size in input_sizes:
        comp = f'{size}'
        times = []
        for _, _, _id in modules:
            time = size_measurement[_id]
            times.append(time)
            time_info = f'{time:3f}'
        speedup = times[1] / times[0]
        lines.append(f'| {comp} |')
    lines.append('\n')


print('\n'.join(lines))

```
</details>

### `ndims=2`

| Size | CuPy (ms) | SciPy (ms) | Speedup |
|:----:|:---------:|:----------:|:-------:|
| 5 | 0.818687 | 0.031569 | 0.038561 |
| 10 | 1.133810 | 0.031387 | 0.027682 |
| 100 | 1.933604 | 0.042085 | 0.021765 |
| 1000 | 3.041713 | 0.218105 | 0.071705 |
| 5000 | 5.039459 | 1.072039 | 0.212729 |
| 10000 | 6.195290 | 2.248290 | 0.362903 |
| 50000 | 9.510284 | 14.774096 | 1.553486 |
| 100000 | 17.683976 | 32.306924 | 1.826904 |
| 200000 | 37.212859 | 69.179654 | 1.859026 |
| 500000 | 105.139692 | 184.356776 | 1.753446 |
| 700000 | 174.812453 | 259.726934 | 1.485746 |
| 1000000 | 271.937696 | 380.062228 | 1.397608 |

## Profiler information

imagen

takagi:
Would you add KDtree to the rst file?

takagi:
Does the paper also use lexsort? (The author has a public source code?)

andfoy:
> Does the paper also use lexsort? (The author has a public source code?)

It seems that they now provide a bitonic sort implementation. https://github.com/ingowald/cudaKDTree/blob/master/cukd/builder_bitonic.h, the original reference did use a mix of `thrust::sort` calls

takagi:
`cupy.lexsort` implementation may not be optimal. Is it possible to try calling `thrust::sort` in a raw kernel with a custom comparer as cudaKDTree does?

cupyx/scipy/spatial/_kdtree_utils.py
```diff
@@ -0,0 +1,929 @@
...
+    level = 0
+    for level in range(n_iter):
+        dim = level % dims
+        x_tags[0, :] = x[:, dim]
+        x_tags[1, :] = tags
+        idx = cupy.lexsort(x_tags)
+        x = x[idx]
+        tags = tags[idx]
+        track_idx = track_idx[idx]
+        update_tags((n_blocks,), (block_sz,), (length, level, tags))
```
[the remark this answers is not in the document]
takagi: I can find `thrust::sort` declarations with `__device__` specified. I wonder if it could be called in `RawKernel`, but I have never tried that for myself. Would you take a look at these?
https://thrust.github.io/doc/group__sorting_ga3f47925d80f4970d5730051dba1c5603.html
https://thrust.github.io/doc/group__sorting_gaef9372b75a6920c67c43150627e74225.html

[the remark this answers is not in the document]
takagi: If it doesn't work, let us keep using `lexsort` here. We wouldn't like to have dynamic parallelism this part.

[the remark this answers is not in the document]
andfoy: > I can find `thrust::sort` declarations with `__device__` specified. I wonder if it could be called in `RawKernel`, but I have never tried that for myself. 

According to some sources in other forums, it seems that a device kernel can call `thrust::sort` under the following conditions:

1. It serializes the execution to that single kernel, by setting `thrust::seq` as policy → No parallel gains in our case
2. It spawns further kernels via dynamic parallelism, by setting `thrust::device` as policy → Parallel gains, but dynamic parallelism is involved.
  
In this case our ideal solution would be to define a kernel that performs the sorting process in a single pass with a custom comparator instead of two, which is the current `lexsort` behaviour. I don't know if we can define a pair sorting function that calls `thrust::sort_by_key`, similar to https://github.com/cupy/cupy/blob/1cd6db5566eab966f7d4cd96bf701dba81e784c5/cupy/cuda/cupy_thrust.cu#L388 in a Cython file or similar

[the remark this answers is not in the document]
takagi: Thanks for the investigation. Let me keep the current implementation using `lexsort` for this moment, and revisit in another issue.

tests/cupyx_tests/scipy_tests/spatial_tests/test_kdtree.py
```diff
@@ -0,0 +1,360 @@
...
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_vectorized_query_multiple_neighbors_random(self, xp, scp):
+        s = 23
+        x, tree = create_random_kd_tree(xp, scp, 100, 3, 8)
+        x = x.reshape(2, 4, 3).copy()
+        kk = tree.n + s
+        d, _ = tree.query(x, k=kk)
+        return d
+
+    def test_query_raises_for_k_none(self):
```
takagi, line 215: Would you run this test only with `scipy>=1.9`?

takagi:
Seems windows CIs are failing because of dtype size (`int32` and `int64`?). Do you have `long` in any kernel of this change?

takagi:
One of my team is interested in trying out kdtree internally. If possible, we would appreciate it if you could prioritize this. Thanks!

andfoy:
> Seems windows CIs are failing because of dtype size (int32 and int64?). Do you have long in any kernel of this change?

I did a very close inspection of every kernel defined, as well as their parameters and I'm afraid I wasn't able to find any references to any of `long`, `cupy.int32`nor `cupy.int_` constants/keywords

cupyx/scipy/spatial/_kdtree_utils.py
```diff
@@ -0,0 +1,918 @@
...
+        const int k, const int n, const int n_dims, const double eps,
+        const double p, const double dist_bound, const bool periodic,
+        const T* __restrict__ point, const T* __restrict__ tree,
+        const long long* __restrict__ index,
+        const double* __restrict__ box_bounds,
+        const T* __restrict__ tree_bounds,
+        double* distances, long long* nodes) {
+
+    volatile long long prev = -1;
+    volatile long long curr = 0;
```
andfoy, line 341: Does `volatile` keyword yields random results in Windows?

takagi: Let me try to remove these volatile keywords.

takagi:
I suspect that the cause of the Windows issue is that numpy/cupy's integer array's default dtype is int32 by default on Windows. For example, we call `cupy.arange` here:
https://github.com/cupy/cupy/blob/f25fe0d049530b417dc4e0ff4f8cd13744425ad2/cupyx/scipy/spatial/_kdtree_utils.py#L753
, and `track_idx` has an int32 array. However, this value is eventually passed to `compute_knn` kernel function and its corresponding argument is `long long* index`, which would cause a problem as we see on the CI results. Would you check such API calls that return an integer array?

rgommers:
> I suspect that the cause of the Windows issue is that numpy/cupy's integer array's dtype is int32 by default on Windows.

As a heads up in case you didn't see this yet: NumPy just changed it in `main`, so for 2.0 it'll map to `int64`: https://github.com/numpy/numpy/pull/24224.

takagi:
Thanks, it will make us happier!

takagi:
We have less failures now. Do we still have other suspicious lines?
https://ci.preferred.jp/cupy.win.cuda122/148918/
```
FAILED cupyx_tests\scipy_tests\spatial_tests\test_kdtree.py::TestPeriodic::test_kdtree_ball[0] - AssertionError: ndarrays of different dtypes are returned.
cupy: int64
numpy: int32
FAILED cupyx_tests\scipy_tests\spatial_tests\test_kdtree.py::TestPeriodic::test_kdtree_ball[1] - AssertionError: ndarrays of different dtypes are returned.
cupy: int64
numpy: int32
FAILED cupyx_tests\scipy_tests\spatial_tests\test_kdtree.py::TestPeriodic::test_kdtree_ball[-1] - AssertionError: ndarrays of different dtypes are returned.
cupy: int64
numpy: int32
FAILED cupyx_tests\scipy_tests\spatial_tests\test_kdtree.py::TestPeriodic::test_kdtree_ball_tree - AssertionError: ndarrays of different dtypes are returned.
cupy: int64
numpy: int32
= 4 failed, 107392 passed, 6791 skipped, 676 deselected, 320 xfailed, 2477 warnings in 7285.52s (2:01:25) =
```

takagi:
All windows tests passed! (`cupy.win.cuda114`'s failure status can be ignored) I'm going to kick the full test.

takagi:
`cupy.linux.cuda122`'s failure is not related. `cupy.linux.cuda-example`'s failure seems to be due to scipy's SSL certification expiration, which we can ignore. The remaining is `cupy.linux.cuda-rapids`.

ev-br:
A gentle ping: the remaining CI failures look unrelated?

takagi:
Very sorry for my delay. I could not merge this because the CI with rapids fails with some strange reason that says cupy can not find `cuda_runtime.h`.
```
E           cupy.cuda.compiler.CompileException: /tmp/tmp343o65t_/a4a0edba0b659d6e7103330e177d8e6105b6b860.cubin.cu(2): catastrophic error: cannot open source file "cuda_runtime.h"
```
However, the file does exist at the CUDA directory in the rapids docker image we use for our CI and I could not reproduce the failure well. So, I have pushed a commit that makes the tests skip in CI with rapids. Let me give another CI try with it.

Other failures are not related with this PR.

kmaehashi:
@takagi 
The problem is that RAPIDS Docker image is conda installation, and CUDA_PATH is pointing to the package installation directory instead of `/usr/local/cuda`. Thus, the header files there cannot be discovered by CuPy.

Rather than skipping all tests on RAPIDS, I think it's better to:

1. Move RAPIDS-dependent test cases (`class TestDistance`) from `test_kdtree.py` to `test_kdtree_pylibraft.py`
2. Fix [the list of tests to be run on RAPIDS image](https://github.com/cupy/cupy/blob/1eb425adeb4af057e05a8456add94d64fcf38518/.pfnci/linux/tests/cuda-rapids.sh#L15-L17). i.e., remove `cupyx_tests/scipy_tests/spatial_tests` and add `cupyx_tests/scipy_tests/spatial_tests/test_kdtree_pylibraft.py` + `cupyx_tests/scipy_tests/spatial_tests/test_distance.py`.

takagi:
Thanks, it sounds surely reasonable.

takagi:
record: https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=61373f08f178466725f26f75cebdf3d10233479e

.pfnci/linux/tests/cuda-rapids.Dockerfile
```diff
@@ -1,9 +1,11 @@
-ARG BASE_IMAGE="rapidsai/rapidsai-core-dev:23.06-cuda11.8-devel-ubuntu22.04-py3.10"
+ARG BASE_IMAGE="rapidsai/base:23.12-cuda12.0-py3.10"
 FROM ${BASE_IMAGE}
 
+USER root
```
takagi, line 4: Use root user to write to the cache directory that will be mounted to be owned by the host user.
```
ccache: error: Failed to create temporary file for /tmp/cupy_cache/.ccache/tmp/tmp.cpp_stdout.gnumMQ: Permission denied
```

takagi (approved):
LGTM!

takagi:
Thanks, @andfoy!

ybagdasa:
Is this available in cupy 13.2.0? I see it on github but I cannot import KDTree.

jakirkham:
As noted above, this will be in CuPy 14, which is not yet released

PR #7679, opened 2023-07-01
Fix memory pool to try resolve fragmentation when limit is set

### LINKED ISSUE #7678: OOM when memory pool limit is set
### Description

When the memory pool limit is set, after allocating and freeing a first array, an OOM appears when allocating a second array. However, the memory pool states that the number of used bytes is equal to 0.

Is this a desired behaviour?

### To Reproduce

```py
import cupy as cp

# Retrieve default memory pool and set memory limit to ~1 GB
mempool = cp.get_default_memory_pool()
mempool.set_limit(1024**3)

# Allocate a first array of ~500 MB
array1 = cp.ones((256, 512, 512, 2), dtype=cp.float32)
print(mempool.used_bytes())  # prints: 536870912

# Deallocate the first array
del array1

# Try to allocate a second array of ~800MB
# An error is raised : cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 805,306,368 bytes (allocated so far: 536,870,912 bytes, limit set to: 1,073,741,824 bytes).
array2 = cp.ones((256, 512, 512, 3), dtype=cp.float32)

# But the memory pool states that the number of used bytes is equal to 0
print(mempool.used_bytes()) # prints: 0
print(mempool.free_bytes()) # prints: 536870912
print(mempool.total_bytes()) # prints: 536870912
```

### Installation

Source (`pip install cupy`)

### Environment

```
NumPy Version                : 1.24.3
SciPy Version                : 1.8.1
Cython Build Version         : 0.29.32
CUDA Build Version           : 12010
CUDA Driver Version          : 12010
CUDA Runtime Version         : 12010
cuFFT Version                : 11002
cuRAND Version               : 10302
cuSOLVER Version             : (11, 4, 5)
NVRTC Version                : (12, 1)
Thrust Version               : 200001
CUB Build Version            : 200001
Jitify Build Version         : 4a37de0
cuDNN Build Version          : 8801
cuDNN Version                : 8900
NCCL Runtime Version         : 21701
Device 0 Name                : NVIDIA GeForce RTX 4090
Device 0 Compute Capability  : 89
Device 0 PCI Bus ID          : 0000:45:00.0
```

### Additional Information

_No response_

kmaehashi:
This fixes the problem that CuPy memory allocator does not call `free_all_blocks()` / `gc.collect()` when the limit is set, causing an OOM error immediately if the memory chunk is fragmented.

Closes #7678.

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1341,12 +1341,22 @@ cdef class SingleDeviceMemoryPool:
         return None
 
     cdef BaseMemory _try_malloc(self, size_t size):
+        cdef size_t limit
+        cdef bint limit_ok
         with LockAndNoGc(self._total_bytes_lock):
-            total_bytes_limit = self._total_bytes_limit
-            total = self._total_bytes + size
-            if total_bytes_limit != 0 and total_bytes_limit < total:
-                raise OutOfMemoryError(size, total - size, total_bytes_limit)
-            self._total_bytes = total
+            limit = self._total_bytes_limit
+            if limit != 0:
+                limit_ok = (self._total_bytes + size) <= limit
+                if not limit_ok:
+                    self.free_all_blocks()
+                    limit_ok = (self._total_bytes + size) <= limit
+                if not limit_ok:
+                    gc.collect()
+                    self.free_all_blocks()
+                    limit_ok = (self._total_bytes + size) <= limit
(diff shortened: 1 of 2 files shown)


kmaehashi:
Test failures are irrelevant.

takagi (approved):
LGTM!

Issue #7548, opened 2023-05-10
Need to create CI base images on our own


kmaehashi:
I just noticed the announcement of the new Container Support Policy for CUDA official docker images:

https://gitlab.com/nvidia/container-images/cuda/blob/master/doc/support-policy.md

According to the new policy, images will be deleted after six months of the Driver's EOL (e.g., CUDA 11.7 docker images will be deleted in November 2023). We will be affected as we are using these images to run CuPy CI 😔

kmaehashi:
The next deletion is scheduled on January 2024.
https://gitlab.com/nvidia/container-images/cuda/-/blob/master/doc/container_tags.pdf

PR #7682, opened 2023-07-02
Bump CUDA minor versions used in CI

kmaehashi:
Ref #7548

Outdated minor versions will be removed from Docker Hub:
https://gitlab.com/nvidia/container-images/cuda/-/blob/master/doc/container_tags.pdf
https://gitlab.com/nvidia/container-images/cuda/-/issues/209


kmaehashi:
Test failures looks irrelevant.

PR #7640, opened 2023-06-16
Add find_peaks, peak_prominences, peak_widths to cupyx.scipy.signal

andfoy:
See https://github.com/cupy/cupy/issues/7403

PR #7694, opened 2023-07-04
Port argrel{extrema, max, min} to cupyx.scipy.signal from cusignal

andfoy:
Depends on https://github.com/cupy/cupy/pull/7640


takagi:
Thanks, @andfoy!

PR #7696, opened 2023-07-05
Port waveforms from cusignal to cupyx.scipy.signal

andfoy:
See #7403

### DIFF
--- a/cupyx/scipy/signal/_waveforms.py
+++ b/cupyx/scipy/signal/_waveforms.py
@@ -94,9 +94,11 @@ _sawtooth_kernel = cupy.ElementwiseKernel(
 def sawtooth(t, width=1.0):
     """
     Return a periodic sawtooth or triangle waveform.
+
     The sawtooth waveform has a period ``2*pi``, rises from -1 to 1 on the
     interval 0 to ``width*2*pi``, then drops from 1 to -1 on the interval
     ``width*2*pi`` to ``2*pi``. `width` must be in the interval [0, 1].
+
     Note that this is not band-limited.  It produces an infinite number
     of harmonics, which are aliased back and forth across the frequency
     spectrum.
--- a/cupyx/scipy/signal/_waveforms.py
+++ b/cupyx/scipy/signal/_waveforms.py
@@ -120,6 +120,7 @@ def sawtooth(t, width=1.0):
     Examples
     --------
     A 5 Hz waveform sampled at 500 Hz for 1 second:
+
     >>> from cupyx.scipy import signal
     >>> import matplotlib.pyplot as plt
     >>> t = np.linspace(0, 1, 500)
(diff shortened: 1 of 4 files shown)


cupyx/scipy/signal/_waveforms.py
```diff
@@ -0,0 +1,687 @@
...
+
+    Returns
+    -------
+    y : ndarray
+        Output array containing the sawtooth waveform.
+
+    Examples
+    --------
+    A 5 Hz waveform sampled at 500 Hz for 1 second:
+    >>> from cupyx.scipy import signal
```
takagi, line 123:
```suggestion

    >>> from cupyx.scipy import signal
```

cupyx/scipy/signal/_waveforms.py
```diff
@@ -0,0 +1,688 @@
...
+
+def sawtooth(t, width=1.0):
+    """
+    Return a periodic sawtooth or triangle waveform.
+    The sawtooth waveform has a period ``2*pi``, rises from -1 to 1 on the
+    interval 0 to ``width*2*pi``, then drops from 1 to -1 on the interval
+    ``width*2*pi`` to ``2*pi``. `width` must be in the interval [0, 1].
+    Note that this is not band-limited.  It produces an infinite number
+    of harmonics, which are aliased back and forth across the frequency
+    spectrum.
```
takagi, line 102: Adding blank lines
```suggestion
    Return a periodic sawtooth or triangle waveform.

    The sawtooth waveform has a period ``2*pi``, rises from -1 to 1 on the
    interval 0 to ``width*2*pi``, then drops from 1 to -1 on the interval
    ``width*2*pi`` to ``2*pi``. `width` must be in the interval [0, 1].

    Note that this is not band-limited.  It produces an infinite number
    of harmonics, which are aliased back and forth across the frequency
    spectrum.
```

cupyx/scipy/signal/_waveforms.py
```diff
@@ -0,0 +1,690 @@
...
+    // Convert  phi to radians.
+    phase = cos(temp + phi);
+    """,
+    "_chirp_phase_hyp_kernel",
+    options=("-std=c++11",),
+)
+
+
+def chirp(t, f0, t1, f1, method="linear", phi=0, vertex_zero=True,
+          type="real"):
```
takagi, line 458: Seems SciPy (stable 1.11.2 and development) does not have this argument. Where does it come from?

takagi: Ah, okay. I didn't check cuSignal implementation. Thanks!

takagi:
I already checked the full test was healthy, so I will merge this PR after the mini test finishes.

takagi:
LGTM!

takagi:
Thanks, @andfoy!

PR #7702, opened 2023-07-06
Fix dtype typings for basic creation functions (`cp.empty` etc.)

satoqz:
A few functions in `_creation/basic.py` assign `float` (i.e. the builtin python float) as the default value for `dtype`. This leads to type checking issues further down the line:

```python
cupy.empty(1, dtype=cupy.float32)
```

yields

```
Argument of type "float32" cannot be assigned to parameter "dtype" of type "type[float]" in function "empty"
  "float32" is incompatible with "type[float]"
  Type "float32" cannot be assigned to type "type[float]"
```

with basic typechecking enabled, as the valid arguments for `dtype` are narrowed down to values that share the type of `float`. The same goes for `cupy.ones`, `cupy.zeros` etc.

In case I didn't miss anything here, I would propose changing these occurrences of `dtype=float` to `dtype=None`, since the default argument does not make a difference: `dtype=None` will still lead to the creation of a double precision array.

### DIFF
--- a/cupy/_creation/basic.py
+++ b/cupy/_creation/basic.py
@@ -4,7 +4,7 @@ import cupy
 from cupy._core.internal import _get_strides_for_order_K, _update_order_char
 
 
-def empty(shape, dtype=float, order='C'):
+def empty(shape, dtype=None, order='C'):
     """Returns an array without initializing the elements.
 
     Args:
@@ -88,7 +88,7 @@ def empty_like(a, dtype=None, order='K', subok=None, shape=None):
     return cupy.ndarray(shape, dtype, memptr, strides, order)
 
 
-def eye(N, M=None, k=0, dtype=float, order='C'):
+def eye(N, M=None, k=0, dtype=None, order='C'):
     """Returns a 2-D array with ones on the diagonals and zeros elsewhere.
 
     Args:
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thanks for the pull-request @satoqz! These API signatures are kept consistent with the NumPy ones, so IMHO we should stick with `dtype=float` and instead add type annotation for arguments. I've asked @asi1024 to how we proceed with this adding type annotation.

kmaehashi:
@satoqz btw how do you configure mypy? Currently CuPy is shipped without `py.typed` so I guess the module should be treated as missing.

satoqz:
@kmaehashi I'm using vscode with

```json
{
  "python.analysis.typeCheckingMode": "basic"
}
```

set in `.vscode/settings.json`

This seems like a vscode (=pylance) issue as it doesn't appear possible to have it ignore modules without a `py.typed`... It either aggressively typechecks every module or none at all.

Using mypy directly via cli like

```sh
mypy test.py --ignore-missing-imports
```

correctly picks up the missing `py.typed` and does not show errors.

kmaehashi:
Thanks, I've confirmed the problem reproduces on my end. I also confirmed the error disappear by setting `"python.analysis.useLibraryCodeForTypes": false`; this option seems to enable type checking only for libraries with `py.typed`.

kmaehashi:
xref https://github.com/cupy/cupy/issues/4831#issuecomment-1641834027

satoqz:
While `"python.analysis.useLibraryCodeForTypes": false` does resolve the issue, it also disables all other language server functionality for the library, such as completions and signature hinting, which is suboptimal..

Should I close this PR now or would you prefer to keep it around? (Alternatively reopen it as an issue)

PR #7708, opened 2023-07-11
Temporarily disable CUB histogram

### LINKED ISSUE #7698: cupy.bincount can give incorrect results on CuPy 12.x with CUB enabled
### Description

I encountered some cases where bincount is giving incorrect results. It showed up on CI for cuCIM when using CUDA 12.0 and CuPy 12.3, but does not occur on our CUDA 11 CI runs.

I was also able to reproduce the issue locally from CuPy main and CUDA 12.2.

### To Reproduce

Here is an example taking bincount of `arange` so that all elements of the result should be 1. It can be seen that the output has zeros in some locations:

```py
import cupy as cp

cp.bincount(cp.arange(65536, dtype=cp.int32))
```
gives
```
array([1, 1, 1, ..., 0, 0, 0])
```
and `cp.bincount(cp.arange(65536, dtype=cp.int32)).sum()` is 32768 instead of the expected 65536.

Running the same thing in NumPy or in CuPy with `CUPY_ACCELERATORS=""` set gives the expected result that is all ones.

### Installation

None

### Environment

OS                           : Linux-5.15.0-76-generic-x86_64-with-glibc2.31
CuPy Version                 : 13.0.0a1
NumPy Version                : 1.24.4
SciPy Version                : 1.11.1
Cython Build Version         : 0.29.30
Cython Runtime Version       : 0.29.35
CUDA Build Version           : 12020
CUDA Runtime Version         : 12020
cuSOLVER Version             : (11, 5, 0)
Thrust Version               : 200101
CUB Build Version            : 200101
Jitify Build Version         : 4a37de0
cuDNN Build Version          : 8902
cuDNN Version                : 8902
NCCL Build Version           : 21803
NCCL Runtime Version         : 21803
Device 0 Name                : NVIDIA RTX A6000
Device 0 PCI Bus ID          : 0000:17:00.0

### Additional Information

I will report the issue to the CUB devs as it seems it is likely an issue in CUB itself.

leofang:
Close #7698.

### DIFF
--- a/cupy/_statistics/histogram.py
+++ b/cupy/_statistics/histogram.py
@@ -5,10 +5,6 @@ import numpy
 
 import cupy
 from cupy import _core
-from cupy._core import _accelerator
-from cupy.cuda import cub
-from cupy.cuda import common
-from cupy.cuda import runtime
 
 
 # rename builtin range for use in functions that take a range argument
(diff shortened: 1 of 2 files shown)


takagi (approved):
LGTM!

grlee77:
@leofang and @jakirkham, is it now possible to revert this commit and restore the CUB-accelerated `histogram` & `bincount` again? 

(It looks like CuPy is now using a very recent CCCL from NVIDIA)

grlee77:
It seems to be working as expected locally, pushed an MR to try in on CI here:
https://github.com/cupy/cupy/pull/8950

PR #7710, opened 2023-07-11
Fix project update workflow

kmaehashi:
- `@octokit/app.js` dropped support for node v16 and stopped working. https://github.com/octokit/app.js/releases/tag/v14.0.0 Use external action to generate GitHub App token instead of doing it on our own.
- Remove concurrency restriction to avoid noisy "Cancelled" notification to contributors.


kmaehashi:
Let me self-merge as this causes failure notification flood...

PR #7715, opened 2023-07-12
Add 1D signal b-splines to cupyx.scipy.signal

andfoy:

### DIFF
--- a/cupyx/scipy/signal/_bsplines.py
+++ b/cupyx/scipy/signal/_bsplines.py
@@ -42,6 +42,15 @@ def sepfir2d(input, hrow, hcol):
         input, (0, 1), lambda i: filters[i], None, 'reflect', 0)
 
 
+def _quadratic(x):
+    x = abs(cupy.asarray(x, dtype=float))
+    b = BSpline.basis_element(
+        cupy.asarray([-1.5, -0.5, 0.5, 1.5]), extrapolate=False)
+    out = b(x)
+    out[(x < -1.5) | (x > 1.5)] = 0
+    return out
+
+
 def _cubic(x):
     x = cupy.asarray(x, dtype=float)
     b = BSpline.basis_element(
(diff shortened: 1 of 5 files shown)


cupyx/scipy/signal/_bsplines.py
```diff
@@ -37,3 +40,353 @@ def sepfir2d(input, hrow, hcol):
...
+def _cubic(x):
+    x = cupy.asarray(x, dtype=float)
+    b = BSpline.basis_element(
+        cupy.asarray([-2, -1, 0, 1, 2]), extrapolate=False)
+    out = b(x)
+    out[(x < -2) | (x > 2)] = 0
+    return out
+
+
+def _coeff_smooth(lam):
```
takagi, line 63: Can we use kernel fusion to combine these expressions into an ElementwiseKernel?
https://docs.cupy.dev/en/stable/user_guide/kernel.html#kernel-fusion

cupyx/scipy/signal/_bsplines.py
```diff
@@ -37,3 +40,353 @@ def sepfir2d(input, hrow, hcol):
...
+    xi = 1 - 96 * lam + 24 * lam * cupy.sqrt(3 + 144 * lam)
+    omeg = cupy.arctan2(cupy.sqrt(144 * lam - 1), cupy.sqrt(xi))
+    rho = (24 * lam - 1 - cupy.sqrt(xi)) / (24 * lam)
+    rho = rho * cupy.sqrt(
+        (48 * lam + 24 * lam * cupy.sqrt(3 + 144 * lam)) / xi)
+    return rho, omeg
+
+
+def _hc(k, cs, rho, omega):
```
takagi, line 72: ditto(kernel fusion)

cupyx/scipy/signal/_bsplines.py
```diff
@@ -37,3 +40,353 @@ def sepfir2d(input, hrow, hcol):
...
+    return (cs / cupy.sin(omega) * (rho ** k) * cupy.sin(omega * (k + 1)) *
+            cupy.greater(k, -1))
+
+
+def _hs(k, cs, rho, omega):
```
takagi, line 77: ditto(kernel fusion)

cupyx/scipy/signal/_bsplines.py
takagi, line 63: As `lam` is a Python scalar, would it be enough to compute these values on CPU?

takagi:
Seems that these functions take some Python scalars, and computing coefficients on CPU then eventually transferring to GPU might be natural. Let me revisit next week.

andfoy:
The heavy lifting work is done by calling ``sosfilt``, where the IIR computation part takes place

cupyx/scipy/signal/_bsplines.py
takagi, line 63: If we keep this computation on GPU, I'd like to make it fused.
```suggestion
@cupy.fuse()
def _coeff_smooth(lam):
```

cupyx/scipy/signal/_bsplines.py
takagi, line 72: ditto(`cupy.fuse()`)

cupyx/scipy/signal/_bsplines.py
takagi, line 77: ditto(`@cupy.fuse()`)

takagi (commented):
LGTM except for the minor comments.

takagi:
Let me retry `cupy.win.cuda121` test.

takagi (approved):
LGTM!

takagi:
Thanks, @andfoy!

PR #7651, opened 2023-06-20
Avoid using `numpy.find_common_type`

asi1024:
`numpy.find_common_type` has been deprecated in NumPy 1.25.

PR #7721, opened 2023-07-13
Add 2D signal b-splines to cupyx.scipy.signal

andfoy:
Depends on https://github.com/cupy/cupy/pull/7715


cupyx/scipy/signal/_bsplines.py
```diff
@@ -390,3 +390,40 @@ def qspline1d_eval(cj, newx, dx=1.0, x0=0):
...
+    Returns
+    -------
+    res : ndarray
+        filterd input data
+
+    """
+    intype = Iin.dtype.char
+    hcol = cupy.asarray([1.0, 4.0, 1.0], 'f') / 6.0
+    if intype in ['F', 'D']:
+        Iin = Iin.astype('F')
```
ev-br, line 417: This is strange (both here and in scipy) : why does it only work in single precision complex. 

Also below, is there a major reason sepfir2d cannot natively support complex inputs?

andfoy: As you mention it, this is directly copied from SciPy, no modifications so far. We would need to research more about the sepfir implementation.

My guess is that all the implementations in _splines_mod.c.in (Or that template file in SciPy) only define implementations for floating point numbers and not for other data types

cupyx/scipy/signal/_bsplines.py
```diff
@@ -393,3 +406,111 @@ def qspline1d_eval(cj, newx, dx=1.0, x0=0):
         result += cj[indj] * _quadratic(newx - thisj)
     res[cond3] = result
     return res
+
+
+def cspline2d(signal, lamb=0.0, precision=-1.0):
```
takagi, line 411: Seems that scipy uses `lambda` for the name of the second argument.  Any reason to choose `lamb` here?
https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.cspline2d.html

cupyx/scipy/signal/_bsplines.py
```diff
@@ -393,3 +406,111 @@ def qspline1d_eval(cj, newx, dx=1.0, x0=0):
...
+                               axis=0)
+        return out
+
+    r, omega = compute_root_from_lambda(lamb)
+    out = _symiirorder2_nd(signal, r, omega, precision=precision, axis=-1)
+    out = _symiirorder2_nd(out, r, omega, precision=precision, axis=0)
+    return out
+
+
+def qspline2d(signal, lamb=0.0, precision=-1.0):
```
takagi, line 448: ditto (`lambda`?)

takagi:
Some tests seem to have mismatched values. Would you check them?

andfoy:
The numerical errors were caused after 9cc8dc29e108bca975e4b1d57108ea703d4731e3 in https://github.com/cupy/cupy/pull/7651, I'm debugging any call to `result_type` that might got affected

andfoy:
The culprit is `cupy._r`, which now is ignoring floating point arguments and it is casting them to integer. Specific to one of the failing tests:
imagen

When it was passing:
imagen

I don't know if this is a new behaviour or a regression caused after https://github.com/cupy/cupy/pull/7651

andfoy:
I can confirm the tests are fixed after https://github.com/cupy/cupy/pull/7815

cupyx/scipy/signal/_bsplines.py
andfoy: As `lambda` is a reserved keyword in Python, then it is not possible to use it. Since SciPy is generating the signature from C, I think it would be possible to declare lambda as a parameter. However in the practice, any use of `lambda` will trigger a `SyntaxError`. https://github.com/scipy/scipy/blob/897d811c616ec380df4000e117034e0c4cb2815c/scipy/signal/_splinemodule.c#L22

This is a bug in SciPy

takagi:
Thanks, @andfoy!

Issue #7713, opened 2023-07-11
cupyx.scipy.sparse.csr_matrix of dtype bool behaves different from scipy.sparse.csr_matrix


uellue:
### Description

Both `cupyx.scipy.sparse.csr_matrix` and `scipy.sparse.csr_matrix` support non-canonical data, i.e. unsorted column indices with duplicate entries. For dtype `bool` their behavior with duplicates is different: `scipy.sparse.csr_matrix` adds the values, while `cupyx.scipy.sparse.csr_matrix` seems to take the first entry for a specific coordinate. See example reproducer!

### To Reproduce

```py
import numpy as np
import scipy.sparse as sp
import cupy as cp
import cupyx.scipy.sparse as csp

data = np.array((True, True, True, False, True))
indices = np.array((1, 0, 0, 1, 1), dtype=int)
indptr = np.array((0, 2, 5), dtype=int)

csr = sp.csr_matrix((data, indices, indptr))
print(csr.todense())

# [[ True  True]
#  [ True  True]]

c_csr = csp.csr_matrix((cp.array(data), cp.array(indices), cp.array(indptr)))
print(c_csr.todense())

# [[ True  True]
#  [ True False]]
```

### Installation

Source (`pip install cupy`)

### Environment

```
(cupytest311) C:\Users\weber\Documents\src\sparseconverter>python -c "import cupy; cupy.show_config()"
OS                           : Windows-10-10.0.22621-SP0
Python Version               : 3.11.4
CuPy Version                 : 13.0.0a1
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 1.24.4
SciPy Version                : 1.11.1
Cython Build Version         : 0.29.36
Cython Runtime Version       : None
CUDA Root                    : C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1
nvcc PATH                    : C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\bin\nvcc.EXE
CUDA Build Version           : 12010
CUDA Driver Version          : 12010
CUDA Runtime Version         : 12010
cuBLAS Version               : (available)
cuFFT Version                : 11002
cuRAND Version               : 10302
cuSOLVER Version             : (11, 4, 5)
cuSPARSE Version             : (available)
NVRTC Version                : (12, 1)
Thrust Version               : 200001
CUB Build Version            : 200001
Jitify Build Version         : 4a37de0
cuDNN Build Version          : None
[...]

PR #7724, opened 2023-07-14
Fix csr2dense to avoid race conditions

asi1024:
Close https://github.com/cupy/cupy/issues/7713.


asi1024:
CuPy does not currently support bool sparse matrix officially, and not tested.

emcastillo (approved):

PR #7729, opened 2023-07-18
CI: remove explicit Cython installation

kmaehashi:
CI failing as Cython 3.0.0 is out. Fix the version to Cython 0.29.* for now.

PR #7731, opened 2023-07-18
[backport] CI: remove explicit Cython installation

kmaehashi:
Backport of #7729 by @kmaehashi


leofang:
```
intersphinx inventory 'https://docs.nvidia.com/cuda/cuquantum/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 404 Client Error: Not Found for url: https://docs.nvidia.com/cuda/cuquantum/objects.inv
```
this needs #7730 😅

PR #7739, opened 2023-07-20
Add type annotation in _creation.basic

asi1024:
Part of https://github.com/cupy/cupy/issues/4831.

Blocked by ~https://github.com/cupy/cupy/pull/7738~ (merged).

### DIFF
--- a/.github/workflows/pretest.yml
+++ b/.github/workflows/pretest.yml
@@ -36,6 +36,13 @@ jobs:
       run: |
         pre-commit run -a --show-diff-on-failure
 
+    - name: Typing
+      run: |
+        # TODO: Test against various NumPy/SciPy versions.
+        pip install numpy==1.25.* scipy==1.10.*
+        pip install types-setuptools==57.4.14 pytest>=7.2
+        pytest -v tests/typing_tests
+
   build-cuda:
     runs-on: ubuntu-22.04
     env:
--- a/tests/typing_tests/.gitignore
+++ b/tests/typing_tests/.gitignore
@@ -1,1 +1,0 @@
-*_numpy.pyi
--- a/tests/typing_tests/.gitignore
+++ b/tests/typing_tests/.gitignore
@@ -1,0 +1,1 @@
+*_numpy.pyi
--- a/tests/typing_tests/cupy_tests/creation_tests/test_basic.pyi
+++ b/tests/typing_tests/cupy_tests/creation_tests/test_basic.pyi
@@ -1,24 +1,24 @@
-import numpy as np
-import cupy as cp
+import numpy
+import cupy
 import cupy as xp
 
 
-x = cp.array([1, 2, 3])
+x = cupy.array([1, 2, 3])
 
 xp.empty(10)
 xp.empty((10, 20))
-xp.empty(np.array([10, 20]))
-cp.empty()  # E: Missing positional argument "shape" in call to "empty"  [call-arg]
-cp.empty(10, 20)  # E: Argument 2 to "empty" has incompatible type "int"; [arg-type]
-xp.empty(cp.array([10, 20]))  # TODO(asi1024): Fix to fail typecheck
+xp.empty(numpy.array([10, 20]))
+cupy.empty()  # E: Missing positional argument "shape" in call to "empty"  [call-arg]
+cupy.empty(10, 20)  # E: Argument 2 to "empty" has incompatible type "int"; [arg-type]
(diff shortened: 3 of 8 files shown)


kmaehashi:
@asi1024 Thanks for starting to work on this!

* Can we keep the type signature consistent with NumPy?
  * https://github.com/numpy/numpy/blob/e0c8bc50b262b4b8ce80b332b5349e165319cf42/numpy/core/multiarray.pyi#L254
  * https://github.com/numpy/numpy/blob/e0c8bc50b262b4b8ce80b332b5349e165319cf42/numpy/__init__.pyi#L981
* I think we need test cases to ensure annotations are valid, like done in NumPy. https://github.com/numpy/numpy/tree/main/numpy/typing/tests/data
* Maybe we need adjustment Sphinx to become signatures less noisy: https://cupy--7739.org.readthedocs.build/en/7739/reference/generated/cupy.empty.html#cupy-empty

.github/workflows/pretest.yml
```diff
@@ -36,6 +36,12 @@ jobs:
       run: |
         pre-commit run -a --show-diff-on-failure
 
+    - name: Typing
+      run: |
+        pip install numpy==1.25.* scipy==1.10.*
```
kmaehashi, line 41: * Let's start with this combination and later expand the variation 😄 
* Maybe better to pin mypy version as well?
* Will `types-setuptools` be required?

```suggestion
        # TODO: Test against various NumPy/SciPy versions.
        pip install numpy==1.25.* scipy==1.10.*
```

.pfnci/generate.py
```diff
@@ -175,6 +175,7 @@ def generate_dockerfile(self) -> str:
             f'RUN pyenv install {py_spec} && \\',
             f'    pyenv global {py_spec} && \\',
             '    pip install -U setuptools pip wheel',
+            '    pip install mypy==1.4.1',
```
kmaehashi, line 178: Let's test typing things in GitHub Actions instead of FlexCI with GPU, which is expensive.

```suggestion
```

.pfnci/windows/test.ps1
```diff
@@ -78,7 +78,7 @@ function Main {
 
     echo "Building..."
     $build_retval = 0
-    RunOrDie python -m pip install -U "numpy<1.24" "scipy<1.10.0"
+    RunOrDie python -m pip install -U "numpy<1.24" "scipy<1.10.0" "mypy==1.4.1"
```
kmaehashi, line 81:
```suggestion
    RunOrDie python -m pip install -U "numpy<1.24" "scipy<1.10.0"
```

tests/typing_tests/.gitignore
```diff
@@ -0,0 +1 @@
+*_numpy.pyi
```
kmaehashi, line 1: Question: Is it possible to generate files under the temporary directory? Or will that invalidate mypy cache?

tests/typing_tests/cupy_tests/creation_tests/test_basic.pyi
```diff
@@ -0,0 +1,95 @@
+import numpy as np
+import cupy as cp
+import cupy as xp
```
kmaehashi, line 3: I felt difficult distinguishing `xp`, `cp`, `np` in test cases 😅 
How about this?

```suggestion
import cupy as xp

import numpy as _np
import cupy as _cp
```

tests/typing_tests/test_typing.py
kmaehashi, line 1: Do we need `tests/typing_tests/mypy.ini` to define the rule we should obey?

kmaehashi:
Hmm Sphinx docs seems still noisy 😢 
https://cupy--7739.org.readthedocs.build/en/7739/reference/generated/cupy.empty_like.html#cupy-empty-like

tests/typing_tests/cupy_tests/creation_tests/test_basic.pyi
asi1024: I agree about not using `np` and `cp`, but I prefer `import numpy` and `import cupy` without aliasing.

kmaehashi: Sounds nice!

asi1024:
@kmaehashi @emcastillo Ready for review!

emcastillo (approved):
LGTM! Only the `pyi` nuance :)

asi1024:
CI failure seems unrelated.

kmaehashi:
LGTM!

kmaehashi (approved):

PR #7740, opened 2023-07-20
MAINT: centralize np.roots calls

ev-br:
Trivial code health tweak: remove a duplicate wrapper for `np.roots`, so that all calls go through the _polyutils version. Will also be easier to remove if/when CuPy grows a general unsymmetric eigenvalue solver.


takagi (approved):
LGTM!

Issue #7759, opened 2023-07-27
Calls into Thrust/CUB Kernels should release the GIL


wlruys:
### Description

When running CuPy in a threaded setting, calling `cp.sort` on two threads, even on different devices,  will never overlap. 
This is because the calls are blocking and they do not release the GIL. 

This fix should be adding the Cython `nogil` annotations to the wrapper and extern definitions:
 https://github.com/cupy/cupy/blob/ae82162be14de46e69b3cd2ab3327fbc94d3eb65/cupy/cuda/thrust.pyx#L71

This may also effect functions wrapped in: cupyx/jit/thrust.py

```python
import cupy as cp
from time import perf_counter as time
import argparse 
import nvtx


parser = argparse.ArgumentParser()
parser.add_argument('-ngpus', type=int, default=1)
parser.add_argument('-n', type=int, default=500000000)
parser.add_argument('-repeat', type=int, default=3)
args = parser.parse_args()

from multiprocessing.pool import ThreadPool as WorkerPool
#cp.cuda.set_allocator(cp.cuda.MemoryAsyncPool().malloc)

#Initialize Data
X = []
A = []
for i in range(args.ngpus):
    with cp.cuda.Device(i) as device:
        nvtx.push_range(message=f"Create Data on Device {i}", color="blue", domain="app")
        with cp.cuda.Stream(non_blocking=True) as stream:
            X.append(cp.arange(args.n))
            A.append([])
            stream.synchronize()
        nvtx.pop_range(domain="app")

def t1(i):
    nvtx.push_range(message=f"Start Thread {i}", color="yellow", domain="app")
    for k in range(args.repeat):
        with cp.cuda.Device(i):
            with cp.cuda.Stream(non_blocking=True) as stream:

                nvtx.push_range(message=f"Sort {k} on Device {i}", color="red", domain="app")
                t1 =time()
                X[i].sort()
                nvtx.push_range(message=f"Waiting for Sort {k} on Device {i}", color="red", domain="app")
                stream.synchronize()
                nvtx.pop_range(domain="app")

                t2=time()
                nvtx.pop_range(domain="app")
                print("Device ", X[i].device,"len", len(X[i]), "time=%.4E"%(t2-t1), flush=True)
    nvtx.pop_range(domain="app")


pool = WorkerPool(args.ngpus)
start_t = time()
results = pool.map(t1, [i for i in range(args.ngpus)])
pool.close()
pool.join()
end_t = time()
[...]

PR #7760, opened 2023-07-27
Releasing the GIL during Thrust sorts

wlruys:
As most thrust calls (including these sorts are blocking), we can release the GIL to allow concurrent CuPy threads to make progress. Fixes weak scaling for the three sorting functions mentioned in https://github.com/cupy/cupy/issues/7759. 

Very small modification, and it passes sorting function tests on my local machine.


kmaehashi:
Thanks @dialecticDolt! This looks good to me. Let me have some time to try this patch on my local.

As a side note: Thrust functions will need to allocate a workspace from CuPy's memory allocator, and GIL will be acquired for a short period during that allocation.

https://github.com/cupy/cupy/blob/ae82162be14de46e69b3cd2ab3327fbc94d3eb65/cupy/cuda/cupy_thrust.cu#L398-L399

dialecticDolt:
Thanks for the note. 
Yeah, I have also noticed that the memory allocations do not release the GIL (even on different devices & streams)  either with or without the new async mempool. It's generally only noticeable before the mempool is large enough to avoid new allocations but it can add up. 

I'm still not very familiar with the CuPy codebase, but would it be possible/safe in the default mempool to release during the cudaMalloc itself? Iirc the default mempool is separate per device (and per stream). Or is the locked time coming from something else?

In a similar test script to the one linked issue for `cp.sort`, for a `cp.empty` call I'm seeing scaling for a new allocation like the following on RTX Quadro 5000s (n=1000000000 int64):

Default Mempool:
1 GPU: 0.17 s
2 GPU: 0.34 s
4 GPU: 0.50 s

Async Mempool:
1 GPU: 0.05 s
2 GPU: 0.15 s
4 GPU: 0.29 s

dialecticDolt:
hmm, nevermind, it seems like it should be. So I'm not sure what is going on in the above timings. 
https://github.com/cupy/cupy/blob/d8e970ec0403376471ac0f71e006f9623569aa34/cupy_backends/cuda/api/runtime.pyx#L473

dialecticDolt:
After looking into this further, it seems mostly driven by a system/driver lock on ioctl preventing mallocs and frees in the same process from executing in parallel even if they are on different devices/threads.

leofang:
I think what you see is because of CuPy's mempool. Assuming you're still using the test script from https://github.com/cupy/cupy/issues/7759, you create temporary streams and discard them right after. There's nothing wrong in terms of correctness or semantics, but it does have a performance implication. Specifically, CuPy caches allocated memory based on stream IDs, but since the streams are discarded, the cached memory would not be reusable immediately, causing CuPy asking for more memory from the CUDA runtime. 

Try allocating a pool of streams before entering the busy loop, and reusing the allocated streams, it might help.

leofang:
You can also try to switch to the `MemoryAsyncPool` (backed by `cudaMallocAsync`) following the instruction:
https://docs.cupy.dev/en/stable/user_guide/memory.html#changing-memory-pool

dialecticDolt:
_re: the side discussion above_ - yep, using a smaller stream of pools (so memory is reused before being cleared) or setting async mempool helps avoid cudaFree / cudaAlloc being serialized within the same process at the driver / ioctl level

_re: this PR specifically_ - Are there any other changes needed for the GIL release patch on the Thurst sorts?

kmaehashi:
Hi @dialecticDolt, apologies for keeping you waiting :bow:
Let me proceed to merge.

kmaehashi (approved):

PR #7758, opened 2023-07-26
Revert FP16 headers from CUDA 12.2.0 to CUDA 12.1.1

kmaehashi:
Follows-up #7748. FP16 headers from CUDA 12.2.0 added dependency to `vector_types.h`, `vector_functions.h`, `cuda_runtime_api.h`, etc., causing compilation errors in environment without these headers.

CUDA 12.1.1:
```
% grep '#include' /usr/local/cuda-12.1.1/include/cuda_fp16.h
/usr/local/cuda-12.1.1/include/cuda_fp16.h:#include "cuda_fp16.hpp"
```

CUDA 12.2.0:
```
% grep '#include' /usr/local/cuda-12.2.0/include/cuda_fp16.h
/usr/local/cuda-12.2.0/include/cuda_fp16.h:#include "vector_types.h"
/usr/local/cuda-12.2.0/include/cuda_fp16.h:#include "vector_functions.h"
/usr/local/cuda-12.2.0/include/cuda_fp16.h:#include "cuda_fp16.hpp"

% grep '#include' /usr/local/cuda-12.2.0/include/vector_types.h
#include "crt/host_defines.h"

% grep '#include' /usr/local/cuda-12.2.0/include/vector_functions.h
#include "cuda_runtime_api.h"
#include "vector_functions.hpp"
```

PR #7777, opened 2023-08-04
Add notes for CUDA 12.2 on Windows support

kmaehashi:
https://github.com/cupy/cupy/pull/7758#issuecomment-1665202197

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -6,7 +6,7 @@ Requirements
 
 * `NVIDIA CUDA GPU <https://developer.nvidia.com/cuda-gpus>`_ with the Compute Capability 3.0 or larger.
 
-* `CUDA Toolkit <https://developer.nvidia.com/cuda-toolkit>`_: v11.2 / v11.3 / v11.4 / v11.5 / v11.6 / v11.7 / v11.8 / v12.0 / v12.1 / v12.2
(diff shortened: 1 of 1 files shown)


emcastillo (approved):

docs/source/install.rst
```diff
@@ -6,7 +6,7 @@ Requirements
 
 * `NVIDIA CUDA GPU <https://developer.nvidia.com/cuda-gpus>`_ with the Compute Capability 3.0 or larger.
 
-* `CUDA Toolkit <https://developer.nvidia.com/cuda-toolkit>`_: v11.2 / v11.3 / v11.4 / v11.5 / v11.6 / v11.7 / v11.8 / v12.0 / v12.1 / v12.2
+* `CUDA Toolkit <https://developer.nvidia.com/cuda-toolkit>`_: v11.2 / v11.3 / v11.4 / v11.5 / v11.6 / v11.7 / v11.8 / v12.0 / v12.1 / v12.2 [#cuda112windows]_
```
emcastillo, line 9: image
This is the rendered doc, I just give it another thought and maybe it is better to specify (linux only)?
I think it is not very important but just a quick note :)

kmaehashi: The point is that CUDA 12.2 for Windows is totally fine for most users with CUDA Toolkit installed, except for environments without headers...

kmaehashi: It seems only supported footnote syntax is numbers (1, 2, 3, ...) or symbols (`*`, `✝️`, ...)
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#footnotes

image

emcastillo: I guess it is fine, it can be seen at a glance so no prob!

PR #7779, opened 2023-08-07
Add workflow to publish to PyPI

kmaehashi:
Add a workflow to publish packages to PyPI to reduce the release effort.

This removes the need of manually downloading all baked wheels to local.


takagi (approved):
LGTM!

PR #7815, opened 2023-08-22
Fix handling of scalars in cupy.r_

andfoy:
After https://github.com/cupy/cupy/pull/7651, `cupy_r` stopped working with floating point scalars, which were being interpreted as integer ones. This was due to the use of the positions of the values (integer values) as opposed to the values themselves.

For example:
imagen

See https://github.com/cupy/cupy/pull/7721#issuecomment-1688923731

### DIFF
--- a/cupy/_indexing/generate.py
+++ b/cupy/_indexing/generate.py
@@ -56,7 +56,7 @@ class AxisConcatenator(object):
                 raise NotImplementedError
             elif type(k) in numpy.ScalarType:
                 newobj = from_data.array(k, ndmin=ndmin)
-                scalars.append(i)
+                scalars.append(k)
             else:
                 newobj = from_data.array(k, copy=False, ndmin=ndmin)
                 if ndmin > 1:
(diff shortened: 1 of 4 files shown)


cupy/_indexing/generate.py
```diff
@@ -69,7 +69,7 @@ def __getitem__(self, key):
 
         final_dtype = numpy.result_type(*arrays, *scalars)
         if final_dtype is not None:
-            for k in scalars:
+            for k in range(len(scalars)):
```
asi1024, line 72: `k` should be indices of scalars in `objs`. How about the following one?

```py
        final_dtype = numpy.result_type(*arrays, *[objs[k] for k in scalars])
        if final_dtype is not None:
            for k in scalars:
                objs[k] = objs[k].astype(final_dtype)
```

andfoy: We cannot use the scalars as indices since they would not be integers, another alternative would be to store the scalar positions into another list and then use that one

asi1024: Yes, exactly. So how about revert line 59 to `scalars.append(i)` and fix L70-L73 as I suggested?

asi1024:
Could you check the CI failures?

asi1024:
Could you check the CI failures?

asi1024:
The CI failure seems unrelated.

takagi:
LGTM for the changes in _splines.py.

asi1024 (approved):

asi1024:
@andfoy LGTM! Thanks!

PR #7836, opened 2023-08-30
Execution tests for typing tests passing rows in `typing_tests`

asi1024:
Part of https://github.com/cupy/cupy/issues/4831.

### DIFF
--- a/tests/typing_tests/test_run.py
+++ b/tests/typing_tests/test_run.py
@@ -1,26 +1,19 @@
 from pathlib import Path
 
+import pytest
 
-def generate_test(_t: Path) -> None:
-    assert not _t.name.endswith("._numpy.pyi")
-    name = str(_t.relative_to(Path(__file__).parent))
-    name = name.replace("_tests/", "/").replace("/test_", "/")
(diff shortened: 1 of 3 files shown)


tests/typing_tests/test_run.py
```diff
@@ -0,0 +1,26 @@
...
+                try:
+                    exec(_line)
+                except Exception:
+                    print(f"{_t.stem}:{_lineno} {_line}")
+                    raise
+
+    globals()[f"test_{name}"] = test
+
+
+for t in Path(__file__).parent.glob("**/*.pyi"):
```
kmaehashi, line 25: I'd like to avoid generating files on import. Can this be inside `def test_***(self):`?

kmaehashi: Ah ok I misunderstood, this does not generate files but test cases dynamically.
Still that I think it's better to generate a matrix decorator and use `def test_*`.

asi1024: @kmaehashi Fixed!

kmaehashi (approved):

emcastillo (approved):
LGTM!

PR #7840, opened 2023-09-03
Remove `cusolver_enabled`, `cub_enabled`, `thrust_enabled` flags

kmaehashi:
These flags have been marked deprecated in CuPy v8.

Issue #7727, opened 2023-07-17
Discussion: Change to link statically to cudart?


leofang:
I've meant to raise this discussion but didn't get time to do so until now, sorry for delay.

I would like to gauge the team and users' interest in changing the way CuPy links to cudart (CUDA Runtime). Currently, CuPy dynamically links to cudart, likely under the assumption that there is one unique `libcudart.so` (say from the system package manager) throughout the system. But, with either many module-based HPC clusters or Conda it is often not the case. Furthermore, not all Python libraries do this, for example PyTorch has been linking to cudart statically for very long time. The same story goes even for CUDA Python (which re-implements cudart for a different technical reason). This then raises a semantic question: When I query the cudart version, am I guaranteed that this is the same version seen by all DSOs loaded in a user process? 

The answer is no. Even CUDA libraries (cuBLAS & co) do not link to cudart dynamically, but statically (see this [CUDA Q&A](https://docs.nvidia.com/deploy/cuda-compatibility/index.html#faq)). This is done so that CUDA minor version compatibility (MVC) can fully kick in. It eliminates the need of comparing the build-time and run-time versions of cudart, and ensures that as long as the user-mode driver (UMD, meaning `libcuda.so`) is newer than the minimal version required by the CUDA major version, the generated binary would work on regardless of the CUDA Toolkit (CTK) version. Therefore, at least for modules that could work without any CUDA libraries (like `cupy.cuda.cub`), they automatically gain forward (and certain backward) compatibility by just depending on the UMD.

Also, there is no size concern in static linking to cudart, which is very tiny in size.

A significant obstacle for CuPy to make this change would be that in many places CuPy calls `runtime.runtimeGetVersion()`, assuming this is the canonical way to check the current CTK version. It may no longer be the case (probably it has never been the case?), so this is not an easy change. Also, CuPy still has to link to all libraries dynamically. Even the "core" part of CuPy (kernel compilation) requires at least NVRTC and not just the UMD. As a result, unlike some libraries that could see the immediate benefit (by only depending on UMD) I understand this change could be less attractive to CuPy, but at least this would be a discussed choice.

kmaehashi:
[...]

PR #7843, opened 2023-09-03
Lazy import cuSOLVER

kmaehashi:
This is a part of #7620.

I tried to softlink cuSOLVER as I did in NVRTC but found there is a native function calls to cuSOLVER in [`cupy_lapack.h`](https://github.com/kmaehashi/cupy/blob/401c3243a5d3bf88eaa1fe79572050609e788020/cupy_backends/cupy_lapack.h) where softlink doesn't work.
For cuSOLVER I'm thinking of using lazy-import instead of softlink.

Depends on ~#7840~ (merged).

### DIFF
--- a/.github/workflows/pretest.yml
+++ b/.github/workflows/pretest.yml
@@ -90,6 +90,14 @@ jobs:
         SPHINXOPTS="-W --keep-going -j 4" make html
         popd
 
+    - name: Import Test
+      run: |
+        python -c 'import cupy, cupyx'
+
+        # Test lazy-imported CUDA Toolkit library.
+        rm cupy_backends/cuda/libs/cusolver.*.so
+        python -c 'import cupy, cupyx'
+
   build-rocm:
     runs-on: ubuntu-22.04
 
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -18,7 +18,6 @@ from cupy_backends.cuda.api import driver  # NOQA
 from cupy_backends.cuda.api import runtime  # NOQA
 from cupy_backends.cuda.libs import cublas  # NOQA
 from cupy_backends.cuda.libs import curand  # NOQA
-from cupy_backends.cuda.libs import cusolver  # NOQA
 from cupy_backends.cuda.libs import cusparse  # NOQA
 from cupy_backends.cuda.libs import nvrtc  # NOQA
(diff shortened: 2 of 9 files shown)


leofang:
What if I change to use dlopen in `cupy_lapack.h`? Would it help unify the style?

kmaehashi:
TBH, after working on this PR, I'm a bit surprised how it is simple to lazy-import cuSOLVER than I initially expected. Although I still think soft linking (or cuSOLVER/cuSPARSE/cuFFT/... support in CUDA Python) is something we will need to support "one CuPy package for all CUDA versions", this is not mandatory to achieve #7620 in v13.

For the modules listed in #7620, I'm currently assessing only Runtime & Driver require soft linking, and the remaining can be done just by lazy-importing. I'll next try lazy-importing cuSPARSE and see if my assumption is correct.

As Jitify and CUB modules also have C/C++ level dependencies to CTK, we will need lazy-importing for them anyway.

leofang:
> For the modules listed in https://github.com/cupy/cupy/issues/7620, I'm currently assessing only Runtime & Driver require soft linking (which is imported/cimported from many places), and the remaining can be done just by lazy-importing. I'm thinking of next try lazy-importing cuSPARSE and see if my assumption is correct.
>
> As Jitify and CUB modules also have C/C++ level dependencies to CTK, we will need lazy-importing for them anyway.

In light of #7727, I would suggest to always statically link to cudart and soft-link to driver/NVRTC. Then, 
1. this eliminates completely cudart from the discussion
2. this follows CUDA Python's current approach
3. the CUB module no longer has external dependency (cudart is its sole dependency)

leofang:
> Is it safe to mix multiple versions of CUDA runtimes within the same process?

If you're referring to static linking to cudart, yes it is safe by design. As long as all DSOs see the same user-mode driver (libcuda), which is already upheld by design, it'd work just fine. Think of it as removing cudart from the dependency chain:
```
any CUDA library -> libcudart -> libcuda
```

leofang:
(btw NVCC by default statically links to cudart, I suspect we don't see that for any CuPy modules because we use gcc instead of nvcc to link, which is also fine, just FYI)

takagi (approved):
LGTM!

kmaehashi:
Thanks for the clarification, Leo! Just submitted a PR to static link cudart to CUB (#7850)

PR #7850, opened 2023-09-05
Static link to CUDA Runtime in CUB module

kmaehashi:
This is a part of #7620.

As discussed in https://github.com/cupy/cupy/pull/7843#issuecomment-1705200327 and #7727, statically link libcudart to CUB to avoid dependency to CTK.

### DIFF
--- a/install/cupy_builder/_command.py
+++ b/install/cupy_builder/_command.py
@@ -202,7 +202,7 @@ class custom_build_ext(setuptools.command.build_ext.build_ext):
         # Remove device code from list of sources, and instead add compiled
         # object files to link.
         ext.sources = sources_cpp
-        ext.extra_objects += extra_objects
+        ext.extra_objects = extra_objects + ext.extra_objects
 
         # Let setuptools do the rest of the build process, i.e., compile
         # "*.cpp" files and link object files generated from "*.cu".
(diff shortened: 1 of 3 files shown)


install/cupy_builder/_command.py
```diff
@@ -202,7 +202,7 @@ def build_extension(self, ext: setuptools.Extension) -> None:
         # Remove device code from list of sources, and instead add compiled
         # object files to link.
         ext.sources = sources_cpp
-        ext.extra_objects += extra_objects
+        ext.extra_objects = extra_objects + ext.extra_objects
```
kmaehashi, line 205: Note: The order of the command line matters: `libcudart_static.a` must be after `cupy_cub.cu.o` (which requires symbols in CUDA Runtime).

kmaehashi:
Size of `cupy/cuda/cub.cpython-39-x86_64-linux-gnu.so`:
* Without this PR: 11270936 bytes
* With this PR: 12207968 bytes (+0.9 MB)

leofang:
Thanks, Kenichi-san! Yes, I expect there's a tiny increase in the binary size.

leofang (approved):

install/cupy_builder/cupy_setup_build.py
```diff
@@ -254,6 +254,25 @@ def _rpath_base():
         raise Exception('not supported on this platform')
 
 
+def _find_static_library(name: str) -> str:
+    if PLATFORM_LINUX:
+        filename = f'lib{name}.a'
+        libdir = 'lib64'
+    elif PLATFORM_WIN32:
+        filename = f'{name}.lib'
+        libdir = 'lib'
```
kmaehashi, line 263:
```suggestion
        libdir = 'lib\\x64'
```

emcastillo (approved):
Failure is unrelated!

PR #7775, opened 2023-08-04
Add timeout to Windows CI

kmaehashi:
Currently Windows CIs for branch tests take more than the hard limit (10 hours) set in `.pfnci/config.pbtxt`.
This is preventing the kernel cache to grow and taking 8 hours for each pull request tests.

This PR intends to add soft timeout to Windows CI to grow kernel cache incrementally.

Issue #4383, opened 2020-12-01
Windows: Error in compiling CUB block reduction kernels


leofang:
From the new Windows CI https://ci.preferred.jp/cupy.win.cuda100/65528/#L96449 being introduced in #4362. Looks like Windows is not happy with `CUB_NS_PREFIX` (the line at which the first error was raised):
```
cupy/cub/cub/block/block_reduce.cuh(45): error: this declaration has no storage class or type specifier
```

Full log:
```
01:57:31.177306 STDOUT 1912]	_ TestCubReduction_param_31_{backend='block', order_and_axis=('F', None), shape=(10, 20, 30, 40)}.test_cub_argmin _
01:57:31.177306 STDOUT 1912]		
01:57:31.177306 STDOUT 1912]	self = <<cupy.testing._bundle.TestCubReduction_param_31_{backend='block', order_and_axis=('F', None), shape=(10, 20, 30, 40)} testMethod=test_cub_argmin>  parameter: {'backend': 'block', 'order_and_axis': ('F', None), 'shape': (10, 20, 30, 40)}>	
01:57:31.177306 STDOUT 1912]	xp = <module 'cupy' from 'C:\\Windows\\Temp\\flexci\\run-00135266\\work\\src\\cupy\\__init__.py'>	
01:57:31.177306 STDOUT 1912]	dtype = <class 'numpy.int8'>	
01:57:31.177306 STDOUT 1912]		
01:57:31.177480 STDOUT 1912]	    @testing.for_dtypes('bhilBHILefdFD')	
01:57:31.177480 STDOUT 1912]	    @testing.numpy_cupy_allclose(rtol=1E-5, contiguous_check=False)	
01:57:31.177480 STDOUT 1912]	    def test_cub_argmin(self, xp, dtype):	
01:57:31.177480 STDOUT 1912]	        _skip_cuda90(dtype)	
01:57:31.177480 STDOUT 1912]	        a = testing.shaped_random(self.shape, xp, dtype)	
01:57:31.177480 STDOUT 1912]	        if self.order == 'C':	
01:57:31.177480 STDOUT 1912]	            a = xp.ascontiguousarray(a)	
01:57:31.177480 STDOUT 1912]	        else:	
01:57:31.177480 STDOUT 1912]	            a = xp.asfortranarray(a)	
01:57:31.177480 STDOUT 1912]		
01:57:31.177480 STDOUT 1912]	        if xp is numpy:	
[...]

PR #7851, opened 2023-09-05
Bundle CCCL in CuPy

### LINKED ISSUE #6700: Pin CUB upstream version more aggressively?
### Description

After #4897 is concluded CuPy now enables the CUB backend by default. The status quo is
- For CTK < 11.0, CuPy uses the bundled CUB
- For CTK >= 11.0, CuPy uses the CUB from the CUDA Toolkit seen at compile time.

It would be nice if CuPy bundle pins at the CUB (and therefore Thrust) version more aggressively, and does not rely on the CTK. The reason is 2-fold:
- Due to scheduling reasons the CTK's CUB copy usually lags behind the upstream
- Unlike ~2 years ago when we started adding CUB support, now CUB/Thrust are well maintained and constantly having new features/bug fixes added in each release. 

If we lag behind upstream (as the CTK does) we can't enjoy the new additions (ex: supporting #6699 would not be possible as the CUB version needed only comes with CTK 11.6+).

### Additional Information

_No response_

leofang:
Close #6700.

~~As per @jrhemstad we expect CCCL v2.2.0 to be tagged soon-ish, but it's not a blocker for this migration 🙂 For now I manually assemble the `cupy/cccl` directory with the three components (thrust/cub/libcudacxx) at v2.1.0, and it'd be easy to update the git submodules later once v2.2.0 is out.~~

*UPDATE*: This PR bundles CCCL (= CUB + Thrust + libcu++) in CuPy. Specifically,
- the bundled CCCL is used at both build- and run- time
- both Jitify and CCCL need to be patched to support NVRTC
  - CCCL patch is merged upstream (NVIDIA/cccl#412) and I updated the submodule to point to the latest CCCL head that includes this patch. This means for now we're a bit advanced than CCCL v2.2.0.
  - Jitify patch to upstream is still work in progress. I managed to find a way to patch it in CuPy, see the updated `cupy/cuda_workaround.h`.
- all header includes are purposely converted to match the CTK layout (ex: `#include <cub/...>`, `#include <thrust/...>`, `#include <cuda/...>`) to avoid issues; also, this ensures that we can pick up from the CuPy bundle, not from CTK, by fixing the header include precedence.
- with the new CCCL layout, many libcu++ headers are picked up for CUB algorithms, and it slows down the first Jitify call to ~50 s on my machine. To mitigate this slowdown, I created a persistent cache similar to the existing kernel cache, see `cupy/cuda/jitify.pyx`. It can be disabled by the new env var `CUPY_DISABLE_JITIFY_CACHE`.

Going forward, all `type_traits` patches in CuPy/Jitify should be converted to include `<cuda/std/type_traits>`. I'll open a new issue to track this. It's not a blocker.


leofang:
TODO: Modify include paths fed to NVRTC at run time.

leofang:
Working on a CUB patch so that it's NVRTC-compilable: https://github.com/NVIDIA/cccl/pull/412

leofang:
@kmaehashi Am I actually waiting in the CI queue, or somehow I don't have permission to launch all jobs? 🤔

kmaehashi:
:eyes: https://github.com/cupy/cupy/actions/runs/6123693803/job/16622120429#step:6:1

To launch all tests, you need to say:
/test full

leofang:
ahhh silly me... thanks!

leofang:
I am clueless why the Windows CIs would fail at build time...
```
copying cupy\_core\include\cupy\cccl\libcudacxx\libcxx\test\std\experimental\memory\memory.polymorphic.allocator.class\memory.polymorphic.allocator.mem\construct_pair_const_lvalue_pair.pass.cpp -> build\lib.win-amd64-cpython-39\cupy\_core\include\cupy\cccl\libcudacxx\libcxx\test\std\experimental\memory\memory.polymorphic.allocator.class\memory.polymorphic.allocator.mem
error: could not create 'build\lib.win-amd64-cpython-39\cupy\_core\include\cupy\cccl\libcudacxx\libcxx\test\std\experimental\memory\memory.polymorphic.allocator.class\memory.polymorphic.allocator.mem\construct_pair_const_lvalue_pair.pass.cpp': No such file or directory
```

kmaehashi:
It's copying a file from:
```
C:\Windows\TEMP\flexci\run-00292523\work\src\cupy\_core\include\cupy\cccl\libcudacxx\libcxx\test\std\experimental\memory\memory.polymorphic.allocator.class\memory.polymorphic.allocator.mem\construct_pair_const_lvalue_pair.pass.cpp
```
to:
```
C:\Windows\TEMP\flexci\run-00292523\work\src\build\lib.win-amd64-cpython-39\cupy\_core\include\cupy\cccl\libcudacxx\libcxx\test\std\experimental\memory\memory.polymorphic.allocator.class\memory.polymorphic.allocator.mem\construct_pair_const_lvalue_pair.pass.cpp
```
and the latter exceeds Windows's MAX_PATH (260) chars. I'll try updating the registry key to workaroud the issue.

leofang:
(so the tags are taken as a union, not an intersection 😂)

leofang:
> Working on a CUB patch so that it's NVRTC-compilable: [NVIDIA/cccl#412](https://github.com/NVIDIA/cccl/pull/412)

Update: this is merged and I updated the submodule to point to the latest CCCL head that includes this patch. A question in discussion internally is whether we'd release CCCL v2.2.x, if so I'd ask for backporting this patch.

leofang:
@kmaehashi what's the right way of addressing CI timeout? Just retry until the cache is populated?

kmaehashi:
The failing CI `cupy.win.cuda112` seems to be a real error rather than timeout?

```
00:10:42.750065 STDERR 1048]	  c:\windows\temp\flexci\run-00293065\work\src\cupy\_core\include\cupy\cccl\libcudacxx\include\cuda\std\detail\libcxx\include\__utility/unreachable.h(28): error: identifier "__builtin_unreachable" is undefined	
```

leofang:
Yes that one I'd have to address later, don't know what happens, but I think the rest is working.

kmaehashi:
Now that #7775 got merged, the cache started to grow. It should be fully cashed after some more runs in the main branch.

leofang:
> It should be fully cached after some more runs in the main branch.

@kmaehashi sorry I may have misunderstood: Do you mean that the cache would only be populated after merged into main, not in PRs? If so I should stop retrying and instead focus on fixing Win + cuda112.

leofang:
All CI (except for win.cuda112) failed due to timeout.

leofang:
> win.cuda112 has quite some test failures, but because the CI timed out, the error log was truncated and I could not see the detail, not sure what to do... 😅

It should be hopefully fixed by commit ecb6270, which I planned to do in #7869. It seems we got #4383 fixed for free by ensuring that NVRTC sees a consistent set of CCCL headers.

leofang:
This is ready!

kmaehashi:
> Do you mean that the cache would only be populated after merged into main, not in PRs?

Sorry I dropped the ball. Yes, the cache is updated only in the branch CI to avoid pollution.

leofang:
I see. Is there a way we can temporarily increase the timeout limit? I'd prefer to see all CIs green before merging (though if not possible I can always send follow-up PRs to fix them) 🙂

kmaehashi:
Could you try extending the timeout [here](https://github.com/cupy/cupy/blob/5c32e40af32f6f9627e09d47ecfeb7e9281ccab2/.pfnci/config.pbtxt#L16) (the overall timeout - including cache download + build + pytest) and [here](https://github.com/cupy/cupy/blob/5c32e40af32f6f9627e09d47ecfeb7e9281ccab2/.pfnci/linux/tests/actions/unittest.sh#L26) (pytest timeout - 18000 sec)?

leofang:
it seems we still hit timeout on windows?

leofang:
@kmaehashi any suggestion? 🙂

kmaehashi:
```
00:20:12.284651 STDERR 2628]	RunWithTimeout : Cannot validate argument on parameter 'Timeout'. The 36000	
00:20:12.284651 STDERR 2628]	argument is greater than the maximum allowed range of 32767. Supply an	
00:20:12.284651 STDERR 2628]	argument that is less than or equal to 32767 and then try the command again.	
```

So the maximum timeout on Windows seems 32767 seconds... 😞

cupy/_core/core.pyx
```diff
@@ -2197,8 +2204,22 @@ cpdef function.Module compile_with_cache(
 
     if prepend_cupy_headers:
         source = _cupy_header + source
+    if jitify:
+        source = '#include <cupy/cuda_workaround.h>\n' + source
     extra_source = _get_header_source()
-    options += ('-I%s' % _get_header_dir_path(),)
+
+    for op in options:
+        if '-std' in op:
```
kmaehashi, line 2212: This looks a bit fragile 😅 
User specifying `-I/foo/bar-std/baz` sounds likely.

leofang: Yeah, I think checking `-std=` (with an extra character `=`) should be more robust?

leofang: or `-std=c++`...

kmaehashi: > `-std=c++`

This sounds good to me!
(Strictly speaking `('-std', 'c++11')` is also a valid option but I guess no one do this)

leofang: Done with [e381c2a](https://github.com/cupy/cupy/pull/7851/commits/e381c2a780573e731090d9d76c19ad0dda5dec52).

leofang:
> So the maximum timeout on Windows seems 32767 seconds... 😞

@kmaehashi so should I revert or modify commit https://github.com/cupy/cupy/pull/7851/commits/fd00150534c0aef3f803487c7a09262d31dcc124?

kmaehashi:
Could you try replacing `36000` with `32767` (9.1 hours) in this line?

```
    $test_retval = RunWithTimeout -timeout 36000 -output ../cupy_test_log.txt -- python -m pytest -rfEX @pytest_opts .
```

If this attempt fails, probably we need to remove the timeout...

leofang:
@kmaehashi it seems windows is all green (including cuda122!)

kmaehashi (approved):

PR #7857, opened 2023-09-10
CI: Remove path length limitation on Windows CI image

kmaehashi:
https://github.com/cupy/cupy/pull/7851#issuecomment-1712702124

### DIFF
--- a/.pfnci/windows/_flexci.ps1
+++ b/.pfnci/windows/_flexci.ps1
@@ -102,3 +102,7 @@ function PrioritizeFlexCIDaemon() {
         throw "Failed to change priority of daemon (exit code = $LastExitCode)"
     }
 }
+
+function EnableLongPaths() {
+    Set-ItemProperty "Registry::HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem" -Name LongPathsEnabled -value 1
+}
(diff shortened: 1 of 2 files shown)


kmaehashi:
Succeeded to create a long path file :tada:

```
    cmd /c "echo foo > C:\Windows\TEMP\flexci\run-00292523\work\src\build\lib.win-amd64-cpython-39\cupy\_core\include\cupy\cccl\libcudacxx\libcxx\test\std\experimental\memory\memory.polymorphic.allocator.class\memory.polymorphic.allocator.mem\construct_pair_const_lvalue_pair.pass.cpp"
```

leofang (approved):

leofang:
Thanks for speedy fix, Kenichi-san!

emcastillo (approved):
LGTM!, never thought we need to mess with the registry 😅

leofang:
This needs to be documented, as it impacts Windows users with all sources of installation (pip/conda/git), I'll handle it in #7851.

xref:
- https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation
- https://github.com/conda/conda/issues/7203

kmaehashi:
As for #7851, can't we omit test files from installation? (maybe by adjusting [package_data](https://github.com/cupy/cupy/blob/6ccec449e9695168f89b3c5613b4ab532566a49a/setup.py#L73))

leofang:
Hmmm, interesting, didn't notice it's in the test folder. I'll look into it. I suspect even manipulating path globs would hit the same issue.

PR #7811, opened 2023-08-18
Update `__repr__` to display some information

negin513:
This is a quick update to the `__repr__` method to address issue #6926 and show device information from the repr.

PR #7858, opened 2023-09-10
Add __repr__ in Sparse Matrix Classes

### LINKED ISSUE #7848: Implement string representation (`__repr__`) in sparse matrix classes
CuPy:
```pycon
>>> cupyx.scipy.sparse.coo_matrix((10,10))
<cupyx.scipy.sparse._coo.coo_matrix object at 0x7fa5716da280>
```

SciPy:
```pycon
>>> scipy.sparse.coo_matrix((10,10))
<10x10 sparse matrix of type '<class 'numpy.float64'>'
        with 0 stored elements in COOrdinate format>
```

aadya940:
Closes #7848

### DIFF
--- a/cupyx/scipy/sparse/_coo.py
+++ b/cupyx/scipy/sparse/_coo.py
@@ -169,6 +169,10 @@ class coo_matrix(sparse_data._data_matrix):
             raise ValueError('invalid shape (must be a 2-tuple of int)')
         self._shape = int(shape[0]), int(shape[1])
 
+    def __repr__(self):
+        return f"<{self._shape[0]}x{self._shape[1]} sparse matrix of type '{self.dtype}'\n" \
+               f"with {self.getnnz()} stored elements in COOrdinate format>"
+    
     def _with_data(self, data, copy=True):
         """Returns a matrix with the same sparsity structure as self,
         but with different data.  By default the index arrays
--- a/cupyx/scipy/sparse/_csc.py
+++ b/cupyx/scipy/sparse/_csc.py
@@ -44,7 +44,7 @@ class csc_matrix(_compressed._compressed_sparse_matrix):
     """
 
     format = 'csc'
-
+    
     def get(self, stream=None):
         """Returns a copy of the array on host memory.
 
@@ -79,6 +79,10 @@ class csc_matrix(_compressed._compressed_sparse_matrix):
     def _swap(self, x, y):
         return (y, x)
 
+    def __repr__(self):
+        return f"<{self._shape[0]}x{self._shape[1]} sparse matrix of type '{self.dtype}'\n" \
+               f"with {self.getnnz()} stored elements in Compressed Sparse Column format>"
+    
     def __mul__(self, other):
         from cupyx import cusparse
 
--- a/cupyx/scipy/sparse/_csr.py
+++ b/cupyx/scipy/sparse/_csr.py
@@ -128,6 +128,10 @@ class csr_matrix(_compressed._compressed_sparse_matrix):
             return csr_matrix(out)
         raise NotImplementedError
 
+    def __repr__(self):
+        return f"<{self._shape[0]}x{self._shape[1]} sparse matrix of type '{self.dtype}'\n" \
+               f"with {self.getnnz()} stored elements in Compressed Sparse Row format>"
+    
     def __eq__(self, other):
         return self._comparison(other, operator.eq, '_eq_')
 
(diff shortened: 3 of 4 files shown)


kmaehashi:
Thanks for the PR!

* Can this be done in the base class, like done in SciPy?
* Could you add `device=` to indicate the current CUDA device? c.f. #7811
* I'm wondering if `self.getnnz()` should be called here. Unlike SciPy, this is a costly operation as this require device synchronization. WDYT?

aadya940:
```
_formats = {
        "csc": [0, "Compressed Sparse Column"],
        "csr": [1, "Compressed Sparse Row"],
        "coo": [2, "COOrdinate"],
        "dia": [3, "DIAgonal"],
    }
    
    def __repr__(self, device=0):
        _, format_name = self._formats[self.format]
        return f"<%dx%d sparse matrix of type '%s'\n" \
               "\twith %d on device {device} stored elements in %s format>" % \
               (self._shape + (self.dtype, self.get, format_name))
```
So, something like this?

kmaehashi:
> * Could you add `device=` to indicate the current CUDA device? c.f. [Update `__repr__` to display some information #7811](https://github.com/cupy/cupy/pull/7811)

I meant to add `device=X` in the string representation (where `X` is the index of the CUDA device where the backing ndarray reside).

aadya940:
Hi @kmaehashi , most of the work is done I guess, you can review the changes. I was wondering what I should use in place of `getnnz()`. I was wondering `count_nonzero()` maybe? Thanking you in advance for your time!

cupyx/scipy/sparse/_base.py
```diff
@@ -33,6 +33,23 @@ def __init__(self, maxprint=50):
...
         self.maxprint = maxprint
 
+    def __repr__(self):
+        format_strings = {
+            "coo_matrix": "COOrdinate format",
+            "csr_matrix": "Compressed Sparse Row format",
+            "csc_matrix": "Compressed Sparse Column format",
+            "dia_matrix": "DIAgonal format",
+        }
+        matrix_type = str(self.__class__.__name__)
```
kmaehashi, line 43: Let's use `self.format` as a key instead of class name, like done in SciPy.

cupyx/scipy/sparse/_base.py
```diff
@@ -33,6 +33,23 @@ def __init__(self, maxprint=50):
...
+        stored_elements = self.getnnz() if hasattr(self, "getnnz") \
+            else self.nnz
+        return (
+            f"<{self._shape[0]}x{self._shape[1]} sparse matrix of type"
+            f"'{self.dtype}'\n"
+            f"on device {self.device} with {stored_elements} stored"
```
kmaehashi, line 49: I think it's better to show device ID instead of `self.device`.

kmaehashi (changes requested):

kmaehashi:
Hi @aadya940, sorry for dropping a ball here.

> * I'm wondering if self.getnnz() should be called here. Unlike SciPy, this is a costly operation as this require device synchronization. WDYT?

On the second thought, I think it's ok to use `self.nnz`. This incurs synchronization but the behavior is aligned with `cupy.ndarray.__repr__`.

Could you add at least one test case and try building & running locally?

aadya940:
cupy-contrib

cupyx/scipy/sparse/_base.py
```diff
@@ -33,6 +33,23 @@ def __init__(self, maxprint=50):
...
+        }
+        matrix_type = str(self.format)
+        stored_elements = self.getnnz() if hasattr(self, "getnnz") \
+            else self.nnz
+        return (
+            f"<{self._shape[0]}x{self._shape[1]} sparse matrix " \
+            f"of type '{self.dtype}'\n " \
+            f"on device {cupy.cuda.get_device_id()} with {stored_elements} "
+            f"stored elements in {format_strings[matrix_type]}>"
+        )
```
kmaehashi, line 51: * Align the output format with SciPy's `__repr__`.
* Device ID should be the device where sparse array resides, not the current device.
* Following the coding convention in CuPy, use single quotes.

```suggestion
        format_name = _format.get(self.format, 'unknown format')
        return (
            f'<{self._shape[0]}x{self._shape[1]} sparse matrix '
            f'of type '{self.dtype.type} '
            f'on device {self.device.id}\n'
            f'\twith {self.nnz} stored elements in {format_name}'
        )
```

Please declare static values at module-level:

```
_format = {
    'coo': 'COOrdinate format',
    'csr': 'Compressed Sparse Row format',
    'csc': 'Compressed Sparse Column format',
    'dia': 'DIAgonal format',
}
```

aadya940: Hi! @kmaehashi Tested this locally, `self.device.id` raises a `NotImplementedError` when calling `self.device`.

BHenkemans:
Hi there, I am part of a group of graduate students at the TU Eindhoven participating in a course concerning Software Evolution (in open source software). We were wondering whether or not we can finish this pull request for our course.

PR #7861, opened 2023-09-11
Avoid triggering Project Updates for updates from assignees

kmaehashi:
`github.event.issue.assignees` is an object, whereas `github.actor` is a string (GitHub login ID). `contains` check had to be done in all login IDs in the object.


asi1024 (approved):

kmaehashi:
Issue comment for test: https://github.com/cupy/cupy/pull/7826#issuecomment-1715233873
Confirmed workflow skipped as expected: https://github.com/cupy/cupy/actions/runs/6156790836

PR #7867, opened 2023-09-13
Add max_len_seq to cupyx.scipy.signal

andfoy:


takagi:
The CI failure is not related.

PR #7868, opened 2023-09-14
Clarify sparse .transpose() return type in docstrings

milesvant:
Documentation for `cupyx.scipy.sparse.csc_matrix` and `cupyx.scipy.sparse.csr_matrix` currently does not specify that `.transpose(...)` changes the sparse representation.

### DIFF
--- a/cupyx/scipy/sparse/_csc.py
+++ b/cupyx/scipy/sparse/_csc.py
@@ -341,7 +341,7 @@ class csc_matrix(_compressed._compressed_sparse_matrix):
                 Otherwise, it shared data arrays as much as possible.
 
         Returns:
-            cupyx.scipy.sparse.csr_matrix: Transpose matrix in CSR format.
+            cupyx.scipy.sparse.csr_matrix: p : `self` with the dimensions reversed.
 
         """
         if axes is not None:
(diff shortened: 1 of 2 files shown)


cupyx/scipy/sparse/_csc.py
```diff
@@ -341,7 +341,7 @@ def transpose(self, axes=None, copy=False):
                 Otherwise, it shared data arrays as much as possible.
 
         Returns:
-            cupyx.scipy.sparse.spmatrix: Transpose matrix.
+            cupyx.scipy.sparse.csr_matrix: Transpose matrix in CSR format.
```
takagi, line 344: Would you use the same description as scipy? (The part after colon)

cupyx/scipy/sparse/_csr.py
```diff
@@ -521,7 +521,7 @@ def transpose(self, axes=None, copy=False):
                 Otherwise, it shared data arrays as much as possible.
 
         Returns:
-            cupyx.scipy.sparse.spmatrix: Transpose matrix.
+            cupyx.scipy.sparse.csc_matrix: Transpose matrix in CSC format.
```
takagi, line 524: ditto

cupyx/scipy/sparse/_csc.py
milesvant: copied from https://github.com/scipy/scipy/blob/686422c4f0a71be1b4258309590fd3e9de102e18/scipy/sparse/_base.py#L768

cupyx/scipy/sparse/_csr.py
milesvant: same as above

takagi (approved):

takagi:
Thanks, @milesvant!

PR #7871, opened 2023-09-19
Add mquantiles and scoreatpercentile to cupyx.scipy.stats

andfoy:
This is required by `find_peaks_cwt` as part of #7403

### DIFF
--- a/cupyx/scipy/stats/_mstats.py
+++ b/cupyx/scipy/stats/_mstats.py
@@ -1,0 +1,158 @@
+
+import cupy
+from cupy._core.internal import _normalize_axis_index
+
+
+_quantiles_1d_map = cupy.RawKernel(r'''
+
+extern "C" {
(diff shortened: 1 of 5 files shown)


ev-br:
As an API design question: do we want to expose these  as public functions given that there's likely no chance to support masked arrays, or keep them private helpers for cupyx.scipy.signal functions that need them?

takagi:
Very sorry that I left this so long...

> As an API design question: do we want to expose these as public functions given that there's likely no chance to support masked arrays, or keep them private helpers for cupyx.scipy.signal functions that need them?

Please keep them as private helpers this time as CuPy does not have masked arrays and can not have the same API as SciPy counterparts.

andfoy:
Thanks for your feedback @takagi, I'm closing this one then

takagi:
Thanks!

PR #7877, opened 2023-09-21
Add residual recompute functionality in cupyx.scipy.sparse.linalg.cg

### LINKED ISSUE #7512: cupyx.scipy.sparse.linalg.cg
### Description

In `cupyx.scipy.sparse.linalg.cg`, the residual `r` is never recomputed. So the algorithm here would terminate early sometimes. I thought we need to compute r = b - Ax to find the true residual. And we should do so once a while, say every 50 iterations. 

### To Reproduce

```py
# Write the code here
```

### Installation

None

### Environment

```
# Paste the output here
```

### Additional Information

_No response_

aadya940:
Closes #7512

### DIFF
--- a/cupyx/scipy/sparse/linalg/_iterative.py
+++ b/cupyx/scipy/sparse/linalg/_iterative.py
@@ -10,7 +10,7 @@ from cupyx.scipy.sparse.linalg import _interface
 
 
 def cg(A, b, x0=None, tol=1e-5, maxiter=None, M=None, callback=None,
-       atol=None):
+       atol=None, residual_interval = 50):
     """Uses Conjugate Gradient iteration to solve ``Ax = b``.
 
     Args:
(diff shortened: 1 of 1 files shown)


cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -33,6 +33,8 @@ def cg(A, b, x0=None, tol=1e-5, maxiter=None, M=None, callback=None,
             iteration. It is called as ``callback(xk)``, where ``xk`` is the
             current solution vector.
         atol (float): Tolerance for convergence.
+        residual_interval (int): Iterations after which residual must be
+            recomputed.
```
emcastillo, line 37: Instead of adding a parameter that doesn't exist in the scipy signature of this function, maybe you can try to replicate the behavior and calculate the value at the same time scipy does.

emcastillo (changes requested):
Thanks for the PR! I just left a comment regarding the added parameter. We try to be a 1:1 replica of numpy & scipy APIs when possible so we were wondering if this can be adjusted as I suggested above.
Thanks again!

PR #7881, opened 2023-09-26
Add distributed ndarray

shino16:
Adds `array.DistributedArray` to `cupyx.distributed`.

It provides initial support of

* conversion from/to `ndarray`
* element-wise operations (`ufunc`, `ElementwiseKernel`)
* reduction (max/min/sum/prod)
* matrix multiplication (`matmul`)

in a multi-GPU setting.

### DIFF
--- a/cupyx/scipy/signal/__init__.py
+++ b/cupyx/scipy/signal/__init__.py
@@ -22,6 +22,10 @@ from cupyx.scipy.signal._signaltools import sosfiltfilt  # NOQA
 from cupyx.scipy.signal._signaltools import hilbert  # NOQA
 from cupyx.scipy.signal._signaltools import hilbert2  # NOQA
 
+from cupyx.scipy.signal._resample import resample  # NOQA
+from cupyx.scipy.signal._resample import resample_poly  # NOQA
+from cupyx.scipy.signal._resample import decimate  # NOQA
+
 from cupyx.scipy.signal._polyutils import unique_roots  # NOQA
 from cupyx.scipy.signal._polyutils import invres  # NOQA
 from cupyx.scipy.signal._polyutils import invresz  # NOQA
--- a/cupyx/scipy/signal/_iir_filter_design.py
+++ b/cupyx/scipy/signal/_iir_filter_design.py
@@ -4,7 +4,7 @@ import math
 
 import cupy
 
-from cupyx.scipy.signal import (
+from cupyx.scipy.signal._iir_filter_conversions import (
     lp2bp_zpk, lp2lp_zpk, lp2hp_zpk, lp2bs_zpk, bilinear_zpk, zpk2tf, zpk2sos)
 from cupyx.scipy.signal._iir_filter_conversions import (
     buttap, cheb1ap, cheb2ap, ellipap, buttord, ellipord, cheb1ord, cheb2ord,
--- a/cupyx/scipy/signal/_resample.py
+++ b/cupyx/scipy/signal/_resample.py
@@ -1,0 +1,556 @@
+
+"""
+Signal sampling functions.
+
+Some of the functions defined here were ported directly from CuSignal under
+terms of the MIT license, under the following notice:
+
+Copyright (c) 2019-2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
+
+import operator
+from math import gcd
+
+import cupy
+from cupyx.scipy.fft import fft, rfft, fftfreq, ifft, irfft, ifftshift
+from cupyx.scipy.signal._iir_filter_design import cheby1
+from cupyx.scipy.signal._fir_filter_design import firwin
+from cupyx.scipy.signal._iir_filter_conversions import zpk2sos
+from cupyx.scipy.signal._ltisys import dlti
+from cupyx.scipy.signal._upfirdn import upfirdn, _output_len
+from cupyx.scipy.signal._signaltools import (
+    sosfiltfilt, filtfilt, sosfilt, lfilter)
+from cupyx.scipy.signal.windows._windows import get_window
+
+
+def _design_resample_poly(up, down, window):
+    """
+    Design a prototype FIR low-pass filter using the window method
+    for use in polyphase rational resampling.
+
+    Parameters
+    ----------
+    up : int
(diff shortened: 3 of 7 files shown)


cupy/_core/_reduction.pyx
```diff
@@ -581,6 +581,10 @@ cdef class _SimpleReductionKernel(_AbstractReductionKernel):
     def __call__(self, object a, axis=None, dtype=None, _ndarray_base out=None,
                  bint keepdims=False):
 
+        if hasattr(a, '__cupy_override_reduction_kernel__'):
```
emcastillo, line 584: This is currently only working for `SimpleReductionKernel`
There are other more generic Reduction kernels, I wonder if we can easily support them?

cupy/_core/core.pyx
```diff
@@ -995,6 +995,9 @@ cdef class _ndarray_base:
            :meth:`numpy.ndarray.max`
 
         """
+        if hasattr(self, '__cupy_override_reduction_kernel__'):
```
emcastillo, line 998: I would only like to have this check in the `ReductionKernel` machinery. Do you think it should be possible to do that? thanks!

cupyx/distributed/_nccl_comm.py
```diff
@@ -43,6 +43,18 @@
     _nccl_ops = {}
 
 
+def _get_nccl_dtype_and_count(array, count=None):
```
emcastillo, line 46: nice!

cupyx/distributed/array/_array.py
```diff
@@ -0,0 +1,424 @@
...
+        """Instantiate a distributed array using arguments for its attributes.
+
+        :func:`~cupy.distributed.array.distributed_array` provides a more user
+        friendly way of creating a distributed array from another array.
+
+        Args:
+            shape (tuple of ints): Length of axes.
+            dtype: Data type. It must be an argument of :class:`numpy.dtype`.
+            mode (str or mode object): Mode that determines how overlaps of
+                chunks are interpreted.
```
emcastillo, line 67: I would like a description of the available models in the docstring under a `.. note:` section :)

cupy/_core/_reduction.pyx
shino16: `__cupy_override_reduction_kernel__` has to be called here, before type checks which rule out `DistributedArray`.
I believe we can support `ReductionKernel` in the same way, by adding this hook in its `__call__` method.

cupy/_core/core.pyx
shino16: If we remove this check, `cupy._core._routines_statistics._ndarray_max` tries CUB and cuTENSOR before falling back to `cupy.max`.
So this check should be done before reaching `cupy.max.__call__`, I wonder where this check should ideally be placed.
https://github.com/shino16/cupy/blob/main/cupy/_core/_routines_statistics.pyx#L25-L43

emcastillo: That's a good call. Maybe we should rewrite those checks to make it feasible.
Like having an attribute in the ndarray that's overridden by derived classes.
`array.support_cub_for_routines`

shino16: Now `DistributedArray` overrides all the methods of `ndarray` including max/min/sum/prod, so those hooks in `cupy._core.core.ndarray` are removed. Still your idea sounds reasonable.

examples/distributed/elementwise.py
```diff
@@ -0,0 +1,45 @@
...
+
+INDEX_MAP_A = {
+    0: (slice(200)),  # arr[:200]
+    1: (slice(200, None)),  # arr[200:]
+}
+
+INDEX_MAP_B = {
+    0: (slice(None), slice(None, None, 3)),  # arr[:,  ::2]
+    1: (slice(None), slice(1, None, 2)),  # arr[:, 1::2]
+}
```
leofang, line 19: It'd be nice to comment why we need these index maps, currently this sample assumes that users have the concept of sharding, which may or may not be the case.

examples/distributed/elementwise.py
```diff
@@ -0,0 +1,45 @@
...
+
+b = distributed_array(data_b, INDEX_MAP_B)
+
+with Device(0):
+    s0 = Stream()
+    s0.use()
+
+with Device(1):
+    s1 = Stream()
+    s1.use()
```
leofang, line 34: Quality of life change: This sample assumes there exists 2 GPUs, we need a device count check at the top and do early exit.

examples/distributed/matmul.py
```diff
@@ -0,0 +1,50 @@
...
+
+data_a = numpy.arange(N * M).reshape(N, M)
+data_b = numpy.arange(M * P).reshape(M, P)
+
+#     0    M/3    M
+#   0 +-----+-----+
+#     | 0 1 |  2  |
+# N/2 +-----+-----+
+#     | 0 3 | 1 2 |
+#   N +-----+-----+
```
leofang, line 19: Q: Does this sample require 4 GPUs?

leofang (commented):
Quick drive-by comments 🙂

examples/distributed/matmul.py
emcastillo: Yeah, we need to add a runtime check here! great catch :)

cupyx/distributed/array/_array.py
```diff
@@ -0,0 +1,917 @@
...
+    __init__(self, shape, dtype, chunks_map, mode=_REPLICA_MODE, comms=None)
+
+    Multi-dimensional array distributed across multiple CUDA devices.
+
+    This class implements some elementary operations that :class:`cupy.ndarray`
+    provides. The array content is split into chunks, contiguous arrays
+    corresponding to slices of the original array. Note that one device can
+    hold multiple chunks.
+
+    This direct constructor is designed for internal calls. Users should create
```
emcastillo, line 51:
```suggestion
    This direct constructor is intended as a private API. Users should create
```

cupyx/distributed/array/_array.py
```diff
@@ -0,0 +1,917 @@
...
+        if isinstance(mode, str):
+            if mode not in _modes._MODES.keys():
+                raise RuntimeError(
+                    f'`mode` must be one of {list(_modes._MODES)}')
+            mode = _modes._MODES[mode]
+        obj._mode = mode
+
+        obj._streams = {}
+        obj._comms = comms if comms is not None else {}
+
```
emcastillo, line 95: Several of the assignments here could be moved to __init__

cupyx/distributed/array/_array.py
```diff
@@ -0,0 +1,917 @@
...
+    """
+    if isinstance(array, DistributedArray):
+        if array.mode != mode:
+            array = array.change_mode(mode)
+        if array.index_map != index_map:
+            array = array.reshard(index_map)
+        return DistributedArray(
+            array.shape, array.dtype, array._chunks_map, array._mode,
+            array._comms)
+
```
emcastillo, line 864: TODO
Support regular CuPy arrays and distribute the chunks without needing to synchronize

cupyx/distributed/array/_array.py
```diff
@@ -0,0 +1,917 @@
...
+    def strides(self):
+        """Not supported."""
+        raise NotImplementedError(
+            'DistributedArray currently does not support strides.')
+
+
+def distributed_array(
+    array: ArrayLike,
+    index_map: dict[int, Any],
+    mode: str = 'replica',
```
emcastillo, line 828: maybe better to use enum?

cupyx/distributed/array/_array.py
```diff
@@ -0,0 +1,917 @@
...
+            array = array.change_mode(mode)
+        if array.index_map != index_map:
+            array = array.reshard(index_map)
+        return DistributedArray(
+            array.shape, array.dtype, array._chunks_map, array._mode,
+            array._comms)
+
+    if isinstance(array, (numpy.ndarray, ndarray)):
+        if mode != 'replica':
+            array = array.copy()
```
emcastillo, line 867: TODO: If the array is already in the device, making a copy can cause an OOM, in some cases we may be able to copy only the chunk in the device if the array is contiguous.

cupyx/distributed/array/_chunk.py
```diff
@@ -0,0 +1,228 @@
...
+from cupy.cuda.stream import get_current_stream
+
+from cupyx.distributed.array import _modes
+from cupyx.distributed.array import _index_arith
+from cupyx.distributed.array import _data_transfer
+from cupyx.distributed.array._data_transfer import _Communicator
+
+
+class _ArrayPlaceholder:
+    # Mocks ndarray
```
emcastillo, line 22: Self note, this is used when we are waiting chunks to arrive, to the current device

emcastillo:
@shino16 can you repush your changes and open again the PR? thanks!

kmaehashi:
This pull-request has merged as #7942. Thanks again @shino16 for working on this!

PR #7885, opened 2023-09-28
Add optional argument `device_id=-1` to `get_current_stream`

shino16:
`cupy.cuda.stream._ThreadLocal.get_current_stream` has an optional parameter `device_id`. However its wrapper `cupy.cuda.stream.get_current_stream` currently does not have that parameter, and it can only check the active device.

This change gives the optional parameter `device_id` to `cupy.cuda.stream.get_current_stream` too. It is then directly passed to `_ThreadLocal.get_current_stream`.

This is going to be used in #7881, which involves changing streams on multiple devices.


emcastillo (approved):
LGTM!

Issue #7878, opened 2023-09-21
Inconsistency between fuse and elementwise kernel


Zhyrek:
### Description

Hi,

While I understand that my previous issue ( https://github.com/cupy/cupy/issues/7872 ) is a non-starter because sympy isn't supported, I do believe there is still a separate issue contained in there which is worth looking at.

It seems like conditional elementwise kernels function as expected if manually defined using cupy.ElementwiseKernel, but do not behave properly if initialized using cupy.fuse from a python expression. I've attached two code snippets which, at least in my head, should give the same result, but don't.

### To Reproduce

```py
import cupy as cp

leaky_ReLU_EWK = cp.ElementwiseKernel(

    'float64 x',

    'float64 z',

    '''
    if(x > 0){
        z = x;
    } 
    else {
        z = 0.01*x;
    }
    ''',

    'leaky_ReLU'
)

c1 = cp.array([1., -1.])
leaky_ReLU_EWK(c1)
```
array([ 1.  , -0.01])

```py
@cp.fuse
def leaky_ReLU_fuse(x):
    if(x > 0):
        return x
    else:
        return 0.01*x

leaky_ReLU_fuse(c1)
```
array([ 1., -1.])

### Installation

None

leofang:
@asi1024 can correct me but I doubt Python control flow (if/else/...) is supported by `@cupy.fuse`.

I am pretty sure, despite I haven't tried, that this should work
```
# no fuse
def leaky_ReLU_fuse(x):
    ...

func = cupy.vectorize(leaky_ReLU_fuse)
func(c1)
```

leofang:
@Zhyrek is your question answered?

Zhyrek:
Yes, I've tried out the cupy.vectorize method and it seems to work very nicely, thanks! (In fact, it even seem to work for the original sympy pipeline with a small amount of finagling!)

leofang:
Thank you for sharing, @Zhyrek! It's very encouraging to hear how CuPy is successfully used in downstream.

PR #7887, opened 2023-09-28
Prohibit conversion from Variable to Python scalar in fusion

asi1024:
Reported at https://github.com/cupy/cupy/issues/7878.

This PR fixes to raise TypeError when trying to cast `_VariableProxy` to Python scalar.


emcastillo (approved):
LGTM!

PR #7900, opened 2023-10-03
Fixed Improper Method Call: Replaced `NotImplementedError` with `NotImplemented`

fazledyn-or:
### Details

1. In file: [_csc.py](https://github.com/cupy/cupy/blob/main/cupyx/scipy/sparse/_csc.py#L104), class: `csc_matrix`, there is a special method `__mul__` that raises a `NotImplementedError`. If a special method supporting a binary operation is not implemented it should return `NotImplemented`. On the other hand, `NotImplementedError` should be raised from abstract methods inside user defined base classes to indicate that derived classes should override those methods. iCR suggested that the special method `__mul__` should return `NotImplemented` instead of raising an exception.

2. In file: [_csr.py](https://github.com/cupy/cupy/blob/main/cupyx/scipy/sparse/_csr.py#L268), class: `csr_matrix`, there is a special method `__truediv__` that raises a `NotImplementedError`. If a special method supporting a binary operation is not implemented it should return `NotImplemented`. On the other hand, `NotImplementedError` should be raised from abstract methods inside user defined base classes to indicate that derived classes should override those methods. iCR suggested that the special method `__truediv__` should return `NotImplemented` instead of raising an exception.

An example of how `NotImplemented` helps the interpreter support a binary operation is [shown here](https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations).

### Notes 
- Both the class `csr_matrix` and `csc_matrix` inherits the following classes as below. Among them, a binary operation method, such as `__mul__` is already defined and implemented in class `_base.spmatrix`. As a result, it's suggested that `NotImplemented` should be used instead of raising an exception.

  ```
  - csr_matrix
	  - _compressed._compressed_sparse_matrix
		  - sparse_data._data_matrix
			  - _base.spmatrix
				  ```
			    	    def __mul__(self, other):
        			        return self.tocsr().__mul__(other)
				  ```
                  - sparse_data._minmax_mixin
                	  - object
                  - _index.IndexMixin
                	  - object
  
  - csc_matrix
	  - _compressed._compressed_sparse_matrix (SAME as above)
  ```
- However, for special custom methods, one can opt-in to use `raise NotImplementedError`.

### CLA Requirements:
This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.

All contributed commits are already automatically signed off.

The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
- [Git Commit SignOff documentation](https://developercertificate.org/)

### Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.


kmaehashi:
As per the [definition of `NotImplemented`](https://docs.python.org/3/library/constants.html#NotImplemented):

> A special value which should be returned by the binary special methods (e.g. [__eq__()](https://docs.python.org/3/reference/datamodel.html#object.__eq__), [__lt__()](https://docs.python.org/3/reference/datamodel.html#object.__lt__), [__add__()](https://docs.python.org/3/reference/datamodel.html#object.__add__), [__rsub__()](https://docs.python.org/3/reference/datamodel.html#object.__rsub__), etc.) to indicate that the operation is not implemented with respect to the other type;

I think

```
                else:
                    raise NotImplementedError
```

should be replaced by `raise AssertionError`, not `return NotImplemented`. These `else` clauses are tests against cuSPARSE's feature availability (not testing "with respect to the other type"), and never expected to reach.

fazledyn-or:
@kmaehashi In that case, let's use `AssertionError` instead of `NotImplemented` in the mentioned code. Since I have no prior knowledge regarding the Cupy codebase, I had no idea. 

Let me add new commits to this PR.

kmaehashi:
I guess just doing `raise AssertionError` is fine here, this is the case where users should never see.

cupyx/scipy/sparse/_csr.py
```diff
@@ -262,10 +262,10 @@ def __truediv__(self, other):
             # which can use lots of memory.
             self_dense = self.todense().astype(dtype, copy=False)
             return self_dense / other.todense()
-        raise NotImplementedError
+        raise AssertionError
```
kmaehashi, line 265: This one is due to type checking.

```suggestion
        return NotImplemented
```

kmaehashi (approved):
LGTM!

asi1024 (approved):

asi1024:
@fazledyn-or LGTM. Thanks for your contribution!

PR #7869, opened 2023-09-15
Use NVRTC for compiling kernels calling `cupyx.jit.cub` APIs

leofang:
~~Blocked by #7851.~~

Follow-up of #7851.

CUB (only block/warp) & libcudacxx headers can be compiled by NVRTC + Jitify, but not Thrust headers. So, in this PR a clean-up is done to allow CUB/libcudacxx functions compiled by NVRTC. The old workarounds to supply std libraries is no longer needed.

PR #7903, opened 2023-10-04
Revert CI timeout changes

leofang:
Follow-up of #7851.


leofang:
I wonder if it might be better to merge this PR after #7869, which again touches some core headers and thus invalidates a portion of cached kernels.

PR #7920, opened 2023-10-10
Reorganize directory structure around CCCL

kmaehashi:
~(Do not merge, raising for discussion)~ Ready for review.

The current `main` branch generates a binary package with many files originated from the CCCL repository, which are essentially not required for CuPy's functionalities (C/C++ code, images, documentation, etc.)

This PR tries to mitigate the situation by placing `cccl` to `thrid_party/cccl` and taking only the necessary parts from there.

`python setup.py bdist_wheel` generates a wheel of:
* `main`: 53 MB, containing 4001 files
* This PR: 48 MB, containing 1949 files

Another idea could be to implement a filter in Python (like eb4922ebc0dc46c365149dff1729ebb4d2f82cc9), but I think it is harder to maintain.

If this is acceptable, we can also use the same strategy for Jitify and DLPack (which is not submodule currently due to directory structure) as well.

cc/ @leofang 
refs #7851

### DIFF
--- a/.github/workflows/pretest.yml
+++ b/.github/workflows/pretest.yml
@@ -84,8 +84,7 @@ jobs:
     - name: Docs Build
       run: |
         # Test to detect invalid escape sequences in docstrings (#1619)
-        # Skip checking any Python files from external dependencies
-        python -Werror::DeprecationWarning -m compileall -f -q cupy cupyx examples tests docs -x "include/cupy/cccl/"
+        python -Werror::DeprecationWarning -m compileall -f -q cupy cupyx examples tests docs
         pushd docs
         pip install -r requirements.txt
         SPHINXOPTS="-W --keep-going -j 4" make html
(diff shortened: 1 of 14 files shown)


leofang:
Thanks, @kmaehashi! I have no objection in changing the location, but
- I strongly prefer to not install things under `site-packages/third_party` (if I read the current PR correctly). I think it should be `site-packages/cupy/third_party` or `site-packages/cupy_backends/third_party`, i.e. under a subdirectory that CuPy owns
- Is file pruning done (4001->1949) in the current PR? I may have missed the relevant bits.

kmaehashi:
Thanks for the comment Leo. `third_party` directory will never be included in sdist/wheels. Symbolic links (e.g., `cupy/_core/include/cupy/_cccl/cub` pointing to `../../../../../third_party/cccl/cub/cub`) will be materialized as a real directory when generating packages.

leofang (approved):

kmaehashi:
@asi1024 Could you merge this one? 😃

asi1024:
LGTM!

asi1024 (approved):

PR #7929, opened 2023-10-13
Soft link CUDA Runtime & lazy import Jitify

kmaehashi:
This is a part of #7620. Soft link CUDART.

This PR also:
* adds a debugging feature in SoftLink. When `CUPY_DEBUG_LIBRARY_LOAD=1` env var is set, CuPy prints the diagnostic messages to stderr.
* fixes to load NVRTC symobols in `class NVRTCError`.
* closes #7834

CI will be added after implementing lazy import of Profiler.

Issue #7332, opened 2023-01-19
CUDA 12: Deprecate (or remove) `cudaProfilerInitialize`


kmaehashi:
The function has been marked deprecated for a long time, and is removed in CUDA 12.

PR #7930, opened 2023-10-13
Merge profiler module into runtime

kmaehashi:
This is a part of #7620. Soft link profiler symbols in CUDART SoftLink. Depends on #7929.

This PR also removes `cupy.cuda.profiler.initialize()` function, which was removed in CUDA 12 and marked deprecated since CuPy v12 (#7332). [Code search](https://github.com/search?q=%22cupy%22+%22profiler.initialize%22&type=code) yields no use-case for this function.


kmaehashi:
Superseded by #7954

PR #7934, opened 2023-10-14
[DO NOT MERGE] Allow Jitify to only cache CuPy-owned headers

leofang:
@kmaehashi raised the concern offline that the Jitify cache (introduced in #7851) could also cache user-supplied headers, potentially causing hard-to-debug problems, unless users read the docs and set `CUPY_DISABLE_JITIFY_CACHE=1` to force initializing a fresh cache.

This PR raises an alternative solution as per offline discussion, by caching only the headers that are owned by CuPy (including the CCCL bundle), thereby alleviating the cognition overhead on the user side.

A new issue with this PR is that if any of the CuPy headers is updated (say in a minor CuPy release) but the cache is not, we could still run into issues. The potential solution is to put the burden on us (maintainers) to remember to bump the internal `build_num` in `jitify.pyx` to invalidate the cache (which was already working back in #7851). Still, I'd like to raise for awareness and for stimulating discussion for other alternatives.

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -2188,22 +2188,7 @@ cdef inline str _translate_cucomplex_to_thrust(str source):
     return ''.join(lines)
 
 
-cpdef function.Module compile_with_cache(
-        str source, tuple options=(), arch=None, cachd_dir=None,
-        prepend_cupy_headers=True, backend='nvrtc', translate_cucomplex=False,
-        enable_cooperative_groups=False, name_expressions=None,
-        log_stream=None, bint jitify=False):
-    if translate_cucomplex:
-        source = _translate_cucomplex_to_thrust(source)
-        cupy_header_list.append('cupy/cuComplex_bridge.h')
-        prepend_cupy_headers = True
-
-    if prepend_cupy_headers:
-        source = _cupy_header + source
-    if jitify:
-        source = '#include <cupy/cuda_workaround.h>\n' + source
-    extra_source = _get_header_source()
-
+cpdef tuple assemble_cupy_compiler_options(tuple options):
     for op in options:
         if '-std=c++' in op:
             if op.endswith('03'):
@@ -2264,6 +2249,27 @@ cpdef function.Module compile_with_cache(
 
     if _cuda_path is not None:
         options += ('-I' + os.path.join(_cuda_path, 'include'),)
+
+    return options
+
+
+cpdef function.Module compile_with_cache(
+        str source, tuple options=(), arch=None, cachd_dir=None,
+        prepend_cupy_headers=True, backend='nvrtc', translate_cucomplex=False,
+        enable_cooperative_groups=False, name_expressions=None,
+        log_stream=None, bint jitify=False):
+    if translate_cucomplex:
+        source = _translate_cucomplex_to_thrust(source)
+        cupy_header_list.append('cupy/cuComplex_bridge.h')
+        prepend_cupy_headers = True
+
+    if prepend_cupy_headers:
+        source = _cupy_header + source
+    if jitify:
+        source = '#include <cupy/cuda_workaround.h>\n' + source
+    extra_source = _get_header_source()
+
+    options = assemble_cupy_compiler_options(options)
 
     return cuda.compiler._compile_module_with_cache(
         source, options, arch, cachd_dir, extra_source, backend,
--- a/cupy/cuda/jitify.pyx
+++ b/cupy/cuda/jitify.pyx
@@ -7,12 +7,12 @@ from libcpp.map cimport map as cpp_map
 from libcpp.string cimport string as cpp_str
 from libcpp.vector cimport vector
 
-from cupy.cuda import cub
-
 import atexit
 import os
 import pickle
 import tempfile
+
+from cupy.cuda import cub
 
 
 ###############################################################################
@@ -35,12 +35,12 @@ cdef extern from 'cupy_jitify.h' namespace "jitify::detail" nogil:
 
 
 # We need an internal way to invalidate the cache (say, when cuda_workaround.h
-# or the CCCL bundle is updated) without having to set the environment variable
-# CUPY_DISABLE_JITIFY_CACHE in the CI. This should never be touched by end
-# users.
+# or the CCCL bundle or the warmup_kernel below is updated) without having to
+# set the environment variable CUPY_DISABLE_JITIFY_CACHE in the CI. This should
+# never be touched by end users.
 cdef extern from *:
     """
-    const int build_num = 1;
+    const int build_num = 2;
     """
     const int build_num
 
@@ -58,8 +58,12 @@ def get_build_version():
 # We cache headers found by Jitify. This is initialized with a few built-in
 # JIT-safe headers, and expands as needed to help reduce compile time.
 cdef cpp_map[cpp_str, cpp_str] cupy_headers
-_jitify_cache_dir = None
-_jitify_cache_versions = None
+cdef cpp_map[cpp_str, cpp_str] cupy_headers_for_cache  # snapshot at init time
+
+# Module-level constants
+cdef bint _jitify_init = False
+cdef str _jitify_cache_dir = None
+cdef str _jitify_cache_versions = None
 
 
 cpdef _add_sources(dict sources):
@@ -69,12 +73,15 @@ cpdef _add_sources(dict sources):
 
 @atexit.register
 def dump_cache():
+    if not _jitify_init:
+        return
+
     # Set up a version guard for invalidating the cache. Right now,
     # we use the build-time versions of CUB/Jitify.
     # TODO(leofang): Parse CUB/Thrust/libcu++ versions at process-
     # start time, for enabling CCCL + CuPy developers?
     assert _jitify_cache_versions is not None
-    data = (_jitify_cache_versions, dict(cupy_headers))
+    data = (_jitify_cache_versions, dict(cupy_headers_for_cache))
 
     # Ensure the directory exists
     os.makedirs(_jitify_cache_dir, exist_ok=True)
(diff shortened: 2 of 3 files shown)


cupy/cuda/jitify.pyx
```diff
@@ -90,14 +97,35 @@ def dump_cache():
...
+#include <type_traits>
+#include <string>
+
+#include <cupy/complex.cuh>
+#include <cupy/carray.cuh>
+#include <cupy/complex.cuh>
+#include <cupy/cuComplex_bridge.h>
+#include <cupy/math_constants.h>
+#include <cupy/atomics.cuh>
+#include <cupy/type_dispatcher.cuh>
```
leofang, line 115: I am thinking perhaps we don't need these.

kmaehashi:
Thanks for working on this Leo!

> A new issue with this PR is that if any of the CuPy headers is updated (say in a minor CuPy release) but the cache is not, we could still run into issues. The potential solution is to put the burden on us (maintainers) to remember to bump the internal `build_num` in `jitify.pyx` to invalidate the cache (which was already working back in #7851). Still, I'd like to raise for awareness and stimulate discussion for other alternatives.

How about including CuPy's version as a cache key like we do for kernel caches? Another idea could be to generate a hash string of all `cupy/_core/include/**` files at the build time and set it as a key.

cupy/cuda/jitify.pyx
```diff
@@ -90,14 +97,35 @@ def dump_cache():
...
+#include <cooperative_groups.h>
+#include <cooperative_groups/memcpy_async.h>
+
+
+extern "C" __global__ void jitify_exercise() { }
+"""
 
+
+cdef inline void _init_cupy_headers_from_cache() except*:
     with open(f'{_jitify_cache_dir}/jitify.pickle', 'rb') as f:
```
kmaehashi, line 129: Can we use JSON instead of pickle here? Although it is unlikely to happen, pickle is known to be insecure to arbitrary code execution ([1](https://docs.python.org/3/library/pickle.html), [2](https://numpy.org/doc/stable/reference/generated/numpy.save.html)).

cupy/cuda/jitify.pyx
```diff
@@ -69,12 +73,15 @@ cpdef _add_sources(dict sources):
 
 @atexit.register
```
kmaehashi, line 74: Now that the cache is immutable, how about changing the timing of persisting cache from `atexit` to the time the cache is generated? `atexit` is often unstable (e.g., user hitting Ctrl-C multiple times) and I'd like avoid using it whenever possible.

cupy/cuda/jitify.pyx
leofang: @kmaehashi I was thinking to just dump the bytes as we did for kernel cache. It should be faster for I/O too. Thoughts?

cupy/cuda/jitify.pyx
leofang: Yes, now that the cache is immutable, we shouldn't need any `atexit` work anymore, good catch!

cupy/cuda/jitify.pyx
leofang: Removed in 9420ec1.

cupy/cuda/jitify.pyx
leofang: Took a stab and implemented it in 9420ec1. Let me know what you think.

cupy/cuda/jitify.pyx
leofang: Fixed in 9420ec1.

leofang:
> How about including CuPy's version as a cache key like we do for kernel caches? Another idea could be to generate a hash string of all `cupy/_core/include/**` files at the build time and set it as a key.

Either of these would result in a cache rebuild when updating the CuPy version. Is this desired/necessary? (Asking for scrutinization, as I could be missing something 😄)

leofang:
> How about including CuPy's version as a cache key like we do for kernel caches? 

Added this in 8ef9f71, but I feel we should not do this. Let me know.

kmaehashi:
> > How about including CuPy's version as a cache key like we do for kernel caches? Another idea could be to generate a hash string of all `cupy/_core/include/**` files at the build time and set it as a key.
> 
> Either of these would result in a cache rebuild when updating the CuPy version. Is this desired/necessary? (Asking for scrutinization, as I could be missing something 😄)

I think the latter approach will not require rebuild when updating CuPy. I've implemented it as a build-time constant in https://github.com/leofang/cupy/pull/14.

cupy/cuda/jitify.pyx
kmaehashi: This is a tradeoff between ease of maintenance and performance, but given that this is a one-time overhead per process, I would prefer JSON to binary here. I've shared a quick benchmark here: https://github.com/leofang/cupy/pull/15

leofang: FWIW the serialized json file is only slightly larger than the binary equivalent.

leofang: I didn't go with json because it incurs 2 copies per serialization/deserialization: C++ strings -> Python bytes -> Python strings, but sure if the overhead is acceptable (3x slower --- 8.5 ms -> 25 ms --- is not too bad, and it's one-time only) I'll switch.

leofang: Done in ccf8a1a.

leofang:
> I think the latter approach will not require cache rebuild when updating the CuPy version (unless headers are modified). I've implemented it as a build-time constant in https://github.com/leofang/cupy/pull/14. I assume this can be used for kernel caches as well.

Thanks, @kmaehashi,  switch to use it as part of the cache key in ccf8a1a.

install/cupy_builder/_command.py
```diff
@@ -156,9 +159,27 @@ def build_extensions(self) -> None:
...
 
+        # Calculate cache key for this build
+        print('Generating cache key from header files...')
+        include_pattern = os.path.join(
+            self.build_lib, 'cupy', '_core', 'include', '**')
+        include_files = [
+            f for f in sorted(glob.glob(include_pattern, recursive=True))
+            if os.path.isfile(f)
+        ]
+        hasher = hashlib.sha1()
```
leofang, line 170: Q: Is sha1 OK?

leofang: Yes, that PR has been a while... thanks for reminder!

leofang:
Hmmm not sure why mypy would complain, the typeshed stub looks fine to me

kmaehashi:
The above commit fixed the problem in my local. `usedforsecurity` is new in Python 3.9.

leofang:
Ah great, I was looking at the CI log and thought we already use PY39, didn't realize it's a separate config for mypy 😅

cupy/cuda/jitify.pyx
```diff
@@ -143,15 +180,33 @@ cdef inline void init_cupy_headers() except*:
...
+            'CUPY_CACHE_DIR', os.path.expanduser('~/.cupy/jitify_cache'))
+
+    # Set up a version guard for invalidating the cache. Right now,
+    # we use the build-time versions of CUB/Jitify & CuPy version.
+    # TODO(leofang): Parse CUB/Thrust/libcu++ versions at process-
+    # start time, for enabling CCCL + CuPy developers?
+    global _jitify_cache_versions
+    if _jitify_cache_versions is None:
+        _jitify_cache_versions = (
+            f"{get_build_version()}_{cub.get_build_version()}_"
```
kmaehashi, line 199: Could you add NVRTC version as a cache key as well? I experienced an error when switching between CUDA 12 and CUDA 11 in my local.

leofang: Hmmm, what's the error? We are caching headers here, not device code, so I am surprised that CUDA/NVRTC versions would matter.

kmaehashi: Sorry I couldn't reproduce the error now, but I think it's better to add the version as cache key because headers from CTK are added in the cache:

```
% jq '. | keys[]' ~/.cupy/jitify_cache/jitify_1a0ca0e_200200_2_a664cdd35aad76d128b93dfa58dbc1cf8876ffd2.json | grep 'cooperative_groups.h'
"../cooperative_groups.h"
"cooperative_groups.h"

% jq '. | keys[]' ~/.cupy/jitify_cache/jitify_1a0ca0e_200200_2_a664cdd35aad76d128b93dfa58dbc1cf8876ffd2.json | grep 'cuda_runtime_api.h'
"cuda_runtime_api.h"
```

cupy/_core/core.pyx
```diff
@@ -2258,6 +2243,27 @@ cpdef function.Module compile_with_cache(
     if _cuda_path is not None:
         options += ('-I' + os.path.join(_cuda_path, 'include'),)
```
leofang, line 2244: Perhaps we should opt out this to avoid caching headers from CTK?

cupy/cuda/jitify.pyx
leofang: Ah yes, this remind me to I forgot to raise this discussion with you... I dislike caching CTK headers and think we should purge them (except for `cuda_fp16.h/hpp` which we vendor), see below. Thoughts?

cupy/_core/core.pyx
leofang: Thinking out loud, Jitify wouldn't work without headers from CTK... (ex: `cuda_fp16.h` etc include CTK headers)

cupy/cuda/jitify.pyx
leofang: Ideally we wanna add cudart's version to the cache key, not nvrtc's, because these headers are coming from cudart (more precisely, cuda-cudart-dev). However, once we statically link to cudart, `runtimeGetVerson()` be no longer reflecting the actual (package or CTK) version seen at run time. This is tricky.

kmaehashi: Indeed it's tricky... I think adding NVRTC version is handy here. Another idea could be to calculate the hash of `cuda_runtime_api.h` (jitified code returned from Jitify) and add it to the cache file name. The header includes `CUDART_VERSION` constant.

leofang: I pushed commit [306ec8d](https://github.com/cupy/cupy/pull/7934/commits/306ec8d28b0e39f9d6c0980bca6082357e0a1b97). I parse `CUDART_VERSION` from the header (if found). If it fails, I fall back to use NVRTC's version as the proxy.

leofang:
@kmaehashi I am tweaking the way we pass `CUPY_CACHE_KEY` and notice something weird, I see this message at build time:
```
    Generating cache key from header files...
    Cache key (0 files matching build/lib.linux-x86_64-cpython-310/cupy/_core/include/**): da39a3ee5e6b4b0d3255bfef95601890afd80709
```
It seems that folder is empty. Is it correct to use `self.build_lib` in `include_pattern` to parse files?

leofang:
> It seems that folder is empty. Is it correct to use `self.build_lib` in `include_pattern` to parse files?

I pushed ff167ce, in which `self.build_lib` is replaced by `source_root`, as 1. I think it reflects our intention and 2. I need to get `CUPY_CACHE_KEY` earlier than `build_extensions()` to define it as a compile-time macro. PTAL.

leofang:
I can't see the Jenkins log... 😅

leofang:
Test failure seems to be irrelevant?

PR #7938, opened 2023-10-16
Ensure D2H copies are stream ordered and by default blocking

### LINKED ISSUE #7820: Outdated (?) documentation on `ndarray.get()`/`asnumpy()` synchronization behavior
### Description

Currently, this is what the doc says about the stream argument (in both APIs):
截圖 2023-08-26 上午1 03 46
the problematic part is "otherwise, the copy is synchronous" which is not true, because the current stream is used and the copy is synchronous only when the current stream is the null stream:
https://github.com/cupy/cupy/blob/501dc69aa1246607638ced5298c48572e8bb079f/cupy/_core/core.pyx#L1851-L1858

I suggest that we just fix the docs to match the current behavior (which existed for years) since no one came to complain 🙂

leofang:
Close #7820. 

As discussed in the linked issue, this PR adds a new, optional argument `blocking` to control the copy behavior, and make it default to `True` to avoid data race. A test that can easily expose the data race is added.

Note that during the discussion, it was proposed to name the new argument as `async`. However, `async` is a reserved Python keyword (for coroutines), so I changed it to `blocking`, but I am happy to make changes if requested.

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -808,7 +808,7 @@ def asnumpy(a, stream=None, order='C', out=None, *, blocking=True):
         out (numpy.ndarray): The output array to be written to. It must have
             compatible shape and dtype with those of ``a``'s.
         blocking (bool): If set to ``False``, the copy runs asynchronously
-            on the given (if given) or default stream, and users are
+            on the given (if given) or current stream, and users are
             responsible for ensuring the stream order. Default is ``True``,
             so the copy is synchronous (with respect to the host).
 
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -792,29 +792,35 @@ from cupy._core import fromDlpack  # NOQA
 from cupy._core import from_dlpack  # NOQA
 
 
-def asnumpy(a, stream=None, order='C', out=None):
+def asnumpy(a, stream=None, order='C', out=None, *, blocking=True):
     """Returns an array on the host memory from an arbitrary source array.
 
     Args:
         a: Arbitrary object that can be converted to :class:`numpy.ndarray`.
-        stream (cupy.cuda.Stream): CUDA stream object. If it is specified, then
-            the device-to-host copy runs asynchronously. Otherwise, the copy is
-            synchronous. Note that if ``a`` is not a :class:`cupy.ndarray`
+        stream (cupy.cuda.Stream): CUDA stream object. If given, the
+            stream is used to perform the copy. Otherwise, the current
+            stream is used. Note that if ``a`` is not a :class:`cupy.ndarray`
             object, then this argument has no effect.
         order ({'C', 'F', 'A'}): The desired memory layout of the host
-            array. When ``order`` is 'A', it uses 'F' if ``a`` is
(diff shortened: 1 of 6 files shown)


kmaehashi:
Thank you Leo! `blocking` sounds good to me :+1:

docs/source/upgrade.rst
```diff
@@ -25,6 +25,14 @@ by CCCL (currently CUDA 11 & 12), but this change leads to the following consequ
 As a result of this movement, CuPy now follows the same compiler requirement as CCCL (and, in turn, CUDA Toolkit) and requires C++11 as
 the lowest C++ standard. CCCL expects to move to C++17 in the near future.
 
+Change in :func:`cupy.asnumpy`/:meth:`cupy.ndarray.get` Behavior
+----------------------------------------------------------------
+
+When transferring a CuPy array from GPU to CPU (as a NumPy array), previously the transfer could be nonblocking and not properly ordered,
+leading to potential data race if the resulting array is modified on host immediately after the copy starts. In CuPy v13, the default
```
kmaehashi, line 32: Could it be clarified that only people using user-defiend streams have been affected by this "data race"?

leofang: Done in 93d9c54.

leofang:
(btw I tagged this PR with "blocking" since this is a "no-compat" change.)

cupy/__init__.py
```diff
@@ -792,29 +792,35 @@ def ndim(a):
...
         order ({'C', 'F', 'A'}): The desired memory layout of the host
-            array. When ``order`` is 'A', it uses 'F' if ``a`` is
-            fortran-contiguous and 'C' otherwise.
+            array. When ``order`` is 'A', it uses 'F' if the array is
+            fortran-contiguous and 'C' otherwise. The ``order`` will be
+            ignored if ``out`` is specified.
         out (numpy.ndarray): The output array to be written to. It must have
             compatible shape and dtype with those of ``a``'s.
+        blocking (bool): If set to ``False``, the copy runs asynchronously
+            on the given (if given) or default stream, and users are
```
leofang, line 811:
```suggestion
            on the given (if given) or current stream, and users are
```

cupy/_core/core.pyx
```diff
@@ -1768,19 +1764,23 @@ cdef class _ndarray_base:
...
+                stream is used to perform the copy. Otherwise, the current
+                stream is used.
             order ({'C', 'F', 'A'}): The desired memory layout of the host
                 array. When ``order`` is 'A', it uses 'F' if the array is
                 fortran-contiguous and 'C' otherwise. The ``order`` will be
                 ignored if ``out`` is specified.
             out (numpy.ndarray): Output array. In order to enable asynchronous
                 copy, the underlying memory should be a pinned memory.
+            blocking (bool): If set to ``False``, the copy runs asynchronously
+                on the given (if given) or default stream, and users are
```
leofang, line 1781:
```suggestion
                on the given (if given) or current stream, and users are
```

takagi (approved):
LGTM!

leofang:
Thanks all!

PR #7939, opened 2023-10-17
Accelerate H2D copies when the source is on pinned memory

### LINKED ISSUE #7821: Efficient CPU->GPU copies with pinned memory
### Description

As documented in https://github.com/cupy/cupy/pull/5155#discussion_r621808782, using pinned memory actually doesn't prevent the on-CPU copy to pinned memory before starting the transfer. This is unfortunate, as it adds unnecessary load on the CPU, and thus latency before the GPU work can start.

There have been discussions on this before, for example in #7532 and #3450, but I'd like to open this issue for discussing a different approach.

I understand that the cause for the additional copy is a data race, as the user can change the array that is currently being copied. But in many of our use cases, the data we transfer to the GPU is immutable, for example if it was read from disk, or if it comes out of a pre-processing pipeline. In that case, we will not touch (at least, not write to) the array after starting the transfer, and it would be safe to not copy.

Maybe it would be possible to add an explicit way of omitting this copy. Either by explicitly calling a replacement for `cupy.asarray` that is explicitly documented to only work on immutable data, or by eliding the copy if the `WRITABLE` flag is set to `False` in the source array.

### Additional Information

In `pytorch`, using pinned memory elides the additional copy, for example:

```python
import torch
f = torch.zeros(shape, dtype=torch.float32, device='cpu').pin_memory()
f[:] = some_data
f_dev = torch.as_tensor(f, device='cuda')
```

See also https://github.com/LiberTEM/sparseconverter/issues/20 for some example code and results from a very simple benchmark.

leofang:
Close #7821. ~~Merge after #7938.~~

Starting commit d7c062461ae6d3aefd58cc9f00e2df79bdb0d892 is new.


cupy/_core/core.pyx
```diff
@@ -2397,17 +2398,21 @@ cpdef _ndarray_base array(obj, dtype=None, bint copy=True, order='K',
...
     if hasattr(obj, '__cupy_get_ndarray__'):
         return _array_from_cupy_ndarray(
             obj.__cupy_get_ndarray__(), dtype, copy, order, ndmin)
 
     concat_shape, concat_type, concat_dtype = (
         _array_info_from_nested_sequence(obj))
     if concat_shape is not None:
+        # TODO(leofang): handle blocking?
+        if blocking:
+            raise NotImplementedError
```
kmaehashi, line 2411: I guess just synchronizing the stream here is enough?

kmaehashi (commented):
Thanks! LGTM except for a comment to the TODO.

cupy/_core/core.pyx
leofang: Done in 335aa8d9025332ed0b8358b13b7756135e7fb2d0, PTAL

cupy/_core/core.pyx
```diff
@@ -2406,11 +2406,15 @@ cpdef _ndarray_base array(obj, dtype=None, bint copy=True, order='K',
...
         _array_info_from_nested_sequence(obj))
     if concat_shape is not None:
-        # TODO(leofang): handle blocking?
         if blocking:
-            raise NotImplementedError
+            warnings.warn(
+                'blocking is set to True, but it is unnecessary in the '
+                'current version for handling nested sequence inputs; a '
+                'redundant synchronization would happen',
+                _util.PerformanceWarning)
```
kmaehashi, line 2414: Do we need this warning? The `blocking` argument might make sense in some cases, e.g.:

```py
data = [cupy_arr1, cupy_arr2]
x = cupy.array(data)  # users might want to pass blocking=True here
...
with stream1:
    cupy_arr1.fill(0)
```

cupy/_creation/from_data.py
```diff
@@ -28,6 +29,10 @@ def array(obj, dtype=None, copy=True, order='K', subok=False, ndmin=0):
             array (default).
         ndmin (int): Minimum number of dimensions. Ones are inserted to the
             head of the shape if needed.
+        blocking (bool): Default is ``False``, meaning if a H2D copy is needed
+            it would run asynchronously on the current stream, and users are
+            responsible for ensuring the stream order. If set to ``True``,
+            the copy is synchronous (with respect to the host).
```
kmaehashi, line 35: How about documenting that users would need to set it to True if the source object is might be modified concurrently during copy?

cupy/_core/core.pyx
leofang: In this case it would be better that a stream order is established instead of adding synchronization, i.e.
```py
data = [cupy_arr1, cupy_arr2]
x = cupy.array(data)  # blocking = False
...
e = cp.cuda.get_current_stream.record()
stream1.wait(e)
with stream1:
    cupy_arr1.fill(0)
```
The problem is it's not easy to find a balance in making the warning message detailed yet succinct 😅

leofang: Also `blocking=True` is unnecessary if we're concatenating NumPy arrays (i.e. `data = [numpy_arr1, numpy_arr2]`), this was my original thinking.

cupy/_creation/from_data.py
leofang: See commit 2810dbc.

cupy/_core/core.pyx
```diff
@@ -2544,20 +2564,33 @@ cdef _ndarray_base _array_default(obj, dtype, order, Py_ssize_t ndmin):
...
-        src_cpu[:] = a_cpu.ravel(order)
-        a.data.copy_from_host_async(mem.ptr, nbytes)
-        pinned_memory._add_to_watch_list(stream.record(), mem)
+
+    cdef intptr_t ptr_h = <intptr_t>(a_cpu.ctypes.data)
+    if pinned_memory.is_memory_pinned(ptr_h):
+        if blocking:
+            warnings.warn(
+                'pinned memory is in use but blocking is set to True, an '
+                'unnecessary synchronization would happen',
```
leofang, line 2575: I suppose the same argument (that blocking could be necessary) is also applicable here. Need to phrase it better.

cupy/_core/core.pyx
leofang: Perhaps to not warn at all (as you suggested) would save us the trouble from coming up a good warning message...? 😂

leofang: Removed in commit 6b962f7283e4870811e0024163f42063cfe32f30

cupy/_core/core.pyx
leofang: Removed the warning in commit 6b962f7283e4870811e0024163f42063cfe32f30

sk1p:
I tested this PR with the same notebook as shown in https://github.com/LiberTEM/sparseconverter/issues/20, and indeed the behavior with transferring an already pinned array is now as efficient as with pytorch. Well done!

kmaehashi (approved):
Thanks, LGTM!

PR #7943, opened 2023-10-18
MNT: Suppress CUB compilation warnings 

leofang:
CUB utilities `cub::If` and `cub::Equals` are deprecated.

There's no need to backport this PR, as the v12 branch does not see the same CCCL headers across all CUDA versions.


asi1024:
LGTM! Thanks!

PR #7954, opened 2023-10-21
Static link CUDA Runtime

kmaehashi:
This is a part of #7620.

- Static link CUDA Runtime in each CuPy moduels (Runtime, Jitify, cuRAND, Thrust)
- Lazy import Jitify
- Removes `cupy.cuda.profiler.initialize()` function, which was removed in CUDA 12 and marked deprecated since CuPy v12 (#7332). [Code search](https://github.com/search?q=%22cupy%22+%22profiler.initialize%22&type=code) yields no use-case for this function.


kmaehashi:
@leofang 
This code relies on locally-installed CUDA Toolkit version (specifically `cooperative_groups.h`). Wondering how we can rewrite this...
https://github.com/cupy/cupy/blob/1ef40363dbc0c489486a178f39b7b3b4ed223b7a/cupyx/jit/cg.py#L102-L103

kmaehashi:
> This code relies on locally-installed CUDA Toolkit version (specifically `cooperative_groups.h`).

Fixed in e89e35a9d9d312c29ff76f3c859510e3c4eacb14.  Could you review this commit @asi1024?

leofang:
Yes it should be a compile time error. LGTM.

cupy/cuda/compiler.py
```diff
@@ -166,12 +164,8 @@ def _get_arch_for_options_for_nvrtc(arch=None):
     # generate cubin (SASS) instead of PTX. See #5097 for details.
     if arch is None:
         arch = _get_arch()
-    if driver._is_cuda_python():
-        version = runtime.runtimeGetVersion()
-    else:
-        version = _cuda_hip_version
     if (
-        not _use_ptx and version >= 11010
+        not _use_ptx
```
leofang, line 168: FYI. I think this change is OK. For a more robust treatment for `_get_arch()`, I suggest to take a look at [this helper function](https://github.com/cupy/cupy/blob/1ef40363dbc0c489486a178f39b7b3b4ed223b7a/tests/cupy_tests/core_tests/test_raw.py#L551-L562) that I added recently. We don't have to do it in this PR or even in v13, since right now we almost always generate SASS which is problem free. But it is possible that the nvcc/nvrtc version is newer than the driver version, and the generated PTX would not be loadable. (I hit this when updating the PTX tests in `test_raw.py`, which prompted me to write that helper 🙂)

install/cupy_builder/_features.py
```diff
@@ -462,16 +468,17 @@ def __init__(self, ctx: Context):
...
+            # CUDA Driver
+            ([] if ctx.use_cuda_python else ['cuda']) +
+
+            # CUDA Runtime
+            _cudart_static_libs +
+
+            # CUDA Toolkit
+            ['cublas', 'cufft', 'curand', 'cusparse']
         )
+        self.static_libraries = ['cudart_static']
```
leofang, line 481: Q: Would this make the CUDA Python build also link to cudart statically?

leofang (approved):
LGTM!

leofang:
Q: Perhaps we should add an import test to `.github/workflows/pretest.yml`?

install/cupy_builder/_features.py
kmaehashi: I believe ld is be clever enough to avoid linking static libs if symbols are unused 😄

leofang: Yes I was asking a silly question. CUDA Python does not use any cudart symbols, and indeed the linker would do nothing.

takagi:
Would you have a look at cuda-python test failure?

takagi:
Do we also need to include `include '_runtime_extern.pxi'` when we're using CUDA Python? And, `cupy.linux.cuda122`'s failure is not related.

kmaehashi:
> Do we also need to include `include '_runtime_extern.pxi'` when we're using CUDA Python? And, `cupy.linux.cuda122`'s failure is not related.

Thanks, that's right! Fixed.

/test mini

leofang (commented):
I can take a final look later today. Reminder to add a CI test: https://github.com/cupy/cupy/pull/7954#issuecomment-1774370509

kmaehashi:
> I can take a final look later today. Reminder to add a CI test: [#7954 (comment)](https://github.com/cupy/cupy/pull/7954#issuecomment-1774370509)

Thanks for the reminder! Added.

kmaehashi:
> > I can take a final look later today. Reminder to add a CI test: [#7954 (comment)](https://github.com/cupy/cupy/pull/7954#issuecomment-1774370509)
> 
> Thanks for the reminder! Added.

I noticed that we can't test this PR like other modules, because `cupy_backends.cuda.api.runtime` uses static link and is not lazy-imported. After everything is done I'll add more sophisticated test, e.g., try installing/importing the generated CuPy binary within a vanilla Python docker image.

leofang:
> (2) I noticed `CUDA_VERSION` requires `cuda.h` which cannot be compiled with NVRTC.

This is new to me, thanks for sharing.

leofang (approved):
LGTM, merge?

leofang:
> I noticed that we can't test this PR like other modules, because `cupy_backends.cuda.api.runtime` uses static link and is not lazy-imported. After everything is done I'll add more sophisticated test, e.g., try installing/importing the generated CuPy binary within a vanilla Python docker image.

Another way to test it (but only on Linux) is to do `LD_DEBUG=libs python -c "import cupy"`, parse the stdout, and check if `libcudart.so` is loaded. It's a bit hacky, though.

takagi (approved):
LGTM!

PR #7955, opened 2023-10-21
Fix `NVRTCError` not calling `initialize()`

kmaehashi:
In most cases this bug does not affect users unless users are manually instantiating `NVRTCError` before calling NVRTC routines.


kmaehashi:
Test failures are unrelated and will be fixed in #7957.

takagi (approved):
LGTM!

PR #7962, opened 2023-10-24
Rename 'M' to 'a' in cupy.linalg._product.matrix_power

kerrylu:


kmaehashi:
Thanks! The argname seems to be changed in NumPy 1.15.

PR #7969, opened 2023-10-31
Update cuda-rapids.Dockerfile

sushruta:
use 23.10 instead of 23.06


kmaehashi:
Closing as no activity.

PR #7976, opened 2023-11-07
Add local memory array in addition to shared memory array

Zhyrek:
This pull request roughly attempts to copy the functionality of creating a local memory array that exists in numba (the function "numba.cuda.local.array(size, dtype)"). This function copies the code from the existing SharedMemory() function in cupy, but in the codegen piece it removes the "\_\_shared\_\_" line to instead have it be a local array. This function can be called using "cupyx.jit.local_memory(dtype, size)"

Here is a piece of code built for the current build of cupy, which demonstrates a locally-specified copy of the local_memory function. By changing the definition of the array "la" in the rawkernel to use either local_memory or shared_memory, it demonstrates the difference by writing to a cell, which another thread attempts to read from.

```
import cupy as cp
from cupyx import jit
from cupyx.jit._cuda_types import ArrayBase, TypeBase
from typing import Optional
from cupyx.jit._internal_types import BuiltinFunc, Data
from cupy_backends.cuda.api import runtime
from cupy.cuda import device
from cupyx.jit import _cuda_types
from cupyx.jit import _compile
import numpy

def to_ctype(t) -> _cuda_types.TypeBase:
    if isinstance(t, _cuda_types.TypeBase):
        return t
    return _cuda_types.Scalar(numpy.dtype(t))

class LocalMem(ArrayBase):

    def __init__(
            self,
            child_type: TypeBase,
            size: Optional[int],
            alignment: Optional[int] = None,
    ) -> None:
        if not (isinstance(size, int) or size is None):
            raise 'size of local_memory must be integer or `None`'
        if not (isinstance(alignment, int) or alignment is None):
            raise 'alignment must be integer or `None`'
        self._size = size
        self._alignment = alignment
        super().__init__(child_type, 1)

    def declvar(self, x: str, init: Optional['Data']) -> str:
        assert init is None
        if self._alignment is not None:
            code = f'__align__({self._alignment})'
        else:
            code = ''
        if self._size is None:
            code = f'extern {code} {self.child_type} {x}[]'
        else:
            code = f'{code} {self.child_type} {x}[{self._size}]'
        return code
    
class LocalMemory(BuiltinFunc):

    def __call__(self, dtype, size, alignment=None):
        """Allocates local memory and returns it as a 1-D array.

        Args:
            dtype (dtype):
                The dtype of the returned array.
            size (int or None):
                If ``int`` type, the size of static local memory.
                If ``None``, declares the local memory with extern specifier.
            alignment (int or None): Enforce the alignment via __align__(N).
        """
        super().__call__()

    def call_const(self, env, dtype, size, alignment=None):
        name = env.get_fresh_variable_name(prefix='_lmem')
        ctype = to_ctype(dtype)
        var = Data(name, LocalMem(ctype, size, alignment))
        env.decls[name] = var
        env.locals[name] = var
        return Data(name, _cuda_types.Ptr(ctype))
    
local_memory = LocalMemory()

cp.cuda.Device(2)

@jit.rawkernel(device=True)
def square(a):
    b = local_memory(int, 2) #also demonstrates local_memory array working in a device function
    b[0] = a
    b[1] = a
    return b[0]*b[1]

@jit.rawkernel()
def elementwise_square_ish(x, y, size):
    tid = jit.grid(1)
    ntid = jit.gridsize(1)
    
    #la = jit.shared_memory(int, 10) #computes x^2+1, because threads can interfere. Might have race conditions.
    la = local_memory(int, 10) #computes x^2, because threads do not interfere.
    la[tid] = 1
    
    for i in range(tid, size, ntid):
        x[i] += la[(i+5)%10] #equals 0 with local mem, if shared mem another thread set this to 1, barring race conditions
        x[i] += square(i)

c = cp.zeros(10, dtype=int)
elementwise_square_ish[1, 10](c, c, 10)
print(c)
```

### DIFF
--- a/cupyx/jit/_builtin_funcs.py
+++ b/cupyx/jit/_builtin_funcs.py
@@ -219,6 +219,28 @@ class SharedMemory(BuiltinFunc):
         env.locals[name] = var
         return Data(name, _cuda_types.Ptr(ctype))
 
+class LocalMemory(BuiltinFunc):
+
+    def __call__(self, dtype, size, alignment=None):
+        """Allocates local memory and returns it as a 1-D array.
+
+        Args:
+            dtype (dtype):
+                The dtype of the returned array.
+            size (int or None):
+                If ``int`` type, the size of static local memory.
+                If ``None``, declares the local memory with extern specifier.
+            alignment (int or None): Enforce the alignment via __align__(N).
+        """
+        super().__call__()
+
+    def call_const(self, env, dtype, size, alignment=None):
+        name = env.get_fresh_variable_name(prefix='_lmem')
+        ctype = _cuda_typerules.to_ctype(dtype)
+        var = Data(name, _cuda_types.LocalMem(ctype, size, alignment))
+        env.decls[name] = var
+        env.locals[name] = var
+        return Data(name, _cuda_types.Ptr(ctype))
 
 class AtomicOp(BuiltinFunc):
 
@@ -456,6 +478,7 @@ range_ = RangeFunc()
 syncthreads = SyncThreads()
 syncwarp = SyncWarp()
 shared_memory = SharedMemory()
+local_memory = LocalMemory()
 grid = GridFunc('grid')
 gridsize = GridFunc('gridsize')
 laneid = LaneID()
--- a/cupyx/jit/_cuda_types.py
+++ b/cupyx/jit/_cuda_types.py
@@ -243,6 +243,33 @@ class SharedMem(ArrayBase):
             code = f'{code} __shared__ {self.child_type} {x}[{self._size}]'
         return code
 
+class LocalMem(ArrayBase):
+
+    def __init__(
+            self,
+            child_type: TypeBase,
+            size: Optional[int],
+            alignment: Optional[int] = None,
+    ) -> None:
+        if not (isinstance(size, int) or size is None):
+            raise 'size of local_memory must be integer or `None`'
+        if not (isinstance(alignment, int) or alignment is None):
+            raise 'alignment must be integer or `None`'
+        self._size = size
+        self._alignment = alignment
+        super().__init__(child_type, 1)
+
+    def declvar(self, x: str, init: Optional['Data']) -> str:
+        assert init is None
+        if self._alignment is not None:
+            code = f'__align__({self._alignment})'
+        else:
+            code = ''
+        if self._size is None:
+            code = f'extern {code} {self.child_type} {x}[]'
+        else:
+            code = f'{code} {self.child_type} {x}[{self._size}]'
+        return code
 
 class Ptr(PointerBase):
 


Zhyrek:
I've been working on getting multi-dimensional local/shared arrays to work, would it be easiest to add those to this pull request, or make a separate one? Main concern with the second option is that it would overlap with the local memory definitions here.

asi1024:
@Zhyrek Thanks for your PR! Could you add tests into `tests/cupyx_tests/jit_tests/test_raw.py`?

> I've been working on getting multi-dimensional local/shared arrays to work, would it be easiest to add those to this pull request, or make a separate one? Main concern with the second option is that it would overlap with the local memory definitions here.

Both are OK, but we would like the definitions not to overlap.

Zhyrek:
@asi1024 
> Both are OK, but we would like the definitions not to overlap.

Just to confirm, my vague idea was to implement the following set of array definitions:

- Shared array: 
  - 1D: cupyx.jit.shared_memory(int, 10) (Already implemented in cupy)
  - 2D: cupyx.jit.shared_memory(int, (5, 5))
  - 3D: cupyx.jit.shared_memory(int, (3, 3, 3))
  - etc.
- Local array:
  - 1D: cupyx.jit.local_memory(int, 10) (Already implemented in this pull request)
  - 2D: cupyx.jit.local_memory(int, (5, 5))
  - 3D: cupyx.jit.local_memory(int, (3, 3, 3))
  - etc.

Is this what you mean by the definitions overlapping? 

On the other hand, I can quite easily implement some level of multi-dimensional arrays manually, and not have to deal with cupy seeming to have a hard time interpreting the tuple argument, like so:

- Shared array: 
  - 1D: cupyx.jit.shared_memory(int, 10) (Already implemented in cupy)
  - 2D: cupyx.jit.shared_memory_2d(int, 5, 5)
  - 3D: cupyx.jit.shared_memory_3d(int, 3, 3, 3)
- Local array:
  - 1D: cupyx.jit.local_memory(int, 10) (Already implemented in this pull request)
  - 2D: cupyx.jit.local_memory_2d(int, 5, 5)
  - 3D: cupyx.jit.local_memory_3d(int, 3, 3, 3)

This lacks the flexibility that is present in the implementation similar to that done by numba, but from my point of view I would likely never use local or shared arrays larger than 3 or maybe 4d.

I'll look into making some tests in the meanwhile

Zhyrek:
Added a test to compare behavior of local vs. shared memory, by having threads write to the "wrong" location in memory. Overall behavior is that for local memory this produces a sorted array of ints in order, while for shared memory it produces a reverse sorted array. 

Removed the option to call cupyx.jit.local_memory(dtype, None), as I believe the __extern__ keyword is specific to shared memory. The second argument must be an int now

Zhyrek:
Finally figured out a way to implement n-dimensional local memory. It's rather inefficient because it uses pointers to pointers, rather than a contiguous block of memory, but its probably ok for most situations and could even be more efficient for certain use cases which involve row-swapping of arrays.

For this, I have another test, `test_multidimensional_local_memory()`, which tests 2, 3, and 4d local memory arrays

I went with approximate numba syntax, so these are initialized as cupyx.jit.local_array(int, (2, 3, 4)) to make a 2x3x4 array of ints. Therefore the size parameter can be either an int or a tuple of ints (and for a tuple size of 1, it automatically converts this to an int)

cupyx/jit/_builtin_funcs.py
```diff
@@ -219,6 +219,46 @@ def call_const(self, env, dtype, size, alignment=None):
...
+                for i in range(len(_int_list)):
+                    _int_list[i] = Constant(int(_int_list[i])).obj
+                args[1] = Constant(tuple(_int_list))
+            args = tuple(args)
+        return super().call(env, *args, **kwargs)
+
+    def call_const(self, env, dtype, size, alignment=None):
+        name = env.get_fresh_variable_name(prefix='_lmem')
+        ctype = to_ctype(dtype)
+        var = Data(name, LocalMem(ctype, size, alignment))
```
asi1024, line 254:
```suggestion
        ctype = _cuda_typerules.to_ctype(dtype)
        var = Data(name, _cuda_types.LocalMem(ctype, size, alignment))
```

asi1024 (changes requested):
Could you please check if the test passes in your local environment?

Zhyrek:
Hey sorry for the trouble, I remembered to add those import changes properly on the first request, but forgot it when I re-pasted the code from my test environ with the locally specified classes (the code in the OP). Thanks for catching that.

The test does indeed pass with my local definitions of the classes/tests, so presumably it should work properly with the imports being fixed.

asi1024 (commented):
Compiling the Python code `lmem3 = local_memory(numpy.int32, (2, 2, 2, 2))` produces the following code:

```cpp
int _lmem3_0_0_0[2];
int _lmem3_0_0_1[2];
int* _lmem3_0_0[2] = {_lmem3_0_0_0,_lmem3_0_0_1};
int _lmem3_0_1_0[2];
int _lmem3_0_1_1[2];
int* _lmem3_0_1[2] = {_lmem3_0_1_0,_lmem3_0_1_1};
int** _lmem3_0[2] = {_lmem3_0_0,_lmem3_0_1};
int _lmem3_1_0_0[2];
int _lmem3_1_0_1[2];
int* _lmem3_1_0[2] = {_lmem3_1_0_0,_lmem3_1_0_1};
int _lmem3_1_1_0[2];
int _lmem3_1_1_1[2];
int* _lmem3_1_1[2] = {_lmem3_1_1_0,_lmem3_1_1_1};
int** _lmem3_1[2] = {_lmem3_1_0,_lmem3_1_1};
int*** _lmem3[2] = {_lmem3_0,_lmem3_1};
int**** lmem3 = _lmem3;
```

We don't want to generate this code because it would consume more registers compared to CUDA code `int lmem3[2][2][2][2]`.

How about generating `int lmem3[2][2][2][2]` in such a case and accept indexing syntax only like as `lmem3[i, j, k, l] = 1` to avoid temporary variables?

Zhyrek:
Thanks again for the fixes, I promise next time I submit a pull request I will actually test it properly, not just in a local build...

I originally built the ndarray code intentionally like that, because int**** and int[2][2][2][2] have different types (int[2][2][2][2] has type int*[2][2][2]) and with the default _cuda_types I wasn't able to build that latter type to reassign the variable name. For int****, it was rather easy to just do _cuda_types.Ptr( _cuda_types.Ptr( _cuda_types.Ptr( _cuda_types.Ptr(int)))) or the equivalent

However I wasn't aware of the register issue, seems like threads are capped around only 255 total registers for a thread. I will rebuild it to use int[2][2][2][2] syntax. 

Would there be interest in keeping the pointer to pointer code for small 2d array cases? I could imagine it might have some utility for row-swapping in matrix ops, but could just add to bloat.

Zhyrek:
Given the constraints of the jit compiler, I was unable to specify the type of the contiguous block, as I need to define it as "int* lmem3[2][2][2] = _lmem3", which has text on the RHS of the variable name. 

So I just went with the auto keyword, hope thats ok.

The current build uses a contiguous block array, and when compiles looks like:

```
int _lmem3[2][2][2][2];
auto lmem3 = _lmem3;
```

<details>
<summary>Full compiled code for the multidimensional test:</summary>

```
extern "C" __global__ void f(CArray<long long, 1, true, true> x, CArray<long long, 1, true, true> y, CArray<long long, 1, true, true> z) {
   int _lmem1[2][3];
   int _lmem2[2][2][2];
   int _lmem3[2][2][2][2];
  int i;
  int j;
  int k;
  int l;
  unsigned int tid = threadIdx.x + blockIdx.x * blockDim.x;
  unsigned int ntid = blockDim.x * gridDim.x;
  auto lmem = _lmem1;
  auto lmem2 = _lmem2;
  auto lmem3 = _lmem3;
  for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
    i = __it;
    for (int __it = 0, __stop = 3, __step = 1; __it < __stop; __it += __step) {
      j = __it;
      lmem[i][j] = 1;
    }
    for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
      j = __it;
      for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
        k = __it;
        lmem2[i][j][k] = 1;
        for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
          l = __it;
          lmem3[i][j][k][l] = 1;
        }
      }
    }
  }
  __syncthreads();
  for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
    i = __it;
    for (int __it = 0, __stop = 3, __step = 1; __it < __stop; __it += __step) {
      j = __it;
      { long long &_tmp_4 = x[tid]; _tmp_4 = (_tmp_4 + (long long)(lmem[i][j])); }
    }
    for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
      j = __it;
      for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
        k = __it;
        { long long &_tmp_5 = y[tid]; _tmp_5 = (_tmp_5 + (long long)(lmem2[i][j][k])); }
        for (int __it = 0, __stop = 2, __step = 1; __it < __stop; __it += __step) {
          l = __it;
          { long long &_tmp_6 = z[tid]; _tmp_6 = (_tmp_6 + (long long)(lmem3[i][j][k][l])); }
        }
      }
    }
  }
}
```
</details>

I think given how the compiler currently works 2 registers is the minimum, because defining a function not only calls local_memory, but also compiles an assign_stmt. A fix would have to intercept this assign_stmt in the compilation process and only build one array, solely for this one function.

The tests still pass for me with this new build locally, and I'm pretty sure I managed to avoid ruining any of the proper imports this time.

cupyx/jit/_builtin_funcs.py
```diff
@@ -219,6 +219,46 @@ def call_const(self, env, dtype, size, alignment=None):
...
+            args = list(args)
+            _int_list = args[1].code.split("(")[1].strip(")").split(", ")
+            if(len(_int_list) == 1): #convert 1-element tuple to integer (1d array)
+                args[1] = Constant(int(_int_list[0]))
+            else:
+                for i in range(len(_int_list)):
+                    _int_list[i] = Constant(int(_int_list[i])).obj
+                args[1] = Constant(tuple(_int_list))
+            args = tuple(args)
+        return super().call(env, *args, **kwargs)
```
asi1024, line 249:
```py
@jit.rawkernel()
def f():
    n = 10
    lmem = jit.local_memory(numpy.int32, n)

f[1, 1]()
```

The above code raises `IndexError: list index out of range`. If you intend to support only constant size, you may remove these lines.

asi1024:
Could you add `from cupyx.jit._builtin_funcs import local_memory` into `cupyx/jit/__init__.py`?

asi1024:
We don't want to rely on `auto` keyword because it tend to expose CUDA compilation error.

```py
@jit.rawkernel(device=True)
def g(array):
    pass

@jit.rawkernel()
def f():
    lmem = jit.local_memory(numpy.int32, (10, 10))
    g(lmem)

f[1, 1]()
```

```cpp
__device__ void g_1(auto array) {
  ;
}
extern "C" __global__ void f() {
   int _lmem1[10][10];
  auto lmem = _lmem1;
  g_1(lmem);
}
```

```
cupy.cuda.compiler.CompileException: /tmp/tmpwoz9d1eg/fc45a23e424e3370d4d396265dbb503b096d9eea.cubin.cu(5): error: "auto" is not allowed here
  __device__ void g_1(auto array) {
                      ^

1 error detected in the compilation of "/tmp/tmpwoz9d1eg/fc45a23e424e3370d4d396265dbb503b096d9eea.cubin.cu"
```

How about generating array type explicitly like as follows?

```cpp
__device__ void g_1(int (*array)[10]) {
  ;
}

extern "C" __global__ void f() {
   int _lmem1[10][10];
   int (*lmem)[10] = _lmem1;
   g_1(_lmem1);
}
```

Zhyrek:
Ah, I see the issue with the auto keyword now. That's definitely a big issue to not be able to pass them to device functions

However as far as I can tell there's no easy way with the machinery currently available to get the assign_stmt to build an array type variable as-is. Since this will likely involve modifying cupyx.jit._compile anyways, I may as well try to get local_memory to only use a single register in the first place.

For the index error, I've never tried building a local memory array based on a constant parameter like that, but don't see any reason to restrict it unnecessarily. I'll get that statement to work as well as the others, and will add it to the local_memory test.

Will also add the import statement to __init__

asi1024:
Sorry for my lack of words. I mean that the error message is inappropriate.
I think we can remove the `call` method and just use the `call_const` method to support only constant sizes.

Zhyrek:
Oh, specifically for that, the reason the "call" method is there is so that the code can even support tuple arguments in the first place, otherwise it will throw a "Arguments must be constant" error when it receives a Data block containing a tuple in arg[1].

Maybe the conversion code could be moved to the call_const method, but putting it in call was the first thing that worked for me when I was building it.

asi1024:
CuPy JIT is implemented to handle constant tuples. Could you remove `call` method and confirm again if the same error is raised?

Zhyrek:
Sure thing! Here's two examples of compiling the line `la2 = local_memory(int, (2,3,4))`

Using the code:
```
def call(self, env: 'Environment', *args, **kwargs):
        print(args)
        if(type(args[1]) is Data): #intercept tuple, convert to Constant(tuple)
            args = list(args)
            _int_list = args[1].code.split("(")[1].strip(")").split(", ")
            if(len(_int_list) == 1): #convert 1-element tuple to integer (1d array)
                args[1] = Constant(int(_int_list[0]))
            else:
                for i in range(len(_int_list)):
                    _int_list[i] = Constant(int(_int_list[i])).obj
                args[1] = Constant(tuple(_int_list))
            args = tuple(args)
        print(args)
        print("")
        return super().call(env, *args, **kwargs)
```
`(<Constant obj = "<class 'int'>">, <Data code = "thrust::make_tuple(2, 3, 4)", type = thrust::tuple<int, int, int>>)`
`(<Constant obj = "<class 'int'>">, <Constant obj = "(2, 3, 4)">)`

Using the code:
```
def call(self, env: 'Environment', *args, **kwargs):
        print(args)
        """
        if(type(args[1]) is Data): #intercept tuple, convert to Constant(tuple)
            args = list(args)
            _int_list = args[1].code.split("(")[1].strip(")").split(", ")
            if(len(_int_list) == 1): #convert 1-element tuple to integer (1d array)
                args[1] = Constant(int(_int_list[0]))
            else:
                for i in range(len(_int_list)):
                    _int_list[i] = Constant(int(_int_list[i])).obj
                args[1] = Constant(tuple(_int_list))
            args = tuple(args)
        """
        print(args)
        print("")
        return super().call(env, *args, **kwargs)
```
`(<Constant obj = "<class 'int'>">, <Data code = "thrust::make_tuple(2, 3, 4)", type = thrust::tuple<int, int, int>>)`
`(<Constant obj = "<class 'int'>">, <Data code = "thrust::make_tuple(2, 3, 4)", type = thrust::tuple<int, int, int>>)`
```
Cell In[3], line 134
    131         x[i] += square(i)
    133 c = cp.zeros(10, dtype=int)
--> 134 elementwise_add[1, 10](c, c, 10)
    135 print(c)

File ~/.conda/envs/cupy/lib/python3.10/site-packages/cupyx/jit/_interface.py:138, in _JitRawKernel.__getitem__.<locals>.<lambda>(*args, **kwargs)
    136 if not isinstance(block, tuple):
    137     block = (block, 1, 1)
--> 138 return lambda *args, **kwargs: self(grid, block, args, **kwargs)

File ~/.conda/envs/cupy/lib/python3.10/site-packages/cupyx/jit/_interface.py:96, in _JitRawKernel.__call__(self, grid, block, args, shared_mem, stream)
     94 result = self._cached_codes.get(in_types)
     95 if result is None:
---> 96     result = _compile.transpile(
     97         self._func,
     98         ['extern "C"', '__global__'],
     99         self._mode,
    100         in_types,
    101         _cuda_types.void,
    102     )
    103     self._cached_codes[in_types] = result
    105 fname = result.func_name

File ~/.conda/envs/cupy/lib/python3.10/site-packages/cupyx/jit/_compile.py:182, in transpile(func, attributes, mode, in_types, ret_type)
    180 generated = Generated()
    181 in_types = tuple(in_types)
--> 182 name, return_type = _transpile_func_obj(
    183     func, attributes, mode, in_types, ret_type, generated)
    184 func_name, _ = generated.device_function[(func, in_types)]
    185 code = '\n'.join(generated.codes)

File ~/.conda/envs/cupy/lib/python3.10/site-packages/cupyx/jit/_compile.py:210, in _transpile_func_obj(func, attributes, mode, in_types, ret_type, generated)
    207     name += '_' + str(len(generated.device_function))
    208 generated.device_function[(func, in_types)] = None
--> 210 cuda_code, env = _transpile_function(
    211     tree, name, attributes, mode, consts,
    212     in_types, ret_type, generated, source=source)
    214 generated.device_function[(func, in_types)] = (name, env.ret_type)
    215 generated.add_code(cuda_code)

File ~/.conda/envs/cupy/lib/python3.10/site-packages/cupyx/jit/_compile.py:311, in _transpile_function(func, name, attributes, mode, consts, in_types, ret_type, generated, source)
    308         exc.reraise(source)
    310 # Raises the error out of `except` block to clean stack trace.
--> 311 exc.reraise(source)
    312 assert False

File ~/.conda/envs/cupy/lib/python3.10/site-packages/cupyx/jit/_compile.py:55, in _JitCompileError.reraise(self, pycode)
     51 end = getattr(self.node, 'end_lineno', start)
     52 pycode = '\n'.join([
     53     (f'> {line}' if start <= i + 1 <= end else f'  {line}').rstrip()
     54     for i, line in enumerate(pycode.split('\n'))])
---> 55 raise self.error_type(self.mes + '\n\n' + pycode)

TypeError: Arguments must be constants.

  @jit.rawkernel()
  def elementwise_add(x, y, size):
      tid = jit.grid(1)
      ntid = jit.gridsize(1)

>     la2 = local_memory(int, (2,3,4))
```

Zhyrek:
The source of this appears to be in cupyx.jit._internal_types:

```
class BuiltinFunc(Expr):
    # subclasses must implement:
    # - either call or call_const
    # - `__call__` with a correct signature, which calls the parent's __call__

    def call(self, env: 'Environment', *args, **kwargs) -> Expr:
        for x in itertools.chain(args, kwargs.values()):
            if not isinstance(x, Constant):
                raise TypeError('Arguments must be constants.')
        args = tuple([x.obj for x in args])
        kwargs = dict([(k, v.obj) for k, v in kwargs.items()])
        return self.call_const(env, *args, **kwargs)
```

Zhyrek:
Oh! This is my mistake. Turns out the version I was using for testing was 11.3.0, which did not natively support tuple arguments (hence the Data blocks). I've been migrating to a new environment for ease of testing, and am now on 12.2.0, where the above code shows its actually a constant the whole time.

`(<Constant obj = "<class 'int'>">, <Constant obj = "(2, 3, 4)">)`
`(<Constant obj = "<class 'int'>">, <Constant obj = "(2, 3, 4)">)`

Because of this, the call method for local_memory is indeed unneeded.

Zhyrek:
Summary of changes:

- Local memory properly uses a contiguous block of memory, and should only use 1 register by initializing the variable name directly.
  - This is done by intercepting the declvar method of LocalMem with a comment, and instead using the declvar of the local_memory ctype (ContiguousArray) to emit just the array initialization code during assignment, and not the `= rvalue` part. Ultimately did not have to modify any of `cupyx.jit._compile`
- ndarray local_memory can be indexed properly (e.g. `array[1][2][4]` can be done)
- local_memory is imported in `cupyx.jit.__init__.py`
- Removed the call method from local_memory, as its unneeded in 12.2.0
- Two new tests, one to test creation of new variables based on indexing local_memory arrays and pointers to these arrays, and one to test passing (1d and 3d) local_memory arrays to device functions.

Zhyrek:
Also more of a formatting thing to ponder, for numpy/cupy arrays the convention for argument order is (shape, dtype), while for shared_memory (which this is based on), the order is (dtype, shape). Would it be worth the small(er) pain now to unify the conventions and change shared_memory and local_memory to use the ordering (shape, dtype)?

Zhyrek:
Small add-on, I originally found that higher-order pointers in the `test_local_memory_assignment` test wouldn't work. However, in order to pass higher-order pointers to the device function in `test_local_memory_in_device_functions` test, I ended up fixing this issue, so higher-order pointer creation in the first test is back in, along with an additional test ensuring creation of pointers to these pointers works. E.g.:

```
lmem = jit.local_memory(numpy.int32, (2, 2, 2, 2))
pointer2 = lmem[1]
pointer3 = pointer2[1]
pointer3[0][1] = 6
```
This properly sets the cell lmem[1][1][0][1] equal to 6

cupyx/jit/_builtin_funcs.py
```diff
@@ -219,6 +219,34 @@ def call_const(self, env, dtype, size, alignment=None):
...
 
+class LocalMemory(BuiltinFunc):
+
+    def __call__(self, dtype, size, alignment=None):
+        """Allocates local memory and returns it as a 1-D array.
+
+        Args:
+            dtype (dtype):
+                The dtype of the returned array.
+            size (int or tuple):
```
asi1024, line 230:
```suggestion
            size (int or tuple of ints):
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -219,6 +219,34 @@ def call_const(self, env, dtype, size, alignment=None):
...
+        """
+        super().__call__()
+
+    def call_const(self, env, dtype, size, alignment=None):
+        name = env.get_fresh_variable_name(prefix='_lmem')
+        ctype = _cuda_typerules.to_ctype(dtype)
+        var = Data(name, _cuda_types.LocalMem(ctype, size, alignment))
+        env.decls[name] = var
+        env.locals[name] = var
+        if(type(size) is int):
```
asi1024, line 244: The following code raises `TypeError`.
```py
@jit.rawkernel()
def f():
    lmem = jit.local_memory(numpy.int32, numpy.int64(10))
```

```suggestion
        if not isinstance(size, tuple):
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -219,6 +219,34 @@ def call_const(self, env, dtype, size, alignment=None):
...
+            size = (size,)
+        return_ctype = ctype
+        for i in range(len(size)):
+            return_ctype = _cuda_types.ContiguousArray(return_ctype, size[len(size)-(i+1):])
+        return Data("!!lmem", return_ctype) #non-valid name in order to intercept creation of this
```
asi1024, line 249: The following code exposes CUDA error.
```py
@jit.rawkernel()
def f():
    lmem = jit.local_memory(numpy.int32, (2, 3, 4)) + 10
```

```
cubin.cu(7): error: a value of type "int" cannot be used to initialize an entity of type "int (*)[3][4]"
    int (*lmem)[3][4] = (!!lmem + 10);
```

Zhyrek: Hey, I recognize for this case it is non-ideal because it exposes the !!lmem token designed to catch the creation of local_memory, but to be honest I wouldn't have expected this code to work in the first place. Testing a similar expression in numba doesn't work either:

```
import numba
from numba import cuda, int32

@cuda.jit
def test():
    a = cuda.local.array((2,3,4), int32) + 10
    
test[256, 256]()
```
```
No implementation of function Function(<built-in function add>) found for signature:
 
 >>> add(array(int32, 3d, C), Literal[int](10))
```

Would it be desired to catch this particular case? I could imagine it could be a syntactic way to initialize the array, by generating the following for the above code:
`int lmem[2][3][4] = {10};`

However, this would almost certainly involve modifying something in cupyx.jit._compile to catch this edge case

Thanks again for all your help with this by the way! :)

asi1024: In case of shared memory, it generates a code that adds offset to the pointer.

```py
@jit.rawkernel()
def f():
    smem = jit.shared_memory(numpy.int32, 10) + 1
```

```cpp
extern "C" __global__ void f() {
   __shared__ int _smem1[10];
  int* smem = (_smem1 + 1);
}
```

asi1024:
> Also more of a formatting thing to ponder, for numpy/cupy arrays the convention for argument order is (shape, dtype), while for shared_memory (which this is based on), the order is (dtype, shape). Would it be worth the small(er) pain now to unify the conventions and change shared_memory and local_memory to use the ordering (shape, dtype)?

As your comment suggests, we may indeed need to fix the interface. However, I think it would be better to create another pull request for discussions, and fix it with shared memory,

cupyx/jit/_builtin_funcs.py
leofang: Interesting. If we wrap shared/local memory in an array-like fashion, I'd expect users to index over the arrays, instead of computing the pointer offset this way, which is only kinda intuitive for 1D arrays. Semantics-wise, adding an array with a constant should result in a broadcasting, i.e.
```py
smem = jit.local_memory(numpy.int32, (2,3)) + 1
```
is equivalent to
```c
int smem[2][3];
for (auto i=0; i<2; i++)
  for (auto j=0; j<3; j++)
    smem[i][j] += 1
```

Zhyrek: This can be done in one line as `int smem[2][3] = {1};` I believe, the brace notation with a single value inside should broadcast to the entire array when its created. (Also has the advantage of ensuring the array is indeed guaranteed to be initialized to some value, I've run into an issue in the past where a local array created using `int lmem[10]` had spurious non-zero values sprinkled throughout!)

leofang: You missed the plus sign 🙂 Here, it's not just about initializing the array, it's about all possible, broadcastable arithmetic operations that NumPy/CuPy would permit. Interpreting the math operators as just calculating pointer offsets kills this possibility, and I want to make sure this is a fully-aware decision.

Zhyrek: Done a fair amount of thinking, and ultimately I agree that the syntax should follow python convention rather than pointer offsets. The more pythonic the syntax, the more people who will be able to just pick it up and use it. And cupy is python-based anyways...

However in this case, would it not make more sense to implement the following *as* local_memory to maximize the overlap in code?

```
@jit.rawkernel
def test():
    smem = cupy.zeros([2, 3])+1
```
(e.g.: if cupy.zeros is called by the CPU, it creates a device_array, if it's called by the GPU within a @jit.rawkernel decorated function, it creates a local_memory array)

It could lead to some confusion, but this seems like a natural extension

Zhyrek: For reference, my main use case for cupy is to create a "parallel for" loop of the form:

```
@jit.rawkernel
def kernel(array):
    tid = jit.grid(1)
    ntid = jit.gridsize(1)
    for i in range(tid, array.shape[0], ntid):
        #do stuff at each index in the array
```

Essentially an elementwise kernel, but with an index so each thread can potentially access neighboring data (partial differential equation simulations). 

However, this use case is relatively independent of the choice of going with pythonic syntax, or sticking with c, so as long as a local memory array exists in some capacity, I don't have much skin in the game :)

Zhyrek: Also just learned I was wrong, the syntax `int lmem[2][2] = {1};` apparently will only initialize the first element of the top-level array to one, creating `{{1, 1}, {0, 0}}`!

leofang: > Also just learned I was wrong

Yeah, C++ is hard and full of surprises.

> However in this case, would it not make more sense to implement the following _as_ local_memory to maximize the overlap in code?
> 
> ```
> @jit.rawkernel
> def test():
>     smem = cupy.zeros([2, 3])+1
> ```
> 
> (e.g.: if cupy.zeros is called by the CPU, it creates a device_array, if it's called by the GPU within a @jit.rawkernel decorated function, it creates a local_memory array)
> 
> It could lead to some confusion, but this seems like a natural extension
> 
> (And for shared memory, potentially a flag could be added? `shared_mem = cupy.zeros([2,3], shared_memory=True)`)
> 
> This idea could be extended to all cupy functions: if called by the CPU, it runs a batched algorithm on a device_array. if called by the GPU, it runs the algorithm as if it was the corresponding numpy algorithm, but within a GPU kernel.

I think this is way out of scope for CuPy JIT, let's not overthink here 😅 I agree for your use case as long as `local_memory`/`shared_memory` are present, you're unblocked. The question I raised was mainly about their semantics (are they pythonic arrays, or should they be treated as pointers in disguise?), I didn't mean bringing any of the current host capabilities to `jit.rawkernel`.

asi1024: We don't want to allow element-wise operations in this notation in CuPy JIT, because it behaves defferently from CuPy ndarray routines in terms of parallelism.

leofang: > because it behaves defferently from CuPy ndarray routines in terms of parallelism.

For shared memory, this syntax ambiguity is indeed a concern. For local memory, it's more like a syntactic sugar (see https://github.com/cupy/cupy/pull/7976#discussion_r1399454937). But it's better that we stick to the lowest common denominator and do not consider it.

Zhyrek: To be completely honest, I think the path of least resistance for this one is to tell users "don't do that". 

For the pointer offset, regardless of how you initialize the problem, you end up with two registers, one for the original array, and one for the offset array. May as well just initialize it as:

```
original_pointer = jit.local_memory(numpy.int32, 10)
offset_pointer = original_pointer + 1
```

For sending the array to a function upon creation, you wouldn't even have a variable handle for the array you've created in this case. Would it not exclusively be better to write it as the following (where you would have a variable to manipulate the array afterwards?):

```
@jit.rawkernel(device=True)
def g(array):
    pass

@jit.rawkernel()
def f():
    array = jit.local_memory(numpy.int32, 10)
    g(array)

f[1, 1]()
```

Zhyrek: Was mainly just wondering if that would be alright :) I'll get right on it!

Zhyrek: Oh! In that case this whole business with trying to catch the creation of the local memory array is moot. I'll rework it back to match the syntax of shared memory, in which case it should be able to do pointer offsets and submit to a function as-is like shared memory can.

Zhyrek:
As a follow-up to the conversation about whether to use pointer offsets or broadcasting for adding to a local_memory array, I think I have a strong justification for the former. A near 1-to-1 translation of fortran LAPACK code into a device function would rely on being able to send offset pointers to one block of working memory. I personally need linalg functions inside a kernel (many small lstsq/pinv calls inside an iterative loop), and this could be one way to do so that might have additional utility for others!

cupyx/jit/_cuda_types.py
```diff
@@ -253,6 +275,29 @@ def __str__(self) -> str:
         return f'{self.child_type}*'
 
 
+class ContiguousArray(PointerBase):
+    # used to define ndarray of local memory, in <dtype>[3][4] format
```
asi1024, line 279: Could you fix class name and comments?
```suggestion
class SizedArray(PointerBase):
    # used to define C-array type, in <type>[dim0][dim1] format
```

cupyx/jit/_cuda_types.py
```diff
@@ -253,6 +275,29 @@ def __str__(self) -> str:
...
+    def __init__(self, child_type: TypeBase, size: Tuple[int, ...]) -> None:
```
asi1024, line 280: Duplicate child dimension lengths are stored in `child_type`. Could you fix `__init__` to only take the outermost length?
```suggestion
    def __init__(self, child_type: TypeBase, size: int) -> None:
```

cupyx/jit/_cuda_types.py
```diff
@@ -253,6 +275,29 @@ def __str__(self) -> str:
...
+        self._index_32_bits = True
+
+    def declvar(self, x: str, init: Optional['Data']) -> str:
+        s = f'{self.base_type} (*{x})'
+        for i in range(1, len(self._size)):
+            s += f'[{self._size[i]}]'
+        if (init is None):
+            return s
+        else:
+            return f'{s} = {init.code}'
```
asi1024, line 298: This style of initialization is allowed only if the array size is 1.

asi1024: Please check the type of `init` is compatible with `base_type`.

cupyx/jit/_cuda_types.py
```diff
@@ -253,6 +275,29 @@ def __str__(self) -> str:
...
+    def __init__(self, child_type: TypeBase, size: Tuple[int, ...]) -> None:
+        self.base_type = child_type
+        if (type(self.base_type) is ContiguousArray):
+            self.base_type = child_type.base_type
+
+        super().__init__(child_type)
+        self._size = size
+        self.dtype = child_type
+        self._c_contiguous = True
+        self._index_32_bits = True
```
asi1024, line 289: Unused members.
```suggestion
```

cupyx/jit/_builtin_funcs.py
```diff
@@ -220,6 +220,39 @@ def call_const(self, env, dtype, size, alignment=None):
...
+        """
+        super().__call__()
+
+    def call_const(self, env, dtype, size, alignment=None):
+        name = env.get_fresh_variable_name(prefix='_lmem')
+        ctype = _cuda_typerules.to_ctype(dtype)
+        if (type(size) is int):
+            size = (size,)
+        if not (isinstance(size, tuple)):
+            raise 'size of local_memory must be int, or a tuple of ints'
```
asi1024, line 245:
```suggestion
        if not isinstance(size, tuple):
            size = (size,)
```

PR #7977, opened 2023-11-08
Add debug feature to preloading and softlink

kmaehashi:
This PR adds the new environment variable `CUPY_DEBUG_LIBRARY_LOAD` which allows users to diagnose issues related to library load. This was originally proposed in #7929.


emcastillo (approved):
LGTM!

PR #7985, opened 2023-11-21
Add 2D Delaunay triangulation

andfoy:
This PR intends to add a two dimensional Delaunay triangulation in the `cupyx.scipy.spatial` namespace. This is required as part of https://github.com/cupy/cupy/issues/7186, where some interpolants such as `CloughTocher2DInterpolator` require them.

This implementation is derived from https://doi.org/10.1145/2556700.2556710, where the implementation is reduced from 3D to 2D.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,3 @@ htmlcov/
 .pytest_cache/
 cupy/.data/
 cupy/cuda/thrust.h
-.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ htmlcov/
 .pytest_cache/
 cupy/.data/
 cupy/cuda/thrust.h
+.vscode/
--- a/cupyx/scipy/spatial/_delaunay.py
+++ b/cupyx/scipy/spatial/_delaunay.py
@@ -1,0 +1,158 @@
+
+import cupy
+from cupyx.scipy.spatial.delaunay_2d._tri import GDel2D
+
+
+class Delaunay:
+    """
+    Delaunay(points, furthest_site=False, incremental=False)
+
+    Delaunay tessellation in 2 dimensions.
+
+    Parameters
+    ----------
+    points : ndarray of floats, shape (npoints, ndim)
+        Coordinates of points to triangulate
+    furthest_site : bool, optional
+        Whether to compute a furthest-site Delaunay triangulation. This option
+        will be ignored, since it is not supported by CuPy
+        Default: False
+    incremental : bool, optional
+        Allow adding new points incrementally. This takes up some additional
+        resources. This option will be ignored, since it is not supported by
+        CuPy. Default: False
+
+    Attributes
+    ----------
+    points : ndarray of double, shape (npoints, ndim)
+        Coordinates of input points.
+    simplices : ndarray of ints, shape (nsimplex, ndim+1)
+        Indices of the points forming the simplices in the triangulation.
+        For 2-D, the points are oriented counterclockwise.
+    neighbors : ndarray of ints, shape (nsimplex, ndim+1)
+        Indices of neighbor simplices for each simplex.
+        The kth neighbor is opposite to the kth vertex.
+        For simplices at the boundary, -1 denotes no neighbor.0
+    vertex_to_simplex : ndarray of int, shape (npoints,)
+        Lookup array, from a vertex, to some simplex which it is a part of.
+        If qhull option "Qc" was not specified, the list will contain -1
+        for points that are not vertices of the tessellation.
+    convex_hull : ndarray of int, shape (nfaces, ndim)
+        Vertices of facets forming the convex hull of the point set.
+        The array contains the indices of the points belonging to
+        the (N-1)-dimensional facets that form the convex hull
+        of the triangulation.
+
+        .. note::
+
+           Computing convex hulls via the Delaunay triangulation is
+           inefficient and subject to increased numerical instability.
+           Use `ConvexHull` instead.
+    coplanar : ndarray of int, shape (ncoplanar, 3)
+        Indices of coplanar points and the corresponding indices of
+        the nearest facet and the nearest vertex.  Coplanar
+        points are input points which were *not* included in the
+        triangulation due to numerical precision issues.
+
+        If option "Qc" is not specified, this list is not computed.
+
+        .. versionadded:: 0.12.0
+    vertex_neighbor_vertices : tuple of two ndarrays of int; (indptr, indices)
+        Neighboring vertices of vertices. The indices of neighboring
+        vertices of vertex `k` are ``indices[indptr[k]:indptr[k+1]]``.
+    furthest_site
+        True if this was a furthest site triangulation and False if not.
+
+    Notes
+    -----
+    This implementation makes use of GDel2D to perform the triangulation in 2D.
+    See [1]_ for more information.
+
+    References
+    ----------
+    .. [1] A GPU accelerated algorithm for 3D Delaunay triangulation (2014).
+    Thanh-Tung Cao, Ashwin Nanjappa, Mingcen Gao, Tiow-Seng Tan.
+    Proc. 18th ACM SIGGRAPH Symp. Interactive 3D Graphics and Games, 47-55.
+    """
+
+    def __init__(self, points, furthest_site=False, incremental=False):
+        if points.shape[-1] != 2:
+            raise ValueError('Delaunay only supports 2D inputs at the moment.')
+
+        if furthest_site:
+            raise ValueError(
+                'furthest_site argument is not supported by CuPy.')
+
+        if incremental:
+            raise ValueError(
+                'incremental argument is not supported by CuPy.')
+
+        self.points = cupy.array(points, cupy.float64, copy=True)
+        # self._points = cupy.unique(self._points, axis=0)
+        self._triangulator = GDel2D(self.points)
+        self.simplices, self.neighbors = self._triangulator.compute()
+
+    def _find_simplex_coordinates(self, xi, eps, find_coords=False):
+        """
+        Find the simplices containing the given points.
+
+        Parameters
+        ----------
+        xi : ndarray of double, shape (..., ndim)
+            Points to locate
+        eps: float
+            Tolerance allowed in the inside-triangle check.
+        find_coords: bool, optional
+            Whether to return the barycentric coordinates of `xi`
+            with respect to the found simplices.
+
+        Returns
+        -------
+        i : ndarray of int, same shape as `xi`
+            Indices of simplices containing each point.
+            Points outside the triangulation get the value -1.
+        c : ndarray of float64, same shape as `xi`, optional
+            Barycentric coordinates of `xi` with respect to the enclosing
+            simplices. Returned only when `find_coords` is True.
+        """
+        out = cupy.empty((xi.shape[0],), dtype=cupy.int32)
+        c = cupy.empty(0, dtype=cupy.float64)
+        if find_coords:
+            c = cupy.empty((xi.shape[0], xi.shape[-1] + 1), dtype=cupy.float64)
+
+        out, c = self._triangulator.find_point_in_triangulation(
+            xi, eps, find_coords)
+
+        if find_coords:
+            return out, c
+        return out
+
+    def find_simplex(self, xi, bruteforce=False, tol=None):
+        """
+        Find the simplices containing the given points.
+
+        Parameters
+        ----------
+        xi : ndarray of double, shape (..., ndim)
+            Points to locate
+        bruteforce : bool, optional
+            Whether to only perform a brute-force search. Not used by CuPy
+        tol : float, optional
+            Tolerance allowed in the inside-triangle check.
+            Default is ``100*eps``.
+
+        Returns
+        -------
+        i : ndarray of int, same shape as `xi`
+            Indices of simplices containing each point.
+            Points outside the triangulation get the value -1.
+        """
+        if tol is None:
+            eps = 100 * cupy.finfo(cupy.double).eps
+        else:
+            eps = float(tol)
+
+        if xi.shape[-1] != 2:
+            raise ValueError('Delaunay only supports 2D inputs at the moment.')
+
+        return self._find_simplex_coordinates(xi, eps)
--- a/cupyx/scipy/spatial/delaunay_2d/_kernels.py
+++ b/cupyx/scipy/spatial/delaunay_2d/_kernels.py
@@ -339,6 +339,32 @@ INLINE_H_D PairType getTriPairType( char c )
     return (PairType) (( c >> 2 ) & 3);
 }
 
+enum CheckDelaunayMode
+{
+    CircleFastOrientFast,
+    CircleExactOrientSoS
+};
+
+enum ActTriMode
+{
+    ActTriMarkCompact,
+    ActTriCollectCompact
+};
+
+/////////////////////////////////////////////////////////////////////// Side //
+
+enum Side
+{
+    SideIn   = -1,
+    SideZero = +0,
+    SideOut  = +1
+};
+
+INLINE_H_D Side cicToSide( RealType det )
+{
+    return ( det < 0 ) ? SideOut : ( ( det > 0 ) ? SideIn : SideZero );
+}
+
 template< typename T >
 __forceinline__ __device__ void cuSwap( T& v0, T& v1 )
 {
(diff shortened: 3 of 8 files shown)


andfoy:
The following numbers correspond to the benchmark results when `Delaunay` is called with varying number of points. The gDel2D implementation has a highmark improvement of 55x over SciPy and QHull.

<details>

```python
from itertools import product

import cupy as cp
import numpy as np
from cupy import testing

from cupyx.scipy.spatial import Delaunay as GPUDelaunay
from scipy.spatial import Delaunay as CPUDelaunay

from tqdm import tqdm


# input_sizes = [5, 10, 100, 1000, 5000]
input_sizes = [100, 1000, 5000, 10000, 50000, 100000]
modules = [(cp, GPUDelaunay, 'CuPy'), (np, CPUDelaunay, 'SciPy')]
kwargs = []
# kwargs = [('order', [1, 2, 4, 6, 10])]
measurement = {}

def gather_time(prof):


def input_creation(xp, fn, size, **kwargs):
    points = testing.shaped_random(size + (2,), xp, xp.float64)
    return fn, points


def closure(fn, sig):
    return fn(sig)


for kwarg_comb in kwargs:
    kwarg_comb = dict(kwarg_comb)
        for xp, cls, _id in modules:


lines = []
for kwarg_comb in kwargs:
    lines.append('| Size | CuPy (ms) | SciPy (ms) | Speedup |')
    lines.append('|:----:|:---------:|:----------:|:-------:|')
    for size in input_sizes:
        comp = f'{size}'
        times = []
        for _, _, _id in modules:
            time = size_measurement[_id]
            times.append(time)
            time_info = f'{time:3f}'
        speedup = times[1] / times[0]
        lines.append(f'| {comp} |')
    lines.append('\n')


print('\n'.join(lines))
```
</details>
| Size | CuPy (ms) | SciPy (ms) | Speedup |
|:----:|:---------:|:----------:|:-------:|
| 100 | 3.223754 | 0.372784 | 0.115637 |
| 1000 | 4.865109 | 2.848961 | 0.585590 |
| 5000 | 5.818882 | 15.383042 | 2.643642 |
| 10000 | 6.819357 | 33.797787 | 4.956155 |
| 50000 | 11.550030 | 227.356258 | 19.684474 |
| 100000 | 16.944136 | 517.125425 | 30.519434 |
| 500000 | 59.798636 | 3322.136601 | 55.555391 |

cupyx/scipy/spatial/_delaunay.py
```diff
@@ -0,0 +1,83 @@
...
+        .. versionadded:: 1.4.0
+    """
+
+    def __init__(self, points, furthest_site=False, incremental=False):
+        if points.shape[-1] != 2:
+            raise ValueError('Delaunay only supports 2D inputs at the moment.')
+
+        if furthest_site:
+            warnings.warn(
+                'furthest_site argument is not supported by CuPy. Ignoring')
```
emcastillo, line 76: I think it is better to hard error in this situation. The user may not notice the warning and get wrong results

cupyx/scipy/spatial/_delaunay.py
```diff
@@ -0,0 +1,83 @@
...
+
+        if incremental:
+            warnings.warn(
+                'incremental argument is not supported by CuPy. Ignoring')
```
emcastillo, line 80: ditto

emcastillo:
The memory pool that you include here is likely to conflict with CuPy one, I think you can use the C apis to get memory directly from the CuPy pool when needed, maybe it is better to try merging https://github.com/cupy/cupy/pull/7210 first!

andfoy:
@emcastillo the potential to mistranslation related to this feature can be high, which can potentially delay the implementation by several months, especially when porting over headers and functions that are 20+ years old.

Also, there are some implementation-specific data structures that depend directly on memory management, which can incur in extra copies or other kinds of overheads that might reduce the overall performance.

@ev-br what do you think about this proposition?

andfoy:
Can't this feature be left as an opt-opt at compilation time somehow?

ev-br:
Indeed, was it the original approach? Which showed it's a non-trivial amount of work with the result being worse for maintenance:
- code provenance is worse (as is, it's an external library, vendored verbatim)
- reworking the library is error-prone.

How bad is the compile time hit?

Or is the suggestion to rename the files from `.cu` to something else to hide them from the build system, and read the sources into RawKernels?

andfoy:
The problem of using RawKernel is that automatically you are constraining the host code to be called from Python. This library has several host calls from C++, hence you would need to translate those.

ev-br:
Here's a small non-scientific experiment to see what is the impact on build times and the wheel size:

```
$ git co main
Already on 'main'
Your branch is up to date with 'upstream/main'.

$ time python setup.py bdist_wheel

<snip>

real	11m10.097s
user	43m56.192s
sys	0m37.035s

$ ll dist/
total 67680
drwxrwxr-x  2 ev-br ev-br     4096 Dec 20 13:16 ./
drwxrwxr-x 18 ev-br ev-br     4096 Dec 20 13:16 ../
-rw-rw-r--  1 ev-br ev-br 69293662 Dec 20 13:16 cupy-13.0.0rc1-cp310-cp310-linux_x86_64.whl


===============================================================================

$ git co edgar/add_delaunay_2d
Note: switching to 'edgar/add_delaunay_2d'.

<snip>

$ git clean -xdf

$ time python setup.py bdist_wheel

<snip>
real	11m41.849s
user	51m53.558s
sys	0m52.472s

$ ll dist
total 73988
drwxrwxr-x  2 ev-br ev-br     4096 Dec 20 13:38 ./
drwxrwxr-x 18 ev-br ev-br     4096 Dec 20 13:38 ../
-rw-rw-r--  1 ev-br ev-br 75751275 Dec 20 13:38 cupy-13.0.0rc1-cp310-cp310-linux_x86_64.whl
```

So --- at least on this specific machine --- the build time increases by half a minute (5%) and the wheel size increases by 6Mb (10%).

asi1024:
Although merging only this pull request may not significantly impact build time, we are concerned that if we accept more PRs like this the build time will keep increasing. Allowing this kind of implementation sets a precedent and it will be hard to reject similar PRs if this one is accepted as is.

I believe that a Python-only implementation utilizing CuPy arrays is the most straightforward and maintainable approach. Would you be willing to try to do it this way? we can provide assistance if needed since we understand the changes may be big.

andfoy:
These are the new performance results as of 9ecce11, indeed the initial version (which has a 100% correspondence w.r.t the original C++ implementation) has a drop in performance.

@emcastillo @asi1024 do you think the following numbers are enough? If not, please let me know and I'll optimize as much as possible. The performance drop is about 5x for small cases (i.e., 5000) and 2x for large cases.

<details>

```python
from itertools import product

import cupy as cp
import numpy as np
from cupy import testing

from cupyx.scipy.spatial.delaunay_2d._kernels import DELAUNAY_MODULE
from cupyx.scipy.spatial import Delaunay as GPUDelaunay
from scipy.spatial import Delaunay as CPUDelaunay

from tqdm import tqdm

DELAUNAY_MODULE.compile()

# input_sizes = [5, 10, 100, 1000, 5000]
input_sizes = [100, 1000, 5000, 10000, 50000, 100000, 500000]
# input_sizes = [100000]
modules = [(cp, GPUDelaunay, 'CuPy'), (np, CPUDelaunay, 'SciPy')]
kwargs = []
# kwargs = [('order', [1, 2, 4, 6, 10])]
measurement = {}

def gather_time(prof):


def input_creation(xp, fn, size, **kwargs):
    points = testing.shaped_random(size + (2,), xp, xp.float64)
    return fn, points


def closure(fn, sig):
    return fn(sig)


for kwarg_comb in kwargs:
    kwarg_comb = dict(kwarg_comb)
        for xp, cls, _id in modules:


lines = []
for kwarg_comb in kwargs:
    lines.append('| Size | CuPy (ms) | SciPy (ms) | Speedup |')
    lines.append('|:----:|:---------:|:----------:|:-------:|')
    for size in input_sizes:
        comp = f'{size}'
        times = []
        for _, _, _id in modules:
            time = size_measurement[_id]
            times.append(time)
            time_info = f'{time:3f}'
        speedup = times[1] / times[0]
        lines.append(f'| {comp} |')
    lines.append('\n')


print('\n'.join(lines))
```
</details>


| Size | CuPy (ms) | SciPy (ms) | Speedup |
|:----:|:---------:|:----------:|:-------:|
| 100 | 14.506556 | 0.377294 | 0.026009 |
| 1000 | 23.978261 | 2.838406 | 0.118374 |
| 5000 | 30.043801 | 14.882599 | 0.495363 |
| 10000 | 32.431401 | 33.145071 | 1.022006 |
| 50000 | 36.745877 | 218.159636 | 5.936983 |
| 100000 | 42.818693 | 533.102947 | 12.450239 |
| 500000 | 83.101113 | 3102.933850 | 37.339257 |

asi1024:
@andfoy The performance degradation is a bit higher than we expected. This degradation may be due to the longer CPU execution time of Python compared to the GPU execution time. Do you have any idea of what routines could be a bottleneck for CPU time? We can try to provide help to fix them or do it in a follow up PR :)

cupyx/scipy/spatial/delaunay_2d/_tri.py
```diff
@@ -0,0 +1,570 @@
...
+            self.n_points - 1, self.point_vec, self.points_idx,
+            self.pred_consts)
+
+        self.available_points = self.n_points - 4
+        self.tri_num = 4
+
+    def _init_for_flip(self):
+        min_val = self.points.min()
+        max_val = self.points.max()
+        range_val = max_val - min_val
```
andfoy, line 160: This is marked as a potential bottleneck after running the profiler

cupyx/scipy/spatial/_delaunay.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if furthest_site:
+            raise ValueError(
+                'furthest_site argument is not supported by CuPy.')
+
+        if incremental:
+            raise ValueError(
+                'incremental argument is not supported by CuPy.')
+
+        self.points = cupy.array(points, cupy.float64, copy=True)
```
andfoy, line 90: This copy might be also a bottleneck

cupyx/scipy/spatial/delaunay_2d/_tri.py
```diff
@@ -0,0 +1,570 @@
...
+
+        # Sort the points spatially according to their Morton numbers
+        get_morton_number(self.points, self.n_points - 1, min_val,
+                          range_val, self.values)
+
+        self.values[-1] = 2 ** 31 - 1
+        unique_values, unique_index = cupy.unique(
+            self.values, return_index=True)
```
andfoy, line 168: This unique is also a bottleneck. This is an issue with the original implementation regarding duplicate points, at some point in the reference C++ implementation looped infinitely when trying to insert duplicate points. I haven't tried to remove it

andfoy:
@asi1024, we also suspect that the CPU execution is taking longer after porting the implementation to Python. In particular, I suspect that the major bottleneck occurs in the `_init_for_flip` () method, which calls `max`,  `min` and `unique`, that can be seen on this profiling flamegraph, from the 6.86s to the 7.00s mark. Another significant bottleneck relates to copying the values to a new array in order to prevent modifications, this can be also removed, given that the points are being copied into another array in order to do the calculations (no need to double copy).

Both of these regions account for at least 50% of the computation time.

imagen

These are the benchmark results when removing the `unique` and the `copy` detailed above. However, there is not much of an improvement (3x).

| Size | CuPy (ms) | SciPy (ms) | Speedup |
|:----:|:---------:|:----------:|:-------:|
| 100 | 19.535612 | 0.530353 | 0.027148 |
| 1000 | 32.956266 | 3.878908 | 0.117699 |
| 5000 | 41.986632 | 21.957715 | 0.522969 |
| 10000 | 44.652589 | 46.640094 | 1.044510 |
| 50000 | 51.745101 | 274.452174 | 5.303926 |
| 100000 | 58.017069 | 614.829551 | 10.597391 |
| 500000 | 93.568521 | 3752.015339 | 40.099120 |

This is the flamegraph after removing these two sections, there is still the max/min bottleneck. The unique bottleneck has been replaced by an `argmax` which is required in order to sort the points in a Z-curve (according to the reference paper results, sorting helped reducing the overall runtime), When unique was called, the sorted results returned were used instead of the `argmax`. Eliminating the extra copy did help a little bit, however, the main blockage occurs in `max/min`  

imagen

andfoy:
Before profiling both implementations, I reproduced the benchmark results for the C++ implementation 

| Size | CuPy (ms) | SciPy (ms) | Speedup |
|:----:|:---------:|:----------:|:-------:|
| 100 | 3.066400 | 0.387718 | 0.126441 |
| 1000 | 4.782042 | 2.825747 | 0.590908 |
| 5000 | 6.048918 | 14.925128 | 2.467405 |
| 10000 | 6.623285 | 49.202431 | 7.428705 |
| 50000 | 11.051237 | 229.440171 | 20.761493 |
| 100000 | 16.243913 | 616.848576 | 37.974136 |
| 500000 | 54.174184 | 3606.194448 | 66.566659 |

Next, we proceed to profile both implementations individually, for this, I used dedicated prints for the most critical functions in GDel, including splitting and flipping the triangles (With an input of 500000 points). Then, the values produced by both programs were compared, where the following differences are found:

<details>
<summary>Timing scripts</summary>

```c++
// timing is an std::vector<std::tuple<std::string, double>>
auto start = std::chrono::system_clock::now();

// Function body

auto end = std::chrono::system_clock::now();
std::chrono::duration<double> elapsed_seconds = end-start;
timing.push_back(std::make_tuple("_compute", elapsed_seconds.count()));

// At the end of the execution
for(int i = 0; i < timing.size(); i++) {
        std::string fnName;
        double secs;
        std::tie(fnName, secs) = timing[i];
        std::cout << fnName << ": " << secs << std::endl;
}
```

For the Python version, a simple timing decorator:
```python
def print_time(f):
    def wrapped_f(self, *args, **kwargs):
        start_time = time.time()
        result = f(self, *args, **kwargs)
        print(f'{f.__name__}: {time.time() - start_time}')
        return result
    return wrapped_f

# Function to benchmark
@print_time
def func(*args, **kwargs):
      ...
```
</details>

<details>
<summary>Comparison script</summary>

```python
import numpy as np

def gather_times(filename):
    perf_map = {}
    with open(filename, 'r') as f:
        for line in f.readlines():
            key, value = line.strip().split(':')
            key_values = perf_map.get(key, [])
            key_values.append(float(value))
            perf_map[key] = key_values
    return {k: np.asarray(perf_map[k]) for k in perf_map}


cpp_perf = gather_times('out_cpp.txt')
py_perf = gather_times('out_py.txt')
print('| Function name | Avg C++ runtime | Avg Python runtime | Avg slowdown | Avg abs diff | Avg rel diff |')
print('|:-------------:|:---------------:|:------------------:|:------------:|:------------:|:------------:|')

for func_name in py_perf:
    py_values = py_perf[func_name]
    cpp_values = cpp_perf[func_name]
    slowdown = np.mean(py_values / cpp_values).item()
    abs_diff = np.mean(np.abs(py_values - cpp_values)).item()
    rel_diff = np.mean(np.abs(py_values - cpp_values) / cpp_values).item()
    avg_cpp = cpp_values.mean().item()
    avg_py = py_values.mean().item()
    print(f'| {func_name} | {avg_cpp:.2e} | {avg_py:.2e} | {slowdown:.2f} | {abs_diff:.2e} | {rel_diff:.2e} |')

```
</details>

| Function name | Avg slowdown |
|:-------------:|:---------------:|
| _construct_initial_triangles | 4.603449 |
| _init_for_flip | 4.244037 |
| _split_tri | 2.109865 |
| _do_flipping | 6.255789 |

In this case, we can appreciate that one of the major differences occur on the `_init_for_flip/_construct_initial_triangles`, which where the max/min/sub occurs. Other sections that report differences are `_split_tri` and `_do_flipping`, which have several cupy calls besides the shared kernel implementations.

andfoy:
After further inspection, these are the top-10 calls that are causing slowdowns (compute is given as a reference, all times are in milliseconds):

| Func | Avg slowdown | Avg Py | Avg C | Avg abs diff | Time % | # Calls |
|:----:|:------------:|:------:|--------:|:------------:|:------:|:-------:|
| compute | 2.509 | 152.027 | 60.603 | 91.424 | 100.00% | 1 |
| _init_for_flip (unique/argmax) | 16.512 | 42.086 | 2.549 | 39.537 | 27.68% | 1 |
| _do_flipping (mark compact) | 18.713 | 0.687 | 0.036 | 0.651 | 18.09% | 40 |
| _do_flipping (compact_if_neg) | 11.294 | 0.449 | 0.035 | 0.414 | 14.48% | 49 |
| _split_tri (split_tri_vec) | 22.117 | 0.824 | 0.037 | 0.787 | 9.21% | 17 |
| _do_flipping (collect compact) | 5.995 | 0.182 | 0.030 | 0.151 | 3.11% | 26 |
| _split_tri (ins_tri_map) | 9.238 | 0.187 | 0.031 | 0.156 | 2.09% | 17 |
| _construct_initial_triangles (min/max/x) | 8.331 | 2.003 | 0.240 | 1.763 | 1.32% | 1 |
| _do_flipping (flip) | 2.486 | 0.039 | 0.085 | 0.080 | 1.23% | 48 |
| _do_flipping (tri_vote_vec) | 3.273 | 0.037 | 0.018 | 0.020 | 1.18% | 49 |
| _do_flipping (update_opp) | 1.653 | 0.030 | 0.081 | 0.072 | 0.95% | 48 |


As we previously discussed, the call to `unique/argmax` is causing most of the performance damage in this case, given that it is a single call that accounts for almost 30% of the total `compute` time and is about 16x slower than `thrust::sort_by_key` and `thrust::unique_by_key`.  I don't know if there is a way to optimize this call, which could yield the biggest gains.

Next, the calls to `mark compact` and `compact_if_neg` consist on applying a boolean mask to a index tensor:

```python
# Mark compact
# This yields a boolean mask
active_tri = is_tri_active(self.triangle_info)

# Get the indices that comply with the mask
self._act_tri = cupy.arange(0, active_tri.shape[0])
self._act_tri = self._act_tri[active_tri].astype(cupy.int32)

# Compact if neg
# Filter the values that are greater than zero
flip_to_tri = flip_to_tri[flip_to_tri >= 0]
```

Once again, any suggestion is welcome in order to improve this kind of operations.

cupyx/scipy/spatial/delaunay_2d/_tri.py
```diff
@@ -0,0 +1,570 @@
...
+                self.triangles, self.triangle_opp, tri_vote, exact_check_vi,
+                self.counters, self.n_points - 1, self.point_vec,
+                self.points_idx, self.pred_consts)
+
+    def _do_flipping(self, check_mode):
+        tri_num = self.triangles.shape[0]
+        if self._act_tri_mode == ActTriMode.ActTriMarkCompact:
+            active_tri = is_tri_active(self.triangle_info)
+            self._act_tri = cupy.arange(0, active_tri.shape[0])
+            self._act_tri = self._act_tri[active_tri].astype(cupy.int32)
```
andfoy, line 375: This is the `mark compact` section discussed in the benchmarks.

cupyx/scipy/spatial/delaunay_2d/_tri.py
```diff
@@ -0,0 +1,570 @@
...
+        flip_to_tri = cupy.empty(org_act_num, dtype=cupy.int32)
+        mark_rejected_flips(
+            self._act_tri, self.triangle_opp, tri_vote, self.triangle_info,
+            flip_to_tri, org_act_num)
+
+        del tri_vote
+
+        # Compact flips
+        flip_to_tri = flip_to_tri[flip_to_tri >= 0]
+        flip_num = flip_to_tri.shape[0]
```
andfoy, line 413: This is the `compact_if_neg` section discussed in the benchmark results.

andfoy:
In other updates, if `unique` is removed, then the performance drops, which is consistent with respect to the reference paper experiments

| Size | CuPy (ms) | SciPy (ms) | Speedup |
|:----:|:---------:|:----------:|:-------:|
| 100 | 16.699395 | 0.459509 | 0.027517 |
| 1000 | 29.736610 | 3.527434 | 0.118623 |
| 5000 | 33.911562 | 19.526128 | 0.575796 |
| 10000 | 38.000970 | 40.853217 | 1.075057 |
| 50000 | 43.425706 | 241.185625 | 5.553983 |
| 100000 | 49.232984 | 537.830085 | 10.924182 |
| 500000 | 116.398724 | 3173.629539 | 27.265157 |

.gitignore
```diff
@@ -24,3 +24,4 @@ htmlcov/
 .pytest_cache/
 cupy/.data/
 cupy/cuda/thrust.h
+.vscode/
```
asi1024, line 27: Could you remove unrelated change?

```suggestion
.vscode/
```

asi1024 (commented):
Thank you very much for your help! I have created an issue for performance improvement of `cupy.unique`: https://github.com/cupy/cupy/issues/8172.
LGTM ecxcepts one comment.

cupyx/scipy/spatial/delaunay_2d/_tri.py
asi1024: I think the reason these lines are slow is essentially the same as the `unique`: both `unique` and advanced indexing seem to call slow `cupy.nonzero`. 😞

cupyx/scipy/spatial/delaunay_2d/_kernels.py
```diff
@@ -0,0 +1,2713 @@
...
+    ker_split_points_exact((PRED_BLOCK_SZ,), (PRED_N_BLOCKS,), (
+        vert_tri, tri_to_vert, tri, ins_tri_map,
+        exact_check, counter, int(tri_num), int(ins_tri_num), inf_idx, points,
+        points_idx, pred_consts))
+
+
+def split_tri(split_tri, tri, tri_opp, tri_info, ins_tri_map, tri_to_vert,
+              tri_num, ins_tri_num):
+    ker_split_tri = DELAUNAY_MODULE.get_function('kerSplitTri')
+    ker_split_tri((N_BLOCKS,), (32,), (
```
asi1024, line 2553: Is this typo?

```suggestion
    ker_split_tri((N_BLOCKS,), (BLOCK_SZ,), (
```

andfoy: Yes, this is an implementation detail that comes from the reference C++ code

asi1024 (approved):

asi1024:
@andfoy LGTM. Thanks for your contribution!

Timothy-Anders0n:
Hello everyone! Just checking, this will be part of the 13.1 March release right?
I've been looking forward to @andfoy 's 2D Delaunay implementation but I'm having issues building CuPy from source so I'm waiting for its release.

asi1024:
@Timothy-Anders0n This feature will be part of v14.0.0a1 to be released in March, but will not be backported to the v13 branch.

kmaehashi:
Once v14.0.0a1 is out, you can try installing the precompilled wheel for pre-releases.
https://github.com/cupy/cupy?tab=readme-ov-file#installation

Timothy-Anders0n:
Thank you @asi1024 and @kmaehashi. Reading #7555 I thought that the next release was going to be v13.1.0 rather than v14.0.0.
Regardless, as long as this PR is in there I'll be immensely happy. :)

rgommers:
@andfoy this probably deserves an entry in `docs/source/reference/scipy_spatial.rst`, right?

andfoy:
Thanks @rgommers for catching this one! There's a PR now to expose the docs: https://github.com/cupy/cupy/pull/8294

PR #7994, opened 2023-11-27
Softlink CUDA Driver

kmaehashi:
Part of #7620. Softlink CUDA driver.


leofang:
LGTM

kmaehashi:
Hmm, only tests using constant memory are failing, both on Linux and Windows. 🤔 

```
01:11:55.935116 STDOUT 2796]	=========================== short test summary info ===========================	
01:11:55.935377 STDOUT 2796]	FAILED cupy_tests\core_tests\test_raw.py::TestRaw_param_0_{backend='nvrtc', in_memory=False}::test_const_memory - cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context	
01:11:55.935377 STDOUT 2796]	FAILED cupy_tests\core_tests\test_raw.py::TestRaw_param_1_{backend='nvrtc', in_memory=True}::test_const_memory - cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context	
01:11:55.935377 STDOUT 2796]	FAILED cupy_tests\core_tests\test_raw.py::TestRaw_param_2_{backend='nvrtc', clean_up=True, in_memory=True}::test_const_memory - cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context	
01:11:55.935377 STDOUT 2796]	FAILED cupy_tests\core_tests\test_raw.py::TestRaw_param_3_{backend='nvrtc', in_memory=False, jitify=True}::test_const_memory - cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context	
01:11:55.935377 STDOUT 2796]	FAILED cupy_tests\core_tests\test_raw.py::TestRaw_param_4_{backend='nvrtc', in_memory=True, jitify=True}::test_const_memory - cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context	
01:11:55.935377 STDOUT 2796]	FAILED cupy_tests\core_tests\test_raw.py::TestRaw_param_5_{backend='nvrtc', clean_up=True, in_memory=True, jitify=True}::test_const_memory - cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context	
01:11:55.935377 STDOUT 2796]	FAILED cupy_tests\core_tests\test_raw.py::TestRaw_param_6_{backend='nvcc', in_memory=False}::test_const_memory - cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context	
01:11:55.935377 STDOUT 2796]	= 7 failed, 105932 passed, 6917 skipped, 676 deselected, 1614 xfailed, 2558 warnings in 3234.39s (0:53:54) =	
```

The error is:

```
E   cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_INVALID_CONTEXT: invalid device context
```

emcastillo:
I could reproduce this locally :)

emcastillo:
The error seems to be happening when calling `cuModuleGetGlobal`
Other similar functions like `moduleGetFunction` work fine ...

leofang:
@kmaehashi I am thinking the treatment in `cupy_backends/cuda/api/_driver_extern.pxi` may not be appropriate. Fetching driver symbols is quite nontrivial (need to take care PTDS, different API versions, etc), which is why CUDA Python's [driver binding](https://github.com/NVIDIA/cuda-python/blob/main/cuda/_cuda/ccuda.pyx.in) is so complicated. Can we instead hide it via the import mechanism, as done for other libraries?

leofang:
(Sorry I wasn't thinking carefully yesterday morning in my first pass...)

kmaehashi:
> Fetching driver symbols is quite nontrivial (need to take care PTDS, different API versions, etc)

Thanks for the hint, Leo! I think this describes why `cuModuleGetGlobal` call fails with `CUDA_ERROR_INVALID_CONTEXT`. (`cuModuleGetGlobal` is incompatible with version of the context implicitly created by the Runtime, and `cuModuleGetGlobal_v2` should be used)

> Can we instead hide it via the import mechanism, as done for other libraries?
This sounds better, I'll try this approach.

kmaehashi:
@leofang I tried to hide driver dependency via import but it looks almost infeasible, because there are so many modules depending on driver directly or indirectly (e.g., `cupy._core.core` -> `cupy.cuda.function` -> `cupy_backends.cuda.api.driver`). Also these modules are often `cimported` for performance, and I'd avoid replacing it with `import` to avoid performance penalty on kernel launch etc.

I come up with another idea to use `cudaGetDriverEntryPoint` to retrieve driver API pointers from CUDART statically linked to CuPy. I think the only drawback is that we should drop CUDA 11.2 support, as the API is added in CUDA 11.3. WDYT?
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT_1gcf55d143722ccfa9252758181701c876

leofang:
Yeah, it should only be a build-time constraint.

What about something simpler? We check the precise symbol names that we should use (ex: `cuModuleGetGlobal_v2`), and still use softlink to fetch them?
- the diff (compared to the present PR) should be minimal
- the driver API surface that we cover is small (compared to runtime), so hopefully not much work?
- avoid the complexity of entry point
- the driver APIs we cover are stable enough, we unlikely need to update the symbol names, so it should be one-time only work

kmaehashi:
> What about something simpler? We check the precise symbol names that we should use (ex: cuModuleGetGlobal_v2), and still use softlink to fetch them?

I think that is the fastest path. Is it ok to just use the latest ones (for example `cuCtxCreate` seems to have three variants: `cuCtxCreate`, `cuCtxCreate_v2`, `cuCtxCreate_v3`) and ignore `_ptds`/`_ptsz`?

leofang:
Below are the driver symbols that we currently used, I checked with both `-cuda11x` and `-cuda12x`:
```
                 U cuCtxCreate_v2
                 U cuCtxDestroy_v2
                 U cuCtxGetCurrent
                 U cuCtxGetDevice
                 U cuCtxSetCurrent
                 U cuDevicePrimaryCtxRelease_v2
                 U cuFuncGetAttribute
                 U cuFuncSetAttribute
                 U cuGetErrorName
                 U cuGetErrorString
                 U cuLaunchCooperativeKernel
                 U cuLaunchKernel
                 U cuLinkAddData_v2
                 U cuLinkAddFile_v2
                 U cuLinkComplete
                 U cuLinkCreate_v2
                 U cuLinkDestroy
                 U cuModuleGetFunction
                 U cuModuleGetGlobal_v2
                 U cuModuleLoad
                 U cuModuleLoadData
                 U cuModuleUnload
                 U cuOccupancyMaxActiveBlocksPerMultiprocessor
                 U cuOccupancyMaxPotentialBlockSize
                 U cuStreamGetCtx
```
I think this should be a good starting point?

Re: Regarding PTDS, I think to support it needs some careful thoughts that might make us end up recreating a mini CUDA Python, so probably we don't want to do that in this PR? We can follow up after rc1 is out, if needed. To do it right, we need to:
- At build time, honor the macro (`CUDA_API_PER_THREAD_DEFAULT_STREAM`) and the option `--default-stream per-thread` passed to NVCC, but I am not sure how we can intercept these and pass to gcc, since the Cython module doesn't use NVCC. Perhaps we need users to set `CUPY_CUDA_PER_THREAD_DEFAULT_STREAM` at build time to make it easier?
- At run time, check `CUPY_CUDA_PER_THREAD_DEFAULT_STREAM` and decide which symbol to load.

kmaehashi:
Thanks Leo, added `_v2` suffix.

/test mini

leofang:
All green 🎉

emcastillo:
And in it went!

Issue #7983, opened 2023-11-21
Support cuTENSOR 2.0


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read Release Notes ([cuTENSOR](https://docs.nvidia.com/cuda/cutensor/index.html#changelog) / [cuSPARSELt](https://docs.nvidia.com/cuda/cusparselt/release_notes.html) / [cuDNN](https://docs.nvidia.com/deeplearning/cudnn/release-notes/index.html) / [NCCL](https://docs.nvidia.com/deeplearning/sdk/nccl-release-notes/index.html)).
- [ ] Fix code and CI to support the new vesrion.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`. #7984
  - **Library Installer**: Update `cupyx/tools/install_library.py`. #7997
- [x] ~Backport the above PR.~ No backport to v12.
- [ ] Fix documentation.
  - Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html)
- [x] ~Backport the above PR.~ No backport to v12.
- [ ] Implement new features using the version, if any. #7984

leofang:
cc: @anaruse

leofang:
Done in #7984.

leofang:
Realize there's still one item (docs) on the checklist, reopening... Sorry!

PR #7997, opened 2023-11-27
Use cuTENSOR 2.0 in library installer

kmaehashi:
Part of #7983. Depends on #7984.

Although it is now recommended for users to setup cuTENSOR through wheel (#7025), the library installer needs to be updated as it is still used for wheel generation in cupy-release-tools.


kmaehashi:
I noticed it's included in #7984 (https://github.com/cupy/cupy/pull/7984/commits/826634de909a34854639f3d8ad34a01887e1d93c) 😄

Issue #7990, opened 2023-11-25
CUPY sosfilt_zi some SOS coefficients are not supported


LizbethEsther:
### Description

I want to  use scipy.signal.bessel  for highpass or lowpass filtering,and then pass the sos into sosfilt_zi。If the order of the Bessel filter is greater than one, it can be calculated normally, but if it is equal to one, the ZI value will become NAN。 But this problem doesn't come up in scipy.signal.sosfilt_zi

### To Reproduce

```py
# Write the code here
```

### Installation

None

### Environment

```
# Paste the output here
```

LizbethEsther:
if self.task_dict['highpass']:
            if self.task_dict['highpass'] is True:
                n, wn = 1, 0.5
            else:
                n, wn = 1, self.task_dict['highpass']
            high_sos = signal.bessel(n, wn, 'hp', output='sos', fs=fs) 
            sos_list.append(high_sos)

        if self.task_dict['highpass'] or self.task_dict['remove_power_freq']:     
            self.sos = np.concatenate(sos_list, axis=0)
            self.sos = cp.asarray(self.sos)
            self.zi = sosfilt_zi(self.sos)   # this is cupy.sosfilt_zi and zi is [[1,1,nan,nan]] which is resulting in NAN

takagi:
Can we have the result of `python -c "import cupy; cupy.show_config()"` for smoother investigation? Thanks!

I can get the following result with this script. Would you take a look into this issue, @andfoy @ev-br?
```
sos_list = []
n, wn = 1, 0.5
high_sos = signal.bessel(n, wn, 'hp', output='sos')  # use default fs
sos_list.append(high_sos)
sos_np = np.concatenate(sos_list, axis=0)
print('scipy: ', end='')
print(sosfilt_zi_sp(sos_np))
sos_cp = cp.asarray(sos_np)
print('cupy : ', end='')
print(sosfilt_zi_cp(sos_cp))
# scipy: [[-0.5  0. ]]
# cupy: [[ 1.  1. nan nan]]
```

takagi:
Can we have the result of `python -c "import cupy; cupy.show_config()"` on your environment for smoother investigation? Thanks!

Can I ask you to take a look into this issue, @andfoy @ev-br?

LizbethEsther:
Hello, this is the function display result~









------------------&nbsp;原始邮件&nbsp;------------------
[...]

PR #7998, opened 2023-11-27
Fix issues with the initial state when a SOS filter has no IIR part

andfoy:
Fixes https://github.com/cupy/cupy/issues/7990

The issue presented by the OP raised some NaN values, which were caused when `sosfilt_zi` tried to solve the equation `0x = 0`, which was due to a lack of IIR coefficients in the example `bessel` filter. This change is analogous to this call in `lfilter`:  https://github.com/cupy/cupy/blob/ff9b39bb107864f086b7a049562d4d48a9720834/cupyx/scipy/signal/_signaltools.py#L930


takagi (approved):
LGTM!

PR #8017, opened 2023-12-06
Speed up cupy environment duplicate detection

hauntsaninja:
cupy is slow to import, taking about 740ms for me. About 300ms of that is the duplicate env check. This PR makes that check about 1.5x faster in my venv, and should scale better with larger environments and more known packages.

The trick here is that `importlib.metadata.distributions(name=name)` usually ends up doing most of the work that `importlib.metadata.distributions` does. So we might as well do that once and save.

Also note that the `.name` property access is perhaps surprisingly not super cheap because it looks up metadata, so we make sure to do it only once.

```
In [14]: %timeit cupy._environment._detect_duplicate_installation()
274 ms ± 1.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [15]: %timeit [name for name in known if list(importlib.metadata.distributions(name=name))]
272 ms ± 1.19 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [16]: %timeit [d for d in importlib.metadata.distributions() if d.name in known]
179 ms ± 265 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```


cupy/_environment.py
```diff
@@ -476,10 +476,9 @@ def _detect_duplicate_installation():
         'cupy-rocm-4-2',
         'cupy-rocm-4-3',
         'cupy-rocm-5-0',
-    ]
-    cupy_installed = [
-        name for name in known
-        if list(importlib.metadata.distributions(name=name))]
+    }
+    installed_names = {d.name for d in importlib.metadata.distributions()}
```
hauntsaninja, line 480: Fix for older Python
```suggestion
    installed_names = {d.metadata["Name"] for d in importlib.metadata.distributions()}
```

cupy/_environment.py
```diff
@@ -476,10 +476,10 @@ def _detect_duplicate_installation():
         'cupy-rocm-4-2',
         'cupy-rocm-4-3',
         'cupy-rocm-5-0',
-    ]
-    cupy_installed = [
-        name for name in known
-        if list(importlib.metadata.distributions(name=name))]
+    }
+    installed_names = {d.metadata["Name"]
+                                     for d in importlib.metadata.distributions()}
```
hauntsaninja, line 481: autopep8 but account for weird github indenting
```suggestion
    installed_names = {d.metadata["Name"]
                       for d in importlib.metadata.distributions()}
```

kmaehashi:
Hi @hauntsaninja, thanks for the PR! This looks great, I'll test & merge after the release planned today.

asi1024 (approved):

PR #8022, opened 2023-12-07
Add `cupyx.signal.pulse_compression` from cuSignal's non SciPy-compat API

takagi:
Here is a pointer to the cuSignal's original implementation of `pulse_compression`:
https://github.com/rapidsai/cusignal/blob/136453a3c8ef381f502a1edd51ca741f6c159f4a/python/cusignal/radartools/radartools.py#L28


takagi:
Thanks! Btw, we'd like to make this exceptionally backported to v13 so existing cuSignal users can get CuPy version API earlier.

PR #8034, opened 2023-12-18
Fix lfilter_zi and sosfilt_zi when any IIR coefficient is zero

### LINKED ISSUE #8028: CUPY sosfilt_zi some SOS coefficients are not supported2
### Description

 I asked this question two weeks ago and got an answer, but when I verified the effect of the fix today(I downloaded the corrected file separately for experimentation), I found that the error still exists.         https://github.com/cupy/cupy/pull/7998
from _signaltools import sosfilt_zi,sosfilt
from scipy import signal

low_sos = signal.bessel(1, 0.5, 'lowpass', output='sos', fs=250)
low_sos_cupy = cp.asarray(low_sos)
zi_cupy = sosfilt_zi(low_sos_cupy)
zi_scipy = signal.sosfilt_zi(low_sos)
print(zi_cupy)
print(zi_scipy)
data = creat_sinwav([(37,50)],fs=250,sample_time=10)   #Generate a sine wave function
filtered_cupy, dss_cupy = sosfilt(low_sos_cupy, data, axis=-1, zi=zi_cupy)  #filtered_cupy=inf……nan………………
filtered_scipy,dss_scipy = signal.sosfilt(low_sos,data,axis=-1,zi=zi_scipy)  #value

The inf value also causes my calculations to be Nan and INF， But using scipy calculations gives normal results (without inf and nan)

ps:
def sin_wave(amp=1, f=None, fs=1000, phi=0, t=1000):
    '''
    Generate a sine wave function
    :params amp:   
    :params f:    
    :params fs:  
    :params phi:  
    :params t:    
    '''
    ts = 1 / fs
    n = t / ts
    n = np.arange(n)
    y = amp * np.sin(2 * np.pi * f * n * ts + phi * (np.pi / 180))
    return y

def creat_sinwav(sin_info, fs, sample_time=200):
    datas = np.zeros(int(sample_time * fs))
    for (amp, f) in sin_info:
        datas += sin_wave(amp, f, fs=fs, t=sample_time)
    return datas

### To Reproduce

```py
# Write the code here
```

### Installation

None

### Environment

```
# Paste the output here
```

### Additional Information

_No response_

andfoy:
Fixes #8028 

This PR fixes a major issue when initial state conditions were computed for filters whose IIR coefficients contained any zero value. Such cases caused the solver to handle a singular matrix, which in turn, produced NaNs and Inf values as output. Now, a least square solution is produced whenever an IIR coefficient is zero.

This PR also fixes the order of the initial state produced by the `lfilter_zi` and `sosfilt_zi` functions, which were reversed prior to this fix. We didn't detected this error previously, since most of the time, the initial conditions are symmetrical, thus are invariant with respect to the order. However, when zero coefficients are defined, then the state will contain zeros in the same positions as the missing coefficients, therefore order does matter.

Regression tests were added in order to test for IIR coefficients that are zero.

### DIFF
--- a/cupyx/scipy/signal/_signaltools.py
+++ b/cupyx/scipy/signal/_signaltools.py
@@ -918,6 +918,7 @@ def lfilter_zi(b, a):
         y, _ = lfilter(b, a, cupy.ones(num_a + 1), zi=zi_t)
         y1 = y[:num_a]
         y2 = y[-num_a:]
+        zero_coef = cupy.where(a_r == 0)[0]
 
         C = compute_correction_factors(a_r, a_r.size + 1, a_r.dtype)
         C = C[:, a_r.size:]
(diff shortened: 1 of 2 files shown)


takagi (approved):
LGTM!

takagi:
Thanks, @andfoy!

PR #8038, opened 2023-12-19
Pass by reference

Zhyrek:
This pull request aims to add pass-by-reference capabilities for ordinary scalar variables, which by default are passed by value in c++.

The code:
```py
@jit.rawkernel(device=True)
def test2(a):
    a += 1.

@jit.rawkernel()
def test(arr):
    a = 1.2
    b = jit.ref(a)
    c = 2.2
    b = c
    test2(b)
    arr[0] = a
    
c = cp.zeros(1, dtype=np.float64)
test[1,1](c)
```
sets c[0] to 3.2, and emits the following c++ code
```
__device__ void test2_1(double &a) {
  { double &_tmp_1 = a; _tmp_1 = (_tmp_1 + 1.0); }
}
extern "C" __global__ void test(CArray<double, 1, true, true> arr) {
  double a = 1.2;
  double &b = a;
  double c = 2.2;
  b = c;
  test2_1(b);
  arr[0] = a;
}
```

A test for this is included in tests/cupyx_tests/jit_tests/test_raw.py. 

Minor issue that may need to be corrected: the first instance of a device_function being called defines whether variables are passed by reference or value, and subsequent attempts to call the device_function with different choices still use the first definition, e.g.:

```py
func(jit.ref(a))
func(a)
```
would pass a by reference to the function func both times.

### DIFF
--- a/tests/cupyx_tests/jit_tests/test_raw.py
+++ b/tests/cupyx_tests/jit_tests/test_raw.py
@@ -359,6 +359,46 @@ class TestRaw:
         assert bool((y == 15).all())
         assert bool((z == 28).all())
 
+    def test_pass_by_reference(self):
+        @jit.rawkernel(device=True)
+        def func(a, b, c):
+            a += 1
+            b = 1  # this throws a flake error, but is valid c++
+            c -= 1
+
+        @jit.rawkernel()
+        def f(x, y, z):
+            tid = jit.grid(1)
+
+            a = 4
+            b = 4
+            c = 4
+
+            if (tid == 1):
+                func(jit.ref(a), jit.ref(b), c)
+            else:
+                # func is already defined,
+                # a, b will still be passed by reference
+                # c will still be passed by value
+                func(a, b, jit.ref(c))
+
+            # still test pass-by-reference with c
+            d = jit.ref(c)
+            e = 7
+            d += e
+
+            x[tid] = a
+            y[tid] = b
+            z[tid] = c
+
+        x = cupy.zeros(2, dtype=int)
+        y = cupy.zeros(2, dtype=int)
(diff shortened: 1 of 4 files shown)


tests/cupyx_tests/jit_tests/test_raw.py
```diff
@@ -359,6 +359,46 @@ def f(x, y, z):
...
+                func(a, b, jit.ref(c))
+
+            # still test pass-by-reference with c
+            d = jit.ref(c)
+            e = 7
+            d += e
+
+            x[tid] = a
+            y[tid] = b
+            z[tid] = c
```
asi1024, line 392: I think this style of syntax is confusing. In this test code, mutating argument `b` in func is different from Python's behavior.

How about introducing `jit.pointer_of` like as follows?
```py
@jit.rawkernel(device=True)
def func(a, b, c):
    a.value += 1
    b.value = 1
    c -= 1

@jit.rawkernel()
def f(x, y, z):
    ...
    func(jit.pointer_of(a), jit.pointer_of(b), c)
    ...
```

leofang: FWIW, Cython uses custom operators `address` (for `ref`/`pointer_of` here) and `dereference` (for `.value` in the above snippet). I agree we need two operators to avoid ambiguity.
https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#c-operators-not-compatible-with-python-syntax

Zhyrek: I originally thought it would be fine because you can't reassign a reference variable once created (at least not without the new keyword), but I do agree it diverges significantly from python syntax. I'll rework it

Issue #8049, opened 2023-12-24
Use of NumPy type aliases to be removed in NumPy 2.0


lucascolley:
https://github.com/numpy/numpy/pull/24376 removed some type aliases for NumPy 2.0

[In SciPy](https://github.com/scipy/scipy/issues/19754) (through an unusual test we have) we found errors due to the use of `numpy.float_` in CuPy. From a brief search I see that there are also uses of `numpy.complex_`.

_(This shouldn't be failing SciPy's test suite, just putting it on your radar for NumPy 2.0 compat)_

kmaehashi:
Thanks! `numpy.float_` and `complex_` usages in CuPy looks only here:
https://github.com/cupy/cupy/blob/cb4e5127e38c1cb2bdd01ad5117d324fac1762fd/cupy/_core/_scalar.pyx#L64-L65

PR #8050, opened 2023-12-25
Remove usages of `numpy.float_` and `numpy.complex_`

asi1024:
Close https://github.com/cupy/cupy/issues/8049.

### DIFF
--- a/cupy/_core/_scalar.pyx
+++ b/cupy/_core/_scalar.pyx
@@ -61,8 +61,8 @@ cdef object _numpy_float32 = numpy.float32
 cdef object _numpy_float64 = numpy.float64
 cdef object _numpy_complex64 = numpy.complex64
 cdef object _numpy_complex128 = numpy.complex128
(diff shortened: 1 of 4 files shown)


cupy/_core/_scalar.pyx
```diff
@@ -61,8 +61,8 @@ cdef object _numpy_float32 = numpy.float32
 cdef object _numpy_float64 = numpy.float64
 cdef object _numpy_complex64 = numpy.complex64
 cdef object _numpy_complex128 = numpy.complex128
-cdef object _numpy_float_ = numpy.float_
-cdef object _numpy_complex_ = numpy.complex_
+cdef object _numpy_float_ = numpy.float64
+cdef object _numpy_complex_ = numpy.complex128
```
lucascolley, line 65: Just to check, they were previously aliases for `double` and `cdouble`. I think that doesn't matter here but just in case you were wanting a straight alias swap

kmaehashi (approved):

Issue #3452, opened 2020-06-18
Mapped memory functionality (zero-copy)


mnicely:
Does CuPy have an implementation of _numba.cuda.mapped_array_ (https://numba.pydata.org/numba-doc/dev/cuda-reference/memory.html)

The idea is to use zero-copy on Jetson devices. 

An example, let's say I loaded a 2GB file into the global memory space of a TX2. I want to give the GPU access to the data without make a deep copy.
https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#zero-copy

mnicely:
Pulling on that thread some more, I'm wondering if I using the CuPy pinned memory correctly. Below is a screenshot compare CuPy and Numba to pin memory. The CuPy is taking twice as long.

![numba_vs_cupy_pinned](https://user-images.githubusercontent.com/50021634/85127873-3fe4f300-b1fe-11ea-8570-b264d35df428.png)

### CuPy code
```python
# https://github.com/cupy/cupy/blob/master/examples/stream/cupy_memcpy.py
def _pin_memory(array):
    mem = cp.cuda.alloc_pinned_memory(array.nbytes)
    ret = np.frombuffer(mem, array.dtype, array.size).reshape(array.shape)
    ret[...] = array
    return ret


def get_pinned_array(size, dtype):
    pinned_memory_pool = cp.cuda.PinnedMemoryPool()
    cp.cuda.set_pinned_memory_allocator(pinned_memory_pool.malloc)

    x_cpu_dst = np.empty(size, dtype)
    x_pinned_cpu_dst = _pin_memory(x_cpu_dst)

    return x_pinned_cpu_dst

buffer = cusignal.get_pinned_array(binary.shape, cp.ubyte)
```

### Numba code
```python
def get_shared_mem(shape, dtype=np.float32, strides=None, order='C', stream=0,
                   portable=False, wc=True):
    return cuda.mapped_array(shape, dtype=dtype, strides=strides, order=order,
                             stream=stream, portable=portable, wc=wc)

buffer = cusignal.get_shared_mem(binary.shape, cp.ubyte)
```

leofang:
@mnicely I think for the CuPy code you allocated host memory twice. You shouldn't need the line `x_cpu_dst = np.empty(size, dtype)` in `get_pinned_array()` and should just allocate the memory directly using `cp.cuda.alloc_pinned_memory()`. I think it's why it took twice as long.

mnicely:
[...]

PR #8056, opened 2023-12-27
Add `cupyx.signal{write_bin,read_bin}` and related utils

emcastillo:
`get_shared_mem` was implemented as per
https://github.com/cupy/cupy/issues/3452#issuecomment-903212530
since cusignal directly called numba
https://github.com/rapidsai/cusignal/blob/branch-23.10/python/cusignal/utils/arraytools.py#L87


cupyx/signal/_io/_writer.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # Get current stream, default or not.
+    stream = cupy.cuda.get_current_stream()
+
+    if buffer is None:
+        buffer = cupy.asnumpy(binary)
+    else:
+        binary.get(out=buffer)
+
+    if append is True:
```
asi1024, line 34:
```suggestion
    if append:
```

Issue #8046, opened 2023-12-22
Support SciPy 1.12


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [ ] Read Release Notes.
- [ ] Fix code and CI to support the new vesrion.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - **Update Requirements**: For NumPy/SciPy, bump the pinned minimum/maximum version in `setup.py` (see #4773.)
- [ ] Backport the above PR.
- [ ] For NumPy update, reflect the changes in `numpy.array_api` to `cupy.array_api`.
- [ ] Backport the above PR.
- [ ] Fix documentation.
  - Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) (note: the baseline version must be updated only in the development branch).
  - Update `docs/requirements.txt` to use the latest versions.
- [ ] Backport the above PR.
- [ ] Apply fixes needed to follow the specification changes.

kmaehashi:
Tests failing against SciPy v1.12.0rc1:

https://ci.preferred.jp/cupy.linux.cuda-head-stable/151407/
https://ci.preferred.jp/cupy.linux.cuda-head/151404/

asi1024:
https://github.com/cupy/cupy/pull/8127

kmaehashi:
These tasks seem overlooked? @asi1024

> Update Requirements: For NumPy/SciPy, bump the pinned minimum/maximum version in setup.py (see https://github.com/cupy/cupy/issues/4773.)
> Update docs/requirements.txt to use the latest versions.

kmaehashi:
> These tasks seem overlooked?

Will be done in #8500

PR #8059, opened 2023-12-27
Fix `scp.signal.{medfilt,medfilt2d}` to raise ValueError for complex64 inputs

asi1024:


takagi (approved):
LGTM!

PR #8061, opened 2023-12-27
Deprecate `cupyx.scipy` wavelet functions

asi1024:
According to [the SciPy 1.12 release notes](https://scipy.github.io/devdocs/release/1.12.0-notes.html#deprecated-features), all wavelet functions will be deprecated and it is recommended to use PyWavelets instead. Should we move the wavelet functions from `cupyx.signal` to the `cupyx.pywt` namespace?

### DIFF
--- a/cupyx/scipy/signal/_wavelets.py
+++ b/cupyx/scipy/signal/_wavelets.py
@@ -33,7 +33,8 @@ from cupyx.scipy.signal._signaltools import convolve
 
 
 _deprecate_msg = (
-    "All wavelet functions have been deprecated since CuPy v14. "
+    "Following the change in SciPy 1.12, all wavelet functions have been "
+    "deprecated in CuPy v14 and are planned to be removed in the future. "
     "To request continued support of the features, "
     "please leave a comment at https://github.com/cupy/cupy/pull/8061."
 )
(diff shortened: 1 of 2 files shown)


kmaehashi:
Thanks for catching up with the SciPy 1.12 changes!

> Should we move the wavelet functions from `cupyx.signal` to the `cupyx.pywt` namespace?

This sounds a rescope of our project. Since we don't have a plan to host PyWavelet functions in CuPy (as of now, at least), I'm opposed to introducing the new namespace. I think it's ok to mark these APIs deprecated like we had done in the past cases.

If someone really wants the GPU version of wavelet functions after SciPy removes them, we can move them to `cupyx.signal` namespace which hosts non-SciPy compatible signal processing routines. I think it's nice to include a link to this PR (or another "RFC" issue) in the DeprecationWarning message to know if there is a user who wants them.

cupyx/scipy/signal/_wavelets.py
```diff
@@ -25,12 +24,21 @@
...
+import warnings
+
 import cupy
 import numpy as np
 
 from cupyx.scipy.signal._signaltools import convolve
 
 
+_deprecate_msg = (
+    "All wavelet functions have been deprecated since CuPy v14. "
```
kmaehashi, line 36: How about adding more info like this?

```
Following the change in SciPy 1.12, all wavelet functions have been deprecated in CuPy v14 and are planned to be removed in the future.
```

kmaehashi (approved):
LGTM except for the minor message comment. Please proceed to merge after test pass.

kmaehashi (approved):

PR #8063, opened 2023-12-27
Fix `csrmatrix.__pow__` to raise ValueError for non-int other

asi1024:


takagi (approved):
LGTM!

PR #8064, opened 2023-12-27
Fix scipy.linalg not to raise DeprecationWarning for zero-size inputs

asi1024:


takagi (approved):
LGTM!

PR #8037, opened 2023-12-19
Add `cupyx.signal.convolve1d3o` from cuSignal's non SciPy-compat API

takagi:
Please merge #8022 and #8047 first.

The followings are the original cuSignal API top and its CUDA kernel implementation.
https://github.com/rapidsai/cusignal/blob/136453a3c8ef381f502a1edd51ca741f6c159f4a/python/cusignal/convolution/convolve.py#L620
https://github.com/rapidsai/cusignal/blob/136453a3c8ef381f502a1edd51ca741f6c159f4a/cpp/src/convolution/_convolution.cu#L695

PR #8078, opened 2024-01-04
Refactor convolve1d3o

asi1024:
https://github.com/cupy/cupy/pull/8037

Verified that there is no performance degradation using the following code.

```py
from cupy import testing
from cupyx.signal import convolve1d3o

data = testing.shaped_random((2 ** 25,), dtype='float32')
kernel = testing.shaped_random((4, 4, 4), dtype='float32')
print(benchmark(convolve1d3o, (data, kernel), max_duration=3))

data = testing.shaped_random((2 ** 10,), dtype='float32')
kernel = testing.shaped_random((64, 64, 64), dtype='float32')
print(benchmark(convolve1d3o, (data, kernel), max_duration=3))
```


takagi (approved):
LGTM!

Issue #7404, opened 2023-02-14
Tracker: cupyx.scipy.signal.windows


andfoy:
## Description
This is a tracking issue regarding the current and future state of compatibility between `scipy.signal.windows` and `cupyx.scipy.signal.windows`. The sections denoted below are aligned with respect to SciPy's reference API: https://docs.scipy.org/doc/scipy/reference/signal.windows.html

### Window functions
- [ ] `get_window`
- [ ] `barthann`
- [ ] `bartlett`
- [ ] `blackman`
- [ ] `blackmanharris`
- [ ] `bohman`
- [ ] `boxcar`
- [ ] `chebwin`
- [ ] `cosine`
- [ ] `dpss`
- [ ] `exponential`
- [ ] `flattop`
- [ ] `gaussian`
- [ ] `general_cosine`
- [ ] `general_gaussian`
- [ ] `general_hamming`
- [ ] `hamming`
- [ ] `hann`
- [ ] `kaiser`
- [ ] `kaiser_bessel_derived`
- [ ] `lanczos`
- [ ] `nuttall`
- [ ] `parzen`
- [ ] `taylor`
- [ ] `triang`
- [ ] `tukey`

leofang:
cc: @awthomp

jakirkham:
It looks like all, but one are implemented. Should we close and convert the remaining function to a new issue?

PR #8081, opened 2024-01-04
ENH: signal: add `lanczos` and `kaiser_bessel_derived` windows

ev-br:
cross-ref https://github.com/cupy/cupy/issues/7404

Both windows are trivial enough so that just copy pasting scipy implementations seems enough.

### DIFF
--- a/cupyx/scipy/signal/windows/_windows.py
+++ b/cupyx/scipy/signal/windows/_windows.py
@@ -1450,6 +1450,7 @@ def kaiser_bessel_derived(M, beta, sym=True):
     See Also
     --------
     kaiser
+    scipy.signal.windows.kaiser_bessel_derived
 
     Notes
     -----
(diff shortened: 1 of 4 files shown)


takagi:
Seems that `lanczos` and `kaiser_bessel_derived` windows are introduced since SciPy 1.10. Would you mark the tests to require newer SciPy?

ev-br:
Done in [2e52b82](https://github.com/cupy/cupy/pull/8081/commits/2e52b820781e6d1b2ce97f73e2751cca7afc614d).

takagi:
Thanks! I've kicked the CI again.

cupyx/scipy/signal/windows/_windows.py
```diff
@@ -1424,6 +1424,91 @@ def kaiser(M, beta, sym=True):
...
+
+
+    References
+    ----------
+    .. [1] Bosi, Marina, and Richard E. Goldberg. Introduction to Digital
+           Audio Coding and Standards. Dordrecht: Kluwer, 2003.
+    .. [2] Wikipedia, "Kaiser window",
+           https://en.wikipedia.org/wiki/Kaiser_window
+
+    Examples
```
takagi, line 1468: Would you fix this examples section to work with cupy? Or I suppose removing this section is also okay as users can refer to scipy's.

cupyx/scipy/signal/windows/_windows.py
```diff
@@ -1424,6 +1424,91 @@ def kaiser(M, beta, sym=True):
     return _truncate(w, needs_trunc)
 
 
+def kaiser_bessel_derived(M, beta, sym=True):
```
takagi, line 1427:
```suggestion
def kaiser_bessel_derived(M, beta, *, sym=True):
```

cupyx/scipy/signal/windows/_windows.py
```diff
@@ -2036,6 +2121,64 @@ def taylor(M, nbar=4, sll=30, norm=True, sym=True):
     return _truncate(w, needs_trunc)
 
 
+def lanczos(M, sym=True):
```
takagi, line 2124: To get aligned with SciPy.
```suggestion
def lanczos(M, *, sym=True):
```

cupyx/scipy/signal/windows/_windows.py
ev-br: This breaks `get_window` which is not equipped to handle keyword-only arguments. Note that the bulk of the windowing APIs come from cuSignal, and probably predates both introduction of these two windows and the keywod-only stuff

cupyx/scipy/signal/windows/_windows.py
ev-br: Removed and added a `See Also`  link to the scipy version.

cupyx/scipy/signal/windows/_windows.py
takagi: I got it. So lets leave the signatures not to have keyword-only parameters in this PR. Thanks!

takagi (approved):
LGTM!

PR #8088, opened 2024-01-05
Fix scipy>=1.12 filter condition

asi1024:
https://ci.preferred.jp/cupy.linux.cuda121/152363/


kmaehashi:
This fixes main branch CI failures (added in PRs for #8046)

Issue #2616, opened 2019-11-08
Adding `writeable` to `flags`


jakirkham:
NumPy arrays have the `writeable` option in `flags`. This can be useful for checking to see if one can write to an array or not. Also it can be useful to ensure an array is not written to (by setting this flag to `False`). Would be useful to have this flag on CuPy arrays as well.

leofang:
I suppose you know this already? 

Changing `WRITEABLE` (to `True`) in Python is now considered dangerous and deprecated in NumPy: 
https://docs.scipy.org/doc/numpy/release.html#writeable-flag-of-c-api-wrapped-arrays 

But are you suggesting the opposite operation should be allowed?

I could dig this myself, but perhaps it's faster to just ask 🙂: How does this flag get honored at the C level in NumPy? Does every NumPy operation need to consult it? The change sounds like some non-trivial task to me, but perhaps I overthink.

FYI, in CUDA there's constant memory guaranteed not writable once initiated, but it's a scarce hardware resource and almost always not enough for serious computations. It'd be nice for `RawKernel` users to be able to access it (see #1703 and #2510), but it's a different story.

jakirkham:
Right the interest here would be to mark arrays as read-only. 🙂

Yeah NumPy honors this at the C level as well. It does bit masking against the flags (an `int`) to check. Only operations that are manipulating the data itself.

I'm not sure. Would think `__setitem__` and the inplace math operations would be the main issues. Are there other cases outside of these that would manipulate data in CuPy? You know the code here better than I. 😉

That seems like a useful feature to include. Though agree it is a different concern.

kmaehashi:
There are some case that NumPy return non-writable ndarray (e.g. `xp.imag(xp.ones(5))`), so it's better to have this feature also in terms of the compatibility. As suggested by Leo, I also think the change is not trivial, though. Specifically I don't come up with an idea to protect an array memory passed to `RawKernel`/`RawModule`.

leofang:
In addition to RawKernel, I suppose any CUDA libraries that permit in-place operation (like cuFFT) would also require some nontrivial guard.

I fantasize asking NVIDIA folks to allow setting a writable flag in the pointer attribute would probably be  the easiest (though highly unlikely) approach 🤣

kmaehashi:
Discussed in the dev team today, the reasonable implementation is to to support writable flag only in ElementwiseKernel/ReductionKernel (i.e., writable arrays cannot be specified as an output array).

[...]

PR #8118, opened 2024-01-17
[WIP] Add `writeable` to `flags`

asi1024:
Close https://github.com/cupy/cupy/issues/2616.
Blocked by https://github.com/cupy/cupy/pull/8104.

Adds `writeable` attributes to `cupy.ndarray.flags`. Please note that using read-only ndarray does not improves performance, since it currently does not automatically generate code with `const` or `__restrict__`.


kmaehashi:
Thanks! Some high-level comments:
* We need to decide how the flag is inherited between NumPy/CuPy.
* Also needs interoperability consideration, e.g. DLPack https://github.com/numpy/numpy/issues/20742.
* The documentation must state that (1) race condition may happen due to async behavior of CUDA (data may change after changing the writeable flag to False) and (2) flags may be ignored in some scenarios, e.g., when passing the read-only ndarrays to CuPy functions using RawKernel/RawModule.

PR #8126, opened 2024-01-19
[WIP] Upstream cuCIM vendored filter changes to CuPy

jakirkham:
Takes changes to the CuPy filter currently vendored in cuCIM and upstreams them.

cc @grlee77 @leofang

### DIFF
--- a/cupyx/scipy/ndimage/_filters_core.py
+++ b/cupyx/scipy/ndimage/_filters_core.py
@@ -10,25 +10,30 @@ from cupyx.scipy.ndimage import _util
 
 
 def _origins_to_offsets(origins, w_shape):
-    return tuple(x//2+o for x, o in zip(w_shape, origins))
+    return tuple(x // 2 + o for x, o in zip(w_shape, origins))
 
 
-def _check_size_footprint_structure(ndim, size, footprint, structure,
-                                    stacklevel=3, force_footprint=False):
+def _check_size_footprint_structure(
+    ndim, size, footprint, structure, stacklevel=3, force_footprint=False
+):
     if structure is None and footprint is None:
         if size is None:
             raise RuntimeError("no footprint or filter size provided")
-        sizes = _util._fix_sequence_arg(size, ndim, 'size', int)
+        sizes = _util._fix_sequence_arg(size, ndim, "size", int)
         if force_footprint:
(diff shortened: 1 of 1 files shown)


leofang:
@jakirkham Any chance we can omit the format changes and only keep the real ones?

grlee77:
> @jakirkham Any chance we can omit the format changes and only keep the real ones?

Thanks, for porting the code upstream John. Unfortunately on closer review, I think only a subset of changes will make sense to backport and we will need to maintain some functionality that is specific to cuCIM within its vendored code.

## Style changes
The one style change that I would encourage to make is switching to f-string use within `_generate_nd_kernel` as I think it makes it easier to read. I can split that off into a small separate MR without other formatting changes.

## Functional Changes
I reviewed this to refresh my memory on why these changes were made on the cuCIM side. There are two separate functional changes here:
- addition of `sizes` argument to `_check_nd_args`
- addition of `**filter_kwargs` to `_run_1d_filters`

Unfortunately, neither of those are useful to CuPy itself with the change to this file alone, as they would not be used by the higher level code in the module. We can potentially backport the cuCIM features that motivated those API additions in separate MRs if there is interest. I will describe them below.

### "sizes" argument
The `sizes` argument was added to be able to specify a square or rectangular footprint merely by supplying a "sizes" tuple describing its shape. This is an optimization to avoid the overhead of allocating the tiny footprint array on the device. We did this primarily to enable morphology functions to pass a tuple directly to indicate a rectangular footprint. This tuple-based footprint is not supported in SciPy itself, though, so it is a deviation in behavior. It pays off more in cuCIM because we also have a lot of supporting code (as in scikit-image) that [decomposes large footprints into an equivalent set of smaller footprints](https://github.com/scikit-image/scikit-image/pull/5482) that would be applied in sequence. 

### **filter_kwargs
`**filter_kwargs` was added to be able to have a way of using an additional "algorithm" kwarg for some separable filters that would be used internally to allow them to choose an alternative set of kernels when advantageous. To use this form, CuPy would require porting additional alternative implementations here, but I am not sure it will be worth the additional maintenance burden.

grlee77:
@jakirkham, I think we can close this in favor of #8311 and #8312

jakirkham:
Thanks Greg! 🙏

PR #3374, opened 2020-05-27
Fix cuTENSOR routines to raise `ValueError` for invalid arguments

asi1024:
Depends on #2772.

PR #8136, opened 2024-01-24
Fix#8133, Update cutensor.pyx

jemiryguo:
Remove the contiguous check in cutensor/contraction to enable the contaction performing on in-contiguous arrays.

### DIFF
--- a/cupy/_core/include/cupy/special/.clang-format
+++ b/cupy/_core/include/cupy/special/.clang-format
@@ -1,0 +1,10 @@
+BasedOnStyle: LLVM
+Standard: Cpp11
+UseTab: Never
+IndentWidth: 4
+BreakBeforeBraces: Attach
+Cpp11BracedListStyle: true
+NamespaceIndentation: Inner
+AlwaysBreakTemplateDeclarations: true
+SpaceAfterCStyleCast: true
+ColumnLimit: 120
--- a/cupy/_core/include/cupy/special/config.h
+++ b/cupy/_core/include/cupy/special/config.h
@@ -1,0 +1,136 @@
+#pragma once
+
+#ifdef __CUDACC__
+#define SPECFUN_HOST_DEVICE __host__ __device__
+
+// Define math constants if they are not available
+#ifndef M_E
+#define M_E 2.71828182845904523536
+#endif
+
+#ifndef M_LOG2E
+#define M_LOG2E 1.44269504088896340736
+#endif
+
+#ifndef M_LOG10E
+#define M_LOG10E 0.434294481903251827651
+#endif
+
+#ifndef M_LN2
+#define M_LN2 0.693147180559945309417
+#endif
+
+#ifndef M_LN10
+#define M_LN10 2.30258509299404568402
+#endif
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+
+#ifndef M_PI_2
+#define M_PI_2 1.57079632679489661923
+#endif
+
+#ifndef M_PI_4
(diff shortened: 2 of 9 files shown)


leofang:
Thanks, @jemiryguo, could you check the pre-test failures? You can run
```
pre-commit run -a --show-diff-on-failure
```
locally and fix the formatting issue following the hints.

leofang:
Other than formatting, I think the fix is on the right track. Once the formatting is fixed, I'll check with CI.

kmaehashi:
@leofang Could you review this one if possible, Leo?

leofang:
@asi1024 AFAICT the to-remove contiguity check was added in https://github.com/cupy/cupy/pull/3374, any chance you still remember the reason? 🙂 cuTENSOR was designed to handle strided tensor operations, so checking contiguity shouldn't be necessary with hindsight.

leofang (approved):

asi1024:
@leofang I added a check for contiguity because the test for non-contiguous input raised illegal memory access. If the test for non-contiguous input passes now, we can remove the check.

leofang:
Got it, seems like it's safe to remove now. Anything else that needs to be addressed before merging, @asi1024? Like adding extra tests etc?

asi1024:
Yes, could you add a contiguity parameter to `TestCuTensorContraction` in `tests/cupyx_tests/test_cutensor.py`?

wxj6000:
@leofang Could alignment_req check be removed too? 
 https://github.com/cupy/cupy/blob/f5d8ed48487f1ce6f31e524d5361bf6f47a4a2e4/cupyx/cutensor.pyx#L247

leofang:
Please keep the alignment requirement for now. There are some subtitles regarding it. For views, if the view does not offset from the original pointer address, the alignment shouldn't be an issue.

wxj6000:
> Please keep the alignment requirement for now. There are some subtitles regarding it. For views, if the view does not offset from the original pointer address, the alignment shouldn't be an issue.

Yes, there are some issues when the view offset from the original address. Just out of curiosity, 

And the contiguous check also appears in reduction, elementwise_binary, and elementwise_trinary. Not sure if it is safe to remove them too.

jemiryguo:
I will give another PR after i finishing all related changs

jemiryguo:
> Please keep the alignment requirement for now. There are some subtitles regarding it. For views, if the view does not offset from the original pointer address, the alignment shouldn't be an issue.

I think we can add a func like this:
```py
cpdef int cal_align_req(_ndarray_base a, _ndarray_base b, _ndarray_base c):
    """Calculate the required alignment to support incontiguous input/output.

    Args:
        a (cupy.ndarray): tensor involved in contraction.
        b (cupy.ndarray): tensor involved in contraction.
        c (cupy.ndarray): tensor involved in contraction.

    Returns:
        (int): required alignment.
    """
    cdef int pa = a.data.ptr
    cdef int pb = b.data.ptr
    cdef int pc = c.data.ptr

    cdef int alignment_req = 255
    cdef int i

    for i in range(5):
        if (pa & alignment_req) + (pb & alignment_req) + (pc & alignment_req) == 0:
            return alignment_req+1
        alignment_req >>= 1
    return 16
    ```

and calculate the required aligned before creating the descriptor:
```
    alignment_req = cal_align_req(A, B, C)
    desc_A = create_tensor_descriptor(A, alignment_req)
    desc_B = create_tensor_descriptor(B, alignment_req)
    desc_C = create_tensor_descriptor(C, alignment_req)
```

This will only effect the alignment for contraction, thus no worry about  "some subtitles regarding it".

Issue #8133, opened 2024-01-24
Additional contiguous check preventing the cutensor performing on in-contiguous arrays.


jemiryguo:
### Description

In principle the `cutensor` lib are support contraction on in-contiguous arrays vias custome `strides` parameters without any extra copy. e.g. Take the $100\times100$ matrix multiplication, a trival contraction, as an example, $c_{ik}=\sum a_{ij}*b_{jk}$ with all `a`, `b`, and `c` in form `arr[:80]`. `cutensor` can handle this without copying `a[:80]` into contiguous form.

However, in https://github.com/cupy/cupy/blob/main/cupyx/cutensor.pyx#L786-L787:
```python
    if not (A._c_contiguous and B._c_contiguous and C._c_contiguous):
        raise ValueError('The inputs should be contiguous arrays.')
```
The tempt to use this feature will be blocked by a unnessasary contiguous check.

After remove this two line, and rebuild the cupy, i found all went will and correct answer is obtained. The test code for `cupy 12.3.0`(newer versions of `cupy` are similar) are as follow.

The expected output (without these teo lines) should be:
```
Used: 0.00GB Total 0.00GB
Used: 4.66GB Total 4.66GB
Used: 9.31GB Total 9.31GB
Used: 13.97GB Total 13.97GB
Used: 13.97GB Total 13.99GB
Correctness check: True
```
The `a`, `b`, and `c` arrays all need 4.66GB, counting for total 13.97GB. Performing cutensor on in-contiguous array only increase the total memory from 13.97GB to 13.99GB. This is due to the workspace and far less than the makeing a contiguous copy.

### To Reproduce

```py
# Write the code here
import numpy
import cupy
import cupy_backends.cuda.libs.cutensor as cutensorlib
from cupy import cutensor

def contraction(inda, a, indb, b, indc, c=None, alpha=1.0, beta=0.0, dtype=None, buf=None, alg=cutensorlib.ALGO_DEFAULT,wspref=cutensorlib.WORKSPACE_MIN):
    modea = tuple(inda)
    modeb = tuple(indb)
    modec = tuple(indc)
    if c is None:
        dtype = numpy.result_type(a, b) if dtype is None else dtype
        cshape = [a.shape[modea.index(c)] if c in modea else b.shape[modeb.index(c)] for c in modec]
        if buf is None:
            c = cupy.empty(cshape, dtype=dtype)
        else:
            if numpy.prod(buf.shape) < numpy.prod(cshape):
[...]

PR #8149, opened 2024-01-31
add incontiguous support for cutensor functions

jemiryguo:
Related issue: https://github.com/cupy/cupy/issues/8133

Related PR: https://github.com/cupy/cupy/pull/8136

### DIFF
--- a/cupyx/cutensor.pyx
+++ b/cupyx/cutensor.pyx
@@ -242,8 +242,8 @@ cpdef TensorDescriptor create_tensor_descriptor(_ndarray_base a):
         (TensorDescriptor): A instance of class TensorDescriptor.
     """
     handle = _get_handle()
-    key = (handle.ptr, a.dtype, tuple(a.shape), tuple(a.strides))
     alignment_req = a.itemsize
+    key = (handle.ptr, a.dtype, tuple(a.shape), tuple(a.strides), alignment_req)
     if a.data.ptr & (alignment_req - 1) != 0:
         raise ValueError("Missaligned array")
     if key not in _tensor_descriptors:
--- a/cupyx/cutensor.pyx
+++ b/cupyx/cutensor.pyx
@@ -243,7 +243,8 @@ cpdef TensorDescriptor create_tensor_descriptor(_ndarray_base a):
     """
     handle = _get_handle()
     alignment_req = a.itemsize
-    key = (handle.ptr, a.dtype, tuple(a.shape), tuple(a.strides), alignment_req)
+    key = (handle.ptr, a.dtype, tuple(a.shape),
+           tuple(a.strides), alignment_req)
     if a.data.ptr & (alignment_req - 1) != 0:
         raise ValueError("Missaligned array")
     if key not in _tensor_descriptors:
--- a/cupyx/cutensor.pyx
+++ b/cupyx/cutensor.pyx
@@ -243,7 +243,7 @@ cpdef TensorDescriptor create_tensor_descriptor(_ndarray_base a):
     """
     handle = _get_handle()
     key = (handle.ptr, a.dtype, tuple(a.shape), tuple(a.strides))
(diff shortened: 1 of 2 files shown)


wxj6000:
Nice work!

cupyx/cutensor.pyx
```diff
@@ -243,7 +243,7 @@ cpdef TensorDescriptor create_tensor_descriptor(_ndarray_base a):
     """
     handle = _get_handle()
     key = (handle.ptr, a.dtype, tuple(a.shape), tuple(a.strides))
```
leofang, line 245: If we want to relax the alignment requirement, we need to add it as part of the cache key. For different alignments, the tensor descriptor cannot be reused.

leofang (changes requested):
In order to relax the alignment requirement, we need to make it a part of cache key for all objects that we cache internally, as long as creation of such objects require the presence of valid tensor descriptors, either directly or indirectly. This includes
- tensor descriptors,
- all operators (elementwise/reduction/contraction/etc)
- plans built on top of them

They need to be re-created as needed, if the new alignment mismatches with that used at object creation time.

cupyx/cutensor.pyx
jemiryguo: I got it, I will change that

jemiryguo: I found that we only need to change the key to cache descripters. For other cached object, desc.ptr is used in keys thus naturally be recreated.

leofang: I see, it's a good observation, so we're using the tensor descriptor pointer as a proxy to propagate the alignment requirement. As long as we don't have dangling pointers, this should work fine.

leofang:
@jemiryguo could you check the linter error please?

tests/cupyx_tests/test_cutensor.py
```diff
@@ -337,3 +337,133 @@ def test_contraction(self):
...
+        r_dtype = dtype
+        if dtype == numpy.complex64:
+            r_dtype = numpy.float32
+        elif dtype == numpy.complex128:
+            r_dtype = numpy.float64
+        a = testing.shaped_random(shape, cupy, dtype=r_dtype, scale=1)
+        if dtype.char in 'FD':
+            a = a + 1j * \
+                testing.shaped_random(shape, cupy, dtype=r_dtype, scale=1)
+        return a
```
leofang, line 362: This should not be necessary, `testing.shaped_random` can generate complex random arrays too, so just use it everywhere.

tests/cupyx_tests/test_cutensor.py
```diff
@@ -337,3 +337,133 @@ def test_contraction(self):
                              self.c, mode_c)
         cupy.testing.assert_allclose(self.c, self.c_ref,
                                      rtol=self.tol, atol=self.tol)
+
+
+@testing.parameterize(*testing.product({
+    'dtype_char': ['e', 'f', 'd', 'F', 'D'],
+    'shape': [(30, 40, 30, 35)],
+    'alpha': [0.5, 1.0],
+    'beta': [0.0, 1.0],
```
leofang, line 346: Let's also test different orders 
```suggestion
    'beta': [0.0, 1.0],
    'order': ('C', 'F'),
```
(need to pass `self.order` to `shaped_random`).

leofang (changes requested):
Thanks, @jemiryguo. The tests look good overall. I suggested two minor changes. Once they're in I'll kick off the CI.

leofang (approved):

leofang:
@asi1024 I think this is ready 🙂 btw I added the backport label to unblock @jemiryguo and his collaborators sooner, as they have memory-demanding applications.

asi1024 (approved):

asi1024:
LGTM. Thanks!

PR #8163, opened 2024-02-05
Add complex support for the digamma function

steppi:
This PR adds support for complex values to the `digamma` function. This is made possible by the recent translation of SciPy's complex valued `digamma` implementation to CUDA compatible C++, https://github.com/scipy/scipy/pull/19834. This PR also makes an improvement to the real valued `digamma` implementation which was made in SciPy but had not been ported to CuPy (https://github.com/scipy/scipy/pull/8129).

I've copied over the entirety of the SciPy special C++ library as it exists now, despite it not all being used for `digamma`, but if you'd like, I can go through and try to add only the minimum number of headers that are actually needed. I just thought it would be easier to maintain if SciPy and CuPy vendor the same version of this set of files.

`polevl_definition` in `_digamma.py` is no longer used for the `digamma` function but I've left it in place because it is imported in other files.

### DIFF
--- a/tests/cupyx_tests/scipy_tests/special_tests/test_digamma.py
+++ b/tests/cupyx_tests/scipy_tests/special_tests/test_digamma.py
@@ -9,7 +9,7 @@ import numpy
 class TestDigamma(unittest.TestCase):
 
     @testing.with_requires('scipy>=1.1.0')
-    @testing.for_all_dtypes(no_complex=False)
+    @testing.for_all_dtypes()
     @testing.numpy_cupy_allclose(atol=1e-13, rtol=1e-15, scipy_name='scp')
     def test_arange(self, xp, scp, dtype):
         import scipy.special  # NOQA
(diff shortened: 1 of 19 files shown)


tests/cupyx_tests/scipy_tests/special_tests/test_digamma.py
```diff
@@ -10,7 +10,7 @@ class TestDigamma(unittest.TestCase):
 
     @testing.with_requires('scipy>=1.1.0')
     @testing.for_all_dtypes(no_complex=True)
```
asi1024, line 12: Could you remove `no_complex=True` option to test with complex inputs?

```suggestion
    @testing.for_all_dtypes()
```

steppi:
Thanks @asi1024, it was oversight on my part not to set `no_complex=False`. I've found there is a failing test due to differences in how division by complex infinity is handled in the C++ standard library on my system vs in `thrust::complex`. 

For large `z`, (including infinite `z`) the code uses an asymptotic expansion `log(z) - 0.5/z` (plus additional terms if needed). If the input is `inf + x*1i`, for finite `x`, then in the C++ standard library `-0.5/z` evaluates to `0i`, but for `thrust` complex numbers it evaluates to `nan + nan*1i`. I've found the C++ standard library does not specify the behavior of division by complex infinities; the behavior is implementation dependent. The fix will be to make sure that the results do not depend on the outcome of any divisions by complex infinity. I'll update the code to check for infinities and handle them separately.

asi1024 (approved):

asi1024:
@steppi LGTM! Thanks!

PR #8170, opened 2024-02-08
MAINT: cupyx/interpolate: remove a duplicate function

ev-br:
Just a cleanup: a helper function was vendored from scipy twice in two source files.


takagi (approved):
LGTM!

takagi:
Thanks, @ev-br!

PR #8201, opened 2024-02-22
Fix: always switch to the submodule dir before checking git tag/commit

### LINKED ISSUE #8185: Jitify build version is wrong in CuPy v13
### Description

When debugging #8171 I noticed the reported Jitify build version makes no sense. It turns out that when executing this line
https://github.com/cupy/cupy/blob/34f8edb19ada737b3afefbc61043f6569f20c091/install/cupy_builder/install_build.py#L584
in a symlink (since https://github.com/cupy/cupy/pull/7956 moved all 3rd party headers to a separate folder and symlink back), what we get is CuPy's commit hash, not Jitify's. I'll fix it.

### To Reproduce

_No response_

### Installation

None

### Environment

_No response_

### Additional Information

_No response_

leofang:
Close #8185.

### DIFF
--- a/install/cupy_builder/install_build.py
+++ b/install/cupy_builder/install_build.py
@@ -528,7 +528,9 @@ def check_cub_version(compiler, settings):
         # could be in a git submodule?
         try:
             # CuPy's bundle
-            cupy_cub_include = _cub_path
+            cupy_cub_include = os.path.join(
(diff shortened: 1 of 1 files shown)


kmaehashi (approved):
LGTM!

PR #8229, opened 2024-03-07
Fix build system for Thrust detection

leofang:
Recent Thrust has changed such that it's no longer safe to use the host compiler to compile the headers. Making a new choice based on the discussion in https://github.com/NVIDIA/cccl/issues/1373. This also aligns the treatment with ROCm builds.


takagi (approved):
LGTM!

Issue #6663, opened 2022-04-19
JIT: Allow users to register external device code?


leofang:
### Description

During the past course of exposing CUDA C++ APIs to CuPy JIT I've gradually realized a big missing piece, which is to allow users to "bring in" their device kernels to CuPy JIT with ease. Currently it requires to go through the entire JIT stack to
- provide the calling signature in Python
  - the signature could deviate from C++'s, as it's hard to deal with raw pointers in Python
- make the types recognized by JIT
- special-case the treatment in the Python AST -> CUDA C++ translation

I feel the already-endless chase of CUDA built-in C++ APIs alone is tough enough, not to mention if users have device kernels that they'd like to interact with from Python. Such examples (that would interest multiple parties) include
- Exposing [NVSHMEM device-side communication APIs](https://docs.nvidia.com/hpc-sdk/nvshmem/api/docs/introduction.html#gpu-initiated-communication-and-strong-scaling) (cc: @dalcinl)
- Exposing CUDA math lib device APIs (ex: cuFFTDx) (cc: @mnicely)

It'd be nice -- in the future -- to add a registration system for users to register external device code and include the needed headers in CuPy JIT kernels, though I do not have good advices for implementation 😅 

A point of reference is Numba which will add [similar support](https://github.com/numba/numba/pull/7621), but there they already allow users to register foreign C/C++ types (through ctypes if I am not mistaken), so I am not sure it's something we could learn from...

asi1024:
@leofang Would the following interface be enough to call external device code?

```py
code = '''
__device__ float device_func(float* x) { ... }
'''

@cupyx.jit.rawkernel(preamble=[code])
def kern(array):
    ...
    f = cupyx.jit.get_function_from_preamble('device_func', return_type=numpy.float32)
    y = f(array.data.ptr)  # No typecheck for arguments
    ...
```

leofang:
Thanks, @asi1024. I will do a case survey and get back to you. My current concerns would be
- Don't we need to at least declare the number of inputs?
- Without type-checking input parameters, do we know how to generate the right code?

asi1024:
I am considering the following implementation.
```py
class ExternalFunc(BuiltinFunc):
    def __init__(self, name, return_type='void'):
        self.name = name
        self.return_type = to_ctype(return_type)  # to_ctype is not implemented currently

    def call(self, env, *args):
        params = ', '.join([str(Data.init(x)) for x in args])
[...]

PR #8242, opened 2024-03-14
Support new cuFFT callbacks

leofang:
cuFFT now supports a new callback mechanism that is not only more performant but also more friendly to dynamic languages like Python and Julia, currently under [public preview](https://docs.nvidia.com/cuda/cufft/ltoea/index.html) and will be released in a future CUDA version. 

The new approach does not require any ugly compilation of host & device code; instead, we just need to compile a user-provided device function (currently supporting raw CUDA C++ strings; once we support user-provided types in `@cupyx.jit.rawkernel` (https://github.com/cupy/cupy/issues/6663) we can also support pure Python functions as device functions) into LTO IR, and pass it to the new cuFFT API `cufftXtSetJITCallback` which will be responsible for linking the LTO IR with the cuFFT kernel.

The existing callback manager (`_CallbackManager`) is renamed to `_LegacyCallbackManager`, to be distinguished with the new `_JITCallbackManager`. My suggestion is once this feature is officially released (generally accessible), we add a deprecation warning to `_LegacyCallbackManager`. Users can currently pick either callback approach via the new option `cp.fft.config.set_cufft_callbacks(..., cb_ver=...)`.

Currently, this feature requires users loading libcufft EA version: `LD_PRELOAD=/path/to/EA/libcufft.so python ...`, including running the test suite or sample code, so that the libcufft copy from CUDA Toolkit is not used.


jakirkham:
Stumbled across this recently when doing some testing with CuPy. This seems like a nice addition

Also it looks like this is now out of preview. Meaning we could start relying on this

> The cuFFT LTO EA preview is deprecated and will be removed in the near future.
> 
> The main feature of the preview, LTO callbacks, has been integrated into cuFFT as part of the [CUDA Toolkit 12.6 Update 2](https://docs.nvidia.com/cuda/cufft/#cufft-callback-routines). This update comes with significant improvements compared to the cuFFT LTO EA preview, including the ability of giving callback device functions arbitrary names, as well as improved performance.

ref: https://docs.nvidia.com/cuda/cufft/ltoea/index.html

leofang:
@kmaehashi If I clean this up within ~1 week, would it be possible to merge and backport to v13.5?

.pfnci/matrix.yaml
```diff
@@ -730,26 +730,26 @@
 # CUDA Python Test
 ###############################################################################
 
-# CUDA 11.x / CUDA Python | Linux
-- project: "cupy.linux.cuda11x-cuda-python"
-  target: "cuda11x-cuda-python"
+# CUDA 12.x / CUDA Python | Linux
+- project: "cupy.linux.cuda12x-cuda-python"
```
leofang, line 530: @kmaehashi I need to move the cuda-python CI to 12.x because we need a new NVRTC API from 12.x. But it seems a new `cupy.linux.cuda12x-cuda-python` pipeline cannot be registered by making changes in the codebase? I assume there's a backend that need to be updated as well? 🙂

leofang:
All tests pass! The new callback supports both Linux/Windows, and I've adjusted the codebase to make it tested on both platforms. Oh, it also supports T4 so we don't have to skip the tests in the CI 😉 

@kmaehashi this is ready for review (modulo https://github.com/cupy/cupy/pull/8242#discussion_r2270019578). I've added deprecation warnings to the old/legacy callback path because the new one is much more accessible (see above) and performant.

cupy/cuda/cufft.pyx
```diff
@@ -3,16 +3,80 @@ from cpython.mem cimport PyMem_Malloc, PyMem_Free
...
+    _L = _get_softlink()
+
+    global _cufftXtSetJITCallback
+    if CUPY_CUDA_VERSION < 13000:
+        # __cufftXtSetJITCallback_12_7
+        _cufftXtSetJITCallback = <F_cufftXtSetJITCallback>_L.get(
+            'XtSetJITCallback_12_7')
+    else:
+        _cufftXtSetJITCallback = <F_cufftXtSetJITCallback>_L.get(
+            'XtSetJITCallback')
```
leofang, line 46: Leaving a note here. The reason we fetch the internal symbol `__cufftXtSetJITCallback_12_7` instead of `cufftXtSetJITCallback` is because the latter has the wrong signature (lacking `callback_name`) with cuFFT from CTK 12.x, which is corrected in 13.0+. The internal symbol was provided by the cuFFT team to assist the transition, and is removed in 13.0.

.pfnci/matrix.yaml
kmaehashi: I've configured the pipeline!

cupy/cuda/cufft.pyx
```diff
@@ -3,16 +3,80 @@ from cpython.mem cimport PyMem_Malloc, PyMem_Free
...
+
+# ****************** SoftLink utilities ******************
+
+cdef SoftLink _L = None
+
+cdef inline void initialize() except *:
+    global _L
+    if _L is not None:
+        return
+    _initialize()
```
kmaehashi: The assignment must be done here to prevent race condition (#8774)

```suggestion
    _L = _initialize()
```

cupy/cuda/cufft.pyx
```diff
@@ -3,16 +3,80 @@ from cpython.mem cimport PyMem_Malloc, PyMem_Free
...
+
+cdef inline void _initialize() except *:
+    global _L
```
kmaehashi:
```suggestion
cdef SoftLink _initialize() except *:
```

cupy/cuda/cufft.pyx
kmaehashi, line 46:
```suggestion
            'XtSetJITCallback')

    return _L
```

.pfnci/matrix.yaml
leofang: Thanks, Kenichi-san! I'll rerun the CI after 13.6.0 is out, to avoid resource contention.

leofang:
@kmaehashi I wonder why `cupy.linux.cuda11x-cuda-python` still shows up in the CI list, but not the cuda12x one?

leofang:
Friendly nudge, @kmaehashi 🙂

.pfnci/matrix.yaml
leofang: @kmaehashi for some reason the new pipeline still cannot be triggered by the `/test` command (despite I can see it's available in https://ci.preferred.jp/cupy.linux.cuda12x-cuda-python/). I grabbed the codebase and saw no more `cupy.linux.cuda11x-cuda-python` in my local branch, so I should already purge it thoroughly. Maybe the new pipeline needs to be registered on GitHub...?

kmaehashi:
Fixed in https://github.com/cupy/cupy/pull/9368, new tests added in PRs can now be trigged within that PR.

cupy/cuda/cufft.pyx
leofang: Ah thanks Kenich-san, I missed that PR

leofang:
> Fixed in #9368, new tests added in PRs can now be trigged within that PR.

Thanks @kmaehashi! I see it can be triggered now, but cuda11x is still shown as "Required" whereas cuda12x is not... 🤔 
截圖 2025-09-11 下午12 00 04

leofang:
We're in a rabbit hole now... `cudaGraphInstatiate` changed its signature since CUDA 12, but for some reason we've been using the old signature even on CUDA 13. What's worse, ROCm does not follow and still use the old signature. 

Since this is no longer a quick fix, let me extract the infra bits and move them to a separate PR 😞

leofang:
> Since this is no longer a quick fix, let me extract the infra bits and move them to a separate PR 😞

#9377

leofang:
This looks like OOM to me, not sure if it's flaky on Windows @kmaehashi?
```
=========================== short test summary info ===========================
FAILED cupy_tests\core_tests\test_raw.py::TestRawPicklable_param_3_{compile=True, raw='ker'}::test_raw_picklable - subprocess.CalledProcessError: Command '['C:\\Development\\Python\\Python310\\python.exe', 'TestRawPicklable.py']' returned non-zero exit status 3221226505.
FAILED cupy_tests\core_tests\test_raw.py::TestRawPicklable_param_4_{compile=True, raw='mod'}::test_raw_picklable - subprocess.CalledProcessError: Command '['C:\\Development\\Python\\Python310\\python.exe', 'TestRawPicklable.py', 'test_sum']' returned non-zero exit status 3221226505.
FAILED example_tests\test_custom_struct.py::TestCustomStruct::test_builtin_vectors - subprocess.CalledProcessError: Command '('C:\\Development\\Python\\Python310\\python.exe', 'C:\\Windows\\TEMP\\flexci\\run-00410212\\work\\src\\tests\\example_tests\\..\\..\\examples\\custom_struct/builtin_vectors.py')' returned non-zero exit status 3221226505.
= 3 failed, 104202 passed, 8043 skipped, 768 deselected, 776 xfailed, 2582 warnings in 3502.94s (0:58:22) =
```
https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/199739/cupy_test_log.txt

leofang:
ping @kmaehashi 😅

leofang:
CI is green! 🎉

leofang:
ping @asi1024

cupy/fft/_callback.pyx
```diff
@@ -467,53 +498,215 @@ cdef class _CallbackManager:
...
+                <intptr_t>plan, self.cb_store_name, self.cb_store_lto,
+                cb_store_type, cb_store_ptr)
+
+        return <intptr_t>plan
+
+
+# LHS is our legacy convention
+# RHS is a valid C identifier pattern
+cdef object cupy_callback_pattern = re.compile(
+    r'd_(?:load|store)CallbackPtr\s*=\s*([A-Za-z_]\w*)\s*;')
```
asi1024: How about separating the patterns to handle cases where users might pass arguments in the wrong order?

```suggestion
cdef object cupy_load_callback_pattern = re.compile(
    r'd_loadCallbackPtr\s*=\s*([A-Za-z_]\w*)\s*;')

cdef object cupy_store_callback_pattern = re.compile(
    r'd_storeCallbackPtr\s*=\s*([A-Za-z_]\w*)\s*;')
```

leofang:
@asi1024 ready for re-review!

asi1024 (approved):

asi1024:
LGTM!

leofang:
Looks like @kmaehashi also needs to approve 🙂

PR #8244, opened 2024-03-15
Fix Generator API parallelism

kmaehashi:
Fixes the problem in random Generator API that the number of threads is unexpectedly fixed at 8 * 256 * 16 regardless of the SM count on the hardware.
`cudaDevAttrMultiProcessorCount` is a constant (16) for querying the attribute and should not have been used directly.

cc/ @emcastillo


leofang (approved):

emcastillo:
great catch!!!
LGTM!

takagi (approved):

Issue #8029, opened 2023-12-13
Helper class for IPC memory handling?


kmaehashi:
Regarding #8027 and the recent question on Gitter/Matrix, I'm wondering whether it will make sense to provide the feature.

The initial idea: allow users to do `a_ipc = cupy.arange(10).ipc()`. Users will get `a_ipc`, an instance of a serializable class that holds array information (shape, strides, dtype) & IPC handle, and can be used to pass it via multiprocessing or other processes.

c.f. https://numba.pydata.org/numba-doc/dev/cuda/ipc.html

Devdoot57:
Hi @kmaehashi, I actually needed this feature for a project I was working on, and I'm sure it could be valuable for others too. I will try to implement a method to get an instance of handle class using `a_ipc = cupy.arange(10).ipc()`. One can get back the original array using `a = a_ipc.get_array()`. Before proceeding further, I wanted to confirm the expected behavior with you.

kmaehashi:
@Devdoot57 Thanks! I think it's better to name it `get_ipc_handle()` (name borrowed from Numba) as it is not NumPy-compatible function. `numpy.ndarray.ipc()` is unlikely to happen, but I'd avoid the short name.

```python
# Source process
a_ipc = cupy.arange(10).get_ipc_handle()

# Destination process
a = a_ipc.get()  # open the IPC and invalidate the instance to avoid re-using the instance
del a  # close the IPC
```

Devdoot57:
Do you want it to throw an error if someone tries to call `a_ipc.get()` on the source process or calls it more than once on the destination process?

kmaehashi:
They sound nice to have 😃

Devdoot57:
Hi @kmaehashi, I noticed that one cannot open a handle using `cupy.cuda.runtime.ipcOpenMemHandle(handle)` on the source process or on a process created using the **fork** method. It throws the `cudaErrorInitializationError` error. A handle can only be opened in a process that has been created using the **spawn** method or the **forkserver** method. I think it makes sense to check the method before trying to open the handle when `get()` is called.

PR #8245, opened 2024-03-15
Add IPCHandle class for inter-process communication

Devdoot57:
This pull request adds a new `IPCHandle` class to facilitate inter-process communication (IPC) for sharing Cupy arrays between processes (#8029). The `IPCHandle` class provides methods for creating IPC handles and managing shared memory, allowing for efficient data transfer between processes.
The `.get_ipc_handle()` method returns an instance of `IPCHandle`, which can be passed to other processes (e.g. via multiprocessing). On the destination process, use the `.get()` method to create a new `ndarray` object that shares the allocation from the original process.

```python
# On source process
import cupy
arr = cupy.ones((3,3))
arr_ipc = arr.get_ipc_handle()

# On destination process
arr = arr_ipc.get()
```


cupy/_core/core.pyx
```diff
@@ -2056,6 +2058,38 @@ cdef class _ndarray_base:
...
+        Example:
+            >>> # Source process
+            >>> import cupy
+            >>> arr = cupy.ones((3,3))
+            >>> arr_ipc = arr.get_ipc_handle()
+            >>> # Destination process
+            >>> arr = arr_ipc.get()
+        """
+        handle = ipcGetMemHandle(self.data.ptr)
+        return IPCHandle(self.shape, self.size, self.dtype, self.strides, handle)
```
kmaehashi, line 2091:
```suggestion
        return IPCMemoryHandle(self)
```

I think it's better to retain the reference to `self` in the handle instance. Otherwise `a = cupy.arange(10).get_ipc_handle()` won't work as the backing ndarray will be freed immediately.

https://docs.nvidia.com/cuda/archive/9.0/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g01050a29fefde385b1042081ada4cde9

> Calling [cudaFree](https://docs.nvidia.com/cuda/archive/9.0/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ga042655cbbf3408f01061652a075e094) on an exported memory region before calling [cudaIpcCloseMemHandle](https://docs.nvidia.com/cuda/archive/9.0/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g02bb3632b5d223db6acae5f8744e2c91) in the importing context will result in undefined behavior.

It is also better to document that users are responsible for keeping the IPC instance alive in the source process for the lifetime of the corresponding array in the destination process.

cupy/_core/ipc_handle.py
```diff
@@ -0,0 +1,82 @@
+import os
+from cupy.cuda.memory import UnownedMemory, MemoryPointer
+from cupy.cuda.runtime import ipcOpenMemHandle, ipcCloseMemHandle
+from multiprocessing import get_start_method
+
+
+class IPCHandle:
+    """
+    A serializable class representing an Inter-process Communication (IPC)
+    handle for sharing Cupy arrays between processes.
```
kmaehashi, line 10: 😄 
```suggestion
    handle for sharing CuPy arrays between processes.
```

cupy/_core/ipc_handle.py
```diff
@@ -0,0 +1,82 @@
...
+        self.opened = False
+        self.source_pid = os.getpid()
+        self.arr_ptr = -1
+
+    def __del__(self):
+        """
+        Closes the handle when object is deleted or goes out of scope.
+        """
+        if self.arr_ptr != -1:
+            ipcCloseMemHandle(self.arr_ptr)
```
kmaehashi, line 38:
```suggestion
        self.close()
```

cupy/_core/ipc_handle.py
```diff
@@ -0,0 +1,82 @@
...
+            Cupy array: The shared Cupy array.
+
+        Raises:
+            RuntimeError: If get() is called on the source process or if
+            get() has already been called, or if the process was created
+            using the 'fork' method.
+        """
+        if get_start_method() == "fork":
+            raise RuntimeError(
+                "The handle cannot be spawned on a forked process.")
```
kmaehashi, line 62: This check might be inaccurate, because it is not guaranteed that the process that created an IPC handle is the same one as the fork parent of the current process.
```suggestion
```

cupy/_core/ipc_handle.py
kmaehashi, line 1: How about putting this file as `cupy/cuda/_ipc_handle.py`? I think this one is not a part of core.

kmaehashi (changes requested):
Thanks @Devdoot57!

cupy/_core/ipc_handle.py
```diff
@@ -0,0 +1,82 @@
...
+        if self.opened:
+            raise RuntimeError(
+                "The handle was already opened for this process.")
+
+        from cupy._core.core import ndarray
+
+        self.arr_ptr = ipcOpenMemHandle(self.handle)
+        mem = UnownedMemory(self.arr_ptr, self.size, owner=self)
+        memptr = MemoryPointer(mem, offset=0)
+        arr = ndarray(shape=self.shape, dtype=self.dtype,
```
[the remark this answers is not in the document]
Devdoot57: Ok, so you want `get()` to return an instance of whatever class (ndarray or its subclass) was used while creating the handle?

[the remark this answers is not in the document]
kmaehashi: I was thinking of that, but on second thought, I think it's difficult to support it without modifying pickling behavior in the base class (`ndarray`). Maybe adding a warning in `get_ipc_handle()` if it is a subclass instance would be sufficient for now 🙂

cupy/_core/ipc_handle.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def close(self):
+        """
+        Closes the IPC handle.
+        """
+        if self.arr_ptr != -1:
+            ipcCloseMemHandle(self.arr_ptr)
+            self.arr_ptr = -1
+
+    def get(self):
```
kmaehashi, line 48: I think `def open(self)` was better, considering `close()` is provided as an opposing concept 😅

cupy/_core/ipc_handle.py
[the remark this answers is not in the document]
Devdoot57: Ok, so for now one can still get a handle for a subclass, but `open()` will always return an instance of `ndarray`. I'll push the requested changes

cupy/_core/ipc_handle.py
Devdoot57: Got it. But, should I add an additional check because it throws `cudaErrorInitializationError` anyway?

kmaehashi: Sounds nice to show hint message for users in that case!

charlesbmi:
@kmaehashi @Devdoot57 this would be very helpful for my project. are there any remaining points I could help with?

cupy/cuda/_ipc_handle.py
```diff
@@ -0,0 +1,109 @@
+import os
+import warnings
+from multiprocessing import get_start_method
+
+from cupy.cuda.memory import UnownedMemory, MemoryPointer
+
+from cupy.cuda.runtime import ipcGetMemHandle
+from cupy.cuda.runtime import ipcOpenMemHandle
+from cupy.cuda.runtime import ipcCloseMemHandle
```
leofang, line 9:
```suggestion
from cupy_backends.cuda.apis.runtime import (
    ipcGetMemHandle, ipcOpenMemHandle, ipcCloseMemHandle)
```

cupy/cuda/_ipc_handle.py
```diff
@@ -0,0 +1,109 @@
...
+                          "subclass of cupy.ndarray, not cupy.ndarray "
+                          "itself. The .open() method will always return "
+                          "a cupy.ndarray object.")
+
+        if not isinstance(ndarray_instance, ndarray):
+            raise TypeError(
+                "The provided object is neither an instance of "
+                "cupy.ndarray, nor an instance of a subclass of "
+                "cupy.ndarray."
+            )
```
leofang, line 40:
```suggestion
        if type(ndarray_instance) is not ndarray:
            if isinstance(ndarray_instance, ndarray):
                warnings.warn("The provided object is an instance of a "
                    "subclass of cupy.ndarray, not cupy.ndarray "
                    "itself. The .open() method will always return "
                    "a cupy.ndarray object.")
            else:
                raise TypeError(
                    "The provided object is neither an instance of "
                    "cupy.ndarray, nor an instance of a subclass of "
                    "cupy.ndarray.")
```

cupy/cuda/_ipc_handle.py
```diff
@@ -0,0 +1,109 @@
...
+        self.handle = ipcGetMemHandle(ndarray_instance.data.ptr)
+        self.opened = False
+        self.source_pid = os.getpid()
+        self.arr_ptr = -1
+
+    def __del__(self):
+        """
+        Closes the handle when object is deleted or goes out of scope.
+        """
+        self.close()
```
leofang, line 55: Our experience shows that a bare `__del__` has terrible issues during Python interpreter shutdown. Please modify this with either [`is_shutting_down`](https://github.com/cupy/cupy/blob/03b45e5ce22599ce1f167a37124592660adf2620/cupy/cuda/stream.pyx#L130) ([preferred](https://github.com/cupy/cupy/pull/8315#issuecomment-3303982492)) or a [weakref finalizer](https://github.com/cupy/cupy/pull/8315).

leofang (changes requested):
@Devdoot57 sorry for late reply. I have some suggestions.

FWIW, exporting device memory via `ipcOpenMemHandle` & co is considered the legacy IPC approach. `cuda-core` recently adds the modern IPC support (though currently it's Linux-only). I'll think about how to enable CuPy after this PR is merged.

cupy/_core/ipc_handle.py
leofang: There is a way to tap into CPython's undocumented internal mechanism and ensure that at the time of serialization (pickling) the correct spawn method is used (raise if it is not). However, it is a bit nerve wrecking and I too think this is OK 😛

cupy/_core/ipc_handle.py
```diff
@@ -0,0 +1,82 @@
+import os
+from cupy.cuda.memory import UnownedMemory, MemoryPointer
+from cupy.cuda.runtime import ipcOpenMemHandle, ipcCloseMemHandle
+from multiprocessing import get_start_method
+
+
+class IPCHandle:
```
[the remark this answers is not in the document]
leofang: Yes. IPC event handles are necessary for stream-ordered IPC. It's good to make it clear.

leofang:
@Devdoot57 let us know if you'd like us to take over, or if you'd have some time to make updates!

Devdoot57:
Hey @leofang @kmaehashi , it’s been a while! I honestly lost track of this issue, but I’ve finally got some time to circle back and wrap this up.
I have made the requested changes locally. The new `.pre-commit-config.yaml` blocks direct commits to main. Since this old PR is tied to my fork's main branch, it prevents me from pushing the updates.
Should I use `--no-verify` to bypass it and update this existing PR, or would you prefer I open a new PR?

PR #8252, opened 2024-03-22
Support fallback allgorithm for spgemm

### LINKED ISSUE #7699: Use different cuSparse algorithm for SpGEMM
### Description

As noticed in issue #7044 the memory usage for SpGEMM can easily exceed the usage of GEMM on dense matrices.
The documentation of [CuSparse](https://docs.nvidia.com/cuda/cusparse/index.html#cusparsespgemm) specify that the memory usage is very high. As of CUDA 12.1.0 there are other options for the spgemm algorithm to lower the memory usage. Is it possible to fall back to `CUSPARSE_SPGEMM_ALG3` when `cusparseSpGEMM_workEstimation` fails with code 11? (`CUSPARSE_STATUS_INSUFFICIENT_RESOURCES`)
https://github.com/cupy/cupy/blob/3d80d9194530bdb02551f19f8119fa8b946c0e15/cupyx/cusparse.py#L2044

### Additional Information

_No response_

lowener:
This PR closes #7699. It is adding support for SPGEMM ALG2 of cusparse that requires a much lower memory usage than the default, used only when the default algorithm would fail.

### DIFF
--- a/cupy_backends/cuda/libs/cusparse.pxd
+++ b/cupy_backends/cuda/libs/cusparse.pxd
@@ -141,8 +141,13 @@ cpdef enum:
     # cusparseSpSMAlg_t
     CUSPARSE_SPSM_ALG_DEFAULT = 0
 
-    # ...
+    # cusparseSpGEMMAlg_t
     CUSPARSE_SPGEMM_DEFAULT = 0
+    CUSPARSE_SPGEMM_CSR_ALG_DETERMINITIC = 1
+    CUSPARSE_SPGEMM_CSR_ALG_NONDETERMINITIC = 2
+    CUSPARSE_SPGEMM_ALG1 = 3 # Default, fastest algorithm
+    CUSPARSE_SPGEMM_ALG2 = 4 # Balance between memory and time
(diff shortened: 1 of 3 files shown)


leofang:
Q: Is there a way we can test the new code path? Maybe you can add a known problem size to the test suite?

lowener:
I added a test to do `spgemm` on matrices (`100k x 50` - `50 x 100k`).
This problem size triggers the usage of the lower memory algorithm on my RTX 3070Ti (8GB) and on my Quadro RTX 8000 (48GB).
I disabled the element-wise comparison with scipy to not slow down the CI.

takagi (approved):
LGTM!

takagi:
Let me kick the full test to check with older CTKs.

takagi:
Would you check the result of static-checks and fix it?
https://github.com/cupy/cupy/actions/runs/8510603439/job/23642667154?pr=8252

takagi:
record: https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=b0674103c5e9e63eaddfecbaaa096e94d53c1eb4

takagi (approved):
LGTM!

Issue #8255, opened 2024-03-24
`test_sos_freqz_against_mp`  test trying to import nonexisting local `mpsig` module


Micket:
The following test fail for me in my build, and I'm confused as to how it was ever intended to work;
https://github.com/cupy/cupy/blob/028889eef3ba110829d677726348aa0f75aadb4e/tests/cupyx_tests/scipy_tests/signal_tests/test_filter_design.py#L642-L648

There is no mpsig submodule, there is no definition of zpkfreqz or butter_lp anywhere in the code.
Maybe it was intended to be part of https://github.com/cupy/cupy/pull/7537 by  @ev-br but wasn't staged into a commit?

takagi:
Seems that this test does not work. Can I ask you to check this, @ev-br?

ev-br:
gh-8259 fixes this locally for me. 
An alternative is to remove the whole test --- the original is checking scipy vs mpmath, and the CuPy test suite checks cupyx.scipy.signal vs scipy.

PR #8258, opened 2024-03-25
Use mpmath in test

takagi:
Following up #7537. See also #8255.

mpmath looks like a pure python library.


takagi:
The test should fail because of the issue posted on #8255.

kmaehashi:
Let do this in #8259!

PR #8263, opened 2024-03-30
ENH: cupyx/scipy/interpolate: add make_lsq_spline

ev-br:
cross-ref https://github.com/cupy/cupy/issues/7186

Add `make_lsq_spline`, lay the groundwork for FITPACK translations (splrep/splprep and *UnivariateSpline classes). Those will come is a follow-up PR, to keep the diff sizes manageable.

Algorithmically, this is solving a linear least-squares problem with banded matrices. Since banded linear algebra is not available from CUDA toolkit, we add a specialized kernel. As discussed offline, the kernel is essentially sequential.
(also note that even in cuSolver, a possibly helpful function, `csrlsqvqr` is also only available for CPU only).

### DIFF
--- a/cupyx/scipy/interpolate/_bspline2.py
+++ b/cupyx/scipy/interpolate/_bspline2.py
@@ -701,7 +701,7 @@ qr_reduce(double *a, int m, int nz, // a(m, nz), packed
     } // for(i = ...
 }
 
-'''
+'''  # NOQA
 
 
 TYPES = ['double']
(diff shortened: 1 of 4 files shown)


cupyx/scipy/interpolate/_bspline2.py
```diff
@@ -529,3 +528,345 @@ def _make_periodic_spline(x, y, t, k, axis):
...
+            }
+        }
+        if (i < nc) {
+            offset[i] = i;
+        }
+
+    } // for(i = ...
+}
+
+'''
```
takagi, line 704: Just suppressing the flake8 check.
```suggestion
'''  # NOQA
```

takagi (dismissed):

ev-br:
Merged as a part of https://github.com/cupy/cupy/pull/8267, closing.

PR #8267, opened 2024-04-02
ENH: cupyx/scipy/interpolate: add *UnivariateSpline for 1D smoothing splines

ev-br:
Add UnivariateSpline family of classes.  

SciPy plans to refactor / rework the matching part of scipy.interpolate, the current SciPy proposal is to add two two new functions, `make_splrep` and `make_splprep` and slowly phase out `splrep`, `splprep` and `*UnivariateSpline`s, see https://github.com/scipy/scipy/pull/19970 . Since this is still under discussion in SciPy, here we add two layers: 

- new-style `make_splrep` : add internally but do not expose to the public API. If/when this lands in SciPy, CuPy will only need to add them to `__init__` to expose them.
- backwards compatible `*UnivariateSpline` classes. These reuse the `make_splrep` internally and adjust the API to match the venerable SciPy API. 

The differences to the SciPy API are minimal:
- FITPACK in SciPy has a limitation that the spline order is `k <= 5`. We do not have this limitation here.
- `UnivariateSpline.roots` is not implemented. This is similar to other `.roots` methods which rely on solving the non-symmetric eigenvalue problem.

Also note that this PR only deals with construction of an interpolant.  Once constructed, evaluations go through `BSpline.__call__` method, so the runtime performance relates to `BSpline`.

This PR is on top of https://github.com/cupy/cupy/pull/8263, which adds a needed low-level kernel. 

cross-ref #7186 for the tracker.

### DIFF
--- a/cupyx/scipy/interpolate/_bspline2.py
+++ b/cupyx/scipy/interpolate/_bspline2.py
@@ -701,7 +701,7 @@ qr_reduce(double *a, int m, int nz, // a(m, nz), packed
     } // for(i = ...
 }
 
-'''
+'''  # NOQA
 
 
 TYPES = ['double']
--- a/cupyx/scipy/interpolate/_bspline2.py
+++ b/cupyx/scipy/interpolate/_bspline2.py
@@ -6,10 +6,9 @@ from numpy.core.multiarray import normalize_axis_index
 import cupy
 from cupyx.scipy import sparse
 from cupyx.scipy.sparse.linalg import spsolve
-from cupyx.scipy.interpolate._bspline import _get_dtype, _as_float_array
-
 from cupyx.scipy.interpolate._bspline import (
-    _get_module_func, INTERVAL_MODULE, D_BOOR_MODULE, BSpline)
+    _get_dtype, _as_float_array, _get_module_func, INTERVAL_MODULE,
+    D_BOOR_MODULE, BSpline)
 
 
 #################################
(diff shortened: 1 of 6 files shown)


cupyx/scipy/interpolate/_fitpack_repro.py
```diff
@@ -0,0 +1,1590 @@
...
+
+    NB: a new knot is in fact an `x` value at the middle of the interval.
+    So *the knots are a subset of `x`*.
+
+    This routine is an analog of
+    https://github.com/scipy/scipy/blob/v1.11.4/scipy/interpolate/fitpack/fpcurf.f#L190-L215
+    (cf _split function)
+
+    and https://github.com/scipy/scipy/blob/v1.11.4/scipy/interpolate/fitpack/fpknot.f
+    """
```
takagi, line 194:
```suggestion
    """  # NOQA
```

cupyx/scipy/interpolate/_bspline2.py
```diff
@@ -529,3 +528,345 @@ def _make_periodic_spline(x, y, t, k, axis):
...
+            }
+        }
+        if (i < nc) {
+            offset[i] = i;
+        }
+
+    } // for(i = ...
+}
+
+'''
```
takagi, line 704:
```suggestion
'''  # NOQA
```

takagi (approved):
LGTM!

takagi:
Thanks, @ev-br!

PR #7564, opened 2023-05-15
Port periodogram, welch and csd from CuSignal to cupyx.signal

andfoy:
See https://github.com/cupy/cupy/issues/7403
Depends on https://github.com/cupy/cupy/pull/7563 to ensure that the files do not get merge conflicts
Depends on https://github.com/cupy/cupy/pull/7536 to implement `csd`
Depends on https://github.com/cupy/cupy/pull/7568 to implement `spectral_analysis`, it depends on `get_window`

This port implies Python-only functions and therefore can be copied 1:1 from the original CuSignal implementation.

PR #8268, opened 2024-04-02
Replace license notice in cupyx.scipy.signal._spectral

andfoy:
See https://github.com/cupy/cupy/pull/7564#discussion_r1545755447


takagi (approved):
LGTM!

PR #8276, opened 2024-04-04
add NdBSpline based interpolation methods to RGI

ev-br:
Add a constructor for N-dimensional b-splines from data, which is new in SciPy 1.13.0. 

The user API matches that of SciPy 1.13.0:
- the constructor is exposed via the RegularGridInterpolator only
- there is a slight backwards compat change: RGI(..., method="cubic") now goes through the NdBSpline. Previous behavior is available via `RGI(..., method="cubic_legacy")`
- the new way is parametrically slower for construction of an interpolator (previously: essentially zero work; currently: construct and solve a large sparse linear system)
- the new way is parametrically faster for evaluations: previously the evaluation went through a python loop over the data points, and for each data point a loop over the dimensions; the new way delegates to NdBSpline, which has a GPU kernel for evaluations.
- new methods can compute the derivatives.

cross-ref https://github.com/cupy/cupy/issues/7186

### DIFF
--- a/cupyx/scipy/interpolate/_ndbspline.py
+++ b/cupyx/scipy/interpolate/_ndbspline.py
@@ -398,12 +398,12 @@ def colloc_nd(xvals, t, len_t, k):
     # Precompute the shape and strides of the coefficients array.
     # This would have been the NdBSpline coefficients; in the present context
     # this is a helper to compute the indices into the collocation matrix.
-    c_shape = tuple(len(t[d]) - k1_shape[d] for d in range(ndim))
+    c_shape = len_t - cupy.asarray(k1_shape, dtype=len_t.dtype)
 
     # The computation is equivalent to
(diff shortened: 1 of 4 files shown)


ev-br:
This PR now includes Edgar's fixes for NdBSpline.design_matrix (thanks!) and is passing all tests locally.

takagi:
Would you add `make_ndbspl` on scipy_interpolate.rst?

ev-br:
> Would you add make_ndbspl on scipy_interpolate.rst?

It's private in SciPy, so this would go beyond the scipy 1.13.0 API at least. Yes/no?

takagi:
> It's private in SciPy, so this would go beyond the scipy 1.13.0 API at least. Yes/no?

Thanks, we don't need it on the current document.

takagi (approved):
LGTM!

PR #8283, opened 2024-04-11
Preload cuTENSORMg

kmaehashi:
Support for cuTENSORMg has been added in #8212 but the library was not loaded. As a result `cupy_backends.cuda.libs.cutensor` import was failing as the library could not be found.

No backport required as cuTENSORMg support is new in v14.


takagi:
LGTM!

PR #8287, opened 2024-04-12
Implement “excluded” option in vectorize.py

### LINKED ISSUE #8166: cupy.vectorize currently does not support "excluded" option
### Description

At the moment I am trying to use cp.vectorize to improve the performance over np.vectorize. However the function that I am trying to vectorize accepts an array like argument that I want to be consistent for each function call, so I dont want to vectorize over this argument, here is a minimal working example:

```
import numpy as np
from bisect import bisect_left

def myfunc(a,b):
    return bisect_left(a, b)

vfunc = np.vectorize(myfunc, excluded = ['a'])
vfunc(a= np.array([1,2,3,4]),b=np.array([1.5,2.5]))
```

In other words, I would like the following code to work as expected:

```
import cupy as cp
from bisect import bisect_left

def myfunc(a,b):
    return bisect_left(a, b)

vfunc = cp.vectorize(myfunc, excluded = ['a'])
vfunc(a= cp.array([1,2,3,4]),b=cp.array([1.5,2.5]))
```

Note that `help(cp.vectorize)` produces the following output:

> Help on class vectorize in module cupy._functional.vectorize:
> class vectorize(builtins.object)
>  |  vectorize(pyfunc, otypes=None, doc=None, excluded=None, cache=False, signature=None)
>  |  
>  |  Generalized function class.
>  |  
>  |  .. seealso:: :class:`numpy.vectorize`
>  |  
>  |  Methods defined here:
>  |  
>  |  __call__(self, *args)
>  |      Call self as a function.
>  |  
>  |  __init__(self, pyfunc, otypes=None, doc=None, excluded=None, cache=False, signature=None)
>  |      Args:
>  |          pyfunc (callable): The target python function.
>  |          otypes (str or list of dtypes, optional): The output data type.
>  |          doc (str or None): The docstring for the function.
>  |          **excluded: Currently not supported.**
>  |          cache: Currently Ignored.
>  |          signature: Currently not supported.

I would like to request that the `excluded` option be implemented. 

In the meantime, I tried to workaround this by vectorizing a function with arbitrary number of arguments *args, but cp.vectorize does not support this either:

```
import cupy as cp
from bisect import bisect_left

def myfunc(b, *a):
    return bisect_left(a, b)
testlist = [4,5,6,7,8]
vfunc = cp.vectorize(myfunc)
vfunc(cp.array([1,2,3,4]),*testlist)
```
Output:
> NotImplementedError: `*args` is not supported currently.

**_Any other potential workarounds would be greatly appreciated!_**

### Additional Information

_No response_

ryan-kersten:
Why:
Fixes #8166 
TLDR: Implements a option numpy recently introduced in vectorize.py

How: Take the arguments for the vectorized function and copy the “excluded” parameters into a data structure that is the same shape as the non-”excluded” parameters. Pass these modified args into the function generated by ElementwiseKernel like normal.

### DIFF
--- a/cupy/_functional/vectorize.py
+++ b/cupy/_functional/vectorize.py
@@ -45,14 +45,15 @@ class vectorize(object):
         if otypes is not None:
             self.otypes = ''.join([numpy.dtype(t).char for t in otypes])
 
-        if excluded is not None:
-            raise NotImplementedError(
-                'cupy.vectorize does not support `excluded` option currently.')
-
         if signature is not None:
             raise NotImplementedError(
                 'cupy.vectorize does not support `signature`'
                 ' option currently.')
+        args = pyfunc.__code__.co_varnames[:pyfunc.__code__.co_argcount]
+        if excluded is not None:
+            self.excluded = \
+                [args.index(arg) for arg in args if arg in excluded]
+            # args are passed in positionally later, so this is best format
 
     @staticmethod
     def _get_body(return_type, call):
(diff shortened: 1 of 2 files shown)


asi1024:
This seems to only support for excluded arrays with `x.size == 1`.

RyanI70I:
I tried making 2 testcases with the excluded array being different shapes, but I wasn't able to find the problem

Do you have a test case in mind that exposes this?

asi1024:
In your two new test cases, the `excluded` option seems to be internally disabled. Could you add another test for inputs that cannot be broadcast?

PR #8289, opened 2024-04-12
ENH: cupyx/interpolate: port interp1d from scipy

ev-br:
Port `interp1d` routine from `scipy.interpolate`.
The routine itself is not very interesting: for `nearest`-like modes it delegates to `cupy.searchsorted`; for `cubic`-like modes it delegates to `make_interp_spline`. And it adds heaps and heaps of overhead on top.
It's public in SciPy though, and is used under the hood by `griddata`. So it's easier to port it and be done with it.

cross-ref https://github.com/cupy/cupy/issues/7186

PR #8220, opened 2024-03-04
Add NearestNDInterpolator to cupyx.scipy.interpolate

andfoy:
See #7186
Depends on #8208 for module ordering

Issue #7240, opened 2022-12-14
`TestBSpline::test_bspline_derivative_jumps` sometimes fail


kmaehashi:
https://ci.preferred.jp/cupy.linux.cuda112/118641/

```
00:23:27.206890 STDOUT 1590]	NumPy Version                : 1.21.6	
00:23:27.206972 STDOUT 1590]	SciPy Version                : 1.7.3	
...

01:11:21.153199 STDOUT 1590]	FAILED cupyx_tests/scipy_tests/interpolate_tests/test_bspline.py::TestBSpline::test_bspline_derivative_jumps[_param_3_{c=[[-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1]], extrapolate=True}]	
```

<details>

```
01:11:20.600785 STDOUT 1590]	_ TestBSpline.test_bspline_derivative_jumps[_param_3_{c=[[-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1]], extrapolate=True}] _	
01:11:20.600792 STDOUT 1590]		
01:11:20.600802 STDOUT 1590]	args = (<<cupyx_tests.scipy_tests.interpolate_tests.test_bspline.TestBSpline object at 0x7fb06ee07e50>  parameter: {'c': [[-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1]], 'extrapolate': True}>,)	
01:11:20.600809 STDOUT 1590]	kw = {}	
01:11:20.600817 STDOUT 1590]	cupy_result = [array([5.48813486, 6.40238698, 5.83469979]), array([5.48813486, 6.40238698, 5.83469979]), array(4.23654795), array(0.), array([-0.74950663, -0.38586775]), array([-0.74950663, -0.38586775]), ...]	
01:11:20.600823 STDOUT 1590]	cupy_error = None	
01:11:20.600832 STDOUT 1590]	numpy_result = [array([5.48813486, 6.40238698, 5.83469979]), array([5.48813486, 6.40238698, 5.83469979]), array(4.23654795), array(0.), array([-0.74950663, -0.38586775]), array([-0.74950663, -0.38586775]), ...]	
01:11:20.600837 STDOUT 1590]	numpy_error = None	
[...]

PR #8290, opened 2024-04-12
ENH: cupyx/scipy/interpolate: add `griddata`

ev-br:
Port `griddata` function from `scipy.interpolate`. This is a pure python convenience wrapper of`LinearNDInterpolator`, `NearestNDInterpolator` and `CloughTocher2DInterpolator`, so no need to benchmark performance.

This PR is on top of https://github.com/cupy/cupy/pull/8220 for the NearestNDInterpolator and https://github.com/cupy/cupy/pull/8289 for `interp1d`. Will need a rebase or two when these two PRs land.

### DIFF
--- a/cupyx/scipy/interpolate/__init__.py
+++ b/cupyx/scipy/interpolate/__init__.py
@@ -14,6 +14,7 @@ from cupyx.scipy.interpolate._interpnd import LinearNDInterpolator  # NOQA
 from cupyx.scipy.interpolate._interpnd import (  # NOQA
     CloughTocher2DInterpolator)  # NOQA
 from cupyx.scipy.interpolate._ndgriddata import NearestNDInterpolator  # NOQA
+from cupyx.scipy.interpolate._ndgriddata import griddata  # NOQA
 
 # 1-D Splines
 from cupyx.scipy.interpolate._bspline import BSpline, splantider, splder  # NOQA
--- a/cupyx/scipy/interpolate/_ndgriddata.py
+++ b/cupyx/scipy/interpolate/_ndgriddata.py
@@ -4,7 +4,8 @@ Convenience interface to N-D interpolation
 
 import cupy
 from cupyx.scipy.interpolate._interpnd import (
-    NDInterpolatorBase, _ndim_coords_from_arrays)
+    NDInterpolatorBase, _ndim_coords_from_arrays,
+    LinearNDInterpolator, CloughTocher2DInterpolator)
 from cupyx.scipy.spatial import KDTree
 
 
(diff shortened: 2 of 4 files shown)


ev-br:
Both dependencies are merged, the PR is rebased on main and is ready for review.

takagi (approved):
LGTM! Waiting for CI pass.

takagi:
Seems that CI fails with illegal memory access. Does this occurs on your local? Disabling CuPy's memory pool `cupy.cuda.set_allocator(None)` may help.
```
E   cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered
```

ev-br:
Sadly, no, I don't see this locally. `$ pytest tests/cupyx_tests/scipy_tests/interpolate_tests/ -v` passes for me (on CUDA 12 and scipy 1.13) modulo a test failure  due to a unrelated scipy 1.13 change, https://github.com/scipy/scipy/pull/20364.

Also given that this PR only adds a thin pure python wrapper, the the root problem is likely somewhere else. Where does the error originate?

takagi:
Seems that its first occurence in CI log is `TestGriddata.test_fill_value `, but as you say it may not the root problem. I will also check this issue.

ev-br:
TBH, I don't see how a pure python addition of this PR can trigger this. The only possiblility is that a nan assignment is broken somehow, in this line: https://github.com/cupy/cupy/pull/8290/files#diff-e6ca1b203fa4c1b7e7f6170aa16844edb19aa250d65f43d99d4bd3ee26187b36R256 with the default of `fill_value=cupy.nan`. But this is definitely not the only place nans appear, so it's unlikely too.

kmaehashi:
I've tried running `compute-sanitizer --tool memcheck python -m pytest -v -s -k test_fill_value tests/cupyx_tests/scipy_tests/interpolate_tests/test_ndgriddata.py` on my local with memory pool disabled by commenting out [these lines](https://github.com/cupy/cupy/blob/d8df1f0405bb645c84274551a7630cf0247d6dd5/cupy/__init__.py#L877-L878). It's not always reproducing (it even sometimes hangs) but I got this error reported:

```
========= Invalid __global__ read of size 4 bytes
=========     at kerFindClosestTri(Point2 *, int, Tri *, int, TriOpp *, long long *, unsigned int *, Point2 *, Point2 *, const double *, const double *, Point2 *, Point2 *, double, bool, int *, double *)+0x85d0
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x7f2ec2a007f4 is out of bounds
=========     and is 12 bytes before the nearest allocation at 0x7f2ec2a00800 of size 48 bytes
=========     Host Frame: [0x32e94f]
```

takagi:
@andfoy Can I ask you to check if `kerFindClosestTri` could make any out-of-bounds access again?

andfoy:
Thanks for the report @takagi, the reported segfault should be fixed by https://github.com/cupy/cupy/pull/8363

takagi:
Now that #8363 is merged, let me kick the CI again.

andfoy:
@takagi @ev-br, I think this one requires a rebase? I tested with `compute-sanitizer` locally and no errors were flagged

image

ev-br:
rebased

takagi:
#8363 is merged and used in CI. I also tested with compute-sanitizer locally and it makes no report for several trials.

takagi:
While I make a wider range of tests, I've found that compute-sanitizer reports `d_boor` kernel may have illegal memory access. Would you check this when it is convenient to you, @andfoy?

```
$ compute-sanitizer --tool memcheck --padding 32 python -m pytest -v  tests/cupyx_tests/scipy_tests/interpolate_tests/test_bspline2.py  -k test_minimum_points_and_deriv
```

```
========= COMPUTE-SANITIZER
============================= test session starts ==============================
platform linux -- Python 3.10.13, pytest-7.4.3, pluggy-1.3.0 -- /home/mtakagi/cupy.cuda-12.1/venv/bin/python
cachedir: .pytest_cache
rootdir: /home/mtakagi/cupy.cuda-12.1
configfile: setup.cfg
collecting ... collected 73 items / 72 deselected / 1 selected

tests/cupyx_tests/scipy_tests/interpolate_tests/test_bspline2.py::TestInterp::test_minimum_points_and_deriv
========= Invalid __global__ write of size 8 bytes
=========     at 0xa70 in void d_boor<double>(const double *, const T1 *, int, int, const double *, const long long *, T1 *, double *, int, int, int)
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x7f85d8e03620 is out of bounds
=========     and is 1 bytes after the nearest allocation at 0x7f85d8e03600 of size 32 bytes
...
```

ev-br:
makes me wonder if https://github.com/cupy/cupy/issues/7240 is related.

takagi:
I will kick the CI again once #8388 is merged.

takagi:
I'm trying compute-sanitizer again.

Issue #6018, opened 2021-11-05
Doc: `cupyx.profiler` is not in the module list of API reference


kmaehashi:

Kaushik-Iyer:
Should I work on this

kmaehashi:
Sure, go ahead 😄

Kaushik-Iyer:
Can I not work on this without installing cuda? while making the html I'm encountering the module not found error for cupy

Taejin101:
Hi is there any guide on how to write .rst file

arkdong:
Hi @kmaehashi, where and how did the documentation page [https://docs.cupy.dev/en/stable/py-modindex.html](Python Module Index) being generated? I didn't see any .rst file in the folder `docs/source/reference` that `* :ref: modindex` in `docs/source/reference/index.rst` referenced to. Or is it auto generated by Sphinx?

kmaehashi:
The doc is generated automaticaly by Sphinx.

leofang:
Don't we already have it in the docs?
https://github.com/cupy/cupy/blob/150c903890bc14ec6c2432813374d9fde1b7528c/docs/source/reference/ext.rst?plain=1#L36-L44

arkdong:
@leofang `ext.rst` are the ones on the page API reference [https://docs.cupy.dev/en/stable/reference/ext.html](url). Isn't the python module index page [https://docs.cupy.dev/en/stable/py-modindex.html](url) that needs to be updated with `cupyx.profiler`?

leofang:
I see, my guess is we need to insert `.. module:: cupyx.profiler` to the section that I linked to, and perhaps modindex would pick it up automatically, feel free to give it a shot!

arkdong:
I will try to implemented, thank you for your help!

arkdong:
Hi, inserting `.. module:: cupyx.profiler` does not solve the problem. When building the html page with `sphinx-build -b html . _build`, it threw an exception that no module named `cupyx.profiler.cupy`. While for `cupyx.optimizing`, inserting `.. module:: cupyx.optimizing` works perfectly.

Therefore I would like to ask is a module that the error mention a .rst file, or a directory/python file?

PR #8297, opened 2024-04-17
Add `cupyx.profiler` to python module API list

arkdong:
I've add `cupyx.profiler` to python module API list via adding `.. module:: cupyx.profiler` in the file `est.rst`, regarding to issue #6018 

Initially just adding the line of code threw an exception that no module named `cupyx.profiler`, but after trying different implementation I found that when moving the hole code piece to the bottom of the file above `.. _kernel_param_opt:` will work.

I've tested using `sphinx-build -b html . _build` to build the html page, and opened with `xdg-open _build/py-modindex.html`. The hyperlink of profiler work as expected, and the code below for `cupyx.optimizing` will not be affected. The only thing that changed on the page for [cupy-specific fuctions](https://docs.cupy.dev/en/stable/reference/ext.html#cupy-specific-functions), is the order of each section.

### DIFF
--- a/docs/source/_comparison_generator.py
+++ b/docs/source/_comparison_generator.py
@@ -259,6 +259,22 @@ def generate():
         'Sparse Matrices',
         'scipy.sparse', 'cupyx.scipy.sparse', 'SciPy', exclude=['test'])
     buf += _section(
+        'Sparse Matrices: COOrdinate Format',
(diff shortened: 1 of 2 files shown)


kmaehashi:
Thanks, but I think this should be doable without moving the block. When covering functions from multiple modules in a single page `.. currentmodule::` is needed to switch the module context.

PR #8312, opened 2024-04-28
improve efficiency of binary morphology functions by accepting a tuple for `footprint`

grlee77:
This MR backports an efficiency improvement for binary morphology functions that were made in cuCIM's vendored copy of `cupyx.scipy.ndimage` code. See additional background in https://github.com/cupy/cupy/pull/8126#issuecomment-2081554814

The CUDA kernels themselves do not have to change to support this as we already previously merged support for not providing a weights array. The changes here just allow the higher level binary morphology functions to make use of that existing optimization.

For NumPy, allocating a small (e.g. 3x3) binary array of ones takes on the order of 1 microsecond, so this would be a micro-optimization if proposed for SciPy itself. For CuPy, though, the array creation time is > 10 microseconds and also involves synchronization.  In addition to the synchronization during array creation, the change here will also avoid a second synchronization that occurs during an internal check of whether all elements of the footprint array are true.

### DIFF
--- a/cupyx/scipy/ndimage/_filters.py
+++ b/cupyx/scipy/ndimage/_filters.py
@@ -769,6 +769,10 @@ def _min_or_max_filter(input, size, ftprnt, structure, output, mode, cval,
     # structure is used by morphology.grey_erosion() and grey_dilation()
     # and not by the regular min/max filters
 
+    if isinstance(ftprnt, tuple) and size is None:
+        size = ftprnt
+        ftprnt = None
+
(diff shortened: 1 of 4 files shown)


takagi (approved):
LGTM!

grlee77:
Let me see if I can reproduce and resolve the failures locally (I think I only ran the full set of morphology tests locally previously, but failures here are mostly in min/max/rank filters)

takagi (approved):

PR #8315, opened 2024-04-30
Use weakref.finalize instead of __del__ for RandomState._generator destruction

mroeschke:
Sometimes upon interpreter exit and using `RandomState`, the following exception can occur

```python
Exception ignored in: <function RandomState.__del__ at 0x7f0cc41dadd0>
  File "/home/coder/.conda/envs/rapids/lib/python3.10/site-packages/cupy/random/_generator.py", line 65, in __del__
ModuleNotFoundError: import of cupy_backends.cuda.libs halted; None in sys.modules
```

In order to more robustly destroy `RandomState._generator` upon exit, a `weakref.finalize` should be used https://docs.python.org/3/library/weakref.html#comparing-finalizers-with-del-methods


bdice:
Thanks for this @mroeschke! We see this a lot in RAPIDS CI. A fix would be really helpful.

leofang:
@mroeschke have you found a reliable way to reproduce the shundown warning? This treatment has a long history (the earliest I could find is https://github.com/cupy/cupy/issues/2806) but it seems `wearkref` had never been considered for some reason. If we can reproduce the behavior locally and then check with your patch, it'd be great!

mroeschke:
Sorry for the delay here @leofang. Here's a reproducer

Before:
```python
>>> import cupy
>>> import sys
>>> cupy.__version__
'13.0.0rc1'
>>> rng = cupy.random.RandomState(1)
>>> sys.exit(1)
Exception ignored in: <function RandomState.__del__ at 0x7fc40cd675b0>
  File "/home/nfs/mroeschke/miniforge3/lib/python3.10/site-packages/cupy/random/_generator.py", line 65, in __del__
ImportError: sys.meta_path is None, Python is likely shutting down
``` 

After (with the following local modification)

```diff
diff --git a/cupy/random/_generator.py b/cupy/random/_generator.py
index 04e6c67b0..f2ddb5145 100644
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -59,7 +59,11 @@ class RandomState(object):
         if method is None:
             method = curand.CURAND_RNG_PSEUDO_DEFAULT
         self._generator = curand.createGenerator(method)
-        self._finalizer = weakref.finalize(self, curand.destroyGenerator, self._generator)
+        
+        def destroyer(*args, **kwargs):
+            print("destroying generator")
+            curand.destroyGenerator(*args)
+        self._finalizer = weakref.finalize(self, destroyer, self._generator)
         self.method = method
         self.seed(seed)
```

```python
>>> import cupy
>>> import sys
>>> rng = cupy.random.RandomState(1)
>>> sys.exit(1)
destroying generator
```

mroeschke:
@leofang following up here to see if you have any other comments

leofang (approved):
Thanks @mroeschke, sorry for delay, I can reproduce with https://github.com/cupy/cupy/pull/8315#issuecomment-2123630361.

leofang:
Unfortunately I cannot trigger the CI for first-time contributors, @kmaehashi would you mind kicking off the CI?

leofang:
@mroeschke could you check the linter error?

cupy/random/_generator.py
```diff
@@ -58,18 +59,10 @@ def __init__(self, seed=None, method=None):
         if method is None:
             method = curand.CURAND_RNG_PSEUDO_DEFAULT
         self._generator = curand.createGenerator(method)
+        self._finalizer = weakref.finalize(self, curand.destroyGenerator, self._generator)
```
mroeschke, line 62:
```suggestion
        self._finalizer = weakref.finalize(
            self, curand.destroyGenerator, self._generator)
```

leofang:
@kmaehashi @asi1024 this is ready

kmaehashi:
Hi, sorry for keeping you waited! LGTM :+1:

leofang:
> This treatment has a long history (the earliest I could find is #2806) but it seems `wearkref` had never been considered for some reason. 

Circling back to this. For Python classes, using the linked treatment (`is_shutting_down`) is preferred over weakref, because @emcastillo and I found that the latter's performance is not ideal. `RandomState` is unlikely to appear in a hot loop, so this PR might be OK, but generally using  `is_shutting_down` is probably better.

PR #8314, opened 2024-04-30
WIP: support NumPy 2.0 : NEP 52 (Python API cleanup)

ev-br:
Work towards https://github.com/cupy/cupy/issues/8306.

The current checklist is

- [x] make sure `cupy` imports with `numpy==2.0.0rc1`
- [x] make sure `pytest tests/cupy` passes the collection stage
- [x] make sure `pytest tests/cupyx` passes the collection stage
- [ ] make main `cupy` and `numpy` namespaces match
- [ ] make `ndarray` methods match
- [ ] match submodule namespaces:
     - [ ] `numpy.core`,
     - [ ] `numpy._lib` 
     - [ ] `numpy.exceptions`
     - [ ] `numpy.strings`

Issue #8306, opened 2024-04-24
Support NumPy 2.0


kmaehashi:
Make CuPy APIs compatible with NumPy 2.0 🚀 Ref: #8269

Major tasks include:

* New Type Promotion Rules for Python scalars ([NEP 50](https://numpy.org/neps/nep-0050-scalar-promotion.html))
* Python API Cleanup for NumPy 2.0 ([NEP 52](https://numpy.org/neps/nep-0052-python-api-cleanup.html))
* Adopting Array API Compatibility in the main NumPy namespace ([NEP 56](https://numpy.org/neps/nep-0056-array-api-main-namespace.html))

rgommers:
Just to make sure: the plan is for 14.0.0 to completely match NumPy 2.0? Which probably means only testing against NumPy 2.0 in all CI, since there is no way to match 2.0 and 1.26 at the same time?

kmaehashi:
Yes, our current plan is to match CuPy v14 API with NumPy 2.0, and only test against NumPy 2.0+.

ev-br:
I'm going to work towards this, starting with the python api, NEP 52.

kmaehashi:
Regarding NumPy 2.0 support, CuPy currently:

1. requires [`numpy>=1.22,<1.29`](https://github.com/cupy/cupy/blob/6eb9fddaa842c67fe46df88b5444f59147a33102/setup.py#L26), and
2. imports APIs to be removed in NumPy 2.0.

This means users cannot install/import CuPy in NumPy v2 environment.

I'm thinking it's better to ship CuPy v13.2.0 release that includes:

1. updated requirement `numpy>=1.22,<2.3`
2. make imports NumPy 2.0-safe, so that CuPy v13 can be imported with both NumPy v1/v2 environment

before the NumPy 2 ship date.

rgommers:
That would be great. I think not matching exactly all the promotion semantics and missing a few new functions is fine, as long as it imports correctly indeed.

kmaehashi:
We have received several bug reports regarding NumPy v2 support in CuPy v13. They will be tracked under [np2-v13](https://github.com/cupy/cupy/labels/np2-v13 ) label.

agm-eratosth:
Per https://github.com/rapidsai/build-planning/issues/38 this should be supported in 24.10

jakirkham:
> Per https://github.com/rapidsai/build-planning/issues/38 this should be supported in 24.10

[...]

PR #8323, opened 2024-05-09
NEP50

ev-br:
Split off https://github.com/cupy/cupy/pull/8314 to keep the diff sizes under control.

Adapt the codebase to NEP 50 (https://numpy.org/neps/nep-0050-scalar-promotion.html). Unlike gh-8314, this PR can be run under NumPy 1.x using 

```
$ export NPY_PROMOTION_STATE=weak
$ pytest ...
```

The main change is that under NEP 50, we have to keep track of "weak" scalars:

```
In [5]: (np.int8(1) + np.uint8(1)).dtype
Out[5]: dtype('int16')

In [6]: (1 + np.uint8(1)).dtype
Out[6]: dtype('uint8')
```

Implementation-wise, CuPy seems to convert python scalars to numpy scalars rather early in the game; thus we keep track of which arguments of ufuncs are python scalars, and take this into account when selecting ufunc loops. 
So far am only doing it in ufuncs, will need to see if ElementwiseKernels / fusion loops / reductions are affected. 

The goal is to make tests pass locally first with numpy 1.26.x & weak promotion rules, so that one can think of switching this on the CI. So far it's not even close: need to weed away lots of assumptions about the promotion rules in code and tests.

### DIFF
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -1614,33 +1614,15 @@ cdef class _Ops:
 
             # check for overflow in operands. Consider `np.uint8(1) + 300`.
             # Per NEP 50 this raises OverflowError because 300 overflows uint8.
-            # We can only check it after the output dtype is known, which is
-            # op.out_types here.
+            # We can only check it after the operand types are known
             out_type = op.out_types[0]
-            if issubclass(out_type, numpy.integer):
-                iinfo = numpy.iinfo(out_type)
-                for i in range(len(in_args)):
-                    arg = in_args[i]
-                    typ = arg.dtype.type
+            for i, weak_t in enumerate(weaks):
+                if weak_t is not int:
+                    continue
+                # Note: For simplicity, check even if output is actually a float
+                integer_argument = int(in_args[i])
+                out_type(integer_argument)  # Check if user input fits loop
 
-                    if weaks[i] and issubclass(typ, numpy.integer):
-                        if issubclass(typ, numpy.signedinteger):
-
-                            if issubclass(out_type, numpy.signedinteger):
-                                ok = (<int64_t>iinfo.min <=
-                                      <int64_t>arg <=
-                                      <int64_t>iinfo.max)
-                            else:
-                                ok = ((arg >= 0) and
-                                      (<uint64_t>arg <= <uint64_t>iinfo.max))
-
-                        elif issubclass(typ, numpy.unsignedinteger):
-                            # arg >= 0
-                            ok = <uint64_t>arg <= <uint64_t>iinfo.max
-
-                        if not ok:
-                            raise OverflowError(f"Python integer {arg} out of "
-                                                f"bounds for {out_type}.")
             return op
 
         if dtype is None:
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -10,7 +10,6 @@ from cupy import _util
 cimport cython  # NOQA
 
 from libcpp cimport vector
-from libc.stdint cimport uint64_t, int64_t
 
 from cupy.cuda cimport device
 from cupy.cuda cimport function
@@ -1097,7 +1096,9 @@ cdef inline int _get_kind_score(type kind):
     return 3
 
 
-cdef inline bint _check_should_use_weak_scalar(tuple in_types, tuple weaks) except? -1:
+cdef inline bint _check_should_use_weak_scalar(
+    tuple in_types, tuple weaks
+) except? -1:
     """The promotion strategy of finding the first matching loop is not
     equipped to deal with correct promotion when mixing weak scalars and
     arrays/strong types.
@@ -1107,8 +1108,8 @@ cdef inline bint _check_should_use_weak_scalar(tuple in_types, tuple weaks) exce
     This prevents e.g. `uint8(1) + 0.` from picking a float64 loop.
     """
     cdef int kind, max_array_kind, max_scalar_kind
-    cdef bint all_scalars
-    all_scalars = True
+    cdef bint all_scalars_or_arrays
+
     max_array_kind = -1
     max_scalar_kind = -1
     for in_t, w_t in zip(in_types, weaks):
@@ -1625,7 +1626,7 @@ cdef class _Ops:
             for i in range(len(in_args)):
                 if weaks[i] is not int:
                     continue
-                # Note: For simplicity, check even if output is actually a float
+                # Note: For simplicity, check even if `in_type` is e.g. float
                 integer_argument = int(in_args[i])
                 in_type = op.in_types[i]
                 in_type(integer_argument)  # Check if user input fits loop
@@ -1638,7 +1639,8 @@ cdef class _Ops:
                         (dtype, name))
 
     cpdef _Op _guess_routine_from_in_types(
-            self, tuple in_types, tuple weaks=None, object can_cast=_numpy_can_cast
+            self, tuple in_types, tuple weaks=None,
+            object can_cast=_numpy_can_cast
     ):
         cdef _Op op
         cdef tuple op_types
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -1592,10 +1592,17 @@ cdef class _Ops:
             self, str name, dict cache, list in_args, tuple weaks, dtype,
             _Ops out_ops):
         cdef _Ops ops_
+
+        if weaks is None:
+            weaks = (False,) * len(in_args)
+
         if dtype is None:
             assert all([isinstance(a, (_ndarray_base, numpy.generic))
                         for a in in_args])
             in_types = tuple([a.dtype.type for a in in_args])
+
+            if not _check_should_use_weak_scalar(in_types, weaks):
+                weaks = (False,) * len(in_args)
 
             op = cache.get((in_types, weaks), ())
             if op is ():
@@ -1615,13 +1622,13 @@ cdef class _Ops:
             # check for overflow in operands. Consider `np.uint8(1) + 300`.
             # Per NEP 50 this raises OverflowError because 300 overflows uint8.
             # We can only check it after the operand types are known
-            out_type = op.out_types[0]
-            for i, weak_t in enumerate(weaks):
-                if weak_t is not int:
+            for i in range(len(in_args)):
+                if weaks[i] is not int:
                     continue
                 # Note: For simplicity, check even if output is actually a float
                 integer_argument = int(in_args[i])
-                out_type(integer_argument)  # Check if user input fits loop
+                in_type = op.in_types[i]
+                in_type(integer_argument)  # Check if user input fits loop
 
             return op
 
@@ -1638,15 +1645,12 @@ cdef class _Ops:
         cdef Py_ssize_t n = len(in_types)
         cdef Py_ssize_t i
 
-        if weaks is None or not _check_should_use_weak_scalar(in_types, weaks):
-            weaks = (False,) * n
-
         for op in self.ops:
             op_types = op.in_types
             for i in range(n):
                 it = in_types[i]
                 ot = op_types[i]
-                weak_t = weaks[i]
+                weak_t = weaks[i] if weaks is not None else False
 
                 # XXX: Remove assert (reachable only for pre NEP 50 logic)
                 assert(not isinstance(it, tuple))
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -10,6 +10,7 @@ from cupy import _util
 cimport cython  # NOQA
 
 from libcpp cimport vector
+from libc.stdint cimport uint64_t, int64_t
 
 from cupy.cuda cimport device
 from cupy.cuda cimport function
@@ -104,6 +105,33 @@ cdef inline int _get_kind_score(int kind):
     return -1
 
 
+cdef inline int _get_kind_score2(dtype):
+    """Use different scores for floating real and complex."""
+    _dct = {
+        cupy.bool_: 0,
+        # unsigned int
+        cupy.uint8: 1,
+        cupy.uint16: 1,
+        cupy.uint32: 1,
+        cupy.uint64: 1,
+        cupy.ulonglong: 1,
+        # signed int
+        cupy.int8: 1,
+        cupy.int16: 1,
+        cupy.int32: 1,
+        cupy.int64: 1,
+        cupy.longlong: 1,
+        # float
+        cupy.float16: 2,
+        cupy.float32: 2,
+        cupy.float64: 2,
+        # complex
+        cupy.complex64: 3,
+        cupy.complex128: 3,
+        }
+    return _dct[dtype]
+
+
 @cython.profile(False)
 cpdef inline _check_peer_access(_ndarray_base arr, int device_id):
     if arr.data.device_id == device_id:
@@ -125,6 +153,7 @@ cpdef inline _check_peer_access(_ndarray_base arr, int device_id):
 
 
 cdef inline _preprocess_arg(int dev_id, arg, bint use_c_scalar):
+    is_weak = False
     if isinstance(arg, _ndarray_base):
         s = arg
         _check_peer_access(<_ndarray_base>s, dev_id)
@@ -137,16 +166,17 @@ cdef inline _preprocess_arg(int dev_id, arg, bint use_c_scalar):
         s = arg.__cupy_get_ndarray__()
         _check_peer_access(<_ndarray_base>s, dev_id)
     else:  # scalars or invalid args
+        is_weak = type(arg) in [int, float, complex]
         if use_c_scalar:
             s = _scalar.scalar_to_c_scalar(arg)
         else:
             s = _scalar.scalar_to_numpy_scalar(arg)
         if s is None:
             raise TypeError('Unsupported type %s' % type(arg))
-    return s
+    return s, is_weak
 
 
-cdef list _preprocess_args(int dev_id, args, bint use_c_scalar):
+cdef tuple _preprocess_args(int dev_id, args, bint use_c_scalar):
     """Preprocesses arguments for kernel invocation
 
     - Checks device compatibility for ndarrays
@@ -155,9 +185,12 @@ cdef list _preprocess_args(int dev_id, args, bint use_c_scalar):
       - If use_c_scalar is False, into NumPy scalars.
     """
     cdef list ret = []
+    cdef list weaks = []
     for arg in args:
-        ret.append(_preprocess_arg(dev_id, arg, use_c_scalar))
-    return ret
+        p_arg, is_weak = _preprocess_arg(dev_id, arg, use_c_scalar)
+        ret.append(p_arg)
+        weaks.append(is_weak)
+    return ret, tuple(weaks)
 
 
 cdef list _preprocess_optional_args(int dev_id, args, bint use_c_scalar):
@@ -169,11 +202,12 @@ cdef list _preprocess_optional_args(int dev_id, args, bint use_c_scalar):
       - If use_c_scalar is False, into NumPy scalars.
     """
     cdef list ret = []
+    cdef list weaks = []
     for arg in args:
         if arg is None:
             ret.append(None)
         else:
-            ret.append(_preprocess_arg(dev_id, arg, use_c_scalar))
+            ret.append(_preprocess_arg(dev_id, arg, use_c_scalar)[0])
     return ret
 
 
@@ -866,7 +900,7 @@ cdef class ElementwiseKernel:
                 return arg.__cupy_override_elementwise_kernel__(
                     self, *args, **kwargs)
         dev_id = device.get_device_id()
-        arg_list = _preprocess_args(dev_id, args, True)
+        arg_list, weaks = _preprocess_args(dev_id, args, True)
 
         out_args = arg_list[self.nin:]
         # _broadcast updates shape
@@ -1084,42 +1118,10 @@ cdef function.Function _get_ufunc_kernel(
         loop_prep=loop_prep)
 
 
-cdef inline bint _check_should_use_min_scalar(list in_args) except? -1:
-    cdef int kind, max_array_kind, max_scalar_kind
-    cdef bint all_scalars
-    all_scalars = True
-    max_array_kind = -1
-    max_scalar_kind = -1
-    for i in in_args:
-        kind = _get_kind_score(ord(i.dtype.kind))
-        if isinstance(i, _ndarray_base):
-            all_scalars = False
-            max_array_kind = max(max_array_kind, kind)
-        else:
-            max_scalar_kind = max(max_scalar_kind, kind)
-    return (max_scalar_kind != -1 and
-            not all_scalars and
-            max_array_kind >= max_scalar_kind)
-
-
 cdef dict _mst_unsigned_to_signed = {
     i: (numpy.iinfo(j).max, (i, j))
     for i, j in [(numpy.dtype(i).type, numpy.dtype(i.lower()).type)
                  for i in "BHILQ"]}
-cdef _numpy_min_scalar_type = numpy.min_scalar_type
-
-cdef _min_scalar_type(x):
-    # A non-negative integer may have two locally minimum scalar
-    # types: signed/unsigned integer.
-    # Return both for can_cast, while numpy.min_scalar_type only returns
-    # the unsigned type.
-    t = _numpy_min_scalar_type(x)
-    dt = t.type
-    if t.kind == 'u':
-        m, dt2 = <tuple>_mst_unsigned_to_signed[dt]
-        if x <= m:
-            return dt2
-    return dt
 
 
 cdef class ufunc:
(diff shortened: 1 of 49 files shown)


leofang:
cc: @seberg for vis

kmaehashi:
Thank you for tackling this, @ev-br! As for OOMs, I'd suggest excluding slow-labeled tests (`pytest -m 'not slow'`) during development. They are for testing functionality against huge ndarray and I think they are safely ignored when working on type promotional rules.

cupy/linalg/_einsum.py
```diff
@@ -150,6 +150,9 @@ def _parse_einsum_input(args):
                 msg + ' operands provided to einstein sum function than '
                 'specified in the subscripts string')
 
+        # NumPy ignores 'weak' scalars and always returns i64/f64
+        operands = [cupy.asarray(op) for op in operands]
```
seberg, line 154: Maybe should change that one day, but not today.  `i64` I assume is true, `f64` I would think not if inputs are float32?

I suppose since scalars are an odd input to einsum, not changing it should be OK or maybe we should fix this one in NumPy?

tests/cupy_tests/random_tests/test_sample.py
```diff
@@ -114,7 +114,7 @@ class TestRandintDtype(unittest.TestCase):
     def test_dtype(self, dtype):
         size = (1000,)
         low = numpy.iinfo(dtype).min
-        high = numpy.iinfo(dtype).max + 1
+        high = numpy.iinfo(dtype).max
```
seberg, line 117: Hmmm, this change the test, Maybe need at least a new xfail here since that should work?

cupy/_core/_kernel.pyx
```diff
@@ -1584,21 +1613,18 @@ cdef class _Ops:
...
-                    else _min_scalar_type(a)
-                    for a in in_args])
-            else:
-                in_types = tuple([a.dtype.type for a in in_args])
-            op = cache.get(in_types, ())
+            assert all([isinstance(a, (_ndarray_base, numpy.generic))
+                        for a in in_args])
+            in_types = tuple([a.dtype.type for a in in_args])
+
+            op = cache.get((in_types, weaks), ())
```
seberg, line 1624: Makes sense, I suspect one could use `in_types[weak_pos] = int` or so also here, but I think using the `weak` tuple works well for starters.
(I might want to push for a cleanup after/with my other PR, but this is more urgent.)

cupy/_core/_kernel.pyx
```diff
@@ -1584,21 +1613,18 @@ cdef class _Ops:
         return _Ops(tuple(ops_))
 
     cpdef _Op guess_routine(
-            self, str name, dict cache, list in_args, dtype, _Ops out_ops):
+            self, str name, dict cache, list in_args, tuple weaks, dtype,
+            _Ops out_ops):
         cdef _Ops ops_
         if dtype is None:
-            use_raw_value = _check_should_use_min_scalar(in_args)
```
seberg, line 1590: Not for me to decide.  If we wish to support both 1.x and 2.x (for the most part), it could make sense to keep the `_check_should_use_min_scalars` here.

As a very subtle point, NumPy still (usually) uses the same check as in `if all(weak): weak = [False] * len(weak)`.

(Which is arguably annoying, and only has an effect for weird integer inputs, though, so I think it is OK to ignore it.)

cupy/_core/_kernel.pyx
```diff
@@ -1610,6 +1636,35 @@ cdef class _Ops:
...
+                                ok = ((arg >= 0) and
+                                      (<uint64_t>arg <= <uint64_t>iinfo.max))
+
+                        elif issubclass(typ, numpy.unsignedinteger):
+                            # arg >= 0
+                            ok = <uint64_t>arg <= <uint64_t>iinfo.max
+
+                        if not ok:
+                            raise OverflowError(f"Python integer {arg} out of "
+                                                f"bounds for {out_type}.")
```
seberg, line 1667: This looks like a lot of code that really ends up doing the same as: `out_type(int(arg))` or `out_type(arg.item())`?
I suppose one could check safe-cast first, although I am not sure it would be faster.

cupy/_core/_kernel.pyx
```diff
@@ -1618,20 +1673,49 @@ cdef class _Ops:
...
+                if weaks[i] and it == numpy.complex128 and max_prec == 2:
+                    it = numpy.complex64
+                    is_weak = False
+
                 if isinstance(it, tuple):
+                    # XXX: account for weaks
                     if not can_cast(it[0], ot) and not can_cast(it[1], ot):
                         break
-                elif not can_cast(it, ot):
+                elif not is_weak and not can_cast(it, ot):
```
seberg, line 1718: I know it is a bit slow, but it should be cached, so I would suggest replacing the full above logic logic with:
```
try:
    safe = np.result_type(weak(0), ot) == ot
except TypeError:
    safe = False
```
Might not be the prettiest ever, but I think it is preferable to custom implementation of the category logic and more flexible for future additions.

I am suggesting another trick:  Let's make `weaks = [False, int, float, complex]` i.e. store either `False` _or_ the type, so that we can use the type information directly from there?

cupy/_core/_routines_math.pyx
```diff
@@ -1037,11 +1037,15 @@ _subtract = create_arithmetic(
...
+# NB: Cannot define loops with short ints in the NEP 50 world. Consider
+# `cupy.arange(3, dtype=cp.uint8) / (-2)`. It would select the 'BB->d' loop,
+# and the kernel would have a declaration `uint8_t in1;`, this converts
+# -2 to uint8_t at initialization (modulo UINT8_MAX, likely).
 _true_divide = create_ufunc(
     'cupy_true_divide',
-    ('bb->d', 'BB->d', 'hh->d', 'HH->d', 'ii->d', 'II->d', 'll->d', 'LL->d',
-     'qq->d', 'QQ->d', 'ee->e', 'ff->f', 'dd->d', 'FF->F', 'DD->D'),
-    'out0 = (out0_type)in0 / (out0_type)in1',
+    ('qq->d', 'QQ->d',
```
seberg, line 1046: Right, this has a slight miss for mixed signs, but that seems OK.

In principle it would be nice to solve this with a custom promotion step for the ufunc.  This could be very nice, because for (almost!) all NumPy ufunc, we should be able to at least try and use:
```
dtypes = ufunc.resolve_dtypes()
```

seberg (commented):
Thanks so much for starting this @ev-br!  Sorry about the slower followup on our discussions.
This is nicer and really quite minimal, and I think we can make it even more minimal!

As said below, I think passing the `weak` tuple is probably the pragmatic solution right now and may well be the best solution in the long run also. 

A few notes (some of which are inline):
* It might be nice to have at least a single per-ufunc/kernel "promotion" step.  I.e. a custom `guess_loop`.  That would solve many issues, because you could call `np.ufunc.resolve_dtypes()`.  I will note that this function *is* currently slow and also is only available since 1.24 or so.
* I have suggested some simplifications, which I *think* should make things even shorter (I may be missing reasons for why not though).
* In NumPy I had to add some "very special" paths unfortunately.  I think it makes sense to skip that for this PR.  However, it would probably make sense to at least for `uint_array > -1` add a special path to make it work.
  (i.e. if it is tricky to do for the comparison ufuncs, make it work for the comparisons.  For NumPy downstream testing code ran into it often enough and it is somewhat surprising to fail...)

cupy/linalg/_einsum.py
ev-br: Indeed, `s/always/sometimes`:

```
In [5]: np.einsum('i->', np.arange(8, dtype=np.float32)).dtype
Out[5]: dtype('float32')

In [6]: np.einsum(', i->', 3, np.arange(8, dtype=np.float32)).dtype
Out[6]: dtype('float64')

In [7]: np.__version__
Out[7]: '2.0.0rc2'
```

numpy bug?

cupy/_core/_kernel.pyx
ev-br: Not for me to decide either. This PR so far labored under assumption that cupy 14 only supports numpy 2.0, per https://github.com/cupy/cupy/issues/8306#issuecomment-2076426835.
BTW, this implies that cupyx.scipy requires scipy 1.13

cupy/_core/_kernel.pyx
ev-br: Not sure what you mean, mind expanding? This code is meant to check `iinfo.min <= arg <= iinfo.max` where arg can be signed or unsigned. Cython is trying to be helpful, so need explicit casts to avoid overflows.

seberg: What I mean is, write this whole thing as:
```
for out_type, arg, weak for zip(out_types, in_args, weaks):
    if not weak is not int:  # assume we change what is stored in `weaks`
        continue

    out_type(int(in_arg))  # check if Python integer fits
```

seberg: now that I look again, I think it also needs to loop `in_types` and not rely on `out_type` being identical.

cupy/_core/_kernel.pyx
ev-br: Currently, `in_types` entries are numpy scalars types already, after the call to https://github.com/cupy/cupy/blob/622cdf4ae126d8bf611eeb0e0dd77a5e069b5cd5/cupy/_core/_kernel.pyx#L180 at https://github.com/cupy/cupy/blob/622cdf4ae126d8bf611eeb0e0dd77a5e069b5cd5/cupy/_core/_kernel.pyx#L1313

Which is the main reason to have an additional `weaks` tuple to be able to distinguish `int8(1) + 1` from `int8(1) + int8(1)`.

So your suggestion is to keep python types to be python types all the way instead of converting them to numpy types early on?

cupy/linalg/_einsum.py
ev-br: So this patch correctly avoids using weak promotion in einsum.

tests/cupy_tests/random_tests/test_sample.py
ev-br: Indeed. Have to xfail for numpy 1.26 + remove short ufunc loops here as well:
https://github.com/cupy/cupy/pull/8323/commits/7217c8dc2ff9269c4aeffaa31945ef8c13b8705c

ev-br:
Progress: `@fuse` reuses the ufunc machinery for type promotion. `core_tests/fusion_tests` all pass, at least locally.

ev-br:
Status:

-  locally,``$ for f in `ls tests/cupy_tests; do echo $f && pytest tests/cupy_tests/$f -n 8 -m'not slow' --tb=no; done`` all passes modulo two OOMs in `test_numpy_interop`. OOMs are likely irrelevant for the promotion state changes;
- @seberg's comments are all addressed --- mostly by Sebastian himself, his commits are included into the PR;
- there are several skips/xfails for things which are broken in numpy 1.26.x + weak promotion and are fixed in numpy 2.0.0rc2;
- there might be more issues in the `cupyx` namespace, but then it also depends on e.g. the scipy version;
- there's a question on how to enable this, also in relation to gh-8314:
     - do we want to push this PR to be CI green (then how to add `export NPY_PROMOTION_STATE=weak` on CI?) , then work on NumPy 2.0 API changes separately, or
     - can we tolerate some CI failures while this is in flux, or
     - do we lump these two PRs together.

cupy/_core/fusion.pyx
```diff
@@ -597,34 +558,40 @@ class _FusionHistory(object):
             raise ValueError('non-broadcastable output operand')
 
         # Typecast and add an operation
-        can_cast = can_cast1 if _should_use_min_scalar(var_list) else can_cast2
-        # TODO(asi1024): Fix to use ``guess_routine``.
-        for op in ufunc._ops.ops:
+        in_sctypes = tuple([a.dtype.type for a in in_vars])
+        weaks = tuple([a.is_weak if hasattr(a, 'is_weak') else False
+                       for a in in_vars])
```
seberg, line 563:
```suggestion
        weaks = tuple([getattr(a, 'is_weak', False) for a in in_vars])
```

tests/cupyx_tests/jit_tests/test_cooperative_groups.py
```diff
@@ -78,12 +78,15 @@ def test_grid(x):
...
 
         x = cupy.empty((16,), dtype=cupy.uint64)
-        x[:] = -1  # = 2**64-1
+        x[:] = -1
         test_grid[2, 32](x)
         assert x[0] == 1
         assert x[1] == 64
         assert (x[2], x[3], x[4]) == (2, 1, 1)
-        assert (x[5:] == 2**64-1).all()
+        # TODO (ev-br): revert back to x[5:] == -1 after the edge case of
```
seberg, line 86:
```suggestion
        # XXX: np2.0: revert back to x[5:] == -1 after the edge case of
```
Let's make sure this is grep'able

seberg (commented):
From my perspective, I think the logic here should be big enough step forward that it would be great to push it forward/merge soon.

The main issue (as @ev-br said) seems now to ensure that tests pass for developers and CI.  What we need for that may be:
* Force testing with NumPy 1.26 and `NPY_PROMOTION_STATE="weak"`.  We can do that in `conftest.py`.
* Fixing CI!  Unfortunately, I am not sure what is all needed for that.  I would think that could be split out in a separate PR.

Otherwise, I think it should be at most be small fixups (a few new xfails, and maybe making sure all xfails/odd or temporary paths are easy to grep for).

There are some more complicated issues, however, I think it is probably better to defer them:
* Many current xfails actually work in NumPy 2 (should we use a skipif?)
* NumPy 2 has some special logic to e.g. make `uint_arr > -1` work.
  (I am not sure how vital this is, but to fix it we will have to add a distinct layer of logic and that seems better in a new PR with an issue.)

---

The goal here is currently to create a CuPy v14 that matches NumPy 2, but fails to match NumPy 1.x w.r.t. to promotion.
That means users will have to stay with CuPy v13 if they have NumPy 1.x (and need it to match).
I think this aligns with the current plans of the CuPy team, but I am not sure.

I will note that I am sure we could implement *both* versions side-by-side without too much hassle and I would be happy to do this if it seems helpful/desirable (even if it is just helpful with getting CI going).

install/cupy_builder/_features.py
```diff
@@ -422,7 +422,7 @@ def get_features(ctx: Context) -> Dict[str, Feature]:
         features = [
             CUDA_cuda(ctx),
             _from_dict(CUDA_cusolver, ctx),
-            _from_dict(CUDA_cudnn, ctx),
+            # _from_dict(CUDA_cudnn, ctx),
```
ev-br, line 425: Note this is needed for me locally (my system has an incompatible cudnn version), should revert the corresponding commit before merging if this goes in.

ev-br:
With a slew of fixups/skips/xfails in the cupyx namespace, this is ready from my side.

jakirkham:
It might be worthwhile to merge in `main`. A few fixes for NumPy 2 went in recently

ev-br:
Rebased on main.

ev-br:
Updated the PR to add a fix in fusion.pyx from @seberg to hopefully fix some of the CI failures.
The majority of other CI "failures" are in fact XPASSes in cupyx.scipy.sparse on scipy 1.13 --- there has been a bit of activity in scipy.sparse recently, and supporting both scipy 1.13 and scipy <= 1.12 gets a little annoying, esp if the test suite is in the strict mode. 

On that note, it would be actually great to have some feedback as to
- whether the general direction of this PR aligns with the project?
- what is the target range of supported numpy/scipy versions. Some CI runs use fairly ancient numpies, also scipy 1.10 etc

asi1024:
@ev-br 
Thanks for your investigation!

> whether the general direction of this PR aligns with the project?

The general direction of this PR looks fine, but there's a concern about potential degradation of the host side performance.
- _kernel.pyx: Performance is a significant concern in this situation. Could you make a evaluation by comparing the execution time with the small elementwise kernel call?
- fusion: Since trace only runs on the initial run, performance isn't a matter of concern in this scenario.

> what is the target range of supported numpy/scipy versions. Some CI runs use fairly ancient numpies, also scipy 1.10 etc

We are planning to drop the support of SciPy<=1.12 in CuPy v14.

ev-br:
Here's a small very much not scientific benchmark @asi1024 : just the docstring example of https://docs.cupy.dev/en/stable/reference/generated/cupyx.profiler.benchmark.html, with either two arrays or an array and a scalar. It seems that the performance differences between main and this PR are within errorbars:


```
==== on main:

In [18]: import cupy as cp
    ...: from cupyx.profiler import benchmark
    ...: 
    ...: def f(a, b):
    ...:     return 3 * cp.sin(-a) * b
    ...: 
    ...: for n in [10, 100, 1000, 1000]:
    ...:    a = 0.5 - cp.random.random((n,))
    ...:    b = 42 #cp.random.random((n,))
    ...:    print(benchmark(f, (a, b), n_repeat=1000))
    ...: 
f                   :    CPU:    49.006 us   +/-  1.191 (min:    47.510 / max:    57.828) us     GPU-0:    52.712 us   +/-  1.366 (min:    50.944 / max:    61.760) us
f                   :    CPU:    49.129 us   +/-  1.011 (min:    47.620 / max:    55.701) us     GPU-0:    52.846 us   +/-  1.505 (min:    50.176 / max:    81.824) us
f                   :    CPU:    49.977 us   +/-  1.001 (min:    48.501 / max:    64.172) us     GPU-0:    53.684 us   +/-  1.160 (min:    51.296 / max:    69.312) us
f                   :    CPU:    49.644 us   +/-  1.037 (min:    48.250 / max:    63.950) us     GPU-0:    53.324 us   +/-  1.151 (min:    51.200 / max:    67.680) us

In [19]: import cupy as cp
    ...: from cupyx.profiler import benchmark
    ...: 
    ...: def f(a, b):
    ...:     return 3 * cp.sin(-a) * b
    ...: 
    ...: for n in [10, 100, 1000, 1000]:
    ...:    a = 0.5 - cp.random.random((n,))
    ...:    b = cp.random.random((n,))
    ...:    print(benchmark(f, (a, b), n_repeat=1000))
    ...: 
f                   :    CPU:    58.835 us   +/- 16.967 (min:    35.310 / max:   109.483) us     GPU-0:    63.930 us   +/- 18.142 (min:    37.792 / max:   117.920) us
f                   :    CPU:    37.784 us   +/-  1.030 (min:    36.580 / max:    49.932) us     GPU-0:    41.473 us   +/-  1.157 (min:    39.424 / max:    53.600) us
f                   :    CPU:    38.974 us   +/-  1.241 (min:    37.283 / max:    52.651) us     GPU-0:    42.646 us   +/-  1.407 (min:    39.808 / max:    57.248) us
f                   :    CPU:    38.408 us   +/-  1.047 (min:    36.532 / max:    47.401) us     GPU-0:    42.245 us   +/-  1.186 (min:    39.872 / max:    51.136) us


on branch:

In [1]: import cupy as cp
   ...: from cupyx.profiler import benchmark
   ...: 
   ...: def f(a, b):
   ...:     return 3 * cp.sin(-a) * b
   ...: 
   ...: for n in [10, 100, 1000, 1000]:
   ...:    a = 0.5 - cp.random.random((n,))
   ...:    b = 42 #cp.random.random((n,))
   ...:    print(benchmark(f, (a, b), n_repeat=1000))
   ...: 
f                   :    CPU:    48.949 us   +/-  0.904 (min:    47.570 / max:    56.149) us     GPU-0:    52.535 us   +/-  1.100 (min:    50.080 / max:    59.552) us
f                   :    CPU:    48.677 us   +/-  1.390 (min:    46.792 / max:    78.181) us     GPU-0:    52.313 us   +/-  1.509 (min:    49.440 / max:    81.920) us
f                   :    CPU:    49.150 us   +/-  1.051 (min:    47.840 / max:    64.112) us     GPU-0:    52.745 us   +/-  1.204 (min:    51.168 / max:    67.904) us
f                   :    CPU:    49.279 us   +/-  0.853 (min:    48.069 / max:    62.831) us     GPU-0:    52.903 us   +/-  1.017 (min:    51.200 / max:    67.552) us

In [2]: import cupy as cp
   ...: from cupyx.profiler import benchmark
   ...: 
   ...: def f(a, b):
   ...:     return 3 * cp.sin(-a) * b
   ...: 
   ...: for n in [10, 100, 1000, 1000]:
   ...:    a = 0.5 - cp.random.random((n,))
   ...:    b = cp.random.random((n,))
   ...:    print(benchmark(f, (a, b), n_repeat=1000))
   ...: 
f                   :    CPU:    66.001 us   +/- 15.602 (min:    38.402 / max:   115.901) us     GPU-0:    71.266 us   +/- 16.489 (min:    41.440 / max:   122.464) us
f                   :    CPU:    40.369 us   +/-  0.802 (min:    39.240 / max:    53.020) us     GPU-0:    44.104 us   +/-  1.001 (min:    42.656 / max:    56.800) us
f                   :    CPU:    39.809 us   +/-  0.745 (min:    38.641 / max:    46.888) us     GPU-0:    43.662 us   +/-  0.935 (min:    41.728 / max:    51.200) us
f                   :    CPU:    40.016 us   +/-  0.899 (min:    38.831 / max:    55.661) us     GPU-0:    43.793 us   +/-  1.090 (min:    41.696 / max:    59.680) us
```

ev-br:
Thanks for the review @asi1024 ! I believe I addressed all your comments + added several small fixes for tests which were failing on CI, and a skip for what appears to be a numpy 1.26 bug, fixed in 2.0.
Remaining CI failures are all on numpy < 1.24 which does not have `_set_promotion_state`. So maybe it's about time to bump not only the minimum scipy  version (https://github.com/cupy/cupy/pull/8323#issuecomment-2270502040), but also a minimum numpy version to 1.25 or thereabouts.

tests/cupy_tests/core_tests/test_elementwise.py
```diff
@@ -92,14 +92,14 @@ class TestElementwiseType(unittest.TestCase):
     @testing.numpy_cupy_array_equal()
     def test_large_int_upper_1(self, xp, dtype):
         a = xp.array([0], dtype=xp.int8)
-        b = xp.iinfo(dtype).max
+        b = xp.asarray(xp.iinfo(dtype).max)
```
asi1024, line 95: Sorry, I overlooked this during my previous review.
These tests are for the type promotion of ndarray and Python primitive scalar. could you please fix it as follows to check if numpy and cupy throw the same exception?

```py
    @testing.for_int_dtypes(no_bool=True)
    @testing.numpy_cupy_array_equal(accept_error=OverflowError)
    def test_large_int_upper_1(self, xp, dtype):
        a = xp.array([0], dtype=xp.int8)
        b = xp.iinfo(dtype).max
        return a + b
```

asi1024:
We are planning to merge a pull request that drops support for NumPy 1.22 and 1.23 in few days. https://github.com/cupy/cupy/pull/8491

tests/cupy_tests/core_tests/test_elementwise.py
ev-br: OK, done.

asi1024: Could you fix also test_large_int_upper_{2,3,4}?

ev-br: OK, fixed test_large_int_upper_2 and test_large_int_lower_{1,2}. 
_{3,4} variants are identical here and in main and all pass, at least locally.

asi1024:
@ev-br In numpy==1.24, the TestFusionScalar::test_numpy_scalar_l test case causes a SEGV during fusion trace due to a bug in NumPy.

Reproducer:

```py
import numpy as np
import cupy as cp

np._set_promotion_state("weak")

@cp.fuse
def func(array):
    np_scalar = np.complex128(1)
    return np_scalar + array

array = cp.arange(12, dtype=np.complex128).reshape(3, 4)
print(func(array))
```

asi1024:
Could you add `@testing.with_requires('numpy>=1.25')` to `test_numpy_scalar_l`?

tests/cupy_tests/linalg_tests/test_product.py
```diff
@@ -468,9 +469,12 @@ def test_matrix_power_large(self, xp, dtype):
         a = xp.eye(23, k=17, dtype=dtype) + xp.eye(23, k=-6, dtype=dtype)
         return xp.linalg.matrix_power(a, 123456789123456789)
 
+    @pytest.mark.skipif(sys.platform == "nt",
```
asi1024, line 472:
```suggestion
    @pytest.mark.skipif(sys.platform == "win32",
```

ev-br:
Done @asi1024

asi1024:
These errors seem to be dtype mismatches that occur only with NumPy 1.24. It appears acceptable to skip them if `numpy<1.25`.

```
=========================== short test summary info ============================
FAILED cupy_tests/functional_tests/test_vectorize.py::TestVectorizeExprs::test_vectorize_ufunc_call_dtype
FAILED cupy_tests/lib_tests/test_polynomial.py::TestPolyvalDtypesCombination::test_polyval_diff_types_array_scalar
= 2 failed, 104003 passed, 6627 skipped, 1503 deselected, 718 xfailed, 2559 warnings in 3773.62s (1:02:53) =
```

ev-br:
thanks @asi1024 , the last commit adds these skips.

asi1024:
@ev-br Pushed two commits that add skips and relax tolerances.

asi1024 (approved):
@ev-br LGTM! Thank you for your contribution!

ev-br:
Thank you @asi1024 !

PR #8341, opened 2024-05-28
Bump version in main branch for v14 development

### LINKED ISSUE #8340: The version number on main seems to be out of date
It looks like on `main` the version number is still `13.0.0rc1`.

https://github.com/cupy/cupy/blob/22694a14896b691c6150c08d0d3f6c6f8c7c4039/cupy/_version.py#L1

In https://github.com/scipy/scipy/pull/20701, I've added tests to SciPy which check that our special function implementations will compile in CuPy. Since SciPy's test suite is typically run in environments where CuPy is not present, we use a function `check_version` to only run these particular tests when CuPy is installed. I set the minimum version to `13.0.0` and was surprised to see that the tests weren't picked up when I had the development version of CuPy installed, even though the latest release was `13.1.0`.

kmaehashi:
Closes #8340.

`a0` is valid as per [PEP 440](https://peps.python.org/pep-0440/):

> The pre-release segment consists of an alphabetical identifier for the pre-release phase, along with a non-negative integer value.


takagi (approved):
LGTM!

PR #8363, opened 2024-06-07
Delaunay: Prevent checking neighboring triangles if nTri <= 1

andfoy:
See https://github.com/cupy/cupy/pull/8290#issuecomment-2146687693

This PR ensures that `kerFindClosestTri` doesn't segfault when the number of triangles in the grid is one.


asi1024 (approved):

leofang:
Q: Is this a fix that should be backported to v13? Does v13 already have Delaunay support (I didn't check)? @andfoy

takagi:
Delaunay is introduced since v14, so we don't need to make this backported.

Issue #6757, opened 2022-05-31
Support cuSPARSELt v0.6.0


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read Release Notes ([cuTENSOR](https://docs.nvidia.com/cuda/cutensor/index.html#changelog) / [cuSPARSELt](https://docs.nvidia.com/cuda/cusparselt/release_notes.html) / [cuDNN](https://docs.nvidia.com/deeplearning/cudnn/release-notes/index.html) / [NCCL](https://docs.nvidia.com/deeplearning/sdk/nccl-release-notes/index.html)).
- FlexCI
  - [x] Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`
  - [ ] ~Backport.~
- Library Installer
  - (n/a) Update `cupyx/tools/install_library.py`.
  - (n/a) ~Backport.~
- Documentation
  - [x] Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html)
  - [ ] ~Backport.~
- [x] Implement new features using the version, if any.

leofang:
There is an API breaking change, see the release note.

takagi:
Thanks for sharing!

gdaisukesuzuki:
Excuse me, I'm an outsider. There seems to be no progress on this ISSUE.The following is a dirty patch, but I'll share it with you as I think it will probably work (I checked it with matmul.py placed under example).

```

*** cupy/cupy_backends/cuda/libs/cusparselt.pyx-dist    2022-06-06 22:13:35.118163775 +0900
--- cupy/cupy_backends/cuda/libs/cusparselt.pyx 2022-06-06 22:31:28.290615626 +0900
***************
*** 98,104 ****
          const void* data, size_t ataSize)
      cusparseStatus_t cusparseLtMatmulGetWorkspace(
          const cusparseLtHandle_t* handle,
!         const cusparseLtMatmulAlgSelection_t* algSelection,
          size_t* workspaceSize)
      cusparseStatus_t cusparseLtMatmulPlanInit(
          const cusparseLtHandle_t* handle,
--- 98,104 ----
          const void* data, size_t ataSize)
      cusparseStatus_t cusparseLtMatmulGetWorkspace(
          const cusparseLtHandle_t* handle,
!         const cusparseLtMatmulPlan_t* plan,
          size_t* workspaceSize)
[...]

Issue #8013, opened 2023-12-02
Allow using CUDA libraries installed from pip wheels


FlorinAndrei:
### Description

JAX, PyTorch and TensorFlow all can use CUDA libraries installed as Python modules. For example, if you install JAX with the pip version of CUDA...

```
pip install --upgrade --user "jax[cuda11_pip]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```

...that will also install the relevant CUDA libraries as Python modules:

```
nvidia-cublas-cu11                11.11.3.6
nvidia-cuda-cupti-cu11            11.8.87
nvidia-cuda-nvcc-cu11             11.8.89
nvidia-cuda-nvrtc-cu11            11.8.89
nvidia-cuda-runtime-cu11          11.8.89
nvidia-cudnn-cu11                 8.9.6.50
nvidia-cufft-cu11                 10.9.0.58
nvidia-cusolver-cu11              11.4.1.48
nvidia-cusparse-cu11              11.7.5.86
nvidia-nccl-cu11                  2.19.3
```

Then you can install TensorFlow as well...

```
pip install --upgrade --user tensorflow=2.14.1
```

...and then PyTorch...

```
pip install --upgrade --user torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118
```

...and they all work perfectly without needing a separate CUDA install.

However, if you install `cupy-cuda11x` and try to run it, it claims it cannot find CUDA. It only works if you give it a completely separate installation of CUDA, in `/usr/local` or something.

Having CUDA installed via pip is easier and more convenient, and may save some storage.

### Additional Information

I'm using Ubuntu 22.04 with an RTX 3090.

kmaehashi:
Hi @FlorinAndrei, thanks for feedback! This is definitely what we want. I'll start investigating changes required to support these PyPI distributed CUDA Toolkit.

HelloCard:
Still not working, still relying on NVIDIA CUDA suite instead of Python components.
```
Traceback (most recent call last):
  File "/root/miniconda3/lib/python3.11/site-packages/cupy/__init__.py", line 17, in <module>
    from cupy import _core  # NOQA
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/lib/python3.11/site-packages/cupy/_core/__init__.py", line 3, in <module>
    from cupy._core import core  # NOQA
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #8364, opened 2024-06-08
Update install.py for downloading and installing libcusparselt 0.6.1

gdaisukesuzuki:
Regarding issue #6757, I tried editing install.py to make it compatible with cusparselt. At this point, it is only grammatically correct, and I haven't tested it on all patterns, so I leave it up to you to decide whether to merge it or not.

### DIFF
--- a/cupyx/tools/install_library.py
+++ b/cupyx/tools/install_library.py
@@ -22,6 +22,7 @@ import urllib.request
 
 _cudnn_records = []
 _cutensor_records = []
+_cusparselt_records = []
 _nccl_records = []
 library_records = {}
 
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thanks @gdaisukesuzuki! CuPy is moving to use PyPI-hosted wheels instead of installing shared libraries on our own (#8013, #6999). As for cuSPARSELt, it looks like only a stub package ( https://pypi.org/project/nvidia-cusparselt-cu12) is available. @leofang, are there any plans to have cuSPARSELt wheels on PyPI?

leofang:
We can't upload it to PyPI yet because of the size limit (https://github.com/pypi/support/issues/3904). I can check with @ethanhs next week and see if we can use pypi.nvidia.com as a temporary solution. Will get back once we discuss.

leofang:
Sorry for my late response. I did discuss with the cuSPARSELt team and they have reasons to not publish to pypi.nvidia.com. Fortunately this is moot now that the size-bump request was granted. @ethanhs told me he should be able to get the wheels pushed to PyPI in a few days. Will keep this thread updated.

leofang:
Update: We've released cuSPARSELt v0.6.2 wheels on PyPI (https://pypi.org/project/nvidia-cusparselt-cu12/), would you be able to give it a try @gdaisukesuzuki?

kmaehashi:
As discussed above, we are moving away from our own installer, so let me close this PR. I've made an issue to track the feature: #8604. Thanks again @gdaisukesuzuki for the contribution!

PR #8365, opened 2024-06-08
eigsh doc correction _eigen.py

YanivDorGalron:
option 'SA' is also possible on 'which' argument


kmaehashi:
Thanks @YanivDorGalron! LGTM except for the style check error, we have an 80-char-folding rule.

YanivDorGalron:
Hey @kmaehashi , I think it should work now.

kmaehashi (approved):

PR #8347, opened 2024-06-03
[Do not merge] try rocm 6.1 ci

takagi:
https://github.com/cupy/cupy/pull/7964
https://github.com/cupy/cupy/pull/8316

PR #8366, opened 2024-06-09
Fix pull request project board workflows

kmaehashi:
Use the latest/standard GitHub actions.

* `actions/create-github-app-token@v1` (official equivalent) instead of `tibdex/github-app-token`
* `actions/download-artifact@v4` (now supports downloading artifacts from other workflow runs) instead of our own implementation
* `actions/upload-artifact@v4` instead of `actions/upload-artifact@v3` to avoid deprecation warnings ([example](https://github.com/cupy/cupy/actions/runs/9437786961))

### DIFF
--- a/.github/workflows/pr-project.yml
+++ b/.github/workflows/pr-project.yml
@@ -17,37 +17,19 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Generate Token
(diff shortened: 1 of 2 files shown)


takagi (approved):
LGTM!

kmaehashi:
Tested with #8347, confirmed it works as expected.
https://github.com/cupy/cupy/actions/runs/9559633289/job/26350299781

PR #8368, opened 2024-06-11
Add NumPy 2.0 on document

takagi:
Should we make this at this moment?


kmaehashi (approved):
Thanks, LGTM! Please send a separate PR for `main` branch that updates baseline API (`CuPy v14 is based on NumPy 2.0 and SciPy 1.13`)

PR #8369, opened 2024-06-12
Fix `ndarray.get()` not honoring current stream when layout is not contiguous

leofang:
This appears to be a very old bug that we managed to expose somehow 😄 

When the layout is not contiguous, we call `_internal_ascontiguousarray` or `_internal_asfortranarray`, which calls the ufunc `elementwise_copy`, which then fetches the current stream before launching. However, if the stream is passed as an argument to `.get()`, it's not set as a current stream, thus not used by the ufunc.


asi1024:
LGTM. Thanks for your PR!

PR #8379, opened 2024-06-13
Preserve stream set with .use() when exiting stream context manager

### LINKED ISSUE #8377: Stream context manager does not preserve previous cupy.cuda.Stream.use() call
### Description

We have a code where we set the stream at program startup to a non-default CUDA stream (via the `use` option). We discovered that if somewhere later on in the code a `with` statement with a different stream is called, the stream resets to the CUDA default stream once we leave the scope of the context manager! This is unexpected and unintuitive behavior. 

Some comments in https://github.com/cupy/cupy/blob/main/cupy/cuda/stream.pyx state that mixing `use` and `with` is an "anti-pattern" but I think it is reasonable for a user to `use` a stream to set a new default stream and then use `with` to temporarily change streams during the program duration. I would agree that using a `use` within a `with` block makes much less sense.

### To Reproduce

```py
import cupy as cp

s1 = cp.cuda.Stream()
s2 = cp.cuda.Stream()
print(f"s1 = {s1}")
print(f"s2 = {s2}")

print(cp.cuda.get_current_stream())

# Set current stream to s1
s1.use()
print(cp.cuda.get_current_stream())

# Temporarily set stream to s2
with s2:
  print(cp.cuda.get_current_stream())

# Stream resets to default stream but should reset to s1
print(cp.cuda.get_current_stream())
```

### Installation

### Environment

```
OS                           : Linux-5.4.0-169-generic-x86_64-with-glibc2.35
CuPy Version                 : 13.1.0
NumPy Version                : 1.26.4
SciPy Version                : 1.13.0
CUDA Root                    : /opt/nvidia/hpc_sdk/Linux_x86_64/24.1/compilers
nvcc PATH                    : /opt/nvidia/hpc_sdk/Linux_x86_64/24.1/compilers/bin/nvcc
CUDA Runtime Version         : 12040 (linked to CuPy) / 12030 (locally installed)
cuFFT Version                : 11012
cuRAND Version               : 10304
cuSOLVER Version             : (11, 5, 4)
NVRTC Version                : (12, 3)
NCCL Runtime Version         : 21805
cuTENSOR Version             : 20000
Device 0 Name                : Tesla V100-DGXS-16GB
Device 0 PCI Bus ID          : 0000:07:00.0
Device 1 Name                : Tesla V100-DGXS-16GB
Device 1 Compute Capability  : 70
Device 1 PCI Bus ID          : 0000:08:00.0
Device 2 Name                : Tesla V100-DGXS-16GB
Device 2 Compute Capability  : 70
Device 2 PCI Bus ID          : 0000:0E:00.0
Device 3 Name                : Tesla V100-DGXS-16GB
Device 3 Compute Capability  : 70
Device 3 PCI Bus ID          : 0000:0F:00.0
```

### Additional Information

The problem seems to lie in the handling of `current_stream_stack` in the implementation of `Stream`.  The initial/base value of the stack is always initialized to the CUDA default stream. Only calls using `with` push or pop streams to this stack so once we leave the scope of the `with` statement, the stream gets reset to the base value in the stack which is the CUDA default stream, not the one currently set with `use`.

romerojosh:
Fixes #8377.

This PR fixes the behavior described in #8377, however it may need some additional input. The current version of this change disallows the usage of `.use()` within `with` blocks, causing a couple of tests to fail. 

`test_mix_use_context` fails as I explicitly throw a new `RuntimeError` in this case. This can be adjusted to preserve existing behavior if desired.

`test_with_statement` fails in `ptds` mode because that test expects that the `ptds` stream gets reset to the `null` stream after the `with` blocks are exited. This is behavior I noted as a "bug" in #8377 so this test seems to indicate that this behavior is expected.


romerojosh:
@seberg I went ahead and rebased this PR to the latest main. Running the stream tests locally, I find that this PR introduces 3 failures (same failures as noted in the original PR description), but they are expected:
  - test_with_statement[ptds] still expects .use() to be forgotten after leaving a context.
  - Both test_mix_use_context variants expect .use() inside a stream context to work, but the branch now raises RuntimeError intentionally.

cupy/cuda/stream.pyx
```diff
@@ -53,6 +53,17 @@ cdef class _ThreadLocal:
         self.set_current_stream(prev_stream)
         assert len(self.current_stream_stack[device_id]) >= 1
 
+    cdef void replace_base_stream(self, stream, int device_id) except*:
+        assert device_id >= 0
+        # should not replace base stream if there are any other streams
+        # pushed onto the stack. e.g. .use() should not be called inside
+        # "with" blocks
```
seberg: Thanks, and I should have seen yesterday already.

But I don't think this is a good change, we should just support it (we could consider discouraging it in documentation maybe).

I realize this was also my very first thought, but that was in part because currently the current stream is _also_ `stream_stack[-1]`, which it _shouldn't_ be.

Instead, can you make it so that:
* `push_stream()` fetches the *current* stream, pushes it and then sets the current stream to the new one.
   * ~EDIT: subtlety: If setting the stream fails, we should ideally `pop()` it again directly here (assume `pop()` can't fail) -- or deal with errors in some other way.~ (Sorry, silly me, this isn't a driver call, it can't fail.)
* `pop_stream()` is very strange, what it should be doing is: `prev_stream = self.current_tream_stack[device_id].pop()`.

Once we have that, please add the use of `.use()` inside the context manager to the test.

romerojosh: Ok, I made these changes. I also adjusted a few of the tests impacted by this behavior change to handle the new behavior.

cupy/cuda/stream.pyx
```diff
@@ -41,17 +41,16 @@ cdef class _ThreadLocal:
...
         # record device_id to prevent from popping the wrong stream at exit
         self.current_device_id_stack.append(device_id)
         self.set_current_stream(stream)
 
     cdef void pop_stream(self) except*:
         cdef int device_id = self.current_device_id_stack.pop()
-        self.current_stream_stack[device_id].pop()
-        prev_stream = self.current_stream_stack[device_id][-1]
+        prev_stream = self.current_stream_stack[device_id].pop()
         self.set_current_stream(prev_stream)
```
seberg:
```suggestion
        self.set_current_stream(prev_stream, device_id)
```

seberg (dismissed):
Thanks, LGTM.  Including the old test fixups (which also cover the `use()` inside the block, nice!

Holding off though until tomorrow, I might push a test for that one suggestion.

I think this is a pre-existing bug, and it only kicks in if you do things you shouldn't be doing:
```
with stream_on_dev0:
    Device(1).use()
```
(to the point that we could probably spew out errors and warning when `device_id != current_device)

cupy/cuda/stream.pyx
```diff
@@ -29,7 +29,7 @@ cdef class _ThreadLocal:
         for i in range(num_devices):
             default_stream = get_default_stream()
             self.current_stream.append(default_stream)
-            self.current_stream_stack.append([default_stream])
+            self.current_stream_stack.append([])
```
leofang, line 32: I think we can also drop line 30 now? (The linter probably cannot find unused variables...)

cupy/cuda/stream.pyx
leofang: @seberg do you mean we should extend this method to accept two args? Right now it only accepts one.

seberg: Ah, I hadn't noticed that part :). But yes: `self.set_current_stream()` can't reliably infer the correct device here *if the code does things it shouldn't* (it also is just unnecessary):
```
stream  # stream with `.device == -1` for "any device".
with dev0:
    with stream:  # __enter__ sets stream on dev0
        dev1.use()  # change current device
  # __exit__ pops stream from dev1
```

Might go as far as requiring the input to `set_current_stream()` and rejecting `-1` for clarity (although, I guess this may be the only place it matters, unless we'd add a `device.set_current_stream(stream)`.).

cupy/cuda/stream.pyx
seberg: No, the line above it still uses it and it is necessary because that keeps the stream alive.

Let me do these two fixes and then follow-up to try and disentangle this a bit more (dunno, maybe merge the backend-stream and stream TLS).

It seems weird that we have two TLS classes one hold the stream pointer and one for the stream object.

seberg:
I pushed the fix I mentioned, I think it is obvious enough.  I'll defer further clean-ups to possible follow-ups (there are definitely some).

@leofang: marked this as "no-compat" as when this was changed in v10, it was actually documented in `upgrade.rst`.  My feeling is a bit that it is a rather ugly bug we should just fix but mention.
That is found on my believe that there is practically no way to rely on the old, bad, behavior (I.e. the number of affected users should be close to zero).

/test mini

seberg (approved):

cupy/cuda/stream.pyx
leofang: Ah, ok, missed it when browsing on my phone, thx.

leofang:
> @leofang: marked this as "no-compat" as when this was changed in v10, it was actually documented in `upgrade.rst`. My feeling is a bit that it is a rather ugly bug we should just fix but mention.
> That is found on my believe that there is practically no way to rely on the old, bad, behavior (I.e. the number of affected users should be close to zero).

So do you mean that we should still ship this in v14.x and just release-note it as a major bug we must fix?

seberg:
That is my small tendency right now.  But v15 isn't far off and maybe I am stretching things too far (at least until we change the BC docs a bit).

That tendency relies on guessing that it is really hard to unintentionally rely on the current behavior for correctness. And at the same time considering it a rather bad bug.

The `no-compat`: I guess I think it may make sense to document this at least a bit.  Even if just by me following up with a `.. note::` in the `Stream` docs.

leofang:
Yeah not sure this is strictly a no-compact, but it's definitely a bug fix.

PR #8386, opened 2024-06-19
Remove hard coded yshape in the ndimage interpolation cuda kernels

martinResearch:


martinResearch:
it seems that converting the tuple yshape to a cupy array that can be provided as an argument has a non-negligeable runtime cost, is that expected? Is there a faster solution?

takagi:
> it seems that converting the tuple yshape to a cupy array that can be provided as an argument has a non-negligeable runtime cost, is that expected? Is there a faster solution?

We have to transfer the data to gpu at calling-time.

takagi (dismissed):
LGTM!

leofang:
cc: @grlee77 for vis

takagi:
Seems that CI fails with the following message:
```
E       TypeError: _generate_interp_custom() missing 1 required positional argument: 'yshape'
```
Would you remove `yshape` argument here?
https://github.com/cupy/cupy/blob/64c3b26df94907716f8f24f72dcec9676c75533a/cupyx/scipy/ndimage/_interp_kernels.py#L226

martinResearch:
"is the primary goal of this to avoid having to compile/cache a new kernel for each shape?" Yes that was the goal. We may want to benchmark and document the overhead added by having the send the shape each time we call the kernel.

martinResearch:
@takagi the tests are passing. Any other things that I would need to do before it can be approved?

takagi (dismissed):
LGTM!

martinResearch:
@takagi  am struggling to compile cupy locally so I cannot run the tests on my machine, and I do not know how to trigger the tests on pfn in this PR. I hope my last commit fixes the problems, but I cannot check that.

takagi:
Thanks for your patient effort. I have checked that all of `TestAffineTransform::test_affine_transform_float` tests pass on my local, and the CI should pass this time. Let me kick it again.

takagi (approved):

takagi:
Seems we need to pass y's shape to the kernel we're constructing.
https://ci.preferred.jp/cupy.linux.cuda126/173785/#L254038

> am struggling to compile cupy locally so I cannot run the tests on my machine, and I do not know how to trigger the tests on pfn in this PR.

Only CuPy maintainers can trigger the CI. Would you share some log you're seeing on building cupy on your side so I can help you run the test for this PR?

kmaehashi:
Could you takeover the assignee, @asi1024?

Issue #8384, opened 2024-06-19
Finding the cuda folder by searching for the nvrtc dll


martinResearch:
### Description

CuPy has a mechanism implemented [here](https://github.com/cupy/cupy/blob/0fef5f6b3751da044895dea47a84da2a7e1c165a/cupy/_environment.py#L114C5-L114C19) to find the folder that contain the CUDA binaries when CUDA_PATH is not set by searching for the `nvcc` executable. 
In practice it seems possible to use CuPy without the `nvcc` executable by using the `nvrtc` dll as backend instead (see [here](https://github.com/cupy/cupy/blob/0fef5f6b3751da044895dea47a84da2a7e1c165a/cupy/cuda/compiler.py#L277)). While doing so one currently still needs to either set CUDA_PATH  or have the `nvcc` executable installed in their environement for CuPy to be able to find the cuda path.
We could relax this constraint by searching for the folder containing the nvrtc dll too using ctypes.

kmaehashi:
Thanks for the feedback and PR @martinResearch! Could you elaborate more on the motivation behind this?

martinResearch:
I am using a conda environment in which the `nvrtc` dlls is installed trough a conda package but not `nvcc`. I cannot set the CUDA_PATH variable in advance in the conda `environment.yml` file because I do not know in advance where this environment is going to be created by the contributors to my project. I would prefer not to ask the contributeurs to my project to have to manually add a CUDA_PATH variable in the conda environment or at the system level because that is error-prone. Without the change in this PR CuPy is not able the find the cuda folder. With the changes in this PR, CuPy finds the nvrtc dll and its path is used to define the cuda path and I do not need to manually define the CUDA_PATH variable.

leofang:
If you also install CuPy from conda-forge, library loading should already be taken care for you. conda knows where to find shared libraries. I would like to know better what is missing for your use case. Can you share with us, e.g. what is the error that you see without your PR?

martinResearch:
Although I use a conda environment, I install my dependencies from pypi.org using pip because I use pip-tools to pin my dependencies versions.
The error I am getting is `nvrtc: error: failed to open nvrtc-builtins64_118.dll.`

leofang:
> Although I use a conda environment, I install my dependencies from pypi.org using pip ...

This is the kind of information that would have helped us understand the issue sooner if you had filled out the issue template as requested 🙂 

[...]

PR #8387, opened 2024-06-19
use nvrtc dll to find cuda path

martinResearch:


cupy/_environment.py
```diff
@@ -122,12 +129,25 @@ def _get_cuda_path():
     if nvcc_path is not None:
         return os.path.dirname(os.path.dirname(nvcc_path))
 
+    # Use nvrtc dll path
+    nvrtc_path = get_nvrtc_path()
+    if nvrtc_path is not None:
+        return os.path.dirname(os.path.dirname(nvrtc_path))
```
leofang, line 135: This line does not work if the goal is to support NVRTC wheel (nvidia-cuda-nvrtc-cuXX), because the DLL is installed to a subdir under site-packages: On Linux it's nvidia/cuda_nvrtc/lib. So you need 3 dirname calls not 2. But even if you find the root dir nvidia this way it still would not work for CuPy to find other CUDA components installed via pip.

xref: https://github.com/cupy/cupy/issues/8384#issuecomment-2187833916

leofang (changes requested):

kmaehashi:
Closing as inactive for a while. As Leo suggested, the fix is insufficient to support CUDA wheels on PyPI.

PR #8388, opened 2024-06-20
Fix spline temp container size in make_interp_spline

andfoy:
See https://github.com/cupy/cupy/pull/8290#issuecomment-2175220453

This prevents the `d_boor` kernel from performing an out-of-bounds array access, this was caused due to the size of `temp` in `make_interp_spline`, which was shorter than the expected value. After this PR, the confirmed segfault caused when running the following test no longer appears:

```
compute-sanitizer --tool memcheck --padding 32 python -m pytest -v  tests/cupyx_tests/scipy_tests/interpolate_tests/test_bspline2.py  -k test_minimum_points_and_deriv
```

imagen


takagi (approved):
LGTM!

takagi:
CI failures are not related. Thanks, @andfoy!

PR #8407, opened 2024-07-01
Fix `RandomState.seed()` for NumPy 2 compatibility

### LINKED ISSUE #8392: cupy.random.seed return conversion error
### Description

With the following code: 
```
cp.random.seed(1234)
```

The following error is returned:
`TypeError: Cannot cast scalar from dtype('int64') to dtype('uint64') according to the rule 'safe'`

### To Reproduce

```py
cp.random.seed(1234)
```

### Installation

### Environment

```
S                           : Linux-5.15.0-84-generic-x86_64-with-glibc2.35
CuPy Version                 : 13.2.0
NumPy Version                : 2.0.0
SciPy Version                : 1.14.0
CUDA Driver Version          : 12050
CUDA Runtime Version         : 12040 (linked to CuPy) / 12050 (locally installed)
cuFFT Version                : 11203
cuRAND Version               : 10306
cuSOLVER Version             : (11, 6, 2)
NVRTC Version                : (12, 5)
NCCL Runtime Version         : 22105
Device 0 Name                : NVIDIA A30
Device 0 Compute Capability  : 80
Device 0 PCI Bus ID          : 0000:41:00.0

```

### Additional Information

_No response_

takagi:
Close #8392.

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -809,7 +809,12 @@ class RandomState(object):
             if isinstance(seed, numpy.ndarray):
                 seed = int(hashlib.md5(seed).hexdigest()[:16], 16)
             else:
-                seed = int(numpy.asarray(seed, dtype=numpy.uint64))
+                seed_passed = numpy.asarray(seed)
+                # Cast allow integer overflows:
+                seed = seed_passed.astype(numpy.uint64, casting='same_kind')
+                if (seed != seed_passed).any():
+                    raise ValueError('Seed must be between 0 and 2**64 - 1')
+                seed = int(seed)
 
         curand.setPseudoRandomGeneratorSeed(self._generator, seed)
         if (self.method not in (curand.CURAND_RNG_PSEUDO_MT19937,
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -809,8 +809,7 @@ class RandomState(object):
             if isinstance(seed, numpy.ndarray):
                 seed = int(hashlib.md5(seed).hexdigest()[:16], 16)
             else:
-                seed = int(
-                    numpy.asarray(seed).astype(numpy.uint64, casting='safe'))
+                seed = int(numpy.asarray(seed, dtype=numpy.uint64))
 
         curand.setPseudoRandomGeneratorSeed(self._generator, seed)
         if (self.method not in (curand.CURAND_RNG_PSEUDO_MT19937,


cupy/random/_generator.py
```diff
@@ -809,8 +809,7 @@ def seed(self, seed=None):
             if isinstance(seed, numpy.ndarray):
                 seed = int(hashlib.md5(seed).hexdigest()[:16], 16)
             else:
-                seed = int(
-                    numpy.asarray(seed).astype(numpy.uint64, casting='safe'))
+                seed = int(numpy.asarray(seed, dtype=numpy.uint64))
```
kmaehashi, line 812: This changes the behavior; the new code implicitly casts `cupy.random.seed(1.23)` to uint64, but it should raise an error, like numpy does.

seberg: The NumPy code actually checks ranges manually, an alternative would be to do:
```
seed_passed = numpy.asarray(seed)
# Cast allow integer overflows:
seed = seed_passed.astype(numpy.uint64, casting='same_kind')
if (seed != seed_passed).any():
    raise ValueError("Seed must be between 0 and 2**64 - 1")
```
(NumPy actually requires 32 bit seeds, but maybe no point in changing behavior here).

seberg: Really, sorry... the `same_kind` casting doesn't work, maybe this:
```
                seed_arr = numpy.asarray(seed)
                if seed_arr.dtype.kind not in "iu":
                    raise TypeError("Seed must be an integer.")
                seed = int(seed_arr)
                # Check that no integer overflow occurred during the cast
                if seed < 0 or seed >=2**64:
                    raise ValueError('Seed must be an integer between 0 and 2**64 - 1')
```

I would suggest just replacing the whole thing with `seed = operator.index()` and the range check, but I am not 100% if that might create some other regression.

seberg:
Would it be a lot of work to push this (and the other copyto one) through and create a 13.2.1 release?
It was unfortunate that we didn't notice these earlier, but unfortunately they do block downstream projects tests to succeed with NumPy 2.

seberg:
To note: unfortunately, my first suggestion was incorrect.  It only works on NumPy 1.x due to the value-based casting safety check, but unfortunately fails on 2.0.

If it helps, I can create a PR with the new suggestion.

takagi:
> If it helps, I can create a PR with the new suggestion.

Can I ask you to create the new PR? Thanks!

Issue #8393, opened 2024-06-27
`AttributeError` on fft code


cliffburdick:
### Description

With the following code:
```
        amf_2d = nfreq * cp.abs(cp.fft.fftshift(
            cp.fft.ifft(new_ynorm, nfreq, axis=1), axes=1))
```

An error is reported:
```
AttributeError: `np.compat` was removed in the NumPy 2.0 release. There's no replacement, as Python 2 is no longer supported.
```

### To Reproduce

```py
        amf_2d = nfreq * cp.abs(cp.fft.fftshift(
            cp.fft.ifft(new_ynorm, nfreq, axis=1), axes=1))
```

`new_ynorm` is a 2D complex array, and `nfreq`  is a scalar

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
OS                           : Linux-5.15.0-84-generic-x86_64-with-glibc2.35
Python Version               : 3.10.12
CuPy Version                 : 13.2.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.0.0
SciPy Version                : 1.14.0
Cython Build Version         : 0.29.36
Cython Runtime Version       : None
CUDA Root                    : /usr/local/cuda
nvcc PATH                    : /usr/local/cuda/bin/nvcc
CUDA Build Version           : 12040
CUDA Driver Version          : 12050
CUDA Runtime Version         : 12040 (linked to CuPy) / 12050 (locally installed)
cuBLAS Version               : (available)
cuFFT Version                : 11203
cuRAND Version               : 10306
cuSOLVER Version             : (11, 6, 2)
cuSPARSE Version             : (available)
NVRTC Version                : (12, 5)
Thrust Version               : 200200
CUB Build Version            : 200200
Jitify Build Version         : <unknown>
cuDNN Build Version          : (not loaded; try `import cupy.cuda.cudnn` first)
cuDNN Version                : (not loaded; try `import cupy.cuda.cudnn` first)
NCCL Build Version           : 21602
NCCL Runtime Version         : 22105
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA A30
Device 0 Compute Capability  : 80
Device 0 PCI Bus ID          : 0000:41:00.0

```

lucascolley:
Was this closed by gh-8430?

cliffburdick:
[...]

PR #8408, opened 2024-07-01
Fix `copyto` for NumPy 2 compatibility

### LINKED ISSUE #8391: `cp.full` with `np.nan` broken
### Description

It seems like a `numpy` 2 thing, but using `cp.full` with `fill_value=np.nan` is broken, and previously worked.

### To Reproduce

```py
import numpy as np, cupy as cp

out = cp.full((2, 2), np.nan)
```

### Installation

### Environment

```
OS                           : Linux-4.18.0-513.11.1.el8_9.x86_64-x86_64-with-glibc2.28
Python Version               : 3.11.7
CuPy Version                 : 13.2.0
NumPy Version                : 2.0.0
CUDA Driver Version          : 12030
CUDA Runtime Version         : 12040 (linked to CuPy) / 12030 (locally installed)
cuFFT Version                : 11012
cuRAND Version               : 10304
cuSOLVER Version             : (11, 5, 4)
NVRTC Version                : (12, 3)
cuDNN Build Version          : 8801
cuDNN Version                : 8500
NCCL Build Version           : (not loaded; try `import cupy.cuda.nccl` first)
NCCL Runtime Version         : (not loaded; try `import cupy.cuda.nccl` first)
Device 0 Name                : Tesla V100-SXM3-32GB
Device 0 PCI Bus ID          : 0000:C0:00.0
```

### Additional Information

Version info:
```
asttokens==2.4.1
cupy-cuda12x==13.2.0
decorator==5.1.1
executing==2.0.1
fastrlock==0.8.2
ipython==8.25.0
jedi==0.19.1
matplotlib-inline==0.1.7
numpy==2.0.0
parso==0.8.4
pexpect==4.9.0
prompt-toolkit==3.0.47
ptyprocess==0.7.0
pure-eval==0.2.2
pygments==2.18.0
six==1.16.0
stack-data==0.6.3
traitlets==5.14.3
typing-extensions==4.12.2
wcwidth==0.2.13
```
and this works with:
```
numpy==1.26.4
```

takagi:
Close #8391

### DIFF
--- a/cupy/_manipulation/basic.py
+++ b/cupy/_manipulation/basic.py
@@ -35,7 +35,7 @@ def copyto(dst, src, casting='same_kind', where=None):
         fusion._FusionVarScalar, _fusion_interface._ScalarProxy)
     if src_is_python_scalar:
         src_dtype = numpy.dtype(type(src))
-        can_cast = numpy.can_cast(src, dst.dtype, casting)
+        can_cast = numpy.can_cast(src_dtype, dst.dtype, casting)
     elif isinstance(src, numpy.ndarray) or numpy.isscalar(src):
         if src.size != 1:
             raise ValueError(
--- a/cupy/_manipulation/basic.py
+++ b/cupy/_manipulation/basic.py
@@ -35,7 +35,11 @@ def copyto(dst, src, casting='same_kind', where=None):
         fusion._FusionVarScalar, _fusion_interface._ScalarProxy)
     if src_is_python_scalar:
         src_dtype = numpy.dtype(type(src))
-        can_cast = numpy.can_cast(src_dtype, dst.dtype, casting)
+        # When the cast is safe, we check wether the value fits (for now)
+        can_cast = numpy.can_cast(
+            src_dtype, dst.dtype, casting if casting != "safe" else "same_kind")
+        if casting == "safe":
+            can_cast = numpy.array(src, dtype=dst.dtype) == src
     elif isinstance(src, numpy.ndarray) or numpy.isscalar(src):
         if src.size != 1:
             raise ValueError(
--- a/cupy/_manipulation/basic.py
+++ b/cupy/_manipulation/basic.py
@@ -38,7 +38,7 @@ def copyto(dst, src, casting='same_kind', where=None):
         # When the cast is safe, we check whether the value fits (for now).
         # See #8408
         can_cast = numpy.can_cast(
-            src_dtype, dst.dtype,
+            src, dst.dtype,
             casting if casting != "safe" else "same_kind")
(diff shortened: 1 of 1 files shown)


cupy/_manipulation/basic.py
```diff
@@ -35,7 +35,7 @@ def copyto(dst, src, casting='same_kind', where=None):
         fusion._FusionVarScalar, _fusion_interface._ScalarProxy)
     if src_is_python_scalar:
         src_dtype = numpy.dtype(type(src))
-        can_cast = numpy.can_cast(src, dst.dtype, casting)
+        can_cast = numpy.can_cast(src_dtype, dst.dtype, casting)
```
seberg, line 38: Maybe the pragmatic thing to just do this.

The more correct thing would probably be to do:
```
# When the cast is safe, we check whether the value fits (for now)
can_cast = numpy.can_cast(
    src_dtype, dst.dtype,
    casting if casting != "safe" else "same_kind
)
if casting == "safe":
    can_cast = np.array(src, dtype=dst.dtype) == src
```
You could skip the second part, but it will give you a change in behavior for NumPy 1.x and NumPy 2 currently does this anyway.  (There is no ideal solution, I think, so this is fine.)

takagi:
Close #8393

kmaehashi (dismissed):
LGTM!

cupy/_manipulation/basic.py
```diff
@@ -35,7 +35,13 @@ def copyto(dst, src, casting='same_kind', where=None):
         fusion._FusionVarScalar, _fusion_interface._ScalarProxy)
     if src_is_python_scalar:
         src_dtype = numpy.dtype(type(src))
-        can_cast = numpy.can_cast(src, dst.dtype, casting)
+        # When the cast is safe, we check wether the value fits (for now).
```
jakirkham, line 38:
```suggestion
        # When the cast is safe, we check whether the value fits (for now).
```

jakirkham:
Do we need to update some of the tests as well?

It looks like there are cases in `cupy_tests/manipulation_tests/test_basic.py` where it is using two different types in a way that doesn't appear to work with the new casting behavior

jakirkham:
> Close #8393

Think this needs to be in the top-level OP to work

takagi:
> Do we need to update some of the tests as well?

It looks like the tests are fine and I should pass the `src` ndarray to 'can_cast' as the original code did.

seberg:
I'll open a new PR for this in a bit.  I suppose the tests failed with the first version due to `int -> uint` not being same-kind as well.  Unfortunately, the current fix won't side-step the original problems on NumPy 2, I think.

PR #8409, opened 2024-07-01
MAINT: Avoid using np.compat.integer_types

seberg:
This was removed in 2.0, since it is just `int` anyway since Python 2 is not a thing anymore.

A better pattern is probably:
```
        try:
            axes = (operator.index(axes),)
        except TypeError:
            pass
```
(broadening this to non Python integers).  But it doens't matter much and this is the more trivial change.


kmaehashi:
Thanks @seberg!

/test mini

kmaehashi (approved):

kmaehashi:
Test failures are irrelevant.

PR #8410, opened 2024-07-02
Fix type dispatcher for arm64

emcastillo:
In ARM, char is an unsigned int which causes several tests to fail.

https://developer.arm.com/documentation/den0013/latest/Porting/Miscellaneous-C-porting-issues/unsigned-char-and-signed-char


leofang:
FYI this is one of the hardest bugs to hunt down, thanks to Emilio for figuring it out 🙂

kmaehashi (approved):
Thanks, LGTM!

PR #8411, opened 2024-07-02
Skip some tests in aarch64 CI

emcastillo:
Some of the routines here rely in scipy functions that don't return the same results in x86/arm. Seems that the CUDA routines are written to mach the results in x86 causing several tests to fail.

https://github.com/scipy/scipy/issues/20158
https://github.com/scipy/scipy/issues/20160

### DIFF
--- a/cupyx/fallback_mode/fallback.py
+++ b/cupyx/fallback_mode/fallback.py
@@ -321,7 +321,12 @@ class ndarray(object):
                     if self._numpy_array is None:
                         self._numpy_array = base._numpy_array.view(type=_type)
                         self._numpy_array.shape = self._cupy_array.shape
-                        self._numpy_array.strides = self._cupy_array.strides
+                        # In some scenarios, the numpy array strides may be
+                        # different. Copying the strides directly will cause
+                        # cause errors. This has been observed in ARM systems
+                        cupy_strides = self._cupy_array.strides
+                        if self._numpy_array.strides != cupy_strides:
+                            self._numpy_array.strides = cupy_strides
         else:
             # not cupy-compatible
             if base is not None:
--- a/cupyx/fallback_mode/fallback.py
+++ b/cupyx/fallback_mode/fallback.py
@@ -321,12 +321,7 @@ class ndarray(object):
                     if self._numpy_array is None:
                         self._numpy_array = base._numpy_array.view(type=_type)
                         self._numpy_array.shape = self._cupy_array.shape
-                        # In some scenarios, the numpy array strides may be
-                        # different. Copying the strides directly will cause
-                        # cause errors. This has been observed in ARM systems
-                        cupy_strides = self._cupy_array.strides
-                        if self._numpy_array.strides != cupy_strides:
-                            self._numpy_array.strides = cupy_strides
+                        self._numpy_array.strides = self._cupy_array.strides
         else:
             # not cupy-compatible
             if base is not None:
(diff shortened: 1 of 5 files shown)


tests/cupyx_tests/scipy_tests/special_tests/test_polygamma.py
```diff
@@ -42,6 +44,9 @@ def test_scalar(self, xp, scp, dtype):
         return scp.special.polygamma(
             dtype(2.), dtype(1.5)).astype(numpy.float32)
 
+    @pytest.mark.xfail(
+        platform.processor() == "aarch64",
+        reason="aarch64 scipy does not match cupy/x86 see Scipy")
```
jakirkham, line 49: Is this meant to have an issue #?

cupyx/fallback_mode/fallback.py
```diff
@@ -321,7 +321,12 @@ def _update_numpy_array(self):
                     if self._numpy_array is None:
                         self._numpy_array = base._numpy_array.view(type=_type)
                         self._numpy_array.shape = self._cupy_array.shape
-                        self._numpy_array.strides = self._cupy_array.strides
+                        # In some scenarios, the numpy array strides may be
```
kmaehashi, line 324: Can I have an example to reproduce this?

leofang:
Note: This PR (#8411) and #8410 were patches needed when we worked on #8442 and ran the full test suite on G+H. None of these patches are specific to G+H, I believe they were just oversights due to lack of aarch64 CIs.

leofang:
@emcastillo could you take a look at the questions above when you get a chance? 🙂 Thanks!

tests/cupyx_tests/scipy_tests/special_tests/test_polygamma.py
emcastillo: Sorry, I think it was https://github.com/scipy/scipy/issues/20159
Will update it

cupyx/fallback_mode/fallback.py
emcastillo: Actually this error seems to be gone now, probably it was some kind of numpy/cupy incompatibility when I tested it the first time.

emcastillo: Sorry, this is related to #8442, When running in a system with unified memory we get the following error

```FAILED tests/cupyx_tests/fallback_mode_tests/test_fallback.py::TestInplaceSpecialMethods::test_put_along_axis - ValueError: strides is not compatible with available memory
FAILED tests/cupyx_tests/fallback_mode_tests/test_fallback.py::TestArrayVariants::test_inplace - ValueError: strides is not compatible with available memory


 def _update_numpy_array(self):
        """
        Updates _numpy_array from _cupy_array.
        To be executed before calling numpy function.
    

        if self._supports_cupy:
            # cupy-compatible
            if base is None:
             _array)
            else:
                if not base._remember_numpy:

                        self._numpy_array.shape = self._cupy_array.shape
>                       self._numpy_array.strides = self._cupy_array.strides
E                       ValueError: strides is not compatible with available memory

cupyx/fallback_mode/fallback.py:324: ValueError
```

I think we can remove this fix from here and add it in the corresponding PR, or another option is to skip fallback mode tests when using the system allocator.

I discussed it a while ago with @leofang but is fallback mode being currently used? with the extremely high coverage that CuPy has, maybe it is worth deprecating it

kmaehashi (approved):

jakirkham:
Thanks all! 🙏

Looks like CI is green 😄

Issue #8209, opened 2024-02-23
Pre-populate and ship a Jitify cache?


leofang:
Heard some people complaining about the one-time only cost 
https://docs.cupy.dev/en/stable/upgrade.html#modernized-cccl-support-and-requirement
which appears as if it hangs. 

As a stop-gap solution, we should print a warning "jitify cache building, it may take a while..." 

Short-term, I am thinking to extend [@kmaehashi's idea](https://github.com/cupy/cupy/pull/7934#issuecomment-1763772965) that we implemented
> generate a hash string of all `cupy/_core/include/**` files at the build time and set it as a key

and build the json-based Jitify cache at build time, and ship it as part of the wheel and conda package. 

Long term, I am afraid Jitify's default compile-fail-search loop is too expansive and we might have to either implement it ourselves or use the WIP (https://github.com/NVIDIA/jitify/pull/131).

kmaehashi:
IIRC the problem with pre-generated cache is that some headers depend on files under `/usr/local/cuda/include` which is user environment dependent. 😢

leofang:
I am thinking about this too. I think the bigger problem is that the cache keys can contain includes with relative paths (or truncated paths in the bracket form `<...>` that requires the compiler to search, which makes it very hard to identify the source of the headers. If we normalize all keys to absolute path, maybe it'd make things easier/safer? However, we'll pay a price during cache lookup (de-normalizing an absolute path to relative/truncated path).

kmaehashi:
> If we normalize all keys to absolute path, maybe it'd make things easier/safer?

I think there's no formal way to do that... If `A/B/C.h` has a `#include <D.h>` line, we only get `D.h` as a key, so we don't have knowledge of which file tried to include `D.h` nor which directory `D.h` is in.

leofang:
Right, so when I noted
>  we might have to either implement it ourselves or use the WIP ([NVIDIA/jitify#131](https://github.com/NVIDIA/jitify/pull/131)).

I meant we'll no longer rely on NVRTC to find headers, but instead roll out a search logic that mimics the compiler behavior. This is the only way we can gain knowledge about header location. I think NVIDIA/jitify#131 did quite a bit of the work, I just don't know if in the long term we'd still want to use Jitify, or just vender that search loop implementation in CuPy and move on.

ryanshrott:
Mine has been suck here for 5 hours. How to fix this? @leofang 
[...]

PR #8221, opened 2024-03-06
Do not use plain thrust namespace in complex clone

miscco:
Thrust has introduced a wrapped versioning namespace. This is colliding with the namespace introduced in `cupy/complex`

To avoid that issue pull in the thrust config and use the proper thrust namespace macro

PR #8412, opened 2024-07-02
Disable jitify for cub & Bump CCCL

emcastillo:
Based on Leo's comment in https://github.com/cupy/cupy/issues/8209#issuecomment-2155784474
The CUB generic reduction routines are enabled by default and they use jitify, thus initializing the jitify cache.
Since CUB now fully supports NVRTC we can disable the jitify compilation and avoid the jitify initialization just for running CuPy regular routines.

This PR needs to bump cccl to allow cub to be compatible with nvrtc without using jitify.
I have used the latest commit available in cupy's cccl fork, which seems to include the msvc patch.
However, this patch was greatly changed in https://github.com/NVIDIA/cccl/pull/1420 and I am not sure if there is a regression or not, so let me kick all the CIs.

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -932,7 +932,7 @@ configs {
       image: "windows"
     }
     time_limit {
-      seconds: 36000
+      seconds: 72000
     }
     checkout_strategy {
       include_dot_git: true
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -13,7 +13,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda112"
@@ -31,7 +31,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda112.multi"
@@ -49,7 +49,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda113"
@@ -67,7 +67,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda113.multi"
@@ -85,7 +85,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda114"
@@ -103,7 +103,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda114.multi"
@@ -121,7 +121,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda115"
@@ -139,7 +139,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda115.multi"
@@ -157,7 +157,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda116"
@@ -175,7 +175,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda116.multi"
@@ -193,7 +193,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda117"
@@ -211,7 +211,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda117.multi"
@@ -229,7 +229,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda118"
@@ -247,7 +247,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda118.multi"
@@ -265,7 +265,7 @@ configs {
       gpu: 1
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda120"
@@ -283,7 +283,7 @@ configs {
       gpu: 2
     }
     time_limit: {
-      seconds: 21600
+      seconds: 43200
     }
     environment_variables { key: "GPU" value: "2" }
     command: ".pfnci/linux/main-flexci.sh cuda120.multi"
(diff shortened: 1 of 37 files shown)


emcastillo:
Strange, I can compile and run tests locally with cuda 12.4, I guess it may be the CI gcc version. I am running it with 11.4

kmaehashi:
I've synced https://github.com/cupy/cccl main branch with upstream.

emcastillo:
Thanks! I was able to reproduce the errors in my local env after doing a build from 0.

I will try with the current head, but probably will hit the thrust issues again. I am thinking on having a stable branch with the old cccl and the new nvrtc CUB if nothing else works.

Another quick solution while we harmonize everything would be to disable the CUB backed reduction by default. This will avoid the jitify initialization problem for the majority of users.

emcastillo:
I merged this with #8221, fixed it for NVRTX & and bumped CCCL to the latest commit.

leofang:
Lots of timeouts here, likely because the kernel cache is invalidated. We can tentatively bump the timeout limit as done previously in commit fd00150534c0aef3f803487c7a09262d31dcc124 and revert it once this PR is merged.

kmaehashi:
It's ok to increase the timeout as Leo suggested. Besides that, there are errors reported (before the timeout) which are likely related to this PR, e.g.:

https://ci.preferred.jp/cupy.linux.cuda120/164308/
```
05:17:52.808601 STDOUT 1303]	FAILED cupy_tests/core_tests/test_raw.py::TestRawJitify_param_0_{jitify=False}::test_jitify1	
```

https://ci.preferred.jp/cupy.linux.cuda123/164314/
```
04:52:56.265113 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_0_{block=64, n=10}::test_grid_sync_rawkernel	
04:52:56.265117 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_0_{block=64, n=10}::test_grid_sync_rawmodule	
04:52:56.265124 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_1_{block=64, n=100}::test_grid_sync_rawkernel	
04:52:56.265126 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_1_{block=64, n=100}::test_grid_sync_rawmodule	
04:52:56.265129 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_2_{block=64, n=1000}::test_grid_sync_rawkernel	
04:52:56.265136 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_2_{block=64, n=1000}::test_grid_sync_rawmodule	
04:52:56.265138 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_3_{block=256, n=10}::test_grid_sync_rawkernel	
04:52:56.265154 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_3_{block=256, n=10}::test_grid_sync_rawmodule	
04:52:56.265175 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_4_{block=256, n=100}::test_grid_sync_rawkernel	
04:52:56.265177 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_4_{block=256, n=100}::test_grid_sync_rawmodule	
04:52:56.265180 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_5_{block=256, n=1000}::test_grid_sync_rawkernel	
04:52:56.265186 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawGridSync_param_5_{block=256, n=1000}::test_grid_sync_rawmodule	
04:52:56.292626 STDOUT 1298]	FAILED cupy_tests/core_tests/test_raw.py::TestRawJitify_param_0_{jitify=False}::test_jitify1	
```

kmaehashi:
Btw I noticed timeouts are not happening in CUDA 12.3+ CIs. Maybe thanks to the compile time improvement mentioned in [CUDA 12.3 release notes](https://docs.nvidia.com/cuda/archive/12.3.0/cuda-toolkit-release-notes/index.html#cuda-compilers)? This is great 🎉

leofang:
Discussed with Emilio, `test_jitify1` can be removed since the assumption (that Jitify must be run first) no longer holds. Will check other failures once either of us gets a chance

leofang:
> Will check other failures once either of us gets a chance

The remaining of the failing tests are:
1. using nvcc,
2. including only `cooperative_groups.h` and nothing else
3. have the following bf16/fp16 conversion issues
```
/home/leof/miniforge3/envs/cupy_cuda125_dev/bin/../targets/x86_64-linux/include/cuda_bf16.h(4977): error: no instance of overloaded function "__half::__half" matches the specified type
  __half::__half(const __nv_bfloat16 f)
          ^

1 error detected in the compilation of "/tmp/tmpnoptin0u/4c03e54824ab5700c6f58a6b4b706396b4c17527.cu".
```
This happens because CuPy-bundled fp16 headers (added in CUDA 12.2) are incompatible with the latest bf16 headers (that somehow are loaded unconditionally by nvcc, despite that `cooperative_groups.h` and its dependencies do not include any bf16/fp16 headers.

@emcastillo @kmaehashi May I suggest a two-step approach? I'd really like to get this PR merged asap so that we have time to get it backported to v13.3.0 given its perf/UX improvement. So,
1. We skip the failing grid-sync tests for nvcc >12.2, and get this merged/backported
2. In a separate PR, we add fp16 headers from all CUDA 12.x minor releases (we did this in the old days). Given that the fp16/bf16 headers are rapidly evolved by the math libs team and they move in lock steps, but we can't redistribute bf16 headers, this is necessary.

In the near future, once we depend on nvmath-python, we rely on it to search for headers from pip wheels, conda packages, or system CTK (following [the logic done here](https://github.com/NVIDIA/nvmath-python/blob/7c485842d0f3300e03ec780056936503913910fe/nvmath/device/common_mathdx.py#L35-L90), which might later be refactored into cuda-python). Alternatively, we can also copy/paste the logic here, though I prefer to let us (NVIDIA) maintain this. In any case, once this is done then the bundled headers can again be removed, at the price of a few extra (small) NVIDIA wheels becoming a dependency.

leofang:
@kmaehashi could you grant me write/maintainer access to the cupy/cccl fork? I have a minor fix to address https://github.com/NVIDIA/cccl/issues/2095.

kmaehashi:
> @kmaehashi could you grant me write/maintainer access to the cupy/cccl fork? I have a minor fix to address [NVIDIA/cccl#2095](https://github.com/NVIDIA/cccl/issues/2095). See [cupy/cccl#1](https://github.com/cupy/cccl/pull/1).

Added write permission!

leofang:
It seems the CI failures are in 3 categories
1. Early termination: CI just ended prematurely with the error "execution function exceeded the context's deadline"
2. Memory error: All tests passed normally, the kernel cache statistics was shown, but the CI returned an error "excessive memory usage is detected"
3. Actual test failures: Looking into this...

leofang:
> 2\. In a separate PR, we add fp16 headers from all CUDA 12.x minor releases (we did this in the old days). Given that the fp16/bf16 headers are rapidly evolved by the math libs team and they move in lock steps, but we can't redistribute bf16 headers, this is necessary.

It seems we'll have to do this at least for CUDA 12.5... I can reproduce the `test_vectorize.py` failures from `pfn-public-ci/cupy.linux.cuda125` locally.

miscco:
I tested locally and it should build with cccl main. 

I am not sure whether we need the thrust fix anymore

leofang:
Discussed with @miscco to ensure we're on the same page. Here's what happened with this PR in the past two days:
- I found the NVRTC+Thrust issue, I reported it in https://github.com/NVIDIA/cccl/issues/2095
- I pushed a patch to CuPy's CCCL fork: https://github.com/cupy/cccl/pull/1
- @miscco fixed it in CCCL main the next day: https://github.com/NVIDIA/cccl/pull/2102
- I switched to CCCL main (commit 08e6a3c) to avoid patching myself, and hit the `thrust::less` issue (https://github.com/NVIDIA/cccl/issues/2109)
- I reverted back to use the fork (which is CCCL main as of 2 days ago + my NVRTC patch)

leofang:
Things are looking good, will check back later today for the remaining running CIs. 

@kmaehashi I've eliminated the other two classes of errors and we're now left with the benign one:
> * Memory error: All tests passed normally, the kernel cache statistics was shown, but the CI returned an error "excessive memory usage is detected"

any advice for how to proceed here? Is the CI memory limit configurable on your side?

leofang:
> Is the CI memory limit configurable on your side?

Ah it seems to be configurable in `.pfnci/config.pbtxt`, @kmaehashi is it ok that I bump it to 60 GB? Any bad consequences (such as increased fees)? 😛

leofang:
> Things are looking good, will check back later today for the remaining running CIs.

Confirmed all tests passed 🎉 Now we just need to figure out a path forward for merging (due to the excessive memory error raised at the end of the CI runs).

kmaehashi:
As all CIs have passed let's proceed to merge. If the excessive memory error happens in main branch as well let's consider increasing the size.

(But not sure why this started to happen... In our CI docker image is on the RAM disk, so the cause may be due to increase of the assets and/or real memory usage increase)

jakirkham:
When building Docker images, do you squash layers? Or do you build everything in one layer? If not, that might be a way to cutdown on image size

leofang:
> As all CIs have passed let's proceed to merge. If the excessive memory error happens in main branch as well let's consider increasing the size.
> 
> (But not sure why this started to happen... In our CI docker image is on the RAM disk, so the cause may be due to increase of the assets and/or real memory usage increase)

Note that the OOM errors only happened with Linux + CUDA 11.8 - 12.3; with CUDA 12.4 and above or any CUDA on Windows (that got run in the CI) does not have this issue. Therefore my thinking is it might be related to NVRTC and is something that can be improved in future CUDA. 

Also, I suspect we only need a temporary memory limit increase; once the kernels are cached I hope to see less OOM, though this part is harder to investigate properly yet quickly.

kmaehashi (approved):

leofang:
Thanks to all, @emcastillo @miscco @kmaehashi!

PR #8415, opened 2024-07-03
Bump NumPy/SciPy versions in cuda-example CI

kmaehashi:
This is required because `matplotlib` 3.9.0 (installed [here](https://github.com/cupy/cupy/blob/64c3b26df94907716f8f24f72dcec9676c75533a/.pfnci/linux/tests/actions/example.sh#L10)) now requires NumPy 1.23+.


takagi (approved):
LGTM!

PR #8425, opened 2024-07-12
Add Incomplete Elliptic Integrals to special

steppi:
This PR adds the functions `ellipkinc` and `ellipeinc` to special for Incomplete Elliptic Integrals of the first and second kind respectively.


kmaehashi (approved):
Thanks, LGTM!

PR #8429, opened 2024-07-16
Fix `RandomState.seed()` for NumPy 2 compatibility

seberg:
Without value based casting allowing "safe" casts from signed to unsigned integer scalars, NumPy 2 rejects practically all inputs here.

Unfortunately, even same-kind does not allow for signed -> unsigned so we need to do a larger dance.

The given diff was used together with cudf and cuml to successfully run tests with NumPy 2 and cupy 13.2.

Closes gh-8407, gh-8392

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -810,7 +810,7 @@ class RandomState(object):
                 seed = int(hashlib.md5(seed).hexdigest()[:16], 16)
             else:
                 seed_arr = numpy.asarray(seed)
-                if seed_arr.dtype.kind not in 'iu':
+                if seed_arr.dtype.kind not in 'biu':
                     raise TypeError('Seed must be an integer.')
                 seed = int(seed_arr)
                 # Check that no integer overflow occurred during the cast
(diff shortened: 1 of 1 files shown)


takagi (dismissed):
LGTM!

seberg:
Seems, I missed that booleans are also allowed as seeds, just a one character fix, but...

leofang:
Maybe the changes made in the other repo need to be propagated to here? https://github.com/cupy/cupy-release-tools/pull/381

jakirkham:
Good suggestion. Thanks Leo! 🙏

Also noticed there is `sed` logic, which sometimes doesn't work. So suggested adding new repo files and disabling the old ones: https://github.com/cupy/cupy-release-tools/pull/381#discussion_r1681763282

kmaehashi:
Thanks for the suggestion, that sounds better! (I'm trained to use the `sed` hack because I was often trapped by a chicken-and-egg problem that `yum-config-manager` had to be installed by `yum`, but it was a problem around RHEL 5 or 6 and I guess it's no longer the case)

jakirkham:
Hooray! 🥳

It's all green 🟢

Thanks Kenichi and Masayuki! 🙏

takagi (approved):
LGTM!

takagi:
Thanks all!

PR #8430, opened 2024-07-16
Fix copyto for NumPy 2 compatibility

seberg:
NumPy two has some bad definitions with `copyto` and NEP 50 in that `copyto` currently ignores it. That is a trickier NumPy problem.

However, it is very hard to write code that does the right thing on both versions and even harder to do the same thing as NumPy 2 (because it is not quite consistent).

Short of a version switch, the only good way seems to be to just call `copyto` directly for Python scalar input.

Closes gh-8408, gh-8391

(I'll test this with cuml to be sure.)

---

As you can see, I gave up on trying to find a solution that doesn't just try with NumPy directly.  Yes, unfortunately `can_cast` is almost twice as fast as `copyto` (~200ns for me).
Hopefully that is just small compared to all the other overheads here?

The alternative would be to special case integers and check for integer dtype.

### DIFF
--- a/cupy/_manipulation/basic.py
+++ b/cupy/_manipulation/basic.py
@@ -35,6 +35,7 @@ def copyto(dst, src, casting='same_kind', where=None):
         # NumPy 1.x and 2.0 make implementing copyto cast safety hard, so
         # test whether NumPy copy allows the copy operation:
         numpy.copyto(dst_arr, src, casting=casting)
+        can_cast = True
         src_is_scalar = True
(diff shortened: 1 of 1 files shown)


takagi:
Thanks for this change. I understand that the current numpy copyto has tricky work to replicate, and I suppose the overhead here is acceptable.

takagi (approved):
LGTM!

jakirkham:
Looks like this is passing! 🥳

Anything else needed here? 🙂

jakirkham:
Would it be alright if we merged this?

jakirkham:
Thanks all! 🙏

Issue #8431, opened 2024-07-17
nvcc.exe console popping up while running cupy code which have been cythonized


monzelr:
### Description

I have a python code that uses cupy. This code is being cythonized and built during deployment. After that while running the deployed code whenever cupy needs to compile .cu files (and make use of nvvc.exe), nvvc.exe keep opening and closing in a console on it's own. This is an undesired behavior for a desktop application with an user interface.

### To Reproduce

It is difficult to reproduce since I do not know which cupy functions needs just-in-time compilation of cu files. If you know a function, I can make a small python package, which will be cythonized so we have an example.

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
OS                           : Windows-10-10.0.19045-SP0
Python Version               : 3.8.10
CuPy Version                 : 12.3.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 1.24.4
SciPy Version                : 1.10.1
Cython Build Version         : 0.29.36
Cython Runtime Version       : 0.29.23
CUDA Root                    : C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.6
nvcc PATH                    : C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.6\bin\nvcc.EXE
CUDA Build Version           : 11080
CUDA Driver Version          : 11080
CUDA Runtime Version         : 11060
cuBLAS Version               : (available)
cuFFT Version                : 10600
cuRAND Version               : 10209
cuSOLVER Version             : (11, 3, 2)
cuSPARSE Version             : (available)
NVRTC Version                : (11, 6)
Thrust Version               : 101501
CUB Build Version            : 101501
Jitify Build Version         : 4a37de0
cuDNN Build Version          : (not loaded; try `import cupy.cuda.cudnn` first)
cuDNN Version                : (not loaded; try `import cupy.cuda.cudnn` first)
NCCL Build Version           : None
NCCL Runtime Version         : None
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA T500
Device 0 Compute Capability  : 75
Device 0 PCI Bus ID          : 0000:01:00.0

```

### Additional Information

[...]

PR #8432, opened 2024-07-17
Update compiler.py to avoid the popup of the nvcc.exe console

monzelr:
The popup of the nvcc.exe console only happens in cythonized code.

This will fix issue #8431

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -62,7 +62,7 @@ def _run_cc(cmd, cwd, backend, log_stream=None):
         log = subprocess.check_output(cmd, cwd=cwd, env=env,
                                       stderr=subprocess.STDOUT,
                                       universal_newlines=True,
-                                      creationflags=subprocess.CREATE_NO_WINDOW)
+                                      creationflags=(subprocess.CREATE_NO_WINDOW if _win32 else None))
         if log_stream is not None:
             log_stream.write(log)
         return log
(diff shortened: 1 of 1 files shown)


cupy/cuda/compiler.py
```diff
@@ -61,7 +61,8 @@ def _run_cc(cmd, cwd, backend, log_stream=None):
                 env['PATH'] = path
         log = subprocess.check_output(cmd, cwd=cwd, env=env,
                                       stderr=subprocess.STDOUT,
-                                      universal_newlines=True)
+                                      universal_newlines=True,
+                                      creationflags=subprocess.CREATE_NO_WINDOW)
```
leofang, line 65: This constant is only provided on Windows, I think you might need this
```suggestion
                                      creationflags=(subprocess.CREATE_NO_WINDOW if _win32 else None))
```

cupy/cuda/compiler.py
```diff
@@ -61,7 +61,8 @@ def _run_cc(cmd, cwd, backend, log_stream=None):
                 env['PATH'] = path
         log = subprocess.check_output(cmd, cwd=cwd, env=env,
                                       stderr=subprocess.STDOUT,
-                                      universal_newlines=True)
+                                      universal_newlines=True,
+                                      creationflags=(subprocess.CREATE_NO_WINDOW if _win32 else None))
```
kmaehashi, line 65: The default value looks `0` here 😄
https://docs.python.org/3/library/subprocess.html#subprocess.Popen
```suggestion
                                      creationflags=(subprocess.CREATE_NO_WINDOW if _win32 else 0))
```

kmaehashi (approved):

kmaehashi (approved):

cupy/cuda/compiler.py
leofang: Ah good catch, thanks!

Issue #8440, opened 2024-07-25
Accessing "Name" from importlib.metadata.distributions can error


hmaarrfk:
https://github.com/cupy/cupy/blob/4b7b5ce6e8723b9ad07af64fa076f268f86fd1fc/cupy/_environment.py#L490

```
    installed_names = {d.metadata["Name"]
                       for d in importlib.metadata.distributions()}
```

this code can error if you have importlib_metadata >8.0.0 installed.

I believe that this is due to:
https://github.com/python/importlib_metadata/blob/main/NEWS.rst#deprecations-and-removals
> Message.__getitem__ now raises a KeyError on missing keys. (#371)

I'm sorry I can't really provide a reproducer (yet) ive tried for a bit to remove dependencies from my enviroment but here is a backtrace that shows `KeyError`

```
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/dask/array/__init__.py", line 34, in <module>
    from dask.array import backends, fft, lib, linalg, ma, overlap, random
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/dask/array/backends.py", line 8, in <module>
    from dask.array.core import Array
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/dask/array/core.py", line 30, in <module>
    from dask.array.chunk_types import is_valid_array_chunk, is_valid_chunk_type
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/dask/array/chunk_types.py", line 110, in <module>
    import cupy
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/cupy/__init__.py", line 10, in <module>
    _environment._detect_duplicate_installation()  # NOQA
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/cupy/_environment.py", line 482, in _detect_duplicate_installation
    installed_names = {d.metadata["Name"]
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/cupy/_environment.py", line 482, in <setcomp>
    installed_names = {d.metadata["Name"]
  File "/home/mark/miniforge3/envs/dev/lib/python3.10/site-packages/importlib_metadata/_adapters.py", line 54, in __getitem__
    raise KeyError(item)
KeyError: 'Name'
```

I hope this helps.

jakirkham:
Hey Mark, what kind of filesystem is this build happening on? Is it NFS?

[...]

PR #8441, opened 2024-07-25
Address KeyErrors from importlib_metadata

hmaarrfk:
Closes: https://github.com/cupy/cupy/issues/8440

### DIFF
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -487,7 +487,13 @@ def _detect_duplicate_installation():
         'cupy-rocm-4-3',
         'cupy-rocm-5-0',
     }
-    installed_names = {d.metadata["Name"]
+    # use metadata.get to be resilient to namespace packages
+    # that may be leftover in the user's path???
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thanks @hmaarrfk! This PR itself looks fine and I'm happy to merge, but curious why this happens, because CuPy uses `importlib.metadata` and not `importlib_metadata` (which is from the PyPI package). Is there a platform doing `importlib.metadata = importlib_metadata` somewhere?

kmaehashi (approved):

hmaarrfk:
Hmmm. I understand that difference but all I can tell is that downgrading importlib_metadata helped resolve things for me.

I can investigate more but I feel like it is related to leftover files after a package is uninstalled. Which is really hard to simulate.

kmaehashi:
Thanks @hmaarrfk, let's merge as anyways it's needed for Python 3.14 according to https://github.com/python/importlib_metadata/issues/497 😄

PR #8442, opened 2024-07-25
Support system allocated memory

rongou:
Support system allocated memory (SAM). This is a new way to allocate device-accessible memory in two environments:
* HMM: Heterogeneous Memory Management, a software solution to support SAM in x86 PCIe based systems.
* ATS: Address Translation Service, a hardware solution on the Grace Hopper superchip.

Mostly based on @leofang's draft change with some cleanups.

@emcastillo


cupy/_core/_dtype.pyx
```diff
@@ -124,3 +124,41 @@ cpdef void _raise_if_invalid_cast(
...
+    elif dtype == "e":
+        buf.format = 'e'
+    elif dtype == "f":
+        buf.format = 'f'
+    elif dtype == "d":
+        buf.format = 'd'
+    elif dtype == "F":
+        buf.format = 'Zf'
+    elif dtype == "D":
+        buf.format = 'Zd'
```
jakirkham, line 162: It might be worth putting this in a `dict` and just using that here

For example, in this Cython code, we added [a `dict` for lookups]( https://github.com/rapidsai/ucx-py/blob/5ceff39ee6069f03a80e054a2c612c8d9c45edc3/ucp/_libs/arr.pyx#L32 ), which we use [in this block]( https://github.com/rapidsai/ucx-py/blob/5ceff39ee6069f03a80e054a2c612c8d9c45edc3/ucp/_libs/arr.pyx#L104-L113 )

This is actually pretty performant especially when Python `intern`s the strings because then the hash of the `str` is known and can be used to grab the key right away

jakirkham:
cc @seberg (as this may be of interest)

rongou:
@leofang @emcastillo merged current main to the draft change to preserve commit history.

cupy/_core/_dtype.pyx
rongou: Done.

leofang:
It seems I don't have privilege to trigger CIs for first-time contributors lol

jakirkham:
Maybe Rong could do a small doc or bug fix in another PR so that he establishes his contributor cred?

leofang:
@rongou could you undo commit c95585dc15b0b88fa3d90ecb6c1e834326e24fa8? It is needed in order to ensure it's functional on using malloc'd memory on G+H. I'll probably have to find time to write a full PR description to explain some implementation details. Will share with you a note offline momentarily in the meanwhile.

rongou:
> @rongou could you undo commit [c95585d](https://github.com/cupy/cupy/commit/c95585dc15b0b88fa3d90ecb6c1e834326e24fa8)? It is needed in order to ensure it's functional on using malloc'd memory on G+H. I'll probably have to find time to write a full PR description to explain some implementation details. Will share with you a note offline momentarily in the meanwhile.

From the CI results looks likt it breaks windows builds. We'd need to either add a windows implementation, or guard it against compiling under windows.

leofang:
> From the CI results looks likt it breaks windows builds. We'd need to either add a windows implementation, or guard it against compiling under windows.

Ah, interesting. I see two ways out
- To support Windows, we need to change to `_aligned_malloc` and add a `free` wrapper that calls `_aligned_free`
   - xref: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=msvc-170
- Or, we add line 131 in `install/cupy_builder/_features.py` conditioned on whether the build/target platform is Windows

leofang:
Perhaps it's best to skip Windows (option 2) since managed memory is not doing great on Windows and there's no Windows support for G+H anyway.

rongou:
Added back with windows support.

emcastillo:
I will try to make any changes needed to land this PR. So far is there any concern that hasn't been addressed?

@kmaehashi can you assign me to get permission to write to @rongou's branch? Thanks!

kmaehashi:
> @kmaehashi can you assign me to get permission to write to @rongou's branch? Thanks!

I think the permission is already granted 😃

leofang:
Sorry for late reply. Can I also get access to @rongou's branch, please? There are few changes I'd like to add.

kmaehashi:
> Sorry for late reply. Can I also get access to @rongou's branch, please? There are few changes I'd like to add.

This is related to the admin permission to the main CuPy repo itself and I'd needs some assessment before doing this. Would you mind opening another PR for now? 🙇🏼

cupy/_core/core.pyx
```diff
@@ -52,9 +56,6 @@ from cupy_backends.cuda.api cimport runtime
 from cupy_backends.cuda.libs cimport nvrtc
 
 
-NUMPY_1x = numpy.__version__ < '2'
-
-
```
leofang, line 57: TODO: revert

cupy/_core/core.pyx
```diff
@@ -91,6 +92,18 @@ cdef tuple _HANDLED_TYPES
 
 cdef object _null_context = contextlib.nullcontext()
 
+cdef bint _is_sam_enabled = (int(os.environ.get('CUPY_ENABLE_SAM', '0')) != 0)
```
leofang, line 95: I think the renaming from HMM (what we had) to SAM is problematic, as SAM is specific to only `malloc` while HMM can be also used to refer to managed memory (which I didn't have time to cover). However I'll let go for now.

cupy/_core/core.pyx
```diff
@@ -342,6 +354,44 @@ cdef class _ndarray_base:
...
 
+    def __getbuffer__(self, Py_buffer* buf, int flags):
+        # TODO(leofang): use flags
+        if (not is_sam_supported(self.data.device_id)
+                or not self.is_host_accessible()):
+            raise RuntimeError(
+                'Accessing a CuPy ndarry on CPU is not allowed except when '
+                'using system memory (on HMM or ATS enabled systems, need to '
+                'set CUPY_ENABLE_SAM=1) or managed memory')
+
```
leofang, line 365: Note: I had a TODO here mentioning the possibility of prefetching. However, it seems prefetching is slow and I did not have time to build a proper heuristic for prefetching.

cupy/_core/internal.pxd
```diff
@@ -64,3 +64,9 @@ cpdef int _update_order_char(
     bint is_c_contiguous, bint is_f_contiguous, int order_char)
 
 cpdef tuple _broadcast_shapes(shapes)
+
+cdef bint _is_layout_expected(
+        const bint c_contiguous, const bint f_contiguous,
+        expected_order) except*
+
+cdef bint _is_alignment_expected(intptr_t ptr, int min_size) noexcept
```
leofang, line 72: unused
```suggestion
```

cupy/_core/internal.pyx
```diff
@@ -535,3 +535,30 @@ cpdef tuple _broadcast_shapes(shapes):
...
+
+
+cdef bint _is_alignment_expected(intptr_t ptr, int min_size) noexcept:
+    cdef int alignment = 1
+    # TODO(leofang): use the chunk size from memory.pyx and don't hard-code?
+    while ptr % alignment == 0 and alignment < 256:
+        alignment *= 2
+    if alignment < min_size:
+        return False
+    return True
```
leofang, line 564: unused
```suggestion
```

cupy/_core/core.pyx
```diff
@@ -2579,22 +2644,91 @@ cdef _ndarray_base _array_from_nested_cupy_sequence(
...
+    obj_flags = obj.flags
+    if not internal._is_layout_expected(
+            obj_flags.c_contiguous, obj_flags.f_contiguous, order):
+        return None
+
+    cdef intptr_t ptr = obj.ctypes.data
+#    # NumPy ndarrays do not guarantee alignment unless a custom
+#    # allocator is in use
+#    if not _is_alignment_expected(ptr, obj_dtype.itemsize):
+#        return None
```
leofang, line 2681: unused
```suggestion
```

cupy/_core/core.pyx
```diff
@@ -2579,22 +2644,91 @@ cdef _ndarray_base _array_from_nested_cupy_sequence(
...
+
+
 cdef _ndarray_base _array_default(
-        obj, dtype, order, Py_ssize_t ndmin, bint blocking):
+        obj, dtype, bint copy, order, Py_ssize_t ndmin, bint blocking):
+    cdef _ndarray_base a
+
+    # Fast path: zero-copy a NumPy array if possible
+    if not blocking:
+        a = _try_skip_h2d_copy(obj, dtype, copy, order, ndmin)
```
leofang, line 2708: Note: we were wondering if this can be called unconditionally, regardless of `blocking`. Should be revisited in another PR.

cupy/_core/core.pyx
```diff
@@ -2579,22 +2644,91 @@ cdef _ndarray_base _array_from_nested_cupy_sequence(
...
+    if a is not None:
+        return a
+
     if order is not None and len(order) >= 1 and order[0] in 'KAka':
         if isinstance(obj, numpy.ndarray) and obj.flags.fnc:
             order = 'F'
         else:
             order = 'C'
-    copy = False if NUMPY_1x else None
```
leofang, line 2589: TODO: we need to reconcile this for NumPy 2.

cupy/_core/internal.pxd
```diff
@@ -1,6 +1,6 @@
 from libcpp cimport bool as cpp_bool
 from libcpp cimport vector
-from libc.stdint cimport uint16_t
+from libc.stdint cimport intptr_t, uint16_t
```
leofang, line 3:
```suggestion
from libc.stdint cimport uint16_t
```

cupy/_core/internal.pyx
```diff
@@ -3,7 +3,7 @@
 cimport cpython  # NOQA
 cimport cython  # NOQA
 from libcpp cimport bool as cpp_bool
-from libc.stdint cimport uint32_t
+from libc.stdint cimport intptr_t, uint32_t
```
leofang, line 6:
```suggestion
from libc.stdint cimport uint32_t
```

cupy/_core/core.pyx
emcastillo: I think is better to decide a name and stick to it now than change it later and having "deprecated" options

cupy/_core/numpy_allocator.pyx
```diff
@@ -0,0 +1,77 @@
...
+        void* _calloc(size_t nmemb, size_t size);
+        void* _malloc(size_t size);
+        void* _realloc(void *ptr, size_t size);
+        void _free(void* ptr);
+    }
+    """
+    void* _calloc(size_t nmemb, size_t size) nogil
+    void* _malloc(size_t size) nogil
+    void* _realloc(void *ptr, size_t size) nogil
+    void _free(void* ptr) nogil
```
leofang, line 43: I think I was probably wrong in suggesting this in a past inline comment. We should just set `CYTHON_EXTERN_C` at build time, see e.g.
- https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#c-public-declarations
- https://github.com/rapidsai/rmm/pull/1350

leofang (changes requested):

cupy/cuda/memory.pyx
```diff
@@ -188,11 +199,21 @@ cdef class UnownedMemory(BaseMemory):
...
         self.ptr = ptr
         self._owner = owner
 
+    def __dealloc__(self):
+        # Note: Cannot raise in the destructor! (cython/cython#1613)
+        if _is_sam_enabled:
+            # we don't own the memory, we must sync before free to avoid any
+            # race condition
+            runtime.streamSynchronize(stream_module.get_current_stream_ptr())
+
```
leofang, line 216: Q: @rongou this is new to me. How would this destructor be triggered?

emcastillo: Unowned memory is not pooled right? so it should trigger once the gc kicks in but not 100% sure

docs/source/user_guide/memory.rst
```diff
@@ -198,3 +198,60 @@ Be sure to do this before any other CuPy operations.
 
    # Disable memory pool for pinned memory (CPU).
    cupy.cuda.set_pinned_memory_allocator(None)
+
+
+System allocated memory (SAM) support (**experimental!**)
```
leofang, line 203: ditto, we might not want to call it SAM here

docs/source/user_guide/memory.rst
```diff
@@ -198,3 +198,60 @@ Be sure to do this before any other CuPy operations.
...
+.........................................................
+
+It is possible to make both NumPy and CuPy use and share system allocated memory on Heterogeneous Memory Management
+(HMM) enabled systems or the NVIDIA Grace Hopper Superchip. To activate this capability, currently you need to:
+
+1. Build CuPy from source and install `numpy_allocator <https://github.com/inaccel/numpy-allocator/>`_
```
leofang, line 209: The doc was written during development, now that we're upstreaming it we can simplify it 🙂 
```suggestion
1. Install `numpy_allocator <https://github.com/inaccel/numpy-allocator/>`_
```

docs/source/user_guide/memory.rst
```diff
@@ -198,3 +198,60 @@ Be sure to do this before any other CuPy operations.
...
+
+For CuPy, alternatively you can use the CuPy allocator from RAPIDS Memory Manager (RMM):
+
+.. code-block:: py
+
+    import cupy as cp
+    import rmm
+    mr = rmm.mr.SystemMemoryResource()
+    rmm.mr.set_current_device_resource(mr)
+    cp.cuda.set_allocator(rmm.allocators.cupy.rmm_cupy_allocator)
```
leofang, line 241: This is where I don't get. Right now we rely on `mem.identity` to return `"SystemMemory"` or `"ManagedMemory"`. Does RMM also tag the memory pointer object in the same way? If not, the fast path would not be triggered.

docs/source/user_guide/memory.rst
```diff
@@ -198,3 +198,60 @@ Be sure to do this before any other CuPy operations.
...
+set its preferred location to the GPU if the allocation would not eat into the headroom:
+
+.. code-block:: py
+
+    import cupy as cp
+    import rmm
+    headroom = 1 << 30  # 1 GiB
+    mr = rmm.mr.SamHeadroomMemoryResource(headroom=headroom)
+    rmm.mr.set_current_device_resource(mr)
+    cp.cuda.set_allocator(rmm.allocators.cupy.rmm_cupy_allocator)
```
leofang, line 255: We should have a follow-up PR adding an example with CuPy's mempool.

tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
```diff
@@ -325,8 +325,7 @@ def test_firls(self, xp, scp):
         N = 11  # number of taps in the filter
         a = 0.1  # width of the transition band
         # design a halfband symmetric low-pass filter
-        h = scp.signal.firls(N, [0, a, 0.5-a, 0.5], [1, 1, 0, 0], fs=1.0)
-        return h
+        return scp.signal.firls(N, [0, a, 0.5-a, 0.5], [1, 1, 0, 0], fs=1.0)
```
leofang, line 328: Q: why do we change it here?

cupy/cuda/memory.pyx
leofang: Right, but for the purpose of this PR we don't use `UnownedMemory` so I am confused..

emcastillo: I think this may be due to an external library interaction, probably RMM interop?

tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
emcastillo: Probably this is a left over of when I was fixing the tests

cupy/_core/core.pyx
leofang: How about `CUPY_ENABLE_UMP`? UMP stands for "Unified Memory Programming", which puts emphasis on not only the change of underlying memory resource (ex: CuPy users can already use managed memory today), but also the change of the programming model (no distinction in memory space anymore, only in execution space).

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#unified-memory-programming

cupy/cuda/memory.pyx
leofang: Yeah I feel this change was to mirror that from RMM: https://github.com/rapidsai/rmm/pull/1655 but this is problematic because if [RMM uses `UnownedMemory`](https://github.com/rapidsai/rmm/blob/ab6e2961d7b8f833f688775e941c4e2ed2bd4d8a/python/rmm/rmm/allocators/cupy.py#L39) then the fast path would not be triggered (https://github.com/cupy/cupy/pull/8442#discussion_r1774375148).

cupy/_core/core.pyx
```diff
@@ -342,6 +354,44 @@ cdef class _ndarray_base:
...
+        cpython.Py_INCREF(self)
+
+        stream_module.get_current_stream().synchronize()
+
+    def __releasebuffer__(self, Py_buffer* buf):
+        stdlib.free(buf.shape)  # frees both shape & strides
+        cpython.Py_DECREF(self)
+
+    cdef inline bint is_host_accessible(self) except*:
+        return self.data.mem.identity in ('SystemMemory', 'ManagedMemory')
```
leofang, line 393: Note: This is where we have a strict requirement for what kinds of memories that can be used for unified np/cp programming

tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
leofang: ah I see, will fix it to minimize diff, thanks Emilio

cupy/cuda/memory.pyx
leofang: Also this destructor would cause unconditional synchronization when CUPY_ENABLE_HMM=1 for all unowned memory, regardless of its source (i.e. whether it's system memory, or whether it comes from RMM/RAPIDS). I think it causes more harm than the perf it gains.

cupy/_core/core.pyx
leofang: I am actually confused here. When this feature was developed, NumPy 2 was not released yet so we must be using NumPy 1. However, I do not believe `copy` is always `False` at this point, so I think @rongou removing this line actually is the right move. @emcastillo please poke holes 😄

cupy/_core/core.pyx
emcastillo: I like this

tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
leofang: fixed in [7af49ef](https://github.com/cupy/cupy/pull/8442/commits/7af49ef482d76913bc0e4ca14d386d1e75657a06)

cupy/cuda/memory.pyx
leofang: fixed in [73d0fe8](https://github.com/cupy/cupy/pull/8442/commits/73d0fe85d5ad53ff5c7123baf6648f205c7113e9)

cupy/_core/numpy_allocator.pyx
leofang: fixed in [c9298fb](https://github.com/cupy/cupy/pull/8442/commits/c9298fba9500b3a0c619b13596f90bdc25cd8fdc)

cupy/_core/core.pyx
leofang: changed to UMP in [846682f](https://github.com/cupy/cupy/pull/8442/commits/846682f1d23deaf109b3edbf7d49fb5f520d735e)

cupy/_core/core.pyx
leofang: xref: https://github.com/cupy/cupy/pull/8442#discussion_r1774510340

docs/source/user_guide/memory.rst
leofang: changed to UMP in [846682f](https://github.com/cupy/cupy/pull/8442/commits/846682f1d23deaf109b3edbf7d49fb5f520d735e)

docs/source/user_guide/memory.rst
leofang: This is still a TODO, but can be addressed in a separate PR

leofang:
> I'll probably have to find time to write a full PR description to explain some implementation details. 

In this PR, we add support for HMM/ATS memory systems so that users can opt-in and use `malloc` or `cudaMallocManaged` to allocate memory that is accessible to both CPU and GPU. This is a programming model feature allowing users of NVIDIA Grace Hopper systems to take full advantage of the coherent memory system.

This opt-in mechanism is currently guarded in a few steps, as outlined in both the doc change as well as test setup:
1. Set `CUPY_ENABLE_UMP=1`
2. Install `numpy_allocator`
   - note: I just found it's currently blocked on Python 3.12 ☹️ https://github.com/inaccel/numpy-allocator/issues/3
4. Set an allocator for both NumPy (CPU) and CuPy (GPU)

The net effect is we use NumPy and CuPy to represent CPU and GPU **execution** spaces, respectively, but the **memory** space is now unified, and memory transfer between CPU and GPU will become no-op if conditions are met. A proper stream synchronization is inserted wherever applicable, in order to ensure CPU/GPU do not have race condition.
In particular, the D2H copies are skipped by supporting the Python buffer protocol, which is generic enough for any host access not just NumPy.

Allocators for both NumPy and CuPy are needed because
- CuPy: we need a mechanism to 
   - allocate system memory (via `malloc`)
   - identify where the memory is from, quickly (via `mem.identity`)
- NumPy: we need to ensure allocated memory has right alignment for CuPy's kernels to consume, in order to avoid data corruption

Due to `numpy_allocator`'s requirement, we expose the memory allocation routines' symbols via `cdef public`, but with C instead of C++ linkage.

leofang:
Rendered document: https://cupy--8442.org.readthedocs.build/en/8442/user_guide/memory.html#unified-memory-programming-ump-support-experimental

cupy/cuda/memory.pyx
rongou: Yes RMM uses UnownedMemory. We probably should change that.

I think I originally added this because some cupy allocated buffers were causing problems. Later ran into issue with cuml c++ allocated buffers causing similar problems. Now that we synchronize the stream in rmm, we don't need to synchronize here any more.

rongou:
I'm not sure `cudaMallocManaged` is relevant here. Although a buffer allocated through it can be accessed from both cpu and gpu, we'd have to explicitly call it, which any non-CUDA code is definitely not doing. But with HMM/ATS, any system `malloc` allocated memory can be potentially accessed from the gpu, which is main issue addressed in this PR. That's why I think spelling out `SAM` is helpful to clarify the issue.

leofang:
Calling out `cudaMallocManaged` is important because from the allocator (ex: RMM 🙂) perspective it does not really matter if under the hood it's `malloc` or `cudaMallocManaged` in use. As long as a memory resource is both host-/device- accessible, they can be used here. (Though, I think there are a few pieces missing to fully support `cudaMallocManaged` as far as unified memory programming is concerned. We need to address this in a separate PR.) Another important reason is that in the early G+H driver (~CUDA 12.2 IIRC) the performance of `cudaMallocManaged` was better than `malloc` in certain use cases. The gap is closing but still not on par AFAIK.

rongou:
But isn't this PR mainly about avoiding unnecessary copies with SAM? For example, `numpy` would never allocate memory through `cudaMallocManaged`, so you do need to copy in that case.

leofang:
This PR already has the necessary mechanism in place (https://github.com/cupy/cupy/pull/8442#discussion_r1774390960) to handle zero-copy managed memory. One of the missing things that we can add later is, as you pointed out, a different `numpy_allocator` C APIs that calls `cudaMallocManaged` instead of `malloc` for NumPy to use. (And, the alignment treatment would be simpler there, because `cudaMallocManaged` is also 256B-aligned IIRC, same as `cudaMalloc`.)

cupy/_core/core.pyx
leofang: @emcastillo perhaps we should run the test suite twice locally? One with NumPy 1.26 and another with 2.0...

leofang:
https://developer.nvidia.com/blog/simplifying-gpu-application-development-with-heterogeneous-memory-management/

cupy/cuda/memory.pyx
```diff
@@ -188,6 +199,13 @@ cdef class UnownedMemory(BaseMemory):
             runtime._ensure_context()
             ptr_attrs = runtime.pointerGetAttributes(ptr)
             device_id = ptr_attrs.device
+            if device_id == runtime.cudaCpuDeviceId:
+                # this happens with SystemMemory...
+                device_id = device.get_device_id()
+            # CUDA doesn't track memory allocated through the system malloc
+            if device_id == runtime.cudaInvalidDeviceId and _is_ump_enabled:
```
leofang, line 206: @emcastillo I am wondering if we can just remove `_is_ump_enabled`. It seems I did so too in all the `prefetch` changes I did in this file. Thoughts?

For the record, that `cudaInvalidDeviceId` is raised is likely related to nvbugs 4300350.

docs/source/user_guide/memory.rst
```diff
@@ -198,3 +198,51 @@ Be sure to do this before any other CuPy operations.
...
+
+    import cupy as cp
+    cp.cuda.set_allocator(cp.cuda.MemoryPool(cp.cuda.memory.malloc_system).malloc)
+
+4. Switch to the aligned allocator for NumPy to draw system memory
+
+.. code-block:: py
+   
+    import cupy._core.numpy_allocator as ac
+    import numpy_allocator
```
emcastillo, line 224: This doesn't work with numpy 2.0 :( Seems like they compiled it using numpy v1.x. Also when trying to build the package from scratch it resorts to numpy v1.22 for the build

cupy/cuda/memory.pyx
emcastillo: I am ok with removing it. I need to test the case where UMP is not enabled and we pass a garbage pointer. If the execution fails before reaching this point I assume it is safe to do so.

docs/source/user_guide/memory.rst
leofang: xref: https://github.com/inaccel/numpy-allocator/issues/4

leofang: To close the loop the upstream has fixed the issue and released new wheels and conda packages! https://pypi.org/project/numpy-allocator/1.2.1/

cupy/cuda/memory.pyx
```diff
@@ -243,6 +263,89 @@ cdef class ManagedMemory(BaseMemory):
...
+    """
+
+    def __init__(self, size_t size):
+        self.size = size
+        # TODO(leofang): using the GPU id may not be ideal, but setting it
+        # to cudaCpuDeviceId (-1) would require a lot of changes
+        self.device_id = device.get_device_id()
+        self.ptr = 0
+        if size > 0:
+            self.ptr = <intptr_t>c_malloc(size)
```
kmaehashi, line 292: Q: Is it ok even if `size` is unaligned to 512?

cupy/cuda/memory.pyx
```diff
@@ -243,6 +263,89 @@ cdef class ManagedMemory(BaseMemory):
...
+        self._owner = None
```
kmaehashi, line 293: I think it's better to have another flag to indicate the ownership of the pointer. My  concern is that if `from_external(ptr, size, owner=None)` is called, `ptr` to be freed by the destructor, which is unexpected.

(Or, adding `assert owner is None` in `from_external` may also be fine)

cupy/cuda/memory.pyx
```diff
@@ -799,6 +937,29 @@ cdef class PooledMemory(BaseMemory):
...
+        if isinstance(chunk.mem, Memory):
+            self.identity = "Memory"
+        elif isinstance(chunk.mem, MemoryAsync):
+            self.identity = "MemoryAsync"
+        elif isinstance(chunk.mem, UnownedMemory):
+            self.identity = "UnownedMemory"
+        elif isinstance(chunk.mem, SystemMemory):
+            self.identity = "SystemMemory"
+            self.prefetch = <SystemMemory>(chunk.mem).prefetch
+            self.advise = <SystemMemory>(chunk.mem).advise
```
kmaehashi, line 951: Q: Who calls these functions (`prefetch` and `advise`)?

docs/source/user_guide/memory.rst
```diff
@@ -198,3 +198,51 @@ Be sure to do this before any other CuPy operations.
 
    # Disable memory pool for pinned memory (CPU).
    cupy.cuda.set_pinned_memory_allocator(None)
+
+
+Unified memory programming (UMP) support (**experimental!**)
```
kmaehashi, line 203: Could you change the title to make it clear that the feature is unrelated to managed memory (which is often referred as [Unified Memory](https://developer.nvidia.com/blog/unified-memory-cuda-beginners/))?
This is the doc often referred by, so I'd avoid confusing majority of users who don't own G+H 😅

tests/cupy_tests/lib_tests/test_polynomial.py
```diff
@@ -5,6 +5,7 @@
 from cupy.cuda import driver
 from cupy.cuda import runtime
 import cupyx
+import cupyx.lapack
```
kmaehashi, line 8: Seems unrelated to this PR?

tests/cupyx_tests/tools_tests/test_install_library.py
```diff
@@ -19,6 +19,9 @@ def _get_supported_cuda_versions(lib):
 
 class TestInstallLibrary:
 
+    @pytest.mark.skipif(
+        platform.processor() == "aarch64",
```
kmaehashi, line 23:
```suggestion
        platform.machine() == "aarch64",
```

tests/cupyx_tests/tools_tests/test_install_library.py
```diff
@@ -27,11 +30,17 @@ def test_install_cudnn(self, cuda):
     @pytest.mark.skipif(
         platform.system() == 'Windows',
         reason='NCCL is only available for Linux')
+    @pytest.mark.skipif(
+        platform.processor() == "aarch64",
```
kmaehashi, line 34:
```suggestion
        platform.machine() == "aarch64",
```

tests/cupyx_tests/tools_tests/test_install_library.py
```diff
@@ -27,11 +30,17 @@ def test_install_cudnn(self, cuda):
...
+        reason="FIXME")  # TODO(leofang)
     @pytest.mark.parametrize('cuda', _get_supported_cuda_versions('nccl'))
     @testing.slow
     def test_install_nccl(self, cuda):
         self._test_install('nccl', cuda)
 
+    @pytest.mark.skipif(
+        platform.processor() == "aarch64",
```
kmaehashi, line 42:
```suggestion
        platform.machine() == "aarch64",
```

cupy/_core/_dtype.pyx
```diff
@@ -124,3 +124,28 @@ cpdef void _raise_if_invalid_cast(
...
+    intern("B"): intern("B"),
+    intern("H"): intern("H"),
+    intern("I"): intern("I"),
+    intern("L"): intern("L"),
+    intern("Q"): intern("Q"),
+    intern("e"): intern("e"),
+    intern("f"): intern("f"),
+    intern("d"): intern("d"),
+    intern("F"): intern("Zf"),
+    intern("D"): intern("Zd"),
```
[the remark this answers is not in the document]
leofang: No, not stupid at all, it appears to be a NumPy thing, but because we're using NumPy dtypes I just followed along.
https://github.com/numpy/numpy/blob/d1724e3a5f37a2860dc9ad2b29f0e5a3501d9848/numpy/_core/src/multiarray/buffer.c#L396-L397
Example:
```python
>>> import numpy as np
>>> a = np.empty(10, dtype=np.complex128)
>>> m = memoryview(a)
>>> m.format
'Zd'
```

cupy/cuda/memory.pyx
leofang: xref: https://github.com/cupy/cupy/pull/8442#discussion_r1774360929

I had `prefetch`/`advise` added when experimenting the support but I commented them out later due to performance degradation (it adds too much overhead to small arrays). The comment was removed in commit f618bab75.

tests/cupy_tests/lib_tests/test_polynomial.py
leofang: Yeah for some reason local tests failed without adding this, I was puzzled too..

cupy/_core/_dtype.pyx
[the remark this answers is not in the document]
seberg: I actually couldn't find good docs for it, it is here: https://peps.python.org/pep-3118/#additions-to-the-struct-string-syntax
(Sorry, very randomly saw this, but may look at this a bit more some time.  If you think I should take a close look, please give a ping.)

cupy/_core/core.pxd
```diff
@@ -24,6 +24,7 @@ cdef class _ndarray_base:
         readonly _ndarray_base base
 
     cdef _init_fast(self, const shape_t& shape, dtype, bint c_order)
+    cdef inline bint is_host_accessible(self) except*
```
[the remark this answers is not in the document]
leofang: I can rename it if desired, but `cdef` methods are already "private" as far as Python users are concerned, no?

cupy/cuda/memory.pyx
leofang: Excellent question! We are protected by `_round_size()` which always rounds up to units of 512B before we arrive here.

cupy/_core/_dtype.pyx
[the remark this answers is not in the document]
leofang: Nice, thanks @seberg! I see `Zd` is called out here:
https://peps.python.org/pep-3118/#examples-of-data-format-descriptions

cupy/cuda/memory.pyx
```diff
@@ -243,6 +263,89 @@ cdef class ManagedMemory(BaseMemory):
...
+            owner (object): Reference to the owner object to keep the memory
+                alive.
+        """
+        cdef SystemMemory self = SystemMemory.__new__(SystemMemory)
+        self.size = size
+        # TODO(leofang): using the GPU id may not be ideal, but setting it
+        # to cudaCpuDeviceId (-1) would require a lot of changes
+        self.device_id = device.get_device_id()
+        self.ptr = ptr
+        self._owner = owner
```
leofang, line 311:
```suggestion
        assert owner is not None, 'must provide an owner'
        self._owner = owner
```

cupy/cuda/memory.pyx
leofang: Added an assert in commit 00c3716.

docs/source/user_guide/memory.rst
leofang: This was purposely changed to reflect the importance of including managed memory here, see the discussion with Rong [starting here](https://github.com/cupy/cupy/pull/8442#issuecomment-2374444257), and the linked devblog. From CUDA Programming Model's perspective (and therefore CuPy), it does not matter if we are using 
- malloc on G+H
- cudaMallocManaged on G+H
- cudaMallocManaged on x84-64

everything here (after we fix a few minor things in a subsequent PR) will work. The performance characteristics may be different, but it is a separate discussion.

cupy/_core/core.pyx
```diff
@@ -185,14 +201,10 @@ cdef class _ndarray_base:
...
         del shape
 
+        # this would raise if order is not recognized
         cdef int order_char = (
             b'C' if order is None else internal._normalize_order(order))
 
-        # `strides` is prioritized over `order`, but invalid `order` should be
-        # checked even if `strides` is given.
-        if order_char != b'C' and order_char != b'F':
-            raise ValueError('order not understood. order=%s' % order)
```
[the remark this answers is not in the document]
leofang: I think this is dead code that we'd never reach, see the comment at line 204 that I added above.

cupy/_core/core.pyx
```diff
@@ -1838,7 +1892,21 @@ cdef class _ndarray_base:
...
-        else:
+
+        if a_cpu is None:
+            # we don't check is_ump_supported() etc here because it'd be
+            # done later
+            if _is_ump_enabled:
+                try:
+                    # return self to use the same memory and avoid copy
+                    a_cpu = numpy.asarray(self, order=order)
+                except TypeError:
```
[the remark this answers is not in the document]
leofang: IIRC I only check if the user has set the env var to enable zero copy, but I skip (defer) the check for hw support until the buffer protocol is tried. If the hw does not support zero copy, buffer creation would fail, and NumPy ends up raising this error.

cupy/cuda/memory.pyx
leofang: Let me know if you prefer to not expose this for now.

cupy/_core/core.pxd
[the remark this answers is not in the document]
kmaehashi: ah you're correct 😄 
I think adding `_` better aligns with `_init_fast` but not mandatory.

tests/cupy_tests/lib_tests/test_polynomial.py
kmaehashi: I see... Then could you add a comment here? Contributors with a clever linter setup may remove it by chance 😅

cupy/cuda/memory.pyx
kmaehashi: Thanks, got it. Let's keep it as is.

tests/cupy_tests/lib_tests/test_polynomial.py
leofang: I am now confused why the linter wouldn't complain about unused imports... @emcastillo do you recall what error we ran into without this change?

emcastillo: Sorry, I can't remember the exact reason. I got some weird import error when kicking the tests.
I will try to check if they run without this line in the most recent commit.

cupy/_core/core.pyx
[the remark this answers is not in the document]
kmaehashi: Hmm, I tried to raise `RuntimeError` in `__getbuffer__` but not getting TypeError here. Am I missing something?

```py
# dummy_buf.pyx
# To compile: cythonize -b -3 dummy_buf.pyx

import numpy as np

cdef class Dummy:
    def __getbuffer__(self, Py_buffer *buffer, int flags):
        print('__getbuffer__ called')
        raise RuntimeError
    def __releasebuffer__(self, Py_buffer *buffer):
        print('__releasebuffer__ called')
        raise RuntimeError

def test():
    d = Dummy()
    print(d)
    a = np.asarray(d)
    print(type(a), a.dtype)
```

```sh
$ python -c 'import dummy_buf; dummy_buf.test()'
<dummy_buf.Dummy object at 0x7fe3e718ac00>
__getbuffer__ called
<class 'numpy.ndarray'> object
```

tests/cupy_tests/lib_tests/test_polynomial.py
emcastillo: @leofang tests pass now without this. It should be ok to remote it

kmaehashi: It seems flake8 overlooks this case. I guess `import cupyx.lapack` is considered in-use because `cupyx` is used in code.

Simple reproducer:

```
import os
import os.path
import sys  # flake8 only warns about this


print(os.uname())
```

cupy/_core/core.pyx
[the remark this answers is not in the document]
leofang: @kmaehashi I think I missed the second half of the story (sorry it's been a while since I worked on this feature...). `TypeError` is raised in `__array__`:
https://github.com/cupy/cupy/blob/678add80cdd82e89f2b751f1722035c53997ad03/cupy/_core/core.pyx#L1489-L1493, which can be easily reproduced by adding a dummy `def __array__(self): raise TypeError` to your snippet above. We do not allow implicitly converting CuPy arrays to NumPy (object) arrays.

tests/cupy_tests/lib_tests/test_polynomial.py
leofang:
```suggestion
```

leofang:
Q: Are there any lightweight perf regression tests that we can run to confirm we don't add overhead due to this change when UMP is not in use (which the majority of CuPy use cases)?

cupy/_core/core.pyx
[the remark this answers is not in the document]
leofang: @kmaehashi sorry I posted this in the wrong thread. I think I missed the second half of the story (sorry it's been a while since I worked on this feature...). `TypeError` is raised in `__array__`:
https://github.com/cupy/cupy/blob/678add80cdd82e89f2b751f1722035c53997ad03/cupy/_core/core.pyx#L1489-L1493
which can be easily reproduced by adding a dummy `def __array__(self): raise TypeError` to your snippet above. We do not allow implicitly converting CuPy arrays to NumPy (object) arrays.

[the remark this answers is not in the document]
kmaehashi: Ah, got it, thanks!

kmaehashi:
> Q: Are there any lightweight perf regression tests that we can run to confirm we don't add overhead due to this change when UMP is not in use (which the majority of CuPy use cases)?

I guess https://github.com/cupy/cupy-performance (work done by @emcastillo) can be used for that purpose?

kmaehashi (approved):

emcastillo:
Thanks!!

I will run the performance tests tomorrow to check if we have regressions! lets wait until its done for merging 😇

emcastillo:
main
```
time_add             - case        100:    CPU:     5.329 us   +/-  0.171 (min:     5.050 / max:     6.110) us     GPU-0:     7.850 us   +/-  0.405 (min:     6.816 / max:     8.416) us
time_add             - case        200:    CPU:     5.420 us   +/-  0.175 (min:     5.131 / max:     6.154) us     GPU-0:    15.181 us   +/-  0.516 (min:    14.336 / max:    16.192) us
time_add             - case        500:    CPU:     5.327 us   +/-  0.179 (min:     4.963 / max:     5.752) us     GPU-0:    61.522 us   +/-  0.471 (min:    60.320 / max:    63.040) us
time_add             - case       1000:    CPU:     5.445 us   +/-  0.256 (min:     5.082 / max:     6.513) us     GPU-0:   225.683 us   +/-  0.517 (min:   224.992 / max:   227.008) us
time_divide          - case        100:    CPU:     5.305 us   +/-  0.119 (min:     5.002 / max:     5.567) us     GPU-0:    13.009 us   +/-  0.530 (min:    12.288 / max:    13.984) us
time_divide          - case        200:    CPU:     5.391 us   +/-  0.212 (min:     4.903 / max:     6.474) us     GPU-0:    31.530 us   +/-  0.524 (min:    30.720 / max:    32.768) us
time_divide          - case        500:    CPU:     5.363 us   +/-  0.115 (min:     5.097 / max:     5.690) us     GPU-0:   158.026 us   +/-  0.478 (min:   157.504 / max:   159.392) us
time_divide          - case       1000:    CPU:     5.383 us   +/-  0.145 (min:     5.006 / max:     5.830) us     GPU-0:   608.937 us   +/-  0.526 (min:   608.256 / max:   610.080) us
time_divmod          - case        100:    CPU:     6.665 us   +/-  0.625 (min:     6.349 / max:    10.940) us     GPU-0:    15.572 us   +/-  0.865 (min:    14.464 / max:    20.000) us
time_divmod          - case        200:    CPU:     6.638 us   +/-  0.131 (min:     6.376 / max:     7.155) us     GPU-0:    39.015 us   +/-  0.342 (min:    38.624 / max:    40.352) us
time_divmod          - case        500:    CPU:     6.636 us   +/-  0.172 (min:     6.343 / max:     7.385) us     GPU-0:   199.698 us   +/-  0.602 (min:   198.656 / max:   200.544) us
time_divmod          - case       1000:    CPU:     6.546 us   +/-  0.283 (min:     6.192 / max:     7.851) us     GPU-0:   774.010 us   +/-  0.553 (min:   772.800 / max:   775.872) us
```

This PR
```
time_add             - case        100:    CPU:     6.097 us   +/-  2.922 (min:     5.311 / max:    26.124) us     GPU-0:     8.548 us   +/-  2.633 (min:     6.976 / max:    26.336) us
time_add             - case        200:    CPU:     5.566 us   +/-  0.135 (min:     5.325 / max:     6.150) us     GPU-0:    15.398 us   +/-  0.619 (min:    14.336 / max:    16.384) us
time_add             - case        500:    CPU:     5.578 us   +/-  0.123 (min:     5.311 / max:     5.881) us     GPU-0:    61.663 us   +/-  0.358 (min:    60.512 / max:    62.688) us
time_add             - case       1000:    CPU:     5.600 us   +/-  0.179 (min:     5.325 / max:     6.300) us     GPU-0:   225.651 us   +/-  0.501 (min:   225.152 / max:   227.136) us
time_divide          - case        100:    CPU:     5.573 us   +/-  0.590 (min:     5.285 / max:     9.615) us     GPU-0:    13.924 us   +/-  2.598 (min:    12.288 / max:    31.168) us
time_divide          - case        200:    CPU:     5.455 us   +/-  0.144 (min:     5.183 / max:     6.174) us     GPU-0:    31.781 us   +/-  0.558 (min:    30.720 / max:    32.768) us
time_divide          - case        500:    CPU:     5.497 us   +/-  0.164 (min:     5.180 / max:     6.370) us     GPU-0:   158.040 us   +/-  0.465 (min:   157.440 / max:   159.168) us
time_divide          - case       1000:    CPU:     5.472 us   +/-  0.168 (min:     5.221 / max:     6.019) us     GPU-0:   609.254 us   +/-  0.630 (min:   608.256 / max:   610.304) us
time_divmod          - case        100:    CPU:     6.782 us   +/-  0.188 (min:     6.550 / max:     7.526) us     GPU-0:    15.693 us   +/-  0.577 (min:    14.432 / max:    16.640) us
time_divmod          - case        200:    CPU:     7.320 us   +/-  3.447 (min:     6.380 / max:    31.090) us     GPU-0:    39.539 us   +/-  3.396 (min:    38.688 / max:    63.232) us
time_divmod          - case        500:    CPU:     6.844 us   +/-  0.167 (min:     6.587 / max:     7.422) us     GPU-0:   199.846 us   +/-  0.593 (min:   198.752 / max:   200.704) us
time_divmod          - case       1000:    CPU:     6.702 us   +/-  0.256 (min:     6.363 / max:     7.934) us     GPU-0:   774.319 us   +/-  0.461 (min:   773.888 / max:   776.064) us
```

There is some impact but is around ~0.2 us.

jakirkham (commented):
Adding a rough first pass of how we might use [NumPy's custom allocator framework]( https://numpy.org/doc/stable/reference/c-api/data_memory.html ) directly

cc @seberg (who likely will identify improvements on what I wrote)

emcastillo:
Yeah, we are being pinged a lot about this, so it would be great if we can merge it in the current step and work on optimizations in a follow up PR 😁

kmaehashi:
Thanks to all involved in the PR! Let me proceed to merge as discussed above.

emcastillo:
Thanks a lot to @rongou & @leofang who did most of the work :)!!

leofang:
Thousand thanks to @emcastillo @seberg @rongou @kmaehashi @jakirkham for pushing this through the finish line, glad to see this lands!

docs/source/user_guide/memory.rst
leofang: > everything here (after we fix a few minor things in a subsequent PR) will work

The needed effort is tracked in #8711.

leofang:
> One of the missing things that we can add later is, as you pointed out, a different `numpy_allocator` C APIs that calls `cudaMallocManaged` instead of `malloc` for NumPy to use.

This is tracked in #8711.

PR #8444, opened 2024-07-31
Fix the size of temporary CUB output space to consider its alignment

takagi:
This should be one of potential causes of illegal memory access we're facing at #8290.

ref:
- https://github.com/NVIDIA/cccl/blob/7a3dae72b36d2ce939de1d1b68e0b251d287074f/cub/cub/util_type.cuh#L667
- https://github.com/NVIDIA/cccl/blob/694e963e72c60384d85abdcf7affed504aa9ed04/thrust/thrust/complex.h#L76

### DIFF
--- a/cupy/cuda/cub.pyx
+++ b/cupy/cuda/cub.pyx
@@ -172,7 +172,8 @@ def device_reduce(_ndarray_base x, op, tuple reduce_axis, tuple out_axis,
         y = _core.ndarray((), x.dtype)
     else:  # argmin and argmax
         # cub::KeyValuePair has 1 int + 1 arbitrary type
-        kv_bytes = (4 + x.dtype.itemsize)
+        # Note that the key may be padded to make the value aligned.
+        kv_bytes = (x.dtype.alignment + x.dtype.itemsize)
         y = _core.ndarray((kv_bytes,), numpy.int8)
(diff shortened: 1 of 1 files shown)


asi1024 (dismissed):

takagi:
I've fixed the case that the size of input dtype is smaller than that of int.

leofang:
Hmmm... Maybe I missed something, but I am curious how this would help https://github.com/cupy/cupy/pull/8290? My understanding is: When we allocate the workspace, and when the mempool is in use (which is default), we actually up-allocate to the closets units of 512B ([here](https://github.com/cupy/cupy/blob/4b7b5ce6e8723b9ad07af64fa076f268f86fd1fc/cupy/cuda/memory.pyx#L1090)). It means we'd allocate more than whatever value we compute here (although it's still a good catch/fix, in case someone turns off the mempool).

kmaehashi:
I guess this one may not be related to #8290 as you suggested. As we are seeing odd `cudaErrorIllegalAddress` errors in #8290 CI (which looks irrelevant to the PR), we're currently running `compute-sanitizer --tool memcheck --padding 32` against all CuPy test code with memory pool disabled. This PR is one of the findings, and few more to come.

asi1024 (approved):

PR #8319, opened 2024-05-02
Support ROCm 6

Azusachan:
ROCm 6 introduced various changes on its API. In particular,
* Removal of gcnarch from hipDeviceProp_t structure
* Renaming of ‘memoryType’ in hipPointerAttribute_t structure to ‘type’

This patch provides support on ROCm 6.0.0 and above.

Issue #4610, opened 2021-02-03
Support Cython 3.0


kmaehashi:
Cython 3.0 is in alpha releases and we should get ready for it.
https://pypi.org/project/Cython/#history

```[tasklist]
### Tasks
- [ ] https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html
- [ ] Remove `IF`s and `DEF`s: https://github.com/cython/cython/issues/4310
```

kmaehashi:
https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html

leofang:
Some module-level `cpdef` variables will need to be fixed for supporting Cython 3. I checked with an upstream developer that there is no good way to support `cpdef` variables at the module level (cython/cython#3959).
```
00:05:42.478428 STDOUT 2075]	    warning: cupy_backends/cuda/api/runtime.pxd:829:6: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables	
00:05:42.478439 STDOUT 2075]	    warning: cupy_backends/cuda/api/runtime.pxd:830:6: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables	
00:05:42.478474 STDOUT 2075]	    warning: cupy_backends/cuda/api/runtime.pxd:831:6: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables
00:05:45.147461 STDOUT 2075]	    warning: cupy/core/internal.pyx:190:10: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables	
00:05:45.147485 STDOUT 2075]	    warning: cupy/core/internal.pyx:190:10: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables	
00:05:45.147845 STDOUT 2075]	    warning: cupy/core/internal.pyx:190:10: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables	
00:05:45.147902 STDOUT 2075]	    warning: cupy/core/internal.pyx:191:10: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables	
00:05:45.147914 STDOUT 2075]	    warning: cupy/core/internal.pyx:191:10: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables
[...]

PR #8457, opened 2024-08-06
Port to Cython 3.0

littlewu2508:
These should be the necessary changes for Cython-3:

1. Resolve some compilation errors
2. Enable a backward compatible mode of Cython: `special binary operator` set to True.

I applied these 2 patches on top of https://github.com/cupy/cupy/pull/8319, and built cupy-13.2.0 on ROCm 6. My cupy application run without errors.

I also observed one deprecation warning regarding Cython, which is currently not a blocker but may become one in the future:

> The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310


leofang:
Thanks for your help!
> I also observed one deprecation warning regarding Cython, which is currently not a blocker but may become one in the future:
> 
> > The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See [cython/cython#4310](https://github.com/cython/cython/issues/4310)

My suggestion is to use the compiler flag added in https://github.com/cython/cython/pull/6243 to silence the build-time warning for the time being. IIRC (though I can't find the thread where this was discussed...) we wanted to investigate some lightweight codegen to address this, but I dunno if it'd be prioritized any time soon. The deprecation of `IF`/`DEF` was well aware and tracked in #4610.

leofang:
> My suggestion is to use the compiler flag added in [cython/cython#6243](https://github.com/cython/cython/pull/6243)

It seems the flag has not made it to the 3.0.x releases...

EarlMilktea:
MEMO: Encountering many performance warnings like this:

> performance hint: cupy_backends/cuda/api/driver.pyx:122:32: Exception check after calling 'cuCtxGetCurrent' will always require the GIL to be acquired. Declare 'cuCtxGetCurrent' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.

```cython
cpdef intptr_t ctxGetCurrent() except? 0:
    initialize()
    cdef Context ctx
    with nogil:
        status = cuCtxGetCurrent(&ctx) # <- HERE!
    check_status(status)
    return <intptr_t>ctx
```

kmaehashi:
> Exception check after calling 'cuCtxGetCurrent' will always require the GIL to be acquired. Declare 'cuCtxGetCurrent' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.

It seems invocations of function pointers (assigned to variables) are emitting this warning. I'm curious if anything has changed in Cython 3.0, or there was a slient overhead even in Cython 0.29...

EarlMilktea:
Fortunately it seems that we can easily resolve most of the warnings by just replacing `nogil` by `noexcept nogil` in

- `cupy_backends/cuda/libs/_cnvrtc.pxi`
- `cupy_backends/cuda/api/_driver_extern.pxi`.

Of course we should check this replacement is valid, but I feel it's OK as these files just wrap the low-level CUDA APIs, which are also available in C codes...

leofang:
FYI @kmaehashi we noticed (https://github.com/conda-forge/cupy-feedstock/pull/295#issuecomment-2564498963) that this PR blocks the Python 3.13 support (https://github.com/cupy/cupy/issues/8651) even for the regular (with-GIL) builds, because Cython 0.29.x simply has no wheel or conda package released for 3.13.

kmaehashi:
Hi @littlewu2508, sorry for taking time to review this one. It seems several Cython warnings are emitted (including ones reported by @EarlMilktea), so I'm now taking a deeper look.

As for the linkage issue you mentioned:
> Cython didn't implement the cdef public externc, so I will try just use extern "C++" in cupy_thrust.cu

Cython allows control the linkage by `CYTHON_EXTERN_C` macro (as [documented](https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#c-public-declarations)), so I'll keep the C linkage (a9721826006f56d800944d7fb56c45b6ed385412). This is also required for `cupy._core.numpy_allocator` module introduced in #8442.

kmaehashi:
> Fortunately it seems that we can easily resolve most of the warnings by just replacing `nogil` by `noexcept nogil` in
> 
> * `cupy_backends/cuda/libs/_cnvrtc.pxi`
> * `cupy_backends/cuda/api/_driver_extern.pxi`
> 
> (Check [EarlMilktea@3c1a3bd](https://github.com/EarlMilktea/cupy/commit/3c1a3bdfb9ef7d680f5dee2f1ecee0609136eb63) for more details).
> 
> Of course we should check this replacement is valid, but I feel it's OK as these files just wrap the low-level CUDA APIs, which are also available in C codes...

Thanks for the investigation @EarlMilktea! I've just borrowed your commit 😄

kmaehashi:
Here is a quote of the `python ../cupy-performance/prof.py --plot ../cupy-performance/benchmarks/bench_ufunc_cupy.py` result. There seems additional 5-10% CPU overhead :thinking:


### commit 8c411811d04f5fc01ca9cc9a683b09b4f5c66196 + Cython 0.29.37:

```
time_add             - case        100:    CPU:    18.558 us   +/-  0.672 (min:    18.086 / max:    22.357) us     GPU-0:    22.262 us   +/-  0.999 (min:    21.504 / max:    26.624) us
time_add             - case        200:    CPU:    19.059 us   +/-  0.809 (min:    18.237 / max:    22.413) us     GPU-0:    22.855 us   +/-  1.033 (min:    21.504 / max:    26.624) us
time_add             - case       5000:    CPU:    18.831 us   +/-  0.822 (min:    18.114 / max:    23.521) us     GPU-0:   619.763 us   +/-  2.899 (min:   615.424 / max:   628.736) us
time_add             - case      10000:    CPU:    19.864 us   +/-  0.698 (min:    18.973 / max:    23.287) us     GPU-0:  2396.750 us   +/-  5.202 (min:  2386.944 / max:  2408.448) us
```


### commit 8c411811d04f5fc01ca9cc9a683b09b4f5c66196 + 6db06e4b68219a5c80c1bf2b1d86ab7ff60f8608 (this PR) + Cython 3.0.12:

```
time_add             - case        100:    CPU:    20.164 us   +/-  0.266 (min:    19.746 / max:    21.094) us     GPU-0:    23.799 us   +/-  0.601 (min:    22.528 / max:    25.600) us
time_add             - case        200:    CPU:    21.090 us   +/-  2.221 (min:    19.798 / max:    31.464) us     GPU-0:    24.863 us   +/-  2.412 (min:    22.528 / max:    35.840) us
time_add             - case       5000:    CPU:    20.638 us   +/-  1.296 (min:    19.804 / max:    27.575) us     GPU-0:   623.083 us   +/-  3.222 (min:   614.400 / max:   629.760) us
time_add             - case      10000:    CPU:    21.175 us   +/-  0.769 (min:    20.165 / max:    25.300) us     GPU-0:  2391.365 us   +/-  5.118 (min:  2379.776 / max:  2404.352) us
```

kmaehashi:
Tests passed, merging! Thanks again, @litlewu2508 and @EarlMilktea!

kmaehashi:
I'll take care of backport after merging #8951

jakirkham:
Wow! This is fantastic 🤩

Thank you all for your hard work on Cython 3 compatibility 👏

PR #8461, opened 2024-08-07
Decrease number of threads to avoid "system error: excessive memory usage is detected"

kmaehashi:
It seems the memory usage has increased during the build phase.

e.g. Example CI in main branch failing with `[ABORTED] excessive memory usage is detected`:
https://ci.preferred.jp/cupy.linux.cuda-example/166617/

### DIFF
--- a/.pfnci/linux/tests/actions/build.sh
+++ b/.pfnci/linux/tests/actions/build.sh
@@ -2,4 +2,4 @@
 
 set -uex
 
-time CUPY_NUM_NVCC_THREADS=8 CUPY_NUM_BUILD_JOBS=8 python3 -m pip install --user -v ".[test]"
+time CUPY_NUM_NVCC_THREADS=5 CUPY_NUM_BUILD_JOBS=8 python3 -m pip install --user -v ".[test]"


leofang:
So it's not about exceeding memory limit, but about using too much system resource (number of threads)? That might explain the observation in #8412 too. 

NVRTC and NVCC have been improved to allow internal multithreading compilation, ex:
https://docs.nvidia.com/cuda/archive/12.3.2/cuda-toolkit-release-notes/index.html#cuda-compilers-new-features
though I was under the impression that it's an opt-in feature and we haven't requested it...

leofang:
@kmaehashi this seems really promising! All Linux CIs that previously failed have passed now. How did you know this is the culprit?

leofang (approved):

kmaehashi:
Just a wild guess 😅 
NVCC emits `max($CUPY_NUM_NVCC_THREADS, number_of_architectures)` processes to build for each architecture in parallel, so I assumed each process's memory consumption has increased.

leofang:
> Just a wild guess 😅

Genius idea!

> NVCC emits `min($CUPY_NUM_NVCC_THREADS, number_of_architectures)` processes to build for each architecture in parallel, so I assumed each process's memory consumption has increased.

I see, so it's still about memory consumption, and this happens only at the build stage not at run time. Seems like a behavior that has the same root cause (OOM) as core dumps in Azure CI: https://github.com/conda-forge/cupy-feedstock/pull/208#issuecomment-1581027447 There, when we increased parallelism the build also failed. But instead of core dumps, pfnCI was able to get it through and raise OOM error at the end (which arguably makes it harder to debug...).

leofang:
In any case, glad to know we can rule out concerns on the memory consumption of JIT compilation.

takagi (approved):
LGTM!

jakirkham:
FWIW saw the same issue in conda-forge: https://github.com/conda-forge/cupy-feedstock/pull/272#discussion_r1722536824

Unfortunately the error was less clear as CI simply died without a clear message

We had to set `CUPY_NUM_BUILD_JOBS=1` in several jobs to resolve the issue

PR #8467, opened 2024-08-08
Avoid using Jitify everywhere inside CuPy

leofang:
Follow-up of #8412. This PR addresses the first item from https://github.com/cupy/cupy/issues/8209#issuecomment-2272753019:
> 1. There are a few places where the similar treatment can be done to avoid using Jitify

This is not critical to be merged/backported if time is short, just a nice to have.


asi1024 (approved):

asi1024:
LGTM! Thanks!

Issue #8452, opened 2024-08-05
Better subsetting of CSR and CSC matrices of minor axis.


Intron7:
### Description

Currently if you subset a `CSR` matrix over the minor axis, the matrix get internally converted to `CSC` to subset and than converted back to `CSR`. This very memory and time intensive and I ran into a lot of out of memory errors because of this.

In my very limited testing using a function that is based around `eliminate_zeros` might work best. This approach would also work for a sorting subset.

### Additional Information

```
from cuml.common.kernel_utils import cuda_kernel_factory
import math
_inplace_subset_gene = r"""
(int *indptr, int * indices, {0} *data,
                    int nrows, bool * mask, int* new_idx) {
        int row = blockDim.x * blockIdx.x + threadIdx.x;

        if(row >= nrows)
            return;

        int start_idx = indptr[row];
        int stop_idx = indptr[row+1];
        for(int i = start_idx; i < stop_idx; i++){
            int gene_indx = indices[i];
            indices[i] =new_idx[gene_indx];
            if(mask[gene_indx]){
                data[i] =0;
            }
  
        }
    }
"""

def _mul_csr(dtype):
    return cuda_kernel_factory(_inplace_subset_gene, (dtype,), "_inplace_subset_gene")
X = adata.X.copy() #copy the matrix
mask = cp.array(adata.var["n_cells_by_counts"] < 10) #get the mask
new_idx = cp.cumsum(~mask).astype(cp.int32)-1
mul_kernel = _mul_csr(X.dtype)
mul_kernel(
    (math.ceil(X.shape[0] / 32),),
    (32,),
    (X.indptr,X.indices, X.data, X.shape[0], mask,new_idx),
)
out = sparse.csr_matrix((X.data,X.indices,X.indptr),shape=(X.shape[0],X.shape[1]-mask.sum()))
out.eliminate_zeros()
out.shape
```

PR #8468, opened 2024-08-08
Add short cut for subsetting along the minor axis

Intron7:
This address #8452 

I added a shortcut for sorted `idx` for the subsetting or when the `X.indices` are not sorted to begin with.


cupyx/scipy/sparse/_compressed.py
```diff
@@ -416,17 +416,24 @@ def _major_index_fancy(self, idx):
...
         if self.nnz == 0 or N == 0:
             return self.__class__(new_shape, dtype=self.dtype)
-
         if idx.size * M < self.nnz:
             # TODO (asi1024): Implement faster algorithm.
             pass
 
-        return self._tocsx()._major_index_fancy(idx)._tocsx()
+        idx_is_sorted = cupy.all(idx[:-1] <= idx[1:])
+        if not self.has_sorted_indices or idx_is_sorted:
```
asi1024, line 429: `idx_is_sorted` is a `cupy.ndarray`, so it requires DtoH synchronization. Aligning with CuPy's policy, we provide features as possible while minimizing the amount of DtoH synchronization.
How about sorting `idx` without checking it is already sorted?

Intron7: I feel like sorting the index would create problems for the user. Because some people relay on sub-setting to sort there array.

asi1024: How about copying `idx`?

Intron7: I'm sorry, but I don't understand why copying would help with the issue. I added this:
```
        idx_is_sorted = cupy.all(idx[:-1] <= idx[1:])
        if not self.has_sorted_indices or idx_is_sorted:
```
because I want the user to get sorted `X.indices` if `X.has_sorted_indices`. If we don't care about that, we could leave this out and just run with the shortcut in every case. However, calling `X.sort_indices()` is more expensive than the old solution (which still has one DtoH that I save).

Intron7:
Benchmarks on an RTX 3090:

X.shape = (200,000, 27,998)
X_new.shape = (200,000, 19,214)
Old implementation = 500 ms
New implementation = 120 ms

Additionally, the shortcut is significantly more memory efficient. With 400,000 rows, the old implementation results in a memory error, while the new version completes without an issue.

cupyx/scipy/sparse/_compressed.py
asi1024: Hmm, I would prefer not to perform additional computation on the GPU to decide whether to use a shortcut path or not. Could you modify to allow shortcuts only to the extent that they can be branched by simple calculations on the host side?

Intron7:
@leofang as mentioned, this PR is important for the single-cell community as the old implementation is a significant bottleneck and responsible for many oom errors due to CSR to CSC conversion.

cupyx/scipy/sparse/_compressed.py
Intron7: Let me clarify the intent and functionality of the current implementation because I feel like we are misaligned:

1. Unsorted Indices and Canonical Format:
My current implementation explicitly supports subsetting with an unsorted index array. In other words, it correctly returns columns exactly in the order specified by the user, even if the provided indices are unsorted. However, in such cases, the resulting CSX matrix will no longer be in canonical format (sorted column indices for CSR).

2. Why Sorting Doesn't Always Apply:
Sorting the index will still not help with the creation of a canonical CSX matrix. Even an unsorted index results in the correct order of columns in a CSR matrix just not in canonical format. I think if a user subsets a Canonical matrix they want a canonical matrix out.

3. Reason for the Existing Check:
The only reason the current check (idx_is_sorted) exists is to preserve performance in cases where users subset from a canonical CSR matrix using an  unsorted index. The old implementation is faster than using my implementation and than calling sort_indices() afterwards.

leofang: I've been thinking about this, and what Severin said makes sense to me. @asi1024 could you share any implementation opportunity that you see in achieving what you suggested (simple shortcut based only on host logics)?

Intron7:
@leofang @asi1024 

After extensive testing and reviewing SciPy’s implementation, I found that my earlier version failed to handle duplicate indices such as [5, 5, 3, 2, 1, 1]. SciPy simply slices the axis without enforcing canonical CSR order, so unsorted input indices produce unsorted output rows. I have now re-implemented SciPy’s approach for CuPy with custom kernels that correctly support duplicates, eliminate the costly post-sort, run significantly faster, and use less memory. Let me know whether you’d prefer me to update the existing pull request with this implementation or create a new one.

leofang:
If it is still in-scope for the original intent, feel free to just reuse the same PR. TBH either way is fine.

Intron7:
@leofang @asi1024 Thank you for your patience with this one. All `cupyx.scipy.sparse` tests are now passing on my system, and I’ve tried to keep the kernel formatting as close as possible to cupy.

leofang:
@asi1024 would it be possible for you to take another look? All tests have passed. Thanks! 🙂

cupyx/scipy/sparse/_compressed.py
```diff
@@ -413,20 +413,232 @@ def _major_index_fancy(self, idx):
...
+            ker_name = 'fill_B_complex<{}>'.format(
+                _scalar.get_typename(self.data.real.dtype),
+            )
+            fillB = self._fill_B_complex.get_function(ker_name)
+        else:
+            ker_name = 'fill_B<{}>'.format(
+                _scalar.get_typename(self.data.dtype),
+            )
+            fillB = self._fill_B.get_function(ker_name)
+        threads = 32
```
asi1024, line 622: Different thread counts are specified for `calc_Bp_minor` and `fillB`. Is this intentional?

Intron7: Yes that is intentional based on how `_fill_B`& `_fill_B_complex` use the shared row to update the nnz per row.

asi1024:
@Intron7 @leofang LGTM! Thank you for your contributions!

Issue #7654, opened 2023-06-21
Remove use of `pkg_resources`


kmaehashi:
Replace it with `importlib.metadata` after dropping support for  Python 3.7.
https://docs.python.org/ja/3/library/importlib.metadata.html#module-importlib.metadata

kmaehashi:
Moved milestone from v13 to v14 and changed priority from high to med. The only remaining usage of `pkg_resources` in CuPy is `cupy/testing/_helper.py`, and is only imported within unit tests.

jakirkham:
Was looking at this recently. If I have time, may try a PR

jakirkham:
Should be fixed by PR: https://github.com/cupy/cupy/pull/8480

PR #8480, opened 2024-08-13
Get rid of `pkg_resources`

### LINKED ISSUE #8479: Get rid of `pkg_resources`
Replace it with modern alternatives (ex. `importlib`).

EarlMilktea:
Resolves #8479 and #7654.

### DIFF
--- a/cupy/testing/_helper.py
+++ b/cupy/testing/_helper.py
@@ -60,7 +60,10 @@ def installed(*specifiers: str) -> bool:
 
     for spec in specifiers:
         req = Requirement(spec)
-        found = metadata.version(req.name)
+        try:
+            found = metadata.version(req.name)
+        except metadata.PackageNotFoundError:
+            return False
         expected = req.specifier
         # If no constrait is given, skip
         if expected and (not expected.contains(found, prereleases=True)):
--- a/cupy/testing/_helper.py
+++ b/cupy/testing/_helper.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
 import contextlib
+import importlib.metadata
 import inspect
 import unittest
 import warnings
 from collections.abc import Callable
-from importlib import metadata
 from importlib.metadata import PackageNotFoundError
 from unittest import mock
 
@@ -62,7 +62,7 @@ def installed(*specifiers: str) -> bool:
     for spec in specifiers:
         req = Requirement(spec)
         try:
-            found = metadata.version(req.name)
+            found = importlib.metadata.version(req.name)
         except PackageNotFoundError:
             return False
         expected = req.specifier
--- a/cupy/testing/_helper.py
+++ b/cupy/testing/_helper.py
@@ -9,7 +9,6 @@ from importlib import metadata
 from unittest import mock
 
 import numpy
-from packaging.requirements import Requirement
 
 import cupy
 import cupyx
@@ -56,6 +55,9 @@ def installed(*specifiers: str) -> bool:
     Args:
         specifiers: Version specifiers (e.g., `numpy>=1.20.0`).
     """
+    # Make `packaging` a soft requirement
+    from packaging.requirements import Requirement
+
     for spec in specifiers:
         req = Requirement(spec)
         found = metadata.version(req.name)
--- a/cupy/testing/_helper.py
+++ b/cupy/testing/_helper.py
@@ -6,6 +6,7 @@ import unittest
 import warnings
 from collections.abc import Callable
 from importlib import metadata
+from importlib.metadata import PackageNotFoundError
 from unittest import mock
 
 import numpy
@@ -62,7 +63,7 @@ def installed(*specifiers: str) -> bool:
         req = Requirement(spec)
         try:
             found = metadata.version(req.name)
-        except metadata.PackageNotFoundError:
+        except PackageNotFoundError:
             return False
         expected = req.specifier
         # If no constrait is given, skip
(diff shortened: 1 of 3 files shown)


cupy/testing/_helper.py
```diff
@@ -1,37 +1,42 @@
...
-from typing import Callable
 import unittest
-from unittest import mock
 import warnings
+from collections.abc import Callable
+from importlib import metadata
+from unittest import mock
 
 import numpy
+from packaging.requirements import Requirement
```
kmaehashi, line 12: [`cupy.testing`](https://docs.cupy.dev/en/latest/reference/testing.html) is a public API (counter part of [numpy.testing](https://numpy.org/doc/stable/reference/routines.testing.html), and would like to avoid introducing new dependency here. How about importing `packaging` within `installed`?

jakirkham:
Could you please add issue ( https://github.com/cupy/cupy/issues/7654 ) to the fixed issues in the top-level post of this PR?

cupy/testing/_helper.py
jakirkham:
```suggestion
```

cupy/testing/_helper.py
```diff
@@ -40,36 +45,34 @@ def with_requires(*requirements):
...
     package requirement.
 
     Args:
         specifiers: Version specifiers (e.g., `numpy>=1.20.0`).
     """
-    # Delay import of pkg_resources because it is excruciatingly slow.
-    # See https://github.com/pypa/setuptools/issues/510
-    import pkg_resources
-
     for spec in specifiers:
```
jakirkham, line 59:
```suggestion
    # Make `packaging` a soft requirement
    from packaging.requirements import Requirement

    for spec in specifiers:
```

setup.py
```diff
@@ -45,6 +45,7 @@
     'test': [
         # 4.2 <= pytest < 6.2 is slow collecting tests and times out on CI.
         # pytest < 7.2 has some different behavior that makes our CI fail
+        'packaging',
```
jakirkham, line 48: Think it makes sense to keep `packaging` here if it is needed to run the tests (this doesn't affect runtime dependencies)

jakirkham (commented):
Perhaps something like this would work?

jakirkham:
Sorry was bit by a `pkg_resources` issue recently. So am eager to help move this fix forward 😅

jakirkham:
Seeing [a few errors on CI]( https://ci.preferred.jp/cupy.linux.cuda118/167503/#L14427 )

```python
/opt/pyenv/versions/3.11.0/lib/python3.11/importlib/metadata/__init__.py:561: in from_name	
    raise PackageNotFoundError(name)	
E   importlib.metadata.PackageNotFoundError: No package metadata was found for opencv-python
```

```python
/opt/pyenv/versions/3.11.0/lib/python3.11/importlib/metadata/__init__.py:561: in from_name	
    raise PackageNotFoundError(name)	
E   importlib.metadata.PackageNotFoundError: No package metadata was found for matplotlib
```

Maybe we need to tweak the logic in the package check somehow?

EarlMilktea:
@jakirkham 

I submitted a fix in 9efd74d9244b50ecced79a2c1c3370dbde344a85.
Now `PackageNotFoundError` results in `return False` instead of abort.

cupy/testing/_helper.py
```diff
@@ -1,37 +1,41 @@
+from __future__ import annotations
+
 import contextlib
 import inspect
-from typing import Callable
 import unittest
-from unittest import mock
 import warnings
+from collections.abc import Callable
+from importlib import metadata
```
jakirkham, line 8:
```suggestion
from importlib.metadata import PackageNotFoundError, version
```

cupy/testing/_helper.py
```diff
@@ -40,36 +44,40 @@ def with_requires(*requirements):
...
+    # Make `packaging` a soft requirement
+    from packaging.requirements import Requirement
 
     for spec in specifiers:
+        req = Requirement(spec)
         try:
-            pkg_resources.require(spec)
-        except pkg_resources.ResolutionError:
+            found = metadata.version(req.name)
+        except metadata.PackageNotFoundError:
```
jakirkham, line 65:
```suggestion
            found = version(req.name)
        except PackageNotFoundError:
```

jakirkham (commented):
Thank you! 🙏

That seems like a good fix

Had a minor suggestion on `import`s

cupy/testing/_helper.py
EarlMilktea: I'd like to use `metadata.version` since the word `version` is too generic and prone to name conflict.

jakirkham: That seems reasonable

Was mainly scratching my head thinking, "What is `metadata`? Have we defined it?" This despite knowing we were using `importlib.metadata` and having used it myself periodically

So was debating whether we go the verbose route and use `importlib.metadata` in a few places or `import` names

That said, no strong feelings here. Happy to go with whatever consensus is on this point

EarlMilktea: I see, then how about using the fully-qualified name: `importlib.metadata.version` ?

jakirkham: SGTM

EarlMilktea: Done.

jakirkham (approved):

jakirkham:
Thanks for all of your work here! 🙏

Looks like it is now passing 🥳

Let's see what Kenichi thinks 🙂

kmaehashi:
Thanks @EarlMilktea and @jakirkham, LGTM!

kmaehashi (approved):

Issue #8477, opened 2024-08-12
cupy.random.choice uses N^2 temporary storage?


truher:
### Description

```
cp.random.choice(10000, size=10000, p=cp.ones(10000)/10000)
```
produces

```
cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 800,000,000 bytes (allocated so far: 160,080,384 bytes).
```

this looks to me like random.choice() wants 8*n^2 bytes for n items.  am i reading that right?

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
OS                           : Linux-6.5.0-45-generic-x86_64-with-glibc2.35
Python Version               : 3.10.12
CuPy Version                 : 13.2.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 1.26.2
SciPy Version                : 1.10.1
Cython Build Version         : 0.29.36
Cython Runtime Version       : 3.0.6
CUDA Root                    : /usr/local/cuda
nvcc PATH                    : /usr/local/cuda/bin/nvcc
CUDA Build Version           : 12040
CUDA Driver Version          : 12050
CUDA Runtime Version         : 12040 (linked to CuPy) / 12060 (locally installed)
cuBLAS Version               : (available)
cuFFT Version                : 11206
cuRAND Version               : 10307
cuSOLVER Version             : (11, 6, 4)
cuSPARSE Version             : (available)
NVRTC Version                : (12, 6)
Thrust Version               : 200200
CUB Build Version            : 200200
Jitify Build Version         : <unknown>
cuDNN Build Version          : 8801
cuDNN Version                : 8907
NCCL Build Version           : (not loaded; try `import cupy.cuda.nccl` first)
NCCL Runtime Version         : (not loaded; try `import cupy.cuda.nccl` first)
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA GeForce GTX 950
Device 0 Compute Capability  : 52
Device 0 PCI Bus ID          : 0000:01:00.0

```

### Additional Information

```
>>> mempool.total_bytes()  
160080384
```

asi1024:
@truher Thanks for your reporting! This issue will be resolved with https://github.com/cupy/cupy/pull/8483.

```py
>>> import cupy as cp
>>> cp.random.choice(10000, size=10000, p=cp.ones(10000)/10000)
array([7274, 9171, 7195, ..., 5042, 3886, 4395])
[...]

PR #8483, opened 2024-08-13
Improving performance of `cupy.random.choice`

asi1024:
Fix https://github.com/cupy/cupy/issues/8477.

current master:

```
CPU:   528.909 us   +/-  8.817 (min:   513.830 / max:   556.690) us     GPU-0: 22175.946 us   +/- 21.370 (min: 22132.736 / max: 22215.679) us
```

This PR:

```
CPU:   542.260 us   +/- 226.564 (min:   510.556 / max:  2789.806) us     GPU-0:   548.843 us   +/- 227.403 (min:   517.088 / max:  2804.736) u
```

Reproducer:

```py
import cupy as cp

def f():
    cp.random.choice(10000, size=10000, p=cp.ones(10000)/10000)

print(benchmark(f, (), n_repeat=20))
```

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -1136,12 +1136,11 @@ class RandomState(object):
             raise NotImplementedError
 
         if p is not None:
-            p = cupy.broadcast_to(p, (size, a_size))
-            index = cupy.argmax(cupy.log(p) +
-                                self.gumbel(size=(size, a_size)),
-                                axis=1)
-            if not isinstance(shape, int):
-                index = cupy.reshape(index, shape)
+            # https://github.com/numpy/numpy/blob/v2.0.1/numpy/random/mtrand.pyx#L1013  # NOQA
+            cdf = p.cumsum()
+            cdf /= cdf[-1]
+            uniform_samples = self.random_sample(shape)
+            index = cdf.searchsorted(uniform_samples, side='right')
         else:
             if a_size == 0:  # TODO: (#4511) Fix `randint` instead
                 a_size = 1


kmaehashi (approved):

seberg:
This PR changes the stream of random numbers (noticed in a cuML doctest, I suspect that isn't a huge deal, once can set up the doctests to only run on newer cupy presumably).

I am not sure if CuPy has much of a policy around stream stability or if you may want to document the change.

Old stream:
```
In [5]: cp.random.RandomState(cp.uint32(10)).choice([0, 1], size=20, p=[0.5, 0.5])
Out[5]: array([1, 0, 1, 0, 1, 1, 1, 1, 0, 1, 0, 1, 0, 0, 0, 1, 1, 1, 0, 0])
```
New stream:
```
In [2]: cp.random.RandomState(cp.uint32(10)).choice([0, 1], size=20, p=[0.5, 0.5])
Out[2]: array([0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 0, 0, 0, 1, 1, 1, 1, 1, 0, 1])
```

jakirkham:
Curious what the result is if we drop `p` (given there is equal probability of either option)?

kmaehashi:
Thanks for the catch! There's no explicit policy, but I think in general users would expect reproducibility between minor versions with fixed seeds, so I'll make a revert PR for v13 backport. cc/ @asi1024

jakirkham:
Sounds good. Thanks all! 🙏

Given this code is more like NumPy (and it is more performant), think that is what we want longer term. We probably just need a little more time to understand the differences between the two

So reverting in v13 seems like the right approach

PR #8510, opened 2024-08-18
Support `half` option in `scipy.signal.minimum_phase`

asi1024:
Follow https://github.com/scipy/scipy/pull/19706

### DIFF
--- a/cupyx/scipy/signal/_fir_filter_design.py
+++ b/cupyx/scipy/signal/_fir_filter_design.py
@@ -783,7 +783,7 @@ def _dhtm(mag):
     return recon
 
 
-def minimum_phase(h, method='homomorphic', n_fft=None):
+def minimum_phase(h, method='homomorphic', n_fft=None, half=True):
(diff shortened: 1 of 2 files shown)


kmaehashi:
This needs to be backported to make results compatible with SciPy 1.14.
https://github.com/scipy/scipy/pull/19706
> It always bothered me that scipy.signal.minimum_phase gave a filter half the length with (roughly) half the effect on the magnitude spectrum. This PR adds a half=True backward-compatible kwarg to scipy.signal.minimum_phase that you can set to half=False to get a filter the same length as h with the same magnitude spectrum. This matches the reference case in Oppenheim and Shafer. Also fixes a small bug where previously we used len(h) rather than n_fft to decide where to truncate the cepstral filter, though the effects seem minimal.

jakirkham:
Thank you both! 🙏

Looks like this is passing 🥳

kmaehashi (approved):

PR #8511, opened 2024-08-19
Increase MAX_NDIM to 64

asi1024:
Part of https://github.com/cupy/cupy/issues/8306.
https://numpy.org/devdocs/release/2.0.0-notes.html


jakirkham:
Thanks Akifumi! 🙏

leofang:
(On my cell, sorry for brevity.) I noticed this PR will have `sizeof(CArray)>=1040` which is a bit enormous, since the default kernel buffer size is 4096 bytes. I have been discussing this internally before I went on leave, and we think it should be possible to use variable length array to work around this. I can help draft a PR after I return from leave if this is not urgent.

leofang:
It seems we missed to update one last test, I raised a PR here: https://github.com/asi1024/cupy/pull/4

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -144,11 +144,17 @@ def test_unsupported_type(self):
...
+    def test_upper_limit_ndim(self, xp):
+        shape = [1 for i in range(64)]
+        return xp.zeros(shape, dtype=xp.int8)
+
     def test_excessive_ndim(self):
         for xp in (numpy, cupy):
             with pytest.raises(ValueError):
                 xp.ndarray(
-                    shape=[1 for i in range(33)], dtype=xp.int8)
+                    shape=[1 for i in range(65)], dtype=xp.int8)
```
leofang, line 157: IIUC this test relies on NumPy 2.0+, perhaps we need `@testing.with_requires('numpy>=2.0')` here as well?

asi1024: Thanks! Fixed in 555e35a.

kmaehashi (approved):

asi1024:
@leofang Thank you for your help too!

Issue #8466, opened 2024-08-08
Recent fp16 headers (from CUDA 12.2+) depend on CUDA runtime headers


leofang:
Starting #8412 we bundled fp16 headers for CUDA 12.x minor releases, in order to address several JIT compilation issues. However, it sorta breaks the current deployment story (that a full CTK installation is not strictly required). For example on the main branch:
```python
>>> import cupy as cp
>>> 2 * cp.empty(10)
Traceback (most recent call last):
  File "/local/home/leof/dev/cupy_cuda125/cupy/cuda/compiler.py", line 679, in compile
    nvrtc.compileProgram(self.ptr, options)
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 125, in cupy_backends.cuda.libs.nvrtc.compileProgram
    cpdef compileProgram(intptr_t prog, options):
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 138, in cupy_backends.cuda.libs.nvrtc.compileProgram
    check_status(status)
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 53, in cupy_backends.cuda.libs.nvrtc.check_status
    raise NVRTCError(status)
cupy_backends.cuda.libs.nvrtc.NVRTCError: NVRTC_ERROR_COMPILATION (6)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy/_core/core.pyx", line 1286, in cupy._core.core._ndarray_base.__mul__
    return _math._multiply(x, y)
  File "cupy/_core/_kernel.pyx", line 1375, in cupy._core._kernel.ufunc.__call__
    kern = self._get_ufunc_kernel(dev_id, op, arginfos, has_where)
  File "cupy/_core/_kernel.pyx", line 1402, in cupy._core._kernel.ufunc._get_ufunc_kernel
    kern = _get_ufunc_kernel(
  File "cupy/_core/_kernel.pyx", line 1082, in cupy._core._kernel._get_ufunc_kernel
    return _get_simple_elementwise_kernel(
  File "cupy/_core/_kernel.pyx", line 94, in cupy._core._kernel._get_simple_elementwise_kernel
    return _get_simple_elementwise_kernel_from_code(name, code, options)
  File "cupy/_core/_kernel.pyx", line 82, in cupy._core._kernel._get_simple_elementwise_kernel_from_code
    module = compile_with_cache(code, options)
[...]

PR #8489, opened 2024-08-14
Search header files from CTK wheel

kmaehashi:
Closes #8466.

cc/ @leofang

PR #8519, opened 2024-08-19
Extend runtime header search logic to conda

jakirkham:
As CuPy also needs a bit of help searching for headers in the conda case, extend the logic used to search for wheel headers in PR ( https://github.com/cupy/cupy/pull/8489 ) to also search for conda headers. In particular this searches the targets directory for the platform where CuPy is being run on so that it can find headers there.

Fixes https://github.com/cupy/cupy/issues/8466

### DIFF
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -7,6 +7,7 @@ import importlib.metadata
 import json
 import os
 import os.path
+import platform
 import re
 import shutil
 import sys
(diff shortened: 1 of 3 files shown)


kmaehashi (approved):
Thanks John, LGTM!

jakirkham:
Thanks Kenichi! 🙏

PR #8525, opened 2024-08-20
Fix order 'K' with shape given for *_like array creation

takagi:
Fix allocation size on *_like array creation, which caused potentially illegal memory access. We had wrong allocation size when the order is K and the shape is different from that of the prototype array.

### DIFF
--- a/cupy/_creation/basic.py
+++ b/cupy/_creation/basic.py
@@ -1,3 +1,4 @@
+import math
 from typing import Any, Optional
 
 import numpy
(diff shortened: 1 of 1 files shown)


kmaehashi (approved):

kmaehashi:
Thanks! For record, here is the reproducer (minified version of the [test code](https://github.com/cupy/cupy/blob/3565c7a57e0a25a81c32a8bae67b0437850cd584/tests/cupy_tests/creation_tests/test_basic.py#L384))

```py
a = cupy.arange(24).reshape((2, 3, 4))
a = a[:, ::2, :].swapaxes(0, 1)  # a is now non-contiguous, shape (2, 2, 4)
b = cupy.empty_like(a, order='K', shape=(2, 100, 4))
b.fill(1)
```

I confirmed `compute-sanitizer --tool memcheck --padding 16 $(pyenv which python) test.py` no longer warns with this PR.

PR #8526, opened 2024-08-20
Fix `sparse.linalg` function signatures following SciPy 1.14

asi1024:

https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html#expired-deprecations


takagi (approved):
LGTM!

PR #8527, opened 2024-08-20
Fix `cupyx.scipy.sparse` tests for SciPy 1.14

asi1024:

- Use `.toarray()` instead of `.A`
- `sparse._base._spbase()` => `sparse._base._spbase(dummy)`
- `str(spmatrix)` is changed

https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html#scipy-sparse-improvements


takagi (approved):
LGTM!

Issue #6935, opened 2022-07-29
Support new kwargs `initial` and `where` in reduction functions


leofang:
### Description

It seems 
- `numpy.min()`
- `numpy.amin()`
- `numpy.max()`
- `numpy.amax()`
- ...?

have added additional kwargs `initial` and `where`, but CuPy's counterparts do not. Would be nice to catch up, so opening this issue to track.

ps. I haven't checked if this is a duplicate issue.

leofang:
Ah I guess this issue is a superset of #6326.

pri1311:
Ahh yes, this is finally here, I was struggling with #6396 for long and thought I was making some error!

ikrommyd:
I'm surprised this hasn't gotten more attention so far. I was trying to have some code that uses `max` work with `numpy`, `cupy` and `jax` at the same time and was surprised to find that in the cupy case, I need an extra ufunc call to `cupy.minimum` because `initial` is not present. I assume more people will be happy to have this so I'm wondering whether it would be possible to do at some point within the cupy 14 dev cycle.

seberg:
I'll milestone it for v15 (thinking mainly of min/max `initial=`).

Especially `initial` may be pretty straight forward for reductions implemented by CuPy (since CuPy doesn't implement weird "reductions" like `subtract`).
But even if the general turns out not straight-forward one can focus on min/max: min/max are special because the `initial` value becomes the identity of the operation/result. (I.e. the reason why min/max doesn't define an identity is that any value could be a reasonable identity depending on context.)

ikrommyd:
Yeah min/max is what I had in mind especially. Also extremely useful if you hit length 0 arrays. You need the identity in those cases.

PR #8533, opened 2024-08-20
Skip `special.logsumexp` test for empty input

asi1024:
We should add a support for empty inputs after reduction functions support the `initial` option: https://github.com/cupy/cupy/issues/6935


takagi (approved):
LGTM!

PR #8537, opened 2024-08-21
Add support for tensor D for cutensor(Mg)

jemiryguo:
In the cutensor lib, contraction is D = alpha * op(a) * op(b) + beta * op(c). It support the case the output array D is different with C, which is disabled in cupy wrapper. According to the cutensor documentation, currently modeD should be identical to modeC in both cutensor and cutensorMg, i change the logical to make this follows the cutensor documentation.


asi1024:
@jemiryguo Sorry for my response. Could you add tests in tests/cupyx_tests/test_cutensor.py?

PR #8538, opened 2024-08-21
Properly allocate in RNG when specified dtype is neither float32/float64

takagi:
Fix to properly allocate output arrays for some RNG methods when dtype is specified. This potentially had illegal memory access.

### DIFF
--- a/cupy/random/_generator_api.pyx
+++ b/cupy/random/_generator_api.pyx
@@ -152,6 +152,10 @@ class Generator:
         if out is not None:
             self._check_output_array(dtype, size, out)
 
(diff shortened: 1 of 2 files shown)


cupy/random/_generator_api.pyx
```diff
@@ -152,6 +152,10 @@ class Generator:
         if out is not None:
             self._check_output_array(dtype, size, out)
 
+        dtype_ = numpy.dtype(dtype)
```
kmaehashi, line 155: I think if desired `dtype` is neither float64/float32, it's better to generate using float32 and down cast to the dtype.

kmaehashi (approved):

PR #8539, opened 2024-08-21
Replace the cublas to 64-bit API

jemiryguo:
CUDA cublas lib currently support the 64-bit interface, which enable the very large linalg operations. But cupy use the 32-bit as cublas backend, which will raise a error `value is too large to convert to int`. This PR replace all cublas function into its 64-bits version if exist.


takagi:
Thanks for your work! Can I ask you to keep 32-bit API and add 64-bit ones instead of replacing? We have `cupy_backends` module as public and would like to keep 32-bit API on that level. Then, `cupy` module will turn to call its 64-bit API.

leofang:
My suggestions:
1. Let us not increase the binding surface (and therefore maintenance burden) in CuPy, now that we (NVIDIA) are aggressively covering them. In particular, all the 64-bit cuBLAS APIs that you need are offered in nvmath-python: https://docs.nvidia.com/cuda/nvmath-python/latest/bindings/cublas.html. Please give it a try and let us know if you encounter any issue.
2. Always calling the 64-bit APIs could incur a slight overhead (cuBLAS internally dispatches to the most performant kernel depending on the problem size), so for CuPy's APIs they may or may not be of concern. It's best to do some benchmarks before making a change.
3. 64-bit APIs are not provided until CUDA 12, so for CUDA 11.x users we still need to use the old APIs.

leofang:
> I have tried the nvmath-python, and found it meets all my requirements. Thanks a lot! BTW, is there a paln that cupy just based on `nvmath-python` in the future? The lack of support to 64-bit is still inconvient.

Glad to know it worked for you. I will create an issue to track the integration plan. cc: @samaid for vis

PR #8540, opened 2024-08-21
Added a GPU implemenation of scipy's lsq_linear BVLS (bounded value l…

craigwarner-ufastro:
…east

squares) which solves:
        minimize 0.5 * ||A x - b||**2
        subject to lb <= x <= ub

The scipy verison works on A with shape (m,n) and b with shape (m) to solve for x with shape (n).  This implementation vectorizes that to a 3-d A with shape (z, m, n) and b with shape (z, m) to solve for x with shape (z, n) - that is to simultaneously solve z independent matrices.

For z of a few hundred this can see a speed gain over scipy's BVLS on the CPU; for a couple thousand, it can be a significant speed gain.

Included is a python unittest tests/cupyx_tests/scipy_tests/optimize_tests/test_lsq_linear.py

As well as a stand-alone tester demonstrating the syntax: tests/cupyx_tests/scipy_tests/optimize_tests/standalone_tests/test_gpu_bvls.py


craigwarner-ufastro:
Hi @linshokaku @kmaehashi I was just wondering if there's been any progress in reviewing my addition and if there's anything additional you need from me to proceed.  Cheers!

Issue #8541, opened 2024-08-21
Conda build issue with CuPy `v13`


jakirkham:
Have been building Conda development versions of CuPy `v13` to aid testing efforts. This has been occurring in this PR: https://github.com/conda-forge/cupy-feedstock/pull/272

As of yesterday was able to build CuPy's `v13` branch using commit: https://github.com/cupy/cupy/commit/ea7c6470bfcf9a2f100f0ce21f6e0e897f344c37

However today ran into issues when building CuPy's `v13` using commit: https://github.com/cupy/cupy/commit/ea5c4e715602213bfcde3d78d7d0c8240db14884

Here is the diff between the two builds: https://github.com/cupy/cupy/compare/ea7c6470bfcf9a2f100f0ce21f6e0e897f344c37...ea5c4e715602213bfcde3d78d7d0c8240db14884

The recipe used for the builds was unchanged between the two occurrences. So think one of the changes may be the cause. Need to dig further to identify which one and how to address

For now am raising primarily for awareness

jakirkham:
Ok am starting to untangle this in PR: https://github.com/cupy/cupy/pull/8544

leofang:
xref: https://github.com/conda-forge/cupy-feedstock/issues/279#issuecomment-2303641454

jakirkham:
To summarize discussion elsewhere, this appears to be due to Distutils changes in Setuptools vendored copy that went out in 72.2.0. That package showed up when we did the rebuild yesterday in conda-forge, which is why we saw it then

PR https://github.com/cupy/cupy/pull/8544 adapts logic we have in CuPy for conda-forge builds for this newer Setuptools & Distutils

Should add the changes in Distutils make it easier to work with C++ compilers, which seems like a good change. It just means we need to adapt our code accordingly 

xref: https://github.com/cupy/cupy/pull/8544#discussion_r1726342921

PR #8544, opened 2024-08-21
Update conda-build CUDA detection logic for Setuptools 72.2.0

jakirkham:
Fixes https://github.com/cupy/cupy/issues/8541

# Background

Recently Leo added a patch to improve the Conda building experience in PR: https://github.com/cupy/cupy/pull/7603 (which is fantastic as this was [a large patch in the cupy-feedstock]( https://github.com/conda-forge/cupy-feedstock/blob/6db86965e825788262d22a5f3c8b9aa6b93f9d96/recipe/support_new_cuda_layout.diff ). Thanks Leo! 🙏)

This worked ok in our testing of CuPy 13.3.0 builds (not to mention previous CuPy builds where we applied this patch in conda-forge). However (as later identified after further investigation), this broke due to the recently released Setuptools [72.2.0]( https://github.com/pypa/setuptools/releases/tag/v72.2.0 ).

In particular the breakage resulted in the following error [on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1007325&view=logs&j=d1321064-f6c3-56b7-0172-6d994d01c836&t=39022612-2533-571e-7b41-94814e7d38bb&l=1339 ):

```python
command 'c++' failed: No such file or directory
```

On the one hand this is a true statement as conda-forge intentional uses [a triple naming scheme with compilers and associated tools]( https://github.com/conda-forge/ctng-compilers-feedstock/blob/8853ee7c0d15a9a73b35503c384f5894cf38c379/recipe/get_cpu_arch.sh#L18-L31 ) to include the architecture, OS type, and OS version. So technically `c++` does not exist. However what it points to is that a command in Distutils is not using the conda-forge compilers during the build, which is the real problem. Reading the CI log led to this code in CuPy as where the error arose

<hr>

# Aligning compilers with the Python interpreter build

Most of the time when an error like this shows up with `distutils` the answer is to ensure [`distutils.sysconfig`'s `customize_compiler`]( https://docs.python.org/3.11/distutils/apiref.html#distutils.sysconfig.customize_compiler ) is called on the new `CCompiler` object created by `distutils`. What this does (as the legacy `distutils` documentation in Python 3.11 linked above explains), is update the `distutils` `CCompiler` object based on settings used to build Python itself. IOW in conda-forge `python` is also built with conda-forge's compiler toolchain. So residual information from that build experience lives in the `python` package conda-forge produces. Thus `customize_compiler` then updates `CCompiler` to work with the conda-forge toolchain. Should add `customize_compiler` only does something in the UNIX case currently (it could potentially be updated to do something on Windows). This would be done by a change like that shown below. Normally this would be enough

```diff
     compiler = ccompiler.new_compiler()
+    distutils.sysconfig.customize_compiler(compiler)
```

<hr>

# Handling cross-compilers

However as the goal of the logic in this part of the CuPy code base is to build an artifact for the native platform and inspect it, that won't work with these compiler settings. Why is that? Well conda-forge uses a cross-compiling toolchain. Meaning the compilers are able to perform a build on a different platform than where the final package will be deployed. So building on Linux `x86_64` (plentiful in the world of CI currently) to create packages for Linux ARM (less available currently). IOW this is done to balance resource availability with user demand. So calling `customize_compiler` in conda-forge actually configures the compiler to target the platform where packages will be deployed (called `host_platform`). However for CuPy's use case here, we need to target `build_platform`, which may be different

As a result a manual patch is needed to configure the build tools for `build_platform`. The logic Leo added here did that based on [what `distutils` had in Python 3.11]( https://github.com/python/cpython/blob/795f2597a4be988e2bb19b69ff9958e981cb894e/Lib/distutils/unixccompiler.py#L49-L63 ). So what were we missing that resulted in new errors?

<hr>

# Setuptools, Distutils & the evolving Python packaging landscape

For a long time now, `setuptools` has monkeypatched `distutils`. In the past this meant there needed to be a specific `import` order for this to work. Namely `setuptools` first and then `distutils`. Though this caused some consternation for users and maintainers ( https://github.com/pypa/setuptools/issues/2259 ). So `setuptools` changed this to always prefer its own vendored `distutils` ( https://github.com/pypa/setuptools/pull/2305 ). Why does this matter? Well it means we never actually get `distutils` from the Python stdlib. We always get it from what `setuptools` vendored. This also explains why upstream removing its copy of `distutils` in Python 3.12 (please see [PEP 632]( https://peps.python.org/pep-0632 )) did not affect us (as we were not using it)

Though why do we care how `setuptools` supplies `distutils`. Well that [source code lives outside CPython]( https://github.com/pypa/distutils ) and we get updates whenever `setuptools` refreshes [its vendored `distutils`]( https://github.com/pypa/setuptools/tree/v72.0.0/setuptools/_distutils ). IOW just because the `distutils` source is updated doesn't mean it will be in `setuptools` yet. The vendored copy needs to be refreshed as well, which may happen out-of-sync relative to the `distutils` source. These pieces are important when trying to make sense of changes to `distutils` and how/when they will affect us

<hr>

# Distutils learns about C++ compilers

So what changed in `distutils`? Earlier this year `distutils` included updates for C++ compiler handling via PR ( https://github.com/pypa/distutils/pull/228 ). This expanded [the executables in `CCompiler`]( https://github.com/pypa/distutils/blob/fcef96367a0aef5bcf58027553ca36d50805210c/distutils/unixccompiler.py#L111-L129 ) by adding a C++ equivalent executable for every C associated one previously. This actually improves working with C++ compilers in `distutils`, which is pretty important for CUDA as well. So this is a helpful fix

<hr>

# Software supply chains and their logistics

Even though that `distutils` improvement went in at the beginning of August, we did not see it right away. This is because Setuptools refreshed `distutils` a week later with PR ( https://github.com/pypa/setuptools/pull/4539 ). The next Setuptools release, [72.2.0]( https://github.com/pypa/setuptools/releases/tag/v72.2.0 ), was a week after that (so 2 weeks after the `distutils` PR was merged). Then conda-forge waited a week to confirm there were not user errors showing up before shipping ( https://github.com/conda-forge/setuptools-feedstock/pull/355 ). So we only started seeing this error 3 weeks after the `distutils` change

<hr>

# Resolution

In any event, now that we have learned about these changes and understood the source, we can apply an appropriate patch. Namely...

* Configure the `distutils` `CCompiler` to match how Python is built
* Patch toolchain to target the build machine where it is run
* Make sure to cover the new Distutils C++ compilers

This PR includes these needed fixes for Setuptools 72.2.0


install/cupy_builder/cupy_setup_build.py
```diff
@@ -343,15 +343,23 @@ def make_extensions(ctx: Context, compiler, use_cython):
             # that are executable in the build environment, not the target
             # environment.
             compiler = ccompiler.new_compiler()
+            sysconfig.customize_compiler(compiler)
```
jakirkham, line 346: In Conda we use this to pick up compiler details from the Python build. Seems CuPy uses this in most places except here. Likely this is because of the heavy patching to perform tests with the build platform compiler instead of host compiler, which is otherwise used

However without this, we get errors like `c++` not found, which can be a bit cryptic

Now if we get an error it points to the C++ compiler used (like the host one), which we will more easily identify as wrong and fix

Also there can be other details that are useful to fill in for the compiler here

install/cupy_builder/cupy_setup_build.py
```diff
@@ -343,15 +343,23 @@ def make_extensions(ctx: Context, compiler, use_cython):
...
-            compiler.linker_exe = [cc, f'-B{os.environ["BUILD_PREFIX"]}/bin']
+            compiler.compiler_cxx = [cxx,]
+            compiler.compiler_so_cxx = [cxx,]
             compiler.linker_so = [cc, f'-B{os.environ["BUILD_PREFIX"]}/bin',
                                   '-shared']
-
+            compiler.linker_so_cxx = [cxx, f'-B{os.environ["BUILD_PREFIX"]}/bin',
+                                      '-shared']
+            compiler.linker_exe = [cc, f'-B{os.environ["BUILD_PREFIX"]}/bin']
+            compiler.linker_exe_cxx = [cxx, f'-B{os.environ["BUILD_PREFIX"]}/bin']
```
jakirkham, line 360: These were resorted to match [the order in Distutils]( https://github.com/pypa/distutils/blob/30b7331b07fbc404959cb37ac311afdfb90813be/distutils/unixccompiler.py#L117-L129 ), which should hopefully make it easier to compare and update when needed

kmaehashi:
Does conda build use sdist on PyPI, or `v13` branch of the repository? The internal PyPI build workflow (which prepares all the sdist/wheel packages) has already been completed, so I'm wondering if we need to rekick the workflow after merging this one. (asking because the workflow will take another 3-4 hours 😅)

jakirkham:
We typically use [the `sdist` from PyPI]( https://github.com/conda-forge/cupy-feedstock/blob/6db86965e825788262d22a5f3c8b9aa6b93f9d96/recipe/meta.yaml#L27-L28 )

We have been using the git repo for the development builds in PR ( https://github.com/conda-forge/cupy-feedstock/pull/272 ). Unfortunately the Windows builds have issues resolving symlinks between `git` repos, which causes the Windows builds to fail. Spent a fair amount of time debugging the Windows build from `git`, but had to give up. For development packages where most folks are on Linux (or can use Linux easily), this isn't an issue. However we would be stuck building on Windows for a release without it

Completely understand not wanting to regenerate the wheels (don't think we need that). Is it possible to just regenerate the `sdist`? It appears this file doesn't get installed. So the wheels would be unchanged by this patch (it only affects the Conda build process anyways)

jakirkham:
Sorry I think I was insufficiently caffeinated when I wrote that 😅

If there is an `sdist`, we can absolutely apply this as a patch on it

If it is at all possible to include this in the `sdist`, that would be preferable. If that is easy to do, great! If not, please let me know

Regardless it would be nice to get this merged and backported so that it is included in future releases at a minimum

Does that sound reasonable?

kmaehashi:
@jakirkham Could you apply this as a patch this time? The current build workflow has a strong assumption that sdist/wheels are built from the same commit. 😞 
I'll make sure this one gets merged after the release!

jakirkham:
Yes I can do that

install/cupy_builder/cupy_setup_build.py
```diff
@@ -336,21 +336,35 @@ def make_extensions(ctx: Context, compiler, use_cython):
...
+            cc = os.environ['CC_FOR_BUILD']
+            cxx = os.environ['CXX_FOR_BUILD']
+            ar = os.environ['BUILD'] + "-ar"
+            compiler.preprocessor = None
             compiler.compiler = [cc,]
-            compiler.compiler_cxx = [cxx,]
             compiler.compiler_so = [cc,]
-            compiler.linker_exe = [cc, f'-B{os.environ["BUILD_PREFIX"]}/bin']
+            compiler.compiler_cxx = [cxx,]
+            compiler.compiler_so_cxx = [cxx,]
```
jakirkham, line 357: @leofang please see this line specifically. Without this we get the error

```python
command 'c++' failed: No such file or directory
```

Previously Distutils was part of CPython as of 3.11. When it was, this was [not defined]( https://github.com/python/cpython/blob/795f2597a4be988e2bb19b69ff9958e981cb894e/Lib/distutils/unixccompiler.py#L55-L63 )

However Python 3.12 removed Distutils. Please see [PEP 632]( https://peps.python.org/pep-0632 )

Think we mostly don't notice as [Setuptools ships its own Distutils]( https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html ), which it uses by default. We probably should move to it explicitly in CuPy as Distutils (even within Setuptools) [is deprecated]( https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools )

Since Setuptools has taken over Distutils support, they have been making changes like the C++ changes we adapted here, which were included in PR: https://github.com/pypa/distutils/pull/228

Now why did this only affect us now? Think Setuptools updated its vendored Distutils recently ( like commit: https://github.com/pypa/setuptools/commit/2c937116cc0dcd9b26b6070e89a3dc5dcbedc2ae ) and this went out [72.2.0]( https://github.com/pypa/setuptools/releases/tag/v72.2.0 ), which conda-forge dutifully packaged recently ( https://github.com/conda-forge/setuptools-feedstock/pull/355 )

jakirkham: Sorry I edited that message a few times to fill in more details. Hopefully that helps clarify. Please ask more questions as needed

install/cupy_builder/cupy_setup_build.py
```diff
@@ -336,21 +336,35 @@ def make_extensions(ctx: Context, compiler, use_cython):
             os.symlink(f'{os.environ["BUILD_PREFIX"]}/x86_64-conda-linux-gnu/'
                        'bin/x86_64-conda-linux-gnu-ld',
                        f'{os.environ["BUILD_PREFIX"]}/bin/ld')
-        if (int(os.environ.get('CONDA_BUILD_CROSS_COMPILATION', 0)) == 1 or
-                os.environ.get('CONDA_OVERRIDE_CUDA', '0').startswith('12')):
+        if (PLATFORM_LINUX and (
+                int(os.environ.get('CONDA_BUILD_CROSS_COMPILATION', 0)) == 1 or
+                os.environ.get('CONDA_OVERRIDE_CUDA', '0').startswith('12'))):
```
leofang, line 341: Can we restore the change made to the if-condition? Previously, this if branch can be triggered for Windows + CUDA 12 as well.

install/cupy_builder/cupy_setup_build.py
```diff
@@ -336,21 +336,35 @@ def make_extensions(ctx: Context, compiler, use_cython):
...
             # that are executable in the build environment, not the target
             # environment.
             compiler = ccompiler.new_compiler()
-            cc = os.environ['CC_FOR_BUILD' if PLATFORM_LINUX else 'CC']
-            cxx = os.environ['CXX_FOR_BUILD' if PLATFORM_LINUX else 'CXX']
+            sysconfig.customize_compiler(compiler)
+            # Need to match and replace these
+            # https://github.com/pypa/distutils/blob/30b7331b07fbc404959cb37ac311afdfb90813be/distutils/unixccompiler.py#L117-L129
+            cc = os.environ['CC_FOR_BUILD']
+            cxx = os.environ['CXX_FOR_BUILD']
```
leofang, line 351: Does this work on Windows? IIRC `{CC,CXX}_FOR_BUILD` are not defined on Windows.

install/cupy_builder/cupy_setup_build.py
leofang: So does it mean adding this line alone is enough to fix the c++ not found error, and that the rest of the changes are mostly cosmetic?

leofang (commented):
Thanks, John! Looks good over all. Left a few questions.

Now that we found the root cause, perhaps we should update the PR title/description to mention this is to support newer setuptools (72.2.0+)?

install/cupy_builder/cupy_setup_build.py
jakirkham: The patching here doesn't really make sense on Windows as `distutils` doesn't use it

Please see [this MSVC logic in `distutils`]( https://github.com/pypa/distutils/blob/30b7331b07fbc404959cb37ac311afdfb90813be/distutils/msvccompiler.py#L233-L238 ) compared to the [UNIX compiler logic in `distutils`]( https://github.com/pypa/distutils/blob/30b7331b07fbc404959cb37ac311afdfb90813be/distutils/unixccompiler.py#L117-L129 )

So don't think it makes sense to keep running this there

install/cupy_builder/cupy_setup_build.py
jakirkham: No, it means that this loads the executables targeting the `host_platform`. AIUI we should only use executables targeting the `build_platform`. So this makes it easier to see when we are missing something from the patch as references to the `host_platform` will give us a clue

install/cupy_builder/cupy_setup_build.py
jakirkham: For simplicity, let's discuss in this thread: https://github.com/cupy/cupy/pull/8544#discussion_r1730647260

We can then decide what changes to make after

install/cupy_builder/cupy_setup_build.py
leofang: Thanks for the explanation, John! I tried to refresh my memory by tracing the feedstock commit history, and I guess you were right, this is irrelevant on Windows (also confirmed with an actual run in https://github.com/conda-forge/cupy-feedstock/pull/285.

install/cupy_builder/cupy_setup_build.py
leofang: Thanks, John, good catch! I think the doc could have been more clear that a `new_compiler()` call should be always followed by `customize_compiler()`. I do see it's always done correctly in CuPy so far ([link](https://github.com/search?q=repo%3Acupy%2Fcupy+new_compiler&type=code)) except here.

leofang (approved):
LGTM, thanks, John!

install/cupy_builder/cupy_setup_build.py
leofang: We concluded above that nothing to change here.

jakirkham:
@kmaehashi could you please take another look? 🙂

jakirkham:
Friendly reminder @kmaehashi 😉

Also hope you had a good break! 😄

kmaehashi (approved):

kmaehashi:
Thanks John, LGTM!

jakirkham:
Thanks Kenichi! 🙏

PR #8548, opened 2024-08-22
[v13] Bump version in Dockerfile for v13.3.0

kmaehashi:


kmaehashi:
I noticed `install` package was unexpectedly installed in the Docker image, which is now removed from PyPI and marked archived https://github.com/eugenekolo/pip-install. I've assessed the code and Docker images we published, but I see no harm in the code.

PR #8559, opened 2024-08-24
fixed arg order to match NumPy

dakofler:


kmaehashi:
Thanks, this looks like a change in NumPy v1.24 (@asi1024 found this 😄).
NumPy v1.23: https://numpy.org/doc/1.23/reference/generated/numpy.histogram.html#numpy.histogram
NumPy v1.24: https://numpy.org/doc/1.24/reference/generated/numpy.histogram.html#numpy.histogram

PR #8562, opened 2024-08-26
add nccl.broadcast 64-bit support

jemiryguo:
In the `cupy_backends/cuda/libs/nccl.pyx`, all functions except `broadcast` use `int` as the type of `count`, others are all `size_t`. I think this is a typo, which prevent broadcasting larger arrays.


leofang:
Good catch, thanks!

kmaehashi (approved):
Thanks, LGTM!

Issue #8514, opened 2024-08-19
Support NumPy 2.1


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [ ] Read Release Notes.
- [ ] Fix code and CI to support the new vesrion.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - **Update Requirements**: For NumPy/SciPy, bump the pinned minimum/maximum version in `setup.py` (see #4773.)
- [ ] Backport the above PR.
- [ ] For NumPy update, reflect the changes in `numpy.array_api` to `cupy.array_api`.
- [ ] Backport the above PR.
- [ ] Fix documentation.
  - Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) (note: the baseline version must be updated only in the development branch).
  - Update `docs/requirements.txt` to use the latest versions.
- [ ] Backport the above PR.
- [ ] Apply fixes needed to follow the specification changes.

----

Release notes: https://github.com/numpy/numpy/blob/maintenance/2.1.x/doc/source/release/2.1.0-notes.rst
* New API: `numpy.unstack`
* Deprecation: Passing non-integer inputs as the first argument of bincount is now deprecated
* Expired Dreprecation: Scalars and 0D arrays are disallowed for numpy.nonzero and numpy.ndarray.nonzero.

kmaehashi:
We already covered NumPy v2.1 in our CI a long ago. Closing.

PR #8565, opened 2024-08-26
Add NumPy 2.x CI for Linux

asi1024:


docs/source/install.rst
```diff
@@ -25,7 +25,7 @@ Python Dependencies
 
 NumPy/SciPy-compatible API in CuPy v14 is based on NumPy 2.0 and SciPy 1.13, and has been tested against the following versions:
```
kmaehashi, line 26:
```suggestion
NumPy/SciPy-compatible API in CuPy v14 is based on NumPy 2.1 and SciPy 1.14, and has been tested against the following versions:
```

kmaehashi:
Thanks for the hard work @asi1024! LGTM!

Issue #8568, opened 2024-08-28
CuPy fails with setuptools 74.0.0


kmaehashi:
setuptools v74.0 (specifically https://github.com/pypa/setuptools/pull/4600) removed `setuptools.msvc` module, which has been there for 8 years. Currently, CuPy depends on the module for both build-time and runtime:

* build-time: https://github.com/cupy/cupy/blob/69a2137ec7577adb03407ed3e8db373ee2118b4c/install/cupy_builder/_compiler.py#L265-L266
* runtime (Windows only): https://github.com/cupy/cupy/blob/69a2137ec7577adb03407ed3e8db373ee2118b4c/cupy/cuda/compiler.py#L102

Until the problem is resolved, we will pin the setuptools version in the CI. (#8567)

### Workarounds (for users)

* Windows users seeing `Failed to auto-detect cl.exe path` warning: Do one of the following workarounds:
    1. Downgrade setuptools to v73.x or earlier
    2. Upgrade setuptools to v74.1.0 or later, and do `import setuptools.msvc` along with `import cupy`
    3. Manually add `cl.exe` directory (e.g., `C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\bin\HostX86\x64\`) to `PATH`
* Linux/Windows users experiencing `ModuleNotFoundError: No module named 'setuptools.msvc'` error when building CuPy from source: This problems happens with setuptools v74.0.0 only. Try upgrading setuptools to v74.1.0 or later (or downgrade setuptools to v73.x or earlier).

Both problems will be fixed in the future release (#8583).

### Additional Context

It seems scikit-build is also affected by this change: https://github.com/pypa/setuptools/issues/4608, https://github.com/scikit-build/scikit-build/issues/1114

henryiii:
Is this fixed with 0.18.1?

kmaehashi:
Hi @henryiii, actually our build doesn't use scikit-build and I'm just surveying how other projects will catch-up with this change. Sorry for the confusion!

PR #8574, opened 2024-08-29
Support building CuPy with setuptools 74

kmaehashi:
This PR fixes the build-time usage of setuptools. Part of #8568.


kmaehashi:
Reverted to use the latest setuptools in `pretest.yml` (partial revert of #8567).

PR #8583, opened 2024-09-03
Support setuptools 74.0.0 or later

### LINKED ISSUE #8568: CuPy fails with setuptools 74.0.0

kmaehashi:
Closes #8568. Follows-up #8574. Reverts #8567.

edit: This PR adopts recent setuptools changes, specifically v74.0.0 (which [removed](https://github.com/pypa/setuptools/pull/4600) `setuptools.msvc`) and v74.1.0+ (which [reverted](https://github.com/pypa/setuptools/pull/4626) `setuptools.msvc` but requires explicit module import instead of `import setuptools`).

<details>
<summary>Original description</summary>

On Windows platform with setuptools 74+, it will now take 2-3 seconds to find `cl.exe`. This overhead happens only once per process, and only when the use of `nvcc` is requested.

`setuptools.msvc` (which we rely on for years and is now removed) was able to discover `cl.exe` very quickly as it directly reads the Windows registry. The new approach introduced in this PR (which only uses the public/documented setuptools API) retrives the compiler path through `vcvarsall.bat`, which incurs subprocess invocation.
</detail>


kmaehashi:
Oops, I just noticed `setuptools.msvc` removal has been reverted...

https://github.com/pypa/setuptools/pull/4626

kmaehashi:
/test mini

This one is review ready!

takagi (approved):
LGTM!

PR #8614, opened 2024-09-18
Extended installation resources with pyx files

akochepasov:
The addtional pyx files allow native calls to cublas from 3rd party cython native cimports

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,6 +73,26 @@ cupy_package_data = [
     'cupyx/scipy/ndimage/cuda/LICENSE',
     'cupyx/scipy/ndimage/cuda/pba_kernels_2d.h',
     'cupyx/scipy/ndimage/cuda/pba_kernels_3d.h',
+    # for providing cupy native resources to 3rd party cython
+    'cupy/cuda/__init__.pxd',
+    'cupy/cuda/common.pxd',
+    'cupy/cuda/device.pxd',
+    'cupy/cuda/memory.pxd',
+    'cupy/cuda/pinned_memory.pxd',
+    'cupy/cuda/stream.pxd',
+    'cupy_backends/cuda/__init__.pxd',
+    'cupy_backends/cuda/libs/__init__.pxd',
+    'cupy_backends/cuda/libs/cublas.pxd',
+    'cupy_backends/cuda/api/__init__.pxd',
+    'cupy_backends/cuda/api/_driver_enum.pxd',
+    'cupy_backends/cuda/api/_driver_extern.pxi',
+    'cupy_backends/cuda/api/_driver_typedef.pxi',
+    'cupy_backends/cuda/api/_runtime_enum.pxd',
+    'cupy_backends/cuda/api/_runtime_extern.pxi',
+    'cupy_backends/cuda/api/_runtime_softlink.pxi',
+    'cupy_backends/cuda/api/_runtime_typedef.pxi',
+    'cupy_backends/cuda/api/driver.pxd',
+    'cupy_backends/cuda/api/runtime.pxd',
 ] + [
     x for x in glob.glob('cupy/_core/include/cupy/**', recursive=True)
     if os.path.isfile(x)


kmaehashi:
Thanks for the PR! Our cuBLAS binding only implements what we need inside CuPy and is not essentially intended to be used by third parties. I think https://github.com/NVIDIA/nvmath-python best serves for that purpose, as it provides the Cython binding of complete and latest CUDA APIs. (cc/ @leofang)

leofang:
Thanks for contribution, @akochepasov. As @kmaehashi noted, CuPy's cuBLAS bindings aren't complete and nvmath-python is NVIDIA's official solution for pythonic access to math libraries. Low-level library bindings (similar to CuPy's) are also provided as an experimental by-product (used by the pythonic APIs). Also, CuPy considers the Cython layer as an implementation detail and has never guaranteed API stability, so things could break between releases. It would be better to avoid accessing CuPy's Cython modules.

To help nvmath-python assess the demand of stabilizing the bindings by its 1.0 release, it'd be great if you could file an issue in the nvmath-python repo, with your intended usage. 

My question on this particular PR is: CuPy's bindings are `cpdef`'d, so you only save a most-likely-negligible overhead by `cimport`'ing them. Couldn't you just access the bindings from within Python?

akochepasov:
Thanks @leofang  and @kmaehashi for review and recommendations. nvmath is recent and I didn't know about it. 
While this package may be promising, is in the beta stage. I also tried another nvidia package `cunumeric` and it didn't support windows and was not going to.
CuPy bindings are `cdef`d, this is why `pxd` exports needed. Bubbling up from native calls and derouting over python makes the whole process noticeably slower (I'd say ~2x), it is not negligible.
I used cupy with pxd for a few years. It works really well. The interface is stable and usable. It is also multiplatform, much smaller than nvmath and works with existing cuda/cublas installation. 
You guys did a great job.

kmaehashi:
How about using `cythonize(..., include_path=['/path/to/cupy_source_tree'])`? This should allow building third-party Cython packages that cimports from CuPy. In this case the version of the `cupy_source_tree` and the version of CuPy actually installed must match, for which the users hold responsibility.

leofang:
> While this package may be promising, is in the beta stage. 

And with an eventual goal of moving toward the v1.0 stable release 🙂 Also, we've been busy working at NVIDIA to ensure CuPy will be able to depend on a future nvmath-python release, which does support both Linux and Windows as the underlying CUDA does. So, the sooner you start testing it, using it, and sharing your feedbacks with us, the better support we will be able to offer and the quicker to make it stable! (cc @samaid for vis)

> I also tried another nvidia package `cunumeric` and it didn't support windows and was not going to.

cuNumerics is never meant to be used for the purpose in discussion here, nvmath-python is. cuNumerics's mission is distinct from the majority of Python GPU libraries: Support hyper scaling to allow painless transition of NumPy-based workloads from a desktop to a HPC cluster. NVIDIA offers many packages and I suggest to not let one package influence your decision for using another one, as they serve for different purposes and audience :) 

> CuPy bindings are `cdef`d, this is why `pxd` exports needed.

I am sorry, but I do not understand what functions you're referencing. All functions are `cpdef`'d, e.g.
https://github.com/cupy/cupy/blob/cc27124aa06010da5b9ca4e04b86d0e467c6e9e6/cupy_backends/cuda/libs/cublas.pxd#L106
because `cdef`'d functions would not allow us to call them from within Python, which is needed in various corners of CuPy. My impression with `cpdef` functions (also [backed by Cython documentation](https://cython.readthedocs.io/en/latest/src/tutorial/clibraries.html#testing-the-result)) is that calling them from Cython does save a little (but only a little) compared to from Python. Seeing a 2x difference is suspicious to me.

akochepasov:
Since NVIDIA offers many packages I'm not sure letting one package influence a decision for using one particular is a good idea. They serve for different purposes and audiences. Using cupy have been a best strategy for me for quite a while and I hoped to help extending cupy ability with pdx files similar to `scipy` . 

> How about using cythonize(..., include_path=['/path/to/cupy_source_tree'])

Currently, I just have copied pdxes inside the installed cupy from the respective source for more than 3 years and it works well. The pdx correspond to the I can use a cupy source tree with the same result. This makes it significantly more complex to deploy from sources though, as I need to carry two source trees, and adds moving parts.

I understand the change may be unwelcomed by some reasons and may probably search for another package. There are not many which have such native interface and well maintained. A switch from cupy to nvmath would take me a few months due to another workloads.

If you don't think this is a reasonable addition, feel free to reject the PR.

kmaehashi:
Thanks for the follow up comment @akochepasov.

From my personal experience of over ten years maintaining several Python libraries, I can say that while making APIs public is easy, it is often irreversible and can lead to constant challenges for the maintainers, affecting the velocity of the development. Even clearly labelling APIs as "private" (or "undocumented"), it can still make for a tough decision to change or remove it, and once it's done, we inevitably receive a barrage of bug reports claiming the library is broken. Actually it took us years to remove the `compile_with_cache` (#5297) or private namespaces (#3420).

Once again, to clear up any possible misunderstanding, your request is completely understandable and legitimate. In fact, I was the one who submitted https://github.com/cupy/cupy/issues/79 when I was merely an early user of CuPy.

I've cross-liked this PR in #130 to collect and track the needs for the Cython-level APIs. When it comes to exposing a public Cython API, I believe we will need:

1. A clear distinction between public and private APIs. This might require some refactoring.
2. Documentation to communicate users about the compatibility level they can expect from the Cython API, like SciPy does.
3. Continuous integration to test the Cython API. As of now, we can't determine whether or not the Cython cimport dependencies are closed within the PXD files you've added in this pull request.

As we see it, these are our responsibilities in providing a stable product to our users and keeping the project maintainable.

akochepasov:
@kmaehashi Impressive answer anyway. Thank you for the examples. It's a work just to pull all the history. Well, I didn't win this time. cupy is great anyway.

PR #8622, opened 2024-09-24
Add `__binsparse__` protocol.

hameerabbasi:


asi1024:
@hameerabbasi We will merge this PR after SciPy releases a version of sparse matrix that supports `__binsparse__` protocol.

leofang:
Indeed, it'd be better that this effort starts with SciPy then propagates to CuPy, not the other way around.

Also it'd be nice to cross-reference the efforts in different projects; a data exchange protocol without community collaboration/coordination does not make much sense 🙂  
- https://github.com/pydata/sparse/pull/764
- https://github.com/data-apis/array-api/issues/840
- https://github.com/GraphBLAS/binsparse-specification/issues/4

leofang:
cc @eriknw for vis

Issue #8606, opened 2024-09-17
Support ROCm 6.3


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read [ROCm Release Notes](https://docs.amd.com/).
- [x] Update AMD driver in Jenkins test infrastructure (ask @kmaehashi).
- [ ] Fix code and CI to support the new vesrion.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`. (https://github.com/cupy/cupy/pull/8623)
  - **Wheel Package Detection**: Add the package to the [duplicate detection](https://github.com/cupy/cupy/blob/master/cupy/_environment.py). (**TBD**)
- [ ] Backport the above PR.
- [ ] Fix `cupy-release-tools` to support the new version. (https://github.com/cupy/cupy-release-tools/pull/396)
- [ ] Backport the above PR.
- [ ] Fix documentation.
  - Add new wheel package to the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) and `README.md`.
  - Update requirements in the installation guide.
- [ ] Backport the above PR.
- [ ] Add new wheel package to the [website](https://cupy.dev).
- [ ] Implement or create an issue to support new features, if applicable.

kmaehashi:
ROCm 6.3 is out.

kmaehashi:
We haven't fully confirmed everything but I think it should work.

leofang:
Superseded by #9529.

PR #8623, opened 2024-09-25
Add CI for ROCm 6.2

kmaehashi:
Ref #8606


.github/workflows/self-hosted-ci.yml
```diff
@@ -2,6 +2,8 @@ name: Self-Hosted CI
 on:
   issue_comment:
     types: [created]
+  pull_request_review_comment:
+    types: [created]
```
kmaehashi, line 6: /test rocm

PR #8628, opened 2024-09-26
Update precommit

EarlMilktea:
This PR updates pre-commit CI for better style checking.

### Description of changes

- Upgrade hooks in `.pre-commit-config.yaml`
- Remove deprecated `mirrors-autopep8`
- Unpin `types-setuptools` because `setuptools` itself is not pinned
- Fix linter warnings
  - ~Ignoring rule E721 for now~
- Change/upgrade stylecheck deps. (deprecated)
  - Drop `pycodestyle` as it's shipped with `autopep8`

### DIFF
--- a/cupyx/scipy/sparse/linalg/_iterative.py
+++ b/cupyx/scipy/sparse/linalg/_iterative.py
@@ -339,7 +339,6 @@ def _make_fast_matvec(A):
     from cupy_backends.cuda.libs import cusparse as _cusparse
     from cupyx import cusparse
 
-    matvec = None
     if _csr.isspmatrix_csr(A) and cusparse.check_availability('spmv'):
         handle = device.get_cusparse_handle()
         op_a = _cusparse.CUSPARSE_OPERATION_NON_TRANSPOSE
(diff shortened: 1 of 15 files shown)


cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -339,7 +339,6 @@ def _make_fast_matvec(A):
     from cupy_backends.cuda.libs import cusparse as _cusparse
     from cupyx import cusparse
 
-    matvec = None
```
asi1024, line 342: `None` should be returned if `_csr.isspmatrix_csr(A) and cusparse.check_availability('spmv')` is False.

EarlMilktea: Fixed in f7219c58432367e48ccb65bdfc8d55ed35235c50 .

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -1,10 +1,10 @@
-import numpy
-
 import cupy
+import numpy
```
asi1024, line 2: Could you roll back this modification in accordance with rule H306?

```
Import statements must be organized into three parts: standard libraries, third-party libraries, and internal imports. [H306]
```
https://docs.cupy.dev/en/latest/contribution.html#coding-guidelines

EarlMilktea: Fixed in d2f93976a56c33f8440c637a3cb568a34787bf8f .
(This modification was accidentally inserted by vscode)

asi1024 (approved):

asi1024:
LGTM!

PR #8655, opened 2024-10-09
Switch to using platform.machine() instead of platform.processor()

dagardner-nv:
* Some systems return an empty string for `platform.processor()` (https://docs.python.org/3.10/library/platform.html#platform.processor)

Closes #8654

Issue #8654, opened 2024-10-09
cupy fails to locate `vector_types.h` on Debian 6.1


dagardner-nv:
I believe the issue is that `platform.processor()` returns an empty string on some systems (https://docs.python.org/3.10/library/platform.html#platform.processor), and as such on my machine in a Conda env the `cupy._environment._get_include_dir_from_conda_or_wheel` method returns an invalid path:
```
>>> from cupy import _environment
>>> _environment._get_include_dir_from_conda_or_wheel(12, 5)
['/home/dagardner/work/conda/envs/morpheus/targets/-linux/include', '/home/dagardner/work/conda/envs/morpheus/include']
```

One possible work-around would be to call `platform.machine()` on systems that return an empty string:
```
>>> import platform
>>> platform.processor()
''
>>> platform.machine()
'x86_64'
```

Repro:
```
import cupy as cp
cp.array(cp.zeros((1,)))
```

Error:
```
Traceback (most recent call last):
  File "/home/dagardner/work/conda/envs/morpheus/lib/python3.10/site-packages/cupy/cuda/compiler.py", line 692, in compile
    nvrtc.compileProgram(self.ptr, options)
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 125, in cupy_backends.cuda.libs.nvrtc.compileProgram
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 138, in cupy_backends.cuda.libs.nvrtc.compileProgram
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 53, in cupy_backends.cuda.libs.nvrtc.check_status
cupy_backends.cuda.libs.nvrtc.NVRTCError: NVRTC_ERROR_COMPILATION (6)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dagardner/work/conda/envs/morpheus/lib/python3.10/site-packages/cupy/_creation/from_data.py", line 53, in array
    return _core.array(obj, dtype, copy, order, subok, ndmin, blocking)
  File "cupy/_core/core.pyx", line 2408, in cupy._core.core.array
  File "cupy/_core/core.pyx", line 2418, in cupy._core.core.array
  File "cupy/_core/core.pyx", line 2449, in cupy._core.core._array_from_cupy_ndarray
  File "cupy/_core/core.pyx", line 518, in cupy._core.core._ndarray_base.astype
[...]

PR #8657, opened 2024-10-10
Use `platform.machine()` instead of `platform.processor()`

kmaehashi:
See #8654. Follows-up #8655.

### DIFF
--- a/install/cupy_builder/install_build.py
+++ b/install/cupy_builder/install_build.py
@@ -705,7 +705,7 @@ def get_cusparselt_version(formatted=False):
 def conda_get_target_name():
     out = None
     if PLATFORM_LINUX:
-        plat = platform.processor()
+        plat = platform.machine()
         if plat == "aarch64":
             out = "sbsa-linux"
         else:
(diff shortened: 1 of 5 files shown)


leofang (approved):

leofang:
Yeah I think this is the right move. On Windows (+miniforge) `platform.processor()` does not even do what we think it should do. 
截圖 2024-10-10 下午4 06 55
(Sorry for posting a screenshot; I can't copy/paste text from this VM.)

jakirkham (approved):
Thanks Kenichi! 🙏

Agree this is a good idea 🙂

jakirkham:
Side note: Looks like the same Windows build that we have seen issues with also ran into issues here

jakirkham:
Hmm...actually the failure here looks legitimate. Though don't think it is related to this change (unless I'm missing something)

```python
=========================== short test summary info ===========================
FAILED cupyx_tests\scipy_tests\interpolate_tests\test_bspline.py::TestBSpline::test_bspline_derivative_jumps[_param_0_{c=[-1, 2, 0, -1], extrapolate=True}] - AssertionError:	
Not equal to tolerance rtol=1e-07, atol=0	
	
x and y nan location mismatch:	
 x: array([-1.206633,  0.363639, -0.413208,       nan])	
 y: array([-1.206633,  0.363639, -0.413208,  0.      ])	
= 1 failed, 101393 passed, 9227 skipped, 677 deselected, 2025 xfailed, 2609 warnings in 3674.17s (1:01:14) =	
------------------------------------------------------------------------------------------
```

kmaehashi:
It's a known flaky one (#7240), so I believe it's unrelated to this PR.

takagi (approved):
LGTM!

Issue #8557, opened 2024-08-23
Make `NpzFile` API compatible with `numpy.lib.npyio.NpzFile`


kmaehashi:
### Description

Currently, `NpzFile` returned from `cupy.load` does not implement `Mapping` protocol (e.g., `get()`, `items()`, etc. are missing). It would be better to support these methods.

CuPy: https://github.com/cupy/cupy/blob/v13.3.0/cupy/_io/npz.py#L11-L29
NumPy: https://numpy.org/devdocs/reference/generated/numpy.lib.npyio.NpzFile.html

Also it looks NumPy 2.0 made the class documented. We should follow that as well.

### Additional Information

I noticed this when porting [llama3.np](https://github.com/likejazz/llama3.np) for CuPy. If this is implemented, `llama3.np` will run on CuPy just by rewriting `import numpy as np` to `import cupy as np`.

arashandishgar:
Hi,

Currently, numpy `NpzFile` inherits from `collections.abc.Mapping, `but cupy `NpzFile` inherits from object. If cupy `NpzFile` were to inherit from collections.abc.Mapping and implement the required methods which include `__iter__` and `__len__`, this problem would be completely resolved.

Thus, I believe there are two potential solutions for this issue. One option is to implement the methods you've mentioned (`get()`,`items()`, etc.). The second option is to have `NpzFile` inherit from `collections.abc.Mapping.`

Which solution do you think would be better for this issue?

kmaehashi:
I think the latter is the better choice.

arashandishgar:
Okay. I've already implemented it. Give it me a day to send a pull request

PR #8660, opened 2024-10-11
implement mapping protocol in NpzFile

andishgar:
Fixing issue #8557

### DIFF
--- a/cupy/_io/npz.py
+++ b/cupy/_io/npz.py
@@ -11,14 +11,7 @@ _support_allow_pickle = (numpy.lib.NumpyVersion(numpy.__version__) >= '1.10.0')
 
 class BagObj:
     """
-        BagObj(obj)
-
-        Convert attribute look-ups to getitems on the object passed in.
-
-        Parameters
-        ----------
-        obj : class instance
-            Object on which attribute look-up is performed.
+    Convert attribute look-ups to getitems on the object passed in.
     """
 
     def __init__(self, obj):
--- a/cupy/_io/npz.py
+++ b/cupy/_io/npz.py
@@ -9,7 +9,7 @@ import cupy
 _support_allow_pickle = (numpy.lib.NumpyVersion(numpy.__version__) >= '1.10.0')
 
 
-class BagObj(object):
+class BagObj:
     """
         BagObj(obj)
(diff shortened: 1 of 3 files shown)


cupy/_io/npz.py
```diff
@@ -1,17 +1,47 @@
...
 
 import cupy
 
-
 _support_allow_pickle = (numpy.lib.NumpyVersion(numpy.__version__) >= '1.10.0')
 
 
-class NpzFile(object):
+if numpy.lib.NumpyVersion(numpy.__version__) >= '2.0.0':
+    from numpy.lib._npyio_impl import BagObj
```
kmaehashi, line 12: Could you copy the `BagObj` class from NumPy? I think it's better not to rely on NumPy's implementation details.

kmaehashi (changes requested):

cupy/_io/npz.py
arashandishgar: "Initially, I came up with the idea of having a separate BagObj class. However, the entire implementation tries to use NumPy's existing code, so I tried to follow the same approach. If you think it would be better to have a separate BagObj, I can implement it."

docs/source/reference/io.rst
```diff
@@ -12,6 +12,8 @@ NumPy binary files (NPY, NPZ)
 .. autosummary::
    :toctree: generated/
 
+   _io.npz.NpzFile
+   _io.npz.BagObj
```
kmaehashi, line 16: Let's add them once NpzFile is fully compatible with NumPy.

```suggestion
```

cupy/_io/npz.py
```diff
@@ -24,9 +72,55 @@ def __getitem__(self, key):
...
+        return iter(self.npz_file.files)
+
+    def __len__(self):
+        return len(self.npz_file.files)
+
+    def __contains__(self, key):
+        return self.npz_file.__contains__(key)
+
+    def __del__(self):
+        self.npz_file.__del__()
```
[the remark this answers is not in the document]
arashandishgar: You are correct, and it even does something worse by calling the __del__ method in NumPy's `NpzFile` twice.

Issue #8481, opened 2024-08-13
Use Ruff for better styling


EarlMilktea:
### Description

Use [Ruff](https://github.com/astral-sh/ruff) instead of `flake8` and `autopep8`, and also introduce `isort`-equivalent formatting.

## Description of Changes

- `flake8`: Replaced by `ruff check`. It is designed to be a drop-in replacement of `flake8` without any config.
- `autopep8`: Replaced by `ruff format`.
- Ruff rule `I`: Sorts import as PEP8 specifies (a.k.a. `isort`)

#### Pros

- Reduce diffs. by aggressive formatting
- Reduce dependency
- Better performance
- Can introduce more linting rules in the future
- Can format example codes in docstrings
- Support `.ipynb`

#### Cons

- Still under active development: may need to pin version
- Introduce massive changes on first formatting

PR #8671, opened 2024-10-16
Replace `flake8` with `ruff`

EarlMilktea:
Related to #8481 .

This PR replaces `flake8` integrated in `.pre-commit-config.yaml` with `ruff`.

### Changes

- create `.ruff.toml` mostly compatible with the current linting configuration
- fix/silence linter warnings that only `ruff` can detect
- fix invalid `noqa` directives

### Note

- we don't have to update the contribution guide as the details of `pre-commit` CI are intentionally hidden


.pre-commit-config.yaml
```diff
@@ -21,11 +21,14 @@ repos:
...
+    rev: v0.6.9
+    hooks:
+      - id: ruff
+
   - repo: https://github.com/PyCQA/flake8
     rev: 7.1.1
     hooks:
-      - id: flake8
-        # use config from setup.cfg
       - id: flake8
```
kmaehashi, line 32: I think it's now OK to remove flake8 against Cython code as well. We've introduced cython-lint over a year ago (#7508) and it's working perfectly.

.ruff.toml
```diff
@@ -0,0 +1,15 @@
...
+  "cupy/array_api",
+  "docs/source",
+  "tests/cupy_tests/array_api_tests",
+  "third_party",
+]
+
+[lint]
+select = ["E", "F", "W"]
+ignore = [
+  "E501", # line-too-long (let formatters handle this)
```
kmaehashi, line 14: Is it ok to ignore this?

kmaehashi (commented):
LGTM except for comments!

.pre-commit-config.yaml
EarlMilktea: Resolved in f62318a866963b3501e7c3f67b0917645df547c7 .

.ruff.toml
EarlMilktea: It is OK if `ruff format` is present, but seems to be not with `autopep8`.
Let me restore it.

EarlMilktea: Fixed in 3ac295ef9e11a6f4f315456e0ec6d77db201cb8c .

kmaehashi:
Thanks, LGTM!

kmaehashi (approved):

PR #8675, opened 2024-10-17
Add backport reminder

kmaehashi:
This PR adds a script, which has been running on on-premise Jenkins for years, to CuPy repository, because we are moving away from Jenkins.
https://github.com/kmaehashi/chainer-tools/blob/master/task-checkers/backport_checker.py


asi1024 (approved):

kmaehashi:
This had to be backported as the script is referred in both v13/main branches

PR #8677, opened 2024-10-17
Fix sosfilt state output shape when ndim < 2

andfoy:
See https://github.com/scipy/scipy/pull/21713#discussion_r1802139152

This PR ensures that `sosfilt` returns the stage state in the correct shape when the input is not 2-dimensional. This was causing compatibility issues when testing against SciPy


takagi (approved):
LGTM!

takagi:
I'll merge this after the preempted(?) linux cuda126 test finishes.

Issue #6253, opened 2021-12-17
Implement the sparse array API introduced in SciPy 1.8


takagi:
`scipy.sparse` has introduced an array API compatible with NumPy arrays, and its use is recommended for all new work. `cupyx.scipy.sparse` should also implement it.

https://scipy.github.io/devdocs/reference/sparse.html

saswatpp:
I will try this one :)

takagi:
https://github.com/scipy/scipy/pull/14822 is the change made in SciPy. Would you look into it, @saswatpp?

saswatpp:
Yes I will look. Thanks for the reference :)

tobiasbecher:
Are there any current plans on implementing this?

acosmicflamingo:
Could someone assign this to me? I'd like to implement this feature. Also, should the test I write take the approach of asserting to static values, or values returned by scipy via `testing.numpy_cupy_allclose` decorator?

acosmicflamingo:
Never mind; it looks like @eriknw is already working on the feature in this [PR](https://github.com/cupy/cupy/pull/10019).

seberg:
As a note, this should be ready for usage (but of course is very new). The hopefully only missing part is that that the new API is not part of the `.rst` docs yet.  I'll look into that, though.

leofang:
The doc is in: https://github.com/cupy/cupy/pull/10154. Closing. Many thanks for @eriknw @seberg for driving this major feature across the finish line 🎉

PR #8682, opened 2024-10-21
Remove deprecated APIs in `scipy.sparse`

EarlMilktea:
See https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html#expired-deprecations for details.


leofang (changes requested):
Sorry I did not notice this PR. I think this is too aggressive, normally we'd have one major release for deprecation (say, v14) and remove the code in another major release (v15). 

For context of this particular case, this removal is not appropriate so deprecation is not even an option we could consider. `scipy.sparse` did this because they've incrementally added support for the new sparse **arrays**, whereas `cupyx.sparse` has not caught up yet (https://github.com/cupy/cupy/issues/6253) and still only offers sparse **matrices**.

EarlMilktea:
@leofang So this PR itself is inappropriate (should not remove them at all), am I right?

EarlMilktea:
Let me turn this PR into draft state.

leofang:
@eriknw would you mind taking over this PR? (No need to rush for v14.1.0.) We need two things IIUC:
1. Once the sparse array is merged, we raise deprecation warnings for the APIs removed in SciPy. This needs to be a non-breaking change so that we can backport to the v14 branch.
2. By the time v15 is imminent, we raise another PR to remove the API. This will break so cannot be backported.

eriknw:
Thanks all, I finally got a chance to dig into this today.

My take is to first make a surgical PR that can target the next cupy release. This is done in #9900. Then, we can do more with the sparray PR later.

Issue #7306, opened 2023-01-15
Update DLPack support


leofang:
There are two tasks:
- [x] Update to DLPack v0.8 (the last compatible version before API/ABI break): #7307 
- [x] Update to DLPack v1.0 (#8683)

I think we should do the first one now, before CuPy v11.5 (or v11.6) is released, and discuss the plan for the ABI break.

Feel free to assign this task to me, thanks!

kmaehashi:
v0.8 support merged in #7307 (v12) and #7376 (v11).

leofang:
The new DLPack 1.0 is imminent. I'll start working on it after array API standard v2023 is finalized (which includes DLPack 1.0 support).

leofang:
Done in #8683!

PR #8683, opened 2024-10-21
ENH: Implement dlpack v1

seberg:
This implements dlpack version 1 and cleans up some of the error handling a little bit.

Some notes:
* I did not add support for `from_dlpack(..., dl_device!=None)`, because the main point of that is currently .
* We support device requests if the device matches exactly _or_ when the request is to copy to the CPU.
* I grudgingly ignore the readonly flag right now on import, because we may need a way to ignore it.  To some degree, I think this is a producer problem (i.e. a project that produces read-only, should maybe provide a way to get a writeable version or ignore it).
  The reason is basically: This can be deprecated easily enough at some point, but I dunno if it may be a regression.
* The UnitTest/pytest interaction confused the heck out of me, so I replaced everything with fixtures.

I think the tests cover things fairly well, although of course not cleanup in the CPU exposing path.  I am not too worried about that, though.
EDIT: Another untested point is that we set the `IS_COPIED` flag when we copied to CPU.

This means that `np.from_dlpack(cupy_arr, device="cpu")` works (with NumPy 2).  The tests indirectly cover this.


leofang:
btw this will close #7306

cupy/_core/dlpack.pyx
```diff
@@ -19,139 +14,213 @@ import warnings
...
+        owner = array
+        device.device_type = kDLCPU
+        device.device_id = 0
     else:
-        raise ValueError('Unknown dtype')
-    dtype.lanes = <uint16_t>1
-    dtype.bits = <uint8_t>(array.dtype.itemsize * 8)
+        # We need to create a CPU copy.  Assumes owner.dtype == array.dtype.
+        # TODO: As noted at calling site, this does not honor the "stream".
+        owner = array.get(stream=None, order='A')
```
leofang, line 128: This will be causing race conditions if host access follows right after hand-shaking. Need to sync internally:
```suggestion
        owner = array.get(order='A')  # this copies on the current stream and then sync
```

cupy/_core/dlpack.pyx
```diff
@@ -19,139 +14,213 @@ import warnings
...
+
+    cdef int32_t ndim = array._shape.size()
+
+    if not to_cpu:
+        owner = array
+    elif not ensure_copy and device.device_type == kDLCUDAManaged:
+        # Managed memory is CPU accessible but consumer may expect kDLCPU.
+        owner = array
+        device.device_type = kDLCPU
+        device.device_id = 0
```
leofang, line 124: I worry that after some round-trips the information regarding memory identity would be lost. Do we have to do this? IIRC NumPy also supports `kDLCUDAManaged` since the beginning?

leofang (commented):
(WIP review)

cupy/_core/dlpack.pyx
seberg: Ah, thanks for the note.  IIRC, I added `stream=None`, because Cython complained about it.  Maybe it needs to be `stream=arr.stream`?
(but maybe I should just retry)

cupy/_core/dlpack.pyx
seberg: Yeah, I wasn't quite sure about what is best here.  The user explicitly asked for "cpu", so I am not super worried about round-tripping.
OTOH, we could clarify that the result must be *compatible* with CPU (i.e. consumer doesn't need to check the device).

cupy/_core/dlpack.pyx
```diff
@@ -19,139 +14,213 @@ import warnings
...
-        void (*deleter)(DLManagedTensor*)  # noqa: E211
-
-
-def get_build_version():
-    return str(DLPACK_VERSION)
+
+cdef const char* CAPSULE_NAME = "dltensor"
+cdef const char* CAPSULE_NAME_VER = "dltensor_versioned"
+cdef const char* USED_CAPSULE_NAME = "used_dltensor"
+cdef const char* USED_CAPSULE_NAME_VER = "used_dltensor_versioned"
```
leofang, line 23: Q: I was under the impression that Cython 0.29.x does not support `cdef const`? Does this compile fine? (CuPy has not migrated yet: #4610)

cupy/_core/dlpack.pyx
```diff
@@ -19,139 +14,213 @@ import warnings
...
+    cdef DLManagedTensorVersioned *dlm_tensor_ver
+
     # Do not invoke the deleter on a used capsule
-    if cpython.PyCapsule_IsValid(dltensor, 'dltensor'):
-        dlm_tensor = <DLManagedTensor*>cpython.PyCapsule_GetPointer(
-            dltensor, 'dltensor')
+    if cpython.PyCapsule_IsValid(dltensor, CAPSULE_NAME):
+        dlm_tensor = <DLManagedTensor*>(
+            cpython.PyCapsule_GetPointer(dltensor, CAPSULE_NAME))
         dlm_tensor.deleter(dlm_tensor)
```
leofang, line 34: I noticed this seems to be a long-standing bug in CuPy: `deleter` can be null as per the C spec
https://dmlc.github.io/dlpack/latest/c_api.html#c.DLManagedTensor.deleter
```suggestion
        if dlm_tensor.deleter:
            dlm_tensor.deleter(dlm_tensor)
```

cupy/_core/dlpack.pyx
```diff
@@ -19,139 +14,213 @@ import warnings
...
+    elif cpython.PyCapsule_IsValid(dltensor, CAPSULE_NAME_VER):
+        dlm_tensor_ver = <DLManagedTensorVersioned*>(
+            cpython.PyCapsule_GetPointer(dltensor, CAPSULE_NAME_VER))
+        dlm_tensor_ver.deleter(dlm_tensor_ver)
```
leofang, line 38: ditto
```suggestion
        if dlm_tensor_ver.deleter:
            dlm_tensor_ver.deleter(dlm_tensor_ver)
```

leofang (changes requested):
WIP review part 2

cupy/_core/dlpack.pyx
seberg: Yeah, compiled fine.  Was it just that it ignored it, so in C it is gone (I recall issues, but honestly not what)?

cupy/_core/dlpack.pyx
```diff
@@ -19,139 +14,213 @@ import warnings
...
-        uint64_t byte_offset
-
-    ctypedef struct DLManagedTensor:
-        DLTensor dl_tensor
-        void* manager_ctx
-        void (*deleter)(DLManagedTensor*)  # noqa: E211
-
-
-def get_build_version():
-    return str(DLPACK_VERSION)
```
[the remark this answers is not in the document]
seberg: I'll reinstate tomorrow.  It seemed unused to me, but it doesn't matter either way for this PR.

cupy/_core/dlpack.pyx
```diff
@@ -1,11 +1,6 @@
 cimport cpython  # NOQA
 
 from libc cimport stdlib
-from libc.stdint cimport uint8_t
-from libc.stdint cimport uint16_t
-from libc.stdint cimport int32_t
-from libc.stdint cimport int64_t
-from libc.stdint cimport uint64_t
```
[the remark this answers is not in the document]
seberg: I wanted to move the `cdef extern` and that uses this.

cupy/_core/core.pyx
```diff
@@ -285,8 +285,42 @@ cdef class _ndarray_base:
...
+        if dl_device is None or dl_device == self.__dlpack_device__():
+            # We chose the device or the device matches, so export normally.
+            if copy is True:
+                # Could be implemented, but there may be some subtleties to
+                # consider here.
+                raise BufferError("copy=True only supported for copy to CPU.")
+        elif dl_device == (dlpack.kDLCPU, 0):
+            # The user explicitly requested CPU device export.
+            # NOTE:
+            # * We effectively ignore the stream here for now!
```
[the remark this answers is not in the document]
seberg: I'll put this in for now.  When we copy to numpy, I am not quite sure about the stream handling right now (or if we should in theory even handle it).
But if we don't copy, this makes a lot of sense for sure, I think.

cupy/_core/dlpack.pyx
leofang: Ahhh ok, your code was right, just need to remove the comment. You hit a Cython limitation when it comes to invoking cdef'd functions with non-default kw args.
https://github.com/cupy/cupy/blob/7cbf5f57b79ae047ec7774bf0568ca5fb3c561cb/cupy/_core/core.pyx#L1798

leofang: Passing stream=None would mean to use the current stream.

cupy/_core/dlpack.pyx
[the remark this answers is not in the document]
leofang: Got it, I see it now, thanks!

cupy/_core/dlpack.pyx
leofang: I now change my mind again and agree that this is a TODO. This is very tricky as I keep revisiting this part, and I'd like to share my understanding here to aid code review.

When this `toDlpack()` function was called, the producer/consumer streams are ordered, that's why we did not need to pass any stream argument. 

This is no longer true when CPU copy is requested. In the case of NumPy, [`stream=None` is provided](https://github.com/numpy/numpy/blob/816fd7c0d1569bef00b3765d47528b093cf55839/numpy/_core/src/multiarray/dlpack.c#L536), but CuPy turns it to `streamLegacy` in `__dlpack__` and orders it after CuPy's current stream (whatever it is), so we're now on the consumer's (legacy) stream. We need a way to pass this info to `get(stream=consumer_stream, order='A')` to make a blocking copy of the array on the legacy stream to CPU. The easiest way to ensure correctness is to pass it through `toDLPack()`.

cupy/_core/dlpack.pyx
```diff
@@ -20,138 +15,213 @@ import cupy
...
+        dl_tensor = &dlm_tensor_ver.dl_tensor
+        capsule_name = CAPSULE_NAME_VER
+
+        dlm_tensor_ver.manager_ctx = <void*>owner
+        cpython.Py_INCREF(owner)
+        dlm_tensor_ver.deleter = deleter_ver
+
+        # "Versioned" specific initialization:
+        dlm_tensor_ver.version.major = 1
+        dlm_tensor_ver.version.minor = 0
```
leofang, line 162: nit: we can avoid hard-coding the values?
```suggestion
        dlm_tensor_ver.version.major = DLPACK_MAJOR_VERSION
        dlm_tensor_ver.version.minor = DLPACK_MINOR_VERSION
```

cupy/_core/dlpack.pyx
```diff
@@ -20,138 +15,213 @@ import cupy
...
+        dl_tensor.data = <void *><intptr_t>owner.ctypes.data
+        shape = owner.shape
+        strides = owner.strides
 
-    dlm_tensor.manager_ctx = <void*>array
-    cpython.Py_INCREF(array)
-    dlm_tensor.deleter = deleter
+        for n in range(ndim):
+            shape_strides[n] = shape[n]
+        dl_tensor.shape = shape_strides
```
leofang, line 220: Q: isn't this already done above (line 191)?

cupy/_core/dlpack.pyx
```diff
@@ -20,138 +15,213 @@ import cupy
...
+
+        for n in range(ndim):
+            shape_strides[n + ndim] = array._strides[n] // dtype_itemsize
     else:
-        raise ValueError('Unknown dtype')
-    dtype.lanes = <uint16_t>1
-    dtype.bits = <uint8_t>(array.dtype.itemsize * 8)
+        # Same as above, but we got a NumPy array, so go through Python
+        # in the off-chance that the copy has changed the strides.
+        dl_tensor.data = <void *><intptr_t>owner.ctypes.data
```
leofang, line 214: nit: I might be paranoid but I feel I've been bitten by operator precedence before...
```suggestion
        dl_tensor.data = <void *><intptr_t>(owner.ctypes.data)
```

cupy/_core/dlpack.pyx
```diff
@@ -163,53 +233,79 @@ cdef class DLPackMemory(memory.BaseMemory):
...
                                    'DLPack tensor')
         else:
-            if dlm_tensor.dl_tensor.device.device_type not in (
+            if self.dl_tensor.device.device_type not in (
                     kDLCUDA, kDLCUDAManaged):
                 raise RuntimeError('CuPy is built against CUDA, different '
                                    'from the backend that backs the incoming '
                                    'DLPack tensor')
 
-        self.dltensor = dltensor
```
leofang, line 190: Q: This always confuses me. We don't hold onto the (used) capsule because if it's used its deleter is not kicked in, until self (`DLPackMemory`) has a 0 refcount?

cupy/_core/dlpack.pyx
```diff
@@ -320,28 +416,24 @@ cdef inline _ndarray_base _dlpack_to_cupy_array(dltensor) except +:
...
     for i in range(ndim):
         strides_vec.push_back(strides[i] * (bits // 8))
 
-    # Make sure this capsule will never be used again.
-    cpython.PyCapsule_SetName(mem.dltensor, 'used_dltensor')
     return core.ndarray(shape_vec, cp_dtype, mem_ptr, strides=strides_vec)
 
 
-cpdef from_dlpack(array):
+def from_dlpack(array, *, dl_device=None, copy=None):
```
leofang, line 436: TODO: document the new args

leofang (commented):
WIP review part 3 (I am left with the test changes!)

Indeed for tests we want to move away from `unittest` to `pytest`. Thanks for modernizing the test infra 🙏

cupy/_core/dlpack.pyx
seberg: So, now that I think about it.  Let me recap two things:
1. The existing stream logic ensures that the GPU copy is available with respect to the consumer stream order.
2. If we just pass on `stream=consumer_stream` here, the copy to the CPU will also be correctly ordered w.r.t. to the consumer stream.

Lastly: If the user did not provide a `stream=`, we have to use `blocking=True` (default) to ensure the CPU copy is immediately accessible.  But we can pass `blocking=False` otherwise.

One other thing, though:  If we have managed memory, we actually need to add an explicit synchronization to ensure that the memory is CPU available?

cupy/_core/dlpack.pyx
```diff
@@ -20,138 +15,213 @@ import cupy
...
+        if dlm_tensor_ver.deleter:
+            dlm_tensor_ver.deleter(dlm_tensor_ver)
+    else:
+        # No cleanup necessary, capsule was "consumed" (renamed).
+        pass
 
 
 cdef void deleter(DLManagedTensor* tensor) with gil:
-    if tensor.manager_ctx is NULL:
-        return
```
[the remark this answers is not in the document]
seberg: This is our own capsule, and I took care to make sure the capsule is fully initialized before any error/cleanup can happen.  But I am fine with adding it as a defensive check.

cupy/_core/dlpack.pyx
seberg: We transfer `dlm_tensor` ownership from the capsule to `DLPackMemory` (via renaming).  After that, the capsule serves no purpose anymore and can/should be deleted.

cupy/_core/dlpack.pyx
seberg: Maybe a `cpdef` for these?  I don't like fetching these directly from the header, since we should be able to update the header without bumping the version incorrectly accidentally.
OTOH, this repeats in a more than 2 places, so probably good to centralize.

leofang: These macros are exposed to Cython in the dlpack.pxd file. Isn't it the expect behavior (that if we update to a 1.1 header the version exported would be 1.1)? What could be the alternative?

cupy/_core/dlpack.pyx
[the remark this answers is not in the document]
leofang: You're right, it's for our own, and we took care of it 👍

cupy/_core/dlpack.pyx
```diff
@@ -20,138 +15,213 @@ import cupy
...
-cpdef object toDlpack(_ndarray_base array) except +:
-    cdef DLManagedTensor* dlm_tensor = \
-        <DLManagedTensor*>stdlib.malloc(sizeof(DLManagedTensor))
-
-    cdef size_t ndim = array._shape.size()
-    cdef DLTensor* dl_tensor = &dlm_tensor.dl_tensor
-    cdef intptr_t data_ptr = array.data.ptr
-    dl_tensor.data = <void*>data_ptr
-    dl_tensor.ndim = ndim
+cdef void deleter_ver(DLManagedTensorVersioned* tensor) with gil:
```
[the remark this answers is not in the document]
leofang: This can be ignored as per https://github.com/cupy/cupy/pull/8683/files#r1810114359

cupy/_core/dlpack.pyx
leofang: I am thinking it might be better that we retain this info, as a managed-memory-backed NumPy array could be handled over to another GPU library (other than CuPy) and if the information is retained it'd work nicely back to GPU.

cupy/_core/dlpack.pyx
seberg: If there is a new flag in version 1.1, and we update the header but don't change anything else, we would be wrongly advertising 1.1 when we are still exporting 1.0

tests/cupy_tests/core_tests/test_dlpack.py
```diff
@@ -74,6 +89,79 @@ def test_conversion(self, dtype):
         testing.assert_array_equal(
             orig_array.data.ptr, out_array.data.ptr)
 
+    @pytest.mark.parametrize("kwargs", [
+        {}, {"max_version": None}, {"max_version": (1, 0)},
+        {"max_version": (10, 10)}]
```
leofang, line 94: Q: What happens if we pass, say, `max_version = (0, 8)`?

tests/cupy_tests/core_tests/test_dlpack.py
```diff
@@ -74,6 +89,79 @@ def test_conversion(self, dtype):
...
+        arr2 = numpy.from_dlpack(
+            DLDummy(orig_array.__dlpack__(dl_device=(1, 0)), device=(1, 0)))
+
+        numpy.testing.assert_array_equal(orig_array.get(), arr1)
+        assert orig_array.dtype == arr1.dtype
+        # Arrays share the same memory exactly when memory is managed.
+        assert numpy.may_share_memory(arr1, arr2) == (self.memory == "managed")
+
+        arr_copy = numpy.from_dlpack(DLDummy(
+            orig_array.__dlpack__(dl_device=(1, 0), copy=True), device=(1, 0)))
```
leofang, line 149: Q: Does `numpy.from_dlpack(..., copy=True)` require NumPy 2.x?

tests/cupy_tests/core_tests/test_dlpack.py
```diff
@@ -38,25 +52,26 @@ def test_conversion(self, dtype):
...
-    'memory': ('device', 'managed')
-}))
-class TestNewDLPackConversion(unittest.TestCase):
-
-    old_pool = None
-    new_pool = None
+class TestNewDLPackConversion:
 
-    def setUp(self):
+    @pytest.fixture(autouse=True, params=["device", "managed"])
```
leofang, line 57: Q: Should this fixture also have a class scope?

leofang (commented):
Tests look good, thanks for putting all these together!

tests/cupy_tests/core_tests/test_dlpack.py
seberg: We export the unversioned one.

tests/cupy_tests/core_tests/test_dlpack.py
seberg: I'll try.  This was now only active for the tests in the class, I think.  But making it class scope didn't work, because it didn't parametrize the tests.
(I.e. I think this was right and did the right thing; but it's the first time I am putting fixtures into a class :))

tests/cupy_tests/core_tests/test_dlpack.py
seberg: I think I can expand the test in this regard.

cupy/_core/dlpack.pyx
seberg: More importantly, `dl_device=` is wrong, it has to be `device=` here :).

tests/cupy_tests/core_tests/test_dlpack.py
leofang: I see, I saw you had `scope='class'` in another test class, and was assuming the semantics is "the class-scoped fixture's lifetime spans over each class test method's invocation", but maybe I am mistaken.

cupy/_core/dlpack.pyx
seberg: I tried to make the pattern slightly more obvious, by adding a slightly weird:
```
try:
    create capsule
except:
    cleanup
    raise
else:
    # finalize initialization code that must not fail.
```
Where the else is mostly nice because it is a nice place to put the incref (the decref had a bug!).

tests/cupy_tests/core_tests/test_dlpack.py
seberg: OK, will not try it, here is the explenation that I hope is correct:
The scope indicates on which level the fixture is applied.  `scope="class"` means we run the fixture only once per class.

Now, that works in the second test case, but I think it is actually less clear, so rather than adding it here, I'll remove it there.

cupy/_core/dlpack.pyx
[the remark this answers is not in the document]
seberg: Actually, the version has been (correctly) deleted from the header, and I can't find a use case in the repository.

So I'll keep it deleted unless you know that a 3rd party will rely on it.  In which case we would have to fake the _old_ version here, and maybe make sure that some default code paths keep returning old unversioned dlpack tensors.

cupy/_core/dlpack.pyx
seberg: CPU users cannot sync unless they know where the data comes from, which means passing in the `stream` (which we support, even if it may be a bit fuzzy).

So, I think we must sync if a stream wasn't passed in, that is logic I need to add.  If a stream was passed in, we should copy on that stream (and that is all the sync we need).

cupy/_core/dlpack.pyx
leofang: Thinking about it more, shouldn't this be matching what's requested (i.e. `max_version` passed to `__dlpack__`)? Perhaps having a bool `use_versioned` argument is a bit restrictive?

Alternatively, how about just locking the major version?
```suggestion
        dlm_tensor_ver.version.major = DLPACK_MAJOR_VERSION
        dlm_tensor_ver.version.minor = 0
```

Not sure what `cpdef` solution you had in mind, maybe it follows a similar line?

cupy/_core/dlpack.pyx
```diff
@@ -397,7 +502,15 @@ cpdef from_dlpack(array):
             cupy.cuda.runtime.setDevice(dev_id)
             assert runtime._is_hip_environment
             stream = stream_module.get_current_stream_ptr()
-            dltensor = array.__dlpack__(stream=stream)
+
+            # For backwards compatibility we catch TypeErrors for now.
+            try:
+                dltensor = array.__dlpack__(
+                    stream=stream, max_version=(1, 0), copy=copy)
```
leofang, line 509: dito, re: https://github.com/cupy/cupy/pull/8683/files#r1809755389 (hard-wiring the `max_version`)

cupy/_core/dlpack.pyx
```diff
@@ -388,7 +485,15 @@ cpdef from_dlpack(array):
             stream = stream_module.get_current_stream_ptr()
             if stream == 0:
                 stream = stream_module.get_default_stream_ptr()
-            dltensor = array.__dlpack__(stream=stream)
+
+            # For backwards compatibility we catch TypeErrors for now.
+            try:
+                dltensor = array.__dlpack__(
+                    stream=stream, max_version=(1, 0), copy=copy)
```
leofang, line 492: dito, re: https://github.com/cupy/cupy/pull/8683/files#r1809755389 (hard-wiring the `max_version`)

cupy/_core/dlpack.pyx
seberg: Sorry, will push in a second (already did the echange).  I mean creating a global in this file (rather than getting it from the header).

> Shouldn't this be matching what's requested

No, it must match exactly with what is exported, and that is currently `1, 0`.  The only thing is that we should make sure it is smaller than `max_version` (which it clearly is).

cupy/_core/dlpack.pyx
seberg: I have added some logic for this see last commit.  But it isn't really tested so needs careful review maybe.
It also is just generally confusing :(.  The first commit should be more straightforward addressing the review comments.

I am not familiar with how one may test the stream logic (the code paths are covered mostly at least, but nothing checks the stream order is right).

cupy/_core/dlpack.pyx
seberg: Here, it's actually fine, since it is our capsule and our dltensor.  Kept it for now as well.

But, we did have the same bug in the DLMemory implementation, so fixed it there (forced pushed, I had thought I had pushed, but forgot the applied suggestions and didn't notice it didn't go through).

cupy/_core/dlpack.pyx
```diff
@@ -18,140 +13,246 @@ import warnings
...
+        dlm_tensor.manager_ctx = <void*>owner
+        dlm_tensor.deleter = deleter
+
+    # Create capsule now. After the else, the capsule will clean up on error.
+    # (Note that it is good to handle expected BufferErrors early.)
+    try:
+        capsule = cpython.PyCapsule_New(
+            dlm_tensor_ptr, capsule_name, pycapsule_deleter)
+    except BaseException:
+        stdlib.free(dlm_tensor)
```
kmaehashi, line 212:
```suggestion
        stdlib.free(dlm_tensor_ptr)
```

cupy/_core/dlpack.pyx
```diff
@@ -18,140 +13,246 @@ import warnings
...
+
+    if not to_cpu:
+        owner = array
+    elif not ensure_copy and device.device_type == kDLCUDAManaged:
+        # Managed memory is CPU accessible.  Note that the consumer may expect
+        # `kDLCPU` here.  We only honor this request in spirit, but not
+        # strictly (because e.g. NumPy will remember the managed part).
+        owner = array
+        device.device_type = device.device_type
+        device.device_id = device.device_id
```
kmaehashi, line 150: Q: What are these lines for?

cupy/_core/dlpack.pyx
```diff
@@ -18,140 +13,246 @@ import warnings
...
+        device.device_id = 0
+
+    cdef void *dlm_tensor_ptr = stdlib.malloc(
+        sizeof(DLManagedTensorVersioned) if use_versioned
+        else sizeof(DLManagedTensor)
+    )
+    if dlm_tensor_ptr == NULL:
+        raise MemoryError()
+
+    # Note: could coalesc this with the previous allocation in principle
```
kmaehashi, line 170:
```suggestion
    # Note: could coalesce this with the previous allocation in principle
```

kmaehashi (commented):
Thanks for the hard work, @seberg!

cupy/_core/dlpack.pyx
seberg, line 150:
```suggestion
```
Nothing, I previously changed it to `kDLCPU` here and didn't pay attention changing it.

kmaehashi (approved):

cupy/_core/dlpack.pyx
```diff
@@ -18,140 +13,244 @@ import warnings
...
+    elif not ensure_copy and device.device_type == kDLCUDAManaged:
+        # Managed memory is CPU accessible.  Note that the consumer may expect
+        # `kDLCPU` here.  We only honor this request in spirit, but not
+        # strictly (because e.g. NumPy will remember the managed part).
+        owner = array
+        if stream is None:
+            # The user did not request a stream to synchronize on.  We have to
+            # assume they don't even know this is GPU data, so must fully
+            # synchronize on the current stream.
+            py_stream_module.get_current_stream().synchronize()
```
leofang, line 153:
```suggestion
        if stream is not None:
            # When reaching here the user requests a safe access on CPU but
            # without any copy. We ensure this by sync'ing on either CuPy's
            # current stream or the consumer stream.
            stream.synchronize()
```

cupy/_core/dlpack.pyx
```diff
@@ -18,140 +13,244 @@ import warnings
...
+    else:
+        # We need to create a CPU copy.  Assumes owner.dtype == array.dtype.
+        owner = array.get(
+            stream=stream, order='A', out=None, blocking=stream is None)
```
leofang, line 157: This reads: We copy over `stream` and then synchronize.
```suggestion
        owner = array.get(stream=stream, order='A')
```

leofang (commented):
Thanks, Sebastian/Kenichi-san, I think this should help simplify the stream handling. Let me know what you think.

cupy/_core/dlpack.pyx
```diff
@@ -18,140 +13,244 @@ import warnings
...
-        dtype.code = <uint8_t>kDLComplex
-    elif array.dtype.kind == 'b':
-        dtype.code = <uint8_t>kDLBool
+
+    return device
+
+
+# The name of this function is following the framework integration guide of
+# TensorComprehensions.
+cpdef object toDlpack(
```
leofang, line 104: I don't like that it's exposed to Python (it's already done before this PR), but it seems we might need a deprecation cycle before making it `cdef`'d... 😞 (https://github.com/cupy/cupy/issues/8703)

cupy/_core/dlpack.pyx
```diff
@@ -18,140 +13,244 @@ import warnings
...
+    dl_tensor.dtype.code = dtype_code
+    dl_tensor.dtype.lanes = <uint16_t>1
+    dl_tensor.dtype.bits = <uint8_t>(dtype_itemsize * 8)
+
+    dl_tensor.device = device
+
+    # Fill in the shape and strides information (depends on GPU vs CPU array).
+    # (assumes strides are a multiple of itemsize, that should be OK for cupy.)
+    if owner is array:
+        dl_tensor.data = <void *>array.data.ptr
```
leofang, line 233: This is even more never-wrecking 😅
```suggestion
        dl_tensor.data = <void *><intptr_t>(array.data.ptr)
```
(Previously, the conversion is done through an intermediate assignment.)

cupy/_core/dlpack.pyx
```diff
@@ -320,34 +448,34 @@ cdef inline _ndarray_base _dlpack_to_cupy_array(dltensor) except +:
...
     for i in range(ndim):
         strides_vec.push_back(strides[i] * (bits // 8))
 
-    # Make sure this capsule will never be used again.
-    cpython.PyCapsule_SetName(mem.dltensor, 'used_dltensor')
     return core.ndarray(shape_vec, cp_dtype, mem_ptr, strides=strides_vec)
 
 
-cpdef from_dlpack(array):
+def from_dlpack(array, *, device=None, copy=None):
```
leofang, line 468: Note: Changing `cpdef` to `def` would cut out the Cython exposure. I don't suppose any users to use this from Cython, so I am fine with it. Just wanna call it out.

tests/cupy_tests/core_tests/test_dlpack.py
```diff
@@ -74,6 +89,96 @@ def test_conversion(self, dtype):
...
             orig_array.data.ptr, out_array.data.ptr)
 
+    def test_from_dlpack_and_conv_errors(self):
+        orig_array = _gen_array("int8")
+
+        with pytest.raises(NotImplementedError):
+            cupy.from_dlpack(orig_array, device=orig_array.device)
+
+        with pytest.raises(BufferError):
+            # `__dlpack__` only allows `copy=True` for host copies.
```
leofang, line 99: Took me a while to think about this. Let's expand the comment a bit
```suggestion
            # Currently CuPy's `__dlpack__` only allows `copy=True`
            # for host copies.
```
(to emphasize it's not because the protocol disallows it, but because we don't want to overcomplicate it).

leofang:
@seberg LGTM, I think only https://github.com/cupy/cupy/pull/8683#discussion_r1818319735 needs immediate attention; my other suggestions can be directly applied, but they're all internal optimizations only, and if we want to expedite merging we can save them for another PR.

leofang:
(btw @kmaehashi I dunno what access level it'd require, but I cannot "resolve conversations" as I review the PRs, so many threads above still remain open...)

cupy/_core/dlpack.pyx
seberg: I think this is already `intptr_t` since it should stay in Cython.  But indeed no need to require the reader guess at that!

seberg:
@leofang I unsure about the refactor suggestion, although I agree that it is a bit overly complex and smells like it should be simplified.

The reason is that, the logic as I intended it differentiates (both for consumer and internal):
* `stream=None`:
  * synchronize after copying (or block until all operations are done if managed memory).
* `stream=current_stream` (or any other stream):
  * Consumer synchronizes.  The operation is never blocking.

And I think the suggested change would change that to always synchronize.

leofang:
The linux-cuda126 CI failed, but it doesn't seem relevant. Is it a known failure?
```
[01:07:58.018443 STDOUT 1378]](https://ci.preferred.jp/cupy.linux.cuda126/176891/#L24929)	=========================== short test summary info ============================	
[01:07:58.018452 STDOUT 1378]](https://ci.preferred.jp/cupy.linux.cuda126/176891/#L24930)	FAILED cupyx_tests/scipy_tests/interpolate_tests/test_bspline.py::TestBSpline::test_bspline_derivative_jumps[_param_5_{c=[[-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1], [-1, 2, 0, -1]], extrapolate='periodic'}]	
[01:07:58.018459 STDOUT 1378]](https://ci.preferred.jp/cupy.linux.cuda126/176891/#L24932)	= 1 failed, 105495 passed, 5808 skipped, 1479 deselected, 785 xfailed, 2518 warnings in 2956.03s (0:49:16) =	27s
```

leofang:
> `Resolve conversation` can only be done by pull-request author or repo admin. 😞

It seems granting "write" access should be sufficient, no need to bump to "maintain" or "admin":
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#resolving-conversations

leofang:
> And I think the suggested change would change that to always synchronize.

@seberg Yes, it was the intention:
- generally, to access managed memory on CPU safely, a synchronization is needed, DLPack assumes a safe handover is done, implying users should not need to do anything themselves
    - it is nontrivial for requesting CPU access + passing in a custom stream (since `from_dlpack()` does not take a `stream` arg), it's a corner case
    - if a consumer library went great length and managed to do this, it should just pass `stream=-1` not a custom stream, if it does not want CuPy (as producer) to manage stream order

Therefore, I suggested to keep it clean and always synchronize, reserving `stream=None` (used internally) only for `stream=-1` (used at handshaking).

seberg:
Hmmm, if we synchronize, is there a possible reason in passing `stream!=None`?  Maybe it is just as well to not implement it even?

I can see semi-usecases for wanting an async copy (such as a when copying multiple arrays), but when paired with `from_dlpack()` there seems unlikely to be a reason to pass `stream`.

TL;DR: sorry to ask for one more nudge: but happy to change it in either direction given that nudge :).

cupy/_core/core.pyx
```diff
@@ -285,13 +285,50 @@ cdef class _ndarray_base:
...
+        #       We can (and must) assume that it is compatible with our device.
         curr_stream = stream_module.get_current_stream()
         curr_stream_ptr = curr_stream.ptr
 
         # stream must be an int for CUDA/ROCm
-        if not runtime._is_hip_environment:  # CUDA
+        if to_cpu and stream is None:
+            # We will use the current stream to copy/sync later.
+            stream = None
+        elif not runtime._is_hip_environment:  # CUDA
```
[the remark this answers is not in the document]
leofang: I think I was wrong 😞 Here, `stream` is an argument passed to `__dlpack__`, and based on the doc `stream=None` means to use the legacy default stream, and it should not be affected by whether `to_cpu` is true, so the old treatment actually is already OK:
```suggestion
        if not runtime._is_hip_environment:  # CUDA
```

leofang (commented):
> TL;DR: sorry to ask for one more nudge: but happy to change it in either direction given that nudge :).

No worries, it is necessary that we discuss as it can be nontrivial to get right (see below). It's also why we want to provide `cuda.core.StridedMemoryView` as a long-term solution... it's just too painful to implement DLPack support 😢 

> Hmmm, if we synchronize, is there a possible reason in passing `stream!=None`? Maybe it is just as well to not implement it even?
> 
> I can see semi-usecases for wanting an async copy (such as a when copying multiple arrays), but when paired with `from_dlpack()` there seems unlikely to be a reason to pass `stream`.

One potential use case is an async array library (like PyTorch) supporting both CPU/GPU access, then even on the CPU side they could have a `stream` that can be passed to the producer's `__dlpack__()`. 

I now think that my suggestion (with a correction below) was necessary, because `__dlpack__(..., stream=None)` has a special meaning. If we just sync over the current stream in the present form, it'd be wrong because the stream order was not established before calling `toDlpack()` and we're just sync'ing over the consumer stream without ordering it after the producer's. Let me know if it makes better sense now.

leofang:
> I now think that my suggestion (with a correction below) was necessary, because `__dlpack__(..., stream=None)` has a special meaning. If we just sync over the current stream in the present form, it'd be wrong because the stream order was not established before calling `toDlpack()` and we're just sync'ing over the consumer stream without ordering it after the producer's. Let me know if it makes better sense now.

A quick note before I forgot: We had a chat offline and Sebastian convinced me that the needed synchronization still happens (on the producer stream in `toDlpack()`), so it is safe.

cupy/_core/core.pyx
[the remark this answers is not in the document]
seberg: I am resolving this after discussing with Leo in sync.  Probably just repeating arguments from before, but:
* If we synchronize on a consumer passed stream, then we gain nothing: synchronizing with `stream=None` will achieve the same thing.  (all operations happen on the producer stream).
* If the user passes in a stream explicitly, the data should be safe to use on that stream (whether that makes sense or not).  Existing logic ensures that.

So, I think allowing `stream!=None` is only useful, if no synchronization happens, which is the current state.  Allowing that may be unnecessarily complicated, admittedly.

leofang (approved):

cupy/_core/core.pyx
```diff
@@ -285,8 +285,42 @@ cdef class _ndarray_base:
...
+                # consider here.
+                raise BufferError("copy=True only supported for copy to CPU.")
+        elif dl_device == (dlpack.kDLCPU, 0):
+            # The user explicitly requested CPU device export.
+            # NOTE:
+            # * We effectively ignore the stream here for now!
+            # * We implement it by copying to NumPy, but we must indicate
+            #   the copy, so will construct the dlpack ourselves.
+            if copy is False and (dlpack.get_dlpack_device(self).device_type
+                                  != dlpack.kDLCUDAManaged):
```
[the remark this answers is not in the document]
seberg: This would presumably be quite a bit nicer, but deferring it, since I think this PR is going to be backported, while 8442 is probably not.

cupy/_core/dlpack.pyx
jakirkham: Think this was resolved with the addition of

* `IMPL_VER_MAJOR`
* `IMPL_VER_MINOR`

Should we resolve this thread?

If there are still things to discuss, presumably it would be on those `IMPL_VER_*` values

leofang: Yeah unfortunately I can't resolve the threads (that I created!) myself, and it's been making re-reviews a bit challenging... https://github.com/cupy/cupy/pull/8683#issuecomment-2440480623

kmaehashi (approved):

kmaehashi:
CIs passed, merging! Thanks for working on this one, Sebastian and Leo!

PR #8685, opened 2024-10-21
Fix script name of backport reminder

kmaehashi:
Follows up #8675, the script name mismatch with the name specified in the workflow...


kmaehashi:
Let me self merge this.

PR #8692, opened 2024-10-23
Use `functools.cache` and `pytest.warns`

### LINKED ISSUE #8691: Replace `testing.assert_warns` with `pytest.warns`
### MEMO

- Not planning to remove `testing.assert_warns` itself for now, because it is a part of public APIs.

EarlMilktea:
Closes #8691 .

### Changes

- Replace `_util.memoize()` (no per-device cache option) with `functools.cache`
  - This resolves circular import between `_util.pyx` and `device.pyx` 
- Replace `testing.assert_warns` with `pytest.warns`

### DIFF
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -1,3 +1,4 @@
+import functools
 import string
 import warnings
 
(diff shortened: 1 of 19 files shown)


kmaehashi:
Cache created `cupy.memoize()` are cleared through `cupy.clear_memo()` so I'd prefer keeping as-is.
https://docs.cupy.dev/en/latest/reference/generated/cupy.clear_memo.html#cupy-clear-memo

EarlMilktea:
I see, but still, we need to break the loop between `_util.pyx` and `device.pyx` since `import` reordering leads to visible error (import fails).
Let me work on this problem in another (more specific) PR.

PR #8705, opened 2024-10-28
Resolve import loop

EarlMilktea:
This PR resolves the circular import between `cupy/_util.pyx` and `cupy/cuda/device.pyx`.

### DIFF
--- a/cupy/_util.pyx
+++ b/cupy/_util.pyx
@@ -6,7 +6,7 @@ import os
 import warnings
 
 import cupy
-from cupy.cuda cimport device
+from cupy_backends.cuda.api import runtime
 
 
 DEF CYTHON_BUILD_VER = CUPY_CYTHON_VERSION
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thanks, LGTM!

kmaehashi (dismissed):

cupy/_util.pyx
```diff
@@ -6,7 +6,7 @@ import os
 import warnings
 
 import cupy
-from cupy.cuda cimport device
+from cupy_backends.cuda.api import runtime
```
kmaehashi, line 9:
```suggestion
from cupy_backends.cuda.api cimport runtime
```

kmaehashi (approved):

kmaehashi:
The current codebase does not impose the problem, but this circular import may cause an error depending on the import order, so I'm supportive to the fix.

kmaehashi:
CI failures are unrelated.

Issue #8336, opened 2024-05-23
how to properly import cupyx submodules


ev-br:
### Description

Are cupyx.scipy submodules meant to be available if 
- the top-level `cupyx` name is imported
- the `cupyx.scipy` name is imported
- the `cupyx.scipy.<submodule>`  name is imported 

The latter seems to be the case ATM (with `special` being, uhm, special, see below), and the question is what's the actual API ? 

The origin of the question is https://github.com/scipy/scipy/pull/20772 where we're having to programmatically `importlib.import_module(f"cupyx.scipy{submodule}")`.

### To Reproduce

```
In [1]: import cupyx
   ...: getattr(cupyx.scipy.special, 'gammainc', None)
Out[1]: <ufunc 'cupyx_scipy_gammainc'>

In [2]: import cupyx
   ...: getattr(cupyx.scipy.signal, 'convolve', None)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 2
      1 import cupyx
----> 2 getattr(cupyx.scipy.signal, 'convolve', None)

AttributeError: module 'cupyx.scipy' has no attribute 'signal'

In [3]: import cupyx
   ...: getattr(cupyx.scipy.linalg, 'solve', None)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], line 2
      1 import cupyx
----> 2 getattr(cupyx.scipy.linalg, 'solve', None)

AttributeError: module 'cupyx.scipy' has no attribute 'linalg'

In [4]: import cupyx
   ...: getattr(cupyx.scipy.ndimage, 'correlate', None)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 2
      1 import cupyx
----> 2 getattr(cupyx.scipy.ndimage, 'correlate', None)

AttributeError: module 'cupyx.scipy' has no attribute 'ndimage'
```

### Installation

None

### Environment

```
# Paste the output here
```

### Additional Information

Of course this is not a bug report. Am asking here not on gitter mainly to have a public record to refer to going forward.

kmaehashi:
Ideally I think it's nice to mimic the behavior of SciPy at this point too. IIRC `import scipy` will allow access to all submodules while actual module loading is performed lazily?

lucascolley:
> IIRC `import scipy` will allow access to all submodules while actual module loading is performed lazily?

yep 👍

PR #8706, opened 2024-10-28
Make `cupyx.scipy` submodule imports lazy

ev-br:
closes https://github.com/cupy/cupy/issues/8336

Two changes:

First, formerly, `import cupyx` eagerly imported `cupyx.scipy.fft`, `linalg` and `scpecial`. This led to strange effects downstream: `import cupyx` made `cupyx.scipy.special` available but not `cupyx.scipy.ndimage`.
Fix this by moving corresponding imports from the module level to call sites.

Second, implement a module `__getattr__` for `cupyx.scipy`, so that for all cupyx.scipy submodules the following works:

```
>>> import cupyx
>>> cupyx.scipy.ndimage.convolve
<function convolve at 0x7f4445220c20>
```

### DIFF
--- a/cupy/sparse/__init__.py
+++ b/cupy/sparse/__init__.py
@@ -1,5 +1,5 @@
-import sys
 import warnings
+
 
 def __getattr__(name):
     import cupyx.scipy.sparse
@@ -8,4 +8,4 @@ def __getattr__(name):
         warnings.warn(msg, DeprecationWarning)
         return getattr(cupyx.scipy.sparse, name)
     raise AttributeError(
-         "module 'cupy.sparse' has no attribute {!r}".format(name))
+        "module 'cupy.sparse' has no attribute {!r}".format(name))
--- a/cupy/sparse/__init__.py
+++ b/cupy/sparse/__init__.py
@@ -1,22 +1,11 @@
 import sys
 import warnings
 
-import cupyx.scipy.sparse
-
-
-# Raise a `DeprecationWarning` for `cupy.sparse` submodule when its functions
-# are called. We could raise the warning on importing the submodule, but we
-# use module level `__getattr__` function here as the submodule is also
-# imported in cupy/__init__.py. Unfortunately, module level `__getattr__` is
-# supported on Python 3.7 and higher, so we need to keep the explicit import
-# list for older Python.
-if (3, 7) <= sys.version_info:
-    def __getattr__(name):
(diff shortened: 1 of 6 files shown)


ev-br:
A gentle ping. While this change is of course totally not critical, it will be helpful for downstream (scipy, at least) to simplify the internals of delegation to `cupyx.scipy`, https://github.com/scipy/scipy/blob/main/scipy/ndimage/_support_alternative_backends.py#L41

cupy/sparse/__init__.py
leofang, line 1: @kmaehashi I am thinking we should just remove this trampoline module entirely, now that we're crossing the major version boundary and that this has been deprecated for a long time. WDYT?

cupy/testing/_helper.py
```diff
@@ -186,6 +184,9 @@ def shaped_sparse_random(
         The sparse matrix with given shape, array module,
     """
     import scipy.sparse
+    import cupyx.scipy.sparse
+    if sp is None:
```
leofang, line 190: nit:
```suggestion
    if sp is None:
        import cupyx.scipy.sparse
```

ev-br: Indeed, thank you.

ev-br:
Rebased on main to resolve the conflict.

leofang (approved):

ev-br:
Pushed an update with an attempt to fix the missing import, could you restrart the CI please?

leofang:
@ev-br it seems the CI failures are real issues, could you take a look? 🙂

cupy/testing/_helper.py
```diff
@@ -187,6 +185,9 @@ def shaped_sparse_random(
         The sparse matrix with given shape, array module,
     """
     import scipy.sparse
+    if sp is None:
+        import cupyx.scipy.sparse
```
seberg:
```suggestion
    import cupyx.scipy.sparse

    if sp is None:
```

seberg:
```suggestion
    import cupyx.scipy.sparse

    if sp is None:
```
It's needed below because there is a check for `sp == cupyx.scipy.sparse`.  I am not 100% sure it is the only error, but I think it may be (with different Python versions spelling it slightly differently).

seberg:
/test mini

EDIT: Just to note, above change fixed the test failure.

PR #8726, opened 2024-11-07
Fix `cupy.percentile` for NumPy 2.x

asi1024:
Part of https://github.com/cupy/cupy/issues/8306.

### DIFF
--- a/cupy/_statistics/order.py
+++ b/cupy/_statistics/order.py
@@ -1,5 +1,7 @@
 import warnings
 
+import numpy
+
 import cupy
 from cupy import _core
 from cupy._core import _routines_statistics as _statistics
@@ -178,6 +180,9 @@ def _quantile_unchecked(a, q, axis=None, out=None,
                         overwrite_input=False,
                         method='linear',
                         keepdims=False):
+    dtype = cupy.result_type(a, q)
+    q = cupy.asarray(q)
+
     if q.ndim == 0:
         q = q[None]
         zerod = True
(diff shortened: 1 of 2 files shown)


cupy/_statistics/order.py
```diff
@@ -323,9 +327,9 @@ def percentile(a, q, axis=None, out=None,
     if interpolation is not None:
         method = _check_interpolation_as_method(
             method, interpolation, 'percentile')
-    if not isinstance(q, cupy.ndarray):
-        q = cupy.asarray(q, dtype='d')
-    q = cupy.true_divide(q, 100)
+    if isinstance(q, (tuple, list)):
+        q = numpy.asarray(q)
+    q = q / 100
     if not _quantile_is_valid(q):  # synchronize
```
kmaehashi, line 333:
```suggestion
    if not _quantile_is_valid(q):
```

cupy/_statistics/order.py
```diff
@@ -370,8 +374,8 @@ def quantile(a, q, axis=None, out=None,
     if interpolation is not None:
         method = _check_interpolation_as_method(
             method, interpolation, 'quantile')
-    if not isinstance(q, cupy.ndarray):
-        q = cupy.asarray(q, dtype='d')
+    if isinstance(q, (tuple, list)):
+        q = numpy.asarray(q)
     if not _quantile_is_valid(q):  # synchronize
```
kmaehashi, line 379:
```suggestion
    if not _quantile_is_valid(q):
```

cupy/_statistics/order.py
```diff
@@ -323,9 +327,9 @@ def percentile(a, q, axis=None, out=None,
     if interpolation is not None:
         method = _check_interpolation_as_method(
             method, interpolation, 'percentile')
-    if not isinstance(q, cupy.ndarray):
-        q = cupy.asarray(q, dtype='d')
-    q = cupy.true_divide(q, 100)
+    if isinstance(q, (tuple, list)):
```
kmaehashi, line 330: I discussed offline about the change with @asi1024 
* This PR also improves the logic to avoid synchronization when `q` is given as `numpy.ndarray` or Python scalar (`list`/`tuple`)
* `float` is intentionally excluded here to compute the correct output dtype in `_quantile_unchecked`

PR #8735, opened 2024-11-13
Prevent cython from generating `numpy_allocator.h`

EarlMilktea:
This PR replaces `cdef public` with `cdef extern`.

### DIFF
--- a/cupy/_core/numpy_allocator.pyx
+++ b/cupy/_core/numpy_allocator.pyx
@@ -28,7 +28,7 @@ ELSE:
 DEF ALIGNMENT = 512
 
 
-cdef public void* _calloc(size_t nmemb, size_t size) nogil:
+cdef extern void* _calloc(size_t nmemb, size_t size) nogil:
     errno.errno = 0
     cdef void* buf = aligned_alloc(ALIGNMENT, nmemb * size)
     if buf and errno.errno == 0:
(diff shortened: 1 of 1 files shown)


leofang:
Hi @EarlMilktea I am not sure if this makes sense, but perhaps I can explain the intention from #8442 first, and then you can tell me if I missed anything 🙂 

First of all, we own these functions; they are not defined somewhere else, so technically `cdef extern` is a bit confusing. But more importantly, we need them to be exposed with C linkage, so that we can load these symbols via `ctypes` or other means (ex: `dlopen`) by (unmangled) name, ex: 
https://github.com/cupy/cupy/blob/0714d176d3098587cf1518cdf4510ecd93e0c295/tests/conftest.py#L93-L96
In order to ensure we have C instead of C++ linkage, we used [`CYTHON_EXTERN_C`](https://cython.readthedocs.io/en/stable/src/userguide/external_C_code.html#c-public-declarations), which I am not sure if it'd still work after changing these to `cdef extern`. 

Maybe we can start by confirming the symbols are still exposed under the same names, and then add a simple test coverage? If this change can still expose the symbols just fine, I agree we should not need to generate/package a header.

EarlMilktea:
@leofang (CC: @kmaehashi )

Thank you for pointing out the problems!
How about closing this PR and adding the file to `.gitignore` for better DX for now, as the role of the file is much more important and complex than I expected?

kmaehashi:
> I thought `cdef extern` can be used to expose symbols but will double check.

The symbol is exported using `cdef extern` but as C++ linkage, which is not what we want. I'll add it to gitignore in #8813.

Issue #8747, opened 2024-11-15
`cupy.array_api.linalg.vecdot` deviates from the standard Array API specification


Saransh-cpp:
### Description

`cupy.array_api.linalg.vecdot` does not behave the same as `np.linalg.vecdot` or `jax.numpy.linalg.vecdot` (or even `cupy.dot`) even when identical arguments are passed to these functions. The `cupy.array_api` function throws a shape error (`ValueError`) in the example below, whereas the other implementations don't.

I am aware that there are subtle differences between NumPy and CuPy, so I am not sure if this is a bug or if this is intended. It looks like a bug as passing the same arguments to `cupy.dot` works, but `cupy.array_api.linalg.vecdot(..., axis=1)` (should behave as `dot`) doesn't.

## Array API compatible NumPy implementation (works)
```py
import numpy as np

rng = np.random.default_rng(seed=42)

a = rng.standard_normal((100, 20))
b = rng.standard_normal((100,))

_, n = a.shape
x = np.zeros(n)

w = np.linalg.vecdot(b - a @ x, a, axis=0)  # works
```

## Array API compatible JAX implementation (works)
```py
import numpy as np
import jax.numpy as jnp

rng = np.random.default_rng(seed=42)

a = rng.standard_normal((100, 20))
b = rng.standard_normal((100,))

a_j = jnp.array(a)
b_j = jnp.array(b)

_, n_j = a_j.shape
x_j = jnp.zeros(n_j)

w_j = jnp.linalg.vecdot(b_j - a_j @ x_j, a_j, axis=0)  # works
```

## Array API non-compatible CuPy (works)

Using `dot` instead of `linalg.vecdot(..., axis=1)` as the latter does not exist -

```py
import numpy as np
import cupy

rng = np.random.default_rng(seed=42)

a = rng.standard_normal((100, 20))
b = rng.standard_normal((100,))

a_cu = cupy.asarray(a)
b_cu = cupy.asarray(b)

_, n_cu = a_cu.shape
x_cu = cupy.zeros(n_cu)

w_cu = cupy.dot(b_cu - a_cu @ x_cu, a_cu)  # works
```

## Array API compatible CuPy (does not work)
```py
import numpy as np
import cupy.array_api as cp

rng = np.random.default_rng(seed=42)

a = rng.standard_normal((100, 20))
b = rng.standard_normal((100,))

a_c = cp.asarray(a)
b_c = cp.asarray(b)

_, n_c = a_c.shape
[...]

PR #8750, opened 2024-11-17
Add notice about `cupy.array_api` removal

### LINKED ISSUE #8470: Clarify the future of the `cupy.array_api ` submodule and the `array_api_compat.cupy` alternative
### Description

https://docs.cupy.dev/en/stable/reference/array_api.html recommends the use of the experimental `cupy.array_api` submodule as a mirror to the `numpy.array_api` experimental module.

In NumPy, the `numpy.array_api` experimental submodule introduced in 1.6 (or 1.5, I am not sure) has been removed in 2.0 as the top-level `numpy` module itself is expected to be array API compliant.

In scikit-learn, we test against both `cupy.array_api` and `cupy` directly via the `array_api_compat.cupy` wrapper. However `cupy.array_api` is not fully compliant with the latest version of the standard and we might need to invest efforts in maintaining our own wrapper (for instance to add the `clip` function). However we would rather not maintain a compat wrapper for something that is meant go away in the future.

If cupy is to mirror numpy, then I would also expect the `cupy.array_api` submodule to go away in the future. If this is the case, I think it would be worth updating the doc to state this explicitly and instead point to `array_api_compat.cupy` as a pragmatic way for early array API users / consuming libraries to address the remaining compliance issues with the top level `cupy` module.

kmaehashi:
context: https://github.com/cupy/cupy/issues/8470#issuecomment-2285383328.

Closes #8470 and #8747.

cc/ @leofang

### DIFF
--- a/docs/source/reference/array_api.rst
+++ b/docs/source/reference/array_api.rst
@@ -1,6 +1,11 @@
 Python Array API Support
 ========================
 
+.. note::
+   ``cupy.array_api`` will be removed in CuPy v14 because its NumPy counterpart ``numpy.array_api`` has been removed.
+   The root module ``cupy.*`` is now compatible with the Array API specification as it mirrors the NumPy v2 API.
+   Use the `Array API compatibility library <https://data-apis.org/array-api-compat/>`_ to develop applications compatible with various array libraries, including CuPy, NumPy, and PyTorch.
+
 The `Python array API standard <https://data-apis.org/array-api/2021.12/>`_ aims to provide a coherent set of
 APIs for array and tensor libraries developed by the community to build upon. This solves the API fragmentation
 issue across the community by offering concrete function signatures, semantics and scopes of coverage, enabling
--- a/docs/source/reference/array_api_array.rst
+++ b/docs/source/reference/array_api_array.rst
@@ -1,6 +1,11 @@
 Array API Compliant Object
 ==========================
 
+.. note::
+   ``cupy.array_api`` will be removed in CuPy v14 because its NumPy counterpart ``numpy.array_api`` has been removed.
+   The root module ``cupy.*`` is now compatible with the Array API specification as it mirrors the NumPy v2 API.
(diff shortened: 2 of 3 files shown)


docs/source/reference/array_api.rst
```diff
@@ -1,6 +1,11 @@
 Python Array API Support
 ========================
 
+.. note::
+   ``cupy.array_api`` will be removed in CuPy v14 because its NumPy counterpart ``numpy.array_api`` has been removed.
+   The root module ``cupy.*`` is now compatible with the Array API specification as it mirrors the NumPy v2 API.
```
Saransh-cpp, line 6:
```suggestion
   The root module ``cupy.*`` is now compatible with the Array API specification as it mirrors the NumPy v2.1 API.
```

docs/source/reference/array_api_array.rst
```diff
@@ -1,6 +1,11 @@
 Array API Compliant Object
 ==========================
 
+.. note::
+   ``cupy.array_api`` will be removed in CuPy v14 because its NumPy counterpart ``numpy.array_api`` has been removed.
+   The root module ``cupy.*`` is now compatible with the Array API specification as it mirrors the NumPy v2 API.
```
Saransh-cpp, line 6:
```suggestion
   The root module ``cupy.*`` is now compatible with the Array API specification as it mirrors the NumPy v2.1 API.
```

Saransh-cpp (commented):
Thanks for adding this in! I am not aware of the features going into CuPy `v14`, but should this say `v2.1`? NumPy started supporting Array API `2023.12` from `v2.1`. Please ignore if `v2` is the right version here.

leofang:
No, supporting array API has been one of the goals of NumPy 2(.0), and it's been delivered
https://numpy.org/doc/stable/release/2.0.0-notes.html#highlights

asi1024 (approved):

PR #8766, opened 2024-11-26
Add NumPy 2.x CI for Windows

asi1024:

### DIFF
--- a/cupy/_functional/vectorize.py
+++ b/cupy/_functional/vectorize.py
@@ -6,13 +6,10 @@ from cupyx.jit import _cuda_types
 
 
 def _get_input_type(arg):
-    if isinstance(arg, int):
-        return 'l'
-    if isinstance(arg, float):
-        return 'd'
-    if isinstance(arg, complex):
-        return 'D'
-    return arg.dtype.char
+    if hasattr(arg, 'dtype'):
+        return arg.dtype.char
+    else:
+        return numpy.dtype(type(arg)).char
 
 
 class vectorize(object):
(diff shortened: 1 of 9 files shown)


cupy/_functional/vectorize.py
```diff
@@ -6,13 +6,10 @@
...
-        return 'l'
-    if isinstance(arg, float):
-        return 'd'
-    if isinstance(arg, complex):
-        return 'D'
-    return arg.dtype.char
+    if hasattr(arg, 'dtype'):
+        return arg.dtype.char
+    else:
+        return numpy.dtype(type(arg)).char
```
kmaehashi, line 12: This line seems like a slower replacement of the existing one?

cupy/random/_generator.py
```diff
@@ -212,12 +212,14 @@ def geometric(self, p, size=None, dtype=int):
         p = cupy.asarray(p)
         if size is None:
             size = p.shape
+        if dtype is None:
+            dtype = numpy.dtype('l')
```
kmaehashi, line 216: Can we avoid creating a new dtype instance everytime? (or use `numpy.int_`?)

cupy/_functional/vectorize.py
asi1024: That returns `l` on Linux, `q` on Widnows.

cupy/random/_generator.py
asi1024: Replaced with `numpy.long`. (`numpy.int_` is `q` on Windows)

kmaehashi (approved):

asi1024:
cupy.win.cuda112 timeout seems unrelated.

PR #8785, opened 2024-12-03
Disable typecheck with NumPy v1 in CI

asi1024:


kmaehashi:
Some tests looks still failing :eyes:

asi1024:
To run tests with `numpy._set_promotion_state(‘weak’)` disabled in NumPy 1.x, we need to skip not only typecheck but also many tests related to OverflowError. We will close this pull request.

PR #8792, opened 2024-12-06
Trim whitespace using `pre-commit`

EarlMilktea:
### Changes

- add `trim-whitespace` pre-commit hook
- slightly adjust `.pfnci/generate.py` to incorporate the change


kmaehashi:
Thanks, but several files in our repository, such as `cupyx/scipy/ndimage/cuda/pba_kernels_2d.h` `cupy/_core/include/` are vendored so we'd avoid modifying them.

Issue #8806, opened 2024-12-13
Bug Report: `quantile` and `percentile` crash with `keepdims` and `axis`


AnonymousPlayer2000:
### Description

`quantile` and `percentile` crash with `keepdims` and `axis`

#### CuPy
```py
import cupy as xp

a1 = xp.array([[1., 2.]])
a2 = xp.quantile(a1, 0.5, axis=0, keepdims=True)
print(a2)
```
Output:
```py
Traceback (most recent call last):
  File "/test.py", line 4, in <module>
    a2 = xp.quantile(a1, 0.5, axis=0, keepdims=True)
  File "/usr/local/lib/python3.10/dist-packages/cupy/_statistics/order.py", line 379, in quantile
    return _quantile_unchecked(
  File "/usr/local/lib/python3.10/dist-packages/cupy/_statistics/order.py", line 194, in _quantile_unchecked
    for ax in axis:
TypeError: 'int' object is not iterable
```
#### NumPy
```py
import numpy as xp

a1 = xp.array([[1., 2.]])
a2 = xp.quantile(a1, 0.5, axis=0, keepdims=True)
print(a2)
```
Output:
```py
[[1. 2.]]
```

#### CuPy
```py
import cupy as xp

a1 = xp.array([[1., 2.]])
a2 = xp.percentile(a1, 10, axis=0, keepdims=True)
print(a2)
```
Output:
```py
Traceback (most recent call last):
  File "/test_p.py", line 4, in <module>
    a2 = xp.percentile(a1, 10, axis=0, keepdims=True)
  File "/usr/local/lib/python3.10/dist-packages/cupy/_statistics/order.py", line 333, in percentile
    return _quantile_unchecked(
  File "/usr/local/lib/python3.10/dist-packages/cupy/_statistics/order.py", line 194, in _quantile_unchecked
    for ax in axis:
TypeError: 'int' object is not iterable
```
#### NumPy
```py
import numpy as xp

a1 = xp.array([[1., 2.]])
a2 = xp.percentile(a1, 10, axis=0, keepdims=True)
print(a2)
```
Output:
```py
[[1. 2.]]
```

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
OS                           : Linux-6.8.0-49-generic-x86_64-with-glibc2.35
Python Version               : 3.10.12
CuPy Version                 : 13.3.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.1.0
[...]

PR #8808, opened 2024-12-13
Fix crashes of quantile and percentile

AnonymousPlayer2000:
Closes https://github.com/cupy/cupy/issues/8806. Convert axis to a tuple before iterating over it if it is int


asi1024 (approved):
LGTM!

PR #8810, opened 2024-12-13
Switch from `.A` attribute to `.todense()` method

jakirkham:
The sparse matrix `.A` attribute is deprecated in SciPy (and [removed in 1.14.0]( https://docs.scipy.org/doc/scipy-1.14.1/release/1.14.0-notes.html#expired-deprecations )) and is sometimes used with NumPy & SciPy in the tests. So swap it out for the `.todense()` method, which is not deprecated. The behavior and results should be the same. This just avoids deprecation warnings or errors from SciPy.

xref: https://github.com/scipy/scipy/pull/14822
xref: https://github.com/scipy/scipy/pull/20499

### DIFF
--- a/cupyx/scipy/sparse/linalg/_norm.py
+++ b/cupyx/scipy/sparse/linalg/_norm.py
@@ -89,9 +89,9 @@ def norm(x, ord=None, axis=None):
             raise ValueError('Invalid axis %r for an array with shape %r' %
                              (axis, x.shape))
         if ord == numpy.inf:
-            return abs(x).max(axis=a).A.ravel()
+            return abs(x).max(axis=a).todense().ravel()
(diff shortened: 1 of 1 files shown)


jakirkham:
Noticed some other code uses `toarray()` ( for example: https://github.com/cupy/cupy/pull/8508 )

Happy to switch to that if preferred

kmaehashi:
Thanks John! `toarray` sounds better as CuPy does not have `cp.matrix`.

jakirkham:
Thanks Kenichi! 🙏

Sounds good. Included the change.

kmaehashi (approved):
Thanks, LGTM!

jakirkham:
Thanks Kenichi! 🙏

Issue #8497, opened 2024-08-15
[RFC] Removing NumPy Fallback Mode in CuPy v14


kmaehashi:
The [NumPy fallback mode](https://github.com/cupy/cupy/tree/main/cupyx/fallback_mode) in CuPy was first introduced in CuPy v7 (2019), aiming to ease the porting of NumPy-based code to CuPy by providing a mechanism to automatically delegate "missing" APIs in CuPy to NumPy. However, now that we have [covered](https://docs.cupy.dev/en/latest/reference/comparison.html) almost all the NumPy API, I think this layer is no longer playing its role.

As far as I can tell from the [GitHub search resutls](https://github.com/search?q=%22cupyx%22+%22fallback_mode%22+language%3APython&type=code&l=Python), no public code is relying on `cupyx.fallback_mode`, so I'm thinking of removing the feature in CuPy v14. Please let us know here if you have use-cases relying on the fallback mode.

kmaehashi:
Let's proceed as we haven't received any negative feedbacks.

jcoffi:
I know this is years old. But I can tell you, that if more people knew about it, they would have used it. I use it all the time. I use 'from cupyx.fallback_mode import numpy as np' instead of import numpy as np everywhere on my private projects.

kmaehashi:
Hi @jcoffi, thank you for providing the feedback. We recognize that our efforts to promote the fallback mode may have fallen short, and we really appreciate hearing from users like you.

We're sorry for any disruption this change may cause, but we hope you will continue to leverage `import cupy` namespace directly. If you find any lacking features you need within the main namespace, please don't hesitate to share your feedback (majority of them are tracked in #6078).

PR #8816, opened 2024-12-15
Remove fallback mode

asi1024:
Close https://github.com/cupy/cupy/issues/8497.
Removes `cupyx.fallback` in CuPy v14.


kmaehashi:
Thanks! How about removing it in v14 as we don't see any negative responses in #8497?

asi1024:
I agree. Fixed to remove all `cupyx.fallback` module.

kmaehashi:
Looks good, thanks!

kmaehashi (approved):

PR #8817, opened 2024-12-16
Skip some tests incompatible with NumPy 2.2

asi1024:
Temporary skip for failing tests.

### DIFF
--- a/cupy/_manipulation/add_remove.py
+++ b/cupy/_manipulation/add_remove.py
@@ -155,10 +155,10 @@ def trim_zeros(filt, trim='fb'):
     .. seealso:: :func:`numpy.trim_zeros`
 
     """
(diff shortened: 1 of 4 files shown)


cupy/_manipulation/add_remove.py
```diff
@@ -155,10 +155,10 @@ def trim_zeros(filt, trim='fb'):
     .. seealso:: :func:`numpy.trim_zeros`
 
     """
+    if filt.ndim == 0:
+        return filt
```
kmaehashi, line 159: Seems it requires a copy:

```
>>> x = np.array(0)
>>> id(x)
140510775702896
>>> id(np.trim_zeros(np.array(0)))
140510776653360
```

PR #8832, opened 2024-12-17
CI: support "skip-ci" label

kmaehashi:
This PR introduces "skip-ci" label support, which omits triggering full CIs when merging a PR.

PR #8840, opened 2024-12-18
CI: Fix FlexCI compatibility

kmaehashi:
Follow-up #8832. FlexCI's pull_request event support should not be used in this case.


kmaehashi:
Let me self-merge and test if the FlexCI trigger as expected.

PR #8843, opened 2024-12-18
Fix sparse matrix indexing to raise IndexError for shape mismatch

asi1024:
Fix for SciPy 1.13.

### DIFF
--- a/cupyx/scipy/sparse/_index.py
+++ b/cupyx/scipy/sparse/_index.py
@@ -454,7 +454,7 @@ class IndexMixin(object):
 
     def _parse_indices(self, key):
         M, N = self.shape
-        row, col = _unpack_index(key)
+        row, col = _unpack_index(key, self.shape)
 
         if self._is_scalar(row):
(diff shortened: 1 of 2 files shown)


cupyx/scipy/sparse/_index.py
```diff
@@ -569,7 +569,18 @@ def _try_is_scipy_spmatrix(index):
...
 
 
-def _unpack_index(index):
+def _check_mask_shape(mask, shape):
+    if mask is None:
+        return
+    assert mask.dtype == numpy.bool_
+    if mask.shape != shape:
+        raise IndexError(
+            f"boolean column index has incorrect length: "
```
lijinf2, line 578: 'column' does not look very accurate when the function is called on row vector, such as  `_check_mask_shape(bool_row, (shape[0],))`. Maybe replace 'column' with 'array'?

lijinf2 (approved):
A nit. 
You need to address the conflict with upstream main.

PR #8845, opened 2024-12-18
Fix handling of pinned memory

### LINKED ISSUE #8789: Lifetime hazards from asynchronous H2D copies
### Description

The documentation for `cp.array` says

> blocking ([bool](https://docs.python.org/3/library/functions.html#bool)) – Default is False, meaning if a H2D copy is needed it would run asynchronously on the current stream, and users are responsible for ensuring the stream order. For example, writing to the source obj without proper ordering while copying would result in a race condition. If set to True, the copy is synchronous (with respect to the host).

However, this doesn't mention that the function doesn't keep the source obj alive while the asynchronous copy happens. So even if you don't modify the data through `obj`, simply releasing the last reference to `obj` can cause it to be garbage collected, and the memory can then be reallocated from the memory pool (or from a lower level, if not using a pool).

This is mostly a risk when `obj` is backed by pinned memory (which is where I got bitten by this, and spent a few days chasing the race condition). Looking at the source code, when the `obj` is pinned, a temporary transfer buffer is allocated (`_alloc_async_transfer_buffer`) and the data is synchronously copied to that buffer, in which case there is no risk. However, if allocation of the transfer buffer fails, the fallback path directly does an asynchronous copy from the source data, and the risk remains, so this is not purely something people allocating their own pinned memory need to be concerned about.

_No response_

kmaehashi:
Closes #8789.

- Keep reference to source when doing async H2D copy to avoid lifetime hazard.
- Check for event completion in `free_all_blocks` to free more blocks that are no longer in use.

### DIFF
--- a/cupy/cuda/pinned_memory.pyx
+++ b/cupy/cuda/pinned_memory.pyx
@@ -330,6 +330,7 @@ cdef class PinnedMemoryPool:
 
     cpdef free_all_blocks(self):
         """Release free all blocks."""
+        _watcher.check_and_release()
         rlock.lock_fastrlock(self._lock, -1, True)
         try:
             self._free.clear()
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -2797,6 +2797,7 @@ cdef _ndarray_base _array_default(
     cdef intptr_t ptr_h = <intptr_t>(a_cpu.ctypes.data)
     if pinned_memory.is_memory_pinned(ptr_h):
         a.data.copy_from_host_async(ptr_h, nbytes, stream)
+        pinned_memory._add_to_watch_list(stream.record(), a_cpu)
     else:
         # The input numpy array does not live on pinned memory, so we allocate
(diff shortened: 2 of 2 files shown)


cupy/cuda/pinned_memory.pyx
```diff
@@ -330,6 +330,7 @@ cdef class PinnedMemoryPool:
 
     cpdef free_all_blocks(self):
         """Release free all blocks."""
+        _watcher.check_and_release()
```
leofang, line 333: Q: it seems this method also needs to acquire the rlock. How about calling `_check_and_release_without_lock()` in the try block below?

kmaehashi: These two locks are the different things. Both Watcher and PinnedMemoryPointer has each lock instance to protect their attributes.

leofang: Ah, I see! Thanks for pointing it out.

leofang (approved):

asi1024 (approved):

PR #8846, opened 2024-12-19
Use a custom Min/Max instead of specializing CUB

bernhardmgruber:
In similar spirit as #8446, users should not specialize `cub::Min` and `cub::Max`. We recently changed their definition to be aliases to `cuda::minimum` and `cuda::maximum`, which will break cupy when it upgrades to CCCL 2.8. This PR provides a fix for that.

* Avoid specializing cub::Min/Max
* Use cuda::minimum/maximum when available

### DIFF
--- a/cupy/cuda/cupy_cub.cu
+++ b/cupy/cuda/cupy_cub.cu
@@ -10,6 +10,8 @@
 #include <cub/device/device_histogram.cuh>
 #include <cub/iterator/counting_input_iterator.cuh>
 #include <cub/iterator/transform_input_iterator.cuh>
+#include <cuda/functional>
+#include <cuda/std/functional>
 #else
 #include <hipcub/device/device_reduce.hpp>
 #include <hipcub/device/device_segmented_reduce.hpp>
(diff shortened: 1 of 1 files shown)


cupy/cuda/cupy_cub.cu
```diff
@@ -10,6 +10,8 @@
 #include <cub/device/device_histogram.cuh>
 #include <cub/iterator/counting_input_iterator.cuh>
 #include <cub/iterator/transform_input_iterator.cuh>
+#include <cuda/functional>
+#include <cuda/std/functional>
```
leofang, line 14: Q: Do we need both?

leofang (approved):

cupy/cuda/cupy_cub.cu
bernhardmgruber: Yes. `<cuda/functional>` for `cuda::maximum<>` etc. and `<cuda/std/functional>` for `cuda::std::multiplies<>`.

asi1024 (approved):

asi1024:

Issue #8681, opened 2024-10-21
Document the arguments of `cupyx.jit.rawkernel`


kmaehashi:
https://github.com/cupy/cupy/issues/8674#issuecomment-2421431598

PR #8857, opened 2025-01-01
Documented the arguments of cupyx.jit.rawkernel 

averagestud:
Documentation for the issue: #8681 
```py
cupyx.jit.rawkernel(*, mode='cuda', device=False)
```
**Documentation added**
mode ('numpy' or 'cuda'): The rule for typecast.
device (int or cupy.cuda.Device): Index of the device to be used. Be careful that the device ID (a.k.a. GPU ID) is zero origin. The current device is selected by default.

### DIFF
--- a/cupyx/jit/_interface.py
+++ b/cupyx/jit/_interface.py
@@ -169,6 +169,11 @@ class _JitRawKernel:
 
 def rawkernel(*, mode='cuda', device=False):
     """A decorator compiles a Python function into CUDA kernel.
+    Args:
+        mode ('numpy' or 'cuda'): The rule for typecast.
+        device (int or cupy.cuda.Device): Index of the device to be used. 
+        Be careful that the device ID (a.k.a. GPU ID) is zero origin. The 
+        current device is selected by default.
     """
     cupy._util.experimental('cupyx.jit.rawkernel')
 


cupyx/jit/_interface.py
```diff
@@ -169,6 +169,11 @@ def cached_code(self):
 
 def rawkernel(*, mode='cuda', device=False):
     """A decorator compiles a Python function into CUDA kernel.
+    Args:
+        mode ('numpy' or 'cuda'): The rule for typecast.
+        device (int or cupy.cuda.Device): Index of the device to be used.
+        Be careful that the device ID (a.k.a. GPU ID) is zero origin.
+        The current device is selected by default.
```
leofang, line 176: Content aside, this formatting is incorrect. The 2nd and further lines should be indented.
```suggestion
        device (int or cupy.cuda.Device): Index of the device to be used.
            Be careful that the device ID (a.k.a. GPU ID) is zero origin.
            The current device is selected by default.
```
During PR review, a preview of the doc updates can be checked in the CI, e.g. https://cupy--8857.org.readthedocs.build/en/8857/reference/generated/cupyx.jit.rawkernel.html. I suggest to build the docs locally to confirm the changes, before pushing to a PR branch.

leofang: `device` here does not mean that this decorator takes a `Device` object or a device ID, but rather indicating this is a device function, not a kernel.

leofang (changes requested):
Thanks for your contribution! I left some comments.

kmaehashi:
Let me close as this PR is inactive for a while. Feel free to request reopening if necessary.

PR #8871, opened 2025-01-10
Handling of scalar ()-shaped out parameter in quantile and percentile #8827

RakshithaKowlikar:
- Fixed an issue where scalar `out` caused errors by directly assigning values when both `out` and `q` are scalars.
- Added a check to raise an error if `out` has fewer dimensions than needed.


asi1024:
Thanks for your contribution! Could you add tests in  `tests/cupy_tests/statistics_tests/test_order.py`?

kmaehashi:

Issue #8866, opened 2025-01-07
Support SciPy 1.15


asi1024:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [ ] Read Release Notes.
- [ ] Fix code and CI to support the new version.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - **Update Requirements**: For NumPy/SciPy, bump the pinned minimum/maximum version in `setup.py` (see #4773.)
- [ ] Backport the above PR.
- [ ] For NumPy update, reflect the changes in `numpy.array_api` to `cupy.array_api`.
- [ ] Backport the above PR.
- [ ] Fix documentation.
  - Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) (note: the baseline version must be updated only in the development branch).
  - Update `docs/requirements.txt` to use the latest versions.
- [ ] Backport the above PR.
- [ ] Apply fixes needed to follow the specification changes.

asi1024:
- [x] tests/cupyx_tests/scipy_tests/fft_tests/test_fftlog.py
  * https://github.com/cupy/cupy/pull/8883
- [x] tests/cupyx_tests/scipy_tests/interpolate_tests/test_interpnd.py
  * #8887
  * DeprecationWarning: `scipy.interpolate.interpnd.estimate_gradients_2d_global` is deprecated along with the `scipy.interpolate.interpnd` namespace. `scipy.interpolate.interpnd.estimate_gradients_2d_global` will be removed in SciPy 1.16.0, and the `scipy.interpolate.interpnd` namespace will be removed in SciPy 2.0.0.
  * https://github.com/scipy/scipy/pull/21754
- [x] tests/cupyx_tests/scipy_tests/linalg_tests/test_special_matrices.py
  * https://github.com/cupy/cupy/pull/8885
  * DeprecationWarning: `kron` has been deprecated in favour of `numpy.kron` in SciPy 1.15.0 and will be removed in SciPy 1.17.0.
  * https://github.com/scipy/scipy/issues/20077
- [x] tests/cupyx_tests/scipy_tests/signal_tests/test_bsplines.py
  * A value mismatch occurs in one test case.
[...]

PR #8877, opened 2025-01-20
Raise ValueError upon attempts to create 3-dim sparse array

asi1024:


kmaehashi (approved):

kmaehashi:
/test mini

Backport should be test only.

PR #8880, opened 2025-01-20
Disable typecheck for CSR sparse matrix norm

asi1024:

In SciPy 1.15, the result of the matrix norm of CSR sparse matrix is returned in Python primitive float, which is different from the behaviour in SciPy 1.14 and earlier. However, to avoid device-to-host synchronization, CuPy should continue to return a 0-dim `cupy.ndarray`.

### DIFF
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
@@ -80,9 +80,9 @@ class TestLsqr(unittest.TestCase):
 @testing.with_requires('scipy')
 class TestMatrixNorm:
 
-    @testing.numpy_cupy_allclose(rtol=1e-3, atol=1e-4, sp_name='sp',
-                                 accept_error=(ValueError,
-                                               NotImplementedError))
+    @testing.numpy_cupy_allclose(
+        rtol=1e-3, atol=1e-4, sp_name='sp', type_check=False,
+        accept_error=(ValueError, NotImplementedError))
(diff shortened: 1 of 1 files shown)


kmaehashi:
This itself looks good to me but I suspect this is an unexpected change in SciPy v1.15? cc/ @ev-br

ev-br:
This is one other case where I would ask @dschult to please weight in.

asi1024:
Sorry, The PR description lacks ample detail, so I am including some additional information here. In SciPy 1.15.1, `scipy.sparse.linalg.norm` **occasionally** returns a Python primitive float value. Below is a sample code that reproduces this behavior.

```py
>>> import numpy
>>> import scipy
>>> a = (numpy.arange(9) - 4.0).reshape(3, 3)
>>> x = scipy.sparse.csr_matrix(a)
>>> type(scipy.sparse.linalg.norm(x))
<class 'numpy.float64'>
>>> type(scipy.sparse.linalg.norm(x, ord=numpy.inf))
<class 'float'>
```

dschult:
This should be fixed in scipy/scipy#22372 which has been merged and flagged for backporting to the next big-fix release.  Thanks very much!

asi1024:
@dschult Thank you for addressing the issue promptly!

leofang:
@asi1024 @dschult thanks -- so do we still need this PR, or is it moot now?

dschult:
The issue on the SciPy side is fixed as indicated above by [scipy/scipy#22372](https://github.com/scipy/scipy/pull/22372).

PR #8881, opened 2025-01-20
[WIP] Support matmul for sparse matrix and 3 or more dim dense array

asi1024:

### DIFF
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_csc.py
@@ -843,12 +843,13 @@ class TestCscMatrixScipyComparison:
         x = xp.arange(8).reshape(4, 2).astype(self.dtype)
         return m * x
 
-    def test_mul_dense_ndim3(self):
-        for xp, sp in ((numpy, scipy.sparse), (cupy, sparse)):
-            m = self.make(xp, sp, self.dtype)
-            x = xp.arange(24).reshape(4, 2, 3).astype(self.dtype)
-            with pytest.raises(ValueError):
-                m * x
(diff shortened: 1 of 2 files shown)


asi1024:
In SciPy 1.15, `csr_matrix.__mul__(dense_array)` returns object-dtype ndarray as follows:

```py
>>> import numpy
>>> import scipy
>>> x = scipy.sparse.csr_matrix((3, 4))
>>> y = numpy.zeros((4, 5, 6))
>>> out = x * y
>>> out.shape
(4, 5, 6)
>>> out.dtype
dtype('O')
>>> out[0][0][0]
<Compressed Sparse Row sparse matrix of dtype 'float64'
        with 0 stored elements and shape (3, 4)>
>>> out[0][0][0].dtype
dtype('float64')
>>> out[0][0][0].shape
(3, 4)
```

I suspect that due to the fix of `csr_matrix.__mul__` to return `NotImplemented` instead of raising `TypeError` in this case, `ndarray.__rmul__` is treating `csr_matrix` as an object, consequently computing scalar_object * (4, 5, 6)-shape ndarray. Is this expected behaviour? cc/ @ev-br

dschult:
Indeed, (and Thank You!!)
Returning `NotImplemented` inside `__mu__` in the >2d case is causing this trouble (and other troubles scipy/scipy#22347). 

I'll put together a fix with a backport. Not sure how long it will be until the next bugfix release.

ev-br:
Sadly, scipy 1.5.2 is likely to be a thing, and is likely sooner than later, I guess. This one + ndimage rank_filter fix or revert, maybe something else on top.

dschult:
Scipy 1.15.2 has been released with a fix for this!

asi1024:
@dschult Thank you very much! I confirmed that the tests pass without the fixes in this pull request!

PR #8899, opened 2025-01-23
Support CUDA 12.8 + Blackwell GPUs (sm_100, sm_120)

leofang:
Public release of CUDA Toolkit 12.7 was skipped, only CUDA driver 12.7 was released.

Issue #6783, opened 2022-06-14
Support aarch64 in library installer / wheel preloads


kmaehashi:

leofang:
With #9204 I suppose this is no longer needed (the library installer will be removed). As for library loading, it's been supported using the pathfinder. Closing.

PR #8914, opened 2025-01-30
Bump library installers for CUDA 12.8

kmaehashi:
Part of #8903. xref https://github.com/cupy/cupy/pull/8899#issuecomment-2621647610


jakirkham:
Thanks Kenichi! 🙏

Would it be possible to add this to our backports for 13.4.0?

kmaehashi:
Yes, I'll backport this to v13.

jakirkham:
Thanks again for working on this Kenichi! 🙏

This looks reasonable. Think everything is up on our end now

Do we need to do anything for ARM?

Also does this get tested on CI?

kmaehashi:
> Do we need to do anything for ARM?

Nothing to do for now, because support for these libraries are currently disabled in ARM binary packages (#6783).

> Also does this get tested on CI?

Yes.

/test mini

asi1024 (approved):

PR #8907, opened 2025-01-24
Fix boxcox_llf for SciPy 1.13

xuefeng-xu:
Follow up #8095

Details see https://github.com/scipy/scipy/pull/20021

PR #8919, opened 2025-02-03
Use CCCL 2.8.x branch + Use `CUPY_CACHE_KEY` in hash keys

### LINKED ISSUE #8588: Incorporate CCCL version in cache hash?
Follow-up of #8412. 

Currently the same CCCL version is used at both build- and run- time, which is great for consistency. However, one missing step is every time the CCCL version is updated, we need a way to invalidate the kernel cache. A challenge in the cache invalidation, however, is we don't know which headers are actually used by the compiler, and to make it robust we might need to pay some run-time cost in attempting to mimic the compiler behavior, parse, and hash the used headers.

I am thinking we could start with something simpler, at the price of potentially more aggressive invalidation than needed. We can generate a hash at build time by extending the current `CUPY_CACHE_KEY` coverage to also cover the CCCL headers:
https://github.com/cupy/cupy/blob/9506efc49080c76a9662485e20cf7d95cd03394c/install/cupy_builder/_context.py#L60-L75
Since the same CCCL version is also used at run-time, this hash is valid. Then, we can dump this hash in, say, a json file that's shipped as part of the CuPy wheels and read in `cupy/cuda/_compiler.py` first time compiling a kernel in the process, and use it as part of the kernel cache key. This is considered aggressive because it might be possible that the used headers are unchanged in a newer CCCL version but the cache is still invalidated because the (global) hash is changed.

leofang:
Follow-up of #8899. Close #8588.

This PR switches to use the [2.8.x](https://github.com/NVIDIA/cccl/tree/branch/2.8.x) branch, as discussed in https://github.com/cupy/cupy/pull/8899#issuecomment-2626174099. Currently I manually backported https://github.com/NVIDIA/cccl/pull/3644 on top of it, using the existing `cupy/cccl` fork. If it gets done tomorrow we can switch back.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "third_party/cccl"]
 	path = third_party/cccl
-	url = https://github.com/NVIDIA/cccl.git
+	url = https://github.com/cupy/cccl.git
 [submodule "third_party/jitify"]
 	path = third_party/jitify
 	url = https://github.com/NVIDIA/jitify.git
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -60,6 +60,20 @@ from cupy.exceptions import ComplexWarning
 NUMPY_1x = numpy.__version__ < '2'
 
 
+cdef extern from *:
+    """
+    #define _str_(s) #s
+    #define _xstr_(s) _str_(s)
+    const char* cupy_cache_key = _xstr_(CUPY_CACHE_KEY);
+    #undef _xstr_
+    #undef _str_
+    """
+    const char* cupy_cache_key  # set at build time
+
+
+CUPY_CACHE_KEY = bytes(cupy_cache_key).decode()
+
+
 # If rop of cupy.ndarray is called, cupy's op is the last chance.
 # If op of cupy.ndarray is called and the `other` is cupy.ndarray, too,
 # it is safe to call cupy's op.
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -155,6 +155,12 @@ def _get_nvrtc_version():
     return _nvrtc_version
 
 
+@_util.memoize()
+def _get_cupy_cache_key():
+    from cupy._core import core
+    return core.CUPY_CACHE_KEY
+
+
 # Known archs for Tegra/Jetson/Xavier/etc
 _tegra_archs = ('32', '53', '62', '72', '87')
 
@@ -580,7 +586,8 @@ def _compile_with_cache_cuda(
         base = _preprocess('', options, arch, backend)
         _empty_file_preprocess_cache[env] = base
 
-    key_src = '%s %s %s %s' % (env, base, source, extra_source)
+    key_src = '%s %s %s %s %s' % (
+        env, base, source, extra_source, _get_cupy_cache_key())
     key_src = key_src.encode('utf-8')
     name = _hash_hexdigest(key_src) + '.cubin'
 
--- a/tests/cupyx_tests/scipy_tests/stats_tests/test_morestats.py
+++ b/tests/cupyx_tests/scipy_tests/stats_tests/test_morestats.py
@@ -149,7 +149,7 @@ class TestBoxcox_llf:
         not sys.platform.startswith('linux'),
         reason="Return value of scipy.stats.boxcox_llf has large error")
     @testing.with_requires('scipy>=1.13')
-    @testing.numpy_cupy_allclose(scipy_name='scp', atol=atol, rtol=rtol)
+    @testing.numpy_cupy_allclose(scipy_name='scp', atol=2e-12, rtol=rtol)
     def test_instability_around_zero(self, xp, scp):
         data = xp.asarray([2003, 1950, 1997, 2000, 2009], dtype=numpy.float64)
         return scp.stats.boxcox_llf(1e-8, data)
--- a/.pfnci/coverage.rst
+++ b/.pfnci/coverage.rst
@@ -3125,7 +3125,7 @@ CuPy CI Test Coverage
      - 
    * - 
      - cutensor
-     - 🚨
+     - 2
      - 
      - 
      - 
@@ -3146,8 +3146,8 @@ CuPy CI Test Coverage
      - 
      - 
      - 
-     - 
-     - 
+     - ✅
+     - ✅
      - 
      - 
      - 
@@ -3207,7 +3207,7 @@ CuPy CI Test Coverage
      - 
    * - 
      - cutensor,cub
-     - 28
+     - 26
      - 
      - 
      - 
@@ -3228,8 +3228,8 @@ CuPy CI Test Coverage
      - ✅
      - ✅
      - ✅
-     - ✅
-     - ✅
+     - 
+     - 
      - ✅
      - ✅
      - ✅
--- a/.pfnci/linux/tests/cuda123.multi.sh
+++ b/.pfnci/linux/tests/cuda123.multi.sh
@@ -9,7 +9,7 @@ ACTIONS="$(dirname $0)/actions"
 
 export NVCC="ccache nvcc"
 
-export CUPY_ACCELERATORS="cutensor,cub"
+export CUPY_ACCELERATORS="cutensor"
 
 "$ACTIONS/build.sh"
 export OMPI_ALLOW_RUN_AS_ROOT=1
--- a/.pfnci/linux/tests/cuda123.sh
+++ b/.pfnci/linux/tests/cuda123.sh
@@ -9,7 +9,7 @@ ACTIONS="$(dirname $0)/actions"
 
 export NVCC="ccache nvcc"
 
-export CUPY_ACCELERATORS="cutensor,cub"
+export CUPY_ACCELERATORS="cutensor"
 
 "$ACTIONS/build.sh"
 "$ACTIONS/unittest.sh" "not slow and not multi_gpu"
--- a/.pfnci/matrix.yaml
+++ b/.pfnci/matrix.yaml
@@ -322,7 +322,7 @@
   mpi4py: null
   cython: "0.29"
   cuda-python: null
-  env:CUPY_ACCELERATORS: "cutensor,cub"
+  env:CUPY_ACCELERATORS: "cutensor"
   test: "unit"
 
 # CUDA 12.3 (Multi-GPU) | Linux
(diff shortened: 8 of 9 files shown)


kmaehashi:
Thanks Leo!

tests/cupyx_tests/scipy_tests/stats_tests/test_morestats.py
```diff
@@ -149,7 +149,7 @@ def test_array_empty_neg_lmb(self, xp, scp, dtype):
         not sys.platform.startswith('linux'),
         reason="Return value of scipy.stats.boxcox_llf has large error")
     @testing.with_requires('scipy>=1.13')
-    @testing.numpy_cupy_allclose(scipy_name='scp', atol=atol, rtol=rtol)
+    @testing.numpy_cupy_allclose(scipy_name='scp', atol=2e-12, rtol=rtol)
```
leofang, line 152: xref: https://github.com/cupy/cupy/pull/8907#issuecomment-2615656310 (tested against SciPy 1.14.1)

cupy/cuda/compiler.py
```diff
@@ -580,7 +586,8 @@ def _compile_with_cache_cuda(
         base = _preprocess('', options, arch, backend)
         _empty_file_preprocess_cache[env] = base
 
-    key_src = '%s %s %s %s' % (env, base, source, extra_source)
+    key_src = '%s %s %s %s %s' % (
+        env, base, source, extra_source, _get_cupy_cache_key())
```
leofang, line 590: Technically we don't need `extra_source` here anymore, because `CUPY_CACHE_KEY` is a superset of it (hash of all 1st-party & 3rd-party headers). However I decided to do minimal work and leave it here for now. Perhaps we could revisit this and clean up the `extra_source` related code later?

leofang:
Re-ran the whole test suite on B100 + CTK 12.8 with this branch. No issues found. Also checked `test_raw.py` on other systems too.

leofang:
Since this PR invalidates the kernel cache, it is no surprise that CIs failed in the first pass due to timeout. I expect all of them will turn green when we run the 2nd time, except for the CUDA 12.3 ones which now show errors everywhere beyond just `test_raw.py` (the kernel cache likely protected the issues from being exposed). 

Based on internal conversation it seems to be a real compiler bug and irrelevant of the CCCL branch choice (`main` or `2.8.x`). Due to the impact scope it is unlikely that we can find a reasonable fix. @kmaehashi is it OK if we document CUDA 12.3 is no longer supported due to compiler bug? If so, should we also detect NVRTC version and warn for 12.3 users?

kmaehashi:
> Since this PR invalidates the kernel cache, it is no surprise that CIs failed in the first pass due to timeout. I expect all of them will turn green when we run the 2nd time, except for the CUDA 12.3 ones which now show errors everywhere beyond just `test_raw.py` (the kernel cache likely protected the issues from being exposed).

:+1:

> Based on internal conversation it seems to be a real compiler bug and irrelevant of the CCCL branch choice (`main` or `2.8.x`). Due to the impact scope it is unlikely that we can find a reasonable fix. @kmaehashi is it OK if we document CUDA 12.3 is no longer supported due to compiler bug? If so, should we also detect NVRTC version and warn for 12.3 users?

Is it still the case even if we bump from 12.3.0 (which we use in CI docker image) to 12.3.1 or 12.3.2?, Or, will disabling CUB accelerator by default for CUDA 12.3 fixes the situation?

leofang:
> Is it still the case even if we bump from 12.3.0 (which we use in CI docker image) to 12.3.1 or 12.3.2?, Or, will disabling CUB accelerator by default for CUDA 12.3 fixes the situation?

Interesting ideas! I found a machine with 12.3.2 pre-installed, so gave it a shot. No luck. However, setting `CUPY_ACCELERATORS=` (empty) does allow `test_raw.py` to finish completely. How do I configure the CI to disable CUB accelerator?

kmaehashi:
`CUPY_ACCELERATORS` is specified here:
https://github.com/cupy/cupy/blob/a111a1eecdfd9490ab64d4048511aa38c9d54d71/.pfnci/matrix.yaml#L325

Then do `./generate.py` to update the [test script](https://github.com/cupy/cupy/blob/a111a1eecdfd9490ab64d4048511aa38c9d54d71/.pfnci/linux/tests/cuda123.sh#L12).

leofang:
Thanks! How about Windows? I noticed the Windows config is not touched.

leofang:
> Thanks! How about Windows? I noticed the Windows config is not touched.

I made a change to `.pfnci/windows/_flexci.ps1` to set `CUPY_ACCELERATORS` explicitly.

cupy/_core/core.pyx
```diff
@@ -60,6 +60,20 @@ from cupy.exceptions import ComplexWarning
...
+    #define _str_(s) #s
+    #define _xstr_(s) _str_(s)
+    const char* cupy_cache_key = _xstr_(CUPY_CACHE_KEY);
+    #undef _xstr_
+    #undef _str_
+    """
+    const char* cupy_cache_key  # set at build time
+
+
+CUPY_CACHE_KEY = bytes(cupy_cache_key).decode()
```
jakirkham, line 74: Cython knows how to decode directly from `const char*` so it is possible to skip copying to `bytes`

```suggestion
CUPY_CACHE_KEY = cupy_cache_key.decode()
```

cupy/_core/core.pyx
```diff
@@ -60,6 +60,20 @@ from cupy.exceptions import ComplexWarning
 NUMPY_1x = numpy.__version__ < '2'
 
 
+cdef extern from *:
+    """
+    #define _str_(s) #s
+    #define _xstr_(s) _str_(s)
+    const char* cupy_cache_key = _xstr_(CUPY_CACHE_KEY);
+    #undef _xstr_
+    #undef _str_
```
jakirkham, line 69: Am curious why all the indirection is needed here?

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "third_party/cccl"]
 	path = third_party/cccl
-	url = https://github.com/NVIDIA/cccl.git
+	url = https://github.com/cupy/cccl.git
```
jakirkham, line 3: Is the plan to switch to a friendly fork of CCCL or is this just temporary until the fixes we need are upstreamed?

jakirkham (commented):
Thanks Leo! 🙏

Had a couple questions/suggestions below

cupy/_core/core.pyx
leofang: `CUPY_CACHE_KEY` is a compiled-time variable that needs to stringification.

.gitmodules
leofang: In the PR description I wrote
> This PR switches to use the [2.8.x](https://github.com/NVIDIA/cccl/tree/branch/2.8.x) branch, as discussed in https://github.com/cupy/cupy/pull/8899#issuecomment-2626174099. Currently I manually backported https://github.com/NVIDIA/cccl/pull/3644 on top of it, using the existing cupy/cccl fork. If it gets done tomorrow we can switch back.

However, I've been trying the whole day to get the backport https://github.com/NVIDIA/cccl/pull/3653 merged, but ran into various issues, so it literally just got merged. At this point once CI is green here, we should just merge and not touch the submodule to avoid costly re-run (once the submodule is updated the hash key will change, invalidating all kernel caches). 

In case it is not clear, we're running full CI sweep to rebuild the kernel caches and validate the changes, and it is taking a lot longer than usual CI runs. We should wait for another window to get the submodule updated.

cupy/_core/core.pyx
leofang: Thanks, would you mind sending a follow-up PR to address this? I prefer to not re-run the CIs as explained below.

leofang: > Thanks, would you mind sending a follow-up PR to address this? I prefer to not re-run the CIs as [explained below](https://github.com/cupy/cupy/pull/8919#discussion_r1940423577).

kmaehashi:
It seems GCP is congested and instances are getting preempted. Let's merge this one after checking CUDA 12.3 test results. I've canceled other CIs to push them forward.

leofang:
> It seems GCP is congested and instances are getting preempted. Let's merge this one after checking CUDA 12.3 test results. I've canceled other CIs to push them forward.

@kmaehashi I have a question. For those already completed, it seems the tests still timed out? Are the kernel caches not updated in the case of timeout? 😱 It's been a while since last time we burned the caches so I forgot the behavior... If so we may have to increase the timeout limit and retry...

Good news is cupy.linux.cuda123.multi passed, and cupy.linux.cuda123 ran 5 hrs (instead of hitting 500 errors and returning early). I'm looking into the failures in the later and see how we can skip those tests (some of them are `cupyx.jit.cub` tests so we know they would not work).

kmaehashi:
Ah yes... cache is not uploaded until the code is merged to avoid pollution. I believe this PR itself is OK so let's merge and let the cache grow. 

> Good news is cupy.linux.cuda123.multi passed, and cupy.linux.cuda123 ran 5 hrs (instead of hitting 500 errors and returning early). I'm looking into the failures in the later and see how we can skip those tests (some of them are `cupyx.jit.cub` tests so we know they would not work).

According to the list of failed tests (https://ci.preferred.jp/cupy.linux.cuda123/184670/#L68145), `cupyx_tests/jit_tests/test_cooperative_groups.py` looks like `#include <cuda/barrier>` issue, and others seems CUB-related ones.

leofang:
@kmaehashi, a quick question: @miscco deciphered the encrypted error message with NVRTC 12.3, and confirmed that it's a compiler bug that could be worked around (https://github.com/NVIDIA/cccl/pull/3666). Would you like me to include his patch so that we can restore CUB support for 12.3, or you think it's OK for now?

cupy/_core/core.pyx
jakirkham: Done in PR: https://github.com/cupy/cupy/pull/8923

kmaehashi:
That's great! Could you push the workaround to our fork?

jakirkham:
Looks like it was added recently: https://github.com/NVIDIA/cccl/commit/7229e0bb8aeb1e49a16bad87405a655c73384361

leofang:
Right, so should we use the fork or just https://github.com/NVIDIA/cccl/commit/3a388b7b01512d48474b98389a3e776c8d8f817a (from the latest 2.8.x branch)?

jakirkham:
Put together a PR to update: https://github.com/cupy/cupy/pull/8924

Please let me know what you think

PR #8923, opened 2025-02-04
Drop unneeded `bytes` copy of `CUPY_CACHE_KEY`

jakirkham:
Cython knows how to [`decode` a `const char*` directly to `str`]( https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#decoding-bytes-to-text ) without needing to create a `bytes` object. So this drops the use of `bytes` here when constructing the Python value of `CUPY_CACHE_KEY`.

xref: https://github.com/cupy/cupy/pull/8919#discussion_r1940412605

cc @leofang

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -71,7 +71,7 @@ cdef extern from *:
     const char* cupy_cache_key  # set at build time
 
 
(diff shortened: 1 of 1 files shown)


leofang:
Thanks, John 🙏

PR #8950, opened 2025-02-13
restore CUB histogram and bincount

grlee77:
This MR reverts #7708 now that it seems CuPy bundles a recent version of CUB from CCCL instead of relying on a version shipped with CUDA.

One test case was added to cover the original issue reported in #7698

### DIFF
--- a/cupy/_statistics/histogram.py
+++ b/cupy/_statistics/histogram.py
@@ -5,6 +5,10 @@ import numpy
 
 import cupy
 from cupy import _core
+from cupy._core import _accelerator
+from cupy.cuda import cub
+from cupy.cuda import common
+from cupy.cuda import runtime
 
 
 # rename builtin range for use in functions that take a range argument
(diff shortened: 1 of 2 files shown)


jakirkham:
@kmaehashi could you please review?

Also would it be alright if we backport this too?

kmaehashi (approved):

jakirkham:
Found a similar bug in NumPy with a NumPy array. Filed as issue: https://github.com/numpy/numpy/issues/28354

cc @seberg (for awareness)

seberg:
Right, in practice, NumPy allows this type of cast in a few places (e.g. indexing, where it is technically unsafe).  For `bincount` it even seems fully safe, so diverging is perfectly fine, IMO.

jakirkham:
Thanks Sebastian! 🙏

jakirkham:
Looks like this is passing! 🥳

PR #8954, opened 2025-02-14
ndimage: fix bug in morphology `axes` support when `structure=None`

grlee77:
There was a bug in the recently added axes support for some morphology functions. The default structuring element in that case must have `ndim=len(axes)` and rather than the full image dimension.

The tests were updated to also run cases with `structure=None`

### DIFF
--- a/cupyx/scipy/ndimage/_morphology.py
+++ b/cupyx/scipy/ndimage/_morphology.py
@@ -363,7 +363,8 @@ def binary_erosion(input, structure=None, iterations=1, mask=None, output=None,
 
     .. seealso:: :func:`scipy.ndimage.binary_erosion`
     """
-    structure, _, _ = _prep_structure(structure, input.ndim)
+    axes = _util._check_axes(axes, input.ndim)
+    structure, _, _ = _prep_structure(structure, len(axes))
     return _binary_erosion(input, structure, iterations, mask, output,
                            border_value, origin, 0, brute_force, axes=axes)
 
@@ -415,9 +416,9 @@ def binary_dilation(input, structure=None, iterations=1, mask=None,
 
     .. seealso:: :func:`scipy.ndimage.binary_dilation`
     """
-    structure, structure_shape, symmetric = _prep_structure(structure,
-                                                            input.ndim)
     axes = _util._check_axes(axes, input.ndim)
+    structure, structure_shape, symmetric = _prep_structure(structure,
+                                                            len(axes))
     origin = _util._fix_sequence_arg(origin, len(axes), 'origin', int)
     # no point in flipping if already symmetric
     if not symmetric:
@@ -480,7 +481,8 @@ def binary_opening(input, structure=None, iterations=1, output=None, origin=0,
 
     .. seealso:: :func:`scipy.ndimage.binary_opening`
     """
-    structure, _, _ = _prep_structure(structure, input.ndim)
+    axes = _util._check_axes(axes, input.ndim)
+    structure, _, _ = _prep_structure(structure, len(axes))
     tmp = binary_erosion(input, structure, iterations, mask, None,
                          border_value, origin, brute_force, axes=axes)
     return binary_dilation(tmp, structure, iterations, mask, output,
@@ -537,7 +539,8 @@ def binary_closing(input, structure=None, iterations=1, output=None, origin=0,
 
     .. seealso:: :func:`scipy.ndimage.binary_closing`
     """
-    structure, _, _ = _prep_structure(structure, input.ndim)
+    axes = _util._check_axes(axes, input.ndim)
+    structure, _, _ = _prep_structure(structure, len(axes))
     tmp = binary_dilation(input, structure, iterations, mask, None,
                           border_value, origin, brute_force, axes=axes)
     return binary_erosion(tmp, structure, iterations, mask, output,
--- a/cupyx/scipy/ndimage/_morphology.py
+++ b/cupyx/scipy/ndimage/_morphology.py
@@ -587,12 +587,12 @@ def binary_hit_or_miss(input, structure1=None, structure2=None, output=None,
 
     .. seealso:: :func:`scipy.ndimage.binary_hit_or_miss`
     """
-    if structure1 is None:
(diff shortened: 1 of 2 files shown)


jakirkham:
Think it would be good to backport this one

`structure=None` is the default and it doesn't work currently

asi1024 (dismissed):

asi1024:
Thank you for the fix! I have added the backport label.

asi1024:
@grlee77 Some tests seem to fail with the following error when run with SciPy 1.15. Could you check this?
`RuntimeError: structure and input must have same dimensionality`

cupyx/scipy/ndimage/_morphology.py
```diff
@@ -363,7 +363,8 @@ def binary_erosion(input, structure=None, iterations=1, mask=None, output=None,
 
     .. seealso:: :func:`scipy.ndimage.binary_erosion`
     """
-    structure, _, _ = _prep_structure(structure, input.ndim)
+    axes = _util._check_axes(axes, input.ndim)
+    structure, _, _ = _prep_structure(structure, len(axes))
```
jakirkham, line 367: Couldn't there be fewer `axes` than `input.ndim`? If so, should we use `input.ndim` here?

```suggestion
    structure, _, _ = _prep_structure(structure, input.ndim)
```

cupyx/scipy/ndimage/_morphology.py
```diff
@@ -415,9 +416,9 @@ def binary_dilation(input, structure=None, iterations=1, mask=None,
 
     .. seealso:: :func:`scipy.ndimage.binary_dilation`
     """
-    structure, structure_shape, symmetric = _prep_structure(structure,
-                                                            input.ndim)
     axes = _util._check_axes(axes, input.ndim)
+    structure, structure_shape, symmetric = _prep_structure(structure,
+                                                            len(axes))
```
jakirkham, line 421:
```suggestion
                                                            input.ndim)
```

cupyx/scipy/ndimage/_morphology.py
```diff
@@ -480,7 +481,8 @@ def binary_opening(input, structure=None, iterations=1, output=None, origin=0,
 
     .. seealso:: :func:`scipy.ndimage.binary_opening`
     """
-    structure, _, _ = _prep_structure(structure, input.ndim)
+    axes = _util._check_axes(axes, input.ndim)
+    structure, _, _ = _prep_structure(structure, len(axes))
```
jakirkham, line 485:
```suggestion
    structure, _, _ = _prep_structure(structure, input.ndim)
```

cupyx/scipy/ndimage/_morphology.py
```diff
@@ -537,7 +539,8 @@ def binary_closing(input, structure=None, iterations=1, output=None, origin=0,
 
     .. seealso:: :func:`scipy.ndimage.binary_closing`
     """
-    structure, _, _ = _prep_structure(structure, input.ndim)
+    axes = _util._check_axes(axes, input.ndim)
+    structure, _, _ = _prep_structure(structure, len(axes))
```
jakirkham, line 543:
```suggestion
    structure, _, _ = _prep_structure(structure, input.ndim)
```

jakirkham (commented):
Thanks Greg! 🙏

Had a question below

cupyx/scipy/ndimage/_morphology.py
grlee77: The intent is for the user-provided structure to have <= input.ndim dimensions (matching the subset of axes being filtered). CuPy is supposed to handle inserting any necessary singleton dimensions so that `ndim` axes are eventually present by the time the kernel is called.

jakirkham: How will it know where to add the singleton dimensions without providing `input.ndim` and `axes`?

cupyx/scipy/ndimage/_morphology.py
grlee77: All of these should stay as is. I had missed one case of changing `input.ndim -> len(axes)` in `binary_hit_or_miss` (now fixed in bf2647c)

grlee77:
> @grlee77 Some tests seem to fail with the following error when run with SciPy 1.15. Could you check this?
RuntimeError: structure and input must have same dimensionality

Thanks @kmaehashi , there was still a bug in generation of the default `structure1` in `binary_hit_or_miss`. It is fixed by the latest commit.

cupyx/scipy/ndimage/_morphology.py
jakirkham: Let's keep this conversation in one thread for simplicity

Given the discussion that has already happened above, let's stick with this one: https://github.com/cupy/cupy/pull/8954#discussion_r1969297034

Please take a look at the question there

cupyx/scipy/ndimage/_morphology.py
grlee77: `_prep_structure` just prepares the structure with a shape matching the number of filtered axes (which will be <= ndim). The actual code inserting the extra dimensions was already present in a `_utils.expand_footprint` utility function used in `_binary_erosion` here:
https://github.com/cupy/cupy/blob/418a9169bc9994574a22e6bfb07edb39d91899c6/cupyx/scipy/ndimage/_morphology.py#L189-L193

or for the grayscale morphology filters it is here:
https://github.com/cupy/cupy/blob/418a9169bc9994574a22e6bfb07edb39d91899c6/cupyx/scipy/ndimage/_filters.py#L933

jakirkham (approved):
Thanks Greg! 🙏

jakirkham:
@asi1024 when you have a moment, could you please take another look? 🙂

asi1024:
I confirmed the tests also pass with SciPy 1.15. Thank you for your contribution!

asi1024 (approved):

grlee77:
@jakirkham : it looks like we don't need to backport this as `axes` support for morphology functions is a new 14.0 feature (it is not in the 13.x branch)

PR #8177, opened 2024-02-12
Add `apply_over_axes` API

syheliel:
Hello, the community! This PR is to contribute a little step forward #6078. It's my first time to get involved in `cupy` :)

implement Code is from: https://github.com/numpy/numpy/blob/8f7e4c57931569760a036599ade6955ecc387c84/numpy/lib/_shape_base_impl.py#L413

Test code is from: https://github.com/numpy/numpy/blob/8f7e4c57931569760a036599ade6955ecc387c84/numpy/lib/tests/test_shape_base.py#L274

PR #8955, opened 2025-02-14
Implemented apply over axes function and successfully tested using py…

O-sama12:
## Add apply_over_axes function to CuPy
- Implemented `apply_over_axes`, similar to NumPy’s version.
- Registered the function inside `_manipulation/__init__.py`.
- Added test cases to `test_dims.py` for verification.
- Ensured compatibility with NumPy output using `cupy.allclose`.

**Tests passed ✅**

### DIFF
--- a/cupy/_manipulation/__init__.py
+++ b/cupy/_manipulation/__init__.py
@@ -1,2 +1,3 @@
 # Functions from the following NumPy document
 # https://numpy.org/doc/stable/reference/routines.array-manipulation.html
+from cupy._manipulation.dims import apply_over_axes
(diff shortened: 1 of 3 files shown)


kmaehashi:
Thanks for contribution!

* When porting a NumPy functions please refer to the NumPy's implementation and port it to CuPy. https://github.com/numpy/numpy/blob/e20317a43d3714f9085ad959f68c1ba6bc998fcd/numpy/lib/_shape_base_impl.py#L420
* For `apply_over_axes` please place the function under `cupy/lib/_shape_base.py`, mirroring the NumPy's directory structure.

kmaehashi:
Hi @O-sama12, but after discussing with @asi1024 we found that there is an ongoing pull request for `apply_over_axes` at #8177 which is almost ready to merge. I'm afraid that I couldn't point that at the first round of the review. :bow:

PR #8957, opened 2025-02-14
Fix buffer protocol to raise TypeError when it is not meant to be supported

leofang:
In Python it seems `TypeError` is expected:
```python
>>> memoryview(123)
TypeError: memoryview: a bytes-like object is required, not 'int'
```
which was discovered by @ZzEeKkAa offline. cc @gmarkall for vis


leofang:
This PR does not need to be backported, because the support only exists in the main branch.

kmaehashi (approved):

kmaehashi:
xref #8442

PR #8960, opened 2025-02-15
[v13] Regenerate coverage matrix

kmaehashi:
Fix the static-check CI failure. The coverage table became outdated due to merge order of pull requests.


kmaehashi:
@asi1024 Let me self-merge as this is trivial but blocking all the other v13 PRs.

PR #8963, opened 2025-02-16
Applying funct in shape base.py

O-sama12:
Successfully removed the apply_over_axes from dims.py and apply it in _shape_base.py
Registering apply_over_axes under __init__.py located in lib folder.
Test case passed.


kmaehashi:
Please update the original PR #8955 instead of creating a separate PR for the same topic.

PR #8970, opened 2025-02-18
Fix `cupyx.scipy.spatial.distance`'s `cdist` for RAPIDS 24.12 compatibility

jakirkham:
Fixes a bug in `cupyx.scipy.spatial.distance`'s implementation of `cdist`, which relies on RAPIDS. The current implementation was broken with RAPIDS 24.12 due to a bug in RAPIDS. This includes a fix to workaround the RAPIDS 24.12 bug and remain compatible with RAPIDS pre-24.12 and 25.02+. 

Also update RAPIDS CI testing to use released RAPIDS 25.02 Docker images (instead of nightlies).

Note: RAPIDS dropped CUDA 12.5 images and replaced them with CUDA 12.8.

ref: https://docs.rapids.ai/notices/rsn0042/

cc @cjnolet @grlee77

### DIFF
--- a/cupyx/scipy/spatial/distance.py
+++ b/cupyx/scipy/spatial/distance.py
@@ -114,8 +114,8 @@ _METRICS_NAMES = list(_METRICS.keys())
 def check_soft_dependencies():
     if not cuvs_available:
         if not pylibraft_available:
-            raise RuntimeError('cuVS >= 24.12 or pylibraft < '
-                               '24.12 should be installed to use this feature')
+            raise RuntimeError('cuVS >= 25.02 or pylibraft < '
+                               '25.02 should be installed to use this feature')
 
 
 def minkowski(u, v, p):
--- a/cupyx/scipy/spatial/distance.py
+++ b/cupyx/scipy/spatial/distance.py
@@ -628,7 +628,7 @@ def cdist(XA, XB, metric='euclidean', out=None, **kwargs):
             output_arr = out if out is not None else cupy.zeros((mA, mB),
                                                                 dtype=XA.dtype,
                                                                 order=XA_order)
-            pairwise_distance(XA, XB, output_arr, metric, p=p)
+            pairwise_distance(XA, XB, output_arr, metric, p)
             return output_arr
         else:
             raise ValueError('Unknown Distance Metric: %s' % mstr)
--- a/cupyx/scipy/spatial/distance.py
+++ b/cupyx/scipy/spatial/distance.py
@@ -114,8 +114,8 @@ _METRICS_NAMES = list(_METRICS.keys())
 def check_soft_dependencies():
     if not cuvs_available:
         if not pylibraft_available:
-            raise RuntimeError('cuVS >= 25.02 or pylibraft < '
-                               '25.02 should be installed to use this feature')
+            raise RuntimeError('cuVS >= 24.12 or pylibraft < '
+                               '24.12 should be installed to use this feature')
 
 
 def minkowski(u, v, p):
--- a/.pfnci/linux/tests/cuda-rapids.Dockerfile
+++ b/.pfnci/linux/tests/cuda-rapids.Dockerfile
@@ -1,11 +1,11 @@
-ARG BASE_IMAGE="rapidsai/base:25.02a-cuda12.5-py3.11"
+ARG BASE_IMAGE="rapidsai/base:25.02-cuda12.8-py3.11"
 FROM ${BASE_IMAGE}
 
 USER root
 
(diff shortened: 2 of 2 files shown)


jakirkham:
We will want to backport this as well since `v13` has the same issue

https://github.com/cupy/cupy/blob/f5e36bfd179cdddbbf49e1ca8d622665f0170373/cupyx/scipy/spatial/distance.py#L114-L118

jakirkham:
cc @kmaehashi (for awareness)

grlee77:
Good timing John, I had just mentioned this in #8971 😄 

It seems that cuVS 24.12 did not support the `p` kwarg for `cupyx.scipy.spatial.distance` functions:
```
  File "~/src/public/cupy/cupyx/scipy/spatial/distance.py", line 631, in cdist
    pairwise_distance(XA, XB, output_arr, metric, p=p)
  File "resources.pyx", line 110, in cuvs.common.resources.auto_sync_resources.wrapper
  File "~/mambaforge/envs/rapids-24.12-v2/lib/python3.12/site-packages/pylibraft/common/outputs.py", line 83, in wrapper
    ret_value = f(*args, **kwargs)
                ^^^^^^^^^^^^^^^^^^
  File "distance.pyx", line 57, in cuvs.distance.distance.pairwise_distance
```

cupyx/scipy/spatial/distance.py
```diff
@@ -114,8 +114,8 @@ def __init__(self, canonical_name=None, aka=None,
 def check_soft_dependencies():
     if not cuvs_available:
         if not pylibraft_available:
-            raise RuntimeError('cuVS >= 24.12 or pylibraft < '
-                               '24.12 should be installed to use this feature')
+            raise RuntimeError('cuVS >= 25.02 or pylibraft < '
+                               '25.02 should be installed to use this feature')
```
grlee77, line 118: I think pylibraft 24.12 does not have the function, so there may not be a working option for the 24.12 release!

(There is [no `distance` folder in here](https://github.com/rapidsai/raft/tree/branch-24.12/python/pylibraft/pylibraft) in the 24.12 tag, but it is there for 24.10 and before)
                               '25.02 should be installed to use this feature')

grlee77, line 118:
```suggestion
                               '24.12 should be installed to use this feature')
```

jakirkham: Good point Greg. Let's let Corey weigh in on how to handle 24.12

grlee77: right, we might have to explicitly raise an exception if only 24.12 is available

cjnolet:
@jakirkham 

On [this](https://github.com/cupy/cupy/blob/c7fbe8e4592e46df4c557d8a2b3d267d6773ce8c/cupyx/scipy/spatial/distance.py#L631) line, when cuVS 24.12 is used, the argument was changed to `metric_arg` (as [here](https://github.com/rapidsai/cuvs/blob/cf2885c9d0b8a5d839378939a29154a4d165fefe/python/cuvs/cuvs/distance/distance.pyx#L59)) before we changed it back to `p` in 25.02. This was a mistake that went unnoticed in 24.12, but the argument is named `p` in cuVS 25.02+ and in all versions of RAFT.

cupyx/scipy/spatial/distance.py
```diff
@@ -628,7 +628,7 @@ def cdist(XA, XB, metric='euclidean', out=None, **kwargs):
             output_arr = out if out is not None else cupy.zeros((mA, mB),
                                                                 dtype=XA.dtype,
                                                                 order=XA_order)
-            pairwise_distance(XA, XB, output_arr, metric, p=p)
+            pairwise_distance(XA, XB, output_arr, metric, p)
```
jakirkham, line 631: The simplest solution is just to pass this argument positionally. That appears to work across versions

cupyx/scipy/spatial/distance.py
jakirkham: Looks like just reverting to the original error is the right approach. So have done that

cupyx/scipy/spatial/distance.py
cjnolet: Yeah, if this is allowed then it's definitely the path of least resistence

jakirkham:
CI passes 🥳

jakirkham:
Have updated the description and title to hopefully make it clearer what this change fixes

kmaehashi (approved):
Thanks, LGTM!

jakirkham:
Thanks Kenichi! 🙏

PR #8971, opened 2025-02-18
cupyx.scipy.distance: initialize output array with empty instead of zeros

grlee77:
cc @cjnolet , @jakirkham 

This MR is to avoid overhead of explicitly writing zeros into the output array before calling the cuVS/raft `pairwise_distance` function. Unit tests continue to pass when I tried this locally.

Checking in cuVS and raft sources, it seems
- cuVS defaults to [empty-initialized `out` array](https://github.com/rapidsai/cuvs/blob/v25.02.00/python/cuvs/cuvs/distance/distance.pyx)
- at least since RAFT 22.12, it also uses an [empty `out` array](https://github.com/rapidsai/raft/blob/v24.10.00/python/pylibraft/pylibraft/distance/pairwise_distance.pyx#L181) by default.

Given that, I think this is likely safe to merge for CuPy 14.x.

---

Aside from the performance issue, I found that the recommendation of "cuVS>=24.12" does not work for 24.12 (25.02 worked as expected): see 
https://github.com/cupy/cupy/pull/8970#issuecomment-2667149022

### DIFF
--- a/cupyx/scipy/spatial/distance.py
+++ b/cupyx/scipy/spatial/distance.py
@@ -114,7 +114,7 @@ _METRICS_NAMES = list(_METRICS.keys())
 def check_soft_dependencies():
     if not cuvs_available:
         if not pylibraft_available:
-            raise RuntimeError('cuVS >= 24.12 or pylibraft < '
+            raise RuntimeError('cuVS >= 25.02 or pylibraft < '
                                '24.12 should be installed to use this feature')
 
 
--- a/cupyx/scipy/spatial/distance.py
+++ b/cupyx/scipy/spatial/distance.py
@@ -114,7 +114,7 @@ _METRICS_NAMES = list(_METRICS.keys())
 def check_soft_dependencies():
     if not cuvs_available:
         if not pylibraft_available:
-            raise RuntimeError('cuVS >= 25.02 or pylibraft < '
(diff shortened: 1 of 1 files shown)


grlee77:
In conversation with @jakirkham about this earlier today, he mentioned we should bump from 25.02a to 25.02 with cuda 12.8 in this file:
https://github.com/cupy/cupy/blob/main/.pfnci/linux/tests/cuda-rapids.Dockerfile

Let me know if I should change it in this MR

cupyx/scipy/spatial/distance.py
```diff
@@ -114,7 +114,7 @@ def __init__(self, canonical_name=None, aka=None,
 def check_soft_dependencies():
     if not cuvs_available:
         if not pylibraft_available:
-            raise RuntimeError('cuVS >= 24.12 or pylibraft < '
+            raise RuntimeError('cuVS >= 25.02 or pylibraft < '
```
jakirkham, line 117: Went ahead and did this along with CI changes in PR: https://github.com/cupy/cupy/pull/8970

Let's go ahead and drop this from here

```suggestion
            raise RuntimeError('cuVS >= 24.12 or pylibraft < '
```

kmaehashi (approved):
LGTM!

jakirkham:
Ah ok

Thanks Kenichi and Greg! 🙏

PR #8854, opened 2024-12-27
Implement `cupy.linalg.eig` / `cupy.linalg.eigvals`

EarlMilktea:
This closes #3255 and also closes #6359 .

This PR uses `cusolverDnXgeev` to solve non-hermitian eigenvalue problems on GPU.

PR #8980, opened 2025-02-24
Implement cupy.linalg.eig / cupy.linalg.eigvals clone of PR #8854

mfoerste4:
This is PR #8854 with an updated main branch
CC @leofang


leofang:
@mfoerste4 found that the OOM issue is due to unguarded calls to the new cuSOLVER routines, which is only available in recent CUDA 12.x (which one I forgot, it's in release notes and the linked issue). The evidence is CUDA 12.8 CIs passed.

asi1024:
Thank you for taking over the pull request. Considering the `_geev` function is called repeatedly by the `eig` function, I think some redundant calculations can be eliminated.

- My understanding is that `work_device_size` and `work_host_size` return the same value for all `_geev` calls. If that is correct, we could allocate `work_device` and `work_host` in advance and reuse them.
- We can eliminate the copy when cocatenating the return values of `_geev` with `cupy.stack`. Could you fix to allocate the concatenated memory in advance?
- Please avoid repeatedly calling `all(w.imag == 0.0)`. It should be checked against the combined result. Also, instead of using the Python primitive all, rewrite it to use the `cupy.ndarray` method, like `(w.imag == 0.0).all()`.
- Could you convert the input array to Fortran format in advance before calling `_geev`?

leofang:
> We need to add a version guard like what we did in the past for other cuSOLVER routines.

Looks like it's fixed now! (The remaining CI failures are irrelevant.)

> I think some redundant calculations can be eliminated.

@mfoerste4 could you kindly take a look at Akifumi-san's comments?

EarlMilktea:
@mfoerste4 @leofang Sorry for making wait you long (was very busy recently). Please keep me up to date!

cupy/linalg/_eigenvalue.py
```diff
@@ -141,7 +207,53 @@ def eigh(a, UPLO='L'):
         return _syevd(a, UPLO, True)
 
 
-# TODO(okuta): Implement eigvals
+def eig(a):
```
asi1024, line 210: How about transpose the input as follows?
```py
a = a.transpose(0, 2, 1).copy(order='C').transpose(0, 2, 1)
```
The output array can also be constructed in the same way.

asi1024:
@mfoerste4 Could you fix for https://github.com/cupy/cupy/pull/8980#discussion_r1991077589?

asi1024:
@mfoerste4 The output dtype for `_geev` is `complex_dtype` as defined here:
https://github.com/cupy/cupy/blob/dfdf9a37a281e931195dfc3f341667346f7b1d4f/cupy/linalg/_eigenvalue.py#L136
After completing all _geev calls, we can check in a single `all()` call whether all the imaginary parts are zero.

kmaehashi:
@asi1024 kindly ping

ev-br:
It's great to see this, I'm looking forward to be able to use it!

For the dtype of eigenvalues: it can be argued that numpy's decision to have value-dependent dtype is a mistake. Other array libraries (notably, jax.numpy, pytorch) always return eigenvalues as a complex array, and it's up to a user to decide if they want to check for zero imaginary parts.
So maybe CuPy could consider to do the same and avoid numpy's "return w.real if w.imag==0 else w" contraption.
For completeness, here's a link to an Array API discussion: https://github.com/data-apis/array-api/issues/935

kmaehashi:
Thanks @mfoerste4! The change looks good to me, let me kick the CI again.
@asi1024 Do you have any other suggestions on this PR?

/test mini

Issue #8933, opened 2025-02-07
LinearNDInterpolator / Delaunay: Illegal Memory Access


lreisinger:
### Description

Hello!
LinearNDInterpolator raises CUDA_ERROR_ILLEGAL_ADDRESS depending on the input I provide. My use case is interpolating sparse depth images and I get the errors on many but not all images.
For inputs that trigger this error, it crashes instantly or I may successfully run the interpolation in a loop a few times until it finally crashes. Weirdly enough, _sometimes_ it doesn't crash at all. During execution, the used vram stayed at~910mb.

I created a small code example to illustrate the issue using random data where i emulated sparsity using NaNs.

### To Reproduce

```py
import numpy as np
import cupy as cp
from cupyx.scipy.interpolate import LinearNDInterpolator


def interpolate_nans(arr: np.ndarray):
    """
    Parameters:
        arr (np.ndarray): A 2D array of shape (H, W) representing depth values.

    Returns:
        np.array: A 2D array with missing values filled.
    """

    memory_pool = cp.get_default_memory_pool()

    arr = np.ascontiguousarray(arr)
    valid_mask = ~np.isnan(arr)
    valid_mask = np.ascontiguousarray(valid_mask)

    with cp.cuda.using_allocator(memory_pool.malloc):
    # with cp.cuda.using_allocator(None):

        stream = cp.cuda.Stream()
        with stream:
            arr_cp = cp.asarray(arr)
            arr_cp = cp.ascontiguousarray(arr_cp, dtype=cp.float64)

            valid_mask_cp = cp.asarray(valid_mask)
            valid_mask_cp = cp.ascontiguousarray(valid_mask_cp)

            # Coordinates of valid and invalid points
            y_valid, x_valid = cp.nonzero(valid_mask_cp)
            y_invalid, x_invalid = cp.nonzero(~valid_mask_cp)

            # Linear interpolation using valid points
            valid_points = cp.stack((x_valid, y_valid), axis=1)
            valid_points = cp.ascontiguousarray(valid_points, dtype=cp.float64)

            valid_values = arr_cp[y_valid, x_valid]
            valid_values = cp.ascontiguousarray(valid_values, dtype=cp.float64)
            
            invalid_points = cp.stack((x_invalid, y_invalid), axis=1)
            invalid_points = cp.ascontiguousarray(invalid_points, dtype=cp.float64)

            ip = LinearNDInterpolator(valid_points, valid_values, fill_value=cp.nan, rescale=False)
[...]

PR #8983, opened 2025-02-25
Fix illegal memory access in LinearNDInterpolator

asi1024:
Close https://github.com/cupy/cupy/issues/8933.

`nextNearest` can become `-1` when the all opposites of `nearest` are visited. This PR stops the search for the closest points and concludes that there is no triangle covering the query point in such a case.
cc: @andfoy

### DIFF
--- a/cupyx/scipy/spatial/delaunay_2d/_kernels.py
+++ b/cupyx/scipy/spatial/delaunay_2d/_kernels.py
@@ -2948,6 +2948,7 @@ RealType*           coords
                 }
 
                 if(!isInTri) {
+                    if (nextNearest == -1) break;
                     nearest = triOpp[nextNearest];
                     visited[off + 1] = nextNearest;
                 }


hovavalon:
I stumbled upon the same bug, and this fix solves it for me as well.
I am currently applying it locally.
@kmaehashi, when is this fix expected to enter the main branch/stable pre release?

cupyx/scipy/spatial/delaunay_2d/_kernels.py
```diff
@@ -2948,6 +2948,7 @@
                 }
 
                 if(!isInTri) {
+                    if (nextNearest == -1) break;
```
andfoy, line 2951: Makes sense! Thanks a lot for the fix @asi1024

kmaehashi (approved):
Stamped. Thanks for review @andfoy!

kmaehashi:
Backport unnecessary as it's new in v14 (#7985)

PR #8984, opened 2025-02-25
Mark xfails in some spline tests for SciPy 1.15

asi1024:


kmaehashi:
Q: Is this a bug in our side or in SciPy (or unsure)?

asi1024:
@kmaehashi Probably our side, but further investigation is needed. I will add a comment to the test.

PR #8949, opened 2025-02-13
[DNM] Diagnose CI failures on CUDA 12.x (Windows)

kmaehashi:
Experiments:
* Run all tests in full matrix: https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=fd9c2a7788f1dbefff3f4f042b7d4eade1969fe9 -> failed (reproduced the main branch result)
* Limit tests to signaltools only: https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=b022feb028477758764ce36376051111ad70d8d3 -> passed
* Limit tests to signal-related CIs only: https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=a03e398ed4b794ad4383129090dd576705c3bb3e -> passed
* Run all tests with increased host memory, show GPU memory usage (nvidia-smi) before running example: https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=e1819327f57206756d6b159464a3373178914fc8

PR #8992, opened 2025-02-26
CI: Do not run full CI on CUDA 12.0/12.1/12.2 + Windows

kmaehashi:
It seems that on these platforms the compilation is extremely slow that could not complete the full combination tests within reasonable resources.

Here is the latest CI results for the recent commit in the `main` branch: https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=d44993a49f920f6e53812791ad2e85bbc04b76fd
image

I attempted several approaches (#8949 etc.) but couldn't find the root cause. The CUDA 12.3.0 release notes mentions the compiler performance improvement, so I guess this got fixed somehow in the recent versions of CUDA.
https://docs.nvidia.com/cuda/archive/12.3.0/cuda-toolkit-release-notes/index.html#cuda-compilers

> * Improved compile time in some common scenarios:
>   * Extended split compilation to cubin for LTO.
>   * Turned on concurrent NVVM processing by default, with documented fallback to serialized compilation.
>   * Reduced NVRTC compile time for small programs via moving CUDA C++ builtin function declarations into compiler.
>   * Moved cuda_fp16.h and cuda_bf16.h into compiler bitcode.

### DIFF
--- a/.pfnci/windows/test.ps1
+++ b/.pfnci/windows/test.ps1
@@ -127,6 +127,10 @@ function Main {
     if (-Not $is_pull_request) {
         $Env:CUPY_TEST_FULL_COMBINATION = "1"
     }
+    # Skip full test for these CUDA versions as compilation seems so slow
+    if (($cuda -eq "12.0") -or ($cuda -eq "12.1") -or ($cuda -eq "12.2")) {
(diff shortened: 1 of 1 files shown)


asi1024 (approved):

kmaehashi:
Test failure is unrelated to this PR. #7240

PR #9007, opened 2025-03-03
[v13] API: Revert toDlpack() default to the old unversioned one

seberg:
This was an oversight in the original PR, even if this is basically deprecated, we should not have changed the default here.

We noticed this in downstream code, but the impact was not very big (use `__dlpack__()` and not common code) for now.

As this reverts behavior, I think it would probably only be useful if backported to a 13.4.1 in the not-so-far future.
(So if that is unlikely to happen, we can probably just close it. Those running into the issue should use `__dlpack__()`.)

---

CC @leofang.


kmaehashi:
Test failures are unrelated to this PR.

PR #9009, opened 2025-03-03
Decrease Windows build jobs from 16 to 8

jakirkham:
Reduce the number of parallel build jobs used in Windows builds. With this change, it now matches what is used on Linux.

https://github.com/cupy/cupy/blob/418a9169bc9994574a22e6bfb07edb39d91899c6/.pfnci/linux/tests/actions/build.sh#L5

We may also consider reducing the number of threads used by `nvcc`.

### DIFF
--- a/.pfnci/windows/test.ps1
+++ b/.pfnci/windows/test.ps1
@@ -74,7 +74,7 @@ function Main {
     ActivatePython $python
 
     # Setup build environment variables
-    $Env:CUPY_NUM_BUILD_JOBS = "16"
+    $Env:CUPY_NUM_BUILD_JOBS = "8"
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thanks John, but I think this isn't related to failures we saw in the CI, because this [env var](https://docs.cupy.dev/en/latest/reference/environment.html#for-installation) only affects the build-time parallelism, and we are seeing excessive memory usage at runtime.

jakirkham:
Thanks Kenichi! 🙏

Are we using parallelism at runtime? If so, maybe we could try decreasing parallelism at runtime instead

kmaehashi:
No, in runtime each kernel is built and launched sequentially.

leofang:
Shall we close this PR then?

Issue #8536, opened 2024-08-21
`cupy.fuse` unable to promote Python scalars in some cases (NEP 50?)


jakirkham:
### Description

A colleague identified an issue where Python scalars are sometimes not promoted to the appropriate device type ( https://github.com/rapidsai/cucim/issues/742#issuecomment-2299877295 ). Have included the code he used below

Though have also reproduced the same issue on Google Colab

### To Reproduce

Code:

```py
import cupy as cp

delta_phi = cp.full((10, 10), 9.999999e-7, dtype=cp.float32)
dt = cp.asarray(0.5, dtype=cp.float32)
mu = cp.asarray(0.1, dtype=cp.float32)

@cp.fuse()
def fused_kernel(dt, mu, delta_phi):
    return 1 + mu * dt * delta_phi

fused_kernel(dt, mu, delta_phi)
```

Error message:

```python
Traceback (most recent call last):
  File "cupy/_core/fusion.pyx", line 891, in cupy._core.fusion.Fusion.__call__
  File "cupy/_core/fusion.pyx", line 732, in cupy._core.fusion._FusionHistory.get_fusion
  File "<string>", line 10, in fused_kernel
  File "cupy/_core/fusion.pyx", line 198, in cupy._core.fusion._FusionVarScalar.__radd__
  File "cupy/_core/_kernel.pyx", line 1241, in cupy._core._kernel.ufunc.__call__
  File "cupy/_core/_fusion_thread_local.pyx", line 35, in cupy._core._fusion_thread_local.call_ufunc
  File "cupy/_core/fusion.pyx", line 949, in cupy._core.fusion._call_ufunc
  File "cupy/_core/fusion.pyx", line 605, in cupy._core.fusion._FusionHistory.call_ufunc
  File "cupy/_core/fusion.pyx", line 554, in cupy._core.fusion._FusionHistory.call_ufunc.can_cast1
TypeError: can_cast() does not support Python ints, floats, and complex because the result used to depend on the value.
This change was part of adopting NEP 50, we may explicitly allow them again in the future.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 12, in <module>
  File "cupy/_core/fusion.pyx", line 897, in cupy._core.fusion.Fusion.__call__
  File "cupy/_core/new_fusion.pyx", line 162, in cupy._core.new_fusion.Fusion.__call__
[...]

PR #9012, opened 2025-03-04
[v13] BUG: Hot fix for numpy 2 support in some fusion paths

seberg:
Scalars are a problem, and I am not convinced that this is a complete fix (for example w.r.t. integers).  OTOH v14 should already be working here, I believe.

But I think this is safe and could be back-portable.

---

Closes gh-8536

~Need to still check if v14 is indeed fine, we may want to forward port the test to make sure.~  v14 is fine as expected.


kmaehashi:
This is `to-be-backported` to v13.4.1 branch.

kmaehashi:
This is backported to v13.4.1 in #9016.
(The bot doesn't handle the irregular hotfix case 😅)

PR #9022, opened 2025-03-07
Fix rocThrust build for ROCm 6.3

Azusachan:
This commit fixes `/opt/rocm/include/thrust/version.h:34:10: fatal error: cuda/version: No such file or directory` error when try to build cupy for ROCm 6.3. See https://github.com/ROCm/rocThrust/issues/502. 

It also addresses the C++ version warning by changing compiler argument from c++11 to c++14


kmaehashi:
Thanks!

Issue #9019, opened 2025-03-06
sparse eigen value decomposition difference


khyeonm:
### Description

I am doing eigen value decomposition(EVD) with cupy. I encountered Out Of Memory issue and I have to solve this problem. So I converted data from dense to sparse and did EVD but dense and sparse have different results. 

### To Reproduce

```py
### numpy, sparse difference
import numpy as np
from scipy.sparse import csr_matrix
from scipy.sparse.linalg import eigsh
from scipy.linalg import eigh

# random data
np.random.seed(42)
data = np.random.rand(100, 100)

# change to sparse
data_sparse = csr_matrix(data)

# dense
L_dense, V_dense = eigh(data)

# sparse
L_sparse, V_sparse = eigsh(data_sparse, k=min(data_sparse.shape) - 1) 

print("Numpy EVD (first 5 eigen values):", L_dense[-5:])  
print("Scipy sparse EVD (first 5 eigen values):", L_sparse[-5:])


### dask + cupy sparse, dask + cupy difference
import cupy as cp
import cupyx.scipy.sparse as sparse
import dask.array as da
from cupyx.scipy.sparse.linalg import eigsh
from cupy.linalg import eigh

# random data
cp.random.seed(42)
data = cp.random.rand(100, 100)

# change to sparse
data_sparse = sparse.csr_matrix(data)

# change to dask array (sparse matrix)
data_dask_sparse = da.from_array(data_sparse.toarray(), chunks=(50, 50))
data_dask_sparse = data_dask_sparse.map_blocks(sparse.csr_matrix)

# change to dask array (dense matrix)
data_dask_dense = da.from_array(data, chunks=(50, 50))

# EVD (Dask + CuPy Sparse)
k = min(data_sparse.shape) // 2 
L_sparse, V_sparse = eigsh(data_dask_sparse.compute(), k=k)

# EVD (Dask + CuPy Dense)
L_dense, V_dense = eigh(data_dask_dense.compute())

# compare results
print("sparse EVD :", L_sparse) 
print("dense EVD :", L_dense)

```

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
# Paste the output here
OS                           : Linux-6.1.0-28-amd64-x86_64-with-glibc2.36
Python Version               : 3.12.8
CuPy Version                 : 13.4.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.0.2
SciPy Version                : 1.15.1
Cython Build Version         : 3.0.12
[...]

PR #9027, opened 2025-03-11
sparse eigen value decomposition difference

khyeonm:
Part of #9019


kmaehashi:
Let me close as there's no activity for a while.

PR #9045, opened 2025-03-19
Update cutensornet accelerator based on cuquantum-python 25.03 deprecation

yangcal:
`cutensornet` from `cuquantum-python` has been an optional accelerator for `cupy.einsum` via `cupy/linalg/_einsum_cutn.py`. 

Starting `cuquantum-python==25.03`, the `cuquantum.cutensornet` module is deprecated with bindings migrated to `cuquantum.bindings.cutensornet` and pythonic APIs, e.g, `cuquantum.cutensornet.contract` migrated to `cuquantum.tensornet.contract`. The deprecated module will remain functional for a few releases before being removed completely. Release notes can be found [here](https://docs.nvidia.com/cuda/cuquantum/latest/python/release-notes.html#cuquantum-python-v25-03-0).

This PR update `_einsum_cutn.py` to use the new APIs. 

Note that `DeprecationWarning` is still expected after this PR is merged due to `cutensornet` module being imported at the top level of `cuquantum`. The warnings will be fully removed when the deprecated modules are removed.


yangcal:
@leofang for vis

kmaehashi (approved):

PR #9053, opened 2025-03-21
Support FIPS enabled machines with MD5 hashing

MattTheCuber:
FIPS enabled machines prohibit MD5 hashing for security reasons. Using hashlib.md5 in Python will throw the error ValueError: [digital envelope routines] unsupported. There are typically 2 resolutions to this problem: use the usedforsecrutiy=False flag or use SHA1 hashing. The usedforsecurity flag is easier to implement and was introduced in Python 3.9, which is the minimum Python version for this library, so that is the method I chose.

### DIFF
--- a/cupy/fft/_callback.pyx
+++ b/cupy/fft/_callback.pyx
@@ -364,7 +364,8 @@ cdef class _CallbackManager:
         # are considered identical regardless of which plan is actually
         # executed at the time of generation
         mod_name = 'cupy_callback_'
-        mod_name += hashlib.md5(keys.encode(), usedforsecurity=False).hexdigest()
+        mod_name += hashlib.md5(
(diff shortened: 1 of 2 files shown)


leofang:
Thanks for good catch! Could you apply the same treatment as done in https://github.com/cupy/cupy/pull/5988? In particular, the same helper function defined in `cupy/cuda/compiler.py` can be reused in both places.

kmaehashi (approved):

MattTheCuber:
Want me to refactor away the `_hash_digest` as well?

kmaehashi:
Thanks @MattTheCuber, merging.

> Want me to refactor away the `_hash_digest` as well?

Yeah, would be great if you could!

PR #9057, opened 2025-03-24
Refactor hashing

MattTheCuber:
Follow up of #9053

Removed `_hash_hexdigest` and switched MD5 hashing to SH1 for speed and  consistency.

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -318,8 +318,11 @@ def _jitify_prep(source, options, cu_path):
     return options, headers, include_names
 
 
-_hash_length = len(hashlib.sha1(
-    b'', usedforsecurity=False).hexdigest())  # 40 for SHA1
+def _hash_hexdigest(value):
+    return hashlib.sha1(value, usedforsecurity=False).hexdigest()
+
+
+_hash_length = len(_hash_hexdigest(b''))  # 40 for SHA1
 
 
 def compile_using_nvrtc(source, options=(), arch=None, filename='kern.cu',
@@ -579,7 +582,7 @@ def _compile_with_cache_cuda(
     key_src = '%s %s %s %s %s' % (
         env, base, source, extra_source, _get_cupy_cache_key())
     key_src = key_src.encode('utf-8')
-    name = hashlib.sha1(key_src, usedforsecurity=False).hexdigest() + '.cubin'
+    name = _hash_hexdigest(key_src) + '.cubin'
 
     mod = function.Module()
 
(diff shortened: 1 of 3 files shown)


cupy/cuda/compiler.py
```diff
@@ -318,18 +318,8 @@ def _jitify_prep(source, options, cu_path):
...
-    if _has_usedforsecurity:
-        hashobj = hashlib.sha1(value, usedforsecurity=False)
-    else:
-        hashobj = hashlib.sha1(value)
-    return hashobj.hexdigest()
-
-
-_hash_length = len(_hash_hexdigest(b''))  # 40 for SHA1
+_hash_length = len(hashlib.sha1(
+    b'', usedforsecurity=False).hexdigest())  # 40 for SHA1
```
MattTheCuber, line 322: Should this be removed and just use `40` when its used instead?

cupy/cuda/compiler.py
```diff
@@ -318,18 +318,8 @@ def _jitify_prep(source, options, cu_path):
     return options, headers, include_names
 
 
-_has_usedforsecurity = (sys.version_info >= (3, 9))
-
-
-def _hash_hexdigest(value):
```
kmaehashi, line 324: Could you retain `_hash_hexdigest` and `_hash_length`? These functions exist so that we can easily change the algorithm in future when needed.

cupy/random/_generator.py
```diff
@@ -801,7 +801,7 @@ def seed(self, seed=None):
                 seed = (time.time() * 1000000) % _UINT64_MAX
         else:
             if isinstance(seed, numpy.ndarray):
-                seed = int(hashlib.md5(
+                seed = int(hashlib.sha1(
```
kmaehashi, line 804: Please keep using `md5` here, as this change will (unnecesarily) break reproducibility of existing users with fixed seed.

kmaehashi (changes requested):
Thanks @MattTheCuber!

cupy/cuda/compiler.py
MattTheCuber: I thought that was what you were saying here: https://github.com/cupy/cupy/pull/9053#issuecomment-2747426375?

kmaehashi: Sorry, my bad. I mistakenly thought you were referring to remove away `_has_usedforsecurity` flag.

To share the context, we once changed the hash algorithm from MD5 to SHA1 in the past (#5988), and later noticed that the hash length was not updated (which was fixed in #6023). To make the code robust to future changes, I'd prefer to keep using `_hash_hexdigest` and `_hash_length` here.

MattTheCuber: Sure, I was just thinking you probably wouldn't change back since MD5 breaks FIPS :shrug:. I'll add it back.

kmaehashi (approved):

PR #9070, opened 2025-03-30
add ndenumerate function for cupy.

pererakavindu:
about issue 6078
add new functionality
tested locally.

### DIFF
--- a/cupy/_indexing/iterate.py
+++ b/cupy/_indexing/iterate.py
@@ -4,6 +4,14 @@ import cupy
 from cupy import _core
 from cupy._core import internal
 
+def ndenumerate(a:cupy.ndarray):
+    """ enumerate a given array in cupy """
+    try:
+        shape=a.shape
+    except:
+        raise ValueError("array should be a cupy array")
+    for i in cupy.ndindex(shape):
+        yield i,a[i]
 
 class flatiter:
     """Flat iterator object to iterate over arrays.


leofang:
Hi, could you please push code to the same branch so that the changes are contained in the same PR, instead of creating multiple PRs?

pererakavindu:
I couldn't find the most suitable place to add this function, but it should accomplish the task as intended.

kmaehashi:
Could you port the implementation from NumPy to keep the behavior consistent with NumPy?

https://github.com/numpy/numpy/blob/a0462960f7bcdcae3f14d0d27df77774712db5d7/numpy/lib/_index_tricks_impl.py#L594-L640

Also please add unit tests.

cupy/_indexing/iterate.py
```diff
@@ -4,6 +4,14 @@
 from cupy import _core
 from cupy._core import internal
 
+def ndenumerate(a:cupy.ndarray):
+    """ enumerate a given array in cupy """
```
a-holm, line 8: he current docstring is minimal. Could you please expand it to follow the style of other functions/classes in CuPy (like `flatiter` in the same file)? It should ideally include parameter descriptions, return value explanation, and potentially a simple example or a `.. seealso::` directive pointing to `numpy.ndenumerate` or `cupy.ndindex`.

PR #9075, opened 2025-04-01
[v13] Skip some tests for numerical error from NumPy 2

asi1024:
CuPy v13 does not guarantee the compatibility with NumPy 2.x.


asi1024:
CI failure seems unrelated
https://ci.preferred.jp/cupy.win.cuda124/

kmaehashi (approved):

PR #9030, opened 2025-03-13
Allow discovering cuTENSOR using major version

kmaehashi:
Allow loading cuTENSOR 2.x wheels regardless of the version used at build time (currently 2.1). Fixes #9017.

This PR also moves the generation logic of wheel metadata file (`cupy/.data/_wheel.json`) from [cupy-release-tools](https://github.com/cupy/cupy-release-tools/blob/3e75f4c5b20f62f6898770ed5f190ba1e2fb221e/dist.py#L75-L96) to cupy, so that CuPy developers can generate the wheel metadata by

```
$ python cupyx/tools/_generate_wheel_metadata.py --cuda 12.x --library cutensor --library cudnn --library nccl > cupy/.data/_wheel.json
```

and easily test the preload mechanism like:

```
$ export CUPY_DEBUG_LIBRARY_LOAD=1
$ python -c 'import cupy'
```

PR #9078, opened 2025-04-03
Allow specifing no libraries when generating wheel metadata

kmaehashi:
Follows-up #9030.

AArch64 wheels are currently not built with support for optional libraries such as cuDNN/cuTENSOR. `_generate_wheel_metadata.py` must be able to generate a wheel metadata containing no preload libraries for that case.


kmaehashi:
Locally confirimed:

```
% python cupyx/tools/_generate_wheel_metadata.py --cuda 12.x
{
    "cuda": "12.x",
    "packaging": "pip"
}
```

kmaehashi:
@asi1024 Could you take a look? This one is mandatory for release.

PR #8916, opened 2025-01-31
Add cupyx.scipy.ndimage.find_objects

grlee77:
This MR implements the `find_objects` functions from SciPy. This function finds the Cartesian bounding boxes for all integer-labeled objects in an image as in the example image shown below:

find_objects_bounding_boxes

So, conceptually the kernel is very simple and just needs to find the minimum and maximum coordinate index associated with each label value. In the first commit, a simple elementwise kernel was created for this. 

This was refined in a subsequent commit which updated the kernel to process multiple sequential pixels per GPU thread to reduce the overhead of atomic operations. This makes the kernel harder to read, but has substantial performance benefit (see benchmark results below)

A simple Cython function was also implemented to perform the final loop over the coordinates that creates the Python slice objects that are returned.

## Benchmarks

Benchmarks were run at a few different 2D and 3D image sizes, where for each size three different object densities were investigated. Examples of what these density images look like for the shape (512, 512) shape is shown here (they are created with a third-party `cucim.skimage.data.binary_blobs` function for generating blobs of various sizes/density):

find_objects_volume_fractions

#### For initial commit (1 pixel per GPU thread) and no Cython slice object creation

|   shape   |  num_labels   |   Duration (skimage)  |  Duration (cuCIM)  |  Acceleration  |
|-----------|---------------|-----------------------|--------------------|----------------|
| (512, 512) | 2747 | 0.00307 | 0.00555 | 0.55 |
| (512, 512) | 309 | 0.00103 | 0.00060 | 1.71 |
| (512, 512) | 41 | 0.00092 | 0.00029 | 3.19 |
| (3840, 2160) | 1558 | 0.02739 | 0.00573 | 4.78 |
| (3840, 2160) | 179 | 0.02504 | 0.00515 | 4.86 |
| (3840, 2160) | 26 | 0.02546 | 0.00629 | 4.05 |
| (192, 192, 192) | 7710 | 0.05011 | 0.02828 | 1.77 |
| (192, 192, 192) | 626 | 0.03022 | 0.00819 | 3.69 |
| (192, 192, 192) | 60 | 0.02493 | 0.00708 | 3.52 |

#### For intermediate version with multiple pixels per thread but no Cython code
|   shape   |  num_labels   |   Duration (skimage)  |  Duration (cuCIM)  |  Acceleration  |
|-----------|---------------|-----------------------|--------------------|----------------|
| (512, 512) | 2747 | 0.00419 | 0.00682 | 0.61 |
[...]

PR #9081, opened 2025-04-03
Include `pxd` files in `cupyx`

kmaehashi:
Follows-up #8916. `pxd` files must be explicitly added via `MANIFEST.in`, otherwise source build from sdist does not work.


kmaehashi:
@asi1024 This one is also blocking the release build :bow:

asi1024 (approved):

PR #9082, opened 2025-04-03
Implement  numpy.geomspace

BHenkemans:
Tracking issue #6078

### DIFF
--- a/cupy/_creation/ranges.py
+++ b/cupy/_creation/ranges.py
@@ -3,7 +3,7 @@ import math
 import numpy
 
 import cupy
-from cupy import _core
+from cupy import _core, result_type
 
 
 def arange(start, stop=None, step=1, dtype=None):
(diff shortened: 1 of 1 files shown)


cupy/_creation/ranges.py
```diff
@@ -254,6 +254,67 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None,
...
+            Use ``-1`` to get an axis at the end.
+    Returns:
+        cupy.ndarray: The 1-D array of ranged values.
+
+    .. seealso:: :func:`numpy.geomspace`
+    """
+    start = cupy.asanyarray(start)
+    stop = cupy.asanyarray(stop)
+
+    if cupy.any(start == 0) or cupy.any(stop == 0):
```
kmaehashi, line 284: This incurs device synchronization which significantly impacts the performance. Considering `geomspace` is often used with scalar start/end, maybe need a fast path for such case.

BHenkemans: Added!

kmaehashi: Thanks! I think it's better to make the scalar case as completely different code path, because doing scalar sign/div/log10 on GPU is highly likely slower than doing it on CPU due to overhead, and it is a hot path (i.e. frequently used case).

Could you implement the CPU path using NumPy? This will allow us to change the condition to use the path. For example we might want to use the fast path for smaller inputs like `start=[1,2]` as well in future.

c.f. `linspace` has a specialized case for scalar inputs: https://github.com/cupy/cupy/blob/37d412f2df6b67f1bbe9923c497bcf58b812bf45/cupy/_creation/ranges.py#L160-L161

BHenkemans: Let me know if you want it implemented in another way.

BHenkemans:
@kmaehashi do you need any additional things from me to complete this issue?

cupy/_creation/ranges.py
```diff
@@ -254,6 +254,104 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None,
...
+        num: Number of elements.
+        endpoint (bool): If ``True``, the stop value is included as the last
+            element. Otherwise, the stop value is omitted.
+        dtype: Data type specifier. It is inferred from the start and stop
+            arguments by default.
+    Returns:
+        cupy.ndarray: The 1-D array of ranged values.
+
+    .. seealso:: :func:`numpy.geomspace`
+    """
```
kmaehashi, line 274: Removing docs for this private functions as redundant.

```suggestion
```

cupy/_creation/ranges.py
```diff
@@ -254,6 +254,104 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None,
...
+    # Fast track if parameters are scalars
+    if cupy.isscalar(start) and cupy.isscalar(stop):
+        return _geomspace_scalar(start, stop, num, endpoint, dtype)
+
+    start = cupy.asanyarray(start)
+    stop = cupy.asanyarray(stop)
+    if cupy.any(start == 0) or cupy.any(stop == 0):
+        raise ValueError('Geometric sequence cannot include zero')
+
+    dt = cupy.result_type(start, stop, float(num), cupy.zeros((), dtype=dtype))
```
kmaehashi, line 324: `cupy.zeros((), dtype=dtype)` unnecesarilly allocates the memory. Could you avoid this?

cupy/_creation/ranges.py
```diff
@@ -254,6 +254,104 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None,
...
+            element. Otherwise, the stop value is omitted.
+        dtype: Data type specifier. It is inferred from the start and stop
+            arguments by default.
+        axis (int):  The axis in the result to store the samples.  Relevant
+            only if start or stop are array-like.  By default ``0``, the
+            samples will be along a new axis inserted at the beginning.
+            Use ``-1`` to get an axis at the end.
+    Returns:
+        cupy.ndarray: The 1-D array of ranged values.
+
```
kmaehashi, line 308: Please include this warning here.
https://github.com/cupy/cupy/blob/v13.3.0/cupy/_indexing/indexing.py#L144-L146

cupy/_creation/ranges.py
```diff
@@ -254,6 +254,104 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None,
...
+    .. seealso:: :func:`numpy.geomspace`
+    """
+    if start == 0 or stop == 0:
+        raise ValueError('Geometric sequence cannot include zero')
+
+    np_result = numpy.geomspace(start, stop, int(
+        num), endpoint=endpoint, dtype=dtype)
```
kmaehashi, line 279: Could you implement the `numpy.geomspace` equivalent for scalar, like done in https://github.com/cupy/cupy/blob/37d412f2df6b67f1bbe9923c497bcf58b812bf45/cupy/_creation/ranges.py#L64-L119? I think most of the code can be copied from [NumPy](https://github.com/numpy/numpy/blob/v2.1.0/numpy/_core/function_base.py#L323-L470)'s implementation but calling `cupy.logspace`.

cupy/_creation/ranges.py
kmaehashi: Sorry for the delay in reviewing. Please check my follow-up review comments.

kmaehashi (changes requested):

PR #9088, opened 2025-04-04
Implement the scipy.stats.gmean function for cupy using cupy routines.

Henkdebutler:
In the comparison table (https://docs.cupy.dev/en/latest/reference/comparison.html#statistical-functions), the function scipy.stats.gmean did not have a cupy counterpart, so I adapted the scipy implementation to work with cupy functions. Includes some testcases.


kmaehashi:
Thanks. `nan_policy` and `keepdims` keyword arguments seems missing?

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gmean.html

Henkdebutler:
I implemented these keyword arguments for the gmean function, but scipy has some generalized way of dealing with these two keyword arguments, and it has a generalized way of dealing with edge case behaviour for some nan policies of its stats functions. If more stats functions are implemented for cupy, it is probably a good idea to do the same (but this is very much beyond the scope of this PR).

Henkdebutler:
@kmaehashi I made the changes you requested a while ago. Could you please take a look?

PR #9102, opened 2025-04-18
Implement lazy load for cuquantum

yangcal:
This PR implements lazy load for `cuquantum` in `cupy.einsum`.

Currently `cuquantum/cutensornet` is an optional accelerator of `cupy.einsum`. However the optional module will be loaded regardless of whether it has been set as `CUPY_ACCELERATOR` if it's installed. Since `cuquantum` has its own dependencies with some overhead during import,  we here switch to lazy import for `cuquantum` such that non-einsum users will not need to load it into their environment.

### DIFF
--- a/cupy/linalg/_einsum_cutn.py
+++ b/cupy/linalg/_einsum_cutn.py
@@ -1,20 +1,6 @@
 import threading
 import warnings
 
-try:
-    import cuquantum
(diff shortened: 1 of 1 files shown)


yangcal:
@leofang for vis

kmaehashi (approved):

Issue #9087, opened 2025-04-04
Cusolver potrs/potrs_batched


stavoltafunzia:
### Description

I see that the (batched) `potrf` routine is implemented in `cupy.linalg._decomposition`, however the corresponding solver routine (batched) `potrs` is currently not implemented anywere. There is `posv` in `cupyx.lapack` but it does not allow to call `potrs` only.

Am I wrong or the (batched) `potrs` routine is still not exposed? I'm open to try sending a pull request if this is something that make sense to be implemented.

leofang:
Hi @stavoltafunzia currently the non-batched POTRS is also used in `cupyx.lapack.posv()` routine when the matrix A is only 2D:
https://github.com/cupy/cupy/blob/1e8ade16f7f0af01e748e706e8c8302ffd2fc873/cupyx/lapack.py#L301-L320
Does it not work?

stavoltafunzia:
Yes I did notice that, but my point was that if you already have the cholesky decomposition and you just need to find the roots for multiple `rhs`, then it's more efficient to call only `potrs` many times rather than calling both `potrf` and `potrs` (as `posv` does) multiple times. Am I missing anything?

stavoltafunzia:
That could be a solution indeed. Though, in my opinion, it would be cleaner to have some kind of standalone `potrs` function exposed, then `posv` can be a simple call to `potrf` and `potrs`. People coming from the Lapack/Cusolver world would look into Cupy for `potrf` and `potrs`, and it may not be obvious to search into `posv`. I personally find this approach more structural, the former something like a quick patch. As I said, I could try making a PR, if what I'm saying makes sense.

leofang:
> Though, in my opinion, it would be cleaner to have some kind of standalone `potrs` function exposed, then `posv` can be a simple call to `potrf` and `potrs`.

It would likely take the same amount of refactoring work as what I suggested, while still keeping the codebase maintainable, so it is not a bad idea. 

However, a common challenge for all CuPy routines is that there are lots of boilerplate code for e.g. input checking and intermediate call preparation. If we break `posv` into two calls, how do we keep the run-time overhead the same as it is today? We certainly don't want to introduce regression for a simple refactoring.

> People coming from the Lapack/Cusolver world would look into Cupy for `potrf` and `potrs`, and it may not be obvious to search into `posv`. 

[...]

PR #9116, opened 2025-04-28
[WIP] Implements Lapack potrs

stavoltafunzia:
Addresses #9087 and, partially, #6324. Implements Lapack `potrs` and also exposes the `scipy.linalg.cho_solve` interface.

As explained in #9087, `potrs` is useful when multiple linear systems have to be solved for the same symmetric matrix A. Multiple calls to `posv` are suboptimal because the same Cholesky decomposition is recalculated multiple times. This happens, for example, in Gaussian Processes.

`potrs` is here implemented inside `cupyx.lapack`, offering a Lapack-like interface. It supports batched systems and also more than one RHS (cusolver does not support NRHS > 1 for the batched version, here a workaround is implemented). Next to that, a scipy-like function in added in `cupy.linalg.cho_solve` which keeps the same interface as the corresponding `scipy` equivalent.

I intentionally did not modify the existing `cupyx.lapack.posv` function to make use of the new `potrs` implementation so as not to add too much boilerplate code to it, as requested by @leofang [here](https://github.com/cupy/cupy/issues/9087#issuecomment-2814600368).

This PR is marked as WIP for this reason: I noted that if non C-contiguous arrays are returned, unittests failed. Thus I modified the return statements to ensure C-contiguity. I would like to avoid this because it may introduce unnecessary memory copies (the end-users may simply use an F-contiguous array), but I don't know if it's allowed by cupy common practices as unittests were failing.

### DIFF
--- a/cupy/linalg/__init__.py
+++ b/cupy/linalg/__init__.py
@@ -37,6 +37,7 @@ from cupy.linalg._solve import lstsq  # NOQA
 from cupy.linalg._solve import inv  # NOQA
 from cupy.linalg._solve import pinv  # NOQA
 from cupy.linalg._solve import tensorinv  # NOQA
+from cupy.linalg._solve import cho_solve
 
 # -----------------------------------------------------------------------------
 # Exceptions
@@ -46,4 +47,4 @@ from numpy.linalg import LinAlgError  # NOQA
 
 __all__ = ["matrix_power", "cholesky", "qr", "svd", "eigh", "eigvalsh", "norm",
            "det", "matrix_rank", "slogdet", "solve", "tensorsolve", "inv",
-           "pinv", "tensorinv", "LinAlgError"]
+           "pinv", "tensorinv", "LinAlgError", "cho_solve"]
--- a/cupyx/lapack.py
+++ b/cupyx/lapack.py
@@ -346,3 +346,189 @@ def posv(a, b):
         potrs, dev_info)
 
     return _cupy.ascontiguousarray(b.reshape(b_shape))
+
+
+def potrs(L, b, lower):
+    """ Implements lapack XPOTRS through cusolver.potrs. Solves linear system
+    A x = b given the cholesky decomposition of A, namely L. Supports also
+    batches of linear systems and more than one right-hand side (NRHS > 1).
+
+    Args
+    ----------
+    L (cupy.ndarray): Array of Cholesky decomposition of real symmetric or
+        complex hermitian matrices with dimension (..., N, N).
+
+    b (cupy.ndarray): right-hand side (..., N) or (..., N, NRHS). Note that
+        this array may be modified in place, as usually done in LAPACK
+
+    lower (bool): If True, L is lower triangular. If False, L is upper
+        triangular.
+
+    Returns
+    -------
+        cupy.ndarray: The solution to the linear system. Note this may point to
+            the same memory as b, since b may be modified in place.
+
+    .. warning::
+        This function calls one or more cuSOLVER routine(s) which may yield
+        invalid results if input conditions are not met.
+        To detect these invalid results, you can set the `linalg`
+        configuration to a value that is not `ignore` in
+        :func:`cupyx.errstate` or :func:`cupyx.seterr`.
+
+    """
+
+    from cupy_backends.cuda.libs import cusolver as _cusolver
+
+    # Check if batched should be used
+    if len(L.shape) > 2:
+        return _batched_potrs(L, b, lower)
+
+    # Check input arguments
+    assert L.ndim == 2, "L is not a matrix"
+    assert L.shape[0] == L.shape[1], "L is not a square matrix"
+    n = L.shape[-1]
+    b_shape = b.shape
+    if b.ndim == 1:
+        b = b[:, None]
+    assert b.ndim == 2, "b is not a vector or a matrix"
+    assert b.shape[0] == n, "length of arrays in b does not match size of L"
+
+    # Check memory order and type
+    dtype = _numpy.promote_types(L.dtype, b.dtype)
+    dtype = _numpy.promote_types(dtype, 'f')
+    L, b = L.astype(dtype, copy=False), b.astype(dtype, copy=False)
+    if (not L.flags.f_contiguous) and (not L.flags.c_contiguous):
+        L = _cupy.asfortranarray(L)
+    if L.flags.c_contiguous:
+        lower = not lower  # Cusolver assumes F-order
+        # For complex types, we need to conjugate the matrix
+        if b.size < L.size:  # Conjugate the one with lower memory footprint
+            b = b.conj()
+        else:
+            L = L.conj()
+    if (b.dtype != dtype) or (not b.flags.f_contiguous):
+        b = _cupy.asfortranarray(b)
+
+    # Take correct dtype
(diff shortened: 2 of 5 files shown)


stavoltafunzia:
Just for clarification: if someone confirms that cupy functions (specifically, the ones added in this PR) should always return C-contiguous arrays, then this PR is no more WIP and I don't expect further changes other than what requested by reviewers. If returning F-contiguous arrays is allowed, then I'd like to know how to configure unittests so as not to fail them. I was not able to find something about C or F-contiguity in the documentation.

cupy/linalg/__init__.py
```diff
@@ -37,6 +37,7 @@
 from cupy.linalg._solve import inv  # NOQA
 from cupy.linalg._solve import pinv  # NOQA
 from cupy.linalg._solve import tensorinv  # NOQA
+from cupy.linalg._solve import cho_solve
```
leofang: `cho_solve` should live in `cupyx.scipy.linalg` since there is no NumPy counterpart

cupyx/lapack.py
```diff
@@ -346,3 +346,189 @@ def posv(a, b):
...
+        b.data.ptr,
+        b.shape[0],  # ldB
+        dev_info.data.ptr)
+    _cupy.linalg._util._check_cusolver_dev_info_if_synchronization_allowed(
+        potrs, dev_info)
+
+    # Conjugate back if necessary
+    if L.flags.c_contiguous and b.size < L.size:
+        b = b.conj()
+    return _cupy.asarray(b, order='C').reshape(b_shape)
```
leofang: Just reshape, no need to run `asarray`

tests/cupy_tests/linalg_tests/test_solve.py
```diff
@@ -339,3 +340,83 @@ def test_invalid_shape(self):
...
+        if (
+            numpy.lib.NumpyVersion(numpy.__version__) < "2.0.0"
+            or a.shape[:-1] != b.shape
+        ):
+            return numpy.linalg.solve(a, b)
+        b = b[..., numpy.newaxis]
+        return numpy.linalg.solve(a, b)[..., 0]
+
+    @testing.for_dtypes('fdFD')
+    @testing.numpy_cupy_allclose(atol=1e-5)
```
leofang: set `contiguous_check` to `False` to avoid comparing strides/contiguity. In general, CuPy does not guarantee following NumPy's contiguity requirement.

leofang (changes requested):
Sorry for delay. Leaving some quick comments.
> I noted that if non C-contiguous arrays are returned, unittests failed. 

see below, it should be relaxed. No need to compare the strides.

stavoltafunzia:
@leofang thanks for the comments, they should now be addressed. No worries for delay, in the end this is an open source project and we all have other activities/jobs.

stavoltafunzia:
I see that cuda docs build is currently failing. @leofang do you have any insights why? Seems to complain about this `/home/runner/work/cupy/cupy/cupy/__init__.py:docstring of cupy.RawKernel.__call__:1: WARNING: Inline emphasis start-string without end-string.` but I don't get how it's related to this PR. I've seen another PR with the same issue, probably something not in scope here.

cupyx/lapack.py
```diff
@@ -348,3 +348,189 @@ def posv(a, b):
...
+        To detect these invalid results, you can set the `linalg`
+        configuration to a value that is not `ignore` in
+        :func:`cupyx.errstate` or :func:`cupyx.seterr`.
+
+    """
+
+    from cupy_backends.cuda.libs import cusolver as _cusolver
+
+    # Check if batched should be used
+    if len(L.shape) > 2:
```
leofang:
```suggestion
    if L.ndim > 2:
```

cupyx/lapack.py
```diff
@@ -348,3 +348,189 @@ def posv(a, b):
...
+    Args
+    ----------
+    L (cupy.ndarray): Array of Cholesky decomposition of real symmetric or
+        complex hermitian matrices with dimension (..., N, N).
+
+    b (cupy.ndarray): right-hand side (..., N) or (..., N, NRHS). Note that
+        this array may be modified in place, as usually done in LAPACK
+
+    lower (bool): If True, L is lower triangular. If False, L is upper
+        triangular.
```
leofang: I think we use Google-style docstrings:
```suggestion
    Args:
        L (cupy.ndarray): Array of Cholesky decomposition of real symmetric or
            complex hermitian matrices with dimension (..., N, N).
        b (cupy.ndarray): right-hand side (..., N) or (..., N, NRHS). Note that
            this array may be modified in place, as usually done in LAPACK.
        lower (bool): If True, L is lower triangular. If False, L is upper
            triangular.
```

cupyx/lapack.py
```diff
@@ -348,3 +348,189 @@ def posv(a, b):
...
+
+    Returns
+    -------
+        cupy.ndarray: The solution to the linear system. Note this may point to
+            the same memory as b, since b may be modified in place.
```
leofang:
```suggestion
    Returns:
        cupy.ndarray: The solution to the linear system. Note this may point to
            the same memory as b, since b may be modified in place.
```

cupyx/lapack.py
```diff
@@ -348,3 +348,189 @@ def posv(a, b):
...
+        This function calls one or more cuSOLVER routine(s) which may yield
+        invalid results if input conditions are not met.
+        To detect these invalid results, you can set the `linalg`
+        configuration to a value that is not `ignore` in
+        :func:`cupyx.errstate` or :func:`cupyx.seterr`.
+
+    """
+
+    from cupy_backends.cuda.libs import cusolver as _cusolver
+
```
leofang, line 380:
```suggestion
    from cupy_backends.cuda.libs import cusolver as _cusolver

    _util = _cupy.linalg._util
    _util._assert_cupy_array(L, b)
    _util._assert_stacked_2d(L)
    _util._assert_stacked_square(L)

```

cupyx/lapack.py
```diff
@@ -348,3 +348,189 @@ def posv(a, b):
...
+    # Check if batched should be used
+    if len(L.shape) > 2:
+        return _batched_potrs(L, b, lower)
+
+    # Check input arguments
+    assert L.ndim == 2, "L is not a matrix"
+    assert L.shape[0] == L.shape[1], "L is not a square matrix"
```
leofang: move the checks to earlier
```suggestion
```

cupyx/lapack.py
```diff
@@ -348,3 +348,189 @@ def posv(a, b):
...
+
+def _batched_potrs(L, b, lower: bool):
+    from cupy_backends.cuda.libs import cusolver as _cusolver
+    import cupyx.cusolver
+
+    if not cupyx.cusolver.check_availability('potrsBatched'):
+        raise RuntimeError('potrsBatched is not available')
+
+    # CHeck input arrays
+    assert L.shape[-1] == L.shape[-2], "L is not a batch of square matrices"
```
leofang: ditto
```suggestion
```

leofang (approved):
LGTM! Sorry for delay and thanks to your contribution @stavoltafunzia! There are some Python overheads that can be optimized out in a separate PR. I don't want to delay this further. Let's get it in.

PR #9117, opened 2025-04-30
BUG: cupyx.scipy.signal: make gammatone return arrays

ev-br:
Currently, one of `gammatone` return values is an array or a list, depending on other inputs. Fix it to always return arrays.

Observed in https://github.com/scipy/scipy/pull/22914, which delegates to `cupyx.scipy.signal.gammatone` from `scipy.signal.gammatone`.


kmaehashi (approved):

kmaehashi:
Thanks, @ev-br!

PR #9119, opened 2025-05-02
Enhance query_pairs to support output_type='set'

Charis-E-Shiny:
Added a new test case test_query_pairs_set_output_type in the TestPairs class to verify the correctness of the query_pairs method when output_type='set' is specified. This ensures that the function returns results as a set of unique index pairs (i, j) where i < j, aligning with expected behavior. This test improves coverage for different output types supported by query_pairs.

### DIFF
--- a/cupyx/scipy/spatial/_kdtree.py
+++ b/cupyx/scipy/spatial/_kdtree.py
@@ -425,19 +425,22 @@ class KDTree:
             branches are added in bulk if their furthest points are nearer
             than ``r * (1+eps)``.  `eps` has to be non-negative.
         output_type : string, optional
-            Choose the output container, 'set' or 'ndarray'. Default: 'ndarray'
-            Note: 'set' output is not supported.
+            Choose the output container, options are:
+            - 'ndarray': Returns the pairs as an ndarray of size (total_pairs, 2)
+            - 'set': Returns the pairs as a set of unique pairs (i, j)
+            Default: 'ndarray'
 
         Returns
         -------
-        results : ndarray
-            An ndarray of size ``(total_pairs, 2)``, containing each pair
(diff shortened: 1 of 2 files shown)


Charis-E-Shiny:
Hi @cupy/maintainers 👋, 
The CI checks are running, and initial tests have passed ✅.
Could you please take a look and review this PR when convenient? 🙏 
Thank you!

Charis-E-Shiny:
> Hi @cupy/maintainers 👋, The CI checks are running, and initial tests have passed ✅. Could you please take a look and review this PR when convenient? 🙏 Thank you!


Hi @leofang @maehashi,

This PR has passed the pre-review checks but the pfn-public-ci jobs are still showing as "Waiting for status to be reported." Could you please help trigger the full CI?
Please let me know if there’s anything else needed from my side. 🙂

Thank you! 🙏

kmaehashi:
Hi @Charis-E-Shiny, I think decomposing `ndarray` to `set` doesn't much sense in GPU implementation. In small data scenarios people can just use SciPy's KDTree.

Charis-E-Shiny:
> Hi @Charis-E-Shiny, I think decomposing `ndarray` to `set` doesn't much sense in GPU implementation. In small data scenarios people can just use SciPy's KDTree.

Hi @kmaehashi, thank you for the feedback! 
You're absolutely right — for small datasets, SciPy's KDTree is more efficient.
However, this change aims to support more general workflows, including GPU-parallelized cases or scenarios where ndarray-to-set conversion helps with deduplication or preprocessing.

Would you consider merging this as an optional enhancement? I’d be happy to guard it behind a flag or add a clear note in the docs to clarify its intended use case.

Let me know what you think!

Issue #9138, opened 2025-05-22
`cupyx.linalg.expm` does


ncrubin:
### Description

The expm code that implements the [13/13] Pade approximation to the matrix exponential has a typo.  https://github.com/cupy/cupy/blob/118ade4a146d1cc68519f7f661f2c145f0b942c9/cupyx/scipy/linalg/_matfuncs.py#L145 uses `A` instead of `A2` as is described by Algorithm 2.3 line 19 in ( N. Higham, SIAM J. MATRIX ANAL. APPL. Vol. 26(4), p. 1179 (2005) https://doi.org/10.1137/04061101X)

### To Reproduce

I use the fact that exponentiation of antisymmetric matrices returns a unitary.  I compute equivalence (up to phases) by the trace overlap of the unitaries and compare to the base scipy implementation. I mock out the `_expm_inner` problematic function with `_fixed_expm_inner`.  

```py
import numpy as np
import cupy as cp
from cupyx.scipy.linalg import expm
from scipy.linalg import expm as sp_expm
import unittest.mock

dim = 10
A = np.random.randn(dim**2).reshape((dim, dim)) + 1j * np.random.randn(dim**2).reshape((dim, dim))
A = A - A.conj().T
assert np.allclose(-A, A.conj().T)

cp_unitary = expm(cp.array(A))
sp_unitary = sp_expm(A)
overlap_cp_cp = cp.abs(cp.trace(cp_unitary.conj().T @ cp_unitary))
overlap_sp_sp = np.abs(np.trace(sp_unitary.conj().T @ sp_unitary))
print(f"{overlap_sp_sp=}")  # will equal the dimension up to numberical precision
print(f"{overlap_cp_cp=}")  # should equal to dim but it will not be.

def _fixed_expm_inner(E, A, A2, A4, A6, b):
    u1 = b[13]*A6 + b[11]*A4 + b[9]*A2
    u2 = b[7]*A6 + b[5]*A4 + b[3]*A2 + b[1]*E

    v1 = b[12]*A6 + b[10]*A4 + b[8]*A2
    v2 = b[6]*A6 + b[4]*A4 + b[2]*A2 + b[0]*E
    return u1, u2, v1, v2

with unittest.mock.patch('cupyx.scipy.linalg._matfuncs._expm_inner') as mocked_fixed_expm_inner:
    mocked_fixed_expm_inner.side_effect = _fixed_expm_inner
    cp_fixed_unitary = expm(cp.array(A)).get()
[...]

PR #9142, opened 2025-05-26
BUG: fix `cupyx.scipy.linalg.expm`

ev-br:
closes gh-9138

Fix a math error in `expm` implementation, where a formula from the paper was incorrectly translated from math to python.


kmaehashi (approved):
Thanks, LGTM!

PR #9148, opened 2025-05-30
[WIP] Annotate `cupy.ndarray` and related APIs

EarlMilktea:
Add preliminary typing supports to core functionalities.

### DIFF
--- a/cupy/typing/_standalone.py
+++ b/cupy/typing/_standalone.py
@@ -10,6 +10,7 @@ from typing import (
     Literal,
     Protocol,
     SupportsIndex,
+    TypeAlias,
     TypeVar,
 )
 
@@ -62,22 +63,23 @@ class _NestedSequence(Protocol[_T_co]):
 
 
 # Miscellaneous types
+_NumpyArrayT = TypeVar("_NumpyArrayT", bound=numpy.ndarray)
 _ScalarT = TypeVar("_ScalarT", bound=numpy.generic)
-_ScalarLike_co = complex | numpy.generic
 _DTypeT_co = TypeVar("_DTypeT_co", bound=numpy.dtype, covariant=True)
 _ShapeT_co = TypeVar("_ShapeT_co", bound=tuple[int, ...], covariant=True)
 _DTypeT = TypeVar("_DTypeT", bound=numpy.dtype)
-_ShapeLike = SupportsIndex | Sequence[SupportsIndex]
-_NonNullKACF = Literal["K", "A", "C", "F"]
-_NonNullCAF = Literal["A", "C", "F"]
-_NonNullCF = Literal["C", "F"]
-_OrderKACF = _NonNullKACF | None
-_OrderCAF = _NonNullCAF | None
-_OrderCF = _NonNullCF | None
-_DTypeLike = type[_ScalarT] | numpy.dtype[_ScalarT]
-_ModeKind = Literal["raise", "wrap", "clip"]
-_SortSide = Literal["left", "right"]
-_NumpyArrayT = TypeVar("_NumpyArrayT", bound=numpy.ndarray)
+
+_ScalarLike_co: TypeAlias = complex | numpy.generic
+_ShapeLike: TypeAlias = SupportsIndex | Sequence[SupportsIndex]
+_NonNullKACF: TypeAlias = Literal["K", "A", "C", "F"]
+_NonNullCAF: TypeAlias = Literal["A", "C", "F"]
+_NonNullCF: TypeAlias = Literal["C", "F"]
+_OrderKACF: TypeAlias = _NonNullKACF | None
+_OrderCAF: TypeAlias = _NonNullCAF | None
+_OrderCF: TypeAlias = _NonNullCF | None
+_DTypeLike: TypeAlias = type[_ScalarT] | numpy.dtype[_ScalarT]
+_ModeKind: TypeAlias = Literal["raise", "wrap", "clip"]
+_SortSide: TypeAlias = Literal["left", "right"]
 
 # Typed scalars
 _FloatT = TypeVar("_FloatT", bound=numpy.floating)


kmaehashi:
Could you make this commit 5fa7f711318669501e3633fff6fb32a8cbe89b62 a separate PR? This PR sems too large and hard to review.

EarlMilktea:
> Could you make this commit [5fa7f71](https://github.com/cupy/cupy/commit/5fa7f711318669501e3633fff6fb32a8cbe89b62) a separate PR? This PR seems too large and hard to review.

Moved to #9154 .

cupy/_core/core.pyi
```diff
@@ -0,0 +1,870 @@
+from collections.abc import Callable, Iterable, Iterator, Mapping
+from typing import (
+    Any,
+    ClassVar,
+    Generic,
+    Literal,
+    SupportsIndex,
+    TypeVar,
```
jorenham: since you're using the PEP 696 `default=`, `TypeVar` should be imported  from `typing_extensions` instead (assuming that cupy supports Python < 3.13)

cupy/_core/core.pyi
```diff
@@ -0,0 +1,870 @@
...
+    @property
+    def ndim(self) -> int: ...
+    @property
+    def itemsize(self) -> int: ...
+    @property
+    def nbytes(self) -> int: ...
+    @property
+    def T(self) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
+    @property
+    def mT(self) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
```
jorenham, line 182: assuming that you won't be using `Literal` int's in shape types (as there's no use in doing so), then you could also return `typing.Self` here

cupy/typing/_array.py
```diff
@@ -0,0 +1,59 @@
...
+)
+_IntegralArrayT = TypeVar(
+    "_IntegralArrayT", bound=NDArray[numpy.integer | numpy.bool]
+)
+
+_ArrayUInt_co = NDArray[numpy.unsignedinteger | numpy.bool]
+_ArrayInt_co = NDArray[numpy.integer | numpy.bool]
+_ArrayFloat_co = NDArray[numpy.floating | numpy.integer | numpy.bool]
+_ArrayComplex_co = NDArray[numpy.inexact | numpy.integer | numpy.bool]
+_ArrayNumber_co = NDArray[numpy.number | numpy.bool]
```
jorenham: these two are almost equivalent

cupy/typing/__init__.py
```diff
@@ -1,4 +1,7 @@
-from cupy.typing._types import ArrayLike  # NOQA
-from cupy.typing._types import DTypeLike  # NOQA
-from cupy.typing._types import NBitBase  # NOQA
-from cupy.typing._types import NDArray  # NOQA
+from __future__ import annotations
+
```
jorenham, line 2: no need for this here

```suggestion
```

cupy/typing/__init__.py
```diff
@@ -1,4 +1,7 @@
...
+from numpy.typing import DTypeLike as DTypeLike
```
jorenham: this is explicitly exported in `__all__`, so no need for an implicit re-export

```suggestion
from numpy.typing import DTypeLike
```

cupy/cuda/stream.pyi
```diff
@@ -0,0 +1,67 @@
...
+
+class _BaseStream:
+    ptr: int
+    device_id: int
+
+    def __init__(self, ptr: int, device_id: int) -> None: ...
+    def __eq__(self, other: object) -> bool: ...
+    def __enter__(self) -> Self: ...
+    def __exit__(self, *args: object) -> None: ...
+    def __hash__(self) -> int: ...
```
jorenham: `__hash__` and `__eq__`  are already present in `builtins.object` (in typeshed's stdlib stubs), so there's no need to also add them here

cupy/cuda/memory.pyi
```diff
@@ -0,0 +1,95 @@
...
+    ptr: int
+    def __init__(self, mem: BaseMemory, offset: int) -> None: ...
+    def __int__(self) -> int: ...
+    @property
+    def device(self) -> Device: ...
+    # MEMO: args renamed
+    def __add__(self, offset: int) -> MemoryPointer: ...
+    def __iadd__(self, offset: int) -> Self: ...
+    def __sub__(self, offset: int) -> MemoryPointer: ...
+    def __isub__(self, offset: int) -> Self: ...
```
jorenham, line 67: It can help to make special dunder parameters like this positional only, e.g. in case of subclassing or structural typing.

cupy/cuda/stream.pyi
```diff
@@ -0,0 +1,67 @@
...
+class Stream(_BaseStream):
+    null: ClassVar[Stream]
+    ptds: ClassVar[Stream] # Missing in HIP
+
+    def __init__(
+        self,
+        null: bool = ...,
+        non_blocking: bool = ...,
+        ptds: bool = ...,
+        priority: int | None = ...,
```
jorenham, line 64: It's no problem to fill in the defaults for simple literals like `None`, `False`, `42`, etc. That way, users can directly see the defaults when they hover it in IDE's such as vscode + pylance, which in my experience is very helpful.

The (IMO bad) trend of using `= ...` in stubs is nothing but a consequence of limitations in mypy's `stubgen` tool.

cupy/cuda/stream.pyi
```diff
@@ -0,0 +1,67 @@
...
+    def __eq__(self, other: object) -> bool: ...
+    def __enter__(self) -> Self: ...
+    def __exit__(self, *args: object) -> None: ...
+    def __hash__(self) -> int: ...
+    def use(self) -> Self: ...
+    @property
+    def done(self) -> bool: ...
+    def synchronize(self) -> None: ...
+    def add_callback(
+        self, callback: Callable[[Stream, int, _T], None], arg: _T
```
jorenham, line 40: Be careful with the *contravariant* `Callable` input types here: This will currently reject it if you pass e.g.

```
def f(s: MyStreamSubclass, i: int, t: _T) -> None: ...
```

So because of the contravariance, this requires that `Stream` is assignable to `MyStreamSubclass`. But that's not the case here (it's other way around).

But I'm not familiar enough with cupy to know whether that's appropriate here, so please ignore this if this was intentional

cupy/cuda/memory.pyi
jorenham, line 1: Nitpick: It could help to annotate any immutable attribute with `typing.Final` (for the lack of a better alternative), e.g. for the `device_id`s.

cupy/cuda/graph.pyi
```diff
@@ -0,0 +1,7 @@
+from cupy.cuda.stream import Stream
+
+
+class Graph:
+    def launch(self, stream: Stream| None = ...) -> None: ...
```
jorenham:
```suggestion
    def launch(self, stream: Stream | None = ...) -> None: ...
```

(although using `= None` as default would be better here)

cupy/cuda/device.pyi
```diff
@@ -0,0 +1,38 @@
...
+def get_cusolver_sp_handle() -> int: ...
+def get_cusparse_handle() -> int: ...
+def get_compute_capability() -> str: ...
+
+class Device:
+    id: int
+
+    def __init__(self, device: int | None = ...) -> None: ...
+    @classmethod
+    def from_pci_bus_id(cls, pci_bus_id: str) -> Device: ...
```
jorenham: `Self` also works works with `@classmethod` :)

```suggestion
    def from_pci_bus_id(cls, pci_bus_id: str) -> Self: ...
```

cupy/_core/internal.pyi
```diff
@@ -0,0 +1,12 @@
+# TODO: Add entries when necessary
+
+from typing import Any
+import cupy
+from cupy.typing._array import NDArray
+
+def _get_strides_for_order_K(
+    x: NDArray[Any], dtype: cupy.dtype, shape: list[int] | None = ...
```
jorenham: shape types are usually tuples, is this correct?

cupy/_core/flags.pyi
```diff
@@ -0,0 +1,21 @@
...
+    @property
+    def owndata(self) -> bool: ...
+
+    @property
+    def fnc(self) -> bool: ...
+
+    @property
+    def forc(self) -> bool: ...
+
+    def __getitem__(self, name: str) -> bool: ...
```
jorenham:
```suggestion
    def __getitem__(self, name: str, /) -> bool: ...
```

cupy/_core/core.pyx
```diff
@@ -1617,10 +1617,9 @@ cdef class _ndarray_base:
             'Please use `.get()` to construct a NumPy array explicitly.')
 
     @classmethod
-    def __class_getitem__(cls, tuple item):
-        from types import GenericAlias
-        item1, item2 = item
-        return GenericAlias(ndarray, (item1, item2))
+    def __class_getitem__(cls, _):
+        # Discard generic parameters on runtime.
+        return cls
```
jorenham: Runtime type-checkers won't like that:

```pycon
>>> import numpy as np
>>> np.__version__
'2.3.0'
>>> np.dtype
<class 'numpy.dtype'>
>>> np.dtype[np.float64]
numpy.dtype[numpy.float64]
```

So this could cause issues for users of e.g. [beartype](https://github.com/beartype/beartype) or [typeguard](https://github.com/agronholm/typeguard)

---

If you don't mind eagerly importing `types`, then you could use this little trick:

```py
class _ndarray_base:
   # --snip--

    __class_getitem__ = classmethod(types.GenericAlias)
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,870 @@
...
+    ) -> numpy.float64: ...
+    @overload
+    def std(
+        self,
+        axis: _ShapeLike | None = ...,
+        dtype: DTypeLike | None = ...,
+        out: None = ...,
+        ddof: float = ...,
+        keepdims: bool = ...,
+    ) -> Any: ...
```
jorenham, line 667: These parameters don't look very compatible. I'm not sure what the intention is here, but I'm guessing that you want `ddof` to be keyword-only here.

The `var` method above here has the same issue btw

cupy/_core/core.pyi
```diff
@@ -0,0 +1,870 @@
...
+    @overload
+    def __abs__(self: NDArray[numpy.complex64]) -> NDArray[numpy.float64]: ...
+    @overload
+    def __abs__(self: NDArray[numpy.complex128]) -> NDArray[numpy.float64]: ...
+    def __invert__(self: _IntegralArrayT) -> _IntegralArrayT: ...
+    # TODO: Annotate binary operators
+    def conj(self) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
+    def conjugate(self) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
+    real: ClassVar[_SupportsRealImag]
+    imag: ClassVar[_SupportsRealImag]
```
jorenham, line 809: Mypy and pyright special cased `@property`, so there's a chance that this custom descriptor won't work as intended here (i.e. similar to a `@property`).

But either way, I think this is a pretty neat way to work around `@property` not being overloadable. It's not the only way, though :)

cupy/_core/core.pyi
```diff
@@ -0,0 +1,870 @@
...
+    def any(
+        self,
+        axis: int | tuple[int, ...] | None = ...,
+        *,
+        out: _ArrayT,
+        keepdims: SupportsIndex = ...,
+    ) -> _ArrayT: ...
+    def __nonzero__(self) -> bool: ...
+    def __neg__(self: _NumericArrayT) -> _NumericArrayT: ...
+    def __pos__(self: _NumericArrayT) -> _NumericArrayT: ...
```
jorenham: Aren't all cupy arrays numeric?

cupy/_core/core.pyi
```diff
@@ -0,0 +1,870 @@
...
+    @overload
+    def __abs__(self: _RealArrayT) -> _RealArrayT: ...
+    @overload
+    def __abs__(self: NDArray[numpy.complex64]) -> NDArray[numpy.float64]: ...
```
jorenham: Typo, or intentional?

```suggestion
    def __abs__(self: NDArray[numpy.complex64]) -> NDArray[numpy.float32]: ...
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,870 @@
...
+    @overload
+    def astype(
+        self,
+        dtype: DTypeLike,
+        order: _OrderKACF = ...,
+        copy: bool = ...,
+    ) -> ndarray[Any, numpy.dtype[Any]]: ...
+    def copy(
+        self, order: _OrderKACF = ...
+    ) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
```
jorenham, line 212:
```suggestion
    def copy(
        self, *, order: _OrderKACF = ...
    ) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
```

jorenham (commented):
I left a couple of comments; I hope you don't mind.

kmaehashi:
Thanks for joining the review, @jorenham!
@EarlMilktea Could you take a look?

We are also curious how to test these type stubs. Does NumPy or SciPy stub have any kind of tests to ensure consistency with codebase?

cupy/_core/core.pyi
EarlMilktea: > not the only way

I'm personally very interested in the other ways.
Could you show me some?

cupy/_core/core.pyi
EarlMilktea: I once used `Self` for annotations but erased most of them because shape supports can impact them in the future.
Specifically for the annotations here you're right.

cupy/typing/__init__.py
EarlMilktea: I won't remove it because `from __future__ import annotations` are everywhere enforced by the CI so that deprecated annotation utilities are never used.

jorenham:
> Does NumPy or SciPy stub have any kind of tests to ensure consistency with codebase?

In `scipy-stubs` we mostly rely on mypy's *stubtest* for this. It can be rather strict and sometimes kinda frustrating, but it's absolutely worth it. To get started, you can have it generate an "allowlist", that's basically a list of all known stubtests errors. That way, when running stubtest in CI, you pass that allowlist and it will ignore those errors, and report an error if there are unused allowlist entries. This way, you can incrementally work on writing the stubs until the allowlist is no longer needed. 
Currently we're not using stubtest in NumPy itself, but [numtype](https://github.com/numpy/numtype) (an experimental rework of numpy's stubs) we do (and have minimized the [allowlist](https://github.com/numpy/numtype/blob/6d2c612264d5770d91a7c8385dd1308b720203c4/tool/allowlists/common.txt)).

---

Scipy-stubs, NumPy, and NumType also use manual *type-testing*. We do this by writing `.pyi` "tests" that are statically validated using type-checkers (so with e.g. pyright and mypy). You don't need something like pytest to do this. There are two main flavors of test:

- Acceptance tests, which roughly speaking are a bunch of `typing.assert_type` statements, which ensure that certain input types result in the expected output types. Usually writing these by hand is most appropriate, but there are also situations where it makes sense to use code generation.
- Rejection tests, which verify that type-checkers report an error in case of unwanted input. The trick is to configure mypy and pyright to report an error for unused `# type: ignore[...]` or `# pyright: ignore[...]`. It is also important that they require specific error codes in the type-ignore's, so [blanket type ignores](https://docs.astral.sh/ruff/rules/blanket-type-ignore/) should not be allowed. With this, you can write `only_ints_pls("sleventy")  # type: ignore[arg-type]` (assuming you just use mypy) to ensure that your `only_ints_pls` function rejects strings. Because if it doesn't, then that `# type: ignore` will be reported as an unused ignore.

Some type-testing examples:

- https://github.com/numpy/numtype/tree/main/src/numpy-stubs/%40test/static (very similar to numpy)
- https://github.com/numpy/numtype/tree/main/src/numpy-stubs/%40test/generated
- https://github.com/scipy/scipy-stubs/tree/master/tests

Some type-checker configurations that work:

- https://github.com/numpy/numtype/blob/6d2c612264d5770d91a7c8385dd1308b720203c4/pyproject.toml#L111-L158
- https://github.com/scipy/scipy-stubs/blob/ce8466b17d7b7233751b4e010dc53f9a2e599082/pyproject.toml#L106-L146
- https://github.com/numpy/numpy/blob/main/numpy/typing/tests/data/mypy.ini

---

It's also a good idea to not rely on one type-checker (like numpy currently does). I'd recommend using at least mypy and pyright (used by pylance in vscode) at the moment, which is what most users are currently using. In situations where where mypy and pyright disagree with each other, then that's usually a bug, and knowing that is pretty important (and in most cases there are workarounds). Sometimes it's not a bug, but be undefined behavior (given the [typing spec](https://typing.python.org/en/latest/spec/index.html)), and you probably want to avoid that. Generally speaking it's a good idea to minimize the differences between type-checkers.

cupy/cuda/stream.pyi
EarlMilktea: I decided to use `...` because default value annotations are essentially duplicates of the original sources (and thus we manually keep the consistency between them) though I admit `...` could be sometimes user-unfriendly.

cupy/_core/core.pyi
EarlMilktea: Currently yes but not in the future.
@kmaehashi Am I correct?

cupy/_core/internal.pyi
EarlMilktea: This function is called with `list[int]`, for example in `cupy/_creation/basic.py` .
(Actually I'm tempted to use `collections.abc` stuff here but also not sure whether it's appropriate or not for cython modules...)

cupy/_core/core.pyx
EarlMilktea: LGTM. Let me try.
(Note that `Generic` inheritance didn't work here)

cupy/_core/core.pyi
EarlMilktea: Is it for readability? I suspect `.copy(True)` (or similar positional calls) is not prohibited in the source.

cupy/_core/core.pyi
EarlMilktea: I added these special overloads to provide non-`Any` return values for the most conventional usages, such as:

```py
x = cp.asarray([0, -1, 3], dtype=cp.float64)
print(x.std(ddof=1))
```

cupy/cuda/stream.pyi
EarlMilktea: I completely forgot about contravariance. Thank you for the comment!
Anyway, for that case, what are correct alternatives to `Stream` here?

cupy/cuda/stream.pyi
EarlMilktea: Fixed. Thanks!

cupy/typing/_array.py
EarlMilktea: Removed the latter.

cupy/_core/core.pyi
EarlMilktea: Fixed. Thanks!

cupy/_core/core.pyi
jorenham: That seems like a good idea to me. But at the moment, calling `x.std(1)` is ambiguous, as type-checkers do not know whether `1` is passed to `ddof` or `axis`. By requiring the `ddof` in the first overload to be keyword only, the ambiguity disappears.

cupy/cuda/memory.pyi
EarlMilktea: I would like to know your own (or Python standards, if any) criteria of whether or not adding `/` for dunders, since for instance blindly adding `/` for `__init__` is likely to be what you wanted here.

cupy/_core/core.pyx
EarlMilktea: It's likely to be working (at least no compilation errors). Thank you!

cupy/_core/core.pyi
jorenham: I think the "easiest" method would be to defer to the scalar types themselves, which also implement `.real` and `.imag` (in numpy). Roughly speaking that could look something like this (using Python 3.12+ syntax)

```pyi
# -- phantom types --

@typing.type_check_only
class HasReal[RT](typing.Protocol):
    @property
    def real(self, /) -> RT: ...

@typing.type_check_only
class HasType[T](typing.Protocol):  # T should not have an upper bound
    @property
    def type(self, /) -> type[T]: ...

@typing.type_check_only
class HasDType[DT](typing.Protocol):  # DT should not have an upper bound
    @property
    def dtype(self, /) -> DT: ...

type ScalarType[T] = HasDType[HasType[T]]  # _.dtype.type

# -- stubs --

class ndarray:
   @property
   def real[RT](self: ScalarType[HasReal[RT]], /) -> RT: ...
```

So this returns `_T.dtype.type.real` s.t. `self: _T`, and uses polymorphism rather than overloading.

This trick can also be useful for methods like `__abs__()` and `.item()` btw :)

cupy/_core/core.pyi
EarlMilktea: Understood your implication. Fixed, thanks!

cupy/_core/core.pyi
jorenham: Oh huh I don't know why I suggested this actually. `order` can indeed be passed positionally (in numpy at least), so there should indeed be no `*` here.

cupy/_core/core.pyi
jorenham: I guess it depends on what a "shape type" means. If you only consider it to describe the rank (ndim), then this `Self` is correct. But it also describes individual axes (which currently isn't possible to do in a useful way, btw), then it isn't.

cupy/typing/__init__.py
jorenham: Ah I see. Just keep in mind that runtime type-checkers might have some trouble with it.

cupy/cuda/memory.pyi
jorenham: I'm not aware of any official criteria for this, but I know that mypy's stubtest doesn't mind it when you use `self, /` in stub methods (even if it doesn't match runtime). It also allows it for `@property` setters value parameters, and also for most arithmetic `__dunder__` methods (I'm not sure about the 3rd `__pow__` and `__round__` params). And strangely, stubtest doesn't like it if you `cls, /` in a `@classmethod` for some reason.

But to be honest, I usually just try it, and then see what stubtest has to say about it 🤷🏻. So if you're not using it, then I suppose it's the dealer's choice.

cupy/cuda/stream.pyi
jorenham: Note that the typing spec explicitly recommends annotating defaults, as well: https://typing.python.org/en/latest/guides/writing_stubs.html#functions-and-methods

> For default values, use the literal values of “simple” default values (`None`, bools, ints, bytes, strings, and floats). Use the ellipsis literal `...` in place of more complex default values.

cupy/cuda/stream.pyi
jorenham: > I completely forgot about contravariance. Thank you for the comment!

Yea it's an easy one to miss. Stubtest also wouldn't have helped, so that's why type-testing is a good idea.

> Anyway, for that case, what are correct alternatives to Stream here?

I'm not really familiar with cupy I'm afraid. Are there a finite amount of allowed `Steam` subtypes that are accepted? Because in that case, you could union several `Callable` types with different inputs. Otherwise, the easy solution would be to use `Any` or `Never`. But I suppose you could also write a **generic** protocol with a (non-generic) `__call__` method.

cupy/_core/internal.pyi
jorenham: If it's always a list of ints, then this is no problem. But in case it could also accept a `list[bool]` or something, then a `Sequence` might indeed be more appropriate.

cupy/_core/core.pyx
jorenham: > Note that `Generic` inheritance didn't work here

Yea there's a lot of magic involved with `Generic`, and I can imagine that not being very cython-friendly.

cupy/_core/core.pyi
asi1024: We are considering supporting string dtypes and structured arrays in the future: https://github.com/cupy/cupy/pull/7462

jorenham: Well, good luck typing that then 😅

cupy/typing/_standalone.py
```diff
@@ -0,0 +1,90 @@
...
+_ShapeLike = SupportsIndex | Sequence[SupportsIndex]
+_NonNullKACF = Literal["K", "A", "C", "F"]
+_NonNullCAF = Literal["A", "C", "F"]
+_NonNullCF = Literal["C", "F"]
+_OrderKACF = _NonNullKACF | None
+_OrderCAF = _NonNullCAF | None
+_OrderCF = _NonNullCF | None
+_DTypeLike = type[_ScalarT] | numpy.dtype[_ScalarT]
+_ModeKind = Literal["raise", "wrap", "clip"]
+_SortSide = Literal["left", "right"]
```
jorenham: This may help a bit

```suggestion
_ScalarLike_co: TypeAlias = complex | numpy.generic
_DTypeT_co = TypeVar("_DTypeT_co", bound=numpy.dtype, covariant=True)
_ShapeT_co = TypeVar("_ShapeT_co", bound=tuple[int, ...], covariant=True)
_DTypeT = TypeVar("_DTypeT", bound=numpy.dtype)
_ShapeLike: TypeAlias = SupportsIndex | Sequence[SupportsIndex]
_NonNullKACF: TypeAlias = Literal["K", "A", "C", "F"]
_NonNullCAF: TypeAlias = Literal["A", "C", "F"]
_NonNullCF: TypeAlias = Literal["C", "F"]
_OrderKACF: TypeAlias = _NonNullKACF | None
_OrderCAF: TypeAlias = _NonNullCAF | None
_OrderCF: TypeAlias = _NonNullCF | None
_DTypeLike: TypeAlias = type[_ScalarT] | numpy.dtype[_ScalarT]
_ModeKind: TypeAlias = Literal["raise", "wrap", "clip"]
_SortSide: TypeAlias = Literal["left", "right"]
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+from cupy.cuda.device import Device
+from cupy.cuda.memory import MemoryPointer
+from cupy.cuda.stream import Stream
+from cupy.typing import DTypeLike
+from cupy.typing._standalone import (
+    _T,
+    _Buffer,
+    _DTypeLike,
+    _DTypeT,
+    _DTypeT_co,
```
jorenham, line 30: Nit: Personally I try to avoid importing `TypeVar`s from other modules. Mostly because theoretically type-parameters only have meaning within the context of the given generic type or generic function. And once you drop Python<3.12 it'll also be easier to migrate to [PEP 695](https://peps.python.org/pep-0695/) syntax if you don't have to worry about your `TypeVar`s being imported in some other module.

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+    _ScalarLike_co,
+    _ScalarT,
+    _ShapeLike,
+    _SortSide,
+    _SupportsFileMethods,
+)
+
+# TODO: tuple vs. Sequence
+
+@typing.runtime_checkable
```
jorenham: Since this doesn't exist at runtime, the `@runtime_checkable` is meaningless. 
I'd instead use `@type_check_only` for such protocols and types (which only exists if `TYPE_CHECKING`), otherwise some IDE's might think this exists at runtime and list it as an auto-completion suggestion or something.

```suggestion
@typing.type_check_only
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+)
+_DualArrayLike: TypeAlias = (
+    _SupportsArray[_DTypeT]
+    | _NestedSequence[_SupportsArray[_DTypeT]]
+    | _T
+    | _NestedSequence[_T]
+)
+
+# Anything castable to ndarray with specified dtype (work in progress)
+_ArrayLikeBool_co: TypeAlias = _DualArrayLike[numpy.dtype[numpy.bool], bool]
```
jorenham, line 64: I'm not sure if cupy still supports `numpy<2`, but if it does, then it'd be safer to use `bool_`:

```suggestion
_ArrayLikeBool_co: TypeAlias = _DualArrayLike[numpy.dtype[numpy.bool_], bool]
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+_ArrayLikeInt_co: TypeAlias = _DualArrayLike[
+    numpy.dtype[numpy.bool | numpy.integer], int
+]
+_ArrayLikeFloat_co: TypeAlias = _DualArrayLike[
+    numpy.dtype[numpy.bool | numpy.integer | numpy.floating], float
+]
+_ArrayLikeComplex_co: TypeAlias = _DualArrayLike[
+    numpy.dtype[numpy.bool | numpy.number], complex
+]
+_ArrayLikeNumber_co = _ArrayLikeComplex_co
```
jorenham, line 77: In `numpy._typing` this alias only exists (internal) backwards-compatibility reasons, and I'll probably get rid of it sometime soon. So you might as well remove it here as well.

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+
+# Index-like
+_ToIndex: TypeAlias = (
+    SupportsIndex | slice | EllipsisType | _ArrayLikeInt_co | None
+)
+_ToIndices: TypeAlias = _ToIndex | tuple[_ToIndex, ...]
+
+# numpy.typing.ArrayLike minus str/bytes
+ArrayLike: TypeAlias = _Buffer | _DualArrayLike[numpy.dtype[Any], complex]
+NDArray: TypeAlias = ndarray[Any, numpy.dtype[_ScalarT]]
```
jorenham, line 87: The difference is subtle, but this can matter quite a bit for shape-typing (theoretically this shouldn't matter because of the ["gradual guarantee"](https://typing.python.org/en/latest/spec/concepts.html#the-gradual-guarantee), but in practice that unfortunately doesn't hold, at least not in case of mypy and pyright).

```suggestion
NDArray: TypeAlias = ndarray[tuple[Any, ...], numpy.dtype[_ScalarT]]
```

This `tuple[Any, ...]` is what I call a "gradual shape-type", which, unlike `tuple[int, ...]` is assignable to specific shape-types like `tuple[int]` (`ndim == 1`) and `tuple[int, int, *tuple[int, ...]]` (`ndim >= 2`).

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+_ArrayUInt_co: TypeAlias = NDArray[numpy.unsignedinteger | numpy.bool]
+_ArrayInt_co: TypeAlias = NDArray[numpy.integer | numpy.bool]
+_ArrayFloat_co: TypeAlias = NDArray[
+    numpy.floating | numpy.integer | numpy.bool
+]
+_ArrayComplex_co: TypeAlias = NDArray[
+    numpy.inexact | numpy.integer | numpy.bool
+]
+
+class _SupportsRealImag(Protocol):
```
jorenham, line 110:
```suggestion
@typing.type_check_only
class _SupportsRealImag(Protocol):
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+    def __get__(
+        self, instance: NDArray[numpy.complex128], owner: type | None = ...
+    ) -> NDArray[numpy.float64]: ...
+    @overload
+    def __get__(
+        self, instance: None, owner: type | None = ...
+    ) -> _SupportsRealImag: ...
+    def __set__(self, instance: NDArray[Any], value: ArrayLike) -> None: ...
+
+@typing.runtime_checkable
```
jorenham:
```suggestion
@typing.type_check_only
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+_ShapeWithDefaultT_co = TypeVar(
+    "_ShapeWithDefaultT_co",
+    bound=tuple[int, ...],
+    default=tuple[Any, ...],
+    covariant=True,
+)
+_DTypeWithDefaultT_co = TypeVar(
+    "_DTypeWithDefaultT_co",
+    bound=numpy.dtype,
+    default=numpy.dtype[Any],
```
jorenham: Note that since numpy 2.2 (or perhaps 2.3, I'm not sure), _: numpy.dtype equivs numpy.dtype[Any]. Before that, its type parameter did not have a default (viz. Any). So on earlier numpy versions this will cause mypy and pyright to complain about a missing generic type argument (in strict mode).

```suggestion
    bound=numpy.dtype[Any],
    default=numpy.dtype[Any],
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+)
+
+# MEMO: Some methods have special overloads for most-conventional use cases.
+
+# TODO: Add shape support (currently Any)
+class ndarray(Generic[_ShapeWithDefaultT_co, _DTypeWithDefaultT_co]):
+    def __init__(
+        self,
+        shape: _ShapeLike,
+        dtype: DTypeLike = ...,
```
jorenham: Since NumPy 2.4 `numpy.typing.DTypeLike` no longer accepts `None`

```suggestion
        dtype: DTypeLike | None = ...,
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+        memptr: MemoryPointer | None = ...,
+        strides: _ShapeLike | None = ...,
+        order: str | None = ...,
+    ) -> None: ...
+    # Attributes
+    base: ndarray[Any, _DTypeWithDefaultT_co] | None
```
jorenham, line 159: This is of same shape, no?

```suggestion
    base: ndarray[_ShapeWithDefaultT_co, _DTypeWithDefaultT_co] | None
```

If not, then I'd write this as 

```suggestion
    base: ndarray[tuple[Any, ...], _DTypeWithDefaultT_co] | None
```

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+    dtype: _DTypeWithDefaultT_co
+    size: int
+    # TODO: Annotate dlpack interface
+    def __cuda_array_interface__(self) -> dict[str, Any]: ...
+    @property
+    def flags(self) -> Flags: ...
+    @property
+    def shape(self) -> tuple[int, ...]: ...
+    @shape.setter
+    def shape(self, newshape: _ShapeLike) -> None: ...
```
jorenham: nit:

```suggestion
    def shape(self, newshape: _ShapeLike, /) -> None: ...
```

jorenham: but bear in mind that allowing users to modify the shape in-place is very type-unsafe. For example:

```py
a: ndarray[tuple[int], dtype[int8]] = arange(4)  # 1d
a.shape = (2, 2)  # currently allowed

def only_1d_pls(vec: ndarray[tuple[int], dtype[int8]]):
    assert vec.ndim == 1

only_1d_pls(a)  # type-checkers will allow this
```

To avoid this, you could either remove this setter altogether, or annotate it as something like  `newshape: typing_extensions.Never` (or equivalently as `typing.NoReturn`)

cupy/_core/core.pyi
```diff
@@ -0,0 +1,975 @@
...
+    @property
+    def ndim(self) -> int: ...
+    @property
+    def itemsize(self) -> int: ...
+    @property
+    def nbytes(self) -> int: ...
+    @property
+    def T(self) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
+    @property
+    def mT(self) -> ndarray[Any, _DTypeWithDefaultT_co]: ...
```
jorenham, line 182: Although I'd personally go with `Self` here, if you don't want to include the shape-type then it'd be better to use `tuple[Any, ...]`:

```suggestion
    def T(self) -> ndarray[tuple[Any, ...], _DTypeWithDefaultT_co]: ...
    @property
    def mT(self) -> ndarray[tuple[Any, ...], _DTypeWithDefaultT_co]: ...
```

It might be handy to have an `_AnyShape: TypeAlias = tuple[Any, ...]`, like we have in numpy as `numpy._typing._AnyShape`.

EarlMilktea:
@jorenham 

Thank you for the comments! Let me examine later.
For the shape supports I'd say I'm tempted to keep them as simple as possible (simple `Any` everywhere etc.), so that we can prototype preliminary typing supports as soon as possible.

cupy/_core/core.pyi
EarlMilktea: I have exactly the same opinion here and would like to use PEP 695 if possible.
Do you think PEP 695 is now mature enough? I suppose it wasn't when I started to compose this PR.

@kmaehashi Any comments?

cupy/_core/core.pyi
EarlMilktea: It no longer does. @kmaehashi Am I correct?

cupy/_core/core.pyi
EarlMilktea: Nice catch, thank you (they're all from my dirty copy-pasting, my apologies)!

Issue #8998, opened 2025-02-28
Support free-threaded build Python


kmaehashi:
Provide a support for Python 3.13 free-threaded build.

At least Cython 3.1.0a1 is required for Python 3.13t.
https://cython.readthedocs.io/en/latest/src/changes.html#alpha-1-2024-11-08

leofang:
Possibly a blocker: https://github.com/scoder/fastrlock/issues/19

bdice:
@leofang I am looking a bit at fastrlock. I see we use its C API. Do you (or others) know if cupy require C API performance here?

I began to look at making fastrlock compatible with free-threading but I am unsure if the C API performance (via Cython) is required, or if it would be better to move to `threading.RLock`. The speed would go down a bit but it would allow CuPy to drop a dependency.

kmaehashi:
Locking is required for each de/allocation, so it is a hot path. I'm under the impression that calling Python's `threading.RLock` from our memory allocator written in Cython will increase the overhead, but can't say anything without numbers.

bdice:
Thanks @kmaehashi, that was my thought as well. I think updating fastrlock is probably the cleanest solution to ensure overhead is minimal. I have a draft locally that I think I can make into a PR for fastrlock. I still need to iron out a few details.

kmaehashi:
I've made an experimental branch #9149 to test `threading.RLock` performance.

ngoldbaum:
Hi all. I'm one of the main authors of https://py-free-threading.github.io, which might be useful as the maintainers of cupy think about supporting the free-threaded build.

> Possibly a blocker: https://github.com/scoder/fastrlock/issues/19

Starting with Python 3.14 fastrlock might not be necessary, since @kumaraditya303 updated `threading.Rlock` to be backed by a `PyMutex`, which is substantially faster than the old platform-specific code that was there before. https://github.com/python/cpython/pull/125144/. Probably worth benchmarking whether fastrlock is necessary on 3.14.

kmaehashi:
Hi @ngoldbaum, thanks you for maintaining the tracker and adding CuPy to the list! It's also great to know that the `threading.RLock` performance has improved. I will try benchmarking again using Python 3.14rc1 (c.f. [benchmark results in Python 3.13](https://github.com/cupy/cupy/pull/9149#issuecomment-2993562268)), but I'm wondering that the Python overhead can be still a bottleneck since we are currently directly calling fastrlock's C API  from our Cython code.

[...]

PR #9149, opened 2025-05-30
[DNM] Experiment: Python 3.13 free-threaded build

kmaehashi:
xref #8998

Do not merge. This PR intends to:
* Try running all CuPy tests under no-GIL mode
* See whether we can use `threading.RLock` instead of `fastrlock` in terms of performance


bdice:
@kmaehashi I was curious how this effort was going. If `threading.RLock` is fast enough, we don't need to update `fastrlock` to support free-threading.

leofang:
@kmaehashi sorry I somehow missed this PR 😓 

I don't think we're testing the real free-threading build, though. Most likely we're still testing the single-threaded Cython extension modules. We need to build CuPy with the `freethreading_compatible` directive turned on, e.g.
```python
from Cython.Compiler.Version import version as cython_version
from packaging.version import Version

compiler_directives = {}
if Version(cython_version) >= Version("3.1.0"):
    compiler_directives["freethreading_compatible"] = True

setup(
    ext_modules=cythonize(
        extensions,
        compiler_directives=compiler_directives,
    )
)
```
xref:
- https://py-free-threading.github.io/porting-extensions/#declaring-free-threaded-support
- https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives

kmaehashi:
Thanks for comments @bdice and @leofang!

I've ran micro benchmark that repeatedly allocates/frees memory allocation:

```py
import timeit
import cupy


def bench(p):
    for i in range(100):
        m = p.malloc(1024)
        del m

p = cupy.get_default_memory_pool()
bench(p)  # warmup
print(timeit.repeat('bench(p)', number=1000, globals=globals()))
```

And here is the output:

| Commit | Python | Result |
| ------ | ------ | ------ |
| main (081a76f8c) | 3.13.3 | `[0.12636320595629513, 0.1284299350809306, 0.1285113380290568, 0.13066443102434278, 0.12848363094963133]` |
| This PR (a58675b81) | 3.13.3 | `[0.34424653300084174, 0.3429534309543669, 0.34284601896069944, 0.3441855148412287, 0.34328238596208394]` |
| This PR (a58675b81) | 3.13.3t | `[0.42291335109621286, 0.42466054391115904, 0.42652632179670036, 0.428276525111869, 0.42930753203108907]` |

The result obviously indicates there is a need for fastrlock's C API. As for the "This PR + 3.13.3t" combination, this apparently due to the [overhead for single-threaded use-cases in free-threaded build](https://docs.python.org/3/howto/free-threading-python.html#single-threaded-performance).

> We need to build CuPy with the `freethreading_compatible` directive turned on

I used `python -X gil=0` (which enforces Python to ignore "freethreading-incompatible" flag) when testing so I think it's running under non-GIL mode. Anyways I added `freethreading_compatible=1` flag to this PR 😃

leofang:
Superseded by #9861 (cc @seberg for vis).

PR #9150, opened 2025-05-31
Use SCM version to indicate development versions

kmaehashi:
Use [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/) to generate the version automatically.

Supersedes #3478.


EarlMilktea:
@kmaehashi 

How about setting up `.git_archival.txt` and `.gitattributes` ?

Please check https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives for details.

kmaehashi:
This requires some additional things to work:

* Compatibility with `actions/checkout` (which does not check out tags by default)
* An option to force specifying the version when building release wheels

Let me close until I have a bandwidth to work on this.

Issue #4557, opened 2021-01-18
Add random distributions to new Generator


emcastillo:
The current generator lacks of several methods

- [ ] integers
- [ ] random
- [ ] choice
- [ ] bytes
- [ ] shuffle
- [ ] permutation
- [ ] beta
- [ ] binomial
- [ ] chisquare
- [ ] dirichlet
- [ ] exponential
- [ ] f
- [ ] gamma
- [ ] geometric
- [ ] gumbel
- [ ] hypergeometric
- [ ] laplace
- [ ] logisitic
- [ ] lognormal
- [ ] logseries
- [ ] multinomial
- [ ] multivariate_hypergeometric
- [ ] multivariate_normal
- [ ] negative_binomial
- [ ] noncentral_chisquare
- [ ] noncentral_f
- [ ] normal
- [ ] paretto
- [ ] poisson
- [ ] power
- [ ] rayleigh
- [ ] standard_cauchy
- [ ] standard_exponential
- [ ] standard_gamma
- [ ] standard_normal
- [ ] standard_t
- [ ] triangular
- [ ] uniform
- [ ] vonmises
- [ ] wald
- [ ] weibull
- [ ] zipf

povinsahu1909:
Hi @emcastillo, could you please guide me on how to add these non implemented distributions. As I got confused between Generators and RandomState.

emcastillo:
Hi, thanks!
I will send a sample PR so you can get some guidance.

emcastillo:
You can check https://github.com/cupy/cupy/pull/4885
for reference!

povinsahu1909:
Currently, I am trying to add poisson and binomial distribution. Should I wait until #4904 is merged?

emcastillo:
neither poisson or binomial supports out kwarg so you don't need to wait
https://numpy.org/doc/stable/reference/random/generator.html

emcastillo:
Hi!, I changed a bit the way tests work in #4907 but you can just post your PRs now and I can take care of them later :)

povinsahu1909:
I have made PR #4915 for an exponential distribution.

povinsahu1909:
Can I take this issue for my GSoC project?

takagi:
We want to have these distributions earlier than GSoC period, if possible, half of them before v9 release (Apr. 22)

kmaehashi:
@emcastillo Could you update the list to the latest status?

anurag629:
[...]

PR #9151, opened 2025-06-04
adding permuted function to cupy

ChunkyPanda03:
Just want to add this function to cupy it is useful for shuffling individual rows of a matrix which is available in [numpy](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.permuted.html) it is fairly close to the permutation method but sorts by rows not by the whole array.

#4557

I was having problems as it was saying that my new method permuted was not found I could use some guidance on this end. I will attach a python script.
[test.py.txt](https://github.com/user-attachments/files/20598359/test.py.txt)
 that shows its use just scroll to bottom

### DIFF
--- a/cupy/random/__init__.py
+++ b/cupy/random/__init__.py
@@ -97,6 +97,7 @@ from cupy.random._generator import RandomState  # NOQA
 from cupy.random._generator import reset_states  # NOQA
 from cupy.random._generator import seed  # NOQA
 from cupy.random._generator import set_random_state  # NOQA
+from cupy.random._permutations import permuted  # NOQA
 from cupy.random._permutations import permutation  # NOQA
 from cupy.random._permutations import shuffle  # NOQA
 from cupy.random._sample import choice  # NOQA
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -1183,10 +1183,25 @@ class RandomState(object):
         array = cupy.argsort(sample)
         return array
 
+    def permuted(a:cp.ndarray,axis=0,type=cp.int32):
+	"""Returns a pertated array for multi dimensional array"""
+        #added type to tweek the sample datatype if you run out of memory
+        #check axis
+        if axis > a.ndim or axis<0:
+            raise TypeError('Axis not valid')
+        num = a.shape
+
+        sample = cp.empty(num,dtype=type) #empty memory of equal shape to array a
+        curand.generate(self._generator,sample.data.ptr,num[axis]) #give data and number to generator to rearrange in place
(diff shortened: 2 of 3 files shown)


ChunkyPanda03:
Please let me know how I can improve this PR I haven't really done any for cupy so I don't know the requirements. I just saw this feature would be easy to implement.

ChunkyPanda03:
also as a note in _permuted the sample array uses a 32 bit integer type for shuffling the array I find that in my case that this is bloated and uses lots of memory. I think that maybe a method for determining the minimum type required to efficiently shuffle based on the size of the axis or just using a smaller datatype and shuffling it multiple times would be a better approach (I think this would be best).

kmaehashi:
`numpy.random.RandomState` does not have `permuted`. #4557 is for new `cupy.random.Generator` API:
https://github.com/cupy/cupy/blob/main/cupy/random/_generator_api.pyx

> Please let me know how I can improve this PR I haven't really done any for cupy so I don't know the requirements. I just saw this feature would be easy to implement.


Please refer to this documentation for the guidance. https://docs.cupy.dev/en/latest/contribution.html

ChunkyPanda03:
@kmaehashi I see now I have to edit the random __INIT__.py to import the permuted method from _permuted, _permutation.py add a call to permuted in _generator.py. I will also clean it up to look a bit nicer

ChunkyPanda03:
@kmaehashi if you take a look at the changes I made I still get 
```Python
temp = cp.random.default_rng()
temp.permuted(temp2,1)
```
```
AttributeError                            Traceback (most recent call last)
Cell In[19], line 1
----> 1 temp.permuted(temp2,1)

AttributeError: 'Generator' object has no attribute 'permuted'
```
I did find that I can access it here 
```Python
cp.random.RandomState(10210234,100).permuted(temp2,1)
```

ChunkyPanda03:
ok I just got it the #4557 is for adding to the _generator_api.pyx and glued to c++ code. Yeah that is beyond the scope of what I can do.

ChunkyPanda03:
I know that I need to format this to fit pep8 standards but if someone could let me know if this is sound or it needs more work otherwise

ChunkyPanda03:
May have found a major memory allocation error when making this #9170 I don't know if it is directly tied to this method so I don't know if this can be merged before the other issue is resolved.

kmaehashi:
Closing as there are no plan to add `cupy.random.permuted` nor `cupy.random.RandomState.permuted` in CuPy, because we are following the NumPy specification and there are no `numpy.random.permuted` nor `numpy.random.RandomState.permuted`.

PR #9152, opened 2025-06-05
Fix rocThrust build on ROCm 6.4

Azusachan:
I've checked #9099 and it seems we need to define an extra macro to pretend it's HIP to mitigate the issue from https://github.com/ROCm/rocThrust/blob/cce4f169de50228a401df104d97b7c90433a87dd/thrust/detail/config/config.h#L24. By using this method we don't need to define THRUST_VERSION 0, which makes it a bit less hacky.


kmaehashi:
I think `__HIP__` macro must only be defined by the device compiler and should not be used during host code compilation.

Issue #8303, opened 2024-04-19
RFC: Seeking feedback on adding new special functions from SciPy


steppi:
Hi CuPy devs!

You may have seen that in `scipy.special` we're working on translating the scalar kernels for special functions into C++ which can run on both CPU and on GPUs with CUDA. The hope is that this could be a common code base for SciPy and CuPy allowing us to add functions from SciPy which are currently missing in CuPy, and have a common place to make bug fixes and improvements. So far https://github.com/cupy/cupy/pull/8140, and https://github.com/cupy/cupy/pull/8163 have landed in CuPy for Lambert's W function, and Digamma of a complex argument respectively.

There's been a tremendous burst of activity in `special` lately, and within the next week we should have dozens of new functions ready to PR to CuPy. My questions are:

1. Would you be happy to add any special function currently available in SciPy's public API, or would you prefer to select only from a subset of the more important ones?
2. At what rate would you like to see new functions added? Would a single PR adding everything currently translated be OK, should we stick to one function per time, or something in-between like PRs for clusters of related functions?
3. Do you have an approximate timeline for the 14.0 release date, so we can have a sense of the pace we should work at to get as many wanted things in as possible before then?
4. Would you be supportive of replacing existing special function implementations from CuPy with the ones we are now writing for SciPy to make it easier to benefit from improvements and fixes made in SciPy?
5. Do you have any thoughts on the idea of separating out the scalar kernels we are working on in SciPy into a separate library which is included as a submodule of both SciPy and CuPy? For the time-being we plan to carry on by just copying code from SciPy into CuPy, but we are hoping to make this a separate special functions library in the future.

Thank you. I look forward to hearing your thoughts.

cc @izaid @rgommers

takagi:
Thanks for raising this issue before you make PRs to include the new special functions into CuPy.

1. We'd like to have any special function available in SciPy's public API
2. We prefer smaller PRs to make them smoothly reviewed and merged. One or a few functions per time would be nice.
3. We have not set the timing of the 14.0 release yet, but I suppose it will be scheduled roughly between six months and a year later from now.
4. Yes.
5. I agree that having a common submodule in the future would be beneficial.

@kmaehashi, do you have any clarifications or corrections? Also thoughts regarding 5.?

kmaehashi:
Hi @steppi, thank you for raising the discussion! Having and managing almost the same code in both SciPy and CuPy is redundant, so the common library is definitely the best solution.

[...]

PR #9159, opened 2025-06-10
Add xsf as submodule for special function scalar kernels

steppi:
This PR adds the header only library [xsf](https://github.com/scipy/xsf) as a submodule, as discussed in https://github.com/cupy/cupy/issues/8303. xsf was recently added as a submodule of SciPy here, https://github.com/scipy/scipy/pull/22854. If interested, see https://github.com/scipy/xsf/issues/1 some history and extended summary.

This PR only switches to a submodule, but does not add any new special functions, and only uses xsf for the special functions which already relied on the vendored xsf that was inside `cupy/cupy/_core/include/cupy/xsf`. I've marked this as draft because it is blocked until https://github.com/scipy/xsf/pull/39 lands in xsf. That PR makes a couple changes that are needed to avoid errors in CuPy, and also adds a GPU CI job for testing that xsf scalar kernels work when used in CuPy. In its current state, my draft PR to CuPy includes my personal fork of xsf as a submodule, checked out at the branch from https://github.com/scipy/xsf/pull/39. When https://github.com/scipy/xsf/pull/39 lands, I'll cut a new xsf release and change the CuPy PR to include upstream xsf as submodule checked out at the new release.

I'm interested in feedback on whether I've structured things correctly here. Is there anything that I've missed?

Also, I have a question. xsf is a C++17 codebase, but I've found that `cupy._core.create_ufunc` uses the C++11 language standard and seems to have no option to change that. `cupy.ElementwiseKernel` has an `options` kwarg for compiler options to pass to NVRTC that lets one set the language standard, but not `cupy._core.create_ufunc`.  You can see here in `_get_ufunc_kernel` that `_get_simple_elementwise_kernel` is called without allowing `options` to be passed to it. 

https://github.com/cupy/cupy/blob/6e3c9b7b1043e165f4efad022196a078eb11abc4/cupy/_core/_kernel.pyx#L1082-L1084

The default language standard when no options are passed is C++11.

This PR doesn't add anything that will require C++17 features, but for future work, is it fine if I add the ability to pass compiler options through `cupy._core.create_ufunc`? Or perhaps the language standard for `create_ufunc` in particular could be updated to C++17? The primary features xsf uses from C++17 are structured bindings, `if constexpr`, and some type traits stuff. After this PR, I can follow up with one to add the Gaussian hypergeometric function to tackle https://github.com/cupy/cupy/blob/6e3c9b7b1043e165f4efad022196a078eb11abc4/cupy/_core/_kernel.pyx#L1082-L1084, but doing so without having to do complicated workarounds will require allowing C++17.


leofang:
Many thanks @steppi! Very excited! I've been meaning to ask if you guys can help migrate CuPy now that you made it modularized in SciPy! 🙂 

I am in favor of bumping the minimal standard to C++17 everywhere. For the pre-compiled kernels we've moved to C++17 a while ago for exactly the same needs (`if constexpr`):
https://github.com/cupy/cupy/blob/fc5ee663380ff211eadbf7890096f2d46c639922/install/cupy_builder/_compiler.py#L215-L217
However, if it'd make your life easier, feel free to plumb it through the ufunc machinery so that you can pass additional compiler options. I think as of today this PR is small enough that you can piggyback on additional things. If you want to send another PR for that it's also fine. If you just want to raise it to C++17 unconditionally it's also fine with me (@kmaehashi makes the final call, though!).

One thing to note is that CCCL 3.0 will raise the bar to C++17. So this adds another reason for consideration.
https://github.com/NVIDIA/cccl/blob/main/docs/cccl/3.0_migration_guide.rst#platform-support

kmaehashi:
Yeah, I don't see any problem bumping up to C++17 as Leo suggested.

steppi:
https://github.com/scipy/xsf/pull/39 landed in `xsf` and I've updated this PR to use the latest `scipy/xsf` release in the submodule instead of a branch on my fork.

I want to make sure I understand what to do for updating the C++ language standard. It looks like the default gets set here, https://github.com/cupy/cupy/blob/fc5ee663380ff211eadbf7890096f2d46c639922/cupy/_core/core.pyx#L2336, 

so is it fine to update this to `c++17`? There are also some scattered places where a language standard less than `c++17` is passed, see: https://github.com/search?q=repo%3Acupy%2Fcupy%20--std&type=code, and at a glance I'm not sure what to about these.

Or for now, I could just pass the option `c++17` specifically in `_get_ufunc_kernel` here https://github.com/cupy/cupy/blob/fc5ee663380ff211eadbf7890096f2d46c639922/cupy/_core/_kernel.pyx#L1077-L1079.

cupy/_core/include/cupy/xsf
```diff
@@ -0,0 +1 @@
+/home/birbir/cupy/third_party/xsf/include/xsf
```
kmaehashi:
```suggestion
../../../../third_party/xsf/include/xsf
```

leofang:
Since we updated the compiler option, the caches in the CI are invalidated and we hit a timeout. We can update the CI timeout limit and retry. (On my cell, need to find it later.)

leofang:
This is one example PR showing how to adjust the CI timeout limit: https://github.com/cupy/cupy/pull/8564

steppi:
> This is one example PR showing how to adjust the CI timeout limit: #8564

Great! Thanks @leofang!

leofang:
@steppi some of the CI failures are due to timeout, but at least one has scipy.special related failures:
https://ci.preferred.jp/cupy.linux.cuda118/193866/
Could you go through the CI logs?

steppi:
> @steppi some of the CI failures are due to timeout, but at least one has scipy.special related failures: https://ci.preferred.jp/cupy.linux.cuda118/193866/ Could you go through the CI logs?

Sure. I'll investigate.

steppi:
It turns out the problem was missing execution space annotations, which are added in `xsf` with the macro `XSF_HOST_DEVICE`. There were a few functions that were missing them. When using CUDA toolkit 12, which xsf uses in CI, the tests all still pass, but the missing annotaitons caused compiler errors when using CUDA toolkit 11. I was going to set up CUDA 11 jobs in `xsf`, but since CUDA 11 will soon be deprecated, I only tested locally to make these fixes.

Currently the submodule is checked out on a feature branch for this PR, https://github.com/scipy/xsf/pull/45.  If everything goes green, I'll merge that PR, cut a new xsf release and pin the commit for the submodule to that.

asi1024:
Let's check the CI is passing!

leofang:
Looks like the fix works, we just run into timeouts.

steppi:
> Looks like the fix works, we just run into timeouts.

Should I bump the timeout times a further?

leofang:
I think there's a hard limit set by CSP. @kmaehashi any suggestions?

kmaehashi:
I'm attempting to introduce cache in PR CIs to solve this timeout issue (#9234). Sorry for keeping you waited, I'll try to land this fix shortly.

kmaehashi:
/test mini

Reverting the timeout bump to allow Linux per-PR cache to grow.

edit: [The first run of `cupy.linux.cuda112`](https://ci.preferred.jp/cupy.linux.cuda112/195453/) took 6h42m and [the second run](https://ci.preferred.jp/cupy.linux.cuda112/195492/) took only 1h22m. Seems the per-PR cache is working as expected. For other CIs we might need to retry several more times for the per-PR cache to grow.

kmaehashi:
CI passed!

asi1024 (approved):

asi1024:
LGTM!

PR #9176, opened 2025-06-18
Add `cupy.polynomial.polynomial.polyadd`

edicius2002:


asi1024:

asi1024:
It seems to overwrite the input.
```py
>>> x = cupy.array([1, 2, 3], dtype=cupy.float64)
>>> cupy.polynomial.polynomial.polyadd(x, x)
array([2., 4., 6.])
>>> x
array([2., 4., 6.])
```

cupy/polynomial/polyutils.py
```diff
@@ -5,6 +5,29 @@
...
+    ----------
+    c1, c2 : array_like
+        1-d arrays of polynomial coefficients to add.
+
+    Returns
+    -------
+    out : ndarray
+        The sum of the inputs, with trailing zeros removed.
+    """
+    [c1, c2] = as_series([c1, c2])
```
asi1024: How about adding the option trimseq=False to avoid DtoH synchronization?
```suggestion
    [c1, c2] = as_series([c1, c2], trimseq=False)
```

PR #9180, opened 2025-06-19
Add cupy.polynomial.polynomial.polydiv

jphants:


asi1024:

asi1024:
The implementation of this PR includes a Python while loop, so when `c1` is large and `c2` is small, it does not benefit much from the GPU's speed advantage. Would you mind considering implementing a different algorithm that is not NumPy's one for GPUs?

PR #9201, opened 2025-06-24
Support loading NCCL from Pip packages

kmaehashi:
Allow loading NCCL from pip/wheel pacakges. This will give users flexibility of using arbitrary NCCL versions installed in the user's environment with CuPy.

We already supports this for cuTENSOR (#6999/#7025).

### DIFF
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -418,7 +418,7 @@ def _find_compatible_wheel(
     is_compatible = (
         actual[0] == expected[0] and
         (
-            actual[1] >= expected[1] or
+            actual[1] > expected[1] or
             (actual[1] == expected[1] and actual[2] >= expected[2])
         )
     )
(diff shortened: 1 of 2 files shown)


cupy/_environment.py
```diff
@@ -394,62 +397,98 @@ def _parse_version(version: str) -> tuple[int, int, int]:
...
 
-    actual = _parse_version(cutensor_dist.version)
+    actual = _parse_version(dist.version)
     expected = _parse_version(version)
     is_compatible = (
         actual[0] == expected[0] and
-        actual[1] >= expected[1] and
-        actual[2] >= expected[2]
+        (
+            actual[1] >= expected[1] or
```
kmaehashi:
```suggestion
            actual[1] > expected[1] or
```

kmaehashi:
Test passed.

asi1024 (approved):

PR #9200, opened 2025-06-24
Support CUDA 12.9

kmaehashi:
xref #9124

Takign over #9139 (cc/ @jakirkham).

PR #9213, opened 2025-06-26
Add `cupy.win.cuda129` CI

kmaehashi:
Follows-up #9200. Without this configuration CUDA 12.9 + Windows matrix will not be triggered.


kmaehashi:
@asi1024 Let me self-merge as it is trivial and blocking. :bow:

Issue #9120, opened 2025-05-02
Cupy's Delaunay fails while Scipy's doesn't


Rodrigo-Tenorio:
### Description

Hi, 

Let me preface this by thanking the Cupy team for exposing the Delaunay interface on the latest release (👏).

I've encountered a few situations in which Scipy's CPU Delaunay code
gives a sensible result (see attached) while Cupy's Delaunay refuses as it claims
the points to be coplanar. 

While I do agree that there's a set of three points in this plot that are almost
colinear, is this discrepancy expected? 

I've checked the code and it looks like the tolerance for colinearity is [baked in](https://github.com/cupy/cupy/blob/main/cupyx/scipy/spatial/delaunay_2d/_kernels.py#L2478) with no possible user input. Could this be changed?

Thank you very much for your help,

![Image](https://github.com/user-attachments/assets/4f58a0cd-e785-4c49-a2a3-6ecad65a506d)

### To Reproduce

```py
import numpy as np
import cupy as cp

from scipy.spatial import Delaunay as Delaunay
from cupyx.scipy.spatial import Delaunay as GPUDelaunay


points = np.array(
    [
        [-5.88538145, -4.29185831],
        [6.16914963, 4.99256718],
        [-3.00473339, -2.71106304],
        [-0.14600758, 1.82462075],
        [1.15481746, -0.56118624],
        [0.41423983, 1.01326104],
    ]
)

print("Computing Scipy's Delaunay...")
Delaunay(points)
print("Ok")

print("Computing Cupy's Delaunay...")
gpu_points = cp.asarray(points)
print("And now this will fail...")
GPUDelaunay(gpu_points)


```

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
pip install cupy-cuda12x --pre -U -f https://pip.cupy.dev/pre
OS                           : Linux-3.10.0-1160.102.1.el7.x86_64-x86_64-with-glibc2.17
Python Version               : 3.12.9
CuPy Version                 : 14.0.0a1
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.2.4
SciPy Version                : 1.15.2
Cython Build Version         : 3.0.12
Cython Runtime Version       : None
CUDA Root                    : /usr/local/cuda-10.0
nvcc PATH                    : /usr/local/cuda-10.0/bin/nvcc
CUDA Build Version           : 12080
CUDA Driver Version          : 12030
[...]

PR #9219, opened 2025-07-02
Fix coplanarity check in Delaunay triangulation

hovavalon:
This PR fixes the issue described in #9120.
I am having problems testing it on my machine due to pip installation issues so it is not tested yet.
Since this is a very small change, I would appreciate it if someone with a working local pip installation can help my by running the unit tests after this update.


kmaehashi:
@andfoy Could you take a look? :pray:

hovavalon:
@andfoy what do you think?
I have updated the PR so now the tests pass.

Issue #9143, opened 2025-05-28
Cupy's `nan_to_num` can't take an array as its `nan` argument but Numpy can


ikrommyd:
For example:
```
In [1]: import cupy

In [2]: from math import inf, nan

In [3]: rho = cupy.array([39.72022144,  6.45279762, 19.3253414 , 49.08952218, 76.53535633, 41.62411281])

In [4]: z = cupy.array([200.22881345, -123.05876649,  305.25768238,  420.72237365, 217.63714915, 254.213766])

In [5]: cupy.nan_to_num(
   ...:         cupy.arcsinh(z / rho),
   ...:         nan=cupy.nan_to_num((z != 0) * inf, posinf=nan),
   ...:         posinf=inf,
   ...:         neginf=-inf,
   ...:     )
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 cupy.nan_to_num(
      2         cupy.arcsinh(z / rho),
      3         nan=cupy.nan_to_num((z != 0) * inf, posinf=nan),
      4         posinf=inf,
      5         neginf=-inf,
      6     )

File /opt/conda/lib/python3.10/site-packages/cupy/_math/misc.py:422, in nan_to_num(x, copy, nan, posinf, neginf)
    420     out = cupy.empty_like(x) if copy else x
    421 dtype = out.dtype
--> 422 nan = _check_nan_inf(nan, dtype)
    423 posinf = _check_nan_inf(posinf, dtype, False)
    424 neginf = _check_nan_inf(neginf, dtype, True)

File /opt/conda/lib/python3.10/site-packages/cupy/_math/misc.py:402, in _check_nan_inf(x, dtype, neg)
    400 elif x is None and neg is not None:
    401     x = cupy.finfo(dtype).min if neg else cupy.finfo(dtype).max
--> 402 elif cupy.isnan(x):
    403     x = cupy.nan
    404 elif cupy.isinf(x):

File cupy/_core/core.pyx:1293, in cupy._core.core._ndarray_base.__nonzero__()

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
Indeed, even the numpy docs say that `nan` should be an int or a float: https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html. However, numpy is fine with an array :
```
In [6]: import numpy

[...]

PR #9240, opened 2025-07-13
feat: make `cupy.nan_to_num` broadcast `nan`, `posinf`, and `neginf` kwargs

ikrommyd:
Fixes https://github.com/cupy/cupy/issues/9143

Following numpy here: https://github.com/numpy/numpy/blob/3ccbc6c29bb56795d32ee685bccdc1e9ca8ae47f/numpy/lib/_type_check_impl.py#L469-L492,
I don't think that cupy should check for `cupy.isnan(x)` and `cupy.isinf(x)` for the user defined arguments `nan`, `posinf`, and `neginf`. Numpy doesn't care about them, and does the `copyto` anyway. I believe that `cupy.asanyarray(x, dtype)` is enough here.

cc @leofang since we went over this at the SciPy sprints.

### DIFF
--- a/cupy/_math/misc.py
+++ b/cupy/_math/misc.py
@@ -416,10 +416,6 @@ def _check_nan_inf(x, dtype, neg=None):
         x = 0
     elif x is None and neg is not None:
         x = cupy.finfo(dtype).min if neg else cupy.finfo(dtype).max
-    elif cupy.isnan(x):
-        x = cupy.nan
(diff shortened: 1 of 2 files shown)


leofang:
@asi1024 this is ready for review 🙂

asi1024 (approved):

asi1024:
LGTM!

Issue #9249, opened 2025-07-15
Missing `cupy.bool` from Array API


TomAugspurger:
### Description

According to https://docs.cupy.dev/en/stable/reference/array_api_functions.html#module-cupy.array_api, the root `cupy.*` namespace should be compatible with the NumPy 2.x / the Array API standard.

Reading https://data-apis.org/array-api/latest/API_specification/data_types.html, I *think* that the namespace should have a `bool` data type

> A conforming implementation of the array API standard must provide and support the following data types (“dtypes”) in its array object, and as data type objects in its main namespace under the specified names:

But cupy is still using `cupy.bool_` (which NumPy now aliases to `bool`, which is now a NumPy dtype rather than Python's `bool`; it's confusing 😄 ):

### To Reproduce

```python
In [1]: import cupy

In [2]: cupy.bool
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 cupy.bool

File test/lib/python3.13/site-packages/cupy/__init__.py:1087, in __getattr__(name)
   1084 if name in _deprecated_apis:
   1085     return getattr(_numpy, name)
-> 1087 raise AttributeError(f"module 'cupy' has no attribute {name!r}")

AttributeError: module 'cupy' has no attribute 'bool'
```

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
OS                           : Linux-5.4.0-182-generic-x86_64-with-glibc2.31
Python Version               : 3.13.3
CuPy Version                 : 13.5.1
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.3.1
SciPy Version                : None
Cython Build Version         : 3.0.12
Cython Runtime Version       : None
CUDA Root                    : /usr/local/cuda
nvcc PATH                    : /usr/local/cuda/bin/nvcc
CUDA Build Version           : 12090
CUDA Driver Version          : 12020
CUDA Runtime Version         : 12090 (linked to CuPy) / 12040 (locally installed)
CUDA Extra Include Dirs      : []
cuBLAS Version               : 120405
cuFFT Version                : 11201
cuRAND Version               : 10305
cuSOLVER Version             : (11, 6, 1)
cuSPARSE Version             : 12301
NVRTC Version                : (12, 4)
Thrust Version               : 200800
CUB Build Version            : 200800
Jitify Build Version         : <unknown>
cuDNN Build Version          : None
cuDNN Version                : None
[...]

PR #9253, opened 2025-07-18
Add `cupy.bool`, `cupy.long` and `cupy.ulong`

asi1024:
Close https://github.com/cupy/cupy/issues/9249.

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -1065,6 +1065,17 @@ def issctype(rep):
         return False
 
 
+# np 2.0: XXX shims for things newly added in np 2.0
+if _numpy.__version__ < "2":
+    from numpy import bool_ as bool  # NOQA
+    from numpy import int_ as long  # NOQA
+    from numpy import uint as ulong  # NOQA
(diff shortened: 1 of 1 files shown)


kmaehashi:
I think this fails with NumPy v1. Soft import (as done in the last part of `__init__.py`) is needed.

cupy/__init__.py
```diff
@@ -1065,6 +1065,12 @@ def issctype(rep):
         return False
 
 
+if _numpy.__version__ >= "2":
+    from numpy import bool  # NOQA
+    from numpy import long  # NOQA
+    from numpy import ulong  # NOQA
```
kmaehashi: Our policy for v14 is to provide shims to keep consistent API for both NumPy v1/v2 environment. Could you add these for NumPy v1 case, and also add XXX comment as seen below?

```py
from numpy import bool_ as bool
from numpy import int_ as long
from numpy import uint as ulong
```

TomAugspurger:
Thanks!

PR #9266, opened 2025-07-29
Speed up dlpack export and ignore stream for CPU requests

seberg:
The stream ignoring is just because not ignoring it came from me forgetting my own argument:  The stream should be tied to the requested device and the requested device has no stream defined. (Because of that, the old way was safe but I doubt there is much need for it, so I think we should simplify it and not give others the wrong idea.)

Otherwise some micro optimizations for certain degrees of micro.  Not all may be wanted, so happy to revert some if that feels better:
* Coalescing allocation and using PyMem* probably doesn't do much but also doesn't make things worse to read to me.
* The new `pointerGetMemoryType` is likely the most unclear.  Avoids an object creation and attribute lookup (probably around 40ns).
* I suspect that the thread-local to C is about 20ns each (if hit) maybe not worth it, but I thought it likely gets quite often.
* The type-number list avoids one attribute lookup and the conversion, probably around 30ns, so simple structure but maybe rather verbose.

The biggest single thing is just the loop variable `n` I think, which Cython didn't auto-optimize (probably because of the second loop using the same variable?).

All of this got me from around 900ns to around 350ns, almost 100ns of that may just be the `cudaPointerGetAttributes` so I suspect there isn't too much left to achieve here.
(The cuda stream/device handling just adds some complexity in the end.)

Just a note, though.  When streams are used, the stream order logic will add almost 4us for me, meaning that all of this is unimportant...


cupy/_core/core.pyx
```diff
@@ -278,16 +278,18 @@ cdef class _ndarray_base:
...
-            'descr': self.dtype.descr,
+            'typestr': str,
+            'descr': descr,
         }
         cdef int ver = _util.CUDA_ARRAY_INTERFACE_EXPORT_VERSION
         cdef intptr_t stream_ptr
 
         if ver == 3:
-            stream_ptr = stream_module.get_current_stream_ptr()
+            stream_ptr = stream_module.get_current_stream().ptr
```
seberg: Not sure this is good:  It seems faster(?!) and is what dlpack uses so I kept it, but I am not sure about the distinction at all here.

EDIT: If this is technically identical, I guess it may make sense to change what `get_current_stream_ptr()` does.

cupy/_core/_dtype.pyx
```diff
@@ -151,3 +151,29 @@ cdef dict dtype_format = {
...
+    if res is not None:
+        return res
+
+    res = (dtype.str, dtype.descr)
+    if dtype.isbuiltin:
+        # `isbuiltin` is a bit weird, but if not zero, we should be able to
+        # cache (In practice, it should give what we want here)
+        _num_to_str_and_descr[dtype.num] = res
+
+    return res
```
seberg, line 368: Yeah, this is a weird function, but about half the time in `__cuda_array_interface__` was just this, so felt it _may_ be worthwhile.

In the end, CAI is a bit slower than dlpack, building the tuples and dicts just adds up.  (Many could be inlined here to avoid a decent amount of overhead due to access via Python attribute, but OK.)

cupy/_core/core.pyx
asi1024: Hmm... It looks identical.
https://github.com/cupy/cupy/blob/397fd017e235d0db153655a46ea0fd2f28f8e445/cupy/cuda/stream.pyx#L63-L67
https://github.com/cupy/cupy/blob/397fd017e235d0db153655a46ea0fd2f28f8e445/cupy_backends/cuda/stream.pyx#L34-L43

seberg: Sorry, I guess I'll revert it.  Maybe this just includes the `curr_stream == 0` check already.  Or I just timed wrong.

(I think I looked at `_stream_module.get_current_stream_ptr()` maybe, or just slipped up.)

seberg:
Pushed small fixup (nothing serious but slight correctness) and some new tests (mostly vibed, I admit) -- mainly wanted an agent to review so that's that.

/test mini

cupy_backends/cuda/api/runtime.pxd
```diff
@@ -250,6 +250,7 @@ cpdef memPrefetchAsync(intptr_t devPtr, size_t count, int dstDevice,
                        intptr_t stream)
 cpdef memAdvise(intptr_t devPtr, size_t count, int advice, int device)
 cpdef PointerAttributes pointerGetAttributes(intptr_t ptr)
+cpdef int pointerGetMemoryType(intptr_t ptr) except -1
```
leofang, line 254: If this is internal-use only, let's make it cdef instead of cpdef

seberg:
~/test mini~

Accidentally pushed from wrong computer without pulling and lost commits, will fix a bit later :/

asi1024 (approved):

seberg:
Thanks for having a look @asi1024!

PR #9269, opened 2025-07-30
[v13] Missing backport of special function tests

asi1024:
Some test skips are missing in #9246.


asi1024:
Special function tests of head CI have passed!

Issue #9074, opened 2025-03-31
`( __CUDACC_VER_MAJOR__ >= 11 ) && ( __CUDACC_VER_MINOR__ >= 2 )`


SonOfLilit:
I'm not familiar with the codebase, but I've noticed this pattern of `( __CUDACC_VER_MAJOR__ >= 11 ) && ( __CUDACC_VER_MINOR__ >= 2 )` in a few places, and I suspect it might be a mistyping of `( __CUDACC_VER_MAJOR__ == 11 ) || ( __CUDACC_VER_MAJOR__ == 11 &&  __CUDACC_VER_MINOR__ >= 2 )`?

https://github.com/cupy/cupy/blob/53127099fe5c441fec97567f57974504342f304d/cupyx/scipy/signal/_upfirdn.py#L62
https://github.com/cupy/cupy/blob/53127099fe5c441fec97567f57974504342f304d/cupyx/scipy/signal/_upfirdn.py#L175
https://github.com/cupy/cupy/blob/53127099fe5c441fec97567f57974504342f304d/cupyx/scipy/signal/_upfirdn.py#L182
https://github.com/cupy/cupy/blob/53127099fe5c441fec97567f57974504342f304d/cupyx/scipy/signal/_upfirdn.py#L188
https://github.com/cupy/cupy/blob/53127099fe5c441fec97567f57974504342f304d/cupy/cuda/cupy_cub.h#L49

kmaehashi:
Thanks for reporting @SonOfLilit! We no longer support CUDA 11.1 or earlier we can just remove these `#if`s.

DanielFrantes:
> Thanks for reporting @SonOfLilit! We no longer support CUDA 11.1 or earlier so we can just remove these `#if`s.

Hi canI do it?

Ellecee:
I would like to help with removing those by opening a PR if no one else is working on this?

Ellecee:
> Thanks for reporting [@SonOfLilit](https://github.com/SonOfLilit)! We no longer support CUDA 11.1 or earlier so we can just remove these `#if`s.

I would like to help with removing those by opening a PR if no one is working on it?

PR #9281, opened 2025-08-03
CUDA 11.1 or earlier is no longer supported

Ellecee:
This is the fix for Issue [#9074](https://github.com/cupy/cupy/issues/9074).

Since cupy no longer supports CUDA 11.1 or earlier, the #if checks from [cupy/cupyx/scipy/signal/_upfirdn.py](https://github.com/cupy/cupy/blob/53127099fe5c441fec97567f57974504342f304d/cupyx/scipy/signal/_upfirdn.py#L62) and [cupy/cupyx/scipy/signal/_upfirdn.py](https://github.com/cupy/cupy/blob/53127099fe5c441fec97567f57974504342f304d/cupyx/scipy/signal/_upfirdn.py#L175) have been removed.

### DIFF
--- a/cupy/cuda/cupy_cub.h
+++ b/cupy/cuda/cupy_cub.h
@@ -46,9 +46,7 @@ __device__ long long atomicAdd(long long *address, long long val) {
 }
 #endif // __CUDA_ARCH__
 
-#if (defined(_MSC_VER) && (__CUDACC_VER_MAJOR__ == 11 && __CUDACC_VER_MINOR__ == 2))
-  #define __builtin_unreachable() __assume(false)
-#endif
+#define __builtin_unreachable() __assume(false)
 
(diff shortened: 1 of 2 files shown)


cupy/cuda/cupy_cub.h
```diff
@@ -46,10 +46,6 @@ __device__ long long atomicAdd(long long *address, long long val) {
 }
 #endif // __CUDA_ARCH__
 
-#if (defined(_MSC_VER) && (__CUDACC_VER_MAJOR__ == 11 && __CUDACC_VER_MINOR__ == 2))
-  #define __builtin_unreachable() __assume(false)
-#endif
-
```
kmaehashi: This is required as we still support CUDA 11.2.

Issue #4421, opened 2020-12-07
Separate compilation not working on Windows


kmaehashi:
The following code runs fine on Linux but fails on Windows:

```py
import cupy
kernel = cupy.RawKernel('extern "C" __global__ void test() {}', 'test', backend='nvcc', enable_cooperative_groups=True)
kernel.compile()
```

Error raised in Windows:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy\core\raw.pyx", line 282, in cupy.core.raw.RawKernel.compile
  File "cupy\core\raw.pyx", line 110, in cupy.core.raw.RawKernel._kernel
  File "cupy\cuda\function.pyx", line 234, in cupy.cuda.function.Module.get_function
  File "cupy\cuda\function.pyx", line 175, in cupy.cuda.function.Function.__init__
  File "cupy_backends\cuda\api\driver.pyx", line 262, in cupy_backends.cuda.api.driver.moduleGetFunction
  File "cupy_backends\cuda\api\driver.pyx", line 124, in cupy_backends.cuda.api.driver.check_status
cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_NOT_FOUND: named symbol not found
```

Internally, the above code does the following nvcc separate compilation:

```
nvcc -gencode=arch=compute_70,code=sm_70 -ftz=true -rdc=true -Xcompiler -fPIC -shared -o test_rdc.o test.cu
nvcc -gencode=arch=compute_70,code=sm_70 --cubin -ftz=true -rdc=true -Xcompiler -fPIC -shared test_rdc.o --device-link -o test_rdc.cubin
```

It generates a correct cubin on Linux:

```
% cuobjdump -symbols test_rdc.cubin

symbols:
STT_FUNC         STB_GLOBAL STO_ENTRY      test
```

But on Windows it generates an empty (?) cubin, which is causing `named symbol not found` error:

```
C:\Users\chainer\Documents>cuobjdump -all -symbols test_rdc.cubin

symbols:
```

It seems we need some other commands to do separate compilation on Windows.

cc/ @leofang @anaruse

anaruse:
This might be a dumb question, but why do we need a separate compilation?

leofang:
Two examples are cooperative group or dynamic parallelism, where a device linking (to the static device runtime library) must happen, and so a separate compilation is required.

leofang:
[...]

PR #9289, opened 2025-08-08
Add CUDA 13 support

emcastillo:
API changes: Fix prefetch/memAdvise and some device properties.

These changes are 99,9% not mine. I am just upstreaming what folks (@robertmaynard) at Nvidia did.

- There is no CUDA 13 package for cuTensor for the install_library script.
- CUDA 13 is only supported by cudnn 9.12.0.46, and this version doesn't seem to be compatible with CuPy.
- cuTENSOR cuda13 library is pending release
- CUDA 13 docker image is pending release

xref #9286

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -246,7 +246,8 @@ cdef class ManagedMemory(BaseMemory):
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
         flags = 0
-        runtime.memPrefetchAsync(self.ptr, self.size, device_id, flags, stream.ptr)
+        runtime.memPrefetchAsync(
+            self.ptr, self.size, device_id, flags, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """(experimental) Advise about the usage of this memory.
@@ -324,7 +325,8 @@ cdef class SystemMemory(BaseMemory):
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
         flags = 0
-        runtime.memPrefetchAsync(self.ptr, self.size, device_id, flags, stream.ptr)
+        runtime.memPrefetchAsync(
+            self.ptr, self.size, device_id, flags, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """Advise about the usage of this memory.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -236,7 +236,7 @@ cdef class ManagedMemory(BaseMemory):
         if size > 0:
             self.ptr = runtime.mallocManaged(size)
 
-    def prefetch(self, stream, *, int device_id=runtime.cudaInvalidDeviceId):
+    def prefetch(self, stream, *, int device_id=runtime.cudaInvalidDeviceId, int flags=0):
         """(experimental) Prefetch memory.
 
         Args:
@@ -245,7 +245,7 @@ cdef class ManagedMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
-        runtime.memPrefetchAsync(self.ptr, self.size, device_id, stream.ptr)
+        runtime.memPrefetchAsync(self.ptr, self.size, device_id, flags, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """(experimental) Advise about the usage of this memory.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -245,8 +245,7 @@ cdef class ManagedMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
-        runtime.memPrefetchAsync(
-            self.ptr, self.size, device_id, stream.ptr)
+        runtime.memPrefetchAsync(self.ptr, self.size, device_id, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """(experimental) Advise about the usage of this memory.
@@ -323,8 +322,7 @@ cdef class SystemMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
-        runtime.memPrefetchAsync(
-            self.ptr, self.size, device_id, stream.ptr)
+        runtime.memPrefetchAsync(self.ptr, self.size, device_id, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """Advise about the usage of this memory.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -236,7 +236,7 @@ cdef class ManagedMemory(BaseMemory):
         if size > 0:
             self.ptr = runtime.mallocManaged(size)
 
-    def prefetch(self, stream, *, int device_id=runtime.cudaInvalidDeviceId, int flags=0):
+    def prefetch(self, stream, *, int device_id=runtime.cudaInvalidDeviceId):
         """(experimental) Prefetch memory.
 
         Args:
@@ -245,6 +245,7 @@ cdef class ManagedMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
+        flags = 0
         runtime.memPrefetchAsync(self.ptr, self.size, device_id, flags, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
@@ -322,7 +323,8 @@ cdef class SystemMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
-        runtime.memPrefetchAsync(self.ptr, self.size, device_id, stream.ptr)
+        flags = 0
+        runtime.memPrefetchAsync(self.ptr, self.size, device_id, flags, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """Advise about the usage of this memory.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -245,9 +245,8 @@ cdef class ManagedMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
-        flags = 0
         runtime.memPrefetchAsync(
-            self.ptr, self.size, device_id, flags, stream.ptr)
+            self.ptr, self.size, device_id, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """(experimental) Advise about the usage of this memory.
@@ -324,9 +323,8 @@ cdef class SystemMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
-        flags = 0
         runtime.memPrefetchAsync(
-            self.ptr, self.size, device_id, flags, stream.ptr)
+            self.ptr, self.size, device_id, stream.ptr)
 
     def advise(self, int advise, device.Device dev):
         """Advise about the usage of this memory.
--- a/cupy_backends/cuda/api/runtime.pxd
+++ b/cupy_backends/cuda/api/runtime.pxd
@@ -247,7 +247,7 @@ cpdef memcpy3DAsync(intptr_t Memcpy3DParmsPtr, intptr_t stream)
 cpdef memset(intptr_t ptr, int value, size_t size)
 cpdef memsetAsync(intptr_t ptr, int value, size_t size, intptr_t stream)
 cpdef memPrefetchAsync(intptr_t devPtr, size_t count, int dstDevice,
-                       int flags, intptr_t stream)
+                       intptr_t stream)
 cpdef memAdvise(intptr_t devPtr, size_t count, int advice, int device)
 cpdef PointerAttributes pointerGetAttributes(intptr_t ptr)
 cpdef intptr_t deviceGetDefaultMemPool(int) except? 0
--- a/cupy_backends/cuda/api/runtime.pxd
+++ b/cupy_backends/cuda/api/runtime.pxd
@@ -247,7 +247,7 @@ cpdef memcpy3DAsync(intptr_t Memcpy3DParmsPtr, intptr_t stream)
 cpdef memset(intptr_t ptr, int value, size_t size)
 cpdef memsetAsync(intptr_t ptr, int value, size_t size, intptr_t stream)
 cpdef memPrefetchAsync(intptr_t devPtr, size_t count, int dstDevice,
-                       intptr_t stream)
+                       int flags, intptr_t stream)
 cpdef memAdvise(intptr_t devPtr, size_t count, int advice, int device)
 cpdef PointerAttributes pointerGetAttributes(intptr_t ptr)
 cpdef intptr_t deviceGetDefaultMemPool(int) except? 0
--- a/cupy_backends/cuda/api/runtime.pyx
+++ b/cupy_backends/cuda/api/runtime.pyx
@@ -743,8 +743,8 @@ cpdef memAdvise(intptr_t devPtr, size_t count, int advice, int device):
         raise RuntimeError('Managed memory requires ROCm 4.3+')
     IF CUPY_CUDA_VERSION < 13000:
         with nogil:
-            status = cudamemadvise(<void*>devptr, count,
-                                   <memoryadvise>advice, device)
+            status = cudaMemAdvise(<void*>devPtr, count,
+                                   <MemoryAdvise>advice, device)
     ELSE:
         cdef _MemLocation loc_c
         c_memset(&loc_c, 0, sizeof(_MemLocation))
--- a/cupy_backends/cuda/api/runtime.pyx
+++ b/cupy_backends/cuda/api/runtime.pyx
@@ -210,12 +210,28 @@ cpdef int deviceGetAttribute(int attrib, int device) except? -1:
 
 cpdef getDeviceProperties(int device):
     cdef DeviceProp props
+    cdef int ret
     cdef int status = cudaGetDeviceProperties(&props, device)
     check_status(status)
 
     cdef dict properties = {'name': b'UNAVAILABLE'}  # for RTD
 
     # Common properties to CUDA 9.0, 9.2, 10.x, 11.x, and HIP
+    IF CUPY_CUDA_VERSION >= 13000:
+        clockRate = deviceGetAttribute(cudaDevAttrClockRate, device)
+        kernelExecTimeoutEnabled = deviceGetAttribute(
+            cudaDevAttrKernelExecTimeout, device)
+        memoryClockRate = deviceGetAttribute(
+            cudaDevAttrMemoryClockRate, device)
+        computeMode = deviceGetAttribute(cudaDevAttrComputeMode, device)
+        cooperativeMultiDeviceLaunch = False
+    ELSE:
+        clockRate = props.clockRate
+        kernelExecTimeoutEnabled = props.kernelExecTimeoutEnabled
+        memoryClockRate = props.memoryClockRate
+        computeMode = props.computeMode
+        cooperativeMultiDeviceLaunch = props.cooperativeMultiDeviceLaunch
+
     IF CUPY_CUDA_VERSION > 0 or CUPY_HIP_VERSION > 0:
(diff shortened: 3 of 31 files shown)


cupy/cuda/memory.pyx
```diff
@@ -245,7 +245,8 @@ cdef class ManagedMemory(BaseMemory):
         """
         if device_id == runtime.cudaInvalidDeviceId:
             device_id = self.device_id
-        runtime.memPrefetchAsync(self.ptr, self.size, device_id, stream.ptr)
+        flags = 0
+        runtime.memPrefetchAsync(self.ptr, self.size, device_id, flags, stream.ptr)
```
agirault: Is this backward compatible with cuda 12? I assume need the if else?

cupy_backends/cuda/api/runtime.pxd
```diff
@@ -247,7 +247,7 @@ cpdef memcpy3DAsync(intptr_t Memcpy3DParmsPtr, intptr_t stream)
 cpdef memset(intptr_t ptr, int value, size_t size)
 cpdef memsetAsync(intptr_t ptr, int value, size_t size, intptr_t stream)
 cpdef memPrefetchAsync(intptr_t devPtr, size_t count, int dstDevice,
-                       intptr_t stream)
+                       int flags, intptr_t stream)
```
agirault: No cuda version check?

cupy_backends/cuda/api/runtime.pyx
```diff
@@ -695,20 +720,40 @@ cpdef memsetAsync(intptr_t ptr, int value, size_t size, intptr_t stream):
     check_status(status)
 
 cpdef memPrefetchAsync(intptr_t devPtr, size_t count, int dstDevice,
-                       intptr_t stream):
+                       int flags, intptr_t stream):
```
agirault: Same question

cupy_backends/cuda/libs/_cnvrtc.pxi
```diff
@@ -131,6 +131,12 @@ cdef SoftLink _get_softlink():
                 libname = 'libnvrtc.so.12'
             else:
                 libname = 'nvrtc64_120_0.dll'
+        elif runtime_version == 13:
+            # CUDA 12.x
```
agirault:
```suggestion
            # CUDA 13.x
```

install/cupy_builder/_compiler.py
```diff
@@ -84,16 +84,17 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
...
-                         ('compute_70', 'sm_70'),
-                         ('compute_75', 'sm_75'),
-                         ('compute_80', 'sm_80'),
+            arch_list = [('compute_80', 'sm_80'),
                          ('compute_86', 'sm_86'),
                          ('compute_89', 'sm_89'),
-                         ('compute_90', 'sm_90'),]
+                         ('compute_90', 'sm_90')]
+            if cuda_version < 13000:
+                arch_list += [('compute_50', 'sm_50'),
```
agirault: Seems quite different to the way it's done below, a bit harder to read too, and seems to miss 87 for both cuda 12 and 13 aarch64. 

I can help with the list if that's helpful.

cupy/cuda/memory.pyx
emcastillo: Yeah, the IF is done inside the runtime.pyx call, it will accept the flag always and dispatch to the correct binding.

https://github.com/cupy/cupy/pull/9289/commits/5c491b23dd4b0f886a3a0fb2ba9d63e18929492c#diff-82cb8875a3644f631939ffd4b7e0c9fe9b596cc8af3437222e5c8e17b77e48faR701

The problem is that the `runtime.memPrefetchAsync` has now a different signature but I don't think we can easily avoid that. @kmaehashi what do you think

cupy_backends/cuda/api/runtime.pxd
emcastillo: We add it unconditionally to the API and then use it or not in the body depending on the version.

install/cupy_builder/_compiler.py
emcastillo: Good catch! let me rethink it

cupy_backends/cuda/libs/cusparse.pyx
```diff
@@ -1381,6 +1381,11 @@ ELIF 12000 <= CUPY_CUDA_VERSION < 13000:
         _libname = 'libcusparse.so.12'
     else:
         _libname = 'cusparse64_12.dll'
+ELIF 13000 <= CUPY_CUDA_VERSION < 14000:
+    if _sys.platform == 'linux':
```
robertmaynard, line 1387: I don't know if we should add a comment that libcusparse SOVERSION are independent of the CUDA Toolkit version and therefore that `12` is correct.

install/cupy_builder/_compiler.py
```diff
@@ -83,6 +83,19 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
...
 
         aarch64 = (platform.machine() == 'aarch64')
+        if cuda_version >= 13000:
+            arch_list = [('compute_80', 'sm_80'),
+                         ('compute_86', 'sm_86'),
+                         ('compute_89', 'sm_89'),
+                         ('compute_90', 'sm_90'),
+                         ('compute_100f', 'sm_100'),
+                         ('compute_120f', 'sm_1020'),
+                         'compute_100']
```
robertmaynard:
```suggestion
            arch_list = [
                         ('compute_75', 'sm_75'),
                         ('compute_80', 'sm_80'),
                         ('compute_86', 'sm_86'),
                         ('compute_89', 'sm_89'),
                         ('compute_90', 'sm_90'),
                         ('compute_100f', 'sm_100'),
                         ('compute_120f', 'sm_1020'),
                         'compute_100']
```

I don't know if we need every supported in the arch_list if so we also need to add `103` and `121` for sure. We should also look into `88` which was also added in CUDA 13

install/cupy_builder/_compiler.py
```diff
@@ -83,6 +83,19 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
...
+            if aarch64:
+                # JetPack 
+                arch_list += [
+                    ('compute_87', 'sm_87'),  # Jetson (Orin)
```
robertmaynard:
```suggestion
                    ('compute_87', 'sm_87'),  # Jetson (Orin)
                   ('compute_110', 'sm_110'),  # Jetson
```

install/cupy_builder/_compiler.py
agirault: Why not `compute_120` @robertmaynard ? Also got a typo on l84, should be `sm102` not `sm_1020`.

> I don't know if we need every supported in the arch_list if so we also need to add 103 and 121 for sure. We should also look into 88 which was also added in CUDA 13


100 will run on 103, 120 and 120f will run on 121, but at the same time, 80 would also run on 86 and 89. It depends on the features needed, I suppose would need to understand if these features are used in cupy? For example, torch never builds 89 generally because it's so specific and 86 is sufficient, and only builds 89 for kernels that need fp8: https://github.com/pytorch/pytorch/issues/152690#issuecomment-2847723785

install/cupy_builder/_compiler.py
agirault:
```suggestion
                    ('compute_87', 'sm_87'),    # Jetson (Orin)
                    ('compute_110', 'sm_110'),  # Jetson (Thor)
```

install/cupy_builder/_compiler.py
```diff
@@ -83,6 +83,21 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
...
 
         aarch64 = (platform.machine() == 'aarch64')
+        if cuda_version >= 13000:
+            arch_list = [('compute_75', 'sm_75'),
+                         ('compute_80', 'sm_80'),
+                         ('compute_86', 'sm_86'),
+                         ('compute_89', 'sm_89'),
+                         ('compute_90', 'sm_90'),
+                         ('compute_100f', 'sm_100'),
+                         ('compute_120f', 'sm_1020'),
```
leofang:
```suggestion
                         ('compute_120f', 'sm_120'),
```

install/cupy_builder/_compiler.py
```diff
@@ -83,6 +83,21 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
...
+                         'compute_100']
```
leofang:
```suggestion
                         'compute_120']
```

cupy/cuda/memory.pyx
kmaehashi: Doing IF in `runtime.pyx` sounds reasonable in this case.

install/cupy_builder/_compiler.py
agirault: This is PTX right? Might as well do `compute_121` actually, that's the latest, doesn't need to be major version only. Right?

agirault: And maybe out of scope, but 121 also latest for 12.9, and 120 latest for 12.8: https://gist.github.com/agirault/d0d66c12c85dbc2662c95569ef039a1d

install/cupy_builder/_compiler.py
agirault: FWIW: https://gist.github.com/agirault/d0d66c12c85dbc2662c95569ef039a1d

install/cupy_builder/_compiler.py
```diff
@@ -83,6 +83,23 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
         #   https://docs.nvidia.com/cuda/cuda-for-tegra-appnote/index.html#deployment-considerations-for-cuda-upgrade-package
 
         aarch64 = (platform.machine() == 'aarch64')
+        if cuda_version >= 13000:
+            arch_list = [('compute_75', 'sm_75'),
+                         ('compute_80', 'sm_80'),
+                         ('compute_86', 'sm_86'),
+                         ('compute_87', 'sm_87'),
```
agirault:
```suggestion
```

This is jetson orin so aarch64 only

install/cupy_builder/_compiler.py
```diff
@@ -83,6 +83,23 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
...
+                         ('compute_89', 'sm_89'),
+                         ('compute_90', 'sm_90'),
+                         ('compute_100f', 'sm_100'),
+                         ('compute_120f', 'sm_120'),
+                         'compute_120']
+            if aarch64:
+                # JetPack
+                arch_list += [
+                    ('compute_87', 'sm_87'),  # Jetson (Orin)
+                    ('compute_110', 'sm_110'),  # Jetson
```
agirault:
```suggestion
                    ('compute_87', 'sm_87'),    # Jetson (Orin)
                    ('compute_110', 'sm_110'),  # Jetson (Thor)
```

cupy_backends/cuda/api/runtime.pxd
leofang: But wouldn't this be a breaking change? Since we'll backport this PR to v13 (and cut v13.6.0), and `flags` must be zero for now, how about we keep the signature unchanged until CuPy v14.0.0 is ready to ship?

emcastillo: yeah, like @kmaehashi said, lets do the version check in `runtime.pyx`

leofang: yeah, runtime.pxd should not be changed

install/cupy_builder/_compiler.py
```diff
@@ -83,6 +83,23 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
...
+                ]
         if cuda_version >= 12000:
```
kmaehashi:
```suggestion
        elif cuda_version >= 12000:
```

install/cupy_builder/_compiler.py
```diff
@@ -93,7 +110,7 @@ def _nvcc_gencode_options(cuda_version: int) -> list[str]:
                          ('compute_80', 'sm_80'),
                          ('compute_86', 'sm_86'),
                          ('compute_89', 'sm_89'),
-                         ('compute_90', 'sm_90'),]
+                         ('compute_90', 'sm_90')]
```
kmaehashi:
```suggestion
                         ('compute_90', 'sm_90'),]
```

cupy/cuda/memory.pyx
emcastillo: Fixed now :) we don't pass flags and inside runtime.pyx, the `IF` sets it to 0.

cupy_backends/cuda/api/runtime.pxd
emcastillo: ok! I think i fixed it :) PTAL

kmaehashi:
Testing new Windows CI image here:

* ~b6cad5148b8a4ec4a5e47ca678d71a4fa25a77ae: https://ci.preferred.jp/cupy.win.cuda130/197956/~
* ~18e2d5716df48df52d0bd4a5aae5303083e4b5d5: https://ci.preferred.jp/cupy.win.cuda130/197957/~
* ~a13f4018fd0ef9c0eea0cef47c809d393994aacb: https://ci.preferred.jp/cupy.win.cuda130/197958/~

cupy_backends/cuda/api/runtime.pxd
leofang: LGTM!

cupy/cuda/memory.pyx
leofang: LGTM!

cupy_backends/cuda/libs/cusparse.pyx
agirault: Fully should.

kmaehashi:
Pushed some fixes.

### CUDA 13 CI status

c63fb2d0411cefd232f00c46c1fd2ac57f426239:
* https://ci.preferred.jp/cupy.linux.cuda130/197965/ ❌
* https://ci.preferred.jp/cupy.linux.cuda130.multi/197966/ ❌
* https://ci.preferred.jp/cupy.win.cuda130/197964/ ❌

6e7dabe4e6374784f855bd97a572682bf867e4a8:
* https://ci.preferred.jp/cupy.linux.cuda130/197967/ ❌
* https://ci.preferred.jp/cupy.linux.cuda130.multi/197968/ ✅ 

f50a36dc40efcf4764f91ca0c48aacd0e5deb8bf:
* https://ci.preferred.jp/cupy.win.cuda130/197989/ ✅ 

8d919203d4865bafb13964c3e263d598ecb080f1:
* https://ci.preferred.jp/cupy.linux.cuda130/197990/ ❌

64b933baadce8e6a12cbe78e56de215b3f62f3c0:
* https://ci.preferred.jp/cupy.linux.cuda130/198024/
* https://ci.preferred.jp/cupy.linux.cuda130.multi/198026/
* https://ci.preferred.jp/cupy.win.cuda130/198025/

kmaehashi (changes requested):

install/cupy_builder/_compiler.py
kmaehashi: I guess this is aarch64 specific?
```suggestion
```

leofang:
FYI it seems the 13.0 docker images are up!

kmaehashi:
> FYI it seems the 13.0 docker images are up!

I guess our CI is one of the very first user 😄

kmaehashi:
I discussed with @leofang offline and found that we need to adjust CUDA header discovery logic to support the change in PyPI CUDA packages. I'm looking into it.
https://developer.nvidia.com/blog/whats-new-and-important-in-cuda-toolkit-13-0/#wheel_package_changes_to_cuda_130

edit: done in 3ad708a428d14284d4d0be6d953e0443f6984924

cupy_backends/cuda/api/runtime.pyx
```diff
@@ -210,12 +210,28 @@ cpdef int deviceGetAttribute(int attrib, int device) except? -1:
 
 cpdef getDeviceProperties(int device):
     cdef DeviceProp props
+    cdef int ret
```
kmaehashi: Seems unused?

kmaehashi:
> [6e7dabe](https://github.com/cupy/cupy/commit/6e7dabe4e6374784f855bd97a572682bf867e4a8):
> 
> * https://ci.preferred.jp/cupy.linux.cuda130/197967/

Seems number of tests are failing only on Linux. (full log is here: https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/197967/log.txt)

emcastillo:
Tests are failing because the fft static library doesn't support sm_75, I will push a fix

```
$ cuobjdump --list-elf /usr/local/cuda-13/lib64/libcufft_static.a

ELF file    1: libcufft_static.1.sm_80.cubin
ELF file    2: libcufft_static.2.sm_90.cubin
ELF file    3: libcufft_static.3.sm_100.cubin
ELF file    4: libcufft_static.4.sm_110.cubin
ELF file    5: libcufft_static.5.sm_103.cubin
ELF file    6: libcufft_static.6.sm_120.cubin
```

Seems the flexCI instances has CC 7.5 GPUs so we need to skip these tests :D

emcastillo:
```
FAILED cupy_tests/core_tests/test_raw.py::TestRaw_param_3_{backend='nvcc', in_memory=False}::test_dynamical_parallelism
FAILED cupy_tests/core_tests/test_raw.py::TestRawPicklable_param_0_{compile=False, raw='ker'}::test_raw_picklable
FAILED cupy_tests/core_tests/test_raw.py::TestRawPicklable_param_1_{compile=False, raw='mod'}::test_raw_picklable
FAILED cupy_tests/core_tests/test_raw.py::TestRawPicklable_param_2_{compile=False, raw='mod_ker'}::test_raw_picklable
FAILED cupy_tests/core_tests/test_raw.py::TestRawPicklable_param_3_{compile=True, raw='ker'}::test_raw_picklable
FAILED cupy_tests/core_tests/test_raw.py::TestRawPicklable_param_4_{compile=True, raw='mod'}::test_raw_picklable
FAILED cupy_tests/core_tests/test_raw.py::TestRawPicklable_param_5_{compile=True, raw='mod_ker'}::test_raw_picklable
FAILED cupy_tests/cuda_tests/test_compiler.py::TestNvrtcArch::test_compile_cuda9
```

`test_compile_cuda9` seems to be already fixed, the `RawKernel` tests work fine in my local machine with cc 7.5, may be related to flexCI settings?

kmaehashi:
I've re-kicked CUDA 13 CI with the latest commit. Let's see how it goes...
https://github.com/cupy/cupy/pull/9289#issuecomment-3170383700

kmaehashi:
https://ci.preferred.jp/cupy.linux.cuda130/197990/ is the CI results for the commit 8d919203d4865bafb13964c3e263d598ecb080f1.

In addition to `Test1dCallbacks`, `TestNdCallbacks` needs to be skipped on `sm_75` as well. I've pushed a fix for that.

For `test_raw.py` test failures, I could reproduce locally using `NVCC="ccache nvcc"`. The problem is described in https://github.com/cupy/cupy/issues/4421#issuecomment-749491667 but I'm not sure why this started to happen in CUDA 13 :thinking:
Let's kick other CIs for older CUDA versions to isolate the problem.

/test mini

leofang:
> For test_raw.py test failures, I could reproduce locally using NVCC="ccache nvcc". The problem is described in https://github.com/cupy/cupy/issues/4421#issuecomment-749491667 but I'm not sure why this started to happen in CUDA 13 🤔

Not sure what happened... Looks like an I/O error, could it be out of space?
```
            if code_type == 'ptx':
>               with open(result_path, 'rb') as ptx_file:
                     ^^^^^^^^^^^^^^^^^^^^^^^
E               FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpzktuvmkn/preprocess.ptx'
```

kmaehashi:
No, this is a longstanding bug in ccache that cannot handle `--ptx -dc preprocess.cu` or `--ptx --device-c preprocess.cu` option. For some reason ccache automatically appends `-o preprocess.o` when proxying to `nvcc`, where it should not (or should append `-o preprocess.ptx`). I'm still unsure why this started happen here though... Anyway I've added a workaround.

/test mini

kmaehashi:
Kicked CUDA 13 CIs again.

https://ci.preferred.jp/cupy.linux.cuda130/198024/
https://ci.preferred.jp/cupy.linux.cuda130.multi/198026/
https://ci.preferred.jp/cupy.win.cuda130/198025/

-> all passed!

PR #9301, opened 2025-08-11
Support building NVTX on Windows without Nsight Systems

kmaehashi:
This PR allows building NVTX support in CuPy under environment that does not have Nsight Systems installed.
NVTX3 is placed under `$CUDA_PATH/include` so we can just use it. (I'm not sure why I overlooked this in https://github.com/cupy/cupy/pull/7304#issuecomment-1383431537 ...)

This is a blocker for the next release. Our new Windows CI machine image does not have Nsight installed, thus the wheel build CI is failing (https://ci.preferred.jp/cupy-release-tools.win/198100/).


kmaehashi:
Test failures are unrelated.

PR #9309, opened 2025-08-12
Fix CI failure in CUDA 12.4

kmaehashi:
Fixes regression in #9295. It seems `cuda_fp16.h` bundled with CUDA 12.4.0 has a problem and causing the following error:

https://ci.preferred.jp/cupy.linux.cuda124/198208/

```
ptxas fatal   : Unresolved extern function '_ZN6__halfC1Ef'
```

I locally confirmed this is fixed in CUDA 12.4.1.


leofang (approved):

kmaehashi:
Confirmed CI issue resolved. Error shown is timeout and is irrelevant.

Issue #9324, opened 2025-08-22
Stop importing cupyx in cupy


vyasr:
### Description

Currently importing cupy automatically imports cupyx. Given that cupyx is extensions built on top of cupy, this requirement seems backwards. It would be nice for cupy to remove that requirement and leave importing cupyx for users who want it. This approach has a few advantages:
- It removes deprecated code paths. It has been a couple of releases since [a number of bits of cupy functionality were moved to cupyx and deprecated](https://github.com/cupy/cupy/commit/cd14be90db4cf2b67003a7074bbc9a055a0eec3a). Users have known for a while now that cupyx was the place to go for extended functionality.
- It better parallels the separation between corresponding host libraries. numpy does not import scipy. While cupyx does contain more than just scipy functionality, to a reasonable first approximation cupy is "numpy on the GPU" and cupyx is "scipy+ on the GPU".
- cupy is on the slower side of libraries to import, taking about half a second on my machine where (for comparison) numpy takes about 0.1 s. From a look at `python -Ximporttime cupy`, it seems like cupyx is between 70-90% of that import time. Those numbers aren't super stable, and from what I can tell the absolute minimum time to import any package (including a built-in) is around 0.05s, but it does indicate a lot of room for improvement if cupyx was removed.

As a starting point, removing some of the deprecated cupy modules would get us part of the way there and seems long overdue. I am happy to open a PR for just that part if that is desired and we can continue discussing other current uses of cupyx inside cupy (I see a few, but not too many).

leofang:
Feel free to open a PR, Vyas! It should be fine to remove those for which we've raised warnings for years, especially it should be safe in v14. We do need to mention them in the migration guide https://docs.cupy.dev/en/latest/upgrade.html.

For completeness, could you share how to reproduce your observation on import time?

vyasr:
Sure! A simple import on the CLI shows this:
```
❯ time python -c "import cupy"
python -c "import cupy"  0.58s user 0.09s system 75% cpu 0.888 total
❯ time python -c "import numpy"
python -c "import numpy"  0.12s user 0.01s system 99% cpu 0.125 total
```

To get more granular information, I started with `python -Ximporttime -c "import cupy"` and inspected the outputs. Some of the most relevant lines there are:
```
import time:       151 |    1550175 |       cupyx.scipy.fft
import time:       122 |      85262 |     cupyx.linalg
import time:    148298 |    1993306 | cupy
```
[...]

PR #9325, opened 2025-08-23
Accelerate duplicate installation check

vyasr:
By only constructing the known distributions to check we can speed up import times pretty substantially. This approach has the benefit of being linear in the number of known cupy distributions, which is ~fixed and small, instead the number of distributions in an environment, which is arbitrarily large. Note that importlib.util.find_spec is even faster and therefore tempting to reach for, but not suitable for this case where we are explicitly searching for multiple distribution packages corresponding to the same import package. 

I verified that this approach works in a conda environment where I install cupy and then I pip install various other distributions of cupy like cupy-cuda12x and cupy-cuda11x. The expected warning is produced with the full list of incorrect packages.

Here is the import times of cupy before this change on my machine in a larger environment (`conda list | wc -l` outputs 680, so if you assume that half of those are non-Python binaries then we're looking at ~350 Python packages):
```bash
❯ time python -c "import cupy"
python -c "import cupy"  0.58s user 0.09s system 75% cpu 0.888 total
```
and after:
```bash
❯ time python -c "import cupy"
python -c "import cupy"  0.43s user 0.07s system 99% cpu 0.501 total
```


vyasr:
OK here's a bit more rigorous information. I set up a development environment with
```
mamba install -y "setuptools>=77" wheel "Cython>=3,<3.2" "fastrlock>=0.5" cuda=12 pyinstrument sccache numpy
export NVCC="sccache nvcc"
pip install --no-build-isolation -e . --no-deps
```

With the latest main branch, I get
```
❯ pyinstrument -c "import cupy"

  _     ._   __/__   _ _  _  _ _/_   Recorded: 16:16:18  Samples:  771
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.855     CPU time: 10.419
/   _/                      v5.1.1

Program: -c

0.361 <module>  cupy/__init__.py:1                                                                                                                                                                                                                                     [14/1905]
├─ 0.150 <module>  numpy/__init__.py:1
│     [37 frames hidden]  numpy, typing
├─ 0.044 <module>  cupy/testing/__init__.py:1
│  ├─ 0.026 <module>  cupy/testing/_helper.py:1
│  │  └─ 0.025 <module>  unittest/mock.py:1
│  │        [3 frames hidden]  asyncio, ssl
│  └─ 0.015 <module>  cupy/testing/_array.py:1
│     └─ 0.015 <module>  numpy/testing/__init__.py:1
│           [3 frames hidden]  unittest, numpy, concurrent
├─ 0.044 <module>  cupy/_core/__init__.py:1
│  ├─ 0.023 <module>  cupy/cuda/__init__.py:1
│  │  └─ 0.016 <module>  cupy/cuda/compiler.py:1
│  └─ 0.010 [self]  cupy/_core/__init__.py
├─ 0.026 _detect_duplicate_installation  cupy/_environment.py:566
│  └─ 0.024 PathDistribution.metadata  importlib/metadata/__init__.py:448
│        [9 frames hidden]  email, importlib
...
```

So in this environment I don't see environment detection as having a huge impact. Next, I do a `mamba install -y rapids -c rapidsai`. I work on RAPIDS, so I chose this as a representative "complex" environment with a lot of packages. In that environment, I get
```
0.852 <module>  cupy/__init__.py:1
├─ 0.241 _detect_duplicate_installation  cupy/_environment.py:556
│     [26 frames hidden]  importlib, email, pathlib
├─ 0.219 <module>  cupyx/__init__.py:1
│     [35 frames hidden]  cupyx, scipy, numpy, re
├─ 0.152 <module>  numpy/__init__.py:1
│     [24 frames hidden]  numpy, typing
├─ 0.130 <module>  cupy/lib/_routines_poly.py:1
│     [23 frames hidden]  cupyx, scipy, importlib
├─ 0.038 <module>  cupy/_core/__init__.py:1
│     [3 frames hidden]  cupy
├─ 0.029 <module>  cupy/testing/__init__.py:1
│     [4 frames hidden]  cupy, unittest, asyncio
└─ 0.011 <module>  cupy/_functional/vectorize.py:1
   └─ 0.011 <module>  cupyx/jit/__init__.py:1
```
Now we see environment detection as having spiked the import time, as expected. If I apply the changes in this PR, the difference to the original environment without rapidsai installed is minimal, almost negligible.
```
0.356 <module>  cupy/__init__.py:1
...
├─ 0.017 _detect_duplicate_installation  cupy/_environment.py:566
│  └─ 0.017 distribution  importlib/metadata/__init__.py:954
...
```
I would put that delta (0.026 to 0.017) as within the margin of noise. However, in the environment with rapidsai installed, we instead see 
```
0.761 <module>  cupy/__init__.py:1
...
├─ 0.020 _detect_duplicate_installation  cupy/_environment.py:566
│  └─ 0.020 distribution  importlib/metadata/__init__.py:954
│        [7 frames hidden]  importlib, <built-in>
...
```
Here we see a significant drop. As expected. The more complex environment is the one where you get a more meaningful speedup.

Note that since cupy handles some conditional imports gracefully, the cupyx import time also spikes in the rapidsai environment because scipy gets installed. Similarly, cupy.testing [conditionally imports pytest](https://github.com/cupy/cupy/blob/main/cupy/testing/_pytest_impl.py#L7), which would contribute significant additional import time if it were installed. The former would be improved by addressing https://github.com/cupy/cupy/issues/9324 or by lazy loading scipy as in https://github.com/cupy/cupy/pull/8706. I'll open a separate PR to address the latter.

kmaehashi:
Thanks @vyasr!

The followings can be safely removed from the list, as these packages only supports Python 3.9 or earlier and thus will never coexist with CuPy v14.

```
        'cupy-cuda80',
        'cupy-cuda90',
        'cupy-cuda91',
        'cupy-cuda92',
        'cupy-cuda100',
        'cupy-cuda101',
```

The followings can also be removed as they were not shipped actually.

```
        'cupy-cuda118',
        'cupy-rocm-4-1',
```

vyasr:
> Thanks @vyasr!
> 
> The followings can be safely removed from the list, as these packages only supports Python 3.9 or earlier and thus will never coexist with CuPy v14.
> 
> ```
>         'cupy-cuda80',
>         'cupy-cuda90',
>         'cupy-cuda91',
>         'cupy-cuda92',
>         'cupy-cuda100',
>         'cupy-cuda101',
> ```
> 
> The followings can also be removed as they were not shipped actually.
> 
> ```
>         'cupy-cuda118',
>         'cupy-rocm-4-1',
> ```

Done, thanks!

kmaehashi (approved):

PR #9331, opened 2025-08-26
Random sampling without replacement from range(a)

### LINKED ISSUE #9320: Reduce memory usage of `cupy.random.choice(N, replace=False)`
### Description

Currently `cupy.random.choice(N, size=M, replace=False)` generates a temporary array of indices of length `N`:

https://github.com/cupy/cupy/blob/541b9c6af4045172cd97a3c045e375189c811519/cupy/random/_generator.py#L1125-L1126

For large `N` this can result in a very large allocation, even if the output itself is small. For example:

```python
import cupy as cp

# Choice 10 unique integers in [0, 100_000_000)
cp.random.choice(100_000_000, size=10, replace=False)
```

That allocates a ~0.75 GiB temporary array, even though the output itself is only 80B.

`numpy.random.choice` uses a [more efficient implementation](https://github.com/numpy/numpy/blob/eb0e901e638e498ecba639b8e3e22bb2208f5be1/numpy/random/_generator.pyx#L963-L987) when selecting small number of unique values from a larger range, and doesn't suffer from this issue.

```python
# This doesn't allocate a huge temporary array
np.random.choice(100_000_000, size=10, replace=False)
```

For now I can make use of the latter, but it would be nice to have a memory efficient implementation in cupy too.

### Additional Information

_No response_

sanuzu:
Fixes #9320

Optimizes `cupy.random.choice(N, size=M, replace=False)` by implementing random sampling of indices without the need for a temporary array.

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -1137,7 +1137,7 @@ class RandomState:
                     'Cannot take a larger sample than population when '
                     '\'replace=False\'')
             if isinstance(a, int):
-                # Floyd's algorithm for sampling without replacement
+                # Random sampling without replacement from range(a)
 
                 # Ensure size is an integer before using it
                 size = int(size)
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -1074,6 +1074,20 @@ class RandomState:
         self._update_seed(y.size)
         return y
 
+    def _gen_mask(self, max_val):
+        """Generates the smallest power of 2 larger than or equal to max_val"""
+        # Start with max_val
+        mask = max_val
+        # Propagate 1's through all higher bits to ensure the mask covers all bits greater than max_val
+        mask |= mask >> 1
+        mask |= mask >> 2
+        mask |= mask >> 4
+        mask |= mask >> 8
+        mask |= mask >> 16
+        mask |= mask >> 32
+        # Return the smallest power of 2 that is greater than or equal to max_val
+        return mask
+
     def choice(self, a, size=None, replace=True, p=None):
         """Returns an array of random values from a given 1-D array.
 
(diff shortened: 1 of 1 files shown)


cupy/random/_generator.py
```diff
@@ -1123,7 +1137,36 @@ def choice(self, a, size=None, replace=True, p=None):
...
+                # To store selected indices
+                indices = cupy.empty(size, dtype=cupy.int64)
+                for j in range(a_size - size, a_size):
+                    # Generate random index
+                    val = self.randint(0, j)
+                    loc = val & mask  # Hashing the value
+
+                    # Resolve collision with linear probing
+                    while hash_set[loc] != -1 and hash_set[loc] != val:
+                        loc = (loc + 1) & mask
```
leofang, line 1159: 1. We need to avoid looping in Python, this would launch many many small kernels
2. Having `while` with GPU values would cause data transfer and host-device synchronization, making the perf even worse

leofang (changes requested):
Hi @sanuzu, appreciate for your enthusiasm and contribution! However, please make sure you run benchmarks for performance optimization tasks like this and include the result in the PR description. I think you'll see that this code performs unfortunately way worse than the status quo and is not something we can accept. Most likely, we need a custom kernel here to generate k random indices from [0, N).

cupy/random/_generator.py
sanuzu: Thank you for the feedback! I see now that I misunderstood the performance considerations. This is my first time contributing to CuPy, and I didn’t fully realize the impact of the changes I made.

Regarding the custom kernel, could you please point me to a reference file or an example where a similar kernels are implemented? That would help me understand how to write the kernel and implement it efficiently.

Thanks again for your guidance!

kmaehashi: For an example of raw kernel usage, see https://github.com/cupy/cupy/tree/main/examples/gemm. However, there isn't a comparable or "similar" thing to this specific issue in CuPy, and if you're new to writing CUDA kernels, this may not be the best first task to tackle.

seberg:
Thanks for the attempt at this @sanuzu, this was solved with a lower level and much faster approach in another PR now though.

Issue #9333, opened 2025-08-27
Update bundled CCCL to v3


leofang:
It was an extremely rare (and pleasant!) case that we had a smooth transition to CUDA 13 #9286 #9289 with the bundled CCCL version happening to support **THREE** CUDA major versions (11, 12, 13) and thus unchanged during the transition. It was always much more brittle than this time.

We should not take this as granted. In particular, the bundled version (somewhere close to the latest v2.x) will soon start fighting against future CUDA 13.x. The upstream has dropped CUDA 11 support. We should also follow and update to v3.x as part of #9319.

PR #8952, opened 2025-02-14
Broaden usage of C++17

jakirkham:
Follow up on this discussion in PR: https://github.com/cupy/cupy/pull/8951

Switch more C++14 flags to C++17.

This should help with compilation issues we have seen (like Windows CUDA 11.2).

PR #8940, opened 2025-02-11
Use C++17 in JIT compile

kmaehashi:
This looks required to support the latest CCCL 2.8.x in CUDA 11.2 on Windows.
xref: https://github.com/cupy/cupy/pull/8915#issuecomment-2649595959

cc/ @leofang @jakirkham

PR #9334, opened 2025-08-28
Change default c++ RTC standard from 11 to 14

gpinkert:
To help modernize development, AMD would like to bump the default RTC c++ standard from 11 to 14. 

This would give projects using cupy and AMD hardware zero day support for ROCm 7.0.0 and encourage
development using newer standards. Some new features in c++ 14:

* Generic lambdas (`[] (auto x, auto y) { … }`)
* Lambda init-capture (`[p = std::move(p)] { … }`)
* Return type deduction for normal functions (`auto f(){ … }`)
* Relaxed `constexpr` (loops/branches allowed in `constexpr` functions)
* Variable templates (`template<class T> constexpr T k = …;`)
* `std::integer_sequence` / `std::index_sequence`
* `_t` type aliases (e.g., `std::enable_if_t`, `std::decay_t`, `std::remove_reference_t`)
* `std::exchange`
* Binary literals & digit separators (`0b1010`, `1'000'000`)
* `[[deprecated]]` attribute
* More literal/`constexpr`-friendly aggregates and `std::array`

Given that there are no deprecations, tests should not need updating. Once extra CUDA test failures have been identified in the PN CI, I can triage them and create a report on how difficult it would be to upgrade and fix those tests such that we can still upgrade the default.

I have also created two more functions to help improve readability. I understand that this is a personal preference, so I can revert back to placing all c++ standard logic into the assemble_cupy_compiler_options function.

### DIFF
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -2326,14 +2326,33 @@ cdef inline str _translate_cucomplex_to_thrust(str source):
     return ''.join(lines)
 
 
+cpdef bint use_default_std(tuple options):
+    cdef str opt
+    for opt in options:
+        if ('-std=' in opt) or ('--std=' in opt):
+            return False
+    return True
+
+cpdef void warn_on_unsupported_std(tuple options):
+    cdef str opt
+    for opt in options:
+        if _is_hip:
+            major, minor = nvrtc.getVersion()
+            if '-std=c++11' in opt and minor == 0 and major == 7:
+                warnings.warn(
+                    'hipRTC on some ROCm 7.0 builds have a known bug '
+                    'that causes RTC to break when the standard is set '
+                    'to c++11. Please use c++14, or use ROCm >= 7.0'
+                )
+        elif '-std=c++03' in opt:
+            warnings.warn('CCCL requires c++11 or above')
+
+
 cpdef tuple assemble_cupy_compiler_options(tuple options):
-    for op in options:
-        if '-std=c++' in op:
-            if op.endswith('03'):
-                warnings.warn('CCCL requires c++11 or above')
-            break
+    if use_default_std(options):
+        options += ('--std=c++14',)
     else:
-        options += ('--std=c++11',)
+        warn_on_unsupported_std(options)
 
     # make sure bundled CCCL is searched first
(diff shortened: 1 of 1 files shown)


cupy/_core/core.pyx
```diff
@@ -2326,14 +2326,33 @@ cdef inline str _translate_cucomplex_to_thrust(str source):
...
+cpdef void warn_on_unsupported_std(tuple options):
+    cdef str opt
+    for opt in options:
+        if _is_hip:
+            major, minor = nvrtc.getVersion()
+            if '-std=c++11' in opt and minor == 0 and major == 7:
+                warnings.warn(
+                    'hipRTC on some ROCm 7.0 builds have a known bug '
+                    'that causes RTC to break when the standard is set '
+                    'to c++11. Please use c++14, or use ROCm >= 7.0'
```
kmaehashi:
```suggestion
                    'to c++11. Please use c++14, or use ROCm > 7.0'
```

cupy/_core/core.pyx
```diff
@@ -2326,14 +2326,33 @@ cdef inline str _translate_cucomplex_to_thrust(str source):
...
+                )
+        elif '-std=c++03' in opt:
+            warnings.warn('CCCL requires c++11 or above')
```
kmaehashi, line 2348: note: C++17 is required once we bump to CCCL v3 (#9333) cc/ @leofang

cupy/_core/core.pyx
gpinkert: Nice catch

kmaehashi:
Merging, test failures are unrelated to the PR.

cupy/_core/core.pyx
leofang: Thanks for the ping @kmaehashi, just returned from vacation and catching up 🙂 Yes, when we start working on #9333 bumping to C++17 across the repo is a must. NVIDIA has dropped the support for C++14 and below. CuPy has been prepared for this transition anyway (ex: https://github.com/cupy/cupy/pull/8952, https://github.com/cupy/cupy/pull/8940, https://github.com/cupy/cupy/pull/9159), so I don't expect any hiccups.

PR #9335, opened 2025-08-28
[feat] Fix failing pytest import and add test_raw tests

gpinkert:
Previously, the @unittest.skipIf decorator in test_raw.py was causing an import‑time assertion in find_compiler_ver(), which made pytest abort collection with errors instead of skipping tests.

This commit fixes this by moving the is_hip check up first. Since the pytest skipIfs are evaluated short circuit, there is no call to find_nvcc_ver() which is what caused the import error.


kmaehashi (approved):
Thanks, LGTM!

kmaehashi:
Merging, test failures are unrelated to the PR.

PR #9336, opened 2025-08-29
Lazy load the testing module

vyasr:
As part of the import time investigations I have been doing regarding #9324 and #9325 I also noticed that cupy.testing adds significant testing overhead in environments where pytest is installed. I considered making the pytest import lazy or trying to hide its cost in other ways, but ultimately I decided that since most users of cupy do not need the testing module directly the best option would be to make the testing module import in cupy's `__init__.py` lazy and hide all of the overhead. The module load is fast enough that for users actually accessing the testing module the first attribute access (which triggers the load, and is therefore made slower by making this module load lazy) is still less than 0.1 s so it doesn't degrade that experience much (in exchange for making the majority use case of importing cupy without using testing faster).

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -1,4 +1,6 @@
 from __future__ import annotations
+import sys
+import importlib.util
 
 import functools as _functools
 import sys as _sys
@@ -45,8 +47,6 @@ from cupy import polynomial  # NOQA
 from cupy import random  # NOQA
 # `cupy.sparse` is deprecated in v8
 from cupy import sparse  # NOQA
-from cupy import testing  # NOQA  # NOQA
-
 
 # import class and function
 from cupy._core import ndarray  # NOQA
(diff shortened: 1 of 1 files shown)


cupy/__init__.py
```diff
@@ -1146,8 +1146,21 @@ def _embed_signatures(dirs):
 
 
 _embed_signatures(globals())
-_embed_signatures(fft.__dict__)
-_embed_signatures(linalg.__dict__)
-_embed_signatures(random.__dict__)
-_embed_signatures(sparse.__dict__)
-_embed_signatures(testing.__dict__)
```
vyasr, line 1153: I removed these because by inspection all the ufuncs are handled by the `globals()` call above. None of the other module dicts actually contain ufuncs. I verified this by adding print statements in `_embed_signatures`.

kmaehashi:
Thanks @vyasr! I've pushed minor tweaks to avoid exposing extra attributes under `cupy.*` namespace (c1feacbe420bd19c5ef9c02fd86bc616f537a6e6).

PR #9337, opened 2025-08-30
Remove deprecated modules

vyasr:
These modules were deprecated in cupy 12 (in https://github.com/cupy/cupy/commit/cd14be90db4cf2b67003a7074bbc9a055a0eec3a) almost three years ago and are suitable for removal now. I updated the upgrade guide to indicate the simple fix of importing them from cupyx instead. This change contributes to #9324.

PR #9338, opened 2025-08-30
Delay all imports of cupyx inside cupy

vyasr:
This PR moves almost all cupyx imports inside cupy to be function-local. The remaining imports are already being removed in other ways (such as #9337). This PR is meant as an experiment to facilitate #9324, namely to see what it would take for cupy to stop importing cupyx. It seems like there is a significant chunk of cupy functionality that relies on cupyx, so I am not certain how realistic it is to completely remove the dependency. It may be possible, but I would like some more input from developers before proceeding any further. At minimum, I think moving to function-local imports like this and delaying the import of cupyx until it is really needed is worthwhile since many use cases for cupy will never require cupyx, so the function-local imports of cupyx will never trigger.


vyasr:
@kmaehashi @leofang I'm curious what you think of this PR. It seems like there is some nontrivial cupy functionality that currently depends on cupyx. If the goal is #9324, then should some of these functions actually be moved from cupyx to cupy? Are there some cases where the deferred import approach used in this PR is more suitable (for example, where cupy use cutensor for acceleration since that is an optional feature)? Is there another solution that you would prefer?

vyasr:
@kmaehashi what are your thoughts on this PR? It gets us part of the way to #9324 by making all cupyx imports lazy, and it gives us the immediate import time improvements. I don't think I would want this PR to be the final state and it would still be good to work through actually removing the dependency, but I don't know cupy well enough to be able to judge each of the usages of cupyx in this PR and say where code should be moved from cupyx to cupy, when imports should be left as optional, or something else.

@leofang had some ideas on some modules that may make sense to move from cupyx to cupy, and @emcastillo  a test offline so I added a simple one in https://github.com/cupy/cupy/pull/9338/commits/53be1fad007c5d24a17c11797fee20bd098e579c.

leofang:
@vyasr sorry for delay. This looks fine to me, I do want to make sure import cupy does not import cupyx right away, but @mdboom recently told me that hiding imports like this could incur extra runtime overhead. Any chance you could pick a function, benchmark it, and see if this overhead is really noticeable?

vyasr:
> @vyasr sorry for delay.

No problem.

> This looks fine to me, I do want to make sure import cupy does not import cupyx right away, but @mdboom recently told me that hiding imports like this could incur extra runtime overhead.

Yes, delayed imports like this mean that there are `PyImport_*` calls happening in each function. However, this cost is quite low after the first time since later imports are effectively just a lookup in sys.modules. For instance, here's a trivial example:
```python
In [1]: import sys

In [2]: def f():
   ...:     sys.path
   ...:

In [3]: %timeit f()
14.7 ns ± 0.19 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [4]: def g():
   ...:     import sys
   ...:     sys.path
   ...:

In [5]: %timeit g()
56.6 ns ± 0.203 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

So with this trivial example the overhead is about 40 ns per call. In the context of cupy calls, probably negligible.

> Any chance you could pick a function, benchmark it, and see if this overhead is really noticeable?

In practice, here's an example. I picked `cupy.linalg.cond` since that function seems like it should be one of the simpler ones that was updated. Here's the performance in this PR, looking at both trivially small and somewhat nontrivial matrices to span some range of work:
```python
In [1]: import cupy as cp

In [2]: mat = cp.random.rand(100, 100)

In [3]: %timeit cp.linalg.cond(mat)
3.56 ms ± 6.66 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: mat = cp.random.rand(2, 2)

In [5]: %timeit cp.linalg.cond(mat)
337 μs ± 1.37 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

And here's what it looks like on main:
```python
In [1]: import cupy as cp

In [2]: mat = cp.random.rand(100, 100)

In [3]: %timeit cp.linalg.cond(mat)
3.59 ms ± 7.73 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: mat = cp.random.rand(2, 2)

In [5]: %timeit cp.linalg.cond(mat)
337 μs ± 873 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

So not really measurable there.

vyasr:
Note that I'm not sure we should mark this as actually closing #9324 unless we think that there's no hope of removing the imports altogether. This PR just delays them.

leofang:
> @leofang had some ideas on some modules that may make sense to move from cupyx to cupy

I don't remember what I said lol But I think we should get this PR merged and revisit the refactoring in another PR.

cupy/lib/_routines_poly.py
```diff
@@ -162,6 +161,7 @@ def _polypow_direct(x, n):
 
 
 def _polypow(x, n):
+    import cupyx.scipy.fft
```
leofang: This can be moved to where `next_fast_len` is called

cupy/linalg/_einsum.py
```diff
@@ -398,25 +380,36 @@ def reduced_binary_einsum(arr0, sub0, arr1, sub1, sub_others):
...
+        if accelerator != _accelerator.ACCELERATOR_CUTENSOR:
+            continue
+        try:
+            import cupy_backends.cuda.libs.cutensor  # NOQA
+            from cupyx import cutensor
+        except ImportError:
+            continue
+        else:
+            if (
+                not cutensor.check_availability('contraction')
```
leofang: this can be removed (`cutensorContraction` always exists)

cupy/lib/_routines_poly.py
vyasr: Sure.

cupy/linalg/_einsum.py
vyasr: OK sure!

vyasr:
> > @leofang had some ideas on some modules that may make sense to move from cupyx to cupy
> 
> I don't remember what I said lol But I think we should get this PR merged and revisit the refactoring in another PR.

I agree, I just mean that you linked this PR to the issue so now merging this PR will close the issue and I don't think we want that. I'll find our previous discussion and add your notes to the issue so that we can continue further discussion there.

vyasr:
Left a note at https://github.com/cupy/cupy/issues/9324#issuecomment-3500994586

vyasr:
@leofang let me know if you need anything else here.

leofang (approved):

seberg:
Hmmm, strange, CI seems to be failing due to:
```
Cleaning up the execution context...	
01:18:12.143781 INFO command finished: code=1, status=[OK]
```
after the tests pass.

EDIT: Sorry, I looked wrong, the problem is that the new `cupyx` import test is failing.

.pfnci/linux/tests/actions/unittest.sh
```diff
@@ -43,3 +43,6 @@ case ${test_retval} in
         exit $test_retval
         ;;
 esac
+
+# Validate that importing cupy does not import cupyx
+python3 -Ximporttime -c "import cupy" |& grep -q cupyx && (echo "cupyx was imported by cupy"  && exit 1)
```
seberg:
```suggestion
# grep returns 0 if it finds anything so invert the result.
! $(python3 -Ximporttime -c "import cupy" |& grep -q cupyx) || (echo "Found forbidden import 'cupyx'" && exit 1)
```
Maybe there is a nicer bash way, but the above seems to fail. The reason is that if `grep -q cupyx` finds nothing, that is also the return value of the full command.

So the tests fails even though everything is fine.

(I checked this locally with a minimal script and it does seem to cause a 0 (with `$?`) exit return for the happy case and `exit 1` with a 1 return for the non-happy one.)

seberg:
Let's just try the fix, I really think it should be fine and had tested locally that it fails if changed to `cupy` rather than `cupyx`.

/test mini

.pfnci/linux/tests/actions/unittest.sh
vyasr: Yikes sorry thanks for the fix @seberg! You're definitely right that the issue here is with grep's return code.

seberg:
Thanks Vyas, had look through the changes once more, let's get this in, since I think the only reason it wasn't was the CI anyway.

Issue #7431, opened 2023-02-27
Array Interface for HIP/ROCm


JBlaschke:
### Description

We're working on a GPU-accelerated Python code which will run on NVIDA and AMD GPUs. Some of our accelerated kernels are written in Pybind11, and expose a cuda array interface. We're working on moving data between kernels and to CuPy without copying to the host first, but at the end of the pipeline the result needs to be converted to a numpy array.

On AMD, we're running into:
```
"/gpfs/alpine/chm137/scratch/djh992/spinifel_crusher_122022/setup/conda/envs/myenv/lib/python3.8/site-packages/cupy/_creation/from_data.py", line 76, in asarray
    return _core.array(a, dtype, False, order)
...
RuntimeError: HIP/ROCm does not support cuda array interface
```

Which looks like it's running into this:
https://github.com/cupy/cupy/blob/a5b24f91d4d77fa03e6a4dd2ac954ff9a04e21f4/cupy/core/core.pyx#L2480

Any thoughts about how to address this? From what I can see, ROCm and CUDA device pointers look very similar (might even be the same). 

@leofang I see you are thinking along the same lines -- even with Numba no longer support ROCm (for now -- fingers crossed), exposing a `__cuda_array_interface__` for ROCm. Are there any reasons why not to do this? Even if it's something stupid like this: https://github.com/numba/numba/issues/4566#issuecomment-682915186

@ax3l @elliottslaughter tagging you also.

Related: 
1. https://github.com/cupy/cupy/issues/4644
2. https://github.com/numba/numba/issues/4566

leofang:
I am currently in a position where I'd feel uneasy to engage in any ROCm-related discussion 😅 

That said, I think the community is moving to use DLPack, and in my previous life I made sure ROCm is considered. I don't think you should overload `__cuda_array_interface__`, and if you still want to pursue this, you should move this discussion to the Numba repo, not here.

Refs:
1. DLPack: https://dmlc.github.io/dlpack/latest/
2. CAI: https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html

ax3l:
x-ref: #4458 by @emcastillo

domcharrier:
@JBlaschke A couple of recent developments from the ROCm side that might be able to help you if this issue is still relevant.

* https://github.com/ROCm/numba-hip
  * Runtime is based on: https://github.com/ROCm/hip-python
[...]

PR #9340, opened 2025-08-31
Dev/gpinkert/add cuda array interface

gpinkert:
This PR enables cuda array interface. Great thanks from Philipp Samfass and Dominic Charrier

### DIFF
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -347,7 +347,7 @@ class TestNdarrayCudaInterfaceStream(unittest.TestCase):
     def setUp(self):
         if cupy.cuda.runtime.is_hip and self.stream == 'ptds':
             self.skipTest('HIP does not support PTDS')
-        
+
         if self.stream == 'null':
             self.stream = cuda.Stream.null
         elif self.stream == 'new':
--- a/tests/cupy_tests/core_tests/test_ndarray.py
+++ b/tests/cupy_tests/core_tests/test_ndarray.py
@@ -387,24 +387,6 @@ class TestNdarrayCudaInterfaceStream(unittest.TestCase):
                 assert iface['stream'] == stream.ptr
 
 
-class TestNdarrayCudaInterfaceAttr(unittest.TestCase):
-
-    def setUp(self):
-        self.arr = cupy.zeros(shape=(2, 3), dtype=cupy.float64)
-
-    def test_cuda_array_interface_hasattr(self):
-        assert hasattr(self.arr, '__cuda_array_interface__')
-
-    def test_cuda_array_interface_getattr(self):
-        try:
-            interface = getattr(self.arr, '__cuda_array_interface__')
-            self.assertIsNotNone(interface)
-        except AttributeError as e:
-            self.fail(
-                "__cuda_array_interface__ should be present, "
-                f"but got an AttributeError: {str(e)}")
-
-
 @testing.parameterize(
     *testing.product({
         'indices_shape': [(2,), (2, 3)],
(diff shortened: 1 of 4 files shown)


tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -388,20 +387,22 @@ def test_cuda_array_interface_stream(self):
                 assert iface['stream'] == stream.ptr
 
 
-@pytest.mark.skipif(not cupy.cuda.runtime.is_hip,
-                    reason='This is supported on CUDA')
-class TestNdarrayCudaInterfaceNoneCUDA(unittest.TestCase):
+class TestNdarrayCudaInterfaceAttr(unittest.TestCase):
```
kmaehashi: I think it's ok to remove this entire test class. It does not make sense as `__cuda_array_interface__` no longer raises `AttributeError`.

kmaehashi (commented):
LGTM except for nits!

kmaehashi:
Please also fix cosmetic errors reported in static-checks CI.

leofang:
FYI, my recommendation in https://github.com/cupy/cupy/issues/7431#issuecomment-1447038466 still stands:
- using DLPack is preferred and it should be already working with CuPy, except perhaps for managed memory for which I suggested Dominic to raise a simple patch upstream: https://github.com/dmlc/dlpack/issues/138
- CAI is not fully compatible with ROCm/HIP as-is, at least with regard to the `stream` support

Now, here's a new update I can provide since I left my comment more than two years ago 🙂 We (wearing my NVIDIA hat) are in the process of owning Numba's CUDA target (see [here](https://numba.readthedocs.io/en/stable/cuda/overview.html#built-in-cuda-target-deprecation-and-maintenance-status) and [here](https://github.com/numba/numba/pull/10138)), including by nature the CAI spec which has a new home [here](https://nvidia.github.io/numba-cuda/user/cuda_array_interface.html). So it no longer makes sense to move the discussion upstream (Numba) as I originally suggested. It is fine that for the ROCm/HIP backend to piggyback on CAI, but I suggest @gpinkert and the ROCm/HIP team to provide a linkable webpage on the ROCm/HIP website explaining the semantic differences when using CAI with the ROCm/HIP backend. Then, we (wearing back my CuPy hat...) can link to it in our [Interoperability page](https://docs.cupy.dev/en/stable/user_guide/interoperability.html).

gpinkert:
@leofang When referring to:
> semantic differences when using CAI with the ROCm/HIP backend

Do you mean how cupy on ROCm doesn't support PTDS? I have an internal branch that adds PTDS support and I plan on upstreaming it as quickly as possible. 

I am still new to the python side of GPU programming, so apologies if I am missing something obvious.

leofang:
> When referring to:
> 
> > semantic differences when using CAI with the ROCm/HIP backend
> 
> Do you mean how cupy on ROCm doesn't support PTDS? I have an internal branch that adds PTDS support and I plan on upstreaming it as quickly as possible.

Yes IIRC I think this is the main one. CAI specifies that `stream=1` uses the legacy default stream and `stream=2` uses the PTDS, and disallows `stream=0` since it's ambiguous. (FWIW, DLPack also inherited this bit of design from CAI.) Back then, ROCm did not support PTDS and did not have a construct that maps `1` to the legacy one, so it could not support the CAI spec fully. 

The other aspect still stands: It'd be nice to have an official page in ROCm/HIP docs to explain how to interpret the CAI spec, even if it is just a one-liner stating "it's fully supported."

gpinkert:
@leofang Gotcha, just wanted to make sure I understand. I can get on that page. In the meantime, is this a blocker for getting ROCm CAI merged?

PR #9345, opened 2025-09-01
Fix percentile/quantile empty axes handling and added tests

nitinvunnam:
The PR fixes the crashes and illegal memory access that occurs when working with empty arrays with cupy.percentile and cupy.quartile. Added checks in order.py to raise an IndexError when operating on empty axes to match NumPy. Added test cases but had to mark two of them as xfails because of the reshape bug (ValueError: cannot reshape array of size 0 into shape (0, -1)). If needed can update implementation so xfails are resolved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ htmlcov/
 cupy/.data/
 cupy/cuda/thrust.h
 cupy/_core/numpy_allocator.h
+.vscode/
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "git.enabled": true
+}
--- a/cupy/_statistics/meanvar.py
+++ b/cupy/_statistics/meanvar.py
@@ -62,6 +62,16 @@ def nanmedian(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     .. seealso:: :func:`numpy.nanmedian`
 
     """
+    a = cupy.asarray(a)
+    if a.ndim == 0 and axis is not None:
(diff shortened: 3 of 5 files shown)


nitinvunnam:
ignore the handling zerodim and chore accidentally pr'd this off that branch instead of main

seberg:
Thanks for looking into this @nitinvunnam.  Could you move this fix down one level?  And maybe be up for a bit more maintenance here?
The `_routines_manipulation.pyx` file has a `_normalize_axis_tuple`, I think that could be moved into `cupy/_core/internal.pyx` together with `_normalize_axis_indices`.

After that we could use it here.  Otherwise, as a minimal fix, I would suggest to just call `_normalize_axis_indices()` in `nanmedian` (since `median` seems to get things right).  That also fixes e.g. repeated entries.
Just adding this check in front works, but adds a layer in front of existing checks, which makes this harder to maintain long-term.

nitinvunnam:
Yeah sorry I'm a little busy right now so I'll fix it in the next couple of days. Thanks for the help and suggestions.

seberg:
@nitinvunnam noticed that this was a while ago. I realize I asked for a slightly larger fix, but gentle ping in case you are still planning to look into it.

nitinvunnam:
> @nitinvunnam noticed that this was a while ago. I realize I asked for a slightly larger fix, but gentle ping in case you are still planning to look into it.

sorry about not responding, ive been a little busy in the next week or two i can look into it and try and get it done. once again, sorry about taking a while

PR #9351, opened 2025-09-03
feat(import) Raise import error

gpinkert:
There are many issues with RTC and hipcc compilation if ROCM_HOME is not set. This commit makes it impossible to import cupy unless ROCM_HOME has been set. Missing ROCM_HOME error messages can take a long time to triage,. and the error messages from hipcc/hiprtc are usually not helpful for users to debug. 

Hopefully, this will result in less back and forth among development teams.

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import os as _os
 
 import functools as _functools
 import sys as _sys
@@ -51,6 +52,11 @@ from cupy import testing  # NOQA  # NOQA
 # import class and function
 from cupy._core import ndarray  # NOQA
 from cupy._core import ufunc  # NOQA
+
+if getattr(cuda.runtime, "is_hip", False) and "ROCM_HOME" not in _os.environ:
+    raise ImportError("ROCM_HOME is not set but a HIP/ROCm backend was "
+                      "detected. Please set with `export ROCM_HOME=/opt/rocm`"
+                      " or similar.")
 
 
 # =============================================================================


kmaehashi:
Thanks @gpinkert. What would be an ideal strategy to detect ROCm directory? Currently we are using the following methods (in order of priority) to discover the toolkit path even when the env var is not set, which is aligned between CUDA and ROCm.

* For CUDA: `CUDA_PATH` env var -> parent directory of `nvcc` -> `/usr/local/cuda`
* For ROCm: `ROCM_HOME` env var -> parent directory of `hipcc` -> `/opt/rocm`

https://github.com/cupy/cupy/blob/92c282384a22f9c72af98303542f64a4a2085bc6/cupy/_environment.py#L146-L161

I guess it would be better to fail if none of the above paths could not be found, rather than quickly failing import when `ROCM_HOME` is unset?

Another thing I would like to discuss with you is this issue: https://github.com/cupy/cupy/issues/4493#issuecomment-883401443
I guess it's better to use `ROCM_PATH` to better align with ecosystem (convention of ROCm toolkit and other libraries)?

gpinkert:
Hi @kmaehashi - to be honest there doesn't seem to be clear guidance on which is preferred from the ROCm side. I looked internally, and there are some projects that set ROCM_HOME and some that set ROCM_PATH. 

I made a small test to see if setting either variable makes a difference when compiling simple kernels with hipRTC or hipcc. Long story short, it doesn't make a difference.

Up to you - I think supporting two env vars is confusing and would prefer to consolidate the logic using one env var. I am leaning towards ROCM_PATH so it is more similar to Nvidia, but up to you.

kmaehashi:
> I looked internally, and there are some projects that set ROCM_HOME and some that set ROCM_PATH.

I see, then let's keep as is as there's no strong reasons to break the existing behavior.

gpinkert:
> > I looked internally, and there are some projects that set ROCM_HOME and some that set ROCM_PATH.
> 
> I see, then let's keep as is as there's no strong reasons to break the existing behavior.

Makes sense to me. Moving forward, does that mean removing references to ROCM_HOME, or erroring when there are conflicting ROCM_PATH/ROCM_HOME? I think it is likely confusing supporting both, and error message explaining the difference and how things could go wrong would be beneficial IMO

Issue #8448, opened 2024-08-01
cupyx.scipy.signal.upfirdn failing on 2D inputs


bmerry:
### Description

Using cupyx.scipy.signal.upfirdn is failing CUDADriverError for some combinations of parameters when the input signal is 2D. It seems to depend on both the types and the up/down factors.

### To Reproduce

```py
#!/usr/bin/env python3

import cupy as cp
import cupyx.scipy.signal
import numpy as np
import scipy.signal

h = np.zeros(101, np.float32)
x = np.zeros((8, 500), np.complex64)
print(scipy.signal.upfirdn(h, x, 3, 2))

h = cp.asarray(h)
x = cp.asarray(x)
cupyx.scipy.signal.upfirdn(h, x, 3, 2)
```

Output:
```
/home/bmerry/.pyenv/versions/env/lib/python3.12/site-packages/cupyx/jit/_interface.py:173: FutureWarning: cupyx.jit.rawkernel is experimental. The interface can change in the future.
  cupy._util.experimental('cupyx.jit.rawkernel')
[[0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 ...
 [0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]]
Traceback (most recent call last):
  File "/home/bmerry/work/bugs/cupy-upfirdn/./upfirdn2d.py", line 14, in <module>
    cupyx.scipy.signal.upfirdn(h, x, 3, 2)
  File "/home/bmerry/.pyenv/versions/env/lib/python3.12/site-packages/cupyx/scipy/signal/_upfirdn.py", line 500, in upfirdn
    return ufd.apply_filter(x, axis)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bmerry/.pyenv/versions/env/lib/python3.12/site-packages/cupyx/scipy/signal/_upfirdn.py", line 417, in apply_filter
[...]

PR #9352, opened 2025-09-04
[BugFix] Fix upfirdn kernel launch bug for 2D arrays

IdanFried:
Fixed https://github.com/cupy/cupy/issues/8448

This PR changed the launch parameters of the upfirdn function to fix the issue.
It also adds tests for the 2D case.


tests/cupyx_tests/scipy_tests/signal_tests/test_resample.py
```diff
@@ -190,6 +190,22 @@ def test_resample_methods(self, method, ext, padtype, xp, scp):
 
         return results
 
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_resample2d(self, xp, scp):
+        results = []
+        for nx in range(10000, 100000, 10000):
```
seberg: NIT: Curious if this test is slow, but doesn't seem over the top.
```suggestion
        for nx in range(10_000, 100_000):
```
or add another size, but one is repeated.

seberg (dismissed):
Haven't looked super close, but looks like a clear typo to swap blocks/threads there and the test looks like it should cover things.  So LGTM modulo that test typo.

tests/cupyx_tests/scipy_tests/signal_tests/test_resample.py
IdanFried: This test crushed my IDE twice (but I have a pretty old GPU).
What do you mean by "add another size, but one is repeated"? No size is repeated.

seberg: sorry must have been asleep when looking at it, read it as an explicit list

> This test crushed my IDE twice (but I have a pretty old GPU).

Maybe it is a bit too excessive/large problems, I doubt this very large sizes matter for the fix, so it may be better to just make it smaller.
(Unless you mean my suggestion, which clearly was nonsense, sorry about that.)

IdanFried: No problem :)
The current test works fine. I mean your suggestion.

tests/cupyx_tests/scipy_tests/signal_tests/test_resample.py
```diff
@@ -190,6 +190,22 @@ def test_resample_methods(self, method, ext, padtype, xp, scp):
...
+            for down in [2, 3, 8]:
+                for ny in range(1, 50, down):
+                    for axis in [0, 1]:
```
seberg: @IdanFried are you still happy to update this?  I am happy to just put this in, but this test needs to be thinned out by a factor of ~100 (or more).

Running it locally, it takes 85 seconds and I don't see that is needed, CI is expensive, so let's try to get it to <1s and if a bit more extensive testing is really relevant than that may be OK but with a `@testing.slow` marker.

EDIT: The best option may be to parametrize it to ~4 interesting explicit cases (and probably also make them smaller than the current ones).

IdanFried: I narrowed it down to 3 explicit tests.
The runtime is now ~900ms on my machine; if it runs for longer on the CI machines, I can add the `@testing.slow` decorator.

seberg:
> I narrowed it down to 3 explicit tests.
> The runtime is now ~900ms on my machine; if it runs for longer on the CI machines, I can add the @testing.slow decorator.

Thanks a lot, especially for following up! While I am a bit new around here, I think it should be OK.  It's <1s now if I don't run it as an individual test (the first test is always slower).

(All time is spend in the first parametrization actually, so not sure if it's not just the due to it being the first time we launch this kernel. -- IIRC that is also the one parametrization that fails without the fix.)

/test mini

seberg (approved):

PR #9380, opened 2025-09-13
feat(CI) ROCm 6.4.3

gpinkert:
Added support for ROCm 6.4.3 in CI. Using the complete image so that we don't have to download additional dependencies. 

Confirmed that the build compiles on gfx90a. Let me know if there is anything else that I have to do.

### DIFF
--- a/.github/workflows/self-hosted-ci.yml
+++ b/.github/workflows/self-hosted-ci.yml
@@ -92,7 +92,7 @@ jobs:
         token: ${{ steps.generate-token.outputs.token }}
         client-payload: |
           {
-            "targets": ["rocm-6-2"],
+            "targets": ["rocm-6-4"],
             "head_sha": "${{ needs.pre-dispatch.outputs.head_sha }}",
             "merge_commit_sha": "${{ needs.pre-dispatch.outputs.merge_commit_sha }}",
             "url": "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}",
--- a/.pfnci/matrix.yaml
+++ b/.pfnci/matrix.yaml
@@ -619,6 +619,31 @@
   env:CUPY_ACCELERATORS: null
   test: "unit"
 
+  # ROCm 6.4 | Linux
+  # The latest ROCm version matrix is intended to cover the highest supported combination.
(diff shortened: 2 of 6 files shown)


.pfnci/matrix.yaml
```diff
@@ -619,6 +619,31 @@
   env:CUPY_ACCELERATORS: null
   test: "unit"
 
+  # ROCm 6.4 | Linux
+  # The latest ROCm version matrix is intended to cover the highest supported combination.
```
kmaehashi:
```suggestion
# ROCm 6.4 | Linux
# The latest ROCm version matrix is intended to cover the highest supported combination.
```

kmaehashi:
Thanks! Could you update [this line](https://github.com/cupy/cupy/blob/67c5887c914029db3a0217657db0b6451def3c0c/.github/workflows/self-hosted-ci.yml#L95) as well to let `/test rocm` comment trigger `rocm-6-4`?

.pfnci/matrix.yaml
gpinkert: fixed

gpinkert:
> Thanks! Could you update [this line](https://github.com/cupy/cupy/blob/67c5887c914029db3a0217657db0b6451def3c0c/.github/workflows/self-hosted-ci.yml#L95) as well to let `/test rocm` comment trigger `rocm-6-4`?

fixed

kmaehashi:
`/test` commands requires reviewer privilege in CuPy. Let me kick.

.github/workflows/self-hosted-ci.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         token: ${{ steps.generate-token.outputs.token }}
         client-payload: |
           {
-            "targets": ["rocm-6-2"],
+            "targets": ["rocm-6-4"],
```
kmaehashi, line 95: /test rocm

kmaehashi:
Ah, this can't be triggered in PR as the workflow uses secrets. Let me merge this and check if it works as expected afterwards.

kmaehashi:
https://github.com/cupy/self-hosted-ci/actions/runs/17756028712

PR #9377, opened 2025-09-11
Build against `cuda-bindings` 12.x

leofang:

PR #9381, opened 2025-09-13
feat(tests): fix `mvdr` failure on float32

gpinkert:
The test was comparing results from `mvdr` against `[-2, 1.5]` with an extremely tight tolerance (`rtol=1e-7, atol=0`). On **float32** this is stricter than the natural rounding error of the type, so values like `[-1.9999995, 1.4999998]` (mathematically equal for single precision) caused a failure.

This change makes the test match the precision of the data type:

* Build the expected array in the **same dtype** as the output.
* Use **dtype-aware tolerances** (looser for float32, tighter for float64).

No algorithmic changes—only the assertion threshold. This keeps the test meaningful while avoiding false negatives due to normal single-precision rounding.

Example before vs after (float32):

* before: expected `[-2, 1.5]` vs actual `[-1.9999995, 1.4999998]`
* after: same values within float32-appropriate tolerance

This should also reduce flakiness across different GPUs/compilers.

### DIFF
--- a/tests/cupyx_tests/signal_tests/radartools_tests/test_radartools.py
+++ b/tests/cupyx_tests/signal_tests/radartools_tests/test_radartools.py
@@ -158,4 +158,8 @@ def test_mvdr(dtype):
     sv = cupy.array([1, 2], dtype=dtype)
     out = signal.mvdr(x, sv)
     assert out.dtype == dtype
-    testing.assert_allclose(out, [-2, 1.5])
+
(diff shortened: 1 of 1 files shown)


gpinkert:
@kmaehashi the build failed, unrelated to this PR. Should I just retest?

kmaehashi:
Merging as the failure is unrelated to the PR. `cuda12x-cuda-python` is a new test matrix introduced in #9377, thus timeout is expected for several runs, until the kernel cache grows.

PR #9390, opened 2025-09-15
Add function bitwise_count

0xtoward:
#6078
- Adds GPU ufunc cupy.bitwise_count and exposes cupy.bit_count as an alias for old version of Numpy.
- return uint8

### DIFF
--- a/tests/cupy_tests/binary_tests/test_elementwise.py
+++ b/tests/cupy_tests/binary_tests/test_elementwise.py
@@ -29,6 +29,7 @@ class TestElementwise(unittest.TestCase):
     def test_bitwise_xor(self):
         self.check_binary_int('bitwise_xor')
 
+    @testing.with_requires('numpy>=2.0.0')
     @testing.for_int_dtypes(no_bool=True)
     @testing.numpy_cupy_array_equal()
     def test_bitwise_count(self, xp, dtype):


cupy/__init__.py
```diff
@@ -307,6 +307,8 @@ def is_available():
 from cupy._binary.elementwise import bitwise_or  # NOQA
 from cupy._binary.elementwise import bitwise_xor  # NOQA
 from cupy._binary.elementwise import bitwise_not  # NOQA
+from cupy._binary.elementwise import bitwise_count  # NOQA
+from cupy._binary.elementwise import bit_count  # NOQA
```
kmaehashi: As far as I'm aware `np.bit_count` has never existed in NumPy.

tests/cupy_tests/binary_tests/test_elementwise.py
```diff
@@ -29,6 +29,17 @@ def test_bitwise_or(self):
     def test_bitwise_xor(self):
         self.check_binary_int('bitwise_xor')
 
+    def test_bitwise_count(self):
```
kmaehashi: Use `check_unary_int`, and please also test boundaries e.g. INT_MIN.

kmaehashi (changes requested):

cupy/_core/_routines_binary.pyx
```diff
@@ -64,6 +64,37 @@ cdef _invert = create_ufunc(
...
+    doc='''Computes the number of 1-bits in each element.
+
+    Only integer arrays are handled.
+
+    .. seealso:: :data:`numpy.bit_count`
+
+    ''',
+    preamble='''
+template <typename T>
+__device__ inline unsigned int _cupy_bitcount(T x) {
```
leofang: It seems this incurs extra type casts which is unnecessary?
- popc returns `int` (or int32)
- This cast int32 to uint32
- The ufunc casts uint32 to uint8 (as required by NumPy)

```suggestion
__device__ inline int _cupy_bitcount(T x) {
```

leofang:
Could you fix the linter errors?

cupy/_core/_routines_binary.pyx
```diff
@@ -64,6 +64,37 @@ cdef _invert = create_ufunc(
...
+
+    ''',
+    preamble='''
+template <typename T>
+__device__ inline int _cupy_bitcount(T x) {
+    if (sizeof(T) <= 4) {
+    unsigned int ux = static_cast<unsigned int>(x);
+    if ((T)(-1) < (T)0 && x < 0) {
+      ux = ~ux + 1u;
+    }
```
leofang: It seems the type check can be done at compile time instead of run time. Could you use `[cuda::]std::is_signed`? Something like
```c++
if constexpr (std::is_signed_v<T>) {
  if (x < 0) {
    ux = ~ux + 1u;
  }
}
```
Need to add `is_signed_v` to https://github.com/cupy/cupy/blob/main/cupy/_core/include/cupy/cuda_workaround.h and include the header.

0xtoward: Hi! Thanks for your review. I just fix the linters and simply changed `if (T)(-1) < (T)0` with `constexpr (std::is_signed_v<T>) `      : )

leofang: Looks like I mistakenly suggested using `is_signed_v`, we should probably just use `is_signed` in case the C++ standard is not new enough. But @0xtoward you should have caught my mistake if you had run tests locally 😅

0xtoward: > Looks like I mistakenly suggested using `is_signed_v`, we should probably just use `is_signed` in case the C++ standard is not new enough. But @0xtoward you should have caught my mistake if you had run tests locally 😅

Nah I guess I figured out my local error (by recompiling and bump build_num maybe). I did pass the unit test, and thought it was what it should be, lol!  Got it, there will be a quick fix

tests/cupy_tests/binary_tests/test_elementwise.py
```diff
@@ -29,6 +29,23 @@ def test_bitwise_or(self):
...
+        info = np.iinfo(dtype)
+        if np.issubdtype(dtype, np.signedinteger):
+            obj = np.array([
+                0, -1, 1, info.min, info.min + 1,
+                info.max, info.max - 1, info.max // 2,
+            ], dtype=object)
+        else:
+            obj = np.array([
+                0, 1, info.max, info.max - 1, info.max // 2,
+            ], dtype=object)
```
leofang: Q: Do we have to use `dtype=object` here? I think NumPy array constructor would take care of all conversions properly. There are no values out of bound, anyway.

0xtoward: > Q: Do we have to use `dtype=object` here? I think NumPy array constructor would take care of all conversions properly. There are no values out of bound, anyway.

as KM mentioned:

>Use check_unary_int, and please also test boundaries e.g. INT_MIN. 

so I added the corner case tests. Will our kernel code run out of boundaries? 

Btw I didn't step into "check_unary_int" and maybe I should add it?

Thanks for ur guide and sorry for my poor engineering experience.  I just messed up my local env and took me much more time than I thought to fix.

leofang: You can check locally but I believe changing `dtype=object` to `dtype=dtype` is better. Both should generate the same input after the `asarray` call below.

Asking because it is quite confusing to use `dtype=object`. Unlike NumPy, CuPy cannot support non-numerical types.

0xtoward: Thanks for your carefull review! I just replace `object `with `dtype `and add `check_unary_int`.

0xtoward: Hi @leofang  
I’ve pushed updates and local CI is passing. Could you rereview it when convenient? Thanks very much!

tests/cupy_tests/binary_tests/test_elementwise.py
```diff
@@ -29,6 +29,24 @@ def test_bitwise_or(self):
     def test_bitwise_xor(self):
         self.check_binary_int('bitwise_xor')
 
+    @testing.for_int_dtypes(no_bool=True)
+    @testing.numpy_cupy_array_equal()
+    def test_bitwise_count(self, xp, dtype):
+        import numpy as np
```
leofang: We already have `xp` supplied by the `numpy_cupy_array_equal` decorator which compares np vs cp 
```suggestion
```

tests/cupy_tests/binary_tests/test_elementwise.py
```diff
@@ -29,6 +29,24 @@ def test_bitwise_or(self):
...
+        info = np.iinfo(dtype)
```
leofang: ditto
```suggestion
        info = xp.iinfo(dtype)
```

tests/cupy_tests/binary_tests/test_elementwise.py
```diff
@@ -29,6 +29,24 @@ def test_bitwise_or(self):
...
+            obj = np.array([
+                0, -1, 1, info.min, info.min + 1,
+                info.max, info.max - 1, info.max // 2,
+            ], dtype=dtype)
+        else:
+            obj = np.array([
+                0, 1, info.max, info.max - 1, info.max // 2,
+            ], dtype=dtype)
+        a = xp.asarray(obj, dtype=dtype)
+        self.check_unary_int('bitwise_count')
```
leofang: @kmaehashi can confirm but I don't think this is needed
```suggestion
```

tests/cupy_tests/binary_tests/test_elementwise.py
```diff
@@ -29,6 +29,24 @@ def test_bitwise_or(self):
...
+        if np.issubdtype(dtype, np.signedinteger):
+            obj = np.array([
+                0, -1, 1, info.min, info.min + 1,
+                info.max, info.max - 1, info.max // 2,
+            ], dtype=dtype)
+        else:
+            obj = np.array([
+                0, 1, info.max, info.max - 1, info.max // 2,
+            ], dtype=dtype)
+        a = xp.asarray(obj, dtype=dtype)
```
leofang: simplify:
```suggestion
        if xp.issubdtype(dtype, xp.signedinteger):
            a = xp.asarray([
                0, -1, 1, info.min, info.min + 1,
                info.max, info.max - 1, info.max // 2,
            ], dtype=dtype)
        else:
            a = xp.asarray([
                0, 1, info.max, info.max - 1, info.max // 2,
            ], dtype=dtype)
```

leofang:
Could you check the CI failures? I think we need to constrain the NumPy version using `@testing.with_requires(...)`, since this API is new in NumPy.

leofang:
Sorry, we only allow maintainers to trigger the CI pipelines in this repo 😅

asi1024:
LGTM! Thanks for your contribution!

PR #9395, opened 2025-09-16
Fix tf2sos failing for constant transfer function.

### LINKED ISSUE #9376: tf2sos fails when given constant transfer function.
### Description

I found this when working on testing against alternative backends in the `scipy.signal` test suite. `tf2sos` fails in the following case:

```python
import cupy
from cupyx.scipy.signal import tf2sos

# Numerator and denominator of transfer function are just the constant 1.
b, a = cupy.array([1.0]), cupy.array([1.0])
tf2sos(b, a)
```

with error

```
```

What's going on is that `tf2sos` simply does `zpk2sos(*tf2zpk(b, a))` and `tf2zpk` in this case returns a tuple

```
(array([], dtype=float64), array([], dtype=float64), array(1.))
```

with a 0d array as the last element.

When fed to `zpk2sos` this hits the code block:

https://github.com/cupy/cupy/blob/25e552d5d679dcdc6f7cc3d81310a9b265463137/cupyx/scipy/signal/_iir_filter_conversions.py#L182-L186

and the failure occurs due to a 0d array being included in an element of the list passed to `cupy.array`. One possible fix would be to instead write this as

```python
    if len(z) == len(p) == 0:
        if not analog:
            return cupy.array([[k.get(), 0., 0., 1., 0., 0.]])
        else:
            return cupy.array([[0., 0., k.get(), 0., 0., 1.]])

```

though their may be a way that's more idiomatic to the CuPy code base.

### To Reproduce

```py
import cupy
from cupyx.scipy.signal import tf2sos

b, a = cupy.array([1.0]), cupy.array([1.0])
sos = tf2sos(b, a)
```

### Installation

### Environment

```
# Paste the output here
OS                           : Linux-6.8.0-79-generic-x86_64-with-glibc2.39
CuPy Version                 : 13.6.0
NumPy Version                : 2.3.2
Cython Runtime Version       : 3.1.3
CUDA Root                    : /usr/local/cuda-12.4
nvcc PATH                    : /usr/local/cuda-12.4/bin/nvcc
CUDA Runtime Version         : 12090 (linked to CuPy) / 12040 (locally installed)
cuBLAS Version               : 120405
cuFFT Version                : 11201
cuRAND Version               : 10305
cuSOLVER Version             : (11, 6, 1)
cuSPARSE Version             : 12301
NVRTC Version                : (12, 4)
NCCL Build Version           : 22501
NCCL Runtime Version         : 22703
Device 0 Name                : NVIDIA RTX A4000
```

### Additional Information

_No response_

steppi:
Closes #9376

This PR makes `tf2sos` work when a constant transfer function is passed. The problem occurred due to array construction in `zpk2sos` which assumes that the system gain `k` is a float and fails when `k` is a 0d CuPy array. To fix this, I added a check for if `k` is a 0d CuPy array, and call `k.get()` to convert to a 0d NumPy array, which does work for such array construction. I made it check specifically for a 0d CuPy array to try to make the error message less confusing if a higher dimensional CuPy array is passed for `k`, e.g. to avoid an error message referring to NumPy arrays.

I've also copied the tests for `tf2sos` over from SciPy, including the ones that failed due to #9376.


leofang (approved):
Sorry for delay. LGTM but I'd like to have @asi1024's blessing, since this corner case causes a host-device synchronization.

cupyx/scipy/signal/_iir_filter_conversions.py
```diff
@@ -182,6 +182,10 @@ def zpk2sos(z, p, k, pairing=None, *, analog=False):
                          'pairing must be "minimal"')
 
     if len(z) == len(p) == 0:
+        if isinstance(k, cupy.ndarray) and k.ndim == 0:
+            # check ndim == 0 to make error message simpler in
+            # case where a non-0d CuPy array is mistakenly passed.
+            k = k.get()
```
asi1024: We can avoid DtoH synchronization with the following modifications.

```suggestion
            out = cupy.array([[0, 0., 0., 1., 0., 0.]])
            index = 2 if analog else 0
            out[index] = k
```

leofang: What's `out`?

steppi: Thanks @asi1024. I had this same thought when I saw the comment about triggering an unnecessary device synchronization. It should be `out[0][index] = k` since out has shape `(1, 6)`.  I'll make the fix here. @leofang, `out` is just what gets returned in the `if len(z) == len(p) == 0:` branch.

steppi:
```python
    if len(z) == len(p) == 0:
        if not analog:
            out = cupy.array([[0., 0., 0., 1., 0., 0.]])
            out[0][0] = k
            return out
        out = cupy.array([[0., 0., 0., 0., 0., 1.]])
        out[0][2] = k
        return out
```

This works.

asi1024 (approved):

asi1024:
LGTM! Thanks!

PR #9399, opened 2025-09-17
Fix cuTENSOR workspace size query

### LINKED ISSUE #9350: Fix cuTENSOR workspace size query
Based on the investigation raised by the cuTENSOR team:

> we are not using the cuTENSOR API properly in cupy (a bug in cuTENSOR 2.3.0 which was released a couple of weeks ago is likely going to exacerbate the situation, but luckily cuTENSOR 2.3.1 (patch) is almost out to fix the issue).
>
> The issue is that cupy uses the estimated workspace size ([here](https://github.com/cupy/cupy/blob/72ea1f14cd454f7f2aa22a0866a6bafeed632532/cupyx/cutensor.pyx#L752)) to determine how much workspace to allocate. The estimated workspace can be much larger than the actual workspace size required (which can be queried using cutensorPlanGetAttribute after planning, see example [here](https://github.com/NVIDIA/CUDALibrarySamples/blob/a2433ef214c238e0abdb5e4307ba4b1862f81d1a/cuTENSOR/contraction.cu#L286))
>
> (the bug in 2.3.0 grossly over-estimated the workspace size required for certain contractions)

@yangcal will take this up once he's freed up.

yangcal:
closes #9350

This MR updates cutensor contraction backend with improvement on the workspace size inquiry. 

An additional call to `cutensorPlanGetAttribute` is made to reduce the workspace size actually needed to execute the contraction. A few extra bindings were added to support this improvement. Example usage from [here](https://github.com/NVIDIA/CUDALibrarySamples/blob/a2433ef214c238e0abdb5e4307ba4b1862f81d1a/cuTENSOR/contraction.cu#L286)

### DIFF
--- a/cupy_backends/cuda/libs/cutensor.pxd
+++ b/cupy_backends/cuda/libs/cutensor.pxd
@@ -184,7 +184,7 @@ cpdef intptr_t createPlan(
     intptr_t pref,
     uint64_t workspaceSizeLimit) except? 0
 
-cpdef getPlanAttribute(
+cpdef planGetAttribute(
     intptr_t handle,
     intptr_t plan,
     int attr,
--- a/cupy_backends/cuda/libs/cutensor.pxd
+++ b/cupy_backends/cuda/libs/cutensor.pxd
@@ -48,7 +48,7 @@ cpdef enum:
     WORKSPACE_MAX = 3          # NOQA, All algorithms will be available
 
     # cutensorPlanAttribute_t
-    PLAN_REQUIRED_WORKSPACE = 0 # NOQA, The required workspace size for the plan
+    PLAN_REQUIRED_WORKSPACE = 0  # NOQA, The required workspace size for the plan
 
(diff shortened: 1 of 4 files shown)


leofang:
Thanks, @yangcal! I think we also need to add a stub to here: 
https://github.com/yangcal/cupy/blob/cutensor-workspace-fix/cupy_backends/stub/cupy_cutensor.h

yangcal:
> Thanks, @yangcal! I think we also need to add a stub to here: https://github.com/yangcal/cupy/blob/cutensor-workspace-fix/cupy_backends/stub/cupy_cutensor.h

yangcal:
@leofang Thanks, stubs are added in a762049d09c6e6842b1aef640462a447727ae394

leofang:
Need to fix linter failures: `pre-commit run -a`

kmaehashi (approved):

PR #9411, opened 2025-10-06
[nccl] add nccl comm split

### LINKED ISSUE #9096: ncclCommSplit exposed in nccl communicator
### Description

Hi guys, is there any plans for exposing ncclCommSplit function for the cupy users?
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/communicators.html#creating-more-communicators
I'd be happy if it was possible as I need to create a clique from existing comms without the need to create whole new communicators with separate comm_id and extra memory usage for maintaining separate set of temp buffers.

### Additional Information

_No response_

tianyi-ge:
# Target
This PR aims to add commSplit support to cupy and close #9096 

# Implementation:
Given color, key and optional nccl_config, a new communicator will be created based on an old larger one. It's only supported if nccl version >= 2.18.1

```python
from cupy.cuda import nccl
comm = nccl.NcclCommunicator(world_size, uid, rank)
new_comm = comm.commSplit(color, key)
if new_comm is not None:
    # use new_comm for collective communication
    new_comm.destroy()
comm.destroy()
```


cupy_backends/cuda/libs/nccl.pyx
```diff
@@ -492,3 +498,53 @@ cdef class NcclCommunicator:
...
+                    new_comm.destroy()
+                comm.destroy()
+
+        .. seealso:: `ncclCommSplit`_
+
+        .. _ncclCommSplit:
+            https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcommsplit
+        """
+        if NCCL_VERSION_CODE < 2700:
+            raise RuntimeError('ncclCommSplit is not available in this version')
```
seberg: Thanks, but I think there is still a bit of work to do here.

This shouldn't actually work, because you need the corresponding header changes or it cannot work on older nccl versions (please search the code for this).  It also looks incorrect, since it was added in `2.18.1`.  Also, to make that work, you do need to modify all the related headers.

Additionally, the way `NCCL_SPLIT_NOCOLOR` is used doesn't look right to me.  Considering that, unless it is very involved it would be good to add a test and ideally do a local test run for the nccl tests.

tianyi-ge: Thanks for your suggestions. I have a few questions to make things work.

> Also, to make that work, you do need to modify all the related headers.

May I ask what the "header changes" you mentioned is? Is it the way to calculate the version code? Do you expect to adapt this api to older version?

> Additionally, the way NCCL_SPLIT_NOCOLOR is used doesn't look right to me.

In nccl code, `NCCL_SPLIT_NOCOLOR` seems to indicate that this rank is not involved in this split operation. So all the ranks have to execute ncclCommSplit. I'll change this

cupy_backends/cuda/libs/nccl.pyx
```diff
@@ -72,6 +73,18 @@ cdef extern from '../../cupy_nccl.h':
...
+        ctypedef struct ncclConfig_t:
+            # expose more fields when necessary
+            int splitShare
+
+        ncclResult_t ncclCommInitRankConfig(ncclComm_t* comm, int ndev,
+                                            ncclUniqueId commId, int rank,
+                                            ncclConfig_t* config) nogil
+        ncclResult_t ncclCommSplit(ncclComm_t comm, int color, int key,
+                                   ncclComm_t* newcomm,
+                                   ncclConfig_t* config) nogil
```
seberg: If you search the code base, you find that these already exist in `cupy_backends/cupy_nccl.h`.  I suppose this may work, but I think it is better to move the `#IF` to there

cupy_backends/cuda/libs/nccl.pyx
```diff
@@ -273,13 +319,28 @@ cdef class NcclCommunicator:
...
         for i in range(NCCL_UNIQUE_ID_BYTES):
             _uniqueId.internal[i] = commId[i]
-        with nogil:
-            status = ncclCommInitRank(&self._comm, ndev, _uniqueId, rank)
+
+        if NCCL_VERSION_CODE < 21801:
+            if config is not None:
+                warnings.warn(
+                    'NcclConfig is ignored in this version of NCCL',
+                    RuntimeWarning)
```
seberg: This can just raise an.  In fact, the `NcclConfig` constructor should already raise.
(but a sanity check before calling `ncclCommInitRankConfig` makes sense).

cupy_backends/cuda/libs/nccl.pyx
```diff
@@ -243,6 +256,32 @@ cpdef groupEnd():
...
+
+        cdef:
+            ncclConfig_t _config
+
+        def __cinit__(self):
+            self._config.splitShare = 0
+
+        property split_share:
+            def __get__(self): return self._config.splitShare
+            def __set__(self, int v): self._config.splitShare = v
```
seberg: I doubt `NcclConfig` should be a mutable object.  If anything it should have a large `__cinit__` that sets the right things.  It can then have properties to query it.

cupy_backends/cuda/libs/nccl.pyx
tianyi-ge: I see. `cupy_backends/cupy_nccl.h` provides dummy structs and functions for older nccl to pass the compilation!

tianyi-ge:
Hi @seberg thanks for the suggestions. I fixed them and by the way, updated the way to calculate NCCL_VERSION. Could you review this again at your convenience

cupy_backends/cuda/libs/nccl.pyx
```diff
@@ -492,3 +545,58 @@ cdef class NcclCommunicator:
...
+                assigned to the same communicator. If color is ``-1``,
+                the process is not included in any communicator.
+            key (int): Controls the rank assignment within
+                the new communicator. The process with the lowest key
+                value is assigned rank 0.
+            config (NcclConfig): Configuration for communicator creation.
+                NULL by default.
+
+        Returns:
+            NcclCommunicator: A new communicator.
```
seberg: It can also return `None` if color is `-1`.

tests/cupy_tests/cuda_tests/test_nccl.py
```diff
@@ -64,6 +64,18 @@ def test_comm_size(self):
         comm = nccl.NcclCommunicator(1, id, 0)
         assert 1 == comm.size()
 
+    @unittest.skipUnless(nccl_version >= 21801, 'Using old NCCL')
+    def test_comm_split(self):
+        id = nccl.get_unique_id()
+        config = nccl.NcclConfig()
+        config.split_share = 1
```
seberg: We removed the setter, so this test can't pass.  It would be nice to add a small test just to check `NcclConfig` and the `split_share` getter.

For here, it may also be nice to test both config and no-config (I don't think I care if you have the identical things).  You could use `@pytest.mark.parametrize("config", [..., None])`.
(It _might_ be nice, to explicitly just test `nccl.NcclCommunicator(1, id, 0, config)`, then here it doesn't even matter if you pass config or not.)

tianyi-ge: Thanks. I added parametrize to simplify comm split tests. I also added split_no_color case and config test.

I agree with your idea to extract a new python binding to nccl. but so far it seems ok for cupy to provide this

tianyi-ge:
I fixed the compilation for nccl==2.17, which provides `ncclConfig_t` but does not support `splitShare` (since nccl 2.18). I cannot intrusively modify a nccl native struct. So I created a fake `_ncclConfig_t` with `splitShare` for 2.17 or earlier, and `typedef ncclConfig_t _ncclConfig_t` for 2.18+

It allows nccl==2.17 to have a `splitShare` field and pass the compilation, even though it will never be used (stopped by version check). If in the future more fields of `ncclConfig_t` need to be exposed for nccl==2.17, just add them in `_ncclConfig_t`.

please take a look and perhaps run ci tests again, thanks @seberg

seberg:
OK, let's see if this fixed all failures (I hope so), thanks.  A bit annoying about needing the `_` version, I would be almost tempted to just ignore that precise versioning, but now we have it.

/test mini

tianyi-ge:
`ncclConfig_t` needs to be intialized via `NCCL_CONFIG_INITIALIZER`, otherwise config magic number check will fail https://github.com/NVIDIA/nccl/blob/834ef7231913ecf22f5cad29d7e26a6596f36452/src/init.cc#L1748

I'll fix this and test it locally

tianyi-ge:
Tested on 2.27.3
```
pytest -s test_nccl.py::TestCommSplit
================================================ test session starts =================================================
platform linux -- Python 3.12.3, pytest-8.1.1, pluggy-1.6.0
rootdir: /data2/g00608016/cupy
configfile: pyproject.toml
plugins: anyio-4.9.0, shard-0.1.2, rerunfailures-15.1, hypothesis-6.130.8, xdist-3.7.0, flakefinder-1.1.0, xdoctest-1.0.2
collecting ... 
collected 4 items                                                                                                    
Running 4 items in this shard

test_nccl.py ....

============================================ 4 passed in 70.99s (0:01:10) ============================================```

**Comment by tianyi-ge** [comment]
Tested a real comm split + allreduce for 4 processes and passed
Is it appropriate to add this to `test_nccl.py`?

```python
class TestCommSplit:
    @staticmethod
    def _worker(id, rank, world_size):
        cuda.Device(rank).use()
        global_comm = nccl.NcclCommunicator(
            world_size, id, rank, nccl.NcclConfig(split_share=1)
        )
        color = 0 if (rank < world_size // 2) else 1
        sub_comm = global_comm.commSplit(color, rank)
        assert sub_comm is not None
        send_buf = cupy.array([rank], dtype=cupy.int8)
        recv_buf = cupy.empty_like(send_buf)
        sub_comm.allReduce(
            send_buf.data.ptr, recv_buf.data.ptr, send_buf.size,
            nccl.NCCL_INT8, nccl.NCCL_SUM, cuda.Stream.null.ptr,
        )
        cuda.Stream.null.synchronize()
        result = sum(range(world_size // 2)) if color == 0 else sum(range(world_size // 2, world_size))
        expected = cupy.array([result], dtype=cupy.int8)
        testing.assert_array_equal(recv_buf, expected)
        sub_comm.destroy()
        global_comm.destroy()

    @testing.multi_gpu(4)
    def test_4_processes_2_subcomm(self):
        import multiprocessing as mp
        mp.set_start_method("spawn", force=True)
        world_size = 4
        id = nccl.get_unique_id()
        procs = [mp.Process(target=TestCommSplit._worker, args=(id, i, world_size,)) for i in range(world_size)]
        for p in procs:
            p.start()
        for p in procs:
            p.join()
```

seberg:
I am assuming that is similar to existing tests there (in both style and runtime).  In which case, I would say: yes, no harm in adding!

tianyi-ge:
tested both on nccl==2.7 and 2.27, to make sure it's compatible. PTAL thanks!

cupy_backends/cupy_nccl.h
```diff
@@ -166,4 +172,44 @@ ncclResult_t ncclRecv(void* recvbuff, size_t count, ncclDataType_t datatype,
...
+    return ncclSuccess;
+}
+
+#else // #if (NCCL_VERSION_CODE >= NCCL_VERSION(2, 18, 0))
+// reuse the original struct
+typedef ncclConfig_t _ncclConfig_t;
+
+static inline _ncclConfig_t ncclConfigInitializer() {
+    return (_ncclConfig_t)NCCL_CONFIG_INITIALIZER;
+}
```
seberg: Could probably just do:
```
#ifndef NCCL_CONFIG_INITIALIZER
#define NCCL_CONFIG_INITIALIZER _ncclConfig_t{0}
#endif
```
I suspect (the `#ifdef` since IIRC the config addition doesn't align with 2.18).  But maybe not worth re-running CI over.

seberg, line 202: Could probably just do:
```
#ifndef NCCL_CONFIG_INITIALIZER
#define NCCL_CONFIG_INITIALIZER _ncclConfig_t{0}
#endif
```
I suspect (the `#ifdef` since IIRC the config addition doesn't align with 2.18).  But maybe not worth re-running CI over.

Anyway, don't want to keep iterating for such a small improvement.  So will have another brief look, probably run tests tomorrow and then this should be good.

seberg:
OK, let's see if tests pass, and then it's probably good.  Sorry, got lost in some other things.

/test mini

tianyi-ge:
Hi @kmaehashi could you take a look and merge this?

seberg:
Thanks @tianyi-ge, especially for the nice cleanups and tests!

I decided I don't want to bike-shed API, since we may start suggesting to move to nccl4py soon enough anyway.
But if anyone feels we should tweak this, I'll be happy to follow up.

Issue #9128, opened 2025-05-11
Support Cython 3.1


kmaehashi:
It seems we are getting this error:

```
    File "//setup_cuda_runtime_headers.py", line 5, in <module>
      import cupy
    File "/tmp/.local/lib/python3.12/site-packages/cupy/__init__.py", line 16, in <module>
      from cupy import _core  # NOQA
      ^^^^^^^^^^^^^^^^^^^^^^
    File "/tmp/.local/lib/python3.12/site-packages/cupy/_core/__init__.py", line 3, in <module>
      from cupy._core import core  # NOQA
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "cupy/_core/core.pyx", line 1, in init cupy._core.core
    File "/tmp/.local/lib/python3.12/site-packages/cupy/cuda/__init__.py", line 9, in <module>
      from cupy.cuda import compiler  # NOQA
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/tmp/.local/lib/python3.12/site-packages/cupy/cuda/compiler.py", line 14, in <module>
      from cupy.cuda import device
    File "cupy/cuda/device.pyx", line 105, in init cupy.cuda.device
    File "cupy/_util.pyx", line 52, in cupy._util.memoize.decorator
    File "/opt/pyenv/versions/3.12.0/lib/python3.12/functools.py", line 56, in update_wrapper
      setattr(wrapper, attr, value)
  AttributeError: attribute '__name__' of 'builtin_function_or_method' objects is not writable
```

## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [x] Read [ChangeLog](http://docs.cython.org/en/latest/src/changes.html).
- [x] Fix code and CI to support the new version.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - **Package Metadata**: Only when needed: change the minimum Cython version in `setup.py`.
- [x] Backport the above PR.
- [ ] Fix `cupy-release-tools` to support the new version.
- [ ] Backport the above PR.
- [x] Fix documentation.
  - Update Requirements section in the [Installation Guide](https://docs.cupy.dev/en/latest/install.html)
- [x] Backport the above PR.

kmaehashi:
[...]

PR #9414, opened 2025-10-07
Replace fastrlock with C++ recursive_mutex 

seberg:
It seems that fastrlock tends to be a problem with new Python versions
and also freethreading.
This proposes to replace it with a C++ mutex, making it recursive/re-entrant
just because I am not 100% sure about things.

I am not sure if there are very subtle issues (subinterpreters?)
if using a C++ mutex, but I don't think those are of a big concern.
FWIW, NumPy also uses a lot of C++ mutexes for free-threading support at least.

Since Python 3.10 has direct GC C-API functions in the stable API just added
those, I didn't time if it is relevant, but it seemed just as well/nice.

---

Tested the performance with the test from: https://github.com/cupy/cupy/pull/9149#issuecomment-2993562268 and saw no difference (beyond just random fluctuations).
I can see that `fastrlock` may have some advantages, but just avoiding the dependency is maybe nice neough.


kmaehashi:
Wow this is great, thanks @seberg!

Let me trigger CI, especially I'm curious if there's any roadblocks on Windows

/test mini

seberg:
Hmmm, seems Cython 3.0.x didn't include the `recursive_mutex` defines (and I am not quite sure if non-recursive is enough).
Do you have a clear preference: Include the defines (with a comment) or be a bit aggressive and require Cython 3.1 (and try to remove the upper limit soon)?

kmaehashi:
It sounds good to me to require Cython 3.1 as it is build-time-only requirement. We deferred that decision because there were bug fix release planned at the time we added support for it (https://github.com/cupy/cupy/issues/9128#issuecomment-2890110649), but I think 3.1.x series today are enough stable.

seberg:
Think it should be good now and also just to try triggering CI.  (I do suspect we don't need a recursive mutex but seemed like a good step to do like-for-like first.)

/test mini

seberg:
And once more:

/test mini

cupy/cuda/memory.pyx
```diff
@@ -1021,53 +1028,75 @@ cpdef size_t _bin_index_from_size(size_t size):
...
+cdef lock_and_no_gc(recursive_mutex& lock):
+    cdef _LockAndNoGc self = _LockAndNoGc.__new__(_LockAndNoGc)
+    self._lock = &lock
+    return self
+
+
+def _test_lock_and_no_gc():
+    # Test function defined here as it requires the C++ mutex
+    # unfortunately we can only test the GC part of the manager
+    # easily because C++ locks can't introspected.  We would need
```
bdice:
```suggestion
    # easily because C++ locks can't be introspected.  We would need
```

cupy/cuda/memory.pyx
```diff
@@ -1021,53 +1028,75 @@ cpdef size_t _bin_index_from_size(size_t size):
...
+cdef bint _lock_no_gc(recursive_mutex& lock):
     """Lock to ensure single thread execution and no garbage collection.
 
     Returns:
         bool: Whether GC is disabled.
     """
-    rlock.lock_fastrlock(lock, -1, True)
+    if not lock.try_lock():
+        with nogil:
+            lock.lock()
```
kmaehashi, line 1039: I suspect this is unsafe (until we make the memory pool free-threaded-safe), because this will give other threads a chance to call memory pool APIs.

seberg: @kmaehashi are you worried that this is not a good replacement?  It should be a direct translation I think (and free-threading safe, but agree we probably need to fixup up/audit the logic).

The thing to keep in mind is that we have two mutexes `lock` and the GIL. And the important part with two mutexes is that we always lock them in the same order.

The dance here ensures that: we always lock `lock` before the GIL. I.e. every blocking `lock.lock()` call is inside a `with nogil:` section guaranteeing that the order is always `lock.lock()` and then grabbing the GIL (leaving `with nogil`).

If `try_lock()` succeeds (which seems likely) with the GIL already held, that is OK since the order only matters to avoid dead-locking which clearly can't occur if we are not blocking).

EDIT: And since we (effectively) always lock `lock` first, it's OK that we may be unlocking the GIL later in the function (indirectly).

kmaehashi: Sorry, I wasn't aware that fastrlock frees GIL when locking. I agree this is indeed a direct translation.

I still suspect there is a potential deadlock ~here~, since we use `gc.collect()` which may trigger [`cudaFree` that also frees/re-acquires GIL](https://github.com/cupy/cupy/blob/b82e6d361fcb64d6bd914f8e0650b018767081eb/cupy_backends/cuda/api/runtime.pyx#L598-L600). The example scenario is: 

|Time|Thread 1|Thread 2|
|--|--|--|
|1|User code calls malloc [Lock GIL]||
|2|(_malloc)||
|3|(_try_malloc)||
|4|Lock _total_bytes_lock||
|5|(gc.collect)||
|6|(cudaFree) [Unlock GIL]||
|7||User code calls free_all_blocks [Lock GIL]|
|8||Lock _free_lock|
|9||(_compact_index)|
|10||Try-Lock _total_bytes_lock (fail)|
|11||[Unlock GIL]|
|12||Lock _total_bytes_lock (deadlock)|
|13|[Lock GIL]||
|14|Try-Lock _free_lock (fail)||
|15|[Unlock GIL]||
|16|Lock _free_lock (deadlock)|

However I think it's ok to proceed merging with this PR because this is not a new issue in this PR, have never reported from users, and will be fixed once we support free-threading support.

PR #9415, opened 2025-10-08
Improve CI docs

kmaehashi:
Improving CI docs now that it is more likely that many contributors modify CI matrix.

### DIFF
--- a/.pfnci/README.md
+++ b/.pfnci/README.md
@@ -1,6 +1,6 @@
 # CuPy CI
 
-CuPy uses two infrastructures for GPU tests.
+CuPy uses the following two infrastructures for GPU tests.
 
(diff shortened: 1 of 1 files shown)


.pfnci/README.md
```diff
@@ -38,6 +39,17 @@ pip install PyYAML
 ./generate.py
 ```
 
+### Timeouts
+
+CuPy CI has a two layers of timeouts:
+
+* Timeout set by CI system
+  * Configured by `time_limit` in `config.pbtxt` (for PFN CI) and `timeout-minutes` in workflow files (for GitHub Actions workflows including self-hosted CIs).
```
seberg, line 47: I guess timeout-minutes is just the default right now.

Issue #9089, opened 2025-04-04
Can't use `cupy.ndarray`'s `__setitem__` when the `value` is an array-like object that implements `__cuda_array_interface__` and `__cupy_get_ndarray__` but also implements `__eq__`.


ikrommyd:
### Description

Can't use `cupy.ndarray`'s `__setitem__` when the `value` is an array-like object that implements `__cuda_array_interface__` and `__cupy_get_ndarray__` but also implements `__eq__`.

The reproduce below works with `numpy`
```py
x = Array(np.array([1, 1, 2]))
y = np.array([1, 2, 3])
y[:] = x
y
```
So I expect it to work for `cupy` as well and not raise
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[20], line 28
     26 x = Array(cp.array([1, 1, 2]))
     27 y = cp.array([1, 2, 3])
---> 28 y[:] = x
     29 y

File cupy[/_core/core.pyx:1641](http://localhost:8888/_core/core.pyx#line=1640), in cupy._core.core._ndarray_base.__setitem__()

File cupy[/_core/_routines_indexing.pyx:51](http://localhost:8888/_core/_routines_indexing.pyx#line=50), in cupy._core._routines_indexing._ndarray_setitem()

File cupy[/_core/_routines_indexing.pyx:1017](http://localhost:8888/_core/_routines_indexing.pyx#line=1016), in cupy._core._routines_indexing._scatter_op()

File cupy[/_core/core.pyx:789](http://localhost:8888/_core/core.pyx#line=788), in cupy._core.core._ndarray_base.fill()

File cupy[/_core/core.pyx:1293](http://localhost:8888/_core/core.pyx#line=1292), in cupy._core.core._ndarray_base.__nonzero__()

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

### To Reproduce

```py
import numpy as np
import cupy as cp

class Array:
    def __init__(self,  data):
        self.data = data

    def __array__(self, *args, **kwargs):
        return self.data.__array__(*args, **kwargs)

    def __cupy_get_ndarray__(self, *args, **kwargs):
        return self.data

    @property
    def __array_interface__(self):
        return self.data.__array_interface__

    @property
    def __cuda_array_interface__(self):
        return self.data.__cuda_array_interface__

    def __eq__(self, other):
        return self.data == other


x = Array(cp.array([1, 1, 2]))
[...]

PR #9418, opened 2025-10-10
ENH,MAINT: Allow cupy-like protocols in setitem and consolidate

seberg:
This allows `cupy_arr[...] = cupy_like` where cupy-like supports the CUDA array interface or `__cupy_get_ndarray__`.

Since there were a few (not many!) places that have a similar logic I, opionatedly, consolidated them into a `cdef inline` function. Similarly moved the test helper classes to not have to repeat them.

The biggest part of the diff is just moving those helpers!

I am very sure that `cdef inline` must be in the `.pyd` file to actually be inlined (which means many are currenlty not).

Closes gh-9089


asi1024:
LGTM!

Issue #8352, opened 2024-06-06
Should `cp.ndarray.__getitem__(some_object)` check to see if `some_object` implements `__cuda_array_interface__` or DLPack?


jpivarski:
### Description

Currently, if you do

```python
cupy_array[some_iterable]
```

CuPy will interrogate `some_iterable` to see if it implements `__array_struct__`, `__array_interface__`, `__array__` (in that order) before concluding that `some_iterable` is not an array and proceed to iterate over it as a Python object.

The [Interoperability documentation](https://docs.cupy.dev/en/stable/user_guide/interoperability.html) says that CuPy produces (for Numba) and consumes (from PyTorch) the `__cuda_array_interface__` protocol, as well as DLPack. But if `some_iterable` produces either or both of these protocols, CuPy does not check.

Should it? Is this a feature request or a bug report?

kmaehashi:
Thanks for the feedback! This feature request looks legit to me for the `__cuda_array_interface__` but would like to check how other libraries are handling it, just in case I'm missing something.

As for the DLPack, explicitly using [`from_dlpack`](https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.from_dlpack.html) is the standard way of importing foreign arrays, so `cupy_array[cupy.from_dlpack(xxx)]` is the expected usage.

seberg:
This seems to go down to `cp.asarray([obj, obj])` failing (i.e. cupy currently accepts anything as an index)
```python
import cupy as cp

class test:
    a = cp.arange(10)
    @property
    def __cuda_array_interface__(self):
        return self.a.__cuda_array_interface__
    def __array__(self):
        print("converting via numpy")
        return self.a.get()

cp.asarray(test())  # via interface (prints nothing_
cp.asarray([test()])  # prints: converting via numpy
```

This can be fixed by checking for `__cuda_array_interface__` in `_compute_concat_info_impl` and getting the shape and dtype directly from it.  Caveat: If there is real work happening when feetching `__cuda_array_interface__` that would happen twice.

@jpivarski this also means there is a work-around available, though:  You can add `__cupy_get_ndarray__` which _is_ supported here.  You do have to have valid `.shape` and `.dtype` defined though.

seberg:
Going to close this since I forgot to tag the PR.

Nested CAI will now be supported (v14); i.e. nesting an object should be supported if it is supported by `cupy.asarray()`, I think.
[...]

PR #9419, opened 2025-10-10
ENH: Accept nested CAI arrays in cupy.asarray and indexing 

seberg:
**~DRAFT: Based on previous PR gh-9418~**

This fixes nested CAI coercion and indirectly gh-8352.  (I am not sure how much of a priority this is in the end!)

I changed the flattening code to append recursively. It is still
depth first.
The reason is that it is now necessary to convert to a cupy array
earlier or `obj.ndim` failed in the tests for CAI objects.

The `min_scalar_type` call was the original source of the issue and problematic.  While it is fixed, I opted to remove it from the indexing code anyway.

EDIT: Updated to reflect change in PR content.


cupy/_core/_routines_indexing.pyx
```diff
@@ -262,15 +263,31 @@ cpdef list _prepare_slice_list(slices):
...
+                        '(or boolean) type')
+                # Unlikely/impossible, but there was another error re-raise
+                raise
+
+            try:
+                numpy.asarray(s)  # check if NumPy can convert the index
+            except Exception:
+                # Can't convert, raise original (probably identical) error.
+                # For example `[1, [2]]` is "ragged" and fails this way.
+                raise e from None
```
seberg, line 285: I admit, this got a bit longer than I hoped for and `min_scalar_type` should convert rarely.

But since the `min_scalar_type` looked unclear to me tried to re-organize it to not try ahead of time.

(For reference it was added in gh-7286.)

cupy/_core/core.pyx
```diff
@@ -3072,8 +3079,10 @@ cpdef min_scalar_type(a):
 
     .. seealso:: :func:`numpy.min_scalar_type`
     """
-    if isinstance(a, ndarray):
-        return a.dtype
+    cdef _ndarray_base arr = _convert_from_cupy_like(a, error=False)
+    if arr is not None:
+        return arr.dtype
```
seberg, line 3084: The thing is that `_array_info_from_nested_sequence` actually _requires_ a sequence.
We could use `_compute_concat_info_impl` instead.  But it currently assumes that an object with `__cupy_get_ndarray__` also has a `.shape` and `.dtype` attribute.

So, I assumed that just using the simple conversion pattern is good enough here.

cupy/_core/_routines_indexing.pyx
```diff
@@ -262,15 +263,31 @@ cpdef list _prepare_slice_list(slices):
...
         try:
             s = core.array(s, dtype=None, copy=None)
-        except ValueError:
-            # "Unsupported dtype"
+        except ValueError as e:
+            # Conversion failed, presumably with "Unsupported dtype".
+
+            # If this is a NumPy array, it should have an unsupportd dtype
+            # raise that as as not "integer or boolean" dtype.
+            if isinstance(s, numpy.ndarray):
```
asi1024, line 272: This is a special edge case, but since `s` is a tensor of another CUDA tensor library that supports the CUDA array interface and has a dtype not supported by CuPy, this line might need to evaluate to `True`.

seberg: Hmmm good point, it's surprisingly tricky (to the point that I wouldn't bother making NumPy 100%)!

I'll let it sink in for a moment. `min_scalar_type` is probably pragmatic here in the end.
I think I'll apply this and that should get us as close as we reasonably can:
```
    try:
        dtype = min_scalar_type(s)
    except Exception:
        raise e from None  # raise original error

    if dtype.char in _dtype.all_type_chars:
        raise  # Conversion failed but dtype is fine, original error.
    elif dtype.char != "O" or isinstance(s, numpy.ndarray):
        raise IndexError('arrays used as indices...')
    
     # Object dtype but not a NumPy array is probably an arbitrary
     # object (that can be wrapped into a 0-D array).
     raise IndexError(generic_message)
```

seberg: @asi1024 I am going back on my statement.  You are right, but the TL;DR is: I am tempted to just delete this for simplicity or otherwise would keep it approximately as is.

NumPy also raises the generic `IndexError` for `__array_interface__` objects.  The specific error is only used for `numpy.ndarray`.

So matching that, we could say that only `cupy.ndarray` should get this specific error and that is already the case.
Or we keep broadening it up to `numpy.ndarray` input as well (which this attempts).

The problem is that for example `arr[1.0]` would also go into the "arrays used as indices" path there...  NumPy may have a fighting chance to make this specific, but I don't really think even there it is worth the trouble.

EDIT: I'll simplify this to check for `dtype.kind not in {'b', 'i', 'u'}` though.

asi1024: Indeed, as you say, this case is unlikely to become an issue. Let's revisit it if someone opens an issue.

asi1024:
LGTM!

Issue #6702, opened 2022-05-05
cupy.is_available() raises CUDARuntimeError


dmargala:
### Description

We've encountered an unexpected error while attempting to run cpu-only unit tests on a shared server that does not have GPUs. The code uses something like the following to decide it should run tests that require cupy + GPUs:

```python
is_cupy_available = False
try:
    import cupy
    is_cupy_available = cupy.is_available()
except ImportError:
    pass
```

However, `cupy.is_available()` is unexpectedly raising a CUDARuntimeError:

```
(cupy-test) dmargala@cori06:~> python -c "import cupy; print(cupy.is_available())"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/global/homes/d/dmargala/.conda/envs/cupy-test/lib/python3.10/site-packages/cupy/__init__.py", line 36, in is_available
    return cuda.is_available()
  File "/global/homes/d/dmargala/.conda/envs/cupy-test/lib/python3.10/site-packages/cupy/cuda/__init__.py", line 84, in is_available
    _available = runtime.getDeviceCount() > 0
  File "cupy_backends/cuda/api/runtime.pyx", line 357, in cupy_backends.cuda.api.runtime.getDeviceCount
  File "cupy_backends/cuda/api/runtime.pyx", line 360, in cupy_backends.cuda.api.runtime.getDeviceCount
  File "cupy_backends/cuda/api/runtime.pyx", line 142, in cupy_backends.cuda.api.runtime.check_status
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorMpsConnectionFailed: MPS client failed to connect to the MPS control daemon or the MPS server
```

Some other user on the system left a CUDA MPS daemon running (why they have CUDA MPS running on a server without GPUs is another question):

```
(cupy-test) dmargala@cori06:~> ps aux | grep cuda-mps | grep -v grep
xxxxx  22343  0.0  0.0  10676  1508 ?        Ss   May02   0:08 nvidia-cuda-mps-control -d
```

It seems like `numba.cuda.is_available()` can handle this so I was wondering if `cupy.is_available()` could catch the CUDARuntimeError for us?

```
(cupy-test) dmargala@cori06:~> python -c "import numba.cuda; print(numba.cuda.is_available())"
False
```

Redirecting the default CUDA MPS pipe directory prevents an uncaught exception from occuring:

```
[...]

Issue #9091, opened 2025-04-06
`CUDARuntimeError` raised on first CuPy call in session when no GPU is present; second call returns expected `False`


geezah:
### Description

When invoking a Python interpreter on a machine with no CUDA drivers installed and running a CUDA availability check (e.g., cupy.is_available()), the first call raises a CUDARuntimeError. However, calling the same function again in the same interpreter session returns False (as expected).

This behavior is unexpected—ideally, it should return False even on the first call, without raising an exception.

### To Reproduce

1. Start a new Python interpreter (e.g., python3). Make sure to create and activate an environment where `cupy` is installed.
3. Run the following commands:
```py
import cupy
cupy.is_available()
cupy.is_available()
```

The first invocation throws the following error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/path/to/.venv/lib/python3.11/site-packages/cupy/__init__.py", line 34, in is_available
    return cuda.is_available()
           ^^^^^^^^^^^^^^^^^^^
  File "/path/to/.venv/lib/python3.11/site-packages/cupy/cuda/__init__.py", line 93, in is_available
    _available = runtime.getDeviceCount() > 0
                 ^^^^^^^^^^^^^^^^^^^^^^^^
  File "cupy_backends/cuda/api/runtime.pyx", line 391, in cupy_backends.cuda.api.runtime.getDeviceCount
  File "cupy_backends/cuda/api/runtime.pyx", line 394, in cupy_backends.cuda.api.runtime.getDeviceCount
  File "cupy_backends/cuda/api/runtime.pyx", line 146, in cupy_backends.cuda.api.runtime.check_status
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version
```

The second invocation correctly returns `False`.

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
/path/to/.venv/lib/python3.13/site-packages/cupyx/_runtime.py:316: UserWarning: Failed to detect number of GPUs: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version
OS                        : Linux-6.8.0-57-generic-x86_64-with-glibc2.39
Python Version            : 3.13.1
CuPy Version              : 13.4.1
CuPy Platform             : NVIDIA CUDA
NumPy Version             : 2.2.4
SciPy Version             : None
Cython Build Version      : 3.0.12
[...]

PR #9420, opened 2025-10-10
Make `cuda.is_available()` guard against (almost) all errors

seberg:
Based on the issues, it seems best if `cuda.is_available()` can in practice never fail.
Since it seems to have no docs entry, added it to the CUDA docs section and some brief docs.

It is a bit unfortunate that one has to do random cupy calls for diagnosing problems, but it didn't seem very useful to add `is_available(show_failure=True)` or document that.

Closes gh-6702, gh-9091

(Sanity checked with `CUDA_VISIBLE_DEVICES=""`, although that error is of course the one that was unproblematic before!)

### DIFF
--- a/tests/import_tests/test_import.py
+++ b/tests/import_tests/test_import.py
@@ -8,7 +8,7 @@ Test to ensure that this file can be imported without CUDA Toolkit.
 """
 
 # Running without CUDA tookit, is_available should return False.
-assert not cupy.is_available()
+assert isinstance(cupy.is_available(), bool)
 
 
 @cupy.memoize()


kmaehashi:
Thanks @seberg, looks good! Could you add `cupy.is_available()` line to this test?: https://github.com/cupy/cupy/blob/main/tests/import_tests/test_import.py
This is a test for special set of APIs that are guaranteed to work on CPU-only environment (even without CUDA driver/toolkit installed), which is tested by cuda-build CI matrix: https://github.com/cupy/cupy/blob/main/.pfnci/linux/tests/cuda-build.sh

tests/import_tests/test_import.py
```diff
@@ -7,6 +7,9 @@
 Test to ensure that this file can be imported without CUDA Toolkit.
 """
 
+# Running without CUDA tookit, is_available should return False.
+assert not cupy.is_available()
```
kmaehashi: Actually this file is also imported within GPU CI. I think it's better to add a guard but for now let's check  if it's bool.

```suggestion
assert isinstance(cupy.is_available(), bool)
```

kmaehashi (approved):

kmaehashi:
@asi1024 Could you approve this one as this is BC-breaking?

asi1024 (approved):

kmaehashi:
/test force-skip

All test failures are timeouts unrelated to this PR and known flaky (#7240) https://ci.preferred.jp/r/job/200800

PR #9421, opened 2025-10-15
Fix Python version requirements in `pyproject.toml`

kmaehashi:
Closes #9405.

PR #9423, opened 2025-10-15
Fix for Ruff UP041

asi1024:
Related to https://github.com/cupy/cupy/pull/9421.


asi1024:
Resolved the conflict with #9424.
/test mini

PR #9406, opened 2025-10-02
Drop support of CUDA 11.x and NumPy v1.x

asi1024:
Close https://github.com/cupy/cupy/issues/9319.

Issue #9204, opened 2025-06-24
[RFC] Remove `cupyx.tools.install_library` in CuPy v15


kmaehashi:
CuPy supports several "satellite" CUDA libraries: cuTENSOR, NCCL, and cuDNN. Historically CuPy's binary packages have been built with specific versions of these libraries, and we were asking users to install that version of libraries using `cupyx.tools.install_library` tool, which downloads library archives from NVIDIA website and extract them under `~/.cupy`.

However, today these libraries are also available on PyPI (e.g. https://pypi.org/project/cutensor-cu12/ and https://pypi.org/project/nvidia-nccl-cu12/), and it is becoming common to use the PyPI installation among CUDA/Python projects (such as PyTorch). And now that CuPy supports loading both cuTENSOR and NCCL from PyPI wheels (#9201), it does not make much sense to continue providing the tool. Note that cuDNN support is going to be removed as discussed in a separate RFC issue (#8215).

Therefore I am thinking of removing the tool in CuPy v14. I believe this doesn't impact most of the use-cases but will keep this RFC issue open for a while to allow anyone to jump in the discussion.

kmaehashi:
If we do this the steps would be:

- [ ] Add wheel metadata as static files (instead of dynamically generating them by `install_library.py` and `_generate_wheel_metadata.py`)
- [ ] Update `cupy-release-tools` to install satellite libraries via pip instead of `install_library` (note: libraries must be available when building wheel packages)
    - Note: Wheel packages do not contain `XXX.so` (symlink to `XXX.so.12`) as wheels do not support symlinks. We will need a workaround for that.
- [ ] Remove `install_library.py` and `_generate_wheel_metadata.py`, and update relevant docs

kmaehashi:
The tool is marked deprecated in CuPy v14 (#9432) and will be removed in the future release, possibly in v15.

leofang:
Note to self: Before we completely drop the dependency on NCCL/cuTENSOR, we should move the code to `ci/tools/`. Today we use them in the CI to build wheels and they are handy.

PR #9432, opened 2025-10-18
Deprecate `cupyx.tools.install_library` tool

kmaehashi:
Deprecate `cupyx.tools.install_library` tool as discussed in #9204.

The initial plan was to remove the tool completely, however I'd keep this in deprecated state for CuPy v14 because:
* It requires some effort to completely remove the tool since the tool is also used in cupy-release-tools for package build
* It is better to give users time for migration

This PR also includes docs updates that were overlooked in #9200 and #9406.


kmaehashi:
The test passed. Can anyone approve? @seberg @leofang

seberg:
Let's put this in then, thanks Kenichi-san!

PR #9433, opened 2025-10-19
Drop support for ROCm 6 or earlier

kmaehashi:
* Removes support for ROCm 6 or earlier.
* Add unofficial fork (`amd-cupy` package) to duplicate detection.

cc/ @gpinkert

### DIFF
--- a/.github/workflows/pretest.yml
+++ b/.github/workflows/pretest.yml
@@ -107,4 +107,5 @@ jobs:
 
     - name: Build & Test
       run: |
-        docker run --rm -v "${PWD}:/src" -w /src "rocm/dev-ubuntu-22.04:7.0.2-complete" bash .github/workflows/scripts/pretest-rocm-test.sh
+        # Note: "-complete" Docker image variants exceeds disk size on GitHub hosted runners.
+        docker run --rm -v "${PWD}:/src" -w /src "rocm/dev-ubuntu-22.04:7.0.2" bash .github/workflows/scripts/pretest-rocm-test.sh


docs/source/install.rst
```diff
@@ -486,21 +470,20 @@ Limitations
 
 The following features are not available due to the limitation of ROCm or because that they are specific to CUDA:
 
-* CUDA Array Interface
 * cuTENSOR
 * Handling extremely large arrays whose size is around 32-bit boundary (HIP is known to fail with sizes `2**32-1024`)
```
gpinkert, line 474: This is starting to get fixed in ROCm 6 and 7 builds

docs/source/install.rst
```diff
@@ -486,21 +470,20 @@ Limitations
...
 * Atomic addition in FP16 (``cupy.ndarray.scatter_add`` and ``cupyx.scatter_add``)
 * Multi-GPU FFT and FFT callback
 * Some random number generation algorithms
```
gpinkert, line 477: Can we specify the random number generator algorithms?

docs/source/install.rst
```diff
@@ -486,21 +470,20 @@ Limitations
 
 The following features are not available due to the limitation of ROCm or because that they are specific to CUDA:
 
-* CUDA Array Interface
 * cuTENSOR
 * Handling extremely large arrays whose size is around 32-bit boundary (HIP is known to fail with sizes `2**32-1024`)
 * Atomic addition in FP16 (``cupy.ndarray.scatter_add`` and ``cupyx.scatter_add``)
 * Multi-GPU FFT and FFT callback
```
gpinkert, line 476: https://github.com/ROCm/rocm-libraries/tree/develop/projects/hipfft

docs/source/install.rst
```diff
@@ -409,7 +409,7 @@ Requirements
 
 * `AMD GPU supported by ROCm <https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html>`_
 
-* `ROCm <https://rocm.docs.amd.com/en/latest/>`_ 4.x / 5.x / 6.x
+* `ROCm <https://rocm.docs.amd.com/en/latest/>`_ 7.x
```
gpinkert, line 412: Maybe pin to the latest docs - https://rocm.docs.amd.com/en/docs-7.0.2/ ? 
Otherwise this link will point to version eight whenever that gets released.

.github/workflows/pretest.yml
```diff
@@ -107,4 +107,4 @@ jobs:
 
     - name: Build & Test
       run: |
-        docker run --rm -v "${PWD}:/src" -w /src "rocm/dev-ubuntu-22.04:6.4" bash .github/workflows/scripts/pretest-rocm-test.sh
+        docker run --rm -v "${PWD}:/src" -w /src "rocm/dev-ubuntu-22.04:7.0.2-complete" bash .github/workflows/scripts/pretest-rocm-test.sh
```
gpinkert: Has this been tested?

.github/workflows/pretest.yml
```diff
@@ -107,4 +107,5 @@ jobs:
 
     - name: Build & Test
       run: |
-        docker run --rm -v "${PWD}:/src" -w /src "rocm/dev-ubuntu-22.04:7.0.2-complete" bash .github/workflows/scripts/pretest-rocm-test.sh
+        # Note: "-complete" Docker image variants exceeds disk size on GitHub hosted runners.
```
gpinkert, line 110: Oh, that is unfortunate :(

docs/source/install.rst
kmaehashi: I don't have a list in hand, everything we mark `xfail` in unittests on ROCm CI has failed in the past, but we are not checking the status on latest ROCm.

docs/source/install.rst
kmaehashi: We might want to support multiple versions (7.x and 8.x) in the future so let's keep it as is.

.github/workflows/pretest.yml
kmaehashi: This failed and fixed to use the minimal image instead.

Issue #9430, opened 2025-10-17
NCCL Errors on _init_with_tcp_store


priyammaz:
### Description

I am struggling to get distributed Cupy to work on a system. I have it working on one (its a 2xa6000 system) but cannot get it on another (2xGH200). Im not sure if its an architecture difference but I get the error:

```bash
Traceback (most recent call last):
  File "/u/priyamm2/MyTorch/tests/all_reduce_test_wo_launch.py", line 27, in <module>
    main(args.rank, args.world_size)
  File "/u/priyamm2/MyTorch/tests/all_reduce_test_wo_launch.py", line 8, in main
    comm = NCCLBackend(n_devices=world_size, rank=rank, host="127.0.0.1", port=13333)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/priyamm2/.conda/envs/sidetorch/lib/python3.12/site-packages/cupyx/distributed/_nccl_comm.py", line 82, in __init__
    self._init_with_tcp_store(n_devices, rank, host, port)
  File "/u/priyamm2/.conda/envs/sidetorch/lib/python3.12/site-packages/cupyx/distributed/_nccl_comm.py", line 105, in _init_with_tcp_store
    shifted_nccl_id = bytes([b + 128 for b in nccl_id])
```

Here is the system information:

```bash
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.163.01             Driver Version: 550.163.01     CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GH200 120GB             On  |   00000019:01:00.0 Off |                    0 |
| N/A   23C    P0            140W /  900W |     677MiB /  97871MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA GH200 120GB             On  |   00000029:01:00.0 Off |                    0 |
| N/A   22C    P0            122W /  900W |     675MiB /  97871MiB |      0%      Default |
[...]

PR #9434, opened 2025-10-20
NCCL ID Fix for ARM systems

priyammaz:
This is a fix for this issue here:

https://github.com/cupy/cupy/issues/9430

But also, if our only goal is to make sure its within 8 bits, a better fix may be:

```python
shifted_nccl_id = bytes([b & 0xFF for b in nccl_id])
nccl_id = tuple([int(b) for b in shifted_nccl_id])
```

I can update it if that is more correct (and will work for all systems regardless of how they encode chars)

### DIFF
--- a/cupyx/distributed/_nccl_comm.py
+++ b/cupyx/distributed/_nccl_comm.py
@@ -104,13 +104,15 @@ class NCCLBackend(_Backend):
             # get_unique_id return negative values due to cython issues
             # with bytes && c strings. We shift them by 128 to
             # make them positive and send them as bytes to the proxy store
-            shifted_nccl_id = bytes([b + 128 for b in nccl_id])
+            #shifted_nccl_id = bytes([b + 128 for b in nccl_id])
+            shifted_nccl_id = bytes([b for b in nccl_id])
             self._store_proxy['nccl_id'] = shifted_nccl_id
             self._store_proxy.barrier()
         else:
             self._store_proxy.barrier()
             nccl_id = self._store_proxy['nccl_id']
-            nccl_id = tuple([int(b) - 128 for b in nccl_id])
+            #nccl_id = tuple([int(b) - 128 for b in nccl_id])
+            nccl_id = tuple([int(b) for b in nccl_id])
         self._comm = nccl.NcclCommunicator(n_devices, nccl_id, rank)
 
     def _check_contiguous(self, array):


cupyx/distributed/_nccl_comm.py
```diff
@@ -104,13 +104,15 @@ def _init_with_tcp_store(self, n_devices, rank, host, port):
             # get_unique_id return negative values due to cython issues
             # with bytes && c strings. We shift them by 128 to
             # make them positive and send them as bytes to the proxy store
-            shifted_nccl_id = bytes([b + 128 for b in nccl_id])
+            #shifted_nccl_id = bytes([b + 128 for b in nccl_id])
+            shifted_nccl_id = bytes([b for b in nccl_id])
```
seberg, line 108: Thanks a lot!

Yeah, this shouldn't work for non-ARM as you said. For the way back (below), I think we can just use `<char>b` (no `int()` needed). Since that is what `get_unique_id` uses as well (and I _think_ cython must allow conversion from byte-string to `char` even if signed).

For the one above, I think we need to do `<unsigned char><char>b` and explain in the comment: Ensure positive values for conversion to bytes via Python.
(this would be less awkward if `nccl.get_unique_id()` was a `cdef` function or just returned the `bytes` the bytes...).

What would be amazing is if you could ad a very small test for this in `tests/cupy_tests/cuda_tests/test_nccl.py`, I assume we are missing that and that will also proof that everything is still good on non-ARM system.

(I would also rename the variable to just `nccl_id_bytes` or so, I think the comment is enough, and "shift" reads weird when it's a cast.)

seberg: Thanks, after there was another issue about this (or something similar lower)...

I am thinking we should fix this at the lower level, the next CuPy version is v14, so a change that _might_ break some niche users seems OK to me.

That is: Make `nccl.get_unique_id()` return a bytes string which will remove this dance and simplify the code generally.

If you are interested in pursuing this, that is great. But we should fix this pretty soon for CuPy v14, so I would pick it up very soon.

asi1024:
@priyammaz ~Could you update this PR according to @seberg’s feedback? Also, please make sure that the pre-review tests pass successfully.~ Sorry, I just noticed that @seberg has already sent an update in the PR https://github.com/cupy/cupy/pull/9438.

seberg:
Yeah, thanks for opening a PR @priyammaz! I hope you don't mind that I fixed it at a deeper level in the other PR, but this is now unnecessary.

priyammaz:
thank you! Sorry i was swamped but it is all working on my end with the manual change! glad I was able to find an error though!

PR #9437, opened 2025-10-22
BUG,MAINT: Restructure SingleDeviceMemoryPool and locking

seberg:
**I'll keep this as draft for now, it should work OK but I want to re-think and maybe even roll back a few changes.**

This restructures the single device memory pool, trying to keep it "moderate".  But it does have some significant changes:

1. Arena lifetime is now tied to the chunks which own the arena. This is circular (if the chunk is free and part of the arena) although that is invisible to Python.

   The reason twofold:
   * Prevent cleanup of an arena with active chunks.
   * It is nice to avoid re-entrant/recursive locking. That requires that `free()` can safely put the chunk somewhere that is (effectively) atomic. But I would like that somewhere to live on the `arena` and `arena` discovery _might_ need locking (at least if we may randomly delete the arena).

2. Use atomics for any limit calculations to avoid having to think about which lock to use in what order. (May not matter in the end, but doesn't hurt, I think.)

3. Restructure the internals.  ~I had far more maximal thoughts, but for now keep it slightly more minimal. I did opt for using an `std::map` since it should be good at finding the next "large enough" chunk. I think the removed complexity much outweighs the added need for manual INCREF/DECREF.~
   Changed this to an `std::set` which is also good at finding the next slot and has a `uintptr_t` pointing to the actual chunk as the second part of the key. We can thus immediately find the next free chunk and also find the chunk for merging.
   (To be clear, I think a bucket system can probably be slightly more performant, but unless we do a maximal solution this is just as performance and much simpler.)

This means that the code now definitely doesn't need a re-entrant/recursive mutex anymore.

One other small thing to note maybe: Python containers should be pretty safe (also in free-threading), but iterating them would not be.

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1424,7 +1424,7 @@ cdef class SingleDeviceMemoryPool:
             finally:
                 self._arena_mutex.unlock()
         else:
-            chunk.arena.add_pending_free_atomic(self)
+            chunk.arena.add_pending_free_atomic(chunk)
 
     cpdef free_all_blocks(self, stream=None):
         """Free all **non-split** blocks for one or all arenas.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -954,12 +954,14 @@ cdef class PooledMemory(BaseMemory):
     cdef:
         readonly object pool
         readonly str identity
+        _Chunk chunk
         dict __dict__
 
     def __init__(self, _Chunk chunk, pool):
         self._init(chunk, pool)
 
     cdef _init(self, _Chunk chunk, pool):
+        self.chunk = chunk
         self.ptr = chunk.ptr()
         self.size = chunk.size
         self.device_id = chunk.mem.device_id
@@ -1017,7 +1019,7 @@ cdef class PooledMemory(BaseMemory):
                                          mem_ptr=ptr,
                                          pmem_id=pmem_id)
                 try:
-                    (<SingleDeviceMemoryPool>pool).free(ptr, size)
+                    (<SingleDeviceMemoryPool>pool).free(self.chunk)
                 finally:
                     for hook in hooks.values():
                         hook.free_postprocess(device_id=device_id,
@@ -1025,7 +1027,7 @@ cdef class PooledMemory(BaseMemory):
                                               mem_ptr=ptr,
                                               pmem_id=pmem_id)
                 return
-        (<SingleDeviceMemoryPool>pool).free(ptr, size)
+        (<SingleDeviceMemoryPool>pool).free(self.chunk)
 
     def __dealloc__(self):
         if _exit_mode:
@@ -1220,12 +1222,6 @@ cdef class SingleDeviceMemoryPool:
     cdef:
         object _allocator
 
-        # Map from memory pointer of the chunk (intptr_t) to the corresponding
-        # Chunk object. All chunks currently allocated to the application from
-        # this pool are stored.
-        # No locking is necessary but update `_in_use_bytes` when modifying.
-        dict _in_use
-
         # Arenas are stored as weak references inside this dict (very minimal
         # WeakValueDict). They must only be taken via `_arena(ident)` or the
         # `_Chunk` attribute that keeps the arena alive.
@@ -1254,7 +1250,6 @@ cdef class SingleDeviceMemoryPool:
     def __init__(self, allocator=None):
         if allocator is None:
             allocator = _malloc
-        self._in_use = {}
         self._arenas = {}
         self._allocator = allocator
         self._weakref = weakref.ref(self)
@@ -1371,7 +1366,6 @@ cdef class SingleDeviceMemoryPool:
             chunk = _Chunk.__new__(_Chunk)
             chunk._init(mem, 0, size, arena)
 
-        self._in_use[chunk.ptr()] = chunk
         self._in_use_bytes += chunk.size
 
         pmem = PooledMemory.__new__(PooledMemory)
@@ -1380,14 +1374,7 @@ cdef class SingleDeviceMemoryPool:
         ret._init(pmem, 0)
         return ret
 
-    cpdef free(self, intptr_t ptr, size_t size):
-        cdef _Chunk chunk
-
-        try:
-            chunk = self._in_use.pop(ptr)
-        except KeyError:
-            raise RuntimeError('Cannot free out-of-pool memory')
-
+    cdef free(self, _Chunk chunk):
         self._in_use_bytes -= chunk.size
 
         # Make sure freeing is always safe, but if we can lock do it.
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -997,14 +997,17 @@ cdef class PooledMemory(BaseMemory):
         buffer to the memory pool for reuse.
 
         """
+        cdef cpython.PyObject *pool_ref
         cdef intptr_t ptr
         ptr = self.ptr
         if ptr == 0:
             return
         self.ptr = 0
-        pool = self.pool()
-        if pool is None:
-            return
+        if cpython.PyWeakref_GetRef(self.pool, &pool_ref) == 1:
+            pool = <object>pool_ref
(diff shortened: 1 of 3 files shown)


seberg:
Going to mark as ready for review, although I'll probably need to udpate it after merging the first free-threading PR.

I may be hitting the deadlock here (occasionally), but maybe more prominently, I am hitting the `gc.disable()` race-condition, where the GC can be off forever.

seberg:
I fixed a small bug and added a stress-test (which found that bug).  I'll note that while I think the single-thread performance of the stress-test is slightly improved, the multi-thread congestion may actually be very slightly worse :shrug:.

I was not able to trigger crashes/hangs in either old or new code though with this stress test yet...

EDIT: In the test, the work is almost nothing related to these changes (cudaFree/cudaMalloc get called quite often also). But even without that there is basically no difference. It's just that the unrelated work is dominant.

EDIT2: I may actually see the bug that the `gc` can get permanently disabled without this. At least for some stress tests, the new version doesn't crash with OOM where the old one does.

cupy/cuda/memory.pyx
```diff
@@ -1195,26 +1223,32 @@ cdef class SingleDeviceMemoryPool:
...
+        # Arenas are stored as weak references inside this dict (very minimal
+        # WeakValueDict). They must only be taken via `_arena(ident)` or the
+        # `_Chunk` attribute that keeps the arena alive.
+        # NOTE: If you work with any arena you must hold the `_arena_mutex`.
+        # The only safe operation is `add_pending_free_atomic` (note the
+        # atomic # in the name).
         dict _arenas
+        # NOTE: Never use `.lock()` outside a `with nogil:` statement as it
+        # may deadlock (GIL may be unlocked and another thread also locks).
+        cpp_mutex _arena_mutex
```
kumaraditya303, line 1238: if you use cython.pymutex instead of C++ lock then this wouldn't be an issue

seberg: Hmmm, thanks good point. I guess coming from the recursive mutex and PyMutex being 3.13, I didn't think of it. But Cython does backports it (at least unless we aim to do stable API builds soon).

(I may update it later, or try and remember to follow-up soon. It would be nice to also change the pinned-memory here, which uses the `recursive_mutex` for historic reasons, but I don't think it needs to).

kumaraditya303: > But Cython does backports it (at least unless we aim to do stable API builds soon).

On older versions without `PyMutex`, cython fallbacks to `PyThread_type_lock` so you can use `cython.pymutex` on all versions.

cupy/cuda/memory.pyx
```diff
@@ -990,176 +1022,177 @@ cdef class PooledMemory(BaseMemory):
...
-cdef lock_and_no_gc(recursive_mutex& lock):
-    cdef _LockAndNoGc self = _LockAndNoGc.__new__(_LockAndNoGc)
-    self._lock = &lock
-    return self
+    cdef insert_chunk(self, _Chunk chunk, bint merge=True):
+        if merge:
+            chunk = self.try_merge_chunk(chunk)
 
+        self.index.insert(index_type(chunk.size, <uintptr_t><void *>chunk))
+        cpython.Py_INCREF(chunk)  # index holds a reference now
```
kumaraditya303, line 1111: This probably does not matter for this case because of locking, but in general it is better to incref the object first then insert as otherwise some other thread could see an object which is borrowed.

seberg: Yeah, let me change that, it's good habit. I guess I would care more in C and there you should try to use `Py_SETREF()` :).
(In Cython it is hard to get it wrong because what you pass in to `Py_INCREF` *must* be a strong reference!)

seberg:
More interesting review wise: I did actually add one thing here.  I removed the ablity to call `arr.data.mem.pool().free(ptr, size)`.  There is no other way to call that function, I am very sure and there is no reason to ever call it as you have to call `arr.data.mem.free()` to free the memory.

The main reason that is very useful is that it removes the `_in_use` dict entirely.

So I think all this assumes is that nobody does:
```
m = malloc(100)
ptr = m.ptr
size = m.size
m.ptr = 0
free = m.pool().free
del m

# code
free(ptr, size)
```

And a nice thing: Together with the thread-local PR (maybe there is more to be done there, but I think it's a good step), that actually makes `malloc()` ~50% faster.

cupy/cuda/memory.pyx
```diff
@@ -1192,53 +1225,75 @@ cdef class SingleDeviceMemoryPool:
...
-        # `_total_bytes_lock` must be acquired to access it.
-        size_t _total_bytes_limit
+        # NOTE: You MUST use _try_block_total_bytes to increase the value
+        # (This ensures correct checks for whether memory is available).
+        std_atomic[size_t] _total_bytes
+        # Number of used bytes, modify with normal +=/-=.
+        std_atomic[size_t] _in_use_bytes
+        # Upper limit of the amount to be allocated by this pool, we don't
+        # care too much about thread-safety for it, but make it atomic anyway.
+        std_atomic[size_t] _total_bytes_limit
```
kumaraditya303, line 1248: You can use relaxed ordering for this as there is no real ordering necessary, it would be more performant on weaker memory models like arm64.

cupy/cuda/memory.pyx
```diff
@@ -990,176 +1022,177 @@ cdef class PooledMemory(BaseMemory):
...
-    The purpose of disabling GC is to prevent unexpected recursion.
-    See gh-2074 for details.
-    """
+    cdef add_pending_free_atomic(self, _Chunk chunk):
+        """Add a chunk back to the arena as free. If you are holding the lock,
+        use `insert_chunk()` directly (it will merge unless `merge=False`).
+        """
+        # In theory we could try to lock here, but in practice the caller
+        # should handle that when relevant.
+        self._pending_free.append(chunk)
```
leofang, line 1081: Dumb question: How is the caller locking today? IIUC `list.append` requires a lock?

seberg: My understanding is that lists/dict, etc. are atomic on new builds.  It _would_ be really nice if these things were documented clearer.  There is a small unknown for me whether there might be some crazy issues on old non-free-threaded builds, but I don't really think it can matter (since the GIL should protect us).

(What is not atomic probably is dict iterators maybe, but those always barfed in threading scenarios, although maybe not crashed.)

leofang:
Sorry for delay @seberg. This is a very nice and compact PR that I relied on Copilot to guide me understand while I was struggling with distraction. To keep the ball rolling I have generated two reports:
- https://gist.github.com/leofang/a4ac97555057fb12c83881a1aebbc8fd#-copilot-11
- [wed_mar_11_2026_review_of_cu_py_pr_9437_concerns.zip](https://github.com/user-attachments/files/25890679/wed_mar_11_2026_review_of_cu_py_pr_9437_concerns.zip)
   - there is a `.hpp` file in which I asked Copilot to vibe-generate an equivalent C++17 implementation, which looks interesting to me but again I did not follow through. It did point out two issues after the direct translation.

Feel free to take a look / dismiss if you think LLM is not making sense here. (Sorry for inconvenience; I don't want to let it comment directly in the PR. I worry it'd add too much noise with little value...)

seberg:
Trying to have a look, harder to go tyhrough and say "nah" than I thought.
A. No, lists are atomic in this sense, this is poorly documented, but the AI is just wrong.
B. `use_bytes` (and `free_bytes()`) is immaterial for the allocation logic itself, so the value only needs to be consistent not a correct snapshot, and that it is.
C. I would shrug this off.  Hooks (or even more custom allocators) recursing seems like something "you aren't gonna need it" to me.
D. To me this is just silly, it is completely correct and we aren't gonna expand the `finally:`
E. Yes (there is no issue pointed out, it's subtler but it has to be).
   * I suppose it would be interesting if a GC run ends up triggering an *allocation*.  I may not have thought of that possibility. It would be a bit strange, but should we worry about it (and just keep using a re-entrant lock)?

The rest just seems like small possible updates or silly comments to me that I may do in another iteration.

leofang:
CI failures are due to either timeout or network glitch.

/test force-skip

leofang (approved):

Issue #9435, opened 2025-10-20
`cupy.cuda.nccl.get_unique_id()` returns signed ints on x86 but unsigned on arm64


yhtang:
### Description

`get_unique_id()` returns a list of integers corresponding to `char` (aka `typedef struct { char internal[NCCL_UNIQUE_ID_BYTES]; } ncclUniqueId`). The choice of the  underlying int type is implementation-defined in C/C++. It happens to be int8 on x86_64 but uint8 on aarch64 for CuPy's cuda12x wheels.

While in most cases this should have no visible effects, it may break code with OverflowError when we try to cross the arch boundary by creating NCCL unique IDs on one arch but consume it on another.

**Expected**: A stable, cross-platform representation—e.g., documented as 128 bytes (0–255) or return a bytes object.

### To Reproduce

```py
import cupy.cuda.nccl as nccl, platform
uid = nccl.get_unique_id()
print(platform.machine(), uid[:8])  # show negatives on x86_64, non-negative on aarch64
```

Example output:
```
x86_64 (-115, 9, -45, 78, 83, -96, -111, 41)
aarch64 (84, 242, 48, 227, 114, 193, 138, 130)
```

### Installation

Wheel (`pip install cupy-cuda12x`)

### Environment

aarch64
```
OS                           : Linux-6.11.0-1013-nvidia-64k-aarch64-with-glibc2.39
Python Version               : 3.12.3
CuPy Version                 : 13.6.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.2.6
SciPy Version                : 1.16.2
Cython Build Version         : 3.0.12
Cython Runtime Version       : None
CUDA Root                    : /usr/local/cuda
nvcc PATH                    : /usr/local/cuda/bin/nvcc
CUDA Build Version           : 12090
CUDA Driver Version          : 13000
CUDA Runtime Version         : 12090 (linked to CuPy) / 12090 (locally installed)
CUDA Extra Include Dirs      : []
cuBLAS Version               : 120901
cuFFT Version                : 11401
cuRAND Version               : 10310
cuSOLVER Version             : (11, 7, 5)
cuSPARSE Version             : 12510
NVRTC Version                : (12, 9)
Thrust Version               : 200800
CUB Build Version            : 200800
Jitify Build Version         : <unknown>
cuDNN Build Version          : None
cuDNN Version                : None
NCCL Build Version           : 22501
NCCL Runtime Version         : 22703
cuTENSOR Version             : None
[...]

PR #9438, opened 2025-10-22
Change nccl get_unique_id to return a bytes string

seberg:
This simplifies things slightly, but mostly avoids any sensitivity to systems with different char signedness.
I.e. casting the `char` to an integer tuple means we get negative values on many platforms (e.g. intel/AMD CPUs) while ARM system use unsigned char and would give signed values.

Using a bytes string, means that we can exchange that directly in the communicator and don't need to do the (incorrect) `+128` (rather than a cast to unsigned int).

Since CuPy v14 is a major version and since nothing should change for most use-cases this seems sensible to try to just change to me.

Closes gh-9435, gh-9430

(I should probably to confirm this on an ARM system, although if anything the systems would fail probably and Cython generates `memcpy` and `FromBytesAndSize`, so it should be safe.)

### DIFF
--- a/cupy_backends/cuda/libs/nccl.pyx
+++ b/cupy_backends/cuda/libs/nccl.pyx
@@ -330,7 +330,7 @@ cdef class NcclCommunicator:
             raise ValueError(
                 f"commId length {len(commId)} != {NCCL_UNIQUE_ID_BYTES}.")
 
-        _uniqueId.internal[:] = <char *>commId  # memcpy (size checked above)
+        _uniqueId.internal[0:NCCL_UNIQUE_ID_BYTES] = <char *>commId
 
         if NCCL_VERSION_CODE < NCCL_VERSION(2, 18, 0):
             if config is not None:


cupy_backends/cuda/libs/nccl.pyx
```diff
@@ -273,11 +276,13 @@ cdef class NcclCommunicator:
...
+    def __init__(self, int ndev, bytes commId not None, int rank):
         cdef ncclUniqueId _uniqueId
-        assert len(commId) == NCCL_UNIQUE_ID_BYTES
-        for i in range(NCCL_UNIQUE_ID_BYTES):
-            _uniqueId.internal[i] = commId[i]
+        if len(commId) != NCCL_UNIQUE_ID_BYTES:
+            raise ValueError(
+                f"commId length {len(commId)} != {NCCL_UNIQUE_ID_BYTES}.")
+
+        _uniqueId.internal[:] = <char *>commId  # memcpy (size checked above)
```
leofang: We learned this the hard way in cuda.core that Cython won't generate correct code... Either we copy with bounds, something like (untested!)
```cython
_uniqueId.internal[0:NCCL_UNIQUE_ID_BYTES] = <char *>commId
```
or we call `PyBytes_FromStringAndSize`.

cupyx/distributed/_nccl_comm.py
```diff
@@ -101,16 +101,11 @@ def _init_with_tcp_store(self, n_devices, rank, host, port):
         if rank == 0:
             self._store.run(host, port)
             nccl_id = nccl.get_unique_id()
-            # get_unique_id return negative values due to cython issues
-            # with bytes && c strings. We shift them by 128 to
-            # make them positive and send them as bytes to the proxy store
-            shifted_nccl_id = bytes([b + 128 for b in nccl_id])
-            self._store_proxy['nccl_id'] = shifted_nccl_id
+            self._store_proxy['nccl_id'] = nccl_id
```
leofang, line 104: @emcastillo can correct me, but I think the original comment still applies even after this PR?

cupy_backends/cuda/libs/nccl.pyx
seberg: The `<char *>` cast was sufficient magic to make it work here, translates to:
```
  __pyx_t_11 = __Pyx_PyBytes_AsWritableString(__pyx_v_commId); if (unlikely((!__pyx_t_11) && PyErr_Occurred())) __PYX_ERR(0, 333, __pyx_L1_error)
  memcpy(&(__pyx_v__uniqueId.internal[0]), ((char *)__pyx_t_11), sizeof(__pyx_v__uniqueId.internal[0]) * (NCCL_UNIQUE_ID_BYTES - 0));
```
but let me add the `0:NCCL_UNIQUE_ID_BYTES`, it is definitely clearer (it's very confusing to reason about whether cython uses null terminated strings or not in these!)

cupyx/distributed/_nccl_comm.py
seberg: A Python bytes string always looks like unsigned integers.  So from the perspective of a Python user (which this is) changing to bytes makes it always unsigned.

leofang: Ah, thanks Sebastian, I misread the change.

seberg:
> We will want to mention this in docs/source/upgrade.rst

Do prefer I add it here or we just gather it later since it's not much and avoids merge conflicts?  (I had tagged it as no-compat.)

leofang:
No opinion 🙂 But let's give @asi1024 time to take a second look before merging (since this is a no-compat).

asi1024 (approved):
I agree with this change! Thank you for your pull request!

Issue #9015, opened 2025-03-04
Casting an array to a byteswapped dtype does a view not a cast


jcrist:
### Description

When casting a cupy array to a [byteswapped](https://numpy.org/doc/stable/user/byteswapping.html) dtype, cupy incorrectly returns an array viewing the original bytes in the new dtype, when it should return an array casting the values to the new dtype.

### To Reproduce

```python
import cupy as cp
import numpy as np


# Create two identical float32 arrays with native byteorder
x = np.array([1.5, 2.5, 3.5], dtype="float32")
cu_x = cp.array(x)
assert x.dtype == cu_x.dtype

# A byteswapped float32
dtype = np.dtype("float32").newbyteorder()

# Cast to non-native byteorder
x2 = x.astype(dtype)
cu_x2 = cu_x.astype(dtype)

# dtypes are byteswapped, as intended
assert x2.dtype == dtype
assert cu_x2.dtype == dtype

# But for cupy, the values are _viewed_ as the byteswapped dtype,
# while for numpy the values are cast to the byteswapped dtype.
# This leads this check to fail.
cp.testing.assert_allclose(cu_x2, x2)

# The cupy behavior instead matches a `.view` call (which is wrong)
cp.testing.assert_allclose(cu_x2, x.view(dtype))
```

Output:

```
Traceback (most recent call last):
  File "/raid/jcristharif/cuml/bug.py", line 24, in <module>
    cp.testing.assert_allclose(cu_x2, x2)
  File "/raid/jcristharif/miniforge3/envs/cuml-dev/lib/python3.12/site-packages/cupy/testing/_array.py", line 24, in assert_allclose
    numpy.testing.assert_allclose(
  File "/raid/jcristharif/miniforge3/envs/cuml-dev/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 1684, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/raid/jcristharif/miniforge3/envs/cuml-dev/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/raid/jcristharif/miniforge3/envs/cuml-dev/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 885, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Not equal to tolerance rtol=1e-07, atol=0

[...]

PR #9440, opened 2025-10-24
ENH: Minimal support for structured dtypes

seberg:
This is at a PoC stage right now. It provides minimal support for structured dtypes, i.e. conversion from NumPy arrays and indexing a single field.

The "trick" is that the array returned by the single field index is very much usable, though.
Plus, I guess the whole array is usable by numba kernels.

CC @leofang just in case you are curious.

---

Draft for now, needs tests and some thoughts... Notes:
* Nothing succeeds with this, not even copies!  Only indexing and assignment works, but of course those some of the most important things to do!
* I could implement multi-field support easily, but I doubt there is much use.
* I might like protection against unaligned dtypes (at index time or even earlier)?  NumPy structured dtypes are unaligned and you can easily get an array where all operations crash with a bad alignment error.

---

Edit: The re-arranging here of checks means that gh-9015 would lead to an error now, because byte-swappedness is checked basically always after it.

### DIFF
--- a/.github/workflows/auto-cc.yml
+++ b/.github/workflows/auto-cc.yml
@@ -12,7 +12,7 @@ jobs:
         github.repository_owner == 'cupy' &&
         contains('|hip|array-api|', github.event.label.name)
     env:
-      NOTIFY_TARGETS: '{"hip": "@amathews-amd", "array-api": "@asmeurer @leofang"}'
+      NOTIFY_TARGETS: '{"hip": "@gpinkert", "array-api": "@asmeurer @leofang"}'
       ISSUE_NUMBER: ${{ (github.event_name == 'issues' && github.event.issue.number) || github.event.pull_request.number }}
       ISSUE_BODY_TAG: "<!-- Added by Auto Notify Bot (${{ github.event.label.name }}) -->"
     runs-on: ubuntu-22.04
--- a/docs/source/user_guide/kernel.rst
+++ b/docs/source/user_guide/kernel.rst
@@ -3,8 +3,8 @@
 User-Defined Kernels
 ====================
 
-CuPy provides easy ways to define three types of CUDA kernels: elementwise kernels, reduction kernels and raw kernels.
-In this documentation, we describe how to define and call each kernels.
+CuPy provides easy ways to define three types of CUDA kernels: elementwise kernels, reduction kernels, and raw kernels.
+In this documentation, we describe how to define and call each kernel.
 
 
 Basics of elementwise kernels
@@ -58,7 +58,7 @@ Type-generic kernels
 --------------------
 
 If a type specifier is one character, then it is treated as a **type placeholder**.
-It can be used to define a type-generic kernels.
+It can be used to define a type-generic kernel.
 For example, the above ``squared_diff`` kernel can be made type-generic as follows:
 
 .. doctest::
@@ -69,7 +69,7 @@ For example, the above ``squared_diff`` kernel can be made type-generic as follo
    ...     'z = (x - y) * (x - y)',
    ...     'squared_diff_generic')
 
-Type placeholders of a same character in the kernel definition indicate the same type.
+Type placeholders of the same character in the kernel definition indicate the same type.
 The actual type of these placeholders is determined by the actual argument type.
 The ElementwiseKernel class first checks the output arguments and then the input arguments to determine the actual type.
 If no output arguments are given on the kernel invocation, then only the input arguments are used to determine the type.
@@ -87,7 +87,7 @@ The type placeholder can be used in the loop body code:
    ...     ''',
    ...     'squared_diff_generic')
 
-More than one type placeholder can be used in a kernel definition.
+More than one type of placeholder can be used in a kernel definition.
 For example, the above kernel can be further made generic over multiple arguments:
 
 .. doctest::
@@ -98,7 +98,7 @@ For example, the above kernel can be further made generic over multiple argument
    ...     'z = (x - y) * (x - y)',
    ...     'squared_diff_super_generic')
 
-Note that this kernel requires the output argument explicitly specified, because the type ``Z`` cannot be automatically determined from the input arguments.
+Note that this kernel requires the output argument to be explicitly specified, because the type ``Z`` cannot be automatically determined from the input arguments.
 
 
 Raw argument specifiers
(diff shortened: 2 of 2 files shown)


cupy/_core/internal.pyx
```diff
@@ -547,3 +545,37 @@ cdef bint _is_layout_expected(
...
+
+    NOTE: We currently assume structured dtypes are always aligned for lack
+    of a clearer definition of what it means for them to be aligned.
+    """
+    cdef Py_ssize_t itemsize = arr.dtype.itemsize
+    cdef Py_ssize_t pow_2_mask = itemsize - 1
+    cdef Py_ssize_t i
+
+    if arr.dtype.char == "V":
+        return True
```
seberg, line 562: This is unfortunate, but I don't think there is any obvious thing to put here, unfortunately.
NumPy has an alignment for `strict=True` but it's not whatever the GPU side will want...

cupy/_core/internal.pyx
```diff
@@ -547,3 +545,37 @@ cdef bint _is_layout_expected(
...
+
+cdef bint check_aligned(_ndarray_base arr) except -1:
+    """Check if an array is aligned with its dtype. We usually assume
+    that this is the case, since cuda allocations are aligned.
+
+    NOTE: We currently assume structured dtypes are always aligned for lack
+    of a clearer definition of what it means for them to be aligned.
+    """
+    cdef Py_ssize_t itemsize = arr.dtype.itemsize
+    cdef Py_ssize_t pow_2_mask = itemsize - 1
```
seberg, line 558: This means we check for an alignment to the itemsize.  I am not actually 100% sure that is correct, but I _think_ that cuda will align and assume alignment of itemsize?

I thought briefly to just not check, but this leads to a `CUDARuntimeError` and that may be fine, but it is a sticky error, i.e. once you run into it, you have to restart the whole session.

cupy/_core/_routines_indexing.pyx
```diff
@@ -47,6 +50,13 @@ cdef _ndarray_base _ndarray_getitem(_ndarray_base self, slices):
 cdef _ndarray_setitem(_ndarray_base self, slices, value):
     if isinstance(value, _ndarray_base):
         value = _squeeze_leading_unit_dims(value)
+
+    if isinstance(slices, str) and self.dtype.fields is not None:
+        # TODO: This is kinda right, but not quite due to finalization
+        # for subclasses.
```
seberg, line 56:
```suggestion
```
This should be OK, I think...  NumPy does the same thing even IIRC (although I prefer a subclass that does non-trivial things to just implement `__setitem__`.

cupy/_core/_dtype.pyx
```diff
@@ -31,6 +31,41 @@ cdef dict _dtype_dict = {}
 cdef _dtype = numpy.dtype
 
 
+cdef bint check_supported_dtype(dtype, bint error) except -1:
+    """ Returns true on success but otherwise raises an error. """
+    cdef str dtype_char = dtype.char
+    if dtype_char in all_type_chars:
+        pass
```
leofang: Return True?

cupy/_core/_dtype.pyx
```diff
@@ -31,6 +31,41 @@ cdef dict _dtype_dict = {}
...
+            raise ValueError(
+                f"Unsupported dtype {dtype} because it contains references "
+                "which cannot be supported by CuPy.\n"
+                "This may happen even if a dtype only contains basic types "
+                "if the original array contained e.g. object dtype.")
+
+        # We can represents the underlying bytes in CuPy, although it is very
+        # possible that the included fields will not be usable in the end.
+        # The simplest path is to inform users
+        pass
```
leofang: ditto

cupy/_core/internal.pyx
leofang: I assume this is what you meant in the docstring?
>  NOTE: We currently assume structured dtypes are always aligned for lack
>  of a clearer definition of what it means for them to be aligned.

Time to push https://github.com/NVIDIA/cuda-python/issues/734 across the finish line 🙂

cupy/_core/internal.pyx
leofang: Lots of the alignment issues are buried by the fact that CuPy overaligns in most cases (we upcast to allocate in units of 512B). Once we relax that, yes for example nvmath folks do see issues and this is something we will want to think about.

seberg: Right, in practice everything is aligned to itemsize (the allocation higher). But on the CPU an `f4,f8` struct is usually aligned. We assume here that can never be aligned for any basic datatype.

EDIT: That is the `f8` part of it on its own is aligned even though it has a 4 byte offset (a kernel using the full struct would definitely have larger alignment anyway, I think).

cupy/_core/_dtype.pyx
seberg: Intentional but not sure why I put the byte-order check at the end (V has a `|` byte-order, so checking it doesn't matter much either way).
(I think this closes https://github.com/cupy/cupy/issues/9015, although enforcing no big endian arrays is a small change.)

cupy/_core/_dtype.pyx
```diff
@@ -31,6 +31,42 @@ cdef dict _dtype_dict = {}
...
+    cdef str dtype_char = dtype.char
+
+    if dtype.byteorder == ">":
+        if not error:
+            return False
+        raise ValueError(
+            f'Unsupported dtype {dtype} with big-endian byte-order')
+
+    if dtype_char in all_type_chars:
+        return True
```
seberg, line 45: Random realization: This currently tightens the checks for supported dtypes in things like `cupy.empty()` where previously you could abuse things, so long you didn't come from NumPy.

Makes me wonder if I should just only do the second part here (check the flags).  Then rely on the "unsupported dtype" error that should happen on the kernel call for any weird dtype?

tests/cupy_tests/core_tests/test_structured.py
```diff
@@ -0,0 +1,100 @@
...
+                iface = super().__cuda_array_interface__
+                iface["typestr"] = arr.dtype.str
+                iface["descr"] = arr.dtype.descr
+                return iface
+
+        with pytest.raises(ValueError):
+            func(ByteSwapped(base))
+
+
+class TestFieldAccess:
```
leofang, line 48: Can we add a slicing test, something like `return a[::2]`?

seberg: Added, although if you do `a[::2]` you get a strided view so it can't even be copied back to the CPU without launching a kernel first) to gather it into a contiguous one.
And that still fails for now.  So you can do `a[::2]` fine and `a[::2]["field"]` is good, but not `a[::2].get()`

seberg:
Had not run the `cupyx` test suite, there was a stray test copy-pasted from SciPy that used a big endian `">f4"` dtype, which doesn't make sense: just wrapped it into a lazy `pytest.raises(ValueError)`.

leofang:
There is one test failing, is it known?

seberg:
Seems like it: https://github.com/cupy/cupy/issues/7240

PR #9444, opened 2025-10-26
Support CUDA wheels

### LINKED ISSUE #8013: Allow using CUDA libraries installed from pip wheels

leofang:
Close #8013.

This PR adds `cuda-pathfinder` as a dependency to CuPy's CUDA builds, and use it to load needed CUDA libraries. Turns out this is fairly easy: We just need to inject a pathfinder call prior to importing the Python extension module that offers the bindings to the C library. Then, the pathfinder will first look for possible locations of CUDA wheels, if installed. This allows us to support CUDA wheels. The needed CUDA component wheels can be installed optionally via ~~`[all]`~~ `[ctk]`.

It is not clear to me if any of the library loading helpers in `cupy/_environment.py` needs any changes or removal. My local test shows they can stay and do not interfere with the pathfinder.

A separate PR to the release tool implements the actual dependency injection: https://github.com/cupy/cupy-release-tools/pull/466. The CuPy sdist and ROCm builds are not affected.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -89,7 +89,7 @@ By default, the above command only installs CuPy itself, assuming a CUDA Toolkit
 smaller installation footprint and better interoperability with other Python GPU libraries), you can pass ``[ctk]`` to install them all as
 optional dependencies, e.g.::
 
-   $ pip install cupy-cuda12x[ctk]
+   $ pip install "cupy-cuda12x[ctk]"
 
 .. note::
 
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ optional_dependencies = {
     ],
 }
 if not ctx.use_hip:
-    dependencies.append("cuda-pathfinder>=1.3.2,<2.0a0")
+    dependencies.append("cuda-pathfinder>=1.3.2,==1.*")
     if ext_modules:
         try:
             cuda_major = ctx.features["cuda"].get_version() // 1000


leofang:
Questions to decide:
1. Dependency on `cuda-pathfinder`: In order to avoid installing it in the ROCm builds, I have to find various places to inject the dependency (in both this PR and the release tool PR). Right now we still has one hole that is when building from source/sdist. Maybe we should make it a required dependency unconditionally?
2. Do we want to piggyback on the `[all]` tag for installing CUDA wheels? It also installs a few other things (ex: SciPy). If not, what is the alternative name? `[cuda]` might be too verbose because it'd then be `pip install cupy-cuda13x[cuda]`, but maybe it's OK.

docs/source/upgrade.rst
```diff
@@ -9,6 +9,13 @@ Please see also the :ref:`compatibility_matrix` for supported environments of ea
 CuPy v14
 ========
 
+Support for NVIDIA CUDA component wheels
+----------------------------------------
+
+CuPy v14 can be installed together with a minimal CUDA installation all from PyPI (``pip install``), allowing quickly
```
seberg: Might be cool to include a complete incantation.

cupy_backends/cuda/libs/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+)
+
+
+def __getattr__(name):
+    if not _runtime.is_hip and name in _submodules:
+        try:
+            _pathfinder.load_nvidia_dynamic_lib(name)
+        except _pathfinder.DynamicLibNotFoundError as e:
+            if (not (_os.environ.get('READTHEDOCS') == 'True') and
+                    not (_os.environ.get('CUPY_CI') is not None)):
```
seberg, line 31: I wonder if using `not cupy.isavailable():` makes sense here? Just to avoid hard-coding specific environments (if it suffices).

seberg (commented):
Just some tiny comments. I wonder if the different import is (semi) public and could be noted somewhere, but that could be a follow-up when the question comes up too.

> `[cuda]` might be too verbose because it'd then be `pip install cupy-cuda13x[cuda]`, but maybe it's OK.

If it adds up (not sure it does) `cudatk` or `cudart` might be less confusing (even if less to the point maybe)?

> It is not clear to me if any of the library loading helpers in `cupy/_environment.py` needs any changes or removal.

Hmm, unfortunately, I don't have a thought/knowledge on this.  Maybe `_environment.py` should use cuda-pathfinder (but it seems a bit different with that `wheels.json`)?

cupy_backends/cuda/libs/__init__.py
leofang: No, unfortunately it'd incur circular import (which is already the case, e.g. between `cupy.cuda.compiler` and `cupy._core.core` and editable install is not working in the main branch as a result). We need everything in `cupy_backends/` to be at the bottom of the import tree.

cupy/cuda/__init__.py
```diff
@@ -71,6 +73,16 @@ def __getattr__(key):
...
+    elif key == 'cufft':
+        if not runtime.is_hip:
+            from cuda import pathfinder
+            try:
+                pathfinder.load_nvidia_dynamic_lib("cufft")
+            except pathfinder.DynamicLibNotFoundError as e:
+                if (not (_os.environ.get('READTHEDOCS') == 'True') and
+                        not (_os.environ.get('CUPY_CI') is not None)):
+                    raise ImportError(str(e)) from e
+        return _importlib.import_module('cupy.cuda.cufft')
```
leofang, line 85: It seems this is interfering with the cuFFT legacy callback support... for some reason `.set_callback()` seems to not use the statically linked symbol 🤔

kmaehashi:
Thanks Leo for working on this!

> It is not clear to me if any of the library loading helpers in `cupy/_environment.py` needs any changes or removal. My local test shows they can stay and do not interfere with the pathfinder.

I believe nothing needs to be done with `cupy/_environment.py`. It handles cutensor/nccl preloading, but they are not (yet?) supported in pathfinder. On Windows it also adds `%CUDA_PATH%\bin\x64` to the DLL search path. The path is searched for by the pathfinder but users may rely on this behavior to discover cuTENSOR.

> A separate PR to the release tool implements the actual dependency injection: [cupy/cupy-release-tools#466](https://github.com/cupy/cupy-release-tools/pull/466). The CuPy sdist and ROCm builds are not affected.

> 1. Dependency on `cuda-pathfinder`: In order to avoid installing it in the ROCm builds, I have to find various places to inject the dependency (in both this PR and the release tool PR). Right now we still has one hole that is when building from source/sdist.

I think this can be done in CuPy repo alone. Add `dependencies` to `dynamic` list of `pyproject.toml`: https://github.com/cupy/cupy/blob/7445da8d4ee47065c028a1c8e2cfff9ab3e87fef/pyproject.toml#L37. Then in `setup.py` generate `install_requires` list dynamically using build context, then pass it as `setup(..., install_requires=install_requires)`. I guess we need to remove [this fast-path hack](https://github.com/cupy/cupy/blob/7445da8d4ee47065c028a1c8e2cfff9ab3e87fef/setup.py#L56-L61) for this to work.

> 2. Do we want to piggyback on the `[all]` tag for installing CUDA wheels? It also installs a few other things (ex: SciPy). If not, what is the alternative name? `[cuda]` might be too verbose because it'd then be `pip install cupy-cuda13x[cuda]`, but maybe it's OK.

I think use case for using `all` tag and CUDA-wheels tags are orthogonal, so +1 for adding a separate tag. I came up with `toolkit` or `with-toolkit`, but `cuda` also sounds OK.

cupy/cuda/__init__.py
leofang: Fixed in commit 2991485. Turns out this is a bug exposed by this PR. We should have passed `-Bsymbolic` to the linker when statically linking to cuFFT, otherwise depending on how libcufft is loaded we could still be using symbols from the shared library instead of the static copy.

docs/source/upgrade.rst
leofang: done in commit e0851df60

leofang:
Thanks, @kmaehashi san!
> I think this can be done in CuPy repo alone. Add `dependencies` to `dynamic` list of `pyproject.toml`:
> 
> https://github.com/cupy/cupy/blob/7445da8d4ee47065c028a1c8e2cfff9ab3e87fef/pyproject.toml#L37
> 
> . Then in `setup.py` generate `install_requires` list dynamically using build context, then pass it as `setup(..., install_requires=install_requires)`. 

Ah! So we inject the dependency dynamically in the old way. I had forgotten that we could do this. It's done now. 

> I guess we need to remove [this fast-path hack](https://github.com/cupy/cupy/blob/7445da8d4ee47065c028a1c8e2cfff9ab3e87fef/setup.py#L56-L61) for this to work.

I am not sure why it's needed? It works locally for me without touching the quoted lines.

> > 2. Do we want to piggyback on the `[all]` tag for installing CUDA wheels? It also installs a few other things (ex: SciPy). If not, what is the alternative name? `[cuda]` might be too verbose because it'd then be `pip install cupy-cuda13x[cuda]`, but maybe it's OK.
> 
> I think use case for using `all` tag and CUDA-wheels tags are orthogonal, so +1 for adding a separate tag. I came up with `toolkit` or `with-toolkit`, but `cuda` also sounds OK.

I changed it to `pip install cupy-cuda12x[ctk]`, and propagated the change to both this PR and the release tool PR. Let me know what you think. 

`cuda` is a little bit ambiguous. It could mean either the Toolkit or the driver (and we meant the former). `toolkit` is a bit lengthy (7 characters). So I went with `ctk`.

cupy_backends/cuda/api/runtime.pyx
```diff
@@ -165,16 +164,8 @@ cpdef int runtimeGetVersion() except? -1:
...
-        # the real runtime version.
-        # https://github.com/NVIDIA/cuda-python/issues/16
-        cdef int major, minor
-        (major, minor) = nvrtc.getVersion()
-        version = major * 1000 + minor * 10
-    ELSE:
-        status = cudaRuntimeGetVersion(&version)
-        check_status(status)
+    status = cudaRuntimeGetVersion(&version)
+    check_status(status)
```
leofang, line 168: This is a bug fix (removing the circular import is only a happy by-product). Since cuda-bindings also statically links to cudart, the branching to use NVRTC as the proxy is unnecessary.

leofang:
@kmaehashi this is ready!

kmaehashi:
I think it's better to inject `ctk` tag in `setup.py` as well instead of dynamically editing pyproject.toml during wheel build. PRed to your branch: https://github.com/leofang/cupy/pull/20

leofang:
Merged! I'll update the release tool PR. The injection there is no longer needed.

leofang:
It seems both Windows pipelines timed out, maybe their cache was fully invalidated?

leofang:
@kmaehashi this is ready. I think upon merging the Windows 129/130 failures will go away.

docs/source/install.rst
```diff
@@ -82,6 +84,13 @@ Package names are different depending on your CUDA Toolkit version.
    * - **v13.x** (x86_64 / aarch64)
      - ``pip install cupy-cuda13x``
 
+By default, the above command only installs CuPy itself, assuming a CUDA Toolkit is already installed on the system. To use NVIDIA's CUDA component wheels
+(so as to quickly spinning up a fresh virtual environment without installing a system-wide CUDA Toolkit -- only the CUDA driver is needed -- and allowing
+smaller installation footprint and better interoperability with other Python GPU libraries), you can pass ``[ctk]`` to install them all as
+optional dependencies, e.g.::
+
+   $ pip install cupy-cuda12x[ctk]
```
kmaehashi: nit: adding quote for zsh users (zsh try to expand `[]`)

```suggestion
   $ pip install "cupy-cuda12x[ctk]"
```

docs/source/install.rst
```diff
@@ -307,6 +313,7 @@ If you have installed CUDA on the non-default directory or multiple CUDA version
 
 CuPy uses the first CUDA installation directory found by the following order.
 
+#. ``cuda-pathfinder``'s `documented search order <https://nvidia.github.io/cuda-python/cuda-pathfinder/latest/generated/cuda.pathfinder.load_nvidia_dynamic_lib.html>`_.
```
kmaehashi, line 316: This can be out of the scope of this PR, but this description is slightly inaccurate. In runtime we rely on several aspects of CTK and their discovery strategy is a bit cluttered:

* CTK shared libraries
  * Fully rely on ``cuda-pathfinder``'s documented search order, after this PR get merged
* CTK headers
  1. `$DISCOVERED_CUDA_PATH/include`
  2. NVIDIA Pip packages: `/path/to/site-packages/nvidia/cuda_runtime/include` (for CUDA 12) or `/path/to/site-packages/nvidia/cuda/cu13/include` (for CUDA 13)
* NVCC
  1. `$NVCC` env var
  2. `$DISCOVERED_CUDA_PATH/bin/nvcc`
  3. OS default strategy (`$PATH`)

For CTK headers I think it's better to migrate to `cuda.pathfinder.find_nvidia_header_directory` at some point.

setup.py
```diff
@@ -68,10 +68,41 @@
...
+    ],
+    "test": [
+        "packaging",
+        "pytest>=7.2",
+        "hypothesis>=6.37.2,<6.55.0",
+        "mpmath",
+    ],
+}
+if not ctx.use_hip:
+    dependencies.append("cuda-pathfinder>=1.3.2,<2.0a0")
```
kmaehashi: This is unsafe as using pre-release specifier will allow pip to pick pre-releases even without `--pre`.

```suggestion
    dependencies.append("cuda-pathfinder>=1.3.2,==1.*")
```

https://pip.pypa.io/en/stable/news/#id104:
> `<2.0dev` can include `1.0rc1`

setup.py
```diff
@@ -68,10 +68,41 @@
...
+    if ext_modules:
+        try:
+            cuda_major = ctx.features["cuda"].get_version() // 1000
+        except AssertionError as e:
+            assert 'not configured yet' in str(e)
+        else:
+            optional_dependencies["ctk"] = [
+                f"cuda-toolkit[nvrtc,cublas,cufft,cusolver,cusparse,curand]=={cuda_major}.*"  # NOQA
+            ]
```
kmaehashi, line 97: I think we don't need a try-except here, because it is guarded by `if ext_modules` thus it should always be configured.

```suggestion
        cuda_major = ctx.features["cuda"].get_version() // 1000
        optional_dependencies["ctk"] = [
            f"cuda-toolkit[nvrtc,cublas,cufft,cusolver,cusparse,curand]=={cuda_major}.*"  # NOQA
        ]
```

kmaehashi (commented):
Sorry for the delay in reviewing. Thanks for the hard work Leo!

docs/source/install.rst
leofang: Yes I'll work on the header bits after this is merged.

setup.py
leofang: I had in mind that pre-releases (specifically, `2.0devN`) can be allowed, while there is no plan for this it might come handy if there is ever such needs, but I can be convinced 🙂

setup.py
leofang: No, we need it here because `ctx.features["cuda"].get_version()` could fail in, e.g., nocuda builds:
https://github.com/cupy/cupy/actions/runs/18915803945/job/53998725226#step:6:145
It was observed immediately after merging https://github.com/leofang/cupy/pull/20

leofang:
@kmaehashi PTAL!

setup.py
kmaehashi: I see! Will follow up afterwards.

docs/source/install.rst
leofang: Tracked in https://github.com/cupy/cupy/issues/9501.

leofang:
Many thanks @kmaehashi!

PR #9445, opened 2025-10-27
Add IPCHandle helper class for inter-process memory handling (#8029)

### LINKED ISSUE #8029: Helper class for IPC memory handling?

RamsaiPolisetti:
This PR introduces a new helper class, IPCHandle, that simplifies inter-process communication (IPC) of GPU memory in CuPy.

It allows users to share GPU memory between processes by creating and passing IPC handles safely and conveniently.

This implements the feature discussed in issue #8029.

Key Features

Adds cupy.get_ipc_handle() function to obtain an IPCHandle instance.

Provides IPCHandle.get() to open shared GPU memory in another process.

Prevents invalid reuse of the same handle.

Includes initial test cases under tests/cupy_tests/core_tests/test_ipc_handle.py.

Testing

Added a minimal unit test for IPCHandle creation.

Verified that the handle cannot be reused after .get() is called.

Basic functionality tested locally on CUDA-compatible GPU.

Related Issue

Fixes #8029

Notes

This implementation is a starting point. Future improvements may include:

Enhanced validation of process creation methods (spawn / forkserver)

Error handling for invalid IPC reuse

Integration tests across processes

### DIFF
--- a/cupy/_core/_ipc_handle.py
+++ b/cupy/_core/_ipc_handle.py
@@ -1,0 +1,32 @@
+import cupy
+from cupy.cuda import runtime
+
+class IPCHandle:
+    """Helper class for sharing GPU memory between processes using IPC."""
+
+    def __init__(self, array):
+        self._shape = array.shape
+        self._strides = array.strides
+        self._dtype = array.dtype
+        self._handle = runtime.ipcGetMemHandle(array.data.ptr)
+        self._closed = False
+
+    def get(self):
+        """Open the IPC handle and return a CuPy array."""
+        if self._closed:
+            raise RuntimeError("IPC handle has already been used or closed.")
+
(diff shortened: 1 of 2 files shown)


leofang (changes requested):
Thanks for your contribution!  There is already a WIP in #8245 and it'd be nice to consolidate instead of duplicating the efforts.,

PR #9452, opened 2025-10-29
Fix OOM in slow CI

kmaehashi:
`testing.shaped_arange` utility first creates an array using NumPy. This fixes host OOM triggered in CI due to excessive memory usage.

e.g. https://ci.preferred.jp/cupy.linux.cuda-slow/201785/

xref #8031


asi1024:
The cuda-slow CI is failing with 9 test cases, but I’m merging this PR since the OOM error has been resolved.

Issue #7818, opened 2023-08-23
`ndarray` constructor behavior mismatches when `strides` is given


leofang:
### Description

As title.

### To Reproduce

```py
>>> import numpy as np
>>> a = np.ndarray((1,), strides=(8,), dtype=np.float64)  # this works
>>>
>>> import cupy as cp
>>> a = cp.ndarray((1,), strides=(8,), dtype=cp.float64)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy/_core/core.pyx", line 136, in cupy._core.core.ndarray.__new__
  File "cupy/_core/core.pyx", line 213, in cupy._core.core._ndarray_base._init
ValueError: memptr is required if strides is given.
```

### Installation

Source (`pip install cupy`)

leofang:
Maybe NumPy no longer raises in recent releases, so the old comment https://github.com/cupy/cupy/pull/1611#discussion_r218683200 is no longer applicable?

leofang:
@seberg would you mind triaging this issue?

seberg:
Seems like a very strange feature that I doubt is used.  But allowing this isn't much more than changing the order things so probably might as well change it.

NumPy does check that the memory bounds fit into the allocation if passed, which CuPy is currently not.  We could add that for the Python side constructor (the Cython/internal code paths don't need to verify).

PR #9453, opened 2025-10-29
ENH,BUG: Allow strides without mem, fix empty byte-bounds

seberg:
It ended up a bit more than I thought... on the plus side:
* Speed up creation slightly by avoiding use of Python functions for the bounds calculation.
* Use cython side helper for byte-bounds, which also fixes a bug (NumPy 2 moved this function, but a different thing.)
* Allow allocation if `strides=` is passed, closes gh-7818 (Limited to what NumPy does, whether thought through or not...)

All of these are very small improvements, but since I started, let's finish it I guess.


leofang (approved):

seberg:
Hmmmm, lots of CI timeouts, just chance or real risk of a slowdown?

leofang:
> Hmmmm, lots of CI timeouts, just chance or real risk of a slowdown?

It's fixed by #9460.

asi1024:
I restarted some CIs that resulted in a timeout and confirmed that all tests passed.

Issue #9448, opened 2025-10-28
Remove the experimental tag on `MemoryAsyncPool`


leofang:
I've been asked multiple times in NVIDIA why this thing is still experimental
https://github.com/cupy/cupy/blob/cb44b9bd063051ddd40c94bad3653f52c58e7aa5/cupy/cuda/memory.pyx#L1737-L1738
which has been working fine (and getting better perf) since CUDA 11.2. I don't think there is any reason to keep it experimental. We added it only due to abundant caution.

kmaehashi:
Sounds good!

PR #9455, opened 2025-10-29
DOC: Remove experimental on async allocator

seberg:
Small doc change to remove experimental notes from async allocator and close gh-9448, since that seems to have agreement.

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -26,7 +26,6 @@ from cupy_backends.cuda.api cimport driver
 from cupy_backends.cuda.api cimport runtime
 
 from cupy_backends.cuda.api.runtime import CUDARuntimeError
-from cupy import _util
 
 
 cdef extern from "Python.h":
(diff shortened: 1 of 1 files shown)


leofang:
Thanks for stealing work from me, Sebastian 🙇 

This line can also be removed I think, the rest LGTM!
https://github.com/cupy/cupy/blob/ed2d9f87d9ace2fa9aff3579cafed02bf214acf0/cupy/cuda/memory.pyx#L900

seberg:
> This line can also be removed I think

Ah, I missed that!  Seems like the tests don't care about these FutureWarnings, I guess (I don't see any ignores or similar).

kmaehashi:
`build-rocm` test failure seems like a tentative network error -  https://repo.radeon.com/amdgpu/7.0.2/ubuntu/dists/jammy/ sometimes return 404, possibly in some CDN edges. I've retriggered the workflow run, and now it looks working.

I think it's ok to skip GPU CIs for this one

/test skip

leofang:
> and now it looks working.

It still fails (at the time of this PR was merged). I created an issue to track it: #9456

Issue #9236, opened 2025-07-13
[RFC] Introduce dependencies on CUDA Python projects


leofang:
# Introduction
For many years, CuPy has been the one-stop shop for using NVIDIA CUDA GPUs in Python. It provides many functionalities, way beyond what’s publicly known (“NumPy/SciPy for GPUs”), for example Python bindings to CUDA libraries, JIT compilation (for both CUDA C++ and pythonic kernels, the latter of which also expose Thrust/CUB device APIs), basic access to CUDA runtime, to name a few. The [CuPy docs](https://docs.cupy.dev/en/stable/overview.html) and the [DeepWiki page](https://deepwiki.com/cupy/cupy/) (autogenerated by AI) offer a good overview on what parts of and how CUDA Toolkit are exposed by CuPy.

However, such extensive coverage is no free lunch; it imposes an increasing amount of challenges as both the project scope and its user base grow, in particular in maintaining and evolving of CuPy to keep up with CUDA updates in both software and hardware. Examples include:
- How to use the thread block clusters feature on Hopper+? ([link](https://github.com/cupy/cupy/issues/8780))
- How to make CuPy use CUDA wheels? ([link](https://github.com/cupy/cupy/issues/8013))
- How to expose new CUDA library APIs to CuPy? ([link](https://github.com/cupy/cupy/issues/6359#issuecomment-2425994358))
- How to replace deprecated APIs with new alternatives? ([link](https://github.com/cupy/cupy/issues/9003))
- How to better utilize CUDA graphs? ([link](https://github.com/cupy/cupy/issues/8575))
- How to enable segmented sort in Python? ([link](https://github.com/cupy/cupy/issues/6699))
- …

It is a big burden for a small engineering team plus a handful of OSS volunteers to take on. CuPy is barely keeping up.

Back in 2020, the wider CuPy team (including myself affiliated with neither NVIDIA nor PFN) foresaw some of such challenges, and started a conversation with NVIDIA to discuss the feasibility of NVIDIA taking over the maintenance burden of some components (["the CUDA backend"](https://github.com/cupy/cupy/issues/3597)). Eventually, it evolved and gave birth to the “cuda-python” package (known as [`cuda.bindings`](https://nvidia.github.io/cuda-python/cuda-bindings/latest/) today); however, it was never made as the default backend of CuPy, nor did it solve any major challenges. The CuPy team is still responsible for maintaining all of the component integrations.

[...]

PR #9458, opened 2025-10-29
Adopt `cuda.bindings` new module layout

### LINKED ISSUE #8594: Adopt cuda-python new module layout
We are starting some drastic (but backward compatible) changes in cuda-python and I'd like to get major downstream projects prepared 🙂

One such change is the module layout change as outlined in https://github.com/NVIDIA/cuda-python/issues/75 in order to reserve `cuda` as a root namespace for future NVIDIA Python project. For CuPy, all these mean is that we need to change the `cimport` paths. The existing module paths are deprecated and will be removed in a future release.

This may or may not be blocked by #8457, since Cython 3 has better support for namespace extension modules.

leofang:
Close #8594.


leofang:
CI is green!

leofang:
Yeah we currently pin at 12.x so it is OK 
https://github.com/cupy/cupy/blob/74a61a532a191e33c24131b4a806fbfbdb91445e/.pfnci/matrix.yaml#L432
I will draft a plan for us to use `cuda-bindings` by default in the context of #9236, but I don't think it'd fit in v14.

PR #9454, opened 2025-10-29
Fixed some typos in the documentation

haru-44:

PR #9460, opened 2025-10-30
Revert typo fix in `cupy/_core/include/cupy/complex/complex.h`

kmaehashi:
xref #9454


leofang:
Note: We reverted #9454 partially because by touching the internal (Thrust) headers it invalidated the kernel cache and caused all CI pipelines to time out due to re-compilation. We missed this during code review. In general we want to be very cautious about cache invalidation because the CI costs real money. I will revert this PR when I start working on https://github.com/cupy/cupy/issues/9333.

PR #9462, opened 2025-10-31
Add `numpy.isdtype` support and tests

kiranlahiri:
Add support for numpy.isdtype by importing it from NumPy.

Implements `numpy.isdtype` from #6078

Added tests covering all dtype kinds (bool, integers, floating point, numeric) and edge cases (dtype matching, tuple kinds).

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -395,6 +395,7 @@ from numpy import finfo  # NOQA
 from numpy import iinfo  # NOQA
 
 from numpy import issubdtype  # NOQA
+from numpy import isdtype  # NOQA
 
 from numpy import mintypecode  # NOQA
 from numpy import typename  # NOQA
--- a/tests/cupy_tests/test_type_routines.py
+++ b/tests/cupy_tests/test_type_routines.py
@@ -91,3 +91,73 @@ class TestResultType(unittest.TestCase):
         ret = xp.result_type(input1, input2)
         assert isinstance(ret, numpy.dtype)
         return ret
+
+
+class TestIsDtype(unittest.TestCase):
+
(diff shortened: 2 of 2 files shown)


cupy/__init__.py
```diff
@@ -395,6 +395,7 @@ def result_type(*arrays_and_dtypes):
 from numpy import iinfo  # NOQA
 
 from numpy import issubdtype  # NOQA
+from numpy import isdtype  # NOQA
```
leofang, line 398: Could you add a doc entry to here?
https://github.com/cupy/cupy/blob/6b054938305186c00a058c88764d060d756d355b/docs/source/reference/dtype.rst?plain=1#L39-L45

tests/cupy_tests/test_type_routines.py
leofang, line 1: @asi1024 @seberg can correct me, but it seems we don't add tests for functions re-exported from numpy as-is. For example, we don't have tests for `issubdtype`, `iinfo`, ..., so perhaps we just purge the tests added here?

(mainly the decorators can be improved but I am thinking we shouldn't even sweat here)

seberg: yeah dunno what the old policy is, but to me it makes little sense to test.  _Maybe_ a single test that checks identity with NumPy?

kmaehashi: I think we don't have a test for other re-exported functions. +1 to purge tests added.

leofang:
Hi @kiranlahiri let's address the review and get this PR merged 🙂

leofang:
Closing due to inactivity

PR #9470, opened 2025-11-03
Clean up mempool-related issues

### LINKED ISSUE #5349: MemoryAsyncPool test issues (tracker issue)

leofang:
Close #5349.

- Clean up outdated test skip conditions with the upcoming (v14) CUDA & ROCm support matrices in mind
- Add an adaptor for hooking a `cuda.core.{experimental.}DeviceMemoryResource` with `MemoryAsyncPool`
   - We have been working with the CUDA memory team and exposing more and more knobs for customizing a mempool to Python via `cuda.core`. I suggest CuPy to avoid repeating this work (which is nontrivial) and simply provide an adaptor to interoperate with `cuda.core`. The placeholder `handle == "create"` path is now fully removed.

### DIFF
--- a/cupy/cuda/__init__.py
+++ b/cupy/cuda/__init__.py
@@ -18,6 +18,7 @@ from cupy.cuda import pinned_memory  # NOQA
 from cupy.cuda import profiler  # NOQA
 from cupy.cuda import stream  # NOQA
 from cupy.cuda import texture  # NOQA
+from cupy.cuda.memory_adaptor import cuda_core_device_memory_resource_adaptor  # NOQA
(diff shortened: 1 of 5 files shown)


seberg (approved):
The changes look all good to me. The only thing would be the function name/exact exposure, I think.

(Or maybe whether `MemoryAsyncPoolAdaptor` can just be public with a `__getattr__` rather than a function to create it. But maybe that is a bit unclear, and may also end up as an `AttributeError`, I am not exactly sure.)

PR #9471, opened 2025-11-04
MAINT: Specify texture address precisely

seberg:
I tried using `profile=True` and I think I also ran into this once before (not sure due to what).
In some cases Cython generates code to expose this to Python and then it fails casting to `int*` when compiling the generated C++ code.

Maybe whether it fails compiling is dependent on more things, since it doesn't happen in CI?

### DIFF
--- a/cupy_backends/cuda/api/_runtime_typedef.pxi
+++ b/cupy_backends/cuda/api/_runtime_typedef.pxi
@@ -93,7 +93,7 @@ cdef extern from *:
         MemoryKind kind
 
     ctypedef struct TextureDesc 'cudaTextureDesc':
-        int addressMode[3]
+        TextureAddressMode addressMode[3]
         int filterMode
         int readMode
         int sRGB


cupy_backends/cuda/api/_runtime_typedef.pxi
```diff
@@ -93,7 +93,7 @@ cdef extern from *:
         MemoryKind kind
 
     ctypedef struct TextureDesc 'cudaTextureDesc':
-        int addressMode[3]
+        TextureAddressMode addressMode[3]
```
leofang, line 96: It is puzzling that you ran into issues 🤔 `TextureAddressMode` is just a typedef of `int` (line 30) so this change alone should make no difference as far as Cython is concerned?

seberg: Cython is happy, the C++ compiler wasn't about casting `int*` to `TextureAddressMode*`?

leofang (approved):

PR #9479, opened 2025-11-12
[bug] Include type_traits in filters

gpinkert:
test_filters.py relies on the type_traits library. When JIT compiling code that uses something `std::enable_if` in these tests, users are met with a compilation error saying that `std::enable_if` can't be found and to try using the namespace `__hip_internal::enable_if` instead. This is likely because `std::enable_if` was previously pulled in via some other headers or defined elsewhere in ROCm versions < 7, but is no longer the case.

When printing the failing generated code, it became clear that the type_traits header was not included, which caused the compilation failure.

This is easily fixed by including `type_traits` in the filters JIT strings.


gpinkert:
test_interpolate.py segfaulting is not a new error, it has been observed locally before this change @kmaehashi @asi1024

asi1024:
LGTM!

asi1024 (approved):

PR #9481, opened 2025-11-12
BUG,PERF: Fixup and simplify GUFuncs for better support and speed

seberg:
This reduces the GUfunc overhead of a simple matmul from ~60% of the operation to probably 10% or a bit more. I expect around 1/3 of the time if the matmul core itself is well optimized.

Mainly, also fixes a bunch of bugs, unfortunately... this is rather complex code so it is hard to review.
(I am sure it fixes a lot of bugs more than it opens, but it's hard to be sure everything is covered...)

It does have one larger change (also a bug fix) that I am aware of: The old code incorrectly broadcast _core_ dimensions, this is not allowed by NumPy unless the dimension is specified as `n|1`. For matmul this just changes the error if someone actually relied on it, they would have to switch to `|1` on newer versions.
EDIT: I think we probably changed/fixed that at some point in NumPy.

---

**Honestly, I hate to make this type of large PR, it's practically impossible to review well.  The `matmul` tests proofed actually good (and found that I initially forgot to expand optional dims). I _think_ the tests I added cover a lot, but besides squashing more bugs than adding new ones, it's hard to jduge it.**

Also, I'll say that some things like vector product via `matmul` adds dimensions and so will be slower. This is enough for now to reduce the "setup" overhead, maybe forever, but a fast-path may still make sense.
(But it rubbed me wrong to keep this very messy and slow and just try to side-step it...)

### DIFF
--- a/cupy/_core/_gufuncs.py
+++ b/cupy/_core/_gufuncs.py
@@ -6,6 +6,7 @@ import numpy
 
 import cupy
 import cupy._core._routines_manipulation as _manipulation
+from cupy._core.internal import _normalize_axis_indices
 from cupy._core._dtype import get_dtype, _raise_if_invalid_cast
 from cupy._core import internal
 
@@ -13,7 +14,7 @@ from cupy._core import internal
 # Signature parsing code and dimension accessing has been borrowed
 # from dask
 # https://github.com/dask/dask/blob/61b578f5a3ad88cbc6a8b9a73ce08c551bd969fa/dask/array/gufunc.py#L12-L55
-_DIMENSION_NAME = r'\w+\?*'
+_DIMENSION_NAME = r'\w+(\?|\|1)?'
 _CORE_DIMENSION_LIST = '(?:{0:}(?:,{0:})*,?)?'.format(_DIMENSION_NAME)
 _ARGUMENT = r'\({}\)'.format(_CORE_DIMENSION_LIST)
 _INPUT_ARGUMENTS = '(?:{0:}(?:,{0:})*,?)?'.format(_ARGUMENT)
@@ -25,6 +26,13 @@ _SIGNATURE = '^{:}->{:}$'.format(_INPUT_ARGUMENTS, _OUTPUT_ARGUMENTS)
 
 
 def _parse_gufunc_signature(signature):
+    """Parse and prepare gufunc signature for internal use, we convert
+    each operand into a tuple of dimensions and for each dimension return
+    (name, optional, broadcastable).
+    Additionally, returns the number of unique core dimensions (for later).
+    """
+    coredims = set()
+
     # The code has been modified from dask to support optional dimensions
     if not isinstance(signature, str):
         raise TypeError('Signature is not a string')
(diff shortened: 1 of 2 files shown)


cupy/_core/_gufuncs.py
```diff
@@ -387,12 +316,98 @@ def __init__(self, func, signature, **kwargs):
...
+            # Output axes may be ommitted, fill them in if this is the case.
+            # This seems actually more relaxes as NumPy as of NumPy 2.4.
+            axes.extend([None] * self._nout)
+        elif len(axes) != self._nin + self._nout:
+            raise ValueError(
+                'The number of `axes` entries is not equal to the number of '
+                'input and output arguments. Outputs may only be omitted if '
+                'they are all scalar.')
+
+        return axes[:self._nin], axes[self._nin:], keepdims
```
seberg, line 381: This one is actually a bit similar, but moved wholesale.  The big difference is to not fill to fill in `None` as axes when neither is passed.
That information is useful later and also allows side-stepping the transposes a bit more easily.

cupy/_core/_gufuncs.py
```diff
@@ -387,12 +316,98 @@ def __init__(self, func, signature, **kwargs):
...
+                f'Number of axes passed for argument {i} does not match the '
+                f'the number of core dimensions.')
+
+        axs = _normalize_axis_indices(axs, ndim, sort_axes=False)
+        if axs == tuple(range(ndim-n_core_dims, ndim)):
+            return None  # normalize for no transpose needed
+
+        return tuple(i for i in range(ndim) if i not in axs) + axs
+
+    def _apply_func_to_inputs(self, func, outer_shape, args, outs):
```
seberg, line 405: Basically unchanged, just changed the recursion scheme to chop away `outer_shape`.

cupy/_core/_gufuncs.py
```diff
@@ -405,120 +420,167 @@ def _apply_func_to_inputs(self, func, dim, sizes, dims, args, outs):
...
+                                   for i in transpose[-n_core_dims:])
+
+            n_skipped = len(coredims) - n_core_dims
+            ommitted_coredims = []
+            i_coredim = 0
+            for length in core_shape:
+                # Process the core dimension shape and store the result into
+                # `core_dims` for each dimension.
+
+                # If there are optional coredims skip them from the front
```
seberg, line 487: The old code supported at-front or at-end, this supports both.  The way this works in NumPy is that we skip at the front, which is done here.

(Could probably support mixing `n?` and `n` properly, I don't think it is but doens't matter.)

cupy/_core/_gufuncs.py
```diff
@@ -405,120 +420,167 @@ def _apply_func_to_inputs(self, func, dim, sizes, dims, args, outs):
...
+
+            # The above may not have processed all optional dimensions, do it
+            # to ensure errors for missing ones (and simplify output shape).
+            while i_coredim < len(coredims):
+                ommitted_coredims.append(i_coredim)
+                self._update_dims(i, core_dims, coredims[i_coredim], -1)
+                i_coredim += 1
+
+            if ommitted_coredims:
+                args[i] = cupy.expand_dims(arg, axis=tuple(ommitted_coredims))
```
seberg, line 506: `expand_dims` here should be well tested by `matmul`.  `expand_dims` sets output dimensions, which fits to the `i_coredim` counter above.
(On this branch, the number of core-dimensions must be 0.)

cupy/_core/_gufuncs.py
```diff
@@ -405,120 +420,167 @@ def _apply_func_to_inputs(self, func, dim, sizes, dims, args, outs):
...
-        return args, dimsizess, loop_output_dims, outs, missing_dims
+                _raise_if_invalid_cast(
+                    ret_dtype, out.dtype, casting, "out dtype")
+            else:
+                if trans is not None:
+                    itrans = sorted(range(len(trans)), key=lambda i: trans[i])
+                    shape = tuple(shape[i] for i in itrans)
+                # Note: Order logic here may be weird/wrong as core dims should
+                # be preferred contiguous.
+                out = cupy.empty(shape, dtype=ret_dtype, order=filter_order)
```
seberg, line 571: Old code had some stride filling in for `K` order.  This seemed just wrong so I omitted it.  I am pretty sure we should actually always just use `C` order here or, if anything, try to figure out the order on the transposed args (and really, only for the outer dimensions!  But it's too much for here and probably not vital...).

cupy/_core/_gufuncs.py
```diff
@@ -538,7 +600,10 @@ def _determine_order(self, args, order):
         else:
             raise RuntimeError(f'Unknown order {order}')
 
-    def __call__(self, *args, **kwargs):
+    def __call__(
+            self, *args,
+            axes=None, axis=None, keepdims=False, casting='same_kind',
+            dtype=None, signature=None, order='K', out=None):
```
seberg, line 606: I guess some old support for Python 2.x...

seberg:
Hmmm, the profiling above was maybe with few iterations or so (it's tricky to get nice profiles, since this doesn't take long but if you do too many iterations cuda waits?).

This code is probably 2-3 times faster but it's still around 50% of the total `matmul` call, so I am tempted to say this is worth it.
But, to be really comparable to `torch` in this one would require a fast-pathing here _and_ cleaning up the `tensordot_call` code.

EDIt: For completeness, Python side comparison in details

<details>

image

emcastillo:
I am all in for adding a fast path for the most common use cases.

In general the changes LGTM, I initially just ripped the Dask code and hacked my way through it to make it work,
This PR fixes a lot of caveats and makes things much better. Thanks!

emcastillo (approved):
LGTM

leofang:
For some reason the GitHub web UI wouldn't let me check the CI status and merge ability in this PR. Since #9482 also contains the same commit, how about we close this PR and merge #9482 instead?

seberg:
Funny, but sure. I just split it out because the fast-path could be applied on it's own if we are worried about this.

Issue #8191, opened 2024-02-16
Higher Kernel Launch CPU Overhead


rajagond:
I have observed a higher CPU launch overhead when using CuPy's matrix multiplication compared to PyTorch. While the GPU computation is almost similar for both, the CPU overhead for launching the kernel in CuPy appears to be larger. 

### To Reproduce
```py
import cupy as cp
import time

def matmul_cpu_overhead_cupy():
    BL = 32 * 2048
    hidden_size = 12288
    inter_size = (4 * hidden_size) // 2
    warmup_iters = 50
    active_iters = 250
    cp.cuda.Device(1).use()
    input = cp.random.rand(BL, inter_size).astype(cp.float16)
    weights = cp.random.rand(inter_size, hidden_size).astype(cp.float16)
    output = cp.zeros((BL, hidden_size), dtype=cp.float16)
    start_event = cp.cuda.Event(disable_timing=False)
    end_event = cp.cuda.Event(disable_timing=False)
    cp.cuda.Stream.null.synchronize()  
    # Warm-up  
    for _ in range(warmup_iters):
        cp.matmul(input, weights, out=output) 
    
    # Measure the launch overhead  
    start_time = time.time()
    start_event.record()
    for _ in range(active_iters):
        cp.matmul(input, weights, out=output)
    end_event.record()
    end_time = time.time()  
    
    end_event.synchronize()  
    gpu_time = cp.cuda.get_elapsed_time(start_event, end_event) / active_iters  # in milliseconds
    cpu_time = (end_time - start_time) / active_iters
    total_time = cpu_time * 1000.0  + gpu_time  
    print(f"Cupy: Total time including GPU computation: {total_time} milliseconds")
    
    # CPU launch overhead time  
    cpu_launch_overhead = cpu_time * 1000.0
    print(f"Cupy: CPU launch overhead: {cpu_launch_overhead} milliseconds")

def matmul_cpu_overhead_torch():
    import torch  
    import time  
    
    device = torch.device("cuda:1")  
    torch.cuda.set_device(device)
    BL = 32 * 2048
    hidden_size = 12288
    inter_size = (4 * hidden_size) // 2
    warmup_iters = 50
    active_iters = 250
    input = torch.randn(BL, inter_size, device=device, dtype=torch.bfloat16)
    weights = torch.randn(inter_size, hidden_size, device=device, dtype=torch.bfloat16)
    output = torch.zeros((BL, hidden_size), device=device, dtype=torch.bfloat16)
[...]

PR #9482, opened 2025-11-13
Add fast-path for gufunc (specifically matmul) 

seberg:
Since simple `cupy.matmul` calls, half of the overhead is still inside the gufuncs even with the previous PR, this suggests adding a fast-path mechanism.

I would love to avoid the fast-path here, but I doubt it is worthwhile to carefully Cythonize the gufuncs and a 2-3x speeup is just still a bit underwhelming.
(Fast-paths are always an opportunity for bugs after all...)

The core is to propose a mechanism of:
```
def try_fastcall(*args, **kwargs):
     if not supports_kwargs_and_args:
         return NotImplemented
     # do the real thing if possible
```
for "GUFuncs".

---

**~Draft as based on gh-9481 and I assume we can put this since @emcastillo already had a look.~** whoops, seems I forgot the draft...

This maybe closes gh-8191, from my timings we are still at about 2x slower than `torch`, though.  That is lost in various places, the biggest chunk is probably a `rollaxis` (creating a new array).

### DIFF
--- a/cupy/_core/_gufuncs.py
+++ b/cupy/_core/_gufuncs.py
@@ -259,6 +259,10 @@ class _GUFunc:
         supports_out (bool, optional):
             If the wrapped function supports out as one of its kwargs.
             Defaults to `False`.
+        try_fastcall_func (None, optional):
+            If not None this function is called with same arguments as the
+            gufunc itself.  Must return the final result or ``NotImplemented``
+            if the fast-path is not possible.
         signatures (list of tuple of str):
             Contains strings in the form of 'ii->i' with i being the char of a
             dtype. Each element of the list is a tuple with the string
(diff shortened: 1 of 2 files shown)


emcastillo (approved):
LGTM!

leofang:
With two 4x4 matrices we can see this PR clearly shows significant reduction of the host overhead:
- commit 5652098:
```python
In [10]: cupyx.profiler.benchmark(cp.matmul, (a, b), {}, n_warmup=1, n_repeat=100000)
Out[10]: matmul              :    CPU:    16.795 us   +/-  1.019 (min:    15.959 / max:    95.660) us     GPU-0:    22.484 us   +/-  1.088 (min:    20.480 / max:   100.352) us
```
- commit 02ec96b (from #9481):
```python
In [7]: cupyx.profiler.benchmark(cp.matmul, (a, b), {}, n_warmup=1, n_repeat=100000)
Out[7]: matmul              :    CPU:    34.240 us   +/-  1.566 (min:    32.781 / max:   115.801) us     GPU-0:    38.415 us   +/-  1.725 (min:    36.832 / max:   119.936) us
```
- commit d36e5afb (latest main):
```python
In [6]: cupyx.profiler.benchmark(cp.matmul, (a, b), {}, n_warmup=1, n_repeat=100000)
Out[6]: matmul              :    CPU:    60.918 us   +/-  2.131 (min:    59.191 / max:   163.941) us     GPU-0:    65.124 us   +/-  2.260 (min:    62.624 / max:   173.024) us
```

PR #9484, opened 2025-11-13
`cupyx.scipy.signal.{cspline1d_eval,qspline1d_eval}` throw exception for empty cj array

steppi:
This PR ports over https://github.com/scipy/scipy/pull/22992 from SciPy, adding explicit guards for `cj` being empty in `cspline1d_eval` and `qspline1d_eval`, and raising a `ValueError` with descriptive message if `cj` is empty. Without the guards, `cspline1d_eval` and `qspline1d_eval` will get locked in endless recursion and hit the recursion limit when `cj` is empty.

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_bsplines.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_bsplines.py
@@ -75,12 +75,13 @@ class TestQSpline:
         return scp.signal.qspline1d(xp.asarray([1., 2, 3, 4, 5]))
 
 
-@testing.with_requires('scipy')
 class TestCSplineEval:
+    @testing.with_requires('scipy')
     @testing.numpy_cupy_allclose(scipy_name='scp')
     def test_cspline_eval_zero(self, xp, scp):
         return scp.signal.cspline1d_eval(xp.asarray([0., 0]), xp.asarray([0.]))
 
+    @testing.with_requires('scipy')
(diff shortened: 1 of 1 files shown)


leofang:
@steppi could you check the linter errors plz? 🙂

steppi:
> @steppi could you check the linter errors plz? 🙂

Oops. Will do. I usually rely on Emacs to lint things as I edit, but my CuPy dev setup involves working on a remote GPU workstation, and I haven't figured out how to get lsp-mode to work remotely for Python dev yet. I used to use something called Elpy instead of lsp-mode, but it's no longer maintained. If I'm not able to get this working, I'm probably going to have to switch to vscode or something.

In the meantime, I'll make a habit of running `ruff check` before committing anything.

asi1024:
@steppi Could you rewrite this test so that it follows the same structure as `TestSplineFilter::test_spline_filter_lambda_zero`? We want to verify that both scipy and cupyx.scipy raise the same error type.

seberg:
/test mini

Using parametrization now and testing both scipy/cupy, seems OK to me.  Thanks @steppi, will put it in (if tests pass).

seberg (approved):

seberg:
@steppi we need to also skip the test on older SciPy versions that have the bug :/.

steppi:
> @steppi we need to also skip the test on older SciPy versions that have the bug :/.

Sorry, right. Skips added now.

PR #9486, opened 2025-11-17
BUG: Fix incorrect initialization in bspline kernel

seberg:
This fixes the use of uninitialized memory in the bspline kernel `d_boor` kernel.
Testing is unfortunately not really possible directly, this leads to occasionally test failures.

I did test it locally, by replacing all ``empty`` calls with either filling with NaN or the minimum integer, though.

Closes gh-7240

---

I am not really worried about it being wrong (the original code always uses uninitialized data here), but maybe @andfoy you can have a brief look? In some paths it seems it only matters that this value isn't NaN or similar (i.e. probably it leads to `value - value` which is mathematically be zero but isn't for NaN/inf).
But I think the design-matrix tests wanted one here?

### DIFF
--- a/.github/workflows/self-hosted-ci.yml
+++ b/.github/workflows/self-hosted-ci.yml
@@ -18,8 +18,9 @@ jobs:
           ( github.event_name == 'issue_comment' && github.event.issue.pull_request ) ||
           github.event_name == 'pull_request_review_comment'
         ) && (
+          github.event.comment.author_association == 'OWNER' ||
           github.event.comment.author_association == 'MEMBER' ||
-          github.event.comment.author_association == 'OWNER'
+          contains('|gpinkert|', github.event.comment.user.login)
         ) && (
           contains(github.event.comment.body, '/test ')
         )
--- a/cupy/_core/_dtype.pxd
+++ b/cupy/_core/_dtype.pxd
@@ -2,8 +2,9 @@ from cpython cimport Py_buffer
 
 
 cdef str all_type_chars
+cdef bint check_supported_dtype(dtype, bint error) except -1
 cpdef get_dtype(t)
-cpdef tuple get_dtype_with_itemsize(t)
+cpdef tuple get_dtype_with_itemsize(t, bint check_support)
 cpdef int to_cuda_dtype(dtype, bint is_half_allowed=*) except -1
 
 cpdef void _raise_if_invalid_cast(
@@ -14,3 +15,13 @@ cpdef void _raise_if_invalid_cast(
 ) except *
 
 cdef void populate_format(Py_buffer* buf, str dtype) except*
+
+
+cdef inline normalize_dtype(dtype):
+    """Given an existing NumPy dtype normalize it for cupy.
+
+    Right now, this is a small helper to ensure little-endian.
+    Use this when converting from NumPy/CPU arrays where cupy helps the
+    user by byte-swapping automatically.
+    """
+    return dtype.newbyteorder("<")
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -31,6 +31,42 @@ cdef dict _dtype_dict = {}
 cdef _dtype = numpy.dtype
 
 
+cdef bint check_supported_dtype(dtype, bint error) except -1:
+    """ Returns true on success but otherwise raises an error. """
+    cdef str dtype_char = dtype.char
+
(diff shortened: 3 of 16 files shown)


asi1024:
@seberg I think it is not necessary to initialize `h[1]`. In the first iteration of the outer loop: L115-L132 (when `j = 1`), `h[1]` is assigned at L130 without being read up to that point.

- the inner loop - `for (int p = 0; p < j; p++) { hh[p] = h[p]; }` does not be executed, so `h[1]` is not referenced.
- the inner loop - `for (n = 1; n <= j; n++) { ... }` assigns `h[1]` at `h[n] = w * (xp - xa);` when `n = 1`.

seberg:
> @seberg I think it is not necessary to initialize h[1].

I admit, I haven't followed the logic so I am very much not sure that this makes sense. By now, I think there is plausibly a deeper issue (although maybe this is still partly correct, dunno).

Here is the observation:
```
diff --git a/cupyx/scipy/interpolate/_bspline.py b/cupyx/scipy/interpolate/_bspline.py
index a220fb1a8..4b120ef5c 100644
--- a/cupyx/scipy/interpolate/_bspline.py
+++ b/cupyx/scipy/interpolate/_bspline.py
@@ -265,7 +266,7 @@ def _evaluate_spline(t, c, k, xp, nu, extrapolate, out):
 
     # Compute interpolation
     num_c = int(np.prod(c.shape[1:]))
-    temp = cupy.empty(xp.shape[0] * (2 * k + 1))
+    temp = cupy.full(xp.shape[0] * (2 * k + 1), fill_value=cupy.nan)
     d_boor_kernel = _get_module_func(D_BOOR_MODULE, 'd_boor', c)
     d_boor_kernel(((xp.shape[0] + 128 - 1) // 128,), (128,),
                   (t, c, k, nu, xp, intervals, out, temp, num_c, 1,
```
That diff will _always_ trigger the test failure from CI (i.e. last value is NaN) in the test.
So far so good.  Initializing `h[1]` to almost any value fixes that part.  Somewhere we must be effectively doing `h[1] - h[1]` without initializing it, but where?!  It could even be in the linear interpolation at the end I guess.

However, setting `h[1]` to any value except 1 (e.g. 0) fails:
```
FAILED cupyx_tests/scipy_tests/interpolate_tests/test_bspline.py::TestBSpline::test_design_matrix_same_as_BSpline_call[_param_0_{c=[-1, 2, 0, -1], extrapolate=True}] - AssertionError: 
```
The result ends up as `[1, val, val, ...]` rather than `[1, 1, 1, ...]`?

That part didn't raise my warning flags enough yesterday.  But it seems strange/like a memory issue?!  Is there some memory corruption or execution order madness that causes `h[1] = 0.0/NaN` to fail that test but pre-initializing the temporary array to `NaN` is OK?

seberg:
@asi1024 OK, I think I found the actual issue, I guess... This one looks like a clear typo and makes a lot more sense.

I won't claim to have understood why this issue leads to the above observations or the algorithm, but it's clear that `x[n]` must be filled in...

seberg:
We should wait a bit for @ev-br, we have been chatting a bit and he is looking at the SciPy code.  SciPy has the same code although it (recently) started to zero the input.

But... let's see maybe @ev-br finds something more here.  It still looks like a typo to me, but it sounds like there may be a bit more complexity here.

seberg:
OK, I think we agreed to (and Evgeni found a bit more to back it up), that deBoors formula has a funny "use zero when it would be divide by zero" which is the reason for the path, so that the fix is right.
(It _is_ still surprising that this error seems very subtle since the test suite doesn't notice a thing here for example!)

Evgeni figures the reason why this is that it is practically impossible to hit this path to begin with: The test is very strange.

ev-br:
To summarize, I indeed believe this is the right fix. A typo (mu<->n) led to using uninitialized memory _in some cases_.
The typo originates from scipy, where it traces back to the original commit which introduced the `de_boor_d` function, in 2007. 
The spirit of the fix is indeed to enforce the "result is zero if denominator is zero" caveat for the recursive definition of b-splines.
Why tests do not catch it otherwise: it is rather difficult to hit this code path. The flaky test which caught it in CuPy has several features which helped catching it (mostly by chance): for a quadratic spline (k=2), it has a triple knot, and there are only two boundary knots next to it.

asi1024:
@seberg Many thanks for looking into the flaky failure! LGTM!

PR #9488, opened 2025-11-17
BUG,TST: Fix `cupy_tests/core_tests` runs for `pytest-run-parallel`

seberg:
Thought a small PR is maybe good, this is **start** to test with [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel).  The `cupy_tests/core_tests` (mostly?) run with this.
I did free-threading earlier, but for now it doesn't really matter, these are just threading problems in the tests.

Some notes:
* Must use `python` 3.14, because the `mock` things (mostly?) work threadsafe there (also warning filters).
* Avoid global `rng` state (`_helper`) fix.
* SetUp/TearDown are not thread-safe, but for some cases `setup_class`/`teardown_class` seems OK and good enough for `pytest-run-parallel`.
  * In general, have to avoid state on the test class now.
* Some tests just can't work, start marking them with `@pytest.mark.thread_unsafe`

---

For future reference, I ran this ASAN (parts with TSAN at an earlier point) with:
```
CUPY_TEST_GPU_LIMIT=8 ASAN_OPTIONS=detect_leaks=0,protect_shadow_gap=0 pytest "cupy_tests/core_tests" -n8 --parallel-threads=20 -s -vv -k "not TestRawJitify and not TestRawJitifyJitify"
```
(Jitify tests are not just slow, I was using ASAN and that ran into issues with any failed test. -- in practice mostly run without any `test_raw.py`.)

### DIFF
--- a/cupy/_core/_routines_indexing.pyx
+++ b/cupy/_core/_routines_indexing.pyx
@@ -543,6 +543,12 @@ _nonzero_kernel = ElementwiseKernel(
 
 
 _take_kernel_core = '''
+using idx_t = decltype(a)::index_t;
+idx_t index_range = static_cast<idx_t>(index_range_);
+idx_t ldim = static_cast<idx_t>(ldim_);
+idx_t cdim = static_cast<idx_t>(cdim_);
+idx_t rdim = static_cast<idx_t>(rdim_);
+
 ptrdiff_t out_i = indices % index_range;
 if (out_i < 0) out_i += index_range;
 if (ldim != 1) out_i += (i / (cdim * rdim)) * index_range;
@@ -552,33 +558,43 @@ out = a[out_i];
 
 
 _take_kernel = ElementwiseKernel(
-    'raw T a, S indices, uint32 ldim, uint32 cdim, uint32 rdim, '
-    'int64 index_range',
+    'raw T a, S indices, int64 ldim_, int64 cdim_, int64 rdim_, '
+    'int64 index_range_',
     'T out', _take_kernel_core, 'cupy_take')
 
 
 _take_kernel_scalar = ElementwiseKernel(
-    'raw T a, int64 indices, uint32 ldim, uint32 cdim, uint32 rdim, '
-    'int64 index_range',
+    'raw T a, int64 indices, int64 ldim_, int64 cdim_, int64 rdim_, '
+    'int64 index_range_',
     'T out', _take_kernel_core, 'cupy_take_scalar')
 
 
 _choose_kernel = ElementwiseKernel(
-    'S a, raw T choices, int32 n_channel',
+    'S a, raw T choices, int64 n_channel',
     'T y',
-    'y = choices[i + n_channel * a]',
+    '''
+      using idx_t = decltype(choices)::index_t;
+      y = choices[i + static_cast<idx_t>(n_channel) * static_cast<idx_t>(a)];
+    ''',
     'cupy_choose')
 
 
 _choose_clip_kernel = ElementwiseKernel(
-    'S a, raw T choices, int32 n_channel, int32 n',
+    'S a, raw T choices, int64 n_channel_, int64 n_',
     'T y',
     '''
-      S x = a;
+      using idx_t = decltype(choices)::index_t;
+      idx_t n_channel = static_cast<idx_t>(n_channel_);
+      idx_t n = static_cast<idx_t>(n_);
+
+      idx_t x;
       if (a < 0) {
         x = 0;
       } else if (a >= n) {
         x = n - 1;
+      }
+      else {
+        x = static_cast<idx_t>(a);
       }
       y = choices[i + n_channel * x];
     ''',
@@ -589,12 +605,14 @@ cdef _put_raise_kernel = ElementwiseKernel(
     'S ind, raw T vals, int64 n_vals, int64 n',
     'raw U data, raw bool err',
     '''
+      using v_idx_t = decltype(vals)::index_t;
+
       ptrdiff_t ind_ = ind;
       if (!(-n <= ind_ && ind_ < n)) {
         err[0] = 1;
       } else {
         if (ind_ < 0) ind_ += n;
-        data[ind_] = (U)(vals[i % n_vals]);
+        data[ind_] = (U)(vals[i % static_cast<v_idx_t>(n_vals)]);
       }
     ''',
     'cupy_put_raise')
@@ -604,10 +622,12 @@ cdef _put_wrap_kernel = ElementwiseKernel(
     'S ind, raw T vals, int64 n_vals, int64 n',
     'raw U data',
     '''
+      using v_idx_t = decltype(vals)::index_t;
+
       ptrdiff_t ind_ = ind;
       ind_ %= n;
       if (ind_ < 0) ind_ += n;
-      data[ind_] = (U)(vals[i % n_vals]);
+      data[ind_] = (U)(vals[i % static_cast<v_idx_t>(n_vals)]);
     ''',
     'cupy_put_wrap')
 
@@ -616,28 +636,35 @@ cdef _put_clip_kernel = ElementwiseKernel(
     'S ind, raw T vals, int64 n_vals, int64 n',
     'raw U data',
     '''
+      using v_idx_t = decltype(vals)::index_t;
+
       ptrdiff_t ind_ = ind;
       if (ind_ < 0) {
         ind_ = 0;
       } else if (ind_ >= n) {
         ind_ = n - 1;
       }
-      data[ind_] = (U)(vals[i % n_vals]);
+      data[ind_] = (U)(vals[i % static_cast<v_idx_t>(n_vals)]);
     ''',
     'cupy_put_clip')
 
 
 cdef _create_scatter_kernel(name, code):
     return ElementwiseKernel(
-        'T v, S indices, int32 cdim, int32 rdim, int32 adim',
+        'T v, S indices, int64 cdim_, int64 rdim_, int64 adim_',
         'raw T a',
         string.Template('''
-            S wrap_indices = indices % adim;
+            using i_idx_t = decltype(i);
+            using a_idx_t = decltype(a)::index_t;
+            i_idx_t cdim = static_cast<i_idx_t>(cdim_);
+            a_idx_t rdim = static_cast<a_idx_t>(rdim_);
+            a_idx_t adim = static_cast<a_idx_t>(adim_);
+
+            a_idx_t wrap_indices = indices % adim;
             if (wrap_indices < 0) wrap_indices += adim;
-            ptrdiff_t li = i / (rdim * cdim);
-            ptrdiff_t ri = i % rdim;
+            a_idx_t li = i / (rdim * cdim);
+            a_idx_t ri = i % rdim;
             T &out0 = a[(li * adim + wrap_indices) * rdim + ri];
-            T &in0 = out0;
             const T &in1 = v;
             ${code};
         ''').substitute(code=code),
--- a/cupy/_core/include/cupy/carray.cuh
+++ b/cupy/_core/include/cupy/carray.cuh
@@ -619,6 +619,7 @@ private:
 
 public:
   static const int ndim = 0;
+  typedef typename cupy::type_traits::conditional<_use_32bit_indexing, int, ptrdiff_t>::type index_t;
 
   __device__ CArray() : data_(NULL), size_(1) { }
   
--- a/cupy/_creation/basic.py
+++ b/cupy/_creation/basic.py
@@ -68,7 +68,7 @@ def _new_like_order_and_strides(
 
 def empty_like(
         prototype: NDArray[Any],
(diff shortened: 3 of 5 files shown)


tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -403,44 +403,45 @@ class _TestRawBase:
...
+        cls.mod2 = cupy.RawModule(
             code=_test_source2,
-            backend=self.backend, jitify=self.jitify)
-        self.mod3 = cupy.RawModule(
+            backend=cls.backend, jitify=cls.jitify)
+        cls.mod3 = cupy.RawModule(
             code=_test_source3,
             options=('-DPRECISION=2',),
-            backend=self.backend, jitify=self.jitify)
+            backend=cls.backend, jitify=cls.jitify)
```
seberg: Code is now run only once, but these modules/kernels are only used in a single test each I think, likely that was always the intention (I being surprised `setUp` runs before every test).

tests/cupy_tests/core_tests/test_userkernel.py
```diff
@@ -339,7 +337,7 @@ def test_texture_input(self):
         width, height, depth = self.dimensions
         dim = 3 if depth != 0 else 2 if height != 0 else 1
 
-        texobj = self._prep_texture()
+        texobj, data, shape = self._prep_texture()
```
seberg, line 340: Avoid state on the test class/instance which can race.

tests/cupy_tests/core_tests/test_carray.py
```diff
@@ -57,6 +58,7 @@ def test_getitem_idx(self):
     {'size': 2 ** 32},
     {'size': 2 ** 32 + 1024},
 )
+@pytest.mark.thread_unsafe(reason="too large allocations")
```
seberg, line 61: Will need a lot of these sprinkled in eventually (although, maybe even more `setUp()` changes).

tests/conftest.py
```diff
@@ -39,6 +39,17 @@ def pytest_configure(config):
...
+        # There is an unclear issue somewhere that causes tests fail in random
+        # ways (usually while doing a host-to-device copy from a NumPy array).
+        # Doing this once on the main thread before spawning new threads seems
+        # to work around it.
+        # `pytest-run-parallel` (if using parallel runs) runs into this very
+        # reliable for some tests when run with multiple threads.
+        # (Added 2025-12, in the future try to remove this again!)
+        import cupy as cp
+
+        cp.array([0])
```
seberg: This is now the annoying work-around I found after way too many tries... With that I think all of the `cupy_tests` (not slow) actually pass wth `pytest-run-parallel` (with or without free-threaded).
(EDIT: `cupy_tests/core_tests`, I guess I hadn't run the fully `cupy_tests`)

Of course, nothing is ever 100%, the more threads the hihgher the chance to find issues (and right now we know there are dead-locks in the allocator that I might have seen over many runs).

Running in CI I have to still think about, running most tests in many threads is likely too slow for us here.

EDIT: By now I have seen this failure even with this... so just very rare or so...

**EDIT: This is wrong, I'll fix it tomorrow probably, maybe in a separate PR. The issue was _probably_ the use of `defaultdict` in the memory pool (this may be fixed in Python, but there is no big reason for a defaultdict)**

cupy/cuda/memory.pyx
```diff
@@ -1603,8 +1602,10 @@ cdef class MemoryPool:
     def __init__(self, allocator=None):
         if allocator is None:
             allocator = _malloc
-        self._pools = collections.defaultdict(
-            lambda: SingleDeviceMemoryPool(allocator))
+
+        dev_counts = runtime.getDeviceCount()
+        self._pools = tuple(
+            SingleDeviceMemoryPool(allocator) for i in range(dev_counts))
```
seberg: This will probably fixed in Python, but doesn't hurt even now. It is the one true bug-fix here (except on future 3.14.x versions presumably).

It seems my very long odyssey was caused by `defaultdict`. The mechanism is that two threads can inintialize a defaultdict at the same time.
In that case, one thread will get a `SingleDeviceMemoryPool` that is immediately discarded again. Howver, `SingleDEviceMemoryPool` owns the actual allocations currently which causes havoc later.
(And at least often not all that obvious havoc...)

seberg:
This has grown a bit. I think I may have occasional hangs (I am hoping this is due to the known memory allocator deadlock), and nothing is 100% (testing is slow, but naturally we are testing for race conditions to begin with).

But, this now makes `cupy_tests/core_tests` pass with free-threading. I do have further fixes beyond this, but I think this should go in first (some of this may be complicated enough).

leofang:
@seberg let's fix the pre-test failure and kick off CI!

cupy/cuda/memory.pyx
seberg: Ahh... the `defaultdict` was of course there for a reason. The default `MemoryPool` get's created on import, but we can't do `getDeviceCount()` or other `SingleDeviceMemoryPool` setup at that point :(.
So we need some locking while accessing/creating `_pools`.

That was also the reason why I thought the `pytest-xdist` hook didn't work right...

seberg:
OK, the fact that instantiating has to be deferred makes this bit more tedious (I actually needed a `critical_section` in free-threaded).
But let's give it a shot now...

/test mini

cupy/cuda/memory.pxd
```diff
@@ -58,7 +58,8 @@ cpdef get_allocator()
 cdef class MemoryPool:
 
     cdef:
-        object _pools
+        cdef tuple _pools
+        cdef object _allocator
```
seberg:
```suggestion
       tuple _pools
       object _allocator
```
oops, can also fix this up in the mempool PR (which should have some small conflicts with this).

tests/cupy_tests/core_tests/test_carray.py
```diff
@@ -65,13 +67,15 @@ class TestCArray32BitBoundary(unittest.TestCase):
     # This test requires approx. 8 GiB GPU memory to run.
     # See https://github.com/cupy/cupy/pull/882 for detailed discussions.
 
-    def tearDown(self):
-        # Free huge memory for slow test
+    @classmethod
+    def teardown_class(self):
```
leofang: Q: Regarding this and all other setup/teardown changes, shouldn't we use `setup_method` and `teardown_method` instead so as to preserve the semantics of `setUp`/`tearDown`?

seberg: Hmmm, I have to review much of these it seems. `setUp` and `setup_method` should behave much the same way I think, but I thought these were called like a decorator.

I.e. I thought these were called per-thread, thus causing race conditions. But it seems that isn't actually the case, the problems must be caused by the tests mutating `self` only.
Which means, I should be able to undo many of these changes.

tests/cupy_tests/core_tests/test_carray.py
```diff
@@ -49,39 +49,40 @@ def test_getitem_idx(self):
...
+    2 ** 31 - 1024,
+    2 ** 31,
+    2 ** 31 + 1024,
+    2 ** 32 - 1024,
+    2 ** 32,
+    2 ** 32 + 1024
+])
+@pytest.mark.slow
+@pytest.mark.thread_unsafe(reason="too large allocations")
+class TestCArray32BitBoundary:
```
seberg, line 62: This one shouldn't need any change except the new marker, but I like the idea of using pytest, so kept it, it seems simple enough.

(It seems that something is disrupting `-n8` `xdist` with `CUPY_TEST_GPU_LIMIT=8` after all, which was confusing me.)

tests/cupy_tests/core_tests/test_core.py
```diff
@@ -117,15 +118,14 @@ def test_cupy_likes_and_nested(self, cupy_like):
 @testing.parameterize(*testing.product({
     'cxx': (None, '--std=c++17'),
 }))
-class TestCuPyHeaders(unittest.TestCase):
-
-    def setUp(self):
+class TestCuPyHeaders:
+    def setup_method(self):
```
seberg, line 119: I think `setup_method` is the nicer spelling. The issue is that it sometimes clashes with some custom `@testing.<...>` decorators.

tests/cupy_tests/core_tests/test_raw.py
```diff
@@ -1312,14 +1318,7 @@ def test_raw_picklable(self):
 # Recent CCCL has made Jitify cold-launch very slow, see the discussion
 # starting https://github.com/cupy/cupy/pull/8899#issuecomment-2613022424.
 # TODO(leofang): Further refactor the test suite?
-@testing.parameterize(*testing.product({
-    'jitify': (False, True),
-}))
-@unittest.skipIf(cupy.cuda.runtime.is_hip,
-                 'Jitify does not support ROCm/HIP')
-@testing.slow
-class TestRawJitify(unittest.TestCase):
```
seberg, line 1321: The reason for splitting this is that only one of them needs skipping (the jitify branch fails with multiple threads)

tests/cupy_tests/core_tests/test_carray.py
seberg: OK, I think I reverted some (but changed this specific one to get rid of `unittest` and some custom `testing.parametrize` stuff anyway in the end).

It seems that `setup_method` is racy came from discussions in NumPy, but for now I'll assume it isn't (i.e. runs exactly once before/after the threaded tests) -- the same should be true for `setUp` and `tearDown`, so I can probably undo some things in follow-up PRs as well.

tests/cupy_tests/core_tests/test_cub_reduction.py
```diff
@@ -15,17 +17,18 @@
 # This test class and its children below only test if CUB backend can be used
 # or not; they don't verify its correctness as it's already extensively covered
 # by existing tests
-@unittest.skipIf(_environment.get_cub_path() is None, 'CUB not found')
 class CubReductionTestBase(unittest.TestCase):
```
seberg, line 20: Admittedly, moving the skip is a left-over of trying to get rid of `unnittest`. Unfortunately, that turned out a bit annoying because it would probably have required modifying the later `@testing.for_contiguous_axes()` (which means touching more files).

tests/cupy_tests/core_tests/test_cub_reduction.py
```diff
@@ -15,17 +17,18 @@
...
     """
     Note: call self.can_use() when arrays are already allocated, otherwise
     call self._test_can_use().
     """
 
     def setUp(self):
```
leofang, line 26: So no `setup_method` here?

cupy/cuda/memory.pxd
```diff
@@ -70,6 +71,19 @@ cdef class MemoryPool:
     cpdef set_limit(self, size=?, fraction=?)
     cpdef size_t get_limit(self)
 
+    cdef _ensure_pools_and_return_device_pool(self)
+
+    cdef inline device_pool(self):
+        # This criticial section should strictly speaking be necessary since
+        # `self._pools` could in theory be ub an inconsistent state
+        # (not fully written itself as the `_pools` should be finished).
+        # Although, on the systems we care about that is probably not possible.
```
leofang: I can't parse this sentence... Do you mean the critical section is absolutely needed?

cupy/cuda/memory.pxd
```diff
@@ -70,6 +71,19 @@ cdef class MemoryPool:
...
+        with cython.critical_section(self):
+            if self._pools is not None:
+                return self._pools[device.get_device_id()]
+
+        return self._ensure_pools_and_return_device_pool()
```
leofang, line 85: it seems we acquire a mutex again here? (leaving the critical section above, and then enter again when calling this function) Would it cause a slowdown? I wonder if we could `return` in the same section.

cupy/cuda/memory.pxd
seberg: Ah, yeah that's mangled. I mean it is probably not needed in practice, will update. (for this type of things having clearer atomics might be nice.)

cupy/cuda/memory.pxd
seberg: Yeah, but this is the one time initialization path, the happy path is always the first (which is also why I put that code in its own function).

tests/cupy_tests/core_tests/test_cub_reduction.py
seberg: Right, that would require some further cleanups here, unfortunately, due to how the cupy decorators work with unittest vs. non-unittest classes.

seberg:
Triggering merge for @kmaehashi: The test changes are pretty straight forward if slight churn.
Otherwise, it is a pretty straight forward replacement of the memory pools per-device `defaultdict` with a tuple (and the required logic to set it up on first access).

Ah, nvm. I pushed a small docs change, need an approval of course still :).

leofang (approved):

leofang:
Linux CIs timed out because the CI caches haven't fully rebuilt. Let's just get it done on the main/v14 branches.

PR #9494, opened 2025-11-21
ENH: `ml_dtypes.bfloat16` support

seberg:
Add optional support for `bfloat16` via `ml_dtypes`.  The mechanism for this is:
* We the dtype coming from `ml_dtypes` and if that can be imported, add it to our C type maps.  (The corresponding scalars are supported now via an earlier refactor.)
* We add a new header and type. This header is only included if `bfloat16` is used in any of the kernel arguments.
* Add a `*bf16_loop()` helper, which expands to the loop if `ml_dtypes` can be found, otherwise the loop is omitted.  (This should be fine, in part because there will always be an "e" loop in front, so we don't have to worry much about weirder promotion corner cases.)
* Attempts to add tests in CI (not sure I got this right).

---

I think this should be considered *experimental*, in the sense that corner cases/differences probably exist and it may not always be obvious what the computation type should be for `bfloat16` input (or e.g. half input is explicitly using float32 and it is missing for now).

### DIFF
--- a/tests/cupy_tests/core_tests/test_bfloat16.py
+++ b/tests/cupy_tests/core_tests/test_bfloat16.py
@@ -79,7 +79,10 @@ def test_i0(xp):
 def test_angle(xp):
     # XXX(seberg): CuPy doesn't handle NaN correctly, so don't test.
     a = xp.array(TEST_VALUES[:-1], dtype=BF16)
-    return xp.angle(a, deg=False)
+    res = xp.angle(a, deg=False)
+    if xp == numpy:
+        res = res.astype(BF16)  # some versions of NumPy return float32
+    return res
 
 
 @pytest.mark.parametrize('func', [


leofang:
FYI #9503 is merged

seberg:
Now that the scalar changes are merged, this makes sense to briefly look at to discuss the approach, I think.  The changes are now rather minimal (many lines just being matrix multiple support).
* For the first try, I shoehorned the `#include <cuda_bf16.h>` into the `complex` header, which doesn't really make sense. Although, I am not actually sure where to best put it.
  * If we don't want to always include, in my string-start, I tagged additional headers on to the dtype `get_typename()`, which may be an option.
    (I think that makes sense long-term, but not sure it is worthwhile short-term.)
  * ROCm is unhappy about this unsurprisingly. I guess there is likely a simple alternative (or we have to make this conditional for now), although not sure what :).
* I am not actually sure that `__nv_bfloat16` is the right C++ type to use.
* This allows registring a new loop with a list of dtypes:
  ```
        E = numpy.dtype(ml_dtypes.bfloat16)
        EE_E = ([(E, E), (E,)],)
  ```
  rather than strings. But maybe we have a better idea.  (For bfloat16 itself, I think it is OK to conditionally insert the loop if we are OK with checking if `ml_dtypes` exists at import time.)

cupy/_core/_routines_linalg.pyx
```diff
@@ -607,13 +607,13 @@ cpdef _ndarray_base tensordot_core(
         c = c.view()
         c.shape = (n, m)
 
-    if dtype not in 'efdFD':
+    if dtype.kind in 'biu':
```
leofang, line 615: Q: what's `biu`?

leofang: and how do we avoid (or allow?) bf16 here?

cupy/_core/_routines_math.pyx
```diff
@@ -866,11 +866,20 @@ cdef create_arithmetic(
...
+        # Name unfortunately collides with complex32 a bit.
+        E = numpy.dtype(ml_dtypes.bfloat16)
+        EE_E = ([(E, E), (E,)],)
+
     return create_ufunc(
         'cupy_' + name,
         (('??->?', boolop),
          'bb->b', 'BB->B', 'hh->h', 'HH->H', 'ii->i', 'II->I', 'll->l',
-         'LL->L', 'qq->q', 'QQ->Q', 'ee->e', 'ff->f', 'dd->d', 'FF->F',
+         'LL->L', 'qq->q', 'QQ->Q', 'ee->e', *EE_E, 'ff->f', 'dd->d', 'FF->F',
```
leofang: Q: What does the asterisk do?

cupy/_core/include/cupy/complex.cuh
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cupy/complex/complex.h>
+#include <cuda_bf16.h>
```
leofang: Q: Why do we need this?

btw this would trigger another cache rebuild... although if the intention is justified it should not be a blocker

cupy/_core/_routines_linalg.pyx
seberg: `biu` is boolean, integer, or unsigned kind. This swaps the logic from the inexact/float types to the integer ones.
(Of course if we add narrow integers from `ml_dtypes` this might also need adjusting again.)

cupy/_core/_routines_math.pyx
seberg: `EE_E` is a tuple, so that it can be an empty tuple if `ml_dtypes` is not available. I.e. it's a way to deal with the optional availability.

cupy/_core/include/cupy/complex.cuh
seberg: Yeah, not sure this is the right header, but I do need something:
```
/tmp/tmpciaqhscv/241011ceb72572e9df829b80bbd57247465ca771.cubin.cu(16): error: incomplete type "const in0_type" is not allowed
          const in0_type in0(_raw_in0[_ind.get()]);
                         ^

/tmp/tmpciaqhscv/241011ceb72572e9df829b80bbd57247465ca771.cubin.cu(17): error: incomplete type "const in1_type" is not allowed
  const in1_type in1(_raw_in1[_ind.get()]);
                 ^

/tmp/tmpciaqhscv/241011ceb72572e9df829b80bbd57247465ca771.cubin.cu(18): error: incomplete type "out0_type" (aka "__nv_bfloat16") is not allowed
  out0_type out0;
            ^

3 errors detected in the compilation of "/tmp/tmpciaqhscv/241011ceb72572e9df829b80bbd57247465ca771.cubin.cu".
```

cupy/_core/_routines_math.pyx
leofang: Oh gosh it's not even a string, I just realized. Thanks!

cupy/_core/include/cupy/complex.cuh
seberg: OK, trying to get a bit more coverage, I think we'll likely want a few small header additions.
While that could be a done in `carray.h` (as it is for `float16`), I am really tempted to go the `get_typename_and_preamble()` route that I did for the string dtype PoC a long time ago.

For ufuncs+ElementWiseKernel we add additional headers from the dtypes if we find those dtypes.
There will be a few places that may need to do this explicitly (or fail JIT compiling on types that need it), but I doubt that it is a huge amount.

seberg:
This is now in a state where a look make sense. Not small anymore, but also not _that_ big.

One larger change is that there is now a `bfloat16.cuh` and it is only included if `bfloat16` is involved in the kernel, that is rather than `_get_typename()` to get the C type, we now need `_get_typename_and_preamble()` to also know that we need to include `bfloat16.cuh`.

Some open points/homework for me:
* [x] New headers need some thoughts (I don't like the way float16 <-> bfloat16 casts work but it may be OK)
* [x] A few small bugs (some matmul code fails).
* [x] Need to fix fusion support (if not too complicated), including special value support (e.g. `inf`).
* [x] It may be nice to see if we can avoid rebuilding caches too much.

*EDIT: I won't claim there won't be holes, but I think I covered all of those points :tada:.  The one thing would probably still do is rename **`param_preambles`** to **`type_headers`** because it is just much clearer (and the name preambles is used elsewhere.*

cupy/_core/include/cupy/bfloat16.cuh
leofang, line 1: Could you remind me the reason that CuPy needs wrappers for the underlying fp16/bf16 types? It's been a while...

cupy/_core/_cub_reduction.pyx
```diff
@@ -208,18 +208,27 @@ __global__ void ${name}(${params}) {
...
 }
 '''
+    type_headers = set()
+    params = _get_cub_kernel_params(params, arginfos, type_headers)
+    type_preambles = type_map.get_typedef_code(type_headers)
+
+    if not type_headers:
+        type_headers = ''
+    else:
+        type_headers = '\n'.join(sorted(type_headers)) + "\n\n"
```
leofang, line 218: Q: Why do we need to sort the type declarations? What if they have dependency and need to be inserted in order?

cupy/_core/_fusion_kernel.pyx
```diff
@@ -100,7 +102,12 @@ cdef class FusedKernel:
         for i, op in enumerate(op_list):
             if i > 0:
                 codes.append('_cg::sync(_grid);')
-            codes.append(op.emit_code())
+            codes.append(op.emit_code(type_headers))
+
+        if type_headers:
+            # Insert parameter/type specific headers
+            submodule_code = (
+                "\n".join(sorted(type_headers)) + "\n\n" + submodule_code)
```
leofang, line 110: ditto

cupy/_core/_kernel.pyx
```diff
@@ -1516,23 +1528,31 @@ cdef class _Op:
...
-            str typ, object routine, object error_func):
+            typ, object routine, object error_func):
         # TODO(niboshi): Write type mapping specification.
-        types = typ.split('->')
-        if len(types) == 1:
-            in_types = out_types = tuple(types)
+        if isinstance(typ, list):
+            # TODO: A bit of a hack for now, use dtype(...).type to normalize.
+            in_types = tuple(numpy.dtype(t).type for t in typ[0])
+            out_types = tuple(numpy.dtype(t).type for t in typ[1])
```
leofang: Q: We don't want to catch bf16 in `_dtype.pyx` and just call `get_dtype()` here?

tests/cupy_tests/core_tests/test_bfloat16.py
leofang, line 1: This can wait, but it'd be nice later to add `no_bfloat16=False` to `for_all_dtypes`, `for_float_dtypes`, ... in `cupy/testing/_loops.py`  to get it auto-tested

tests/cupy_tests/core_tests/test_bfloat16.py
```diff
@@ -0,0 +1,145 @@
...
+])
+@testing.numpy_cupy_allclose(rtol=TOL, atol=TOL)
+def test_matmul(xp, shapes):
+    shape_a, shape_b = shapes
+    a = testing.shaped_arange(shape_a, xp, dtype=BF16)
+    b = testing.shaped_arange(shape_b, xp, dtype=BF16)
+    res = xp.matmul(a, b)
+    if xp == numpy:
+        # NumPy returns float32 we don't
+        res = res.astype(BF16)
```
leofang, line 163: So we return bf16 here? 👍  cc @samaid for vis

tests/cupy_tests/core_tests/test_bfloat16.py
```diff
@@ -0,0 +1,145 @@
...
+        return getattr(xp, func)(a, axis=1)
+
+
+@pytest.mark.parametrize('func', ['sum', 'prod', 'min', 'max', 'mean'])
+@pytest.mark.parametrize('data', [
+    numpy.arange(1, 10_000),
+    numpy.random.uniform(0.5, 2, size=10_000),
+])
+@testing.numpy_cupy_allclose(rtol=TOL, atol=TOL)
+def test_reductions_large(xp, func, data):
```
leofang, line 181: I am wondering which path this takes, JIT (`_cub_reduction.pyx`) or AOT (`cub.pyx`)? I don't see any code change to the latter so I assume this is all JIT? Seems like a happy path (that we don't need to increase pre-compiled kernels).

cupy/_core/include/cupy/bfloat16.cuh
seberg: In part, some of the ufuncs are just written that way. The generated code also relies on `=` assignment casts working.

The assignment is likely the most annoying, other stuff would probably just be slightly different ufunc code.

Actually, it may well be that the assignment doesn't really matter if we didn't use a custom `float16` (or one that inherited from cuda headers directly?).

So, I guess it's a hmmm... Maybe we _could_ get away with using just the cuda header and `nv_bfloat16` by tweaking some other things.

cupy/_core/_cub_reduction.pyx
seberg: I figure parameters shouldn't have a order anyway, and wanted a `set()` (unordered) to not duplicate them.

cupy/_core/_kernel.pyx
seberg: Good spot, yeah this can (and should) just use the identical code from below.

seberg:
One other thing to confirm here: I am not yet sure the `bfloat16` tests are running in any of the CI jobs.

cupy/_core/_routines_math.pyx
```diff
@@ -917,10 +921,10 @@ _angle = create_ufunc(
 
 _angle_deg = create_ufunc(
     'cupy_angle_deg',
-    ('?->d', 'e->e', 'f->f', 'd->d',
+    ('?->d', 'e->e', *bf16_loop(), 'f->f', 'd->d',
      ('F->f', 'out0 = arg(in0) * (180.0 / M_PI)'),
      ('D->d', 'out0 = arg(in0) * (180.0 / M_PI)')),
-    'out0 = in0 >= 0 ? 0 : 180.0',
+    'out0 = in0 >= decltype(in0){0} ? 0 : 180.0',
```
seberg: Hmmm, I need to undo (or change this), but it needs some header fix as well (it thinks the `__half` comparison may also match).

FWIW, I will also check other CI testing `bfloat16` though, I suspect I added `ml_dtypes` to the matrix but have to add it to actually be installed maybe?

tests/cupy_tests/core_tests/test_bfloat16.py
seberg: I think it all goes through `_cub_reduction` or `_reduction`, I'll add one simple reduction to the fusion examples (that was missing).  (Or the statistical routines for some of them.)

I am not sure if there may be paths where we should have `cub.pyx` usage right now.

cupy/_core/_routines_math.pyx
seberg: In the end, I made this two lines, but really just used `decltype(+in0){0}`. The `+` seemed like the easiest way to fix the fusion, which uses e.g. `double &in0` and we need to get rid of the `&`.

I couldn't think of a way to keep this code unchanged that would avoid the operator to the header, but I didn't want to add a full family of them...

(As an aside: `type_in0_raw` mechanism isn't compatible with fusion, maybe fusion could also define it. Although I like the `decltype` at least if it wasn't for the `+in0` hack.)

cupy/_core/include/cupy/bfloat16.cuh
leofang: I think it's fine to keep this structure.

tests/cupy_tests/core_tests/test_bfloat16.py
leofang: Ah nvm. That was a dumb question because I get rusty in the way we cover CUB. We do checks like this to decide if we can dispatch to the pre-compiled kernels
https://github.com/cupy/cupy/blob/0362252fe60b82efc7f6d38715a131a8bfa7b00a/cupy/cuda/cub.pyx#L420
and since we do not touch `cub.pyx` in this PR (nor should we), the dispatch will fallback to the JIT path. Resolving.

cupy/_core/_routines_math.pyx
```diff
@@ -911,7 +911,11 @@ _angle = create_ufunc(
     ('?->d', 'e->e', *bf16_loop(), 'f->f', 'd->d',
      ('F->f', 'out0 = arg(in0)'),
      ('D->d', 'out0 = arg(in0)')),
-    'out0 = in0 >= decltype(in0){0} ? 0 : M_PI',
+    '''
+    // For bfloat16 compare within type. Unary + strips ref (for fusion).
+    const decltype(+in0) zero = 0;
```
leofang: Not a blocker, but it seems like we could use `cuda::std::remove_reference` here?

seberg: Nice thought.
```
using cmp_type = cuda::std::remove_reference(decltype(in0))::type;
```
seems to at least require some additional includes, though.

seberg: Hmmm, I forgot a bunch of functions, and some more use this pattern. I should re-think whether the header definitions are quite as they should be :/.

seberg:
OK, an unfortunately big update, so this is becoming more a bit more unwieldy:
* I forgot functions (quite a few), so added those. Which flushed out that the old approach wasn't good.
  * Mostly straight-forward, though. Just `nextafter` needed a custom implementation, and the `float16` version is slightly off.
* I added explicit conversions in each direction to the type right now.  Guarded with an `enable_if` so that it will be delayed (i.e. you don't need both headers).  Maybe there is a better way?
* I got rid of implicit conversions, and instead added an `operator=` template that uses the *explicit* conversion.  That way cast kernels that use `arr[i] = val` can convert between the two.
* Talking with Leo, `float16` now uses the identical pattern and has its own file. That means float16 *isn't* normally included!
  * Unfortunately, I hadn't covered the fusion fully (I skipped the old-fusion), for now I just injected the new header clumsily there.  I guess that should be pretty safe for starters.
  * I still put forward declarations into the `carray.h` for now, not sure that is best. The atomic version (and conversion) are guarded with an `enable_if` templating, so that the compiler doesn't complain if `float16` isn't actually fully defined.
    (Is there a non-templating trick for this?!)

**At least core tests are (mostly?) fine, but let's see. Open issues:**
* Before the last changes, tests were failing on CUDA 12, maybe that is even OK to say bfloat16 just fails there, but unless my header changes helped that may be easiest.
* ROCm is untested and I have no idea how much it might work
* This made me noticed that `atomicAdd` is needed for scatter ops, which would be failing here for bfloat16.

EDIT: Nvm. the code is probably almost good, but there is at least an error in the `atomic` part.  Either way I'll look at it more tomorrow.

seberg:
OK, I think I have fixed the templating, I want to see if I can clean up the fusion code a bit and maybe add a test for the atomicAdd (although, maybe not, in the end we should expand the template to include bfloat16 probably).

But, I am really curious about ROCm even without all that, so:

/test rocm

seberg:
/test rocm

Just an aside, I _think_ things work now **except** `cupyx/jit`, I haven't figured out how to insert the float16 header there as a hot-fix for making the header optional in most contexts... 

EDIT: Found the right place, just not pushing because I want to see the ROCm result and not potentially abort it.

cupy/_core/include/cupy/bfloat16.cuh
```diff
@@ -0,0 +1,162 @@
...
+#define __nv_bfloat16 __hip_bfloat16
+
+#else
+
+#include <cuda_bf16.h>
+
+#endif
+
+
+// Wrapper class that inherits from __nv_bfloat16 to get all operators
```
seberg:
```suggestion
```
oops wrong, now.

cupy/_core/include/cupy/bfloat16.cuh
```diff
@@ -0,0 +1,153 @@
+#pragma once
+
+#include "cupy/carray.cuh"
+
+#if defined(__HIPCC_RTC__) || defined(__HIPCC__)
+
+#include <hip/hip_bf16.h>
+#define __nv_bfloat16 __hip_bfloat16
```
seberg, line 8: @gpinkert hopefully you have a quick insight. CI doesn't like this due to:
```
2026-01-23T09:50:40.8561021Z E           /opt/rocm-7.0.2/include/hip/amd_detail/amd_hip_bf16.h:674:66: error: redefinition of default argument
2026-01-23T09:50:40.8564699Z E             674 |                                                        const int width = warpSize) {
2026-01-23T09:50:40.8567455Z E                 |                                                                  ^       ~~~~~~~~
2026-01-23T09:50:40.8572025Z E           /root/.local/lib/python3.11/site-packages/cupy/_core/include/cupy/hip_workaround.cuh:9:49: note: expanded from macro '__shfl_down_sync'
2026-01-23T09:50:40.8575587Z E               9 | #define __shfl_down_sync(mask, ...) __shfl_down(__VA_ARGS__)
2026-01-23T09:50:40.8576889Z E                 |                                                 ^~~~~~~~~~~
2026-01-23T09:50:40.8578524Z E           /opt/rocm-7.0.2/include/hip/amd_detail/amd_hip_bf16.h:649:57: note: previous definition is here
2026-01-23T09:50:40.8580334Z E             649 |                            unsigned int lane_delta, int width = warpSize) {
2026-01-23T09:50:40.8581698Z E                 |                                                         ^       ~~~~~~~~
2026-01-23T09:50:40.8583438Z E           /opt/rocm-7.0.2/include/hip/amd_detail/amd_hip_bf16.h:691:61: error: use of undeclared identifier 'mask'
2026-01-23T09:50:40.8585496Z E             691 |   u.ui = __shfl_down_sync<unsigned long long, unsigned int>(mask, u.ui, delta, width);
2026-01-23T09:50:40.8587009Z E                 |                                                             ^
2026-01-23T09:50:40.8588680Z E           /opt/rocm-7.0.2/include/hip/amd_detail/amd_hip_bf16.h:701:61: error: redefinition of default argument
2026-01-23T09:50:40.8590464Z E             701 |                                                   const int width = warpSize) {
```

Likely we need to apply a small fix to `hip_workaround.cuh`?!

(Of course also happy if this is an XY problem and the include/pattern here is just not the right/best approach.)

seberg: OK, I figured out that newer ROCm just defines these, we probably got away with just redefining it before (because why not), while now the `bfloat16` header extends those functions which then fails.

Tested it with the csr tests in CI and this seemed OK.

It would be nice if you can briefly check `hip_workaround.cuh` if it looks right.  But I think we should be fine here.

seberg: There is one other caveat.  The `matmul` (with matrices) ~fails~ crashed in the HIP path, but works with cublas locally (although newer version than in the branch, so I am not 100% sure!).

It may just be that older versions don't support bf16, or the bf16+f32 compute type incantation?

I had to skip the test for now because of this.

seberg:
Want to see if this works (or if I might have to adapt the minimum version in the tests).

/test cuda122,linux

tests/cupy_tests/core_tests/test_bfloat16.py
```diff
@@ -0,0 +1,231 @@
...
+
+import numpy
+import pytest
+
+import cupy
+from cupy.cuda import runtime
+from cupy import testing
+
+
+if not runtime.is_hip and runtime.runtimeGetVersion() < 12020:
```
leofang, line 11: We can fix this in the next PR, not a blocker. But it's better to use either NVRTC version or `cupy.cuda.get_local_runtime_version()`; `runtimeGetVersion()` does not represent the version of either the compiler or the headers.

cupy/_core/include/cupy/carray.cuh
```diff
@@ -44,178 +49,20 @@ namespace cupy {
...
-        ret_raw_.x = x_raw_.x + 1;
-      }
-    } else if(!(y_raw_.x & 0x8000u) || (x_raw_.x & 0x7fffu) > (y_raw_.x & 0x7fffu)) {
-      ret_raw_.x = x_raw_.x - 1;
-    } else {
-      ret_raw_.x = x_raw_.x + 1;
-    }
-    return float16(ret_raw_);
-  }
+#endif  // #ifdef __HIPCC_RTC__
```
leofang, line 59: Comment only, no need to address in this PR: Shouldn't this be `__HIPCC__`?

leofang (approved):

asi1024:
LGTM! Thanks!

tests/cupy_tests/core_tests/test_bfloat16.py
leofang: Addressed in #9660.

PR #9498, opened 2025-11-24
MAINT,PERF: Add fast-path and avoid errors in operators/array-ufunc

seberg:
This re-organizes things very slightly to avoid unnecessary attribute checking and errors. In the first case, just avoid an actual error, in the second check for cupy arrays, because otherwise we build the full `__array_interface__` unnecessarily there.

Randomly checked a profile for `arr + 1` and this saves 10-20% or so. (Profiling on this machine is very unreliable.)
I admit `arr + 1` should likely side-step `__array_ufunc__` completely, but after this it isn't a big problem (for now).

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -398,7 +398,7 @@ configs {
   value {
     requirement {
       cpu: 8
-      memory: 50
+      memory: 60
       disk: 10
       gpu: 1
     }
--- a/.pfnci/coverage.rst
+++ b/.pfnci/coverage.rst
@@ -809,7 +809,7 @@ CuPy CI Test Coverage
      - 
    * - nccl
      - null
-     - 1
+     - 3
      - 
      - 
      - 
@@ -830,8 +830,8 @@ CuPy CI Test Coverage
      - 
      - 
      - 
-     - 
-     - 
+     - ✅
+     - ✅
      - ✅
      - 
      - 
@@ -1129,7 +1129,12 @@ CuPy CI Test Coverage
      - 
    * - 
      - 2.27
-     - 7
+     - 3
+     - 
+     - 
+     - 
+     - 
+     - 
      - 
      - 
      - 
@@ -1150,20 +1155,13 @@ CuPy CI Test Coverage
      - 
      - ✅
      - ✅
-     - 
-     - 
-     - 
-     - ✅
-     - ✅
      - ✅
      - 
-     - ✅
-     - ✅
+     - 
+     - 
    * - 
      - 2.28
-     - 2
-     - 
-     - 
+     - 4
      - 
      - 
      - 
@@ -1191,6 +1189,8 @@ CuPy CI Test Coverage
      - 
      - 
      - 
+     - ✅
+     - ✅
    * - cutensor
      - null
      - 5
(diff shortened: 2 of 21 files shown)


cupy/_core/core.pyx
```diff
@@ -1826,11 +1827,12 @@ cdef class _ndarray_base:
...
                 # Except for numpy.ndarray, types should be supported by
                 # `_kernel._preprocess_args`.
-                check = (hasattr(x, '__cuda_array_interface__')
-                         or hasattr(x, '__cupy_get_ndarray__'))
-                if (not check
-                        and not type(x) in _scalar.scalar_type_set
-                        and not isinstance(x, numpy.ndarray)):
+                if (
+                    not type(x) in _scalar.scalar_type_set
+                    and not isinstance(x, (_ndarray_base, numpy.ndarray))
```
leofang, line 1832: Q: Is it expected that we add `_ndarray_base` here? Do we need to special-casing for CuPy arrays themselves?

cupy/_core/core.pyx
```diff
@@ -97,19 +101,16 @@ CUPY_CACHE_KEY = cupy_cache_key.decode()
...
-    except AttributeError:
-        # NEP 13's recommendation is `return False`.
-        xp = getattr(x, '__array_priority__', 0)
-        yp = getattr(y, '__array_priority__', 0)
-        return xp < yp
-    return y_ufunc is None
+    # _HANDLED_TYPES is just a random private object as "singleton".
+    y_ufunc = getattr(y, "__array_ufunc__", _HANDLED_TYPES)
+    if y_ufunc is not _HANDLED_TYPES:
+        return y_ufunc is None
```
leofang, line 107: This is interesting... I would think that try-except is faster than `getattr`!

cupy/_core/core.pyx
leofang: Ah, missed the PR description:
> in the second check for cupy arrays, because otherwise we build the full __array_interface__ unnecessarily there.

leofang (approved):

cupy/_core/core.pyx
seberg: try/except relies on an exception object being created and set. `getattr` often/usually avoids that on failure. (On success I am not sure, but I would not expect much of a difference.)

leofang: Ah ok, for some reason I was thinking `__array_ufunc__` is an attribute/property, but it's actually a method (obviously). 

FWIW it is still slower by a tiny margin:
```python
In [4]: %timeit getattr(a, "__array_ufunc__")
53 ns ± 0.0876 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
vs
```python
In [6]: %%timeit
   ...: try:
   ...:     a.__array_ufunc__
   ...: except:
   ...:     pass
   ...: 
   ...: 
37.4 ns ± 0.186 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

leofang:
Thanks!

cupy/_core/core.pyx
seberg: Ahhh, good point! So this is actually a trade-off. Because it is a method, if `__array_ufunc__` is defined we lose some time because we have to bind it.
If it doesn't exist, the `try/except` would add a lot more overhead due to the exception.

I suppose I'll wager that this is still better since we don't go here for CuPy `ndarray`'s.

leofang: right, I wouldn't sweat for 15 ns 🙂

PR #9503, opened 2025-11-26
MAINT,ENH: Simplify `CScalar` handling and ready it for arbitrary dtypes

seberg:
This tries to simplify the scalar handling. In part just for maintenance and a small speed boost, but largely to make it easier to support arbitrary dtypes in the scalar code-path (i.e. split out from the `ml_dtypes` draft PR).

The simplification ideas ideas are:
1. Rely on NumPy C-API to convert Python scalars to C data that we can use for kernel launches.
2. Simplify handling by pushing NEP 50 "weak scalar" handling into the `CScalar` and removing the distinction between "numpy scalar/CScalar" path.
3. Make CScalar simply hold the original scalar

The other change is making NumPy a *build-time* dependency.

**I'll add some comments in-line, so discussions are easier to focus on each topic.**

### DIFF
--- a/install/cupy_builder/cupy_setup_build.py
+++ b/install/cupy_builder/cupy_setup_build.py
@@ -345,7 +345,8 @@ def make_extensions(ctx: Context, compiler, use_cython):
     settings['define_macros'].append(('CYTHON_EXTERN_C', 'extern "C"'))
 
     # We use NumPy 2.x only C-API, so need to define this:
-    settings['define_macros'].append(('NPY_TARGET_VERSION', 'NPY_2_0_API_VERSION'))
+    settings['define_macros'].append(
+        ('NPY_TARGET_VERSION', 'NPY_2_0_API_VERSION'))
 
     if ctx.linetrace:
         settings['define_macros'].append(('CYTHON_TRACE', '1'))


tests/cupyx_tests/jit_tests/test_cooperative_groups.py
```diff
@@ -115,7 +115,7 @@ def test_grid(x):
             g.sync()  # this should just work!
 
         x = cupy.empty((16,), dtype=cupy.uint64)
-        x[:] = -1
+        x[:] = cupy.int64(-1)  # wrap-around cast to uint
```
seberg, line 118: Maybe clearer to comment here, but this is an actual behavior change.  Previously, we checked for things like `uint8_arr + (-1)` raising an error now in NumPy in `guess_routine`, now it is checked one level lower.

That means that direct elementwise kernel calls now also see the new behavior. This `-1` behavior especially can be tedious, though. It's certainly possible to restore the old behavior if we think it is likely to create hassle.
(Or move it later if we find it does in practice after a release.)

cupy/_core/_scalar.pyx
```diff
@@ -1,32 +1,41 @@
...
+
+    // Defined for NumPy 2.x builds, this define allows a local 1.x build.
+    #ifndef PyDataType_GetArrFuncs
+    static inline PyArray_ArrFuncs *
+    PyDataType_GetArrFuncs(const PyArray_Descr *descr)
+    {
+        return descr->f;
+    }
+    #endif
+    """
```
seberg: I'll confirm that this works for NumPy 1.x locally.  As mentioned, we can simplify this dance to just the `cdef PyArray_Pack()` if we _hard_ require NumPy 2 (i.e. with `NPY_TARGET_VERSION=NPY_2_0_API_VERSION`, NumPy will generate an error if importing with NumPy 1.x).

But, I wasn't sure that we should do that _hard_ yet and while you may have to be me to know that this is all fine, I am me :).

cupy/_core/_scalar.pyx
```diff
@@ -97,249 +96,104 @@ cdef set _numpy_scalar_type_set = set(_typenames.keys())
...
+        # If we ever support dtypes larger than this (e.g. strings)
+        # we will have to introduce a conditional allocation here and
+        # should memset memory to NULL (must if dtype NEEDS_INIT).
+        assert self.descr.itemsize < sizeof(self._data)
+        self.ptr = <void *>self._data  # make sure ptr points to _data.
+
+        # NOTE(seberg): This uses assignment logic, which is very subtly
+        # different from casting by rejecting nan -> int. This is *only*
+        # relevant for `casting="unsafe"` passed to ufuncs with `dtype=`.
+        # It also means we fail for out of bound integers (NEP 50 change).
```
seberg, line 139: Maybe to explain this: This uses the same logic as `arr_with_dtype[0] = value` and that is _exceedingly subtly_ different from a casting.
That truly only matters for things like `cp.add(cp.float64(np.nan), 1, casting="unsafe", dtype=int)` style call. So I think it's safe to ignore :).

cupy/_core/_dtype.pyx
```diff
@@ -61,10 +63,14 @@ cdef bint check_supported_dtype(dtype, bint error) except -1:
...
-    else:
-        raise ValueError(f'Unsupported dtype {dtype}')
+
+    try:
+        _scalar.get_typename(dtype)  # allow if we know a C typename.
+    except (ValueError, KeyError):
+        if not error:
+            return False
+        else:
+            raise ValueError(f'Unsupported dtype {dtype}') from None
```
asi1024, line 74: Should we return `True` when `scalar.get_typename(dtype)` does not raise an exception?

cupy/_core/_dtype.pyx
```diff
@@ -61,10 +63,14 @@ cdef bint check_supported_dtype(dtype, bint error) except -1:
         # possible that the included fields will not be usable in the end.
         # The simplest path is to inform users
         return True
-    elif not error:
-        return False
-    else:
-        raise ValueError(f'Unsupported dtype {dtype}')
+
+    try:
+        _scalar.get_typename(dtype)  # allow if we know a C typename.
```
seberg, line 68:
```suggestion
        _scalar.get_typename(dtype)  # allow if we know a C typename.
        return True
```

cupy/_core/_dtype.pyx
seberg: Yes, good catch. This was prep for `ml_dtypes` and would be wrong for those so we can string them into `get_typename()` and not in many places.

cupy/_core/_dtype.pyx
```diff
@@ -1,5 +1,7 @@
 cimport cython  # NOQA
 
+from . cimport _scalar
```
leofang, line 3: nit: use absolute import

cupy/_core/_dtype.pyx
```diff
@@ -8,6 +10,7 @@ from cupy.exceptions import ComplexWarning
 
 
 cdef str all_type_chars = '?bhilqBHILQefdFD'
+cdef bytes all_type_chars_b = b'?bhilqBHILQefdFD'
```
leofang: nit: use the same source of truth?
```suggestion
cdef bytes all_type_chars_b = all_type_chars.encode()
```

tests/cupyx_tests/jit_tests/test_cooperative_groups.py
leofang: @asi1024 would know this a lot better since he built the whole CuPy JIT machinery, but I think somewhere in the compiler we have a way to change how Python scalars should be interpreted inside a JIT kernel. Maybe it's this:
https://github.com/cupy/cupy/blob/f60edcfdf481fd3c18da7494f7a8d1843d390b53/cupyx/jit/_cuda_typerules.py#L111-L137
My thinking is: Maybe there is a way to leave the CuPy JIT default unchanged by this PR?

cupy/_core/internal.pxd
leofang, line 1: To confirm my understanding: We don't need these fp16 conversion helpers, because they are handled by NumPy out of box?

cupy/_core/_fusion_trace.pyx
```diff
@@ -1,6 +1,7 @@
 import numpy
 
 from cupy._core import _kernel
+from cupy._core import _scalar
```
leofang, line 4: Q: Would it be better if we `cimport _scalar`?

cupy/_core/_kernel.pxd
```diff
@@ -166,6 +166,6 @@ cdef list _get_out_args_with_params(
 
 cpdef _check_peer_access(_ndarray_base arr, int device_id)
 
-cdef tuple _preprocess_args(int dev_id, args, bint use_c_scalar)
+cdef list _preprocess_args(int dev_id, args)
```
leofang, line 169: Q: I thought using `tuple` is slightly faster? Is this needed for mutating the returned sequence?

cupy/_core/_kernel.pyx
```diff
@@ -1305,20 +1294,19 @@ cdef class ufunc:
                         f'Cannot cast array data from {x.dtype!r} to '
                         f'{get_dtype(bool)!r} according to the rule \'safe\'')
             else:
-                # NumPy does not seem raising TypeError.
+                # NumPy does not seem raising TypeError, so just `bool()`.
                 # CuPy does not have to support `where=object()` etc. and
                 # `_preprocess_args` rejects it anyway.
-                where_args[0] = _scalar.CScalar.from_numpy_scalar_with_dtype(
-                    x, numpy.bool_)
+                where_args[0] = _scalar.CScalar(bool(where_args[0].value))
```
leofang: nit:
```suggestion
                where_args[0] = _scalar.CScalar(bool(x.value))
```

cupy/_core/_kernel.pyx
```diff
@@ -1592,14 +1577,33 @@ cdef class _Ops:
...
         if dtype is None:
-            assert all([isinstance(a, (_ndarray_base, numpy.generic))
-                        for a in in_args])
-            in_types = tuple([a.dtype.type for a in in_args])
+            weaks_l = []
+            in_types_l = []
+            for a in in_args:
+                if type(a) is _scalar.CScalar:
+                    # .typeobj is the C-level type (as a PyTypeObject *)
+                    t = <object>(<_scalar.CScalar>a).descr.typeobj
```
leofang: This is a bit nerve-wrecking to me. Is it this
```suggestion
                    # .typeobj is the C-level type (as a PyTypeObject *)
                    t = (<object>(<_scalar.CScalar>a)).descr.typeobj
```
or this?
```suggestion
                    # .typeobj is the C-level type (as a PyTypeObject *)
                    t = <object>((<_scalar.CScalar>a).descr.typeobj)
```

cupy/_core/_kernel.pyx
```diff
@@ -1592,14 +1577,33 @@ cdef class _Ops:
...
+                    weak_t = (<_scalar.CScalar>a).weak_t
+                    if weak_t is not False:
```
leofang, line 1594: Q: At this point, can `weak_t` be None?

cupy/_core/_scalar.pyx
leofang: Q: Remind me what's the conclusion here? That we still allow importing with 1.x, we just wouldn't claim full support for it?

cupy/_core/_scalar.pyx
```diff
@@ -97,249 +96,104 @@ cdef set _numpy_scalar_type_set = set(_typenames.keys())
...
-            ret.kind = b'f'
-            ret.size = 8
-        elif typ is complex:
-            (<double complex*>ret.ptr)[0] = x
-            ret.kind = b'c'
-            ret.size = 16
+    def __init__(self, value, dtype=None):
+        self.value = value
+        if dtype is not None:
+            self.value = value
```
leofang: nit: this is done above
```suggestion
```

cupy/_core/_scalar.pyx
```diff
@@ -97,249 +96,104 @@ cdef set _numpy_scalar_type_set = set(_typenames.keys())
...
+        self.ptr = <void *>self._data
+        (<int32_t *>self.ptr)[0] = value
+        return self
+
+    cdef _store_c_value(self):
+        # If we ever support dtypes larger than this (e.g. strings)
+        # we will have to introduce a conditional allocation here and
+        # should memset memory to NULL (must if dtype NEEDS_INIT).
+        assert self.descr.itemsize < sizeof(self._data)
+        self.ptr = <void *>self._data  # make sure ptr points to _data.
```
leofang: I am panicking about operator precedence 😅 
 ```suggestion
         self.ptr = <void *>(self._data)  # make sure ptr points to _data.
```

File: cupy/_core/_scalar.pyx [hunk 15]
```diff
@@ -97,249 +96,104 @@ cdef set _numpy_scalar_type_set = set(_typenames.keys())
...
-            ret.size = 1
-        elif ret.kind == b'c':
-            (<double complex*>ret.ptr)[0] = x
-            ret.size = 16
+    cdef CScalar from_int32(int32_t value):
+        cdef CScalar self = CScalar.__new__(CScalar)
+        self.value = None
+        self.descr = _numpy_int32
+        self.ptr = <void *>self._data
+        (<int32_t *>self.ptr)[0] = value
```
leofang:
```suggestion
        self.ptr = <void *>(self._data)
        (<int32_t *>(self.ptr))[0] = value
```

cupy/_core/_scalar.pyx
```diff
@@ -97,249 +96,104 @@ cdef set _numpy_scalar_type_set = set(_typenames.keys())
...
-cdef object scalar_to_numpy_scalar(object x):
-    # Converts a Python or NumPy scalar to a NumPy scalar.
-    # Returns None if the argument is not a scalar.
-    typ = type(x)
-    if typ in _python_scalar_type_set:
-        return _python_scalar_to_numpy_scalar(x)
-    elif typ in _numpy_scalar_type_set:
-        return x
-    return None
+        return <object>self.descr.typeobj  # typeobj is the C-level .type
```
leofang:
```suggestion
        return <object>(self.descr.typeobj)  # typeobj is the C-level .type
```

tests/cupyx_tests/jit_tests/test_cooperative_groups.py
seberg: In that case it is unchanged since the JIT kernel discovers a reasonable type here (`int64`) which is also the final kernel type.

The change only occurs for places where the kernel C-type is explicitly a `uint` or narrower int.

cupy/_core/_kernel.pxd
seberg: The function builds a list so I thought it is probably not helpful to create a new object. I guess access might be faster on free-threaded (where lists need locking).

cupy/_core/_fusion_trace.pyx
seberg: Doubt you'll notice the difference, so just stuck with what `_kernel`/others used, but happy to change also.

cupy/_core/_kernel.pyx
seberg: It's never `None`. The `weaks` tuple input could be None previously (but now is folded into the types instead).

cupy/_core/internal.pxd
seberg: Right, we use the NumPy implementation for: `np.empty(1, dtype=float16)[0] = value`, which works for any NumPy dtype.

cupy/_core/_scalar.pyx
seberg: Ah, I never removed this :/! I think we said we don't need <2.0, it still felt a bit strange to me to enforce strictly in this PR. But let me just do this.

It is easy to restore this if there is any doubt about in the end.  (Makes this a lot cleaner, all we'll have left is the Cython define for `PyArray_Pack`).

cupy/_core/_kernel.pyx
leofang: So I assume this is equivalent too?
```suggestion
                    if weak_t is True:
```

leofang:
LGTM! CI is green too. Though Kenichi-san mentioned there is a code freeze now. Let merge it after the freeze and also give @asi1024 a bit more time in case he wants to chime in 🙂

leofang:
@seberg forgot to ask, is there a simple reproducer for us to check the perf difference before and after this PR? Would be nice to know the expected ballpark improvement.

cupy/_core/_kernel.pyx
seberg: `weak_t` is one of: `int, float, complex, False` maybe None rather than False is clearer?

leofang: Ah I see, thx for explanation.

leofang:
> Let merge it after the freeze and also give @asi1024 a bit more time in case he wants to chime in 🙂

Code freeze is lifted. Let me get this merged. @asi1024 please let us know if you have any concern and we can follow up in a separate PR.

> @seberg forgot to ask, is there a simple reproducer for us to check the perf difference before and after this PR? Would be nice to know the expected ballpark improvement.

Would still be nice to keep a record in this PR for future reference, in case people come and ask for what drove the decision of making NumPy a build-time dependency.

seberg:
> is there a simple reproducer for us to check the perf difference before and after this PR? 

Not sure about this, but things like `cp.add(f, f)` (with `f = np.float32(1.))` is e.g. unchanged compared to v13 (`cp.add(1, 1)` maybe very slightly faster).
I think it's basically a wash. The call to NumPy costs a few ns maybe, OTOH, I optimized an allocation away, which may even safe some in the end.

On the grand scheme, the perf differences here are just very small I think.

Issue #9491, opened 2025-11-19
Sparse matrix (cupyx.scipy.sparse.csr_matrix) attribute 'device' shows not implemented error


ninoyrahman:
### Description

When the 'device' attribute for the sparse matrix cupyx.scipy.sparse.csr_matrix is called, NotImplementedError is shown.

### To Reproduce

```py
# Write the code here
```
import cupy as cp
import cupyx.scipy as cps

row  = cp.array([0, 0, 1, 2, 2, 2], dtype=cp.int32)
col   = cp.array([0, 2, 2, 0, 1, 2], dtype=cp.int32)
data = cp.array([1, 2, 3, 4, 5, 6], dtype=cp.float32)
A      = cps.sparse.csr_matrix((data, (row, col)), dtype=cp.float32)
print(A.device)

### Installation

Wheel (`pip install cupy-cuda12x)

### Environment

```
# Paste the output here
```
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
Cell In[8], line 8
      6 data = cp.array([1, 2, 3, 4, 5, 6], dtype=cp.float32)
      7 A = cps.sparse.csr_matrix((data, (row, col)), dtype=cp.float32)
----> 8 print(A.device)

File [~\AppData\Local\Programs\Python\Python311\Lib\site-packages\cupyx\scipy\sparse\_base.py:42](http://localhost:8888/lab/tree/~/AppData/Local/Programs/Python/Python311/Lib/site-packages/cupyx/scipy/sparse/_base.py#line=41), in spmatrix.device(self)
     39 @property
     40 def device(self):
     41     """CUDA device on which this array resides."""
---> 42     raise NotImplementedError

NotImplementedError:

leofang:
It has never been implemented since the debut, so it is not a bug:
https://github.com/cupy/cupy/blob/f05ce6a7b64ad35253a95d6d02e7481a4f3e700d/cupyx/scipy/sparse/_base.py#L43-L46

ninoyrahman:
Hope this feature will be implemented someday.

BarnikRB:
Hello! I have already sent a PR addressing the issue. #9504

PR #9504, opened 2025-11-26
Implemented device property for sparse matrices and necessary changes

BarnikRB:
# Implement `.device` for All CuPy Sparse Matrix Types

## Summary

This PR implements the `.device` property for CuPy sparse matrices.
Previously, accessing `.device` on any sparse matrix raised a `NotImplementedError`, even though this attribute exists for dense `cupy.ndarray`.

---

## Background

All CuPy sparse formats (`csr`, `csc`, `coo`, `dia`, etc.) share the same inheritance structure:

```
spmatrix
   ↓
_data_matrix
   ↓
specific sparse class (csr_matrix, csc_matrix, ...)
```

Key points:

* `_data_matrix` is the first class in the hierarchy where `self.data` is guaranteed to exist.
* `self.data` is always a `cupy.ndarray` stored on the GPU.
* Therefore, `self.data.device` is always valid.
* The abstract base class `spmatrix` declared a `.device` property but did not implement it.
* Since subclasses did not override it, calling `.device` resulted in a `NotImplementedError`.

This prevented users from determining which GPU a sparse matrix resides on.

---

## Fix

This PR implements `.device` inside `_data_matrix`, where the underlying data array is available.

This ensures:

* `.device` works consistently across all sparse matrix formats.
* `.device` returns a proper `cupy.cuda.Device` instance.
* Behavior matches that of `cupy.ndarray.device`.

**Example:**

```python
A = cupyx.scipy.sparse.csr_matrix(...)
print(A.device)   # now prints <CUDA Device 0>
```

---

## Tests

A new test file has been added:

```
cupy/tests/cupyx_tests/scipy_tests/sparse_tests/test_sparse_device.py
```

The tests cover:

* `csr_matrix`
* `csc_matrix`
* `coo_matrix`
* `dia_matrix`

Each test verifies:

* `A.device` returns a valid `cupy.cuda.Device`.
* `A.device == A.data.device`.
* Accessing `.device` does not raise an exception.

---

## Related Issue

Fixes [https://github.com/cupy/cupy/issues/9491](**#<9491>**)

---

### DIFF
--- a/tests/cupyx_tests/scipy_tests/sparse_tests/test_common.py
+++ b/tests/cupyx_tests/scipy_tests/sparse_tests/test_common.py
@@ -4,6 +4,7 @@ import cupy as cp
 from cupy import testing
 import cupyx.scipy.sparse as sp
 
+
 @testing.parameterize(*testing.product({
     'format': ['csr', 'csc', 'coo', 'dia'],
 }))


tests/cupyx_tests/scipy_tests/sparse_tests/test_sparse_device.py
```diff
@@ -0,0 +1,52 @@
...
+    makers = [make_csr, make_csc, make_coo, make_dia]
+
+    for maker in makers:
+        A = maker()
+
+        # device must exist and be a cupy.cuda.Device
+        assert isinstance(A.device, cp.cuda.Device)
+
+        # device must match the underlying data device
+        assert A.device == A.data.device
```
seberg: I think the creation needs to be simplified (maybe check the other files for inspiration) presumably with a parametrization.  And if there is no better file, the file should at least be named more generic, maybe `test_common.py`.

BarnikRB: Thanks for the feedback @seberg! By "creation" are you referring to the creation of the dummy matrices? Should I embed the tests  within the test file for each sparse matrix instead of writing a separate script?

seberg: I think other test files already solve this, you should simplify it to create the matrices inline in a very short way that doesn't special case each format.

BarnikRB: I have modified the tests and renamed it to test_common.py

BarnikRB: @seberg could you please let me know if anything else needs to be modified for this pr?

cupyx/scipy/sparse/_data.py
```diff
@@ -26,6 +26,11 @@ def dtype(self):
         """Data type of the matrix."""
         return self.data.dtype
 
+    @property
+    def device(self):
+        """CUDA device on which this array resides."""
+        return self.data.device
```
leofang, line 32: My take is this is technically not robust enough. In the case of CSR, for example, it has one constructor that takes 3 arrays to create it: data, indptr, indices. We need to check (and cache, for later queries here) that they all reside on the same device. 

All sparse matrices have a number of constructors. Need to hunt them down. Then, we need to measure how much CPU overhead we're adding to the matrix creation.

BarnikRB: Thanks for the comments. If I have understood correctly it is necessary to verify during construction that all the required arrays are in the same gpu and if not we should throw an error?

Issue #7482, opened 2023-03-26
CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered


curtcorum:
### Description

---------------------------------------------------------------------------
CUDARuntimeError                          Traceback (most recent call last)
~/miniconda3/envs/cuda_11_8/lib/python3.9/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

~/miniconda3/envs/cuda_11_8/lib/python3.9/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    392                         if cls is not object \
    393                                 and callable(cls.__dict__.get('__repr__')):
--> 394                             return _repr_pprint(obj, self, cycle)
    395 
    396             return _default_pprint(obj, self, cycle)

~/miniconda3/envs/cuda_11_8/lib/python3.9/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    698     """A pprint that just redirects to the normal repr function."""
    699     # Find newlines and replace them with p.break_()
--> 700     output = repr(obj)
    701     lines = output.splitlines()
    702     with p.group():

cupy/_core/core.pyx in cupy._core.core._ndarray_base.__repr__()

cupy/_core/core.pyx in cupy._core.core._ndarray_base.get()

cupy/cuda/memory.pyx in cupy.cuda.memory.MemoryPointer.copy_to_host()

cupy_backends/cuda/api/runtime.pyx in cupy_backends.cuda.api.runtime.streamIsCapturing()

cupy_backends/cuda/api/runtime.pyx in cupy_backends.cuda.api.runtime.check_status()

CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered

### To Reproduce

```py
# needed for supressing gpu memory leak after previous run, CAC 220505
#%reset -f

import numpy as np
import matplotlib.pyplot as plt

import sigpy as sp
import sigpy.plot as pl

from tkinter import Tk
from tkinter import filedialog as fd
root = Tk() # pointing root to Tk() to use it as Tk() in program.
root.withdraw() # Hides small tkinter window.
root.attributes('-topmost', True) # Opened windows will be active. above all windows despite of selection.

import cupy as cp
import multiprocessing

from tqdm.auto import trange

[...]

PR #9507, opened 2025-11-27
BUG: Make sure indexing routines are OK with huge arrays 

seberg:
Many of these routines used int32 for some shapes internally.
Usually, that works out just fine, since all of these shapes
represent only a part of the array, it is unlikely to be larger
than an int32. But it very much can be larger.

Closes gh-7482

---

~I admit, my env got broken somehow and one of the tests didn't work, but I confirmed it was broken at least.~ (fixed in a new env)

I saw some similar tests that said ROCm doesn't support these large indices?  I wasn't sure if that is still valid, nor am I sure if the `MemoryError` try/except is helpful or doesn't matter.

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -398,7 +398,7 @@ configs {
   value {
     requirement {
       cpu: 8
-      memory: 50
+      memory: 60
       disk: 10
       gpu: 1
     }
--- a/.pfnci/coverage.rst
+++ b/.pfnci/coverage.rst
@@ -809,7 +809,7 @@ CuPy CI Test Coverage
      - 
    * - nccl
      - null
-     - 1
+     - 3
      - 
      - 
      - 
@@ -830,8 +830,8 @@ CuPy CI Test Coverage
      - 
      - 
      - 
-     - 
-     - 
+     - ✅
+     - ✅
      - ✅
      - 
      - 
@@ -1129,7 +1129,12 @@ CuPy CI Test Coverage
      - 
    * - 
      - 2.27
-     - 7
+     - 3
+     - 
+     - 
+     - 
+     - 
+     - 
      - 
      - 
      - 
@@ -1150,20 +1155,13 @@ CuPy CI Test Coverage
      - 
      - ✅
      - ✅
-     - 
-     - 
-     - 
-     - ✅
-     - ✅
      - ✅
      - 
-     - ✅
-     - ✅
+     - 
+     - 
    * - 
      - 2.28
-     - 2
-     - 
-     - 
+     - 4
      - 
      - 
      - 
@@ -1191,6 +1189,8 @@ CuPy CI Test Coverage
      - 
      - 
      - 
+     - ✅
+     - ✅
    * - cutensor
      - null
      - 5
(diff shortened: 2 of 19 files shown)


emcastillo:
I am not sure if the following applies to this, but I think it is worth double checking.

In ancient times, indexing with 64-bit values caused a huge perf degradation in some kernels.
Mostly due to int64 divisions in the indexer.

Probably for these routines, the change is fine, but I'd suggest to run a benchmark to detect potential performance regressions.

May be related: https://github.com/cupy/cupy/pull/3448

Since the host know the integer values, maybe it would be better to dispatch to either int32 or int64?

seberg:
Good point, those `ptrdiff_t` casts made me think it probably doesn't matter, but the division/modulo happens _before_ that.
FWIW, timing on a V100, it seems to make about a 20% difference for the a very simple advanced indexing assignment (my solution has identical perf.).

Anyway, found a nice solution by using `decltype(...)::index_t` and casting within the kernel here (seems a bit nicer than trying to pass the scalars as the correct type, but if you think that may work out nicer, I am all ears.).

EDIT: One note is that we may want to use `fast_div_mod` in the nearer future at which point I guess this might be 64bit only again, but probably better anyway.

cupy/_core/_routines_indexing.pyx
```diff
@@ -552,34 +558,44 @@ out = a[out_i];
...
 
 _choose_clip_kernel = ElementwiseKernel(
-    'S a, raw T choices, int32 n_channel, int32 n',
+    'S a, raw T choices, int64 n_channel_, int64 n_',
     'T y',
     '''
-      S x = a;
+      using idx_t = decltype(choices)::index_t;
+      idx_t n_channel = static_cast<idx_t>n_channel_;
+      idx_t n = static_cast<indx_t>n_;
```
leofang: typo?

cupy/_core/_routines_indexing.pyx
```diff
@@ -552,34 +558,44 @@ out = a[out_i];
...
+
+      idx_t x;
       if (a < 0) {
         x = 0;
       } else if (a >= n) {
         x = n - 1;
       }
+      else {
+        x = static_cast<indx_t>(a);
```
leofang: typo?

cupy/_core/_routines_indexing.pyx
seberg: Hmmm, yeah. I also got the `i_idx_t` wrong in the scatter kernel. Must have somehow not run tests after final changes :/.

Should be fixed now with tests passing.  (There was also one unused line of code in the the scatter kernel.)

seberg:
Hmmm, I had only spot-run tests which missed a few paths, including that `CArray` with 0-dims didn't actually define `index_t`, so had to add that.

/test mini

cupy/_core/include/cupy/carray.cuh
```diff
@@ -619,6 +619,7 @@ private:
 
 public:
   static const int ndim = 0;
+  typedef typename cupy::type_traits::conditional<_use_32bit_indexing, int, ptrdiff_t>::type index_t;
```
leofang, line 622: I think this diff will likely trigger a full rebuild of the kernel caches, because `carray.cuh` is included in virtually all kernels in CuPy. However, I don't see any alternative... Let's wait for the CI to confirm if I could be wrong.

seberg: Hmmm, I guess you are right, this burned a lot of cycles and timed out on a few jobs :(.

leofang:
I wish there is a way to say "test mini,linux" and trigger only the subset of the pipelines that meet all the requirements 😅 It seems, surprisingly, that only Linux pipelines fail; the Windows ones passed (after approaching the timeout limit).

Issue #9473, opened 2025-11-07
`cupy.cuda.get_local_runtime_version()` fails with `AttributeError` after importing `cupy.cuda.runtime`


cchapmanbird:
### Description

To check the `cuda` version, running 

```py
import cupy
cupy.cuda.get_local_runtime_version()
```

works as expected. However, if I instead write
```py
import cupy
import cupy.cuda.runtime
cupy.cuda.get_local_runtime_version()
```

I now see the error `AttributeError: module 'cupy.cuda.runtime' has no attribute '_getLocalRuntimeVersion'`.

This seems like uninstended behaviour.

### To Reproduce

```py
import cupy
import cupy.cuda.runtime
cupy.cuda.get_local_runtime_version()
```

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
OS                           : Linux-6.1.0-15-amd64-x86_64-with-glibc2.36
Python Version               : 3.12.12
CuPy Version                 : 13.6.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.3.4
SciPy Version                : 1.16.3
Cython Build Version         : 3.0.12
Cython Runtime Version       : None
CUDA Root                    : /usr/local/cuda
nvcc PATH                    : /usr/local/cuda/bin/nvcc
CUDA Build Version           : 12090
CUDA Driver Version          : 12030
CUDA Runtime Version         : 12090 (linked to CuPy) / 12030 (locally installed)
CUDA Extra Include Dirs      : []
cuBLAS Version               : 120304
cuFFT Version                : 11012
cuRAND Version               : 10304
cuSOLVER Version             : (11, 5, 4)
cuSPARSE Version             : 12200
NVRTC Version                : (12, 3)
Thrust Version               : 200800
CUB Build Version            : 200800
Jitify Build Version         : <unknown>
cuDNN Build Version          : None
cuDNN Version                : None
NCCL Build Version           : None
NCCL Runtime Version         : None
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA A30
Device 0 Compute Capability  : 80
Device 0 PCI Bus ID          : 0000:01:00.0
Device 1 Name                : NVIDIA A30
Device 1 Compute Capability  : 80
Device 1 PCI Bus ID          : 0000:21:00.0
```

seberg:
This has a bit convoluted reason:
[...]

PR #9510, opened 2025-12-04
Fix AttributeError in get_local_runtime_version after importing cupy.cuda.runtime

samay2504:
When users import `cupy.cuda.runtime` directly, it replaces the `runtime` attribute in `cupy.cuda`, causing `cupy.cuda.get_local_runtime_version()` to fail with `AttributeError: module 'cupy.cuda.runtime' has no attribute '_getLocalRuntimeVersion'`.

## Problem
The issue occurs because:
1. `cupy/cuda/__init__.py` imports `from cupy_backends.cuda.api import runtime` (line 24)
2. `cupy/cuda/__init__.py` defines `get_local_runtime_version()` which calls `runtime._getLocalRuntimeVersion()` (line 101)
3. When a user does `import cupy.cuda.runtime`, Python replaces the `runtime` attribute with the `cupy.cuda.runtime` module
4. `cupy/cuda/runtime.py` does `from cupy_backends.cuda.api.runtime import *` which doesn't import private symbols like `_getLocalRuntimeVersion`
5. This causes `cupy.cuda.get_local_runtime_version()` to fail

## Solution
Explicitly import `_getLocalRuntimeVersion` in `cupy/cuda/runtime.py` to ensure it's available even when the module is imported directly.

## Changes
- Modified `cupy/cuda/runtime.py` to explicitly import `_getLocalRuntimeVersion` from `cupy_backends.cuda.api.runtime`
- Added test in `tests/cupy_tests/cuda_tests/test_runtime.py` to verify the fix

## Testing
Added a test that:
1. Imports `cupy.cuda.runtime` to trigger the module replacement
2. Verifies that `cupy.cuda.get_local_runtime_version()` still works correctly

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
       # Git
       - id: check-added-large-files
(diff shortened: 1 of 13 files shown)


leofang:
@samay2504 Q: Is this for #9473?

seberg:
@samay2504 sorry for not pinging you earlier, but I think the test failure is actually real, although admittedly a test that monkey-patches exactly this. We can likely adapt the test.

PR #9511, opened 2025-12-06
Add %gpu_timeit IPython magic using cupyx.profiler.benchmark

### LINKED ISSUE #9466: Add a Jupyter cell magic that wraps a cell with `cupyx.profiler.benchmark`
As title. The `%timeit` magic is very useful but it is CPU-only, and often causes confusions for first-time GPU users programming in an IPython or Jupyter environment. Based on the teaching experience @brycelelbach suggested offline that we could add a Jupyter magic like, say, `%gpu_timeit`, that wraps a cell with `cupyx.profiler.benchmark` for GPU timing purposes.

aman-coder03:
# Add `%gpu_timeit` IPython magic using `cupyx.profiler.benchmark` (#9466)

This PR adds a new Jupyter/IPython magic `%gpu_timeit`, as requested in issue #9466.  
While `%timeit` measures only CPU execution, `%gpu_timeit` provides **GPU-synchronized timing** using `cupyx.profiler.benchmark`.
---
## Summary
implements `%gpu_timeit` as both **line and cell magic**
Uses `cupyx.profiler.benchmark` for correct GPU timing
Similar to `%timeit`, but GPU-aware
Implemented in `cupyx/tools/_gpu_timeit.py`
Automatically enabled via:
```python
%load_ext cupy
```

fixes #9466

### DIFF
--- a/cupyx/__init__.py
+++ b/cupyx/__init__.py
@@ -32,12 +32,10 @@ def __getattr__(key):
         return cupyx.lapack
 
     raise AttributeError(
-        "module '{}' has no attribute '{}'".format(__name__, key)
-    )
+        "module '{}' has no attribute '{}'".format(__name__, key))
+
 
 # Auto-register IPython magics only when running in IPython/Jupyter
-
-
 try:
     from IPython import get_ipython
 


aman-coder03:
Thanks for the guidance! @leofang 
I have now implemented the changes!
please let me know if any further adjustments are needed.

leofang:
@aman-coder03 thanks. Please address the previous comment:
> * please do not change the submodules

Anything under the `third_party/` directory should remain unchanged.

aman-coder03:
@leofang you can check now!

cupyx/__init__.py
```diff
@@ -32,4 +32,19 @@ def __getattr__(key):
         return cupyx.lapack
 
     raise AttributeError(
-        "module '{}' has no attribute '{}'".format(__name__, key))
+        "module '{}' has no attribute '{}'".format(__name__, key)
+    )
+
+# Auto-register IPython magics only when running in IPython/Jupyter
+
+
```
leofang:
```suggestion
        "module '{}' has no attribute '{}'".format(__name__, key))


# Auto-register IPython magics only when running in IPython/Jupyter
```

tests/cupyx_tests/tools_tests/test_gpu_timeit.py
```diff
@@ -0,0 +1,44 @@
...
+def dummy_benchmark(func):
+    func()
+    return DummyBenchmarkResult()
+
+
+fake_profiler.benchmark = dummy_benchmark
+
+sys.modules["cupy"] = fake_cupy
+sys.modules["cupyx"] = types.ModuleType("cupyx")
+sys.modules["cupyx.profiler"] = fake_profiler
```
leofang, line 29: Q: Why would we need to fake it? Can't we just run real CuPy code?

aman-coder03: I initially mocked `cupy `and `cupyx.profiler` because the test wasn’t being run inside an environment with CuPy built from source on my machine, so importing the real modules failed locally.

I can revise the test to:
import real `cupy `and `cupyx.profiler.benchmark`
use a very lightweight GPU/CPU snippet that is safe across configurations
check only that the magic is registered and produces benchmark-style output (without mocking)

Let me know if you'd prefer:
A minimal GPU operation like `cp.arange(10)`, or
A no-op function inside the benchmark (still using real benchmark)
I can update the test accordingly.

leofang: Let us not mock it. It is important that we test real-world use cases. Also, if you mock it this way then all subsequent tests run in the same pytest process are ruined. Please switch to do lightweight CuPy tests, say create two tiny arrays and do a matmul.

cupyx/__init__.py
leofang, line 1: A UX question for @kmaehashi: In the present form, the magics are loaded only when `cupyx` is imported. I can see us going either way:
1. we delay the magic loading until we actually need it
   - but then the question is: How do we signal this need? By importing something? (what?)
2. we always load the magics as soon as CuPy is imported
   - but then how can we avoid in the meantime importing cupyx (https://github.com/cupy/cupy/issues/9324)?

leofang: Thinking out loud: What if we move the code added in this PR from `cupyx/` to a new dir, say, `cupy_magics/`?

leofang:
Continued in #9561.

PR #9512, opened 2025-12-08
ENH: allow python scalars in the 2nd argument of searchsorted

ev-br:
The upcoming Array API revision 2025.12 will allow the second argument of searchsorted to be a python scalar, cf
https://github.com/data-apis/array-api/pull/982, so here's a matching proposal for CuPy.

### DIFF
--- a/tests/cupy_tests/sorting_tests/test_search.py
+++ b/tests/cupy_tests/sorting_tests/test_search.py
@@ -699,7 +699,8 @@ class TestSearchSorted:
         x = testing.shaped_arange(self.shape, xp, dtype)
         bins = xp.array(self.bins)
         y = xp.searchsorted(bins, x, side=self.side)
-        y1 = xp.searchsorted(bins, 42.5, side=self.side)   # python scalar for `v`
+        # python scalar for `v`
+        y1 = xp.searchsorted(bins, 42.5, side=self.side)
         return y, y1
 
     @testing.for_all_dtypes(no_bool=True)
@@ -708,6 +709,15 @@ class TestSearchSorted:
         x = testing.shaped_arange(self.shape, xp, dtype)
         bins = xp.array(self.bins)
         y = bins.searchsorted(x, side=self.side)
+        return y,
+
+
+class TestSearchSortedScalar:
+    @pytest.mark.parametrize('side', ['left', 'right'])
+    @pytest.mark.parametrize('v', [-1.0, 0.0, 1.0, 1.5, 2.0, 4.0, 4.5])
+    @testing.numpy_cupy_array_equal()
+    def test_searchsorted_scalar(self, xp, v, side):
+        y = xp.searchsorted(xp.arange(3), v)
         return y,
 
 


cupy/_sorting/search.py
```diff
@@ -430,7 +430,10 @@ def _searchsorted(a, v, side, sorter, assume_increasing):
         raise NotImplementedError('Only int or ndarray are supported for a')
 
     if not isinstance(v, cupy.ndarray):
-        raise NotImplementedError('Only int or ndarray are supported for v')
+        if not isinstance(v, int | float | complex):
+            raise NotImplementedError(
+                'Only python scalars or ndarrays are supported for v')
+        v = cupy.asarray(v, dtype=a.dtype)
```
seberg: Thanks!

Technically, this differs from NumPy, because NumPy would support `searchsorted([1, 2, 3], 1.9) != searchsorted([1, 2, 3], 1)` (or complex).

Right now NumPy actually supports out-of-bound integers as well (this is in part because of not-great NEP 50 support), but considering that out-of-bound int's are not a big issue for searchsorted, maybe it's best to just not add the `dtype=a.dtype`?

ev-br: Oh good catch!

Given 

```
In [1]: import jax.numpy as jnp

In [2]: import torch

In [3]: import numpy as np

In [4]: np.searchsorted(np.arange(3), 1.9)
Out[4]: np.int64(2)

In [5]: np.__version__
Out[5]: '2.2.6'

In [6]: jnp.searchsorted(jnp.arange(3), 1.9)
Out[6]: Array(2, dtype=int32)

In [7]: torch.searchsorted(torch.arange(3), 1.9)
Out[7]: tensor(2)
```

and

```
In [8]: cp.round(1.9).astype(int)
Out[8]: array(2)
```

WDYT about this mouthful: 

```
if cupy.issubdtype(a, cupy.integer):
    v = cupy.round(cupy.asarray(v)).astype(a.dtype)
else:
    v = cupy.asarray(v, dtype=a.dtype)
```

ev-br: Oh of course not, scratch my previous suggestion. Need some more coffee. Removed the `dtype=a.dtype` meanwhile.

seberg: I think you would have to use a `ceil` or `floor` depending on `side=`, though? For out-of-range value, I guess things are OK, in the sense that they would just error.
I suppose float64->float32 rounding might have similar effects, although I am not sure that we should worry about it.
(I.e. I am not sure what NumPy does right now there, but it may be fine to change it to NEP 50 behavior of downcasting the scalar.)

tests/cupy_tests/sorting_tests/test_search.py
```diff
@@ -699,7 +699,8 @@ def test_searchsorted(self, xp, dtype):
         x = testing.shaped_arange(self.shape, xp, dtype)
         bins = xp.array(self.bins)
         y = xp.searchsorted(bins, x, side=self.side)
-        return y,
+        y1 = xp.searchsorted(bins, 42.5, side=self.side)   # python scalar for `v`
```
seberg:
```suggestion
        # python scalar for `v`
        y1 = xp.searchsorted(bins, 42.5, side=self.side)
```
Linter is unhappy.

cupy/_sorting/search.py
ev-br: Given that searchsorted always returns integers, maybe there's no need to go full NEP 50?

Just doing `v = np.asarray(v)   # no dtype=` seems to work under an added test, at least shows that it's in lockstep with what numpy does. Integer overflows is something else, but indeed do we need to worry about them. 

```
$ pytest tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar -v
================================= test session starts ==================================
platform linux -- Python 3.11.0, pytest-9.0.1, pluggy-1.6.0 -- /home/ev-br/.conda/envs/cupyx/bin/python3.11
cachedir: .pytest_cache
rootdir: /home/ev-br/repos/cupy_ng
configfile: pyproject.toml
collected 14 items                                                                     

tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[-1.0-left] PASSED [  7%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[-1.0-right] PASSED [ 14%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[0.0-left] PASSED [ 21%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[0.0-right] PASSED [ 28%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[1.0-left] PASSED [ 35%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[1.0-right] PASSED [ 42%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[1.5-left] PASSED [ 50%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[1.5-right] PASSED [ 57%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[2.0-left] PASSED [ 64%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[2.0-right] PASSED [ 71%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[4.0-left] PASSED [ 78%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[4.0-right] PASSED [ 85%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[4.5-left] PASSED [ 92%]
tests/cupy_tests/sorting_tests/test_search.py::TestSearchSortedScalar::test_searchsorted_scalar[4.5-right] PASSED [100%]

================================== 14 passed in 2.93s ==================================

```

seberg: Yes, agreed it is actually better to not use NEP 50 promotion, as there is indeed no reason to do so from a behavior oerspective.

ev-br:
> Not sure if a kernel launch with a scalar would be better

FWIW I thought about it. My impression is it will complicate the kernel without giving tangible speed-ups. Happy to implement it if there are strong reasons to though.

seberg:
Thanks @ev-br, seems safe to me, so let me put it in. If we someone wants a follow-up, I am happy to do that.

PR #9514, opened 2025-12-09
DOC: Add Nsight Compute profiling tutorial for CuPy kernels

### LINKED ISSUE #8450: Option to pass rootpath to `cp.RawModule` over using tmp
### Description

For profiling it's very common to analyze the PTC or SASS along with the source code, to check which lines translate to which instructions and the time each one takes. So keeping the source file that generated the NVTX can be of interest.
Currently the file is generated automatically and saved in a tmpdir, with default commands in the source
 
https://github.com/cupy/cupy/blob/49135f8310441be6ad8d7a717f539a63dee79e9d/cupy/cuda/compiler.py#L315C1-L322C66

Request:

Add an argument to the functions `compile_using_nvcc` and `compile_using_nvrtc` to pass a root directory, otherwise a temporary one is used, and also update the other functions and modules that interact with it, such as `cupy.RawModule`

### Additional Information

Example of debugging with NCU where source info is not available:
Screenshot from 2024-08-02 16-54-21

leofang:
Close #8450.

### DIFF
--- a/docs/source/user_guide/performance.rst
+++ b/docs/source/user_guide/performance.rst
@@ -86,14 +86,7 @@ When profiling CuPy kernels (whether they are part of the CuPy codebase or :ref:
 it is often useful to correlate the source code with performance metrics in tools like NVIDIA Nsight Compute. 
 This allows you to analyze which lines translate to which instructions and see the time each one takes.
 
-The trick is to leverage CuPy's ability to dump kernel source code to disk and then load it into the profiler.
-
-Prerequisites
-.............
-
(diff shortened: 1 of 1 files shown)


leofang:
Rendered docs: https://cupy--9514.org.readthedocs.build/en/9514/user_guide/performance.html#profiling-kernels-with-nsight-compute

seberg (approved):
Should try it out once in full, but reads very well, and will probably just merge it as is in a bit. No need to run CI on this.

/skip ci
/ci skip

kmaehashi (approved):

PR #9516, opened 2025-12-10
MAINT: simplify and fix `CUPY_TEST_GPU_LIMIT`

seberg:
Not that `CUPY_TEST_GPU_LIMIT` is a game-changer probably, but it didn't actually work reliably (I assume it used to work). I am not sure why exactly, it may be that the fixture is too late because it is run after test collection (which then may e.g.
create cupy arrays).

This pattern using the environment variables was probably not available when it was introduced initially, but works now.

(N.B. at least for high-memory tests using two runners on the same GPU probably doesn't work, xdist groupings may be a way to solve this.)

---

Funny observation is that it isn't necessarily faster to use many GPUs...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,8 +77,8 @@ filterwarnings = [
   "ignore::UserWarning",
   "error::DeprecationWarning",
   "error::PendingDeprecationWarning",
-  "error::cupy.exceptions.VisibleDeprecationWarning",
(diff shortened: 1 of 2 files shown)


seberg:
I'll close this for now, might be a nice simplification on newer pytest-xdist, but doesn't matter...

The problem was that I accidentally caused `import cupy` to actually initialize the cuda context.

PR #9518, opened 2025-12-12
Fix CI (mini)

EarlMilktea:
## Changes

- Upgrade hooks
- Resolve or ignore newly-detected linter errors
- Prune a deprecated rule from prproject

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -398,7 +398,7 @@ configs {
   value {
     requirement {
       cpu: 8
-      memory: 50
+      memory: 60
       disk: 10
       gpu: 1
     }
--- a/.pfnci/windows/_flexci.ps1
+++ b/.pfnci/windows/_flexci.ps1
@@ -1,7 +1,5 @@
 function ActivatePython($version) {
-    if ($version -eq "3.9") {
-        $pydir = "Python39"
-    } elseif ($version -eq "3.10") {
+    if ($version -eq "3.10") {
         $pydir = "Python310"
     } elseif ($version -eq "3.11") {
         $pydir = "Python311"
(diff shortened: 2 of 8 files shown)


cupy/_core/_fusion_variable.pyx
```diff
@@ -143,10 +143,10 @@ class _TraceVariable:
     def format(self, form, **kwargs):
         """Returns a string following the format taken as an input.
         """
-        kwargs = dict([
+        kwargs = {
             (k, get_typename(v) if isinstance(v, numpy.dtype) else v)
```
seberg: Are these intentional here? Just a quick note either way, it's `k: get_typename(v)` also below.  (I may want to check the generated C code once, to make sure the list isn't there for Cython reasons.)

EarlMilktea: My bad! Thank you for the comment.

EarlMilktea: @seberg Could you check the code again? Thank you in advance!

cupy/_core/fusion.pyx
```diff
@@ -928,8 +928,11 @@ def _call_reduction(fusion_op, *args, **kwargs):
         raise TypeError(mes.format(fusion_op._ops.name, len(args)))
 
     arg = args[0]
-    kwargs = dict([(key, value) for key, value in kwargs.items()
-                   if (key in ('axis', 'out') and value is not None)])
+    kwargs = {
+        key: value
+        for key, value in kwargs.items()
+        if (key in ('axis', 'out') and value is not None)
+    }
```
seberg, line 935: Feels a bit much for updating two values, but let's keep it.

cupy_backends/cuda/api/runtime.pyx
```diff
@@ -217,11 +217,11 @@ cpdef getDeviceProperties(int device):
...
-        clockRate = props.clockRate
-        kernelExecTimeoutEnabled = props.kernelExecTimeoutEnabled
-        memoryClockRate = props.memoryClockRate
-        computeMode = props.computeMode
-        cooperativeMultiDeviceLaunch = props.cooperativeMultiDeviceLaunch
+        clockRate = props.clockRate  # no-cython-lint
+        kernelExecTimeoutEnabled = props.kernelExecTimeoutEnabled  # no-cython-lint
+        memoryClockRate = props.memoryClockRate  # no-cython-lint
+        computeMode = props.computeMode  # no-cython-lint
+        cooperativeMultiDeviceLaunch = props.cooperativeMultiDeviceLaunch  # no-cython-lint
```
seberg, line 224: I guess these are only used in other `ELSE:` blocks and cython lint thinks it's possibly unused?

pyproject.toml
```diff
@@ -132,7 +132,6 @@ select = ["E", "F", "W", "UP", "I002", "FA", "PYI", "PLC01"]
 extend-ignore = [
   "UP031", # printf-string-formatting
   "UP032", # f-string
-  "UP038", # Use `X | Y` in `isinstance` call instead of `(X, Y)`
```
seberg, line 135: If someone confirms to me that `X | Y` has the same speed in this context in Cython, I wouldn't mind adopting it. But, I guess that would be a lot of churn.

cupy_backends/cuda/api/runtime.pyx
EarlMilktea: I think so too. Do you come up with situations where this block is actually hit?

pyproject.toml
EarlMilktea: This rule itself was removed from `ruff` so this change is necessary.

cupy_backends/cuda/api/runtime.pyx
seberg: Yeah, nvm... Feels like this may be a small `cython-lint` issue, unfortunate, but not a real worry.

seberg:
Thanks @EarlMilktea

PR #9521, opened 2025-12-14
Fill in compatibility matrix upper bound for CuPy v13

kmaehashi:
Filled upper limit for CuPy v13 as we are going to cut v14 release.

Also replaced `~` with `-` to clearly indicate it is a range.


kmaehashi:
Preview: https://cupy--9521.org.readthedocs.build/en/9521/upgrade.html#compatibility-matrix

/test skip

Issue #9513, opened 2025-12-08
Make cupyx.scipy.ndimage.gaussian_filter1d CUDA graph compatible


jhill2:
### Description

Similar to #9489 I am trying to use gaussian_filter1d in a cupy CUDA graph but I can't capture the graph because of an internal H2D transfer. I would expect to be able to use this operation in a CUDA graph. I am using v14.0.0a1 version of cupy which made other filters in ndimage CUDA graphable but not gaussian_filter1d. Here is the stack trace:
```bash
Traceback (most recent call last):
  File "/tmp/cupy_error.py", line 16, in <module>
    output_data = cupyx.scipy.ndimage.gaussian_filter1d(input_data, sigma)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/lm/venv/python311/lib64/python3.11/site-packages/cupyx/scipy/ndimage/_filters.py", line 359, in gaussian_filter1d
    weights = _gaussian_kernel1d(
              ^^^^^^^^^^^^^^^^^^^
  File "/opt/lm/venv/python311/lib64/python3.11/site-packages/cupyx/scipy/ndimage/_filters.py", line 462, in _gaussian_kernel1d
    return cupy.asarray(phi_x)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/lm/venv/python311/lib64/python3.11/site-packages/cupy/_creation/from_data.py", line 92, in asarray
    return _core.array(a, dtype, copy, order, blocking=blocking)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cupy/_core/core.pyx", line 2550, in cupy._core.core.array
  File "cupy/_core/core.pyx", line 2577, in cupy._core.core.array
  File "cupy/_core/core.pyx", line 2810, in cupy._core.core._array_default
  File "cupy/cuda/memory.pyx", line 587, in cupy.cuda.memory.MemoryPointer.copy_from_host_async
RuntimeError: the current stream is capturing, so H2D transfers are disallowed
```

### To Reproduce

```py
import cupy as cp
import cupyx.scipy.ndimage

# Create a CuPy array on the GPU
input_data = cp.arange(10, dtype=cp.float32)
sigma = 1

# Create a stream for capturing the graph
stream = cp.cuda.Stream(non_blocking=True)

[...]

Issue #8689, opened 2024-10-21
Kernel fusion cannot fuse array constructors?


leofang:
In an offline chat with @Brycelelbach we noticed this function
```python
@cp.fuse
def rmmmi_fused(n: int) -> int:
    return cp.sum((cp.arange(1, n) % 10) * 2 + 1)
```
would result in 6 kernels (`arange`, `%`, `*`, `+`, and 2-pass CUB reduction) being launched instead of 2 (for reduction-only). It turns out `arange` was not fusible, presumably because it allocates new buffer and `@fuse` cannot handle this? But we should be able to see through the fused ops and allocate the final buffer, right?

It'd be nice if this is doable, as it's a bit surprising to see no perf gain without/with `@fuse` for this case, but it's just a nice to have as we already figured out a workaround (by passing in an `arange`'d array from outside of the fusion context).

leofang:
I probably overlooked another possibility why this is not doable: The final buffer size is dynamic (determined by the input `n`)?

leofang:
Quoting @asi1024 from https://github.com/cupy/cupy/issues/9002#issuecomment-2696304704:
> * If constructing an ndarray within `cupy.fuse`, it is not possible to detect shape mismatches if the axis lengths differ.

We should probably document the limitation of `@cupy.fuse`. I'll create a separate tracking issue.

PR #9522, opened 2025-12-15
Add CUDA graph support to _gaussian_kernel1d

jhill2:
This addresses the issue #9513 where CUDA graph capture will fail if wrapping _gaussian_kerne1d because it performs H2D memory transfers.

The fix is to keep the Numpy operations for regular use but use Cupy operations if running a CUDA graph capture.


leofang (commented):
@jhill2 Sorry for late reply. I think the fix is fine. But I do see a huge optimization opportunity left on the table and if you want to support CUDA graphs due to perf reasons, you might be interested in this. 

When Jeff checked in this code, the TODO was to use `@fuse` to fuse all the small kernels (https://github.com/cupy/cupy/pull/3505#issuecomment-654582605). I suspect it did not work back then because `arange` is called (https://github.com/cupy/cupy/issues/8689), but the remaining array operations should be fusible. 

Let me know if you'd like to look into this in this PR, or table it for the future.

jhill2:
@leofang Thank you for the suggestion. I am not very familiar with cupy.fuse and have a few questions. I haven't eliminated the arange from the function so based on your linked threads, I wouldn't get an improvement from trying to fuse those. Do you therefore suggest breaking out certain lines from the _gaussian_kernel1d function that would benefit from fusing into their own functions, wrapping them with @fuse and calling them from _gaussian_kernel1d? Or are you suggesting that wrapping all of _gaussian_kernel1d with @fuse will do the best it can by fusing the portions of the function that it is able to? If the latter is acceptable then why aren't more internal cupy functions already wrapped in @fuse?

I'm happy to dig into the @fuse option, but I am also satisfied with the performance of the code as it currently is in the PR. This function is only one small piece in my overall desired CUDA graph, so I am not currently hurt by the lack of full optimization. But I also love getting the most out of cupy, so I'm willing to help if I can be steered in the right direction.

leofang:
> Do you therefore suggest breaking out certain lines from the `_gaussian_kernel1d` function that would benefit from fusing into their own functions, wrapping them with `@fuse` and calling them from `_gaussian_kernel1d`? 

Yes. I guess we can exclude `arange`, fuse the rest, and call them from `_gaussian_kernel1d`.

> If the latter is acceptable then why aren't more internal cupy functions already wrapped in `@fuse`?

We did not have a strict policy when CuPy started, we did however recognize the importance of fusion and many internal kernels are fused. Furthermore, a casual fusion could silently fail (as in -- if a fusion cannot happen a regular dispatch would occur, causing performance mismatching developer expectation).

PR #9523, opened 2025-12-15
Make FFT config more thread-safe and adatapt FFT tests

seberg:
This contains two sets of changes around thread-safety and FFTs.

First, it modifies `cupy.fft.config` to be an object rather than a module to allow making `.devices` thread-local, this may be up for discussion but for the FFT config it seemed sensible to me.

I decided to use contextvars for this, which I think is the best pattern (thread and context safe), but they remain private for now.

The tests also need fairly extensive changes anyway to make them thread-safe, one with custom decoartors (fixtures are not run for each thread in `pytest-run-parallel`).
They were also digging into the FFT internals so needed to adapt to that.

(My local runs had some problems with the callback tests with or without threading, so this may not be 100%.)

---

Specifically `import cupy.fft.config` is broken by this of course as it would try to import the module and not the `config` object.

This is one of multiple next steps to gh-9488, but luckily it is probably the largest non-test change.

### DIFF
--- a/docs/source/reference/fft.config.rst
+++ b/docs/source/reference/fft.config.rst
@@ -1,0 +1,13 @@
+``cupy.fft.config``
+===================
+
+The ``cupy.fft.config`` object provides configuration options for FFTs and
+FFT plans.
+
+.. autoclass:: cupy.fft::config.set_cufft_callbacks
+   :members:
+   :special-members: __enter__, __exit__
+.. autofunction:: cupy.fft::config.set_cufft_gpus
+.. autofunction:: cupy.fft::config.get_plan_cache
+.. autofunction:: cupy.fft::config.show_plan_cache_info
+
--- a/docs/source/reference/fft.rst
+++ b/docs/source/reference/fft.rst
@@ -59,15 +59,12 @@ Helper routines
 CuPy-specific APIs
 ------------------
 
-See the description below for details.
+CuPy has some additional configuration options available:
 
-.. autosummary::
-   :toctree: generated/
+.. toctree::
+   :maxdepth: 2
 
-   config.set_cufft_callbacks
-   config.set_cufft_gpus
-   config.get_plan_cache
-   config.show_plan_cache_info
+   fft.config
 
 
 Normalization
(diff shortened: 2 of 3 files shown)


cupy/fft/_config.py
```diff
@@ -0,0 +1,89 @@
...
+        .. seealso:: `Multiple GPU cuFFT Transforms`_
+
+        .. _Multiple GPU cuFFT Transforms:
+            https://docs.nvidia.com/cuda/cufft/index.html#multiple-GPU-cufft-transforms
+        '''
+        _util.experimental('cupy.fft.config.set_cufft_gpus')
+
+        if isinstance(gpus, int):
+            devs = [i for i in range(gpus)]
+        elif isinstance(gpus, (list, tuple)):
```
seberg, line 59: This I changed to also allow tuples (simplifies restoring).

seberg:
As per Leo's suggestion, I added a weird, but (apparently) working hack to add `cupy.fft.cache` to `sys.modules` directly, which makes `import cupy.fft.config` or `from cupy.fft.config import get_plan_cache` work.

With that, the change here the multi-GPU config is a contextvar.  That means it is not inherited by threads (except on free-threaded and probably future Python versions).

cupy/fft/_config.py
```diff
@@ -0,0 +1,89 @@
...
+        # make it hashable
+        self._devices.set(tuple(devs))
+
+    @property
+    def enable_nd_planning(self):
+        return self._enable_nd_planning.get()
+
+    @enable_nd_planning.setter
+    def enable_nd_planning(self, value):
+        self._enable_nd_planning.set(value)
```
leofang, line 81: I think we should deprecate this flag and always enable nd planning by default

tests/cupy_tests/cuda_tests/test_cufft.py
```diff
@@ -37,8 +37,8 @@ class TestMultiGpuPlan1dNumPy(unittest.TestCase):
 
     @multi_gpu_config(gpu_configs=[[0, 1], [1, 0]])
     @testing.for_complex_dtypes()
-    def test_fft(self, dtype):
-        _skip_multi_gpu_bug(self.shape, self.gpus)
+    def test_fft(self, dtype, gpus):
```
leofang: Q: How is `gpus` passed to the tests? I don't see any new test fixture being used?

seberg: A problem with fixtures is that they are only run once per test also in `pytest-run-parallel`, so I changed the `multi_gpu_setup` to pass in the kwarg rather than add `self.` and made it a normal decorator and not a fixture.
(It would be nice if pytest-run-parallel figured out a way to make fixtures that ran every time.)

seberg: Ah, this was a bit silly.  It's only used for `_skip_multi_gpu_bug`, I'll have that check the config state instead.

seberg: Actually this was only needed for CUDA 11.0-11.2, so I think we can just drop it entirely.

cupy/fft/_config.py
```diff
@@ -0,0 +1,96 @@
...
+    def use_multi_gpus(self):
+        return self._use_multi_gpus.get()
+
+    @use_multi_gpus.setter
+    def use_multi_gpus(self, value):
+        self._use_multi_gpus.set(bool(value))
+
+    @property
+    def devices(self):
+        return self._devices.get() if self._use_multi_gpus.get() else None
```
leofang, line 93: I don't think we need to expose this (it wasn't a public API), but I am fine either way.

cupyx/scipy/fftpack/_fft.py
```diff
@@ -52,8 +52,7 @@ def get_fft_plan(a, shape=None, axes=None, value_type='C2C'):
             # alternatively:
             y = cupyx.scipy.fftpack.fftn(x, plan=plan)  # pass plan explicitly
 
-        In the first case, no cuFFT plan will be generated automatically,
-        even if ``cupy.fft.config.enable_nd_planning = True`` is set.
+        In both cases the user created ``plan`` will be used.
```
leofang, line 55: 👍

leofang (approved):
LGTM!

cupy/fft/_config.py
seberg: Hmmm, good point, it's the smaller change to hide it away (I think I added it to make the internal usage more convenient).

OTOH, since the `config` has these properties, this doesn't seem like bad API. (We could add a setter in the future as well).

Maybe @kmaehashi has a quick opinion?

leofang: `set_cufft_gpus` is our setter. Adding a getter here is fine.

leofang (commented):
CI failure is irrelevant, let's merge

/test force-skip

PR #9525, opened 2025-12-16
Support viewing CuPy arrays as `cuda::std::mdspan` on host/device

leofang:
This PR introduces a way for CuPy to interoperate with CUDA C++, by supporting emitting `cuda::std::mdspan` on the host that can be passed by value to the device. Same as how we have been treating CArray over the years, we exploit the `cuda::std::mdspan` ABI stability to allow dynamically generating the struct.

With this PR, it really helps reduce overhead of generating kernel args (compared to e.g. DLPack which has O(1 us) overhead):
```python
In [1]: import cupy as cp

In [2]: arr = cp.arange(32).reshape(4, 8)

In [3]: %timeit mdspan = arr.mdspan(index_type=cp.int32)
199 ns ± 4.23 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

xref: https://github.com/NVIDIA/cuda-python/pull/1387

### DIFF
--- a/cupy/cuda/function.pyx
+++ b/cupy/cuda/function.pyx
@@ -52,7 +52,7 @@ cdef class CNumpyArray(CPointer):
 
 cdef inline CPointer _pointer(x):
     if x is None:
-        return CPointer()
+        return CIntptr(0)
     if isinstance(x, _ndarray_base):
         return (<_ndarray_base>x).get_pointer()
     if isinstance(x, _carray.Indexer):


tests/cupy_tests/core_tests/test_mdspan.py
```diff
@@ -0,0 +1,121 @@
...
+                    f'verify_mdspan_layout_stride<{dtype_str}, {index_str}>'
+                )
+
+        self.mod = cupy.RawModule(
+            code=code_verify,
+            options=('--std=c++17',),
+            name_expressions=name_expressions
+        )
+
+    # TODO(leofang): it does not seem 'AK' makes any difference here?
```
seberg, line 322: You would need to transpose a 3-D array and `shaped_random` has no logic to intentionally mangle it up.
I.e. if you truly want this, you need to test 3-D and then parametrize a transpose with `(0, 1, 2), (2, 1, 0), (2, 0, 1)` where the last would be neither C nor F.

(There is a helper for "contiguous axes", but I am not sure there is one for this.)

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
 from cupy._core cimport internal
 
 
+cdef class mdspan(function.CPointer):
+
+    cdef int init(
+            self, void* data_ptr, int itemsize,
+            const shape_t& shape, const strides_t& strides,
+            int index_itemsize) except?-1:
```
seberg: Should we expose the `index_itemsize` in some way?

cupy/_core/core.pyx
```diff
@@ -586,6 +586,21 @@ cdef class _ndarray_base:
...
+
+        Args:
+            index_type: The data type for extent and stride indices.
+                Defaults to cupy.int64. Can be cupy.int32 or cupy.int64.
+
+        Returns:
+            mdspan: An mdspan view of the array.
+        """
+        if index_type is None:
+            import cupy
```
seberg:
```suggestion
```

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
...
+                raise ValueError(
+                    f"Unsupported index_itemsize: {index_itemsize}"
+                )
+        assert offset == total_size
+
+    def __cinit__(self):
+        self.ptr = 0
+
+    def __dealloc__(self):
+        if self.ptr != 0:
```
seberg, line 85: Note or nit: No need for the check (PyMem_Free is OK with 0). I think technically, also no need to `__cinit__` (because NULLing everything is the default), but doesn't hurt.

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
...
+            raise MemoryError
+        self.ptr = <intptr_t>data
+
+        cdef size_t offset = 0
+        cdef int i
+        memcpy(<char*>(data) + offset, &data_ptr, sizeof(data_ptr))
+        offset += sizeof(data_ptr)
+        if ndim != 0:
+            if index_itemsize == 4:
+                for i in range(ndim):
```
seberg: Should we add a check that size fits `uint32` (apparently that is the technical requirement that the size has to fit the corresponding unsigned int: `https://en.cppreference.com/w/cpp/container/mdspan/size.html`).
(I suppose technically, if we use unsigned an individual dimensions might still overflow the signed one.)

Some with strides. We could check for overflow.

cupy/_core/core.pyx
```diff
@@ -586,6 +586,21 @@ cdef class _ndarray_base:
         """
         return _CArray_from_ndarray(self)
 
+    def mdspan(self, index_type=None):
```
seberg: So one thing to consider IMO is to think of gh-9588. If the `mdspan` pointer object would get a `ctype` attribute then a non-raw kernel could use that.

But in that case we could default to whatever is best for the current array, i.e. use a C-/F-contig layouting option and figure out the best index type for the user.

This may not matter here to much, except when it comes to what defaults would be best for `arr.mdspan()`.

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
...
+                    (<long long*>(<char*>(data) + offset))[0] = (
+                        <long long>(strides[i] // itemsize)
+                    )
+                    offset += sizeof(long long)
+            else:
+                raise ValueError(
+                    f"Unsupported index_itemsize: {index_itemsize}"
+                )
+        assert offset == total_size
+
```
seberg, line 80: Maybe add an `__init__` that raises an error that we can't instantiate it directly?

cupy/_core/core.pyx
```diff
@@ -586,6 +586,60 @@ cdef class _ndarray_base:
...
+                ) {
+                    // Access: arr(i, j)
+                }
+
+            **Static extents are NOT supported.** For example, using
+            ``extents<int, 4, 8>`` will result in incorrect memory layout
+            and undefined behavior.
+
+            The mdspan uses element strides (not byte strides) and supports
+            arbitrary stride patterns including sliced and transposed arrays.
```
mhoemmen: `layout_stride` does not support arbitrary strides.  The strides must be all positive and the resulting layout must be unique (which excludes some cases of all-positive strides, as well as "broadcasting" layouts).  Please refer to https://isocpp.org/files/papers/P3959R0.html for a discussion.

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
...
+        offset += sizeof(data_ptr)
+        if ndim != 0:
+            if index_itemsize == 4:
+                for i in range(ndim):
+                    (<int*>(<char*>(data) + offset))[0] = <int>shape[i]
+                    offset += sizeof(int)
+
+                for i in range(ndim):
+                    (<int*>(<char*>(data) + offset))[0] = (
+                        <int>(strides[i] // itemsize)
```
mhoemmen: `layout_stride` cannot use negative strides directly, because layout mappings must always return a nonnegative number in [0, `required_span_size()`).  That's why we wrote `layout_stride_relaxed` (please see [this pull request](https://github.com/NVIDIA/cccl/pull/7175)) and why that has an offset as well as strides.

Negative strides in Python mean iterating backwards.  That means that the start pointer might actually be the end pointer.  mdspan actually casts the result of the layout mapping to `size_t`.  That means it turns your negative layout mapping result (counting backwards) to a nonnegative layout mapping result by casting to unsigned.

That's actually bad.  You'll be accessing elements outside the array.

That's why `layout_stride_relaxed` has an offset instead.  The pointer is the start pointer and the offset points to the end pointer, so that the negative strides can count backwards and still get a nonnegative layout mapping result.

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
...
+        cdef size_t ndim = shape.size()
+        assert ndim == strides.size()
+        assert ndim <= MAX_NDIM
+
+        cdef size_t total_size = \
+            sizeof(void*) + ndim * 2 * index_itemsize
+        cdef void* data = PyMem_Malloc(total_size)
+        if data == NULL:
+            raise MemoryError
+        self.ptr = <intptr_t>data
```
mhoemmen, line 53: In C++, one would normally convert the pointer to `uintptr_t` in order to treat it as an integer.  I'm not sure what the usual Python idiom is.

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
...
+
+        cdef size_t offset = 0
+        cdef int i
+        memcpy(<char*>(data) + offset, &data_ptr, sizeof(data_ptr))
+        offset += sizeof(data_ptr)
+        if ndim != 0:
+            if index_itemsize == 4:
+                for i in range(ndim):
+                    (<int*>(<char*>(data) + offset))[0] = <int>shape[i]
```
mhoemmen: This seems to assume that the index type is `int`, when all we know here is that it is 4 bytes.  It could also be `unsigned`.

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,65 @@ from cupy.cuda cimport function
...
+                    offset += sizeof(int)
+
+                for i in range(ndim):
+                    (<int*>(<char*>(data) + offset))[0] = (
+                        <int>(strides[i] // itemsize)
+                    )
+                    offset += sizeof(int)
+            elif index_itemsize == 8:
+                for i in range(ndim):
+                    (<long long*>(<char*>(data) + offset))[0] = (
```
mhoemmen: This seems to assume that the index type is `long long`, when all we know here is that it is 8 bytes. It could also be `unsigned long long`.

cupy/_core/_carray.pyx
mhoemmen: Never mind, I see that you require `cupy.int32` below.

cupy/_core/_carray.pyx
mhoemmen: Never mind, I see that you require `cupy.int64` below.

tests/cupy_tests/core_tests/test_mdspan.py
```diff
@@ -0,0 +1,474 @@
...
+
+template<typename T, typename IndexType>
+__global__ void verify_mdspan_2d(
+    cuda::std::mdspan<
+        T,
+        cuda::std::extents<
+            IndexType,
+            cuda::std::dynamic_extent,
+            cuda::std::dynamic_extent>,
+        cuda::std::layout_stride> arr_in,
```
mhoemmen: `mdspan` can't use negative strides directly without a positive offset, because layout mappings must always return a nonnegative number in [0, `required_span_size()`).

This part is important.  Negative strides in Python mean iterating backwards.  That means that the start pointer might actually be the end pointer.  `mdspan` actually casts the result of the layout mapping to `size_t`.  That means it turns your negative offset (counting backwards) to a nonnegative offset by casting to unsigned.

That's actually bad.  You'll be accessing elements outside the array.

That's why `layout_stride_relaxed` (see https://github.com/NVIDIA/cccl/pull/7175 ) has an offset instead.  The pointer is the start pointer and the offset points to the end pointer, so that the negative strides can count backwards and still get a nonnegative layout mapping result.

mhoemmen: Negative `required_span_size()` is bad for two reasons.

1. Users use `required_span_size()` to allocate an array large enough to hold all of an mdspan's elements.
2. If an mdspan has `is_exhaustive()` and `is_unique()` both `true`, then users can iterate over its elements by bypassing the mapping and using the accessor to access [0, `required_span_size()`).  It's valid to do the latter with a `size_t` index, because that's what accessor's `access` function takes.  Casting -1 to `size_t` will result in a huge number.

cupy/_core/_carray.pyx
leofang: In CuPy we use `intptr_t` for memory pointers so as to allow pointer arithmetic.

cupy/_core/core.pyx
leofang: disabled by default in commit d0fdde118

cupy/_core/_carray.pyx
leofang: disabled by default in commit d0fdde118

tests/cupy_tests/core_tests/test_mdspan.py
leofang: disabled by default in commit d0fdde118

leofang:
Thanks to Mark for offline discussion. This PR is ready for review. I'll expand the PR description tonight/tomorrow.

cupy/_core/_carray.pxd
```diff
@@ -19,6 +19,15 @@ cdef struct _CArray:
     Py_ssize_t size
 
 
+@cython.final
+cdef class mdspan(function.CPointer):
+
```
seberg, line 24: NIT: could add a `__init__` that errors just for the sake of it.

cupy/_core/_carray.pyx
```diff
@@ -5,6 +5,88 @@ from cupy.cuda cimport function
...
+cdef inline int populate_shape_strides(
+        index_t val, size_t ndim, bint allow_unsafe,
+        void* data, size_t& offset, int itemsize,
+        const shape_t& shape, const strides_t& strides) except?-1:
+    cdef int i
+
+    for i in range(ndim):
+        if not allow_unsafe and shape[i] == 0:
+            raise RuntimeError(
+                f"{i}-th dimension has size zero")
```
seberg, line 22: NIT: Since we have one error, maybe add a check for too large values as well (that is for 64bit exported incorrectly as 32bit).

cupy/_core/core.pyx
```diff
@@ -586,6 +586,64 @@ cdef class _ndarray_base:
         """
         return _CArray_from_ndarray(self)
 
+    def mdspan(self, index_type=None, allow_unsafe=False):
```
seberg: Just to write down my thought from yesterday's discussion:

I wonder if we may have a future where the returned `mdspan` object has:
```
mdspan.ptr
mdspan.ctype == "mdspan<index_t, ..., layout_*>"  # depends on actual layout
```
(or similar).  So an `ElementwiseKernel` could use that `.ctype`.

---

The thing that this might change is that if we optimize for use with `ElementwiseKernel` that actually checks `.ctype`, then the default `index_type` could be whatever is best for the array.

But... if we expect a lot of "raw" usage, then I think this default is probably best. And likely the future would be adding a `layout="best"` option and in that case there is no API convern *right now*.

cupy/_core/core.pyx
```diff
@@ -586,6 +586,64 @@ cdef class _ndarray_base:
...
+                            cuda::std::dynamic_extent,
+                            cuda::std::dynamic_extent>,
+                        cuda::std::layout_stride> arr
+                ) {
+                    // arr is a 2D mdspan
+                    // Access: arr(i, j)
+                }
+
+            **Static extents are NOT supported.** For example, using
+            ``extents<int, 4, 8>`` will result in undefined behavior.
```
seberg, line 644: Curious, this is a problem even when the dimensions of the array actually match?  (E.g. due to the calling convention passing the `mdspan` differently and thus mangling things up?)

cupy/_core/core.pyx
```diff
@@ -2215,6 +2273,29 @@ cdef class _ndarray_base:
...
 
 
+cdef inline _carray.mdspan _mdspan_from_ndarray(
+        _ndarray_base arr, index_type, bint allow_unsafe):
+    # Creates mdspan from ndarray.
+    # Note that this function cannot be defined in _carray.pxd because that
+    # would cause cyclic cimport dependencies.
+    cdef _carray.mdspan carr = _carray.mdspan.__new__(_carray.mdspan)
+    cdef int index_itemsize
+    if index_type == cupy.int32:
```
seberg: The speed discussion (from elsewhere) made me realize that technically, this should use `get_type(index_type) == cupy.int32` maybe?  Then we also allow writing `"int32"`, but mainly we can't run into problems because `np.int != np.long` even when they are both 32bit.

cupy/cuda/function.pxd
```diff
@@ -2,7 +2,8 @@ from libc.stdint cimport intptr_t
 
 
 cdef class CPointer:
-    cdef void* ptr
+    cdef:
+        readonly intptr_t ptr
```
seberg, line 6: Hah, yeah, this should be fine and nice!

tests/cupy_tests/core_tests/test_mdspan.py
```diff
@@ -0,0 +1,794 @@
...
+        extents=extents,
+        layout=layout,
+        body=body
+    )
+
+
+# Kernel with compile-time size validation
+code_verify_with_validation = r"""
+#include <cuda/std/mdspan>
+#include <cupy/carray.cuh>  // TODO(leofang): replace this by fp16 header once available
```
seberg, line 109: Ah yeah, this is the type of things we'll break in the other PR now :).

seberg (commented):
I lost a bit focus on the tests, but this looks good to me. Adding `mdspan` seems fine.  And I think I'll brush away my API concerns, by considering that we can add a `layout="auto"` in some future (and if we care a lot, we can probably pull off transitioning to it as the default too).

A few small comments, nits otherwise.

cupy/_core/core.pyx
kmaehashi: Wondering it would be better to make `index_type` and `allow_unsafe` kwarg-only, just in case if we want to be ready to change the API in the future.

cupy/_core/_carray.pyx
leofang: The public-facing input to influence this is `index_type`. This is internal-only.

cupy/_core/core.pyx
leofang: Yes. When extents are static, they become part of the mdspan type, and have no physical memory backing. So the memory layout would become incompatible (`(ptr, strides)` in the case of `layout_stride` instead of `(ptr, extents, strides)`). 

We could consider relaxing this in a follow-up PR, since it is compatible with `layout_{left/right}`. But I don't want to overcomplicate the API for now (and I agree with Kenichi above that we should use kw-only args).

cupy/_core/_carray.pyx
leofang: Ah yes this reminds me I haven't added the support for auto-selecting int32/int64.

cupy/_core/_carray.pyx
leofang: Yeah I follow the same pattern in other CPointer subclasses. No harm in keeping these.

cupy/_core/_carray.pyx
leofang: All subclasses in this file are private classes. The only way to construct an instance is through public APIs like `ndarray.mdspan()`.

cupy/_core/_carray.pxd
leofang: See https://github.com/cupy/cupy/pull/9525/files#r2717187833. Also `__init__` cannot be added to `.pxd` IIRC.

cupy/_core/_carray.pyx
seberg: Sorry, I was wondering about an `index_itemsize` attribute here.  But let's defer that for now.

tests/cupy_tests/core_tests/test_mdspan.py
leofang: yup once #9494 lands we can change this

cupy/_core/core.pyx
seberg: Sorry, I meant to delete some of these comments from yesterday and forgot (I think github didn't show them).

cupy/_core/_carray.pyx
seberg: FWIW, if we default to int64 (which is sane) I think we can punt on this!

cupy/_core/core.pyx
mhoemmen: My view is that it could make sense to separate "just give me a correct `mdspan` with the most general layout mapping possible" from "please check the preconditions for converting this `layout_stride_relaxed` `mdspan` to an `mdspan` with a more specific layout mapping, and do the conversion if the preconditions are met."

I say this because it sounds like we care a lot about performance of conversion to `mdspan`.  It's helpful to know that the conversion will always succeed.  Some users might not worry too much about the layout mapping being so general.

cupy/_core/_carray.pyx
mhoemmen: @seberg wrote:
> FWIW, if we default to int64 (which is sane) I think we can punt on this!

This is reasonable.  Developers can check extents on the C++ side and convert their `int64` `mdspan` to an `int32` `mdspan` by invoking `mdspan`'s converting constructor (which is `explicit` in this case).

cupy/_core/_scalar.pyx
```diff
@@ -1,3 +1,5 @@
+from libc.stdint cimport intptr_t
```
mhoemmen, line 1: Should we consider using `uintptr_t` instead of `intptr_t`?  The usual representation of pointers as integers in C++ is `uintptr_t`.

cupy/_core/_carray.pyx
leofang: Sorry, I changed my mind and as of commit b864f86c5 this
> Ah yes this reminds me I haven't added the support for auto-selecting int32/int64.

is not pursued. Instead, as part of the move to use kw-only args I require the type to be given explicitly, since when this method is called we have 0 visibility into what could be required on the C++ side, and users should take the full responsibility. Once we have a safer way to detect it in the future, this restriction can be relaxed without breaking APIs or semantics (which is the benefit of kw-only args).

> Since we have one error, maybe add a check for too large values as well

This is done.

cupy/_core/core.pyx
leofang: > Wondering it would be better to make `index_type` and `allow_unsafe` kwarg-only,

This is done as of commit b864f86c5. I will table the rest of this thread (which is great suggestion, btw!) to a later PR.

leofang: (tracked in #9627)

cupy/_core/_scalar.pyx
leofang: see https://github.com/cupy/cupy/pull/9525#discussion_r2714685158

cupy/_core/_carray.pyx
leofang: I added a simple check in commit b864f86c, let me know if it is not enough

seberg:
(timeouts here as well)

/test mini

cupy/_core/_carray.pyx
seberg: I guess technically we could still overflow a single dimension if we have a shape of `(0, 2**34)` or so. But I'm happy to not worry about it (at least not in this iteration).

leofang: Zero-extent would be blocked by default (`allow_unsafe=False`). If flipped to `True`, yeah it could overflow but users express the intent for us to not do any checks, so it should be fine I think.

mhoemmen: @seberg I would recommend going to the C++ Standard Draft for mdspan information, rather than cppreference.  cppreference has been read-only for a while so we haven't been able to improve the documentation there.

Here is the mdspan section in the Draft: https://eel.is/c++draft/views.multidim

mhoemmen: @seberg wrote:
> ... apparently that is the technical requirement that the size has to fit the corresponding unsigned int....

It is an important feature that users be able to construct an `extents` object where each extent fits in `index_type`, but the product of the extents does not necessarily fit in `index_type`.  It's the layout mappings that impose the size requirement.  See e.g., [[mdspan.layout.left.overview-4]](https://eel.is/c++draft/views.multidim#mdspan.layout.left.overview-4) and [[mdspan.layout.left.cons-1]](https://eel.is/c++draft/views.multidim#mdspan.layout.left.cons-1).

leofang:
Fixed test failures due to changing `CPointer.ptr` from `void*` to `intptr_t`.

/test mini

kmaehashi:
LGTM. Ignoring timeouts.
/test skip

PR #9535, opened 2025-12-18
Fix piecewise to handle list of boolean conditions

megha-darda:
This PR will fix: https://github.com/cupy/cupy/issues/9267
cupy.piecewise returned incorrect results when condlist contained booleans in a list (e.g., [True, False]). Now, such lists are properly handle boolean arrays, matching NumPy behavior. Also added test case for it.

I picked up this issue as a way to get started with contributing to CuPy and to better understand its codebase. Please let me know if there are improvements I can make to this PR or if there are other small issues/features I could contribute to next.

Thank you.

### DIFF
--- a/cupy/_functional/piecewise.py
+++ b/cupy/_functional/piecewise.py
@@ -35,9 +35,13 @@ def piecewise(x, condlist, funclist):
 
         .. seealso:: :func:`numpy.piecewise`
         """
-    if cupy.isscalar(condlist):
+    x = cupy.asanyarray(x)
+
+    if cupy.isscalar(condlist) or (
+            not isinstance(condlist[0], (list, cupy.ndarray)) and x.ndim != 0):
         condlist = [condlist]
 
+    condlist = cupy.asarray(condlist, dtype=bool)
     condlen = len(condlist)
     funclen = len(funclist)
     if condlen == funclen:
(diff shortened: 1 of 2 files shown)


cupy/_functional/piecewise.py
```diff
@@ -35,9 +35,13 @@ def piecewise(x, condlist, funclist):
 
         .. seealso:: :func:`numpy.piecewise`
         """
-    if cupy.isscalar(condlist):
+    x = cupy.asanyarray(x)
+
+    if cupy.isscalar(condlist) or (
+            not isinstance(condlist[0], (list, cupy.ndarray)) and x.ndim != 0):
```
seberg, line 41: I guess this is came straight from NumPy. But to me this looks just like pretty bad guess work to me.

Do you use this function at all? I don't really appreciate the guess work, maybe it would be nice to actually fix more in NumPy (first/as well).

I would be tempted to try and deprecate all ambiguous paths, although I am not 100% sure what those are.  My suspicion would be that this is if `asarray(condlist[0]).ndim != x.ndim` (which actually implies the shape must match exactly).
Having a smaller ndim works, but is not documented in what it does and probably also differs in behavior between NumPy and CuPy.  So I would be tempted to just deprecate it _in_ NumPy (which would also deprecate this path, although the path is scarier, because it may or may not kick in in-practice).

megha-darda: Thanks so much for the feedback, that makes a lot of sense. I agree that the current logic is heuristic and not very explicit about the intended behavior.

For this PR, my main goal was to address the specific NumPy incompatibility reported in the issue by aligning CuPy with NumPy’s current behavior, rather than to redesign how ambiguous condition handling works. I understand this may not be the ideal long-term solution.

I’d be happy to adjust or drop this approach if it’s not suitable, and I’d really appreciate any suggestions for smaller, well-scoped issues tests, documentation, or parity fixes that could be a good starting point for a first contribution. I’m contributing as part of a coursework assignment and want to learn more about the CuPy codebase while making a helpful, manageable contribution.

PR #9536, opened 2025-12-19
MAINT: Use only union for half-infinity constexpr

seberg:
GCC seems to accept the reinterpret cast, but clang refuses it as it is technically not a constexpr.
The nicest thing would be `std::bit_cast`, but while I am not sure that the union is the best solution, it seems like a good enough solution that should work without the need for `#ifdefs` (I don't see a reason why the `#ifdef` would be helpful).

(This was the second small diff to make compiling with clang work locally for me.)

### DIFF
--- a/cupy/cuda/cupy_cub.cu
+++ b/cupy/cuda/cupy_cub.cu
@@ -115,23 +115,21 @@ class numeric_limits<__half> {
   public:
     static __host__ __device__ constexpr __half infinity() noexcept {
         unsigned short inf_half = 0x7C00U;
-        #if (defined(_MSC_VER) && _MSC_VER >= 1920)
-        #if CUDA_VERSION < 11030
-        // WAR: CUDA 11.2.x + VS 2019 fails with __builtin_bit_cast
+        // With C++20, should use std::bit_cast, until then use __builtin_bit_cast when
+        // we know it is supported.
+        // Otherwise use a union (which is technically not standard compliant).
+        #if (  \
+            (defined(_MSC_VER) && _MSC_VER >= 1920) ||  \
+            (defined(__clang__) && __clang_major__ >= 9) || \
+            (defined(__GNUC__) && (__GNUC__ > 11 || (__GNUC__ == 11 && __GNUC_MINOR__ >= 1)))  \
+        )
+        return __builtin_bit_cast(__half, inf_half);
+        #else
         union caster {
             unsigned short u_;
             __half h_;
         };
         return caster{inf_half}.h_;
-        #else  // CUDA_VERSION < 11030
-        // WAR:
-        // - we want a constexpr here, but reinterpret_cast cannot be used
-        // - we want to use std::bit_cast, but it requires C++20 which is too new
-        // - we use the compiler builtin, fortunately both gcc and msvc have it
-        return __builtin_bit_cast(__half, inf_half);
-        #endif
-        #else
-        return *reinterpret_cast<__half*>(&inf_half);
         #endif
     }
 


leofang:
I guess this would be a question for @kmaehashi. If CuPy can move to use VS2022+ On Windows, I think we can use `std::bit_cast` unconditionally?

kmaehashi:
"bit_cast by union" is an undefined behavior in terms of the C++ standard:

https://en.cppreference.com/w/cpp/language/union.html
> It is undefined behavior to read from the member of the union that wasn't most recently written.

The current code was added (by me 😛) under "but it works" basis since otherwise there were no way to workaround the compiler issue which happens only in extremely specific environment (Windows + VS 2019 + CUDA 11.2). So I didn't expect this code path to be used universally.

> If CuPy can move to use VS2022+ On Windows, I think we can use `std::bit_cast` unconditionally?

This also affects Linux builds. We currently use RHEL 7 for CUDA 12 wheel build (`manylinux2014`) which use g++ 7 that only supports `-std=c++17`.

leofang:
> This also affects Linux builds. We currently use RHEL 7 for CUDA 12 wheel build (`manylinux2014`) which use g++ 7 that only supports `-std=c++17`.

~~Ah, if this helps the decision making: since CUDA 12.x (forget which minor version) the support for glibc 2.17 is dropped and now 2.28 is the minimal requirement. As a result, CUDA Python has bumped up our minimal supported glibc since around this summer.~~

EDIT: I had a brain fart. The glibc version is irrelevant. We want `std::bit_cast` but it's C++20. We need to support C++17 as the minimum.

seberg:
Heh, undefined behavior is a funny monster, seems a bit unlikely that a compiler might do something undefined here, but let me see if we can't detect whether `std::bit_cast` or `__builtin_bit_cast` is available in a nice way. That would be the best... (there is a `memcpy` work-around, but I doubt that is strictly `constexpr` compatible.)

leofang:
> but let me see if we can't detect whether `std::bit_cast` or `__builtin_bit_cast` is available in a nice way. 

@seberg maybe let's just check if Clang has `__builtin_bit_cast`? Both GCC and MSVC support it (see the inline comment removed in this PR).

seberg:
OK, I just hardcoded known good compiler versions to use `__builtin_bit_cast`, I guess in practice we may not care about any older ones. Tested via: [godbold](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:16,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:24,positionColumn:2,positionLineNumber:24,selectionStartColumn:2,selectionStartLineNumber:24,startColumn:2,startLineNumber:24),source:'%23include+%3Ccstdint%3E%0A%0Afloat+func(int32_t+j)+%7B%0A++++++++//+With+C%2B%2B20,+should+use+std::bit_cast,+until+then+use+__builtin_bit_cast+when%0A++++++++//+we+know+it+is+supported.+Otherwise+use+a+union+(technically+not+standard+compliant).%0A++++++++//+WAR:+CUDA+11.2.x+%2B+VS+2019+fails+with+__builtin_bit_cast%0A++++++++%23if+(++%5C%0A++++++++++++(defined(_MSC_VER)+%26%26+_MSC_VER+%3E%3D+1920)+%7C%7C++%5C%0A++++++++++++(defined(__clang__)+%26%26+__clang_major__+%3E%3D+9)+%7C%7C+%5C%0A++++++++++++(defined(__GNUC__)+%26%26+(__GNUC__+%3E+11+%7C%7C+(__GNUC__+%3D%3D+11+%26%26+__GNUC_MINOR__+%3E%3D+1)))++%5C%0A++++++++)%0A++++++++return+__builtin_bit_cast(float,+j)%3B%0A++++++++%23else%0A++++++++union+caster+%7B%0A++++++++++++int32_t+u_%3B%0A++++++++++++float+h_%3B%0A++++++++%7D%3B%0A++++++++return+caster%7Bj%7D.h_%3B%0A++++++++%23endif%0A%7D%0A%0Aint+main(void)+%7B%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:51.24782356355194,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:arm64g1050,filters:(b:'0',binary:'0',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-O2',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+ARM64+gcc+10.5.0+(Editor+%231)',t:'0')),k:50,l:'4',m:53.58480470840021,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'NVCC+11.7.0',editorid:1,fontScale:14,fontUsePx:'0',j:2,wrap:'1'),l:'5',n:'0',o:'Output+of+ARM64+gcc+10.5.0+(Compiler+%232)',t:'0')),header:(),l:'4',m:46.41519529159979,n:'0',o:'',s:0,t:'0')),k:48.75217643644806,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).
I think the CUDA 11 path shouldn't matter anymore (i.e. we dropped it).

PR #9539, opened 2025-12-20
Implement hazen, weibull, median_unbiased, normal_unbiased quantile m…

Harishjitu:
…ethods

Add support for four parametric quantile estimation methods from Hyndman & Fan (1996):
- `hazen` (H&F type 5): α=0.5, β=0.5
- `weibull` (H&F type 6): α=0, β=0  
- `median_unbiased` (H&F type 8): α=1/3, β=1/3
- `normal_unbiased` (H&F type 9): α=3/8, β=3/8

These methods use the general formula:
`virtual_index = q * (n - α - β + 1) + α - 1`

## Changes
- Added `_QUANTILE_PARAMS` dict in `cupy/_statistics/order.py`
- Implemented the four methods using shared formula
- Enabled existing tests for these methods in `test_order.py`

Partial fix for #6229

### DIFF
--- a/tests/cupy_tests/statistics_tests/test_order.py
+++ b/tests/cupy_tests/statistics_tests/test_order.py
@@ -173,9 +173,8 @@ class TestQuantileMethods:
                 with pytest.raises(ValueError):
                     xp.percentile(a, q, axis=-1, method=method)
 
-    @testing.for_all_dtypes()
     @testing.for_all_dtypes(no_float16=True, no_bool=True, no_complex=True)
-    @testing.numpy_cupy_allclose()
+    @testing.numpy_cupy_allclose(rtol=1e-6)
     def test_quantile_defaults(self, xp, dtype, method):
         a = testing.shaped_random((2, 3, 8), xp, dtype)
         q = testing.shaped_random((3,), xp, scale=1)


leofang (approved):
LGTM! Thanks for your contribution! Will trigger the CI once we rebuild the cache.

leofang:
@Harishjitu could you check the CI failures (ignore cuda131 for now)? Seems like a tolerance issue.

Harishjitu:
The CI failures were due to dtype mismatch with numpy. Fixed it to match numpy behavior
@leofang could you please take a look?

leofang:
Hi @Harishjitu, sorry for late reply. The latest commit looks odd to me: The test suite does already check if returned dtypes match between CuPy/NumPy, so it's not a matter of dtype mismatch, but simply a tolerance issue. `atol` and/or `rtol` in the failing tests' `@testing.numpy_cupy_allclose` decorator should be slightly relaxed. I see that the same file already does this in a few places, could you take a look?

Harishjitu:
@leofang You are right!

I made the fix.

I also removed a duplicate decorator, please check.

leofang (approved):

leofang:
Thanks for your contribution @Harishjitu!

Issue #8716, opened 2024-10-31
cupyx.scipy.signal.resample_poly fails for 2D arrays


naperman:
### Description

For 2D input arrays, `cupyx.scipy.signal.resample_poly` works for small upsampling rates (see Example 1), but fails for large upsampling rates( see Example 2).

### To Reproduce

```py
# Example 1 (works)
import cupy as cp
from cupyx.scipy.signal import resample, resample_poly

x = cp.linspace(0, 10, 32, endpoint=False)
z = cp.cos(-x**2/6.0) + 1j*cp.sin(-x**2/6.0)
y = cp.concatenate((z[cp.newaxis,...], z[cp.newaxis,...]))
print(f"y.shape = {y.shape}")

f_fft = resample(y, 96, axis=1)
f_poly = resample_poly(y, 3, 1, axis=1)
xnew = cp.linspace(0, 10, 96, endpoint=False)

plt.plot(cp.asnumpy(xnew), cp.asnumpy(cp.squeeze(f_fft[0,:])), 'b.-', cp.asnumpy(xnew), cp.asnumpy(cp.squeeze(f_poly[0,:])), 'r.-')
plt.plot(cp.asnumpy(x), cp.asnumpy(cp.squeeze(y[0,:])), 'ko-')
plt.plot(10, cp.asnumpy(y[0,0]), 'bo', 10, 0., 'ro')  # boundaries
plt.legend(['resample', 'resamp_poly', 'data'], loc='best')
plt.grid(True)
plt.show()

plt.plot(cp.asnumpy(xnew), cp.asnumpy(cp.squeeze(f_fft[1,:])), 'b.-', cp.asnumpy(xnew), cp.asnumpy(cp.squeeze(f_poly[1,:])), 'r.-')
plt.plot(cp.asnumpy(x), cp.asnumpy(cp.squeeze(y[1,:])), 'ko-')
plt.plot(10, cp.asnumpy(y[1,0]), 'bo', 10, 0., 'ro')  # boundaries
plt.legend(['resample', 'resamp_poly', 'data'], loc='best')
plt.grid(True)
plt.show()

```

```py
# Example 2 (does not work)
import cupy as cp
from cupyx.scipy.signal import resample, resample_poly

x = cp.linspace(0, 10, 32, endpoint=False)
A = cp.zeros([8, 32], dtype=cp.complex64)

A[:,...] = cp.cos(-x**2/6.0) + 1j*cp.sin(-x**2/6.0)

[...]

PR #9540, opened 2025-12-20
Add N-D array support to upfirdn

Harishjitu:
Extend cupyx.scipy.signal.upfirdn to support N-dimensional arrays, matching SciPy's behavior. Previously limited to 1D/2D arrays.

Implementation:
- Extract 2D kernel code into _apply_filter_2d helper method
- For N-D arrays: move target axis to end, reshape to 2D, apply existing kernel, reshape back to original dimensions

Ref #8716


leofang (approved):

leofang:
cuda120 CI failure seems irrelevant, and other failures will be fixed in #9551 and #9562.

PR #9549, opened 2025-12-24
Support NumPy 2.4

asi1024:


cupy/_statistics/order.py
```diff
@@ -328,6 +327,7 @@ def percentile(a, q, axis=None, out=None,
     if interpolation is not None:
         method = _check_interpolation_as_method(
             method, interpolation, 'percentile')
+    dtype = numpy.common_type(a)
```
seberg: Hmmm, this is slightly surprising to me. Since I guess it now ignores the dtype of `q`?
That is correct for `nearest`, `lower`, and `higher` in principle, but it seems that has been ignored for a long time, so not relevant for NumPy 2.4.

If `q` is not floating is the problem that the result dtype is inferred as integral?  Maybe we could add `if method (not) in ...: np.result_type(a, q, 0.0)` to work around that.

But right now I am not sure sure what changed here that needs an update.

kmaehashi: @asi1024 Was this the one you mentioned in our offline discussion?: https://github.com/numpy/numpy/pull/29105

asi1024: Yes, the change in the dtype of the output array is likely due to numpy/numpy#29105. It is not highlighted in the NumPy 2.4 release notes.

seberg: I'll investigate, I didn't recall this change. It seems meant to just bump precision a little for `float16`/`float32` arrays to preserve precision for intermediate values derived from `q`.

But, I think that PR is buggy and incorrectly and unintentinoally changed the result type, where the precision of `q` itself was previously taken into account.

EDIT: OK, I think the return type change is just a NumPy bug to be fixed there, but we should probably do the same bump of the internal `q` precision.

leofang: xref: https://github.com/numpy/numpy/issues/30586

seberg: Just to update, I have a change for NumPy to revert, although it might also land on calling it an intentional change, in which case this seems good (not quite sure if it is 100% identical, but I am dubious of the NumPy logic as well).

EDIT: 2.4.1 should hit this weekend, so once it does, we can do whichever is included there.

seberg: OK, this has been reverted. Glancing at the code, I think the behavior in 2.4.1 should now be closer to what was in CuPy than previously (i.e. hopefully now change is necessary).
We could tweak the `q` logic to read `if type(q) in (int, float)` rather than checking for tuple/list change (because `int` and `float` are what is special with NEP 50).

seberg: Sorry, I made things worse in NumPy by incorrectly reverting one line... I wish I had tested it against CuPy then.

I have pushed a commit here that patches NumPy, since I suspect we have to fix that there either way (links in the code).  If something else happens, I guess we will notice it in test failures eventually and fix it here.

seberg:
Seems timeouts (some of them when things were basically done).

/test mini

seberg:
OK, the two failing CIs at least ran to 99% or 91% without errors, so I think it's OK to skip them.  Might be good if someone else can have a brief scroll-through, but I think it is good now.

(I added a monkey-patch to make tests pass on NumPy 2.4.1 explicitly, kept it minimal, no changes to quantile for now, that would make sense, but I don't think it's needed here.)

/test force-skip

.pfnci/matrix.yaml
```diff
@@ -275,7 +275,7 @@
   cutensor: "2.4"
   cusparselt: "0.8.1"
   python: "3.14"
-  numpy: "2.3"
+  numpy: "2.4"
```
leofang, line 278: nit: by our convention this should be applied to CUDA 13.1 now, not 13.0

asi1024:
@seberg Thank you for updating the tests for NumPy 2.4.1! I have modified the test to use a fixture so that the change does not affect other test cases.

/test mini

seberg:
The two runs timed out at 97% and 72% respectively, just skip probably?

PR #9552, opened 2025-12-24
Fix dtype error when float32 used for cupyx.scipy.signal.square

### LINKED ISSUE #9541: signal.square fails with float32 inputs
### Description

See https://github.com/scipy/scipy/actions/runs/20399317235/job/58619346597?pr=24201
```
FAILED scipy/signal/tests/test_waveforms.py::TestSquareWaveform::test_dtype[cupy] - cupy.cuda.compiler.CompileException: /tmp/tmp5hvbace9/f645fd6891a1c3fa7c03a5d8e659fa0bfc2a49ad.cubin.cu(23): error: more than one instance of overloaded function "fmod" matches the argument list:
            function "fmod(float, float)" (declared implicitly)
            function "fmod(double, double)" (declared implicitly)
            argument types are: (const T, double)
      const T tmod { fmod( t, 2.0 * M_PI ) };
```

### To Reproduce

```py
# Write the code here
def test_dtype(self, xp):
        waveform = square(xp.asarray(1, dtype=xp.float32),
                                    duty=xp.asarray(0.5, dtype=xp.float32))
        assert waveform.dtype == xp.float64

        waveform = square(xp.asarray(1, dtype=xp.float64))
        assert waveform.dtype == xp.float64
```
This test from SciPy fails with CuPy and the implementation of `square` delegates to CuPy

### Installation

### Environment

```
# Paste the output here
```
Sorry this in a CI job and I don't have access to a machine with CuPy

### Additional Information

_No response_

BarnikRB:
Addresses https://github.com/cupy/cupy/issues/9541. When period is double and time in float, cuda fmod throw error, so this fix ensures that period has same type as the inputs. Also added dtype tests as in original issue.


tests/cupyx_tests/scipy_tests/signal_tests/test_waveforms.py
```diff
@@ -188,6 +188,15 @@ def test_square(self, duty, xp, scp):
         t = xp.linspace(0, 1, 500)
         return scp.signal.square(t, duty)
 
+    @pytest.mark.parametrize("dtype_t", [cupy.float32, cupy.float64])
```
leofang: Could you use `@testing.for_float_dtypes()` instead? I dunno if SciPy supports fp16, if not we may want to set `no_float16=True`.

tests/cupyx_tests/scipy_tests/signal_tests/test_waveforms.py
```diff
@@ -188,6 +188,15 @@ def test_square(self, duty, xp, scp):
...
+    def test_square_dtype(self, dtype_t):
```
leofang: I am thinking this test can be combined with the one above. Could you check?

leofang (changes requested):

tests/cupyx_tests/scipy_tests/signal_tests/test_waveforms.py
j-bowhay: No SciPy doesn't support `fp16`

BarnikRB:
@leofang I have addressed your comments.

tests/cupyx_tests/scipy_tests/signal_tests/test_waveforms.py
```diff
@@ -182,11 +182,21 @@ def test_sawtooth(self, width, xp, scp):
...
     @pytest.mark.parametrize('duty', [1.0, 0.5, 3.0])
-    @testing.numpy_cupy_allclose(scipy_name="scp")
-    def test_square(self, duty, xp, scp):
-        t = xp.linspace(0, 1, 500)
-        return scp.signal.square(t, duty)
+    @testing.for_float_dtypes(no_float16=True)
+    @testing.numpy_cupy_allclose(scipy_name='scp')
+    def test_square(self, duty, dtype, xp, scp):
+        t = xp.linspace(0, 1, 500, dtype=dtype)
+        duty_arr = xp.asarray(duty, dtype=dtype)
```
leofang: The original test covers `duty` as a Python scalar. Why do we need to convert it to an array here? It's already done in the `square` implementation.

tests/cupyx_tests/scipy_tests/signal_tests/test_waveforms.py
```diff
@@ -182,11 +182,21 @@ def test_sawtooth(self, width, xp, scp):
...
+
+        y = scp.signal.square(t, duty=duty_arr)
+
+        # CuPy-specific dtype rule
+        if xp is cupy:
+            assert y.dtype == cupy.float64
```
leofang: I believe the allclose decorator also does type checks, so if the test passes then it means we align with SciPy's return type. I see no need to do this extra check. WDYT?

cupyx/scipy/signal/_waveforms.py
```diff
@@ -141,8 +141,9 @@ def sawtooth(t, width=1.0):
         y = nan("0xfff8000000000000ULL");
     }
 
-    const T tmod { fmod( t, 2.0 * M_PI ) };
-    const bool mask2 { ( ( 1 - mask1 ) && ( tmod < ( w * 2.0 * M_PI ) ) ) };
+    const T period = (T)(2.0 * M_PI);
```
leofang: nit: C++ style casting + constexpr
```suggestion
    constexpr T period = static_cast<T>(2.0 * M_PI);
```

leofang (changes requested):

tests/cupyx_tests/scipy_tests/signal_tests/test_waveforms.py
BarnikRB: I tried to combine my initial test function for dtype with this one and for that i required to change dtype of duty for which i converted it to array. but now i will modify using dtype() directly

tests/cupyx_tests/scipy_tests/signal_tests/test_waveforms.py
BarnikRB: Yes, I also agree with you.

PR #9555, opened 2025-12-28
Fix conda compiler detection

leofang:
A minor refactoring is done to ensure the fix can be applied to all places.

### DIFF
--- a/install/cupy_builder/install_build.py
+++ b/install/cupy_builder/install_build.py
@@ -1,8 +1,8 @@
 # mypy: ignore-errors
 from __future__ import annotations
 
-
 import contextlib
+import functools
 import logging
 import os
 import platform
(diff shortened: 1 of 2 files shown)


install/cupy_builder/install_build.py
```diff
@@ -684,14 +684,23 @@ def conda_get_target_name():
     return out
 
 
+@functools.cache
+def is_conda_cross_compiling() -> bool:
+    is_cross_compiling = os.environ.get('CONDA_BUILD_CROSS_COMPILATION', '0')
+    # Recent conda compilers might set CONDA_BUILD_CROSS_COMPILATION to empty.
+    if is_cross_compiling == '':
+        is_cross_compiling = '0'
```
leofang, line 692: This is the fix (to handle the case when the env var is set to an empty string).

PR #9557, opened 2025-12-30
Fix intersect1d crash with empty arrays

### LINKED ISSUE #8937: Empty array cuases `intersect1d` to crash Cupy
### Description

If the second argument of `intersect1d` is empty, this will cause `CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered` error and any subsequent use of Cupy (e.g. `cp.array([0])`) will also cause the same error.

### To Reproduce

```py
import cupy as cp
cp.intersect1d(cp.array([0]), cp.array([]))
```

### Installation

### Environment

```
OS                           : Linux-6.8.0-41-generic-x86_64-with-glibc2.39
Python Version               : 3.12.2
CuPy Version                 : 13.3.0
NumPy Version                : 2.2.2
Cython Build Version         : 0.29.37
CUDA Build Version           : 12060
CUDA Runtime Version         : 12060 (linked to CuPy) / RuntimeError('CuPy failed to load libcudart.so.12: OSError: libcudart.so.12: cannot open shared object file: No such file or directory') (locally installed)
CUDA Extra Include Dirs      : ['/home/theyoke/miniconda3/targets/x86_64-linux/include', '/home/theyoke/miniconda3/include']
cuBLAS Version               : 120803
cuFFT Version                : 11303
cuSOLVER Version             : (11, 7, 2)
Thrust Version               : 200500
CUB Build Version            : 200600
Device 0 Name                : NVIDIA GeForce RTX 3060
Device 0 PCI Bus ID          : 0000:0B:00.0
```

### Additional Information

_No response_

Harishjitu:
Fix illegal memory access when either input array is empty. The CUDA kernel was accessing non-existent memory, corrupting the CUDA context for subsequent operations.

Add early return with proper dtype handling using result_type() to match NumPy behavior.

Fixes #8937

### DIFF
--- a/cupy/_logic/truth.py
+++ b/cupy/_logic/truth.py
@@ -162,6 +162,7 @@ def intersect1d(arr1, arr2, assume_unique=False, return_indices=False):
     numpy.intersect1d
 
     """
+    common_dtype = cupy.result_type(arr1.dtype, arr2.dtype)
     if not assume_unique:
         if return_indices:
             arr1, ind1 = cupy.unique(arr1, return_index=True)
@@ -175,11 +176,11 @@ def intersect1d(arr1, arr2, assume_unique=False, return_indices=False):
 
     if not return_indices:
         mask = _search._exists_kernel(arr1, arr2, arr2.size, False)
-        return arr1[mask]
+        return arr1[mask].astype(common_dtype)
 
     mask, v1 = _search._exists_and_searchsorted_kernel(
         arr1, arr2, arr2.size, False)
-    int1d = arr1[mask]
+    int1d = arr1[mask].astype(common_dtype)
     arr1_indices = cupy.flatnonzero(mask)
     arr2_indices = v1[mask]
 
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -365,6 +365,12 @@ _exists_kernel = _core.ElementwiseKernel(
     'S x, raw T bins, int64 n_bins, bool invert',
     'bool out',
     '''
+    // Handle empty bins array to avoid illegal memory access
+    if (n_bins == 0) {
+        out = (invert ? true : false);
+        return;
+    }
+
     const bool assume_increasing = true;
     const bool side_is_right = false;
     long long y;
@@ -379,6 +385,13 @@ _exists_and_searchsorted_kernel = _core.ElementwiseKernel(
     'S x, raw T bins, int64 n_bins, bool invert',
     'bool out, int64 y',
     '''
+    // Handle empty bins array to avoid illegal memory access
+    if (n_bins == 0) {
+        out = (invert ? true : false);
+        y = 0;
+        return;
+    }
+
     const bool assume_increasing = true;
     const bool side_is_right = false;
     '''
--- a/tests/cupy_tests/logic_tests/test_truth.py
+++ b/tests/cupy_tests/logic_tests/test_truth.py
@@ -263,6 +263,36 @@ class TestIntersect1d:
         b = xp.array([4, 6, 2, 5, 7, 6], dtype=dtype)
         return xp.intersect1d(a, b, return_indices=True)
 
+    @testing.numpy_cupy_array_equal()
+    def test_intersect1d_both_empty(self, xp):
+        return xp.intersect1d(xp.array([]), xp.array([]))
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_intersect1d_empty_array(self, xp, dtype):
+        a = xp.array([], dtype=dtype)
+        b = xp.array([0], dtype=dtype)
+        return xp.intersect1d(a, b, return_indices=True)
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_intersect1d_second_empty_array(self, xp, dtype):
+        a = xp.array([0], dtype=dtype)
+        b = xp.array([], dtype=dtype)
+        return xp.intersect1d(a, b, return_indices=True)
+
+    @testing.numpy_cupy_array_equal()
+    def test_intersect1d_mixed_dtypes_empty(self, xp):
+        a = xp.array([0], dtype=xp.int64)
+        b = xp.array([], dtype=xp.float64)
+        return xp.intersect1d(a, b)
+
+    @testing.numpy_cupy_array_equal()
+    def test_intersect1d_mixed_dtypes_empty_with_indices(self, xp):
+        a = xp.array([0], dtype=xp.int64)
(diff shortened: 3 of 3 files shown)


leofang (changes requested):
Hi @Harishjitu Thanks for your contribution! I have two comments:
1. Could you please add a test, similar to [NumPy's](https://github.com/numpy/numpy/blob/c89cf2cd367bfa97e5c96ce90d3ae4534892a0c3/numpy/lib/tests/test_arraysetops.py#L36) and what's added in #8997, so that we capture this edge case?
2. Could you please try to avoid special-casing at the Python-level, and fix the kernel instead? @zazabap gave a patch in the #8997 PR description. I was able to discuss this with Copilot, which gave a similar patch (I did not verify, so please kindly do so before checking in the changes 🙂).

<details>
<summary>CLICK ME for the chat record</summary>

Here's the diff to fix the kernels in `cupy/_sorting/search.py`:

```diff
diff --git a/cupy/_sorting/search.py b/cupy/_sorting/search.py
index 7f0731d.. XXXXXXX 100644
--- a/cupy/_sorting/search. py
+++ b/cupy/_sorting/search.py
@@ -257,6 +257,11 @@ _hip_preamble = r'''
 
 _searchsorted_code = '''
     #ifdef __HIP_DEVICE_COMPILE__
     bool is_done = false;
     #endif
+
+    // Handle empty bins array to avoid illegal memory access
+    if (n_bins == 0) {
+        no_thread_divergence( y = 0 , true )
+    }
 
     // Array is assumed to be monotonically
@@ -364,10 +369,15 @@ _hip_preamble = r'''
 _exists_kernel = _core. ElementwiseKernel(
     'S x, raw T bins, int64 n_bins, bool invert',
     'bool out',
     '''
+    // Handle empty bins array to avoid illegal memory access
+    if (n_bins == 0) {
+        out = (invert ?  true : false);
+        return;
+    }
+
     const bool assume_increasing = true;
     const bool side_is_right = false;
     long long y;
     '''
     + _searchsorted_code + '''
@@ -378,10 +388,15 @@ _exists_kernel = _core. ElementwiseKernel(
 _exists_and_searchsorted_kernel = _core. ElementwiseKernel(
     'S x, raw T bins, int64 n_bins, bool invert',
     'bool out, int64 y',
     '''
+    // Handle empty bins array to avoid illegal memory access
+    if (n_bins == 0) {
+        out = (invert ? true : false);
+        y = 0;
+        return;
+    }
+
     const bool assume_increasing = true;
     const bool side_is_right = false;
     '''
     + _searchsorted_code + '''
```

**Key changes:**

1. **In `_searchsorted_code`** (line 258+): Added early return when `n_bins == 0` to set `y = 0` safely
2. **In `_exists_kernel`** (line 367+): Added guard to return `false` (or `true` if inverted) when `n_bins == 0`
3. **In `_exists_and_searchsorted_kernel`** (line 381+): Added guard to return both `out` and `y` safely when `n_bins == 0`

This way, **no changes are needed to `intersect1d` in `truth. py`** - the kernels handle empty arrays gracefully, and the algorithm naturally produces empty results with the correct dtype through array indexing operations like `arr1[mask]`.

This approach truly mirrors NumPy's philosophy where the algorithm handles edge cases implicitly rather than through explicit size checks at the Python level. 

</details>

Harishjitu:
Hi @leofang!
Thanks for the feedback! I've updated the PR:

1. **Fixed the kernels** in `cupy/_sorting/search.py`:
   - Added empty array guard to `_searchsorted_code`
   - Added empty array guard to `_exists_kernel`
   - Added empty array guard to `_exists_and_searchsorted_kernel`

2. **Added tests**:
   - `test_intersect1d_both_empty` — similar to NumPy's `intersect1d([], [])`
   - `test_intersect1d_empty_array` — similar to #8997 (first array empty)
   - `test_intersect1d_second_empty_array` — original bug case (second array empty)

Tested locally and all cases pass.

cupy/_sorting/search.py
```diff
@@ -260,6 +260,11 @@ def argwhere(a):
     bool is_done = false;
     #endif
 
+    // Handle empty bins array to avoid illegal memory access
+    if (n_bins == 0) {
+        no_thread_divergence( y = 0 , true )
+    }
+
```
leofang: Could you check if this special handling is really needed? Since we already touched both kernels below, I suspect this is redundant (and that Copilot made a mistake here -- do not trust any LLM generated code as-is!)

cupy/_logic/truth.py
leofang, line 1: The whole point of modifying the kernels is to avoid special-casing at the Python level. Please undo the change in this file.

cupy/_sorting/search.py
Harishjitu: You are right!
_exists_kernel and _exists_and_searchsorted_kernel already return early before _searchsorted_code runs, so that if check was redundant

I removed that check

cupy/_logic/truth.py
Harishjitu: Fixed

leofang (dismissed):
Thanks, @Harishjitu! I can confirm locally that the illegal instruction is fixed. However, there is one edge case where NumPy and CuPy (with this PR) disagrees in the returned array's dtype:
```python
>>> import cupy as cp
>>> import numpy as np
>>> cp.intersect1d(cp.array([0]), cp.array([]))
array([], dtype=int64)
>>> np.intersect1d(np.array([0]), np.array([]))
array([], dtype=float64)
```
I think it is likely due to the two inputs' dtypes differ (int64 vs float64), but I did not dig deep. If you have any thoughts, feel free to fix. Otherwise, I think this PR does fix the issue and I'd like to get it merged once we sort out the CI issues (discussing offline with the team, cc @kmaehashi for vis).

Harishjitu:
Thanks @leofang 

NumPy's `intersect1d` concatenates both arrays first promoting int64 + float64 → float64
CuPy returns `arr1[mask]` which preserves `arr1`'s dtype.

I actually had this fix in my original PR before moving the crash fix to the kernels.

```python
common_dtype = cupy.result_type(arr1.dtype, arr2.dtype)
result = arr1[mask]
if result.dtype != common_dtype:
    result = result.astype(common_dtype)
return result
```

I could do something like this in `truth.py` and add it to this PR. Do you think this is the correct solution?

leofang:
I see. NumPy does `concatenate` which casts to the same kind:
https://github.com/numpy/numpy/blob/2bc398ed454f47f4d6c9b054d8747046977224b6/numpy/lib/_arraysetops_impl.py#L736
while we use custom kernels to avoid excessive memory usage. I think your suggestion makes sense. Perhaps `.astype()` can be called unconditionally (without the if comparison).

Harishjitu:
@leofang I have made the dtype fix. Please have a look!

PR #9561, opened 2025-12-31
Add `%gpu_timeit` IPython magic

### LINKED ISSUE #9466: Add a Jupyter cell magic that wraps a cell with `cupyx.profiler.benchmark`

leofang:
Takeover of #9511. Close #9466.

### DIFF
--- a/cupyx/profiler/_timeit_magic.py
+++ b/cupyx/profiler/_timeit_magic.py
@@ -1,139 +1,99 @@
-"""IPython magic for GPU-aware timing using cupyx.profiler.benchmark.
-
-To use this extension in IPython/Jupyter:
-    %load_ext cupyx.profiler
-
-Then you can use:
-    %gpu_timeit my_array.sum()
-
-Or as a cell magic:
-    %%gpu_timeit
-    x = cp.random.random((1000, 1000))
-    x @ x.T
-
-You can also pass options to control benchmark behavior:
-    %gpu_timeit -n 100 --max-duration 5 cp.arange(1024**3).sum()
-"""
 from __future__ import annotations
 
-import inspect as _inspect
+import functools
+import inspect
 import textwrap
 
-try:
-    from IPython.core.magic import Magics, magics_class, line_cell_magic
-    from IPython.core.magic_arguments import (
-        argument, magic_arguments, parse_argstring
-    )
-    _IPYTHON_AVAILABLE = True
-except ImportError:
-    _IPYTHON_AVAILABLE = False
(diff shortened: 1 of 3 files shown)


cupyx/profiler/_timeit_magic.py
```diff
@@ -0,0 +1,99 @@
...
+        args = parse_argstring(self.gpu_timeit, line)
+
+        if cell is None:
+            # Line magic
+            code = ' '.join(args.code).strip()
+
+            def run():
+                return eval(code, self.shell.user_ns, local_ns)
+        else:
+            # Cell magic
```
seberg, line 87: It seems that `args.code` is still active but ignored. Maybe add an error for that?
(Mainly, because `%timeit` supports that as setup code I think).

EDIT: For clarity, I mean:
```
%%gpu_timeit -n100 a = cp.ones(100)
a.sum()
```
the `a = cp.ones(100)` is just ignored, but if it was used would be an error (and if it was called, should be setup code).

If it is trivial to do as setup code, that would be even cooler of course.

tests/cupyx_tests/profiler_tests/test_timeit_magic.py
```diff
@@ -0,0 +1,141 @@
...
+            ipython_shell, 'line', '-n 10 -w 2 cp.ones(100).sum()'
+        )
+        _verify_benchmark_output(output)
+
+    def test_cell_magic_simple(self, ipython_shell):
+        """Test cell magic with simple code."""
+        cell_code = """
+x = cp.random.random((100, 100))
+y = x @ x.T
+"""
```
seberg, line 96: Might be tempted to use `textwrap.dedent()`, but doesn't matter.

cupyx/profiler/_timeit_magic.py
leofang: Thanks, Sebastian! I'd like to address this asap. I've been reading how `%timeit` and `%time` were implemented and have some thoughts.  But let me merge and follow up in a separate PR.

leofang: Tracked in https://github.com/cupy/cupy/issues/9573.

PR #9562, opened 2026-01-02
Fix cuda-python CI & the test status retrieval in Windows CI

leofang:


leofang:
The cuda-python CI is fixed!

.pfnci/windows/_flexci.ps1
```diff
@@ -18,23 +16,7 @@ function ActivatePython($version) {
...
-        $Env:CUDA_PATH = $Env:CUDA_PATH_V11_5
-    } elseif ($version -eq "11.6") {
-        $Env:CUDA_PATH = $Env:CUDA_PATH_V11_6
-    } elseif ($version -eq "11.7") {
-        $Env:CUDA_PATH = $Env:CUDA_PATH_V11_7
-    } elseif ($version -eq "11.8") {
-        $Env:CUDA_PATH = $Env:CUDA_PATH_V11_8
-    } elseif ($version -eq "11.x") {
-        $Env:CUDA_PATH = $Env:CUDA_PATH_V11_8
-    } elseif ($version -eq "12.0") {
```
seberg, line 37: Nice cleanup!

.pfnci/config.pbtxt
```diff
@@ -398,7 +398,7 @@ configs {
   value {
     requirement {
       cpu: 8
-      memory: 50
+      memory: 60
```
seberg, line 401: Was this the actual fix?

seberg (approved):
LGTM, all look like nice cleanups assuming things pass (and it looked like they do).
I am a bit surprised about what the actual fix is :)?

.pfnci/config.pbtxt
leofang: Yes. In the CI we hit OOM (somewhere, I suspect the CI runner monitors the high watermark). I noticed that all CUDA 12 + Linux CIs have bumped the memory limit to 60, so this also aligns with the existing treatment.

PR #9567, opened 2026-01-04
Add CUDA Stream Protocol support

### LINKED ISSUE #9401: RFC: Support CUDA Stream Protocol
For new Python projects, we encourage them to just use `cuda.core.<experimental>.Stream`. 

For existing Python projects, transitioning to `cuda.core` may or may not be desired. As a result, we encourage projects that already expose a CUDA stream to Python to follow the CUDA Stream protocol:
https://nvidia.github.io/cuda-python/cuda-core/latest/interoperability.html#cuda-stream-protocol
and add a `__cuda_stream__` method to the stream class, so as to improve interoperability without introducing extra `ExternalStream`-like types.

For CuPy, I suggest we further deprecate `ExternalStream` in favor of the protocol.

leofang:
Close #9401.


kmaehashi:
I'm completely in favor of bring CUDA stream protocol support to CuPy v14 but I'm wondering if it's better to defer the deprecation of `ExternalStream` to CuPy v15 or later, considering that not all libraries are supporting the protocol yet, and there might be a use-case that require importing stream via pointers from foreign libraries (e.g. `cv::cuda::Stream`).

leofang:
> but I'm wondering if it's better to defer the deprecation of `ExternalStream` to CuPy v15 or later

Yeah I've been pondering about this too. @kmaehashi my current thinking is 
- we raise `DeprecationWarning` in v14 (status quo in this PR)
- we raise `VisibleDeprecationWarning` in v15
- we remove it by v16

WDYT?

leofang:
FWIW this is the current adoption tracker: https://github.com/NVIDIA/cuda-python/issues/1000

cupy/_core/core.pyx
```diff
@@ -406,7 +406,10 @@ cdef class _ndarray_base:
             # Establish no stream order for now (for `stream=None` do it later)
             stream = None
         elif stream != curr_stream_ptr:
-            stream = stream_mod.ExternalStream(stream)
+            # Suppress deprecation warning for internal use of ExternalStream
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", DeprecationWarning)
+                stream = stream_mod.ExternalStream(stream)
```
seberg: I think we could just use a `_BaseStream()` also, since this is internal to avoid the warning context.

cupy/cuda/stream.pyx
```diff
@@ -495,20 +512,101 @@ class Stream(_BaseStream):
...
+        if self._foreign_stream_ref is not None:
+            self._foreign_stream_ref = None
+            return
         if self.ptr not in (0, runtime.streamLegacy, runtime.streamPerThread):
             runtime.streamDestroy(self.ptr)
         # Note that we can not release memory pool of the stream held in CPU
         # because the memory would still be used in kernels executed in GPU.
 
+    @classmethod
+    def from_external(cls, obj):
```
seberg, line 529: Would be cool to support this in `__init__` so that `Stream(external_object)` works directly.  But, I suspect we might want to transition to kwargs only for that to look nicer.

So something for the future, I think. Even in such a future, a `from_external()` method makes sense.

cupy/cuda/stream.pyx
```diff
@@ -495,20 +512,101 @@ class Stream(_BaseStream):
...
+
+        Examples:
+            >>> # Assuming torch_stream is a PyTorch CUDA stream
+            >>> cupy_stream = cupy.cuda.Stream.from_external(torch_stream)
+
+        """
+        try:
+            version, stream_ptr = obj.__cuda_stream__()
+        except AttributeError as e:
+            raise AttributeError(
```
seberg: I would suggest to use a `TypeError` here as well.  From the perspective of this function, the input type is incorrect (this is indicated by the `AttributeError`).

(Also change in docs, which means we can simplify it to `TypeError: If the object does not implement the CUDA stream protocol.)

cupy/cuda/stream.pyx
```diff
@@ -495,20 +512,101 @@ class Stream(_BaseStream):
...
+                method.
+
+        Returns:
+            Stream: A CuPy Stream wrapping the external stream.
+
+        Raises:
+            AttributeError: If the object does not implement
+                ``__cuda_stream__``.
+            TypeError: If ``__cuda_stream__`` does not return a valid
+                2-tuple.
```
seberg: (NIT: but if you like, could also be `ValueError` for unsupported protocol version.)

cupy/cuda/stream.pyx
```diff
@@ -495,20 +512,101 @@ class Stream(_BaseStream):
...
+
 
 class ExternalStream(_BaseStream):
 
     """CUDA stream not managed by CuPy.
 
+    .. deprecated:: 14.0
+        :class:`~cupy.cuda.ExternalStream` is deprecated. Use
+        :meth:`cupy.cuda.Stream.from_external` instead to interoperate with
+        external streams that implement the CUDA stream protocol.
```
seberg, line 607: I might consider honestly punting on this one for a bit. The reason is just that downstream probably can't adopt the new protocol yet, which makes it hard to avoid the `DeprecationWarning`.

(But no strong opinion, `DeprecationWarnings` are not very intrusive.)

docs/source/user_guide/interoperability.rst
```diff
@@ -394,12 +411,21 @@ CuPy can validate that for you if you pass ``device_id`` argument when creating
...
+   version, stream_ptr = s.__cuda_stream__()
+   print(version, stream_ptr)  # => 0 93997451352336
+   
+   # Or directly access the pointer attribute
+   print(s.ptr)  # => 93997451352336
+
+The CUDA stream will be destroyed when the :class:`~cupy.cuda.Stream`
+(``s``) gets destructed.
+You must keep the :class:`~cupy.cuda.Stream` instance alive while the
+pointer is in use by other libraries.
```
seberg, line 431: Not if they import it to the other one via `__cuda_stream__()`, but I guess writing an example for this is hard?
(I.e. `torch` doesn't have a `from_external()` type API yet.)

tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -154,12 +164,15 @@ def test_per_device_failure(self):
             with pytest.raises(RuntimeError):
                 stream0.use()
 
-    def test_mix_use_context(self):
+    @pytest.mark.parametrize('stream_name', ['null', 'ptds'])
+    def test_mix_use_context(self, stream_name):
+        stream = self._get_stream(stream_name)
```
seberg, line 169: Maybe your approach is nicer (to just inline this, IIRC, I created a decorator for it instead.)

tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -279,3 +297,175 @@ def test_launch_host_func(self):
...
+        with pytest.raises(TypeError, match=r'must return \(int, int\)'):
+            cuda.Stream.from_external(BadStream4())
+
+    def test_from_external_invalid_version(self):
+        # Test unsupported protocol version
+        class BadStream5:
+            def __cuda_stream__(self):
+                return (1, 12345)  # Version 1 is not supported
+
+        with pytest.raises(ValueError, match='unsupported version'):
```
seberg: NIT: Would be nice to parametrize the bad `__cuda_stream__` returns, will just have to parametrize the error type and message as well.

tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -279,3 +297,175 @@ def test_launch_host_func(self):
...
+            cuda.Stream.from_external(BadStream5())
+
+    def test_from_external_keeps_stream_alive(self):
```
seberg, line 390: This is a nice test, although I admit, since we already check that `_foreign_stream_ref`, I am not sure it adds much.
(But it exists now, so no big opinion.)

cupy/cuda/stream.pyx
leofang: I did this to `cuda.core.Stream` but I am not super happy about it. The constructor logic becomes a bit messy. Deferring this decision to future was indeed the idea.

cupy/cuda/stream.pyx
leofang: xref: https://github.com/cupy/cupy/pull/9567#issuecomment-3745311597, @kmaehashi would be nice to get your input 🙂 

I'll address the remaining comments once we reach a conclusion here

kmaehashi: Raising DeprecationWarnings sounds good as it is invisible to most users. Let's think about when to move further depending on community adoption and/or user feedback.

cupy/_core/core.pyx
leofang: done in commit 6917b69e6

cupy/cuda/stream.pyx
leofang: Good idea, done in commit 6917b69e6

cupy/cuda/stream.pyx
leofang: done in commit 6917b69e6

tests/cupy_tests/cuda_tests/test_stream.py
leofang: done in commit 6917b69e6

kmaehashi:

/test skip

PR #9570, opened 2026-01-05
Avoid allocating an intermediate array in `cupy.random.choice`

leofang:
Turns out we can't use `cuda::shuffle_iterator` or `thrust::shuffle_iterator` for a few reasons:
1. They are not available in CCCL 3.1.2 that we just updated to
2. Using them as-is is challenging because
   - We need pre-compiled kernels + C/Cython bindings (also noted in the inline comment)
   - We need to figure out a way to influence its RNG there
3. Both have serious bugs that just got fixed upstream (I propagated the fixes to here)
   - uniform randomness
   - cipher bit initialization

The solution here is arguably very simple: We initialize the state (a POD struct) on host and pass it to the device as a kernel argument. Then, the bijection device function can be invoked straightforwardly.

### DIFF
--- a/cupy/random/_generator_api.pyx
+++ b/cupy/random/_generator_api.pyx
@@ -1129,7 +1129,7 @@ cdef class FeistelBijection:
         self.params = _FeistelBijectionParam()
         params_struct = &self.params.struct
 
-        # Round up to at least 4 bits, then to next power of 2
+        # Round up to at least 8 bits, then to next power of 2
         # Note: This is a bug fix to cuda::shuffle_iterator (NVIDIA/cccl#7073).
         total_bits = get_cipher_bits(num_elements)
         self.arr_size = num_elements


cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+        """Initialize Feistel bijection with pre-generated random keys.
+
+        Args:
+            num_elements: Size of the domain (will be rounded up to power of 2)
+            keys_array: Array of 24 uint32 random numbers to use as keys
+        """
+        cdef uint64_t total_bits
+
+        # Round up to at least 4 bits, then to next power of 2
+        total_bits = get_cipher_bits(num_elements)
```
leofang: Note: This is a bug fix to `cuda::shuffle_iterator` (https://github.com/NVIDIA/cccl/issues/7073).

leofang:
Discovered a bug locally. Let's see if Copilot can spot it 🙂

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+                
+                // 24 rounds of Feistel network
+                for (uint32_t i = 0; i < 24; i++) {
+                    constexpr uint64_t __m0  = 0xD2B74407B1CE6E93;
+                    const uint64_t __product = __m0 * high;
+                    const uint32_t __high    = static_cast<uint32_t>(__product >> 32);
+                    uint32_t __low           = static_cast<uint32_t>(__product);
+                    __low                    = (__low << (params.right_side_bits - params.left_side_bits)) | low >> params.left_side_bits;
+                    high                     = ((__high ^ params.keys[i]) ^ low) & params.left_side_mask;
+                    low                      = __low & params.right_side_mask;
```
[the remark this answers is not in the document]
leofang: This is a literal copy of the libcudacxx code. Let's keep it as-is for easier cross-reference.

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+        # Half bits rounded down
+        self.param.left_side_bits = total_bits // 2
+        self.param.left_side_mask = (1 << self.param.left_side_bits) - 1
+
+        # Half the bits rounded up
+        self.param.right_side_bits = total_bits - self.param.left_side_bits
+        self.param.right_side_mask = (1 << self.param.right_side_bits) - 1
+
+        # Copy keys from the input array
+        assert len(keys_array) == 24
```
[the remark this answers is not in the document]
leofang: @copilot There is no such a thing as debug vs optimized builds in Python.

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+        # TODO(leofang): use NumPy C API
+        params = numpy.empty(1, dtype=_feistel_bijection_dtype)
+        assert (
+            _feistel_bijection_dtype.itemsize == sizeof(_FeistelBijection))
+        memcpy(<void*><intptr_t>(params.ctypes.data),
+               &(self.param),
+               sizeof(_FeistelBijection))
+        return params
+
+    cdef get_kernel(self):
```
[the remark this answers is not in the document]
leofang: cdef instead of property is used here to avoid Python overhead.

cupy/random/_generator.py
```diff
@@ -1136,8 +1137,13 @@ def choice(self, a, size=None, replace=True, p=None):
...
             if isinstance(a, int):
-                indices = cupy.arange(a, dtype='l')
+                # Use memory-efficient bijection approach
+                keys = numpy.random.randint(
+                    0, _UINT32_MAX + 1, size=24, dtype=cupy.uint32)
+                bijection = FeistelBijection(a_size, keys)
+                indices = bijection(size, a_size)
             else:
                 indices = a.copy()
             self.shuffle(indices)
```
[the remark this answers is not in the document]
leofang: Yes, this is the bug I found locally!

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+                if (tid >= arr_size)
+                    return;
+                
+                // Apply Feistel bijection to tid
+                uint64_t val = static_cast<uint64_t>(tid);
+                uint32_t high = (uint32_t)(val >> params.right_side_bits);
+                uint32_t low = (uint32_t)(val & params.right_side_mask);
+                
+                // 24 rounds of Feistel network
+                for (uint32_t i = 0; i < 24; i++) {
```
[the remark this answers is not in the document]
leofang: The kernel body is largely kept intact for easier reference to the original libcudacxx code.

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+        """Get bijection parameters as a structured array for device code.
+
+        Returns:
+            numpy.ndarray: Structured array containing all bijection parameters
+        """
+        # Create and populate the params array
+        # TODO(leofang): use NumPy C API
+        params = numpy.empty(1, dtype=_feistel_bijection_dtype)
+        assert (
+            _feistel_bijection_dtype.itemsize == sizeof(_FeistelBijection))
```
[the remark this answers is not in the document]
leofang: This is Python assert (in Cython) not C assert.

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+        Returns:
+            cupy.RawKernel: The compiled CUDA kernel for the bijection
+        """
+        global _feistel_bijection_with_cutoff_kernel
+        if _feistel_bijection_with_cutoff_kernel is None:
+            # FIXME(leofang): currently only supports a_size <= 2**32
+            # FIXME(leofang): move kernel to separate file
+            # Feistel bijection kernel for choice without replacement
+            _feistel_bijection_with_cutoff_kernel = cupy.RawKernel(
+                r'''  // # noqa: E501
```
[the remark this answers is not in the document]
leofang: @copilot This is the only way to silence cython-lint. I don't want the C++ code to be linted.

cupy/random/_generator.py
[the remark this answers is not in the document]
leofang: fixed

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+                    const uint32_t __high    = static_cast<uint32_t>(__product >> 32);
+                    uint32_t __low           = static_cast<uint32_t>(__product);
+                    __low                    = (__low << (params.right_side_bits - params.left_side_bits)) | low >> params.left_side_bits;
+                    high                     = ((__high ^ params.keys[i]) ^ low) & params.left_side_mask;
+                    low                      = __low & params.right_side_mask;
+                }
+                
+                // Combine left and right sides
+                long long idx = (long long)((static_cast<uint64_t>(high) << params.right_side_bits) | static_cast<uint64_t>(low));
+                if (tid < cutoff_size) {
```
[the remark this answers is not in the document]
leofang: fixed

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+                    out[tid] = idx; 
+                }
```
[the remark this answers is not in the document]
leofang: fixed

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1062,168 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+
+        # Launch kernel
+        block_size = 256
+        grid_size = (a_size + block_size - 1) // block_size
+        kernel(
+            (grid_size,), (block_size,),
+            (indices, params, size, a_size)
+        )
+
+        return indices
```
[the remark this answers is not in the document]
leofang: This is a private class. We do the checks on the call site.

cupy/random/_generator.py
```diff
@@ -1136,12 +1137,18 @@ def choice(self, a, size=None, replace=True, p=None):
                 raise ValueError(
                     'Cannot take a larger sample than population when '
                     '\'replace=False\'')
+
             if isinstance(a, int):
-                indices = cupy.arange(a, dtype='l')
+                # Use memory-efficient bijection approach
+                keys = numpy.random.randint(
+                    0, _UINT32_MAX + 1, size=24, dtype=numpy.uint32)
```
leofang: Note: This honors `numpy.random.seed()`. But, it might be better to use (and update) `self._rk_seed`?

leofang:
Currently blocked on the uniformness discussion (https://github.com/NVIDIA/cccl/pull/7062#issuecomment-3715733827).

cupy/random/_generator_api.pyx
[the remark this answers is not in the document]
leofang: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/cupy/cupy/pull/9570#discussion_r2662926609)

cupy/random/_generator.py
leofang: addressed in commit 30840f846

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1071,191 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+        Warning:
+            The returned array references self's internal memory without
+            holding a reference. It must be used immediately while self
+            is alive. Do not store or return it to Python code.
+        """
+        # Create and populate the params array
+        assert (
+            _feistel_bijection_dtype.itemsize == sizeof(_FeistelBijection))
+        cdef cnp.npy_intp[1] dims = [1]
+        Py_INCREF(_feistel_bijection_dtype)
```
leofang: @seberg this `Py_INCREF` is needed to avoid segfault, but it's a bit nerve wrecking...

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1071,191 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+                uint64_t __R_bits_;
+                uint64_t __L_bits_;
+                uint64_t __R_mask_;
+                uint64_t __L_mask_;
+                uint32_t __keys_[{_FEISTEL_NUM_ROUNDS}];
+            }};
+            
+            extern "C" __global__ void feistel_bijection_choice(
+                long long* out,
+                const FeistelParams params,
```
leofang, line 1172: Note: Having to pass this struct (by value) is the only reason that we use `RawKernel` instead of `ElementwiseKernel` here (https://github.com/cupy/cupy/issues/9588).

cupy/random/_generator.py
```diff
@@ -1126,7 +1127,11 @@ def choice(self, a, size=None, replace=True, p=None):
             raise NotImplementedError(
                 'choice() without specifying size is not supported yet')
         shape = size
-        size = numpy.prod(shape)
+        try:
+            size = _core.internal.prod(shape)
```
leofang, line 1131: Perf optimization (speed up by a few us)

cupy/random/_generator_api.pyx
seberg: Yeah, but correct, terrible API from a time long past, when stealing refs was common...
```suggestion
        Py_INCREF(_feistel_bijection_dtype)  # NewFromDescr steals dtype.
```

leofang: Once reviewed/approved and CI turning green, I will apply the diff and force-merge to save CI resources. It seems we have a long queue for Windows runners for some reason.

leofang:
cc @asi1024 for vis (in case you have time to take a look 🙂)

tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -918,6 +918,235 @@ def test_bound(self):
...
+    def test_multidim_shape_uniqueness(self):
+        """Test multi-dimensional outputs maintain uniqueness."""
+        val = self.generate(a=self.a, size=self.size, replace=False).get()
+
+        # Check shape
+        assert val.shape == self.size
+
+        # Check uniqueness across all dimensions
+        val_flat = val.flatten()
+        assert numpy.unique(val_flat).size == val_flat.size
```
seberg: This one also, might as well just make a comment on the parametrization that this tests multi dimensional, the first test can do this.

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1071,191 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+            '''  # noqa: E501
+            , 'feistel_bijection_choice')
+        return _feistel_bijection_with_cutoff_kernel
+
+    def __call__(self, size):
+        params = self.get_params()
+        kernel = self.get_kernel()
+
+        # Allocate output
+        # TODO(leofang): check if this equals to dtype='l' on Windows
```
seberg, line 1217: Hmmmmpf, this should not be the case. OTOH, the new generator always returns 64bit...

I can see an argument in diverging from NumPy here and just using 64bit integers always.

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1071,191 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+        """Initialize Feistel bijection with pre-generated random keys.
+
+        Args:
+            num_elements: Size of the domain (will be rounded up to power of 2)
+            keys_array: Array of :attr:`num_rounds` uint32 random numbers to
+                use as keys
+        """
+        cdef uint64_t total_bits
+
+        # Round up to at least 4 bits, then to next power of 2
```
RAMitchell: Now at least 8 bits.

cupy/random/_generator_api.pyx
seberg: I took the opinionated action and pushed (sorry, with a rebase!) two changes:
1. Replace this with a `CPointer` pattern to avoid the array. I think this is also a subtle bug fix maybe, because the NumPy array didn't own the memory and the base was not set.  (But mainly, I think I like this pattern.)
2. Very opinionated: I made the tests much more aggressively parametrized.  Admittedly, that does lose the structure a little bit.

It looks good, though (and Rory is having a look).  The `l` choice is a bit of a downer :/, should we add a cast on windows for now.

cupy/random/_generator_api.pyx
```diff
@@ -1061,3 +1071,191 @@ cdef void _launch_dist(bit_generator, func, out, args) except*:
...
+    ('__R_mask_', numpy.uint64),
+    ('__L_mask_', numpy.uint64),
+    ('__keys_', numpy.uint32, (_FEISTEL_NUM_ROUNDS,))
+], align=True)
+
+
+cdef object _feistel_bijection_with_cutoff_kernel = None
+
+
+cdef inline uint64_t get_cipher_bits(uint64_t m) nogil noexcept:
```
RAMitchell, line 1090: I think we should write tests that the bijection size is as exected, including edge cases e.g. m=0.

RAMitchell (approved):
Feistel code looks correct.

seberg (approved):
LGTM as it is, the only real thing is that windows long issue...

cupy/random/_generator_api.pyx
leofang: > Replace this with a `CPointer` pattern to avoid the array. 

Ah right, I forgot this is internal code and we get to use internal utilities 👍 

> I think this is also a subtle bug fix maybe, because the NumPy array didn't own the memory and the base was not set.  

Yes. Copilot warned both locally and in one of the threads above IIRC. But it was safe because we guarantee by construction that the underlying C struct outlives the temporary NumPy array.

> (But mainly, I think I like this pattern.)

Agreed.

cupy/random/_generator_api.pyx
leofang:
```suggestion
        # Round up to at least 8 bits, then to next power of 2
```

leofang:
@seberg looks like I invalidated your approval so we need to approve again lol

seberg:
@leofang I'll mark it as no-compat. It is a smaller change (the 32->64bit is windows only). It is a random stream change, though.  So it could be added to the final release notes.

cupy/random/_generator_api.pyx
leofang: Ahh ok I see what you meant:
- With the old RNG, on Windows NumPy returns int32
- With the new RNG, on Windows NumPy returns int64

It seems CuPy shares the implementation for both old/new RNG. I think it is the best to align and mark this PR as no-compat indeed.

PR #9579, opened 2026-01-08
TYP: Fix small typing issues found in new pre-commit env

seberg:
Some versions changes seems to have flushed these out locally. The `like`/`full` functions explicitly support `None` of course while the `astype` helper function does indeed not support it.

### DIFF
--- a/cupy/_creation/basic.py
+++ b/cupy/_creation/basic.py
@@ -68,7 +68,7 @@ def _new_like_order_and_strides(
 
 def empty_like(
         prototype: NDArray[Any],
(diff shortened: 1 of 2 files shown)


leofang (approved):

leofang:
I think the pre-checks should suffice and no need to run the GPU tests for typing changes?

seberg:
> I think the pre-checks should suffice and no need to run the GPU tests for typing changes?

Yeah, agreed, this only affects mypy, let's see.

/test skip

PR #9582, opened 2026-01-10
ENH: make `broadcast_arrays`, `meshgrid` return a tuple not list

ev-br:
NumPy changed the return type in NumPy 2.0. 
Array API follows suit in the upcoming 2025.12 revision.

Would CuPy prefer to match numpy or keep cupy-specific backwards compatibility?

### DIFF
--- a/tests/cupy_tests/creation_tests/test_ranges.py
+++ b/tests/cupy_tests/creation_tests/test_ranges.py
@@ -324,8 +324,8 @@ class TestMeshgrid(unittest.TestCase):
         x = xp.arange(2).astype(dtype)
         result = xp.meshgrid(x, indexing=self.indexing, sparse=self.sparse,
                              copy=self.copy)
-        if self.copy and xp == numpy:
-            # XXX cf numpy#30641
+        if xp == cupy:
+            # XXX cf numpy#30641, should be fixed in NumPy 2.5
             assert isinstance(result, tuple)
         return result
 


seberg:
I think we should match NumPy 2.x for the upcoming release. Might be nice to have a test.

ev-br:
One other function which saw the numpy 2.0 return change from list to a tuple is `meshgrid`. Want me to change it, too, while we're at it --- this PR, a separate PR, or not at all?

leofang:
Same PR is fine. Saves a bit of resources :)

ev-br:
Done then.

seberg:
@ev-br the test failures are unfortunately real. We missed the earlier branch in meshgrid that still returns a list.

ev-br:
Yes, sorry about that. Fixed the empty tuple return, and what's left is this beauty:

```
In [1]: import numpy as np

In [2]: x, indexing, sparse, copy = np.array([0, 1]), 'xy', True, False

In [5]: np.meshgrid(x, indexing=indexing, sparse=sparse, copy=copy)
Out[5]: [array([0, 1])]

In [6]: np.__version__
Out[6]: '2.3.5'
```

which leads to

```
====================================== FAILURES =======================================
____ TestMeshgrid_param_1_{copy=False, indexing='xy', sparse=True}.test_meshgrid1 _____

self = <<cupy_tests.creation_tests.test_ranges.TestMeshgrid_param_1_{copy=False, indexing='xy', sparse=True} testMethod=test_meshgrid1>  parameter: {'copy': False, 'indexing': 'xy', 'sparse': True}>
xp = <module 'numpy' from '/home/ev-br/.conda/envs/cupyx/lib/python3.11/site-packages/numpy/__init__.py'>
dtype = <class 'numpy.float64'>

    @testing.for_all_dtypes()
    def test_meshgrid1(self, xp, dtype):
        x = xp.arange(2).astype(dtype)
        result = xp.meshgrid(x, indexing=self.indexing, sparse=self.sparse,
                             copy=self.copy)
>       assert isinstance(result, tuple)
E       AssertionError: Only numpy raises error
E       
E       Traceback (most recent call last):
E         File "/home/ev-br/repos/cupy_ng/cupy/testing/_loops.py", line 45, in _call_func
E           result = impl(*args, **kw)
E                    ^^^^^^^^^^^^^^^^^
E         File "/home/ev-br/repos/cupy_ng/tests/cupy_tests/creation_tests/test_ranges.py", line 327, in test_meshgrid1
E           assert isinstance(result, tuple)
E       AssertionError: assert False
E        +  where False = isinstance([array([0., 1.])], tuple)

tests/cupy_tests/creation_tests/test_ranges.py:327: AssertionError
-------------------------------- Captured stdout call ---------------------------------
dtype is <class 'numpy.float64'>
=============================== short test summary info ===============================
FAILED tests/cupy_tests/creation_tests/test_ranges.py::TestMeshgrid_param_1_{copy=False, indexing='xy', sparse=True}::test_meshgrid1 - AssertionError: Only numpy raises error
```

seberg (approved):

seberg:
Sorry, should have checked what you wrote: Is the test still failing now due to NumPy also missing a path?
Seems OK to adapt the test for that (maybe just test the return type separately?).

ev-br:
I think it's https://github.com/numpy/numpy/issues/30641

ev-br:
Added a workaround for the numpy issue.

tests/cupy_tests/creation_tests/test_ranges.py
```diff
@@ -316,14 +316,18 @@ class TestMeshgrid(unittest.TestCase):
...
     @testing.for_all_dtypes()
     @testing.numpy_cupy_array_equal()
     def test_meshgrid1(self, xp, dtype):
         x = xp.arange(2).astype(dtype)
-        return xp.meshgrid(x, indexing=self.indexing, sparse=self.sparse,
-                           copy=self.copy)
+        result = xp.meshgrid(x, indexing=self.indexing, sparse=self.sparse,
+                             copy=self.copy)
+        if self.copy and xp == numpy:
+            # XXX cf numpy#30641
```
seberg:
```suggestion
        if xp == cupy:
            # XXX cf numpy#30641, should be fixed in NumPy 2.5
```
The if may be right, but disables the test :).

seberg (approved):
Thanks Evgeni, especially for the follow-ups and opening the NumPy issue.

PR #9584, opened 2026-01-12
DEV: Make `CUPY_TEST_GPU_LIMIT` more reliable.

seberg:
Not that `CUPY_TEST_GPU_LIMIT` is a game-changer probably, but it didn't actually work reliably (I assume it used to work). I am not sure why exactly, it may be that the fixture is too late because it is run after test collection (which then may e.g. create cupy arrays).

This pattern using the environment variables was probably not available when it was introduced initially, but works now.

(N.B. at least for high-memory tests using two runners on the same GPU probably doesn't work, xdist groupings may be a way to solve this.)

---

Re-opened this, since I realized that the old pattern doesn't work for the core tests. (Fun fact: it's not even necessarily faster to use many GPUs at least.)


leofang:
> (Fun fact: it's not even necessarily faster to use many GPUs at least.)

Yeah I don't think using multi GPUs to run tests is about accelerating it. It's to cover multi-GPU functionalities in CuPy.

leofang:
Would be nice if @kmaehashi could take a look 🙂

tests/conftest.py
```diff
@@ -37,48 +35,39 @@ def pytest_configure(config):
...
-        print(f'CUDA_VISIBLE_DEVICES={devices}')
+            # We'll edit the environment variable `CUDA_VISIBLE_DEVICES` for
+            # each worker which needs to happen before the CUDA init.
+            n_gpu = int(n_gpu)
+
+            devices = os.environ.get('CUDA_VISIBLE_DEVICES')
+            if devices is None:
+                devices = [str(k) for k in range(n_gpu)]
+            else:
+                devices = devices.split(',')[:n_gpu]
```
leofang, line 57: Q: If `CUDA_VISIBLE_DEVICES` is set, we may want to compare it with `PYTEST_XDIST_WORKER_COUNT`?
https://github.com/NVIDIA/cuQuantum/blob/4e097de09c9025f3a10a2f574d9fcd95167c1e6d/benchmarks/tests/nv_quantum_benchmarks_tests/test_run.py#L32-L39

seberg: Seems unnecassary even in that code, total workers is just to branch, but the second branch does the same thing anyway as the % would be a no-op?

leofang: Ah. I didn't quite get what the original code was doing (and this PR does not change the device assignment logic, just exercises it in a different way). I see now that it preserves the number of visible devices and just rotates the list

In the linked code, the intention was to assign each worker a GPU. It serves for a different purpose from what we need here.

seberg:
Maybe gentle ping, I think it makes to just to do this as it is a bit simpler and works as intended.
(I am not actually sure that distributing tests across GPUs is all that useful in practice.)

I think I would like to also follow-up e.g. with automatically setting the `GPU_TEST_LIMIT` for convenience.

tests/conftest.py
```diff
@@ -37,48 +35,39 @@ def pytest_configure(config):
...
+                      'setting CUPY_TEST_GPU_LIMIT environment variable.\n')
         else:
-            devices = devices.split(',')[:n_gpu]
-        devices = collections.deque(devices)
-        devices.rotate(w)
-        devices = ','.join(devices)
-        os.environ['CUDA_VISIBLE_DEVICES'] = devices
-        # With PyTest's default, the print will be shown as
-        # "--- Captured stdout setup ---" on failure.
-        print(f'CUDA_VISIBLE_DEVICES={devices}')
```
leofang, line 81: nit: this print seems nice to keep (for debugging/logging purposes)?

seberg: There is a new print (which unlike this one should actually be shown more reliably :)).

PR #9585, opened 2026-01-12
TST: Reorganize cublas tests for thread-safety

seberg:
This modifies the cublas tests to ensure they are thread safe if run in parallel (with `pytest-run-parallel`)

---

Changes are a bit more than strictly needed.  The important part here is just that e.g. mutating `self.mode` is not allowed. But it seemed more obvious to stop relying on `self` attributes.
(Unfortunately, part of the diff doesn't show quite as straight forward as it is.)

### DIFF
--- a/tests/cupy_tests/test_cublas.py
+++ b/tests/cupy_tests/test_cublas.py
@@ -186,55 +186,49 @@ class TestLevel1Functions:
         cupy.testing.assert_allclose(x, ref, rtol=self.tol, atol=self.tol)
 
 
-@testing.parameterize(*testing.product({
-    'dtype': ['float32', 'float64', 'complex64', 'complex128'],
-    'shape': [(10, 9), (9, 10)],
-    'trans': ['N', 'T', 'H'],
-    'order': ['C', 'F'],
-    'mode': [None, numpy, cupy],
-}))
+@pytest.mark.parametrize('dtype',
+                         ['float32', 'float64', 'complex64', 'complex128'])
+@pytest.mark.parametrize('shape', [(10, 9), (9, 10)])
+@pytest.mark.parametrize('trans', ['N', 'T', 'H'])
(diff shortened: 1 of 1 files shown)


tests/cupy_tests/test_cublas.py
```diff
@@ -263,78 +257,79 @@ def test_sbmv(self, dtype):
...
+    def test_ger(self, dtype, shape, order, mode):
+        alpha, tol, x, y, a = self.get_data(dtype, shape, order)
 
-@testing.parameterize(*testing.product({
-    'nk': [(5, 9), (9, 5)],
-    'transa': ['N', 'T'],
-    'ordera': ['F', 'C'],
-    'orderc': ['F', 'C'],
-    'lower': [0, 1],
-    'mode': [None, numpy, cupy]
```
leofang, line 336: `mode` is gone for `TestSyrk`?

leofang (commented):
Thanks, @seberg, looks fine to me! So is the lesson here that `@testing.parameterize` should be deprecated/removed in favor of `@pytest.mark.parametrize` in the era of free threading?

tests/cupy_tests/test_cublas.py
seberg: didn't remember deleting this one, but yeah: It was an unused parametrization.

seberg:
> So is the lesson here that @testing.parameterize should be deprecated/removed in favor of @pytest.mark.parametrize in the era of free threading?

I guess not strictly, but I am not sure I have much love for the pattern without `unittest.TestCase`.
In some paths the way it inserts a parametrization that mutes `self` was brittle and here, `self.mode` is replaced which doesn't work in free-threading.
(Of course a `mode += 1` would still not work with normal parametrization.)

So it _is_ a bit a matter of taste, in that I see small downsides but no big advantage besides the historic `unitest.TestCase` usage.

PR #9551, opened 2025-12-24
Fix CUDA 13.1 CI

leofang:

PR #9592, opened 2026-01-13
CI: Revert to use NCCL 2.28 in CUDA 13.1 CI

kmaehashi:
Follows up #9551.


leofang:
ah, they released new packages 👍 @kmaehashi need to resolve conflicts

leofang (approved):

Issue #9569, opened 2026-01-05
Move from unittest to pytest


leofang:
This is not an RFC, but an umbrella issue tracking the status quo. We've been slowly moving tests from unittest to pytest over the years, but I don't think there is a tracking issue yet.

New tests should always be written using pytest.

When old tests are being updated, if it is trivial to change we should also migrate them to pytest.

We'll close this issue once unittest is no longer imported in the CuPy test suite.

seberg:
This would be great.  I also am not sure we should keep around the custom `testing.parametrize` really.  It is convenient, but in the end, I don't think repeating the parameters in every test function is all that bad.
(Which is the main difference to using `pytest.mark.parametrize` on the class.)

megha-darda:
Hi, I’m interested in contributing.  
If there’s a recommended file or folder to start with, I can begin there and open a small PR.

acosmicflamingo:
If no one is currently working on this, could I be assigned this ticket? I think I was able to figure out a way to migrate the test suite to using `pytest.mark.parametrize` in a way where it keeps lots of the helper functions intact. For example, in `test_array.py`, one can use pytest fixture's `autouse=True` in this fashion:
```diff
diff --git a/tests/cupy_tests/core_tests/fusion_tests/test_array.py b/tests/cupy_tests/core_tests/fusion_tests/test_array.py
index b73cad84f..1f94c7459 100644
--- a/tests/cupy_tests/core_tests/fusion_tests/test_array.py
+++ b/tests/cupy_tests/core_tests/fusion_tests/test_array.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import pytest
+
 import numpy
 
 from cupy import testing
@@ -8,6 +10,11 @@ from cupy_tests.core_tests.fusion_tests import fusion_utils
 
 class FusionArrayTestBase:
 
+    @pytest.fixture(autouse=True)
+    def set_values(self, left_value, right_value):
+        self.left_value = left_value
+        self.right_value = right_value
+
     def _get_argument(self, xp, dtype, seed, value_type):
         dtype = numpy.dtype(dtype)
         if value_type == 'array':
```

Then our tests can look like this:
```python
@pytest.mark.parametrize(
    "left_value,right_value",
    [
        ('array', 'array'),
        ('array', 'scalar'),
        ('array', 'primitive'),
        ('scalar', 'array'),
        ('primitive', 'array'),
    ]
)
[...]

PR #9596, opened 2026-01-13
TST: migrate tests from unittest to pytest

megha-darda:
This PR migrates several existing tests from `unittest`.

Updates tests include:
- elementwise tests in binary_tests
- packing tests in binary_tests
- array function tests in core_tests
- core tests in core_tests
- internal tests in core_tests

The changes are mechanical:
- remove `unittest.TestCase` and `import unittest`
- replace `self.assertRaises` with `pytest.raises`
- no intended behavior changes

Part of #9569.
I’m happy to continue with other test files if helpful.

### DIFF
--- a/tests/cupy_tests/core_tests/test_internal.py
+++ b/tests/cupy_tests/core_tests/test_internal.py
@@ -4,7 +4,6 @@ import numpy
 import pytest
 
 from cupy._core import internal
-from cupy import testing
 
 
 class TestProd:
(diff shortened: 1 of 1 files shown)


tests/cupy_tests/core_tests/test_internal.py
```diff
@@ -215,7 +213,7 @@ def test_invalid_stop_type(self):
     {'x': 2 ** 40 - 1, 'expect': 2 ** 40},
     {'x': 2 ** 40 + 1, 'expect': 2 ** 41},
 )
```
seberg, line 218: I would be tempted to also convert this parametrization to `pytest.mark.parametrize`, but we could also do that as separate PR.

seberg (approved):
LGTM to me already, might be nice to remove that parametrization while we are at it (since in this case it looks straight forward).

tests/cupy_tests/core_tests/test_internal.py
megha-darda: Thanks! I’ve updated so that `testing.parameterize` is replaced with `pytest.mark.parametrize` in `test_internal.py`.

If there are other test files you consider higher priority for this migration, please let me know. Otherwise I can continue incrementally in follow up PRs.

seberg (approved):

tests/cupy_tests/core_tests/test_internal.py
seberg: Thanks! Nothing in particular, I have a bit of work on tests in gh-9607, but that shouldn't be a big deal (even if there is a conflict, it'll be easy).

(One related thing I stumbled on was that `testing.for_contiguous_axes` required the user of `self` and classes. I don't think it is used in very many places, so it might be nice to use a pytest fixture now.)

PR #9600, opened 2026-01-13
ENH: do not unload modules/code that have been used

seberg:
This is a minimal solution to make graph capture mostly more robust in a threaded environment.

I have a maximal solution that ensures we will never compile+unload unnecessary.
This may seem unnecessary, but unfortunately, this solution is not sufficient when running with `compute-sanitizer` (it seems `compute-sanitizer + graph capture` has problems with unloading here, I am not sure why).

---

Opening this, in case we feel we can backport this sooner, I am not sure how important it is.
I had the maximal solution here and was holding off on creating a PR (for the other one): https://github.com/cupy/cupy/compare/main...seberg:cupy:compile-once

(The maximal solution isn't terribly difficult, but adds a custom `ExactlyOnceDict`, which is a bit awkward.)

### DIFF
--- a/tests/cupy_tests/core_tests/test_multithreading.py
+++ b/tests/cupy_tests/core_tests/test_multithreading.py
@@ -1,0 +1,104 @@
+from __future__ import annotations
+
+import concurrent.futures
+import threading
+
+import pytest
+
+import cupy
+
+
+pytestmark = pytest.mark.thread_unsafe(
+    reason="tests in this module are already explicitly multi-threaded"
+)
+
+
+def run_threaded(func, max_workers=8, pass_count=False,
+                 pass_barrier=False, outer_iterations=1,
+                 prepare_args=None):
+    """Runs a function many times in parallel
+
+    This function has been taken from NumPy:
+    https://github.com/numpy/numpy/blob/a90ef57574c501a780fe834123b20fcea1329f90/numpy/testing/_private/utils.py#L2807
+    """
+    for _ in range(outer_iterations):
+        with (concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
+              as tpe):
+            if prepare_args is None:
+                args = []
+            else:
+                args = prepare_args()
+            if pass_barrier:
+                barrier = threading.Barrier(max_workers)
+                args.append(barrier)
+            if pass_count:
+                all_args = [(func, i, *args) for i in range(max_workers)]
+            else:
+                all_args = [(func, *args) for i in range(max_workers)]
+            try:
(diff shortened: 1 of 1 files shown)


leofang:
I am not sure I follow, how does using setdefault prevents unloading?

seberg:
>  how does using setdefault prevents unloading?

It doesn't prevent the _unloading_ itself, thus this is not a 100% fix, and I would be happy to the linked diff.
However, it prevents unloading of a _used_ function. Without this the following can happen, which is obviously bad:
* Thread 1: compiles `add()` as `add1` and stores it
* Thread 2: compiles `add()` as `add2` and stores it (replacing `add1`)
* Thread 1: Captures `add1(x, y)`
* Thread 1: Unloads `add1`, captured graph is clearly bad.

The `setdefault` doesn't prevent the unloading. However, it prevents *Thread 2* from replacing `add1`. I.e. the function we actually end up using is guaranteed to not be unloaded again.

Now, for some reasons, unloading by itself (at least within the graph capture?) can lead to crashes in graph capture. So this approach doesn't fix that, but is trivial.

leofang:
Ah ok, if we unconditionally set the value we could purge the previously set value. The current fix looks fine to me. I don't have any strong opinion favoring this or the `ExactlyOnceDict` fix. I guess it'd be up to you and @cupy/code-owners 🙂

leofang:
Q: Is there a way we can construct a test for this guarantee? Seems a bit challenging to me... 🤔

seberg:
Maybe it's time to add some explicitly threaded tests. Graph capture tests notice when using pytest-run-parallel.
I think we can e.g. create a new ufunc and check if calling `_get_ufunc_kernel()` twice returns the identical object. If the test is threaded that should fail. (Or maybe by inspecting the cache.)

leofang:
Ah that sounds good to me! We do already have a few `threading`-based tests:
https://github.com/search?q=repo%3Acupy%2Fcupy%20threading%20path%3Atests%2F&type=code

seberg:
Added a `test_multithreaded.py` with the pattern from NumPy (outer iterations are needed, the tests seem surprisingly unflaky even on multi-threading.).

seberg:
> I don't have any strong opinion favoring this or the ExactlyOnceDict fix.

I was considering to do this quickly and put up the other PR then for discussion and don't worry too much if it stalls a bit.
The issue I have seen is just with graph-capture within `compute-sanitizer`, so I suspect it isn't super urgent, while this helps and is very simple.

leofang (approved):

PR #9606, opened 2026-01-14
Fix compute capability list and remove CUDA 11 related code

kmaehashi:
* Follows-up #9200. Fix max compute capability for CUDA 12.9+. It should be `121` instead of `120`.
* Follows-up #9406. Correctly detect and fail when build is attempted under CUDA 11.
* Clean up code related to CUDA 11.

### DIFF
--- a/cupy/cuda/compiler.py
+++ b/cupy/cuda/compiler.py
@@ -169,22 +169,15 @@ _tegra_archs = ('32', '53', '62', '72', '87')
 @_util.memoize()
 def _get_max_compute_capability():
     major, minor = _get_nvrtc_version()
-    if major < 11:
-        # CUDA 10.2
-        nvrtc_max_compute_capability = '75'
-    elif major == 11 and minor == 0:
-        # CUDA 11.0
-        nvrtc_max_compute_capability = '80'
-    elif major == 11 and minor < 8:
-        # CUDA 11.1 - 11.7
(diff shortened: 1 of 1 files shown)


install/cupy_builder/_features.py
```diff
@@ -474,9 +475,9 @@ def configure(self, compiler: Any, settings: Any) -> bool:
 
         self._version = int(out)
 
-        if self._version < self.minimum_cuda_version:
+        if self._version < self._minimum_cuda_version:
             utils.print_warning(
-                'CUDA version is too old: %d' % self._version,
-                'CUDA 11.2 or newer is required')
+                f'CUDA version is too old: {self._version}',
+                f'CUDA {self._minimum_cuda_version_str} or newer is required')
```
kmaehashi, line 481: Tested locally with CUDA 11.8:

```
      -------- Configuring Module: cuda --------
      **************************************************
      *** WARNING: CUDA version is too old: 11080
      *** WARNING: CUDA 12.0 or newer is required
      **************************************************
```

seberg (approved):
Awesome cleanup, thanks!

seberg:
Hmmm, timeouts, not sure this is the best course of action, but:

/test mini

leofang:
> Hmmm, timeouts, not sure this is the best course of action

For some reason that is not obvious to me, this PR invalidated all CI caches... 🤔

PR #9607, opened 2026-01-14
TST: Skip many tests when running with pytest-run-parallel

seberg:
This skips many tests when running in free-threading mostly. It also does some smaller re-organizations at times.

The random tests (at least repeats) are judiciously skipped, it is just a bit of a tricky setup and there should be quite a lot of tests that still run anyway.

One change that may be larger is setting warnings to "always" which seemed necessary for some warnings tests, if that leads to a lot of spam in CI, we may want to skip that for now.

**EDIT: With the FFT PR, this is likely enough (or at least very close) for the `cupy_tests` to pass with `pytest-run-parallel`.**


tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -12,24 +12,29 @@
...
 
-    def tearDown(self):
-        self._prev_stream.use()
 
+@testing.parameterize(
+    *testing.product({
+        'stream_name': ['null', 'ptds'],
+    }))
+class TestStream(unittest.TestCase):
+    @set_and_restore_stream
```
seberg: The decorator-style is needed, since we must set the stream for each thread that is launched. I.e. the non-decorator version would set it before launching the threads and have no effect.

tests/cupy_tests/math_tests/test_sumprod.py
```diff
@@ -393,8 +395,8 @@ def _mitigate_cumprod(self, xp, dtype, result):
     reason='The cuTENSOR routine is not enabled')
 class TestCuTensorReduction:
 
-    @pytest.fixture(autouse=True)
-    def setUp(self):
+    @pytest.fixture(autouse=True, scope='class')
+    def setup(self):
```
seberg, line 401: I guess this does nothing, but once per-class should be OK. (probably part of me thinking `setUp` may be bad).

tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -240,21 +257,19 @@ def test_create_with_priority(self):
...
-    def tearDown(self):
-        cuda.runtime.streamDestroy(self.stream_ptr)
+class TestExternalStream:
+    @pytest.fixture
+    def stream(self):
+        # In pytest-run-parallel, shared between threads.
+        stream_ptr = cuda.runtime.streamCreate()
+        stream = cuda.ExternalStream(stream_ptr)
+        yield stream
+        cuda.runtime.streamDestroy(stream_ptr)
```
seberg: No real change in the end, but the fixture is nice with the `yield`...

tests/cupy_tests/cuda_tests/test_stream.py
```diff
@@ -192,7 +209,7 @@ def f1(barrier, errors):
                 print(f'error in {tid}: {e}')
 
         def f2(barrier, errors):
-            global s1
+            nonlocal s1
```
seberg: An actual fix, the `global` did what was promised and actually used a true global.
(Admittedly, the test is already threaded, so dunno if there is a point to thread it twice.)

leofang:
Ah... I just realized I also touched `test_stream.py` in #9567, not sure which one should be merged first?

seberg:
Let's not sweat it, I doubt it is tedious to resolve either way around.

seberg:
Hmmm, running the stream tests in parallel on a non free-threaded Python pointed out potential deadlocks with the callbacks. I added a commit to add `with nogil:` (and `with gil`) as needed to prevent those deadlocks (and maybe slightly more).  But happy to pull that out of this PR.

cupy_backends/cuda/api/runtime.pyx
leofang, line 1: good catch, I dunno why we missed it 👍

tests/cupy_tests/cuda_tests/test_memory.py
leofang, line 1: Not a blocker but we probably should follow up in a separate PR and make sure the mempool works in multithreading env. If this requires revising the tests, so be it.

leofang (dismissed):

leofang:
Let's get the CI pass!

tests/cupy_tests/cuda_tests/test_memory.py
seberg: The mempool PR includes some new multi-threaded tests.

seberg:
> Let's get the CI pass!

Ah, merge conflict.  The new test for swalling python errors strikes again on the `runtime.pyx` changes.
But I think that should be right, just need to ignore it.  Host callbacks are executed through the runtime and cannot propagate errors to Python.

cupy_backends/cuda/api/runtime.pyx
```diff
@@ -920,25 +924,28 @@ cpdef streamSynchronize(intptr_t stream):
...
-    func, arg = obj
-    func(<intptr_t>hStream, status, arg)
-    cpython.Py_DECREF(obj)
+cdef void _streamCallbackFunc(driver.Stream hStream, int status,
+                              void* func_arg) nogil noexcept:
+    with gil:
+        obj = <object>func_arg
+        cpython.Py_DECREF(obj)
+        func, arg = obj
+        func(<intptr_t>hStream, status, arg)
```
leofang, line 933: Q: if we decref first, wouldn't the gc kick in prematurely?

seberg: No, this is more a `DECREF(func_arg)` (we need `object`, not `PyObject *`).
I.e. `obj` owns a reference and will get `DECREF`ed by Cython at function exit.

(I just now briefly considered swapping it, but I think here it is no better since the `DECREF` must happen. If we, admittedly impossible, say `func, args = obj` would error then swapping would be wrong.)

leofang (approved):

PR #9608, opened 2026-01-14
Implement remaining quantile methods: inverted_cdf, averaged_inverted…

### LINKED ISSUE #6229: Support more interpolation methods in `cupy.percentile` and `cupy.quantile`

Harishjitu:
…_cdf, closest_observation, interpolated_inverted_cdf

Add support for the final four quantile estimation methods from NumPy 1.22:
- inverted_cdf (H&F type 1): Discontinuous, takes j if g > 0, else i
- averaged_inverted_cdf (H&F type 2): Discontinuous, takes j if g > 0, else average(i, j)
- closest_observation (H&F type 3): Discontinuous, rounds to nearest observation
- interpolated_inverted_cdf (H&F type 4): Continuous with alpha=0, beta=1

Closes #6229

### DIFF
--- a/cupy/_statistics/order.py
+++ b/cupy/_statistics/order.py
@@ -13,10 +13,11 @@ from cupy._logic import content
 # Quantile method parameters (alpha, beta) from Hyndman & Fan (1996)
 # Used for continuous interpolation methods in percentile/quantile
 _QUANTILE_PARAMS = {
-    'hazen': (0.5, 0.5),            # H&F type 5
-    'weibull': (0, 0),              # H&F type 6
-    'median_unbiased': (1/3, 1/3),  # H&F type 8
-    'normal_unbiased': (3/8, 3/8),  # H&F type 9
+    'interpolated_inverted_cdf': (0, 1),  # H&F type 4
+    'hazen': (0.5, 0.5),                  # H&F type 5
+    'weibull': (0, 0),                    # H&F type 6
+    'median_unbiased': (1/3, 1/3),        # H&F type 8
+    'normal_unbiased': (3/8, 3/8),        # H&F type 9
 }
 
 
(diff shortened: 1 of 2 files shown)


BarnikRB:
but the tests have not been implemented

cupy/_statistics/order.py
```diff
@@ -234,12 +235,56 @@ def _quantile_unchecked(a, q, axis=None, out=None,
...
+        )(idx_i, idx_j, g, ap, ap.shape[-1] if ap.ndim > 1 else 0, ret)
+        ret = cupy.rollaxis(ret, -1)
+
+        if zerod:
+            ret = ret.squeeze(0)
+        if keepdims:
+            if q.size > 1:
+                keepdim = (-1,) + keepdim
+            ret = ret.reshape(keepdim)
+        return _core._internal_ascontiguousarray(ret)
```
seberg, line 280: Sorry for the late reply.  While the custom kernel may be a bit faster, I would prefer to do what NumPy does (roughly speaking) and prepare `indices` in a way that the generic code path works fine.

(We can discuss adding it if you think it's worthwhile, my initial feeling is that we would need to do a lot more in the kernel for it to be and even the the difference may be small.)

cupy/_statistics/order.py
```diff
@@ -234,12 +235,56 @@ def _quantile_unchecked(a, q, axis=None, out=None,
...
+
+    elif method == 'closest_observation':
+        # H&F type 3: Round to nearest observation
+        virtual_indices = q * Nx
+        indices = cupy.around(virtual_indices) - 1
```
seberg, line 285: I think this is good, might be nice to confirm that we have a test that hits the rounding points, i.e. `virtual_indices` is always exactly `0.5, 1.5, 2.5`.
(The direction of rounding doesn't have to match NumPy strictly to be fair, so this may be more curiosity. -- The NumPy code doesn't actually use `around` and I am not sure why, it seems to do a manual round-to-even, but that is what `around` should be doing.)

Issue #9529, opened 2025-12-18
Support ROCm 7.1


kmaehashi:
## [Tasks](https://github.com/cupy/cupy/wiki/Actions-Needed-for-Dependency-Update)

- [ ] Read [ROCm Release Notes](https://docs.amd.com/).
- [ ] Update AMD driver in Self-Hosted CI infrastructure (ask @kmaehashi).
- [ ] Fix code and CI to support the new version.
  - **FlexCI**: Update `.pfnci/schema.yaml` and `.pfnci/matirx.yaml`.
  - **Wheel Package Detection**: Add the package to the [duplicate detection](https://github.com/cupy/cupy/blob/master/cupy/_environment.py).
- [ ] Backport the above PR.
- [ ] Fix `cupy-release-tools` to support the new version.
- [ ] Backport the above PR.
- [ ] Fix documentation.
  - Add new wheel package to the [Installation Guide](https://docs.cupy.dev/en/latest/install.html) and `README.md`.
  - Update requirements in the installation guide.
- [ ] Backport the above PR.
- [ ] Add new wheel package to the [website](https://cupy.dev).
- [ ] Implement or create an issue to support new features, if applicable.

reneleonhardt:
It seems https://rocm.docs.amd.com/en/latest/about/release-notes.html contain the Release Notes for ROCm 7.1.1 😉

gpinkert:
Hey - as far as I can tell, cupy will build already on ROCm 7.1. Internally, we are still using cupy v13.5 (with the ROCm 7 patches) so there hasn't been a push to update our pipelines. I will add it to my sprint to get this checked and handled and backported in the next week or so.

yanite:
Request support for Windows ROCM 7.1

reneleonhardt:
Is it possible to add support for ROCm 7.2 too?
https://www.phoronix.com/news/AMD-ROCm-7.2-Released
https://rocm.docs.amd.com/en/latest/about/release-notes.html

VaterVato:
Request support for Windows ROCM 7.X

reneleonhardt:
ROCm 7.2 is a month old, when will it be used in pretest.yml?

pzeiger:
Hi @gpinkert,
thank you for your great work!

[...]

PR #9615, opened 2026-01-16
Upgrade rocm test environment to 7.1

gpinkert:
xref #9529

### DIFF
--- a/.pfnci/coverage.rst
+++ b/.pfnci/coverage.rst
@@ -2698,7 +2698,7 @@ CuPy CI Test Coverage
 .. _t21: https://ci.preferred.jp/cupy.linux.cuda131.multi/
 .. _d21: linux/tests/cuda131.multi.Dockerfile
 .. _s21: linux/tests/cuda131.multi.sh
-.. _t22: https://jenkins.preferred.jp/job/chainer/job/cupy_main/TEST=rocm-7-0,label=mnj-mi50/
+.. _t22: https://jenkins.preferred.jp/job/chainer/job/cupy_main/TEST=rocm-7-1,label=mnj-mi50/
 .. _d22: linux/tests/rocm-7-1.Dockerfile
 .. _s22: linux/tests/rocm-7-1.sh
 .. _t23: https://ci.preferred.jp/cupy.linux.cuda-slow/
--- a/.pfnci/matrix.yaml
+++ b/.pfnci/matrix.yaml
@@ -348,7 +348,7 @@
 # The lowest ROCm version matrix is intended to cover the lowest supported combination.
 # The latest ROCm version matrix is intended to cover the highest supported combination.
 - project: "cupy.linux.rocm-7-1"
-  _url: "https://jenkins.preferred.jp/job/chainer/job/cupy_main/TEST=rocm-7-0,label=mnj-mi50/"
+  _url: "https://jenkins.preferred.jp/job/chainer/job/cupy_main/TEST=rocm-7-1,label=mnj-mi50/"
   tags: null  # self-hosted
   target: "rocm-7-1"
   system: "linux"


gpinkert:
Unfortunately  it will take me a while to get permissions to clone cupy-releaes-tools. If it takes much longer I will just clone it under my amd profile and update it there

gpinkert:
https://github.com/cupy/cupy-release-tools/pull/479

gpinkert:
Let me know if there is anything else I need to do @kmaehashi other than the documentation. I will handle that after this is merged.

cupy/_environment.py
```diff
@@ -570,7 +570,7 @@ def _detect_duplicate_installation():
         'cupy',
         'cupy-cuda12x',
         'cupy-cuda13x',
-        'cupy-rocm-7-0',
+        'cupy-rocm-7-1',
```
leofang, line 574: @kmaehashi would probably want to confirm:  Do we ship `cupy-rocm-7-0` or `cupy-rocm-7-1` by CuPy v14.0.0?
- If it's the former, this file should not be touched
- If it's the latter, then this diff is fine

kmaehashi: `cupy-rocm-7-0` shouldn't be removed anyway as we already shipped it in v14rc1 release (although it's only in our private index). Let's just add `cupy-rocm-7-1` here. I want to get this one and https://github.com/cupy/cupy-release-tools/pull/479 in for v14 final release if the time allows.

kmaehashi:
@gpinkert Could you also update the build-rocm CI here?
https://github.com/cupy/cupy/blob/0569cd65381a36b0cd726fb0d3599baff3dc5eb9/.github/workflows/pretest.yml#L100-L112

cupy/_environment.py
leofang: I have added a blocking label to this PR. To conclude this thread I think we need this patch:
```suggestion
        'cupy-rocm-7-0',
        'cupy-rocm-7-1',
```

gpinkert: @leofang sorry if I am missing something, but is that not what the current diff shows?

leofang: It seems a recent push has changed to what's suggested. This thread was made when it was not yet the case 😅

leofang (approved):

leofang:
@gpinkert it seems the CI has failed, could you take a look? 🙂 
https://github.com/cupy/self-hosted-ci/actions/runs/21641760007/job/62383092674

.github/workflows/self-hosted-ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         token: ${{ steps.generate-token.outputs.token }}
         client-payload: |
           {
-            "targets": ["rocm-7-0"],
+            "targets": ["rocm-7-1"],
```
leofang, line 96: hmmm it seems the `client-payload.targets` is already updated, but `rocm-7-0` is still in use: 
https://github.com/cupy/self-hosted-ci/actions/runs/21657825134/job/62436214048#step:1:17

@kmaehashi is this the same security limitation that such changes aren't honored until merged into main?

leofang: Or, we also need to update this line (but we don't do `workflow_dispatch` here do we? 🤔)
https://github.com/cupy/self-hosted-ci/blob/98d1bfc6450903ca3bc8fd44ed0349bedb6499ff/.github/workflows/ci.yml#L10

gpinkert:
@kmaehashi do we have to let this merge before the CI will pick it up?

.github/workflows/self-hosted-ci.yml
kmaehashi: Currently we can't test the workflow updated in PR without merging it first, due to the security restriction of GitHub Actions. I have an idea to relax this restriction, will do that separately.

kmaehashi:
Thanks Grant, I think this is ready to merge! Let's see how it goes after merging.

kmaehashi (approved):

PR #9621, opened 2026-01-19
BUG: Fix RandomState thread safety (and memory initialization)

seberg:
This adds a lock to the RandomState and refactors it to be thread-safe. Technically, the lock on the `_generator` should only be needed on free-threaded Python.
However, compared to the full overhead, adding the lock seems relatively small, so I figured it is worth the simplicity/clarity.

Running the test on it's own did flush out a mistake in my thinking. The critical section in the memory-pool init was not sufficient (it seems it got released in the tuple expression).

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -1604,17 +1604,18 @@ cdef class MemoryPool:
             allocator = _malloc
         self._allocator = allocator
 
-    @cython.critical_section  # ensure we are only one setting _pools
     cdef _ensure_pools_and_return_device_pool(self):
-        if self._pools is None:
-            # It would actually be OK if we create pools more than once
-            # as long we do not set `self._pools` exactly once and always
-            # return the pool from those in `self._pools`.
-            n_gpu = runtime.getDeviceCount()
-            pools = tuple(
-                SingleDeviceMemoryPool(self._allocator) for i in range(n_gpu))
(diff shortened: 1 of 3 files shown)


cupy/cuda/memory.pyx
leofang, line 1: I don't get why the previous implementation is buggy 😅

seberg: Honestly, I also don't but that is because we don't know how critical sections are implemented, I guess...

If this was the GIL, then building the tuple/creating it might unlock the GIL, so it isn't 100% safe with it (in principle).
My new understanding is that critical sections unlock on some/all(?) C-API calls to prevent deadlocks, but I am not quite sure (yet).

tests/cupy_tests/random_tests/test_distributions.py
```diff
@@ -852,3 +854,116 @@ class TestDistributionsZipf(RandomDistributionsTestCase):
...
+                 id='permutation'),
+    pytest.param(lambda rs: rs.randint(0, 10, size=10),
+                 id='randint'),
+    pytest.param(lambda rs: rs.randn(10),
+                 id='randn'),
+])
+@pytest.mark.thread_unsafe("already multi-threaded")
+def test_multithreaded(dist_func):
+    n_threads = 10
+    rs = cupy.random.RandomState(seed=0)
```
leofang, line 950: I see, so we test multithreading by making all threads sharing the same random state.

Issue #9629, opened 2026-01-23
`TestChoiceReplaceFalse` fails on Windows


leofang:
```
05:36:09.318052 STDOUT 4968]	=========================== short test summary info ===========================	
05:36:09.318052 STDOUT 4968]	FAILED cupy_tests\random_tests\test_generator.py::TestChoiceReplaceFalse_param_0_{a=5, size=2}::test_dtype_shape - AssertionError: assert dtype('int64') == 'long'	
05:36:09.318052 STDOUT 4968]	 +  where dtype('int64') = array([0, 4]).dtype	
05:36:09.318052 STDOUT 4968]	FAILED cupy_tests\random_tests\test_generator.py::TestChoiceReplaceFalse_param_1_{a=5, size=(2, 2)}::test_dtype_shape - AssertionError: assert dtype('int64') == 'long'	
05:36:09.318052 STDOUT 4968]	 +  where dtype('int64') = array([[0, 4],\n       [3, 1]]).dtype	
05:36:09.318747 STDOUT 4968]	FAILED cupy_tests\random_tests\test_generator.py::TestChoiceReplaceFalse_param_2_{a=5, size=()}::test_dtype_shape - AssertionError: assert dtype('int64') == 'long'	
05:36:09.318780 STDOUT 4968]	 +  where dtype('int64') = array(0).dtype	
05:36:09.318780 STDOUT 4968]	= 3 failed, 108982 passed, 3995 skipped, 775 deselected, 598 xfailed, 2613 warnings in 14648.42s (4:04:08) =	
05:36:09.318780 STDOUT 4968]	------------------------------------------------------------------------------------------
```
https://ci.preferred.jp/cupy.win.cuda129/207386/

PR #9631, opened 2026-01-25
Fix Windows CI + Tightening up `/test` logic

leofang:
1. https://github.com/cupy/cupy/issues/9629 revealed a bug in the CI script that even if the pipeline fails/terminates, it'd still report success. I think the bug was introduced in #9562, so this PR is a partial revert.
2. I often find the need of only triggering a subset of runners, but the current `/test` command would trigger more than needed. This PR tightens up the logic from finding a superset to a subset. For example:
   - `/test mini,linux` → I want this to trigger only Linux mini tests, not all `mini` tests and all `linux` tests
   - `/test mini,windows` → same logic
   - `/test cuda130,linux` → same logic

### DIFF
--- a/.github/workflows/scripts/flexci_dispatcher.py
+++ b/.github/workflows/scripts/flexci_dispatcher.py
@@ -236,7 +236,7 @@ def main(argv: Any) -> int:
     projects_dispatch: set[str] = set()
     projects_skip: set[str] = set()
     for project, tags in project_tags.items():
-        dispatch = (len(set(tags) & requested_tags) != 0)
+        dispatch = (requested_tags <= set(tags))
         if dispatch:
             projects_dispatch.add(project)
         else:
--- a/.pfnci/config.tags.json
+++ b/.pfnci/config.tags.json
@@ -3,6 +3,7 @@
         "@push",
         "full",
         "mini",
+        "linux",
         "cuda120"
     ],
     "cupy.linux.cuda120.multi": [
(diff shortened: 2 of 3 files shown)


.github/workflows/scripts/flexci_dispatcher.py
```diff
@@ -236,7 +236,7 @@ def main(argv: Any) -> int:
     projects_dispatch: set[str] = set()
     projects_skip: set[str] = set()
     for project, tags in project_tags.items():
-        dispatch = (len(set(tags) & requested_tags) != 0)
+        dispatch = (requested_tags <= set(tags))
```
leofang, line 239: Ah. Alright. I was wondering why the superset of runners are still triggered. It seems this change won't become effective until after it's merged 😓 Below is Copilot answer.

Looking at the workflow:

     - Line 37-40: The workflow checks out the base branch (upstream) first
     - Line 42-47: For issue_comment events (i.e., /test commands), it checks out the PR code to _pr_untrusted/
     - Line 62: PROJECTS_ROOT_DIR is set to _pr_untrusted for issue_comment events
     - Line 64: The dispatcher script is run from the base branch (./.github/workflows/scripts/flexci_dispatcher.py)
     - Line 67: But config.tags.json is read from the PR branch (${PROJECTS_ROOT_DIR}/.pfnci/config.tags.json)

   The problem: The flexci_dispatcher.py script is always from the base branch (upstream), not the PR branch. So even though my changes to flexci_dispatcher.py are in the PR, they won't be used until merged.

.pfnci/windows/test.ps1
leofang, line 1: With the revert we should be able to trigger #9629 and cause CI failure.

leofang:
> (although, maybe if we need super-set we should have a spelling for both.
> (Such as using `|` or `;` to split groups.)

Yeah, I am wondering if we could just do this via issuing multiple `/test` comments.

.github/workflows/scripts/flexci_dispatcher.py
kmaehashi: Yes. For security reasons we don't trust infra code within PRs.

kmaehashi:
/test force-skip
Timeouts.

PR #9644, opened 2026-01-27
CI: Update cache handling in CI

kmaehashi:
After investigating the cache details, I found that there are many cache files that are not cleaned up. For example, CUDA 12.4 Linux CI cache contained:

* 6.0G: FFT callback cache (e.g. `cupy_callback_0ab11671b92a58bf381e386efe1f7152.cpython-312-x86_64-linux-gnu.so`)
* 4.9G: kernel cache (e.g. `0000747945674e27d08f6d586270db6b5e93e3ff.cubin`) -- previously `cleanup.sh` only takes care of this
* 506M: Jitify cahce (e.g. `jitify_unknown_200600_12040_3_d65d3d46fed76d4a6f7f6d143f355659f84a958e.json`)
* 310M: Pruned cuFFT static library (e.g. `libcufft_static_11201_sm7.a`)

I assume this gradually increased the cache size, and started to cause `no space left on device` errors in recent CI runs. Note that disk size required is twice the cache size during un/archiving cache directory to TAR/ZIP. This PR adjusts the cache management in CI as follows:

* Clean up all of the above cache files.
* Adjust cache expiry and strategy to keep the cache size slim.
* Always run cleanup even if build or unittest step fails.

### DIFF
--- a/.pfnci/windows/test.ps1
+++ b/.pfnci/windows/test.ps1
@@ -28,9 +28,10 @@ function DownloadCache([String]$gcs_dir, [String]$cupy_kernel_cache_file) {
 
 function UploadCache([String]$gcs_dir, [String]$cupy_kernel_cache_file) {
     # Limit *.cubin cache by total size.
-    # TODO: Also clean up $CUPY_CACHE_DIR/callback_cache/*.ltoir
+    # Note: --expiry cannot be used as access time is not updated on Windows.
+    # TODO: Also clean up ~/.cupy/callback_cache/*.ltoir
     echo "Trimming kernel cache..."
-    RunOrDie python .pfnci\trim_cupy_kernel_cache.py --max-size 6442450944 --rm
+    RunOrDie python .pfnci\trim_cupy_kernel_cache.py --max-size 4831838208 --rm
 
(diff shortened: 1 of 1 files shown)


kmaehashi:
/test force-skip

.pfnci/linux/tests/actions/cleanup.sh
```diff
@@ -2,5 +2,18 @@
...
 
-python3 .pfnci/trim_cupy_kernel_cache.py --max-size $((3*1024*1024*1024)) --rm
+# Note: Linux CI image uses relatime.
+
+# Expire misc cache in 30 days.
+echo "Cleaning up misc cache:"
+find "${CUPY_CACHE_DIR}" -type f \( \
+    -name "jitify_*.json" -or \
+    -name "cupy_callback_*.so" -or \
+    -name "libcufft_static_*.a" \
```
leofang, line 12: I think the `.a` files can be removed after each run. As long as the `.so` files exist, we won't need them; they are only used to generate the `.so` files.

leofang: tracked in #9666

PR #9647, opened 2026-01-28
Remove cupyx.scipy.linalg.kron

asi1024:
Part of #9690.
https://docs.scipy.org/doc/scipy/release/1.17.0-notes.html#expired-deprecations


kmaehashi:

/test force-skip

PR #9648, opened 2026-01-28
Support SciPy 1.17

asi1024:
Part of #9690


seberg:
@eriknw if you have some time, maybe you can have a look as well?  Should maybe add a test for the new helpers as well.
This is a rebase of the original to add SciPy 1.17 plus test fixups.  I added two helpers (private and hidden right now):
* `is_after_baseline`
* `skip_if_after_baseline`

The idea being that we can use `@skip_if_after_baseline(scipy="1.17")` to skip tests that break.  But when we bump the baseline to `1.17+` we _hopefully_ update the constant in the test code and the tests will start failing.
(Because I am worried about skipping tests now but missing skips that should be updated when the baseline is incremented... If we don't want to I could also remove this and maybe add a comment or so.)

seberg:
@asi1024 do you have time to briefly look at my changes, I am hesitant to self-merge, but this should move and keeps getting conflicts.

asi1024:
@seberg Thanks for resolving the conflicts! LGTM.

seberg (approved):

Issue #2031, opened 2019-02-14
Passing a list dtype fails


mrocklin:
It looks like cupy supports record arrays, but fails to accept a list dtype that was not properly packed as a dtype object.   My guess is that this could 

```python
In [1]: import numpy as np

In [2]: import cupy

In [3]: x = cupy.empty(shape=5, dtype=np.dtype([('a', int), ('b', float)]))

In [4]: x
Out[4]:
array([(0, 0.), (0, 0.), (0, 0.), (0, 0.), (0, 0.)],
      dtype=[('a', '<i8'), ('b', '<f8')])

In [5]: x = cupy.empty(shape=5, dtype=[('a', int), ('b', float)])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-8478cd56259b> in <module>
----> 1 x = cupy.empty(shape=5, dtype=[('a', int), ('b', float)])

~/cupy/cupy/creation/basic.py in empty(shape, dtype, order)
     20
     21     """
---> 22     return cupy.ndarray(shape, dtype, order=order)
     23
     24

~/cupy/cupy/core/core.pyx in cupy.core.core.ndarray.__init__()

~/cupy/cupy/core/_dtype.pyx in cupy.core._dtype.get_dtype_with_itemsize()

TypeError: unhashable type: 'list'
```

Numpy handles this ok.  My guess is that this could be resolved by a quick typecheck, or just always calling `np.dtype` on the input to any `dtype=` keyword.

mrocklin:
So the following change is probably needed and welcome on the dask side

```diff
-   result = empty(n.shape, dtype=[('total', total.dtype), ('n', n.dtype)])
+   result = empty(n.shape, dtype=np.dtype([('total', total.dtype), ('n', n.dtype)]))
```

mrocklin:
More generally, is there a nice way to construct an array with structured dtypes from arrays?

```python
In [1]: import numpy as np

In [2]: x = np.empty(shape=5, dtype=np.dtype([('a', int), ('b', float)]))

In [3]: a = np.ones(5, dtype=int)

In [4]: b = np.ones(5, dtype=float)

In [5]: x['a'] = a

In [6]: x['b'] = b

In [7]: x
Out[7]:
array([(1, 1.), (1, 1.), (1, 1.), (1, 1.), (1, 1.)],
      dtype=[('a', '<i8'), ('b', '<f8')])
```

kmaehashi:
[...]

PR #9650, opened 2026-01-28
ENH: (almost) full structured dtype support

seberg:
**~This is quite far along, but needs cleaning up (and is based on the other PR). So it is a WIP only, just for visibility.~**

Based on gh-9494, this adds the missing support for structured dtypes, that is, allowing casts and multi-field access/assignments.

Major open points:
* We need to thread through the `dtype` much more. This is OK, but we need to refactor things to make use of the NumPy C-API to reduce overheads again.
  * It may make sense to do a bit larger refactors around this.  *EDIT:* Actually, I don't think these changes cause any measurable speed regression, some things may be nice but not much needed here.
* NumPy doesn't help us align types properly, so we'll need a helper to do so.
  * I am curious if we couldn't query the `alignof()` through the JIT to implement this? (plus memoization).
  * This will currently lead to some incorrect paths even for correct inputs in theory (i.e. the comparison compares via a promoted type and that promoted type may be unaligned)
* ~It needs tests of course!~

Smaller points:
* I would like this to be interoperable with a plain struct, if such a plain struct can work. That logic exists, but likely doesn't quite work yet.
* This PR does *not* implement subarray support.

To implement this I added a `resolve_dtypes()` function. This is a concept that NumPy also has, although I made it a bit more raw here. (The divergence has two reasons. First casts as a ufunc don't work well without it, although a cast could maybe be a simple `ElementwiseKernel`. But also the functionality used in NumPy to actual implement it nicely isn't public, oops.)

Addresses gh-2031 (but one could argue that subarray support is missing)

### DIFF
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -206,7 +206,7 @@ cdef void populate_format(Py_buffer* buf, str dtype) except*:
     buf.format = dtype_format[dtype]
 
 
-def _make_gpu_aligned_dtype(
+def _make_aligned_dtype(
         dtype, *, int alignment=-1, bint recurse=True):
     # Internal version that also return the final alignment for recursion.
     cdef Py_ssize_t final_alignment = 1
@@ -232,7 +232,7 @@ def _make_gpu_aligned_dtype(
             # (A title is an alternative field name...)
             if offset < min_offset:
                 raise ValueError(
-                    "make_gpu_aligned_dtype() only supports well behaved "
+                    "make_aligned_dtype() only supports well behaved "
                     "in order fields as it ignores field offsets).")
 
             # Keep track of field offset to reject non-ordered inputs.
@@ -251,7 +251,7 @@ def _make_gpu_aligned_dtype(
                 # Must assume the alignment of the subdtype makes sense.
                 subalignment = subdtype.alignment
             else:
-                subdtype, subalignment = _make_gpu_aligned_dtype(
+                subdtype, subalignment = _make_aligned_dtype(
                     subdtype, recurse=recurse)
 
             if curr_offset % subalignment != 0:
@@ -274,7 +274,7 @@ def _make_gpu_aligned_dtype(
             metadata = {"metadata": {"__cuda_alignment__": alignment}}
         else:
             raise ValueError(
-                f"make_gpu_aligned_dtype(): given alignment={alignment} "
+                f"make_aligned_dtype(): given alignment={alignment} "
                 f"smaller than minimum alignment {final_alignment}"
             )
 
@@ -296,7 +296,7 @@ def _make_gpu_aligned_dtype(
     return cnp.dtype(dtype_info, align=True, **metadata), final_alignment
 
 
-def make_gpu_aligned_dtype(
+def make_aligned_dtype(
         dtype, *, int alignment=-1, bint recurse=True):
     """Create a new structured dtype from a NumPy dtype or dtype-like with
     sufficient algnment for GPU use.
@@ -326,10 +326,10 @@ def make_gpu_aligned_dtype(
     Notes:
         By default this function recurses into nested structures as if
         `alignment=-1` is passed for these.  You can nest a dtype with
-        larger alignment by creating it with ``make_gpu_aligned_dtype()``.
+        larger alignment by creating it with ``make_aligned_dtype()``.
 
         NumPy promotion (e.g. in concatenate) may "canonicalize" the dtype
         and drop the struct layout and CuPy alignment metadata.
     """
-    return _make_gpu_aligned_dtype(
+    return _make_aligned_dtype(
         dtype, alignment=alignment, recurse=recurse)[0]
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -276,7 +276,7 @@ def make_gpu_aligned_dtype(
                 subalignment = dtype.alignment
                 if dtype.metadata:
                     subalignment = subdtype.metadata.get(
-                        "__cupy_alignment__", subalignment)
+                        "__cuda_alignment__", subalignment)
 
             if curr_offset % subalignment != 0:
                 curr_offset += subalignment - (curr_offset % subalignment)
@@ -295,7 +295,7 @@ def make_gpu_aligned_dtype(
     if alignment != -1:
         if alignment >= final_alignment:
             final_alignment = alignment
-            metadata = {"metadata": {"__cupy_alignment__": alignment}}
+            metadata = {"metadata": {"__cuda_alignment__": alignment}}
         else:
             raise ValueError(
                 f"make_gpu_aligned_dtype(): given alignment={alignment} "
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -206,37 +206,9 @@ cdef void populate_format(Py_buffer* buf, str dtype) except*:
     buf.format = dtype_format[dtype]
 
 
-def make_gpu_aligned_dtype(
-        dtype, *, int alignment=-1, field_alignments=None, int recurse=False):
-    """Create a new structured dtype from a NumPy dtype or dtype-like with
-    sufficient algnment for GPU use.
-
-    Args:
-        dtype: Data type specifier compatible with NumPy.
-        alignment: Desired alignment of the resulting dtype. The dtype
-            will be padded to ensure this alignment for CuPy.
-
-            .. note::
-                When the requested alignment is smaller than the minimal
-                inferred one an error will be raised.
-                When it is larger, CuPy will attach this alignment as
-                metadata to the structured dtype. This is used for structured
-                dtypes in the kernel and when nesting.
-
-                Note that metadata may be lost in many operations.
-
-    Returns:
-        cupy.ndarray: A view of the array with reduced dimensions.
-
-
-    Notes:
-        By default this function recurses into nested structures as if
-        `alignment=-1` is passed for these.  You can nest a dtype with
-        larger alignment by creating it with ``make_gpu_aligned_dtype()``.
-
-        NumPy promotion (e.g. in concatenate) may "canonicalize" the dtype
-        and drop the struct layout and CuPy alignment metadata.
-    """
+def _make_gpu_aligned_dtype(
+        dtype, *, int alignment=-1, bint recurse=True):
+    # Internal version that also return the final alignment for recursion.
     cdef Py_ssize_t final_alignment = 1
     cdef Py_ssize_t itemsize = 0
     cdef Py_ssize_t curr_offset = 0
@@ -266,17 +238,21 @@ def make_gpu_aligned_dtype(
             # Keep track of field offset to reject non-ordered inputs.
             min_offset = offset + subdtype.itemsize
 
-            if subdtype.num != cnp.NPY_VOID or subdtype.fields is None:
+            subalignment = None
+            if subdtype.metadata:
+                subalignment = subdtype.metadata.get("__cuda_alignment__")
+
+            if subalignment is not None:
+                # __cuda_alignment__ is defined and overrides everything else.
+                pass
+            elif subdtype.num != cnp.NPY_VOID or subdtype.fields is None:
                 subalignment = _scalar.get_cuda_alignment(subdtype)
             elif not recurse:
-                # We assume the alignment of the subdtype makes sense.
+                # Must assume the alignment of the subdtype makes sense.
                 subalignment = subdtype.alignment
             else:
-                subdtype = make_gpu_aligned_dtype(subdtype, recurse=recurse)
-                subalignment = dtype.alignment
-                if dtype.metadata:
-                    subalignment = subdtype.metadata.get(
-                        "__cuda_alignment__", subalignment)
+                subdtype, subalignment = _make_gpu_aligned_dtype(
+                    subdtype, recurse=recurse)
 
             if curr_offset % subalignment != 0:
                 curr_offset += subalignment - (curr_offset % subalignment)
@@ -317,4 +293,43 @@ def make_gpu_aligned_dtype(
         dtype_info["itemsize"] = itemsize
 
     # Create a new dtype enforcing the newly computed offsets.
-    return cnp.dtype(dtype_info, align=True, **metadata)
+    return cnp.dtype(dtype_info, align=True, **metadata), final_alignment
+
+
+def make_gpu_aligned_dtype(
+        dtype, *, int alignment=-1, bint recurse=True):
+    """Create a new structured dtype from a NumPy dtype or dtype-like with
+    sufficient algnment for GPU use.
+
+    Args:
+        dtype: Data type specifier compatible with NumPy.
+        alignment: Desired alignment of the resulting dtype. The dtype
+            will be padded to ensure this alignment for CuPy.
+
+            .. note::
+                When the requested alignment is smaller than the minimal
+                inferred one an error will be raised.
+                When it is larger, CuPy will attach this alignment as
+                metadata to the structured dtype. This is used for structured
+                dtypes in the kernel and when nesting.
+
+                Note that metadata may be lost in many operations.
+
+        recurse: Whether to recurse into nested structures, defaults to True.
+            When ``False``, the nested struct is assumed to be already
+            sufficientlyaligned.
+
+    Returns:
+        cupy.ndarray: A view of the array with reduced dimensions.
+
+
+    Notes:
+        By default this function recurses into nested structures as if
+        `alignment=-1` is passed for these.  You can nest a dtype with
+        larger alignment by creating it with ``make_gpu_aligned_dtype()``.
+
+        NumPy promotion (e.g. in concatenate) may "canonicalize" the dtype
+        and drop the struct layout and CuPy alignment metadata.
+    """
+    return _make_gpu_aligned_dtype(
+        dtype, alignment=alignment, recurse=recurse)[0]
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -62,7 +62,7 @@ cdef bint check_supported_dtype(cnp.dtype dtype, bint error) except -1:
 
         # We can represents the underlying bytes in CuPy, although it is very
         # possible that the included fields will not be usable in the end.
-        # The simplest path is to inform users
+        # I.e. an error may be raised when launching a kernel with this.
         return True
 
     try:
@@ -204,3 +204,117 @@ cdef dict dtype_format = {
 @cython.profile(False)
 cdef void populate_format(Py_buffer* buf, str dtype) except*:
     buf.format = dtype_format[dtype]
+
+
+def make_gpu_aligned_dtype(
+        dtype, *, int alignment=-1, field_alignments=None, int recurse=False):
+    """Create a new structured dtype from a NumPy dtype or dtype-like with
+    sufficient algnment for GPU use.
+
+    Args:
+        dtype: Data type specifier compatible with NumPy.
+        alignment: Desired alignment of the resulting dtype. The dtype
+            will be padded to ensure this alignment for CuPy.
+
+            .. note::
+                When the requested alignment is smaller than the minimal
+                inferred one an error will be raised.
+                When it is larger, CuPy will attach this alignment as
+                metadata to the structured dtype. This is used for structured
+                dtypes in the kernel and when nesting.
+
+                Note that metadata may be lost in many operations.
+
+    Returns:
+        cupy.ndarray: A view of the array with reduced dimensions.
+
+
+    Notes:
+        By default this function recurses into nested structures as if
+        `alignment=-1` is passed for these.  You can nest a dtype with
+        larger alignment by creating it with ``make_gpu_aligned_dtype()``.
+
+        NumPy promotion (e.g. in concatenate) may "canonicalize" the dtype
+        and drop the struct layout and CuPy alignment metadata.
+    """
+    cdef Py_ssize_t final_alignment = 1
+    cdef Py_ssize_t itemsize = 0
+    cdef Py_ssize_t curr_offset = 0
+    cdef Py_ssize_t min_offset = 0  # offset before additional padding
+    cdef cnp.dtype descr = cnp.dtype(dtype, align=True)
+
+    if alignment != -1 and (alignment <= 0 or alignment.bit_count() != 1):
+        raise ValueError("Reasonable alignments must be >0 and a power of 2.")
+
+    if descr.type_num != cnp.NPY_VOID or (<object>descr).fields is None:
+        # Allow this when `alignment=` is passed, we try to add the meatadat
+        dtype_info = descr  # unchanged for now
+        final_alignment = descr.alignment
+    else:
+        names = []
+        offsets = []
+        subdtypes = []
+
+        for name, (subdtype, offset, *_) in (<object>descr).fields.items():
+            # The fields tupe can contain a 4th title, we ignore it.
+            # (A title is an alternative field name...)
+            if offset < min_offset:
+                raise ValueError(
+                    "make_gpu_aligned_dtype() only supports well behaved "
+                    "in order fields as it ignores field offsets).")
+
+            # Keep track of field offset to reject non-ordered inputs.
+            min_offset = offset + subdtype.itemsize
+
+            if subdtype.num != cnp.NPY_VOID or subdtype.fields is None:
+                subalignment = _scalar.get_cuda_alignment(subdtype)
+            elif not recurse:
+                # We assume the alignment of the subdtype makes sense.
+                subalignment = subdtype.alignment
+            else:
+                subdtype = make_gpu_aligned_dtype(subdtype, recurse=recurse)
+                subalignment = dtype.alignment
+                if dtype.metadata:
+                    subalignment = subdtype.metadata.get(
+                        "__cupy_alignment__", subalignment)
+
+            if curr_offset % subalignment != 0:
+                curr_offset += subalignment - (curr_offset % subalignment)
+
+            final_alignment = max(final_alignment, subalignment)
+
+            names.append(name)
+            subdtypes.append(subdtype)
+            offsets.append(curr_offset)
+            curr_offset += subdtype.itemsize
+
+        dtype_info = dict(names=names, offsets=offsets, formats=subdtypes,
+                          itemsize=itemsize)
+
+    metadata = {}
+    if alignment != -1:
+        if alignment >= final_alignment:
+            final_alignment = alignment
+            metadata = {"metadata": {"__cupy_alignment__": alignment}}
+        else:
+            raise ValueError(
+                f"make_gpu_aligned_dtype(): given alignment={alignment} "
+                f"smaller than minimum alignment {final_alignment}"
+            )
+
+    itemsize = (
+        (curr_offset + final_alignment - 1) // final_alignment
+        * final_alignment)
+
+    if descr.type_num != cnp.NPY_VOID:
+        if descr.itemsize != itemsize:
+            raise ValueError(
+                "Alignment larger than itemsize for non-structured dtype.")
+    else:
+        if descr.itemsize > itemsize:
+            raise ValueError(
+                "Input descriptor had larger itemsize than inferred.")
+        dtype_info["itemsize"] = itemsize
+
+    # Create a new dtype enforcing the newly computed offsets.
+    return cnp.dtype(dtype_info, align=True, **metadata)
--- a/cupy/_core/_dtype.pyx
+++ b/cupy/_core/_dtype.pyx
@@ -222,6 +222,7 @@ def _make_aligned_dtype(
         # Allow this when `alignment=` is passed, we try to add the metadata
         dtype_info = descr  # unchanged for now
         final_alignment = _scalar.get_cuda_alignment(descr)
+        curr_offset = descr.itemsize  # asserts itemsize multiple of alignment
     else:
         names = []
         offsets = []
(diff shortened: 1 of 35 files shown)


cupy/_core/include/cupy/structview.cuh
```diff
@@ -0,0 +1,192 @@
...
+
+//#include <cstddef>
+//#include <type_traits>
+//#include <utility>
+
+namespace sv {
+
+// Hack to replace std::false_type and std::true_type
+template<bool Value>
+struct my_integral_constant {
```
leofang: FYI `std::integral_constant` is available through `cupy/workaround.h`

cupy/_core/_dtype.pyx
```diff
@@ -203,3 +203,117 @@ cdef dict dtype_format = {
 @cython.profile(False)
 cdef void populate_format(Py_buffer* buf, str dtype) except*:
     buf.format = dtype_format[dtype]
+
+
+def make_gpu_aligned_dtype(
```
leofang: Two questions:
1. How does this relate to GPU, other than that we ensure alignment? I assume `make_aligned_dtype` is fine too?
2. How about moving this to `cuda.core`? We need this function for libraries lower than CuPy in the stack (https://github.com/NVIDIA/cuda-python/issues/734).

seberg: Moving sounds good. CPU alignment is smaller than GPU one often. While GPU is (typically?) itemsize aligned. An example that always works is complex numbers, which CUDA uses the full itemsize alignment for copies, I think, while for the CPU the compiler will only require the floats alignment.

FWIW, I am not sure that encoding a custom larger alignment is super useful. It seems mainly needed for nested structs which dould maybe be solve directly by passing field alignment constraints offset (that are only used to compute offsets).

leofang: Yeah for 1 I mainly wanted to clarify for the function name. It seems calling it `make_aligned_dtype()` (removing "gpu") is fine. There is nothing GPU-specific.

For 2, my suggestion is
- move it to cuda.core (I can take over this if you like, or if you have time to check it there it's also fine 🙂 )
- try-import it in CuPy, if cuda.core is new enough to offer it, use it
- otherwise, use a private copy vendored in CuPy (and we can remove it in 1-2 major releases)

seberg: Sorry yeah, I'll work on moving it. Thought I may have an idea about how to signal alignment, but likely this is pretty good (and we only embed if users asked for somdthing special).

(I suppose we cannot move the C++ type construction, might befun one day, but not a priority.)

seberg:
Ok, this grew to 1k lines, so maybe should split it up. But while it still needs some polishing, I think this should be good for review.

The basic changes are:
* Introduce a header for field based struct view (and one for unstructured, we don't have `std::array`, I think)
* Create a helper `cupy.make_aligned_dtype`, we need this internally as well.
* Create Python code to build the C type from the structured dtype (i.e. `get_typename()`), this is fairly complicated!
  * Currently, it builds a struct if possible. This may serve no real purpose, so maybe we should just remove it (and maybe add field names via `.at<"name">` if that works.)
* We need a `resolve_dtypes` function, this is needed to figure out the right loop dtypes for comparisons.
* CScalars may have arbitrary size now and in a few places we needed to propagate them better.
* `get_typename(..., type_headers)` is now `type_decls` (changed everywhere).  Dependencies matter now, so it needs a topological sort.  Feels a bit overpowered, but also kinda OK...

I am sure there will be places with some holes (if just due to an `.item()`), and e.g. we won't notice if there should be a `ComplexWarning` due to a nested cast, but that seems fine...

The main questions here may well be around `__cuda_alignment__` or so.

tests/cupy_tests/core_tests/test_structured.py
```diff
@@ -46,11 +48,114 @@ def __cuda_array_interface__(self):
         # is the case, this always passes.
         # with pytest.raises(ValueError):
         #     func(ByteSwapped(base))
```
leofang: Q: Is this comment invalid now?

tests/cupy_tests/core_tests/test_structured.py
```diff
@@ -46,11 +48,114 @@ def __cuda_array_interface__(self):
...
-                func(ByteSwapped(base))
-        else:
-            assert func(ByteSwapped(base)).dtype == f"V{base.dtype.itemsize}"
+        assert func(ByteSwapped(base)).dtype == f"V{base.dtype.itemsize}"
+
+
+class TestFieldCasting:
+    @testing.numpy_cupy_array_equal()
+    def test_casting_simple(self, xp):
+        a = xp.array([(1, 2.), (3, 4.)], dtype="i8,f8")
```
leofang, line 63: Q: if we use `"i4,f8"` instead, will we get to test the padding behavior?

tests/cupy_tests/core_tests/test_structured.py
```diff
@@ -118,3 +243,76 @@ def test_slicing(self, xp, sl):
...
         return (a[sl]["f0"], a[sl]["f1"])
+
+
+class TestKernelStructAccess:
+    def test_elementwise_kernel(self):
+        # When the data can be defined as a normal struct, then
+        # we allow access via `.data`.
+        # TODO/NOTE(seberg): Is this even desired?!
+        kernel = cupy.ElementwiseKernel(
+            'T in', 'T out', 'out.data.f1 = -in.data.f0;',
```
leofang: Q: How do we determine the field names on the device side?

cupyx/signal/_filtering/_filtering.py
```diff
@@ -293,7 +293,7 @@ def freq_shift(x, freq, fs):
 #include <cooperative_groups.h>
 #include <cooperative_groups/reduce.h>
 #include <cupy/complex.cuh>
-#include <cupy/float16.cuh>  // TODO(seberg): Add this via type_headers?
+#include <cupy/float16.cuh>  // TODO(seberg): Add this via type_decls?
```
leofang, line 296: > Add this via type_decls?

If we are ready to rebuild the world (since changing comments globally would trigger it anyway..) I suppose the answer is yes. It might also help reduce the compile time of these modules. WDYT?

cupy/_core/_fusion_trace.pyx
```diff
@@ -115,10 +116,13 @@ def _guess_routine(func, args, dtype):
...
             obj = core.ndarray((0,), x.dtype)
         dummy_args.append(obj)
+        dtypes.append(x.dtype)
 
     op = func._ops.guess_routine(
         func.name, func._routine_cache, dummy_args, dtype, None)
-    return op.get_in_dtypes(), op.get_out_dtypes(), op.routine
+
+    in_out_dtypes = op.resolve_dtypes(dtypes[:func.nin], dtypes[func.nin:])
+    return in_out_dtypes[0], in_out_dtypes[1], op.routine
```
leofang: nit:
```suggestion
    iin_dtypes, out_dtypes = op.resolve_dtypes(dtypes[:func.nin], dtypes[func.nin:])
    return in_dtypes, out_dtypes, op.routine
```

cupy/_core/_routines_logic.pyx
```diff
@@ -53,7 +56,28 @@ cdef _any = create_reduction_func(
     'false', '')
 
 
-cpdef create_comparison(name, op, doc='', no_complex_dtype=True):
+def struct_compare_resolution(op, in_dtypes, out_dtypes):
```
leofang, line 59: Q: I assume there exists C APIs that we can switch to in the future, or does staying in pure Python actually not matter here?

leofang (commented):
I am half-way done!

tests/cupy_tests/core_tests/test_structured.py
seberg: I am not auto-padding for `xp.array()` right now, so it would just fail on this test.  There are some tests below (mainly using complex types, because they are always float aligned in NumPy/C and itemsize aligned in CUDA) that test the padding paths.

tests/cupy_tests/core_tests/test_structured.py
seberg: They come directly from the dtype field names, *if* the dtype can be represented by a C-struct, then I generate the `typedef struct {}` and use it for `.data` to make them available.
We could allow something like `out.get<field>` but it seems C++20 or so and we don't actually _need_ field names for our kernels, it is only interesting for custom kernels.

Commenting also because we should decide whether we want this at all.  We could just delete the `.data` attribute (as we don't use it).
So this is a convenience feature for custom kernels avoiding careful dtype checking + reinterpret casts in the kernel.

cupyx/signal/_filtering/_filtering.py
seberg: Changing this here will add a some `cupyx` churn. If you prefer sure, but it'll only cause rebuilding of the cupyx `RawKernel`s and is rather distinct.  But yeah, would be nice to finish that for 14.1.

(It is probably pretty straight-forward, although I am confused by the direct use of `half` and `__half` (ROCm) rather than `float16`.)

cupy/_core/_routines_logic.pyx
seberg: Yeah, we could make it a bit more light-weight with C, the reason for the Python `def` is passing it into `create_ufunc`.  Right now it may matter a bit, since this isn't cached.
But since we cache kernels aggressively, this part may also be cache-able in principle (I am not jumping at it right now, seeming structs as a less relevant for overheads, but I am not actually sure about that part).

tests/cupy_tests/core_tests/test_structured.py
leofang: > We could just delete the .data attribute (as we don't use it).

So instead of 
```c++
'out.data.f1 = -in.data.f0;'
```
we write
```c+++
'out.f1 = -in.f0;'
```
? If so, it sounds reasonable to me. 

This C++ feature seems very internal-ish and I expect only the team (+ contributors) would be using it within the codebase for now, so I am not too worried about it. I am aware some users reaching into `CArray` stuff, but it's really undocumented and offers no guarantee.

tests/cupy_tests/core_tests/test_structured.py
leofang: Ah. What if it's `np.dtype([('f1', np.int32), ('f2', np.float64)], align=True)` instead? Would it also fail?

cupy/_core/_scalar.pyx
```diff
@@ -37,21 +43,234 @@ cdef object _numpy_float64 = numpy.dtype(numpy.float64)
...
+
+    # Formatting the is unfortunately not as simple as `sorted()` as it can
+    # be nested, etc.
+    cdef dict declarations = {}
+    # Recursively flatten the type declarations into a dict
+    _flatten_type_decls(type_decls, declarations)
+    # Sort the dictionary by it's keys (to achieve a stable order)
+    declarations = dict(sorted(declarations.items()))
+
+    ts = graphlib.TopologicalSorter(declarations)
```
leofang, line 82: 👍

cupy/_core/_scalar.pyx
```diff
@@ -37,21 +43,234 @@ cdef object _numpy_float64 = numpy.dtype(numpy.float64)
...
+            # Alignment is inherited from the field dtype
+            return get_cuda_alignment(dtype.base)
+
+    if _alignment_kernel is None:
+        _alignment_kernel = cupy.ElementwiseKernel(
+            "T in", "int64 out",
+            "using in_t = decltype(in); out = alignof(in_t);",
+        )
+
+    alignment = int(_alignment_kernel(cupy.empty((), dtype=dtype))[()])
```
leofang, line 187: Applying an inline comment we use throughout the codebase... though I wonder if we can avoid it... For example, this would break CUDA graph and also hurt perf.
```suggestion
    # synchronize!
    alignment = int(_alignment_kernel(cupy.empty((), dtype=dtype))[()])
```

cupy/_core/_dtype.pyx
```diff
@@ -204,3 +204,132 @@ cdef dict dtype_format = {
...
+        dtype_info["itemsize"] = itemsize
+
+    # Create a new dtype enforcing the newly computed offsets.
+    return cnp.dtype(dtype_info, align=True, **metadata), final_alignment
+
+
+def make_aligned_dtype(
+        dtype, *, int alignment=-1, bint recurse=True):
+    """Create a new structured dtype from a NumPy dtype or dtype-like with
+    sufficient algnment for GPU use.
```
leofang:
```suggestion
    sufficient alignment for GPU use.
```

cupy/_core/_dtype.pyx
```diff
@@ -204,3 +204,132 @@ cdef dict dtype_format = {
...
+                inferred one an error will be raised.
+                When it is larger, CuPy will attach this alignment as
+                metadata to the structured dtype. This is used for structured
+                dtypes in the kernel and when nesting.
+
+                Note that metadata may be lost in many operations.
+
+        recurse: Whether to recurse into nested structures, defaults to True.
+            When ``False``, the nested struct is assumed to be already
+            sufficientlyaligned.
```
leofang:
```suggestion
            sufficiently aligned.
```

cupy/_core/_dtype.pyx
```diff
@@ -204,3 +204,132 @@ cdef dict dtype_format = {
...
+
+    Returns:
+        cupy.ndarray: A view of the array with reduced dimensions.
```
leofang:
```suggestion
        numpy.dtype: A dtype object with the desired alignment.
```

cupy/_core/_dtype.pyx
```diff
@@ -204,3 +204,132 @@ cdef dict dtype_format = {
...
+        offsets = []
+        subdtypes = []
+
+        for name, (subdtype, offset, *_) in (<object>descr).fields.items():
+            # The fields tupe can contain a 4th title, we ignore it.
+            # (A title is an alternative field name...)
+            if offset < min_offset:
+                raise ValueError(
+                    "make_aligned_dtype() only supports well behaved "
+                    "in order fields as it ignores field offsets).")
```
leofang:
```suggestion
                    "in-order fields as it ignores field offsets.")
```

cupy/_core/include/cupy/structview.cuh
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  // Construct from "scalar" by broadcasting.
+  template<typename T>
+  explicit StructView(const T& value) : StructType{} {
+    assign_broadcast<0>(value);
+  }
+
+  StructView(const StructType& s) : StructType(s) {}
+  StructView(StructType&& s) : StructType(std::move(s)) {}
```
leofang: This seems odd to me... shouldn't `data` be mentioned in the initializer?

cupy/_core/_scalar.pyx
```diff
@@ -37,21 +43,234 @@ cdef object _numpy_float64 = numpy.dtype(numpy.float64)
...
+        descr = <cnp.dtype>dtype
+        if descr.type_num == cnp.NPY_VOID:
+            if cnp.PyDataType_HASFIELDS(descr):
+                # NOTE: Caching this may not be trivial if we use metadata.
+                name, *_ = _build_struct_typename(dtype, type_decls)
+                return name
+            elif not cnp.PyDataType_HASSUBARRAY(descr) and descr.itemsize > 0:
+                # Unstructured void is just a blob bytes.
+                if type_decls is not None:
+                    type_decls.add('#include "cupy/unstructued_void.cuh"')
```
leofang:
```suggestion
                    type_decls.add('#include "cupy/unstructured_void.cuh"')
```

cupy/_core/include/cupy/unstructured_void.cuh
leofang, line 1: filename: `unstructued_void.cuh` -> `unstructured_void.cuh`

cupy/_core/include/cupy/unstructued_void.cuh
```diff
@@ -0,0 +1,46 @@
...
+    const char* other_bytes = reinterpret_cast<const char*>(&other);
+    memcpy(data, other_bytes, min_length);
+    if (Size > min_length) {
+      memset(data + min_length, '\0', Size - min_length);
+    }
+    return *this;
+  }
+};
+
+} // namespace scupy
```
leofang:
```suggestion
} // namespace cupy
```

cupy/_core/_scalar.pyx
```diff
@@ -37,21 +43,234 @@ cdef object _numpy_float64 = numpy.dtype(numpy.float64)
...
+            "T in", "int64 out",
+            "using in_t = decltype(in); out = alignof(in_t);",
+        )
+
+    alignment = int(_alignment_kernel(cupy.empty((), dtype=dtype))[()])
+    _cuda_alignments[dtype] = alignment
+    return alignment
+
+
+def _build_struct_typename(dtype, type_decls):
```
leofang, line 192: Q: Should we use functool.cache or the internal _utils.cache here?

cupy/_core/_scalar.pyx
```diff
@@ -37,21 +43,234 @@ cdef object _numpy_float64 = numpy.dtype(numpy.float64)
...
+                subdtype, subtype_decls)
+        else:
+            subalignment = get_cuda_alignment(subdtype)
+            subname = struct_name = get_typename(subdtype, subtype_decls)
+            if subdtype.metadata:
+                # If manually overridden, use that alignment:
+                subalignment = subdtype.metadata.get(
+                    "__cuda_alignment__", subalignment)
+
+        assert (subalignment - 1) & subalignment == 0
```
leofang: nit:
```suggestion
        assert ((subalignment - 1) & subalignment) == 0
```

tests/cupy_tests/core_tests/test_structured.py
seberg: Ah, right now it is `out.data.f1`, I could maybe subclass the structured dtype to achieve writing `out.f1`.  I.e. `.data` is the normal struct. I had a subclassing solution initially, but then thought if we don't need it much, subclassing may add a bit much complexity.

The alternative would be `out.get<f1>() = -in.get<f0>()` if C++20 is available.

> would be using it within the codebase for now, so I am not too worried about it.

So maybe it is just a bit "experimental" which seems fine. But not sure if we are likely to need it internally (_maybe_ in `cupyx`).

cupy/_core/_scalar.pyx
seberg: Unfortunately, caching needs to use the identity, so I didn't add it yet. But yeah, maybe I should (it needs to be a custom caching, unfortunately, which may mean cache misses are pretty common), since this is going to be slowish.

cupy/_core/_scalar.pyx
seberg: I liked making it generic but, while a bit theoretical, that is a downside I hadn't thought of; making it generic here is really not all that necessary.

So I'm tempted to just hard-code the values for now?  (I can't see a neat way to get it without the kernel launch right now, at least not simple enough considering how OK hard-coding is for now.)

cupy/_core/include/cupy/structview.cuh
seberg: Hmmm, I probably asked the agent to fill this in.  Let me try if `StructView(const StructType& s) default` didn't work anyway.
(IIRC, I need to add this to avoid confusion with the templated version but using `default` likely works)

tests/cupy_tests/core_tests/test_structured.py
seberg: In principle yes, float64 CPU alignment can be 4 bytes.  That said, it may be that it isn't on any hardware we support in practice (I am not sure it is on any 64bit platforms).

tests/cupy_tests/core_tests/test_structured.py
seberg: Yeah, I hadn't fixed that yet here. Let's see, it might not be much more than a 2 line change to make it work well enough.

cupy/_core/include/cupy/structview.cuh
seberg: Actually, I just deleted this and also the `as_struct` below. If `.data` is exposed then I think there is little reason to have it.
(Yeah, the constructor _might_ be useful, but overall, I doubt it.)

EDIT: For clarity, this was a left-over from when I thought I just subclass, rather than just wrapping a `data` field.

tests/cupy_tests/core_tests/test_structured.py
seberg: OK, I added this.  It means that CAI now strictly checks for supported dtypes.  But, since we support unstructured void dtypes now, that shouldn't be much of a limitation.

seberg: OK, I added support for this, but with slightly extra logic to work with the necessary padding more reliably (NumPy doesn't do this well, but also doesn't pad by default, I need to think about what API to add in NumPy to improve this).

The `asarray()` path still doesn't raise, because CAI doesn't strictly check the dtype compatibility right now.  (tried enabling it, but I forgot about `datetime`)

cupy/_core/_scalar.pyx
leofang: What do you have in mind specifically about hard-coding?

seberg: Right now, `alignment == itemsize` for all dtypes we support, even if we add string support as an explicit branch, for the majority of types we can just hard-code that.

(Or use the largest power of two that is smaller than the itemsize as a guess for unknown dtypes, doesn't work for strings, but works for all other current dtypes in practice.)

cupy/_core/include/cupy/structview.cuh
leofang: I mean, having copy/move constructors defined is correct (rule of three/five). What was confusing to me is the initializer  `StructType(s)` and `StructType(std::move(s))`, which did not seem to be correct syntax or convention.

tests/cupy_tests/core_tests/test_structured.py
leofang: yeah let's keep it c++17 compatible for now

leofang:
(Let me merge with main to catch up with recent CI cache update)

asi1024 (dismissed):

leofang:
@seberg let's resolve the conflicts and re-test!

seberg:
Rebased and cleaned up slightly.  I did two changes:
1. For now only accept structured dtypes in `make_aligned_dtype()`, that isn't necessary I think, but I doubt that we need it for other dtypes (it would attach `__cuda_alignment__`).
2. I removed the `.data` being an actual struct in some cases.  So field access is now solely by field number internally.
   The pattern is now `StructView<typename type, typename... Fields>`, but we always use `StructView<cupy::raw_structview_storage<itemsize, alignment>>`
   * This way, I think we can restore the old path passing a struct (with explicit padding).

Otherwise a few fixes based on what codex noticed during a review.

/test mini

seberg:
Erik had given me some claude red-team tasks.  I applied most of them (the largest diff being that it suggested to re-organize the header to allow more than 250 fields; not really vital but sounds good).
There were also two issues with the "expose `.data` as an actual C-struct" with field names part.  But since I removed that, those went away.

A few of the suggestion seemed wrong, or way to niche to do something about (biggest niche one is support for `dtype([])`, I also leaned to not early reject `"i4,>i4"` although also happy to do -- it's super niche either way).

/test mini

cupy/_core/core.pyx
```diff
@@ -3190,6 +3194,36 @@ cpdef _ndarray_base asfortranarray(_ndarray_base a, dtype=None):
...
 
+cdef inline get_cai_dtype(dict desc):
+    dtype = numpy.dtype(desc['typestr'])
+    if <cnp.dtype>(dtype).num != cnp.NPY_VOID:
+        return dtype
+
+    # extract dtype from descr, this is slightly "smarter" than some
+    # NumPy versions (at least 2.4), but much like `np.load` logic.
+    # (For CuPy it is much more likely to use aligned and thus padded dtypes.)
+    descr = desc['descr']
```
eriknw: Very minor edge case with a small fix (if desired). `descr` is optional according to CAI v3, so for compliance we could probably do this:
```python
    descr = desc.get('descr')
    if descr is None:
        return dtype
```
In practice, the most likely case where this would be encountered is from third-party CAI producers that emit e.g. `|V<N>`, so probably not super likely. Also, raising a `KeyError` instead of returning a `V<N>` array _may_ be more helpful to nudge third-party providers to provide an informative `descr`. But if this is ever encountered, there may be pushback that CuPy should be more compliant. Either way, I don't think this is particularly important, but it would be easy to change.

eriknw (approved):
LGTM, primarily from a user perspective.

Thanks for applying the appropriate updates from my Claude's suggestions. I agree with your judgement about which to ignore.

The one other comment I left is more about technical compliance with CAI v3 that I think we should be aware of (and these comments leave a paper trail for). Probably fine as is though, so I'm approving the PR :+1:

cupy/_core/core.pyx
seberg: Hmmmm, my tendency is that it is optional, but only if the result isn't void :).  The problem is that right now we hit this for bfloat16 as well :(.

I think we should just hack bfloat16 for cupy FWIW, I.e. set `typestr` to `bfloat16` for it and be done with it...
(I want to fix this for NumPy, but I don't think I'll manage for 2.5 :()

(I'll have a brief check tomorrow if there is a regression here, in that we previously just returned the void version -- which then allows to do a manual `.view()` in theory, which... who knows if someone might be using such tricks)

seberg: Yeah, you (or claude) are right.  Better to just keep the old behavior and return the unstructured void here.  If this was from scratch I might prefer the error but it may be a regression (and honestly, that may not even be super unlikely).

seberg:
This causes a full rebuild, so unfortunately needs at least one more iteration.

/test mini

seberg:
(Was fine before except one test not finishing yet, so skipping for the doc change.)

/test skip

seberg:
Let me put this in then, thanks for the reviews!  I think `make_aligned_dtype` is fine, the least sure I am is about the `__cuda_alignment__` metadata, but it does make nesting work well and doesn't seem like a bad choice.

The doc addition is rather trivial and renders fine but we can follow-up of course.

seberg:
Ah, silly me, I can't actually merge, need someone to do a final approve as the one got stale :).

cupy/_core/include/cupy/unstructured_void.cuh
```diff
@@ -0,0 +1,48 @@
...
+
+#include "cupy/carray.cuh"
+
+
+namespace cupy {
+
+template<size_t Size>
+class UnstructuredVoid {
+public:
+  char data[Size];
```
leofang, line 12: `UnstructuredVoid<Size>` inherits the alignment of `char` (i.e. 1). That means `CArray<UnstructuredVoid<8>>` will also get alignment 1, so device accesses to multi-byte voids won't be naturally aligned. Consider adding `alignas(...)` — e.g. the largest power-of-two ≤ `min(Size, 16)` — to match the alignment CUDA would give an equivalent-sized trivial type.

cupy/_core/_scalar.pyx
```diff
@@ -152,13 +358,24 @@ cdef class CScalar(CPointer):
...
+
     cdef _store_c_value(self):
-        # If we ever support dtypes larger than this (e.g. strings)
-        # we will have to introduce a conditional allocation here and
-        # should memset memory to NULL (must if dtype NEEDS_INIT).
-        assert self.descr.itemsize < sizeof(self._data)
-        # make sure ptr points to _data.
-        self.ptr = <intptr_t><void *>(self._data)
+        # No current support for dtypes that may require initialization
+        assert not self.descr.flags & 0x08
```
leofang: Nit: `0x08` is a magic number. Using a named constant (e.g. `NPY_NEEDS_INIT` or at minimum a comment spelling out which flag this is) would make this more readable and resilient to future NumPy flag changes.

cupy/_core/include/cupy/structview.cuh
```diff
@@ -0,0 +1,198 @@
...
+
+public:
+  using storage_type = StorageType;
+  static constexpr size_t size = sizeof(storage_type);
+  static constexpr size_t alignment = alignof(storage_type);
+  static constexpr size_t field_count = sizeof...(Fields);
+
+  __device__ StructView() : data_{} {
+    init();
+  }
```
leofang: `data_{}` value-initializes (zero-fills) the storage, then `init()` default-constructs each field into it. For trivial field types this is a redundant double-write. Would `data_{}` alone suffice, or does `init()` carry additional semantics beyond zero-init?

leofang (commented):
A few observations from reading through the PR.

cupy/_core/_scalar.pyx
seberg: suggestion already applied, but forgot to resolve.

cupy/_core/include/cupy/structview.cuh
seberg: Ah good find, yeah, we could only init the fields but zeroing everything is the easier and must be correct for any field in practice.

(changing to the default constructor removing the `init()`)

cupy/_core/include/cupy/unstructured_void.cuh
seberg: It would be nice to use a higher alignment when possible, but in principle there is nothing stopping you from having a `?,V4` dtype or so, where the `V4` is random data that has no alignment guarantees.
Maybe we'll have to either allow users to override it explicitly, or use actual runtime alignment for the kernel building (I like the latter if it works).

cupy/_core/_scalar.pyx
seberg: Changed to use the actual enum (although accessed via `_scalar.NPY_...` in `_dtype.pyx`.

leofang (approved):

Issue #9559, opened 2025-12-31
Make Cython compilation warning-free


leofang:
An increasing number of Cython warnings are raised due to Cython 3.x. We should identify the warning classes and address them systematically group by group.

ManuCorrea:
Hello @leofang! I am willing to tackle this contribution since I thought might be a good starting point to me to contribute.

I started with the file _runtime_typedef.pxi:
https://github.com/cupy/cupy/blob/main/cupy_backends/cuda/api/_runtime_typedef.pxi

Is that a good strategy? Would you recommend starting somewhere else? I have been struggling with it just managed to migrate a single #IF deprecation warning.

I am using the following strategy for the #IF warnings migration, would that be a good solution?
```
cdef extern from *:
    """
    #if CUPY_CUDA_VERSION > 0
        typedef struct cudaMemLocation MemLocation;
        typedef struct cudaMemPoolProps MemPoolProps;
    #else
        /* Dummy definitions for HIP/RTD to prevent compilation errors */
        typedef struct { int type; int id; } MemLocation;
        typedef struct { int allocType; int handleTypes; MemLocation location; } MemPoolProps;
    #endif

...
    """
```

The warnings I found of Cython are:
Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
The 'IF' statement is deprecated and will be removed in a future Cython version. 

The noexcept should be straightforward, can I assume that all of them are effectively not raising an exception to Python or should require a double check?

seberg:
Yeah, that is the right pattern for the `#IF` migration and cupy uses it in parts. If it bloats things a lot, I am slightly unsure if I would prioritize it. In the NCCL code those defines are in actual header files, but dunno if that matters much.

As to the `noexcept`, this would be very cool to address for sure!  I suspect you can find larger parts where just mass adding it is good.  Unfortunately, I started looking at it a bit at some point and there were also quite a few places where it seemed subtly wrong to assume that no error can happen.
So, I think those are *very* worthwhile to start addressing but it might make sense to try and focus on the obvious cases where just adding `noexcept` works.
(Happy to also add `except?` or so otherwise though, in some cases that might warrant some thought I guess.)

ManuCorrea:
[...]

PR #9654, opened 2026-01-29
Cython Compilation Warnings of implicit noexcept

ManuCorrea:
### Summary

This PR addresses part of the Cython deprecation warnings reported in #9559, specifically those related to implicit `noexcept` declarations

This PR focuses solely on resolving deprecation warnings and does not introduce functional or behavioral changes.

### What was changed

- Fixed warnings of the form:
  - `"Implicit noexcept declaration is deprecated"`
  - `"noexcept clause is ignored for function returning Python object"`
- Updated function signatures to explicitly declare exception behavior:
  - Used `noexcept` for pure C / CUDA wrapper functions that cannot raise Python exceptions.
  - Used `except *` for functions whose execution path may invoke Python APIs and therefore may raise exceptions.
- Adjusted function signatures to place `nogil` at the end, as recommended by Cython, removing minor related warnings.

### Build
```
  Python 3.11.14
  Modules:
    cuda      : Yes (version 12090)
    cusolver  : Yes
    nccl      : No
      -> Include files not found: ['nccl.h']
      -> Check your CFLAGS environment variable.
    nvtx      : Yes
    cutensor  : No
      -> Include files not found: ['cutensor.h']
      -> Check your CFLAGS environment variable.
    cub       : Yes (version 300102)
    jitify    : Yes (version 1a0ca0e)
    random    : Yes
    thrust    : Yes (version 300102)
    cusparselt: No
      -> Include files not found: ['cusparseLt.h']
      -> Check your CFLAGS environment variable.
    dlpack    : Yes

Compile-time constants: {
      "CUPY_USE_CUDA_PYTHON": false,
      "CUPY_CUFFT_STATIC": false,
      "CUPY_CYTHON_VERSION": "3.1.8",
      "CUPY_CUDA_VERSION": 12090,
      "CUPY_HIP_VERSION": 0
  }
```

### Testing
I ran `python -m pytest tests/install_tests`  fully
And partially `python -m pytest` to make sure CuPy ran, should I run the whole test suit or with the tests of the installation of modules is enough?

Verified removal of the reported deprecation warnings in the full log file of build.

### DIFF
--- a/cupy/_core/dlpack.pxd
+++ b/cupy/_core/dlpack.pxd
@@ -40,7 +40,7 @@ cdef extern from './include/cupy/_dlpack/dlpack.h' nogil:
         int32_t device_id
 
     # Enable using `except DLDevice_err` (there is no direct way)
-    const DLDevice DLDevice_err "((DLDevice){(DLDeviceType)-1, -1})"
+    const DLDevice DLDevice_err "(DLDevice{(DLDeviceType)-1, -1})"
 
     ctypedef enum DLDataTypeCode:
         kDLInt


cupy/_core/_kernel.pyx
```diff
@@ -345,7 +346,8 @@ cdef shape_t _reduce_dims(list args, tuple params, const shape_t& shape):
     return _reduced_view_core(args, params, shape)
 
 
-cdef shape_t _reduced_view_core(list args, tuple params, const shape_t& shape):
+cdef shape_t _reduced_view_core(
+        list args, tuple params, const shape_t& shape) noexcept:
```
seberg: Technically, this isn't true, since this function _can_ run out of memory in the vectors or new objects.

My suggestion is to omit here and then refactor it to modify `shape` in-place, since I think that is always acceptable (that way we could make it a `bint ... except 0` function).
We could also add `except *`, I am not exactly sure how hot this path is and how much overhead `PyErr_Occurred()` adds, it may be nice to avoid, but it may be more reasonable to just add `except *` and maybe look at trying to remove the number of `except *` in the future.

(Omit, just because this is huge either way, easier to split out any non-trivial changes.)

cupy/_core/_kernel.pyx
```diff
@@ -1073,7 +1075,7 @@ cdef dict _mst_unsigned_to_signed = {
                  for i in "BHILQ"]}
 
 
-cdef inline int _get_kind_score(type kind):
+cdef inline int _get_kind_score(type kind) except *:
```
seberg: This one actually cannot (reasonably) fail.  But it returns an `int` so `except *` is unnecessary, it can be `except -1`.

cupy/_core/_reduction.pyx
```diff
@@ -475,7 +476,7 @@ cdef class _AbstractReductionKernel:
             self, out_block_num, block_size, block_stride,
             in_args, out_args, in_shape, out_shape, type_map,
             map_expr, reduce_expr, post_map_expr, reduce_type,
-            stream, params):
+            stream, params) noexcept:
```
seberg: There is way too much going on here for it to be noexcept.  Now, it might be that due to it being actually `inline` it just doesn't matter what it says here.

cupy/_core/core.pyx
```diff
@@ -2446,14 +2446,14 @@ cdef inline str _translate_cucomplex_to_thrust(str source):
...
-cpdef bint use_default_std(tuple options):
+cpdef bint use_default_std(tuple options) noexcept:
     cdef str opt
     for opt in options:
         if ('-std=' in opt) or ('--std=' in opt):
             return False
     return True
 
-cpdef void warn_on_unsupported_std(tuple options):
+cpdef void warn_on_unsupported_std(tuple options) noexcept:
```
seberg: Also here, warnings are *by definition* not noexcept.  Either use a `bint` pattern or just `except*` (since this is clearly not on a hot path).

cupy/_core/dlpack.pyx
```diff
@@ -80,7 +80,7 @@ cdef uint8_t get_dlpack_dtype_code(dtype) except? 255:
         raise BufferError('dtype is not supported for dlpack export')
 
 
-cdef DLDevice get_dlpack_device(_ndarray_base array):
+cdef DLDevice get_dlpack_device(_ndarray_base array) noexcept:
```
seberg: I suspect this function can at least raise earlier asynchronous CUDA errors. Doesn't `except -1` work?

cupy/cuda/memory.pxd
```diff
@@ -64,12 +64,13 @@ cdef class MemoryPool:
...
     cpdef free_all_free(self)
-    cpdef size_t n_free_blocks(self)
-    cpdef size_t used_bytes(self)
-    cpdef size_t free_bytes(self)
-    cpdef size_t total_bytes(self)
+    # they call device_pool so they may raise error.
+    cpdef size_t n_free_blocks(self) except *
+    cpdef size_t used_bytes(self) except *
+    cpdef size_t free_bytes(self) except *
+    cpdef size_t total_bytes(self) except *
```
seberg: All of these should clearly use `except -1`, IMO (yeah, it's an unsigned int, but the maximum value should be OK, or if you don't like that fact, make it `except -1?`

cupy/_core/_kernel.pyx
ManuCorrea: The use of issubclass() can raise a Python error, that is why I add the except *
https://docs.python.org/3/library/functions.html#issubclass

It may be hard to trigger but from what I understood as long as something could raise a Python error it should include the except *

cupy/_core/core.pyx
ManuCorrea: But those don't raise a Python Exception unless is configured to* they would just print the message right?

* https://docs.python.org/3/library/warnings.html#warning-filter

ManuCorrea:
From the Cython docs:
´´´
cdef void spam() except *:
    ...
´´´

This form causes Cython to generate a call to [PyErr_Occurred()](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Occurred) after every call to spam, regardless of what value it returns. Calling spam() is roughly translated to the following C code:

----

I am wondering, should we take care or benchmark once we are done with this PR? We could may be including some overhead since PyErr_Occurred() is call every time each function finish.

cupy/_core/dlpack.pyx
ManuCorrea: I got cupy/_core/dlpack.pxd:73:61: Cannot assign type 'long' to 'DLDevice' since its a struct, the return type must match the error type, changing to except *

seberg:
Yeah, I am OK with going with `except *` when there is no direct alternative and considering it a follow up.  A brief benchmark for some of these would be nice (only relatively hot ones), but overall, I don't expect a real difference in any relevant real-world code examples, so while nicer not to, yeah, I think we can go with `except *` here.

cupy/_core/_reduction.pyx
ManuCorrea: I agree but are we willing to include some overhead on the launch of kernels? I think it could impact performance (I commented about on the main thread of the PR, maybe we need to consider benchmarking once all is done?).

Let me know what you think and will change accordingly.

cupy/_core/_kernel.pyx
seberg: Yeah, but not here and in Cython, it can only raise for abstract classes.  Unless the NumPy module was monkeypatched anyway.  But sure, use `except -1`, it's clean.

cupy/_core/_reduction.pyx
seberg: You can just make it `bint ... except 0` and ignore the return value as I said in my earlier post, I believe.

(This is OK when it's clearly simple private functions, don't do it for public ones.)

ManuCorrea: Just saw your comment on the PR thread, changing it to except *

cupy/_core/core.pyx
seberg: yes, but:
1. they call complex Python code, which should always be assumed to be able to raise.
2. It's the **users** choice, not our choice how warnings are configured (in particular it's common and good practice to raise warnings in tests during development).

cupy/_core/dlpack.pyx
seberg: Ah, yeah. I think one can define a `const DLDevice dldevice_err = (-1, -1)` and use that, but doubt that complexity is worthwhile.

cupy/_core/_kernel.pyx
leofang: How about marking it `except +*` (propagating both C++ and Python exceptions)?

leofang:
It is nerve-wrecking that we have `noexcept` added to so many places. Are we sure these are guaranteed to not raise? If not, any `noexcept` would make exceptions unhandled and terminate the program.

ManuCorrea:
I understand the worry about having `noexcept` added in so many places. But it is not changing any behavior from before since it was implicit when there was no declaration. Now it is a great chance to review all the functions that were assumed as noexcept so we can improve such declarations.

If it raises an exception and is labeled as `noexcept` according to Cython docs: "If a noexcept function does finish with an exception then it will print a warning message but not allow the exception to propagate further." So it makes me wonder, can we catch those warnings in the unit testing process?

I will go over all the changes to review those `noexcept` since now I have a better understanding than when I started and also got your feedback.

cupy_backends/cuda/api/driver.pyx
```diff
@@ -320,7 +320,7 @@ cpdef funcSetAttribute(intptr_t f, int attribute, int value):
 ###############################################################################
 
 cpdef int occupancyMaxActiveBlocksPerMultiprocessor(
-        intptr_t func, int blockSize, size_t dynamicSMemSize):
+        intptr_t func, int blockSize, size_t dynamicSMemSize) except *:
```
seberg: I would really like you to always prefer except `-1` or so when it makes sense.  If there is any doubt, just add the `?`.

cupy_backends/cuda/libs/cublas.pyx
```diff
@@ -1441,9 +1444,10 @@ cpdef int cgetrsBatched(intptr_t handle, int trans, int n, int nrhs,
             <cuComplex**>Barray, ldb, <int*>info, batchSize)
     check_status(status)
 
-cpdef int zgetrsBatched(intptr_t handle, int trans, int n, int nrhs,
-                        size_t Aarray, int lda, size_t devIpiv,
-                        size_t Barray, int ldb, size_t info, int batchSize):
+cpdef int zgetrsBatched(
+        intptr_t handle, int trans, int n, int nrhs,
+        size_t Aarray, int lda, size_t devIpiv,
+        size_t Barray, int ldb, size_t info, int batchSize) except *:
```
seberg: Why not use `except -1` if these functions already have an unused `int` return?

cupy/_core/_memory_range.pyx
ManuCorrea, line 32: Here there is the following warning:
```
  warning: cupy/_core/_memory_range.pyx:33:66: local variable 'left' referenced before assignment
  warning: cupy/_core/_memory_range.pyx:33:72: local variable 'right' referenced before assignment
```
They get a value assigned in get_range:
```
cdef get_range(
        Py_ssize_t itemsize, shape_t& shape, strides_t & strides,
        Py_ssize_t& out_left, Py_ssize_t& out_right):
    """Discover the byte range (out_left, out_right] (without ptr offset).
    """
    cdef Py_ssize_t tmp, i
    out_left = 0
    out_right = itemsize
```
Should I initialize the variables to avoid the warning or other warnings are out of scope for this PR?

cupy/_core/internal.pyx
```diff
@@ -374,7 +375,7 @@ cdef _broadcast_core(list arrays, shape_t& shape):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cpdef bint _contig_axes(tuple axes):
+cpdef bint _contig_axes(tuple axes) noexcept:
```
ManuCorrea: So in the following code as example. We have the indexing `axes[n], axes[n - 1] ` which could raise an IndexError, but checking the code which is well written that should never happen. Would you rather I label it as `except *` even if the code should not raise an error? @seberg @leofang 

I also have a doubt, can receive other types than tuples? I think another type error may be raised on runtime so **I am assuming it's always receiving the correct type.** That makes it safe to call len() for example. Am I right or I should assume any type could be received because then there are many scenarios that would be labeled with `noexcept` since len() is called in many functions.

cupy/cuda/memory.pyx
```diff
@@ -398,7 +398,7 @@ cdef class _Chunk:
         self.size = size
         self.stream_ident = stream_ident
 
-    cpdef intptr_t ptr(self):
+    cpdef intptr_t ptr(self) noexcept:
         return self.mem.ptr + self.offset
```
ManuCorrea, line 402: Could be possible to have an overflow here? If so should we handle it and change to except?

cupy/cuda/memory.pyx
```diff
@@ -2044,18 +2044,20 @@ cdef class MemoryAsyncPool:
...
+ctypedef void*(*malloc_func_type)(void*, size_t, int) noexcept
+ctypedef void(*free_func_type)(void*, void*, int) noexcept
 
 
 cdef intptr_t _call_malloc(
-        intptr_t param, intptr_t malloc_func, Py_ssize_t size, int device_id):
+        intptr_t param, intptr_t malloc_func, Py_ssize_t size,
+        int device_id) noexcept:
     return <intptr_t>(
         (<malloc_func_type>malloc_func)(<void*>param, size, device_id))
```
ManuCorrea, line 2055: I guess those call [cudaMalloc](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g37d37965bfb4803b6d4e59ff26856356) which can return 1 o 2 as error. Should use runtime.pyx `check_status` before returning it and adding the proper except?

Or is it using runtime.pyx malloc?

ManuCorrea:
In runtime.pyx
```
cpdef intptr_t malloc(size_t size) except? 0:
    cdef void* ptr
    with nogil:
        status = cudaMalloc(&ptr, size)
    check_status(status)
    return <intptr_t>ptr
```
It calls check_status which other functions that call it have except -1? I am a bit confused in this scenario, shouldn't it have an except -1 as well since check_status raises an error? The except 0 is for checking the NULL pointer if I understood well right?

cupy_backends/cuda/api/_runtime_softlink.pxi
```diff
@@ -8,7 +8,7 @@ Load CUDA Runtime shared library via SoftLink to allow probing CUDA
 version installed in users' environment.
 """
 
-ctypedef int (*F_cudaRuntimeGetVersion)(int* runtimeVersion) nogil
+ctypedef int (*F_cudaRuntimeGetVersion)(int* runtimeVersion) noexcept nogil
```
ManuCorrea, line 11: How could we check the cudaError in here? It's a direct link to the CUDA API right? Should we make a pyx function and apply the get_status?

cupy/_core/internal.pyx
seberg: The problem is only if someone passes nonsense, like `("1", "2")` which then fails later.

So yeah, this can fail, but not because of indexing.

This function has a `bint` return value, so it's trivial and always correct to add `except -1` (to be honest, I thought that was already the default for `bint`).

cupy_backends/cuda/libs/_cnvrtc.pxi
```diff
@@ -57,10 +57,12 @@ cdef F_nvrtcGetNumSupportedArchs nvrtcGetNumSupportedArchs
 ctypedef nvrtcResult (*F_nvrtcGetSupportedArchs)(int* supportedArchs) noexcept nogil  # NOQA
 cdef F_nvrtcGetSupportedArchs nvrtcGetSupportedArchs
 
-ctypedef nvrtcResult (*F_nvrtcGetLTOIRSize)(nvrtcProgram prog, size_t *ltoirSizeRet) nogil  # NOQA
+ctypedef nvrtcResult (*F_nvrtcGetLTOIRSize)(
+    nvrtcProgram prog, size_t *ltoirSizeRet) noexcept nogil  # NOQA
```
ManuCorrea, line 61: How should we check `nvrtcResult`? Seems a direct link to the API

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -3077,7 +3081,7 @@ cpdef int zgesvdaStridedBatched_bufferSize(
         intptr_t handle, int jobz, int rank, int m, int n, intptr_t d_A,
         int lda, long long int strideA, intptr_t d_S, long long int strideS,
         intptr_t d_U, int ldu, long long int strideU, intptr_t d_V, int ldv,
-        long long int strideV, int batchSize):
+        long long int strideV, int batchSize) noexcept:
```
seberg: This is clearly incorrect, since the function calls `check_status` which will set an error.

cupy/cuda/memory.pyx
seberg: No, this is C-code, it can't overflow (and pointer operations are always OK, if this overflows it would be due to the using signed integers, which is OK -- if maybe theoretically undefined).

cupy/_core/internal.pyx
```diff
@@ -84,7 +85,7 @@ cpdef inline bint vector_equal(
 @cython.profile(False)
 cdef void get_reduced_dims(
         shape_t& shape, strides_t& strides, Py_ssize_t itemsize,
-        shape_t& reduced_shape, strides_t& reduced_strides):
+        shape_t& reduced_shape, strides_t& reduced_strides) noexcept:
```
seberg: I think I commented on other things. If you are resizing vectors things can at least theoretically fail.

(FWIW, I might just make it `bint ... except 0` or `except -1` since this is very internal and not used a lot.)

cupy_backends/cuda/libs/_cnvrtc.pxi
seberg: you don't check it. It's clearly NVRTC C-API, not Cython.  `check_result()` will always be called on this.

cupy/cuda/memory.pyx
ManuCorrea: Sorry I think I did not explain myself properly. Yes the function as it is it doesn't raise an exception. What I meant is that if we should manually check that the pointer logic doesn't overflow so we always have safe addresses. And it case it overflows then we manually raising an error.

Something like:
```
if offset > 0 and (ptr + offset < ptr): # check if it wrapped around
    raise OverflowError("Pointer arithmetic overflow")
```

But after researching a bit just saw that it's quite unlikely to hit such situation and that by checking it can affect performance.

cupy_backends/cuda/libs/cublas.pyx
ManuCorrea: Because check_status could raise Python exception:
```
@cython.profile(False)
cpdef inline check_status(int status):
    if status != 0:
        raise CUBLASError(status)
```

cupy_backends/cuda/api/driver.pyx
ManuCorrea: It calls check_status which could do `raise CUDADriverError(status)`, should I proceed with `except -1`?

cupy/_core/internal.pyx
ManuCorrea: I changed several functions that had bint already from `noexcept` to `except -1`

cupy_backends/cuda/libs/cublas.pyx
seberg: yes, of course.  `except -1` means that Cython promises to return -1 if (and only if without the `?`) a Python exception occurred.

It is about how *this* function signals errors, not about what kind of error happened.

matusvalo:
Just going around - adding just few comments...

The ideal approach would be to remove `legacy_implicit_noexcept` argument
https://github.com/cupy/cupy/blob/d21cbea587afaec110695ab073b4e230e71eb1b9/install/cupy_builder/_command.py#L117
and make all function without `except` clause as `noexcept`. The resulting C code should be the same. The `legacy_implicit_noexcept` argument was introduced to help projects to migrate to Cython 3 and it would be great to remove it.

> he use of issubclass() can raise a Python error, that is why I add the except *

The `except *` is not mandatory for propagating exceptions in Cython. The same can be achieved also with `except -1`/`except? -1`(or other integer which makes sense). The main usage of `except *` is when there is no value returned that can be tested against (e.g. `void` return type, struct type or union type). Even easier will be when `legacy_implicit_noexcept` argument is removed. In this case Cython automatically injects exception handling to function call (equivalent to `except? -1`) not having `except` or `noexcept` clause.

seberg:
> ven easier will be when legacy_implicit_noexcept argument is removed. In this case Cython automatically injects exception handling to function call (equivalent to except? -1) not having except or noexcept clause.

Ahh, right, thanks for the note @matusvalo , there was a reason I thought `bint` always had an `except -1` :facepalm:.

We still need to review all functions, which is a bit hard. But I suppose we could actually limit this PR to just adding `noexcept` to functions that are clearly noexcept.
Then I need to review locally that the others should get exception handling (although missing one here or there isn't a huge deal).

(Of course for some of these an explicit `except -1` without the `?` makes sense, but I guess it doesn't matter much.)

ManuCorrea:
Thank you for the clarification @matusvalo.

Just one more question to ensure my implementation is correct. When a function calls check_status, I originally thought the caller must also use except *.

However, my current understanding is that the caller should use the except clause that matches its own return type. For example, if the caller returns an int, it should use except -1 (or except? -1), even though it calls an except * function like check_status inside. Is that correct? I want to make sure I'm not using except * unnecessarily when a more specific sentinel value is available.

```
@cython.profile(False)
cpdef inline check_status(int status):
    if status != 0:
        # to reset error status
        cudaGetLastError()
```

@seberg once this is confirmed I will do a final recheck and I think all should be ready.

While waiting for reviews, is there any other issue that you think I could tackle while? I am looking to improve my CUDA skills and this issue was perfect for getting an overview of the project.

Regarding the IF warning migration. I’ve done some initial tryouts but ran into compilation errors. If someone could point me toward the preferred pattern of a correct implementation I think I could handle it.

seberg:
Yes exactly. The body of the function doesn't matter (except to decide whether to decide if `noexcept` is OK because the body cannot possibly raise).

The only annoying thing is that we really should get rid of `except? -1` in the end, but I somewhat like to see a warning-free state to be sure we reviewed all functions :/.
But... I am OK with deferring that indefinitely (the important part is more to not keep cargo-culting it to new functions after we are done).

I think there are quite a few open issues, some of them may touch straight forward cuda kernel code (e.g. scipy ones).

ManuCorrea:
@seberg I implemented the feedback according the last comments, I hope all is good now!

cupy/_core/_kernel.pyx
seberg:
```suggestion
        list args, tuple params, const shape_t& shape) except *:
```

cupy/_core/_reduction.pyx
```diff
@@ -157,7 +157,7 @@ cpdef tuple _get_axis(object axis, Py_ssize_t ndim):
 
 cpdef shape_t _get_out_shape(
         const shape_t& shape, tuple reduce_axis, tuple out_axis,
-        bint keepdims):
+        bint keepdims) noexcept:
```
seberg:
```suggestion
        bint keepdims) except *:
```
This allocates a shape, plus is exposed to Python, so indexing is shady at least for that.

cupy/_core/_reduction.pxd
```diff
@@ -10,7 +10,8 @@ cdef Py_ssize_t _block_size
 cpdef tuple _get_axis(object axis, Py_ssize_t ndim)
 
 cpdef shape_t _get_out_shape(
-    const shape_t& shape, tuple reduce_axis, tuple out_axis, bint keepdims)
+    const shape_t& shape, tuple reduce_axis, tuple out_axis,
+    bint keepdims) noexcept
```
seberg:
```suggestion
    bint keepdims) except *
```

cupy/_core/_reduction.pyx
```diff
@@ -471,11 +472,11 @@ cdef class _AbstractReductionKernel:
             best.user_attrs['block_stride'],
             best.params['out_block_num'])
 
-    cdef inline void _launch(
+    cdef inline bint _launch(
             self, out_block_num, block_size, block_stride,
             in_args, out_args, in_shape, out_shape, type_map,
             map_expr, reduce_expr, post_map_expr, reduce_type,
-            stream, params):
+            stream, params) except 0:
```
seberg:
```suggestion
            stream, params) except -1:
```
I guess I suggested that, but let's go with `-1`. Otherwise we have to explicitly return 1 on success.

cupy/_core/_reduction.pxd
```diff
@@ -31,11 +32,11 @@ cdef class _AbstractReductionKernel:
         bint keepdims, bint reduce_dims, int device_id,
         stream, bint try_use_cub=*, bint sort_reduce_axis=*)
 
-    cdef void _launch(
+    cdef bint _launch(
         self, out_block_num, block_size, block_stride,
         in_args, out_args, in_shape, out_shape, types,
         map_expr, reduce_expr, post_map_expr, reduce_type,
-        stream, params)
+        stream, params) except 0
```
seberg:
```suggestion
        stream, params) except -1
```

cupy/_core/core.pyx
```diff
@@ -2446,14 +2446,14 @@ cdef inline str _translate_cucomplex_to_thrust(str source):
...
-cpdef bint use_default_std(tuple options):
+cpdef bint use_default_std(tuple options) except -1:
     cdef str opt
     for opt in options:
         if ('-std=' in opt) or ('--std=' in opt):
             return False
     return True
 
-cpdef void warn_on_unsupported_std(tuple options):
+cpdef void warn_on_unsupported_std(tuple options) except *:
```
seberg:
```suggestion
cpdef warn_on_unsupported_std(tuple options):
```
Let's just return an object here, even on old Cython that always uses NULL to indicate error.

cupy/_core/internal.pyx
```diff
@@ -12,23 +12,24 @@ from cupy.exceptions import AxisError
...
+cpdef inline Py_ssize_t prod(const vector.vector[Py_ssize_t]& args) noexcept:
     cdef Py_ssize_t n = 1
     for i in range(args.size()):
         n *= args[i]
     return n
 
 
 @cython.profile(False)
-cpdef inline Py_ssize_t prod_sequence(object args):
+cpdef inline Py_ssize_t prod_sequence(object args) except *:
```
seberg:
```suggestion
cpdef inline Py_ssize_t prod_sequence(object args) except? -1:
```

cupy/_core/internal.pxd
```diff
@@ -7,30 +7,31 @@ from cupy._core._carray cimport strides_t
 from cupy._core.core cimport _ndarray_base
 
 
-cpdef Py_ssize_t prod(const vector.vector[Py_ssize_t]& args)
+cpdef Py_ssize_t prod(const vector.vector[Py_ssize_t]& args) noexcept
 
-cpdef Py_ssize_t prod_sequence(object args)
+cpdef Py_ssize_t prod_sequence(object args) except *
```
seberg:
```suggestion
cpdef Py_ssize_t prod_sequence(object args) except? -1
```

cupy/cuda/memory.pyx
```diff
@@ -1009,19 +1009,19 @@ cdef class PooledMemory(BaseMemory):
...
 
-cpdef size_t _bin_index_from_size(size_t size):
+cpdef size_t _bin_index_from_size(size_t size) noexcept:
     """Returns appropriate bins index from the memory size."""
     # avoid 0 div checking
     return (size - 1) // ALLOCATION_UNIT_SIZE
 
 
-cdef bint _lock_no_gc(recursive_mutex& lock):
+cdef bint _lock_no_gc(recursive_mutex& lock) except *:
```
seberg:
```suggestion
cdef bint _lock_no_gc(recursive_mutex& lock) except -1:
```
Doubt this can fail, but doesn't matter.

cupy/cuda/memory.pyx
```diff
@@ -1128,7 +1128,7 @@ cdef class _Arena:
         free_list.add(chunk)
         self._flag[index] = 1
 
-    cdef bint remove_from_free_list(self, _Chunk chunk):
+    cdef bint remove_from_free_list(self, _Chunk chunk) except *:
```
seberg:
```suggestion
    cdef bint remove_from_free_list(self, _Chunk chunk) except -1:
```

cupy/cuda/memory.pyx
```diff
@@ -1375,7 +1375,7 @@ cdef class SingleDeviceMemoryPool:
             DeprecationWarning)
         self.free_all_blocks()
 
-    cpdef size_t n_free_blocks(self):
+    cpdef size_t n_free_blocks(self) except *:
```
seberg:
```suggestion
    cpdef size_t n_free_blocks(self) except -1:
```

cupy/cuda/memory.pyx
```diff
@@ -1390,7 +1390,7 @@ cdef class SingleDeviceMemoryPool:
             self._free_lock.unlock()
         return n
 
-    cpdef size_t used_bytes(self):
+    cpdef size_t used_bytes(self) except *:
```
seberg:
```suggestion
    cpdef size_t used_bytes(self) except -1:
```

cupy/cuda/memory.pyx
```diff
@@ -1403,7 +1403,7 @@ cdef class SingleDeviceMemoryPool:
             self._in_use_lock.unlock()
         return size
 
-    cpdef size_t free_bytes(self):
+    cpdef size_t free_bytes(self) except *:
```
seberg:
```suggestion
    cpdef size_t free_bytes(self) except -1:
```

cupy/cuda/memory.pyx
```diff
@@ -1422,7 +1422,7 @@ cdef class SingleDeviceMemoryPool:
             self._free_lock.unlock()
         return size
 
-    cpdef size_t total_bytes(self):
+    cpdef size_t total_bytes(self) except *:
```
seberg:
```suggestion
    cpdef size_t total_bytes(self) except -1:
```

cupy/cuda/memory.pyx
```diff
@@ -1450,7 +1450,7 @@ cdef class SingleDeviceMemoryPool:
         with lock_and_no_gc(self._total_bytes_lock):
             self._total_bytes_limit = size
 
-    cpdef size_t get_limit(self):
+    cpdef size_t get_limit(self) except *:
```
seberg:
```suggestion
    cpdef size_t get_limit(self) except -1:
```

cupy/cuda/memory.pyx
```diff
@@ -1929,7 +1929,7 @@ cdef class MemoryAsyncPool:
         # to reserve at least 0 bytes
         runtime.memPoolTrimTo(pool, 0)
 
-    cpdef size_t n_free_blocks(self):
+    cpdef size_t n_free_blocks(self) except *:
```
seberg:
```suggestion
    cpdef size_t n_free_blocks(self) except? -1:
```

cupy_backends/cuda/_softlink.pyx
```diff
@@ -32,7 +32,7 @@ cdef class SoftLink:
                     raise RuntimeError(msg) from e
                 warnings.warn(msg)
 
-    cdef func_ptr get(self, str name):
+    cdef func_ptr get(self, str name) except *:
```
seberg:
```suggestion
    cdef func_ptr get(self, str name) except NULL:
```

cupy_backends/cuda/_softlink.pxd
```diff
@@ -1,9 +1,9 @@
...
+ctypedef int (*func_ptr)(...) noexcept nogil  # NOQA
 
 cdef class SoftLink:
     cdef:
         object error
         str prefix
         str _libname
         object _cdll
-        func_ptr get(self, str name)
+        func_ptr get(self, str name) except *
```
seberg:
```suggestion
        func_ptr get(self, str name) except NULL
```

cupy_backends/cuda/stream.pxd
```diff
@@ -1,7 +1,8 @@
 from libc.stdint cimport intptr_t
 
-cdef intptr_t get_current_stream_ptr()
-cdef intptr_t get_stream_ptr(int device_id)
+# Uses Python threading
+cdef intptr_t get_current_stream_ptr() except *
+cdef intptr_t get_stream_ptr(int device_id) except *
```
seberg:
```suggestion
cdef intptr_t get_current_stream_ptr() except? -1
cdef intptr_t get_stream_ptr(int device_id) except? -1
```

cupy_backends/cuda/stream.pyx
```diff
@@ -43,7 +44,7 @@ cdef class _ThreadLocal:
         return curr_stream
 
 
-cdef intptr_t get_current_stream_ptr():
+cdef intptr_t get_current_stream_ptr() except *:
```
seberg:
```suggestion
cdef intptr_t get_current_stream_ptr() except? -1:
```

cupy_backends/cuda/stream.pyx
```diff
@@ -52,7 +53,7 @@ cdef intptr_t get_current_stream_ptr():
     tls = _ThreadLocal.get()
     return <intptr_t>tls.get_current_stream_ptr()
 
-cdef intptr_t get_stream_ptr(int device_id):
+cdef intptr_t get_stream_ptr(int device_id) except *:
```
seberg:
```suggestion
cdef intptr_t get_stream_ptr(int device_id) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2085,7 +2085,7 @@ cpdef size_t shgesv_bufferSize(intptr_t handle, int n, int nrhs, size_t dA,
 
 cpdef int zzgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg: Sorry, but I am a bit unsure about the return value, so let's just make it ? -1
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2098,7 +2098,7 @@ cpdef int zzgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int zcgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2111,7 +2111,7 @@ cpdef int zcgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int zygesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2124,7 +2124,7 @@ cpdef int zygesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int zkgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2137,7 +2137,7 @@ cpdef int zkgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int ccgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2150,7 +2150,7 @@ cpdef int ccgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int cygesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2163,7 +2163,7 @@ cpdef int cygesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int ckgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2176,7 +2176,7 @@ cpdef int ckgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int ddgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2189,7 +2189,7 @@ cpdef int ddgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int dsgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2202,7 +2202,7 @@ cpdef int dsgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int dxgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2215,7 +2215,7 @@ cpdef int dxgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int dhgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2228,7 +2228,7 @@ cpdef int dhgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int ssgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2241,7 +2241,7 @@ cpdef int ssgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int sxgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2254,7 +2254,7 @@ cpdef int sxgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
 
 cpdef int shgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except -1:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy_backends/cuda/libs/cusolver.pyx
```diff
@@ -2439,7 +2439,7 @@ cpdef size_t shgels_bufferSize(intptr_t handle, int m, int n, int nrhs,
 
 cpdef int zzgels(intptr_t handle, int m, int n, int nrhs, size_t dA, int ldda,
                  size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork, size_t dInfo):
+                 size_t dwork, size_t lwork, size_t dInfo) except *:
```
seberg:
```suggestion
                 size_t dwork, size_t lwork, size_t dInfo) except? -1:
```

cupy/_core/_reduction.pyx
ManuCorrea: I thought this code was unlikely to fail, but added except *

cupy_backends/cuda/stream.pyx
ManuCorrea: I was keeping it except * because the use of the _ThreadLocal class. I am unsure between both options, I'll do as you say.

cupy_backends/cuda/libs/cusolver.pyx
ManuCorrea: Since that is not Python code underneath that wouldn't trigger:

The ? indicates that the value -1 may signal an error.

In this case, Cython generates a call to [PyErr_Occurred()](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Occurred) if the exception value is returned, to make sure it really received an exception and not just a normal result.

Seems that is positive integers are the ones that signal error: https://forums.developer.nvidia.com/t/cusolverstatus-t/214845

Should we create a check_status function for cusolver?

ManuCorrea: Since that calls cusolver API there is no Python code underneath, so the error would not trigger anything if I got right the behaviour:
"The ? indicates that the value -1 may signal an error.

In this case, Cython generates a call to PyErr_Occurred() if the exception value is returned, to make sure it really received an exception and not just a normal result."

Seems that what signals error are positive integer values, should we make a check_status function for cusolver?
https://forums.developer.nvidia.com/t/cusolverstatus-t/214845

cupy/_core/_reduction.pyx
```diff
@@ -157,7 +157,7 @@ cpdef tuple _get_axis(object axis, Py_ssize_t ndim):
 
 cpdef shape_t _get_out_shape(
         const shape_t& shape, tuple reduce_axis, tuple out_axis,
```
ManuCorrea, line 170: @seberg

> Suggested change:
> 
>         bint keepdims) except *:  
> This allocates a shape, plus is exposed to Python, so indexing is shady at least for that.

I thought this code was unlikely to fail, but added except *

cupy_backends/cuda/libs/cusolver.pxd
```diff
@@ -291,46 +291,46 @@ cpdef size_t shgesv_bufferSize(intptr_t handle, int n, int nrhs, size_t dA,
 
 cpdef int zzgesv(intptr_t handle, int n, int nrhs, size_t dA, int ldda,
                  size_t dipiv, size_t dB, int lddb, size_t dX, int lddx,
-                 size_t dwork, size_t lwork_bytes, size_t dInfo)
+                 size_t dwork, size_t lwork_bytes, size_t dInfo) except -1
```
ManuCorrea: @seberg committing your changes outdated all comments so I am unsure if you were able to see my replies, just in case:

> Sorry, but I am a bit unsure about the return value, so let's just make it ? -1


Since that calls cusolver API there is no Python code underneath, so the error would not trigger anything if I got right the behaviour:
"The ? indicates that the value -1 may signal an error.

In this case, Cython generates a call to PyErr_Occurred() if the exception value is returned, to make sure it really received an exception and not just a normal result."

Seems that what signals error are positive integer values, should we make a check_status function for cusolver?
https://forums.developer.nvidia.com/t/cusolverstatus-t/214845

cupy_backends/cuda/stream.pyx
```diff
@@ -31,7 +31,8 @@ cdef class _ThreadLocal:
             device_id = runtime.getDevice()
         self.current_stream[device_id] = ptr
 
-    cdef intptr_t get_current_stream_ptr(self, int device_id=-1):
+    # getDevice may return an error
+    cdef intptr_t get_current_stream_ptr(self, int device_id=-1) except? -1:
```
ManuCorrea: @seberg:
> Suggested change:
> 
> cdef intptr_t get_stream_ptr(int device_id) except? -1:  

I was keeping it except * because the use of the _ThreadLocal class. I am unsure between both options, I'll do as you say.

cupy_backends/cuda/libs/cusolver.pxd
seberg: I already did, I'll force push away any other changes (including a rebase).

You are reading it the wrong way around.  Any value that isn't -1 means there is no error. Adding `except? -1` may generate an unnecessary `PyErr_Occurred()`, but only if the return value happens to be `-1` (and we don't care about speed on error).

So this isn't ideal in theory if we know that -1 is a very typical return value (and others are not).  We know that -1 is *impossible* if there is no error.

But... I am opting for this anyway, because it is the default.  I.e. after switching to Cython 3 proper, we can delete them again (I just like to see things warning free once I think -- maybe I'll include the commits to undo it directly here though).

seberg:
@ManuCorrea please don't add any commits, I am just making a mass edit on my side.  I don't really want to merge this before CuPy v14 code freeze, but directly after I would go through once more (fix merge conflicts) and get it in.

seberg:
As a small point, I went through now fixing things as they look right to me.
* Sometimes just remove the `void` return (returning Python `None` is OK and can signal errors).  Since it's shorter also using this rather than another return.
* Added `except? -1` very generously even for things that small like they will only return positive numbers.  In part to be safe and in part because that is the default if you removed it on Cython 3 (i.e. now).

The (currently) **last commit removes the Cython directive after verifying that no warnings exist**.  I don't really want to merge this before v14, but after that my plan would be to go over once more with fresh eyes.
(It would be nice to follow up with removing the default `except? -1` again, but I am tempted to do that in a second PR, even if it is a bit more churny)

ManuCorrea:
@seberg perfect! Sounds good, thank you for the review and feedback. Let me know if you would like something else on my side.

seberg:
OK, I think this is settling. I realized that Cython (currently) always includes code with `Pyx_WriteUnraisable`. That is expected in `__deallocate__` and capsule deleters, but otherwise we hope to never see it.
(There is a Cython issue about adding a warning in those cases. It's not always problematic, but the refactors seemed OK if that allows the CI.)

So this way there is actual proof that all `noexcept` marked functions are actually `noexcept`, but it also means that the second to last commit is slightly non-trivial changes.

seberg:
This is nerve wrecking... every time you find something else small.

I added another commit that undoes some changes (although not the whitespace parts) where I hadded added `except? -1` as that is the default now. There also were 1-2 typos.

The other thing is allocators:
* The C-allocator is a mess, I kept it a mess and added pretty random `except` clauses (it is also untested).
* The thrust sets an allocator that I added to the `noexcept` exceptions, because short of raising a C++ exception, I suspect that is actually best.

Since we don't have *incorrect* `noexcept` now with CI, the main regression would be speed for `noexcept *` uses in hot paths.  That is for later!

ManuCorrea:
@seberg thank you for the deep dive into the Cython internals. I was unaware of the `Pyx_WriteUnraisable` glad the errors can be catch in the CI.

Should an issue be open for improving the c/thrust allocators since you mentioned they are a mess? Or is it something we can just life with?

I’m happy to take on a follow-up PR to look into the performance "hot paths" for the remaining except * cases, since I’m already familiar with the changes.

cupy/_core/internal.pyx
```diff
@@ -12,7 +12,7 @@ from cupy.exceptions import AxisError
 
 
 @cython.profile(False)
-cpdef inline Py_ssize_t prod(const vector.vector[Py_ssize_t]& args):
+cpdef inline Py_ssize_t prod(const vector.vector[Py_ssize_t]& args) noexcept:
     cdef Py_ssize_t n = 1
     for i in range(args.size()):
```
leofang, line 17: While we're here... are we sure this variable `i` is a C variable and not a Python one? Maybe we need to declare `cdef int i` to be safe?

cupy/_core/internal.pyx
```diff
@@ -82,7 +83,7 @@ cpdef inline bint vector_equal(
 
 
 @cython.profile(False)
-cdef void get_reduced_dims(
+cdef get_reduced_dims(
```
leofang: We could declare this as 
```cython
cdef int get_reduced_dims(...) except? -1:
```
so that we don't generate any Python wrapper. (Without a return type, Cython by default assumes a Python `None` is returned.)

cupy/_core/internal.pxd
```diff
@@ -7,30 +7,31 @@ from cupy._core._carray cimport strides_t
...
     Py_ssize_t itemsize, shape_t& reduced_shape,
     strides_t& reduced_strides)
 
 # Computes the contiguous strides given a shape and itemsize.
 # Returns the size (total number of elements).
 cdef Py_ssize_t get_contiguous_strides_inplace(
-    const shape_t& shape, strides_t& strides,
-    Py_ssize_t itemsize, bint is_c_contiguous, bint zeros_for_zerosize)
+    const shape_t& shape, strides_t& strides, Py_ssize_t itemsize,
+    bint is_c_contiguous, bint zeros_for_zerosize)
```
leofang: Q: I am puzzled. We changed the signature here by reordering the args (which is fine), but I don't see any corresponding changes in all call sites?

cupy/_core/dlpack.pxd
```diff
@@ -70,7 +70,7 @@ cdef extern from './include/cupy/_dlpack/dlpack.h' nogil:
         DLTensor dl_tensor
 
 
-cdef DLDevice get_dlpack_device(_ndarray_base array)
+cdef DLDevice get_dlpack_device(_ndarray_base array) except *
```
leofang: FYI, for functions that return a C type (ex: enum) like this, we figured out a way to avoid unconditional exception check `except*`. The trick is to define a dummy value (ex: `DLDevice(-1)`) and use it for `except?`. Example:
https://github.com/NVIDIA/cuda-python/blob/0bd28ccf52bb022dab5c9383d24119684782063f/cuda_bindings/cuda/bindings/cycufile.pxd#L367-L385

cupy/cuda/pinned_memory.pyx
```diff
@@ -90,7 +90,7 @@ cdef class PinnedMemoryPointer:
         return self.__iadd__(-offset)
 
     cpdef size_t size(self):
-        return self.mem.size - (self.ptr - self.mem.ptr)
+        return <size_t>self.mem.size - (self.ptr - <intptr_t>self.mem.ptr)
```
leofang: I am overly paranoid, please always use parenthesis for casting
```suggestion
        return <size_t>(self.mem.size) - (self.ptr - <intptr_t>(self.mem.ptr))
```

cupy_backends/cuda/libs/cublas.pxd
leofang, line 1: Can we please not make whitespace changes?

cupy_backends/cuda/libs/cublas.pyx
leofang, line 1: ditto

cupy_backends/cuda/libs/cusolver.pxd
leofang, line 1: ditto

cupy/_core/internal.pxd
seberg: Ack, not changed, but there was an except clause intermittently (to get to warning free)

cupy/_core/internal.pyx
seberg: I figure None is OK since it is only an unnecessary decref at the callsite really but also happy to go with bint or so int.
(The thing thatis nicer is not having to write `return 0` explicitly.)

But yeah, that was also my first thought, so happy to change if you prefer?

cupy/_core/dlpack.pxd
seberg: Yeah, I was aware a bit, but wasn't sure if it's worth it (there is also one nvrtc struct).
But I didn't think of using a cast, that probably makes it nicer!

seberg: Nice!  I was a bit creative to make it slightly more compact, but applied.  I re-ordered the result for a similar one in `nvtx` it isn't exposed in the `.pxd` file so private (here it makes sense maybe, but the struct there is akward large, I think).

cupy/_core/internal.pyx
seberg: Yeah, otherwise my test would have failed, but let me just declare it...

cupy/_core/internal.pyx
leofang: yup let's do that

seberg: OK, changed it for any `cdef` function (not `cpdef`, feels odd there).

cupy/_core/dlpack.pxd
```diff
@@ -31,6 +39,9 @@ cdef extern from './include/cupy/_dlpack/dlpack.h' nogil:
         DLDeviceType device_type
         int32_t device_id
 
+    # Enable using `except DLDevice_err` (there is no direct way)
+    const DLDevice DLDevice_err "((DLDevice){(DLDeviceType)-1, -1})"
```
leofang: This, and a few other instances, seems to choke MSVC lol
```
error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
```

seberg:
Had missed that. I guess using `DLDevice{(DLDeviceType)-1, -1}` is nicer and should work as well.

/test mini

leofang (approved):

PR #9660, opened 2026-02-02
MAINT: Fix cuda-version check and disable bfloat16 on NumPy <2.1.2

seberg:
NumPy versions 2.0 to 2.1.1 had a reference counting bug when promoting user old style dtypes (i.e. `bfloat16`).
The obvious trigger for this was code like `1 / cupy_array` where we eventually see if the `bfloat16` loop works and run into the NumPy bug.

This means that just the existance of bfloat16 loops trigger this and the sane thing to me seems to me to just require NumPy >=2.1.2.

This was not found in the PR stage, because the tests did not generate a case where a Python integer was devided.
I.e. it would probably have triggered all over, but only with `CUPY_TEST_FULL_COMBINATION=1`!

EDIT: I checked locally that the failing cuda 12.0 CI run should be working with this.

### DIFF
--- a/cupy/_util.pyx
+++ b/cupy/_util.pyx
@@ -223,6 +223,11 @@ def is_shutting_down(_shutting_down=_shutting_down):
 
 
 try:
+    # We cannot add loops for bfloat16 on because the loops existance alone
+    # means NumPy code-paths are used that will lead to segfaults.
+    # (I.e. these crashes happen even without bfloat16 ever being used.)
+    if numpy.lib.NumpyVersion(numpy.__version__) < "2.1.2":
+        raise ImportError("NumPy had a critical result-type issue")
     import ml_dtypes
 except ImportError:
     def bf16_loop(in_types=1, out_types=1, code=None):
--- a/cupy/_core/_scalar.pyx
+++ b/cupy/_core/_scalar.pyx
@@ -70,18 +70,21 @@ cdef _setup_type_dict():
     for t in ('cudaTextureObject_t',):
         _typenames[t] = (t, None)
 
-    try:
-        import ml_dtypes
-    except ImportError:
-        pass
-    else:
-        dt = numpy.dtype(ml_dtypes.bfloat16)
-        _dtype_kind_size_dict[dt] = ("V", 2)
-        _typenames[dt] = (
-            "bfloat16", '#include "cupy/bfloat16.cuh"')
-        _dtype_kind_size_dict[dt.type] = ("V", 2)
-        _typenames[dt.type] = (
-            "bfloat16", '#include "cupy/bfloat16.cuh"')
+    # See also _util.pyx. older NumPy versions will cause crashes if we add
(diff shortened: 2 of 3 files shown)


cupy/_util.pyx
```diff
@@ -223,6 +223,11 @@ def is_shutting_down(_shutting_down=_shutting_down):
 
 
 try:
+    # We cannot add loops for bfloat16 on because the loops existance alone
+    # means NumPy code-paths are used that will lead to segfaults.
+    # (I.e. these crashes happen even without bfloat16 ever being used.)
+    if numpy.lib.NumpyVersion(numpy.__version__) < "2.1.2":
+        raise ImportError("NumPy had a critical result-type issue")
```
seberg, line 230: Yeah, a bit weird to raise, but I didn't feel like adding an indentation level below.

Could refactor it though, to define `bf16_loop` versions and then alias the one we actually want to use.

leofang: Instead of raising an import error, would it be possible to just make the loop a no-op? Would it be sufficient  to avoid the NumPy bug?

leofang (commented):
I am wondering if we can keep the lower bound at 2.0.

cupy/_util.pyx
seberg: Oh, the error is caught?  But yeah, maybe nicer to not raise an error for control flow.

leofang: ahhh sorry I was reading this diff on my cell, I see it now.

leofang:
Thanks, @seberg! 

Q: Do we trigger the mini CI? Or is there a particular pipeline we want to confirm? Since the segfault only happens in the main branch, I am not sure how useful triggering per-PR CIs versus just merging and testing on main...

seberg:
I would be tempted to just wing it here and see if it passes on main.

I am not sure we can run `CUPY_TEST_FULL_COMBINATION=1` easily, which would be the interesting part (besides from not accidentally disabling `bfloat16` support).

leofang:
> I am not sure we can run `CUPY_TEST_FULL_COMBINATION=1` easily, which would be the interesting part (besides from not accidentally disabling `bfloat16` support).

We could adjust this line
https://github.com/cupy/cupy/blob/513b954dcac7cdb70a725e8eed1af58642a79972/.pfnci/linux/tests/actions/_environment.sh#L26-L27
but given that the cache is not fully rebuilt on main yet, testing in the PR would still require one or two retries. I incline to merge and monitor.

seberg:
Agreed, let's just do that, seems simpler.

/test skip

leofang:
#9659 is not merged yet, so we'll have to backport manually.

leofang:
To complete the circle: After #9494 was merged, we observed a segfault on the main branch (ex: https://ci.preferred.jp/r/job/208659). This PR is to fix the segfault (due to a NumPy bug, see the PR description). Sebastian has backported this PR to #9659.

PR #9661, opened 2026-02-02
Fixed cupy clip behaviour to match numpy clip and added tests

### LINKED ISSUE #9031: `cupy.clip` behaves differently with `numpy.clip` on `nan`
### Description

`cupy.clip` behaves differently with `numpy.clip` on `nan`

### To Reproduce

```py
import numpy as np
import cupy as cp

print(np.clip(np.array([1,2]),a_min=np.nan,a_max=np.nan))
print(cp.clip(cp.array([1,2]),a_min=cp.nan,a_max=cp.nan))
```
```bash
[nan nan]
[1. 2.]
```

### Installation

### Environment

```
OS                           : Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
CuPy Version                 : 13.4.0
NumPy Version                : 2.2.3
SciPy Version                : 1.15.2
CUDA Build Version           : 12080
CUDA Runtime Version         : 12080 (linked to CuPy) / 12020 (locally installed)
cuBLAS Version               : 120201
cuSOLVER Version             : (11, 5, 2)
cuSPARSE Version             : 12101
Device 0 Name                : NVIDIA GeForce GTX 1650

```

### Additional Information

_No response_

ManuCorrea:
This code changes fixes: https://github.com/cupy/cupy/issues/9031 making cupy clip to have the same behavior as numpy clip.

In order to fix it I created a different function for the float types that checks if any of the min/max arguments are nan. Only applied to float values since nan is a float type.

### DIFF
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -1143,12 +1143,33 @@ _sqrt = create_ufunc(
     ''')
 
 
+# Define the specialized logic for floats
+_float_routine = '''
+if (isnan(in1) || isnan(in2)) {
+    out0 = (out0_type)nan("");
+} else {
+    out0 = in1 > in2 ? in2 : (in0 < in1 ? in1 : (in0 > in2 ? in2 : in0));
+}
+'''
+
+# Define the default logic for integers
+_default_routine = (
+    'out0 = in1 > in2 ? in2 : '
+    '(in0 < in1 ? in1 : (in0 > in2 ? in2 : in0))'
+)
+
+
 _clip = create_ufunc(
     'cupy_clip',
-    ('???->?', 'bbb->b', 'BBB->B', 'hhh->h', 'HHH->H', 'iii->i', 'III->I',
-     'lll->l', 'LLL->L', 'qqq->q', 'QQQ->Q',
-     'eee->e', 'fff->f', 'ddd->d'),
-    'out0 = in1 > in2 ? in2 : (in0 < in1 ? in1 : (in0 > in2 ? in2 : in0))')
+    (
+        '???->?', 'bbb->b', 'BBB->B', 'hhh->h', 'HHH->H', 'iii->i', 'III->I',
+        'lll->l', 'LLL->L', 'qqq->q', 'QQQ->Q',
+        ('eee->e', _float_routine),
+        ('fff->f', _float_routine),
+        ('ddd->d', _float_routine)
+    ),
+    _default_routine
+)
 
 
 # Variables to expose to Python
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -1144,7 +1144,7 @@ _sqrt = create_ufunc(
 
 
 # Define the specialized logic for floats
-_float_routine = '''
+_float_clip = '''
 if (isnan(in1) || isnan(in2)) {
     out0 = (out0_type)nan("");
 } else {
@@ -1164,9 +1164,9 @@ _clip = create_ufunc(
     (
         '???->?', 'bbb->b', 'BBB->B', 'hhh->h', 'HHH->H', 'iii->i', 'III->I',
         'lll->l', 'LLL->L', 'qqq->q', 'QQQ->Q',
-        ('eee->e', _float_routine),
-        ('fff->f', _float_routine),
-        ('ddd->d', _float_routine)
+        ('eee->e', _float_clip),
+        ('fff->f', _float_clip),
+        ('ddd->d', _float_clip)
     ),
     _default_routine
 )
--- a/tests/cupy_tests/math_tests/test_misc.py
+++ b/tests/cupy_tests/math_tests/test_misc.py
@@ -149,6 +149,38 @@ class TestMisc:
         a_max = xp.array([[10], [9], [8]], dtype=dtype)
         return a.clip(a_min, a_max)
 
+    @testing.for_float_dtypes()
+    @testing.numpy_cupy_allclose(atol=1e-5)
+    def test_clip_nan_in_array(self, xp, dtype):
+        # NaN in array is preserved; finite bounds clip other values
+        nan = float('nan')
+        a = xp.array([1.0, nan, 5.0, -1.0, 3.0], dtype=dtype)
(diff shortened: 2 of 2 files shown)


ManuCorrea:
Hello @seberg I think the CI error may be unrelated to the changes but to the CI itself? Should we rerun it?

```
The job was not acquired by Runner of type hosted even after multiple attempts
Internal server error. Correlation ID: c245d51b-6957-4ee8-b712-6af655d8490a
```

cupy/_core/_routines_math.pyx
```diff
@@ -1143,12 +1143,33 @@ _sqrt = create_ufunc(
     ''')
 
 
+# Define the specialized logic for floats
+_float_routine = '''
```
seberg:
```suggestion
_float_clip = '''
```

cupy/_core/_routines_math.pyx
```diff
@@ -1143,12 +1143,33 @@ _sqrt = create_ufunc(
...
+} else {
+    out0 = in1 > in2 ? in2 : (in0 < in1 ? in1 : (in0 > in2 ? in2 : in0));
+}
+'''
+
+# Define the default logic for integers
+_default_routine = (
+    'out0 = in1 > in2 ? in2 : '
+    '(in0 < in1 ? in1 : (in0 > in2 ? in2 : in0))'
+)
```
seberg, line 1159: Just keep it inlined?

cupy/_core/_routines_math.pyx
```diff
@@ -1143,12 +1143,33 @@ _sqrt = create_ufunc(
     ''')
 
 
+# Define the specialized logic for floats
+_float_routine = '''
+if (isnan(in1) || isnan(in2)) {
+    out0 = (out0_type)nan("");
+} else {
+    out0 = in1 > in2 ? in2 : (in0 < in1 ? in1 : (in0 > in2 ? in2 : in0));
+}
```
seberg, line 1152: Do you know if this has a performance impact? 

I wonder if there are other re-arrangements that might be better:
```
    out0 = in0;
    if (out0 < in1 || isnan(in1)) {
        out0 = in1;
    }
    if (out0 < in2) || isnan(in2)) {
        out0 = in2;
    }
```
Assuming we don't care about *which* NaN is propagated.

seberg (commented):
This looks fine and maybe I am off base a bit. But: I think this is a silly niche case and the original issue smells a lot like "I ran an automated tools and found this difference!!1!".

If there is almost no performance hit, we should do this. Otherwise, I am not sure we should care about the min/max value containing NaNs -- although I may be off-base here.

seberg:
> The job was not acquired by Runner of type hosted even after multiple attempts

Dunno which job, but the interesting tests only run once things seem settled and a maintainer asked for it (they are rather expensive).

cupy/_core/_routines_math.pyx
ManuCorrea: Cython linter complained about the line too long, so I had to split it

cupy/_core/_routines_math.pyx
ManuCorrea: I haven't benchmark it but if we want to proceed with the implementation I can benchmark the different possibilities.

Also another way for the else part of floats could be:
`out0 = fminf(in2, fmaxf(in1, in0))`
Using CUDA functions could benefit from hardware acceleration right?

Sadly the way they handle NAN are not valid for this case: https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__DOUBLE.html#_CPPv44fmaxdd

In both fmax and fmin:

- If both arguments are NaN, returns NaN.
- If one argument is NaN, returns the numeric argument.

seberg: Right that is the definition for `fmin`/`fmax` as per C/C++ standards. My opinion is somewhat that it would be OK to be a bit sloppy or different around NaN handling here compared to NumPy.
So if there is no real speed impact, I don't see a reason why not to match NumPy.  But if there is then I am not sure.

That also goes a bit for the case of `min > max` inputs, which seems also rather "undefined" to me.

(But I have to say that my gut feeling for these choices may still need a bit alignment with what CuPy did in the past.)

cupy/_core/_routines_math.pyx
seberg: The cython linter is pretty strict.  I would probably still just use an inline:
```
    '''
    <code block>
    '''
```
, I think.  but not a big deal.

ManuCorrea:
> This looks fine and maybe I am off base a bit. But: I think this is a silly niche case and the original issue smells a lot like "I ran an automated tools and found this difference!!1!".
> 
> If there is almost no performance hit, we should do this. Otherwise, I am not sure we should care about the min/max value containing NaNs -- although I may be off-base here.

I agree with you looks like was found with automated tools. I am also unsure how likely are people to hit a case like this. I would proceed to either:

1. Document this numpy difference in the docs in case somewhere in the world there is a person suffering because their migration from numpy to cupy is not working out as expected.
2. I can benchmark the changes if we think is worth implementing.

> Dunno which job, but the interesting tests only run once things seem settled and a maintainer asked for it (they are rather expensive).

That was inside the "Pre-review Tests / build-cuda/rocm (pull_request)"

I am also considering, another implementation would be to check if any of both min/max arguments is NAN and just return an array with the same shape full of NAN? Single check on CPU and then just building the array of NAN, I am aware this moves away from GPU implementation and is niche use case so it may be not desirable . I checked on numpy the behavior and seems that any of both min/max as NAN makes the whole array NAN:
```
>>> print(np.clip(np.array([1,2]),a_min=np.nan,a_max=np.nan))
[nan nan]
>>> print(np.clip(np.array([1,2,3]),a_min=2,a_max=np.nan))
[nan nan nan]
>>> print(np.clip(np.array([1,2,3]),a_min=np.nan,a_max=2))
[nan nan nan]
>>> print(np.clip(np.array([1,2,3]),a_min=2,a_max=2))
[2 2 2]
```

seberg:
Sorry, I seem to not have followed up earlier, because I think it still stands roughly at the same spot:
We need to know what the performance impact is. Then we can decide where to go there.

PR #9672, opened 2026-02-09
MAINT: A few small micro-optimizations

seberg:
I was looking a bit at launch overheads end of last week. Bigger fish will come in other PRs eventually. This one does:
* `cpdef` is actually slowish on hot paths for light-weight functions. (It checks for monkey-patching, which is _some_ work.)
* `1 << 31` overflows int's. Cython needs a hint to use a 64bit integer (or it uses Python).
* I think due to `memcpy` alignment, the `memcpy` seemed not inlined. Barely noticable, but I also prefered this pattern slightly.

All of this is maybe 5% of the overhead int `cp.add(arr1d, arr1d)`. Not much, but after a while a few 5% improvements will add up.

---

No rush just thought I'd get the ball rolling.  More interesting will be e.g. less trivial refactors in `kernel.pyx` mostly.

### DIFF
--- a/cupy/_core/_carray.pyx
+++ b/cupy/_core/_carray.pyx
@@ -53,7 +53,7 @@ cdef class mdspan(function.CPointer):
         self.ptr = <intptr_t>data
 
         cdef size_t offset = 0
-        memcpy(<char*>(data) + offset, &data_ptr, sizeof(data_ptr))
+        (<void**>(<char*>(data) + offset))[0] = data_ptr
         offset += sizeof(data_ptr)
         if ndim != 0:
             try:
@@ -104,10 +104,11 @@ cdef class CArray(function.CPointer):
         self.ptr = <intptr_t>data
 
         cdef size_t offset = 0
-        memcpy(<char*>(data) + offset, &data_ptr, sizeof(data_ptr))
+        (<void **>(<char*>(data) + offset))[0] = data_ptr
         offset += sizeof(data_ptr)
-        memcpy(<char*>(data) + offset, &data_size, sizeof(data_size))
+        (<Py_ssize_t*>(<char*>(data) + offset))[0] = data_size
         offset += sizeof(data_size)
+
         if ndim != 0:
             memcpy(<char*>(data) + offset,
                    shape.data(),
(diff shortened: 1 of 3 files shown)


cupy/_core/_carray.pyx
```diff
@@ -104,19 +104,19 @@ cdef class CArray(function.CPointer):
...
-                   sizeof(Py_ssize_t) * ndim)
-            offset += sizeof(Py_ssize_t) * ndim
-            memcpy(<char*>(data) + offset,
-                   strides.data(),
-                   sizeof(Py_ssize_t) * ndim)
-            offset += sizeof(Py_ssize_t) * ndim
+
+        for i in range(ndim):
+            (<Py_ssize_t*>(<char*>(data) + offset))[i] = shape[i]
+        offset += ndim * sizeof(Py_ssize_t)
```
leofang: Interesting, a for-loop assignment is faster than memcpy?

cupy/_core/_carray.pyx
```diff
@@ -165,7 +164,7 @@ cdef class Indexer:
     cdef void init(self, const shape_t& shape):
         self.shape = shape
         self.size = internal.prod(shape)
-        self._index_32_bits = self.size <= (1 << 31)
+        self._index_32_bits = self.size <= <Py_ssize_t>(1 << 31)
```
leofang, line 169: For my own curiosity, would this work equivalently?

```suggestion
        self._index_32_bits = self.size <= (1ULL << 31)
```

cupy/_core/_carray.pyx
leofang, line 1: FYI I also applied the same memcpy pattern to mdspan, not sure if we want to take care of it at once

cupy/_core/_carray.pyx
seberg: Yeah, that would work as well.

cupy/_core/_carray.pyx
seberg: Nvm... glancing at I thought it maybe is. But actually, the non-inlining of the first part even seems just due to debug mode or so (I thought it was due to missing alignment information.)

Either way, it doesn't matter probably. I kept the single assignment ones, although not sure it is really nicer.

PR #9674, opened 2026-02-09
TST: Slightly bump SVD test tolerance (but tighten it for float64)

seberg:
Locally, I see a very slightly larger need for tolerance, 1.2e-4 seems to be sufficient (even just on atol).
Those finfo values come down to:
* 2.2737367544323206e-13  (float64/complex128)
* 0.00012207031 (float32/complex64)

while I am not sure about the history, that seems like an OK relaxation to me, and it is good to tighten it for float64.


leofang (approved):

kmaehashi:

/test force-skip

PR #9676, opened 2026-02-10
TST: Protect graph tests from failing when GPU is busy

seberg:
I started using `-nX` a lot (very nice for `-k "not slow"`). But when X is large, the graph tests can currently fail because the random number generation may still be in progress by the time we launch the kernels (leading to different results).

Opted for a mix of syncing early vs. late depending on what worked for the specific test.


kmaehashi (approved):

seberg:
Failures were just timeouts.  (I realize this will spam backport PRs that are irrelevant for 14.0.1 and should just hang for now)

/test force-skip

PR #9680, opened 2026-02-11
TST: Skip bfloat16 dlpack test on old CUDA versions

seberg:
Yeah, this really needs a better solution.  But to be ready for release freeze, I think I prefer just to go with the simplest change.


leofang:
let's merge it for v14.0.0 and fix it by v14.1.0

PR #9554, opened 2025-12-27
Improve pathfinder support

leofang:
Close #9501. Close #9537. Close #8604.

Below is an LLM-generated summary.

This pull request introduces significant improvements to how CuPy locates and loads CUDA libraries and headers. The main changes involve integrating the `cuda-pathfinder` utility for more robust and accurate library discovery, simplifying library loading to avoid duplicate loads, and improving support for conda-based CUDA installations. Additionally, the runtime information reporting has been enhanced to include library paths. These changes aim to make CuPy more reliable across different environments and reduce issues with mismatched or missing CUDA libraries.

**Library Loading and Discovery Improvements**
- Transitioned dynamic library loading (e.g., for `cudart`, `cufft`, `cusolver`, `cusparse`, `nvrtc`) to use `cuda-pathfinder`, passing library handles directly to avoid loading the same library multiple times and risking version mismatches. The `SoftLink` class now accepts a handle and uses it to attach to already-loaded libraries. (`cupy_backends/cuda/_softlink.pyx`, `cupy_backends/cuda/api/_runtime_softlink.pxi`, `cupy_backends/cuda/libs/_cnvrtc.pxi`, `cupy_backends/cuda/libs/cusolver.pyx`, `cupy_backends/cuda/libs/cusparse.pyx`, `cupy/cuda/cufft.pyx`) [[1]](diffhunk://#diff-cbc27dbc38858d936362f3f65b9e090ff6192afedaf00dd14b899b1bf4963257L14-R31) [[2]](diffhunk://#diff-6cb8381aee6141f41efe652d4a194c261c3bf38092c14a931303bc3b5d985786R36-R48) [[3]](diffhunk://#diff-08d1fbf5fd5501384a37161816ad3b3b2b042370c5bb8a278d7eaa9ad6bc6939R119-R126) [[4]](diffhunk://#diff-08d1fbf5fd5501384a37161816ad3b3b2b042370c5bb8a278d7eaa9ad6bc6939L165-R150) [[5]](diffhunk://#diff-16ac314758c2277313ec2eaccc5fe6eeae81086bfaa8f1a0df3b9ee83384a975L1007-R1018) [[6]](diffhunk://#diff-30585ad4d0aa93771c86e3f3d1b49d77a7f9bd2f0e83e6f48188d5f09bfeab7dL1374-R1389) [[7]](diffhunk://#diff-f5ad605b6a808958307974d6802c4b960d424dcab7948ffc026105da57237a8eR54-R74)

[...]

PR #9684, opened 2026-02-13
Add guard for unsupported CUDA version in cuFFT `setJITCallback`

kmaehashi:
https://github.com/cupy/cupy/pull/9554#discussion_r2803152708


kmaehashi:
Test failures are due to timeout.

/test force-skip

PR #9532, opened 2025-12-18
Support CUDA 13.1 and NCCL 2.28

kmaehashi:
xref #9515 #9531

(I'd not make this v14.0.0rc1 blocker)

PR #9686, opened 2026-02-13
CI: Add `cupy.win.cuda131`

kmaehashi:
Part of #9515. Follows up #9532.

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -9,7 +9,7 @@ configs {
     requirement {
       cpu: 8
       memory: 50
-      disk: 10
+      disk: 30
       gpu: 1
     }
     time_limit: {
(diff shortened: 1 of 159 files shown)


.pfnci/config.pbtxt
```diff
@@ -801,3 +801,23 @@ configs {
...
+      gpu: 2
+      image: "windows"
+    }
+    time_limit {
+      seconds: 72000
+    }
+    checkout_strategy {
+      include_dot_git: true
+    }
+    environment_variables { key: "GPU" value: "2" }
```
leofang, line 820: Q: Do we need 2 GPUs?

kmaehashi: There are several tests using multiple GPUs. But this indeed doubles the CI cost so worth revisiting. (maybe it's sufficient by doing multi-gpu tests only in one Windows CI variant)

PR #9691, opened 2026-02-13
Fill bfloat16 gaps around arange and some reduction

seberg:
There are a few more ufunc gaps, notably rounding (but ml_dtypes won't support that either) and possible another 1-2 (I hope) that are more niche (I think maybe divmod, but I think that one isn't the end of the world either way).

Maybe arange isn't all that critical in actual code, but is just very obvious to try :(
(I didn't want to do much here, but especially arange seemed a bit embarrassing.)

### DIFF
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -807,6 +807,7 @@ _sum_keep_dtype = create_reduction_func(
     ('?->?', 'b->b', 'B->B', 'h->h', 'H->H', 'i->i', 'I->I', 'l->l', 'L->L',
      'q->q', 'Q->Q',
      ('e->e', (None, None, None, 'float')),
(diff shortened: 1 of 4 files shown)


seberg:
@leofang dunno where we stand on freeze, but if this happens to slip in, that might be nice.

leofang:
I dunno if @kmaehashi has already started building packages, I'll let Kenichi-san answer

leofang:
#9683 is not merged yet, so we might still have time 😆

seberg:
Yeah. If it works cool. If it would be create a hassle, no worries. I don't want to stress about it.

kmaehashi:
Let's get this into v14.0!

seberg:
/test force-skip

3 windows jobs just timed out.

PR #9496, opened 2025-11-24
Update to CCCL v3.1.2

leofang:
All diffs are explained by the excellent [CCCL 3.0 Migration Guide](https://nvidia.github.io/cccl/cccl/3.0_migration_guide.html):
- CUB SpMV is removed. 
   - The CCCL team and the cuSPARSE team suggested that cuSPARSE should be as performant if not more than CCCL, as a result the CUB version has been removed in CCCL v3.0 ([link](https://github.com/NVIDIA/cccl/blob/65050fef2ab8bf10e0d2b053ea6481efea72292c/docs/cccl/3.0_migration_guide.rst?plain=1#L107)).
- `FpLimits` is removed in favor of `cuda::std::numeric_limits` 
- `<thrust/optional>` is removed in favor of `cuda::std::optional` 
- ``cub::TransformInputIterator``: Use ``thrust::transform_iterator`` instead
- ``cub::CountingInputIterator``: Use ``thrust::counting_iterator`` instead
- `cub::{Sum,Min,Max}` are replaced by `cuda::std::plus` and `cuda::{minimum,maximum}`
- Bump C++ standard everywhere (I think!) to C++17 due to CCCL 3.0 requirement

Additional changes:
- `atomicAdd_block` is needed by CUB's recent histogram change ([link](https://github.com/NVIDIA/cccl/pull/4973/))

PR #9693, opened 2026-02-14
Fix CCCL v3 deprecation warnings in CUB and Thrust wrappers

### LINKED ISSUE #9558: Make CCCL compilation warning-free
Follow-up of #9496. When compiling the `cupy.cuda.{thrust, cub}` modules, a few deprecation warnings are raised. We should try to migrate the code to newer constructs in CCCL v3.x so as to eliminate the warnings.

sdace9719:
This update addresses build-time deprecation warnings raised in  `cupy.cuda.{thrust, cub}`  when compiling against CCCL v3.x

Changes:
- cupy/cuda/cupy_cub.cu: Updated `DeviceReduce::ArgMin` and `ArgMax` calls. The interface using `KeyValuePair*` is deprecated; switched to the new interface that accepts separate iterators for values and indices (`&pair->value`, `&pair->key`).
- cupy/cuda/cupy_thrust.cu: Replaced usage of `thrust::less` with `cuda::std::less`.

See #9496


sdace9719:
fixes #9558

leofang:
@sdace9719 Thanks for your contribution! Do you think you can fix the ROCm build? No worries if not, I'll find help 🙂

sdace9719:
Hi, ill work on it.

sdace9719:
Hi @leofang seems that test is passing now.

leofang (approved):

leofang:
@sdace9719 many test failures occur, could you take a look?

Issue #9586, opened 2026-01-12
Revisit the splitting of the ufunc & ElementwiseKernel machineries


leofang:
In #2920 we attempted to merge the two machineries, which seemed a natural move, but had to revert it in #3296 due to performance regression. However, to the best of my knowledge we did not actually figure out what caused the regression, nor was there any follow-up work. Would be nice to revisit this, since I suspect it'd help simplify work in accepting kernel arguments with arbitrary NumPy dtypes (narrow precision, structured, ...).

leofang:
@cupy/code-owners @emcastillo any chance any of you still remember what happened? 😛

emcastillo:
Yeah, can't remember why, didn't get enough time to dive in it too.
My wild guess is some python object manipulations causing the interpreter to kick-in

seberg:
This sounds like we should be able to figure out with some profiling and a bit more time than last time around.
Such a large increase in how much time it took, feels like there might have been a simple flaw such as not caching a part that should have been cached.

Anyway, we should probably look into this on our back-burner.  Unifying might even go hand-in-hand with micro-optimizing the kernel launch in the end.

PR #9705, opened 2026-02-17
MAINT,ENH: Refactor kernel argument handling into `KernelArguments` class

seberg:
This is a bit related to gh-9586 and the previous attempt on it in gh-2920. I.e. to make ElementWise kernels and ufuncs (and reductions) more similar one part of the old PR was also to introduce a new `Args` class to encapsulate kernel arguments.

This PR introduces a `KernelArguments` and merges the broadcasting logic into `find_and_apply_shape()`.
There is still quite a bit of duplication, but it helps to make the main `__call__` paths easier to follow, I think.  I opted for `.get_in(0)` getters for many things, because I would like to hide a bit how the arguments are stored.

---

Should reduce overheads for simple ufuncs by around 5-10%, but for me it is more about helping with follow-ups on merging the ufunc/elementwise kernel code or just moving more things into `KernelArguments`.

Two concrete thoughts I have here is that `KernelArguments` could already create `CArray`s, which could give us a pretty big boost when broadcasting or `reduce_dims()` happens (avoid creating more `ndarray`s).
Another thought I have is to speed up `ArgInfo`s a bit by merging them (the cache lookup takes quite a lot longer than it should).

### DIFF
--- a/cupy/_core/_kernel.pxd
+++ b/cupy/_core/_kernel.pxd
@@ -87,7 +87,6 @@ cdef class _TypeMap:
     cdef str get_typedef_code(self, type_headers=*)
 
 
-
 cdef class KernelArguments:
     # A kernels friendly helper to organize preparation of input and
     # output arguments.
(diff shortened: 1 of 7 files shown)


seberg:
I'll push the milestone on this.  I'll probably still fix the conflicts, but may need needs an iteration or two and I doubt we'll get there before 14.3 with partial availability.
(If yes, then great, I just don't see go high enough on the priority queue.)

PR #9714, opened 2026-02-18
Allow using PCH via `CUPY_NVRTC_USE_PCH=1` and use it for tests

seberg:
Locally, this is just massively useful, the question is almost whether we should just always add it and have the env variable to disable it (users could also disable by setting the cache size to zero, but that seems less clean)?

Since this adds a compile option, it will lead to rebuilding of CI where PCH is supported. Will be interesting how much it speeds up.

### DIFF
--- a/.pfnci/generate.py
+++ b/.pfnci/generate.py
@@ -211,7 +211,9 @@ class LinuxGenerator:
             'ENV PATH "${PYENV_ROOT}/shims:${PYENV_ROOT}/bin:${PATH}"',
             f'RUN pyenv install {py_spec} && \\',
             f'    pyenv global {py_spec} && \\',
-            '    pip install -U setuptools pip wheel',
+            '    pip install -U setuptools pip wheel && \\',
+            # For GCP kernel cache backend
+            '    pip install -U google-cloud-storage',
             '',
         ]
 
--- a/.pfnci/linux/main-flexci.sh
+++ b/.pfnci/linux/main-flexci.sh
@@ -28,7 +28,7 @@ STAGES="cache_get build test"
 if [[ "${TARGET}" == "benchmark" ]]; then
     STAGES="cache_get build benchmark"
 fi
-BENCHMARK_DIR=/tmp/benchmark CACHE_DIR=/tmp/cupy_cache PULL_REQUEST="${pull_req}" "$(dirname ${0})/run.sh" "${TARGET}" "${STAGES}" 2>&1 | tee "${LOG_FILE}"
+BENCHMARK_DIR=/tmp/benchmark CACHE_DIR=/tmp/cupy_cache CACHE_KERNEL_TO_GCS=1 PULL_REQUEST="${pull_req}" "$(dirname ${0})/run.sh" "${TARGET}" "${STAGES}" 2>&1 | tee "${LOG_FILE}"
 test_retval=${PIPESTATUS[0]}
 
 echo "****************************************************************************************************"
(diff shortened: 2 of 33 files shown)


seberg:
I am too curious about what it does to CI times:

/test cuda130

**EDIT: oops, I don't think this causes cache misses at all.  But that is actually for the better behavior wise...**

leofang:
hmmm, doesn't seem to help Windows much...?

seberg:
Yeah, it had no effect at all. For linux I think that may just be the case because the caches are hot (the option is added after hash key creation).
But for windows... it either has no effect here or the slowness has nothing to do with disk cache misses!?
(I mean, it could still be the disk cache being extremely slow itself or so, but not a miss?  I am almost curious what happens if I rerun it _without_ the disk cache here...)

**EDIT: Sorry, nvm. This windows run failed to download the cache, it really seems like there was no effect on at least the windows run. (compared to similar runs also stops at 60-70% of tests)**

seberg:
I am curious about what happens with `CUPY_CACHE_IN_MEMORY`.  (may just end up being a silly try)

/test cuda130

seberg:
One observation locally that surprises me is that `CUPY_CACHE_IN_MEMORY=1` seems a decent amount (~1.5x) slower to run the tests than running with an initially empty file cache.
I.e. I expected that all identical code is also memoized by:
```
@_util.memoize()
def _get_elementwise_kernel_code(...)
```
although ufuncs don't hit that cache -- but it seems surprising to have a mass of ufuncs that miss their per-Op cache but hit the file cache.
EDIT: `long` vs. `long long` can be one reason, but I guess the main one may be `RawKernel` (at least for the core tests with `test_raw.py`, plus `test_raw.py` cleans up all memoizations.).

But... my later tries here were with `CUPY_CACHE_IN_MEMORY=1`, so it might explain at least a bit why things are just not as massive as I may have hoped (if PCH is just ~2x in CI but 1.5x slower due to the file-cache effect, the expected speedup is only 1.5x...).

I want to confirm windows works, but I suspect the thing to do here might be just to merge and see what happens.

seberg:
OK, I removed all the experiments.  I am not exactly sure, but I think it may well be that this will get us to a full cache being rebuild in 2 CI runs rather than 3.

Unless we want to lower API/expose the cache directory options?  (E.g. here, auto-PCH takes care of the fact that `options` change and we need multiple PCH files.)

I suppose the main question is still whether PCH use should be the default (in some setups?) because it is very much worthwhile when we have to compile a lot of kernels.

cupy/cuda/compiler.py
```diff
@@ -379,6 +379,10 @@ def _compile(
                 # Starting with CUDA 12.0, even without using jitify, some
                 # tests cause an error if the following option is not included.
                 options += ('--device-as-default-execution-space',)
+            if ((major_version >= 13 or
+                    (major_version == 12 and minor_version >= 8)) and
+                    int(os.environ.get('CUPY_NVRTC_USE_PCH', 0))):
```
leofang: Nit: can we cache this query instead of checking the env var every time?

cupy/cuda/compiler.py
```diff
@@ -379,6 +379,10 @@ def _compile(
                 # Starting with CUDA 12.0, even without using jitify, some
                 # tests cause an error if the following option is not included.
                 options += ('--device-as-default-execution-space',)
```
leofang: nit: not related to this PR, but this can now be set unconditionally since we dropped CUDA 11 support

leofang:
CUDA 12.0. failures were fixed by https://github.com/cupy/cupy/pull/9771, and 13.1 by #9782.

/test force-skip

PR #9734, opened 2026-02-20
ENH: Reduce overheads for flags and imports related to FFT

seberg:
The flags is worth it I think.  Avoiding the import is surprisingly worthwhile but of course also a micro-optimization, but for light-weight FFT calls a noticable one.
(The import is 160ns vs 30ns for the attribute lookups for me.)

---

Not like it is much of a difference, but it struck me in the profiling (array creation is a far bigger chunk for example).  And since I had done most of the change, let's just do it.

I think flags cythonizing just makes sense even if pretty minor too.

### DIFF
--- a/cupy/fft/_fft.py
+++ b/cupy/fft/_fft.py
@@ -62,7 +62,7 @@ def _cook_shape(a, s, axes, value_type, order='C'):
 
 
 def _convert_fft_type(dtype, value_type):
-    from cupy.cuda import cufft
+    cufft = cupy.cuda.cufft  # lazy import
 
     if value_type == 'C2C' and dtype == np.complex64:
         return cufft.CUFFT_C2C
@@ -82,7 +82,7 @@ def _convert_fft_type(dtype, value_type):
 
 def _exec_fft(a, direction, value_type, norm, axis, overwrite_x,
               out_size=None, out=None, plan=None):
-    from cupy.cuda import cufft
+    cufft = cupy.cuda.cufft  # lazy import
 
     fft_type = _convert_fft_type(a.dtype, value_type)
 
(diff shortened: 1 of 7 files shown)


cupy/fft/_fft.py
```diff
@@ -82,7 +82,7 @@ def _convert_fft_type(dtype, value_type):
 
 def _exec_fft(a, direction, value_type, norm, axis, overwrite_x,
               out_size=None, out=None, plan=None):
-    from cupy.cuda import cufft
+    cufft = cupy.cuda.cufft  # lazy import
```
leofang, line 85: Does the import style affect performance? 👀

leofang (dismissed):
Sorry I dropped the ball. LGTM, but I am curious what's the saving we gain from this PR? Do you recall a ballpark number?

leofang:
> (The import is 160ns vs 30ns for the attribute lookups for me.)

Oh I missed this. This is referring to the flag change?

seberg:
> Oh I missed this. This is referring to the flag change?

That is referring to `from cupy.cuda import cufft` (160 ns for the import) vs. `cupy.cuda.cufft` (30ns for two attribute lookups).
I don't recall how much the difference was for the flags :/.

Let me close/reopen for readthedocs and then test.

(I would like to use this lazy import pattern more, but there was also a free-threading bug on Python that is probably fixed by now, I don't think it's a rush for v14.1.)

seberg:
/test mini

EDIT: heh, the test failures are real, there is a `cupyx/scipy` test that sets `arr.flags.writeable` which works for a Python class but doens't work now unless I add a `__dict__`.

seberg:
Added a `__dict__` for now.  The difference for something like `arr.flags.c_contiguous` on my computer is about `104->44ns` (without the dict it's 10ns faster, but I don't want to push that through this shortly, I think).

/test mini

PR #9736, opened 2026-02-21
DOC: Fix `scipy.linalg.*` comparison table

kmaehashi:
When generating a comparison table, we've been assuming `numpy.linalg.*` functions are directly exposed under `scipy.linalg.*`, but it seems no longer the case (https://github.com/scipy/scipy/pull/24085). Thus all `scipy.linalg.*` functions must be enumerated.

This PR also made the generator to self-executable to make testing easier.


kmaehashi:
Doc rendered as expected: https://cupy--9736.org.readthedocs.build/en/9736/reference/comparison.html
/test skip

Issue #9665, opened 2026-02-05
CI: Improve how kernel caches are updated/pruned


leofang:
Today the CI kernel cache works as follows:
- Each branch (ex: `main`, `v14`, ...) has its own cache per CUDA major.minor version. Every PR merge triggers the CI pipeline against the target branch: In the CI we load the cache from GCP, run tests, and store the (updated) cache back to GCP.
- In every PR, the base cache is the target branch's cache, plus additional kernels generated/updated by code changes in the PR. The per-PR cache aggregates both base and additions and is stored in GCP.

The problems: 
1. If we are rebuilding the whole world in parallel (see #9664 for why massive cache invalidation could happen), the per-PR caches blow up, potentially running out of disk space in the CI runners.
2. By the time the CI is green and the PR is merged, this per-PR cache is thrown away, instead of merged into the branch cache immediately. So in the target branch we rebuild the cache again. When the said cache invalidation happens, this doubles the costs in time and CI resource.

Possible solution(s):
- It might make sense to merge the per-PR cache into that of the target branch to speed up the cache rebuild process.
- If there is a way to do a more fine-grained CI retry, it'd save more resources. Some pipelines require more retries than others. (Currently, every retry launches the full CI, since we use GHA to call the dispatcher.)

leofang:
Another possible solution: Instead of storing the whole cache as a single GB-size tarball, we use GCP object store at the per-cubin level. Then, there is no need to worry about per-PR cache versus branch cache, because they all share the same store. We also don't need to pack/unpack tarballs, which on Windows can be extremely slow (~1hr) and also lead to the cache-merging challenge. @kmaehashi is kindly exploring this solution: https://github.com/kmaehashi/cupy/issues/83 🙂

kmaehashi:
> * It might make sense to merge the per-PR cache into that of the target branch to speed up the cache rebuild process.

I once considered and rejected this idea because this will easily increase the size of the branch cache. (Suppose a PR received `/test mini` comments three times, each invalidating all kernel caches...)

> * If there is a way to do a more fine-grained CI retry, it'd save more resources. Some pipelines require more retries than others. (Currently, every retry launches the full CI, since we use GHA to call the dispatcher.)

I also want to have this (it's convenient aside from cache issue). Maybe argument-like options? e.g. `/test --failed` (to retrigger failed CIs only) or `/test --pattern *linux*` should be doable.

leofang:
[...]

PR #9737, opened 2026-02-21
Implement kernel cache save/load abstraction

kmaehashi:
Part of #9665. Introduces an abstraction layer to kernel cache.

This PR is just a refactoring. I'll introduce GCP-backed cache using this API in a separate PR to make sure this refactoring does not break anything :)

Disclosure: The initial implementation was done by Copilot in https://github.com/kmaehashi/cupy/pull/84.

### DIFF
--- a/tests/cupy_tests/core_tests/test_raw.py
+++ b/tests/cupy_tests/core_tests/test_raw.py
@@ -354,11 +354,11 @@ extern "C" __global__ void my_func(void* input, int N) {
 def use_temporary_cache_dir():
     # Note uses mock, so not thread-safe (except at class/method level)
     # tempdir fixture could be used instead.
-    target1 = 'cupy.cuda.compiler.get_cache_dir'
+    target1 = 'cupy.cuda.compiler._kernel_cache_backend._cache_dir'
     target2 = 'cupy.cuda.compiler._empty_file_preprocess_cache'
     temp_cache = {}
     with tempfile.TemporaryDirectory() as path:
-        with mock.patch(target1, lambda: path):
+        with mock.patch(target1, path):
             with mock.patch(target2, temp_cache):
                 yield path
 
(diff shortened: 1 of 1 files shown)


cupy/cuda/_compiler_cache.py
```diff
@@ -0,0 +1,153 @@
...
+    def __init__(self, cache_dir: str | None = None):
+        """Initialize the disk cache backend.
+
+        Args:
+            cache_dir (str, optional): Directory to store cache files.
+                Defaults to CUPY_CACHE_DIR environment variable or
+                ~/.cupy/kernel_cache.
+        """
+        if cache_dir is None:
+            cache_dir = os.environ.get('CUPY_CACHE_DIR', _default_cache_dir)
```
seberg, line 86: Looking through the code, it seems to me that the `cache_dir=` parameter to the compile functions is never used and now effectively ignore.
(It doesn't seem to be part of any public API, though.)

I think that is totally fine (would be nice to remove it), but there are a few thats that "mock" `cupy.cuda.compiler.get_cache_dir` (raw and `test_callback.py`).

kmaehashi:
> Almost makes me wonder if cache-in-memory should just be a no-op backend.

Currently cache-in-memory is only effective for NVRTC compilation (ie ignored for NVCC compilation), so I think this needs some consideration beyond refactoring.

> The one comment I have is that there are a couple of tests that actually modify the cache folder and they do it by monkey-patching `get_cache_dir` and the code actually checks that every time. The tests should be easy to change to use monkey-patch/mock `_kernel_cache_backend` instead. As for the `cache_dir=` kwarg, I suspect it isn't public and could actually just be removed everywhere?

I've removed `cache_dir` arg and updated tests accordingly.

seberg (approved):
LGTM, thanks! Technically, `get_cache_dir()` might have been public, but it seems so niche/useless, I am not even sure it is worth noting in the upgrade guide.

kmaehashi:
I think it's unnecessary to treat this no-compat as it has never been a documented API. I looked through code search result for `get_cache_dir` (https://github.com/search?q=%22cupy%22+%22compiler%22+%22get_cache_dir%22+language%3Apython&type=code&p=1), and most usages are fork repos or vendored copies of CuPy. The only exception is cupyimg (https://github.com/mritools/cupyimg/blob/d64168280b2b46a7e2aaf8d16f4b6b07d6554f82/cupyimg/_misc.py#L94) but I assume no one will try to use it with CuPy v14. I also see no usage under rapidsai org.

kmaehashi:
Test failures are due to timeout.

/test force-skip

PR #9738, opened 2026-02-21
CI: Use GCP-backed kernel cache in Windows CI

kmaehashi:
Part of #9665. Based on #9737, this PR introduces GCP-backed kernel cache to Windows CI, which will be activated if `CUPY_CI_ENABLE_GCP_KERNEL_CACHE=1` env var is set when invoking pytest.

Disclosure: The initial implementation was done by Copilot in https://github.com/kmaehashi/cupy/pull/84.

----

Observations:
* First run (no cache available): https://ci.preferred.jp/cupy.win.cuda130/210536/ -> timed out around 38%
* Second run: https://ci.preferred.jp/cupy.win.cuda130/210537/ -> timed out around 65%
* Third run: https://ci.preferred.jp/cupy.win.cuda130/210538/ -> All tests completed (ie cache fully populated)
* Fourth run: https://ci.preferred.jp/cupy.win.cuda130/210543/ -> All tests completed in 2h13m. (roughly 13m for build, 2h for unit test)

PR #9739, opened 2026-02-23
CI: Use GCP-backed kernel cache in Linux CI

kmaehashi:
Linux CI counterpart of #9738. Note that, unlike Windows CI, tarball download/upload is still required for ccache. (Maybe we can migrate to sccache, but will leave it as future work for now)

~~6624775eb71253ab5e4b04743b168e33905cc5a6 needs to be reverted before merging.~~ **Update**: Done.

### DIFF
--- a/.pfnci/linux/tests/actions/cleanup.sh
+++ b/.pfnci/linux/tests/actions/cleanup.sh
@@ -20,8 +20,13 @@ find "${CUPY_CACHE_DIR}" -type f \
     -name "jitify_*.json" \
     -atime +30 -mtime +30 -ctime +30 -print -delete | wc -l
 
-# Purge all local *.cubin cache (they are stored separately via GCS kernel cache backend)
-python3 .pfnci/trim_cupy_kernel_cache.py --max-size 0 --rm
+if [[ "${CUPY_CI_ENABLE_GCP_KERNEL_CACHE:-0}" == "1" ]]; then
+    # Purge all local *.cubin cache (they are stored separately via GCS kernel cache backend)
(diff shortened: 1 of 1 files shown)


.pfnci/linux/tests/actions/cleanup.sh
```diff
@@ -20,8 +20,8 @@ find "${CUPY_CACHE_DIR}" -type f \
     -name "jitify_*.json" \
     -atime +30 -mtime +30 -ctime +30 -print -delete | wc -l
 
-# Expire *.cubin cache in 30 days, and also limit by total size.
-python3 .pfnci/trim_cupy_kernel_cache.py --max-size $((3*1024*1024*1024)) --expiry $((30*24*60*60)) --rm
+# Purge all local *.cubin cache (they are stored separately via GCS kernel cache backend)
+python3 .pfnci/trim_cupy_kernel_cache.py --max-size 0 --rm
```
leofang: Let's purge this only when GCP is enabled:
```suggestion
if [[ "${CUPY_CI_ENABLE_GCP_KERNEL_CACHE:-0}" == "1" ]]; then
    # Purge all local *.cubin cache (they are stored separately via GCS kernel cache backend)
    python3 .pfnci/trim_cupy_kernel_cache.py --max-size 0 --rm
else
    # Expire *.cubin cache in 30 days, and also limit by total size.
    python3 .pfnci/trim_cupy_kernel_cache.py --max-size $((3*1024*1024*1024)) --expiry $((30*24*60*60)) --rm
fi
```

leofang (approved):

PR #9748, opened 2026-02-25
Fix hip mask errors

### LINKED ISSUE #9742: ROCm 7.x: Boolean indexing fails to compile on CuPy 14.0.1 due to 32-bit mask passed to `__shfl_*_sync`
## Summary

On ROCm 7.x, CuPy 14.0.1 fails at runtime when performing boolean indexing due to HIPRTC compilation failure. The generated kernel passes a 32-bit literal mask (`0xffffffff`) to `__shfl_*_sync`, but HIP requires a 64-bit mask type.  Expected behavior is that boolean indexing operations succeed without any HIPRTC compilation errors.

This affects both RDNA3 and CDNA3 GPUs.

## Observed Example Error

```
In file included from <built-in>:1:
/tmp/comgr-15736-12-b48b05/include/hiprtc_runtime.h:8136:62: error: static assertion failed due to requirement 'sizeof(unsigned int) == 8': The mask must be a 64-bit integer. Implicitly promoting a smaller integer is almost always an error.
 8136 |   static_assert(__hip_internal::is_integral<MaskT>::value && sizeof(MaskT) == 8,
      |                                                              ^~~~~~~~~~~~~~~~~~
/tmp/comgr-15736-12-b48b05/input/tmp/tmpok6ksj1i/16ca1c7ca24b14a20631daa84c0c853cde516b85.hsaco.cu:26:18: note: in instantiation of function template specialization '__shfl_xor_sync<unsigned int, int>' requested here
   26 |             x += __shfl_xor_sync(0xffffffff, x, j, 64);
      |                  ^
/tmp/comgr-15736-12-b48b05/include/hiprtc_runtime.h:8136:76: note: expression evaluates to '4 == 8'
 8136 |   static_assert(__hip_internal::is_integral<MaskT>::value && sizeof(MaskT) == 8,
      |                                                              ~~~~~~~~~~~~~~^~~~
1 error generated when compiling for gfx942.
```
Full traceback omitted here for brevity; it originates from:
```
_cupy_bsum_shfl → scan → boolean indexing
```
and fails during HIPRTC compilation.

### To Reproduce

Save reproducer as `repro_bool_index.py`:
```py
import cupy as cp

x = cp.arange(2048, dtype=cp.int32)
mask = (x & 1) == 0  # bool mask, half True

y = x[mask]

print("y.shape:", y.shape, "y[:5]:", y[:5].get() if hasattr(y, "get") else y[:5])
```

Clear kernel cache before running:
```bash
rm -rf ~/.cupy/kernel_cache
env ROCR_VISIBLE_DEVICES=[device index here] ROCM_HOME="/opt/rocm" python repro_bool_index.py
```

### Installation

Built from GitHub source

### Environment

```
OS                        : Linux-5.15.0-170-generic-x86_64-with-glibc2.35
Python Version            : 3.10.12
CuPy Version              : 14.0.1
CuPy Platform             : AMD ROCm
NumPy Version             : 2.2.6
SciPy Version             : 1.15.3
Cython Build Version      : 3.1.8
Cython Runtime Version    : 3.1.8
CUDA Root                 : /usr
HIPCC PATH                : /usr/bin/hipcc
CUDA Build Version        : 70226015
CUDA Driver Version       : 70226015
CUDA Runtime Version      : 70226015 (linked to CuPy) / 70226015 (locally installed)
CUDA Extra Include Dirs   : None
cuBLAS Version            : 50200
cuFFT Version             : 10036
cuRAND Version            : 400200
cuSOLVER Version          : (3, 32, 0)
cuSPARSE Version          : 400200
NVRTC Version             : (9, 0)
Thrust Version            : 0
CUB Build Version         : 400200
Jitify Build Version      : None
NCCL Build Version        : 22707
NCCL Runtime Version      : 22707
cuTENSOR Version          : None
cuSPARSELt Build Version  : None
Device 0 Name             : AMD Instinct MI300X VF
Device 0 Arch             : gfx942:sramecc+:xnack-
Device 0 PCI Bus ID       : 0000:ff:00.0
```

### Additional Information

Above reproducer python script was tested on:
* CuPy 14.0.1 PyPI prebuilt wheel with ROCm 7.0.2 system libraries
* CuPy 14.0.1 built from source first with ROCm 7.1.1 on local system, and then with ROCm 7.2.0 in cloud MI300X VM (with the necessary adjustments from [#9622](https://github.com/cupy/cupy/pull/9622) to build against ROCm 7.2.0)

GPUs:
* RX 7600 XT (gfx1102; PyPI wheel tested using `HSA_OVERRIDE_GFX_VERSION=11.0.0`)
* MI300X (gfx942)

Both GPUs fail in the same way.

---

## Root Cause

CuPy generates shuffle intrinsics like:

```cpp
__shfl_xor_sync(0xffffffff, x, j, warp_size);
```

On HIP (ROCm 7.x), `_sync` shuffle intrinsics require the mask argument to be a 64-bit integer type. Passing a 32-bit literal (`0xffffffff`) causes a compile-time static assertion failure in HIPRTC.

This occurs in at least:

* `_cupy_bsum_shfl` (used by `scan` in `cupy/_core/_routines_math.pyx`)
* `_nonzero_kernel_incomplete_scan` (in `cupy/_core/_routines_indexing.pyx`)

Other shuffle-based kernels may be similarly affected.

---

## Notes

* This does not appear to be specific to a single ROCm minor version.
* It affects boolean indexing (`x[mask]`) and any code path that triggers `scan`.
* CUDA builds do not appear to be affected.

gpinkert:
There are many crashes related to using an incorrect parameter for some of the shfl intrinsics when setting a full mask. This PR adds support for hip by defining a full mask when the warp/wavefront size is 64.

### DIFF
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -64,7 +64,7 @@ else:
     _full_mask = 0xffffffff
 
 
-def _full_mask_hex():
+cpdef str _full_mask_hex():
     if runtime._is_hip_environment:
         return hex(_full_mask) + 'ULL'
     return hex(_full_mask)
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -55,19 +55,24 @@ def _get_warpsize():
 _is_hip_7_plus = (runtime._is_hip_environment
                   and _driver.get_build_version() >= 7_00_00000)
 
-# Full-lane mask: value is warp-size-dependent on HIP, always 32-bit on CUDA.
-# On HIP the mask *type* must be 64-bit (ROCm 7+ enforces via static_assert),
-# so _full_mask_hex() appends 'ULL' to the C literal.
-if runtime._is_hip_environment:
-    _full_mask = (1 << _get_warpsize()) - 1
-else:
-    _full_mask = 0xffffffff
 
-
-cpdef str _full_mask_hex():
+# Full-lane mask: warp-size-dependent on HIP, always 32-bit on CUDA.
+# Computed lazily (per device) so that importing CuPy does not require a
+# functional ROCm/CUDA runtime.
+@_util.memoize(for_each_device=True)
+def _full_mask():
     if runtime._is_hip_environment:
-        return hex(_full_mask) + 'ULL'
-    return hex(_full_mask)
+        return (1 << _get_warpsize()) - 1
+    return 0xffffffff
+
+
+# On HIP the mask *type* must be 64-bit (ROCm 7+ enforces via static_assert),
+# so the C literal is suffixed with 'ULL'.
+cpdef str _full_mask_hex():
+    cdef str s = hex(_full_mask())
+    if runtime._is_hip_environment:
+        return s + 'ULL'
+    return s
 
 
 cdef str _get_simple_elementwise_kernel_code(
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -28,6 +28,7 @@ from cupy._core.core cimport compile_with_cache
 from cupy._core.core cimport _ndarray_base
 from cupy._core cimport internal
 from cupy_backends.cuda.api cimport runtime
+from cupy_backends.cuda.api import driver as _driver
 
 try:
     from cupy_backends.cuda.libs import cutensor as cuda_cutensor
@@ -48,6 +49,25 @@ cdef inline bint _contains_zero(const shape_t& v) except? -1:
 def _get_warpsize():
     device_id = runtime.getDevice()
     return runtime.getDeviceProperties(device_id)['warpSize']
+
+
+# ROCm 7+ requires a 64-bit mask type for __shfl_*_sync / __any_sync.
+_is_hip_7_plus = (runtime._is_hip_environment
+                  and _driver.get_build_version() >= 7_00_00000)
+
+# Full-lane mask: value is warp-size-dependent on HIP, always 32-bit on CUDA.
+# On HIP the mask *type* must be 64-bit (ROCm 7+ enforces via static_assert),
+# so _full_mask_hex() appends 'ULL' to the C literal.
+if runtime._is_hip_environment:
+    _full_mask = (1 << _get_warpsize()) - 1
+else:
+    _full_mask = 0xffffffff
+
+
+def _full_mask_hex():
+    if runtime._is_hip_environment:
+        return hex(_full_mask) + 'ULL'
+    return hex(_full_mask)
 
 
 cdef str _get_simple_elementwise_kernel_code(
--- a/cupy/_core/_routines_indexing.pyx
+++ b/cupy/_core/_routines_indexing.pyx
@@ -6,7 +6,8 @@ import numpy
 import cupy
 import cupy._core.core as core
 from cupy.exceptions import AxisError
-from cupy._core._kernel import ElementwiseKernel, _get_warpsize, _full_mask_hex
+from cupy._core._kernel import ElementwiseKernel, _get_warpsize
+from cupy._core._kernel cimport _full_mask_hex
 from cupy._core._ufuncs import elementwise_copy
 
 from libcpp cimport vector
--- a/cupy/_core/_routines_indexing.pyx
+++ b/cupy/_core/_routines_indexing.pyx
@@ -6,7 +6,7 @@ import numpy
 import cupy
 import cupy._core.core as core
 from cupy.exceptions import AxisError
-from cupy._core._kernel import ElementwiseKernel, _get_warpsize
+from cupy._core._kernel import ElementwiseKernel, _get_warpsize, _full_mask_hex
 from cupy._core._ufuncs import elementwise_copy
 
 from libcpp cimport vector
@@ -490,7 +490,7 @@ def _nonzero_kernel_incomplete_scan(block_size, warp_size=32):
         S x = 0;
         if (i < a.size()) x = a[i];
         for (int j = 1; j < ${warp_size}; j *= 2) {
-            S tmp = __shfl_up_sync(0xffffffff, x, j, ${warp_size});
+            S tmp = __shfl_up_sync(${full_mask}, x, j, ${warp_size});
             if (lane_id - j >= 0) x += tmp;
         }
         if (lane_id == ${warp_size} - 1) smem[warp_id] = x;
@@ -499,7 +499,7 @@ def _nonzero_kernel_incomplete_scan(block_size, warp_size=32):
             S y = 0;
             if (lane_id < n_warp) y = smem[lane_id];
             for (int j = 1; j < n_warp; j *= 2) {
-                S tmp = __shfl_up_sync(0xffffffff, y, j, ${warp_size});
+                S tmp = __shfl_up_sync(${full_mask}, y, j, ${warp_size});
                 if (lane_id - j >= 0) y += tmp;
             }
             int block_id = i / ${block_size};
@@ -510,7 +510,7 @@ def _nonzero_kernel_incomplete_scan(block_size, warp_size=32):
         }
         __syncthreads();
         x += smem[warp_id];
-        S x0 = __shfl_up_sync(0xffffffff, x, 1, ${warp_size});
+        S x0 = __shfl_up_sync(${full_mask}, x, 1, ${warp_size});
         if (lane_id == 0) {
             x0 = smem[warp_id];
         }
@@ -523,7 +523,8 @@ def _nonzero_kernel_incomplete_scan(block_size, warp_size=32):
                 j = j_next;
             }
         }
-    """).substitute(block_size=block_size, warp_size=warp_size)
+    """).substitute(block_size=block_size, warp_size=warp_size,
+                    full_mask=_full_mask_hex())
     return cupy.ElementwiseKernel(in_params, out_params, loop_body,
                                   'cupy_nonzero_kernel_incomplete_scan',
                                   loop_prep=loop_prep)
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -5,7 +5,8 @@ import numpy
 
 import cupy
 from cupy._core._reduction import create_reduction_func
-from cupy._core._kernel import create_ufunc, _get_warpsize, _full_mask_hex
+from cupy._core._kernel import create_ufunc, _get_warpsize
+from cupy._core._kernel cimport _full_mask_hex
 from cupy._core._scalar import get_typename
 from cupy._core._ufuncs import elementwise_copy
 import cupy._core.core as core
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -5,7 +5,7 @@ import numpy
 
 import cupy
 from cupy._core._reduction import create_reduction_func
-from cupy._core._kernel import create_ufunc, _get_warpsize
+from cupy._core._kernel import create_ufunc, _get_warpsize, _full_mask_hex
 from cupy._core._scalar import get_typename
 from cupy._core._ufuncs import elementwise_copy
 import cupy._core.core as core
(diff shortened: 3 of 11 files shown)


cupy/_core/_routines_indexing.pyx
```diff
@@ -476,6 +477,12 @@ cdef tuple _view_getitem(_ndarray_base a, list slice_list):
     return v, 0
 
 
+# ROCm 7.2+ requires a 64-bit mask type for __shfl_*_sync / __any_sync.
+# ~0ULL sets all bits, covering any wavefront size (32 or 64).
+_full_mask = ('~0ULL'
+              if _runtime.is_hip else '0xffffffff')
```
leofang: Let's consolidate these in e.g. `cupy/_core/_kernel.pyx` or somewhere else, and import from there. For example, there is already a `_get_warpsize`.

leofang: btw for my own curiosity: 
1. Is this change universal to all ROCm-supported devices, or is it device-dependent?   
2. Does this change break older (<7.2) ROCm users?

cupy/_core/_routines_math.pyx
```diff
@@ -155,6 +155,11 @@ cdef _ndarray_base _ndarray_clip(_ndarray_base self, a_min, a_max, out):
 _op_char = {scan_op.SCAN_SUM: '+', scan_op.SCAN_PROD: '*'}
 _identity = {scan_op.SCAN_SUM: 0, scan_op.SCAN_PROD: 1}
 
+# ROCm 7.2+ requires a 64-bit mask type for __shfl_*_sync / __any_sync.
+# ~0ULL sets all bits, covering any wavefront size (32 or 64).
+_full_mask = ('~0ULL'
+              if runtime._is_hip_environment else '0xffffffff')
```
leofang: ditto

cupyx/jit/_builtin_funcs.py
```diff
@@ -379,9 +380,13 @@ def call(self, env, mask, var, val_id, *, width=None):
...
         except Exception:
             raise TypeError('mask must be an integer')
-        if runtime.is_hip:
+
+        _hip_72_plus = (runtime.is_hip
+                        and driver.get_build_version() >= 7_02_00000)
+        if runtime.is_hip and not _hip_72_plus:
             warnings.warn(f'mask {mask} is ignored on HIP', RuntimeWarning)
-        elif not (0x0 <= mask <= 0xffffffff):
+        max_mask = 0xffffffffffffffff if _hip_72_plus else 0xffffffff
```
leofang: ditto

leofang (changes requested):

cupy/_core/_routines_indexing.pyx
gpinkert: 1) This is likely universal to all ROCm supported devices. I haven't tested others but for all warpsize 64 devices this is neccessary. 
2) No, so we don't have to gate on 7.2. It actually enables ROCm 7.0 >

gpinkert:
@leofang apologies for the delay - I hope you find the updates satisfactory.

alexschroeter:
@gpinkert Thanks for looking into this and for the notes on hipcim as I mentioned in the Issue I was surprised to learn that rocm/dev-ubuntu-22.04:latest comes with ROCm 6.2 and I didn't think about hipcim. I will fix that thank you.

Regarding this PR, I might have identified 2 more issues.

## 1. The 6.2 GAP

ROCm 7.0.2 and ROCm 6.1 pass all tests. ROCm 6.2 fails with use of undeclared identifier '__shfl_xor_sync' and __shfl_down_sync in the hipcc compilation path.

The root cause is a version gap. https://github.com/cupy/cupy/blob/b316c6d35dc483b1e75bf792463bc89854dde492/cupy/_core/include/cupy/hip_workaround.cuh#L8 On ROCm < 6.2, the workaround macros in hip_workaround.cuh are active and rewrite __shfl_*_sync(mask, ...) to __shfl(...), stripping the mask entirely, so any mask value works. On ROCm 7.0+, the native __shfl_*_sync functions are available and accept 64-bit masks. But ROCm 6.2 falls in between: the workaround macros are disabled (because HIP_VERSION >= 60200000) yet the native __shfl_*_sync symbols are not available in all compilation paths, specifically hipcc. So kernels compiled via hipcc on ROCm 6.2 fail to find these functions.

## 2. _is_hip_7_plus behaviour

The PR's _is_hip_7_plus version check doesn't address this gap. The _full_mask_hex() function also unconditionally appends ULL on all HIP versions, which means even if the function were found, the 64-bit literal would be passed on ROCm 6.2 where it may not be expected.

gpinkert:
@alexschroeter 

>I was surprised to learn that rocm/dev-ubuntu-22.04:latest comes with ROCm 6.2 

Yeah, so am I. I would recommend using the rocm-complete images. That way you know which version of ROCm you are using. 

>  the 64-bit literal would be passed on ROCm 6.2 where it may not be expected.

64 bits is always expected in the shfl intrinsics on ROCm platforms. Prior to ROCm 7 there just wasn't an error AFAIK. I think it had something to do with an implicit conversion that no longer happens.  

Also, thank you for pointing out the gap on ROCm 6.2. As it turns out, there is also a gap for 6.3 and 6.4. Also, the defines that gate those shfl functions change from ROCm 6 to 7 - it changes from opt in to opt out. I've updated the code to check for both macros depending on the ROCm versions. If you can see here, the change is working on ROCm 6.2:

image

Here, you can see the change working on ROCm 7: 

image

Also, is there a reason that you are using ROCm 6.2? I don't get a chance to talk to customers very often so this is exciting for me. I ask because I would recommend updating to latest ROCm if you can. Every latest version will come with a variety of bug fixes. We (by we, I mean internal cupy development at AMD) don't officially support ROCm 6 anymore AFAIK. Right now development efforts are focused on ROCm 7 and above, and we plan on supporting ROCm 7 a year after ROCm 8 is released. Seven will be much more stable for the foreseeable future. 

Anyways, thanks again for bringing this to our attention! 

@leofang, do you know when we can get this PR re-reviewed?

alexschroeter:
@gpinkert There is a reason some of my setup is pinned to 6.2 (see https://github.com/ROCm/ROCm/issues/5948). It is the only version that can currently run GROMACS without leaking memory. But regarding cupy, I simply didn't realize that the container came with ROCm 6.2, and yes, I will update to 7.2.

While I have your attention, I was wondering if you wouldn't mind taking a look at another issue I was facing. https://github.com/cupy/cupy/issues/9830

leofang:
Sorry I was OOTO until this week. Will try to catch up asap!

leofang:
In the meanwhile, @alexschroeter would you mind re-testing this PR locally and check if it fixes for you on ROCm 6.2?

cupy/_core/_kernel.pyx
```diff
@@ -50,6 +51,25 @@ def _get_warpsize():
...
+# Full-lane mask: value is warp-size-dependent on HIP, always 32-bit on CUDA.
+# On HIP the mask *type* must be 64-bit (ROCm 7+ enforces via static_assert),
+# so _full_mask_hex() appends 'ULL' to the C literal.
+if runtime._is_hip_environment:
+    _full_mask = (1 << _get_warpsize()) - 1
+else:
+    _full_mask = 0xffffffff
+
+
+def _full_mask_hex():
```
leofang: nit: let's make this function a `cpdef str _full_mask_hex` so that in Cython modules we can cimport it instead of import.

tests/cupyx_tests/jit_tests/test_raw.py
```diff
@@ -654,10 +654,11 @@ def test_shfl_down(self, dtype):
         N = 5
         # __shfl_down() on HIP does not seem to have the same behavior...
         block = cupy._core._get_warpsize()
+        full_mask = (1 << block) - 1
```
leofang: Q: can we avoid any recompilation now that we centralize it?

cupy/_core/_kernel.pyx
```diff
@@ -50,6 +51,25 @@ def _get_warpsize():
...
 
+# ROCm 7+ requires a 64-bit mask type for __shfl_*_sync / __any_sync.
+_is_hip_7_plus = (runtime._is_hip_environment
+                  and _driver.get_build_version() >= 7_00_00000)
+
+# Full-lane mask: value is warp-size-dependent on HIP, always 32-bit on CUDA.
+# On HIP the mask *type* must be 64-bit (ROCm 7+ enforces via static_assert),
+# so _full_mask_hex() appends 'ULL' to the C literal.
+if runtime._is_hip_environment:
+    _full_mask = (1 << _get_warpsize()) - 1
```
leofang: Looks like this breaks CI because ROCm is required functional at import time. In general, we defer initialization of CUDA or ROCm as late as possible, so that CuPy is importable in CPU-only projects. Let's make this value lazily initialized. 

In fact, after reading the change I don't think we need to keep this variable. The whole lazy init can be done in the `_full_mask_hex` below (need to add `@_util.memoize(for_each_device=True)`). This value is only needed by one Python test, which can be easily retrieved by `int(_full_mask_hex())`.

leofang (changes requested):

alexschroeter:
> In the meanwhile, @alexschroeter would you mind re-testing this PR locally and check if it fixes for you on ROCm 6.2?

Everything passes. From my side the issue is fixed.

| ROCm | Result |
| --- | --- |
|6.1 (Docker) | All 7 tests passed|
|6.2 (Docker) | All 7 tests passed|
|7.0.2 (host) | All 7 tests passed|

cupy/_core/_kernel.pyx
gpinkert: done

cupy/_core/_kernel.pyx
gpinkert: Done

tests/cupyx_tests/jit_tests/test_raw.py
gpinkert: I think in practice this probably doesn't buy us much? My understanding is that CuPy processes are usually pinned to a single backend (CUDA or HIP) by the wheel, and within a process the visible GPUs tend to be the same arch / warp size, so `_full_mask()` would resolve to a single value and the per-device memoize would end up with a one-entry cache. 

My understanding is the kernel module cache then gives us a one-time compile per kernel per process anyway (and I think the on-disk cache usually covers even that on subsequent runs?). So centralizing the helper wouldn't reduce compiles? so centralizing the helper might not actually reduce compiles, but I still think it's worth doing for consistency across call sites. 

Happy to be corrected if I'm wrong about any of that.

gpinkert:
@alexschroeter can you test again?

leofang (approved):
/test mini

tests/cupyx_tests/jit_tests/test_raw.py
leofang: No, it wouldn't help in terms of avoiding re-compilation. Sorry I wasn't making sense at all. I was probably just thinking about avoiding non-trivial expressions like `(1 << block) -1`, while multitasking 😅

alexschroeter:
> @alexschroeter can you test again?

Retested against HEAD (e783ef493) on all three ROCm versions — all 7 tests pass on ROCm 6.1, 6.2, and 7.0.2 (host). Confirmed _full_mask() returns 0xffffffffffffffff and _full_mask_hex() returns 0xffffffffffffffffULL on HIP.

leofang:
Hmmmm several CUDA CIs timed out. It is unclear to me why this innocent PR would invalidate CI caches... @gpinkert if possible would you mind taking a look? We try to avoid massive CI rebuilds unless absolutely necessary (it causes a lot of churns and real money). I would think a refactoring PR should be easy to test 😅

gpinkert:
@leofang sure, unsure when I can get time this week to look into it. Got assigned to a higher priority project in the short term. I don't know how much time I will have until after the 30th

leofang:
> Hmmmm several CUDA CIs timed out. It is unclear to me why this innocent PR would invalidate CI caches... 

Sorry no worries Grant, Sebastian pointed out to me that it's due to our cache key not being granular enough (and likely hard to make it so). This PR touches `cupy/_core/include/cupy/hip_workaround.cuh` which changes the cache key, but there is no other way to proceed so no need to sweat. We will get this through!

leofang:
The Windows test failure is irrelevant, let's get this in.

PR #9751, opened 2026-02-27
fix(hip): add clear error message for AQL work-item overflow on kernel launch

gpinkert:
When a raw kernel launch on HIP/ROCm exceeds the uint32_t limit of the HSA AQL dispatch packet (gridDimX * blockDimX > UINT32_MAX), the HIP runtime returns an opaque "hipErrorInvalidValue: invalid argument" with no indication of the cause.

Add a pre-launch check on HIP builds that detects the overflow and raises a ValueError with an actionable message explaining the thread limit and suggesting the user reduce the grid size and have each thread handle multiple elements in a loop.

### DIFF
--- a/cupy/cuda/function.pyx
+++ b/cupy/cuda/function.pyx
@@ -5,6 +5,7 @@ import warnings
 
 from libc.stdint cimport intptr_t
 from libc.stdint cimport uintmax_t
+from libc.stdint cimport UINT32_MAX
 from libcpp cimport vector
 
 from cupy._core cimport _carray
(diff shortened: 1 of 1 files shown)


cupy/cuda/function.pyx
```diff
@@ -5,6 +5,7 @@ import warnings
 
 from libc.stdint cimport intptr_t
 from libc.stdint cimport uintmax_t
+from libc.stdint cimport UINT32_MAX
```
leofang, line 8: I think we hit this issue (closed as won't fix): https://github.com/MarcoGorelli/cython-lint/issues/70. Let's try this:
```suggestion
from libc.stdint cimport UINT32_MAX  # no-cython-lint
```

PR #9752, opened 2026-02-27
BUG: Fix regression for 32bit index flag in `.real` and broadcast

### LINKED ISSUE #9750: _index_32_bits incorrectly set for non-contiguous views (e.g. .real on complex arrays >2^31 bytes)
### Description

I noticed that taking `.real` of a complex array that exceeds 2^31 bytes can lead to an illegal memory access. 

The reason for this seems to be, that cupy does not account for the strides when setting `_index_32_bits` of a view (see reproducing example). 

The bug seems to come from here:
https://github.com/cupy/cupy/blob/78bdcc3553c6ba0e1f65d3baedf25cbd5c58f107/cupy/_core/core.pyx#L269
https://github.com/cupy/cupy/blob/78bdcc3553c6ba0e1f65d3baedf25cbd5c58f107/cupy/_core/core.pyx#L275

In the case of taking `.real` of a `complex64`, itemsize will be 4, but the stride will be 8. Which can lead to `_index_32_bits` wrongfully being True.

**Possible Fix:**

Replace `alloc_size = self.size * itemsize` by `alloc_size = right - left`

I can open a PR if this fix is acceptable.

### To Reproduce

```py
import cupy as cp

# 268435456 * 8 bytes (complex64) = exactly 2^31 bytes → works
arr = cp.ones(268435456, dtype=cp.complex64)
arr_real = arr.real
print(f"_index_32_bits: {arr_real._index_32_bits}")  # True
res = arr_real.sum()
cp.cuda.Device().synchronize()
print("PASS: 2^31 bytes")

# 268435457 * 8 bytes = 2^31 + 8 bytes → crashes
arr = cp.ones(268435457, dtype=cp.complex64)
arr_real = arr.real
print(f"_index_32_bits: {arr_real._index_32_bits}")  # Will be True, but should be False
res = arr_real.sum()
cp.cuda.Device().synchronize()
print("PASS: 2^31 + 8 bytes")  # never reached
```

### Installation

### Environment

```
OS                           : Linux-6.14.0-37-generic-x86_64-with-glibc2.39
SciPy Version                : 1.17.1
Cython Build Version         : 3.1.4
NVCC PATH                    : /usr/local/cuda/bin/nvcc
CUDA Driver Version          : 12080
CUDA Runtime Version         : 12090 (linked to CuPy) / 12080 (locally installed)
CUDA Runtime Path            : /usr/local/cuda/lib64/libcudart.so.12 (locally installed)
CUDA Extra Include Dirs      : ['/usr/local/cuda/include']
cuBLAS Version               : 120803
cuBLAS Path                  : /usr/local/cuda/lib64/libcublas.so.12
cuFFT Version                : 11303
cuFFT Path                   : /usr/local/cuda/lib64/libcufft.so.11
cuRAND Path                  : /usr/local/cuda/lib64/libcurand.so.10
cuSOLVER Version             : (11, 7, 2)
cuSOLVER Path                : /usr/local/cuda/lib64/libcusolver.so.11
cuSPARSE Path                : /usr/local/cuda/lib64/libcusparse.so.12
NVRTC Path                   : /usr/local/cuda/lib64/libnvrtc.so.12
NCCL Build Version           : 22501
NCCL Path                    : /lib/x86_64-linux-gnu/libnccl.so.2
Device 0 Name                : NVIDIA GeForce RTX 5090
Device 0 Compute Capability  : 120

```

### Additional Information

_No response_

seberg:
I suspect I misunderstood the flag, as the offset is used to calculate the pointer offset as well and not just for index manipulation.

Adding a few tests, I noticed that broadcasting also got things wrong, one could also proof it via something like:
```
arr = cp.broadcast_arrays(cp.random.randint(255, size=2**30+100, dtype=cp.uint8), cp.empty((2, 1)))[0]
cp.put(arr, cp.arange(arr.size//2, arr.size), 1)
print(arr)
```

I audited places for `._view()` and `_set_shape_and_strides()`, but I am a bit wondering whether it this couldn't use a bigger refactor

Maybe I should refactor this into a flag to `_view` and `_set_shape_and_strides`?

Closes gh-9750

### DIFF
--- a/cupy/_core/_routines_manipulation.pyx
+++ b/cupy/_core/_routines_manipulation.pyx
@@ -499,6 +499,8 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
     view = array.view()
     # TODO(niboshi): Confirm update_x_contiguity flags
     view._set_shape_and_strides(_shape, strides, True, True)
+    if view.size * view.itemsize > 2**31:
+        view._index_32_bits = False
     return view
 
 
--- a/cupy/_core/core.pxd
+++ b/cupy/_core/core.pxd
@@ -15,7 +15,8 @@ cdef class _ndarray_base:
         public strides_t _strides
         readonly bint _c_contiguous
         readonly bint _f_contiguous
-        # To do fast indexing in the CArray class
+        # Whether the array memory can be addressed with 32bit signed
+        # integers. To do fast indexing in the CArray class.
         readonly bint _index_32_bits
         readonly object dtype
         readonly memory.MemoryPointer data
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -266,11 +266,13 @@ cdef class _ndarray_base:
                         f"contiguous size of {self.size * itemsize} but got "
                         f"{alloc_size}.")
             else:
-                alloc_size = self.size * itemsize
+                alloc_size = right - left  # actual memory extend
         else:
             self._set_contiguous_strides(itemsize, order_char == b'C')
-            alloc_size = self.size * itemsize
+            alloc_size = self.size * itemsize  # contiguous
 
+        # Arrays can e.g. broadcast, so check that all pointer offsets fit and
+        # the size fit int32. (seberg: `* itemsize` may be unnecessary)
         max_diff = max(alloc_size, self.size * itemsize)
         self._index_32_bits = max_diff <= <Py_ssize_t>(1 << 31)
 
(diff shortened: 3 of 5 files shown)


leofang (dismissed):

seberg:
Arg, my local GPU had way too much memory, the parametrization must work lazily with such large allocations.

/test mini

cupy/_core/_routines_manipulation.pyx
```diff
@@ -500,6 +500,8 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
     view = array.view()
     # TODO(niboshi): Confirm update_x_contiguity flags
     view._set_shape_and_strides(_shape, strides, True, True)
+    if view.size * view.itemsize > 2**31:
```
leofang, line 503: Q: `>` or `>=`?

seberg: `>` is right (sizes not a a index). Other code uses `<=` to match:
```
self._index_32_bits = max_diff <= <Py_ssize_t>(1 << 31)
```

leofang (approved):

PR #9757, opened 2026-03-02
BUG: Introduce a "promotion" step and fix integer comparisons

### LINKED ISSUE #9753: Comparison between `uint64` max value and `uint64` array result in `OverflowError`
### Description

Comparison operations (e.g. `==`, `<`, ...) between the maximum `uint64` value represented as a Python `int` and a CuPy `uint64` array result in an `OverflowError`. In fact, the same error occurs for all Python `int`s greater than the `int64` maximum value, so I suspect that the Python `int` goes through a `int64` conversion (instead of `uint64` conversion) before the comparison is performed. Manually converting to `uint64` array first works fine.

### To Reproduce

```py
import cupy as cp

(cp.iinfo(cp.uint64).max) == cp.asarray(1, dtype=cp.uint64)
# OverflowError: int too big to convert

(cp.iinfo(cp.uint64).max // 2 + 1) == cp.asarray(1, dtype=cp.uint64)
# OverflowError: int too big to convert

cp.asarray(cp.iinfo(cp.uint64).max, dtype=cp.uint64) == cp.asarray(1, dtype=cp.uint64)
# array(False)

(cp.iinfo(cp.uint64).max // 2) == cp.asarray(1, dtype=cp.uint64)
# array(False)
```

### Installation

### Environment

```
(scipy-dev) C:\Users\matth\Desktop\marray>python -c "import cupy; cupy.show_config(_full=True)"
meson-python: building scipy: meson compile
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\ninja.EXE
[2/356] Generating subprojects/highs/highs/HConfig.h with a custom command
OS                           : Windows-11-10.0.26200-SP0
Python Version               : 3.13.7
SciPy Version                : 1.18.0.dev0+git20260227.dc6795a
Cython Build Version         : 3.1.8
Cython Runtime Version       : 3.1.4
CUDA Root                    : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library
NVCC PATH                    : None
CUDA Driver Version          : 13000
CUDA Runtime Version         : 12090 (linked to CuPy) / 12090 (locally installed)
CUDA Runtime Path            : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\cudart64_12.dll (locally installed)
CUDA Extra Include Dirs      : ['C:\\Users\\matth\\.local\\share\\mamba\\envs\\scipy-dev\\Library\\include']
cuBLAS Version               : 120901
cuBLAS Path                  : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\cublas64_12.dll
cuFFT Version                : 11401
cuFFT Path                   : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\cufft64_11.dll
cuRAND Version               : 10310
cuRAND Path                  : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\curand64_10.dll
cuSOLVER Version             : (11, 7, 5)
cuSOLVER Path                : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\cusolver64_11.dll
cuSPARSE Version             : 12510
cuSPARSE Path                : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\cusparse64_12.dll
NVRTC Version                : (12, 9)
NVRTC Path                   : C:\Users\matth\.local\share\mamba\envs\scipy-dev\Library\bin\nvrtc64_120_0.dll
NCCL Path                    : None
Device 0 Name                : Quadro P400
Device 0 PCI Bus ID          : 0000:01:00.0```
```

### Additional Information

Discovered while testing compatibility between [MArray](https://github.com/mdhaber/marray) and CuPy.

seberg:
This does two things (both are needed to some degree):
* Introduce a `promote_types` function to customize the normal linear type lookup loop.  I could modify that (to do a double pass) but I prefer this (so long it is not considered public, in which case I would think more about the signature).
* Fix the actual mixed comparison, which unfortunately needs some branching here.  C promotes to unsigned which gives incorrect results.

(This will be incompatible with the caching changes in 9673, but that can be adjusted and is inivetable..)

Closes gh-9753

### DIFF
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -1573,7 +1573,7 @@ cdef class _Op:
 
 cdef class _Ops:
 
-    def __init__(self, tuple ops):
+    def __init__(self, tuple ops, promote_types=None):
         assert len(ops) > 0
         nin = ops[0].nin
         nout = ops[0].nout
@@ -1583,9 +1583,10 @@ cdef class _Ops:
         self.ops = ops
         self.nin = nin
         self.nout = nout
+        self.promote_types = promote_types
 
     @staticmethod
-    cdef _Ops from_tuples(object ops, routine):
+    cdef _Ops from_tuples(object ops, routine, object promote_types=None):
         ops_ = []
         for t in ops:
             if isinstance(t, tuple):
@@ -1607,7 +1608,7 @@ cdef class _Ops:
                         f"invalid op {t}, expected string or list")
                 typ, rt = t, routine
             ops_.append(_Op.from_type_and_routine(typ, rt))
-        return _Ops(tuple(ops_))
+        return _Ops(tuple(ops_), promote_types=promote_types)
 
     cpdef _Op guess_routine(
             self, str name, dict cache, list in_args, dtype, _Ops out_ops):
@@ -1671,15 +1672,17 @@ cdef class _Ops:
         cdef Py_ssize_t n = len(in_types)
         cdef Py_ssize_t i
 
+        if self.promote_types is not None:
+            # NOTE(seberg): This works fine, but should maybe be considered
+            # in flux. NumPy needs more (but mostly for loop registration).
+            in_types, weaks = self.promote_types(in_types, weaks)
+
         for op in self.ops:
             op_types = op.in_types
             for i in range(n):
                 it = in_types[i]
                 ot = op_types[i]
                 weak_t = weaks[i] if weaks is not None else False
-
-                # XXX: Remove assert (reachable only for pre NEP 50 logic)
-                assert(not isinstance(it, tuple))
 
                 if not can_cast(it, ot):
                     if not weak_t:
@@ -1711,9 +1714,11 @@ cdef class _Ops:
 
 cpdef create_ufunc(name, ops, routine=None, preamble='', doc='',
                    default_casting=None, loop_prep='', out_ops=None,
-                   cutensor_op=None, scatter_op=None):
-    ops_ = _Ops.from_tuples(ops, routine)
-    _out_ops = None if out_ops is None else _Ops.from_tuples(out_ops, routine)
+                   cutensor_op=None, scatter_op=None,
+                   promote_types=None):
+    ops_ = _Ops.from_tuples(ops, routine, promote_types)
+    _out_ops = None if out_ops is None else _Ops.from_tuples(
+        out_ops, routine, promote_types)
     return ufunc(
         name, ops_.nin, ops_.nout, ops_, preamble,
         loop_prep, doc, default_casting=default_casting, out_ops=_out_ops,
--- a/cupy/_core/_kernel.pyx
+++ b/cupy/_core/_kernel.pyx
@@ -1711,8 +1711,7 @@ cdef class _Ops:
     ):
         cdef _Op op
         cdef tuple op_types
-        cdef Py_ssize_t n = len(in_types)
-        cdef Py_ssize_t i
+        cdef int i
 
         if self.promote_types is not None:
             # NOTE(seberg): This works fine, but should maybe be considered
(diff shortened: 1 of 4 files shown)


leofang:
Q: Should we merge the structured dtype support first?

seberg:
Maybe slightly nicer, but I think if there are conflicts they should be small enough to be unproblematic no matter the order.

cupy/_core/_kernel.pyx
```diff
@@ -1671,16 +1672,18 @@ cdef class _Ops:
         cdef Py_ssize_t n = len(in_types)
```
eriknw: `n = ...` could totally be moved to after the "if" statement below. It's probably okay as is, but my ignorant, paranoid self would feel more comfortable initializing _after_ `in_types` gets updated in the if statement

eriknw (changes requested):
Stripping the weak `int` flag unconditionally fixes the integer dispatch but breaks NEP 50 parity for non-integer arrays:

```python
>>> cp.array([2048.0], dtype=cp.float16) == 2049
array([False])  # this PR

>>> np.array([2048.0], dtype=np.float16) == 2049
array([ True])  # NEP 50: 2049 rounds to 2048 in float16
```

A regression test for float dtypes vs Python ints would have caught this.

Also, this PR undersells what it does. More than the loud `OverFlowError`, this PR fixes silent wrong results! See:

```python
>>> cp.less(cp.array([0, 1, 2**63, 2**64-1], dtype='u8'), -1)
# main: array([ True,  True,  True, False])  # WRONG: C-promoted -1 to UINT64_MAX
# PR:   array([False, False, False, False])  # correct
```

Scary, huh? Glad it's fixed.

My Claude had some choice words about some of the comments, git commit messages, and lack of some tests. You can ask your AI about it if you like.

Other than the NEP 50 thing above, this PR looks correct to me.

seberg:
> Stripping the weak int flag unconditionally fixes the integer dispatch but breaks NEP 50 parity for non-integer arrays:

Hmmmpf, yeah... This is also somewhat unfortunate in NumPy (for comparisons specifically, not generally).  But since NumPy/CuPy promotes float64 and int64 to float64 _this_ specific approach only does something for float32 and float16 anyway, which is also silly.

So limited the special handling to `.kind in "ui"`.  The bool behavior diverges from NumPy very mildly, because we and up being OK with using `uint64` for the comparison as well, and NumPy (currently) rejects that.  I think that is fine, just can't fold it into a numpy-cupy comparison test.

cupy/_core/_kernel.pyx
seberg: Yeah, all the same, but we actually have `self.nin` defined and the class (as an `int`)... So might as well just use that, I think.

eriknw (commented):
I think this is shippable, but I do have one suggestion: please address the 282 warnings that this PR introduces. These clutter CI logs.

My Claude has some cosmetic recommendations. Nothing I think is serious.

seberg:
Ah, overflow warnings in the tests.  Added `@numpy.errstate(over='ignore')` (also to two old tests in the file spewing out warnings).

/test mini

eriknw (approved):

seberg:
Failure is:
```
FAILED cupy_tests\core_tests\test_raw.py::TestRawPicklable_param_3_{compile=True, raw='ker'}::test_raw_picklable - subprocess.CalledProcessError: Command '['C:\\Development\\Python\\Python310\\python.exe', 'TestRawPicklable.py']' returned non-zero exit status 3221226505.
```
again :/.

/test force-skip

leofang (approved):

PR #9764, opened 2026-03-03
docs: remove global footnote state in comparison generator

ChAbdulWahhab:
## Summary
This PR removes global mutable footnote state from `docs/source/_comparison_generator.py` and makes footnote handling local to each generation run.

## Changes
- Removed module-level `_footnotes` dictionary.
- Added `footnote_registry` parameter to:
  - `_generate_comparison_rst(...)`
  - `_section(...)`
- Initialized `footnote_registry` inside `generate()` and passed it through section generation.
- Updated footnote ID creation to use the local registry.
- Updated final footnote rendering to read from the local registry.

## Why
The previous global registry could leak state across repeated invocations in the same Python process, causing non-deterministic footnote numbering/content.

## Validation
- Ran:
  - `python -m py_compile docs/source/_comparison_generator.py`
- Result: passed.

### DIFF
--- a/docs/source/_comparison_generator.py
+++ b/docs/source/_comparison_generator.py
@@ -4,9 +4,6 @@ import inspect
 import numpy
 
 
-_footnotes = {}
-
-
 def _get_functions(obj, exclude=None):
     return set([
         n for n, target in [(n, getattr(obj, n)) for n in dir(obj)]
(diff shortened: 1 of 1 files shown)


seberg:
Thanks, and maybe others will disagree, but this one I'll just close. I see no real motivation here that I can see beyond _possible_ slight code smell improvements.
That's fine, but I am not sure I want too many of such PRs unless it is systematic and agreed upon.

ChAbdulWahhab:
Thanks for the review!

The motivation wasn't just 'style' - using a global dictionary causes state leakage across repeated calls in the same process (e.g., during tests or parallel builds), leading to duplicate/incorrect footnotes.

Totally respect your decision if you'd prefer to keep it as-is for now, though. Appreciate the feedback!

PR #9767, opened 2026-03-04
Fix NumPy signature mismatch for cupy.argmax

Punit200320:
This PR fixes the function signature mismatch between NumPy and CuPy for `cupy.argmax`.

The `dtype` argument has been removed from the CuPy API to align with NumPy's signature:

numpy.argmax(a, axis=None, out=None, keepdims=False)

This improves API compatibility between NumPy and CuPy.

### DIFF
--- a/cupy/_sorting/search.py
+++ b/cupy/_sorting/search.py
@@ -10,7 +10,6 @@ from cupy._core import _routines_indexing as _indexing
 from cupy._core import _routines_statistics as _statistics
 
 
-
 def argmax(a, axis=None, out=None, keepdims=False):
     """Returns the indices of the maximum along an axis.
 


seberg:
Thanks for looking into this.  I am confirming with others, but to me `dtype=` seems working and OK to keep around.
I would say we should change the scope slightly (sorry). That is modify the signature to something like: `argmax(a, axis=None, _dtype=None, out=None, keepdims=False, dtype=None)`
If `_dtype is not None` we then give a `DeprecationWarning` that in the Future out, keepdims, and dtype must be passed as keyword arguments.
(I.e. the future will be `argmax(a, axis=None, *, out=None, keepdims=False, dtype=None)` which differs from NumPy in the extra `dtype` and `out` being kwarg only, but I think both are fine.)

We also need to make sure to do two other things:
1. If we modify argmax, we should make sure to modify argmin and the methods in `core.pyx` also aligned with identical changes (they may just be able to forward to a single warning).
2. Any change should get some new tests to confirm it.

PR #9771, opened 2026-03-05
CI: Fix cuda120 CI failures due to `FutureWarning`

leofang:
After #9739 I noticed the cuda120 CIs were broken, because they use Python 3.10 which trigger a `FutureWarning`.

### DIFF
--- a/.pfnci/generate.py
+++ b/.pfnci/generate.py
@@ -205,7 +205,6 @@ class LinuxGenerator:
             raise ValueError('Python cannot be null')
 
         py_spec = self.schema['python'][matrix.python]['spec']
-        # For GCP kernel cache backend, install google-cloud-storage.
         lines += [
             'RUN git clone https://github.com/pyenv/pyenv.git /opt/pyenv',
             'ENV PYENV_ROOT "/opt/pyenv"',
@@ -213,6 +212,7 @@ class LinuxGenerator:
             f'RUN pyenv install {py_spec} && \\',
             f'    pyenv global {py_spec} && \\',
             '    pip install -U setuptools pip wheel && \\',
+            # For GCP kernel cache backend
             '    pip install -U google-cloud-storage',
             '',
         ]


seberg:
Thanks, I think I would slightly prefer adding an ignore to:
https://github.com/cupy/cupy/blob/main/pyproject.toml#L79
(I am assuming that is sufficient here.)

.pfnci/generate.py
```diff
@@ -205,14 +205,14 @@ def generate_dockerfile(self) -> str:
...
+        # For GCP kernel cache backend, install google-cloud-storage.
         lines += [
             'RUN git clone https://github.com/pyenv/pyenv.git /opt/pyenv',
             'ENV PYENV_ROOT "/opt/pyenv"',
             'ENV PATH "${PYENV_ROOT}/shims:${PYENV_ROOT}/bin:${PATH}"',
             f'RUN pyenv install {py_spec} && \\',
             f'    pyenv global {py_spec} && \\',
             '    pip install -U setuptools pip wheel && \\',
-            # For GCP kernel cache backend
             '    pip install -U google-cloud-storage',
```
leofang, line 216:
```suggestion
            # For GCP kernel cache backend
            '    pip install -U google-cloud-storage',
```

.pfnci/generate.py
```diff
@@ -205,14 +205,14 @@ def generate_dockerfile(self) -> str:
             raise ValueError('Python cannot be null')
 
         py_spec = self.schema['python'][matrix.python]['spec']
+        # For GCP kernel cache backend, install google-cloud-storage.
```
leofang:
```suggestion
```

leofang:
Let's give it a shot!

/test cuda120

leofang:
I wish we could have the option to do squash merge, I'd like to increase squash on my own PRs 😛 @asi1024 do you think we can turn it on but make it optional (default is still to generate a merge commit)?

PR #9729, opened 2026-02-19
Cython 'IF' statement deprecation

ManuCorrea:
### Summary
This PR addresses part of the Cython deprecation warnings reported in https://github.com/cupy/cupy/issues/9559, specifically those related to `IF` deprecation warnings.

The work done so far focuses solely on resolving deprecation warnings. Functional or behavioral changes may need to be included, opening this PR as draft so we can discuss the changes.

The remaining #IFs can't be changed into Python's ifs since they include type definitions, imports, etc that need to be resolved at compile time.

### What was changed
- Fixed warnings of the form:
  - `"The 'IF' statement is deprecated and will be removed in a future Cython version."`

### Options
The options I have considered so far are:
- Separate extension modules per variant.
- Build-time code generation. A script that reads the config and generates the correct .pyx, .pxd files.

Do we have any preference? Maybe a mix of both strategies work best depending on the file.

### Remaining warnings
```
  warning: cupy/cuda/cufft.pxd:13:4: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
  warning: cupy/cuda/cufft.pyx:1268:4: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
  warning: cupy/cuda/cufft.pyx:191:0: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
  warning: cupy_backends/cuda/api/_runtime_enum.pxd:122:0: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
  warning: cupy_backends/cuda/api/_runtime_enum.pxd:239:4: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
  warning: cupy_backends/cuda/api/_runtime_extern.pxi:163:4: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310
[...]

PR #9774, opened 2026-03-05
Ingnore Cython IF warnings and avoid DEF uses

ManuCorrea:
This PR aims to address the concerns expressed in https://github.com/cupy/cupy/issues/9559 regarding making Cython compilation warning free.

Discarded previous draft PR https://github.com/cupy/cupy/pull/9729 addressing IF since solution for warning deprecation since it needs further team discussion and strategy.

Included short term fix as suggested by @leofang.

### DIFF
--- a/.pfnci/linux/tests/actions/_environment.sh
+++ b/.pfnci/linux/tests/actions/_environment.sh
@@ -15,6 +15,7 @@ export CUPY_NUM_BUILD_JOBS=8
 # Setup CuPy env vars (for runtime)
 export CUPY_CACHE_DIR="${_cache_dir}/.cupy/kernel_cache"
 export CUPY_DUMP_CUDA_SOURCE_ON_ERROR="1"
+export CUPY_NVRTC_USE_PCH="1"
 
 # Setup CuPy env vars (for unit tests)
 export CUPY_TEST_GPU_LIMIT="${GPU:--1}"
--- a/.pfnci/windows/test.ps1
+++ b/.pfnci/windows/test.ps1
@@ -101,6 +101,7 @@ function Main {
 
     $Env:CUPY_TEST_GPU_LIMIT = $Env:GPU
     $Env:CUPY_DUMP_CUDA_SOURCE_ON_ERROR = "1"
+    $Env:CUPY_NVRTC_USE_PCH = "1"
 
     # Unit test
     if ($test -eq "build") {
--- a/cupy/_environment.py
+++ b/cupy/_environment.py
@@ -119,6 +119,8 @@ _PLATFORM_WIN32 = sys.platform.startswith('win32')
 def _get_conda_cuda_path():
     """This works since CUDA 12.0+."""
     conda_prefix = os.environ.get('CONDA_PREFIX')
+    if conda_prefix is None:
+        return None
     if _PLATFORM_LINUX:
         plat = platform.machine()
         if plat == 'aarch64':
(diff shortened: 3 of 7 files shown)


install/cupy_builder/_command.py
```diff
@@ -114,6 +114,8 @@ def _cythonize(self, nthreads: int) -> None:
             # Keep the behavior same as Cython 0.29.x.
             # https://github.com/cupy/cupy/pull/8457#issuecomment-2656568499
             'binding': False,
+            'warn.deprecated.IF': False,   # suppress IF deprecation warning
+            'warn.deprecated.DEF': False,  # suppress DEF deprecation warning
```
seberg: Thanks for looking at other filters!  From a quick glance, it seems like there are only very few `DEF` in `.pyx` files?  So I think they can be safely replaced with `cdef const int`.
(which gets translated to `static const int`).

ManuCorrea: You are welcome! I can take a look and fix them, should I use this PR for fixing those so we exclude `'warn.deprecated.DEF': False, ` or rather a different one to not mix up?

seberg: I don't care in this case, they should be all very small fixes (if it isn't many).
Some will create merge conflicts with other things I have in-flight, but let's not worry.

ManuCorrea: I did not find the DEF deprecation warning in my logs (without the Cython ignore warnings), unsure why but processing to fix them.

I am going to apply the changes for this PR.

ManuCorrea:
@seberg I applied the suggested changes, I built it successfully locally and ran some tests but didn't do a deep check since I need to leave now, I will check the following days better (in case there is something wrong). Thanks for the feedback!

cupy/_util.pyx
```diff
@@ -12,7 +12,7 @@ import cupy
 from cupy_backends.cuda.api cimport runtime
 
 
-DEF CYTHON_BUILD_VER = CUPY_CYTHON_VERSION
+cdef const char* CYTHON_BUILD_VER = CUPY_CYTHON_VERSION
 cython_build_ver = CYTHON_BUILD_VER
```
seberg:
```suggestion
cython_build_ver = CUPY_CYTHON_VERSION
```
I am almost surprised it works fine, but this is the no-op change, since we don't need `CYTHON_BUILD_VER`.
(The other change made it a `char*` and converting that to Python ends up with bytes which prints slightly less nice)

seberg:
Don't think this needs a full test, so let me pick one at random:

/test cuda128

seberg (approved):
Thanks, if things work this will be really nice.  The `noexcept` fix already reduced the amount of scrolling on cython failures a lot.

leofang:
The pre-test failed, not sure why... 🤔

seberg:
Heh, it's the new, weird, "noexcept" check.  The problem is that cython doesn't realize these values are not `0` so it inserts a division by zero check (which can raise).
Adding @cython.cdivision should be OK, I think.

seberg:
(Poking again, was hanging for 8+ hours, maybe it failed to trigger?)

/test windows,cuda128

leofang:
I think the last run had succeeded? https://ci.preferred.jp/cupy.linux.cuda128/211640/

seberg:
> I think the last run had succeeded? 

Ohhh, maybe it just failed to report!

leofang:
Let me just merge it and get it across the finish line.

Issue #9723, opened 2026-02-18
`show_config` is too verbose when libraries do not exist


leofang:
Example with `apidsai/devcontainers:latest-cpp-gcc13-cuda12.0`:
```
OS                           : Linux-6.14.0-37-generic-x86_64-with-glibc2.35
Python Version               : 3.10.12
CuPy Version                 : 14.0.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 2.2.6
SciPy Version                : None
Cython Build Version         : 3.1.4
Cython Runtime Version       : None
CUDA Root                    : /usr/local/cuda
NVCC PATH                    : /usr/local/cuda/bin/nvcc
CUDA Build Version           : 12090
CUDA Driver Version          : 13000
CUDA Runtime Version         : 12090 (linked to CuPy) / 12000 (locally installed)
CUDA Runtime Path            : /usr/local/cuda/targets/x86_64-linux/lib/libcudart.so.12 (locally installed)
CUDA Extra Include Dirs      : ['/usr/local/cuda/include']
[...]

PR #9775, opened 2026-03-05
Making show_config less verbose when libraries don't exist

ManuCorrea:
Addressing issue: [show_config is too verbose when libraries do not exist](https://github.com/cupy/cupy/issues/9723)

If the value of a path, version, etc is None it won't be printed.

Also as @leofang mentioned in the comment the `ImportError` should be `None`, hence with the current implementation excluded from the information. Let me know if this is the desired behavior.

### DIFF
--- a/cupyx/_runtime.py
+++ b/cupyx/_runtime.py
@@ -23,6 +23,8 @@ def _eval_or_error(func, errors):
     # representing the error.
     try:
         return func()
+    except ImportError:
+        return None
     except errors as e:
         return repr(e)
 
@@ -405,6 +407,8 @@ class _RuntimeInfo:
                     arch = ('Device {} Compute Capability'.format(device_id),
                             device.compute_capability)
                 records += [name, arch, pci_bus]
+
+        records = [(k, v) for (k, v) in records if v is not None]
 
         width = max([len(r[0]) for r in records]) + 2
         fmt = '{:' + str(width) + '}: {}\n'


leofang (changes requested):
Thanks for working on this @ManuCorrea!

The `except ImportError: return None` change in `_eval_or_error` is exactly right — it swallows the verbose cuda-pathfinder `ImportError` messages and turns them into `None`, which is what we want.

However, the `None`-filtering in `__str__` is too aggressive:

```python
records = [(k, v) for (k, v) in records if v is not None]
```

This hides **every** row whose value is `None`, not just the ones that came from `ImportError`. Rows like `SciPy Version: None`, `NCCL Build Version: None`, and `cuTENSOR Version: None` would all disappear from the output. That information is useful — a user running `show_config` should be able to see which libraries are not installed. If the row is simply absent, it's ambiguous whether the library isn't installed or whether `show_config` doesn't check for it.

The intent from #9723 was to replace the verbose `ImportError(...)` dump with a short `None` — not to hide all `None` entries. With just the `_eval_or_error` change (and without the filter), the output would show e.g. `cuBLAS Version: None` instead of the multi-line directory listing, which is the desired behavior.

**Requested changes:**

1. **Remove the `None`-filter line** from `__str__`. The `_eval_or_error` change alone is sufficient to fix the issue.

2. **Add a test** in `tests/cupyx_tests/test_runtime.py`. There's already a `test_error` case that mocks functions to raise `CUDARuntimeError`. Please add a similar test that mocks a library function (e.g. `cublas.getVersion`) to raise `ImportError` and verifies:
   - The verbose `ImportError` text does **not** appear in `str(runtime)`
   - The corresponding row still shows `None` in the output (i.e. it is not hidden)

Issue #9777, opened 2026-03-05
CuPy v14.0.0+ is buggy in a conda env


leofang:
### Description

This line:
https://github.com/cupy/cupy/blob/bdacbd07c17d7babd5d5c2c4a3fcdc1db20395e2/cupy/_environment.py#L121
could be problematic because there is no guarantee that `CONDA_PREFIX` would exist. Here is one example traceback:
```python
File /opt/conda/lib/python3.12/site-packages/cupy/__init__.py:18
     14 _environment._preload_library('cutensor')  # NOQA
     17 try:
---> 18     from cupy import _core  # NOQA
     19 except ImportError as exc:
     20     raise ImportError(f'''
     21 ================================================================
     22 {_environment._diagnose_import_error()}
   (...)     26 ================================================================
     27 ''') from exc

File /opt/conda/lib/python3.12/site-packages/cupy/_core/__init__.py:5
      1 # mypy: ignore-errors
      2 from __future__ import annotations
----> 5 from cupy._core import core  # NOQA
      6 from cupy._core import fusion  # NOQA
      7 from cupy._core import internal  # NOQA

File cupy/_core/core.pyx:1, in init cupy._core.core()

File cupy/_core/_routines_math.pyx:7, in init cupy._core._routines_math()

File cupy/_core/_reduction.pyx:1, in init cupy._core._reduction()

File cupy/_core/_cub_reduction.pyx:260, in init cupy._core._cub_reduction()

File /opt/conda/lib/python3.12/site-packages/cupy/_environment.py:86, in get_nvcc_path()
     84 global _nvcc_path
     85 if _nvcc_path == '':
---> 86     _nvcc_path = _get_nvcc_path()
     87 return _nvcc_path

File /opt/conda/lib/python3.12/site-packages/cupy/_environment.py:198, in _get_nvcc_path()
    195     return nvcc_path
    197 # Lookup <CUDA>/bin
--> 198 cuda_path = get_cuda_path()
    199 if cuda_path is None:
    200     return None

File /opt/conda/lib/python3.12/site-packages/cupy/_environment.py:78, in get_cuda_path()
     76 global _cuda_path
     77 if _cuda_path == '':
---> 78     _cuda_path = _get_cuda_path()
     79 return _cuda_path

File /opt/conda/lib/python3.12/site-packages/cupy/_environment.py:151, in _get_cuda_path()
    149 config = get_preload_config()
    150 if config is not None and config['packaging'] == 'conda':
[...]

PR #9779, opened 2026-03-05
Guard against None conda prefix in _get_conda_cuda_path()

leofang:
Fix https://github.com/cupy/cupy/issues/9777


seberg (approved):
Probably the best thing, yeah. Even if it only happens due to half activatedconda envs.
Thanks!

PR #9781, opened 2026-03-05
If `CONDA_PREFIX` is undefined, use `sys.prefix`

### LINKED ISSUE #9777: CuPy v14.0.0+ is buggy in a conda env

jakirkham:
Fixes https://github.com/cupy/cupy/issues/9777

The `sys.prefix` points to the top of the environment that Python lives in. So in most cases this is what we want. If the environment is not active and so `CONDA_PREFIX` is not defined, fallback to `sys.prefix` as that is a good choice.

Longer term we might consider standardizing and streamlining the codebase to make use of `sys.prefix` throughout and dropping specialized Conda logic. This will ensure one code path works well in a variety of common environments without as much specialization. It will also help be more robust and easier to test.


seberg:
Thanks John, I went ahead with this: https://github.com/cupy/cupy/pull/9779 before noticing your PR.

In the CuPy case, things are even weirder, because a later `pip install` claims to be conda even though it isn't (at least it seems to me that way).

I think that fix should be fine, because we probably still find the right `nvrtc` and the right path, although the search priority probably differs in a badly activated conda env.

jakirkham:
Thanks Sebastian! 🙏


Probably the long term goal would be to better integrate these different code paths in a more standardized way. Admittedly easy to say, but requires some finesse to do

Could you please share more about the use cases that you are encountering? That will help inform the next steps

seberg:
We have had it happen in CI based on the rapids containers (I am aware of a few places where we ran into it, but I think they were all based on rapids containers).

In the CuPy case, we `pip install .` from source on top of a previous `conda-forge` install.  Which is a bit curious, because it seems that removing the original `conda-forge` install meant that the package was detected as `pip` installed.

But, I don't think all of this matters too much, in practice all solutions seem good enough.
It isn't clear to me if the best path (also in `cuda-pathfinder`) is to just ignore conda if `CONDA_PREFIX` isn't set (now merged PR), do this approach, or even warn that the conda environment isn't properly activated.

FWIW, I am happy to go with any (even slight) preference you have here.

jakirkham:
Thanks for the additional context Sebastian! 🙏

Certainly there can be subtleties when mixing Conda & Pip. Though agree the approach you are using here is reasonable. If one wants to be a bit more cautious one, they can remove the existing Conda package first (though leave the dependencies) with `conda remove --force-remove` first. If one wants to be extra careful, they can run `pip install --no-deps --ignore-installed --no-index` to bypass installing anything else other than the specified packages

The `sys.prefix` just comes from the running Python. In general this refers to the top-level directory where the underlying directories for binaries, libraries, etc. live. For Conda specifically, this is the same as the environment's prefix. So this is a safe choice to use

In fact [Numba makes use of `sys.prefix` in its Conda path discovery logic]( https://github.com/numba/numba/blob/be990a818c377408a1f41c37c7720bc2ba0fb89a/numba/cuda/cuda_paths.py#L126 ) as [does Numba-CUDA]( https://github.com/NVIDIA/numba-cuda/blob/893dde7ed75ade4ef44b42ed0163e53982fbcf49/numba_cuda/numba/cuda/cuda_paths.py#L244 )

Of course it is better for users to activate Conda environments. Sometimes there are situations where that isn't possible. So where we can trying to improve the unactivated experience where we can is a good thing

That all being said, am more interested in how we can streamline the logic here so we can rely less on Conda specifics where possible and more general path detection logic from Python's configuration. With the goal of a better experience for all parties involved 🙂

seberg:
Coming back after a while, I am very slightly leaning towards just merging if numnba uses it, since I suspect if we take this path (recognize things as conda installed) it is reasonable to try `sys.prefix`.

I want to give Leo a chance to comment, but if not will put it in eventually (and mark it as blocker for now).

leofang:
I’d prefer to not use sys.prefix if possible (that is, closing, instead of merging, this PR). It is not tested with pathfinder which we now heavily rely on, and pathfinder does not honor sys.prefix (for good reasons subject to debate). I’d prefer that we move the sys.prefix discussion to NVIDIA/cuda-python and resolve there.

Issue #8590, opened 2024-09-09
cupyx.scipy.sparse.linalg.lsmr's argument is different from scipy's counterpart


so298:
### Description

There is some differences in cupyx.scipy's `lsmr()` and scipy's original `lsmr()` with their arguments.

CuPy
https://docs.cupy.dev/en/latest/reference/generated/cupyx.scipy.sparse.linalg.lsmr.html
```py
lsmr(A, b, x0=None, damp=0.0, atol=1e-06, btol=1e-06, conlim=100000000.0, maxiter=None)
```

SciPy
https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.lsmr.html
```py
lsmr(A, b, damp=0.0, atol=1e-06, btol=1e-06, conlim=100000000.0, maxiter=None, show=False, x0=None)
```

I found the following differences:

- Position of `x0`.
- CuPy's lsmr does not have `show` argument.
- When `maxiter=None` (default behavior), SciPy lsmr sets `maxiter=min(m, n)` (where (m, n) is the shape of A) automatically while CuPy's does not do so.

### To Reproduce

```py
# Write the code here
```

### Installation

None

### Environment

```
# Paste the output here
```

PR #9792, opened 2026-03-10
Align lsmr signature and show option with SciPy

sk-226:
## Summary
- Updated `cupyx.scipy.sparse.linalg.lsmr` to follow SciPy's argument order: `damp, atol, btol, conlim, maxiter, show, x0`
- Added support for the `show` argument so that `lsmr(..., show=True)` prints iteration logs in a SciPy-like format
- Aligned the default `maxiter` behavior with SciPy by changing it from `min(m, n) * 5` to `min(m, n)`
- Updated the docstring accordingly
- Add tests for `cupyx.scipy.sparse.linalg.lsmr(show=True)`

## Related issue
- Related to [Issue 8590](https://github.com/cupy/cupy/issues/8590)

## Current status
- Updated `cupyx/scipy/sparse/linalg/_solve.py`
- Confirmed the implementation direction locally
- Tests are not added yet

## Why this is a WIP PR
I am opening this PR early to confirm that the implementation direction is acceptable before adding tests and polishing the rest.

## Feedback requested
- Whether changing the argument order to match SciPy is acceptable here
- Whether the current `show` behavior is sufficient
- Whether there are any additional compatibility concerns I should address before adding tests

## Additional points to confirm
- This change breaks positional compatibility for `lsmr(A, b, x0)`, since `x0` is now moved to the SciPy-compatible position after `show`
- In the initial `show=True` logging path, `test1` is currently printed as `1`; I would like to confirm whether it should instead be computed as `normr / normb` when `x0` is not zero
- SciPy also has an early return for `if normb == 0:`; I have not added it in this PR, so I would like to confirm whether CuPy should follow that behavior here

## Remaining work
- [x] Add tests for `lsmr`

### DIFF
--- a/cupyx/scipy/sparse/linalg/_solve.py
+++ b/cupyx/scipy/sparse/linalg/_solve.py
@@ -84,8 +84,8 @@ def lsqr(A, b):
     return ret
 
 
-def lsmr(A, b, x0=None, damp=0.0, atol=1e-6, btol=1e-6, conlim=1e8,
-         maxiter=None):
+def lsmr(A, b, damp=0.0, atol=1e-6, btol=1e-6, conlim=1e8, maxiter=None,
+         show=False, x0=None):
     """Iterative solver for least-squares problems.
 
     lsmr solves the system of linear equations ``Ax = b``. If the system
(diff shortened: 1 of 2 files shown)


cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -84,8 +84,8 @@ def lsqr(A, b):
     return ret
 
 
-def lsmr(A, b, x0=None, damp=0.0, atol=1e-6, btol=1e-6, conlim=1e8,
-         maxiter=None):
+def lsmr(A, b, damp=0.0, atol=1e-6, btol=1e-6, conlim=1e8, maxiter=None,
+         show=False, x0=None):
```
leofang, line 88: 👍  this is good catch, apparently the signature has been like this for many SciPy releases, looks like we never caught it because we always pass kw-args correctly (instead of treating them as positional args).

This reordering is technically a breaking change (due to potential misuse of kw-args), but I would like to think it's a bug fix instead.

leofang (approved):

PR #9793, opened 2026-03-10
CI: Make sure local cache is warmed up at job start time

### LINKED ISSUE #9791: CI: Ensure runner local cache is initialized?
Follow-up of #9665.

With the new GCP cache backend I've noticed that the e2e job run time on the main/v14 branches has become longer than before (when the old file-based backend was in use), see the screenshot below. It is now longer 40-60 mins longer (2h->3h). I suspect this has to do with the new backend triggering numerous network access in every fresh CI run that starts with an empty local cache. My hunch is we need to figure out a way to populate the local cache using the GCP cache when the job starts.

The red line in the screenshot separates (roughly) CI runs before/after switching to the GCP backend.
Image

leofang:
Close #9791.

### DIFF
--- a/tests/cupy_tests/_gcp_kernel_cache_backend.py
+++ b/tests/cupy_tests/_gcp_kernel_cache_backend.py
@@ -15,8 +15,6 @@ Requirements:
 
 from __future__ import annotations
 
-import os
-import tempfile
 import time
 import warnings
 
@@ -31,9 +29,7 @@ except ImportError:
     transfer_manager = None  # type: ignore
     gcp_exceptions = None  # type: ignore
 
-from cupy.cuda._compiler_cache import (  # noqa
-    DiskKernelCacheBackend, _hash_hexdigest, _hash_length,
-)
+from cupy.cuda._compiler_cache import DiskKernelCacheBackend  # noqa
 from cupy.cuda.compiler import _get_cupy_cache_key
 from cupy_backends.cuda.libs import nvrtc as _nvrtc
 
@@ -126,18 +122,13 @@ class GCPStorageCacheBackend(DiskKernelCacheBackend):
         prefix_len = len(self._nvrtc_prefix)
 
         try:
-            # list_blobs returns a lazy HTTPIterator; we materialise it here
-            # to separate listing errors from download errors and to build the
-            # filtered list that is passed to download_many_to_path.
+            # list_blobs returns a lazy HTTPIterator; materialise it to
+            # separate listing errors from download errors and to build the
+            # list passed to download_many_to_path.
             blob_iter = self._bucket.list_blobs(prefix=self._nvrtc_prefix)
-            # blob_names are relative to self._nvrtc_prefix (the prefix is
-            # passed separately to download_many_to_path as blob_name_prefix).
-            missing_blob_names = [
-                blob.name[prefix_len:] for blob in blob_iter
-                if blob.name[prefix_len:]  # skip the prefix "directory" entry
-                and not os.path.exists(
-                    os.path.join(self._cache_dir, blob.name[prefix_len:]))
-            ]
+            # blob_names are relative to self._nvrtc_prefix (passed separately
+            # as blob_name_prefix so it is stripped from the destination path).
+            blob_names = [blob.name[prefix_len:] for blob in blob_iter]
         except Exception as e:
             warnings.warn(
                 f"Failed to list GCS objects for cache initialization: "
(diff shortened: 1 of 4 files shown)


tests/cupy_tests/_gcp_kernel_cache_backend.py
```diff
@@ -64,6 +68,10 @@ def __init__(
 
         self._gcp_enabled = False
         self._prefix = prefix
+        # Narrow the GCS namespace to this pipeline's NVRTC (CUDA) version so
+        # that initialize_local_cache only downloads what this pipeline needs.
+        major, minor = _nvrtc.getVersion()
+        self._nvrtc_prefix = f'{self._prefix}CUDA_{major}_{minor}/'
```
seberg: As discussed offline, don't we need the cuda cache key here?
(Not even necessarily to reduce the download, but rather to naturally stop touching outdated versions – I am assuming that cleanup is based on the last used time.)

tests/cupy_tests/_gcp_kernel_cache_backend.py
```diff
@@ -86,6 +94,82 @@ def __init__(
...
+            # executor.map lets GCS page fetching and blob downloads interleave:
+            # futures for already-listed blobs are submitted while subsequent
+            # pages are still being fetched from GCS.
+            blob_iter = self._bucket.list_blobs(prefix=self._nvrtc_prefix)
+            with concurrent.futures.ThreadPoolExecutor(
+                    max_workers=max_workers) as executor:
+                # list() must be called inside the with-block so that all
+                # futures are submitted and their results collected before the
+                # executor shuts down.
+                results = list(executor.map(_download_if_missing, blob_iter))
```
seberg: This looks good. But I just noticed that this already subclasses `DiskKernelCacheBackend`.
Maybe we can just use [`transfer_manager.download_many_to_path()`](https://docs.cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.transfer_manager.html#google_cloud_storage_transfer_manager_download_many_to_path) even if that may lose a bit of reporting details?

seberg (commented):
Just some comments, let's see how this trial goes after the cache is actually filled in the next iteration...

seberg:
/test linux,cuda129

I'll prod it to see if cache is used now (it came to 69%).  Although, I am confused why I don't see any printout about cache use (when no other stdout output looks missing).

leofang:
> Although, I am confused why I don't see any printout about cache use (when no other stdout output looks missing).

It's likely because I undid the change that passed `-s` to pytest.

tests/cupy_tests/_gcp_kernel_cache_backend.py
leofang: This is a great idea, now implemented. I did review the APIs of `Blob` and `Bucket` and look for a one-stop shop, but I found nothing and did not realize there is a standalone function for this.

tests/cupy_tests/_gcp_kernel_cache_backend.py
leofang: both `CUPY_CACHE_KEY` and MAJOR_MINOR are used now

leofang:
> I'll prod it to see if cache is used now (it came to 69%).

Seems to be the case (5h26m -> 2h19m):
https://ci.preferred.jp/r/job/211986

seberg (dismissed):
Approving just in case it is settled fully. Thanks for the decode/encode refactor!

Since the earlier try used ~2h, it seems like this should do the trick...

leofang:
> Since the earlier try used ~2h, it seems like this should do the trick...

Per-PR CI runs faster because the test matrix is smaller (than that of the main/v14 branches). With the old cache, per-PR runs only took about 1.5 hr. The new cache (before this PR) took ~2 hr, so this is still within the margin. I'll retry a few times to make sure it helps.

seberg:
3h21m this time, slow, but expected. Triggering again

/test cuda129

EDIT: ooops, sorry forgot the `linux`...

seberg:
OK, it should have been hot in the last run and the result is: **test succeeded in 1h41m32.154482406s**  so it might still be slightly slower than the first time around.
Let me do one more, for the statistics/curiosity:

/test linux,cuda129

**EDIT:** This run: test succeeded in 1h38m25.625151001s

leofang:
Yeah I think Linux is good, but let me check Windows one more time

/test windows,cuda129

tests/cupy_tests/_gcp_kernel_cache_backend.py
```diff
@@ -64,6 +69,14 @@ def __init__(
...
         self._gcp_enabled = False
         self._prefix = prefix
+        # Narrow the GCS namespace to this pipeline's cache key and NVRTC
+        # (CUDA) version so that initialize_local_cache only downloads what
+        # this pipeline needs, and so that kernels compiled with different
+        # CuPy builds do not collide or interfere with each other.
+        major, minor = _nvrtc.getVersion()
+        self._nvrtc_prefix = (
+            f'{self._prefix}{_get_cupy_cache_key()}/CUDA_{major}_{minor}/'
+        )
```
leofang, line 96: We discussed offline. Looks like the cubin hashes are very sensitive to OS/platform differences (through indirect means), so that Windows ends up not reusing the blobs generated from Linux. 

In theory, a CUBIN is OS/platform-agnostic. It runs everywhere as long as the driver and GPU arch are the same. However, if reuse does not happen, then the way we call `initialize_local_cache` would bring in more blobs than needed. This prefix is not granular enough and we should add the OS/plat identifier here
- `linux-64`
- `linux-aarch64`
- `win-64`

seberg (approved):

leofang:
Windows had warmed up and it only took 1.5 hr! https://ci.preferred.jp/r/job/212151

I triggered the CI again due to the prefix change. I expect to hit timeouts. I'll merge and let it retry in the main branch and the backport PR, to accelerate cache warm-up.

PR #9795, opened 2026-03-10
BUG: Fix bug with minimum_phase

larsoner:
Meant as a companion to https://github.com/scipy/scipy/pull/24749 . TL;DR is that there is a bug with how Nyquist is handled and this fixes it.

No idea if I've implemented the test correctly, and I expect it to fail until https://github.com/scipy/scipy/pull/24749 is merged and it's tested against that.

Maintainers feel free to push commits to fix this as needed (or close and open your own PR), figured getting the ball rolling was better than just opening an issue!

### DIFF
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
@@ -447,3 +447,15 @@ class TestMinimumPhase:
         if xp == cupy:
             h_linear = cupy.asarray(h_linear)
         return scp.signal.minimum_phase(h_linear, method="hilbert")
+
+    @testing.numpy_cupy_allclose(scipy_name="scp", atol=1e-6)
+    @pytest.mark.parametrize("N", (963, 964))
+    @pytest.mark.parametrize("dtype", ("float32", "float64"))
+    def test_nyquist(self, N, dtype, xp, scp):
+        fc = xp.asarray(10)
+        fs = 100
+        h = scp.signal.firwin(
+            N, fc, window="hann", pass_zero="lowpass", scale=False, fs=fs
+        )
+        h = h.astype(dtype)
+        return scp.signal.minimum_phase(h, method="homomorphic", n_fft=N, half=False)
--- a/tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
+++ b/tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
@@ -408,7 +408,7 @@ class TestMinimumPhase:
         assert_raises(ValueError, signal.minimum_phase,
                       cupy.ones(10), method='foo')
 
-    @testing.with_requires("scipy>=1.14")
+    @testing.with_requires("scipy>=1.18.0")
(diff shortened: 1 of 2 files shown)


tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
```diff
@@ -447,3 +447,18 @@ def test_hilbert(self, xp, scp):  # , n):
         if xp == cupy:
             h_linear = cupy.asarray(h_linear)
         return scp.signal.minimum_phase(h_linear, method="hilbert")
+
+    @testing.numpy_cupy_allclose(scipy_name="scp", atol=1e-6)
+    @pytest.mark.parametrize("N", (963, 964))
+    @pytest.mark.parametrize("dtype", ("float32", "float64"))
+    def test_nyquist(self, N, dtype, xp, scp):
```
leofang, line 455: I assume we want this?
```suggestion
    @testing.with_requires("scipy>=1.18.0")
    def test_nyquist(self, N, dtype, xp, scp):
```

larsoner: Yeah makes sense to me, pushed!

leofang (dismissed):

leofang:
@larsoner could you check the CI failures? Do we need to apply the test requirement to other tests?

larsoner:
Yeah I added/adjusted two more requires, feel free to tell it to run tests :crossed_fingers: 

I would test locally but I hit https://github.com/cupy/cupy/issues/8124#issuecomment-4032283452

leofang (approved):
Thanks, @larsoner! If this fix ends up not being part of 1.18, please do let us know!

tests/cupyx_tests/scipy_tests/signal_tests/test_fir_filter_design.py
```diff
@@ -447,3 +447,19 @@ def test_hilbert(self, xp, scp):  # , n):
         if xp == cupy:
             h_linear = cupy.asarray(h_linear)
         return scp.signal.minimum_phase(h_linear, method="hilbert")
+
+    @testing.with_requires("scipy>=1.18.0")
+    @testing.numpy_cupy_allclose(scipy_name="scp", atol=1e-6)
+    @pytest.mark.parametrize("N", (963, 964))
+    @pytest.mark.parametrize("dtype", ("float32", "float64"))
+    def test_nyquist(self, N, dtype, xp, scp):
+        fc = xp.asarray(10)
```
seberg, line 456: @larsoner if you have a moment to confirm on this fix:

It's OK if CuPy and SciPy diverge here (this test is failing). That is because the input ends up with a zero at the nyquist frequency (while CuPy ends up with 1e-15 or so) and the `h_temp[h_temp > 0].min()` blows things up, because it means the smallest value can differ a lot in magnitude.

(If I hear nothing, that is what I'll do, seems all right, but this added test wasn't super long ago, so... :))

larsoner: Yeah it seems fine to have the conditional be (not) "numerically zero-ish" instead of strictly (not) zero

PR #9801, opened 2026-03-13
Support caching CUBINs generated with `name_expressions`

### LINKED ISSUE #9667: Load kernels with mangled name from cache
### Description

Currently, when `name_expressions` is passed to the `RawModule` constructor, the kernel cache on the filesystem is never considered (see comment and implementation in [compiler.py](https://github.com/cupy/cupy/blob/d21cbea587afaec110695ab073b4e230e71eb1b9/cupy/cuda/compiler.py#L604). E.g. the following:

```python
program = """
template<typename T>
__global__ void kernel(T* inout)
{
    const int tid = blockDim.x * blockIdx.x + threadIdx.x;
    inout[tid] = static_cast<T>(tid);
}
"""

name = "kernel<float>"
module = cp.RawModule(code=program, name_expressions=[name], options=("-std=c++14",))
kernel = module.get_function(name)
print(kernel.name) # explained later
```

gets compiled every time the script runs.

Obviously, due to the use of the function template, `extern "C"` cannot be added. However, the resulting binary program still gets cached, and it is possible to retrieve the the instantiated symbol by its mangled name (printed in the above example). E.g. running the next example after the previous works, and is a cache hit.

```python
program = """
template<typename T>
__global__ void kernel(T* inout)
{
    const int tid = blockDim.x * blockIdx.x + threadIdx.x;
    inout[tid] = static_cast<T>(tid);
}
"""

module = cp.RawModule(code=program, options=("-std=c++14",))
kernel = module.get_function("_Z6kernelIfEvPT_") # mangled name from previous example
```

From a birds eye view, it should be possible either to generate the mangled name of the C++ function, henceforth retrieving it from the cache, or to save the C++ to mangled name mapping to the cache. Applications operating with kernel templates would benefit from reduced compilation times on repeated runs.

### Additional Information

_No response_

leofang:
Close #9667. Close #6159.

This helps reduce test time greatly. For example with `test_mdspan.py`:
|   | main | this PR
|---|---|---|
1st run (cold cache) | 66s| 66s | 
2nd run (warm cache) | 36s | **3s** |


leofang:
The win.cuda120 pipeline failure is irrelevant.

seberg:
I have to look closer at the code once, but overall looks fine and it is honestly cool to demangle the names.
But even after reading the issue discussion, I have to ask: If this is only about caching, is there much of a reason to not just store a `<hash>-name_expressions.json` that contains the mapping `name->mangled_name` from the original pass?  Try to read it and if all names are found, we are good?

leofang:
I prefer a solution that relies entirely on CUDA's offering so that we do not need any customization. We finally have it by CUDA 12. I am not 100% happy about having to do `normalize_name_expr` to strip out the argument list (+ handling Thrust internal namespace), I will talk to the compiler team about it. The executable version of `cu++filt` does offer a `-p` flag to only print function names so it should be totally doable. 

I don't want to add more files to disk because, in addition to increase the disk usage (and changing our caching scheme), it is a problem with concurrent access. On an HPC system we've seen accessing on-disk cache could add stress to the filesystem. (`cache_in_memory` was an extreme workaround provided to cope with file systems having a bad day, so that users can still run CuPy code, but on good days the disk cache should still be preferred, as long as we don't stress it out.)

That said, I recall now the reason we punted on #6159: We do not have a way to parse all user-included headers and we probably should not (it hurts perf). But if we don't, the CUBIN hash would not be fully accurate. This PR does not solve that case either. I don't have good suggestions other than documenting the limitation.

cupy/cuda/compiler.py
```diff
@@ -639,16 +669,29 @@ def _compile_with_cache_cuda(
...
             cubin = _kernel_cache_backend.load(name)
             if cubin is not None:
                 if to_ltoir:
                     return cubin
                 else:
                     mod.load(cubin)
-                    return mod
+                    if name_expressions:
+                        mod._enumerate_and_build_mapping(
+                            tuple(name_expressions))
```
seberg, line 686: It's unlikely or almost implausible to fail (except if unsupported, and we shouldn't call it then).
It also seems like reasonable public API, so how about folding it into `load` and `load_file`?  The problem I do see is that `load` seems like public Cython API, though.
So unless it is clear that CuPy doesn't consider that public (plus unlikely anyone uses it), please ignore...

That way, we can move the
```
try:
    mod.load()
except Exception:
    pass
```
avoiding the try/except there and the unload here.

Or, maybe nicer: `def __init__(*, binary=None, filename=None, name_expressions=None):` (of course the below one can't use it, or need to call `_set_mapping` afterwards rather than before).

cupy/cuda/function.pyx
```diff
@@ -18,6 +19,178 @@ from cupy.cuda.texture cimport TextureObject, SurfaceObject
...
+IF CUPY_CUDA_VERSION > 0:
+    cdef extern from "../../cupy_backends/cupy_backend.h" nogil:
+        pass
+
+    # Platform-abstracted C++ demangling via NVIDIA's libcufilt.
+    # Linux: __cu_demangle is linked directly from libcufilt.a.
+    # Windows: cupy_cufilt.dll (a /MT trampoline) is loaded at runtime
+    #          to avoid the /MT vs /MD CRT mismatch with cufilt.lib.
+    cdef extern from *:
+        """
```
seberg:
```suggestion
        r"""
```
Should work to reduce the backslash repetition, but but doesn't matter either way.

cupy/cuda/function.pyx
```diff
@@ -207,6 +380,56 @@ cdef class Module:
...
+
+            try:
+                num_functions = driver.moduleGetFunctionCount(self.ptr)
+                function_handles = driver.moduleEnumerateFunctions(
+                    self.ptr, num_functions)
+
+                for i in range(function_handles.size()):
+                    mangled_cstr = driver.funcGetName(
+                        <intptr_t>(function_handles[i]))
+                    mangled_str = mangled_cstr.decode('utf-8')
```
seberg, line 422: This would be better inside `demangle_cxx_name`, I think.  (i.e. it's passing the same thing twice)

I also don't understand why the function returns the original string on failure, why not return `None`?

cupy/cuda/function.pyx
```diff
@@ -18,6 +19,178 @@ from cupy.cuda.texture cimport TextureObject, SurfaceObject
...
+    to
+      ``"square<complex<double>>"``.
+
+    Steps:
+      1. Strip the ``void`` return-type prefix (CUDA kernels always return
+         void; non-template kernels are demangled without one).
+      2. Strip the parameter list ``(...)`` that follows the function name
+         (and any template arguments).
+      3. Replace ``thrust::*::complex`` with ``complex`` (CuPy's bundled
+         Thrust namespace).
```
seberg, line 163: I guess this works well enough. In principle, some other/user type aliases might not work?  (I see that float16 works fine, though)

cupy/cuda/function.pyx
```diff
@@ -207,6 +380,56 @@ cdef class Module:
...
+                    mangled_cstr = driver.funcGetName(
+                        <intptr_t>(function_handles[i]))
+                    mangled_str = mangled_cstr.decode('utf-8')
+                    demangled = demangle_cxx_name(mangled_cstr, mangled_str)
+                    normalized = normalize_name_expr(demangled)
+                    original = norm_to_user.get(normalized)
+                    if original is not None:
+                        mapping[original] = mangled_str
+            except Exception:
+                pass
```
seberg, line 429: Unclear why it added this defensive code.  The interesting part would be if we don't find an entry in `norm_to_user.get()` or maybe if demangling fails?
In that case it would seems sensible to raise an error, because we are missing expressions that may fail usage later.

cupy/cuda/function.pyx
```diff
@@ -207,6 +380,56 @@ cdef class Module:
...
+        .. note::
+            This function requires CUDA driver 12.4 or later.
+            On HIP/ROCm, this raises NotImplementedError.
+        """
+        IF CUPY_CUDA_VERSION > 0:
+            if self.mapping is not None:
+                return  # Already built
+
+            cdef vector.vector[driver.Function] function_handles
+            cdef unsigned int i, num_functions
```
seberg:
```suggestion
            cdef Py_ssize_t i, num_functions
```
sure, doesn't matter in practice, but let's use the native type.

cupy_backends/cuda/api/driver.pyx
```diff
@@ -257,6 +258,63 @@ cpdef intptr_t moduleGetGlobal(intptr_t module, str varname) except? 0:
...
+    initialize()
+    cdef unsigned int count = 0
+    with nogil:
+        status = cuModuleGetFunctionCount(&count, <Module>module)
+    check_status(status)
+    return count
+
+
+cdef vector.vector[Function] moduleEnumerateFunctions(
+        intptr_t module, unsigned int count) except +:
```
seberg: only thing to avoid the `except *` would be to pass in the vector (and add a `.clear()`, I guess).
Would honestly make sense to me.  The return value could then be an ignored `int` (or nothing, but IIRC you prefer the `int`).

cupy_backends/cuda/api/driver.pyx
```diff
@@ -257,6 +258,63 @@ cpdef intptr_t moduleGetGlobal(intptr_t module, str varname) except? 0:
...
+
+    Returns:
+        Vector of function handles.
+    """
+    initialize()
+    cdef vector.vector[Function] result
+    result.resize(count)
+
+    if count == 0:
+        return result
```
seberg: Not that it matters, but if you know: does `cuModuleEnumerateFunctions` really not like 0?

cupy_backends/cuda/api/driver.pxd
```diff
@@ -65,6 +66,10 @@ cpdef intptr_t moduleLoadData(bytes image) except? 0
 cpdef moduleUnload(intptr_t module)
 cpdef intptr_t moduleGetFunction(intptr_t module, str funcname) except? 0
 cpdef intptr_t moduleGetGlobal(intptr_t module, str varname) except? 0
+cdef unsigned int moduleGetFunctionCount(intptr_t module) except +
+cdef vector[Function] moduleEnumerateFunctions(  # NOQA
+    intptr_t module, unsigned int count) except +
```
seberg:
```suggestion
cdef unsigned int moduleGetFunctionCount(intptr_t module)
cdef vector[Function] moduleEnumerateFunctions(
    intptr_t module, unsigned int count) except *
```
Interestingly, I guess Cython must have added the correct non `except +`, default `except` clause anyway.  (The `except *` is also the default)

cupy_backends/cuda/api/driver.pyx
```diff
@@ -257,6 +258,63 @@ cpdef intptr_t moduleGetGlobal(intptr_t module, str varname) except? 0:
...
+
+    with nogil:
+        status = cuModuleEnumerateFunctions(
+            result.data(), count, <Module>module)
+    check_status(status)
+
+    return result
+
+
+cdef const char* funcGetName(intptr_t func) except? NULL:
```
seberg, line 297: `except? NULL` should be the default now, but OK to spell out.

murdock-grewar:
I notice a broader pattern with `cupy` that there is a lot of friction when trying to use `cupy` with separately maintained `.cu` source files on the filesystem. One issue is that the `.cu` files cannot be compiled with local inclusions, e.g. `#include local_header.h`, because `cupy` copies the `.cu` file alone to a temporary directory, missing out on any local dependencies (see also #8450). 

I'm encountering a related issue when generating custom `np.dtype`s on-the-fly, which I generate `struct` code for (similar to [this example file](https://github.com/cupy/cupy/blob/main/examples/custom_struct/complex_struct.py)). Because `cupy` only supports limited parameter types for kernels out-of-the-box, the code complexity of CUDA kernels is difficult to manage without custom data structures. Implementing these custom data structures requires `C` code generation, mainly in the form of creating new `struct`s. It would be easier to maintain that code if these custom `struct`s could be kept in their own header files, but then the caching issue arises. 

It's not necessarily impossible to work around these issues, but it makes it difficult to maintain readable code, including `.cu` files on the system which can be statically analyzed. 

> We do not have a way to parse all user-included headers and we probably should not (it hurts perf). But if we don't, the CUBIN hash would not be fully accurate. This PR does not solve that case either. I don't have good suggestions other than documenting the limitation.

Maybe you could include something like a `hashfiles` keyword in the `cupy.RawModule` and `cupy.RawKernel` functions. The user could pass in a tuple of relevant files which should be included in the hash. E.g.
```
import cupy as cp
module = cp.RawModule(code='...', hashfiles=('custom_struct.h'.))
```
However, it seems that it would be better to encourage the user to maintain actual `.cu` files which can be edited and statically analyzed. Maybe a new syntax is called for, like
```
import cupy as cp
module = cp.RawModule(sourcepath='./module.cu')
```
This would reduce the friction of maintaining separate `.cu` files (which a user ought to be doing anyway, if they can) and would open up the possibility of doing static analysis on the `.cu` file to determine its dependencies and include them in the hash.

seberg (commented):
A few comments many leave it or take it. I am still a bit unsure about the approach, but it seems fine.

cupy/cuda/compiler.py
leofang: - We do not offer Cython APIs; any Cython-level APIs that work today are only happy coincidence 🙂 
- I in particular do not want to expose any public APIs in this PR, at least for now. I am thinking cuda-core support and public exposure should probably go there.

cupy/cuda/function.pyx
leofang: I am surprised too. Turns out that the bot was pretty stubborn on this, but with a good reason: This is exactly what C++ demanglers do (same with the CUDA C++ demangler):
```shell
$ c++filt 12345
12345
```
that is, if something cannot be demangled, the original string is returned as-is.

cupy/cuda/function.pyx
leofang: This is really to cope with the fact that Thrust/CUB have internal, hidden namespaces to encode the ABI information. It is unusual for average C++ libraries.

cupy/cuda/function.pyx
leofang: One reason is that the driver APIs loaded via SoftLink could raise (if func ptrs do not exist). I forget if there is other reason.

cupy_backends/cuda/api/driver.pyx
leofang: Probably just to save one API call, I don't recall 😅

cupy/cuda/function.pyx
leofang: Another reason is: Here, we swallow all issues since nothing is user-actionable. If we cannot load for any reason, this function becomes no-op (does not set the mapping dict) which would trigger re-compilation on the call site.

cupy/cuda/function.pyx
leofang: Done in commit a6b7614.

cupy/cuda/function.pyx
leofang: Done in commit a6b7614.

cupy_backends/cuda/api/driver.pyx
leofang: Done in commit a6b7614.

cupy_backends/cuda/api/driver.pxd
leofang: Done in commit a6b7614.

leofang:
This PR is ready for review. The Windows failure is already fixed in #9902.

cupy/cuda/function.pyx
```diff
@@ -207,6 +380,56 @@ cdef class Module:
...
+
+                for i in range(function_handles.size()):
+                    mangled_cstr = driver.funcGetName(
+                        <intptr_t>(function_handles[i]))
+                    mangled_str = mangled_cstr.decode('utf-8')
+                    demangled = demangle_cxx_name(mangled_cstr, mangled_str)
+                    normalized = normalize_name_expr(demangled)
+                    original = norm_to_user.get(normalized)
+                    if original is not None:
+                        mapping[original] = mangled_str
```
seberg, line 427: Claude itself also found this.  If `original is None` for any reason, then the final `mapping` has no entry `mapping["original"]` which could be required later on, though.

Could be fixed by just returning `None` here, or at the call site check that the length of the mapping matches the length of the `name_expressions`.

cupy/cuda/function.pyx
```diff
@@ -18,6 +19,178 @@ from cupy.cuda.texture cimport TextureObject, SurfaceObject
...
+                return (_cupy_demangle != NULL);
+            _cupy_cufilt_loaded = 1;
+
+            /* Locate cupy_cufilt.dll next to this extension module. */
+            if (!GetModuleHandleExA(
+                    GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
+                    GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+                    (LPCSTR)&_cupy_cufilt_init, &hSelf))
+                return 0;
+            if (!GetModuleFileNameA(hSelf, path, MAX_PATH))
```
seberg: Claude is saying that `GetModuleFileNameW` is better and that `MAX_PATH` is in theory insufficient on some windows builds?  I.e. it wants a dance to check if the buffer size is insufficient.

I am not sure that is plausible, but it may make sense to at least check if the result `< MAX_PATH` (in which case truncation occurred).
At least the `...W` version actually returns the true required size.

cupy/cuda/function.pyx
```diff
@@ -18,6 +19,178 @@ from cupy.cuda.texture cimport TextureObject, SurfaceObject
...
+        static int _cupy_cufilt_loaded = 0;
+
+        static int _cupy_cufilt_init(void) {
+            HMODULE hSelf, hDll;
+            char path[MAX_PATH];
+            char *sep;
+
+            if (_cupy_cufilt_loaded)
+                return (_cupy_demangle != NULL);
+            _cupy_cufilt_loaded = 1;
```
seberg: Claude is rightly pointing out that this isn't thread-safe.  It's suggesting (not XP, but I guess that is fine) to use a:
```
    InitOnceExecuteOnce(&_cupy_cufilt_init_once, _cupy_cufilt_do_init,
                        NULL, NULL);
```
pattern to call this function. At the start of the function after prodding it a bit (`std::atomics` would also work to write this, but if this is fine, it seems pretty nice...).

Your bot can probably come up with something decent.  Alternative, it's prbably also enough to just move this to the end of the function and set both pointers via:
```
InterlockedExchangePointer((void**)&_cupy_dfree, (void*)local_free);
```
(so that the pointer setting is atomic and they are guaranteed set by the time `_cupy_cufilt_loaded` is visible as 1)

seberg (commented):
Generally, looks good, I am still a bit wondering if integrating it into the file isn't better.
However, this doesn't block us from doing that in the future if we get into problems and it should be a pretty big improvement!

I'll be curious if we see speed improvements in CI here on re-runs now (that would be awesome and also a decent test that it works).

Claude found some nitpicks that I agree with, though (to me, I wonder if the try/except should be on the callers around `self._enumerate_and_build_mapping()` but either way seems fine).

cupy/cuda/function.pyx
leofang: Good catch. Added a `len(mapping) == len(name_expressions)` guard in commit 1a6ac9c — if the mapping is incomplete, we don't set it, which triggers recompilation via the original code path (and produces proper error messages for truly missing symbols).

cupy/cuda/function.pyx
leofang: Switched the entire Windows path to wide-char APIs in commit 1a6ac9c: `GetModuleHandleExW`, `GetModuleFileNameW`, `LoadLibraryW`, `wcsrchr`, `wcscat`, etc. Also added a `len >= MAX_PATH` truncation check on the return value. (`GetProcAddress` stays ANSI since function export names are always ASCII.)

cupy/cuda/function.pyx
leofang: Replaced the manual `_cupy_cufilt_loaded` flag with `std::call_once` / `std::once_flag` in commit 1a6ac9c. The init function is now `void _cupy_cufilt_do_init(void)` (called exactly once), and `cupy_demangle` checks `_cupy_demangle != NULL` after `call_once` returns. This is portable (works on both Windows and Linux) and avoids the TOCTOU race.

seberg:
Close/re-open again, let's see if readthedocs recovers...

seberg (approved):
Thanks, nice to use an output parameter for the function return there.  Let's give this a shot (I am curious if it affects CI).

leofang:
Test failure is irrelevant.

/test force-skip

leofang:
> (I am curious if it affects CI).

Let's monitor the main branch. The per-PR CI runs don't seem to get faster much.

Issue #9239, opened 2025-07-13
Hypergeometric nsample = 0 returns ngood


njwfish:
### Description

This is a fairly simple issue, I think ideally it would be fixed in the rk_hypergeometric kernel. Running this:

```
import cupy as cp
N = cp.broadcast_to(cp.array(100), (2, 2))
B_1 = cp.broadcast_to(cp.array(30), (2, 2))
p = cp.broadcast_to(cp.array(.5), (2, 2))
N_t = cp.random.binomial(N, p, dtype=cp.int32)
B_t = cp.random.hypergeometric(
    ngood=B_1,
    nbad=N - B_1,
    nsample=0 * N_t, 
    dtype=cp.int32
)
```
Returns:

```
[[30 30]
 [30 30]]
```

And it should pretty clearly return zero, consistent with the intuition and the numpy implementation.

### To Reproduce

```py
# Write the code here
```

### Installation

Wheel (`pip install cupy-***`)

### Environment

```
# Paste the output here
```

kmaehashi:
Thanks, in case `nsample` is scalar we can check and fail if `nsample<1` case. PR welcomed.

PR #9805, opened 2026-03-13
9804: Prevent hypergeometric infinite loops and other consequences of invalid inputs

### LINKED ISSUE #9804: Hypergeometric does not prevent invalid input; worst case causes infinite loop
### Description

Any hypergeometric call where `nsample > (ngood + nbad)`, and `nsample > 10`, will result in an infinite loop.

https://github.com/cupy/cupy/blob/9c85c501f29c7e2283e61c1aeea9d0b4be9fa49c/cupy/random/cupy_distributions.cu#L334

dispatches to `rk_hypergeometric_hrua` without performing any checks for valid inputs. Then

https://github.com/cupy/cupy/blob/9c85c501f29c7e2283e61c1aeea9d0b4be9fa49c/cupy/random/cupy_distributions.cu#L296

`m` goes negative, `d11` goes negative, and then `W` is always negative, so 

https://github.com/cupy/cupy/blob/9c85c501f29c7e2283e61c1aeea9d0b4be9fa49c/cupy/random/cupy_distributions.cu#L313

the `(W < 0.0)` is always `true`, we always hit `continue`, and we keep looping forever.

Whether this loops infinitely in practice depends on whether there's a watchdog timer keeping an eye on things. For T4s I don't think one exists, so this never times out, but RTX3090 eventually killed it with a `CUDA_ERROR`.

Numpy does a bunch of python-side validation, e.g.
https://github.com/numpy/numpy/blob/c2702141082ef1ba1cfad198994ac362a020365e/numpy/random/_generator.pyx#L3616

We should probably just add the same. I can raise a PR.

### To Reproduce

```py
cp.random.hypergeometric(10, 10, 21, size=5)
```

### Installation

### Environment

```
# Paste the output here
```

### Additional Information

Kinda related to https://github.com/cupy/cupy/issues/9239 - we just need to match numpy's input validation to prevent invalid inputs.

jberg5:
Should fix https://github.com/cupy/cupy/issues/9804 and also https://github.com/cupy/cupy/issues/9239.

Context: If `nsample > (ngood + nbad)` and `nsample > 10` we would end up in an infinite loop. Good news is that that's invalid input, and numpy just checks this case and some others before running the kernel.

I basically tried to port numpy's checks here. Added both Python-level and kernel level input validation for defense in depth, but I'm happy to change the approach though. I understand raising validation exceptions from Python at that point will sync the device, but given that this only happens on invalid input, I think it's ok?

Full disclosure: I had claude (opus 4.6) write the test cases. I think they look pretty good and match the overall style, but happy to tweak!

**EDIT(seberg): Marked as no-compat, since the random stream could change here, just for a brief note in release notes.**

### DIFF
--- a/cupy/random/_kernels.py
+++ b/cupy/random/_kernels.py
@@ -791,8 +791,7 @@ __device__ long rk_hypergeometric(
     rk_state *state, long good, long bad, long sample)
 {
     if (sample <= 0) return 0;
-    if (sample >= good + bad) return good;
-    if (sample > 10)
+    if ((sample > 10) && (sample <= good + bad - 10))
     {
         return rk_hypergeometric_hrua(state, good, bad, sample);
     } else
--- a/cupy/random/_kernels.py
+++ b/cupy/random/_kernels.py
@@ -790,6 +790,8 @@ __device__ long rk_hypergeometric_hrua(
 __device__ long rk_hypergeometric(
     rk_state *state, long good, long bad, long sample)
 {
+    if (sample <= 0) return 0;
+    if (sample >= good + bad) return good;
     if (sample > 10)
     {
         return rk_hypergeometric_hrua(state, good, bad, sample);
--- a/tests/cupy_tests/random_tests/test_generator.py
+++ b/tests/cupy_tests/random_tests/test_generator.py
@@ -245,29 +245,30 @@ class TestHypergeometric(
     pass
 
 
-class TestHypergeometricValidation(unittest.TestCase):
+class TestHypergeometricValidation:
 
-    def setUp(self):
-        self.rs = _generator.RandomState(seed=testing.generate_seed())
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.rs = _generator.RandomState(seed=0)
 
     def test_hypergeometric_ngood_negative(self):
-        with self.assertRaises(ValueError):
+        with pytest.raises(ValueError):
             self.rs.hypergeometric(-1, 10, 5, size=10)
 
     def test_hypergeometric_nbad_negative(self):
-        with self.assertRaises(ValueError):
+        with pytest.raises(ValueError):
             self.rs.hypergeometric(10, -1, 5, size=10)
 
     def test_hypergeometric_nsample_zero(self):
-        with self.assertRaises(ValueError):
+        with pytest.raises(ValueError):
             self.rs.hypergeometric(10, 10, 0, size=10)
 
     def test_hypergeometric_nsample_negative(self):
-        with self.assertRaises(ValueError):
+        with pytest.raises(ValueError):
             self.rs.hypergeometric(10, 10, -1, size=10)
 
     def test_hypergeometric_nsample_too_large(self):
-        with self.assertRaises(ValueError):
+        with pytest.raises(ValueError):
             self.rs.hypergeometric(5, 10, 16, size=10)
 
     def test_hypergeometric_nsample_equals_total(self):
--- a/tests/cupy_tests/random_tests/test_generator.py
+++ b/tests/cupy_tests/random_tests/test_generator.py
@@ -245,6 +245,45 @@ class TestHypergeometric(
     pass
 
 
+class TestHypergeometricValidation(unittest.TestCase):
+
+    def setUp(self):
+        self.rs = _generator.RandomState(seed=testing.generate_seed())
+
+    def test_hypergeometric_ngood_negative(self):
+        with self.assertRaises(ValueError):
+            self.rs.hypergeometric(-1, 10, 5, size=10)
+
+    def test_hypergeometric_nbad_negative(self):
+        with self.assertRaises(ValueError):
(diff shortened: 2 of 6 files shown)


cupy/random/_kernels.py
```diff
@@ -790,6 +790,8 @@
 __device__ long rk_hypergeometric(
     rk_state *state, long good, long bad, long sample)
 {
+    if (sample <= 0) return 0;
+    if (sample >= good + bad) return good;
```
seberg: Isn't this unnecessary now with the checks coming in? (also below?)

tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -245,6 +245,45 @@ class TestHypergeometric(
     pass
 
 
+class TestHypergeometricValidation(unittest.TestCase):
```
seberg:
```suggestion
class TestHypergeometricValidation:
```
We try to phase out unit tests, so let's just not use it (it's not like the actual seed matters for these tests, but `setUp` should be called much the same way by pytest).

tests/cupy_tests/random_tests/test_generator_api.py
```diff
@@ -292,6 +292,45 @@ class TestHypergeometric(
     pass
 
 
+class TestHypergeometricValidation(unittest.TestCase):
```
seberg:
```suggestion
class TestHypergeometricValidation:
```

seberg (commented):
Thanks, the additional checks look good. Just one question if it's worthwhile to add the branching into the kernels now?

The tests look fine, could parametrize for briefness, but they seem brief enough to me.

tests/cupy_tests/random_tests/test_generator.py
jberg5: Happy to hear that, I personally prefer pytest too :) will update

cupy/random/_kernels.py
jberg5: Theoretically yes. I'm intentionally redundant here because 1) numpy does it and 2) defense in depth in case someone adds some code elsewhere in cupy that calls directly into this kernel for some reason. Let me know what you think - happy to remove, I can also just replace with a comment saying that callers are responsible for verifying inputs are valid

jberg5: This is the specific bit of numpy validation I was looking at: https://github.com/numpy/numpy/blob/80bcb8b03ed0b2699901e20aafaa149c9ec5a527/numpy/random/src/distributions/random_hypergeometric.c#L251

seberg: Yeah OK. I suspect it is fair to assume that these checks are pretty irrelevant compared to the whole kernel speed.

seberg (dismissed):
Thanks, LGTM, and synchronizing checks ahead of time is normal behavior for rngs.

cupy/random/_kernels.py
seberg: Hmmm, the NumPy code just uses the second branch also for `sample <= good + bad - 10` rather than just avoiding the out of range?

Might be a bit nicer then?  This will change the random number stream though for cases that take the branch now but didn't earlier.
Or am I missing something about it?  I don't think CuPy has a very strict policy about that though (and we can make a short note in the release notes).

jberg5: Agreed it would be a bit nicer. Initially I didn't want to change the condition too much, but I think you're right that changing it and routing some more cases to the other kernel should be fine.

seberg (approved):
Thanks, went through once more, and while spelled differently, the algorithms are identical to NumPy, so I think it makes sense to just align.
I'll just leave the branch, since I don't think it should matter compared to whatever comes later.

leofang:
> I understand raising validation exceptions from Python at that point will sync the device, but given that this only happens on invalid input, I think it's ok?

Sorry I was out and just started catching up. We rarely do input validation that synchronizes the device. This would kill the performance entirely. It is also not true that this "only happens" on invalid input -- the new checks always synchronize regardless if the inputs are invalid or not. Can we think of a better way?

For example, @kmaehashi suggested (https://github.com/cupy/cupy/issues/9239#issuecomment-3067817920) that we check for `nsamples`<1 if it is a scalar (on host), which would be a nice check to have that does not synchronize. We need to do better than this PR currently does. I suggest that we revert this PR for now and work on a better approach. Thoughts @seberg @jberg5?

jberg5:
Hey @leofang thanks for the feedback! Totally fair, I misunderstood the flow on the host side. Is this closer to what you had in mind? https://github.com/cupy/cupy/pull/9822

seberg:
> We need to do better than this PR currently does. I suggest that we revert this PR for now and work on a better approach. Thoughts

Ah sorry, the random module does it for a few functions, but maybe we should avoid introducing it then.
One could return an invalid result (assuming that is very cheap), but since it isn't `NaN`, dunno how nice it is -1?).

Just doing it for scalars seems fine (and honestly the most common path), a bit suprising maybe that it doesn't kick in always, but as it is probably by far the most common than an array input.

PR #9807, opened 2026-03-17
MAINT,BUG: cleanup pending, simplify PooledMemory, use pymutex

seberg:
This does remove the pinned memory pool from the pxd, but a code search found a single use in a Cythn file that didn't cimport it. I'll wager it's fine, there is no good reason to cimport this or use it directly?
(One reason is that Cython seems to have a subtle bug leading to compile issues).

Otherwise, make sure to clean up pending free every time. I first thought it's pretty harmless but allows the pool to grow unnecessarly and the CuPy pool being too full can interfere with out of pool allocations.


cupy/cuda/memory.pyx
```diff
@@ -1165,6 +1164,9 @@ cdef class _Arena:
         cdef std_set[index_type].iterator it
         cdef _Chunk chunk = None
 
+        if len(self._pending_free):
+            self._commit_pending_free()
```
seberg, line 1169: I had forgotten this.  As mentioned, we have to prevent this from growing indefinitely, it seems fast enough to just always check.

cupy/cuda/pinned_memory.pxd
```diff
@@ -15,32 +14,8 @@ cdef class PinnedMemoryPointer:
...
-        # NOTE: Never use `lock()` outside a nogil statement, because
-        # almost anything could release the GIL and then deadlocks happen
-        # if another thread tries to lock also (without the GIL released).
-        # You can try with `try_lock()` first though.
-        recursive_mutex _lock
-
-    cpdef PinnedMemoryPointer malloc(self, size_t size)
-    cpdef free(self, intptr_t ptr, size_t size)
-    cpdef free_all_blocks(self)
-    cpdef n_free_blocks(self)
```
seberg, line 43: I think I mentioned it, a github code search found no cython usage of this and I don't see a reason why anyone would do so (it's public more conveniently and documented in Python).

This also changed `free()` to be `cdef` only.  Same as for the memory pool.

seberg (commented):
General note: Adding to lists is atomic, and we always hold a lock when removing. So lists can change while we delete, but only by adding new elements at the end.

seberg:
Ack, forgot to put in a last fixup.

/test mini

cupy/cuda/pinned_memory.pyx
```diff
@@ -104,14 +106,15 @@ cdef class PinnedMemoryPointer:
         buffer.itemsize = 1
         buffer.len = size
         buffer.ndim = 1
-        buffer.obj = self
         buffer.readonly = 0
         buffer.shape = self._shape
         buffer.strides = self._strides
         buffer.suboffsets = NULL
+        cpython.Py_INCREF(self)
+        buffer.obj = self
```
seberg, line 114: Nasty little bug.  Changing the `free()` logic to remove `_in_use` surfaced that the ref-counting was wrong.  This made multi-threaded tests crash hard, but didn't affect typical runs bad enough...

(E.g. some fft tests, I am not sure exactly which path, maybe via `_pinned_array.py` utils.)

cupy/cuda/memory.pyx
```diff
@@ -1235,6 +1238,7 @@ cdef class SingleDeviceMemoryPool:
         dict _arenas
         # NOTE: Never use `.lock()` outside a `with nogil:` statement as it
         # may deadlock (GIL may be unlocked and another thread also locks).
+        # We currently use a mutex for the ability to use `try_lock`.
```
seberg, line 1241: I am not sure the pattern is worth it or not, but without much profiling, I'll just default to "no change is just as well".

seberg:
This is blocking (not for the pooledmemory changes so much, but for the `_commit_pending_free()` so I could pull that out if desired).

cupy/cuda/pinned_memory.pyx
```diff
@@ -299,52 +297,39 @@ cdef class PinnedMemoryPool:
...
-        finally:
-            self._lock.unlock()
+    cdef free(self, mem, size):
+        cdef list free = self._free.get(size)
+        if free is None:
+            free = self._free.setdefault(size, [])
+        # OK to append to list (atomic) while another threads may pop().
+        free.append(mem)
+
+        self._in_use.remove(mem)
```
leofang, line 311: Q: we don't need to hold a lock here? What if thread 1 mallocs and thread 2 frees (of the same `mem`)?

seberg: Ah, good spot!  I think we should be OK without a lock (I am slightly worried about dead-locks if there is, although a critical section would likely be right as well.)

But, I think we have to swap these two.  It should be fine if the block is not `_in_use`, but not yet available for re-use via the `free` list.  (But it is not fine in this order).

(Also means double free errors on the `remove()` before anything else weird might happen.)

leofang (approved):

PR #9809, opened 2026-03-17
TST: Fixup some more tests (mainly cupyx) for free-threading

seberg:
Fixes a few smaller things, mainly two categories:
* Accidental changes of `self.` which are incompatible but commonly arise for non-unittest custom parametrization.
* Use of `RandomState` rather than global seeding.

And a few smaller ones.  This covers cupyx, but two other small things flushed up when I was running things now.
(Since this can be flaky, there is a bit of a long tail.)

### DIFF
--- a/tests/cupyx_tests/profiler_tests/test_timeit_magic.py
+++ b/tests/cupyx_tests/profiler_tests/test_timeit_magic.py
@@ -65,6 +65,7 @@ def _verify_benchmark_output(output):
         number_pattern, output), "Output should contain timing numbers"
 
 
+@pytest.mark.thread_unsafe(reason="ipython_shell fixture not thread-safe")
 class TestGPUTimeitMagic:
     """Test %gpu_timeit magic functionality."""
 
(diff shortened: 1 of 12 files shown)


tests/cupyx_tests/profiler_tests/test_timeit_magic.py
```diff
@@ -65,6 +65,7 @@ def _verify_benchmark_output(output):
         number_pattern, output), "Output should contain timing numbers"
 
 
+@pytest.mark.thread_unsafe(reason="ipython_shell fixture not thread-safe")
```
leofang, line 68: Interesting... I did not know that!

leofang (approved):
LGTM!

tests/cupyx_tests/profiler_tests/test_timeit_magic.py
seberg: Oh, it may be that it works for some tests. I forgot (and didn't look super closely), it may well be that it was more about one test seeing the output from another (since they would share the same shell).

PR #9810, opened 2026-03-18
ENH,BUG: Replace `int` with `size_t` in cub reduce/segmented reduce/scan

### LINKED ISSUE #9780: Illegal memory access in cupy.sum (cupy 13.6.0)
### Description

`cupy.sum()` crashes on H200 for large arrays. (cupy 13.6)

```
  File "/nfs/roberts/project/pi_tz324/cgh42/si-gpu/err2.py", line 4, in <module>                                        
    cp.asnumpy(b)
  File "/nfs/roberts/project/pi_tz324/shared/software/cupy/13.6.0/cupy/__init__.py", line 807, in asnumpy               
    return a.get(stream=stream, order=order, out=out, blocking=blocking)                                                
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                
  File "cupy/_core/core.pyx", line 1873, in cupy._core.core._ndarray_base.get                                           
  File "cupy/_core/core.pyx", line 1978, in cupy._core.core._ndarray_base.get                                           
  File "cupy/cuda/memory.pyx", line 690, in cupy.cuda.memory.MemoryPointer.copy_to_host_async                           
  File "cupy_backends/cuda/api/runtime.pyx", line 636, in cupy_backends.cuda.api.runtime.memcpyAsync                    
  File "cupy_backends/cuda/api/runtime.pyx", line 146, in cupy_backends.cuda.api.runtime.check_status                   
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered      
[cgh42@a1124u19n01.bouchet si-gpu]$ vim err2.py                                                                         
[cgh42@a1124u19n01.bouchet si-gpu]$ python err2.py                                                                      
[cgh42@a1124u19n01.bouchet si-gpu]$ vim err2.py                                                                         
[cgh42@a1124u19n01.bouchet si-gpu]$ python err2.py                                                                      
  File "/nfs/roberts/project/pi_tz324/cgh42/si-gpu/err2.py", line 4, in <module>                                        
    cp.asnumpy(b)
  File "/nfs/roberts/project/pi_tz324/shared/software/cupy/13.6.0/cupy/__init__.py", line 807, in asnumpy               
    return a.get(stream=stream, order=order, out=out, blocking=blocking)                                                
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                
  File "cupy/_core/core.pyx", line 1873, in cupy._core.core._ndarray_base.get                                           
  File "cupy/_core/core.pyx", line 1978, in cupy._core.core._ndarray_base.get                                           
  File "cupy/cuda/memory.pyx", line 690, in cupy.cuda.memory.MemoryPointer.copy_to_host_async                           
  File "cupy_backends/cuda/api/runtime.pyx", line 636, in cupy_backends.cuda.api.runtime.memcpyAsync                    
  File "cupy_backends/cuda/api/runtime.pyx", line 146, in cupy_backends.cuda.api.runtime.check_status                   
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered
```

### To Reproduce

```py
import cupy as cp                                                                                                       
j3c = cp.zeros((40000, 400, 200), dtype=cp.float32)
b = j3c.sum(axis=2)
cp.asnumpy(b)
```

### Installation

Built from GitHub source

(Also tried `pip install cupy-cuda12x` in fresh venv)

### Environment

```
OS                           : Linux-4.18.0-553.87.1.el8_10.x86_64-x86_64-with-glibc2.28                                
CuPy Version                 : 13.6.0
SciPy Version                : 1.17.1
Cython Build Version         : 3.1.8
CUDA Root                    : /apps/software/2024a/software/CUDA/12.6.0
nvcc PATH                    : /apps/software/2024a/software/CUDA/12.6.0/bin/nvcc
CUDA Build Version           : 12060
CUDA Driver Version          : 12080
CUDA Runtime Version         : 12060 (linked to CuPy) / 12060 (locally installed)
cuBLAS Version               : 120600
cuFFT Version                : 11206
cuRAND Version               : 10307
cuSOLVER Version             : (11, 6, 4)
cuSPARSE Version             : 12502
NVRTC Version                : (12, 6)
Jitify Build Version         : 1a0ca0e
NCCL Build Version           : 22602
cuTENSOR Version             : 20500
cuSPARSELt Build Version     : 801
Device 0 Name                : NVIDIA H200
Device 0 Compute Capability  : 90
Device 0 PCI Bus ID          : 0000:CB:00.0
```

### Additional Information

Build process:
```
export CUPY_NVCC_GENERATE_CODE="arch=compute_90,code=sm_90"
python -m build --wheel --no-isolation
```

seberg:
This replaces the `int` size related parameters with `size_t` parameters in CUB to remove call limitation and fix existing missing check.

Commit created in part with cursor.

---

Closes gh-9780, **note: if there is any uncertainty here or seems nontrivial to review, as discussed elsewhere: we can just close this and I'll instead do the one-line fix from the issue.  (the reasoning is that mid to long-term we'll want a more comprehensive fix anyway)**

### DIFF
--- a/tests/cupy_tests/math_tests/test_sumprod.py
+++ b/tests/cupy_tests/math_tests/test_sumprod.py
@@ -409,46 +409,58 @@ class TestReductionSizeOverInt32Max:
         cupy.get_default_memory_pool().free_all_blocks()
         cupy.get_default_pinned_memory_pool().free_all_blocks()
 
-    @pytest.mark.parametrize('shape,axis', [
-        ((INT32_MAX + 1024,), None),
-        ((4, 2**30 + 512), 1),
-        ((INT32_MAX + 1024, 2), 0),
-    ], ids=['full', 'axis1', 'axis0'])
+    @pytest.mark.parametrize('shape,axis,part', [
+        ((INT32_MAX + 1024,), None, "first-part"),
+        ((4, 2**30 + 512), 1, "second-part"),
+        ((INT32_MAX + 1024, 2), 0, "first-part"),
+    ])
     @pytest.mark.parametrize('dtype', [
         numpy.int8, numpy.int32, numpy.float32,
     ])
-    def test_sum_max_min_prod(self, shape, axis, dtype):
+    def test_sum_max_min_prod(self, shape, axis, dtype, part):
         try:
             a = cupy.ones(shape, dtype=dtype)
             # Make first and last element along each slice interesting
             if axis is None:
                 a[[0, -1]] = [3, -1]
             elif axis == 0:
-                a[[0, -1], :] = [3, -1]
+                a[[0, -1], :] = [[3], [-1]]
             else:
-                a[:, [0, -1]] = [3, -1]
+                a[:, [0, -1]] = [[3, -1]]
 
-            if axis is None:
-                # Full reduction: one segment, one 2 and (size-1) ones
-                assert a.sum() == a.size
-                assert a.max() == 3
-                assert a.min() == -1
-                assert a.prod() == -3
+            # Test only half of the reductions per test for better speed
+            # (it is still very slow.)
+            if part == "first-part":
+                if axis is None:
+                    # Full reduction: one segment, one 2 and (size-1) ones
+                    assert a.sum() == a.size
+                    assert a.max() == 3
+                    assert a.argmin() == a.size - 1
+                else:
+                    s = a.sum(axis=axis)
+                    expected_sum = shape[axis]
+                    testing.assert_array_equal(s, cupy.full(
+                        s.shape, expected_sum, dtype=s.dtype))
+                    testing.assert_array_equal(
+                        a.max(axis=axis), cupy.full(s.shape, 3, dtype=dtype))
+                    testing.assert_array_equal(
+                        a.argmin(axis), cupy.full(s.shape, a.shape[axis] - 1))
             else:
-                s = a.sum(axis=axis)
-                expected_sum = shape[axis]
-                testing.assert_array_equal(s, cupy.full(
-                    s.shape, expected_sum, dtype=s.dtype))
-                testing.assert_array_equal(
-                    a.max(axis=axis), cupy.full(s.shape, 3, dtype=dtype))
-                testing.assert_array_equal(
-                    a.min(axis=axis), cupy.full(s.shape, -1, dtype=dtype))
-                testing.assert_array_equal(
-                    a.prod(axis=axis), cupy.full(s.shape, -3, dtype=dtype))
+                if axis is None:
+                    # Full reduction: one segment, one 2 and (size-1) ones
+                    assert a.prod() == -3
+                    assert a.min() == -1
+                    assert a.argmax() == 0
+                else:
+                    p = a.prod(axis=axis)
+                    testing.assert_array_equal(p, cupy.full(
+                        p.shape, -3, dtype=p.dtype))
+                    testing.assert_array_equal(
+                        a.min(axis=axis), cupy.full(p.shape, -1, dtype=dtype))
+                    testing.assert_array_equal(
+                        a.argmax(axis), cupy.full(p.shape, 0))
         except MemoryError:
             pytest.skip("out of memory in test.")
-
-        cupy.get_default_memory_pool().free_all_blocks()
 
     @pytest.mark.parametrize('dtype', [numpy.int8, numpy.int32, numpy.float32])
     def test_cumsum_size_over_int32_max(self, dtype):
@@ -467,8 +479,6 @@ class TestReductionSizeOverInt32Max:
         except MemoryError:
             pytest.skip("out of memory in test.")
 
-        cupy.get_default_memory_pool().free_all_blocks()
-
     @pytest.mark.parametrize('dtype', [numpy.int8, numpy.int32, numpy.float32])
     def test_cumprod_size_over_int32_max(self, dtype):
         """CUB device_scan (cumprod) with size > INT32_MAX."""
@@ -481,8 +491,6 @@ class TestReductionSizeOverInt32Max:
             assert out[-1] == 6  # product of array
         except MemoryError:
             pytest.skip("out of memory in test.")
-
-        cupy.get_default_memory_pool().free_all_blocks()
 
 
 # This class compares cuTENSOR results against NumPy's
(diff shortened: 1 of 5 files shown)


seberg:
Nvm... I added a test and after improving it (the agent version wasn't very thorough) it now just fails and I noticed that `argmin/argmax` also needs fixing probably.
So, I think I'll go with the minimal version for now (although... hopefully the new tests pass _that_)

seberg:
OK, the issue was in the test itself so I'll give it one more chance and reject argmin/argmax for those paths.
The tests won't run in CI to a degree, they just need too much memory. But are passing locally (they are pretty slow though).

leofang (commented):
We need a performance study because CUB algorithms are extremely sensitive to the number of items (which is partly why it took a long time for large-size support to land in CUB). In the worst case scenario, we need to dispatch to two different kernel implementations and favor `int` over `size_t` unless `num_items` is too large, though it'd inflate the compile time and binary size, which we also need to study.

seberg:
Hmmm, good point, locally for summing it seems to be identical (unless it is hardware dependent -- for arange, I guess there is a bigger chance of a difference, but it is unchanged for now).  I'll try and confirm that CUB picks the best implementation at runtime

Actual timings in details (first run is new branch, but it is identical either way -- confirmed on T400 and RTX PRO6000):

<details>

```
In [3]: import cupy as cp

In [4]: a = cp.arange(1000000.)

In [5]: %timeit a.sum()
10.2 μs ± 0.455 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [6]: %timeit a.reshape(1000, 1000).sum(axis=0)
12.3 μs ± 0.544 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [7]: %timeit a.reshape(1000, 1000).sum(axis=1)
8.19 μs ± 0.514 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [8]: %timeit a.sum()
10.2 μs ± 0.44 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [9]: %timeit a.reshape(1000, 1000).sum(axis=0)
12.3 μs ± 0.527 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [10]: %timeit a.reshape(1000, 1000).sum(axis=1)
8.19 μs ± 0.434 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [11]:                                                                                                                          
Do you really want to exit ([y]/n)? 
(cupy-dev) sebastianb@sebastianb-workstation:~/forks/cupy/tests$ ipython
Python 3.14.3 | packaged by conda-forge | (main, Feb  9 2026, 21:56:02) [GCC 14.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.9.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: You can use `%hist` to view history, see the options with `%history?`

In [1]: import cupy as cp

In [2]: a = cp.arange(1000000.)

In [3]: %timeit a.sum()
10.2 μs ± 0.453 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit a.reshape(1000, 1000).sum(axis=0)
12.3 μs ± 0.535 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [5]: %timeit a.reshape(1000, 1000).sum(axis=1)
8.2 μs ± 0.541 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

</details>

It seems to me that larger indices are implemented by launching smaller kernels for the larger sizes, rather than two kernel types:
https://github.com/NVIDIA/cccl/pull/3764/changes#diff-7c890d0fb2241798e5e82f69cfc78e3308e2830ebcc13f5cec23f0c0a4785d3eR831

leofang (dismissed):
Approving since I will get disconnected soon and I don't wanna block this from merging 😛 but before merging it'd better to check perf regression on data center GPUs like A100, H100, and B100, because it's where CUB tuning really shines.

leofang:
btw when testing the perf we ~~show~~ **should** sweep through input sizes in log scale (10^2, 10^3, ...) because CUB performs differently in different regions.

seberg:
I ran a local benchmark and realized that the histogram doesn't actually work (and the size checks weren't removed anyway).

I would still need to a run a datacenter GPU benchmark.  Otherwise a local one showed no difference.  As mentioned above, I think that is expected: scan, reduce, segmented reduce look like they run in chunks of integer size so that the `OffsetT` doesn't affect the kernels themselves.

However... Trying to benchmark histograms ran into illegal memory address issues, so I pulled out those changes (it was incomplete anyway before).
 
(The most unclear is `scan` which I think [uses](https://github.com/NVIDIA/cccl/blob/bfb94f236150f94cae250b8bccf465cee51a399e/cub/cub/device/dispatch/dispatch_scan.cuh#L383-L384):
```
      int tile_size = policy.Scan().BlockThreads() * policy.Scan().ItemsPerThread();
      int num_tiles = static_cast<int>(::cuda::ceil_div(num_items, tile_size));
```
which seems pretty save as `tile_size` looks like a larger factor)

/test mini

tests/cupy_tests/math_tests/test_sumprod.py
```diff
@@ -387,6 +387,110 @@ def _mitigate_cumprod(self, xp, dtype, result):
...
+            if axis is None:
+                a[[0, -1]] = [3, -1]
+            elif axis == 0:
+                a[[0, -1], :] = [[3], [-1]]
+            else:
+                a[:, [0, -1]] = [[3, -1]]
+
+            # Test only half of the reductions per test for better speed
+            # (it is still very slow.)
+            if part == "first-part":
```
leofang, line 431: typo

leofang:
As discussed offline, let's close this one in favor of #9867 (for reasons discussed both above and there).

seberg:
Yeah, OK, let's close it.  We can revive it when someone wants to spend some time and dive deeper to ensure no perf regression.

PR #9812, opened 2026-03-19
BUG: Don't use `--device-as-default-execution-space` for hip

seberg:
We removed a version check here, but running rocm tests once showed that this option isn't supported there.


seberg:
:shrug: there are more failures, but I think some have open PRs, maybe others not. But it does fix the issue it set out to fix...

leofang:
Yeah I've noticed too that the ROCm CI has been failing for a long time. Not sure if @gpinkert has time to help us fix it 😅 but as long as it improves the situation we should merge.

leofang (approved):

PR #9815, opened 2026-03-19
Fix `delete` incompatibilities with NumPy

### LINKED ISSUE #9035: `cupy.delete` behaves differently with `numpy.delete`
### Description

`cupy.delete` behaves differently with `numpy.delete`

### To Reproduce

```py
import numpy as np
import cupy as cp

print(np.delete(np.array([1,2,3]),(0,True),axis=None))
print(cp.delete(cp.array([1,2,3]),(0,True),axis=None))
```
```bash
[3]
[2 3]
```

### Installation

### Environment

```
OS                           : Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
CuPy Version                 : 13.4.0
NumPy Version                : 2.2.3
SciPy Version                : 1.15.2
CUDA Build Version           : 12080
CUDA Runtime Version         : 12080 (linked to CuPy) / 12020 (locally installed)
cuBLAS Version               : 120201
cuSOLVER Version             : (11, 5, 2)
cuSPARSE Version             : 12101
Device 0 Name                : NVIDIA GeForce GTX 1650
```

### Additional Information

_No response_

dufp2002:
This PR fixes `cupy.delete` compatibility issues for:
- boolean mask inputs
- empty Python list/tuple `obj`
- bool scalar error behavior
- mixed tuple input such as `(0, True)`

It also adds tests covering these cases.

Fixes #9035

### DIFF
--- a/cupy/_manipulation/add_remove.py
+++ b/cupy/_manipulation/add_remove.py
@@ -9,39 +9,14 @@ import math
 from cupy import _core
 
 
-def _delete_mask(obj, size):
-    if isinstance(obj, slice):
-        mask = cupy.ones(size, dtype=cupy.bool_)
-        mask[obj] = False
-        return mask
-
-    raw_obj = obj
-    obj = cupy.asarray(obj)
-
-    if obj.dtype == cupy.bool_:
-        if obj.ndim != 1 or obj.shape[0] != size:
-            raise ValueError(
-                'boolean array argument obj to delete must be one '
-                f'dimensional and match the axis length of {size}')
-        return ~obj
-
-    # Match NumPy behavior for empty Python lists/tuples.
-    if obj.size == 0 and not isinstance(raw_obj, cupy.ndarray):
-        obj = obj.astype(cupy.intp)
-
-    mask = cupy.ones(size, dtype=cupy.bool_)
-    mask[obj] = False
-    return mask
-
-
-def delete(arr, obj, axis=None):
+def delete(arr, indices, axis=None):
     """
     Delete values from an array along the specified axis.
 
     Args:
         arr (cupy.ndarray):
             Values are deleted from a copy of this array.
-        obj (slice, int or array of ints):
+        indices (slice, int or array of ints):
             These indices correspond to values that will be deleted from the
             copy of `arr`.
             Boolean indices are treated as a mask of elements to remove.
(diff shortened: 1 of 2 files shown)


tests/cupy_tests/manipulation_tests/test_add_remove.py
```diff
@@ -53,6 +58,87 @@ def test_delete_with_indices_as_int(self, xp):
...
+
+    @testing.numpy_cupy_array_equal()
+    def test_delete_with_int_array_obj(self, xp):
+        arr = xp.array([10, 20, 30])
+        return xp.delete(arr, xp.array([0, 2]))
+
+    @testing.numpy_cupy_array_equal(accept_error=IndexError)
+    def test_delete_with_float_array_obj_error(self, xp):
+        arr = xp.array([10, 20, 30])
+        xp.delete(arr, xp.array([0., 2.]))
```
seberg: It would be nice if you could parametrize at least some of these tests just to make them a bit shorter and quicker to read.
(I.e. adding `@pytest.parametrize()`.)

cupy/_manipulation/add_remove.py
```diff
@@ -34,25 +59,19 @@ def delete(arr, indices, axis=None):
...
-
-        arr = arr.ravel()
-
-        if isinstance(indices, cupy.ndarray) and indices.dtype == cupy.bool_:
-            return arr[~indices]
-
-        mask = cupy.ones(arr.size, dtype=bool)
-        mask[indices] = False
-        return arr[mask]
+    arr = cupy.asarray(arr)
```
seberg: Hmmm, the old code didn't call `asarray()` here.  Unlike NumPy, I think it may make sense to keep that as it was (roughly speaking):

```suggestion
    if not isinstance(arr, cupy.ndarray):
        raise ValueError('delete(): arr must be a cupy.ndarray')
```

cupy/_manipulation/add_remove.py
```diff
@@ -34,25 +59,19 @@ def delete(arr, indices, axis=None):
...
+        axis = _core.internal._normalize_axis_index(axis, arr.ndim)
 
-        if isinstance(indices, cupy.ndarray) and indices.dtype == cupy.bool_:
-            return cupy.compress(~indices, arr, axis=axis)
-
-        mask = cupy.ones(arr.shape[axis], dtype=bool)
-        mask[indices] = False
-        return cupy.compress(mask, arr, axis=axis)
+    size = arr.shape[axis]
+    mask = _delete_mask(obj, size)
```
seberg: Could you check whether the tests pass just the same if you only modify this one line of code to:
```
mask[indices,] = False
```
I think the only issue is that NumPy assumes a single index (makes some sense, as we work on a single axes), while cupy here doesn't.  Adding the `,` should fix that.
If that works, it would keep things much simpler even if the error message you created is very slightly nicer.

(NumPy has a bit of fast paths, but I doubt that is worthwhile.)

seberg (commented):
Thanks, sorry for the slightly late reply.  Overall, this looks nice, but I think the tests could be more concise.

I also suspect that we can actually simplify this and maybe don't even need the helper at all.  The insight is that `mask[obj]` and `mask[obj,]` are two very different things!

seberg:
OK, NumPy is a bit weird and we need some special handling for slice/ints (ideally) in order to raise the identical error for boolean inputs.  But otherwise, I think it makes sense to defer as much as possible to indexing anyway.

Did that, rejected non-cupy arrays (identical), undid the parameter rename (yes, matching NumPy with `obj` is nicer, but it is a silent change), and had my agent merge practically all new tests into a single parametrization.

seberg:
The test hit the NumPy regression, let's just re-run...

/test windows,cuda132

seberg (approved):
@eriknw maybe you can have a very brief glance so that I don't effectively self merge my code?

(Arg, sorry, I mistyped the username first :/!)

eriknw (approved):
LGTM! Thanks @dufp2002 (and @seberg). This does indeed fix the issue, and the code looks good to me. I have two things to note:
- I noticed a pre-existing bugs where out-of-bound indices, beyond [-size, size-1], give _silent wrong_ results in cupy, but numpy raises IndexError. Addressing this is probably beyond the scope of this PR
- It might be worth noting that cupy, unlike numpy, doesn't accept array-like values (only cupy arrays) and raises ValueError instead

seberg:
Thanks, putting it in then.

> I noticed a pre-existing bugs where out-of-bound indices, beyond [-size, size-1], give silent wrong results in cupy, but numpy raises IndexError. Addressing this is probably beyond the scope of this PR

Likely OK.  CuPy doesn't do this for indexing, because raise an error requires synchronization, so I guess it is the same divergence as for indexing.

PR #9822, opened 2026-03-22
Validate hypergeometric inputs without syncing

jberg5:
Improve upon https://github.com/cupy/cupy/pull/9805, which had more extensive host-side checks but synced the device on every call. Now we only check scalars, and leave it to the kernel dispatcher to handle anything else.

### DIFF
--- a/cupy/random/_generator.py
+++ b/cupy/random/_generator.py
@@ -254,19 +254,12 @@ class RandomState:
         """  # NOQA
         # Array inputs are not validated here; the kernel dispatcher
         # handles out-of-range values safely.
-        if isinstance(ngood, (int, float)):
-            if ngood < 0:
-                raise ValueError('ngood < 0')
-        if isinstance(nbad, (int, float)):
-            if nbad < 0:
-                raise ValueError('nbad < 0')
-        if isinstance(nsample, (int, float)):
-            if nsample < 1:
-                raise ValueError('nsample < 1')
-        if (isinstance(ngood, (int, float)) and isinstance(nbad, (int, float))
-                and isinstance(nsample, (int, float))):
-            if ngood + nbad < nsample:
-                raise ValueError('ngood + nbad < nsample')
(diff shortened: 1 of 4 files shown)


jberg5:
cc @seberg @leofang , appreciate your thoughts as always!

seberg:
Can we check if e.g. returning `-1` (or the minimum possible value) as an invalid result has practically no speed impact? Maybe that would be a reasonable solution that doesn't diverge scalars vs. array inputs.

(There are other parts that do these checks, for most returning `NaN` would be the thing to do in that philosophy -- here we need an "invalid" integer, though)

EDIT: But maybe we should just do this also, since it probably covers the vast majority of uses.

jberg5:
@seberg good question, I can't imagine a check like `good < 0 || bad < 0 || sample < 0 || sample > good + bad` would take more than ~1 cycle (I haven't benchmarked this though!). That said I can't find any precedent in cupy for returning a sentinel value. `-1` makes sense to me as a single user, but as for the question of introducing this pattern generally, I'll leave that decision up to you :)

Anyway I'd be on board with having python-side checks for scalars, and a kernel guard -> sentinel value to catch everything else. I can benchmark returning -1 if you'd like, but if you're happy with the reasoning, I can just push the additional check to my branch.

seberg (approved):
@jberg5 if you want to follow up on the -1 return line of thought soon, please do so. But otherwise, this should be OK and is a pattern used in some other parts as well.

/test mini

seberg:
Ah, thanks for triggering mini again, enabled auto-merge, let's just get this in.

leofang:
@seberg should we backport #9805 first?

seberg:
Ah, I forgot that we didn't backport that one for exactly that reason. But yeah, no reason not to backport both.

seberg:
(Or actually, you marked it as draft, for precisely this reason.)

PR #9824, opened 2026-03-23
Add bicgstab solver for sparse linear systems

marco-pas:
added bicgstab in the sparse solvers directory. This feature was missing, and was adapted from scipy's https://github.com/scipy/scipy/blob/main/scipy/sparse/linalg/_isolve/iterative.py 

The solver build on the cgs solver which has some issues in converging. feature needed for some solver benchmarking in nonlinear PDEs using single and double precisions. test passed on cuda device tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py::TestBicgstab

### DIFF
--- a/cupyx/scipy/sparse/linalg/_iterative.py
+++ b/cupyx/scipy/sparse/linalg/_iterative.py
@@ -342,7 +342,7 @@ def bicgstab(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
     rtilde = r.copy()
 
     # Initialize vars to prevent linter warnings
-    rho_prev, omega, alpha, p, v = None, None, None, None, None
+    rho_prev, omega, alpha, p, v = 0.0, 0.0, 0.0, 0.0, 0.0
     s = cupy.empty_like(r)
 
     iters = 0
@@ -353,14 +353,15 @@ def bicgstab(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
         if iters >= maxiter:  # reached max iters
             break
 
-        rho = dotprod(rtilde, r)
+        # pull from device to host
+        rho = dotprod(rtilde, r).item()
 
-        # Breakdown checks ensure CuPy doesn't generate NaNs
-        if cupy.abs(rho) < rhotol:  # rho tol breakdown
+        # Breakdown checks ensure no NaNs
+        if abs(rho) < rhotol:  # rho tol breakdown on host
             return x, -10
 
         if iters > 0:
-            if cupy.abs(omega) < omegatol:  # omega tol breakdown
+            if abs(omega) < omegatol:  # omega tol breakdown
                 return x, -11
 
             beta = (rho / rho_prev) * (alpha / omega)
@@ -372,24 +373,25 @@ def bicgstab(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
 
         phat = psolve(p)
         v = matvec(phat)
-        rv = dotprod(rtilde, v)
+        rv = dotprod(rtilde, v).item()  # pull to host
 
         if rv == 0:
             return x, -11
 
         alpha = rho / rv
         r -= alpha * v
-        s[:] = r[:]
 
         # does a half-step check
         if cupy.linalg.norm(s) <= atol:
             x += alpha * phat
             break
 
-        shat = psolve(s)
+        shat = psolve(r)
         t = matvec(shat)
-        omega = dotprod(t, s) / dotprod(t, t)
 
+        omega = dotprod(t, r).item() / dotprod(t, t).item
+
+        # host scalar times device array
         x += alpha * phat
         x += omega * shat
         r -= omega * t
--- a/cupyx/scipy/sparse/linalg/_iterative.py
+++ b/cupyx/scipy/sparse/linalg/_iterative.py
@@ -343,7 +343,6 @@ def bicgstab(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
 
     # Initialize vars to prevent linter warnings
     rho_prev, omega, alpha, p, v = 0.0, 0.0, 0.0, 0.0, 0.0
-    s = cupy.empty_like(r)
 
     iters = 0
     while True:
@@ -382,14 +381,14 @@ def bicgstab(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
         r -= alpha * v
 
         # does a half-step check
-        if cupy.linalg.norm(s) <= atol:
+        if cupy.linalg.norm(r) <= atol:
             x += alpha * phat
             break
 
         shat = psolve(r)
         t = matvec(shat)
 
-        omega = dotprod(t, r).item() / dotprod(t, t).item
+        omega = dotprod(t, r).item() / dotprod(t, t).item()
 
         # host scalar times device array
         x += alpha * phat
--- a/cupyx/scipy/sparse/linalg/_iterative.py
+++ b/cupyx/scipy/sparse/linalg/_iterative.py
@@ -388,7 +388,7 @@ def bicgstab(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
         shat = psolve(r)
         t = matvec(shat)
 
-        omega = dotprod(t, r).item() / dotprod(t, t).item()
+        omega = (dotprod(t, r) / dotprod(t, t)).item()
 
         # host scalar times device array
         x += alpha * phat
--- a/cupyx/scipy/sparse/linalg/_iterative.py
+++ b/cupyx/scipy/sparse/linalg/_iterative.py
@@ -323,11 +323,11 @@ def bicgstab(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
     n = A.shape[0]
     if n == 0:
         return cupy.empty_like(b), 0
-    
+
     b_norm = cupy.linalg.norm(b)
     if b_norm == 0:
         return b, 0
-        
+
     atol = max(float(atol), rtol * float(b_norm))
     if maxiter is None:
         maxiter = 10 * n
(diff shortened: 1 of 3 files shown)


marco-pas (commented):
added comment

marco-pas:
Hello! I've updated the code and triggered the new build, but the CI seems to be pending since yesterday. Could a maintainer please take a look at it? Thank you all!

leofang:
Hi @marco-pas, thanks for your contribution. Do you actually get any speedup for this pure Python solver implementation over the SciPy counterpart? There are reasons they don't get implemented so far, because it can be very tricky. Could you provide some benchmark data against SciPy?

marco-pas:
@leofang Hello, thanks for the reply! yes I do get speedups compared to the SciPy version. I have been using my own version for about 10 days on a nonlinear solver and it works great. I am send you here some benchmarks for a 2d Poisson system.

bicgstab_speedup_benchmark

### Final Speedup Summary (CuPy vs. SciPy)

| Grid | Unknowns | FP32 Avg | FP32 Std | FP64 Avg | FP64 Std |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **64x64** | 4,096 | 0.09x | 0.03 | 0.11x | 0.00 |
| **128x128** | 16,384 | 0.21x | 0.00 | 0.68x | 0.31 |
| **256x256** | 65,536 | 0.88x | 0.01 | 1.74x | 0.09 |
| **512x512** | 262,144 | **102.66x** | 0.31 | **7.01x** | 0.17 |

Note: This test was repeated on 10 runs on a 5080 RTX, CPU is AMD Ryzen 9 7900X. 

I found that in some problems there can be issues in GMRES convergence when you get strong nonlinearities in PDEs. This usually does not occur as much in BiCGStab, so that is why it would be nice to have it in the repo. Additionally solvers from other libraries (e.g., JAX) are not as optimized, so it is always worth is to use something like DLPack and move back to a CuPy solver if the interest is running on an accelerator.

By the way, I do get similar speed ups on my own solver (which combines different libraries but runs the CuPy solver) as the ones shown in the table. Especially, in FP32 the results are great!

Please, let me know if anything else is needed.

leofang:
Thanks, @marco-pas! Looks promising, though there might be room for improvement. I will think about this (i am currently traveling). 

In the meanwhile could you share your benchmark script so if others on the team have time they can play around with it? 🙂

marco-pas:
@leofang I agree, there is definitely room for improvement!! In the meanwhile here is the benchmark script.

[bicgstabBenchmarkPoisson.py](https://github.com/user-attachments/files/26325806/bicgstabBenchmarkPoisson.py)
[bicgstabCuPy.py](https://github.com/user-attachments/files/26325762/bicgstabCuPy.py)

I've set a ignore warnings for convergence issues, this should not impact the raw performance as the erratic convergence is inherently in BiCGstab. Would be nice to get additional data on other devices and scanning over solver tolerances.

Safe travels! ;)

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,129 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+        phat = psolve(p)
+        v = matvec(phat)
+        rv = dotprod(rtilde, v)
+
+        if rv == 0:
+            return x, -11
+
+        alpha = rho / rv
+        r -= alpha * v
+        s[:] = r[:]
```
eriknw, line 382: Do we need to use `s`? It looks like it's always an identical copy of `r` at the point it's used. `s[:] = r[:]` is a full memcpy kernel that does nothing useful.

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,129 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+        r_norm = cupy.linalg.norm(r)
+        if r_norm <= atol:  # reached abs tol
+            break
+        if iters >= maxiter:  # reached max iters
+            break
+
+        rho = dotprod(rtilde, r)
+
+        # Breakdown checks ensure CuPy doesn't generate NaNs
+        if cupy.abs(rho) < rhotol:  # rho tol breakdown
```
eriknw, line 359: I wonder if scalars such as `rho`, `omega`, `rv`, `alpha`,  etc. would be better as Python scalars. You already have to sync to evaluate breakdown checks like `if cupy.abs(rho) < rhotol`. Instead of keeping `rho` as a GPU scalar and launching a `cupy.abs` kernel, we can extract it with `.item()` right after the dot product.

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,129 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+            return x, -10
+
+        if iters > 0:
+            if cupy.abs(omega) < omegatol:  # omega tol breakdown
+                return x, -11
+
+            beta = (rho / rho_prev) * (alpha / omega)
```
eriknw, line 366: Once values are Python scalars, I think this can be all be done on the CPU.

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,129 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+    dotprod = cupy.vdot if x.dtype.kind == 'c' else cupy.dot
+
+    rhotol = cupy.finfo(x.dtype.char).eps ** 2
+    omegatol = rhotol
+
+    r = b - matvec(x)
+    rtilde = r.copy()
+
+    # Initialize vars to prevent linter warnings
+    rho_prev, omega, alpha, p, v = None, None, None, None, None
```
eriknw, line 345: Super-minor nit: it may be nicer and more accurate type-wise to initialize these to 0.0. I don't really care though, so up to you!

tests/cupyx_tests/scipy_tests/sparse_tests/test_linalg.py
```diff
@@ -466,6 +466,144 @@ def test_invalid(self):
         with pytest.raises(TypeError):
             sp.linalg.cg(ng_a, b, atol=self.atol)
 
+@testing.parameterize(*testing.product({
```
eriknw, line 469: Linting: there should be two blank lines between classes (pep8). Please add one.

eriknw (changes requested):
This looks pretty good overall, thanks @marco-pas! Implementation appears correct. I left a few comments for some tweaks to consider--they _may_ reduce per-iteration overhead (might be nice to measure tho).

I wonder if you have any thoughts or guidance for when the cupy solver is faster/better.

marco-pas (commented):
Hello! Thanks for the comments.

Updated bicgstab solver with cpu scalars, changed initialization to match type and fixed linting issues.

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,131 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+
+        # does a half-step check
+        if cupy.linalg.norm(s) <= atol:
+            x += alpha * phat
+            break
+
+        shat = psolve(r)
+        t = matvec(shat)
+
+        omega = dotprod(t, r).item() / dotprod(t, t).item
```
eriknw: Forgot parentheses:
```suggestion
        omega = dotprod(t, r).item() / dotprod(t, t).item()
```

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,131 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+
+    rhotol = cupy.finfo(x.dtype.char).eps ** 2
+    omegatol = rhotol
+
+    r = b - matvec(x)
+    rtilde = r.copy()
+
+    # Initialize vars to prevent linter warnings
+    rho_prev, omega, alpha, p, v = 0.0, 0.0, 0.0, 0.0, 0.0
+    s = cupy.empty_like(r)
```
eriknw: `s` may be deleted now:
```suggestion
```

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,131 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+        rv = dotprod(rtilde, v).item()  # pull to host
+
+        if rv == 0:
+            return x, -11
+
+        alpha = rho / rv
+        r -= alpha * v
+
+        # does a half-step check
+        if cupy.linalg.norm(s) <= atol:
```
eriknw: `r`, not `s`:
```suggestion
        if cupy.linalg.norm(r) <= atol:
```

eriknw (changes requested):
Thanks @marco-pas! I left a few suggestions.

marco-pas:
bicgstab_speedup_benchmark

Ran a new benchmark test and results are comparable to previous ones

cupyx/scipy/sparse/linalg/_iterative.py
```diff
@@ -286,6 +286,130 @@ def cgs(A, b, x0=None, *, rtol=1e-5, atol=0.0, maxiter=None, M=None,
...
+
+        # does a half-step check
+        if cupy.linalg.norm(r) <= atol:
+            x += alpha * phat
+            break
+
+        shat = psolve(r)
+        t = matvec(shat)
+
+        omega = dotprod(t, r).item() / dotprod(t, t).item()
```
eriknw: Aha, one more tweak to do one sync instead of two:
```suggestion
        omega = (dotprod(t, r) / dotprod(t, t)).item()
```

eriknw (approved):
LGTM! I left another minor suggestion (one less sync), but this shouldn't hold things up.

leofang (approved):

leofang:
Thanks a lot @marco-pas for your contribution and @eriknw for helping push this across the finish line!

PR #9825, opened 2026-03-23
Add int64 index support to `cupyx.scipy.sparse`

### LINKED ISSUE #3513: Support sparse matrices with more than 2^31-1 elements

eriknw:
Closes #3513.

Sparse matrices (CSR, CSC, COO) now support int64 `indices`/`indptr`/`row`/`col`, matching `scipy.sparse` semantics. Index dtype is chosen automatically via `get_index_dtype(check_contents=True)` — int32 when values fit, int64 when they don't. All existing int32 behavior is preserved.

Matrices with dimensions or nnz exceeding ~2.1 billion (INT32_MAX) previously failed silently or raised unhelpful errors. This is a prerequisite for large-scale sparse workloads (e.g., single-cell genomics via rapids-singlecell, large graph analytics). There is already a companion PR that uses this branch: https://github.com/scverse/rapids-singlecell/pull/620

Note that this PR was made with assistance of Claude Code (models: Sonnet 4.6, Opus 4.6). I have been _heavily_ involved every step of the way, which I hope is clear from the level of hardening already present. Nevertheless, as is typical with AI-assisted work, I am part author and part reviewer, and I am still reviewing these changes.

### How

- **cuSPARSE Generic API** (spmv, spmm, spGEMM) already supports int64 via `SpMatDescr` — no changes needed for these paths
- **Pure-CuPy fallbacks** for Legacy API functions that are int32-only (csr2cscEx2, xcoo2csr, xcsrsort, csrgeam2, etc.) using searchsorted, lexsort, and cumsum
- **CUDA kernel templating** — all inline kernels that used `int*` or `int32` parameters now use template parameter `I`
- **`_from_parts()` classmethods** on CSR/CSC/COO for internal construction without `check_contents` downcast, replacing ad-hoc `object.__new__` patterns
- **Shared helpers** (`_indptr_to_coo`, `_build_indptr`) consolidate repeated patterns

Also fixed along the way: float32 precision bug in `__setitem__` for nnz > 16M, several int32 overflow issues in binopt/setdiag/DIA kernels that affected large matrices regardless of index dtype, and `random()`/`rand()` O(m*n) memory usage for large shapes (noted by @samaid in #3513).

### What remains

- Request updates to the cuSPARSE generic API so that we would need fewer pure-CuPy fallback implementations
- ~Run everything and experiment with a larger, newer GPU. Current development has been done on Tesla V100-SXM2-32GB with CUDA 12.8, but since the scope is also to support large sparse arrays that need more than 32 GB, some things need a beefier GPU~ Done
- ~Verify SpGEMM native int64 path on CUDA 13.0+~ Done
- Activate SpGEAM path when it ships in a public CUDA release. I would also like to test this locally
- `dense2csr` bool-dtype fallback still int32 (guarded with ValueError, medium-effort fix)
- `cupy.repeat` upstream fixes would simplify several workarounds (but the workaround are fast, and not too bad). I plan to create a PR that support `cupy.repeat` with cupy arrays as `repeats` argument (I am nearly finished with this) (**In progress:** https://github.com/cupy/cupy/pull/9828)

Future work is to add support for sparse _arrays_ (not matrix classes). `scipy.sparse` semantics for sparse arrays is to s
imply use the dtype of the indices that the user provided (and then if everything isn't int32, then upcast to int64 as appropriate). Supporting mixed dtypes for `indices`/`indptr`/`row`/`col` for sparse matrix classes is something we could consider, but I would prefer to do that in another PR. Also note that sparse matrix classes are being deprecated in scipy and may be removed in 2027.

This is a large PR. It's also my first major contribution to CuPy. I can break this into multiple PRs if it would help. I value feedback.

Here is a review by Claude Code that explains some of the implementation details: [line_by_line_full.md](https://github.com/user-attachments/files/26196419/line_by_line_full.md)


eriknw:
This work began with https://github.com/samaid/cupy/commit/3014b99c6495f8bad490d24499ec3b794b9fd7eb from @samaid, which is the "spiritual" predecessor to this PR and helped inform things.

cupy_backends/cuda/libs/cusparse.pxd
```diff
@@ -69,6 +69,21 @@ cdef extern from *:
...
+    #ifndef CUSPARSE_SPGEAM_ALG_DEFAULT
+    /* SpGEAM not in this cuSPARSE; provide opaque forward decls. */
+    typedef void* cusparseSpGEAMDescr_t;
+    typedef int cusparseSpGEAMAlg_t;
+    #define CUSPARSE_SPGEAM_ALG_DEFAULT 0
+    #define CUSPARSE_SPGEAM_ALG1 1
+    #endif
+    """
+    ctypedef void* SpGEAMDescr 'cusparseSpGEAMDescr_t'
+    ctypedef int SpGEAMAlg 'cusparseSpGEAMAlg_t'
```
eriknw: I have verified the code paths to use SpGEAM works against the dev version of cuSPARSE 🎉

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -519,6 +519,10 @@ def spsolve(A, b):
         warnings.warn('CSR format is required. Converting to CSR format.',
                       sparse.SparseEfficiencyWarning)
         A = A.tocsr()
+
+    from cupyx.cusparse import _check_int32_indices
+    _check_int32_indices(A, 'spsolve')
```
seberg:
```suggestion
    cupyx.cusparse._check_int32_indices(A, 'spsolve')
```

cupyx/scipy/sparse/_construct.py
```diff
@@ -301,7 +317,14 @@ def bmat(blocks, format=None, dtype=None):
     shape = (row_offsets[-1], col_offsets[-1])
 
     data = cupy.empty(nnz, dtype=dtype)
-    idx_dtype = _sputils.get_index_dtype(maxval=max(shape))
+    # Propagate int64 from input blocks: if any input block had int64
+    # indices (checked before COO conversion), the output should too.
```
seberg, line 327: Seems not a big deal, but maybe you know: Is this logic even needed?  SciPy seems to have it (after coo conversion) but even there I don't see that it isn't discarded anyway.

cupyx/scipy/sparse/_construct.py
```diff
@@ -366,14 +389,64 @@ def random(m, n, density=0.01, format='coo', dtype=None,
...
+    _use_rejection = _choice_bytes > (1 << 28) and k < mn // 10
+    if _use_rejection:
+        ind = _rejection_sample_indices(mn, k, random_state, tp)
+    else:
+        try:
+            ind = random_state.choice(mn, size=k, replace=False)
+            ind = ind.astype(tp, copy=False)
+        except cupy.cuda.memory.OutOfMemoryError:
+            cupy.get_default_memory_pool().free_all_blocks()
+            ind = _rejection_sample_indices(mn, k, random_state, tp)
```
seberg: I guess this comes from the initial PR?  CuPy v14 improved the `random.choice` code to avoid this (and return `int64`).

So I expect this change is now unnecessary.

cupyx/scipy/sparse/_construct.py
```diff
@@ -366,14 +389,64 @@ def random(m, n, density=0.01, format='coo', dtype=None,
...
+        flat_new = (random_state.rand(sample_size) * mn).astype(tp)
+        flat = cupy.unique(cupy.concatenate([flat, flat_new]))
+
+        if len(flat) >= k:
+            flat = flat[:k]
+            remaining = 0
+        else:
+            remaining = k - len(flat)
+
+    return flat
```
seberg: (Also this helper should be unnecessary now)

cupyx/scipy/sparse/_construct.py
```diff
@@ -301,7 +317,14 @@ def bmat(blocks, format=None, dtype=None):
     shape = (row_offsets[-1], col_offsets[-1])
 
     data = cupy.empty(nnz, dtype=dtype)
-    idx_dtype = _sputils.get_index_dtype(maxval=max(shape))
+    # Propagate int64 from input blocks: if any input block had int64
+    # indices (checked before COO conversion), the output should too.
+    # Also check the flat-index product as a maxval heuristic.
+    _maxval = int(shape[0]) * int(shape[1]) - 1
```
seberg: I am a bit confused by this `_maxval`?  We are (initially) going to COO, shouldn't it just be `max(shape[0], shape[1])`?

(The `-1` here makes sense, and seems missing in some places below. I suspect SciPy does the same thing though, and off by one in this direction is safe anyway.)

cupyx/scipy/sparse/_csr.py
```diff
@@ -101,9 +101,59 @@ def _comparison(self, other, op, op_name):
...
+                'Comparing a sparse matrix with {} using {} is '
+                'inefficient. Try using {} instead.'
+                .format(scalar, sym, alt),
+                _base.SparseEfficiencyWarning)
+            idx_dtype = self.indices.dtype
+            other = csr_matrix._from_parts(
+                data,
+                cupy.zeros((1,), dtype=idx_dtype),
+                cupy.arange(2, dtype=idx_dtype),
+                (1, 1))
```
seberg, line 161: Seems like a nice improvement!

cupyx/scipy/sparse/_csr.py
```diff
@@ -290,8 +366,14 @@ def _maximum_minimum(self, other, cupy_op, op_name, dense_check):
...
-                return csr_matrix((new_data, self.indices, self.indptr),
-                                  shape=self.shape, dtype=dtype)
+                new_data = new_data.astype(dtype, copy=False)
+                return csr_matrix._from_parts(
+                    new_data, self.indices, self.indptr,
+                    self.shape,
+                    has_canonical_format=getattr(
+                        self, '_has_canonical_format', None),
+                    has_sorted_indices=getattr(
+                        self, '_has_sorted_indices', None))
```
seberg, line 392: I guess to preserve the input index type?

cupyx/scipy/sparse/_csr.py
```diff
@@ -512,10 +597,18 @@ def transpose(self, axes=None, copy=False):
...
+            indices = self.indices.copy()
+            indptr = self.indptr.copy()
+        else:
+            data, indices, indptr = self.data, self.indices, self.indptr
+        return _csc.csc_matrix._from_parts(
+            data, indices, indptr, shape,
+            has_canonical_format=getattr(
+                self, '_has_canonical_format', None),
+            has_sorted_indices=getattr(
+                self, '_has_sorted_indices', None))
```
seberg, line 627: Isn't this just `_with_data`, which seems like a nice helper?  Although, I think that isn't doing the `_has_sorted_indices`, yet, but that can be added and is clearer there.)

I have a feeling that there are quite a few of these where `_with_data` will work nicely.

cupyx/scipy/sparse/_csr.py
```diff
@@ -1149,45 +1273,65 @@ def dense2csr(a):
...
-    cupy_dense2csr_step1()(m, n, a, indptr, info)
-    indptr = cupy.cumsum(indptr, dtype=numpy.int32)
-    info = cupy.cumsum(info, dtype=numpy.int32)
-    nnz = int(indptr[-1])
-    indices = cupy.empty(nnz, dtype=numpy.int32)
+    indptr = cupy.zeros(m + 1, dtype=idx_dtype)
+    info = cupy.zeros(mn + 1, dtype=idx_dtype)
+    it = numpy.dtype(idx_dtype).type
+    cupy_dense2csr_step1()(it(m), it(n), a, indptr, info)
+    from cupyx.cusparse import _cumsum_int64
```
seberg: I guess the cub fix should make this unnecessary (I need to check and finalize that!).

cupyx/scipy/sparse/_csr.py
```diff
@@ -1149,45 +1273,65 @@ def dense2csr(a):
...
+_DENSE2CSR_ATOMICADD_PREAMBLE = '''
+template <typename I>
+__device__ inline void _d2c_atomic_add(I* addr) {
+    atomicAdd(addr, (I)1);
+}
+template <>
+__device__ inline void _d2c_atomic_add<long long>(long long* addr) {
+    atomicAdd((unsigned long long int*)addr,
+              (unsigned long long int)1);
+}
```
seberg: I guess OK to copy, but maybe name both `_atomic_add_one`?  Or maybe don't bother...

(ideally/unless you can use the same `atomic_add_one` preamble for both)

cupyx/scipy/sparse/_data.py
```diff
@@ -162,20 +170,24 @@ def _min_or_max_axis(self, axis, min_or_max, explicit):
 
         # Do the reduction
         value = mat._minor_reduce(min_or_max, axis, explicit)
-        major_index = cupy.arange(M)
+        idx_dtype = _sputils.get_index_dtype(maxval=max(self.shape))
```
seberg: Isn't this `maxval=M`?  But dunno, maybe there is some need to preserve the original type somewhere?

cupyx/cusparse.py
```diff
@@ -56,6 +56,96 @@ def _cast_common_type(*xs):
...
 
+def _check_int32_indices(a, func_name):
+    """Raise ValueError if *a* has int64 indices."""
+    if a.indices.dtype == _cupy.int64:
+        raise ValueError(
+            '{} does not support int64 indices '
+            '(cuSPARSE {} is int32-only). '
+            'Cast to int32 if values fit: '
+            'a.indices.astype(cupy.int32)'.format(
+                func_name, func_name))
```
seberg, line 65: This instruction seems incorrect, is there a better one?  (It seems very hard to even achieve an array that could be int32 but isn't?)

cupyx/cusparse.py
```diff
@@ -56,6 +56,96 @@ def _cast_common_type(*xs):
...
+        _cupy.add.at(
+            indptr[1:].view(_cupy.uint64),
+            row_indices, _cupy.uint64(1))
+    else:
+        _cupy.add.at(indptr[1:], row_indices, 1)
+    _cumsum_int64(indptr)
+    return indptr
+
+
+def _with_indices_dtype(m, dtype):
```
seberg, line 113: Maybe this could just live next to `_with_values`?

seberg (commented):
I had a first pass, a lot of nice cleanups, too.  I should check the tests more carefully for coverage (I am wondering if some of them should just be marked globally as `slow` even if they aren't very slow, in case they use a huge amount of memory.)

cupyx/scipy/sparse/linalg/_solve.py
eriknw: Done. It was easier for me to do it, sorry.

cupyx/scipy/sparse/_construct.py
eriknw: Yup. Verified--`RandomState.choice(10**12, 1000)` now works and returns int64. Removed the early-switch heuristic and OOM fallback. The rejection sampling helper is also removed. Good catch--thanks!

cupyx/scipy/sparse/_construct.py
eriknw: Removed.

cupyx/scipy/sparse/_construct.py
eriknw: Agreed. Oops. COO indices represent row/col positions, so maxval should be `max(shape) - 1`, not the flat product. Changed. Updated tests that depended on the old product-based heuristic.

cupyx/scipy/sparse/_csr.py
eriknw: Yes--the public constructor calls `check_contents=True` which downcasts int64 to int32 when values are small. `_from_parts` preserves the dtype.

cupyx/scipy/sparse/_csr.py
eriknw: Close--`_with_data` replaces data but preserves indices/indptr. The transpose returns a CSC matrix that shares the same physical arrays, reinterpreted. `_with_data` doesn't change the matrix class (CSR→CSC). We could add a transpose-specific helper but the current pattern is clear and explicit.

cupyx/scipy/sparse/_csr.py
eriknw: Yes--PR #9810 widens the CUB scan wrapper from `int` to `long long`, fixing `cupy.cumsum` for arrays > 2^32 elements.  Once merged, `_cumsum_int64` can be removed and all call sites reverted to plain `cupy.cumsum`.

cupyx/scipy/sparse/_csr.py
eriknw: Removed entirely! I just enabled `cupy.add.at` for int64 targets by relaxing the type guard in `_routines_indexing.pyx` (CuPy's `atomics.cuh` already provided `atomicAdd(long long*)` via reinterpret_cast). With that fix, the `_atomic_add_one` template and its `long long` specialization are no longer needed--the CUDA kernel code now uses `atomicAdd(&addr, (I)1)` directly, which resolves to the `atomics.cuh` wrapper when `I=long long`. I'm pretty sure this is safe to do 🤞 

Also removed the `view(uint64)` workaround in `_build_indptr` (cusparse.py) and `_insert_many` (_compressed.py)--both now call `cupy.add.at` directly on int64 arrays.

cupyx/scipy/sparse/_data.py
eriknw: Ha, yes--`M` is the major axis dimension which is the correct maxval for the output indices. Changed from `max(self.shape)` to `M`.

cupyx/cusparse.py
eriknw: Removed the misleading "Cast to int32 if values fit" suggestion. The error now just states the limitation without suggesting a potentially unsafe workaround.

cupyx/cusparse.py
eriknw: `_with_indices_dtype` is in `cusparse.py` because it's used by the cuSPARSE dispatch code (SpGEMM mixed-type upcast).  `_with_data` is on the sparse matrix classes. Moving it would create a circular import. Left as-is. Happy to revisit if you have a different preference.

eriknw:
> ... more carefully for coverage ...

This little comment caused me to go on a decent-sized tangent to look for coverage gaps. I added a few more tests to improve important coverage. I think I should probably consolidate the tests added in this PR so they're not quite so verbose.

cupyx/scipy/sparse/linalg/_solve.py
```diff
@@ -519,6 +519,10 @@ def spsolve(A, b):
         warnings.warn('CSR format is required. Converting to CSR format.',
                       sparse.SparseEfficiencyWarning)
         A = A.tocsr()
+
+    import cupyx.cusparse
```
leofang: Q: Why do we defer the import here?

cupyx/cusparse.py
```diff
@@ -56,6 +56,79 @@ def _cast_common_type(*xs):
...
+    carry = arr.dtype.type(0)
+    for start in range(0, n, _CHUNK):
+        end = min(start + _CHUNK, n)
+        chunk = arr[start:end]
+        _cupy.cumsum(chunk, out=chunk)
+        if start > 0:
+            chunk += carry
+        carry = chunk[-1].item()  # synchronize!
+        carry = arr.dtype.type(carry)
+    return arr
```
leofang: TODO: Merge https://github.com/cupy/cupy/pull/9810 first.

cupyx/cusparse.py
```diff
@@ -56,6 +56,79 @@ def _cast_common_type(*xs):
...
+
+
+def _build_indptr(row_indices, n_rows, dtype):
```
leofang: Can we make this clearer? I suppose this only works (and is only needed) for dtype=int64? How about calling it `_build_indptr_int64` for clarity?

cupy_backends/cuda/libs/cusparse.pxd
```diff
@@ -69,6 +69,21 @@ cdef extern from *:
     # CSR2CSC
     ctypedef int Csr2CscAlg 'cusparseCsr2CscAlg_t'
 
+# TODO(cuSPARSE): remove stubs when SpGEAM ships in a public release.
```
leofang: Note: For the TODO comments, the convention is `TODO(id)`, in which case it should be `TODO(eriknw)` 🙂

leofang: Let's do a sweep and update all TODO comments before merging.

cupyx/cusparse.py
leofang: Update: #9867 (instead of #9810) was merged. @eriknw this helper can be removed now.

cupyx/scipy/sparse/_compressed.py
```diff
@@ -675,12 +863,7 @@ def _minor_slice(self, idx, copy=False):
         new_shape = self._swap(M, N)
 
         if N == 0 or self.nnz == 0:
-            return self.__class__(new_shape, dtype=self.dtype)
-        if step == 1:
-            return self.__class__(
-                _index._get_csr_submatrix_minor_axis(
-                    self.data, self.indices, self.indptr, start, stop),
-                shape=new_shape, copy=False)
```
seberg, line 683: Hmmm, claude is asking where the `step == 1` fast path went, and it isn't clear to me immediately either.
(Maybe removed when cumsum was broken?)

cupyx/scipy/sparse/_compressed.py
```diff
@@ -297,17 +327,72 @@ def __init__(self, arg1, shape=None, dtype=None, copy=False):
...
+        """
+        if indices.dtype != indptr.dtype:
+            raise ValueError(
+                'indices and indptr must have the same dtype, '
+                'got {} and {}'.format(indices.dtype, indptr.dtype))
+        A = cls.__new__(cls)
+        sparse_data._data_matrix.__init__(A, data)
+        A.indices = indices
+        A.indptr = indptr
+        from cupyx.cusparse import MatDescriptor
```
seberg, line 382: I feel Leo might have made a similar comment somewhere.  We can defer also, the nicest thing might be to just add `cusparse` to the `scipyx.__getattr__` (which we already have).

Then we can just write `cupyx.cusparse.MatDescriptor` which is lazy, feels cleaner, and is even faster.

(but yeah, another semi-related improvement, so happy to follow up)

cupyx/scipy/sparse/_compressed.py
```diff
@@ -558,65 +663,89 @@ def _major_index_fancy(self, idx):
...
+        # For very large minor axis (N > INT32_MAX), the O(N)
+        # col_counts allocation would be prohibitive and the int32
+        # col_counts/col_order arrays can't represent the positions.
+        # Fall back to the sort-based O(nnz) path.
+        if N > numpy.iinfo(numpy.int32).max:
+            return self._minor_index_fancy_sorted(
+                idx, M, n_idx, new_shape)
+
+        idx_dtype = self.indices.dtype
+        idx_tname = self._idx_type_name(idx_dtype)
```
seberg: Since you are using it elsewhere also, maybe just use `_scalar.get_typename()`?

(If it wasn't used elsewhere, it wouldn't matter, I think... since this is the typename used by the RawModule `name_expressions` and not the name in an `ElementwiseKernel`.)

cupyx/scipy/sparse/_compressed.py
```diff
@@ -633,12 +762,60 @@ def _minor_index_fancy(self, idx):
...
+        total_nnz = int(cnt.sum())  # synchronize!
+        if total_nnz == 0:
+            return self._empty_like(new_shape)
+
+        out_src = cupy.repeat(
+            cupy.arange(self.nnz, dtype=cupy.int64), cnt)
+        # offset: position within each repeated group (grouped arange)
+        cum_cnt = cupy.zeros(self.nnz + 1, dtype=cupy.int64)
+        cupy.cumsum(cnt, out=cum_cnt[1:])
+        offset = cupy.arange(total_nnz, dtype=cupy.int64) \
```
seberg: Isn't it nicer to write that as `arange(-cum_cnt[out_src], -cum_cnt[out_src] + total_nnz, dtype=cupy.int64)`? (Anyway, silly nit, I mostly saw the `\` ;))

cupyx/scipy/sparse/_compressed.py
```diff
@@ -633,12 +762,60 @@ def _minor_index_fancy(self, idx):
...
+        sort_order = cupy.argsort(idx)
+        sorted_idx = idx[sort_order]
+
+        lo = cupy.searchsorted(sorted_idx, self.indices, side='left')
+        hi = cupy.searchsorted(sorted_idx, self.indices, side='right')
+        cnt = (hi - lo).astype(cupy.int64)
+
+        total_nnz = int(cnt.sum())  # synchronize!
+        if total_nnz == 0:
+            return self._empty_like(new_shape)
```
seberg, line 810: NIT: isn't this just the same as `prod(new_shape) == 0` (and that is already checked earlier)?

(Basically, the `cnt.sum()` seems unnecessary to me either way, as the `repeat()` already does an implicit sum as `len(out_src)`.

cupyx/scipy/sparse/_compressed.py
```diff
@@ -742,17 +925,20 @@ def _set_many(self, i, j, x):
         i, j, M, N = self._prepare_indices(i, j)
         x = cupy.array(x, dtype=self.dtype, copy=True, ndmin=1).ravel()
 
-        new_sp = cupyx.scipy.sparse.csr_matrix(
-            (cupy.arange(self.nnz, dtype=cupy.float32),
-             self.indices, self.indptr), shape=(M, N))
+        # Temporary CSR mapping each stored element to its flat offset.
+        # Use _from_parts to avoid check_contents D2H syncs; the
+        # indices/indptr are already validated (they come from self).
```
seberg, line 966: Should this also pass `has_canonical_format=getattr(self, "_has_canonical_format")` then?

cupyx/scipy/sparse/_compressed.py
```diff
@@ -1012,20 +1200,25 @@ def _minor_reduce(self, ufunc, axis, nonzero):
...
-                      self.indptr[1:], cupy.int64(self.shape[axis]),
-                      out))
+        mod, fname = {
+            (cupy.amax, False): (self._max_reduction_mod, 'max_reduction'),
+            (cupy.amin, False): (self._min_reduction_mod, 'min_reduction'),
+            (cupy.amax, True): (self._max_nonzero_reduction_mod,
+                                'max_nonzero_reduction'),
+            (cupy.amin, True): (self._min_nonzero_reduction_mod,
+                                'min_nonzero_reduction'),
+        }[(ufunc, nonzero)]
```
seberg, line 1257: Super nit, but feels a bit strange to me to build a dict (but also was old code either way, the new path seems really cleaner even if a nested if seems almost as good).

cupyx/scipy/sparse/_construct.py
```diff
@@ -262,6 +266,14 @@ def bmat(blocks, format=None, dtype=None):
     brow_lengths = numpy.zeros(M+1, dtype=numpy.int64)
     bcol_lengths = numpy.zeros(N+1, dtype=numpy.int64)
 
+    # Check if any input block has int64 indices before conversion
+    # to COO (the COO constructor may downcast via check_contents).
+    _has_int64 = any(
+        getattr(b, 'indices', getattr(b, 'row', None)) is not None
+        and getattr(b, 'indices', getattr(b, 'row', None)).dtype
+        == cupy.int64
```
seberg: IIRC, my claude asked about DIA inputs here (i.e. it having `offsets` and not `indices` or `row`), but I am not quite convinced this is a big issue.

(I suppose this was the main reason for the question of whether there should be a property or so to figure this out without trying all the different formats here. But if this is the only real place where we need to do this crazy dance, I am not sure it is worthwhile; especially with a future where indptr and indices might have different types...)

cupyx/scipy/sparse/_construct.py
```diff
@@ -366,7 +388,12 @@ def random(m, n, density=0.01, format='coo', dtype=None,
     if data_rvs is None:
         data_rvs = random_state.rand
 
+    mn = m * n
+
+    tp = numpy.int64 if mn > numpy.iinfo(numpy.int32).max \
```
seberg, line 399: Use `_sputils.get_index_dtype`?  But maybe not/doesn't matter...

cupyx/scipy/sparse/_csr.py
```diff
@@ -101,9 +101,59 @@ def _comparison(self, other, op, op_name):
...
+            if not zero_cmp:
+                self.sum_duplicates()
+                new_data = op(self.data, scalar)
+                # Keep only True entries (filter explicit False).
+                mask = new_data
+                if mask.all():  # synchronize!
+                    return csr_matrix._from_parts(
+                        new_data, self.indices.copy(),
+                        self.indptr.copy(), self.shape,
+                        has_sorted_indices=True)
```
seberg: You called `sum_duplicates` so it's `has_canonical_format=True` (which implies sorted indices)?

(same below)

cupyx/scipy/sparse/_csr.py
```diff
@@ -155,7 +205,11 @@ def __mul__(self, other):
         elif isspmatrix_csr(other):
             self.sum_duplicates()
             other.sum_duplicates()
-            if cusparse.check_availability('spgemm'):
+            # int64: always route through cusparse.spgemm, which has a
+            # pure-CuPy fallback when cuSPARSE lacks int64 SpGEMM.
+            is_int64 = (self.indices.dtype == cupy.int64
+                        or other.indices.dtype == cupy.int64)
```
seberg, line 216: Actually, if we add an `_is_int64` property (or `__init__` time attribute) to all of the classes, than also all of these types of checks get nicer (not just block)?

cupyx/scipy/sparse/_csr.py
```diff
@@ -619,15 +726,21 @@ def multiply_by_dense(sp, dn):
             indptr *= n
 
     # out = sp * dn
-    cupy_multiply_by_dense()(sp.data, sp.indptr, sp.indices, sp_m, sp_n,
-                             dn, dn_m, dn_n, indptr, m, n, data, indices)
+    idx_dtype = sp.indptr.dtype
+    it = idx_dtype.type
```
seberg, line 735: N.B.: I _suspect_ you don't actually need `it()` with NEP 50 rules now, but not sure it's worth trying.

cupyx/scipy/sparse/_csr.py
```diff
@@ -870,11 +986,16 @@ def binopt_csr(a, b, op_name):
     a_nnz = a.nnz * (m // a_m) * (n // a_n)
     b_nnz = b.nnz * (m // b_m) * (n // b_n)
 
-    a_info = cupy.zeros(a_nnz + 1, dtype=a.indices.dtype)
-    b_info = cupy.zeros(b_nnz + 1, dtype=b.indices.dtype)
+    idx_dtype = numpy.result_type(a.indices.dtype, b.indices.dtype)
```
seberg: (Could also use `promote_types` here, since it's both simple dtypes and it's used below. -- just a bit faster)

cupyx/scipy/sparse/_csr.py
```diff
@@ -870,11 +986,16 @@ def binopt_csr(a, b, op_name):
...
+    a_indptr = a.indptr.astype(idx_dtype, copy=False)
+    a_indices = a.indices.astype(idx_dtype, copy=False)
+    b_indptr = b.indptr.astype(idx_dtype, copy=False)
+    b_indices = b.indices.astype(idx_dtype, copy=False)
+    a_info = cupy.zeros(a_nnz + 1, dtype=idx_dtype)
+    b_info = cupy.zeros(b_nnz + 1, dtype=idx_dtype)
```
seberg, line 1008: Hmmm, this feels a bit too bad, but I don't have a good idea.  We could type it differently in the `ElementwiseKernel` but not sure it's worthwhile.
(In that case, I guess we might have to add casts to the output type for promotion to it.)

But... I suppose that mixing isn't going to be super common, so unless you are curious if ends up nice and short, I doubt it's worth it.
(Ufuncs have machinery to add casts automatically, but as these use `raw`, I don' thitnk there is any gain.)

seberg (commented):
Submitting slightly early, I don't think there is much in there, mostly silly comments (no need to action).
I am still wondering now if a property isn't best (can be private), but that may be it.

We may want re-consider the exact kernel setup (when is casting done), but I think that also can wait.

I didn't add comments ont he tests, they feel verbose, but seem fine just scrolling through.
(Nor did I paste more claude comments, since you are asking it to review already anyway.)

cupyx/scipy/sparse/_compressed.py
eriknw: Good catch, you and Claude! It was removed back before #9867 was merged, and wasn't added back. Restoring the fast path (it's correct either way).

cupyx/scipy/sparse/_compressed.py
eriknw: Yeah, I agree it's cleaner, but I'd feel more comfortable doing this in a follow-up (there _will_ be followups).

cupyx/scipy/sparse/_compressed.py
eriknw: Good catch. Done.

cupyx/scipy/sparse/_compressed.py
eriknw: I cleaned up the `\`.

Using 0-d cupy arrays actually doesn't work well in `arange`:
```python
In [1]: import cupy as cp

In [2]: a = cp.arange(5)

In [3]: cp.arange(a[0], a[-1])
Cell In[3], line 1
----> 1 cp.arange(a[0], a[-1])

File /raid/erwelch/projects/cupy13/cupy/_creation/ranges.py:63, in arange(start, stop, step, dtype)
     61 ret = cupy.empty((size,), dtype=dtype)
     62 typ = numpy.dtype(dtype).type
---> 63 _arange_ufunc(typ(start), typ(step), ret, dtype=dtype)
     64 return ret

File /raid/erwelch/projects/cupy13/cupy/_core/core.pyx:1757, in cupy._core.core._ndarray_base.__array__()
   1755 # configure flag that allows implicit conversions to NumPy array.
   1756 # (See https://github.com/cupy/cupy/issues/589 for the detail.)
-> 1757 raise TypeError(
   1758     'Implicit conversion to a NumPy array is not allowed. '
   1759     'Please use `.get()` to construct a NumPy array explicitly.')

```
Also, I think may way reads better (no negatives).

cupyx/scipy/sparse/_compressed.py
eriknw: Right--`cupy.repeat(arr, cnt)` already syncs internally to size its output, so reading `out_src.size` afterwards gives `cnt.sum()` for free.  Replaced

```python
total_nnz = int(cnt.sum())  # synchronize!
```

with

```python
out_src = cupy.repeat(...)
total_nnz = out_src.size
```

cupyx/scipy/sparse/_compressed.py
eriknw: Yes for the gather-rows path (`_csr_row_index` preserves each row's column ordering).  Applied via `_from_parts` kwargs.

Not for the histogram path of `_minor_index_fancy` — the output's canonicity depends on whether `idx` is sorted-no-dup, which the helper doesn't validate.  Conservative leave-unset is correct there (a future `has_canonical_format` access lazy-computes).

cupyx/scipy/sparse/_compressed.py
eriknw: Agreed it's a wash. Old code; not changing this late in this PR. Happy to flatten in a follow-up cleanup

cupyx/scipy/sparse/_construct.py
eriknw: This is the right shape for a follow-up refactor: it touches every sparse format and benefits from a consistent treatment of mixed indptr / indices / offsets dtypes (which this PR explicitly doesn't tackle).

cupyx/scipy/sparse/_construct.py
eriknw: Perfectly reasonable suggestion. Once again, this would be a pure refactor on pre-existing code; happy to fold into a follow-up.

cupyx/scipy/sparse/_csr.py
eriknw: Yes--both branches preserve canonical structure (filtering by a per-stored-entry mask doesn't reorder).  Updated both `_from_parts` calls in `_comparison` to pass `has_canonical_format=True` instead of `has_sorted_indices=True` (the setter propagates canonical → sorted).

cupyx/scipy/sparse/_csr.py
eriknw: Sounds good. Filing as a stand-alone refactor PR--adding it here would touch every sparse class plus 8+ call-site updates and obscure the int64 work. Pairs naturally with the DIA-offsets simplification above.

cupyx/scipy/sparse/_csr.py
eriknw: Probably right, but I'd want to verify with the kernel-launch path on a few CUDA versions before dropping the explicit cast. The kernel template-instantiation cache keys on the exact dtype passed, so changing it right now seems like an unnecessary risk. My vote: out of scope for this PR.

cupyx/scipy/sparse/_csr.py
eriknw: Done.  Both ``binopt_csr`` and ``multiply_by_csr``.

cupyx/scipy/sparse/_csr.py
eriknw: Hmm, agreed, I think. Casting at the kernel boundary is the simplest correct approach for now. Templating the dtype on the ElementwiseKernel would require parallel rewrites of the step1/step2 kernels. Defer to a future kernel-templating cleanup.

kmaehashi (approved):
Thank you for the hard work, looks good!

seberg:
OK, I also did another brief look through and I think the best course is to get things in and do follow-ups (if needed).
Thanks @eriknw big PR, but also an important new feature and a lot of iterations/fixups (and unrelated bug fixes for that matter)!

One thing, I am not quite sure but I think we shouldn't hoist the `cusparse` import up to avoid loading `cusparse` unless we need it?
But, I suspect the nicest way to fix that is to use a module level `__getattr__` (either for `cupyx.cusparse` or maybe even in `cusparse.py` itself for all of the `cupy_backends` imports)

(EDIT: Although, maybe it is fine, since at this point we imported `cupyx.sparse`)

jcrist:
Thanks y'all!!!! We're very excited to get to use this in cuml. Excellent work!

cupyx/scipy/sparse/_dia.py
```diff
@@ -55,7 +56,8 @@ def __init__(self, arg1, shape=None, dtype=None, copy=False):
 
         data = cupy.array(data, dtype=dtype, copy=copy)
         data = cupy.atleast_2d(data)
-        offsets = cupy.array(offsets, dtype='i', copy=copy)
+        off_dtype = _sputils.get_index_dtype(maxval=max(shape))
+        offsets = cupy.array(offsets, dtype=off_dtype)
```
leofang, line 60: The nvmath-python team pointed out to me offline that `copy=copy` is dropped and they caught this in their tests (nvbugs 6255938).

PR #9828, opened 2026-03-25
Allow cupy.ndarray as repeats argument to cupy.repeat

### LINKED ISSUE #3849: Support array objects as the repeats argument to cupy.repeat
For consistency with NumPy, it would be nice to allow passing a CuPy array (or perhaps an ndarray-like object) as the `repeats=` argument to `cupy.repeat`.

I can do the following with NumPy:
```python
import numpy as np
np.repeat(a=np.arange(3), repeats=np.array([2,2,2]))
array([0, 0, 1, 1, 2, 2])
```

But the analogous action is not implemented for CuPy:
https://github.com/cupy/cupy/blob/3c4f481c2eca55045119bc2b96b3335deea2e6da/cupy/core/_routines_manipulation.pyx#L466-L468

eriknw:
cupy.repeat now accepts a CuPy ndarray for the repeats parameter, matching NumPy's API. Scalar-like inputs (size-1 or 0-D arrays) use an efficient broadcast-copy kernel. Multi-element inputs use cumsum + searchsorted + take on the GPU. List/tuple inputs with len > 1, previously O(n * kernel_launch), now use the same GPU path.

Closes #3849. There are probably good reasons why this addition hasn't been done before (there _is_ a D2H transfer) given how long #3849 has been open for and how much interest there has been in it, so I'm curious to find out why! This feature has actually been on my wishlist for a few years too.

Claude Code (Opus 4.6) was used to assist in the development of this PR, but I was also heavily involved. These changes were actually spawned from my efforts to improve sparse matrix classes in #9825, and like in that PR I'm still doing my final human review of these changes.


cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +503,128 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+      ``reps`` array to validate non-negative values and determine output
+      size on the CPU; then ``cumsum + searchsorted + take`` entirely on
+      the device.
+
+    Both paths perform O(total) work on the GPU.  ``searchsorted`` is the
+    best available approach for the general case: benchmarks show it is 2x
+    faster than scatter-add + cumsum because CUDA atomic operations have
+    high per-operation overhead relative to parallel binary search.
+    """
+    import cupy as _cupy
```
seberg: Seems a bit strange to do it here. Might as well do it top-level?

cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +503,128 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+
+    if not numpy.issubdtype(reps.dtype, numpy.integer):
+        raise ValueError(
+            "'repeats' array must have integer dtype, got {}".format(
+                reps.dtype))
```
seberg: I might suggest to just use `reps = reps.astype(np.intp, casting="safe")`
(safe is what NumPy seems to use, but `same_kind` actually makes more sense maybe, since it allows casting from `uint64` even if shady in theory.)

That replaces the `!= intp` check below and raises the error here also.

cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +503,128 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+
+    reps = reps.ravel()
```
seberg: I don't think NumPy does this ravel? a bit unclear why it is here?

cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +503,128 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+
+    if axis is None:
+        a = a.ravel()
+        axis = 0
+
+    axis = internal._normalize_axis_index(axis, a.ndim)
```
seberg:
```suggestion
    else:
        axis = internal._normalize_axis_index(axis, a.ndim)
```

cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +503,128 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+        # This is faster than two separate GPU reductions (min + sum)
+        # followed by scalar D2H syncs, because each sync incurs a
+        # full device synchronization round-trip (~2-5 ms on typical
+        # systems).  CPU min + sum on the host copy is negligible for
+        # typical reps sizes (< 1 M elements).
+        reps_cpu = reps.get()
+        if reps_cpu.min() < 0:
+            raise ValueError(
+                "all elements of 'repeats' should not be negative")
+        total = int(reps_cpu.sum())
```
seberg: Not sure I understand why we would do this on the CPU rather than GPU if we already made it a GPU array?
I mean, sure it may be pretty fast, but we do a cumsum below on the GPU, it seems like a weird mix of things.

Also, `boundaries[-1]` is also the total sum, so we don't have to do this twice.

cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +503,128 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+    cdef Py_ssize_t n_reps = reps.size
+    cdef Py_ssize_t total = 0
+
+    if n_reps == 1:
+        # Broadcast: one D2H transfer to retrieve the scalar.
+        rep_val = int(reps.get()[0])
+        if rep_val < 0:
+            raise ValueError(
+                "'repeats' should not be negative: {}".format(rep_val))
+        return _broadcast_repeat(a, rep_val, axis)
```
seberg: This seems like claude added defensive nonsense?  I.e. it is a logic error if we reach here with size == 1. But it is either fine, or should give some more critical error?

cupy/_core/_routines_manipulation.pyx
seberg: But looking below, it also feels like defensive nonsense to do this here.  At this point it should already be guaranteed a 1-D `intp` array.

cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +503,128 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+    ret_shape[axis] = total
+    if total == 0:
+        return core.ndarray(ret_shape, dtype=a.dtype)
+
+    # boundaries[i] = sum(reps[0..i]).
+    # searchsorted(boundaries, k, 'right') returns the index i such that
+    # boundaries[i-1] <= k < boundaries[i], i.e., k falls in segment i.
+    boundaries = _cupy.cumsum(reps)
+    positions = _cupy.arange(total, dtype=numpy.intp)
+    src_idx = _cupy.searchsorted(boundaries, positions, side='right')
```
seberg: Ah neat trick, I wonder if a custom kernel could be a lot faster, but `repeat` isn't the most trivial to do on the GPU.
And I guess this is a _huge_ improvement already.

cupy/_core/_routines_manipulation.pyx
```diff
@@ -517,71 +633,60 @@ cpdef _ndarray_base _repeat(_ndarray_base a, repeats, axis=None):
     .. seealso:: :func:`numpy.repeat`
 
     """
-    cdef _ndarray_base ret
+    cdef Py_ssize_t rep_val
+    cdef int norm_axis
+
+    # numpy.ndarray instances are rejected before PySequence_Check, which
+    # returns True for numpy arrays (they implement __len__/__getitem__).
+    if isinstance(repeats, numpy.ndarray):
```
seberg: Hmmm, I guess the old code effectively rejected it because `repeats < 1` would fail for all but 0-D arrays (which are rare).

But does it matter to reject it if we don't reject lists (maybe, I don't care either way)?

cupy/_core/_routines_manipulation.pyx
```diff
@@ -517,71 +633,60 @@ cpdef _ndarray_base _repeat(_ndarray_base a, repeats, axis=None):
...
+        raise TypeError(
+            "'repeats' does not accept numpy.ndarray; "
+            "convert with cupy.array() first")
 
     if isinstance(repeats, _ndarray_base):
```
seberg: There were some places that didn't support things broadly, I would suggest also here using the new conversion helper I added:
```
_convert_from_cupy_like(arg, error=False)
```
(it either returns `None` or an `_ndarray_base`.)

cupy/_core/_routines_manipulation.pyx
```diff
@@ -517,71 +633,60 @@ cpdef _ndarray_base _repeat(_ndarray_base a, repeats, axis=None):
...
+        # repeat count, then fall through to the broadcast path below.
+        rep_scalar = int(reps_flat.get()[0])
+        if rep_scalar < 0:
+            raise ValueError(
+                "'repeats' should not be negative: {}".format(rep_scalar))
+        rep_val = rep_scalar
+    elif isinstance(repeats, int) or (
+            hasattr(repeats, 'dtype') and
+            numpy.ndim(repeats) == 0 and
+            numpy.issubdtype(repeats.dtype, numpy.integer)):
```
seberg: I would suggest to just use:
```
try:
    reps = operator.index()
except TypeError:
    pass
```
Admittedly, that is a bit more broad but not much, so if you don't like it.
(I don't love try/excepts, but things like `numpy.ndim` and `numpy.issubdtype` are also all shady and slow)

This would let numpy ndarrays pass maybe...

cupy/_core/_routines_manipulation.pyx
```diff
@@ -517,71 +633,60 @@ cpdef _ndarray_base _repeat(_ndarray_base a, repeats, axis=None):
...
+            if repeats[0] < 0:
+                raise ValueError(
+                    "'repeats' should not be negative: {}".format(repeats[0]))
+            rep_val = repeats[0]
         else:
-            a = a.ravel()
-            axis = 0
+            # Non-broadcast sequence (including empty): GPU per-element path.
+            return _repeat_ndarray_repeats(
+                a, core.array(repeats, dtype=numpy.intp), axis)
```
seberg: Using `core.array(repeats, dtype=numpy.intp)` means we actually accept a list of floats indirectly.
This is maybe fine, not doing so is actually a bit tedious (you need to special case for `len(repeats) == 0`, and otherwise use `numpy.asarray()`, and then do a `same_kind` cast.

tests/cupy_tests/manipulation_tests/test_tiling.py
```diff
@@ -26,19 +26,499 @@ def test_array_repeat(self, xp):
...
-            a.repeat(repeats)
+# ---------------------------------------------------------------------------
+# Helpers shared by ndarray-repeats test classes
+# ---------------------------------------------------------------------------
+
+def _reps(xp, vals):
+    """Make a reps array on the right platform."""
+    if xp is cupy:
+        return cupy.array(vals)
+    return numpy.array(vals)
```
seberg: This function is just silly :), since it is just `xp.array(vals)`

tests/cupy_tests/manipulation_tests/test_tiling.py
```diff
@@ -26,19 +26,499 @@ def test_array_repeat(self, xp):
...
+    # --- all ones (identity) ---
+    {'shape': (5,),   'reps': [1, 1, 1, 1, 1], 'axis': 0},
+    # --- broadcast scalar equivalence ---
+    {'shape': (2, 3), 'reps': [0], 'axis': 0},
+    {'shape': (2, 3), 'reps': [1], 'axis': 1},
+    # --- 4-D ---
+    {'shape': (2, 3, 4, 5), 'reps': [2, 1, 3], 'axis': 1},
+    {'shape': (2, 3, 4, 5), 'reps': [3],        'axis': 3},
+)
+class TestRepeatNdarrayRepeats(unittest.TestCase):
```
seberg: Let's get rid of `unittest` and use normal `@pytest.parametrize`, (all it does is not use `shape` as an argument rather than `self.shape`).

Just as a long-term hope to use a more typical pattern in general.

tests/cupy_tests/manipulation_tests/test_tiling.py
```diff
@@ -26,19 +26,499 @@ def test_array_repeat(self, xp):
...
+        # Verify dtype is preserved
+        assert result.dtype == x.dtype
+        return result
+
+
+def _int_reps(xp, vals):
+    """Make an integer reps array (int32 dtype), even for empty lists."""
+    if xp is cupy:
+        return cupy.array(vals, dtype=numpy.int32)
+    return numpy.array(vals, dtype=numpy.int32)
```
seberg: Also feels like an unnecessary helper?

seberg (commented):
Looks good in general, I have a lot of nitpicking comments, some can be ignored and/or are more general "could be a bit nicer".  As usual, tests seem a bit verbose, but I don't mind much.

(I should look through at it once more, but thought I'd post.)

eriknw:
Thanks for the initial review and nits--keep 'em coming! Seem valid to me. I'll give my human replies on Monday

cupy/_core/_routines_manipulation.pyx
eriknw: Done. Moved to top-level, matching `_routines_sorting.pyx` and other `_core` files.

cupy/_core/_routines_manipulation.pyx
eriknw: Good calls. I reorganized some things.

I used `numpy.can_cast(reps.dtype, intp, casting='safe')` for the check--CuPy's Cython `astype` cpdef requires all positional args, making `casting=` impractical inline.

Chose `safe` over `same_kind` to match NumPy's actual behavior: NumPy's `PyArray_ContiguousFromAny(op, NPY_INTP, 0, 1)` uses safe casting, and `np.repeat(a, np.array([1,2,3], dtype=np.uint64))` raises `TypeError: Cannot cast ... according to the rule 'safe'`.

This also correctly accepts bool (bool -> intp is a safe cast).

cupy/_core/_routines_manipulation.pyx
eriknw: Removed from here. NumPy's C code uses `PyArray_ContiguousFromAny(op, NPY_INTP, 0, 1)` which rejects ndim > 1. Added an explicit ndim > 1 rejection with NumPy's exact error message: `"object too deep for desired array"`. The only remaining ravel is at the scalar extraction point in step 3 (`reps_arr.ravel().get()[0]`) to handle 0-D arrays, which can't be indexed with `[0]` directly.

cupy/_core/_routines_manipulation.pyx
eriknw: Yup. Reworked it. Removed from the inner function. `_repeat` handles size-1 arrays in step 3 (extracts scalar, dispatches to `_broadcast_repeat`). `_repeat_perelement` only receives multi-element arrays.

cupy/_core/_routines_manipulation.pyx
eriknw: Thanks for calling out weird things! This PR was pretty raw Claude, sorry. Anyway. Switched to GPU. `reps.min()` and `cupy.cumsum(reps)` are launched back-to-back so both kernels queue before we sync. The first `int()` waits for both; the second is just a memcpy. Measured: 107 us total for 10K elements (vs 158 us for the bulk D2H approach).

cupy/_core/_routines_manipulation.pyx
eriknw: Agreed--a fused kernel could eliminate the intermediate `arange` and `src_idx` arrays and reduce to 1 launch. Left as future optimization. It'd be faster, but not sure by how much. The current approach is already ~340x faster than the old Python loop.

cupy/_core/_routines_manipulation.pyx
eriknw: I kept the rejection (`TypeError`). CuPy convention: `cupy.put`, `cupy.place`, etc. also reject numpy arrays to prevent silent host-to-device copies. Numpy *scalars* are accepted via `operator.index`. This is a "don't care either way" choice per the reviewer; rejection is the conservative default.

cupy/_core/_routines_manipulation.pyx
eriknw: Done. First check in `_repeat`. Handles cupy ndarrays, `__cuda_array_interface__`, and `__cupy_get_ndarray__`.

cupy/_core/_routines_manipulation.pyx
eriknw: Done. Yeah, I've been using this trick (that I totally borrowed) from NumPy for years. Sorry again for the raw Claude code. Replaces the `isinstance(int) or (hasattr(dtype) and ...)` chain. Handles Python ints, bools, numpy integer scalars, 0-D numpy integer arrays. Rejects floats with `TypeError`.

cupy/_core/_routines_manipulation.pyx
eriknw: Left as-is (for now, up to you). The truncation matches NumPy's behavior. Strict rejection would require special-casing for `len(repeats) == 0`.

tests/cupy_tests/manipulation_tests/test_tiling.py
eriknw: Haha, yeah. Removed. Tests use `xp.array(vals)` directly.

tests/cupy_tests/manipulation_tests/test_tiling.py
eriknw: Understood. Removed `unittest.TestCase` inheritance from all new test classes. Kept `@testing.parameterize` for consistency with the rest of CuPy's test suite (thousands of tests use this pattern, it's not _too_ bad); migrating to `@pytest.parametrize` would require reworking the `@testing.numpy_cupy_*` decorators and is better done as a broader effort across the codebase.

tests/cupy_tests/manipulation_tests/test_tiling.py
eriknw: Removed.

eriknw (commented):
Thanks again for all the feedback @seberg! I wasn't expecting quite so much given that I hadn't yet shaken this down and cleaned it up--it was pretty raw Claude (I was heavily involved with experiments, but not so close to the code before). I'll try to be more clear about this in the future.

## Claude's Summary of changes

**Implementation (`_routines_manipulation.pyx`):**
- Three functions: `_repeat` (dispatch), `_broadcast_repeat` (scalar),
  `_repeat_perelement` (per-element GPU)
- 3-step dispatch: convert -> normalize axis -> dispatch
- `_convert_from_cupy_like` + `operator.index` + `PySequence_Check`
- Async GPU validation (min + cumsum back-to-back, one effective sync)
- `can_cast(safe)` for dtype validation (accepts bool, rejects float/uint64)
- ndim > 1 rejection matching NumPy ("object too deep for desired array")
- `cdef` for internal functions (no Python call overhead)
- Sequence recursion (lists -> cupy array -> re-enter cupy-like path)

**Tests (`test_tiling.py`):**
- 84 tests, 0 failures
- Cross-validation against NumPy for all error types
- Bool per-element and broadcast tests
- uint32 acceptance / uint64 rejection tests
- ndim > 1 rejection test
- Method interface test (`.repeat()` vs `cupy.repeat()`)
- No `unittest.TestCase`, no helper functions

**Docstring (`tiling.py`):**
- Updated `repeats` parameter to document cupy.ndarray support

cupy/_core/_routines_manipulation.pyx
seberg: uggg, a small issue in NumPy in my opinion, but one that apparently nobody ever noticed, hehe.
(NumPy now has a funny `same_value` casting level, but I think that is too awkward for CuPy to adopt any time soon.)

Either way, sounds good to just follow NumPy, resolving.

tests/cupy_tests/manipulation_tests/test_tiling.py
seberg: Ah, yeah, there may be some weird little things, anyway, I agree it isn't too important.
(The only important thing is to never mutate `self`, which this doesn't do.)

cupy/_core/_routines_manipulation.pyx
```diff
@@ -517,71 +577,63 @@ cpdef _ndarray_base _repeat(_ndarray_base a, repeats, axis=None):
...
-            return ret.ravel()
+        a = a.ravel()
+        norm_axis = 0
+    norm_axis = internal._normalize_axis_index(axis, a.ndim)
+
+    # --- Step 3: dispatch ---
+
+    if reps_arr is not None:
+        if reps_arr.size == 1:
+            rep_val = int(reps_arr.ravel().get()[0])
```
seberg:
```suggestion
            rep_val = reps_arr.item()
```
should work fine (but doesn't matter much).

cupy/_core/_routines_manipulation.pyx
```diff
@@ -503,12 +506,69 @@ cpdef _ndarray_base broadcast_to(_ndarray_base array, shape):
...
+        ret_shape = list(a.shape)
+        ret_shape[axis] = 0
+        return core.ndarray(ret_shape, dtype=a.dtype)
+
+    # Launch min and cumsum back-to-back so both GPU kernels are
+    # queued before we sync.  The first int() waits for both;
+    # the second is just a memcpy (~0.02 ms).
+    reps_min = reps.min()
+    boundaries = cupy.cumsum(reps)
+    total = int(boundaries[-1])
```
seberg:
```suggestion
    total = int(boundaries[-1])  # synchronize!
```
I just learned recently that this is a CuPy code-base convention.

cupy/_core/_routines_manipulation.pyx
```diff
@@ -517,71 +577,63 @@ cpdef _ndarray_base _repeat(_ndarray_base a, repeats, axis=None):
...
+    elif isinstance(repeats, numpy.ndarray):
+        raise TypeError(
+            "'repeats' does not accept numpy.ndarray; "
+            "convert with cupy.array() first")
+    else:
+        try:
+            rep_val = operator.index(repeats)
+            reps_arr = None
+        except TypeError:
+            if not cpython.PySequence_Check(repeats):
```
seberg, line 605: Maybe a bit awkward to recurse, especially in the `except:` clause?  But of course the alternative is to set `rep_val = -1` and do an `if rep_val < 0` after the `try/except: pass`.
The `reps_arr is not None:` checks could (partially?) be moved later as well to avoid recursing.

tests/cupy_tests/manipulation_tests/test_tiling.py
```diff
@@ -9,6 +9,8 @@
 from cupy import testing
 
 
+# ---- Pre-existing repeat tests (scalar and list repeats) ----
```
seberg: Very opinionated and nitpicky, but I think we might as well delete all of these comments maybe.  "Pre-existing" won't date well.

seberg (commented):
Thanks a few more nitpicks, but I think this is basically goo and a big improvement. Do you want a large test 32bit overflow in (result?) elements?
(I don't care too much, if we add it, mark it with `@pytest.mark.slow`.)

CI is also failing for some silly lint issue.

If this causes regressions for anyone due to synchronization, I think we should go the JAX route and add a parameter for the correct result size.
(But that can be a follow-up probably.)

tests/cupy_tests/manipulation_tests/test_tiling.py
```diff
@@ -104,6 +82,200 @@ def test_repeat_failure(self):
...
+        for xp in (numpy, cupy):
+            with pytest.raises(TypeError):
+                xp.repeat(xp.arange(3), xp.array([1.0, 1.0, 1.0]), 0)
+
+    def test_uint64_matches_numpy(self):
+        # Both reject uint64 (unsigned → signed is unsafe)
+        for xp in (numpy, cupy):
+            with pytest.raises(TypeError):
+                xp.repeat(xp.arange(3),
+                           xp.array([1, 2, 3], dtype=numpy.uint64))
```
seberg:
```suggestion
                xp.repeat(xp.arange(3),
                          xp.array([1, 2, 3], dtype=numpy.uint64))
```

seberg (approved):
Thanks, the recursion seems still slightly awkward, but I think it is good enough.

Going to add the "breaking" label, since it adds synchronization.  This also breaks NumPy array repeat values implicitly, which I am not sure about whether that seems acceptable or not (let's discuss it very briefly today).

Otherwise, I think this is a big improvement and it looks good to me to merge now.

EDIT: Sorry, nvm, numpy ndarrays already fail on some weird thing before reaching the sequence path.

seberg:
Thanks Erik!  If anyone runs into the synchronization as a serious issue, then we should add the JAX kwarg only argument that would be the total length (and disable many of the sanity checks).

leofang:
Thanks @eriknw @seberg for getting this across the finish line, and @eriknw for your first (merged) PR to CuPy! 🎉

PR #9831, opened 2026-03-25
Issues #8434 fixed.

### LINKED ISSUE #8434: Square of complex number with zero imaginary part has nonzero imaginary part?
### Description

The code below demonstrates the issue. I expected the square of the complex number with nonzero imaginary part to have exactly zero imaginary part, but it has a small nonzero imaginary part that got amplified in my application.

I know floating point math can be quirky,  and I can reproduce the problem with a PyTorch tensor on the GPU, so this is probably a known CUDA thing. But I couldn't find reports of the same problem, so I thought I'd go ahead and send it to you.

### To Reproduce

```py
import cupy as cp
cp.asarray([-1 + 0j])**2  # array([1.-2.4492936e-16j])
```

### Installation

### Environment

```
(scipy-dev) C:\Users\matth\scipy>python -c "import cupy; cupy.show_config()"
OS                           : Windows-11-10.0.22631-SP0
CuPy Version                 : 13.1.0
NumPy Version                : 1.26.4
SciPy Version                : 1.15.0.dev0+1218.6afa9f1
Cython Build Version         : 0.29.37
Cython Runtime Version       : 3.0.10
CUDA Runtime Version         : 12040 (linked to CuPy) / RuntimeError("CuPy failed to load cudart64_12.dll: FileNotFoundError: Could not find module 'cudart64_12.dll' (or one of its dependencies). Try using the full path with constructor syntax.") (locally installed)
cuFFT Version                : 11203
cuRAND Version               : 10306
cuSOLVER Version             : (11, 6, 2)
NVRTC Version                : (12, 5)
Device 0 Name                : NVIDIA GeForce RTX 2060 SUPER
```

### Additional Information

The imaginary part is proportional to the square of the real part, and it only seems to happen when the real part is negative.
```python3
array([1.+0.j])  # array([1.+0.j])
cp.asarray([-1 - 0j])**2  # array([1.-2.4492936e-16j])
cp.asarray([-10 - 0j])**2  # array([100.-2.4492936e-14j])
cp.asarray([-100 - 0j])**2  # array([10000.-2.4492936e-12j])
```

It doesn't happen on my CPU.

```python3
(-1. - 0.j)**2  # (1+0j)

import numpy as np
np.asarray([(-1. - 0.j)])**2  # array([1.-0.j])
```

dufp2002:
This PR fixes the power operator issues:

Added fast paths in cupy.power for cases where complex inputs lie on the real axis, reducing complex computation to real pow when possible. The change preserves complex edge-case behavior more carefully.

Fixes #8434

### DIFF
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -1024,7 +1024,29 @@ inline __device__ T integral_power(T in0, T in1) {
 
 template <typename T>
 inline __device__ T complex_power(T in0, T in1) {
-    return in1 == T(0) ? T(1): pow(in0, in1);
+    if (in1 == T(0)) {
+        return T(1);
+    }
+
+    if (in1 == T(1)) {
+        return in0;
+    }
+
+    using value_type = decltype(in0.real());
+    if (in0 == T(0)) {
+        if (in1.real() > value_type(0)) {
+            return T(0);
+        }
+
+        value_type nanv = value_type(nan(""));
+        return T(nanv, nanv);
+    }
+
+    if (in0.imag() == value_type(0) && in1.imag() == value_type(0) && (in0.real() > value_type(0) | in1.real() > value_type(1))) {
+        return T(pow(in0.real(), in1.real()), value_type(0));
+    }
+
+    return pow(in0, in1);
 }
 '''
 
--- a/cupy/_core/_routines_math.pyx
+++ b/cupy/_core/_routines_math.pyx
@@ -1042,7 +1042,8 @@ inline __device__ T complex_power(T in0, T in1) {
         return T(nanv, nanv);
     }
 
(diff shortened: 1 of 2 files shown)


seberg:
Thanks @dufp2002, can you do some initial timings for when this path isn't hit?  NumPy does indeed use fast-paths like this, but only uses them for integers or if the second input is constant (stride 0, although I think it would be fair to just limit this to scalar, in which case we specialize -1, 0., 0.5, 1., 2. -- for floats, for ints only 0, 1, 2).

Even if we go just with scalar, we may want to time it once, but I would suspect with that it is rather safe.  In principle, we could also just specialize the `**` operator (which is what NumPy _used_ to do for quite some time, I think.)

This is really nice, but we may have to move such special handling into the actual `ElementwiseKernel` at least unless timings show that it is completely fine.  (or higher/approach it from the opposite side)

PR #9832, opened 2026-03-25
fix: raise AttributeError for NumPy 2.0 removed type aliases

### LINKED ISSUE #9818: TypeError: Alias 'bool8' was removed in NumPy 2.0.
### Description

TypeError: Alias 'bool8' was removed in NumPy 2.0. Use a name without a digit at the end.
Am using numpy==2.2.0

### To Reproduce

```py
# Write the code here
```

### Installation

### Environment

```
# Paste the output here
```

### Additional Information

_No response_

thakoreh:
Fixes #9818

bool8, int0, and uint0 were removed in NumPy 2.0. The module
__getattr__ shim still tried to forward these to numpy via
getattr(np, name), which raises TypeError in NumPy 2.0 instead of
the expected AttributeError.

This moves them out of the general deprecated list into a separate
_removed_apis dict that raises AttributeError with a helpful message
suggesting the replacement type.

### DIFF
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -939,7 +939,7 @@ def show_config(*, _full=False):
     _sys.stdout.flush()
 
 
-_deprecated_apis = []
+_deprecated_apis: list[str] = []
 
 # APIs removed in NumPy 2.0 that we previously shimmed
 _removed_apis = {
--- a/cupy/__init__.py
+++ b/cupy/__init__.py
@@ -939,11 +939,14 @@ def show_config(*, _full=False):
     _sys.stdout.flush()
 
 
-_deprecated_apis = [
-    'int0',
-    'uint0',
-    'bool8',
-]
+_deprecated_apis = []
+
+# APIs removed in NumPy 2.0 that we previously shimmed
+_removed_apis = {
+    'bool8': 'bool',
+    'int0': 'int',
+    'uint0': 'uint',
+}
 
 
 # np 2.0: XXX shims for things removed in np 2.0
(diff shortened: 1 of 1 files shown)


seberg:
Sorry for not looking at this earlier.  If you have some time, could you just delete this entirely?

Unlike in NumPy itself, the shimming here was just to avoid the warning at import time.  This already is a plain `AttributeError` right now (since all supported NumPy versions have an `AttributeError`).  So we can just delete the shims entirely.

seberg:
Maybe to be more clear: The issue report was spurious anyway, the only reason to change this is to clean up the code eventually. (Although I think right now this actually already inherits the slightly nicer error message NumPy does.)

We should just remove it entirely (and maybe not backport), or do nothing.

cupy/__init__.py
```diff
@@ -1133,14 +1123,6 @@ def safe_eval(*args, **kwds):  # type: ignore [misc]
 
 
 def __getattr__(name):
```
seberg:
```suggestion
```

cupy/__init__.py
```diff
@@ -1133,14 +1123,6 @@ def safe_eval(*args, **kwds):  # type: ignore [misc]
...
-    if name in _removed_apis:
-        raise AttributeError(
-            f"module 'cupy' has no attribute {name!r}. "
-            f"Use '{_removed_apis[name]}' instead."
-        )
-    if name in _deprecated_apis:
-        return getattr(_numpy, name)
-
     raise AttributeError(f"module 'cupy' has no attribute {name!r}")
```
seberg:
```suggestion
```
Might as well remove the whole geattr for now, I think.

seberg (dismissed):
Thanks, can just get this in, I think... We don't need the getattr for now, I think we can re-add it when we do.

seberg:
This is simple now, just deleting the `__getattr__` that we currently don't need.  I'll put it in, I just hadn't activated auto-merge.

Skipping, since it just wants to run the new free-threaded tests now.
/test skip

seberg (approved):

Issue #9838, opened 2026-03-27
NumPy 2.4: `linalg.solve` now raises LinAlgError for singular matrix with empty RHS


leofang:
## Description

`test_solve_singular_empty` in `cupy_tests/linalg_tests/test_solve.py` fails on NumPy 2.4.x (observed with NumPy 2.4.3, Python 3.14.0, CUDA 13.0).

## Root cause

The test calls `xp.linalg.solve(a, b)` with a singular matrix `a = xp.zeros((3, 3))` and an empty RHS `b = xp.empty((3, 0))`. It expects both NumPy and CuPy to silently return an empty array (since `nrhs = 0`).

However, **NumPy 2.4 changed behavior**: `numpy.linalg.solve` now checks for singularity and raises `LinAlgError("Singular matrix")` even when the RHS has 0 columns. CuPy still returns silently, so the `numpy_cupy_allclose` test helper reports `AssertionError: Only numpy raises error`.

## Failing tests (4 total)

```
FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_0_{batched_gesv_limit=None, order='C'}::test_solve_singular_empty
FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_1_{batched_gesv_limit=None, order='F'}::test_solve_singular_empty
FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_2_{batched_gesv_limit=0, order='C'}::test_solve_singular_empty
FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_3_{batched_gesv_limit=0, order='F'}::test_solve_singular_empty
```

## Traceback

```
    @testing.numpy_cupy_allclose()
    def test_solve_singular_empty(self, xp):
        a = xp.zeros((3, 3))  # singular
        b = xp.empty((3, 0))  # nrhs = 0
        # LinAlgError("Singular matrix") is not raised
>       return xp.linalg.solve(a, b)

numpy.linalg.LinAlgError: Singular matrix
AssertionError: Only numpy raises error
```

## CI log

https://storage.googleapis.com/chainer-artifacts-pfn-public-ci/cupy-ci/213251/log.txt

## Discussion

[...]

PR #9840, opened 2026-03-27
Skip test_solve_singular_empty on NumPy >= 2.4

leofang:
NumPy 2.4 changed `linalg.solve` to check for singularity even when `nrhs=0` (empty RHS), raising `LinAlgError("Singular matrix")`. CuPy intentionally skips the singularity check by default (`errstate(linalg='ignore')`) for performance, so the comparison test is no longer meaningful on NumPy >= 2.4.

See #9838 for the full fix plan when the NumPy baseline is bumped.

-- Leo's bot

### DIFF
--- a/tests/cupy_tests/linalg_tests/test_solve.py
+++ b/tests/cupy_tests/linalg_tests/test_solve.py
@@ -63,6 +63,7 @@ class TestSolve(unittest.TestCase):
             with pytest.raises(error_type):
                 xp.linalg.solve(a, b)
 
+    @testing.with_requires('numpy<2.4')
     @testing.numpy_cupy_allclose()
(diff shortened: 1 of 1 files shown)


seberg:
Thanks, let's get this in so CI stops failing.

Yeah, I guess the full fix is to either make sure we also raise (with that the error mode), or just accept both.
Since the matrix _is_ singular, I don't think there is a reason to change NumPy.

seberg (approved):

leofang:
Right, we'll fix it when bumping the baseline to NumPy 2.4 (see Part 2 in https://github.com/cupy/cupy/issues/9838#issuecomment-4144721016).

Issue #9830, opened 2026-03-25
CuPy-on-ROCm: `compute_capability` collision causes kernel cache corruption on heterogeneous multi-GPU systems


alexschroeter:
### Description

# CuPy-on-ROCm: `compute_capability` collision causes kernel cache corruption on heterogeneous multi-GPU systems

## Summary

`cupy.cuda.Device(i).compute_capability` returns the same string for GPUs
with different GCN architectures (e.g. gfx906 and gfx908 both return
`"90"`).  CuPy uses this value as part of its kernel disk-cache key, so a
binary compiled for one GPU architecture is incorrectly loaded on another,
producing `hipErrorNoBinaryForGpu` at runtime.

## Environment

| Component | Version |
|-----------|---------|
| CuPy | 14.0.1 |
| ROCm / HIP runtime | 6.2.41134 |
| GPU 0 | AMD Radeon VII (gfx906:sramecc+:xnack-) |
| GPU 1 | AMD Instinct MI100 (gfx908:sramecc+:xnack-) |
| OS | Linux 6.8.0-87-generic |
| Python | 3.11.0rc1 |

## Reproducer

```python
import cupy as cp

# Step 1: Show that compute_capability does not distinguish the GPUs
for i in range(cp.cuda.runtime.getDeviceCount()):
    props = cp.cuda.runtime.getDeviceProperties(i)
    gcn = props.get("gcnArchName", b"").decode()
    cc = cp.cuda.Device(i).compute_capability
    print(f"Device {i}: gcnArchName={gcn}  compute_capability={cc}")
# Output:
#   Device 0: gcnArchName=gfx906:sramecc+:xnack-  compute_capability=90
#   Device 1: gcnArchName=gfx908:sramecc+:xnack-  compute_capability=90

# Step 2: Trigger the cache collision
cp.cuda.Device(0).use()
a = cp.zeros(10)          # compiles & caches elementwise kernel for gfx906

cp.cuda.Device(1).use()
b = cp.zeros(10)          # loads cached gfx906 binary on gfx908 -> crash
```

## Error

```
CUDADriverError: hipErrorNoBinaryForGpu: no kernel image is available
for execution on the device
```

## Root cause

CuPy computes `compute_capability` from the HIP `gcnArch` property via
integer division that discards the last digit of the architecture number:

| GPU | gcnArchName | gcnArch (expected) | compute_capability |
|-----|-------------|--------------------|--------------------|
[...]

PR #9847, opened 2026-04-01
fix(hip): use gcnArchName for cache keying and return sentinel compute_capability on HIP

alexschroeter:
See Isssue https://github.com/cupy/cupy/issues/9830

### DIFF
--- a/cupy/cuda/device.pyx
+++ b/cupy/cuda/device.pyx
@@ -205,9 +205,15 @@ cdef class Device:
     def compute_capability(self):
         """Compute capability of this device.
 
-        The capability is represented by a string containing the major index
-        and the minor index. For example, compute capability 3.5 is represented
-        by the string '35'.
+        On CUDA the capability is represented by a string containing the
+        major index and the minor index. For example, compute capability
+        3.5 is represented by the string ``'35'``.
+
+        On HIP/ROCm this returns ``'9999'`` so that all CUDA
+        feature-level threshold checks (e.g. ``int(cc) >= 70``) pass
+        unconditionally.  The real GCN architecture name is used
+        internally for kernel cache keying via
+        :func:`cupy.cuda.compiler._get_arch`.
 
         """
         if self.id in _compute_capabilities:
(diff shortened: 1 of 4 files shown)


cupy/cuda/device.pyx
```diff
@@ -215,11 +216,25 @@ cdef class Device:
...
+                # precision (e.g. gfx906 and gfx908 both map to "90").
+                # Use gcnArchName to get the full architecture number.
+                props = runtime.getDeviceProperties(self.id)
+                try:
+                    arch_name = props['gcnArchName'].decode()
+                except KeyError:  # ROCm < 3.6.0
+                    arch_name = 'gfx' + str(props['gcnArch'])
+                # Arch names may contain hex letters (e.g. gfx90a)
+                # and feature flags (e.g. gfx906:sramecc+:xnack-)
+                m = re.match(r'gfx([0-9a-fA-F]+)', arch_name)
```
gpinkert: Something like this seems simpler and less likely to break than a regex:

```python 
arch = gcn_arch_name.split(b':')[0].decode()
```

You could also just hash the entire string arch string and use that as the key so you don't have to worry about changing the code in the future to handle different types of string formatting.

alexschroeter: That is a good point, this is unnecessarily complicated. It seems the `higher compute capability` comparison that CUDA can do is not possible with the arch name under ROCm. 

But I think there is a bigger issue, where the compute_capability is cast to int. In some cases like '90a' as well as 'gfx90a' this would fail for rocm during the conversion. Here is an example but there are many more occasions:

https://github.com/cupy/cupy/blob/b316c6d35dc483b1e75bf792463bc89854dde492/cupy/cuda/cufft.pyx#L1140-L1141

At least here the solution seems to be guarding against hip like this:

https://github.com/cupy/cupy/blob/b316c6d35dc483b1e75bf792463bc89854dde492/cupyx/cusolver.pyx#L120-L121

The question if this should be guarded everywhere as well or if there is a different solution.

The option that makes the most sense to me at the moment would, only changing the cache key and keeping the broken shortening for the compute_capability case. But I see this breaking very soon for the gfx11xx, gfx12xx.

alexschroeter: Thinking about the issue a bit more I decided that these are actually multiple problems which are somewhat related.

- The caching path error due to the shortening of the ROCm arch name ("gfx906" -> 90, "gfx908" -> 90, "gfx90a" -> 90).
- The matter in which the compute_capability is shortened. ("gfx908" -> 90, "gfx1102" -> 11 or 2 - two digits).
- And that the compute_capability comparison for ROCm does not work that way.

Which can be split into two main issues: 1. caching 2. compute_capability

I believe 
1. is best solved as @gpinkert suggested by using the full arch name. This solves all issues regarding shortening of the device name and the multidevice caching that brought me to this issue.

2. Here I see a problem coming up real soon when new devices like the "gfx1102" are shortened to 11 and the compute_capability check present in many files will fail because the numerical comparison makes no sense. Until this becomes a real issue meaning, there also needs to be a check for compute_capability under ROCm I believe the best solution is to return a decently high value when hip is used.

@gpinkert, @leofang Do you have any thoughts on this?

leofang:
@alexschroeter one logistic ask: Please kindly make PR titles more descriptive to help us triage PRs/issues faster :)

cupy/cuda/device.pyx
leofang: Sorry, I do not have good advices here. I added the ROCm support to CuPy, but I have never quite figured out the relation between the different archs on the ROCm side. (And I had a lot less archs to worry about back then.)

From CUDA's perspective, the "compute capability" carries semantic meanings, but I am unaware of if a HIP/ROCm counterpart exists. @kmaehashi might know better...

From CuPy's perspective, this PR might be introducing a breaking change (as far as ROCm users are concerned)?

PR #9849, opened 2026-04-03
fix(hip): use event-based sync for cross-device D2D copies

gpinkert:
On HIP/ROCm, cross-device memcpy with hipMemcpyAsync requires the active device to be set to the source device, per hip_runtime_api.h: "For multi-gpu or peer-to-peer configurations, it is recommended to use a stream which is attached to the device where the src data is physically located."

Without this, the D2D copy runs on the caller's current stream with no ordering guarantee on the destination device's stream. Any subsequent operation on the destination device (e.g. a .get() readback) can race with the in-flight copy, producing corrupted data. This manifests as flaky failures in multi-GPU tests, especially when tests run in parallel.

Both ndarray.copy() and copyto() now:
1. Switch to the source device context before issuing the copy
2. Record an event on the source device's stream after the copy
3. Switch to the destination device and wait on that event

This ensures correct cross-stream ordering between the D2D copy and any subsequent work on the destination device.

Same-device copies and all CUDA paths are unaffected.

### DIFF
--- a/cupy/_manipulation/basic.py
+++ b/cupy/_manipulation/basic.py
@@ -101,7 +101,7 @@ def copyto(dst, src, casting='same_kind', where=None):
         return
 
     if _can_memcpy(dst, src):
-        if src.device != dst.device and runtime.is_hip:
+        if runtime.is_hip and src.device != dst.device:
             # From hip_runtime_api.h (hipMemcpyAsync): "For multi-gpu
             # or peer-to-peer configurations, it is recommended to
             # use a stream which is attached to the device where the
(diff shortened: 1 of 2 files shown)


cupy/_manipulation/basic.py
```diff
@@ -100,7 +101,21 @@ def copyto(dst, src, casting='same_kind', where=None):
         return
 
     if _can_memcpy(dst, src):
-        dst.data.copy_from_async(src.data, src.nbytes)
+        if src.device != dst.device and runtime.is_hip:
```
leofang: nit: check HIP first
```suggestion
        if runtime.is_hip and src.device != dst.device:
```

gpinkert: done

leofang (approved):

leofang:
Smoking gun test:

/test cuda131

PR #9769, opened 2026-03-04
Use expected precision in `cupyx.scipy.ndimage` interpolation functions (zoom, shift, rotate, affine_transform, map_coordinates)

grlee77:
It is expected that ndimage interpolation functions use:
- float32 precision for float32 inputs and 8 or 16-bit integer or unsigned integer inputs
- float64 precision for float64 inputs and larger integer dtypes

Currently, the expected precision is being used during spline prefiltering, but the main interpolation kernels involve many undesired promotion to double dtypes. This is due to a combination of double precision floating point literals and some variables being declared with `double` type regardless of image precision.

## Benchmark results

The results below are for `rotate` with various interpolation orders. `prefilter` was set to False to better isolation the kernel being changed in this MR. As expected, changing computations from float64->float32 when applicable results in substantial performance benefit (on RTX A6000). There is no regression in performance for the 64-bit case.

| shape         | order | prefilter | dtype   | duration (ms) | duration, main (ms) | acceleration |
|---------------|-------|-----------|---------|---------------|---------------------|--------------|
| (3840, 2160)) |   0   | False     | uint8   |     0.489     |        1.043        |     2.13     |
| (3840, 2160)) |   1   | False     | uint8   |     0.502     |        1.682        |     3.35     |
| (3840, 2160)) |   2   | False     | uint8   |     0.620     |        2.874        |     4.64     |
| (3840, 2160)) |   3   | False     | uint8   |     0.722     |        5.873        |     8.13     |
| (3840, 2160)) |   4   | False     | uint8   |     0.836     |        7.935        |     9.49     |
| (3840, 2160)) |   5   | False     | uint8   |     1.397     |       11.010        |     7.88     |
| (3840, 2160)) |   0   | False     | float32 |     0.417     |        0.870        |     2.09     |
| (3840, 2160)) |   1   | False     | float32 |     0.420     |        1.489        |     3.54     |
| (3840, 2160)) |   2   | False     | float32 |     0.518     |        2.494        |     4.81     |
| (3840, 2160)) |   3   | False     | float32 |     0.621     |        5.244        |     8.44     |
| (3840, 2160)) |   4   | False     | float32 |     0.746     |        7.049        |     9.45     |
| (3840, 2160)) |   5   | False     | float32 |     1.312     |        9.784        |     7.46     |
| (3840, 2160)) |   0   | False     | float64 |     0.891     |        0.876        |     ~same    |
[...]

PR #9857, opened 2026-04-09
Fix remaining floating-point inconsistencies and improve tests for `cupyx.scipy.ndimage.interpolation`

grlee77:
This is a small follow-up to #9769, with a fixes for a couple remaining issues in dtype usage.

In the prior PR, testing was done primarily with `rotate` and a few places with undesired use of float64 casting of the `shift`, `zoom` or `matrix` coefficient arrays remained in other functions. Also, the `float64_dtype` parameter defaulted to True for `affine_matrix`, which was inconsistent with the float32 default always used internally by `rotate`, `zoom` and `shift`.

There are no changes to the CUDA kernels themselves in this MR. Test cases are updated to make sure float32 precision is tested as expected for all of these functions.

cc @seberg since you already reviewed the prior PR.

### DIFF
--- a/cupyx/scipy/ndimage/_interpolation.py
+++ b/cupyx/scipy/ndimage/_interpolation.py
@@ -322,7 +322,7 @@ def map_coordinates(input, coordinates, output=None, order=3,
 
 def affine_transform(input, matrix, offset=0.0, output_shape=None, output=None,
                      order=3, mode='constant', cval=0.0, prefilter=True, *,
-                     texture_memory=False, float64_coords=True):
+                     texture_memory=False, float64_coords=False):
     """Apply an affine transformation.
 
     Given an output image pixel index vector ``o``, the pixel value is
@@ -379,6 +379,8 @@ def affine_transform(input, matrix, offset=0.0, output_shape=None, output=None,
             - ``order=0`` (nearest neighbor) and ``order=1`` (linear
                 interpolation)
             - NVIDIA CUDA GPUs
+        float64_coords (bool): If True, force double precision computations
+            internally as in scipy.ndimage.
 
     Returns:
         cupy.ndarray or None:
@@ -434,13 +436,13 @@ def affine_transform(input, matrix, offset=0.0, output_shape=None, output=None,
     float_dtype = cupy.float64
     if not float64_coords:
         float_dtype = cupy.promote_types(input.real.dtype, cupy.float32)
+    matrix = matrix.astype(float_dtype, copy=False)
     if mode == 'opencv' or mode == '_opencv_edge':
         if matrix.ndim == 1:
             matrix = cupy.diag(matrix)
-        coordinates = cupy.indices(output_shape, dtype=cupy.float64)
+        coordinates = cupy.indices(output_shape, dtype=float_dtype)
         coordinates = cupy.dot(matrix, coordinates.reshape((input.ndim, -1)))
         coordinates += cupy.expand_dims(cupy.asarray(offset), -1)
-        coordinates = coordinates.astype(float_dtype, copy=False)
         ret = _util._get_output(output, input, shape=output_shape)
         ret[:] = map_coordinates(input, coordinates, ret.dtype, order, mode,
                                  cval, prefilter).reshape(output_shape)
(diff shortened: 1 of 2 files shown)


cupyx/scipy/ndimage/_interpolation.py
```diff
@@ -434,13 +436,13 @@ def affine_transform(input, matrix, offset=0.0, output_shape=None, output=None,
     float_dtype = cupy.float64
     if not float64_coords:
         float_dtype = cupy.promote_types(input.real.dtype, cupy.float32)
+    matrix = matrix.astype(float_dtype, copy=False)
     if mode == 'opencv' or mode == '_opencv_edge':
         if matrix.ndim == 1:
             matrix = cupy.diag(matrix)
-        coordinates = cupy.indices(output_shape, dtype=cupy.float64)
+        coordinates = cupy.indices(output_shape, dtype=float_dtype)
```
seberg, line 443: heh, much nicer :).

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
```diff
@@ -178,43 +249,50 @@ def _affine_transform(self, xp, scp, a, matrix):
             matrix[-1, 0:-1] = 0
             matrix[-1, -1] = 1
         affine_transform = scp.ndimage.affine_transform
+
+        # use double precision for comparison to SciPy results
+        # extra_kwargs = dict(float64_coords=True) if xp == cupy else {}
+        extra_kwargs = {}
```
seberg, line 255: Not actually used?!  So remove it again, or is there a reason to keep it?

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
```diff
@@ -178,43 +249,50 @@ def _affine_transform(self, xp, scp, a, matrix):
...
-    @testing.numpy_cupy_allclose(rtol=1e-6, atol=1e-5, scipy_name='scp')
+    @testing.numpy_cupy_allclose(rtol={"default": 1e-7, numpy.complex64: 1e-4},
+                                 atol=1e-4, scipy_name='scp')
     @testing.with_requires('scipy>=1.6.0')
     def test_affine_transform_complex_float(self, xp, scp, dtype):
-        if self.output == numpy.float64:
+        if self.output == numpy.float32:
             # must promote output to a complex dtype
-            self.output = numpy.complex128
+            self.output = numpy.complex64
```
seberg, line 284: We have to avoid any mutation of `self` for free-threaded testing, unfortunately.

Which means we'll have to pass it in as a parameter to `_affine_transform`  (although, it could default to `self.output` if `None` is passed in)

seberg (commented):
Thanks LGTM, the tolerance changes, I assume are fine and the new tests are nice (I didn't look at the matrix generation too closely).

I guess the `float64_coords` change was just going back to what we had?  (Just a reminder, I'll trust the choice.)

The biggest issue is that unfortunately we can't modify `self.output` in all the tests, will need to pass that around (unless you can push it into the helper function).

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
grlee77: Can remove it, thanks

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
grlee77: Good to know, let me update to avoid the issue

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
grlee77: removed

tests/cupyx_tests/scipy_tests/ndimage_tests/test_interpolation.py
grlee77: refactored to avoid mutation of `self`

seberg (approved):
Thanks @grlee77, LGTM, let's get it in.

PR #9859, opened 2026-04-10
scipy.ndimage.label index overflow

### LINKED ISSUE #9851: cypyx.scipy.ndimage.label misbehaves for large arrays
### Description

I noticed that the results of cypyx.scipy.ndimage.label are inconsistent for large arrays. If I have an array "masks" of shape (N, H, W), the output of cupyx.scipy.ndimage.label(masks, structure=structure) with the correct structure is consistent with doing the same slice by slice, i.e. [cupyx.scipy.ndimage.label(masks[i]) for i in range(len(masks))]. The actual label numbers are different, but the number of labels and their groupings is the same. But if the masks array is large enough, both approaches are only consistent up to a point, and then cupyx.scipy.ndimage.label(masks, structure=structure) yields only 0 for every subsequent slice.
It seems to be related to the total number of elements in the array and not to the size in a particular dimension, e.g. (5000,768,768) fails but (10000,100,100) does not.

I've tested the same with scipy.ndimage.label and this behaviour does not occur.

### To Reproduce

```py
import cupy as cp
import cupyx.scipy.ndimage
import matplotlib.pyplot as plt
import numpy as np

# Make synthetic data, large uint8 array with random number of binary circles
N = 5000

# Image size
h, w = 768, 768
masks = cp.zeros((N, h, w), dtype=np.uint8)
# Create a coordinate grid once (efficient)
Y, X = cp.ogrid[:h, :w]
for i in range(N):
    # Random number of circles
    num_circles = cp.random.randint(5, 20)
    for _ in cp.arange(num_circles):
        # Random center and radius
        cx = cp.random.randint(0, w)
        cy = cp.random.randint(0, h)
        r = cp.random.randint(10, 80)
        # Create a mask for the circle
        mask = (X - cx) ** 2 + (Y - cy) ** 2 <= r * r
        masks[i][mask] = 1

# masks now contains your 3D image with circles
plt.imshow(masks[0].get(), cmap="gray")

structure = cp.array([[[0, 0, 0], [0, 0, 0], [0, 0, 0]], [[0, 1, 0], [1, 1, 1], [0, 1, 0]], [[0, 0, 0], [0, 0, 0], [0, 0, 0]]], dtype="uint8")
for b in [10,50,100,500,1000,5000]:
    labels_2d, num_features_2d = zip(*[cupyx.scipy.ndimage.label(masks[i]) for i in range(b)])
    labels_2d = cp.array(labels_2d)
    num_features_2d = sum(num_features_2d)
    labels_3d, num_features_3d = cupyx.scipy.ndimage.label(masks[:b], structure=structure)

    # Adjust values so that labels_2d and labels_3d have the same range of values
    for i in range(1, b):
        labels_2d[i][labels_2d[i] > 0] += labels_2d[i - 1].max()
    print(f"For N={b}")
    print(f" - Number of features in 2D: {num_features_2d}, Number of features in 3D: {num_features_3d}")
    print(f" - Label arrays are equal: {cp.array_equal(labels_2d, labels_3d)}")
    print(f" - Number of features are equal: {num_features_2d == num_features_3d}")
    print(f" - Max label in 2D: {labels_2d.max()}, Max label in 3D: {labels_3d.max()}")
    if not cp.array_equal(labels_2d, labels_3d):
        for i in range(b):
            if not cp.array_equal(labels_2d[i], labels_3d[i]):
                print(f"Difference found in slice {i}")
                print(f" - Max label in 2D slice {i}: {labels_2d[i].max()}, Max label in 3D slice {i}: {labels_3d[i].max()}")
    print()
```

### Installation

### Environment

```
OS                           : Linux-6.12.70-1-lts-x86_64-with-glibc2.43
Python Version               : 3.13.5
NumPy Version                : 2.1.2
SciPy Version                : 1.16.0
Cython Build Version         : 3.1.4
NVCC PATH                    : None
CUDA Build Version           : 13000
CUDA Driver Version          : 13010
CUDA Runtime Version         : 13000 (linked to CuPy) / 12080 (locally installed)
CUDA Runtime Path            : /home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/cuda_runtime/lib/libcudart.so.12 (locally installed)
CUDA Extra Include Dirs      : ['/home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/cuda_runtime/include']
cuBLAS Version               : ImportError('libcublas.so.13: cannot open shared object file: No such file or directory')
cuBLAS Path                  : /home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/cublas/lib/libcublas.so.12
cuFFT Version                : None
cuFFT Path                   : None
cuRAND Path                  : /home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/curand/lib/libcurand.so.10
cuSOLVER Version             : ImportError('libcusolver.so.12: cannot open shared object file: No such file or directory')
cuSOLVER Path                : /home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/cusolver/lib/libcusolver.so.11
cuSPARSE Path                : /home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/cusparse/lib/libcusparse.so.12
NVRTC Path                   : /home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/cuda_nvrtc/lib/libnvrtc.so.12
Jitify Build Version         : None
NCCL Build Version           : 22707
NCCL Path                    : /home/user/miniconda3/envs/myenv/lib/python3.13/site-packages/nvidia/nccl/lib/libnccl.so.2
Device 0 Name                : NVIDIA RTX A6000
Device 0 PCI Bus ID          : 0000:E6:00.0
Device 1 Name                : Quadro RTX 5000
Device 1 Compute Capability  : 75
Device 1 PCI Bus ID          : 0000:D5:00.0
```

### Additional Information

_No response_

jeremyfirst22:
## Summary

Fixes an index overflow bug in `cupyx.scipy.ndimage.label` that leads to incorrect results or CUDA illegal memory access errors for large images.

### Root cause

The union-find implementation in `_kernel_connect` and `_kernel_count` uses 32-bit integers (`int`) for indexing flattened array positions. For sufficiently large arrays, these indices overflow, leading to incorrect connectivity, silent mislabeling, or GPU memory access errors.

## What changed

- Promote indexing variables (`j`, `k`, `rest`, `stride`, `pos`) to 64-bit (`long long`)
- Store intermediate label array `y` as `int64` to support large flattened indices
- Upgrade 'local' array to int64 (stores flat indices of roots)
- Use explicit 64-bit casting for `atomicCAS`, since it only supports `unsigned long long` for 64-bit (link: [CUDA docs](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=atomicCAS#atomiccas)):
  ```cpp
  atomicCAS((unsigned long long*)&y[k],
            (unsigned long long)k,
            (unsigned long long)j);

## Validation
The following test constructs a tiled 3D sphere packing where each tile should become a single connected component.
```python
#!/usr/bin/env python3

import time
import cupy as cp
import cupyx.scipy.ndimage as ndi
import numpy as np


INT32_MAX = 2**31 - 1  # 2,147,483,647


def _make_sphere_cell(cell_size, radius):
    c = cell_size // 2
    ix, iy, iz = np.mgrid[:cell_size, :cell_size, :cell_size]
    return (ix - c) ** 2 + (iy - c) ** 2 + (iz - c) ** 2 <= radius**2


def make_sphere_pack(tile_counts, cell_size=10, radius=4):
    cell = _make_sphere_cell(cell_size, radius)
    return cp.tile(cp.asarray(cell), tile_counts)


def run_test(tile_counts):
    cell_size = 10
    expected = tile_counts[0] * tile_counts[1] * tile_counts[2]

    img = make_sphere_pack(tile_counts, cell_size=cell_size)
    print(img.shape)

    t0 = time.perf_counter()
    _, n_labels = ndi.label(img)
    cp.cuda.Device().synchronize()
    dt = time.perf_counter() - t0

    status = "PASS" if int(n_labels) == expected else "FAIL"
    print(
        f"  expected={expected:,}  got={int(n_labels):,}  " f"[{status}]  ({dt:.1f}s)"
    )


SIZES = [
    ((100, 100, 100), "1.0e9 voxels — below int32 overflow threshold (baseline)"),
    ((130, 130, 130), "2.197e9 voxels — exceeds int32 max (overflow case)"),
]


print(f"CuPy {cp.__version__}")
print(f"GPU: {cp.cuda.runtime.getDeviceProperties(0)['name'].decode()}\n")

for tile_counts, desc in SIZES:
    print(f"[{desc}]")
    run_test(tile_counts)
```
### Before fix
```text
CuPy 15.0.0a1
GPU: NVIDIA H100 80GB HBM3

[1.0e9 voxels — below int32 overflow threshold (baseline)]
(1000, 1000, 1000)
  expected=1,000,000  got=1,000,000  [PASS]  (0.1s)
[2.197e9 voxels — exceeds int32 max (overflow case)]
(1300, 1300, 1300)
  expected=2,197,000  got=2,146,300  [FAIL]  (0.1s)
```

### After fix
```text
CuPy 15.0.0a1
GPU: NVIDIA H100 80GB HBM3

[1.0e9 voxels — below int32 overflow threshold (baseline)]
(1000, 1000, 1000)
  expected=1,000,000  got=1,000,000  [PASS]  (0.5s)
[2.197e9 voxels — exceeds int32 max (overflow case)]
(1300, 1300, 1300)
  expected=2,197,000  got=2,197,000  [PASS]  (0.4s)
```

## Related issue

Fixes: #9851

### DIFF
--- a/cupyx/scipy/ndimage/_measurements.py
+++ b/cupyx/scipy/ndimage/_measurements.py
@@ -45,9 +45,14 @@ def label(input, structure=None, output=None):
     structure = numpy.array(structure, dtype=bool)
     if structure.ndim != input.ndim:
         raise RuntimeError('structure and input must have equal rank')
-    for i in structure.shape:
-        if i != 3:
+    for ii in structure.shape:
+        if ii != 3:
             raise ValueError('structure dimensions must be equal to 3')
+
+    # Use 32 bits if it's large enough for this image.
+    # _ni_label.label() needs two entries for background and
+    # foreground tracking
+    need_64bits = input.size >= (2**31 - 2)
 
     if isinstance(output, cupy.ndarray):
         if output.shape != input.shape:
@@ -56,30 +61,40 @@ def label(input, structure=None, output=None):
     else:
         caller_provided_output = False
         if output is None:
-            output = cupy.empty(input.shape, numpy.int32)
+            output = cupy.empty(input.shape, cupy.intp if need_64bits else cupy.int32)
         else:
+            if output not in (cupy.int32, cupy.int64):
+                # Guard against invalid atomicAdd calls
+                raise TypeError("output dtype must be int32 or int64")
             output = cupy.empty(input.shape, output)
 
-    if input.size == 0:
-        # empty
-        maxlabel = 0
-    elif input.ndim == 0:
-        # 0-dim array
-        maxlabel = 0 if input.item() == 0 else 1
-        output.fill(maxlabel)
-    else:
-        if output.dtype != numpy.int32:
-            y = cupy.empty(input.shape, numpy.int32)
+    # handle scalars, 0-D arrays
+    if input.ndim == 0 or input.size == 0:
+        if input.ndim == 0:
+            # scalar
+            maxlabel = 1 if (input != 0) else 0
+            output.fill(maxlabel)
         else:
-            y = output
-        maxlabel = _label(input, structure, y)
-        if output.dtype != numpy.int32:
-            _core.elementwise_copy(y, output)
+            # 0-D
+            maxlabel = 0
+        if caller_provided_output:
+            return maxlabel
+        else:
+            return output, maxlabel
+
+    if not (need_64bits and output.dtype != cupy.int64):
+        max_label = _label(input, structure, output)
+    else:
+        # Allocate temporary 64-bit array, then copy results to output buffer
+        tmp_output = cupy.empty(input.shape, cupy.int64 if need_64bits else cupy.int32)
+        max_label = _label(input, structure, tmp_output, max_label=2**31 - 1)
+        _core.elementwise_copy(tmp_output, output)
 
     if caller_provided_output:
-        return maxlabel
+        # result was written in-place
+        return max_label
     else:
-        return output, maxlabel
+        return output, max_label
 
 
 def _generate_binary_structure(rank, connectivity):
@@ -92,7 +107,7 @@ def _generate_binary_structure(rank, connectivity):
     return output <= connectivity
 
 
-def _label(x, structure, y):
+def _label(x, structure, y, max_label=None):
     elems = numpy.where(structure != 0)
     vecs = [elems[dm] - 1 for dm in range(x.ndim)]
     offset = vecs[0]
@@ -103,15 +118,19 @@ def _label(x, structure, y):
     dirs = cupy.array(dirs, dtype=numpy.int32)
     ndirs = indxs.shape[0]
     y_shape = cupy.array(y.shape, dtype=numpy.int32)
-    count = cupy.zeros(2, dtype=numpy.int32)
+    count = cupy.zeros(2, dtype=y.dtype)
+
     _kernel_init()(x, y)
-    _kernel_connect()(y_shape, dirs, ndirs, x.ndim, y, size=y.size)
+    atomic_t = "unsigned long long" if y.dtype == cupy.int64 else "unsigned int"
+    _kernel_connect(atomic_t)(y_shape, dirs, ndirs, x.ndim, y, size=y.size)
     _kernel_count()(y, count, size=y.size)
-    maxlabel = int(count[0])
-    labels = cupy.empty(maxlabel, dtype=numpy.int32)
+    nlabels = int(count[0])
+    if max_label and nlabels > max_label:
+        raise RuntimeError("insufficient bit-depth in requested output type")
+    labels = cupy.empty(nlabels, dtype=y.dtype)
     _kernel_labels()(y, count, labels, size=y.size)
-    _kernel_finalize()(maxlabel, cupy.sort(labels), y, size=y.size)
-    return maxlabel
+    _kernel_finalize()(nlabels, cupy.sort(labels), y, size=y.size)
+    return nlabels
 
 
 def _kernel_init():
@@ -120,19 +139,20 @@ def _kernel_init():
         'cupyx_scipy_ndimage_label_init')
 
 
-def _kernel_connect():
+def _kernel_connect(atomic_t):
     return _core.ElementwiseKernel(
         'raw int32 shape, raw int32 dirs, int32 ndirs, int32 ndim',
         'raw Y y',
+        f"typedef {atomic_t} atomic_t;"
         '''
         if (y[i] < 0) continue;
         for (int dr = 0; dr < ndirs; dr++) {
-            int j = i;
-            int rest = j;
-            int stride = 1;
-            int k = 0;
-            for (int dm = ndim-1; dm >= 0; dm--) {
-                int pos = rest % shape[dm] + dirs[dm + dr * ndim];
+            Y j = i;
+            Y rest = j;
+            Y stride = 1;
+            Y k = 0;
+            for (int dm = ndim - 1; dm >= 0; dm--) {
+                Y pos = rest % shape[dm] + dirs[dm + dr * ndim];
                 if (pos < 0 || pos >= shape[dm]) {
                     k = -1;
                     break;
@@ -143,17 +163,18 @@ def _kernel_connect():
             }
             if (k < 0) continue;
             if (y[k] < 0) continue;
+
             while (1) {
                 while (j != y[j]) { j = y[j]; }
                 while (k != y[k]) { k = y[k]; }
                 if (j == k) break;
                 if (j < k) {
-                    int old = atomicCAS( &y[k], k, j );
+                    Y old = atomicCAS((atomic_t*)&y[k], k, j);
                     if (old == k) break;
                     k = old;
                 }
                 else {
-                    int old = atomicCAS( &y[j], j, k );
+                    Y old = atomicCAS((atomic_t*)&y[j], j, k);
                     if (old == j) break;
                     j = old;
                 }
@@ -165,10 +186,10 @@ def _kernel_connect():
 
 def _kernel_count():
     return _core.ElementwiseKernel(
-        '', 'raw Y y, raw int32 count',
+        '', 'raw Y y, raw Y count',
         '''
         if (y[i] < 0) continue;
-        int j = i;
+        Y j = i;
         while (j != y[j]) { j = y[j]; }
         if (j != i) y[i] = j;
         else atomicAdd(&count[0], 1);
@@ -178,10 +199,10 @@ def _kernel_count():
 
 def _kernel_labels():
     return _core.ElementwiseKernel(
-        '', 'raw Y y, raw int32 count, raw int32 labels',
+        '', 'raw Y y, raw Y count, raw Y labels',
         '''
         if (y[i] != i) continue;
-        int j = atomicAdd(&count[1], 1);
+        Y j = atomicAdd(&count[1], 1);
         labels[j] = i;
         ''',
         'cupyx_scipy_ndimage_label_labels')
@@ -189,16 +210,16 @@ def _kernel_labels():
 
 def _kernel_finalize():
     return _core.ElementwiseKernel(
-        'int32 maxlabel', 'raw int32 labels, raw Y y',
+        'Y maxlabel', 'raw Y labels, raw Y y',
         '''
         if (y[i] < 0) {
             y[i] = 0;
             continue;
         }
-        int yi = y[i];
-        int j_min = 0;
-        int j_max = maxlabel - 1;
-        int j = (j_min + j_max) / 2;
+        Y yi = y[i];
+        Y j_min = 0;
+        Y j_max = maxlabel - 1;
+        Y j = (j_min + j_max) / 2;
         while (j_min < j_max) {
             if (yi == labels[j]) break;
             if (yi < labels[j]) j_max = j - 1;
--- a/cupyx/scipy/ndimage/_measurements.py
+++ b/cupyx/scipy/ndimage/_measurements.py
@@ -61,7 +61,9 @@ def label(input, structure=None, output=None):
     else:
         caller_provided_output = False
         if output is None:
-            output = cupy.empty(input.shape, cupy.intp if need_64bits else cupy.int32)
+            output = cupy.empty(
+                input.shape, cupy.intp if need_64bits else cupy.int32
+            )
         else:
             if output not in (cupy.int32, cupy.int64):
                 # Guard against invalid atomicAdd calls
@@ -86,7 +88,9 @@ def label(input, structure=None, output=None):
         max_label = _label(input, structure, output)
     else:
         # Allocate temporary 64-bit array, then copy results to output buffer
-        tmp_output = cupy.empty(input.shape, cupy.int64 if need_64bits else cupy.int32)
+        tmp_output = cupy.empty(
+            input.shape, cupy.int64 if need_64bits else cupy.int32
+        )
         max_label = _label(input, structure, tmp_output, max_label=2**31 - 1)
         _core.elementwise_copy(tmp_output, output)
 
@@ -121,7 +125,9 @@ def _label(x, structure, y, max_label=None):
     count = cupy.zeros(2, dtype=y.dtype)
 
     _kernel_init()(x, y)
-    atomic_t = "unsigned long long" if y.dtype == cupy.int64 else "unsigned int"
+    atomic_t = (
+        "unsigned long long" if y.dtype == cupy.int64 else "unsigned int"
+    )
     _kernel_connect(atomic_t)(y_shape, dirs, ndirs, x.ndim, y, size=y.size)
     _kernel_count()(y, count, size=y.size)
     nlabels = int(count[0])
--- a/cupyx/scipy/ndimage/_measurements.py
+++ b/cupyx/scipy/ndimage/_measurements.py
@@ -65,10 +65,9 @@ def label(input, structure=None, output=None):
                 input.shape, cupy.intp if need_64bits else cupy.int32
             )
         else:
-            if output not in (cupy.int32, cupy.int64):
-                # Guard against invalid atomicAdd calls
-                raise TypeError("output dtype must be int32 or int64")
             output = cupy.empty(input.shape, output)
+    if output.dtype.kind not in 'iu':
+        raise TypeError("output dtype must be an integer type")
 
     # handle scalars, 0-D arrays
     if input.ndim == 0 or input.size == 0:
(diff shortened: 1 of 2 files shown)


cupyx/scipy/ndimage/_measurements.py
```diff
@@ -189,13 +196,13 @@ def _kernel_labels():
 
 def _kernel_finalize():
     return _core.ElementwiseKernel(
-        'int32 maxlabel', 'raw int32 labels, raw Y y',
+        'int32 maxlabel', 'raw int64 labels, raw Y y',
```
seberg: The `maxlabel` can now also clearly be `int64`, I think.  (but I think we can parametrize it to the same type as `labels`.

cupyx/scipy/ndimage/_measurements.py
seberg: I expect the count could now also overflow and should be the same dtype as `y`?

seberg, line 145: Just as a note, this is all fine.  The `shape` could in theory overflow, but in practice it isn't plausible and should error out earlier.

cupyx/scipy/ndimage/_measurements.py
```diff
@@ -148,12 +149,18 @@ def _kernel_connect():
                 while (k != y[k]) { k = y[k]; }
                 if (j == k) break;
                 if (j < k) {
-                    int old = atomicCAS( &y[k], k, j );
+                    long long old = (long long)atomicCAS(
```
seberg: Hmmm, I guess if the type of `Y` can be `long long` we need this cast :/.

Maybe there is a better work-around, but worst case, maybe we can do a `using atomic_t = std::is_same_v<t, long long> ? unsigned long long : t;`  (untested)...

cupyx/scipy/ndimage/_measurements.py
```diff
@@ -127,12 +128,12 @@ def _kernel_connect():
...
-            int j = i;
-            int rest = j;
-            int stride = 1;
-            int k = 0;
-            for (int dm = ndim-1; dm >= 0; dm--) {
-                int pos = rest % shape[dm] + dirs[dm + dr * ndim];
+            long long j = i;
+            long long rest = j;
+            long long stride = 1;
+            long long k = 0;
```
seberg: It's a bit funny, but I think all of these are just `Y`?  (Because I think `Y` needs to be either int32 or int64)

(Technically, `decltype(y)::index_t` or so may also be right, but I am not 100% sure and `Y` is more conservative.)

seberg (commented):
Thanks for looking into this! a few smaller comments that hopefully help. The main things however are:
1. We should adopt the SciPy to decide between int32 and int64 here.
2. This really needs some tests (probably marked as `slow`) that are just large enough to hit the size issue (and maybe also the count/number of overflows if you use an all zero structured element, all pixels get a different label, I think).

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: Agreed, and fixed. `maxlabel`, `labels`, and `y` all parameterized to same type now.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: Agreed. It is only used in `atomicAdd` operations, so I do not think we need type casting (unless we are required to support older (<6.0) CUDA versions?). Either way, `count` can be made unsigned if we need to force `unsigned long long` for `atomicAdd`.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: I tried this, but we do not have access to `std::is_same_v` in an `ElementwiseKernel`, and I figured it would be bad practice to include the appropriate header. 

The cleanest thing I could come up with is determine the appropriate type on the Python side, and then inject the typedef with string formatting on the kernel definition string.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: Really good point, thank you, and simplifies everything a bit. Fixed, and now flat indices are parameterized the same as `y`

jeremyfirst22:
@seberg I've updated the commit to address your comments. Overall I think is simpler now, and directly mimics scipy's implementation, so thank you for the push and the comments. Let me know what you think. 

> 1. We should adopt the SciPy to decide between int32 and int64 here.

I agree. I used this as an opportunity to bring the `label` driver function into alignment with the current `scipy` structure. Because of the large memory footprint of the CUDA thrust sort (which caused issues on the unit test, see below), I decided to change the try/except block so that `_label` throws a `RuntimeError` directly, before sorting, if the output dtype is insufficient. 

> 2. This really needs some tests (probably marked as `slow`) that are just large enough to hit the size issue (and maybe also the count/number of overflows if you use an all zero structured element, all pixels get a different label, I think).

I converted the snippet from the description into a pytest and added it. I test both below and above the int32 limit across output options, and test that supplying an int32 output buffer raises a `RuntimeError` if there are more than 2^31 labels, via the zero-structured test you recommended. Great idea.

cupyx/scipy/ndimage/_measurements.py
```diff
@@ -45,41 +45,60 @@ def label(input, structure=None, output=None):
...
+        if caller_provided_output:
+            return maxlabel
         else:
-            y = output
-        maxlabel = _label(input, structure, y)
-        if output.dtype != numpy.int32:
-            _core.elementwise_copy(y, output)
+            return output, maxlabel
+
+    if not (need_64bits and output.dtype != cupy.int64):
```
seberg:
```suggestion
    if not need_64bits or output.dtype == cupy.int64:
```
seems clearer to me.

cupyx/scipy/ndimage/_measurements.py
```diff
@@ -45,41 +45,60 @@ def label(input, structure=None, output=None):
...
+        max_label = _label(input, structure, output)
+    else:
+        # Allocate temporary 64-bit array, then copy results to output buffer
+        tmp_output = cupy.empty(
+            input.shape, cupy.int64 if need_64bits else cupy.int32
```
seberg: After rewriting the above `if`, it is clear that the comment is right and `need_64bits` is never true.

cupyx/scipy/ndimage/_measurements.py
```diff
@@ -45,41 +45,60 @@ def label(input, structure=None, output=None):
...
     if isinstance(output, cupy.ndarray):
         if output.shape != input.shape:
             raise ValueError("output shape not correct")
         caller_provided_output = True
     else:
         caller_provided_output = False
         if output is None:
-            output = cupy.empty(input.shape, numpy.int32)
+            output = cupy.empty(
+                input.shape, cupy.intp if need_64bits else cupy.int32
```
seberg:
```suggestion
                input.shape, cupy.int64 if need_64bits else cupy.int32
```
just a bit clearer (it's the same thing on all systems we support).

(But if this is just stolen from SciPy, let's keep it)

cupyx/scipy/ndimage/_measurements.py
```diff
@@ -45,41 +45,60 @@ def label(input, structure=None, output=None):
...
+            )
         else:
+            if output not in (cupy.int32, cupy.int64):
+                # Guard against invalid atomicAdd calls
+                raise TypeError("output dtype must be int32 or int64")
```
seberg, line 70: I suppose this isn't a new restriction or at least shared by SciPy?  Because if it is, we may just want to make sure to make that output copy below (or support it).

cupyx/scipy/ndimage/_measurements.py
seberg: Ah, yeah, I think I would slightly prefer this:
```
using atomic_t = typename cupy::type_traits::conditional<
    cupy::type_traits::is_same<Y, long long>::value, unsigned long long, Y>::type;
```
I.e. there are some custom definitions for this in `carray.cuh` which are available (although a few more might be nice to have...)

It certainly isn't prettier, but I do slightly prefer to have this as part of the kernel and not generate two different ones.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
```diff
@@ -672,3 +672,55 @@ def test_find_objects_non_contiguous(self, xp, scp, expansion_factor,
...
+        s = self._CELL_SIZE
+        c = s // 2
+        r = self._RADIUS
+        ix, iy, iz = numpy.mgrid[:s, :s, :s]
+        cell = (ix - c) ** 2 + (iy - c) ** 2 + (iz - c) ** 2 <= r ** 2
+        return cupy.tile(cupy.asarray(cell), tile_counts)
+
+    @pytest.mark.parametrize('tile_counts', [
+        ((100, 100, 100)),  # ~1e9 voxels — below int32 limit
+        ((130, 130, 130)),  # ~2.197e9 voxels — above int32 limit
```
seberg: This should be `(1300, 1300, 1300)` no?

seberg (commented):
Generally looks good now, thanks.  Mainly a few nitpicks and I think I prefer pushing the `atomic_t` in some form into the kernel (even if it might be a bit uglier code wise).

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
seberg: That said... This will take a _lot_ of memory if things are OK in CI then maybe it is fine, but there is a chance we need:
```
try:
   ...
except MemoryError:
   pytest.skip("OOM")
```
eventually.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: Agreed, this would be preferred, but this is stolen direcly from SciPy [here](https://github.com/scipy/scipy/blob/main/scipy/ndimage/_measurements.py#L199). Keeping as is.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: Agreed, fixed.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: Good catch. I had taken this from [here](https://github.com/scipy/scipy/blob/main/scipy/ndimage/_measurements.py#L222), but after rewriting this as a simply `if else` block, rather than a `try except` block, you are correct, we do not need conditional data typing for the temporary workspace. It should always be `int64`. Fixed.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
jeremyfirst22: This is 130x130x130 cells, not image dimension, but I can see how that is not clear at first glance. I've updated the comments to make this explicit in a table. 

I agree, this takes a lot of memory (16 GB just for the temporary array). I'm not sure what kind of hardware the CI/CD pipeline runs on, but this appears to have passed, and it passes locally on an 80 GB H100 node. I'm happy to add the OOM catcher if you think it is prudent.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: This restriction is an artifact of the CUDA kernels. If `output` is `int8` or `int16`, the kernel fails since there is no CUDA support for `unsigned char` or `unsigned short` in `atomicAdd`. The same thing applies for the `float` dtypes.

We could in principle support `float32` and `float64` output types, but it will require some complicated type conversions and memory alignment. This guard is the simplest solution.

cupyx/scipy/ndimage/_measurements.py
jeremyfirst22: I agree, the string injection was hacky at best. I did not know about these carray.cuh internals, so thank you. Fixed.

jeremyfirst22:
> Generally looks good now, thanks. Mainly a few nitpicks and I think I prefer pushing the `atomic_t` in some form into the kernel (even if it might be a bit uglier code wise).

@seberg I've updated the PR with your suggestions and responded to each. Thank you for your time to review. Let me know what you think.

tests/cupyx_tests/scipy_tests/ndimage_tests/test_measurements.py
seberg: Ah, thanks.  CI is running with a T4, so 16GiB, so it should be out of luck (but somehow though it had already passed).
But I guess it means we may need a (large) try/except.

cupyx/scipy/ndimage/_measurements.py
seberg: Yeah, OK.  I suppose I am slightly worried about regressions, even if it wasn't actually _fast_, i.e. it seems the old code let just about anything pass (int16, float32, ...).  Presumably by just assuming assinging back to `output` will work fine.

I'll admit, I think that might just make sense.  Since we are only NumPy 2+, we could maybe just make the range check (which is neat) into:
```
try:
    output.dtype.type(max_label)
except (OverflowError, ValueError):
    ValueError("...")
```
(Should be an `OverflowError`, but dunno if hedging a bit more doesn't make sense. For floats, that won't catch all possible errors, but since SciPy doesn't seem to raise...)

seberg (commented):
Thanks for the updates, I guess this one small thing we should maybe change.  Let me run CI anyway, to see if there are other fixes needed (and if just avoiding OoM errors).

/test mini

seberg:
Sorry, I was about to merge it at least once and then remembered that this changed things a bit: previously an output of `int8` or `int16`, etc. worked just fine even if it wasn't ideal.
I restored that and changed the max-labels error to a ValueError (if slightly awkward, but OK).

I also hacked the test for `try/except MemoryError` they are not run in CI, but do pass locally on GPUs with a huge amount of memory.  (Also not ideal, but I think also acceptable)

But, with that I think it's all good, thanks a lot!

/test mini

seberg (approved):

lucascolley:
Hey @seberg, https://github.com/cupy/cupy/pull/9859/changes/0f00c5a47fa792cdfad57ea8f28284d41aa42ac5 looks suspicious on `git blame` for the failures surfacing at https://github.com/scipy/scipy/actions/runs/28133346482/job/83314587858?pr=25495.

seberg:
Hmmm, yeah.  I reverted that type narrowing for `out`/`dtype` mostly, but missed this (or thought it was OKish).
@jeremyfirst22 are  yo interested in following up?  Although, I am slightly wondering if you event want this to work, using floats as labels seems a bit strange?

lucascolley:
> Although, I am slightly wondering if you event want this to work, using floats as labels seems a bit strange?

I guess that's a discussion for a new SciPy issue... would be good to fix the regression in the meantime.

PR #9861, opened 2026-04-10
Advertise free-threading support and add linux CI run

seberg:
Advertise free-threading support in CuPy builds.

Additionally, this PR does a few things:
* Adds a py314t CI run (linux)
  * To make this test run in a reasonable time, a mechanism for heavily subsample parametrizations and `for_all_dtypes` has been added as `CUPY_TEST_RANDOM_SUBSAMPLE=1`. This does _not_ replace full test runs!
* Some final threading fixes in the test suite
* Cython is bumped to 3.2, although this is probably not vital for the free-threading (I thought this earlier).


seberg:
Well, that run was a cryptic mess (and infinitely slow)... Best hope is due to old cython, but let's see.
(One of the first, or the first failiure, was a dict that didn't got invalidated or so.  And while that code looks not ideal, it doesn't look all that wrong.)

/test linux,cuda130

seberg:
Hmmm, the slowness got me, but this time around it didn't look super slow.  Either way, one fusion fix (I find it strange this didn't show up earlier ever) and `cutensor` was never thread-safe (had only one handle per device), but I guess I never ran that locally!

Let's try again:
/test linux,cuda130

seberg:
Hmmmmpf, timed out after 5 hours on second try.  I guess we need to find some thought to extremely reduce the amount of tests run :/.

cupy/_core/_fusion_interface.py
```diff
@@ -23,15 +23,18 @@ def _set_dtype_to_astype_dict():
     This function is called at most once.
     """
     global _dtype_to_astype_dict
-    _dtype_to_astype_dict = {}
+    dtype_to_astype_dict = {}
```
leofang, line 26: We should at least check this if we don't want to use any lock: If it's not None, return early (already populated by another thread).

leofang:
Yeah unclear to me what's going on. Is it because the kernel cache is not populated, or because the runner has limited cores, or because the 314t build is inherited slower (and we only turn on 2 processes), or something else...? 🤔

seberg:
I'll run it once more to confirm (and should try locally), but the build is fine ~17 minutes and I don't see the cache being off either really.

I am suprised about at large slowdown, but need to time it also locally to get a feel here.  `pytest-run-parallel` does e.g. spawn new threads for every single test so a more than 2x slowdown isn't surprising.
Maybe the fact that we have quite a bit of thread-local state makes things even worse?

Either way, my guess right now is that this would need it's own CI run maybe and then slashing down parametrization a lot, maybe?
(I don't have a much better thought on how to run fewer tests but CuPy is ridiculously parametrize with the `for_all_dtypes` loops.)

/test linux,cuda130

EDIT: Ack, silly me had `/test mini`, I guess too late now :(.

seberg:
Yeah, definitely need to cut the number of tests run down for this on CI :(, unless I am missing something bigger...

(The cutensor stuff also needs more massaging then I hoped; I'll split out a new PR, I think...)

seberg:
OK, the last run is a bit lobotomized (only two threads in parallel) and introduces random subsampling...
... but, I think this shows that we can set up basic CI for free-threading.

(I will try to figure out if there is a specific reason here for the slowness, one theory is that spawning threads is just slow with cuda involved, but only profiling will tell.)

seberg:
Let's see found two harmless (or optional tests) that needed skipping.  But one test might have disabled the normal cub reduction path for the full run.  And that might have caused the generic cub reduce path to be used, which seems to compile a lot more.

(I think one level of caching is failing there, and should be fixed. But I am not sure yet why it looks like a lot of re-compilation as that has a second layer of caching...)

/test linux,cuda130

seberg:
/test linux,cuda130

EDIT: For reference, the previous run took 1h:26m 
EDIT2: New time 1h:18m so, honestly, just within the range of random fluctuation anyway.

seberg:
(should include the FT run now)

/test mini

seberg:
Re-running once, it was very slow, but I expect it is just because it's not sharing the cache (which I guess it could in principle).

/test cupy.linux.cuda13x-py314t

seberg:
Ooops, should use the tag, not the full name:

/test py314t

EDIT: Hmmmm, still a bit on the slow side at 1h52m

seberg:
I bumped subsampling a bit as a hot-fix... would be nice to get close to 1h:30m, I think...
(since the GCP cache should be shared I don't think it can be a caching issue really.)

/test py314t

EDIT: OK  test succeeded in 1h24m1.588791291s should be good enough, IMO.

seberg:
OK, so we figured out the speed curve here (or most likely cause).  `cuda131` isn't in the `mini` matrix anymore, so this job may actually have had it's own cache.
Since it doesn't run the identical tests every time, it might be that even the second run still takes close to two hours.

But I suspect that is OK, since also for normal jobs we need 2 runs for things to settle.  (We could avoid this by just using cuda132 for the new job.

tests/conftest.py
```diff
@@ -100,3 +102,44 @@ class my_allocator(metaclass=numpy_allocator.type):
...
+        # (and maybe to be picked up the `for_all_dtypes`).
+        import secrets
+
+        seed = secrets.randbits(64)
+        os.environ['_CUPY_TEST_SUBSAMPLE_SEED'] = str(seed)
+
+    rng = random.Random(seed)
+
+    def pytest_collection_modifyitems(session, config, items):
+        original_len = len(items)
```
seberg, line 120:
```suggestion
        original_len = len(items)
        if original_len == 0:
            return
```
I had my bot have a look, and it said `/ original_length` could otherwise fail.  Doesn't really matter, but we could apply this.

.pfnci/schema.yaml
```diff
@@ -344,7 +349,7 @@ mpi4py:
       - "3.14"
 cython:
   "3.1":
```
leofang: Is the Cython 3.1→3.2 bump specifically required for free-threading support (e.g. `freethreading_compatible` directive or FT-related codegen fixes only available in 3.2)? If so, it'd be helpful to note that in the PR description since this change touches every CI Dockerfile.

Also — the key here is still `"3.1"` but the spec is now `"==3.2.*"`. Should the key be updated to `"3.2"` as well, or is it intentionally left as `"3.1"` for matrix backward compatibility?

tests/conftest.py
```diff
@@ -100,3 +102,44 @@ class my_allocator(metaclass=numpy_allocator.type):
...
+        # (This also overrides the global pyproject setting.)
+        # TODO(seberg): Could audit strict xfail to remove this.
+        for item in items:
+            for node in item.listchain():
+                for i, marker in enumerate(node.own_markers):
+                    if marker.name == 'xfail':
+                        kwargs = dict(marker.kwargs)
+                        kwargs['strict'] = False
+                        node.own_markers[i] = pytest.mark.xfail(
+                            *marker.args, **kwargs).mark
```
leofang, line 148: This blanket `strict=False` override means that if a previously-failing parametrization starts passing (e.g. bug fixed upstream), CI won't flag it — the xfail will silently become a no-op instead of erroring as `XPASS`.

I see the TODO — could this be scoped more narrowly? For example, only relax `strict` on tests whose parametrization set was actually reduced by subsampling, rather than all tests unconditionally?

.pfnci/linux/tests/cuda13x-py314t.Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ENV LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/libcusparseLt/13:${LD_LIBRARY_PATH}
+RUN git clone https://github.com/pyenv/pyenv.git /opt/pyenv
+ENV PYENV_ROOT "/opt/pyenv"
+ENV PATH "${PYENV_ROOT}/shims:${PYENV_ROOT}/bin:${PATH}"
+RUN pyenv install 3.14t && \
+    pyenv global 3.14t && \
+    pip install -U setuptools pip wheel && \
+    pip install -U google-cloud-storage
+
+RUN pip install -U 'numpy==2.3.*' 'scipy==1.16.*' 'optuna==4.*' 'ml_dtypes==0.5.*' 'cython==3.2.*'
```
leofang: This pins `numpy==2.3.*` while `cuda130`/`cuda132` use `numpy==2.4.*`. For a free-threading CI job, wouldn't it be better to use the latest numpy (2.4) which likely has more FT fixes?

tests/conftest.py
leofang, line 120: +1 to @seberg's self-review suggestion — `original_len` can be 0, which would cause a `ZeroDivisionError` in the print statement below.

leofang (commented):
Nice work getting free-threading CI up and running for CuPy — the thread-safety fixes (fitpack `_res` → instance attr, cub_reduction tuple removal, test_filters avoiding `self` mutation) all look correct, and the subsampling approach is well-designed (seeded RNG, xdist-compatible).

A few inline comments on:
- Whether the Cython 3.1→3.2 bump is FT-motivated (and the schema key name mismatch)
- The blanket `xfail strict=False` override scope
- numpy version choice for the FT Dockerfile
- The `original_len == 0` guard

Also worth updating the PR description — it still says "just a draft to experiment" but this has evolved into a solid, multi-faceted change.

tests/conftest.py
seberg: > set was actually reduced by subsampling, rather than all tests unconditionally?

The problem is that it probably won't be obvious generally (unless we explicitly mark it).  Since `for_all_dtypes()` can be used as a loop within a test even.

cupyx/scipy/interpolate/_fitpack_repro.py
```diff
@@ -986,9 +986,9 @@ def make_splrep(x, y, *, w=None, xb=None, xe=None, k=3, s=0, t=None,
     if s == 0:
```
leofang, line 987: Can't comment on the above lines, but this seems like a nice fix too? Why did we set a function attribute `make_splrep._res`? 😱 We should update the inline comments.

.pfnci/schema.yaml
seberg: Honestly... If that seems fine I'll just bump 3.1 -> 3.2 (and fix it here).  But, while I thought that was needed at some point, I am not sure.  3.2 presumably has fixes, but I doubt we'll notice them.

install/cupy_builder/_command.py
```diff
@@ -115,6 +115,7 @@ def _cythonize(self, nthreads: int) -> None:
             # https://github.com/cupy/cupy/pull/8457#issuecomment-2656568499
             'binding': False,
             'warn.deprecated.IF': False,   # suppress IF deprecation warning
+            'freethreading_compatible': True,
```
leofang, line 118: Note: This is a no-op if the host Python is not FT-enabled.

tests/conftest.py
seberg: I would suggest that we make this a follow-up issue if we deem it relevant enough.  Because IIRC, it was quite unwieldy.

Although, the way I tried to do it was to actually parametrize things with `pytest.param(..., marks=xfail())` for the individual parameters that _actually_ fail.  Which is just quite a bit of refactor.

.pfnci/schema.yaml
leofang: I am not sure how sensitive our CPU overhead is, with regard to Cython versions. I would recommend running a few quick tests to confirm. Otherwise, LGTM!

seberg: Trying around with a few simple ops, I think it's basically the same (there are some very small fluctuations but they seem mostly also to happen if I just start a fresh session).

But hmmmm, not actually sure what to put as dependency now `>3.2` or the full range `3.2-3.3` (but in that case, it would make sense to keep both in CI).

(Will have to update cupy-release-tool as well, but that is simple.)

leofang (approved):
LGTM!

PR #9862, opened 2026-04-10
sparse: work around cuSPARSE SpMM gridDim.y overflow (#9850)

### LINKED ISSUE #9850: [Bug] Sparse-dense matmul returns zero for F-contiguous dense input (float64)
# [Bug] Sparse-dense matmul returns zero for F-contiguous dense input (float64)

I've found a silent failure in Cupy's sparse-dense multiplication. When multiplying a large F-contiguous dense matrix with a CSR matrix in float64, the result is exactly zero, even though both inputs are non-zero.

### Environment
- Cupy: 14.0.1 (cupy-cuda12x)
- CUDA: 12.x
- GPU: RTX 4090
- OS: Linux

### Reproduction
This script reproduces the bug. It uses a dense matrix of shape (1380300, 377) and a 377x377 CSR matrix.

```python
import cupy as cp
import cupyx.scipy.sparse as cpx_sparse
import numpy as np
import scipy.sparse as sp

def reproduce():
    n_grid = 1380300
    n_ao = 377
    sparsity = 0.05
    
    # Sparse CSR (float64)
    dm = sp.random(n_ao, n_ao, density=sparsity, format='csr', dtype=np.float64)
    dm_gpu = cpx_sparse.csr_matrix(dm)
    
    # Dense F-contiguous (float64)
    ao_gpu = cp.random.rand(n_grid, n_ao, dtype=np.float64)
    ao_gpu_F = cp.asfortranarray(ao_gpu)
    
    print(f"DM sum: {cp.sum(cp.abs(dm_gpu.data)):.6e}")
    print(f"AO sum: {cp.sum(cp.abs(ao_gpu_F)):.6e}")

    # This returns exactly 0.0
    res = ao_gpu_F @ dm_gpu
    print(f"Result sum: {cp.sum(cp.abs(res)):.6e}")
    
    assert cp.sum(cp.abs(res)) > 0, \"Error: matmul returned zero\"

if __name__ == \"__main__\":
    reproduce()
```

### Observations
- `Result sum: 0.000000e+00`

eriknw:
Fixes #9850.

cuSPARSE csrmm_alg1 sets gridDim.y = ceil(n/16) where n is the number of columns in the dense output. When n > 65535*16 the grid dimension exceeds the hardware limit and the kernel silently produces zeros. Chunk the dense matrix along columns on CUDA < 13.1 where the bug is present.

I've confirmed the cuSPARSE fix is present on a system with `_runtime.runtimeGetVersion()` of 13020, but I'm unsure in which exact version it became available.

### DIFF
--- a/cupyx/cusparse.py
+++ b/cupyx/cusparse.py
@@ -1439,9 +1439,7 @@ def spmv(a, x, y=None, alpha=1, beta=0, transa=False):
 
 # cuSPARSE csrmm_alg1 gridDim.y overflow threshold (#9850).
 # The kernel sets gridDim.y = ceil(n / 16); hardware max is 65535.
-# Fixed in CUDA 13.2+; affects all CUDA 12.x. Applied unconditionally
-# since the chunking overhead is negligible and there is no reliable
-# way to detect the fix at runtime (cusparseGetVersion doesn't change).
+# Fixed in cuSPARSE 12.7.20+ (CUDA 13.2+).
 _SPMM_MAX_DENSE_COLS = 65535 * 16  # 1,048,560
 
 
@@ -1499,7 +1497,7 @@ def spmm(a, b, c=None, alpha=1, beta=0, transa=False, transb=False):
         c.fill(0)
         return c
 
-    if n > _SPMM_MAX_DENSE_COLS:
+    if n > _SPMM_MAX_DENSE_COLS and getVersion() < 12720:
         # Workaround for cuSPARSE csrmm_alg1 gridDim.y overflow (#9850)
         for col0 in range(0, n, _SPMM_MAX_DENSE_COLS):
             col1 = min(col0 + _SPMM_MAX_DENSE_COLS, n)
--- a/cupyx/cusparse.py
+++ b/cupyx/cusparse.py
@@ -1437,6 +1437,11 @@ def spmv(a, x, y=None, alpha=1, beta=0, transa=False):
     return y
 
 
+# cuSPARSE csrmm_alg1 gridDim.y overflow threshold on CUDA <13.1 (see #9850).
+# The kernel sets gridDim.y = ceil(n / 16); hardware max is 65535.
+_SPMM_MAX_DENSE_COLS = 65535 * 16  # 1,048,560
+
+
 def spmm(a, b, c=None, alpha=1, beta=0, transa=False, transb=False):
     """Multiplication of sparse matrix and dense matrix.
 
@@ -1489,6 +1494,19 @@ def spmm(a, b, c=None, alpha=1, beta=0, transa=False, transb=False):
         raise ValueError('dimension mismatch')
     if a.nnz == 0:
         c.fill(0)
+        return c
+
+    if n > _SPMM_MAX_DENSE_COLS and _runtime.runtimeGetVersion() < 13020:
+        # Workaround for cuSPARSE csrmm_alg1 gridDim.y overflow (see #9850)
+        for col0 in range(0, n, _SPMM_MAX_DENSE_COLS):
+            col1 = min(col0 + _SPMM_MAX_DENSE_COLS, n)
+            if transb:
+                b_chunk = _cupy.asfortranarray(b[col0:col1, :])
+            else:
+                b_chunk = b[:, col0:col1]
+            c_chunk = c[:, col0:col1]
+            spmm(a, b_chunk, c=c_chunk, alpha=alpha, beta=beta,
+                 transa=transa, transb=transb)
         return c
 
     desc_a = SpMatDescriptor.create(a)
--- a/cupyx/cusparse.py
+++ b/cupyx/cusparse.py
@@ -1437,8 +1437,11 @@ def spmv(a, x, y=None, alpha=1, beta=0, transa=False):
     return y
 
 
-# cuSPARSE csrmm_alg1 gridDim.y overflow threshold on CUDA <13.1 (see #9850).
+# cuSPARSE csrmm_alg1 gridDim.y overflow threshold (#9850).
 # The kernel sets gridDim.y = ceil(n / 16); hardware max is 65535.
+# Fixed in CUDA 13.2+; affects all CUDA 12.x. Applied unconditionally
+# since the chunking overhead is negligible and there is no reliable
+# way to detect the fix at runtime (cusparseGetVersion doesn't change).
 _SPMM_MAX_DENSE_COLS = 65535 * 16  # 1,048,560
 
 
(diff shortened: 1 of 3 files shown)


cupyx/cusparse.py
```diff
@@ -1491,6 +1496,19 @@ def spmm(a, b, c=None, alpha=1, beta=0, transa=False, transb=False):
         c.fill(0)
         return c
 
+    if n > _SPMM_MAX_DENSE_COLS and _runtime.runtimeGetVersion() < 13020:
```
leofang: Important: We need to use `cusparseGetVersion` instead of `runtimeGetVersion`, because in CuPy (and CUDA Python in general) the cudart version is a poor indicator.

leofang: Hit send too fast... The technical reason is we statically link cudart to CuPy so this is the build-time version, not run-time. 

Even if we get the run-time version via `runtimeGetLocalVersion`, it is still incorrect because technically package managers like conda and pip do not enforce a strict version lock (by design in conda and by nature in pip), so cudart and cusparse need not come from the same CTK major.minor.patch version.

cupyx/cusparse.py
```diff
@@ -1491,6 +1496,19 @@ def spmm(a, b, c=None, alpha=1, beta=0, transa=False, transb=False):
...
+        # Workaround for cuSPARSE csrmm_alg1 gridDim.y overflow (see #9850)
+        for col0 in range(0, n, _SPMM_MAX_DENSE_COLS):
+            col1 = min(col0 + _SPMM_MAX_DENSE_COLS, n)
+            if transb:
+                b_chunk = _cupy.asfortranarray(b[col0:col1, :])
```
leofang, line 1506: Q: If we call `spmm` recursively for chunking, does it matter that we use `asfortranarray` here? I would think `spmm` should figure out how to handle it?

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -2141,3 +2141,18 @@ def test_setdiag_invalid(self):
...
+    @pytest.mark.parametrize('dtype', [numpy.float32, numpy.float64])
+    def test_dense_at_sparse_large_rows(self, dtype):
+        n_grid = 1048561  # just above 65535 * 16
+        n_ao = 5
+        dm = sparse.random(n_ao, n_ao, density=0.5,
+                           format='csr', dtype=dtype)
+        ao = cupy.ones((n_grid, n_ao), dtype=dtype)
+        result = ao @ dm
+        expected = ao @ dm.toarray()
+        cupy.testing.assert_allclose(result, expected, rtol=1e-5)
```
leofang: Let's add it to two places for consistency (e.g. reusing the test utils including parametrization) 
- in this file, move this test to `test_mul_dense_matrix_large_rows`
- in the other test file for `cupyx/cusparse`, add it as a method to the [`TestSpMM`](https://github.com/cupy/cupy/blob/fce00140702ac588c5fdccd6373c1ba79257b330/tests/cupyx_tests/test_cusparse.py#L739) class

In both cases, mark the test with `@testing.slow`.

cupyx/cusparse.py
leofang: btw is `< 13020` correcT? I think it's fixed by 13.2 so it should be "<=" instead?

eriknw: Good points. I was wondering about this (I'm new here :) ), so thanks for helping out. I am able to verify the fix on a system with 13020. So, I'm not sure there's a reliable way to determine the appropriate limit at runtime, and I don't think it's worth the effort to try, because the overhead of the workaround is negligible--one extra F-contiguous column slice per ~1M columns, no data copies for the non-transb case.

cupyx/cusparse.py
eriknw: `spmm` requires F-contiguous input (`assert b.flags.f_contiguous` at the top).

Added comments to clarify this in the code.

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
eriknw: Done.

eriknw (commented):
Notes:
- **CSC path is also covered**: `csc.__mul__(dense)` calls `spmm(csc.T, b, transa=True)`, which goes through the same chunked `spmm`. Verified.
- **`csrmm2` (Legacy API) is NOT affected**: different kernel implementation. Only available on CUDA 11.x anyway.
- **`spmv` is NOT affected**: n=1 column, gridDim.y=1.

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -866,6 +866,19 @@ def test_mul_dense_matrix(self, xp, sp):
         x = xp.arange(8).reshape(4, 2).astype(self.dtype)
         return m * x
 
+    @testing.slow
+    @pytest.mark.parametrize('dtype', [numpy.float32, numpy.float64])
```
leofang: This parametrization is not needed because the class is already parametrized (please verify). Use `self.dtype` instead.

leofang: Also, we should use `@testing.numpy_cupy_allclose(sp_name='sp')` here

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -866,6 +866,19 @@ def test_mul_dense_matrix(self, xp, sp):
...
+    def test_mul_dense_matrix_large_rows(self, dtype):
+        """cuSPARSE SpMM gridDim.y overflow (#9850)."""
+        n_grid = 1048561  # just above 65535 * 16
+        n_ao = 5
+        dm = sparse.random(n_ao, n_ao, density=0.5,
+                           format='csr', dtype=dtype)
+        ao = cupy.ones((n_grid, n_ao), dtype=dtype)
+        result = ao @ dm
+        expected = ao @ dm.toarray()
+        cupy.testing.assert_allclose(result, expected, rtol=1e-5)
```
leofang: ditto, please use the allclose decorator, this will need to be re-structured a bit (see the surrounding tests)

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
```diff
@@ -2141,3 +2154,7 @@ def test_setdiag_invalid(self):
                 scipy_a.setdiag(x, k=k)
             with pytest.raises(ValueError):
                 cupyx_a.setdiag(x, k=k)
+
+
+class TestSpMMLargeColumns:
+    """Regression test for cuSPARSE SpMM gridDim.y overflow (#9850)."""
```
leofang: leftover code?

cupyx/cusparse.py
leofang: It is better to keep the WAR in-place because in the future when we drop old CUDA versions, we could remove the WAR entirely. It is simple to find out which version is functional locally by creating CUDA 13.0, 13.1, and 13.2 conda envs. Or just ask the cuSPARSE team. This API can be used to check the cuSPARSE library version:
https://github.com/cupy/cupy/blob/9263620693c892864c52d2703b862dcf1c264c6a/cupy_backends/cuda/libs/cusparse.pyx#L1554

cupyx/cusparse.py
leofang: Ah OK, it's a bit odd to me. Thanks.

leofang (changes requested):

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
leofang:
```suggestion
```

leofang:
Ah sorry, my bad, I thought it's ready for re-review and pushed a trivial change. Please feel free to overwrite...

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
eriknw: Done

tests/cupyx_tests/scipy_tests/sparse_tests/test_csr.py
eriknw: Done. Somehow I had thought I addressed all comments earlier, sorry for the delay!

leofang (approved):

PR #9205, opened 2025-06-25
[WIP] Add cumulative_sum

vin-prabhakar:
Hello,

This is a **work-in-progress** PR for implementing the `cupy.cumulative_sum` function.

### Current Progress

- The function has been ported from NumPy, following a similar implementation pattern.
- One key difference from NumPy’s implementation is how the identity value is handled:
  - In NumPy, `add.identity` is used when `include_initial=True`.
  - In CuPy, using `func.identity` raises an `AttributeError` since CuPy’s ufuncs do not have an `identity` attribute.
  - To resolve this, the identity value is passed explicitly to the `_cumulative_func` function. This also allows the same helper to be reused for `cupy.cumulative_prod`.
- Doc string written

### Pending Work

1. **Tests**  
   I need some clarification on whether the test cases for `cupy.cumulative_sum` should:
   - Be integrated into the existing test suite for `cupy.cumsum`, similar to how it's done in NumPy ([reference](https://github.com/numpy/numpy/blob/1fefc5c6b767e86c2642641e7ba7a22ab0cf554b/numpy/lib/tests/test_function_base.py#L734)), 
   OR
   - Be implemented as a separate test class.

2. **Documentation**

PR #9864, opened 2026-04-11
Add `cumulative_sum` and `cumulative_prod` (Array API)

DanielCohenn:
## Summary

Add `cupy.cumulative_sum` and `cupy.cumulative_prod` — the Array-API-compatible
counterparts of `cupy.cumsum` / `cupy.cumprod`, mirroring `numpy.cumulative_sum`
and `numpy.cumulative_prod` added in NumPy 2.0.

Both functions are implemented behind a single shared helper (`_cumulative_op`)
so they share axis handling, `out` validation, and the `include_initial` path.
This resolves the reviewer feedback on the two prior WIP PRs (#9205 and the
original prod PR) to "concentrate effort on one PR/approach, and implement both
at the same time."

## Public API

```python
cupy.cumulative_sum(x, /, *, axis=None, dtype=None, out=None, include_initial=False)
cupy.cumulative_prod(x, /, *, axis=None, dtype=None, out=None, include_initial=False)
```


DanielCohenn:
Hello! I'm a new contributor.
This PR adds `cupy.cumulative_prod`, the Array API compatible counterpart of
`cupy.cumprod` (following the same pattern as #9205 for `cumulative_sum`).
Tests and documentation will be added in a follow-up. Opening as a draft/WIP for now.

cupy/_math/sumprod.py
```diff
@@ -220,6 +220,68 @@ def nancumprod(a, axis=None, dtype=None, out=None):
...
+    x_ndim = x.ndim
+
+    if axis is None:
+        if x_ndim >= 2:
+            raise ValueError("For arrays which have more than one dimension "
+                             "``axis`` argument is required.")
+        
+        axis = 0
+    
+    if out is not None and include_initial:
```
leofang: - If `out` is not None, we need to check if its (`include_initial`-dependent) shape is compatible with the requirement
- in any case
  - if`include_initial` is true, we should just write 1 to the start, no need to concat (which increases memory usage for holding intermediate tensors)
  - otherwise, we need to allocate an empty array and write 1 to the start

leofang (changes requested):
Thanks for your contribution, @DanielCohenn! Please add tests and doc coverage as well.

cupy/_math/sumprod.py
DanielCohenn: Thanks for the CR! I will work on it tomorrow and hopefully should have it finished by the end of the week.

eriknw:
Thanks @DanielCohenn and @vprabhakar12! I'm curious why we need two separate PRs (this one and #9205) for `*_prod` and `*_sum`. Can we concentrate efforts on one PR/approach, and implement both at the same time? How do you two want to proceed, and what guidance would you like?

With a helper function as is being done in #9205, it would be straightforward to support both. The functions bodies could be e.g.:
```python
# prod
return _cumulative_op(x, cumprod, 1, axis, dtype, out, include_initial)

# sum 
return _cumulative_op(x, cumsum, 0, axis, dtype, out, include_initial)
```

The biggest gap in both PRs is testing. Here's an example to get started:
```python
@testing.with_requires('numpy>=2.0')
@pytest.mark.parametrize('func', ['cumulative_sum', 'cumulative_prod'])
class TestCumulativeSumProd:

    @testing.for_all_dtypes()
    def test_1d(self, xp, dtype, func):
        a = testing.shaped_arange((5,), xp, dtype)
        return getattr(xp, func)(a)
```

If `include_initial` is False, then the helper function should more-or-less be a straight pass-through to `cumsum` or `cumprod`.

Also, don't forget to add entries to `docs/source/reference/math.rst`.

cupy/_math/sumprod.py
```diff
@@ -220,6 +220,68 @@ def nancumprod(a, axis=None, dtype=None, out=None):
...
+        out[tuple(item)] = 1
+    
+        return out
+    
+    res = cumprod(x, axis=axis, dtype=dtype, out=out)
+
+    if include_initial:
+        initial_shape = list(x.shape)
+        initial_shape[axis] = 1
+        res = cupy.concat(
```
eriknw:
```suggestion
        res = cupy.concatenate(
```
`concatenate` is the standard form and what the rest of the codebase uses. `concat` is an alias.

DanielCohenn: Hey ! 
Yeah this implementation looks more straightforward and inclusive for both functions. I can go ahead and implement them both in the same PR.

DanielCohenn: I will also update the PR message after I am done implementing.

DanielCohenn:
Both functions are now implemented through one inclusive helper, `_cumulative_op`.

One subtle bug showed up in `_cumulative_op` for `include_initial=True` and `out=None`:

```python
if out is None:
    res_dtype = dtype if dtype is not None else x.dtype
    res = cupy.empty(expected_shape, dtype=res_dtype)
```

If `dtype` is omitted and `x` is `int8`, this allocates an `int8` buffer. NumPy instead follows its usual rule for cumulative ops without an explicit `dtype`: narrow integer inputs are promoted to the default integer type (`int64` on 64-bit), so CuPy and NumPy diverged (overflow vs correct values). The resolution is to derive the output dtype from the result of the inner `cumsum / cumprod` before building the extended array.

Tests and documentation are included in this PR as well.

@eriknw @leofang

DanielCohenn:
Hey, 
@asmeurer @leofang @eriknw pinging you guys to understand the current status of this PR and when it could be merged to main.

tests/cupy_tests/math_tests/test_sumprod.py
```diff
@@ -1246,3 +1246,86 @@ def test_trapz_1dim_with_x_and_dx(self, xp, dtype):
         a = testing.shaped_arange((5,), xp, dtype)
         x = testing.shaped_arange((5,), xp, dtype)
         return xp.trapezoid(a, x=x, dx=0.1)
+
+
+@testing.with_requires('numpy>=2.0')
```
eriknw:
```suggestion
@testing.with_requires('numpy>=2.1')
```
My bad earlier; it was actually numpy 2.1 that introduced these. See: https://numpy.org/doc/stable/release/2.1.0-notes.html#new-features

eriknw (changes requested):
Sorry for my delay @DanielCohenn, I was afk some last week! Thanks for making the updates--looks good. I left one minor suggestion (my initial mistake) that needs in.

There are two more things I would consider doing (CC @leofang for a quick sanity check on these):
- Add a `if not isinstance(x, cupy.ndarray):` check at the very beginning that raises a TypeError (otherwise, `cupy.atleast_1d` could silently do a H2D transfer, which is a footgun cupy tries to avoid)
- `cp.cumsum` and `cumprod` have slightly odd behavior (which may be considered pre-existing bug, dunno): when `out` is not None, the returned value is a _view_ over the values of `out`, not `out` itself. Per the Array API standard, we can make these functions return `out`
  - For example, we could add `return out if out is not None else result`

cupy/_math/sumprod.py
```diff
@@ -220,6 +220,145 @@ def nancumprod(a, axis=None, dtype=None, out=None):
...
+        item[axis] = slice(1, None)
+        op(x, axis=axis, dtype=dtype, out=res[tuple(item)])
+        return res
+
+    inner = op(x, axis=axis, dtype=dtype)
+    res = cupy.empty(expected_shape, dtype=inner.dtype)
+    item[axis] = 0
+    res[tuple(item)] = identity
+    item[axis] = slice(1, None)
+    res[tuple(item)] = inner
```
leofang: Q: Wouldn't this cause an intermediate copy?

eriknw: Yes, I believe it does. Good catch!

Perhaps the pattern that would fix this _and_ the issue with not returning `out` as I mentioned above would be to create `out = cupy.empty_like(...)` if `out` is None, then there would only be this one branch to handle for the rest of the implementation.

tests/cupy_tests/math_tests/test_sumprod.py
DanielCohenn: Hey ! 

Thanks, I fixed it.

DanielCohenn: Also, fixed the code according to your previous suggestions you are welcome to check the commit `[d9b25b5](https://github.com/cupy/cupy/pull/9864/commits/d9b25b5220471372c0184cfd0bfbf1745d102360)[d9b25b5](https://github.com/cupy/cupy/pull/9864/commits/d9b25b5220471372c0184cfd0bfbf1745d102360)`

cupy/_math/sumprod.py
DanielCohenn: Hey ! 

Good catch, I indeed saw the D2D copy using Nsight Systems and some dummy code. Fixed it in the most recent commit.

DanielCohenn:
@leofang @eriknw I went over your comments and fixed the code accordingly. Used some dummy code with Nsight Systems to check if the `h2d` and `d2d` copies were indeed resolved (and they are now).

DanielCohenn:
Hey ! @eriknw @leofang  pinging you guys.

DanielCohenn:
Hey @eriknw still waiting for your review

eriknw:
> Hey @eriknw still waiting for your review

Yup, I know. Near the top of my list. Sorry for my delay--I'm currently on vacation 😁. I should be able to reply early next week

DanielCohenn:
Hey, 
pinging @eriknw for a pending CR after fixes.

leofang (approved):

leofang:
Thanks @DanielCohenn @eriknw !

PR #9866, opened 2026-04-14
BUG: Make `cutensor` bindings threadsafe (and some small fixes)

seberg:
Trying to run free-threading tests on CI, I noticed that I never ran the tests with `cutensor` locally...
And, `cutensor` was never thread-safe at all, this fixes that.

I did use claude for some of the fixes, it is mostly churn, but roughly:
* Move the `handle` on to the device thread-local storage.
  * This is slightly awkard, because we change the `Handle` definition and because for the others the handle is an `intptr_t` only, while here it is a Python object.
* Most other caches are moved to the handle itself so that they are thread-local
* (Some caches are just updated to use `setdefault()` for nicer threading.)

The tests were also problematic... I had to have claude pull all `self.<var>` into the local namespace.  This just makes sense but is churn.
The `empty()` behavior for pinned memory was unreliable/odd but initializing the array fixes it.

(This should be all good, but I'll make a pass myself to check claude before merging.)

### DIFF
--- a/cupy/_core/_fusion_interface.py
+++ b/cupy/_core/_fusion_interface.py
@@ -23,7 +23,7 @@ def _set_dtype_to_astype_dict():
     This function is called at most once.
     """
     global _dtype_to_astype_dict
-    _dtype_to_astype_dict = {}
+    dtype_to_astype_dict = {}
 
     dtype_list = [numpy.dtype(type_char) for type_char in '?bhilqBHILQefdFD']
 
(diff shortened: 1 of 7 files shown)


leofang:
I think this makes sense to me based on my understanding in cuTENSOR, but I'd like to have @yangcal to take another look.

yangcal (approved):
All changes LGTM

leofang (approved):

PR #9877, opened 2026-04-18
Prepare mdspan for layout_stride_relaxed support

### LINKED ISSUE #8207: [BUG] ScanTileState static_assert fails for types with is_primitive=true and sizeof >= 16
## Summary

After #7575 (commit b9a0994c1c, "Support more types in decoupled lookback fastpath"), `cub::DeviceScan::InclusiveSum` fails to compile for types that are marked as primitive via `cub::NumericTraits` but have `sizeof(T) >= 16`, such as `thrust::complex<double>`.

## Root cause

#7575 changed the default template parameter of `ScanTileState` to include `is_primitive<T>` in the `SingleWord` condition:

```cpp
// single_pass_scan_operators.cuh, line ~657
bool SingleWord = detail::is_primitive<T>::value
                   || ::cuda::std::is_trivially_copyable_v<T>
                   && sizeof(T) < detail::largest_atomic_message_size
```

Due to C++ operator precedence (`&&` binds tighter than `||`), this evaluates as:

```cpp
is_primitive<T>::value || (is_trivially_copyable_v<T> && sizeof(T) < 16)
```

When `is_primitive<T>::value` is `true`, `SingleWord` is unconditionally `true` regardless of `sizeof(T)`. This selects the `ScanTileState<T, true>` specialization, which then hits:

```cpp
static_assert(sizeof(TileDescriptor) <= detail::largest_atomic_message_size);
```

For `thrust::complex<double>` (`sizeof == 16`), `TileDescriptor = {StatusWord, T}` exceeds 16 bytes.

## Who is affected

Projects like [CuPy](https://github.com/cupy/cupy) specialize `cub::NumericTraits` for complex types to enable CUB device algorithms:

```cpp
template <>
struct NumericTraits<thrust::complex<double>>
    : BaseTraits<FLOATING_POINT, true, unsigned long long,
                 thrust::complex<double>> {};
```

The second `BaseTraits` template parameter (`true`) marks the type as "primitive". This pattern worked on all prior CCCL releases and breaks after #7575.

## Minimal reproducer

```cpp
#include <thrust/complex.h>
#include <cuda/std/limits>

// Trait specializations to enable CUB device algorithms on complex types
// (same pattern used by CuPy)
namespace cuda {
namespace std {
template <>
class numeric_limits<thrust::complex<double>> {
  public:
    static __host__ __device__ thrust::complex<double> max() noexcept {
        return thrust::complex<double>(
            cuda::std::numeric_limits<double>::max(),
            cuda::std::numeric_limits<double>::max());
    }
    static __host__ __device__ thrust::complex<double> lowest() noexcept {
        return thrust::complex<double>(
            -cuda::std::numeric_limits<double>::max(),
            -cuda::std::numeric_limits<double>::max());
    }
    static constexpr bool is_specialized = true;
};
}  // namespace std
template <>
inline constexpr bool is_floating_point_v<thrust::complex<double>> = true;
}  // namespace cuda

#include <cub/util_type.cuh>
using namespace cub;
template <>
struct NumericTraits<thrust::complex<double>>
    : BaseTraits<FLOATING_POINT, true, unsigned long long,
                 thrust::complex<double>> {};

#include <cub/device/device_scan.cuh>
#include <cstdio>

int main() {
    using T = thrust::complex<double>;
    const int n = 1024;
    T* d_in = nullptr;
    T* d_out = nullptr;
    void* d_tmp = nullptr;
    size_t tmp_bytes = 0;

    cudaMalloc(&d_in, n * sizeof(T));
    cudaMalloc(&d_out, n * sizeof(T));

    // This triggers the static_assert failure
    auto err = cub::DeviceScan::InclusiveSum(
        d_tmp, tmp_bytes, d_in, d_out, n);

    printf("err = %d\n", (int)err);
    cudaFree(d_in);
    cudaFree(d_out);
    return 0;
}
```

Compile with:
```
nvcc --std=c++17 --expt-relaxed-constexpr \
  -I <cccl>/cub -I <cccl>/thrust -I <cccl>/libcudacxx/include \
  cccl_scan_complex_repro.cu
```

Error:
```
single_pass_scan_operators.cuh(693): error: static assertion failed
    static_assert(sizeof(TileDescriptor) <= detail::largest_atomic_message_size);
```

## Suggested fix

The `SingleWord` condition should ensure the size check applies to both `is_primitive` and `is_trivially_copyable` types. For example:

```cpp
bool SingleWord = (detail::is_primitive<T>::value
                   || ::cuda::std::is_trivially_copyable_v<T>)
                   && sizeof(T) < detail::largest_atomic_message_size
```

Or alternatively, add `sizeof(T) < detail::largest_atomic_message_size` to the `is_primitive` branch explicitly.

-- Leo's bot

### LINKED ISSUE #9772: Prepare to support `mdspan`'s `layout_stride_relaxed`
The CCCL team is working on a new `layout_stride_relaxed` (https://github.com/NVIDIA/cccl/pull/7175) which is more compatible with Python needs (supporting negative/zero strides). Once it's out, the existing `mdspan` would break because it needs a new tailing `offset` member, thanks to @oleksandr-pavlyk and @fbusato for pointing this out to me.

Specifically, it requires the following memory layout (in the case where all extents/strides are dynamic):
```c
struct {
  void* ptr;
  IndexType[N] extents;
  StrideType[N] strides;
  StrideType offset.
} mdspan;
```

leofang:
## Summary

Add support for CCCL's `cuda::layout_stride_relaxed` to CuPy's mdspan, enabling correct handling of negative strides (reversed array views) and zero strides (broadcasting).

### Struct layout

The packed mdspan struct now includes an offset field:

```
[void* ptr | IndexType extents[N] | IndexType strides[N] | ptrdiff_t offset]
```

The offset is always `ptrdiff_t` (8 bytes) to match `layout_stride_relaxed::mapping`'s default `offset_type`. This is important for the `IndexType=int32` case where the extents and strides are 4 bytes each but the offset remains 8 bytes:

| Component | IndexType=int32 | IndexType=int64 |
|---|---|---|
| `ptr` | 8 B | 8 B |
| `extents[N]` | N × 4 B | N × 8 B |
| `strides[N]` | N × 4 B | N × 8 B |
| `offset` | 8 B (ptrdiff_t) | 8 B (ptrdiff_t) |

xref: https://github.com/NVIDIA/cccl/blob/dee407140bde57111dd1fa9311434cb26429f035/libcudacxx/include/cuda/__fwd/mdspan.h#L82-L85

### Pointer and offset semantics

- **Non-negative strides** (or `allow_unsafe=True`): `ptr` = logical data pointer, `offset` = 0. Backward compatible with `layout_stride`, `layout_left`, `layout_right` (they ignore the trailing offset).
- **Negative strides** with `allow_unsafe=False` (default): `ptr` = base of allocation (`arr.data.mem.ptr`), `offset` = element distance from base to logical origin. The kernel must use `cuda::layout_stride_relaxed`.
- **`allow_unsafe=True`** with negative strides: legacy convention (`ptr=data.ptr`, `offset=0`) for backward compatibility with `layout_stride` kernels.

Negative strides no longer require `allow_unsafe=True`. Only zero strides and zero-size dimensions still do.

### Other changes

- Remove `NumericTraits` specializations for complex types — they declared `complex<float/double>` as primitive (`is_primitive=true`), causing UB in CUB's decoupled lookback scan (torn reads/writes). CUB reductions still work correctly without these traits. (Fixes NVIDIA/cccl#8207)
- Bump CCCL submodule to `dee407140b` (includes `layout_stride_relaxed`)
- Add `TestMdspanStrideRelaxed` test class with negative-stride kernel tests
- Update validation test to use `layout_stride_relaxed` with correct `sizeof` check

## Merge blocker

The CCCL submodule currently points to a `main` commit (`dee407140b`), not a stable release. Before merging, it should be updated to the first stable CCCL release that includes `layout_stride_relaxed`. Bumping CCCL is disruptive (invalidates CI cache) so we want to do it only once.

## Test plan

- [x] All 108 mdspan tests pass (0 failures, 0 skips)
- [x] CUB reduction tests pass for complex types after `NumericTraits` removal (80/80)
- [x] Existing mdspan tests (0D, 1D, 2D, 3D with all layouts and dtypes)
- [x] New `TestMdspanStrideRelaxed` tests (1D/2D negative strides, both-axes reversed, sliced+reversed)
- [x] Validation: negative strides accepted by default, zero strides rejected
- [ ] Full CI pass

Closes #9772

-- Leo's bot

### DIFF
--- a/cupy/_core/_carray.pyx
+++ b/cupy/_core/_carray.pyx
@@ -1,3 +1,4 @@
+from libc.stddef cimport ptrdiff_t
 from libc.string cimport memcpy
 from cpython.mem cimport PyMem_Malloc, PyMem_Free
 
@@ -49,7 +50,7 @@ cdef class mdspan(function.CPointer):
         # The offset is always ptrdiff_t (8 bytes) to match
         # layout_stride_relaxed's default offset_type.
         cdef size_t total_size = \
-            sizeof(void*) + ndim * 2 * index_itemsize + sizeof(long long)
+            sizeof(void*) + ndim * 2 * index_itemsize + sizeof(ptrdiff_t)
         cdef void* data = PyMem_Malloc(total_size)
         if data == NULL:
             raise MemoryError
@@ -64,7 +65,7 @@ cdef class mdspan(function.CPointer):
                 break
 
         cdef void* stored_ptr
-        cdef long long mem_offset
+        cdef ptrdiff_t mem_offset
         if has_negative_stride and not allow_unsafe:
             # Safe path for layout_stride_relaxed: store the base-of-
             # allocation pointer and compute the element offset from base
@@ -107,8 +108,8 @@ cdef class mdspan(function.CPointer):
 
         # Offset for layout_stride_relaxed support (see cupy/cupy#9772).
         # Always packed as ptrdiff_t (8 bytes) to match the C++ default.
-        (<long long*>(<char*>(data) + offset))[0] = mem_offset
-        offset += sizeof(long long)
+        (<ptrdiff_t*>(<char*>(data) + offset))[0] = mem_offset
+        offset += sizeof(ptrdiff_t)
 
         assert offset == total_size
 
--- a/cupy/_core/_carray.pyx
+++ b/cupy/_core/_carray.pyx
@@ -13,7 +13,8 @@ cdef fused index_t:
 cdef inline int populate_shape_strides(
         index_t val, size_t ndim, bint allow_unsafe,
         void* data, size_t& offset, int itemsize,
-        const shape_t& shape, const strides_t& strides) except?-1:
+        const shape_t& shape, const strides_t& strides,
+        long long mem_offset) except?-1:
     cdef int i
 
     for i in range(ndim):
@@ -24,13 +25,20 @@ cdef inline int populate_shape_strides(
         offset += sizeof(index_t)
 
     for i in range(ndim):
-        if not allow_unsafe and strides[i] <= 0:
+        if not allow_unsafe and strides[i] == 0:
             raise RuntimeError(
-                f"{i}-th dimension has non-positive stride")
+                f"{i}-th dimension has zero stride")
         (<index_t*>(<char*>(data) + offset))[0] = (
             <index_t>(strides[i] // itemsize)
         )
         offset += sizeof(index_t)
+
+    # Offset for layout_stride_relaxed support (see cupy/cupy#9772).
+    # For non-negative strides this is 0. For negative strides, this is the
+    # element distance from the base of the allocation to the logical origin,
+    # ensuring the layout mapping always returns a non-negative index.
+    (<index_t*>(<char*>(data) + offset))[0] = <index_t>mem_offset
+    offset += sizeof(index_t)
 
     return 0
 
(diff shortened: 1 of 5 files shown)


leofang:
This PR is blocked until there is a public CCCL release that includes `layout_stride_relaxed`.

cupy/_core/_carray.pyx
```diff
@@ -38,23 +39,52 @@ cdef inline int populate_shape_strides(
...
+        cdef void* stored_ptr
+        cdef ptrdiff_t mem_offset
+        if has_negative_stride and not allow_unsafe:
+            # Safe path for layout_stride_relaxed: store the base-of-
+            # allocation pointer and compute the element offset from base
+            # to logical origin, ensuring the layout mapping always
+            # returns a non-negative index.
+            stored_ptr = mem_ptr
+            mem_offset = (
+                (<intptr_t>data_ptr - <intptr_t>mem_ptr) // itemsize
```
stiepan, line 76: Can we trust that mem_ptr is a _base ptr_? As an example, packages don't bother to distinguish between data ptr and base ptr when exporting memory with dlpack, so if the array converted here comes from elsewhere - this won't give us the offset. The elsewhere can by cupy itself:

```
>>> x = cp.arange(3 * 5, dtype=cp.int8)[::-1]
>>> x.data.ptr - x.data.mem.ptr
14
>>> y = cp.from_dlpack(x.__dlpack__())
>>> y.data.ptr - y.data.mem.ptr  # both point at the last element of the array
0
```


Idea: ignore data.mem.ptr. Instead, compute `min_offset=sum((e - 1) * s for e, s in zip(shape, strides) if s < 0 and e > 0)` (note it is zero or negative), then take `offset=-min_offset` and `mem_ptr = data.ptr + min_offset`.

PR #9890, opened 2026-04-24
CI,TST: Improve xdist GPU distribution and try using for CI

seberg:
This modifies the setup slightly, because I noticed that otherwise the note is printed even when xdist isn't used (which is a bit confusing).

Otherwise, it sets `CUPY_GPU_MEMORY_LIMIT` as I suspect this may have been part of the reason why pytest-xdist wasn't reliable in CI previously.

Then attempt using pytest-xdist in CI (this will have to be an experiment, it is very possible that the same old issues appear).

*Draft until it seems good, if it seems to work well, should probably also use it on windows.*

### DIFF
--- a/.pfnci/linux/tests/actions/unittest.sh
+++ b/.pfnci/linux/tests/actions/unittest.sh
@@ -21,6 +21,8 @@ if [[ "${MARKER}" =~ "not slow" ]]; then
     # I we are not running slow tests (and not high memory)
     # running tests in parallel should be safe.
     pytest_opts+=("-n2")
+    # Use nvidia-cuda-mps-control -d to start MPS server and hopefully help execution speed on T4
+    nvidia-cuda-mps-control -d
 fi
 
 # TODO: support coverage reporting
--- a/.pfnci/linux/tests/actions/unittest.sh
+++ b/.pfnci/linux/tests/actions/unittest.sh
@@ -17,6 +17,12 @@ if [[ "${MARKER}" != "" ]]; then
     pytest_opts+=(-m "${MARKER}")
 fi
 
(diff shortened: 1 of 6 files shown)


seberg:
Hmmmm, seems about 45 instead of 65 minutes.  The memory test can be worked around (although there is a cryptic rare one I didn't understand yet).
The other one that fails due to MemoryError can probably just be marked as slow and into the slow tests.  (Although, I am not actually sure when we run slow tests, just locally?)

seberg:
One more try, tests are now grouped, and the one MemoryError test tweaked to maybe just get away with things.

/test linux,cuda130

EDIT: Well, this is slow, or even hangs... although, we don't have to group that agressively.  Seems like the mysterious hangs that caused pytest-xdist to not be used in CI might still exists, despite it working well locally :/

seberg:
Hmmm, not sure why the test deadlocked there... I did change it while testing to copy to host more (and maybe use more pinned memory). But due to previous tests that shouldn't have been problematic.
I am wondering if there is another deadlock for GPU memory allocation, but there isn't all that much going on, so really not sure.
(I did learn that `try_lock()` as used in the normal allocator is not be safe on windows if the Python GC can kick in at weird times -- not sure it can but have to think about cleaning that up.  But... that can't be the issue here.)

So, I currently have no good idea why `test_asnumpy_blocking` might hang/crash, but maybe reducing the memory use helps?  Since OoM, or close to it, I can at least imagine bad things...

/test linux,cuda130

seberg:
Hmmm, `loadscoped` just isn't that fast?  Try again with grouped (just the memory tests need it, maybe).
Also PTDS can lead to dead-locks.

/test linux,cuda130

PR #9899, opened 2026-04-29
Fix silent corruption in thrust sort/argsort/lexsort under OOM

### LINKED ISSUE #9894: thrust sort/argsort silently produces wrong results when workspace allocation fails
## Summary

`cupy.sort()` and `cupy.argsort()` can silently return incorrect results when GPU memory is under pressure and thrust workspace allocation fails. The root cause is that `cupy/cuda/thrust.pyx` declares `cupy_malloc()` as `noexcept with gil`. When `memory.alloc()` raises `OutOfMemoryError`, Cython swallows the exception, prints a warning to stderr, and returns NULL. The C++ allocator does not validate this NULL pointer, so thrust receives a garbage workspace buffer, silently corrupts the sort output, and returns normally without raising an exception.

## Expected Behavior

`cupy.sort()` and `cupy.argsort()` should raise `OutOfMemoryError` when thrust workspace allocation fails, preventing silent data corruption.

## Actual Behavior

When VRAM utilization crosses a threshold (varies by GPU architecture):
- A message is printed to stderr: `Exception ignored in: 'cupy.cuda.thrust.cupy_malloc'` followed by the full `OutOfMemoryError` traceback
- The `sort()` / `argsort()` call returns **without raising an exception**
- The returned array contains **incorrect values**
- Downstream computation silently uses corrupted results

## Environment

| Property | Value |
|----------|-------|
| **CuPy version** | 13.6.0 (confirmed; bug likely present in v9+) |
| **CUDA version** | 11.8 |
| **Driver version** | 580.126.09 |
| **GPUs tested** | NVIDIA V100-SXM2 (32GB), L20 (48GB), RTX 3090 (24GB) |
| **OS** | Linux |

## Reproduction Script

```python
import cupy
import numpy as np

# Step 1: Create ground truth on CPU
rng = np.random.default_rng(42)
cpu_data = rng.standard_normal(140_000_000).astype(np.float32)
cpu_sorted_indices = np.argsort(cpu_data)
cpu_ground_truth = cpu_data[cpu_sorted_indices]

# Step 2: Run on GPU with memory pressure
# Allocate padding to reduce available VRAM (adjust size for your GPU)
# For 32GB V100: allocate ~25-30 GB to trigger corruption at 100M elements
padding = cupy.zeros((28 * 1024**3) // 4, dtype=cupy.float32)

# Step 3: Sort the same data on GPU
gpu_data = cupy.asarray(cpu_data)
gpu_sorted_indices = cupy.argsort(gpu_data)  # no exception raised
gpu_sorted_data = gpu_data[gpu_sorted_indices]

# Step 4: Compare — silent corruption if stderr shows "Exception ignored"
matches = int(cupy.sum(gpu_sorted_indices == cupy.asarray(cpu_sorted_indices)))
total = len(cpu_data)
correctness = 100 * matches / total

print(f"GPU correctness: {matches}/{total} ({correctness:.1f}%)")
print(f"CPU ground truth (first 10 sorted values): {cpu_ground_truth[:10]}")
print(f"GPU result (first 10 sorted values): {cupy.asnumpy(gpu_sorted_data[:10])}")

del padding  # free for cleanup
```

**Expected stdout**: `GPU correctness: 140000000/140000000 (100.0%)`

**Actual stdout under memory pressure**:
```
GPU correctness: 0/140000000 (0.0%)
CPU ground truth (first 10 sorted values): [-4.57... -4.32... -4.20...]
GPU result (first 10 sorted values): [1.23... 5.67... -2.34...]
```

**Observed stderr** (printed by Cython, NOT by user code):
```
cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 7,024,805,376 bytes (allocated so far: 19,021,614,080 bytes).
  File "cupy/cuda/memory.pyx", line 875, in cupy.cuda.memory.alloc
  File "cupy/cuda/memory.pyx", line 1579, in cupy.cuda.memory.MemoryPointer.__init__
  File "cupy/cuda/memory.pyx", line 1600, in cupy.cuda.memory.MemoryPointer.__init__
  File "cupy/cuda/memory.pyx", line 1271, in cupy.cuda.memory._malloc
  File "cupy/cuda/memory.pyx", line 1292, in cupy.cuda.memory._malloc
  File "cupy/cuda/memory.pyx", line 1537, in cupy.cuda.memory.MemoryPool.malloc
  File "cupy/cuda/memory.pyx", line 1540, in cupy.cuda.memory.MemoryPool.malloc
cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 7,024,805,376 bytes (allocated so far: 19,021,614,080 bytes).
Exception ignored in: 'cupy.cuda.thrust.cupy_malloc'
```

### Corruption Threshold by Architecture

The corruption threshold scales with available VRAM. Testing on three architectures shows the pattern:

| GPU | Total VRAM | Corruption starts when | Allocation size | Allocated so far |
|-----|-----------|----------------------|-----------------|-----------------|
| V100-SXM2 | 32 GB | ~100-110 GB allocated | 7,024,805,376 B | 25,000,000,000 B |
| L20 | 48 GB | ~150 GB allocated | 7,024,805,376 B | 35,000,000,000 B |
| RTX 3090 | 24 GB | ~110 days of unsharded eval | 7,024,805,376 B | 19,021,614,080 B |

The corruption occurs when thrust cannot allocate its workspace buffer (~7 GB), but the host application has not yet hit a hard OOM exception. This creates a window where:
- Signal computation arrays consume 75-85% of VRAM
- Thrust workspace request fails
- No exception propagates to Python
- Sort operates on garbage memory

### Code Path Analysis

**In `cupy/cuda/thrust.pyx` (~line 30):**

```cython
cdef public char* cupy_malloc(void *m, size_t size) noexcept with gil:
    if size == 0:
        return <char *>0
    cdef _MemoryManager mm = <_MemoryManager>m
    mem = memory.alloc(size)  # <-- raises OutOfMemoryError
    mm.memory[mem.ptr] = mem
    return <char *>mem.ptr
```

The `noexcept` annotation tells Cython: "this function will never raise a Python exception." When `memory.alloc()` raises `OutOfMemoryError`, Cython's exception handler:
1. Catches the exception (violating the `noexcept` contract)
2. Prints "Exception ignored in: 'cupy.cuda.thrust.cupy_malloc'" to stderr
3. Returns the default value for `char*`, which is NULL (pointer 0)

The exception is **not propagated** to Python; the function appears to succeed from the caller's perspective.

**In `cupy/cuda/cupy_thrust.cu`:**

```cpp
class cupy_allocator {
    char *allocate(size_t num_bytes) {
        return cupy_malloc(memory, num_bytes);  // No null check
    }
};
```

The allocator does not validate the returned pointer. A NULL workspace buffer is passed directly to thrust.

**Thrust behavior:**

Thrust sort/argsort operates on the garbage memory at address 0x0, produces undefined/wrong output, and returns normally. There is no error-handling code path in thrust for invalid workspace pointers — it assumes the allocator returned valid memory.

### Existing Code Comment

A comment in the existing source code acknowledges this issue:

```cython
# NOTE(seberg): On failure, thrust may expect a C++ exception, `noexcept`
# prints it out and returns NULL (not great, but maybe OK).
```

The comment underestimates the severity. **Returning NULL with silent data corruption is not OK** — it violates the principle that sorting should either succeed with correct output or fail with an exception. Silent correctness violations are among the most dangerous types of bugs.

## Impact

This is a **silent data corruption bug**. 

Users get mathematically wrong results with no exception or warning visible in their Python code. The only indicator is a message to stderr, which is:
- Easy to miss in batch/cluster environments
- Often not monitored in production
- Frequently buffered or discarded

The sort operation appears to complete successfully, making the corruption difficult to detect until downstream analysis reveals inconsistent results. On GPUs with limited VRAM (24-48 GB), this can occur during normal workloads (not just pathological edge cases).

## Affected Versions

The `noexcept` declaration in `cupy_malloc()` has been present for several years. This bug affects CuPy v9+, likely including all current releases through v13.6.0.

### Option A: Add NULL check in C++ allocator (Recommended)

Minimal and preferred. Requires only C++ changes:

```cpp
class cupy_allocator {
    char *allocate(size_t num_bytes) {
        if (num_bytes == 0) return nullptr;
        char* ptr = cupy_malloc(memory, num_bytes);
        if (ptr == nullptr) {
            throw std::bad_alloc();
        }
        return ptr;
    }
};
```

When thrust catches `std::bad_alloc`, it will propagate the error appropriately instead of using corrupted memory.

### Option B: Remove noexcept and add explicit exception handling in Cython

Requires changes to both Cython and C++. More invasive but ensures the exception type is preserved:

```cython
cdef public char* cupy_malloc(void *m, size_t size) with gil:  # Remove noexcept
    if size == 0:
        return <char *>0
    cdef _MemoryManager mm = <_MemoryManager>m
    try:
        mem = memory.alloc(size)
    except MemoryError:
        # Raise an exception that thrust can handle
        raise
    mm.memory[mem.ptr] = mem
    return <char *>mem.ptr
```

**Option A is preferred** because it is minimal, maintains the existing Cython/C++ interface boundaries, and is less likely to introduce new issues.

## Testing Recommendations

A regression test should verify one of the following for each affected operation (`sort`, `argsort`, etc.):

1. **Correctness test**: Allocate nearly all available GPU memory, call the operation on a large array, and verify that either:
   - The operation completes correctly, OR
   - An `OutOfMemoryError` is raised (not just printed to stderr)

2. **Stderr check**: Verify that no `"Exception ignored in: 'cupy.cuda.thrust.cupy_malloc'"` message appears in stderr during normal operation.

3. **Result validation**: For small test arrays under memory pressure, compare GPU results against CPU (NumPy) reference to catch silent corruption.

Example test structure:

```python
def test_sort_under_memory_pressure():
    """Verify sort raises OutOfMemoryError instead of silently corrupting data."""
    import cupy
    import numpy as np
    
    # Create test array
    cpu_data = np.random.randn(100_000_000).astype(np.float32)
    cpu_result = np.argsort(cpu_data)
    
    # Allocate nearly all available GPU memory
    gpu_total = cupy.cuda.Device().mem_info[0]
    padding = cupy.zeros((int(0.95 * gpu_total)) // 4, dtype=cupy.float32)
    
    gpu_data = cupy.asarray(cpu_data)
    
    # Should raise OutOfMemoryError, not silently corrupt
    with pytest.raises(cupy.cuda.memory.OutOfMemoryError):
        gpu_result = cupy.argsort(gpu_data)
    
    del padding
```

## Additional Context

This bug was discovered in production during numerical signal processing on panel data with millions of elements. The corruption manifested as a divergence between chunked GPU evaluation (correct, scores ~-2.75) and unsharded GPU evaluation (corrupted, scores ~-1.04) on the same input data. CPU oracle confirmed the correct value, eliminating floating-point arithmetic as the cause.

Testing across three GPU architectures (V100, L20, RTX 3090) shows the corruption threshold scales with VRAM, confirming the root cause is the memory pressure condition, not a GPU-specific or CUDA version issue.

eriknw:
Thrust's allocator callback `cupy_malloc` is declared `noexcept` on the Cython side: when `memory.alloc` raises `OutOfMemoryError`, the exception is swallowed (printed to stderr) and NULL is returned.  The C++ `cupy_allocator` did not check for NULL, so thrust silently operated on a NULL workspace and produced corrupt results.

- `cupy_thrust.cu`: throw `std::bad_alloc` when `cupy_malloc` returns NULL.
- `thrust.pyx`: add `except +` to the thrust externs so `std::bad_alloc` surfaces as `MemoryError`; catch `MemoryError` in `cupy_malloc` to suppress the noisy "Exception ignored" stderr trace.
- Add regression tests using `cupy.cuda.using_allocator` that fail the last allocation of each routine (always inside thrust's workspace request).

I measured virtually no performance impact.

Fixes cupy/cupy#9894.

### DIFF
--- a/cupy/cuda/cupy_thrust.cu
+++ b/cupy/cuda/cupy_thrust.cu
@@ -8,6 +8,7 @@
 #include <thrust/sort.h>
 #include <thrust/tuple.h>
 #include <thrust/execution_policy.h>
+#include <new>
 #include <type_traits>
 
 
@@ -37,7 +38,15 @@ public:
     cupy_allocator(void* memory) : memory(memory) {}
 
     char *allocate(size_t num_bytes) {
-        return cupy_malloc(memory, num_bytes);
+        if (num_bytes == 0) return nullptr;
+        // cupy_malloc returns NULL on out-of-memory (see thrust.pyx).
+        // Without this check thrust would sort into a NULL workspace
+        // and produce corrupt results (cupy/cupy#9894).
+        char *ptr = cupy_malloc(memory, num_bytes);
+        if (ptr == nullptr) {
+            throw std::bad_alloc();
+        }
+        return ptr;
     }
 
     void deallocate(char *ptr, size_t n) {
--- a/cupy/cuda/thrust.pyx
+++ b/cupy/cuda/thrust.pyx
@@ -24,13 +24,21 @@ cdef class _MemoryManager:
         self.memory = dict()
 
 
-# NOTE(seberg): On failure, thrust may expect a C++ exception, `noexcept`
-# prints it out and returns NULL (not great, but maybe OK).
+# MemoryError is caught explicitly so `noexcept` does not print an
+# "Exception ignored in: cupy_malloc" trace to stderr.  The C++ allocator
+# (cupy_thrust.cu) checks for NULL and throws std::bad_alloc so thrust
+# unwinds cleanly; the externs below use `except +` to convert that into
+# a Python MemoryError visible to the caller.  Any other (unexpected)
+# Python exception is left to noexcept to surface for diagnosis.
+# See cupy/cupy#9894.
 cdef public char* cupy_malloc(void *m, size_t size) noexcept with gil:
     if size == 0:
         return <char *>0
     cdef _MemoryManager mm = <_MemoryManager>m
-    mem = memory.alloc(size)
+    try:
+        mem = memory.alloc(size)
+    except MemoryError:
+        return <char *>0
     mm.memory[mem.ptr] = mem
     return <char *>mem.ptr
 
@@ -48,12 +56,16 @@ cdef public int cupy_free(void *m, char* ptr) except -1 with gil:
 ###############################################################################
 
 cdef extern from 'cupy_thrust.h' nogil:
+    # `except +` propagates std::bad_alloc thrown by the C++ allocator
+    # (when cupy_malloc returns NULL) as a Python MemoryError instead of
+    # letting thrust silently operate on garbage memory.  See cupy/cupy#9894.
     void thrust_sort(int, void *, size_t *, const vector.vector[ptrdiff_t]&,
-                     intptr_t, void *)
+                     intptr_t, void *) except +
     void thrust_lexsort(
-        int, size_t *, void *, size_t, size_t, intptr_t, void *)
-    void thrust_argsort(int, size_t *, void *, void *,
-                        const vector.vector[ptrdiff_t]&, intptr_t, void *)
+        int, size_t *, void *, size_t, size_t, intptr_t, void *) except +
+    void thrust_argsort(
+        int, size_t *, void *, void *,
+        const vector.vector[ptrdiff_t]&, intptr_t, void *) except +
 
     # Build-time version
     int THRUST_VERSION
--- a/tests/cupy_tests/sorting_tests/test_sort.py
+++ b/tests/cupy_tests/sorting_tests/test_sort.py
@@ -433,6 +433,84 @@ class TestSort_complex(unittest.TestCase):
         return a, xp.sort_complex(a)
 
 
+class TestThrustWorkspaceOOM:
+    """Regression tests for cupy/cupy#9894.
+
+    When thrust's workspace allocation fails, sort/argsort/lexsort must
+    raise ``MemoryError`` instead of silently producing corrupt results.
+
+    Each op may make several pre-thrust allocations (e.g. ``data.copy()``
+    and ``idx_array``) before reaching thrust.  Failing the *first*
+    allocation only exercises pre-existing OOM behavior, not this fix.
+    To target thrust's workspace specifically, we count allocations during
+    a successful run, then re-run with the *last* allocation forced to
+    fail.  Since thrust is called last in each routine, the final
+    allocation is always inside thrust's workspace request.
+    """
+
+    @staticmethod
+    def _verify_workspace_oom_raises(op):
+        pool = cupy.get_default_memory_pool()
+        n = [0]
+
+        def counting(size):
+            n[0] += 1
+            return pool.malloc(size)
+
+        with cupy.cuda.using_allocator(counting):
+            op()
+        assert n[0] >= 1, "expected at least one allocation"
+        total = n[0]
+
+        seen = [0]
+
+        def fail_on_last(size):
+            seen[0] += 1
+            if seen[0] >= total:
+                raise cupy.cuda.memory.OutOfMemoryError(size, 0, 0)
+            return pool.malloc(size)
+
+        with cupy.cuda.using_allocator(fail_on_last):
+            with pytest.raises(MemoryError):
+                op()
+
+    def test_sort_workspace_oom(self):
+        self._verify_workspace_oom_raises(
+            lambda: cupy.arange(100_000, dtype=cupy.float32).sort()
+        )
+
+    def test_argsort_workspace_oom(self):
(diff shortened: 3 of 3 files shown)


seberg:
Nice thanks for cleaning that up! Clearly not a good hope that thrust likes a `NULL` return (but even then the `except +` is a nice spot!).
That feels long, but it is impressively thorough :) -- a nice pattern to keep in mind if we run into other alloc failures.

/test mini

seberg (approved):

seberg:
Same three failures as in gh-9897 :/

/test force-skip

PR #9900, opened 2026-04-29
Deprecate sparse matrix APIs removed in SciPy 1.14

eriknw:
Emit DeprecationWarning on spmatrix.A, .H, asfptype, getformat, getmaxprint, and set_shape. Non-breaking; backportable to v14.

Reroute shape.setter through reshape() so `m.shape = ...` skips the deprecated set_shape path. Add a maxprint class-attribute default to fix a latent AttributeError on subclasses that bypass spmatrix.__init__.

getnnz and get_shape are deferred pending a _getnnz/_get_shape refactor (currently called internally by .nnz/.size/.shape).

Takes over #8682.

@leofang, this is a surgical PR targeting cupy 14.1.0 release (if you want to). Each deprecation has a working migration target on spmatrix today, so there is no need to wait for sparray imho (and it'll be straightforward to include deprecations in sparray PR too).

In which version might the deprecations be removed?

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -392,6 +392,41 @@ configs {
   }
 }
 
+# CUDA 13.2 | Linux
+configs {
+  key: "cupy.linux.cuda132"
+  value {
+    requirement {
+      cpu: 8
+      memory: 60
+      disk: 30
+      gpu: 1
+    }
+    time_limit: {
+      seconds: 21600
+    }
+    environment_variables { key: "GPU" value: "1" }
+    command: ".pfnci/linux/main-flexci.sh cuda132"
+  }
+}
+# CUDA 13.2 (Multi-GPU) | Linux
+configs {
+  key: "cupy.linux.cuda132.multi"
+  value {
+    requirement {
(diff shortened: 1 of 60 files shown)


eriknw:
Oops. I'm going to need to update this. Many things in scipy went through a deprecation cycle and were removed from sparse arrays, but still exist on sparse matrix w/o deprecation warnings.

eriknw:
Okay, this PR has been trimmed down after the latest merge with main branch.  Basically, `.A` and `.H` attributes were _removed_ in scipy 1.14 from sparse matrix. This PR deprecates them. The other methods on scipy matrices don't raise warnings as of scipy 1.17, so we don't either anymore.

I also found it interesting that some tests require scipy <1.14. CI only tests scipy 1.14-1.16 currently.

Anyway, hopefully this PR should be quick to review now. I kept the other tweaks in it for quality of life improvements.

seberg (approved):
Thanks, confirmed that this is long deprecated/removed on the SciPy side and we discussed that it should be good to just add this (and maybe even finalize relatively quickly considering how long SciPy doesn't have it)

PR #9782, opened 2026-03-06
Update `test_assumed_runtime_version` for Windows + CUDA >=13.0

leofang:
Close cupy/cupy#9776.

PR #9902, opened 2026-04-30
Remove `test_assumed_runtime_version`

### LINKED ISSUE #9898: Windows runtime test failure
It seems https://github.com/cupy/cupy/pull/9782 is not robust...
https://ci.preferred.jp/cupy.win.cuda131/215774/
```
FAILED cupy_tests\cuda_tests\test_runtime.py::test_assumed_runtime_version - assert 13020 in (13000, 13010)	
```
I need to follow up with the Windows driver team again.

leofang:
## Summary

Remove `test_assumed_runtime_version` which is not reliably testable.

## Details

The test compared `_getLocalRuntimeVersion()` against a value derived from `nvrtc.getVersion()`. This comparison involves three independent version sources that can all disagree:

| API | Version returned |
|---|---|
| `nvrtc.getVersion()` | Run-time NVRTC library version (from local CTK, pip wheel, or conda package) |
| `runtime.runtimeGetVersion()` | Compile-time `CUDART_VERSION` baked into whichever cudart CuPy links against (CuPy's own build-time CTK, or `cuda.bindings`' build-time CTK) |
| `runtime._getLocalRuntimeVersion()` | Dynamically-loaded `cudaRuntimeGetVersion()` from the run-time cudart. On Linux or Windows + CUDA 12.x this comes from `libcudart.so` / `cudart64_*.dll` (the run-time cudart, not build-time). On Windows + CUDA >= 13.0 this returns the display driver's runtime version instead, which is decoupled from the toolkit (nvbugs 5955788, 5523579). |

These three values are only guaranteed to share the same *major* version. There is no combination of assertions that is correct across all configurations (Linux/Windows, with/without `cuda.bindings`, matching vs mismatched driver versions). The previous fix (#9782) hard-coded expected values that broke as soon as the CI driver was updated (#9898).

Close #9898.

-- Leo's bot


seberg (approved):
LGTM, and we need to get CI green, so let's get it in.

Its not 100% clear to me if this test tested something that was actually relevant, but I don't really think so (or at least not particularly).

leofang:
Yeah it used to be relevant in the beginning, but as both CuPy and cuda-bindings moved to statically link to cudart, it no longer makes sense to run this test. The change in the Windows driver 13.x is the last straw.

PR #9906, opened 2026-05-04
Use nvmath bindings for CUDA library bindings

yangcal:
## Summary

This PR enables the `CUPY_USE_CUDA_PYTHON` build path, which routes selected CUDA library bindings through `nvmath.bindings` instead of CuPy's header-based declarations.

All migrated modules follow the same pattern: when `CUPY_USE_CUDA_PYTHON=1`, the types and prototypes are `cimport`ed from the corresponding `nvmath.bindings.cy*` module; otherwise the existing header-based declarations are used, so default CUDA, HIP, and stub builds are unchanged.

| Module | nvmath binding |
| --- | --- |
| cuBLAS | `nvmath.bindings.cycublas` |
| cuRAND | `nvmath.bindings.cycurand` |
| cuSolverDn | `nvmath.bindings.cycusolverDn` |
| cuSolverSp | `nvmath.bindings.cycusolverSp` |
| cuSPARSE | `nvmath.bindings.cycusparse` |

## Notes

- In the `CUPY_USE_CUDA_PYTHON` build, **all** CUDA symbols are loaded dynamically at runtime (via nvmath + cuda-pathfinder).
- **cuSPARSE** is the only module that still routes some symbols through CuPy's own `SoftLink` rather than nvmath: the `cusparseSpGEAM_*` APIs are not in any public release yet and `nvmath` does not expose them, so they are loaded via `SoftLink` (over the same pathfinder-loaded handle) unconditionally, while all nvmath-provided symbols keep their `SoftLink` entries only under `IF not CUPY_USE_CUDA_PYTHON`. Because the compile-time `CUSPARSE_VERSION` macro is unavailable on this path, `get_build_version()` returns `0` and a new `is_cuda_python_build()` helper lets `cupyx.cusparse.check_availability` fall back to the runtime `getVersion()` for feature gating.
- **cuSolver** — `cupyx.cusolver` batched helpers are refactored so call sites stay shared across the CUDA-Python and header-based builds.


yangcal:
@leofang @seberg `nvmath-python` 1.0.0 just got released yesterday and I have edited the yaml files with Dockerfiles updated, can we enable tests on this PR?

seberg:
Yes, let me poke CI:

/test mini

yangcal:
@seberg seems there is a CI glitch with `
flexCI execution error | [DEADLINE_EXCEEDED] deadline exceeded`, shall we re-launch?

cupy_backends/cuda/libs/cusparse.pyx
```diff
@@ -982,8 +1237,18 @@ cpdef int getVersion(intptr_t handle) except? -1:
...
+    return CUPY_USE_CUDA_PYTHON
+
+
 def get_build_version():
-    return CUSPARSE_VERSION
+    # In the CUPY_USE_CUDA_PYTHON mode cusparse is not required at build time,
+    # so there is no compile-time version available. Callers should rely on the
+    # runtime version (getVersion) for API availability checks instead.
+    IF CUPY_USE_CUDA_PYTHON:
+        return 0
```
seberg, line 1249: My agent is pointing out that this is not great.  A few paths use this to disable unsupported things.  Setting it to zero means that we _always_ disable them.

It seems to me this really needs to use a runtime version check lookup (I'll doubt it needs memoization, since the API call is probably just a tiny function with no locks, etc.).

cupyx/cusolver.pyx
```diff
@@ -33,37 +33,452 @@ import cupyx as _cupyx
...
+        float
+        double
+        cuComplex
+        cuDoubleComplex
+
+    cdef int gesvd_loop_internal(
+            cusolver.Handle handle, char jobu, char jobvt, int m, int n,
+            solver_dtype* A, void* s_ptr, solver_dtype* U,
+            solver_dtype* VT, solver_dtype* Work, int buffersize,
+            int* devInfo, int batch_size) except * nogil:
```
seberg: Can you mark this as `noexcept`?  This isn't good and if the reason for `noexcept` not working is the bindings I'd like to understand why.
(Note that I added a bit of a crazy check to the cupy CI; if that check doesn't fail when you just add `noexcept`, I think we are fine -- if it fails, we _may_ be fine.)

Otherwise, there are three options, pick your poison:
* Remove it entirely to be an implicit `except -1?`.  Which is fine, since a -1 return will be _some_ form of error (Python one *or* a bad CUDA status).
* Use `except -1` and inline the CUDA error check (downside is that it is a bit funky that we need the double check for the other path, though).

The current version would be an emergency solution to me because `except * nogil` feels a bit silly: `PyErr_Occurred()` needs the GIL, so we have to acquire the GIL to check for an error that probably never happens (I am assuming Cython implicitly does this, although I wonder if it warns about it...).

cupyx/cusolver.pyx
```diff
@@ -33,37 +33,452 @@ import cupyx as _cupyx
...
+        static int cupyx_zungqr_loop(
+                intptr_t handle, int m, int n, int k, intptr_t a_ptr,
+                int lda, intptr_t tau_ptr, intptr_t w_ptr, int buffersize,
+                intptr_t info_ptr, int batch_size, int origin_n) {
+            return orgqr_loop<cuDoubleComplex>(
+                handle, m, n, k, a_ptr, lda, tau_ptr, w_ptr, buffersize,
+                info_ptr, batch_size, origin_n);
+        }
+        """
+        int sgesvd_loop 'cupyx_sgesvd_loop'(
```
seberg: I guess we may not be able to avoid this fully (i.e. the cuda-python path really needs to use Cython, I assume?).

But I think one thing to make this a bit less painful is to just do:

```suggestion
        int sgesvd_loop 'gesvd_loop<float>'(
```
I am pretty sure that works and at least removes the in-inline C++ code block.

(of course, if we just always use the Cython version, I guess this'll just go away)

cupyx/cusolver.pyx
```diff
@@ -33,37 +33,452 @@ import cupyx as _cupyx
...
+        double
+        cuComplex
+        cuDoubleComplex
+
+    cdef int gesvd_loop_internal(
+            cusolver.Handle handle, char jobu, char jobvt, int m, int n,
+            solver_dtype* A, void* s_ptr, solver_dtype* U,
+            solver_dtype* VT, solver_dtype* Work, int buffersize,
+            int* devInfo, int batch_size) except * nogil:
+        cdef int i, status = 0
```
seberg, line 81: My bot pointed out that this is a bug fix.  Sorry for piggy-backing, but maybe we can just add that 0 init to the C++ code as well while here?

Although, honestly, this makes me now think that maybe we should just *always* use this function.  Make it check the error too probably and just delete the C++ version.
Besides not changing anything, I don't really see much reason why the C++ version is better (well, beyond maybe that that the fused type is a tad annoying).

EDIT: Actually moving this to Cython always should also help with fixing gh-10095

cupyx/cusparse.py
```diff
@@ -295,7 +295,12 @@ def _get_avail_version_from_spec(x):
 def check_availability(name):
     if not _runtime.is_hip:
         available_version = _available_cusparse_version
-        version = _cusparse.get_build_version()
+        if _cusparse.is_cuda_python_build():
+            # In the CUDA Python build, cusparse is not required at build time;
+            # all symbols are loaded at runtime, so use the runtime version.
+            version = getVersion()
+        else:
+            version = _cusparse.get_build_version()
```
seberg, line 303: No need to change, although I am wondering if it would be "obvious" to just always use the runtime `getVersion()`.

seberg (commented):
Thanks!  Let's just ignore the failure (if we restart the CI it'll rerun anyway).

Me and my agent found two things that need addressed I think (error return and runtime version checking).  And I really wonder if we shouldn't just always use the Cython version for the `*_loop` helpers.
(Other things are small comments right now; I'll probably want Leo to have a look if he wants to before merging.)

cupyx/cusparse.py
yangcal: Good question. I was thinking the same thing and went back to the git history. Turns out this block was once performing runtime check and then later changed to build time version check. I think the motivation here might be that `get_build_version` is cheaper as `getVersion` requires creating a cusparse handle if not cached already and thus incur more overhead.

cupyx/cusolver.pyx
yangcal: let's try `noexcept` first, committed in eb2740c79e1f25e77a8f80d899cc35f0f1f63aa3

cupyx/cusolver.pyx
yangcal: Good idea, locally verified that this works for the default header build, committed in 9118f077aa7b891edc708cdf8777d6fbc2362f0f

cupyx/cusolver.pyx
yangcal: good catch, I pushed the initialization in de61b8aa70d64b49a8a65d599eae53ad9ba35f69

RE: switching out of the c++ version, maybe it's better to put it in a followup MR? (or just wait until the cuda_python switch is flipped on)

seberg: Thanks and yeah let's follow up.  It might be the nicest solution to gh-10095 but you are right to not figure it out here.

cupyx/cusparse.py
seberg: Yeah, nvm.  It may also be that if the function isn't available at build time, we don't link it, even if it _might_ be available at runtime.
With a (presumably always correct) assumption runtime >= buildtime.

cupyx/cusolver.pyx
seberg: Thanks, this even runs in the preview tests, so appears these functions use no Python API.

cupy_backends/cuda/libs/cusparse.pyx
yangcal: Forgot to respond to this message. This should be fixed in 530fbf892f472959c4980e4d9734f8cfee8b0d44. I think all the missed checks should be stale now as cupy requires cuda 12+.

seberg (approved):
Thanks for the fixups, I think this is ready now.  We need to touch that new cython helper code to solve gh-10095 so I am happy to keep that as is now.

(For now this changes only things with `CUPY_USE_CUDA_PYTHON`.)

CC @leofang will put this in very soon but I think you are aware of the big picture anyway.

seberg:
Going to put this in to simplify trying follow up to avoid direct linking (i.e. probably always using that cython code).

PR #9909, opened 2026-05-06
Remove legacy cuSparse bindings from CUDA build

yangcal:
## Summary

This PR removes cuSPARSE bindings for APIs that are no longer shipped in CUDA 12+ headers from **CUDA builds.**

The removed APIs are kept under `IF CUPY_CUDA_VERSION == 0`, so HIP and stub builds continue to compile the same bindings as before. CUDA builds now only expose cuSPARSE APIs that are available with the CUDA 12+ headers CuPy currently supports.

This also includes a small bugfix for `cupyx.cusparse`: `dense2csc` and `dense2csr` are now capped at CUDA 12.0 in the availability table, matching the actual CUDA header/API surface and avoiding dispatch to missing backend symbols.

## Clarifications

- This does not remove the HIP/stub implementation paths. In CuPy’s build configuration, `CUPY_CUDA_VERSION == 0` covers HIP and stub builds.
- High-level sparse code already prefers newer generic cuSPARSE APIs such as `denseToSparse` / `sparseToDense` where available so there is no affect on functionality.
- Type declarations that do not require CUDA 12 removed function symbols are left intact where they are still useful for shared HIP/stub compatibility.

<details>
<summary>cuSPARSE wrapper APIs moved to HIP/stub-only builds</summary>

- `sgthr`, `dgthr`, `cgthr`, `zgthr`
- `scsrmv`, `dcsrmv`, `ccsrmv`, `zcsrmv`
- `csrmvEx_bufferSize`, `csrmvEx`
- `createCsrsv2Info`, `destroyCsrsv2Info`
- `scsrsv2_bufferSize`, `dcsrsv2_bufferSize`, `ccsrsv2_bufferSize`, `zcsrsv2_bufferSize`
- `scsrsv2_analysis`, `dcsrsv2_analysis`, `ccsrsv2_analysis`, `zcsrsv2_analysis`
- `scsrsv2_solve`, `dcsrsv2_solve`, `ccsrsv2_solve`, `zcsrsv2_solve`
- `xcsrsv2_zeroPivot`
- `scsrmm`, `dcsrmm`, `ccsrmm`, `zcsrmm`
- `scsrmm2`, `dcsrmm2`, `ccsrmm2`, `zcsrmm2`
- `createCsrsm2Info`, `destroyCsrsm2Info`
- `scsrsm2_bufferSizeExt`, `dcsrsm2_bufferSizeExt`, `ccsrsm2_bufferSizeExt`, `zcsrsm2_bufferSizeExt`
- `scsrsm2_analysis`, `dcsrsm2_analysis`, `ccsrsm2_analysis`, `zcsrsm2_analysis`
- `scsrsm2_solve`, `dcsrsm2_solve`, `ccsrsm2_solve`, `zcsrsm2_solve`
- `xcsrsm2_zeroPivot`
- `xcsrgeamNnz`
- `scsrgeam`, `dcsrgeam`, `ccsrgeam`, `zcsrgeam`
- `xcsrgemmNnz`
- `scsrgemm`, `dcsrgemm`, `ccsrgemm`, `zcsrgemm`
- `createCsrgemm2Info`, `destroyCsrgemm2Info`
- `scsrgemm2_bufferSizeExt`, `dcsrgemm2_bufferSizeExt`, `ccsrgemm2_bufferSizeExt`, `zcsrgemm2_bufferSizeExt`
- `xcsrgemm2Nnz`
- `scsrgemm2`, `dcsrgemm2`, `ccsrgemm2`, `zcsrgemm2`
- `scsc2dense`, `dcsc2dense`, `ccsc2dense`, `zcsc2dense`
- `scsr2csc`, `dcsr2csc`, `ccsr2csc`, `zcsr2csc`
- `scsr2dense`, `dcsr2dense`, `ccsr2dense`, `zcsr2dense`
- `sdense2csc`, `ddense2csc`, `cdense2csc`, `zdense2csc`
- `sdense2csr`, `ddense2csr`, `cdense2csr`, `zdense2csr`
- `createCooAoS`, `cooAoSGet`
- `spMatSetStridedBatch`
- `constrainedGeMM_bufferSize`, `constrainedGeMM`

</details>

This PR could partially unblock nvmath cusparse binding integration in https://github.com/cupy/cupy/pull/9906. (The deprecated cusparse API that's currently missing in nvmath still need to be addressed)


yangcal:
@eriknw @leofang @seberg for vis. I wanted to remove these legacy bindings from cusparse to make it easier for nvmath integration. CuPy already requires CUDA12+. They are not functional now and currently exist only due to the compatibility layer with HIP and stub builds. Any thoughts?

leofang:
Wow! This is freaking good (and scary) cleanup! Is the reason that we never hit issues at build time due to the stubs shadowing the real CUDA header...?
https://github.com/cupy/cupy/blob/8abef5766deeb5f02604892efd199331122d8eed/cupy_backends/cuda/cupy_cusparse.h#L660-L662

yangcal:
> Wow! This is freaking good (and scary) cleanup! Is the reason that we never hit issues at build time due to the stubs shadowing the real CUDA header...?
> 
> https://github.com/cupy/cupy/blob/8abef5766deeb5f02604892efd199331122d8eed/cupy_backends/cuda/cupy_cusparse.h#L660-L662

Yes, these stubs in the cuda headers can also be removed now

seberg:
Glancing through, this is nice.  Should we just do the cleanup (as a follow-up?) to delete everything that is CUDA <12?
The way these stubs are written doing nothing but returning success, makes me think that basically none of these can be used in a correct way by downstream (even in theory)...

yangcal:
> This pull request is now in conflicts. Could you fix it @yangcal? 🙏

done

yangcal:
@seberg Should we just do the cleanup (as a follow-up?) to delete everything that is CUDA <12?

My understanding is that we can not just remove them now from the cython extension files since `cupy_backends` also needs to  support `HIP` backend. These `cuXXXX.pyx/pxd` is a common layer for CUDA, HIP and Stub build. (Stub build is when there is neither CUDA nor HIP in the environment). If there are APIs that's still supported by HIP, I don't think we can delete them without breaking HIP build.

 However, after the migration of `cuda-python` completes and becomes default, I think we can completely get rid of the headers for the cuda build as well as stub build (stub can just use `cuda-python` build)

seberg:
> If there are APIs that's still supported by HIP, I don't think we can delete them without breaking HIP build.

OK, yeah.  There are a few here that we use via python in `cusparse` it seems.  I wonder how important they are given that they just result in a `RuntimeError` on CUDA since forever, but...

From a CUDA header perspective, you are right, we might as well just sit it out until we can delete the whole thing.  For the stubs, we could just omit all of these, but it doesn't matter (also cleaned up in the future automatically as you said).

From a HIP perspective, it may still be nice to weed out things a bit, but I thought it may be as easy as deleting everything when it isn't, so no point here, I agree.

(I'll approve, but leave open for Leo to have a quick glance.)

/test mini

seberg (approved):

leofang:
I don't want to block this PR, but to validate that the cython bindings can be built correctly for CUDA >=12, it is the best that we clean up `cupy_backends/cuda/cupy_cusparse.h`, otherwise it is shadowing the real header. 

@yangcal I'll review/merge tonight but it'd be nice if you could follow up 🙂

leofang:
We don't need to touch the main header `cupy_backends/cupy_sparse.h` or the HIP stub `cupy_backends/hip/cupy_hipsparse.h`, just the CUDA stub and the RTD stub `cupy_backends/stub/cupy_cusparse.h` is enough.

PR #9910, opened 2026-05-06
Add sparse array classes (csr_array, csc_array, coo_array, dia_array)

eriknw:
Stacked on #9825 (int64 sparse indices). See https://github.com/eriknw/cupy/pull/2 for smaller diff against that branch

Targets `int64_sparse_indices` because both branches change the same hot paths in `_compressed.py`, `_coo.py`, and `_construct.py`; merging on top keeps the diff reviewable. Adds the SciPy-1.11+ sparse-array API surface to `cupyx.scipy.sparse`, plus the SciPy 1.17 alignment work that depends on it.

## What's new

- **Sparse array classes**: `csr_array`, `csc_array`, `coo_array`, `dia_array`. Mirrors SciPy's mixin architecture: `_spbase` is the common base; `sparray` mixin provides element-wise `*` and `**`; `spmatrix` overrides keep legacy matmul-`*` and matrix-power-`**`.  Format logic lives in shared bases (`_csr_base`, `_csc_base`, ...); leaf classes are thin wrappers — ` / `permute_dims`, `coords` on COO, `prune()`, `astype(copy=False)` short-circuit, SciPy-style `__repr__`, `maxprint=` kwarg, `__round__`, sparse-singleton `__setitem__`, kind preservation through `tril`/`triu`/`find`, `safely_cast_index_arrays` export.
- **SciPy 1.14 deprecations**: `.A`, `.H`, `asfptype`, `get_shape`, `getformat`, `getmaxprint`, `set_shape` warn on `spmatrix` only; bool `__neg__` raises `NotImplementedError`.
- **Internal migration**: every `isspmatrix(x)` / `isinstance(x, csr_matrix)` site in `cupyx.cusparse`, `cusolver`, NCCL, and `sparse.linalg` now uses format helpers (`_is_csr_type`, etc.) so the same code paths accept both arrays and matrices.

## Semantics

```python
import cupyx.scipy.sparse as csp

A = csp.csr_array([[1, 2], [3, 4]])
A * A   # element-wise   (csr_matrix would matmul here)
A @ A   # matmul         (both kinds)
A ** 2  # element-wise   (csr_matrix would matrix-power)
```

## Tests

- New `test_sparray.py`: construction, format conversion, arithmetic dispatch (sparray `*` vs spmatrix `*`), indexing, reductions, dtype preservation, kind preservation across operations, cross-format ops, SciPy parity.
- `test_sparse_int64_indices.py` extended for sparray paths.
- Pre-existing sparse suite unchanged

## Out of scope (follow-ups)

- 1-D sparse arrays (SciPy 1.16+). This PR is 2-D only; documented in the user-facing docs page.
- Bool-dtype generalisation. cuSPARSE accepts only `?fdFD` for data; bool input through arithmetic falls back inconsistently.

### DIFF
--- a/cupyx/distributed/_nccl_comm.py
+++ b/cupyx/distributed/_nccl_comm.py
@@ -456,19 +456,34 @@ class _DenseNCCLCommunicator:
         nccl.groupEnd()
 
 
-def _make_sparse_empty(dtype, sparse_type):
+def _make_sparse_empty(dtype, sparse_type, *, sparray=False):
     data = cupy.empty(1, dtype)
     a = cupy.empty(1, 'i')
     b = cupy.empty(1, 'i')
     if sparse_type == 'csr':
-        return sparse.csr_matrix((data, a, b), shape=(0, 0))
+        cls = sparse.csr_array if sparray else sparse.csr_matrix
+        return cls((data, a, b), shape=(0, 0))
     elif sparse_type == 'csc':
-        return sparse.csc_matrix((data, a, b), shape=(0, 0))
+        cls = sparse.csc_array if sparray else sparse.csc_matrix
+        return cls((data, a, b), shape=(0, 0))
     elif sparse_type == 'coo':
-        return sparse.coo_matrix((data, (a, b)), shape=(0, 0))
+        cls = sparse.coo_array if sparray else sparse.coo_matrix
+        return cls((data, (a, b)), shape=(0, 0))
     else:
         raise TypeError(
             'NCCL is not supported for this type of sparse matrix')
+
+
+def _empty_like(model):
+    """Return an empty sparse object that matches the type of ``model``.
+
+    Preserves the array-vs-matrix distinction so operator semantics
+    (``*`` is matmul on matrices, element-wise on arrays) stay
+    consistent when reductions combine ``model`` with the placeholder.
+    """
+    return _make_sparse_empty(
+        model.dtype, _get_sparse_type(model),
+        sparray=isinstance(model, sparse.sparray))
 
 
 def _get_sparse_type(matrix):
@@ -613,8 +628,7 @@ class _SparseNCCLCommunicator:
                 raise ValueError(
                     'in_array and out_array must be the same format')
             result = in_array
-            partial = _make_sparse_empty(
-                in_array.dtype, _get_sparse_type(in_array))
+            partial = _empty_like(in_array)
             # each device will send and array with a different size
             for peer, ss in enumerate(shape_and_sizes):
                 shape = tuple(ss[0:2])
@@ -683,8 +697,7 @@ class _SparseNCCLCommunicator:
             raise ValueError(
                 'in_array must be a list or a tuple of sparse matrices')
         for s_m in in_array:
-            partial_out_array = _make_sparse_empty(
-                s_m.dtype, _get_sparse_type(s_m))
+            partial_out_array = _empty_like(s_m)
             cls.reduce(comm, s_m, partial_out_array, root, op, stream)
             reduce_out_arrays.append(partial_out_array)
         cls.scatter(comm, reduce_out_arrays, out_array, root, stream)
@@ -702,7 +715,7 @@ class _SparseNCCLCommunicator:
         cls.gather(comm, in_array, gather_out_arrays, root, stream)
         if comm.rank != root:
             gather_out_arrays = [
-                _make_sparse_empty(in_array.dtype, _get_sparse_type(in_array))
+                _empty_like(in_array)
                 for _ in range(comm._n_devices)
             ]
         for arr in gather_out_arrays:
@@ -787,8 +800,7 @@ class _SparseNCCLCommunicator:
         # out_array is a list of sparse matrices
         if comm.rank == root:
             for peer in range(comm._n_devices):
-                res = _make_sparse_empty(
-                    in_array.dtype, _get_sparse_type(in_array))
+                res = _empty_like(in_array)
                 if peer != root:
                     cls.recv(comm, res, peer, stream)
                 else:
@@ -832,7 +844,5 @@ class _SparseNCCLCommunicator:
             for a in r_arrays:
                 cls._recv(comm, a, i, a.dtype, a.size, stream)
             nccl.groupEnd()
-            out_array.append(_make_sparse_empty(
-                in_array[i].dtype,
-                _get_sparse_type(in_array[i])))
+            out_array.append(_empty_like(in_array[i]))
             cls._assign_arrays(out_array[i], r_arrays, shape)
--- a/cupyx/distributed/_nccl_comm.py
+++ b/cupyx/distributed/_nccl_comm.py
@@ -472,25 +472,24 @@ def _make_sparse_empty(dtype, sparse_type):
 
 
 def _get_sparse_type(matrix):
-    if sparse.isspmatrix_coo(matrix):
-        return 'coo'
-    elif sparse.isspmatrix_csr(matrix):
-        return 'csr'
-    elif sparse.isspmatrix_csc(matrix):
-        return 'csc'
-    else:
+    if not sparse.issparse(matrix):
         raise TypeError(
             'NCCL is not supported for this type of sparse matrix')
+    if matrix.format in ('coo', 'csr', 'csc'):
+        return matrix.format
+    raise TypeError(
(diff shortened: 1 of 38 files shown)


tests/cupyx_tests/scipy_tests/sparse_tests/test_sparray.py
```diff
@@ -0,0 +1,1482 @@
...
+        result = sparse.tril(mat_pair[0])
+        assert isinstance(result, sparse.spmatrix)
+
+    def test_triu_array(self, arr_pair):
+        result = sparse.triu(arr_pair[0])
+        assert isinstance(result, sparse.sparray)
+
+    def test_triu_matrix(self, mat_pair):
+        result = sparse.triu(mat_pair[0])
+        assert isinstance(result, sparse.spmatrix)
```
seberg, line 1287: If your agent wants a :cookie: from me, it should parametrizes, maybe via lambdas (although can just create list pairs):
```
lambda arg1, _: sparse.tril(arg1)
```
(and the fixture can be nice, but could be a `sptype` parameter and then create the pair depending on that to squash down the two classes.)

cupyx/scipy/sparse/_base.py
```diff
@@ -619,16 +625,225 @@ def transpose(self, axes=None, copy=False):
...
+
+    @property
+    def _csc_container(self):
+        from cupyx.scipy.sparse._csc import csc_matrix
+        return csc_matrix
+
+    @property
+    def _coo_container(self):
+        from cupyx.scipy.sparse._coo import coo_matrix
+        return coo_matrix
```
seberg, line 727: Can't we just import `from cupyx.scipy import sparse (as _sparse)` at the top and return `_sparse.<object>` here?
(Might be just a mood of mine to avoid inline import's if possible.)

cupyx/scipy/sparse/_base.py
```diff
@@ -619,16 +625,225 @@ def transpose(self, axes=None, copy=False):
...
+        # Cooperative MI: forward to the next __init__ in the MRO.
+        nxt = super().__init__
+        if nxt is not object.__init__:
+            nxt(*args, **kwargs)
+        elif args or kwargs:
+            # Direct instantiation of ``spmatrix(args)`` with no concrete
+            # format subclass; scipy raises here, so do the same.
+            raise TypeError(
+                'cannot instantiate spmatrix directly; use a format '
+                'subclass such as csr_matrix')
```
seberg, line 663: Not really worth thinking much about, it but that `nxt =` dance is funky and I wonder if it can't just be:
```
if type(self) is spmatrix:
    raise TypeError(...)

super().__init__(*args, **kwargs)
```
for practically the same effect (yeah, I guess not in some multiple inheritance scenarios, but gotta draw a line).

(I expected to find this code in SciPy and didn't, so not sure that scipy still has it around but doesn't matter.)

cupyx/scipy/sparse/_base.py
```diff
@@ -619,16 +554,175 @@ def transpose(self, axes=None, copy=False):
...
+                pass
+
+    # Matrix semantics: * is matmul (overrides _spbase element-wise default)
+    def __mul__(self, other):
+        return self._matmul_dispatch(other)
+
+    def __rmul__(self, other):
+        return self._rmatmul_dispatch(other)
+
+    def __pow__(self, other):
```
seberg, line 591: Followup: But I guess this should move to `linalg.matrix_power` (or at least add it) so that we have an alternative when switching to arrays.

cupyx/scipy/sparse/_construct.py
```diff
@@ -615,3 +624,99 @@ def kronsum(A, B, format=None):
...
+
+    Returns:
+        cupyx.scipy.sparse.sparray
+
+    .. seealso:: :func:`scipy.sparse.eye_array`
+
+    """
+    if n is None:
+        n = m
+    m, n = int(m), int(n)
```
seberg, line 675: I'd shout `operator.index`, but SciPy missed the chance.

cupyx/distributed/_nccl_comm.py
```diff
@@ -472,25 +472,24 @@ def _make_sparse_empty(dtype, sparse_type):
...
 
 def _get_sparse_type(matrix):
-    if sparse.isspmatrix_coo(matrix):
-        return 'coo'
-    elif sparse.isspmatrix_csr(matrix):
-        return 'csr'
-    elif sparse.isspmatrix_csc(matrix):
-        return 'csc'
-    else:
+    if not sparse.issparse(matrix):
```
seberg, line 475: Nitpick alert: `if sparse.ismatrix(matrix) and matrix.format in ('coo', 'csr', 'csc): return matrix.format` fits on a single line to avoid duplicating the error.

cupyx/scipy/sparse/csgraph/_traversal.py
```diff
@@ -46,8 +46,12 @@ def connected_components(csgraph, directed=True, connection='weak',
...
         raise ValueError('graph should have two dimensions')
 
-    if not cupyx.scipy.sparse.isspmatrix_csr(csgraph):
-        csgraph = cupyx.scipy.sparse.csr_matrix(csgraph)
+    if not (cupyx.scipy.sparse.issparse(csgraph)
+            and csgraph.format == 'csr'):
+        if cupyx.scipy.sparse.issparse(csgraph):
+            csgraph = csgraph.tocsr()
+        else:
+            csgraph = cupyx.scipy.sparse.csr_matrix(csgraph)
```
seberg, line 54: Above you didn't bother with preserving the array/matrix distinction, but here you do when I don't _think_ it matters for either function?
(i.e. just `csgraph = cupyx.scipy.sparse.csr_array(cscraph)` should be OK)

cupyx/scipy/sparse/_csr.py
```diff
@@ -201,53 +172,53 @@ def __le__(self, other):
     def __ge__(self, other):
         return self._comparison(other, operator.ge, '_ge_')
 
-    def __mul__(self, other):
+    def _as_csr_type(self, result):
+        """Re-wrap a cuSPARSE result to match ``type(self)``."""
+        if type(result) is type(self) or not _is_csr(result):
```
seberg, line 177: Do you need that `or not _is_csr(result):` it feels a bit shady to me, since when that happens we may need to still make sure we return the right type (matrix vs. array)?

tests/cupyx_tests/scipy_tests/sparse_tests/test_sparray.py
```diff
@@ -0,0 +1,895 @@
...
+        A_dense[cupy.arange(n), cupy.arange(n)] = 4
+        A_dense[cupy.arange(n - 1), cupy.arange(1, n)] = 1
+        A_dense[cupy.arange(1, n), cupy.arange(n - 1)] = 1
+        A = sparse.csr_array(A_dense)
+        b = cupy.arange(1, n + 1, dtype=numpy.float64)
+        x = spsolve(A, b)
+        cupy.testing.assert_allclose(A @ x, b, rtol=1e-10)
+
+    def test_spsolve_csr_matrix(self):
+        from cupyx.scipy.sparse.linalg import spsolve
```
seberg, line 886: Why do agents seem to think it's nice to do inline imports in tests?

seberg (commented):
At this point these are mostly random comments but I did a scroll through most things and it looks like good patterns and straight forward mostly.
As I tend to, the tests look always not super concinse, but it may well be fine.

I wonder if we shouldn't just have a `_utils._issparse(..., format_or_formats=None):` function (if avoidance of circular import is needed, just use it via the module) -- i.e. that can take a string or tuple/set of strings.
But... since SciPy doesn't, I don't actually care too much, just seemed like it could normalize and shorten the checks a bit.

Will you manage to split out things a bit?  That may still be nice to do a bit more thorough reviews.
I didn't throw an agent at it yet, because I figured you would anyway, but at least I have a high level idea of things now.

eriknw:
> Will you manage to split out things a bit? That may still be nice to do a bit more thorough reviews.

Yeah, just made the first PR in https://github.com/cupy/cupy/pull/10019 . Sorry for the delay--got sick again 😭 . I hope that PR is mostly mechanical and straightforward. I'll keep the current PR up to date with changes to the split PRs.

Thanks for the initial batch of comments!

eriknw:
This is now fully superseded by #10019 #10030 #10031. Closing

PR #9911, opened 2026-05-07
CI: Use only 1 GPU in Windows runners

### LINKED ISSUE #9896: CI: Reduce GPU count to 1 for Windows runners
Ex:
https://github.com/cupy/cupy/blob/5d5ff56c8f973e6a694b3fd7cf6b4de9b658c718/.pfnci/config.pbtxt#L611
My understanding is that this was a typo and reducing it might help save some costs.

leofang:
Close #9896.

### DIFF
--- a/.pfnci/config.pbtxt
+++ b/.pfnci/config.pbtxt
@@ -28,12 +28,12 @@ configs {
       cpu: 8
       memory: 50
       disk: 30
-      gpu: 2
+      gpu: 1
     }
     time_limit: {
       seconds: 21600
     }
-    environment_variables { key: "GPU" value: "2" }
+    environment_variables { key: "GPU" value: "1" }
     command: ".pfnci/linux/main-flexci.sh cuda120.multi"
   }
 }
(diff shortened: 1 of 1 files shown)


seberg:
Hmmmm, now that I am looking through it.  Unlike the linux CI which has explicitly multi-gpu runs, the windows CI does not: It seems it currently runs 2 GPU tests.

So this makes a lot of sense, but we may lose all windows multi GPU CI coverage.

kmaehashi:
I'd prefer to keep (at least one axis of) GPU=2 setup for Windows. This will ensure e.g. inter-device memory copy work on Windows.

leofang:
Does it make sense if we follow Linux and set up "multi" runners? It would be idea I think, but I cannot add runner backends myself 😅

kmaehashi:
Sounds good, how about `cupy.win.cuda13x.multi`? I can create a project once we agree on the name.

leofang:
Yeah I think adding two multi runners (for latest CUDA 12.x & 13.x) sounds good! 👍

PR #9913, opened 2026-05-08
Fix cupy.kron raising ValueError on empty arrays + refactor and performance improvement

### LINKED ISSUE #9904: cupyx.scipy.linalg.kron raises ValueError on empty arrays (incompatible with scipy)
I noticed `scipy.linalg.kron` handles empty arrays (e.g., shape (n, 0)), but `cupyx.scipy.linalg.kron` raises a ValueError.
Is this incompatibility intentional? Just wanted to check.

For example:
```
import numpy as np
import scipy.linalg
import cupy as cp
from cupyx.scipy import linalg as clinalg

A_sp = np.array([[]])
B_sp = np.array([[1, 2], [3, 4]])
result_sp = scipy.linalg.kron(A_sp, B_sp)
print(result_sp)

A_cp = cp.array([[]]) 
B_cp = cp.array([[1, 2], [3, 4]])    
result_cp = clinalg.kron(A_cp, B_cp)
print(result_cp)
```

eriknw:
The concatenate-based path iterates each intermediate's leading axis; when either operand has size 0 that axis is empty and concatenate_method raises. Short-circuit on empty inputs and return cupy.empty with the shape/dtype numpy.kron would produce.

Fixes #9904

I may have gotten a little excessive with the tests. Also, note that sparse kron behaves properly.

_**edit:**_

After @seberg's feedback, I ended up refactoring `kron` to match numpy's implementation, which is faster and uses less peak memory

### DIFF
--- a/cupy/linalg/_product.py
+++ b/cupy/linalg/_product.py
@@ -500,6 +500,15 @@ def kron(a, b):
             b_shape = (1,) * (a_ndim - b_ndim) + b_shape
             ndim = a_ndim
 
+    if a.size == 0 or b.size == 0:
+        # The concatenate-based path below iterates each intermediate's
+        # leading axis; once that axis has length 0 (the next iteration when
+        # `a` is empty, or the first one for some shape combinations) it
+        # raises. Bypass it for empty inputs.
+        out_shape = tuple(s_a * s_b for s_a, s_b in zip(a_shape, b_shape))
+        return cupy.empty(
+            out_shape, dtype=numpy.result_type(a.dtype, b.dtype))
+
     axis = ndim - 1
     out = _core.tensordot_core(
         a, b, None, a.size, b.size, 1, a_shape + b_shape)
--- a/cupy/linalg/_product.py
+++ b/cupy/linalg/_product.py
@@ -490,32 +490,15 @@ def kron(a, b):
     if a_ndim == 0 or b_ndim == 0:
         return cupy.multiply(a, b)
 
-    ndim = b_ndim
-    a_shape = a.shape
-    b_shape = b.shape
-    if a_ndim != b_ndim:
-        if b_ndim > a_ndim:
-            a_shape = (1,) * (b_ndim - a_ndim) + a_shape
-        else:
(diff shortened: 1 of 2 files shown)


cupy/linalg/_product.py
```diff
@@ -500,6 +500,15 @@ def kron(a, b):
...
             ndim = a_ndim
 
+    if a.size == 0 or b.size == 0:
+        # The concatenate-based path below iterates each intermediate's
+        # leading axis; once that axis has length 0 (the next iteration when
+        # `a` is empty, or the first one for some shape combinations) it
+        # raises. Bypass it for empty inputs.
+        out_shape = tuple(s_a * s_b for s_a, s_b in zip(a_shape, b_shape))
+        return cupy.empty(
+            out_shape, dtype=numpy.result_type(a.dtype, b.dtype))
```
seberg: Puh, this is hard to reason about, so maybe this is fine as is...

Sorry, I got nerd-sniped:  I think in principle the `concatenate_method` is a bit weird.  It should be possible to do some form of:
```
out = out.transpose(...).reshape(out_shape)
```
and avoid both the multiple copies (for each `ndim` iteration) and solve the empty array issue.

eriknw: your instinct was right, the concatenation was awkward and hard to think about (as it was also awkward to describe). The initial changes I made in this PR were to put a bandaid on it.

I played with the approach you were sketching, and also the numpy implementation. The NumPy approach is faster and uses less peak memory, so I went with it. I think it's pretty clean now, and we get handling of empty arrays "for free".

seberg (approved):
Perfect, now _this_ is much nicer.  Glancing, I wasn't sure if tensordot use might be faster for some reason, but I suppose for kron it shouldn't really matter.

If it seems faster and less memory use to you, then this is great (and much simpler!).

Thanks!

(EDIT: Ahh, after thinking about it, I guess this path never has to do a copy, a transpose+reshape would do another copy, so even if the tensordot is might be faster that would kill perf.)

PR #9915, opened 2026-05-08
perf: int8 matmul via cublasGemmEx with Tensor Cores (fixes #6611)

### LINKED ISSUE #6611: CuPy int8 matmul takes much longer time than float32 and float16 on GTX 3090

devferrazZz:
## Problem

Fixes #6611. `int8` matmul on a 3090 was **~2.4x slower than fp32 and ~6.9x slower than fp16** despite int8 being the natural Tensor Core dtype on Turing/Ampere.

## Why #6703 is not sufficient

PR #6703 (merged 2022-06-15) improved the existing MAGMA-derived custom CUDA kernel by tuning `BLK_M`/`BLK_N`/`DIM_X`/`DIM_Y`. That is a +7/-7 line change that does **not** introduce Tensor Cores or cuBLAS. Benchmark on a T4 (sm_75) against upstream `main` (which already includes #6703):

| Shape         | upstream int8 vs fp16 | this branch vs upstream int8 |
|---------------|-----------------------|------------------------------|
| 1024×1024×1024 | **16.81×** slower    | 3.15× faster                 |
| 3000×3000×3000 | **8.55×** slower     | 5.17× faster                 |
| 4096×4096×4096 | **10.55×** slower    | 6.08× faster                 |

After this PR, int8 is **1.65–1.77× slower than fp16** (cuBLAS int8 Tensor Core throughput is inherently lower than fp16 due to narrower accumulation), which is the expected physical ceiling on T4.

## What this PR does

- Routes `int8` 2-D matmul through **`cublasGemmEx`** with `CUBLAS_COMPUTE_32I` and `CUDA_R_8I` operands, using the IMMA Tensor Core path on sm_75+ and integer accumulation on sm_61–sm_74.
- **Alignment guard**: cuBLAS IMMA requires `k % 4 == 0`. When `k` is unaligned, a `PerformanceWarning` is emitted and we fall back to the existing custom CUDA kernel. Same fallback is taken if `cublasGemmEx` raises `CUBLAS_STATUS_NOT_SUPPORTED` at runtime.
- **SM gating**: the cuBLAS path is entered only on NVIDIA hardware with compute capability ≥ 6.1 (minimum for `CUBLAS_COMPUTE_32I`). HIP/ROCm skips it unconditionally.
- No change to fp16/fp32/fp64 paths.

### Files changed

| File | Change |
|------|--------|
| `cupy/_core/_routines_linalg.pyx` | int8 cuBLAS dispatch, alignment guard, SM gate, fallback (+74/-4) |
| `cupy/_core/_dtype.pyx` | int8 dtype mapping additions (+4) |
| `cupy_backends/cuda/api/_runtime_enum.pxd` | runtime enum additions (+2) |
| `tests/cupy_tests/math_tests/test_matmul.py` | `TestMatmulInt8`, `TestMatmulInt8SetComputeTypeError` (+93) |
| `tests/cupy_tests/linalg_tests/test_product.py` | `TestInt8Tensordot` (+26) |

## Test coverage

- IMMA-aligned and unaligned `k` shapes (operator `@` and `tensordot`)
- Transposed (Fortran-order) inputs exercising `_mat_to_cublas_contiguous`
- int32→int8 overflow wraps mod 256, matching NumPy semantics
- `PerformanceWarning` emitted before fallback when `cublasGemmEx` is mocked to raise `CUBLAS_STATUS_NOT_SUPPORTED`
- `set_compute_type` guard that rejects non-DEFAULT compute types for integer dtypes

CI: https://github.com/devferrazZz/cupy/actions/runs/25559441581


seberg:
Thanks, interesting that we can just add this!  For reference, cuBLAS documents int8 via int32 all the way back to cuda toolkit 12.0: https://docs.nvidia.com/cuda/archive/12.0.0/cublas/index.html#cublasgemmex so feature wise, we should be on the safe side.
I suppose we do need the larger intermediate int32 result, but usually that is probably a good trade-off.

One thing that I would like to test is what happens for the case where even the intermediate int32 overflows (mainly because there is an int32 and int32_pedantic mode and I wonder what the difference is) -- although I am not sure if it matters much in practice.

(We'll have to have a closer look, for example, I am not sure we should give a performance warning, although that is mainly because I am not 100% familiar with the prior art.)

devferrazZz:
Thanks for the review!

**int32 overflow** — you're right that the int32 accumulator can overflow in principle. For int8×int8 each product is bounded by 127×128 = 16,256, so worst-case overflow only kicks in around k ≈ 132,100. The patch uses `CUBLAS_COMPUTE_32I` (non-pedantic), mirroring how CuPy already defaults float paths to `CUBLAS_COMPUTE_32F` / `_64F` rather than their `_PEDANTIC` variants (`_core/_routines_linalg.pyx:791`). Exposing `CUBLAS_COMPUTE_32I_PEDANTIC` via `cupy.cuda.device.set_compute_type()` is doable — it would mean relaxing the guard I added that currently restricts integer dtypes to `COMPUTE_TYPE_DEFAULT`. Happy to do that in this PR or a follow-up if you'd prefer it gated behind an explicit opt-in.

**performance warning** — to clarify, the `PerformanceWarning` only fires on the *fallback* path (k not divisible by 4, or `CUBLAS_STATUS_NOT_SUPPORTED` at runtime); the cuBLAS Tensor Core path is silent. I modeled it on `cupy/_core/core.pyx:3098`, where CuPy warns when pinned-memory allocation fails and it falls back to synchronous transfer — same pattern of "tried the fast path, had to fall back, here's why." Happy to drop it if you find it noisy.

**CUDA support** — confirmed, `cublasGemmEx` with `CUDA_R_8I` operands is documented in CUDA 12.0 (your link) and earlier versions back to at least CUDA 11.0, so feature availability across supported CUDA versions should be fine.

emcastillo (approved):
Take this with a grain of salt.

Skimmed over the PR and didn't see anything wrong with it or potential issues. I think it is great and the correct approach. The only "suggestion" I have would be to mock the cublas call in the test to see if the function is being called correctly.

leofang:
@devferrazZz could you check the test failures?

devferrazZz:
@leofang thanks — dug into the failures. Three distinct root causes, all fixed in `8419aad`:

1. **NumPy 2.x dtype `.name`** — `to_compute_type_index` read `dtype.name` directly, but in NumPy 2.x the scalar dtype types (`numpy.int8` etc.) no longer expose `.name`; only a `numpy.dtype` instance does. Wrapped it as `numpy.dtype(dtype).name`, mirroring the existing `numpy.dtype(dtype).char` two lines above. This is what broke the `cuda12x`/`cuda13x`/`cuda12x-cuda-python` and Windows runs.
2. **`_integral_tensordot_core` arg type** — it expects the dtype *char code*, but `tensordot_core` was passing the dtype object, so Cython raised `TypeError`. Now passing `dtype.char` at all three call sites.
3. **`py314t` thread-safety** — `test_fallback_emits_performance_warning` patches `cublas.gemmEx` via `mock.patch.object` on a shared symbol, which races under `pytest-run-parallel`. Marked it `@pytest.mark.thread_unsafe` so the runner serializes it; the other int8 tests still run in parallel.

Pre-validated all 21 int8 tests green on CUDA 12, CUDA 13, and free-threaded 3.14t (the last under `--parallel-threads=4`, confirming the marker forces serial execution of the mocked test). Ready for another `/test mini` whenever you have a slot.

devferrazZz:
@leofang is any pending actions from my side to land this?

PR #9916, opened 2026-05-09
Find static libs under conda's targets/<target>/lib path

eriknw:
I started having issues building locally in conda environments after #9801 was merged.

In conda CUDA 12+ envs, static libs (e.g. libcufilt.a) live only under targets/<target>/lib/. After PR #9801 began linking libcufilt.a, source builds in conda CUDA 12+ envs fail at setup time with "Could not find libcufilt.a".

Append targets/<target>/lib to the libdir search list. System-CUDA (lib64/) and conda-build (CUDA_PATH=$PREFIX/targets/<target>) layouts resolve via earlier libdirs and are unaffected.

### DIFF
--- a/install/cupy_builder/cupy_setup_build.py
+++ b/install/cupy_builder/cupy_setup_build.py
@@ -275,7 +275,11 @@ def _find_static_library(name: str) -> str:
             cuda_path = (f'{build.get_cuda_path()}/targets/'
                          f'{build.conda_get_target_name()}/')
         else:
-            libdirs = ['lib64', 'lib']
+            # On conda CUDA 12+, shared libs are symlinked into lib/ but
+            # static libs (e.g. libcufilt.a) live only under
+            # targets/<target>/lib/.
+            libdirs = ['lib64', 'lib',
+                       f'targets/{build.conda_get_target_name()}/lib']
             cuda_path = build.get_cuda_path()
     elif PLATFORM_WIN32:
         filename = f'{name}.lib'


install/cupy_builder/cupy_setup_build.py
```diff
@@ -275,7 +275,11 @@ def _find_static_library(name: str) -> str:
             cuda_path = (f'{build.get_cuda_path()}/targets/'
                          f'{build.conda_get_target_name()}/')
         else:
-            libdirs = ['lib64', 'lib']
+            # On conda CUDA 12+, shared libs are symlinked into lib/ but
+            # static libs (e.g. libcufilt.a) live only under
+            # targets/<target>/lib/.
+            libdirs = ['lib64', 'lib',
+                       f'targets/{build.conda_get_target_name()}/lib']
```
leofang, line 282: I think we need to guard against conda-build. Let's use the same check that is already used in the build system:
```suggestion
            # On conda CUDA 12+, shared libs are symlinked into lib/ but
            # static libs (e.g. libcufilt.a) live only under
            # targets/<target>/lib/.
            libdirs = ['lib64', 'lib']
            if os.environ.get('CONDA_OVERRIDE_CUDA', '0').startswith(('12', '13')):
                libdirs.append(f'targets/{build.conda_get_target_name()}/lib')
```

eriknw: Isn't `CONDA_OVERRIDE_CUDA` only set when building with conda-build? The issue I encountered is for building cupy _in_ a regular conda environment, which your suggested guard _prevents_. Should I be doing something different?

leofang: Hmmm yeah. 

My suggestion is still needed [when we release on conda-forge](https://github.com/conda-forge/cupy-feedstock/tree/main/recipe), though. The question is why you needed this locally. On my system I see
```shell
$CONDA_PREFIX/targets/x86_64-linux/lib/libcufilt.a
```
and in fact #9801 was developed in a conda env as well. I think the existing logic should already cover it if you set 
```shell
export CUDA_PATH=$CONDA_PREFIX/targets/x86_64-linux/
```
at CuPy build time. In fact, local build cannot succeed if this is not set, and perhaps we should document this. Could you check?

eriknw: Ah, I've been using `export CUDA_PATH=$CONDA_PREFIX`, which was previously working for me (but maybe it shouldn't be?).

Your suggestions works _only if_ I _also_ set `NVCC`, e.g.:
```bash
export CUDA_PATH="${CONDA_PREFIX}/targets/${cuda_target}/"
export NVCC="${CONDA_PREFIX}/bin/nvcc"
```
i'm happy to treat this as a learning exercise and to close the PR. I do wonder whether documentation could be better.

PR #9920, opened 2026-05-11
Document env vars for source builds in Conda envs

eriknw:
As learned in #9916, add two notes to "Installing CuPy from Source":

- Conda envs need CUDA_PATH=$CONDA_PREFIX/targets/<target>/ and NVCC pointing at the real binary; otherwise the build fails with "fatal error: cuda_runtime.h: No such file or directory" because Conda packages place headers/static libs under targets/<target>/ and invoking nvcc through the symlinked $CUDA_PATH/bin/nvcc breaks its header lookup.

- Recommend CUPY_NVCC_GENERATE_CODE=current and the parallelism vars (CUPY_NUM_BUILD_JOBS, CUPY_NUM_NVCC_THREADS) to avoid the slow default of compiling for every supported architecture.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -194,6 +194,38 @@ If you want to install the latest development version of CuPy from a cloned Git 
    Cython 3 is required to build CuPy from source.
    It will be automatically installed during the build process if not available.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


seberg (approved):
Thanks good tips/notes!

/test skip

PR #9923, opened 2026-05-13
Use cuSPARSE type aliases consistently

yangcal:
## Summary

This PR cleans up the `cuSPARSE` Cython bindings by using CuPy's type aliases consistently.

Currently, the cuSPARSE bindings mix local aliases such as `Handle`, `Status`, and `SpMatDescr` with the original cuSPARSE type names such as `cusparseHandle_t`, `cusparseStatus_t`, and `cusparseSpMatDescr_t` with duplications. This PR updates the cuSPARSE bindings to use the aliases consistently, matching the style used by the other CUDA library binding modules.


seberg (approved):
Thanks, I was initially a bit unsure if this is public API, but it's clearly not, we don't even ship the `pxd` files.

Good to consolidate this!

PR #9926, opened 2026-05-14
Add missing parameters to reduction functions (NumPy 1.22)

### LINKED ISSUE #6326: Add missing parameters to the nan<x> functions
Rel. #6198.

NumPy 1.22 added missing parameters for some `nan<x>` functions.

> A number of the nan<x> functions previously lacked parameters that were present in their <x>-based counterpart, e.g. the where parameter was present in numpy.mean but absent from numpy.nanmean.

https://numpy.org/devdocs/release/1.22.0-notes.html#missing-parameters-have-been-added-to-the-nan-x-functions

- [ ] nanmin: initial & where
- [ ] nanmax: initial & where
- [ ] nanargmin: keepdims & out
- [ ] nanargmax: keepdims & out
- [ ] nansum: initial & where
- [ ] nanprod: initial & where
- [ ] nanmean: where
- [ ] nanvar: where
- [ ] nanstd: where

subhramit:
Closes https://github.com/cupy/cupy/issues/6326

Covers both normal and NaN reductions:

- `sum`, `prod`: added `initial`, `where`
- `amin`, `amax`: added `initial`, `where`
- `nanmin`, `nanmax`: added `initial`, `where` (delegating to `amin`/`amax`)
- `nansum`, `nanprod`: added `initial`, `where` (delegating to `sum`/`prod`; note as opposed to the issue: they have natural identities (0 and 1), so `where` does not require `initial`)
- `nanmean`, `nanvar`, `nanstd`: added `where`

Also note that no changes were needed to `nanargmin` and `nanargmax`, as both of them already carry `out` and `keepdims` in their signatures and pass them through to the internal `_statistics` implementations.

 Changes are covered by respective unit tests.


cupy/_math/sumprod.py
```diff
@@ -105,13 +115,29 @@ def nansum(a, axis=None, dtype=None, out=None, keepdims=False):
...
 
-def nanprod(a, axis=None, dtype=None, out=None, keepdims=False):
+    return res
+
+
+def nanprod(a, axis=None, dtype=None, out=None, keepdims=False,
+            initial=None, where=True):
     """Returns the product of an array along given axes treating Not a Numbers
-    (NaNs) as zero.
+    (NaNs) as one.
```
subhramit, line 198: Noticed this so corrected it as well.

subhramit:
@seberg @eriknw hi, apologies for the tag (I was unable to find an explicit `MAINTAINERS` file so went with active contributors), but would it be possible for someone to take a look and see if these changes are alright?

subhramit:
> @seberg @eriknw hi, apologies for the tag (I was unable to find an explicit `MAINTAINERS` file so went with active contributors), but would it be possible for someone to take a look and see if these changes are alright?

cc @takagi @asi1024

seberg:
Overall, this seems fair, but I am not sure that we should really do this before we don't do it for normal reductions.
The thing is that once we implement it for normal reductions (where masking seems more useful to me) then this will probably make this much nicer to implement as well.

subhramit:
> Overall, this seems fair, but I am not sure that we should really do this before we don't do it for normal reductions. The thing is that once we implement it for normal reductions (where masking seems more useful to me) then this will probably make this much nicer to implement as well.

I went through your [comment](https://github.com/cupy/cupy/issues/6326#issuecomment-4612768467) on the issue as well and it makes sense.
So for the scope of this PR, would it be better if we remove `initial` from all nan functions for now, keeping only `where`? Or should this PR be closed altogether? I'm fine with either.

seberg:
I think we should explore both for normal reductions.  FWIW, I am not sure there is much of an argument to implement `where` for NaN functions but not for normal reductions (but happy if someone has one).
I actually see more of an argument for `initial=` being useful, but for `initial=` I would also more like to see some support for normal reductions (at least for min/max).
Now there is a point that the NaN functions need some special handling for `initial` anyway, because `initial=NaN` is slightly funny/different in NumPy.

subhramit:
> I think we should explore both for normal reductions. FWIW, I am not sure there is much of an argument to implement `where` for NaN functions but not for normal reductions (but happy if someone has one). I actually see more of an argument for `initial=` being useful, but for `initial=` I would also more like to see some support for normal reductions (at least for min/max). Now there is a point that the NaN functions need some special handling for `initial` anyway, because `initial=NaN` is slightly funny/different in NumPy.

Okay - makes sense, added support for normal reductions and delegated to them. Updated the PR desc as well.
Can you take a look now?

seberg:
Sorry, if I lose track please do ping me again in a bit.  I have to think about the approaches here...
We don't really support weird reductions, I think, so for `initial=` it may actually be fine to post-process, but it may also work to just fill the output array or so.

For `where=` I am more interested in lowering it into the kernel itself, though.  But with CUB reductions that may be trickier (I am not sure yet).
Maybe have to ask others what they think, i.e. whether we should do the masking path to provide the functionality or it is important that if we provide this we make sure it is as fast as it should be.

subhramit:
> For `where=` I am more interested in lowering it into the kernel itself, though. But with CUB reductions that may be trickier (I am not sure yet).
> Maybe have to ask others what they think, i.e. whether we should do the masking path to provide the functionality or it is important that if we provide this we make sure it is as fast as it should be.

For this PR, would you prefer if changes concerning `where` are reverted?
As it was marked a "good first issue", I did not expect to have to dig into the kernel, but would be happy to do so with some guidance/instructions.
Is there is a documented guide on what to keep in mind when touching the kernel?

subhramit:
> > For `where=` I am more interested in lowering it into the kernel itself, though. But with CUB reductions that may be trickier (I am not sure yet).
> > Maybe have to ask others what they think, i.e. whether we should do the masking path to provide the functionality or it is important that if we provide this we make sure it is as fast as it should be.
> 
> For this PR, would you prefer if changes concerning `where` are reverted?
> As it was marked a "good first issue", I did not expect to have to dig into the kernel, but would be happy to do so with some guidance/instructions.
> Is there is a documented guide on what to keep in mind when touching the kernel?

@seberg gentle ping :)

PR #9928, opened 2026-05-15
Test with and document support for cusparseLt v0.9.1

gdaisukesuzuki:
A small fix following the release of cusparseLt v0.9.1.

- No API change from v0.9.0.

Reference: [Release Notes for cuSPARSELt v0.9.1](https://docs.nvidia.com/cuda/cusparselt/release_notes.html#cusparselt-v0-9-1)


seberg (approved):
LGTM, and simple bump.  We also only test 0.8.1, so it seems OK to only test 0.9.1 now that it is out.

PR #9930, opened 2026-05-17
TST: Work around NumPy 2.4.5 regression in conj().

seberg:
NumPy has a pretty bad regression in `arr.conj()` causing it to use the ufunc path almost always.
The ufunc path has two problems (a) it creates copies, and (b) it has no bool version leading to knock on effects in SciPy.


leofang (approved):

seberg:
Pickable crash again :/, but let me put it in... It's plausible NumPy release 2.4.6 today (and maybe yank 2.4.5) and this would be unnecessary.  But this isn't so bad and red CI isn't pretty :).

/test force-skip

PR #9932, opened 2026-05-19
TST: Assert cupy.linalg.solve throws LinAlgError

acosmicflamingo:
Addresses #9838 

`numpy.linalg.solve` started throwing a `LinAlgError` exception in `2.4.3` when they bumped `scipy-openblas` from `0.3.31.22.0` to `0.3.31.126.1`. OpenBLAS moved the early exit for `NRHS=0` after the `GETRF` call in their `gesv` function. Removing the early-out logic in `cupy` side gives us same behavior, but isn't being caught because exceptions are intentionally being ignored (using `cupyx.errstate(linalg='raise')` introduces synchronization). To ensure numpy parity, let's enable them in a specific test.

Add identity matrix test as suggested by @seberg.


tests/cupy_tests/linalg_tests/test_solve.py
```diff
@@ -63,12 +63,33 @@ def check_shape(self, a_shape, b_shape, error_type):
...
+            a = xp.zeros((3, 3))  # singular
+            b = xp.empty((3, 0))  # nrhs = 0
+            with pytest.raises(numpy.linalg.LinAlgError):
+                if xp == cupy:
+                    # errstate is 'ignore' by default since enabling it causes
+                    # synchronization
+                    with cupyx.errstate(linalg='raise'):
+                        xp.linalg.solve(a, b)
+                else:
+                    xp.linalg.solve(a, b)
```
seberg: I think we can test this directly, even with `numpy_cupy_allclose(accept_error=LinAlgError)` when NumPy is 2.4.3+.  That way you don't have to iterate numpy/cupy explicitly and can just use the same `xp` based code in both (plus the cupyx errstate).

Would be nice to add an explicit test for the non-error case as well maybe (could just test that it returns the same thing as the `eye` version?).

acosmicflamingo: Hah, `accept_error` is perfect; didn't know about that feature! Including the cupy-specific logic in the `xp` flow does clean up the test quite a bit; great suggestion.

I can add a simple test for the non-error case. Something like this what you had in mind?
```python
def test_solve_singular(self, xp):
    a = xp.array([[1, 2], [3, 5]])
    b = xp.array([1, 2])
    return xp.linalg.solve(a, b)
```

acosmicflamingo: @seberg P.S. Pretty neat that you contributed to code that went into Mars' Ingenuity helicopter and Arctic Code Vault!

seberg: We already have one for the non error case (the eye?).  For the error case, don't compare to numpy, since it'll error on newer versions.

acosmicflamingo: Ah, so if I understand you correctly (and correct me if I'm wrong), we have already tested that assertions are expectedly thrown (`test_solve_singular_empty__assert_raises`), for basic `xp.linalg.solve` functionality (`test_solve_singular_empty__identity`), but we are missing actually asserting what `cupy.linalg.solve` should be when errors are suppressed. Given that `cupy.linalg.solve` returns an empty matrix when a is both singular and non-singular, all that should be needed is asserting that the matrices are equal like this:
```python
@testing.with_requires('numpy>=2.4.3')
def test_solve_singular_empty__ignore_assert(self):
    a1 = cupy.zeros((3, 3))  # singular
    b1 = cupy.empty((3, 0))  # nrhs = 0
    c1 = cupy.linalg.solve(a1, b1)

    a2 = cupy.eye(3)  # non-singular
    b2 = cupy.empty((3, 0))  # nrhs = 0
    c2 = cupy.linalg.solve(a2, b2)

    # Need to tweak this since it throws "DeprecationWarning:
    # The truth value of an empty array is ambiguous"
    assert c1 == c2
```
Is this what you mean?

seberg: Yeah, something like that... Or just hard-code the expected array.  There is no numpy requirement on that one.

(We should finalize that DeprecationWarning, but I guess better in 15.0... it must have been around for a while.)

acosmicflamingo: Gotcha, great!

I can push this change if you'd like:
```diff
diff --git a/cupy/_core/core.pyx b/cupy/_core/core.pyx
index 3b1aca806..95e8bbebb 100644
--- a/cupy/_core/core.pyx
+++ b/cupy/_core/core.pyx
@@ -1513,11 +1513,9 @@ cdef class _ndarray_base:
 
     def __nonzero__(self):
         if self.size == 0:
-            msg = ('The truth value of an empty array is ambiguous. Returning '
-                   'False, but in future this will result in an error. Use '
-                   '`array.size > 0` to check that an array is not empty.')
-            warnings.warn(msg, DeprecationWarning)
-            return False
+            msg = ('The truth value of an empty array is ambiguous. '
+                'Use `array.size > 0` to check that an array is not empty.')
+            raise ValueError(msg)
         elif self.size == 1:
             return bool(self.get())
         else:
```

Or did you mean that by 15.0, the message should be updated to give a concrete version when the deprecation actually occurs?

tests/cupy_tests/linalg_tests/test_solve.py
```diff
@@ -63,12 +63,36 @@ def check_shape(self, a_shape, b_shape, error_type):
...
+    @testing.numpy_cupy_allclose(accept_error=numpy.linalg.LinAlgError)
+    def test_solve_singular_empty__assert_raises(self, xp):
+        a = xp.zeros((3, 3))  # singular
+        b = xp.empty((3, 0))  # nrhs = 0
+        # errstate is 'ignore' by default since enabling it causes
+        # synchronization
+        with cupyx.errstate(linalg='raise'):
+            xp.linalg.solve(a, b)
+
+    @testing.with_requires('numpy>=2.4.3')
```
seberg: Sorry, I wasn't clear (and I admit it is a bit of an opinion). Either way, it looks all good though, and I can just finish this off soon (release is about to happen, so after that).

This test doesn't need a `@testing.with_requires('numpy>=2.4.3')`.  And since it covers the same thing as `test_solve_singular_empty` I would just replace `test_solve_singular_empty` with this one (and no NumPy version narrowing).

tests/cupy_tests/linalg_tests/test_solve.py
```diff
@@ -63,12 +63,36 @@ def check_shape(self, a_shape, b_shape, error_type):
...
+    @testing.with_requires('numpy<2.4.3')
     @testing.numpy_cupy_allclose()
     def test_solve_singular_empty(self, xp):
         a = xp.zeros((3, 3))  # singular
         b = xp.empty((3, 0))  # nrhs = 0
-        # LinAlgError("Singular matrix") is not raised
+        return xp.linalg.solve(a, b)
+
+    # TODO: replace above test_solve_singular_empty with test below after
+    # bumping numpy baseline to 2.4 (https://github.com/cupy/cupy/issues/9838)
```
seberg: We can just delete the comment and already live in the future now, I think.

seberg (dismissed):
LGTM, I would just tweak the tests once more so they need no adjustment in the future.  But I could also do a small follow-up for that before merging (probably after the 14.1 release is done).

tests/cupy_tests/linalg_tests/test_solve.py
acosmicflamingo: Done!

tests/cupy_tests/linalg_tests/test_solve.py
acosmicflamingo: No worries! And goes without saying that this can be merged whenever :)

Done!

acosmicflamingo:
@seberg thanks for all the feedback! I incorporated them all in the latest push. The only lowest priority item is whether you think I should tweak the deprecation message so that way in the `14.2` release, it mentions it will be deprecated in `15.0`, or I leave the deprecation warning alone entirely.

seberg:
Thanks, let's just go with this.  If NumPy can get away with it in a bug-fix release and nobody really notices (except us), then I don't think we have reason to worry about doing it in a minor release.

/test mini

acosmicflamingo:
Uh oh, seeing some tests fail :(
```pytest

01:17:53.694520 STDOUT 1440] | 29.55s call     tests/cupyx_tests/scipy_tests/interpolate_tests/test_rgi.py::TestInterpN::test_nonscalar_values_2[quintic] |  
-- | -- | --
01:17:53.694522 STDOUT 1440] | 28.52s call     tests/cupyx_tests/scipy_tests/interpolate_tests/test_rgi.py::TestRegularGridInterpolator::test_nonscalar_values_2[True-quintic] |  
01:17:53.694531 STDOUT 1440] | 28.34s call     tests/cupyx_tests/scipy_tests/interpolate_tests/test_rgi.py::TestRegularGridInterpolator::test_nonscalar_values_2[False-quintic] |  
01:17:53.694551 STDOUT 1440] | 27.19s call     tests/cupy_tests/fft_tests/test_callback.py::Test1dCallbacks::test_irfft_load_store_aux[_param_0_{cb_ver='legacy', n=None, norm=None, shape=(10, 7)}] |  
01:17:53.694562 STDOUT 1440] | 27.08s call     tests/cupy_tests/fft_tests/test_callback.py::Test1dCallbacks::test_rfft_load_store_aux[_param_0_{cb_ver='legacy', n=None, norm=None, shape=(10, 7)}] | 7.8s
01:18:01.505638 STDOUT 1440] | 26.96s call     tests/cupy_tests/fft_tests/test_callback.py::Test1dCallbacks::test_fft_load_aux[_param_0_{cb_ver='legacy', n=None, norm=None, shape=(10, 7)}] |  
01:18:01.505664 STDOUT 1440] | 26.92s call     tests/cupy_tests/fft_tests/test_callback.py::Test1dCallbacks::test_irfft_load_store[_param_0_{cb_ver='legacy', n=None, norm=None, shape=(10, 7)}] |  
01:18:01.505677 STDOUT 1440] | 26.88s call     tests/cupy_tests/fft_tests/test_callback.py::Test1dCallbacks::test_irfft_store[_param_0_{cb_ver='legacy', n=None, norm=None, shape=(10, 7)}] |  
01:18:01.505686 STDOUT 1440] | 26.84s call     tests/cupy_tests/fft_tests/test_callback.py::Test1dCallbacks::test_rfft_load_store[_param_0_{cb_ver='legacy', n=None, norm=None, shape=(10, 7)}] |  
01:18:01.505732 STDOUT 1440] | 26.77s call     tests/cupy_tests/fft_tests/test_callback.py::Test1dCallbacks::test_rfft_load[_param_0_{cb_ver='legacy', n=None, norm=None, shape=(10, 7)}] |  
01:18:01.505754 STDOUT 1440] | =========================== short test summary info ============================ |  
01:18:01.505767 STDOUT 1440] | FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_0_{batched_gesv_limit=None, order='C'}::test_solve |  
01:18:01.505769 STDOUT 1440] | FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_1_{batched_gesv_limit=None, order='F'}::test_solve |  
01:18:01.505772 STDOUT 1440] | FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_2_{batched_gesv_limit=0, order='C'}::test_solve |  
01:18:01.505782 STDOUT 1440] | FAILED cupy_tests/linalg_tests/test_solve.py::TestSolve_param_3_{batched_gesv_limit=0, order='F'}::test_solve |  
01:18:01.505784 STDOUT 1440] | = 4 failed, 101651 passed, 3251 skipped, 1630 deselected, 583 xfailed, 20729 warnings in 2987.32s (0:49:47) =
```
Let me see if I can reproduce locally...

acosmicflamingo:
I have confirmed locally that removing the early-out may have fixed the throwing error issue, but it introduced a regression in those tests above. Given that these tests usually test corner-cases (and I recall seeing a condition somewhere that tested for the lhs matrix size), my intuition told me to just change the early-out logic so it returns if `a.size` is 0, and that did fix the regression. Additionally, I checked to see what OpenBLAS was doing and it turns out they were checking the sizes of both lhs and rhs the whole time: https://github.com/OpenMathLib/OpenBLAS/pull/5663/changes#diff-9d0964a1800813c1b3c40e92dfc586f56235c24ffe41a14fddaaff7876f29b6dL102.

I believe that this below is what the fix should actually be
```diff
diff --git a/cupy/cublas.py b/cupy/cublas.py
index 384635ee0..c3db95758 100644
--- a/cupy/cublas.py
+++ b/cupy/cublas.py
@@ -54,7 +54,7 @@ def batched_gesv(a, b):
             'or (..., M, K)')
 
     dtype, out_dtype = _util.linalg_common_type(a, b)
-    if b.size == 0:
+    if a.size == 0:
         return cupy.empty(b.shape, out_dtype)
 
     if dtype == 'f':
diff --git a/cupy/linalg/_solve.py b/cupy/linalg/_solve.py
index f815dd741..8e4c2a083 100644
--- a/cupy/linalg/_solve.py
+++ b/cupy/linalg/_solve.py
@@ -61,7 +61,7 @@ def solve(a, b):
         return batched_gesv(a, b)
 
     dtype, out_dtype = _util.linalg_common_type(a, b)
-    if b.size == 0:
+    if a.size == 0:
         return cupy.empty(b.shape, out_dtype)
 
     if a.ndim == 2:
```

seberg:
Yeah that makes sense that a 0x0 could create issues for us (and most CPU blas versions are fine with it).

seberg:
I pushed a small test change, I hope you don't mind.  I realized that I would prefer to not rely on NumPy on this one because it is related to the OpenBLAS version here...
Which means if someone else packages NumPy+CuPy, they might have failing CuPy tests if we assume NumPy definitely raises.

Anywho, thanks again, hopefully this is it :).

/test mini

acosmicflamingo:
@seberg don't mind at all; good thinking! Glad I can help in any way 😄

acosmicflamingo:
fyi is there a certain process for backporting changes like this? Don't know if it entails merging main back to v14 or cherry-picking individual patches (I can certainly handle the latter 😉).

seberg:
There is a bot that should auto backport.  CuPy's main branch is 15.0 but the minor releases get new features mostly, so most PRs are backported.  The label will make the bot trigger (there is a tiny chance it complains if things don't apply cleanly only).

I think this is safe enough to backport so I tagged it (in part to keep differences between main and v14 lower also).

seberg:
(but there is something weird with the merge, github thinks this isn't mergable but it looks like it should be and right now I am not sure what is going on...)

seberg (approved):
Ah, forgot a final approving review, hehe...

(And nope... same problem as on other PR, I can't actually merge. -- we'll figure it out, maybe just a github fluke)

acosmicflamingo:
Neat! 🎊

leofang:
> (And nope... same problem as on other PR, I can't actually merge. -- we'll figure it out, maybe just a github fluke)

It took us a while to debug but I think @kmaehashi finally fixed it... let me test on the main branch

acosmicflamingo:
Sorry my PR didn't cleanly merge! Was it the way I merged upstream/main to my branch a few times locally? Wondering if I was something wrong on my end.

leofang:
Not your problem at all. We made a mistake when updating the rulesets 😛

acosmicflamingo:
Phew! Thanks y'all :)

PR #9937, opened 2026-05-20
Tolerate hipRAND MT19937 implementation in test_methods

gpinkert:
tests/cupy_tests/random_tests/test_generator.py::TestRandomState:: test_methods iterates over every cuRAND pseudorandom-generator method and constructs a cupy.random.RandomState from each. hipRAND historically did not implement MT19937 -- the constructor for that method would raise KeyError(1000) (HIPRAND_STATUS_NOT_IMPLEMENTED) -- so the HIP branch of the test was a 'pytest.raises(KeyError)' assertion that specifically checked for that failure and 'continue'd past the 'rs.normal()' call.

The test even forecast its own future failure, with a comment explaining the design choice:

    # We use  here so that we will be able to find it
    # once hipRAND implement MT19937 as the imperative
    # immediately rewinds the control flow and does not run the test.

That day arrived: hipRAND on ROCm 7.x ships an MT19937 implementation, so the constructor now succeeds, the 'pytest.raises' block sees no exception, and the test fails with

    Failed: DID NOT RAISE <class 'KeyError'>

Replace the unconditional 'pytest.raises' with a try/except that accepts EITHER outcome:

  * Old hipRAND -- KeyError(1000) is raised; assert the args match the HIPRAND_STATUS_NOT_IMPLEMENTED code we used to see, then continue.
  * New hipRAND -- the constructor succeeds; fall through to rs.normal() and exercise the generator like any other method.

Reproduced the failure in isolation against the installed cupy, then verified the patched test passes against the same hipRAND. The KeyError(1000) compatibility branch can be deleted entirely once cupy drops support for hipRAND versions that lacked MT19937.

### DIFF
--- a/tests/cupy_tests/random_tests/test_generator.py
+++ b/tests/cupy_tests/random_tests/test_generator.py
@@ -140,17 +140,6 @@ class TestRandomState(unittest.TestCase):
         ]
 
         for method in methods:
-            if (runtime.is_hip and
-                    method == cupy.cuda.curand.CURAND_RNG_PSEUDO_MT19937):
-                # hipRAND fails for MT19937 with the status code 1000,
-                # HIPRAND_STATUS_NOT_IMPLEMENTED. We use `pytest.raises` here
-                # so that we will be able to find it once hipRAND implement
-                # MT19937 as the imperative `pytest.xfail` immediately rewinds
-                # the control flow and does not run the test.
-                with pytest.raises(KeyError) as e:
-                    rs = cupy.random.RandomState(method=method)
-                assert e.value.args == (1000,)
-                continue
             rs = cupy.random.RandomState(method=method)
             rs.normal()
 


tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -142,14 +142,15 @@ def test_methods(self):
...
+                # ROCm 7+ implements MT19937; tolerate both old (KeyError) and new.
+                try:
                     rs = cupy.random.RandomState(method=method)
-                assert e.value.args == (1000,)
+                except KeyError as e:
+                    assert e.args == (1000,), (
+                        'expected only HIPRAND_STATUS_NOT_IMPLEMENTED'
+                        f' (1000) on older hipRAND, got {e.args!r}')
+                    continue
+                rs.normal()
```
kmaehashi: I guess it's ok to remove the these lines entirely now that we dropped support for ROCm 6?

kmaehashi: ping @gpinkert

gpinkert: Done

leofang:
@gpinkert could you turn on "allow maintainer to edit" so that we can move forward?

PR #9944, opened 2026-05-20
cuda_workaround.h: enable std::is_signed et al. on HIPRTC

gpinkert:
NVRTC and HIPRTC both ship minimal STL: standard type-trait templates (std::is_signed, std::enable_if, std::declval, ...) that CuPy kernels routinely need aren't in scope by default. Each runtime needs a different shim:

  * NVRTC -- already handled. cuda_workaround.h pulls libcu++ in via <cuda/std/type_traits> et al. and re-exports each trait into namespace std via using-declarations, so kernel code can spell `std::is_signed<T>` and have it resolve to cuda::std::is_signed.

  * HIPRTC -- new. Clang's HIP mode picks up libstdc++'s host headers via the system gcc include search paths, but only if something explicitly includes them. Pull <type_traits> and <utility> in from this header so the same `std::is_signed<T>` spelling works on HIPRTC without any cupy-specific namespace at the call site.

Net result: kernel code that includes cuda_workaround.h can use `std::is_signed<T>`, `std::declval<T>()`, etc. uniformly on both backends.

### DIFF
--- a/cupy/_core/include/cupy/hip_workaround.cuh
+++ b/cupy/_core/include/cupy/hip_workaround.cuh
@@ -1,6 +1,15 @@
 #ifndef INCLUDE_GUARD_CUPY_HIP_WORKAROUND_H
 #define INCLUDE_GUARD_CUPY_HIP_WORKAROUND_H
 
+#ifdef __HIPCC_RTC__
+
+// Pull in libstdc++ headers so kernel code can use std::is_signed,
+// std::declval, etc. under HIPRTC just like on NVRTC.
+#include <type_traits>
+#include <utility>
+
+#endif  // __HIPCC_RTC__
+
 #ifdef __HIP_DEVICE_COMPILE__
 
 // Determine whether native __shfl_*_sync functions are available.


kmaehashi (dismissed):

kmaehashi:
LGTM but let me adjust merge timing as this will invalidate all CI cache

leofang (changes requested):
Hi @gpinkert I think this is the wrong file to edit. This should go to `hip_workaround.cuh`, which is meant to be included when targeting the ROCm/HIP backend. If it's not included, it is a bug. In addition, as Kenichi noted touching this file would invalidate all CUDA CI caches, which is not desirable.

gpinkert:
@leofang is it possible to get this fix in without invalidating the cuda CI caches? To be honest, I don't understand what triggers the CI rebuilds, but I also haven't investigated or thought it through (it is almost midnight here and my brain isn't working at high occupancy, too many HBM reads)

gpinkert:
@leofang sorry to bump this. Is the only fix I need to do is put the change in `hip_workaround.cuh`? Or does that also trigger CI cache rebuilds and therefore blocks the PR from merge?

kmaehashi:
The fix should live in `hip_workaround.cuh`. All changes in `cupy/_core/include/` will invalidate the CI cache, but I think this is unavoidable in this case. Cache invalidation itself does not block the merge but will affect overall development velocity (ie, CI for other PRs will take much much longer time), so I'd find the right time to get this in.

gpinkert:
@kmaehashi , does the change look ok now?

leofang (approved):

leofang:
@gpinkert could you turn on "allow maintainer to edit" so that we can move forward?

PR #9949, opened 2026-05-20
test_ndarray_complex_ops: drop identity assert from test_conj_pass

gpinkert:
The failure
-----------
On NumPy 2.x:
  tests/cupy_tests/core_tests/test_ndarray_complex_ops.py
  ::TestConj::test_conj_pass        FAIL  (AssertionError: Only numpy raises)
  ::TestConj::test_conjugate_pass   FAIL  (AssertionError: Only numpy raises)

both with

    assert x is y
        for x, y = real-typed ndarray, x.conj()/x.conjugate()

The values are still identical -- conj() of a real number is the number itself -- so the harness's numpy_cupy_array_almost_equal value check would have passed. The test was specifically asserting *object identity* (`is`, not `==`), and that contract no longer holds on the numpy side.

Identity vs equality
--------------------
`==` asks: "do these two arrays hold the same numbers?" `is` asks: "are these two names pointing to the same Python object /
            the same underlying memory block?"

For x.conj() on a real array, the values are mathematically required to equal x. What changed in NumPy 2.x is the *memory*: it now hands you a fresh allocation with the same numbers copied into it, instead of handing back the original array.

                  +- x --+                        +- x --+    +- y --+
NumPy 1.x / CuPy: |1 2 3 |  <- y also points here |1 2 3 |    |1 2 3 |  NumPy 2.x
                  +------+                        +------+    +------+
                  (one buffer, two names)         (two buffers, same numbers)

Why NumPy 2.x dropped the optimisation
--------------------------------------
Three reasons, all aimed at making ndarray semantics less surprising:

  1. Consistency with the ufunc. numpy.conjugate is a ufunc and ufuncs always allocate (or write into the user-supplied `out=`). The method x.conj() used to take a shortcut and `return self` when the dtype was real -- so the method and the ufunc behaved differently for the same input:

         # NumPy 1.x
         x = np.array([1.0, 2.0])
         x.conj() is x         # True   (method optimisation)
         np.conj(x) is x       # False  (ufunc always allocates)

         # NumPy 2.x
         x.conj() is x         # False  (both paths agree now)
         np.conj(x) is x       # False

  2. Avoiding accidental aliasing bugs. With the optimisation in effect, this was a long-standing footgun:

         x = np.array([1.0, 2.0])
         y = x.conj()
         y[0] = 999     # silently mutates x too, because y IS x

     NumPy 2.x removed the optimisation specifically so .conj() now gives the user a fresh, independent array -- same defensive-copy semantics as .copy() or .astype().

  3. Cleaner NEP-50 type promotion. NumPy 2.0 tightened how dtypes propagate through operations. Several "this used to be a no-op so we returned self" optimisations were retired so that the type-promotion machinery could be the single source of truth. Bool is the most visible casualty (it now promotes to int8 for conj, which is the matching fix in cupy/_core/_routines_math.pyx in this branch), but the identity-preservation removal is the same cleanup applied to the real-typed path.

Why CuPy still preserves identity
---------------------------------
CuPy makes a different tradeoff. On a GPU, allocating a fresh ndarray means:
  * a device memory allocation (mempool hit, or worse, a real malloc
    + cudaMalloc/hipMalloc-class call),
  * a device-to-device copy of the data buffer.

For real dtypes where conj is mathematically the identity, that allocation+copy is pure waste. CuPy keeps the `return self` fast path in cupy._core._routines_math::_ndarray_conj because the GPU cost of always allocating is much higher than the CPU cost numpy was willing to pay (the wider rationale block is in that commit, see the matching fix in this branch's preceding commit).

The risk traded off is the aliasing surprise from (2) above -- CuPy users hitting that get the same footgun numpy users used to. Whether that is the right tradeoff long-term is a real design question we explicitly leave open for now.

What this commit does
---------------------
test_conj_pass / test_conjugate_pass were exercising both numpy and cupy under numpy_cupy_array_almost_equal and then asserting x is y. Under NumPy 2.x the assert fires only on the numpy side, the harness sees "Only numpy raises error", and the test fails despite CuPy still implementing the documented optimisation correctly. The cross-backend comparison the harness was meant to enforce (values agree) wasn't even reached.

Convert both tests to pure-CuPy assertions of the optimisation contract: drop the numpy_cupy_array_almost_equal harness, drop the `xp` parameter, call cupy directly, and keep `assert x is y` as the sole intent of the test (verifying that CuPy's _ndarray_conj fast path returns the input array unchanged for non-complex, non-bool dtypes).

Also exclude bool from the parametrisation via no_bool=True -- the matching commit in this branch (cupy/_core/_routines_math.pyx, "core: bool.conj() promotes to int8 (match NumPy 2.x)") *deliberately* allocates a fresh int8 array for bool inputs, so for bool `assert x is y` correctly does not hold.

Verified locally on ROCm 7.2.0 / gfx90a + NumPy 2.4.5:
  4 passed in 0.89s.

### DIFF
--- a/tests/cupy_tests/core_tests/test_ndarray_complex_ops.py
+++ b/tests/cupy_tests/core_tests/test_ndarray_complex_ops.py
@@ -17,15 +17,14 @@ class TestConj:
             return x  # NumPy 2.4.5 had a regression for bool_arr.conj()
         return x.conj()
 
-    @testing.for_all_dtypes(no_complex=True)
-    @testing.numpy_cupy_array_almost_equal()
-    def test_conj_pass(self, xp, dtype):
-        x = testing.shaped_arange((2, 3), xp, dtype)
+    @testing.for_all_dtypes(no_complex=True, no_bool=True)
+    def test_conj_pass(self, dtype):
+        # CuPy keeps the same-object fast path; numpy 2.x dropped it.
+        x = testing.shaped_arange((2, 3), cupy, dtype)
         y = x.conj()
         if xp is numpy and numpy.__version__ == "2.4.5":
             return x  # NumPy 2.4.5 had a regression for bool_arr.conj()
         assert x is y
-        return y
 
     @testing.for_all_dtypes()
     @testing.numpy_cupy_array_almost_equal()
@@ -35,15 +34,14 @@ class TestConj:
             return x  # NumPy 2.4.5 had a regression for bool_arr.conj()
         return x.conjugate()
 
-    @testing.for_all_dtypes(no_complex=True)
-    @testing.numpy_cupy_array_almost_equal()
-    def test_conjugate_pass(self, xp, dtype):
-        x = testing.shaped_arange((2, 3), xp, dtype)
+    @testing.for_all_dtypes(no_complex=True, no_bool=True)
+    def test_conjugate_pass(self, dtype):
+        # See test_conj_pass.
+        x = testing.shaped_arange((2, 3), cupy, dtype)
         y = x.conjugate()
         if xp is numpy and numpy.__version__ == "2.4.5":
             return x  # NumPy 2.4.5 had a regression for bool_arr.conj()
         assert x is y
-        return y
 
 
 class TestAngle:


seberg:
Closing this one, thanks.  That was an accidental regression in NumPy 2.4.5 which I fixed for 2.4.6 (the CuPy tests also have a skip for 2.4.5 now).

